PR #29, opened 2017-03-23
Mount the container for Copy

nalind:
In Copy(), instead of flagging a container not being mounted as an error, do what we do in Run(), which is to mount it first and then unmount when we're done.

### DIFF
--- a/add.go
+++ b/add.go
@@ -40,13 +40,20 @@ func addUrl(destination, srcurl string) error {
 	return nil
 }
 
-// Add copies contents into the container's root filesystem, optionally
-// extracting contents of local files that look like non-empty archives.
+// Add copies the contents of the specified sources into the container's root
+// filesystem, optionally extracting contents of local files that look like
+// non-empty archives.
 func (b *Builder) Add(destination string, extract bool, source ...string) error {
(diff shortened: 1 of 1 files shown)


rhatdan:
Should we be checking if it is already mounted?  We don't want to umount it if someone else has done a buildah mount?

add.go
```diff
@@ -43,10 +43,16 @@ func addUrl(destination, srcurl string) error {
 // Add copies contents into the container's root filesystem, optionally
```
TomSweeneyRedHat: I might be light on the amount of tea that I've had today, but I'm struggling with this comment.  Should it be "Add the source's contents into the containter's root ..."

nalind:
The storage library keeps refcounts of mounts using information that it keeps under the `--runroot` directory, so we shouldn't have to worry about accidentally unmounting a container that someone else is using, even if the filesystem doesn't look busy.

add.go
nalind: Went with "Add copies the contents of the specified sources".

rhatdan:
:+1:

nalind:
Thanks, merging.

nalind:

PR #30, opened 2017-03-23
Always make sure the working directory exists

nalind:
Always make sure the working directory exists before attempting to run anything inside of it, and before attempting to copy contents into it or one of its subdirectories.

### DIFF
--- a/run.go
+++ b/run.go
@@ -108,6 +108,9 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	if spec.Process.Cwd == "" {
 		spec.Process.Cwd = DefaultWorkingDir
 	}
+	if err := os.MkdirAll(filepath.Join(mountPoint, b.Workdir), 0755); err != nil {
+		return fmt.Errorf("error ensuring working directory %q exists: %v)", b.Workdir, err)
+	}
 	mounts := options.Mounts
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

run.go
```diff
@@ -108,6 +108,9 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	if spec.Process.Cwd == "" {
 		spec.Process.Cwd = DefaultWorkingDir
 	}
+	if err := os.MkdirAll(filepath.Join(mountPoint, b.Workdir), 0755); err != nil {
+		return fmt.Errorf("error ensuring working directory %q exists: %v)", b.Workdir, err)
```
TomSweeneyRedHat: I know it's unlikely we'll ever hit this, but do we need differing verbiage between these two functions so we could tell more easily where the error is if it ever does pop?    Perhaps just changing ' exists:' to ' exists in Run():'.  Not a biggy either way, thought I'd throw it out there for consideration.

nalind: The two hunks in this PR ended up using slightly different errors messages, though it wasn't something I was concerned about at the time.

nalind:
Thanks, merging.

nalind:

PR #31, opened 2017-03-23
Teach "copy" to allow copying to a destination file

nalind:
The current Copy() mplementation assumes that the destination location is (or is intended to be) a directory, because I thought that was how the COPY instruction worked.  On reading example Dockerfiles, it looks like that was incorrect.  Building on #30, this changes Copy() to default to creating a file when the item being copied is a file and the destination name doesn't exist yet, which is more in line with the `cp` command.

### DIFF
--- a/add.go
+++ b/add.go
@@ -55,12 +55,20 @@ func (b *Builder) Add(destination string, extract bool, source ...string) error 
 		}
 		dest = filepath.Join(dest, b.Workdir, destination)
 	}
-	// Make sure the destination is usable.
-	if fi, err := os.Stat(dest); err == nil && !fi.Mode().IsDir() {
-		return fmt.Errorf("%q already exists, but is not a subdirectory)", dest)
+	// Make sure the destination's parent directory is usable.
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

add.go
```diff
@@ -79,11 +93,14 @@ func (b *Builder) Add(destination string, extract bool, source ...string) error
 			return fmt.Errorf("error reading %q: %v", src, err)
 		}
 		if fi.Mode().IsDir() {
```
TomSweeneyRedHat: Ok, personal twitch here.  I'm not at all a fan of one or two character variable names in code, especially so when the short name is part of a regular word like 'fi', and even more so when the variable is used more than 4 or 5 lines away from where it's first declared.  I much prefer longer/clearer variable names as it makes long term maintenance much easier.  I'll now crawl under my rock.

nalind: Renamed it to srcfi because I have bad habits.

nalind:
Thanks, merging.

nalind:

PR #32, opened 2017-03-23
Rename "list" and "delete" to "containers" and "rm"; add "images" and "rmi" commands

nalind:
Add some minimal image management commands.  Neither is really in scope for a building tool, but it reduces the number of use cases where we need additional tools.

### DIFF
--- a/cmd/buildah/images.go
+++ b/cmd/buildah/images.go
@@ -1,0 +1,69 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/urfave/cli"
+)
+
+var (
+	imagesFlags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "quiet, q",
+			Usage: "display only container IDs",
+		},
+		cli.BoolFlag{
+			Name:  "noheading, n",
+			Usage: "do not print column headings",
+		},
+	}
+	imagesDescription = "Lists locally stored images."
+	imagesCommand     = cli.Command{
+		Name:        "images",
+		Usage:       "List images in local storage",
+		Description: imagesDescription,
+		Flags:       imagesFlags,
+		Action:      imagesCmd,
+		ArgsUsage:   " ",
+	}
+)
+
+func imagesCmd(c *cli.Context) error {
+	store, err := getStore(c)
+	if err != nil {
+		return err
+	}
+
+	quiet := false
+	if c.IsSet("quiet") {
+		quiet = c.Bool("quiet")
+	}
+	noheading := false
+	if c.IsSet("noheading") {
+		noheading = c.Bool("noheading")
+	}
+
+	images, err := store.Images()
+	if err != nil {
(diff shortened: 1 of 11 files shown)


cmd/buildah/main.go
```diff
@@ -68,6 +68,14 @@ func main() {
 	}
 	app.Commands = []cli.Command{
 		{
+			Name:        "images",
```
TomSweeneyRedHat: Thought this was going to be list?

cmd/buildah/images.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+var (
+	imagesFlags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "quiet",
+			Usage: "omit column headings",
+		},
+		cli.BoolFlag{
+			Name:  "qq",
```
TomSweeneyRedHat: Depending on the online context, qq means 'crying eyes' or 'just quit the game as you've no skills'.  Perhaps --onlyIDs  or --short or --min?

cmd/buildah/main.go
nalind: We're already using "list" for listing containers that buildah is working with, though I could be swayed to "list-images" to parallel with "delete-image".

cmd/buildah/images.go
nalind: Oh.  I was going for something along the lines of 'docker images -q', but am not opposed to renaming the flag.  Any preference?

TomSweeneyRedHat: At first I thought just --q would be good, but then remembered docker images uses --quiet or -q to only show the ids.  Can I vote to change the flags to --omit for the column headings and --quiet for the image ids only?

cmd/buildah/main.go
TomSweeneyRedHat: My bad, ignore this.  I'd forgotten list was for containers.  Images is good.

cmd/buildah/images.go
nalind: The --quiet flag is what we already for headings in "list", that'll need to be changed there to keep them consistent.  Updating.

rhatdan:
I would prefer not to add additional verbs, just use buildah list and buildah delete.

TomSweeneyRedHat:
LGTM.  For images docker uses 'images' for list and 'rmi' to remove.  For containers 'ps' and 'rm'.  Should we emulate that or do you think the code should recognize the object and do the right thing?

nalind:
We can overload delete so that it looks for a container first, but with free-form names around, we're inevitably going to need some way to specify what type of thing was intended.  We have distinctly different things to display for images and containers, and every output format I can come up with that encompasses both just looks messier if you only care about one or the other.

rhatdan:
Can't we just force user to specify --image or --container when their is conflict.

rhatdan:
@rh-atomic-bot r+  	0e702d6

PR #33, opened 2017-03-23
Strip the image name down more for container names

nalind:
When computing a default container name from the name of a source image, drop tags and any leading components from the image name before attempting to use it as part of a container name.

### DIFF
--- a/new.go
+++ b/new.go
@@ -2,6 +2,7 @@ package buildah
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/Sirupsen/logrus"
 	is "github.com/containers/image/storage"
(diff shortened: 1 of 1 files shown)


new.go
```diff
@@ -28,7 +29,20 @@ func newBuilder(store storage.Store, options BuilderOptions) (*Builder, error) {
...
+				prefix = s[0]
+			}
+			s = strings.Split(prefix, "@")
+			if len(s) > 0 {
+				prefix = s[0]
+			}
+			name = prefix + "-" + name
 		}
 	}
 	if name != "" {
```
TomSweeneyRedHat: Is working-container a possibly valid value for a container name?  Do we want to error out in an else here if it is still working-container and/or set name initially to "" at line 23 and error out if there's no name?  I'm not sure, putting it up for consideration.

nalind: If we end up not passing in a name, the container gets created anyway, but can only be referred to by its ID.  This bit of logic is mainly about providing a friendly default if one isn't specified.

TomSweeneyRedHat: ok, thx for the 411.

nalind: Though in the case of "scratch", we could as easily decide to keep that as a prefix for a name that we generate.

rhatdan:
@rh-atomic-bot r+ 59d508d

PR #34, opened 2017-03-23
Give Run() commands access to the host's network

nalind:
These changes put commands that we Run() in the host's network namespace, and defaults to giving it a read-only bind mounted copy of /etc/hosts and /etc/resolv.conf.

### DIFF
--- a/run.go
+++ b/run.go
@@ -104,6 +104,7 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	}()
 	g.SetRootPath(mountPoint)
 	g.SetProcessTerminal(true)
+	g.RemoveLinuxNamespace("network")
 	spec := g.Spec()
 	if spec.Process.Cwd == "" {
 		spec.Process.Cwd = DefaultWorkingDir
--- a/run.go
+++ b/run.go
@@ -113,6 +113,22 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 		return fmt.Errorf("error ensuring working directory %q exists: %v)", b.Workdir, err)
 	}
 	mounts := options.Mounts
+	boundMounts := []specs.Mount{}
+	for _, boundFile := range []string{"/etc/hosts", "/etc/resolv.conf"} {
+		for _, mount := range mounts {
+			if mount.Destination == boundFile {
+				// Already have an override for it, so skip this one.
+				continue
+			}
+		}
+		boundMount := specs.Mount{
+			Source:      boundFile,
+			Destination: boundFile,
+			Type:        "bind",
+			Options:     []string{"rbind", "ro"},
+		}
+		boundMounts = append(boundMounts, boundMount)
+	}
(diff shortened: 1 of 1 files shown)


run.go
```diff
@@ -97,18 +97,39 @@ func (b *Builder) Run(command []string, options RunOptions) error {
...
 	mounts := options.Mounts
+	for _, boundFile := range []string{"/etc/hosts", "/etc/resolv.conf"} {
+		override := false
+		for _, mount := range mounts {
+			if mount.Destination == boundFile {
+				// Already have an override for it, so skip this one.
+				continue
+			}
+		}
+		if override {
```
TomSweeneyRedHat: Unless I've been staring at the screen too long today, I'm not seeing how override could be true here.  Do we need this if or even the boolean?

run.go
```diff
@@ -97,18 +97,39 @@ func (b *Builder) Run(command []string, options RunOptions) error {
...
+		boundMount := specs.Mount{
+			Source:      boundFile,
+			Destination: boundFile,
+			Type:        "bind",
+			Options:     []string{"rbind", "ro"},
+		}
+		boundMounts = append(boundMounts, boundMount)
+	}
 	for _, specMount := range spec.Mounts {
 		override := false
```
TomSweeneyRedHat: I'm not seeing where override is being used after this line.

run.go
```diff
@@ -97,18 +97,39 @@ func (b *Builder) Run(command []string, options RunOptions) error {
...
+				continue
+			}
+		}
+		if override {
+			continue
+		}
+		boundMount := specs.Mount{
+			Source:      boundFile,
+			Destination: boundFile,
+			Type:        "bind",
```
TomSweeneyRedHat: Again maybe staring too much at the screen, but shouldn't Source: be set to 'mount' here rather than boundFile?  Aren't you just adding hosts and resolv.conf in over and over and nothing else?

run.go
nalind: Doh.  Yeah, that's a copy/paste error.  Removing it.

run.go
nalind: Source is the path on the host that we want to bind into the container.  In this case we're defaulting to using read-only bind mounts of the host's `/etc/hosts` and `/etc/resolv.conf`, refraining from doing so for either of them if any of the `Mount` values passed in via `options` also specifies the same destination.

run.go
nalind: It wasn't.  Removing it.

nalind:
Updated.

nalind:
@rh-atomic-bot r+ 91958d4fb674c21e2215b4dfa54161357d7f59ea

rhatdan:
@rh-atomic-bot r+ 743989c

rhatdan:
@rh-atomic-bot r+ 7eee83d

rhatdan:
@rh-atomic-bot retry

nalind:

PR #39, opened 2017-03-24
Cleanup `buildah from` CLI

rhatdan:
Remove --image since image is an argument
Remove --mount, since I don't see a case where you would not want to do this.

### DIFF
--- a/tests/basic.bats
+++ b/tests/basic.bats
@@ -3,42 +3,43 @@
 load helpers
 
 @test "from" {
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json --image alpine)
+  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
   buildah delete $cid
   cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json         alpine)
   buildah delete $cid
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json         alpine  i-love-naming-things)
+  cid=$(buildah from alpine --pull --signature-policy ${TESTSDIR}/policy.json --name i-love-naming-things)
   buildah delete i-love-naming-things
 }
 
 @test "from-defaultpull" {
-  cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json --image alpine)
+  cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json alpine)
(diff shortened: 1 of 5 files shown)


nalind:
This also appears to be removing the option of specifying a container's name as a second argument?

rhatdan:
@nalind Yes, you can do this via the --name option.  In our proposal we said the second argument is for mountpoint, but I am not sure that is a great idea.

Might rather just do

```
$ ctr=$(buildah from fedora)
$ buildah mount $ctr /mnt
$ dnf install httpd --installroot /mnt
```

rhatdan:
I added back in the --mount option since it was blowing up the tests.  I think we need to consider how this should work.

If `builder from` Always mounts the image, then we would want 

```rootfs=$(builder mount $cid)```
To succeed if the $cid is mounted and return the rootfs, as well as mount the object and return the rootfs.

File: tests/basic.bats [hunk 1]
```diff
@@ -3,42 +3,43 @@
 load helpers
 
 @test "from" {
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json --image alpine)
+  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
   buildah delete $cid
   cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json         alpine)
   buildah delete $cid
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json         alpine  i-love-naming-things)
+  cid=$(buildah from alpine --pull --signature-policy ${TESTSDIR}/policy.json --name i-love-naming-things)
```
TomSweeneyRedHat: Will this work with alpine in the first position? I see alpine in the last position in other lines below.  It's fine if it does, just a bit unusual.

nalind:
[[Buildah] (8) buildah CLI overhaul](https://trello.com/c/LULizZlk/442-buildah-8-buildah-cli-overhaul)

tests/basic.bats
rhatdan: It seems to work.  Only problem I have found that does not work is

OPTIONS ARGS OPTIONS

If the options are all after or before the args CLI seems to work

nalind:
LGTM, pending CI tests.

nalind:
@rh-atomic-bot r+ b8ab8aef63b24790e45cbca8a2c1e56b9a329512

PR #40, opened 2017-03-24
Remove --destination flag from `buildah add|copy`

rhatdan:
We need to match the syntax of Dockerfile, three forms

buildah copy $CID SOURCE
buildah copy $CID SOURCE DESTINATION
buildah copy $CID SOURCE SOURCE SOURCE DESTINATION


nalind:
LGTM.  If we aren't checking that specifying what would be a file as a destination produces an error when we try to copy more than one thing to the container, we should add a test for that.

rhatdan:
Added tests.

nalind:
Merging.

@rh-atomic-bot r+ 37fd317abac0519d2d51a4468bdff12803282d8d

nalind:
[[Buildah] (8) buildah CLI overhaul](https://trello.com/c/LULizZlk/442-buildah-8-buildah-cli-overhaul)

PR #41, opened 2017-03-24
Only attempt to shut down a Store if we opened one

nalind:
Only attempt to Shutdown() the default storage.Store if we actually opened it to begin with.  This avoids some unnecessary churn with storage in cases where argument parsing flags an invocation error.


rhatdan:
@rh-atomic-bot r+ 5152a04

nalind:

PR #44, opened 2017-03-25
Add some obvious short options

rhatdan:
I chose to only add traditional ones like

--quiet, -q
and 
--user -u
And for some which could be specified multiple times I added 
-v, -e, -l

### DIFF
--- a/cmd/buildah/list.go
+++ b/cmd/buildah/list.go
@@ -12,7 +12,11 @@ var (
 	listFlags = []cli.Flag{
 		cli.BoolFlag{
 			Name:  "quiet, q",
-			Usage: "omit column headings",
+			Usage: "list only container image id's",
+		},
+		cli.BoolFlag{
+			Name:  "noheading, n",
+			Usage: "do not print column headings",
 		},
 	}
 	listDescription = "Lists containers which appear to be " + buildah.Package + " working containers, their\n   names and IDs, and the names and IDs of the images from which they were\n   initialized"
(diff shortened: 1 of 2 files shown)


cmd/buildah/list.go
```diff
@@ -10,7 +10,7 @@ import (
 var (
 	listFlags = []cli.Flag{
 		cli.BoolFlag{
-			Name:  "quiet",
+			Name:  "quiet, q",
```
TomSweeneyRedHat: Thought we were doing 'omit' here to kill the column heading and 'quiet' to show just the id column.  fwiw, docker ps --quiet caused it to display on the numeric ids.

rhatdan: That sounds good to me.  Basically you want to add an option though correct, since there will still be a quiet option.  You just want to change the meaning.

TomSweeneyRedHat: Yes, I'd like to add an "omit" option here that would keep the table headers from printing when used.  "quiet" would be used to show just the image id's, much like docker does.  My main confusion here is I'd thought Nalin had already done this, but perhaps it wasn't merged?

rhatdan:
@TomSweeneyRedHat I added a --noheading switch and changed quiet to work the way you wanted.

cmd/buildah/list.go
nalind: In #32, yes.

rhatdan: Well lets do it here or in a separate pull request.

nalind:
LGTM.

rhatdan:
@rh-atomic-bot r+  	e5fe19c

PR #45, opened 2017-03-27
Move all command definitions out of main.go into implementation source

rhatdan:
This will allow us to more easily share these source modules between
kpod and buildah as well as move keep the definition with the implementation.


cmd/buildah/mount.go
```diff
@@ -8,6 +8,13 @@ import (
 
 var (
 	mountDescription = "Mounts a working container's root filesystem for manipulation"
+	mountCommand     = cli.Command{
+		Name:        "mount",
+		Usage:       "Mount a working container's filesystem root",
```
TomSweeneyRedHat: should this be "root filesystem" like line 10?

cmd/buildah/umount.go
```diff
@@ -8,6 +8,14 @@ import (
 
 var (
 	umountDescription = "Unmounts a working container's root filesystem"
+	umountCommand     = cli.Command{
+		Name:        "umount",
+		Aliases:     []string{"unmount"},
+		Usage:       "Unmount a working container's filesystem root",
```
TomSweeneyRedHat: should this be root filesystem like line 10 too?

cmd/buildah/main.go
```diff
@@ -69,82 +69,16 @@ func main() {
...
-			ArgsUsage:   "CONTAINER-NAME-OR-ID IMAGE",
-		},
-		{
-			Name:        "delete",
-			Usage:       "Delete working container(s)",
-			Description: deleteDescription,
-			Action:      deleteCmd,
-			ArgsUsage:   "CONTAINER-NAME-OR-ID [...]",
-		},
+		addCommand,
```
TomSweeneyRedHat: Do you need an import for these commands as they're in other source files?  I'm not seeing at least a '/cmd' in the imports and thought there'd be one (but may be showing my lack of go xp).

rhatdan: As long as they are in the same package, these fields are global.

TomSweeneyRedHat: kk, thx for the 411.

rhatdan:
@rh-atomic-bot r+ ad2d49b

PR #47, opened 2017-03-27
Add support for images as well as containers to buildah list & delete

rhatdan:
This patch set will cleanup the output of buildah list adding truncate
flags as well as support images.

### DIFF
--- a/cmd/buildah/delete.go
+++ b/cmd/buildah/delete.go
@@ -7,10 +7,10 @@ import (
 )
 
 var (
-	deleteDescription = "Deletes working container(s), unmounting them if necessary"
+	deleteDescription = "Deletes images and working container(s), unmounting them if necessary"
 	deleteCommand     = cli.Command{
 		Name:        "delete",
-		Usage:       "Deletes working container(s)",
+		Usage:       "Deletes images and working container(s)",
 		Description: deleteDescription,
 		Action:      deleteCmd,
 		ArgsUsage:   "CONTAINER-NAME-OR-ID [...]",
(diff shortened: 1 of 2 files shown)


rhatdan:
Alternative to https://github.com/projectatomic/buildah/pull/32

cmd/buildah/delete.go
```diff
@@ -30,12 +30,16 @@ func deleteCmd(c *cli.Context) error {
 	for _, name := range args {
 		builder, err := openBuilder(store, name)
 		if err != nil {
-			return fmt.Errorf("error reading build container %q: %v", name, err)
+			/* Maybe this was an image attempt to delete it */
+			_, err = store.DeleteImage(name, true)
+			return fmt.Errorf("error deleting image or container container %q: %v", name, err)
```
TomSweeneyRedHat: "container container"?

cmd/buildah/list.go
```diff
@@ -38,19 +50,76 @@ func listCmd(c *cli.Context) error {
...
+	if images {
+		images, err := store.Images()
+		if err != nil {
+			return err
+		}
+
+		if len(images) > 0 && !quiet {
+			fmt.Printf("\nIMAGES\n\n")
+			if truncate {
+				fmt.Printf("%-12s %-64s\n", "IMAGE ID", "IMAGE NAME")
```
TomSweeneyRedHat: do you want to do the full 64 for the name or a shorter version here?

rhatdan: Since we are trying to stay within 80 chars, I thought we would just do 64, also the Name is somewhat important while the ID is pretty much only used by the computer.
We have a bug in this patch in that you can not use the short name for delete operations.

rhatdan:
@nalind PTAL.  This one is a little more controversial in that we are expanding the list command to handle both containers and images.

rhatdan:
@TomSweeneyRedHat Could you take a look at getting the delete command to handle the short names?

cmd/buildah/list.go
TomSweeneyRedHat: That's works for me, especially given the bug,  I asked because I thought you shortened it at line 93.  However i need to read more carefully, you didn't shorten it there either.

TomSweeneyRedHat:
Sure, I'll see what I can mess up with short name deletes.

rhatdan:
Will rename buildah list to
buildah containers
and 
buildah images.

PR #49, opened 2017-03-28
Accept commit targets which are just store refs

nalind:
Accept commit target names which don't include a transport name by checking if they parse as valid containers-storage references after they fail to parse as a general reference.


rhatdan:
@rh-atomic-bot r+ bd0efd1

PR #52, opened 2017-03-28
Add truncation to buildah containers|images

rhatdan:


nalind:
LGTM.  We could stand to pick one or the other in terms of "do not"/"don't" and whether flags that start with "no" should be hyphenated or not, but that doesn't have to happen here.

cmd/buildah/containers.go
```diff
@@ -17,6 +17,10 @@ var (
 			Name:  "noheading, n",
 			Usage: "do not print column headings",
 		},
+		cli.BoolFlag{
+			Name:  "no-trunc",
```
TomSweeneyRedHat: Are you supposed to truncate no-truncate?  ;^p

rhatdan: Fine with me.

rhatdan:
Went with --noheading and --notruncate
and 
Do not ...

rhatdan:
@rh-atomic-bot r+  	efad757

PR #55, opened 2017-03-28
Use "scratch" for tests where we never "run"

nalind:
We can use "scratch" now as a source for building images, so it should speed things up a bit if we use it instead of pulling alpine in tests where it isn't expected to affect the test result.


rhatdan:
LGTM
@rh-atomic-bot r+	c9e142b

PR #56, opened 2017-03-28
Handle ADD/COPY where the destination ends with '/'

nalind:
When the destination for an ADD or COPY operation ends with a path separator, take that as an indicator that the destination should be a directory, that we should create it if it doesn't already exist, and that any files we're copying should be placed in the directory.


nalind:
[[Buildah] Add "dockah build-dockerfile"](https://trello.com/c/AyBXqoRY/443-buildah-add-dockah-build-dockerfile)

rhatdan:
LGTM
@rh-atomic-bot r+ 68d934a

PR #57, opened 2017-03-29
Fix copy and paste error on container-image

rhatdan:


rhatdan:
@TomSweeneyRedHat PTAL

rhatdan:
@rh-atomic-bot r+ 5017a57
Tom when they are obvious, just do the above...

PR #58, opened 2017-03-29
Apply Last-Modified time as mtime when saving URLs

nalind:
When saving the contents of a URL to a local file, attempt to set mtime based on the response's Last-Modified header, if there is one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,4 +11,4 @@ before_install:
     - sudo apt-get -qq install bats btrfs-tools libdevmapper-dev libgpgme11-dev
 script:
     - make
-    - cd tests; sudo ./test_runner.sh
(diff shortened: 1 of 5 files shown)


rhatdan:
:+1:

nalind:
Telling Travis to run the tests under sudo without resetting $PATH, so the go compiler should be found at test-time.

nalind:
Merging, since the last revision changed the invocation of sudo in the Travis config, and nothing else.
@rh-atomic-bot +r f65c3061e3e28fb3ef1efb749db549744343a27d

nalind:
@rh-atomic-bot r+ f65c3061e3e28fb3ef1efb749db549744343a27d

PR #59, opened 2017-03-29
Initial build-using-dockerfile implementation

nalind:
Add a `build-using-dockerfile` command which uses `openshift/imagebuilder` to wrap parsing and handle most of the dispatching, and the logic behind the other subcommands to implement assorted instructions.


cmd/buildah/build.go
```diff
@@ -0,0 +1,136 @@
...
+		cli.StringSliceFlag{
+			Name:  "arg",
+			Usage: "`argument=value` to supply to the builder",
+		},
+		cli.StringFlag{
+			Name:  "t",
+			Usage: "`tag` to apply to the built image",
+		},
+		cli.StringSliceFlag{
+			Name:  "f",
```
rhatdan: Name:  "file, f",

rhatdan: Do we want this or do we just specify Dockerfile in the command.

buildah bud Dockerfile  
Or do you want to match docker syntax totally>

cmd/buildah/build.go
```diff
@@ -0,0 +1,136 @@
...
+			Usage: "`pathname` of a Dockerfile",
+		},
+	}
+	buildDescription = "Builds an OCI image using instructions in one or more Dockerfiles."
```
rhatdan: Does this accept more then one Dockerfiles?  Or is this a copy/paste error.

cmd/buildah/build.go
```diff
@@ -0,0 +1,136 @@
...
+	buildCommand     = cli.Command{
+		Name:        "build-using-dockerfile",
+		Aliases:     []string{"bud"},
+		Usage:       "Build an image using instructions in a Dockerfile",
+		Description: buildDescription,
+		Flags:       buildFlags,
+		Action:      buildCmd,
+		ArgsUsage:   "CONTEXT-DIRECTORY",
```
rhatdan: In Docker build this is a PATH-or-URL

cmd/buildah/build.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+)
+
+func buildCmd(c *cli.Context) error {
+	output := ""
+	if c.IsSet("t") {
+		output = c.String("t")
```
rhatdan: tag

cmd/buildah/build.go
```diff
@@ -0,0 +1,136 @@
...
+				delete(args, av[0])
+			}
+		}
+	}
+	quiet := false
+	if c.IsSet("quiet") {
+		quiet = c.Bool("quiet")
+	}
+	dockerfiles := []string{}
+	if c.IsSet("f") {
```
rhatdan: file

nalind:

cmd/buildah/build.go
```diff
@@ -0,0 +1,136 @@
...
+		PullAlways:          pullAlways,
+		Registry:            registry,
+		Compression:         archive.Gzip,
+		Quiet:               quiet,
+		SignaturePolicyPath: signaturePolicy,
+		Args:                args,
+		Output:              output,
+	}
+
+	return imagebuildah.BuildDockerfileFromFile(store, options, dockerfiles...)
```
rhatdan: Should this just be BuildDockerfile(...)

imagebuildah/build.go
```diff
@@ -0,0 +1,331 @@
...
+	runtime                        string
+	runtimeArgs                    []string
+	transientMounts                []specs.Mount
+	compression                    archive.Compression
+	output                         string
+	signaturePolicyPath            string
+	systemContext                  *types.SystemContext
+}
+
+func getSystemContext(signaturePolicyPath string) *types.SystemContext {
```
rhatdan: Should this function be just
systemContext(...)

imagebuildah/build.go
```diff
@@ -0,0 +1,331 @@
...
+	sc := &types.SystemContext{}
+	if signaturePolicyPath != "" {
+		sc.SignaturePolicyPath = signaturePolicyPath
+	}
+	return sc
+}
+
+func (b *buildahExecutor) Preserve(path string) error {
```
rhatdan: We need to add comments to all externally defined functions.  We need to start running golint and govalidate against our code.  I have no idea what Preserve does and it seems to be not implemented yet.

imagebuildah/build.go
```diff
@@ -0,0 +1,331 @@
...
+	logrus.Debugf("PRESERVE %q", path)
+	return nil
+}
+
+func (b *buildahExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) error {
```
rhatdan: Comment.

imagebuildah/build.go
```diff
@@ -0,0 +1,331 @@
...
+			}
+		}
+		if err := b.builder.Add(copy.Dest, false, sources...); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (b *buildahExecutor) Run(run imagebuilder.Run, config docker.Config) error {
```
rhatdan: Comment

rhatdan:
BTW This is pretty awesome.

cmd/buildah/build.go
```diff
@@ -0,0 +1,136 @@
+package main
```
nalind: bud's a lot less typing, so I'll rename it to that.

cmd/buildah/build.go
```diff
@@ -0,0 +1,136 @@
...
+
+	"github.com/containers/storage/pkg/archive"
+	"github.com/projectatomic/buildah/imagebuildah"
+	"github.com/urfave/cli"
+)
+
+var (
+	buildFlags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "quiet",
```
nalind: 👍

cmd/buildah/build.go
```diff
@@ -0,0 +1,136 @@
...
+		cli.StringFlag{
+			Name:  "registry",
+			Usage: "prefix to prepend to the image name in order to pull the image",
+			Value: DefaultRegistry,
+		},
+		cli.BoolTFlag{
+			Name:  "pull",
+			Usage: "pull the image if not present",
+		},
+		cli.BoolFlag{
```
nalind: This defaults to true, and --pull-always defaults to false, like "buildah from" does.  I'll note that "docker build"'s --pull is equivalent to --pull-always, so maybe we need to drop --pull from both, and rename --pull-always to --pull?

cmd/buildah/build.go
TomSweeneyRedHat: 👍

imagebuildah/build.go
TomSweeneyRedHat: fwiw, I prefer keeping the get in the name here.

cmd/buildah/build.go
TomSweeneyRedHat: 👍

imagebuildah/build.go
```diff
@@ -0,0 +1,334 @@
...
+		return err
+	}
+	if err = b.Commit(ib, node); err != nil {
+		return err
+	}
+	return nil
+}
+
+// BuildDockerfile builds an image using instructions from the named file.
+func BuildDockerfileFromFile(store storage.Store, options BuildDockerfileOptions, dockerfile ...string) error {
```
TomSweeneyRedHat: If you change the name to BuildDockerfileFromFile in build.go, you probably want to do the same here.

cmd/buildah/build.go
```diff
@@ -0,0 +1,136 @@
...
+			Name:  "pull-always",
+			Usage: "pull the image, even if a version is present",
+		},
+		cli.StringFlag{
+			Name:  "signature-policy",
+			Usage: "signature policy path (not usually used)",
+		},
+		cli.StringSliceFlag{
+			Name:  "arg",
+			Usage: "`argument=value` to supply to the builder",
```
nalind: This is used to supply a value for a variable named by an ARG instruction in a Dockerfile, after which the variable's value can be referred to in other instructions, and will be added to the environment list we get when we're told to process a RUN instruction.  Renaming this to `--build-arg`.

cmd/buildah/build.go
```diff
@@ -0,0 +1,136 @@
...
+		},
+		cli.StringFlag{
+			Name:  "t",
```
nalind: 👍

cmd/buildah/build.go
nalind: Not totally, but I think we at least need this flag to handle cases where a directory that we're pointed to as a build context is one of those where we're expected to able to select among multiple Dockerfiles with names like Dockerfile.${something}.

cmd/buildah/build.go
nalind: 👍

cmd/buildah/build.go
nalind: 👍

cmd/buildah/build.go
nalind: Traditional imagebuilder (which I'd like to reuse some of this logic in when we get there) does, to a degree: the FROM instructions in files after the first are effectively ignored, but other instructions are not, and the image is committed after instructions from all files have been executed.  I'll clean up the logic to mimic this.

imagebuildah/build.go
rhatdan: I am fine with get, but I think @mrunalp pointed out that there is some kind of GOLANG standard that says this is discouraged.  I think golint and goverify used in CRI-O and runc give you errors when you start functions with get.

mrunalp: Here is the suggested style reference: https://golang.org/doc/effective_go.html#Getters

nalind: Hmm, that's focused more on wrappers for reading and writing object members, but in terms of what it does, it probably makes more sense to call it makeSystemContext().

imagebuildah/build.go
nalind: See #62 for adding formatting and style checking to CI.  The `Preserve` method is used to signal the `Executor` that the path has been marked as part of a volume, which means we need to keep instructions other than an explicit ADD or COPY from modifying anything below that point.  The imagebuilder dockerclient implementation uses caching/restoring around its handling of RUN instructions to accomplish this.  Adding that.

cmd/buildah/build.go
nalind: Support for remote context archives or Dockerfiles wasn't in there.  Adding.

nalind:
Updated:
* rebased
* golint clean
* tests no longer fail without #61 
* `Preserve()` is now implemented, so handling of volumes during bud should be more in line with expectations
* the bud command's `--arg` flag is now named `--build-arg`
* the bud command's `-t`, `-f`, and `-q` flags now also have long names `--tag`, `--file`, and `--quiet`
* Dockerfiles can be specified using http/https URLs
* context directories can be specified as "git://" URLs, "github.com/..." repositories, or as http/https URLs of archives
* handling of multiple Dockerfiles is more in line with how `openshift/imagebuildah` uses them (in order, only the first FROM: counts)
* renamed `getSystemContext()` to `makeSystemContext()`
* more tests

rhatdan:
Lets remove the 1.6 tests, in a separate patch.

rhatdan:
LGTM

Do we eventually want to turn ImageBuilder to use Buildah?  Might want to move this code out of ImageBuilder into Buildah at that point.

nalind:
Moving the tell-Travis-to-stop-using-1.6 patch to #65 and rebasing before merging.

nalind:
I think the imagebuilder CLI could be taught to toggle between its current dockerclient Executor and the one we implement here, as @smarterclayton suggested in https://github.com/openshift/imagebuilder/issues/33.  We don't yet support all of the options and hooks that the imagebuilder CLI tool wants to be able to specify, but that's fixable.

smarterclayton:
You should also be able to snip more of the dependency tree that we pull in - let me know if there are changes that need to be done to make the core imagebuilder code not depend on docker dependencies.

cmd/buildah/bud.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/containers/storage/pkg/archive"
+	"github.com/projectatomic/buildah/imagebuildah"
+	"github.com/urfave/cli"
+)
+
+var (
+	budFlags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "quiet, q",
+			Usage: "refrain from announcing build instructions",
```
rhatdan: Should we reverse this and change it to verbose, and then default it to value so we don't announce build instructions?

rhatdan:
man pages and bash completions?

nalind:
Added a man page and updated the bash-completion script.

docs/buildah-bud.md
```diff
@@ -0,0 +1,79 @@
+## buildah-bud "1" "April 2017" "buildah"
+
+## NAME
+buildah bud - Build an image using instructions from Dockerfiles.
+
+## SYNOPSIS
+**buildah** **bud | build-using-dockerfile** [*options* [...]] [**context**]
```
rhatdan: Is the context directory optional?

rhatdan: I guess it is if you specify a Dockerfile.

rhatdan:
@rh-atomic-bot r+ 390defd

docs/buildah-bud.md
nalind: If at least one of the Dockerfiles is a local file, we assume the directory that it's in is the context directory.

PR #60, opened 2017-03-29
buildah should be install in /usr/bin not /usr/sbin

rhatdan:


nalind:
LGTM pending CI tests.

rhatdan:
@rh-atomic-bot r+ a92e1e1

PR #61, opened 2017-03-29
Broaden what "rmi" can accept, and add tests

nalind:
Teach "rmi" to handle arbitrary image references and image references that implicitly refer to the local store, and add tests for both, falling back to remove-by-ID for anything left after we finish the basic tests.  This _should_ fix the errors in the self-test which currently cause CI to fail for #59.

### DIFF
--- a/cmd/buildah/rmi.go
+++ b/cmd/buildah/rmi.go
@@ -3,6 +3,11 @@ package main
 import (
 	"fmt"
 
+	"github.com/Sirupsen/logrus"
+	"github.com/containers/image/storage"
+	"github.com/containers/image/transports"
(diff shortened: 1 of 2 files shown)


rhatdan:
Will this handle truncated ID's?

rhatdan:
@rh-atomic-bot r+ 	fc27a56

nalind:
Truncated IDs should work as soon as we get https://github.com/containers/storage/pull/6, or something like it in.

nalind:
@rh-atomic-bot retry

rhatdan:
@rh-atomic-bot  r+ e583b9c

PR #62, opened 2017-04-03
Add some validation checks

nalind:
Add checks using gofmt, gometalinter, and git-validate, and fixup the things that they flag.

### DIFF
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -108,8 +108,8 @@ func updateConfig(builder *buildah.Builder, c *cli.Context) {
 		}
 	}
 	if c.IsSet("env") {
-		for _, envSpec := range c.StringSlice("env") {
-			builder.Env = append(builder.Env, envSpec)
+		if envSpec := c.StringSlice("env"); len(envSpec) > 0 {
+			builder.Env = append(builder.Env, envSpec...)
 		}
 	}
 	if c.IsSet("entrypoint") {
@@ -129,8 +129,8 @@ func updateConfig(builder *buildah.Builder, c *cli.Context) {
 		}
 	}
 	if c.IsSet("volume") {
-		for _, volSpec := range c.StringSlice("volume") {
-			builder.Volumes = append(builder.Volumes, volSpec)
+		if volSpec := c.StringSlice("volume"); len(volSpec) > 0 {
+			builder.Volumes = append(builder.Volumes, volSpec...)
 		}
 	}
 	if c.IsSet("label") {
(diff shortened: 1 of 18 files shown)


import.go
```diff
@@ -26,13 +25,13 @@ func importBuilder(store storage.Store, options ImportOptions) (*Builder, error)
...
 		}
-		src, err := ref.NewImage(systemContext)
-		if err != nil {
-			return nil, fmt.Errorf("error instantiating image: %v", err)
+		src, err3 := ref.NewImage(systemContext)
+		if err3 != nil {
+			return nil, fmt.Errorf("error instantiating image: %v", err3)
 		}
 		defer src.Close()
 		config, err = src.ConfigBlob()
```
TomSweeneyRedHat: Not an issue, but I'm not getting why you added numbers to some but not all 'err' variables.

cmd/buildah/config.go
```diff
@@ -129,8 +129,8 @@ func updateConfig(builder *buildah.Builder, c *cli.Context) {
 		}
 	}
 	if c.IsSet("volume") {
-		for _, volSpec := range c.StringSlice("volume") {
-			builder.Volumes = append(builder.Volumes, volSpec)
+		if volSpec := c.StringSlice("volume"); len(volSpec) > 0 {
+			builder.Volumes = append(builder.Volumes, volSpec...)
```
TomSweeneyRedHat: Neat trick using the ..., will have to try and remember that.

import.go
nalind: These avoid shadowing warnings from one of the linters, by removing cases where we declare a variable in a block when a variable with that same name already exists in an enclosing block.  https://golang.org/ref/spec#Short_variable_declarations also covers why we don't have to do this when we redeclare a variable from the same block.

cmd/buildah/config.go
nalind: Agreed, it's a very handy language feature.

nalind:

import.go
TomSweeneyRedHat: Ty for the 411!

nalind:
Rebased.

rhatdan:
@rh-atomic-bot r+ c842847

nalind:
@rh-atomic-bot retry

nalind:
Reworked some of the install targets to merge install-tools and .install.md2man and avoid having "make all" install things outside of the build tree.

nalind:
@TomSweeneyRedHat, @rhatdan, PTAL again for that last patch, which is kind of small to be on its own, since we're already messing around with the top-level Makefile.

rhatdan:
@rh-atomic-bot r+ 540eb22

PR #63, opened 2017-04-04
Teach "Run" to dig user IDs out of containers

nalind:
When we have cgo, use fgetpwent() to try to look up user IDs and primary GIDs in containers.  If that fails for any reason (or if we don't have cgo), fall back to doing what we were doing before (i.e., trying to look up the information on the host).

### DIFF
--- a/user.go
+++ b/user.go
@@ -1,33 +1,78 @@
 package buildah
 
 import (
+	"fmt"
 	"os/user"
 	"strconv"
+	"strings"
 
 	"github.com/opencontainers/runtime-spec/specs-go"
 )
 
-// TODO: we should doing these lookups using data that's actually in the container.
-func getUser(username string) (specs.User, error) {
-	if username == "" {
+func getUser(rootdir, userspec string) (specs.User, error) {
+	var gid64 uint64
+	var gerr error = user.UnknownGroupError("error looking up group")
+
+	spec := strings.SplitN(userspec, ":", 2)
+	userspec = spec[0]
+	groupspec := ""
+	if userspec == "" {
 		return specs.User{}, nil
 	}
-	runuser, err := user.Lookup(username)
-	if err != nil {
-		return specs.User{}, err
+	if len(spec) > 1 {
+		groupspec = spec[1]
 	}
-	uid, err := strconv.ParseUint(runuser.Uid, 10, 32)
-	if err != nil {
-		return specs.User{}, nil
+
+	uid64, uerr := strconv.ParseUint(userspec, 10, 32)
+	if uerr == nil && groupspec == "" {
+		// We parsed the user name as a number, and there's no group
+		// component, so we need to look up the user's primary GID.
+		name := ""
+		name, gid64, gerr = lookupGroupForUIDInContainer(rootdir, uid64)
+		if gerr == nil {
+			userspec = name
+		} else {
+			if userrec, err := user.LookupId(userspec); err == nil {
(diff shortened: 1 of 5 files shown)


user_linux_cgo.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"path/filepath"
+	"sync"
+	"unsafe"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/opencontainers/runtime-spec/specs-go"
+)
+
+func getUser(rootdir, username string) (specs.User, error) {
```
TomSweeneyRedHat: Ok, showing my golang ignorance here.  What does specs.user and error in the second parens do?

nalind: That's the syntax for declaring that a function returns multiple values.

nalind:

user_linux_cgo.go
TomSweeneyRedHat: Thx again, just found it in the golang spec.

user_linux_cgo.go
```diff
@@ -0,0 +1,51 @@
...
+	var m sync.Mutex
+	m.Lock()
+	defer m.Unlock()
+
+	pwdfile := C.CString(filepath.Join(rootdir, "/etc/passwd"))
```
nalind: Hmm, yes, it could.  Will have to add some checks for that.

user_linux_cgo.go
nalind: I empathize.  For handling the equivalent of a RUN instruction while building an image, though, I think it'll have to be sufficient.  It mirrors what we get from `docker build` when evaluating a Dockerfile, and we're trying to avoid surprising people who are used to that.

user_unix_cgo.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/Sirupsen/logrus"
+	"github.com/opencontainers/runtime-spec/specs-go"
+)
+
+func getUser(rootdir, username string) (specs.User, error) {
+	var st, lst syscall.Stat_t
+	var m sync.Mutex
+	m.Lock()
+	defer m.Unlock()
```
nalind: This mutex would never have helped anything.  Rewriting.

nalind:
@rh-atomic-bot, retest this please.

user.go
```diff
@@ -1,33 +1,73 @@
...
 )
 
-// TODO: we should doing these lookups using data that's actually in the container.
-func getUser(username string) (specs.User, error) {
-	if username == "" {
+func getUser(rootdir, username string) (specs.User, error) {
+	var gid64 uint64
+	var gerr error = user.UnknownGroupError("error looking up group")
+
+	userspec := strings.SplitN(username, ":", 2)
```
rhatdan: How about 
user, group := strings.SplitN(username, ":", 2)
Then you don' t need to deal with the [0] stuff all over the place.

nalind: SplitN() returns a slice, so the compiler rejects that change as an assignment count mismatch.

rhatdan: Ok how about you do something like

userspec := strings.SplitN(username, ":", 2)
user := userspec[0]
group := userspec[1]

Unless their is a fancy way to convert this in one line.

nalind: I don't know of one.  Revising.

nalind:
Rebasing.

rhatdan:
@rh-atomic-bot r+ baa1aa2

PR #64, opened 2017-04-04
Make Run() use a terminal only if it has one

nalind:
Make Run() run the command with a terminal if we're being run with a terminal.  This lets us avoid a weird workaround in the tests.

### DIFF
--- a/run.go
+++ b/run.go
@@ -23,6 +23,12 @@ const (
 	DefaultRuntime = "runc"
 )
 
+const (
+	DefaultTerminal = iota
+	WithoutTerminal
+	WithTerminal
+)
+
 // RunOptions can be used to alter how a command is run in the container.
 type RunOptions struct {
 	// Hostname is the hostname we set for the running container.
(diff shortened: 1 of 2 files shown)


nalind:

run.go
```diff
@@ -45,6 +45,10 @@ type RunOptions struct {
 	Entrypoint []string
 	// NetworkDisabled puts the container into its own network namespace.
 	NetworkDisabled bool
+	// ForceTerminal overrides the should-we-use-a-pty logic with a "yes".
```
rhatdan: Feels like we should have one value here, which takes three option.
FORCETERM, DEFAULT (DONOTHING), FORCENOTERM

UseTerminal 

This way we avoid the case where someone sets both of your options to True.

nalind: Okay, changing it.

rhatdan:
@rh-atomic-bot r+ ca3ce93

nalind:
@rh-atomic-bot retry

PR #68, opened 2017-04-13
rm/rmi: Don't exit until we've tried them all

nalind:
If we're told to remove more than one container or image, actually try to remove them all, but still return an error if we fail at any of them.

### DIFF
--- a/cmd/buildah/rm.go
+++ b/cmd/buildah/rm.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/urfave/cli"
 )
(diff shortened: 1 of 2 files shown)


rhatdan:
I think the other patch set just merged both?

nalind:
This one duplicates the man page part, as I was going to move that hunk from the larger man page PR to here, but since we're about to merge that I can undo that.

TomSweeneyRedHat:
LGTM.  Obviously we're only returning the last error if there are multiples, but I think that's ok.  Regardless, this is much better than the previous functionality.

rhatdan:
@rh-atomic-bot r+ 82a34a1

PR #71, opened 2017-04-17
Clean up image pulling policy

nalind:
Change how we represent image pulling policy from a pair of booleans to a single field which can take three values (never, if-missing, always), with the default value being if-missing.


rhatdan:
@rh-atomic-bot r+ d2c5325

PR #72, opened 2017-04-17
Fix handling of man page for make install

rhatdan:


nalind:
Can we more closely mirror what automake would do here?  By that I mean only referencing $(DESTDIR) in install-related targets, making $(MANDIR) relative to $(DATADIR), itself relative to $(PREFIX), and installing man pages to $(MANDIR)/man1/.

rhatdan:
nalind, this worked with the spec file so I think it is ready to be merged.

nalind:
LGTM.
@rh-atomic-bot r+ ff61a844070fb112363be338d17f9958dae03a6e

PR #73, opened 2017-04-17
imagebuildah: Provide logging callbacks

nalind:
Add a Log() callback and Out and Err io.Writers to the set of options that imagebuildah can accept.  If Log is not specified, log to Err.  If Err is not specified, log to os.Stderr.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -61,6 +61,14 @@ type BuildOptions struct {
 	Args map[string]string
 	// Name of the image to write to.
 	Output string
+	// Log is a callback that will print a progress message.  If no value
+	// is supplied, the message will be sent to Err (or os.Stderr, if Err
+	// is nil) by default.
+	Log func(format string, args ...interface{})
+	// Out is a place where non-error log messages are sent.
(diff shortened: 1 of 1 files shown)


imagebuildah/build.go
```diff
@@ -61,6 +61,14 @@ type BuildOptions struct {
...
 	// Name of the image to write to.
 	Output string
+	// Log is a callback that will print a progress message.  If no value
+	// is supplied, the message will be sent to Err (or os.Stderr, if Err
+	// is nil) by default.
+	Log func(format string, args ...interface{})
+	// Out is a place where low-priority status messages are sent.
+	Out io.Writer
+	// Err is a place where higher-priority status messages are sent.
+	Err io.Writer
```
rhatdan: Seems a bit weird to call this Err, if all it is getting is "higher-priority status messages"

nalind: Yeah.  Revising that to be "error log messages", since more often than not it tends to be set to `os.Stderr`.

rhatdan:
@rh-atomic-bot r+ cb6df7a

PR #74, opened 2017-04-18
imagebuildah: Commit() doesn't need a parser node

nalind:
The imagebuildah Executor.Commit() method doesn't use its parser node argument, and the Commit() method in the dockerclient builder in imagebuilder doesn't, so switch to match its signature.


rhatdan:
@rh-atomic-bot r+ 42b4406

PR #75, opened 2017-04-18
Internalize some Docker types

nalind:
Internalize the type declaration of a Docker image, so that we don't need to pull in code and a pile of additional dependencies when we just want to know how to parse out a JSON object.  Carry a copy of the license that covered the sources from which the definitions came.

### DIFF
--- a/config.go
+++ b/config.go
@@ -6,8 +6,8 @@ import (
 	"time"
 
 	"github.com/Sirupsen/logrus"
-	docker "github.com/docker/docker/image"
 	ociv1 "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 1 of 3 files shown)


docker/LICENSE
```diff
@@ -0,0 +1,191 @@
+
```
rhatdan: Do we actually need this since Buildah is Apache licensed?

nalind: Fair point.  I don't see us wanting to change the license, so I can drop that file.

rhatdan:
LGTM

nalind:
@rh-atomic-bot r+ fbf1b83adee1a70e67105075e316dfa57b8e2a6e

PR #76, opened 2017-04-18
bud: Add flags for specifying the runtime command

nalind:
So long as it's possible to have an incompatible version of runc installed, we're going to want to easily point to another version, as we already allow for with the "run" command.

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/Sirupsen/logrus"
 	"github.com/containers/storage/pkg/archive"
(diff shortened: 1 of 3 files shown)


rhatdan:
man pages & bash completions.

nalind:
It looks like I inadvertently already merged those in with earlier changes to the completions and man pages, though it's worth clarifying that we're talking about an OCI-compatible runtime.

rhatdan:
LGTM

nalind:
@rh-atomic-bot r+ 2986b3106ed1334c90856853ad312e72d9045500

PR #83, opened 2017-04-21
imagebuildah: Prepare() accepts a source image now

nalind:
Update to mimic an API change in imagebuilder, which now allows us to provide the FROM image directly via an API, potentially overriding the first one named in the Dockerfile.


rhatdan:
LGTM, except it seems like the test infrastructure is failing.

nalind:
Rebased.

rhatdan:
@rh-atomic-bot r+  	fd41585

PR #84, opened 2017-04-21
imagebuildah: Reexport some things

nalind:
Have imagebuildah reexport some constants and its own Mount type, to reduce the number of our dependencies that a prospective consumer of this package would also need to import directly.


rhatdan:
@rh-atomic-bot r+ 1e9a02a

PR #86, opened 2017-04-24
Remove --mount from buildah-from

rhatdan:
We don't have a good way of identifying to a user where a container
was mounted on the file system when doing a buildah from.

A user can just do a buildah mount CID /mnt, after the from line to
handle this.

### DIFF
--- a/cmd/buildah/from.go
+++ b/cmd/buildah/from.go
@@ -37,10 +37,6 @@ var (
 			Name:  "signature-policy",
 			Usage: "`pathname` of signature policy file (not usually used)",
 		},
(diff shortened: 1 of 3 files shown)


rhatdan:
Fixes https://github.com/projectatomic/buildah/issues/5

nalind:
LGTM.

TomSweeneyRedHat:
LGTM

nalind:
@rh-atomic-bot r+ b3efb80f4d5c4bfc97ba8f8c1cb3034c79a02b3d

PR #88, opened 2017-05-01
imagebuildah: add AdditionalTags

nalind:
Add an AdditionalTags field to the imagebuildah options structure, to provide for additional tags which we'll want to apply to the image that we're writing.

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -1,8 +1,12 @@
 package buildah
 
 import (
+	"fmt"
+
 	"github.com/containers/image/copy"
+	"github.com/containers/image/docker/reference"
 	"github.com/containers/image/signature"
+	"github.com/containers/image/storage"
 	"github.com/containers/image/types"
 	"github.com/containers/storage/pkg/archive"
 )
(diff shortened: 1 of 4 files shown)


commit.go
```diff
@@ -37,5 +67,20 @@ func (b *Builder) Commit(dest types.ImageReference, options CommitOptions) error
 		return err
 	}
 	err = copy.Image(policyContext, dest, src, getCopyOptions())
+	switch dest.Transport().Name() {
+	case storage.Transport.Name():
+		tags, err := expandTags(options.AdditionalTags)
+		if err != nil {
```
TomSweeneyRedHat: Do you want to also spin off some kind of error message for these similar to what you did above in expandTags?  Or will the returned err have enough info in it to tell which of these three conditions the error sprang from?

nalind: Yeah, it should be doing that for the SetNames() call, at least.  I'll fix that.

nalind:
Added some context to the error message returned if we get an error from SetNames().

nalind:
Removed a redundant type assertion that @TomSweeneyRedHat spotted out of band.

commit.go
rhatdan: Did you fix this?

nalind: Yes, not long after.

rhatdan:
@rh-atomic-bot r+ 3c3174c

PR #91, opened 2017-05-03
Add a "go vet" validation step

nalind:
This adds another health check.


rhatdan:
@rh-atomic-bot r+ 1d43802

PR #98, opened 2017-05-09
Report pull/commit progress by default

### LINKED ISSUE #94: buildah does not report progress when pulling images
I thought `skopeo` needed to copy images from, e.g. DockerHub, and make them available before `buildah` runs, but it turns out that you can just: `buildah from centos` and it'll pull the image from DockerHub if needed. That's really great from a UX perspective, but it's hard to tell that's what is happening, since there's no progress indication.

Can we add some sort of status information (e.g. pulling image, pulling layers, etc) similar to how `docker` does it with its build/run commands?

nalind:
Have `from`, `commit`, and `build-using-dockerfile` report progress via stderr (so that capturing output from `from` and `commit` still works as expected) unless `--quiet` is used to suppress the reporting.  Should fix #94.

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -15,7 +15,7 @@ var (
 	budFlags = []cli.Flag{
 		cli.BoolFlag{
 			Name:  "quiet, q",
-			Usage: "refrain from announcing build instructions",
+			Usage: "refrain from announcing build instructions and image read/write progress",
 		},
 		cli.StringFlag{
 			Name:  "registry",
@@ -202,6 +202,9 @@ func budCmd(c *cli.Context) error {
 		AdditionalTags:      tags,
 		Runtime:             runtime,
 		RuntimeArgs:         runtimeFlags,
+	}
+	if !quiet {
+		options.ReportWriter = os.Stderr
 	}
 
 	return imagebuildah.BuildDockerfiles(store, options, dockerfiles...)
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/containers/image/storage"
 	"github.com/containers/image/transports/alltransports"
@@ -19,6 +20,10 @@ var (
 		cli.StringFlag{
 			Name:  "signature-policy",
 			Usage: "`pathname` of signature policy file (not usually used)",
+		},
+		cli.BoolFlag{
+			Name:  "quiet, q",
+			Usage: "don't output progress information when writing images",
 		},
 	}
 	commitDescription = "Writes a new image using the container's read-write layer and, if it is based\n   on an image, the layers of that image"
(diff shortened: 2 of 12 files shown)


nalind:
[[buildah] Report progress when pulling or committing images, like skopeo does](https://trello.com/c/vVLoq3On/479-buildah-report-progress-when-pulling-or-committing-images-like-skopeo-does)

cmd/buildah/bud.go
```diff
@@ -203,6 +203,9 @@ func budCmd(c *cli.Context) error {
 		Runtime:             runtime,
 		RuntimeArgs:         runtimeFlags,
 	}
+	if !quiet {
+		options.ReportWriter = os.Stderr
```
rhatdan: does skopeo default to stderr?

cmd/buildah/commit.go
```diff
@@ -20,6 +21,10 @@ var (
 			Name:  "signature-policy",
 			Usage: "`pathname` of signature policy file (not usually used)",
 		},
+		cli.BoolFlag{
+			Name:  "quiet",
```
rhatdan: add q as well as quiet

cmd/buildah/from.go
```diff
@@ -37,6 +38,10 @@ var (
 			Name:  "signature-policy",
 			Usage: "`pathname` of signature policy file (not usually used)",
 		},
+		cli.BoolFlag{
+			Name:  "quiet",
```
rhatdan: add q

cmd/buildah/commit.go
rhatdan: Should we make quiet a global option? In case there are other things that we want to be quiet in the future?

rhatdan:
Need changes to man pages and bash completions.

cmd/buildah/bud.go
nalind: skopeo uses stdout, but we've previously been attempting to make it useful to capture the stdout of the `from` and `commit` commands, so stderr seemed to be a better choice.

cmd/buildah/from.go
nalind: Fixing.

cmd/buildah/bud.go
rhatdan: Right , makes sense.

ctr=$(buildah from fedora)

Will still work.

cmd/buildah/commit.go
nalind: Fixed.  I'd shy away from making --quiet a global option if in a number of cases it would have no effect.

nalind:
Okay, man pages and completions should be up to date.

pull.go
```diff
@@ -54,6 +54,6 @@ func pullImage(store storage.Store, options BuilderOptions, sc *types.SystemCont
 
 	logrus.Debugf("copying %q to %q", spec, name)
 
-	err = copy.Image(policyContext, destRef, srcRef, getCopyOptions())
+	err = copy.Image(policyContext, destRef, srcRef, getCopyOptions(options.ReportWriter))
```
rhatdan: Should we pass in just options to make this future proof?

TomSweeneyRedHat: I think that's a good idea, but I'd vote to submit this as is and then add that later.

cmd/buildah/bud.go
nalind: Right, but in practice a script will probably still want to use --quiet to avoid the stderr spew.

rhatdan:
@rh-atomic-bot r+ 5f09cf5

pull.go
rhatdan: Ok I looked into this a little deeper, and I see that we pass different options from Buidah to Commit.  I now think we keep it the way it did.

PR #99, opened 2017-05-09
Don't bother setting redundant image names

nalind:
When we pull down an image to use as a starting point, we no longer need to attach any additional names to it, as the image library can now find it correctly without requiring us to do so.


rhatdan:
@rh-atomic-bot r+ 244b4de

PR #102, opened 2017-05-16
Improve handling of image configuration in build-using-dockerfile

nalind:
Previously, build-using-dockerfile wasn't taking sufficient care to communicate a source image's configuration to the imagebuilder object, nor to extract updated settings from it when committing the image.  This PR makes an effort to avoid discarding configuration information from source images, uses as much of that information as we keep to initialize the imagebuilder object, and reads those settings back out when a build is complete.

### DIFF
--- a/config.go
+++ b/config.go
@@ -2,29 +2,36 @@ package buildah
 
 import (
 	"encoding/json"
+	"path/filepath"
 	"runtime"
-	"time"
+	"strings"
 
-	"github.com/Sirupsen/logrus"
+	digest "github.com/opencontainers/go-digest"
 	ociv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/projectatomic/buildah/docker"
 )
 
-func copyDockerImageConfig(dimage *docker.Image) (ociv1.Image, error) {
+// makeOCIv1Image builds the best OCIv1 image structure we can from the
+// contents of the docker image structure.
+func makeOCIv1Image(dimage *docker.Image) (ociv1.Image, error) {
+	config := dimage.Config
+	if config == nil {
+		config = &dimage.ContainerConfig
+	}
 	image := ociv1.Image{
 		Created:      dimage.Created.UTC(),
 		Author:       dimage.Author,
 		Architecture: dimage.Architecture,
 		OS:           dimage.OS,
 		Config: ociv1.ImageConfig{
-			User:         dimage.Config.User,
+			User:         config.User,
 			ExposedPorts: map[string]struct{}{},
-			Env:          dimage.Config.Env,
-			Entrypoint:   dimage.Config.Entrypoint,
-			Cmd:          dimage.Config.Cmd,
-			Volumes:      dimage.Config.Volumes,
-			WorkingDir:   dimage.Config.WorkingDir,
-			Labels:       dimage.Config.Labels,
+			Env:          config.Env,
+			Entrypoint:   config.Entrypoint,
+			Cmd:          config.Cmd,
+			Volumes:      config.Volumes,
+			WorkingDir:   config.WorkingDir,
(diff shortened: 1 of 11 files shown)


nalind:
[(3) [buildah] Fix build-using-dockerfile's handling of image configurations](https://trello.com/c/nYY2wUAF/478-3-buildah-fix-build-using-dockerfile-s-handling-of-image-configurations)

rhatdan:
LGTM other then the question about docker/docker.

nalind:
I'm afraid I can't find that question.  Can you link to it in this conversation?

rhatdan:
Oops for got to press submit button.

config.go
```diff
@@ -59,89 +65,402 @@ func copyDockerImageConfig(dimage *docker.Image) (ociv1.Image, error) {
...
+
+// Labels returns a set of key-value pairs from the image's runtime
+// configuration.
+func (b *Builder) Labels() map[string]string {
+	return copyStringStringMap(b.OCIv1.Config.Labels)
+}
+
+// SetLabel adds or overwrites a key's value from the image's runtime
+// configuration.
+func (b *Builder) SetLabel(k string, v string) {
```
TomSweeneyRedHat: Something for your consideration.  Any value, perhaps in the future, to having a separate function for each of these?  Perhaps like below and then calling them both from here?  Just in case you might need to update just one in the future.    Ditto the next 2 or three related functions. 

func (b *Builder) SetOCIv1Label(k string, v string) {
	b.OCIv1.Config.Labels[k] = v
}

func (b *Builder) SetDockerLabel(k string, v string) {
	b.Docker.Config.Labels[k] = v
}

TomSweeneyRedHat: micro nit.  SetLabel is setting Labels.  Perhaps "SetLabels"

imagebuildah/build.go
```diff
@@ -514,6 +542,30 @@ func (b *Executor) Commit(ib *imagebuilder.Builder) (err error) {
 	} else {
 		imageRef, err = is.Transport.ParseStoreReference(b.store, "@"+stringid.GenerateRandomID())
 	}
+	config := ib.Config()
```
TomSweeneyRedHat: Should you return err here before this processing rather than down below?

TomSweeneyRedHat:
Nice chunk of code, I'm good with merging now and then picking off nits if appropriate.

config.go
nalind: In the future, perhaps.  This PR is geared toward ensuring that the values we set are kept in sync between the two configuration structures, but if we find a use case for those, we can add them.

nalind: These functions deal with one item at a time, though?

docker/types.go
```diff
@@ -44,6 +44,70 @@ type History struct {
...
+
+// github.com/docker/go-connections/nat/nat.go
+// PortSet is a collection of structs indexed by Port
+type PortSet map[Port]struct{}
+
+// github.com/docker/go-connections/nat/nat.go
+// Port is a string containing port number and protocol in the format "80/tcp"
+type Port string
+
+// github.com/docker/docker/api/types/container/config.go
```
nalind: The repository was renamed, as you know.  Adding a patch to update the pointers in comments.

imagebuildah/build.go
nalind: Yes, I think so.  Fixing.

config.go
TomSweeneyRedHat: Your call.  Yep, it's dealing with one label at a time, but you've a setter for a Labels field.  That's why I leaned toward the plural, but not a big lean at all.

rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+ 4be121b

PR #109, opened 2017-05-18
Add an initial "inspect" command

nalind:
Add an "inspect" command, which can be used to dump the contents of the Buildah object for a working container.  This'll make it easier to write integration tests that need to look at a container while we're working on it.

### DIFF
--- a/cmd/buildah/inspect.go
+++ b/cmd/buildah/inspect.go
@@ -1,0 +1,103 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"text/template"
+
+	"github.com/projectatomic/buildah"
+	"github.com/urfave/cli"
+)
+
+const (
+	defaultFormat = `Container: {{.Container}}
+ID: {{.ContainerID}}
+`
+	inspectTypeContainer = "container"
+	inspectTypeImage     = "image"
+)
+
+var (
+	inspectFlags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "type, t",
+			Usage: "look at the item of the specified `type` (container or image) and name",
+		},
+		cli.StringFlag{
+			Name:  "format, f",
+			Usage: "use `format` as a Go template to format the output",
+		},
+	}
+	inspectDescription = "Inspects a build container's or built image's configuration."
+	inspectCommand     = cli.Command{
+		Name:        "inspect",
+		Usage:       "Inspects the configuration of a container or image",
+		Description: inspectDescription,
+		Flags:       inspectFlags,
+		Action:      inspectCmd,
+		ArgsUsage:   "CONTAINER-OR-IMAGE",
+	}
+)
+
+func inspectCmd(c *cli.Context) error {
+	var builder *buildah.Builder
+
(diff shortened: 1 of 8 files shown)


cmd/buildah/inspect.go
```diff
@@ -0,0 +1,81 @@
...
+
+	raw := false
+	if c.IsSet("raw") {
+		raw = c.Bool("raw")
+	}
+	format := defaultFormat
+	if c.IsSet("format") {
+		format = c.String("format")
+	}
+	t := template.Must(template.New("format").Parse(format))
```
TomSweeneyRedHat: Is it possible to throw an error here and should we handle it?

docs/buildah-inspect.md
```diff
@@ -0,0 +1,27 @@
...
+**buildah** **inspect** [*options* [...] --] **containerID**
+
+## DESCRIPTION
+Prints information about a working container's configuration.
+
+## OPTIONS
+
+**--format** *template*
+
+Use *template* as a Go template when formatting the output.
```
TomSweeneyRedHat: Are there any example templates we could include here and/or other documentation?

cmd/buildah/inspect.go
nalind: Must() panics if the Parse() call fails, so we should be good there.

rhatdan:
Would you inspect an image?

docs/buildah-inspect.md
rhatdan: Should be able to steal something from the docker inspect man page?

nalind:
The format of the manifest (which skopeo displays directly in `--raw` mode) and configurations will vary, but I suppose we could build up a Buildah object using the configuration and display that.

nalind:
[(3) [buildah] Add a way to inspect working containers](https://trello.com/c/g5hLYG4M/501-3-buildah-add-a-way-to-inspect-working-containers)

docs/buildah-inspect.md
nalind: The description now includes:

Users of this option should be familiar with the [*text/template* package](https://golang.org/pkg/text/template/) in the Go standard library, and of internals of buildah's implementation.

If we want to go bigger, we can, but I don't have anything in mind.

nalind:
Inspecting of images (or rather, the builder configuration which we'd start with for a given image) should be working now.

contrib/completions/bash/buildah
```diff
@@ -520,6 +520,23 @@ return 1
      esac
  }
 
+ _buildah_inspect() {
+     local options_with_args="
+       --format
+       -f
+       --type
```
rhatdan: Can you add a completion so -t <tab> shows image and container?

rhatdan:
@bot retest

rhatdan:
@rh-atomic-bot r+ fc6e8ef

contrib/completions/bash/buildah
nalind: Makes sense, though I'll have to look that one up.  We'll probably want that for a --format option later anyway.

PR #110, opened 2017-05-19
Add a "tag" command

### LINKED ISSUE #106: Add buildah tag like Docker tag
There is no way to tag an existing image to other name, how about to add a command `buildah tag` to do it?

nalind:
Add a "tag" command, for adding names to locally-stored images.

Fixes #106.

### DIFF
--- a/cmd/buildah/tag.go
+++ b/cmd/buildah/tag.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/projectatomic/buildah/util"
+	"github.com/urfave/cli"
+)
+
+var (
+	tagDescription = "Adds one or more additional names to locally-stored image"
+	tagCommand     = cli.Command{
+		Name:        "tag",
+		Usage:       "Add an additional name to a local image",
+		Description: tagDescription,
+		Action:      tagCmd,
+		ArgsUsage:   "IMAGE-NAME [IMAGE-NAME ...]",
+	}
+)
+
+func tagCmd(c *cli.Context) error {
+	args := c.Args()
+	if len(args) < 2 {
+		return fmt.Errorf("image name and at least one new name must be specified")
+	}
+	store, err := getStore(c)
+	if err != nil {
+		return err
+	}
+	img, err := util.FindImage(store, args[0])
+	if err != nil {
+		return fmt.Errorf("error finding local image %q: %v", args[0], err)
+	}
+	err = util.AddImageNames(store, img, args[1:])
+	if err != nil {
+		return fmt.Errorf("error adding names %v to image %q: %v", args[1:], args[0], err)
+	}
+	return nil
+}
--- a/tests/tag.bats
+++ b/tests/tag.bats
@@ -1,0 +1,19 @@
+#!/usr/bin/env bats
+
+load helpers
+
+@test "tag" {
+  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
+  run buildah commit --signature-policy ${TESTSDIR}/policy.json "$cid" scratch-image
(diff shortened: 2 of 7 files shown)


nalind:
[(3) [buildah] Add a way to add names to images in local storage](https://trello.com/c/sbv2ARez/502-3-buildah-add-a-way-to-add-names-to-images-in-local-storage)

rhatdan:
LGTM

cmd/buildah/tag.go
```diff
@@ -0,0 +1,39 @@
...
+		Usage:       "Add an additional name to a local image",
+		Description: tagDescription,
+		Action:      tagCmd,
+		ArgsUsage:   "IMAGE-NAME [IMAGE-NAME ...]",
+	}
+)
+
+func tagCmd(c *cli.Context) error {
+	args := c.Args()
+	if len(args) == 0 {
```
TomSweeneyRedHat: Don't you need to add both an image and a name or names here?  I thought you'd want to check len(args) > 1 here.

nalind: Yes, that should be expecting at least two names.  Fixing.

cmd/buildah/tag.go
```diff
@@ -0,0 +1,39 @@
...
+		Usage:       "Add an additional name to a local image",
+		Description: tagDescription,
+		Action:      tagCmd,
+		ArgsUsage:   "IMAGE-NAME [IMAGE-NAME ...]",
+	}
+)
+
+func tagCmd(c *cli.Context) error {
+	args := c.Args()
+	if len(args) < 2 {
```
TomSweeneyRedHat: Ty and better than my backwards > 1 comment earlier.  DOH!

tests/tag.bats
```diff
@@ -0,0 +1,17 @@
...
+
+@test "tag" {
+  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
+  run buildah commit "$cid" scratch-image
+  [ "$status" -eq 0 ]
+  run buildah inspect --type image tagged-image
+  [ "$status" -ne 0 ]
+  run buildah tag scratch-image tagged-image tagged-also-image
+  [ "$status" -eq 0 ]
+  run buildah inspect --type image tagged-image
```
TomSweeneyRedHat: IDK why, but 3 has always been the magical painful number for me.  If you could add a test with three tags added and tested for I'd feel  better.  Not required, but appreciated if you do.

nalind: Might as well, since CI is failing in a way that it doesn't on my workstation.

util/util.go
```diff
@@ -0,0 +1,54 @@
...
+	return img, nil
+}
+
+// AddImageNames adds the specified names to the specified image.
+func AddImageNames(store storage.Store, image *storage.Image, addNames []string) error {
+	names, err := ExpandTags(addNames)
+	if err != nil {
+		return err
+	}
+	err = store.SetNames(image.ID, append(image.Names, names...))
```
TomSweeneyRedHat: Does Set Names not add a tag if it already exists as a tag?  I don't think we need it now given the way the code flow is now, but if we were able to call this some time in the future after the commit, it might be nice to not have duplicate tags.  Something for your consideration, no change necessary.

TomSweeneyRedHat:
LGTM, assuming the latests test runs come back happy and cheerful.

util/util.go
nalind: SetNames prunes out duplicates for its callers, and we're leaning on that feature here.

TomSweeneyRedHat: Super, I was hoping that was the case and thx for the confirmation.

rhatdan:
@rh-atomic-bot r+  	512b192

PR #114, opened 2017-05-22
Make sure maps in config blobs are always created

nalind:
When we decode map elements, either from our own data, or when importing from elsewhere, maps that don't contain any keys aren't created automatically, so we need to check, and potentially create, those maps when we want to add something to them.


nalind:

rhatdan:
@rh-atomic-bot r+  	d991edb

PR #115, opened 2017-05-22
Add more detail to errors returned from Commit()

nalind:
Add some context to errors returned from the Commit() method.

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -1,12 +1,14 @@
 package buildah
 
 import (
+	"fmt"
 	"io"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/containers/image/copy"
 	"github.com/containers/image/signature"
 	"github.com/containers/image/storage"
+	"github.com/containers/image/transports"
 	"github.com/containers/image/types"
 	"github.com/containers/storage/pkg/archive"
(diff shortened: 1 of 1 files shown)


nalind:

commit.go
```diff
@@ -47,22 +48,22 @@ func (b *Builder) Commit(dest types.ImageReference, options CommitOptions) error
...
-		return err
+		return fmt.Errorf("error copying layers and metadata: %v", err)
 	}
 	if len(options.AdditionalTags) > 0 {
 		switch dest.Transport().Name() {
 		case storage.Transport.Name():
 			img, err := storage.Transport.GetStoreImage(b.store, dest)
 			if err != nil {
-				return err
+				return fmt.Errorf("error locating just-written image: %v", err)
```
TomSweeneyRedHat: Could you add the value from b.store to the message to let the user know what it was looking at in particular when there was an error?  Or do you think that wouldn't help the debugging process later on down the road?

nalind: github.com/containers/image/transports.ImageName() should include that information, adding its value to the error message.

rhatdan:
@rh-atomic-bot r+  	328d5ad

PR #116, opened 2017-05-22
Initialize the builder's annotations from OCI data

nalind:
If we're building on an OCI image, initialize the Builder object's annotations using the annotations from the source image.

### DIFF
--- a/config.go
+++ b/config.go
@@ -136,6 +136,13 @@ func (b *Builder) initConfig() {
 		b.OCIv1 = image
 		b.Docker = dimage
 	}
+	if len(b.Manifest) > 0 {
+		// Attempt to recover format-specific data from the manifest.
+		v1Manifest := ociv1.Manifest{}
+		if json.Unmarshal(b.Manifest, &v1Manifest) == nil {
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -136,6 +136,13 @@ func (b *Builder) initConfig() {
 		b.OCIv1 = image
 		b.Docker = dimage
 	}
+	if len(b.Manifest) > 0 {
```
TomSweeneyRedHat: If we've a docker image, will b.Manifest still be > 0 ?  Are the manifests the same regardless of having a docker or  ociv1 image so no special handling is necessary in the below unmarshalling?

nalind: The manifest is the top level of data in the image, and while the data type that it encodes differs depending on whether we're using docker or oci formats, it _should_ always be there.

nalind:

config.go
TomSweeneyRedHat: Thx for the 411.

rhatdan:
@rh-atomic-bot r+  	edd0b04

PR #118, opened 2017-05-22
Support writing image manifests and configurations in both oci and docker formats

nalind:
This patch set adds `--format` options to the `commit` and `build-using-dockerfile` commands (also `-f` for `commit`) to allow selecting either docker v2s2 or oci v1 as output formats for the manifest and configuration blobs that we write to images.  It also adds initial support for importing settings from docker v2s1 images, though that's not as well tested since we don't produce that format.

This also helps us properly test the `config` command and the logic that imports settings from images in the two formats.

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -47,6 +47,10 @@ var (
 			Name:  "runtime-flag",
 			Usage: "add global flags for the container runtime",
 		},
+		cli.StringFlag{
+			Name:  "format",
+			Usage: "`format` of the built image's manifest and metadata",
+		},
 		cli.StringSliceFlag{
 			Name:  "tag, t",
 			Usage: "`tag` to apply to the built image",
@@ -130,6 +134,17 @@ func budCmd(c *cli.Context) error {
 	if c.IsSet("file") || c.IsSet("f") {
 		dockerfiles = c.StringSlice("file")
 	}
+	format := "oci"
+	if c.IsSet("format") {
+		format = strings.ToLower(c.String("format"))
+	}
+	if strings.HasPrefix(format, "oci") {
+		format = imagebuildah.OCIv1ImageFormat
+	} else if strings.HasPrefix(format, "docker") {
+		format = imagebuildah.Dockerv2ImageFormat
+	} else {
+		return fmt.Errorf("unrecognized image type %q", format)
+	}
 	contextDir := ""
 	cliArgs := c.Args()
 	if len(cliArgs) > 0 {
@@ -202,6 +217,7 @@ func budCmd(c *cli.Context) error {
 		AdditionalTags:      tags,
 		Runtime:             runtime,
 		RuntimeArgs:         runtimeFlags,
+		OutputFormat:        format,
 	}
 	if !quiet {
 		options.ReportWriter = os.Stderr
--- a/config.go
+++ b/config.go
@@ -13,7 +13,7 @@ import (
 
 // makeOCIv1Image builds the best OCIv1 image structure we can from the
 // contents of the docker image structure.
-func makeOCIv1Image(dimage *docker.Image) (ociv1.Image, error) {
+func makeOCIv1Image(dimage *docker.V2Image) (ociv1.Image, error) {
 	config := dimage.Config
 	if config == nil {
 		config = &dimage.ContainerConfig
@@ -42,7 +42,7 @@ func makeOCIv1Image(dimage *docker.Image) (ociv1.Image, error) {
 	for port, what := range config.ExposedPorts {
 		image.Config.ExposedPorts[string(port)] = what
 	}
-	RootFS := docker.RootFS{}
+	RootFS := docker.V2S2RootFS{}
 	if dimage.RootFS != nil {
 		RootFS = *dimage.RootFS
 	}
@@ -67,8 +67,8 @@ func makeOCIv1Image(dimage *docker.Image) (ociv1.Image, error) {
 
 // makeDockerV2S2Image builds the best docker image structure we can from the
 // contents of the OCI image structure.
-func makeDockerV2S2Image(oimage *ociv1.Image) (docker.Image, error) {
-	image := docker.Image{
+func makeDockerV2S2Image(oimage *ociv1.Image) (docker.V2Image, error) {
+	image := docker.V2Image{
 		V1Image: docker.V1Image{Created: oimage.Created.UTC(),
 			Author:       oimage.Author,
 			Architecture: oimage.Architecture,
@@ -84,11 +84,11 @@ func makeDockerV2S2Image(oimage *ociv1.Image) (docker.Image, error) {
 				Labels:       oimage.Config.Labels,
 			},
 		},
-		RootFS: &docker.RootFS{
+		RootFS: &docker.V2S2RootFS{
 			Type:    "",
 			DiffIDs: []digest.Digest{},
 		},
-		History: []docker.History{},
+		History: []docker.V2S2History{},
 	}
 	for port, what := range oimage.Config.ExposedPorts {
 		image.ContainerConfig.ExposedPorts[docker.Port(port)] = what
@@ -98,13 +98,13 @@ func makeDockerV2S2Image(oimage *ociv1.Image) (docker.Image, error) {
 		for _, id := range oimage.RootFS.DiffIDs {
 			d, err := digest.Parse(id)
 			if err != nil {
-				return docker.Image{}, err
+				return docker.V2Image{}, err
 			}
 			image.RootFS.DiffIDs = append(image.RootFS.DiffIDs, d)
 		}
 	}
 	for _, history := range oimage.History {
-		dhistory := docker.History{
+		dhistory := docker.V2S2History{
 			Created:    history.Created.UTC(),
 			CreatedBy:  history.CreatedBy,
 			Author:     history.Author,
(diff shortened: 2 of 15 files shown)


nalind:

cmd/buildah/bud.go
```diff
@@ -130,6 +134,17 @@ func budCmd(c *cli.Context) error {
 	if c.IsSet("file") || c.IsSet("f") {
 		dockerfiles = c.StringSlice("file")
 	}
+	format := "oci"
+	if c.IsSet("format") {
+		format = c.String("format")
```
rhatdan: Nit, could call strings.ToLower() here, and make the code below look a little neeter.

image.go
```diff
@@ -24,15 +24,28 @@ import (
 	"github.com/projectatomic/buildah/docker"
 )
 
+const (
+	// OCIv1ImageManifest is the MIME type of an OCIv1 image manifest,
+	// suitable for specifying as a value of the PreferredManifestType
+	// member of a CommitOptions structure.  It is also the default.
+	OCIv1ImageManifest  = v1.MediaTypeImageManifest
+	defaultManifestType = OCIv1ImageManifest
```
rhatdan: Not sure defaultManifestType adds any value here?

image.go
```diff
@@ -57,17 +70,30 @@ func (i *containerImageRef) NewImage(sc *types.SystemContext) (types.Image, erro
...
-			break
+	}
+	// Look for a supported format in the acceptable list.
+	if manifestType == "" {
+		for _, mt := range manifestTypes {
+			if mt == v1.MediaTypeImageManifest {
+				manifestType = mt
+				break
+			}
+			if mt == docker.V2S2MediaTypeManifest {
```
rhatdan: Why not combine these together into one statement?

rhatdan:
@mtrmac @runcom PTAL

cmd/buildah/bud.go
nalind: Changing it.

image.go
nalind: It's not exported, so probably not.  Dropping it.

image.go
nalind: No particularly good reason.  Changing it.

image.go
```diff
@@ -175,48 +224,93 @@ func (i *containerImageRef) NewImageSource(sc *types.SystemContext, manifestType
...
+			MediaType: dmediaType,
 			Digest:    destHasher.Digest(),
 			Size:      size,
 		}
-		manifest.Layers = append(manifest.Layers, layerDescriptor)
+		dmanifest.Layers = append(dmanifest.Layers, dlayerDescriptor)
 		lastLayerDiffID := destHasher.Digest().String()
-		image.RootFS.DiffIDs = append(image.RootFS.DiffIDs, lastLayerDiffID)
+		oimage.RootFS.DiffIDs = append(oimage.RootFS.DiffIDs, lastLayerDiffID)
+		dimage.RootFS.DiffIDs = append(dimage.RootFS.DiffIDs, destHasher.Digest())
```
mtrmac: The `DiffIDs` values should be digests of uncompressed data, AFAICS `destHasher` digests the possibly compressed stream.

config.go
```diff
@@ -117,9 +119,60 @@ func makeDockerV2S2Image(oimage *ociv1.Image) (docker.Image, error) {
...
+	// Build a filesystem history.
+	rootFS := &docker.V2S2RootFS{
+		Type:    docker.TypeLayers,
+		DiffIDs: []digest.Digest{},
+	}
+	history := []docker.V2S2History{}
+	for i := range manifest.FSLayers {
+		d := manifest.FSLayers[i].BlobSum
+		// Prepend this layer, because the original manifest's list has the earliest layer last.
+		rootFS.DiffIDs = append([]digest.Digest{d}, rootFS.DiffIDs...)
```
mtrmac: (Note that this is not correct in general: BlobSum is a digest of the possibly compressed version, DifffIDs of the uncompressed version. Is there a reason to expect the v2s1 image to only use uncompressed layers?)

mtrmac: The original upstream code does not add `V1Compatibility[].ThrowAway` layers to v2s2 `DiffIDs` and manifest; I’m not sure that matters.

image.go
mtrmac: (This has already been fixed to use `srcHasher`.)

config.go
nalind: Ah, so that explains the discrepancy in https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-1.md.  Fixing to decode v1 compat data correctly.

nalind: It's not an intentional limitation.  I don't see the digest of the uncompressed data in the manifest, though.  If we have to decompress or regenerate the layer data to get the right value, I'd be reluctant to do that.  We compute the right values when we're exporting the image, so if nobody's going to be reading the blatantly-wrong-if-compression-was-used value, then I'd lean hard in the direction of adding a comment about it here and leaving it otherwise.

mtrmac: > We compute the right values when we're exporting the image, so if nobody's going to be reading the blatantly-wrong-if-compression-was-used value, then I'd lean hard in the direction of adding a comment about it here and leaving it otherwise.

Works for me. (`docker`’s pull path does enforce that the uncompressed data match the digests in `DiffIDs`, so if it ever broke, it would be very visible.)

Or perhaps create an empty array, or an array of empty strings, to make extra certain that the incorrect values are never used by anything? Not necessary by any means.

nalind:
Updated parsing of `v1Compatibility` fields in schema 1 manifests to use the right type and preserve `ThrowAway` information.

I hadn't taken that close a look at format conversion in UpdatedImage(), so yeah, there's probably some duplication there.  I think we need that level of control over the manifest and configuration contents, though.
The `containerImageSource` implementation also recomputes the checksums of the compressed layers, which we need to do in order to push the layers anywhere.

config.go
nalind: I don't have a strong feeling about it either way.  If you do, I'm happy to change it to match.

mtrmac:
LGTM, to the very limited extent I understand the repo.

runcom:
LGTM as well

config.go
```diff
@@ -117,9 +119,63 @@ func makeDockerV2S2Image(oimage *ociv1.Image) (docker.Image, error) {
...
+		return docker.V2Image{}, fmt.Errorf("error parsing image configuration from history")
+	}
+	// Build a filesystem history.
+	rootFS := &docker.V2S2RootFS{
+		Type:    docker.TypeLayers,
+		DiffIDs: []digest.Digest{},
+	}
+	history := []docker.V2S2History{}
+	for i := range manifest.FSLayers {
+		// The DiffID list is intended to contain the sums of _uncompressed_ blobs, so if the blob is
```
TomSweeneyRedHat: I'm struggling a bit with this comment and probably a lack of docker internals knowledge on my part.  Could a blob be compressed and if so, is there a check that we need here before we prepend it?  I'm not seeing a mention of compression here: https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-1.md, but don't know if that's the right spot to look.

docker/types.go
```diff
@@ -108,6 +120,22 @@ type Config struct {
 	Shell           strslice.StrSlice   `json:",omitempty"` // Shell for shell-form of RUN, CMD, ENTRYPOINT
 }
 
+// github.com/docker/distribution/manifest/schema1/config_builder.go
```
TomSweeneyRedHat: I think this location may have changed to github.com/docker/distribution/blob/master/manifest/schema1/config_builder.go.  At least I'm finding the file there (added blob/master) and not at this location on the web.

image.go
```diff
@@ -53,17 +68,31 @@ func (i *containerImageRef) NewImage(sc *types.SystemContext) (types.Image, erro
 }
 
 func (i *containerImageRef) NewImageSource(sc *types.SystemContext, manifestTypes []string) (src types.ImageSource, err error) {
```
TomSweeneyRedHat: When maintaining two years hence, searching for 'i' will be a bear through out this function.  Any chance in your next pass through you could change this to at least 'img' to make it more markable?   I'd really love 'containerImage', but willing to settle for at least a few more characters.

TomSweeneyRedHat:
LGTM, a couple of possible touch ups for follow ons.

config.go
nalind: Blobs can be either data (such as a JSON-encoded config blob) or layers, and layers can be uncompressed or compressed.  Compression is usually done gzip or bzip2.  I expect bzip2 is rare (I've literally never seen it used) because the Go compress/bzip2 implementation can decompress but not compress.  Most layers in the wild are compressed with gzip compression.

For Docker v2 schema 2, they're tagged with one of the MIME types listed in https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md.  In OCI v1, they're tagged with one of the MIME types listed at https://github.com/opencontainers/image-spec/blob/master/media-types.md.  For Docker v2 schema 1, I think we don't have that information, but we can look at the initial segment of the blob and make a guess.  Problem is that we don't want to take the time to decompress a possibly-very-large blob to calculate the correct value, so I'm starting to lean toward @mtrmac's suggestion that it's better to leave the DiffID list empty to avoid creating errors if we ever try to use this data.

docker/types.go
nalind: The github web site's URL isn't going to perfectly mirror the tree layout that you get when cloning the repository, and I've been going for the latter here.

nalind:
Switched to leaving the diffID list empty when reading v2s1 images.

rhatdan:
@rh-atomic-bot r+  	a3f4a14

PR #117, opened 2017-05-22
Makefile: add source deps for the CLI tool

nalind:
The CLI tool also incorporates types from the 'docker' subdirectory, so add it as a build-time dependency to check.

PR #119, opened 2017-05-23
Increase the timeout for metalinter from 2m to 4m

nalind:
Increase the timeout that we set for running the metalinter from 2 minutes to 4 minutes, for cases where the calling environment is more heavily loaded than we expected.

As of this writing, this appears to be what's causing CI to reject #117.

### DIFF
--- a/tests/validate/gometalinter.sh
+++ b/tests/validate/gometalinter.sh
@@ -16,5 +16,5 @@ exec gometalinter.v1 \
 	--disable=gas \
 	--disable=aligncheck \
 	--cyclo-over=40 \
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM, just wondering if we should just bump it to 10 minutes to be extra safe.

nalind:
We might have to eventually, though I'm told that speediness is desirable for CI tests, so if we get to that point we might have to make other tradeoffs.

rhatdan:
@rh-atomic-bot r+  	5c734a7

PR #120, opened 2017-05-23
Cleanup: use FileInfo's own IsDir() method

nalind:
Use os.FileInfo's IsDir() method instead of Mode().IsDir(), h/t @rhatdan.


rhatdan:
@rh-atomic-bot r+ 6131340

nalind:
@rh-atomic-bot retry

PR #121, opened 2017-05-24
Use Hostname settings, avoid mounting tmpfs over volumes with mounts

nalind:
Go ahead and default to a hostname from the source image if it included one, and avoid mounting a tmpfs at a volume location if the list of mounts already includes an entry that mounts something else there.

### DIFF
--- a/run.go
+++ b/run.go
@@ -64,6 +64,58 @@ type RunOptions struct {
 	Terminal int
 }
 
+func setupMounts(spec *specs.Spec, optionMounts []specs.Mount, bindFiles []string, volumes []string) error {
+	// The passed-in mounts matter the most to us.
+	mounts := make([]specs.Mount, len(optionMounts))
+	copy(mounts, optionMounts)
+	haveMount := func(destination string) bool {
+		for _, mount := range mounts {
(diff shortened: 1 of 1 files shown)


run.go
```diff
@@ -185,6 +182,19 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 		}
 	}
 	spec.Mounts = append(mounts, boundMounts...)
+	for _, volume := range b.Volumes() {
+		override := false
+		for _, mount := range spec.Mounts {
+			if mount.Destination == volume {
+				// Already mounting something there, no need for a tmpfs.
+				override = true
+			}
```
rhatdan: Should break here.

nalind: I'm going to have to restructure this to get CI to accept it.

nalind:
Okay, the logic that sets the list of mounts in the runtime spec is now its own function.

rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+  	1eb1736

rhatdan:
@rh-atomic-bot r+  9e94205

PR #124, opened 2017-05-28
buildah mount command should list mounts when no arguments are given.

rhatdan:
buildah mount should work like the mount command and list all mount points
when no options are specified.

Need buildah umount to remove mount point from the database when a mount point
is umounted.

### DIFF
--- a/buildah.go
+++ b/buildah.go
@@ -68,9 +68,6 @@ type Builder struct {
 	// MountPoint is the last location where the container's root
 	// filesystem was mounted.  It should not be modified.
 	MountPoint string `json:"mountpoint,omitempty"`
-	// Mounts is a list of places where the container's root filesystem has
-	// been mounted.  It should not be modified.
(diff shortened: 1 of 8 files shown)


rhatdan:
Fixes https://github.com/projectatomic/buildah/issues/107

rhatdan:
@nalind We have a b.Mounts structure which I am not taking care of right now, Should we just remove this since we don't allow the container image to be mounted in more then one postition now?

nalind:
@rhatdan Yeah, the Mounts string slice in there is probably a leftover from when we dropped the various `--mount` flags, so I don't think we're doing anything with that any more.  Might as well lose it, then.

rhatdan:
Mounts has been removed.

nalind:
LGTM.
@rh-atomic-bot r+ b6c7475f0d7a2c2fbfd0cd81935c5611e770fba8

PR #127, opened 2017-05-30
Finish wiring --format into build-using-dockerfile

nalind:
We weren't properly passing the preferred output format to the Commit() method when committing images that we were building using build-with-dockerfile.  This fixes that and adds a test for it.


rhatdan:
@rh-atomic-bot r+  	b124b1a

PR #131, opened 2017-06-02
Help people figure out runc compatibility

nalind:
Output the versions of the image-spec and runtime-spec that we were built with as part of our version information, produced when we are invoked with the `--help` or `--version` flags, and note that we need a compatible version of `runc` in our README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,10 @@ Then to install buildah follow the steps in this example:
   buildah --help
 ```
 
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

nalind:
@rh-atomic-bot r+ 1538d5e04afa0842781a805104f712ad0a982ff2

PR #138, opened 2017-06-07
Fix inspect and tag to accept truncated image IDs

nalind:
Teach inspect about image IDs, truncated and otherwise.

### DIFF
--- a/import.go
+++ b/import.go
@@ -5,6 +5,7 @@ import (
 	"github.com/containers/image/types"
 	"github.com/containers/storage"
 	"github.com/pkg/errors"
+	"github.com/projectatomic/buildah/util"
 )
 
 func importBuilderDataFromImage(store storage.Store, systemContext *types.SystemContext, imageID, containerName, containerID string) (*Builder, error) {
@@ -85,20 +86,16 @@ func importBuilderFromImage(store storage.Store, options ImportFromImageOptions)
 		return nil, errors.Errorf("image name must be specified")
 	}
 
-	ref, err := is.Transport.ParseStoreReference(store, options.Image)
+	img, err := util.FindImage(store, options.Image)
 	if err != nil {
-		return nil, errors.Wrapf(err, "error parsing reference to image %q", options.Image)
-	}
-	img, err := is.Transport.GetStoreImage(store, ref)
-	if err != nil {
-		return nil, errors.Wrapf(err, "unable to locate image")
+		return nil, errors.Wrapf(err, "error locating image %q for importing settings", options.Image)
 	}
 
 	systemContext := getSystemContext(options.SignaturePolicyPath)
 
 	builder, err := importBuilderDataFromImage(store, systemContext, img.ID, "", "")
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrapf(err, "error importing build settings from image %q", options.Image)
 	}
 
 	return builder, nil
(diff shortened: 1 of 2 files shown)


nalind:
[(3) [buildah] Add a way to inspect working containers](https://trello.com/c/g5hLYG4M/501-3-buildah-add-a-way-to-inspect-working-containers)

import.go
```diff
@@ -91,6 +91,13 @@ func importBuilderFromImage(store storage.Store, options ImportFromImageOptions)
 	}
 	img, err := is.Transport.GetStoreImage(store, ref)
 	if err != nil {
+		img2, err2 := store.Image(options.Image)
```
rhatdan: If you do this with 
i
f img2, err := store.Image(options.Image); err == nil {
     img=img2
}
}

Does that override the outside err, or does it create a new instance only?

rhatdan: Other then this question, this LGTM.

nalind: It'd create a new variable in the scope of the block, leaving the `err` defined on line 92 unchanged.  We want to clear that one, though.  The "validate" make target would also fail because it'd trigger a warning from the 'vetshadow' linter.

rhatdan: Ok.  I was thinking you could have thrown the err from inside of the block preventing the second check.
```
if err2 == nil {
		img = img2
		err = nil
} else {
  		return nil, errors.Wrapf(err, "unable to locate image") 
}
```

rhatdan: Then no need to reset

nalind: Yeah, that'd work, too.  I thought one of the linters would complain about returning from one leg of an if-then-else (since you can rewrite it as if-then-return, and just have the else bit be the continuation), but it didn't.

rhatdan: Make the change into both and we can merge it.

nalind: Will do.

nalind:
Okay, I think this looks less complicated than what it looked like before, and merges in the changes that were in #139.

import.go
```diff
@@ -81,17 +81,23 @@ func importBuilder(store storage.Store, options ImportOptions) (*Builder, error)
...
+	var img *storage.Image
+
 	if options.Image == "" {
 		return nil, errors.Errorf("image name must be specified")
 	}
 
 	ref, err := is.Transport.ParseStoreReference(store, options.Image)
-	if err != nil {
-		return nil, errors.Wrapf(err, "error parsing reference to image %q", options.Image)
+	if err == nil {
```
TomSweeneyRedHat: We lose out on the  "error parsing reference to image"  error, any value to keeping it?  Even if only a debug message now.  I'm on the fence, thought I'd throw it up for consideration.

TomSweeneyRedHat:
The changes LGTM, test error is puzzling.  Look like a issue in the manpages?  Maybe a rebuild in hopes it's a hiccup?

import.go
nalind: Sure, we can infer based on whether or not `ref` contains a value.

nalind:
Cleared the "ineffectual assignment of err" that the linter caught, customized that error message based on whether or not we had a non-nil `ref`.

import.go
TomSweeneyRedHat: 👍

nalind:
Realized that at this point it's easier to have `importBuilderFromImage()` just call `util.FindImage()` instead of duplicating its logic exactly.

TomSweeneyRedHat:
LGTM, assuming happy tests.

rhatdan:
@rh-atomic-bot r+  	bc583a8

Issue #136, opened 2017-06-07
Unable to tag the image ID to new image


gouyang:
```
[root@fedora buildah]# ./buildah images
IMAGE ID     IMAGE NAME
b812babf7f65 docker.io/library/redis:latest
b812babf7f65 docker.io/gouyang/redis:latest
b812babf7f65 docker.io/testing/redis:latest
b812babf7f65 docker.io/foo/redis:latest
5965adf86a6e docker.io/library/nginx:latest
223c6ba7409d docker.io/kubernetes/pause:latest
e236280e3ef7 docker.io/library/redis:alpine
a9a3ca121508 registry.access.redhat.com/rhel7/rhel:latest
8dc56e5fe235 docker.io/library/busybox:latest
f8ca008a9d31 docker.io/library/busybox3:latest
d0a32bc20460 docker.io/library/redis2:latest
9334afa88fd3 
[root@fedora buildah]# ./buildah  tag 9334afa88fd3 gouyang/nginx
ERRO[0000] error finding local image "9334afa88fd3": unable to locate image: image not known 
```

rhatdan:
Any idea how you created an image without a tag?  If you specify the untruncated name does it work?

gouyang:
> Any idea how you created an image without a tag? 

`./buildah bud .` 

> If you specify the untruncated name does it work?

```
# ./buildah tag 9334afa88fd31ec8f0bff28cd4a435748c22200824337b8bc2818d2363d02d46 gouyang/nginx
ERRO[0000] error finding local image "9334afa88fd31ec8f0bff28cd4a435748c22200824337b8bc2818d2363d02d46": error parsing reference to image "9334afa88fd31ec8f0bff28cd4a435748c22200824337b8bc2818d2363d02d46": invalid repository name (9334afa88fd31ec8f0bff28cd4a435748c22200824337b8bc2818d2363d02d46), cannot specify 64-byte hexadecimal strings 
```

rhatdan:
Nalin I am not sure if this is a problem in buildah bud not adding a default tag, or in the lookup function, which is looking for a tag with
docker.io//library/ID/latest

rhatdan:
@nalind This is something we need to fix ASAP though. since it can leave your database in a weird state.

Well maybe not that bad since
./buildah rmi ID 

Does work.

gouyang:
It works now.

```
[root@fedora buildah]# ./buildah  images
IMAGE ID     IMAGE NAME
[...]

PR #139, opened 2017-06-07
Fix tag-by-truncated-image-ID

nalind:
Teach tag about image IDs, truncated and otherwise.


nalind:
It most likely would have been, but I wasn't aware of this when opening #138.

nalind:
Should fix #136.

nalind:
Merging into #138.

PR #141, opened 2017-06-09
Split copying to external locations ("pushing") from "commit"

nalind:
This patch set modifies `Commit()` to take shortcuts when it notices that the location to which we're committing the new image is local storage: the manifest blob sums are not recalculated, and a bit of lower-level trickery is used to copy only the container's most recent layer to a read-only layer that becomes the top level for the new image.  A separate "push" path that never takes those shortcuts is added so that when we want to export things, we actually do the necessary digesting and copying.

### DIFF
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -3,6 +3,7 @@ package main
 import (
 	"os"
 	"strings"
+	"time"
 
 	"github.com/containers/image/storage"
 	"github.com/containers/image/transports/alltransports"
@@ -15,7 +16,7 @@ import (
 var (
 	commitFlags = []cli.Flag{
 		cli.BoolFlag{
-			Name:  "disable-compression",
+			Name:  "disable-compression, D",
 			Usage: "don't compress layers",
 		},
 		cli.StringFlag{
@@ -25,6 +26,11 @@ var (
 		cli.StringFlag{
 			Name:  "format, f",
 			Usage: "`format` of the image manifest and metadata",
+		},
+		cli.StringFlag{
+			Name:   "reference-time",
+			Usage:  "set the timestamp on the image to match the named `file`",
+			Hidden: true,
 		},
 		cli.BoolFlag{
 			Name:  "quiet, q",
@@ -73,6 +79,15 @@ func commitCmd(c *cli.Context) error {
 	if c.IsSet("format") {
 		format = c.String("format")
 	}
+	timestamp := time.Now().UTC()
+	if c.IsSet("reference-time") {
+		referenceFile := c.String("reference-time")
+		finfo, err := os.Stat(referenceFile)
+		if err != nil {
+			return errors.Wrapf(err, "error reading timestamp of file %q", referenceFile)
+		}
+		timestamp = finfo.ModTime().UTC()
+	}
 	if strings.HasPrefix(strings.ToLower(format), "oci") {
 		format = buildah.OCIv1ImageManifest
 	} else if strings.HasPrefix(strings.ToLower(format), "docker") {
@@ -103,6 +118,7 @@ func commitCmd(c *cli.Context) error {
 		PreferredManifestType: format,
 		Compression:           compress,
 		SignaturePolicyPath:   signaturePolicy,
+		HistoryTimestamp:      &timestamp,
 	}
 	if !quiet {
 		options.ReportWriter = os.Stderr
--- a/cmd/buildah/push.go
+++ b/cmd/buildah/push.go
@@ -1,0 +1,84 @@
+package main
+
+import (
+	"os"
+
+	"github.com/containers/image/transports/alltransports"
+	"github.com/containers/storage/pkg/archive"
+	"github.com/pkg/errors"
+	"github.com/projectatomic/buildah"
+	"github.com/urfave/cli"
+)
+
+var (
+	pushFlags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "disable-compression, D",
+			Usage: "don't compress layers",
+		},
+		cli.StringFlag{
+			Name:  "signature-policy",
+			Usage: "`pathname` of signature policy file (not usually used)",
+		},
+		cli.BoolFlag{
+			Name:  "quiet, q",
+			Usage: "don't output progress information when pushing images",
+		},
+	}
+	pushDescription = "Pushes an image to a specified location."
+	pushCommand     = cli.Command{
+		Name:        "push",
+		Usage:       "Push an image to a specified location",
+		Description: pushDescription,
+		Flags:       pushFlags,
+		Action:      pushCmd,
+		ArgsUsage:   "IMAGE [TRANSPORT:]IMAGE",
+	}
+)
+
+func pushCmd(c *cli.Context) error {
+	args := c.Args()
+	if len(args) < 2 {
+		return errors.New("source and destination image IDs must be specified")
+	}
+	src := args[0]
+	destSpec := args[1]
+
+	signaturePolicy := ""
+	if c.IsSet("signature-policy") {
+		signaturePolicy = c.String("signature-policy")
+	}
+	compress := archive.Uncompressed
+	if !c.IsSet("disable-compression") || !c.Bool("disable-compression") {
+		compress = archive.Gzip
+	}
+	quiet := false
+	if c.IsSet("quiet") {
+		quiet = c.Bool("quiet")
+	}
+
+	store, err := getStore(c)
+	if err != nil {
+		return err
+	}
+	dest, err := alltransports.ParseImageName(destSpec)
+	if err != nil {
+		return err
+	}
+
+	options := buildah.PushOptions{
+		Compression:         compress,
+		SignaturePolicyPath: signaturePolicy,
+		Store:               store,
+	}
+	if !quiet {
+		options.ReportWriter = os.Stderr
+	}
+
+	err = buildah.Push(src, dest, options)
+	if err != nil {
+		return errors.Wrapf(err, "error pushing image %q to %q", src, destSpec)
+	}
+
+	return nil
+}
--- a/commit.go
+++ b/commit.go
@@ -8,9 +8,10 @@ import (
 	"github.com/Sirupsen/logrus"
 	cp "github.com/containers/image/copy"
 	"github.com/containers/image/signature"
-	"github.com/containers/image/storage"
+	is "github.com/containers/image/storage"
 	"github.com/containers/image/transports"
 	"github.com/containers/image/types"
+	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/stringid"
 	"github.com/opencontainers/go-digest"
@@ -56,6 +57,25 @@ type CommitOptions struct {
 	HistoryTimestamp *time.Time
 }
 
+// PushOptions can be used to alter how an image is copied somewhere.
+type PushOptions struct {
+	// Compression specifies the type of compression which is applied to
+	// layer blobs.  The default is to not use compression, but
+	// archive.Gzip is recommended.
+	Compression archive.Compression
+	// SignaturePolicyPath specifies an override location for the signature
+	// policy which should be used for verifying the new image as it is
+	// being written.  Except in specific circumstances, no value should be
+	// specified, indicating that the shared, system-wide default policy
+	// should be used.
+	SignaturePolicyPath string
+	// ReportWriter is an io.Writer which will be used to log the writing
+	// of the new image.
+	ReportWriter io.Writer
+	// Store is the local storage store which holds the source image.
+	Store storage.Store
+}
+
 // shallowCopy copies the most recent layer, the configuration, and the manifest from one image to another.
 // For local storage, which doesn't care about histories and the manifest's contents, that's sufficient, but
 // almost any other destination has higher expectations.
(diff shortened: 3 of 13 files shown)


rhatdan:
https://github.com/mtrmac @runcom PTAL

rhatdan:
@mtrmac PTAL

image.go
```diff
@@ -70,36 +75,34 @@ func (i *containerImageRef) NewImage(sc *types.SystemContext) (types.Image, erro
...
 
-func (i *containerImageRef) NewImageSource(sc *types.SystemContext, manifestTypes []string) (src types.ImageSource, err error) {
-	manifestType := ""
-	// If we have a preferred format, and it's in the acceptable list, select that one.
-	for _, mt := range manifestTypes {
-		if mt == i.preferredManifestType {
-			manifestType = mt
-			break
+func selectManifestType(preferred string, acceptable, supported []string) string {
+	for _, accept := range acceptable {
```
rhatdan: Could make this a bit more efficient.
```
default : = preferred
for _, accept := range acceptable {
                if preferred == accept {
			return preferred
 		}
                for _, support := range supported {
			if accept == support {
				default = accept
 			}
 		}
}
return default

File: commit.go [hunk 2]
```diff
@@ -38,6 +43,187 @@ type CommitOptions struct {
...
+	}
+	defer func() {
+		if err2 := tmpRef.DeleteImage(systemContext); err2 != nil {
+			logrus.Debugf("error deleting temporary image %q: %v", tmpName, err2)
+		}
+	}()
+	// Make a temporary directory to use as a blob source.
+	path, err := ioutil.TempDir(os.TempDir(), Package)
+	if err != nil {
+		return err
```
rhatdan: Should we wrap to say that we are in shallowCopy?

commit.go
```diff
@@ -38,6 +43,187 @@ type CommitOptions struct {
...
+	// Build an empty filesystem layer.
+	rc, err := archive.Tar(path, archive.Gzip)
+	if err != nil {
+		return errors.Wrapf(err, "error reading dummy layer")
+	}
+	defer rc.Close()
+	// Open the source for reading and a temporary image for writing.
+	srcImage, err := src.NewImageSource(systemContext, nil)
+	if err != nil {
+		return errors.Wrapf(err, "error computing layer digest and building metadata")
```
rhatdan: Should we include image Name?

commit.go
```diff
@@ -38,6 +43,187 @@ type CommitOptions struct {
...
+	}
+	defer srcImage.Close()
+	configImage, err := src.NewImage(systemContext)
+	if err != nil {
+		return errors.Wrapf(err, "error reading configuration for image")
```
rhatdan: Image Name?

commit.go
```diff
@@ -38,6 +43,187 @@ type CommitOptions struct {
...
+	}
+	defer configImage.Close()
+	tmpImage, err := tmpRef.NewImageDestination(systemContext)
+	if err != nil {
+		return errors.Wrapf(err, "error opening temporary image for writing")
```
rhatdan: Ditto

rhatdan: As well as the rest.

commit.go
```diff
@@ -36,6 +43,209 @@ type CommitOptions struct {
...
+	// of the new image.
+	ReportWriter io.Writer
+	// Store is the local storage store which holds the source image.
+	Store storage.Store
+}
+
+// shallowCopy copies the most recent layer, the configuration, and the manifest from one image to another.
+// For local storage, which doesn't care about histories and the manifest's contents, that's sufficient, but
+// almost any other destination has higher expectations.
+func (b *Builder) shallowCopy(dest types.ImageReference, src types.ImageReference, systemContext *types.SystemContext) error {
```
TomSweeneyRedHat: just seems odd to have destination as the first argument, then src as the second.  I think you've mostly gone in the other order in most other functions.

commit.go
TomSweeneyRedHat: 👍

commit.go
TomSweeneyRedHat: 👍  and the other image names below too.

commit.go
```diff
@@ -36,6 +43,209 @@ type CommitOptions struct {
...
+		_, err2 := b.store.DeleteImage(image.ID, true)
+		if err2 != nil {
+			logrus.Debugf("error removing image %q: %v", image.ID, err2)
+		}
+		return errors.Wrapf(err, "error assigning metadata to new image")
+	}
+	// Move the target name(s) from the temporary image to the new image.
+	err = util.AddImageNames(b.store, image, names)
+	if err != nil {
+		_, err2 := b.store.DeleteImage(image.ID, true)
```
runcom: This block is repeated and could be very well done in a defer

contrib/completions/bash/buildah
```diff
@@ -437,7 +438,7 @@ return 1
      esac
  }
 
- _buildah_umount() {
+ _buildah_unmount() {
      _buildah_umount $@
```
TomSweeneyRedHat: does this one need to change too?

cmd/buildah/push.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/containers/storage/pkg/archive"
+	"github.com/pkg/errors"
+	"github.com/projectatomic/buildah"
+	"github.com/urfave/cli"
+)
+
+var (
+	pushFlags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "disable-compression",
```
rhatdan: Should we have a short?  -D?

cmd/buildah/push.go
```diff
@@ -0,0 +1,84 @@
...
+		cli.StringFlag{
+			Name:  "signature-policy",
+			Usage: "`pathname` of signature policy file (not usually used)",
+		},
+		cli.BoolFlag{
+			Name:  "quiet, q",
+			Usage: "don't output progress information when pushing images",
+		},
+	}
+	pushDescription = "Copies an image to a specified location."
```
rhatdan: Pushes an image ...

cmd/buildah/push.go
```diff
@@ -0,0 +1,84 @@
...
+	pushCommand     = cli.Command{
+		Name:        "push",
+		Usage:       "Copy an image to a specified location",
```
rhatdan: Pushes...

contrib/completions/bash/buildah
nalind: That would create a recursion problem in the shell function.  To be fair, this hunk really belongs in its own PR.  I'll move it to one.

nalind: Moved it to #143.

commit.go
nalind: It's basically arbitrary, but in this case it's using the same order as the image library's `Copy`.

cmd/buildah/commit.go
```diff
@@ -26,6 +27,10 @@ var (
 			Name:  "format, f",
 			Usage: "`format` of the image manifest and metadata",
 		},
+		cli.StringFlag{
+			Name:  "reference-time",
+			Usage: "set the timestamp on the image to match the named `file`",
```
rhatdan: If this is for self tests, can we hide the option and not document or command completion it.

image.go
nalind: Changing it.

commit.go
nalind: Changing.

commit.go
nalind: Adding it.

commit.go
nalind: Adding it.

commit.go
nalind: Adding them.

cmd/buildah/commit.go
nalind: The self-test for this checks that we get the same results from commit-then-push and the original push-while-committing paths, and after a while, the timestamps were the only difference in the results, so we needed some way to control them.  Using a reference file rather than offering to parse a date seemed more scripting-friendly (`touch` already does that, arguably better than we can), but it was a guess.

rhatdan: Lets just hide it, then we don't have to explain or support it.

cmd/buildah/push.go
nalind: No harm in adding one.

cmd/buildah/push.go
nalind: Changing it.

contrib/completions/bash/buildah
rhatdan: I am merging #143

cmd/buildah/push.go
nalind: Changing it.

cmd/buildah/commit.go
nalind: Okay, changing it.

commit.go
nalind: Changing it.

nalind:
Updated.

image.go
```diff
@@ -425,16 +483,63 @@ func (b *Builder) makeContainerImageRef(manifestType string, compress archive.Co
...
+		created:               created,
 		createdBy:             b.CreatedBy(),
 		annotations:           b.Annotations(),
 		preferredManifestType: manifestType,
+		exporting:             exporting,
+	}
+	return ref, nil
+}
+
+func (b *Builder) makeImageImageRef(compress archive.Compression, names []string, layerID string) (types.ImageReference, error) {
```
mtrmac: This and `makeContainerImageRef` seems to share a enough code that it might be possible to extract a shared helper from the two. OTOH the code is simple enough that this might not be worth it, I have no idea.

commit.go
```diff
@@ -36,6 +43,202 @@ type CommitOptions struct {
...
+		if err2 := tmpRef.DeleteImage(systemContext); err2 != nil {
+			logrus.Debugf("error deleting temporary image %q: %v", tmpName, err2)
+		}
+	}()
+	// Make a temporary directory to use as a blob source.
+	path, err := ioutil.TempDir(os.TempDir(), Package)
+	if err != nil {
+		return errors.Wrapf(err, "error creating temporary directory under %q", os.TempDir())
+	}
+	logrus.Debugf("using %q to hold temporary data", path)
```
nalind: Ooh, I like that idea.  Switching to that.

commit.go
```diff
@@ -36,6 +43,202 @@ type CommitOptions struct {
...
+	// of the new image.
+	ReportWriter io.Writer
+	// Store is the local storage store which holds the source image.
+	Store storage.Store
+}
+
+// shallowCopy copies the most recent layer, the configuration, and the manifest from one image to another.
+// For local storage, which doesn't care about histories and the manifest's contents, that's sufficient, but
+// almost any other destination has higher expectations.
+func (b *Builder) shallowCopy(dest types.ImageReference, src types.ImageReference, systemContext *types.SystemContext) error {
```
nalind: Adding a note to the comment: `We assume that "dest" is a reference to a local image (specifically, a containers/image/storage.storageReference), and will fail if it isn't.`

commit.go
```diff
@@ -36,6 +43,209 @@ type CommitOptions struct {
...
+		}
+	}()
+	// Build an empty filesystem layer.
+	rc, err := archive.Tar(path, archive.Gzip)
+	if err != nil {
+		return errors.Wrapf(err, "error reading dummy layer")
+	}
+	defer rc.Close()
+	// Open the source for reading and a temporary image for writing.
+	srcImage, err := src.NewImageSource(systemContext, nil)
```
nalind: Ah, right!  Making both of those changes.

commit.go
```diff
@@ -50,28 +253,84 @@ func (b *Builder) Commit(dest types.ImageReference, options CommitOptions) error
 	if err != nil {
 		return err
 	}
-	src, err := b.makeContainerImageRef(options.PreferredManifestType, options.Compression)
+	// Check if we're keeping everything in local storage.  If so, we can take certain shortcuts.
+	exporting := dest.Transport().Name() != is.Transport.Name()
```
nalind: I'm not wedded to string comparison there, so making that change.

commit.go
nalind: Yeah, a different version of this might have the image library provide a way for us drop layers on top of an existing image when writing a new one.  And if we need to modify `containers/image/storage.StorageImageDestination` at some point, I could definitely see the use of adding that.

image.go
nalind: No harm in factoring that out, though.

nalind:
Updated.

commit.go
rhatdan: Is this something we should fix in this pull request or open an issue to fix it later?

rhatdan: :+1:

image.go
rhatdan: :+1:

rhatdan:
LGTM

commit.go
mtrmac: From my POV, \*shrug\* this is fragile but it can work just fine. Whether to open an issue still depends on how seriously you care about that kind of architectural cleanliness, _if_ you agree at all that centralizing the mechanisms for storing images in containers/storage would be cleaner.

I personally would lean towards not tracking this as an issue. If a real need to refactor arises, someone will know :)

rhatdan:
@rh-atomic-bot r+  	3bbcd2f

rhatdan:
@rh-atomic-bot r+ 2f72163

PR #143, opened 2017-06-12
Fix "unmount" bash completion

nalind:
Fix the infinitely self-recursing, but immediately replaced, version of the "umount" completion function with the missing "unmount" completion function.


rhatdan:
@rh-atomic-bot r+  	8a718c2

PR #146, opened 2017-06-14
Update the version number we identify as to 0.1.

nalind:
Adjust the order in which `--help` output lists commands to keep them sorted, and change our claimed version number to 0.1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ buildah - a tool which facilitates building OCI container images
 [![Go Report Card](https://goreportcard.com/badge/github.com/projectatomic/buildah)](https://goreportcard.com/report/github.com/projectatomic/buildah)
 [![Travis](https://travis-ci.org/projectatomic/buildah.svg?branch=master)](https://travis-ci.org/projectatomic/buildah)
 
-Note: this package is in alpha.
+Note: this package is in alpha, but is close to being feature-complete.
 
 The buildah package provides a command line tool which can be used to
 * create a working container, either from scratch or using an image as a starting point
--- a/buildah.go
+++ b/buildah.go
@@ -19,7 +19,7 @@ const (
 	// identify working containers.
 	Package = "buildah"
 	// Version for the Package
-	Version       = "0.0.1"
+	Version       = "0.1"
 	containerType = Package + " 0.0.1"
 	stateFile     = Package + ".json"
 )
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -4,7 +4,7 @@ buildah - a tool which facilitates building OCI container images
 [![Go Report Card](https://goreportcard.com/badge/github.com/projectatomic/buildah)](https://goreportcard.com/report/github.com/projectatomic/buildah)
 [![Travis](https://travis-ci.org/projectatomic/buildah.svg?branch=master)](https://travis-ci.org/projectatomic/buildah)
 
-Note: this package is in alpha.
+Note: this package is in alpha, but is close to being feature-complete.
```
rhatdan: I think we should say we are entering alpha?

buildah.go
```diff
@@ -19,7 +19,7 @@ const (
 	// identify working containers.
 	Package = "buildah"
 	// Version for the Package
-	Version       = "0.0.1"
+	Version       = "0.1"
```
rhatdan: Should we add a rc1?

cmd/buildah/main.go
```diff
@@ -86,8 +86,8 @@ func main() {
 		rmCommand,
 		rmiCommand,
 		runCommand,
-		umountCommand,
 		tagCommand,
+		umountCommand,
```
rhatdan: Cheeting, but this is fine.

buildah.go
```diff
@@ -19,7 +19,7 @@ const (
...
 	containerType = Package + " 0.0.1"
```
TomSweeneyRedHat: does 0.0.1 need to change in line 23 too?

rhatdan: Can't Container-type be Package + Version

rhatdan:
We will be updating the actual spec file in Fedora/RHEL, so this is just an example.  But I see no reason not to update it.

buildah.go
nalind: It can be.  We use this to check if the data we've stored for a working container looks like it matches our version, and that data structure hasn't changed incompatibly since we moved to keeping both docker-style and OCI-style configuration structures separately.

README.md
nalind: It previously said it was in alpha, though.

buildah.go
rhatdan: So are we leaving this alone? Or modifying it.  If we leave it alone, you need to add a comment explaining what it is used for?

nalind: Adding a comment about that.

rhatdan:
@rh-atomic-bot r+  	d25bb49

PR #147, opened 2017-06-14
Add a quick note about state version values

nalind:
Here's that part of #146 that I didn't get in fast enough.

### DIFF
--- a/buildah.go
+++ b/buildah.go
@@ -19,9 +19,17 @@ const (
 	// identify working containers.
 	Package = "buildah"
 	// Version for the Package
-	Version       = "0.1"
+	Version = "0.1"
+	// The value we use to identify what type of information, currently a
+	// serialized Builder structure, we are using as per-container state.
+	// This should only be changed when we make incompatible changes to
+	// that data structure, as it's used to distinguish containers which
+	// are "ours" from ones that aren't.
 	containerType = Package + " 0.0.1"
-	stateFile     = Package + ".json"
+	// The file in the per-container directory which we use to store our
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+ 79b0610

buildah.go
```diff
@@ -19,9 +19,17 @@ const (
...
 	Package = "buildah"
 	// Version for the Package
-	Version       = "0.1"
+	Version = "0.1"
+	// The value we use to identify what type of information, currently a
+	// serialized Builder structure, we are using as per-container state.
+	// This should only be changed when we make incompatible changes to
+	// that data structure, as it's used to distinguish containers which
+	// are "ours" from ones that aren't.
 	containerType = Package + " 0.0.1"
```
TomSweeneyRedHat: IDK as I'm flip fllopping, so I'll just ask.   Should/do we need to state the basis of the Builder structure?  i.e. based on Docker's Builder structure or whatever?  Do we want to identify who "They" is somewhere in here?  I'm getting at this as we might make incompatible changes, but isn't it likely "They" might too?

nalind: We're talking about the parts of the structure that we pick up from other packages (the OCIv1 configuration structure), right?  That's defined as part of the image-spec, and I guess we could stand to call it out.

TomSweeneyRedHat: If you get a chance to add verbiage later about the ociv1 spec here, I think it would make it clear, thx.

PR #148, opened 2017-06-14
Add a -a/--all option to "buildah containers"

nalind:
Add a --all option to "buildah containers" that causes it to go through the full list of containers, providing information about the ones that aren't buildah containers in addition to the ones that are.

### DIFF
--- a/cmd/buildah/containers.go
+++ b/cmd/buildah/containers.go
@@ -22,6 +22,10 @@ var (
 			Name:  "notruncate",
 			Usage: "do not truncate output",
 		},
+		cli.BoolFlag{
+			Name:  "all, a",
+			Usage: "also list non-buildah containers",
+		},
 	}
 	containersDescription = "Lists containers which appear to be " + buildah.Package + " working containers, their\n   names and IDs, and the names and IDs of the images from which they were\n   initialized"
 	containersCommand     = cli.Command{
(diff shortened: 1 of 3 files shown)


cmd/buildah/containers.go
```diff
@@ -64,18 +72,57 @@ func containersCmd(c *cli.Context) error {
...
 			if truncate {
-				fmt.Printf("%-12.12s %-12.12s %-10s %s\n", builder.ContainerID, builder.FromImageID, builder.FromImage, builder.Container)
+				fmt.Printf("%-12.12s %-12.12s %-10s %s\n", containerID, imageID, image, container)
+			} else {
+				fmt.Printf("%-64s %-64s %-10s %s\n", containerID, imageID, image, container)
+			}
+		}
+	}
+	seenContainers := make(map[string]struct{})
+	for _, builder := range builders {
```
rhatdan: Would this list some containers twice?  Should we have build container interleafed with regular containers?

nalind: We add IDs to the `seenContainers` map when we list them normally, and use that to skip them when we're walking the whole list.

rhatdan: I think there should be an indicator of a buildcontainer and not a build container then.
Does the list indicate anything. Oldest to newest?  

I rely on docker images and docker ps giving the newest at the top,

nalind: We don't do any specific sorting of the lists, but this PR ends up listing buildah containers before other types.  I don't think we're currently tracking creation times for... well, anything.

TomSweeneyRedHat: Just wondering if it makes sense to do the list of just buildah containers (lines ~75->95) under an "if !all" statement then make the current "if all" an else for that and remove the seenImages functionality.  I'm not sure how painful the lookups could be under a big load.

rhatdan: I would like that better.  The seen seems a little easy for someone in the future to screw up.

nalind: Okay, changing it, using inlines to factor some more things out.

nalind:
Updated.

docs/buildah-containers.md
```diff
@@ -24,6 +24,11 @@ Do not truncate IDs in output.
 
 Displays only the container IDs.
 
+**--all, -a**
+
+Also list information about containers which don't appear to be buildah working
```
rhatdan: Nit. Remove Also.

`List information about all containers including containers images which were not created by buildah.`

rhatdan:
I still think the list should contain and indicator that this container was not created by buildah.

docs/buildah-containers.md
nalind: Making that change.

nalind: Changed the description to `List information about all containers, including those which were not created by and are not being used by buildah.`

nalind:
Inserted a column after the first which marks buildah containers with an asterisk.

rhatdan:
@rh-atomic-bot r+  	9165d4c

TomSweeneyRedHat:
LGTM, thanks a bunch for adding the "Mark of Buildah!"

PR #149, opened 2017-06-15
Make tag names for "buildah bud" optional again

nalind:
Go back to having the set of names to assign to the resulting image being an optional argument when building a package using a Dockerfile.


nalind:
Needs work.

PR #150, opened 2017-06-15
Make shallowCopy() not use a temporary image

nalind:
Modify `shallowCopy()` to not use a temporary image.
We assume that the big data items that we formerly added to the temporary image are small enough that we can just hang on to them.
We now write everything to the destination image instead of a temporary image, read it all back using the low level APIs, delete the image, and then recreate the image using the new layer and the saved items and names.
This lets us lift the requirement that we `shallowCopy()` only to images with names, so that `build-using-dockerfile` will work again when it's not being given tags.

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -13,7 +13,6 @@ import (
 	"github.com/containers/image/types"
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/archive"
-	"github.com/containers/storage/pkg/stringid"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	"github.com/projectatomic/buildah/util"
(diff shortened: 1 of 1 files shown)


commit.go
```diff
@@ -82,38 +81,27 @@ type PushOptions struct {
...
-	tmpName := stringid.GenerateRandomID() + "-tmp-" + Package + "-commit"
-	tmpRef, err := is.Transport.ParseStoreReference(b.store, tmpName)
-	if err != nil {
-		return err
-	}
-	defer func() {
-		if err2 := tmpRef.DeleteImage(systemContext); err2 != nil {
-			logrus.Debugf("error deleting temporary image %q: %v", tmpName, err2)
+	if dest.DockerReference() != nil {
+		expandedNames, err := util.ExpandTags([]string{dest.DockerReference().String()})
```
rhatdan: Why do you need expandedNames at all, why not just set it to names?

rhatdan:
Test failures also.

commit.go
nalind: Oh, right.  The name in the reference should already have been canonicalized during parsing.

rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+  	e783ffd

Issue #152, opened 2017-06-19
Volume in the created image has different permissions


vrutkovs:
buildah version 0.1 (image-spec 1.0.0-rc5, runtime-spec 1.0.0-rc5) from buildah-0.1.0-2.git597d2ab.fc26.x86_64

Dockerfile:
```
FROM busybox
RUN mkdir /vol && touch /vol/foo
VOLUME /vol
RUN touch /vol/bar
CMD ["cat", "/vol/bar"]
```

Permissions in the image using docker build are `drwxr-xr-x`:
```
$ sudo docker run -ti test/04:dockerapi ls -la /vol
total 8
drwxr-xr-x    2 root     root          4096 Jun 19 18:00 .
```

Buildah image has `drw-------`:
```
$ sudo docker run -ti test/04:buildah ls -la /vol
total 8
drw-------    2 root     root          4096 Jun 19 18:03 .
```

vrutkovs:
Closed by #154

Issue #151, opened 2017-06-19
Dockerfile with VOLUME instruction fails to build


vrutkovs:
buildah version 0.1 (image-spec 1.0.0-rc5, runtime-spec 1.0.0-rc5) from buildah-0.1.0-2.git597d2ab.fc26.x86_64

Dockerfile fails to build:
```
FROM busybox
RUN echo "a" > /a
VOLUME /b
CMD ["cat", "/a"]
```
```
$ buildah --storage-driver overlay2 bud --tag test/03:buildah /home/vrutkovs/src/image-comparison/tests/03
STEP 1: FROM busybox
STEP 2: RUN echo "a" > /a
STEP 3: VOLUME /b
time="2017-06-19T19:40:19+02:00" level=error msg="error building: error building at step {Env:[PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin] Command:volume Args:[/b] Flags:[] Attrs:map[] Message:VOLUME /b Original:VOLUME /b}: stat /var/lib/containers/storage/overlay2/cf989ed6334eeb5a90874264c960220306577d7a07d0d0a48ff933d8ff272d7d/merged/b: no such file or directory"
```

rhatdan:
Since this patch got merged this issue should be fixed.

PR #154, opened 2017-06-20
Ensure volume points get created, and with perms

nalind:
Ensure that volume mount points are created, if they don't exist, when they're defined in a Dockerfile (#151), and that if we create them, we create them with 0755 permissions (#152).

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -153,7 +153,15 @@ func (b *Executor) Preserve(path string) error {
 	logrus.Debugf("PRESERVE %q", path)
 	if b.volumes.Covers(path) {
 		// This path is already a subdirectory of a volume path that
-		// we're already preserving, so there's nothing new to be done.
+		// we're already preserving, so there's nothing new to be done
+		// except ensure that it exists.
+		archivedPath := filepath.Join(b.mountPoint, path)
+		if err := os.MkdirAll(archivedPath, 0755); err != nil {
+			return errors.Wrapf(err, "error ensuring volume path %q exists", archivedPath)
+		}
+		if err := b.volumeCacheInvalidate(path); err != nil {
+			return errors.Wrapf(err, "error ensuring volume path %q is preserved", archivedPath)
+		}
 		return nil
 	}
 	// Figure out where the cache for this volume would be stored.
@@ -166,9 +174,15 @@ func (b *Executor) Preserve(path string) error {
 	// Save info about the top level of the location that we'll be archiving.
 	archivedPath := filepath.Join(b.mountPoint, path)
 	st, err := os.Stat(archivedPath)
+	if os.IsNotExist(err) {
+		if err = os.MkdirAll(archivedPath, 0755); err != nil {
+			return errors.Wrapf(err, "error ensuring volume path %q exists", archivedPath)
+		}
+		st, err = os.Stat(archivedPath)
+	}
 	if err != nil {
 		logrus.Debugf("error reading info about %q: %v", archivedPath, err)
-		return err
+		return errors.Wrapf(err, "error reading info about volume path %q", archivedPath)
 	}
 	b.volumeCacheInfo[path] = st
 	if !b.volumes.Add(path) {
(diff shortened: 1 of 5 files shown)


imagebuildah/build.go
```diff
@@ -241,6 +255,9 @@ func (b *Executor) volumeCacheSave() error {
 		if !os.IsNotExist(err) {
 			return errors.Wrapf(err, "error checking for cache of %q in %q", archivedPath, cacheFile)
 		}
+		if err := os.MkdirAll(archivedPath, 0755); err != nil {
```
rhatdan: Will this work on a READONLY Image?

rhatdan:
Does buildah ignore when a dockerfile attempts to write to the directory or if content is created in the directory.  I guess I am not sure what happens in docker when this happens.

rhatdan:
@runcom @mrunalp PTAL

runcom:
In CRI-O we create those VOLUME(s) with 0644 actually https://github.com/kubernetes-incubator/cri-o/blob/master/server/container_create.go#L610 :confused:

rhatdan:
They should be consistent, although creating a directory as 644 means that you could not search through it, 755, is probably correct.

mrunalp:
I just tested volume behavior with docker:
```
[root@localhost test_volume]# docker build -t test_volume .
Sending build context to Docker daemon 2.048 kB
Step 1 : FROM busybox:latest
 ---> e02e811dd08f
Step 2 : VOLUME /mydata
 ---> Running in c6a07ccebdfd
 ---> 7af942b3d1e1
Removing intermediate container c6a07ccebdfd
Step 3 : RUN echo hi > /mydata/hi.txt
 ---> Running in 0cda24be6cfb
 ---> a893656b0a28
Removing intermediate container 0cda24be6cfb
Successfully built a893656b0a28
[root@localhost test_volume]# docker run -it --rm test_volume sh
/ # ls -l /mydata/
total 0
/ # 
```
So, it looks like docker isn't saving contents written to a volume during a build. This would mean that we would want to use bind mounts here. Otherwise, we cannot mimic this behavior if there already existed a directory in the image that a volume is added over.

rhatdan:
I think in CRI-O they are mounting a tmpfs there now, which would probably work for buildah.  CRI-O is looking to handle it differently since OpenShift wants us to stop users from using storage outside of volumes handed to the user.

mrunalp:
We are also doing a mkdir in CRI-O right now, but I plan to make it all configurable as part of my current work. There could be these options in the config for admin:
1. ImageVolume=ignore
2. ImageVolume=tmpfs
3. ImageVolume=mount
4. ImageVolume=directory

nalind:
@mrunalp Apparently RUN instructions which modify the contents of a directory that's been marked as a volume are discarded, but contents can still be added using ADD or COPY instructions (noted at https://github.com/openshift/imagebuilder/blob/master/builder.go#L213, and experimentally confirmed).

imagebuildah/build.go
nalind: At this point we should be operating on a read-write layer, so I don't expect whether or not the original image is on a writable filesystem to matter.

mrunalp:
@nalind okay, I guess it is hard to support RUN as commands could be arbitrary but with COPY / ADD it is easy to parse the destination and check which would explain the behavior.

nalind:
Updated to use the contents of a volume mount point to initialize a directory that we bind mount in from the host if nothing is specified for a volume location.  This appears to match the behavior of "docker run" and "docker start" for images that contain data in volume directories.

run.go
```diff
@@ -98,17 +100,37 @@ func setupMounts(spec *specs.Spec, optionMounts []specs.Mount, bindFiles, volume
...
+			srcPath := filepath.Join(mountPoint, volume)
+			if err = archive.CopyWithTar(srcPath, volumePath); err != nil {
+				return errors.Wrapf(err, "error populating directory %q for volume %q in container %q using contents of %q", volumePath, volume, b.ContainerID, srcPath)
+			}
+
+		}
+		// Add the bind mount.
 		mounts = append(mounts, specs.Mount{
-			Source:      "tmpfs",
+			Source:      volumePath,
```
nalind: Yes, it is cleaned up by the storage library.

rhatdan:
@rh-atomic-bot r+ ed4cd48

Issue #133, opened 2017-06-05
buildah rmi tagged images also remove the original image


gouyang:
After `buildah tag`, the tagged image ID is same with the original image ID, so remove the tagged image also remove the original image, which is not reasonable.

nalind:
Can you elaborate on this?  I'm not sure exactly what you're doing that is triggering the behavior you're seeing.

gouyang:
In below testing, remove `busybox2`, `busybox` got removed also. 

```
[root@fedora buildah]# ./buildah  images
IMAGE ID     IMAGE NAME
b812babf7f65 docker.io/library/redis:latest
b812babf7f65 docker.io/gouyang/redis:latest
b812babf7f65 docker.io/testing/redis:latest
b812babf7f65 docker.io/foo/redis:latest
5965adf86a6e docker.io/library/nginx:latest
223c6ba7409d docker.io/kubernetes/pause:latest
e236280e3ef7 docker.io/library/redis:alpine
a9a3ca121508 registry.access.redhat.com/rhel7/rhel:latest
8dc56e5fe235 docker.io/library/busybox:latest
8e1b8fedede7 docker.io/library/fedora:latest

[root@fedora buildah]# ./buildah tag docker.io/library/busybox docker.io/library/busybox2
           
[root@fedora buildah]# ./buildah rmi docker.io/library/busybox2
docker.io/library/busybox2

[root@fedora buildah]# ./buildah images
IMAGE ID     IMAGE NAME
b812babf7f65 docker.io/library/redis:latest
b812babf7f65 docker.io/gouyang/redis:latest
b812babf7f65 docker.io/testing/redis:latest
b812babf7f65 docker.io/foo/redis:latest
5965adf86a6e docker.io/library/nginx:latest
223c6ba7409d docker.io/kubernetes/pause:latest
e236280e3ef7 docker.io/library/redis:alpine
a9a3ca121508 registry.access.redhat.com/rhel7/rhel:latest
8e1b8fedede7 docker.io/library/fedora:latest
```

Docker just untagged the image.

```
[root@fedora buildah]# docker images
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
docker.io/redis     latest              e32ef7250bc1        6 weeks ago         183.6 MB
docker.io/busybox   latest              00f017a8c2a6        3 months ago        1.11 MB
docker.io/redis     alpine              83638a6d3af2        3 months ago        19.82 MB
[...]

PR #155, opened 2017-06-21
update 'buildah images' and 'buildah rmi' commands

14rcole:
add more flags to `buildah images` and `buildah rmi`, and write tests

### DIFF
--- a/cmd/buildah/common.go
+++ b/cmd/buildah/common.go
@@ -1,14 +1,27 @@
 package main
 
 import (
+	"encoding/json"
 	"os"
+	"strings"
+	"time"
 
 	is "github.com/containers/image/storage"
+	"github.com/containers/image/types"
 	"github.com/containers/storage"
 	"github.com/pkg/errors"
 	"github.com/projectatomic/buildah"
 	"github.com/urfave/cli"
 )
+
+type imageMetadata struct {
+	Tag            string              `json:"tag"`
+	CreatedTime    time.Time           `json:"created-time"`
+	ID             string              `json:"id"`
+	Blobs          []types.BlobInfo    `json:"blob-list"`
+	Layers         map[string][]string `json:"layers"`
+	SignatureSizes []string            `json:"signature-sizes"`
+}
 
 var needToShutdownStore = false
 
@@ -70,4 +83,32 @@ func openImage(store storage.Store, name string) (builder *buildah.Builder, err 
 		return nil, errors.Errorf("error mocking up build configuration")
 	}
 	return builder, nil
+}
+
+func parseMetadata(image storage.Image) (imageMetadata, error) {
+	var im imageMetadata
+
+	dec := json.NewDecoder(strings.NewReader(image.Metadata))
+	if err := dec.Decode(&im); err != nil {
+		return imageMetadata{}, err
+	}
+	return im, nil
+}
+
+func getSize(image storage.Image, store storage.Store) (int64, error) {
+
+	is.Transport.SetStore(store)
+	storeRef, err := is.Transport.ParseStoreReference(store, "@"+image.ID)
+	if err != nil {
+		return -1, err
+	}
+	img, err := storeRef.NewImage(nil)
+	if err != nil {
+		return -1, err
+	}
+	imgSize, err := img.Size()
+	if err != nil {
+		return -1, err
+	}
+	return imgSize, nil
 }
--- a/cmd/buildah/common_test.go
+++ b/cmd/buildah/common_test.go
@@ -1,0 +1,99 @@
+package main
+
+import (
+	"os/user"
+	"testing"
+
+	"flag"
+
+	is "github.com/containers/image/storage"
+	"github.com/containers/storage"
+	"github.com/urfave/cli"
+)
+
+func TestGetStore(t *testing.T) {
+	// Make sure the tests are running as root
+	failTestIfNotRoot(t)
+
+	set := flag.NewFlagSet("test", 0)
+	globalSet := flag.NewFlagSet("test", 0)
+	globalSet.String("root", "", "path to the root directory in which data, including images,  is stored")
+	globalCtx := cli.NewContext(nil, globalSet, nil)
+	command := cli.Command{Name: "imagesCommand"}
+	c := cli.NewContext(nil, set, globalCtx)
+	c.Command = command
+
+	_, err := getStore(c)
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+func TestParseMetadata(t *testing.T) {
+	// Make sure the tests are running as root
+	failTestIfNotRoot(t)
+
+	store, err := storage.GetStore(storage.DefaultStoreOptions)
+	if err != nil {
+		t.Fatal(err)
+	} else if store != nil {
+		is.Transport.SetStore(store)
+	}
+
+	images, err := store.Images()
+	if err != nil {
+		t.Fatalf("Error reading images: %v", err)
+	} else if len(images) == 0 {
+		t.Fatalf("no images with metadata to parse")
+	}
+
+	_, err = parseMetadata(images[0])
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+func TestGetSize(t *testing.T) {
+	// Make sure the tests are running as root
+	failTestIfNotRoot(t)
+
+	store, err := storage.GetStore(storage.DefaultStoreOptions)
+	if err != nil {
+		t.Fatal(err)
+	} else if store != nil {
+		is.Transport.SetStore(store)
+	}
+
+	images, err := store.Images()
+	if err != nil {
+		t.Fatalf("Error reading images: %v", err)
+	}
+
+	_, err = getSize(images[0], store)
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+func failTestIfNotRoot(t *testing.T) {
+	u, err := user.Current()
+	if err != nil {
+		t.Log("Could not determine user.  Running without root may cause tests to fail")
+	} else if u.Uid != "0" {
+		t.Fatal("tests will fail unless run as root")
+	}
+}
+
+func pullTestImage(imageName string) error {
+	set := flag.NewFlagSet("test", 0)
+	set.Bool("pull", true, "pull the image if not present")
+	globalSet := flag.NewFlagSet("globaltest", 0)
+	globalCtx := cli.NewContext(nil, globalSet, nil)
+	command := cli.Command{Name: "imagesCommand"}
+	c := cli.NewContext(nil, set, globalCtx)
+	c.Command = command
+	c.Set("pull", "true")
+	c.Args = append(c.Args, imageName)
+
+	return fromCommand(c)
+}
--- a/cmd/buildah/images.go
+++ b/cmd/buildah/images.go
@@ -2,8 +2,15 @@ package main
 
 import (
 	"fmt"
+	"os"
+	"strings"
+	"text/template"
+	"time"
 
 	"encoding/json"
+
+	is "github.com/containers/image/storage"
+	"github.com/containers/storage"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
 )
@@ -11,6 +18,24 @@ import (
 type jsonImage struct {
 	ID    string   `json:"id"`
 	Names []string `json:"names"`
+}
+
+type imageOutputParams struct {
+	ID        string
+	Name      string
+	Digest    string
+	CreatedAt string
+	Size      string
+}
+
+type filterParams struct {
+	dangling         string
+	label            string
+	beforeImage      string // Images are sorted by date, so we can just output until we see the image
+	sinceImage       string // Images are sorted by date, so we can just output until we don't see the image
+	beforeDate       time.Time
+	sinceDate        time.Time
+	referencePattern string
 }
 
 var (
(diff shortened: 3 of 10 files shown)


rhatdan:
bot, add author to whitelist

rhatdan:
bot, test pull request

rhatdan:
@14rcole Sign your image.  This is a patch for buildah not kpod, so change the title.

cmd/buildah/images.go
```diff
@@ -48,38 +83,293 @@ func imagesCmd(c *cli.Context) error {
...
+	if c.IsSet("format") {
+		formatString = c.String("format")
+		hasTemplate = true
+	}
+
+	name := ""
+	if len(c.Args()) == 1 {
+		name = c.Args().Get(0)
+	} else if len(c.Args()) > 1 {
+		return errors.New("'kpod images' requires at most 1 argument")
```
rhatdan: kpod?

cmd/buildah/rmi_test.go
```diff
@@ -0,0 +1,168 @@
...
+	"strings"
+	"testing"
+
+	is "github.com/containers/image/storage"
+	"github.com/containers/storage"
+)
+
+func TestProperImageRefTrue(t *testing.T) {
+	// Pull an image so we know we have it
+	cmd := exec.Command("crioctl", "image", "pull", "busybox:latest")
```
rhatdan: crioctl is not part of buildah.

cmd/buildah/rmi_test.go
```diff
@@ -0,0 +1,168 @@
...
+	if err != nil {
+		t.Errorf("could not match image: %v", err)
+	} else if imgRef == nil {
+		t.Error("Returned nil Image Reference")
+	}
+}
+
+func TestProperImageRefFalse(t *testing.T) {
+	// Pull an image so we know we have it
+	cmd := exec.Command("crioctl", "image", "pull", "busybox:latest")
```
rhatdan: You should probably use the from calls for this or internal calls to storage.

cmd/buildah/common_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	is "github.com/containers/image/storage"
+	"github.com/containers/storage"
+	"github.com/urfave/cli"
+)
+
+func TestGetStore(t *testing.T) {
+	u, err := user.Current()
+	if err != nil {
+		t.Log("Could not determine user.  Running as root may cause tests to fail")
```
TomSweeneyRedHat: Here you're saying you can't run as root, but at line 19 you're saying you must run at root in the else for this if?

cmd/buildah/common_test.go
```diff
@@ -0,0 +1,88 @@
...
+	_, err = getStore(c)
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+func TestParseMetadata(t *testing.T) {
+	u, err := user.Current()
+	if err != nil {
+		t.Log("Could not determine user.  Running as root may cause tests to fail")
```
TomSweeneyRedHat: same root/not root bit here.

cmd/buildah/rmi.go
```diff
@@ -43,55 +57,38 @@ func rmiCmd(c *cli.Context) error {
...
+					if err2 != nil {
+						logrus.Debugf("Error parsing image ID: %v\n", err2)
+					}
+					containers, err2 := store.Containers()
+					if err2 != nil {
+						logrus.Debugf("Error getting associated containers: %v\n", err)
+					}
+					for i := 0; i < len(containers); i++ {
+						if containers[i].ImageID == image.ID {
+							fmt.Printf("Found the container")
```
TomSweeneyRedHat: Not real helpful to have 1 to n "Found the container" 's.  If you think it's important to note, perhaps change this to something like fmt.Printf("Deleting container %s", containers[i].ID)

cmd/buildah/common_test.go
14rcole: It should all require root.  I'll pull all of those checks out into another function so that I don't have as much duplicate code running around

cmd/buildah/rmi.go
14rcole: Good catch.  That was actually a debugging output that I had forgotten to delete

rhatdan:
@14rcole tests are crashing.

14rcole:
My most recent addition to this PR should resolve #133

rhatdan:
@14rcole tests crashing.

cmd/buildah/common_test.go
```diff
@@ -0,0 +1,99 @@
...
+		t.Fatalf("Error reading images: %v", err)
+	}
+
+	_, err = getSize(images[0], store)
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+func failTestIfNotRoot(t *testing.T) {
```
TomSweeneyRedHat: This is a good, but I'm sure finger numbing change.  Thanks for making it.

rhatdan:
LGTM
@nalind PTAL

cmd/buildah/rmi.go
```diff
@@ -35,75 +47,192 @@ func rmiCmd(c *cli.Context) error {
...
+	}
+	for _, ctr := range containers {
+		if ctr.ImageID == image.ID {
+			ctrIDs = append(ctrIDs, ctr.ID)
+		}
+	}
+	return ctrIDs, nil
+}
+
+func removeRunningContainers(ctrIDs []string, store storage.Store) error {
```
TomSweeneyRedHat: Should this be named 'removeContianers'?  I'm not seeing any checks to see if the container is running within the function.

cmd/buildah/rmi.go
```diff
@@ -35,75 +47,192 @@ func rmiCmd(c *cli.Context) error {
...
+	var ref types.ImageReference
+	var err error
+	if ref, err = alltransports.ParseImageName(id); err == nil {
+		if img, err2 := ref.NewImage(nil); err2 == nil {
+			img.Close()
+			return ref, nil
+		}
+		return nil, fmt.Errorf("error confirming presence of image %q: %v", transports.ImageName(ref), err)
+	}
+	return nil, fmt.Errorf("error parsing %q as a store reference: %v", id, err)
```
TomSweeneyRedHat: should this be "image reference" rather than "store reference"?

cmd/buildah/rmi.go
```diff
@@ -35,75 +47,192 @@ func rmiCmd(c *cli.Context) error {
...
+	var ref types.ImageReference
+	var err error
+	if ref, err = is.Transport.ParseStoreReference(store, id); err == nil {
+		if img, err2 := ref.NewImage(nil); err2 == nil {
+			img.Close()
+			return ref, nil
+		}
+		return nil, fmt.Errorf("error confirming presence of image %q: %v", transports.ImageName(ref), err)
+	}
+	return nil, fmt.Errorf("error parsing %q as a store reference: %v", id, err)
```
TomSweeneyRedHat: I'd suggest "storage image reference" over "store reference"

cmd/buildah/rmi.go
```diff
@@ -35,75 +47,192 @@ func rmiCmd(c *cli.Context) error {
...
+// it and check if it corresponds to an image that actually exists.
+func storageImageRef(store storage.Store, id string) (types.ImageReference, error) {
+	var ref types.ImageReference
+	var err error
+	if ref, err = is.Transport.ParseStoreReference(store, id); err == nil {
+		if img, err2 := ref.NewImage(nil); err2 == nil {
+			img.Close()
+			return ref, nil
+		}
+		return nil, fmt.Errorf("error confirming presence of image %q: %v", transports.ImageName(ref), err)
```
TomSweeneyRedHat: change image to "storrage image reference" perhaps?

cmd/buildah/rmi.go
```diff
@@ -35,75 +47,192 @@ func rmiCmd(c *cli.Context) error {
...
+// can't hurt either.
+func storageImageID(store storage.Store, id string) (types.ImageReference, error) {
+	var ref types.ImageReference
+	var err error
+	if ref, err = is.Transport.ParseStoreReference(store, "@"+id); err == nil {
+		if img, err2 := ref.NewImage(nil); err2 == nil {
+			img.Close()
+			return ref, nil
+		}
+		return nil, fmt.Errorf("error confirming presence of image %q: %v", transports.ImageName(ref), err)
```
TomSweeneyRedHat: First error is 'image", second is "image reference".  I'd prefer "storage image reference" for both, but will settle for consistency.

cmd/buildah/rmi.go
14rcole: Yes, good catch

rhatdan:
@14rcole Please rebase.

14rcole:
Rebasing didn't update my PR.  Do I need to trigger something manually?

14rcole:
Never mind, it seems to have just been delayed

cmd/buildah/common.go
```diff
@@ -1,15 +1,28 @@
...
 )
 
+type imageMetadata struct {
+	Tag            string              `json:"tag"`
+	CreatedTime    time.Time           `json:"created-time"`
+	ID             string              `json:"id"`
+	Blobs          []types.BlobInfo    `json:"blob-list"`
+	Layers         map[string][]string `json:"layers"`
+	SignatureSizes []string            `json:"signature-sizes"`
+}
```
nalind: I wouldn't recommend parsing the metadata that the image library records - it intentionally doesn't make that a public type, so that it can change the layout to fit its needs.  In the future, it very likely will be missing the information we're trying to read, since it will be duplicating information that newer versions of the storage library will cache in publicly-defined structures.

cmd/buildah/images.go
```diff
@@ -13,6 +18,23 @@ type jsonImage struct {
...
+	Digest    string
+	CreatedAt string
+	Size      string
+}
+
+type filterParams struct {
+	dangling         string
+	label            string
+	beforeImage      string // Images are sorted by date, so we can just output until we see the image
+	sinceImage       string // Images are sorted by date, so we can just output until we don't see the image
```
nalind: The storage library doesn't take steps to ensure anything about the ordering of images that it returns, and it's probably a bug that the creation date that the image library metadata records for a pulled image is the time at which the image was pulled -- if the image's config blob includes a datestamp indicating when it was built on whichever system it was built, I think we'd want to use that value instead.

cmd/buildah/common.go
nalind: We can _probably_ get away with this for now, but when we pull in a newer storage library, we'll want to remember to switch to using the public information.

cmd/buildah/images.go
```diff
@@ -24,22 +46,35 @@ var (
...
+			Usage: "pretty-print images using a Go template. will override --quiet",
+		},
+		cli.StringFlag{
+			Name:  "filter, f",
+			Usage: "filter output based on conditions provided (default [])",
+		},
 	}
-	imagesDescription = "Lists locally stored images."
+
+	imagesDescription = "lists locally stored images."
```
nalind: This makes the capitalization stop matching the other commands.  What's the intent here?

cmd/buildah/images.go
```diff
@@ -24,22 +46,35 @@ var (
...
 	imagesCommand     = cli.Command{
 		Name:        "images",
-		Usage:       "List images in local storage",
+		Usage:       "list images in local storage",
```
nalind: This makes the capitalization stop matching the rest of the `buildah --help` output.  What's the intent here?

cmd/buildah/images.go
```diff
@@ -24,22 +46,35 @@ var (
...
 		Description: imagesDescription,
 		Flags:       imagesFlags,
 		Action:      imagesCmd,
-		ArgsUsage:   " ",
+		ArgsUsage:   "",
```
nalind: This causes the help output to include `[arguments...]` after `[command options]` in the usage section.  It's preferable to indicate what type of value we expect to be given, as we do for the other commands.

cmd/buildah/images.go
```diff
@@ -24,22 +46,35 @@ var (
 			Usage: "do not print column headings",
 		},
 		cli.BoolFlag{
-			Name:  "notruncate",
+			Name:  "no-trunc, notruncate",
```
nalind: What's the rationale for adding this alias for this option to this command, but not to any of the other commands that offer a `--notruncate` option?

cmd/buildah/common.go
14rcole: What would you suggest as an alternative?  Something like unmarshaling the data to a `map[string]interface{}`?

cmd/buildah/images.go
14rcole: These were all unintentional changes, I'll revert them

cmd/buildah/images.go
14rcole: I hadn't looked at any of the other commands.  When I wrote this code for kpod, it was supposed to match the docker cli, which uses `--no-trunc`.  I chose not to remove the option when I ported the code to buildah.  I can remove it or add the option to the other commands if you would like

cmd/buildah/images.go
14rcole: Okay, I'll fix this

cmd/buildah/images.go
nalind: @rhatdan You've got stronger opinions on the CLI than I do.  WDYT?

cmd/buildah/common.go
nalind: We're not providing anything better (yet), so this is probably going to have to do until we update the storage and image libraries to provide this information as part of their public APIs.

cmd/buildah/images.go
rhatdan: I think we should match the CLI of kpod and docker.   No value in the alternate spelling.

14rcole: Should I change this in the other commands that offer a `--notruncate` option then?  Or should that be a separate PR?

14rcole:
@rhatdan is there a way to manually re-trigger the checks?  It looks like an internal error occurred on the Red Hat CI

rhatdan:
@14rcole I usually just do a 
```
git commit -a --amend -s
git push rhatdan --force
```

cmd/buildah/images.go
rhatdan: Lets open separate pull requests.

rhatdan:
@nalind you ok with merging?

rhatdan:
@14rcole you need to update buildah.md and README.md with these new commands.

14rcole:
@rhatdan The commands were already listed in buildah.md and README.md, and I updated both buildah-images.md and buildah-rmi.md to include the flags I added.  Is there something else in those files that needs to be changed?

rhatdan:
Ok Sorry I thought they were missing.

rhatdan:
I am going to merge...
@rh-atomic-bot r+ c7c207b

TomSweeneyRedHat:
Also fixes #232

PR #157, opened 2017-06-22
Add 'buildah version' command

umohnani8:
Added "buildah version" command following the same format as "docker version" and "kpod version"

### DIFF
--- a/cmd/buildah/version.go
+++ b/cmd/buildah/version.go
@@ -1,0 +1,48 @@
+package main
+
+import (
+	"fmt"
+	"runtime"
+	"strconv"
+	"time"
+
+	ispecs "github.com/opencontainers/image-spec/specs-go"
+	rspecs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/projectatomic/buildah"
+	"github.com/urfave/cli"
+)
+
+//Overwritten at build time
+var (
+	gitCommit string
+	buildInfo string
+)
+
+//Function to get and print info for version command
+func versionCmd(c *cli.Context) error {
+
+	//converting unix time from string to int64
+	buildTime, err := strconv.ParseInt(buildInfo, 10, 64)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Version:      ", buildah.Version)
+	fmt.Println("Go Version:   ", runtime.Version())
+	fmt.Println("Image Spec:   ", ispecs.Version)
+	fmt.Println("Runtime Spec: ", rspecs.Version)
+	fmt.Println("Git Commit:   ", gitCommit)
+
+	//Prints out the build time in readable format
+	fmt.Println("Built:        ", time.Unix(buildTime, 0).Format(time.ANSIC))
+	fmt.Println("OS/Arch:      ", runtime.GOOS+"/"+runtime.GOARCH)
+
+	return nil
+}
+
+//cli command to print out the version info of buildah
+var versionCommand = cli.Command{
+	Name:   "version",
+	Usage:  "Display the Buildah Version Information",
+	Action: versionCmd,
+}
--- a/docs/buildah-version.md
+++ b/docs/buildah-version.md
@@ -1,0 +1,27 @@
+## buildah-version "1" "June 2017" "buildah"
+
+## NAME
+buildah version - Display the Buildah Version Information.
+
+## SYNOPSIS
+**buildah version**
+[**--help**|**-h**]
+
+## DESCRIPTION
+Shows the following information: Version, Go Version, Image Spec, Runtime Spec, Git Commit, Build Time, OS, and Architecture.
+
+## OPTIONS
+
+**--help, -h**
+  Print usage statement
+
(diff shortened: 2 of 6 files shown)


rhatdan:
@umohnani8 Sign your commit.

rhatdan:
@umohnani8 Not compiling.

umohnani8:
I fixed the issue and signed the commit.

TomSweeneyRedHat:
One of the files is a binary buildah file.  I don't think we want that in this commit.

TomSweeneyRedHat:
OH FUDGE!  I didn't meant to close this, just comment on it.

umohnani8:
oh right, buildah appeared while I was testing. Should I just delete the file or is there another way to do it?

TomSweeneyRedHat:
Hopefully
cd to your sandbox
git rm buildah
git commit --amend -s

Then the normal fetch and push stuff as appropriate.

rhatdan:
LGTM

rhatdan:

cmd/buildah/version.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+//cli command to print out the full version of kpod
+var versionCommand = cli.Command{
+	Name:  "version",
+	Usage: "Display the Buildah Version Information",
+	Action: func(context *cli.Context) error {
+		versionCmd()
+		return nil
+	},
```
nalind: No need for this wrapper if `versionCmd()` is changed to return an error -- either one received from `strconv.ParseInt()` or `nil`, if `versionCmd()` runs without problems.

docs/buildah-version.md
```diff
@@ -0,0 +1,27 @@
...
+
+## NAME
+buildah version - Display the Buildah Version Information.
+
+## SYNOPSIS
+**buildah version**
+[**--help**|**-h**]
+
+## DESCRIPTION
+Shows the following information for the Client: Version, Go Version, Image Spec, Runtime Spec, Git Commit, Build Time, OS, and Architecture.
```
nalind: We're not a client-server program, so referring to it as a client in the description is confusing.

cmd/buildah/version.go
```diff
@@ -0,0 +1,49 @@
...
+
+	fmt.Println("Version:	", buildah.Version)
+	fmt.Println("Go Version:	", runtime.Version())
+	fmt.Println("Image Spec:	", ispecs.Version)
+	fmt.Println("Runtime Spec:	", rspecs.Version)
+	fmt.Println("Git Commit:	", gitCommit)
+
+	//Prints out the build time in readable format
+	fmt.Println("Built:		", time.Unix(buildTime, 0).Format(time.ANSIC))
+	fmt.Println("OS/Arch:	", runtime.GOOS+"/"+runtime.GOARCH)
```
nalind: Using tabs for indentation in the printed value makes the alignment harder to check, for example in the github web UI.  Can we change these to spaces?

cmd/buildah/version.go
umohnani8: Used versionCmd() to return the error instead and took out the wrapper. Fixed!

docs/buildah-version.md
umohnani8: Oops - I fixed it in the code, but forgot to in the docs. Fixed!

cmd/buildah/version.go
umohnani8: Used spaces instead of tabs. Fixed!

rhatdan:
@umohnani8 Merge your pull requests together.

cmd/buildah/version.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	fmt.Println("Version:      ", buildah.Version)
+	fmt.Println("Go Version:   ", runtime.Version())
+	fmt.Println("Image Spec:   ", ispecs.Version)
+	fmt.Println("Runtime Spec: ", rspecs.Version)
+	fmt.Println("Git Commit:   ", gitCommit)
+
+	//Prints out the build time in readable format
+	fmt.Println("Built:	      ", time.Unix(buildTime, 0).Format(time.ANSIC))
```
TomSweeneyRedHat: Should the right quote on this line be in the same position as the ones above/below it?  Is everything using tabs or spaces after the ":" and does it line up nicely when run?

umohnani8: So if the right quote is lined up with the rest, the output doesn't line up and is shifted to the right by 1 space. I placed the right quote one character before the other right quotes - unsure why its showing more on the right. I will fix that, but everything lines up nicely when run.

umohnani8: Ok, I pushed it again but git is still showing it one space to the right even though its not. It definitely lines up nicely on my machine - unsure why git is showing it like that though.

umohnani8: Oh right! I missed that. Fixed it so that they are all aligned now.

rhatdan:
@rh-atomic-bot r+  	27823fe

PR #160, opened 2017-06-23
Handle "run" without an explicit command correctly

nalind:
When "run" isn't explicitly given a command, mix the command and entrypoint options and configured values together correctly.


rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+ 7be7aa9

PR #161, opened 2017-06-26
cmd/buildah/images.go: Add JSON output option

baude:
The Atomic CLI will eventually need to be able to consume
structured output (in something like JSON).  This commit
adds a -j option to output to trigger JSON output of
images.

Example output:
```
[
    {
        "id": "aa66247d48aedfa3e9b74e4a41d2c9e5d2529122c8f0d43417012028a66f4f3b",
        "names": [
            "docker.io/library/busybox:latest"
        ]
    },
    {
        "id": "26db5ad6e82d85265d1609e6bffc04331537fdceb9740d36f576e7ee4e8d1be3",
        "names": [
            "docker.io/library/alpine:latest"
        ]
    }
]
```

### DIFF
--- a/cmd/buildah/images.go
+++ b/cmd/buildah/images.go
@@ -3,9 +3,15 @@ package main
 import (
 	"fmt"
 
+	"encoding/json"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
 )
+
+type jsonImage struct {
+	ID    string   `json:"id"`
+	Names []string `json:"names"`
+}
 
 var (
 	imagesFlags = []cli.Flag{
@@ -20,6 +26,10 @@ var (
 		cli.BoolFlag{
 			Name:  "notruncate",
 			Usage: "do not truncate output",
+		},
+		cli.BoolFlag{
+			Name:  "json",
+			Usage: "output in JSON format",
 		},
 	}
 	imagesDescription = "Lists locally stored images."
@@ -39,6 +49,11 @@ func imagesCmd(c *cli.Context) error {
 		return err
 	}
 
+	images, err := store.Images()
+	if err != nil {
+		return errors.Wrapf(err, "error reading images")
+	}
+
 	quiet := false
 	if c.IsSet("quiet") {
 		quiet = c.Bool("quiet")
@@ -51,11 +66,18 @@ func imagesCmd(c *cli.Context) error {
 	if c.IsSet("notruncate") {
 		truncate = !c.Bool("notruncate")
 	}
-	images, err := store.Images()
-	if err != nil {
-		return errors.Wrapf(err, "error reading images")
+	if c.IsSet("json") {
+		JSONImages := []jsonImage{}
+		for _, image := range images {
+			JSONImages = append(JSONImages, jsonImage{ID: image.ID, Names: image.Names})
+		}
+		data, err := json.MarshalIndent(JSONImages, "", "    ")
+		if err != nil {
+			return err
+		}
+		fmt.Printf("%s\n", data)
+		return nil
 	}
-
 	if len(images) > 0 && !noheading && !quiet {
 		if truncate {
 			fmt.Printf("%-12s %s\n", "IMAGE ID", "IMAGE NAME")
--- a/contrib/completions/bash/buildah
+++ b/contrib/completions/bash/buildah
@@ -554,6 +554,7 @@ return 1
      local boolean_options="
      --help
      -h
+     --json
      --quiet
      -q
      --noheading
--- a/docs/buildah-images.md
+++ b/docs/buildah-images.md
@@ -11,6 +11,9 @@ Displays locally stored images, their names, and their IDs.
 
 ## OPTIONS
 
+**--json**
+Display the output in JSON format.
+
 **--noheading, -n**
 
 Omit the table headings from the listing of images.
@@ -26,6 +29,8 @@ Lists only the image IDs.
 ## EXAMPLE
 
 buildah images
+
+buildah images --json
 
 buildah images --quiet
 


cmd/buildah/images.go
```diff
@@ -5,8 +5,14 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
+	"encoding/json"
 )
 
+type JsonImage struct {
+	Id string `json:"id"`
```
runcom: `Id` should be `ID` (all uppercase) - you would also want an `omitempty` in case id isn't there? (`json:"id,omitempty"`)

cmd/buildah/images.go
```diff
@@ -5,8 +5,14 @@ import (
...
+	Names []string `json:"names"`
```
runcom: omitempty?

cmd/buildah/images.go
```diff
@@ -21,6 +27,10 @@ var (
 			Name:  "notruncate",
 			Usage: "do not truncate output",
 		},
+		cli.BoolFlag{
+			Name: "json, j",
```
runcom: I'd drop the short `-j`, just `--json` is fine

cmd/buildah/images.go
```diff
@@ -51,11 +61,26 @@ func imagesCmd(c *cli.Context) error {
 	if c.IsSet("notruncate") {
 		truncate = !c.Bool("notruncate")
 	}
+	json_out := false
```
runcom: no sneak_case, camelCase needed by go

cmd/buildah/images.go
```diff
@@ -51,11 +61,26 @@ func imagesCmd(c *cli.Context) error {
...
+	if c.IsSet("json") {
```
runcom: you can remove this `if` - when you do `c.Bool`, if the flag isn't set it will still return `false` so no need for this.

cmd/buildah/images.go
```diff
@@ -51,11 +61,26 @@ func imagesCmd(c *cli.Context) error {
...
+		json_out = c.Bool("json")
+	}
 	images, err := store.Images()
 	if err != nil {
 		return errors.Wrapf(err, "error reading images")
 	}
-
+	if json_out {
+		var JsonImages = []JsonImage{}
```
runcom: `JSONImages := []JSONImage{}` (no var)

runcom: oh, also, this must be lowercase, no uppercase for vars

cmd/buildah/images.go
```diff
@@ -5,8 +5,14 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
+	"encoding/json"
 )
 
+type JsonImage struct {
```
runcom: this isn't exported, it must start with a lowercase

cmd/buildah/images.go
```diff
@@ -21,6 +27,10 @@ var (
 			Name:  "notruncate",
 			Usage: "do not truncate output",
 		},
+		cli.BoolFlag{
+			Name:  "json, j",
```
runcom: drop `j`

cmd/buildah/images.go
```diff
@@ -3,10 +3,16 @@ package main
 import (
 	"fmt"
 
+	"encoding/json"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
 )
 
+type JsonImage struct {
```
runcom: start it with lowercase, it's unexported

cmd/buildah/images.go
```diff
@@ -3,10 +3,16 @@ package main
...
+	Id    string   `json:"id"`
```
runcom: Id should be ID (all uppercase)

cmd/buildah/images.go
baude: For the consumer, I think best to allow the key to remain but have the value be ""  ... then the consumer wont get key errors with simple iteration.  Will fix the Id casing.

cmd/buildah/images.go
baude: same as above.

cmd/buildah/images.go
baude: Done

rhatdan:
Missing Bash completions.

cmd/buildah/images.go
baude: Fixed.

rhatdan:
@baude you need to sign your pull request.

cmd/buildah/images.go
baude: Leaving the same to be consistent with the other conditionals.

cmd/buildah/images.go
baude: Done

cmd/buildah/images.go
baude: done

cmd/buildah/images.go
baude: done

cmd/buildah/images.go
baude: fixed

baude:
Added bash completions

cmd/buildah/images.go
```diff
@@ -51,11 +61,26 @@ func imagesCmd(c *cli.Context) error {
...
+	JsonOut := false
+	if c.IsSet("json") {
+		JsonOut = c.Bool("json")
+	}
 	images, err := store.Images()
 	if err != nil {
 		return errors.Wrapf(err, "error reading images")
 	}
-
+	if JsonOut {
```
rhatdan: I agree with @runcom use this here.
```
if c.IsSet("json") {
```

And eliminate the code above.  If we are using this syntax elsewhere we should clean those up.

rhatdan:
@baude Still not signed.

git commit -a --amend -s

rhatdan:
@rh-atomic-bot r+ b85c234

PR #162, opened 2017-06-26
Better support for testing with different storage drivers

nalind:
This change updates the test helper macros to pass the right flags to tools to use the storage driver named in `$STORAGE_DRIVER` instead of `vfs`, if the variable is set when the tests are run.  It adds debug logging support to the `imgtype` test helper, and ensures that it actually shuts down storage, even when it exits with errors.

### DIFF
--- a/tests/imgtype.go
+++ b/tests/imgtype.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/Sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

rhatdan:
@TomSweeneyRedHat PTAL

tests/imgtype.go
```diff
@@ -79,60 +97,74 @@ func main() {
 
 		ref, err := is.Transport.ParseStoreReference(store, image)
 		if err != nil {
-			logrus.Fatalf("error parsing reference %q: %v", image, err)
-		}
-
-		src, err := ref.NewImageSource(systemContext, []string{expectedManifestType})
```
TomSweeneyRedHat: Is this test/check no longer needed?  I don't see it in the new version.

nalind: The code for handling an error result from `ParseStoreReference`?  The diff is assuming one `if err != nil` is equivalent to another, but we should still be handling that case in the proposed version.

TomSweeneyRedHat: I wasn't referring to the Transport.ParseStoreReference(), but the call to ref.NewImageSource on line 85 in the old code.  I didn't see that elsewhere.

nalind: Ah, yes.  Realized that we can grab the manifest contents from the Image object, so there's no need to create an ImageSource object to read it when we already have an Image object.

rhatdan:
@TomSweeneyRedHat Can we merge?

rhatdan:
@rh-atomic-bot r+ ccfde8a

PR #163, opened 2017-06-26
Update the example commit target to skip transport

nalind:
Update the target name that we use when committing an image in the example script to not mention the local storage transport, since the default is to use it anyway, and if we ever rename it, it's one fewer edit to make.


rhatdan:
@rh-atomic-bot r+  	89471f9

PR #170, opened 2017-06-27
Add buildah export support

rhatdan:
Will export the contents of a container as a tar ball.

### DIFF
--- a/cmd/buildah/export.go
+++ b/cmd/buildah/export.go
@@ -1,0 +1,87 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/containers/storage/pkg/archive"
+	"github.com/pkg/errors"
+	"github.com/projectatomic/buildah"
+	"github.com/urfave/cli"
+)
+
+var (
+	exportFlags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "output, o",
+			Usage: "write to a file, instead of STDOUT",
+		},
+	}
+	exportCommand = cli.Command{
+		Name:  "export",
+		Usage: "Export container's filesystem contents as a tar archive",
+		Description: `This command exports the full or shortened container ID or container name to
+   STDOUT and should be redirected to a tar file.
+       `,
+		Flags:     exportFlags,
+		Action:    exportCmd,
+		ArgsUsage: "CONTAINER",
+	}
+)
+
+func exportCmd(c *cli.Context) error {
+	var builder *buildah.Builder
+
+	args := c.Args()
+	if len(args) == 0 {
+		return errors.Errorf("container name must be specified")
+	}
+	if len(args) > 1 {
+		return errors.Errorf("too many arguments specified")
+	}
+
+	name := args[0]
+
+	store, err := getStore(c)
+	if err != nil {
+		return err
+	}
+
+	builder, err = openBuilder(store, name)
+	if err != nil {
+		return errors.Wrapf(err, "error reading build container %q", name)
+	}
+
+	mountPoint, err := builder.Mount("")
+	if err != nil {
+		return errors.Wrapf(err, "error mounting %q container %q", name, builder.Container)
+	}
+	defer func() {
+		if err := builder.Unmount(); err != nil {
+			fmt.Printf("Failed to umount %q: %v\n", builder.Container, err)
+		}
+	}()
+
+	input, err := archive.Tar(mountPoint, 0)
+	if err != nil {
+		return errors.Wrapf(err, "error reading directory %q", name)
+	}
+
+	outFile := os.Stdout
+	if c.IsSet("output") {
+		outfile := c.String("output")
+		outFile, err = os.Create(outfile)
+		if err != nil {
+			return errors.Wrapf(err, "error creating file %q", outfile)
+		}
+		defer outFile.Close()
+	}
+	if logrus.IsTerminal(outFile) {
+		return errors.Errorf("Refusing to save to a terminal. Use the -o flag or redirect.")
+	}
+
+	_, err = io.Copy(outFile, input)
+	return err
+}
--- a/docs/buildah.md
+++ b/docs/buildah.md
@@ -58,6 +58,7 @@ Print the version
 | buildah-config(1)     | Update image configuration settings.                                                                 |
 | buildah-containers(1) | List the working containers and their base images.                                                   |
 | buildah-copy(1)       | Copies the contents of a file, URL, or directory into a container's working directory.               |
+| buildah-export(1)     | Export the contents of a container's filesystem as a tar archive                                     |
 | buildah-from(1)       | Creates a new working container, either from scratch or using a specified image as a starting point. |
 | buildah-images(1)     | List images in local storage.                                                                        |
 | buildah-inspect(1)    | Inspects the configuration of a container or image                                                   |
--- a/tests/export.bats
+++ b/tests/export.bats
@@ -1,0 +1,22 @@
+#!/usr/bin/env bats
+
+load helpers
+
+@test "extract" {
+  touch ${TESTDIR}/reference-time-file
+  for source in scratch alpine; do
+    cid=$(buildah from --pull=true --signature-policy ${TESTSDIR}/policy.json ${source})
+    mnt=$(buildah mount $cid)
+    touch ${mnt}/export.file
+    tar -cf - --transform s,^./,,g -C ${mnt} . | tar tf - | grep -v "^./$" | sort > ${TESTDIR}/tar.output
+    buildah umount $cid
+    buildah export "$cid" > ${TESTDIR}/${source}.tar
+    buildah export -o ${TESTDIR}/${source}1.tar "$cid"
+    diff ${TESTDIR}/${source}.tar ${TESTDIR}/${source}1.tar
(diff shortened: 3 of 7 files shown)


rhatdan:
Need to add tests, but other then that this seems to work.

docs/buildah.md
```diff
@@ -60,6 +60,7 @@ Print the version
 | buildah-copy(1)       | Copies the contents of a file, URL, or directory into a container's working directory.               |
 | buildah-from(1)       | Creates a new working container, either from scratch or using a specified image as a starting point. |
 | buildah-images(1)     | List images in local storage.                                                                        |
+| buildah-export(1)     | Export the contents of a container's filesystem as a tar archive                                     |
```
nalind: Seems to be in the wrong place?

docs/buildah-export.1.md
```diff
@@ -0,0 +1,43 @@
...
+  Print usage statement
+
+**-o**, **--output**=""
+  Write to a file, instead of STDOUT
+
+# EXAMPLES
+Export the contents of the container called angry_bell to a tar file
+called angry_bell.tar:
+
+    # buildah export angry_bell > angry_bell.tar
```
TomSweeneyRedHat: angry_bell should == pahked_cah

docs/buildah.md
```diff
@@ -58,6 +58,7 @@ Print the version
 | buildah-config(1)     | Update image configuration settings.                                                                 |
 | buildah-containers(1) | List the working containers and their base images.                                                   |
 | buildah-copy(1)       | Copies the contents of a file, URL, or directory into a container's working directory.               |
+| buildah-export(1)     | Export the contents of a container's filesystem as a tar archive                                     |
```
TomSweeneyRedHat: I think you don't have Tomasek's change here that added the links.

TomSweeneyRedHat:
I think you missed a merge on buildah.md.  The code changes look ok to me, don't know what's what with the test though.

docs/buildah.md
rhatdan: The links are in the README.md, Not in the man page, but I will add this there also.  Good point

TomSweeneyRedHat: Doh!  Yep, you're right I'd the wrong file.  It would be a nice add here too though.

rhatdan:
Added tests, I think this is ready for merge/review.

@nalind @TomSweeneyRedHat @14rcole @umohnani8 PTAL

cmd/buildah/export.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+func exportCmd(c *cli.Context) error {
+	var builder *buildah.Builder
+
+	args := c.Args()
+	if len(args) == 0 {
+		return errors.Errorf("container name must be specified")
+	}
+	if len(args) > 1 {
```
TomSweeneyRedHat: nit, else instead of if here?

tests/export.bats
```diff
@@ -0,0 +1,19 @@
...
+
+@test "extract" {
+  touch ${TESTDIR}/reference-time-file
+  for source in scratch alpine; do
+    cid=$(buildah from --pull=true --signature-policy ${TESTSDIR}/policy.json ${source})
+    mnt=$(buildah mount $cid)
+    touch ${mnt}/export.file
+    buildah umount $cid
+    buildah export "$cid" > ${TESTDIR}/${source}.tar
+    buildah export "$cid" > ${TESTDIR}/${source}.tar
```
TomSweeneyRedHat: It's fine, but did you mean to do the same export twice?  Are you just testing the overwrite of the tar?

TomSweeneyRedHat:
Assuming happy tests, LGTM for merging.

tests/export.bats
rhatdan: Nope this is a bug.

cmd/buildah/export.go
```diff
@@ -0,0 +1,89 @@
...
+			Name:  "output, o",
+			Usage: "write to a file, instead of STDOUT",
+		},
+	}
+	exportCommand = cli.Command{
+		Name:  "export",
+		Usage: "Export the contents of a container's filesystem as a tar archive",
+		Description: `Export the contents of a container's filesystem using the full or shortened container ID or container name. The output is exported to STDOUT and can be redirected to a tar file.
+
+   Stream to a file instead of STDOUT by using -o.
```
nalind: Unintended paste?

cmd/buildah/export.go
```diff
@@ -0,0 +1,89 @@
...
+	defer func() {
+		if err := builder.Unmount(); err != nil {
+			fmt.Println("Failed to umount %q: %v", builder.Container, err)
+		}
+	}()
+
+	input, err := archive.Tar(mountPoint, 0)
+	if err != nil {
+		return errors.Wrapf(err, "error reading directory %q", name)
+	}
```
nalind: input is a ReadCloser, so it needs to be closed.

cmd/buildah/export.go
rhatdan: No this was intended, it gives a better description in buildah export --help.

cmd/buildah/export.go
```diff
@@ -0,0 +1,89 @@
...
+	outFile, err := os.Create(outfile)
+	if err != nil {
+		return errors.Wrapf(err, "error creating file %q", outfile)
+	}
```
nalind: I think it'd be better to reuse `os.Stdout` if `--output` isn't used, as is done above.

cmd/buildah/export.go
```diff
@@ -0,0 +1,89 @@
...
+
+	outfile := "/dev/stdout"
+	if c.IsSet("output") {
+		outfile = c.String("output")
+	} else {
+		stat, _ := os.Stdout.Stat()
+		if (stat.Mode() & os.ModeCharDevice) != 0 {
+			return errors.Errorf("Refusing to save to a terminal. Use the -o flag or redirect.")
+		}
```
nalind: Is this test also intended to cover cases where `--output` is given a device name?

nalind:
We're probably going to overlap some with kpod here, but should we be able to export non-buildah containers as well?

cmd/buildah/export.go
rhatdan: Fixed

cmd/buildah/export.go
rhatdan: Well it is intended to handle the case where you write to the tty.  Not sure if there is a better way to see if os.Stdout == tty.

cmd/buildah/export.go
nalind: The options section immediately follows the description in the help output, though, and this would be the only description that duplicates information about any options.

rhatdan:
Well as soon as this is approved we want kpod export, so they should be the same.

cmd/buildah/export.go
nalind: We use github.com/Sirupsen/logrus.IsTerminal() elsewhere, so it might as well get used here, too.

tests/export.bats
```diff
@@ -0,0 +1,18 @@
...
+
+@test "extract" {
+  touch ${TESTDIR}/reference-time-file
+  for source in scratch alpine; do
+    cid=$(buildah from --pull=true --signature-policy ${TESTSDIR}/policy.json ${source})
+    mnt=$(buildah mount $cid)
+    touch ${mnt}/export.file
+    buildah umount $cid
+    buildah export "$cid" > ${TESTDIR}/${source}.tar
+    buildah export -o ${TESTDIR}/${source}1.tar "$cid"
```
nalind: Can we add a `run tar tf` on one of these and check that the output matches the expected contents of the tarball?

cmd/buildah/export.go
rhatdan: OK I can remove the first line.

rhatdan: This is what it looks like currently
```
uildah
FC26:buildah: ./buildah export --help
NAME:
   buildah export - Export the contents of a container's filesystem as a tar archive

USAGE:
   buildah export [command options] CONTAINER

DESCRIPTION:
   Export the contents of a container's filesystem using the full or shortened container ID or container name. The output is exported to STDOUT and can be redirected to a tar file.

   Stream to a file instead of STDOUT by using -o.
       

OPTIONS:
   --output value, -o value  write to a file, instead of STDOUT
```

rhatdan: Cleaned it up to fix better in an 80 Char tty.

```
 ./buildah export --help
NAME:
   buildah export - Export container's filesystem contents as a tar archive

USAGE:
   buildah export [command options] CONTAINER

DESCRIPTION:
   Export the contents of a container's filesystem using the full or shortened 
   container ID or container name. The output is exported to STDOUT and can be 
   redirected to a tar file.

   Stream to a file instead of STDOUT by using -o.
       

OPTIONS:
   --output value, -o value  write to a file, instead of STDOUT
```

tests/export.bats
rhatdan: Working on it

nalind:
I think ideally we'd handle the `--output` argument in one place rather than two -- the terminal check can currently be bypassed by passing `/dev/tty` as a value, which is probably not the intention.
I still don't like duplicating the description of the only implemented option in the description section of the `--help` output and the man page, where they're immediately followed by the section that describes the available options, but if you've got a good reason for that, I'll let that go.

nalind:
Okay, LGTM pending CI.

rhatdan:
@rh-atomic-bot r+ 424592b

PR #174, opened 2017-06-28
Bump and pin containers/storage and containers/image

nalind:
This bumps and pins the versions of containers/storage and containers/image that we use, pulls in the newer image-spec that the newer containers/image needs, and adds a copy of ostree-go that containers/image also needs.  It then updates buildah's sources to handle changes in the updated versions.


rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+ 4e9f5db

PR #181, opened 2017-07-07
add a way to disable PTY allocation

### LINKED ISSUE #179: add a way to disable PTY allocation
One of the commands Ansible invokes on target system is `/bin/sh -c '/usr/bin/python && sleep 0'` which is used to determine whether `python` executable is present on the target. Unfortunately, if the pty is allocated, the REPL is invoked which causes the process to fail.

Would it be possible to add an option to disable PTY allocation? I am okay creating the patch myself: unfortunately I haven't figured out the way how it's possible to change the runtime spec.

TomasTomecek:
Fixes #179

This patch resolves the original issue: https://github.com/ansible/ansible/pull/26170#issuecomment-312591504

Signed-off-by: Tomas Tomecek <ttomecek@redhat.com>

### DIFF
--- a/docs/buildah-run.md
+++ b/docs/buildah-run.md
@@ -14,6 +14,13 @@ the *buildah config* command.
 
 ## OPTIONS
 
+**--tty**
+
+By default a pseudo-TTY is allocated only when buildah's standard input is
+attached to a pseudo-TTY. Setting the `--tty` option to `true` will cause a
+pseudo-TTY to be allocated inside the container. Setting the `--tty` option to
+`false` will prevent the pseudo-TTY from being allocated.
+
 **--runtime** *path*
 
 The *path* to an alternate OCI-compatible runtime.
--- a/cmd/buildah/run.go
+++ b/cmd/buildah/run.go
@@ -28,6 +28,10 @@ var (
 			Name:  "volume, v",
 			Usage: "bind mount a host location into the container while running the command",
 		},
+		cli.BoolFlag{
+			Name:  "tty",
+			Usage: "allocate a pseudo-TTY in the container",
+		},
 	}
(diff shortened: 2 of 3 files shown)


docs/buildah-run.md
```diff
@@ -14,6 +14,12 @@ the *buildah config* command.
 
 ## OPTIONS
 
+**--tty**
```
rhatdan: How about 
```
**--tty**

By default a pseudo-TTY is allocated only when buildah's standard output is
attached to a pseudo-TTY.  Setting the `--tty` option to `true` will cause a
pseudo-TTY to be allocated inside the container. Setting the `--tty` option to `false` will prevent the pseudo-TTY from being allocated.
```

rhatdan:
Do we need the same feature for buildah bud?


We also need bash completion.

docs/buildah-run.md
nalind: ... standard output is attached to a terminal.  Though we might want to change it to point to check standard input instead.

TomasTomecek: I like your suggestions.

TomasTomecek:
> Do we need the same feature for buildah bud?

`docker build` doesn't have such option. I suspect it may not be needed. The only problem may be if some tools try to be interactive (that's what I run into with `run`). Not sure if PTY is allocated by default.

I think I addressed all the comments.

rhatdan:
LGTM

rhatdan:
@nalind Do you want to merge?

TomSweeneyRedHat:
LGTM

nalind:
@rhatdan Yup, looks good to me.  Merging.

nalind:
@rh-atomic-bot r+ 62b79edc26c95ac85dfe898af91ad8145ed90377

PR #189, opened 2017-07-14
Add flag to remove containers on commit

rhatdan:
I think this would be good practice to eliminate wasted disk space.

### DIFF
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -36,6 +36,10 @@ var (
 			Name:  "quiet, q",
 			Usage: "don't output progress information when writing images",
 		},
+		cli.BoolFlag{
+			Name:  "rm",
+			Usage: "remove the container and its content after committing it to an image. Default leaves the container and its content in place.",
+		},
 	}
 	commitDescription = "Writes a new image using the container's read-write layer and, if it is based\n   on an image, the layers of that image"
 	commitCommand     = cli.Command{
@@ -128,5 +132,8 @@ func commitCmd(c *cli.Context) error {
 		return errors.Wrapf(err, "error committing container %q to %q", builder.Container, image)
 	}
 
+	if c.Bool("rm") {
+		return builder.Delete()
+	}
 	return nil
 }
--- a/docs/buildah-commit.md
+++ b/docs/buildah-commit.md
@@ -33,11 +33,15 @@ Control the format for the image manifest and configuration data.  Recognized
 formats include *oci* (OCI image-spec v1.0, the default) and *docker* (version
 2, using schema format 2 for the manifest).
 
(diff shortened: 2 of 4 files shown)


rhatdan:
I first brought this up back in https://github.com/projectatomic/buildah/issues/2

cmd/buildah/commit.go
```diff
@@ -128,5 +132,8 @@ func commitCmd(c *cli.Context) error {
 		return errors.Wrapf(err, "error committing container %q to %q", builder.Container, image)
 	}
 
+	if c.IsSet("reference-time") {
```
nalind: Copy-paste error?

tests/basic.bats
```diff
@@ -95,6 +95,12 @@ load helpers
   cmp ${TESTDIR}/other-randomfile $yetanothernewroot/other-randomfile
   buildah delete $yetanothernewcid
 
+  newcid=$(buildah from new-image)
+  buildah commit --rm --signature-policy ${TESTSDIR}/policy.json $newcid containers-storage:remove-container-image
+  newroot=$(buildah mount $newcid)
+  [ "$status" -ne 0 ]
```
nalind: Unless I'm mistaken, `bats(7)` says that `$status` is only updated when you use its `run` helper, so I don't think this'd be checking the exit status of the right command.  The `run` helper also captures the output of the command as `$output`, so there's no need for a subshell here.

docs/buildah-commit.md
```diff
@@ -33,11 +33,14 @@ Control the format for the image manifest and configuration data.  Recognized
 formats include *oci* (OCI image-spec v1.0, the default) and *docker* (version
 2, using schema format 2 for the manifest).
 
+**--rm**
+Remove the container content after committing it to an image.  Default leaves the container content.
```
TomSweeneyRedHat: I'd suggest:  "Remove the container and its content after committing it to an image.  Default leaves the container and its content in place.

cmd/buildah/commit.go
TomSweeneyRedHat: 👍

rhatdan: Fixed

tests/basic.bats
rhatdan: Fixed

docs/buildah-commit.md
rhatdan: Sounds good.

rhatdan:

nalind:
LGTM.

rhatdan:
@rh-atomic-bot r+ f03bfb1

nalind:
@rh-atomic-bot retry

PR #194, opened 2017-07-17
Add/Copy need to support glob syntax

rhatdan:
This patch allows users to do
buildah add $ctr * /dest

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -206,3 +206,16 @@ load helpers
   run buildah --debug=false images -q
   [ "$output" = "" ]
 }
+
+@test "bud-from-glob" {
+  target=alpine-image
+  buildah bud --signature-policy ${TESTSDIR}/policy.json -t ${target} -f Dockerfile2.glob ${TESTSDIR}/bud/from-multiple-files
+  cid=$(buildah from ${target})
+  root=$(buildah mount ${cid})
+  cmp $root/Dockerfile1.alpine ${TESTSDIR}/bud/from-multiple-files/Dockerfile1.alpine
+  cmp $root/Dockerfile2.withfrom ${TESTSDIR}/bud/from-multiple-files/Dockerfile2.withfrom
+  buildah rm ${cid}
+  buildah rmi $(buildah --debug=false images -q)
+  run buildah --debug=false images -q
+  [ "$output" = "" ]
+}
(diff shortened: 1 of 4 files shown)


rhatdan:
Fixes https://github.com/projectatomic/buildah/issues/188

nalind:
This needs a test.  The example in the first comment would let the shell do the globbing, so it would either need to be quoted to avoid that, or tested using a Dockerfile.

tests/bud.bats
```diff
@@ -206,3 +206,22 @@ load helpers
   run buildah --debug=false images -q
   [ "$output" = "" ]
 }
+
+@test "bud-from-glob" {
+  # This Dockerfile needs us to be able to handle a working RUN instruction.
```
TomSweeneyRedHat: cut/paste error?

TomSweeneyRedHat: nvm, not enough tea.

TomSweeneyRedHat:
LGTM

tests/copy.bats
```diff
@@ -21,6 +21,12 @@ load helpers
   buildah copy $cid ${TESTDIR}/randomfile
   buildah copy $cid ${TESTDIR}/other-randomfile ${TESTDIR}/third-randomfile ${TESTDIR}/randomfile /etc
   buildah rm $cid
+
+  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+  root=$(buildah mount $cid)
+  buildah config --workingdir / $cid
+  buildah copy $cid ${TESTDIR}/*randomfile /etc
```
nalind: Please put double quotes around `${TESTDIR}/*randomfile` so that the shell doesn't do the globbing, and add a loop to verify that copies all of the matching files ended up in the right place with the right contents.

tests/copy.bats
```diff
@@ -21,6 +21,13 @@ load helpers
   buildah copy $cid ${TESTDIR}/randomfile
   buildah copy $cid ${TESTDIR}/other-randomfile ${TESTDIR}/third-randomfile ${TESTDIR}/randomfile /etc
   buildah rm $cid
+
+  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+  root=$(buildah mount $cid)
+  buildah config --workingdir / $cid
+  buildah copy $cid "${TESTDIR}/*randomfile" /etc
+  (cd ${TESTDIR}; for i in *randomfile; do stat ${root}/etc/$i; done)
```
nalind: Please use `cmp` here.  (Aside: `stat` could be replaced here by `test -s`, which we use elsewhere, but then `cmp` is going to flag an error if the file doesn't exist, too.)

rhatdan: fixed

nalind:
LGTM pending CI results.

nalind:
@rh-atomic-bot r+ 79358a96a0003afaa33bf8fd2696044a4cd697cf

PR #195, opened 2017-07-17
Only print heading once for `buildah images`

rhatdan:
The current buildah images command prints the heading twice, this
bug was introduced when --json flag was added.


nalind:
LGTM, looks alright when I run it, too.

rhatdan:
@rh-atomic-bot r+ 9c8c3a1

PR #197, opened 2017-07-17
Add support for -- ending options parsing to buildah run

rhatdan:
If you specify an option in a buildah run command, the command fails.
The proper syntax for this is to add --

buildah run $ctr -- ls -l /

### DIFF
--- a/cmd/buildah/run.go
+++ b/cmd/buildah/run.go
@@ -51,6 +51,9 @@ func runCmd(c *cli.Context) error {
 	}
 	name := args[0]
 	args = args.Tail()
+	if len(args) > 0 && args[0] == "--" {
+		args = args[1:]
+	}
 
 	runtime := ""
 	if c.IsSet("runtime") {
--- a/tests/run.bats
+++ b/tests/run.bats
@@ -19,6 +19,17 @@ load helpers
 	buildah run        $cid cp /tmp/randomfile /tmp/other-randomfile
 	test -s $root/tmp/other-randomfile
 	cmp ${TESTDIR}/randomfile $root/tmp/other-randomfile
+	run buildah run $cid echo -n test
+	[ $status != 0 ]
+	run buildah run $cid echo -- -n test
+	[ $status != 0 ]
+	run buildah run $cid -- echo -n -- test
+	[ "$output" = "-- test" ]
+	run buildah run $cid -- echo -- -n test --
+	[ "$output" = "-- -n -- test --" ]
+	run buildah run $cid -- echo -n "test"
+	[ "$output" = "test" ]
+
 	buildah unmount $cid
 	buildah rm $cid
 }
--- a/docs/buildah-run.md
+++ b/docs/buildah-run.md
@@ -1,3 +1,4 @@
+
 ## buildah-run "1" "March 2017" "buildah"
 
 ## NAME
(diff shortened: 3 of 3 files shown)


rhatdan:
Fix https://github.com/projectatomic/buildah/issues/196

cmd/buildah/run.go
```diff
@@ -51,6 +51,9 @@ func runCmd(c *cli.Context) error {
 	}
 	name := args[0]
 	args = args.Tail()
+	if len(args) > 0 && args[0] == "--" {
```
TomSweeneyRedHat: Is the '--' standard?  I'm not seeing this in docker run.  gdb does --args, I'd prefer that as it's clearer and easier to search in a man page.

rhatdan: "--" is the CLI standard.  And it built into the CLI parsing in GOLANG.
https://unix.stackexchange.com/questions/55753/how-to-deal-with-end-of-options-in-getopts

rhatdan:
retest

rhatdan:
bot retest

cmd/buildah/run.go
TomSweeneyRedHat: ok thx for the background.  Your call but I still prefer --args.

rhatdan:
Added a test for parsing of --.

cmd/buildah/run.go
rhatdan: Docker run must be using a different parser, but I will still with "--"

rhatdan:
@nalind now passes the test.

rhatdan:
https://github.com/clcollins please verify this fixes your issue.

tests/run.bats
```diff
@@ -19,6 +19,7 @@ load helpers
 	buildah run        $cid cp /tmp/randomfile /tmp/other-randomfile
 	test -s $root/tmp/other-randomfile
 	cmp ${TESTDIR}/randomfile $root/tmp/other-randomfile
+	buildah run $cid -- rm -f /tmp/randomfile /tmp/other-randomfile
```
nalind: Might be better to test with 'echo' here, so that we can capture the output and verify that it matches what we expect (and specifically, that it doesn't include "--" when it shouldn't).  Otherwise LGTM.

rhatdan: So you want
run buildah run $cid --  echo -n foobar
[  "$output" == "foobar" ]
Do you also want a negative test
run buildah run $cid  echo -n foobar
[  $status != 1 ]

nalind: For a start.  Also might as well add
* run buildah run $cid echo -- -n foobar
* run buildah run $cid -- echo -n foobar --
* run buildah run $cid -- echo -- -n foobar --

Just to make sure that we're not interfering with or running afoul of any reordering that the cli package is trying to do on our behalf.

nalind:
LGTM.
@rh-atomic-bot r+ 75f616b6e926b1f2c0b0552ca32f3c5d3bae0068

rhatdan:
buildah-0.2.0-1.gitac2aad6.fc26 buildah-0.2.0-1.gitac2aad6.fc25 are in updates testing.

rhatdan:
@clcollins Thanks, keep finding those bugs...

PR #200, opened 2017-07-19
Improve buildah bash completions to return images or containers

rhatdan:
These changes will make it easier for users to select containers/images
to use rather then remembering their names.


rhatdan:
@nalind PTAL

rhatdan:
@rh-atomic-bot r+ 73b8d14

PR #201, opened 2017-07-20
Update image-spec and runtime-spec to v1.0.0

nalind:
Update to just-released versions of image-spec and runtime-spec, and the latest version of runtime-tools.  The runtime-spec dropped the `Platform` field, so stop trying to set it when calling out to a runtime as part of handling a `run` command or `RUN` instruction.

### DIFF
--- a/run.go
+++ b/run.go
@@ -153,12 +153,6 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	}()
 	g := generate.New()
 
-	if b.OS() != "" {
-		g.SetPlatformOS(b.OS())
-	}
-	if b.Architecture() != "" {
-		g.SetPlatformArch(b.Architecture())
-	}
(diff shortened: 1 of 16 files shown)


run.go
```diff
@@ -153,12 +153,6 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	}()
 	g := generate.New()
 
-	if b.OS() != "" {
```
TomSweeneyRedHat: Just curious, was this changed for the spec update or just something you meant to change?

nalind: The field was dropped from the runtime spec, and the runtime-tools API no longer provides the methods we were calling here.

TomSweeneyRedHat: LGTM and thx for the 411!

rhatdan:
@rh-atomic-bot r+ 16eb3c6

PR #211, opened 2017-07-25
Fixes for rmi

nalind:
* Switch to using storage APIs that don't bypass locking and synchronization.
* Properly close `Image` objects in a couple of places that we weren't.
* Look up images that may be specified using truncated IDs in `rmi`.
* Wrap some errors.

### DIFF
--- a/cmd/buildah/common.go
+++ b/cmd/buildah/common.go
@@ -96,7 +96,6 @@ func parseMetadata(image storage.Image) (imageMetadata, error) {
 }
 
 func getSize(image storage.Image, store storage.Store) (int64, error) {
-
 	is.Transport.SetStore(store)
 	storeRef, err := is.Transport.ParseStoreReference(store, "@"+image.ID)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM, should we add a check for rmi a truncated ID?

rhatdan:
@rh-atomic-bot r+ https://github.com/projectatomic/buildah/pull/211/commits/d3fb719d7318c3461d8b358bbe930548900e988f
Ok lets add these tests once your patches land.

rhatdan:
@jlebon @cgwalters We seem to be hung up?

cgwalters:
Hm, yeah the queue doesn't have the approval: https://homu-projectatomic-ci-infra.e8ca.engint.openshiftapps.com/queue/buildah

@rh-atomic-bot r=rhatdan d3fb719

jlebon:
@rhatdan Make sure to type in the commit id, not the URL of the commit (the final comment looks the same because of GitHub's autolinking, but rh-atomic-bot expects the SHA there). We also need to restart the deployment to get the fix for SHA-less r+s in: https://github.com/cgwalters/homu/pull/1.

PR #187, opened 2017-07-13
Update containers/storage and containers/image

nalind:
This bumps the versions of containers/storage and containers/image that we use, and updates the shortcuts that we take when committing to local storage to compensate for assumptions that it made which are no longer true (now it makes different ones).

PR #218, opened 2017-07-26
Avoid parsing metadata stored by the image library

nalind:
* Retrieve image creation dates from new public fields provided by the storage library.
* Compute image digests using the image manifest.

### DIFF
--- a/cmd/buildah/images.go
+++ b/cmd/buildah/images.go
@@ -190,11 +190,7 @@ func setFilterDate(images []storage.Image, imgName string) (time.Time, error) {
 		for _, name := range image.Names {
 			if matchesReference(name, imgName) {
 				// Set the date to this image
-				im, err := parseMetadata(image)
-				if err != nil {
-					return time.Time{}, errors.Wrapf(err, "could not get creation date for image %q", imgName)
-				}
-				date := im.CreatedTime
+				date := image.Created
 				return date, nil
 			}
 		}
@@ -222,7 +218,7 @@ func outputImages(images []storage.Image, format string, store storage.Store, fi
 		if err != nil {
 			fmt.Println(err)
 		}
-		createdTime := imageMetadata.CreatedTime.Format("Jan 2, 2006 15:04")
+		createdTime := image.Created.Format("Jan 2, 2006 15:04")
 		digest := ""
 		if len(imageMetadata.Blobs) > 0 {
 			digest = string(imageMetadata.Blobs[0].Digest)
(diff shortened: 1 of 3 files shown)


nalind:
Updated to dig image creation dates out of inspection data.

cmd/buildah/images.go
```diff
@@ -214,16 +215,15 @@ func outputHeader(truncate, digests bool) {
...
-			fmt.Println(err)
-		}
-		createdTime := image.Created.Format("Jan 2, 2006 15:04")
-		digest := ""
-		if len(imageMetadata.Blobs) > 0 {
-			digest = string(imageMetadata.Blobs[0].Digest)
+		createdTime := image.Created
+
+		inspectedTime, digest, size, _ := getDateAndDigestAndSize(image, store)
+		if !inspectedTime.IsZero() {
```
rhatdan: Is it ok if we have a createdTime, but inspectedTime.IsZero() is true?  Should we log this situation?

cmd/buildah/images.go
```diff
@@ -246,12 +246,11 @@ func outputImages(images []storage.Image, format string, store storage.Store, fi
 				ID:        image.ID,
 				Name:      name,
 				Digest:    digest,
-				CreatedAt: createdTime,
+				CreatedAt: createdTime.Format("Jan 2, 2006 15:04"),
```
rhatdan: Any significance to this date?

rhatdan:
Will this fix the issue where I see datas showing up as 1970?

cmd/buildah/images.go
nalind: If inspectedTime is zero, then the image probably didn't include a configuration blob, since IIRC Inspect() looks there.  In that case, using the date attached to the image is probably the most correct thing, since it's either the date the image record was created locally, or (after #187 or something like it gets merged) the same value we got back from Inspect().

cmd/buildah/images.go
nalind: The godoc for the `time` package explains that the values of the numeric fields are used to determine which field in a reference date structure is meant to be placed at that location in the output string.

rhatdan:
:+1:

nalind:
Yes, it should fix that.

rhatdan:
@TomSweeneyRedHat PTAL

rhatdan:
I am  going to merge.  Need to get a new buildah version out.

@rh-atomic-bot r+ c1e39f5

PR #226, opened 2017-07-27
Switch Travis from trusty to xenial

nalind:
Request a later platform for running our tests in Travis.  Since the `flatpak` PPA includes packages for `libostree-dev`, we no longer need to build our own.  CC @lsm5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,10 @@ dist: xenial
 sudo: required
 before_install:
     - sudo add-apt-repository -y ppa:duggan/bats
+    - sudo add-apt-repository -y ppa:alexlarsson/flatpak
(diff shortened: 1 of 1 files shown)


lsm5:
doesn't look like it's switching to xenial, since it can't locate libostree-dev from ppa. Might be because of https://github.com/travis-ci/travis-ci/issues/5821

lsm5:
the key fetches in travis logs also seem to be for trusty and not xenial. I guess we can leave this hanging until travis enables xenial (or until I enable buildah ppa for trusty, ughhh)

nalind:
It doesn't look like we can use xenial for sudo-requiring tests like ours, at least not yet.  We'll have to revisit this later.

PR #230, opened 2017-07-28
Bash completions should list containers or images by default

rhatdan:
This patch will get bash completions to list a container or an image
when a user hits the <tab> button without a leading -.

### DIFF
--- a/docs/buildah-images.md
+++ b/docs/buildah-images.md
@@ -12,15 +12,17 @@ Displays locally stored images, their names, and their IDs.
 ## OPTIONS
 
 **--json**
+
 Display the output in JSON format.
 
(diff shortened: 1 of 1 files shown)


rhatdan:
@TomSweeneyRedHat PTAL

TomSweeneyRedHat:
LGTM.  Two questions to possibly chase for later.  Should buildah run have a " __buildah_list_containers" near line 411?  Also buildah inspect can be used for containers or images, you've containers there which I think is right but thought I'd throw that by you to mull over.

rhatdan:
Yes I thought about the inspect, I will fix that and run should have it also.

Issue #229, opened 2017-07-28
document obtaining images from local dockerd


TomasTomecek:
I had a brief irc conversation with @nalind back then:

```
ttomecek>   nalind, hi, I'm trying buildah: I built a custom docker using
    docker and I want to use this image in buildah, so I'm running `sudo buildah
    from fwp` and it fails with docker registry error
ttomecek>   denied: requested access to the resource is denied
ttomecek>   unauthorized: authentication required
ttomecek>   is it possible to disable pull somehow? (so it just copies the
    image from local docker daemon?)
nalind> ttomecek: it doesn't share local storage with dockerd, so to pull an
    image from the local daemon, you could refer to the image, at least the
    first time, as 'docker-daemon:fwp:latest' and let it grab a copy
```

I am trying this out:

```
$ buildah from --pull=false docker-daemon:fedora:26-with-python
ERRO[0000] error parsing reference to image "docker-daemon:fedora:26-with-python": invalid reference format

$ docker images fedora                                                                                                                          1 ↵
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
fedora              26-with-python      57603fc684b5        10 days ago         409 MB
docker.io/fedora    26                  ac96429054d6        2 weeks ago         231 MB
fedora              25-with-python      0881717adf14        2 weeks ago         447 MB
docker.io/fedora    25                  15895ef0b3b2        3 months ago        231 MB
```

Can you help me out here?

rhatdan:
@mtrmac PTAL

rhatdan:
@TomasTomecek What happens if you remove pull=false?

TomasTomecek:
doesn't work :/

```
$ buildah from docker-daemon://fedora:26-with-python
ERRO[0000] error parsing reference to image "docker-daemon://fedora:26-with-python": invalid reference format
$ buildah from docker-daemon:fedora:26-with-python
ERRO[0000] error parsing reference to image "docker-daemon:fedora:26-with-python": invalid reference format
$ buildah from docker-daemon:fedora:26
ERRO[0000] error parsing reference to image "docker-daemon:fedora:26": invalid reference format
```

```
$ rpm -q buildah
buildah-0.3-2.gitb9b2a8a.fc27.x86_64
```

rhatdan:
Does skopeo copy work

skopeo copy docker-daemon:fedora:26-with-python dir:/tmp/fedora

TomasTomecek:
yup, that works just fine:

```
[...]

PR #234, opened 2017-07-31
Remove --transport flag

rhatdan:
This is no simpler then putting the transport in the image page,
we should default to the registry specified in containers/image
and not override it.  People are confused by this option, and I
see no value.

### DIFF
--- a/contrib/completions/bash/buildah
+++ b/contrib/completions/bash/buildah
@@ -340,7 +340,6 @@ return 1
   "
 
      local options_with_args="
-     --transport
      --signature-policy
      --runtime
      --runtime-flag
@@ -621,7 +620,6 @@ return 1
      --cert-dir
      --creds
      --name
-     --transport
      --signature-policy
      --tls-verify
   "
--- a/docs/buildah-from.md
+++ b/docs/buildah-from.md
@@ -37,12 +37,6 @@ Defaults to *true*.
 
 Pull the image even if a version of the image is already present.
 
-**--transport** *transport*
-
-A prefix to prepend to the image name in order to pull the image.  The default
-value is "docker://".  Note that no separator is implicitly added when the
-values are combined.
-
 **--signature-policy** *signaturepolicy*
 
 Pathname of a signature policy file to use.  It is not recommended that this
@@ -59,13 +53,13 @@ If an image needs to be pulled from the registry, suppress progress output.
 
 ## EXAMPLE
 
-buildah from imagename --pull --transport "docker://myregistry.example.com/"
+buildah from imagename --pull
 
 buildah from docker://myregistry.example.com/imagename --pull
 
 buildah from imagename --signature-policy /etc/containers/policy.json
 
-buildah from imagename --pull-always --transport "docker://myregistry.example.com/" --name "mycontainer"
+buildah from docker://myregistry.example.com/imagename --pull-always --name "mycontainer"
 
 buildah from myregistry/myrepository/imagename:imagetag --creds=myusername:mypassword
 
--- a/docs/buildah-from.md
+++ b/docs/buildah-from.md
@@ -8,7 +8,28 @@ buildah from - Creates a new working container, either from scratch or using a s
 
 ## DESCRIPTION
 Creates a working container based upon the specified image name.  If the
-supplied image name is "scratch" a new empty container is created.
+supplied image name is "scratch" a new empty container is created. Image names
+uses a "transport":"details" format.
+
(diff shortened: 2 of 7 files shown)


rhatdan:
https://github.com/projectatomic/buildah/issues/108

contrib/completions/bash/buildah
```diff
@@ -279,6 +279,9 @@ return 1
          -*)
              COMPREPLY=($(compgen -W "$boolean_options $options_with_args" -- "$cur"))
              ;;
+         *)
```
TomSweeneyRedHat: Didn't these changes for the bash completions go in with a different PR?

rhatdan:
I think we might have some other fixes to do to make this work.

@nalind Can I do a 

buildah from registry.access.redhat.com/fedora/rawhide

mtrmac:
This patch requires everyone to say **`docker://`**`registry.access.redhat.com/fedora/rawhide`.  I don’t mind :) but I guess interacting with docker/distribution registries by default is considered desirable?

#229 was not really about the flag being useless, rather about being unintuitively named.

rhatdan:
No I think we should allow the configuration to state where the content will go.  We can default to docker://  But adding

--registry registry.access.redhat.com fedora/rawhide 
Is not a great situation.  

I would prefer we follow the skopeo model, with a default transport being stored in a config file. not needing to have it specified on the command line.

mtrmac:
`--registry` is _not_ being used like `--registry registry.access.redhat.com fedora/rawhide`; that’s one of the ways it is confusing. It is used to specify the containers/image _transport_ (docker/oci-archive/oci-layout/atomic/containers-storage… , along with a possible prefix of the transport-specific location (which may be the `//` syntax, or perhaps a host name, or perhaps even more).

Skopeo actually always requires the transport to be specified; maybe `atomic` does not.

(Also see `kpod`; IIRC some of the commands try to use the input as transport:description, or `docker://`description, depending on what works. I’m not at all sure this is semantically sound / unambiguous, but if various parts of the CLI have to have ambiguous / guessing behavior, it would be nice for all of them to behave consistently.)

rhatdan:
Ok, I think we should default to the docker:// transport all over the place, and allow users to override by specifying the transport in the image name.  Not on some random misunderstood option.

TomasTomecek:
@rhatdan that would make it clear and predictable

rhatdan:
@nalind PTAL

docs/buildah-from.md
```diff
@@ -47,13 +41,13 @@ If an image needs to be pulled from the registry, suppress progress output.
...
 
-buildah from imagename --pull --transport "docker://myregistry.example.com/"
+buildah from imagename --pull
 
 buildah from docker://myregistry.example.com/imagename --pull
 
 buildah from imagename --signature-policy /etc/containers/policy.json
 
-buildah from imagename --pull-always --transport "docker://myregistry.example.com/" --name "mycontainer"
+buildah from docker://myregistry.example.com/imagename --pull-always --name "mycontainer"
```
TomasTomecek: It would be awesome to point to skopeo's manpage wrt different transports.

rhatdan: We have added some of that information to buildah man pages, but we could probably add more.

rhatdan: Checkout buildah-push man page.

TomasTomecek: That man page rocks! I think you should mention the push man page from man pages which operate on images.

rhatdan: I could also cut and paste it into the other man pages.

rhatdan:
@TomasTomecek Adding transport information to buildah from man page.  PTAL

rhatdan:

rhatdan:
@nalind Can I merge this?

rhatdan:
@nalind PTAL

docs/buildah-from.md
mtrmac: > I could also cut and paste it into the other man pages.

Eventually there probably should be a centralized man page, something like `container-transports(5)`, which everything could refer to; as we add new transports (or worse, change the semantics of existing transports, as is now happening with the `oci` image tag→reference name), there would be only one page to update instead of dozens; and if that man page lived in containers/image, it could be updated by the PR which changes the implementation.  Only… no such page exists yet I’m afraid.

nalind:
Alright, let's merge this.

rhatdan:
@rh-atomic-bot r+ 5b596a9

PR #235, opened 2017-07-31
Improve "from" behavior with unnamed references

nalind:
Fix our instantiation behavior when the source image reference is not a named reference.  Take the opportunity to rename the `--registry` CLI flag to `--transport`.

### DIFF
--- a/docs/buildah-from.md
+++ b/docs/buildah-from.md
@@ -29,10 +29,11 @@ Defaults to *true*.
 
 Pull the image even if a version of the image is already present.
 
-**--registry** *registry*
+**--transport** *transport*
 
-A prefix to prepend to the image name in order to pull the image.  Default
-value is "docker://"
+A prefix to prepend to the image name in order to pull the image.  The default
+value is "docker://".  Note that no separator is implicitly added when the
+values are combined.
 
 **--signature-policy** *signaturepolicy*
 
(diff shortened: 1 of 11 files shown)


new.go
```diff
@@ -18,97 +21,113 @@ const (
...
+			return nil, errors.Wrapf(err, "error reading image configuration for %q", transports.ImageName(ref))
 		}
 		manifest, _, err = src.Manifest()
 		if err != nil {
-			return nil, errors.Wrapf(err, "error reading image manifest")
+			return nil, errors.Wrapf(err, "error reading image manifest for %q", transports.ImageName(ref))
 		}
 	}
 
+	name := "working-container"
```
rhatdan: Should this be made into a function?

docs/buildah-from.md
```diff
@@ -46,13 +46,13 @@ If an image needs to be pulled from the registry, suppress progress output.
 
 ## EXAMPLE
 
-buildah from imagename --pull --registry "myregistry://"
+buildah from imagename --pull --transport "docker://myregistry.example.com/"
```
rhatdan: Why is this easier than 

buildah from docker://myregistry.example.com/imagename

nalind: Both should provide equivalent results, but the `--transport` option is necessary for `build-using-dockerfile` anyway.

nalind:
Fixed an inconsistency (was `--default-transport` in `build-using-dockerfile`, instead of `--transport`, as it should have been), and added a note stressing that we don't insert any additional separator when combining the transport name with the image name.

nalind:
What errors are you getting back when you try to commit or push to a `docker-daemon:` destination?

rhatdan:
I am talking about https://github.com/projectatomic/buildah/issues/229

docs/buildah-from.md
rhatdan: Why can't I do 
buildah from docker://myregistry.example.com/imagename

nalind:
Yes, this should fix those pull errors.

docs/buildah-from.md
nalind: You can.

rhatdan: If I can, why do I need --transport?

nalind:
Hmm, no it doesn't.

rhatdan:
Ok I will approve this patch and open another one to remove --transport.  Just so we can fix the issue.
@rh-atomic-bot r+ fb32f53

PR #238, opened 2017-08-01
atomic transport is being deprecated, so we should not document it.

rhatdan:
containers/image now fully supports pushing images and signatures to an
openshift/atomic registry using the docker:// transport.


nalind:
LGTM

rhatdan:
@rh-atomic-bot r+ 52df43e

PR #241, opened 2017-08-10
Run(): create the right working directory

nalind:
When ensuring that the working directory exists before running a command, make sure we create the location that we actually set in the configuration file that we passed to runc.


rhatdan:
@rh-atomic-bot r+ 1214c58

PR #248, opened 2017-08-18
Run: don't complain about missing volume locations

nalind:
Don't worry about not being able to populate temporary volumes using the contents of the location in the image where they're expected to be mounted if we fail to do so because that location doesn't exist.

### DIFF
--- a/run.go
+++ b/run.go
@@ -120,7 +120,7 @@ func (b *Builder) setupMounts(mountPoint string, spec *specs.Spec, optionMounts 
 				return errors.Wrapf(err, "error creating directory %q for volume %q in container %q", volumePath, volume, b.ContainerID)
 			}
 			srcPath := filepath.Join(mountPoint, volume)
-			if err = archive.CopyWithTar(srcPath, volumePath); err != nil {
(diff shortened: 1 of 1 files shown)


run.go
```diff
@@ -120,7 +120,7 @@ func (b *Builder) setupMounts(mountPoint string, spec *specs.Spec, optionMounts
 				return errors.Wrapf(err, "error creating directory %q for volume %q in container %q", volumePath, volume, b.ContainerID)
 			}
 			srcPath := filepath.Join(mountPoint, volume)
-			if err = archive.CopyWithTar(srcPath, volumePath); err != nil {
+			if err = archive.CopyWithTar(srcPath, volumePath); err != nil && !os.IsNotExist(err) {
```
nalind: Define a volume in the image, but don't include its directory in any of the filesystem layers, and try to run it using `buildah run` without assigning a directory on the host to use as the volume.

`Run()` will create a temporary directory on the host to use for the volume, and then try to populate it with the contents of the volume's location as provided by the image, so that it can bind it to the volume's location, as would be consistent with the semantics of the Dockerfile `RUN` instruction.  It'll fail at the second step.

rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+ 87eb550

PR #250, opened 2017-08-24
Add authentication to commit and push

TomSweeneyRedHat:

Adds authentication to commit and push, including script that can be run at the command line to test.  Updates to *.bats tests will be added in a follow up.

### DIFF
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -20,9 +20,18 @@ var (
 			Usage: "don't compress layers",
 		},
 		cli.StringFlag{
+			Name:  "cert-dir",
+			Value: "",
+			Usage: "use certificates at the specified path to access the registry",
+		},
+		cli.StringFlag{
 			Name:  "creds",
 			Value: "",
 			Usage: "use `username[:password]` for accessing the registry",
+		},
+		cli.BoolTFlag{
+			Name:  "tls-verify",
+			Usage: "Require HTTPS and verify certificates when accessing the registry",
 		},
 		cli.StringFlag{
 			Name:  "signature-policy",
@@ -123,11 +132,17 @@ func commitCmd(c *cli.Context) error {
 		dest = dest2
 	}
 
+	systemContext, err := systemContextFromOptions(c)
+	if err != nil {
+		return errors.Wrapf(err, "error building system context")
+	}
+
 	options := buildah.CommitOptions{
 		PreferredManifestType: format,
 		Compression:           compress,
 		SignaturePolicyPath:   signaturePolicy,
 		HistoryTimestamp:      &timestamp,
+		SystemContext:         systemContext,
 	}
 	if !quiet {
 		options.ReportWriter = os.Stderr
--- a/cmd/buildah/push.go
+++ b/cmd/buildah/push.go
@@ -20,6 +20,20 @@ var (
 			Usage: "don't compress layers",
 		},
 		cli.StringFlag{
+			Name:  "cert-dir",
+			Value: "",
+			Usage: "use certificates at the specified path to access the registry",
+		},
+		cli.StringFlag{
+			Name:  "creds",
+			Value: "",
+			Usage: "use `username[:password]` for accessing the registry",
+		},
+		cli.BoolTFlag{
+			Name:  "tls-verify",
+			Usage: "Require HTTPS and verify certificates when accessing the registry",
+		},
+		cli.StringFlag{
 			Name:  "signature-policy",
 			Usage: "`pathname` of signature policy file (not usually used)",
 		},
@@ -79,10 +93,16 @@ func pushCmd(c *cli.Context) error {
 		return err
 	}
 
+	systemContext, err := systemContextFromOptions(c)
+	if err != nil {
+		return errors.Wrapf(err, "error building system context")
+	}
+
 	options := buildah.PushOptions{
 		Compression:         compress,
 		SignaturePolicyPath: signaturePolicy,
 		Store:               store,
+		SystemContext:       systemContext,
 	}
 	if !quiet {
 		options.ReportWriter = os.Stderr
--- a/commit.go
+++ b/commit.go
@@ -54,6 +54,9 @@ type CommitOptions struct {
 	// HistoryTimestamp is the timestamp used when creating new items in the
 	// image's history.  If unset, the current time will be used.
 	HistoryTimestamp *time.Time
+	// github.com/containers/image/types SystemContext to hold credentials
+	// and other authentication/authorization information.
+	SystemContext *types.SystemContext
 }
 
 // PushOptions can be used to alter how an image is copied somewhere.
@@ -73,6 +76,9 @@ type PushOptions struct {
 	ReportWriter io.Writer
 	// Store is the local storage store which holds the source image.
 	Store storage.Store
+	// github.com/containers/image/types SystemContext to hold credentials
+	// and other authentication/authorization information.
+	SystemContext *types.SystemContext
 }
 
 // shallowCopy copies the most recent layer, the configuration, and the manifest from one image to another.
@@ -249,8 +255,7 @@ func (b *Builder) Commit(dest types.ImageReference, options CommitOptions) error
 	}
 	if exporting {
 		// Copy everything.
-		// TODO: add credsContext
-		err = cp.Image(policyContext, dest, src, getCopyOptions(options.ReportWriter, nil, nil))
+		err = cp.Image(policyContext, dest, src, getCopyOptions(options.ReportWriter, nil, options.SystemContext))
 		if err != nil {
 			return errors.Wrapf(err, "error copying layers and metadata")
 		}
@@ -322,8 +327,7 @@ func Push(image string, dest types.ImageReference, options PushOptions) error {
 		return errors.Wrapf(err, "error recomputing layer digests and building metadata")
 	}
 	// Copy everything.
-	// TODO: add credsContext
-	err = cp.Image(policyContext, dest, src, getCopyOptions(options.ReportWriter, nil, nil))
+	err = cp.Image(policyContext, dest, src, getCopyOptions(options.ReportWriter, nil, options.SystemContext))
 	if err != nil {
 		return errors.Wrapf(err, "error copying layers and metadata")
 	}
--- a/contrib/completions/bash/buildah
+++ b/contrib/completions/bash/buildah
@@ -291,9 +291,12 @@ return 1
           --quiet
           -q
           --rm
+          --tls-verify
   "
 
      local options_with_args="
+          --cert-dir
+          --creds
           --signature-policy
           --format
           -f
@@ -471,9 +474,12 @@ return 1
           -D
           --quiet
           -q
+          --tls-verify
   "
 
      local options_with_args="
+          --cert-dir
+          --creds
           --signature-policy
   "
 
@@ -614,6 +620,7 @@ return 1
      --pull-always
      --quiet
      -q
+     --tls-verify
   "
 
      local options_with_args="
@@ -621,7 +628,6 @@ return 1
      --creds
      --name
      --signature-policy
-     --tls-verify
   "
 
 
--- a/docs/buildah-commit.md
+++ b/docs/buildah-commit.md
@@ -13,6 +13,14 @@ specified, an ID is assigned, but no name is assigned to the image.
 
 ## OPTIONS
 
+**--cert-dir** *path*
+
+Use certificates at *path* (*.crt, *.cert, *.key) to connect to the registry
+
+**--creds** *creds*
+
+The username[:password] to use to authenticate with the registry if required.
+
 **--disable-compression, -D**
 
 Don't compress filesystem layers when building the image.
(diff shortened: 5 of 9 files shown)


commit.go
```diff
@@ -249,8 +255,7 @@ func (b *Builder) Commit(dest types.ImageReference, options CommitOptions) error
 	}
 	if exporting {
 		// Copy everything.
-		// TODO: add credsContext
-		err = cp.Image(policyContext, dest, src, getCopyOptions(options.ReportWriter, nil, nil))
+		err = cp.Image(policyContext, dest, src, getCopyOptions(options.ReportWriter, nil, options.SystemContext))
```
TomSweeneyRedHat: Question for @nalind .  cp.Image() does dest than src.  getCopyOptions does src then dest.  Can we change cp.Image or should I change getCopyOptions for the ordering?  I'm happy to do either, just wondering if changing cp.Image makes you twitch too much.

rhatdan:
I prefer src, dest

docs/buildah-commit.md
```diff
@@ -13,6 +13,14 @@ specified, an ID is assigned, but no name is assigned to the image.
 
 ## OPTIONS
 
+**--cert-dir** *path*
+
+Use certificates at *path* (*.crt, *.cert, *.key) to connect to the registry
```
rhatdan: Is there a default?

docs/buildah-commit.md
```diff
@@ -13,6 +13,14 @@ specified, an ID is assigned, but no name is assigned to the image.
...
+
+**--creds** *creds*
+
+The username and password to use to authenticate with the registry if required.
```
rhatdan: Should we specify format username[:password]

cmd/buildah/push.go
```diff
@@ -20,6 +20,20 @@ var (
...
+			Name:  "cert-dir",
+			Value: "",
+			Usage: "use certificates at the specified path to access the registry",
+		},
+		cli.StringFlag{
+			Name:  "creds",
+			Value: "",
+			Usage: "use `username[:password]` for accessing the registry",
+		},
+		cli.StringFlag{
```
rhatdan: Shouldn't this be a bool?  And it needs to default to true, according to man page below.

cmd/buildah/commit.go
```diff
@@ -20,11 +20,20 @@ var (
...
+			Name:  "cert-dir",
+			Value: "",
+			Usage: "use certificates at the specified path to access the registry",
+		},
+		cli.StringFlag{
 			Name:  "creds",
 			Value: "",
 			Usage: "use `username[:password]` for accessing the registry",
 		},
 		cli.StringFlag{
```
rhatdan: Shouldn't this be a bool? And it needs to default to true, according to man page below.

docs/buildah-commit.md
```diff
@@ -39,13 +51,21 @@ Default leaves the container and its content in place.
...
+This example saves an image based on the container.
+ `buildah commit containerID`
+
+This example saves an image named newImageName based on the container.
+ `buildah commit --rm containerID newImageName`
+
 
-buildah commit --rm containerID newImageName
+This example saves an image based on the container disabling compression and specifying a signature policy (not recommended).
+ `buildah commit --disable-compression --signature-policy '/etc/containers/policy.json' containerID`
```
rhatdan: Why are we specifying --signature-policy, isn't this the default?  I actually thought this was a hidden field.  Specifying in each example, makes it seem like something useful for a user to do, but we expect them to almost never do it.

docs/buildah-push.md
```diff
@@ -40,6 +40,14 @@ Image stored in local container/storage
 
 ## OPTIONS
 
+**--cert-dir** *path*
+
+Use certificates at *path* (*.crt, *.cert, *.key) to connect to the registry
```
rhatdan: Default?

docs/buildah-push.md
```diff
@@ -40,6 +40,14 @@ Image stored in local container/storage
...
+
+**--creds** *creds*
+
+The username and password to use to authenticate with the registry if required.
```
rhatdan: Format username[:password]

docs/buildah-commit.md
TomSweeneyRedHat: It doesn't default and there's some wonkiness going on that I need to talk over with Miloslav about which may be my lack of openssl cli knowledge.   I'm like this as is, but I'm fine with changing this here if you wish.  But it's the same verbiage as in skopeo and we'd probably want to touch that up too.

docs/buildah-commit.md
TomSweeneyRedHat: yeah, we should.  I'll make it so.

docs/buildah-push.md
TomSweeneyRedHat: See comment in commit plz.

docs/buildah-push.md
TomSweeneyRedHat: yes, same as commit.  Ty for picking this one up, I missed that.

cmd/buildah/commit.go
TomSweeneyRedHat: I think you're right.  WIll go play with that.

cmd/buildah/push.go
TomSweeneyRedHat: Most likely, will play.

TomSweeneyRedHat:
@Spindel another buildah auth submit that may be of interest.

cmd/buildah/push.go
TomSweeneyRedHat: Fixed, good catch ty.  Changed to BoolTFlag which defaults to true.

cmd/buildah/commit.go
TomSweeneyRedHat: Fixed, good catch ty.  Changed to BoolTFlag which defaults to true.

docs/buildah-commit.md
TomSweeneyRedHat: @nalind thoughts on this.  I primarily just added the comment over each and the one new example for the auth.

commit.go
nalind: TBH, I'm not at all bothered if they stay as they are.

contrib/completions/bash/buildah
```diff
@@ -284,13 +284,16 @@ return 1
 
  _buildah_commit() {
      local boolean_options="
+          --cert-dir
+          --creds
```
nalind: These aren't boolean options.

contrib/completions/bash/buildah
```diff
@@ -466,12 +469,15 @@ return 1
 
  _buildah_push() {
      local boolean_options="
+          --cert-dir
+          --creds
```
nalind: These aren't boolean options.

contrib/completions/bash/buildah
```diff
@@ -609,12 +615,15 @@ return 1
 
  _buildah_from() {
      local boolean_options="
+     --cert-dir
+     --creds
```
nalind: These aren't boolean options.

contrib/completions/bash/buildah
```diff
@@ -609,12 +615,15 @@ return 1
...
      --help
      -h
      --pull
      --pull-always
      --quiet
      -q
+     --tls-verify
```
nalind: This _is_ a boolean option, but it duplicates a value listed in `$options_with_args`, which should be removed from that while we're adding this one here.

tests/test_buildah_authentication.sh
```diff
@@ -0,0 +1,181 @@
+#!/bin/bash
+# test_buildah_authentication
+# A script to be run at the command line with Buildah installed.
```
nalind: Extra whitespace at the end of this line.

tests/test_buildah_authentication.sh
```diff
@@ -0,0 +1,181 @@
...
+########
+docker ps --all
+
+docker images
+
+buildah containers
+
+buildah images
+
+
```
nalind: Extra whitespace at the end of this line, which can be removed.

tests/test_buildah_authentication.sh
```diff
@@ -0,0 +1,181 @@
...
+buildah containers
+
+buildah images
+
+########
+# Clean up Buildah
+########
+buildah rm $ctrid
+buildah rmi -f localhost:5000/my-alpine:latest
```
nalind: Extra whitespace at the end of this line.

docs/buildah-commit.md
nalind: If I'm reading things (https://github.com/containers/image/blob/master/docker/docker_client.go#L228, https://github.com/containers/image/blob/master/docker/docker_client.go#L136, and the two test cases at https://github.com/containers/image/blob/master/docker/reference/reference_test.go#L45) right, what we're setting will lead us to a subdirectory of `/etc/docker/certs.d` named after the host and port (if specified) of the registry which we're contacting.

docs/buildah-commit.md
nalind: It is the default.  The option is documented higher up with a note that its use is not recommended, so that the system-wide default will be used.  Removing it from the examples should be fine.

nalind:
Docs and command completion nits, otherwise LGTM.

contrib/completions/bash/buildah
TomSweeneyRedHat: Oh shoot, messed that up.  Ty, fixed.

contrib/completions/bash/buildah
TomSweeneyRedHat: fixed and ty.

contrib/completions/bash/buildah
TomSweeneyRedHat: fixed.

contrib/completions/bash/buildah
TomSweeneyRedHat: touched this up as well, fixed ty.

docs/buildah-commit.md
TomSweeneyRedHat: ok, I'll remove it from all but one of the examples here then.

TomSweeneyRedHat: Actually I lied, I'll remove it entirely from the examples.

tests/test_buildah_authentication.sh
TomSweeneyRedHat: fixed

tests/test_buildah_authentication.sh
TomSweeneyRedHat: I've removed the line entirely.

tests/test_buildah_authentication.sh
TomSweeneyRedHat: fixed

docs/buildah-commit.md
rhatdan: We probably need to change this code to start using /etc/containers/certs.d as well.  But not in the pull request.  I guess I would rather not point out this default in the man page.

docs/buildah-commit.md
```diff
@@ -39,13 +51,21 @@ Default leaves the container and its content in place.
...
+This example saves an image based on the container disabling compression.
+ `buildah commit --disable-compression containerID`
 
-buildah commit --disable-compression --signature-policy '/etc/containers/policy.json' containerID
+This example saves an image named newImageName based on the container disabling compression.
+ `buildah commit --disable-compression containerID newImageName`
 
-buildah commit --disable-compression --signature-policy '/etc/containers/policy.json' containerID newImageName
+This example commits the container to the image on the local registry using credentials and certificates for authentication.
+ `buildah commit --cert-dir ~/auth  --tls-verify true --creds=username:password containerID docker://localhost:5000/imageId`
```
rhatdan: nit: A example where you turn tls-verify to false, would be more likely, since the default is on.

rhatdan:
Overall LGTM

docs/buildah-commit.md
TomSweeneyRedHat: Ok, sounds good.  I've created Issue #251 so we can track this.

docs/buildah-commit.md
TomSweeneyRedHat: Good point, adding.

rhatdan:
@nalind PTAL

cmd/buildah/commit.go
```diff
@@ -123,11 +132,17 @@ func commitCmd(c *cli.Context) error {
 		dest = dest2
 	}
 
+	systemContext, err := systemContextFromOptions(c)
+	if err != nil {
+		return errors.Errorf("error building system context [%v]", err)
```
nalind: Change this to return `errors.Wrapf(err, "error building system context")`.

cmd/buildah/push.go
```diff
@@ -79,10 +93,16 @@ func pushCmd(c *cli.Context) error {
 		return err
 	}
 
+	systemContext, err := systemContextFromOptions(c)
+	if err != nil {
+		return errors.Errorf("error building system context [%v]", err)
```
nalind: Change this to return `errors.Wrapf(err, "error building system context")`.

nalind:
A couple of changes to wrap errors to be more helpful for debugging, otherwise LGTM.

cmd/buildah/commit.go
TomSweeneyRedHat: Fixed

cmd/buildah/push.go
TomSweeneyRedHat: Fixed.  Circled back and changed in from.go too.

nalind:
LGTM.  Merging.
@rh-atomic-bot r+

TomSweeneyRedHat:
@nalind The last change that I pushed touched up the errors to use Wrapf as you requested.  I did the same for from.go too.

rhatdan:
@rh-atomic-bot r+ 888c08e

PR #252, opened 2017-08-28
Auto-set build tags for ostree and selinux

nalind:
Try to use pkg-config to check for 'ostree-1' and 'libselinux'.

If ostree-1 is not found, use the containers_image_ostree_stub build tag to not require it, at the cost of not being able to use or write images to the 'ostree' transport.

If libselinux is found, build with the 'selinux' tag,


rhatdan:
LGTM

nalind:
Question: should we doing something similar with apparmor and seccomp?

rhatdan:
If they have buildtags, yes.

rhatdan:
@rh-atomic-bot r+ a247474

nalind:
On further investigation, we probably want to expect libseccomp anywhere, and we're installing libapparmor in Travis for the sake of runc, so there's nothing to be done for them at this time.

PR #255, opened 2017-08-31
Avoid trying to print a nil ImageReference

nalind:
When we fail to pull an image, don't try to include the name of the image that pullImage() returned in the error text - it will have returned nil for the pulled reference in most cases.  Instead, use the name of the image as we were told it.


nalind:
Thanks!  Merging.
@rh-atomic-bot r+

PR #260, opened 2017-09-09
Add default transport to push if not provided

TomSweeneyRedHat:

Currently with 'buildah push' if the transport is not defined, it will fail.  This change will retry if a transport was not found by adding "docker://" to the front of the destination image and retrying.

I tweaked a test in the test script to test for this condition and also converted the creation of htpasswd from Docker to Buildah.  I tried running the registry with Buildah too, but 'docker login' wasn't happy.  More investigation there in  a follow up.

### DIFF
--- a/cmd/buildah/push.go
+++ b/cmd/buildah/push.go
@@ -88,9 +88,20 @@ func pushCmd(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
+
 	dest, err := alltransports.ParseImageName(destSpec)
+	// add the docker:// transport to see if they neglected it.
 	if err != nil {
-		return err
+		if strings.Contains(destSpec, "://") {
+			return err
+		}
+
+		destSpec = "docker://" + destSpec
+		dest2, err2 := alltransports.ParseImageName(destSpec)
+		if err2 != nil {
+			return err
+		}
+		dest = dest2
 	}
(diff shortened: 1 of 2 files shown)


cmd/buildah/push.go
```diff
@@ -88,9 +88,20 @@ func pushCmd(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
+
 	dest, err := alltransports.ParseImageName(destSpec)
+	// add the docker:// transport to see if they neglected it.
 	if err != nil {
-		return err
+		if !strings.Contains(destSpec, "://") {
```
rhatdan: Reverse the error
if err strings.Container(destSpec,"://") {
    return err
}

rhatdan:
LGTM

TomSweeneyRedHat:
@nalind PTAL

nalind:
LGTM

rhatdan:
@rh-atomic-bot r+ f1b6b46

rhatdan:
@rh-atomic-bot retest

rhatdan:
@rh-atomic-bot retry

rhatdan:
@rh-atomic-bot retry

rhatdan:
@rh-atomic-bot retest

PR #267, opened 2017-09-22
Cleanup buildah code

rhatdan:
1. Sort options so they are in alphabet order
2. Remove extra lines of code for options parsing that really do not accomplish anything.
3. Remove variables when they are not necesary, IE Don't create a variable to hold an
option that is only used once, use the option instead.

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -13,9 +13,17 @@ import (
 
 var (
 	budFlags = []cli.Flag{
-		cli.BoolFlag{
-			Name:  "quiet, q",
-			Usage: "refrain from announcing build instructions and image read/write progress",
+		cli.StringSliceFlag{
+			Name:  "build-arg",
+			Usage: "`argument=value` to supply to the builder",
+		},
+		cli.StringSliceFlag{
+			Name:  "file, f",
+			Usage: "`pathname or URL` of a Dockerfile",
+		},
+		cli.StringFlag{
+			Name:  "format",
+			Usage: "`format` of the built image's manifest and metadata",
 		},
 		cli.BoolTFlag{
 			Name:  "pull",
@@ -25,13 +33,9 @@ var (
 			Name:  "pull-always",
 			Usage: "pull the image, even if a version is present",
 		},
-		cli.StringFlag{
-			Name:  "signature-policy",
-			Usage: "`pathname` of signature policy file (not usually used)",
-		},
-		cli.StringSliceFlag{
-			Name:  "build-arg",
-			Usage: "`argument=value` to supply to the builder",
+		cli.BoolFlag{
+			Name:  "quiet, q",
+			Usage: "refrain from announcing build instructions and image read/write progress",
 		},
 		cli.StringFlag{
 			Name:  "runtime",
@@ -43,18 +47,15 @@ var (
 			Usage: "add global flags for the container runtime",
 		},
 		cli.StringFlag{
-			Name:  "format",
-			Usage: "`format` of the built image's manifest and metadata",
+			Name:  "signature-policy",
+			Usage: "`pathname` of signature policy file (not usually used)",
 		},
 		cli.StringSliceFlag{
 			Name:  "tag, t",
 			Usage: "`tag` to apply to the built image",
 		},
-		cli.StringSliceFlag{
-			Name:  "file, f",
-			Usage: "`pathname or URL` of a Dockerfile",
-		},
 	}
+
 	budDescription = "Builds an OCI image using instructions in one or more Dockerfiles."
 	budCommand     = cli.Command{
 		Name:        "build-using-dockerfile",
@@ -77,35 +78,14 @@ func budCmd(c *cli.Context) error {
 			tags = tags[1:]
 		}
 	}
-	pull := true
-	if c.IsSet("pull") {
-		pull = c.BoolT("pull")
-	}
-	pullAlways := false
-	if c.IsSet("pull-always") {
-		pull = c.Bool("pull-always")
-	}
-	runtimeFlags := []string{}
-	if c.IsSet("runtime-flag") {
-		runtimeFlags = c.StringSlice("runtime-flag")
-	}
-	runtime := ""
-	if c.IsSet("runtime") {
-		runtime = c.String("runtime")
-	}
-
 	pullPolicy := imagebuildah.PullNever
-	if pull {
+	if c.BoolT("pull") {
 		pullPolicy = imagebuildah.PullIfMissing
 	}
-	if pullAlways {
+	if c.Bool("pull-always") {
 		pullPolicy = imagebuildah.PullAlways
 	}
 
-	signaturePolicy := ""
-	if c.IsSet("signature-policy") {
-		signaturePolicy = c.String("signature-policy")
-	}
 	args := make(map[string]string)
 	if c.IsSet("build-arg") {
 		for _, arg := range c.StringSlice("build-arg") {
@@ -117,14 +97,8 @@ func budCmd(c *cli.Context) error {
 			}
 		}
 	}
-	quiet := false
-	if c.IsSet("quiet") {
-		quiet = c.Bool("quiet")
-	}
-	dockerfiles := []string{}
-	if c.IsSet("file") || c.IsSet("f") {
-		dockerfiles = c.StringSlice("file")
-	}
+
+	dockerfiles := c.StringSlice("file")
 	format := "oci"
 	if c.IsSet("format") {
 		format = strings.ToLower(c.String("format"))
@@ -200,16 +174,16 @@ func budCmd(c *cli.Context) error {
 		ContextDirectory:    contextDir,
 		PullPolicy:          pullPolicy,
 		Compression:         imagebuildah.Gzip,
-		Quiet:               quiet,
-		SignaturePolicyPath: signaturePolicy,
+		Quiet:               c.Bool("quiet"),
+		SignaturePolicyPath: c.String("signature-policy"),
 		Args:                args,
 		Output:              output,
 		AdditionalTags:      tags,
-		Runtime:             runtime,
-		RuntimeArgs:         runtimeFlags,
+		Runtime:             c.String("runtime"),
+		RuntimeArgs:         c.StringSlice("runtime-flag"),
 		OutputFormat:        format,
 	}
-	if !quiet {
+	if !c.Bool("quiet") {
 		options.ReportWriter = os.Stderr
 	}
 
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -15,10 +15,6 @@ import (
 
 var (
 	commitFlags = []cli.Flag{
-		cli.BoolFlag{
-			Name:  "disable-compression, D",
-			Usage: "don't compress layers",
-		},
 		cli.StringFlag{
 			Name:  "cert-dir",
 			Value: "",
@@ -29,17 +25,18 @@ var (
 			Value: "",
 			Usage: "use `username[:password]` for accessing the registry",
 		},
-		cli.BoolTFlag{
-			Name:  "tls-verify",
-			Usage: "Require HTTPS and verify certificates when accessing the registry",
-		},
-		cli.StringFlag{
-			Name:  "signature-policy",
-			Usage: "`pathname` of signature policy file (not usually used)",
+		cli.BoolFlag{
+			Name:  "disable-compression, D",
+			Usage: "don't compress layers",
 		},
 		cli.StringFlag{
 			Name:  "format, f",
 			Usage: "`format` of the image manifest and metadata",
+			Value: "oci",
+		},
+		cli.BoolFlag{
+			Name:  "quiet, q",
+			Usage: "don't output progress information when writing images",
 		},
 		cli.StringFlag{
 			Name:   "reference-time",
@@ -47,12 +44,16 @@ var (
 			Hidden: true,
 		},
 		cli.BoolFlag{
-			Name:  "quiet, q",
-			Usage: "don't output progress information when writing images",
-		},
-		cli.BoolFlag{
 			Name:  "rm",
 			Usage: "remove the container and its content after committing it to an image. Default leaves the container and its content in place.",
+		},
+		cli.StringFlag{
+			Name:  "signature-policy",
+			Usage: "`pathname` of signature policy file (not usually used)",
+		},
+		cli.BoolTFlag{
+			Name:  "tls-verify",
+			Usage: "Require HTTPS and verify certificates when accessing the registry",
 		},
 	}
 	commitDescription = "Writes a new image using the container's read-write layer and, if it is based\n   on an image, the layers of that image"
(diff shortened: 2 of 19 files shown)


rhatdan:
bot, retest please

jlebon:
bot, retest this please

cmd/buildah/bud.go
```diff
@@ -43,18 +43,15 @@ var (
...
-			Name:  "format",
-			Usage: "`format` of the built image's manifest and metadata",
+			Name:  "signature-policy",
+			Usage: "`pathname` of signature policy file (not usually used)",
 		},
 		cli.StringSliceFlag{
 			Name:  "tag, t",
 			Usage: "`tag` to apply to the built image",
 		},
-		cli.StringSliceFlag{
```
TomSweeneyRedHat: Don't we still need file?

rhatdan: Yes we do.  Nice catch.  I wish this would be caught by the compiler.

Fixed

cmd/buildah/run.go
```diff
@@ -78,25 +65,19 @@ func runCmd(c *cli.Context) error {
...
-	if c.IsSet("tty") {
-		if c.Bool("tty") {
-			options.Terminal = buildah.WithTerminal
-		} else {
-			options.Terminal = buildah.WithoutTerminal
-		}
+	if c.Bool("tty") {
+		options.Terminal = buildah.WithTerminal
+	} else {
+		options.Terminal = buildah.WithoutTerminal
```
TomSweeneyRedHat: Six of one, half dozen of another.  Perhaps:
options.Terminal = buildah.WithoutTerminal
if c.Bool("tty") {
    opitions.Terminal = buildah.WithTerminal
}

TomSweeneyRedHat:
Looks like the test is suffering from some kind of version wonkiness.
not ok 45 buildah version up to date in .spec file

# (in test file ./version.bats, line 15)

#   `test "$bversion" = "$rversion"' failed

cmd/buildah/run.go
rhatdan: Well we have a lot of code that does this method.  From a anal performance point of view,  The current code is better.  From the total lines of code poit of view, yours is better.  :^)

rhatdan:
Lets merge https://github.com/projectatomic/buildah/pull/269 to fix the version issue, which I caused.  :^(

rhatdan:
@TomSweeneyRedHat @nalind Now passes Tests PTAL
@umohnani8 PTAL

cmd/buildah/images.go
```diff
@@ -93,28 +93,10 @@ func imagesCmd(c *cli.Context) error {
...
-	formatString := ""
-	hasTemplate := false
-	if c.IsSet("format") {
-		formatString = c.String("format")
-		hasTemplate = true
-	}
+	quiet := c.Bool("quiet")
+	truncate := !c.Bool("no-trunc")
+	digests := c.Bool("digests")
+	hasTemplate := !c.IsSet("format")
```
TomSweeneyRedHat: I think this one should not have the '!'.

cmd/buildah/run.go
```diff
@@ -84,19 +71,17 @@ func runCmd(c *cli.Context) error {
...
 	}
 
-	for _, volumeSpec := range volumes {
+	for _, volumeSpec := range c.StringSlice("volume") {
```
TomSweeneyRedHat: I'm going back/forth.  Should "volume" be "volumes"?  Just seems odd to have a range against something that's singular.  We are dealing with two volumes, but one string, so I guess you could make a case for either?  I'll stop thinking.

TomSweeneyRedHat:
One possible small change, and a fence sitting question.  Otherwise LGTM.

cmd/buildah/images.go
rhatdan: Fixed

cmd/buildah/run.go
rhatdan: I think we should leave this.

rhatdan:
@TomSweeneyRedHat Fixed.

rhatdan:
@mheon PTAL

cmd/buildah/bud.go
```diff
@@ -43,18 +47,15 @@ var (
 			Usage: "add global flags for the container runtime",
 		},
 		cli.StringFlag{
-			Name:  "format",
-			Usage: "`format` of the built image's manifest and metadata",
+			Name:  "signature-policy",
+			Usage: "`pathname` of signature policy file (not usually used)",
```
mheon: Suggestion: change to "pathname of signature policy file to override default (not usually used)" - makes it more apparently why it's not usually used

rhatdan: Should we just make this hidden?

cmd/buildah/bud.go
```diff
@@ -77,35 +78,14 @@ func budCmd(c *cli.Context) error {
...
-		runtime = c.String("runtime")
-	}
-
 	pullPolicy := imagebuildah.PullNever
-	if pull {
+	if c.BoolT("pull") {
 		pullPolicy = imagebuildah.PullIfMissing
 	}
-	if pullAlways {
+	if c.Bool("pull-always") {
```
mheon: Are these options exclusive? If so, you should check for that. If not, could you add to the description of `pull-always` that it overrides `pull` if both are specified?

rhatdan: No pull means pull if it is not there,  pull-always means pull even if it is there, I believe.

cmd/buildah/bud.go
rhatdan: @nalind WDYT?

cmd/buildah/commit.go
```diff
@@ -29,31 +25,36 @@ var (
...
-			Name:  "quiet, q",
-			Usage: "don't output progress information when writing images",
-		},
-		cli.BoolFlag{
 			Name:  "rm",
 			Usage: "remove the container and its content after committing it to an image. Default leaves the container and its content in place.",
 		},
+		cli.StringFlag{
+			Name:  "signature-policy",
+			Usage: "`pathname` of signature policy file (not usually used)",
```
mheon: Whatever happens to the signature policy flag above should also be applied here

cmd/buildah/config.go
```diff
@@ -18,42 +18,50 @@ const (
 
 var (
 	configFlags = []cli.Flag{
+		cli.StringSliceFlag{
+			Name:  "annotation, a",
+			Usage: "`annotation` e.g. annotation=value, for the target image",
```
mheon: Nit: would make more sense as `"add annotation e.g. annotation=value to target image"`

cmd/buildah/config.go
```diff
@@ -18,42 +18,50 @@ const (
...
+			Name:  "arch",
+			Usage: "`architecture` of the target image",
+		},
 		cli.StringFlag{
 			Name:  "author",
 			Usage: "image author contact `information`",
 		},
 		cli.StringFlag{
+			Name:  "cmd",
+			Usage: "`command` for containers based on image",
```
mheon: This isn't really clear at all. Suggest "set command to run as entrypoint for containers using this image" or similar

cmd/buildah/bud.go
TomSweeneyRedHat: 👍

cmd/buildah/run.go
TomSweeneyRedHat: 👍

cmd/buildah/images.go
```diff
@@ -42,32 +42,32 @@ type filterParams struct {
...
-			Usage: "display only image IDs",
+			Name:  "digests",
+			Usage: "show digests",
 		},
-		cli.BoolFlag{
-			Name:  "noheading, n",
-			Usage: "do not print column headings",
+		cli.StringFlag{
+			Name:  "filter, f",
+			Usage: "filter output based on conditions provided (default [])",
```
mheon: Maybe "default none" instead of "default []"

mheon:
As someone who hasn't used buildah much, I have some issues with some of the argument usage strings, which don't really explain much to me about how they're used. Code looks like a massive improvement, though.

cmd/buildah/bud.go
rhatdan: Lets handle signature-policy in a different PR.

cmd/buildah/bud.go
rhatdan: Updated to say that it overrides pull.

cmd/buildah/config.go
rhatdan: Add "add" or "set" on each one of these.

cmd/buildah/config.go
rhatdan: Modified to talk about setting default command.

cmd/buildah/images.go
rhatdan: Removed (default []).
I think users will understand without a filter they will get everything.

cmd/buildah/bud.go
TomSweeneyRedHat: @rhatdan  👍 to doing the signature policy in a different PR.  Personally I'd vote to leave it as is, but if we decide to change it there are a number of places that need changing.  The hard part is we really don't want to encourage people to mess with this param outside of testing, but I don't think we want to hide it entirely either.

nalind: I think this is documented in the man page, so a brief summary is the best we can do here.

rhatdan: Right I am suggesting that we hide it and remove the description in the man page.  I don't see why we want to support this.  At least  until customers ask for it.

rhatdan:
@mheon @TomSweeneyRedHat PTAL, 
Lets keep signaturepolicy confersation separate from this patch, I will open a new PR as soon as this gets merged.

mheon:
LGTM. Thanks for cleaning up the usage text @rhatdan - it's a lot more clear now than it was previously

cmd/buildah/bud.go
TomSweeneyRedHat: The one case that I've needed it so far is for running a Buildah container that is itself running a Docker registry inside of Travis CI.  But then again I've not gotten that up and running completely.  I don't know if there are or are not other likely use cases for an end user.

cmd/buildah/config.go
```diff
@@ -18,58 +18,58 @@ const (
...
 		cli.StringFlag{
-			Name:  "cmd",
-			Usage: "`command` for containers based on image",
+			Name:  "user, u",
+			Usage: "set `user` to run containers based on image as",
 		},
 		cli.StringSliceFlag{
 			Name:  "volume, v",
-			Usage: "`volume` to create for containers based on image",
+			Usage: "add `volume` to create for containers based on image",
```
TomSweeneyRedHat: I'd personally vote to leave this as is, or at least not use 'add'.  Perhaps 'set'  if you feel the need?  The manpage has: 
```
--volume volume

Specifies a location in the directory tree which should be marked as a volume in any images which will be built using the specified container.
```
Which I'm not crazy about, but we're not really adding or creating,

cmd/buildah/config.go
```diff
@@ -18,58 +18,58 @@ const (
...
+		},
+		cli.StringSliceFlag{
+			Name:  "port, p",
+			Usage: "add `port` to expose when running containers based on image",
 		},
 		cli.StringFlag{
-			Name:  "cmd",
-			Usage: "`command` for containers based on image",
+			Name:  "user, u",
+			Usage: "set `user` to run containers based on image as",
```
TomSweeneyRedHat: "based on the the image as" perhaps?

cmd/buildah/from.go
```diff
@@ -21,30 +31,20 @@ var (
 		},
 		cli.BoolFlag{
 			Name:  "pull-always",
-			Usage: "pull the image even if one with the same name is already present",
+			Usage: "pull the image even if one with the same name is already present (supercedes pull option)",
```
TomSweeneyRedHat: Thought these lines were supposed to be < 80 char.  The 80th char is the first 'l' in pull.  I'm personally fine with out adding this verbiage.  But if you think it's important to have that in, perhaps:
"pull the image even if the image is already present (supersedes pull option)"

Either way, please touch up supercedes to supersedes.

cmd/buildah/config.go
rhatdan: My idea was to use "add" when you can call it more then once, and `set` when you calling it once is all that makes sense.  Since you can have multiple volumes, I do add, but if you are useing working_dir, it can only be `set`.  But we are bike shedding.

TomSweeneyRedHat: Your call.  Truth be told I was back/forth on this as well and thought set was the better option.

cmd/buildah/config.go
rhatdan: Usage: "set default `user` to run inside containers based on image",
Fixed

cmd/buildah/config.go
rhatdan: How about?
			Usage: "add builtin `volume` path to be created for containers based on image",

TomSweeneyRedHat: I like!

cmd/buildah/from.go
rhatdan: Reworded to fix within 80 chars.
			Usage: "pull the image even if named image is present in store (supercedes pull option)",

rhatdan:
I do like when you tpyo typo :^)

cmd/buildah/from.go
TomSweeneyRedHat: supersedes not supercedes right?  The new verbiage looks good to me otherwise.  FWIW. supercede is a go to typo/mis-spelling for me too.

rhatdan:
https://answers.yahoo.com/question/index;_ylt=A0LEViR6UMpZMc8Ax1UnnIlQ;_ylu=X3oDMTBydWNmY2MwBGNvbG8DYmYxBHBvcwM0BHZ0aWQDBHNlYwNzcg--?qid=20081121134801AAkf5YF&p=supercedes

 they are both right

"Supercede has occurred as a spelling variant of supersede since the 17th century, and it is common in current published writing. It continues, however, to be widely regarded as an error."

" Supercede comes from the same Latin root as intercede, cessation etc
SuperSede is a new dumb Americanism
Who prints the dictionaries these days? Check an English dictionary!!"

rhatdan:
http://www.dictionary.com/browse/supercede?s=t

cmd/buildah/config.go
```diff
@@ -18,58 +18,58 @@ const (
...
-			Usage: "`description` of how the image was created",
-			Value: DefaultCreatedBy,
+			Name:  "author",
+			Usage: "set image author contact `information`",
 		},
 		cli.StringFlag{
-			Name:  "arch",
-			Usage: "`architecture` of the target image",
+			Name:  "cmd",
+			Usage: "Sets the default `command` to run for containers based on the image",
```
TomSweeneyRedHat: ugh, hate to catch this now, but "sets the default"  lower case 's'

rhatdan: Fixed

rhatdan:
@TomSweeneyRedHat Also changed to supersedes, since I had to repush.

TomSweeneyRedHat:
LGTM, one small casing issue that can be fixed later if you want

TomSweeneyRedHat:
oops didn't see your latest comment.   LGTM, ship the bad boy

rhatdan:
@rh-atomic-bot r-@TomSweeneyRedHat 90630a5

rhatdan:
@rh-atomic-bot -r@TomSweeneyRedHat 90630a5

rhatdan:
@rh-atomic-bot r@TomSweeneyRedHat 90630a5

TomSweeneyRedHat:
I'll bite, what's the last comment supposed to do?  The changes called out in the  90630a5  link LGTM fwiw.  If you need me to push other buttons, please let me know.

jlebon:
Heh, yeah it's tricky. The syntax here is: `@rh-atomic-bot r=<username_without_leading_@> [optional SHA]`. Note that the `[optional SHA]` has to be just that, not a link to a GitHub commit. See https://homu-projectatomic-ci-infra.e8ca.engint.openshiftapps.com/ for full documentation.

PR #269, opened 2017-09-22
Update buildah spec file to match new version

rhatdan:
When I bumped the version, I forgot to update the spec file.


rhatdan:
@TomSweeneyRedHat This should fix the version problem, I think.

rhatdan:
@rh-atomic-bot r+ a58cafb

PR #270, opened 2017-09-27
Cleanup

rhatdan:
Only print logrus if in debug mode 

 Validate options

If an string option is passed in, and it is not followed by a value,
then error out with a message saying the option requires a value.

For example

buildah from --creds --pull dan
option --creds requires a value

### DIFF
--- a/cmd/buildah/common.go
+++ b/cmd/buildah/common.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"os"
+	"reflect"
 	"strings"
 	"syscall"
 	"time"
@@ -157,4 +158,35 @@ func getDockerAuth(creds string) (*types.DockerAuthConfig, error) {
 		Username: username,
 		Password: password,
 	}, nil
+}
+
+// validateFlags searches for StringFlags or StringSlice flags that never had
+// a value set.  This commonly occurs when the CLI mistakently takes the next
+// option and uses it as a value.
+func validateFlags(c *cli.Context, flags []cli.Flag) error {
+	for _, flag := range flags {
+		switch reflect.TypeOf(flag).String() {
+		case "cli.StringSliceFlag":
+			{
+				f := flag.(cli.StringSliceFlag)
+				name := strings.Split(f.Name, ",")
+				val := c.StringSlice(name[0])
+				for _, v := range val {
+					if v != "" && v[0] == '-' {
+						return errors.Errorf("option --%s requires a value", name[0])
+					}
+				}
+			}
+		case "cli.StringFlag":
+			{
+				f := flag.(cli.StringFlag)
+				name := strings.Split(f.Name, ",")
+				val := c.String(name[0])
+				if val != "" && val[0] == '-' {
+					return errors.Errorf("option --%s requires a value", name[0])
+				}
+			}
+		}
+	}
+	return nil
 }
--- a/cmd/buildah/main.go
+++ b/cmd/buildah/main.go
@@ -13,6 +13,8 @@ import (
 )
 
 func main() {
+	debug := false
+
 	var defaultStoreDriverOptions *cli.StringSlice
 	if buildah.InitReexec() {
 		return
@@ -55,6 +57,7 @@ func main() {
 	app.Before = func(c *cli.Context) error {
 		logrus.SetLevel(logrus.ErrorLevel)
 		if c.GlobalBool("debug") {
+			debug = true
 			logrus.SetLevel(logrus.DebugLevel)
 		}
 		return nil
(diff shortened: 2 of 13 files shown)


rhatdan:
Fixes https://trello.com/c/NLJtPsvv/607-1-validate-and-fix-buildah-params-that-take-values?menu=filter&filter=targeted-rh

rhatdan:
Probably want the same patches for kpod.

cmd/buildah/main.go
```diff
@@ -90,7 +93,11 @@ func main() {
...
 	err := app.Run(os.Args)
 	if err != nil {
-		logrus.Errorf("%v", err)
-		os.Exit(1)
+		if debug {
+			logrus.Errorf(err.Error())
+		} else {
+			fmt.Fprintln(os.Stderr, err.Error())
+		}
+		cli.OsExiter(1)
```
TomSweeneyRedHat: I might be losing my marbles, but wasn't this change already made?

cmd/buildah/common.go
```diff
@@ -158,3 +159,17 @@ func getDockerAuth(creds string) (*types.DockerAuthConfig, error) {
 		Password: password,
 	}, nil
 }
+
+func validateFlags(c *cli.Context, flags []cli.Flag) error {
+	for _, flag := range flags {
+		if reflect.TypeOf(flag).String() == "cli.StringFlag" {
```
TomSweeneyRedHat: do we need to check booleans too?

cmd/buildah/common.go
```diff
@@ -158,3 +159,17 @@ func getDockerAuth(creds string) (*types.DockerAuthConfig, error) {
...
+			f := flag.(cli.StringFlag)
```
TomSweeneyRedHat: maybe it's late and I'm being confused, but I'm not seeing a check for the actual value?  It looks like you're checking the name and failing if it begins with a dash.  Can you tell me what value is expected in 'f.Name' and 'name' if it checks 'buildah pull' with these params "--creds=myusername:mypassword" and  "--cert-dir ~/auth"?

cmd/buildah/common.go
rhatdan: I don't think so. Booleans are not set that way.  --bool --foobar will not accidently set the bool to --foobar.

cmd/buildah/common.go
rhatdan: That type of check would happen per option. I am just looking for the case where a user forgot to enter a value fro a string field.

--format --foobar  
Should fail since --format was expecting a value but instead the CLI handed it a "option".

cmd/buildah/main.go
rhatdan: It was made in kpod but not in buildah.

rhatdan:
@mheon @nalind PTAL

rhatdan:
@umohnani8 PTAL

umohnani8:
LGTM

cmd/buildah/common.go
```diff
@@ -158,3 +159,17 @@ func getDockerAuth(creds string) (*types.DockerAuthConfig, error) {
 		Password: password,
 	}, nil
 }
+
+func validateFlags(c *cli.Context, flags []cli.Flag) error {
```
mheon: I think this definitely needs a comment explaining intent. It took me a good 5 minutes and the comments on this pull request to understand what you were doing here.

cmd/buildah/common.go
mheon: No checks for string slice flags, though? We do have a lot of those.

cmd/buildah/common.go
TomSweeneyRedHat: 👍 @mheon comment on comment.

I played around with it a bunch this morning.  This does catch corner cases only from what I could see.  Most of the error checking for the params would pick things off after it would  go through here.  One example that I ran was 
```
buildah inspect --type --type image george
```
The only downside is it goes through this code every single invocation and will probably only catch an error less than 1% of the time.  Without this change in place the above command told me I'd "too many argument specified".  All that said, it is nice when it catches a corner case and the overhead is minimal. 

The one boolean flag that I was thinking about was tls-verify which takes a boolean value.  But if one is not specified, it defaults to True, so I don't think we need to check that.

I'll say LGTM once Matt's comments are addressed.

rhatdan: Added comment.

cmd/buildah/common.go
rhatdan: I added the StringSliceFlags check also.  I originally thought it would not happen but if you do 

buildah config --env --env foo=bar containerid 

It will give you a wacky error.  Fixed now.

rhatdan:
@mheon PTAL

mheon:
I'm slightly concerned we might someday have someone try and pass a legitimate argument that begins with a hyphen. But I can't think of any cases where this would actually be something people do. I suppose we can handle it if someone actually files a bug report about it.

LGTM.

TomSweeneyRedHat:
LGTM, ship it!

rhatdan:
@rh-atomic-bot rTomSweeneyRedHat cdc5b95

rhatdan:
@rh-atomic-bot r+ cdc5b95

PR #278, opened 2017-10-02
use Makefile var for go compiler

lsm5:
This will allow compilation with a custom go binary,
for example /usr/lib/go-1.8/bin/go instead of /usr/bin/go on Ubuntu
16.04 which is still version 1.6

@nalind @rhatdan @TomSweeneyRedHat PTAL .. similar for skopeo was merged earlier https://github.com/projectatomic/skopeo/pull/424


nalind:
There's a call in the `runc` target which this should probably also change.

lsm5:
ah yes, thanks, pushing a new one..

lsm5:
done

rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #284, opened 2017-10-06
Fix validateOptions to match kpod

rhatdan:
This patch will allow for the possibility of a valid "-"
option value.  This is often used for taking from STDIN,
and should future proof us from this breaking if ever added.


rhatdan:

rhatdan:
Merging.

PR #294, opened 2017-10-20
Add proper SELinux labeling to buildah run

rhatdan:


rhatdan:

new.go
```diff
@@ -25,8 +27,36 @@ const (
...
+		if err != nil {
+			return err
+		}
+
+		for _, c := range containers {
+			if id == c.ID {
+				continue
+			} else {
+				b, err := OpenBuilder(store, c.ID)
+				if err != nil {
```
TomSweeneyRedHat: If we can't get the container, do we need to adjust the values in containers somehow, or is that likely happening when we get this error?

new.go
```diff
@@ -25,8 +27,36 @@ const (
...
+				}
+				// Prevent containers from using same MCS Label
+				if err := label.ReserveLabel(b.ProcessLabel); err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
```
TomSweeneyRedHat: Ok, maybe I've not had enough tea, but I'm struggling with this function.  Shouldn't we also be checking the ProcessLabel for 'id' rather than just continuing?  Is mcsList in selinux.go something that we can persist, or is it a map that we have to build every time we create a container?

rhatdan: The idea is to go through all of the defined containers and reserve their MCS Labels.  Then when you create a new container we get a fresh unigue MCS Label.  I don't think persisting this is a great idea.  Eventually we need to coordinate with kpod/crio/buildah.

new.go
rhatdan: I have seen this error, when I corrupted the database.  Perhaps this should be a warning, and it should be easier to cleanup.

new.go
TomSweeneyRedHat: If @nalind is hip with this, then I'm fine.  Maybe I'm just terribly misreading the code, but aren't we going through this code and touching every container every time we create a new container?  If so, it seems like a lot of overhead, but I'm not sure how to make it better.  That's the basis for my concern here fwiw.

rhatdan: Yes you are right,  and we will have to do this for CRI-O also and kpod run (We do this on Docker now also.).  Unless we eventually move to some light weight database,  or start persisting a store of allocated MCS Labels, we will have to do this.

TomSweeneyRedHat: Ok, thanks this might be the one place a BFD might be handy.  LW db sounds like an interesting side project, I'll try poking around at that.

rhatdan:
@nalind What do you think of committing this, so that @umohnani8 can move forward with here secrets patch.

rhatdan:
@nalind Could you take a look at this PR?

new.go
```diff
@@ -25,8 +27,36 @@ const (
 	DefaultTransport = "docker://"
 )
 
+func reserveSELinuxLabels(store storage.Store, id string) error {
+
```
nalind: Extra blank line?

run.go
```diff
@@ -182,7 +182,9 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	} else if b.Hostname() != "" {
 		g.SetHostname(b.Hostname())
 	}
-	mountPoint, err := b.Mount("")
+	g.SetProcessSelinuxLabel(b.ProcessLabel)
+	g.SetLinuxMountLabel(b.MountLabel)
+	mountPoint, err := b.Mount(b.MountLabel)
```
nalind: The label we determine in `reserveSELinuxLabels()` isn't recorded except in buildah's own bookkeeping data, so if another tool has already mounted the filesystem, it may have a different context.  Can we detect this?

tests/selinux.bats
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bats
+
+load helpers
+
+@test "selinux test" {
+  if ! which selinuxenabled ; then
+    skip
+   elif ! /usr/sbin/selinuxenabled; then
+     skip
```
nalind: These two lines are indented inconsistently.

tests/selinux.bats
```diff
@@ -0,0 +1,26 @@
...
+  fi
+  image=alpine
+  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json $image)
+  firstlabel=$(buildah --debug=false run $cid cat /proc/1/attr/current)
```
nalind: I don't think this `--debug=false` is necessary here.

tests/selinux.bats
```diff
@@ -0,0 +1,26 @@
...
+  run buildah --debug=false run $cid cat /proc/1/attr/current
+  [ "$status" -eq 0 ]
+  [ "$output" -eq $firstlabel ]
```
nalind: `test -eq` is used for numeric comparisons; you need `=` here.

tests/selinux.bats
```diff
@@ -0,0 +1,26 @@
...
+
+  cid1=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json $image)
+  run buildah --debug=false run $cid1 cat /proc/1/attr/current
+  [ "$status" -eq 0 ]
+  [ "$output" -ne $firstlabel ]
```
nalind: `test -ne` is used for numeric comparisons; you need `!=` here.

tests/selinux.bats
```diff
@@ -0,0 +1,26 @@
...
+
+  buildah rm $cid
+  buildah rm $cid1
+}
+
```
nalind: The test looks mostly alright (don't be afraid to add comments!), but fails when I run it locally, I expect because the "vfs" driver doesn't support mount labels.  The test is skipped in Travis and in Red Hat CI.

nalind: Not verified: would adding "libselinux-utils" or "/usr/sbin/selinuxenabled" to the list of packages to install in `.papr.sh` help?

new.go
rhatdan: Fixed

run.go
rhatdan: The kernel will refuse to mount the same file system with a different mountlabel.

tests/selinux.bats
rhatdan: Fixed

tests/selinux.bats
rhatdan: Fixed

tests/selinux.bats
rhatdan: Fixed

tests/selinux.bats
rhatdan: Wonder why the tests don't fail.

nalind: The CI environments are skipping this test.

run.go
nalind: What should we do when that happens?

tests/selinux.bats
rhatdan: I will try.

run.go
rhatdan: Error out.  Not much you can do but report the error.  We eventually need to get kpod/buildah/cri-o to all agree on the reserved labels to try to prevent it.  

BTW If this is an OverlayMount the label should be unique for the overlay.  you should be able to share the lower, but the upper needs to be unique.

nalind:
We might need to up the --deadline value used in `tests/validate/gometalinter.sh` a bit.

rhatdan:
@nalind passed the tests with the increase of timeout.

nalind:
CI environments are still skipping the test in selinux.bats.

tests/selinux.bats
nalind: Might be useful to add a message to `skip` (e.g. `skip "couldn't find selinuxenabled"`) indicating why the test is being skipped.

tests/selinux.bats
rhatdan: Add libselinux-utils

rhatdan:
Ok the problem is not the VM's, the problem is we are running the tests in a docker container and libselinux tells tools in a container that SELinux is disabled.  I will attempt to get the docker command to not fool libselinux.

rhatdan:
Adding the following to the docker run command gets libselinux to understand that SELinux is enabled.  
-v /sys/fs/selinux/:/sys/fs/selinux -v /etc/selinux:/etc/selinux

Now we will see what this breaks.

rhatdan:
Well this is not going to work.  I think it is probably best if we move the packages out of the container to run the SELinux tests.

The problem with turning on SELinux inside the container is rpm will start labeling the content and this is causing it to blow up.

nalind:
@rh-atomic-bot r+

PR #295, opened 2017-10-20
Add secrets patch to buildah

umohnani8:

### DIFF
--- a/docs/buildah.md
+++ b/docs/buildah.md
@@ -22,6 +22,10 @@ The buildah package provides a command line tool which can be used to:
 
 Print debugging information
 
+**--default-mounts-file**
+
+path to default mounts file (default path: "usr/share/containers/mounts.conf")
+
 **--help, -h**
 
 Show help
--- a/run.go
+++ b/run.go
@@ -100,6 +100,26 @@ func (b *Builder) setupMounts(mountPoint string, spec *specs.Spec, optionMounts 
 			Options:     []string{"rbind", "ro"},
 		})
 	}
+
+	cdir, err := b.store.ContainerDirectory(b.ContainerID)
+	if err != nil {
+		return errors.Wrapf(err, "error determining work directory for container %q", b.ContainerID)
+	}
+
+	// Add secrets mounts
+	mountsFiles := []string{OverrideMountsFile, b.DefaultMountsFilePath}
+	for _, file := range mountsFiles {
+		secretMounts, err := secretMounts(file, "", cdir)
+		if err != nil {
+			logrus.Warn("error mounting secrets, skipping...")
+		}
+		for _, mount := range secretMounts {
+			if haveMount(mount.Destination) {
+				continue
+			}
+			mounts = append(mounts, mount)
+		}
+	}
 	// Add temporary copies of the contents of volume locations at the
 	// volume locations, unless we already have something there.
 	for _, volume := range volumes {
@@ -107,13 +127,8 @@ func (b *Builder) setupMounts(mountPoint string, spec *specs.Spec, optionMounts 
 			// Already mounting something there, no need to bother.
 			continue
 		}
-		cdir, err := b.store.ContainerDirectory(b.ContainerID)
-		if err != nil {
-			return errors.Wrapf(err, "error determining work directory for container %q", b.ContainerID)
-		}
 		subdir := digest.Canonical.FromString(volume).Hex()
 		volumePath := filepath.Join(cdir, "buildah-volumes", subdir)
-		logrus.Debugf("using %q for volume at %q", volumePath, volume)
 		// If we need to, initialize the volume path's initial contents.
 		if _, err = os.Stat(volumePath); os.IsNotExist(err) {
 			if err = os.MkdirAll(volumePath, 0755); err != nil {
--- a/secrets.go
+++ b/secrets.go
@@ -1,0 +1,196 @@
+package buildah
+
+import (
+	"bufio"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/selinux/go-selinux/label"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	// DefaultMountsFile holds the default mount paths in the form "host_path:container_path"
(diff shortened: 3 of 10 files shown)


umohnani8:
Everything works except the relabelling. The mount is unable to get the correct label type. I am passing in "" to the label.Relabel function for the value of mountLabel. Cannot find where exactly the mountLabel is generated.
@nalind suggested I test the sha256sum of the files mounted in the container and on the host. If they are equal, it should work fine even if the relabeling doesn't work.
@rhatdan any ideas?
Tested in a RHEL box also - works, except the relabelling.

rhatdan:
https://github.com/projectatomic/buildah/pull/294 should help with the relabeling, once it gets merged, you will be able to use b.MountLabel as the label.

rhatdan:
Fix the int issues.

cmd/buildah/config.go
```diff
@@ -71,6 +71,11 @@ var (
 			Name:  "workingdir",
 			Usage: "set working `directory` for containers based on image",
 		},
+		cli.StringFlag{
```
rhatdan: This flag should be for the buildah command not for the config command, I believe.

rhatdan: Should not be hidden, but should be defaulted.  /usr/share/containers/

run.go
```diff
@@ -66,7 +66,7 @@ type RunOptions struct {
 	Terminal int
 }
 
-func (b *Builder) setupMounts(mountPoint string, spec *specs.Spec, optionMounts []specs.Mount, bindFiles, volumes []string) error {
+func (b *Builder) setupMounts(mountPoint, mountLabel string, spec *specs.Spec, optionMounts []specs.Mount, bindFiles, volumes []string) error {
```
rhatdan: mountLabel will be part of the Builder struct.

tests/secrets.bats
```diff
@@ -0,0 +1,36 @@
...
+    echo "$output"
+	[ "$status" -eq 0 ]
+	mounts="$output"
+    grep "secrets" <<< "$mounts"
+     echo "$output"
+    [ "$status" -eq 0 ]
+	buildah rm $cid
+    rm -rf $TESTSDIR/containers
+    rm -rf $TESTSDIR/rhel
+}
```
rhatdan: new line

tests/secrets.bats
```diff
@@ -0,0 +1,36 @@
...
+	runc --version
+	cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+	buildah config "$cid" --default-mounts-file "$MOUNTS_PATH"
+	run buildah --debug=false run $cid ls /run
+    echo "$output"
+	[ "$status" -eq 0 ]
+	mounts="$output"
+    grep "secrets" <<< "$mounts"
+     echo "$output"
+    [ "$status" -eq 0 ]
```
nalind: The values of `$output` and `$status` are only changed by `run`, which hasn't been called since the last time they were printed and checked.  Without the `run` error, this test will fail if `grep` doesn't find a match, since it will exit with non-zero status in that case.

tests/secrets.bats
```diff
@@ -0,0 +1,36 @@
...
+    touch $TESTSDIR/rhel/secrets/test.txt
+    echo "Testing secrets mounts. I am mounted!" > $TESTSDIR/rhel/secrets/test.txt
+}
+
+@test "bind secrets mounts to container" {
+    if ! which runc ; then
+		skip
+	fi
+	runc --version
+	cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
```
nalind: The `bulidah` helper function that runs the `buildah` binary runs it with debugging enabled, so you probably want to add `--debug=false` to undo that, when the output is being captured for use this way.

cmd/buildah/config.go
umohnani8: That is what I thought, but I cannot find like a common config struct for buildah - I might be blind.
The only global options are that of the store, which come from storage.
That is why I made the flag for config.
Any ideas on how I should approach this?

secrets.go
```diff
@@ -0,0 +1,188 @@
...
+
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/selinux/go-selinux/label"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	DefaultMountsFile  = "/usr/share/containers/mounts.conf"
+	OverrideMountsFile = "/etc/conainers/mounts.conf"
```
nalind: Typo: `conainers`

cmd/buildah/config.go
nalind: Global options are in `app.Flags` in `cmd/buildah/main.go`, though I could see this being an option that only matters to `from` and `run`.

umohnani8: fixed

tests/secrets.bats
umohnani8: fixed.

tests/secrets.bats
umohnani8: fixed.

tests/secrets.bats
umohnani8: fixed.

umohnani8:
waiting on https://github.com/projectatomic/buildah/pull/294/commits for proper labelling.

docs/buildah.md
```diff
@@ -22,6 +22,10 @@ The buildah package provides a command line tool which can be used to:
 
 Print debugging information
 
+**--default-mounts-file**
+
+path to default mounts file (default path: "usr/share/containers/mounts.conf")
```
nalind: The path in the help message is missing the initial `/` that the definition in `secrets.go` includes.

secrets.go
```diff
@@ -0,0 +1,196 @@
...
+	"github.com/opencontainers/selinux/go-selinux/label"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	// DefaultMountsFile holds the default mount paths in the form "host_path:container_path"
+	DefaultMountsFile = "/usr/share/containers/mounts.conf"
+	// OverrideMountsFile holds the default mount paths in the form "host_path:container_path" overriden by the user
+	OverrideMountsFile = "/etc/conainers/mounts.conf"
```
nalind: Typo: `conainers`

rhatdan:
https://github.com/projectatomic/buildah/pull/315 replaces this patch.

Issue #292, opened 2017-10-19
Cannot pull parent image from insecure registry


lcarva:
```
$ sudo /usr/local/bin/buildah build-using-dockerfile .
STEP 1: FROM my-insecure-registry.example.com:8888/rhel7:7-released
error building: error creating build container: error pulling image "my-insecure-registry.example.com:8888/rhel7:7-released": Error initializing image from source docker://my-insecure-registry.example.com:8888/rhel7:7-released: pinging docker registry returned: Get https://my-insecure-registry.example.com:8888/v2/: http: server gave HTTP response to HTTPS client
```

Pulling of the image via docker daemon is fine:

```
$ sudo docker pull my-insecure-registry.example.com:8888/rhel7:7-released
Trying to pull repository my-insecure-registry.example.com:8888/rhel7 ... 
sha256:87046178b950978a423047d58c11f822f8b0b3c30adf9441d42c779701a8a09f: Pulling from my-insecure-registry.example.com:8888/rhel7
26e5ed6899db: Already exists 
66dbe984a319: Already exists 
Digest: sha256:87046178b950978a423047d58c11f822f8b0b3c30adf9441d42c779701a8a09f
Status: Image is up to date for my-insecure-registry.example.com:8888/rhel7:7-released
```

Note: Output has been modified to rename internal server name to `my-insecure-registry.example.com:8888`.

TomSweeneyRedHat:
As a test/workaround.  Can you try:
```
buildah from --tls-verify=false my-insecure-registry.example.com:8888/rhel7:7-released
```
Then try the builldah bud again.  Hopefully it will pick up the image that you downloaded in the from.  If that works, then we probably have a tls-verification change that we need to do for the build-using-dockerfile command.

TomSweeneyRedHat:
@lcarva Sorry! Forgot to include your name in the last one, and more importantly forgot to thank you for the report.

lcarva:
That fixes it!

```
$ sudo /usr/local/bin/buildah from --tls-verify=false my-insecure-registry.example.com:8888/rhel7:7-released
Getting image source signatures
Copying blob sha256:26e5ed6899dbf4b1e93e0898255e8aaf43465cecd3a24910f26edb5d43dafa3c
 71.40 MiB / 71.40 MiB [====================================================] 9s
Copying blob sha256:66dbe984a319ca6d40dc10c2c561821128a0bd8967e0cbd8cc2a302736041ffb
[...]

PR #297, opened 2017-10-21
Add tls-verify to bud command

TomSweeneyRedHat:

Adds the --tls-verify flag to the Buildah bud command.

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -54,6 +54,10 @@ var (
 			Name:  "tag, t",
 			Usage: "`tag` to apply to the built image",
 		},
+		cli.BoolTFlag{
+			Name:  "tls-verify",
+			Usage: "Require HTTPS and verify certificates when accessing the registry",
+		},
 	}
 
 	budDescription = "Builds an OCI image using instructions in one or more Dockerfiles."
@@ -179,6 +183,7 @@ func budCmd(c *cli.Context) error {
 		Compression:         imagebuildah.Gzip,
 		Quiet:               c.Bool("quiet"),
 		SignaturePolicyPath: c.String("signature-policy"),
+		SkipTLSVerify:       !c.Bool("tls-verify"),
 		Args:                args,
 		Output:              output,
 		AdditionalTags:      tags,
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -95,6 +95,8 @@ type BuildOptions struct {
 	// specified, indicating that the shared, system-wide default policy
 	// should be used.
 	SignaturePolicyPath string
+	// SkipTLSVerify denotes whether TLS verification should not be used.
+	SkipTLSVerify bool
 	// ReportWriter is an io.Writer which will be used to report the
 	// progress of the (possible) pulling of the source image and the
 	// writing of the new image.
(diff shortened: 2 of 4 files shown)


TomSweeneyRedHat:
Fixes  #292

imagebuildah/build.go
```diff
@@ -136,11 +139,14 @@ type Executor struct {
 	reportWriter                   io.Writer
 }
 
-func makeSystemContext(signaturePolicyPath string) *types.SystemContext {
+func makeSystemContext(signaturePolicyPath string, tlsVerify bool) *types.SystemContext {
 	sc := &types.SystemContext{}
 	if signaturePolicyPath != "" {
 		sc.SignaturePolicyPath = signaturePolicyPath
 	}
+	if tlsVerify {
```
rhatdan: Why is this check necessary?  This says it will only modify the systemcontext if tleVerify is true?  Why not always set it below.

rhatdan:
Should we add an example to the man page.

--tls-verify false

imagebuildah/build.go
TomSweeneyRedHat: You're right, removing it.  I should not have had that there.

TomSweeneyRedHat:
@nalind PTAL

rhatdan:
@nalind Can you take a look at this so we can get it merged.

imagebuildah/build.go
```diff
@@ -95,6 +95,9 @@ type BuildOptions struct {
 	// specified, indicating that the shared, system-wide default policy
 	// should be used.
 	SignaturePolicyPath string
+	// TlsVerify denotes whether TLS verification should or should not
+	// be used.
+	TlsVerify bool
```
nalind: Please change this to `SkipTLSVerify`, to line up better with the flag names that the image library expects us to set, inverting the logic where necessary to keep the CLI the same.

nalind:
A nit on the field's naming, otherwise LGTM.

TomSweeneyRedHat:
Green lights and have adjusted per @nalind's comments.

cmd/buildah/bud.go
```diff
@@ -179,6 +183,7 @@ func budCmd(c *cli.Context) error {
 		Compression:         imagebuildah.Gzip,
 		Quiet:               c.Bool("quiet"),
 		SignaturePolicyPath: c.String("signature-policy"),
+		SkipTLSVerify:       c.Bool("tls-verify"),
```
rhatdan: Since tls-verify is defaulted to true, shouldn't this be 
SkipTLSVerify:  ! c.Bool("tls-verify")

imagebuildah/build.go
```diff
@@ -95,6 +95,9 @@ type BuildOptions struct {
 	// specified, indicating that the shared, system-wide default policy
 	// should be used.
 	SignaturePolicyPath string
+	// SkipTLSVerify denotes whether TLS verification should or should not
```
rhatdan: // SkipTLSVerify denotes whether TLS verification should not be used

imagebuildah/build.go
```diff
@@ -136,11 +139,12 @@ type Executor struct {
 	reportWriter                   io.Writer
 }
 
-func makeSystemContext(signaturePolicyPath string) *types.SystemContext {
+func makeSystemContext(signaturePolicyPath string, tlsVerify bool) *types.SystemContext {
```
rhatdan: Change this to the new name SkipTlsVerify and then reverse the logic

cmd/buildah/bud.go
TomSweeneyRedHat: Good point.  I just grabbed the name suggestion from Nalin, plugged it in, and didn't consider that.  Will adjust per your suggestions.

TomSweeneyRedHat: Fixed

imagebuildah/build.go
TomSweeneyRedHat: fixed

imagebuildah/build.go
TomSweeneyRedHat: fixed

rhatdan:
LGTM.  Merge once it passes tests.

TomSweeneyRedHat:
Going to close and reopen to respin the tests.  Looks like a doc build error which this shouldn't effect.

nalind:
@rh-atomic-bot r+

PR #305, opened 2017-11-01
Add go 1.9.x to .travis.yml and remove go tip build temporarily

TomSweeneyRedHat:
Signed-off-by: TomSweeneyRedHat <tsweeney@redhat.com>

Add go v1.9.x to the mix and remove go tip temporarily.  This will fix #304's build issues.

PR #306, opened 2017-11-02
Run Travis on latest released versions of go.

rhatdan:
Do not run travis builds with ostree.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 language: go
 go:
-    - 1.7
-    - 1.8
-    - tip
+    - 1.7.x
+    - 1.8.x
+    - 1.9.x
+    - Tip
 dist: trusty
 sudo: required
 services:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,8 +1,9 @@
 language: go
 go:
-    - 1.7
-    - 1.8
-    - tip
+    - 1.7.x
+    - 1.8.x
+    - 1.9.x
+    - Tip
```
nalind: `Tip` doesn't seem to be recognized as a version that `gimme` supports.

new.go
```diff
@@ -45,6 +44,7 @@ func newBuilder(store storage.Store, options BuilderOptions) (*Builder, error) {
 
 	imageID := ""
 	if image != "" {
+		var err error
```
nalind: Is this accidentally copied from #294?

TomSweeneyRedHat:
FWIW, I'm doing similar changes in travis.yml in #305.  Except there I'm trying to ignore the tip failures.

new.go
rhatdan: oops

rhatdan:
Ok lets concentrate on Tom's but I think we should stop using ostree on Ubuntu for testing.

TomSweeneyRedHat:
ok, will see how this current run goes.  Will circle back to kill ostree in ubuntu.

PR #308, opened 2017-11-02
Add go tip to build, but allow it to have failures

TomSweeneyRedHat:

Add tip back in and teach travis ci to ignore any failure while using it as that's golang's unstable and under development version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,21 @@
 language: go
-go:
+dist: trusty
+sudo: required
+go: 
     - 1.7
     - 1.8
     - 1.9.x
-dist: trusty
-sudo: required
+    - tip
+
+matrix:
+  # If the latest unstable development version of go fails, that's OK.
+  allow_failures:
+    - go: tip
+
+  # Don't hold on the tip tests to finish.  Mark tests green if the
+  # stable versions pass.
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
What a per-snickety piece of syntax that's not well documented.  @rhatdan @nalind this is ready to go.  

Open up the checks and then the Details link for Travis.  You'll see a Red X for "Go: Tip", green the rest of the way and the overall tests pass as "Go: Tip" is in the "Allowed Failures" section.  If we're still Red in a week or so, we can try to figure out the issues there.

If you'd like, I can change the other projects to do the same thing.

rhatdan:
Tip gofmt is breaking in libpod and crio also.  Nice maybe we should do this for all of the repos.

TomSweeneyRedHat:
@rhatdan The only downside to doing this here or elsewhere is you don't get the Go: tip failure blaring in your eyeballs.  You've got to click down a level or two to see the failure.  The upside is we won't have build failures as frequently and we wont' get them due to some underdevelopment go funkiness.  Personally I think this is the way to go, but up to you, @nalind and others.

rhatdan:
LGTM
I will merge since this only adds information.

PR #311, opened 2017-11-03
add new command: pull

TomasTomecek:
As you may know, I am trying to implement buildah backend for [ansible-container](https://github.com/ansible/ansible-container/pull/760).

One of the interfaces defined in a-c is [pull method](https://github.com/ansible/ansible-container/blob/ab8d4bdf6199f20969f09b42843cfb4f75ffa9c7/container/docker/engine.py#L613)
which returns ID to the pulled image. This is why I implemented dedicated pull
command for buildah. Unfortunately I wasn't able to figure out how to get this
ID/digest of the image which was just pulled. Could you please help me out?
Does it even make sense? a-c is using this digest for caching purposes -- if the
image didn't change, we don't need to rebuild.

The other issue I have actually is that one cannot do `buildah from <image-id>`.

- [ ] I also need to write a manpage.
- [ ] CI failures

### DIFF
--- a/buildah.go
+++ b/buildah.go
@@ -158,6 +158,11 @@ func NewBuilder(store storage.Store, options BuilderOptions) (*Builder, error) {
 	return newBuilder(store, options)
 }
 
+// pull selected image and return its reference
+func PullImage(store storage.Store, options BuilderOptions, sc *types.SystemContext) (types.ImageReference, error) {
+	return pullImage(store, options, sc)
+}
+
 // ImportBuilder creates a new build configuration using an already-present
 // container.
 func ImportBuilder(store storage.Store, options ImportOptions) (*Builder, error) {
--- a/cmd/buildah/main.go
+++ b/cmd/buildah/main.go
@@ -90,6 +90,7 @@ func main() {
 		tagCommand,
 		umountCommand,
 		versionCommand,
+		pullCommand,
 	}
 	err := app.Run(os.Args)
 	if err != nil {
--- a/cmd/buildah/pull.go
+++ b/cmd/buildah/pull.go
@@ -1,0 +1,95 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/containers/image/transports"
+	"github.com/pkg/errors"
+	"github.com/projectatomic/buildah"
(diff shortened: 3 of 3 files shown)


rhatdan:
Would `kpod pull` work for you?

TomasTomecek:
Just looked at dependencies and they seem to be similar for both projects. It's really up to upstream maintainers of Ansible Container whether they would be okay with adding another dependency. CC @j00bar @chouseknecht

nalind:
FWIW, image IDs will be stable once we merge #187, and it will also be able to refer to already-pulled images by ID without requiring an `@` as a prefix.

TomasTomecek:
@nalind what are your plans to have that PR merged?

nalind:
The bulk of #187 is vendoring changes which are still in progress in the image library (https://github.com/containers/image/pull/305).  Once those are sorted out there, I expect to update #187 and merge it.

TomasTomecek:
So I used `kpod pull` and it works well. I'm closing this since you don't seem to be interested in this feature.

TomasTomecek:
since this is closed and kpod turned into podman: what are the plans with including podman in fedora?

```
$ dnf install /usr/bin/podman
No match for argument: /usr/bin/podman
Error: Unable to find a match

$ dnf search libpod
No matches found.
```

TomSweeneyRedHat:
@TomasTomecek the work to get Podman include into Fedora, and for that matter other distros, is currently in progress.  No promises, but hopefully something will be out in the very near future.

baude:
@TomasTomecek -> https://koji.fedoraproject.org/koji/packageinfo?packageID=26289  process is started ...

we also have https://copr.fedorainfracloud.org/coprs/baude/Upstream_CRIO_Family/packages/ for tracking upstream builds ... podman is plugged in there and working.  As always, the Fedora builds are the *real* ones and the copr builds are convenience.

TomasTomecek:
👍 thank you gentlemen

rhatdan:
podman is actually in Fedora right now. Although we plan on changing the database backend on Monday, so you can play with this podman, but expect it to change in incompatible ways next week.  We will make more of an announcement after we swap out the database.

PR #313, opened 2017-11-07
Resolve run USER names in a chroot, change default run group to 0

nalind:
This update changes things so that we chroot() before attempting to read `passwd` or `group` files in a container's root, to try to avoid symlink traversal issues.

It also corrects the default behavior when a numeric UID is specified for a container's run, but no group information is provided, to use group 0 rather than attempting to convert the UID to a user name and primary group, or looking up the UID as if it were a name.


rhatdan:
We are going to need this for kpod run also...

user_linux_cgo.go
```diff
@@ -0,0 +1,216 @@
...
+
+	// Start the lookup at the chroot directory.
+	rootfd, err := unix.Open(rootdir, unix.O_DIRECTORY|unix.O_PATH, 0)
+	if err != nil {
+		return "", nil, err
+	}
+	defer unix.Close(rootfd)
+
+	// Start resolving the pathname.
+	components := strings.Split(filepath.Clean(filename), string(os.PathSeparator))
```
rhatdan: Why are you cleaning filename a second time?

user_linux_cgo.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	defer C.fclose(f)
+
+	lookupUser.Lock()
+	defer lookupUser.Unlock()
+
+	pwd := C.fgetpwent(f)
+	for pwd != nil {
+		if uint64(pwd.pw_uid) != userid {
+			pwd = C.fgetpwent(f)
```
rhatdan: Won't this go into in infinite loop?

user_linux_cgo.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	defer C.fclose(f)
+
+	lookupGroup.Lock()
+	defer lookupGroup.Unlock()
+
+	grp := C.fgetgrent(f)
+	for grp != nil {
+		if C.strcmp(grp.gr_name, name) != 0 {
+			grp = C.fgetgrent(f)
```
rhatdan: Infinite loop?

tests/run.bats
```diff
@@ -154,6 +156,20 @@ load helpers
...
+	buildah run -- $cid id
+	run buildah --debug=false run -- $cid id -u
+	[ "$output" = $testotheruid ]
+	run buildah --debug=false run -- $cid id -g
+	[ "$output" = $testgroupid ]
+
+	buildah config $cid -u ${testotheruid}
+	buildah run -- $cid id
+	run buildah --debug=false run -- $cid id -u
+	[ "$output" = $testotheruid ]
```
rhatdan: Should you check status as well?

user_linux_cgo.go
nalind: Yeah, that's a mistake.

user_linux_cgo.go
nalind: It shouldn't?  We either (read another entry, and either iterate or fall out of the loop) or (return).

user_linux_cgo.go
nalind: It shouldn't?  We either (read another entry, and either iterate or fall out of the loop) or (return).

rhatdan:
LGTM.

tests/run.bats
nalind: Doh.  Yeah.  Fixing.

user_linux_cgo.go
rhatdan: Ok never mind, I forgot whti fgetpwent was doing.

user_linux_cgo.go
rhatdan: Yup, me being dumb.

rhatdan:
@umohnani8 @mitrmac PTAL

TomSweeneyRedHat:
LGTM

nalind:
Rewrote the parsing logic to use less hand-rolled code and take advantage of bufio.

nalind:
Simplified further.

rhatdan:
@mheon PTAL

user_linux.go
```diff
@@ -0,0 +1,239 @@
...
+	for isPrefix {
+		// We didn't get a whole line.  Keep reading chunks until we find an end of line, and discard them.
+		for isPrefix {
+			_, isPrefix, err = rc.ReadLine()
+			if err != nil {
+				return nil
+			}
+		}
+		// That last read was the end of a line, so now we try to read the (beginning of?) the next line.
+		line, isPrefix, err = rc.ReadLine()
```
TomSweeneyRedHat: OK, it's past my "it's 8:00, don't ask questions" time, but I gotta ask.  If the first line to be read is too long for the buffer and we spin through to the end of the line, we then do another ReadLine.  Won't that first line be ignored?  I'm guessing it's rarely if ever going to be an issue for group or passwd, but I'm curious.

TomSweeneyRedHat:
LGTM, one question on ReadLine functionality which is most likely going to be educational for me.

user_linux.go
```diff
@@ -0,0 +1,239 @@
...
+	uid  uint64
+	gid  uint64
+}
+type lookupGroupEntry struct {
+	name string
+	gid  uint64
+}
+
+func parseNextPasswd(rc *bufio.Reader) *lookupPasswdEntry {
+	line, isPrefix, err := rc.ReadLine()
```
mheon: This line retrieval code looks to be shared between parseNextPasswd and parseNextGroup - can probably be moved into a function that returns a line and error that both of them call

mheon:
Code mostly looks fine, though I think I share the concerns of @TomSweeneyRedHat about potentially discarding incompletely-read lines. I feel like this might happen on occasion, considering we're reading from another process's STDOUT

user_linux.go
nalind: Yup, this is an attempt to avoid trying to parse any part of a line that we can't read in its entirety.  The older version of this function used an 8192 byte buffer, but bufio defaults to something smaller, and particularly for group entries, I'm wary of lines that are too long to fit (i.e., groups with an astonishing number of members).

TomSweeneyRedHat: OK, I'm a little weary about skipping a line.  Can/should you a debug log here to note which lines are skipped?

umohnani8:
LGTM

nalind:
Rebased, added a debug log message, and factored the ReadLine() calling into its own function.

user_linux.go
nalind: Refactored it.

rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+

nalind:
[(3) [buildah] Fix potential symlink traversal problem when looking up numeric IDs for USER settings](https://trello.com/c/Go7pxsyV/480-3-buildah-fix-potential-symlink-traversal-problem-when-looking-up-numeric-ids-for-user-settings)

PR #316, opened 2017-11-07
Add secrets patch to buildah

rhatdan:

### DIFF
--- a/secrets.go
+++ b/secrets.go
@@ -1,0 +1,198 @@
+package buildah
+
+import (
+	"bufio"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	rspec "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/selinux/go-selinux/label"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	// DefaultMountsFile holds the default mount paths in the form
+	// "host_path:container_path"
+	DefaultMountsFile = "/usr/share/containers/mounts.conf"
+	// OverrideMountsFile holds the default mount paths in the form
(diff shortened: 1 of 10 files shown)


tests/secrets.bats
```diff
@@ -0,0 +1,33 @@
...
+    echo "$output"
+	[ "$status" -eq 0 ]
+	mounts="$output"
+    run grep "secrets" <<< "$mounts"
+    echo "$output"
+    [ "$status" -eq 0 ]
+	buildah rm $cid
+    rm -rf $TESTSDIR/containers
+    rm -rf $TESTSDIR/rhel
+}
```
nalind: The indentation in this shell function mixes tabs and spaces inconsistently.

secrets.go
```diff
@@ -0,0 +1,198 @@
...
+		ctrDirOnHost := filepath.Join(containerWorkingDir, ctrDir)
+		if err = os.RemoveAll(ctrDirOnHost); err != nil {
+			return nil, fmt.Errorf("remove container directory failed: %v", err)
+		}
+
+		if err = os.MkdirAll(ctrDirOnHost, 0755); err != nil {
+			return nil, fmt.Errorf("making container directory failed: %v", err)
+		}
+
+		hostDir, err = resolveSymbolicLink(hostDir)
```
nalind: Is this step necessary?

nalind:
A couple of nits, but otherwise it looks reasonable.

secrets.go
rhatdan: I would guess this came from the original patch used for docker.

tests/secrets.bats
rhatdan: Fixed

rhatdan:
Will merge when the tests pass.  And then will build new version of buildah

PR #320, opened 2017-11-09
integration tests: default to /var/tmp

nalind:
Default to running integration tests using /var/tmp as scratch space, since it's more likely to support proper SELinux labeling than /tmp, which is more likely to be on a tmpfs.


TomSweeneyRedHat:
LGTM, assuming happy tests.  Is this something we should add to kpod and maybe others?

nalind:
bot, retest this please

nalind:
Okay, merging this, since I think it's a problem for other open PRs.
@rh-atomic-bot r+

rhatdan:
Do we have similar issues with CRI-O and libpod?

PR #321, opened 2017-11-09
WIP: Add manifest type conversion to buildah push

umohnani8:
Add manifest type conversion to buildah push with dir transport

User can select from 3 manifest types: oci, v2s1, or v2s2
e.g buildah push --format v2s2 alpine dir:my-directory

Did not add the `compress` flag because buildah push compresses by default and has the `disable-compression` flag to push the uncompressed form of the image


umohnani8:
@rhatdan @nalind PTAL

umohnani8:
https://github.com/containers/image/pull/377 needs to be merged upstream and vendored in first.

vendor.conf
```diff
@@ -1,7 +1,8 @@
 github.com/BurntSushi/toml master
 github.com/Nvveen/Gotty master
 github.com/blang/semver master
-github.com/containers/image 063852766c3e82ec8359ce5f6612e056f3efaa76
+# github.com/containers/image 063852766c3e82ec8359ce5f6612e056f3efaa76
+github.com/containers/image dir_transport https://github.com/umohnani8/image
```
TomSweeneyRedHat: Will this change before you do a final submit?

umohnani8: yes, once it gets accepted to upstream I will change the vendor.

docs/buildah-push.md
```diff
@@ -52,6 +52,10 @@ The username[:password] to use to authenticate with the registry if required.
 
 Don't compress copies of filesystem layers which will be pushed.
 
+**--format, -f**
+
+`MANIFEST TYPE` (oci, v2s1, or v2s2) to use when saving image to directory using the 'dir:' transport (default is manifest type of source)
```
TomSweeneyRedHat: May have missed this in earlier reviews, but I think this probably ought to be "Manifest type" or "Manifest Type" if you really want to make it stand out a little.

cmd/buildah/push.go
```diff
@@ -29,6 +31,10 @@ var (
 			Name:  "disable-compression, D",
 			Usage: "don't compress layers",
 		},
+		cli.StringFlag{
+			Name:  "format, f",
+			Usage: "`MANIFEST TYPE` (oci, v2s1, or v2s2) to use when saving image to directory using the 'dir:' transport (default is manifest type of source)",
```
TomSweeneyRedHat: I'd make "MANIFEST TYPE" all lower to match the other commands.  Unless @nalind has objections, I'd shorten this to:
"manifest type (oci, v2s1, or v2s2) to use when saving image"

They can dig for details in the man page if that doesn't jog their memory.

umohnani8: fixed

docs/buildah-push.md
umohnani8: fixed

rhatdan:
LGTM
I think we can merge once vendor is fixed.

cmd/buildah/push.go
```diff
@@ -104,8 +110,26 @@ func pushCmd(c *cli.Context) error {
...
+	transport := strings.SplitN(args[1], ":", 2)[0]
+	if transport == "dir" {
+		if c.IsSet("format") {
+			switch c.String("format") {
+			case "oci":
+				manifestType = imgspecv1.MediaTypeImageManifest
+			case "v2s1":
+				manifestType = manifest.DockerV2Schema1SignedMediaType
+			case "v2s2":
+				manifestType = manifest.DockerV2Schema2MediaType
```
nalind: The set of values we recognize here doesn't match the set of values for `--format` that `commit` accepts.  Can add  `docker` as an alias for `v2s2`?

tests/push.bats
```diff
@@ -18,3 +18,24 @@ load helpers
...
   done
 }
+
+@test "push with manifest type" {
+  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+  run buildah push --signature-policy ${TESTSDIR}/policy.json --format oci alpine dir:my-dir
+  echo "$output"
+  [ "$status" -eq 0 ]
+  manifest=$(cat my-dir/manifest.json)
+  run grep "application/vnd.oci.image.config.v1+json" <<< "$manifest"
```
nalind: Is there a particular reason we're using redirection here, instead of giving `grep` the name of the file to read?

cmd/buildah/push.go
umohnani8: added `docker` as an alias for `v2s2`. Fixed

tests/push.bats
umohnani8: No reason, I was just being silly. Fixed

rhatdan:
LGTM

umohnani8:
vendor updated, ready to merge once tests pass.

PR #323, opened 2017-11-12
Set option.terminal appropriately in run

TomSweeneyRedHat:

The options.terminal field was never set to DefaultTerminal.  Unless the --tty was specified, the options.terminal field defaults to "WithoutTerminal" or the --tty=false value.  It should instead use the current terminal unless --tty is specified to use a pseudo terminal (or not).

This behaviour follows the man page description.

Testing results.  Notice in the --tty=false case, there is not a prompt from the container.  Prior to this change that prompt was not present when the --tty parameter was not specified.  It is now present as illustrated in the fourth example.
```
[root@localhost buildah]# ./buildah run $container bash --tty=false
exit
[root@localhost buildah]# ./buildah run $container bash --tty=true
[root@mrsdalloway /]# exit
exit
[root@localhost buildah]# ./buildah run $container bash --tty
[root@mrsdalloway /]# exit
exit
[root@localhost buildah]# ./buildah run $container bash
[root@mrsdalloway /]# exit
```

### DIFF
--- a/cmd/buildah/run.go
+++ b/cmd/buildah/run.go
@@ -79,10 +79,12 @@ func runCmd(c *cli.Context) error {
 		Args:     c.StringSlice("runtime-flag"),
 	}
 
-	if c.Bool("tty") {
-		options.Terminal = buildah.WithTerminal
-	} else {
-		options.Terminal = buildah.WithoutTerminal
+	if c.IsSet("tty") {
+		if c.Bool("tty") {
+			options.Terminal = buildah.WithTerminal
+		} else {
+			options.Terminal = buildah.WithoutTerminal
+		}
 	}
 
 	for _, volumeSpec := range c.StringSlice("volume") {


rhatdan:
@nalind PTAL

cmd/buildah/run.go
```diff
@@ -79,10 +79,12 @@ func runCmd(c *cli.Context) error {
 		Args:     c.StringSlice("runtime-flag"),
 	}
 
-	if c.Bool("tty") {
+	options.Terminal = buildah.DefaultTerminal
```
nalind: This shouldn't be necessary, as `DefaultTerminal` is the zero value, which `options` should already include as a default.

cmd/buildah/run.go
```diff
@@ -79,10 +79,12 @@ func runCmd(c *cli.Context) error {
...
+	if c.IsSet("tty") {
```
nalind: Does this path get taken when we're passed `--tty=false`?

cmd/buildah/run.go
TomSweeneyRedHat: I'd not thought of that.  I'll remove that line, retest and will cut a new PR.

cmd/buildah/run.go
TomSweeneyRedHat: I'd not specifically watched this code, but the top level run code was only getting options.Terminal=buildah.WithoutTerminal when I set the flag to false.  I'll instrument this code during my testing to verify.

cmd/buildah/run.go
```diff
@@ -79,10 +79,11 @@ func runCmd(c *cli.Context) error {
...
 	}
 
-	if c.Bool("tty") {
+	if c.IsSet("tty") {
 		options.Terminal = buildah.WithTerminal
-	} else {
-		options.Terminal = buildah.WithoutTerminal
+		if !c.Bool("tty") {
+			options.Terminal = buildah.WithoutTerminal
+		}
```
TomSweeneyRedHat: @nalind I tested this code by adding a few println's (see below) and the second if statement here is invoked with tty==false and not in other cases.

```
        if c.IsSet("tty") {
                fmt.Println("TOM: In tty IsSet")
                options.Terminal = buildah.WithTerminal
                if !c.Bool("tty") {
                        fmt.Println("TOM: In tty not Boolean")
                        options.Terminal = buildah.WithoutTerminal
                }
        }

and saw:
[root@localhost buildah]# ./buildah run fedora-working-container /bin/bash --tty=false
TOM: In tty IsSet
TOM: In tty not Boolean
exit
[root@localhost buildah]# ./buildah run fedora-working-container /bin/bash --tty
TOM: In tty IsSet
[root@mrsdalloway /]# exit
exit
[root@localhost buildah]# ./buildah run fedora-working-container /bin/bash --tty=true
TOM: In tty IsSet
[root@mrsdalloway /]# exit
exit
[root@localhost buildah]# ./buildah run fedora-working-container /bin/bash 
[root@mrsdalloway /]# exit
exit
```

nalind:
LGTM.

cmd/buildah/run.go
```diff
@@ -79,10 +79,12 @@ func runCmd(c *cli.Context) error {
 		Args:     c.StringSlice("runtime-flag"),
 	}
 
-	if c.Bool("tty") {
+	options.Terminal = buildah.DefaultTerminal
+	if c.IsSet("tty") {
 		options.Terminal = buildah.WithTerminal
-	} else {
-		options.Terminal = buildah.WithoutTerminal
+		if !c.Bool("tty") {
```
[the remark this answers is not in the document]
TomSweeneyRedHat: fixed, new version inc.

rhatdan:
@rh-atomic-bot r+

PR #327, opened 2017-11-16
Ignore errors when trying to read containers buildah.json

rhatdan:
Since containers can be created using other tools then buildah
we can not fail when they don't have buildah config.

### DIFF
--- a/new.go
+++ b/new.go
@@ -2,6 +2,7 @@ package buildah
 
 import (
 	"fmt"
+	"os"
 	"strings"
 
 	is "github.com/containers/image/storage"
@@ -40,7 +41,9 @@ func reserveSELinuxLabels(store storage.Store, id string) error {
 			} else {
 				b, err := OpenBuilder(store, c.ID)
 				if err != nil {
-					if err == storage.ErrContainerUnknown {
+					if os.IsNotExist(err) {
+						// Ignore not exist errors since containers probably created by other tool
+						// TODO, we need to read other containers json data to reserve their SELinux labels
 						continue
 					}
 					return err


umohnani8:
LGTM

new.go
```diff
@@ -40,10 +40,9 @@ func reserveSELinuxLabels(store storage.Store, id string) error {
 			} else {
 				b, err := OpenBuilder(store, c.ID)
 				if err != nil {
-					if err == storage.ErrContainerUnknown {
-						continue
-					}
-					return err
+					// Ignore errors since containers can be created by other tools
+					// TODO, we need to read other containers json data to reserve their SELinux labels
+					continue
```
nalind: Were we getting storage.ErrContainerUnknown here?  The logic in https://github.com/projectatomic/buildah/blob/master/cmd/buildah/common.go#L48 is checking for different errors from `OpenBuilder()`.

new.go
```diff
@@ -40,10 +40,9 @@ func reserveSELinuxLabels(store storage.Store, id string) error {
 			} else {
 				b, err := OpenBuilder(store, c.ID)
 				if err != nil {
-					if err == storage.ErrContainerUnknown {
```
nalind: This error can be wrapped, so I suggest changing the test to
```
if errors.Cause(err) == storage.ErrContainerUnknown || os.IsNotExist(errors.Cause(err)) {
```
or something similar.  We probably also want to check for these at https://github.com/projectatomic/buildah/blob/master/cmd/buildah/common.go#L48.

nalind:
LGTM.

TomSweeneyRedHat:
LGTM, with happy tests.

nalind:
@rh-atomic-bot r+

PR #331, opened 2017-11-21
Mention docker login in documentation for authentication

umohnani8:
Since we fall back to reading the credentials from $HOME/.docker/config
set by docker login when kpod login doesn't have the credentials

### DIFF
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -16,8 +16,8 @@ to a temporary location.
 
 **--authfile** *path*
 
-Path of the authentication file. Default is ${XDG_RUNTIME_DIR}/containers/auth.json
-which is set using `kpod login`
+Path of the authentication file. Default is ${XDG_RUNTIME\_DIR}/containers/auth.json, which is set using `kpod login`.
+If the authorization state is not found there, $HOME/.docker/config.json is checked, which is set using `docker login`.
 
 **--build-arg** *arg=value*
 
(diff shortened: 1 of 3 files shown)


umohnani8:

rhatdan:
Add the man page reference at the bottom of the man pages for docker-login.

docs/buildah-bud.md
```diff
@@ -16,8 +16,8 @@ to a temporary location.
 
 **--authfile** *path*
 
-Path of the authentication file. Default is ${XDG_RUNTIME_DIR}/containers/auth.json
-which is set using `kpod login`
+Path of the authentication file. Default is ${XDG_RUNTIME\_DIR}/containers/auth.json, which is set using `kpod login`.
```
TomSweeneyRedHat: is the \_DIR correct?

TomSweeneyRedHat: Hmm, md ate up my slash.  Is RUNTIME(backslash)_DIR correct?  It appears to resolve ok, but I don't think the backslash is necessary

rhatdan: I agree.

TomSweeneyRedHat: Same comment in the other spots.

umohnani8: in mark down when you put a word between underscores (_word_), it gets italicized. The back slash is to escape the _, so that doesn't happen.

TomSweeneyRedHat: Gotta luv md!  Thanks for the info, something new learned today!

umohnani8:
Tests passed

rhatdan:
@rh-atomic-bot r+

PR #333, opened 2017-11-22
Need to block access to kernel files systems in /proc and /sys

rhatdan:


umohnani8:
LGTM

TomSweeneyRedHat:
Same questions as from https://github.com/projectatomic/libpod/pull/61/files, if that's all hip, then LGTM.

rhatdan:

rhatdan:
Yes if they are missing it is ok.  We should ask about the sysrq one on CRI-O or in docker, if it is also readonly there?

Issue #66, opened 2017-04-12
Potential symlink/directory traversal issue in user ID extraction


fweimer:
Commit b1bb73e01c9bf0b1b75e50a2d1947b14a8174eee (“ Teach "Run" to dig user IDs out of containers”) added the `fopenContainerFile` function. It does not correctly deal with symbolic links, which could point to something outside of the container root.

nalind:
Are you suggesting something like https://github.com/nalind/buildah-1/commit/97ca24a7b4cc2739edf8b6b12f1c87ab103a0caa?  What did we not catch with the (possibly over-restrictive) lstat-after-opening method?

nalind:
![](https://github.trello.services/images/mini-trello-icon.png) [(8) [Buildah] Add "dockah build-using-dockerfile" aka "buildah bud"](https://trello.com/c/AyBXqoRY/443-8-buildah-add-dockah-build-using-dockerfile-aka-buildah-bud)

nalind:
![](https://github.trello.services/images/mini-trello-icon.png) [(8) [Buildah] Add "dockah build-using-dockerfile" aka "buildah bud"](https://trello.com/c/AyBXqoRY/443-8-buildah-add-dockah-build-using-dockerfile-aka-buildah-bud)

nalind:
@fweimer The patch with the fix I'm looking at had a conflict in it, so I've rebased it to https://github.com/nalind/buildah-1/commit/2c387528f2af7655d231a4989f67b21c90e1c385.  I'd appreciate any guidance you can offer.

fweimer:
@nalind, when this code is run, is the file system directory tree dead and inaccessible to unprivileged processes, or has concurrent modification of the tree to be expected?

nalind:
The filesystem tree is mounted on the host, so modification is possible for users with the necessary privileges.

fweimer:
@nalind, in this case, the code still has a time-of-check-time-of-use race condition because after symbolic link resolution, the file system tree could change and contain symbolic links again.

You either have to use `chroot` to restrict pathname lookups (no longer needs root in current Fedora), or reimplement kernel pathname lookup using `openat` and `O_PATH`.

nalind:
[...]

PR #336, opened 2017-11-27
Add --chown option to add/copy

bertinatto:
Hi there,

I came across a permission error while I was automating the migration of a Redis setup into a container. Basically, `buildah copy` was copying my file into the container and leaving it owned by `root`.

While I know that an extra `buildah run containerID chown redis:redis /etc/redis.conf` would solve that issue, I would love to have a `--chown` option in `buildah add/copy`. I know that docker added something similar recently, but I haven't looked into their implementation.

This PR is simply a PoC to to get a feedback from you. It still needs some improvements, so if you think I can go ahead, I'll look into adding in some tests, docs and refactor the code.

This is an example of how it works:

```
# Create container
$ sudo ./buildah from registry.access.redhat.com/rhscl/redis-32-rhel7
redis-32-rhel7-working-container-15

# Copy some data into the container, setting the owner
$ sudo ./buildah copy --chown redis:redis redis-32-rhel7-working-container-15 contrib /data/contrib
$ sudo ./buildah copy --chown redis:redis redis-32-rhel7-working-container-15 add.go /data/
$ sudo ./buildah copy --chown redis:redis redis-32-rhel7-working-container-15 add.go /data/
$ sudo ./buildah copy --chown redis:redis redis-32-rhel7-working-container-15 https://github.com/projectatomic/buildah/blob/master/README.md /data/

# Owner is properly set
$ sudo ./buildah run redis-32-rhel7-working-container-15 sh
sh-4.2$ ls /data -l
total 64
-rw-------. 1 redis redis 45981 Nov 27 14:41 README.md
-rw-rw-r--. 1 redis redis  8779 Nov 27 14:35 add.go
drwxr-xr-x. 4 redis redis  4096 Nov 27 14:40 contrib
sh-4.2$ ls /data/contrib/ -l
completions/ rpm/
sh-4.2$ ls /data/contrib/rpm/buildah.spec -l
-rw-rw-r--. 1 redis redis 5347 Nov 23 13:49 /data/contrib/rpm/buildah.spec
sh-4.2$

```

Thanks for looking into this!

### DIFF
--- a/add.go
+++ b/add.go
@@ -12,9 +12,15 @@ import (
 	"time"
 
 	"github.com/containers/storage/pkg/archive"
+	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
+
+//AddAndCopyOptions holds options for add and copy commands.
+type AddAndCopyOptions struct {
+	Chown string
+}
 
 // addURL copies the contents of the source URL to the destination.  This is
 // its own function so that deferred closes happen after we're done pulling
@@ -58,7 +64,7 @@ func addURL(destination, srcurl string) error {
 // Add copies the contents of the specified sources into the container's root
 // filesystem, optionally extracting contents of local files that look like
 // non-empty archives.
-func (b *Builder) Add(destination string, extract bool, source ...string) error {
+func (b *Builder) Add(destination string, extract bool, options AddAndCopyOptions, source ...string) error {
 	mountPoint, err := b.Mount(b.MountLabel)
 	if err != nil {
 		return err
@@ -100,6 +106,11 @@ func (b *Builder) Add(destination string, extract bool, source ...string) error 
 	if len(source) > 1 && (destfi == nil || !destfi.IsDir()) {
 		return errors.Errorf("destination %q is not a directory", dest)
 	}
+	// Find out which user (and group) the destination should belong to.
+	user, err := b.user(mountPoint, options)
+	if err != nil {
+		return err
+	}
 	for _, src := range source {
 		if strings.HasPrefix(src, "http://") || strings.HasPrefix(src, "https://") {
 			// We assume that source is a file, and we're copying
@@ -116,6 +127,9 @@ func (b *Builder) Add(destination string, extract bool, source ...string) error 
 				d = filepath.Join(dest, path.Base(url.Path))
 			}
 			if err := addURL(d, src); err != nil {
+				return err
+			}
+			if err := setOwner(d, user); err != nil {
 				return err
 			}
 			continue
@@ -146,6 +160,9 @@ func (b *Builder) Add(destination string, extract bool, source ...string) error 
 				if err := copyWithTar(gsrc, d); err != nil {
 					return errors.Wrapf(err, "error copying %q to %q", gsrc, d)
 				}
+				if err := setOwner(d, user); err != nil {
+					return err
+				}
 				continue
 			}
 			if !extract || !archive.IsArchivePath(gsrc) {
@@ -161,6 +178,9 @@ func (b *Builder) Add(destination string, extract bool, source ...string) error 
 				if err := copyFileWithTar(gsrc, d); err != nil {
 					return errors.Wrapf(err, "error copying %q to %q", gsrc, d)
 				}
+				if err := setOwner(d, user); err != nil {
+					return err
+				}
 				continue
 			}
 			// We're extracting an archive into the destination directory.
@@ -169,6 +189,39 @@ func (b *Builder) Add(destination string, extract bool, source ...string) error 
 				return errors.Wrapf(err, "error extracting %q into %q", gsrc, dest)
 			}
 		}
+	}
+	return nil
+}
+
+// user returns the user (and group) information which the destination should belong to.
+func (b *Builder) user(mountPoint string, options AddAndCopyOptions) (specs.User, error) {
+	if options.Chown != "" {
+		return getUser(mountPoint, options.Chown)
+	}
+	return getUser(mountPoint, b.User())
+}
+
+// setOwner sets the uid and gid owners of a given path.
+// If path is a directory, recursively changes the owner.
+func setOwner(path string, user specs.User) error {
+	fi, err := os.Stat(path)
+	if err != nil {
+		return errors.Wrapf(err, "error reading %q", path)
+	}
+	if fi.IsDir() {
+		err2 := filepath.Walk(path, func(p string, info os.FileInfo, we error) error {
+			if err3 := os.Lchown(p, int(user.UID), int(user.GID)); err3 != nil {
+				return errors.Wrapf(err3, "error setting ownership of %q", p)
+			}
+			return nil
+		})
+		if err2 != nil {
+			return errors.Wrapf(err2, "error walking dir %q to set ownership", path)
+		}
+		return nil
+	}
+	if err := os.Lchown(path, int(user.UID), int(user.GID)); err != nil {
+		return errors.Wrapf(err, "error setting ownership of %q", path)
 	}
 	return nil
 }
--- a/cmd/buildah/addcopy.go
+++ b/cmd/buildah/addcopy.go
@@ -2,10 +2,17 @@ package main
 
 import (
 	"github.com/pkg/errors"
+	"github.com/projectatomic/buildah"
 	"github.com/urfave/cli"
 )
 
 var (
+	addAndCopyFlags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "chown",
+			Usage: "Set the user and group ownership of the destination content",
+		},
+	}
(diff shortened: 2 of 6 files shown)


TomSweeneyRedHat:
@bertinatto Thanks for the PR/PoC.  The change sounds good to me, but I'll let @nalind and @rhatdan give the definitive answer.  If we do move forward, we'll want to add some tests and man page changes too.

The test failure is not related to this change.  @rhatdan the error is these two versions aren't matching:
# buildah version: 0.8
# buildah rpm version: 0.7

Does buildah.spec need to have it's version bumped?  It's set at 0.7 atm.

rhatdan:
bot, add author to whitelist

rhatdan:
Since this made it into Docker, I agree that it should be implemented.

rhatdan:
https://github.com/moby/moby/pull/34263

add.go
```diff
@@ -161,6 +168,24 @@ func (b *Builder) Add(destination string, extract bool, source ...string) error
 				if err := copyFileWithTar(gsrc, d); err != nil {
 					return errors.Wrapf(err, "error copying %q to %q", gsrc, d)
 				}
+
+				// Set permissions. For now both uid and gid need to be set and they have to be numbers.
+				if options.Chown[0] != "" && options.Chown[1] != "" {
+					// TODO: figure out uid/id if user/group names was passed in instead of the real uid/gid.
```
rhatdan: I believe there is code for this already written to handle passing of of USER string.

add.go
```diff
@@ -16,6 +17,12 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+//AddAndCopyOptions holds options for add and copy commands.
+type AddAndCopyOptions struct {
+	// These are strings because I want to accept both ids and names.
+	Chown [2]string
```
rhatdan: Since  you have a struct, why not make it
USER string, 
Group String,

add.go
```diff
@@ -161,6 +168,24 @@ func (b *Builder) Add(destination string, extract bool, source ...string) error
...
+						return errors.Wrapf(err, "could not get uid %d", options.Chown[0])
+					}
+
+					gid, err := strconv.Atoi(options.Chown[1])
+					if err != nil {
+						return errors.Wrapf(err, "could not get gid %d", options.Chown[1])
+					}
+
+					if err := os.Chown(d, uid, gid); err != nil {
+						return errors.Wrapf(err, "error setting permissions")
```
rhatdan: error changing ownership %s:%s", options.Chown[0], options.Chown[1]

Although I would prefer options.User, options,Group

cmd/buildah/addcopy.go
```diff
@@ -1,18 +1,28 @@
...
 	"github.com/pkg/errors"
+	"github.com/projectatomic/buildah"
 	"github.com/urfave/cli"
 )
 
 var (
+	addAndCopyFlags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "chown",
+			Usage: "Set the user and group ownership of the file",
```
rhatdan: of the destination content.

cmd/buildah/addcopy.go
```diff
@@ -52,7 +67,16 @@ func addAndCopyCmd(c *cli.Context, extractLocalArchives bool) error {
 		return errors.Wrapf(err, "error reading build container %q", name)
 	}
 
-	err = builder.Add(dest, extractLocalArchives, args...)
+	options := buildah.AddAndCopyOptions{}
+	if chown := c.String("chown"); chown != "" {
```
rhatdan: Why does this have to be two, I would think you should just do a 
if c.String("chown") != "" {
Then split it. I don't think we need to care of group is not set.

--chown 1234
Should be ok, and then just leave the other component alone.
}

add.go
```diff
@@ -69,6 +78,12 @@ func (b *Builder) Add(destination string, extract bool, source ...string) error
 		}
 	}()
 	dest := mountPoint
+
+	uid, gid, err := findUserGroupIDs(mountPoint, options.Chown)
```
nalind: I think it'd be better to use `getUser()` from `util.go` here.

bertinatto: Wow, too bad I hadn't see `getUser()` before! Thanks for pointing this out.

bertinatto:
Thank you for the initial look! 

I've made a few adjustments, added a few tests and updated the man pages. Would you mind taking another look?

The second commit is just to make the tests pass. I tried to make my code simpler to avoid this change, but I' afraid there isn't much I can do. Any suggestions?

add.go
```diff
@@ -68,6 +74,18 @@ func (b *Builder) Add(destination string, extract bool, source ...string) error
 			logrus.Errorf("error unmounting container: %v", err2)
 		}
 	}()
+	// Set ownership of the destination.  If this information is
```
TomSweeneyRedHat: It's a fairly long time before the user variable is used from this point and a number of other errors that could happen before then.  I'm just wondering if it would make sense to move this new block of code down to be run just before the for statement on line 121.

tests/validate/gometalinter.sh
```diff
@@ -16,6 +16,6 @@ exec gometalinter.v1 \
 	--disable=gotype \
 	--disable=gas \
 	--disable=aligncheck \
-	--cyclo-over=40 \
+	--cyclo-over=45 \
```
TomSweeneyRedHat: Not sure what this change helps with?

add.go
bertinatto: That's right, I missed that!

tests/validate/gometalinter.sh
bertinatto: Commit 9227692 introduced a few `if`s in the `Add()` function, increasing it its  "cyclomatic complexity". As a result, `make validate` was failing with:

`add.go:67::warning: cyclomatic complexity 41 of function (*Builder).Add() is high (> 40) (gocyclo)`

I increased the limit in `gometalinter.sh` only to make the tests pass, however, I know this isn't a proper solution. Any suggestions on how to address this?

add.go
```diff
@@ -100,6 +106,18 @@ func (b *Builder) Add(destination string, extract bool, source ...string) error
 	if len(source) > 1 && (destfi == nil || !destfi.IsDir()) {
 		return errors.Errorf("destination %q is not a directory", dest)
 	}
+	// Find out which user/group should the destination belong to.
```
rhatdan: Create a separate function 
```
func (b *Builder) user(options AddAndCopyOptions) (string,  error) {
	// If this information is passed in through --chown, use it;
	// otherwise use the user/group as whom the container should run.
	if options.Chown != "" {
		return getUser(mountPoint, options.Chown)
	}
	 return getUser(mountPoint, b.User())
}
```

bertinatto: That's right! Thanks!

rhatdan:
LGTM
@nalind WDYT?

add.go
```diff
@@ -100,6 +106,11 @@ func (b *Builder) Add(destination string, extract bool, source ...string) error
 	if len(source) > 1 && (destfi == nil || !destfi.IsDir()) {
 		return errors.Errorf("destination %q is not a directory", dest)
 	}
+	// Find out which user (and group) should the destination belong to.
```
TomSweeneyRedHat: nit of a nit. "should the destination" to "the destination should".  But don't worry about it unless there are other needed changes.

add.go
```diff
@@ -172,3 +192,36 @@ func (b *Builder) Add(destination string, extract bool, source ...string) error
...
+// setOwner sets the uid and gid owners of a given path.
+// If path is a directory, recursively changes the owner.
+func setOwner(path string, user specs.User) error {
+	fi, err := os.Stat(path)
+	if err != nil {
+		return errors.Wrapf(err, "error reading %q", path)
+	}
+	if fi.IsDir() {
+		err2 := filepath.Walk(path, func(p string, info os.FileInfo, we error) error {
+			if err3 := os.Chown(p, int(user.UID), int(user.GID)); err3 != nil {
```
nalind: `os.Chown()` follows symlinks, and since we're not doing any of this in a chrooted subprocess, we could open ourselves up to something like #66 again.  I think using `os.Lchown()` is a first step here.

bertinatto: Hm, interesting perspective, thank you!

bertinatto:
Failing test seems to be unrelated to this patch:

```
# error pulling image "registry.fedoraproject.org/fedora:27": Error initializing image from source docker://registry.fedoraproject.org/fedora:27: pinging docker registry returned: error pinging repository, response code 503
```

TomSweeneyRedHat:
@bertinatto I concur that it's unrelated.  The fedora site seems to be hiccuping like crazy today.  I've seen this error in other PR's this morning and I am having difficulty pulling kits from there outside of GitHub.

rhatdan:
I talked to @nalind offline and he expresses concern about symlink attacks.
os.Chown() follows symlinks, we need to at least switch to os.Lchown, so it will not follow symlinks, and perhaps it is best if we did the chown within a chroot.

bertinatto:
Cool, tests are passing now!

rhatdan:
Ok I will merge, the question is do we need more protection with the chown call then just lchown.

rhatdan:
@bertinatto thanks for this.
@rh-atomic-bot r+

PR #340, opened 2017-11-28
Make CI tests pass again

nalind:
This corrects a version mismatch that CI should have caught, and works around a problem that we can't undo without rewriting history.


rhatdan:
@rh-atomic-bot r+

PR #341, opened 2017-11-28
Allow push to use the image id

TomSweeneyRedHat:
Signed-off-by: TomSweeneyRedHat <tsweeney@redhat.com>

Allows this command to work:
```
./buildah push bea0ba3aafbf oci:/tmp/image:test
```
where  bea0ba3aafbf is the ImageId for the image.  Prior when we passed 'image' to ParseStoreReference it would package up the image name as "docker.io/library/bea0ba3aafbf:latest" rather than "docker.io/library/alpine:latest" which is the valid name.

At line 312 of the commit.go file below we do:

builder, err := importBuilderFromImage(options.Store, importOptions)

The value returned from that in the builder.FromImage variable is "docker.io/library/alpine:latest" which ParseStoreReference parses appropriately and the push now succeeds.

Fixes #324

PR #342, opened 2017-11-30
tests/rpm.bats: use Fedora 27

nalind:
Update tests/rpm.bats to use Fedora 27, and add "make" as a build requirement in the .spec file that it uses, which I don't think is part of the latest set of images.


rhatdan:
I don't see the make change?

nalind:
That got quietly added in #341, so doing it here would just add a duplicate.

rhatdan:
@rh-atomic-bot r+

PR #354, opened 2017-12-11
include Debian install docs

anarcat:
This does small changes I have found necessary while installing Buildah on Debian stretch.

First, we splits install docs more clearly across different OSes - at first glance, you're tempted to copy-paste duplicate instructions, even across RHEL/Fedora. Then we explicitly run `make install` as root, showing how the build itself can be ran as non-root, and finally, include Debian-specific install docs, for which `apt-add-repository` unfortunately fails.

### DIFF
--- a/install.md
+++ b/install.md
@@ -103,3 +103,18 @@ Then to install Buildah on Ubuntu follow the steps in this example:
   make install
   buildah --help
 ```
+
+## Debian
+
+To install the required dependencies, you can use those commands, tested under Debian GNU/Linux amd64 9.3 (stretch):
+
+```
+gpg --recv-keys 0x018BA5AD9DF57A4448F0E6CF8BECF1637AD8C79D
+gpg --export 0x018BA5AD9DF57A4448F0E6CF8BECF1637AD8C79D >> /usr/share/keyrings/projectatomic-ppa.gpg
+echo 'deb [signed-by=/usr/share/keyrings/projectatomic-ppa.gpg] http://ppa.launchpad.net/projectatomic/ppa/ubuntu zesty main' > /etc/apt/sources.list.d/projectatomic-ppa.list
+apt update
+apt -y install -t stretch-backports libostree-dev 
+apt -y install bats btrfs-tools git libapparmor-dev libdevmapper-dev libglib2.0-dev libgpgme11-dev libseccomp-dev libselinux1-dev skopeo-containers go-md2man golang-1.8
+``` 
+
+The build steps on Debian are otherwise the same as Ubuntu, above.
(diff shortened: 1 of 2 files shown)


install.md
```diff
@@ -94,6 +100,20 @@ Then to install Buildah on Ubuntu follow the steps in this example:
...
 ```
+
+## Debian
+
+To install the required dependencies, you can use those commands, tested under Debian GNU/Linux amd64 9.3 (stretch):
+
+```
+gpg --recv-keys 0x018BA5AD9DF57A4448F0E6CF8BECF1637AD8C79D
+gpg --export 0x018BA5AD9DF57A4448F0E6CF8BECF1637AD8C79D >> /usr/share/keyrings/projectatomic-ppa.gpg
+echo 'deb [signed-by=/usr/share/keyrings/projectatomic-ppa.gpg] http://ppa.launchpad.net/projectatomic/ppa/ubuntu zesty main' > /etc/apt/sources.list.d/projectatomic-ppa.list
```
anarcat: note that we only add a single PPA here - projectatomic's. Debian stretch already has golang 1.8 (not 1.8.1 mind you, in backports), but I wonder if the flatpak PPA is also necessary... The only thing I can see would be required is libostree-dev, but that's also available in Debian (2017.12-2~bpo9+1 in backports). It would be nice to know which version of ostree is needed here as well...

anarcat:
incidentally, what's up with `make` here? why can't we just `go get github.com/projectatomic/buildah` anyways?

rhatdan:
bot, add author to whitelist

install.md
rhatdan: @giuseppe PTAL

install.md
```diff
@@ -50,6 +52,8 @@ Then to install Buildah on Fedora follow the steps in this example:
   buildah --help
 ```
 
+## RHEL 7
```
rhatdan: Probably should add CENTOS 7 here as well.

rhatdan:
@lsm5 Could you take a look at this.

rhatdan:
@lsm5 Could you take a look at this.

install.md
anarcat: alright, fixed up to say "RHEL, CentOS" instead. i haven't changed the specific instructions, assuming they are the same in CentOS...

rhatdan:
LGTM

rhatdan:
@umohnani8 PTAL

umohnani8:
LGTM

anarcat:
hold the presses

anarcat:
okay, now it's good - sorry for the noise, but there was one last tricky bit about the golang package I hadn't noticed.... `apt install golang-1.8` doesn't actually install the `/usr/bin/go` binary, that's the job of the metapackage, which needs to be installed from backports.

phew!

nalind:
The Makefile sets some build tags for the storage library so that it can build with older versions of some of the C libraries that it uses, and sets the commit ID and build date that are output by `buildah version`.  
The build tags not being present, if they're needed, will cause the compile to fail, but 'go get github.com/projectatomic/buildah/cmd/buildah' can produce a binary when the tags aren't necessary.  The `buildah version` command will fail when the commit ID and build date aren't set, though, so I can't recommend it as an alternative to using the makefile, at least not yet.

rhatdan:
@rh-atomic-bot r+

PR #360, opened 2017-12-13
Use configured registries to resolve image names

### LINKED ISSUE #356: buildah run works with symbolic image names but not hex
This works:

    root@curie:~# buildah from docker.io/anarcat/debian-airsonic:latest
    debian-airsonic-working-container-2

But this doesn't:

    root@curie:~# buildah from e87186c3d14
    error pulling image "docker.io/library/e87186c3d14:latest": Error initializing image from source docker://e87186c3d14:latest: errors:
    denied: requested access to the resource is denied

It looks like buildah is trying to fetch the container instead of looking in the local image storage, which has the image:

```
root@curie:~# buildah images
9ad0f3c4a8cd         docker.io/library/debian:stretch                         Nov 4, 2017 05:26      95.46 MB
e87186c3d14e         docker.io/anarcat/debian-airsonic:latest                 Nov 4, 2017 05:26      1.442 KB
2a90789e080d         docker.io/projectatomic/atomicapp:latest                 Oct 6, 2016 20:28      278.4 MB
```

nalind:
When locating an image for pulling, inspection, or pushing, if we're given an image name that doesn't include a domain/registry, try building a set of candidate names using the configured registries as domains, and then pull/inspect/push using the first of those names that works.

If a name that we're given corresponds to a prefix of the ID of a local image, skip completion and use the ID directly instead, which should fix #356.

### DIFF
--- a/import.go
+++ b/import.go
@@ -16,9 +16,9 @@ func importBuilderDataFromImage(store storage.Store, systemContext *types.System
 	imageName := ""
 
 	if imageID != "" {
-		ref, err := is.Transport.ParseStoreReference(store, "@"+imageID)
+		ref, err := is.Transport.ParseStoreReference(store, imageID)
 		if err != nil {
-			return nil, errors.Wrapf(err, "no such image %q", "@"+imageID)
+			return nil, errors.Wrapf(err, "no such image %q", imageID)
 		}
 		src, err2 := ref.NewImage(systemContext)
 		if err2 != nil {
@@ -68,7 +68,7 @@ func importBuilder(store storage.Store, options ImportOptions) (*Builder, error)
 		return nil, err
 	}
 
-	systemContext := getSystemContext(options.SignaturePolicyPath)
+	systemContext := getSystemContext(&types.SystemContext{}, options.SignaturePolicyPath)
 
 	builder, err := importBuilderDataFromImage(store, systemContext, c.ImageID, options.Container, c.ID)
 	if err != nil {
@@ -95,21 +95,27 @@ func importBuilder(store storage.Store, options ImportOptions) (*Builder, error)
 }
 
 func importBuilderFromImage(store storage.Store, options ImportFromImageOptions) (*Builder, error) {
+	var img *storage.Image
+	var err error
+
 	if options.Image == "" {
 		return nil, errors.Errorf("image name must be specified")
 	}
 
-	img, err := util.FindImage(store, options.Image)
-	if err != nil {
-		return nil, errors.Wrapf(err, "error locating image %q for importing settings", options.Image)
+	systemContext := getSystemContext(options.SystemContext, options.SignaturePolicyPath)
+
+	for _, image := range util.ResolveName(options.Image, "", systemContext, store) {
+		img, err = util.FindImage(store, image)
+		if err != nil {
+			continue
+		}
+
+		builder, err2 := importBuilderDataFromImage(store, systemContext, img.ID, "", "")
+		if err2 != nil {
+			return nil, errors.Wrapf(err2, "error importing build settings from image %q", options.Image)
+		}
+
+		return builder, nil
 	}
-
-	systemContext := getSystemContext(options.SignaturePolicyPath)
-
-	builder, err := importBuilderDataFromImage(store, systemContext, img.ID, "", "")
-	if err != nil {
-		return nil, errors.Wrapf(err, "error importing build settings from image %q", options.Image)
-	}
-
-	return builder, nil
+	return nil, errors.Wrapf(err, "error locating image %q for importing settings", options.Image)
 }
--- a/pull.go
+++ b/pull.go
@@ -53,27 +53,17 @@ func localImageNameForReference(store storage.Store, srcRef types.ImageReference
 	return name, nil
 }
 
-func pullImage(store storage.Store, options BuilderOptions, sc *types.SystemContext) (types.ImageReference, error) {
-	name := options.FromImage
-
-	spec := name
-	if options.Registry != "" {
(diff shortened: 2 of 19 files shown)


util/util.go
```diff
@@ -1,27 +1,122 @@
...
+	}
+	if named.String() == name {
+		// Parsing produced the same result, so there was a domain name in there to begin with.
+		return []string{name}
+	}
+	if reference.Domain(named) != "" && RegistryDefaultPathPrefix[reference.Domain(named)] != "" {
+		// If this domain can cause us to insert something in the middle, check if that happened.
+		repoPath := reference.Path(named)
+		defaultPrefix := RegistryDefaultPathPrefix[reference.Domain(named)] + "/"
+		if strings.HasPrefix(repoPath, defaultPrefix) {
```
ripcurld0: I think you can avoid the nested-ifs:

```Go
if strings.HasPrefix(repoPath, defaultPrefix) && path.Join(reference.Domain(named), repoPath[len(defaultPrefix):]) == name {
   ...
}
```

WDYT?

docs/buildah-from.md
```diff
@@ -67,6 +67,22 @@ Pull the image even if a version of the image is already present.
 
 If an image needs to be pulled from the registry, suppress progress output.
 
+**--registries-conf** *path*
+
+Pathname of the configuration file which specifies which registries should
+be consulted when completing image names which do not include a registry or
+domain portion.  It is not recommended that this option be used, as the default
+behavior of using the system-wide configuration (frequently
```
rhatdan: Lets just say default is /etc/containers/registiers.conf.

I have no idea what frequiently...* preferred means as a user.

cmd/buildah/commit.go
```diff
@@ -51,6 +51,14 @@ var (
 			Name:  "signature-policy",
 			Usage: "`pathname` of signature policy file (not usually used)",
 		},
+		cli.StringFlag{
```
rhatdan: Should we just make these global options of the buildah command rather then adding them to a bunch of commands.

pull.go
```diff
@@ -53,27 +53,17 @@ func localImageNameForReference(store storage.Store, srcRef types.ImageReference
...
-		spec = options.Registry + spec
-	}
-	spec2 := spec
-	if options.Transport != "" {
-		spec2 = options.Transport + spec
-	}
-
-	srcRef, err := alltransports.ParseImageName(name)
+func pullImage(store storage.Store, imageName string, options BuilderOptions, sc *types.SystemContext) (types.ImageReference, error) {
+	spec := imageName
```
TomSweeneyRedHat: could the imageName be just the ID at this point?  If so I think we still end up with a srcRef that has that id in it rather than converting it the right one.   i.e.:

docker.io/library/e87186c3d14:latest

rather than 
docker.io/anarcat/debian-airsonic:latest

I had added a 'importBuilderFromImage() call here and that converted the srcRef appropriately for me, but I was trying to figure out how to get the code to recognize that the image was stored locally instead of the original pull location at github.

tests/byid.bats
```diff
@@ -0,0 +1,102 @@
...
+
+    # Get the image's ID.
+    run buildah --debug=false images -q $IMAGE
+    echo "$output"
+    [ $status -eq 0 ]
+    [ $(wc -l <<< "$output") -eq 1 ]
+    iid="$output"
+
+    # Use the image's ID to push it.
+    run buildah push $iid dir:$TARGET
```
TomSweeneyRedHat: Looks like you need a signature-policy here too, probably at line 96 as well.

cmd/buildah/commit.go
nalind: Hmm, no harm in that, it'd just be ignored a lot of the time.

docs/buildah-from.md
nalind: Changing it.

pull.go
nalind: It shouldn't be - if it matches an image that we have locally, we'll skip attempts to search registries for it.  Otherwise, we'll fall through and try to treat the ID as a name, but I don't think the registry protocol lets us pull an image without a name.

util/util.go
nalind: Yeah, that'd work.  Changing it.

tests/byid.bats
nalind: Doh.  Adding it.

pull.go
TomSweeneyRedHat: I just took another look.  What I was concerned about is if  you passed in the id (e87186c3d14) into the call for ResolveName() it would just find the long version of that id and potentially send that back to pullAndFindImage() which would set srcRef to the image id.  But that only happens in newBuilder(), I thought elsewhere, so that should be OK and we're probably not passing in an ID at that point anyway.

rhatdan:
@nalind Needs rebase.

import.go
```diff
@@ -95,21 +95,27 @@ func importBuilder(store storage.Store, options ImportOptions) (*Builder, error)
...
+		if err != nil {
+			continue
+		}
 
-	builder, err := importBuilderDataFromImage(store, systemContext, img.ID, "", "")
-	if err != nil {
-		return nil, errors.Wrapf(err, "error importing build settings from image %q", options.Image)
-	}
+		builder, err2 := importBuilderDataFromImage(store, systemContext, img.ID, "", "")
+		if err != nil {
```
TomSweeneyRedHat: should this be err2?

nalind: Yup.  Fixing.

rhatdan:
@rh-atomic-bot r+

PR #361, opened 2017-12-14
Add a test that 'rmi' works with truncated image IDs

nalind:
Add a test to ensure that 'buildah rmi' works if passed a truncated version of an image's ID.  (This really should have been part of #360, but I didn't think to add it there.)


rhatdan:
LGTM
@rh-atomic-bot r+

PR #364, opened 2017-12-15
Update vendor on urfave/cli to support single character booleans

rhatdan:
Many other fixes have been made in urfave/cli


TomSweeneyRedHat:
Changes LGTM, but it looks like the change upset a couple of bats tests.

nalind:
Do we need to be setting `UseShortOptionHandling` in our `cli.Command` structures to allow combining of short options?

rhatdan:
@nalind You are taking this one over correct?

Issue #363, opened 2017-12-15
Add function `string` to format the output


gouyang:
```
[root@fedora ~]# buildah inspect -t image -f '{{ .Config }}' docker.io/library/busybox 
[...]

PR #368, opened 2017-12-18
Display Config and Manifest as strings

ripcurld0:
Builder has the Config and the Manifest fields which are `[]byte`. The type is `[]byte` because it's eaiser to serialize them that way. However, when these fields are displayed in the following way:

```
$ buildah inspect -f '{{.Config}} {{.Manifest}}' IMAGE
```

they might be shown as a byte slice.

This patch suggests to use a struct that wraps Builder's exposed fields such as Config, Manifest and Container as strings and thus make them readable when inspecting specific field.

related to #363

### DIFF
--- a/buildah.go
+++ b/buildah.go
@@ -95,6 +95,46 @@ type Builder struct {
 	DefaultMountsFilePath string `json:"defaultMountsFilePath,omitempty"`
 }
 
+// BuilderInfo are used as objects to display container information
+type BuilderInfo struct {
+	Type                  string
+	FromImage             string
+	FromImageID           string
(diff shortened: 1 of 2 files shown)


rhatdan:
bot, add author to whitelist

rhatdan:
LGTM

TomSweeneyRedHat:
Change LGTM, tests are unhappy with some kind of libostree issue.  I don't think it's at all related.

rhatdan:
@rh-atomic-bot r+

ripcurld0:
Thank you

Issue #240, opened 2017-08-02
Issues with Centos/ version 3.* Kernels


TomSweeneyRedHat:
Pinesh Shah Reports as a reply to the buildah blog:

> Hi Tom,
> 
> It was really a nice and simple article to have a quick view of buildah at work.
> 
> Although i tried setting up buildah on centos, the buildah command is up and now i am able to see the buildah version information
> 
> $buildah --version
> buildah version 0.3 (image-spec 1.0.0, runtime-spec 1.0.0)
> 
> Although when i try a simple buildah from <containername> for eg.
> $buildah from centos
> ERRO[0000] driver not supported
> driver not supported
> 
> Not sure why did you face something similar to this or any idea where i should start exploring on this.
> 
> Thanks & Regards,
> Pinesh Shah

In addition I tried myself and received a kernel age error.  I suspect that Pinesh's issue is due to runc being older than buildah is expecting.  I'm not as sure about the error that I received, it may be a configuration issue as I threw the vm together pretty quickly.

# buildah from centos
ERRO[0000] kernel too old to provide multiple lowers feature for overlay: driver not supported
kernel too old to provide multiple lowers feature for overlay: driver not supported 

# cat /etc/*release
CentOS Linux release 7.3.1611 (Core) 
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"

CentOS Linux release 7.3.1611 (Core) 
CentOS Linux release 7.3.1611 (Core) 
[root@localhost buildah]# rpm -qa | grep runc
runc-1.0.0-6.gite800860.el7.x86_64

rhatdan:
buildah --storage-driver=devicemapper ...

Should fix his issue.  There is a change I just got merged into containers/storage that was forcing overlay on CentOS and RHEL, which is causing these issues, I believe.

TomSweeneyRedHat:
[...]

PR #372, opened 2018-01-02
Add kernel version requirement to install.md and touchups

TomSweeneyRedHat:

Added a kernel version requirement to install.md to satisfy #240.  Also moved the runc requirements from README.md to install.md where I think they belong.  Then some general polishing of install.md.

### DIFF
--- a/install.md
+++ b/install.md
@@ -1,6 +1,28 @@
 # Installation Instructions
 
-Prior to installing Buildah, install the following packages on your linux distro:
+## System Requirements
+
+### Kernel Version Requirements
+To run Buildah on Red Hat Enterprise Linux or CentOS, version 7.4 or higher is required.
+On other Linux distributions Buildah requires a kernel version of 4.0 or
+higher in order to support the OverlayFS filesystem.  The kernel version can be checked
+with the 'uname -a' command.
+
+### runc Requirement
+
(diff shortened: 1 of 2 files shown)


install.md
```diff
@@ -1,6 +1,27 @@
 # Installation Instructions
 
-Prior to installing Buildah, install the following packages on your linux distro:
+## System Requirements
+
+### Kernel Version Required
+
+Buildah requires that the kernel be at version 4.0 or higher in order to support the OverlayFS filesystem.
```
rhatdan: This is wrong, since it eliminates Centos and RHEL systems, which run fine with the OverlayFS.  Should probably point out that on RHEL and CENTOS 7.4 and greater it also works.

TomSweeneyRedHat: OK, I've changed to:

### Kernel Version Requirements
To run Buildah on Red Hat Enterprise Linux or CentOS, version 7.4 or higher is required.
On other Linux distributions Buildah requires a kernel version of 4.0 or
higher in order to support the OverlayFS filesystem.  The kernel version can be checked
with the 'uname -a' command.

Holler if that needs tweaking still.

rhatdan:
LGTM
@rh-atomic-bot r+

PR #378, opened 2018-01-03
Increase compile spead using go build -i

rhatdan:
Brent found a feature in go that implements --incremental builds
Basically this flag caused builds to happen much quicker, making
developers lifes happier.

### DIFF
--- a/tests/validate/gometalinter.sh
+++ b/tests/validate/gometalinter.sh
@@ -11,6 +11,7 @@ exec gometalinter.v1 \
 	--enable-gc \
 	--exclude='error return value not checked.*(Close|Log|Print).*\(errcheck\)$' \
 	--exclude='.*_test\.go:.*error return value not checked.*\(errcheck\)$' \
+	--exclude='declaration of.*err.*shadows declaration.*\(vetshadow\)$'\
 	--exclude='duplicate of.*_test.go.*\(dupl\)$' \
 	--exclude='vendor\/.*' \
 	--disable=gotype \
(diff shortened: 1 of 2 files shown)


rhatdan:
@nalind @TomSweeneyRedHat @baude PTAL

baude:
LGTM as long as make clean will wipe the cache.

tests/validate/gometalinter.sh
```diff
@@ -11,6 +11,7 @@ exec gometalinter.v1 \
 	--enable-gc \
 	--exclude='error return value not checked.*(Close|Log|Print).*\(errcheck\)$' \
 	--exclude='.*_test\.go:.*error return value not checked.*\(errcheck\)$' \
+	--exclude='declaration of.*err.*shadows declaration.*\(vetshadow\)$'\
```
nalind: Is this part of a different PR?

TomSweeneyRedHat: I don't recall this being there when I reviewed earlier, a merge hiccup perhaps?

rhatdan: It was failing on testing for some reason.  Lint was checking on vetshadow of err.  
BTW I am seeing no speed increase by adding this patch.  So I think we should ignore it and add the vetshadow patch separately.

rhatdan:
I am going to close this one, since I don't see the speedup from build -i.

tests/validate/gometalinter.sh
nalind: You wouldn't see a speed improvement the first time, since `-i` is more or less doing a `go install`, saving compiled packages to disk so that they don't have to be recompiled the second time -- compare by running `go build -v -i -o buildah ./cmd/buildah` twice.

rhatdan: Ok I will reopen.
The question @baude had was whether a make clean would reverse the go install code so everything would be rechecked.

nalind: It doesn't, but adding a `go clean -i -r ./cmd/buildah` does.  I'm not really a fan of installing things under $GOPATH as a side-effect of compiling our binaries, though.

PR #379, opened 2018-01-03
Make lint was complaining about some vetshowed err

rhatdan:
We often use err as a variable inside of subblocks, and
we don't want golint to complain about it.


rhatdan:
@nalind PTAL

nalind:
The feature's experimental (i.e., known to not be 100% correct), so okay.
@rh-atomic-bot +r

nalind:
Oh, right.
@rh-atomic-bot r+

PR #380, opened 2018-01-04
When we say we skip a secrets config file, do so

nalind:
When we warn about not processing a secrets configuration file, actually skip anything we might have salvaged from it to make our behavior match the warning.


TomSweeneyRedHat:
@rh-atomic-bot r

rhatdan:
@rh-atomic-bot r+

PR #381, opened 2018-01-04
Use NewImageSource() instead of NewImage()

nalind:
Use NewImageSource() instead of NewImage() when checking if an image is actually there, since it makes the image library do less work while answering the same question for us.


rhatdan:
@rh-atomic-bot r+

Issue #374, opened 2018-01-02
Failure to build image in 0.10-2


smarterclayton:
CentOS 7.4

```
$ runc --version
runc version 1.0.0-rc4+dev
commit: a031838b23e9990f67a3b4c0e89bf3d935d1298b-dirty
spec: 1.0.0
$ buildah version
Version:       0.10
Go Version:    go1.8.3
Image Spec:    1.0.0
Runtime Spec:  1.0.0
Git Commit:    129fb10
Built:         Tue Dec 26 12:37:51 2017
OS/Arch:       linux/amd64
```

Attempted to build https://github.com/openshift/release projects/origin-release/base with buildah bud, got this repeatable failure when trying to commit

```
STEP 5: COMMIT containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage]@151c26a38f49601ec2cc77514f47d79a0166f1512b93cd64a5d667e6a29b1d4a
error building: error copying layer and metadata: error committing new image "containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage]@151c26a38f49601ec2cc77514f47d79a0166f1512b93cd64a5d667e6a29b1d4a": error determining uncompressed digest for blob "sha256:e9c4879b5b6b53531b4585f3c5d0bb54784b77560612a1c186eb7af94d9a14dc"
```

Other containers were able to commit.  What debugging info can I provide (is 100% reproducible)

nalind:
Can you attach the full output from 'buildah images' followed by 'buildah --debug bud ...'?  That'll tell us whether blob `sha256:e9c4879b5b6b53531b4585f3c5d0bb54784b77560612a1c186eb7af94d9a14dc` was downloaded and stored by buildah, or by something else that ran before it.
If it is a layer that was pulled down by a different package which was built using older versions of the storage or image libraries, that information may have not been recorded.  In that case, it'd be worth checking if removing the local copy of the `centos:7` image and re-trying, forcing buildah to pull down the "from" image, allows it to succeed.

smarterclayton:
Just remove from /var/run/containers/storage and
/var/lib/containers/storage be cleaned up?

nalind:
If no container filesystems are mounted up, that's fine.  Alternately, something like `buildah containers -q | xargs buildah rm` followed by `buildah images -q | xargs buildah rmi` would also work.

rhatdan:
[...]

PR #382, opened 2018-01-04
Add --all to remove containers

TomSweeneyRedHat:

Add the --all flag to buildah rm.  Will follow up for images in a separate patch/pr.  Addresses at least part of #374

### DIFF
--- a/cmd/buildah/rm.go
+++ b/cmd/buildah/rm.go
@@ -10,19 +10,26 @@ import (
 
 var (
 	rmDescription = "Removes one or more working containers, unmounting them if necessary"
-	rmCommand     = cli.Command{
+	rmFlags       = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "all, a",
+			Usage: "Remove all containers",
+		},
+	}
+	rmCommand = cli.Command{
 		Name:        "rm",
 		Aliases:     []string{"delete"},
 		Usage:       "Remove one or more working containers",
 		Description: rmDescription,
 		Action:      rmCmd,
 		ArgsUsage:   "CONTAINER-NAME-OR-ID [...]",
+		Flags:       rmFlags,
 	}
 )
 
 func rmCmd(c *cli.Context) error {
 	args := c.Args()
-	if len(args) == 0 {
+	if len(args) == 0 && !c.Bool("all") {
 		return errors.Errorf("container ID must be specified")
 	}
 	store, err := getStore(c)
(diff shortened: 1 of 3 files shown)


cmd/buildah/rm.go
```diff
@@ -31,27 +39,45 @@ func rmCmd(c *cli.Context) error {
...
+			id := builder.ContainerID
+			err = builder.Delete()
+			if e == nil {
+				e = err
+			}
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "error removing container %q: %v\n", builder.Container, err)
+				continue
+			}
+			fmt.Printf("%s\n", id)
```
TomSweeneyRedHat: diff wasn't very kind here.  The code in the else was all the pre-existing code and didn't change except for an extra leading tab.

nalind:
LGTM

cmd/buildah/rm.go
```diff
@@ -31,27 +38,45 @@ func rmCmd(c *cli.Context) error {
 	}
 
 	var e error
-	for _, name := range args {
-		builder, err := openBuilder(store, name)
-		if e == nil {
-			e = err
-		}
+	if c.Bool("all") {
+		builders, err := openBuilders(store)
```
nalind: This returns the set of containers which buildah is using, but would not include those which it is not (for example, any created by cri-o).  Is this the intention?

TomSweeneyRedHat: I thought it was just to do Buildah containers.  @rhatdan?

TomSweeneyRedHat: That is I meant it to remove the containers shown with "buildah containers", holler if this doesn't hit that mark, seems to in my testing, but I didn't try adding a container via Podman in my tests.

rhatdan: Yes we need to be careful with this going forward,  We have to figure out when we merge cri-o/podman/buildah together, what is the expected output of this.

If buildah images

List all images, which I think it should eventually.  I would want buildah rmi -a to remove them all whether they are created via buildah or other tools.

rhatdan:
@rh-atomic-bot r+

cmd/buildah/rm.go
nalind: We set a `*` in the `BUILDER` column in the `buildah containers` output if the container has some buildah-specific data associated with it, and containers which don't (i.e., created by cri-o, and not yet touched by buildah) are not listed unless `buildah containers` was passed the `-a` flag.  The two `-a` flags don't seem to work the same way.

TomSweeneyRedHat: Hmm, sold.  I've to do rmi too, I'll test/look into the containers without the '*' tag too.

Issue #371, opened 2017-12-26
Whale Dockerfile causes non-empty layers error


TomSweeneyRedHat:
The Dockerfile:
```# cat Dockerfile
FROM docker/whalesay:latest
RUN apt-get -y update && apt-get install -y fortunes
CMD /usr/games/fortune -a | cowsay
```

When used by Buildah to create an image and then that image is pushed by Buildah to Docker Hub and then pulled from there by "docker run" doesn't run.  It shows the error:

```
# docker run docker.io/tomsweeneyredhat/blog:whale
Unable to find image 'docker.io/tomsweeneyredhat/blog:whale' locally
Trying to pull repository docker.io/tomsweeneyredhat/blog ...
sha256:435e92163ed7cefb978156a3ca9a8d67016af58c94d3e1a1ceb3b396f28fecf6: Pulling from docker.io/tomsweeneyredhat/blog
188ab0d3421d: Pull complete
079c8ca7aeb2: Pull complete
278f7f06ff44: Pull complete
4f4fb700ef54: Pull complete
b6675f4fe3bd: Pull complete
6f247a045fef: Pull complete
243e1699db27: Pull complete
81376b9db5bb: Pull complete
7d39d8990dc2: Pull complete
/usr/bin/docker-current: too many non-empty layers in History section.
See '/usr/bin/docker-current run --help'.
```

If the image is created by Buildah and pushed it can be run by Buildah.  If the image is created by Docker and pushed by Docker, it can be run by Docker.

More details in the following e-mails

> On 12/21/2017 07:53 PM, Tom Sweeney wrote:
>
>> Hi Nalin,
>>
>> Late this summer I was putting a blog together that Dan requested where
>> I'd create an image in Buildah, push it to a repo on docker.io, then run
>> it from Docker.  At the time it was failing due to issues that we
>> suspected Buildah #187 would fix.
>>
>> I've just installed a fresh Fedora, cloned and built Buildah, installed
>> Docker, and then did 'dnf -y update --enablerepo updates-testing" to try
>> and get the latest bits.  The build and push work just fine, but I'm
>> getting an error from Docker that looks to be happening just before the
>> container is run saying "too many non-empty layers in History section.".
>>
>> The test run is at:  http://pastebin.test.redhat.com/542756 if you want
>> to see it.  The error is different that before.  The blog in case you
>> want to see it in it's very rough state is at:
[...]

PR #383, opened 2018-01-04
Ignore sequential duplicate layers in v2s1 images

nalind:
The image library drops sequential duplicate layers from images when converting manifests from v2s1 to other formats, so we need to do the same when we do it ourselves.

### DIFF
--- a/config.go
+++ b/config.go
@@ -139,23 +139,30 @@ func makeDockerV2S1Image(manifest docker.V2S1Manifest) (docker.V2Image, error) {
 	}
 	// Build a filesystem history.
 	history := []docker.V2S2History{}
+	lastID := ""
 	for i := range manifest.History {
-		h := docker.V2S2History{
-			Created:    time.Now().UTC(),
-			Author:     "",
-			CreatedBy:  "",
-			Comment:    "",
-			EmptyLayer: false,
-		}
+		// Decode the compatibility field.
 		dcompat := docker.V1Compatibility{}
-		if err2 := json.Unmarshal([]byte(manifest.History[i].V1Compatibility), &dcompat); err2 == nil {
-			h.Created = dcompat.Created.UTC()
-			h.Author = dcompat.Author
-			h.Comment = dcompat.Comment
-			if len(dcompat.ContainerConfig.Cmd) > 0 {
-				h.CreatedBy = fmt.Sprintf("%v", dcompat.ContainerConfig.Cmd)
-			}
-			h.EmptyLayer = dcompat.ThrowAway
+		if err = json.Unmarshal([]byte(manifest.History[i].V1Compatibility), &dcompat); err != nil {
+			return docker.V2Image{}, errors.Errorf("error parsing image compatibility data (%q) from history", manifest.History[i].V1Compatibility)
+		}
+		// Skip this history item if it shares the ID of the last one
+		// that we saw, since the image library will do the same.
+		if i > 0 && dcompat.ID == lastID {
+			continue
+		}
+		lastID = dcompat.ID
+		// Construct a new history item using the recovered information.
+		createdBy := ""
+		if len(dcompat.ContainerConfig.Cmd) > 0 {
+			createdBy = fmt.Sprintf("%s", strings.Join(dcompat.ContainerConfig.Cmd, " "))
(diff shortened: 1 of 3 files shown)


nalind:
This should fix https://github.com/projectatomic/buildah/issues/371.

config.go
```diff
@@ -156,7 +158,14 @@ func makeDockerV2S1Image(manifest docker.V2S1Manifest) (docker.V2Image, error) {
 				h.CreatedBy = fmt.Sprintf("%v", dcompat.ContainerConfig.Cmd)
 			}
 			h.EmptyLayer = dcompat.ThrowAway
+			id = dcompat.ID
 		}
+		// Skip this history item if it shares the ID of the last one
+		// that we saw, since the image library will do the same.
+		if id != "" && id == lastID {
```
rhatdan: Is id capable of still being "" without being  an error?  If id == "" and last ID == "", should that be a continue?

nalind: That's a good point.  The decode failing could reasonably be a reason to bail on importing the image, in which case `id` would always be set.  I'll rework that bit.

nalind:
Added a test to `digest.bats` that exercises the sanity check we're adding here.

TomSweeneyRedHat:
Just pulled the latest and it looks to have solved #371

TomSweeneyRedHat:
LGTM

config.go
```diff
@@ -139,23 +139,31 @@ func makeDockerV2S1Image(manifest docker.V2S1Manifest) (docker.V2Image, error) {
...
-		if err2 := json.Unmarshal([]byte(manifest.History[i].V1Compatibility), &dcompat); err2 == nil {
-			h.Created = dcompat.Created.UTC()
-			h.Author = dcompat.Author
-			h.Comment = dcompat.Comment
-			if len(dcompat.ContainerConfig.Cmd) > 0 {
-				h.CreatedBy = fmt.Sprintf("%v", dcompat.ContainerConfig.Cmd)
-			}
-			h.EmptyLayer = dcompat.ThrowAway
+		err := json.Unmarshal([]byte(manifest.History[i].V1Compatibility), &dcompat)
+		if err != nil {
```
rhatdan: nit: Do this in one line.
if err := json ...

config.go
```diff
@@ -139,23 +139,31 @@ func makeDockerV2S1Image(manifest docker.V2S1Manifest) (docker.V2Image, error) {
...
+			return docker.V2Image{}, errors.Errorf("error parsing image compatibility data (%q) from history", manifest.History[i].V1Compatibility)
+		}
+		// Skip this history item if it shares the ID of the last one
+		// that we saw, since the image library will do the same.
+		if lastID != "" && dcompat.ID == lastID {
```
rhatdan: I thought you were going to fix the situation where lastID==""

config.go
```diff
@@ -139,23 +139,31 @@ func makeDockerV2S1Image(manifest docker.V2S1Manifest) (docker.V2Image, error) {
...
+			continue
+		}
+		lastID = dcompat.ID
+		// Construct a new history item using the recovered information.
+		createdBy := ""
+		if len(dcompat.ContainerConfig.Cmd) > 0 {
+			createdBy = fmt.Sprintf("%v", dcompat.ContainerConfig.Cmd)
```
rhatdan: Not fully understanding the code, but this looks weird.  You are setting createdBy to a Cmd?

config.go
nalind: Fixing.

config.go
nalind: Not sure I follow your meaning here.  The first time through the loop, lastID is still empty, but we can equally just check the loop index, so I'll change that.

config.go
nalind: Yup.  The field is confusingly named, but that's the information that should be in it.  Comparing the output of `buildah inspect --type=image` for a v2s1 image like `kubernetes/pause` or `docker/whalesay` with the output of a v2s2 image like `fedora:27` is probably the most instructive, and I've tweaked it a bit to increase the resemblance.

rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #384, opened 2018-01-06
Add --all functionality to rmi

TomSweeneyRedHat:

Add --all functionality to the rmi command to remove all containers.  This does a force to remove all of the associated containers for the images too.

### DIFF
--- a/cmd/buildah/rmi.go
+++ b/cmd/buildah/rmi.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"os"
 
 	is "github.com/containers/image/storage"
 	"github.com/containers/image/transports"
@@ -16,6 +17,10 @@ import (
 var (
 	rmiDescription = "removes one or more locally stored images."
 	rmiFlags       = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "all, a",
+			Usage: "remove all images",
+		},
 		cli.BoolFlag{
 			Name:  "force, f",
 			Usage: "force removal of the image",
@@ -33,11 +38,16 @@ var (
 
 func rmiCmd(c *cli.Context) error {
 	force := c.Bool("force")
+	removeAll := c.Bool("all")
 
 	args := c.Args()
-	if len(args) == 0 {
+	if len(args) == 0 && !removeAll {
 		return errors.Errorf("image name or ID must be specified")
 	}
+	if len(args) > 0 && removeAll {
+		return errors.Errorf("when using the --all switch, you may not pass any images names or IDs")
+	}
+
 	if err := validateFlags(c, rmiFlags); err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


TomSweeneyRedHat:
I was going back and forth about using force under the covers for the rmi -all.  I decided to do so thinking that if someone wanted to remove all images, they'd also be thinking to remove all the containers too.  I could instead throw up a warning, list the running container(s) associated with image and then continue on.  But I thought the ---all should be omni-potent.

Also I did not address Nalin's concerns about rm and non Buildah containers.  I need to do so.  I discovered an error in rm today, it needs to also remove the containers from the list of running containers associated with the image.  Got a bit lost in that and will address both of those in a separate PR.

rhatdan:
@TomSweeneyRedHat In podman you have to specify podman rmi -f -a to get force, I think we should do the same for buildah.

cmd/buildah/rmi.go
```diff
@@ -33,11 +38,16 @@ var (
 
 func rmiCmd(c *cli.Context) error {
 	force := c.Bool("force")
+	removeAll := c.Bool("all")
 
 	args := c.Args()
-	if len(args) == 0 {
+	if len(args) == 0 && !c.Bool("all") {
```
ripcurld0: why not using `removeAll` instead of `c.Bool("all")`

TomSweeneyRedHat: I should, thx for the ptr!

TomSweeneyRedHat:
The tests are failing due to no status coming back from rmi.  I'll have to cook up another way to validate the removals later today or tomorrow.  I also have a weird issue with alpine  in the first rmi test.

TomSweeneyRedHat:
Green buttons and ready for review.  I've worked through the bats testing issues that I encountered.

cmd/buildah/rmi.go
```diff
@@ -47,26 +57,58 @@ func rmiCmd(c *cli.Context) error {
...
+	}
+
+	for _, id := range imagesToDelete {
+		image, err := getImage(id, store)
+		if err != nil || image == nil {
+			if lastError != nil {
+				fmt.Fprintln(os.Stderr, lastError)
+			}
+			if err == nil {
+				err = errors.New("Image does not Exist")
```
rhatdan: How about.
err=storage.ErrNotAnImage

cmd/buildah/rmi.go
```diff
@@ -47,26 +57,58 @@ func rmiCmd(c *cli.Context) error {
...
+				}
+				lastError = errors.Wrapf(err, "error getting running containers for image %q", id)
+				continue
 			}
 			if len(ctrIDs) > 0 && len(image.Names) <= 1 {
 				if force {
 					err = removeContainers(ctrIDs, store)
 					if err != nil {
-						return errors.Wrapf(err, "error removing containers %v for image %q", ctrIDs, id)
+						if lastError != nil {
```
rhatdan: Should we check if the error was container is not running?  Or does removeContainers take care of this for us?  Otherwise we have a race condition.

cmd/buildah/rmi.go
TomSweeneyRedHat: Sold, I'll also change the errors.New a few lines down to use storage.ErrImageUsedByContainer instead of Container Exists.

cmd/buildah/rmi.go
TomSweeneyRedHat: removeContainers is a stub that calls store.DeleteContainers.  That appears to delete containers without a "running" check.  I'll let @nalind slap me around if I'm off-base.

I setup a registry that was running in a container and did "buildah rmi --all --force" and it wiped out the images and the associated container including the one running the registry.  However the registry was still up but not functioning very happily until I <ctrl>-c'd it.

I setup  another container that printed to stdout, slept a second, then looped back to the print.  After I removed the container, the prints kept happening but the sleeps did not as /usr/sbin/sleep returned "No such file or directory".  It died by itself when the loop finished up

So long story short, we delete running containers, they're left in a haphazard state until they die.  Do we need to adjust?  Personally I think perhaps a stronger warning in the man page would suffice.

If you're interested, here's the rmi process for the registry below.

```
# buildah images
177391bcf802         docker.io/library/registry:2                             Dec 1, 2017 22:15      34.1 MB
[root@docker buildah]# buildah containers
9fad5cae923a     *                  scratch                          working-container
5638b62b89ec     *     177391bcf802 docker.io/library/registry:2     registry-working-container
[root@docker buildah]# buildah rmi -f --all
177391bcf8024d4db90069c89920576acd00a4eaba72e53ca0dc8e9a7501f093
[root@docker buildah]# buildah images
[root@docker buildah]# buildah containers
9fad5cae923a     *                  scratch                          working-container
```

rhatdan:
@rh-atomic-bot r+

TomSweeneyRedHat:
Forgot to add that this will help with  #374

PR #389, opened 2018-01-13
Don't overwrite directory permissions on --chown

bertinatto:
In a previous PR ( #336) I failed to check if my changes were changing the ownership of:

1) existent target directory
2) existing files in the target directory

As a result, I'm proposing these changes to fix that. 

I'm happy to re-work my patch if you think the approach isn't appropriate.

Thanks.

### DIFF
--- a/add.go
+++ b/add.go
@@ -75,12 +75,17 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 			logrus.Errorf("error unmounting container: %v", err2)
 		}
 	}()
+	// Find out which user (and group) the destination should belong to.
+	user, err := b.user(mountPoint, options.Chown)
+	if err != nil {
+		return err
+	}
 	dest := mountPoint
 	if destination != "" && filepath.IsAbs(destination) {
 		dest = filepath.Join(dest, destination)
 	} else {
-		if err = os.MkdirAll(filepath.Join(dest, b.WorkDir()), 0755); err != nil {
-			return errors.Wrapf(err, "error ensuring directory %q exists)", filepath.Join(dest, b.WorkDir()))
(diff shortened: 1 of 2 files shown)


nalind:
bot, add author to whitelist

add.go
```diff
@@ -210,26 +210,45 @@ func (b *Builder) user(mountPoint string, userspec string) (specs.User, error) {
...
+}
+
+// ensureDir creates a directory if it doesn't exist, setting ownership and permissions as passed by user and perm.
+func ensureDir(path string, user specs.User, perm os.FileMode) error {
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		if err := os.MkdirAll(path, perm); err != nil {
+			return errors.Wrapf(err, "error ensuring directory %q exists", path)
+		}
+		if err := os.Chown(path, int(user.UID), int(user.GID)); err != nil {
+			return errors.Wrapf(err, "error ensuring directory %q exists", path)
```
rhatdan: SHouldn't this be
"error setting ownership of %q"

add.go
```diff
@@ -154,14 +154,14 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 				// to create it first, so that if there's a problem,
 				// we'll discover why that won't work.
 				d := dest
-				if err := os.MkdirAll(d, 0755); err != nil {
```
rhatdan: Nit, should just remove `d` and use `dest` everywhere.

add.go
bertinatto: Thanks for reviewing!

rhatdan:
LGTM
@nalind PTAL

rhatdan:
@rh-atomic-bot r+

PR #391, opened 2018-01-16
Touchup rm and container man pages

TomSweeneyRedHat:

Clarifies what 'buildah rm --all' does and adds some examples to the containers page.

### DIFF
--- a/docs/buildah-containers.md
+++ b/docs/buildah-containers.md
@@ -15,7 +15,7 @@ IDs, and the names and IDs of the images from which they were initialized.
 **--all, -a**
 
 List information about all containers, including those which were not created
-by and are not being used by Buildah.
+by and are not being used by Buildah.  Containers created by Buildah are denoted with an '*' in the 'BUILDER' column.
 
 **--json**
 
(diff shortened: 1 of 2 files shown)


docs/buildah-containers.md
```diff
@@ -36,12 +36,43 @@ Displays only the container IDs.
...
 buildah containers
+```
+CONTAINER ID  BUILDER  IMAGE ID     IMAGE NAME                       CONTAINER NAME
+29bdb522fc62     *     3fd9065eaf02 docker.io/library/alpine:latest  alpine-working-container
+c6b04237ac8e     *     f9b6f7f7b9d3 docker.io/library/busybox:latest busybox-working-container
+```
 
 buildah containers --quiet
+```
+29bdb522fc62d43fca0c1a0f11cfc6dfcfed169cf6cf25f928ebca1a612ff5b0
```
ripcurld0: I thought `buildah containers --quiet` should truncate `IMAGE ID`. Am I wrong?

TomSweeneyRedHat: 'buildah containers --quiet" shows only the container id, I don't believe it was ever designed to show the short version of the id, but will let @nalind correct me on that.

TomSweeneyRedHat:
Travis error is only in golang 1.7 and looks like a build environment issue.  

```
$ sudo apt-get -qq install bats btrfs-tools git libapparmor-dev libdevmapper-dev libglib2.0-dev libgpgme11-dev libselinux1-dev
E: Unable to locate package bats
The command "sudo apt-get -qq install bats btrfs-tools git libapparmor-dev libdevmapper-dev libglib2.0-dev libgpgme11-dev libselinux1-dev" failed and exited with 100 during .
```

docs/buildah-containers.md
ripcurld0: My bad. I meant `CONTAINER ID`.

From the below code it looks like even if `nontruncate` is false `CONTAINER ID` will never be truncated.

https://github.com/projectatomic/buildah/blob/fe2de4f491147e5063d12ff80431d80d42a37a0d/cmd/buildah/containers.go#L81-L82

nalind: That's correct.  I expected the output of `--quiet` to get piped to 'xargs buildah rm/rmi' more often than not.  This was back when we weren't as good about handling truncated IDs for images, so it made a lot of sense to not require a `--notruncate=false` to get back values that we could use for that case.  Since we're already defaulting to `--notruncate` for `--quiet`, I'm not really sure of a good approach for adding the ability to have the output truncate IDs.

ripcurld0: Thanks @nalind, I just wanted to be sure.

rhatdan:
I think they should be truncated by default.  This is what podman and docker do, don't see a real reason why buildah should be different.

rhatdan:
But that should be fixed in a different PR.

rhatdan:
@rh-atomic-bot r+

Issue #390, opened 2018-01-15
Cannot push image to ostree


gouyang:
it's a regression issue as it works before.

```
# buildah push docker.io/library/busybox:latest ostree:busybox:latest
Getting image source signatures
Copying blob sha256:ae25bfd8dcdeab5892b4efa7364c0a1bba59e11491e10415893b8905e3467799
 1.29 MiB / 1.29 MiB [======================================================] 0s
Copying config sha256:807fd4df40d170bfa2b3176ba4524cda95d9ec8073aac3b0d52b3e2d65559565
 1.46 KiB / 1.46 KiB [======================================================] 0s
Writing manifest to image destination
Storing signatures
error pushing image "docker.io/library/busybox:latest" to "ostree:busybox:latest": error copying layers and metadata: Error committing the finished image: gzip: invalid header


# buildah version
Version:       0.10
Go Version:    go1.9.2
Image Spec:    1.0.0
Runtime Spec:  1.0.0
Git Commit:    adfb256
Built:         Mon Jan 15 00:56:51 2018
OS/Arch:       linux/amd64
```

giuseppe:
it is addressed in https://github.com/containers/image/pull/401 by: "ostree: use archive.DecompressStream instead of gzip.NewReader"

giuseppe:
this can be closed as of https://github.com/projectatomic/buildah/pull/393

ripcurld0:
Thanks @gouyang and @giuseppe

PR #393, opened 2018-01-16
vendor containers/image to 386d6c33c9d622ed84baf14f4b1ff1be86800ccd

ripcurld0:
A follow-up to https://github.com/projectatomic/buildah/pull/385 and https://github.com/containers/image/pull/390.


nalind:
LGTM.  @giuseppe, this should also fix #390, right?

giuseppe:
@nalind yes correct.  It should fix #390

PR #399, opened 2018-01-18
Fix: setting the container name to the image

### LINKED ISSUE #396: buildah from image create unexpected container name
It should create container name `busybox2-working-container` rather than the name `busybox-working-container-4`.

```
[root@fedora buildah]# buildah from busybox2
busybox-working-container-4
```

ripcurld0:
In commit [47ac96155f1573e](https://github.com/projectatomic/buildah/commit/47ac96155f1573e1bbf7804461f6850ff67e8747) the image name that is used for setting the container name is taken from the resolved image unless it is empty.

The resolved image has the `Names` field and right now the first name is taken. However, when the image is a tagged image, the container name will end up using the original name instead of the given one.

For example:
```sh
$ buildah tag busybox busybox1
$ buildah from busybox1
```

Will set the name of the container as `busybox-working-container` while it was expected to be `busybox1-working-container`.

This patch fixes this particular issue.
fixes #396

### DIFF
--- a/new.go
+++ b/new.go
@@ -77,6 +77,24 @@ func pullAndFindImage(store storage.Store, imageName string, options BuilderOpti
 	return img, ref, nil
 }
 
+func getImageName(name string, img *storage.Image) string {
+	imageName := name
+	if len(img.Names) > 0 {
+		imageName = img.Names[0]
+		// When the image used by the container is a tagged image
+		// the container name might be set to the original image instead of
+		// the image given in the "form" command line.
+		// This loop is supposed to fix this.
+		for _, n := range img.Names {
+			if strings.Contains(n, name) {
+				imageName = n
+				break
+			}
+		}
+	}
+	return imageName
+}
+
 func imageNamePrefix(imageName string) string {
 	prefix := imageName
 	s := strings.Split(imageName, "/")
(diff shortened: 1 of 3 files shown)


new.go
```diff
@@ -202,6 +202,16 @@ func newBuilder(store storage.Store, options BuilderOptions) (*Builder, error) {
 	if img != nil {
 		if len(img.Names) > 0 {
 			image = img.Names[0]
+			// When the image used by the container is a tagged image
```
rhatdan: Create a function called getImageName and move this code to it, that will fix your tests failures.

ripcurld0: Yep, I did it and run the `make validation`. I just want to add some test(s) to that change too.
Thanks for the review, though.

rhatdan:
LGTM

rhatdan:

TomSweeneyRedHat:
The code changes LGTM and it tests out well.  I do have a few questions about the test though.  Can that be moved into the tests directory and is it run automatically with each CI?  Our other tests are all bats based and I'd prefer to keep the test in the same format if possible.  @nalind WDYT?

The test that you came up with does cover the cases that need to be tested though.

ripcurld0:
@TomSweeneyRedHat moving this to "tests" shouldn't be a problem.
If you prefer that way, I can make it happen.
Should we wait on @nalind ?

rhatdan:
I am fine with the unit test, but I would like to see at least one test that uses actual storage.  BTW CRI-O and Buildah are moving away from bats testing, we should probably follow suit.

ripcurld0:
@rhatdan @TomSweeneyRedHat you got it. Added a test to `from.bats`.

TomSweeneyRedHat:
@rhatdan, ack on the move away from bats.  I just wanted to get the functionality in the bats now so we didn't neglect moving it to the new test suite.  

@ripcurld0 thanks for the add to from.bats.  My OCD is kicking in and I'd really like to have the new test moved into the tests directory, but that's just me.

LGTM

rhatdan:
@rh-atomic-bot r+

ripcurld0:
Thank you so much

PR #406, opened 2018-01-23
Return a better error message when failed to resolve an image

### LINKED ISSUE #401: Incorrect error with from --creds un:pwd when creds are invalid
**Description**

buildah pull from --creds un:pwd gives a bad error message when the creds are invalid.  Rather than showing an error in the creds, it says image not found.

1.  Setup a private repository with credentials
2.  push an image to it
3.  pull image with creds that are incorrect.

You can use https://raw.githubusercontent.com/projectatomic/buildah/master/tests/test_buildah_authentication.sh and comment out the last five line in the comment section.  On a machine with Docker and Buildah (0.11) installed run:

```
/bin/bash -v test_buildah_authentication.sh
buildah from localhost:5000/my-commit-alpine --pull-always --cert-dir /root/auth  --tls-verify=true --creds=testuser:badpassword
```

```
# buildah from localhost:5000/my-commit-alpine --pull-always --cert-dir /root/auth  --tls-verify=true --creds=testuser:testpasswo
no such image "localhost:5000/my-commit-alpine": image not known
```

A better error message is displayed showing the errror.

```
buildah-0.11-1.git6bad262.fc27.x86_64
```

```
# buildah version
Version:       0.11
Go Version:    go1.9.2
Image Spec:    1.0.0
Runtime Spec:  1.0.0
Git Commit:    6bad262
Built:         Wed Jan 17 08:49:14 2018
OS/Arch:       linux/amd64
```

```
# cat /etc/*release
Fedora release 27 (Twenty Seven)
NAME=Fedora
VERSION="27 (Server Edition)"
ID=fedora
VERSION_ID=27
PRETTY_NAME="Fedora 27 (Server Edition)"
ANSI_COLOR="0;34"
CPE_NAME="cpe:/o:fedoraproject:fedora:27"
REDHAT_BUGZILLA_PRODUCT_VERSION=27
REDHAT_SUPPORT_PRODUCT_VERSION=27
VARIANT="Server Edition"
VARIANT_ID=server
Fedora release 27 (Twenty Seven)
Fedora release 27 (Twenty Seven)

```

**Output of `uname -a`:**

```
# uname -a
Linux localhost.localdomain 4.14.3-300.fc27.x86_64 #1 SMP Mon Dec 4 17:18:27 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

```
# cat /etc/containers/storage.conf

[storage]

# Default Storage Driver
driver = "overlay"

# Temporary storage location

[storage.options]
additionalimagestores = [
]

size = ""

```

ripcurld0:
During the creation of a new builder object there are errors that are only logged into `logrus.Debugf`.

If in the end of the process `ref` or `img` are `nil` and options.FromImage` is set then it means that there was an issue. By default, it was assumed that the image name is wrong.
Yet, this assumption isn't always correct. For example, it might fail due to authorization or connection errors.

In this patch, I am attempting to fix this problem by checking the last error stored in the `err` variable and returning the cause of the failure.

Fixes #401

### DIFF
--- a/new.go
+++ b/new.go
@@ -144,6 +144,7 @@ func newBuilder(store storage.Store, options BuilderOptions) (*Builder, error) {
 			pulledImg, pulledReference, err2 := pullAndFindImage(store, image, options, systemContext)
 			if err2 != nil {
 				logrus.Debugf("error pulling and reading image %q: %v", image, err2)
+				err = err2
 				continue
 			}
 			ref = pulledReference
@@ -155,11 +156,13 @@ func newBuilder(store storage.Store, options BuilderOptions) (*Builder, error) {
 		if err2 != nil {
 			if options.Transport == "" {
 				logrus.Debugf("error parsing image name %q: %v", image, err2)
+				err = err2
 				continue
 			}
 			srcRef2, err3 := alltransports.ParseImageName(options.Transport + image)
 			if err3 != nil {
 				logrus.Debugf("error parsing image name %q: %v", image, err2)
+				err = err3
 				continue
 			}
 			srcRef = srcRef2
@@ -186,6 +189,7 @@ func newBuilder(store storage.Store, options BuilderOptions) (*Builder, error) {
 			pulledImg, pulledReference, err2 := pullAndFindImage(store, image, options, systemContext)
 			if err2 != nil {
 				logrus.Debugf("error pulling and reading image %q: %v", image, err2)
+				err = err2
 				continue
 			}
 			ref = pulledReference
@@ -195,7 +199,10 @@ func newBuilder(store storage.Store, options BuilderOptions) (*Builder, error) {
 	}
 
 	if options.FromImage != "" && (ref == nil || img == nil) {
-		return nil, errors.Wrapf(storage.ErrImageUnknown, "no such image %q", options.FromImage)
+		// If options.FromImage is set but we ended up
+		// with nil in ref or in img then there was an error that
+		// we should return.
+		return nil, util.GetFailureCause(err, errors.Wrapf(storage.ErrImageUnknown, "no such image %q", options.FromImage))
 	}
 	image := options.FromImage
 	imageID := ""
--- a/util/util.go
+++ b/util/util.go
@@ -1,6 +1,8 @@
 package util
 
 import (
+	"fmt"
+	"net/url"
 	"path"
 	"strings"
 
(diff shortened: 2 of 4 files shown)


new.go
```diff
@@ -114,6 +116,17 @@ func imageManifestAndConfig(ref types.ImageReference, systemContext *types.Syste
 	return nil, nil, nil
 }
 
+func getFailureCause(err, defaultError error) error {
```
TomSweeneyRedHat: Is this a function we could (should?) put into buildah/util.go?  The same file that has ResolveName in it, this might be useful in other situations me thinks.

ripcurld0: Yeah, why not.

util/util.go
```diff
@@ -152,3 +154,17 @@ func AddImageNames(store storage.Store, image *storage.Image, addNames []string)
...
+
+// GetFailureCause checks the type of the error "err" and returns a new
+// error message that reflects the reason of the failure.
+// In case err type is not a familiar one the error "defaultError" is returned.
+func GetFailureCause(err, defaultError error) error {
+	switch nErr := errors.Cause(err).(type) {
+	case errcode.Error:
+		return nErr
+	case *url.Error:
+		return nErr
```
rhatdan: Is it intentional to return nErr in both cases?

ripcurld0:
@TomSweeneyRedHat thank you for noticing this!
I will look at this.

util/util.go
ripcurld0: Yes, I tried to use `falltrhough` but I got compiled error that said I can't do it.
Maybe I can change it a little bit so it will be more readable.

ripcurld0:
OK, I updated the code. PTAL.

ripcurld0:
The failure in Travis is not related to the commit:
> The command "sudo apt-get update" failed and exited with 100 during .

util/util.go
rhatdan: This should work.
```
case errcode.Error, *url.Error:
      return nErr
```

rhatdan:
Tests now pass.  @TomSweeneyRedHat can you confirm that you like the current handling.

TomSweeneyRedHat:
The end result looks very good on the tests now:
```
########
# Use bad password on from/pull, this should fail
########
buildah from localhost:5000/my-commit-alpine --pull-always --cert-dir /root/auth  --tls-verify=true --creds=testuser:badpassword
```
I'm a little concerned with the hack to make it so, but not sure there's much choice.  @nalind WDYT?

If @nalind is hip, I'm hip.

ripcurld0:
@TomSweeneyRedHat I will try to open a PR to `containers/images` to return an `errcode` in that part so we could remove the hack.

nalind:
LGTM

rhatdan:
Ok this is better then what we have.  Merging
@rh-atomic-bot r+

ripcurld0:
Thank you @nalind @TomSweeneyRedHat  @rhatdan

PR #408, opened 2018-01-23
Make bud be really quiet

TomSweeneyRedHat:

This changes makes the  'buildah bud -q .' command really be quiet.  Prior a portion but not all of the output from the build progression was shown when you did a build.  Now it's completely quiet, no output.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -380,6 +380,7 @@ func (b *Executor) Run(run imagebuilder.Run, config docker.Config) error {
 		Entrypoint:      config.Entrypoint,
 		Cmd:             config.Cmd,
 		NetworkDisabled: config.NetworkDisabled,
+		Quiet:           b.quiet,
 	}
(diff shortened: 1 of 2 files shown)


imagebuildah/build.go
```diff
@@ -380,6 +380,7 @@ func (b *Executor) Run(run imagebuilder.Run, config docker.Config) error {
 		Entrypoint:      config.Entrypoint,
 		Cmd:             config.Cmd,
 		NetworkDisabled: config.NetworkDisabled,
+		QuietBuild:      b.quiet,
```
rhatdan: Why not just call this Quiet?  Change the value to Quiet and then it LGTM

TomSweeneyRedHat: Changed to Quiet and just pushed.  Tests are cranking.

rhatdan:
@rh-atomic-bot r+

PR #422, opened 2018-01-29
Trim the image reference to just the name before calling getImageName

ripcurld0:
When setting a container name the getImageName function goes through all the names of the resolved image and finds the name that contains the given name by the user.

However, if the user is specifying "docker.io/tagged-image" the Docker transport returns "docker.io/library/tagged-image" which makes getImageName returns the original image name because it does not find a match (basically "docker.io/tagged-image" is not a substring of "docker.io/library/tagged-image").

To resolve this issue, before calling getImageName, the image given by the user will be trimmed to be just the image name.


rhatdan:
I have a concern with this in that docker.io/tagged-image might end up returning. 
registry.fedora.com/tagged-image when the user clearly stated he wanted docker.io/library/tagged-image.

ripcurld0:
Will that be a problem?
1) this happens if "registry.fedora.com/tagged-image" is a tag to "docker.io/library/tagged-image" which is OK because they have the same content.
2) either way we are going to use just the image name which is "tagged-image" and ignore the domain.

rhatdan:
What happens if both fedora.registry and docker.io have different fedora images.?

buildah from docker.io/fedora

ripcurld0:
I see what you're saying. You want `fedora.registry/nginx` and `docker.io/nginx` to have different container names.
Right now it will set the same container name because of this [line](https://github.com/ripcurld0/buildah/blob/1d8c27805fd4f4c9ac05eaf40198cb31ed102495/new.go#L241).

Was that what you meant?

rhatdan:
Yes

ripcurld0:
Do you want the domain to be reflected in the container name?

rhatdan:
No I just want to make sure we pull the correct image.  In the example above where there is an docker.io/library/nginx and a fedora.registry.com/nginx, if a user specifies docker.io/nginx, then we better pull the one from docker.  Perhaps you insert library when the user does not specify the full registry path?

ripcurld0:
the `library` is injected only if the [name doesn't contain "/" and the domain is `docker.io`](https://github.com/projectatomic/buildah/blob/c615c3e23d0a43dcb0989eec37edf4ff96430a84/vendor/github.com/containers/image/docker/reference/normalize.go#L61-L75).
Unless you specify the domain (f.e. `fedora.registry.com`) in the image name the domain is set to the default one which is [`docker.io`](https://github.com/projectatomic/buildah/blob/master/vendor/github.com/containers/image/docker/reference/normalize.go#L13).

Anyway, it looks to me that we are pulling the correct image:

`buildah from busybox` will pull `docker.io/library/busybox`
`buildah from ripcurld0/busybox` will pull `docker.io/ripcurld0/busybox`
`buildah from fedora.registry.com/busybox` will pull `fedora.registry.com/busybox`

Anything that I might be missing?

I guess I can make this a little bit better without using the `imageNamePrefix`, though.

ripcurld0:
So I guess that `imageNamePrefix` is a redundancy here.
In a second thought what I should have done in #399 is to remove the following part

```
if len(img.Names) > 0 {
    image = img.Names[0]
}
```

Because If `img != nil` it means the image name given by the user exists and thus there is no need to assign the name again.

@nalind @rhatdan @TomSweeneyRedHat  WDYT?

rhatdan:
No I just want to make sure we pull the correct image.  In the example above where there is an docker.io/library/nginx and a fedora.registry.com/nginx, if a user specifies docker.io/nginx, then we better pull the one from docker.  Perhaps you insert library when the user does not specify the full registry path?

rhatdan:
Can you check that pulling docker.io/fedora wih registry.fedoraproject.org as first in the registry lists, pulls from docker.io.

ripcurld0:
`/etc/containers/registries.d/default.yaml`: 
```
# This is the default signature write location for docker registries.                  
default-docker:                            
  sigstore-staging: file:///var/lib/atomic/sigstore                                    
docker:                                    
  registry.fedoraproject.org:              
    sigstore: https://registry.fedoraproject.org                                       
  docker.io:              
    sigstore: https://docker.io                                       
```

running `sudo ./buildah --debug from docker.io/fedora` prints
```
DEBU[0000] [graphdriver] trying provided driver "overlay"                              
DEBU[0000] overlay: override_kernelcheck=true                                          
DEBU[0000] overlay test mount with multiple lowers succeeded                           
DEBU[0000] backingFs=extfs, projectQuotaSupported=false                                
DEBU[0000] parsed reference to refname into "[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.override_kernel_check=true]docker.io/library/fedora:latest" 
DEBU[0000] parsed reference to refname into "[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.override_kernel_check=true]docker.io/library/fedora:latest" 
DEBU[0000] reference "[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.override_kernel_check=true]docker.io/library/fedora:latest" does not resolve to an image ID 
DEBU[0000] parsed reference to refname into "[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.override_kernel_check=true]docker.io/library/fedora:latest" 
DEBU[0000] parsed reference to refname into "[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.override_kernel_check=true]docker.io/library/fedora:latest" 
DEBU[0000] Using registries.d directory /etc/containers/registries.d for sigstore configuration                                                                               
DEBU[0000]  Using "default-docker" configuration                                       
DEBU[0000]  No signature storage configuration found for docker.io/library/fedora:latest                                                                                      
DEBU[0000] Looking for TLS certificates and private keys in /etc/docker/certs.d/docker.io                                                                                     
DEBU[0000] copying "docker://docker.io/fedora" to "docker.io/library/fedora:latest"    
DEBU[0000] Using registries.d directory /etc/containers/registries.d for sigstore configuration                                                                               
DEBU[0000]  Using "default-docker" configuration                                       
DEBU[0000]  No signature storage configuration found for docker.io/library/fedora:latest                                                                                      
DEBU[0000] Looking for TLS certificates and private keys in /etc/docker/certs.d/docker.io                                                                                     
DEBU[0000] GET https://registry-1.docker.io/v2/                                        
DEBU[0000] Ping https://registry-1.docker.io/v2/ err <nil>                             
DEBU[0000] Ping https://registry-1.docker.io/v2/ status 401                            
DEBU[0001] GET https://registry-1.docker.io/v2/library/fedora/manifests/latest         
DEBU[0002] Source is a manifest list; copying (only) instance sha256:1b9bfb4e634dc1e5c19d0fa1eb2e5a28a5c2b498e3d3e4ac742bd7f5dae08611                                         
DEBU[0002] GET https://registry-1.docker.io/v2/library/fedora/manifests/sha256:1b9bfb4e634dc1e5c19d0fa1eb2e5a28a5c2b498e3d3e4ac742bd7f5dae08611                               
DEBU[0003] IsRunningImageAllowed for image docker:docker.io/library/fedora:latest      
DEBU[0003]  Using default policy section                                               
DEBU[0003]  Requirement 0: allowed                                                     
DEBU[0003] Overall: allowed                                                            
DEBU[0003] Downloading /v2/library/fedora/blobs/sha256:422dc563ca3260ad9ef5c47a1c246f5065d7f177ce51f4dd208efd82967ff182                                                       
DEBU[0003] GET https://registry-1.docker.io/v2/library/fedora/blobs/sha256:422dc563ca3260ad9ef5c47a1c246f5065d7f177ce51f4dd208efd82967ff182                                   
```

which means that you're correct.

rhatdan:
Ok that is what we want.

rhatdan:
@rh-atomic-bot r+

ripcurld0:
thank you @rhatdan

PR #423, opened 2018-01-29
Fix typo then->than

ripcurld0:


rhatdan:
Just merging this since this is a no brainer.
@rh-atomic-bot r+

PR #426, opened 2018-01-29
format should override quiet for images

umohnani8:
quiet was overriding format, but we want format to override quiet
if both the flags are set for buildah images.

Fixes issue https://github.com/projectatomic/buildah/issues/425


rhatdan:
I think we should conflict the options, so that users know that quiet does not work with format.

umohnani8:
Fixed it to conflict instead.

PR #431, opened 2018-01-30
Fix --runtime-flag for buildah run and buildah bud

umohnani8:
The --runtime-flag flag for buildah run would fail
whenever the global flags of the runtime were passed to it.
Changed it to accept the format [global-flag]=[value] where
global-flag would be converted to --[global-flag] in the code.

fixes https://github.com/projectatomic/buildah/issues/428

### DIFF
--- a/cmd/buildah/run.go
+++ b/cmd/buildah/run.go
@@ -73,10 +73,15 @@ func runCmd(c *cli.Context) error {
 		return errors.Wrapf(err, "error reading build container %q", name)
 	}
 
+	runtimeFlags := []string{}
+	for _, arg := range c.StringSlice("runtime-flag") {
+		runtimeFlags = append(runtimeFlags, "--"+arg)
+	}
+
 	options := buildah.RunOptions{
 		Hostname: c.String("hostname"),
 		Runtime:  c.String("runtime"),
-		Args:     c.StringSlice("runtime-flag"),
+		Args:     runtimeFlags,
 	}
 
 	if c.IsSet("tty") {
--- a/docs/buildah-run.md
+++ b/docs/buildah-run.md
@@ -26,7 +26,9 @@ The *path* to an alternate OCI-compatible runtime.
 
 Adds global flags for the container runtime. To list the supported flags, please
 consult manpages of your selected container runtime (`runc` is the default
-runtime, the manpage to consult is `runc(8)`)
+runtime, the manpage to consult is `runc(8)`).
+Note: Do not pass the leading `--` to the flag. To pass the runc flag `--log-format json`
+to buildah run, the option given would be `--runtime-flag log-format=json`.
 
 **--tty**
 
@@ -49,7 +51,9 @@ buildah run containerID -- ps -auxw
 
 buildah run containerID --hostname myhost -- ps -auxw
 
-buildah run containerID --runtime-flag --no-new-keyring -- ps -auxw
+buildah run --runtime-flag log-format=json containerID /bin/bash
+
+buildah run --runtime-flag debug containerID /bin/bash
 
 buildah run --tty containerID /bin/bash
 
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -196,6 +196,11 @@ func budCmd(c *cli.Context) error {
 		return errors.Wrapf(err, "error building system context")
 	}
 
+	runtimeFlags := []string{}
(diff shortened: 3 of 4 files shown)


docs/buildah-run.md
```diff
@@ -26,7 +26,8 @@ The *path* to an alternate OCI-compatible runtime.
 
 Adds global flags for the container runtime. To list the supported flags, please
 consult manpages of your selected container runtime (`runc` is the default
-runtime, the manpage to consult is `runc(8)`)
+runtime, the manpage to consult is `runc(8)`).
+If the global flag takes in a string, it will be `--runtime-flag global-flag=value`.
```
rhatdan: Add something like.  Make sure you do not pass the `--`.  If you want to pass --debug to runc, then the command to buildah would be `--runtime-flag debug`

umohnani8: fixed.

cmd/buildah/run.go
```diff
@@ -73,10 +73,15 @@ func runCmd(c *cli.Context) error {
 		return errors.Wrapf(err, "error reading build container %q", name)
 	}
 
+	runtimeFlags := []string{}
+	for _, arg := range c.StringSlice("runtime-flag") {
```
TomSweeneyRedHat: Can we check the arg and add the '--' only if not present?  I.e. if a user wanted to do buildah run --runtime-flag '--debug' that should be OK me thinks.

umohnani8: That will not work because the CLI tool would think of --debug as another flag and will complain that --runtime-flag does not have a value.

docs/buildah-run.md
TomSweeneyRedHat: But if the prepending if not provided is a pain in the tookus, then +1 on Dan's thought.  Perhaps:

Do not include the leading dashes (--) for the runc arguments.  For instance to set the log format to json, the global flag would be "log-format=json".

cmd/buildah/run.go
TomSweeneyRedHat: Gotta love the CLI tool, OK scratch that thought then.

umohnani8: yup, I added a note in the man page and fixed the examples there. So should be pretty obvious on how to work with the `--runtime-flag` flag.

docs/buildah-run.md
```diff
@@ -26,7 +26,9 @@ The *path* to an alternate OCI-compatible runtime.
 
 Adds global flags for the container runtime. To list the supported flags, please
 consult manpages of your selected container runtime (`runc` is the default
-runtime, the manpage to consult is `runc(8)`)
+runtime, the manpage to consult is `runc(8)`).
+Note: Do not pass the leading `--` to the flag. If you want to pass `--log-format json`
```
TomSweeneyRedHat: Nit of a nit.  I know I've added some, but I'd really rather not have the word 'you' in a man page.  (Note to self, scrub man pages).  Can I be a bear and ask that you change this to:
"To pass the runc flag `--log-format json` to `buildah run`,  the option given would be `--runtime-flag log-format=json`.

TomSweeneyRedHat: My suggestion would replace the line starting with "If you want to pass", the preceding sentence is fine.

TomSweeneyRedHat:
One more parting shot.  It looks like we have this flag in buildah bud too, does it work there?

docs/buildah-run.md
umohnani8: fixed.

docs/buildah-run.md
```diff
@@ -26,7 +26,9 @@ The *path* to an alternate OCI-compatible runtime.
 
 Adds global flags for the container runtime. To list the supported flags, please
 consult manpages of your selected container runtime (`runc` is the default
-runtime, the manpage to consult is `runc(8)`)
+runtime, the manpage to consult is `runc(8)`).
+Note: To pass the runc flag `log-format json` to buildah run, the option given
```
TomSweeneyRedHat: Let me see if I can get on @umohnani8 PITA list.  Can you add the '--' before the first log-format?

Note: To pass the runc flag `--log-format json` to buildah run, the option given

(Tom now dives for cover).

umohnani8: Haha no, my fault should have thought about that. Fixed now.

umohnani8:
Made the same fix to buildah bud as well.

rhatdan:
@rh-atomic-bot r+

PR #442, opened 2018-02-01
Add new line after executing template for buildah inspect

umohnani8:
No new line was returned when using the --format flag for buildah
inspect.

Fixes https://github.com/projectatomic/buildah/issues/427

### DIFF
--- a/cmd/buildah/inspect.go
+++ b/cmd/buildah/inspect.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"text/template"
 
@@ -96,7 +97,13 @@ func inspectCmd(c *cli.Context) error {
 	}
 
 	if c.IsSet("format") {
-		return t.Execute(os.Stdout, buildah.GetBuildInfo(builder))
+		if err := t.Execute(os.Stdout, buildah.GetBuildInfo(builder)); err != nil {
+			return err
+		}
+		if terminal.IsTerminal(int(os.Stdout.Fd())) {
+			fmt.Println()
+		}
+		return nil
 	}
 
 	enc := json.NewEncoder(os.Stdout)
--- a/cmd/buildah/images.go
+++ b/cmd/buildah/images.go
@@ -14,6 +14,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
+	"golang.org/x/crypto/ssh/terminal"
 )
 
 type jsonImage struct {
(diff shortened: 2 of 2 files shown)


umohnani8:

cmd/buildah/inspect.go
```diff
@@ -96,7 +97,12 @@ func inspectCmd(c *cli.Context) error {
 	}
 
 	if c.IsSet("format") {
-		return t.Execute(os.Stdout, buildah.GetBuildInfo(builder))
+		err := t.Execute(os.Stdout, buildah.GetBuildInfo(builder))
```
TomSweeneyRedHat: Does the error case look OK without adding a println?

umohnani8: you mean without adding extra info to the returned error?

TomSweeneyRedHat: Ooops, sorry, too many reviews at once.  I was thinking the error would be printed here and might not have a carriage return as it should before the next shell prompt.  However the error handling code handles that and no issue there.  nvm me.

cmd/buildah/inspect.go
```diff
@@ -96,7 +97,12 @@ func inspectCmd(c *cli.Context) error {
...
+		if err != nil {
```
TomSweeneyRedHat: But before I leave, I'll hit you with one of Dan's.  Can this change to:
```
if err := t.Execute(os.Stdout, buildah.GetBuildInfo(builder)); err !=nil {
    return err
}
```

umohnani8: fixed.

TomSweeneyRedHat:
Change LGTM, Red Hat CI is unhappy again and I'm sure due to unrelated issues.

TomSweeneyRedHat:
bot, retest this please

cmd/buildah/inspect.go
```diff
@@ -96,7 +97,11 @@ func inspectCmd(c *cli.Context) error {
 	}
 
 	if c.IsSet("format") {
-		return t.Execute(os.Stdout, buildah.GetBuildInfo(builder))
+		if err := t.Execute(os.Stdout, buildah.GetBuildInfo(builder)); err != nil {
+			return err
+		}
+		fmt.Println()
```
rhatdan: Do we want this if this is not a tty?  IE If I wanted to pype the output out, could this cause issues?

umohnani8: Just tested it with pipe and writing to a file with ">" and its working fine.

rhatdan: Well it works with a carriage return.

If I do a 
if [ "foobar" == `buildah inspect --format ''{{ .ID }}" ]; then 
...
fi
 
I would bet it would fail because buildah inspect --format will "return foobar\n"

umohnani8: This is how its being done in podman as well for templates. Not sure how else to return a new line.

rhatdan: We could do a terminal check before printing the extra line. since this is just for human beings

umohnani8: Added the terminal check to print the extra line for us human beings XD

rhatdan:
@rh-atomic-bot r+

PR #445, opened 2018-02-05
[WIP] Add support for a few missing options from buildah run

rhatdan:
--add-host
     --cgroup-parent
     --cpu-period
     --cpu-quota
     --cpu-shares
     --cpuset-cpus
     --cpuset-mems
     --memory
     --memory-swap
     --security-opt
     --ulimit

### DIFF
--- a/cmd/buildah/run.go
+++ b/cmd/buildah/run.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/docker/go-units"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	"github.com/projectatomic/buildah"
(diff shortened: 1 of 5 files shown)


rhatdan:
On the plane flight back from Europe, I was adding a bunch of options needed for buildah run and buildah build.  This is how far I got,  If someone would like to take this patch over or break the patch into several smaller patches, that would be welcome.

rhatdan:
I currently commented out support for seccomp, since this will be much more complicated.

umohnani8:
Taking this over.

TomSweeneyRedHat:
@umohnani8 there are a number of trello cards set up for these.  Please grab any that you're going to work on.  I'm on most at the moment, but am not actively working them, I probably should remove myself from them.

Also I'd suggest doing smaller patches and we'll need to also update "podman build".

umohnani8:
@TomSweeneyRedHat yeah, I plan to break this into smaller patches. And will add myself to the trello cards. Thanks for the info :)

rhatdan:
Closing since @umohnani8 has taken it over.

PR #447, opened 2018-02-05
Fix md2man issues

umohnani8:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,12 @@ validate:
 
 .PHONY: install.tools
 install.tools:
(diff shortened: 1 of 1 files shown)


rhatdan:
I have opened an issue with upstream go-md2man package
https://github.com/cpuguy83/go-md2man/issues/34
It looks like it does not like the way we are doing tables.

rhatdan:
This is fine for now, but I would like to revert these once this bug gets fixed.

umohnani8:
@rhatdan Okay, I will open an issue to keep track of this.
Can we get this in for now, so we can test other PRs?

rhatdan:
@rh-atomic-bot r+

PR #452, opened 2018-02-06
Fix unintended reversal of the ignoreUnrecognizedInstructions flag

### LINKED ISSUE #451: Buildah bud ignores Dockerfile non-keywords by default
**Description**

A Dockerfile with lines containing syntax errors - lines starting with non-keywords - does not cause an error and halt the build. Instead it skips the offending line.

Also I think a simple change of setting ignoreUnrecognizedInstructions to false is not the fix. 

i.e. that would have the same effect as commenting out:
//      if !b.ignoreUnrecognizedInstructions {
//              logrus.Debugf("+(UNIMPLEMENTED?) %#v", step)
//              return nil
//      }

But this means that the error output is too verbose because there is no check for log-level.  We probably need to have a if !debug type of check and just output a simple "Buildah Error: unknown instructions: <non-keyword>"

1.  Create a Dockerfile that has commands that need to be run without the RUN keyword at the start. Or a file to be copied without the COPY keyword. 
2.  buildah bud .
3.

Build finishes fast. No instructions are executed. All skipped.
I received empty images because dnf commands were not getting run.

Build error: Unknown instruction: DNF

```
Built from HEAD
```

```
buildah version 0.8 (image-spec 1.0.0, runtime-spec 1.0.0)

```

```
Fedora release 27 (Twenty Seven)

```

**Output of `uname -a`:**

```
Linux localhost.localdomain 4.13.9-300.fc27.x86_64 #1 SMP Mon Oct 23 13:41:58 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

```

[storage]

# Default Storage Driver
driver = "overlay"

# Temporary storage location

[storage.options]
additionalimagestores = [
]

size = ""

```

nalind:
We were interpreting the ignoreUnrecognizedInstructions incorrectly, so fix that, and call out the unrecognized instruction keyword in the error message (or debug message, if we're ignoring it).

Should fix #451.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -389,12 +389,12 @@ func (b *Executor) Run(run imagebuilder.Run, config docker.Config) error {
 // UnrecognizedInstruction is called when we encounter an instruction that the
 // imagebuilder parser didn't understand.
 func (b *Executor) UnrecognizedInstruction(step *imagebuilder.Step) error {
-	if !b.ignoreUnrecognizedInstructions {
-		logrus.Debugf("+(UNIMPLEMENTED?) %#v", step)
+	if b.ignoreUnrecognizedInstructions {
+		logrus.Debugf("+(UNIMPLEMENTED KEYWORD %q) %#v", step.Command, step)
 		return nil
 	}
-	logrus.Errorf("+(UNIMPLEMENTED?) %#v", step)
-	return errors.Errorf("Unrecognized instruction: %#v", step)
+	logrus.Errorf("+(UNIMPLEMENTED KEYWORD %q?) %#v", step.Command, step)
+	return errors.Errorf("Unrecognized instruction %q: %#v", step.Command, step)
 }
 
 // NewExecutor creates a new instance of the imagebuilder.Executor interface.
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM, Red Hat CI seems unhappy with unrelated network issues to the registry.

nalind:
bot, retest this please

rhatdan:
@rh-atomic-bot r+

nalind:
@rh-atomic-bot retry

ipbabble:
May want to have different levels of output for different loglevels. 
```
func (b *Executor) UnrecognizedInstruction(step *imagebuilder.Step) error {
        if b.ignoreUnrecognizedInstructions {
                logrus.Debugf("+(UNIMPLEMENTED?) %#v", step)
                return nil
        }
        if logrus.GetLevel() == logrus.ErrorLevel {
// no DebugF
                return errors.Errorf("Unknown instruction: %v", step.Command)
        } else if logrus.GetLevel() == logrus.DebugLevel {
// add DebugF
                logrus.Debugf("Unknown instruction: %#v", step)
                return errors.Errorf("Unknown instruction: %#v", step)
        } else {
// what else should it check?
                return errors.Errorf("Unknown instruction: %#v", step)
        }

}
````

PR #456, opened 2018-02-07
Implement the following flags for buildah run

umohnani8:
Implement the following flags for buildah run

        --cpu-period
        --cpu-quota
        --cpu-shares
        --cpuset-cpus
        --cpuset-mems
        --memory
        --memory-swap
        --ulimit


umohnani8:

contrib/completions/bash/buildah
```diff
@@ -357,20 +357,11 @@ return 1
      --build-arg
      -c
      --cert-dir
```
TomSweeneyRedHat: are we removing these from buildah bud?  Don't we need to add them to buildah run?

docs/buildah-bud.md
```diff
@@ -13,11 +13,6 @@ build context directory.  The build context directory can be specified as the
 to a temporary location.
 
 ## OPTIONS
-**--add-host**=[]
```
TomSweeneyRedHat: OK, I'm officially confused.  What's the plan here are we removing these options from 'buildah bud'?  I'm not sure why, but if @nalind is hip, then works for me.  If we are removing them, we need to update the bud.go file too.

Also I thought you were adding these to the buildah run command, so I'd expect these to be added to the buildah run man page.

TomSweeneyRedHat: Also if we're removing from buildah bud, we'll need to adjust podman build too as it passes these along at the moment.

umohnani8: Sorry for the confusion! My previous PR added these options to the docs and completions of buildah run and bud, but after trying to implement them for buildah bud, I realised that the code for bud is very different from run and it is going to be harder and take longer to implement these same options for buildah bud.
So I decided to take it out of the documentation and completions for buildah bud just so I don't block the release whenever that happens in the near future.
I changed the commit message but forgot to update it here, sorry!
The PR that added the docs and completions stuff is https://github.com/projectatomic/buildah/pull/446.

TomSweeneyRedHat: @umohnani8 thanks for the 411, I'm easily confuzuled before my first cup of tea.

run.go
```diff
@@ -160,6 +186,35 @@ func (b *Builder) setupMounts(mountPoint string, spec *specs.Spec, optionMounts
...
+		if ul, err = units.ParseUlimit(u); err != nil {
+			return errors.Wrapf(err, "ulimit option %q requires name=SOFT:HARD, failed to be parsed", u)
+		}
+
+		g.AddProcessRlimits("RLIMIT_"+strings.ToUpper(ul.Name), uint64(ul.Hard), uint64(ul.Soft))
+	}
+	return nil
+}
+
+func addHostsToFile(hosts []string) error {
```
TomSweeneyRedHat: Does this add entries to the Host file on the host or the container?

TomSweeneyRedHat: Thanks and I need to remember that.

TomSweeneyRedHat:
LGTM, but would like an eyeball from @rhatdan or @nalind

umohnani8:
Reworking this whole PR.Closing for now.

PR #460, opened 2018-02-08
COPR enablement

baude:
For COPR builds, we will use a a slightly modified spec and the
makesrpm method over SCM builds so we can have dynamic package
names.


baude:
bot, retest this please

rhatdan:
@rh-atomic-bot r+

PR #470, opened 2018-02-12
Removing docs and completions for run options

umohnani8:
Figured that these options need to be in from and bud instead.
Removed the options from the documentation of run and bud for now.
Need to get this in before we cut for release.


umohnani8:
@rhatdan PTAL

rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+

Issue #474, opened 2018-02-13
error messsage not useful when image not found on RHEL


gouyang:
```
# buildah from busybox
unknown: Not Found

# buildah version
Version:       0.12
Go Version:    go1.8.3
Image Spec:    1.0.0
Runtime Spec:  1.0.0
Git Commit:    e073df1
Built:         Tue Feb 13 15:51:59 2018
OS/Arch:       linux/amd64
```

ripcurld0:
@gouyang can't reproduce it.

gouyang:
@ripcurld0 are you reproducing it on RHEL?

ripcurld0:
No, `Fedora 27`

gouyang:
The issue s reported on RHEL, it cannot see this error on Fedora.

gouyang:
It's not happened on `centos`.  it happened on RHEL because of it only search `registry.access.redhat.com` by default. 

```
# cat /etc/containers/registries.conf 
....
[registries.search]
registries = ['registry.access.redhat.com']
```

If add `docker.io` to the registries.search or remove `registry.access.redhat.com`, the image can be pulled and no error occurs. 

The issue is complaining the error message `unknown: Not Found` not very clean to user, not why it happened.

ripcurld0:
@gouyang this info was very helpful.
I could successfully reproduce the problem.

rhatdan:
@gouyang Did the default registries.conf on RHEL not have docker.io, or did you modify it?  

Are you just reporting that you think we should have a better error message to tell the user the image is not found at any of the specified registries?

rhatdan:
On Fedora I see this error when I have a bogus image

```
 buildah from foobar
no such image "foobar": image not known
```
If I remove all registies except for the redhat registry, I get the same error.
```
[registries.search]
#registries = ['docker.io', 'registry.fedoraproject.org', 'registry.access.redhat.com']
registries = ['registry.access.redhat.com']
```

gouyang:
The buildah is built from master.
```
# buildah version
Version:       0.12
Go Version:    go1.8.3
Image Spec:    1.0.0
Runtime Spec:  1.0.0
Git Commit:    d59af12
Built:         Wed Feb 14 06:45:04 2018
OS/Arch:       linux/amd64
```

Test on RHEL7.4, with `registries = ['registry.access.redhat.com'] ` in /etc/containers/registries.conf: 
```
# ./buildah from gouyang/foobar
unknown: Not Found
[...]

PR #485, opened 2018-02-18
Display full error string instead of just status

TomSweeneyRedHat:

Show the full error for commit and pull errors rather than just the error code.   As noted in #474 the error seen if the requested registry did not have an image was:

```
# buildah from busybox
unknown: Not Found
```

With this fix in place it's a lot more verbose, but perhaps not as pretty.  More adjustments might be made in the containers side, or even here.  New errors look like the following.  The first is a bad authentication to a private image, the second is the new error for the image not found situation line in #474.

```
# /usr/local/bin/buildah from docker.io/tomsweeneyredhat/testing:first
Error determining manifest MIME type for docker://tomsweeneyredhat/testing:first: errors:
denied: requested access to the resource is denied

# /usr/local/bin/buildah from busybox
Error determining manifest MIME type for docker://registry.access.redhat.com/busybox:latest: unknown: Not Found
```

@ripcurld0 thoughts?

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -1,7 +1,6 @@
 package util
 
 import (
-	"fmt"
 	"net/url"
 	"path"
 	"strings"
@@ -14,7 +13,6 @@ import (
 	"github.com/docker/distribution/registry/api/errcode"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"github.com/urfave/cli"
 )
 
 const (
@@ -163,16 +161,10 @@ func AddImageNames(store storage.Store, image *storage.Image, addNames []string)
 func GetFailureCause(err, defaultError error) error {
 	switch nErr := errors.Cause(err).(type) {
 	case errcode.Errors:
-		return cli.NewMultiError([]error(nErr)...)
+		return err
 	case errcode.Error, *url.Error:
 		return nErr
 	default:
-		// HACK: In case the error contains "not authorized" like in
-		//       https://github.com/containers/image/blob/master/docker/docker_image_dest.go#L193-L205
-		// TODO(bshuster): change "containers/images" to return "errcode" rather than "error".
-		if strings.Contains(nErr.Error(), "not authorized") {
-			return fmt.Errorf("unauthorized: authentication required")
-		}
 		return defaultError
 	}
 }
--- a/new.go
+++ b/new.go
@@ -220,7 +220,7 @@ func newBuilder(store storage.Store, options BuilderOptions) (*Builder, error) {
 		// If options.FromImage is set but we ended up
 		// with nil in ref or in img then there was an error that
 		// we should return.
-		return nil, util.GetFailureCause(err, errors.Wrapf(storage.ErrImageUnknown, "no such image %q", options.FromImage))
+		return nil, util.GetFailureCause(err, errors.Wrapf(storage.ErrImageUnknown, "no such image %q in registry", options.FromImage))
 	}
 	image := options.FromImage
 	imageID := ""


ripcurld0:
@TomSweeneyRedHat that's a good start.
Although, I think we should change our strategy: check if the cause is "unauthorized" error and if it is, then return the stripped error. Otherwise, return the entire error.

@mtrmac PTAL to make sure we're on the right path.

Thanks.

rhatdan:
LGTM

util/util.go
```diff
@@ -163,13 +162,17 @@ func AddImageNames(store storage.Store, image *storage.Image, addNames []string)
...
-		return nErr
-	default:
 		// HACK: In case the error contains "not authorized" like in
 		//       https://github.com/containers/image/blob/master/docker/docker_image_dest.go#L193-L205
 		// TODO(bshuster): change "containers/images" to return "errcode" rather than "error".
+		// Also in the default stanza.
+		if strings.Contains(nErr.Error(), "not authorized") {
+			return fmt.Errorf("unauthorized: authentication required")
+		}
+		return (err)
```
bertinatto: Minor nit: I think the parentheses are not necessary.

util/util.go
```diff
@@ -163,13 +162,17 @@ func AddImageNames(store storage.Store, image *storage.Image, addNames []string)
 func GetFailureCause(err, defaultError error) error {
 	switch nErr := errors.Cause(err).(type) {
 	case errcode.Errors:
-		return cli.NewMultiError([]error(nErr)...)
-	case errcode.Error, *url.Error:
-		return nErr
-	default:
 		// HACK: In case the error contains "not authorized" like in
 		//       https://github.com/containers/image/blob/master/docker/docker_image_dest.go#L193-L205
 		// TODO(bshuster): change "containers/images" to return "errcode" rather than "error".
```
ripcurld0: I think we can remove these two comments:
```
// HACK: ....
// TODO(bshuster): ...
```

util/util.go
TomSweeneyRedHat: fixed, thanks!

util/util.go
TomSweeneyRedHat: Zapped, thanks!

util/util.go
```diff
@@ -163,13 +162,13 @@ func AddImageNames(store storage.Store, image *storage.Image, addNames []string)
 func GetFailureCause(err, defaultError error) error {
 	switch nErr := errors.Cause(err).(type) {
 	case errcode.Errors:
-		return cli.NewMultiError([]error(nErr)...)
+		if strings.Contains(nErr.Error(), "not authorized") {
```
mtrmac: The `errcode` package has a public `ErrorCodeUnauthorized` variable; it seems to me that this should be possible to do using the actual API instead of matching strings.

… if `not authorized` is even a value which can be returned here: AFAICS this is intended to match the code before https://github.com/containers/image/pull/410 ; so it definitely can’t apply in the `errcode.Errors` case, and could probably be dropped from the `default` case as well if `containers/image` is new enough.

util/util.go
```diff
@@ -163,13 +162,13 @@ func AddImageNames(store storage.Store, image *storage.Image, addNames []string)
...
+			return fmt.Errorf("unauthorized: authentication required")
+		}
+		return err
 	case errcode.Error, *url.Error:
 		return nErr
 	default:
-		// HACK: In case the error contains "not authorized" like in
-		//       https://github.com/containers/image/blob/master/docker/docker_image_dest.go#L193-L205
-		// TODO(bshuster): change "containers/images" to return "errcode" rather than "error".
 		if strings.Contains(nErr.Error(), "not authorized") {
```
mtrmac: Why remove _only_ the comment? The string matching remains a hack, and it seems to me that the TODO item also remains valuable.

Also, per the other comment, after https://github.com/containers/image/pull/410 this probably does not work any more and needs updating.

util/util.go
TomSweeneyRedHat: Thanks @mtrmac.  I've removed the hacks from both here and the default.  On a bad auth, we're hitting the first case and showing an appropriate error now rather than dropping down into the default.

util/util.go
TomSweeneyRedHat: I've actually removed it completely, I don't think it's necessary anymore, but then again I haven't done my one last test yet....

TomSweeneyRedHat:
With the latest changes:

```
# buildah from  docker.io/tomsweeneyredhat/testing:first
Error determining manifest MIME type for docker://tomsweeneyredhat/testing:first: errors:
denied: requested access to the resource is denied

# buildah from junker
Error determining manifest MIME type for docker://registry.access.redhat.com/junker:latest: unknown: Not Found
```

mtrmac:
ACK, this is a clear improvement. _Maybe_ the “authentication required” case could be tidied up a bit by matching on `errcode.*` values, but this is clearly making the code better, so it would be unreasonable to block this PR on that.

rhatdan:
@rh-atomic-bot r+

PR #486, opened 2018-02-20
Vendor in latest containers/storage

rhatdan:
This fixes a large SELinux bug.  Currently if you do the following
commands

ctr=$(buildah from scratch)
mnt=$(buildah mount $ctr)
dnf install --installroot=$mnt httpd
buildah run $ctr touch /test

The last command fails.  The reason for this is the SELinux labels are getting applied
to the mount point, since it was not being mounted as an overlay file system.

Containers/storage was updated to always mount an overlay even if the lower layer is empty
This then causes the mount point to use a context mount, and changes dnf to not apply
labels.  This change then allows buildah run to create confined containers to run code.

### DIFF
--- a/vendor/github.com/containers/storage/drivers/overlay/overlay.go
+++ b/vendor/github.com/containers/storage/drivers/overlay/overlay.go
@@ -142,14 +142,16 @@ func Init(home string, options []string, uidMaps, gidMaps []idtools.IDMap) (grap
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


rhatdan:
This should fix https://github.com/projectatomic/buildah/issues/482

TomSweeneyRedHat:
LGTM

rhatdan:

PR #490, opened 2018-02-22
Add libseccomp-devel to packages to install

TomSweeneyRedHat:

While doing a quick test on a new VM, noted that we'd not added libseccomp-devel to the install list for the dependencies.  A recent PR added that requirement.


rhatdan:
@rh-atomic-bot r+

PR #493, opened 2018-02-23
Currently buildah run is not handling command options correctly

rhatdan:
This patch will allow commands like

buildah run $ctr ls -lZ /

To work correctly.

### DIFF
--- a/cmd/buildah/run.go
+++ b/cmd/buildah/run.go
@@ -43,12 +43,13 @@ var (
 	}
 	runDescription = "Runs a specified command using the container's root filesystem as a root\n   filesystem, using configuration settings inherited from the container's\n   image or as specified using previous calls to the config command"
 	runCommand     = cli.Command{
-		Name:        "run",
-		Usage:       "Run a command inside of the container",
-		Description: runDescription,
-		Flags:       runFlags,
-		Action:      runCmd,
-		ArgsUsage:   "CONTAINER-NAME-OR-ID COMMAND [ARGS [...]]",
+		Name:           "run",
+		Usage:          "Run a command inside of the container",
+		Description:    runDescription,
+		Flags:          runFlags,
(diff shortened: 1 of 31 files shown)


umohnani8:
LGTM

cmd/buildah/run.go
```diff
@@ -43,12 +43,14 @@ var (
...
-		Flags:       runFlags,
-		Action:      runCmd,
-		ArgsUsage:   "CONTAINER-NAME-OR-ID COMMAND [ARGS [...]]",
+		Name:                   "run",
+		Usage:                  "Run a command inside of the container",
+		Description:            runDescription,
+		Flags:                  runFlags,
+		Action:                 runCmd,
+		ArgsUsage:              "CONTAINER-NAME-OR-ID COMMAND [ARGS [...]]",
+		SkipArgReorder:         true,
```
TomSweeneyRedHat: should we add these to the other commands?

rhatdan: THe skip arg reorder is just for executing commands after the image. So not sure for it.  But the supports short names might be usefull for buildah from/buid, if they support single letter options.

rhatdan:

umohnani8:
"Error determining manifest MIME type for docker://registry.fedoraproject.org/fedora-minimal:latest: unauthorized: authentication required" -  fedoraproject.org is down?

rhatdan:
bot, retest this please

rhatdan:
@rh-atomic-bot r+

cgwalters:
FWIW I was previously doing:

`buildah config ${container} --entrypoint blah` etc., and that broke after this PR.

But eh, after reading now I understand why the change was made.

cgwalters:
(It's tricky since the rest of my code is lots of `buildah run ${container} arguments...`)

Issue #440, opened 2018-02-01
support for multiple FROM instructions


vbatts:
**Description**
Upstream now has multiple image builds.
For example, working with the [Grafeas project ](https://github.com/grafeas/grafeas) their [Dockerfile](https://github.com/grafeas/grafeas/blob/master/Dockerfile) does a build container, and then copies the binary to a minimal image.

**Steps to reproduce the issue:**
1. git clone https://github.com/grafeas/grafeas
2. cd grafeas
3. sudo buildah bud -t grafeas .

**Describe the results you received:**
```
error building: error determining starting point for build: multiple FROM statements are not supported
```

**Describe the results you expected:**
an image

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-0.11-1.git6bad262.fc26.x86_64
```

**Output of `buildah version`:**

```
Version:       0.11
Go Version:    go1.8.5
Image Spec:    1.0.0
Runtime Spec:  1.0.0
Git Commit:    6bad262
Built:         Wed Jan 17 08:49:09 2018
OS/Arch:       linux/amd64
```

**Output of `cat /etc/*release`:**

```
Fedora release 26 (Twenty Six)
NAME=Fedora
VERSION="26 (Workstation Edition)"
ID=fedora
VERSION_ID=26
PRETTY_NAME="Fedora 26 (Workstation Edition)"
ANSI_COLOR="0;34"
CPE_NAME="cpe:/o:fedoraproject:fedora:26"
HOME_URL="https://fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=26
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=26
PRIVACY_POLICY_URL=https://fedoraproject.org/wiki/Legal:PrivacyPolicy
VARIANT="Workstation Edition"
VARIANT_ID=workstation
Fedora release 26 (Twenty Six)
Fedora release 26 (Twenty Six)
```

**Output of `uname -a`:**

```
Linux chacha.usersys.redhat.com 4.14.14-200.fc26.x86_64 #1 SMP Fri Jan 19 13:27:06 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

**Output of `cat /etc/containers/storage.conf`:**

```
[...]

PR #495, opened 2018-02-24
Add multi-stage builds support

bertinatto:
This PR adds multi-stage builds support to buildah.

Addresses #440.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	is "github.com/containers/image/storage"
@@ -113,6 +114,7 @@ type BuildOptions struct {
 // Executor is a buildah-based implementation of the imagebuilder.Executor
 // interface.
 type Executor struct {
+	index                          int
 	name                           string
 	named                          map[string]*Executor
 	store                          storage.Store
@@ -146,14 +148,18 @@ type Executor struct {
 }
 
 // withName creates a new child executor that will be used whenever a COPY statement uses --from=NAME.
-func (b *Executor) withName(name string) *Executor {
+func (b *Executor) withName(name string, index int) *Executor {
 	if b.named == nil {
 		b.named = make(map[string]*Executor)
 	}
 	copied := *b
+	copied.index = index
 	copied.name = name
 	child := &copied
 	b.named[name] = child
+	if idx := strconv.Itoa(index); idx != name {
+		b.named[idx] = child
+	}
 	return child
 }
 
@@ -354,7 +360,7 @@ func (b *Executor) Copy(excludes []string, copies ...imagebuilder.Copy) error {
 			if strings.HasPrefix(src, "http://") || strings.HasPrefix(src, "https://") {
 				sources = append(sources, src)
 			} else if len(copy.From) > 0 {
-				if other, ok := b.named[copy.From]; ok {
+				if other, ok := b.named[copy.From]; ok && other.index < b.index {
 					sources = append(sources, filepath.Join(other.mountPoint, src))
 				} else {
 					return errors.Errorf("the stage %q has not been built", copy.From)
@@ -686,7 +692,7 @@ func (b *Executor) Build(stages imagebuilder.Stages) error {
 	}
 	var stageExecutor *Executor
 	for _, stage := range stages {
-		stageExecutor = b.withName(stage.Name)
+		stageExecutor = b.withName(stage.Name, stage.Position)
 		if err := stageExecutor.Prepare(stage.Builder, stage.Node, ""); err != nil {
 			return err
 		}
@@ -701,8 +707,8 @@ func (b *Executor) Build(stages imagebuilder.Stages) error {
 // BuildDockerfiles parses a set of one or more Dockerfiles (which may be
 // URLs), creates a new Executor, and then runs Prepare/Execute/Commit/Delete
 // over the entire set of instructions.
-func BuildDockerfiles(store storage.Store, options BuildOptions, dockerfile ...string) error {
-	if len(dockerfile) == 0 {
+func BuildDockerfiles(store storage.Store, options BuildOptions, paths ...string) error {
+	if len(paths) == 0 {
 		return errors.Errorf("error building: no dockerfiles specified")
 	}
 	var dockerfiles []io.ReadCloser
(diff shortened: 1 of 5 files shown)


rhatdan:
@nalind PTAL

TomSweeneyRedHat:
Travis CI isn't happy @bertinatto 
```
not ok 13 bud-from-multiple-files-two-froms
# (in test file ./bud.bats, line 51)
#   `cmp $root/Dockerfile1 ${TESTSDIR}/bud/from-multiple-files/Dockerfile1.scratch' failed with status 2
```

bertinatto:
Right, I have a question about this failing test (`bud-from-multiple-files-two-froms`).

In a multi-stage build, each `FROM` statement will result in a different stage (i.e., container). I believe this specific test made sense so far, but from now on I'd assume that files added to a stage shouldn't end up in the final stage. 

What do you think about this?

rhatdan:
I am no expert on this. but isn't multi-stage the equivalent of doing two different Dockerfile, where the second Dockerfile from is from the artifact of the first Dockerfile.  Anything that is in the image created by the first `Dockerfile` would end up in the second unless it was at some point removed?

kbaegis:
@rhatdan Here's an example:
https://github.com/gentoo/gentoo-docker-images/blob/master/portage.Dockerfile

This is all based around the flexibility of needing to use one images utilities to bootstrap the toolchain for another (i.e. - using alpine gpg to validate or busybox to copy/untar).

bertinatto:
@rhatdan, AFAIU the idea is that we can selectively copy artifacts from one stage to another, leaving the final image without unnecessary artifacts.

Using the example of #440, the first `FROM` starts a new stage with the base image `golang` where the binary `grafeas-server` is built. The second stage is built using `alpine` as base image and only the file `grafeas-server` is copied to it, leaving behind the Go SDK.

rhatdan:
Do these workloads work with this patch?

bertinatto:
Yup, but unfortunately multi-stage builds support changes the way `buildah` (and `imagebuilder`) work with multiple Dockerfiles (i.e.,  multiple FROM statements).

rhatdan:
@nalind PTAL

rhatdan:
@ipbabble @TomSweeneyRedHat Could you try out a couple of Dockerfiles against this to make sure it works fairly well.

I would like to get this merged along with SHELL Support so we can cut a new version of Buildah.

TomSweeneyRedHat:
FWIW, I tested using both of the below Dockerfiles and these worked.  I tried a couple of other more complex Dockerfiles that I found on the net and they had failures, but the same failure happened in Docker too.  I don't believe it was a Buildah issue.  If you've a pointer to other multi-stage Dockerfiles, happy to try them.
 
```
FROM ubuntu
RUN whoami
FROM ubuntu
RUN echo hi
```
and

```
FROM golang:1.7.3
WORKDIR /go/src/github.com/alexellis/href-counter/
RUN go get -d -v golang.org/x/net/html  
COPY app.go .
RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o app .

FROM alpine:latest  
RUN apk --no-cache add ca-certificates
WORKDIR /root/
COPY --from=0 /go/src/github.com/alexellis/href-counter/app .
CMD ["./app"]  
```

imagebuildah/build.go
```diff
@@ -113,6 +113,8 @@ type BuildOptions struct {
 // Executor is a buildah-based implementation of the imagebuilder.Executor
 // interface.
 type Executor struct {
+	name                           string
```
nalind: Please add a field for the numeric index of the stage which this `Executor` is responsible for building.  We'll need that information to ensure that `COPY` instructions that try to reference this stage or a later stage, can be rejected.

imagebuildah/build.go
```diff
@@ -329,6 +343,12 @@ func (b *Executor) Copy(excludes []string, copies ...imagebuilder.Copy) error {
 		for _, src := range copy.Src {
 			if strings.HasPrefix(src, "http://") || strings.HasPrefix(src, "https://") {
 				sources = append(sources, src)
+			} else if len(copy.From) > 0 {
+				if other, ok := b.named[copy.From]; ok {
```
nalind: We'll need to check that the stage number for `other` is less than the one for `b`.

imagebuildah/build.go
```diff
@@ -647,58 +667,52 @@ func (b *Executor) Commit(ib *imagebuilder.Builder) (err error) {
...
-	for _, this := range node {
-		if err = b.Execute(ib, this); err != nil {
+// over each of the one or more parsed Dockerfiles and stages.
+func (b *Executor) Build(stages imagebuilder.Stages) error {
+	if len(stages) == 0 {
+		errors.New("error building: no stages to build")
+	}
+	var stageExecutor *Executor
+	for _, stage := range stages {
+		stageExecutor = b.withName(stage.Name)
```
nalind: This should take the index of the stage as an argument, since whether a name for the stage is specified with `AS` or not, its contents still need to be accessible by referring to the stage using its numeric index, and we want to know where a given `Executor` is in the list of them.  Converting the number to a string in `withName()` and using it as another key for adding `copied` to the `named` map, since the values are pointers, should be fine.

nalind:
Apologies for the delay.  A couple of notes about stage ordering and numbers, but the rest looks really good.  Thanks!

rhatdan:
@bertinatto Needs a rebase.

bertinatto:
@nalind, thanks for the review. I'll apply your suggestions as soon as I have time (probably over the weekend, if that's not a problem).

@rhatdan, will do, thanks!

ipbabble:
I've been traveling and haven't been able to take a look at this. Sorry.
I've been doing som eof my own testing of complex Dockerfiles.

William Henry
Senior Consulting Software Engineer
Red Hat
Colorado, USA.

imagebuildah/build.go
bertinatto: Done, although I think it was rejecting invalid stages before.

imagebuildah/build.go
bertinatto: Done.

imagebuildah/build.go
bertinatto: Done. Thanks for catching this.

I checked the behaviour of `docker` and it does accept both indexes and names. On the other hand, `imagebuilder` does not. We might report this.

nalind:
LGTM pending CI.

bertinatto:
I had to close/reopen to retest. All green now.

nalind:
Thanks!
@rh-atomic-bot r+

rhatdan:
Time for a new build?  I think this could be buildah 1.0?  @nalind @TomSweeneyRedHat WDYT

bertinatto:
My vote doesn't count, but it would be awesome to have buildah 1.0!

PR #498, opened 2018-02-26
If commonOpts do not exist, we should return rather then segfault

rhatdan:

### DIFF
--- a/run.go
+++ b/run.go
@@ -107,6 +107,10 @@ func addHostsToFile(hosts []string) error {
 }
 
 func addCommonOptsToSpec(commonOpts *CommonBuildOptions, g *generate.Generator) error {
+	if commonOpts == nil {
+		return nil
+	}
+
 	// RESOURCES - CPU
 	if commonOpts.CPUPeriod != 0 {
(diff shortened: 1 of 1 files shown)


rhatdan:
I think this will fix https://github.com/projectatomic/buildah/issues/497

rhatdan:
I know that the logic here is reversed.  But the plane I am on right now will not allow me to push...  Will update PR when I land.

run.go
```diff
@@ -107,6 +107,10 @@ func addHostsToFile(hosts []string) error {
 }
 
 func addCommonOptsToSpec(commonOpts *CommonBuildOptions, g *generate.Generator) error {
```
ripcurld: I wonder if changing from `*CommonBuildOptions` to `CommonBuildOptions` (removing the asterisk) can solve the problem too:
1) It seems to me that we set Linux resources using it but we don't change `commonOpts` during that process so why using the pointer?
2) This will solve the checking if `nil`.
3) Empty `commonOpts` will set all its properties to `0` or to empty strings (depends on the type). So we don't need to initialize it.

mheon:
LGTM

rhatdan:

PR #501, opened 2018-02-27
Add selinux test from #486

TomSweeneyRedHat:

Adds a test to the baseline test suite to make sure an selinux issue is no more.  This is not run by the CI tests systems, only by hand to check on the health of a new kit.


rhatdan:
LGTM

PR #506, opened 2018-03-02
shallowCopy: avoid a second read of the container's layer

### LINKED ISSUE #502: commit fails with "error determining uncompressed digest for blob" when using vfs
I've tried to build an image using the vfs storage driver (that could be used from an user namespace).

```
ctr1=`buildah --storage-driver vfs from --pull fedora:27`
buildah --storage-driver vfs run $ctr1 -- dnf  upgrade -y
buildah --storage-driver vfs run $ctr1 -- dnf install -y lighttpd
buildah --storage-driver vfs config $ctr1 --annotation "com.example.build.host=fedora-27"
buildah --storage-driver vfs config $ctr1 --cmd "/usr/sbin/lighttpd -D -f /etc/lighttpd/lighttpd.conf"
buildah --storage-driver vfs config $ctr1 --port 80
buildah --storage-driver vfs commit $ctr1 giuseppe/lighttpd
```

```
error committing container "fedora-working-container-7" to "giuseppe/lighttpd": error copying layer and metadata: error committing new image "containers-storage:[vfs@/var/lib/containers/storage+/var/run/containers/storage:overlay.override_kernel_check=true]docker.io/giuseppe/lighttpd:latest": error determining uncompressed digest for blob "sha256:c8be87302b4b2a1c84a22a29f43e6b45ca68dff1909f2cce967f372e7ed28b49"
```

"buildah commit" succeded

using buildah from upstream, commit cb42905a7f5e91a4407b090947cd0ea0d375ada0

```
# cat /etc/os-release 
NAME=Fedora
VERSION="27 (Workstation Edition)"
ID=fedora
VERSION_ID=27
PRETTY_NAME="Fedora 27 (Workstation Edition)"
ANSI_COLOR="0;34"
CPE_NAME="cpe:/o:fedoraproject:fedora:27"
REDHAT_BUGZILLA_PRODUCT_VERSION=27
REDHAT_SUPPORT_PRODUCT_VERSION=27
VARIANT="Workstation Edition"
VARIANT_ID=workstation

```

**Output of `uname -a`:**
```
# uname -a
Linux helium 4.14.16-300.fc27.x86_64 #1 SMP Wed Jan 31 19:24:27 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

```
# cat /etc/containers/storage.conf

[storage]

# Default Storage Driver
driver = "overlay"

# Temporary storage location

[storage.options]
additionalimagestores = [
]

size = ""

```

nalind:
Avoid reading the container's layer a second time in `shallowCopy()` (after the first time it's read, in `NewImageSource()`).

Our second copy of the diff isn't going to match the first one if it's of any serious size, because when `NaiveDiff()` is used to generate the diff, whiteout entries have their times set to "now", rather than a
fixed value, which causes both the uncompressed and compressed digests to differ from what `NewImageSource()` records in the manifest.

Instead, use the generic manifest-reading API that the image library now provides to read the layer info list and the config blob info, and use their values to `PutBlob()` them to the destination image.

In the medium-term, we can probably drop `shallowCopy()` completely, since the newer version of the image library has a more reliable `HasBlob()`, which should save us the time that `shallowCopy()` was originally meant to save, but we should probably consider the use of compression for this code path as part of that, too.

Fixes #502.

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -1,20 +1,18 @@
 package buildah
 
 import (
-	"bytes"
 	"fmt"
 	"io"
-	"syscall"
 	"time"
 
 	cp "github.com/containers/image/copy"
+	"github.com/containers/image/manifest"
 	"github.com/containers/image/signature"
 	is "github.com/containers/image/storage"
 	"github.com/containers/image/transports"
 	"github.com/containers/image/types"
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/archive"
-	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	"github.com/projectatomic/buildah/util"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


giuseppe:
great, I've just tried and it solves the issue I've reported!

rhatdan:
@rh-atomic-bot r+
Should we cut a new release?  Anything else we want to get in?

PR #508, opened 2018-03-05
Remove shallowCopy(), which shouldn't be saving us time any more

nalind:
Remove `shallowCopy()`, which shouldn't be saving us time any more since `containers/image`'s storage transport's `HasBlob()` implementation was last updated.


rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #510, opened 2018-03-08

umohnani8:
Adds feature to pull compressed docker-archive files


umohnani8:

TomSweeneyRedHat:

rhatdan:
LGTM
@rh-atomic-bot r+

rhatdan:
@umohnani8 Could you give us a use case where this is important?

umohnani8:
@rhatdan Actually I don't think buildah pulls images saved to docker-archive or oci-archive. It would be cool to add this though so we can do "buildah from docker-archive:alp.tar" and it will create a container from the image in alp.tar. And buildah push supports all the various transports.

PR #514, opened 2018-03-10
WIP - Ignore builtin image volumes when run

TomSweeneyRedHat:

To satisfy the Trello card https://trello.com/c/WpA1TIXP, we'll ignore builtin volumes when doing a run if the user requests.  

I'm not 100% sure if I'm on the right track for this as what CRIO/podman did was slightly different than what Buildah is doing and the description in the card is kind of light.  Due to that, this is a WIP at the moment.  I need updates to man pages and tests before merging.

### DIFF
--- a/run.go
+++ b/run.go
@@ -43,6 +43,8 @@ const (
 
 // RunOptions can be used to alter how a command is run in the container.
 type RunOptions struct {
+	// BuiltInVolume the setting for handling builtin image volumes.
+	BuiltInVolume string
 	// Hostname is the hostname we set for the running container.
 	Hostname string
 	// Runtime is the name of the command to run.  It should accept the same arguments that runc does.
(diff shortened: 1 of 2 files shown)


run.go
```diff
@@ -205,36 +207,38 @@ func (b *Builder) setupMounts(mountPoint string, spec *specs.Spec, optionMounts
...
-			if err = os.MkdirAll(volumePath, 0755); err != nil {
-				return errors.Wrapf(err, "error creating directory %q for volume %q in container %q", volumePath, volume, b.ContainerID)
-			}
-			if err = label.Relabel(volumePath, b.MountLabel, false); err != nil {
-				return errors.Wrapf(err, "error relabeling directory %q for volume %q in container %q", volumePath, volume, b.ContainerID)
-			}
-			srcPath := filepath.Join(mountPoint, volume)
-			if err = copyWithTar(srcPath, volumePath); err != nil && !os.IsNotExist(err) {
-				return errors.Wrapf(err, "error populating directory %q for volume %q in container %q using contents of %q", volumePath, volume, b.ContainerID, srcPath)
+	if (builtInVolume != "ignore") {
```
bertinatto: Would it make sense to make this a `bool` instead?

rhatdan:
Follow the standard set in podman, which is based off the standard set in CRI-O. I believe.  Basically I believe there are three options.  Builtin, Tmpfs, and Ignore.  Or something like that.

rhatdan:
I have talked to @nalind about this, and it could be a little more complicated in buildah.

rhatdan:
@TomSweeneyRedHat @umohnani8 Isn't this fixed in current buildah?

rhatdan:
@umohnani8 Do we have support for this?

TomSweeneyRedHat:
I've messed up my sandbox and this is way outta date, will restart when I get back to it.

PR #517, opened 2018-03-16
Add support for shell

rhatdan:
Also vendor in the latest imagebuilder code and all the packages
that come with it.


config.go
```diff
@@ -343,6 +343,19 @@ func (b *Builder) SetWorkDir(there string) {
...
 
+// Shell returns the default shell for running commands in the
+// container, or in a container built using an image built from this container.
+func (b *Builder) Shell() []string {
+	return b.Docker.Config.Shell
+}
+
+// SetShell sets the default shell for running
+// commands in the container, or in a container built using an image built from
+// this container.
```
nalind: Add a note here like the one we have for `SetHostname()`, so that callers will be less surprised when this doesn't affect the contents of OCI images.

contrib/rpm/buildah.spec
```diff
@@ -49,6 +49,7 @@ BuildRequires:  make
 Requires:       runc >= 1.0.0-6
 Requires:       container-selinux
 Requires:       skopeo-containers
+Requires:       ostree-libs
```
nalind: Is this from a different PR?

docs/buildah-config.md
```diff
@@ -65,6 +65,11 @@ its OS is kept, otherwise the host's OS's name is recorded.
 Specifies a *port* to expose when running containers based on any images which
 will be built using the specified container.
 
+**--shell** *shell*
+
+Specify the *shell* to run inside of the container image
+The shell instruction allows the default shell used for the shell form of commands to be overridden. The default shell on Linux is ["/bin/sh", "-c"]
+
```
nalind: Missing `.` at the ends of sentences.

imagebuildah/build.go
```diff
@@ -721,32 +698,25 @@ func BuildDockerfiles(store storage.Store, options BuildOptions, dockerfile ...s
...
-			}
-			if dinfo.Size() == 0 {
-				contents.Close()
-				return errors.Wrapf(err, "no contents in %q", dfile)
-			}
-			rc = contents
+			parsed, err = imagebuilder.ParseFile(dfile)
+		}
+		if err != nil {
+			return errors.Wrapf(err, "uname to parse %s", dfile)
```
nalind: That error message looks odd.  Should it be `"unable to parse %s"`?

run.go
```diff
@@ -58,6 +58,8 @@ type RunOptions struct {
 	User string
 	// WorkingDir is an override for the working directory.
 	WorkingDir string
+	// Shell is default shell to run in a container
```
nalind: Missing `.` at the end of this sentence.

tests/bud.bats
```diff
@@ -262,3 +262,16 @@ load helpers
...
   [ "$output" = "" ]
 }
+
+@test "bud-shell" {
+  target=alpine-image
+  buildah bud --format docker --signature-policy ${TESTSDIR}/policy.json -t ${target} ${TESTSDIR}/bud/shell
+  run buildah --debug=false inspect --type=image --format '{{.Docker.Config.Shell}}' ${target}
+  echo $output
+  [ "$status" -eq 0 ]
+  [ "$output" = "[/bin/sh -c]" ]
```
nalind: This test should probably also verify that `buildah config --shell` updates the value.

contrib/rpm/buildah.spec
rhatdan: Yes there are two PRs here. I just added this one when I saw it, and rather then opening a separate PR for it.

docs/buildah-config.md
rhatdan: Fixed

imagebuildah/build.go
rhatdan: fixed

run.go
rhatdan: Fixed

tests/bud.bats
```diff
@@ -262,3 +262,23 @@ load helpers
...
+  run buildah --debug=false config --shell "/bin/bash" --shell "\-c" ${target}
```
nalind: This doesn't seem to jive with the man page, and it looks kind of weird.  Would it be simpler to parse it as a shell command, like we do for the `--cmd` and `--entrypoint` options?

tests/bud.bats
```diff
@@ -262,3 +262,25 @@ load helpers
...
+
+@test "bud-shell" {
+  target=alpine-image
+  buildah bud --format docker --signature-policy ${TESTSDIR}/policy.json -t ${target} ${TESTSDIR}/bud/shell
+  run buildah --debug=false inspect --type=image --format '{{.Docker.Config.Shell}}' ${target}
+  echo $output
+  [ "$status" -eq 0 ]
+  [ "$output" = "[/bin/sh -c]" ]
+  ctr=$(buildah from ${target})
+  run buildah --debug=false config --shell "/bin/bash \-c" ${ctr}
```
nalind: Is the `\` necessary here?

tests/bud.bats
```diff
@@ -262,3 +262,25 @@ load helpers
...
+  echo $output
+  [ "$status" -eq 0 ]
+  run buildah --debug=false inspect --type=container --format '{{.Docker.Config.Shell}}' ${ctr}
```
nalind: Please switch this from `'{{.Docker.Config.Shell}}'` to `'{{printf "%q" .Docker.Config.Cmd }}'`, so that we can be sure that strings that should be split on whitespace are getting split correctly.

tests/bud.bats
rhatdan: Sadly yes. The urvave CLI tool will not work without escaping it.

tests/bud.bats
rhatdan: Fixed

config.go
rhatdan: Fixed, also added to man page.

cmd/buildah/config.go
```diff
@@ -100,6 +104,10 @@ func updateConfig(builder *buildah.Builder, c *cli.Context) {
 	if c.IsSet("user") {
 		builder.SetUser(c.String("user"))
 	}
+	if c.IsSet("shell") {
+		shellstr := strings.Replace(c.String("shell"), "\\-", "-", -1)
+		builder.SetShell(strings.Fields(shellstr))
+	}
```
nalind: Try using `shellwords.Parse()` instead, like we do for `--entrypoint` and `--cmd`.  I think the `strings.Replace()` call won't be necessary then.

contrib/rpm/buildah.spec
rhatdan: Fixed

tests/bud.bats
```diff
@@ -262,3 +262,25 @@ load helpers
...
+
+@test "bud-shell" {
+  target=alpine-image
+  buildah bud --format docker --signature-policy ${TESTSDIR}/policy.json -t ${target} ${TESTSDIR}/bud/shell
+  run buildah --debug=false inspect --type=image --format '{{printf "%q" .Docker.Config.Shell}}' ${target}
+  echo $output
+  [ "$status" -eq 0 ]
+  [ "$output" = '["/bin/sh" "-c"]' ]
+  ctr=$(buildah from ${target})
+  run buildah --debug=false config --shell "/bin/bash \-c" ${ctr}
```
nalind: Does this test succeed without the `\` here?

docs/buildah-config.md
```diff
@@ -65,6 +65,13 @@ its OS is kept, otherwise the host's OS's name is recorded.
 Specifies a *port* to expose when running containers based on any images which
 will be built using the specified container.
 
+**--shell** *shell*
+
+Specify the *shell* to run inside of the container image
+The shell instruction allows the default shell used for the shell form of commands to be overridden. The default shell command to run Linux is "/bin/sh -c".
```
nalind: That second sentence would probably read better as 'The default shell for Linux containers is "/bin/sh -c"'.

rhatdan: Yes it does.  The urvave only gets confused when it is "-c", it does not complain with "bash -c"

tests/bud.bats
rhatdan: Yes it does.  The urvave only gets confused when it is "-c", it does not complain with "bash -c"

rhatdan: Fixed

nalind:
LGTM, pending CI.

nalind:
bot, retest this please

rhatdan:
@rh-atomic-bot r+

PR #524, opened 2018-03-19
Add various transport support to buildah from

umohnani8:
buildah from now supports pulling images using the following transports:
docker-archive, oci-archive, and dir.

Fixes issue https://github.com/projectatomic/buildah/issues/511

### DIFF
--- a/cmd/buildah/from.go
+++ b/cmd/buildah/from.go
@@ -3,9 +3,11 @@ package main
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/projectatomic/buildah"
+	util "github.com/projectatomic/buildah/util"
 	"github.com/urfave/cli"
 )
 
@@ -100,8 +102,17 @@ func fromCmd(c *cli.Context) error {
 		return err
 	}
 
+	transport := buildah.DefaultTransport
+	arr := strings.SplitN(args[0], ":", 2)
+	if len(arr) == 2 {
+		if _, ok := util.Transports[arr[0]]; ok {
+			transport = arr[0]
+		}
+	}
+
 	options := buildah.BuilderOptions{
 		FromImage:             args[0],
+		Transport:             transport,
 		Container:             c.String("name"),
 		PullPolicy:            pullPolicy,
 		SignaturePolicyPath:   signaturePolicy,
--- a/docs/buildah-from.md
+++ b/docs/buildah-from.md
@@ -25,7 +25,7 @@ Multiple transports are supported:
   **docker-daemon:**_docker-reference_
   An image _docker-reference_ stored in the docker daemon internal storage.  _docker-reference_ must contain either a tag or a digest.  Alternatively, when reading images, the format can also be docker-daemon:algo:digest (an image ID).
 
-  **oci:**_path_**:**_tag_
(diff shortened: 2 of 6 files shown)


new.go
```diff
@@ -177,7 +177,11 @@ func newBuilder(store storage.Store, options BuilderOptions) (*Builder, error) {
 				err = err2
 				continue
 			}
-			srcRef2, err3 := alltransports.ParseImageName(options.Transport + image)
+			transport := options.Transport
+			if transport != DefaultTransport {
```
rhatdan: Does DefaultTransport have a `:` in it?

cmd/buildah/from.go
```diff
@@ -100,8 +102,17 @@ func fromCmd(c *cli.Context) error {
 		return err
 	}
 
+	transport := ""
+	arr := strings.SplitN(args[0], ":", 2)
+	if len(arr) == 2 {
+		if _, ok := util.Transports[arr[0]]; ok {
+			transport = arr[0]
+		}
```
rhatdan: Shold we return an error if this is not ok?

docs/buildah-from.md
```diff
@@ -272,6 +272,12 @@ buildah from imagename --pull
 
 buildah from docker://myregistry.example.com/imagename --pull
 
+buildah from docker-archive:filename --name mycontainer
+
+buildah from oci-archive:filename
+
+buildah from dir:directoryname --name mycontainer
```
rhatdan: What about docker-daemon?

rhatdan:
BTW Can podman do this now, for podman pull?

new.go
umohnani8: DefaultTransport is `docker://`.

cmd/buildah/from.go
umohnani8: No, that means the user omitted a transport and is using the Default Transport.

docs/buildah-from.md
umohnani8: It is supported, I will add an example.

umohnani8:
@rhatdan yeah, podman pull has this support already.

umohnani8:

cmd/buildah/from.go
umohnani8: No leaving as an empty string just means it will be the Default Transport, which the code handles later on.

TomSweeneyRedHat: Right, but if you set it to DefaultTransport here, the code won't have to handle it later and I think it would make this code clearer.  Your call, it's a nit level.

rhatdan:
@umohnani8 Please answer @TomSweeneyRedHat then merge.

cmd/buildah/from.go
umohnani8: setting transport to DefaultTransport on line 105.

TomSweeneyRedHat:
LGTM, thx for the nit fix @umohnani8

TomSweeneyRedHat:
@rh-atomic-bot r+

rhatdan:
@ipbabble I think calls for a new blog, about using buildah to build containers based on content in docker storage?  Perhaps on oci images on your local disk.
@TomSweeneyRedHat I think it is time to release a new buildah with this new support, but lets merge Multi From support.

PR #528, opened 2018-03-20
If the user overrides the storage driver, the options should be dropped

rhatdan:
Since the driver options are configured in storage.conf based on the selected
driver then the user overrideing them should cause the options to drop.

### DIFF
--- a/docs/buildah.md
+++ b/docs/buildah.md
@@ -51,18 +51,22 @@ be used, as the default behavior of using the system-wide configuration
 **--root** **value**
 
 Storage root dir (default: "/var/lib/containers/storage")
+Default root dir is configured in /etc/containers/storage.conf
 
 **--runroot** **value**
 
 Storage state dir (default: "/var/run/containers/storage")
+Default state dir is configured in /etc/containers/storage.conf
 
 **--storage-driver** **value**
 
-Storage driver
+Storage driver.  Default Storage driver is configured in /etc/containers/storage.conf.  Overriding
+this option, will drop the storage-opt definitions was well from the storage.conf file.  User must
+specify additional options via the `--storage-opt` flag.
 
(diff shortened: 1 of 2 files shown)


rhatdan:
This should fix https://github.com/projectatomic/buildah/issues/526

docs/buildah.md
```diff
@@ -89,5 +92,7 @@ Print the version
 | buildah-run(1)        | Run a command inside of the container.                                                               |
 | buildah-tag(1)        | Add an additional name to a local image.                                                             |
 | buildah-umount(1)     | Unmount a working container's root file system.                                                      |
-| buildah-version(1)    | Display the Buildah Version Information
-                                               |
+| buildah-version(1)    | Display the Buildah Version Information                                    |
+| storage.conf(5) | Syntax of Container Storage configuration file
+ |
+| |                                                  |
```
TomSweeneyRedHat: Something funky going on here , there's a bunch of random lines at the bottom of the page if you hit the "View" button.  I think line 98 needs to go away.

rhatdan: @TomSweeneyRedHat I am trying to figure out how we cat get this to look good in both markdown and man page.  Adding these lines caused the bottom of the man page to look better.

Try compiling the docs as a man page and then run man on it. WIth this patch we get an empty thable line, without this patch we get a table line contianer   .TH

rhatdan: I figured this is the least ugly way to clean that up.  If you want to play with the markdown to see if you can get something cleaner, I am all for it.

TomSweeneyRedHat:
If it's just the md that's a bit ugly, then we can play later.  I say ship it as is and we'll adjust.
LGTM/.

rhatdan:
@rh-atomic-bot r+

PR #529, opened 2018-03-20
Add unit tests to run.go

ripcurld:
Trying to cover `run.go` as much as possible.

* Changed `addHostsToFile` to make it easy to test
* Added unit-tests to `addHosts`, `addHostsToFile` and `addRlimits`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,6 +90,7 @@ test-integration:
 
 .PHONY: test-unit
 test-unit:
+	$(GO) test -v -race $(shell go list ./... | grep -v vendor | grep -v tests | grep -v cmd)
 	tmp=$(shell mktemp -d) ; \
 	mkdir -p $$tmp/root $$tmp/runroot; \
 	$(GO) test -v -tags "$(AUTOTAGS) $(TAGS)" ./cmd/buildah -args -root $$tmp/root -runroot $$tmp/runroot -storage-driver vfs -signature-policy $(shell pwd)/tests/policy.json
--- a/run.go
+++ b/run.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -61,5 +61,6 @@ script:
     # Setting up  Docker Registry is complete, let's do Buildah testing! 
     - make install.tools install.libseccomp.sudo all runc validate TAGS="apparmor seccomp containers_image_ostree_stub"
     - go test -c -tags "apparmor seccomp `./btrfs_tag.sh` `./libdm_tag.sh` `./ostree_tag.sh` `./selinux_tag.sh`" ./cmd/buildah
+    - go test -v -race $(go list ./... | grep -v vendor | grep -v tests | grep -v cmd)
```
TomSweeneyRedHat: Can this be combined with line 63?

ripcurld: I tried to merge them, but that caused the "./*.sh" files to fail before running any test.
I need to look more into it, but I guess this will be in another PR.

TomSweeneyRedHat: Ah, I think I got it, this is going to run run_test.go below right?  If so, then it can't be combined with line 63.  @baude is setting up ginkgo gets for use under tests/e2e, we'll probably want to move run_test.go there eventually and convert it if necessary.

TomSweeneyRedHat:
LGTM assuming happy tests.

.travis.yml
ripcurld: On one hand  I can't move this to line 64 because multiple packages are not allowed when `-c` is set and on the other hand I need to set it with tags which leaves me confused :confused:

ripcurld: I decided not to add anything to `.travis.yml` as I put it under a lot of load.
I wish there was a way to reproduce `Travis-CI` on my local machine. :thinking:

TomSweeneyRedHat:
LGTM and happy green test buttons.  I feel your .travis.yml pain @ripcurld, it's been biting me lately too.

rhatdan:
LGTM
@rh-atomic-bot r+

ripcurld:
Thank you! :balloon: :cloud:

PR #541, opened 2018-03-27
buildah bud walks symlinks

umohnani8:
After a symlink is created, buildah bud would fail if the symlink was
mounted as a volume.

Fixes https://github.com/projectatomic/buildah/issues/439

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -178,6 +178,16 @@ func (b *Executor) Preserve(path string) error {
 	cacheFile := filepath.Join(cacheDir, fmt.Sprintf("volume%d.tar", b.preserved))
 	// Save info about the top level of the location that we'll be archiving.
 	archivedPath := filepath.Join(b.mountPoint, path)
+
+	// Try and resolve the symlink (if one exists)
+	// Set archivedPath and path based on whether a symlink is found or not
+	if symLink, err := resolveSymLink(b.mountPoint, path); err == nil {
+		archivedPath = filepath.Join(b.mountPoint, symLink)
+		path = symLink
+	} else {
+		return errors.Wrapf(err, "error reading symbolic link to %q", path)
+	}
+
 	st, err := os.Stat(archivedPath)
 	if os.IsNotExist(err) {
 		if err = os.MkdirAll(archivedPath, 0755); err != nil {
--- a/imagebuildah/chroot_symlink.go
+++ b/imagebuildah/chroot_symlink.go
@@ -1,0 +1,145 @@
+package imagebuildah
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/containers/storage/pkg/reexec"
+	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	symlinkChrootedCommand = "chrootsymlinks-resolve"
+	maxSymlinksResolved    = 40
+)
+
+func init() {
+	reexec.Register(symlinkChrootedCommand, resolveChrootedSymlinks)
+}
+
+func resolveChrootedSymlinks() {
+	status := 0
+	flag.Parse()
+	if len(flag.Args()) < 1 {
+		os.Exit(1)
+	}
+	// Our first parameter is the directory to chroot into.
+	if err := unix.Chdir(flag.Arg(0)); err != nil {
+		fmt.Fprintf(os.Stderr, "chdir(): %v\n", err)
+		os.Exit(1)
+	}
+	if err := unix.Chroot(flag.Arg(0)); err != nil {
+		fmt.Fprintf(os.Stderr, "chroot(): %v\n", err)
+		os.Exit(1)
+	}
+
+	// Our second paramter is the path name to evaluate for symbolic links
+	symLink, err := getSymbolicLink(flag.Arg(0), flag.Arg(1))
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "error getting symbolic links: %v\n", err)
+		os.Exit(1)
+	}
+	if _, err := os.Stdout.WriteString(symLink); err != nil {
+		fmt.Fprintf(os.Stderr, "error writing string to stdout: %v\n", err)
+		os.Exit(1)
+	}
+	os.Exit(status)
+}
+
+func resolveSymLink(rootdir, filename string) (string, error) {
+	// The child process expects a chroot and one path that
+	// will be consulted relative to the chroot directory and evaluated
+	// for any symbolic links present.
+	cmd := reexec.Command(symlinkChrootedCommand, rootdir, filename)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return "", errors.Wrapf(err, string(output))
+	}
+
+	// Hand back the resolved symlink, will be "" if a symlink is not found
+	return string(output), nil
+}
+
+// getSymbolic link goes through each part of the path and continues resolving symlinks as they appear.
+// Returns what the whole target path for what "path" resolves to.
+func getSymbolicLink(rootdir, path string) (string, error) {
+	var (
+		symPath          string
+		symLinksResolved int
+	)
+
+	// Splitting path as we need to resolve each parth of the path at a time
+	splitPath := strings.Split(path, "/")
+	if splitPath[0] == "" {
+		splitPath = splitPath[1:]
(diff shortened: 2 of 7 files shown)


umohnani8:

imagebuildah/build.go
```diff
@@ -227,6 +244,46 @@ func (b *Executor) Preserve(path string) error {
 	return nil
 }
 
+// getSymbolicLink finds any symbolic links in the path
+// If the path is "/my/path/with/symlinks" and "/my/path" is the symlink
+// getSymbolicLink starts with "/my" then "/my/path" and so forth to find a symlink
+// if one exists. Returns an empty string if a symlink doesn't exist
+func getSymbolicLink(mountPoint, path string) (string, string, error) {
```
rhatdan: Did you look at https://golang.org/pkg/path/filepath/#EvalSymlinks, does everything you need except you are still dangerous.  
We have to do these calls inside of a chroot.

Look at libpod/pkg/chrootuser for examples of how to do this.

umohnani8:
Doing the symlink calls inside of a chroot.
Cannot use `filepath.EvalSymlinks()` as there may be paths with non-existent directories/files and `filepath.EvalSymlinks()` calls `lstat` internally, which throws a directory does not exist error. So doing it with a loop and `Lstat` and `ReadLink`.
I put the chroot code under `imagebuildah` in the file `chroot_symlink.go`, not sure where exactly to put. Can move it around based on where it fits better.
Added tests for the various cases I could think of.
@rhatdan @nalind PTAL

imagebuildah/build.go
```diff
@@ -227,6 +244,56 @@ func (b *Executor) Preserve(path string) error {
 	return nil
 }
 
+// getSymbolicLink finds any symbolic links in the path
```
rhatdan: Remove this section

imagebuildah/chroot_symlink.go
```diff
@@ -0,0 +1,115 @@
...
+
+func resolveChrootedSymlinks() {
+	status := 0
+	flag.Parse()
+	if len(flag.Args()) < 1 {
+		os.Exit(1)
+	}
+	// Our first parameter is the directory to chroot into.
+	if err := unix.Chdir(flag.Arg(0)); err != nil {
+		fmt.Fprintf(os.Stderr, "chdir(): %v", err)
```
rhatdan: Don't these messages need "\n"

rhatdan:
@nalind PTAL

rhatdan:
@nalind PTAL

imagebuildah/build.go
```diff
@@ -38,6 +38,9 @@ const (
 	Uncompressed = archive.Uncompressed
 )
 
+// ErrNoSymlink is returned when the path porvided is not a symbolic link
+var ErrNoSymlink = errors.New("path is not a symbolic link")
```
nalind: This error value doesn't appear to be returned by anything.

imagebuildah/chroot_symlink.go
nalind: I think so.

imagebuildah/chroot_symlink.go
```diff
@@ -0,0 +1,115 @@
...
+		os.Exit(1)
+	}
+	if err := unix.Chroot(flag.Arg(0)); err != nil {
+		fmt.Fprintf(os.Stderr, "chroot(): %v", err)
+		os.Exit(1)
+	}
+
+	// Our second paramter is the mountpoint/rootdir of the container
```
nalind: The second parameter appears to be a pathname under the root directory of the container, rather than the root directory itself, which is the first parameter.

imagebuildah/chroot_symlink.go
```diff
@@ -0,0 +1,115 @@
...
+		if symLinksResolved > 39 {
+			return "", errors.Errorf("have resolved 40 symlinks, something is terribly wrong!")
+		}
+		symPath = filepath.Join(symPath, p)
+		info, err := os.Lstat(symPath)
+		if os.IsNotExist(err) {
+			if err = os.MkdirAll(symPath, 0755); err != nil {
+				return "", errors.Wrapf(err, "error ensuring volume path %q exists", symPath)
+			}
+			info, err = os.Lstat(symPath)
```
nalind: This should check `err`, just in case.  If it's set for whatever reason, and `info` is still `nil`, we'll crash right after.

imagebuildah/chroot_symlink.go
```diff
@@ -0,0 +1,115 @@
...
+
+		// Read the symlink to get what it points to
+		targetDir, err := os.Readlink(symPath)
+		if err != nil {
+			return "", errors.Wrapf(err, "error reading link %q", symPath)
+		}
+		// if symlink points to itself, return an error
+		if targetDir == symPath {
+			return "", errors.Errorf("symlink %q points to itself: %q", symPath, targetDir)
+		}
```
nalind: This particular check shouldn't be necessary, if the check on the maximum number of links we resolve will catch it.

imagebuildah/chroot_symlink.go
```diff
@@ -0,0 +1,115 @@
...
+		// if the symlink points to a relative path, prepend the path till now to the resolved path
+		if !filepath.IsAbs(targetDir) {
+			targetDir = filepath.Join(symPath, targetDir)
+		}
+		// run filepath.Clean to remove the ".." from relative paths
+		symPath = filepath.Clean(targetDir)
```
nalind: At this point, `symPath` could now be resolved to a new pathname which is also, itself, a symbolic link (and on, and on).  Do we handle this?

imagebuildah/build.go
```diff
@@ -177,7 +180,21 @@ func (b *Executor) Preserve(path string) error {
...
-	archivedPath := filepath.Join(b.mountPoint, path)
+	var archivedPath string
+
+	// Try and resolve the symlink (if one exists)
+	// Set archivedPath and path based on whether a symlink is found or not
+	symLink, err := resolveSymLink(b.mountPoint, path)
+	switch err {
+	case nil:
+		archivedPath = filepath.Join(b.mountPoint, symLink)
+		path = symLink
```
nalind: What are the benefits of changing the key that we're using for the `volumeCache` and `volumeCacheInfo` maps?

imagebuildah/chroot_symlink.go
umohnani8: fixed

imagebuildah/build.go
umohnani8: right, removed.

imagebuildah/chroot_symlink.go
umohnani8: Whoops, my bad. Fixed

imagebuildah/chroot_symlink.go
umohnani8: Fixed.

imagebuildah/chroot_symlink.go
umohnani8: Removed check.

imagebuildah/chroot_symlink.go
umohnani8: Just added functionality to handle this. PTAL

imagebuildah/build.go
umohnani8: So if after volume mounting, we decide to add a file to the path where the symlink points to fo example, the build code will fail as it won't be able to find that path in the cache maps. Changing this to be the resolved path of the symlink helps get rid of that error.

imagebuildah/build.go
```diff
@@ -178,6 +178,17 @@ func (b *Executor) Preserve(path string) error {
 	cacheFile := filepath.Join(cacheDir, fmt.Sprintf("volume%d.tar", b.preserved))
 	// Save info about the top level of the location that we'll be archiving.
 	archivedPath := filepath.Join(b.mountPoint, path)
+
+	// Try and resolve the symlink (if one exists)
+	// Set archivedPath and path based on whether a symlink is found or not
+	symLink, err := resolveSymLink(b.mountPoint, path)
```
ripcurld: A small nitpick -
```
if symLink, err := resolveSymLink(b.mountPoint, path); err == nil {
  ...
} else {
 ...
}
```

umohnani8: fixed

ripcurld:
LGTM 2 :+1:

rhatdan:
@rh-atomic-bot r=ripcurld

PR #543, opened 2018-03-29
bash: Update some options

q384566678:

### DIFF
--- a/contrib/completions/bash/buildah
+++ b/contrib/completions/bash/buildah
@@ -164,6 +164,8 @@ return 1
          --version -v
          "
      local options_with_args="
+         --registries-conf
+         --registries-conf-dir
          --root
          --runroot
          --storage-driver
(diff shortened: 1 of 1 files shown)


rhatdan:
Thanks,
LGTM

contrib/completions/bash/buildah
```diff
@@ -164,6 +164,8 @@ return 1
          --version -v
          "
      local options_with_args="
+         --registries-conf
```
TomSweeneyRedHat: Do we need these here since they're in the global options below?

TomSweeneyRedHat:
@q384566678 thanks for the updates.  I think you may have a couple of duplicate entries, otherwise LGTM.

contrib/completions/bash/buildah
rhatdan: I believe we only need the buildah_buildah() ones, the other global options in buildah() should be removed.

q384566678: updated. It's strange why tests fail.

rhatdan:
@rh-atomic-bot r+

PR #540, opened 2018-03-27
docs: fix error example 

q384566678:
Fix the following error example:
```
➜  /data sudo buildah from imagename --signature-policy /etc/containers/policy.json
too many arguments specified
```

Signed-off-by: Zhou Hao <zhouhao@cn.fujitsu.com>

PR #545, opened 2018-03-30
buildah-run.md: fix error SYNOPSIS

q384566678:
I fixed the example of `build-run.md` in #540. When I changed it, I only followed SYNOPSIS, but when I used it today, I found the following error:

```
➜  buildah git:(fix-run) sudo buildah run 819 --hostname zz hostname
container_linux.go:348: starting container process caused "exec: \"--hostname\": executable file not found in $PATH"
```
The correct form is as follows:
```
➜  buildah git:(fix-run) sudo buildah run --hostname zz 819 hostname
zz
```

So I modified SYNOPSIS and related examples.

### DIFF
--- a/docs/buildah-run.md
+++ b/docs/buildah-run.md
@@ -4,7 +4,7 @@
 buildah run - Run a command inside of the container.
 
 ## SYNOPSIS
-**buildah** **run** **containerID** [*options* [...] --] **command**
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@rh-atomic-bot r+

PR #546, opened 2018-03-30

rhatdan:
Some more features.
docker-archive generates docker legacy compatible images
Do not create $DiffID subdirectories for layers with no configs
Ensure the layer IDs in legacy docker/tarfile metadata are unique
docker-archive: repeated layers are symlinked in the tar file
sysregistries: remove all trailing slashes
Improve docker/* error messages
Fix failure to make auth directory
Create a new slice in Schema1.UpdateLayerInfos
Drop unused storageImageDestination.{image,systemContext}
Load a *storage.Image only once in storageImageSource
Support gzip for docker-archive files
Remove .tar extension from blob and config file names
ostree, src: support copy of compressed layers
ostree: re-pull layer if it misses uncompressed_digest|uncompressed_size
image: fix docker schema v1 -> OCI conversion
Add /etc/containers/certs.d as default certs directory

Signed-off-by: Daniel J Walsh dwalsh@redhat.com

### DIFF
--- a/vendor/github.com/containers/image/types/types.go
+++ b/vendor/github.com/containers/image/types/types.go
@@ -132,6 +132,19 @@ type ImageSource interface {
 	LayerInfosForCopy() ([]BlobInfo, error)
 }
 
+// LayerCompression indicates if layers must be compressed, decompressed or preserved
+type LayerCompression int
+
+const (
+	// PreserveOriginal indicates the layer must be preserved, ie
+	// no compression or decompression.
+	PreserveOriginal LayerCompression = iota
+	// Decompress indicates the layer must be decompressed
+	Decompress
+	// Compress indicates the layer must be compressed
+	Compress
+)
+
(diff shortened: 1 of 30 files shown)


vendor/github.com/containers/image/types/types.go
```diff
@@ -292,7 +305,7 @@ type ManifestUpdateInformation struct {
 // for other manifest types.
 type ImageInspectInfo struct {
 	Tag           string
-	Created       time.Time
+	Created       *time.Time
```
ripcurld: it seems like this change breaks the tests. 
I think we need to update `cmd/buildah/common.go` line 116 to
```
created = *inspectInfo.Created
```
and `cmd/buildah/images.go` line 198 to:
```
date := *inspect.Created
```

rhatdan:

TomSweeneyRedHat:

rhatdan:
@rh-atomic-bot r+

PR #551, opened 2018-04-04
Remove requires for ostree-lib in rpm spec file

rhatdan:
I stupidly added this in a previous patch. rpm figures this
out for you so no need to specify.  On RHEL systems ostree-libs
does not exist.


rhatdan:
@rh-atomic-bot r+

PR #554, opened 2018-04-04
Imagename is required for commit atm, update manpage

TomSweeneyRedHat:

This will probably be a short term fix as commit should allow the imageName to not be specified and then set it to <none>.


rhatdan:
@rh-atomic-bot r+

PR #556, opened 2018-04-06
addcopy: add src validation

q384566678:
If the src is not specified, an error should be returned.

### DIFF
--- a/cmd/buildah/addcopy.go
+++ b/cmd/buildah/addcopy.go
@@ -22,7 +22,7 @@ var (
 		Description:    addDescription,
 		Flags:          addAndCopyFlags,
 		Action:         addCmd,
-		ArgsUsage:      "CONTAINER-NAME-OR-ID [[FILE | DIRECTORY | URL] ...] [DESTINATION]",
+		ArgsUsage:      "CONTAINER-NAME-OR-ID [FILE | DIRECTORY | URL] [... [DESTINATION]]",
 		SkipArgReorder: true,
 	}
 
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

cmd/buildah/addcopy.go
```diff
@@ -44,6 +44,9 @@ func addAndCopyCmd(c *cli.Context, extractLocalArchives bool) error {
 	}
 	name := args[0]
 	args = args.Tail()
+	if len(args) == 0 {
```
TomSweeneyRedHat: This should be "== 1".  The command should be `buildah add containerID SRC [[...] DEST]`  So both args[0] and args[1] should be provided.  Good catch though, that definitely was a hole that needed filling!

q384566678:
ping @rhatdan

rhatdan:
@q384566678 Can you update to the latest patches so travis will patch.

cmd/buildah/addcopy.go
TomSweeneyRedHat: Ooops!  Sorry about that @q384566678, I somehow missed the Tail() call a line above.  Ouch.  This is fine.

TomSweeneyRedHat:
@q384566678 as @rhatdan you'll need to rebase your patch so you can pull in the the new .travis.yml from #572

q384566678:
@rhatdan @TomSweeneyRedHat  rebased. PTAL.

PR #560, opened 2018-04-06
Add accessor functions for comment and stop signal

baude:
Podman commit needs the ability to set a comment and stop signal
when committing a container to an image.  Adding set/get functions
for each.

### DIFF
--- a/config.go
+++ b/config.go
@@ -579,3 +579,30 @@ func (b *Builder) SetDomainname(name string) {
 func (b *Builder) SetDefaultMountsFilePath(path string) {
 	b.DefaultMountsFilePath = path
 }
+
+// Comment returns the comment which will be set in the container and in
+// containers built using images built from the container
+func (b *Builder) Comment() string {
+	return b.Docker.Comment
+}
+
+// SetComment sets the Comment which will be set in the container and in
+// containers built using images built from the container.
+// Note: this setting is not present in the OCIv1 image format, so it is
+// discarded when writing images using OCIv1 formats.
+func (b *Builder) SetComment(comment string) {
+	b.Docker.Comment = comment
+}
+
+// StopSignal returns the signal which will be set in the container and in
+// containers built using images buiilt from the container
+func (b *Builder) StopSignal() string {
+	return b.Docker.Config.StopSignal
+}
+
+// SetStopSignal sets the signal which will be set in the container and in
+// containers built using images built from the container.
+func (b *Builder) SetStopSignal(stopSignal string) {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

baude:
sit on this one while i work some edge case issues

baude:
Fixed up, should be good now.

config.go
```diff
@@ -579,3 +579,29 @@ func (b *Builder) SetDomainname(name string) {
 func (b *Builder) SetDefaultMountsFilePath(path string) {
 	b.DefaultMountsFilePath = path
 }
+
+// Comment returns the comment which will be set in the container and in
+//containers built using images buiilt from the container
```
TomSweeneyRedHat: nit: buiilt to built ditto line 597

baude: it is a copy of existing text but can do

rhatdan:
@baude Waiting for the spelling error fix and then we can merge.

config.go
baude: fixed

baude:
fixed spelling error, please merge

config.go
```diff
@@ -579,3 +579,29 @@ func (b *Builder) SetDomainname(name string) {
...
+//containers built using images built from the container
+func (b *Builder) Comment() string {
+	return b.Docker.Comment
+}
+
+// SetComment sets the Comment which will be set in the container and in
+// containers built using images built from the container.
+func (b *Builder) SetComment(comment string) {
+	b.Docker.Comment = comment
+	b.OCIv1.History[0].Comment = comment
```
nalind: The `History` list can be empty if the image is being built from scratch, and when it isn't empty, this is the entry for the base layer.

baude: Adding check for len ... I think adding the comment to the base layer is OK; at least that is what I have seen being done before.  I'll suggest a work-around in my next push.  or is there an alternate way you would want the comment set for v1?

config.go
```diff
@@ -579,3 +579,30 @@ func (b *Builder) SetDomainname(name string) {
 func (b *Builder) SetDefaultMountsFilePath(path string) {
 	b.DefaultMountsFilePath = path
 }
+
+// Comment returns the comment which will be set in the container and in
+//containers built using images built from the container
```
nalind: Add a whitespace after `//` here, to match the rest.

nalind:
Whitespace nit, otherwise LGTM.

config.go
baude: fixed

nalind:
@rh-atomic-bot r+

PR #567, opened 2018-04-11
config: add support for StopSignal

q384566678:
Add a stop-singnal option for buildah-config.

### DIFF
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -64,6 +64,10 @@ var (
 			Usage: "add `shell` to run in containers",
 		},
 		cli.StringFlag{
+			Name:  "stop-signal",
+			Usage: "set `stop signal` for containers based on image",
+		},
+		cli.StringFlag{
 			Name:  "user, u",
 			Usage: "set default `user` to run inside containers based on image",
 		},
@@ -111,6 +115,9 @@ func updateConfig(builder *buildah.Builder, c *cli.Context) {
 		} else {
 			builder.SetShell(shellSpec)
 		}
+	}
+	if c.IsSet("stop-signal") {
+		builder.SetStopSignal(c.String("stop-signal"))
 	}
 	if c.IsSet("port") || c.IsSet("p") {
 		for _, portSpec := range c.StringSlice("port") {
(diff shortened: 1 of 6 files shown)


docs/buildah-config.md
```diff
@@ -72,6 +72,10 @@ The shell instruction allows the default shell used for the shell form of comman
 
 Note: this setting is not present in the OCIv1 image format, so it is discarded when writing images using OCIv1 formats.
 
+**--stop-signal** *signal*
+
+Specifies a *signal* and the system call this *signal* that will be sent to the container to exit.
```
TomSweeneyRedHat: Suggest: "Sets the system call *signal* that will be sent to the container to exit."

TomSweeneyRedHat:
Ditto @ripcurld's thought on test.  The test failures don't seem to be related to this PR, looks like some kind of unrelated issue on the test systems.  @q384566678 thanks for the PR!

rhatdan:
Lets get this in ASAP,  Now that we have Multi-stage builds we want to cut a new release and this is something we should support.

cmd/buildah/config.go
```diff
@@ -63,6 +63,10 @@ var (
 			Name:  "shell",
 			Usage: "add `shell` to run in containers",
 		},
+		cli.StringFlag{
+			Name:  "stop-signal",
+			Usage: "add `signal` to stop containers",
```
rhatdan: set default stop signal for container.  This signal will be sent when container is stopped, default is SIGINT.

docs/buildah-config.md
rhatdan: I would explain it further as I did above.



set default stop signal for container. This signal will be sent when container is stopped, default is SIGINT.

rhatdan:
A good test would be to just check buildah inspect to make sure the stop signal is set.

TomSweeneyRedHat:
Test issues should be fixed with #572

rhatdan:
@q384566678 Could you update your PR to pull in the fix for the docker change.

q384566678:
@rhatdan What's the change for docker?

q384566678:
I updated the explanation and added related tests.

rhatdan:
@q384566678 I think you figured it out, but docker no longer wants you to do `--password foobar` on the command line.

LGTM

rhatdan:
@rh-atomic-bot r+

PR #582, opened 2018-04-12
Fix secrets patch

umohnani8:
The secrets code was just tarring and copying the contents of the secrets directory on host as is.
This meant it was not accounting for any symlinks inside the directory, leading up to the contents
not being copied over.

### DIFF
--- a/secrets.go
+++ b/secrets.go
@@ -3,6 +3,7 @@ package buildah
 import (
 	"bufio"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


umohnani8:
I had modified the secrets code from cri-o and had not thought about the fact that the contents of the secrets directory could be symlinks pointing to other directories. Huge bug on my part, sorry!
Working up the same fix for podman.

umohnani8:
@tchughesiv buildah secrets issue is being fixed here.

TomSweeneyRedHat:
Changes LGTM, any tests that we can add/modify to validate the symlink traversal?

umohnani8:
@TomSweeneyRedHat yup just modified the test to account for this situation.

rhatdan:
LGTM
@rh-atomic-bot r+

PR #583, opened 2018-04-12
Remove tarball as an option from buildah push --help

umohnani8:
pushing to the tarball transport is not supported, so removing
the option from the inline help

### DIFF
--- a/cmd/buildah/push.go
+++ b/cmd/buildah/push.go
@@ -62,7 +62,7 @@ var (
    %s
 
    See buildah-push(1) section "DESTINATION" for the expected format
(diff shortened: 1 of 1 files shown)


umohnani8:
Fixes https://github.com/projectatomic/buildah/issues/557

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM.  The man page for push looks fine to me, but if you could just double check @umohnani8 if you haven't already, I'd appreciate it.

umohnani8:
@TomSweeneyRedHat double checked the man page and it looks good :)

PR #589, opened 2018-04-16
buildah-containers: add validation to 'format'

q384566678:
Avoid the following situations：

```
➜  buildah git:(containers-format) sudo buildah containers --format zz
zz
zz
zz
zz
zz
zz
zz
zz
zz
```

After change:
```
➜  buildah git:(containers-format) sudo ./buildah containers --format zz
error format: zz
```

### DIFF
--- a/cmd/buildah/containers.go
+++ b/cmd/buildah/containers.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"regexp"
 	"strings"
 	"text/template"
 
@@ -239,6 +240,13 @@ func outputContainers(store storage.Store, opts containerOptions, params *contai
 }
 
 func containerOutputUsingTemplate(format string, params containerOutputParams) error {
+	matched, err := regexp.MatchString("{{.*}}", format)
+	if err != nil {
+		return errors.Wrapf(err, "error validating format provided %s", format)
+	} else if !matched {
+		return errors.Errorf("error invalid format provided: %s", format)
+	}
+
 	tmpl, err := template.New("container").Parse(format)
 	if err != nil {
 		return errors.Wrapf(err, "Template parsing error")
--- a/cmd/buildah/containers.go
+++ b/cmd/buildah/containers.go
@@ -242,7 +242,7 @@ func outputContainers(store storage.Store, opts containerOptions, params *contai
 func containerOutputUsingTemplate(format string, params containerOutputParams) error {
 	matched, err := regexp.MatchString("{{.*}}", format)
 	if err != nil {
-		return errors.Wrapf(err, "error validating format provided %s", format)
+		return errors.Wrapf(err, "error validating format provided: %s", format)
 	} else if !matched {
 		return errors.Errorf("error invalid format provided: %s", format)
 	}
--- a/cmd/buildah/containers_test.go
+++ b/cmd/buildah/containers_test.go
@@ -1,0 +1,45 @@
+package main
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestTemplateOutputValidFormat(t *testing.T) {
+	params := containerOutputParams{
+		ContainerID:   "e477836657bb",
+		Builder:       " ",
+		ImageID:       "f975c5035748",
+		ImageName:     "test/image:latest",
+		ContainerName: "test-container",
+	}
+
+	formatString := "Container ID: {{.ContainerID}}"
+	expectedString := "Container ID: " + params.ContainerID
+
+	output, err := captureOutputWithError(func() error {
+		return containerOutputUsingTemplate(formatString, params)
+	})
+	if err != nil {
+		t.Error(err)
+	} else if strings.TrimSpace(output) != expectedString {
+		t.Errorf("Errorf with template output:\nExpected: %s\nReceived: %s\n", expectedString, output)
+	}
+}
+
+func TestTemplateOutputInvalidFormat(t *testing.T) {
+	params := containerOutputParams{
+		ContainerID:   "e477836657bb",
+		Builder:       " ",
+		ImageID:       "f975c5035748",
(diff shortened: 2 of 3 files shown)


cmd/buildah/containers.go
```diff
@@ -239,6 +239,10 @@ func outputContainers(store storage.Store, opts containerOptions, params *contai
 }
 
 func containerOutputUsingTemplate(format string, params containerOutputParams) error {
+	if !strings.HasPrefix(format, "{{.") || !strings.HasSuffix(format, "}}") {
+		return errors.Errorf("error format: %s", format)
```
TomSweeneyRedHat: Can we change the error to ("error invalid format provided: %s", format)

Otherwise LGTM, thanks for the touchup!

ripcurld: I'm afraid if someone will set `--format` to `"container ID: {{.Id}}"` this will return error although it's legit.

rhatdan: I agree. if we want to do the format check to make sure their is json, it should be done with a regex expression.

q384566678: updated, PTAL.

cmd/buildah/containers.go
```diff
@@ -239,6 +240,13 @@ func outputContainers(store storage.Store, opts containerOptions, params *contai
 }
 
 func containerOutputUsingTemplate(format string, params containerOutputParams) error {
+	matched, err := regexp.MatchString("{{.*}}", format)
```
rhatdan: This still does not match the format suggested before.  Perhaps something like
```
matched, err := regexp.MatchString(".*{{.*}}", format)
```
But add some tests of valid formats and invalid formats.

cmd/buildah/containers.go
```diff
@@ -239,6 +240,13 @@ func outputContainers(store storage.Store, opts containerOptions, params *contai
...
+	if err != nil {
+		return err
```
rhatdan: return errors.Wrapf(err, "error validating format provided %s", format)

rhatdan:
Could you add tests including validating that

--format  "container ID: {{.Id}}"

Works, and add this example to the man page.

cmd/buildah/containers.go
q384566678: Don't need change. It works.

```
➜  buildah git:(containers-format) sudo ./buildah containers --format "containerID: {{.ContainerID}}"
containerID: e477836657bba9fd76b80fd71c13318dc18a78711102a32b110e72d37187602e
containerID: 09d1845b52485ce3bc431ff8e57b1e4910723d96af6223bf3a421ed6b48a5d6d
containerID: b025ebe4b1d94887e876f2ebc7e410396f5fe4780ec1f80437d36c39cc3189d8
containerID: 5666da5d29003faecbaf44c951ee465402aa2c500e4659b60767ebb7c94b087e
```

q384566678:
>Could you add tests including validating that
--format "container ID: {{.Id}}"

Do you mean this?

sudo ./buildah containers --format "containerID: {{.ContainerID}}"

cmd/buildah/containers.go
rhatdan: Excellent.

q384566678:
updated, PTAL.

cmd/buildah/containers_test.go
```diff
@@ -0,0 +1,45 @@
...
+		ContainerName: "test-container",
+	}
+
+	formatString := "ContainerID"
+
+	err := containerOutputUsingTemplate(formatString, params)
+	if err == nil || err.Error() != "error invalid format provided: ContainerID" {
+		t.Fatalf("expected error invalid format")
+	}
+}
```
TomSweeneyRedHat: I'm not sure these tests will get run automatically with Travis.  They look great, but is there anyway you could change them into  bats format in our /tests directory?

cmd/buildah/containers.go
```diff
@@ -239,6 +240,13 @@ func outputContainers(store storage.Store, opts containerOptions, params *contai
 }
 
 func containerOutputUsingTemplate(format string, params containerOutputParams) error {
+	matched, err := regexp.MatchString("{{.*}}", format)
+	if err != nil {
+		return errors.Wrapf(err, "error validating format provided %s", format)
```
TomSweeneyRedHat: nit of a nit, can you add a colon ':' after provided please?

cmd/buildah/containers.go
TomSweeneyRedHat: Is this new code OK with this:

sudo ./buildah containers --format "containerID: {{}}"

i.e. does it hit the error that you have at line 247 or does it do something else?

cmd/buildah/containers_test.go
q384566678: > could change them into bats format in our /tests directory?

I also wanted to do it at the beginning, but I was not familiar with bats and did not find a suitable method. So only use this method.
I will learn more about bats and hope to find a suitable solution.

q384566678:
> i.e. does it hit the error that you have at line 247 or does it do something else?

No. It will failed. But it is not because of the change.

```
➜  buildah git:(containers-format) sudo ./buildah containers --format "contaienrID: {{}}"
Template parsing error: template: container:1: missing value for command
```

TomSweeneyRedHat:
Thanks for checking on:

```
 buildah git:(containers-format) sudo ./buildah containers --format "contaienrID: {{}}"
Template parsing error: template: container:1: missing value for command
```
That's a valid error in that situation. I was concerned it might go completely off the rails in that instance and not give enough info to the user to figure out what the issue was.

rhatdan:
bot, retest this please

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

rhatdan:
@rh-atomic-bot r+
@q384566678 Thanks for the PR.

PR #592, opened 2018-04-16
Clarify the use of buildah config options

rhatdan:
Make it clearer to the user which options can be executed
multiple times, versus those that can only used once.

Cleanup some documentation issues

### DIFF
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -20,7 +20,7 @@ var (
 	configFlags = []cli.Flag{
 		cli.StringSliceFlag{
 			Name:  "annotation, a",
-			Usage: "add `annotation` e.g. annotation=value, for the target image",
+			Usage: "add `annotation` e.g. annotation=value, for the target image (default [])",
 		},
 		cli.StringFlag{
 			Name:  "arch",
@@ -32,11 +32,11 @@ var (
 		},
 		cli.StringFlag{
 			Name:  "cmd",
-			Usage: "sets the default `command` to run for containers based on the image",
+			Usage: "set the default `command` to run for containers based on the image",
 		},
 		cli.StringFlag{
 			Name:  "created-by",
-			Usage: "add `description` of how the image was created",
+			Usage: "set `description` of how the image was created",
 			Value: DefaultCreatedBy,
 		},
 		cli.StringFlag{
@@ -45,7 +45,7 @@ var (
 		},
 		cli.StringSliceFlag{
 			Name:  "env, e",
-			Usage: "add `environment variable` to be set when running containers based on image",
+			Usage: "add `environment variable` to be set when running containers based on image (default [])",
 		},
 		cli.StringSliceFlag{
 			Name:  "label, l",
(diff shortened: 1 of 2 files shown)


rhatdan:
Fixed https://github.com/projectatomic/buildah/issues/591

cmd/buildah/config.go
```diff
@@ -73,7 +73,7 @@ var (
 		},
 		cli.StringSliceFlag{
 			Name:  "volume, v",
-			Usage: "add default `volume` path to be created for containers based on image",
+			Usage: "add default `volume` path to be created for containers based on image  (default [])",
```
TomSweeneyRedHat: nit, looks like you've a double space before (default []).

docs/buildah-config.md
```diff
@@ -13,24 +13,24 @@ Updates one or more of the settings kept for a container.
 
 **--annotation** *annotation*
 
-Adds an image *annotation* (e.g. annotation=*annotation*) to the image manifest
-of any images which will be built using the specified container.
+Add an image *annotation* (e.g. annotation=*annotation*) to the image manifest
```
TomSweeneyRedHat: could you whip up an example or three that shows how some of the multiples could be used?  Perhaps an example for annotations, port and volume?

rhatdan:
bot, retest this please

docs/buildah-config.md
```diff
@@ -41,33 +41,33 @@ edits") in any images which will be created using the specified container.
...
+Set the target *operating system* for any images which will be built using
 the specified container.  By default, if the container was based on an image,
 its OS is kept, otherwise the host's OS's name is recorded.
 
 **--port** *port*
 
-Specifies a *port* to expose when running containers based on any images which
-will be built using the specified container.
+Add a *port* to expose when running containers based on any images which
+will be built using the specified container. Can be used multiple times.
```
taqtiqa-mark: Can you clarify the syntax used in this [Redhat documentation](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_atomic_host/7/html/managing_containers/finding_running_and_building_containers_without_docker#creating_images_from_scratch_with_buildah)

````
# buildah config --port 80/tcp working-container
````
Are `/tcp` `/udp` and `/icmp` supported.  If you use this decoration in the config do you have to use that when you launch, for example, a rkt container from the command line?

rhatdan: I would doubt you have to do anything special with rkt on the command line, although I have know clue.

@baude @TomSweeneyRedHat Could you verify the syntax.

taqtiqa-mark: According to the spec the description should be:
> Can be in the format of: `port/tcp`, `port/udp`, `port` with the default protocol being `tcp` if not specified. ICMP is not part of the OCI specification.

Hmm, maybe the documentation burden can be eased by linking to the version of the spec that Buildah is targeting. Example:

> See **OCI spec v1.0.1** [Properties:ExposedPorts](https://github.com/opencontainers/image-spec/blob/v1.0.1/config.md#properties)

If you're happy with that approach I'll add such a reference as I go?

rhatdan: Sure lets merge this and then open a PR with these additional changes to docs.

rhatdan:
@rh-atomic-bot r+

rhatdan:
@taqtiqa-mark Could you add a PR with the changes you want.

PR #594, opened 2018-04-17
buildah images and podman images are listing different sizes

rhatdan:
After reviewing how the sizes are shown.  It looks like podman was
using docker/units to display size.  Its size is based off of multiples
of 1000, which according to Wikipedia is correct for image sizes.

https://en.wikipedia.org/wiki/Binary_prefix

Buildah was using multiples of 1024.  This patch switches buildah to use the same
multiple and to reduce its accuracy to two decimal points so that buildah
podman will return the same sizes.  This also will match the way that docker
reports image sizes.

### DIFF
--- a/cmd/buildah/images_test.go
+++ b/cmd/buildah/images_test.go
@@ -74,14 +74,14 @@ func TestSizeFormatting(t *testing.T) {
 		t.Errorf("Error formatting size: expected '%s' got '%s'", "0 B", size)
 	}
 
-	size = formattedSize(1024)
+	size = formattedSize(1000)
 	if size != "1 KB" {
 		t.Errorf("Error formatting size: expected '%s' got '%s'", "1 KB", size)
 	}
 
-	size = formattedSize(1024 * 1024 * 1024 * 1024 * 1024)
-	if size != "1024 TB" {
-		t.Errorf("Error formatting size: expected '%s' got '%s'", "1024 TB", size)
+	size = formattedSize(1000 * 1000 * 1000 * 1000)
+	if size != "1 TB" {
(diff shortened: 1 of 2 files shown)


rhatdan:
@TomSweeneyRedHat PTAL

rhatdan:
@umohnani8 @baude @mheon @pixdrift PTAL

cmd/buildah/images_test.go
```diff
@@ -74,14 +74,14 @@ func TestSizeFormatting(t *testing.T) {
...
-	size = formattedSize(1024)
+	size = formattedSize(1000)
 	if size != "1 KB" {
 		t.Errorf("Error formatting size: expected '%s' got '%s'", "1 KB", size)
 	}
 
-	size = formattedSize(1024 * 1024 * 1024 * 1024 * 1024)
-	if size != "1024 TB" {
-		t.Errorf("Error formatting size: expected '%s' got '%s'", "1024 TB", size)
+	size = formattedSize(1000 * 1000 * 1000 * 1000)
```
mheon: Did you lose a * 1000 in here? The original multiplies 1000 five times, this only does 4

rhatdan: Yes that was intentional.  I saw no reason to have 1000 TB.

mheon: Should you change the expected below to be 1TB instead of 1000TB then?

rhatdan: Fixed

umohnani8:
LGTM

rhatdan:
@rh-atomic-bot r=umohnani8

Issue #593, opened 2018-04-16
/bin/bash: cannot execute binary file


jwendell:
Generated Image seems to be corrupt.

```sh
ctr=$(buildah from centos:7)
buildah config --workingdir /tmp $ctr
buildah config --entrypoint /bin/bash $ctr
buildah commit $ctr test1

podman run --rm -it test1
/bin/bash: /bin/bash: cannot execute binary file
```
A simple example like this works fine in docker:
```Dockerfile
FROM centos:7
ENTRYPOINT /bin/bash
WORKDIR /tmp
```

**Output of `rpm -q buildah` or `apt list buildah`:**
```
buildah-0.16-1.git82529cd.fc27.x86_64
```

**Output of `buildah version`:**
```
Version:       0.16
Go Version:    go1.9.4
Image Spec:    1.0.0
Runtime Spec:  1.0.0
Git Commit:    82529cd
Built:         Wed Apr  4 11:22:14 2018
OS/Arch:       linux/amd64
```

**Output of `cat /etc/*release`:**
```
Fedora release 27 (Twenty Seven)
```

**Output of `uname -a`:**
```
Linux jw-laptop 4.15.16-300.fc27.x86_64 #1 SMP Mon Apr 9 17:50:06 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

**Output of `cat /etc/containers/storage.conf`:**

```
# storage.conf is the configuration file for all tools
# that share the containers/storage libraries
# See man 5 containers-storage.conf for more information

# The "container storage" table contains all of the server options.
[storage]

# Default Storage Driver
driver = "overlay"

# Temporary storage location
runroot = "/var/run/containers/storage"

# Primary Read/Write location of container storage
graphroot = "/var/lib/containers/storage"

[storage.options]
# AdditionalImageStores is used to pass paths to additional Read/Only image stores
# Must be comma separated list.
additionalimagestores = [
]

# Size is used to set a maximum size of the container image.  Only supported by
# certain container storage drivers.
size = ""

# OverrideKernelCheck tells the driver to ignore kernel checks based on kernel version
override_kernel_check = "true"
```

pixdrift:
This looks like a result of this pull request / change in entrypoint behaviour:
https://github.com/projectatomic/buildah/pull/577

It looks like podman is passing /bin/bash in to execute and because the entrypoint is set it is being passed as the first parameter to the entrypoint (/bin/bash).

[...]

PR #596, opened 2018-04-17
Add /bin/sh -c to entrypoint in config

TomSweeneyRedHat:

Prepend ''/bin/sh -c" to the entrypoint value given by the user for the `buildah config` command.  This addresses #593 

Tests:
```
# ctr=$(buildah from centos:7)
# buildah config --workingdir /tmp $ctr
# buildah config --entrypoint /bin/bash $ctr

# buildah commit $ctr test1
Skipping fetch of repeat blob sha256:43e653f84b79ba52711b0f726ff5a7fd1162ae9df4be76ca1de8370b8bbf9bb0
Copying config sha256:ae0382211cbf19f667ef1b3bbcc97b4195ebc1d5119813ca90cdad78eba83371
 1.24 KiB / 1.24 KiB [======================================================] 0s
Storing signatures

# buildah from test1
test1-working-container

# buildah run test1-working-container
[root@mrsdalloway tmp]# 
```

### DIFF
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -135,13 +135,16 @@ func updateConfig(builder *buildah.Builder, c *cli.Context) {
 		}
 	}
 	if c.IsSet("entrypoint") {
-		entrypointSpec, err := shellwords.Parse(c.String("entrypoint"))
-		if err != nil {
-			logrus.Errorf("error parsing --entrypoint %q: %v", c.String("entrypoint"), err)
+		entrypointSpec := make([]string, 3)
+		entrypointSpec[0] = "/bin/sh"
+		entrypointSpec[1] = "-c"
+		entrypointSpec[2] = c.String("entrypoint")
+		if len(strings.TrimSpace(c.String("entrypoint"))) == 0 {
+			builder.SetEntrypoint(nil)
 		} else {
 			builder.SetEntrypoint(entrypointSpec)
-			builder.SetCmd(nil)
 		}
+		builder.SetCmd(nil)
 	}
 	// cmd should always run after entrypoint; setting entrypoint clears cmd
 	if c.IsSet("cmd") {
--- a/tests/config.bats
+++ b/tests/config.bats
@@ -125,3 +125,11 @@ load helpers
   buildah --debug=false inspect --type=image --format '{{.ImageAnnotations}}' scratch-image-oci | grep ANNOTATION:VALUE
   buildah --debug=false inspect --type=image --format '{{.ImageAnnotations}}' scratch-image-oci | grep ANNOTATION:VALUE
 }
+
+@test "config with entrypoint" {
+  ctr=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+  buildah config --workingdir /tmp $ctr
+  buildah config --entrypoint /bin/sh $ctr
+  buildah commit --signature-policy ${TESTSDIR}/policy.json $ctr test1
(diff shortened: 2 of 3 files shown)


rhatdan:
Can you add a test to verify the problem that this is fixing?

TomSweeneyRedHat:
the buildah run tests aren't happy with this change.  For instance line 83 in run.bats is failing as the value being returned is " " rather than "pwd".  Trying to figure out what's what, but am also spending more time on the BZ that I has with docker ps.  Hope to wrap the bz early this afternoon, then chase the run tests down.

TomSweeneyRedHat:
bot, retest this please

cmd/buildah/config.go
```diff
@@ -135,13 +135,15 @@ func updateConfig(builder *buildah.Builder, c *cli.Context) {
...
-		entrypointSpec, err := shellwords.Parse(c.String("entrypoint"))
-		if err != nil {
-			logrus.Errorf("error parsing --entrypoint %q: %v", c.String("entrypoint"), err)
-		} else {
-			builder.SetEntrypoint(entrypointSpec)
-			builder.SetCmd(nil)
+		entrypointSpec := make([]string, 3)
+		entrypointSpec[0] = "/bin/sh"
+		entrypointSpec[1] = "-c"
+		entrypointSpec[2] = c.String("entrypoint")
```
ripcurld: I am curious why we are not validating `c.String("entrypoint")` as we used to before.

In addition, what happens if someone puts in `entrypoint` such a string `"ls -lha /tmp"`. Can you add a test for such a case?

TomSweeneyRedHat: @ripculd I'm not 100% sure what you mean by "not validating".  I think you mean not calling shellwords.Parse() on the string any more.  The reason for that change is illustrated nicely by your example.    With the current code, if you were to inspect the container after setting the entrypoint as you suggested you'll see:

```
            "Entrypoint": [
                "/bin/sh",
                "-c",
                "ls -lha /tmp"
```
which matches Docker and the result if you use ENTRYPOINT in the Dockerfile passed to `buildah bud`.  Prior to this change when we called shellwords.Parse, the results of `buildah inspect` would be:

```
            "Entrypoint": [
                "/bin/sh",
                "-c",
                "ls",
                "-lha", 
                "/tmp"
```
Which did not match Docker and with some entrypoints such as "ping www.google.com" would cause an unexpected result or failure.  

I'll see if I can work something up for testing.

```
#######
Test:
# buildah config --entrypoint "ls -lha /tmp" $cid
# buildah run $cid
total 0
drwxrwxrwt    2 root     root           6 Jan  9 19:37 .
drwxr-xr-x    1 root     root          28 Apr 23 13:31 ..
```

ripcurld: Got it :bulb: ! Thanks for the clarification.

TomSweeneyRedHat:
@rhatdan @nalind PTAL.  I've messed up my sandbox.  If this is OK I'd like to merge this and I'll add the requested test in  a separate PR.

cmd/buildah/config.go
```diff
@@ -135,13 +135,15 @@ func updateConfig(builder *buildah.Builder, c *cli.Context) {
...
+		builder.SetEntrypoint(entrypointSpec)
```
rhatdan: Nit:  Should this be the else clause of the if statement below.  Saves a command.

rhatdan:
LGTM other then the nit.

TomSweeneyRedHat:
Latest patch updates the nit by @rhatdan and added the requested test for @ripcurld

PR #603, opened 2018-04-18
Add FIPS mode secret to buildah run and bud

umohnani8:
If the host is in FIPS mode and /etc/system-fips exists
/run/secrets/system-fips is created in the container so that
the container can run in FIPS mode as well.
Vendor in libpod/pkg/secrets and remove the duplicate secrets code
in buildah.
Also remove the hidden --default-mounts-file flag that was being used for test,
as it is not needed anymore and makes the code simpler.

Adds the buildah side of this card: https://trello.com/c/UawSMhdW/572-openssl-fips-mode-in-containers


umohnani8:
@rhatdan PTAL

umohnani8:
Are we assuming that /run/secrets would have been mounted by the secrets patch? I was accounting for the case where no secrets are added but /etc/system-fips exists.

rhatdan:
Basically this should be one call to SecretMounts and that returns a mount point if the mounts.conf or the fips-mode flag is set on the host.

umohnani8:
Moved the FIPS mode support to `SecretMounts`. I left the `addFIPSModeSecret` function as that checks for the case there were no secret mounts in the mounts.conf file and the fips mode secret has the create /run/secrets as well as adding the system-fips file to it.
@rhatdan PTAL

run.go
```diff
@@ -211,6 +211,7 @@ func (b *Builder) setupMounts(mountPoint string, spec *specs.Spec, optionMounts
 			mounts = append(mounts, mount)
 		}
 	}
+
```
rhatdan: Remove whitespace

secrets.go
```diff
@@ -131,6 +131,7 @@ func getMountsMap(path string) (string, string, error) {
 // and returns a list of mounts
 func secretMounts(filePath, mountLabel, containerWorkingDir string) ([]rspec.Mount, error) {
 	var mounts []rspec.Mount
+
```
rhatdan: Remove whitespace

secrets.go
```diff
@@ -181,9 +182,59 @@ func secretMounts(filePath, mountLabel, containerWorkingDir string) ([]rspec.Mou
 
 		mounts = append(mounts, m)
 	}
+
+	// Add FIPS mode secret
+	_, err := os.Stat("/etc/system-fips")
+	// The filePath check is to ensure that this secret is created after the
+	// DefaultMountsFile has been looked at, i.e builder.DefaultMountsFile
+	if err == nil && filePath != OverrideMountsFile {
```
rhatdan: You want this to happen even if the OverrideMountsFile ispresent, I think

rhatdan:
@umohnani8 We should move to sharing libpod/pkg/secrets and make this change once and then allow all tools to share?

secrets.go
umohnani8: This check is here because SecretMounts is called twice, once for the override mounts and the other for default causing the FIPS secret code to happen twice. So I put the check to make sure it just happens one time.

umohnani8:
okay, so I should make this change in libpod and then vendor libpod/pkg/secrets into buildah and cri-o?

rhatdan:
Them move it up before the stat, add a comment so the next time I read this, I don't say that makes no sense and remove it.

rhatdan:
While you are in there replace fmt.Errorsf with errors,Wrapf

umohnani8:
tests pass, @rhatdan @nalind @TomSweeneyRedHat PTAL

cmd/buildah/main.go
```diff
@@ -61,12 +61,6 @@ func main() {
 			Usage: "storage driver option",
 			Value: defaultStoreDriverOptions,
 		},
-		cli.StringFlag{
```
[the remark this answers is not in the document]
umohnani8: Since we are using lipbod/pkg/secrets in buildah, the code doesn't support reading this path of the default mounts file from this. It just uses the actual override and default mount file paths, so there was no point in having this as we can't test with it.
I could modify libpod/pkg/secrets for this one case.

tests/secrets.bats
```diff
@@ -31,7 +31,21 @@ function setup() {
     [ "$status" -eq 0 ]
     [[ "$output" =~ "key.pem" ]]
     buildah rm $cid
-    rm -rf $TESTSDIR/containers
+    rm -rf /usr/share/containers
```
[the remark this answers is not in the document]
umohnani8: Was trying to replicate where the mounts.conf would live. I can leave it there and not delete it after the test? This is how the test is written for podman as well.

rhatdan:
@umohnani8 Needs a rebase.

rhatdan:
LGTM
@mheon @TomSweeneyRedHat @nalind @baude PTAL

rhatdan:
@rh-atomic-bot r+

PR #604, opened 2018-04-19
containers_test: Perfect testing

q384566678:
1.Rename some function names and distinguish them from test functions in
images_test.
2.Improve test content in containers_test.

### DIFF
--- a/cmd/buildah/containers_test.go
+++ b/cmd/buildah/containers_test.go
@@ -1,11 +1,12 @@
 package main
 
 import (
+	"fmt"
 	"strings"
 	"testing"
 )
 
-func TestTemplateOutputValidFormat(t *testing.T) {
+func TestContainerTemplateOutputValidFormat(t *testing.T) {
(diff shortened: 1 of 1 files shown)


cmd/buildah/containers_test.go
```diff
@@ -43,3 +44,64 @@ func TestTemplateOutputInvalidFormat(t *testing.T) {
 		t.Fatalf("expected error invalid format")
 	}
 }
+
+func TestContainerTemplateOutputInexistenceField(t *testing.T) {
```
rhatdan: I was not familiar with the work Inexistence, I would have used Nonexistent, but since it is a real word, I am fine with it.

rhatdan:
Can you amend you PR so the name does not have a typo.

s/contaienrs_test/containers_test

q384566678:
updated.

rhatdan:
@rh-atomic-bot r+
Thanks @q384566678

PR #605, opened 2018-04-19
buildah-image: add validation to 'format'

q384566678:
Same as #589.

Duplicate function name `TestTemplateOutputInvalidFormat` was modified in #604.

### DIFF
--- a/cmd/buildah/images.go
+++ b/cmd/buildah/images.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"regexp"
 	"strings"
 	"text/template"
 	"time"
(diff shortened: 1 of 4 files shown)


cmd/buildah/images.go
```diff
@@ -373,6 +374,13 @@ func formattedSize(size int64) string {
 }
 
 func outputUsingTemplate(format string, params imageOutputParams) error {
+	matched, err := regexp.MatchString("{{.*}}", format)
+	if err != nil {
```
rhatdan: Nit: can you combine these two lines together
```
if matched, err := regexp.MatchString("{{.*}}", format); err != nil {
...
```

q384566678: Fine, updated.

rhatdan:
LGTM

cmd/buildah/images_test.go
```diff
@@ -50,6 +50,23 @@ func TestTemplateOutputValidTemplate(t *testing.T) {
 	}
 }
 
+func TestTemplateOutputInvalidFormat(t *testing.T) {
```
TomSweeneyRedHat: I think you're having a collision with this function name and the one in the containers test

TomSweeneyRedHat:
Changes LGTM overall, looks like you need to tweak the name of the test function to make the tests happy.

cmd/buildah/images_test.go
q384566678: updated, PTAL.

rhatdan:
Thanks @q384566678 
@rh-atomic-bot r+

PR #610, opened 2018-04-20
Fix README example of buildah config

TristanCacqueray:
The example needs to use config argument before the container name,
otherwise it is failing with: too many arguments specified.


rhatdan:
bot, add author to whitelist

rhatdan:
@TristanCacqueray Thanks for the PR, but you need to sign it so we can merge.

git commit -a --ammend -s
git push --force

Should fix this problem

LGTM

TristanCacqueray:
Done, thanks.

rhatdan:
@rh-atomic-bot r+

PR #639, opened 2018-04-26
Add --compress, --rm, --squash flags as a noop for bud

rhatdan:
We don't have a client server operation, so there
is no reason to compress the input.

We don't currently support cache so --rm and --squash are NOOPS

### DIFF
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -41,6 +41,12 @@ Default certificates directory is _/etc/containers/certs.d_.
 
 Path to cgroups under which the cgroup for the container will be created. If the path is not absolute, the path is considered to be relative to the cgroups path of the init process. Cgroups will be created if they do not already exist.
 
+**--compress**
+
+This option is added to be aligned with other containers CLIs.
+Buildah doesn't communicate with a daemon or a remote server.
+Thus, compressing the data before sending it is irrelevant to Buildah.
+
 **--cpu-period**=*0*
 
 Limit the CPU CFS (Completely Fair Scheduler) period
@@ -174,6 +180,10 @@ Suppress output messages which indicate which instruction is being processed,
 and of progress when pulling images from a registry, and when writing the
 output image.
 
+**--rm**
+
+Remove intermediate containers after a successful build.  Buildah does not currently support cacheing so this is a NOOP.
+
 **--runtime** *path*
 
 The *path* to an alternate OCI-compatible runtime, which will be used to run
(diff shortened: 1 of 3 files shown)


rhatdan:
Fixes https://github.com/projectatomic/buildah/issues/621

docs/buildah-bud.md
```diff
@@ -41,6 +41,11 @@ Default certificates directory is _/etc/containers/certs.d_.
 
 Path to cgroups under which the cgroup for the container will be created. If the path is not absolute, the path is considered to be relative to the cgroups path of the init process. Cgroups will be created if they do not already exist.
 
+**--compress**
+
+This is legacy option, which has no effect on the image.  Buildah does not use a daemon. There
+is no reason to compress the input.
+
```
TomSweeneyRedHat: I'm finding this confusing and the Docker documentation doesn't enlighten much.  By just reading this, I wouldn't know why the lack of a Buildah daemon is important to this parameter.  Would this hold?:

"This is a legacy option in Docker which no longer has an effect on an image.  As Buildah already compresses the image, there is no need to further compress it."

rhatdan: I don't want to mention `Docker` at all.  As was explained to me, this flag compresses the imput data before sending it to the daemon.  If you were sitting on a MAC/Windows and had a big tar ball that you were doing a COPY or ADD, then this flag would cause the tar ball to the server, where it would be uncompressed.

pixdrift:
For all these flags that are noop, can we add a line of debug output saying as much when the user provides them as a parameter (when debug is specified)?

eg.
_--compress flag provided, but does not change buildah behaviour (NOOP)_

That way, if someone is expecting a change in behaviour when providing the parameter they can see it 'not working' as intentional and raise a bug/issue to resolve it?

docs/buildah-bud.md
TomSweeneyRedHat: OK, would you be hip with my wording then but just drop "in Docker"?

ripcurld: My suggestion:
```
This option is added to be aligned with other containers CLIs. 
Buildah doesn't communicate with a daemon or a remote server. 
Thus, compressing the data before sending it is irrelevant to Buildah.
```

rhatdan: I like @ripcurld Suggestion. Do not want to add Docker to the buildah docs if we can avoid it.

rhatdan:
@ripcurld @TomSweeneyRedHat Added --rm and --squash support also.  PTAL

pixdrift:
FWIW LGTM 👍

ripcurld:
LGTM :+1: 
Let's merge this.

rhatdan:
@rh-atomic-bot r+

PR #641, opened 2018-04-27
Add a method for setting a comment in to-be-created history items

nalind:
This patch adds `HistoryComment()` and `SetHistoryComment()` methods to the builder structure, which provide a way to manipulate which value we'll use when creating history items for images when we need them.

It stops providing a default for the `buildah config --created-by` option, so that if it's invoked a second time after setting the option, we won't override the previously-supplied value with our default.

It adds tests for --history-comment, --shell, --domainname, --hostname, and --comment.


rhatdan:
LGTM
@rh-atomic-bot r+

PR #646, opened 2018-04-30

umohnani8:
Fixes buildah from so a public image can be pulled even if
$XDG_RUNTIME_DIR does not exist. Public images don't require
authentication to be pulled.

Fixes https://github.com/projectatomic/buildah/issues/618


rhatdan:
@rh-atomic-bot r+

PR #647, opened 2018-05-01
Do not print directly to stdout in Commit()

mheon:
Buildah is being used as a library in Podman, and as such should avoid printing directly to standard streams. Instead, return an image ID which can be printed by the caller (if desired)

### DIFF
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"os"
 	"strings"
 	"time"
@@ -159,12 +160,15 @@ func commitCmd(c *cli.Context) error {
 	if !c.Bool("quiet") {
 		options.ReportWriter = os.Stderr
 	}
-	err = builder.Commit(ctx, dest, options)
+	id, err := builder.Commit(ctx, dest, options)
 	if err != nil {
 		return util.GetFailureCause(
 			err,
 			errors.Wrapf(err, "error committing container %q to %q", builder.Container, image),
 		)
+	}
+	if options.IIDFile == "" && id != "" {
+		fmt.Printf("%s\n", id)
 	}
 
 	if c.Bool("rm") {
--- a/commit.go
+++ b/commit.go
@@ -78,16 +78,19 @@ type PushOptions struct {
 
 // Commit writes the contents of the container, along with its updated
 // configuration, to a new image in the specified location, and if we know how,
-// add any additional tags that were specified.
-func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options CommitOptions) error {
+// add any additional tags that were specified. Returns the ID of the new image
+// if commit was successful and the image destination was local
+func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options CommitOptions) (string, error) {
+	var imgID string
+
 	systemContext := getSystemContext(options.SystemContext, options.SignaturePolicyPath)
 	policy, err := signature.DefaultPolicy(systemContext)
 	if err != nil {
-		return errors.Wrapf(err, "error obtaining default signature policy")
+		return imgID, errors.Wrapf(err, "error obtaining default signature policy")
 	}
 	policyContext, err := signature.NewPolicyContext(policy)
 	if err != nil {
-		return errors.Wrapf(err, "error creating new signature policy context")
+		return imgID, errors.Wrapf(err, "error creating new signature policy context")
 	}
 	defer func() {
 		if err2 := policyContext.Destroy(); err2 != nil {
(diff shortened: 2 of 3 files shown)


rhatdan:
LGTM

cmd/buildah/commit.go
```diff
@@ -155,13 +156,14 @@ func commitCmd(c *cli.Context) error {
...
 	}
-	err = builder.Commit(ctx, dest, options)
+	id, err := builder.Commit(ctx, dest, options)
 	if err != nil {
 		return util.GetFailureCause(
 			err,
 			errors.Wrapf(err, "error committing container %q to %q", builder.Container, image),
 		)
 	}
+	fmt.Printf("%s\n", id)
```
rhatdan:
```
if options.iidfile == "" {
fmt.Printf("%s\n", id) 
}
```

mheon: Oops, fixed.

commit.go
```diff
@@ -78,15 +78,16 @@ type PushOptions struct {
...
 // configuration, to a new image in the specified location, and if we know how,
-// add any additional tags that were specified.
-func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options CommitOptions) error {
+// add any additional tags that were specified. Returns the ID of the new image
+// if successful.
+func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options CommitOptions) (string, error) {
 	policy, err := signature.DefaultPolicy(getSystemContext(options.SystemContext, options.SignaturePolicyPath))
 	if err != nil {
-		return errors.Wrapf(err, "error obtaining default signature policy")
+		return "", errors.Wrapf(err, "error obtaining default signature policy")
```
ripcurld: Won't it be easier to declare `var imgID string` in the beginning of the function and return it instead of the empty string `""`?
Obviously in 126 after we get `img` we set `imgID` to `img.ID`?

WDYT?

mheon: @ripcurld Seems like a good way of avoiding the extensive use of the `""` constant

cmd/buildah/commit.go
```diff
@@ -155,13 +156,16 @@ func commitCmd(c *cli.Context) error {
...
 	}
-	err = builder.Commit(ctx, dest, options)
+	id, err := builder.Commit(ctx, dest, options)
 	if err != nil {
 		return util.GetFailureCause(
 			err,
 			errors.Wrapf(err, "error committing container %q to %q", builder.Container, image),
 		)
 	}
+	if options.IIDFile != "" {
```
TomSweeneyRedHat: should this be ==?  Tests are failing all over as it can't get an ID.

rhatdan:
LGTM

ripcurld:
LGTM :heavy_check_mark:

mheon:
This looks like https://github.com/projectatomic/buildah/blob/master/commit.go#L125 throwing an error. Before this patch, this was allowed to fail, and the ID just wouldn't print in that case.
@nalind is it expected that we sometimes can't retrieve the committed image like that? If so, returning the ID would have best-effort, and that error can be made nonfatal.

TomSweeneyRedHat:
@rhatdan fwiw, I think this one should also be in v1.0 if at all possible.

nalind:
It's expected that we can't read the image's ID when the destination isn't in local storage, since we're reading the ID from the record kept by local storage for an image with the target name.

mheon:
Alright, I'll change it to be a best-effort return

commit.go
```diff
@@ -123,16 +126,18 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
...
 
 	img, err := is.Transport.GetStoreImage(b.store, dest)
-	if err == nil {
-		if options.IIDFile != "" {
-			if err := ioutil.WriteFile(options.IIDFile, []byte(img.ID), 0644); err != nil {
-				return errors.Wrapf(err, "failed to write Image ID File %q", options.IIDFile)
-			}
-		} else {
-			fmt.Printf("%s\n", img.ID)
+	if err != nil {
```
TomSweeneyRedHat: @mheon, I've been playing/testing this this morning and unless @nalind yells,  you want to add something like:
```
# git diff
diff --git a/commit.go b/commit.go
index 82e57cd..401f790 100644
--- a/commit.go
+++ b/commit.go
@@ -126,6 +126,11 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
        }
 
        img, err := is.Transport.GetStoreImage(b.store, dest)
+
+       if (err == storage.ErrImageUnknown) {
+               return "", nil
+       }
+
        if err != nil {
                return imgID, errors.Wrapf(err, "failed to retrieve ID of committed image")
        }
```
The only thing a little wonky is with this in place is the 'buildah commit` leaves an extra blank line.

```
# git diff
diff --git a/commit.go b/commit.go
index 82e57cd..401f790 100644
--- a/commit.go
+++ b/commit.go
@@ -126,6 +126,11 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
        }
 
        img, err := is.Transport.GetStoreImage(b.store, dest)
+
+       if (err == storage.ErrImageUnknown) {
+               return "", nil
+       }
+
        if err != nil {
                return imgID, errors.Wrapf(err, "failed to retrieve ID of committed image")
        }
```

With the above in place, the tests pass.

commit.go
```diff
@@ -124,15 +127,16 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 
 	img, err := is.Transport.GetStoreImage(b.store, dest)
 	if err == nil {
+		imgID = img.ID
```
TomSweeneyRedHat: @mheon, made a comment in the prior rev just before you posted this rev.  I'd pref a check to see if the error is unknownimage and then error out if not and I did in my example.  Your call.

mheon: @TomSweeneyRedHat I think that's a good option, let me make the change and repush

TomSweeneyRedHat:
LGTM assuming happy tests.  Thx for working this through @mheon.

rhatdan:
@mheon needs a rebase

mheon:
@rhatdan Done

rhatdan:
@rh-atomic-bot r+

PR #648, opened 2018-05-01
Force "localhost" as a default registry

nalind:
Extend `util.ResolveName()` to prepend "localhost" to the list of registries, and teach `util.FindImage()`, `util.ExpandNames()`, and `util.AddImageNames()` to use `util.ResolveName()`.

### DIFF
--- a/cmd/buildah/rmi.go
+++ b/cmd/buildah/rmi.go
@@ -12,6 +12,7 @@ import (
 	"github.com/containers/storage"
 	"github.com/pkg/errors"
 	"github.com/projectatomic/buildah/pkg/parse"
+	"github.com/projectatomic/buildah/util"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 )
@@ -86,9 +87,13 @@ func rmiCmd(c *cli.Context) error {
 	}
 
 	ctx := getContext()
+	systemContext, err := parse.SystemContextFromOptions(c)
+	if err != nil {
+		return errors.Wrapf(err, "error building system context")
+	}
 
 	for _, id := range imagesToDelete {
-		image, err := getImage(ctx, id, store)
+		image, err := getImage(ctx, systemContext, id, store)
 		if err != nil || image == nil {
 			if lastError != nil {
 				fmt.Fprintln(os.Stderr, lastError)
@@ -148,7 +153,7 @@ func rmiCmd(c *cli.Context) error {
 
 				// Need to fetch the image state again after making changes to it i.e untag
 				// because only a copy of the image state is returned
-				image, err = getImage(ctx, image.ID, store)
+				image, err = getImage(ctx, systemContext, image.ID, store)
 				if err != nil || image == nil {
 					return errors.Wrapf(err, "error getting image after untag %q", image.ID)
 				}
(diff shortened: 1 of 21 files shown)


rhatdan:
@mtrmac @runcom We have decided to use localhost/ to represent locally created images. We know this potentially conflicts with users setting up a registry at localhost, But the user has an easy work around if this causes an issue.  IE I can not tell the difference between locally created images and images I pulled from localhost, by setting up the registry with a different name like localhost.com or something and adding an entry to the /etc/hosts file.

This is much better then what we have now in that we can't tell the difference between images created at locally and those pulled from docker.io.  It also eliminates the need for a breaking change to containers/image.

@baude @mheon These changes will also need to be made to podman once we have it working in buildah.  

Note: Nalin has also found lots of issues where we were not looking through the registries for certain image interactions.

cmd/buildah/rmi.go
```diff
@@ -263,16 +268,12 @@ func properImageRef(ctx context.Context, id string) (types.ImageReference, error
...
-			img.Close()
-			return ref, nil
-		}
+func storageImageRef(ctx context.Context, systemContext *types.SystemContext, store storage.Store, id string) (types.ImageReference, error) {
+	ref, _, err := util.FindImage(store, "", systemContext, id)
+	if err != nil {
 		return nil, errors.Wrapf(err, "error confirming presence of storage image reference %q", transports.ImageName(ref))
 	}
-	return nil, errors.Wrapf(err, "error parsing %q as a storage image reference", id)
+	return ref, err
```
rhatdan: Shouldn't this be 
```
return ref, nil
```
Just to make it clearer that no error happened

rhatdan:
BTW I believe this is the last patch we need to get in so we can declare Buildah 1.0  :1st_place_medal:

rhatdan:
@giuseppe @umohnani8 @TomSweeneyRedHat @mtrmac @baude @runcom PTAL

nalind:
Note that when we're given an image name that doesn't include a registry component, we'll attempt to pull from a registry running on localhost first, before consulting registries listed in the configuration.  If the configuration doesn't exist, then, we'll just fail to pull an image.

TomSweeneyRedHat:
We'll need to document the registry on the localhost if it doesn't work like the multituedd of Docker examples.  There are a number of examples for running a Docker registry on the web, will Buildah no longer talk to those registries without adjustments?  @ipbabble also  wrote a blog on how to set up a local and we'll need to make sure that holds or add an addendum to that blog.

TomSweeneyRedHat:
Travis isn't happy @nalind

rhatdan:
There are several fixes in this patch suite to make sure that buildah uses the registries.conf properly.  So yes this change will not effect what is going on now, unless you don't have a properly populated registries.conf.  But I have a feeling our previous code was pulling from docker.io even if docker.io was not listed in the registry.conf.  Which would be a big bug.

rhatdan:
@TomSweeneyRedHat This should work just fine with a localhost registry.  The only change would be if I do 
```
ctr=buildah from scratch
buildah commit $ctr foobar
```
I will end up with localhost/foobar as my image name.

If I had the same named package in a localhost registry and did a 

```
buildah from foobar
````
Then the original localhost/foobar would be replaced by the one pulled down.

This should not effect pushing, since you need to specify a path to push to.

umohnani8:
LGTM

rhatdan:
@rh-atomic-bot r+

I agree I think we need to sprinkle information in all of the functions that do pull or push.

cgwalters:
This really confused me a lot as I didn't understand why my container build scripts weren't picking up changes, turned out of course they were running the `docker.io/` version - looks like this change landed in the Silverblue release stream recently.

Eh, yeah I guess it's better than doing `docker.io` anyways.

rhatdan:
Sorry about that, tough to change the existing databases.

PR #649, opened 2018-05-01
Drop util.GetLocalTime()

nalind:
util.GetLocalTime() duplicates time.Time.Local(), so drop it.

### DIFF
--- a/cmd/buildah/images.go
+++ b/cmd/buildah/images.go
@@ -15,7 +15,6 @@ import (
 	"github.com/containers/storage"
 	"github.com/pkg/errors"
 	"github.com/projectatomic/buildah/pkg/parse"
-	"github.com/projectatomic/buildah/util"
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 	"golang.org/x/crypto/ssh/terminal"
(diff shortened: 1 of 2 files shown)


cmd/buildah/images.go
```diff
@@ -233,7 +232,7 @@ func outputImages(ctx context.Context, images []storage.Image, format string, st
 				createdTime = inspectedTime
 			}
 		}
-		createdTime = util.GetLocalTime(createdTime)
+		createdTime = createdTime.Local()
```
ripcurld: Won't this change the behavior?
Before we were calculating the duration when timezone is `UTC` and now we are using the local timezone?

Or I didn't understand the code correctly? :grinning:

nalind: We were calculating an offset from UTC using the local timezone information and offsetting the value of "now" by that amount.  If the format string we were using for display later in `outputImages()` included a timezone specifier ("MST"), it would have been misleading.

ripcurld: Thanks for clarifying this.

LGTM

rhatdan:
LGTM
@rh-atomic-bot r+

PR #652, opened 2018-05-02
Touch up auth test commands

TomSweeneyRedHat:

When testing an authentication PR earlier today, I realized that a number of the commands had the target in the middle of the command rather than at the end.  Fixed that up and touched up a comment.

This test is not called during CI testing.  It's only used for testing on VM's from time to time.


rhatdan:
@rh-atomic-bot r+

Issue #622, opened 2018-04-24
buildah bud --force-rm functionality is missing


TomSweeneyRedHat:
**Description**
`buildah bud --force-rm` is missing and needs to be provided.  This is functionality that Docker has.

#599 is the parent issue for this issue.

**Output of `buildah version`:**

Buildah v0.16

rhatdan:
Since we don't have caching this one can not be implemented yet.

pixdrift:
Blocked by #601

pixdrift:
With PR #639 implementing NOOP for `--rm`, the same can probably be done for this flag until caching is implemented in #601.

umohnani8:
Implemented in https://github.com/projectatomic/buildah/pull/788

PR #659, opened 2018-05-06
Adding noop for --force-rm to match --rm

pixdrift:
This adds noop for the `--force-rm` flag in the same way that noop was added for `--rm` in PR #639 .

This is an intermediate fix for #622 .


pixdrift:
Needs another minor fix, will commit now.

pixdrift:
Just spotted a typo.. 'cacheing'.. can't handle this PR being merged with that present, I will update all the noops.

PR #663, opened 2018-05-07
WIP - Add CI tests for cache related noop flags

pixdrift:
Tests to ensure that the cache related flags currently configured for noop are consumed by `buildah bud` but don't change its behaviour.


pixdrift:
Interesting, the --cache-from failed which is the only noop flag with an option provided. Will take a closer look at this as I thought it passed in my local tests.

PR #670, opened 2018-05-09
Add cpu-shares short flag (-c) and cpu-shares CI tests

pixdrift:
Short flag `-c` was missing for `--cpu-shares` in `buildah bud`, this adds the flag to improve compatibility with `docker build`. CI tests included for flag options.

### DIFF
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -65,7 +65,7 @@ Limit the container's CPU usage. By default, containers run with the full
 CPU resource. This flag tell the kernel to restrict the container's CPU usage
 to the quota you specify.
 
-**--cpu-shares**=*0*
+**-c, --cpu-shares**=*0*
 
 CPU shares (relative weight)
 
--- a/pkg/cli/common.go
+++ b/pkg/cli/common.go
@@ -118,7 +118,7 @@ var (
 			Usage: "limit the CPU CFS (Completely Fair Scheduler) quota",
 		},
 		cli.Uint64Flag{
-			Name:  "cpu-shares",
+			Name:  "cpu-shares, c",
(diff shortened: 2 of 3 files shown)


docs/buildah-bud.md
```diff
@@ -65,7 +65,7 @@ Limit the container's CPU usage. By default, containers run with the full
 CPU resource. This flag tell the kernel to restrict the container's CPU usage
 to the quota you specify.
 
-**--cpu-shares**=*0*
+**-c, --cpu-shares**=*0*
```
TomSweeneyRedHat: Just wondering and not a thought or change needed for this PR.  We seem to list the short version first, then the fully specified version.  Alphabetically this throws the list out of order with this change.  I think it might be better if we list the full name first, then the short name as we do in the usage statement and as Docker does in their docs.  Currently our docs seem to follow short first, then fully specified so @pixdrift is following the right pattern.  Thoughts?

TomSweeneyRedHat:
LGTM assuming happy green test buttons.  I'd one generic question that should not hold up this PR, but may result in a small issue.

docs/buildah-bud.md
pixdrift: Agree they should be switched, was following convention. Can raise a separate PR to switch them all at the same time if you like.

pixdrift: @TomSweeneyRedHat, is it blocked on this?

TomSweeneyRedHat:
@pixdrift, no not blocked on this.  However @nalind or @rhatdan usually do the approval/merges.  Both are a bit busy this week so hopefully they'll get a looksee soon.

rhatdan:
@rh-atomic-bot r+
Lets fix the docs in a different PR>

Issue #658, opened 2018-05-04
Buildah build-using-dockerfile ONBUILD support


clcollins:
**Description**

`buildah bud` does not appear to support Dockerfile ONBUILD commands.  As a drop-in replacement for Docker, Buildah is working great, except for this example.  I'd like to get developers to migrate away from the Docker daemon, and "build-using-dockerfile" is the painless way to get them started down that road.  The ONBUILD instruction is a somewhat widely adopted Dockerfile feature, however.

**Steps to reproduce the issue:**
1. Create a Dockerfile with ONBUILD instructions
2. Build an image from the Dockerfile with `buildah bud`
3. Create a Dockerfile "FROM" the image created in the last step
4. Build an image from the new Dockerfile with `buildah bud`
5. Verify that the ONBUILD instruction has not occurred in the new image

_Example:_
```
$ cat << EOF > Dockerfile
FROM fedora:28
RUN touch /foo
ONBUILD RUN touch /bar
EOF

$ sudo buildah bud -f Dockerfile -t onbuild-image .
STEP 1: FROM fedora:28
STEP 2: RUN touch /foo
STEP 3: ONBUILD RUN touch /bar
STEP 4: COMMIT containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay2.override_kernel_check=true]docker.io/library/onbuild-image:latest
Getting image source signatures
Skipping fetch of repeat blob sha256:72fcdba8cff9f105a61370d930d7f184702eeea634ac986da0105d8422a17028
Copying blob sha256:10d17c14659832afdc927fd9028f3cbf2c23159d112302f7b23f7f14bb7973f8
 165 B / 165 B [============================================================] 0s
Copying config sha256:b1d7fb89c28825240f4af5bd32b976d0fbad8f3fae5f008588b12dea6bda5d2e
 1.45 KiB / 1.45 KiB [======================================================] 0s
Writing manifest to image destination
Storing signatures

$ cat << EOF > Dockerfile-2
FROM onbuild-image
RUN touch /baz
EOF

$ sudo buildah bud -f Dockerfile-2 -t result-image .
STEP 1: FROM onbuild-image
STEP 2: RUN touch /baz
STEP 3: COMMIT containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay2.override_kernel_check=true]docker.io/library/result-image:latest
Getting image source signatures
[...]

PR #686, opened 2018-05-14
Add OnBuild support for Dockerfiles

TomSweeneyRedHat:

This is a WIP to hopefully allow @nalind to spot what I'm messing up with this.  The image is not getting the OnBuild value attached to it, and I'm not completely sure why.  Hoping Nalin can quickly spot my mistake and give me a resounding head slap.

### DIFF
--- a/config.go
+++ b/config.go
@@ -331,6 +331,17 @@ func (b *Builder) SetUser(spec string) {
 	b.Docker.Config.User = spec
 }
 
+// OnBuild returns the OnBuild value from the container.
+func (b *Builder) OnBuild() []string {
+	return copyStringSlice(b.Docker.Config.OnBuild)
+}
+
+// SetOnBuild sets a trigger instruction to be executed when the image is used
+// as the base of another image.
+func (b *Builder) SetOnBuild(onBuild []string) {
+	b.Docker.Config.OnBuild = copyStringSlice(onBuild)
+}
+
 // WorkDir returns the default working directory for running commands in the
 // container, or in a container built using an image built from this container.
 func (b *Builder) WorkDir() string {
@@ -348,7 +359,7 @@ func (b *Builder) SetWorkDir(there string) {
 // Shell returns the default shell for running commands in the
 // container, or in a container built using an image built from this container.
 func (b *Builder) Shell() []string {
-	return b.Docker.Config.Shell
+	return copyStringSlice(b.Docker.Config.Shell)
 }
 
 // SetShell sets the default shell for running
@@ -357,7 +368,7 @@ func (b *Builder) Shell() []string {
 // Note: this setting is not present in the OCIv1 image format, so it is
 // discarded when writing images using OCIv1 formats.
 func (b *Builder) SetShell(shell []string) {
-	b.Docker.Config.Shell = shell
+	b.Docker.Config.Shell = copyStringSlice(shell)
 }
 
 // Env returns a list of key-value pairs to be set when running commands in the
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -155,6 +155,8 @@ type Executor struct {
 	labels                         []string
 }
 
+var onBuildCmd string
+
 // withName creates a new child executor that will be used whenever a COPY statement uses --from=NAME.
 func (b *Executor) withName(name string, index int) *Executor {
 	if b.named == nil {
@@ -539,6 +541,7 @@ func (b *Executor) Prepare(ctx context.Context, ib *imagebuilder.Builder, node *
 	for _, v := range builder.Volumes() {
 		volumes[v] = struct{}{}
 	}
+	onBuildCmdSlice := []string{onBuildCmd}
 	dConfig := docker.Config{
 		Hostname:   builder.Hostname(),
 		Domainname: builder.Domainname(),
@@ -552,6 +555,7 @@ func (b *Executor) Prepare(ctx context.Context, ib *imagebuilder.Builder, node *
 		Labels:     builder.Labels(),
 		Shell:      builder.Shell(),
 		StopSignal: builder.StopSignal(),
+		OnBuild:    onBuildCmdSlice,
 	}
 	var rootfs *docker.RootFS
 	if builder.Docker.RootFS != nil {
@@ -669,6 +673,7 @@ func (b *Executor) Commit(ctx context.Context, ib *imagebuilder.Builder) (err er
 	for v := range config.Volumes {
 		b.builder.AddVolume(v)
 	}
+	b.builder.SetOnBuild(config.OnBuild)
 	b.builder.SetWorkDir(config.WorkingDir)
 	b.builder.SetEntrypoint(config.Entrypoint)
 	b.builder.SetShell(config.Shell)
@@ -791,6 +796,13 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
 		return errors.Wrapf(err, "error creating build executor")
 	}
 	b := imagebuilder.NewBuilder(options.Args)
+	// If the this is the ONBUILD line, note the value less the first
+	// eight characters "ONBUILD " retaining only the command.
+	for _, node := range mainNode.Children {
+		if node.Value == "onbuild" && len(node.Original) > 8 {
+			onBuildCmd = node.Original[8:]
+		}
+	}
 	stages := imagebuilder.NewStages(mainNode, b)
 	return exec.Build(ctx, stages)
 }


commit.go
```diff
@@ -104,11 +104,27 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
...
 		return imgID, errors.Wrapf(err, "error copying layers and metadata")
 	}
+	//TOM: Start
+	cdest, ok := dest.(*containerImageRef) 
+	if ok {
+		fmt.Printf("TOM: dest  onbuild is %q\n", cdest.onBuild)
+	}
+	dest.DockerReference().Config.OnBuild = csrc.onBuild
+	//TOM: end
+	fmt.Printf("TOM: AFTER dest onBuild )\n")
```
TomSweeneyRedHat: This destination image never comes back OK.  I thought it would?  I was hoping I could push the OnBuild value into it here, but I'm not sure why I can't get to it.

image.go
```diff
@@ -154,6 +157,7 @@ func (i *containerImageRef) NewImageSource(ctx context.Context, sc *types.System
 		return nil, err
 	}
 	dimage.Created = created
+	dimage.Config.OnBuild = i.onBuild
```
TomSweeneyRedHat: This was my other attempt to get OnBuild into the destination image, but it too doesn't seem to take.

imagebuildah/build.go
```diff
@@ -548,6 +550,8 @@ func (b *Executor) Prepare(ctx context.Context, ib *imagebuilder.Builder, node *
 		Labels:     builder.Labels(),
 		Shell:      builder.Shell(),
 		StopSignal: builder.StopSignal(),
+		//OnBuild:    tomString,
+		OnBuild:    builder.GetOnBuild(),
```
TomSweeneyRedHat: OK here's the weird stuff.  If I set OnBuild to tomString, most everything works.  The example that @clcollins sent us runs fine.  Unfortunately at this point we have only read the FROM line in the Dockerfile and the builder's OnBuild is not yet set.  The strange part here is if I use tomString for the value, the example still runs however when I do `buildah inspect {image}` the two OnBuild values (OCI and Docker) are both listed as null.

buildah.go
```diff
@@ -97,6 +97,8 @@ type Builder struct {
 	Docker docker.V2Image `json:"docker,omitempty"`
 	// DefaultMountsFilePath is the file path holding the mounts to be mounted in "host-path:container-path" format
 	DefaultMountsFilePath string `json:"defaultMountsFilePath,omitempty"`
+	// Adds a trigger instruction to be executed when the image is used as the base of another image.
+	OnBuild           []string            `json:"OnBuild,omitempty"`
```
nalind: No need for a specific field here - the value should be accessible in `Docker.Config.OnBuild`.

commit.go
nalind: The destination isn't a `*containerImageRef`, since it's a reference to an image that's accessed via a transport that's supported by the image library, and image references are usually obtained by asking transport objects by way of `alltransports.ParseImageName()`, which uses the initial portion of the name to select which `ImageTransport` implementation to ask. (The `containers-storage` transport offers additional transport-specific methods for people who ask its `Transport` object directly.)

The `src` reference is a `*containerImageRef`, but it isn't obtained through a transport.  It's just enough of an `ImageReference` implementation that the image library can ask it for an `ImageSource`, which in turn is just enough of an `ImageSource` implementation that the image library can examine it and copy it.

config.go
```diff
@@ -331,6 +331,17 @@ func (b *Builder) SetUser(spec string) {
 	b.Docker.Config.User = spec
 }
 
+// OnBuild returns the OnBuild value from the container.
+func (b *Builder) GetOnBuild() []string {
+	return b.Docker.Config.OnBuild
```
nalind: Please rename this to `OnBuild()`, and use `copyStringSlice()` to make a copy of the value to return.  We don't want people to think that modifying the slice that they get back is a supported way of changing the value we're holding.

config.go
```diff
@@ -331,6 +331,17 @@ func (b *Builder) SetUser(spec string) {
...
+}
+
+// SetOnBuild sets a trigger instruction to be executed when the image is used
+// as the base of another image.
+func (b *Builder) SetOnBuild(onBuild []string) {
+	b.Docker.Config.OnBuild = onBuild
```
nalind: Please use `copyStringSlice()` here to make a copy of `onBuild`.  We don't want people to modify the slice after we get it, and assuming that that's a supported way of updating the value we're holding.  (It looks like `Shell()` and `SetShell()` need similar fixes.  It's not a part of this PR, but I don't want to forget to fix that.)

image.go
nalind: This shouldn't be necessary -- we default to preserving any value that's been set there when reading the image in, and saving the value that's in the structure when writing the config blob as part of committing the image.

image.go
```diff
@@ -528,6 +533,8 @@ func (b *Builder) makeImageRef(manifestType string, exporting bool, compress arc
 		annotations:           b.Annotations(),
 		preferredManifestType: manifestType,
 		exporting:             exporting,
+		onBuild:               b.GetOnBuild(),
```
nalind: This shouldn't be necessary - the value should be written to the image along with the rest of the image configuration.

imagebuildah/build.go
TomSweeneyRedHat: This is my Catch-22 that I'm still trying to figure out..  At this point I've not yet processed (or at least I don't know how to access it) the value that was in the ONBUILD line in the Dockerfile.  If I can figure out what that value is somehow at this point, I can jam it into the 'OnBuild:" line here at 554 and then things work well.  

Thanks for the heads up on the OnBuild field being null in the inspect, I thought that wasn't supposed to be like that once I got this working.

image.go
TomSweeneyRedHat: oops, forgot to note that I've removed this.

buildah.go
TomSweeneyRedHat: Removed this too.   Hmm, what happened to my comments GitHub? ...

config.go
TomSweeneyRedHat: fixed (another missing comment)

TomSweeneyRedHat:
Test Results: 
```
# buildah bud -f ~/Dockerfile -t onbuild-image 
STEP 1: FROM fedora:28
STEP 2: RUN touch bar
STEP 3: ONBUILD RUN touch bar
STEP 4: RUN touch /foo
STEP 5: COMMIT containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.override_kernel_check=true]localhost/onbuild-image:latest
Skipping fetch of repeat blob sha256:891e1e4ef82ad02a4ea1f030831f942d722c7694c4db64ca3239c8163b811c58
Copying blob sha256:3bf52e4f7f804cdf869125bc256fe29e36ce9dc3c44913c132d5b418b74191bb
 147 B / 147 B [============================================================] 0s
Copying config sha256:42cda3f6e105725f507146e1b7ce8a7aad41971567316086149eaea6f9628dba
 1.45 KiB / 1.45 KiB [======================================================] 0s
Storing signatures
42cda3f6e105725f507146e1b7ce8a7aad41971567316086149eaea6f9628dba

# buildah bud -f ~/Dockerfile-2 -t result-image .
STEP 1: FROM onbuild-image
STEP 2: RUN touch /baz
STEP 3: COMMIT containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.override_kernel_check=true]localhost/result-image:latest
Skipping fetch of repeat blob sha256:891e1e4ef82ad02a4ea1f030831f942d722c7694c4db64ca3239c8163b811c58
Skipping fetch of repeat blob sha256:aaead2948a115bdbfc959afa6b15783eec2e7857bde51d9831370cd65b8ce9e1
Copying blob sha256:2e101ff2bd58cf07381fdb4167efba61e215fc48b6e72c5edaf2acfbe0d766c6
 91 B / 91 B [==============================================================] 0s
Copying config sha256:05b73d6d687068c4f9a5894717583a8e2244568f46e0adfec1bd4b4539c027b9
 1.75 KiB / 1.75 KiB [======================================================] 0s
Storing signatures
05b73d6d687068c4f9a5894717583a8e2244568f46e0adfec1bd4b4539c027b9

# ctr=$(buildah from result-image)

# buildah run $ctr ls /bar
/bar
```

TomSweeneyRedHat:
Addresses #658

pixdrift:
@TomSweeneyRedHat 

In this execution
`# buildah bud -f ~/Dockerfile-2 -t result-image .`

Shouldn't the output show the user that it is firing the `ONBUILD RUN` as the first RUN command?

rhatdan:
@pixdrift That is my understanding.  I would want to to make sure that the commands run on the second buildah bud or docker build from the container image.

TomSweeneyRedHat:
I''ll have to play around with Docker to see what exactly is going on there.  As an FYI, the ONBUILD command is creating the `/bar` directory in the first (onbuild-image) container.  Just for the record, that creation was not in the Dockerfile via a RUN command.  I think the end result is correct, just the road travelled may be off, but TBD on that.  Here are the Dockerfiles for reference:

```
# cat Dockerfile
FROM fedora:28
RUN touch /foo
ONBUILD RUN touch /bar

# cat Dockerfile-2
FROM onbuild-image
RUN touch /baz
```

TomSweeneyRedHat:
Docker run.  Note the `# Executing 1 build trigger...` line in the first build.  The /bar directory is only present in the second image, not the first.

```
# docker build  -t onbuild-image .
Sending build context to Docker daemon  2.72 GB
Step 1/3 : FROM fedora:28
Trying to pull repository docker.io/library/fedora ... 
sha256:7ae08e5637170eb47c01e315b6e64e0d48c6200d2942c695d0bee61b38c65b39: Pulling from docker.io/library/fedora
e71c36a80ba9: Pull complete 
Digest: sha256:7ae08e5637170eb47c01e315b6e64e0d48c6200d2942c695d0bee61b38c65b39
Status: Downloaded newer image for docker.io/fedora:28
 ---> cc510acfcd70
Step 2/3 : RUN touch /foo
 ---> Running in a8a45d58d3a7
 ---> 9d475a707a1d
Removing intermediate container a8a45d58d3a7
Step 3/3 : ONBUILD run touch /bar
 ---> Running in 0c8f9fbaea11
 ---> 5a47742d391a
Removing intermediate container 0c8f9fbaea11
Successfully built 5a47742d391a

# docker build -f Dockerfile-2 -t result-image .
Sending build context to Docker daemon  2.72 GB
Step 1/2 : FROM onbuild-image
# Executing 1 build trigger...
Step 1/1 : RUN touch /bar
 ---> Running in 6b7f21a96a7e
 ---> c179eb6e01ee
Removing intermediate container 6b7f21a96a7e
Step 2/2 : RUN touch /baz
 ---> Running in 0aec140ff8ce
 ---> 301e60c5f24a
Removing intermediate container 0aec140ff8ce
Successfully built 301e60c5f24a

# docker run result-image ls /bar
/bar

# docker run onbuild-image ls /bar
ls: cannot access '/bar': No such file or directory
```

rhatdan:
@TomSweeneyRedHat And how does buildah bud compare?

TomSweeneyRedHat:
Had a thought on my way home and unhappily confirmed it.  The bar file is getting created in the first container, not in the second container/image vi the OnBuild command.  Because the file itself got copied during the second image build from the fist image, it looked like OnBuild worked.  In reality if the OnBuild command had been "date >> /tmp/date.txt" the value would be the oriingal date, not an updated one.  Crud.  I've  a suspicion that something in the imagebuilder code isn't quite right.

TomSweeneyRedHat:
@rhatdan.  What's happening with this current code is the OnBuild is set for the first image before it goes into the imagebuilder code and when it's in the imagebuilder code it gets executed.  So in the example, the ONBUILD touch /bar happens at the first image creation time rather than when the second image is created from it.

So I've been trying to figure out how to get the OnBuild value of the first image to be set and stored after it goes through the processing of imagebuilder.  I thought I could do so somewhere in image.go to do this and in the outdated comments you can see that I tried adding near line 160

```
dimage.Config.OnBuild = i.onBuild
```

So the value that I've captured in the onBuildCmd variable in imagebuildah/build.go has the value in it that we need.  If you can figure out where I can push that too, I think we'll be golden.  I've had a bear of a time working my way through all of the intertwined variable and structures representing the image in different stages/code.

rhatdan:
Closing this in favor of #722

PR #687, opened 2018-05-15
Attempt to download file from url, if fails assume Dockerfile

rhatdan:
If you do a
podman bud http://example.com/Dockerfile

podman currently assumes the file is an archive, if it is not an archive
we should assume it is a Dockerfile, download and create a Dockerfile in the
directory.


pixdrift:
Tested a build here and it worked cleanly and '-f <url> .' behaviour is also maintained.

With regard to documentation, I find **[context]** in this initial command description fairly ambiguous.
https://github.com/projectatomic/buildah/blob/master/docs/buildah-bud.md

With this change, the last parameter should be either a `PATH` or `URL` (previously was `PATH`), Do the square brackets denote optional by convention? I don't think bud will execute if `PATH` or `URL` aren't provided, so this is actually a mandatory option?

rhatdan:
Updated the man pages.

docs/buildah-bud.md
```diff
@@ -4,13 +4,19 @@
...
 ## SYNOPSIS
-**buildah** **bud | build-using-dockerfile** [*options* [...]] [**context**]
+**buildah** **bud | build-using-dockerfile** [*options* [...]] **context**
 
 ## DESCRIPTION
 Builds an image using instructions from one or more Dockerfiles and a specified
-build context directory.  The build context directory can be specified as the
-**http** or **https** URL of an archive which will be retrieved and extracted
-to a temporary location.
+build context directory.
```
TomSweeneyRedHat: Really nice additions, thanks for putting it together.

rhatdan:

PR #705, opened 2018-05-18
API: make PullPolicy and TerminalPolicy their own types

nalind:
Make PullPolicy and TerminalPolicy their own types, instead of just integer values, so that we can give them a `String()` method, and return more informative errors from a couple of `containerImageSource` methods when they're asked for particular manifest instances, which it doesn't support.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ LDFLAGS := -ldflags '-X main.gitCommit=${GIT_COMMIT} -X main.buildInfo=${BUILD_I
 
 all: buildah imgtype docs
 
-buildah: *.go imagebuildah/*.go cmd/buildah/*.go docker/*.go util/*.go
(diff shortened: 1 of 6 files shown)


rhatdan:
LGTM
@TomSweeneyRedHat @umohnani8 PTAL

umohnani8:
LGTM

TomSweeneyRedHat:
bot, retest this please

nalind:
Rebased, tweaked how we do the naming of the temporary directory again.

rhatdan:
@nalind Getting a little nuts there.
@rh-atomic-bot r+

nalind:
The main goal is to keep `.` followed by either a digit or `_` from appearing in any component of the pathname, because we're trying to finesse the pathname into part of an image name, and the rules around image names are a bit tighter.

PR #706, opened 2018-05-18
Make "run --terminal" an alias for "run --tty"

nalind:
I swear I can never remember which one it is.


rhatdan:
LGTM
But should we add -t also?

nalind:
Oh, I hadn't even noticed that.  Yeah.

rhatdan:
@rh-atomic-bot r+

PR #707, opened 2018-05-18
Fix buildah from docker.io/centos:7 pull

rhatdan:
Currently buildah does not parse the docker.io syntax correcly.
In the example above it should parse this to docker.io/library/centos:7

The change in this patch will check to see if the user specified a transport
if not, then it will append the default transport "docker://" and containers/image
takes care of the rest.

### DIFF
--- a/new.go
+++ b/new.go
@@ -133,6 +133,18 @@ func imageManifestAndConfig(ctx context.Context, ref types.ImageReference, syste
 	return nil, nil, nil
 }
 
+// Adds default transport if the image is not empty and does not have a transport
+func addTransport(image, transport string) string {
+	if image == "" {
+		return image
+	}
+
+	if _, err := alltransports.ParseImageName(image); err != nil {
+		return fmt.Sprintf("%s%s", transport, image)
+	}
+	return image
+}
+
 func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions) (*Builder, error) {
 	var ref types.ImageReference
 	var img *storage.Image
@@ -146,6 +158,8 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
 	if options.Transport == "" {
 		options.Transport = DefaultTransport
 	}
+
+	options.FromImage = addTransport(options.FromImage, options.Transport)
 
 	systemContext := getSystemContext(options.SystemContext, options.SignaturePolicyPath)
 


rhatdan:
Fixes https://github.com/projectatomic/buildah/issues/692

TomSweeneyRedHat:
LGTM assuming happy tests

rhatdan:
I am not sure this is the correct fix, especially where it is triggering so many other issues.  I will take a closer look on Monday.

mtrmac:
`util.ResolveName`, which uses `util.FromImage`, immediately below, accepts **either**:

1. _transport_`:`_reference_, **only** for _transport_ values `docker-archive`, `oci-archive`, `dir`, tarball`.  In that case _reference_ (i.e. a path) is returned.
2. Valid Docker reference syntax, qualified. Supposed to be returned unmodified.
3. Valid Docker reference syntax, unqualified. Supposed to be turned into a list with `localhost` and the search list and the like.
4. (Any other string, whether valid or arbitrary garbage, is returned unmodified.)

This PR puts all input into category 4, defeating the search list entirely.

---

The bug in this case is a confusion between cases 2 and 3: in
```go
if strings.HasPrefix(repoPath, defaultPrefix) && path.Join(domain, repoPath[len(defaultPrefix):]) == name {
```
the `path.Join()` [which is not the same thing and `strings.Join(…)`, BTW, it does `path.Clean` on the result] uses `repoPath`, which does not include any tag or digest. So the comparison ends up `"docker.io/centos" == name`, which fails when the input `name` has a tag or digest.

I guess something like
```go
domain := reference.Domain(named)
if middle := RegistryDefaultPathPrefix(domain); middle != "" {
   prefix := domain + "/" + middle + "/"
   if strings.HasPrefix(named.String(), prefix) && name == domain + "/" + named.String()[len(prefix):] {
    return []string{name}
  }
}
```
could work (minimally tested! This can and should have unit tests.)

---

The whole interaction of `newBuilder` et al. and `utils.ResolveName` is full of redundancies, computing information or deciding to interpret input in some way, and then throwing that away, BTW.  Look at how:
- The `minimumTruncatedIDLength` logic in `newBuilder` is redundant to what `utils.ResolveName` has already done
- `ResolveName` _already knows_ (or, well, has decided) whether the input is a _transport_`:`_reference_ or a Docker reference to search for (in which case the returned names already have a known transport interpretation, even if implied); it just discards that information for `newBuilder` (or, actually, _modifies the input and discards that information_ in case 1 above) to guess again.
- `pullImage` and `newBuilder` both contain copies of the same `options.Transport`/`DefaultTransport` heuristic.
- `localImageNameFromReference` (which can decompress and parse a multi-gigabyte archive) is called both in `newBuilder` and `pullImage`

and that’s just from a quick look on the surface.

(For the Nth time, *please*, decide what syntax is accepted in each string in the codebase, document that, and design a mechanism, maybe Golang typedefs, to avoid mixing them. *Stop adding random string edits in random places that make a single test case work.*)

rhatdan:
@nalind did a better PR for this https://github.com/projectatomic/buildah/pull/710
closing.

PR #708, opened 2018-05-18
Run(): manage containers more directly

nalind:
Instead of using the runtime's "run" command to have it handle the entire life cycle of a process when we need to launch one, do it in phases, so that we can handle passing the data between our stdio and the container's.  This will also make it possible for us to set up networking using CNI between the "create" and "start" phases.


rhatdan:
LGTM
@umohnani8 @mheon @TomSweeneyRedHat PTAL

rhatdan:
I would like @mheon to look at this one, since he works in this area more often.

run.go
```diff
@@ -452,31 +462,559 @@ func (b *Builder) Run(command []string, options RunOptions) error {
...
+		Options:       options,
+		Spec:          spec,
+		RootPath:      rootPath,
+		BundlePath:    bundlePath,
+		ContainerName: containerName,
+	})
+	if conferr != nil {
+		return errors.Wrapf(conferr, "error encoding configuration for %q", runUsingRuntimeCommand)
+	}
+	cmd := reexec.Command(runUsingRuntimeCommand)
```
mheon: @nalind Why are you using `reexec` here, instead of continuing on with the code in the same process? The code itself looks fine, just curious as to why you're doing this in a fresh main.

mheon: Nevermind, I see a comment below about setting subreaper - makes sense now.

rhatdan:
@rh-atomic-bot r+

Issue #692, opened 2018-05-15
`buildah from` doesn't like the format `docker.io/$NAME:$TAG`


carlwgeorge:
**Description**

Running `buildah from` with an image in the format of `docker.io/$NAME:$TAG` fails.

**Steps to reproduce the issue:**
1. `buildah from docker.io/centos` works
2. `buildah from docker.io/centos:7` does not work

**Describe the results you received:**
```
Error determining manifest MIME type for docker://registry.access.redhat.com/docker.io/centos:7: Error reading manifest 7 in registry.access.redhat.com/docker.io/centos: unknown: Not Found
```

**Describe the results you expected:**
```
centos-working-container
```

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.0-1.git1ab80bc.fc27.x86_64
```

**Output of `buildah version`:**

```
Version:       1.0
Go Version:    go1.9.5
Image Spec:    1.0.0
Runtime Spec:  1.0.0
Git Commit:    1ab80bc
Built:         Mon May  7 09:00:57 2018
OS/Arch:       linux/amd64
```

**Output of `cat /etc/*release`:**

```
Fedora release 27 (Twenty Seven)
NAME=Fedora
VERSION="27 (Twenty Seven)"
ID=fedora
VERSION_ID=27
PRETTY_NAME="Fedora 27 (Twenty Seven)"
ANSI_COLOR="0;34"
CPE_NAME="cpe:/o:fedoraproject:fedora:27"
HOME_URL="https://fedoraproject.org/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=27
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=27
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
Fedora release 27 (Twenty Seven)
Fedora release 27 (Twenty Seven)
```

**Output of `uname -a`:**

```
Linux tassadar 4.16.7-200.fc27.x86_64 #1 SMP Wed May 2 20:33:31 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

**Output of `cat /etc/containers/storage.conf`:**

```
[...]

PR #710, opened 2018-05-18
util.ResolveName(): handle completion for tagged/digested image names

nalind:
When checking if an image name includes a registry name, when checking for cases where parsing it inserts additional path components, handle cases where the provided name includes a tag or digest component.


nalind:
Fixes #692.

rhatdan:
Nice and it gets me out of the hell I was in with my PR.
@rh-atomic-bot r+

PR #725, opened 2018-05-26
add test to inspect

q384566678:

### DIFF
--- a/tests/inspect.bats
+++ b/tests/inspect.bats
@@ -2,6 +2,16 @@
 
 load helpers
 
+@test "inspect" {
+	cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+	run buildah commit --signature-policy ${TESTSDIR}/policy.json "$cid" alpine-image
+	[ "$status" -eq "0" ]
+	out1=$(buildah inspect --format '{{.OCIv1.Config}}' alpine)
+	out2=$(buildah inspect --type image --format '{{.OCIv1.Config}}' alpine-image)
+	[ "$out1" != "" ]
+	[ "$out1" = "$out2" ]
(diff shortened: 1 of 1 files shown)


tests/inspect.bats
```diff
@@ -2,6 +2,16 @@
 
 load helpers
 
+@test "inspect" {
+	cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+	run buildah commit --signature-policy ${TESTSDIR}/policy.json "$cid" alpine-image
+	[ "$status" -eq "0" ]
+	run buildah inspect alpine
```
rhatdan: Would it be possible to check some fields in these images to make sure they are the same.
Would probably make for  a better tests.

Check Cmd, Entrypoint, Env, Labels ...

q384566678: updated, WDYT?

q384566678:
updated.

TomSweeneyRedHat:
LGTM and happy green tests buttons.
If at some point you could add a 'buildah config' command to this or a new test to fill in some fields like author, entrypoint, cmd, etc and then test that they are filled with the values you expect, that's would be great.

rhatdan:
Awesome @q384566678 
Adding tests is something Engineers usually are reticent to do, but really help out in the long run.  I am as guilty as anyone of this.
@rh-atomic-bot r+

q384566678:
>If at some point you could add a 'buildah config' command to this or a new test to fill in some fields like author, entrypoint, cmd, etc and then test that they are filled with the values you expect, that's would be great.

I thought about it at first, but I think this kind of test has been implemented in config.bats. It feels repetitive when you reuse it. I don't know if my understanding is correct.

PR #727, opened 2018-05-28
images: Error if the specified imagename does not exist

q384566678:
The images option allows the image name to be specified. If the specified image does not exist, an error should be output.

Before modification:
```
➜  buildah git:(images-wron-name) sudo buildah images --format {{.ID}} 456
```

After modification:
```
➜  buildah git:(images-wron-name) sudo ./buildah images --format {{.ID}} 456
No such image 456

```


cmd/buildah/images.go
```diff
@@ -268,6 +274,13 @@ func outputImages(ctx context.Context, images []storage.Image, format string, st
 			outputUsingFormatString(truncate, digests, params)
 		}
 	}
+
+	if !found {
```
rhatdan: Nit you can combine these args if statements into a single if

if !found && argName != ""

rhatdan:
Thanks @q384566678 

You need to gofmt your code.

rhatdan:
Thanks @q384566678 

You need to gofmt your code.

cmd/buildah/images.go
q384566678: Thanks, updated, and I changed the related test file.

rhatdan:
bot, retest this please

PR #729, opened 2018-05-29
rm: add validation to args and all option

q384566678:
When arg is greater than one and the all option is used, error need to be output.

### DIFF
--- a/cmd/buildah/rm.go
+++ b/cmd/buildah/rm.go
@@ -35,6 +35,9 @@ func rmCmd(c *cli.Context) error {
 	if len(args) == 0 && !c.Bool("all") {
 		return errors.Errorf("container ID must be specified")
 	}
+	if len(args) > 0 && c.Bool("all") {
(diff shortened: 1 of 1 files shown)


rhatdan:
Why.  If a user specifies an image do delete and accidently adds a `-a`; then the tool would delete all containers, instead of giving an error.  Why would you want the command to continue and potentially do something the user did not intend?

q384566678:
>Why. If a user specifies an image do delete ...

It seems my idea is wrong. 
Updated. Added similar verification in the `rm` function.

TomSweeneyRedHat:
Change LGTM, it would be nice to have a test case.  Build issues are not related, hopefully they will be cleaned up with another round of testing.

rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot  retry

Issue #732, opened 2018-05-29
No way to provide 'exec form' to buildah config --entrypoint


pixdrift:
The Dockerfile ENTRYPOINT parameter provides two 'forms' of operation, that behave slightly differently. These forms are summarised in the Docker documentation here:
https://docs.docker.com/engine/reference/builder/#entrypoint

If an `array` is used for ENTRYPOINT, the command is interpreted as `exec` form
If a `string` is used for ENTRYPOINT, the command is interpreted as `shell` form

Unfortunately, the `shell` form doesn't take parameters in that are passed to the container.

The problem is, `buildah config --entrypoint` can only take a string parameter, so you can't currently configure a container with `buildah config` to use the `exec` (`array`) preferred form.

The differences in the behaviour can be seen in the following Dockerfile examples:

If `ENTRYPOINT` is declared as an array:
```
FROM centos
ADD showparams.sh /
ENTRYPOINT [ "/showparams.sh" ]
```

Results in:
```
            "Entrypoint": [
                "/showparams.sh"
            ],
```

If `ENTRYPOINT` is declared as a string:
```
FROM centos
ADD showparams.sh /
ENTRYPOINT /showparams.sh
```

Results in:
```
            "Entrypoint": [
                "/bin/sh",
                "-c",
                "/showparams.sh"
            ],
```

TomSweeneyRedHat:
Talking with @pixdrift via IRC:

```
tsweeney> any reason to not just do it the Docker style?  
<tsweeney> [ "/showparams.sh", "myinputdata" ]
<tsweeney> as a for instance.
<pixdrift> complex to parse on command line?
<pixdrift> although possibly not..
<pixdrift> maybe if it was hard quoted?
<pixdrift> --entrypoint '[ "/showparams.sh", "myinputdata" ]'
<tsweeney> yeah, something like that with doc.
<pixdrift> although.. you could check if the first character is '[' and parse accordingly
<tsweeney> yep, that too.
```

TomSweeneyRedHat:
@rhatdan and/or @nalind WDYT?

rhatdan:
Tom can you see if you can get this to work.

rhatdan:
@pixdrift Now that we have the PR Merged can we close this issue?

pixdrift:
Yep.

PR #736, opened 2018-05-30
Accept json array input for config entrypoint

pixdrift:
This is a first attempt at taking the agreed exec form of entrypoint for `buildah config --entrypoint` (#732) 

eg.
`buildah config --entrypoint '[ "/showparams.sh" ]' dea5f337c325`

Keen on feedback, as I am sure the `if` branching on the error is probably going to lose points with the judges 😄

### DIFF
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"strings"
 
 	"github.com/mattn/go-shellwords"
@@ -145,16 +146,26 @@ func updateConfig(builder *buildah.Builder, c *cli.Context) {
 		}
 	}
 	if c.IsSet("entrypoint") {
-		entrypointSpec := make([]string, 3)
-		entrypointSpec[0] = "/bin/sh"
-		entrypointSpec[1] = "-c"
-		entrypointSpec[2] = c.String("entrypoint")
 		if len(strings.TrimSpace(c.String("entrypoint"))) == 0 {
 			builder.SetEntrypoint(nil)
 		} else {
-			builder.SetEntrypoint(entrypointSpec)
+			var entrypointJSON []string
+			err := json.Unmarshal([]byte(c.String("entrypoint")), &entrypointJSON)
+
+			if err != nil {
+				// it wasn't a valid json array, fall back to string
+				entrypointSpec := make([]string, 3)
+				entrypointSpec[0] = "/bin/sh"
+				entrypointSpec[1] = "-c"
+				entrypointSpec[2] = c.String("entrypoint")
+
+				builder.SetEntrypoint(entrypointSpec)
+			} else {
+				builder.SetEntrypoint(entrypointJSON)
+			}
+
+			builder.SetCmd(nil)
 		}
-		builder.SetCmd(nil)
 	}
 	// cmd should always run after entrypoint; setting entrypoint clears cmd
 	if c.IsSet("cmd") {
--- a/tests/config.bats
+++ b/tests/config.bats
@@ -2,7 +2,112 @@
 
 load helpers
 
-@test "config entrypoint" {
+@test "config entrypoint using single element in JSON array (exec form)" {
+  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
+  buildah config --entrypoint '[ "/ENTRYPOINT" ]' $cid
+  buildah commit --format dockerv2 --signature-policy ${TESTSDIR}/policy.json $cid entry-image-docker
+  buildah commit --format ociv1 --signature-policy ${TESTSDIR}/policy.json $cid entry-image-oci
+
+  run buildah --debug=false inspect --type=image --format '{{.Docker.Config.Entrypoint}}' entry-image-docker
+  [ "$output" = "[/ENTRYPOINT]" ]
+  [ "$status" -eq 0 ]
+  
+  run buildah --debug=false inspect --type=image --format '{{.Docker.Config.Entrypoint}}' entry-image-oci
+  [ "$output" = "[/ENTRYPOINT]" ]
+  [ "$status" -eq 0 ]
+
+  run buildah --debug=false inspect --type=image --format '{{.OCIv1.Config.Entrypoint}}' entry-image-docker
+  [ "$output" = "[/ENTRYPOINT]" ]
+  [ "$status" -eq 0 ]
+
+  run buildah --debug=false inspect --type=image --format '{{.OCIv1.Config.Entrypoint}}' entry-image-oci
+  [ "$output" = "[/ENTRYPOINT]" ]
+  [ "$status" -eq 0 ]
+
+  buildah rm $cid
+  buildah rmi entry-image-docker entry-image-oci
+}
+
+@test "config entrypoint using multiple elements in JSON array (exec form)" {
+  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
+  buildah config --entrypoint '[ "/ENTRYPOINT", "ELEMENT2" ]' $cid
+  buildah commit --format dockerv2 --signature-policy ${TESTSDIR}/policy.json $cid entry-image-docker
+  buildah commit --format ociv1 --signature-policy ${TESTSDIR}/policy.json $cid entry-image-oci
+
+  run buildah --debug=false inspect --type=image --format '{{.Docker.Config.Entrypoint}}' entry-image-docker
+  [ "$output" = "[/ENTRYPOINT ELEMENT2]" ]
+  [ "$status" -eq 0 ]
+ 
+  run buildah --debug=false inspect --type=image --format '{{.Docker.Config.Entrypoint}}' entry-image-oci
+  [ "$output" = "[/ENTRYPOINT ELEMENT2]" ]
+  [ "$status" -eq 0 ]
+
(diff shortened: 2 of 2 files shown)


TomSweeneyRedHat:
@pixdrift you've some gofmt issues.  After you make your changes, do:

gofmt -s -w ./cmd/buildah/config.go

before doing your git add/commit/fetch/rebase/push commands.

cmd/buildah/config.go
```diff
@@ -145,14 +146,22 @@ func updateConfig(builder *buildah.Builder, c *cli.Context) {
...
+			entrypointSpec[1] = "-c"
+			entrypointSpec[2] = c.String("entrypoint")
+			if len(strings.TrimSpace(c.String("entrypoint"))) == 0 {
+				builder.SetEntrypoint(nil)
+			} else {
+				builder.SetEntrypoint(entrypointSpec)
+			}
 		} else {
-			builder.SetEntrypoint(entrypointSpec)
+			builder.SetEntrypoint(entrypointJson)
```
TomSweeneyRedHat: @pixdrift I think this is pretty darn close.  Like we chatted about in IRC, it doesn't handle 

```
buildah config --entrypoint ["/usr/bin/myscript.sh", "myinputdata"]
```

but it does handle this:

```
buildah config --entrypoint `["/usr/bin/myscript.sh", "myinputdata"]`
```
which @rhatdan and @nalind were good with, so leave that as is. 

The one thing that I think needs to be added is prepending "/bin/sh -c" to the entrypoint spec.  Maybe:
```
        if c.IsSet("entrypoint") {
                entrypointSpec:= []string{"/bin/sh", "-c"}

                var entrypointJson []string
                err := json.Unmarshal([]byte(c.String("entrypoint")), &entrypointJson)

                if err != nil {
                        // it wasn't a valid json array, fall back to string
                        entrypointSpec = append(entrypointSpec, c.String("entrypoint"))
                        if len(strings.TrimSpace(c.String("entrypoint"))) == 0 {
                                builder.SetEntrypoint(nil)
                        } else {
                                builder.SetEntrypoint(entrypointSpec)
                        }
                } else {
                        for _, node := range entrypointJson {
                                entrypointSpec = append(entrypointSpec, node)
                        }
                        builder.SetEntrypoint(entrypointSpec)
                }
                builder.SetCmd(nil)
        }
```

pixdrift: @TomSweeneyRedHat, '/bin/sh -c' is exactly what I am trying to avoid when specifying an ENTRYPOINT using this method. If you look at the ticket #732, you will see in my example that `/bin/sh -c` isn't appended when the Dockerfile ENTRYPOINT is specified as an array.

TomSweeneyRedHat: Really no /bin/sh -c in docker with an array?  How does it get run then?  Every time I've sent stuff to entrypoint and /bin/sh -c isn't in play it's fallen over.  I admittedly have kept it relatively simple, i think I've always just sent a string to Buildah or Docker.  Does Docker expect that the user will also specify the shell themselves if they send in an array?

pixdrift: No requirement to specify the shell if supplying an array. The test to prove the behaviour is easy to replicate, as per #732, just change how ENTRYPOINT is declared in the Dockerfile and `docker build`, `docker inspect`.

pixdrift:
CI failure is a minor nit on Json/JSON, I will fix it tonight and add some additional testing for the new use cases of `buildah config`.

TomSweeneyRedHat:
Addresses #732

pixdrift:
I think I broke the CI, won't pick up my job now 😞

TomSweeneyRedHat:
I've got to run, at home now, need to get to work, will try to kcik the bot to retest and will look again once I get into the office.

TomSweeneyRedHat:
bot, retest this please.

pixdrift:
@TomSweeneyRedHat is there a documentation convention for showing that `--entrypoint` will take the input in the form of array _or_ a string? Was going to update the documentation to mention the shell/exec forms for entrypoint, but can't find an example for the switch description.

pixdrift:
f28 rpmbuild blew up, f27 and el7 were fine, seems odd.

cmd/buildah/config.go
```diff
@@ -145,14 +146,22 @@ func updateConfig(builder *buildah.Builder, c *cli.Context) {
...
+		var entrypointJSON []string
+		err := json.Unmarshal([]byte(c.String("entrypoint")), &entrypointJSON)
+
+		if err != nil {
+			// it wasn't a valid json array, fall back to string
+			entrypointSpec := make([]string, 3)
+			entrypointSpec[0] = "/bin/sh"
+			entrypointSpec[1] = "-c"
+			entrypointSpec[2] = c.String("entrypoint")
+			if len(strings.TrimSpace(c.String("entrypoint"))) == 0 {
```
rhatdan: Can we do this check before we do the json.Unmarshal.

rhatdan:
@pixdrift Nice job on this, only want to rearrange a little bit.  I like the tests.

TomSweeneyRedHat:
@pixdrift the --entrypoint param is not terribly well documented in Buildah at this point.  I've not looked at the latest patch, but the best place to buff that up to start would be in docs/buildah-config.md

cmd/buildah/config.go
TomSweeneyRedHat: Thanks for the reference.  Seems weird that Docker adds /bin/sh -c for a string, but not for an array.

cmd/buildah/config.go
```diff
@@ -145,14 +146,22 @@ func updateConfig(builder *buildah.Builder, c *cli.Context) {
...
+				builder.SetEntrypoint(nil)
+			} else {
+				builder.SetEntrypoint(entrypointSpec)
+			}
 		} else {
-			builder.SetEntrypoint(entrypointSpec)
+			builder.SetEntrypoint(entrypointJSON)
```
TomSweeneyRedHat: Based on the example in the Docker doc that @pixdrift provided in an earlier reply (https://docs.docker.com/engine/reference/builder/#exec-form-entrypoint-example), should we do 'builder.SetCmd(nil)' if we have an array?  I think it still holds for the string case, but I'm not sure about the array case here.

TomSweeneyRedHat:
BTB, I'd remove the 'WIP' at this point from the commit title.

TomSweeneyRedHat:
Taking a first stab at the syntax for the command in the manpage

--entrypoint "string"  | '["array_entry1", "array_entry2", ...]'

and the I'd talk about it accepting either a string or an array in the verbiage directly below it, then include at least one example for each at the bottom of the page.

cmd/buildah/config.go
pixdrift: you are referring to the line 158, len check? the other array isn't used in valid JSON case.

cmd/buildah/config.go
pixdrift: I had difficulty determining what the blanking behaviour was replicating in Docker/bud?

rhatdan:
LGTM

rhatdan:
@umohnani8 @TomSweeneyRedHat PTAL

pixdrift:
@TomSweeneyRedHat, i'm not convinced blanking CMD is correct in either case, can you clarify this?

TomSweeneyRedHat:
I'll let @rhatdann correct me, but we had issues running at times because the value in the cmd was running along with the value from the entrypoint and some unexpected behaviour was occurring.  Thinking about it might be best to not do the nil and let the user set it as they'd like.  @rhatdan any other thoughts.

Changes LGTM, but we need some man page updates still.

rhatdan:
I thought we had looked at what docker did, and I think they blanked out the image cmd, if the user specified entrypoint.

Setup a container with a CMD and then modify it with entrypoint using docker, if they are both in the second image, then remove this behaviour.

pixdrift:
So the Dockerfile behaviour with buildah bud is to blank the CMD for both forms of ENTRYPOINT (if CMD exists in the FROM container).

rhatdan:
We want to follow that behaviour.

pixdrift:
How do we follow that behaviour?

Consider the following test case

buildah bud:
```
FROM centos
CMD "command"
ENTRYPOINT [ "/entrypoint" ]
```

buildah native:
```
ctr=$(buildah from centos)
buildah config --cmd command $ctr
buildah config --entrypoint /entrypoint $ctr
```

In this example, buildah bud leaves the CMD value intact, buildah native strips it. 

But if the CMD is specified in the FROM image (centos in this case), it would be stripped by the ENTRYPOINT command in the Dockerfile.

rhatdan:
I think we should merge this to fix the Entrypoint handling, and then work on the handling of the Cmd with Entrypoint in a separate PR.  Since it is broken in the current buildah code.

With respect to CMD it looks like it should be removed IFF it is in the original image but not if it was added to the container.

rhatdan:
So I think we should only do 
		builder.SetCmd(nil)
When Cmd == the images command, indicating the Cmd flag was never set in the container.  So

pixdrift:
@rhatdan, what about this scenario?

container_1:
```
FROM centos
CMD /bin/bash
```

container_2
```
FROM container_1
CMD /bin/bash
ENTRYPOINT /myscript.sh
```

In this instance, the CMD is set in the same container as the entrypoint, but to the same value as the parent container.. so it would match, and be wiped.. even though it was set in the second container.

rhatdan:
@rh-atomic-bot r+
Lets carry the conversation on in the other PR.

PR #700, opened 2018-05-17
Add support for controlling ID mapping and namespace usage

nalind:
This patch set wires in command line flags to select our strategy for using some of the namespace types when creating containers and running commands in them.
With user namespacing available, it properly sets up ID mapping for user namespaces when asked to do so for a container, and adopts the "intermediate bind mount" strategy from libpod to mount everything.
When creating a network namespace, it can now use CNI plugins to configure networking for the container's namespace.
It now handles passing data back and forth between the controlling terminal and a pseudo-terminal that it creates for a command being executed in the container, in order to have stdio still work when user namespaces with non-default mappings are being used, as the command would not normally be able to write to the controlling terminal.

PR #739, opened 2018-05-31
Run(): process RunOptions.Mounts, and its flags

nalind:
`RunOptions.Mounts` has been mistakenly ignored since #700; handle them.

Process the options on the bind mounts in `RunOptions.Mounts` the same way we handle the ones in `Builder.CommonBuildOpts.Volumes`, so that flags that control read-only/read-write usage, SELinux labeling, and mount propagation will work.

### DIFF
--- a/run.go
+++ b/run.go
@@ -285,8 +285,7 @@ func addCommonOptsToSpec(commonOpts *CommonBuildOptions, g *generate.Generator) 
 
 func (b *Builder) setupMounts(mountPoint string, spec *specs.Spec, optionMounts []specs.Mount, bindFiles map[string]string, builtinVolumes, volumeMounts []string, shmSize string, namespaceOptions NamespaceOptions) error {
 	// Start building a new list of mounts.
-	mounts := make([]specs.Mount, len(optionMounts))
-	copy(mounts, optionMounts)
+	var mounts []specs.Mount
 	haveMount := func(destination string) bool {
 		for _, mount := range mounts {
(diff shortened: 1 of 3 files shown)


rhatdan:
I still don't see why we want these?  It just seems to complicate things and adds very little value.

nalind:
Fixing the API breakage is more important, but if nothing else, having `buildah run` offer the option gives us a simple way to test that it stays fixed.

run.go
```diff
@@ -491,25 +483,46 @@ func runSetupVolumeMounts(mountLabel string, volumeMounts []string) ([]specs.Mou
...
 			Type:        "bind",
-			Source:      spliti[0],
+			Source:      host,
 			Options:     options,
-		})
+		}, nil
+	}
+	// Bind mount volumes specified for this particular Run() invocation
+	for _, i := range optionMounts {
+		mount, err := parseMount(i.Source, i.Destination, append(i.Options, "rbind"))
```
rhatdan: You swithched bind above to rbind below, is that intentional?

nalind: Yes, it brings the behavior for the CLI flag closer to volumes that were already set up using this code path.  I always thought it was weird that we had to pass that as an option from `runCmd()`.

rhatdan:
LGTM
PTAL @TomSweeneyRedHat @umohnani8

rhatdan:
@rh-atomic-bot r+

PR #740, opened 2018-05-31
Run(): only collect error output from stdio pipes if we created some

nalind:
If the runtime's "create" command fails, we try to collect error messages from the pipes that we hooked up to its stdio in anticipation of running a container without a TTY.  We should only bother with that when the container isn't attached to a TTY, which is the only time those pipe descriptor slices are populated, so that we don't panic when we try to read an item from an empty slice.


rhatdan:
LGTM

rhatdan:
@TomSweeneyRedHat @umohnani8 PTAL

umohnani8:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #741, opened 2018-06-01
 Print Warning message if cmd will not be used based on entrypoint 

rhatdan:

### DIFF
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -107,6 +107,34 @@ var (
 	}
 )
 
+func updateEntrypoint(builder *buildah.Builder, c *cli.Context) {
+	if len(strings.TrimSpace(c.String("entrypoint"))) == 0 {
+		builder.SetEntrypoint(nil)
+		return
+	}
+	var entrypointJSON []string
+	err := json.Unmarshal([]byte(c.String("entrypoint")), &entrypointJSON)
+
+	if err == nil {
+		builder.SetEntrypoint(entrypointJSON)
+		if len(builder.Cmd()) > 0 {
+			logrus.Warnf("cmd %q exists and will be passed to entrypoint as a parameter", strings.Join(builder.Cmd(), " "))
+		}
+		return
+	}
+
+	// it wasn't a valid json array, fall back to string
+	entrypointSpec := make([]string, 3)
+	entrypointSpec[0] = "/bin/sh"
+	entrypointSpec[1] = "-c"
+	entrypointSpec[2] = c.String("entrypoint")
+	if len(builder.Cmd()) > 0 {
+		logrus.Warnf("cmd %q exists but will be ignored because of entrypoint settings", strings.Join(builder.Cmd(), " "))
+	}
+
+	builder.SetEntrypoint(entrypointSpec)
+}
+
 func updateConfig(builder *buildah.Builder, c *cli.Context) {
 	if c.IsSet("author") {
 		builder.SetMaintainer(c.String("author"))
@@ -150,26 +178,7 @@ func updateConfig(builder *buildah.Builder, c *cli.Context) {
 		}
 	}
 	if c.IsSet("entrypoint") {
-		if len(strings.TrimSpace(c.String("entrypoint"))) == 0 {
-			builder.SetEntrypoint(nil)
-		} else {
-			var entrypointJSON []string
-			err := json.Unmarshal([]byte(c.String("entrypoint")), &entrypointJSON)
-
-			if err != nil {
-				// it wasn't a valid json array, fall back to string
-				entrypointSpec := make([]string, 3)
-				entrypointSpec[0] = "/bin/sh"
-				entrypointSpec[1] = "-c"
-				entrypointSpec[2] = c.String("entrypoint")
-
-				builder.SetEntrypoint(entrypointSpec)
-			} else {
-				builder.SetEntrypoint(entrypointJSON)
-			}
-
-			builder.SetCmd(nil)
-		}
+		updateEntrypoint(builder, c)
 	}
 	// cmd should always run after entrypoint; setting entrypoint clears cmd
 	if c.IsSet("cmd") {
--- a/docs/buildah-config.md
+++ b/docs/buildah-config.md
@@ -56,9 +56,13 @@ Note: this setting is not present in the OCIv1 image format, so it is discarded 
 **--entrypoint** *entry*
 
 Set the *entry point* for containers based on any images which will be built
-using the specified container.
+using the specified container. buildah supports two formats for entrypoint.  It
+can be specified as a simple string, or as a array of commands.
 
-Note: Setting the entrypoint without setting the --cmd, clears the cmd field in the container.
+Note: When the entrypoint is specified as a string, container runtimes will
+ignore the `cmd` value of the container image.  However if you use the array
+form, then the cmd will be appended onto the end of the entrypoint cmd and be
+executed together.
 
 **--env** *var=value*
 
--- a/vendor.conf
+++ b/vendor.conf
@@ -38,7 +38,7 @@ github.com/ostreedev/ostree-go aeb02c6b6aa2889db3ef62f7855650755befd460
 github.com/pborman/uuid master
 github.com/pkg/errors master
 github.com/pquerna/ffjson d49c2bc1aa135aad0c6f4fc2056623ec78f5d5ac
-github.com/projectatomic/libpod e686269da34ed4208f4ed517c0587ab38e8eaf2c
+github.com/projectatomic/libpod 781eec27b52c842fc83c8b1c97fbf825065f3b0c
 github.com/sirupsen/logrus master
 github.com/syndtr/gocapability master
 github.com/tchap/go-patricia master
--- a/cmd/buildah/main.go
+++ b/cmd/buildah/main.go
@@ -68,7 +68,7 @@ func main() {
 		},
 	}
 	app.Before = func(c *cli.Context) error {
-		logrus.SetLevel(logrus.ErrorLevel)
+		logrus.SetLevel(logrus.WarnLevel)
 		if c.GlobalBool("debug") {
 			debug = true
 			logrus.SetLevel(logrus.DebugLevel)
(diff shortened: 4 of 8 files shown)


giuseppe:
would it make sense to have this behaviour only with `buildah bud` (where we can more easilty check if CMD was already done earlier) for compatibility with Docker but let `buildah --entrypoint` override the ENTRYPOINT everytime as it does now?  I didn't know of this behaviour, but it looks quite confusing.

rhatdan:
Well it will override the behaviour now, the only concern is what happens to the CMD field.

pixdrift:
Let me make a series of Docker tests to confirm what it does. I am not 100% convinced wiping CMD is the right thing to do. The PR it was resolving had another issue with entrypoint.

pixdrift:
**entrypoint:src**
```
FROM centos
CMD /entrypoint.src
```

Produces:
```
            "Cmd": [
                "/bin/sh",
                "-c",
                "/command.src"
            ],

            "Entrypoint": null,
```

Building from the above container:

**entrypoint:samecmd** (child same CMD as parent, introduces ENTRYPOINT)
```
FROM entrypoint:src
CMD /command.src
ENTRYPOINT /entrypoint.new
```

```
            "Cmd": [
                "/bin/sh",
                "-c",
                "/command.src"
            ],
            "Entrypoint": [
                "/bin/sh",
                "-c",
                "/entrypoint.new"
            ],
```

**entrypoint:diffcmd** (child different CMD to parent, introduces ENTRYPOINT)
```
FROM entrypoint:src
CMD /command.new
ENTRYPOINT /entrypoint.new
```

```
            "Cmd": [
                "/bin/sh",
                "-c",
                "/command.new"
            ],
            "Entrypoint": [
                "/bin/sh",
                "-c",
                "/entrypoint.new"
            ],
```

**entrypoint:nocmd** (child no CMD, introduces ENTRYPOINT)
```
FROM entrypoint:src
ENTRYPOINT /entrypoint.new
```

```
            "Cmd": null,
            "Entrypoint": [
                "/bin/sh",
                "-c",
                "/entrypoint.new"
            ],
```

In summary, the only situation where the `CMD` is blanked by entrypoint, is when the parent container contains `CMD`, the child doesn't have a `CMD` specified and specifies an `ENTRYPOINT`

Not sure where I sit with `buildah --config` blanking `CMD` when a user can explicitly blank it themselves if that's what they want. It feels like a destructive default option. If that option stays, could there perhaps be a warning eg.

```
buildah config --entrypoint /entrypoint $cid
Note: CMD entry '/command' was found in the container and removed 
```

cmd/buildah/config.go
```diff
@@ -107,6 +107,63 @@ var (
...
+		} else {
+			builder.SetEntrypoint(entrypointJSON)
+		}
+
+		equal, err := compareImageCmd(c, builder.FromImage, builder.Cmd())
+		if err != nil {
+			logrus.Errorf("error parsing cmd from %q: %v", builder.FromImage, err)
+		}
+		if equal {
+			builder.SetCmd(nil)
```
TomSweeneyRedHat: Rather than setting the command to nil, could we just print out a warning with a recommendation to the user that they should set cmd to nil themselves?  That way we let them know of the potential issue, but don't disallow them from shooting themselves in the foot if they want to do so.

rhatdan: Sure, do you have a description of the risk?  Can you set the cmd to nil, or only to ""?  Is there a difference?

pixdrift: I think their foot might be alright...  

The Docker documentation has the following on Entrypoint:
> The shell form prevents any CMD or run command line arguments from being used

So if using shell form (a string) for ENTRYPOINT, it is going to ignore what's in CMD anyway, so they won't really shoot themselves in the foot.

If they use exec form (an array), existing CMD will be appended to the ENTRYPOINT on execution (which may cause them to shoot themselves in the foot if they have something like `/bin/bash` in CMD). But the way i've seen this used is ENTRYPOINT set in the parent, and then child just modifies CMD to adjust input parameters... but there may well be other use cases.

Example execution methods (punch line at the end!)

__exec form (uses CMD as parameter):__
```
FROM centos
ENTRYPOINT [ "/usr/bin/echo" ]
CMD [ "pixdrift" ]
```
Output
```
pixdrift
```

__shell form (ignores CMD value):__
```
FROM centos
ENTRYPOINT /usr/bin/echo
CMD [ "pixdrift" ]
```

_Produces no output_


So, if removal of CMD is taken out of `buildah config`.. error message could be

When using exec
> Warning: cmd 'pixdrift' exists and will be passed as parameter to entrypoint '/usr/bin/echo'

When using shell
> Warning: cmd 'pixdrift' exists but will be usurped by entrypoint

Probably don't need to leave 'usurped'.. late.. brain.. etc. 😄

TomSweeneyRedHat: I'm good with @pixdrift's suggestions, perhaps with slight tweaks for late brain.  If we go this route, the title for this commit should change.

rhatdan:
bot, retest this please

rhatdan:
@pixdrift @TomSweeneyRedHat PTAL

pixdrift:
@rhatdan, looks good, just want your thoughts on the additional message. Wanted to display that message because at that point, the behaviour of the CMD changes.

Really like the code re-org.. 👍

docs/buildah-config.md
```diff
@@ -56,9 +56,13 @@ Note: this setting is not present in the OCIv1 image format, so it is discarded
 **--entrypoint** *entry*
```
pixdrift: Good opportunity to update the --entrypoint syntax doco (provided by @TomSweeneyRedHat)

`--entrypoint "string"  | '["array_entry1", "array_entry2", ...]'`

cmd/buildah/config.go
```diff
@@ -107,6 +107,32 @@ var (
...
+func updateEntrypoint(builder *buildah.Builder, c *cli.Context) {
+	if len(strings.TrimSpace(c.String("entrypoint"))) == 0 {
+		builder.SetEntrypoint(nil)
+		return
+	}
+	var entrypointJSON []string
+	err := json.Unmarshal([]byte(c.String("entrypoint")), &entrypointJSON)
+
+	if err == nil {
+		builder.SetEntrypoint(entrypointJSON)
```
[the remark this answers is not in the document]
rhatdan: I don't know that is expected behaviour, as long as it is documented in the man page.  We could drop this down to logrus.Warningf then it will not give the user noice.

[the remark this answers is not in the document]
pixdrift: It appears to be the current behaviour of Docker, as shown here:
https://github.com/projectatomic/buildah/pull/741#discussion_r192418132

In this example, 'pixdrift' is being passed to echo if echo is declared in exec form (array).

From Docker documentation (although it's a little ambiguous):
https://docs.docker.com/engine/reference/builder/#entrypoint

> You can use the exec form of ENTRYPOINT to set fairly stable default commands and arguments and then use either form of CMD to set additional defaults that are more likely to be changed.
```
FROM ubuntu
ENTRYPOINT ["top", "-b"]
CMD ["-c"]
```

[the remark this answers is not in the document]
rhatdan: My punctuation was bad on my previous reply.  I know this is the correct and expected behaviour.  I Will drop it down a level so user is not flooded with the message.

rhatdan:
Need to wait for this to get merged 
https://github.com/projectatomic/libpod/pull/882

Before merging this commit. I also think we should wait for this until AFTER 1.1 is released.

cmd/buildah/config.go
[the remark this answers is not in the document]
pixdrift: np, suggestion sounds good 🍺

TomSweeneyRedHat:
Figured out the hold on projectatomic/libpod#882, it's due to the log level settings.

TomSweeneyRedHat:
Weird test errors, going to rekick the bot.

TomSweeneyRedHat:
bot, retest this please.

rhatdan:
@TomSweeneyRedHat @pixdrift @giuseppe PTAL

docs/buildah-config.md
TomSweeneyRedHat: Would like to see the above somewhere in the man page and/or an example of each

vendor.conf
```diff
@@ -38,7 +38,7 @@ github.com/ostreedev/ostree-go aeb02c6b6aa2889db3ef62f7855650755befd460
 github.com/pborman/uuid master
 github.com/pkg/errors master
 github.com/pquerna/ffjson d49c2bc1aa135aad0c6f4fc2056623ec78f5d5ac
-github.com/projectatomic/libpod e686269da34ed4208f4ed517c0587ab38e8eaf2c
+github.com/projectatomic/libpod 781eec27b52c842fc83c8b1c97fbf825065f3b0c
```
TomSweeneyRedHat: Gonna show my golang vendoer ignorance here.  Can we instead vendor in github.com/projectatomic/libpod/pkg?  That seems to be the only code we use from libpod and I just wonder if we could set up some kind of circular nastiness if we vendor in all of libpod as it vendors in Buildah.

docs/buildah-config.md
```diff
@@ -56,9 +56,13 @@ Note: this setting is not present in the OCIv1 image format, so it is discarded
...
 
 Set the *entry point* for containers based on any images which will be built
-using the specified container.
+using the specified container. buildah suppots two formats for entrypoint.  It
```
giuseppe: typo here: suppots

vendor.conf
rhatdan: vndr is smart enough to only pull in the parts that it needs. So unless something in libpod imports certain code from buildah, it will not end up in the vendor.

rhatdan:
@cevich @jlebon Is there something wacky with buildah travis?

jlebon:
Looks like it's working now. I don't have much experience with Travis, so your guess would've been as good as mine. :)

rhatdan:

cevich:
Looks like ``homu`` is wedged?  I'm not sure what that is, what it's for, or who maintains it :(

rhatdan:
Yes travis seems to have come back to life today.  Was in a bad state yesterday.

PR #745, opened 2018-06-01
use absolute path for bundlePath

baude:
When running buildah code on Atomic Hosts, we need to make sure
the absolute path for the bundlePath is used or operations
will fail.

Signed-off-by: baude <bbaude@redhat.com>

PR #746, opened 2018-06-01
run.bats: check that we can run with symlinks in the bundle path

nalind:
Make sure that we don't trigger error messages in `runc` when `$TMPDIR`, which affects `os.TempDir()` (which is the default for `ioutil.TempDir()`), is itself a symbolic link.  Tests #745.


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM.  @mheon will probably want to revendor Podman after this goes in.

rhatdan:
@rh-atomic-bot r+

PR #747, opened 2018-06-02
Add registry errors for pull

TomSweeneyRedHat:

Buffs up the registry errors on pull.  Based on https://github.com/projectatomic/libpod/pull/84 by @baude.

### DIFF
--- a/new.go
+++ b/new.go
@@ -168,7 +168,7 @@ func resolveImage(ctx context.Context, systemContext *types.SystemContext, store
 		if options.PullPolicy == PullAlways {
 			pulledImg, pulledReference, err := pullAndFindImage(ctx, store, image, options, systemContext)
 			if err != nil {
-				logrus.Debugf("error pulling and reading image %q: %v", image, err)
+				logrus.Debugf("unable to pull and read image %q: %v", image, err)
 				continue
 			}
 			ref = pulledReference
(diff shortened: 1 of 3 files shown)


pull.go
```diff
@@ -171,7 +173,21 @@ func pullImage(ctx context.Context, store storage.Store, imageName string, optio
 	if err == nil {
 		return destRef, nil
 	}
-	return nil, err
+
+	// If no image was found, we should handle.  Lets be nicer to the user and see if we can figure out why.
+	registryPath := sysregistries.RegistriesConfPath(&types.SystemContext{})
+	searchRegistries, err := registries.GetRegistries()
+	if err != nil {
+		return nil, err
```
rhatdan: Does this fail if /etc/containers/registies.conf does not exist?

TomSweeneyRedHat: Yes, it throws:

```
unable to parse the registries.conf file: open /etc/containers/registries.conf: no such file or directory
```

rhatdan:
LGTM
@rh-atomic-bot r+

new.go
```diff
@@ -168,7 +168,7 @@ func resolveImage(ctx context.Context, systemContext *types.SystemContext, store
 		if options.PullPolicy == PullAlways {
 			pulledImg, pulledReference, err := pullAndFindImage(ctx, store, image, options, systemContext)
 			if err != nil {
-				logrus.Debugf("error pulling and reading image %q: %v", image, err)
+				logrus.Debugf("unable to pull and read image %q: %v", image, err)
```
TomSweeneyRedHat: @rhatdan, just wanted to note that the errors that are now thrown in the new code in pull.go and util.go below are caught here and thrown out to debug only, not shown to the user directly.  I tried sending it out to the user directly, but that error gets generated for each registry in the list in registries.conf until it is resolved and/or fails looking for the image on localhost.  I thought in the case where it failed looking in a registry and then worked in a second was very misleading so I left the debug here rather than showing the error all the time.

rhatdan:
Homu seems to be hosed, manually merging.

PR #749, opened 2018-06-04
Add parameter judgment for commands that do not require parameters

q384566678:

### DIFF
--- a/cmd/buildah/containers.go
+++ b/cmd/buildah/containers.go
@@ -90,6 +90,9 @@ var (
 )
 
 func containersCmd(c *cli.Context) error {
+	if len(c.Args()) > 0 {
+		return errors.New("'buildah containers' does not accept arguments")
+	}
 	if err := parse.ValidateFlags(c, containersFlags); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


cmd/buildah/containers.go
```diff
@@ -90,6 +90,9 @@ var (
 )
 
 func containersCmd(c *cli.Context) error {
+	if len(c.Args()) > 0 {
+		return errors.New("'buildah containers' accepts no arguments")
```
TomSweeneyRedHat: I think I'd prefer "does not accept arguments" over "accepts no arguments" for both of these errors.

TomSweeneyRedHat: I'm sitting on the fence about the need for this change or just changing it to a warning message and then just continue on.  @rhatdan WDYT?

rhatdan: I prefer Tom's message.  I am fine with this error-ing out.

q384566678: updated, PTAL.

q384566678:
>Code LGTM but I am missing tests

Do we still need to perform a special test on this? Each command has a judgment on the number of input parameters. In this case, the parameters need to be tested for each command. I don't think it is necessary.

TomSweeneyRedHat:
LGTM.  @ripculd I think the tests have been provided in another PR.  Are you OK with this PR?

rhatdan:
@rh-atomic-bot r+

PR #750, opened 2018-06-04
add mount test

q384566678:

### DIFF
--- a/tests/mount.bats
+++ b/tests/mount.bats
@@ -1,0 +1,33 @@
+#!/usr/bin/env bats
+
+load helpers
+
+@test "mount one container" {
+  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+  run buildah --debug=false mount "$cid"
+  [ "${status}" -eq 0 ]
+  buildah rm $cid
+  buildah rmi -f alpine
+}
+
+@test "mount bad container" {
+  run buildah --debug=false mount badcontainer 
+  [ "${status}" -ne 0 ]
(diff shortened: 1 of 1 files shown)


tests/mount.bats
```diff
@@ -0,0 +1,29 @@
...
+
+@test "list currently mounted containers" {
+  cid1=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+  buildah mount "$cid1"
+  cid2=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+  buildah mount "$cid2"
+  cid3=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+  buildah mount "$cid3"
+  run buildah mount
+  [ "${status}" -eq 0 ]
```
TomSweeneyRedHat: It would be good to add a line like:
[[ ${lines[2]} =~ "/var/lib/containers" ]]
To make sure there are at least 3 lines with mounts listed.  I'd really like to verify that 3 lines were returned, but I'm not seeing how to do that offhand.

q384566678: updated, PTAL

rhatdan:
I think you should `echo $output` after each RUN command
Tests are not happy.

q384566678:
Thanks, updated. It's strange why the mount point is in `/tmp`.

rhatdan:
@q384566678 The test sets up all containers and storage to run out of /tmp.

@rh-atomic-bot r+

PR #516, opened 2018-03-14
Add CodeCov to travis

TomSweeneyRedHat:
Signed-off-by: TomSweeneyRedHat <tsweeney@redhat.com>

Adding CodeCov to generate Code Coverage numbers to Buildah.

PR #751, opened 2018-06-05
rmi, rm: add test

q384566678:

### DIFF
--- a/tests/rmi.bats
+++ b/tests/rmi.bats
@@ -27,6 +27,15 @@ load helpers
   [ "$output" == "" ]
 }
 
+@test "remove multiple non-existent images errors" {
+  run buildah --debug=false rmi image1 image2 image3
+  [ "${lines[0]}" == "could not get image \"image1\": identifier is not an image" ]
+  [ "${lines[1]}" == "could not get image \"image2\": identifier is not an image" ]
+  [ "${lines[2]}" == "could not get image \"image3\": identifier is not an image" ]
+  [ $(wc -l <<< "$output") -eq 3 ]
+  [ "${status}" -eq 1 ]
+}
+
 @test "remove all images" {
   cid1=$(buildah from --signature-policy ${TESTSDIR}/policy.json scratch)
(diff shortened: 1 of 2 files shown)


ripcurld:
@rhatdan @TomSweeneyRedHat 
Although `bats` makes it easy to write tests, I prefer writing unit-tests in Go because we can have code coverage and it motivates you to write unit-tests which give confidence to change things in the future.

I know we have unit-tests but looking at the code base it looks like we have more `bats` tests than unit-tests, doesn't it?

pixdrift:
I've seen ways to get coverage for external testing tools for go, it would be awesome to get coverage for bats tests.

tests/rmi.bats
```diff
@@ -27,6 +27,15 @@ load helpers
   [ "$output" == "" ]
 }
 
+@test "remove multiple images errors" {
```
TomSweeneyRedHat: suggest changing "images" to "non-existent images"

TomSweeneyRedHat:
@q384566678 thanks for the PR.  I've one very small suggestion request for a name in one of your tests.  If you could do that and push again, I think the Travis CI test issue will clear itself up as it looks to be a hiccup/flake.

As far as the type of test go, @nalind started the test suite using bats and it grew over time to what it is now.  Since then other folks have added *.go tests here and there.  Longer term I'd like to convert the tests to ginkgo based bdd tests.  They're written in go and should allow for coverage.  That and getting initial code coverage going with CodeCov has been a "background" project of mine that I keep making very slow progress on.  #516 is my current attempt for Buildah.  My main issue there is getting Buildah to run inside of the Travis CI Ubuntu so the ginkgo tests can run.  Podman does most of their testing using gingko having converted most of if not all of their bats, but they use a different test environment spinning up a container to do it within so they've not run into the Ubuntu fun there.  

So in short, we're trying to keep stuff in bats for now for continuity and once we get ginkgo setup, then we can start converting from bats and not have to search all over for the files.

tests/rmi.bats
q384566678: updated， PTAL.

PR #753, opened 2018-06-05
imagebuildah/build.go: write to reportwriter

baude:
when calling build through backends like varlink, writing to stdout or
stderr is not really option.  therefore, we should always write to the
provided reportwriter unless otherwise overriden.  this allows the build output
to be returned through the backend API and prevents it from landing in
journald.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -532,10 +532,10 @@ func NewExecutor(store storage.Store, options BuildOptions) (*Executor, error) {
 		annotations:           append([]string{}, options.Annotations...),
 	}
 	if exec.err == nil {
-		exec.err = os.Stderr
+		exec.err = options.ReportWriter
 	}
 	if exec.out == nil {
-		exec.out = os.Stdout
+		exec.out = options.ReportWriter
 	}
 	if exec.log == nil {
 		stepCounter := 0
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
Changes LGTM, but I want to run a few sanity tests.

rhatdan:
@nalind @umohnani8 PTAL

rhatdan:
LGTM

TomSweeneyRedHat:
@baude unhappy tests.
I'm heading to the home office, will pick up testing after I arrive.

not ok 139 run-user
# (in test file ./run.bats, line 266)
#   `[ "$output" = 0 ]' failed
# /root/gopath/src/github.com/projectatomic/buildah/runc
# CGO_ENABLED=1

imagebuildah/build.go
```diff
@@ -532,10 +532,10 @@ func NewExecutor(store storage.Store, options BuildOptions) (*Executor, error) {
 		annotations:           append([]string{}, options.Annotations...),
 	}
 	if exec.err == nil {
-		exec.err = os.Stderr
+		exec.err = options.ReportWriter
```
nalind: This will cause a crash in quiet mode, because we previously didn't (and still don't) require `ReportWriter` to be non-`nil`, but depend on `exec.err` being non-`nil` below.

rhatdan:
@baude Could you look at the latest changes from @nalind patch to see if this fixes your issue?

rhatdan:
@baude I think @nalind fixed your issue, closing. Reopen if I am wrong.

PR #754, opened 2018-06-05
Provide redirection for stdio

nalind:
Provide `RunOption` fields for callers to give us stdio as an `io.ReadCloser` and a pair of `io.WriteClosers`, or `nil` to use `os.Stdin`/`os.Stdout`/`os.Stderr`.


rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #756, opened 2018-06-06
runCopyStdio(): don't close stdin unless we saw POLLHUP

nalind:
When polling for input on stdin (in Terminal == false cases), don't close the write end of the pipe that we're using to relay data from stdin to the container unless poll() tells us that we got a POLLHUP.

Assuming that was the case when POLLIN wasn't set meant that we'd close it as soon as poll() returned and there was no activity on the descriptor, and if poll() only returned because we had output from the container to relay back, we were doing so prematurely.


rhatdan:
LGTM
@rh-atomic-bot r+

PR #759, opened 2018-06-06
Shouldn't add insecure registries to list of search registries

rhatdan:
The search registries and insecure registries are separate lists,
and should not be merged just because the InsecureSkipTLSVerify is
set

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -112,11 +112,6 @@ func ResolveName(name string, firstRegistry string, sc *types.SystemContext, sto
 		logrus.Debugf("unable to read configured registries to complete %q: %v", name, err)
 		registries = []string{}
 	}
-	if sc.DockerInsecureSkipTLSVerify {
-		if unverifiedRegistries, err := sysregistries.GetInsecureRegistries(sc); err == nil {
-			registries = append(registries, unverifiedRegistries...)
-		}
-	}
 
 	// Create all of the combinations.  Some registries need an additional component added, so
 	// use our lookaside map to keep track of them.  If there are no configured registries, we'll


rhatdan:
@nalind @TomSweeneyRedHat @mtrmac @umohnani8 PTAL

mtrmac:
> ACK as far as

Please ignore, wrong tab.

rhatdan:
@rh-atomic-bot r=mtrmac

util/util.go
```diff
@@ -112,11 +112,6 @@ func ResolveName(name string, firstRegistry string, sc *types.SystemContext, sto
 		logrus.Debugf("unable to read configured registries to complete %q: %v", name, err)
 		registries = []string{}
 	}
```
TomSweeneyRedHat: When do we reference the insecure registries list?  Rather than appending the insecure registries, should we just use only them when DockerInsecureSkipTLSVerify is set?  I.e. just do 
registries = unverifiedRegistries 
at line 117

rhatdan: If you want to search a registry then you add it to the search able list, if you want to treat a registry as insecure, then you put it in the insecure list, If you want them in both then you put them in both. 

@vrotherberg is working on a redesigned config, but this is what we have now.

rhatdan: @TomSweeneyRedHat I believe we have another bug on mishandling of insecure registiries, which might be what you are looking at.

TomSweeneyRedHat:
That's what I was poking at.  I might have missed it, but I only see us reading the secured list from that file.  My thinking, maybe wrong, is we should read the secured list if tls-verify = true, the insecured list if tls-verify=false.

mtrmac:
There _is no_ “secured list”; all registries are secure by default, unless they are in an insecure list.

The search list is, AFAIK, completely orthogonal to the secure/insecure statuses of registries

TomSweeneyRedHat:
@mtrmac you're right, that is listed as "search" list in the registries.conf.  In my noggin I'd it as the secured list.  Regardless, I don't think we're ever reading the registries.insecure list in Buildah.  Do you have thoughts on when we should be including that list in our search of registires if ever?

mtrmac:
Yeah, buildah might need to do something like https://github.com/projectatomic/libpod/blob/0eda60957d593411a144371bb4903c7a74307a59/libpod/image/pull.go#L194 .

Though, ideally (see the conversation around `sysregistriesv2`), containers/image should probably read the configuration automatically and use it for everyone, so that the code in libpod does not have to be copied all over the place.

PR #767, opened 2018-06-08
Add --layers and --no-cache to buildah bud

umohnani8:
This patch adds in the caching feature to buildah bud.

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -70,6 +70,10 @@ func budCmd(c *cli.Context) error {
 	} else {
 		return errors.Errorf("unrecognized image type %q", format)
 	}
+	layers := buildahcli.UseLayers()
+	if c.IsSet("layers") {
+		layers = c.Bool("layers")
+	}
 	contextDir := ""
 	cliArgs := c.Args()
 	if len(cliArgs) > 0 {
@@ -160,10 +164,6 @@ func budCmd(c *cli.Context) error {
 		logrus.Debugf("build caching not enabled so --force-rm flag has no effect")
 	}
 
-	if c.IsSet("no-cache") {
-		logrus.Debugf("build caching not enabled so --no-cache flag has no effect")
-	}
-
 	if c.IsSet("rm") {
 		logrus.Debugf("build caching not enabled so --rm flag has no effect")
 	}
@@ -205,6 +205,8 @@ func budCmd(c *cli.Context) error {
 		Squash:                c.Bool("squash"),
 		Labels:                c.StringSlice("label"),
 		Annotations:           c.StringSlice("annotation"),
+		Layers:                layers,
+		NoCache:               c.Bool("no-cache"),
 	}
 
 	if c.Bool("quiet") {
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	is "github.com/containers/image/storage"
 	"github.com/containers/image/transports"
@@ -20,6 +21,7 @@ import (
 	"github.com/containers/storage/pkg/stringid"
 	"github.com/docker/docker/builder/dockerfile/parser"
 	docker "github.com/fsouza/go-dockerclient"
+	"github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/openshift/imagebuilder"
 	"github.com/pkg/errors"
@@ -141,6 +143,11 @@ type BuildOptions struct {
 	Annotations []string
 	// OnBuild commands to be run by images based on this image
 	OnBuild []string
+	// Layers tells the builder to create a cache of images for each step in the Dockerfile
+	Layers bool
+	// NoCache tells the builder to build the image from scratch without checking for a cache.
+	// It creates a new set of cached images for the build.
+	NoCache bool
 }
 
 // Executor is a buildah-based implementation of the imagebuilder.Executor
@@ -187,6 +194,9 @@ type Executor struct {
 	labels                         []string
 	annotations                    []string
 	onbuild                        []string
+	layers                         bool
+	topLayers                      []string
+	noCache                        bool
 }
 
 // withName creates a new child executor that will be used whenever a COPY statement uses --from=NAME.
@@ -539,6 +549,8 @@ func NewExecutor(store storage.Store, options BuildOptions) (*Executor, error) {
 		squash:                options.Squash,
 		labels:                append([]string{}, options.Labels...),
 		annotations:           append([]string{}, options.Annotations...),
+		layers:                options.Layers,
+		noCache:               options.NoCache,
 	}
 	if exec.err == nil {
 		exec.err = os.Stderr
@@ -646,6 +658,9 @@ func (b *Executor) Prepare(ctx context.Context, ib *imagebuilder.Builder, node *
 	}
 	b.mountPoint = mountPoint
 	b.builder = builder
+	// Add the top  layer of this image to b.topLayers so we can keep track of them
+	// when building with cached images.
+	b.topLayers = append(b.topLayers, builder.TopLayer)
 	return nil
 }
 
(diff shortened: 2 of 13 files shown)


umohnani8:
Working on tests, completions and docs.
@rhatdan @TomSweeneyRedHat  can you please pull these changes in and play around with it a bit. I have tried to hit every case I could find, but it is possible i missed/messed up something.

TomSweeneyRedHat:
@umohnani8 looks like test unhappiness.  I'm pulling now to test fwiw.

TomSweeneyRedHat:
Looks like `buildah from scratch` isn't happy with his patch.  The other tests that don't use scratch in the baseline tests passed with this patch though.

```
# buildah from scratch
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0xa136db]

goroutine 1 [running]:
github.com/projectatomic/buildah.newBuilder(0xcbaa40, 0xc4200260b0, 0xcc9080, 0xc420486c60, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/tmp/tsweeney/workspaces/buildah/src/github.com/projectatomic/buildah/new.go:338 +0xd4b
github.com/projectatomic/buildah.NewBuilder(0xcbaa40, 0xc4200260b0, 0xcc9080, 0xc420486c60, 0x7ffdc8ac05d3, 0x7, 0x0, 0x0, 0x0, 0x0, ...)
	/tmp/tsweeney/workspaces/buildah/src/github.com/projectatomic/buildah/buildah.go:365 +0x97
main.fromCmd(0xc4200c54a0, 0x0, 0xf3)
	/tmp/tsweeney/workspaces/buildah/src/github.com/projectatomic/buildah/cmd/buildah/from.go:224 +0x654
github.com/projectatomic/buildah/vendor/github.com/urfave/cli.HandleAction(0xb12c60, 0xc51f78, 0xc4200c54a0, 0x0, 0xc42036de00)
	/tmp/tsweeney/workspaces/buildah/src/github.com/projectatomic/buildah/vendor/github.com/urfave/cli/app.go:501 +0xc8
github.com/projectatomic/buildah/vendor/github.com/urfave/cli.Command.Run(0xc17a44, 0x4, 0x0, 0x0, 0x0, 0x0, 0x0, 0xc389ef, 0x2c, 0x0, ...)
	/tmp/tsweeney/workspaces/buildah/src/github.com/projectatomic/buildah/vendor/github.com/urfave/cli/command.go:165 +0x47d
github.com/projectatomic/buildah/vendor/github.com/urfave/cli.(*App).Run(0xc420020700, 0xc42001e090, 0x3, 0x3, 0x0, 0x0)
	/tmp/tsweeney/workspaces/buildah/src/github.com/projectatomic/buildah/vendor/github.com/urfave/cli/app.go:259 +0x6e8
main.main()
	/tmp/tsweeney/workspaces/buildah/src/github.com/projectatomic/buildah/cmd/buildah/main.go:107 +0xbff

```

imagebuildah/build.go
```diff
@@ -675,42 +693,304 @@ func (b *Executor) Execute(ib *imagebuilder.Builder, node *parser.Node) error {
...
+			}
+			continue
+		}
+
+		var (
+			cacheID string
+			err     error
+			imgID   string
+		)
+		// checkForCache will be true if b.cache is true and a cached intermediate image is found.
```
rhatdan: b.layers?
Should checkForCache now be useLayers?

imagebuildah/build.go
```diff
@@ -675,42 +693,304 @@ func (b *Executor) Execute(ib *imagebuilder.Builder, node *parser.Node) error {
...
+			files, err := ioutil.ReadDir(b.contextDir)
+			if err != nil {
+				return nil, errors.Wrapf(err, "error reading context directory %q", b.contextDir)
+			}
+			for _, file := range files {
+				if strings.HasPrefix(file.Name(), currNode.Value[:len(currNode.Value)-2]) {
+					src = append(src, filepath.Join(b.contextDir, file.Name()))
+				}
+			}
+		} else if strings.Contains(currNode.Value, "?") {
```
rhatdan: You should use regex for these checks.  I think the code would be much cleaner.

imagebuildah/build.go
```diff
@@ -675,42 +693,304 @@ func (b *Executor) Execute(ib *imagebuilder.Builder, node *parser.Node) error {
...
+	if fi.ModTime().After(*historyTime) {
+		return true, nil
+	}
+	if fi.Mode().IsDir() {
+		files, err := ioutil.ReadDir(item)
+		if err != nil {
+			return false, errors.Wrapf(err, "error retrieving contents of directory %q", item)
+		}
+		for _, file := range files {
+			return timeStampAfter(filepath.Join(item, file.Name()), historyTime)
```
rhatdan: Not sure if using recursion is a good idea, should we just walk the directory?  Ask @nalind

pkg/cli/common.go
```diff
@@ -119,9 +119,13 @@ var (
 			Name:  "label",
 			Usage: "Set metadata for an image (default [])",
 		},
+		cli.BoolFlag{
+			Name:  "layers",
+			Usage: "cache intermediate layers during build.",
```
rhatdan: Could you add an environment variable BUILDAH_LAYERS and then have bool function useLayers() that checks if this is set to true/false (Default to false if not set)  Then set the default value here
Value: useLayers()

rhatdan:
Missing man pages and command completion.

TomSweeneyRedHat:
missing tests too

imagebuildah/build.go
umohnani8: Will it make much of a difference? Won't I have to have two different regex expressions to check against for the two different conditions I am checking for?

imagebuildah/build.go
umohnani8: Changed to `checkForLayers`.

imagebuildah/build.go
umohnani8: Realised that filpath has a Walk functions, so using that instead to walk to file tree and compare the time stamps. filpath.Walk walks the file tree recursively I believe though.

pkg/cli/common.go
umohnani8: Added the env variable.

cmd/buildah/common.go
```diff
@@ -133,3 +133,11 @@ func defaultFormat() string {
 	}
 	return "oci"
 }
+
+func useLayers() bool {
+	layers := os.Getenv("BUILDAH_LAYERS")
+	if layers == "true" {
```
rhatdan: Add 
strings.ToLower(layers) and check also against "1"

cmd/buildah/bud.go
```diff
@@ -70,6 +70,10 @@ func budCmd(c *cli.Context) error {
 	} else {
 		return errors.Errorf("unrecognized image type %q", format)
 	}
+	layers := useLayers()
```
rhatdan: Set this in default during option definition, then always just do.

layers = c.Bool("layers")

pkg/cli/common.go
rhatdan: You need to add
```
Value: useLayers()
```
Here.

Then if I do buildah bud --help
It will show me the default as true or false, depending on if the environment is set.

imagebuildah/build.go
rhatdan: Actually the best way to handle this is with filepath.Glob
https://golang.org/pkg/path/filepath/#Glob

rhatdan:
Still missing man pages and completions.

cmd/buildah/bud.go
umohnani8: BoolFlag doesn't have a `Value` field. I can set the EnvVar field, which shows which environment variable controls the layers flag.

rhatdan: Can you change the description then to add a default "%s", useLayers

cmd/buildah/common.go
umohnani8: Fixed

umohnani8:
Using filepath.Glob instead for pattern matching. Add docs, completions, and tests.

umohnani8:
Fixes https://github.com/projectatomic/buildah/issues/601

umohnani8:
Tests are happy. @rhatdan @TomSweeneyRedHat @nalind PTAL

rhatdan:
Awesome job @umohnani8 
LGTM

rhatdan:
@pixdrift PTAL

rhatdan:
@rh-atomic-bot r+

TomSweeneyRedHat:
I'd a test machine hiccup with Docker, but after I fixed that, the baseline tests passed.

zvonkok:
Wouldn't caching be also useful for "normal" builds, when not using a Dockerfile? 
I am heavily using bash scripts for building my containers. 

/cc @jeremyeder

rhatdan:
@zvonkok Open an issue for this and we can talk about it.  
You can do buildah commit, whenever you want and then base the next image off of the content.  So you have control over this in your scripts.

PR #768, opened 2018-06-08
Switch buildah back to ErrorLevel

rhatdan:

### DIFF
--- a/cmd/buildah/main.go
+++ b/cmd/buildah/main.go
@@ -68,7 +68,7 @@ func main() {
 		},
 	}
 	app.Before = func(c *cli.Context) error {
-		logrus.SetLevel(logrus.WarnLevel)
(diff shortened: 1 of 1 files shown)


rhatdan:
@rh-atomic-bot r+

pixdrift:
Is there any context around why this was reverted? Did something break?

TomSweeneyRedHat:
One of the new changes listed a lot of crufty warning messages.  I don't recall the exact wording but something like:

WARNING variable is an int and being cast to a double.

type of thing many times over in the output of a buildah command.

pixdrift:
ack and cheers @TomSweeneyRedHat

rhatdan:
Yes we need to do a deeper dive to make sure we drop Warning messages down to Info, that are not that important.

PR #769, opened 2018-06-08
Bump to Version 1.1

rhatdan:


release.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/sh
+#
+# Cut a libpod release.  Usage:
```
TomSweeneyRedHat: buildah not libpod

release.sh
```diff
@@ -0,0 +1,73 @@
...
+	sed -i "s/^\(Version: *\).*/\1${LOCAL_VERSION}/" contrib/rpm/buildah.spec
+}
+
+write_makefile_epoch()
+{
+	LOCAL_EPOCH="$1"
+	sed -i "s/^\(EPOCH_TEST_COMMIT ?= \).*/\1${LOCAL_EPOCH}/" Makefile
+}
+
+write_changelog()
```
TomSweeneyRedHat: neat!  Much easier than the gronking I've done.

buildah.go
```diff
@@ -23,7 +23,7 @@ const (
 	Package = "buildah"
 	// Version for the Package.  Bump version in contrib/rpm/buildah.spec
 	// too.
-	Version = "1.0"
+	Version = "1.2-dev"
```
TomSweeneyRedHat: should this be 1.1 or does this get bumped up by one due to it being seen in master?

contrib/rpm/buildah.spec
```diff
@@ -26,7 +26,7 @@
 
 Name:           buildah
 # Bump version in buildah.go too
-Version:        1.0
+Version:        1.2
```
TomSweeneyRedHat: 1.1?

buildah.go
rhatdan: There are two Patches here.  First one bumps to 1.1, second one to 1.2-dev.  We will tag and release the first one, the second one will just tell us if someone is running buildah off of master

When we release 1.2 we will do the same thing again.  This is how podman releases work.

TomSweeneyRedHat: Thanks for the 411.

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
Well as soon as I sent the LGTM, the Travis CI flipped from green to red on my screen....

TomSweeneyRedHat:
Looks like the version.bats tests aren't happy with something.

```
not ok 160 buildah version current in .spec file Version
# (in test file ./version.bats, line 15)
#   `[ "$status" -eq 0 ]' failed
not ok 161 buildah version current in .spec file changelog
# (in test file ./version.bats, line 21)
#   `[ "$status" -eq 0 ]' failed
/home/travis/.travis/job_stages: line 169: 24709 Terminated              travis_jigger $! $timeout $cmd
```

pixdrift:
@TomSweeneyRedHat, it's because the output from `buildah version` doesn't match what's in the changelog / spec file, which is correct behaviour for the test.

The version going in this commit should be stamped 1.1, then another commit to bump to 1.2?

pixdrift:
The most recent changelog entry in the spec needs to be 1.2 to pass.

rhatdan:
@pixdrift I am trying to allow adding -dev to version number so we can identify when people are testing from Master branch versus something we shipped.

pixdrift:
@rhatdan, that's fine. But the check is checking that the latest changelog entry matches the version number returned by buildah. In this commit the version is 1.2-dev, but the latest commit in the spec file is 1.1-1

contrib/rpm/buildah.spec
```diff
@@ -91,6 +91,99 @@ make DESTDIR=%{buildroot} PREFIX=%{_prefix} install install.completions
 %{_datadir}/bash-completion/completions/*
 
 %changelog
+* Sat Jun 9 2018 Dan Walsh <dwalsh@redhat.com> 1.1-1
```
pixdrift: This line here needs 1.2-dev to pass the check as that is what buildah is outputting, really should be 1.2-dev-1... but anyway.

tests/version.bats
```diff
@@ -11,12 +11,12 @@ load helpers
...
 	rversion=$(cat ${TESTSDIR}/../contrib/rpm/buildah.spec | awk '/^Version:/ { print $NF }')
-	run test "${bversion}" = "${rversion}"
+	run test "${bversion}" = "${rversion}" -o "${bversion}" = "${rversion}-dev"
 	[ "$status" -eq 0 ]
 }
 
 @test "buildah version current in .spec file changelog" {
 	bversion=$(buildah version | awk '/^Version:/ { print $NF }')
-	run bash -c "grep -A1 ^%changelog ${TESTSDIR}/../contrib/rpm/buildah.spec | grep -q \" ${bversion}-.*$\""
+	run bash -c "grep -A1 ^%changelog ${TESTSDIR}/../contrib/rpm/buildah.spec | grep -q \" ${bversion}.*$\""
```
pixdrift: Can I nit, and ask that this be reverted? ${bversion}- is correct, in this case 1.2-dev with -1. This was to stop people just pasting in the version number, when you should also include the package release number. I can revert it after the commit as I know this commit is something you want to get through.

pixdrift:
It's quite peculiar that CentOS 7 and Fedora 27 passed on that RPM version check, but Fedora 28 failed.

rhatdan:
bot, retest this please

rhatdan:
Finally.

pixdrift:
Well done :tada:

pixdrift:
Looks good @rhatdan 👍 quite the changelog 😄

tests/rpm.bats
```diff
@@ -167,7 +167,7 @@ load helpers
         echo "buildah version: $bv"
         echo "buildah rpm version: $rv"
         test $shortcommit = $id
-        test $bv = $rv
+	test $bv = ${rv} -o $bv = ${rv}-dev
```
TomSweeneyRedHat: :+1:

PR #771, opened 2018-06-10
Minor update to buildah config documentation for entrypoint

pixdrift:


docs/buildah-config.md
```diff
@@ -53,7 +53,7 @@ using the specified container.
 
 Note: this setting is not present in the OCIv1 image format, so it is discarded when writing images using OCIv1 formats.
 
-**--entrypoint** *entry*
+**--entrypoint** *"string"* | *'["array_entry1", "array_entry2", ...]'*
```
rhatdan: How about 
**--entrypoint** *"string"* | *'["command", "arg1", ...]'*

pixdrift: Cheers. Done.

pixdrift: For consistency, should "string" be "command" also?

rhatdan: Yes I was thinking that also.

ripcurld: I think only 

**--entrypoint** "command" "arg1",...

is the best.

pixdrift: @ripcurld this doesn't show that entrypoint takes string _or_ array.

rhatdan: @ripcurld I agree with @pixdrift 
we are trying to show that --entrypoint takes input as either a simple command string, or as an array of command followed by argumets '[ "command", "arg1", "arg3" ]' 

--entrypoint "command" "arg1" 
is invalid


--entrypoint "command"
or 
--entrypoint  '[ "command", "arg1", "arg3" ]' 
Are valid

Issue #684, opened 2018-05-13
buildah bud fails at commit stage for layered images from private registry with internal error


Conan-Kudo:
**Description**

Whenever I'm building a layered image from my private registry (OpenShift Origin 1.5 registry), it breaks at the end with weird messages like this:

```
STEP 4: COMMIT containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.override_kernel_check=true]@ac887cd1c3698c3247f983681a24daeb2e35e9497f0a9e77fc120a310c43cf7b
error copying layers and metadata: Error initializing source containers-storage:centos-buildah-working-container: internal error: history lists 5 non-empty layers, but we have 10 layers on disk
```

**Steps to reproduce the issue:**
1. Set up Origin 1.5
2. Build a container in it
3. Use `buildah bud` to build a container layered on top of that one.

Example "base" container:
```
FROM centos:7
RUN yum upgrade -y && yum -y install buildah && rm -rf /var/cache/yum/*
CMD /bin/bash
```

Example "layered" container:
```
FROM registry.localhost.localdomain/test/centos-buildah
RUN yum -y install docker && rm -rf /var/cache/yum/*
CMD /bin/bash
```

**Describe the results you received:**
It fails to build due to internal error with history of layers, as noted in the description.

**Describe the results you expected:**
I expected the commit stage to succeed and the image to build.

**Output of `rpm -q buildah`:**

On my CentOS 7 box:
```
buildah-0.11-3.git49095a8.el7.x86_64
```
On my Fedora 28 box:
```
buildah-0.16-1.git82529cd.fc28.x86_64
```

**Output of `buildah version`:**

On my CentOS 7 box:
```
Version:       0.11
Go Version:    go1.8.3
Image Spec:    1.0.0
Runtime Spec:  1.0.0
Git Commit:    49095a8
Built:         Wed Mar  7 17:03:34 2018
OS/Arch:       linux/amd64
```

On my Fedora 28 box:
```
Version:       0.16
Go Version:    go1.10
Image Spec:    1.0.0
Runtime Spec:  1.0.0
Git Commit:    82529cd
Built:         Wed Apr  4 10:22:20 2018
OS/Arch:       linux/amd64
```

**Output of `cat /etc/*release`:**

CentOS 7 box:
```
CentOS Linux release 7.4.1708 (Core)
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
[...]

PR #762, opened 2018-06-07
Do not add EmptyLayer entries to DiffID

mtrmac:
This is an immediate fix for the assertion failure in #684, but it’s not at all clear to me that this is the _right_ fix.

@nalind RFC

PR #775, opened 2018-06-11
DO NOT MERGE TEST ONLY MODIFIES VENDOR: Ignore EmptyLayer layers on storage input

mtrmac:
TESTING ONLY this approach for fixing #684.

### DIFF
--- a/vendor/github.com/containers/image/README.md
+++ b/vendor/github.com/containers/image/README.md
@@ -65,7 +65,9 @@ the primary downside is that creating new signatures with the Golang-only implem
 - `containers_image_ostree_stub`: Instead of importing `ostree:` transport in `github.com/containers/image/transports/alltransports`, use a stub which reports that the transport is not supported. This allows building the library without requiring the `libostree` development libraries. The `github.com/containers/image/ostree` package is completely disabled
 and impossible to import when this build tag is in use.
 
-## Contributing
+## [Contributing](CONTRIBUTING.md)**
+
+Information about contributing to this project.
 
 When developing this library, please use `make` (or `make … BUILDTAGS=…`) to take advantage of the tests and validation.
 
--- a/vendor/github.com/containers/image/copy/copy.go
+++ b/vendor/github.com/containers/image/copy/copy.go
@@ -347,6 +347,9 @@ func checkImageDestinationForCurrentRuntimeOS(ctx context.Context, sys *types.Sy
 
 // updateEmbeddedDockerReference handles the Docker reference embedded in Docker schema1 manifests.
 func (ic *imageCopier) updateEmbeddedDockerReference() error {
+	if ic.c.dest.IgnoresEmbeddedDockerReference() {
+		return nil // Destination would prefer us not to update the embedded reference.
+	}
 	destRef := ic.c.dest.Reference().DockerReference()
 	if destRef == nil {
 		return nil // Destination does not care about Docker references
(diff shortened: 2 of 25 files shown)


mtrmac:
Note to self: This may break existing installations - if the top layers are `EmptyLayer`, the topmost layer for an image will now be a the topmost non-empty one, and the code expects images with the same config to have the same topmost layer.  I.e. `pull foo; update c/image; pull foo` could break.

mtrmac:
@nalind RFC.  Instead of expecting `EmptyLayer` layers to exist, expect them not to exist, and modify c/image/storage to ignore them when pulling images.

This is pretty much the opposite to the attempt in #762 (which expected `EmptyLayer` to exist, as they do in v2s1, but they don’t exist at all in v2s2/OCI).

See also https://github.com/projectatomic/buildah/pull/775#issuecomment-396321518 , the
```go
		if img.TopLayer != lastLayer {
			logrus.Debugf("error creating image: image with ID %q exists, but uses different layers", intendedID)
			return errors.Wrapf(storage.ErrDuplicateID, "image with ID %q already exists, but uses a different top layer", intendedID)
		}
```
check could be broken.  We could just remove it, or maybe compute both the top non-empty and top possibly-empty layer, and allow reusing images with either.

mtrmac:
This is known to be insufficient for @pixdrift ’s reproducer from #684 , see #776 for the extra needed fix.

rhatdan:
@mtrmac Since https://github.com/projectatomic/buildah/pull/776 has merged, can this move forward?

mtrmac:
> @mtrmac Since #776 has merged, can this move forward?

It’s very likely this breaks corner cases (pulling the same image after upgrade, because the top layer ID changes; pulling v2s1 and pushing it out again, because the intermediate “throwaway” layers are not stored), I am still looking into that.

I’d also _very much_ appreciate a review by someone who understands c/storage and knows the context I don’t know I don’t know.

rhatdan:
@nalind we need you to review this. ASAP

TomSweeneyRedHat:
I might be off base, but thought this PR was not going to be merged.  Rather @mtrmac was going to open one in containers/image and then Buildah would have to vendor that in.  The changes s/b the same though, so a review would be good regardless.

mtrmac:
> I might be off base, but thought this PR was not going to be merged. Rather @mtrmac was going to open one in containers/image and then Buildah would have to vendor that in.

Good point, yes, that’s the very likely outcome. But right now I don’t have a comprehensive c/image PR, so this is the only place I where can currently ask for a review of the core idea.

mtrmac:
Please see https://github.com/containers/image/pull/472 for what could be a comprehensive fix. This PR now only vendors that branch to run `buildah` tests.

**NOTE**: That PR punts on handling pre-existing v2s1 images; users may have to delete/untag them and re-pull using the fixed code. Because comprehensively supporting the old representation would be a fair amount of complex code (and mean a commitment to add such support to any future layer-walking code), and because quite a few uses of v2s1 haven’t worked correctly anyway, on balance it seems to me that giving up is the better long-term choice.  OTOH possible deployments that only run containers (k8s nodes?), and are configured to re-pull the same image again for some reason, may be affected; please speak up in https://github.com/containers/image/pull/472 if this is not acceptable.

mtrmac:
#860 is equivalent, closing this one to simplify the PR list.

PR #779, opened 2018-06-12
tests: Add tests

q384566678:
After adding these two tests, all commands have corresponding test files.

### DIFF
--- a/tests/images.bats
+++ b/tests/images.bats
@@ -1,0 +1,70 @@
+#!/usr/bin/env bats
+
+load helpers
+
+@test "images" {
+  cid1=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+  cid2=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json busybox)
+  run buildah --debug=false images
+  [ $(wc -l <<< "$output") -eq 3 ]
+  [ "${status}" -eq 0 ]
+  buildah rm -a
+  buildah rmi -a -f
+}
+
+@test "images filter test" {
+  cid1=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+  cid2=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json busybox)
+  run buildah --debug=false images --filter since=alpine
+  [ $(wc -l <<< "$output") -eq 2 ]
+  [ "${status}" -eq 0 ]
+  buildah rm -a
+  buildah rmi -a -f
+}
+
+@test "images format test" {
+  cid1=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+  cid2=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json busybox)
+  run buildah --debug=false images --format "{{.Name}}"
+  [ $(wc -l <<< "$output") -eq 1 ]
+  [ "${status}" -eq 0 ]
+  buildah rm -a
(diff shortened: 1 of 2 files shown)


tests/images.bats
```diff
@@ -0,0 +1,70 @@
...
+  [ "${status}" -eq 0 ]
+  buildah rm -a
+  buildah rmi -a -f
+}
+
+@test "images format test" {
+  cid1=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+  cid2=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json busybox)
+  run buildah --debug=false images --format "{{.Name}}"
+  [ $(wc -l <<< "$output") -eq 1 ]
```
q384566678: My expectation is to output two lines. But I don't understand why only one line was output in the test.

rhatdan: Did you fix this?

rhatdan:
LGTM

rhatdan:
@TomSweeneyRedHat @umohnani8 @nalind PTAL

tests/images.bats
TomSweeneyRedHat: I'm seeing two lines when I run this "by hand", which is the result I'd expect to see.

```
# buildah images --format "{{.Name}}"
docker.io/library/alpine:latest
docker.io/library/busybox:latest
```

umohnani8:
LGTM

rhatdan:
@rh-atomic-bot r+

tests/images.bats
TomSweeneyRedHat: @rhatdan something funky with this.  This should be "-eq 2" not "-eq 1", I'm not sure why it's passing.

TomSweeneyRedHat:
OK, curiosity killed the t-cat.  When we do `buildah images --format ...`  we do this after each image:

```
        if terminal.IsTerminal(int(os.Stdout.Fd())) {
                fmt.Println()
        }
```
at https://github.com/projectatomic/buildah/blob/master/cmd/buildah/images.go#L403.  In the bats environment where the tests run, there's no associated tty so the carriage return is not printed out and the end result is one line of images jammed together, that's why the format test failed checking for mor than one line as expected.  In a standard terminal, it returns two lines.

@rhatdan @nalind @umohnani8 unless you can think of a reason to keep this term check, I'm going to remove it.  It seems to work fine in both a standard terminal and in the bats tests without it.

umohnani8:
@TomSweeneyRedHat I remember putting that there because it was jamming everything into one line. It was not returning a new line after each line. I just tested it without that and I am still seeing the same thing.
With the check:
```
➜  buildah git:(cache) ✗ sudo buildah images --format "{{.ID}}"
cc510acfcd701a409014118d5f417f0022520802a26c650866b8a9594d75f3a7
3fd9065eaf02feaf94d68376da52541925650b81698c53c6824d92ff63f98353
```
Without the check:
```
➜  buildah git:(cache) ✗ sudo buildah images --format "{{.ID}}"
cc510acfcd701a409014118d5f417f0022520802a26c650866b8a9594d75f3a73fd9065eaf02feaf94d68376da52541925650b81698c53c6824d92ff63f98353%
```

I use zshell, so could be because of that, not sure...

rhatdan:
Sounds familiar.  Is there a way to only do this if when there is multiple lines of output?

TomSweeneyRedHat:
Created #782 to hopefully address the one line return for the formats test in bats.

PR #786, opened 2018-06-12
Add environment variable BUILDAH_RUNTIME

rhatdan:
Allow user to setup alternate runtimes to use rather then
runc.

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"net/url"
+	"os"
 	"path"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


util/util.go
```diff
@@ -233,3 +234,12 @@ func WriteError(w io.Writer, err error, lastError error) error {
...
 	return err
 }
+
+// Runtime is the default command to use to run the container.
+func Runtime() string {
+	runtime := os.Getenv("BUILDAH_RUNTIME")
+	if runtime != "" {
+		return runtime
+	}
+	return DefaultRuntime
```
TomSweeneyRedHat: I'm not seeing this being set anywhere, only some removals from old code....

rhatdan: It is in util/types.go

rhatdan:
@TomSweeneyRedHat Theoretically we support all OCI Runtimes.

TomSweeneyRedHat:
Thx for the pointer @rhatdan, LGTM

rhatdan:
@umohnani8 @pixdrift PTAL

rhatdan:
@rh-atomic-bot r+

Issue #785, opened 2018-06-12
ERRO[0030] unable to write container terminal output: resource temporarily unavailable


carlwgeorge:
**Description**

While installing packages via `buildah run`, the program appears to lock up.  This error message is displayed in the terminal.

```
ERRO[0030] unable to write container terminal output: resource temporarily unavailable
```

It just stays on that error permanently until I ctrl+c the process.  With the error on the screen the working container is up.

```
# buildah containers
CONTAINER ID  BUILDER  IMAGE ID     IMAGE NAME                       CONTAINER NAME
1c9542922661     *     801894bc0e43 registry.fedoraproject.org/fedora:27 fedora-working-container
```

**Steps to reproduce the issue:**
1. `buildah from registry.fedoraproject.org/fedora:27`
2. `buildah run fedora-working-container -- dnf --assumeyes install bash-completion dnf-utils less procps-ng python2-rpm-macros python3-rpm-macros rpmdevtools rpmlint tree vim-enhanced wget`

It does *not* happen every time, but rather about half of the times I try to reproduce it on my system.

**Describe the results you received:**
`ERRO[0030] unable to write container terminal output: resource temporarily unavailable`

**Describe the results you expected:**
command to finish successfully

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.1-1.gitfbf46d3.fc27.x86_64
```

<details>
<summary><strong>Output of <code>buildah version</code>:</strong></summary>

```
Version:         1.1
Go Version:      go1.9.6
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.3.1
libcni Version:  v0.6.0
Git Commit:      fbf46d3
Built:           Sun Jun 10 16:30:19 2018
OS/Arch:         linux/amd64
```
</details>

<details>
<summary><strong>Output of <code>cat /etc/*release</code>:</strong></summary>

```
Fedora release 27 (Twenty Seven)
NAME=Fedora
VERSION="27 (Twenty Seven)"
ID=fedora
VERSION_ID=27
PRETTY_NAME="Fedora 27 (Twenty Seven)"
ANSI_COLOR="0;34"
CPE_NAME="cpe:/o:fedoraproject:fedora:27"
HOME_URL="https://fedoraproject.org/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
[...]

PR #787, opened 2018-06-13
Separate stdio buffering from writing

nalind:
When relaying stdio data to and from a container, separate the "reading and buffering from a read descriptor" step from the "relaying to a write descriptor" step, so that we can try to flush buffers that have data in them even when there's no new data to be read.

Treat `EAGAIN` as a recoverable error when writing, since we're now able to come back and try again later.  I haven't been able to reproduce #785, but this _should_ address it.


TomSweeneyRedHat:
LGTM, but want another head not on this one.

Passes baseline tests and with it, I'm not able to duplicate the issue in #785.  However I was not able to duplicated #785 without this fix either.  Regardless, I think this is a better way to go, nice work @nalind

PR #788, opened 2018-06-13
Add --rm and --force-rm to buildah bud

umohnani8:
--rm removes all intermediate containers if the build was successful.
If unsuccessful, the intermediate container at which the failure occured
will persist in the container store.
--force-rm removes all intermediate containers even if the build was
unsuccessful.


umohnani8:

imagebuildah/build.go
```diff
@@ -1047,7 +1066,12 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) error
 		if err := stageExecutor.Prepare(ctx, stage.Builder, stage.Node, ""); err != nil {
 			return err
 		}
-		defer stageExecutor.Delete()
+		// Always remove intermediate containers if either b.forceRmIntermediateCtrs is true
+		// or if building without layers.
+		// Will remove intermediate containers even if build fails.
+		if b.forceRmIntermediateCtrs || (!b.layers && !b.noCache) {
```
rhatdan: I think b.layers and this flag should conflict, so this situation would not happen.
--layers=false --rm 
--layers=false --force-rm
Should return an error

Not sure about the b.noCache flag here?  Doesn't this mean the user chose to not use the cache during a rebuild?  Wouldn't he still want the layers removed?

tests/bud/use-layers/Dockerfile.fail-case
```diff
@@ -0,0 +1,5 @@
+FROM alpine
+RUN mkdir /hello
+RUN touch blah.txt
+COPY non-existent world/
+ENV foo=bar
```
rhatdan: nit: add return

tests/bud.bats
```diff
@@ -33,6 +33,36 @@ load helpers
...
+
+  buildah rm -a
+  buildah rmi -a
+}
+
+@test "bud with --force-rm flag" {
+  run buildah bud --signature-policy ${TESTSDIR}/policy.json --force-rm --layers -t test1 -f Dockerfile.fail-case ${TESTSDIR}/bud/use-layers
+  echo "$output"
+  [ "$status" -ne 0 ]
+  ctrs1=$(buildah containers | grep "" -c)
```
rhatdan: nit: Weird way to count lines.
`wc -l`
Would make more sense to me.

pkg/cli/common.go
```diff
@@ -147,9 +147,9 @@ var (
 			Name:  "quiet, q",
 			Usage: "refrain from announcing build instructions and image read/write progress",
 		},
-		cli.BoolFlag{
+		cli.BoolTFlag{
 			Name:  "rm",
-			Usage: "Remove intermediate containers after a successful build. The build process does not currently support caching so this is a NOOP.",
+			Usage: "Remove intermediate containers after a successful build (default true)",
```
rhatdan: I was suprised to see this defaulted to true.  Does this mean that if I run 
buildah bud --layers .
If the build is successful and I go add a new line to the Dockerfile at the end, and run the same command again, it will go through every step, nothing will be cached?

imagebuildah/build.go
umohnani8: Right, so if we are not building with layers, this if condition shouldn't be true as the removal of the container will be handled by the defer. This is to ensure that if a build fails and --force-rm=false then the container where the failure occurred persists.

pkg/cli/common.go
umohnani8: No, this just means that the containers created during the build process will be removed. After each stem the container is committed to an intermediate image so the cache uses those images. The intermediate containers don't really come into play unless you want to keep them around for debugging I guess.

TomSweeneyRedHat:
LGTM other than @rhatdan's questions.

tests/bud.bats
umohnani8: Using wc instead.

tests/bud/use-layers/Dockerfile.fail-case
umohnani8: Fixed

imagebuildah/build.go
umohnani8: Reworded the comment, hopefully it is clearer now.

umohnani8:
Made `layers` and `no-cache` conflict. Also if `force-rm` or `rm` is set without either `layers` or `no-cache` it will error out.
Added examples to the doc.

umohnani8:
Rebased.

docs/buildah-bud.md
```diff
@@ -163,7 +163,7 @@ context.
 
 **--force-rm**
 
-Always remove intermediate containers after a build. Buildah does not currently support caching so this is a NOOP.
+Always remove intermediate containers after a build, even if the build is unsuccessful..
```
TomSweeneyRedHat: should line 164 and 272 have bool-value like tls-verify does?

umohnani8: Sure! Fixed :)

rhatdan:
@rh-atomic-bot r+

bendem:
> Made `layers` and `no-cache` conflict.

@umohnani8, I know this was a long time ago, but I'd love some insight into the rational for this (if you can shime in on #3861).

PR #793, opened 2018-06-18
setupSeccomp(): refactor

nalind:
Rework `setupSeccomp()` to use `switch{}` instead of multiple `if{}` tests when deciding how to set the Seccomp configuration for a container.


rhatdan:
@rh-atomic-bot r+

PR #794, opened 2018-06-18
Escape use of '_' in .md docs

nalind:
Escape uses of _ in markdown docs, so that it isn't mistaken for a formatting character.  Fix a couple of copy-paste errors.


rhatdan:
@rh-atomic-bot r+

PR #795, opened 2018-06-18
Changes to terminal handling

nalind:
Default to creating a pseudo-terminal only when all of stdin, stdout, and stderr are connected to terminals.  Remove the "reading" variable from `runCopyStdio()`, which was always just the size of the `relayMap` map.  Add a field to `RunOptions` that lets callers pass in a terminal size for cases where we can't/shouldn't read the size from a controlling terminal.  Move a comment about `SIGWINCH` to the right place.  Stop polling on a descriptor if we get a `POLLNVAL` event for it.

### DIFF
--- a/run.go
+++ b/run.go
@@ -1386,11 +1386,14 @@ func runCopyStdio(stdio *sync.WaitGroup, copyPipes bool, stdioPipe [][]int, copy
 		stdio.Done()
 		finishedCopy <- struct{}{}
 	}()
-	// If we're not doing I/O handling, we're done.
-	if !copyConsole && !copyPipes {
-		return
-	}
-	terminalFD := -1
+	// Map describing where data on an incoming descriptor should go.
+	relayMap := make(map[int]int)
+	// Map describing incoming and outgoing descriptors.
+	readDesc := make(map[int]string)
+	writeDesc := make(map[int]string)
+	// Buffers.
+	relayBuffer := make(map[int]*bytes.Buffer)
+	// Set up the terminal descriptor or pipes for polling.
 	if copyConsole {
 		// Accept a connection over our listening socket.
 		fd, err := runAcceptTerminal(consoleListener, spec.Process.ConsoleSize)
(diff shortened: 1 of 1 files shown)


run.go
```diff
@@ -602,15 +606,24 @@ func setupApparmor(spec *specs.Spec, apparmorProfile string) error {
 	return nil
 }
 
-func setupTerminal(g *generate.Generator, terminalPolicy TerminalPolicy) {
+func setupTerminal(g *generate.Generator, terminalPolicy TerminalPolicy, terminalSize *specs.Box) {
 	switch terminalPolicy {
 	case DefaultTerminal:
-		g.SetProcessTerminal(terminal.IsTerminal(int(os.Stdout.Fd())))
+		onTerminal := terminal.IsTerminal(unix.Stdin) && terminal.IsTerminal(unix.Stdout) && terminal.IsTerminal(unix.Stderr)
```
rhatdan: So if anyone is not a terminal you default everything to not being a terminal>

run.go
```diff
@@ -1549,7 +1560,7 @@ func runCopyStdio(stdio *sync.WaitGroup, copyPipes bool, stdioPipe [][]int, copy
 	}
 }
 
-func runAcceptTerminal(consoleListener *net.UnixListener) (int, error) {
+func runAcceptTerminal(consoleListener *net.UnixListener, consoleSize *specs.Box) (int, error) {
```
rhatdan: nit, why not keep this terminalSize?

run.go
nalind: Yeah.  Trying to get "not a terminal" when we're being invoked directly (so we can't use the `--tty` flag) is a guessing game otherwise.  This ends up changing the default for `build-using-dockerfile`, since there we always connect `stdin` to `/dev/null` for `RUN` instructions.

rhatdan: ok

rhatdan:
LGTM, One Nit.

rhatdan:
LGTM, One Nit.

run.go
nalind: No strong reason either way.  I can change it back if you have a preference.

rhatdan: Well yo ucalled it with terminalSize, so lets be consistant then we can merge.

nalind: Okay, changing it.

nalind:
Updated.

nalind:
bot, retest this please

rhatdan:
LGTM
bot, retest this please

rhatdan:
@umohnani8 @TomSweeneyRedHat @pixdrift @giuseppe PTAL

nalind:
Tweaked the logic - we weren't setting the terminal size when `unix.Stdin` wasn't a terminal, even if a size was explicitly passed to us as part of the configuration.

umohnani8:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #797, opened 2018-06-18
Run(): don't use a callback when a slice will do

nalind:
`moreCreateArgs()` doesn't need to be a function; it can just be a slice.


rhatdan:
LGTM
@umohnani8 @TomSweeneyRedHat @giuseppe PTAL

TomSweeneyRedHat:
LGTM, but thought I'd done so a while ago....

rhatdan:
@rh-atomic-bot r+

giuseppe:
LGTM

PR #799, opened 2018-06-18
Run(): add options for adding and removing capabilities

nalind:
Add `RunOptions` and `BuildOptions` flags for modifying the list of granted capabilities from the default.

Default to dropping CAP_NET_RAW.

### DIFF
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -48,6 +48,26 @@ resulting image's configuration.
 
 Images to utilise as potential cache sources. Buildah does not currently support caching so this is a NOOP.
 
+**--cap-add**=*CAP\_xxx*
+
+When executing RUN instructions, run the command specified in the instruction
+with the specified capability added to its capability set.
+Certain capabilities are granted by default; this option can be used to add
+more.
+
+**--cap-drop**=*CAP\_xxx*
+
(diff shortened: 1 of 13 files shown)


docs/buildah-bud.md
```diff
@@ -48,6 +48,24 @@ resulting image's configuration.
...
+
+When executing RUN instructions, run the command specified in the instruction
+with the specified capability removed from its capability set.
+Certain capabilities are granted by default; this option can be used to remove
+them.
+
+If a capability is specified to both the **--cap-add** and **--cap-drop**
+options, it will be dropped, regardless of the order in which the options were
+given.
+
```
TomSweeneyRedHat: Will need to add these to the podman build man page too, note to self.

rhatdan:
Yes I think the man page should container a list of capabiltiies we are using, and we should just specify  a list in the code rather then relying on the underlying tools to specify a list.

nalind:
Okay, switching away from using runtime-tools's defaults.

nalind:
Updated.

docs/buildah-bud.md
rhatdan: yes. I would love to see us add this list to inspect also, not in this PR though.

rhatdan:
@rh-atomic-bot r+

PR #800, opened 2018-06-18
bind: factor out RecursiveUnmount

nalind:
Factor out logic for recursively unmounting our bind mounts, and make use of the parent/child relationship between mountpoints to decide which order to use, rather than sorting based on their mountpoint paths.

### DIFF
--- a/bind/mount.go
+++ b/bind/mount.go
@@ -6,8 +6,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"sort"
-	"strings"
 	"syscall"
 
 	"github.com/containers/storage/pkg/idtools"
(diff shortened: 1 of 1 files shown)


rhatdan:
Seems rather brutal.  But LGTM

bind/mount.go
```diff
@@ -201,3 +198,97 @@ func leaveBindMountAlone(mount specs.Mount) bool {
 	}
 	return false
 }
+
+// RecursiveUnmount unmounts the given mountpoints and anything that's hanging
+// off of them, rather aggressively.  If a mountpoint also appears in the
+// mountpointsToRemove slice, the mountpoints are removed after they are
+// unmounted.
+func RecursiveUnmount(mountpoint string, mountpointsToRemove []string) error {
```
TomSweeneyRedHat: nit on the name, not really recursive is it?  FullUnmount, UmountAll, HardUmount, or something similar instead?

nalind: The function isn't recursive, true.  The name mirrors that of the similar function in the `github.com/containers/storage/pkg/mount` package, which does something similar, but which selects the order in which it unmounts things by sorting the directory paths of the mountpoints, as we did previously, and which doesn't detach busy filesystems.  If you have a preference for a different name, I'm happy to change it.

TomSweeneyRedHat: Would UnmountMountpoints work for you?

nalind: Sure.  Making the change.

nalind:
Rebased.

TomSweeneyRedHat:
LGTM, thanks for the funtion name change to keep my OCD at bay!  :^)

rhatdan:
@rh-atomic-bot r+

PR #802, opened 2018-06-18
Run(): don't set a Pdeathsig for the runtime

nalind:
Don't try to set the `Pdeathsig` attribute on the runtime process when we call the runtime.  Whether we should try to do that is debatable, and it seems to cause us to interact badly with strace(1).


rhatdan:
LGTM
@umohnani8 @pixdrift @TomSweeneyRedHat @giuseppe PTAL

umohnani8:
LGTM

TomSweeneyRedHat:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #803, opened 2018-06-18
Allow --userns-uid-map/--userns-gid-map to be global options

nalind:
Allow --userns-uid-map and --userns-gid-map to be specified as global options, which can be overridden by options specified to commands which know them as non-global options.

This will mainly benefit tools that want to set ID mappings for buildah to use, but which don't want to have to figure out whether or not they're invoking it with a subcommand that recognizes these options.

### DIFF
--- a/cmd/buildah/common.go
+++ b/cmd/buildah/common.go
@@ -11,6 +11,7 @@ import (
 	digest "github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
 	"github.com/projectatomic/buildah"
+	"github.com/projectatomic/buildah/util"
 	"github.com/urfave/cli"
 )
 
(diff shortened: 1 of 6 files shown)


cmd/buildah/main.go
```diff
@@ -61,6 +61,14 @@ func main() {
 			Usage: "storage driver option",
 			Value: defaultStoreDriverOptions,
 		},
+		cli.StringSliceFlag{
+			Name:  "userns-uid-map",
+			Usage: "default `ctrID:hostID:length` UID mapping to use",
+		},
```
rhatdan: Should we make these hidden options so we don't need to document?

nalind: It's tempting (the language in the man pages about how the mapping options interact could stand to be simplified, but it's about as simple as I can make it), but I'm planning to have `buildah` itself use this in a later PR, and I think we'd be obliged to document it then.

cmd/buildah/common.go
```diff
@@ -33,6 +34,22 @@ func getStore(c *cli.Context) (storage.Store, error) {
...
+		uopts := c.GlobalStringSlice("userns-uid-map")
+		gopts := c.GlobalStringSlice("userns-gid-map")
+		if len(uopts) > 0 && len(gopts) > 0 {
+			uidmap, gidmap, err := util.ParseIDMappings(uopts, gopts)
+			if err != nil {
+				return nil, err
+			}
+			options.UIDMap = uidmap
+			options.GIDMap = gidmap
+		}
```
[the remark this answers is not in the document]
nalind: Seems unlikely to me, but no harm in returning an error for that case.

nalind:
Added checks for `c.GlobalStringSlice()` returning zero-length slices for `--userns-uid-map`/`--userns-gid-map` arguments.

rhatdan:
What happens if you specify the options both globally and locally?

podman --uidmap 1000,0,100 build --uidmap 2000,0.100 ~

rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #804, opened 2018-06-19
Makefile: add uninstall

q384566678:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,12 @@ install:
 	install -D -m0755 buildah $(DESTDIR)/$(BINDIR)/buildah
 	$(MAKE) -C docs install
 
+.PHONY: uninstall
+uninstall:
+	rm -f $(DESTDIR)/$(BINDIR)/buildah
+	rm -f $(PREFIX)/share/man/man1/buildah*.1
+	rm -f $(DESTDIR)/${BASHINSTALLDIR}/buildah
+
 .PHONY: install.completions
 install.completions:
 	install -m 644 -D contrib/completions/bash/buildah $(DESTDIR)/${BASHINSTALLDIR}/buildah


TomSweeneyRedHat:
LGTM, thought we'd need to remove the completions too, but they're apparently not installed with `make install`

Makefile
```diff
@@ -82,6 +82,11 @@ install:
 	install -D -m0755 buildah $(DESTDIR)/$(BINDIR)/buildah
 	$(MAKE) -C docs install
 
+.PHONY: uninstall
+uninstall:
+	rm -f $(DESTDIR)/$(BINDIR)/buildah
+	rm -f $(PREFIX)/share/man/man1/buildah*.1
```
rhatdan: What about completions and other stuff that you installed?

TomSweeneyRedHat: We only install these two with `make install`.  Should this remove everything that we could potentially install other commands like `make install.completions`?  @rhatdan should we add completions to the standard `make install`?

rhatdan: No I don't think so, but the reason for a make uninstall is to remove cruft from your system, I guess although it is seldom done.
If we are going to add a make uninstall, we should have either a make uninstall.completions... or just have make uninstall all of the stuff that the Makefile could have added.

q384566678: updated, Not sure if it is correct.

Makefile
```diff
@@ -82,6 +82,14 @@ install:
 	install -D -m0755 buildah $(DESTDIR)/$(BINDIR)/buildah
 	$(MAKE) -C docs install
 
+.PHONY: uninstall
+uninstall:
+	rm -f $(DESTDIR)/$(BINDIR)/buildah
+	rm -f $(DESTDIR)/$(BINDIR)/runc
```
rhatdan: runc did not come from this package.

Makefile
```diff
@@ -82,6 +82,14 @@ install:
...
+	rm -f $(PREFIX)/share/man/man1/buildah*.1
+	rm -f $(DESTDIR)/${BASHINSTALLDIR}/buildah
+	rm -r /opt/cni/bin
```
rhatdan: I would also leave this alone, runc and /opt/cni/bin can be used by other packages.

q384566678:
updated.

TomSweeneyRedHat:

rhatdan:
@rh-atomic-bot r+

PR #807, opened 2018-06-19
Restore CAP_NET_RAW

nalind:
Add CAP_NET_RAW back to the default list of granted capabilities.

Add capabilities lists to `inspect` output.


pixdrift:
Interested in how/why this came about.
Did something break that required this capability?
How is the default list of kernel caps determined?

rhatdan:
Me too.  I know that it is needed for ping, but not sure of anything else.  I want to remove it from CRI-O by default also.

TomSweeneyRedHat:
The actual changes LGTM, but am interested in the conversation started by others.

nalind:
This won't help unprivileged users, who currently won't get CAP_NET_RAW abilities on the host's network namespace, and privileged users are served equally well by #818, so dropping this one in favor of that.

PR #809, opened 2018-06-22
commit: recognize committing to second storage locations

nalind:
Recognize cases when we're committing to local storage, but in a location other than where the working container is, by checking if the base image (by ID, in case it's tagged and the tag points to a different image with different layers) is present in the destination's store.

In those cases, we can't just assume that we can skip exporting the base image's layers.


rhatdan:
LGTM
@umohnani8 @TomSweeneyRedHat @pixdrift PTAL

nalind:
Huh.  I was wondering if I broke the "quiet" test.

nalind:
bot, retest this please.

TomSweeneyRedHat:
Change LGTM, but some tests aren't hip.

umohnani8:
LGTM

nalind:
bot, retest this please.

rhatdan:
@rh-atomic-bot r+

PR #815, opened 2018-06-25
Makefile: fix dependencies for buildah

giuseppe:
add *.go files under vendor/ to the list of dependencies for buildah.

I've noticed that make didn't trigger a build after editing some files under vendor.


rhatdan:
I am not crazy about doing this, since it slows everyones compiles down, for little benefit, since the vendor directory should change seldom.

TomSweeneyRedHat:
Ditto @rhatdan's thought.  FWIW., what I've done is made a change in vendor, then just a small change somewhere in the "real" buildah files to make it compile.

giuseppe:
thanks, I am going to close this

PR #816, opened 2018-06-25
Fix ARGS parsing for run commands

rhatdan:
Currently we are not adding the ARGS passed in via Dockerfile
or --build-args into the running container as environment variables.

### DIFF
--- a/buildah.go
+++ b/buildah.go
@@ -105,6 +105,8 @@ func (p NetworkConfigurationPolicy) String() string {
 type Builder struct {
 	store storage.Store
 
+	// Args define variables that users can pass at build-time to the builder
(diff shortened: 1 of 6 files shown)


rhatdan:
Fixes https://github.com/projectatomic/buildah/issues/813

rhatdan:
I don't think so, args is not configuration for the container image, it is more about passing args around in a Dockerfile.

rhatdan:
@giuseppe @umohnani8 @nalind PTAL

umohnani8:
LGTM

rhatdan:
@rh-atomic-bot r=umohnani8

Issue #810, opened 2018-06-22
Fresh install "buildah from fedora" tries localhost registry


LemonIceScream:
**Description**

buildah is unable to pull an image using the examples:

    [pomelo ~] buildah from fedora
    Get https://localhost/v2/: dial tcp 127.0.0.1:443: connect: connection `refused`

I see references in the documentation to /etc/containers but it did not exist. After installing skopeo it appeared but continued to not work.

Running both "skopeo inspect docker://fedora" and "skopeo inspect docker://docker.io/fedora" look like they're working (produce two results from docker.io).

I feel like I am missing something. What do I need to configure? What's the documentation? Where can I find it?

**/etc/containers configuration**
```
[pomelo /etc/containers] find .
.
./registries.d
./registries.d/default.yaml
./policy.json
[pomelo /etc/containers] cat policy.json 
{
    "default": [
        {
            "type": "insecureAcceptAnything"
        }
    ],
    "transports":
        {
            "docker-daemon":
                {
                    "": [{"type":"insecureAcceptAnything"}]
                }
        }
}
[pomelo /etc/containers] cat registries.d/default.yaml 
# This is a default registries.d configuration file.  You may
# add to this file or create additional files in registries.d/.
#
# sigstore: indicates a location that is read and write
# sigstore-staging: indicates a location that is only for write
#
# sigstore and sigstore-staging take a value of the following:
#   sigstore:  {schema}://location
#
# For reading signatures, schema may be http, https, or file.
# For writing signatures, schema may only be file.

# This is the default signature write location for docker registries.
default-docker:
#  sigstore: file:///var/lib/atomic/sigstore
  sigstore-staging: file:///var/lib/atomic/sigstore

# The 'docker' indicator here is the start of the configuration
# for docker registries.
#
# docker:
#
#   privateregistry.com:
#    sigstore: http://privateregistry.com/sigstore/
#    sigstore-staging: /mnt/nfs/privateregistry/sigstore
```

**Output of `rpm -q buildah` or `apt list buildah`:**

```
Installed via Arch: https://www.archlinux.org/packages/community/x86_64/buildah/
skopeo: https://www.archlinux.org/packages/community/x86_64/skopeo/
```

**Output of `buildah version`:**

```
[pomelo ~] buildah version
Version:       1.0
Go Version:    go1.10.2
Image Spec:    1.0.0
Runtime Spec:  1.0.0
Git Commit:    1ab80bcd
Built:         Fri May 11 07:14:04 2018
[...]

PR #817, opened 2018-06-25
Add info on registries.conf to from manpage

TomSweeneyRedHat:

Add a DEPENDENCIES sub-section to the `buildah-from` man page.  I realize it's a bit unconventional, but I think we've had enough folks outside of Fedora/RHEL who've stumbled into this.  Hopefully this will help address this.  Hopefully addresses #810

Also snuck in a small example into `buildah inspect` to show how to format the returned data from that command.

### DIFF
--- a/docs/buildah-inspect.md
+++ b/docs/buildah-inspect.md
@@ -33,5 +33,7 @@ buildah inspect --type container containerID
 
 buildah inspect --type image imageID
 
+buildah inspect --format '{{.OCIv1.Config.Env}}' alpine
+
 ## SEE ALSO
 buildah(1)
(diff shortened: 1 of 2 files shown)


docs/buildah-inspect.md
```diff
@@ -33,5 +33,7 @@ buildah inspect --type container containerID
 
 buildah inspect --type image imageID
 
+buildah inspect --format '{{.OCIv1.Config.Env}}' alpine
```
rhatdan: This seems a little random, but I am fine with adding it.

rhatdan:
LGTM
@rh-atomic-bot r+

docs/buildah-inspect.md
TomSweeneyRedHat: I'm happy to change it to something else, I was mainly looking to get a multi-level OCI image field as an example as we'd no format example..

PR #818, opened 2018-06-25
Allow ping command without NET_RAW Capabilities

rhatdan:

### DIFF
--- a/run.go
+++ b/run.go
@@ -792,6 +792,11 @@ func setupNamespaces(g *generate.Generator, namespaceOptions NamespaceOptions, i
 			}
 		}
 	}
+	if configureNetwork {
+		for name, val := range util.DefaultNetworkSysctl {
+			g.AddLinuxSysctl(name, val)
+		}
+	}
 	return configureNetwork, configureNetworks, configureUTS, nil
 }
 
--- a/util/types.go
+++ b/util/types.go
@@ -27,4 +27,9 @@ var (
 		"CAP_SETUID",
 		"CAP_SYS_CHROOT",
 	}
+	// DefaultNetworkSysctl is the list of Kernel parameters which we
+	// grant by default to containers which are running under UID 0.
+	DefaultNetworkSysctl = map[string]string{
+		"net.ipv4.ping_group_range": "0 0",
+	}
 )


rhatdan:
@nalind What do you think of this as an alternative to https://github.com/projectatomic/buildah/pull/807

run.go
```diff
@@ -781,6 +781,7 @@ func setupNamespaces(g *generate.Generator, namespaceOptions NamespaceOptions, i
 				return false, nil, false, errors.Wrapf(err, "error adding new %q namespace for run", string(specs.NetworkNamespace))
 			}
 			configureNetwork = (policy != NetworkDisabled)
+			g.Spec().Linux.Sysctl = util.DefaultNetworkSysctl
```
nalind: I think you'd need to duplicate this at line 732 to get the same effect.

util/types.go
```diff
@@ -27,4 +27,9 @@ var (
 		"CAP_SETUID",
 		"CAP_SYS_CHROOT",
 	}
+	// DefaultNetworkSysctl is the list of Kernel parameters which we
+	// grant by default to containers which are running under UID 0.
+	DefaultNetworkSysctl = map[string]string{
+		"net.ipv4.ping_group_range": "0 1",
```
nalind: ICBW, but I _think_ this can be "0 0".

run.go
nalind: This should probably be call to `g.AddLinuxSysctl()`, since we try not to mix using the Generator object's methods with direct modifications to a Spec structure.

nalind:
What's the plan for net==host?

rhatdan:
net==host would mean that we instruct the user to set that himself.

util/types.go
rhatdan: I could not tell if this was starting and number versus start and end.

run.go
rhatdan: I looked for that but did not see it.

util/types.go
nalind: I'm not super-pumped about using this if the documentation is this scarce, but it's a range of values - the GID of a user would need to be greater than or equal to the first value, and less than or equal to the second value.  The default of `[1 0]` chosen so that when left unconfigured, access can't be granted by this mechanism.

nalind:
LGTM
@rh-atomic-bot r+

PR #819, opened 2018-06-25
build without seccomp

baude:
If the seccomp build tag is disabled or selinux is not detected, build without
them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 AUTOTAGS := $(shell ./btrfs_tag.sh) $(shell ./libdm_tag.sh) $(shell ./ostree_tag.sh) $(shell ./selinux_tag.sh)
-TAGS := seccomp
+TAGS ?= seccomp
 PREFIX := /usr/local
 BINDIR := $(PREFIX)/bin
 BASHINSTALLDIR=${PREFIX}/share/bash-completion/completions
--- a/config_noseccomp.go
+++ b/config_noseccomp.go
@@ -1,0 +1,11 @@
+// +build !seccomp
+
+package buildah
+
+import "github.com/opencontainers/runtime-spec/specs-go"
+
+func setupSeccomp(spec *specs.Spec, seccompProfilePath string) error {
+	// If no seccomp is being used, the Seccomp profile in the Linux spec
+	// is not set
+	return nil
+}
--- a/config_seccomp.go
+++ b/config_seccomp.go
@@ -1,0 +1,35 @@
+// +build seccomp
+
+package buildah
+
+import (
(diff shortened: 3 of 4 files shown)


baude:
A couple of comments on this:

1. I choose to separate by function (selinux/seccomp) rather than by OS (linux, darwin, windows).  If you prefer by OS, we will need to figure out a solution for Ubuntu which is linux but no selinux.
2. I could add a seccomp.sh script to the auto part of the Make if you prefer that approach.  Currently I think it is hardcoded or overridden by TAGS=

rhatdan:
Why do you need this?  SELinux is already isolated via the selinux code? Seems you are just making this more complicated?

nalind:
I think we need this to be able to eliminate a dependency on the libseccomp shared library, at least.  If that isn't a problem we have with libselinux, I'm fine with dropping it from the PR.

LGTM either way.

baude:
When buildah is vendored into podman, selinux is not isolated and cannot be isolated when vendored in.  I'm not looking to make anything difficult.  The very fact that selinux is imported in new.go causes compiling complaints.

TomSweeneyRedHat:
@baude travis ci isn't hip

Makefile
```diff
@@ -1,5 +1,5 @@
 AUTOTAGS := $(shell ./btrfs_tag.sh) $(shell ./libdm_tag.sh) $(shell ./ostree_tag.sh) $(shell ./selinux_tag.sh)
-TAGS := seccomp
+TAGS ?= seccomp
```
TomSweeneyRedHat: Makefile ignorance questtion.  What does the '?=' do vs ':='  ?

rhatdan:
What is the SELinux compiler complaint?  The SELinux bindings are all Go Based, at least as far as I know.

Makefile
nalind: I had to look this one up in the 'make' info pages: this defines TAGS if it isn't already defined.  We only define TAGS here, so I don't know that we gain anything by it, but it shouldn't break anything, either.

rhatdan:
LGTM

Although I want to create a new repository for github.com/containers/seccomp and move this out of Docker package.

Makefile
TomSweeneyRedHat: Thanks Nalin, I search for 15 or 20 minutes last night and never found it.

baude:
the upside is if you want to pass TAGS in, this allows you to do so.

nalind:
For reference, variables can be overridden [on the command line](https://www.gnu.org/software/make/manual/make.html#Overriding).

PR #820, opened 2018-06-25
Run(): don't include any default environment variables

nalind:
When copying environment variables from the image's configuration to the spec that we'll pass to the runtime, clear out any defaults that the generate package might be supplying.  Currently, that's "$TERM".

### DIFF
--- a/run.go
+++ b/run.go
@@ -795,6 +795,53 @@ func setupNamespaces(g *generate.Generator, namespaceOptions NamespaceOptions, i
 	return configureNetwork, configureNetworks, configureUTS, nil
 }
 
+// Search for a command that isn't given as an absolute path using the $PATH
+// under the rootfs.  We can't resolve absolute symbolic links without
+// chroot()ing, which we may not be able to do, so just accept a link as a
+// valid resolution.
+func runLookupPath(g *generate.Generator, command []string) []string {
+	// Look for the configured $PATH.
+	spec := g.Spec()
+	envPath := ""
+	for i := range spec.Process.Env {
+		if strings.HasPrefix(spec.Process.Env[i], "PATH=") {
+			envPath = spec.Process.Env[i]
+		}
+	}
+	// If there is no configured $PATH, supply one.
+	if envPath == "" {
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
Change LGTM, but unhappy test results.

nalind:
bot, retest this please

nalind:
Added a variation on how `exec.LookPath()` resolves commands to do so using the PATH we're going to set and the rootfs that we've just mounted.

run.go
```diff
@@ -795,6 +795,53 @@ func setupNamespaces(g *generate.Generator, namespaceOptions NamespaceOptions, i
...
+// chroot()ing, which we may not be able to do, so just accept a link as a
+// valid resolution.
+func runLookupPath(g *generate.Generator, command []string) []string {
+	// Look for the configured $PATH.
+	spec := g.Spec()
+	envPath := ""
+	for i := range spec.Process.Env {
+		if strings.HasPrefix(spec.Process.Env[i], "PATH=") {
+			envPath = spec.Process.Env[i]
+		}
```
rhatdan: Should you break at this point? Can the array have more then one PATH=?  In that case you only get the last.

run.go
```diff
@@ -795,6 +795,53 @@ func setupNamespaces(g *generate.Generator, namespaceOptions NamespaceOptions, i
...
+	if envPath == "" {
+		defaultPath := "/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/bin:/sbin"
+		envPath = "PATH=" + defaultPath
+		g.AddProcessEnv("PATH", defaultPath)
+	}
+	// No command, nothing to do.
+	if len(command) == 0 {
+		return command
+	}
+	// Command is already an absolute path, use it as-is.
```
rhatdan: Move to top of function?  Or do you want to make sure PATH is set?

rhatdan:
bot, retest this please

rhatdan:
LGTM
@TomSweeneyRedHat @umohnani8 @giuseppe @pixdrift PTAL

rhatdan:
@rh-atomic-bot r+

PR #822, opened 2018-06-25
Implement basic recognition of the "--isolation" option

nalind:
Add the basics of handling the "--isolation" option, though at the moment, the only recognized option is "oci", which is our default.


rhatdan:
Should the default be runc versus oci?  Is the goal of this to support kata or givisord containers?  Or are you looking at something else?

rhatdan:
Ok
LGTM

rhatdan:
Of course theoretically we could build a oci-chroot. :^)

rhatdan:
@umohnani8 @giuseppe @TomSweeneyRedHat PTAL

PR #823, opened 2018-06-25
main: if unprivileged, reexec in a user namespace

nalind:
If our CLI is invoked as an unprivileged user (uid != 0),
* create a namespace using our own UID and GID as "0" followed by the ranges matching our name and our primary group's name that we find in `/etc/subuid` and `/etc/subgid` (the latter by way of using `newuidmap` and `newgidmap`)
* re-exec ourselves inside of that user namespace, prepending global CLI arguments that:
  * override the driver from storage.conf with "vfs"
  * override the storage root from storage.conf with "$HOME/containers/storage"
  * override the storage runroot from storage.conf with "/var/run/user/$uid/containers/storage"
  * set default ID mapping settings to map all of the ranges matching our name and our primary group's name that we found in `/etc/subuid` and `/etc/subgid`
  * can still be overridden using the command line

Add a `buildah unshare` CLI that will start an arbitrary command in the first namespace, so that manual cleanup of locations used by the second namespace will be possible.

### DIFF
--- a/docs/buildah-unshare.md
+++ b/docs/buildah-unshare.md
@@ -1,0 +1,32 @@
+# buildah-unshare "19" "June 2018" "buildah"
+
+## NAME
+buildah\-unshare - Run a command inside of a modified user namespace.
+
+## SYNOPSIS
+**buildah** **unshare** [*options* [...] --] [**command**]
+
+## DESCRIPTION
+Launches a process (by default, *$SHELL*) in a new user namespace.  The user
+namespace is configured so that the invoking user's UID and primary GID appear
+to be UID 0 and GID 0, respectively.  Any ranges which match that user and
+group in /etc/subuid and /etc/subgid are also mapped in as themselves with the
+help of the *newuidmap(1)* and *newgidmap(1)* helpers.
+
+This is mainly useful for troubleshooting unprivileged operations and for
+manually clearing storage and other data related to images and containers.
+
+## EXAMPLE
+
+buildah unshare id
+
+buildah unshare pwd
+
+buildah unshare cat /proc/self/uid\_map
+
+buildah unshare cat /proc/self/gid\_map
+
+buildah unshare rm -fr $HOME/.local/share/containers/storage /var/run/user/\`id -u\`/run
+
+## SEE ALSO
+buildah(1), namespaces(7), newuidmap(1), newgidmap(1), user\_namespaces(7)
--- a/docs/buildah.md
+++ b/docs/buildah.md
@@ -46,23 +46,23 @@ be used, as the default behavior of using the system-wide configuration
 
 **--root** **value**
 
-Storage root dir (default: "/var/lib/containers/storage")
+Storage root dir (default: "/var/lib/containers/storage" for UID 0, "$HOME/.local/share/containers/storage" for other users)
 Default root dir is configured in /etc/containers/storage.conf
 
 **--runroot** **value**
 
-Storage state dir (default: "/var/run/containers/storage")
+Storage state dir (default: "/var/run/containers/storage" for UID 0, "/var/run/user/$UID/run" for other users)
 Default state dir is configured in /etc/containers/storage.conf
 
 **--storage-driver** **value**
 
-Storage driver.  Default Storage driver is configured in /etc/containers/storage.conf.  Overriding
-this option, will drop the storage-opt definitions was well from the storage.conf file.  User must
+Storage driver.  The default storage driver for UID 0 is configured in /etc/containers/storage.conf, and is *vfs* for other users.
+Overriding this option will cause the *storage-opt* settings in /etc/containers/storage.conf to be ignored.  The user must
 specify additional options via the `--storage-opt` flag.
 
 **--storage-opt** **value**
 
-Storage driver option, Default Storage driver options are configured in /etc/containers/storage.conf
+Storage driver option, Default storage driver options are configured in /etc/containers/storage.conf
 
 **--userns-uid-map** *mapping*
 
@@ -107,8 +107,9 @@ Print the version
 | buildah-run(1)        | Run a command inside of the container.                                                               |
 | buildah-tag(1)        | Add an additional name to a local image.                                                             |
 | buildah-umount(1)     | Unmount a working container's root file system.                                                      |
-| buildah-version(1)    | Display the Buildah Version Information                                    |
-| storage.conf(5) | Syntax of Container Storage configuration file |
+| buildah-unshare(1)    | Launch a command in a user namespace with modified ID mappings.                                      |
+| buildah-version(1)    | Display the Buildah Version Information                                                              |
+| storage.conf(5)       | Syntax of Container Storage configuration file                                                       |
 
 
 ## Files
@@ -134,7 +135,7 @@ Print the version
 	registries.conf is the configuration file which specifies which container registries should be consulted when completing image names which do not include a registry or domain portion.
 
 ## SEE ALSO
-podman(1), mounts.conf(5), registries.conf(5), storage.conf(5)
+podman(1), mounts.conf(5), newuidmap(1), newgidmap(1), registries.conf(5), storage.conf(5)
 
 ## HISTORY
 December 2017, Originally compiled by Tom Sweeney <tsweeney@redhat.com>
--- a/util/util.go
+++ b/util/util.go
@@ -7,6 +7,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"path/filepath"
 	"strconv"
 	"strings"
 
@@ -330,7 +331,8 @@ func getHostIDMappings(path string) ([]specs.LinuxIDMapping, error) {
 	return mappings, nil
 }
 
-// GetHostIDMappings reads mappings for the current process from the kernel.
+// GetHostIDMappings reads mappings for the specified process (or the current
+// process if pid is "self" or an empty string) from the kernel.
 func GetHostIDMappings(pid string) ([]specs.LinuxIDMapping, []specs.LinuxIDMapping, error) {
 	if pid == "" {
 		pid = "self"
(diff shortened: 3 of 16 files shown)


nalind:
bot, retest this please.

AkihiroSuda:
Why not unshare mountns?

giuseppe:
for podman we agreed on using these directories for the storage:

- `$XDG_DATA_DIR/containers/storage` if `XDG_DATA_DIR` is specified, otherwise:
- `$HOME/.containers/storage`.

Could we use these directories for buildah as well so the two tools work well together?

https://github.com/projectatomic/libpod/blob/master/cmd/podman/libpodruntime/runtime.go#L22-L39

rhatdan:
Yes lets use the same storage.

docs/buildah-unshare.md
```diff
@@ -0,0 +1,32 @@
...
+
+buildah unshare id
+
+buildah unshare pwd
+
+buildah unshare cat /proc/self/uid\_map
+
+buildah unshare cat /proc/self/gid\_map
+
+buildah rm -fr $HOME/containers/storage /var/run/user/`id -u`/containers/storage
```
TomSweeneyRedHat: I'm not seeing unshare in this line and I'm not sure how it fits?

docs/buildah-unshare.md
```diff
@@ -0,0 +1,32 @@
+# buildah-unshare "19" "June 2018" "buildah"
+
+## NAME
```
TomSweeneyRedHat: Buildah is all about containers and images, but not a mention of either in this man page.  Can you tie in how this command would be useful to either or both in this page?

nalind:
The mount namespace is already unshared in `SetupIntermediateMountNamespace()`, which is called from inside of the user namespace.

nalind:
I see `XDG_DATA_DIRS` (note the _S_ on the end) in [the spec](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html), but not `XDG_DATA_DIR`, but if the permissions and ownerships are compatible, it makes sense to share the storage.  Changing it.

nalind:
Switched to use ($XDG_DATA_HOME with a fallback to $HOME/.local/share)"/containers/storage" and ($XDG_RUNTIME_DIR with a fallback to /var/run/user/$UID)"/run".

docs/buildah-unshare.md
nalind: Updated to note that this is mainly for troubleshooting and manual clearing of storage areas, and updated the example that deletes everything to reflect the updated locations.

util/util.go
```diff
@@ -428,3 +430,37 @@ func ParseIDMappings(uidmap, gidmap []string) ([]idtools.IDMap, []idtools.IDMap,
...
+
+// UnsharedRootPath returns a location under ($XDG_DATA_HOME/containers/storage,
+// or $HOME/.local/share/containers/storage, or
+// (the user's home directory)/.local/share/containers/storage, or an error.
+func UnsharedRootPath(homedir string) (string, error) {
+	// If $XDG_DATA_HOME is defined...
+	if envDataHome, haveDataHome := os.LookupEnv("XDG_DATA_HOME"); haveDataHome {
+		return filepath.Join(envDataHome, "containers", "storage"), nil
+	}
+	// If $XDG_DATA_DIRS is not defined, but $HOME is defined...
```
giuseppe: XDG_DATA_HOME in the comment?

util/util.go
```diff
@@ -428,3 +430,37 @@ func ParseIDMappings(uidmap, gidmap []string) ([]idtools.IDMap, []idtools.IDMap,
...
+	if envHomedir, haveHomedir := os.LookupEnv("HOME"); haveHomedir {
+		// Default to the user's $HOME/.local/share/containers/storage subdirectory.
+		return filepath.Join(envHomedir, ".local", "share", "containers", "storage"), nil
+	}
+	// If we know where our home directory is...
+	if homedir != "" {
+		// Default to the user's homedir/.local/share/containers/storage subdirectory.
+		return filepath.Join(homedir, ".local", "share", "containers", "storage"), nil
+	}
+	return "", errors.New("unable to determine a --root location: neither $XDG_DATA_DIRS nor $HOME is set")
```
giuseppe: same here in the error

util/util.go
nalind: Eek, fixing.

Conan-Kudo:
This code looks like it would not work properly in a container environment normally. Within my GitLab CI environment, we're running everything in container environments already, but neither `podman` nor `buildah` can handle the container storage lower layer being on OverlayFS. Is there some kind of config file thing we can drop into the home directory to redirect it to another location that's volume mounted in?

rhatdan:
@Conan-Kudo You need to volume mount in /var/lib/containers/storage into the container, if you want to use overlay.  overlay on overlay is not supported.

Conan-Kudo:
@rhatdan I'm aware, but unprivileged builds seem to use a path other than that. I currently volume mount `/var/lib/containers/storage` now, but it'd be nice if in unprivileged mode I could fix it to a specific path as an override.

rhatdan:
The fixes for non root usage is not in yet. Our first priority is to get buildah to work in a non-privileged root container to be able to run under openshift, with volumes mounted in.  But @giuseppe and @nalind Are also working on getting good "non-root" support in.

docs/buildah.md
```diff
@@ -46,23 +46,23 @@ be used, as the default behavior of using the system-wide configuration
 
 **--root** **value**
 
-Storage root dir (default: "/var/lib/containers/storage")
+Storage root dir (default: "/var/lib/containers/storage" for UID 0, "$HOME/containers/storage" for other users)
```
giuseppe: this should be updated to reflect the last changes, `$HOME/containers/storage` is not used anymore

giuseppe:
@rhatdan the re-exec to userNS code is slightly different than what we are doing in podman, but it LGTM.

docs/buildah.md
nalind: Whoops, fixing.

rhatdan:
@giuseppe @nalind Can you guys work on making them more compatable?

nalind:
Corrected a bug: entries in /etc/subgid are keyed by user name, not by name of the user's primary group.

nalind:
Rebased.

nalind:
Added a hunk to force ownership of `/etc/hosts` and `/etc/resolv.conf` to 0:0.  When we `run`, letting the ID mapping logic map from the host-level owner to the right UID/GID pair in the container doesn't work because our "host" namespace is one where the owner isn't mapped, so it isn't 0:0 to begin with.

rhatdan:
LGTM

docs/buildah-unshare.md
TomSweeneyRedHat: ty!

rhatdan:
@rh-atomic-bot r+

q384566678:
>Our first priority is to get buildah to work in a non-privileged root container to be able to run under openshift, with volumes mounted in.

I have a question to ask. Currently, `buildah` can be executed by non-privileged users, but only in  `vfs` format. So what we need to do is support other types of drivers? Or there's something else to do?

giuseppe:
@q384566678 an unprivileged user cannot mount other kind of file systems.
In Linux 4.18 it will be possible to mount a FUSE file system in an user namespace, so we can take advantage of it and emulate overlay in user space: https://github.com/containers/fuse-overlayfs.

There is already support for it plugged into containers/storage and it is enabled with `--storage-opt  overlay.mount_program=/usr/bin/fuse-overlayfs`.  If you'd like you can already play with it on Rawhide where Linux 4.18 is available

q384566678:
@giuseppe Thanks for your explanation.
Now there is a [development plan](https://github.com/projectatomic/buildah/blame/master/developmentplan.md#L7) in the community that is related to this. I thought that the type of driver could be improved to complete this plan, but it doesn't seem to be.
I also want to advance this work, but I have no direction at present. Could you give me some advice? Thank you very much!

giuseppe:
I am not sure of that, @rhatdan or @nalind might know better how much it is important for Buildah to be explicitly added to the development plan; IMHO anything that can make the rootless case better can be considered part of 'Integration with User Namespace'

rhatdan:
I am not really sure of the question here.  We are working to integrate buildah into OpenSHift workflow for building containers with as little privs as possible.  Actually I think it can run the gamut, from fully privileged writing the the containers/storage on the host that the actual container that is running buildah is running, to fully locked down where the container storage is only used with the the cotnainer file system.

Obviously there are speed versus security concerns with this.  Any thing that we can take advantage of to make the more secure options faster, we want to take advantage of, whether that is reflink on top of VFS, or fuse-overlayfs or advances in the kernel overlayfs to be allowed by non priv users.

q384566678:
@rhatdan @nalind After thinking for a while, I still don't quite understand. After the patch is merged, buildah can be performed by non-root users, so has the following plan been completed?

> Integration into Kubernetes and potentially other tools. The biggest requirement for this is to be able run Buildah within a standard linux container without SYS_ADMIN privileges. This would allow Buildah to run non-privileged containers inside of Kubernetes, so you could distribute your container workloads.

I'm not sure what more needs to be done about the plan.

My understanding of this plan is that you can perform buildah with non-root users. Now it's done.

I think the plan should do more than that, but at present I have no idea to carry out the plan. Could you give me some advice? TYVM!

Conan-Kudo:
> I think the plan should do more than that, but at present I have no idea to carry out the plan. Could you give me some advice? TYVM!

Well, right now, if a container build runs `sudo`, it fails. So something needs to happen to fix that. This case triggers because a container may be layered on top of another one that makes it so it doesn't run as `root` by default, but has passwordless sudo configured.

rhatdan:
@Conan-Kudo Could you give me a Dockerfile where you see this failure?

Conan-Kudo:
@rhatdan Sure.

Base container Dockerfile:

```Dockerfile
FROM ubuntu:16.04

RUN apt-get update && apt-get -y install sudo 

RUN useradd -G sudo docker && echo "docker:docker" | chpasswd && \
    chown -R docker:docker /home/docker && \
    echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers

USER docker
CMD /bin/bash

```

Layered container:

```Dockerfile
FROM 1604dockeruser # this is the base container above

RUN sudo apt-get -y install tree

```

rhatdan:
@Conan-Kudo Then you run this with a non privileged buildah correct?

Conan-Kudo:
@rhatdan It fails in both cases (privileged and unprivileged).

rhatdan:
Sorry one more thing could you open a new issue for this, just include the dockerfile(s) above and explain that it fails.  I will look into it or have @TomSweeneyRedHat look at this tomorrow,

Conan-Kudo:
@rhatdan Done. https://github.com/projectatomic/buildah/issues/996

PR #825, opened 2018-06-26
umount: add all option to umount all mounted containers

q384566678:
Umount all of the currently mounted containers.
After change:

```
➜  buildah git:(umount-all) sudo ./buildah umount --all
43e674866c4cb46dfc64988b95a8247d6ae5fd17f0b15a46f16a81cb62a55353
14d5fbebf8eadf8056f72271cff1e86147fe40990331c5bf76e251f3cdd25b81
07d02e23e58503a33ec1886c1bfd6e9831fe05c5ac2ea8c6d9e2a939cb40b9fc
```

### DIFF
--- a/cmd/buildah/umount.go
+++ b/cmd/buildah/umount.go
@@ -5,11 +5,16 @@ import (
 	"os"
 
 	"github.com/pkg/errors"
-	"github.com/projectatomic/buildah/util"
 	"github.com/urfave/cli"
 )
 
 var (
+	umountFlags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "all, a",
+			Usage: "umount all of the currently mounted containers",
+		},
+	}
 	umountCommand = cli.Command{
 		Name:           "umount",
 		Aliases:        []string{"unmount"},
@@ -17,15 +22,20 @@ var (
 		Description:    "Unmounts the root file system on the specified working containers",
 		Action:         umountCmd,
 		ArgsUsage:      "CONTAINER-NAME-OR-ID [...]",
+		Flags:          umountFlags,
 		SkipArgReorder: true,
 	}
 )
 
 func umountCmd(c *cli.Context) error {
+	umountAll := c.Bool("all")
 	umountContainerErrStr := "error unmounting container"
 	args := c.Args()
-	if len(args) == 0 {
+	if len(args) == 0 && !umountAll {
 		return errors.Errorf("at least one container ID must be specified")
+	}
+	if len(args) > 0 && umountAll {
+		return errors.Errorf("when using the --all switch, you may not pass any container IDs")
 	}
 
 	store, err := getStore(c)
@@ -34,20 +44,50 @@ func umountCmd(c *cli.Context) error {
 	}
 
 	var lastError error
-	for _, name := range args {
+	if len(args) > 0 {
+		for _, name := range args {
+			builder, err := openBuilder(getContext(), store, name)
+			if err != nil {
+				if lastError != nil {
+					fmt.Fprintln(os.Stderr, lastError)
+				}
+				lastError = errors.Wrapf(err, "%s %s", umountContainerErrStr, name)
+				continue
+			}
+			if builder.MountPoint == "" {
+				continue
+			}
 
-		builder, err := openBuilder(getContext(), store, name)
+			id := builder.ContainerID
+			if err = builder.Unmount(); err != nil {
+				if lastError != nil {
+					fmt.Fprintln(os.Stderr, lastError)
+				}
+				lastError = errors.Wrapf(err, "%s %q", umountContainerErrStr, builder.Container)
+				continue
+			}
+			fmt.Printf("%s\n", id)
+		}
+	} else {
+		builders, err := openBuilders(store)
 		if err != nil {
-			lastError = util.WriteError(os.Stderr, errors.Wrapf(err, "%s %s", umountContainerErrStr, name), lastError)
-			continue
+			return errors.Wrapf(err, "error reading build Containers")
 		}
+		for _, builder := range builders {
+			if builder.MountPoint == "" {
+				continue
+			}
 
-		id := builder.ContainerID
-		if err = builder.Unmount(); err != nil {
-			lastError = util.WriteError(os.Stderr, errors.Wrapf(err, "%s %q", umountContainerErrStr, builder.Container), lastError)
-			continue
+			id := builder.ContainerID
+			if err = builder.Unmount(); err != nil {
+				if lastError != nil {
+					fmt.Fprintln(os.Stderr, lastError)
+				}
+				lastError = errors.Wrapf(err, "%s %q", umountContainerErrStr, builder.Container)
+				continue
+			}
+			fmt.Printf("%s\n", id)
 		}
-		fmt.Printf("%s\n", id)
 	}
 	return lastError
 }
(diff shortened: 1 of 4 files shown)


cmd/buildah/umount.go
```diff
@@ -10,44 +10,73 @@ import (
...
 		return err
 	}
 
 	var lastError error
-	for _, name := range args {
+	if len(args) > 0 {
+		for _, name := range args {
+			builder, err := openBuilder(getContext(), store, name)
+			if err != nil {
+				lastError = util.WriteError(os.Stderr, errors.Wrapf(err, "%s %s", umountContainerErrStr, name), lastError)
```
rhatdan: You need to check if lastError is already set and output it before replacing the lastError with the new one.
```
if lastError != nil {
           logrus.Error(lastError)
}
```

cmd/buildah/umount.go
```diff
@@ -10,44 +10,73 @@ import (
...
+				continue
+			}
 
-		builder, err := openBuilder(getContext(), store, name)
+			id := builder.ContainerID
+			if err = builder.Unmount(); err != nil {
+				lastError = util.WriteError(os.Stderr, errors.Wrapf(err, "%s %q", umountContainerErrStr, builder.Container), lastError)
```
rhatdan: Print lastError here also if set.

cmd/buildah/umount.go
```diff
@@ -10,44 +10,73 @@ import (
...
+				continue
+			}
 
-		id := builder.ContainerID
-		if err = builder.Unmount(); err != nil {
-			lastError = util.WriteError(os.Stderr, errors.Wrapf(err, "%s %q", umountContainerErrStr, builder.Container), lastError)
-			continue
+			id := builder.ContainerID
+			if err = builder.Unmount(); err != nil {
+				lastError = util.WriteError(os.Stderr, errors.Wrapf(err, "%s %q", umountContainerErrStr, builder.Container), lastError)
```
rhatdan: Print  lastError here also if it is set.

docs/buildah-umount.md
```diff
@@ -9,11 +9,23 @@ buildah\-umount - Unmount the root file system on the specified working containe
...
+
 ## EXAMPLE
 
 buildah umount containerID
 
 buildah umount containerID1 containerID2 containerID3
 
+buildah umount --all
+
+43e674866c4cb46dfc64988b95a8247d6ae5fd17f0b15a46f16a81cb62a55353
```
TomSweeneyRedHat: We don't have example output from the other commands here and I'm not sure it adds much to have it  here.  I'd drop it, but if we are going to keep it then we ought to add it to the other examples too.

cmd/buildah/umount.go
q384566678: updated, thanks.

docs/buildah-umount.md
q384566678: updated, I'm ok whether add or not.

rhatdan:
LGTM
@TomSweeneyRedHat @umohnani8 @pixdrift PTAL

cmd/buildah/umount.go
```diff
@@ -6,48 +6,87 @@ import (
...
 func umountCmd(c *cli.Context) error {
+	umountAll := c.Bool("all")
 	umountContainerErrStr := "error unmounting container"
 	args := c.Args()
-	if len(args) == 0 {
+	if len(args) == 0 && !umountAll {
 		return errors.Errorf("at least one container ID must be specified")
 	}
+	if len(args) > 0 && umountAll {
+		return errors.Errorf("when using the --all switch, you may not pass anycontainers IDs")
```
TomSweeneyRedHat: apologies for not spotting this sooner if this was in prior, "anycontainers" to "any container"

docs/buildah-umount.md
```diff
@@ -9,6 +9,11 @@ buildah\-umount - Unmount the root file system on the specified working containe
 ## DESCRIPTION
 Unmounts the root file system on the specified working containers.
 
+## OPTIONS
+**--all, -a**
+
+All of the currently mounted containers will be unmounted.
+
```
TomSweeneyRedHat: I'd add the example 'buildah umount --all' here, just don't include the output from the command too.

umohnani8:
LGTM. Can we get the same thing in podman umount as well.

TomSweeneyRedHat:
@umohnani8 sounds like a good idea, I'll whip up a PR.

rhatdan:
Don't we already have it?

TomSweeneyRedHat:
Nope, podman only umounts the one single container handed in.  Need to work through a compile issue, then the code should be ready.

rhatdan:
I was thinking of this https://github.com/projectatomic/buildah/pull/825

cmd/buildah/umount.go
```diff
@@ -6,48 +6,87 @@ import (
...
+					logrus.Error(lastError)
+				}
+				lastError = util.WriteError(os.Stderr, errors.Wrapf(err, "%s %q", umountContainerErrStr, builder.Container), lastError)
+				continue
+			}
+			fmt.Printf("%s\n", id)
 		}
-		fmt.Printf("%s\n", id)
 	}
 	return lastError
```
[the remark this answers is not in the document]
rhatdan: No the lastError handling is correct, you want main to print it out if it is an error, and exit with value != 0

@q384566678 You need to check if the container is mounted before unmountings, which should remove the bogus errors.  There is still a chance at a race condition, but I think just checking if it is mounted should be enough.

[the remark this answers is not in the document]
q384566678: > one less ERRO, which makes me wonder if we should or should not do the logrus if statement at all.

if you run `buildah` with debug option, you'll see all of the ERROR.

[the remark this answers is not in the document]
q384566678: updated, PTAL.

cmd/buildah/umount.go
```diff
@@ -6,48 +6,90 @@ import (
...
 	var lastError error
-	for _, name := range args {
+	if len(args) > 0 {
+		for _, name := range args {
+			builder, err := openBuilder(getContext(), store, name)
+			if err != nil {
+				if lastError != nil {
+					logrus.Error(lastError)
+				}
+				lastError = util.WriteError(os.Stderr, errors.Wrapf(err, "%s %s", umountContainerErrStr, name), lastError)
```
rhatdan: This looks wrong, I believe the util.WriteError will print the error to the screen so you will end up with lastErrror reported twice.  I thin the code should probbaly look like.

if lastError != nil {
		util.WriteError(os.Stderr, lastError)
}
lastError = errors.Wrapf(err, "%s %s", umountContainerErrStr, name)

q384566678: Done.

cmd/buildah/umount.go
```diff
@@ -2,52 +2,91 @@ package main
...
+			}
 
-		id := builder.ContainerID
-		if err = builder.Unmount(); err != nil {
-			lastError = util.WriteError(os.Stderr, errors.Wrapf(err, "%s %q", umountContainerErrStr, builder.Container), lastError)
-			continue
+			id := builder.ContainerID
+			if err = builder.Unmount(); err != nil {
+				if lastError != nil {
+					fmt.Println(lastError)
```
rhatdan: These lines should still be 
fmt.Fprintln(os.Stderr, lastError)
Otherwise the error message goes to stdout, we want it at stderr.

q384566678:
updated, thanks.

rhatdan:
@rh-atomic-bot r+

PR #828, opened 2018-06-26
vendor in latest runtime-tools

baude:
The latest runtime-tools is aware of other OS's than Linux.  Libpod needs the newer
version to compile on darwin.  Unfortunately, the API for generator.New() changed
and requires a string representation of the OS; furthermore, it also returns a
a generator and an error so code had to be adjusted for this too.


run.go
```diff
@@ -177,8 +177,11 @@ func DefaultNamespaceOptions() NamespaceOptions {
 		{Name: string(specs.UserNamespace), Host: true},
 		{Name: string(specs.UTSNamespace), Host: true},
 	}
-	g := generate.New()
-	spec := g.Spec()
+	g, err := generate.New("linux")
+	if err != nil {
+		return options, err
+	}
+	spec := g.Config
```
wking: nit: This is fine, but you may want to consider renaming `spec` -> `config` as well.  I don't know why the runtime-spec authors decided to use `Spec` for the structure backing `config.json`, but that JSON is a configuration based on a specification, not a specification in its own right (also why the file is not `specification.json` ;).

run.go
```diff
@@ -177,8 +177,11 @@ func DefaultNamespaceOptions() NamespaceOptions {
 		{Name: string(specs.UserNamespace), Host: true},
 		{Name: string(specs.UTSNamespace), Host: true},
 	}
-	g := generate.New()
-	spec := g.Spec()
+	g, err := generate.New("linux")
```
wking: I'm fine with hard-coding this here, because any cross-platform buildah implementation would have to switch OSes at a higher level, and only call `DefaultNamespaceOptions` on Linux (since that function makes no sense on other OSes).

baude: Right, thats the conclusion I came to as well

TomSweeneyRedHat:
LGTM and happy green test buttons

rhatdan:
@rh-atomic-bot r+

rhatdan:
Seems homu is hosed.  Merging manually.

PR #832, opened 2018-06-28
Fix buildah bud --layers

umohnani8:
When building with layers, the last step wouldn't get implemented
if a cache already existed. This fix checks if every step in the dockerfile
is the same, and if it is it just creates a copy of the existing image
with the new name passed in by the user. The images will have the same
IDs and the new one will just be another tag of the original image.
This is what docker build does as well.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"time"
 
+	cp "github.com/containers/image/copy"
 	is "github.com/containers/image/storage"
 	"github.com/containers/image/transports"
 	"github.com/containers/image/transports/alltransports"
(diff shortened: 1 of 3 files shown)


umohnani8:
@rhatdan PTAL

rhatdan:
LGTM, 
Why did you have to change the tests to buildah rmi -a -f?

umohnani8:
@rhatdan because some of the tests run build again without changing the Dockerfile, so we end up with two images with the same ID. To delete those we need to do rmi -f for buildah. Podman handles it better, it just untags it, I guess we don't have that in buildah yet.

imagebuildah/build.go
```diff
@@ -862,6 +930,20 @@ func historyMatches(children []*parser.Node, history []v1.History) bool {
 	return true
 }
 
+// getPolicyContext sets up, initializes and returns a new context for the specified policy
```
TomSweeneyRedHat: This feels like a function that should live in a utility class, not here in build.  Besides that, is there a way to get to SignaturePolicyPath from here rather than pulling again?

TomSweeneyRedHat:
In general LGTM, just curious about the signaturepolicy stuff.

imagebuildah/build.go
umohnani8: Moved to util. Not sure what you mean by not pulling the SignaturePolicyPath, I don't see where that is happening?

TomSweeneyRedHat: I'm not sure about your question about my question!  ;^)   What I was trying to say is we rarely (never?) need to get the signature policy after the first time we get it during the init process. I think that value is set in the SignaturePolicyPath variable,  I'm just not sure if you can get to that variable from this code or not.  If you can, it would see preferable over doing the new function that you put together.  

Then again a ton of reviews today and I maybe off base.

rhatdan:
@umohnani8 Can you open an issue on the untagging.

@rh-atomic-bot r+

PR #834, opened 2018-06-28
runConfigureNetwork(): remove an unused parameter

nalind:
Remove the configureNetwork parameter from runConfigureNetwork(), which was only called if the value was true, and which runConfigureNetwork() itself never used.


rhatdan:
LGTM
@umohnani8 @TomSweeneyRedHat @pixdrift PTAL

umohnani8:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #836, opened 2018-06-28
Add --isolation chroot

nalind:
Add an `IsolationChroot` that trades flexibility and isolation for being able to do what it does in a host environment that's already been isolated to the point where we're not allowed to set up some of that isolation (using the host's cgroup, ipc, network, and pid namespaces, which generally means we don't have privileges for configuring cgroups, and we can't control things like networking and the size of /dev/shm), producing a result that leans more toward chroot(1) than runc(1) does.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,10 @@ go:
     - 1.10.x
     - tip
 
+env:
+    - BUILDAH_ISOLATION=oci
+    - BUILDAH_ISOLATION=chroot
+
 matrix:
   # If the latest unstable development version of go fails, that's OK.
   allow_failures:
@@ -21,7 +25,7 @@ services:
 before_install:
     - sudo add-apt-repository -y ppa:duggan/bats
     - sudo apt-get update
-    - sudo apt-get -qq install bats btrfs-tools git libapparmor-dev libdevmapper-dev libglib2.0-dev libgpgme11-dev libselinux1-dev
+    - sudo apt-get -qq install bats btrfs-tools git libapparmor-dev libc-dev libdevmapper-dev libglib2.0-dev libgpgme11-dev libselinux1-dev
     - sudo apt-get -qq remove libseccomp2
     - sudo apt-get -qq update
     - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-ce
--- a/chroot/run.go
+++ b/chroot/run.go
@@ -1,0 +1,1211 @@
+// +build linux
+
+package chroot
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strconv"
+	"strings"
+	"sync"
+	"syscall"
+	"unsafe"
+
+	"github.com/containers/storage/pkg/ioutils"
+	"github.com/containers/storage/pkg/mount"
+	"github.com/containers/storage/pkg/reexec"
+	"github.com/opencontainers/runc/libcontainer/apparmor"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
+	"github.com/projectatomic/buildah/bind"
+	"github.com/projectatomic/buildah/unshare"
+	"github.com/projectatomic/buildah/util"
+	"github.com/sirupsen/logrus"
+	"github.com/syndtr/gocapability/capability"
+	"golang.org/x/crypto/ssh/terminal"
+	"golang.org/x/sys/unix"
+)
+
+const (
+	// runUsingChrootCommand is a command we use as a key for reexec
+	runUsingChrootCommand = "buildah-chroot-runtime"
+	// runUsingChrootExec is a command we use as a key for reexec
+	runUsingChrootExecCommand = "buildah-chroot-exec"
+)
+
+func init() {
+	reexec.Register(runUsingChrootCommand, runUsingChrootMain)
+	reexec.Register(runUsingChrootExecCommand, runUsingChrootExecMain)
+}
+
+type runUsingChrootSubprocOptions struct {
+	Spec        *specs.Spec
+	BundlePath  string
+	UIDMappings []syscall.SysProcIDMap
+	GIDMappings []syscall.SysProcIDMap
+}
+
+type runUsingChrootExecSubprocOptions struct {
+	Spec       *specs.Spec
+	BundlePath string
+}
+
+// RunUsingChroot runs a chrooted process, using some of the settings from the
+// passed-in spec, and using the specified bundlePath to hold temporary files,
+// directories, and mountpoints.
+func RunUsingChroot(spec *specs.Spec, bundlePath string, stdin io.Reader, stdout, stderr io.Writer) (err error) {
+	var confwg sync.WaitGroup
+
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	// Write the runtime configuration, mainly for debugging.
+	specbytes, err := json.Marshal(spec)
+	if err != nil {
+		return err
+	}
+	if err = ioutils.AtomicWriteFile(filepath.Join(bundlePath, "config.json"), specbytes, 0600); err != nil {
+		return errors.Wrapf(err, "error storing runtime configuration")
+	}
+
+	// Run the grandparent subprocess in a user namespace that reuses the mappings that we have.
+	uidmap, gidmap, err := util.GetHostIDMappings("")
+	if err != nil {
+		return err
+	}
+	for i := range uidmap {
+		uidmap[i].HostID = uidmap[i].ContainerID
+	}
+	for i := range gidmap {
+		gidmap[i].HostID = gidmap[i].ContainerID
+	}
+
+	// Default to using stdin/stdout/stderr if we weren't passed objects to use.
+	if stdin == nil {
+		stdin = os.Stdin
+	}
+	if stdout == nil {
+		stdout = os.Stdout
+	}
+	if stderr == nil {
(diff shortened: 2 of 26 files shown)


rhatdan:
@giuseppe @mheon @wking @umohnani8 @TomSweeneyRedHat PTAL

.travis.yml
```diff
@@ -7,6 +7,10 @@ go:
     - 1.10.x
     - tip
 
+env:
+    - BUILDAH_ISOLATION=oci
+    - BUILDAH_ISOLATION=chroot
```
rhatdan: Does this cause two tests to run, or just override the first ENV with the second?

chroot/run.go
```diff
@@ -0,0 +1,1208 @@
+// +build linux
```
rhatdan: Can you add a run_nonlinux.go

pkg/parse/parse.go
```diff
@@ -535,6 +535,8 @@ func defaultIsolation() buildah.Isolation {
 	isolation := os.Getenv("BUILDAH_ISOLATION")
 	if strings.HasPrefix(strings.ToLower(isolation), "oci") {
 		return buildah.IsolationOCI
+	} else if strings.HasPrefix(strings.ToLower(isolation), "chroot") {
+		return buildah.IsolationChroot
```
rhatdan: Should we throw an error if environment is set but not set to chroot or oci>

.travis.yml
nalind: Each test configuration we had becomes two, now.

pkg/parse/parse.go
nalind: Yes, we should.  Fixing.

giuseppe:
from a quick glance it looks good, but it is a lot of new code :-)  I'll review more in details.  There were recently a lot of improvements in runc to run from an isolated environment.
What are the issues you have encountered?  What is the reference environment that Buildah will run from?  I see we are setting up an intermediate mount namespace, so it seems that at least the default seccomp policy for a Docker/Podman container was tweaked.

rhatdan:
@giuseppe what is blocked by seccomp?  Our version allows mount/umount syscalls?

giuseppe:
That is nice, I didn't see that changed with docker 1.13.  It was blocked on docker 1.12 (same for `unshare`).

That is helpful, so root in the container can re-gain capabilities in a new userNS and still be able to map all the uids/gids without requiring `newuid`/`newgidmap`

rhatdan:
I think that is valid to research but for now, I think we move forward with chroot, so we can get buildah running under OpenShift sooner rather then later.

TomSweeneyRedHat:
fwiw @nalind, unhappy tests

giuseppe:
I am only worried that it will be a lot of new stuff to maintain as it is almost a new runtime (although we could replace the new code with the OCI variant once it is fully working).  I've done a quick test with dropping all the namespaces from the OCI configuration and runc requires only a small change to run the container (drop a control that `/proc` is an invalid destination where chroot is used).

giuseppe:
said so, I don't want to hold this PR if this can get unprivileged builds faster into OpenShift.

If you'd like to consider the alternative way I suggested, these are the changes I added to `run.go` (it is an old patch I've used for playing with the FUSE filesystem+dropping namespaces) to get buildah running in a container.

For simplicity I've hardcoded `rootless` to `true`.

In addition to the patch I needed:

- this change in runc: https://github.com/opencontainers/runc/pull/1832
- and I've manually created an userNS with the full range `0 0 4294967295` so that root could gain again all the capabilities in the namespace before running buildah.  This should be also possible to add to the new re-exec code.

```
diff --git a/run.go b/run.go
index 97d247fd..50b5b2f8 100644
--- a/run.go
+++ b/run.go
@@ -797,6 +797,7 @@ func setupNamespaces(g *generate.Generator, namespaceOptions NamespaceOptions, i
 
 // Run runs the specified command in the container's root filesystem.
 func (b *Builder) Run(command []string, options RunOptions) error {
+	rootless := true
 	var user specs.User
 	p, err := ioutil.TempDir("", Package)
 	if err != nil {
@@ -817,6 +818,44 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	gp := generate.New()
 	g := &gp
 
+	if rootless {
+		g.RemoveMount("/dev/pts")
+		devPts := specs.Mount{
+			Destination: "/dev/pts",
+			Type:        "devpts",
+			Source:      "devpts",
+			Options:     []string{"nosuid", "noexec", "newinstance", "ptmxmode=0666", "mode=0620"},
+		}
+		g.AddMount(devPts)
+
+		g.RemoveMount("/sys")
+		sysMnt := specs.Mount{
+			Destination: "/sys",
+			Type:        "bind",
+			Source:      "/sys",
+			Options:     []string{"nosuid", "noexec", "nodev", "ro", "rbind"},
+		}
+		g.AddMount(sysMnt)
+
+		g.RemoveMount("/proc")
+		procMnt := specs.Mount{
+			Destination: "/proc",
+			Type:        "bind",
+			Source:      "/proc",
+			Options:     []string{"nosuid", "noexec", "nodev", "rw", "rbind"},
+		}
+		g.AddMount(procMnt)
+
+		g.RemoveMount("/dev/mqueue")
+		mqueueMnt := specs.Mount{
+			Destination: "/dev/mqueue",
+			Type:        "bind",
+			Source:      "/dev/mqueue",
+			Options:     []string{"nosuid", "noexec", "nodev", "rw", "rbind"},
+		}
+		g.AddMount(mqueueMnt)
+	}
+
 	for _, envSpec := range append(b.Env(), options.Env...) {
 		env := strings.SplitN(envSpec, "=", 2)
 		if len(env) > 1 {
@@ -854,8 +893,10 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 		}
 	}()
 
-	setupMaskedPaths(g)
-	setupReadOnlyPaths(g)
+	if !rootless {
+		setupMaskedPaths(g)
+		setupReadOnlyPaths(g)
+	}
 
 	g.SetRootPath(mountPoint)
 
@@ -875,7 +916,7 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 		return err
 	}
 
-	if configureUTS {
+	if !rootless && configureUTS {
 		if options.Hostname != "" {
 			g.SetHostname(options.Hostname)
 		} else if b.Hostname() != "" {
@@ -900,9 +941,16 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 
 	// Now grab the spec from the generator.  Set the generator to nil so that future contributors
 	// will quickly be able to tell that they're supposed to be modifying the spec directly from here.
+	if rootless {
+		g.ClearLinuxNamespaces()
+	}
 	spec := g.Spec()
 	g = nil
 
+	if rootless {
+		spec.Linux.Resources = nil
+	}
+
 	// Remove capabilities if not running as root
 	if user.UID != 0 {
 		var caplist []string
@@ -1075,6 +1123,8 @@ func runUsingRuntimeMain() {
 }
 
 func runUsingRuntime(options RunOptions, configureNetwork bool, configureNetworks []string, spec *specs.Spec, rootPath, bundlePath, containerName string) (wstatus unix.WaitStatus, err error) {
+	rootless := true
+
 	// Lock the caller to a single OS-level thread.
 	runtime.LockOSThread()
 
@@ -1170,6 +1220,9 @@ func runUsingRuntime(options RunOptions, configureNetwork bool, configureNetwork
 	// Build the commands that we'll execute.
 	pidFile := filepath.Join(bundlePath, "pid")
 	args := append(append(append(options.Args, "create", "--bundle", bundlePath, "--pid-file", pidFile), moreCreateArgs...), containerName)
+	if rootless {
+		args = append(append(append(options.Args, "--rootless=true", "create", "--bundle", bundlePath, "--pid-file", pidFile, "--no-new-keyring"), moreCreateArgs...), containerName)
+	}
 	create := exec.Command(runtime, args...)
 	create.Dir = bundlePath
 	stdin, stdout, stderr := getCreateStdio()
```

nalind:
bot, retest this please

rhatdan:
@nalind Please comment on @giuseppe question.

nalind:
I don't know what @giuseppe is asking.  Right now we expect to be able to set ID mappings when launching a process (we've mapped the user's own ID to 0 for ourselves so that we can set things up, but we don't map it into the namespace we're setting up for the process that we're starting), and the rootless mode changes don't do that.

giuseppe:
I was asking if we can use `runc --rootless=true`, the flag was added recenly that will skip configuring cgroups and tweak the OCI configuration to simplify this PR as I think the runtime should be able to support this setup.  If we don't want to map the current UID inside of the process userNS we could use runc to set up a new userNS as a subset of the one we created?

I opened a PR for runc for the issue I've encountered when using chroot instead of a mount namespace, but the PR is waiting for review.  Anyway, this is something we could come back at later if it is going to cause delays.

nalind:
It sounds like `runc --rootless=true` is closer to being usable for this case (some of the mountpoint changes made above are things we already have to do when namespaces are set to "host"), but it's not there yet.  I don't see `--isolation oci` going away as the default, so that's fine.

giuseppe:
> but it's not there yet

what more would we miss from `runc --rootless=true`?

rhatdan:
Well we could always drop support for "chroot" if/when runc gets enough support, over time.

rhatdan:
@nalind Needs a rebase.

rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot retry

nalind:
@rh-atomic-bot retry

rhatdan:
@rh-atomic-bot r+

nalind:
@rh-atomic-bot retry

PR #838, opened 2018-06-29
Bump to v1.2

TomSweeneyRedHat:

First time using release.sh script, we'll see what Travis CI thinks.

### DIFF
--- a/contrib/rpm/buildah.spec
+++ b/contrib/rpm/buildah.spec
@@ -26,7 +26,7 @@
 
 Name:           buildah
 # Bump version in buildah.go too
-Version:        1.2
+Version:        1.3
 Release:        2.git%{shortcommit}%{?dist}
 Summary:        A command line tool used to creating OCI Images
 License:        ASL 2.0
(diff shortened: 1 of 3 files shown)


contrib/rpm/buildah.spec
```diff
@@ -92,6 +92,85 @@ make DESTDIR=%{buildroot} PREFIX=%{_prefix} install install.completions
...
+- images: fix usage
+- fix make clean error
+- Change 'registries' to 'container registries' in man
+- add commit test
+- Add(): learn to record hashes of what we add
+- Minor update to buildah config documentation for entrypoint
+- Bump to v1.2-dev
+- Add registries.conf link to a few man pages
+
 * Sat Jun 9 2018 Dan Walsh <dwalsh@redhat.com> 1.2-dev-1
```
pixdrift: I think this line needs to be dropped from the changelog.

TomSweeneyRedHat: Why does it need to be dropped?  This new release is 1.2-1 and 1.3-dev-1

pixdrift: Specifically this line:
`* Sat Jun 9 2018 Dan Walsh <dwalsh@redhat.com> 1.2-dev-1`

It has no changelog entries and the rpm spec delinter probably won't like that.

contrib/rpm/buildah.spec
```diff
@@ -26,7 +26,7 @@
 
 Name:           buildah
 # Bump version in buildah.go too
-Version:        1.2
+Version:        1.3
 Release:        2.git%{shortcommit}%{?dist}
```
pixdrift: This was also bumped to 2.git by nalind after the shadow-utils change in the Requires. I think it should be bumped back to 1.git with the new version number.

rhatdan:
Replaced by https://github.com/projectatomic/buildah/pull/862

PR #839, opened 2018-06-29
docs: Follow man-pages(7) suggestions for SYNOPSIS

wking:
[`man-pages(7)`][1] has:

> For commands, this shows the syntax of the command and its arguments (including options); boldface is used for as-is text and italics are used to indicate replaceable arguments. Brackets (`[]`) surround optional arguments, vertical bars (`|`) separate choices, and ellipses (`...`) can be repeated.

I've adjusted our SYNOPSIS entries to match that formatting, and generally tried to make them more consistent with the precedent set by the man-pages project.  Outside of the SYNOPSIS entry, I prefer using `backticks` for literals, although in some places I've left the **bolding** to keep things visually similar to a nearby SYNOPSIS entry.

I've also simplified a few placeholders, e.g. "containerID" -> "container", because I didn't think the additional bit was providing much additional context.  If there is ambiguity about the representation, it should be addressed in the DESCRIPTION instead of with an "ID" or "Name" suffix.

I've also proposed very similar changes for libpod in projectatomic/libpod#1027.

[1]: http://man7.org/linux/man-pages/man7/man-pages.7.html

### DIFF
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -4,7 +4,11 @@
 buildah\-bud - Build an image using instructions from Dockerfiles.
 
 ## SYNOPSIS
-**buildah** **bud | build-using-dockerfile** [*options* [...]] **context**
+**buildah build-using-dockerfile** [*options*] *context*
+
+**buildah bud** [*options*] *context*
+
+**bud** is an alias for **build-using-dockerfile**.
 
 ## DESCRIPTION
 Builds an image using instructions from one or more Dockerfiles and a specified
--- a/docs/buildah-mount.md
+++ b/docs/buildah-mount.md
@@ -4,9 +4,7 @@
 buildah\-mount - Mount a working container's root filesystem.
 
 ## SYNOPSIS
-**buildah** **mount**
-
-**buildah** **mount** [**containerID** [...]]
+**buildah mount** [*container* ...]
 
 ## DESCRIPTION
 Mounts the specified container's root file system in a location which can be
--- a/docs/buildah-add.md
+++ b/docs/buildah-add.md
@@ -4,7 +4,7 @@
 buildah\-add - Add the contents of a file, URL, or a directory to a container.
 
 ## SYNOPSIS
-**buildah** **add** **containerID** **SRC** [[...] **DEST**]
+**buildah add** [*options*] *container* *src* [[*src* ...] *dest*]
 
 ## DESCRIPTION
 Adds the contents of a file, URL, or a directory to a container's working
--- a/docs/buildah-commit.md
+++ b/docs/buildah-commit.md
@@ -4,12 +4,12 @@
 buildah\-commit - Create an image from a working container.
 
 ## SYNOPSIS
-**buildah** **commit** [*options* [...]] **containerID** **imageName**
+**buildah commit** [*options*] *container* *image*
 
 ## DESCRIPTION
 Writes a new image using the specified container's read-write layer and if it
-is based on an image, the layers of that image.  If *imageName* does not begin
-with a registry name component, *localhost* will be added to the name.
+is based on an image, the layers of that image.  If *image* does not begin
+with a registry name component, `localhost` will be added to the name.
 
 ## RETURN VALUE
 The image ID of the image that was created.  On error, 1 is returned and errno is returned.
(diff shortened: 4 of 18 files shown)


TomSweeneyRedHat:
bot, add author to whitelist

TomSweeneyRedHat:
The changes in general LGTM, but I'd rather containerID and imageID insted of container and image.  What to others think?

wking:
> I'd rather containerID and imageID insted of container and image. 

Some of the previous examples were using `imageName`, not `imageID`.  But are there other ways to identifiy containers and images (e.g. container names or image hashes)?  If so, do the commands that used to have `containerID`, etc. only support the ID form?

docs/buildah-bud.md
```diff
@@ -4,7 +4,9 @@
 buildah\-bud - Build an image using instructions from Dockerfiles.
 
 ## SYNOPSIS
-**buildah** **bud | build-using-dockerfile** [*options* [...]] **context**
+**buildah build-using-dockerfile** [*options*] *context*
```
rhatdan: I would rather list both.
**buildah bud** [*options*] *context*

docs/buildah-mount.md
```diff
@@ -4,9 +4,7 @@
 buildah\-mount - Mount a working container's root filesystem.
 
 ## SYNOPSIS
-**buildah** **mount**
-
-**buildah** **mount** [**containerID** [...]]
+**buildah** **mount** [**container** ...]
```
rhatdan: Remove middle ** **

docs/buildah-bud.md
wking: > I would rather list both.
> **buildah bud** [*options*] *context*

How do you feel about [the belt-and-suspenders approach][1]?

[1]: https://github.com/projectatomic/buildah/blob/4a2b66a55ba513480c1dd7b35055c9f724f6d310/docs/buildah-bud.md#synopsis

docs/buildah-mount.md
wking: > Remove middle ** **

Done, here and in `buildah-from.md`, with 1a6e426 -> 4a2b66a.

TomSweeneyRedHat:
@wking buildah command should support either the id or the name unless there's a bug running around.    I've preferred documenting only the ID in the synopsis as it's cleaner to document.  Also we've never had issues looking up by id and over the course of development the name hasn't always been hip.  That's my $.02 anywho.

wking:
> buildah command should support either the id or the name unless there's a bug running around.

In that case, I'd prefer the more general `container` to the overly-specific `containerID`.

> Also we've never had issues looking up by id and over the course of development the name hasn't always been hip.

It may be too early in the buildah process to be documenting bugs in man pages instead of fixing them.  But if a command is intended to support both container names and IDs, but names may be buggy, that seems like fodder for [a `BUGS` section][1], not a reason to leave the name support undocumented.

[1]: http://man7.org/linux/man-pages/man7/man-pages.7.html

TomSweeneyRedHat:
My personal problem with leaving it as "container" is when you look at the output of "buildah containers" which column in a container?  "containerID" is very evident and quick to find.  As  a bonus, it's concise and has always worked.  Similar thinking with images.  My $.02.

wking:
> "containerID" is very evident and quick to find.

But [the `CONTAINER NAME` column][0] would work too, right?  There doesn't seem to be a column whose name contains "container" that would not work.

To add another option to the `container` vs. `containerID` debate, [`buildah add --help` will show][1] `CONTAINER-NAME-OR-ID`.  Using CAPS there makes sense without the man page's italics availabe, but whatever we settle on here should probably be otherwise consistent.  In order to keep this PR managable, I suggest we punt `--help` consistency to follow-up-work, but we should consider the list-all-the-possibilities approach for the man pages.

My preferences are (in decreasing order):

1. `container` (short & generic)
2. `containerNameOrID` (generic)
3. `containerID` (too specific, except where names aren't acceptable)

[0]: https://github.com/projectatomic/buildah/blob/v1.1/cmd/buildah/containers_test.go#L95
[1]: https://github.com/projectatomic/buildah/blob/v1.1/cmd/buildah/addcopy.go#L26

rhatdan:
Ok I am going to merge this since the cleanup looks good, and you guys can continue the debate of what to call this in a different PR>

rhatdan:
@rh-atomic-bot r+

PR #840, opened 2018-06-29
allow buildah cross compile for a darwin target

baude:
the goal here is to allow ourselves to cross compile buildah for a darwin
target.  we are doing this to eventually protect from regressions that could
creep into buildah so we don't dig ourselves a deeper hole.

the simplified and non-variable approach to the make darwin was done with
intent to keep this simple until we can exploit things a little more.

once this PR merges, i will create a CI test that will test for regressions
on a make darwin.  we should also be doing a gofmt with a darwin target so the
!linux|darwin tagged files are also checked for completeness.  initially the
test can be optional for passing with the long-term idea that it be made
a firm requirement at the buildah maintainers behest.


rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #842, opened 2018-07-01
add optional CI test fo darwin

baude:
run gofmt and make for darwin to prevent regressions from new code for a potential
darwin build.


baude:
ok this is ready for review

.papr.yml
```diff
@@ -47,3 +47,39 @@ tests:
...
+    - git
+    - glib2-devel
+    - gnupg
+    - golang
+    - libassuan-devel
+    - make
+    - ostree-devel
+    - skopeo-containers
+
+required: false
```
rhatdan: Why is this optional, Don't we want to block PRS that break this?

baude: I was thinking optional initially ... if others want it required, easy change to make here or later.

rhatdan:
Well I still think it should be required, but maybe we wait until we get buildah 1.2 out.
@rh-atomic-bot

PR #845, opened 2018-07-03
Add a --loglevel option to build-with-dockerfile

nalind:
Add a --loglevel option to build-with-dockerfile and use it to add a value from -2 to 3 to the default logrus LogError logging level.  This makes --loglevel 3 equivalent in effect to the global --debug option.

OpenShift defaults to appending --loglevel 0 (or --loglevel $BUILD_LOGLEVEL, if the variable is set) to the command for the builder container, and this lets us accept that.

### DIFF
--- a/cmd/buildah/main.go
+++ b/cmd/buildah/main.go
@@ -76,7 +76,7 @@ func main() {
 		},
 	}
 	app.Before = func(c *cli.Context) error {
-		logrus.SetLevel(logrus.ErrorLevel)
+		logrus.SetLevel(logrus.ErrorLevel + logrus.Level(c.Int("loglevel")))
 		if c.GlobalBool("debug") {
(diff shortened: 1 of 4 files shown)


rhatdan:
bot, retest this please
@nalind Should these be hidden options?  Do you see users ever using this in the real world?

TomSweeneyRedHat:
@nalind tests aren't hip

nalind:
@rhatdan I don't expect the option to be used often, but I could say the same for a good number of options we don't hide, including a few no-op options, so I don't see much point in making this a hidden one.

rhatdan:
bot, retest this please

TomSweeneyRedHat:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #854, opened 2018-07-06
Fix handling of --registries-conf

nalind:
Instead of ignoring the global --registries-conf option and using only $REGISTRIES_CONFIG_PATH, use it for the option default.

### DIFF
--- a/pull.go
+++ b/pull.go
@@ -167,14 +167,14 @@ func pullImage(ctx context.Context, store storage.Store, imageName string, optio
 	}()
 
 	logrus.Debugf("copying %q to %q", spec, destName)
-	err = cp.Image(ctx, policyContext, destRef, srcRef, getCopyOptions(options.ReportWriter, options.SystemContext, nil, ""))
+	err = cp.Image(ctx, policyContext, destRef, srcRef, getCopyOptions(options.ReportWriter, sc, nil, ""))
 	if err == nil {
 		return destRef, nil
 	}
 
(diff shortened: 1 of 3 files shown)


pull.go
```diff
@@ -167,14 +167,14 @@ func pullImage(ctx context.Context, store storage.Store, imageName string, optio
 	}()
 
 	logrus.Debugf("copying %q to %q", spec, destName)
-	err = cp.Image(ctx, policyContext, destRef, srcRef, getCopyOptions(options.ReportWriter, options.SystemContext, nil, ""))
+	err = cp.Image(ctx, policyContext, destRef, srcRef, getCopyOptions(options.ReportWriter, sc, nil, ""))
```
TomSweeneyRedHat: In new.go:newBuilder, we tease out the systemContext from options and then pass it down to here without using it prior that I can see.  Could we just drop systemContext as a param to several functions and just pull the value out from options here in this code?  No a change necessary for this PR, but maybe a touchup issue for later.

nalind: We'd have to call `getSystemContext()` to override the default signature policy path with the configured one for each case, but that's not a deal-breaker.

TomSweeneyRedHat: yeah, and signaturepolicy should be in the the options too.

TomSweeneyRedHat:
bot, retest this please.

nalind:
Basically everything needs #855 at this time.

rhatdan:
@rh-atomic-bot r+

PR #855, opened 2018-07-06
image filter test: use kubernetes/pause as a "since"

nalind:
Use `kubernetes/pause`, which is not expected to be updated regularly, as a reference image for the "since" filter, instead of `alpine`, which is updated regularly, and which therefore can be newer than `busybox`.  (I was tempted to use references with digests to make a permanent fix, but that would also implicitly hard code an architecture.)

### DIFF
--- a/tests/images.bats
+++ b/tests/images.bats
@@ -29,10 +29,11 @@ load helpers
 }
 
 @test "images filter test" {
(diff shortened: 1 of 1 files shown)


nalind:
@rh-atomic-bot r+

PR #856, opened 2018-07-06
Don't pass a nil error to errors.Wrapf()

nalind:
Don't pass a `nil` error value to `errors.Wrapf()` when we want to report an error, since it's documented as returning `nil` for that case.


TomSweeneyRedHat:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #857, opened 2018-07-09
Add rename command

zhouhao3:
Add rename command to rename a local container.

### DIFF
--- a/cmd/buildah/rename.go
+++ b/cmd/buildah/rename.go
@@ -1,0 +1,52 @@
+package main
+
+import (
+	"github.com/pkg/errors"
+	"github.com/projectatomic/buildah"
+	"github.com/urfave/cli"
+)
+
+var (
+	renameDescription = "Rename a local container"
+	renameCommand     = cli.Command{
+		Name:           "rename",
+		Usage:          "Rename a container",
+		Description:    renameDescription,
+		Action:         renameCmd,
+		ArgsUsage:      "CONTAINER-NAME-OR-ID CONTAINER-NAME",
+		SkipArgReorder: true,
+	}
+)
+
+func renameCmd(c *cli.Context) error {
+	var builder *buildah.Builder
+
+	args := c.Args()
+	if len(args) != 2 {
+		return errors.Errorf("container and it's new name must be specified'")
+	}
+
+	name := args[0]
+	newName := args[1]
+
+	store, err := getStore(c)
+	if err != nil {
+		return err
+	}
+
+	builder, err = openBuilder(getContext(), store, name)
+	if err != nil {
+		return errors.Wrapf(err, "error reading build container %q", name)
+	}
+
+	oldName := builder.Container
+	if oldName == newName {
+		return errors.Errorf("renaming a container with the same name as its current name")
+	}
+
+	err = store.SetNames(builder.ContainerID, []string{newName})
+	if err != nil {
+		return errors.Wrapf(err, "error renaming container %q to the name %q", oldName, newName)
+	}
+	return nil
+}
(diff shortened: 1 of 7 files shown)


nalind:
This is probably going to make things confusing, because historically we've based our copy of the container's name on the name that we initially set using the storage library, and we're probably not being careful about one or the other changing.

Any chance we could make `Rename()` a method of the builder that updates the `Container` field and also calls the builder's `store.SetNames()` method to update the names that the storage library has on file?  We could also consider resetting `Container` based on the value that the storage library has as a part of opening the builder, which would keep us from being surprised when someone else changed the storage library's set of names for the container.  What do you think?

q384566678:
```
➜  buildah git:(add-rename) sudo buildah containers
828e0b099441     *     3fd9065eaf02 docker.io/library/alpine:latest  alpine-working-container
➜  buildah git:(add-rename) sudo ./buildah rename alpine-working-container test-container
➜  buildah git:(add-rename) sudo ./buildah containers                                    
828e0b099441     *     3fd9065eaf02 docker.io/library/alpine:latest  alpine-working-container
➜  buildah git:(add-rename) sudo ./buildah run alpine-working-container bash
error reading build container "alpine-working-container": error reading build container: container not known
➜  buildah git:(add-rename) sudo ./buildah containers --all                 
828e0b099441     *     3fd9065eaf02 docker.io/library/alpine:latest  test-container
```

Updated.  After a change, Because the structure of the call is different, only the --all option displays the name of the container that has been properly modified.

q384566678:

cmd/buildah/rename.go
```diff
@@ -0,0 +1,55 @@
...
+		ArgsUsage:      "CONTAINER-NAME-OR-ID CONTAINER-NAME",
+		SkipArgReorder: true,
+	}
+)
+
+func renameCmd(c *cli.Context) error {
+	var builder *buildah.Builder
+
+	args := c.Args()
+	if len(args) < 2 {
```
rhatdan: Nit: Could simplify this code to say
if len(args) == 2 {
          return errors.Errorf("container and it's new name must be specified'")
}

cmd/buildah/rename.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	builder, err = openBuilder(getContext(), store, name)
+	if err != nil {
+		return errors.Wrapf(err, "error reading build container %q", name)
+	}
+
+	oldName := builder.Container
+	if oldName == newName {
+		return errors.Errorf("Renaming a container with the same name as its current name")
```
rhatdan: Errors should be lower case.

rhatdan:
You also need to update buildah.md man page.

rhatdan:
Also add command to README.md

cmd/buildah/rename.go
TomSweeneyRedHat: s/b '!=2` no?  Or are the cold meds kicking in for me?

cmd/buildah/rename.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	err = store.SetNames(builder.ContainerID, []string{newName})
+	if err != nil {
+		return errors.Wrapf(err, "error set name %q to container %q", newName, oldName)
```
TomSweeneyRedHat: suggest:  "error renaming container %q to the name %q", oldName, newName

q384566678:
Rebased.

rhatdan:
@rh-atomic-bot r+
@q384566678 I am not sure this can be done for podman.  @mheon What do you think?

mheon:
@rhatdan Not easily, Podman container configs are as immutable as I could get away with, so it would require major changes to allow name changes.

rhatdan:
Ok I guess we could put this at the bottom of the barrell of issues.

PR #871, opened 2018-07-19
Fix rmi to remove intermediate images associated with an image

umohnani8:
buildah rmi now removes cached images associated with an image
when the image is removed only if the cached image is not required
by another image.
If an image is a parent and has a name then untag, otherwise return
an error stating the image can't be remoed as it has dependent
child images.

### DIFF
--- a/cmd/buildah/common.go
+++ b/cmd/buildah/common.go
@@ -154,3 +154,65 @@ func defaultFormat() string {
 	}
 	return "oci"
 }
+
+// imageIsParent goes through the layers in the store and checks if i.TopLayer is
+// the parent of any other layer in store. Double check that image with that
+// layer exists as well.
+func imageIsParent(store storage.Store, topLayer string) (bool, error) {
+	children, err := getChildren(store, topLayer)
+	if err != nil {
+		return false, err
+	}
+	return len(children) > 0, nil
+}
+
+// getParent returns the image ID of the parent. Return nil if a parent is not found.
+func getParent(store storage.Store, topLayer string) (*storage.Image, error) {
+	images, err := store.Images()
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to retrieve images from store")
+	}
+	layer, err := store.Layer(topLayer)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to retrieve layers from store")
+	}
+	for _, img := range images {
+		if img.TopLayer == layer.Parent {
+			return &img, nil
+		}
+	}
+	return nil, nil
+}
+
+// getChildren returns a list of the imageIDs that depend on the image
+func getChildren(store storage.Store, topLayer string) ([]string, error) {
+	var children []string
+	images, err := store.Images()
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to retrieve images from store")
+	}
+	layers, err := store.Layers()
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to retrieve layers from store")
+	}
+
+	for _, layer := range layers {
+		if layer.Parent == topLayer {
+			if imageID := getImageOfTopLayer(images, layer.ID); len(imageID) > 0 {
+				children = append(children, imageID...)
+			}
+		}
+	}
+	return children, nil
+}
+
+// getImageOfTopLayer returns the image ID where layer is the top layer of the image
+func getImageOfTopLayer(images []storage.Image, layer string) []string {
+	var matches []string
+	for _, img := range images {
+		if img.TopLayer == layer {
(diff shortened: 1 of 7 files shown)


umohnani8:

tests/rmi.bats
```diff
@@ -87,6 +87,7 @@ load helpers
   run buildah --debug=false images -q
   [ $(wc -l <<< "$output") -eq 2 ]
 
+  buildah rmi --all --force
```
rhatdan: Why run this twice?

rhatdan:
Who found this issue?

tests/rmi.bats
umohnani8: Now that we have the check for `isParent`, when we do `buildah rmi -a`, it is possible for it to hit an image that has children so in the first rmi -a run the children will be removed, then we can run it again to remove the remaining images.
This check is necessary so that users don't delete images that have dependent children.

umohnani8:
@rhatdan no one, it is just something we needed to implement to account for the cached images. Without this, the intermediate images will just be lying around in the store after you delete the final image. Did the same in podman, https://github.com/projectatomic/libpod/pull/1055.

rhatdan:
Looks like you have to break out the functions in rmi.go into multiple functions.

cmd/buildah/common.go
```diff
@@ -154,3 +154,65 @@ func defaultFormat() string {
...
+// getImageOfTopLayer returns the image ID where layer is the top layer of the image
+func getImageOfTopLayer(images []storage.Image, layer string) []string {
+	var matches []string
+	for _, img := range images {
+		if img.TopLayer == layer {
+			matches = append(matches, img.ID)
+		}
+	}
+	return matches
+}
```
TomSweeneyRedHat: Iterating through all of the images in each of these functions worries me from a performance perspective, but I'm not sure how to avoid it.  It would be nice if we could just set up a map in memory somewhere, but I don't think we've a place to store it.

cmd/buildah/rmi.go
```diff
@@ -159,7 +159,19 @@ func rmiCmd(c *cli.Context) error {
 				}
 			}
 
-			if len(image.Names) > 0 {
+			isParent, err := imageIsParent(store, image.TopLayer)
+			if err != nil {
+				return err
+			}
+			if len(image.Names) > 0 || isParent {
```
TomSweeneyRedHat: should this 'if' be moved to line 170?  I don't think 170 can ever be hit if isParent is true now.

cmd/buildah/rmi.go
```diff
@@ -222,9 +234,35 @@ func untagImage(imgArg string, image *storage.Image, store storage.Store) (strin
 }
 
 func removeImage(image *storage.Image, store storage.Store) (string, error) {
+	parent, err := getParent(store, image.TopLayer)
+	if err != nil {
+		return "", err
```
TomSweeneyRedHat: will err be sufficient to let the user know why the removal didn't happen?  Should we have an errros.Wrapf here?

TomSweeneyRedHat: same comment for other returns below.

tests/rmi.bats
TomSweeneyRedHat: IDK, I think if a user does "rmi --all -force", they're not expecting to have to run it again to remove all, they just want them all gone in the first pass....

umohnani8: I can add a check after to see if there are any images left, if so call removeImage() on them again. Does that sound better?

rhatdan: Yes lets try that, but fail after two times.  Don't want to introduce an infinite loop

cmd/buildah/common.go
umohnani8: Yeah, there really is no way around this. Not sure if it is worth setting up a map in memory.

cmd/buildah/rmi.go
umohnani8: Changed the condition so that we don't try to delete a parent image that has been untagged and an image that still has a name associated with it after untagging it.

cmd/buildah/rmi.go
umohnani8: wrapped the errors.

cmd/buildah/common.go
TomSweeneyRedHat: yeah given everything I think you're right unfortunately.

umohnani8:
@nalind @rhatdan no matter how much I increase the timeout, the tests still times out. I increased it to 1500s and it still happened. Am I doing something wrong?

rhatdan:
@rh-atomic-bot r+

PR #876, opened 2018-07-20
tests/run.bats(volumes): use :z when SELinux is enabled

nalind:
When testing volumes as part of checking that "run" is working correctly, use the ":z" flag if we detect SELinux support to avoid false negatives.

### DIFF
--- a/tests/run.bats
+++ b/tests/run.bats
@@ -323,15 +323,21 @@ load helpers
 	if ! which runc ; then
 		skip
 	fi
+	zflag=
+	if which selinuxenabled > /dev/null 2> /dev/null ; then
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
Changes LGTM, but papr is having an odd lint issue...

```
WARNING: deadline exceeded by linter unconvert on cmd/buildah (try increasing --deadline)
make: *** [Makefile:54: validate] Error 2
make: Leaving directory '/root/gopath/src/github.com/projectatomic/buildah'
### EXITED WITH CODE 2 AFTER 1096s
```

nalind:
@rh-atomic-bot r+

PR #880, opened 2018-07-24
Switch to Recommends container-selinux

rhatdan:
This way we can install buildah into a container without pulling in
SELinux policy packages.

### DIFF
--- a/contrib/rpm/buildah.spec
+++ b/contrib/rpm/buildah.spec
@@ -47,8 +47,13 @@ BuildRequires:  glib2-devel
 BuildRequires:  ostree-devel
 BuildRequires:  make
 Requires:       runc >= 1.0.0-6
+%if 0%{?rhel} == 7
 Requires:       container-selinux
 Requires:       skopeo-containers
+%else
+Recommends:     container-selinux
+Requires:       containers-common
+%endif
 Requires:       shadow-utils
 Provides:       %{repo} = %{version}-%{release}
 


contrib/rpm/buildah.spec
```diff
@@ -47,8 +47,8 @@ BuildRequires:  glib2-devel
 BuildRequires:  ostree-devel
 BuildRequires:  make
 Requires:       runc >= 1.0.0-6
-Requires:       container-selinux
-Requires:       skopeo-containers
+Recommends:     container-selinux
+Requires:       containers-common
```
TomSweeneyRedHat: :+1:

rhatdan:
Yes I guess so. (Although I hate dirtying up the spec file, that is just an example and not something we actually ship on RHEL7 or Centos7

pixdrift:
Even if it is just an example, i'd take 'Requiring' a few extra packages over breaking EL7 builds if modifying the .spec isn't an option.

EL7 is where a lot of the existing Docker users exist (RHEL7, CentOS7, Scientific Linux 7, Oracle Linux 7). I'd argue they are the enterprise customers to convince that Buildah is a viable alternative by making access to it as easy as possible.

Another alternative is to include a Fedora and EL7 spec file, rather than including the logic in the same file.

rhatdan:
@pixdrift @lsm5 Will my check work on CENTOS7?

pixdrift:
@rhatdan, yep
```
# cat /etc/redhat-release 
CentOS Linux release 7.5.1804 (Core) 
# grep rhel /etc/rpm/macros.dist
%rhel 7
```
What's the story with `containers-common`? Should that stay as `skopeo-containers` on EL7 until the `containers-common` package is built for EL7?

lsm5:
@fkluknav when you do the next skopeo build for rhel 7.5, could you please rename the skopeo-containers package to containers-common ? Once that change is done, please also rebuild that source for CentOS Virt SIG.

pixdrift:
Is this a good time to ask if it can be changed so the naming is consistent? ie. selinux package is singular _container_ and the common package is plural _container**s**_? 😟 😄

lsm5:
ha, `container-` SGTM. @rhatdan wdyt?

lsm5:
@rhatdan you mean rename to singular `container-` ? Though it's shipped already, we can have the rename done for future builds and have it Provides: containers-common. sound good?

rhatdan:
@lsm5 Fine.
https://paste.fedoraproject.org/paste/Nc4YGhPpjbP5eD0XSOnrQA
Here is the spec changes I was suggesting ^^

rhatdan:
I actually think we should leave it plural since we have
github.com/containers/*
containers/image
containers/storage
containers/skopeo

So the real problem is that we ship container-selinux, perhaps we should change this name.

lsm5:
@rhatdan seems like your diff is on an older spec file. The latest spec file already has `containers-common` change along with your seccomp.json changes. PTAL and let me know..

lsm5:
don't mean to drag this on :\, but there's also `container-storage-setup`. And in other tools, there's `systemd-container` (if that matters), and we might need to include `container-diff` in Fedora.

rhatdan:
@pixdrift @lsm5 @TomSweeneyRedHat PTAL

contrib/rpm/buildah.spec
```diff
@@ -47,8 +47,13 @@ BuildRequires:  glib2-devel
 BuildRequires:  ostree-devel
 BuildRequires:  make
 Requires:       runc >= 1.0.0-6
+%if 0%{?rhel} == 7
 Requires:       container-selinux
```
TomSweeneyRedHat: should this be"Recommends:" too?

rhatdan: Recommends is not supported on RHEL7.

TomSweeneyRedHat: kk, thx for the 411

TomSweeneyRedHat:
LGTM

rhatdan:
@rh-atomic-bot r+

Issue #882, opened 2018-07-24
bud fails when run as non-root - error copying layers and metadata


TomSweeneyRedHat:
**Description**

buildah bud fails when run as non-root user, but completes as root

**Steps to reproduce the issue:**
1.  $ cat Dockerfile
FROM fedora:28
RUN dnf -y update
2.  buildah bud -t myimg .

**Describe the results you received:**
```
$ buildah bud --net=host -t myimg .
STEP 1: FROM fedora:28
STEP 2: RUN dnf -y update
Fedora 28 - x86_64 - Updates                    2.5 MB/s |  20 MB     00:08    
Fedora 28 - x86_64                              892 kB/s |  60 MB     01:09    
Last metadata expiration check: 0:00:14 ago on Tue Jul 24 18:55:08 2018.
Dependencies resolved.
================================================================================
 Package                     Arch   Version                       Repository
                                                                           Size
================================================================================
Upgrading:
 acl                         x86_64 2.2.53-1.fc28                 updates  80 k
 {removed lots of happy update output for brevity}                                             
Complete!
STEP 3: COMMIT containers-storage:[vfs@/home/tom/.local/share/containers/storage+/run/user/1000/run]localhost/myimg:latest
Getting image source signatures
Skipping fetch of repeat blob sha256:891e1e4ef82ad02a4ea1f030831f942d722c7694c4db64ca3239c8163b811c58
Copying blob sha256:a6f9a4c6f5ae68a3090ddb2e73728d2a06778ba5b8fd316c90aac5697af9cc2a
 198.86 MiB / 198.86 MiB [==================================================] 5s
Copying config sha256:f1aaa659f928caf201d375680de469b568c6e8593c4f76cc4de2b187591e2860
 1.56 KiB / 1.56 KiB [======================================================] 0s
Writing manifest to image destination
Storing signatures
error copying layers and metadata: Error committing the finished image: error adding layer with blob "sha256:a6f9a4c6f5ae68a3090ddb2e73728d2a06778ba5b8fd316c90aac5697af9cc2a": ApplyLayer exit status 1 stdout:  stderr: lchown /etc/tcsd.conf: invalid argument
ERRO[0237] exit status 1   
```

**Describe the results you expected:**
A clean build.

**Output of `buildah version`:**

```
# buildah version
Version:         1.2
Go Version:      go1.10.3
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.3.1
[...]

PR #883, opened 2018-07-24
Remove default dev/pts

rhatdan:
This allows us to run buildah without being root

### DIFF
--- a/run.go
+++ b/run.go
@@ -938,6 +938,17 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 
 	b.configureEnvironment(g, options)
 
+	if os.Getuid() != 0 {
+		g.RemoveMount("/dev/pts")
+		devPts := specs.Mount{
+			Destination: "/dev/pts",
+			Type:        "devpts",
+			Source:      "devpts",
+			Options:     []string{"nosuid", "noexec", "newinstance", "ptmxmode=0666", "mode=0620"},
+		}
+		g.AddMount(devPts)
+	}
+
 	if b.CommonBuildOpts == nil {
 		return errors.Errorf("Invalid format on container you must recreate the container")
 	}


rhatdan:
```
cat ~/Dockerfile 
FROM fedora
RUN echo hello
```
```
id 
uid=3267(dwalsh) gid=3267(dwalsh) groups=3267(dwalsh),10(wheel) context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
```
```
 ./buildah bud --net=host ~/
STEP 1: FROM fedora
Skipping fetch of repeat blob sha256:e71c36a80ba912dd7a5a9f2f2d6136c148afa19bc7d024bd616b74a0bc7a2774
Copying config sha256:cc510acfcd701a409014118d5f417f0022520802a26c650866b8a9594d75f3a7
 2.29 KiB / 2.29 KiB [======================================================] 0s
Storing signatures
STEP 2: RUN echo hello
hello
STEP 3: COMMIT containers-storage:[vfs@/home/dwalsh/.local/share/containers/storage+/run/user/3267/run]@94f057525b74621db1b8b002166be617a73ee11ee77062aaaaaecb3264544264
Skipping fetch of repeat blob sha256:891e1e4ef82ad02a4ea1f030831f942d722c7694c4db64ca3239c8163b811c58
Copying config sha256:2f5099516d2ad3e32b69027e94aefb3fc7cb9872c6cbb6776b69216150984647
 1.56 KiB / 1.56 KiB [======================================================] 0s
Storing signatures
--> 94f057525b74621db1b8b002166be617a73ee11ee77062aaaaaecb3264544264
```

rhatdan:
@umohnani8 @TomSweeneyRedHat @nalind @giuseppe PTAL

nalind:
This (essentially dropping "gid=5" from the mount options) shouldn't be necessary if GID 5 is mapped into the container.

rhatdan:
It would be nice to force --net=host when running as non root, until @giuseppe figures out a way to setup network namespaces in rootless mode.  But running buildah this simply as non root is very cool.

run.go
```diff
@@ -938,6 +938,17 @@ func (b *Builder) Run(command []string, options RunOptions) error {
...
+		g.RemoveMount("/dev/pts")
+		devPts := specs.Mount{
+			Destination: "/dev/pts",
+			Type:        "devpts",
+			Source:      "devpts",
+			Options:     []string{"nosuid", "noexec", "newinstance", "ptmxmode=0666", "mode=0620"},
+		}
+		g.AddMount(devPts)
+	}
+
```
TomSweeneyRedHat: LGTM.
It doesn't fix #882 though does it?

giuseppe:
> This (essentially dropping "gid=5" from the mount options) shouldn't be necessary if GID 5 is mapped into the container.

There is an issue in runc where it still complains if it detects when running in an user namespace (rootless mode) and there are not mappings in the OCI file.  This is a workaround for runc

run.go
giuseppe: It should fix: https://github.com/projectatomic/libpod/issues/1147

rhatdan:
Ok I will merge and we can revert when runc is fixed.

rhatdan:
@rh-atomic-bot r=giuseppe

Issue #887, opened 2018-07-27
ENH: 'buildah pull' Do a pull without creating a container.


TomSweeneyRedHat:
This is a feature request/enhancement discussion. The below is a IRC chat that I'd with @pixdrift the other night. I wanted to capture the thoughts so we could discuss further here.

**Description**
```
<pixdrift> another one I also wanted to ask, was has 'buildah pull' been discussed?
<pixdrift> When re-tagging, buildah pull <image>:<tag> && buildah push <image>:<newtag> comes up a bit
<pixdrift> at the moment, the only real option is 'buildah from' which creates the temp container, so the work around has been `buildah rm $(buildah from <image>:<tag>)'
<pixdrift> where buildah pull could essentially do the same as 'buildah from' without the temp container 

```

@pixdrift please feel free to add further details or correct me if any of this is incorrect.

My first thought was to do a new param like --nocontainer for `buildah from` that causes the image to just be pulled and saved locally and not create the container.  Haven't delved into it any further than that.

rhatdan:
I would prefer buildah pull since it is much easier to discover and matches up with buildah push.

TomSweeneyRedHat:
OKD, I'll make it so.  ;^)

TomSweeneyRedHat:
Closing as #889 addresses this.

PR #889, opened 2018-07-27
Create buildah pull command

TomSweeneyRedHat:

The initial implementation of the 'buildah pull' command.  This strictly pulls the image and does not create a container from it.  

This addresses #887 
```
# buildah pull alpine
Copying blob sha256:8e3ba11ec2a2b39ab372c60c16b421536e50e5ce64a0bc81765c2e38381bcff6
 2.10 MiB / 2.10 MiB [======================================================] 0s
Copying config sha256:11cd0b38bc3ceb958ffb2f9bd70be3fb317ce7d255c8a4c3f4af30e298aa1aab
 1.48 KiB / 1.48 KiB [======================================================] 0s
Storing signatures
11cd0b38bc3ceb958ffb2f9bd70be3fb317ce7d255c8a4c3f4af30e298aa1aab


# buildah images
11cd0b38bc3c         docker.io/library/alpine:latest                          Jul 6, 2018 10:14      4.67 MB

# buildah containers
```


docs/buildah-pull.md
```diff
@@ -0,0 +1,112 @@
...
+  An image in local OSTree repository.  _/absolute/repo/path_ defaults to _/ostree/repo_.
+
+### DEPENDENCIES
+
+Buildah resolves the path to the registry to pull from by using the /etc/containers/registries.conf
+file, registries.conf(5).  If the `buildah pull` command fails with an "image not known" error,
+first verify that the registries.conf file is installed and configured appropriately.
+
+## RETURN VALUE
+The image ID of the image that was pulled.  On error, 1 is returned and errno is returned.
```
vrothberg: [...] and errno is **set**?

TomSweeneyRedHat:
bot, retest this please

TomSweeneyRedHat:
Going to close and reset.  @vrothberg will take a look at adding some time if this one fails.

cmd/buildah/pull.go
```diff
@@ -0,0 +1,121 @@
...
+		cli.StringFlag{
+			Name:  "signature-policy",
+			Usage: "`pathname` of signature policy file (not usually used)",
+		},
+		cli.BoolTFlag{
+			Name:  "tls-verify",
+			Usage: "require HTTPS and verify certificates when accessing the registry",
+		},
+	}
+	pullDescription = "Pulls an image from the specified location"
```
rhatdan: How about we change this to.

Pulls and image from the specified registry or one of the registries listed in %s if no registry was specified and stores it in 
container storage in the %s directory.

cmd/buildah/pull.go
```diff
@@ -0,0 +1,121 @@
...
+		SkipArgReorder: true,
+	}
+)
+
+func pullCmd(c *cli.Context) error {
+	args := c.Args()
+	if len(args) == 0 {
+		return errors.Errorf("an image name must be specified")
+	}
+	if len(args) > 1 {
```
rhatdan: Should we allow the specification of multiple images?

cmd/buildah/pull.go
rhatdan: podman pull says

```
Pulls an image from a registry and stores it locally.
An image can be pulled using its tag or digest. If a tag is not
specified, the image with the 'latest' tag (if it exists) is pulled.
```

docs/buildah-pull.md
rhatdan: This seems like a function message. Does errno mean anything to normal users.
On error the non zero errno is set?

docs/buildah-pull.md
```diff
@@ -0,0 +1,112 @@
...
+
+
+## Files
+
+**registries.conf** (`/etc/containers/registries.conf`)
+
+registries.conf is the configuration file which specifies which container registries should be consulted when completing image names which do not include a registry or domain portion.
+
+## SEE ALSO
+buildah(1), podman-login(1), docker-login(1), policy.json(5), registries.conf(5)
```
rhatdan: Add buildah-from(1)

new.go
```diff
@@ -248,6 +248,13 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
 		defer src.Close()
 	}
 
+	// If the pull command was used, we only pull the image,
+	// we don't create a container.
+	if options.ImageOnly {
+		fmt.Printf("%s\n", imageID)
```
rhatdan: I think this should be in the pull.go

cmd/buildah/pull.go
```diff
@@ -0,0 +1,121 @@
...
+		},
+		cli.BoolFlag{
+			Name:  "quiet, q",
+			Usage: "don't output progress information when pulling images",
+		},
+		cli.StringFlag{
+			Name:  "shm-size",
+			Usage: "size of '/dev/shm'. The format is `<number><unit>`.",
+			Value: "65536k",
+		},
```
nalind: Does this get used anywhere?

cmd/buildah/pull.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+	}
+
+	options := buildah.BuilderOptions{
+		FromImage:             args[0],
+		Transport:             transport,
+		PullPolicy:            pullPolicy,
+		SignaturePolicyPath:   signaturePolicy,
+		SystemContext:         systemContext,
+		DefaultMountsFilePath: c.GlobalString("default-mounts-file"),
```
nalind: This seems unnecessary.

cmd/buildah/pull.go
```diff
@@ -0,0 +1,121 @@
...
+		ImageOnly:             true,
+		CommonBuildOpts:       commonOpts,
```
nalind: Do these values get used anywhere?

docs/buildah-pull.md
TomSweeneyRedHat: I cut/pasted this from the 'buildah from' man page.  I'll change the last sentence in both to just  "On error 1 is returned."

docs/buildah-pull.md
TomSweeneyRedHat: Done, also added to similar to buildah-from man page.

cmd/buildah/pull.go
TomSweeneyRedHat: yes, eventually in new.go:newBuilder() via the NewBuilder() call at line 115.

TomSweeneyRedHat: Oops, I misread, you're asking about commonOpts, not 'options'.  commonOpts don't seem to be used in the image creation, I'll try removing.

cmd/buildah/pull.go
rhatdan: Should not need this flag for pull, this is used for podifying containers on the size of the tmpfs inside of them.

cmd/buildah/pull.go
TomSweeneyRedHat: Done

cmd/buildah/pull.go
TomSweeneyRedHat: I was keeping step with 'buildah from' and it looks like podman pull only does one image at a time.  Can we leave it as one for now and come back and revisit?

cmd/buildah/pull.go
TomSweeneyRedHat: Retested and I indeed don't need it , removed.  Earlier testing I still must of had some container creation bits in play as I'd run into errors at one point without it.

new.go
TomSweeneyRedHat: The reason I did this hear is this function return a builder object.  That object isn't created until several lines beyond this point.  However, I probably could create a builder object here containing only the image id and return that .  I'll give that a go.

TomSweeneyRedHat:
@vrothberg I've had about a 50/50 success rate with this PR.  I've bumped the time out as suggested form 60 to 90.  We'll see how that goes.  Thanks for the suggestion/tip.

vrothberg:
@TomSweeneyRedHat Thanks! It seems there is another timeout we need to increase (in `./tests/validate/gometalinter.sh`):
```
WARNING: deadline exceeded by linter vetshadow on cmd/buildah (try increasing --deadline)
```

Those checks are *really* resource hungry and I can feel a `make validate` on my workstation quite a bit. Maybe, we can find a way (future work) to limit the resources for those tests?

TomSweeneyRedHat:
Finally happy Green Test Buttons.  The change in ./tests/validate/gometalinter.sh seems to have done the trick, or I got lucky.....

rhatdan:
@rh-atomic-bot r+

TomSweeneyRedHat:
bot, retest this please

rhatdan:
@rh-atomic-bot retry

docs/buildah-pull.md
```diff
@@ -0,0 +1,112 @@
+# buildah-pull "1" "July 2018" "buildah"
+
+## NAME
+buildah\-pull - Creates a new working container, either from scratch or using a specified image as a starting point.
```
debarshiray: Copy paste fallout? :) Submitted PR #1545 with a fix.

PR #891, opened 2018-07-30
Set the default ulimits to match Docker

rhatdan:
We want to set the maximum number of processes and number of open files
to match the Docker defaults.

nofile=1048576:1048576, nproc=1048576:1048576


rhatdan:
@nalind PTAL

nalind:
What's the rationale setting just these two defaults?  Your testing out-of-band also saw a difference in the limit for queued signals.
Can we update the docs to list the correct syntax for the argument values?  Can we add tests for this?

nalind:
When invoked by an unprivileged user, this appears to trigger an EPERM when it attempts to raise hard limits.

rhatdan:
Yes I was thinking of this causing issues. I guess we should set it to the Hardlimit for the user.

pkg/parse/parse.go
```diff
@@ -40,6 +41,22 @@ func CommonBuildOptions(c *cli.Context) (*buildah.CommonBuildOptions, error) {
...
+			return nil, err
+		}
+		nofile = fmt.Sprintf("nofile=%d:%d", rlimit.Cur, rlimit.Max)
+		if err := unix.Getrlimit(unix.RLIMIT_NPROC, &rlimit); err != nil {
+			return nil, err
+		}
+		nproc = fmt.Sprintf("nproc=%d:%d", rlimit.Cur, rlimit.Max)
+	}
+	defaultLimits := []string{nofile, nproc}
+
```
nalind: I think it'd be less code to initialize `defaultLimits` with `nofile` and `nproc`'s values before checking our privilege level, and just assign `nil` to it if it looks like setting those values would fail.

unshare/unshare.go
```diff
@@ -40,6 +40,7 @@ type Cmd struct {
 // Command creates a new Cmd which can be customized.
 func Command(args ...string) *Cmd {
 	cmd := reexec.Command(args...)
+	cmd.Env = append(os.Environ(), "__rootless__=1")
```
nalind: Setting this field deviates from the behavior of the function from the `os/exec` package that we're mimicking.  This would also be set before the caller could customize the object, including for cases where `UnshareFlags` isn't going to have `CLONE_NEWUSER` set.  It would break the current caller's current logic for setting `_BUILDAH_STARTED_IN_USERNS` in the child's environment, which it does by setting the value it in its own environment after creating the `Command` object.  Is there a reason we're not keying off of that variable?

rhatdan: Nope, I wish you told me about this a while ago.

rhatdan:
@nalind @umohnani8 @pixdrift @giuseppe @TomSweeneyRedHat PTAL

umohnani8:
LGTM

rhatdan:
@rh-atomic-bot r=umohnani8

giuseppe:
LGTM

rhatdan:
@rh-atomic-bot retry

rhatdan:
@rh-atomic-bot -rgiuseppe

rhatdan:
@rh-atomic-bot -r=giuseppe

rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot retry

PR #892, opened 2018-07-31
Fix volume cache issue with buildah bud --layers

umohnani8:
buildah bud would fail at the next step after a VOLUME command
due to buildah trying to cache the volume contents and restore it
later, but the intermediate container used at the VOLUME step didn't
exist anymore.
Changed functionality to delete successful intermediate containers at
the end of the whole build process.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -215,6 +215,7 @@ type Executor struct {
 	noCache                        bool
 	removeIntermediateCtrs         bool
 	forceRmIntermediateCtrs        bool
+	containerIDs                   []string // Stores the IDs of the successful intermediate containers used during layer build
 }
 
 // withName creates a new child executor that will be used whenever a COPY statement uses --from=NAME.
@@ -684,6 +685,7 @@ func (b *Executor) Prepare(ctx context.Context, ib *imagebuilder.Builder, node *
 	// Add the top layer of this image to b.topLayers so we can keep track of them
 	// when building with cached images.
 	b.topLayers = append(b.topLayers, builder.TopLayer)
+	logrus.Debugln("Container ID:", builder.ContainerID)
 	return nil
 }
 
@@ -811,12 +813,8 @@ func (b *Executor) Execute(ctx context.Context, ib *imagebuilder.Builder, node *
 			// it is used to create the container for the next step.
 			imgID = cacheID
 		}
-		// Delete the intermediate container if b.removeIntermediateCtrs is true.
-		if b.removeIntermediateCtrs {
-			if err := b.Delete(); err != nil {
-				return errors.Wrap(err, "error deleting intermediate container")
-			}
-		}
+		// Add container ID of successful intermediate container to b.containerIDs
+		b.containerIDs = append(b.containerIDs, b.builder.ContainerID)
 		// Prepare for the next step with imgID as the new base image.
 		if i != len(children)-1 {
 			if err := b.Prepare(ctx, ib, node, imgID); err != nil {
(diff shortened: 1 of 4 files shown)


imagebuildah/build.go
```diff
@@ -1225,3 +1238,14 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
 	stages := imagebuilder.NewStages(mainNode, b)
 	return exec.Build(ctx, stages)
 }
+
+// deleteSuccessfulIntermediateCtrs goes through the container IDs in b.containerIDs
+// and deletes the containers associated with that ID.
+func (b *Executor) deleteSuccessfulIntermediateCtrs() error {
+	for _, ctr := range b.containerIDs {
+		if err := b.store.DeleteContainer(ctr); err != nil {
+			return errors.Wrapf(err, "error deleting build container %q", ctr)
```
rhatdan: Should log here and not return, so the other layers will be deleted.

imagebuildah/build.go
```diff
@@ -1135,8 +1136,20 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) error
...
-			return err
+			lastErr = err
 		}
+
+		// Delete the successful intermediate containers if an error in the build
+		// process occurs and b.removeIntermediateCtrs is true.
+		if lastErr != nil {
+			if b.removeIntermediateCtrs {
+				if err := stageExecutor.deleteSuccessfulIntermediateCtrs(); err != nil {
+					logrus.Error(err)
```
rhatdan: I would ignore errors here and log in delete

imagebuildah/build.go
umohnani8: Fixed

imagebuildah/build.go
umohnani8: Fixed

imagebuildah/build.go
```diff
@@ -1225,3 +1236,14 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
 	stages := imagebuilder.NewStages(mainNode, b)
 	return exec.Build(ctx, stages)
 }
+
+// deleteSuccessfulIntermediateCtrs goes through the container IDs in b.containerIDs
+// and deletes the containers associated with that ID.
+func (b *Executor) deleteSuccessfulIntermediateCtrs() error {
+	for _, ctr := range b.containerIDs {
```
rhatdan: Add lasterror here, so that it can report the error.  Otherwise this always returns nil.

umohnani8: But this function doesn't need to return anything since we are just logging the errors right? The error is returned if it occurs during the execute or prepare step above.

umohnani8: nvm, I got what you meant. Fixed.

imagebuildah/build.go
```diff
@@ -1135,8 +1136,18 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) error
...
 		if err := stageExecutor.Execute(ctx, stage.Builder, stage.Node); err != nil {
-			return err
+			lastErr = err
 		}
+
+		// Delete the successful intermediate containers if an error in the build
+		// process occurs and b.removeIntermediateCtrs is true.
+		if lastErr != nil {
+			if b.removeIntermediateCtrs {
+				stageExecutor.deleteSuccessfulIntermediateCtrs()
```
TomSweeneyRedHat: do we need to retain the lastErr that's now returned from this function?

umohnani8: We don't need to, all the errors that occur by that function are logged anyways before the function returns so the user should be able to see all the errors that occurred.

rhatdan: Right we just want the fact that it errord to be recorded on the successful build.

imagebuildah/build.go
```diff
@@ -1154,7 +1165,7 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) error
 	// the removal of intermediate/build containers will be handled by the
 	// defer statement above.
 	if b.removeIntermediateCtrs && (b.layers || b.noCache) {
-		return stageExecutor.Delete()
+		return b.deleteSuccessfulIntermediateCtrs()
```
rhatdan: We might want to eat the error here and report a different error.
Something like
```
if err = b.deleteSuccessfulIntermediateCtrs(); err != nil {
           return errors.Error("Failed to cleanup intermediate Containers");
}
```

Otherwize the `LastError` will be reported twice.

umohnani8: Ooh, right. fixed. Thanks!

rhatdan:
@rh-atomic-bot r+

PR #893, opened 2018-08-01
preprocess ".in" suffixed Dockerfiles

vrothberg:
Use CPP(1) on any input Dockerfile with a ".in" suffix.  This allows to
decompose Dockerfiles and make them reusable via the `#include`
directive.  Notice that those Dockerfiles can still be used by other
tools by manually running `cpp -E` on them.

This change affects both, the `imagebuild.BuildDockerfiles(...)` API and
buildah-bud.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -1,11 +1,14 @@
 package imagebuildah
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -1186,6 +1189,8 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
 		}
 	}(dockerfiles...)
 	for _, dfile := range paths {
+		var data io.ReadCloser
+
 		if strings.HasPrefix(dfile, "http://") || strings.HasPrefix(dfile, "https://") {
 			logrus.Debugf("reading remote Dockerfile %q", dfile)
 			resp, err := http.Get(dfile)
@@ -1196,7 +1201,7 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
 				resp.Body.Close()
 				return errors.Errorf("no contents in %q", dfile)
 			}
-			dockerfiles = append(dockerfiles, resp.Body)
+			data = resp.Body
 		} else {
 			if !filepath.IsAbs(dfile) {
 				logrus.Debugf("resolving local Dockerfile %q", dfile)
@@ -1216,8 +1221,19 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
 				contents.Close()
 				return errors.Wrapf(err, "no contents in %q", dfile)
 			}
-			dockerfiles = append(dockerfiles, contents)
+			data = contents
 		}
+
+		// pre-process Dockerfiles with ".in" suffix
+		if strings.HasSuffix(dfile, ".in") {
+			pData, err := preprocessDockerfileContents(data, options.ContextDirectory)
+			if err != nil {
+				return err
+			}
+			data = *pData
+		}
+
+		dockerfiles = append(dockerfiles, data)
 	}
 	mainNode, err := imagebuilder.ParseDockerfile(dockerfiles[0])
 	if err != nil {
@@ -1250,4 +1266,55 @@ func (b *Executor) deleteSuccessfulIntermediateCtrs() error {
 		}
 	}
 	return lastErr
+}
+
+// preprocessDockerfileContents runs CPP(1) in preprocess-only mode on the input
+// dockerfile content and will use ctxDir as the base include path.
+//
+// Note: we cannot use cmd.StdoutPipe() as cmd.Wait() closes it.
+func preprocessDockerfileContents(r io.ReadCloser, ctxDir string) (rdrCloser *io.ReadCloser, err error) {
+	cppPath := "/usr/bin/cpp"
+	if _, err = os.Stat(cppPath); err != nil {
+		if os.IsNotExist(err) {
+			err = errors.Errorf("error: Dockerfile.in support requires %s to be installed", cppPath)
+		}
+		return nil, err
+	}
+
+	stdout := bytes.Buffer{}
+	stderr := bytes.Buffer{}
+
+	cmd := exec.Command(cppPath, "-E", "-iquote", ctxDir, "-")
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	pipe, err := cmd.StdinPipe()
+	if err != nil {
+		return nil, err
+	}
+
+	defer func() {
+		if err != nil {
+			pipe.Close()
+		}
+	}()
+
+	if err = cmd.Start(); err != nil {
+		return nil, err
+	}
+
+	if _, err = io.Copy(pipe, r); err != nil {
+		return nil, err
+	}
+
+	pipe.Close()
+	if err = cmd.Wait(); err != nil {
+		if stderr.Len() > 0 {
+			err = fmt.Errorf("%v: %s", err, strings.TrimSpace(stderr.String()))
+		}
+		return nil, errors.Wrapf(err, "error pre-processing Dockerfile")
+	}
+
+	rc := ioutil.NopCloser(bytes.NewReader(stdout.Bytes()))
+	return &rc, nil
 }
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -16,6 +16,8 @@ build context directory.
 
 The build context directory can be specified as the http(s) URL of an archive, git repository or Dockerfile.
 
+Dockerfiles ending with a ".in" suffix will be preprocessed via CPP(1).  This can be useful to decompose Dockerfiles into several reusable parts that can be used via CPP's **#include** directive.  Notice, a Dockerfile.in file can still be used by other tools when manually preprocessing them via `cpp -E`.
+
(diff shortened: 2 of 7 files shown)


vrothberg:
No tests yet as I am not sure how complex scenarios we should check.

rhatdan:
Also need a requires /usr/bin/cpp in the spec file, although I think we will do a recommends in the Fedora packages.

imagebuildah/build.go
```diff
@@ -1225,3 +1241,43 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
...
+
+// preprocessDockerfileContents runs CPP(1) in preprocess-only mode on the input
+// dockerfile content.
+//
+// Note: we cannot use cmd.StdoutPipe() as cmd.Wait() closes it.
+func preprocessDockerfileContents(r io.ReadCloser) (*io.ReadCloser, error) {
+	stdout := bytes.Buffer{}
+	stderr := bytes.Buffer{}
+
+	cmd := exec.Command("cpp", "-E", "-")
```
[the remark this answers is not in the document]
vrothberg: Good catch. I'll update that.

TomSweeneyRedHat:
Thanks @vrothberg, looks really good overall.

vrothberg:
@rhatdan @TomSweeneyRedHat Thanks! I updated the PR with tests and extended a bit the manpages. One thing I missed before was setting the context directory of buildah-bud as CPP's include path, so the basic requirement is that all includes are relative to build-context directory.

vrothberg:
@TomSweeneyRedHat Also here we hit the validate timeout :-(

vrothberg:
bot, retest this please

rhatdan:
Did you attach the Decomposed.in file?

vrothberg:
@rhatdan Now I did...thanks!

imagebuildah/build.go
```diff
@@ -1225,3 +1241,52 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
...
+		pipe.Close()
+		return nil, err
+	}
+	pipe.Close()
+
+	err = cmd.Wait()
+	if err != nil {
+		pipe.Close()
+		if stderr.Len() > 0 {
+			stderrString := strings.TrimSpace(stderr.String())
```
rhatdan: nit: no real reason for a variable here.

imagebuildah/build.go
```diff
@@ -1225,3 +1241,52 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
...
+		return nil, err
+	}
+
+	if _, err := io.Copy(pipe, r); err != nil {
+		pipe.Close()
+		return nil, err
+	}
+	pipe.Close()
+
+	err = cmd.Wait()
```
rhatdan: Nit
if err := cmd.Wait(); err != nil

imagebuildah/build.go
```diff
@@ -1225,3 +1241,52 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
 	stages := imagebuilder.NewStages(mainNode, b)
 	return exec.Build(ctx, stages)
 }
+
+// preprocessDockerfileContents runs CPP(1) in preprocess-only mode on the input
+// dockerfile content and will use ctxDir as the base include path.
+//
+// Note: we cannot use cmd.StdoutPipe() as cmd.Wait() closes it.
+func preprocessDockerfileContents(r io.ReadCloser, ctxDir string) (*io.ReadCloser, error) {
```
rhatdan: Add err error;

imagebuildah/build.go
```diff
@@ -1225,3 +1241,52 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
...
+
+	cmd := exec.Command(cppPath, "-E", "-iquote", ctxDir, "-")
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	pipe, err := cmd.StdinPipe()
+	if err != nil {
+		return nil, err
+	}
+	if err := cmd.Start(); err != nil {
```
rhatdan: Add
```
defer func() {
           if err != nil {
                pipe.Close()
    }
}()
```

imagebuildah/build.go
```diff
@@ -1225,3 +1241,52 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
...
+		pipe.Close()
+		return nil, err
+	}
+
+	if _, err := io.Copy(pipe, r); err != nil {
+		pipe.Close()
+		return nil, err
+	}
+	pipe.Close()
```
rhatdan: delete

imagebuildah/build.go
```diff
@@ -1225,3 +1241,52 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
...
+	if err != nil {
+		return nil, err
+	}
+	if err := cmd.Start(); err != nil {
+		pipe.Close()
+		return nil, err
+	}
+
+	if _, err := io.Copy(pipe, r); err != nil {
+		pipe.Close()
```
rhatdan: delete

imagebuildah/build.go
```diff
@@ -1225,3 +1241,52 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
...
+	cmd := exec.Command(cppPath, "-E", "-iquote", ctxDir, "-")
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	pipe, err := cmd.StdinPipe()
+	if err != nil {
+		return nil, err
+	}
+	if err := cmd.Start(); err != nil {
+		pipe.Close()
```
rhatdan: delete

TomSweeneyRedHat:
@vrothberg I made the same time change in #889 and it seems to have worked.  Will cross my fingers for your build.

vrothberg:
bot, retest this please

imagebuildah/build.go
rhatdan: This will not work correctly unless you add.

func preprocessDockerfileContents(r io.ReadCloser, ctxDir string) (*io.ReadCloser, err error) {
                                                                                                                                   ^^^

vrothberg: Mixed named return values don't compile. I'll figure out what to do with the ReadCloser.

rhatdan:
Fixes https://github.com/projectatomic/buildah/issues/851

vrothberg:
@rhatdan Updated the named return values. Let's see if we hit timeouts again in CI.

imagebuildah/build.go
```diff
@@ -1251,3 +1267,56 @@ func (b *Executor) deleteSuccessfulIntermediateCtrs() error {
...
+	}
+
+	defer func() {
+		if err != nil {
+			pipe.Close()
+		}
+	}()
+
+	if err = cmd.Start(); err != nil {
+		return
```
rhatdan: I think  you need to return err here now.

THen the err will be set. and you don't have to worry about err.  All returns should return err.

imagebuildah/build.go
```diff
@@ -1251,3 +1267,56 @@ func (b *Executor) deleteSuccessfulIntermediateCtrs() error {
...
+		if stderr.Len() > 0 {
+			err = fmt.Errorf("%v: %s", err, strings.TrimSpace(stderr.String()))
+		}
+		err = errors.Wrapf(err, "error pre-processing Dockerfile")
+		return
+	}
+
+	rc := ioutil.NopCloser(bytes.NewReader(stdout.Bytes()))
+	rdrCloser = &rc
+	return
```
rhatdan: return nil

imagebuildah/build.go
```diff
@@ -1251,3 +1267,56 @@ func (b *Executor) deleteSuccessfulIntermediateCtrs() error {
...
+		return
+	}
+
+	pipe.Close()
+	if err = cmd.Wait(); err != nil {
+		if stderr.Len() > 0 {
+			err = fmt.Errorf("%v: %s", err, strings.TrimSpace(stderr.String()))
+		}
+		err = errors.Wrapf(err, "error pre-processing Dockerfile")
+		return
```
rhatdan: return errors.Wrapf(err, "error pre-processing Dockerfile")

imagebuildah/build.go
```diff
@@ -1251,3 +1267,56 @@ func (b *Executor) deleteSuccessfulIntermediateCtrs() error {
...
+	stdout := bytes.Buffer{}
+	stderr := bytes.Buffer{}
+
+	cmd := exec.Command(cppPath, "-E", "-iquote", ctxDir, "-")
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	pipe, err := cmd.StdinPipe()
+	if err != nil {
+		return
```
rhatdan: return err

imagebuildah/build.go
```diff
@@ -1251,3 +1267,56 @@ func (b *Executor) deleteSuccessfulIntermediateCtrs() error {
...
+			pipe.Close()
+		}
+	}()
+
+	if err = cmd.Start(); err != nil {
+		return
+	}
+
+	if _, err = io.Copy(pipe, r); err != nil {
+		return
```
rhatdan: return err.

imagebuildah/build.go
vrothberg: This won't compile as we have two return values. Or am I missing something? Can't we just do a `defer pipe.Close()` and one explicit one before the `cmd.Wait()`? In the worst case, we have one redundant Close() but the code will be easier to understand (imho). Personally, I find that named return values complicate the maintainability of code.

rhatdan: Sorry 
return nil, err

The values of the returns get populated when you return before the defer gets called.

So you can do 
return nil, errors.Error("foobar") 
and when the defer gets called the err == foobar.

rhatdan:
@rh-atomic-bot r+

PR #900, opened 2018-08-02
unshare: make adjusting the OOM score optional

nalind:
The OOM score adjustment is an optional field in the runtime spec, so only try to set it in our unshare if it's set in the spec.  This _might_ help with a problem that @vbatts is running into attempting to use buildah inside of a kubernetes pod.

### DIFF
--- a/unshare/unshare.go
+++ b/unshare/unshare.go
@@ -33,7 +33,7 @@ type Cmd struct {
 	Setsid                     bool
 	Setpgrp                    bool
 	Ctty                       *os.File
-	OOMScoreAdj                int
+	OOMScoreAdj                *int
 	Hook                       func(pid int) error
 }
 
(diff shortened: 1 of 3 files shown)


unshare/unshare.go
```diff
@@ -235,16 +235,18 @@ func (c *Cmd) Start() error {
...
+		oomScoreAdj, err := os.OpenFile(fmt.Sprintf("/proc/%s/oom_score_adj", pidString), os.O_TRUNC|os.O_WRONLY, 0)
+		if err != nil {
+			fmt.Fprintf(continueWrite, "error opening oom_score_adj: %v", err)
+			return errors.Wrapf(err, "error opening /proc/%s/oom_score_adj", pidString)
+		}
+		if _, err := fmt.Fprintf(oomScoreAdj, "%d\n", c.OOMScoreAdj); err != nil {
+			fmt.Fprintf(continueWrite, "error writing \"%d\" to oom_score_adj: %v", c.OOMScoreAdj, err)
+			return errors.Wrapf(err, "error writing \"%d\" to /proc/%s/oom_score_adj", c.OOMScoreAdj, pidString)
+		}
+		defer oomScoreAdj.Close()
```
rhatdan: Isn't this in the wrong place.  Should be before the Fprintf,  Since an error from there will leave the oomScoreAdj open.

nalind: Agreed, fixing.

vbatts:
this is a decent step. I think we may eventually want to check if env variable `container` is set, and run in an altered state for being _inside_ a container. Like some errors aren't fatal because EPERMs could be expected.

rhatdan:
@rh-atomic-bot r+

vbatts:
just tested this, and it got past that EPERM (and on to the next issue...). So :+1:

rhatdan:
@nalind needs a rebase.

rhatdan:
Replaced by https://github.com/projectatomic/buildah/pull/906

Issue #850, opened 2018-07-05
buildah bud --file, -f regression / change in behaviour


pixdrift:
**Description**
From the `buildah bud` documentation, it is suggested that the `--file / -f` switch can take multiple Dockerfiles and the context is optional.

> Specifies a Dockerfile which contains instructions for building the image, either a local file or an http or https URL. If more than one Dockerfile is specified, FROM instructions will only be accepted from the first specified file.
>
> If a build context is not specified, and at least one Dockerfile is a local file, the directory in which it resides will be used as the build context.

It appears that changes to buildah have made context _required_ which in turn has made the multiple Dockerfile feature less intuitive.

One of the benefits of building using multiple Dockerfiles (or slices/shards/snippets) is code re-use, ie. Multiple Dockerfiles can share a subset of code stored in a separate file. As the option currently works, the context is required and all additional files that are provided are relative paths in the provided context.

**Steps to reproduce the issue:**
A use case I was expecting to use this feature for was
```
container1/Dockerfile
container2/Dockerfile
common/Dockerfile
```

In this scenario, I was expecting to be able to execute (from the root) so the conainerX executes first and then the contents of common is appended:
```
buildah bud -f container1/Dockerfile -f common/Dockerfile
buildah bud -f container2/Dockerfile -f common/Dockerfile
```
This results in:
```
# buildah bud -f container1/Dockerfile -f common/Dockerfile
no context directory or URL specified
```

If you provide a context directory as it is now required, it uses this for the base path of images supplied using `-f`:
```
# buildah bud -f common/Dockerfile container1
error reading "/root/multi/container1/common/Dockerfile": open /root/multi/container1/common/Dockerfile: no such file or directory
```

To make this work, it's somewhat counter intuitive. The context needs to be set, and the Dockerfile specified as the first `-f`, then a drop back to the parent directory. It does work, but it would be much cleaner without the context requirement.
```
buildah bud -f Dockerfile -f ../common/Dockerfile container1
```

**Describe the results you received:**
Errors / issues when executing `buildah bud -f` and specifying multiple images without context.

**Describe the results you expected:**
Ability to build containers from Dockerfiles using shared elements/includes as described in the buildah documentation for `buildah bud -f`.

**Output of `rpm -q buildah`:**

```
buildah-1.2-201807041005.gita59fb7a.el7.x86_64
```

**Output of `buildah version`:**

```
Version:         1.2-dev
Go Version:      go1.9.4
Image Spec:      1.0.0
[...]

PR #905, opened 2018-08-02
bud should not search context directory for Dockerfile

TomSweeneyRedHat:

When `buildah bud` received a Dockerfile value with a relative path, it would prepend the context directory onto it.  In many cases this caused a failure and confusion.  Now the code first checks to see if the Dockerfile specified with the '--file' parameter exists.  If it does exist, bud continues with the processing.  If it does not exist, then it prepends the context directory and uses that.

Testing:
===========
```
# pwd
/root/container2

# buildah  bud -f ../container1/Dockerfile container2
STEP 1: FROM alpine
STEP 2: COMMIT containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.override_kernel_check=true]@e259501bebc58b1859d702c6cc9e24fbe1543326c87240fe2b27747131335dfa
Skipping fetch of repeat blob sha256:73046094a9b835e443af1a9d736fcfc11a994107500e474d0abf399499ed280c
Copying config sha256:55edbe131d4cf07c6fed2ccb9f5a92e5237c1336fe3c6281281653e1c7ce6438
 721 B / 721 B [============================================================] 0s
Storing signatures
--> e259501bebc58b1859d702c6cc9e24fbe1543326c87240fe2b27747131335dfa

```

### DIFF
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -183,9 +183,8 @@ either a local file or an **http** or **https** URL.  If more than one
 Dockerfile is specified, *FROM* instructions will only be accepted from the
 first specified file.
 
-If a build context is not specified, and at least one Dockerfile is a
-local file, the directory in which it resides will be used as the build
-context.
+If a local file is specified as the Dockerfile and it does not exist, the
+context directory will be prepended to the local file value.
 
 If you specify `-f -`, the Dockerfile contents will be read from stdin.
 
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -1203,8 +1203,9 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
 			}
 			data = resp.Body
 		} else {
-			if !filepath.IsAbs(dfile) {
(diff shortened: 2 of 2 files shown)


TomSweeneyRedHat:
Addresses #850

rhatdan:
@TomSweeneyRedHat I think we should do both stat to see if the file is there relative, otherwise look into the context directory.

BTW We need to make sure podman build follows this pattern also.

TomSweeneyRedHat:
@rhatdan ack on the stat.  WIll make that so.  Looks like some but not all of our tests are expecting that anyway.

rhatdan:
@rh-atomic-bot r+

TomSweeneyRedHat:
@rhatdan I think I may have updated the comment at about the same time you did the rh-atomic command, does it need to be issued again?  This PR looks to be in git limbo.

rhatdan:
I think it is waiting for @nalind --rootless patch to merge, but I will ping homu again just in case
@rh-atomic-bot r+

PR #907, opened 2018-08-03
Do not use registries if user specifies transport

rhatdan:

### DIFF
--- a/cmd/buildah/from.go
+++ b/cmd/buildah/from.go
@@ -188,7 +188,7 @@ func fromCmd(c *cli.Context) error {
 		return err
 	}
 
-	transport := buildah.DefaultTransport
+	transport := ""
 	arr := strings.SplitN(args[0], ":", 2)
 	if len(arr) == 2 {
 		if _, ok := util.Transports[arr[0]]; ok {
--- a/new.go
+++ b/new.go
@@ -65,8 +65,8 @@ func reserveSELinuxLabels(store storage.Store, id string) error {
 	return nil
 }
 
-func pullAndFindImage(ctx context.Context, store storage.Store, imageName string, options BuilderOptions, sc *types.SystemContext) (*storage.Image, types.ImageReference, error) {
-	ref, err := pullImage(ctx, store, imageName, options, sc)
+func pullAndFindImage(ctx context.Context, store storage.Store, imageName string, options BuilderOptions, sc *types.SystemContext, useRegistries bool) (*storage.Image, types.ImageReference, error) {
+	ref, err := pullImage(ctx, store, imageName, options, sc, useRegistries)
 	if err != nil {
 		logrus.Debugf("error pulling image %q: %v", imageName, err)
 		return nil, nil, err
@@ -131,7 +131,7 @@ func newContainerIDMappingOptions(idmapOptions *IDMappingOptions) storage.IDMapp
 	return options
 }
 
-func resolveImage(ctx context.Context, systemContext *types.SystemContext, store storage.Store, options BuilderOptions) (types.ImageReference, *storage.Image, error) {
+func resolveImage(ctx context.Context, systemContext *types.SystemContext, store storage.Store, options BuilderOptions, useRegistries bool) (types.ImageReference, *storage.Image, error) {
 	var ref types.ImageReference
 	var img *storage.Image
 	for _, image := range util.ResolveName(options.FromImage, options.Registry, systemContext, store) {
@@ -146,7 +146,7 @@ func resolveImage(ctx context.Context, systemContext *types.SystemContext, store
 		}
 
 		if options.PullPolicy == PullAlways {
-			pulledImg, pulledReference, err := pullAndFindImage(ctx, store, image, options, systemContext)
+			pulledImg, pulledReference, err := pullAndFindImage(ctx, store, image, options, systemContext, useRegistries)
 			if err != nil {
 				logrus.Debugf("unable to pull and read image %q: %v", image, err)
 				switch err := errors.Cause(err).(type) {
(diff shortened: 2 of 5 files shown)


rhatdan:
@nalind @TomSweeneyRedHat @mtrmac @umohnani8 This is a partial solution but it at least fixes the case where user specifies a transport.  Don't use registries.  We need something similar to not use them if the user specifies a host/registryname.

cmd/buildah/from.go
```diff
@@ -188,7 +188,7 @@ func fromCmd(c *cli.Context) error {
 		return err
 	}
 
-	transport := buildah.DefaultTransport
+	transport := ""
 	arr := strings.SplitN(args[0], ":", 2)
 	if len(arr) == 2 {
```
mtrmac: What is `transport := ""` supposed to do? `newBuilder` sets `options.Transport` to `DefaulTransport` if it is `""`, so this doesn’t change behavior at all AFAICT.

If I understand the code right, `BuilderOptions.Transport` needs to exist basically because if `util.ResolveName` detects the _transport_`:`_name_ syntax, it _throws away the transport_ and returns only the _name_ part, which forces `resolveImage` to use `BuilderOptions.Transport` to reconstruct the original input. Working around that the internal implementation throws data away should not be the responsibility of `cmd/buildah` at all; that data should not be thrown away in the first place, i.e. maybe `util.ResolveName` should return a (source to pull, name to use for local storage) pairs instead of just strings.

… OK, now I see `useRegistries :=` now. That’s entirely in the wrong place, an imprecise heuristic to determine whether we used the search list when the code, at some point, _knows precisely_ whether that was the case, and this seemingly-no-op change exists only to make that heuristic work.

rhatdan:
@TomSweeneyRedHat @mtrmac @nalind @umohnani8 @pixdrift PTAL
I just want to get this in for today, and we need to really clean up  the handling of users specifying registry and/or transport.

pull.go
```diff
@@ -173,6 +173,9 @@ func pullImage(ctx context.Context, store storage.Store, imageName string, optio
 	if err == nil {
 		return destRef, nil
 	}
+	if !useRegistries {
```
mtrmac: So, the parameter does not say “registries should be used” at all, as the name suggests, but it affects how errors are reported. That’s confusing.

I don’t think the code below, which talks about short names and search, should be down here in `pullImage` in the first place: it belongs where the search behavior is implemented, i.e. in `resolveImage`.  At that point we can have a design discussion about what `resolveImage` should do on failure:
- maybe collect a list of all (registry, error) pairs, and report them all
- or maybe report the underlying error when exactly one registry is involved
- or maybe continue to do what it does now and just vaguely mention the search list

util/util.go
```diff
@@ -39,8 +40,10 @@ var (
 		"index.docker.io": "library",
 		"docker.io":       "library",
 	}
+
 	// Transports contains the possible transports used for images
 	Transports = map[string]string{
+		docker.Transport.Name():        "",
```
mtrmac: Surprisingly, this probably does not work: the `ResolveName` use below returns the after-`:` part as a docker-like reference string, i.e. `docker://busybox` would return `//busybox`.  (AFAICT the way `ResolveName` and `resolveImage` work together to handle `docker://busybox` is _pretty_ non-obvious.)

new.go
```diff
@@ -222,6 +222,7 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
 	var img *storage.Image
 	var err error
 
+	useRegistries := options.Transport == ""
```
mtrmac: As mentioned elsewhere, primarily the “helpfully tell the user what registries where tried, and unhelpfully be silent about the underlying error” logic, if it should exist at all, should be moved from `pullImage` to `resolveImage`. Then we _might_ find out that the “search registry list was used” information (not “use [search?] registries” option) is not necessary at all.

And even if we do decide that the “search registry list was used” information is desirable:
- This does not compute the right thing (looking at `ResolveName`, something like `FromImage == "example.com/busybox"` would cause `useRegistries` to be `true`, but `ResolveName` does not actually use the search registry list in that case)
- This is not the right place to compute that information; `ResolveName` _knows_ whether it used the search list or not, so it should just return that as an additional return value, instead of `newBuilder` (_very_ far from where the decision is made, pretty much at the opposite end of the call stack) independently determining the same thing.

mtrmac:
> I just want to get this in for today

When faced with non-trivial code with unexpected behaviors in corner cases, my personal preference would be to stop adding more non-trivial code with unexpected corner cases, even when in a hurry :)

> This is a partial solution but it at least fixes the case where user specifies a transport.

I’m not 100% sure what ”the case” being fixed here is; is it all just better error messages, or is there something that just fails and shouldn’t have?

If you want to get _some_ useful error messages, just drop all the heuristics in `pullImage` to be
```go
	if err = cp.Image(ctx, policyContext, destRef, srcRef, getCopyOptions(options.ReportWriter, sc, nil, "")); err != nil {
		return nil, err
	}
	return destRef, nil
 
```
and, I don’t know, have `resolveImage` record the first `pullAndFindImage` error cause and return it at the end of the loop.

Or is it about 
https://github.com/projectatomic/libpod/issues/1212#issuecomment-410350051 ? I haven’t actually _tested_ that but it seems to me that the solution is also to drop the heuristics at the end of `pullImage` (the failure happens because `hasRegistry` at the end of `pullImage` chokes on `docker://…`).

---

BTW, a lot of this `buildah` code and the code in `podman` has obviously shared origins, even if they drifted away over time (and a large part of that drift is my fault). Is there any long-term idea that the two would eventually share code, and if so, in what direction?

E.g. the `buildah` `localImageNameForReference` / `podman` `pullGoalFromImageReference` set of heuristics should very likely be shared.

util/util.go
mtrmac: … ah, it maybe actually does, because `resolveImage` stitches `BuilderOptions.Transport` and `//busybox` back together.  Mostly, the `hasRegistry` at the end of `pullImage` will choke on it again.

_Please_ let’s not add yet another possible syntax of strings returned from `ResolveName` like that.

PR #908, opened 2018-08-03
Add a few debug messages to pull

TomSweeneyRedHat:

Add a few debug statements to find out why a pull failed.  Also if we encounter a URL error, handle it rather than checking for the registry which will cause "invalid reference format" errors.

Testing: 
=====
```
# ./buildah from docker://registry1.fedoraproject.org/fedora-minimal
Get https://registry1.fedoraproject.org/v2/: dial tcp: lookup registry1.fedoraproject.org on 192.168.122.1:53: no such host
```

### DIFF
--- a/pull.go
+++ b/pull.go
@@ -3,6 +3,7 @@ package buildah
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"reflect"
 	"strings"
 
@@ -179,6 +180,12 @@ func pullImage(ctx context.Context, store storage.Store, imageName string, optio
 	case signature.PolicyRequirementError:
 		logrus.Debugf("PolicyRequirementError raised")
 		return nil, err
+	case *url.Error:
+		logrus.Debugf("URL Error raised, tranport: [%v]", options.Transport)
+		if options.Transport == "" {
+			// Not using localhost, a url error that's a problem
+			return nil, err
+		}
 	default:
 		logrus.Debugf("Generic Error raised [%v] [%v]", err, reflect.TypeOf(err))
 	}
@@ -187,13 +194,16 @@ func pullImage(ctx context.Context, store storage.Store, imageName string, optio
 	registryPath := sysregistries.RegistriesConfPath(sc)
 	searchRegistries, err := getRegistries(sc)
 	if err != nil {
+		logrus.Debugf("getRegistries Error for image [%s] [%v] [%v]", imageName, err, reflect.TypeOf(err))
 		return nil, err
 	}
 	hasRegistryInName, err := hasRegistry(imageName)
 	if err != nil {
+		logrus.Debugf("hasRegistry Error for image [%s] [%v] [%v]", imageName, err, reflect.TypeOf(err))
 		return nil, err
 	}
 	if !hasRegistryInName && len(searchRegistries) == 0 {
+		logrus.Debugf("No registry in name for image [%s] [%v] [%v]", imageName, err, reflect.TypeOf(err))
 		return nil, errors.Errorf("image name provided is a short name and no search registries are defined in %s.", registryPath)
 	}
 	return nil, NewGenericPullError(fmt.Sprintf("unable to find image in the registries defined in %q", registryPath))


rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+

TomSweeneyRedHat:
@rhatdan I added a url check to this, I like it better than #907

TomSweeneyRedHat:
forgot PTAL @rhatdan

pull.go
```diff
@@ -187,13 +194,16 @@ func pullImage(ctx context.Context, store storage.Store, imageName string, optio
 	registryPath := sysregistries.RegistriesConfPath(sc)
 	searchRegistries, err := getRegistries(sc)
 	if err != nil {
+		logrus.Debugf("getRegistries Error for image [%s] [%v] [%v]", imageName, err, reflect.TypeOf(err))
```
mtrmac: (FWIW you can use %#v to get a full representation of `err` without having to call `reflect `explicitly)

TomSweeneyRedHat: Neat!  Golang learning of the day for me, thanks @mtrmac!

pull.go
```diff
@@ -179,6 +180,12 @@ func pullImage(ctx context.Context, store storage.Store, imageName string, optio
 	case signature.PolicyRequirementError:
 		logrus.Debugf("PolicyRequirementError raised")
 		return nil, err
+	case *url.Error:
+		logrus.Debugf("URL Error raised, tranport: [%v]", options.Transport)
+		if options.Transport == "" {
+			// Not using localhost, a url error that's a problem
```
mtrmac: - When is the `options.Transport == ""` check true? AFAICT it is always non-empty, both because `cmd/*` set it to `DefaultTransport`, and because `NewBuilder` does.

- I have no idea what “not using localhost” means. Is that `docker://localhost/foo`, `localhost/foo`, both, neither?

- It is not obvious to me at all that various kinds errors should be handled differently in here (as I argue in #907 , most of that should probably be in `resolveImage`, along with the search loop), and similarly to #907, the `options.Transport == ""` check seems like an imprecise way to determine whether the registry search list is involved. `utils.ResolveName` knows precisely whether that happened, no need to heuristically guess again.

(I would be quite fine with this PR without the `options.Transport == ""` check, whether or not the code returns original `err` here.)

TomSweeneyRedHat: The problem is if you do `buildah from alpine` the pull process first tries to pull it from localhost/alpine before going out to the registries.   This call using localhost/alpine returns a url error if a localhost registry isn't set up.   Prior to all these changes we just ignored the error from the localhost lookup  and then tried looking for alpine based on the entries in registries.conf.  Now it fails on that localhost lookup.  Dan found that if the transport isn't specified, then we know someone has given us just 'alpine' and we can then continue on and ignore the url error.

The thing that's biting me with this change is in some cases it's OK to get a failure and others not.  Trying to specify which are good and which are bad errors has proven challenging especially as we're pushing to try and get a release out.

mtrmac: - I can’t see that `options.Transport == ""` can happen (#907 makes it happen for the brief time necessary to set `useRegistries`; but that’s not included here); perhaps I am just being blind.

- Again, `utils.ResolveName` _knows precisely_ whether the input contained a registry or not.

- `options.Transport == ""` (as well as the “input contained a registry” indication potentially returned by `utils.ResolveName`) is constant for all the names returned by `utils.ResolveName`; neither actually answers the question “are we pulling from `localhost`”

- The error handling should be much easier to do in `resolveImage`.

rhatdan:
@TomSweeneyRedHat Needs a rebase, do we still want this?

TomSweeneyRedHat:
@rhatdan we should do something here, I just haven't gotten back roundtoit.  Need to do so.

rhatdan:
@TomSweeneyRedHat Needs a rebase.

TomSweeneyRedHat:
I'm going to close this rather than rebase as I think @nalind's #1072 addressed most if not all of this one.

PR #909, opened 2018-08-04
Pull Error handling: talk about search registries again, more intelligent output, cleanups

mtrmac:
<s>WIP ABSOLUTELY UNTESTED.</s>

Motivated by #907 and #908: Refactor pull error handling; instead of throwing away most errors, go to the other extreme and report all of them when searching multiple registries.

_Mostly_ this is fairly straightforward refactoring and ~simplification. The last commit, though, fairly drastically changes `resolveImage` error handling to first collect the details of all failures to pull the image, and _then_ at the end decide how to report it to the user.

**I am fairly sure the result is not good enough**, but _having that information_ at all in `resolveImage` should allow making more intelligent decisions **EDIT than trying to do the right thing deep in `pullImage` with incomplete information**.

The `resolveImage` loop body is fairly big and maybe should be split now into a separate function.

This might need at least one more step, explicitly recognizing the `localhost` case as likely-to-fail and being more silent about failures to pull it (but not other failures involving it).

See individual commits for details. This CHANGES BEHAVIOR, marked like that in the commit messages.

Feel free to reuse and/or modify all of parts of this as necessary; or tell me to throw it away.

PR #881, opened 2018-07-24
Return policy error on pull

TomSweeneyRedHat:
Signed-off-by: TomSweeneyRedHat <tsweeney@redhat.com>

When a pull failed due to a signature policy violation, the error returned was:  
```
ERRO[0001] image "quay.io/tomsweeneyredhat/myalpine" not found in /etc/containers/registries.conf registries: image not known
```

The PolicyRequirementError that triggered the failure but itself was lost: 
```
Running image docker://quay.io/tomsweeneyredhat/myalpine:latest is rejected by policy.
```
With this change the failure now looks like:

```
# buildah from quay.io/tomsweeneyredhat/myalpine
Source image rejected: Running image docker://quay.io/tomsweeneyredhat/myalpine:latest is rejected by policy.
```
I'm a little leery about doing a return of the error instead of the continues.  I thought about doing a lastError thing, but that go messy too.  However if @nalind thinks that's the better way to go, I'll convert.

This addresses at least in part: #849

PR #910, opened 2018-08-04
Revert pull error handling from 881

TomSweeneyRedHat:

This PR reverts the code put into play for Policy Rejection Error handling with #881.  It leaves the test code that was submitted in that PR in place as that PR found prior issues in our tests.  I'm suggesting that we merge this in now and build v1.3 with this version.  When a bad image is requested to be pulled, the pre-existing error, although not great, is shown:

```
# buildah from docker://registry1.fedoraproject.org/fedora-minima
image "docker://registry1.fedoraproject.org/fedora-minima" not found in /etc/containers/registries.conf registries: image not known
```

After v1.3 ships, we can then revisit using #907, #908, or more likely #909 next week when we've time to investigate more fully.

This passes the bats tests on my vm and also the baseline tests.

@mtrmac and @rhatdan thanks for your help and guidance on these changes.  I truly appreciate it.


rhatdan:
I agree.  We can work with @mtrmac to fix the problem correctly.
@rh-atomic-bot r+

PR #918, opened 2018-08-08
Export a Pull() function

nalind:
Provide a Pull() function that can be called directly, to go with the Push() function.


rhatdan:
@rh-atomic-bot r+

PR #921, opened 2018-08-09
Don't bother with --no-pivot for rootless isolation

nalind:
When running outside of a container, --no-pivot isn't necessary, and when running inside of a container, it's not enough to solve any of the difficulties we're seeing there.  It may trigger an EPERM for unshare()
calls inside of the container that we launch, and we don't want that, so drop it, for now at least.

### DIFF
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -219,11 +219,11 @@ another process.
 Controls what type of isolation is used for running processes as part of `RUN`
 instructions.  Recognized types include *oci* (OCI-compatible runtime, the
 default), *rootless* (OCI-compatible runtime invoked using a modified
-configuration and its --rootless flag enabled, with *--no-new-keyring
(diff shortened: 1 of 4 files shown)


rhatdan:
Is there a reason to use this Option?

nalind:
I had thought it was be necessary for running `runc` inside of a container.  It might still be necessary, but it isn't sufficient AFAICT.

rhatdan:
@giuseppe WDYT?

giuseppe:
yes I agree, I think it is not required for buildah.  I was surprised when I saw it used (and had to implement it for crun to keep it usable with Buildah: https://github.com/giuseppe/crun/commit/d32bf5cb57cb10be5703ca5f3d5336da897fc730)

LGTM

rhatdan:
@rh-atomic-bot r+

PR #922, opened 2018-08-09
Remove an unused srcRef.NewImageSource in pullImage

mtrmac:
- The result is not used anywhere
- In the common case of pulling from `docker://` this does not do anything notable (only sets up configuration, no network checks)
- ... but it can be pretty expensive for compressed archives, creating an uncompressed on-disk copy.
- `copy.Image` soon after the removed code calls `srcRef.NewImageSource` internally anyway.

This should not change behavior on success; it may change which error is reported if there is more than one reason for the pull to fail.

### DIFF
--- a/pull.go
+++ b/pull.go
@@ -173,12 +173,6 @@ func pullImage(ctx context.Context, store storage.Store, imageName string, optio
 		return nil, errors.Wrapf(err, "error parsing image name %q", destName)
 	}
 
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM
I'm going to try and rekick the tests, the error is a memory segv with volumes and I don't think at all related with this change.

TomSweeneyRedHat:
bot, retest this please.

rhatdan:

PR #923, opened 2018-08-10
chroot: bind mount an empty directory for masking

nalind:
In chroot isolation, when we attempt to mask a directory, use a read-only bind mount of an empty directory instead of a read-only mount of a fresh tmpfs with size=0, which is more likely to be be denied by mandatory access controls.

### DIFF
--- a/chroot/run.go
+++ b/chroot/run.go
@@ -1131,6 +1131,15 @@ func setupChrootBindMounts(spec *specs.Spec, bundlePath string) (undoBinds func(
 		}
 	}
 
+	// Create an empty directory for to use for masking directories.
+	roEmptyDir := filepath.Join(bundlePath, "empty")
+	if len(spec.Linux.MaskedPaths) > 0 {
+		if err := os.Mkdir(roEmptyDir, 0700); err != nil {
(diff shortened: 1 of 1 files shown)


chroot/run.go
```diff
@@ -901,6 +901,13 @@ func setupChrootBindMounts(spec *specs.Spec, bundlePath string) (undoBinds func(
 		return err
 	}
 
+	// Create an empty directory for possible use later on.
+	roEmptyDir := filepath.Join(bundlePath, "empty")
+	if err := os.Mkdir(roEmptyDir, 0700); err != nil {
+		return undoBinds, errors.Wrapf(err, "error creating empty directory %q", roEmptyDir)
+	}
+	removes = append(removes, roEmptyDir)
```
TomSweeneyRedHat: removes and roEmptyDir aren't used until the for statement at 990 and there are a few possible errors between here and there.  Does it make sense to move this chunk down just before that for statement?

nalind: No harm in doing so.  Changing it.

rhatdan:
LGTM
@rh-atomic-bot r+

TomSweeneyRedHat:
LGTM, thx for the touch up.

PR #935, opened 2018-08-15
Show lastest container name in containers cmd

TomSweeneyRedHat:

With the new rename command the new name of the container was placed at names[0] however in 'buildah containers` we were showing the name from the builder.container field which was the name of the container at it's creation.  This changes that and reworks the structure of the code a bit to make use of a map and slices in both parts of an if statement.

Testing:
```
# buildah from fedora
# buildah rename fedora-working-container my-fedora

Before:
# buildah containers
6a26f8bdcc75     *     cc510acfcd70 docker.io/library/fedora:latest  fedora-working-container
# buildah containers --all
6a26f8bdcc75     *     cc510acfcd70 docker.io/library/fedora:latest  my-fedora

After:
# buildah containers
6a26f8bdcc75     *     cc510acfcd70 docker.io/library/fedora:latest  my-fedora
[root@localhost buildah]# buildah containers --all
6a26f8bdcc75     *     cc510acfcd70 docker.io/library/fedora:latest  my-fedora
```

### DIFF
--- a/cmd/buildah/containers.go
+++ b/cmd/buildah/containers.go
@@ -147,48 +147,59 @@ func outputContainers(store storage.Store, opts containerOptions, params *contai
 		return seenImages[id]
 	}
 
-	builders, err := openBuilders(store)
-	if err != nil {
-		return errors.Wrapf(err, "error reading build containers")
-	}
 	var (
 		containerOutput []containerOutputParams
 		JSONContainers  []jsonContainer
 	)
+	builders, err := openBuilders(store)
+	if err != nil {
+		return errors.Wrapf(err, "error reading build containers")
+	}
+
+	builderMap := make(map[string]struct{})
+	for _, builder := range builders {
+		builderMap[builder.ContainerID] = struct{}{}
+	}
+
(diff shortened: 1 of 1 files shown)


rhatdan:
@TomSweeneyRedHat You said the old name is still associated with the container?

rhatdan:
I would figure if I rename X->Y, I would be able to recreate a new container named X?

q384566678:
SGTM

TomSweeneyRedHat:
I wasn't sure which way you wanted to go and haven't circled back.  We've the choice of doing it as is in this PR, keeping the original name.  Or we can pretty much chuck all of this and change the name but lose the history.  Is the rename everywhere the approach you'd prefer?

rhatdan:
Yes, We can warn about the loss of history.  But to me Rename means rename.  And this looks more like a tag.
What does Docker rename do?

TomSweeneyRedHat:
Looks like Docker wipes the old name from the container.  I'll back out the change to the containers code then and will change things somewhere in the rename code path to do that.

PR #944, opened 2018-08-21
cmd: remove SkipArgReorder

### LINKED ISSUE #940: 'buildah from' error on image argument placing is unclear
**Description**
Doing ` buildah from scratch --name builder` yields:
```
too many arguments specified
ERRO[0000] exit status 1
```
The error being raised is fine, but the message should instead be something like "image name is not the last argument provided".

1. See description

See description.

An error pointing out that the image needs to be the last argument.

```
# buildah version
Version:         1.4-dev
Go Version:      go1.9.7
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.3.1
libcni Version:  v0.6.0
Git Commit:      00dd72aa
Built:           Thu Aug 16 13:35:08 2018
OS/Arch:         linux/amd64
```

q384566678:
Fixes #940 

Before:
```
➜  buildah git:(from-issue) ✗ sudo buildah from scratch --name builder 
too many arguments specified
```
After:
```
➜  buildah git:(from-issue) ✗ sudo ./buildah from scratch --name builder 
builder
➜  buildah git:(from-issue) ✗ sudo buildah containers    
e173e4ed21d3     *                  scratch                          builder

```

### DIFF
--- a/cmd/buildah/addcopy.go
+++ b/cmd/buildah/addcopy.go
@@ -35,13 +35,12 @@ var (
 	}
 
 	copyCommand = cli.Command{
-		Name:           "copy",
-		Usage:          "Copy content into the container",
-		Description:    copyDescription,
-		Flags:          addAndCopyFlags,
-		Action:         copyCmd,
-		ArgsUsage:      "CONTAINER-NAME-OR-ID [FILE | DIRECTORY | URL] [[...] DESTINATION]",
-		SkipArgReorder: true,
+		Name:        "copy",
+		Usage:       "Copy content into the container",
+		Description: copyDescription,
+		Flags:       addAndCopyFlags,
+		Action:      copyCmd,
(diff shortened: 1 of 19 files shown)


vrothberg:
@q384566678, thanks a lot for the PR. Although it fixes the issue, I have some doubts if that's the way to go.

As mentioned in #940, the image is required to be the last argument to `buildah-from`, which is not the case with the changes in this PR anymore. I am also skeptical towards the reordering of arguments as it can introduce very subtle bugs and some CLI invocations can turn into something quite ambiguous.

TomSweeneyRedHat:
Thanks for the PR @q384566678 , but sorry, I concur with @vrothberg .  We don't want to remove the reorder flag for a number of reasons that he stated.  For the `buildah from` command the last argument should be the image name.  We need to "just" change the error returned in this instance, but I quote the just as I'm not sure it will be a quick and easy fix.

rhatdan:
I agree.  We want to be consistent and have the reorder flag in all the tools

PR #957, opened 2018-08-24
Warn about using Commands in Dockerfile that are not supported by OCI.

rhatdan:

### DIFF
--- a/cmd/buildah/from.go
+++ b/cmd/buildah/from.go
@@ -36,6 +36,12 @@ var (
 			Usage: "use `[username[:password]]` for accessing the registry",
 		},
 		cli.StringFlag{
+			Name:   "format, f",
+			Usage:  "`format` of the image manifest and metadata",
+			Value:  defaultFormat(),
+			Hidden: true,
+		},
+		cli.StringFlag{
 			Name:  "name",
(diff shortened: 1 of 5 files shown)


rhatdan:
@nalind @TomSweeneyRedHat @pixdrift @umohnani8 PTAL

cmd/buildah/from.go
```diff
@@ -211,6 +212,16 @@ func fromCmd(c *cli.Context) error {
 	}
 	namespaceOptions.AddOrReplace(usernsOption...)
 
+	format := c.String("Format")
+	switch format {
+	case "oci":
+		format = imagebuildah.OCIv1ImageFormat
+	case "docker":
+		format = imagebuildah.Dockerv2ImageFormat
+	default:
```
TomSweeneyRedHat: Same comment as from another review, can we merge these definition out of image.go and imagebuildah and into common.go?

rhatdan: I want to get the other one merged first then I can clean this one up.

rhatdan:
@rh-atomic-bot r=umohnani8

rhatdan:
@rh-atomic-bot retry

PR #958, opened 2018-08-24
Cleanup code to more closely match Docker Build images

rhatdan:
Hostname should be set to a random name if not set.
Workdir should default to "" not to "/"
If Volumes and Entrypoint are not set they should return NULL not []

### DIFF
--- a/config.go
+++ b/config.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"os"
-	"path/filepath"
 	"runtime"
 	"strings"
 	"time"
@@ -12,6 +11,7 @@ import (
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/transports"
 	"github.com/containers/image/types"
+	"github.com/containers/storage/pkg/stringid"
 	ociv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/projectatomic/buildah/docker"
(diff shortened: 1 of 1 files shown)


rhatdan:
Fixes most of https://github.com/projectatomic/buildah/issues/953

TomSweeneyRedHat:
Changes LGTM, but I want to run it through a baseline test.  I think the entrypoint change is OK, but want to verify.

TomSweeneyRedHat:
The majority of the tests ran fine, but the test that pushes a bulidah image to Docker and then runs the image using Docker is failing.  

'''
FILE=./runecho.sh
/bin/cat <<EOM >$FILE
#!/bin/bash
for i in {1..9};
do
    echo "This is a new container from ipbabble [" \$i "]"
done
EOM
chmod +x $FILE

# newcontainer=$(buildah from scratch)
# scratchmnt=$(buildah mount $newcontainer)
# dnf install --installroot $scratchmnt --release 28 bash coreutils --setopt install_weak_deps=false -y
# buildah copy $newcontainer $FILE /usr/bin
# buildah config --cmd /usr/bin/runecho.sh $newcontainer
# buildah run $newcontainer /usr/bin/runecho.sh 
# buildah unmount $newcontainer
# buildah commit $newcontainer fedora-bashecho
# buildah push fedora-bashecho docker-daemon:fedora-bashecho:latest

# docker run fedora-bashecho
standard_init_linux.go:178: exec user process caused "permission denied"


'''

TomSweeneyRedHat:
I'm backing out Buildah, installing latest rpm to see if it's a vm config issue...

TomSweeneyRedHat:
Hmm, same error with the Buildah released kit on board, not this change, but still worrisome.  More digging.

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
FWIW, something was funky in the installed docker dirs.  I removed /var/lib/docker, restarted docker and then it worked just fine.

rhatdan:

PR #961, opened 2018-08-27
Add buildah bud conformance test

ypu:
The conformance test for buildah bud is used to check the images
build with buildah against docker. Currently we have two types of
tests:
1. Image build output test for features that support with docker ce
only. Test steps:
    - Prepare Dockerfile
    - Build image with buildah
    - Check the command output
2. General tests. Test steps:
    - Prepare Dockerfile
    - Build image with buildah
    - Push buildah image to docker with podman
    - Check the image id is not changed
    - Build image with docker
    - Check key part of the build command output from buildah and
      docker is the same(optional)
    - Check image info by docker inspect. And compare the output
      from two images. The key items should be the same.
    - Check the files inside the image with container-diff. They
      should be the same.

The tests are written with ginkgo. And request docker, podman and
container-diff installed in the test env.


TomSweeneyRedHat:
@nalind @rhatdan PTAL

rhatdan:
LGTM
We might want to look at how many of these we could make blocking. Without causing the tests to run for a huge amount of time.

rhatdan:
@rh-atomic-bot r+

PR #963, opened 2018-08-28
Remove redundant variable declarations

zhouhao3:
Signed-off-by: Zhou Hao <zhouhao@cn.fujitsu.com>

PR #967, opened 2018-08-29
cmd/rm: restore the correct containerID display

q384566678:
This code was modified in #963. At the time, I didn't notice that the ContainerID would be emptied after the [delete](https://github.com/projectatomic/buildah/blob/master/delete.go#L16) operation,so the ID displayed after deleting the container is empty.
Now the relevant code is restored. I am sorry for my negligence.


rhatdan:
@rh-atomic-bot r+

PR #968, opened 2018-08-29
Make buildah-from error message clear when flags are after arg

### LINKED ISSUE #940: 'buildah from' error on image argument placing is unclear

boaz1337:
**NOTE**: this is my way to solve #940 - without setting `SkipArgOrder` to false(https://github.com/projectatomic/buildah/pull/944). Please let me know what you think.

The error message in buildah-from can be confusing when a flag is specified after the argument (image name).

For example, running the following: 

```bash
$ buildah from scratch -q
too many arguments specified
```

however only one argument was specified`scratch`. the reason we get this error is because anything that comes after "scratch" will be considered as an argument.

To fix this - check if every argument is a flag (flag starts with "-" e.g. "-q", "--ulimit").
If you find at least one argument that is a flag return a much clearer error message:

```
-q should be set before the image name
```

otherwise, return the old error.

closes #940

### DIFF
--- a/cmd/buildah/from.go
+++ b/cmd/buildah/from.go
@@ -160,6 +160,11 @@ func fromCmd(c *cli.Context) error {
 		return errors.Errorf("an image name (or \"scratch\") must be specified")
 	}
 	if len(args) > 1 {
+		for _, arg := range args {
+			if strings.HasPrefix(arg, "-") {
+				return errors.Errorf("%s should be set before the image name", arg)
+			}
+		}
 		return errors.Errorf("too many arguments specified")
 	}
 	if err := parse.ValidateFlags(c, fromFlags); err != nil {
--- a/tests/from.bats
+++ b/tests/from.bats
@@ -2,6 +2,28 @@
 
 load helpers
 
+@test "from bad order of flags" {
+  run buildah from scratch -q
+  [ "$status" -eq 1 ]
+  [[ $output = *"-q should be set before the image name"* ]]
+
+  run buildah from scratch --pull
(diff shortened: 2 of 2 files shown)


TomSweeneyRedHat:
bot, add author to whitelist

TomSweeneyRedHat:
@boaz1337 This works great in about 85+% of the cases.  I did find that if there's no equal sign in the second bad argument, then you get the "too many arguments specified." error.  Can you tweak this to handle this kind of bad request:

# buildah from scratch --cred="fake fake" --name small
too many arguments specified

boaz1337:
@TomSweeneyRedHat nice catch!
I believe I have a better implementation.

boaz1337:
Before you merge, I just want to change the commit message.
Please :smile:

TomSweeneyRedHat:
NP, I was going to let @rhatdan merge it and he's not likely to do so until the morning.

rhatdan:
@rh-atomic-bot r+

PR #973, opened 2018-08-30
cmd/containers: End loop early when using the json option

q384566678:


rhatdan:
LGTM
@TomSweeneyRedHat PTAL

Issue #974, opened 2018-08-30
buildah bud : on commit error copying layers and metadata


olivierGATSNCF:
**Description**

I'm trying to build a buildah image  that has an image from openshift / base-centos7

With docker I can build my image buildah.
With buildah bud, during the commit of the image an error occurs which seems to be due to a bad reading of the history of the manifest of the image from openshift/base-centos7:
```
error copying layers and metadata: Error initializing source containers-storage:base-centos7-working-container: internal error: history lists 5 non-empty layers, but we have 16 layers on disk.
```

buildah version 1.1 or 1.3 the result is the same.

**Steps to reproduce the issue:**
1. buildah bud --format docker --build-arg HTTP_PROXY=http://xxxx --build-arg HTTPS_PROXY=http://xxxx --build-arg http_proxy=http://xxxx --build-arg https_proxy=http://xxxx --tls-verify=false --pull -f Dockerfile --tag registyr/buildtools/buildah_v1:2 --label buildtools.build-date=20180828 --label buildtools.git-commit=f2a6485b785d5e07d92bb1bc55be75279fb620a2 --build-arg no_proxy=registry --build-arg IMAGE_VERSION=2 .

**Describe the results you received:**
STEP 11: CMD ["/bin/bash"]    
STEP 12: COMMIT containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.override_kernel_check=true]docker-eul-releases-nexusm-repo-gpf.apps.eul.sncf.fr/buildtools/buildah_v1:2
EXITCODE   0error copying layers and metadata: Error initializing source containers-storage:base-centos7-working-container: internal error: history lists 5 non-empty layers, but we have 16 layers on disk

**Describe the results you expected:**

**Output of `rpm -q buildah` or `apt list buildah`:**
(buildah used to build the image is  a docker container. )
```
package buildah is not installed
```

**Output of `buildah version`:**

```
buildah version
Version:         1.3
Go Version:      go1.9.4
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.3.1
libcni Version:  v0.6.0
Git Commit:      4888163
Built:           Thu Aug 23 14:48:55 2018
OS/Arch:         linux/amd64
```

**Output of `cat /etc/*release`:**

```
CentOS Linux release 7.3.1611 (Core) 
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
[...]

PR #976, opened 2018-08-30
Update containers/image to fix commit layer issue

TomSweeneyRedHat:

Update container/image to address a commit error when copying layers and metadata.  This change may require users to recreate containers.  This change also required that the xz tools from ulikunitz be vendored too.  It was in the vendor.conf, but it doesn't appear that we actually vendored it.  I was unable to compile with the new container/image without vendoring it to.

@mtrmac PTAL.

Hopefully addresses #974


mtrmac:
ACK on the `vendor/github.com/containers/image` part.

TomSweeneyRedHat:
@rhatdan between this and the container/storage update alone, it's probably wise to get a new kit together soonish.

PR #984, opened 2018-09-03
Add flags/arguments order verification to one-arg commands - easy

boaz1337:
A follow up to https://github.com/projectatomic/buildah/pull/968#issuecomment-417158706

It's a work in progress as I have to add more tests and make sure I covered all the necessary commands.

### DIFF
--- a/pkg/cli/common.go
+++ b/pkg/cli/common.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
 	"github.com/projectatomic/buildah"
 	"github.com/projectatomic/buildah/util"
 	"github.com/urfave/cli"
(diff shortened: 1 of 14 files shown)


pkg/cli/common.go
```diff
@@ -280,3 +281,12 @@ func DefaultIsolation() string {
 	}
 	return buildah.OCI
 }
+
+func VerifyFlagsArgsOrder(args []string) error {
+	for _, arg := range args {
+		if strings.HasPrefix(arg, "-") {
+			return errors.Errorf("%s should be set before the image name", arg)
```
rhatdan: Perhaps make this a little more informative.

"No options (%s) can be specified after the image or container name"

boaz1337: Yep, I like this better. Thanks!

boaz1337:
This PR is trivial: adding verification to only one argument commands such as `commit`, `config` and etc.
The next part is adding this verification to `images`, `mount`, `push`, `rename` and etc.

Commands that don't need this verification are `version` and `containers`.

umohnani8:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #985, opened 2018-09-03
unshare: keep the additional groups of the user

giuseppe:
Do not reset the additional groups for a rootless user when creating a
new user namespace.  In this way, users of the "docker" group are
still able to use the docker-deamon: destination.

Closes: https://github.com/projectatomic/buildah/issues/978

### DIFF
--- a/unshare/unshare.c
+++ b/unshare/unshare.c
@@ -31,7 +31,7 @@ static int _buildah_unshare_parse_envint(const char *envname) {
 
 void _buildah_unshare(void)
 {
-	int flags, pidfd, continuefd, n, pgrp, sid, ctty, allow_setgroups;
+	int flags, pidfd, continuefd, n, pgrp, sid, ctty;
 	char buf[2048];
 
 	flags = _buildah_unshare_parse_envint("_Buildah-unshare");
@@ -83,14 +83,7 @@ void _buildah_unshare(void)
 			_exit(1);
 		}
 	}
-	allow_setgroups = _buildah_unshare_parse_envint("_Buildah-allow-setgroups");
 	if ((flags & CLONE_NEWUSER) != 0) {
-		if (allow_setgroups == 1) {
-			if (setgroups(0, NULL) != 0) {
-				fprintf(stderr, "Error during setgroups(0, NULL): %m\n");
-				_exit(1);
-			}
-		}
 		if (setresgid(0, 0, 0) != 0) {
 			fprintf(stderr, "Error during setresgid(0): %m\n");
 			_exit(1);
--- a/unshare/unshare.go
+++ b/unshare/unshare.go
@@ -84,11 +84,6 @@ func (c *Cmd) Start() error {
 		c.Env = append(c.Env, fmt.Sprintf("_Buildah-ctty=%d", len(c.ExtraFiles)+3))
 		c.ExtraFiles = append(c.ExtraFiles, c.Ctty)
 	}
-	if c.GidMappingsEnableSetgroups {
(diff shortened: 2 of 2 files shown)


unshare/unshare.c
```diff
@@ -85,12 +85,6 @@ void _buildah_unshare(void)
 	}
 	allow_setgroups = _buildah_unshare_parse_envint("_Buildah-allow-setgroups");
```
rhatdan: Remove this line?

rhatdan:
Looks like the allow-setgroups code is set in two places in unshare.go also.
./unshare/unshare.go:		c.Env = append(c.Env, "_Buildah-allow-setgroups=1")
./unshare/unshare.go:		c.Env = append(c.Env, "_Buildah-allow-setgroups=0")

If we are not going to use it any longer, then we should remove it here also.

rhatdan:
@nalind PTAL

cmd/buildah/unshare.go
```diff
@@ -181,6 +181,7 @@ func maybeReexecUsingUserNamespace(c *cli.Context, evenForRoot bool) {
 	cmd.UidMappings = uidmap
 	cmd.UseNewgidmap = uidNum != 0
 	cmd.GidMappings = gidmap
+	//fmt.Printf("AAA %s %s\n", c.GlobalIsSet("enable-setgroups"), c.GlobalBool("enable-setgroups"))
```
rhatdan: Is this debugging?

cmd/buildah/main.go
```diff
@@ -75,6 +75,10 @@ func main() {
 			Usage:  "path to default mounts file",
 			Hidden: true,
 		},
+		cli.BoolTFlag{
+			Name:  "allow-setgroups",
```
rhatdan: Is this used somewhere?

cmd/buildah/unshare.go
giuseppe: argh yes, I've added the modified files with `git add` but then I've used `git commit -a` that included also the changes that weren't supposed to be added.

Fixed now

unshare/unshare.go
```diff
@@ -84,11 +84,6 @@ func (c *Cmd) Start() error {
 		c.Env = append(c.Env, fmt.Sprintf("_Buildah-ctty=%d", len(c.ExtraFiles)+3))
 		c.ExtraFiles = append(c.ExtraFiles, c.Ctty)
 	}
-	if c.GidMappingsEnableSetgroups {
```
nalind: This nullifies any effect that setting the field has, but leaves the field in the structure.  Why?

giuseppe: it is still used to write `allow` or `deny` to the `/proc/PID/setgroups` file

TomSweeneyRedHat:
bot retest this please.

rhatdan:
@giuseppe Can you answer @nalind question?

giuseppe:
> @giuseppe Can you answer @nalind question?

I thought I already did.  Is there any question I've left unanswered?

rhatdan:
@nalind Are you satisfied with the answer?
@giuseppe Please fix the travis errors.

giuseppe:
Travis keeps having flakes, I re-ran the tests several times and everytime they fail in a different place.  I don't think it depends from these changes

rhatdan:
@rh-atomic-bot r+

giuseppe:
@rh-atomic-bot retry

rhatdan:
Seems like Homu is hung.  Merging manually.

PR #988, opened 2018-09-05
run: clear default seccomp filter if not enabled

nalind:
When seccomp is not enabled, make sure to clear any default setting which runtime-tools might have supplied for us.  Likewise, if SELinux is not enabled, don't set a process label or a mount label.


nalind:
@TomSweeneyRedHat it's a metalinter error, fixing.

rhatdan:
LGTM
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot r+

nalind:
@rh-atomic-bot retry

PR #991, opened 2018-09-06
Exclude --layers from the common bug cli flags

baude:
Other applications like podman would like to switch the the default value
for --layers.  Because we re-use this code, we just need to pop layers out
of the budflags.


umohnani8:
LGTM

rhatdan:
@rh-atomic-bot r+
@TomSweeneyRedHat He is going to change the default of --layers on podman.

PR #994, opened 2018-09-07
Push: Allow an empty destination

jwendell:
In this case assume it's the same as source. This makes it
compatible with `docker push`.

Thus `buildah push my_registry/my_image` will behave like
`buildah push my_registry/my_image docker://my_registry/my_image`

### DIFF
--- a/cmd/buildah/push.go
+++ b/cmd/buildah/push.go
@@ -15,6 +15,7 @@ import (
 	buildahcli "github.com/projectatomic/buildah/pkg/cli"
 	"github.com/projectatomic/buildah/pkg/parse"
 	"github.com/projectatomic/buildah/util"
+	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
 )
 
@@ -58,7 +59,7 @@ var (
 	pushDescription = fmt.Sprintf(`
    Pushes an image to a specified location.
 
-   The Image "DESTINATION" uses a "transport":"details" format.
+   The Image "DESTINATION" uses a "transport":"details" format. If not specified, will reuse source IMAGE as DESTINATION.
 
    Supported transports:
    %s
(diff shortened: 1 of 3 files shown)


jwendell:
If you like the idea I can go ahead and update docs. My use case is to be able to replace `docker push` with `buildah push`

cmd/buildah/push.go
```diff
@@ -78,15 +79,23 @@ var (
...
+
+	args := c.Args()
+	var src, destSpec string
+	switch len(args) {
+	case 0:
+		return errors.New("source and destination image IDs must be specified")
+	case 1:
+		src = args[0]
+		destSpec = src
+		logrus.Infof("Destination argument not specified, assuming the same as the source: %s", destSpec)
```
rhatdan: Make this logrus.Debugf

cmd/buildah/push.go
```diff
@@ -78,15 +79,23 @@ var (
...
 		return err
 	}
-	src := args[0]
-	destSpec := args[1]
+
+	args := c.Args()
+	var src, destSpec string
+	switch len(args) {
+	case 0:
+		return errors.New("source and destination image IDs must be specified")
```
rhatdan: Should probably throw an error if the user specifies 3 or more Args.

rhatdan:
This looks ok to me.  Do we need similar changes to podman push?

rhatdan:
You also need to sign your PR.

git commit -a --amend -s
git push --force

cmd/buildah/push.go
TomSweeneyRedHat: This error is no longer true as you only need to specify a source with this change.  I'm struggling with verbiage, but maybe something like:

"At least a source image ID must be specified"

I'm struggling with trying to work the thought of the source becoming the destination if not specified, but can't come up with concise verbiage.  Perhaps the above is enough?

TomSweeneyRedHat:
Along with docs, a test or two would be good too.

rhatdan:
@jwendell We would like to get this into the next buildah release scheduled for tomorrow.  Are you able to continue working on it?

jwendell:
@rhatdan done, PTAL.

TomSweeneyRedHat:
The push tests are breaking, but I'm not sure why.  It doesn't look like it should be due to this change.  Going to try and retest.

TomSweeneyRedHat:
bot, retest this please

rhatdan:
@jwendell That test expect an error like

buildah push img1 dest1 -q
No options (-q) can be specified after the image or container name

What does your patched version of buildah do?

jwendell:
@rhatdan what I am saying is that even on master that test is failing.

rhatdan:
Works on my F28 box?

./tests/test_runner.sh push.bats
++ time bats --tap push.bats
1..4
ok 1 push-flags-order-verification
ok 2 push
ok 3 push with manifest type conversion
ok 4 push with imageid

real	0m14.028s
user	0m3.132s
sys	0m1.648s

jwendell:
@rhatdan you're right! What happened here was that I switched to master branch and re-run the test, and it also failed, because I didn't rebuild buildah binary. So, I assumed master was broken too :/

Anyway, just fixed the error and pushed again.

rhatdan:
@rh-atomic-bot r+

PR #1005, opened 2018-09-11
Use random string for getting a unique container name

### LINKED ISSUE #1004: Race in naming build containers causes build failures
**Description**
The concurrent buildah bud script in podman issue https://github.com/containers/libpod/issues/1146#issuecomment-420428372 sometimes fails to build an image and logs the following error:

`error creating build container: error creating container: the container name "centos-working-container-2" is already in use by "f5bda7beecec6896b09eea6d406d733199a4f8b98afc9150fdac4f883f59cdce". You have to remove that container to be able to reuse that name.: that name is already in use`

The code seems to have a race when choosing an incremented name suffix, between checking for a unique name and claiming that name, it may have been claimed by another buildah bud run.

1. Run http://paste.openstack.org/show/729907/ with concurrency=4
2. Keep running until the above error is logged (hoping you don't hit https://github.com/containers/libpod/issues/1146 )

buildah version 1.4-dev (image-spec 1.0.0, runtime-spec 1.0.0)
(local build of git master ref 08c56308c)

steveb:
Currently a buildah bud unique container name is found by incrementing
a counter and checking if that container exists. This causes a race
where that name may be claimed by another run by the time the
container is created.

This change replaces the incremented counter with a random string, so
that no duplicate name check is required.

Fixes #1004

Signed-off-by: Steve Baker <sbaker@redhat.com>

### DIFF
--- a/new.go
+++ b/new.go
@@ -12,6 +12,7 @@ import (
 	"github.com/containers/image/transports/alltransports"
 	"github.com/containers/image/types"
 	"github.com/containers/storage"
+	"github.com/containers/storage/pkg/stringid"
 	"github.com/opencontainers/selinux/go-selinux"
 	"github.com/opencontainers/selinux/go-selinux/label"
 	"github.com/openshift/imagebuilder"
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
Can you add some test output to see what the new container name looks like?

I really like the older numbering system, I wonder if we could throw up a lock around the old code instead of doing a randomid.  @nalind WDYT?

rhatdan:
@steveb Thanks for this PR.  @nalind Submitted a different version.  And we are going to use his.  Great find.

PR #1007, opened 2018-09-12
Evaluate symbolic links on Add/Copy Commands

rhatdan:
Users expect to be able to use Symbolic links in the
ADD and COPY commands in Dockerfiles.  We need to evaluate
these and pull in the correct content.

### DIFF
--- a/add.go
+++ b/add.go
@@ -168,9 +168,13 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 			return errors.Wrapf(syscall.ENOENT, "no files found matching %q", src)
 		}
 		for _, gsrc := range glob {
-			srcfi, err := os.Stat(gsrc)
+			esrc, err := filepath.EvalSymlinks(gsrc)
 			if err != nil {
-				return errors.Wrapf(err, "error reading %q", gsrc)
+				return errors.Wrapf(err, "error evaluating symlinks %q", gsrc)
+			}
+			srcfi, err := os.Stat(esrc)
+			if err != nil {
+				return errors.Wrapf(err, "error reading %q", esrc)
 			}
 			if srcfi.IsDir() {
 				// The source is a directory, so copy the contents of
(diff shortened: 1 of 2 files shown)


rhatdan:
Fixes https://github.com/projectatomic/buildah/issues/1006
@TomSweeneyRedHat @nalind @pixdrift @vrothberg PTAL

rhatdan:
https://github.com/Apkubatrh PTAL

rhatdan:
@umohnani8 @boaz1337 PTAL

tests/copy.bats
```diff
@@ -139,3 +139,22 @@ load helpers
...
+  root=$(buildah mount $cid)
+  buildah config --workingdir / $cid
+  buildah copy $cid ${TESTDIR}/link-randomfile
+  buildah unmount $cid
+  buildah commit --signature-policy ${TESTSDIR}/policy.json $cid containers-storage:new-image
+  buildah rm $cid
+
+  newcid=$(buildah from --signature-policy ${TESTSDIR}/policy.json new-image)
+  newroot=$(buildah mount $newcid)
+  test -s $newroot/randomfile
```
nalind: Shouldn't we expect `link-randomfile` to keep its name when it's copied?

rhatdan: I don't know, I will check with Docker.

rhatdan:
@nalind PTAL

add.go
```diff
@@ -168,10 +168,14 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 			return errors.Wrapf(syscall.ENOENT, "no files found matching %q", src)
 		}
 		for _, gsrc := range glob {
-			srcfi, err := os.Stat(gsrc)
+			esrc, err := filepath.EvalSymlinks(gsrc)
 			if err != nil {
 				return errors.Wrapf(err, "error reading %q", gsrc)
```
nalind: Please tweak this error message so that we can distinguish it from the next one.

tests/copy.bats
```diff
@@ -139,3 +139,22 @@ load helpers
...
+  root=$(buildah mount $cid)
+  buildah config --workingdir / $cid
+  buildah copy $cid ${TESTDIR}/link-randomfile
+  buildah unmount $cid
+  buildah commit --signature-policy ${TESTSDIR}/policy.json $cid containers-storage:new-image
+  buildah rm $cid
+
+  newcid=$(buildah from --signature-policy ${TESTSDIR}/policy.json new-image)
+  newroot=$(buildah mount $newcid)
+  test -s $newroot/link-randomfile
```
nalind: Please also ensure that the new `link-randomfile` isn't a symbolic link (using `test -h`).

rhatdan: Added test -f

nalind: That looks better than what I was thinking, actually.

nalind:
LGTM

umohnani8:
LGTM

rhatdan:
@rh-atomic-bot r=nalind

PR #1009, opened 2018-09-12
Handle ErrDuplicateName errors from store.CreateContainer()

nalind:
newBuilder(): When store.CreateContainer() returns a duplicate-name error for a name that we just made up, try again with a different made-up name.

### DIFF
--- a/new.go
+++ b/new.go
@@ -262,26 +262,23 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
 	if options.Container != "" {
 		name = options.Container
 	} else {
-		var err2 error
 		if image != "" {
 			name = imageNamePrefix(image) + "-" + name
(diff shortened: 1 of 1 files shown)


nalind:
This should fix part of https://github.com/containers/libpod/issues/1146#issuecomment-420428372.

TomSweeneyRedHat:
Does this replace #1005?

nalind:
Oof.  I hadn't realized, but it would.

TomSweeneyRedHat:
Change LGTM and I think I like this approach better then #1005 for the container name.  @steveb WDYT?

rhatdan:
LGTM
@rh-atomic-bot r+

steveb:
This fixes #1004 for me, thanks

Issue #1011, opened 2018-09-13
Buildah cannot be executed in the non-privileged container of k8s


q384566678:
I created a non-privileged container in k8s, but could not execute buildah. Later, I found that it could be executed correctly before this [patch](https://github.com/projectatomic/buildah/pull/836/commits/f1f087c8e51b3854a1ef2cb9127fe7e7a3f885cb) ,but it could not be executed after this patch modified the `unshare.go` file.

If I add privileges to the `yaml` file, I can execute buildah correctly.

Error info:
```
root@zz-ubuntu-78f657b855-ljrx8:~# buildah images
Error during unshare(CLONE_NEWUSER): Invalid argument
ERRO[0000] error parsing PID "": strconv.Atoi: parsing "": invalid syntax 
ERRO[0000] (unable to determine exit status)  
```

yaml file for creating non-privileged containers in k8s:
```
$ cat ubuntu.yaml 
apiVersion: apps/v1
kind: Deployment
metadata:
  name: zz-ubuntu
  labels: 
    label: ubuntu-1
spec:
  replicas: 1
  selector:
    matchLabels:
      app: my-ubuntu
  template:
    metadata:
      labels:
        app: my-ubuntu
    spec:
      containers:
      - name: test-ubuntu
        image: ubuntu:16.04
        command: ["sleep", "10000"]
```
Output of buildah version:
```
Version:         1.4-dev
Go Version:      go1.8.7
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.4.0
libcni Version:  v0.7.0-alpha1
Git Commit:      07aa077
Built:           Thu Sep 13 10:34:15 2018
OS/Arch:         linux/amd64
```

Output of uname -a:
```
Linux zz-ubuntu-78f657b855-ljrx8 3.10.0-862.11.6.el7.x86_64 #1 SMP Tue Aug 14 21:49:04 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

TomSweeneyRedHat:
@giuseppe PTAL

q384566678:
@nalind PTAL

giuseppe:
could you use the `chroot` isolation?  You can set it with the env variable `BUILDAH_ISOLATION=chroot`, or as a flag to `buildah run`

q384566678:
> could you use the chroot isolation? You can set it with the env variable BUILDAH_ISOLATION=chroot, or as a flag to buildah run

[...]

PR #1012, opened 2018-09-14
unshare: Fix running errors in non-privileged container

q384566678:
When running buildah in a non-privileged container in k8s, the following error occurs:

```
root# buildah images
Error during unshare(CLONE_NEWUSER): Invalid argument
ERRO[0000] error parsing PID "": strconv.Atoi: parsing "": invalid syntax
ERRO[0000] (unable to determine exit status)
```

This patch is to fix this error.

### DIFF
--- a/cmd/buildah/unshare.go
+++ b/cmd/buildah/unshare.go
@@ -16,7 +16,6 @@ import (
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"github.com/syndtr/gocapability/capability"
 	"github.com/urfave/cli"
 )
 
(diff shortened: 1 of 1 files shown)


q384566678:
Fixes #1011 
This is a modification method that I can think of, and there may be a better way.
@TomSweeneyRedHat @rhatdan @nalind @giuseppe PTAL

rhatdan:
@giuseppe PTAL

TomSweeneyRedHat:
Build/Tests issues hopefully fixed with #1020

q384566678:
@giuseppe PTAL

giuseppe:
why is `unshare` failing?  Is the kernel too old?

The tests we are dropping are valid ones, are you able to build a container once you drop these tests in a container?  It'd probably fail on `buildah run` if you haven't `CAP_SYS_ADMIN`

q384566678:
> Is the kernel too old?

Output of uname -a:
```
Linux zz-ubuntu-78f657b855-ljrx8 3.10.0-862.11.6.el7.x86_64 #1 SMP Tue Aug 14 21:49:04 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```
> It'd probably fail on buildah run if you haven't CAP_SYS_ADMIN

Only this one command can't be executed, but other commands can still be executed.

> are you able to build a container once you drop these tests in a container?

This modification guarantees that other commands can be executed, but the same problem will occur with the run command. I wonder if there are other solutions?

giuseppe:
> This modification guarantees that other commands can be executed, but the same problem will occur with the run command. I wonder if there are other solutions?

could you try to use the chroot isolation?  Does that work?

q384566678:
>could you try to use the chroot isolation? Does that work?


I tried it and it still went wrong.
```
root@zz-ubuntu-78f657b855-ljrx8:~# buildah run --isolation=chroot 747d sh
Error during unshare(CLONE_NEWUSER): Invalid argument
                                                     error parsing PID "": strconv.Atoi: parsing "": invalid syntax
```

giuseppe:
using the chroot isolation and having this patch solve the issue?

giuseppe:
In any case, CAP_SYS_ADMIN is required to setup the environment for launching a container.  If user namespaces are not available, you'll probably need a privileged container

theinrichs:
With the fix I don't run into the error. Buildah starts in a container but then exits with 
```
error running container: error creating new mount namespace for [/bin/sh -c apt-get update     && apt-get install -y curl unzip procps     && apt-get clean autoclean     && apt-get autoremove -y     && rm -rf /var/lib/{apt,dpkg,cache,log}/]: operation not permitted
```
This only works if I start buildah with `buildah --storage-driver=vfs bud --isolation=rootless` and a hostmount into the container.

rhatdan:
@theinrichs if you are running buildah inside of a container we are recommending that you use 
--isolation chroot
Rather then using Namespaces. Since you are already in a container, adding additional container separation adds little and causes the issues you are seeing.

theinrichs:
Thanks @rhatdan . --isolation chroot still gives me
`Error during unshare(CLONE_NEWUSER): Operation not permitted`

giuseppe:
That happens because the container has not enough capabilities and chroot as well needs them.  You should add CAP_SYS_ADMIN.

rhatdan:
@giuseppe I think chroot was created to not need CAP_SYS_ADMIN so we can run buildah inside of locked down containers.  I believe @nalind does this.

@q384566678 Looks like the tests passed?

rhatdan:
@q384566678 Were  you able to get buildah to run in k8s with --isolation  chroot without having to add SYS_ADMIN?  If yes can we close this PR.

q384566678:
> Were you able to get buildah to run in k8s with --isolation chroot without having to add SYS_ADMIN?

@rhatdan No, running the following command without modifying the code will still cause an error.

```
root@zz-ubuntu-78f657b855-ljrx8:~# buildah run --isolation=chroot 747d sh
Error during unshare(CLONE_NEWUSER): Invalid argument
ERRO[0000] error parsing PID "": strconv.Atoi: parsing "": invalid syntax 
ERRO[0000] (unable to determine exit status) 
```

rhatdan:
@giuseppe is this true ^^

giuseppe:
it creates the userns if it doesn't have enough capabilities.  Even the chroot isolation needs these capabilities to configure the `podman run` environment.  Creating an user namespace gives CAP_SYS_ADMIN limited to that namespace.  So either userNS are supported or the container should already have enough capabilities that a new userNS is not required.

rhatdan:
@giuseppe So you reject this PR?

giuseppe:
yes, I think it won't solve the problem of running buildah in a non-privileged container when user namespaces are not available.

lollipopman:
I am getting the same error trying to build as a non priveleged user on a debian box, is there a way to grant this capability to a user?:
```bash
~| uname -a
Linux blowpop 4.17.0-3-amd64 #1 SMP Debian 4.17.17-1 (2018-08-18) x86_64 GNU/Linux
~| buildah -v
buildah version 1.5-dev (image-spec 1.0.0, runtime-spec 1.0.0)
~| buildah from ${1:-fedora}
Error during unshare(CLONE_NEWUSER): Operation not permitted
ERRO[0000] error parsing PID "": strconv.Atoi: parsing "": invalid syntax 
ERRO[0000] (unable to determine exit status)
```

giuseppe:
@lollipopman on Ubuntu/Debian userNS for unprivileged users are disabled by default.  You need to enable user namespaces for unprivileged users with (as root) `echo 1 > /proc/sys/kernel/unprivileged_userns_clone`.

lollipopman:
@giuseppe thanks, that fixed it for me, it would be great if buildah could detect whether they are enabled and issue a better error message.

rhatdan:
@lollipopman Great idea.  Also we should update the man pages, and installation docs.

rhatdan:
@lollipopman Could you open an Issue to fix this.

q384566678:
>yes, I think it won't solve the problem of running buildah in a non-privileged container when user namespaces are not available.

So this is still an unresolved issue, we need to find a way to solve this problem.

lollipopman:
@rhatdan issue created, https://github.com/containers/buildah/issues/1068

TomSweeneyRedHat:
Thanks @lollipopman

rhatdan:
@q384566678 There is no way for a non priv user to create multiple UIDs without UserNS.  I think some tools have faked this out, but I am not sure it is worth hacking on this, since we have UserNS>

gvreddym:
is this issue resolved. I am getting same error on Openshift4 and Jenkins with i tried with below build command 

buildah bud --device /dev/fuse --isolation=chroot runtime=crun --format=oci --tls-verify=true --layers -f ${WORKSPACE}/Dockerfile .

Error
time="2021-07-27T07:06:44Z" level=warning msg="error reading allowed ID mappings: error reading subuid mappings for user \"1000200000\" and subgid mappings for group \"1000200000\": No subuid ranges found for user \"1000200000\" in /etc/subuid"
time="2021-07-27T07:06:44Z" level=warning msg="Found no UID ranges set aside for user \"1000200000\" in /etc/subuid."
time="2021-07-27T07:06:44Z" level=warning msg="Found no GID ranges set aside for user \"1000200000\" in /etc/subgid."
Error during unshare(CLONE_NEWUSER): Invalid argument
time="2021-07-27T07:06:49Z" level=error msg="error parsing PID \"\": strconv.Atoi: parsing \"\": invalid syntax"
time="2021-07-27T07:06:49Z" level=error msg="(unable to determine exit status)"

rhatdan:
Looks like openshift needs to be loosened up to run builds.  You might want to look at the openshift builder.  It uses Buildah under the covers.

PR #1021, opened 2018-09-17
nitpick: loop at parse.validateFlags in bud cli

boaz1337:


cmd/buildah/bud.go
```diff
@@ -115,14 +115,10 @@ func budCmd(c *cli.Context) error {
...
-		return err
-	}
-	if err := parse.ValidateFlags(c, buildahcli.LayerFlags); err != nil {
-		return err
-	}
-	if err := parse.ValidateFlags(c, buildahcli.FromAndBudFlags); err != nil {
-		return err
+	for _, flag := range [][]cli.Flag{buildahcli.BudFlags, buildahcli.LayerFlags, buildahcli.FromAndBudFlags} {
+		if err := parse.ValidateFlags(c, flag); err != nil {
+			return nil
```
TomSweeneyRedHat: Shouldn't we still be returning err here and not nil?  Beyond which even though this is more efficient, I think having them broken out like they were makes it easier to find where the issue was, but could be proven wrong on that.

TomSweeneyRedHat:
Typo in the commit message prase -> parse if you please.

cmd/buildah/bud.go
boaz1337: Hi @TomSweeneyRedHat 

Thanks for noticing about the `nil` thing. I put it there to check something and forgot to change it back.
I am not sure what you meant in the last sentence, can you be more specific?

Thanks again :+1:

cmd/buildah/bud.go
```diff
@@ -115,14 +115,10 @@ func budCmd(c *cli.Context) error {
...
-		return err
-	}
-	if err := parse.ValidateFlags(c, buildahcli.LayerFlags); err != nil {
-		return err
-	}
-	if err := parse.ValidateFlags(c, buildahcli.FromAndBudFlags); err != nil {
-		return err
+	for _, flag := range [][]cli.Flag{buildahcli.BudFlags, buildahcli.LayerFlags, buildahcli.FromAndBudFlags} {
+		if err := parse.ValidateFlags(c, flag); err != nil {
+			return err
```
TomSweeneyRedHat: @boaz1337 my only concern is with the old code, if we had an error it would say "Error line 122: {err}" or similar and we'd know that the failure happened when checking the LayerFlags.  With this change would we know which had an issue (BudFlags, LayerFlags or FromAndBudFlags)?  It shouldn't be a problem in production, most likely something we'd hit during development/testing.

boaz1337: Yeah, I see what you're saying.
Maybe adding index to the error, will help?
I don't mind closing this pull request if you want to.

TomSweeneyRedHat: The thing is I really like the change from a code efficiency standpoint.  Anyway to have the flag type that was being checked in the error....

rhatdan:
@rh-atomic-bot r+

PR #1029, opened 2018-09-18
Get rid of dangling whitespace in markdown files

nalind:
Get rid of dangling whitespace in markdown files, and add a quick validation check to error out if we try to add any new ones.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,7 @@ deps: gopath
 .PHONY: validate
 validate:
 	@./tests/validate/gofmt.sh
+	@./tests/validate/whitespace.sh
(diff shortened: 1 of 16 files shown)


rhatdan:
LGTM

nalind:
@TomSweeneyRedHat it doesn't break anything here, but it's something that `git-validation` would be doing for us if its `$GIT_CHECK_EXCLUDE` feature didn't depend on a newer version of git than we have by default on EL7.  (More on that at https://github.com/vbatts/git-validation/issues/37.)

rhatdan:
Travis just refuses to work. Force merging.

PR #1036, opened 2018-09-25
Clear intermediate container IDs after each stage

umohnani8:
Since we delete intermediate containers at the end of the whole build
process, we keep track of the intermediate container IDs.
In the case of a multi stage build, we need to clear the container IDs
before each stage to ensure we don't have a container ID repeated.
Before this, the container IDs were repeated and when we delete them
an error would be thrown as it would try to delete a container that
was already deleted.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -751,6 +751,7 @@ func (b *Executor) Execute(ctx context.Context, ib *imagebuilder.Builder, node *
 	checkForLayers := true
 	children := node.Children
 	commitName := b.output
+	b.containerIDs = nil
 	for i, node := range node.Children {
 		step := ib.Step()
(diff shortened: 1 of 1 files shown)


umohnani8:
@rhatdan @nalind PTAL

rhatdan:
LGTM

nalind:
LGTM.

rhatdan:
@rh-atomic-bot r+

umohnani8:
Looks like atomic-bot is stuck?

umohnani8:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot retry

rhatdan:
Looks like homu is hung.  Going to force merge.

PR #1041, opened 2018-09-27
Export buildah ReserveSELinuxLables so podman can use it

rhatdan:
We want to make sure that buildah and podman don't use the
same SELinux MCS Labels.  So we need to export this function
so libpod can use it.


rhatdan:
@TomSweeneyRedHat @umohnani8 @giuseppe @vrothberg @pixdrift PTAL

TomSweeneyRedHat:
LGTM

rhatdan:
@umohnani8 @giuseppe @pixdrift @boaz1337 @vrothberg @nalind PTAL

boaz1337:
good idea! LGTM :+1:

rhatdan:
@rh-atomic-bot r+

Issue #1047, opened 2018-09-28
buildah mishandles insecure registry config


bparees:
I have `172.30.255.124:5000` configured as an insecure registry in registries.conf.  That seems to work fine for pulling the blob as seen below, but then as soon as the dockerfile processing starts, a failure occurs because buildah tries to access the registry over https.

```
Pulling image 172.30.255.124:5000/openshift/ruby@sha256:a18c8706118a5c4c9f1adf045024d2abf06ba632b5674b23421019ee4d3edcae ...
Getting image source signatures
Copying blob sha256:256b176beaff7815db2a93ee2071621ae88f451bb1e198ca73010ed5bba79b65
 71.23 MiB / 71.23 MiB  3s
Copying blob sha256:efb19fea0fdb7178d5deedd98a3cbf90284dbc01f468e1dc8117b2d1becd34e9
 8.92 MiB / 8.92 MiB  0s
Copying blob sha256:b4570fdc208ced0033bf04a67323eda6cd72d92bcfab46bac200860a587722cc
 4.63 KiB / 4.63 KiB  0s
Copying blob sha256:213e4e250552a0d0e7e7c69ef31e783085154c9cf91cea80ed705b3eb5a35cc0
 157.87 KiB / 157.87 KiB  0s
Copying blob sha256:cb2ae87f9dc0eeb210670522c740de539310cbe565c8775f9438fc5a6dee56d4
 80.23 MiB / 80.23 MiB  3s
Copying blob sha256:8965056fa1335fd78743b362217a12f1e3f85a8f862378adadfa46248ee4551b
 13.18 MiB / 13.18 MiB  0s
Copying blob sha256:0c54dc45450aba434ffdbff89994d7cb5cd7f4fb074d284fb11f8d2239599b33
 1.82 KiB / 1.82 KiB  0s
Copying blob sha256:3834b16e2f5c97ee46bcb71b5a219f1205b906b09cb70cc0c0d8fbea34569c34
 3.92 KiB / 3.92 KiB  0s
Copying blob sha256:ff4625cd4ac9741afe593d4d03a0c4a0820b64bc679d68ae27d9b94f52402e02
 157.52 KiB / 157.52 KiB  0s
Copying config sha256:c7101e5b8b2912bb90834b291d4cc8a114da575e1bf9b9e305070cf2b395dd3d
 16.42 KiB / 16.42 KiB  0s
Writing manifest to image destination
Storing signatures
[...]

PR #1056, opened 2018-10-02
Correctly follow "insecure" and "blocked" settings when pulling images

nalind:
The image library's copy routine doesn't itself consult the registries configuration in order to decide whether or not to disable TLS verification when communicating with a registry, so it's on us to use the name of a source or destination image to decide whether to set the flag for that behavior.

While we're in here, wire up blocked registries.

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -92,6 +92,15 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 	var imgID string
 
 	systemContext := getSystemContext(options.SystemContext, options.SignaturePolicyPath)
+
+	blocked, err := isReferenceBlocked(dest, systemContext)
+	if err != nil {
+		return "", errors.Wrapf(err, "error checking if committing to registry for %q is blocked", transports.ImageName(dest))
+	}
+	if blocked {
+		return "", errors.Errorf("commit access to registry for %q is blocked by configuration", transports.ImageName(dest))
+	}
+
 	policy, err := signature.DefaultPolicy(systemContext)
 	if err != nil {
 		return imgID, errors.Wrapf(err, "error obtaining default signature policy")
@@ -162,6 +171,15 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 // Push copies the contents of the image to a new location.
 func Push(ctx context.Context, image string, dest types.ImageReference, options PushOptions) error {
 	systemContext := getSystemContext(options.SystemContext, options.SignaturePolicyPath)
+
+	blocked, err := isReferenceBlocked(dest, systemContext)
+	if err != nil {
+		return errors.Wrapf(err, "error checking if pushing to registry for %q is blocked", transports.ImageName(dest))
+	}
+	if blocked {
+		return errors.Errorf("push access to registry for %q is blocked by configuration", transports.ImageName(dest))
+	}
+
 	policy, err := signature.DefaultPolicy(systemContext)
 	if err != nil {
 		return errors.Wrapf(err, "error obtaining default signature policy")
--- a/commit.go
+++ b/commit.go
@@ -120,7 +120,7 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 		return imgID, errors.Wrapf(err, "error computing layer digests and building metadata")
 	}
 	// "Copy" our image to where it needs to be.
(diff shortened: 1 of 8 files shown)


nalind:
This should resolve part of #1047.

util.go
```diff
@@ -184,6 +184,58 @@ func getRegistries(sc *types.SystemContext) ([]string, error) {
 	return searchRegistries, nil
 }
 
+// isRegistryInsecure checks if the named registry is marked as not secure
+func isRegistryInsecure(registry string, sc *types.SystemContext) (bool, error) {
+	registries, err := sysregistriesv2.GetRegistries(sc)
+	if err != nil {
+		return false, errors.Wrapf(err, "unable to parse the registries.conf file")
```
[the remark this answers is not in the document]
nalind: It's not told to us by the `sysregistriesv2` package, but the `sysregistries` package offers a value, so we can report that.

util.go
```diff
@@ -184,6 +184,58 @@ func getRegistries(sc *types.SystemContext) ([]string, error) {
...
+		return reginfo.Insecure, nil
+	}
+	return false, nil
+}
+
+// isRegistryBlocked checks if the named registry is marked as blocked
+func isRegistryBlocked(registry string, sc *types.SystemContext) (bool, error) {
+	registries, err := sysregistriesv2.GetRegistries(sc)
+	if err != nil {
+		return false, errors.Wrapf(err, "unable to parse the registries.conf file")
```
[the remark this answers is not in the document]
nalind: It's not told to us by the `sysregistriesv2` package, but the `sysregistries` package offers a value, so we can report that.

nalind:
bot, retest this please

rhatdan:
LGTM
@TomSweeneyRedHat @umohnani8 @giuseppe @vrothberg PTAL

util.go
```diff
@@ -184,6 +185,58 @@ func getRegistries(sc *types.SystemContext) ([]string, error) {
...
+
+// isReferenceBlocked checks if the registry part of a reference is blocked
+func isReferenceBlocked(ref types.ImageReference, sc *types.SystemContext) (bool, error) {
+	if ref != nil && ref.Transport() != nil {
+		switch ref.Transport().Name() {
+		case "docker":
+			return isReferenceSomething(ref, sc, isRegistryBlocked)
+		}
+	}
+	return false, nil
```
TomSweeneyRedHat: Any value to a logrus.debug before the "return false, nil" lines in the above functions?

TomSweeneyRedHat:
LGTM, would like a head nod from @vrothberg

util.go
nalind: Maybe?  Adding it.

rhatdan:
LGTM

commit.go
```diff
@@ -92,6 +92,15 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 	var imgID string
 
 	systemContext := getSystemContext(options.SystemContext, options.SignaturePolicyPath)
+
+	blocked, err := isReferenceBlocked(dest, systemContext)
+	if err != nil {
+		return "", errors.Wrapf(err, "error checking if commiting to registry for %q is blocked", transports.ImageName(dest))
```
vrothberg: small typo s/commiting/committing/ (double t)

common.go
```diff
@@ -14,11 +17,35 @@ const (
 	DOCKER = "docker"
 )
 
-func getCopyOptions(reportWriter io.Writer, sourceSystemContext *types.SystemContext, destinationSystemContext *types.SystemContext, manifestType string) *cp.Options {
+func getCopyOptions(reportWriter io.Writer, sourceReference types.ImageReference, sourceSystemContext *types.SystemContext, destinationReference types.ImageReference, destinationSystemContext *types.SystemContext, manifestType string) *cp.Options {
+	sourceCtx := &types.SystemContext{}
+	if sourceSystemContext != nil {
+		*sourceCtx = *sourceSystemContext
```
vrothberg: This looks fishy and I think it may cause a segfault when being executed.

vrothberg:
LGTM (just the small typo in the error message)

commit.go
nalind: Good eye!  Fixed.

nalind:
Rebasing.

util.go
TomSweeneyRedHat: Well I was thinking just adding a debug before the "return false..." lines, but this works too!

rhatdan:
Did this get successfully merged?

PR #1067, opened 2018-10-04
parse: fix IsolationOption function

q384566678:
1.parse: modify the verification of the isolation value

To verify the value of the isolation, verify that all strings are the same, not just the beginning of the string.
Avoid users mistakenly thinking that the wrong value entered is also a correct value.

```
➜  buildah git:(isolation-fix) ✗ buildah from --isolation chrootcc cc510acfcd70
fedora-working-container-7
```

After Change:
```
➜  buildah git:(isolation-fix) ✗ ./buildah from --isolation chrootcc cc510acfcd70
unrecognized isolation type "chrootcc"
ERRO[0000] exit status 1
```

2.parse: Modify the return value

Code optimization and modification of return values.

### DIFF
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -568,7 +568,7 @@ func IsolationOption(c *cli.Context) (buildah.Isolation, error) {
 		case "chroot":
 			return buildah.IsolationChroot, nil
 		default:
-			return buildah.IsolationDefault, errors.Errorf("unrecognized isolation type %q", c.String("isolation"))
+			return 0, errors.Errorf("unrecognized isolation type %q", c.String("isolation"))
 		}
 	}
 	return defaultIsolation()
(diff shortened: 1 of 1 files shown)


vrothberg:
bot, retest this please

vrothberg:
Thanks for the PR! Overall LGTM. I am not a fan of mixing types (i.e., `Isolation` but returning `0`) but that's not a blocker given we don't have a `IsolationNone` or something similar.

TomSweeneyRedHat:
test failure seems to be a network error pulling alpine, going to retest.

TomSweeneyRedHat:
bot, retest this please.

TomSweeneyRedHat:
LGTM if we can get the test to turn green

pkg/parse/parse.go
```diff
@@ -543,11 +543,11 @@ func NamespaceOptions(c *cli.Context) (namespaceOptions buildah.NamespaceOptions
 func defaultIsolation() (buildah.Isolation, error) {
 	isolation, isSet := os.LookupEnv("BUILDAH_ISOLATION")
 	if isSet {
-		if strings.HasPrefix(strings.ToLower(isolation), "oci") {
+		if strings.ToLower(isolation) == "oci" {
```
rhatdan: Can we do this as a switch statement?

q384566678: updated, PTAL.

rhatdan:
@rh-atomic-bot r+

Issue #1068, opened 2018-10-04
Provide a better error message when user namespaces are not available


lollipopman:
**Description**

When user namespaces are not available buildah outputs the following error:
```
Error during unshare(CLONE_NEWUSER): Operation not permitted
ERRO[0000] error parsing PID "": strconv.Atoi: parsing "": invalid syntax 
ERRO[0000] (unable to determine exit status)
```
It would be great if the error message providing a better indication to the user about the nature of the error, and how it might be fixed, e.g.:
```
ERRO[0000] User namespaces are not enabled for unprivileged users on this
kernel, to enable run `echo 1 > /proc/sys/kernel/unprivileged_userns_clone`
as root.
```

**Steps to reproduce the issue:**
1. Ensure unprivileged user namespaces are disabled:
	```
	echo 0 > /proc/sys/kernel/unprivileged_userns_clone
	```
2. run buildah from
	```
	buildah from ${1:-fedora}
	```

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah/bionic 1.4-1~dev~ubuntu18.04~ppa5 amd64
buildah/bionic 1.4-1~dev~ubuntu18.04~ppa5 i386

```

**Output of `buildah version`:**

```
Version:         1.5-dev
Go Version:      go1.10.4
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.4.0
libcni Version:  v0.7.0-alpha1
Git Commit:      e9d18fc0
Built:           Tue Oct  2 12:18:57 2018
OS/Arch:         linux/amd64

```

**Output of `cat /etc/*release`:**

```
PRETTY_NAME="Debian GNU/Linux buster/sid"
NAME="Debian GNU/Linux"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

```

**Output of `uname -a`:**

```
Linux blowpop 4.17.0-3-amd64 #1 SMP Debian 4.17.17-1 (2018-08-18) x86_64 GNU/Linux

```

**Output of `cat /etc/containers/storage.conf`:**

```
# storage.conf is the configuration file for all tools
# that share the containers/storage libraries
# See man 5 containers-storage.conf for more information

# The "container storage" table contains all of the server options.
[storage]

# Default Storage Driver
driver = "overlay"

# Temporary storage location
runroot = "/var/run/containers/storage"

[...]

Issue #1061, opened 2018-10-03
unshare error during build


bparees:
see error at the end of this log.

```
2018-10-02T19:07:08.412805613Z FROM centos/s2i-base-centos7
2018-10-02T19:07:08.412808634Z ENV "BUILD_LOGLEVEL"="7"
2018-10-02T19:07:08.412811844Z EXPOSE 8080
2018-10-02T19:07:08.412816312Z ENV RUBY_VERSION=2.3
2018-10-02T19:07:08.412819803Z ENV SUMMARY="Platform for building and running Ruby $RUBY_VERSION applications" DESCRIPTION="Ruby $RUBY_VERSION available as container is a base platform for building and running various Ruby $RUBY_VERSION applications and frameworks. Ruby is the interpreted scripting language for quick and easy object-oriented programming. It has many features to process text files and to do system management tasks (as in Perl). It is simple, straight-forward, and extensible."
2018-10-02T19:07:08.412825073Z LABEL summary="$SUMMARY" description="$DESCRIPTION" io.k8s.description="$DESCRIPTION" io.k8s.display-name="Ruby 2.3" io.openshift.expose-services="8080:http" io.openshift.tags="builder,ruby,ruby23,rh-ruby23" com.redhat.component="rh-ruby23-container" name="centos/ruby-23-centos7" version="2.3" usage="s2i build https://github.com/sclorg/s2i-ruby-container.git --context-dir=2.4/test/puma-test-app/ centos/ruby-23-centos7 ruby-sample-app" maintainer="SoftwareCollections.org <sclorg@redhat.com>"
2018-10-02T19:07:08.412832185Z RUN yum install -y centos-release-scl &&     INSTALL_PKGS="rh-ruby23 rh-ruby23-ruby-devel rh-ruby23-rubygem-rake rh-ruby23-rubygem-bundler" &&     yum install -y --setopt=tsflags=nodocs $INSTALL_PKGS && rpm -V $INSTALL_PKGS &&     yum clean all -y
2018-10-02T19:07:08.412836438Z COPY ./s2i/bin/ $STI_SCRIPTS_PATH
2018-10-02T19:07:08.412867023Z COPY ./root/ /
2018-10-02T19:07:08.41287491Z RUN chown -R 1001:0 ${APP_ROOT} && chmod -R ug+rwx ${APP_ROOT} &&     rpm-file-permissions
2018-10-02T19:07:08.412878509Z USER 1001
[...]

Issue #1058, opened 2018-10-02
buildah chokes on dockerfile syntax


bparees:
```
2018-10-02T19:02:14.575036033Z Storing signatures
2018-10-02T19:02:46.852178787Z Generating dockerfile with builder image docker.io/bucharestgold/centos7-s2i-nodejs@sha256:a718a9daf8074aacd08e2de2679c067b39790f8e89e860f31d03e8fadd2d424a
2018-10-02T19:02:46.864097324Z Using buildah to create image temp.builder.openshift.io/e2e-test-prometheus-gj9m5/frontend-1:a8f3ad53
2018-10-02T19:02:46.864461829Z STEP 1: FROM docker.io/bucharestgold/centos7-s2i-nodejs@sha256:a718a9daf8074aacd08e2de2679c067b39790f8e89e860f31d03e8fadd2d424a
2018-10-02T19:02:46.886543093Z STEP 2: LABEL "io.openshift.build.image"="docker.io/bucharestgold/centos7-s2i-nodejs@sha256:a718a9daf8074aacd08e2de2679c067b39790f8e89e860f31d03e8fadd2d424a"       "io.openshift.build.commit.author"="Honza Horak <hhorak@redhat.com>"       "io.openshift.build.commit.date"="Tue Sep 25 07:47:27 2018 +0200"       "io.openshift.build.commit.id"="905416d88f74bfb734efac2a7e835948128a1df1"       "io.openshift.build.commit.ref"="master"       "io.openshift.build.commit.message"="Merge pull request #204 from jeffdyoung/ppc64le"       "io.openshift.build.source-location"="https://github.com/openshift/nodejs-ex.git"
2018-10-02T19:02:46.886662541Z STEP 3: ENV OPENSHIFT_BUILD_NAME="frontend-1"     OPENSHIFT_BUILD_NAMESPACE="e2e-test-prometheus-gj9m5"     OPENSHIFT_BUILD_SOURCE="https://github.com/openshift/nodejs-ex.git"     OPENSHIFT_BUILD_COMMIT="905416d88f74bfb734efac2a7e835948128a1df1"
2018-10-02T19:02:46.886673744Z STEP 4: USER root
2018-10-02T19:02:46.886677656Z STEP 5: COPY upload/src /tmp/src
2018-10-02T19:02:47.274455003Z STEP 6: RUN chown -R 1001:0 /tmp/src
[...]

PR #1069, opened 2018-10-04
chroot: use user namespaces only when necessary, check if they're enabled on error

nalind:
When running with chroot isolation, only create a new user namespace when we have mappings to set.

If we get an error when trying to create a new user namespace, check `/proc/sys/user/max_user_namespaces` to verify that user namespaces are even available.

### DIFF
--- a/unshare/unshare.c
+++ b/unshare/unshare.c
@@ -11,6 +11,8 @@
 #include <termios.h>
 #include <unistd.h>
 
+const char *_max_user_namespaces = "/proc/sys/user/max_user_namespaces";
+
 static int _buildah_unshare_parse_envint(const char *envname) {
 	char *p, *q;
 	long l;
@@ -31,7 +33,10 @@ static int _buildah_unshare_parse_envint(const char *envname) {
 
 void _buildah_unshare(void)
 {
+	FILE *fp;
 	int flags, pidfd, continuefd, n, pgrp, sid, ctty;
+	long max_userns;
+	size_t n_read;
 	char buf[2048];
 
 	flags = _buildah_unshare_parse_envint("_Buildah-unshare");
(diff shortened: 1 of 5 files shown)


nalind:
This should resolve #1061 and resolve #1068.

bparees:
may also be the resolution to https://github.com/containers/buildah/issues/1058 ?

nalind:
I think so, at least the bits where we run afoul of not being able to create user namespaces even in cases where we don't need to do any ID mapping.  The unit test that's failing in CI appears to be only reproducible when running under podman, still digging.

unshare/unshare.c
```diff
@@ -41,6 +46,22 @@ void _buildah_unshare(void)
 	if ((flags & CLONE_NEWUSER) != 0) {
 		if (unshare(CLONE_NEWUSER) == -1) {
 			fprintf(stderr, "Error during unshare(CLONE_NEWUSER): %m\n");
+			fp = fopen(_max_user_namespaces, "r");
+			if (fp != NULL) {
+				memset(buf, 0, sizeof(buf));
+				n_read = fread(buf, 1, sizeof(buf) - 1, fp);
+				if (n_read != 0) {
```
TomSweeneyRedHat: Personal nit, would you mind changing to `n_read > 0`.  Personal scars from using != in the past for zero checking.

nalind: Sure.

TomSweeneyRedHat:
@nalind looks like Red Hat CI isn't happy and it doesn't appear to be a flake.

rhatdan:
LGTM

giuseppe:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #1072, opened 2018-10-05
Attempt to log or return every error

nalind:
Make sure that when attempting to diagnose an error, if we encounter an error during the diagnostic attempt, we return the original error rather than the error encountered in trying to diagnose it.  Log that one.  Try to wrap more errors, and include more information when we do.


TomSweeneyRedHat:
Does this address a recent issue?  I thought we'd one, but I'm not seeing a relevant one on the issues list in a quick scan.

TomSweeneyRedHat:
@nalind looks like some legit test issues.

rhatdan:
@rh-atomic-bot r+

PR #1075, opened 2018-10-08
[WIP] Walk symlinks when checking cached images for copied/added files

umohnani8:
If a cache is found for a copied/added file, check for symlinks
and follow the symlink to get most up to date information of
file status.

Fixes https://github.com/containers/buildah/issues/1063#event-1889938970


imagebuildah/build.go
```diff
@@ -1017,7 +1017,19 @@ func (b *Executor) copiedFilesMatch(node *parser.Node, historyTime *time.Time) (
 		// For local files, walk the file tree and check the time stamps.
 		timeIsGreater := false
 		err := filepath.Walk(item, func(path string, info os.FileInfo, err error) error {
-			if info.ModTime().After(*historyTime) {
+			modTime := info.ModTime()
+			// resolve any symlinks that occur to get most up to date modified time.
+			if resolvedPath, err := resolveSymLink(b.contextDir, path); err == nil {
```
TomSweeneyRedHat: If the path is a file and not a symlink, does the resolvedPat get set to that file?  A comment in resolveSymlink says it returns "" if a symlink isn't found, but that might be a bad comment.

TomSweeneyRedHat:
Red Hat CI is failing, but it doesn't look related to this PR.

TomSweeneyRedHat:
bot, retest this please.

imagebuildah/build.go
rhatdan: I think you should first check to see if path is a symlink before calling resolveSymLink.

umohnani8: Yeah I am working on a patch to better approach this issue. Please ignore this PR till then.

rhatdan:
@umohnani8 Any progress?

umohnani8:
Fixed up the patch. @nalind @rhatdan @TomSweeneyRedHat  PTAL

rhatdan:
LGTM
@nalind @vrothberg @TomSweeneyRedHat @pixdrift @boaz1337 PTAL

TomSweeneyRedHat:
and how'd you manage to get the green butons, woulda thought the gofmt would have been unhappy on Go 11.   Hmmm, sometimes I just love Travis.

rhatdan:
@TomSweeneyRedHat modifiedtimes is the goal of this PR, and I think that was moved from another section of code.  
The bug that this is trying to fix, is when a user is building with --layers, the code is supposed to figure out whether or not the content used in the Dockerfile has changed since the last run.  If the user setup a symbolic link to content, then we would check if the symbolic link changed rather then the content pointed to by the symbolic link.

umohnani8:
@TomSweeneyRedHat yeah the goal is to compare the modified time of a file to the creation time of the cached intermediate image to determine whether to use cache or not. Initially, we were not resolving any symlinks if the file being copied/added was a symlink. This just does that as well as compare the modified time.

TomSweeneyRedHat:
@umohnani8 and @rhatdan tyvm for the background info, that makes sense to me now.
LGTM

imagebuildah/chroot_symlink.go
```diff
@@ -62,32 +65,118 @@ func ResolveSymLink(rootdir, filename string) (string, error) {
 		return "", errors.Wrapf(err, string(output))
 	}
 
-	// Hand back the resolved symlink, will be "" if a symlink is not found
+	// Hand back the resolved symlink, will be filename if a symlink is not found
 	return string(output), nil
 }
 
+func resolveSymlinkTimeModified() {
```
vrothberg: Non-blocker: can we document this function? Having a function without argument and return values is rare.

imagebuildah/chroot_symlink.go
```diff
@@ -145,3 +239,11 @@ func hasSymlink(path string) (bool, string, error) {
 	// run filepath.Clean to remove the ".." from relative paths
 	return true, filepath.Clean(targetDir), nil
 }
+
+// boolToString converts a boolean to its equivalent string form
```
vrothberg: Is there a reason to not use `fmt.*rintf("%v", bool)`?

vrothberg:
Just small nitpicks but LGTM, really nice!

imagebuildah/chroot_symlink.go
umohnani8: Nope, just didn't think about it. Thanks for the tip! :)

imagebuildah/chroot_symlink.go
umohnani8: Added comment before function. Fixed.

rhatdan:
LGTM
@rh-atomic-bot r+

PR #1076, opened 2018-10-09
chroot: expose ResolveSymLink

giuseppe:
make ResolveSymLink public so that it can be used by other packages.


imagebuildah/chroot_symlink.go
```diff
@@ -24,7 +24,7 @@ func init() {
 func resolveChrootedSymlinks() {
 	status := 0
 	flag.Parse()
-	if len(flag.Args()) < 1 {
+	if len(flag.Args()) < 2 {
```
TomSweeneyRedHat: Not sure why you made this change?  Might be good to add a comment in the description section of the PR.

imagebuildah/chroot_symlink.go
```diff
@@ -50,7 +50,7 @@ func resolveChrootedSymlinks() {
 	os.Exit(status)
 }
 
-func resolveSymLink(rootdir, filename string) (string, error) {
+func ResolveSymLink(rootdir, filename string) (string, error) {
```
TomSweeneyRedHat: I thought golang wanted a comment line above a public function like this, but it doesn't seem to be barking about it.

rhatdan: It does, the linter will fail.

imagebuildah/chroot_symlink.go
rhatdan: The description of the PR says why
`two arguments are required by the program`

rhatdan: ALthough I think doing logrus.error would be helpful in this case before exiting.

giuseppe:
travis tests are passing, PAPR seems to be in a bad state:

```

'/etc/yum.repos.d.host/fedora-cisco-openh264.repo' -> '/etc/yum.repos.d/fedora-cisco-openh264.repo'
'/etc/yum.repos.d.host/fedora-updates-testing.repo' -> '/etc/yum.repos.d/fedora-updates-testing.repo'
'/etc/yum.repos.d.host/fedora-updates.repo' -> '/etc/yum.repos.d/fedora-updates.repo'
'/etc/yum.repos.d.host/fedora.repo' -> '/etc/yum.repos.d/fedora.repo'
sh: /code/.papr.sh: Permission denied
```

imagebuildah/chroot_symlink.go
rhatdan: @giuseppe While you clean this up, could you add a logrus.Error message below.

rhatdan:
One nit on a error message, LGTM.

imagebuildah/chroot_symlink.go
giuseppe: pushed a new version with an added error message

giuseppe:
travis is passing now, papr is still broken but I don't think it depends from this PR

giuseppe:
this is blocking: https://github.com/containers/libpod/pull/1609

giuseppe:
tests are passing

rhatdan:
@rh-atomic-bot r+

PR #1077, opened 2018-10-09
buildah: use the same logic for XDG_RUNTIME_DIR as podman

giuseppe:
if XDG_RUNTIME_DIR is not set, try in the order:

- /var/run/user/$UID/run
- $HOME/rundir

also set the XDG_RUNTIME_DIR so that the OCI runtime will
use the same settings.

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"github.com/containers/image/directory"
 	dockerarchive "github.com/containers/image/docker/archive"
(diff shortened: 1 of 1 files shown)


util/util.go
```diff
@@ -472,11 +473,31 @@ func UnsharedRunrootPath(uid string) (string, error) {
 	if envRuntimeDir, haveRuntimeDir := os.LookupEnv("XDG_RUNTIME_DIR"); haveRuntimeDir {
 		return filepath.Join(envRuntimeDir, "run"), nil
 	}
+	var runtimeDir string
 	// If $XDG_RUNTIME_DIR is not defined, but we know our UID...
 	if uid != "" {
-		return filepath.Join("/var/run/user", uid, "run"), nil
+		tmpDir := filepath.Join("/var/run/user", uid, "run")
```
TomSweeneyRedHat: Podman doesn't seem to have "run" on the end, is this correct or did I misread the podman code?

TomSweeneyRedHat:
Red Hat CI isn't happy.  I think, but don't know that it's a flake.

util/util.go
giuseppe: I've left this part similar to the previous code in Buildah.  Should I drop the last "run"?

/cc @nalind

giuseppe: thinking more of it, it seems correct the version without the "run" so I've pushed a new version now.  Thanks for catching this issue!

rhatdan:
bot, retest this please

vbatts:
tested and it works on centos7

rhatdan:
@jlebon any idea what is going on here with papr?

TomSweeneyRedHat:
LGTM, assuming happy tests.

jlebon:
Infra is currently experiencing difficulties. Ops is investigating.

rhatdan:
bot, retest this please

vrothberg:
Once https://github.com/containers/buildah/pull/1090 is merged, tests will be calm again.

vrothberg:
@giuseppe, you can now rebase and test should be fine again

giuseppe:
rebased :arrow_up:

rhatdan:
@rh-atomic-bot r+

PR #1085, opened 2018-10-11
unshare: detect when unprivileged userns are disabled

giuseppe:
On Debian/Ubuntu it is possible to disable unprivileged userns through
the /proc/sys/kernel/unprivileged_userns_clone file.

Detect such case when unshare(CLONE_NEWUSER) fails.

Closes: https://github.com/containers/buildah/issues/1068

### DIFF
--- a/unshare/unshare.c
+++ b/unshare/unshare.c
@@ -9,9 +9,11 @@
 #include <stdlib.h>
 #include <string.h>
 #include <termios.h>
+#include <errno.h>
 #include <unistd.h>
 
-const char *_max_user_namespaces = "/proc/sys/user/max_user_namespaces";
+static const char *_max_user_namespaces = "/proc/sys/user/max_user_namespaces";
+static const char *_unprivileged_user_namespaces = "/proc/sys/kernel/unprivileged_userns_clone";
 
 static int _buildah_unshare_parse_envint(const char *envname) {
 	char *p, *q;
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
Test issues aren't related, I belive @jlebon is still chasing papr fun.

jlebon:
Hmm, PAPR infra should be back up.
Checking the logs I see:

```
sh: /code/.papr.sh: Permission denied
### EXITED WITH CODE 126 AFTER 19s
```

Hypothesis: the [latest FAH28 release](https://lists.projectatomic.io/projectatomic-archives/atomic-devel/2018-October/msg00002.html) includes:

```
podman 0.9.1-3.gitaba58d1.fc28 -> 0.9.3.1-1.git1cd906d.fc28
```

Does that newer podman version suffer from any regressions SELinux-wise? Sounds like it could be https://github.com/containers/libpod/issues/1575, which was fixed by https://github.com/containers/libpod/pull/1576, which is only in master for now.

One way to verify this is to e.g. disable SELinux and see if it passes, or build podman from git master, or fetch prebuilt packages with the patch above and `ostree admin unlock && rpm -Uvh ...`.

TomSweeneyRedHat:
@mheon thoughts on @jlebon's notes?  @jlebon tyvm!

mheon:
Sounds entirely possible.

I think it's possible to downgrade with an overlay in rpm-ostree, so that might be a good option for now. We are hopefully cutting a new Podman release today, so the issue should resolve itself soon.

rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot retry

PR #1090, opened 2018-10-12
fix papr runs

vrothberg:

### DIFF
--- a/.papr.sh
+++ b/.papr.sh
@@ -1,11 +1,10 @@
 #!/bin/bash
 set -xeuo pipefail
+export GOPATH=/go
+export PATH=$HOME/gopath/bin:$PATH:$GOPATH/bin
+export GOSRC=$GOPATH/src/github.com/containers/buildah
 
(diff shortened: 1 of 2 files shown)


vrothberg:
@giuseppe PTAL. The recent hiccups from libpod are now here as well, but the change seems to fix it.

.papr.sh
```diff
@@ -1,11 +1,10 @@
 #!/bin/bash
 set -xeuo pipefail
+export GOPATH=/go
+export PATH=$HOME/gopath/bin:$PATH:$GOPATH/bin
+export GOSRC=$GOPATH/src/github.com/containers/libpod
```
rhatdan: libpod or buildah?

vrothberg: I like both, but it should be buildah. Thanks for checking!

giuseppe:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #1091, opened 2018-10-12
Podman  --privileged selinux is broken

rhatdan:
Current podman does not disable SELinux when running in privileged mode

This PR Should fix this until we get an updated version of podman out.


TomSweeneyRedHat:
LGTM assuming happy tests

rhatdan:
Going to force merge since this PR Has nothing to do with Travis, and I think we see other breakage caused by this.

PR #1097, opened 2018-10-15
Update github.com/containers/image

nalind:
Update `github.com/containers/image` to version 5e5b67d6b1cf43cc349128ec3ed7d5283a6cc0d1, which returns the new image's manifest when we copy an image from one place to another.  Return the digest of that manifest, and a canonical reference, from our `Commit` and `Push` APIs.


cmd/buildah/bud.go
```diff
@@ -233,5 +233,6 @@ func budCmd(c *cli.Context) error {
 		options.ReportWriter = ioutil.Discard
 	}
 
-	return imagebuildah.BuildDockerfiles(getContext(), store, options, dockerfiles...)
+	_, _, err = imagebuildah.BuildDockerfiles(getContext(), store, options, dockerfiles...)
```
TomSweeneyRedHat: Podman calls this, when we merge this there in we'll need to adjust calls there too @mheon

PR #1099, opened 2018-10-16
run: When the value of isolation is set, use the set value instead of the default value

q384566678:
Before change:
```
➜  buildah git:(from-isolation) ✗ buildah from --isolation oci 1650
ubuntu-working-container-2
➜  buildah git:(from-isolation) ✗ buildah run ubuntu-working-container-2 bash
root@zz:/# exit
exit
```
As you can see from the above example, the isolation set in the `from` command has no effect, because the default value will still be used when running the `run` command.
So when the value of isolation is set, the set value should be used instead of the default value.

After Change:
```
➜  buildah git:(from-isolation) ✗ ./buildah run ubuntu-working-container-2 bash
cannot specify gid= mount options for unmapped gid in rootless containers
error running container: error creating container for [/bin/bash]: : exit status 1
ERRO[0000] exit status 1
```

### DIFF
--- a/cmd/buildah/run.go
+++ b/cmd/buildah/run.go
@@ -102,9 +102,12 @@ func runCmd(c *cli.Context) error {
 		return errors.Wrapf(err, "error reading build container %q", name)
 	}
 
-	isolation, err := parse.IsolationOption(c)
-	if err != nil {
-		return err
+	isolation := builder.Isolation
+	if c.IsSet("isolation") {
+		isolation, err = parse.IsolationOption(c)
+		if err != nil {
+			return err
+		}
 	}
 
 	runtimeFlags := []string{}


rhatdan:
LGTM

TomSweeneyRedHat:
Change LGTM, but tests aren't happy.  I'm not sure what's up test wise, might be a problem not related to this.

TomSweeneyRedHat:
#1102 should fix the test issues here.  Most likely will need a rebase after that is merged.

TomSweeneyRedHat:
Ooops #1101 should fix the test issues, not #1102

TomSweeneyRedHat:
@nalind @rhatdan this is also showing a gofmt error only on the Go 1.11 test like #1103

rhatdan:
Can we just remove validate from travis and rely on .papr for these tests?

TomSweeneyRedHat:
@q384566678 can you rebase this please, I think our test system is finally happy.

q384566678:
Rebased, PTAL.

q384566678:
Rebased, PTAL.

q384566678:
Sorry for replying some duplicate content due to the network, it has been deleted.

TomSweeneyRedHat:
LGTM and happy green test buttons finally.
@q384566678 thanks for working through all the non-related test CI issues you ran into doing this PR.

rhatdan:
@rh-atomic-bot r+

rhatdan:
Since Github is not working, merging manually.

PR #1100, opened 2018-10-16
Builder: working container: assign random suffix

vrothberg:
Assign random suffixes to the working containers instead of using an
iterative sequence starting at `1`.  This speeds up the creation of
working containers as the probability of naming conflicts is reduced.
An error is returned after 100 failed attempts to find an unused
name/suffix.

### DIFF
--- a/new.go
+++ b/new.go
@@ -3,6 +3,7 @@ package buildah
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"strings"
 
 	"github.com/containers/buildah/util"
@@ -281,17 +282,18 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
 	coptions := storage.ContainerOptions{}
 	coptions.IDMappingOptions = newContainerIDMappingOptions(options.IDMappingOptions)
 
+	attempts := 0
+	name = fmt.Sprintf("%s-%d", name, rand.Int())
 	container, err := store.CreateContainer("", []string{name}, imageID, "", "", &coptions)
-	suffix := 1
 	for err != nil && errors.Cause(err) == storage.ErrDuplicateName && options.Container == "" {
-		suffix++
-		tmpName := fmt.Sprintf("%s-%d", name, suffix)
-		if container, err = store.CreateContainer("", []string{tmpName}, imageID, "", "", &coptions); err == nil {
-			name = tmpName
+		attempts++
+		name = fmt.Sprintf("%s-%d", name, rand.Int())
+		if container, err = store.CreateContainer("", []string{name}, imageID, "", "", &coptions); err == nil {
+			break
 		}
-	}
-	if err != nil {
-		return nil, errors.Wrapf(err, "error creating container")
+		if attempts > 99 {
+			return nil, fmt.Errorf("error creating container: reached maximum of 100 attempts to create working container")
+		}
 	}
 
 	defer func() {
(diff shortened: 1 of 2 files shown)


new.go
```diff
@@ -282,13 +283,19 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
 	coptions.IDMappingOptions = newContainerIDMappingOptions(options.IDMappingOptions)
 
 	container, err := store.CreateContainer("", []string{name}, imageID, "", "", &coptions)
```
rhatdan: I think the suffix here should be random as well.  Otherwise you end up with a delay on every container after the first image.

new.go
```diff
@@ -282,13 +283,19 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
...
+	tries := 0
 	for err != nil && errors.Cause(err) == storage.ErrDuplicateName && options.Container == "" {
-		suffix++
+		tries++
 		tmpName := fmt.Sprintf("%s-%d", name, suffix)
 		if container, err = store.CreateContainer("", []string{tmpName}, imageID, "", "", &coptions); err == nil {
 			name = tmpName
 		}
+		if tries == 100 {
+			logrus.Debug("reached maximum of 100 ties to create working container")
```
rhatdan: This should probably be a return err, so that the  user knows it was attempted 100 times.

new.go
vrothberg: Thanks, I completely focused on the loop below. I'll fix that.

new.go
vrothberg: I'll fix that and will also change the 100 *ties* to 100 attempts :laughing:

vrothberg:
For one reason or another, Travis has problems firing the tests. Other PRs might be affected as well (Cc @TomSweeneyRedHat ).

TomSweeneyRedHat:
@vrothberg yep, just opened #1101, think I've a cure almost ready.

new.go
```diff
@@ -281,17 +282,18 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
...
-		suffix++
-		tmpName := fmt.Sprintf("%s-%d", name, suffix)
-		if container, err = store.CreateContainer("", []string{tmpName}, imageID, "", "", &coptions); err == nil {
-			name = tmpName
+		attempts++
+		name = fmt.Sprintf("%s-%d", name, rand.Int())
+		if container, err = store.CreateContainer("", []string{name}, imageID, "", "", &coptions); err == nil {
+			break
+		}
+		if attempts == 100 {
```
TomSweeneyRedHat: Could we change this to '> 99', just to be ultra safe.

TomSweeneyRedHat: I'm hemming and hawing.  Is it worth having an environment variable for "100" so the value could be over-riidden if we wanted to make it larger/smaller in production?  IDK, thought I'd solicit other opinions.

vrothberg:
> One nit and a question, otherwise LGTM. The intro tutorial (docs/tutorials/01-intro.md) talks about the "working-container" being returned by default. We might want to tweak that too.

@TomSweeneyRedHat, good idea!

new.go
vrothberg: Will change.

vrothberg: I am not generally opposed to the idea but think we're really safe with 100 random guesses.

TomSweeneyRedHat:
bot, retest this please.

TomSweeneyRedHat:
@vrothberg I think a rebase will clean this up.

vrothberg:
Closing as it is in conflict with the already merged https://github.com/containers/buildah/pull/1102.

PR #1102, opened 2018-10-16
Moving selinux reservations into container storage.

rhatdan:


new.go
```diff
@@ -351,7 +344,7 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
 	}
 
 	if options.Mount {
-		_, err = builder.Mount(mountLabel)
+		_, err = builder.Mount(container.MountLabel())
```
nalind: Does the underlying `Store` handle this for us if we pass "" as the label?

rhatdan:
@nalind This now seems to work from a quick desk check.

new.go
rhatdan: Yes

new.go
```diff
@@ -278,14 +277,16 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
 		}
 	}
 
-	coptions := storage.ContainerOptions{}
-	coptions.IDMappingOptions = newContainerIDMappingOptions(options.IDMappingOptions)
-
+	coptions := storage.ContainerOptions{
+		LabelOpts:        options.CommonBuildOpts.LabelOpts,
+		IDMappingOptions: newContainerIDMappingOptions(options.IDMappingOptions),
+	}
```
nalind: Might as well move this definition into the loop to have it create a new `ContainerOptions` for each call to `CreateContainer()`.

rhatdan:
@nalind @TomSweeneyRedHat @vrothberg @giuseppe This is ready for review.

new.go
```diff
@@ -232,6 +231,27 @@ func resolveImage(ctx context.Context, systemContext *types.SystemContext, store
...
+func containerNameExist(name string, containers []storage.Container) bool {
+	for _, container := range containers {
+		for _, cname := range container.Names {
+			if cname == name {
+				return true
+			}
+		}
+	}
+	return false
+}
```
vrothberg: The lookup is quite expensive with O(n). It comes from storage, but I want to drop the idea here. Maybe, we can come up with a data type (in storage) that abstracts from such tasks, e.g.,
```golang
type Containers struct {
    ctr_slice []container
    ctr_map[string[container
}
```

vrothberg: But certainly not in this PR. I just wanted to use this example to kick off a short discussion.

vrothberg:
The second commit hides a bit the change to the name selection (conflicting with https://github.com/containers/buildah/pull/1100). I am not sure if that is actually superior to https://github.com/containers/buildah/pull/1100 when doing parallel builds where name-conflicts will still appear. That's why I prefer assigning random suffixes as proposed in https://github.com/containers/buildah/pull/1100.

new.go
```diff
@@ -277,23 +297,33 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
...
-	for err != nil && errors.Cause(err) == storage.ErrDuplicateName && options.Container == "" {
-		suffix++
-		tmpName := fmt.Sprintf("%s-%d", name, suffix)
-		if container, err = store.CreateContainer("", []string{tmpName}, imageID, "", "", &coptions); err == nil {
+	for true {
+		coptions := storage.ContainerOptions{
+			LabelOpts:        options.CommonBuildOpts.LabelOpts,
+			IDMappingOptions: newContainerIDMappingOptions(options.IDMappingOptions),
+		}
+		container, err = store.CreateContainer("", []string{tmpName}, imageID, "", "", &coptions)
```
vrothberg: Name conflicts would still appear here in parallel builds. I am not really not found of the incremental suffixes. They look much nicer but it's really expensive when hitting this condition.

vrothberg:
If you want to test it, you can use the following bash script. Run two instances in parallel and compare it to https://github.com/containers/buildah/pull/1100.

```bash
#!/bin/bash
for i in `seq 1 20`;               
do                                 
        sudo ./buildah from scratch &
done
```

new.go
TomSweeneyRedHat: I'd love to have a map available to lookup containers from, I'm just not sure how we create one efficiently that will be used enough times to justify the cost.  But a good discussion to have in case there's a way to do so.

TomSweeneyRedHat:
I've nothing beyond @vrothberg and @nalind's comments.  Look like you've a merge conflict now that needs to be tended to @rhatdan

new.go
rhatdan: Well the conflict is only going to happen for containers that are created since the buildah command is executed.  Only Buildah uses this algorythm.  A conflict would only happen if a second buildah rand creating a container since the first container read the container store.  Unlike the original code this buildah code will only try on a known `good` name based on the list of containers that were created at start.

rhatdan:
Ok I got a little creative, trying to keep the best of both worlds, so you both will probably hate it.
Try @nalind version first, then if conflict, try a random suffix from 1-100,if this fails, go to random 1-1000, if this fails go to 1-10000 ...

Trying the script above I am seeing around 4 seconds to create the 20 containers.
50 Seconds to create 200 containers.

vrothberg:
I am still not a big fan of it but I also want to avoid that we burn too much time on this, so I will drop my final thoughts. There are two things that I don't really like about the proposed changes.

1) The incremental naming is expensive with O(n) finding the initial suffix. As we cannot lock the storage, we may run into another naming loop lifting it to O(n^2) plus getting the list of all containers. Using random suffixes makes it O(1) at the cost of less appealing suffixes.

2) The logic of creating and naming a container should go into containers/storage e.g., `CreateContainerWithSuffix(...)` that resolves conflicts in a similar fashion. The benefit is that we could use locks and avoid the upper problems altogether. Another idea could also be to use the short ID of the container as a suffix.

Regarding execution times, the costs mentioned in 1) are already measurable without having parallel builds. Creating 20 containers with this change takes ~8 seconds on my machine while using random suffixes takes ~5 seconds. My fear is that if we had to speed things up in the future, this is likely to be a code path that profilers would highlight.

rhatdan:
@vrothberg Lets continue to debate the startup time in a separate place then this PR.  This PR should make it better, but not ideal, and this PR is mainly about the SELinux stuff,  which I think everyone approves.

rhatdan:
Since github.com is not working properly I am going to manually merge.

PR #1109, opened 2018-10-17
ReserveSELinuxLabels(): handle wrapped errors from OpenBuilder()

nalind:
`ReserveSELinuxLabels()` checks if an error returned by `OpenBuilder()` is a does-not-exist error, but `OpenBuilder()` returns wrapped errors now, and it wasn't checking the root cause error.

When `newBuilder()` fails, check the right error value when deciding whether or not deleting the partially-constructed container failed.

`OpenBuildersByPath()` shouldn't choke on non-buildah containers, so have it handle does-not-exist errors the same way `OpenAllBuilders()` does.

### DIFF
--- a/buildah.go
+++ b/buildah.go
@@ -457,6 +457,10 @@ func OpenBuilderByPath(store storage.Store, path string) (*Builder, error) {
 		}
 		buildstate, err := ioutil.ReadFile(filepath.Join(cdir, stateFile))
 		if err != nil {
+			if os.IsNotExist(err) {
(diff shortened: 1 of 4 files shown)


nalind:
@TomSweeneyRedHat, @rhatdan PTAL

rhatdan:
LGTM

baude:
LGTM

mheon:
LGTM - solves a nasty Buildah/Podman interaction bug we've been seeing in our CI

rhatdan:
@rh-atomic-bot r+

PR #1110, opened 2018-10-17
Run gofmt only on Go 1.10

TomSweeneyRedHat:

I think this will work to run gofmt only when go is version 1.10.  I'd like to have this in travis.yml, but couldn't figure a clean way to do so.  Any other suggestions/comments gratefully accepted.

Once working, this should cure our go 1.11 CI issues.


TomSweeneyRedHat:
This is failing with a syntax error, I've gone through a number of loops.  @rhatdan or @nalind any thoughts on what bit is out of place?

```
HEAD is now at 2c632d1a Merge pull request #1824 from cyphar/fix-mips-build-devNumber
ln -sf ../../opencontainers/runc/runc
# Run gofmt only on version 1.10
ifneq (,)
/bin/sh: 1: Syntax error: word unexpected (expecting ")")
make: *** [validate] Error 2
```

Makefile
```diff
@@ -5,6 +5,7 @@ BINDIR := $(PREFIX)/bin
 BASHINSTALLDIR=${PREFIX}/share/bash-completion/completions
 BUILDFLAGS := -tags "$(AUTOTAGS) $(TAGS)"
 GO := go
+GOVERSION := go version
```
rhatdan: GOVERSION := $(shell go version)

Makefile
```diff
@@ -48,7 +49,10 @@ deps: gopath
 
 .PHONY: validate
 validate:
-	@./tests/validate/gofmt.sh
+	# Run gofmt only on version 1.10
+	ifneq (,$(findstring 1.10,$(GOVERSION)))
```
rhatdan: We should only run this on 1.11?

TomSweeneyRedHat: gofmt I thought was failing on 1.11, not 1.10, but regardless, I'm still getting a syntax error on this line.

Makefile
TomSweeneyRedHat: fixed, thx

Makefile
rhatdan: Can we fix our code to confirm with what golang 1.11 wants, or is golang 1.11 just broken?

TomSweeneyRedHat:
@q384566678 tried to get the code happy for golang 1.11, but that ended up not working.  I'll try working from that angle too.

TomSweeneyRedHat:
Ugh, I love make, the issue was the if statement was indented, it needed to be all the way to the left otherwise make treats it like a shell script.

TomSweeneyRedHat:
@umohnani8 FYI

rhatdan:
@rh-atomic-bot r+

PR #1111, opened 2018-10-18
build: Fix code format

q384566678:


q384566678:
I raised the version of go to 1.11. I found that the results of these two versions of `gofmt` are not the same. I think this is better, so I changed the format of the code.
But the test results are wrong, we should make a trade-off between 1.10 and 1.11.

TomSweeneyRedHat:
@q384566678 Thanks for the attempt, but this is still having gofmt issues, in fact it's now having them in 1.10 and 1.11.  #1110 is meant to fix the gofmt issue for at least the short term.  However that's having some Makefile syntax issues atm that I'm trying to work through.

TomSweeneyRedHat:
@q384566678 I'm going to close this as I've this in my #1114 PR along with changes in the Makefile to keep other issues from popping up.  Thanks for the PR, it made the diagnosis of the issue much easier.

PR #1114, opened 2018-10-18
Change gofmt so it runs on all but 1.10

TomSweeneyRedHat:

Flip the logic from #1110.  Rather than running gofmt only on 1.11, run on any version that is not 1.10.  gofmt should run on version 1.11 and higher and we currently only have 1.10 and higher in the Test CI system.

Also remove the gofmt for the Darwin CI which was failing and in truth superfluous.


rhatdan:
@rh-atomic-bot r+

Issue #1117, opened 2018-10-18
buildah from fail when using an explicit docker: transport


robmv:
**Description**

`buildah from` fail when using an explicit `docker:` transport. This is a regression, probably caused by  the introduction of `buildah pull`, It worked before 1.4 release.

**Steps to reproduce the issue:**
1. Run `buildah from docker://centos:centos7` or `buildah from docker://docker.io/centos:centos7`

**Describe the results you received:**

Both cases fail with the error `invalid reference format` error when the `docker:` transport is explicit.

**Describe the results you expected:**

It should work as when the `docker:` transport is implicit as `buildah pull docker://centos:centos7` and `buildah pull docker://docker.io/centos:centos7` works perfectly.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.4-3.git608fa84.fc28.x86_64
```

**Output of `buildah version`:**

```
Version:         1.4
Go Version:      go1.10.4
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.4.0
libcni Version:  v0.7.0-alpha1
Git Commit:      608fa84
Built:           Thu Oct  4 12:40:10 2018
OS/Arch:         linux/amd64
```

**Output of `cat /etc/*release`:**

```
Fedora release 28 (Twenty Eight)
NAME=Fedora
VERSION="28 (Workstation Edition)"
ID=fedora
VERSION_ID=28
PLATFORM_ID="platform:f28"
PRETTY_NAME="Fedora 28 (Workstation Edition)"
ANSI_COLOR="0;34"
CPE_NAME="cpe:/o:fedoraproject:fedora:28"
HOME_URL="https://fedoraproject.org/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=28
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=28
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Workstation Edition"
VARIANT_ID=workstation
Fedora release 28 (Twenty Eight)
Fedora release 28 (Twenty Eight)
```

**Output of `uname -a`:**

```
[...]

Issue #1057, opened 2018-10-02
buildah tries to parse incorrect value


bparees:
https://github.com/containers/buildah/blob/master/pull.go#L193-L204

Two problems with this code:

1) pullError should be logged when it's non-nil
2) the hasRegistry() call ultimately fails because it's passing a `docker://blahblah` value to `Parse()` which doesn't expect the protocol prefix and throws an error.  (an error which is *also* not logged in a useful manner since it just reports `error parsing image name` without showing the value it was trying to parse).

@nalind

TomSweeneyRedHat:
@bparees the hasRegistry() function was replaced in https://github.com/containers/buildah/commit/6e0074eeedbc903acddeb7bb04093d615de055e5.  However that was still having issues with handling a fully specified image name like `docker://blahblah`.  I've hopefully addressed that with #1117.

I'll work on logging the non-nil error, I'll see if I can sneak that into #1117 too.

rhatdan:
It looks like this is fixed.

PR #1119, opened 2018-10-19
Don't ResolveName if image is fully specified

TomSweeneyRedHat:

If the user specified a fully qualified image name, i.e. `docker://centos:centos`, then don't try to parse it, just let it try to be resolved as is.  The prior code failed the parsing due to two colons ':' in the  given name.

Addresses: #1117


TomSweeneyRedHat:
I think this also addresses #1057

rhatdan:
I did a different version of this.  We could just check for the defaultransport prefix and then skip this check.

rhatdan:
@TomSweeneyRedHat PTAL at my version of this. https://github.com/containers/buildah/pull/1120

PR #1120, opened 2018-10-19
Resolve image names with default transport in from command

rhatdan:

### DIFF
--- a/pull.go
+++ b/pull.go
@@ -146,11 +146,11 @@ func pullImage(ctx context.Context, store storage.Store, imageName string, optio
 	srcRef, err := alltransports.ParseImageName(spec)
 	if err != nil {
 		if options.Transport == "" {
-			options.Transport = DefaultTransport
+			options.Transport = util.DefaultTransport
 		}
 		logrus.Debugf("error parsing image name %q, trying with transport %q: %v", spec, options.Transport, err)
 		transport := options.Transport
-		if transport != DefaultTransport {
+		if transport != util.DefaultTransport {
 			transport = transport + ":"
 		}
 		spec = transport + spec
@@ -201,6 +201,7 @@ func pullImage(ctx context.Context, store storage.Store, imageName string, optio
 
 	logrus.Debugf("copying %q to %q", spec, destName)
 	if _, err := cp.Image(ctx, policyContext, destRef, srcRef, getCopyOptions(options.ReportWriter, srcRef, sc, destRef, nil, "")); err != nil {
+		logrus.Debugf("error copying src image [%q] to dest image [%q] err: %v", spec, destName, err)
 		return nil, err
 	}
 	return destRef, nil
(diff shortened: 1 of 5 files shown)


pull.go
```diff
@@ -146,11 +146,11 @@ func pullImage(ctx context.Context, store storage.Store, imageName string, optio
 	srcRef, err := alltransports.ParseImageName(spec)
 	if err != nil {
 		if options.Transport == "" {
-			options.Transport = DefaultTransport
+			options.Transport = util.DefaultTransport
 		}
 		logrus.Debugf("error parsing image name %q, trying with transport %q: %v", spec, options.Transport, err)
 		transport := options.Transport
-		if transport != DefaultTransport {
+		if transport != util.DefaultTransport {
```
TomSweeneyRedHat: Can you add the error message from 1119 that's new to pull.go in this?  Or your variant of the message, then we can close off @bparees issue too.

util/util.go
```diff
@@ -89,6 +93,7 @@ func ResolveName(name string, firstRegistry string, sc *types.SystemContext, sto
 		}
 	}
 
+	name = strings.TrimPrefix(name, DefaultTransport)
```
TomSweeneyRedHat: In my review I'd done this at one point, but was concerned that the DefaultTransport might not be the only valid transport at some point and to trust the user.  Was I over thinking?  This did work when I tested it though.

rhatdan: Well the Default Transport is the only one that looks like `NAME://`  All other transports are `NAME:` and are handled separately.

So I don't see this as an issue.

pull.go
rhatdan: Done

rhatdan:
Addresses: #1117

rhatdan:
Fixes https://github.com/containers/buildah/issues/1057

util/util.go
TomSweeneyRedHat: Right, but this code only works for "docker://", would it be valid to have something else there at some point like "oci://"?  Also if they put in "Docker://" they'll get a bad reference message which is probably OK, not sure if you want to lowercase name or not.

pull.go
```diff
@@ -201,6 +201,7 @@ func pullImage(ctx context.Context, store storage.Store, imageName string, optio
 
 	logrus.Debugf("copying %q to %q", spec, destName)
 	if _, err := cp.Image(ctx, policyContext, destRef, srcRef, getCopyOptions(options.ReportWriter, srcRef, sc, destRef, nil, "")); err != nil {
+		logrus.Debugf("error copying src image [%q] to dest image [%q] err: %v", spec, destiName, err)
```
TomSweeneyRedHat: Crud, my bad, builds are failing due to this.  Change "destiName" to "destName", no "i".

rhatdan:

rhatdan:
Since github is not working properly, I will manually merge.

PR #1123, opened 2018-10-22
Use libpod GetDefaultStorage to report proper storage config

rhatdan:

### DIFF
--- a/cmd/buildah/main.go
+++ b/cmd/buildah/main.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/containers/buildah"
+	"github.com/containers/libpod/pkg/util"
 	"github.com/containers/storage"
 	ispecs "github.com/opencontainers/image-spec/specs-go"
 	rspecs "github.com/opencontainers/runtime-spec/specs-go"
@@ -18,6 +19,11 @@ func main() {
 	var defaultStoreDriverOptions *cli.StringSlice
 	if buildah.InitReexec() {
 		return
+	}
+	storageOptions, err := util.GetDefaultStoreOptions()
+	if err != nil {
+		logrus.Errorf(err.Error())
+		cli.OsExiter(1)
 	}
 
 	app := cli.NewApp()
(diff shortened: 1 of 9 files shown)


rhatdan:
@giuseppe PTAL

cmd/buildah/main.go
```diff
@@ -19,6 +20,11 @@ func main() {
 	if buildah.InitReexec() {
 		return
 	}
+	storageOptions, err := util.GetDefaultStoreOptions()
```
giuseppe: I think this won't work in buildah, as it uses the `rootless` package from libpod while buildah uses its own unshare reexec module.

rhatdan: I'm at least seeing the proper content in the help.  And rootless seems to work.

```
$ ./buildah --help
...
   --root value                          storage root dir (default: "/home/dwalsh/.local/share/containers/storage")
   --runroot value                       storage state dir (default: "/run/user/3267/run")
```
Versus
```
$ buildah --help
...
   --root value                          storage root dir (default: "/var/lib/containers/storage")
   --runroot value                       storage state dir (default: "/var/run/containers/storage")
```

giuseppe:
@rhatdan I've opened a PR that adds the missing pieces: https://github.com/containers/buildah/pull/1124

TomSweeneyRedHat:
@giuseppe Do we need to merge #1123 before or after #1124 or does the merge order not matter?

giuseppe:
this is a replacement for #1123, in facts it is based on top of it

docs/buildah.md
```diff
@@ -66,7 +66,7 @@ specify additional options via the `--storage-opt` flag.
 
 **--storage-opt** **value**
 
-Storage driver option, Default storage driver options are configured in /etc/containers/storage.conf. The `STORAGE_OPTS` environment variable overrides the default.  The --storage-opt specified options overrides all.
+Storage driver option, Default storage driver options are configured in /etc/containers/storage.conf (`$HOME/.config/containers/storage.conf` in rootless mode). The `STORAGE_OPTS` environment variable overrides the default. The --storage-opt specified options overrides all.
```
TomSweeneyRedHat: nit of a nit, and not in your changes, but "Default storage" should be "default storage", or we should replace the preceding comma with a period and change to "The default" like in --storage-driver.  Ignoring or doing in a follow up is fine.

PR #1125, opened 2018-10-23
Handle empty built in directory

TomSweeneyRedHat:

For at least the registry:2 image, the volume was existent, but empty:
```
           "ArgsEscaped": true,
            "Image": "sha256:bc0d86131a2db92f18b8413a38da41718748b152024be47a8bc9413fa1d50a95",
            "Volumes": {
                "/var/lib/registry": {}
            },
```
The srcPath was never created on disk by the time the stat happened at 514, so the stat failed.  I'm not sure if this is the right approach or if we need to somehow create the srcPath, I wasn't able to see away to do that unless we move this functionality outside of this function and do the processing after the directory is created.

Prior to this fix, this would fail:
```
registry=$(buildah from registry:2)
buildah --debug run $registry -- htpasswd -Bbn testuser testpassword > /root/auth/htpasswd
```

But if you ran the run command a second time, it would succeed.

### DIFF
--- a/run.go
+++ b/run.go
@@ -512,7 +512,8 @@ func runSetupBuiltinVolumes(mountLabel, mountPoint, containerDir string, copyWit
 			srcPath := filepath.Join(mountPoint, volume)
 			stat, err := os.Stat(srcPath)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
I'm going to close this one as either #1126 or less likely #1127 will replace it.

PR #1126, opened 2018-10-24
runSetupBuiltinVolumes(): break up volume setup

nalind:
Break setup for built-in volumes into independent steps where we create the volume's mount point, the directory that will hold its contents, and if there is content under the mount point, populate the volume with the mount point's contents.

This should help with #1125.


TomSweeneyRedHat:
I'm not sure why the 1.11 builds are failing on this PR as the file they're failing on are not part of this PR and pass in other PR's?
```
Error: source files are not formatted according to recommendations. Run "gofmt -s -w" on:
./cmd/buildah/unshare.go
```

TomSweeneyRedHat:
Quick review and the code looks good,  but I want to test/play with it in the morn.

rhatdan:
@rh-atomic-bot r+

Issue #1121, opened 2018-10-20
buildah run registry fails on first attempt.


TomSweeneyRedHat:
**Description**

When running a container image registry to create a password file, the first attempt fails, but the second attempt succeeds.  The error is /var/lib/registry can't be gotten to on the container.

**Steps to reproduce the issue:**
1. registry=$(buildah from registry:2)
2. buildah --debug run $registry -- htpasswd -Bbn testuser testpassword > /root/auth/htpasswd
3. buildah --debug run $registry -- htpasswd -Bbn testuser testpassword > /root/auth/htpasswd

**Describe the results you received:**
```
# registry=$(buildah from registry:2)
Getting image source signatures
Copying blob sha256:d6a5679aa3cfc52503fac8e2345790783537564df5af43355ec662d7703aacbf
 2.28 MiB / 2.28 MiB [======================================================] 0s
Copying blob sha256:ad0eac849f8f2014e20df11d611e4453c94a365a71eb3a90f02d8500391e1787
 1.94 MiB / 1.94 MiB [======================================================] 0s
Copying blob sha256:2261ba058a15e3f21a467b1b02ea206d40cf0632d6b1aeee2ad200b662ed4aff
 5.98 MiB / 5.98 MiB [======================================================] 0s
Copying blob sha256:f296fda86f10cfcb81d60d5bcb47a7784a8ec4876d6eac7fabd51f2a7e8694aa
 369 B / 369 B [============================================================] 0s
Copying blob sha256:bcd4a541795bb9981999ce8f6e2e1c92e55b0041256153d2458ff35fb9381063
 212 B / 212 B [============================================================] 0s
Copying config sha256:2e2f252f3c88679f1207d87d57c07af6819a1a17e22573bcef32804122d2f305
 3.09 KiB / 3.09 KiB [======================================================] 0s
Writing manifest to image destination
Storing signatures

# buildah --debug run $registry -- htpasswd -Bbn testuser testpassword > /root/auth/htpasswd
DEBU[0000] [graphdriver] trying provided driver "overlay" 
DEBU[0000] overlay: override_kernelcheck=true           
DEBU[0000] overlay test mount with multiple lowers succeeded 
DEBU[0000] backingFs=xfs, projectQuotaSupported=false, useNativeDiff=true 
DEBU[0000] using "/tmp/buildah198005705" to hold bundle data 
[...]

PR #1127, opened 2018-10-24
Add builtin directory on run if non-existent

TomSweeneyRedHat:

I think this will most likely replace #1125, but don't want to close that quite yet.

If an image has a builtin volume attached to it like:

```
            "Volumes": {
                "/var/lib/registry": {}
            },
```
Create the directory if it doesn't exist and fails the first stat, then attempt to continue processing.  

I've one slight concern with this change.  To replicate it prior to this change:

```
registry=$(buildah from registry:2)
buildah --debug run $registry -- htpasswd -Bbn testuser testpassword > /root/auth/htpasswd   # failed
buildah --debug run $registry -- htpasswd -Bbn testuser testpassword > /root/auth/htpasswd   # completed
```

After this change both runs work.  However my concern is what's created on the host.  After the first `buildah run` I see:

```
# find /var/lib/containers/storage -name registry
/var/lib/containers/storage/overlay/1fe8be720d6e3a8540dc649be8542b5304c46a88d5dcfac49729b3e8c267b4ac/diff/bin/registry
/var/lib/containers/storage/overlay/a4c34f0c87b2563d2835c48bffe2b940fea9719850c003c314e487e6af44da53/diff/etc/docker/registry
/var/lib/containers/storage/overlay/241c7d778023cd436c1148c16af2c039c8ae9f27d79bf66ee8dc6386848c6dea/diff/var/lib/registry
```

But I was thinking I should see something under 

```
/var/lib/containers/storage/overlay/241c7d778023cd436c1148c16af2c039c8ae9f27d79bf66ee8dc6386848c6dea/merged
```

too, but I'm not.  Am I overexpecting @nalind ?  

Otherwise with this change I'm seeing the /var/lib/registry directory in the container as I'd expect.

### DIFF
--- a/run.go
+++ b/run.go
@@ -512,7 +512,14 @@ func runSetupBuiltinVolumes(mountLabel, mountPoint, containerDir string, copyWit
 			srcPath := filepath.Join(mountPoint, volume)
 			stat, err := os.Stat(srcPath)
 			if err != nil {
-				return nil, errors.Wrapf(err, "failed to stat %q for volume %q", srcPath, volume)
+				logrus.Debugf("failed to stat %q for volume %q err %v, attempting MkdirAll", srcPath, volume, err)
+				if err = os.MkdirAll(srcPath, 0755); err != nil {
+					return nil, errors.Wrapf(err, "error creating directory %q", srcPath)
+				}
+				stat, err = os.Stat(srcPath)
+				if err != nil {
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
Addresses #1121 and I didn't see @nalind's #1126 until I posted this, may retract this one.

Issue #1081, opened 2018-10-09
can't inspect cgroup limits within a build


bparees:
One of our tests attempts to set memory/cpu limits on the docker build being performed and attempts to confirm the values by running a script as one of the RUN commands, where the script itself inspects the cgroup limits:

https://github.com/openshift/origin/blob/master/test/extended/testdata/builds/build-quota/.s2i/bin/assemble

This is failing under buildah due to permissions issues:

```
          2018-10-05T03:43:54.900581678Z MEMORY=cat: /sys/fs/cgroup/memory/memory.limit_in_bytes: Permission denied
          2018-10-05T03:43:54.902597615Z MEMORYSWAP=cat: /sys/fs/cgroup/memory/memory.memsw.limit_in_bytes: Permission denied
          2018-10-05T03:43:54.904834655Z QUOTA=/tmp/scripts/assemble: line 13: /sys/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us: Permission denied
          2018-10-05T03:43:54.904996984Z cat: /sys/fs/cgroup/cpu,cpuacct/cpu.cfs_quota_us: Permission denied
          2018-10-05T03:43:54.995823041Z SHARES=PERIOD=cat: /sys/fs/cgroup/cpu,cpuacct/cpu.shares: Permission denied
          2018-10-05T03:43:54.9975429Z cat: /sys/fs/cgroup/cpu,cpuacct/cpu.cfs_period_us: Permission denied
```

rhatdan:
SELinux is preventing it.  I believe.

rhatdan:
Nope it looks like podman and buildah can?
```
podman run fedora cat /sys/fs/cgroup/memory/memory.limit_in_bytes
9223372036854771712
```
```
buildah bud ~
STEP 1: FROM fedora
STEP 2: RUN cat /sys/fs/cgroup/memory/memory.limit_in_bytes
9223372036854771712
STEP 3: COMMIT containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.mountopt=nodev]@2a120441bc09faa7a7c683bb66b8e3187891532ce40fc30d6aa01e24e65162f3
Getting image source signatures
Skipping fetch of repeat blob sha256:888cf76dff2b5266ded71f295eb5d891b04ac2f11cfa7e50dc4dea86e753aca8
Copying blob sha256:a17633bd8ec191199638477af1e5d778eda5ca765ed79edaa96ac3c940f64298
[...]

PR #1130, opened 2018-10-26
Allow cgroups to be passed into chroot, read/only

rhatdan:
Interesting finding here.  MS_REC only effects the Propagation flags, it does not apply the "ro" options to the recursive mounts.  So we need to apply them individually.

### DIFF
--- a/chroot/run.go
+++ b/chroot/run.go
@@ -955,6 +955,20 @@ func setRlimits(spec *specs.Spec, onlyLower, onlyRaise bool) error {
 	return nil
 }
 
+func makeReadOnly(mntpoint string, flags uintptr) error {
+	var fs unix.Statfs_t
+	// Make sure it's read-only.
+	if err := unix.Statfs(mntpoint, &fs); err != nil {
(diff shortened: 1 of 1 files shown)


rhatdan:
Fixes https://github.com/containers/buildah/issues/1081
@nalind @bparees PTAL

rhatdan:
@rhvgoyal FYI

TomSweeneyRedHat:
re-ping @bparees @nalind @rhvgoyal PTAL

bparees:
@rhatdan Sorry, i don't have the relevant expertise to review this PR effectively, I was assuming @nalind would cover it.

rhatdan:
@bparees I guess in the end, we need to make sure your tests in OpenShift that look at cgroups work after we merge this patch.

bparees:
> @bparees I guess in the end, we need to make sure your tests in OpenShift that look at cgroups work after we merge this patch.

yeah i certainly plan to do that.

rhatdan:
@rh-atomic-bot r+

PR #1141, opened 2018-10-30
migrate from chrootuser to libpod/pkg/lookup

baude:
for user and group related lookups, we now use libpod/pkg/lookup


nalind:
How does this avoid reintroducing the risks in #66?

baude:
@nalind I'm not sure I follow ... the primary reason we made changes in libpod was for performance and secondarily to use the secure filejoin.  My motivation for submitting this PR was the same.  If you guys feel that this doesn't help or even hurts, lets just close the PR.  I was just trying to act on due diligence.

rhatdan:
The securejoin protects us against a hostile image.  That is what the issue originally complained about, I believe. The chrootuser version does protect against a hostile process running in a container when someone launches this code against the same rootfs. Seems that the community is willing to take this risk for the benefits of speed.

rhatdan:
@baude  needs a rebase.

rhatdan:
Since @nalind is not comfortable with this, we will wait for customer feedback on it being slow.  Closing.

PR #1143, opened 2018-10-31
run --volume: Created if hostdir does not exist.

q384566678:
When using the `volume` option in the `run` command again, create if the specified hostdir does not exist.


rhatdan:
Why do this?  I don't like this behavior since it will make typos work.
If I mean to volume mount in /var/lib/mariadb and I spell it /var/lib/mariaDb everything silently succeeds but It is really broken.  I know that Docker does this, but we rejected a PR to do this in Podman and I think we should here as well.

Sorry @q384566678

q384566678:
Fair enough.

PR #1146, opened 2018-10-31
run: only set up /etc/hosts or /etc/resolv.conf with network

### LINKED ISSUE #1142: Cannot modify `resolv.conf` inside of the container
I'd like to be able to modify the `/etc/resolv.conf` file inside of the container (namely, to be a symlink to `/run/systemd/resolve/resolv.conf`), but `buildah run` bind mounts over this path and it cannot be modified this way. Something like `--net none` should prevent bind-mounting over `/etc/hosts` and `/etc/resolv.conf`

nalind:
Only set up bind mounts of copies of the host's `/etc/hosts` and `/etc/resolv.conf` files if we're setting up networking.  Should resolve #1142.

### DIFF
--- a/run.go
+++ b/run.go
@@ -1060,24 +1060,28 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	}
 	rootIDPair := &idtools.IDPair{UID: int(rootUID), GID: int(rootGID)}
 
-	hostFile, err := b.addNetworkConfig(path, "/etc/hosts", rootIDPair)
-	if err != nil {
-		return err
-	}
-	resolvFile, err := b.addNetworkConfig(path, "/etc/resolv.conf", rootIDPair)
-	if err != nil {
-		return err
+	bindFiles := make(map[string]string)
+	namespaceOptions := append(b.NamespaceOptions, options.NamespaceOptions...)
+	networkNamespace := namespaceOptions.Find(string(specs.NetworkNamespace))
+	if networkNamespace == nil || networkNamespace.Host || networkNamespace.Path != "" {
+		hostFile, err := b.addNetworkConfig(path, "/etc/hosts", rootIDPair)
+		if err != nil {
+			return err
+		}
+		bindFiles["/etc/hosts"] = hostFile
+
+		if err := addHostsToFile(b.CommonBuildOpts.AddHost, hostFile); err != nil {
+			return err
+		}
+
+		resolvFile, err := b.addNetworkConfig(path, "/etc/resolv.conf", rootIDPair)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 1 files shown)


run.go
```diff
@@ -1060,23 +1060,25 @@ func (b *Builder) Run(command []string, options RunOptions) error {
...
-	hostFile, err := b.addNetworkConfig(path, "/etc/hosts", rootIDPair)
-	if err != nil {
-		return err
-	}
-	resolvFile, err := b.addNetworkConfig(path, "/etc/resolv.conf", rootIDPair)
-	if err != nil {
-		return err
-	}
+	bindFiles := make(map[string]string)
+	if configureNetwork {
```
rhatdan: Is configureNetwork set for --net=host?  I thought it was turned off for that as well.  
I might be mistaken, but I saw little difference internally between 

--net=host
and
--net=none (This one actually did not work.  I was still able to do a yum update).

TomSweeneyRedHat:
Change LGTM, just wondering if we should toggle based on --net=none type of param from the user.

run.go
```diff
@@ -1060,23 +1060,25 @@ func (b *Builder) Run(command []string, options RunOptions) error {
...
+			return err
+		}
+		bindFiles["/etc/hosts"] = hostFile
 
-	if err := addHostsToFile(b.CommonBuildOpts.AddHost, hostFile); err != nil {
-		return err
-	}
+		if err := addHostsToFile(b.CommonBuildOpts.AddHost, hostFile); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
nalind: `git show` is giving me tabs.  The highlighting appears to be pointing to the change in indentation level.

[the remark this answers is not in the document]
TomSweeneyRedHat: Thanks for verifying @nalind.  I thought that might be the case, but wasn't sure.

run.go
nalind: Good point.  That controls whether or not we call CNI to do things, so it isn't.

nalind: In rootless mode, we had been discarding `--net=none` by forcing `--net=host` for every case.  Was this how you checked?

nalind: Added a change here to only discard the part of the configuration that specifies configuration names - whether or not we're using the host's network namespace, and if we're either leaving it unconfigured or attempting to attach to another namespace by specifying its path, should be preserved now for rootless containers.  @TomSweeneyRedHat is this what you were thinking?

TomSweeneyRedHat: @nalind, yep, was basing my question off of what @rhatdan had done in https://github.com/containers/libpod/pull/1734.  I'd not seen Dan's comment here until after I punched mine in also.

rhatdan:
@rh-atomic-bot r+

PR #1147, opened 2018-10-31
Update github.com/openshift/imagebuilder

nalind:
This updates our version of imagebuilder and takes into account an API change that lets one of its functions return an error when it couldn't before.


nalind:
Yup!  Closing.

PR #1149, opened 2018-10-31
Add optional blob caching for very specific use cases

nalind:
Add a blob cache reference type which wraps other reference types.  When used to create source and destination images for copying, it returns wrappers for source and destination images created by the wrapped references.

The cache reference can be created with a specified directory, or it will create a temporary directory of its own.  When told to clear its cache, if a temporary directory is being used, the reference will also remove the temporary directory.

Blobs that are written to a cache image destination are stored as-is on disk.  When reading a blob from a cache image source, if a blob with the matching digest (and if specified, a matching size) can be found on disk, read it directly instead of reading it from the real image source.

Add API hooks for designating locations to be used as blob caches when pulling and pushing images.  When we commit read-only copies of container layers for use in images, if we're using blob caching, store a copy of the layer in the blob cache directory so that it can be found.

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -133,6 +133,12 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 		return imgID, nil, "", errors.Wrapf(err, "error computing layer digests and building metadata for container %q", b.ContainerID)
 	}
 	// "Copy" our image to where it needs to be.
+	switch options.Compression {
+	case archive.Uncompressed:
+		systemContext.OCIAcceptUncompressedLayers = true
+	case archive.Gzip:
+		systemContext.DirForceCompress = true
+	}
 	var manifestBytes []byte
 	if manifestBytes, err = cp.Image(ctx, policyContext, dest, src, getCopyOptions(options.ReportWriter, src, nil, dest, systemContext, "")); err != nil {
 		return imgID, nil, "", errors.Wrapf(err, "error copying layers and metadata for container %q", b.ContainerID)
@@ -210,6 +216,12 @@ func Push(ctx context.Context, image string, dest types.ImageReference, options 
 		return nil, "", err
 	}
 	// Copy everything.
+	switch options.Compression {
+	case archive.Uncompressed:
+		systemContext.OCIAcceptUncompressedLayers = true
+	case archive.Gzip:
+		systemContext.DirForceCompress = true
+	}
 	var manifestBytes []byte
 	if manifestBytes, err = cp.Image(ctx, policyContext, dest, src, getCopyOptions(options.ReportWriter, src, nil, dest, systemContext, options.ManifestType)); err != nil {
 		return nil, "", errors.Wrapf(err, "error copying layers and metadata from %q to %q", transports.ImageName(src), transports.ImageName(dest))
--- a/commit.go
+++ b/commit.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"time"
 
+	"github.com/containers/buildah/pkg/blobcache"
 	"github.com/containers/buildah/util"
 	cp "github.com/containers/image/copy"
 	"github.com/containers/image/docker/reference"
@@ -55,6 +56,12 @@ type CommitOptions struct {
 	// Squash tells the builder to produce an image with a single layer
 	// instead of with possibly more than one layer.
 	Squash bool
+	// BlobDirectory is the name of a directory in which we'll look for
+	// prebuilt copies of layer blobs that we might otherwise need to
+	// regenerate from on-disk layers.  If blobs are available, the
+	// manifest of the new image will reference the blobs rather than
+	// on-disk layers.
+	BlobDirectory string
 
 	// OnBuild is a list of commands to be run by images based on this image
 	OnBuild []string
@@ -85,6 +92,11 @@ type PushOptions struct {
 	// ManifestType is the format to use when saving the imge using the 'dir' transport
 	// possible options are oci, v2s1, and v2s2
 	ManifestType string
+	// BlobDirectory is the name of a directory in which we'll look for
+	// prebuilt copies of layer blobs that we might otherwise need to
+	// regenerate from on-disk layers, substituting them in the list of
+	// blobs to copy whenever possible.
+	BlobDirectory string
 }
 
 // Commit writes the contents of the container, along with its updated
@@ -128,9 +140,27 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 			}
 		}
 	}
-	src, err := b.makeImageRef(options.PreferredManifestType, options.Parent, exportBaseLayers, options.Squash, options.Compression, options.HistoryTimestamp)
+	src, err := b.makeImageRef(options.PreferredManifestType, options.Parent, exportBaseLayers, options.Squash, options.BlobDirectory, options.Compression, options.HistoryTimestamp)
 	if err != nil {
 		return imgID, nil, "", errors.Wrapf(err, "error computing layer digests and building metadata for container %q", b.ContainerID)
+	}
+	var maybeCachedSrc types.ImageReference = src
+	var maybeCachedDest types.ImageReference = dest
+	if options.BlobDirectory != "" {
+		compress := types.PreserveOriginal
+		if options.Compression != archive.Uncompressed {
+			compress = types.Compress
+		}
+		cache, err := blobcache.NewBlobCache(src, options.BlobDirectory, compress)
+		if err != nil {
+			return imgID, nil, "", errors.Wrapf(err, "error wrapping image reference %q in blob cache at %q", transports.ImageName(src), options.BlobDirectory)
+		}
+		maybeCachedSrc = cache
+		cache, err = blobcache.NewBlobCache(dest, options.BlobDirectory, compress)
+		if err != nil {
+			return imgID, nil, "", errors.Wrapf(err, "error wrapping image reference %q in blob cache at %q", transports.ImageName(dest), options.BlobDirectory)
+		}
+		maybeCachedDest = cache
 	}
 	// "Copy" our image to where it needs to be.
 	switch options.Compression {
@@ -140,7 +170,7 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 		systemContext.DirForceCompress = true
 	}
 	var manifestBytes []byte
-	if manifestBytes, err = cp.Image(ctx, policyContext, dest, src, getCopyOptions(options.ReportWriter, src, nil, dest, systemContext, "")); err != nil {
+	if manifestBytes, err = cp.Image(ctx, policyContext, maybeCachedDest, maybeCachedSrc, getCopyOptions(options.ReportWriter, maybeCachedSrc, nil, maybeCachedDest, systemContext, "")); err != nil {
 		return imgID, nil, "", errors.Wrapf(err, "error copying layers and metadata for container %q", b.ContainerID)
 	}
 	if len(options.AdditionalTags) > 0 {
@@ -215,6 +245,18 @@ func Push(ctx context.Context, image string, dest types.ImageReference, options 
 	if err != nil {
 		return nil, "", err
 	}
+	var maybeCachedSrc types.ImageReference = src
+	if options.BlobDirectory != "" {
+		compress := types.PreserveOriginal
+		if options.Compression != archive.Uncompressed {
+			compress = types.Compress
+		}
+		cache, err := blobcache.NewBlobCache(src, options.BlobDirectory, compress)
+		if err != nil {
+			return nil, "", errors.Wrapf(err, "error wrapping image reference %q in blob cache at %q", transports.ImageName(src), options.BlobDirectory)
+		}
+		maybeCachedSrc = cache
+	}
 	// Copy everything.
 	switch options.Compression {
 	case archive.Uncompressed:
@@ -223,8 +265,8 @@ func Push(ctx context.Context, image string, dest types.ImageReference, options 
 		systemContext.DirForceCompress = true
 	}
 	var manifestBytes []byte
-	if manifestBytes, err = cp.Image(ctx, policyContext, dest, src, getCopyOptions(options.ReportWriter, src, nil, dest, systemContext, options.ManifestType)); err != nil {
-		return nil, "", errors.Wrapf(err, "error copying layers and metadata from %q to %q", transports.ImageName(src), transports.ImageName(dest))
+	if manifestBytes, err = cp.Image(ctx, policyContext, dest, maybeCachedSrc, getCopyOptions(options.ReportWriter, maybeCachedSrc, nil, dest, systemContext, options.ManifestType)); err != nil {
+		return nil, "", errors.Wrapf(err, "error copying layers and metadata from %q to %q", transports.ImageName(maybeCachedSrc), transports.ImageName(dest))
 	}
 	if options.ReportWriter != nil {
 		fmt.Fprintf(options.ReportWriter, "")
--- a/pkg/blobcache/blobcache.go
+++ b/pkg/blobcache/blobcache.go
@@ -1,0 +1,548 @@
+package blobcache
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sync"
+
+	"github.com/containers/buildah/docker"
+	"github.com/containers/image/docker/reference"
+	"github.com/containers/image/image"
+	"github.com/containers/image/manifest"
+	"github.com/containers/image/transports"
+	"github.com/containers/image/types"
+	"github.com/containers/storage/pkg/archive"
+	"github.com/containers/storage/pkg/ioutils"
+	digest "github.com/opencontainers/go-digest"
+	"github.com/opencontainers/image-spec/specs-go/v1"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	_ types.ImageReference   = &blobCacheReference{}
+	_ types.ImageSource      = &blobCacheSource{}
+	_ types.ImageDestination = &blobCacheDestination{}
+)
+
+const (
+	compressedNote   = ".compressed"
+	decompressedNote = ".decompressed"
+)
+
+// BlobCache is an object which saves copies of blobs that are written to it while passing them
+// through to some real destination, and which can be queried directly in order to read them
+// back.
+type BlobCache interface {
+	types.ImageReference
+	// HasBlob checks if a blob that matches the passed-in digest (and
+	// size, if not -1), is present in the cache.
+	HasBlob(types.BlobInfo) (bool, int64, error)
+	// PutBlob adds a blob to the cache.  The expected digest of the blob
+	// must be provided.
+	PutBlob(ctx context.Context, stream io.Reader, inputInfo types.BlobInfo, isConfig bool) error
+	// Directories returns the list of cache directories.
+	Directory() string
+	// ClearCache() clears the contents of the cache directories.  Note
+	// that this also clears content which was not placed there by this
+	// cache implementation.
+	ClearCache() error
+}
+
+type blobCacheReference struct {
+	reference types.ImageReference
+	directory string
+	compress  types.LayerCompression
+}
+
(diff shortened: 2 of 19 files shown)


TomSweeneyRedHat:
Luv me Red Hat CI, looks like it's not happy today, I don't believe due to this PR.

rhatdan:
bot, retest this please

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,410 @@
...
+			continue
+		}
+		// We don't "own" this directory, so just clear its contents.
+		f, err := os.Open(directory)
+		if err != nil {
+			return errors.Wrapf(err, "error opening directory %q", directory)
+		}
+		names, err := f.Readdirnames(-1)
+		if err != nil {
+			f.Close()
```
rhatdan: Can't we use defer f.Close

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,410 @@
...
+		if d.reference.removeFirstDirectoryOnCleanup && d.reference.directories[0] == "" {
+			temp, err := ioutil.TempDir("", "")
+			if err != nil {
+				return types.BlobInfo{}, errors.Wrapf(err, "error creating new temporary directory for blob cache")
+			}
+			d.reference.directories[0] = temp
+		}
+		filename := filepath.Join(d.reference.directories[0], makeFilename(inputInfo.Digest, isConfig))
+		f, err := os.OpenFile(filename, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, 0600)
+		if err == nil {
```
rhatdan: Are you ignoring err != nil?
Should we only ignore ENOEXIST?

rhatdan:
@vrothberg @mtrmac PTAL

rhatdan:
@nalind Shouldn't we remove the blob-cache when we remove the image?

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,410 @@
...
+}
+
+func (d *blobCacheDestination) PutManifest(ctx context.Context, manifestBytes []byte) error {
+	if len(d.reference.directories) > 0 {
+		manifestDigest, err := manifest.Digest(manifestBytes)
+		if err != nil {
+			logrus.Warnf("error digesting manifest %q: %v", string(manifestBytes), err)
+		} else {
+			filename := filepath.Join(d.reference.directories[0], makeFilename(manifestDigest, false))
+			if err = ioutil.WriteFile(filename, manifestBytes, 0600); err != nil {
```
mtrmac: Does anything read these cached manifests?

pkg/blobcache/blobcache.go
nalind: If we're opening _n_ directories, we'll hold those descriptors open until we've finished cleaning them all.  Probably not worth worrying about that, though, so changing it.

pkg/blobcache/blobcache.go
nalind: Yes, and probably yes.  Adding a debug message for the `!os.IsNotExist()` case.

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,410 @@
...
+	components := strings.SplitN(reference, ":", 2)
+	realRef := ""
+	directories := []string{}
+	switch len(components) {
+	case 1:
+		realRef = components[0]
+	case 2:
+		if transports.Get(components[0]) != nil {
+			realRef = components[0] + ":" + components[1]
+		} else {
```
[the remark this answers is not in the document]
nalind: Yeah, I'm doubting the usefulness of being able to omit directory names.

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,410 @@
...
+}
+
+func (d *blobCacheDestination) PutBlob(ctx context.Context, stream io.Reader, inputInfo types.BlobInfo, isConfig bool) (types.BlobInfo, error) {
+	if inputInfo.Digest != "" && len(d.reference.directories) > 0 {
+		if d.reference.removeFirstDirectoryOnCleanup && d.reference.directories[0] == "" {
+			temp, err := ioutil.TempDir("", "")
+			if err != nil {
+				return types.BlobInfo{}, errors.Wrapf(err, "error creating new temporary directory for blob cache")
+			}
+			d.reference.directories[0] = temp
```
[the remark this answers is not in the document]
nalind: This is another reason to remove that flexibility discussed about above.

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,410 @@
...
+		}
+		filename := filepath.Join(d.reference.directories[0], makeFilename(inputInfo.Digest, isConfig))
```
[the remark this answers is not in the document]
nalind: Good point.  Will have to spool to a temporary file and then `Rename()` it.

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,410 @@
...
+
+func (s *blobCacheSource) Close() error {
+	return s.source.Close()
+}
+
+func (s *blobCacheSource) GetManifest(ctx context.Context, instanceDigest *digest.Digest) ([]byte, string, error) {
+	return s.source.GetManifest(ctx, instanceDigest)
+}
+
+func (s *blobCacheSource) getBlob(ctx context.Context, blobinfo types.BlobInfo) (bool, io.ReadCloser, int64, error) {
```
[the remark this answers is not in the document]
nalind: I think an earlier version also used this for `HasBlob()`, but since it isn't needed now, I'll merge them.

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,410 @@
...
+	present, size, err := s.reference.HasBlob(blobinfo)
+	if err != nil {
+		return false, nil, -1, err
+	}
+	if !present {
```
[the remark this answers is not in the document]
nalind: Ah, yes, that reads better.  Changing it.

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,410 @@
...
+	_, rc, size, err := s.getBlob(ctx, blobinfo)
+	return rc, size, err
+}
+
+func (s *blobCacheSource) GetSignatures(ctx context.Context, instanceDigest *digest.Digest) ([][]byte, error) {
+	return s.source.GetSignatures(ctx, instanceDigest)
+}
+
+func (s *blobCacheSource) LayerInfosForCopy(ctx context.Context) ([]types.BlobInfo, error) {
+	return nil, nil
```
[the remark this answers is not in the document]
nalind: An image that we committed to `containers-storage` will generally use the uncompressed diffIDs in its manifest, since at commit-time, we don't know that we'll be able to supply layer blobs in a compressed form that matches a given digest, so the underlying `LayerInfosForCopy()` doesn't return values that match the blobs that we've cached.
In order to force a copying operation to use the blobs from the cache, we also have to "break" `Commit()` to have it record the original compressed digest values when it creates a manifest, which means that copying the resulting image can fail unless the cache is also there to supply blobs at copy-time.

pkg/blobcache/blobcache.go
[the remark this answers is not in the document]
nalind: Removing any munging of the directories as given in the structure.  (We have to replace `""` with `"."` when passing the directory path to `ioutil.TempFile()`, but we do that only when we need to.)

pkg/blobcache/blobcache.go
nalind: I could see us wanting to cache manifests for non-`nil` instanceDigest values.  Added a check for a cached copy when an `instanceDigest` is supplied to `GetManifest()`.

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,396 @@
...
+func (d *blobCacheDestination) PutBlob(ctx context.Context, stream io.Reader, inputInfo types.BlobInfo, isConfig bool) (types.BlobInfo, error) {
+	var tempfile *os.File
+	var err error
+	if inputInfo.Digest != "" {
+		directory := d.reference.directories[0]
+		if directory == "" {
+			directory = "."
+		}
+		filename := filepath.Join(directory, makeFilename(inputInfo.Digest, isConfig))
+		tempfile, err = ioutil.TempFile(directory, makeFilename(inputInfo.Digest, isConfig))
```
mtrmac: (Previously `directories[0] == ""` meant “create a temporary directory”; that has intentionally gone away?)

- Is it a good default to store temporary caches to `"."`? It’s not _impossible_ but it’s a bit surprising.
- Does `directories == []string{""}` work consistently? AFAICS `StringWithinTransport` will format it as (`blob-cache:`)`:`_ref_, and `ParseReference` interprets that as `[]string{}`, not `[]string{""}`.

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,396 @@
...
+	switch len(components) {
+	case 2:
+		directories = strings.Split(components[0], ",")
+		if len(directories) == 1 && directories[0] == "" {
+			directories = []string{}
+		}
+		realRef = components[1]
+	default:
+		return nil, errors.Errorf("error parsing reference for %q image %q: bad syntax", t.Name(), reference)
+	}
```
mtrmac: (Absolutely non-blocking: this would be shorter with an `if len(components) != 2 { /* return error */ }`, the two variable declarations outside the `switch` scope could be merged with the assignments inside the `switch`.)

pkg/blobcache/blobcache.go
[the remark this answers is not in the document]
mtrmac: > An image that we committed to `containers-storage` will generally use the uncompressed diffIDs in its manifest, since at commit-time, we don't know that we'll be able to supply layer blobs in a compressed form that matches a given digest

AFAICS it’s the opposite: `DesiredLayerCompression`+`PutBlob` pretend to accept the unmodified compressed blobs, and the original manifest referring to compressed blobs is recorded — as well as returned by `GetManifest`.

> so the underlying `LayerInfosForCopy()` doesn't return values that match the blobs that we've cached.

Ah, right — `LayerInfosForCopy` is where the uncompressed blob digests get substituted, and we don’t want that happening here, at least _not always_.

> In order to force a copying operation to use the blobs from the cache, we also have to "break" `Commit()` to have it record the original compressed digest values when it creates a manifest, which means that copying the resulting image can fail unless the cache is also there to supply blobs at copy-time.

AFAICS this implementation forwards to the underlying `Commit` without modification, so I’m not sure what this  is referring to.

---

Overall, though, the choice to blindly use unmodified `LayerInfosForCopy` here seems rather radical to me — now the cached blobs _must_ be present or copying from `containers-storage:` will fail.  That looks brittle and very non-cache-like (in particular, the only time the cache can be enabled is when the local image storage is empty).

(Arguably the outlier here is `containers-storage:` pretending to store the compressed blobs, not the cache. I’m not really strongly arguing for changing that at this point, though.)

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,410 @@
...
+	return transports.ImageName(r.reference)
+}
+
+func (r *blobCacheReference) DockerReference() reference.Named {
+	return r.reference.DockerReference()
+}
+
+func (r *blobCacheReference) PolicyConfigurationIdentity() string {
+	return r.reference.PolicyConfigurationIdentity()
```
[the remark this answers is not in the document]
nalind: Switched this from being a "transport" of its own to being a wrapper for references that creates and returns wrapped image sources and destinations.

pkg/blobcache/blobcache.go
nalind: Now that we can only create caches using `NewBlobCache()`, the directory (of which there can only be one) must be specified, or we return an error.

rhatdan:
@mtrmac PTAL

nalind:
Updated so that we don't change the manifest that `Commit()` commits, and we do substitutions in `LayerInfosForCopy()` based on whether we prefer compressed or uncompressed blobs, though that part's largely guesswork.

rhatdan:
@nalind  needs a rebase.

nalind:
Before we push an image, we _could_ open the destination ourselves and ask its `DesiredLayerCompression()` method what the destination would prefer, and use that to guide whether or not we make adjustments to what we return from the source `LayerInfosForCopy()`, but that might be assuming too much about how `copy.Image()` will continue to do things.

mtrmac:
> Before we push an image, we _could_ open the destination ourselves and ask its `DesiredLayerCompression()` method what the destination would prefer, and use that to guide whether or not we make adjustments to what we return from the source `LayerInfosForCopy()`, but that might be assuming too much about how `copy.Image()` will continue to do things.

Yeah; `NewImageDestination` can in principle be fairly expensive (so far the worst is probably `docker-daemon:`, which is not negligible but not a disaster either), and in `copy.Image` there’s also the `canModify{Manifest,Blob}` logic involved…

At a first glance, it would be much more convenient to move the decision from the source (and `LayerInfosForCopy`) to the destination — after https://github.com/mtrmac/image/tree/blob-info-caching merges `HasBlob` with `ReapplyBlob` to `TryReusingBlob`, which is (unlike the current interface) allowed to replace the source-provided blob with a different semantically equivalent one (changing the digest).

OTOH I can’t remember the full flow of the data in this PR, maybe that is unworkable; also, moving the cache references from `GetBlob` to `TryReusingBlob` would hide copy operation from `copy.Image`, e.g. losing progress reporting.

rhatdan:
@nalind @mtrmac Is this ready to merge and handle the other conversation in a different PR?

rhatdan:
@nalind needs a rebase.

rhatdan:
@rh-atomic-bot r+

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,548 @@
...
+		return nil, errors.Wrapf(err, "error checking if image %q has signatures", transports.ImageName(s.reference))
+	}
+	canReplaceBlobs := !(len(signatures) > 0 && len(signatures[0]) > 0)
+
+	infos, err := s.source.LayerInfosForCopy(ctx)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error getting layer infos for copying image %q through cache", transports.ImageName(s.reference))
+	}
+	if infos == nil {
+		image, err := s.reference.NewImage(ctx, &s.sys)
```
mtrmac: Don’t do this if `!canReplaceBlobs`?

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,548 @@
...
+	}
+	return rc, size, nil
+}
+
+func (s *blobCacheSource) GetSignatures(ctx context.Context, instanceDigest *digest.Digest) ([][]byte, error) {
+	return s.source.GetSignatures(ctx, instanceDigest)
+}
+
+func (s *blobCacheSource) LayerInfosForCopy(ctx context.Context) ([]types.BlobInfo, error) {
+	signatures, err := s.source.GetSignatures(ctx, nil)
```
mtrmac: Don’t do this if `c.reference.compress == types.PreserveOriginal`?

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,548 @@
...
+		return rc, size, errors.Wrapf(err, "error reading blob from source image %q", transports.ImageName(s.reference))
+	}
+	return rc, size, nil
+}
+
+func (s *blobCacheSource) GetSignatures(ctx context.Context, instanceDigest *digest.Digest) ([][]byte, error) {
+	return s.source.GetSignatures(ctx, instanceDigest)
+}
+
+func (s *blobCacheSource) LayerInfosForCopy(ctx context.Context) ([]types.BlobInfo, error) {
```
mtrmac: Wow… this implementation is certainly not giving up — but this seems to me to be working at the wrong location/abstraction level, and it has to work _way_ too hard, giving up a fair amount of the performance benefits AFAICS.

Both the compressed/uncompresssed digest records in c/storage as they exist right now, and in the cache in  https://github.com/containers/image/pull/536; either would AFAICS completely avoid the need for the extra decompression in `saveStream`.

With https://github.com/containers/image/pull/536 letting the core `copy.Image` code take care of allowing substitutions and manifest updates, the extra `NewImage`/`GetSignatures` could be avoided as well.

commit.go
```diff
@@ -128,13 +140,37 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
...
 		return imgID, nil, "", errors.Wrapf(err, "error computing layer digests and building metadata for container %q", b.ContainerID)
 	}
+	var maybeCachedSrc types.ImageReference = src
+	var maybeCachedDest types.ImageReference = dest
+	if options.BlobDirectory != "" {
+		compress := types.PreserveOriginal
+		if options.Compression != archive.Uncompressed {
+			compress = types.Compress
+		}
+		cache, err := blobcache.NewBlobCache(src, options.BlobDirectory, compress)
```
mtrmac: AFAICS with the default behavior of `buildah commit`, `options.Compression == archive.Gzip`, i.e. when the destination is c/storage, this cache wrapper will, in `LayerInfosForCopy`, substitute cached compressed blobs for the uncached ones, and cause the destination to decompress them (twice! Once in `saveStream`, once in `storageImageDestination.PutBlob`), for not benefit whatsoever.

Actually, for the `makeImageRef`→c/storage common case, it seems to me that the caches should not be involved at all; either the data has already been cached, or this is eagerly making extra cached copies that might not ever be used.

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,548 @@
...
+func (r *blobCacheReference) putBlob(ctx context.Context, stream io.Reader, inputInfo types.BlobInfo, isConfig bool, destination types.ImageDestination) (types.BlobInfo, error) {
+	newBlobInfo := inputInfo
+	var tempfile *os.File
+	var err error
+	var n int
+	var alternateDigest digest.Digest
+	wg := new(sync.WaitGroup)
+	defer wg.Wait()
+	compression := archive.Uncompressed
+	if inputInfo.Digest != "" {
```
mtrmac: It’s rather non-obvious that this `if` is what differentiates between the “push” case (writing to `docker://`, compressing on the fly, i.e. `inputInfo.Digest == ""`), and the “pull“ case (writing to c/storage, _decompressing on the fly_, yet `inputInfo.Digest != ""` because the decompression happens inside `storageImageDestination`, invisible to `copy.Image`).  Definitely worth a comment.

rhatdan: Can you open a PR to fix these issues?

mtrmac: Apart from adding comments and the like, not this year, any more, realistically.

Avoiding the decompression and overhead in `LayerInfosForCopy` would be a fairly invasive change, and more or less depends on the still-unmerged https://github.com/containers/image/pull/536 .

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,548 @@
...
+	types.ImageReference
+	// HasBlob checks if a blob that matches the passed-in digest (and
+	// size, if not -1), is present in the cache.
+	HasBlob(types.BlobInfo) (bool, int64, error)
+	// PutBlob adds a blob to the cache.  The expected digest of the blob
+	// must be provided.
+	PutBlob(ctx context.Context, stream io.Reader, inputInfo types.BlobInfo, isConfig bool) error
+	// Directories returns the list of cache directories.
+	Directory() string
+	// ClearCache() clears the contents of the cache directories.  Note
```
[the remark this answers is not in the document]
nalind: Whoops, that's a leftover from when I thought multiple directories would be useful.

pkg/blobcache/blobcache.go
```diff
@@ -0,0 +1,548 @@
...
+			if n > 0 {
+				// Build a Reader that will still return the bytes that we just
+				// read, for PutBlob()'s sake.
+				stream = io.MultiReader(bytes.NewReader(initial[:n]), stream)
+				if n >= len(initial) {
+					compression = archive.DetectCompression(initial[:n])
+				}
+				if compression != archive.Uncompressed {
+					// The stream is compressed, so create a file which we'll
+					// use to store a decompressed copy.
```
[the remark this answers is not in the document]
nalind: That's the main reason it's there; replacing compressed blobs with uncompressed blobs is a possibility, but I don't see it being used very often.

Issue #1138, opened 2018-10-30
Build from is causing error in unprivileged container


olaf-meyer:
**Description**

I'm trying to build a container inside a non privileged container and I don't want to store any images on the host file system.  For this I created a centos container with buildah version 1.2 installed and 
a centos container which builds the latest buildah version 1.5-dev. When I run these container with the capability CAP_SYS_ADMIN and execute the command "buildah from" in the container I get an error message "permission denied" for buildah version 1.5-dev and "image not known" for buildah version 1.2. If I start the container as privileged than everything works fine. Do I need to give the container additional capability besides CAP_SYS_ADMIN? I attached the Dockerfiles that I'm using.

**Steps to reproduce the issue:**
1. Build container with "docker build -t build_container:centos_buildah_latest . -f Dockerfile_centos_buildah_latest" or "docker build -t build_container:centos7_buildah_1.2 . -f Dockerfile_centos_buildah_1.2"
2. Start container with "docker run --cap-add=SYS_ADMIN -it build_container:centos_buildah_latest /bin/bash" or "docker run --cap-add=SYS_ADMIN -it build_container:centos7_buildah_1.2 /bin/bash"
3. Execute inside the container the command "buildah --debug from --tls-verify=false docker.io/library/centos:7"

**Describe the results you received:**
For buildah version 1.2 "ERRO[0006] image "docker.io/library/centos:7" not found in /etc/containers/registries.conf registries: image not known" (see attachment Log_Docker_buildah_1.2_SYS_ADMIN for complete error message)
For buildah version 1.5-dev "ERRO[0006] image "ERRO[0006] Error committing the finished image: error adding layer with blob "sha256:aeb7866da422acc7e93dcf7323f38d7646f6269af33bcdb6647f2094fc4b3bf7": ApplyLayer exit status 1 stdout:  stderr: permission denied" (see attachment Log_Docker_buildah_latest_SYS_ADMIN for complete error message)

**Describe the results you expected:**
Buildah is able to download centos image from docker.io

**Output of `rpm -q buildah` or `apt list buildah`:**
Version 1.2:
```
buildah-1.2-2.gitbe87762.el7.x86_64
```
Version 1.5-dev
No possible, because buildah version was build inside the container.

**Output of `buildah version`:**
Version 1.2
```
Version:         1.2
Go Version:      go1.9.4
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.3.1
libcni Version:  v0.6.0
[...]

PR #1167, opened 2018-11-08
unshare: Fix running errors in non-privileged container

q384566678:
Solve the problem that buildah cannot be run in a non-privileged container. Although the non-privileged container cannot run the container, other commands can be executed, so all commands should not be reported and other commands should be executed.

No commands can be executed (except help and version) before modification:
```
root@zz-ubuntu-6fcc5f5d55-4plr7:~# buildah images
Error during unshare(CLONE_NEWUSER): Invalid argument
User namespaces are not enabled in /proc/sys/user/max_user_namespaces.
ERRO[0000] error parsing PID "": strconv.Atoi: parsing "": invalid syntax
ERRO[0000] (unable to determine exit status)
root@zz-ubuntu-6fcc5f5d55-4plr7:~# buildah pull busybox
Error during unshare(CLONE_NEWUSER): Invalid argument
User namespaces are not enabled in /proc/sys/user/max_user_namespaces.
ERRO[0000] error parsing PID "": strconv.Atoi: parsing "": invalid syntax
ERRO[0000] (unable to determine exit status)
```

After the modification, only the run command cannot be executed (because there is no permission to execute):
```
root@zz-ubuntu-6fcc5f5d55-4plr7:~# buildah pull busybox                                                                                                                        [33/1810]
Copying blob sha256:90e01955edcd85dac7985b72a8374545eac617ccdddcc992b732e43cd42534af
 710.92 KiB / 710.92 KiB [==================================================] 1s
Copying config sha256:59788edf1f3e78cd0ebe6ce1446e9d10788225db3dedcfd1a59f764bad2b2690
 1.46 KiB / 1.46 KiB [======================================================] 0s
Storing signatures
59788edf1f3e78cd0ebe6ce1446e9d10788225db3dedcfd1a59f764bad2b2690
root@zz-ubuntu-6fcc5f5d55-4plr7:~# buildah images
59788edf1f3e         docker.io/library/busybox:latest                         Oct 2, 2018 17:19      1.37 MB
root@zz-ubuntu-6fcc5f5d55-4plr7:~# buildah from 597
busybox-working-container
root@zz-ubuntu-6fcc5f5d55-4plr7:~# buildah run busybox-working-container bash
Error during unshare(CLONE_NEWUSER): Invalid argument
User namespaces are not enabled in /proc/sys/user/max_user_namespaces.
ERRO[0000] error parsing PID "": strconv.Atoi: parsing "": invalid syntax
ERRO[0000] (unable to determine exit status)
```

### DIFF
--- a/cmd/buildah/unshare.go
+++ b/cmd/buildah/unshare.go
@@ -63,8 +63,7 @@ func maybeReexecUsingUserNamespace(c *cli.Context, evenForRoot bool) {
 	if c.NArg() == 0 {
 		return
 	}
-	switch c.Args()[0] {
-	case "help", "version":
+	if c.Args()[0] != "run" {
 		return
 	}
 


q384566678:
Non-privileged container configuration can refer to #1011.
It is a follow-up to #1012. I think that although the container cannot be run in a non-privileged container, this should only be for the `run` command and should not be directed to other commands. Other `buildah` commands should be executed normally.
@rhatdan @TomSweeneyRedHat @giuseppe @nalind @umohnani8 PTAL

cmd/buildah/unshare.go
```diff
@@ -63,8 +63,7 @@ func maybeReexecUsingUserNamespace(c *cli.Context, evenForRoot bool) {
 	if c.NArg() == 0 {
 		return
 	}
-	switch c.Args()[0] {
-	case "help", "version":
+	if c.Args()[0] != "run" {
```
giuseppe: is "run" the only command that requires a namespace?  I think any command that is accessing the storage requires multiple IDs to be available

giuseppe: I don't think `buildah pull` will work without an user namespace

rhatdan:
Any tool that creates storage or reads storage will require UserNS, if there is more then one UID in container storage, the user will not be able to access/create it unless he is in a UserNS>
build, push, pull, run, create, start at least need it.  I think it would be safer to determine the ones that definitely do not need USERNS.

rhatdan:
@q384566678 Could you try to pull a container image that has more then one UID inside of it?  Does this work?  (It might if the container has setuid/setgid)

giuseppe:
this will break when running as rootless, so please check this case.  It will work in a not privileged container when running as root as long as cap_setuid/cap_setgid are available

q384566678:
>Could you try to pull a container image that has more then one UID inside of it? Does this work? (It might if the container has setuid/setgid)

I am not very clear which image meets this requirement. Is there an image that can be directly pulled? Or do I need to create it myself?

Currently I try to pull including ubuntu, centos, busybox and so on. These images can be successfully pulled down.

giuseppe:
> I think you mean to create a non-root user in the non-privileged container in k8s, and then use non-root user to execute buildah?

yes exactly.  It could also be tested without k8s, just try to use buildah as non root user

q384566678:
I added regular users and found that I couldn't execute buildah pull and other commands. It should be the problem of UserNS.
Now I've located the wrong place [here:](https://github.com/containers/buildah/blob/master/vendor/github.com/containers/storage/drivers/vfs/driver.go#L132)
```
	if err := idtools.MkdirAndChown(dir, 0755, rootIDs); err != nil {
		return err
	}
```

I am trying to find a way to solve this problem.

rhatdan:
@q384566678 @giuseppe Is this ready to merge?

q384566678:
This modification only allows the root user to run in a non-privileged container, and normal users cannot execute it, so this modification does not solve this problem very well.

Because buildah will try to create USERNS when it is executed, but now I find that the non-privileged container does not have enough capabilities to create USERNS, so I feel that there is no good way to solve this problem at present, perhaps adding the corresponding capabilities can solve this problem.

I am looking for the possibility of solving this problem in terms of code.

rhatdan:
@q384566678 Any progress?

q384566678:
I found a [blog](https://blog.42.be/2018/10/buildah-inside-centos-container.html) written by @42wim in #1138 [comment.](https://github.com/containers/buildah/issues/1138#issuecomment-435984410) I tried it according to his steps, but I still get an error.  @42wim Could you take a look？
```
[root@79c115f9c6cb /]# buildah bud .
Error during unshare(CLONE_NEWUSER): Operation not permitted
ERRO[0000] error parsing PID "": strconv.Atoi: parsing "": invalid syntax 
ERRO[0000] (unable to determine exit status)  
```

Even if I added some parameters while running, it was still unsuccessful.
```
# docker run --rm -ti --security-opt seccomp=unconfined centos:7 /bin/bash
```
```
[root@b3dd33f831e8 /]# buildah bud .
Error during unshare(...): Operation not permitted
ERRO[0000] exit status 1
```
According to the example in his blog, buildah can build an image in a unprivileged container. This is exactly what I need. If his process works correctly, the PR can be closed.

rhatdan:
You are blocked by CAP_SYS_ADMIN.

If you run the container with --cap-add CAP_SYS_ADMIN it should work.  (Might need to disable SELinux).

buildah also support --isolation chroot, which will not attempt to do a mount so no need for SYS_ADMIN privs.

q384566678:
After adding --isolation chroot, the problem was solved successfully. I will close this PR.
@rhatdan Thank you very much!

q384566678:
@rhatdan When I use the ubuntu system to test, even if I add `SYS_ADMIN`, I will still report an error.

```
root@zz-VirtualBox:/etc# docker run --rm -it --cap-add=SYS_ADMIN --security-opt seccomp=unconfined 8ccd578491be /bin/bash
```
Do not use `chroot`

```
root@2bfb07573789:/# buildah --storage-driver vfs bud .
STEP 1: FROM busybox
Copying blob sha256:90e01955edcd85dac7985b72a8374545eac617ccdddcc992b732e43cd42534af
 710.92 KiB / 710.92 KiB [==================================================] 1s
Copying config sha256:59788edf1f3e78cd0ebe6ce1446e9d10788225db3dedcfd1a59f764bad2b2690
 1.46 KiB / 1.46 KiB [======================================================] 0s
Storing signatures
ERRO[0012] Error while applying layer: ApplyLayer exit status 1 stdout:  stderr: permission denied 
error creating build container: The following failures happened while trying to pull image specified by "busybox" based on search registries in /etc/containers/registries.conf:
* "localhost/busybox": Error determining manifest MIME type for docker://localhost/busybox:latest: pinging docker registry returned: Get https://localhost/v2/: dial tcp 127.0.0.1:443: connect: connection refused
* "registry.fedoraproject.org/busybox": Error determining manifest MIME type for docker://registry.fedoraproject.org/busybox:latest: Error reading manifest latest in registry.fedoraproject.org/busybox: manifest unknown: manifest unknown
* "registry.access.redhat.com/busybox": Error determining manifest MIME type for docker://registry.access.redhat.com/busybox:latest: Error reading manifest latest in registry.access.redhat.com/busybox: unknown: Not Found
* "docker.io/library/busybox": Error committing the finished image: error adding layer with blob "sha256:90e01955edcd85dac7985b72a8374545eac617ccdddcc992b732e43cd42534af": ApplyLayer exit status 1 stdout:  stderr: permission denied
```
Use `chroot`
```
root@2bfb07573789:/# buildah --storage-driver vfs bud --isolation chroot -t test .                                                                                                      
STEP 1: FROM busybox                                                                                                                                                                    
Copying blob sha256:90e01955edcd85dac7985b72a8374545eac617ccdddcc992b732e43cd42534af                                                                                                    
 710.92 KiB / 710.92 KiB [==================================================] 2s                                                                                                        
Copying config sha256:59788edf1f3e78cd0ebe6ce1446e9d10788225db3dedcfd1a59f764bad2b2690
 1.46 KiB / 1.46 KiB [======================================================] 0s
Storing signatures
ERRO[0012] Error while applying layer: ApplyLayer exit status 1 stdout:  stderr: permission denied
error creating build container: The following failures happened while trying to pull image specified by "busybox" based on search registries in /etc/containers/registries.conf:
* "localhost/busybox": Error determining manifest MIME type for docker://localhost/busybox:latest: pinging docker registry returned: Get https://localhost/v2/: dial tcp 127.0.0.1:443:
connect: connection refused
* "registry.fedoraproject.org/busybox": Error determining manifest MIME type for docker://registry.fedoraproject.org/busybox:latest: Error reading manifest latest in registry.fedorapr$
ject.org/busybox: manifest unknown: manifest unknown
* "registry.access.redhat.com/busybox": Error determining manifest MIME type for docker://registry.access.redhat.com/busybox:latest: Error reading manifest latest in registry.access.re
dhat.com/busybox: unknown: Not Found
* "docker.io/library/busybox": Error committing the finished image: error adding layer with blob "sha256:90e01955edcd85dac7985b72a8374545eac617ccdddcc992b732e43cd42534af": ApplyLayer e
xit status 1 stdout:  stderr: permission denied
```
Unless I remove the `apparmor` on the host to perform correctly.
```
apt-get remove apparmor --purge
```
```
root@2ae4c03ed239:/# buildah bud --isolation chroot .
STEP 1: FROM busybox
STEP 2: RUN ls
bin   dev   etc   home  proc  root  sys   tmp   usr   var
STEP 3: COMMIT containers-storage:[vfs@/root/.local/share/containers/storage+/run/user/0]@d7a546c3c334410e8775810d18f4db06e62e5b2f90d5a8e055cb3a18040875bf
Skipping fetch of repeat blob sha256:8a788232037eaf17794408ff3df6b922a1aedf9ef8de36afdae3ed0b0381907b
Copying blob sha256:7e750bc655543fe5e28c56e2ef5d2c33033008783143df8f0525d90391ded658
 116 B / 116 B [============================================================] 0s
Copying config sha256:586d2a315d4cbde4c8814a99489039e9ffe246d306b5195d2663c8b4c91fa28f
 693 B / 693 B [============================================================] 0s
Storing signatures
--> d7a546c3c334410e8775810d18f4db06e62e5b2f90d5a8e055cb3a18040875bf
```
 I don't think this is the best solution.

q384566678:
It looks similar to the problem in #1138 . @olaf-meyer Do you have any suggestions?

rhatdan:
Does it work with --security-opt apparmor=unconfined

q384566678:
>Does it work with --security-opt apparmor=unconfined

Yes. After adding this option, it will run successfully.
This will disable ` apparmor`, will there be security issues?

rhatdan:
I do SELinux, no idea what apparmor is doing as far as protecting against escape.

PR #1172, opened 2018-11-12
Conformance test improvement

ypu:
This pullreq is used for improving conformance test.
1. Fix the wrong regex in multi stage base test
2. Add Hostname and ContainerConfig:Cmd to skip word list for inspect check step.
3. In inspect check step, accept nil equal to empty array or map.
4. Also print command line in SystemExec function to add more debug info.
5. Fix the default docker build command to fit container-diff request

### DIFF
--- a/tests/conformance/conformance_suite_test.go
+++ b/tests/conformance/conformance_suite_test.go
@@ -466,6 +466,7 @@ func (s *BuildahTestSession) LineInOuputContains(term string) bool {
 
 // SystemExec is used to exec a system command to check its exit code or output
 func SystemExec(command string, args []string) *BuildahTestSession {
+	fmt.Printf("Running: %s %s\n", command, strings.Join(args, " "))
 	c := exec.Command(command, args...)
 	session, err := gexec.Start(c, GinkgoWriter, GinkgoWriter)
 	if err != nil {
--- a/tests/conformance/conformance_suite_test.go
+++ b/tests/conformance/conformance_suite_test.go
@@ -341,14 +341,28 @@ func CompareJSON(a, b map[string]interface{}, skip []string) ([]string, []string
 			continue
 		}
 		if reflect.TypeOf(v) != reflect.TypeOf(vb) {
+			if reflect.TypeOf(v) == nil && reflect.ValueOf(vb).Len() == 0 {
+				continue
+			} else if reflect.TypeOf(vb) == nil && reflect.ValueOf(v).Len() == 0 {
+				continue
+			}
 			diffKeys = append(diffKeys, diffDebug(k, v, vb))
 			isSame = false
 			continue
 		}
 		switch v.(type) {
 		case map[string]interface{}:
+			nextSkip := skip
+			for _, s := range skip {
+				if strings.Contains(s, ":") {
+					tmp := strings.Split(s, ":")
+					if tmp[0] == k {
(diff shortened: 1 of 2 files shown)


tests/conformance/conformance_suite_test.go
```diff
@@ -341,14 +341,29 @@ func CompareJSON(a, b map[string]interface{}, skip []string) ([]string, []string
 			continue
 		}
 		if reflect.TypeOf(v) != reflect.TypeOf(vb) {
+			if (reflect.TypeOf(v) == nil && reflect.ValueOf(vb).Len() == 0) {
+				continue
+			} else if (reflect.TypeOf(vb) == nil && reflect.ValueOf(v).Len() == 0) {
```
rhatdan: No need for else here.

tests/conformance/conformance_suite_test.go
```diff
@@ -341,14 +341,29 @@ func CompareJSON(a, b map[string]interface{}, skip []string) ([]string, []string
...
 		}
 		switch v.(type) {
 		case map[string]interface{}:
+			nextSkip := skip
+			for _, s := range skip {
+				if strings.Contains(s, ":") {
+					tmp := strings.Split(s, ":")
+					if tmp[0] == k {
+						nextSkip = append(nextSkip, strings.Join(tmp[1:], ":"))
+
```
rhatdan: nit: remove empty line

rhatdan:
A couple of nits, Fix these and LGTM

TomSweeneyRedHat:
@ypu looks like you need to run `gofmt -s -w ./tests/conformance/conformance_suite_test.go`.  However make sure to use gofmt from go v1.11 or higher, gofmt from v1.10 or lower is no longer valid.

TomSweeneyRedHat:
LGTM assuming happy tests

tests/conformance/conformance_suite_test.go
ypu: Hi @rhatdan 
As currently we call this functions like CompareJSON(dockerimage, buildahimage, skip) we don't need else here. But if we consider this functions as common use function, we don't know everyone will call it with such order. So I'd like keep this here to cover all kinds of situation.

tests/conformance/conformance_suite_test.go
ypu: Got it. Thanks a lot. Will remove it.

ypu:
Thanks @rhatdan and @TomSweeneyRedHat . Pullreq updated.

tests/conformance/conformance_suite_test.go
rhatdan: sorry I wasn't very clear.  All I meant is to remove the word `else`  Since the first block executes a continue. it will not execute the the second block.  
I find `else if` harder to read and don't want them when they are not necessary.
``
		if reflect.TypeOf(v) != reflect.TypeOf(vb) {
			if (reflect.TypeOf(v) == nil && reflect.ValueOf(vb).Len() == 0) {
				continue
			} 
                        if (reflect.TypeOf(vb) == nil && reflect.ValueOf(v).Len() == 0) {
```

ypu: Oh. Thanks a lot for your explanation. Get the point now. Will format a patch to fix this.

Issue #1179, opened 2018-11-15
multi-stage build tags old image / uses cache incorrectly


ansemjo:
**Description**

I am building an application consisting of frontend and backend into a single container with a multi-staged Dockerfile, where the last stage simply copies files from the first two stages. I am using rootless `podman build` to build my images. These images will later be deployed to a [flynn](https://flynn.io/) cluster.

**Steps to reproduce the issue:**
1. have a multi-stage dockerfile with a final stage which only depends on intermediate stages (example appended)
2. build the image with `podman build -t app:1.0 .`
3. make changes to your app sources, i.e. new version: `2.0`
4. rebuild the image with `podman build -t app:2.0 .`
5. check `podman images`

**Describe the results you received:**
The new tag `app:2.0` was tagged on the old image. The build did __not__ use caches for the first two stages since the sources were changed. However __all__ build steps in the final stage used the cache, thus the built image is the same as the old tag. I have two untagged (`<none>`) images built, which I suppose are the intermediate stages.

```
$ podman images
REPOSITORY                              TAG                IMAGE ID       CREATED             SIZE
<none>                                  <none>             fd181c7fbfc0   40 minutes ago      382MB
<none>                                  <none>             5e0e119512d3   43 minutes ago      743MB
localhost/batchcards.js                 2.5.3              269aa8e1878e   About an hour ago   105MB
localhost/batchcards.js                 2.5.2-1-g0d5a57c   269aa8e1878e   About an hour ago   105MB
<none>                                  <none>             45a39d3e6440   About an hour ago   382MB
<none>                                  <none>             f921bc100c63   About an hour ago   743MB
```

**Describe the results you expected:**
The final stage should detect that the first two stages are different and not use any cache.

**Other notes:**
I have two other (unrelated) observations with multi-stage builds:

- `buildah` does not accept `ARG`s before the first `FROM`. With Docker you could make the image used in `FROM` an `ARG` itself, i.e. when all three stages shall use `FROM node:10-alpine` you could use: `ARG IMAGE=node:10-alpine` and `FROM ${IMAGE}` instead.
- If the last stage uses the `AS` keyword (i.e. `FROM node:10-alpine AS runtime`) the built image will not be tagged at all and I will just have a bunch of `<none>` images.

---

**Output of ~`rpm -q buildah` or `apt list buildah`~ `pacman -Qi buildah`:**

```
$ pacman -Qi buildah
Name            : buildah
Version         : 1.4-2
[...]

PR #1181, opened 2018-11-17
Handle 'COPY --from' in Dockerfile

TomSweeneyRedHat:
Signed-off-by: TomSweeneyRedHat <tsweeney@redhat.com>

Handle the "COPY --from" directives in a Dockerfile.  If we discover this line in our Dockerfiles, the code will create a  new FROM directive for the listed image which will cause the image to be pulled prior to the COPY command is executed.  Prior this was not happening.

Fixes: #1175 

Testing:
```
# cat ~/Dockerfile.copyfrom
FROM php:7.2 
COPY --from=composer:latest /usr/bin/composer /usr/local/bin/composer

# buildah bud -f ~/Dockerfile.copyfrom -t tom .
STEP 1: FROM composer:latest
Getting image source signatures
Copying blob sha256:4fe2ade4980c2dda4fc95858ebb981489baec8c1e4bd282ab1c3560be8ff9bde
 2.10 MiB / 2.10 MiB [======================================================] 0s
Copying blob sha256:5cf60c0c0763e6e5b51f7b06d1a681fbc422c4feadba6e28e0354de32fe18743
 1.29 MiB / 1.29 MiB [======================================================] 0s
Copying blob sha256:468219aa46893af0196f1629ab668d8120e76afb69af0c3cc51a0c0caeff594e
 1.22 KiB / 1.22 KiB [======================================================] 0s
Copying blob sha256:4a8b1a3b9e510caf57ea183cacd59ad9f85a026dac10eb86f68fec44d0dfda14
 169 B / 169 B [============================================================] 0s
Copying blob sha256:da2c88c4ca27c759936b47b3d60f6bc7e1996016686abb20992caa9d23e63ad1
 11.60 MiB / 11.60 MiB [====================================================] 1s
Copying blob sha256:c4f408a591350b599d1d9e2f510961e37188d566263750a6fefd764410ec2406
 497 B / 497 B [============================================================] 0s
Copying blob sha256:d65ea60b5d6c1c72c736fe5c73e87c2546ea7a7ff3016aa9742096d4d2b7ac83
 15.27 MiB / 15.27 MiB [====================================================] 2s
Copying blob sha256:7e266b5abba2ce2cf6c6db0e5db503e1929585ecafe556360ba1d339776b890a
 2.13 KiB / 2.13 KiB [======================================================] 0s
Copying blob sha256:bb41dfb37ff5079cb2e644061f6d0eead3bb3c34c9746962ae0624f7064e6718
[...]

PR #1182, opened 2018-11-18
Fix cache issue with COPY/ADD --from

umohnani8:
In multi-stage builds, it is possible to copy data from
a previous build. When using cached images, buildah wasn't
able to detect a change in the previous build and would proceed
to use cached images. Fix buildah bud to be keep track of any
files coming from a previous build and checking whether the cached
image is still valid or not.
Also fix removal of intermediate build containers when building
from cache. We were not keeping track of those containers so they
weren't getting deleted at the end of the build process.

Fixes https://github.com/containers/buildah/issues/1179

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -222,7 +222,7 @@ type Executor struct {
 	forceRmIntermediateCtrs        bool
 	containerIDs                   []string          // Stores the IDs of the successful intermediate containers used during layer build
 	imageMap                       map[string]string // Used to map images that we create to handle the AS construct.
-
+	copyFrom                       string            // Used to keep track of the --from flag from COPY and ADD
 }
 
 // withName creates a new child executor that will be used whenever a COPY statement uses --from=NAME.
@@ -826,6 +826,18 @@ func (b *Executor) Execute(ctx context.Context, stage imagebuilder.Stage) error 
 			err     error
 			imgID   string
 		)
+
+		b.copyFrom = ""
+		// Check if --from exists in the step command of COPY or ADD
+		// If it exists, set b.copyfrom to that value
+		for _, n := range step.Flags {
+			if strings.Contains(n, "--from") && (step.Command == "copy" || step.Command == "add") {
+				arr := strings.Split(n, "=")
+				b.copyFrom = b.named[arr[1]].mountPoint
+				break
+			}
+		}
+
 		// checkForLayers will be true if b.layers is true and a cached intermediate image is found.
 		// checkForLayers is set to false when either there is no cached image or a break occurs where
 		// the instructions in the Dockerfile change from a previous build.
@@ -848,6 +860,7 @@ func (b *Executor) Execute(ctx context.Context, stage imagebuilder.Stage) error 
 			if err := b.copyExistingImage(ctx, cacheID); err != nil {
 				return err
 			}
+			b.containerIDs = append(b.containerIDs, b.builder.ContainerID)
 			break
 		}
 
@@ -1009,6 +1022,11 @@ func (b *Executor) getFilesToCopy(node *parser.Node) ([]string, error) {
 			currNode = currNode.Next
 			continue
 		}
+		if b.copyFrom != "" {
+			src = append(src, filepath.Join(b.copyFrom, currNode.Value))
+			currNode = currNode.Next
+			continue
+		}
 		matches, err := filepath.Glob(filepath.Join(b.contextDir, currNode.Value))
 		if err != nil {
 			return nil, errors.Wrapf(err, "error finding match for pattern %q", currNode.Value)
@@ -1049,7 +1067,12 @@ func (b *Executor) copiedFilesMatch(node *parser.Node, historyTime *time.Time) (
 		// Change the time format to ensure we don't run into a parsing error when converting again from string
 		// to time.Time. It is a known Go issue that the conversions cause errors sometimes, so specifying a particular
 		// time format here when converting to a string.
-		timeIsGreater, err := resolveModifiedTime(b.contextDir, item, historyTime.Format(time.RFC3339Nano))
+		// If the COPY has --from in the command, change the rootdir to mountpoint of the container it is copying from
+		rootdir := b.contextDir
+		if b.copyFrom != "" {
+			rootdir = b.copyFrom
+		}
+		timeIsGreater, err := resolveModifiedTime(rootdir, item, historyTime.Format(time.RFC3339Nano))
 		if err != nil {
 			return false, errors.Wrapf(err, "error resolving symlinks and comparing modified times: %q", item)
 		}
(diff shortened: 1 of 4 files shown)


umohnani8:

umohnani8:
I thinks #1181 is handling a different issue with COPY --from, @TomSweeneyRedHat please correct me if I am mistaken.

TomSweeneyRedHat:
Ugh, thought I'd replied.  Yes this appears to be a different issue.  The issue I'm working on happens when there's a "COPY --from={image}" command in the Dockerfile, but no "FROM {image}" prior to that in the Dockerfile.  I'll go a more thorough review of this after my next meeting, but a quick glance looks good and green test buttons.

rhatdan:
LGTM

imagebuildah/build.go
```diff
@@ -826,6 +826,18 @@ func (b *Executor) Execute(ctx context.Context, stage imagebuilder.Stage) error
 			err     error
 			imgID   string
 		)
+
+		b.copyFrom = ""
+		// Check if --from exists in the step command of COPY or ADD
+		// If it exists, assing b.copyfrom that value
```
TomSweeneyRedHat: I'm sorry, but "assing"?

TomSweeneyRedHat: Maybe "pass"

imagebuildah/build.go
```diff
@@ -826,6 +826,18 @@ func (b *Executor) Execute(ctx context.Context, stage imagebuilder.Stage) error
...
+		for _, n := range step.Flags {
+			if strings.Contains(n, "--from") && (step.Command == "copy" || step.Command == "add") {
+				arr := strings.Split(n, "=")
+				b.copyFrom = b.named[arr[1]].mountPoint
+				break
+			}
+		}
+
```
TomSweeneyRedHat: Is this happy if you have a Dockerfile with multiple 'COPY --from={image}' lines like:

```
FROM scratch as myname
COPY Dockerfile.name /

FROM scratch as myname2
COPY Dockerfile.index /

FROM scratch
COPY --from=myname /Dockerfile.name /Dockerfile.name
COPY --from=myname2 /Dockerfile.index /Dockerfile.index
```

tests/bud/use-layers/Dockerfile.multistage-copy
```diff
@@ -0,0 +1,9 @@
+FROM alpine AS uuid
+COPY uuid /src
+
+FROM alpine AS date
+COPY date /src
+
+FROM alpine
+COPY --from=uuid /src/data /uuid
+COPY --from=date /src/data /date
```
TomSweeneyRedHat: Would be nice to have some checks here to make sure the files got in the right places.

TomSweeneyRedHat: FWIW, you could do something like #1182 

```
ctr=$(buildah --debug=false from --signature-policy ${TESTSDIR}/policy.json ${target})
  mnt=$(buildah --debug=false mount ${ctr})
  run test -e $mnt/usr/local/bin/composer
```
which is something I stole from a @nalind suggestion several reviews ago.

imagebuildah/build.go
umohnani8: yup, that is the what the test Dockerfile looks like.

tests/bud/use-layers/Dockerfile.multistage-copy
umohnani8: oh okay, will add such a check.

imagebuildah/build.go
umohnani8: Lol, didn't mean that, typo.

imagebuildah/build.go
TomSweeneyRedHat: I see, I missed this logic is within the loop for the child, you get a new b.copyFrom with each child, I was thinking it was only one piece of memory for all.  This works.

tests/bud.bats
```diff
@@ -59,6 +59,36 @@ load helpers
...
+  [ $(wc -l <<< "$output") -eq 6 ]
+  [ "${status}" -eq 0 ]
+  buildah bud --signature-policy ${TESTSDIR}/policy.json --layers -t test2 -f Dockerfile.multistage-copy ${TESTSDIR}/bud/use-layers
+  run buildah --debug=false images -a
+  [ $(wc -l <<< "$output") -eq 7 ]
+  [ "${status}" -eq 0 ]
+
+  uuidgen > ${TESTSDIR}/bud/use-layers/uuid/data
+  date > ${TESTSDIR}/bud/use-layers/date/data
+  buildah bud --signature-policy ${TESTSDIR}/policy.json --layers -t test3 -f Dockerfile.multistage-copy ${TESTSDIR}/bud/use-layers
```
TomSweeneyRedHat: It would probably be good to add a 'test4' that dropped the '--layers' param from the buildah bud command.

TomSweeneyRedHat:
I've tested and it looks good.  If you could touch up the comment and tweak the tests a bit as noted, then it will be gtg.

Nice change @umohnani8 !

TomSweeneyRedHat:
@umohnani8 gofmt errors, did you use gofmt v1.10 instead fo gofmt v1.11?

imagebuildah/build.go
```diff
@@ -563,39 +563,39 @@ func NewExecutor(store storage.Store, options BuildOptions) (*Executor, error) {
...
+		commonBuildOptions:      options.CommonBuildOpts,
+		defaultMountsFilePath:   options.DefaultMountsFilePath,
+		iidfile:                 options.IIDFile,
+		squash:                  options.Squash,
+		labels:                  append([]string{}, options.Labels...),
+		annotations:             append([]string{}, options.Annotations...),
+		layers:                  options.Layers,
+		noCache:                 options.NoCache,
+		removeIntermediateCtrs:  options.RemoveIntermediateCtrs,
+		forceRmIntermediateCtrs: options.ForceRmIntermediateCtrs,
```
TomSweeneyRedHat: This smells like gofmt 1.10 vs 1.11+

tests/bud.bats
```diff
@@ -59,6 +59,48 @@ load helpers
...
+  [ "${status}" -eq 0 ]
+  run buildah --debug=false containers
+  [ $(wc -l <<< "$output") -eq 1 ]
+  [ "${status}" -eq 0 ]
+  
+  ctr=$(buildah --debug=false from --signature-policy ${TESTSDIR}/policy.json test3)
+  mnt=$(buildah --debug=false mount ${ctr})
+  run test -e $mnt/uuid
+  [ "${status}" -eq 0 ]
+  run test -e $mnt/date
```
TomSweeneyRedHat: :+1:

TomSweeneyRedHat:
Once this gets happy green test buttons, then this is gtg.

imagebuildah/build.go
umohnani8: yup, vscode messing up again. Fixed

TomSweeneyRedHat:

TomSweeneyRedHat:
@rh-atomic-bot r+

PR #1185, opened 2018-11-19
pull: add iidfile option

q384566678:
Add iidfile option to write the image ID to the specified file.

### DIFF
--- a/cmd/buildah/pull.go
+++ b/cmd/buildah/pull.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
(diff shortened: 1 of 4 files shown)


q384566678:
@vrothberg I refer to the bud's iidfile option. I just think that adding this feature is convenient for the user.
I  can accept both methods, whether it is added or not.

rhatdan:
I think we should just redirect stdio to the file, rather then add another option.  Especially where `docker pull` does not have an equivalent.

PR #1188, opened 2018-11-20
vendor: update selinux

giuseppe:
inherit a change for not failing a recursive relabelling if the file
is removed between the directory is read and the lsetxattr syscall.


rhatdan:
I was just doing a similar vendor.  Will merge this as soon as it passes.
Would fix this issue on buildah as well.
https://github.com/containers/libpod/issues/1826

rhatdan:
@rh-atomic-bot r+

Issue #1199, opened 2018-11-26
images dangling filter does not work in 1.5


carlwgeorge:
**Description**

After updating to buildah 1.5, `buildah images --filter dangling=true` no longer works correctly.

**Steps to reproduce the issue:**
1. `container=$(buildah from scratch)`
2. `buildah commit $container test`
3. `buildah commit $container test` (yes, a second time)
4. `buildah images --filter dangling=true`

**Describe the results you received:**
No images are listed.

**Describe the results you expected:**
One untagged image should be listed.  This is what happens with `buildah-1.4-3.git608fa84.fc29.x86_64`.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.5-1.gite94b4f9.fc29.x86_64
```

<details>
<summary><strong>Output of <code>buildah version</code>:</strong></summary>

```
Version:         1.5
Go Version:      go1.11.2
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.4.0
libcni Version:  
Git Commit:      
Built:           Thu Jan  1 00:00:00 1970
OS/Arch:         linux/amd64
```
</details>

<details>
<summary><strong>Output of <code>cat /etc/*release</code>:</strong></summary>

```
Fedora release 29 (Twenty Nine)
NAME=Fedora
VERSION="29 (Twenty Nine)"
ID=fedora
VERSION_ID=29
PLATFORM_ID="platform:f29"
PRETTY_NAME="Fedora 29 (Twenty Nine)"
ANSI_COLOR="0;34"
CPE_NAME="cpe:/o:fedoraproject:fedora:29"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f29/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=29
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=29
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
Fedora release 29 (Twenty Nine)
Fedora release 29 (Twenty Nine)
```
</details>

<details>
[...]

PR #1207, opened 2018-11-28
Enable -q when --filter is used for images command

TomSweeneyRedHat:

When the images command was reworked recently to handle multiple names for an image, the logic for quiet and json checking was moved up out of a loop that went through all of the multiple names of an image to make things a bit more expedient.  However the filtering logic did not move too so when we now do a `buildah images -q --filter dangling=true` command, the filter is never considered.  Ditto `--json` and `-q` combined.

The code to do the json and quiet checks is still within the loop for multiple names along with the filter checks.  Although less expedient to use there, it is almost immeasurably slower.  The logic for the filtering has some dependencies within the loop and can't be easily moved outside of it.

@QiWang19 PTAL and verify my thinking here.

Finishes addressing #1199 (he hopes!)

### DIFF
--- a/cmd/buildah/images.go
+++ b/cmd/buildah/images.go
@@ -273,20 +273,7 @@ func outputImages(ctx context.Context, images []storage.Image, store storage.Sto
 			// images without names should be printed with "<none>" as the image name
 			names = append(names, "<none>:<none>")
 		}
-		if opts.quiet && argName == "" {
-			fmt.Printf("%-64s\n", image.ID)
-			// We only want to print each id once
-			continue
-		}
-		if opts.json && argName == "" {
-			JSONImage := jsonImage{ID: image.ID, Names: image.Names}
-			data, err2 := json.MarshalIndent(JSONImage, "", "    ")
-			if err2 != nil {
-				return err2
-			}
-			fmt.Printf("%s\n", data)
-			continue
-		}
+		breakOuter := false
 		for name, tags := range imagebuildah.ReposToMap(names) {
 			for _, tag := range tags {
 				if !matchesReference(name+":"+tag, argName) {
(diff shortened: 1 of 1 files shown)


QiWang19:
I think `bud.bats` failed, because this `buildah --debug=false images -q` returns more than `image iD` more than once inside the for loop. That's the corner case.

rhatdan:
LGTM

TomSweeneyRedHat:
Thanks @QiWang19 and hmmm, that worked in my sandbox before I pushed the PR.  Will go verify.

QiWang19:
@TomSweeneyRedHat 🙂 not sure that's a bug in bud.bats or `buildah rmi`, I found current podman doesn't allow `podman rmi -f $(podman images -q) ` if image has multiple tags

TomSweeneyRedHat:
@QiWang19 looks to be a bug with images @QiWang19 which I'm going to guess podman has too.  The issue is:

```
# buildah bud --signature-policy ${TESTSDIR}/policy.json -t ${target} -t docker.io/${target2} -t ${target3} ${TESTSDIR}/bud/from-scratch

caused images to give you: 

# buildah images -q
0e00913c1ac3c6dde8dac5038009025b43eddf863a2dc7c62b6b6503252ec29c
0e00913c1ac3c6dde8dac5038009025b43eddf863a2dc7c62b6b6503252ec29c
0e00913c1ac3c6dde8dac5038009025b43eddf863a2dc7c62b6b6503252ec29c
```
When it should only show one image.  @mheon I bet Podman has the same issue, I'll chase it down next week after my PTO if someone doesn't beat me to it.  The fix appears to be to add a boolean that you set to true when you hit your first quiet and then break out of the secondary `for` loop when the boolean is true.

We'll see if the tests like it, but they look good in my sandbox.

TomSweeneyRedHat:
Happy green test buttons. @QiWang19 and @rhatdan, PTAL

QiWang19:
:+1: looks good to me

rhatdan:
@rh-atomic-bot r+

Issue #987, opened 2018-09-04
"push docker-daemon:" always uses "docker.io/" prefix


tliron:
Tested with version 1.3.

Example:
```
buildah push c1589518a3be docker-daemon:prefix/test:1.0
```
And then running `docker images` you will the name as:
```
docker.io/prefix/test
```
This makes it difficult to use buildah with non-public Docker repositories.

rhatdan:
So you would rather see 
localhost/prefix/test

tliron:
Well, I would rather have no added extra prefix, as this much complicates the rest of the CI/CD chain. Other Docker building tools do not tend to add extra prefixes.

rhatdan:
@mtrmac @umohnani8 PTAL

mtrmac:
This is intentional and necessary for `atomic pull` to work as expected (or as close to it as possible, anyway).  See https://github.com/containers/image/issues/72 for the complex details if you are interested (perhaps start with the tables at the bottom).

hostname-unprefixed destinations for `docker-daemon:` are fundamentally equivalent to `docker.io/*` on upstream Docker, and not interoperable (different from `docker pull` / `docker tag`) on projectatomic/docker.

If buildah wanted to implement `buildah push --to-daemon prefix/test` which implicitly treats it as `localhost/` instead of `docker.io/`, that would be possible (_please_ don’t edit `docker-daemon:` strings to reinterpret their semantics, though); the desired support for ~ `buildah push --to-daemon prefix/test` to not add _any_ prefix is not supportable.

tliron:
Well. All I can say is that the assumption that everyone will want to push to `docker.io` is wrong. Many of us use internal repositories.

I wonder if perhaps we can support multiple use cases via an extra CLI switch? E.g.:
```
buildah push c1589518a3be docker-daemon:prefix/test:1.0 --docker-repository=internal.org
```
Would result in this name:
```
internal.org/prefix/test
```
The default `--docker-repository` could be `docker.io`, so that your other use case (Atomic?) would continue working as it does now.

umohnani8:
@tliron you could also give it the full name including the registry, that way the default docker.io won't be added to the name.
`buildah push image docker-daemon:internal.org/prefix/test:1.0`

rhatdan:
@tliron I think @umohnani8 way to handle this is the best.

tliron:
OK, I did some testing and this indeed does work:
```
buildah push image docker-daemon:internal.org/prefix/test:1.0
```
[...]

PR #1221, opened 2018-12-05
Clarify docker.io default in push with docker-daemon

TomSweeneyRedHat:

Touches up docker-deamon verbiage per @tliron.  Thank you for the suggested verbiage and apologies for taking so long to get this done!

Fixes: #987

### DIFF
--- a/docs/buildah-push.md
+++ b/docs/buildah-push.md
@@ -30,7 +30,7 @@ Image stored in local container/storage
   An image is stored in the `docker save` formatted file.  _docker-reference_ is only used when creating such a file, and it must not contain a digest.
 
   **docker-daemon:**_docker-reference_
-  An image _docker-reference_ stored in the docker daemon internal storage.  _docker-reference_ must contain either a tag or a digest.  Alternatively, when reading images, the format can also be docker-daemon:algo:digest (an image ID).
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
@mtrmac @nalind PTAL

mtrmac:
👍 Thanks for taking care of this!

vrothberg:
LGTM but I have some follow up questions.

- Do we need the same change for libpod?
- Does a similar behavior apply to other transports? If so, we could create a dedicate manpage for those transports (maybe in c/image?).

mtrmac:
The `docker.io/library` normalization is common, but does not typically happen for `Buildah`/`Podman`; and `docker-daemon:` is special in that it has a particularly aggressive version of it, where it does not only _interpret_ short names with this semantics, but actively normalizes the name to explicitly include the hostname when contacting the daemon.

Yes, c/image should be maintaining a documentation of all its transports, and I can’t see it in there.  IIRC historically it lived in https://github.com/containers/skopeo/blob/master/docs/skopeo.1.md , and yes it should be split into a separate man page referenced from the others.

rhatdan:
@rh-atomic-bot r+

Issue #1225, opened 2018-12-06
[rootless & fuse-ovl] mount command makes most directories disappear rendering the container useless


TomasTomecek:
**Description**
`mount` command seems to be a magician:

```
$ buildah from registry.fedoraproject.org/fedora:29
fedora-working-container

$ buildah run fedora-working-container ls /
bin  boot  dev  etc  home  lib  lib64  lost+found  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
```

so far so good

```
$ buildah mount fedora-working-container
/home/tt/.local/share/containers/storage/overlay/ecf5d2cc2ea73f428dd2aef0f5159af1b97e7c1ea555e50a02adb674bc776ed3/merged

$ ls -lha /home/tt/.local/share/containers/storage/overlay/ecf5d2cc2ea73f428dd2aef0f5159af1b97e7c1ea555e50a02adb674bc776ed3/merged
total 0
drwx------. 6 tt tt 51 Dec  6 17:31 .
drwx------. 5 tt tt 69 Dec  6 17:30 ..
drwxr-xr-x. 2 tt tt  6 Dec  6 17:31 dev
drwxr-xr-x. 2 tt tt 38 Dec  6 17:31 etc
drwxr-xr-x. 2 tt tt  6 Dec  6 17:31 proc
drwxr-xr-x. 2 tt tt  6 Dec  6 17:31 sys
```

where is the rest?

```
$ buildah run fedora-working-container ls /
container_linux.go:337: starting container process caused "exec: \"ls\": executable file not found in $PATH"
error running container: error creating container for [ls /]: : exit status 1
error while running runtime: exit status 1
ERRO[0000] exit status 1
```

It's true, whole `/usr` is gone.

Can this be an issue of fuse-overlayfs?
```
$ cat ~/.config/containers/storage.conf
RunRoot = "/run/user/1000"
GraphRoot = "/home/tt/.local/share/containers/storage"
GraphDriverName = "overlay"
GraphDriverOptions = ["overlay.mount_program=/usr/bin/fuse-overlayfs"]

$ buildah version
Version:         1.5
Go Version:      go1.11.2
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.4.0
libcni Version:
Git Commit:
Built:           Thu Jan  1 01:00:00 1970
OS/Arch:         linux/amd64
```

giuseppe:
[...]

PR #1233, opened 2018-12-12
mount: allow mount only when using vfs

giuseppe:
when using a driver different than vfs, the mount is probably in a
different mount namespace thus not accessible from the host.  Avoid
the confusion by not allowing mount when a different driver is used.

Closes: https://github.com/containers/buildah/issues/1225

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ script:
     - docker ps --all
     - docker images
     - ls -alF /home/travis/auth
-    - docker pull alpine
+    - docker pull docker.io/alpine
     - echo testpassword | docker login localhost:5000 --username testuser --password-stdin
     - docker tag alpine localhost:5000/my-alpine
     - docker push localhost:5000/my-alpine
--- a/cmd/buildah/mount.go
+++ b/cmd/buildah/mount.go
@@ -48,6 +48,14 @@ func mountCmd(c *cli.Context) error {
 
 	var lastError error
 	if len(args) > 0 {
+		// Do not allow to mount a graphdriver that is not vfs if we are creating the userns as part
+		// of the mount command.
+		// Differently, allow the mount if we are already in a userns, as the mount point will still
+		// be accessible once "buildah mount" exits.
+		if os.Getenv(startedInUserNS) != "" && store.GraphDriverName() != "vfs" {
+			return fmt.Errorf("cannot mount using driver %s in rootless mode", store.GraphDriverName())
+		}
+
 		for _, name := range args {
 			builder, err := openBuilder(getContext(), store, name)
 			if err != nil {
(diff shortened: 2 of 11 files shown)


TomSweeneyRedHat:
@giuseppe gofmt error, be sure to run gofmt 1.11 or higher.
```
Error: source files are not formatted according to recommendations. Run "gofmt -s -w" on:
./cmd/buildah/unshare_unsupported.go
```

rhatdan:
Can I do a 
```
buildah unshare
mnt=$(buildah mount $ctr)
dnf -y install installroot=$mnt 
cp foo $mnt
buildah umount $mnt
exit
buildah commit $ctr imagename
buildah push imagename foobar
```

This is probably a reason we should allow mechanisms for buildah to specify whether or not it wants to use the OverlayFS when running as nonroot.

giuseppe:
> Can I do a
> 
> ```
> buildah unshare
> mnt=$(buildah mount $ctr)
> dnf -y install installroot=$mnt 
> cp foo $mnt
> buildah umount $mnt
> exit
> buildah commit $ctr imagename
> buildah push imagename foobar
> ```
> 
> This is probably a reason we should allow mechanisms for buildah to specify whether or not it wants to use the OverlayFS when running as nonroot.

yes, you can do that.

Yes, in general VFS is safer, slower but safer.  Fortunately Buildah allows to use both storages so one can always switch between overlay and vfs.

TomasTomecek:
Do you think it would make sense to add this fix also to ansible so that people are able to use the buildah connection plugin with rootless+ovl? I tried to do `buildah unshare` as root, and it seemed to work correctly -- so maybe the unshare could be the default for running commands in that ansible plugin both root and rootless.

giuseppe:
> Do you think it would make sense to add this fix also to ansible so that people are able to use the buildah connection plugin with rootless+ovl? I tried to do `buildah unshare` as root, and it seemed to work correctly -- so maybe the unshare could be the default for running commands in that ansible plugin both root and rootless.

root won't need to do an unshare, as the root user is able to create mounts on the host namespace.

About rootless, it is only requires for `mount` and manually operating on the mounted directory.  It is not required for commands like `add` and `copy`.

Also it is not required when using `bud` as implicitly it does the unshare and operates all the commands in the new created namespaces.

rhatdan:
@giuseppe Why did you have to add the fully qualified PRs?

Also we need to document this change in the man page, and tell people how they could do the buildah mount command while not root.

giuseppe:
@rhatdan tests were failing as buildah couldn't find the image on registry.access.redhat.com

giuseppe:
@rhatdan I've added another commit to update the man page

rhatdan:
@rh-atomic-bot r+

PR #1238, opened 2018-12-14
Handle one line Dockerfile with layers

TomSweeneyRedHat:

Addresses https://github.com/containers/libpod/issues/1993 once merged there.

Podman turns layers on by default with every build. The layer logic does not behave if there's only a one line Dockerfile given to the build, a commit is never done as the layer logic doesn't do a commit because it's waiting for the last of two ore more layers before doing so. If we have only one line in one Dockerfile that we're building, the most layers we will have is one, so just ignore layering.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -1307,7 +1307,12 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (strin
 
 	var imageRef reference.Canonical
 	imageID := ""
-	if !b.layers && !b.noCache {
+
+	// Check if we have a one line Dockerfile making layers irrelevant
+	// or the user told us to ignore layers.
+	ignoreLayers := (len(stages) < 2 && len(stages[0].Node.Children) < 2) || (!b.layers && !b.noCache)
+
+	if ignoreLayers {
 		imgID, ref, err := stageExecutor.Commit(ctx, stages[len(stages)-1].Builder, "")
 		if err != nil {
 			return "", nil, err


TomSweeneyRedHat:
@mheon FYI

imagebuildah/build.go
```diff
@@ -1257,6 +1257,13 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (strin
 		stageExecutor *Executor
 		lastErr       error
 	)
+
+	//Check if we have a one line Dockerfile, if so, layers are irrelevant
+	ignoreLayers := false
```
rhatdan: Couldn't we just set b.layers or b.noCache to true?

TomSweeneyRedHat: I think we could, but I'm always hesitant to overwrite a field that was sent in as input from the caller.

rhatdan: How about 
ignoreLayers =  (!b.layers && !b.noCache) || ignoreLayers) || (len(stages) < 2 && len(stages[0].Node.Children) < 2)

Then just check below 
if ignoreLayers {\}

TomSweeneyRedHat: I'm massaged it a bit more, moving all the logic downwards.  I could combing the setting of the variable and the if statement, but I like having them separate.  Holler if you feel strongly that they should be merged.

TomSweeneyRedHat:
OK, trying this one again.  Travis CI was stuck overnight and this morning in a pending state.  I repushed a couple of times to see if I could make it move.

TomSweeneyRedHat:
Now registry network flakes.  Git has not been kind to me in 2019.

TomSweeneyRedHat:
bot, retest this please

TomSweeneyRedHat:
bot, retest this please.

rhatdan:
LGTM

TomSweeneyRedHat:
@vrothberg PTAL

vrothberg:
LGTM
Great work, @TomSweeneyRedHat. I think you found the needle in the haystack :)

vrothberg:
@rh-atomic-bot r+

PR #1241, opened 2018-12-17
Add some tests for buildah pull

ypu:
Add some tests for pull image from different transport.


TomSweeneyRedHat:
@ypu thanks for the test additions, but it looks like a couple of them are failing.  Also I don't think you signed the PR. After you correct the test issues, can you do 'git commit --amend -s' to sign the commit please?

ypu:
Hi @TomSweeneyRedHat 

Thanks a lot for your reply. Update the test cases, and now the new added tests are passed. But "from-by-digest-s1" is failed in one of the  test 'BUILDAH_ISOLATION=oci DISTRO="ubuntu"' with TLS handshake timeout.  Not sure if this is a network problem or not as in others tests this case passed.  Will try to rerun the tests to see, if it is OK.
And I checked the commit. Seems it is signed in old and new version. Not sure where is the problem. If it is still not signed in your view please let me know. Will reformat it from the beginning.

bot, retest this please.

PR #1252, opened 2018-12-30
Add make depend, no build if there are no changes

afbjorklund:
Generate depend files for make, using a `go list` script.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 docs/buildah*.1
+*.o
+*.d
 /buildah
 /imgtype
 /build/
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ BINDIR := $(PREFIX)/bin
 BASHINSTALLDIR = $(PREFIX)/share/bash-completion/completions
 BUILDFLAGS := -tags "$(BUILDTAGS)"
 BUILDAH := buildah
+PROJECT := github.com/containers/buildah
 GO := go
 GO110 := 1.10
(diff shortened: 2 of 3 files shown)


afbjorklund:
Signed.

rhatdan:
LGTM
@TomSweeneyRedHat @umohnani8 @nalind @vrothberg PTAL

vrothberg:
@afbjorklund, thanks for the PR! I don't think we need this optimization as `go-build` doesn't recompile when there are no changes. However, it does link the objects which is fairly cheap in comparison to compiler optimizations.

What I find very important is sensitivity to changes to the environment, which the proposed script in this PR is not, see below:

```bash
[valentin@nebuchadnezzar buildah]$ TAGS="123" make buildah
make: 'buildah' is up to date.
```

IMHO we should stick with the go tooling. You can try an incremental compilation by first cleaning the cache `rm -rf ~/.cache/go-build` and then perform two compilations. On my machine the time drops from around 14 seconds to around 2.5 seconds.

afbjorklund:
Yeah, I suppose it makes for a bigger difference with big projects like minikube.

afbjorklund:
For me it takes 4.5 seconds to build (with cache). Running make take 0.8 seconds*.
As a comparison it takes 0.5 seconds to list the depends, and 30 seconds without cache.

\* The main reason why make is so "slow", is because of RCS. Running `make -r` is 10x.
https://www.gnu.org/software/make/manual/html_node/Catalogue-of-Rules.html

PR #1253, opened 2019-01-02
We should only warn if user actually requests Hostname be set in image

rhatdan:
When we set the Hostname to match the container id, we don't want to print
the warning, since the user did not request the hostname being set.

### DIFF
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -244,7 +244,11 @@ func updateConfig(builder *buildah.Builder, c *cli.Context) {
 		builder.SetDomainname(c.String("domainname"))
 	}
 	if c.IsSet("hostname") {
-		builder.SetHostname(c.String("hostname"))
+		name := c.String("hostname")
+		if name != "" && builder.Format == buildah.OCIv1ImageManifest {
+			logrus.Errorf("HOSTNAME is not supported for OCI V1 image format, hostname %s will be ignored. Must use `docker` format", name)
+		}
+		builder.SetHostname(name)
 	}
 	if c.IsSet("onbuild") {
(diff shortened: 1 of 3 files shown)


rhatdan:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1661592  in buildah

rhatdan:
https://github.com/adelton PTAL

rhatdan:
@vrothberg @TomSweeneyRedHat @umohnani8 @nalind PTAL

cmd/buildah/config.go
```diff
@@ -244,7 +244,11 @@ func updateConfig(builder *buildah.Builder, c *cli.Context) {
 		builder.SetDomainname(c.String("domainname"))
 	}
 	if c.IsSet("hostname") {
-		builder.SetHostname(c.String("hostname"))
+		name := c.String("hostname")
+		if name != "" && builder.Format != buildah.Dockerv2ImageManifest {
```
vrothberg: Can we instead use `... != buildah.OCIv1ImageManifest`? Currently, it works as there are only two formats but who knows what the future holds?

TomSweeneyRedHat:
LGTM

cmd/buildah/config.go
rhatdan: So you want to only report the warning, if buildah.Format == buildah.OCIv1ImageManifest, since we know that OCIv1ImageManifest does not support it but OCIV2ImageManifest, might so we would not need to change the code.

rhatdan:
@vrothberg @giuseppe @umohnani8 PTAL

vrothberg:
LGTM

vrothberg:
@rh-atomic-bot r+

PR #1256, opened 2019-01-02
imagebuildah: don't walk if rootdir and path are equal

sjenning:
https://github.com/containers/libpod/issues/1973#issuecomment-449392951

@rhatdan @umohnani8 @TomSweeneyRedHat


TomSweeneyRedHat:
Change LGTM, would like a head nod from @nalind, but if this is needed sooner rather than later, I'm fine with a merge soon, then touch up after @nalind returns if necessary.

rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #1261, opened 2019-01-06
Make rootless work under no_pivot_root

afbjorklund:
Previously only done for root isolation

### DIFF
--- a/run.go
+++ b/run.go
@@ -1130,10 +1130,14 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	case IsolationChroot:
 		err = chroot.RunUsingChroot(spec, path, options.Stdin, options.Stdout, options.Stderr)
 	case IsolationOCIRootless:
+		moreCreateArgs := []string{"--no-new-keyring"}
+		if options.NoPivot {
+			moreCreateArgs = append(moreCreateArgs, "--no-pivot")
+		}
 		if err := setupRootlessSpecChanges(spec, path, rootUID, rootGID); err != nil {
 			return err
 		}
-		err = b.runUsingRuntimeSubproc(isolation, options, configureNetwork, configureNetworks, []string{"--no-new-keyring"}, spec, mountPoint, path, Package+"-"+filepath.Base(path))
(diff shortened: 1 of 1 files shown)


afbjorklund:
@giuseppe does this look OK ?

Same original issue as #1071

giuseppe:
LGTM

cmd/buildah/unshare.go
```diff
@@ -148,6 +148,14 @@ func maybeReexecUsingUserNamespace(c *cli.Context, evenForRoot bool) {
 		}
 	}
 
+	// Copy the no_pivot_root setting, for instance when running on tmpfs.
+	if no_pivot_root, present := os.LookupEnv("BUILDAH_NOPIVOT"); !present {
```
rhatdan: Validate check wants you to use noPivotRoot,  Does not like underscores for variables.

afbjorklund:
Right. Think I managed to get the logic wrong, as well ("present")

Wonder why it worked ? Or maybe it didn't, tested the wrong thing ?

afbjorklund:
Wonder if the unshare stuff was needed at all ? Does os.Environment() copy the existing one ?

cmd/buildah/unshare.go
afbjorklund: Need to get "make validate" working here...

afbjorklund:
Right, don't have to copy any variables - the Environment starts out as a copy  (isolation is an _override_)

run.go
```diff
@@ -1130,10 +1130,16 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	case IsolationChroot:
 		err = chroot.RunUsingChroot(spec, path, options.Stdin, options.Stdout, options.Stderr)
 	case IsolationOCIRootless:
+		var moreCreateArgs []string
```
rhatdan: Pain in the but nit:
var moreCreateArgs []string {"--no-new-keyring"}
if options.NoPivot {
			moreCreateArgs = append(moreCreateArgs, "--no-pivot")
}
Saves three lines.

afbjorklund: But adds a syntax error. See what you mean, though

rhatdan:
@rh-atomic-bot r+

PR #1263, opened 2019-01-07
pull: add all-tags option

q384566678:
Add all-tags option to download all tagged images in the repository

After change:
```
➜  buildah git:(add-pull-all) ✗ ./buildah pull --all-tags busybox
Pulling docker://busybox:1-glibc
Skipping fetch of repeat blob sha256:68d65759a692b254073928cce9b3da459b59ee063f4aeb217cd6bcdfac5f838b
Copying config sha256:755a2703667876f4259f24a3225ef503483953ef553fba8758406beefc2ce3f9
 1.46 KiB / 1.46 KiB [======================================================] 0s
Storing signatures
755a2703667876f4259f24a3225ef503483953ef553fba8758406beefc2ce3f9
Pulling docker://busybox:1-musl                  
Skipping fetch of repeat blob sha256:d900fc804a8829d0ea4db613927f60a28a1ef933aa1dbafdaab43630579646c2
Copying config sha256:3cc47384c4cb779466fe40182420bd90ba761a5f26f8564580a114bcd0dfa911
 1.46 KiB / 1.46 KiB [======================================================] 0s
Storing signatures                     
3cc47384c4cb779466fe40182420bd90ba761a5f26f8564580a114bcd0dfa911
Pulling docker://busybox:1-ubuntu
...
```

### DIFF
--- a/cmd/buildah/pull.go
+++ b/cmd/buildah/pull.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"fmt"
 	"os"
 	"strings"
 
@@ -9,13 +8,16 @@ import (
 	buildahcli "github.com/containers/buildah/pkg/cli"
 	"github.com/containers/buildah/pkg/parse"
 	util "github.com/containers/buildah/util"
-	is "github.com/containers/image/storage"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
 )
 
 var (
 	pullFlags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "all-tags, a",
+			Usage: "download all tagged images in the repository",
+		},
 		cli.StringFlag{
 			Name:  "authfile",
 			Usage: "path of the authentication file. Default is ${XDG_RUNTIME_DIR}/containers/auth.json",
@@ -94,6 +96,9 @@ func pullCmd(c *cli.Context) error {
 	transport := util.DefaultTransport
 	arr := strings.SplitN(args[0], ":", 2)
 	if len(arr) == 2 {
+		if c.Bool("all-tags") {
+			return errors.Errorf("tag can't be used with --all-tags")
+		}
 		if _, ok := util.Transports[arr[0]]; ok {
 			transport = arr[0]
 		}
@@ -105,22 +110,12 @@ func pullCmd(c *cli.Context) error {
 		Store:               store,
 		SystemContext:       systemContext,
 		BlobDirectory:       c.String("blob-cache"),
+		AllTags:             c.Bool("all-tags"),
 	}
 
 	if !c.Bool("quiet") {
 		options.ReportWriter = os.Stderr
 	}
 
-	ref, err := buildah.Pull(getContext(), args[0], options)
-	if err != nil {
-		return err
-	}
-
-	img, err := is.Transport.GetStoreImage(store, ref)
-	if err != nil {
-		return err
-	}
-
-	fmt.Printf("%s\n", img.ID)
-	return nil
+	return buildah.Pull(getContext(), args[0], options)
 }
--- a/docs/buildah-pull.md
+++ b/docs/buildah-pull.md
@@ -42,6 +42,10 @@ The image ID of the image that was pulled.  On error 1 is returned.
 
 ## OPTIONS
 
+**--all-tags, a**
+
+All tagged images in the repository will be pulled.
+
 **--authfile** *path*
 
 Path of the authentication file. Default is ${XDG\_RUNTIME\_DIR}/containers/auth.json, which is set using `podman login`.
--- a/pull.go
+++ b/pull.go
@@ -2,12 +2,14 @@ package buildah
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"strings"
 
 	"github.com/containers/buildah/pkg/blobcache"
 	"github.com/containers/buildah/util"
 	cp "github.com/containers/image/copy"
+	"github.com/containers/image/docker"
 	"github.com/containers/image/docker/reference"
 	tarfile "github.com/containers/image/docker/tarfile"
 	ociarchive "github.com/containers/image/oci/archive"
@@ -17,6 +19,7 @@ import (
 	"github.com/containers/image/transports/alltransports"
 	"github.com/containers/image/types"
 	"github.com/containers/storage"
+	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -45,6 +48,9 @@ type PullOptions struct {
 	// store copies of layer blobs that we pull down, if any.  It should
 	// already exist.
 	BlobDirectory string
+	// AllTags is a boolean value that determines if all tagged images
+	// will be downloaded from the repository. The default is false.
+	AllTags bool
 }
 
 func localImageNameForReference(ctx context.Context, store storage.Store, srcRef types.ImageReference, spec string) (string, error) {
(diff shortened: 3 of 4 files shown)


pull.go
```diff
@@ -141,9 +146,53 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
...
+		}
+		for _, tag := range tags {
+			name := spec + ":" + tag
+			names = append(names, name)
+		}
+	} else {
+		names = append(names, spec)
+	}
+	for _, name := range names {
+		fmt.Printf("Pulling %s\n", name)
```
rhatdan: You need to respect --quiet flag.

pull.go
```diff
@@ -141,9 +146,53 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
...
+		}
+		logrus.Debugf("error parsing image name %q, trying with transport %q: %v", spec, options.Transport, err)
+		transport := options.Transport
+		if transport != util.DefaultTransport {
+			transport = transport + ":"
+		}
+		spec = transport + spec
+		srcRef2, err2 := alltransports.ParseImageName(spec)
+		if err2 != nil {
+			return errors.Wrapf(err2, "error parsing image name %q", spec)
```
rhatdan: Should this still be err wrapped?

pull.go
```diff
@@ -141,9 +146,53 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
...
+		}
+		srcRef = srcRef2
+	}
+	var names []string
+	if options.AllTags {
+		tags, err := docker.GetRepositoryTags(ctx, systemContext, srcRef)
```
rhatdan: Will this work on non docker transports?

pull.go
```diff
@@ -141,9 +146,53 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
...
+		if err != nil {
+			return errors.Wrapf(err, "error geting repository tags")
```
rhatdan: geting -> getting

pull.go
```diff
@@ -141,9 +146,53 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
...
+			names = append(names, name)
+		}
+	} else {
+		names = append(names, spec)
+	}
+	for _, name := range names {
+		fmt.Printf("Pulling %s\n", name)
+		ref, err := pullImage(ctx, options.Store, name, options, systemContext)
+		if err != nil {
+			return errors.Wrapf(err, "error pulling image %q", name)
```
rhatdan: Should we pull all of the images we can and save up the ones that fail?

pull.go
```diff
@@ -141,9 +146,53 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
...
+		}
+		img, err := is.Transport.GetStoreImage(options.Store, ref)
+		if err != nil {
+			return err
+		}
+		fmt.Printf("%s\n", img.ID)
```
rhatdan: Does PullImage do this for us?

cmd/buildah/pull.go
```diff
@@ -94,6 +96,9 @@ func pullCmd(c *cli.Context) error {
 	transport := util.DefaultTransport
 	arr := strings.SplitN(args[0], ":", 2)
 	if len(arr) == 2 {
+		if c.Bool("all-tags") {
+			return errors.Errorf("tag can't be used with --all-tags/-a")
```
TomSweeneyRedHat: drop the '/-a' here.

docs/buildah-pull.md
```diff
@@ -42,6 +42,10 @@ The image ID of the image that was pulled.  On error 1 is returned.
 
 ## OPTIONS
 
+**--all-tags, a**
+
+All tagged images in the repository will be downloaded.
```
TomSweeneyRedHat: I'd suggest "pulled" over "downloaded".  I know the Docker dock uses downloaded, but think it should be pulled there too as we never talk about download prior to this.

pull.go
```diff
@@ -45,6 +47,9 @@ type PullOptions struct {
 	// store copies of layer blobs that we pull down, if any.  It should
 	// already exist.
 	BlobDirectory string
+	// AllTags is a value that determines whether or not download all
```
TomSweeneyRedHat: Suggest:  AllTags is a boolean value that determines if all tagged images will be downloaded from the repository.  The default is false.

pull.go
```diff
@@ -141,9 +146,53 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
 }
 
 // Pull copies the contents of the image from somewhere else to local storage.
-func Pull(ctx context.Context, imageName string, options PullOptions) (types.ImageReference, error) {
+func Pull(ctx context.Context, imageName string, options PullOptions) error {
```
TomSweeneyRedHat: I haven't gone through this throughly, but we need to be careful altering this function as the FROM command also uses it.  Please verify that FROM is behaving as expected with this changes, especially so in the tests.

TomSweeneyRedHat:
Left a few comments, need to test/play with it in the morning.

pull.go
q384566678: I noticed that the other commands eventually call the [`pullImage`](https://github.com/containers/buildah/pull/1263/files#diff-27c9dec751c416fdf10e5e5e780b008cR205) function, so modifying this function will have no effect on other commands.

pull.go
q384566678: No, I only solved this type at the moment.

pull.go
q384566678: Because the previous [cmd/buildah/pull.go](https://github.com/containers/buildah/pull/1263/files#diff-4ca994502fc61e8b4456e9a91bcef926L124) did the same, I just copied this code. I think the output image ID of the pull down is more helpful to the user.

pull.go
q384566678: Not quite sure how to respond.

pull.go
rhatdan: Then we should return unsupported.  What is the error message returned if you use --all-tags on a different type of transport?

pull.go
rhatdan: What does the original err message say. 

If it talks about the unmodified spec, that would be better then the modified spec

IE I specify buildah pull --all-tags foobar

I would want something about unable to parse foobar verses unable to parse docker://foobar.

pull.go
```diff
@@ -141,9 +148,58 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
...
+		for _, tag := range tags {
+			name := spec + ":" + tag
+			names = append(names, name)
+		}
+	} else {
+		names = append(names, spec)
+	}
+	var errs *multierror.Error
+	for _, name := range names {
+		if options.ReportWriter == os.Stderr {
```
rhatdan: Should be != nil

pull.go
```diff
@@ -141,9 +148,58 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
...
+			fmt.Printf("Pulling %s\n", name)
```
rhatdan: options.Reportwriter.WriteString("Pulling %s\n", name)

pull.go
rhatdan: ok

pull.go
q384566678: Did not find the method you said, using other methods to achieve.

rhatdan:
LGTM
@TomSweeneyRedHat @baude @vrothberg @giuseppe @umohnani8 @mheon PTAL

TomSweeneyRedHat:
please don't merge this yet, testing and finding something odd in the pull progress reporting.  IDK if it's due to this change or not, trying to figure it out.

pull.go
TomSweeneyRedHat: Oh you're right @q384566678 , I had it my head that From used this too, but this is used only by Pull, good call.

rhatdan:
@rh-atomic-bot r+

PR #1267, opened 2019-01-08
Fix support for multiple Short options

q384566678:

Adds the UseShortOptionHandling  option to several of the commands.  Allowing for multiple short options to be passed together.  i.e. instead of `buildah pull -c -q alpine` the two options could be combined to `buildah pull -cq alpine`


TomSweeneyRedHat:
I don't think it's needed for the add or copy commands as they have only -q at most for short options, but I don't think it hurts.  I've added a bit of description to your submit, in the future, please add that so a few months later we can figure out what the change was a bit more easily.

LGTM

rhatdan:
@rh-atomic-bot retry

PR #1274, opened 2019-01-09
build-using-dockerfile,commit: disable compression by default

nalind:
When committing an image to local storage, compressing the new layers only wastes time because they need to be decompressed again when being written to local storage.


rhatdan:
LGTM
@TomSweeneyRedHat @umohnani8 @vrothberg @giuseppe PTAL

vrothberg:
Awesome change, @nalind! As Dan mentions above, this change impacts the blob cache tests.

cmd/buildah/bud.go
```diff
@@ -178,9 +178,9 @@ func budCmd(c *cli.Context) error {
 		logrus.Debugf("--compress option specified but is ignored")
 	}
 
-	compression := imagebuildah.Gzip
-	if c.Bool("disable-compression") {
-		compression = imagebuildah.Uncompressed
+	compression := imagebuildah.Uncompressed
+	if c.IsSet("disable-compression") && !c.Bool("disable-compression") {
```
vrothberg: `c.IsSet("disable-compression")` we may need to check for more such cases in our tools. `urfave/cli` is full of surprises.

vrothberg:
LGTM (once tests are happy)

cmd/buildah/bud.go
vrothberg: I'd love to do that. `urfave/cli` is pretty much dead and I think we are stretching it a bit until it hits a bit harder. Maybe something to discuss (and execute) during devconf?

rhatdan: I would be fine with switching, But start with Buildah since it is a much smaller CLI.

vrothberg: I fear we have to start with Podman as Buildah is using some flags from there.

rhatdan:
@rh-atomic-bot r+

vrothberg:
(restarted the timed out test)

tests/commit.bats
```diff
@@ -28,9 +28,9 @@ load helpers
 @test "commit format test" {
   cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
   buildah commit --signature-policy ${TESTSDIR}/policy.json $cid alpine-image-oci
-  buildah commit --format docker --signature-policy ${TESTSDIR}/policy.json $cid alpine-image-docker
+  buildah commit --format docker --disable-compression=false --signature-policy ${TESTSDIR}/policy.json $cid alpine-image-docker
 
-  buildah --debug=false inspect --type=image --format '{{.Manifest}}' alpine-image-oci | grep "application/vnd.oci.image.layer.v1.tar+gzip"
+  buildah --debug=false inspect --type=image --format '{{.Manifest}}' alpine-image-oci | grep "application/vnd.oci.image.layer.v1.tar"
```
vrothberg: @nalind, why did the type change? As we use compression, I'd expect this to still be gzip.

rhatdan:
@rh-atomic-bot r+

tests/commit.bats
nalind: We left compression on when committing using the Docker format, but turned it off for the OCI format.

PR #1287, opened 2019-01-17
imagebuildah: make EnsureContainerPath() check/create the right one

nalind:
Update `EnsureContainerPath()` to check for the existence of the passed-in path relative to the working container, and to create it if it doesn't exist.


rhatdan:
LGTM

PR #1300, opened 2019-01-18
Add options for empty-layer history entries

nalind:
Add configuration methods for adding entries which will show up in a committed image's history, both before and after the new layer that we add while committing the image.  Expose them from the CLI in the form of a new --add-history option for the "add", "config", "copy", and "run" commands.

### DIFF
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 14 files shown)


cmd/buildah/config.go
```diff
@@ -17,6 +18,10 @@ import (
 
 var (
 	configFlags = []cli.Flag{
+		cli.BoolFlag{
+			Name:  "add-history",
```
rhatdan: Lets add an environment variable to turn this on or off by default.

nalind: Added `$BUILDAH_HISTORY` for it.

docs/buildah-add.md
```diff
@@ -15,6 +15,14 @@ archive file itself.  If a local directory is specified as a source, its
...
 ## OPTIONS
 
+**--add-history**
+
+Add an entry to the history which will note the digest of the added content.
+Defaults to false.
+
+Note: You can also override the default value of --add-history by setting the
+BUILDAH\_HISTORY environment variable. `export BUILDAH_HISTORY=true`
+
```
TomSweeneyRedHat: :+1:

rhatdan:
@rh-atomic-bot r+

PR #1302, opened 2019-01-23
Add Quiet to PullOptions and PushOptions

TomSweeneyRedHat:

Add Quiet to the PullOptions and PushOptions structures.
If set to true output will be in 'quiet' mode.  This
will primarily be used by callers such as OpenShift.

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -101,6 +101,10 @@ type PushOptions struct {
 	// regenerate from on-disk layers, substituting them in the list of
 	// blobs to copy whenever possible.
 	BlobDirectory string
+	// Quiet is a boolean value that determines if minimal output to
+	// the user will be displayed, this is best used for logging.
+	// The default is false.
+	Quiet bool
 }
 
 // Commit writes the contents of the container, along with its updated
@@ -228,6 +232,9 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 func Push(ctx context.Context, image string, dest types.ImageReference, options PushOptions) (reference.Canonical, digest.Digest, error) {
 	systemContext := getSystemContext(options.SystemContext, options.SignaturePolicyPath)
 
+	if options.Quiet {
+		options.ReportWriter = nil // Turns off logging output
+	}
 	blocked, err := isReferenceBlocked(dest, systemContext)
 	if err != nil {
 		return nil, "", errors.Wrapf(err, "error checking if pushing to registry for %q is blocked", transports.ImageName(dest))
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
@bparees FYI

bparees:
so for BuildOptions (when invoking BuildDockerfiles) you'd just expect us to set ReportWriter=nil?

bparees:
the push api also needs this.

so push, pull, build.  all need an explicit quiet flag (or you can tell us to just always set the reportwriter to nil)

TomSweeneyRedHat:
@bparees I'll go ahead and add to push and build too.

TomSweeneyRedHat:
@bparees I've added to PushOptions, looks like the option is already in play for BuildOptions.  @nalind PTAL

bparees:
thanks

i'm guessing this won't be the end of this... if you look at our 3.x docker-based experience, we had decent logs that showed we were pushing layers so the user had some feedback, but not as verbose as what you get from buildah.  We are probably going to have to find a compromise position in between "no buildah output" and "the current semi verbose buildah output" in the future.  But for now at least this gives us a clean way to enable/disable.

rhatdan:
@rh-atomic-bot r+

bparees:
@TomSweeneyRedHat even w/ quiet output i'm still seeing this output from a build?

```
Copying blob sha256:f972d139738dfcd1519fd2461815651336ee25a8b54c358834c50af094bb262f
Copying blob sha256:5e884d4e3b68029324ec90f6ccf274a9f5b3d954d3715c9fc78f4cc231fe4351
Copying blob sha256:6593d9ef5bb787d7fc78b91cf449305fb29db474b46fd3a14f6b7d5d27312fa7
Copying blob sha256:79c4057dbfa8001aa0ae8f8d8f4abb0f81c692267ab9dc59af0339d023afa4f2
Copying blob sha256:b737e1b974e2ff5c702c643e3c46f328646e83bdb5f37ee5342f854dfc921272
Copying blob sha256:b1d8074a2218384d6974a2c8d4a57c3d541b550062c443650dcde68c5b228cf6
Copying blob sha256:493c57fb457fac16d89437214b1ed226296394575b5dcf0e2aa62772b529b4ab
Copying blob sha256:c4e91bcbbac6b6d2f5dbf95f1451ee9b1f59a138a0f6a4a9d555ed7d68a460f1
Copying blob sha256:36d0ae34ce990249fba5d102fc88bd9de7a54897f2b2475f1c108a2ee4c4a752
Copying config sha256:1726612d31f7126e74cb15c986b971052ac5093b436b48cfa21cd3888efcead2
Storing signatures
```

rhatdan:
@vrothberg Looks like a bug in containers/image?

PR #1306, opened 2019-01-29
Note configuration file requirements

TomSweeneyRedHat:

We've had a couple of comments that the required configuration
files are not as documented as they should be.  Adding a note to
the install and troubleshooting guides to hopefully help.

### DIFF
--- a/install.md
+++ b/install.md
@@ -204,6 +204,9 @@ The build steps on Debian are otherwise the same as Ubuntu, above.
 
 ## Configuration files
 
+The following configuration files are required in order for Buildah to run appropriately.  The
+majority of these files are commonly contained in the `containers-common` package.
+
 ### [registries.conf](https://src.fedoraproject.org/rpms/skopeo/blob/master/f/registries.conf)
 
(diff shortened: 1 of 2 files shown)


install.md
```diff
@@ -204,6 +204,10 @@ The build steps on Debian are otherwise the same as Ubuntu, above.
 
 ## Configuration files
 
+The following configuration files are required in order for Buildah to run appropriately.  The
+majority of these files are contained in the `containers-common` package, please verify that
```
rhatdan: Can you state that this is commonly packaged in containers-common.  No need to force all distros to package that way.

TomSweeneyRedHat: hmm, didn't get e-mail on this.  I've add commonly and have removed ', please verify...' as it tells them to check for that kit.

TomSweeneyRedHat:
pushed latest with requested change.

TomSweeneyRedHat:
@rhatdan, but still doesn't like me, can you retest this please?

rhatdan:
@rh-atomic-bot r+

Issue #1307, opened 2019-01-29
RFE - Report that configuration files are missing


TomSweeneyRedHat:
**Description**

I'm just making an issue so it doesn't get lost in the shuffle.  In #810 @rhatdan noted that we should at least throw a warning if the required [configuration files](https://github.com/containers/buildah/blob/master/install.md#configuration-files) such as registry.conf, storage.conf, policy.json, mount.conf, etc are not found when running a Buildah command.

This is probably a good one for someone getting their feet wet in Buildah to take on.

rhatdan:
I agree. lets setup a library in buildah that checks this.
pkg/utils/
or something like that which validates all of the tools required for setup exists.

rhatdan:
Why do we need this if the errors look pretty good and the tools actually work in some cases without the files.

Lets examine the errors.
# No registries.conf
```
 buildah from fedora
(image name "fedora" is a short name and no search registries are defined in /etc/containers/registries.conf): while pulling "fedora" as "localhost/fedora": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
```

# No Policy.json
```
buildah from fedora:f28
The following failures happened while trying to pull image specified by "fedora:f28" based on search registries in /etc/containers/registries.conf:
* "localhost/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
* "docker.io/library/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
* "registry.fedoraproject.org/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
* "quay.io/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
* "registry.access.redhat.com/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
* "registry.centos.org/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
```

# No /etc/containers/storage.conf

```
# buildah from fedora:28
Getting image source signatures
Copying blob 012cb44e3253: 85.49 MiB / 85.49 MiB [=========================] 26s
Copying config 4c5b21baa7a0: 1.99 KiB / 1.99 KiB [==========================] 0s
Writing manifest to image destination
Storing signatures
fedora-working-container-1
```

QiWang19:
I agree. I don't think this is necessary

[...]

Issue #575, opened 2018-04-12
Error target is unknown on docker pull for image pushed with buildah


gregoryboue:
**Description**

I build an image with *buildah bud* then i push it to a private registry with a *buildah push* command.

The image is visible on registry GUI, and i can list the tag with the endpoint : https://myregistry/v2/MY_IMAGE_NAME/tags/list.

When i try a *docker pull* command i have this message : ` Error response from daemon: target is unknown`

**Note** : It's possible to pull the image with *buildah from* but not with *docker pull*

**Steps to reproduce the issue:**
1. Build an image with *buildah bud*...
2. Push an image to a private registry (nexus3) with `buildah push --creds=${CRED} MY_IMAGE_ID docker://MY_IMAGE_NAME_AND_TAG`
3. Docker login and Pull an image from Docker version 17.03.1-ce, build c6d412e daemon : `docker pull MY_IMAGE_TAG`

**Describe the results you received:**

Error from daemon docker logs :
```
level=debug msg="Calling POST /v1.27/images/create?fromImage=MY_IMAGE_NAME_AND_TAG"
level=debug msg="hostDir: /etc/docker/certs.d/MYREGISTRY"
level=debug msg="hostDir: /etc/docker/certs.d/MYREGISTRY"
level=debug msg="Trying to pull MY_IMAGE_NAME_AND_TAG from https://MYREGISTRY v2"
level=debug msg="Calling GET /v1.30/services"
level=debug msg="Calling GET /v1.30/tasks"
level=error msg="Error trying v2 registry: target is unknown"
level=error msg="Attempting next endpoint for pull after error: target is unknown"
level=debug msg="Skipping v1 endpoint https://MYREGISTRY because v2 registry was detected"
 level=error msg="Handler for POST /v1.27/images/create returned error: target is unknown"
```

**Describe the results you expected:**
Get the image pulled.

**Output of `buildah version`:**

```
Version:       0.16
Go Version:    go1.8.3
Image Spec:    1.0.0
Runtime Spec:  1.0.0
Git Commit:    82529cd
Built:         Fri Apr  6 10:15:58 2018
OS/Arch:       linux/amd64
```

**Output of `cat /etc/*release`:**

```
CentOS Linux release 7.3.1611 (Core) 
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
[...]

PR #1312, opened 2019-01-31
Check if config files exist before buildah command

QiWang19:

### DIFF
--- a/cmd/buildah/main.go
+++ b/cmd/buildah/main.go
@@ -2,9 +2,10 @@ package main
 
 import (
 	"fmt"
-	"github.com/containers/buildah"
 	"os"
 
+	"github.com/containers/buildah"
+	"github.com/containers/buildah/pkg/utils"
 	"github.com/containers/libpod/pkg/util"
 	"github.com/containers/storage"
 	ispecs "github.com/opencontainers/image-spec/specs-go"
@@ -56,6 +57,12 @@ func init() {
 		logrus.Errorf(err.Error())
 		os.Exit(1)
 
+	}
+
+	err = utils.CheckConfigFiles()
+	if err != nil {
+		logrus.Errorf(err.Error())
+		os.Exit(1)
 	}
 
 	if len(storage.DefaultStoreOptions.GraphDriverOptions) > 0 {
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -1,0 +1,39 @@
+package utils
+
+import (
+	"os"
+
+	"github.com/containers/image/pkg/sysregistries"
+	"github.com/containers/image/signature"
+	"github.com/containers/libpod/pkg/rootless"
+	"github.com/containers/storage"
+	"github.com/pkg/errors"
+)
+
+// CheckConfigFiles log messages if one or more files to set up buildah do not exist
+func CheckConfigFiles() error {
+	registriesConfPath := sysregistries.RegistriesConfPath(nil)
+	_, err := os.Stat(registriesConfPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return errors.Errorf("required file %q not found.", registriesConfPath)
+		}
+		return err
+	}
+
+	if !rootless.IsRootless() {
+		storageDefaultPath := storage.DefaultConfigFile
+		_, err = os.Stat(storageDefaultPath)
+		if err != nil {
+			if os.IsNotExist(err) {
+				return errors.Errorf("required file %q not found.", storageDefaultPath)
+			}
+			return err
+		}
+	}
+
+	if _, err = signature.DefaultPolicy(nil); err != nil {
+		return err
+	}
+	return nil
+}


pkg/utils/utils.go
```diff
@@ -0,0 +1,55 @@
...
+			logrus.Errorf("error checking seccomp.json in %s and %s, %s, %s", seccompDefaultPath, seccompOverridePath, errDefault, errOptional)
+		}
+	}
+
+	_, err = signature.DefaultPolicy(nil)
+	if err != nil {
+		if os.IsNotExist(err) {
+			logrus.Errorf("Required file policy.json not found %s", err)
+		} else {
+			logrus.Errorf("error checking policy.json, %s", err)
```
rhatdan: You need to print out the path here.

/etc/containers/policy.json Which I think signature.DefaultPolicy will return?

pkg/utils/utils.go
```diff
@@ -0,0 +1,55 @@
...
+		}
+	}
+
+	seccompDefaultPath := "/usr/share/containers/seccomp.json"
+	seccompOverridePath := "/etc/crio/seccomp.json"
+	_, errDefault = os.Stat(seccompDefaultPath)
+	_, errOptional = os.Stat(seccompOverridePath)
+	if errDefault != nil && errOptional != nil {
+		if os.IsNotExist(errDefault) && os.IsNotExist(errOptional) {
+			logrus.Errorf("Required file seccomp.json not found in %s or %s, please ensure that the containers-common package has been installed.", seccompDefaultPath, seccompOverridePath)
```
rhatdan: We need a path, not just the file.  BTW The user can override this file, so we need to check the default path or the override path.  Not sure if this is required at all, since I think the tool should fail if this file does not exist.

pkg/utils/utils.go
```diff
@@ -0,0 +1,55 @@
...
+	_, err := os.Stat(registriesConfPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			logrus.Errorf("Required file registries.conf not found in %s, %s", registriesConfPath, err)
+		} else {
+			logrus.Errorf("error checking registries.conf in %s, %s", registriesConfPath, err)
+		}
+	}
+
+	mountsDefaultFilePath := "/usr/share/containers/mounts.conf"
```
rhatdan: These	

DefaultMountsFile = "/usr/share/containers/mounts.conf"
OverrideMountsFile = "/etc/containers/mounts.conf"
UserOverrideMountsFile = filepath.Join(os.Getenv("HOME"), ".config/containers/mounts.conf")

 are defined in 
libpod/pkg/secrets/secrets.go

The constants should be used from there.  Also we need to handle the rootless case.  Do things fail if this file does not exists?  If not I say we just ignore it.

pkg/utils/utils.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// CheckConfigFiles log messages if one or more files to set up buildah do not exist
+func CheckConfigFiles() {
+	registriesConfPath := sysregistries.RegistriesConfPath(nil)
+	_, err := os.Stat(registriesConfPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			logrus.Errorf("Required file registries.conf not found in %s, %s", registriesConfPath, err)
```
rhatdan: logrus.Errorf("Required file %s not found", registriesConfPath)

pkg/utils/utils.go
```diff
@@ -0,0 +1,55 @@
...
+		} else {
+			logrus.Errorf("error checking registries.conf in %s, %s", registriesConfPath, err)
```
rhatdan: logrus.Errorf("error checking %s, %s",  registriesConfPath, err)

pkg/utils/utils.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"os"
+
+	"github.com/containers/image/pkg/sysregistries"
+	"github.com/containers/image/signature"
+	"github.com/sirupsen/logrus"
+)
+
+// CheckConfigFiles log messages if one or more files to set up buildah do not exist
+func CheckConfigFiles() {
```
rhatdan: Should this return an error, if anyone are missing?

pkg/utils/utils.go
rhatdan: Better yet, why not put this function into secrets.go and do the check there.  I would also like to see secrets.go moved to buildah so we could stop vendoring libpod into buildah.

QiWang19: I think those commands do not fail if there's no mounts file, I'll ignore it...

cmd/buildah/pull.go
```diff
@@ -87,6 +88,8 @@ func pullCmd(c *cli.Context) error {
 		return errors.Wrapf(err, "error building system context")
 	}
 
+	utils.CheckConfigFiles()
```
TomSweeneyRedHat: I was thinking this call would live in cmd/buildah/main.go up near the top of the main() function.  I just checked and had forgotten that we've a check there for Storage Options already, so probably makes sense to add this type of check there too rather than in each command.

pkg/utils/utils.go
TomSweeneyRedHat: yes please

pkg/utils/utils.go
QiWang19: not all commands fail. `buildah run` fails, but `buildah from` or `buildah pull` can work
the override path can be checked inside the function for the command. 
only need to check the files required for all command?

rhatdan: Lets talk about buildah run failing when their is no seccomp file.  I think we could just warn.

cmd/buildah/pull.go
rhatdan: Agree, it should be in Main.

QiWang19: cmd/buildah/main.go I can only check using the default path?
I'm thinking if the path can override by the user, the method for the command will check it and return errors so we only need to check the default path of the config file required for all commands and maybe that's an error.
but currently, I'm not sure which is must be required

QiWang19:

TomSweeneyRedHat:
Ugh, looks like the test system is failing on every test due to your change.  It's not able to find some of the files in the test system such as policy.json.  We pass that in from the command line.  Not sure off the top of my head how to get around that.

cmd/buildah/main.go
```diff
@@ -26,6 +27,12 @@ func main() {
 		cli.OsExiter(1)
 	}
 
+	err = utils.CheckConfigFiles()
+	if err != nil {
+		logrus.Errorf(err.Error())
+		cli.OsExiter(1)
```
TomSweeneyRedHat: Just as a test, could you comment out this line temporarily and push a new patch?  I'd like to see if the tests are happy that way.  Then we can figure out if we should change this to a warning or figure out a way to turn off this check during testing.

pkg/utils/utils.go
```diff
@@ -0,0 +1,58 @@
...
+	_, errOverride := os.Stat(seccompOverridePath)
+	if errDefault != nil && errOverride != nil {
+		if os.IsNotExist(errDefault) && os.IsNotExist(errOverride) {
+			errMsg = fmt.Sprintf("Required file %s or %s not found.\n %s", seccompDefaultPath, seccompOverridePath, errMsg)
+		} else {
+			errMsg = fmt.Sprintf("error checking %s and %s, %s, %s.\n %s", seccompDefaultPath, seccompOverridePath, errDefault, errOverride, errMsg)
+		}
+	}
+
+	policyDefaultPath := "/etc/containers/policy.json"
```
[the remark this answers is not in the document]
QiWang19: `func defaultPolicyPath(sys *types.SystemContext) string` containers/image/signature/policy_config.go in `containers/image/signature/policy_config.go` is not public, and no sys *types.SystemContext can be used if call this function in main.go, so I think no?

TomSweeneyRedHat:
@QiWang19 I'd like @rhatdan to weigh in on just throwing the error but not exiting there.  But for a quick test, I think it would be OK to do just to see if it gets past the Production Travis CI.

pkg/utils/utils.go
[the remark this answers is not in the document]
TomSweeneyRedHat: @rhatdan thoughts?  Shudder, should we put together a conf file for buildah or put a requirement in on Podman's conf files?  Double shudder.

cmd/buildah/main.go
QiWang19: ok let's see if can pass :eyes:

pkg/utils/utils.go
```diff
@@ -0,0 +1,58 @@
...
+// CheckConfigFiles log messages if one or more files to set up buildah do not exist
+func CheckConfigFiles() error {
+	errMsg := ""
+	registriesConfPath := sysregistries.RegistriesConfPath(nil)
+	_, err := os.Stat(registriesConfPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			errMsg = fmt.Sprintf("Required file %s not found.\n", registriesConfPath)
+		} else {
+			errMsg = fmt.Sprintf("error checking %s, %s", registriesConfPath, err)
```
mtrmac: The code in **EDIT<s>`registries.conf`</s>`c/image/pkg/sysregistriesv2` explicitly supports the situation when the file is missing. It’s not at all obvious to me that buildah should enforce its existence independently — if it is _that_ necessary, `c/image/pkg/sysregistriesv2` should fail if it is missing as well.

To the extent this is about https://github.com/containers/buildah/issues/1307 , buildah already has the `searchRegistriesWereUsedButEmpty` boolean in the search path, so isn’t this just a matter of propagating/displaying that error correctly, or maybe using the same boolean on other code paths (or consolidating them)?

pkg/utils/utils.go
```diff
@@ -0,0 +1,58 @@
...
+
+	storageDefaultPath := "/etc/containers/storage.conf"
+	_, err = os.Stat(storageDefaultPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			errMsg = fmt.Sprintf("Required file %s not found.\n %s", storageDefaultPath, errMsg)
+		} else {
+			errMsg = fmt.Sprintf("error checking %s, %s.\n %s", storageDefaultPath, err, errMsg)
+		}
+	}
```
mtrmac: The code in `c/storage.ReloadConfigurationFile` explicitly supports the situation when the file is missing. Is Buildah doing something special that makes it different from other users that it requires a config file? (E.g. if it requires some non-default configuration option, that option should either become a default, or this code should check that that option is set as necessary.)

pkg/utils/utils.go
```diff
@@ -0,0 +1,58 @@
...
+
+	policyDefaultPath := "/etc/containers/policy.json"
+	_, err = os.Stat(policyDefaultPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			errMsg = fmt.Sprintf("Required file %s not found.\n %s", policyDefaultPath, errMsg)
+		} else {
+			errMsg = fmt.Sprintf("error checking %s, %s.\n %s", policyDefaultPath, err, errMsg)
+		}
+	}
```
mtrmac: Why is existence of this file necessary before processing every single command, including things like `buildah version`?

Code paths that do need this file will eventually fail, so it seems to me that just making sure the error reporting/propagation works correctly should give the users all necessary information, e.g.
```console
$ ./buildah pull --signature-policy /foo/bar foo
error obtaining default signature policy: open /foo/bar: no such file or directory
```
That’s _already_ more useful than the text added above, which says “required file” and does not even say what it is for.  (Sure, the text added above can be easily improved, while “no such file or directory” might not quite as easy to augment.)

pkg/utils/utils.go
rhatdan: @TomSweeneyRedHat 

The question is how user friendly do we want to make the tool. Can the tools diagnose the situation, where the user is likely to get a bad experience.  People are packaging up buildah either from source or for other distributions and are suprised when they don't work correctly.  Then they have to find random config files to make them run correctly.

So the goal was to make this up front to the user.

Lets examine the errors.
# No registries.conf
```
 buildah from fedora
(image name "fedora" is a short name and no search registries are defined in /etc/containers/registries.conf): while pulling "fedora" as "localhost/fedora": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
```

# No Policy.json
```
buildah from fedora:f28
The following failures happened while trying to pull image specified by "fedora:f28" based on search registries in /etc/containers/registries.conf:
* "localhost/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
* "docker.io/library/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
* "registry.fedoraproject.org/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
* "quay.io/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
* "registry.access.redhat.com/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
* "registry.centos.org/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
```

# No /etc/containers/storage.conf

```
# buildah from fedora:28
Copying blob 012cb44e3253: 85.49 MiB / 85.49 MiB [=========================] 26s
Copying config 4c5b21baa7a0: 1.99 KiB / 1.99 KiB [==========================] 0s
Storing signatures
fedora-working-container-1
```

rhatdan:
I am leading toward @mtrmac I don't think this is necessary.

TomSweeneyRedHat:
I noted in #1307 that I think at least a warning message for registries.conf and policy.json non-existence would be good here.   I think the message that you get now are either not terribly clear (can't find the registry in the file when the file doesn't exist), or verbose and have to be waded through.  I think a warning for any command would help a user when things are out of whack and in reality rarely be issued.  

My $.02, I'm good with going with the general consensus.

pkg/utils/utils.go
mtrmac: > # No Policy.json
> ```
> buildah from fedora:f28
> The following failures happened while trying to pull image specified by "fedora:f28" based on search registries in /etc/containers/registries.conf:
> * "localhost/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
> * "docker.io/library/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
> * "registry.fedoraproject.org/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
> * "quay.io/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
> * "registry.access.redhat.com/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
> * "registry.centos.org/fedora:f28": error obtaining default signature policy: open /etc/containers/policy.json: no such file or directory
> ```

BTW this could be easily improved (and made a bit more efficient, avoiding repeated re-parsings of the file, and in the future perhaps redundant up GPG home directory creations) by moving `signature.NewPolicyContext` from `pullImage` out of the loop to `resolveImage`

mtrmac:
Oh, the situation can _definitely_ be improved[1] - it’s just that this PR has the combination of _both_ being imprecise about the problem (only checking file presence vs. expected properties) _and_ being imprecise about the context (checking even for commands that don’t care at all). Fixing this to be precise about at least one of the two would make it much more acceptable to me.

I would be perfectly fine with adding this information to the output of `buildah info`, or maybe even `buildah version`; but making `buildah version`, and even `buildah --help`, report errors before doing anything else, is a bit much I think.

---

[1] Arguably the existence of `errors.Wrapf` leads to implementation-focused instead of end-user-focused errors, and is the underlying problem… but we’ve been very much benefiting from how much easier it makes to report …_not useless_ errors, so ripping that out in favor of thoughtful errors throughout is not likely to happen.

QiWang19:
improve a little bit, check file presence for storage.conf and registries.conf, check expected properties for policy.json by calling signature.DefaultPolicy() (get rid of the vendor PR [#575](https://github.com/containers/image/pull/575)
But directly exit in main.go.

Issue #1046, opened 2018-09-28
Fail to build Dockerfile: unpacking of archive failed on file /etc/hosts: cpio: rename


patrickdillon:
**Description**

I am trying to build a Dockerfile. It builds ok with Docker but fails with buildah.

**Steps to reproduce the issue:**
1. Dockerfile: http://pastebin.test.redhat.com/652536
2. `podman build -t install-test .`

**Describe the results you received:**
Error logs: http://pastebin.test.redhat.com/652525

**Describe the results you expected:**
Expected it to build (works with Docker).

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.3-1.git4888163.fc28.x86_64
```

**Output of `buildah version`:**

```
Version:         1.3
Go Version:      go1.10.3
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.3.1
libcni Version:  v0.6.0
Git Commit:      4888163
Built:           Sun Aug  5 07:31:55 2018
OS/Arch:         linux/amd64
```

**Output of `cat /etc/*release`:**

```
Fedora release 28 (Twenty Eight)
NAME=Fedora
VERSION="28 (Workstation Edition)"
ID=fedora
VERSION_ID=28
PLATFORM_ID="platform:f28"
PRETTY_NAME="Fedora 28 (Workstation Edition)"
ANSI_COLOR="0;34"
CPE_NAME="cpe:/o:fedoraproject:fedora:28"
HOME_URL="https://fedoraproject.org/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=28
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=28
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Workstation Edition"
VARIANT_ID=workstation
Fedora release 28 (Twenty Eight)
Fedora release 28 (Twenty Eight)
```

**Output of `uname -a`:**

```
[...]

PR #1313, opened 2019-01-31
Prepend a comment to files setup in container

rhatdan:
setup.rpm attempts to modify /etc/hosts, if it thinks
it has never been modified.  By adding a #comment to the
front of the file, it should prevent this from blowing up
the build.


rhatdan:
Attempt to fix https://github.com/containers/buildah/issues/1046

@nalind WDYT?

TomSweeneyRedHat:
@mglantz it would be great if you could try this in your environment to see if it solves your issue.

mglantz:
I can get testing right away @TomSweeneyRedHat, do you have automatic builds including RPM packages or do I need to built it myself?

TomSweeneyRedHat:
@mglantz thanks, no auto builds at this time, you'll need to pull the patch and build/install.

mglantz:
I'm testing, I'll should be done latest tomorrow.

mglantz:
I do not get this to work on a completely clean install of F29 and a build of this fix, same problem as before. I used the same Dockerfile as I used before to test.

mglantz:
I'll try and test on a new EL8 test rig as well later.

TomSweeneyRedHat:
@mairin drats!  But thanks for the test run and I'd be interested in hearing the EL8 test results.

TomSweeneyRedHat:
I'm not at all sure how I got to @mairin on my last reply, but that reply was directed towards @mglantz

mglantz:
I'm waiting for people to sort out availability of -devel packages for EL8 so I can compile the fix. Hopefully I can test today.

rhatdan:
@mglantz Are you saying this fixes the issue or not?

rhatdan:
@mglantz Could you try this one.

rhatdan:
@TomSweeneyRedHat @nalind @vrothberg @giuseppe PTAL

mglantz:
> @mglantz Could you try this one.

I tried it out on a fresh F29 install and I see the same problem I'm afraid. Just in case I'm messing this somehow, full test is as such:
```
cd
cat << 'EOF' >Dockerfile
FROM registry.centos.org/che-stacks/centos-stack-base
RUN sudo yum -y update setup
CMD tail -f /dev/null 
EOF
mkdir buildah
cd buildah
export GOPATH=`pwd`
git clone -b copy1 https://github.com/rhatdan/buildah ./src/github.com/containers/buildah
cd src/github.com/containers/buildah
make
sudo make install
cd
podman build -t buildah:test .
```

rhatdan:
@mglantz Could you run a container and cat /etc/hosts file, to make sure my fix took place.

rhatdan:
Actually you need to do this with buildah bud not with podman.  If we fix it here then we need to vendor this into podman.

run.go
```diff
@@ -638,6 +607,58 @@ func (b *Builder) addNetworkConfig(rdir, hostPath string, chownOpts *idtools.IDP
...
+		if len(values) != 2 {
+			return "", errors.Errorf("unable to parse host entry %q: incorrect format", host)
+		}
+		if values[0] == "" {
+			return "", errors.Errorf("hostname in host entry %q is empty", host)
+		}
+		if values[1] == "" {
+			return "", errors.Errorf("IP address in host entry %q is empty", host)
+		}
+		hosts += fmt.Sprintf("%s\t%s\n", values[1], values[0])
```
vrothberg: Performance nit: string concatenations are very slow. Starting with Go 1.10, there's a `strings.Builder` which is much faster. We could also use a `var buf bytes.Buffer` and `buf.WriteString("...")` which is still magnitudes faster than the `+` operator.

run.go
```diff
@@ -638,6 +607,58 @@ func (b *Builder) addNetworkConfig(rdir, hostPath string, chownOpts *idtools.IDP
...
+	}
+
+	if hostname != "" {
+		hosts += fmt.Sprintf("127.0.0.1   %s\n", hostname)
+		hosts += fmt.Sprintf("::1         %s\n", hostname)
+	}
+	cfile := filepath.Join(rdir, filepath.Base(hostPath))
+	if err = ioutils.AtomicWriteFile(cfile, []byte(hosts), stat.Mode().Perm()); err != nil {
```
vrothberg: I think `bytes.Buffer` is the right type for `hosts`.

run.go
```diff
@@ -638,6 +607,58 @@ func (b *Builder) addNetworkConfig(rdir, hostPath string, chownOpts *idtools.IDP
 	return cfile, nil
 }
 
+// generateHosts creates a containers hosts file
+func (b *Builder) generateHosts(rdir, hostname string, addHosts []string, chownOpts *idtools.IDPair) (string, error) {
+	hostPath := "/etc/hosts"
+	stat, err := os.Stat(hostPath)
```
vrothberg: Can we drop a comment on why the `Stat()` call is necessary? It took me a while to find the user below to preserve the permissions.

mglantz:
> Actually you need to do this with buildah bud not with podman. If we fix it here then we need to vendor this into podman.

@rhatdan Aah. I thought podman called on buildah. Using buildah directly:

```
buildah from registry.centos.org-che-stacks/centos-stack-base
buildah run centos-stack-base-working-container-3 -- sudo yum -y update setup
```
Things works without any problem.

rhatdan:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1669197

run.go
TomSweeneyRedHat: Added #1397 to go through and scrub these.  I'm going to let this for now.

run.go
TomSweeneyRedHat: @rhatdan thoughts on @vrothberg 's comment here?

vrothberg: @TomSweeneyRedHat, we can defer that to another PR. At some point, I would love to analyze the code for such and similar issues in the hope to further squeeze some cycles :)

rhatdan: Moved to bytes.Buffer,  Is there something better then using the fmt.Sprintf?

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
@vrothberg one last check please and then merge?

vrothberg:
@rh-atomic-bot r+
LGTM

Issue #1316, opened 2019-01-31
Respect `registries.conf` search list for image pulls


adambkaplan:
`buildah pull` currently does not consult the `registries.conf` search list when doing an image pull. It instead uses "docker-centric" logic to pull from `docker.io` if no repository is specified in an image pull spec.

This can cause builds to break if `docker.io` is not in the `registries.conf` search path. See report in https://github.com/containers/image/issues/572.

rhatdan:
Tom can you check this and make it work the same way podman pull works?

@vrothberg @mtrmac Is this something that is being moved into container/image?

mtrmac:
c/image won’t natively implement the signature-breaking search in the `docker://` transport if I can help it.  We have been speculatively talking about sharing the buildah and Podman implementations, maybe placing that shared code in some explicitly-opt-in subpackage of c/image (better to have one problematic code than two), but for now the two are different (and a bit diverging over time).

FWIW I’ve managed to convince OpenShift to give up on search ( https://github.com/openshift/builder/pull/40 https://github.com/openshift/api/pull/192 ) as it has never worked as expected with authentication, so I’m not sure how urgent this is for buildah.

Still, it’s very reasonable to expect that `buildah from`, `FROM` in `buildah bud`, and `buildah pull` all behave the same.

mtrmac:
> Thanks so much 🙏 I would very much like to deprecate the search registries as they actually serve as mirrorr/pull endpoints for docker.io.

That’s not quite how we think of them; projectatomic/docker has diverged from upstream (e.g. in https://github.com/projectatomic/docker/commit/8193836927adc3cf0f9f84be43680a832855c2ae ) to define (hostname-)“qualified”/“unqualified” names; unqualified mean “use the search list”, but that’s not the same thing as treating the search list as mirrors — notably
1. the pulled image is tagged with the hostname found during the search, not as `docker.io`
2. the pull uses credentials for that hostname, not for `docker.io`.

> Expanding an unqualified image (e.g., `alpine:latest` to `docker.io/library/alpine:latest`) is an implementation detail of Docker but not defined in any spec.

Right, there is a spec for how to pull from a registry, but not for what the docker CLI arguments mean.  The `--add-registry` patch has redefined that… and succeeded enough that we probably can’t just drop the shorthand search notation, I’m afraid.

Ideally, I’d like to have aliases instead of search (`pull $fedora`, with a config file defining `$fedora = registry.fedoraproject.org/…`).

[...]

PR #1319, opened 2019-02-02
 Fix pulling of images within buildah 

rhatdan:
Change references to Transfer to transfer to make it internal only.
It should be determined from the image specification and only determined
in one place.

Make buildah.Pull use registries.conf

Currently buildah pull does not resolve images based on registries.conf
This does not match the behaviour of buildah from or buildah bud

### DIFF
--- a/buildah.go
+++ b/buildah.go
@@ -336,11 +336,6 @@ type BuilderOptions struct {
 	// needs to be pulled and the image name alone can not be resolved to a
 	// reference to a source image.  No separator is implicitly added.
 	Registry string
-	// Transport is a value which is prepended to the image's name, if it
-	// needs to be pulled and the image name alone, or the image name and
-	// the registry together, can not be resolved to a reference to a
-	// source image.  No separator is implicitly added.
-	Transport string
 	// PullBlobDirectory is the name of a directory in which we'll attempt
 	// to store copies of layer blobs that we pull down, if any.  It should
 	// already exist.
--- a/cmd/buildah/pull.go
+++ b/cmd/buildah/pull.go
@@ -86,7 +86,10 @@ func pullCmd(c *cobra.Command, args []string, iopts pullResults) error {
 		BlobDirectory:       iopts.blobCache,
 		AllTags:             iopts.allTags,
 		ReportWriter:        os.Stderr,
-		Quiet:               iopts.quiet,
+	}
+
+	if iopts.quiet {
+		options.ReportWriter = nil // Turns off logging output
 	}
 
 	return buildah.Pull(getContext(), args[0], options)
--- a/cmd/buildah/pull.go
+++ b/cmd/buildah/pull.go
@@ -2,12 +2,10 @@ package main
 
 import (
 	"os"
-	"strings"
 
 	"github.com/containers/buildah"
 	buildahcli "github.com/containers/buildah/pkg/cli"
 	"github.com/containers/buildah/pkg/parse"
-	"github.com/containers/buildah/util"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
@@ -81,27 +79,14 @@ func pullCmd(c *cobra.Command, args []string, iopts pullResults) error {
 		return err
 	}
 
-	transport := util.DefaultTransport
-	arr := strings.SplitN(args[0], ":", 2)
-	if len(arr) == 2 {
-		if iopts.allTags {
-			return errors.Errorf("tag can't be used with --all-tags")
-		}
-		if _, ok := util.Transports[arr[0]]; ok {
-			transport = arr[0]
-		}
-	}
-
 	options := buildah.PullOptions{
-		Transport:           transport,
 		SignaturePolicyPath: iopts.signaturePolicy,
 		Store:               store,
 		SystemContext:       systemContext,
 		BlobDirectory:       iopts.blobCache,
 		AllTags:             iopts.allTags,
-	}
-	if !iopts.quiet {
-		options.ReportWriter = os.Stderr
+		ReportWriter:        os.Stderr,
+		Quiet:               iopts.quiet,
 	}
 
 	return buildah.Pull(getContext(), args[0], options)
--- a/docs/buildah-pull.md
+++ b/docs/buildah-pull.md
@@ -25,6 +25,9 @@ Multiple transports are supported:
   **docker-daemon:**_docker-reference_
   An image _docker-reference_ stored in the docker daemon's internal storage.  _docker-reference_ must include either a tag or a digest.  Alternatively, when reading images, the format can also be docker-daemon:algo:digest (an image ID).
 
+  **oci:**_path_**:**_tag_**
+  An image tag in a directory compliant with "Open Container Image Layout Specification" at _path_.
+
   **oci-archive:**_path_**:**_tag_
   An image _tag_ in a directory compliant with "Open Container Image Layout Specification" at _path_.
 
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -62,11 +62,6 @@ type BuildOptions struct {
 	// needs to be pulled and the image name alone can not be resolved to a
 	// reference to a source image.  No separator is implicitly added.
 	Registry string
-	// Transport is a value which is prepended to the image's name, if it
-	// needs to be pulled and the image name alone, or the image name and
-	// the registry together, can not be resolved to a reference to a
-	// source image.  No separator is implicitly added.
-	Transport string
 	// IgnoreUnrecognizedInstructions tells us to just log instructions we
 	// don't recognize, and try to keep going.
 	IgnoreUnrecognizedInstructions bool
@@ -186,7 +181,6 @@ type Executor struct {
 	builder                        *buildah.Builder
 	pullPolicy                     buildah.PullPolicy
 	registry                       string
-	transport                      string
 	ignoreUnrecognizedInstructions bool
 	quiet                          bool
 	runtime                        string
@@ -582,7 +576,6 @@ func NewExecutor(store storage.Store, options BuildOptions) (*Executor, error) {
 		contextDir:                     options.ContextDirectory,
 		pullPolicy:                     options.PullPolicy,
 		registry:                       options.Registry,
-		transport:                      options.Transport,
 		ignoreUnrecognizedInstructions: options.IgnoreUnrecognizedInstructions,
 		quiet:                          options.Quiet,
 		runtime:                        options.Runtime,
@@ -672,7 +665,6 @@ func (b *Executor) Prepare(ctx context.Context, stage imagebuilder.Stage, from s
 		FromImage:             from,
 		PullPolicy:            b.pullPolicy,
 		Registry:              b.registry,
-		Transport:             b.transport,
 		PullBlobDirectory:     b.blobDirectory,
 		SignaturePolicyPath:   b.signaturePolicyPath,
 		ReportWriter:          b.reportWriter,
@@ -785,7 +777,7 @@ func (b *Executor) resolveNameToImageRef() (types.ImageReference, error) {
 	if b.output != "" {
 		imageRef, err = alltransports.ParseImageName(b.output)
 		if err != nil {
-			candidates, _, err := util.ResolveName(b.output, "", b.systemContext, b.store)
+			candidates, _, _, err := util.ResolveName(b.output, "", b.systemContext, b.store)
 			if err != nil {
 				return nil, errors.Wrapf(err, "error parsing target image name %q", b.output)
 			}
--- a/new.go
+++ b/new.go
@@ -28,15 +28,14 @@ const (
 	minimumTruncatedIDLength = 3
 )
 
-func pullAndFindImage(ctx context.Context, store storage.Store, imageName string, options BuilderOptions, sc *types.SystemContext) (*storage.Image, types.ImageReference, error) {
+func pullAndFindImage(ctx context.Context, store storage.Store, transport string, imageName string, options BuilderOptions, sc *types.SystemContext) (*storage.Image, types.ImageReference, error) {
 	pullOptions := PullOptions{
 		ReportWriter:  options.ReportWriter,
 		Store:         store,
 		SystemContext: options.SystemContext,
-		Transport:     options.Transport,
 		BlobDirectory: options.PullBlobDirectory,
 	}
-	ref, err := pullImage(ctx, store, imageName, pullOptions, sc)
+	ref, err := pullImage(ctx, store, transport, imageName, pullOptions, sc)
 	if err != nil {
 		logrus.Debugf("error pulling image %q: %v", imageName, err)
 		return nil, nil, err
@@ -101,16 +100,16 @@ func newContainerIDMappingOptions(idmapOptions *IDMappingOptions) storage.IDMapp
 	return options
 }
 
-func resolveImage(ctx context.Context, systemContext *types.SystemContext, store storage.Store, options BuilderOptions) (types.ImageReference, *storage.Image, error) {
+func resolveImage(ctx context.Context, systemContext *types.SystemContext, store storage.Store, options BuilderOptions) (types.ImageReference, string, *storage.Image, error) {
 	type failure struct {
 		resolvedImageName string
 		err               error
 	}
-
-	candidates, searchRegistriesWereUsedButEmpty, err := util.ResolveName(options.FromImage, options.Registry, systemContext, store)
+	candidates, transport, searchRegistriesWereUsedButEmpty, err := util.ResolveName(options.FromImage, options.Registry, systemContext, store)
 	if err != nil {
-		return nil, nil, errors.Wrapf(err, "error parsing reference to image %q", options.FromImage)
+		return nil, "", nil, errors.Wrapf(err, "error parsing reference to image %q", options.FromImage)
 	}
+
 	failures := []failure{}
 	for _, image := range candidates {
 		var err error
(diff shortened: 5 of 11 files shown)


TomSweeneyRedHat:
Does this resolve #1316?

rhatdan:
Fixes  #1316

pull.go
```diff
@@ -152,60 +152,60 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
...
-			continue
-		}
-		img, err := is.Transport.GetStoreImage(options.Store, ref)
-		if err != nil {
-			errs = multierror.Append(errs, err)
-			continue
+			if options.ReportWriter != nil {
+				options.ReportWriter.Write([]byte("Pulling " + name + "\n"))
+			}
+			ref, err := pullImage(ctx, options.Store, name, options, systemContext)
```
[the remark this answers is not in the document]
rhatdan: Yes.  
srcRef, _, err := resolveImage(ctx, systemContext, options.Store, boptions)
Actually pulls the image.  The dump-tags is just using the resolved first image name to pull  all of the tags.

TomSweeneyRedHat:
I'd like to take a deeper dive an run some test with this Tues morn.  But if others OK this before, then fine by me.

pull.go
```diff
@@ -152,60 +152,60 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
 
 // Pull copies the contents of the image from somewhere else to local storage.
 func Pull(ctx context.Context, imageName string, options PullOptions) error {
-	spec := imageName
 	systemContext := getSystemContext(options.SystemContext, options.SignaturePolicyPath)
-	srcRef, err := alltransports.ParseImageName(spec)
```
mtrmac: If I am reading `return []string{split[1]}, false, nil` in `ResolveName` correctly, this is now broken — it only works if `options.Transport` matches the transport in the input.

_That_ problem, in turn, is masked by e.g. `buildah from` parsing the input string and setting `options.Transport` to make this all work — but non-CLI callers could be tripped by this, and it’s a very fragile way to structure the code anyway.

pull.go
```diff
@@ -152,60 +152,60 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
...
-		if options.Transport == "" {
-			options.Transport = util.DefaultTransport
-		}
-		logrus.Debugf("error parsing image name %q, trying with transport %q: %v", spec, options.Transport, err)
-		transport := options.Transport
-		if transport != util.DefaultTransport {
-			transport = transport + ":"
-		}
-		spec = transport + spec
-		srcRef2, err2 := alltransports.ParseImageName(spec)
```
mtrmac: (This is also broken, for the more exotic values of `Transport`, but I don’t quite mind `DefaultTransport` etc. going away, the after-transport-specification suffixes are fundamentally not substitutable.)

pull.go
```diff
@@ -152,60 +152,60 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
...
+		SystemContext:       systemContext,
+		PullBlobDirectory:   options.BlobDirectory,
 	}
+
 	if options.Quiet {
-		options.ReportWriter = nil // Turns off logging output
+		boptions.ReportWriter = nil // Turns off logging output
 	}
-	var names []string
+	srcRef, _, err := resolveImage(ctx, systemContext, options.Store, boptions)
```
mtrmac: The reference returned by `resolveImage` does not refer to the source (e.g. a remote registry), but to the destination (c/storage).

_If_ the source is from a `docker.Transport`, the `DockerReference()` value should be the same, so in that case this is primarily a matter of variable naming; if the source is from some other source entirely, this code will pull the other tags from some completely unrelated docker/distribution registry, based on how the pull code decided to name the pulled image.

pull.go
```diff
@@ -152,60 +152,60 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
...
+		return err
+	}
+
+	var errs *multierror.Error
 	if options.AllTags {
 		if srcRef.DockerReference() == nil {
 			return errors.New("Non-docker transport is currently not supported")
 		}
+
+		spec := transport + srcRef.DockerReference().Name()
```
mtrmac: Looking at `resolveName`, the transport value does not include the `:` separator unless it is `docker://` (or `DefaultTransport`, depending on how exactly it is hard-coded); it needs to be added.

Or, well, if you want to hard-code `docker.Transport`, just do so.

rhatdan: I would like to get rid of this code altogether and get it into a function provided by containers/image written by you or @vrothberg.  But for now, I would just like to fix the issue of buildah pull not working like buildah from.

mtrmac: \*shrug\* I don’t quite buy “this is an important feature and that justifies not getting the details right”, but in the end, I’m just pointing out breakage I am noticing; I am not a buildah maintainer and I won’t be approving the PR.

mtrmac: (To be clear, I am not saying that consolidation of the `transport + name` logic into a function is a requirement of this PR; the goal here was getting the `":"` handling right, and consolidation was a side suggestion.)

TomSweeneyRedHat:
When an image is pulled, the imageID is not being printed with this patch as it had been prior.

```
# buildah images
# buildah pull docker.io/busybox
# buildah images
IMAGE NAME                                               IMAGE TAG            IMAGE ID             CREATED AT             SIZE
docker.io/library/busybox                                latest               3a093384ac30         Dec 31, 2018 20:21     1.42 MB

```

TomSweeneyRedHat:
Not only no imageId at the end, but no progress bars.

rhatdan:
Fixed imageid and progress problems.

rhatdan:
@TomSweeneyRedHat Could you merge this one.

TomSweeneyRedHat:
Code and testing LGTM, but I'd like to get a head nod from @mtrmac to make sure his concerns were addressed.  I'm not quite sure about his second comment in the old code in particular having been addressed.

rhatdan:
I would rather get his concerns addressed in a new effort to get this code rewritten into containers/image and then vendored into podman, Buildah and CRI-O to do it the same way in all packages.

TomSweeneyRedHat:
I was hoping @mtrmac would be able to check in by now, but I'll go ahead and merge now.  I agree the right place for this functionality in my book is containers/image and it looks like we're moving in that direction.

TomSweeneyRedHat:
@rh-atomic-bot r+

mtrmac:
> Code and testing LGTM, but I'd like to get a head nod from @mtrmac to make sure his concerns were addressed. I'm not quite sure about his second comment in the old code in particular having been addressed.

AFAICS all of those review comments are still applicable.

In my experience it’s pretty much never worth it to knowingly introduce new bugs in order to get a new feature, _especially_ if introducing those bugs is not at all essential for the feature (as is the case of 
https://github.com/containers/buildah/pull/1319/files#r253626657 , which needs about three more lines; the other comments may require notably more code). But, \*shrug\*, I realize that’s a bit extremist approach, and there may not be time or, well, interest, to avoid such situations — again, not my call, but that decision did depend on me being happy with the PR, I’m afraid I’m still not.

---

On a more conceptual level, does it even make much sense to `buildah pull --all-tags fedora`? Does that mean “I want all `fedora:*` tags from all registries on the search list”, or “I want all `fedora:*` tags from the first registry on the search list that contains a `fedora`, but I don’t care at all which registry it is”? The latter seems a tiny bit more sane, but _both_ seem, fundamentally, rather useless. When would anyone _ever_ want to use either?

So, maybe, at the highest level, there should be a separate `PullAllTags` + `Pull`, with `PullAllTags` not using the search path at all (and minimally modified), and `Pull` modified roughly like this PR? That would also nicely bypass the `":"` and the `srcRef` concerns (but not the other `alltransports.ParseImageName` semantics breakage).

OTOH doing that the easy way would cause `PullAllTags` to (continue to) use the `docker.io/library` normalization, which may be considered very undesirable.

mtrmac:
> I would rather get his concerns addressed in a new effort to get this code rewritten into containers/image and then vendored into podman, Buildah and CRI-O to do it the same way in all packages.

(BTW don’t expect miracles from this — as https://github.com/containers/libpod/blob/9c1b08fd79012c12d478edb13f6057a0414762db/libpod/image/parts.go#L54 shows, rewrites in c/image won’t free libpod/buildah from making decisions about semantics of the input; if anything, it would make it less avoidable for them to make such decisions.)

rhatdan:
@mtrmac @nalind 
I reworked this PR Heavily to remove the handling of Transports from the clients, and moved it into the library.

I made the Transports private to the library to prevent the client passing in a transport different then the image.

Transport=docker
ImageName=ociarchive:/tmp/foobar

Also added OCI As a valid format.

util.ResolveImage now returns the image name without the transport as well as the transport.

docs/buildah-pull.md
```diff
@@ -25,6 +25,9 @@ Multiple transports are supported:
   **docker-daemon:**_docker-reference_
   An image _docker-reference_ stored in the docker daemon's internal storage.  _docker-reference_ must include either a tag or a digest.  Alternatively, when reading images, the format can also be docker-daemon:algo:digest (an image ID).
 
+  **oci:**_path_**:**_tag_**
+  An image tag in a directory compliant with "Open Container Image Layout Specification" at path.
```
TomSweeneyRedHat: nit, s/path/_path_/

TomSweeneyRedHat: Ugh love GitHub auto md sometimes

`s/path/_path_/`

TomSweeneyRedHat:
@nalind and/or @vrothberg PTAL

buildah.go
```diff
@@ -340,7 +340,7 @@ type BuilderOptions struct {
 	// needs to be pulled and the image name alone, or the image name and
 	// the registry together, can not be resolved to a reference to a
 	// source image.  No separator is implicitly added.
-	Transport string
```
TomSweeneyRedHat: Just wondering why you switched this to lower case?  Also in pull.go.  I understand the scoping, but none of the other variables in these structs are lowercase.  In some of the code we have a number of variables named transport(s) being used and having this upper case helps it to stand out better.

rhatdan: I don't want callers to use this field.  We use it to pass the information around withing buildah library, but specifying transport as well as imagename can cause conflicts.  
Bottom line, I only want transport derived in one place and from ImageName.  Before we had it being derived in different places,and not the same way.

pull.go
```diff
@@ -215,13 +222,10 @@ func pullImage(ctx context.Context, store storage.Store, imageName string, optio
 	spec := imageName
 	srcRef, err := alltransports.ParseImageName(spec)
 	if err != nil {
```
mtrmac: The two callers of `pullImage` now explicitly know whether the input should be readable by `ParseImageName` or whether it needs the transport added (and using the other path can misinterpret the input); it would be nice to make this unambiguous.

`pullAndFindImage` AFAICS currently always submits input input with the “concatenate `options.transport`+`imageName`” format, while the `AllTags` path in `Pull` always submits input with the “use `ParseImageName(imageName)` semantics — so, the concatenation code could be moved to `pullAndFindImage` (or maybe even higher to `resolveImage`, and consolidated with the other part there), and `pullImage` could expect input in the `ParseImageName` format.

(This is a pre-existing situation, so non-blocking; the changes to `ResolveName` in this PR now allow making this unambiguous.)

buildah.go
```diff
@@ -340,7 +340,7 @@ type BuilderOptions struct {
...
+	transport string
```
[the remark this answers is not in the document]
rhatdan: Removing.

cmd/buildah/pull.go
```diff
@@ -80,27 +78,14 @@ func pullCmd(c *cobra.Command, args []string, iopts pullResults) error {
 		return err
 	}
 
-	transport := util.DefaultTransport
-	arr := strings.SplitN(args[0], ":", 2)
-	if len(arr) == 2 {
-		if iopts.allTags {
-			return errors.Errorf("tag can't be used with --all-tags")
-		}
```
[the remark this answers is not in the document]
rhatdan: This is tough to check, and seem fairly arbitrary.  If I state I want to pull --alltags on alpine:latest, I don't see this as a big conflict.

imagebuildah/build.go
```diff
@@ -582,7 +577,6 @@ func NewExecutor(store storage.Store, options BuildOptions) (*Executor, error) {
 		contextDir:                     options.ContextDirectory,
 		pullPolicy:                     options.PullPolicy,
 		registry:                       options.Registry,
-		transport:                      options.Transport,
```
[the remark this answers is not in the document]
rhatdan: Removed

util/util.go
```diff
@@ -70,37 +48,39 @@ var (
 // and a boolean which is true iff 1) the list of search registries was used, and 2) it was empty.
```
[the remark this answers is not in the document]
rhatdan: Added Docs.

util/util.go
```diff
@@ -70,37 +48,39 @@ var (
...
+		return nil, "", false, nil
 	}
 
 	// Maybe it's a truncated image ID.  Don't prepend a registry name, then.
 	if len(name) >= minimumTruncatedIDLength {
 		if img, err := store.Image(name); err == nil && img != nil && strings.HasPrefix(img.ID, name) {
 			// It's a truncated version of the ID of an image that's present in local storage;
 			// we need only expand the ID.
-			return []string{img.ID}, false, nil
+			return []string{img.ID}, DefaultTransport, false, nil
```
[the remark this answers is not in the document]
rhatdan: Changed to ""

util/util.go
```diff
@@ -70,37 +48,39 @@ var (
...
 		}
 	}
 
-	// If the image includes a transport's name as a prefix, use it as-is.
-	split := strings.SplitN(name, ":", 2)
-	if len(split) == 2 {
-		if _, ok := Transports[split[0]]; ok {
-			return []string{split[1]}, false, nil
+	if strings.HasPrefix(name, DefaultTransport) {
+		name = strings.TrimPrefix(name, DefaultTransport)
```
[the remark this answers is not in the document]
rhatdan: Fixed.

new.go
```diff
@@ -33,7 +33,7 @@ func pullAndFindImage(ctx context.Context, store storage.Store, imageName string
 		ReportWriter:  options.ReportWriter,
 		Store:         store,
 		SystemContext: options.SystemContext,
-		Transport:     options.Transport,
+		transport:     options.transport,
```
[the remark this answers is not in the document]
rhatdan: Made into a parameter.

new.go
```diff
@@ -250,21 +253,22 @@ func findUnusedContainer(name string, containers []storage.Container) string {
...
+		img *storage.Image
+		err error
+	)
 	if options.FromImage == BaseImageFakeName {
 		options.FromImage = ""
 	}
-	if options.Transport == "" {
-		options.Transport = util.DefaultTransport
+	if options.transport == "" {
+		options.transport = util.DefaultTransport
```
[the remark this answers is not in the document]
rhatdan: Removed

pull.go
```diff
@@ -152,60 +162,57 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
...
-		options.ReportWriter = nil // Turns off logging output
+		boptions.ReportWriter = nil // Turns off logging output
+	}
+	srcRef, transport, img, err := resolveImage(ctx, systemContext, options.Store, boptions)
+	if err != nil {
+		return err
 	}
-	var names []string
+
+	fmt.Printf("%s\n", img.ID)
```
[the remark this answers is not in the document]
rhatdan: Fixed.

pull.go
```diff
@@ -152,60 +162,57 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
...
+		SystemContext:       systemContext,
+		PullBlobDirectory:   options.BlobDirectory,
+		ReportWriter:        options.ReportWriter,
 	}
+
 	if options.Quiet {
-		options.ReportWriter = nil // Turns off logging output
+		boptions.ReportWriter = nil // Turns off logging output
+	}
+	srcRef, transport, img, err := resolveImage(ctx, systemContext, options.Store, boptions)
```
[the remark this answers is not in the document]
rhatdan: If I don't test for the conflict of TAG versus all-tag, then the user must specify a legal tag if the :latest does not exist to do a pull-all. 

Is that adequate for now.

pull.go
```diff
@@ -152,60 +162,57 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
...
-		if err != nil {
-			errs = multierror.Append(errs, err)
-			continue
-		}
-		img, err := is.Transport.GetStoreImage(options.Store, ref)
-		if err != nil {
-			errs = multierror.Append(errs, err)
-			continue
+			if options.ReportWriter != nil {
+				options.ReportWriter.Write([]byte("Pulling " + name + "\n"))
```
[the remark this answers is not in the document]
rhatdan: Removed

pull.go
rhatdan: Lets handle this in a separate PR

util/util.go
```diff
@@ -43,64 +40,47 @@ var (
...
 			// we need only expand the ID.
-			return []string{img.ID}, false, nil
+			return []string{img.ID}, "", false, nil
 		}
 	}
 
+	if strings.HasPrefix(name, DefaultTransport) {
+		return []string{strings.TrimPrefix(name, DefaultTransport)}, DefaultTransport, false, nil
+	}
 	// If the image includes a transport's name as a prefix, use it as-is.
```
mtrmac: Tiny nit: This comment would better fit above the `DefaultTransport` case.

pull.go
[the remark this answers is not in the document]
mtrmac: At the very least, `s/srcRef/destRef/` or `s/srcRef/storageRef/` to not be confusing about the semantics of that value, please.

---

WRT requiring `:latest` to exist:

Well, it’s a breaking change AFAICS; this used to work. And forcing users to specify a tag would make reintroducing the “`--all-tags` with a tag is invalid” error _another_ breaking change.

\*shrug\* I don’t quite understand the utility of `--all-tags` (let alone search+`--all-tags` together) in the first place, so I guess it is acceptable.

pull.go
```diff
@@ -152,76 +154,71 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
...
 		if err != nil {
 			return errors.Wrapf(err, "error getting repository tags")
 		}
 		for _, tag := range tags {
 			name := spec + ":" + tag
-			names = append(names, name)
+			if options.ReportWriter != nil {
+				options.ReportWriter.Write([]byte("Pulling " + name + "\n"))
+			}
+			ref, err := pullImage(ctx, options.Store, transport, name, options, systemContext)
```
mtrmac: Non-blocking: 

For consistency with the `pullAndFindImage` caller, it would be nice if `name` here did not start with `transport`:
```go
repoName := whateverSrcRefIsRenamedTo.DockerReference().Name()
srcRef, err := alltransports.ParseImageName(transport+repoName)
// or srcRef, err := c/image/docker.NewReference(srcRef.DockerReference())
…
name = repoName + ":" + tag
```
or so.

util/util.go
rhatdan: Moved

rhatdan:
@mtrmac Ready to merge.  I will open another PR to further clean this up and eventually get it into podman and maybe add the functions to containers/image

@rh-atomic-bot r+

mtrmac:
CK. Thanks!

mtrmac:
ACK, that is.

Issue #1296, opened 2019-01-18
buildah inside of a podman container runs in rootless mode


rhatdan:
It is running as root but still puts its content in /root/.local ...

It should be using /var/lib/containers/.

rhatdan:
This is breaking some demos.

giuseppe:
is the podman container using namespaces?

TomSweeneyRedHat:
@giuseppe any further thoughts?

giuseppe:
PR here: https://github.com/containers/buildah/pull/1323

PR #1323, opened 2019-02-05
unshare: do not set rootless mode if euid=0

giuseppe:
when doing an unshare, do not set rootless mode unless the euid != 0.

Closes: https://github.com/containers/buildah/issues/1296


rhatdan:
@rh-atomic-bot r+

PR #1324, opened 2019-02-06
main: fix rootless mode

giuseppe:
the args argument is always an empty slice, so it breaks the code for
detecting if we need a new user namespace.  Use directly os.Args[1:].

### DIFF
--- a/cmd/buildah/main.go
+++ b/cmd/buildah/main.go
@@ -98,7 +98,7 @@ func before(cmd *cobra.Command, args []string) error {
 	if globalFlagResults.Debug {
 		logrus.SetLevel(logrus.DebugLevel)
 	}
-	maybeReexecUsingUserNamespace(args, false)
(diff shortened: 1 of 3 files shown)


giuseppe:
hm.. this will probably not work when there are options, still working on it

giuseppe:
it will still create a userns in the case of "buildah --debug version" where "buildah version" doesn't.

@baude do you know why the args is empty?

giuseppe:
fixed now in the new version :arrow_up:

rhatdan:
bot, retest this please
@baude PTAL

baude:
yeah LGTM!

giuseppe:
fixed the Darwin CI

rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot retry

Issue #1305, opened 2019-01-29
umask 0077 breaks build-using-dockerfile with USER statement


praiskup:
BUG REPORT INFORMATION
---------------------------------------------------

```
$ cat Dockerfile
FROM registry.fedoraproject.org/fedora:29
RUN useradd jdoe
USER jdoe
RUN id
$ buildah bud .
STEP 1: FROM registry.fedoraproject.org/fedora:29
STEP 2: RUN useradd jdoe
STEP 3: USER jdoe
STEP 4: RUN id
container_linux.go:344: starting container process caused "chdir to cwd (\"/\") set in config.json failed: permission denied"
error running container: error creating container for [/bin/sh -c id]: : exit status 1
error building at step {Env:[DISTTAG=f29container FGC=f29  PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin] Command:run Args:[id] Flags:[] Attrs:map[] Message:RUN id Original:RUN id}: error while running runtime: exit status 1
ERRO[0001] exit status 1
$ rpm -q buildah
buildah-1.6-2.git5f95bd9.fc29.x86_64
```

**Description**

Ideally, buildah should make sure that the created `/` directory is created a standard way (say chmod 666 per filesystem.rpm content), or if it at least warned properly.

**Describe the results you received:**

`buildah bud` ends with strange permission error with umask 0077 after `USER` command/statement.

**Describe the results you expected:**

`buildah bud` either works under those contitions.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.6-2.git5f95bd9.fc29.x86_64
```

**Output of `buildah version`:**

```
Version:         1.6
Go Version:      go1.11.4
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.4.0
libcni Version:  
Git Commit:      
Built:           Thu Jan  1 01:00:00 1970
OS/Arch:         linux/amd64
```
**Output of `cat /etc/*release`:**

```Fedora release 29 (Twenty Nine)
NAME=Fedora
VERSION="29 (Twenty Nine)"
ID=fedora
VERSION_ID=29
VERSION_CODENAME=""
PLATFORM_ID="platform:f29"
PRETTY_NAME="Fedora 29 (Twenty Nine)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:29"
[...]

PR #1325, opened 2019-02-06
buildah: force umask 022

giuseppe:
an umask more restrictive can break some images.

Closes: https://github.com/containers/buildah/issues/1305

### DIFF
--- a/cmd/buildah/common.go
+++ b/cmd/buildah/common.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"os"
+	"syscall"
 	"time"
 
 	"github.com/containers/buildah"
@@ -13,6 +14,7 @@ import (
 	"github.com/containers/storage"
 	"github.com/opencontainers/go-digest"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
@@ -86,6 +88,11 @@ func getStore(c *cobra.Command) (storage.Store, error) {
 		}
 		options.UIDMap = uidmap
 		options.GIDMap = gidmap
+	}
+
+	oldUmask := syscall.Umask(0022)
+	if (oldUmask & ^0022) != 0 {
+		logrus.Debugf("umask value too restrictive.  Forcing it to 022")
 	}
 
 	store, err := storage.GetStore(options)


cmd/buildah/common.go
```diff
@@ -88,6 +90,11 @@ func getStore(c *cobra.Command) (storage.Store, error) {
 		options.GIDMap = gidmap
 	}
 
+	oldUmask := syscall.Umask(0)
+	if oldUmask != 0 {
+		logrus.Warningf("umask value different than 0.  Forcing it to 0")
```
rhatdan: Doesn't this mean that every running of buildah will cause this to happen?  IE Lots and lots of Warning messages.

rhatdan: When I set my umask to 0, and create a file, I end up creating a file 0666, which is not what we want, it is a security risk.
Should we set the default to 0002, or 0022

[dwalsh@localhost libpod]$ umask 0000; rm -f ~/walsh; touch ~/walsh; ls -l ~/walsh
-rw-rw-rw-. 1 dwalsh dwalsh 0 Feb  6 05:41 /home/dwalsh/walsh
[dwalsh@localhost libpod]$ umask 0022; rm -f ~/walsh; touch ~/walsh; ls -l ~/walsh
-rw-r--r--. 1 dwalsh dwalsh 0 Feb  6 05:41 /home/dwalsh/walsh
[dwalsh@localhost libpod]$ umask 0002; rm -f ~/walsh; touch ~/walsh; ls -l ~/walsh
-rw-rw-r--. 1 dwalsh dwalsh 0 Feb  6 05:41 /home/dwalsh/walsh

rhatdan: Looks like the default login is supposed to be 0022. Although when I setup a new account, I see it as 0002.

giuseppe: I can change it to Debug

praiskup: There's this check in /etc/bashrc, it might be related.
```
    if [ $UID -gt 199 ] && [ "`id -gn`" = "`id -un`" ]; then
       umask 002
    else
       umask 022
    fi
```

rhatdan: I guess someone decided if you have the same Username and Group name by default then files you create should be 
0664
If not they should be
0644

So wouldn't everything work correctly if the umask was 0022

praiskup: There's also the `$UID -gt 199` condition, so **only** system-defined UIDs like `postgres` or `mariadb` have 0022 umask.  I'm not sure, unless there's a reason to be more restrictive (some playing with volumes regarding group permissions, e.g.) I'd probably go with 0002.

giuseppe:
changed it to 022, which is also the default used by systemd

PR #1326, opened 2019-02-06
pull.bats updates

nalind:
Don't use /tmp for temporary files in `pull.bats`, use the per-test temporary directory.  Use our known registry.conf file when ever we communicate with a registry.  Don't run "mkdir" using the `run` function.  Check the exit status of more commands that we run under `run`.

### DIFF
--- a/tests/pull.bats
+++ b/tests/pull.bats
@@ -35,6 +35,7 @@ load helpers
   [ "$status" -eq 0 ]
   run buildah images --format "{{.Name}}:{{.Tag}}"
   echo "$output"
+  [ "$status" -eq 0 ]
   [[ "$output" =~ "busybox:glibc" ]]
   [[ "$output" =~ "busybox:latest" ]]
   run buildah pull --signature-policy ${TESTSDIR}/policy.json quay.io/libpod/alpine_nginx:latest
@@ -42,6 +43,7 @@ load helpers
   [ "$status" -eq 0 ]
   run buildah images --format "{{.Name}}:{{.Tag}}"
   echo "$output"
+  [ "$status" -eq 0 ]
   [[ "$output" =~ "alpine_nginx:latest" ]]
   run buildah rmi quay.io/libpod/alpine_nginx:latest
   echo "$output"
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@TomSweeneyRedHat @umohnani8 @vrothberg @giuseppe PTAL

tests/pull.bats
```diff
@@ -27,102 +27,103 @@ load helpers
...
+  run buildah pull --signature-policy ${TESTSDIR}/policy.json docker-archive:${TESTDIR}/alp.tar
   echo "$output"
   [ "$status" -eq 0 ]
   run buildah images --format "{{.Name}}:{{.Tag}}"
   echo "$output"
-  [[ "$output" =~ "alpine" ]]
-  run rm -f /tmp/alp.tar
-  echo "$output"
   [ "$status" -eq 0 ]
+  [[ "$output" =~ "alpine" ]]
```
TomSweeneyRedHat: does ${TESTDIR}/alp.tar get auto removed?  I think so, but just want to double check.

tests/pull.bats
```diff
@@ -158,24 +163,24 @@ load helpers
...
+  run buildah pull --signature-policy ${TESTSDIR}/policy.json ostree:alpine@${TESTDIR}/ostree-repo
   echo "$output"
   [ "$status" -eq 0 ]
   run buildah images --format "{{.Name}}:{{.Tag}}"
   echo "$output"
-  [[ "$output" =~ "ostree-repo:latest" ]]
-  run rm -rf /tmp/ostree-repo
-  echo "$output"
   [ "$status" -eq 0 ]
+  [[ "$output" =~ "ostree-repo:latest" ]]
```
TomSweeneyRedHat: ditto removal of ${TESTDIR}/ostree-repo is automatic?

TomSweeneyRedHat:
Just two verification questions that are related, otherwise LGTM.  Nice cleanup Nalin.

tests/pull.bats
giuseppe: on teardown TESTDIR is rm completely.

giuseppe:
LGTM

@rh-atomic-bot r+

tests/pull.bats
TomSweeneyRedHat: Thanks for the confirmation @giuseppe , I thought that was the case but couldn't remember last night.

PR #1333, opened 2019-02-12
Fix buildah commit compress by default

steveb:
The change to switch over to cobra had the effect of always setting
compress to imagebuildah.Uncompressed. This is because the conditional
to set it to Gzip always returns false regardless of whether
--disable-compression is set or not.

This change fixes that, and make the logic match the intent of Gzip
being the default option.

### DIFF
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -93,9 +93,9 @@ func commitCmd(c *cobra.Command, args []string, iopts commitInputOptions) error 
 		return errors.Errorf("too many arguments specified")
 	}
 	image := args[0]
-	compress := imagebuildah.Uncompressed
-	if c.Flag("disable-compression").Changed && !iopts.disableCompression {
-		compress = imagebuildah.Gzip
+	compress := imagebuildah.Gzip
+	if iopts.disableCompression {
+		compress = imagebuildah.Uncompressed
 	}
 	timestamp := time.Now().UTC()
 	if c.Flag("reference-time").Changed {
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -214,10 +214,9 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budResults) error {
 		logrus.Debugf("--compress option specified but is ignored")
 	}
 
-	compression := imagebuildah.Uncompressed
-
-	if c.Flag("disable-compression").Changed && !iopts.DisableCompression {
-		compression = imagebuildah.Gzip
+	compression := imagebuildah.Gzip
+	if iopts.DisableCompression {
+		compression = imagebuildah.Uncompressed
 	}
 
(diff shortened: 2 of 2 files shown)


rhatdan:
@steveb You never signed your commit.

git commit -a --amend -s
git push --force

LGTM
@baude @TomSweeneyRedHat @nalind PTAL

TomSweeneyRedHat:
@vrothberg PTAL

cmd/buildah/commit.go
```diff
@@ -93,9 +93,9 @@ func commitCmd(c *cobra.Command, args []string, iopts commitInputOptions) error
 		return errors.Errorf("too many arguments specified")
 	}
 	image := args[0]
-	compress := imagebuildah.Uncompressed
-	if c.Flag("disable-compression").Changed && !iopts.disableCompression {
-		compress = imagebuildah.Gzip
+	compress := imagebuildah.Gzip
+	if iopts.disableCompression {
+		compress = imagebuildah.Uncompressed
```
TomSweeneyRedHat: Maybe I'm light on tea, bu the man page says: 

```
-disable-compression, -D

Don't compress filesystem layers when building the image unless it is required by the location where the image is being written. This is the default setting, because image layers are compressed automatically when they are pushed to registries, and images being written to local storage would only need to be decompressed again to be stored. Compression can be forced in all cases by specifying --disable-compression=false.
```

So I think the default should be Uncompressed at line 96 and then `if !iopts.disableCompression {` (note changing to false), then setting compression to Gzip.

@nalind, WDYT?

TomSweeneyRedHat: FWIW, I think the `c.Flag("disable-compression").Changed is probably not needed.

steveb:
Thanks for the reviews

steveb:
bud.go has the exact same issue, let me see if there are any other obvious cases of this

steveb:
push.go does the right thing, I'll make bud.go and commit.go consistent with that.

steveb:
push.go, for reference https://github.com/containers/buildah/blob/master/cmd/buildah/push.go#L98-L101

cmd/buildah/commit.go
steveb: I went with the same logic in push.go, to make all three consistent

TomSweeneyRedHat:
@steveb I've enough tea in me today, and the change looks good.  However I think we've an issue where disable-compression is being set to false by default.  I'm going to spin another PR up for that.
LGTM

TomSweeneyRedHat:
@rh-atomic-bot r+

rhatdan:
I believe this is wrong, after talking to @nalind We should default all "commits" to uncompressed and only compress on buildah push.

Whether the user is calling buildah bud or buildah commit directly we should not be compressing.  Only compress when sending image over the internet.

PR #1334, opened 2019-02-12
unshare: reexec using a memfd copy instead of the binary

nalind:
When we're unshare()ing, copy `/proc/self/exe` into a memory FD before exec()ing it, so that the descriptor which we leak into our child process's namespace through the /proc/$parent/exe link is one that gets discarded after the process exits, and seal it against modifications as soon as we're finished populating it, mimicking patches for other projects that mitigate CVE-2019-5736.

If `github.com/containers/storage/pkg/reexec` already required `cgo`, so that adding this there didn't add a new build requirement, it would have been better to do this there.

### DIFF
--- a/unshare/unshare.c
+++ b/unshare/unshare.c
@@ -2,6 +2,8 @@
 #include <sys/types.h>
 #include <sys/ioctl.h>
 #include <sys/stat.h>
+#include <sys/syscall.h>
+#include <linux/memfd.h>
 #include <fcntl.h>
 #include <grp.h>
 #include <sched.h>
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@giuseppe PTAL

unshare/unshare.c
```diff
@@ -59,6 +83,120 @@ static void _check_proc_sys_file(const char *path)
...
+	fd = open(list, O_RDONLY);
+	if (fd == -1) {
+		return NULL;
+	}
+	buf = NULL;
+	size = 0;
+	used = 0;
+	for (;;) {
+		new_size = used + BUFSTEP;
+		buf = realloc(buf, new_size);
```
giuseppe: it will probably trigger a warning in the static analysis as we don't free the previous value of `buf` in case of failure

unshare/unshare.c
```diff
@@ -59,6 +83,120 @@ static void _check_proc_sys_file(const char *path)
...
+	if (fd == -1) {
+		fprintf(stderr, "open(\"/proc/self/exe\"): %m\n");
+		return -1;
+	}
+	if (fstat(fd, &st) == -1) {
+		fprintf(stderr, "fstat(\"/proc/self/exe\"): %m\n");
+		return -1;
+	}
+	basename = argv[0];
+	while (strchr(basename, '/') != NULL) {
```
giuseppe: could we just use `basename(3)` here?

nalind: Yes.  Changing it.

unshare/unshare.c
nalind: Worth fixing, then.

giuseppe:
thanks, LGTM

giuseppe:
@rh-atomic-bot r+

Issue #1336, opened 2019-02-13
--disable-compression is set to false by default and should be true.


TomSweeneyRedHat:
**Description**

The man page for `buildah commit` and others state that `--disable-compression` is the default setting or set to true.  However the code has it set to false by default.

***Output of `buildah version`:**

```
buildah v1.7-dev
```

PR #1337, opened 2019-02-13
Set --disable-compression to true as documented

TomSweeneyRedHat:

The man pages for `--disable-compression` indicates that it is
true by default.  However the code has it set to false.  This changes
that.

Fixes: #1336

### DIFF
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -59,7 +59,7 @@ func init() {
 
 	flags.StringVar(&opts.certDir, "cert-dir", "", "use certificates at the specified path to access the registry")
 	flags.StringVar(&opts.creds, "creds", "", "use `[username[:password]]` for accessing the registry")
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
@nalind, PTAL

rhatdan:
@TomSweeneyRedHat After talking to @nalind we should default to noncompression on buildah commit, and only compress on buildah push.

nalind:
The man page was right as of #1274, so we broke the code somewhere.

TomSweeneyRedHat:
OK, thanks @dwalsh and @nalind, I'll try to sort it out tomorrow.

TomSweeneyRedHat:
OK, think it's set now with the latest push if the tests behave.  Commit and bud do NOT compress by default.  --disable-compression  is set to TRUE by default.

Push DOES compress by default.  --disable-compression is set to FALSE by default.

rhatdan:
@rh-atomic-bot r+

PR #1338, opened 2019-02-13
healthcheck start-period: update documentation

nalind:
A re-reading of the spec doesn't agree with our docs - the healthcheck start period isn't a delay in the first invocation of the healthcheck command, but is instead a grace period where failures aren't counted as indications of problems.  Instead, during that grace period, failures are attributed to the container not yet having finished starting.  Once the healthcheck command succeeds, the engine should start counting
failures.

### DIFF
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -77,7 +77,7 @@ func init() {
 	flags.StringVar(&opts.healthcheck, "healthcheck", "", "set a `healthcheck` command for the target image")
 	flags.StringVar(&opts.healthcheckInterval, "healthcheck-interval", "", "set the `interval` between runs of the `healthcheck` command for the target image")
 	flags.IntVar(&opts.healthcheckRetries, "healthcheck-retries", 0, "set the `number` of times the `healthcheck` command has to fail")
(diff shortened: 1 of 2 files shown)


cmd/buildah/config.go
```diff
@@ -77,7 +77,7 @@ func init() {
 	flags.StringVar(&opts.healthcheck, "healthcheck", "", "set a `healthcheck` command for the target image")
 	flags.StringVar(&opts.healthcheckInterval, "healthcheck-interval", "", "set the `interval` between runs of the `healthcheck` command for the target image")
 	flags.IntVar(&opts.healthcheckRetries, "healthcheck-retries", 0, "set the `number` of times the `healthcheck` command has to fail")
-	flags.StringVar(&opts.healthcheckStartPeriod, "healthcheck-start-period", "", "set the amount of `time` to wait after starting a container before running a `healthcheck` command")
+	flags.StringVar(&opts.healthcheckStartPeriod, "healthcheck-start-period", "", "set the amount of `time` to wait after starting a container before a failed `healthcheck` command will count")
```
TomSweeneyRedHat: "will count as a failure" perhaps?

nalind: Updated.

rhatdan:
@rh-atomic-bot r+

PR #1340, opened 2019-02-14
dumpbolt: handle nested buckets

nalind:
When iterating through a bucket, a key with no value is a sub-bucket.  Dump the contents of those buckets after the non-sub-buckets.


TomSweeneyRedHat:
FYI @mheon

mheon:
@nalind TY for looking into this - we use sub-buckets a lot in the libpod schema

TomSweeneyRedHat:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #1352, opened 2019-02-20
Make "images --all" faster

nalind:
Computing `imageIsParent()` for every image can get expensive, and if `--all` was specified, it's entirely unnecessary.


rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #1353, opened 2019-02-20
Users should not be able to override the registries.conf rules

rhatdan:
We should not support home directory registries.conf rules.
If an admin blocks container registries, then it should be blocked
in rootless mode also.  Similarly if he only wants users pulling
from a single container registry we should only have that registry listed.

### DIFF
--- a/common.go
+++ b/common.go
@@ -7,7 +7,6 @@ import (
 
 	cp "github.com/containers/image/copy"
 	"github.com/containers/image/types"
-	"github.com/containers/libpod/pkg/rootless"
 )
 
 const (
@@ -25,11 +24,8 @@ func getCopyOptions(reportWriter io.Writer, sourceReference types.ImageReference
 	if sourceSystemContext != nil {
 		*sourceCtx = *sourceSystemContext
 	} else {
-		if rootless.IsRootless() {
-			if _, err := os.Stat(userRegistriesFile); err == nil {
-				sourceCtx.SystemRegistriesConfPath = userRegistriesFile
-			}
-
+		if _, err := os.Stat(userRegistriesFile); err == nil {
+			sourceCtx.SystemRegistriesConfPath = userRegistriesFile
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
@giuseppe PTAL.  I thought there was some issue getting to the "main" registries.conf when rootless.  If not, then I'm fine with this.

giuseppe:
I think we should merge the two configurations somehow.  Why we wouldn't give users the possibility to customize this file if they wish?

vrothberg:
> I think we should merge the two configurations somehow. Why we wouldn't give users the possibility to customize this file if they wish?

I agree with @giuseppe. Users should be able to configure registries as well. However, an administrator should also be able to enforce certain rules on the users. I think that merging registries is too complex and it can really complex not only in the code but also for a user trying to understand error messages etc.

I suggest extending the registries.conf with a switch, e.g., `enforce_for_all_users=true`. Setting it to true will enforce using the global configuration, while false allows users to use their own configuration.

mtrmac:
> We should not support home directory registries.conf rules.
> If an admin blocks container registries, then it should be blocked
> in rootless mode also. Similarly if he only wants users pulling
> from a single container registry we should only have that registry listed.

Where is the trust boundary / enforcement point for this?

AFAICS `buildah` is an ordinary non-setuid program, so the user can, almost always (except for _extremely strictly_ hardened systems), compile or download a different `buildah` binary that does not implement this admin enforcement.

So, there is no true boundary, at most a convenience hurdle.

(That may still be enough, I just want everyone to be clear about what we are doing.)

rhatdan:
I understand that, but  the same thing happens when users leak the docker socket, but the docker socket is blocked by the way the docker daemon is configured.  Of course the user could override this or download podman or skopeo or any tool.  

We are just putting hurdles in front of the user to make the realize when they are being evil.

rhatdan:
I have lost this battle for now.

PR #1358, opened 2019-02-21
Stop printing default twice in cli --help

rhatdan:
Currently we are printing two defaults.  This change will
let the flags package print out the defaults.

### DIFF
--- a/pkg/cli/common.go
+++ b/pkg/cli/common.go
@@ -125,7 +125,7 @@ func GetNameSpaceFlags(flags *NameSpaceResults) pflag.FlagSet {
 func GetLayerFlags(flags *LayerResults) pflag.FlagSet {
 	fs := pflag.FlagSet{}
 	fs.BoolVar(&flags.ForceRm, "force-rm", false, "Always remove intermediate containers after a build, even if the build is unsuccessful.")
-	fs.BoolVar(&flags.Layers, "layers", false, fmt.Sprintf("cache intermediate layers during build. Use BUILDAH_LAYERS environment variable to override. (default %t)", UseLayers()))
+	fs.BoolVar(&flags.Layers, "layers", UseLayers(), fmt.Sprintf("cache intermediate layers during build. Use BUILDAH_LAYERS environment variable to override."))
 	return fs
 }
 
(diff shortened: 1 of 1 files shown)


rhatdan:
@nalind @TomSweeneyRedHat @giuseppe @vrothberg PTAL

baude:
looks correct! if a value is specified, it will be doc'd as the default value in the help.

TomSweeneyRedHat:
OK, I'm confused, but it could just be tired eyeballs.  Without this change `buildah help` is returning this without this change:
```
      --layers                        cache intermediate layers during build. Use BUILDAH_LAYERS environment variable to override. (default false)
```
and with it, I'm seeing:
```
--layers                        cache intermediate layers during build. Use BUILDAH_LAYERS environment variable to override.
```

I'm not seeing the double default being printed in the original?    Then the new one doesn't seem to show the default, which I think we should keep?

rhatdan:
@TomSweeneyRedHat setenv BUILDAH_LAYERS true.

rhatdan:
@TomSweeneyRedHat Actually the problem only happens when this code is vendored into podman.
```
$ podman build --help | grep default.*default
      --force-rm                                   Always remove intermediate containers after a build, even if the build is unsuccessful. (default true) (default true)
      --rm                                         Remove intermediate containers after a successful build (default true) (default true)
```

TomSweeneyRedHat:
OK, now I'm a bit more confused.  I thought the CLI help in Buildah wasn't used by Podman?  @baude did I miss this?

rhatdan:
Yes this is vendored into podman.

TomSweeneyRedHat:
@rhatdan we talked over IRC and you said the Cobra shows the default value if it is true.  But that seems to only happen on the Podman side, but not the Buildah side?   @baude is there some bit we have to jiggle on the Buildah side to make this happen?

rhatdan:
Anyways can we get this merged?  To me these changes make sense whether or not the double default is happening.

TomSweeneyRedHat:
I'd rather not merge it until we can figure out how to get the default to show on the Buildah help which it's missing now.

TomSweeneyRedHat:
OK, DOH moment here.  I was looking at the --layers output for 'buildah bud' versus 'podman build'.  One was showing the (default true) with these changes and the other wasn't.  Took me a bit to remember that the default for layers is false on Buildah, but true on Podman.  That's why it wasn't showing in both players.  Personally I'd still rather not have the default values showing if they're only true, but the disparity which was worrying me now make sense.

TomSweeneyRedHat:
@rh-atomic-bot r+

PR #1373, opened 2019-02-27
Hide from users command options that we don't want them to use.

rhatdan:
We really want users to use the registries.conf file specified by
the administrator.  We also don't want them overriding the signature-policy.
Thes options are mainly their for CI/CD Systems, so they should be hidden
from the user.


rhatdan:
@giuseppe @vrothberg @TomSweeneyRedHat @nalind @pixdrift PTAL

giuseppe:
LGTM

vrothberg:
LGTM (as they're still usable but hidden). I prefer @TomSweeneyRedHat to ack before merging.

rhatdan:
@TomSweeneyRedHat PTAL

TomSweeneyRedHat:
LGTM
Thanks for the cleanup @rhatdan

TomSweeneyRedHat:
@rh-atomic-bot r+

Issue #1374, opened 2019-02-27
buildah unshare parses flags when it shouldn't [1.7]


TomasTomecek:
**Description**

```
$ buildah unshare ls -l
unknown shorthand flag: 'l' in -l

$ buildah unshare 'ls -l'
ERRO[0000] exec: "ls -l": executable file not found in $PATH
ERRO[0000] (unable to determine exit status)
ERRO[0000] exit status 1
```

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.7-1.git873f001.fc29.x86_64
```

**Output of `buildah version`:**

```
Version:         1.7
Go Version:      go1.11.5
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.4.0
libcni Version:  
Git Commit:      
Built:           Thu Jan  1 01:00:00 1970
OS/Arch:         linux/amd64
```

I went back to [1.6](https://koji.fedoraproject.org/koji/buildinfo?buildID=1181308) and it works just fine.

rhatdan:
You can do.
```
buildah unshare -- ls -l 
```

TomasTomecek:
Thanks, will try.

rhatdan:
BTW I think this is a real bug.

rhatdan:
Fixed in master.

PR #1375, opened 2019-02-27
fix unshare option handling and documentation

rhatdan:
Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

PR #1377, opened 2019-02-28
unshare: fix unshare parses flags

q384566678:
Fixes: #1374 

Fix buildah unshare parses flags when it shouldn't:

```
➜  cmd git:(unshare-fix) buildah unshare ls -l
unknown shorthand flag: 'l' in -l
```

After change:
```
➜  buildah git:(unshare-fix) ./buildah unshare ls -l
total 43028
-rw-rw-r-- 1 root root     8514 Feb 13 10:22 add.go
...
```

### DIFF
--- a/cmd/buildah/unshare.go
+++ b/cmd/buildah/unshare.go
@@ -40,6 +40,8 @@ var (
 
 func init() {
 	unshareCommand.SetUsageTemplate(UsageTemplate())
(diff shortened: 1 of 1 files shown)


rhatdan:
@q384566678 I already opened a PR for this https://github.com/containers/buildah/pull/1375
Could you review that one and we can get it merged.

PR #1002, opened 2018-09-11
Add flags/arguments order verification to other commands

boaz1337:
This is the third part of adding flags/args order verficiation.
In this part I am adding it to command lines that take no arguments or more than 1 arguments such as buildah-images, buildah-copy and etc.

I added some tests too to verify that it does what it's supposed to do.

- [x] add
- [x] copy
- [x] images
- [x] mount
- [x] push
- [x] rm
- [x] rmi
- [x] umount

PR #1379, opened 2019-03-01
Allow buildah bud options and arguments to be interspersed.

rhatdan:
No reason to not allow

buildah bud . -f Dockerfile
versus
buidlah bud -f Dockerfile .

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -56,7 +56,7 @@ func init() {
 	budCommand.SetUsageTemplate(UsageTemplate())
 
 	flags := budCommand.Flags()
-	flags.SetInterspersed(false)
+	flags.SetInterspersed(true)
 
 	// BUD is a all common flags
 	budFlags := buildahcli.GetBudFlags(&budFlagResults)


TomSweeneyRedHat:
I'm leery about changing this back.   It was set up this way at one point and we went through a cycle of issues to set it up the way it is now with the ordering.  

Regardless, this change is blowing up the test, if we are going to keep it, those will need to be adjusted.

rhatdan:
Yes I see the tests. The question I have was did we make the change because of problems with urfave?

TomSweeneyRedHat:
That's a valid question @rhatdan, it was in at least part to do with urfave/cli.  I think #940 was the starting issue and @boaz0 fixed a bit with #968 and then more with #984 and #1002.  I thought there was another one or two, but I'm not seeing it.

I personally like enforcing that stuff is in the expected order, but not enough to not LGTM it if you think it's better and Cobra can handle it.

rhatdan:
I think we can just let it lie for now.  We know how to fix lets see if real customers complain.

PR #1380, opened 2019-03-03
Add --all-tags to push

TomSweeneyRedHat:

Adds the --all-tags to the push command to emulate the one in
the pull command.   Addresses https://github.com/containers/libpod/issues/2369
on the Buildah side, once approved similar changes will be made to Podman.

### DIFF
--- a/cmd/buildah/push.go
+++ b/cmd/buildah/push.go
@@ -13,6 +13,9 @@ import (
 	"github.com/containers/image/manifest"
 	"github.com/containers/image/transports"
 	"github.com/containers/image/transports/alltransports"
+	"github.com/containers/storage"
+	"github.com/containers/storage/pkg/archive"
+	multierror "github.com/hashicorp/go-multierror"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -20,6 +23,7 @@ import (
 )
 
 type pushResults struct {
+	allTags            bool
 	authfile           string
 	blobCache          string
 	certDir            string
@@ -61,6 +65,7 @@ func init() {
 
 	flags := pushCommand.Flags()
 	flags.SetInterspersed(false)
+	flags.BoolVarP(&opts.allTags, "all-tags", "a", false, "push all tagged images to the repository")
 	flags.StringVar(&opts.authfile, "authfile", "", "path of the authentication file. Default is ${XDG_RUNTIME_DIR}/containers/auth.json")
 	flags.StringVar(&opts.blobCache, "blob-cache", "", "assume image blobs in the specified directory will be available for pushing")
 	flags.StringVar(&opts.certDir, "cert-dir", "", "use certificates at the specified path to access the registry")
@@ -98,16 +103,47 @@ func pushCmd(c *cobra.Command, args []string, iopts pushResults) error {
 		return errors.New("Only two arguments are necessary to push: source and destination")
 	}
 
-	compress := imagebuildah.Gzip
-	if iopts.disableCompression {
-		compress = imagebuildah.Uncompressed
-	}
-
 	store, err := getStore(c)
 	if err != nil {
 		return err
 	}
 
+	// Map of image tags found for the specified image.
+	// tags stored in the key
+	var imageTags []string
+	if iopts.allTags {
+		if err := getAllTags(src, &imageTags, store); err != nil {
+			return err
+		}
+		logrus.Debugf("For the image: [%s] found tags: %v", src, imageTags)
+	}
+
+	var errs *multierror.Error
+	imageName := src
+	destTarget := destSpec
+	compress := imagebuildah.Gzip
+	if iopts.disableCompression {
+		compress = imagebuildah.Uncompressed
+	}
+
+	if len(imageTags) > 1 {
+		for _, tagKey := range imageTags {
+			src = imageName + ":" + tagKey
+			destSpec = destTarget + ":" + tagKey
+			if err = pushImage(c, src, destSpec, store, compress, iopts); err != nil {
+				errs = multierror.Append(errs, err)
+			}
+		}
+	} else {
+		if err = pushImage(c, src, destSpec, store, compress, iopts); err != nil {
+			errs = multierror.Append(errs, err)
+		}
+	}
+
+	return errs.ErrorOrNil()
+}
+
+func pushImage(c *cobra.Command, src string, destSpec string, store storage.Store, compress archive.Compression, iopts pushResults) error {
 	dest, err := alltransports.ParseImageName(destSpec)
 	// add the docker:// transport to see if they neglected it.
 	if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/buildah/push.go
```diff
@@ -98,16 +102,46 @@ func pushCmd(c *cobra.Command, args []string, iopts pushResults) error {
...
+
+	var errs *multierror.Error
+	var pushErr error
+	imageName := src
+	destTarget := destSpec
+	if len(imageTags) > 1 {
+		for tagKey := range imageTags {
+			src = imageName + ":" + tagKey
+			destSpec = destTarget + ":" + tagKey
+			pushErr = pushImage(c, src, destSpec, store, iopts)
```
rhatdan: Nit: why not just use err?

cmd/buildah/push.go
```diff
@@ -98,16 +102,46 @@ func pushCmd(c *cobra.Command, args []string, iopts pushResults) error {
...
+			if pushErr != nil {
```
rhatdan: Nit: Do this all on one line.

cmd/buildah/push.go
```diff
@@ -98,16 +102,46 @@ func pushCmd(c *cobra.Command, args []string, iopts pushResults) error {
...
+				errs = multierror.Append(errs, pushErr)
+			}
+		}
+	} else {
+		pushErr = pushImage(c, src, destSpec, store, iopts)
```
rhatdan: Do this on one line with if err = ...

cmd/buildah/push.go
```diff
@@ -179,3 +218,28 @@ func getListOfTransports() string {
...
+	for _, image := range images {
+
+		if len(image.Names) < 1 {
+			continue
+		}
+
+		for _, name := range image.Names {
+			splitName := strings.Split(name, ":")
+			if len(splitName) > 1 {
+				imageTags[splitName[1]] = ""
```
rhatdan: Why use a map rather then a list?

cmd/buildah/push.go
```diff
@@ -179,3 +218,28 @@ func getListOfTransports() string {
 	allTransports := strings.Join(transports.ListNames(), ",")
 	return strings.Replace(allTransports, ",tarball", "", 1)
 }
+
+// getAllTags gets all of the locally tagged images and returns them
+// in a slice.
+func getAllTags(image string, imageTags map[string]string, store storage.Store) error {
```
rhatdan: Just have this return a list of tags.

docs/buildah-push.md
```diff
@@ -45,6 +45,10 @@ If the transport part of DESTINATION is omitted, "docker://" is assumed.
 
 ## OPTIONS
 
+**--all-tags, a**
```
rhatdan: Should be a `-a`

cmd/buildah/push.go
TomSweeneyRedHat: I could.  I think when I was first working on this, err was already in use, but I restructured.

cmd/buildah/push.go
TomSweeneyRedHat: done

cmd/buildah/push.go
TomSweeneyRedHat: some day I'll keep this in my noggin.

docs/buildah-push.md
TomSweeneyRedHat: fixed, will touch up in pull too.  Thx for the eagle eye.

cmd/buildah/push.go
TomSweeneyRedHat: 'cause I love Maps?  Actually thought I'd use it as a quick lookup later, but you're right I ended up iterating so I'll pop it back to a slice.

TomSweeneyRedHat:
Repushed, @mheon FYI, planning to add to Podman once approved here.

mheon:
SGTM

cmd/buildah/push.go
```diff
@@ -148,6 +178,11 @@ func pushCmd(c *cobra.Command, args []string, iopts pushResults) error {
 		}
 	}
 
+	compress := imagebuildah.Gzip
+	if iopts.disableCompression {
+		compress = imagebuildah.Uncompressed
+	}
+
```
TomSweeneyRedHat: hmm, might be better to do this in the calling function and pass it down to pushImage....

TomSweeneyRedHat: moved.

cmd/buildah/push.go
```diff
@@ -179,3 +215,32 @@ func getListOfTransports() string {
...
+	}
+	for _, image := range images {
+
+		if len(image.Names) < 1 {
+			continue
+		}
+		// Name should look like: "docker.io/library/alpine:latest"
+		for _, name := range image.Names {
+			splitName := strings.Split(name, ":")
+			if !strings.HasSuffix(splitName[0], imageName) {
```
rhatdan: Is this enough?  What happens if I am looking for image frank, and I have an image named ballparkfrank?

rhatdan: I think @mtrmac might have a much better way of doing this.

TomSweeneyRedHat: If I put in the image name 'frank' it won't match up to 'ballparkfrank' as I force add a '/' to the imageName up at line 223.  So I think this is OK, but happy to rework it if there's a better way.

cmd/buildah/push.go
```diff
@@ -179,3 +215,32 @@ func getListOfTransports() string {
...
+				break
+			}
+			if len(splitName) > 1 {
+				*imageTags = append(*imageTags, splitName[1])
+			}
+		}
+
+	}
+	return nil
```
mtrmac: (I would find it a bit more natural for `imageTags` to be a return value instead of a reference parameter.)

cmd/buildah/push.go
```diff
@@ -98,16 +103,47 @@ func pushCmd(c *cobra.Command, args []string, iopts pushResults) error {
...
+	}
+
+	// Map of image tags found for the specified image.
+	// tags stored in the key
+	var imageTags []string
+	if iopts.allTags {
+		if err := getAllTags(src, &imageTags, store); err != nil {
+			return err
+		}
+		logrus.Debugf("For the image: [%s] found tags: %v", src, imageTags)
```
mtrmac: Fail with a nice error message if `len(imageTags) == 0`.

cmd/buildah/push.go
```diff
@@ -98,16 +103,47 @@ func pushCmd(c *cobra.Command, args []string, iopts pushResults) error {
...
 	}
 
+	store, err := getStore(c)
+	if err != nil {
+		return err
+	}
+
+	// Map of image tags found for the specified image.
+	// tags stored in the key
+	var imageTags []string
```
mtrmac: This is not a map.

docs/buildah-pull.md
```diff
@@ -45,9 +45,9 @@ The image ID of the image that was pulled.  On error 1 is returned.
 
 ## OPTIONS
 
-**--all-tags, a**
+**--all-tags, -a**
 
-All tagged images in the repository will be pulled.
+All tagged images in the repository will be pulled, not just `:latest`.
```
mtrmac: `:latest` is not the only tag one can pull, it’s only the default if the user specifies no tag and no digest.

docs/buildah-push.md
```diff
@@ -45,6 +45,10 @@ If the transport part of DESTINATION is omitted, "docker://" is assumed.
 
 ## OPTIONS
 
+**--all-tags, -a**
+
+All tagged images in the repository will be pushed, not just `:latest`.  A tag can not be included in the imageID or DESTINATION fields.
```
mtrmac: `:latest` is not the only tag one can pull, it’s only the default if the user specifies no tag and no digest.

cmd/buildah/push.go
```diff
@@ -98,16 +103,47 @@ func pushCmd(c *cobra.Command, args []string, iopts pushResults) error {
...
 		compress = imagebuildah.Uncompressed
 	}
 
-	store, err := getStore(c)
-	if err != nil {
-		return err
+	if len(imageTags) > 1 {
+		for _, tagKey := range imageTags {
+			src = imageName + ":" + tagKey
+			destSpec = destTarget + ":" + tagKey
```
mtrmac: (\*shrug\* This looks fragile, but if `dest` can be an arbitrary destination, it’s a practical compromise I guess.)

cmd/buildah/push.go
mtrmac: (BTW `FindImage` just loops over the result of `ResolveNames`. That’s not obviously _correct_ for transport-qualified names, or for storage IDs, but a similar loop here would at least be consistent with that code.)

rhatdan:
@TomSweeneyRedHat You still working on this?

TomSweeneyRedHat:
I've not been since @mtrmac's comments.  He brought up a lot of good points and I'm not sure if we should go forward with this or not.  I've been mulling it over, but paying attention to other things.

rhatdan:
@TomSweeneyRedHat @mtrmac Should we close this?

TomSweeneyRedHat:
I think yes.  @mtrmac comments pointed out a number of issues that I'd not considered.  I'll close and will let him reopen if he disagrees.

PR #1392, opened 2019-03-07
imagebuildah resolveSymlink: fix handling of relative links

nalind:
We don't need to put in as much effort as `getSymbolicLink()` does when we're already in a chroot, so switch back to using `filepath.EvalSymlinks()` to evaluate the path.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -293,7 +293,7 @@ func (b *Executor) Preserve(path string) error {
 
 	// Try and resolve the symlink (if one exists)
 	// Set archivedPath and path based on whether a symlink is found or not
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM

rhatdan:
@nalind Tests aren't happy.

nalind:
Yeah, that routine included some handling of corner cases that we don't get for free from `filepath.EvalSymlinks()`.  Going to have to restore part of it.

nalind:
This should fix #1393.

rhatdan:
@TomSweeneyRedHat @umohnani8 @giuseppe @vrothberg PTAL

vrothberg:
LGTM

vrothberg:
@rh-atomic-bot r+

PR #1400, opened 2019-03-11
Stop vendoring libpod into buildah

rhatdan:
These patches remove libpod from buildah and mv some pkg/rootless and pkg/chrootuser from libpod to buildah repo.


giuseppe:
it looks all we need in Buildah from rootless is `rootless.IsRootless()` and `rootless.GetRootlessUID()`.

Let's just duplicate these two functions instead of moving the package.  It will be a lot more difficult to fix rootless stuff in Podman if the implementation lives in another repository.

TomSweeneyRedHat:
Are the chroot and rootless pkgs also in libpod still?  If so could/should we make a new project in containers and vendor them from there?  Maybe a jira card for later, it would be nice to untangle this sooner.
Otherwise, LGTM

rhatdan:
bot, retest this please

giuseppe:
the error seems related to files that are still under `vendor/` but not present in `vendor.conf`

rhatdan:
@giuseppe Yes I need to get this vendored into containers/image first.

rhatdan:
Completes: https://github.com/containers/buildah/issues/1437

vrothberg:
We need this PR merged to make others pass again (that vendor recent storage).
LGTM, @giuseppe PTAL.

TomSweeneyRedHat:
LGTM, but would like a head nod from @giuseppe to make sure his rootless concerns are addressed.

vrothberg:
On a general note.

We should not merge PRs over at containers/storage (or any other place) when breaking the API while the follow-up ones on our consumers are not yet ready. In this specific case we cannot update containers/storage in libpod (and soon in CRI-O) because they depend on buildah which is not yet updated to the API breakages in containers/storage. This does not only block work (in this case me) but also blocks any fixes we might need to perform that require quick vendoring.

I think it's more the exception than the rule but I was blocked fairly often by such cases in the past that I want to share my troubles. Especially API breakages should be treated with care and those responsible should do the vendoring into all the consumers.

PR #1402, opened 2019-03-12
buildah: drop dependency on libpod rootless pkg

giuseppe:


giuseppe:
alternative version for: https://github.com/containers/buildah/pull/1400

giuseppe:
nevermind, I didn't see that #1400 is already copying the two functions instead of moving the package, so closing this PR

PR #1406, opened 2019-03-13
imagebuildah.ReposToMap(): move to cmd/buildah, which is its only consumer

nalind:
The imagebuildah.ReposToMap() function is only called from the CLI, so move it there.  This makes it easier to spot a duplicate `<none>:<none>` being added to the list that `buildah images` outputs, so fix that.

### DIFF
--- a/cmd/buildah/images.go
+++ b/cmd/buildah/images.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM
@rh-atomic-bot r+

PR #1408, opened 2019-03-13
bud.bats: use the per-test temp directory

nalind:
Use the per-test temporary directory for tests where we previously wrote to the source tree and later tried to clean up after ourselves, so that we don't leave files around if the tests fail.


TomSweeneyRedHat:
LGTM assuming happy tests.

rhatdan:
@rh-atomic-bot r+

PR #1409, opened 2019-03-13
Build: resolve copyFrom references earlier

nalind:
Resolve a possible copyFrom argument to a location before we possibly call into the imagebuilder `Run()` method that might call back to our `Copy()` method.


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM
@rh-atomic-bot r+

PR #1413, opened 2019-03-18
escape shell variables in README example

praiskup:


TomSweeneyRedHat:
bot, add author to whitelist

TomSweeneyRedHat:
@praiskup Thanks for the PR.  You need to sign you pr, if you could please.  `git commit --amend -s` and then repush it.

praiskup:
Done, PTAL.

TomSweeneyRedHat:
The changes look fine, but both the before and after works fine for me on Fedora.  Can you tell me what triggered the change?

Thanks for the signature and repush, git validation looks happier.

rhatdan:
Well $(command) is the preferred way to handle subshells, so I like that and wrapping of $ctr1 in "" allows contianers with spaces, although I am not sure that is supported.

praiskup:
On Monday, March 18, 2019 1:55:07 PM CET Tom Sweeney wrote:
> Can you tell me what triggered the change?

When I saw the command "ctr1=`buildah from ${1:-fedora}`".  The variable
"$1" might contain white spaces, which would expand as multiple arguments.

I then fixed all variables in the script to be consistent;  because $ctr
can hardly contain whitespaces (unless there's a bug, which is possible as
well).

TomSweeneyRedHat:
OK @praiskup , you sold me, thanks for the the follow up and the PR.
LGTM.

TomSweeneyRedHat:
@rh-atomic-bot r+

PR #1419, opened 2019-03-19
buildah-bud.md: correct a typo, note a default

nalind:
Note the default for the `--force-rm` flag in the man page.

### DIFF
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -199,7 +199,7 @@ If you specify `-f -`, the Dockerfile contents will be read from stdin.
 
 **--force-rm** *bool-value*
 
-Always remove intermediate containers after a build, even if the build is unsuccessful..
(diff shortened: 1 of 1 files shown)


docs/buildah-bud.md
```diff
@@ -199,7 +199,7 @@ If you specify `-f -`, the Dockerfile contents will be read from stdin.
 
 **--force-rm** *bool-value*
 
-Always remove intermediate containers after a build, even if the build is unsuccessful..
+Always remove intermediate containers after a build, even if the build fails (default false).
```
TomSweeneyRedHat: we'll need a similar change in podman-build.1.md sadly.
I think the new Cobra style is to NOT put a default  in the verbiage, but let cobra do so.  Cobra only does that on defaults set to true.

Am I on target @baude ?

rhatdan:
LGTM
@rh-atomic-bot r+

PR #1420, opened 2019-03-19
imagebuilder: split Executor and per-stage execution

nalind:
When building images using multiple stages, we were using a shallow copy of the `Executor`, depending on some fields to be shared between all instances, while replacing others fields with per-stage values.

This patch makes that easier for me to reason about by introducing a `StageExecutor` which holds the stage-specific information, and which can hold a reference to the main `Executor`, which now coordinates stages using `StageExecutors` rather than `Executors`.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -171,9 +171,10 @@ type BuildOptions struct {
 }
 
 // Executor is a buildah-based implementation of the imagebuilder.Executor
-// interface.
+// interface.  It coordinates the entire build by using one StageExecutors to
+// handle each stage of the build.
 type Executor struct {
-	named                          map[string]*Executor
+	stages                         map[string]*StageExecutor
 	store                          storage.Store
 	contextDir                     string
 	pullPolicy                     buildah.PullPolicy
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@vrothberg @giuseppe @TomSweeneyRedHat @umohnani8 PTAL

umohnani8:
LGTM

TomSweeneyRedHat:
build issue appears to be a timeout flake.

TomSweeneyRedHat:
bot, retest this please

TomSweeneyRedHat:
passes baseline and the other tests I threw at it.  LGTM, but would like a head nod from @vrothberg in the morning.

vrothberg:
I restarted the two failed tests, so we can merge the PR once everything's green.

rhatdan:
@rh-atomic-bot r+

nalind:
@rh-atomic-bot retry

nalind:
Rebased.

nalind:
Removed logic to reset the stage's volume cache each time we commit a layer in a stage - the cached contents of parts of the working container's rootfs don't need to be regenerated just because we committed a layer and created a new container from the image, since we know that that portion of the new container;s directory tree is supposed to have the same contents as the old one.

imagebuildah/build.go
```diff
@@ -222,9 +214,18 @@ type Executor struct {
...
+
+	index           int
+	name            string
+	builder         *buildah.Builder
+	preserved       int
+	volumes         imagebuilder.VolumeSet
+	volumeCache     map[string]string
+	volumeCacheInfo map[string]os.FileInfo
+	mountPoint      string
+	copyFrom        string // Used to keep track of the --from flag from COPY and ADD
```
rhatdan: Why are you doing this?

nalind: We were making shallow copies of this structure in `withName()` and then depending on some of the fields to remain shared between them, and some to not be.  It was getting difficult to debug.

rhatdan: Can you add a comment about that in the place of the break to explain it, so someone like me does not come back in and merge them back together.

nalind: Expanded on the comment there.

rhatdan:
@rh-atomic-bot r+

PR #1423, opened 2019-03-20
WIP - Switch to golanci-lint

TomSweeneyRedHat:

Attempt to switch to golangci-lint which is replacing
gometalinter, which is getting deprecated soon.

Stealing heavily from https://github.com/kubernetes-sigs/cri-o/pull/2137
by @saschagrunert


Makefile
```diff
@@ -70,15 +74,14 @@ endif
 	@./tests/validate/whitespace.sh
 	@./tests/validate/govet.sh
 	@./tests/validate/git-validation.sh
-	@./tests/validate/gometalinter.sh . cmd/buildah
+	lint
```
TomSweeneyRedHat: I think I'll probably need to make an equivalent gometalinter.sh for the new lint.  Will work on this more tomorrow.

rhatdan:
@TomSweeneyRedHat Any movement on this?

TomSweeneyRedHat:
@vrothberg FYI, if I get out from under BZ's while you're gone, I'll try to dive back into this.

rhatdan:
Perhaps @saschagrunert Would like to help with this.  HintHint...

vrothberg:
> Perhaps @saschagrunert Would like to help with this. HintHint...

Feel free to cherry-pick from https://github.com/containers/buildah/pull/1624.

.golanc.yml
```diff
@@ -0,0 +1,38 @@
+---
```
saschagrunert: The file has the wrong name and should be `.golangci.yml`

Makefile
```diff
@@ -50,6 +50,10 @@ clean:
 docs: ## build the docs on the host
 	$(MAKE) -C docs
 
+.PHONY: lint
+lint: ## do lint processing 
+	golangci-lint run --build-tags="$(BUILDTAGS)"
```
saschagrunert: I would hard-code the buildtags to produce the same output on every developer machine, like here: 
https://github.com/cri-o/cri-o/blob/97c3b029ad69d7347520e6efc932ac27c8659c33/.golangci.yml#L3-L8

Makefile
saschagrunert: I guess a single invocation of `golangci-lint run` could be enough here.

saschagrunert:
If we rename the file, then I only see some issues left on my machine:
```
> make lint
golangci-lint run --build-tags="seccomp selinux    containers_image_ostree_stub"
bind/util.go:29:6: `dedupeStringSlice` is unused (deadcode)
func dedupeStringSlice(slice []string) []string {
     ^
chroot/run_test.go:107:6: `testNoop` is unused (deadcode)
func testNoop(t *testing.T) {
     ^
chroot/run_test.go:114:6: `testMinimalSkeleton` is unused (deadcode)
func testMinimalSkeleton(t *testing.T) {
     ^
chroot/util.go:5:6: `dedupeStringSlice` is unused (deadcode)
func dedupeStringSlice(slice []string) []string {
     ^
make: *** [Makefile:55: lint] Error 1
```

We could also exclude `deadcode` for now and enable the linters later on including the fixes.

rhatdan:
SGTM

Makefile
TomSweeneyRedHat: I'd like to get @nalind's opinion on this.  He set the calls to a variety of shell scripts to get the value for the platform.  I'm not sure if we can hardcode it without messing up a particular variant on linux.

saschagrunert: That's an argument, so I guess we can also stick to the current approach.

TomSweeneyRedHat:
Replaced by #1686

PR #1426, opened 2019-03-20
golint: make golint happy

dougsland:
should omit type ... from declaration of var ...; it will be
inferred from the right-hand side

Signed-off-by: Douglas Schilling Landgraf <dougsland@redhat.com>


rhatdan:
LGTM

nalind:
LGTM
@rh-atomic-bot r+

PR #1427, opened 2019-03-21
imagebuildah: check for unused args across stages

nalind:
Check for unused arguments across all of the stages of a multi-stage build.  They can only be passed to us once for the whole build, so we shouldn't treat it as an error if a given stage doesn't reference all of
them.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -11,6 +11,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -216,6 +217,7 @@ type Executor struct {
 	imageMap                       map[string]string // Used to map images that we create to handle the AS construct.
 	blobDirectory                  string
 	excludes                       []string
+	unusedArgs                     map[string]struct{}
 }
 
 // StageExecutor bundles up what we need to know when executing one stage of a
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

rhatdan:

TomSweeneyRedHat:
LGTM

imagebuildah/build.go
```diff
@@ -1348,12 +1351,20 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (strin
 			b.imageMap[stage.Name] = imgID
 		}
 	}
+	if len(b.unusedArgs) > 0 {
+		unusedList := make([]string, 0, len(b.unusedArgs))
+		for k := range b.unusedArgs {
+			unusedList = append(unusedList, k)
+		}
+		fmt.Fprintf(b.out, "[Warning] one or more build args %v were not consumed\n", unusedList)
```
vrothberg: Nit of a nit: can we move printing the list to the end of the sentence?

"[Warning] one or more build args were not consumed: %v\n"

I think this is a bit easier to read for longer lists.

nalind: Okay, the list should be at the end of the line now.  We populate it using the keys from the map, so their order wasn't well-defined, so we sort it now, too.

nalind:
Rebased and added an integration test.

rhatdan:
@rh-atomic-bot r+

PR #1428, opened 2019-03-21
imagebuildah: drop the duplicate containerIDs list

nalind:
`StageExecutor`s keep their own lists of the IDs of containers that they create, so we don't need to merge them into a single list in the `Executor` that we'll then consult during clean up.  As long as we make sure we only ever try to delete a given stage's containers once, we can just use their lists directly, and avoid duplicating the information.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -213,7 +213,6 @@ type Executor struct {
 	noCache                        bool
 	removeIntermediateCtrs         bool
 	forceRmIntermediateCtrs        bool
-	containerIDs                   []string          // Stores the IDs of the successful intermediate containers used during layer build
 	imageMap                       map[string]string // Used to map images that we create to handle the AS construct.
 	blobDirectory                  string
 	excludes                       []string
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

rhatdan:

imagebuildah/build.go
```diff
@@ -1569,15 +1566,23 @@ func processCopyFrom(dockerfiles []io.ReadCloser) []io.ReadCloser {
...
+			if err := b.store.DeleteContainer(ctr); err != nil {
+				logrus.Errorf("error deleting build container %q: %v\n", ctr, err)
+				lastErr = err
+			}
 		}
+		// The stages map includes some stages under multiple keys, so
+		// clearing their lists after we process a given stage is
+		// necessary to avoid triggering errors that would occur if we
+		// tried to delete a given stage's containers multiple times.
+		s.containerIDs = nil
```
vrothberg: Doing this would also yield panics and reveal bugs in the code, which I think is a really good thing for such cases.

nalind:
@rh-atomic-bot r=vrothberg

PR #1429, opened 2019-03-21
imagebuildah.resolveNameToImageRef(): take name as a parameter

nalind:
Instead of having the resolveNameToImageRef() method resolve a member value in the object that we set before calling the method, have it resolve a value that we pass directly to it.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -806,24 +806,24 @@ func (s *StageExecutor) Delete() (err error) {
 }
 
 // resolveNameToImageRef creates a types.ImageReference from b.output
-func (b *Executor) resolveNameToImageRef() (types.ImageReference, error) {
+func (b *Executor) resolveNameToImageRef(output string) (types.ImageReference, error) {
 	var (
 		imageRef types.ImageReference
 		err      error
 	)
-	if b.output != "" {
-		imageRef, err = alltransports.ParseImageName(b.output)
+	if output != "" {
+		imageRef, err = alltransports.ParseImageName(output)
 		if err != nil {
-			candidates, _, _, err := util.ResolveName(b.output, "", b.systemContext, b.store)
+			candidates, _, _, err := util.ResolveName(output, "", b.systemContext, b.store)
 			if err != nil {
-				return nil, errors.Wrapf(err, "error parsing target image name %q", b.output)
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

rhatdan:

imagebuildah/build.go
```diff
@@ -1358,7 +1362,7 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (strin
 	ignoreLayers := singleLineDockerfile || !b.layers && !b.noCache
 
 	if ignoreLayers {
-		imgID, ref, err := stageExecutor.Commit(ctx, stages[len(stages)-1].Builder, "")
+		imgID, ref, err := stageExecutor.Commit(ctx, stages[len(stages)-1].Builder, "", b.output)
```
TomSweeneyRedHat: should this and line 1348 just be 'output' and not 'b.output'?

TomSweeneyRedHat:
One questions, otherwise LGTM

imagebuildah/build.go
nalind: This one should be `b.output`, since it's committing the final image, but you're right that 1348 should be `output` rather than `b.output`, since except when it's the final stage, we don't want to be assigning the final image's name to the image that's committed there (though that should only happen when the final stage has an "AS" attached to it, which in any case would be weird).  Nice catch!  Fixed.

imagebuildah/build.go
```diff
@@ -899,11 +899,15 @@ func (s *StageExecutor) Execute(ctx context.Context, stage imagebuilder.Stage) e
...
-			s.output = ""
+			commitName = ""
 		} else {
-			s.output = commitName
+			commitName = s.output
 		}
 
+		// TODO: this makes the tests happy, but it shouldn't be
+		// necessary unless this is the final stage.
+		commitName = s.executor.output
```
rhatdan: This overwrites the command just above?

nalind: Yes.  For the moment, not doing this causes tests to stop passing.
There's a bigger change to the execute loop coming up that'll require updating the tests, and this will be resolved as part of that, but this PR is intended to be a smaller change that doesn't impact tests, so in the interests of that, for now I'm just marking it as a to-do item.

rhatdan:
@rh-atomic-bot r+

Issue #1439, opened 2019-03-25
Image build failed on COPY, VOLUME, WORKDIR instruction use symlinks


Lagrang:
(Edited just to make the buildah example be happy in md)
**Description**

`buildah bud` failed on Dockerfile with WORKDIR instruction: ensuring container path "/usr/local/hbase/bin": mkdir /var/lib/containers/storage/btrfs/subvolumes/8325b8065650778d4abbb211b130511525e668b8b45e0df0f1c3563646c4c674/usr/local/hbase: file exists

I find out, that root cause is symlinks: COPY, VOLUME, WORKDIR instructions cannot handle it.
COPY not support symlink as target and fails with error:
`
COPY --chown=hbase:hadoop conf/hbase-env.sh      conf/log4j.properties      conf/hadoop-metrics2-hbase.properties      conf/hbase-policy.xml      $HBASE_CONF_DIR/}: error creating directory "/var/lib/containers/storage/btrfs/subvolumes/8b473a99a4b0bcc89d0b3a70a25414431dd5063a0987c084c134efcd78be56cb/etc/hbase": chown /var/lib/containers/storage/btrfs/subvolumes/8b473a99a4b0bcc89d0b3a70a25414431dd5063a0987c084c134efcd78be56cb/etc/hbase: no such file or directory
`

VOLUME which point to symlink overwrite target folder permissions to root and cause folder inaccessible for container.

**Steps to reproduce the issue:**
1. Dockerfile from this [repo](https://gitlab.com/hutils/containers/tree/09121d3b7f85e6033b7354c903d209cd815d96d3/hbase/2.x)
2. Run 
```
    sudo buildah bud --no-cache \
             --build-arg HBASE_VERSION=2.1.2 \
             --build-arg HADOOP_VERSION=2.8.5 \
             -t hbase .
```
**Describe the results you received:**
`Command:workdir Args:[/usr/local/hbase/bin] Flags:[] Attrs:map[] Message:WORKDIR $BASE_INSTALL_PATH/hbase/bin Original:WORKDIR $BASE_INSTALL_PATH/hbase/bin}: error ensuring container path "/usr/local/hbase/bin": mkdir /var/lib/containers/storage/btrfs/subvolumes/8325b8065650778d4abbb211b130511525e668b8b45e0df0f1c3563646c4c674/usr/local/hbase: file exists`

**Output of `rpm -q buildah` or `apt list buildah`:**

```
[...]

PR #1443, opened 2019-03-25
Resolve symlink when checking container path

TomSweeneyRedHat:
When checking the container path, check to see if the path is a symlink and convert it if necessary
before writing the file.

Fixes: #1439

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -483,6 +483,19 @@ func (s *StageExecutor) volumeCacheRestore() error {
 // imagebuilder tells us the instruction was "ADD" and not "COPY".
 func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) error {
 	for _, copy := range copies {
+		// If the file exists, check to see if it's a symlink.
+		// If it is a symlink, convert to it's target otherwise
+		// the symlink will be overwritten.
+		fileDest, _ := os.Lstat(filepath.Join(s.mountPoint, copy.Dest))
+		if fileDest != nil {
+			if fileDest.Mode()&os.ModeSymlink != 0 {
+				if symLink, err := resolveSymlink(s.mountPoint, copy.Dest); err == nil {
+					copy.Dest = symLink
+				} else {
+					return errors.Wrapf(err, "error reading symbolic link to %q", copy.Dest)
+				}
+			}
+		}
 		if copy.Download {
 			logrus.Debugf("ADD %#v, %#v", excludes, copy)
 		} else {
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -1140,3 +1140,19 @@ load helpers
   echo "$output"
   test "${#lines[@]}" -eq 0
 }
+
+@test "bud copy to symlink" {
+  target=alpine-image
+  ctr=alpine-ctr
+  run_buildah --debug=false bud --signature-policy ${TESTSDIR}/policy.json -t ${target} ${TESTSDIR}/bud/dest-symlink
+  expect_output --substring "STEP 5: RUN ln -s "
+
+  run_buildah --debug=false from --signature-policy ${TESTSDIR}/policy.json --name=${ctr} ${target}
+  expect_output --substring ${ctr}
+
+  run_buildah --debug=false run ${ctr} ls -alF /etc/hbase
+  expect_output --substring "/etc/hbase -> /usr/local/hbase/"
+
+  run_buildah --debug=false run ${ctr} ls -alF /usr/local/hbase 
+  expect_output --substring "Dockerfile"
+}
--- a/tests/bud/dest-symlink/Dockerfile
+++ b/tests/bud/dest-symlink/Dockerfile
@@ -1,0 +1,9 @@
+FROM alpine
+
+ENV HBASE_HOME="/usr/local/hbase"
+ENV HBASE_CONF_DIR="/etc/hbase"
+
+RUN mkdir $HBASE_HOME
+RUN ln -s $HBASE_HOME $HBASE_CONF_DIR
+
+COPY Dockerfile $HBASE_CONF_DIR


TomSweeneyRedHat:
@Lagrang I believe this will fix your #1439.  I tested with this patch however I filled up my disk space using your test example.  I need to build a bigger vm and retest.  My test got well past your failure point, so I'm relatively positive this fixes the issue you reported.

If by chance you could test and verify on your end, that would be awesome.

Lagrang:
@TomSweeneyRedHat I'll need some time to check this fix. I'll do my best to check it in few days.

imagebuildah/build.go
```diff
@@ -1613,6 +1613,16 @@ func (s *StageExecutor) EnsureContainerPath(path string) error {
 	if err != nil && os.IsNotExist(err) {
 		err = os.MkdirAll(filepath.Join(s.mountPoint, path), 0755)
 	}
+	// Let's see if we're dealing with a symlink
+	if err != nil {
+		var sympath string
+		sympath, err = filepath.EvalSymlinks(path)
```
rhatdan: Should this be on path or s.mountPoint+path?
Should you do this before you do the MkDirALl above, so we only have one call to MkdirALL?

Lagrang:
I built branch with this patch, but it seems that it not fix an error:(
I created more simple Dockerfile to reproduce errors, with extensive comments which I hope helps you:
```
FROM ubuntu:18.10

ENV HBASE_VERSION="2.1.3"
ENV HBASE_HOME="/usr/local/hbase-$HBASE_VERSION"
ENV HBASE_CONF_DIR="/etc/hbase"

RUN groupadd --gid 15000 hadoop \
    && useradd --uid 15100 -m -G hadoop hbase \
    && mkdir $HBASE_HOME $HBASE_HOME/conf \
    && touch $HBASE_HOME/conf/file.conf \	
    && ln -s $HBASE_HOME/conf $HBASE_CONF_DIR \
    && ln -s $HBASE_HOME /hbase \
    && chown -R hbase:hadoop $HBASE_HOME \
    && chmod -R 755 $HBASE_HOME

# this copy works fine
COPY --chown=hbase:hadoop https://gitlab.com/hutils/containers/raw/master/hbase/2.x/conf/hbase-env.sh?inline=false $HBASE_HOME/conf/

# this copy fails because of symlink as target dir
# error message is: 'error creating "/var/lib/containers/storage/overlay/bf01b87ae6a574100bcdbba997b5b2570fb1f7c655dbd8935b5b8e1c7bb80d31/merged/etc/hbase": open /var/lib/containers/storage/overlay/bf01b87ae6a574100bcdbba997b5b2570fb1f7c655dbd8935b5b8e1c7bb80d31/merged/etc/hbase: no such file or directory'
#COPY --chown=hbase:hadoop https://gitlab.com/hutils/containers/raw/master/hbase/2.x/conf/hbase-env.sh?inline=false $HBASE_CONF_DIR

# this volume works fine and show files 'conf/'
VOLUME $HBASE_HOME/conf

# this volume(based on symlink) not show files under 'conf/' dir
#VOLUME $HBASE_CONF_DIR

# this workdir works fine
WORKDIR /usr/local/hbase-$HBASE_VERSION

# this workdir broken, now it returns 'error evaluating "/hbase/bin" for symbolic links when ensuring container path: lstat /hbase: no such file or directory'
#WORKDIR /hbase/bin

USER hbase:hadoop
ENTRYPOINT ["/bin/bash"]
CMD ["ls -l /hbase/conf"]
```

imagebuildah/build.go
TomSweeneyRedHat: Oh shoot, I think your right about joining them.  I've reworked the logic a bit, only one mkdireall now.  I'll repush later.

TomSweeneyRedHat:
@Lagrang I'm not able to completely run your Dockerfile as it fails in step 24 looking for /usr/local/hbase-2.1.2/bin which I don't have. so if you could test/verify, that would be great.

TomSweeneyRedHat:
oops, off by two, it's failing in step # 26

Lagrang:
@TomSweeneyRedHat Do you use Dockerfile from my comment above? It doesn't contain path /usr/local/hbase-2.1.2...
If not, please try it. This Dockerfile simpler and can reproduce all type of errors, if you uncomment some instructions.

imagebuildah/build.go
```diff
@@ -1629,12 +1629,17 @@ func (b *Executor) deleteSuccessfulIntermediateCtrs() error {
 }
 
 func (s *StageExecutor) EnsureContainerPath(path string) error {
-	_, err := os.Stat(filepath.Join(s.mountPoint, path))
+	joinedPath, err := filepath.EvalSymlinks(filepath.Join(s.mountPoint, path))
```
giuseppe: don't we need to ensure `joinedPath` is still under s.mountPoint?

TomSweeneyRedHat:
@Lagrang I was playing with coming up with a small test example and found that Docker doesn't support symlink traversal.  Does this Dockerfile work with Docker.  When I tried traversing a symlink there with this Dockerfile:

```
# cat ~/Dockerfile.followsym
FROM alpine
COPY /mydir /mydir

# ls -alF ./mydir
lrwxrwxrwx. 1 root root 19 Mar 30 16:36 ./mydir -> /root/tomsym/mydir//
# ls mydir/*
mydir/myfakefile.txt
```

Then when I run it

```
# docker build -t tom -f ~/Dockerfile.followsym .
Sending build context to Docker daemon   3.54GB
Step 1/2 : FROM alpine
 ---> 5cb3aa00f899
Step 2/2 : COPY /mydir /mydir
COPY failed: stat /var/lib/docker/tmp/docker-builder349996372/root/tomsym/mydir: no such file or directory

# docker build -t tom -f ~/Dockerfile.followsym .
Sending build context to Docker daemon   3.54GB
Step 1/2 : FROM alpine
 ---> 5cb3aa00f899
Step 2/2 : COPY /mydir /mydir
COPY failed: stat /var/lib/docker/tmp/docker-builder349996372/root/tomsym/mydir: no such file or directory
```
I also dug this up in Docker's GitHub:  https://github.com/moby/moby/issues/1676  which seems to indicate they don't support symlink traversal due to potential security issues and not knowing where a symlink will live from host to host making a Dockerfile a little wonky.

@nalind, @rhatdan thoughts?

Lagrang:
As I understand, you try use symlink as source of COPY. It's reasonable to not support symlinks as source. But I faced issue when symlink is a target of COPY.

TomSweeneyRedHat:
@Lagrang thanks for the clarification.  I'll poke further with that in mind.

TomSweeneyRedHat:
@Lagrang btb, have you had a chance to test with this patch yet?

Lagrang:
@TomSweeneyRedHat  I built branch with your last changes, but build fails with error: 
`STEP 14: WORKDIR /hbase/bin
error building at step {Env:[PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin HBASE_VERSION=2.1.3 HBASE_HOME=/usr/local/hbase-2.1.3 HBASE_CONF_DIR=/etc/hbase] Command:workdir Args:[/hbase/bin] Flags:[] Attrs:map[] Message:WORKDIR /hbase/bin Original:WORKDIR /hbase/bin}: error evaluating container path "": lstat /usr/local/hbase-2.1.3/bin: no such file or directory
`
Same problem with COPY and VOLUME instructions.
I use this [Dockerfile](#issuecomment-476791911)

TomSweeneyRedHat:
@Lagrang darn but thanks for the test and report back.  Will take another look.

ddarrah:
I know I am late to this party, but do we want to provide a way for build to ignore simlinks for security purposes?

rhatdan:
I think as long as the symlinks are checked to be within the image, then we are ok.

EmilienM:
@TomSweeneyRedHat we hit this bug in OpenStack, some containers are failing to build. Do you have an ~ ETA for this PR to land? Thanks!

TomSweeneyRedHat:
@EmilienM I don't have an ETA on this one.  It's one of two that are on the top of the stack of my backlog.

TomSweeneyRedHat:
@Lagrang second time is the charm!  This one I think should fix your and @EmilienM 's issue.  If you've a chance to try it out, that would be great.

@umohnani8 @rhatdan @giuseppe PTAL

EmilienM:
@karelyatin if you have a chance to try before I do.

karelyatin:
@EmilienM @TomSweeneyRedHat i tried with the recent PR and i see similar Error as we have with buidah-1.7.1:-
[zuul@standalone ~]$ buildah --version
buildah version 1.7.1 (image-spec 1.0.0, runtime-spec 1.0.0)
[zuul@standalone ~]$ sudo buildah bud --tls-verify=False  -t 127.0.0.1:8787/tripleomaster/centos-binary-tempest:be90d93c3c5f77f428d12a9a8a2ef97b9dada8f3_5b53d5ba tempest
STEP 1: FROM docker.io/tripleomaster/centos-binary-openstack-base:be90d93c3c5f77f428d12a9a8a2ef97b9dada8f3_5b53d5ba
STEP 2: LABEL maintainer="Kolla Project (https://launchpad.net/kolla)" name="tempest" build-date="20190415"
STEP 3: RUN usermod --append --home /var/lib/tempest --groups kolla tempest     && mkdir -p /var/lib/tempest     && chown -R 42480:42480 /var/lib/tempest
STEP 4: RUN ln -sf /var/lib/tempest /tempest
STEP 5: WORKDIR /tempest
error building at step {Env:[PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin KOLLA_BASE_DISTRO=centos KOLLA_INSTALL_TYPE=binary KOLLA_INSTALL_METATYPE=rdo KOLLA_DISTRO_PYTHON_VERSION=2.7 PS1=$(tput bold)($(printenv KOLLA_SERVICE_NAME))$(tput sgr0)[$(id -un)@$(hostname -s) $(pwd)]$ ] Command:workdir Args:[/tempest] Flags:[] Attrs:map[] Message:WORKDIR /tempest Original:WORKDIR /tempest}: error ensuring container path "/tempest": mkdir /var/lib/containers/storage/overlay/609939041c25b5d318dcba013e5f8d26bfe11a040f579c16fb64958a5d9bb713/merged/tempest: file exists
[zuul@standalone ~]$ /usr/local/bin/buildah --version
buildah version 1.8-dev (image-spec 1.0.0, runtime-spec 1.0.0)
[zuul@standalone ~]$ sudo /usr/local/bin/buildah bud --tls-verify=False  -t 127.0.0.1:8787/tripleomaster/centos-binary-tempest:be90d93c3c5f77f428d12a9a8a2ef97b9dada8f3_5b53d5ba tempest
STEP 1: FROM docker.io/tripleomaster/centos-binary-openstack-base:be90d93c3c5f77f428d12a9a8a2ef97b9dada8f3_5b53d5ba
STEP 2: LABEL maintainer="Kolla Project (https://launchpad.net/kolla)" name="tempest" build-date="20190415"
STEP 3: RUN usermod --append --home /var/lib/tempest --groups kolla tempest     && mkdir -p /var/lib/tempest     && chown -R 42480:42480 /var/lib/tempest
STEP 4: RUN ln -sf /var/lib/tempest /tempest
STEP 5: WORKDIR /tempest
error building at STEP "WORKDIR /tempest": error ensuring container path "/tempest": mkdir /var/lib/containers/storage/overlay/87bbfbded4d6dd20ee1961a664c530b49051dec4429ad1dee055071862f94788/merged/tempest: file exists


[zuul@standalone ~]$ cat tempest/Dockerfile 
FROM docker.io/tripleomaster/centos-binary-openstack-base:be90d93c3c5f77f428d12a9a8a2ef97b9dada8f3_5b53d5ba
LABEL maintainer="Kolla Project (https://launchpad.net/kolla)" name="tempest" build-date="20190415"
RUN usermod --append --home /var/lib/tempest --groups kolla tempest \
    && mkdir -p /var/lib/tempest \
    && chown -R 42480:42480 /var/lib/tempest
RUN ln -sf /var/lib/tempest /tempest
WORKDIR /tempest
RUN chmod 755 /usr/local/bin/kolla_extend_start
USER tempest

TomSweeneyRedHat:
@EmilienM and @karelyatin thanks for the test of the patch and sorry it didn't  fix your issue.  However I think you've discovered a separate but related issue and I've just spun up #1537 to track that and I will hop on that one next.

TomSweeneyRedHat:
OK, I've tried dumbing down the test for the build (removed the STEP reference in the check), it's working in my sandbox, perhaps the numbering is different in the CI?

TomSweeneyRedHat:
OK, tests might work better if I include the required Dockerfile....

imagebuildah/build.go
```diff
@@ -482,6 +482,13 @@ func (s *StageExecutor) volumeCacheRestore() error {
 // imagebuilder tells us the instruction was "ADD" and not "COPY".
 func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) error {
 	for _, copy := range copies {
+		// If the destination is a symlink, convert it to a hardpath
+		// otherwise it will be overwritten.
+		if symLink, err := resolveSymlink(s.mountPoint, copy.Dest); err == nil {
+			copy.Dest = symLink
+		} else {
+			return errors.Wrapf(err, "error reading symbolic link to %q", copy.Dest)
+		}
```
TomSweeneyRedHat: This is having troubles with a clause like
 COPY Dockerfile1.scratch /Dockerfile1
Rather than copying Dockerfile1.scratch to the file /Dockerfile1, it's creating a directory named /Dockerfile1. Will dive further on Monday.

TomSweeneyRedHat: I think this now works when the dest is a symlink with this os.Stat check.

imagebuildah/build.go
```diff
@@ -483,6 +483,19 @@ func (s *StageExecutor) volumeCacheRestore() error {
 // imagebuilder tells us the instruction was "ADD" and not "COPY".
 func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) error {
 	for _, copy := range copies {
+		// If the file exists, check to see if it's a symlink.
+		// If it is a symlink, convert to it's target otherwise
+		// the symlink will be overwritten.
+		fileDest, _ := os.Stat(s.mountPoint + copy.Dest)
+		if fileDest != nil {
+			if !(fileDest.Mode()&os.ModeSymlink == os.ModeSymlink) {
```
edsantiago: Did you really mean `!` ?

imagebuildah/build.go
```diff
@@ -483,6 +483,19 @@ func (s *StageExecutor) volumeCacheRestore() error {
 // imagebuilder tells us the instruction was "ADD" and not "COPY".
 func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) error {
 	for _, copy := range copies {
+		// If the file exists, check to see if it's a symlink.
+		// If it is a symlink, convert to it's target otherwise
+		// the symlink will be overwritten.
+		fileDest, _ := os.Stat(s.mountPoint + copy.Dest)
```
edsantiago: You probably want `Lstat()`, not Stat(). But even if I change that on my end, I get FileMode `1000000777` which makes no sense. Deferring this again for a while, sending back your way. HTH.

tests/bud.bats
```diff
@@ -1140,3 +1140,19 @@ load helpers
   echo "$output"
   test "${#lines[@]}" -eq 0
 }
+
+@test "bud copy to symlink" {
+  target=alpine-image
+  ctr=alpine-ctr
+  run buildah --debug=false bud --signature-policy ${TESTSDIR}/policy.json -t ${target} ${TESTSDIR}/bud/dest-symlink
```
edsantiago: `run_buildah` (with underscore) gives you exit code checking, timeout, and much better logging in case of error.

edsantiago:
The following code produces the outcome I believe is desired:
```go
    fileDest, _ := os.Lstat(filepath.Join(s.mountPoint, copy.Dest))
    if fileDest != nil {
        if (fileDest.Mode()&os.ModeSymlink != 0) {
```
The key insight is that Go `os.ModeSymlink` bears no relation to `S_IFLNK`.  HTH.

tests/bud.bats
TomSweeneyRedHat: fixed

imagebuildah/build.go
TomSweeneyRedHat: yep, changed to Lstat.  Made the other changes too, thanks @edsantiago

imagebuildah/build.go
TomSweeneyRedHat: It seemed to work OK in my testing, but your suggestion is shorter and seems to work just fine, converted.

edsantiago: In UNIX-land it's important to mask then check for `S_IFLNK` because that has two bits set. In Go-land `os.ModeSymlink` is a single bit, so (assuming that never changes) it's safe to `!= 0`. I would welcome feedback from someone more Go-literate, though. (And the `!` definitely needed to go)

rhatdan:
LGTM, if this passes test we will merge.

@Lagrang Could you verify this works for you?

rhatdan:
@edsantiago PTAL

edsantiago:
I'm still uncomfortable about `path + path` instead of `filepath.Join()`, but perhaps there's a guarantee or restriction I'm missing such that the target will never be a relative path. Aside from that, new changes LGTM.

TomSweeneyRedHat:
@edsantiago I should have changed that earlier, just did now.

rhatdan:
@rh-atomic-bot r+

Lagrang:
@TomSweeneyRedHat @rhatdan sorry for late response!
Today I tried to build master branch. Now COPY and VOLUME works with symlinks. But I face an issue with WORKDIR again:
`STEP 9: WORKDIR /hbase/bin
error building at STEP "WORKDIR /hbase/bin": error ensuring container path "/hbase/bin": mkdir /var/lib/containers/storage/overlay/261c8863ece59a0f1c9b01828bb8e4a0370af7130331b1b994a58dc11a9b434d/merged/hbase: file exists`

TomSweeneyRedHat:
@Lagrang I suspect that #1558 will fix your WORKDIR issue too.  I know it's fixed two others.

PR #1446, opened 2019-03-26
imagebuildah.Build: consolidate cleanup logic

nalind:
Consolidate all of the logic related to cleaning up after a build into a helper function in `Build()`.  This lets us ensure that we always clean up images after cleaning up containers, so that we don't have to worry
about ordering when some images may be in use by containers because we wanted access to a root filesystem for a stage whose image was reused.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -951,7 +951,6 @@ func (s *StageExecutor) Execute(ctx context.Context, stage imagebuilder.Stage) (
 			if imgID, ref, err = s.copyExistingImage(ctx, cacheID, commitName); err != nil {
 				return "", nil, err
 			}
-			s.containerIDs = append(s.containerIDs, s.builder.ContainerID)
 			break
 		}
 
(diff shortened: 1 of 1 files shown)


imagebuildah/build.go
```diff
@@ -1353,33 +1390,27 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 		// Always remove the intermediate/build containers, even if the build was unsuccessful.
 		// If building with layers, remove all intermediate/build containers if b.forceRmIntermediateCtrs
 		// is true.
-		if b.forceRmIntermediateCtrs || (!b.layers && b.useCache) {
-			defer stageExecutor.Delete()
+		if b.forceRmIntermediateCtrs || !b.layers {
```
TomSweeneyRedHat: should the b.useCache still remain here?  If we don't have layers and don't have cache, will the cleanup be happy?

nalind: Before this change, we were letting `deleteSuccessfulIntermediateCtrs()` handle removal of the stage's final working container if the stage succeeded, by adding its ID to the `containerIDs` list, sometimes.  This clause handled one case where the removal of the final working container was desired but not handled that way.

This patch changes the logic so that `containerIDs` only ever lists the non-final containers for a stage, so that the decision about whether or not to remove a stage's final container is made in fewer places and controlled by fewer variables.

nalind: Argh, I didn't even answer your question.  If the stage ends up handing us an image that it found already existed, then the stage's final container will be based on that image, and if the build failed and we leave that container around for troubleshooting (which is our default behavior), we'll hit an error when we go to remove it.  We check for that specific error (`storage.ErrImageUsedByContainer`) when `forceRmIntermediateCtrs` is not set, and suppress the error in that case.

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM, @vrothberg if you can PTAL

vrothberg:
LGTM :+1:

vrothberg:
@rh-atomic-bot r+

PR #1466, opened 2019-04-02
WIP - Remove images/containers created in bud failure case

TomSweeneyRedHat:
When a Dockerfile had clauses that caused a commit to occur,
creating an image, and then had an error in other clauses before
the final commit, the image and containers created were left
behind and should not have been.

Fixes: https://github.com/containers/libpod/issues/2823

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -848,7 +848,7 @@ func (b *Executor) resolveNameToImageRef(output string) (types.ImageReference, e
 }
 
 // Execute runs each of the steps in the stage's parsed tree, in turn.
-func (s *StageExecutor) Execute(ctx context.Context, stage imagebuilder.Stage) (imgID string, ref reference.Canonical, err error) {
+func (s *StageExecutor) Execute(ctx context.Context, stage imagebuilder.Stage) (imgID string, ref reference.Canonical, commitIDs []string, err error) {
 	ib := stage.Builder
 	node := stage.Node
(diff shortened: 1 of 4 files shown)


TomSweeneyRedHat:
FYI @mheon @edsantiago

rhatdan:
LGTM
@TomSweeneyRedHat This is not a problem in Buildah?

TomSweeneyRedHat:
It's a problem in Buildah if the option `--layers=true` is used.   By default that's set to false.  Podman is the reverse, by default layers is set to true.  That's why it popped over there.  The Buildah tests were lacking for this case, the ones I added should catch future hiccups.

The one thing that I'd like @nalind to comment/head nod on is this now also removes the intermediate containers when this error arises.  That did not happen prior and I think should have.

nalind:
We generally want to keep intermediate images around when a build fails so that they can be found by subsequent multi-layer builds that don't use the `--no-cache` flag.

rhatdan:
@TomSweeneyRedHat Considering @nalind Comment and work in other areas should we close this PR?

TomSweeneyRedHat:
@rhatdan, I'm going to keep it open for a bit.  I think there may still be a problem when the -t option is used with layers, but want to verify with the latest bits from @nalind.

TomSweeneyRedHat:
I'm going to close this one and will start a new PR if we decide to do further changes here.

Issue #1465, opened 2019-04-02
buildah from --security-opt label=foo,bar - invalid


edsantiago:
This is in `tests/selinux.bats:65`; I'm not sure why CI isn't catching it:
```
# buildah from --security-opt label=level:s0:c1,c2 alpine
Invalid --security-opt name=value pair: "c2"
```
recent master @ 9e1967ae

edsantiago:
Fixed by #1467

PR #1467, opened 2019-04-02
Fix handling of Slices versus Arrays

rhatdan:
Cobra does not allow a "," in a StringSlice, so we should
be using StringArrays.  This fixes issues with setting
security-opts level flags.


rhatdan:
Closes: https://github.com/containers/buildah/issues/1465

TomSweeneyRedHat:
LGTM, assuming happy tests

rhatdan:
@edsantiago added test to make sure this does not happen again.

TomSweeneyRedHat:
Looks like the test @edsantiago was wondering why it was not failing prior is now failing with these changes.

```
not ok 159 from security-opt test
# (in test file ./from.bats, line 379)
#   `[[ $output =~ "s0:c1,c2" ]]' failed
# time="2019-04-02T19:39:58Z" level=debug msg="[graphdriver] trying provided driver "vfs"" 
# mycontainer
# time="2019-04-02T19:39:58Z" level=debug msg="[graphdriver] trying provided driver "vfs"" 
```

edsantiago:
LGTM but I'm seeing weird CI failures in my test setup; looks like docker.io weirdness. Will troubleshoot later this afternoon.

tests/from.bats
```diff
@@ -367,3 +367,20 @@ load helpers
   [ "$status" -eq 0 ]
   buildah rm ${cid}
 }
+
+@test "from security-opt test" {
```
edsantiago: I don't think this test is necessary; that's what the last one in `selinux.bats` is supposed to do, I think. If you agree, feel free to update that test (adding `buildah inspect`); or LMK and I can work it into my big bats PR.

I would like to see added tests for `--annotations` and `--env`, possibly just piggybacked onto the existing tests in `bud.bats`, but it might be easier for me to add that using my new BATS framework once/if that gets approved.

TomSweeneyRedHat:
@rh-atomic-bot r+

tests/from.bats
TomSweeneyRedHat: @edsantiago if you could massage these changes into your PR or a separate one, that would be awesome.

PR #1470, opened 2019-04-03
docs/tutorials/01-intro.md: use a smaller example for demo package in…

rpjday:
…stall

Replace Java with a much smaller example to demonstrate installing
a package in a container; there's no need to install anything as
sizable as Java

Signed-off-by: Robert P. J. Day <rpjday@crashcourse.ca>


rpjday:
Sure, I'm OK with that. I was just trying to speed things up, but as long as the example works, let's leave it.

PR #1472, opened 2019-04-03
BATS tests - extensive but minor cleanup

edsantiago:
This started off as bug fixes necessary to get BATS tests
actually working on RHEL8 (they weren't). It grew. I will
defend my actions in the first comment post.

Primary change: import some helpers from podman BATS tests,
most importantly 'run_buildah' and 'is'. The vast majority
of the changes you'll see here are of the form:

    - run buildah ...
    - [ $status = 0 ]
    - [ check $output ]
    + run_buildah ...            ! automatically checks status
    + is "$output" "..."

Also: chmod'ed some files -x. Necessary because rpmbuild tries
to be oh-so-clever about requirements, and when it sees an
executable file with a shebang line like '#!env bats' it
helpfully adds 'Requires: /usr/bin/bats' to the rpm, which
then fails to install because RHEL8 does not have bats.

Also: refactored duplicate code in a few places, by writing
and invoking module-specific helper functions.

Also: changed a handful of 'buildah's to run_buildah, in
order to get error checking and debug logging.

Also: added descriptive reasons to many "skip"s.

Also: selinux test: some tweakery to make it run on production
system (context is different if /usr/bin/buildah is chcon'ed
appropriately). I can't get this test to pass on Fedora from
a build dir, and I'm actually not convinced that this test
has ever passed, but let's see what CI shows.

Also: selinux test: skip broken test (#1465).

Also: version test: skip parts of it if running w/o sources.

Tests are now passing as root on RHEL8; rootless has numerous
failures which I don't believe are related to this PR.


edsantiago:
This is a long and cumbersome PR; it will be tedious to review. I am sorry. I am using this comment space to justify it.

The conventional BATS workflow leaves much to be desired:
```
    run buildah this and that
    echo "$output"
    [ $status -eq 0 ]
    [ "$output" = "expected output" ]
```
So many reasons:
* Yeah, you see $output, but you never see the command that produced it. In a long test with multiple buildah commands it's impossible to know which command generated what, especially when run with --debug (which produces copious output).
* If `$status` or `$output` are not as expected, this form gives no indication of what they actually are! This makes it impossible to debug situations like an error message changing, or an off-by-one, because you have no idea what the output or error status are -- only that they aren't what was expected.
* No timeout. A hanging command will block CI without explanation.
* Lots of duplication; need for much syntactic sugar; poor maintainability.

Compare to:
```
    run_buildah this and that
    is "$output" "expected output" "what we're looking at and why"
```
This logs the command and all output; checks for 0 exit status; runs buildah with a timeout; and if exit status or `$output` are not as expected, you'll get logs like:
```
$ buildah first command
output from first command
$ buildah this and that
output from the command
#/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
#|     FAIL: what we're looking at and why
#| expected: 'expected output'
#|   actual: 'output from the command'
#\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Note that multiple buildah commands are easily differentiated, with their output.

I believe that this PR will make it much easier to write future tests, to maintain existing ones, and perhaps most importantly to debug the output of failed tests.

rhatdan:
LGTM, but need @nalind Approval.
Tests are very angry.

edsantiago:
>Tests are very angry.

AFAICT the `timeout` command on the CI system doesn't grok `-v`. I'm now spelunking into coreutils to see how recent that is. And just now realizing that it doesn't matter, it's probably just a nice-to-have but not critical to the PR. I'll force-push without -v.

edsantiago:
Looks like `timeout -v` was added in coreutils-8.29, released December 2017. I can't figure out how to determine the version installed on travis, and am giving up on it. We'll see how it runs without -v.

TomSweeneyRedHat:
@edsantiago I took a very quick look.  It looks OK to me, but I would have much preferred seeing the changes to helpers.bat and then one or two test files for the first PR.

edsantiago:
@TomSweeneyRedHat I'll be glad to resubmit like that if you think it would be helpful; I'm just not sure what that would demonstrate (the added code in `helpers.bash` would be essentially a NOP).

edsantiago:
Two CI pass so far, one failure: `/usr/bin/expr: Argument list too long`, triggered by huge `--debug` output and, I assume, some sort of process limits on travis. Now trying to figure out how to work around it.

edsantiago:
travis seems to be stalled. Is this a keep-twiddling-my-thumbs thing, or a kick-it thing? If the latter, how?

rhatdan:
Much more of a twiddle.

edsantiago:
timed out in `bud with copy-from in Dockerfile no prior FROM`. That was a 120s timeout; I've just force-pushed with a 300s default.

edsantiago:
Failures in `Go:tip` all seem to be identical: 
```
docker/types.go:164:2: struct field Parent repeats json tag "parent" also at types.go:137
```
(but the CI job seems to continue despite this; I'm uncertain as to how or why). Each also has two failures in the BATS job, but I'm not sure if it's meaningful given the compile failure.

Other jobs (the ones at top) all pass. This has been an informative day; I've learned much about platform and timing differences.

rhatdan:
Overall this LGTM.
@TomSweeneyRedHat @nalind @vrothberg PTAL

edsantiago:
@cevich could you also PTAL if you have time?

edsantiago:
Rebased, fixed conflicts, incorporated (part of) review feedback; but waiting for #1467 to merge because I'm also adding new tests that rely on it.

rhatdan:
Sadly @edsantiago looks like it needs another rebase.

edsantiago:
@rhatdan what I meant above is that I've done the rebase & conflict resolution, I just haven't pushed it because I want to add (and confirm) tests for #1467. I may have some time for that this morning, but possibly not all of it.

cevich:
Hmmm, it looks like #1467 merged maybe (it's closed?).  @edsantiago changes look good, but I'm not not sure about other issue to slash-lgtm.  Is this ready?

edsantiago:
Nope, not ready; some pythony stuff is failing. I'll take a look at it next week.

tests/helpers.bash
```diff
@@ -48,6 +51,154 @@ function imgtype() {
...
+#
+# Compares given string against expectations, using '=~' to allow patterns.
+#
+# Examples:
+#
+#   is "$actual" "$expected" "descriptive test name"
+#   is "apple" "orange"  "name of a test that will fail in most universes"
+#   is "apple" "[a-z]\+" "this time it should pass"
+#
+function is() {
```
TomSweeneyRedHat: can we change the name of this function?  Looking for a two character string that is a common word and common to many other words is not fun when grepping through an entire project.  Plus I don't find the function name to represent what the function does.  Sorry, I know this is a big ask for this PR.  Perhaps outContains?

edsantiago: Thanks for mentioning it. With my latest changeset I was thinking of renaming this and changing its usage slightly; I think it could be made much friendlier. I will do so next week. Thanks again.

tests/config.bats
```diff
@@ -3,122 +3,78 @@
...
+  local expect=$2
+
+  for image in docker oci; do
+      for which in Docker OCIv1; do
+        run_buildah --debug=false inspect --type=image --format "{{.$which.$setting}}" scratch-image-$image
+        is "$output" "$expect" "scratch-image-$image > .$which.$setting"
+    done
+  done
+}
+
```
TomSweeneyRedHat: neat little function, made me think for a bit @edsantiago

tests/config.bats
```diff
@@ -188,100 +138,40 @@ load helpers
...
+  check_matrix 'Config.User'         'likes:things'
+  check_matrix 'Config.ExposedPorts' 'map[12345:{}]'
+  check_matrix 'Config.Env'          '[VARIABLE=VALUE1,VALUE2]'
+  check_matrix 'Config.Entrypoint'   '[/bin/sh -c /ENTRYPOINT]'
+  check_matrix 'Config.Cmd'          '[COMMAND-OR-ARGS]'
+  check_matrix 'Config.Volumes'      'map[/VOLUME:{}]'
+  check_matrix 'Config.WorkingDir'   '/tmp'
+  check_matrix 'Config.Labels'       '.*LABEL:VALUE'
+  check_matrix 'Config.Labels'       '.*exec:podman run -it --mount=type=bind,bind-propagation=Z,source=foo,destination=bar /script buz'
+  check_matrix 'Config.StopSignal'   'SIGINT'
```
TomSweeneyRedHat: nit of a nit, might be nice to order the checks alphabetically here by the checked value.  ie. Config.User, Config.Env, etc.

TomSweeneyRedHat:
I'm fine with merging this once the test turn green.  This PR is getting so big I'm a bit concerned about changes to the tests by other commiters before this comes in.  I've one function name change that I'd really like to see happen, but I'm perfectly fine with a secondary PR to take care of that.

tests/config.bats
edsantiago: ...which, to me, means: "needs a comment!". Will do; something like "matrix check all combinations of settings in each container". Thanks.

tests/config.bats
edsantiago: Maybe in a future PR? I wanted to keep those diffs (at least partly) readable

TomSweeneyRedHat: yes, I'm fine with doing that or any further changes in a second PR.  If this goes green, I'd like to merge it.

edsantiago:
@TomSweeneyRedHat thanks, and I agree (that name change is important, and that I should limit other changes because it's already unreviewable). I can't work on the rename today (day off, busy with housework and vet visit) but will try to find time over the weekend.

Is there any way to cancel a pending Travis job? I can't figure it out and I need to go AFK again for a while.

edsantiago:
@TomSweeneyRedHat tests are happy; if you or @nalind would like to merge as-is, I'm fine with that. I won't have time for the `is` rewrite until late Saturday at the earliest.

rhatdan:
@edsantiago Sorry needs a rebase.

TomSweeneyRedHat:
Thanks for this @edsantiago, really nice piece of work and improvement overall.

TomSweeneyRedHat:
@rh-atomic-bot r+

Issue #1405, opened 2019-03-13
Multi-layered build-using-dockerfile builds create too many layers


nalind:
When attempting to switch on multi-layered builds for openshift/builder (in [#35](https://github.com/openshift/builder/pull/35)), one of the tests that failed did so because we're creating more layers than `docker build` does for the test's Dockerfile.

The root cause is that we commit a new layer for every instruction that we process in a `Dockerfile`, when the more conformant behavior would be to do so only for `ADD`, `COPY`, and `RUN` instructions.

One point of reference is [buildkit's parser](https://github.com/moby/buildkit/blob/9faab2a616cde39d10b63f0351a987ac93915484/frontend/dockerfile/dockerfile2llb/convert.go), and calls that it makes to its local `commitToHistory()` function for various instructions.

Option 1: modify Commit() so that we can commit without creating a layer.  This... feels like it's going to lose someone's data at a bad time.
Option 2: use #1300's empty layer history entries to skip the expensive commits, while fixing the history for non-layered builds.  I'm working on this.

rhatdan:
Since the patch for this is merged, I am closing.

Issue #1456, opened 2019-03-28
BUG: tagging is broken for multistage builds


h-vetinari:
[Now that I have access to the nightly ppa's (cf. containers/libpod#2250), I'm starting to file some issues against `podman` - and since the issue template there says that `podman build` issues should be filed for buildah, I'm starting here.]

Assume I have a toy dockerfile, say `podman_test.dockrf`:
```
FROM docker.io/library/ubuntu:bionic

RUN apt-get update && apt-get install curl -y

CMD ["bash"]
```

Then `sudo podman build -t podman_test -f podman_test.dockrf .` produces 
![buildah1](https://user-images.githubusercontent.com/33685575/55168014-51159000-5172-11e9-85a7-4704d1858b66.png)

So far so good. If I change the dockerfile slightly:
```diff
-FROM docker.io/library/ubuntu:bionic
+FROM docker.io/library/ubuntu:bionic as base

 RUN apt-get update && apt-get install curl -y

 CMD ["bash"]
```
then I get an untagged image (even though it successfully uses the cache):
![buildah2](https://user-images.githubusercontent.com/33685575/55168094-7dc9a780-5172-11e9-835c-e4f06e28c8f0.png)

Same for `sudo podman build -t podman_test --target=base -f podman_test.dockrf .`

This is quite basic functionality IMO - without tagging, many automation tasks are not possible.

TomSweeneyRedHat:
@h-vetinari thanks for the issue.  This is definitely not acting correctly as you surmised.  I think the <none><none> is ok for image 21fe, but what I think should be happening is we should have another image there named podman_test with and id of d6ca... , the last line from the build output there.

There appears to be something going on with the --layers option that's not letting that complete successfuly, if I add `--layers=false` to the podman build command, then I see that image.

I'll have to dig more in the next day or too, and may need to check in with @nalind

h-vetinari:
Thanks for the response!

> @TomSweeneyRedHat:  I think the `<none>` `<none>` is ok for image 21fe, but what I think should be happening is we should have another image there named podman_test with an id of d6ca... , the last line from the build output there.

Ideally (IMO) both dockerfiles should create exactly the same image, so that 21fe... is not untagged by the build, but stays the same. Builds not being idempotent is another issue I'm encountering (pending further tests).

nalind:
I think #1473 will fix this.

h-vetinari:
@nalind @TomSweeneyRedHat 
[...]

PR #1473, opened 2019-04-03
imagebuildah: reorganize stage and per-stage logic

nalind:
When we're not building with multiple layers, add empty layer items in the image's history so that the instructions from the Dockerfile don't get lost.

Move the preparation for the first container and the commit of the final image in a single-stage build into the per-stage `Execute()` function.

For multi-stage builds, only create a layer for `ADD`, `COPY`, `RUN`, and the final instruction in a stage, which changes the number of cache images that we create, so a number of tests needed to have their expectations updated.

It also affects the step counts that we output, so a test in the last hunk we change in `bud.bats` needed to be updated to expect the changed value.  In the interests of not making this patch bigger, I expect to include the fix and change that back in a later PR.
    
Stop removing the final image for intermediate stages, since they're expected to be present as cached images, even if they don't end up as ancestors of the final image, except for single-stage builds, which we document as not populating the cache.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -680,17 +680,17 @@ func NewExecutor(store storage.Store, options BuildOptions) (*Executor, error) {
 	return &exec, nil
 }
 
-// Prepare creates a working container based on the specified image, or if one
+// prepare creates a working container based on the specified image, or if one
 // isn't specified, the first argument passed to the first FROM instruction we
 // can find in the stage's parsed tree.
-func (s *StageExecutor) Prepare(ctx context.Context, stage imagebuilder.Stage, from string) error {
+func (s *StageExecutor) prepare(ctx context.Context, stage imagebuilder.Stage, from string) error {
 	ib := stage.Builder
 	node := stage.Node
 
 	if from == "" {
 		base, err := ib.From(node)
 		if err != nil {
-			logrus.Debugf("Prepare(node.Children=%#v)", node.Children)
+			logrus.Debugf("prepare(node.Children=%#v)", node.Children)
 			return errors.Wrapf(err, "error determining starting point for build")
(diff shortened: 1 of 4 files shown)


nalind:
This should fix #1405.

nalind:
We shouldn't be leaving unintended containers after #1446, and leaving intermediate images around is consistent with `docker build`, so I'm not sure if we want to be making changes to our behavior there.

https://github.com/containers/libpod/issues/2823 might be [the TODO that it added](https://github.com/containers/buildah/blob/c933fe4bc608676d308ffcb276b7d8561a18e94d/imagebuildah/build.go#L923), which this should resolve.

TomSweeneyRedHat:
@nalind thanks.  I just verified with Docker-ce and they are leaving an image `<none><none>' behind for this, along with the images from the "FROM" clauses.   My #1146 should not be removing those images I also verified that Docker also does not  remove the containers, so perhaps that shouldn't be done either?

Now for the weird part, when I was looking at this earlier, there were no images with '<none><none>' tagged but now there are.  Has some other new code gotten in that might have fixed that or have I completely lost it?

nalind:
Before this PR, we'd remove the image that was created for stages other than the last one (i.e., the intermediate stages) when we were finished, which I recently noticed differed from how `docker build` treats them, so this PR changes that.
Leaving them around allows the cache checking logic to find them, which is how this should resolve the part of https://github.com/containers/buildah/issues/1456 where we create a new image with a new image ID, in cases when a suitable one should have been present in the cache.

rhatdan:
LGTM
@vrothberg PTAL

giuseppe:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #1474, opened 2019-04-03
imagebuildah: deduplicate prepended "FROM" instructions

nalind:
We currently prepend a `FROM` instruction to the full set of instructions for any images which are referenced in `COPY --from` instructions that we don't create during the build.  Make sure that the
list doesn't include any duplicates.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -1559,6 +1559,10 @@ func processCopyFrom(dockerfiles []io.ReadCloser) []io.ReadCloser {
 	// asMap contains the names of the images seen after a "FROM image AS"
 	// line in the Dockefiles.  The boolean value just completes the map object.
 	asMap := make(map[string]bool)
+	// foreignImages is the list of images that are referenced as sources
+	// of "COPY --from" instructions, but which are not the result of
+	// earlier stages in the build.
+	foreignImages := []string{}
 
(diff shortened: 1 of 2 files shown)


nalind:
This should fix the duplication part of https://bugzilla.redhat.com/show_bug.cgi?id=1684427.

rhatdan:
LGTM

imagebuildah/build.go
```diff
@@ -1559,6 +1559,10 @@ func processCopyFrom(dockerfiles []io.ReadCloser) []io.ReadCloser {
 	// asMap contains the names of the images seen after a "FROM image AS"
 	// line in the Dockefiles.  The boolean value just completes the map object.
 	asMap := make(map[string]bool)
+	// foreignImages is the list of images that are referenced as sources
+	// of "COPY --from" instructions, but which are not the result of
+	// earlier stages in the build.
+	foreignImages := []string{}
```
vrothberg: I think it's cheaper and faster to make `foreignImages` a `map[string]bool` and iterate over the keys down in line 1626.

nalind: A version of this patch did that, but I went with this version to make sure that we pull the images in same order in which they are first referenced in the Dockerfile.

vrothberg: Makes sense, thanks!

vrothberg:
@rh-atomic-bot r+

PR #1476, opened 2019-04-03
WIP - Don't recursive mount inside a container

TomSweeneyRedHat:
When inside of a container, don't do a recursive mount.
This causes issues when trying to run a container
within a container.


rhatdan:
We figured out what was going on here and no longer need this patch.

PR #1477, opened 2019-04-03
commit: always set created-by

nalind:
Set the `CreatedBy` field of the new image's new history item to the shell if we don't have a different value to set.


rhatdan:
LGTM

PR #1485, opened 2019-04-05
imagebuildah: only initialize imagebuilder configuration once per stage

nalind:
Only initialize imagebuilder's copy of the image's configuration once per stage, so that `ONBUILD` instructions that we add along with a layer in a multi-layer build won't be prepended to the current stage's list of instructions when we reinitialize our working rootfs using that layer's container, which is not what we want.

Most of the change in `prepare()` should be an indentation change, so if you're reviewing the patch locally, using `git diff`'s `-w` flag will be helpful.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -683,7 +683,7 @@ func NewExecutor(store storage.Store, options BuildOptions) (*Executor, error) {
 // prepare creates a working container based on the specified image, or if one
 // isn't specified, the first argument passed to the first FROM instruction we
 // can find in the stage's parsed tree.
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM
@TomSweeneyRedHat @vrothberg @giuseppe PTAL

vrothberg:
LGTM
@nalind, out of curiosity: did you measure the performance impact of it?

vrothberg:
@rh-atomic-bot r+

nalind:
@vrothberg this shouldn't impact performance much - whether or not we reset imagebuilder's copy of the image's configuration should be dwarfed by any disk I/O.

TomSweeneyRedHat:
Coming in after the barn door shut, but LGTM

Issue #1424, opened 2019-03-20
RFE - let `buildah commit` generate image IDs


vrothberg:
`buildah commit` requires specifying an image ID which is a problem in automation as the user is confronted with generating hopefully unique IDs (see https://github.com/SUSE/kiwi/pull/1013#issuecomment-474383203 for a background discussion).

To make `buildah commit` more friendly in automation environments, I would love to have some flag (e.g., `--generate-id`) that generates an ID automatically, just as `buildah bud` does.

Cc: @davidcassany @TomSweeneyRedHat @nalind @umohnani8 @rhatdan

rhatdan:
@TomSweeneyRedHat We also need similar changes in podman commit.

TomSweeneyRedHat:
My question here is what's expected for the repository:tag of the committed image?  In Docker it sets them to <none>:<none>.  Would that be good here and Podman too?

Also Docker does that by default if  a repository:tag isn't specified.  I think I'd prefer doing that rather than creating an option for this.

rhatdan:
Yes I would follow the Docker behavior make the specification of the image reference optional.

TomasTomecek:
I hope that I understand this right, but the problem is that `buildah commit` requires an image name. It would be awesome if the name argument was optional and buildah would not set any afterwards resulting into a nameless image. E.g. this what bender does now and it clutters the output quite a bit:
```
localhost/packit-20190507-160521                         latest               03c3e1bddf33         Feb 7, 2019 16:05      673 MB                                            
localhost/packit-20190507-160544                         latest               a54d06efbeba         Feb 7, 2019 16:05      676 MB                                            
localhost/packit-20190607-160612                         latest               ed12304334e0         Feb 7, 2019 16:06      745 MB                                            
localhost/packit-20190607-160633                         latest               4b089426a1c4         Feb 7, 2019 16:06      748 MB                                            
localhost/packit-20190607-160651                         latest               6cd6b78dc4db         Feb 7, 2019 16:06      748 MB                                            
localhost/packit-20190707-160707                         latest               e0a35db19c8d         Feb 7, 2019 16:07      748 MB                                            
localhost/packit-20190707-160725                         latest               1e18476d602d         Feb 7, 2019 16:07      748 MB                                            
```
A ton of layers which should have been hidden, ideally.

`docker commit` accepts image name optionally:
```
Usage:  docker commit [OPTIONS] CONTAINER [REPOSITORY[:TAG]]
```

[...]

PR #1486, opened 2019-04-05
commit: make target image names optional

nalind:
Make the name of the image to create an optional parameter.  If none is specified, use a temporary mostly-random name that can't be interpreted as an ID, so that the image copying logic will compute the correct ID to assign to the new image, and remove the temporary name before returning.

### DIFF
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -12,6 +12,7 @@ import (
 	"github.com/containers/buildah/util"
 	"github.com/containers/image/storage"
 	"github.com/containers/image/transports/alltransports"
+	"github.com/containers/image/types"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -46,7 +47,8 @@ func init() {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return commitCmd(cmd, args, opts)
 		},
-		Example: `buildah commit containerID newImageName
+		Example: `buildah commit containerID
+  buildah commit containerID newImageName
   buildah commit containerID docker://localhost:5000/imageId`,
 	}
 	commitCommand.SetUsageTemplate(UsageTemplate())
@@ -80,6 +82,7 @@ func init() {
 }
 
 func commitCmd(c *cobra.Command, args []string, iopts commitInputOptions) error {
+	var dest types.ImageReference
 	if len(args) == 0 {
 		return errors.Errorf("container ID must be specified")
 	}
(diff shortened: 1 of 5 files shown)


cmd/buildah/commit.go
```diff
@@ -130,20 +133,21 @@ func commitCmd(c *cobra.Command, args []string, iopts commitInputOptions) error
...
-		if err != nil {
-			return errors.Wrapf(err, "error parsing target image name %q", image)
-		}
-		if len(candidates) == 0 {
-			return errors.Errorf("error parsing target image name %q", image)
-		}
-		dest2, err2 := storage.Transport.ParseStoreReference(store, candidates[0])
-		if err2 != nil {
-			return errors.Wrapf(err, "error parsing target image name %q", image)
+	if image != "" {
```
rhatdan: Is it possible for image to == ""?

nalind: If we didn't get one passed as a command line argument, that's the value the variable has at this point.

rhatdan: Ok I misread the code above.  I thought you were defaulting the image name to the container name if the user did not specify an image name.

rhatdan:
LGTM
@TomSweeneyRedHat @vrothberg PTAL

nalind:
This should fix #1424.

nalind:
Rebased.

giuseppe:
LGTM

tests/commit.bats
```diff
@@ -91,3 +91,8 @@ load helpers
   [ "${status}" -eq 0 ]
   [ "$output" == "/bin/sh" ]
 }
+
+@test "commit-no-name" {
+  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+  run_buildah commit --signature-policy ${TESTSDIR}/policy.json $cid
+}
```
TomSweeneyRedHat: can you check the status at least here and maybe output?

tests/bud.bats
```diff
@@ -1046,3 +1046,7 @@ load helpers
   buildah umount ${cid}
   buildah rm ${cid}
 }
+
+@test "bud-no-target-name" {
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json ${TESTSDIR}/bud/maintainer
+}
```
TomSweeneyRedHat: status check here, more if possible?

tests/commit.bats
rhatdan: run_output takes care of this for you.  I believe it makes sure that you exit with 0 and always echos $output

rhatdan:
@rh-atomic-bot r+

PR #1487, opened 2019-04-05
imagebuildah: centralize COMMIT and image ID output

nalind:
Only write the image ID file after we commit the final image, not at every commit.

Log all COMMIT steps and image IDs from inside of Execute().  This undoes a change to step counts that previously required us to adjust the bud-target test in #1473 .

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -883,22 +883,46 @@ func (s *StageExecutor) Execute(ctx context.Context, stage imagebuilder.Stage, b
 	}
 	children := stage.Node.Children
 
+	// A helper function to only log "COMMIT" as an explicit step if it's
+	// the very last step of a (possibly multi-stage) build.
+	logCommit := func(output string, instruction int) {
+		if instruction < len(children)-1 || s.index < s.stages-1 {
+			return
+		}
+		commitMessage := "COMMIT"
+		if output != "" {
+			commitMessage = fmt.Sprintf("%s %s", commitMessage, output)
+		}
+		logrus.Debugf(commitMessage)
+		if !s.executor.quiet {
+			s.executor.log(commitMessage)
+		}
+	}
+	logImageID := func(imgID string) {
+		if s.executor.iidfile == "" {
+			fmt.Fprintf(s.executor.out, "--> %s\n", imgID)
+		}
+	}
+
 	if len(children) == 0 {
 		// There are no steps.
 		if s.builder.FromImageID == "" || s.executor.squash {
 			// We either don't have a base image, or we need to
 			// squash the contents of the base image.  Whichever is
 			// the case, we need to commit() to create a new image.
+			logCommit(s.output, -1)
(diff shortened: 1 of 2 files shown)


imagebuildah/build.go
```diff
@@ -879,22 +879,46 @@ func (s *StageExecutor) Execute(ctx context.Context, stage imagebuilder.Stage, b
 	}
 	children := stage.Node.Children
 
+	// A helper function to only log "COMMIT" as an explicit step if it's
+	// the very last step of a (possibly multi-stage) build.
+	logCommit := func(output string, instruction int) {
+		if instruction < len(children)-1 || s.index < s.stages-1 {
+			return
+		}
+		if output != "" {
```
rhatdan: nit How about
```
commitMsg := "COMMIT"
if output != "" {
    commitMsg := fmt.Println("%s %s", commitMsg, ouput)
}
logrus.Debugf(commitMsg)
if !s.executor.quiet {
		s.executor.log(commitMsg)
}
```

imagebuildah/build.go
```diff
@@ -879,22 +879,46 @@ func (s *StageExecutor) Execute(ctx context.Context, stage imagebuilder.Stage, b
...
 			// We don't need to squash the base image, so just
 			// reuse the base image.
+			logCommit(s.output, -1)
 			if imgID, ref, err = s.copyExistingImage(ctx, s.builder.FromImageID, s.output); err != nil {
 				return "", nil, err
 			}
 		}
+		if s.executor.iidfile == "" {
+			fmt.Fprintf(s.executor.out, "--> %s\n", imgID)
+		}
```
TomSweeneyRedHat: Consideration.  Would it make sense to move this if statement to a small function?  Just on the off chance we ever need to adjust it slightly, we'd only have to do it in one spot.

imagebuildah/build.go
```diff
@@ -1370,14 +1401,8 @@ func (s *StageExecutor) commit(ctx context.Context, ib *imagebuilder.Builder, cr
 	if imageRef != nil {
```
TomSweeneyRedHat: Is this if statement  useful anymore?

TomSweeneyRedHat:
Changes LGTM in general, just a few questions for consideration.  Passes baseline test and a few others that I threw at it too.
Red Hat CI is being onerous though, you may want to repush.

imagebuildah/build.go
nalind: At the moment, I think you're right and it can be removed, but #1486 would have to add it back.

imagebuildah/build.go
nalind: Might as well.

imagebuildah/build.go
nalind: Sure.

imagebuildah/build.go
TomSweeneyRedHat: OK, no worries then, leave it be.

rhatdan:
@rh-atomic-bot r+

PR #1489, opened 2019-04-05
imagebuildah: defer pulling images for COPY --from

nalind:
When handling `COPY --from` instructions, defer pulling images and instantiating containers so that we can read content from their root filesystems until we get to the instruction.

Have the main Executor keep track of the Builders that we create for that purpose, and have it clean them up at the end of the build, as it did when they were treated as their own stages.

This is intended to resolve https://bugzilla.redhat.com/show_bug.cgi?id=1684427.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -10,7 +10,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM, passes baseline tests plus a few others I threw at it

rhatdan:
@rh-atomic-bot r+

PR #1490, opened 2019-04-05
Add a couple of tests

nalind:
Add a test of using `COPY --from` in a Dockerfile while using the cache, checking that we get the same final image, indicating that all cache checks resulted in "hits".

Test squashing (which should probably be called "flattening") of the final image while using the cache for most of the build, in cases where we expect both cache hits and cache misses.

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -1019,6 +1019,17 @@ load helpers
   [ "$out" -ne 0 ]
 }
 
+@test "bud with copy-from and cache" {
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
Can we hold this until #1472 gets in?

nalind:
Sure.  I'm not sure they'll even pass... yet.

rhatdan:
Love new tests.
@rh-atomic-bot r+

PR #1491, opened 2019-04-06
add --dns* flags to buildah bud

### LINKED ISSUE #1438: --dns* flags would be nice to have
**Description**

`biuldah bud` should have `--dns`, `--dns-opt`, and `--dns-search` options just like `podman run` does (see also https://github.com/containers/libpod/issues/2755, requesting same for `podman build`).

My use case is trying to hit package repos as part of a build while using the systemd-resolved stub resolver (127.0.0.53) in my /etc/resolv.conf.

It's probably too much to ask at the moment for the better solution of somehow ultimately querying against the host's resolver library, This would be quite the boon for some situations, but the command line arg would be a workable substitute.

QiWang19:
close #1438 
use --dns* flags to change /etc/resolv.conf in the container during the build.

### DIFF
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -185,6 +185,18 @@ This is a Docker specific option to disable image verification to a Docker
 registry and is not supported by Buildah.  This flag is a NOOP and provided
 soley for scripting compatibility.
 
+**--dns**=[]
+
+Set custom DNS servers
+
+**--dns-option**=[]
+
+Set custom DNS options
+
+**--dns-search**=[]
+
+Set custom DNS search domains
+
 **--file, -f** *Dockerfile*
 
 Specifies a Dockerfile which contains instructions for building the image,
@@ -606,6 +618,8 @@ buildah bud --no-cache -t imageName .
 buildah bud --layers --force-rm -t imageName .
 
 buildah bud --no-cache --rm=false -t imageName .
+
+buildah bud --dns-search=example.com --dns=223.5.5.5 --dns-option=use-vc .
 
 ### Building an image using a URL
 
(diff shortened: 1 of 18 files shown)


rhatdan:
Looks like you did not include some files in the pkg/resolvconf directory.
Also we need some tests added to prove this works.

Need updates to man pages and completions.

QiWang19:
Add tests. Fixed docs/buildah-bud.md and completions

pkg/resolvconf/dns/resolvconf.go
```diff
@@ -0,0 +1,28 @@
+// Originally from github.com/docker/libnetwork/resolvconf/dns
```
rhatdan: Why can't we just vendor it from there?

pkg/resolvconf/resolvconf.go
```diff
@@ -0,0 +1,250 @@
+// Package resolvconf provides utility code to query and update DNS configuration in /etc/resolv.conf.
```
rhatdan: Why not just vendor it?

QiWang19:
vendor from vendor/github.com/docker/libnetwork and pass the tests

docs/buildah-bud.md
```diff
@@ -185,6 +185,18 @@ This is a Docker specific option to disable image verification to a Docker
...
+Set custom DNS servers
+
+**--dns-option**=[]
+
+Set custom DNS options
+
+**--dns-search**=[]
+
+Set custom DNS search domains
+
```
TomSweeneyRedHat: Can you add an example or two?

QiWang19: :ok_hand:  will add an example in this patch

rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot r+

PR #1003, opened 2018-09-11
run container-diff tests in Travis

lsm5:
This commit will cause Travis to spawn a fedora container
which will build a Dockerfile using docker and buildah bud
and then compare both images using container-diff

Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>

I'm thinking of moving the stuff in `.travis.yml` to the Fedora container, but that's for later.

@nalind @TomSweeneyRedHat @mrunalp PTAL

PR #1494, opened 2019-04-07
Travis CI: fail upon failure in any step

edsantiago:
Use 'set -e' to exit immediately upon any failure: there is,
e.g., no reason to run BATS tests if the compile fails.

Background: in my brief experience here, the Go:tip jobs
inevitably fail. Although these are marked as allowed to
fail, I tried looking at the results anyway to see if the
failure was my fault; it took me a while to figure out
that the go build was failing, yet the CI job was plodding
on regardless. This seems wrong on infinite levels, and
a number of other people agree:

    https://github.com/travis-ci/travis-ci/issues/1066

I resent this waste of my time, and would like to prevent
others from similarly wasting theirs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,6 +67,8 @@ install:
     - docker run --entrypoint htpasswd registry:2 -Bbn testuser testpassword > /home/travis/auth/htpasswd
     - docker run -d -p 5000:5000 --name registry -v /home/travis/auth:/home/travis/auth:Z -e "REGISTRY_AUTH=htpasswd" -e "REGISTRY_AUTH_HTPASSWD_REALM=Registry Realm" -e REGISTRY_AUTH_HTPASSWD_PATH=/home/travis/auth/htpasswd -e REGISTRY_HTTP_TLS_CERTIFICATE=/home/travis/auth/domain.crt -e REGISTRY_HTTP_TLS_KEY=/home/travis/auth/domain.key registry:2
 script:
+    # Fail fast
+    - set -e
     # Let's do some docker stuff just for verification purposes
     - docker ps --all
     - docker images
@@ -89,12 +91,3 @@ script:
     - tmp=`mktemp -d`; mkdir $tmp/root $tmp/runroot; sudo PATH="$PATH" ./buildah.test -test.v --root $tmp/root --runroot $tmp/runroot --storage-driver vfs --signature-policy `pwd`/tests/policy.json --registries-conf `pwd`/tests/registries.conf
     - cd tests; sudo PATH="$PATH" ./test_runner.sh
     - cd ..
-after_script:
-    - echo "Run conformance tests in Fedora container..."
-    - echo "Travis/host environment:"
-    - env
-    - env TEST_GROUP=conformance
-    - echo
-    - echo "Running tests in SPC using ./hack/run_build.sh"
-    - "./hack/run_build.sh"
-


edsantiago:
Good news: this worked as expected. All the `Go:tip` jobs failed within 8 minutes, at the expected `go build` error, thus saving compute cycles and human effort.

Bad news: all the other CI jobs failed too, in the `after_script` section at the very very end:
```
./hack/spc_build.sh: line 20: dnf: command not found
```
`spc_build.sh` is invoked from `hack/run_build.sh`. 

My gut reaction is to add `set +e` before running this, and thus sweep this problem back under the rug, but I have no idea what either of these scripts are intended to accomplish so I will leave it up to the CI experts to look into this further.

edsantiago:
@lsm5 those scripts seem to be yours; could you comment on how important it is for them to run and pass in CI environment? TIA.

lsm5:
the scripts in hack are for running ubuntu builds on Travis.

lsm5:
But I thought those scripts were *not* merged upstream.

lsm5:
errrr ... my bad, the ones in hack/ that you mention above are for running container-diff tests. The ubuntu autobuilds are on my own fork.

edsantiago:
Thanks, but to rephrase my question: these scripts are failing due to missing `dnf`. It seems likely that they have been failing for a while. Are you monitoring their output? Are they serving any useful purpose by running in CI?

lsm5:
ah ok, nope, I haven't been :( ..and nobody pointed me to these earlier. Also, your PR LGTM.

edsantiago:
What would you suggest, then? Remove the `hack/run_build.sh` line? Remove the entire `after_script`? Fix it so it runs? I don't have enough background info to know what's best here.

lsm5:
hmm, so IIRC, these were set to not block on anything. If we remove those lines, container-diff tests won't run at all. Let me get back to you after I recollect info on this. This was quite a while back.

edsantiago:
From a quick glance, I think part of the problem is
```
hack/run_build.sh:10:  DISTRO="${DISTRO:-fedora}"
```
However, in this CI environment:
```
Setting environment variables from .travis.yml
...
export DISTRO="ubuntu"
...
Preparing to run $SPCMD=./hack/spc_build.sh in a ubuntu SPC.    <--- probably not what's desired
```

edsantiago:
@lsm5 could you take a quick new look, at my 2nd commit and at the new results?

I made a small change to `hack/run_build.sh`, making an assumption that the use of `$DISTRO` was generic, and the collision with same envariable in CI was accidental. That gets the job much further, but all jobs now fail with:
```
Error unpacking rpm package tzdata-2019a-1.fc29.noarch
error: unpacking of archive failed on file /usr/share/zoneinfo/Etc/UTC: cpio: rename failed - Device or resource busy
```
Without knowing the purpose behind these scripts, I'm not sure how to proceed. I'm pretty sure they have never run properly, and am tempted to remove them completely from Travis, but I'd like your feedback.

edsantiago:
Help, please.

I am happy with the `set -e` portion of this PR (my original goal): the broken builds fail in 8 minutes (vs 30), and the tail end of the log clearly shows the failure, without a misleading BATS run.

The `hack/run_build.sh` is proving a challenge. I've tried two ways of getting it to run, and it still fails; only now it's adding 5 minutes to each CI run instead of failing early. 

I don't believe that anyone here wants CI to take five more minutes; I don't believe this step has to be run on each individual job; and I actually don't believe this step should even run at all, given that it has probably been failing since Day One and nobody has noticed. I would like to remove it from `.travis.yml`.

@lsm5 @nalind @TomSweeneyRedHat @rhatdan advice please?

rhatdan:
Makes sense to me.
LGTM

edsantiago:
[CI failure](https://travis-ci.org/containers/buildah/jobs/518902223). Looks like a flake; I see these occasionally but can't reproduce them. Is this related to the cri-o failures discussed in Planning yesterday?
```
$ /home/travis/gopath/src/github.com/containers/buildah/tests/./../buildah --debug=false from --signature-policy /home/travis/gopath/src/github.com/containers/buildah/tests/./policy.json --quiet --pid= alpine
# The following failures happened while trying to pull image specified by "alpine" based on search registries in /home/travis/gopath/src/github.com/containers/buildah/tests/./registries.conf:
# * "localhost/alpine": Error determining manifest MIME type for docker://localhost/alpine:latest: pinging docker registry returned: Get https://localhost/v2/: dial tcp 127.0.0.1:443: connect: connection refused
# * "docker.io/library/alpine": Error reading blob sha256:bdf0201b3a056acc4d6062cc88cd8a4ad5979983bfb640f15a145e09ed985f92: Invalid status code returned when fetching blob 503 (Service Unavailable)
# * "registry.fedoraproject.org/alpine": Error determining manifest MIME type for docker://registry.fedoraproject.org/alpine:latest: Error reading manifest latest in registry.fedoraproject.org/alpine: manifest unknown: manifest unknown
...
```

edsantiago:
ping, I'm happy with this now.

rhatdan:
@mheon @lsm5 @baude @giuseppe @vrothberg @TomSweeneyRedHat PTAL

mheon:
LGTM

TomSweeneyRedHat:
@edsantiago I'd thought the hack/run_build.sh was running the conformance tests OK, I was unaware of them not running or having problems, but that doesn't mean they haven't been.  With this change I don't think they'll run now via the CI in any circumstance correct?

@ypu I believe you put these tests together, do you want to weigh in?  Are you good with the conformance tests not running with each build?

edsantiago:
The tests have not been running for a long time, quite probably have never run. I tried to read the code and understand it, but was unable to, and I'm not motivated to spend any further effort on them. I would like to have this PR approved, because its benefits are immediate: quick failure on CI error. This benefits everyone. If there is someone interested in looking at the code I removed, they can do so as a future PR. I would really not like to make this PR become about fixing something unrelated that has been broken for a long time.

rhatdan:
@rh-atomic-bot r+

TomSweeneyRedHat:
@edsantiago I'm fine with going forward with this now, but would like @ypu to weigh in about possibly getting it back on-line at some point.

edsantiago:
Odd. The history I see relates to #1003 by @lsm5. But I agree entirely, Tom: I would like someone, anyone, to take ownership of those scripts; and, if they're important, make them work.

PR #1497, opened 2019-04-08
commit: provide an image ID more often

nalind:
When we finish committing the image, use the just-written image's manifest and the diffIDs list from its config blob to compute the image's ID, so that it can be known even if it was written to a location other than local storage.

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -221,32 +221,61 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 			logrus.Warnf("don't know how to add tags to images stored in %q transport", dest.Transport().Name())
 		}
 	}
-
-	img, err := is.Transport.GetStoreImage(b.store, dest)
-	if err != nil && err != storage.ErrImageUnknown {
-		return imgID, nil, "", errors.Wrapf(err, "error locating image %q in local storage", transports.ImageName(dest))
-	}
+	// Calculate the image ID, now that we have a real image with a
+	// configuration blob that's in its final format.
+	img, err := dest.NewImage(ctx, systemContext)
 	if err == nil {
-		imgID = img.ID
-		prunedNames := make([]string, 0, len(img.Names))
-		for _, name := range img.Names {
-			if !(nameToRemove != "" && strings.Contains(name, nameToRemove)) {
-				prunedNames = append(prunedNames, name)
-			}
+		var manifestType string
+		manifestBytes, manifestType, err = img.Manifest(ctx)
+		if err != nil {
+			return imgID, nil, "", errors.Wrapf(err, "error reading manifest from new image %q", transports.ImageName(dest))
 		}
-		if len(prunedNames) < len(img.Names) {
-			if err = b.store.SetNames(imgID, prunedNames); err != nil {
-				return imgID, nil, "", errors.Wrapf(err, "failed to prune temporary name from image %q", imgID)
-			}
-			logrus.Debugf("reassigned names %v to image %q", prunedNames, img.ID)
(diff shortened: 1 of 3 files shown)


rhatdan:
@vrothberg @mtrmac PTAL

vrothberg:
See https://github.com/containers/buildah/pull/1497/files#diff-d280db36f672412d557878c23067458dR204.

commit.go
```diff
@@ -221,32 +221,61 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
...
 		}
 	}
-
-	img, err := is.Transport.GetStoreImage(b.store, dest)
-	if err != nil && err != storage.ErrImageUnknown {
-		return imgID, nil, "", errors.Wrapf(err, "error locating image %q in local storage", transports.ImageName(dest))
-	}
+	// Calculate the image ID, now that we have a real image with a
+	// configuration blob that's in its final format.
+	img, err := dest.NewImage(ctx, systemContext)
```
mtrmac: If `dest` is not local storage (and, strictly speaking, even if it were), nothing guarantees that `dest` points to exactly the just written image (e.g. if `dest` is a tag reference, the tag can be reassigned at any time).

That’s, to an extent, a weakness in the `ImageReference` design — but transports that write into mutable shared storage just don’t have stable references to immutable content (even digest references don’t work, because the image may be removed in the meantime), so any write+read is never guaranteed to return the originally written data in principle.

commit.go
```diff
@@ -221,32 +221,61 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
...
+			return imgID, nil, "", errors.Wrapf(err, "error reading configuration from new image %q", transports.ImageName(dest))
+		}
+		// Ask the manifest for the image's ID.
+		imgID, err = mfest.ImageID(config.RootFS.DiffIDs)
+		if err != nil {
+			return imgID, nil, "", errors.Wrapf(err, "error computing image ID for new image %q", transports.ImageName(dest))
+		}
+		// If we attached a temporary name to the image, remove it now.
+		if nameToRemove != "" {
+			img, err := b.store.Image(imgID)
```
mtrmac: This only works if the image exists locally; at that point this code could probably use `b.store.ImagesByDigest(manifest.Digest(manifestBytes/*original from copy.Image*/))` to avoid the re-tagging race, and it doesn’t have to compute the ID.

commit.go
```diff
@@ -221,32 +221,61 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
...
+					return imgID, nil, "", errors.Wrapf(err, "failed to prune temporary name from image %q", imgID)
+				}
+				logrus.Debugf("reassigned names %v to image %q", prunedNames, imgID)
 			}
-			dest = dest2
 		}
+		// Write the new image's ID to a file, if we were asked to.
 		if options.IIDFile != "" {
-			if err = ioutil.WriteFile(options.IIDFile, []byte(img.ID), 0644); err != nil {
+			if err = ioutil.WriteFile(options.IIDFile, []byte(imgID), 0644); err != nil {
```
mtrmac: This only makes sense for storage destinations; no other destination can look up images using this value, so I don’t know what good does writing this file, or caring about the value written into this file, do; it seems to me that this could just as well write `no ID exists, sorry` for non-storage.

commit.go
nalind: Aargh, that's right, the tag can move in the interim.

rhatdan:
@nalind Are you going to close this PR?

nalind:
Yeah.  It's not going to be reliable at all.

PR #1505, opened 2019-04-10
Add hack/build_speed.sh  test speeds on building container images

rhatdan:
The main goal of this script is test and time builds using Buildah or Docker.
 We hope to use it to help optimize Buildah build performance

 It takes two options
 First option tells the type of the container image
 build to do.  Valid options are:
       Docker  - docker build
       Buildah - buildah bud
       Both    - Do docker build followed by buildah bud

 Second Option specifies a directory or cleanup
 The script will 'find' files beginning with Dockerfile, for each Dockerfile
 it finds it will run a build with the Dockerfile and directory for the
 context. When it does the builds, it will call time on them to show how
 long the builds take.  The created image name will be a combination of the
 lowercased Directory name that the Dockerfile was found in plus the lower
 cased dockerfile name.

 if the second field is cleanup, the script will remove all images from the
 specified builder.

 The script does not check for conflicts on nameing.

### DIFF
--- a/hack/build_speed.sh
+++ b/hack/build_speed.sh
@@ -1,0 +1,128 @@
+#! /bin/sh
+
+# The main goal of this script is test and time builds using Buildah or Docker.
+# We hope to use it to help optimize Buildah build performance
+#
+# It takes two options
+# First option tells the type of the container image
+# build to do.  Valid options are:
+#       Docker  - docker build
+#       Buildah - buildah bud
+#       Both    - Do docker build followed by buildah bud
+#
+# Second Option specifies a directory or cleanup
+# The script will 'find' files beginning with Dockerfile, for each Dockerfile
+# it finds it will run a build with the Dockerfile and directory for the
+# context. When it does the builds, it will call time on them to show how
+# long the builds take.  The created image name will be a combination of the
+# lowercased Directory name that the Dockerfile was found in plus the lower
+# cased dockerfile name.
+#
+# if the second field is cleanup, the script will remove all images from the
+# specified builder.
+#
+# The script does not check for conflicts on nameing.
+#
+# Outputs file:
+#
+#
+# cat /tmp/build_speed.json
+# {
+#   "/usr/share/fedora-dockerfiles/redis/Dockerfile": {
+#     "docker": {
+#       "command": "docker build -f /usr/share/fedora-dockerfiles/redis/Dockerfile -t redis_dockerfile /usr/share/fedora-dockerfiles/redis",
+#       "real": "3:28.70"
+#     },
+#     "buildah": {
+#       "command": "buildah bud --layers -f /usr/share/fedora-dockerfiles/redis/Dockerfile -t redis_dockerfile /usr/share/fedora-dockerfiles/redis",
+#       "real": "2:55.48"
+#     }
+#   }
+# }
+#
+# Examples uses
+#     ./build_speed.sh Docker ~/MyImages
+#     ./build_speed.sh Both /usr/share/fedora-dockerfiles/django/Dockerfile
+
+#totalsfile=$(mktemp /tmp/buildspeedXXX.json)
+totalsfile=/tmp/build_speed.json
+commaDockerfile=""
+
+echo -n '{' > $totalsfile
+Dockerfiles() {
+    find -L $1 -name Dockerfile\*
+}
+
+Buildah() {
+    Name=$1
+    Dockerfile=$2
+    Context=$3
+    echo buildah bud --layers -f ${Dockerfile} -t ${Name} ${Context}
+    Time buildah bud --layers -f ${Dockerfile} -t ${Name} ${Context}
+}
+
+Time() {
+    outfile=$(mktemp /tmp/buildspeedXXX)
+    /usr/bin/time -o $outfile --f "%E" $@
+    echo "{\"engine\": \"$1\", \"command\": \"$@\", \"real\": \"$(cat ${outfile})\"}"
+    echo -n "${comma}\"$1\": {\"command\": \"$@\", \"real\": \"$(cat ${outfile})\"}" >> $totalsfile
+    comma=","
+    rm -f $outfile
+}
+
+Docker() {
+    Name=$1
+    Dockerfile=$2
+    Context=$3
+    echo docker build -f ${Dockerfile} -t ${Name} ${Context}
+    Time docker build -f ${Dockerfile} -t ${Name} ${Context}
+}
+
+Both() {
+    comma=""
+    echo -n "${commaDockerfile}\"$2\": {" >> $totalsfile
+    commaDockerfile=","
+    Docker $1 $2 $3
+    Buildah $1 $2 $3
+    echo -n "}" >> $totalsfile
+}
+
+Docker_cleanup() {
+    docker rmi --force $(docker images  -q) 
+}
+
+Buildah_cleanup() {
+    buildah rmi --force --all
+}
+
+Both_cleanup() {
+    Docker_cleanup
+    Buildah_cleanup
+}
+
+Cmd=${1?Missing CMD argument}
+Path=${2?Missing PATH argument}
+
+case "$Cmd" in
+    Docker)   ;;
+    Buildah)  ;;
+    Both)     ;;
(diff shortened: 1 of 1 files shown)


rhatdan:
@mrunalp @TomSweeneyRedHat @nalind @edsantiago PTAL

hack/build_speed.sh
```diff
@@ -0,0 +1,81 @@
...
+    Path=$2
+    for i in $(Dockerfiles ${Path});do
+	echo "DAN"
+	name=$(basename $(dirname $i) | sed -e 's/\(.*\)/\L\1/')
+	name=${name}_$(basename $i | sed -e 's/\(.*\)/\L\1/')
+	echo ${Cmd} $(basename $i) ${name} $(dirname $i)
+	${Cmd} ${name} $(basename $i) $(dirname $i)
+    done
+}
+if $2 == cleanup {
```
edsantiago: Just a hunch, but I'm guessing you didn't test this... the next line is also wrong (it needs braces).

For error-checking's sake, I'd probably get rid of `Builder()`, making that all inline; and I'd write lines 64-onward as:
```bash
Cmd=${1?Missing CMD argument}
Path=${2?Missing PATH argument}

case "$Cmd" in
    Docker)   ;;
    Buildah)  ;;
    Both)     ;;
    *)     echo "Invalid command '$Cmd'; must be Buildah, Docker, or Both"; exit 1;;
esac


if [ "$Path" == "cleanup" ]; then
    ${Cmd}_cleanup
    exit 0
fi

for i in $(Dockerfiles ${Path});do
    echo "DAN"
    name=$(basename $(dirname $i) | sed -e 's/\(.*\)/\L\1/')
    name=${name}_$(basename $i | sed -e 's/\(.*\)/\L\1/')
    echo ${Cmd} $(basename $i) ${name} $(dirname $i)
    ${Cmd} ${name} $(basename $i) $(dirname $i)
done
```

hack/build_speed.sh
```diff
@@ -0,0 +1,81 @@
...
+}
+
+Builder() {
+    Cmd=$1
+    Path=$2
+    for i in $(Dockerfiles ${Path});do
+	echo "DAN"
+	name=$(basename $(dirname $i) | sed -e 's/\(.*\)/\L\1/')
+	name=${name}_$(basename $i | sed -e 's/\(.*\)/\L\1/')
+	echo ${Cmd} $(basename $i) ${name} $(dirname $i)
```
sjug: @edsantiago beat me to the other one, but this also doesn't work. 

This should:
```
echo ${Cmd} ${name} $i $(dirname $i)
${Cmd} ${name} $i $(dirname $i)
```

hack/build_speed.sh
```diff
@@ -0,0 +1,85 @@
...
+# long the builds take.  The created image name will be a combination of the
+# lowercased Directory name that the Dockerfile was found in plus the lower
+# cased dockerfile name.
+#
+# if the second field is cleanup, the script will remove all images from the
+# specified builder.
+#
+# The script does not check for conflicts on nameing.
+#
+
```
TomSweeneyRedHat: It would be nice to have an example or two of execution here.

TomSweeneyRedHat: not the output, just './build_speed.sh Docker ~/MyDockerfiles'  or some such.

hack/build_speed.sh
rhatdan: Sorry I as testing a copy of this, and then committed the older version, I sent updates. But I like your changes I will add them, and remove the echo "DAN", :^)

rhatdan:
@sjug I now have the script create a json file with all timings in it.

```
cat /tmp/buildspeed7Fn.json
[{"engine": "buildah", "command": "buildah bud --layers -f /root/Dockerfile.test -t root_dockerfile.test /root", "real": "2:10.05"},
{"engine": "docker", "command": "docker build -f /root/Dockerfile.test -t root_dockerfile.test /root", "real": "2:10.34"}]
``

**Comment by sjug** [comment]
@rhatdan Nice! Might be nice to have a shared/common value to key off like `name: ${name}` for comparison rather than pulling the value for `-t` out of `command:`. Unless you output one `.json` file per `Dockerfile` build (that has one run from docker & one run from buildah), which could get messy.

**Comment by TomSweeneyRedHat** [comment]
LGTM

File: hack/build_speed.sh [hunk 4]
```diff
@@ -0,0 +1,126 @@
...
+    Dockerfile=$2
+    Context=$3
+    echo buildah bud --layers -f ${Dockerfile} -t ${Name} ${Context}
+    Time buildah bud --layers -f ${Dockerfile} -t ${Name} ${Context}
+}
+
+Time() {
+    outfile=$(mktemp /tmp/buildspeedXXX)
+    /usr/bin/time -o $outfile --f "%E" $@
+    echo "{\"engine\": \"$1\", \"command\": \"$@\", \"real\": \"$(cat ${outfile})\"}"
```
edsantiago: hint: `$(< outfile)` (dollar paren less-than) is equivalent to `$(cat` and, some believe, shorter and more readable.

hack/build_speed.sh
```diff
@@ -0,0 +1,126 @@
...
+    if [ ${first} -ne 1 ]; then
```
edsantiago: hint: a possibly more readable way to do this is to define `comma=` at top, then always do `echo "...${comma}"`, and immediately after the first echo do `comma=","`.

I'm confused about the new change from `first` to `firstBoth`; it seems to have changed in some places but not others?

rhatdan: @edsantiago I am changing the output format to be a map.
{
  "/usr/share/fedora-dockerfiles/redis/Dockerfile": {
    "docker": {
      "command": "docker build -f /usr/share/fedora-dockerfiles/redis/Dockerfile -t redis_dockerfile /usr/share/fedora-dockerfiles/redis",
      "real": "3:28.70"
    },
    "buildah": {
      "command": "buildah bud --layers -f /usr/share/fedora-dockerfiles/redis/Dockerfile -t redis_dockerfile /usr/share/fedora-dockerfiles/redis",
      "real": "2:55.48"
    }
  }
}

rhatdan: So now I need commas between multiple Dockerfiles.

rhatdan: Repushed using comma and commaDockerfile

hack/build_speed.sh
```diff
@@ -0,0 +1,134 @@
...
+
+Time() {
+    outfile=$(mktemp /tmp/buildspeedXXX)
+    /usr/bin/time -o $outfile --f "%E" $@
+    echo "{\"engine\": \"$1\", \"command\": \"$@\", \"real\": \"$(cat ${outfile})\"}"
+    echo ${comma}  >> $totalsfile
+    if [ ${comma} == "" ]; then
+	comma=","
+    fi
+    echo -n "\"$1\": {\"command\": \"$@\", \"real\": \"$(cat ${outfile})\"}" >> $totalsfile
```
edsantiago: What I was trying for is simplicity: there's no need to `if` anything, and no need for extra echo lines. You can replace lines 68-72 with:
```
    echo -n "${comma}\"$1\":....
    comma=","
```
i.e. the comma is printed on all iterations except the first one. (you may want to initialize `comma=` around line 50 as good practice).

hack/build_speed.sh
```diff
@@ -0,0 +1,134 @@
...
+    Time docker build -f ${Dockerfile} -t ${Name} ${Context}
+}
+
+Both() {
+    comma=","
+    echo ${commaDockerfile}  >> $totalsfile
+    if [ ${commaDockerfile} == "" ]; then
+	commaDockerfile=","
+    fi
+    echo -n "\"$2\": {" >> $totalsfile
```
edsantiago: Same here (no `if` needed)

rhatdan: Done

hack/build_speed.sh
rhatdan: done

edsantiago:
Thanks. LGTM.

rhatdan:
@rh-atomic-bot r=edsantiago

PR #1506, opened 2019-04-10
Create buildahimage Dockerfile for Quay

TomSweeneyRedHat:
The Dockerfile in the buildahimage directory will be
used by quay.io/buildah when ever a new tag is created
here in this GitHub repo.  It will build a new buildahimage
image automatically.

At least that's the theory.

### DIFF
--- a/buildahimage/Dockerfile
+++ b/buildahimage/Dockerfile
@@ -1,0 +1,6 @@
+FROM fedora
+RUN yum -y install buildah --exclude container-selinux --enablerepo updates-testing; rm -rf /var/cache /var/log/dnf* /var/log/yum.*
+RUN sed 's|^#mount_program|mount_program|g' -i /etc/containers/storage.conf
+ENV _BUILDAH_STARTED_IN_USERNS="" BUILDAH_ISOLATION=chroot
+COPY Dockerfile /root
+


rhatdan:
Probably need some documentation on how to use this.
LGTM

TomSweeneyRedHat:
Yeah a bit of POC at the moment, will add a README.md once the dust settles.  But in short and in theory, quay.io/buildah will point at this Dockerfile and use it to build a new buildahimage every time the GitHub tag for Buildah changes or when we press a button on the quay.io site to build it "by hand".

rhatdan:
@giuseppe @nalind @vrothberg PTAL

buildahimage/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM fedora
+RUN yum -y install buildah --exclude container-selinux --enablerepo updates-testing; rm -rf /var/cache /var/log/dnf* /var/log/yum.*
+RUN sed 's|^#mount_program|mount_program|g' -i /etc/containers/storage.conf
```
vrothberg: Can we add a comment to line 3 and line 4 explaining why those changes are necessary? I think this would help folks brain-parse the Dockerfile.

TomSweeneyRedHat: I could, but this particular Dockerfile is a proof of concept that I put in place to see if the buildahimage does get built auto-magically on quay.io when a PR is merged in.  I'm not planning to keep this Dockerfile as is long term, I think I'm going to change it to clone upstream buildah, build it and install it.  That way every time a PR goes through, the latest bits could be pulled/tested directly from quay.io.

vrothberg: Sounds good to me :+1:

vrothberg:
@rh-atomic-bot r+

buildahimage/Dockerfile
rhatdan: Can we get two images, One built off of released versions, and one off of master.

TomSweeneyRedHat: I'm worried, I'd the same thought as Dan, but I was thinking three.

1) image from latest push in GitHub  (buildah/buildahimagegit)
2) image from latest release that's stable on Fedora (buildah/buildahimage)
3) image from latest updates testing repo on Fedora, which at times might be the same as 2 (buildah/buildaimagetesting)

Names for each in parens, happy to take suggestions there for any.

PR #1508, opened 2019-04-11
Add Dockerfiles for buildahimages

TomSweeneyRedHat:
Add a Dockerfile for each buildah image that we will
be maintaining in quay.io/buildah.  They are stable,
git and testing.  The resulting image names will be
buldaimage, buildahimagegit, buildahimagetesting in
order.

### DIFF
--- a/buildahimage/README.md
+++ b/buildahimage/README.md
@@ -1,0 +1,47 @@
+![buildah logo](https://cdn.rawgit.com/containers/buildah/master/logos/buildah-logo_large.png)
+
+# buildahimage
+
+## Overview
+
+This directory contains the Dockerfiles necessary to create the three buildahimage container
+images that are housed on quay.io under the buildah account.  All three repositories where
+the images live are public and can be pulled without credentials.  These container images are secured and the
+resulting containers can run safely with privileges within the container.  The container images are built
+using the latest Fedora and then Buildah is installed into them:
+
+  * quay.io/buildah/buildahimage - This image is built using the latest stable version of Buildah in a Fedora based container.  Built with buildahimage/stable/Dockerfile.
+  * quay.io/buildah/buildahimagegit - This image is built using the latest code found in this GitHub repository.  When someone creates a commit and pushes it, the image is created.  Due to that the image changes frequently and is not guaranteed to be stable.  Containers built with this imsage have a Buildah development environment set up in the directory /root/buildah.  Built with buildahimage/git/Dockerfile.
+  * quay.io/buildah/buildahimagetesting - This image is built using the latest version of Buildah that is or was in updates testing for Fedora.  At times this may be the same as the stable image.  This container image will primarily be used by the development teams for verification testing when a new package is created.  Built with buildahimage/testing/Dockerfile.
+
+## Sample Usage
+
+Although not required, it is suggested that [Podman](https://github.com/containers/libpod) be used with these container images.
+
+```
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM
@vrothberg @lsm5 @giuseppe PTAL

buildahimage/README.md
```diff
@@ -0,0 +1,42 @@
...
+## Sample Usage
+
+Although not required, it is suggested that [Podman](https://github.com/containers/libpod) be used with these container images.
+
+```
+podman pull docker://quay.io/buildah/buildahimage:latest
+
+podman run buildahimage buildah version
+
+mkdir /var/lib/mycontainer
```
vrothberg: Can we add a comment why we need this volume?

vrothberg: It also implies we're running as root which seems like a trap users can step into if they try rootless.

buildahimage/Dockerfile.git
```diff
@@ -0,0 +1,51 @@
...
+# Build a Buildah image from the latest stable
+# version of Buildah on the Fedoras Updates System.
+# This builds a secured container that can run
+# safely with privileges within the container.
+#
+FROM fedora
+ENV GOPATH=/root/buildah
+
+# Install the softwware required to build Buildah,
+# Then create a directory and clone from the
```
vrothberg: s/Then/then/

buildahimage/Dockerfile.git
```diff
@@ -0,0 +1,51 @@
...
+     go-md2man \
+     runc \
+     fuse-overlayfs \
+     fuse3 \
+     containers-common; \
+     mkdir /root/buildah; \
+     git clone https://github.com/containers/buildah /root/buildah/src/github.com/containers/buildah; \
+     cd /root/buildah/src/github.com/containers/buildah; \
+     make;\
+     make install
```
vrothberg: A few thoughts on the RUN command above:
- Can we split the the package installation and the git clone + build into separate runs? The thought is that it's a pretty long running RUN task and network flakes after package installation would allow for reusing the package-installation layer.
- We could/should remove `/root/buildah` after `make install` to reduce the image size. My local repo with some remotes is ~65M.

buildahimage/Dockerfile.git
```diff
@@ -0,0 +1,51 @@
+# Dockerfile.stable
```
vrothberg: s/stable/git/

buildahimage/Dockerfile.git
```diff
@@ -0,0 +1,51 @@
...
+#
+# Build a Buildah image from the latest stable
+# version of Buildah on the Fedoras Updates System.
+# This builds a secured container that can run
+# safely with privileges within the container.
```
vrothberg: The description seems to describe another image :^)

buildahimage/Dockerfile.git
```diff
@@ -0,0 +1,51 @@
...
+#
+FROM fedora
+ENV GOPATH=/root/buildah
+
+# Install the softwware required to build Buildah,
```
vrothberg: s/softwware/software/

buildahimage/README.md
TomSweeneyRedHat: I've added a note about the volume.
It's a good implication for root.  At the moment rootless is a bit sketchy.  Will come back later on and rework the examples once that's a bit steadier.

vrothberg:
@rh-atomic-bot r+

Issue #765, opened 2018-06-08
Build/Push/Pull via HTTP Proxy


maxmeyer:
**Description**

**Steps to reproduce the issue:**
1. `sudo buildah build-using-dockerfile --file files/latest/Dockerfile` 

**Describe the results you received:**

Installing packages by `apk add` hangs with no progress

**Describe the results you expected:**

Installing packages works without any build args passed to `buildah` - something similar to https://docs.docker.com/network/proxy/

**Possible solutions**

* Use `--build-args`
* Use config file

If you need to pass proxy configuration via `--build-args` you need to make your build scripts of project XY aware of the HTTP*/http*-variables thus make them depend on the developers environment. So I would prefer to have a config file.

*Example A*:

`bin/build-no-config`:
~~~
#!/usr/bin/env sh

sudo buildah build-using-dockerfile --file files/latest/Dockerfile --build-arg HTTP_PROXY=http://10.0.2.15:3128 
~~~

*Example A*:

`bin/build-with-config`:
~~~
#!/usr/bin/env sh

sudo buildah build-using-dockerfile --file files/latest/Dockerfile
~~~

`~/.buildah/config.json`:
~~~
{
 "proxies":
 {
   "default":
   {
     "httpProxy": "http://10.0.2.15:3128"
   }
 }
}
~~~

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.0-1.git1ab80bc.fc28.x86_64
```

**Output of `buildah version`:**

```
Version:       1.0
Go Version:    go1.10.1
Image Spec:    1.0.0
Runtime Spec:  1.0.0
Git Commit:    1ab80bc
Built:         Mon May  7 16:01:02 2018
OS/Arch:       linux/amd64

```

**Output of `cat /etc/*release`:**

```
Fedora release 28 (Twenty Eight)
NAME=Fedora
VERSION="28 (Workstation Edition)"
ID=fedora
VERSION_ID=28
PLATFORM_ID="platform:f28"
PRETTY_NAME="Fedora 28 (Workstation Edition)"
ANSI_COLOR="0;34"
CPE_NAME="cpe:/o:fedoraproject:fedora:28"
HOME_URL="https://fedoraproject.org/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
[...]

PR #1510, opened 2019-04-12
auto pass http_proxy to container

jamescassell:
works toward fixing #765

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -198,8 +198,6 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budResults) error {
 	if err != nil {
 		return err
 	}
-        // TODO: move this into CommonBuildOptions() above
-	commonOpts.HttpProxy = iopts.HttpProxy
 
 	if c.Flag("layers").Changed && c.Flag("no-cache").Changed {
 		return errors.Errorf("can only set one of 'layers' or 'no-cache'")
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -198,6 +198,8 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budResults) error {
 	if err != nil {
 		return err
 	}
+        // TODO: move this into CommonBuildOptions() above
+	commonOpts.HttpProxy = iopts.HttpProxy
 
 	if c.Flag("layers").Changed && c.Flag("no-cache").Changed {
 		return errors.Errorf("can only set one of 'layers' or 'no-cache'")
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -230,6 +230,8 @@ option to `false`.  The environment variables passed in include `http_proxy`,
 `https_proxy`, `ftp_proxy`, `no_proxy`, and also the upper case versions of
 those.
 
+Defaults to `true`
+
 **--iidfile** *ImageIDfile*
 
 Write the image ID to the file.
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -222,6 +222,14 @@ Recognized formats include *oci* (OCI image-spec v1.0, the default) and
 Note: You can also override the default format by setting the BUILDAH\_FORMAT
 environment variable.  `export BUILDAH_FORMAT=docker`
 
+**--http-proxy**
+
+By default proxy environment variables are passed into the container if set
+for the buildah process.  This can be disabled by setting the `--http-proxy`
+option to `false`.  The environment variables passed in include `http_proxy`,
+`https_proxy`, `ftp_proxy`, `no_proxy`, and also the upper case versions of
+those.
+
 **--iidfile** *ImageIDfile*
 
 Write the image ID to the file.
--- a/docs/buildah-from.md
+++ b/docs/buildah-from.md
@@ -179,6 +179,8 @@ option to `false`.  The environment variables passed in include `http_proxy`,
 `https_proxy`, `ftp_proxy`, `no_proxy`, and also the upper case versions of
 those.
 
+Defaults to `true`
+
 **--ipc** *how*
 
 Sets the configuration for IPC namespaces when the container is subsequently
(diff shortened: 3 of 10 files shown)


run.go
```diff
@@ -939,6 +939,13 @@ func (b *Builder) configureEnvironment(g *generate.Generator, options RunOptions
 		}
 	}
 
+	for _, envSpec := range []string{"http_proxy", "https_proxy", "no_proxy"} {
```
rhatdan: Need an option --http-proxy, which is defaulted to true to turn this off in buildah run and buildah bud.

buildah bud --http-proxy=false ...

rhatdan: Also I thought you wanted to test upper case.

rhatdan:
bot, add author to whitelist

rhatdan:
bot, test pull request

rhatdan:
Would like a test on this.  Just to make sure the environment variables get passed.  
Please make sure the environment variables do not end up in the container image.

Once you add command line options, you need to add them to man pages and command completions.

run.go
jamescassell: I've pushed an update that includes more proxy vars, including upper case.  The initial push was the subset vars that I've found that work in the most common cases.

Still needs an option, which I haven't figured out how to do yet.

jamescassell:
> Would like a test on this. Just to make sure the environment variables get passed.
> Please make sure the environment variables do not end up in the container image.
> 

These env vars are evaluated each time a 'buildah run' command is executed.  If you unset some of these vars between 'buildah run' commands, they cease to be present during the subsequent 'bulidah run command'

I've also confirmed that these do not persist 'buildah commit', both by inspecting the metadata and by 'podman run --rm -it <image> env'

If you meant automated testing, I'm not sure where to start for that.

> Once you add command line options, you need to add them to man pages and command completions.

I might need a little help adding the command line options.  I'm sure I can figure out the man pages and bash completions.

rhatdan:
Tests are failing, because your second PR was not signed.  It is better if you squash your commits together.

git rebase -i origin
This will load a screen which has the "f" or "fixup" option, which you can use to squash your PRs together.

rhatdan:
As far as adding a flag, first thing you need to do is add a build options.

in buildah.go look for

CommonBuildOptions

Add you HttpProxy boolean option to this.

Now in your existing code wrap it with 
```
if b.CommonBuildOptions.HttpProxy {
}
```
 in cmd/buildah/from.go and cmd/buildah/from.go
Set the HttpProxy values in the section that sets up the BuilderOptions.

Now in pkg/cli/common.go look for FromAndBudResults, add a boolean there.  
Now look for how CPUQuota is handled and add similar calls for HttpProxy.

jamescassell:
I've pushed my (failed) attempts to make the functionality optional.  The result complies, but the option is always disabled and the command line flags are not honored.  I'm likely missing something in connecting the setting of the option and reading of the option.

Side note: I also tested the original functionality of this PR in `buildah bud` and it worked exactly as expected.

pkg/cli/common.go
```diff
@@ -54,6 +54,7 @@ type BudResults struct {
 	DisableContentTrust bool
 	File                []string
 	Format              string
+	HttpProxy    bool
```
rhatdan: You need to run your code changes through `gofmt -w -s`, to be able to pass tests.

cmd/buildah/bud.go
```diff
@@ -272,7 +272,6 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budResults) error {
 		ForceRmIntermediateCtrs: iopts.ForceRm,
 		BlobDirectory:           iopts.BlobCache,
 		Target:                  iopts.Target,
-		HttpProxy:               iopts.HttpProxy,
```
rhatdan: Why are you reverting these.  I believe the original was correct.

jamescassell:
I've gotten the '--http-proxy' option to work for 'buildah run' but not for 'buildah from' or 'buildah bud'.

Likely, I just haven't hooked things up properly.  Seems like there are several layers of indirection going on to get those hooked up.

I'll rebase everything in the end once I've gotten the functionality working; such rebase will ensure code formatting and signed-off-by headers are up to standards.

rhatdan:
@jamescassell You would only see the --http-proxy working when doing a buildah run after you used it in the buildah from.

rhatdan:
@rh-atomic-bot Could you squash all of your commits together

git rebase -i origin

Then switch each commit to a "f" after the fist one.

jamescassell:
ready_for_review

Also, is there an easy way to add this to podman to have similar functionality when running containers there?

pkg/cli/common.go
jamescassell: Fixed

docs/buildah-bud.md
```diff
@@ -222,6 +222,14 @@ Recognized formats include *oci* (OCI image-spec v1.0, the default) and
 Note: You can also override the default format by setting the BUILDAH\_FORMAT
 environment variable.  `export BUILDAH_FORMAT=docker`
 
+**--http-proxy**
+
+By default proxy environment variables are passed into the container if set
```
rhatdan: Add `Defaults to true`

docs/buildah-from.md
```diff
@@ -171,6 +171,14 @@ The [username[:password]] to use to authenticate with the registry if required.
 If one or both values are not supplied, a command line prompt will appear and the
 value can be entered.  The password is entered without echo.
 
+**--http-proxy**
+
+By default proxy environment variables are passed into the container if set
```
rhatdan: Add `Defaults to true.`

rhatdan:
The tests don't like the naming of HttpProxy
```
buildah.go:286:2:warning: struct field HttpProxy should be HTTPProxy (golint)
Makefile:66: recipe for target 'validate' failed
make: *** [validate] Error 1
```
Rename it to make golint happy.

Also add --http-proxy options to contrib/completions/bash/buildah

And squash your PRs into one PR.

rhatdan:
@jamescassell We will vendor this code into podman and that will handle the podman build case.  But we will need work to get this into 
podman run and podman create.

jamescassell:
Closing since a squash was requested.  See #1525

PR #1514, opened 2019-04-15
WIP - Update rpm spec file in contrib to pass rpmlint

pixdrift:
Recent changes to the RPM spec file resulted in failure to pass rpmlint validation which was breaking `rpmbuild`. This update fixes minor issues so the spec file once again passes `rpmlint`.

### DIFF
--- a/contrib/rpm/buildah.spec
+++ b/contrib/rpm/buildah.spec
@@ -26,7 +26,7 @@
 
 Name:           buildah
 # Bump version in buildah.go too
-Version:        1.8-dev
+Version:        1.8dev
 Release:        1.git%{shortcommit}%{?dist}
 Summary:        A command line tool used to creating OCI Images
 License:        ASL 2.0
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@TomSweeneyRedHat PTAL

pixdrift:
I am concerned this is going to impact the release script which does string replacement in the spec file off buildah version. As I started digging, it looked like it went deeper than just the spec.

The primary problem is that having the hyphen in the version isn't valid, everything else is cosmetic (changelog dates need to be valid to pass lint).

@TomSweeneyRedHat thoughts on this?

Happy to move it back to WIP until we figure out approach.

TomSweeneyRedHat:
bot, retest this please.

TomSweeneyRedHat:
@pixdrift has some spec changed to disallow a hyphen/dash (-) in a version number somewhere?  I'd like to keep *-dev if possible.

contrib/rpm/buildah.spec
```diff
@@ -26,7 +26,7 @@
 
 Name:           buildah
 # Bump version in buildah.go too
-Version:        1.8-dev
```
TomSweeneyRedHat: The version listed in buildah.go should match this.

pixdrift: What I am proposing is that the '-dev' become 'dev' in the spec/rpm file only.

Below quote is from Max RPM which was published in 2000... so it's been this way for about 20 years.. probably more 😁 

RPM spec reference here:
http://ftp.rpm.org/max-rpm/ch-rpm-file-format.html
> The version number is normally taken verbatim from the package's version. The only restriction placed on the version is that it cannot contain a dash "-".

TomSweeneyRedHat: Guess I was showing my rpm spec lack of knowledge off well!
I'm OK, but now wild with dev smashed up against the number.  Could we change it to 1.8.dev?  or is that no good either?

If possible, I'd like to keep the version the same in both this file and buildah.go.

pixdrift: Yeah '.' should be good, and there is precedent for this in the RHEL repos, not with 'dev' specifically, but string content eg. commit id's.

rhatdan:
@pixdrift @TomSweeneyRedHat What is up with this PR?  Is it still needed?

TomSweeneyRedHat:
I'm going to close this for now.  @pixdrift feel free to reopen or spin up a new PR if you think we need changes still.

PR #1515, opened 2019-04-15
[WIP] Commit if last stage, or stage used in future From

rhatdan:
Currently we commit every stage when the last command happens.
We should not need to do this unless it is the last stage
or this stage is used in a future From command.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -872,7 +872,7 @@ func (s *StageExecutor) getImageRootfs(ctx context.Context, stage imagebuilder.S
 }
 
 // Execute runs each of the steps in the stage's parsed tree, in turn.
-func (s *StageExecutor) Execute(ctx context.Context, stage imagebuilder.Stage, base string) (imgID string, ref reference.Canonical, err error) {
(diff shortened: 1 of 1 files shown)


rhatdan:
@nalin is that what you were thinking?

nalind:
Unfortunately, we need those intermediate images to always be committed so that they can be found in the cache later, per https://github.com/containers/buildah/pull/1473#issuecomment-479592881.

When we conditionalize on whether or not the stage's name is the same as stage's index, we can get an incorrect result for instructions like `COPY --from=0`, which I'd rather avoid.

imagebuildah/build.go
```diff
@@ -1538,8 +1542,22 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
...
+		// a base image later
+		forceCommit := stageIndex == len(stages)-1
+		if !forceCommit {
+			for index, value := range b.stages {
+				if value == stageExecutor {
+					continue
+				}
+				if index != stage.Name {
+					forceCommit = true
+				}
```
nalind: I think that when we have multiple stages, this loop ends up always setting `forceCommit` to `true`?

rhatdan:
@nalind patch is better https://github.com/containers/buildah/pull/1517

nalind:
Thanks!

PR #1517, opened 2019-04-15
imagebuildah: only commit images for intermediate stages if we have to

nalind:
In single-layer mode, if the stage's image isn't going to be used as the basis for a later stage, and it isn't the final stage, don't bother committing it.

In multiple-layer mode, if we're handling the final instruction for a stage, and we've found a suitable image for the stage in the cache, if the container's root filesystem isn't going to be used as the source for a COPY --from instruction in a later stage, don't bother rebasing the working container to match the image.

It turns out that the cache expectation is that we always produce an image at the end of a stage, so remove the TODO comment that suggested that we'd be able to skip it in some cases.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -215,6 +215,8 @@ type Executor struct {
 	forceRmIntermediateCtrs        bool
 	imageMap                       map[string]string           // Used to map images that we create to handle the AS construct.
 	containerMap                   map[string]*buildah.Builder // Used to map from image names to only-created-for-the-rootfs containers.
+	baseMap                        map[string]bool             // Holds the names of every base image, as given.
+	rootfsMap                      map[string]bool             // Holds the names of every stage whose rootfs is referenced in a COPY or ADD instruction.
 	blobDirectory                  string
 	excludes                       []string
 	unusedArgs                     map[string]struct{}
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

imagebuildah/build.go
```diff
@@ -1510,6 +1522,45 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 	}
 	defer cleanup()
 
+	// Build maps of every named base image and every referenced stage root
+	// filesystem.  Individual stages can use them to determine whether or
+	// not they can skip certain steps near the end of their stages.
+	for _, stage := range stages {
+		node := stage.Node
+		for node != nil {
+			for _, node := range node.Children {
```
vrothberg: Can we rename the loop's run variable to `child`? The code is correct but it looked very suspicious, so I had parse it a few times to be sure to understand it's correct.

h-vetinari:
Sounds like this might (help) fix #1456?

imagebuildah/build.go
TomSweeneyRedHat: or if you really don't like child, some other variable name, or perhaps change node to stageNode at line 1529

nalind: `child` is fine.  Changed it.

rhatdan:
@rh-atomic-bot r+

nalind:
@rh-atomic-bot retry

Issue #1219, opened 2018-12-05
Dangling symlink traversal failure when using bud with layer cache


pkubatrh:
**Description**
Image build fails when building an image using the layer cache from a Dockerfile while adding a symlink that will be dangling at build time due to the target not being located in the context directory.

**Steps to reproduce the issue:**
1.cat << EOF >Dockerfile
FROM fedora
COPY run-mysqld /tmp/
EOF
2. ln -s /bin/run-mysqld run-mysqld
3. buildah bud --layers .
4. buildah bud --layers .

**Describe the results you received:**
Second `buildah bud` fails with:
```
error checking if cached image exists from a previous build: error checking if copied files match: error resolving symlinks and comparing modified times: "/tmp/reproducer/run-mysqld": error checking if modified time of resolved symbolic link is greater: error walking file tree "/tmp/reproducer/run-mysqld": error evaluating symlink "run-mysqld": lstat /bin: no such file or directory
: exit status 1
```

**Describe the results you expected:**
Buildah sees that the symlink is dangling and does not try to check the target

**Output of `rpm -q buildah` or `apt list buildah`:**

buildah-1.5-2.gite94b4f9.fc28.x86_64

**Output of `buildah version`:**

Version:         1.5
Go Version:      go1.10.5
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.4.0
libcni Version:  
Git Commit:      
Built:           Thu Jan  1 01:00:00 1970
OS/Arch:         linux/amd64

umohnani8:
@TomSweeneyRedHat no it wouldn't. It is failing because bud expects the symlink to exist so it can compare the timestamps to know whether to use from cache or not. Didn't think of a dangling symlink case. I guess the fix would be to ignore it if it is dangling and run the build process again i.e not use cached layers? Not sure what docker does.

rhatdan:
@pkubatrh Does Docker handle this?

rhatdan:
@umohnani8 @TomSweeneyRedHat @pkubatrh Is this still an issue or has it been fixed?

pkubatrh:
@rhatdan Still seeing it in latest master

```
$ git rev-parse HEAD
36605c29eb258caa9720dbd42622269a81dd1bfa
$ cat << EOF >Dockerfile
FROM fedora
COPY test /tmp/
EOF
$ mkdir test; ln -s /bin/run-mysqld test/run-mysqld
[...]

PR #1519, opened 2019-04-15
Don't fail bud --layers on dangling symlink

umohnani8:
If there is a dangling symlink when building with --layers,
don't fail and continue the build process.
It would fail because it would try to stat the file to check
whether it has been modified since the previous build. And since
it would be a dangling symlink, the stat would fail hence causing
the build to fail.

Fixes https://github.com/containers/buildah/issues/1219

### DIFF
--- a/imagebuildah/chroot_symlink.go
+++ b/imagebuildah/chroot_symlink.go
@@ -154,6 +154,9 @@ func modTimeIsGreater(rootdir, path string, historyTime string) (bool, error) {
 		if info.Mode()&os.ModeSymlink == os.ModeSymlink {
 			// Evaluate any symlink that occurs to get updated modified information
 			resolvedPath, err := filepath.EvalSymlinks(path)
+			if err != nil && os.IsNotExist(err) {
+				return errors.Wrapf(errDanglingSymlink, "%q", path)
+			}
 			if err != nil {
 				return errors.Wrapf(err, "error evaluating symlink %q", path)
 			}
@@ -169,7 +172,12 @@ func modTimeIsGreater(rootdir, path string, historyTime string) (bool, error) {
 		}
 		return nil
 	})
+
 	if err != nil {
+		// if error is due to dangling symlink, ignore error and return nil
+		if errors.Cause(err) == errDanglingSymlink {
+			return false, nil
+		}
 		return false, errors.Wrapf(err, "error walking file tree %q", path)
 	}
 	return timeIsGreater, err
--- a/imagebuildah/errors.go
+++ b/imagebuildah/errors.go
@@ -1,0 +1,7 @@
+package imagebuildah
+
+import "errors"
+
+var (
+	errDanglingSymlink = errors.New("error evaluating dangling symlink")
+)
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -163,6 +163,28 @@ load helpers
   buildah rmi -a -f
 }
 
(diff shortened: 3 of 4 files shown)


umohnani8:

imagebuildah/build.go
```diff
@@ -28,7 +28,7 @@ import (
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/archive"
 	docker "github.com/fsouza/go-dockerclient"
-	"github.com/opencontainers/image-spec/specs-go/v1"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
```
rhatdan: Why is this necessary?

rhatdan:
Other then the question LGTM

imagebuildah/build.go
umohnani8: vscode did that automatically lol, fixed.

rhatdan:
lgtm
@nalind @vrothberg @giuseppe @TomSweeneyRedHat PTAL

imagebuildah/chroot_symlink.go
```diff
@@ -19,6 +19,10 @@ const (
 	maxSymlinksResolved    = 40
 )
 
+var (
+	errDanglingSymlink = errors.New("error evaluating dangling symlink")
```
nalind: Creating this with the `github.com/pkg/errors` package adds a traceback that points to this line whenever we return the error.  Can you change this to be created with the standard library's `errors` package, either by importing it under a different name or making it the sole definition in an `errors.go` that uses the standard `errors` package?

imagebuildah/chroot_symlink.go
```diff
@@ -154,6 +158,9 @@ func modTimeIsGreater(rootdir, path string, historyTime string) (bool, error) {
 		if info.Mode()&os.ModeSymlink == os.ModeSymlink {
 			// Evaluate any symlink that occurs to get updated modified information
 			resolvedPath, err := filepath.EvalSymlinks(path)
+			if err != nil && os.IsNotExist(err) {
+				return errDanglingSymlink
```
nalind: If `errDanglingSymlink` becomes a standard-library-produced `error`, this will probably want to be wrapped with `errors.Wrapf()`...

imagebuildah/chroot_symlink.go
```diff
@@ -169,7 +176,12 @@ func modTimeIsGreater(rootdir, path string, historyTime string) (bool, error) {
 		}
 		return nil
 	})
+
 	if err != nil {
+		// if error is due to dangling symlink, ignore error and return nil
+		if err == errDanglingSymlink {
```
nalind: ... and this will need to compare with `errors.Cause(err)`.

imagebuildah/chroot_symlink.go
umohnani8: Fixed. Created a new file called errors.go and used the standard library's error package.

imagebuildah/chroot_symlink.go
umohnani8: Fixed.

imagebuildah/chroot_symlink.go
umohnani8: Fixed

umohnani8:
@nalind fixed the issues, PTAL

rhatdan:
@rh-atomic-bot r+

Issue #1391, opened 2019-03-07
caches: `buildah bud --layers` sucks on --build-arg=VAR=VAL


praiskup:
**Description**

Changing the `--build-arg` option doesn't invalidate `--layers` caches.

**Steps to reproduce the issue:**

1. provide dockerfile:

    ```
    $ cat Dockerfile
    FROM busybox
    ARG user
    RUN echo $user
    ```

2. first run (no caches => no hit => correct)

    ```
    $ buildah bud --layers --build-arg=user=0 .
    STEP 1: FROM busybox
    Getting image source signatures
    Copying blob 697743189b6d: 738.18 KiB / 738.18 KiB [========================] 0s
    Copying config d8233ab899d4: 1.46 KiB / 1.46 KiB [==========================] 0s
    Writing manifest to image destination
    Storing signatures
    STEP 2: ARG user
    --> 61e60952494142fd9d32723c5eb0dad8adb526ff18f991642d66aed5dc783627
    STEP 3: FROM 61e60952494142fd9d32723c5eb0dad8adb526ff18f991642d66aed5dc783627
    STEP 4: RUN echo $user
    0
    --> 6dfeb0be1b508772f699f32606a04c5f5fdc2909b1684f247f9c2e97de3e33dd
    STEP 5: COMMIT
    ```

3. repeat the build with the same --build-arg=user=0 (cache hit => correct)

    ```
    $ buildah bud --layers --build-arg=user=0 .
    STEP 1: FROM busybox
    STEP 2: ARG user
    --> Using cache 61e60952494142fd9d32723c5eb0dad8adb526ff18f991642d66aed5dc783627
    STEP 3: FROM 61e60952494142fd9d32723c5eb0dad8adb526ff18f991642d66aed5dc783627
    STEP 4: RUN echo $user
    --> Using cache 6dfeb0be1b508772f699f32606a04c5f5fdc2909b1684f247f9c2e97de3e33dd
    STEP 5: COMMIT 
    ```

4. repeat the build with changed --build=arg=1 (cache hit => incorrect!)

    ```
    $ buildah bud --layers --build-arg=user=1 .
    STEP 1: FROM busybox
    STEP 2: ARG user
    --> Using cache 61e60952494142fd9d32723c5eb0dad8adb526ff18f991642d66aed5dc783627
    STEP 3: FROM 61e60952494142fd9d32723c5eb0dad8adb526ff18f991642d66aed5dc783627
    STEP 4: RUN echo $user
[...]

Issue #1507, opened 2019-04-11
BUG: caching does not work correctly with ARG


h-vetinari:
[Continuing from #1456; containers/libpod#2250]

Using the following dockerfile (say, `podman_test.dockrf`):
```
FROM docker.io/library/ubuntu:bionic
ARG DEBIAN_FRONTEND=noninteractive

#RUN apt-get update && apt-get install tzdata -y --no-install-recommends

CMD ["bash"]
```
and build using `sudo podman -t podman_test -f podman_test.dockrf .`.

Afterwards, uncomment the `RUN` line and build again. The package is chosen to normally require user interaction, which would be suppressed at installation time by `DEBIAN_FRONTEND=noninteractive`. But while the build says:
```
$ sudo podman build -f podman-cache.dockrf .
STEP 1: FROM docker.io/library/ubuntu:bionic
STEP 2: ARG DEBIAN_FRONTEND=noninteractive
--> Using cache 0d0d6081636f891eefe257781e2d8445912eb2d321d3d19446cb28b1e6e32f4d
STEP 3: FROM 0d0d6081636f891eefe257781e2d8445912eb2d321d3d19446cb28b1e6e32f4d
STEP 4: RUN apt-get update && apt-get install tzdata -y --no-install-recommends
[...]
```

it ends up running into an error that's plainly due to `DEBIAN_FRONTEND` not being set:
```
debconf: (Can't locate Term/ReadLine.pm in @INC (you may need to install the Term::ReadLine module) (@INC contains: /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.26.1 /usr/local/share/perl/5.26.1 /usr/lib/x86_64-linux-gnu/perl5/5.26 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl/5.26 /usr/share/perl/5.26 /usr/local/lib/site_perl /usr/lib/x86_64-linux-gnu/perl-base) at /usr/share/perl5/Debconf/FrontEnd/Readline.pm line 7.)
debconf: falling back to frontend: Teletype
Configuring tzdata
------------------

Please select the geographic area in which you live. Subsequent configuration
questions will narrow this down by presenting a list of cities, representing
the time zones in which they are located.

  1. Africa      4. Australia  7. Atlantic  10. Pacific  13. Etc
  2. America     5. Arctic     8. Europe    11. SystemV
  3. Antarctica  6. Asia       9. Indian    12. US
Geographic area:
```

[...]

Issue #1503, opened 2019-04-10
BUG: global ARG not shown as build step


h-vetinari:
[Continuing from #1456; containers/libpod#2250]

This is a comparatively small bug, but confusing nevertheless. In multistage builds, it may be necessary to re-use a build-arg for different stages (I'm using the generic `DEBIAN_FRONTEND` in the example below), but setting it per stage would lead to an error: `ARG requires exactly one argument definition`

Docker solves this by being able to specify a global ARG *before* the first FROM, cf. [here](https://github.com/moby/moby/issues/37622#issuecomment-412101935). Podman/buildah *execute*  this correctly, but does not show the ARG before the first FROM as a build step.

Say we have the following dockerfile, and call it `podman_test.dockrf`:
```
ARG DEBIAN_FRONTEND=noninteractive

FROM docker.io/library/ubuntu:bionic as base
ARG DEBIAN_FRONTEND
RUN echo $DEBIAN_FRONTEND

RUN apt-get update && apt-get install curl -y --no-install-recommends

FROM base as main
ARG DEBIAN_FRONTEND
RUN echo $DEBIAN_FRONTEND

RUN apt-get update && apt-get install wget -y --no-install-recommends

CMD ["bash"]
```

Then the build log is
```
$ sudo podman build -t podman_test -f podman_test.dockrf .
STEP 1: FROM docker.io/library/ubuntu:bionic AS base
STEP 2: ARG DEBIAN_FRONTEND
--> 8dbc2806a4f15de339706b0512a21e8a7453da79a07101900b0afe077e029c10
STEP 3: FROM 8dbc2806a4f15de339706b0512a21e8a7453da79a07101900b0afe077e029c10 AS base
STEP 4: RUN echo $DEBIAN_FRONTEND
noninteractive
--> 4c48e1473c102e397ff4381bae5bfdba811313894c1c159a525e19975e0eb9a0
STEP 5: FROM 4c48e1473c102e397ff4381bae5bfdba811313894c1c159a525e19975e0eb9a0 AS base
STEP 6: RUN apt-get update && apt-get install curl -y --no-install-recommends
[snip]
--> abf1a7cf75a5da41939ebb1ac25e74731ea9ce3cf81815466bdc960bd1fa6be8
STEP 7: COMMIT podman_test
--> abf1a7cf75a5da41939ebb1ac25e74731ea9ce3cf81815466bdc960bd1fa6be8
[...]

PR #1521, opened 2019-04-16
Keep track of any build-args used during buildah bud --layers

umohnani8:
Add any build-args used in the build process to the layers history
for future comparison when building with --layers.

Fixes https://github.com/containers/buildah/issues/1391

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -28,7 +28,7 @@ import (
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/archive"
 	docker "github.com/fsouza/go-dockerclient"
-	"github.com/opencontainers/image-spec/specs-go/v1"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/openshift/imagebuilder"
 	"github.com/openshift/imagebuilder/dockerfile/parser"
@@ -220,6 +220,7 @@ type Executor struct {
 	blobDirectory                  string
 	excludes                       []string
 	unusedArgs                     map[string]struct{}
+	buildArgs                      map[string]string
 }
 
 // StageExecutor bundles up what we need to know when executing one stage of a
@@ -662,6 +663,7 @@ func NewExecutor(store storage.Store, options BuildOptions, mainNode *parser.Nod
 		rootfsMap:                      make(map[string]bool),
 		blobDirectory:                  options.BlobDirectory,
 		unusedArgs:                     make(map[string]struct{}),
+		buildArgs:                      options.Args,
 	}
 	if exec.err == nil {
 		exec.err = os.Stderr
@@ -949,7 +951,7 @@ func (s *StageExecutor) Execute(ctx context.Context, stage imagebuilder.Stage, b
 			// squash the contents of the base image.  Whichever is
 			// the case, we need to commit() to create a new image.
 			logCommit(s.output, -1)
-			if imgID, ref, err = s.commit(ctx, ib, getCreatedBy(nil), s.output); err != nil {
+			if imgID, ref, err = s.commit(ctx, ib, s.executor.getCreatedBy(nil), s.output); err != nil {
 				return "", nil, errors.Wrapf(err, "error committing base container")
 			}
 		} else {
@@ -1021,7 +1023,7 @@ func (s *StageExecutor) Execute(ctx context.Context, stage imagebuilder.Stage, b
 				// instruction in the history that we'll write
 				// for the image when we eventually commit it.
 				now := time.Now()
-				s.builder.AddPrependedEmptyLayer(&now, getCreatedBy(node), "", "")
+				s.builder.AddPrependedEmptyLayer(&now, s.executor.getCreatedBy(node), "", "")
 				continue
 			} else {
 				// This is the last instruction for this stage,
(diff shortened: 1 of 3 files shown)


umohnani8:

rhatdan:
LGTM

h-vetinari:
Not sure if the PR somehow separates the treatment of `--build-arg` and plain `ARG`, but maybe it also solves #1507.

rhatdan:
@nalind @giuseppe @TomSweeneyRedHat @vrothberg PTAL

umohnani8:
@h-vetinari I will look into #1507 :)

h-vetinari:
@umohnani8: @h-vetinari I will look into #1507 :)

Cool! :)
Maybe also tangentially related: #1503

imagebuildah/build.go
```diff
@@ -1250,12 +1256,22 @@ func getCreatedBy(node *parser.Node) string {
...
+			buildArgs := b.getBuildArgs()
+			// If a previous image was built with some build-args but the new build process doesn't have any build-args
+			// specified, so compare the lengths of the old instruction with the current one
+			if buildArgs == "" && len(history[i].CreatedBy) > len(instruction)+11 {
+				return false
+			}
+			// There are build-args, so check if anything with the build-args has changed
+			if buildArgs != "" && !strings.Contains(history[i].CreatedBy, buildArgs) {
+				return false
+			}
```
TomSweeneyRedHat: Nice!  Rambling note. I was just looking at this function this weekend in regards to #1391 and found it wasn't working there as I thought it should, although I was approaching it by adding the value from the build args to the applicable arg.  So I changed 'ARG user' to "ARG user=0" if --build-arg user=0 was passed in.  I was tripping over the string contains below with that.  

So end of ramble, will this handle the case where --build-args user=0 on the first run and --build-args=1 on the second run.  Will the code think they should be fresh layers if --layers=true or will will it reuse the cache?

umohnani8: yup, it will handle that case. I am adding any build-args used during the process to the history of that layer, as docker does as well. So this line compares the build-args if build-args are set during the process.
I also have a check where the image was built with a build-arg, but in the second run it wasn't. And when that happens, we want a fresh layer instead of cache.

TomSweeneyRedHat: Thanks @umohnani8 !  The code LGTM on a quick read, but I want to test/play with it.  Looks like Red Hat CI has gone off the rails.  Might be worth a new push.

umohnani8: thanks! and re pushed, hopefully the CI will not die again

imagebuildah/build.go
```diff
@@ -1250,12 +1256,22 @@ func getCreatedBy(node *parser.Node) string {
...
+func (b *Executor) historyMatches(children []*parser.Node, history []v1.History) bool {
 	i := len(history) - 1
 	for j := len(children) - 1; j >= 0; j-- {
 		instruction := children[j].Original
 		if children[j].Value == "run" {
 			instruction = instruction[4:]
+			buildArgs := b.getBuildArgs()
+			// If a previous image was built with some build-args but the new build process doesn't have any build-args
+			// specified, so compare the lengths of the old instruction with the current one
+			if buildArgs == "" && len(history[i].CreatedBy) > len(instruction)+11 {
```
TomSweeneyRedHat: Might want to document and/or create a variable for the 11.

TomSweeneyRedHat: Or, assuming --build-arg user=0, you're comparing the values of 
instruction: [echo $user]
history[i].CreatedBy: [user=0 /bin/sh -c echo $user]

Would it be better to just check and see if history[i].CreatedBy started with /bin/sh or not?

I' m on the fence.

TomSweeneyRedHat:
Code looks fine, but I've spotted a slight difference between Docker and this that we probably ought to touch up.  If you do:
```
# cat ~/Dockerfile.withuser
FROM busybox
ARG user
ARG user2
RUN echo $user
RUN echo $user2
ENV MYBAD=bad

buildah bud --layers -t tom -f ~/Dockerfile.withuser --build-arg user=0 --build-arg user2=3 .
docker build -t tom -f ~/Dockerfile.withuser --build-arg user=1 --build-arg user2=2 .
```

Buildah history shows:
```
buildah inspect tom
...
      {
            "created": "2019-04-17T18:19:15.705509758Z",
            "created_by": "user=0 user2=3 /bin/sh -c echo $user"
        },
        {
            "created": "2019-04-17T18:19:16.844597882Z",
            "created_by": "user=0 user2=3 /bin/sh -c echo $user2"
        },

```

Docker history shows:
```
IMAGE               CREATED             CREATED BY                                      SIZE                COMMENT
420c228b495e        5 seconds ago       /bin/sh -c #(nop)  ENV MYBAD=bad                0B                  
14c4e5a79540        5 seconds ago       |2 user2=2 user=1 /bin/sh -c echo $user2        0B                  
52affc18d01f        6 seconds ago       |2 user2=2 user=1 /bin/sh -c echo $user         0B                  
ce34f4d6e0c1        7 seconds ago       /bin/sh -c #(nop)  ARG user2                    0B                  
9480e5de1551        8 days ago          /bin/sh -c #(nop)  ARG user                     0B                  
af2f74c517aa        2 weeks ago         /bin/sh -c #(nop)  CMD ["sh"]                   0B                  
<missing>           2 weeks ago         /bin/sh -c #(nop) ADD file:6051b0ebe4098ccbb…   1.2MB        
```
not the pipe 2 at the start of the echo history commands.  When I ran it with one build-arg it was "|1" and with 3 args "|3".  Could we add that?

What's of head scratching interest is for the 2 args case the switched the order of the  build-args in the history.  For 3 args, they looked to have moved that first argument to the last position and then kept the others in the order received.

```
5f3d788341bd        5 seconds ago       |3 user2=2 user3=3 user=1 /bin/sh -c echo $u…   0B                  

```

This is with Docker Version:           18.09.0

IDK for sure if we need to change the history order to match, but we probably ought to add the pipe and number of args if possible.

Other than that, really nice piece of code craftsmanship here @umohnani8

umohnani8:
@TomSweeneyRedHat thanks for testing it! And great catch, I will rework it to include the pipe and number of args :)

rhatdan:
@rhatdan Needs a rebase.

umohnani8:
@TomSweeneyRedHat fixed it up and rebased.

TomSweeneyRedHat:
LGTM, nice work @umohnani8

TomSweeneyRedHat:
@rh-atomic-bot r+

PR #1522, opened 2019-04-16
Do not state that buildah always requires root

danmack:
It is very possible today to use buildah as a non-root user (at least on Fedora 29 and probably CentOS 7.6, etc.).  This changes the wording to allow for that nuance.


danmack:
I'll be back after adhering to the contributing guide.  closing this out for now as I have no idea why the CI build failed.

PR #1526, opened 2019-04-17
Add registry mirror support

saschagrunert:
Vendor in the latest containers/image to support registry mirrors.
github.com/etcd-io/bbolt has to be vendored as direct dependency.

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -112,7 +112,7 @@ func ResolveName(name string, firstRegistry string, sc *types.SystemContext, sto
 	}
 	for _, registry := range searchRegistries {
 		if !registry.Blocked {
-			registries = append(registries, registry.URL)
+			registries = append(registries, registry.Prefix)
 		}
 	}
 	searchRegistriesAreEmpty := len(registries) == 0
--- a/vendor.conf
+++ b/vendor.conf
@@ -3,7 +3,8 @@ github.com/blang/semver v3.5.0
 github.com/BurntSushi/toml v0.2.0
 github.com/containerd/continuity 004b46473808b3e7a4a3049c20e4376c91eb966d
 github.com/containernetworking/cni v0.7.0-rc2
-github.com/containers/image f52cf78ebfa1916da406f8b6210d8f7764ec1185
+github.com/containers/image ff926d3c79684793a2135666a2cb738f44ba33dc
+github.com/etcd-io/bbolt v1.3.2
 github.com/vbauerster/mpb v3.3.4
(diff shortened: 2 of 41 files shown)


util/util.go
```diff
@@ -115,7 +115,7 @@ func ResolveName(name string, firstRegistry string, sc *types.SystemContext, sto
 	}
 	for _, registry := range searchRegistries {
 		if !registry.Blocked {
-			registries = append(registries, registry.URL)
+			registries = append(registries, registry.Location)
```
mtrmac: This should be `registry.Prefix` actually, so that later lookups of the result of using this registry find and use the same registry entry.

saschagrunert: Alright, I changed it.

rhatdan:
@saschagrunert Now that the PR is merged, what do you want to do with this PR?

saschagrunert:
@rhatdan The changes introduced with this PR are still needed when vendoring the latest contianers/image commit. On the other side we could also wait for the next released version.

rhatdan:
Ok, I want to get a released version of containers/image out soon, so we can vendor it into buildah and release a new version of buildah.

saschagrunert:
> Ok, I want to get a released version of containers/image out soon, so we can vendor it into buildah and release a new version of buildah.

Alright, so I guess we can leave this open and just adapt the vendored version after a release. Please note that this new version of containers/image contains not backwards compatible changes, like the renaming of `URL` to `Location` in sysregistriesv2. So I guess it is time for 2.0.0 there. :innocent:

vrothberg:
Let's wait with releasing a new c/image until the remaining issues are ironed out. I don't think we do us a favor promoting the mirror support in a rush and forcing users to change existing deployments with hypothetical follow up releases.

rhatdan:
Yes I don't think we will vendor in an update for container/image before the next buildah release.  That way we can keep it simple.  If we decide not to rebase what do we loose?

vrothberg:
* docs: containers-transports(5) manpage
* Add registry mirror support to docker client
* Switch to github.com/etcd-io/bbolt
* storage: fix UID detection in a namespace
* blobinfocache: correctly detect the UID in a namespace
* bump version to 1.7.0-dev
* release v1.6.0
* pkg/docker/config: Remove GetUserLoggedIn

This should not affect Buildah (famous last words?)

rhatdan:
Closing this because https://github.com/containers/buildah/pull/1634 supercedes it.

PR #1528, opened 2019-04-17
Use errors.Cause() when checking for IsNotExist errors

nalind:
Use `errors.Cause()` when asking if `os.IsNotExist()` recognizes an error result from any call that isn't into the standard library or the `golang.org/x/sys` package.


rhatdan:
LGTM

vrothberg:
LGTM

PR #1538, opened 2019-04-22
commit: always set a parent ID

nalind:
Always set a parent ID when we go to commit an image, whether it's as part of build-using-dockerfile or our "commit" CLI.  Coerce the parent image's ID directly into the value that we use instead of digesting it
again.


rhatdan:
LGTM

vrothberg:
@rh-atomic-bot r+

PR #1539, opened 2019-04-23
commit: commit on every instruction, but not always with layers

nalind:
When building an image with multiple layers, go back to committing images for instructions for which we previously wouldn't bother committing an image, but create them without adding a new layer.

This violates some assumptions that we currently make elsewhere, as it's possible for an image that's derived from a base image to add no layers relative to the base image, when previously it was always the case that we'd add at least one whenever we committed it.  This would already have been the case with images built with `docker build`, so we have to make some adjustments to our CLI logic to not be confused by it.

### DIFF
--- a/cmd/buildah/common.go
+++ b/cmd/buildah/common.go
@@ -149,7 +149,7 @@ func openImage(ctx context.Context, sc *types.SystemContext, store storage.Store
 	return builder, nil
 }
 
-func getDateAndDigestAndSize(ctx context.Context, image storage.Image, store storage.Store) (time.Time, string, int64, error) {
+func getDateAndDigestAndSize(ctx context.Context, store storage.Store, image storage.Image) (time.Time, string, int64, error) {
 	created := time.Time{}
 	is.Transport.SetStore(store)
 	storeRef, err := is.Transport.ParseStoreReference(store, image.ID)
--- a/cmd/buildah/common.go
+++ b/cmd/buildah/common.go
@@ -13,6 +13,7 @@ import (
 	"github.com/containers/image/types"
 	"github.com/containers/storage"
 	digest "github.com/opencontainers/go-digest"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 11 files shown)


image.go
```diff
@@ -636,7 +642,7 @@ func (i *containerImageSource) GetBlob(ctx context.Context, blob types.BlobInfo,
 	return ioutils.NewReadCloserWrapper(layerFile, closer), size, nil
 }
 
-func (b *Builder) makeImageRef(manifestType string, exporting bool, squash bool, blobDirectory string, compress archive.Compression, historyTimestamp *time.Time, omitTimestamp bool) (types.ImageReference, error) {
+func (b *Builder) makeImageRef(manifestType string, exporting bool, squash bool, emptyLayer bool, blobDirectory string, compress archive.Compression, historyTimestamp *time.Time, omitTimestamp bool) (types.ImageReference, error) {
```
TomSweeneyRedHat: So many function params.  Not now, but perhaps consider a new struct to pass instead?

rhatdan:
LGTM

cmd/buildah/common.go
```diff
@@ -206,63 +207,152 @@ func defaultFormat() string {
...
+		if a[i].Created == nil && b[i].Created != nil {
+			break
+		}
+		if a[i].Created != nil && b[i].Created == nil {
+			break
+		}
+		if a[i].Created != nil && b[i].Created != nil && !a[i].Created.Equal(*(b[i].Created)) {
+			break
+		}
+		if a[i].CreatedBy != b[i].CreatedBy {
```
TomSweeneyRedHat: I think you could drop the first two checks and add the below here.
```
	if a[i].Created == nil || b[i].Created == nil {
			break
	}
```
But I do have "it's almost dinner time brain".

TomSweeneyRedHat:
LGTM and passes baseline tests.  Would like a @vrothberg looksee if this isn't critical to get in tonight.

cmd/buildah/common.go
rhatdan: No I think the code is correct.  But I believe if both are nil then he wants to continue checking.

TomSweeneyRedHat: oh yeah, if both are nil then they fall through and probably should.  Ignore me.

TomSweeneyRedHat:
Ping @vrothberg

TomSweeneyRedHat:
@nalind, looks like you need a rebase on this one.

image.go
nalind: Have to rebase anyway, adding the change.

cmd/buildah/rmi.go
```diff
@@ -197,7 +197,7 @@ func deleteImages(ctx context.Context, systemContext *types.SystemContext, store
 				lastError = errors.Errorf("unable to delete %q (cannot be forced) - image has dependent child images", image.ID)
 				continue
 			}
-			id, err := removeImage(image, store)
+			id, err := removeImage(ctx, systemContext, image, store)
```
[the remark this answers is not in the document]
nalind: Have to rebase anyway, adding the change.

nalind:
Rebased.

rhatdan:
@rh-atomic-bot r+

vrothberg:
> Ping @vrothberg

Apologies for missing that. I went through ~300 mails yesterday this one slipped through :see_no_evil:

vrothberg:
> > Ping @vrothberg
> 
> Apologies for missing that. I went through ~300 mails yesterday and this one slipped through

rhatdan:
Well since I created most of those emails, Apology accepted.  :^)

PR #1542, opened 2019-04-23
run: check if SELinux is enabled

nalind:
When we're built with support for SELinux, refrain from setting process and mount labels if SELinux isn't detected as enabled at runtime.

### DIFF
--- a/selinux.go
+++ b/selinux.go
@@ -4,9 +4,12 @@ package buildah
 
 import (
 	"github.com/opencontainers/runtime-tools/generate"
+	selinux "github.com/opencontainers/selinux/go-selinux"
 )
 
 func setupSelinux(g *generate.Generator, processLabel, mountLabel string) {
-	g.SetProcessSelinuxLabel(processLabel)
-	g.SetLinuxMountLabel(mountLabel)
+	if processLabel != "" && selinux.GetEnabled() {
+		g.SetProcessSelinuxLabel(processLabel)
+		g.SetLinuxMountLabel(mountLabel)
+	}
 }
--- a/chroot/selinux.go
+++ b/chroot/selinux.go
@@ -13,7 +13,7 @@ import (
 // setSelinuxLabel sets the process label for child processes that we'll start.
 func setSelinuxLabel(spec *specs.Spec) error {
 	logrus.Debugf("setting selinux label")
-	if spec.Process.SelinuxLabel != "" && selinux.EnforceMode() != selinux.Disabled {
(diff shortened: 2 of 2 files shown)


rhatdan:
Are you seeing an issue with this?

selinux.go
```diff
@@ -3,10 +3,13 @@
...
 
 import (
+	selinux "github.com/opencontainers/go-selinux"
 	"github.com/opencontainers/runtime-tools/generate"
 )
 
 func setupSelinux(g *generate.Generator, processLabel, mountLabel string) {
-	g.SetProcessSelinuxLabel(processLabel)
-	g.SetLinuxMountLabel(mountLabel)
+	if processLabel != "" && selinux.EnforceMode() != selinux.Disabled {
```
rhatdan: There is an interface selinux.GetEnabled() should be enough, but I don't think you even need this.  The processLabel will be "" if SELinux is enabled.

nalind:
I think we had a report from IRC where the user had created the container while SELinux was enabled, but hit an error while attempting to run it after disabling SELinux.  It's probably not a 100% fix, but I think it's an improvement.

rhatdan:
Change to selinux.GetEnabled() and then we can merge.

EnforceMode, reads /etc/selinux/config and tells you what is set there,  Has nothing to do with host until the next reboot.

nalind:
@rhatdan are you sure?  The godoc for `EnforceMode()` doesn't mention that, and a casual reading doesn't suggest it.

rhatdan:
```
man selinux_getenforcemode
selinux_getenforcemode(3)                                                                               SELinux API documentation                                                                              selinux_getenforcemode(3)

NAME
       selinux_getenforcemode - get the enforcing state of SELinux

SYNOPSIS
       #include <selinux/selinux.h>

       int selinux_getenforcemode(int *enforce);

DESCRIPTION
       selinux_getenforcemode() Reads the contents of the /etc/selinux/config file to determine how the system was setup to run SELinux.

       Sets the value of enforce to 1 if SELinux should be run in enforcing mode.  Sets the value of enforce to 0 if SELinux should be run in permissive mode.  Sets the value of enforce to -1 if SELinux should be disabled.

RETURN VALUE
       On success, zero is returned.  On failure, -1 is returned.

SEE ALSO
       selinux(8)

dwalsh@redhat.com                                                                                              25 May 2004                                                                                     selinux_getenforcemode(3)
```

rhatdan:
Sadly the golang does not match.

But the interface we want to expose is GetEnabled()

rhatdan:
@rh-atomic-bot r+

PR #1548, opened 2019-04-24
allow portions of buildah to build for windows

baude:
the podman remote client for windows needs to be able to compile certain
portions of buildah for windows.  this seems like the safest way to not
break api for buildah.

as podman matures, we might be able to avoid this by not compiling some
of the core pieces of libpod.  that is a longer term process however.


run_supported.go
```diff
@@ -0,0 +1,1350 @@
+// +build linux darwin
```
rhatdan: Lets call this run_unix.go

run_unsupport.go
```diff
@@ -1,11 +1,32 @@
-// +build !linux
+// +build windows
```
rhatdan: Call this run_windows.go

imagebuildah/chroot_symlink.go
```diff
@@ -1,16 +1,13 @@
 package imagebuildah
```
rhatdan: Can't you just add a #linux to this, and then implement the windows global interfaces?

rhatdan:
Much simpler patch would be to just add

// +build linux to existing files and them implement // +build -linux to implement the non supported ones.

run_unsupport.go
TomSweeneyRedHat: If you do, does @baude need to create a separate run_unsupported with empty functions?

TomSweeneyRedHat:
So I'm curious why podman remote needs to get to Buildah run functionality?  I was expecting to see something closer to @rhatdan 's earlier PR.

imagebuildah/chroot_symlink_supported.go
```diff
@@ -0,0 +1,75 @@
+// +build linux
+
+package imagebuildah
+
+import (
+	"flag"
+	"fmt"
+	"golang.org/x/sys/unix"
```
nalind: Our convention is to group standard library imports together, and everything else together.

nalind:
General comments:  try moving `Run()` and all of its support functions (the ones like `addCommonOptsToSpec()` that would compile for other platforms, but still expect to manipulate paths in /proc, which is Linux-specific, so we're not saving ourselves any trouble by compiling them on other platforms) into `run_linux.go`.  When this is done, `run.go` should only contain type definitions and functions for manipulating those types that libpod needs in order to compile.

Instead of naming files with _supported as a suffix and using a build constraint to only build the file on Linux, use _linux in the filename.

imagebuildah/chroot_symlink.go
```diff
@@ -24,37 +21,6 @@ func init() {
...
-	if _, err := os.Stdout.WriteString(symLink); err != nil {
-		fmt.Fprintf(os.Stderr, "error writing string to stdout: %v\n", err)
-		os.Exit(1)
-	}
-	os.Exit(status)
-}
-
 // resolveSymlink uses a child subprocess to resolve any symlinks in filename
 // in the context of rootdir.
 func resolveSymlink(rootdir, filename string) (string, error) {
```
nalind: Move this to `chroot_symlink_linux.go`, along with `resolveChrootedSymlinks()`, and stub just `resolveSymlink()` out in `chroot_symlink_unsupported.go` - `resolveSymlink()` is the function that gets called from elsewhere, and it'd be better to have it return an error on other platforms than to have it try to run a child process that will just mysteriously exit immediately because we're stubbing out its `main()`.

PR #1551, opened 2019-04-25
allow podman to build a client for windows

baude:
the podman remote-client for windows pulls in some buildah code for
things like commit and build.  we need to perform some slight
refactoring of buildah code to accomodate that build.


rhatdan:
LGTM

nalind:
Move `resolveModifiedTime()`, `modTimeIsGreater()`, `getSymbolicLink()`, and `hasSymlink()` to `chroot_symlink_linux.go`.  Only `resolveSymlink()` and `resolveModifiedTime()` need to be stubbed out in `chroot_symlink_unsupported.go`.  `chroot_symlink.go` should basically be renamed to `chroot_symlink_linux.go` (or to `chroot_symlink_unix.go` with build constraints added, if we intend for the functions to run on other Unixy systems) -- it shouldn't exist any more.

The only functions that should be left in `run.go` are the methods for `TerminalPolicy`, `Isolation`, and `NamespaceOptions` types, along with the definitions of their types, and definitions for the `RunOptions` and `IDMappingOptions` types.  `run_unsupported.go` should then only need to provide a `DefaultNamespaceOptions()` that returns nothing, and a `Run()` method that returns an error.  Literally everything else from `run.go` should be moved into `run_linux.go`.

Move the bits of logic in `cmd/buildah/common.go` that manipulate the umask into a function in `common_unix.go`, and stub it out in a `common_unsupported.go`.  Likewise, the logic in `pkg/parse` that modifies the current process limits and returns a list of them should be moved into a `parse_unix.go` and stubbed out with a version that returns an empty list in `parse_unsupported.go`.

While you're moving things around, go ahead and rename files with names ending in `_unsupport.go` so that they end with `_unsupported.go` -- it's how other packages tend to name source files that serve the same role.

baude:
updated

rhatdan:
@baude Darwin build is failing.

baude:
fixed

rhatdan:
LGTM

nalind:
This is much better, thanks.

Everything from `contains()` on down in `run.go` should just be moved to `run_linux.go`, since they're all details of how the implementations in `run_linux.go` do their thing and not used from elsewhere.  Everything in `chroot_symlink.go` should be moved to `chroot_symlink_linux.go`, for the same reason.

 `common_unix.go` and `parse_unix.go` need to have some build constraints added, as `_unix` suffixes in filenames don't signify anything to the compiler, so on platforms where their `_unsupported` alternates get built, we'll get multiple definitions of things.

rhatdan:
@rh-atomic-bot r+

nalind:
LGTM.  Thanks!

rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot retry

PR #1552, opened 2019-04-26
Bump 1.8

TomSweeneyRedHat:
Bump to 1.8.  I'd thought about 1.7.4, but given the Fedora 30 packaging and the change to c/storage here, I went to 1.8.

### DIFF
--- a/buildah.go
+++ b/buildah.go
@@ -26,7 +26,7 @@ const (
 	Package = "buildah"
 	// Version for the Package.  Bump version in contrib/rpm/buildah.spec
 	// too.
-	Version = "1.8-dev"
+	Version = "1.8"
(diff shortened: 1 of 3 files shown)


vrothberg:
LGTM, I agree that 1.8 seems good. One thing I just noted is that we did not consistently follow semver with the versioning as 1.x should be a 1.x.0.

Can we name it 1.8.0 (also the git tag)? This would help a bit with using `go mod` as we had a consistent versioning scheme.

rhatdan:
Yes this should be v1.8.0.

TomSweeneyRedHat:
Ugh, completely hosed my branch trying to update the first commit.  Going to kill this and start up a new one.

Issue #1537, opened 2019-04-19
Symlink in Workdir cause error


TomSweeneyRedHat:
**Description**

When the WORKDIR is a symlink, the directory is not resolved appropriately.

**Steps to reproduce the issue:**
1.  See received results.

**Describe the results you received:**
As reported by @karelyatin in #1443.

@EmilienM @TomSweeneyRedHat i tried with the recent PR and i see similar Error as we have with buidah-1.7.1:-
```
[zuul@standalone ~]$ buildah --version
buildah version 1.7.1 (image-spec 1.0.0, runtime-spec 1.0.0)

[zuul@standalone ~]$ sudo buildah bud --tls-verify=False  -t 127.0.0.1:8787/tripleomaster/centos-binary-tempest:be90d93c3c5f77f428d12a9a8a2ef97b9dada8f3_5b53d5ba tempest
STEP 1: FROM docker.io/tripleomaster/centos-binary-openstack-base:be90d93c3c5f77f428d12a9a8a2ef97b9dada8f3_5b53d5ba
STEP 2: LABEL maintainer="Kolla Project (https://launchpad.net/kolla)" name="tempest" build-date="20190415"
STEP 3: RUN usermod --append --home /var/lib/tempest --groups kolla tempest     && mkdir -p /var/lib/tempest     && chown -R 42480:42480 /var/lib/tempest
STEP 4: RUN ln -sf /var/lib/tempest /tempest
STEP 5: WORKDIR /tempest
error building at step {Env:[PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin KOLLA_BASE_DISTRO=centos KOLLA_INSTALL_TYPE=binary KOLLA_INSTALL_METATYPE=rdo KOLLA_DISTRO_PYTHON_VERSION=2.7 PS1=$(tput bold)($(printenv KOLLA_SERVICE_NAME))$(tput sgr0)[$(id -un)@$(hostname -s) $(pwd)]$ ] Command:workdir Args:[/tempest] Flags:[] Attrs:map[] Message:WORKDIR /tempest Original:WORKDIR /tempest}: error ensuring container path "/tempest": mkdir /var/lib/containers/storage/overlay/609939041c25b5d318dcba013e5f8d26bfe11a040f579c16fb64958a5d9bb713/merged/tempest: file exists

[zuul@standalone ~]$ /usr/local/bin/buildah --version
[...]

PR #1558, opened 2019-04-26
Handle WORKDIRs that are symlinks

TomSweeneyRedHat:
When a Dockerfile specified a WORKDIR that was
a symlink, the EnsureContainerPath() function
did not resolve the link to it's target and
caused failures.

Fixes: #1537

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -1812,12 +1812,25 @@ func (b *Executor) deleteSuccessfulIntermediateCtrs() error {
 }
 
 func (s *StageExecutor) EnsureContainerPath(path string) error {
-	_, err := os.Stat(filepath.Join(s.mountPoint, path))
+	// Check the path to see if it's a symlink, if
+	// so convert it to it's target for processing.
+	targetPath := filepath.Join(s.mountPoint, path)
+	fileDest, _ := os.Lstat(targetPath)
+	if fileDest != nil {
+		if fileDest.Mode()&os.ModeSymlink != 0 {
+			if symLink, err := resolveSymlink(s.mountPoint, path); err == nil {
+				targetPath = symLink
+			} else {
+				return errors.Wrapf(err, "error reading symbolic link to %q", targetPath)
+			}
+		}
+	}
+	_, err := os.Stat(targetPath)
 	if err != nil && os.IsNotExist(err) {
-		err = os.MkdirAll(filepath.Join(s.mountPoint, path), 0755)
+		err = os.MkdirAll(targetPath, 0755)
 	}
 	if err != nil {
-		return errors.Wrapf(err, "error ensuring container path %q", path)
+		return errors.Wrapf(err, "error ensuring container path %q", targetPath)
 	}
 	return nil
 }
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -1164,3 +1164,19 @@ load helpers
   run_buildah --debug=false run ${ctr} ls -alF /usr/local/hbase 
   expect_output --substring "Dockerfile"
 }
+
+@test "bud WORKDIR isa symlink" {
+  target=alpine-image
+  ctr=alpine-ctr
+  run_buildah --debug=false bud --signature-policy ${TESTSDIR}/policy.json -t ${target} ${TESTSDIR}/bud/workdir-symlink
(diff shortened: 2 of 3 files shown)


TomSweeneyRedHat:
@karelyatin PTAL.  Your patch was dead on for the location, but I had some problems with the chmod command in your Dockerfile as it was written.  So I changed the code a bit to emulate what was done in #1443.  Thanks again for the help with locating and fixing this bug.

TomSweeneyRedHat:
@edsantiago if you're symlink eagle eye is still around, happy to have you review this one too.

rhatdan:
LGTM

rhatdan:
@nalind @edsantiago @giuseppe @vrothberg @umohnani8 PTAL

umohnani8:
LGTM

rhatdan:
@rh-atomic-bot r+
@TomSweeneyRedHat This might be worth releasing as buildah 1.8.1.

karelyatin:
> @karelyatin PTAL. Your patch was dead on for the location, but I had some problems with the chmod command in your Dockerfile as it was written. So I changed the code a bit to emulate what was done in #1443. Thanks again for the help with locating and fixing this bug.

@TomSweeneyRedHat Thanks for making it a go. I tested it and it's working fine, the only thing that i noticed is https://github.com/containers/buildah/blob/master/imagebuildah/build.go#L1830 creating the destination directory on Host if it doesn't exist. It's created inside the container image as well at some other function.

TomSweeneyRedHat:
@rhatdan I too was thinking we ought to get a 1.8.1 out sooner rather than later.  

@karelyatin the directory that's getting created on the host that you're concerned about is getting created somewhere under (assuming root user) /var/lib/containers/storage right?  I think that's OK, @rhatdan or @nalind any concerns with that?

rhatdan:
If the directory is in the containers storage directory it should be fine.

karelyatin:
> @rhatdan I too was thinking we ought to get a 1.8.1 out sooner rather than later.
> 
> @karelyatin the directory that's getting created on the host that you're concerned about is getting created somewhere under (assuming root user) /var/lib/containers/storage right? I think that's OK, @rhatdan or @nalind any concerns with that?

@TomSweeneyRedHat @rhatdan  The directory is not created under container storage, it's under host root path(/var/lib/tempest). For the following Dockerfile:-
[zuul@standalone ~]$ cat alpine/Dockerfile 
FROM alpine
RUN mkdir /var/lib/tempest
RUN ln -sf /var/lib/tempest /tempest
WORKDIR /tempest
RUN touch /etc/notareal.conf
RUN chmod 664 /etc/notareal.conf

[zuul@standalone ~]$ ls -l /var/lib/|grep tempest
[zuul@standalone ~]$ sudo /usr/local/bin/buildah bud --tls-verify=False  -t alpine:be90d93c3c5f77f428d12a9a8a2ef97b9dada8f3_5b53d5ba alpine
STEP 1: FROM alpine
STEP 2: RUN mkdir /var/lib/tempest
STEP 3: RUN ln -sf /var/lib/tempest /tempest
STEP 4: WORKDIR /tempest
STEP 5: RUN touch /etc/notareal.conf
STEP 6: RUN chmod 664 /etc/notareal.conf
STEP 7: COMMIT alpine:be90d93c3c5f77f428d12a9a8a2ef97b9dada8f3_5b53d5ba
Copying blob a464c54f93a9 skipped: already exists
Copying blob 16c18fcaf59d done
Copying config 3eb90397bd done
Storing signatures
--> 3eb90397bd640a29b1c4facfe197ae8a0ad062ecf1093ff101c81052dffb0920
[zuul@standalone ~]$ ls -l /var/lib/|grep tempest
drwxr-xr-x.  2 root    root             6 Apr 29 12:02 tempest
[zuul@standalone ~]$ /usr/local/bin/buildah --version
buildah version 1.9.0-dev (image-spec 1.0.0, runtime-spec 1.0.0)

edsantiago:
Confirmed. Nice catch, @karelyatin

TomSweeneyRedHat:
Thanks for the example and the heads up @karelyatin .  I've created an issue #1562 so that it doesn't get lost in the shuffle.  If you have further information to add, please add it to the issue.I hope to take a deeper dive on this later today.

PR #1564, opened 2019-04-30
imagebuildah: don't remove intermediate images if we need them

nalind:
In multistage builds without caching, if an intermediate stage's image ended up being the final image (i.e., when the last instruction in the Dockerfile is a FROM instruction that references a previous stage), we would remove it when we finished building.  Fix that by modifying the cleanup logic to compare the ID of an image that it's about to delete to the final image's ID, if it has one, and skipping it if they match.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -1558,6 +1558,9 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 		// stages.
 		for i := range cleanupImages {
 			removeID := cleanupImages[len(cleanupImages)-i-1]
+			if removeID == imageID {
+				continue
+			}
 			if _, err := b.store.DeleteImage(removeID, true); err != nil {
 				logrus.Debugf("failed to remove intermediate image %q: %v", removeID, err)
 				if b.forceRmIntermediateCtrs || errors.Cause(err) != storage.ErrImageUsedByContainer {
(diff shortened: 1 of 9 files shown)


rhatdan:
LGTM
@TomSweeneyRedHat @vrothberg @giuseppe @pixdrift @umohnani8 PTAL

rhatdan:
LGTM

rhatdan:
No idea why tests are not running.

nalind:
It looks like this gets https://github.com/openshift/builder/pull/67 to pass.

vrothberg:
@rh-atomic-bot r+

imagebuildah/build.go
```diff
@@ -1663,6 +1666,7 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 			if !b.layers {
 				cleanupImages = append(cleanupImages, imageID)
 			}
+			imageID = ""
```
TomSweeneyRedHat: Is this going to cause us to remember only the first intermediate stage that we run into?  I don't see imageID getting set within this `for stageIndex, stage := range stages ` loop and I'm not sure that's correct?

nalind: Not sure I follow - `imageID` is set to the value that `stageExecutor.Execute()` returns, so clearing it here is mainly about making sure that if we return early due to an error result from `ib.From()` in a subsequent loop iteration, we don't mistakenly leave the most recent stage in place when, because it's a failure case, we want to remove it.

TomSweeneyRedHat: Ugh, false alarm, sorry.  I somehow didn't see imageID being set at line 1645.  I could have sworn I saw it getting cleared here and not reset again within the for loop getting the next stage.  Can I chalk it up to a lack of tea?

nalind:
@rh-atomic-bot retry

Issue #1569, opened 2019-05-02
symlink workdir without target dir fails.


TomSweeneyRedHat:
**Description**
Reported by @karelyatin in #1565

**Steps to reproduce the issue:**
```
# cat ~/Dockerfile
FROM alpine
#RUN mkdir /var/lib/tempest
RUN ln -sf /var/lib/tempest /tempest
WORKDIR /tempest
RUN touch /etc/notareal.conf
RUN chmod 664 /etc/notareal.conf

# buildah bud -t tom -f ~/Dockerfile .
```

**Describe the results you received:**
```
# buildah bud -t tom -f ~/Dockerfile .
STEP 1: FROM alpine
STEP 2: RUN ln -sf /var/lib/tempest /tempest
STEP 3: WORKDIR /tempest
STEP 4: RUN touch /etc/notareal.conf
container_linux.go:345: starting container process caused "process_linux.go:424: container init caused \"mkdir /tempest: file exists\""
error running container: error creating container for [/bin/sh -c touch /etc/notareal.conf]: : exit status 1
error building at STEP "RUN touch /etc/notareal.conf": error while running runtime: exit status 1

```

**Describe the results you expected:**

Successful build.

**Output of `buildah version`:**

```
# buildah version 
Version:         1.9.0-dev
Go Version:      go1.12.2
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.4.0
libcni Version:  v0.7.0-rc2
Git Commit:      5dc376c3
Built:           Thu May  2 15:55:48 2019
OS/Arch:         linux/amd64

```

PR #1576, opened 2019-05-03
Handle WORKDIR with dangling target

TomSweeneyRedHat:
When the WORKDIR is a symlink and the target directory does not exist, RUN
commands would fail.  If the target doesn't exist, create the directory in
the container.    This also converts the code to use SecureJoin so that if
a symlink is in the middle of a filepath, it will be converted correctly.

Fixes: #1569

Tip of the hat to @karelyatin for the initial fix

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -27,6 +27,7 @@ import (
 	"github.com/containers/image/types"
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/archive"
+	"github.com/cyphar/filepath-securejoin"
 	docker "github.com/fsouza/go-dockerclient"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/opencontainers/runtime-spec/specs-go"
@@ -480,22 +481,22 @@ func (s *StageExecutor) volumeCacheRestore() error {
 }
 
 // Copy copies data into the working tree.  The "Download" field is how
-// imagebuilder tells us the instruction was "ADD" and not "COPY".
+// imagebuilder tells us the instruction was "ADD" and not "COPY"
 func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) error {
 	for _, copy := range copies {
-		// If the file exists, check to see if it's a symlink.
-		// If it is a symlink, convert to it's target otherwise
-		// the symlink will be overwritten.
-		fileDest, _ := os.Lstat(filepath.Join(s.mountPoint, copy.Dest))
-		if fileDest != nil {
-			if fileDest.Mode()&os.ModeSymlink != 0 {
-				if symLink, err := resolveSymlink(s.mountPoint, copy.Dest); err == nil {
-					copy.Dest = symLink
-				} else {
-					return errors.Wrapf(err, "error reading symbolic link to %q", copy.Dest)
-				}
-			}
+		// Check the file and see if part of it is a symlink.
+		// Convert it to the target if so.  To be ultrasafe
+		// do the same for the mountpoint.
+		secureMountPoint, err := securejoin.SecureJoin("", s.mountPoint)
+		finalPath, err := securejoin.SecureJoin(secureMountPoint, copy.Dest)
+		if err != nil {
+			return errors.Wrapf(err, "error resolving symlinks for copy destination %s", copy.Dest)
 		}
+		if !strings.HasPrefix(finalPath, secureMountPoint) {
+			return errors.Wrapf(err, "error resolving copy destination %s", copy.Dest)
+		}
+		copy.Dest = strings.TrimPrefix(finalPath, secureMountPoint)
+
 		if copy.Download {
 			logrus.Debugf("ADD %#v, %#v", excludes, copy)
 		} else {
@@ -1819,8 +1820,8 @@ func (b *Executor) deleteSuccessfulIntermediateCtrs() error {
 }
 
 func (s *StageExecutor) EnsureContainerPath(path string) error {
-	targetPath := filepath.Join(s.mountPoint, path)
-	_, err := os.Lstat(targetPath)
+	targetPath, err := securejoin.SecureJoin(s.mountPoint, path)
+	_, err = os.Lstat(targetPath)
 	if err != nil && os.IsNotExist(err) {
 		err = os.MkdirAll(targetPath, 0755)
 	}
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -1198,7 +1198,48 @@ load helpers
   run_buildah --debug=false run ${ctr} ls -alF /tempest
   expect_output --substring "/tempest -> /var/lib/tempest/"
 
-  run_buildah --debug=false run ${ctr} ls -alF /etc/notareal.conf 
+  run_buildah --debug=false run ${ctr} ls -alF /etc/notareal.conf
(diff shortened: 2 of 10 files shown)


imagebuildah/build.go
```diff
@@ -1817,13 +1817,20 @@ func (b *Executor) deleteSuccessfulIntermediateCtrs() error {
...
 		err = os.MkdirAll(targetPath, 0755)
 	}
 	if err != nil {
 		return errors.Wrapf(err, "error ensuring container path %q", path)
 	}
+	// If a symlink, check that the target directory exists in the container.
+	// If not, make it.
+	if joinedPath.Mode()&os.ModeSymlink != 0 {
+		if symTarget, err2 := os.Readlink(targetPath); err2 == nil {
+			os.MkdirAll(filepath.Join(s.mountPoint, symTarget), 0755)
```
rhatdan: I believe you have to make sure that the path is withing the s.mountPoint,  Otherwise the link could create a directory on the host.

rhatdan: libpod is using SecureJoin.
./vendor/github.com/cyphar/filepath-securejoin/join.go

TomSweeneyRedHat: @rhatdan thanks for the pointer.  I was going through how to code that last night in my noggin, looks like it's already there with securejoin.

TomSweeneyRedHat:
Hold reviews on new patch, just spotted an error, will repush

TomSweeneyRedHat:
Repushed a new patch, added SecureJoin and a new test.  All green test buttons.

imagebuildah/build.go
```diff
@@ -480,22 +481,21 @@ func (s *StageExecutor) volumeCacheRestore() error {
...
-					return errors.Wrapf(err, "error reading symbolic link to %q", copy.Dest)
-				}
-			}
+		// Check the file and see if part of it is a symlink.
+		// Convert it to the target if so.
+		finalPath, err := securejoin.SecureJoin(s.mountPoint, copy.Dest)
+		if err != nil {
+			return errors.Wrapf(err, "error resolving symlinks for copy destination %s", copy.Dest)
+		}
+		targetDest := strings.Split(finalPath, s.mountPoint)
```
vrothberg: I don't think this additional check is necessary but maybe I am missing a corner case. If we keep it, could we use [strings.HasPrefix()](https://golang.org/pkg/strings/#HasPrefix) instead? Using strings.Split() could fail when the mountpoint is equal to a substring of the copy.Dest (e.g., mountpoint=/a/b and copy.Dest=c/a/b/d/a/b).

TomSweeneyRedHat: Just double checking you're commenting on the check at line 494 for the len right?  I need to do the split as finalpath at this point would look something like `/var/lib/docker/overlay2/f92fd1c6882516f93635319c0018706ddb46d3aa85283e94b2012ef6aa25d50d/merged/var/lib/tempest/` and we only would want to pass down `/var/lib/tempest` as code further down the stream will prepend the rest of that path back onto it.
The len check in probably superfluous, but I've been burned a few time without checks like this.

I'll play with HasPrefix and will see if I can get that to go instead, good thought thanks!

TomSweeneyRedHat:
I've added a strings.HasPrefix() check before the error and also replaced 'tom' with 'lowerdir' in the tests and related Dockerfiles.  So much for infamy!

TomSweeneyRedHat:
Fixes: #1577
also

tests/bud.bats
```diff
@@ -1186,6 +1186,47 @@ load helpers
...
+  run_buildah --debug=false from --signature-policy ${TESTSDIR}/policy.json --name=${ctr} ${target}
+  expect_output --substring ${ctr}
+
+  run_buildah --debug=false run ${ctr} ls -alF /tempest
+  expect_output --substring "/tempest -> /var/lib/tempest/"
+
+  run_buildah --debug=false run ${ctr} ls /tempest
+  expect_output --substring "Dockerfile-2"
+
+  run_buildah --debug=false run ${ctr} ls -alF /etc/notareal.conf
```
giuseppe: nit: trailing whitespace

imagebuildah/build.go
giuseppe: to use HasPrefix we need to be sure `s.mountPoint` is also resolved, can we be sure at this point that s.mountPoint has no symlinks and no `..` components?

TomSweeneyRedHat: I'd assumed, possibly badly, that s.mountPoint was completely under our (Buildah's) control at this point and a symlink would not be part of that equation.  @nalind thoughts?  Should I verify that s.mountPoint doesn't have a symlink before the hasPrefix?

tests/bud.bats
TomSweeneyRedHat: fixed

TomSweeneyRedHat:
@giuseppe @vrothberg PTAL, I think I've addressed your concerns with the latest.

rhatdan:
LGTM

imagebuildah/build.go
```diff
@@ -1816,8 +1818,8 @@ func (b *Executor) deleteSuccessfulIntermediateCtrs() error {
 }
 
 func (s *StageExecutor) EnsureContainerPath(path string) error {
-	targetPath := filepath.Join(s.mountPoint, path)
-	_, err := os.Lstat(targetPath)
+	targetPath, err := securejoin.SecureJoin(s.mountPoint, path)
+	_, err = os.Lstat(targetPath)
```
karelyatin: os.Stat should also work now as securejoin returns resolved path. iirc Lstat was put earlier as targetPath could be a symlink.

imagebuildah/build.go
```diff
@@ -480,22 +481,23 @@ func (s *StageExecutor) volumeCacheRestore() error {
...
+		// Check the file and see if part of it is a symlink.
+		// Convert it to the target if so.  To be ultrasafe
+		// do the same for the mountpoint.
+		secureMountPoint, err := securejoin.SecureJoin("", s.mountPoint)
+		finalPath, err := securejoin.SecureJoin(secureMountPoint, copy.Dest)
+		if err != nil {
+			return errors.Wrapf(err, "error resolving symlinks for copy destination %s", copy.Dest)
+		}
+		targetDest := strings.Split(finalPath, secureMountPoint)
+		if !strings.HasPrefix(finalPath, secureMountPoint) || len(targetDest) != 2 {
```
vrothberg: The usage of `strings.Split(...)` is still problematic as the code will fail for reoccurring patterns. Instead, we can use [strings.TrimPrefix](https://golang.org/pkg/strings/#TrimPrefix):
```golang
if !HasPrefix(...) { return error}
copy.Dest = trimPrefix(finalPath, secureMountPoint)
```

imagebuildah/build.go
TomSweeneyRedHat: Thanks!  Yeah, we probably being too cautious there with the Lstat, but I think I'll keep it as is just in case.

TomSweeneyRedHat:
Repushed with @vrothberg 's requested HasPrefix/TrimPrefix change.

TomSweeneyRedHat:
All happy green test buttons....  @vortherberg, @rhatdan ready to merge?

vrothberg:
Sorry, it slipped through my emails.

LGTM, thanks a lot @TomSweeneyRedHat :+1:

rhatdan:
@rh-atomic-bot r+

PR #1579, opened 2019-05-08
commit/push: use an everything-allowed policy

nalind:
Ignore the global signature policy, which is used to control whether or not we can trust an image, when we're attempting to commit an image from a container, or pushing an image.  The global policy still affects pulling of base images at the time when we pull them, which I _think_ is correct.

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -106,6 +106,22 @@ type PushOptions struct {
 	Quiet bool
 }
 
+var (
+	// commitPolicy bypasses any signing requirements when committing containers to images
+	commitPolicy = &signature.Policy{
+		Default: []signature.PolicyRequirement{signature.NewPRReject()},
+		Transports: map[string]signature.PolicyTransportScopes{
+			is.Transport.Name(): {
+				"": []signature.PolicyRequirement{
+					signature.NewPRInsecureAcceptAnything(),
+				},
+			},
+		},
+	}
+	// pushPolicy bypasses any signing requirements when pushing (copying) images from local storage
+	pushPolicy = commitPolicy
+)
+
 // Commit writes the contents of the container, along with its updated
 // configuration, to a new image in the specified location, and if we know how,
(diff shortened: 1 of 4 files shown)


nalind:
This should fix https://github.com/openshift/builder/issues/71.

TomSweeneyRedHat:
The changes LGTM
I'm just a little leery about the security implications and would like @rhatdan, @vrothberg or @mtrmac  to weigh in.

commit.go
```diff
@@ -106,6 +106,13 @@ type PushOptions struct {
 	Quiet bool
 }
 
+var (
+	// commitPolicy bypasses any signing requirements when committing containers to images
+	commitPolicy = []byte(`{"default": [ { "type": "insecureAcceptAnything" } ] }`)
+	// pushPolicy bypasses any signing requirements when pushing (copying) images from local storage
+	pushPolicy = commitPolicy
```
mtrmac: 1. Please use the Go API:
    ```go
    &signature.Policy{Default: []signature.PolicyRequirement{signature.NewPRInsecureAcceptAnything()}}
    ```
    so that the code visibly breaks on a typo, or if the policy format ever changed.

2. Yes, the change seems to be correct right now, but for extra paranoia I’d prefer an explicit
    - default = reject
    - containers-storage = insecureAcceptAnything

    configuration, so that if ever any of the code is refactored or extended to allow non-storage or remote sources, it does not accidentally bypass policy. (In that case, I think hard-failing is sufficient, because it “can’t happen”; if you wanted to be fancy, you could load the actual system policy and only override the containers-storage configuration, but I don’t think it’s warranted.

nalind: Very convincing on both counts.  Changed it to create an API struct rather than parsing a hard-coded configuration in text format.

rhatdan:
@rh-atomic-bot r+

PR #1583, opened 2019-05-08
imagebuildah: addHelper: handle symbolic links

nalind:
When filtering copying/adding operations to comply with a .dockerignore file, we should directly handle copying directories and symbolic links.  This update still returns an error when attempting to copy non-directory, non-regular items.
    
Make `DockerIgnoreHelper()` private, since it's not being used by packages that use us, and I expect it to be removed when this gets refactored.


TomSweeneyRedHat:
LGTM
@vrothberg @rhatdan PTAL

rhatdan:
@rh-atomic-bot r+

PR #1588, opened 2019-05-10
Apply SHELL during image build time

### LINKED ISSUE #1581: SHELL instruction isn't applied at image build time
**Description**

1. Use the following `Dockerfile`
```
FROM docker.io/ubuntu
SHELL ["/usr/bin/this/does/not/exist"]
RUN echo "Hello world"
```
2. Build an image `buildah bud --format=docker .`

Image build succeeds. We can, therefore, conclude that shell wasn't actually changed.
```
STEP 1: FROM docker.io/ubuntu
STEP 2: SHELL ["/usr/bin/this/does/not/exist"]
STEP 3: RUN echo "Hello world"
Hello world
STEP 4: COMMIT
Copying blob 762d8e1a6054 skipped: already exists
Copying blob e45cfbc98a50 skipped: already exists
Copying blob d60e01b37e74 skipped: already exists
Copying blob b57c79f4a9f3 skipped: already exists
Copying blob 5f70bf18a086 skipped: already exists
Copying config 9e1fa82b3f done
Storing signatures
--> 9e1fa82b3f37ed3b1d6b0a283133550f65324cbbf93c04ba6257d4d53b7e2085
```

Expected `buildah` to attempt to use the `SHELL` defined and, as a result, fail to build the image.
Docker output to compare:
```
$ docker build -t .
Step 1/3 : FROM docker.io/ubuntu
 ---> 94e814e2efa8
Step 2/3 : SHELL ["/usr/bin/this/does/not/exist"]
 ---> Running in dc5c819df407
Removing intermediate container dc5c819df407
 ---> 0735bcf1de70
Step 3/3 : RUN echo "Hello world"
 ---> Running in a3f4d8c42eb3
OCI runtime create failed: container_linux.go:348: starting container process caused "exec: \"/usr/bin/this/does/not/exist\": stat /usr/bin/this/does/not/exist: no such file or directory": unknown
```

```
Listing... Done
buildah/xenial,now 1.8-1~dev~ubuntu16.04.2~ppa5 amd64 [installed]
```

```
Version:         1.8-dev
Go Version:      go1.10.4
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.4.0
libcni Version:
Git Commit:
Built:           Sat Apr 13 16:42:38 2019
OS/Arch:         linux/amd64
```

```
N/A
```

```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.6 LTS"
NAME="Ubuntu"
VERSION="16.04.6 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.6 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```

**Output of `uname -a`:**

```
Linux scw-35eb3f 4.5.7-docker-4 #1 SMP Tue Jul 12 09:41:14 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
```

```

[storage]

# Default Storage Driver
driver = "overlay"

# Temporary storage location

# Primary read-write location of container storage

[storage.options]
# AdditionalImageStores is used to pass paths to additional read-only image stores
additionalimagestores = [
]

# certain container storage drivers (currently overlay, zfs, vfs, btrfs)
size = ""

```

EricHripko:
# Overview
This PR fixes the behaviour of `SHELL` by making sure that it applies during image build time in addition to existing functionality. Resolves #1581.

# Testing
4 new tests added:
- default shell + OCI format
- default shell + Docker format
- custom shell + OCI format
- custom shell + Docker format

Output:
```
 ✓ bud-shell during build in Docker format
 ✓ bud-shell during build in OCI format
 ✓ bud-shell changed during build in Docker format
 ✓ bud-shell changed during build in OCI format
...
296 tests, 10 failures, 5 skipped
```
10 tests that are failing (as far as I can see) rely on an image registry being setup at `localhost`. Therefore, I don't think they're caused by my changes.

# Signing
Signed-off-by: Eric Hripko <ehripko@bloomberg.net>

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -589,7 +589,11 @@ func (s *StageExecutor) Run(run imagebuilder.Run, config docker.Config) error {
 
 	args := run.Args
 	if run.Shell {
-		args = append([]string{"/bin/sh", "-c"}, args...)
+		if len(config.Shell) > 0 && s.builder.Format == buildah.Dockerv2ImageManifest {
+			args = append(config.Shell, args...)
+		} else {
+			args = append([]string{"/bin/sh", "-c"}, args...)
+		}
 	}
 	if err := s.volumeCacheSave(); err != nil {
 		return err
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -635,6 +635,42 @@ load helpers
   expect_output ""
 }
 
+@test "bud-shell during build in Docker format" {
+  target=alpine-image
+  run_buildah bud --format docker --signature-policy ${TESTSDIR}/policy.json -t ${target} -f ${TESTSDIR}/bud/shell/Dockerfile.build-shell-default ${TESTSDIR}/bud/shell
+  expect_output --substring "SHELL=/bin/sh"
+  buildah rmi -a
+  run_buildah --debug=false images -q
+  expect_output ""
+}
+
+@test "bud-shell during build in OCI format" {
+  target=alpine-image
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json -t ${target} -f ${TESTSDIR}/bud/shell/Dockerfile.build-shell-default ${TESTSDIR}/bud/shell
+  expect_output --substring "SHELL=/bin/sh"
+  buildah rmi -a
+  run_buildah --debug=false images -q
+  expect_output ""
+}
+
+@test "bud-shell changed during build in Docker format" {
+  target=ubuntu-image
+  run_buildah bud --format docker --signature-policy ${TESTSDIR}/policy.json -t ${target} -f ${TESTSDIR}/bud/shell/Dockerfile.build-shell-custom ${TESTSDIR}/bud/shell
+  expect_output --substring "SHELL=/bin/bash"
+  buildah rmi -a
+  run_buildah --debug=false images -q
+  expect_output ""
+}
+
+@test "bud-shell changed during build in OCI format" {
+  target=ubuntu-image
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json -t ${target} -f ${TESTSDIR}/bud/shell/Dockerfile.build-shell-custom ${TESTSDIR}/bud/shell
+  expect_output --substring "SHELL=/bin/sh"
+  buildah rmi -a
+  run_buildah --debug=false images -q
+  expect_output ""
+}
+
 @test "bud with symlinks" {
   target=alpine-image
   run_buildah bud --signature-policy ${TESTSDIR}/policy.json -t ${target} ${TESTSDIR}/bud/symlink
--- a/tests/bud/shell/Dockerfile.build-shell-custom
+++ b/tests/bud/shell/Dockerfile.build-shell-custom
@@ -1,0 +1,3 @@
+FROM ubuntu
+SHELL [ "/bin/bash", "-c" ]
+RUN echo "SHELL=$0"
--- a/tests/bud/shell/Dockerfile.build-shell-default
+++ b/tests/bud/shell/Dockerfile.build-shell-default
@@ -1,0 +1,2 @@
+FROM alpine
+RUN echo "SHELL=$0"


rhatdan:
Thanks @EricHripko Looks great 
LGTM
Could you check if this works properly if you do this via Bash instead of Dockerfile?

```
# ctr=$(buildah from --format docker fedora)
#  buildah config --shell '/bin/sh -c' $ctr 
#  buildah run $ctr  echo "SHELL=$0"
```

EricHripko:
Hi @rhatdan! Thank you for looking at this :+1: 

I tried to get this to work, but I am struggling with some shell expansion. Namely, I get the following:
```
   #|     FAIL: buildah --debug=false run ubuntu-working-container echo SHELL=/usr/lib/bats/bats-exec-test
   #| expected: 'SHELL=/bin/bash'
   #|   actual: 'SHELL=/usr/lib/bats/bats-exec-test
```
It seems that host shell insists on expanding `$0` before it gets inside a working container:
```
$ buildah run ubuntu-working-container echo "SHELL=$0"
SHELL=bash
# ^- should be /bin/sh
```
I've tried various escaping tricks, but it seems like the only 'proper' way to do this is to do something like `buildah run ubuntu-working-container sh -c 'echo SHELL=$0'`. Unfortunately, at this point we are no longer testing the default shell :slightly_frowning_face:

Do you have any suggestions on how to get this working?

rhatdan:
I can not get it to work either.  But how about this patch, which I believe will cover both use cases.

https://paste.fedoraproject.org/paste/~To3nUps4Ti4KuzHKkthTA

rhatdan:
With my patch I am getting.
```
# buildah config --shell 'bash -c' $ctr 
# ./buildah run $ctr  'echo $0'
/usr/bin/bash
# buildah config --shell 'sh -c' $ctr 
# ./buildah run $ctr  'echo $0'
/usr/bin/sh
```

EricHripko:
With this patch, we would be always forcing the shell to take effect. This means that cases below would use the shell whilst they didn't previously. 
- `RUN ["echo", "stuff"]`
- `buildah run $ctr echo "stuff"`
  ( ^- guess that's why we couldn't get the test going, the code path doesn't allow it currently :see_no_evil: )

This would be the case for both custom and default shell - and for both Docker and OCI formats too.
Is this ok?

rhatdan:
Well could you check if we match Docker behaviour?

EricHripko:
Shell is indeed not used for JSON format of the `RUN` instruction. As a result, `$0` isn't expanded at all in Docker:
```
Step 1/3 : FROM ubuntu
 ---> 94e814e2efa8
Step 2/3 : RUN echo $0
 ---> Running in 9ab2d72cf9c3
/bin/sh
Removing intermediate container 9ab2d72cf9c3
 ---> 1246e0ad732b
Step 3/3 : RUN ["echo", "$0"]
 ---> Running in ba2508344e4a
$0
Removing intermediate container ba2508344e4a
 ---> 5fe7ffc812ef
Successfully built 5fe7ffc812ef
```

Shell also doesn't appear to be used for the second case. This makes the current behaviour (without your patch) of `buildah run` match `docker run`:
```
$ docker run --rm ubuntu bash -c 'echo $0'
bash
$ docker run --rm ubuntu sh -c 'echo $0'
sh
```

rhatdan:
My goal here is to make 
buildah run work just like the RUN directive in the Dockerfile.
`docker run` and `podman run` are much different things.

rhatdan:
In a Dockerfile, does docker handle
RUN echo $0
versus
RUN [ "echo", "$0" ]

EricHripko:
Sorry, I initially misunderstood your question.
Yes, `RUN echo $0` correctly prints the current shell.

__Default shell:__
```
...
Step 2/2 : RUN echo $0
 ---> Running in 94f4a2336321
/bin/sh
Removing intermediate container 94f4a2336321
 ---> c87a94c5de69
Successfully built c87a94c5de69
```

__Custom shell:__
```
Step 2/3 : SHELL ["/bin/bash", "-c"]
 ---> Running in ee4e3eec0270
Removing intermediate container ee4e3eec0270
 ---> 53366cb4cafc
Step 3/3 : RUN echo $0
 ---> Running in 39bb8fd8ecae
/bin/bash
Removing intermediate container 39bb8fd8ecae
 ---> 1cbbe22fe598
Successfully built 1cbbe22fe598
```

And, as I've mentioned [earlier](https://github.com/containers/buildah/pull/1588#issuecomment-491616094), `RUN ["echo", "$0"]` doesn't work:
```
Step 3/3 : RUN ["echo", "$0"]
 ---> Running in ba2508344e4a
$0
Removing intermediate container ba2508344e4a
 ---> 5fe7ffc812ef
Successfully built 5fe7ffc812ef
```

rhatdan:
I guess the bottom line then, does buildah bud work the same way with my patch?  Did my patch break anything, since it will allow `buildah run ...` to work the correct way.

EricHripko:
I've tried to apply your patch, but I am getting really different results. Since the patch forces shell interpretation by prepending `sh`, `-c` to a parsed command, only the first argument seems to be passed to the shell. This seems to break multiple things:

- bud
```
STEP 1: FROM docker.io/ubuntu
STEP 2: RUN echo $0
/bin/sh
STEP 3: RUN ["echo", "$0"]

STEP 4: RUN ["apt-get", "install", "vim"]
apt 1.6.8 (amd64)
Usage: apt-get [options] command
...
^- echo and apt-get is called with no arguments
```

- run
```
$ buildah --format=docker from docker.io/ubuntu
ubuntu-working-container-2
$ buildah config --shell 'bash -c' ubuntu-working-container-2
$ buildah run ubuntu-working-container-2 sh -c 'echo $0'
# 
^- new bash shell is spawned since -c is dropped
```

Did you make any other changes apart from the ones on pastebin?

rhatdan:
Ok lets merge your version now and look at the buildah run command separately so we can get this fix in.

rhatdan:
LGTM

rhatdan:
@TomSweeneyRedHat @giuseppe @vrothberg @nalind PTAL

rhatdan:
@TomSweeneyRedHat @giuseppe @vrothberg @nalind PTAL

TomSweeneyRedHat:
@rhatdan @EricHripko sorry, this one slipped off my stack, reviewing now.

TomSweeneyRedHat:
Hold on the merge for a bit, ran into some issues testing that may be self induced, still poking.

TomSweeneyRedHat:
@rh-atomic-bot r+

TomSweeneyRedHat:
And thanks @EricHripko !

PR #1593, opened 2019-05-15
imagebuildah: tighten cache checking back up

nalind:
Since we started adding layers for only certain instructions (somewhere around 74a319586ecbf3207c6bb18e83cfc0557ca47368), history matching for cached images hasn't been precise enough.  Since we now include a full image history, though, we can compare a candidate cache image's history with our base image's exactly, and then check if the current instruction would produce its final history entry.


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM assuming happy tests.
Very nice change @nalind, I like this a lot!

rhatdan:
@rh-atomic-bot r+

Issue #1592, opened 2019-05-13
buildah bud fails on RUN - go build


vtereso:
**Description**

Using the following Dockerfile, I am unable to get successfully run the buildah bud command:
```
# Run using --privileged
FROM quay.io/openshift-pipeline/buildah
ARG GIT_SOURCE="https://github.com/a-roberts/knative-helloworld"
ENV TLS_VERIFY="true"
ENV CONTEXT_PATH="/workspace/git-source"
ENV DOCKERFILE_PATH="/workspace/git-source/Dockerfile"
ENV TAG="testing"
WORKDIR ${CONTEXT_PATH}
# This is the above $GIT_SOURCE repo on my local
COPY git-source/ .
VOLUME "/var/lib/containers"
ENTRYPOINT ["/bin/sh", "-c", "buildah build-using-dockerfile --tls-verify=${TLS_VERIFY} --layers -f ${DOCKERFILE_PATH} -t ${TAG} -- ${CONTEXT_PATH}"]
```

The resulting log is:
```
STEP 1: FROM golang AS builder
Getting image source signatures
Copying blob sha256:c5e155d5a1d130a7f8a3e24cee0d9e1349bff13f90ec6a941478e558fde53c14
Copying blob sha256:221d80d00ae9675aad24913aacbadfac1ce8b7084f9765a6c0813486082c5c69
Copying blob sha256:4250b3117dca5e14edc32ebf1366cd54e4cda91f17610b76c504a86917ff8b95
Copying blob sha256:3b7ca19181b24b87e24423c01b490633bc1e47d2fcdc1987bf2e37949d6789b5
Copying blob sha256:aa24759e848fee3ef333af3dd3ae951eb042e8cd20b5fc0e28a2f3c52cc7e25f
Copying blob sha256:927e9eaeed1922f626e8a34f9a21b6029f36d4112cbb04dbdbd9065e107a59cb
Copying blob sha256:66293f4dacbd8884954f2c9332298ace627830801c3b484ba89ca424c619f374
Copying config sha256:7ced090ee82ee77beabd76ad1ba3b167acd8609b0b10c4ef46cee3ddf6e6fa5f
Writing manifest to image destination
Storing signatures
STEP 2: WORKDIR /go/src/github.com/knative/docs/helloworld
--> 9c3aa441b8638080568faaf59b277d07cb861f48195a3fbeec558dc5b01e2b2b
[...]

PR #1594, opened 2019-05-15
Add default /root if empty on chroot iso

TomSweeneyRedHat:
Checks to see if the $HOME envvar has been set
and if not, trys to set it as best as possible.

Fixes: #1592

### DIFF
--- a/chroot/run.go
+++ b/chroot/run.go
@@ -84,9 +84,18 @@ type runUsingChrootExecSubprocOptions struct {
 // RunUsingChroot runs a chrooted process, using some of the settings from the
 // passed-in spec, and using the specified bundlePath to hold temporary files,
 // directories, and mountpoints.
-func RunUsingChroot(spec *specs.Spec, bundlePath string, stdin io.Reader, stdout, stderr io.Writer) (err error) {
+func RunUsingChroot(spec *specs.Spec, bundlePath, homeDir string, stdin io.Reader, stdout, stderr io.Writer) (err error) {
 	var confwg sync.WaitGroup
-
+	var homeFound bool
+	for _, env := range spec.Process.Env {
+		if strings.HasPrefix(env, "HOME=") {
+			homeFound = true
+			break
+		}
+	}
+	if !homeFound {
+		spec.Process.Env = append(spec.Process.Env, fmt.Sprintf("HOME=%s", homeDir))
+	}
 	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
 
--- a/chroot/run_test.go
+++ b/chroot/run_test.go
@@ -90,7 +90,7 @@ func testMinimal(t *testing.T, modify func(g *generate.Generator, rootDir, bundl
 	}
 
 	output := new(bytes.Buffer)
-	if err := RunUsingChroot(g.Config, bundleDir, new(bytes.Buffer), output, output); err != nil {
+	if err := RunUsingChroot(g.Config, bundleDir, "/", new(bytes.Buffer), output, output); err != nil {
 		t.Fatalf("run: %v: %s", err, output.String())
 	}
 
--- a/pkg/chrootuser/user.go
+++ b/pkg/chrootuser/user.go
@@ -18,7 +18,7 @@ var (
 // it will use the /etc/passwd and /etc/group files inside of the rootdir
 // to return this information.
 // userspec format [user | user:group | uid | uid:gid | user:gid | uid:group ]
-func GetUser(rootdir, userspec string) (uint32, uint32, error) {
+func GetUser(rootdir, userspec string) (uint32, uint32, string, error) {
 	var gid64 uint64
 	var gerr error = user.UnknownGroupError("error looking up group")
 
@@ -26,7 +26,7 @@ func GetUser(rootdir, userspec string) (uint32, uint32, error) {
 	userspec = spec[0]
 	groupspec := ""
 	if userspec == "" {
-		return 0, 0, nil
+		return 0, 0, "/", nil
 	}
 	if len(spec) > 1 {
 		groupspec = spec[1]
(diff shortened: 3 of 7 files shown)


chroot/run.go
```diff
@@ -84,9 +84,17 @@ type runUsingChrootExecSubprocOptions struct {
...
 // directories, and mountpoints.
-func RunUsingChroot(spec *specs.Spec, bundlePath string, stdin io.Reader, stdout, stderr io.Writer) (err error) {
+func RunUsingChroot(spec *specs.Spec, bundlePath, homeDir string, stdin io.Reader, stdout, stderr io.Writer) (err error) {
 	var confwg sync.WaitGroup
-
+	var homeFound bool
+	for _, env := range spec.Process.Env {
+		if strings.HasPrefix(env, "HOME=") {
+			homeFound = true
+		}
```
rhatdan: Add `break`

chroot/run.go
```diff
@@ -84,9 +84,17 @@ type runUsingChrootExecSubprocOptions struct {
...
+	}
+	if !homeFound {
+		spec.Process.Env = append(spec.Process.Env, fmt.Sprintf("HOME=%s", homeDir))
```
rhatdan: Move this to the line where you find HOME= and then remove homeFound variable.

pkg/chrootuser/user.go
```diff
@@ -18,15 +18,19 @@ var (
 // it will use the /etc/passwd and /etc/group files inside of the rootdir
 // to return this information.
 // userspec format [user | user:group | uid | uid:gid | user:gid | uid:group ]
-func GetUser(rootdir, userspec string) (uint32, uint32, error) {
+func GetUser(rootdir, userspec string) (uint32, uint32, string, error) {
 	var gid64 uint64
 	var gerr error = user.UnknownGroupError("error looking up group")
-
+	usr, uerr := user.Current()
```
rhatdan: Nit: why uerr versus err?

run_linux.go
```diff
@@ -129,8 +129,8 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	if err != nil {
 		return err
 	}
-
-	if err := b.configureUIDGID(g, mountPoint, options); err != nil {
+	var homeDir string
+	if homeDir, err = b.configureUIDGID(g, mountPoint, options); err != nil {
```
rhatdan: nit:
I would prefer
```
homeDir, err := b.configureUIDGID(g, mountPoint, options)
if err != nil {
```

pkg/chrootuser/user.go
nalind: This is the user who's running buildah, and not the value that might be set in the `/etc/passwd` in the rootfs, which is the value we want to read, and default to `/` if we can't find anything in there.

chroot/run.go
TomSweeneyRedHat: Good thought, done

chroot/run.go
TomSweeneyRedHat: I don't think I can.  I only want to add the variable if it's not found.  If I move it to 'homeFound = true' (line 92), I'd be adding the envvar only when it existed.  We have to go through the entire slice to make sure it's not there.

pkg/chrootuser/user.go
TomSweeneyRedHat: just to specify user error and we're using the same name below at line 35 in pre-existing code.

TomSweeneyRedHat: Thanks @nalind.  If you've a quick easy way to do that, please let me know, otherwise I'll go gronking.  If there's nothing found in the rootfs /etc/passwd too, is this a reasonable last chance value to add?

run_linux.go
TomSweeneyRedHat: Switched over.

TomSweeneyRedHat:
@vrothberg thx for the test heads up.  I'm a bit surprised 'make' in my sandbox didn't catch that.....

chroot/run.go
rhatdan: Ok I read it wrong. I thought you were adding when it was found...

rhatdan:
Add this patch.
[lookuphome.txt](https://github.com/containers/buildah/files/3188739/lookuphome.txt)

pkg/chrootuser/user.go
nalind: You'll want to look at `parseNextPasswd()` in `chrootuser/user_unix.go`, and bubble the value it returns (and that its callers return) up to where it's needed.  The last-ditch default that `runc` uses (per [`setupUser`](https://github.com/opencontainers/runc/blob/b9b6cc6e47fe4f2aa4f744a1fc62d248c182d28d/libcontainer/init_linux.go#L240)) is "/".

rhatdan: My patch below allows you to get the homedir from the /etc/passwd in the container based on the UID.  Basically what @nalind is suggesting.

TomSweeneyRedHat:
Repushed with @rhatdan's homedir lookup changes.  As a root user, the userspec that's returned at the top of the GetUser() function appears to be empty and we're just defaulting all the time in that root.  I need to try as a non-root user in the morning.

chroot/run_test.go
```diff
@@ -90,7 +90,7 @@ func testMinimal(t *testing.T, modify func(g *generate.Generator, rootDir, bundl
 	}
 
 	output := new(bytes.Buffer)
-	if err := RunUsingChroot(g.Config, bundleDir, new(bytes.Buffer), output, output); err != nil {
+	if err := RunUsingChroot(g.Config, bundleDir, "/root", new(bytes.Buffer), output, output); err != nil {
```
rhatdan: Should this be "/" rather then "/root"

TomSweeneyRedHat: Ayup, missed that, ty.

rhatdan:
bot, retest this please

TomSweeneyRedHat:
@nalind @rhatdan I keep getting a compile error in Darwin that I'm not seeing in Travis or RedHat CI, nor my sandbox.  Any clues as to what I need to do to fix that?

```
### COMPLETED IN 0s
>>> cd $GOSRC/buildah && make darwin
GOOS=darwin go build -ldflags '-X main.GitCommit=03ce3cff -X main.buildInfo=1558263023 -X main.cniVersion=v0.7.0-rc2'  -o buildah.darwin -tags "containers_image_openpgp" ./cmd/buildah
# github.com/containers/buildah/pkg/chrootuser
pkg/chrootuser/user.go:68:18: undefined: lookupHomedirInContainer
make: *** [Makefile:39: darwin] Error 2
### EXITED WITH CODE 2 AFTER 6s
```

rhatdan:
You need to add that function to

buildah/pkg/chrootuser/user_basic.go

As not supported.

TomSweeneyRedHat:
@rhatdan thx, will see how that goes, have just added and repushed.

rhatdan:
@rh-atomic-bot r+

PR #1596, opened 2019-05-16
Avoid fmt.Printf() in the library

nalind:
Avoid calling fmt.Printf() to print things in library logic, which can't be controlled or suppressed by callers.  Prefer returning values and printing them in our CLI wrapper, as callers would.

### DIFF
--- a/cmd/buildah/pull.go
+++ b/cmd/buildah/pull.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM assuming happy tests.

rhatdan:
@rh-atomic-bot r+

Issue #1600, opened 2019-05-18
Build fails on Centos / older kernel


epruesse:
While building skopeo for conda-forge, the 0.1.36 pulling in the newest buildah fails with

```
unshare.c:6:10: fatal error: linux/memfd.h: No such file or directory
```

(See https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=36049&view=logs&jobId=1bf226d3-0e2f-52d8-fa93-7d9e633347b3)

Since `unshare.c` does not actually use the `MFG_HUGE` defines, it should in theory suffice to `#include <sys/mman.h>`?

rhatdan:
Could you open a PR to fix this in Buildah.

rhatdan:
Looks like this is fixed.

PR #1602, opened 2019-05-18
Avoid including linux/memfd.h

epruesse:
The file is not available on older distros.

xref:  #1600, containers/skopeo#643

### DIFF
--- a/pkg/unshare/unshare.c
+++ b/pkg/unshare/unshare.c
@@ -3,7 +3,7 @@
 #include <sys/ioctl.h>
 #include <sys/stat.h>
 #include <sys/syscall.h>
-#include <linux/memfd.h>
+#include <sys/mman.h>
 #include <fcntl.h>
 #include <grp.h>
 #include <sched.h>


epruesse:
Let's hope this builds...

epruesse:
ping @rhatdan @TomSweeneyRedHat

rhatdan:
Looks like ubuntu is not too happy.

rhatdan:
@epruesse Can we do something to support Ubuntu?

giuseppe:
we need to define `MFD_ALLOW_SEALING` and `MFD_CLOEXEC`

epruesse:
Yes.... This was just an attempt. I did not have time to research much.

According to `man memfd_create`, the interface was added with kernel 3.17 and glibc support with 2.27. What exactly does `unshare.c` do actually? Since the interface is linux specific, it won't work on e.g. OSX, so it can't be required. Is it possible to simply disable building it when `memfd.h` is not found?

giuseppe:
The memfd is used to fix a potential vulnerability where the container can get access to the host via `/proc/self/exe`.  Not sure what is the way to check for its presence without something like autoconf.  In any case, `unshare.c` is specific to Linux so portability doesn't matter.  If we want to drop it, it is enough I think, to define the missing `MFD_ALLOW_SEALING` and `MFD_CLOEXEC`

rhatdan:
I agre lets define those if they are not defined after inporting the mman.h file.

rhatdan:
@epruesse Is this still something you are interested in?

rhatdan:
@epruesse I will take this over.

Replaced by  #1655

epruesse:
thanks @rhatdan

PR #1604, opened 2019-05-21
Remove containerstorage workaround, move buildahimage under contrib

TomSweeneyRedHat:
Moves the buildahimage directory from the top-level to under contrib.
Also removes this link workaround for containers storage:

```
RUN mkdir /root/.config && ln -s /etc/containers /root/.config/containers
```

Things seem to work without it, however that directory is not created now and
I'm not 100% sure that's expected.

@rhatdan if any of this doesn't jive, please holler.

Signed-off-by: TomSweeneyRedHat <tsweeney@redhat.com>

PR #1605, opened 2019-05-21
Rework buildahimamges

rhatdan:
mv images to contrib subdir
Remove code to setup link between /root/container->/var/lib/container
This is no longer needed.


rhatdan:
@TomSweeneyRedHat PTAL

TomSweeneyRedHat:
#1604 beat you to the punch, except the entrypoint.  I'll close that one out.

vbatts:
cc myself

TomSweeneyRedHat:
The addition of the entrypoint is blowing up all of our prior examples.  At the very least, the README.md will need to be updated.   The line:

`podman run stable buildah version` has to have 'buildah' removed.  The other example showing the detach call then the execing into the container do not work.  I'm not sure if this will or will not affect what @vbatts does with the image.

In addition, the [blog](https://buildah.io/blogs/2019/05/14/buildah-blocks-images.html) article that just went out doesn't work, and the corresponding [demo](https://github.com/containers/Demos/tree/master/building/myweather) needs to be updated or done away with.

vbatts:
Are there regression tests on these images? There are increasing amounts of
downstream use of these images. These breakages will likely have fan-out
effects.

rhatdan:
The question, I have is whether or not ENTRYPOINT=/usr/bin/buildah is required enough to make this breaking change now or we have to live with this forever.  
We need to get this decide before my big article on Buildah container image goes out.

TomSweeneyRedHat:
I'm not a big fan of ENTRYPOINT.  I know it makes the container tighter and more secure, but I find it makes things a lot less flexible if you just want to play/experiment.  If I'd to make the call I'd not include it, but would add an example to the README.MD showing how you could use it to productize/secure the container even further.

rhatdan:
OK, ENTRYPOINT is removed.  I agree it can be a pain in the butt.

TomSweeneyRedHat:
LGTM
@vbatts can you throw a LGTM on here too and I'll merge?

TomSweeneyRedHat:
@rh-atomic-bot r+

vbatts:
I have updated the configs of my scheduled builds, to build from this new path now, and kicked of builds for the [stable](https://quay.io/repository/openshift-pipeline/buildah/build/43578948-5083-460b-a909-d300dea0916a), [testing](https://quay.io/repository/openshift-pipeline/buildah/build/c34edf37-8714-4831-b91b-908cdf2e26ac) and [upstream](https://quay.io/repository/openshift-pipeline/buildah/build/77ec5188-fd4c-457a-8b02-63d41480ff72) images

PR #931, opened 2018-08-15
Expand variables names for --env

rhatdan:
Fixes https://github.com/projectatomic/buildah/issues/929

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

PR #1607, opened 2019-05-21
config: expand variables only at the command line

nalind:
Skip our own attempt to expand variables in the `SetEnv()` API, since we need to let the Dockerfile parser be the only place that happens when we're building using a Dockerfile in order to fix https://bugzilla.redhat.com/show_bug.cgi?id=1712245.

Move the expansion logic introduced in #931 from `SetEnv()` to the CLI, and have it log an error when it encounters a variable that it can't expand.

### DIFF
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@vrothberg @giuseppe @TomSweeneyRedHat PTAL

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
@rh-atomic-bot r+

Issue #1598, opened 2019-05-17
COPY in Dockerfile fails on a symlink


jkryl:
I have a problem when using Dockerfile to build an image.  It looks as if symlinks in COPY target were not processed properly. I get the following error message:

```
STEP 6: COPY moac /moac/
error building at STEP "COPY moac /moac/": error creating symbolic link to "../esprima/bin/esparse.js" at "/var/lib/containers/storage/overlay/f8c017fd9cb563cf10e1e4b7c0a3f5058fda0660375813cbb8381c78d77d94e7/merged/moac/node_modules/.bin/esparse": symlink ../esprima/bin/esparse.js /var/lib/containers/storage/overlay/f8c017fd9cb563cf10e1e4b7c0a3f5058fda0660375813cbb8381c78d77d94e7/merged/moac/node_modules/.bin/esparse: file exists
```
The situation on disk looks as follows. `node_modules/.bin` contains symlinks which are pointing to executable files inside node modules, so the target is something like `../npm-module/bin/cmd.js`. `docker build` works fine, but buildah fails on these symlinks. Does it ring a bell?

I'm using dockerfile:
```
FROM node:10

RUN mkdir /moac
WORKDIR /moac
COPY moac/package*.json ./
RUN npm install
COPY moac /moac/
RUN mkdir -p /mayastor-grpc/proto
COPY mayastor-grpc/proto /mayastor-grpc/proto

ENTRYPOINT ["/moac/main.js"]
```
buildah version:
```
Version:         1.9.0-dev
Go Version:      go1.10.4
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.4.0
libcni Version:  
Git Commit:      
Built:           Tue May 14 23:36:19 2019
OS/Arch:         linux/amd64
```
uname -a:
```
Linux mayastor-dev 5.0.0-13-generic #14-Ubuntu SMP Mon Apr 15 14:59:14 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```

TomSweeneyRedHat:
Thanks for the issue @jkryl , will take a look

TomSweeneyRedHat:
[...]

PR #1623, opened 2019-05-29
addHelperSymlink: clear the destination on os.IsExist errors

nalind:
If we fail to create a symbolic link because the destination already exists, attempt to remove the destination.  This should help with #1598.


rhatdan:
LGTM
@giuseppe @vrothberg @giuseppe PTAL

giuseppe:
LGTM

@rh-atomic-bot r+

PR #1624, opened 2019-05-29
Lint fixes

vrothberg:


vrothberg:
@TomSweeneyRedHat, here's a bunch of fixes for some reports of golangci-lint. Feel free to cherry-pick the commits in case the PR won't make it in before my PTO.

chroot/run.go
```diff
@@ -491,7 +491,13 @@ func runUsingChroot(spec *specs.Spec, bundlePath string, ctty *os.File, stdin io
 		return 1, err
 	}
 	defer func() {
-		undoIntermediates()
+		if undoErr := undoIntermediates(); undoErr != nil {
+			if err == nil {
```
rhatdan: I don't think this check is necessary.
I think if you do
err = errors.Wrap(nil, undoErr.Error())
It will do the right thing.

cmd/buildah/common_test.go
```diff
@@ -2,11 +2,12 @@ package main
 
 import (
 	"flag"
-	"github.com/spf13/cobra"
 	"os"
 	"os/user"
 	"testing"
 
+	"github.com/spf13/cobra"
+
```
rhatdan: Bad space

chroot/run.go
vrothberg: Unfortunately, it is necessary as `errors.Wrap(nil, ...)` always returns nil. See [docs](https://godoc.org/github.com/pkg/errors#Wrap).

cmd/buildah/common_test.go
vrothberg: Fixed.

cmd/buildah/run.go
```diff
@@ -67,7 +68,9 @@ func init() {
 	// TODO add-third alias for tty
 	flags.BoolVarP(&opts.terminal, "terminal", "t", false, "allocate a pseudo-TTY in the container")
 	flags.BoolVar(&opts.terminal, "tty", false, "allocate a pseudo-TTY in the container")
-	flags.MarkHidden("tty")
+	if err := flags.MarkHidden("tty"); err != nil {
+		panic(fmt.Sprintf("error marking tty as hidden: %v", err))
```
nalind: This should explicitly mention that it's about the "tty flag" to be less confusing.

cmd/buildah/images_test.go
```diff
@@ -506,15 +506,22 @@ func TestMatchesSinceImageFalse(t *testing.T) {
...
+	if err := f(); err != nil {
+		return "", err
+	}
 
 	w.Close()
 	os.Stdout = old
 	var buf bytes.Buffer
-	io.Copy(&buf, r)
+	if _, err := io.Copy(&buf, r); err != nil {
+		return "", err
```
nalind: Do we want to discard any received data here, for example if we get an "unexpected EOF" error?

cmd/buildah/run.go
vrothberg: Good catch!

cmd/buildah/images_test.go
vrothberg: Let's play it safe and revert that. golangci-lint will now ignore this case.

rhatdan:
@vrothberg What is the scoop on this PR?

vrothberg:
> @vrothberg What is the scoop on this PR?

I can continue once I'm back from PTO. Currently just browsing through GitHub recovering from hiking :^)

rhatdan:
Enjoy your last few minutes of Vacation...  :^)

vrothberg:
> Enjoy your last few minutes of Vacation... :^)

Thanks :) Still one week to go though :surfing_man:

rhatdan:
Ok make sure you don't get lost in the woods.

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
@rh-atomic-bot r+

PR #1626, opened 2019-05-30
add: use fileutils.PatternMatcher for .dockerignore

nalind:
Use a fileutiles.PatternMatcher, which works better for this case than filepath.Match, to check if an item that we're copying is marked for exclusion by .dockerignore.

### DIFF
--- a/tests/bud/dockerignore3/.dockerignore
+++ b/tests/bud/dockerignore3/.dockerignore
@@ -1,0 +1,10 @@
+# comment
+*
+!test*
+!src
+**/*.in
+src/etc
+*.md
+!README*.md
+README-secret.md
+test1.txt
--- a/add.go
+++ b/add.go
@@ -14,6 +14,7 @@ import (
 	"github.com/containers/buildah/pkg/chrootuser"
 	"github.com/containers/buildah/util"
 	"github.com/containers/storage/pkg/archive"
+	"github.com/containers/storage/pkg/fileutils"
 	"github.com/containers/storage/pkg/idtools"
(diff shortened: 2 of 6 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
@QiWang19 could you review this and give it a quick test run or three please?

QiWang19:
@TomSweeneyRedHat LGTM

tests/bud/dockerignore3/.dockerignore
```diff
@@ -0,0 +1,9 @@
+# comment
+*
+!test*
+!src
+**/*.in
+*.md
+!README*.md
+README-secret.md
+test1.txt
```
TomSweeneyRedHat: I'm not sure this is covered elsewhere, but could you create a `src/etc` directory with a couple of files in it and add 'src/etc to the end here and make sure those files don't get in?  Or do you think you've that scenario covered and I'm missing it?

TomSweeneyRedHat:
@QiWang19 TYVM!

tests/bud/dockerignore3/.dockerignore
nalind: It certainly can't hurt.  Added.

TomSweeneyRedHat:
Thanks @nalind.  LGTM assuming happy tests.

TomSweeneyRedHat:
Happy tests, going to merge this.

TomSweeneyRedHat:
@rh-atomic-bot r+

PR #1630, opened 2019-05-31
imagebuildah: handle ID mappings for COPY --from

nalind:
Fix handling of ID mapping for COPY: when copying from other containers, use their mappings, and when copying from the host, use host mappings.

### DIFF
--- a/util.go
+++ b/util.go
@@ -1,9 +1,12 @@
 package buildah
 
 import (
+	"archive/tar"
 	"io"
 	"os"
+	"path/filepath"
 
+	"github.com/containers/buildah/util"
 	"github.com/containers/image/docker/reference"
 	"github.com/containers/image/pkg/sysregistries"
 	"github.com/containers/image/pkg/sysregistriesv2"
@@ -12,7 +15,9 @@ import (
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/chrootarchive"
 	"github.com/containers/storage/pkg/idtools"
+	"github.com/containers/storage/pkg/pools"
 	"github.com/containers/storage/pkg/reexec"
+	"github.com/containers/storage/pkg/system"
 	"github.com/opencontainers/image-spec/specs-go/v1"
 	rspec "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/opencontainers/selinux/go-selinux"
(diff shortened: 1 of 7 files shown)


util.go
```diff
@@ -105,19 +110,108 @@ func convertRuntimeIDMaps(UIDMap, GIDMap []rspec.LinuxIDMapping) ([]idtools.IDMa
...
+		tarIDMappingOptions = &IDMappingOptions{
+			HostUIDMapping: true,
+			HostGIDMapping: true,
+		}
+	}
+	return func(src, dest string) error {
+		logrus.Debugf("copyFileWithTar(%s, %s)", src, dest)
+		f, err := os.Open(src)
+		if err != nil {
+			return errors.Wrapf(err, "error opening %q to digest its contents")
```
rhatdan: Missing input for %q

util.go
```diff
@@ -105,19 +110,108 @@ func convertRuntimeIDMaps(UIDMap, GIDMap []rspec.LinuxIDMapping) ([]idtools.IDMa
...
+		}
+		defer func() {
+			if f != nil {
+				f.Close()
+			}
+		}()
+
+		sysfi, err := system.Stat(src)
+		if err != nil {
+			return errors.Wrapf(err, "error reading attributes of %q")
```
rhatdan: Missing input for %q

util.go
```diff
@@ -105,19 +110,108 @@ func convertRuntimeIDMaps(UIDMap, GIDMap []rspec.LinuxIDMapping) ([]idtools.IDMa
...
+		hostUID := sysfi.UID()
+		hostGID := sysfi.GID()
+		containerUID, containerGID, err := util.GetContainerIDs(tarIDMappingOptions.UIDMap, tarIDMappingOptions.GIDMap, hostUID, hostGID)
+		if err != nil {
+			return errors.Wrapf(err, "error mapping owner IDs of %q: %d/%d", src, hostUID, hostGID)
+		}
+
+		fi, err := os.Lstat(src)
+		if err != nil {
+			return errors.Wrapf(err, "error reading attributes of %q")
```
rhatdan: Bad %q

util.go
nalind: Argh, fixed.

rhatdan:
LGTM

PR #1631, opened 2019-06-01
allow 32bit builds

### LINKED ISSUE #1629: buildah ftbfs on 32bit
**Description**
buildah ftbfs on 32bit arches.  this is causing podman to failed to build for the 32bit arches on the ubuntu ppa.

```
# github.com/containers/buildah
./add.go:300:8: cannot use atime.Unix() (type int64) as type int32 in field value
./add.go:300:27: cannot use atime.UnixNano() % 1000000000 (type int64) as type int32 in field value
./add.go:301:8: cannot use mtime.Unix() (type int64) as type int32 in field value
./add.go:301:27: cannot use mtime.UnixNano() % 1000000000 (type int64) as type int32 in field value
Makefile:34: recipe for target 'binary' failed
```
The problem area in the code is here
```
mtime := info.ModTime()
atime := mtime
times := []syscall.Timespec{
    {Sec: atime.Unix(), Nsec: atime.UnixNano() % 1000000000},
    {Sec: mtime.Unix(), Nsec: mtime.UnixNano() % 1000000000},
}
```
It looks like on 32bit systems, a syscall.timespec expects an int32 but .Unix() returns a int64 and therefore the type mismatch.  In the golang docs for syscall, they apparently prefer the use of x/sys/unix instead which happens to have a TimeToTimeSpec which handles the bitness issue (https://godoc.org/golang.org/x/sys/unix#Timespec)

However, when making this change, it required the change of some function input types and eventually spread throughout the code.

This could also be handled by making a func that returns the []syscall.TimeSpec from a _64 and _32 file as well.

I'm fine to do a PR for this, but would like to get some direction first. Should I:

* convert syscall.timespecs to unix.timespecs and follow through on where-ever that might lead me
* just make _32 and  _64 files and move on
* claim buildah should not support 32bit arches and therefore neither should podman?

baude:
on 32-bit systems, we had type mismatches when creating a
syscall.timespec.

resolves #1629

### DIFF
--- a/add.go
+++ b/add.go
@@ -293,8 +293,8 @@ func addHelper(excludes *fileutils.PatternMatcher, extract bool, dest string, de
 					mtime := info.ModTime()
 					atime := mtime
 					times := []syscall.Timespec{
-						{Sec: atime.Unix(), Nsec: atime.UnixNano() % 1000000000},
-						{Sec: mtime.Unix(), Nsec: mtime.UnixNano() % 1000000000},
+						syscall.NsecToTimespec(atime.Unix()),
+						syscall.NsecToTimespec(mtime.Unix()),
 					}
 					if info.IsDir() {
 						return addHelperDirectory(esrc, path, filepath.Join(dest, fpath), info, hostOwner, times)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM, would like to steal a head nod from @nalind

rhatdan:
@nalind Suggested this fix.
@rh-atomic-bot r+

PR #1638, opened 2019-06-03
bud.bats: add a test for WORKDIR and COPY with absolute destinations

nalind:
Add an extra test to ensure that we don't mistakenly treat absolute destination paths for COPY instructions as being relative to WORKDIR.


rhatdan:
LGTM

TomSweeneyRedHat:
@rh-atomic-bot r+

PR #1640, opened 2019-06-03
Write to setgroups when doing single user mapping

kpelzel:
There was a [recent commit](https://github.com/containers/buildah/commit/53be3d39f1cb4a182cfc3559726332f643f335c8) to add support for single user mapping without uidmap. Running buildah as an unprivileged user and without having uidmap installed results in this error:

`ERRO[0000] error writing "0 1000 1\n" to /proc/28584/gid_map: write /proc/28584/gid_map: operation not permitted
ERRO[0000] (unable to determine exit status)`

This is due to /proc/[pid]/setgroups not containing 'deny' which is required for unprivileged single user mapping. 

This pull request writes 'deny' to setgroups when doing single user mapping. It also changes the error messages so they are passed through logrus as warnings to prevent an exit status 1 error.

Signed-off-by: Kevin Pelzel <kevinpelzel22@gmail.com>

### DIFF
--- a/pkg/unshare/unshare.go
+++ b/pkg/unshare/unshare.go
@@ -195,13 +195,25 @@ func (c *Cmd) Start() error {
 				if err == nil {
 					gidmapSet = true
 				} else {
-					fmt.Fprintf(continueWrite, "error running newgidmap: %v: %s", err, g.String())
-					fmt.Fprintf(continueWrite, "falling back to single mapping\n")
+					logrus.Warnf("error running newgidmap: %v: %s", err, g.String())
+					logrus.Warnf("falling back to single mapping")
 					g.Reset()
 					g.Write([]byte(fmt.Sprintf("0 %d 1\n", os.Getegid())))
 				}
 			}
 			if !gidmapSet {
+				if c.UseNewgidmap {
+					setgroups, err := os.OpenFile(fmt.Sprintf("/proc/%s/setgroups", pidString), os.O_TRUNC|os.O_WRONLY, 0)
+					if err != nil {
+						fmt.Fprintf(continueWrite, "error opening /proc/%s/setgroups: %v", pidString, err)
+						return errors.Wrapf(err, "error opening /proc/%s/setgroups", pidString)
+					}
+					defer setgroups.Close()
+					if _, err := fmt.Fprintf(setgroups, "deny"); err != nil {
+						fmt.Fprintf(continueWrite, "error writing 'deny' to /proc/%s/setgroups: %v", pidString, err)
+						return errors.Wrapf(err, "error writing 'deny' to /proc/%s/setgroups", pidString)
+					}
+				}
 				gidmap, err := os.OpenFile(fmt.Sprintf("/proc/%s/gid_map", pidString), os.O_TRUNC|os.O_WRONLY, 0)
 				if err != nil {
 					fmt.Fprintf(continueWrite, "error opening /proc/%s/gid_map: %v", pidString, err)
@@ -232,8 +244,8 @@ func (c *Cmd) Start() error {
 				if err == nil {
 					uidmapSet = true
 				} else {
-					fmt.Fprintf(continueWrite, "error running newuidmap: %v: %s", err, u.String())
-					fmt.Fprintf(continueWrite, "falling back to single mapping\n")
+					logrus.Warnf("error running newuidmap: %v: %s", err, u.String())
+					logrus.Warnf("falling back to single mapping")
 					u.Reset()
 					u.Write([]byte(fmt.Sprintf("0 %d 1\n", os.Geteuid())))
 				}


pkg/unshare/unshare.go
```diff
@@ -195,13 +195,25 @@ func (c *Cmd) Start() error {
 				if err == nil {
 					gidmapSet = true
 				} else {
-					fmt.Fprintf(continueWrite, "error running newgidmap: %v: %s", err, g.String())
-					fmt.Fprintf(continueWrite, "falling back to single mapping\n")
+					logrus.Warnf("error running newgidmap: %v: %s", err, g.String())
```
rhatdan: I don't think these messages show to the user by default?  Do the messages still show on the terminal?

pkg/unshare/unshare.go
```diff
@@ -195,13 +195,25 @@ func (c *Cmd) Start() error {
...
+					logrus.Warnf("falling back to single mapping")
 					g.Reset()
 					g.Write([]byte(fmt.Sprintf("0 %d 1\n", os.Getegid())))
 				}
 			}
 			if !gidmapSet {
+				if c.UseNewgidmap {
+					setgroups, err := os.OpenFile(fmt.Sprintf("/proc/%s/setgroups", pidString), os.O_TRUNC|os.O_WRONLY, 0)
+					if err != nil {
+						fmt.Fprintf(continueWrite, "error opening /proc/%s/setgroups: %v", pidString, err)
```
rhatdan: I believe this message will get displayerd twice.  I don't think you should Fprintf the line.

pkg/unshare/unshare.go
```diff
@@ -195,13 +195,25 @@ func (c *Cmd) Start() error {
...
+						return errors.Wrapf(err, "error opening /proc/%s/setgroups", pidString)
+					}
+					defer setgroups.Close()
+					if _, err := fmt.Fprintf(setgroups, "deny"); err != nil {
+						fmt.Fprintf(continueWrite, "error writing 'deny' to /proc/%s/setgroups: %v", pidString, err)
```
rhatdan: I believe this message will get displayerd twice. I don't think you should Fprintf the line.

rhatdan:
bot, add author to whitelist
@giuseppe PTAL

giuseppe:
LGTM with the changes you've pointed out

giuseppe:
please rebase your patch, do not merge master on top of it

pkg/unshare/unshare.go
kpelzel: The logrus.Warnf messages only appear to the user if the --debug flag is on. Is that how you think this should function? When writing to continueWrite before the command would always exit with status code 1.

pkg/unshare/unshare.go
kpelzel: I was just following the same syntax as lines 219 and 224. Should we remove the Fprintf elsewhere in this function as well?

pkg/unshare/unshare.go
TomSweeneyRedHat: @nalind thoughts on why the Fprintf to continueWrite would exit out here?

kpelzel:
Whoops the merge was done on accident. I reset the the head to before the merge and force pushed to remote branch. Then ran a rebase from master.

pkg/unshare/unshare.go
nalind: [The child](https://github.com/containers/buildah/blob/fa67ecb27af4ca6faad5509d9837e05c32ed63cb/pkg/unshare/unshare.c#L229) treats any data that it reads from the reading end of the pipe as an indication that a fatal error has occurred and that it needs to exit.  Messages that aren't about fatal errors should not be written to the pipe.

kpelzel:
Related to single-user-mapping issues: https://github.com/containers/buildah/pull/1645

giuseppe:
> Whoops the merge was done on accident. I reset the the head to before the merge and force pushed to remote branch. Then ran a rebase from master.

Thanks but I see 3 commits now.  Could you amend two of them and completely drop the merge?

kpelzel:
> > Whoops the merge was done on accident. I reset the the head to before the merge and force pushed to remote branch. Then ran a rebase from master.
> 
> Thanks but I see 3 commits now. Could you amend two of them and completely drop the merge?

Ah shoot I messed that up. Would it just make more sense to close this PR in favor of https://github.com/containers/buildah/pull/1645 or would you like me to fix the commits.

giuseppe:
> Ah shoot I messed that up. Would it just make more sense to close this PR in favor of #1645 or would you like me to fix the commits.

I am fine if you want to carry the change there, but the two can also be treated separately as this is fixing a bug, while the other PR is adding a new feature

rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot retry

rhatdan:
@kpelzel This might need to be rebased and repushed. Not sure what failure is being complained about here.

kpelzel:
> @kpelzel This might need to be rebased and repushed. Not sure what failure is being complained about here.

Okay I rebased and I think I fixed the weird multiple commits issue.

rhatdan:
@rh-atomic-bot r+

PR #1645, opened 2019-06-05
[WIP] Make chown tolerant of errors when doing single user mapping

kpelzel:
**Linked to this PR:** https://github.com/containers/storage/pull/358

### Problem:
When building a container as an unprivileged user and without uidmap tools, buildah will fail when chowning system files. This patch ignores those chown errors if buildah has fallen back to single user mapping mode.

### What this patch changes:
**Buildah side:** 
unshare.go checks if newuidmap and newgidmap is installed by running "which newuidmap". If uidmap tools are not installed it sets an environment variable to pass down to the child.

**containers/storage side:**
diff_unix.go passes in the hosts environment variables (including the single user mapping environment variable) to the "storage-applyLayer" command. This allows checkChownErr in idtools.go to check if it needs to ignore the chown errors.

### Notes:
I'm not completely sure if using environment variables is the best way to get this information down to checkChownErr, but it seemed to work best in terms minimal code change. I would love to hear feedback on better solutions.

### DIFF
--- a/pkg/unshare/unshare.go
+++ b/pkg/unshare/unshare.go
@@ -97,6 +97,28 @@ func (c *Cmd) Start() error {
 		c.ExtraFiles = append(c.ExtraFiles, c.Ctty)
 	}
 
+	// Check if uidmap tools are available
+	useNewgidmap := c.UseNewgidmap
+	useNewuidmap := c.UseNewuidmap
+	if useNewgidmap {
+		if err := checkGIDMap(); err != nil {
+			useNewgidmap = false
+			logrus.Warnf("%v", err)
+		}
+	}
+	if useNewuidmap {
+		if err := checkUIDMap(); err != nil {
+			useNewuidmap = false
+			logrus.Warnf("%v", err)
+		}
+	}
+
+	// Set environment variable if uidmap tools are not available
+	if (c.UseNewgidmap && !useNewgidmap) || (c.UseNewuidmap && !useNewuidmap) {
+		c.Env = append(c.Env, "_CONTAINERS_SINGLE_MAPPING=0")
+		logrus.Warnf("falling back to single mapping")
+	}
+
 	// Make sure we clean up our pipes.
 	defer func() {
 		if pidRead != nil {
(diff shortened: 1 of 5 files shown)


giuseppe:
bot, add author to whitelist

pkg/unshare/unshare.go
```diff
@@ -558,3 +586,25 @@ func ParseIDMappings(uidmap, gidmap []string) ([]idtools.IDMap, []idtools.IDMap,
 	}
 	return uid, gid, nil
 }
+
+func checkGIDMap() error {
+	cmd := exec.Command("which", "newgidmap")
```
giuseppe: I think we want to use `exec.LookPath` here instead of running another process

pkg/unshare/unshare.go
```diff
@@ -558,3 +586,25 @@ func ParseIDMappings(uidmap, gidmap []string) ([]idtools.IDMap, []idtools.IDMap,
...
+	cmd.Stderr = nil
+	err := cmd.Run()
+	if err != nil {
+		return errors.Wrapf(err, "command 'which newgidmap' failed. Is newgidmap in $PATH?")
+	}
+	return nil
+}
+
+func checkUIDMap() error {
+	cmd := exec.Command("which", "newuidmap")
```
giuseppe: same here

kpelzel:
In the most recent commit I added a "single-user-mapping" global flag to buildah. This will set the environment variable that tells idtools to ignore chown errors. I didn't know what to name the flag so feedback is welcome.

I also moved most of the uidmap checking logic into MaybeReexecUsingUserNamespace because I thought it made the code more readable, but let me know if that's a bad idea.

rhatdan:
Changing to WIP flag, to make sure this does not get merged until we get the containers/storage patches merged and figure out how this should interact with containers/storage.

kpelzel:
Closing as this issue is resolved in https://github.com/containers/buildah/pull/1749

rhatdan:
@kpelzel Once we get this support into podman 1.5, we would love to have further feedback on what you would need to make podman a good solution for HPC environments.

PR #1646, opened 2019-06-05
imagebuildah: simplify screening by top layers

nalind:
We only care about the most recent top layer when we're checking if a candidate image might be based on the one that's the base for the current working container, so don't bother tracking more than that.


rhatdan:
LGTM

TomSweeneyRedHat:
@rh-atomic-bot r+

Issue #1635, opened 2019-06-03
Unpredictable build-as-dockerfile caching with --layers and --build-arg


praiskup:
**Description**

Unpredictable caching in `buildah bud`.

**Steps to reproduce the issue:**

**1. create dockerfile**
```
$ cat Dockerfile
FROM registry.fedoraproject.org/fedora:30
MAINTAINER jdoe <jdoe@example.com>
ENV container="docker"

RUN echo this-should-be-cached-but-it-s-not && sleep 3

ARG USERNAME
ARG UID
ARG CODE
ARG PGDATA
ARG PORT=55555

CMD ["/container-run"]
```
**2. Build for the first time:**
```
$ buildah bud --layers --build-arg=UID=17122 --build-arg=CODE=/copr/coprs_frontend --build-arg=USERNAME=praiskup --build-arg=PGDATA=/pgdata .
STEP 1: FROM registry.fedoraproject.org/fedora:30
STEP 2: MAINTAINER jdoe <jdoe@example.com>
--> Using cache e2d50ffd8947a4880814edfdd0a62955c927fafe6201ba2bdbffcf7f9f93868e
STEP 3: ENV container="docker"
--> Using cache 1336ba0d55cc96001a6a87ca7b1a3090b4c4bc7caf79df7259dc194435546237
STEP 4: RUN echo this-should-be-cached-but-it-s-not && sleep 3
--> Using cache da075d5c0c020a836d5bbfd946fdc9f413cf288a6570264c998578d4f2095f3a
STEP 5: ARG USERNAME
--> Using cache 44eb7b2838dac2123ba077da8a5176df6a65ae5daff8ef3f278b944a85b5ea78
STEP 6: ARG UID
--> Using cache 3fc9fb1edaf9baac057e6b5e52fc6e623f487fabafe70c0a8550e006fa8aa675
STEP 7: ARG CODE
--> e55775ddb455bbac98806c94b13caa27d1e685dbab4b4597b131d621af2b1b78
STEP 8: ARG PGDATA
--> b041ea6991b06c87b301131d74f84f911d759092f51161002fb4fdd15a61c3fd
STEP 9: ARG PORT=55555
--> 9f4b2e7c224eb642ca4dd73032757ed906121e4d9124424503b406482f026ca8
STEP 10: CMD ["/container-run"]
STEP 11: COMMIT
--> c921fdec80e4eb1cd5bc856010947437ebb65cafa783fa5f8975a4566b91da77
```
[...]

PR #1647, opened 2019-06-05
imagebuildah: use a stable sort for comparing build args

nalind:
When checking if a history entry in a candidate image that represents a RUN instruction matches what we'll do if the image is not a valid cached image, sort the arguments.  This should fix https://github.com/containers/buildah/issues/1635.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -1355,6 +1355,7 @@ func (b *Executor) getBuildArgs() string {
 			buildArgs = append(buildArgs, k+"="+v)
 		}
 	}
+	sort.Strings(buildArgs)
 	return strings.Join(buildArgs, " ")
 }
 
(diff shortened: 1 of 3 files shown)


nalind:
An alternative would be replacing the `Args` map with a slice of `struct { Name, Value string }` values, which is what go-dockerclient uses for passing them to a remote builder, but switching to that here would require changes both here and in imagebuilder.

rhatdan:
LGTM
Anyway we can add a test for this?

nalind:
It's not a guaranteed trigger, but this should hit it the error in at least some cases.

rhatdan:

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
@rh-atomic-bot r+

PR #1649, opened 2019-06-06
Support --dns=none like podman

jordemort:
This PR adds support for `--dns=none`, which will avoid setting up a bind mount for `/etc/resolv.conf`. This is equivalent to podman's `--dns=none`, which was added in https://github.com/containers/libpod/pull/2747

It looks like it was possible at one point to avoid a bind mount on `/etc/resolv.conf` as a side-effect of passing network configuration options in https://github.com/containers/buildah/pull/1146 but it appears this way has been factored out of the code since then. This gives "I don't want resolv.conf" a dedicated knob to twist, which is shaped and labeled identically to the knob in podman :)

I may eventually want to add an equivalent podman's of `--no-hosts` in a separate PR, but I left that off for now since it would be a bit more involved and since I don't personally need it just yet.

### DIFF
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -189,6 +189,10 @@ soley for scripting compatibility.
 
 Set custom DNS servers
 
+This option can be used to override the DNS configuration passed to the container. Typically this is necessary when the host DNS configuration is invalid for the container (e.g., 127.0.0.1). When this is the case the `--dns` flag is necessary for every run.
+
+The special value **none** can be specified to disable creation of /etc/resolv.conf in the container by Buildah. The /etc/resolv.conf file in the image will be used without changes.
+
 **--dns-option**=[]
 
 Set custom DNS options
--- a/docs/buildah-from.md
+++ b/docs/buildah-from.md
@@ -171,6 +171,22 @@ The [username[:password]] to use to authenticate with the registry if required.
 If one or both values are not supplied, a command line prompt will appear and the
 value can be entered.  The password is entered without echo.
 
+**--dns**=[]
+
+Set custom DNS servers
+
+This option can be used to override the DNS configuration passed to the container. Typically this is necessary when the host DNS configuration is invalid for the container (e.g., 127.0.0.1). When this is the case the `--dns` flag is necessary for every run.
+
+The special value **none** can be specified to disable creation of /etc/resolv.conf in the container by Buildah. The /etc/resolv.conf file in the image will be used without changes.
+
+**--dns-option**=[]
+
+Set custom DNS options
+
+**--dns-search**=[]
+
+Set custom DNS search domains
+
 **--http-proxy**
 
 By default proxy environment variables are passed into the container if set
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -37,6 +37,7 @@ func CommonBuildOptions(c *cobra.Command) (*buildah.CommonBuildOptions, error) {
 	var (
 		memoryLimit int64
 		memorySwap  int64
+		noDNS       bool
 		err         error
 	)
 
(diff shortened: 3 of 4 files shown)


rhatdan:
Can you update the man page of podman from and podman build to include this documentation?  
Should we throw an error if the user specifies two DNS Options with one of them being NONE?

@jordemort You also need to sign your commit.

git commit -a -s --amend
git push --force

TomSweeneyRedHat:
bot, add author to whitelist

TomSweeneyRedHat:
@rhatdan I think you meant update man for buildah from and buildah bud

jordemort:
@rhatdan The commit is signed, with key ID C341950C47B6CE14 - maybe because I wasn't in the whitelist?

I'll add in the documentation changes.

If the user specifies `--dns=none,somethingelse` then the length check won't match and `buildah` will go ahead and try and use `none` as a DNS server.  Eventually this will lead to the following error message when the string `none` is passed to `getDNSIP`:

```error getting DNS servers: invalid IP address none```

**Comment by jordemort** [comment]
Stole the documentation for this out of the podman docs. Also added the `--dns*` options to `buildah-from.md`, which was missing them entirely.

File: docs/buildah-bud.md [hunk 1]
```diff
@@ -189,6 +189,10 @@ soley for scripting compatibility.
 
 Set custom DNS servers
 
+This option can be used to override the DNS configuration passed to the container. Typically this is necessary when the host DNS configuration is invalid for the container (e.g., 127.0.0.1). When this is the case the --dns flag is necessary for every run.
+
+The special value none can be specified to disable creation of /etc/resolv.conf in the container by Buildah. The /etc/resolv.conf file in the image will be used without changes.
```
rhatdan: Could you highlight `none`

docs/buildah-from.md
```diff
@@ -171,6 +171,22 @@ The [username[:password]] to use to authenticate with the registry if required.
 If one or both values are not supplied, a command line prompt will appear and the
 value can be entered.  The password is entered without echo.
 
+**--dns**=[]
+
+Set custom DNS servers
+
+This option can be used to override the DNS configuration passed to the container. Typically this is necessary when the host DNS configuration is invalid for the container (e.g., 127.0.0.1). When this is the case the --dns flag is necessary for every run.
+
+The special value none can be specified to disable creation of /etc/resolv.conf in the container by Buildah. The /etc/resolv.conf file in the image will be used without changes.
```
rhatdan: Highlight `none`

rhatdan:
@jordemort The validator is looking for a string like:

In your commit.

Also can you squash you PR's together.

I think I can do

buildah from --dns none --dns 8.8.8.8 fedora 

Which should throw an error.

jordemort:
@rhatdan Ok, I will squash and add the DCO

Multiple `--dns` options seem to be the same as a single `--dns` option with multiple values:

```
root@a0343e7ab207:/app# buildah from --dns none --dns 8.8.8.8 docker.io/fedora
Copying blob 34b3503aa97f done
Copying config 9754ce1464 done
Storing signatures
fedora-working-container
root@a0343e7ab207:/app# buildah run -- fedora-working-container bash
error getting dns servers: invalid IP address none
```

Note that we don't hit the error until `buildah run`; I wonder if it would be better to add something to catch this and error at `buildah from`?

rhatdan:
@jordemort That is what I want, catch it in buildah bud and buildah run CLI processing rather then waiting for a container to run.  I don't think you can fix it after the fact.

jordemort:
Added some stuff in `pkg/parse/parse.go` to catch conflicts. Go isn't even my 10th best language so please let me know if my code is ugly.

```
root@a0343e7ab207:/app# buildah from --dns none --dns 8.8.8.8 docker.io/fedora
invalid --dns, --dns=none may not be used with any other --dns options
root@a0343e7ab207:/app# buildah from --dns none --dns-search chicken docker.io/fedora
invalid --dns-search, --dns-search may not be used with --dns=none
root@a0343e7ab207:/app# buildah from --dns none --dns-option chicken docker.io/fedora
invalid --dns-option, --dns-option may not be used with --dns=none
root@a0343e7ab207:/app# buildah from --dns none docker.io/fedora
fedora-working-container-1
root@a0343e7ab207:/app# buildah run -- fedora-working-container-1 cat /etc/resolv.conf 
# Generated by NetworkManager
nameserver 192.168.122.1
root@a0343e7ab207:/app# cat /etc/resolv.conf
# This file is included on the metadata iso
nameserver 192.168.65.1
domain webb.haus
root@a0343e7ab207:/app# buildah from --dns 8.8.8.8 docker.io/fedora
fedora-working-container-2
root@a0343e7ab207:/app# buildah run -- fedora-working-container-2 cat /etc/resolv.conf 
nameserver 8.8.8.8
```

jordemort:
Wait, there's a bug here

jordemort:
I had `if len(dnsServers) > 1` instead of `if noDNS && len(dnsServers) > 1`, which would have flat-out prevented multiple DNS servers at all. Fixed.

pkg/parse/parse.go
```diff
@@ -67,9 +68,28 @@ func CommonBuildOptions(c *cobra.Command) (*buildah.CommonBuildOptions, error) {
 		}
 	}
 
+	noDNS = false
 	dnsServers, _ := c.Flags().GetStringSlice("dns")
+	if len(dnsServers) > 0 {
```
rhatdan: No need for this check

jordemort: Check removed

rhatdan:
LGTM
@TomSweeneyRedHat @giuseppe @nalind @vrothberg PTAL

giuseppe:
LGTM

giuseppe:
@rh-atomic-bot r+

Issue #1650, opened 2019-06-07
Piping into `buildah run` never sends EOF on child stdin


jordemort:
**Description**

When I pipe input into `buildah run`, it seems that stdin never reaches EOF in the child process. This problem appears to keep pipelining from working with Ansible's connection module for buildah.

**Steps to reproduce the issue:**
1. `buildah from docker.io/debian:stretch`
2. `echo "echo foo" | buildah run -- debian-working-container /bin/sh`

**Describe the results you received:**

`foo` is printed on my terminal, but `buildah run` never exits.

**Describe the results you expected:**

`foo` is printed on my terminal, and `buildah run` exits, as if I had run `echo "echo foo" | /bin/sh`

**Output of `rpm -q buildah` or `apt list buildah`:**

```
Listing... Done
buildah/stretch,now 1.8.2+github2 amd64 [installed]
N: There is 1 additional version. Please use the '-a' switch to see it
```

(I built this myself from the stretch-backport branch of https://github.com/jordemort/buildah)

**Output of `buildah version`:**

```
Version:         1.8.2
Go Version:      go1.11.5
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.4.0
libcni Version:  
Git Commit:      
Built:           Thu Jun  6 22:33:30 2019
OS/Arch:         linux/amd64
```

**Output of `podman version` if reporting a `podman build` issue:**

```
Version:            1.3.1
RemoteAPI Version:  1
Go Version:         go1.11.5
OS/Arch:            linux/amd64
```

(I built this myself from the stretch-backport branch of https://github.com/jordemort/libpod)

**Output of `cat /etc/*release`:**

```
PRETTY_NAME="Debian GNU/Linux 9 (stretch)"
NAME="Debian GNU/Linux"
VERSION_ID="9"
VERSION="9 (stretch)"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

**Output of `uname -a`:**

```
Linux a0343e7ab207 4.9.125-linuxkit #1 SMP Fri Sep 7 08:20:28 UTC 2018 x86_64 GNU/Linux
```

**Output of `cat /etc/containers/storage.conf`:**

```
# storage.conf is the configuration file for all tools
# that share the containers/storage libraries
# See man 5 containers-storage.conf for more information

# The "container storage" table contains all of the server options.
[...]

PR #1653, opened 2019-06-07
Always close stdin pipe

jordemort:
This is a fix for https://github.com/containers/buildah/issues/1650.  Here is what I observed when running `echo "echo foo" | buildah run -- debian-working-container /bin/sh`:

On the first iteration of the loop, the other side of stdin is closed but our side has waiting data. Because it was closed, `pollFd.Revents&unix.POLLHUP` == `unix.POLLHUP`, so the file descriptor gets added to `removes`. However, because there is data waiting, `pollFd.Revents&unix.POLLIN` != `0`, so the first chance to trigger the logic that closes stdin is not taken. Because there is waiting data, then `n` > `0`, so the second chance to trigger the logic that closes stdin is also not taken. The data is relayed, but the file descriptor is in `removes`, so it gets removed from the polling set and there are no further chances to trigger closing stdin; thus, it hangs forever.

This removes the first chance to close stdin, responds to `n` == `0` by adding the descriptor to `removes`, and then triggers the "close stdin" logic while looping through `removes`.

cc @QiWang19 

Fixes: https://github.com/containers/buildah/issues/1650

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -1155,12 +1155,6 @@ func runCopyStdio(stdio *sync.WaitGroup, copyPipes bool, stdioPipe [][]int, copy
 			}
 			// If the POLLIN flag isn't set, then there's no data to be read from this descriptor.
 			if pollFd.Revents&unix.POLLIN == 0 {
-				// If we're using pipes and it's our stdin and it's closed, close the writing
-				// end of the corresponding pipe.
-				if copyPipes && int(pollFd.Fd) == unix.Stdin && pollFd.Revents&unix.POLLHUP != 0 {
-					unix.Close(stdioPipe[unix.Stdin][1])
-					stdioPipe[unix.Stdin][1] = -1
-				}
 				continue
 			}
 			// Read whatever there is to be read.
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
But I want @nalind to review.

rhatdan:
@giuseppe PTAL

giuseppe:
LGTM, solves the issue that was reported.

Not related to this PR (buildah without the patch has the same behaviour), but a command like:

`seq 100000 | sudo ./buildah run fedora-working-container -- sh -c 'while read i; do echo $i; done'`

seems to hang

TomSweeneyRedHat:
I've spun up #1664 to make sure we look at the issue @giuseppe spotted.

giuseppe:
@TomSweeneyRedHat @rhatdan let's merge this PR and I will take a look at it

rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot retry

rhatdan:
@jordemort Can you rebase this PR and repush. Then we can get it merged.

Not sure what Homu is complaining about

giuseppe:
@rh-atomic-bot retry

giuseppe:
followup PR here: https://github.com/containers/buildah/pull/1668

PR #1662, opened 2019-06-10
fix bug --cpu-shares parsing typo

### LINKED ISSUE #1464: buildah from --cpu-shares: not working, or test is misleading
`tests/from.bats:242` (the cpu-shares test) has code that does:
```
# cid=$(buildah from --cpu-shares=2 --pull alpine)
# buildah run $cid cat /sys/fs/cgroup/cpu/cpu.shares
```
...but all it does is check that the output contains '2'. It does: the output is actually `1024`.

If 1024 is the expected value, please correct the test.

If 2 is the expected value, please investigate buildah.

QiWang19:
close #1464

fix the paring typo cpu-shared of --cpu-shares.
fix expect value in from.bats

### DIFF
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -80,7 +80,7 @@ func CommonBuildOptions(c *cobra.Command) (*buildah.CommonBuildOptions, error) {
 	}
 	cpuPeriod, _ := c.Flags().GetUint64("cpu-period")
 	cpuQuota, _ := c.Flags().GetInt64("cpu-quota")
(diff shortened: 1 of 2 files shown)


rhatdan:
Nice simple fix.  Too bad our tests hard-coded the failure.
LGTM
@TomSweeneyRedHat @giuseppe @nalind @ashley-cui PTAL

TomSweeneyRedHat:
@rh-atomic-bot r+

PR #1666, opened 2019-06-11
WIP - Switch to golangci-lint 2nd

TomSweeneyRedHat:
Changes lint to golangci-lint.  I'd to vendor a number of
packages to make this work.

This replaces #1423.  Since I spun that PR up, my sandbox
got wiped and I'm cautious about merging ontop of that.


TomSweeneyRedHat:
This still leaves 4 golint issues I'm not sure how to handle:

This first one looks perfectly OK to me, maybe I'm being oblivious?
```
# make lint
golangci-lint run --build-tags="seccomp selinux    ostree"
imagebuildah/build.go:325:2: SA4006: this value of `archivedPath` is never used (staticcheck)
	archivedPath := filepath.Join(s.mountPoint, path)
	^
```

This one is complaining as our json has 'parent' in both the V1 and V2 image.  I could change this to 'parentv2` but was concerned that might mess up stuff elsewhere.
```
docker/types.go:164:2: structtag: struct field Parent repeats json tag "parent" also at types.go:137 (govet)
	Parent     ID            `json:"parent,omitempty"`
	^
```

I'm not sure how to handle these.  It's returning a function that returns an error.  If I don't specify it as `_ = defer policyContext.Destroy()`  lint complains.  But if I define it that way go doesn't compile.
```
imagebuildah/build.go:1196:29: Error return value of `policyContext.Destroy` is not checked (errcheck)
	defer policyContext.Destroy()
	                           ^
imagebuildah/build.go:1602:15: Error return value of `cleanup` is not checked (errcheck)
	defer cleanup()
	             ^
make: *** [Makefile:55: lint] Error 1
```

TomSweeneyRedHat:
@saschagrunert PTAL

saschagrunert:
Is it possible that the changes to the Makefile are missing within this PR? 

> ```
> # make lint
> golangci-lint run --build-tags="seccomp selinux    ostree"
> imagebuildah/build.go:325:2: SA4006: this value of `archivedPath` is never used (staticcheck)
> 	archivedPath := filepath.Join(s.mountPoint, path)
> 	^
> ```

It seems that we either overwrite this var or return the error in the following code. So there is no reason to initially assign it.

> ```
> docker/types.go:164:2: structtag: struct field Parent repeats json tag "parent" also at types.go:137 (govet)
> 	Parent     ID            `json:"parent,omitempty"`
> 	^
> ```

Yeah this is tricky and I'm not sure if this is intended. I would assume that it is and add a `// nolint` there.

> ```
> imagebuildah/build.go:1196:29: Error return value of `policyContext.Destroy` is not checked (errcheck)
> 	defer policyContext.Destroy()
> 	                           ^
> imagebuildah/build.go:1602:15: Error return value of `cleanup` is not checked (errcheck)
> 	defer cleanup()
> 	             ^
> make: *** [Makefile:55: lint] Error 1
> ```

I would just log a warning with the error there to see if anything went wrong during the cleanup. :)

TomSweeneyRedHat:
@saschagrunert  I'm not sure where it went, but you're right the Makefile changes are not here.  Sometimes I love git....

.golangci.yml
```diff
@@ -0,0 +1,38 @@
...
+    # - gosec
+    # - govet
+    # - lll
+    # - maligned
+    # - megacheck
+    # - nakedret
+    # - prealloc
+    # - scopelint
+    # - unconvert
+    # - unparam
```
saschagrunert: The latest version of golangci-lint (1.17.1) contains a new checker called `bodyclose`. We should add them to this list too to be complete (either enabled or disabled).

vendor.conf
```diff
@@ -67,3 +67,9 @@ github.com/spf13/cobra v0.0.3
 github.com/spf13/pflag v1.0.3
 github.com/ishidawataru/sctp 07191f837fedd2f13d1ec7b5f885f0f3ec54b1cb
 github.com/etcd-io/bbolt v1.3.2
+
+# Added for conformance testing only
+github.com/hpcloud/tail v1.0.0
+github.com/onsi/ginkgo v1.7.0
```
saschagrunert: Is it intended that we these vendoring deps should be part of this PR?

Makefile
```diff
@@ -70,15 +74,14 @@ endif
 	@./tests/validate/whitespace.sh
 	@./tests/validate/govet.sh
 	@./tests/validate/git-validation.sh
-	@./tests/validate/gometalinter.sh . cmd/buildah
+	lint
```
saschagrunert: Since this is a make target I guess it won't work here. We could add it as dependency to `validate: lint`

TomSweeneyRedHat:
Closing in favor of #1678 and now I won't have a massive rebase!  Thanks @vrothberg !

Issue #1664, opened 2019-06-11
Potential hang when using run


TomSweeneyRedHat:
As noted in #1653 
A command like:

seq 100000 | sudo ./buildah run fedora-working-container -- sh -c 'while read i; do echo $i; done'

seems to hang

giuseppe:
PR opened here: https://github.com/containers/buildah/pull/1668

PR #1668, opened 2019-06-12
run: fix hang when using run

giuseppe:
make the stdin pipe not blocking, so that it won't hang if the other
end is not reading from it.

Closes: https://github.com/containers/buildah/issues/1664

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -1049,6 +1049,18 @@ func runConfigureNetwork(isolation Isolation, options RunOptions, configureNetwo
 	return teardown, nil
 }
 
+func setNonblock(fd int, description string, nonblocking bool) error {
+	err := unix.SetNonblock(fd, nonblocking)
+	if err != nil {
+		if nonblocking {
+			logrus.Errorf("error setting %s to nonblocking: %v", description, err)
+		} else {
+			logrus.Errorf("error setting descriptor %s blocking: %v", description, err)
+		}
+	}
+	return err
+}
+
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
@giuseppe 
Looks like you need to make a function somewhere to avoid gofmt cycle issues:
```
run_linux.go:1052::warning: cyclomatic complexity 46 of function runCopyStdio() is high (> 45) (gocyclo)
Makefile:66: recipe for target 'validate' failed
make: *** [validate] Error 1
```

rhatdan:
LGTM

TomSweeneyRedHat:
looks like a network flake killed it, will spin again.

run_linux.go
```diff
@@ -1049,6 +1049,18 @@ func runConfigureNetwork(isolation Isolation, options RunOptions, configureNetwo
 	return teardown, nil
 }
 
+func setNonblock(fd int, description string, blocking bool) error {
+	err := unix.SetNonblock(fd, true)
```
mheon: Is this right? Shouldn't this be `fd, blocking` instead of unconditional true?

giuseppe: argh, sure.  Surprised the tests passed

rhatdan:
@rh-atomic-bot r+

giuseppe:
@rh-atomic-bot retry

TomSweeneyRedHat:
I've rekicked the two failing travis tests. It looks like they're completing or at least mostly completing but running out of time?

giuseppe:
it seems the same issue is present in chroot, I am looking at it right now

rhatdan:
@rh-atomic-bot r+

Issue #1399, opened 2019-03-11
Ability remove config values


hhenkel:
**Description**

```buildah config``` allows to add various config options, but it seems like it is not possible to unset most of them with the command.

**Steps to reproduce the issue:**
1. buildah config --volume /test123 $containerid
2. buildah inspect $containerid
3. buildah config --volume [] $containerid

**Describe the results you received:**
Help text indicates "[]" is the default for volume, but I'm not able to set it back to that nor am I able to remove certain volumes with ```buildah config --volume```

**Describe the results you expected:**
Command line option to remove a certain setting

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.5-2.gite94b4f9.el7.x86_64
```

**Output of `buildah version`:**

```
buildah version 1.5 (image-spec 1.0.0, runtime-spec 1.0.0)
```

**Output of `cat /etc/*release`:**

```
Red Hat Enterprise Linux Server release 7.6 (Maipo)
```

**Output of `uname -a`:**

```
Linux dc00-sans001t.helvetia.ch 3.10.0-957.5.1.el7.x86_64 #1 SMP Wed Dec 19 10:46:58 EST 2018 x86_64 x86_64 x86_64 GNU/Linux
```

**Comment by rhatdan** [comment]
I would be fine with this, but we need to think what the syntax would be.

We could do something like:
```
--volume='[]' --label='[]' --env='[]' --anotations='[]' --onbuild=[] --port='[]'
Or just add the keywork none to all flags to remove any settings?
```
Which would either remove the container config settings or all settings including those in the `from` image?

nalind:
It's not the easiest thing to spot, but Kubernetes seems to have a convention that "setting" a value that ends with "-" actually unsets it.  A couple of the examples for [kubectl set env](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#-em-env-em-) and the last example in [kubectl label](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#label) might be relevant here.

TomSweeneyRedHat:
`kubernetes label` also seems to have an --overwrite option which clears the field and puts in the new value passed in.   I'm a little leery about the parsing work of having to check a number of options for a trailing '-'.

rhatdan:
So then 
--volume=/source-
Would remove all mount points /source?

TomSweeneyRedHat:
That's my read if you didn't use the `--overwrite` option.

[...]

PR #1670, opened 2019-06-12
Added '-' syntax to remove volume config option

ashley-cui:
When using the --volume flag, putting a '-' after the volume removes the volume from the config. 
`--volume /testvol-` removes /testvol from the config. 
Only implemented on Volume flag as a test. 

#1399

### DIFF
--- a/config.go
+++ b/config.go
@@ -440,8 +440,14 @@ func (b *Builder) AddVolume(v string) {
 // mounted from outside of the container when a container based on an image
 // built from this container is run.
 func (b *Builder) RemoveVolume(v string) {
-	delete(b.OCIv1.Config.Volumes, v)
-	delete(b.Docker.Config.Volumes, v)
+	_, OCIv1Volume := b.OCIv1.Config.Volumes[v]
+	_, DockerVolume := b.Docker.Config.Volumes[v]
+	if !OCIv1Volume && !DockerVolume {
+		logrus.Errorf("Volume \"%s\" cannot be removed from config because it was not set.", v)
+	} else {
+		delete(b.OCIv1.Config.Volumes, v)
+		delete(b.Docker.Config.Volumes, v)
+	}
 }
 
 // ClearVolumes removes all locations from the image's list of locations which
--- a/docs/buildah-config.md
+++ b/docs/buildah-config.md
@@ -180,7 +180,7 @@ If names are used, the container should include entries for those names in its
 
 **--volume** *volume*
 
-Add a location in the directory tree which should be marked as a *volume* in any images which will be built using the specified container. Can be used multiple times.
+Add a location in the directory tree which should be marked as a *volume* in any images which will be built using the specified container. Can be used multiple times. If *volume* has a trailing `-`, and does not exist on disk, then the volume is removed from the config.
 
 **--workingdir** *directory*
 
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -217,9 +217,17 @@ func updateConfig(builder *buildah.Builder, c *cobra.Command, iopts configResult
 	}
 	if c.Flag("volume").Changed {
 		if volSpec := iopts.volume; len(volSpec) > 0 {
-			for _, spec := range volSpec {
-				builder.AddVolume(spec)
-				conditionallyAddHistory(builder, c, "/bin/sh -c #(nop) VOLUME %s", spec)
+			for _, volVal := range volSpec {
+				_, err := os.Stat(volVal) //check if volume exists on disk
+				if err == nil {
+					builder.AddVolume(volVal)
+				} else if strings.HasSuffix(volVal, "-") {
+					rmVol := strings.TrimSuffix(volVal, "-")
+					builder.RemoveVolume(rmVol)
+				} else {
+					logrus.Errorf("Volume cannot be added because it does not exist on disk.")
+				}
+				conditionallyAddHistory(builder, c, "/bin/sh -c #(nop) VOLUME %s", volVal)
 			}
 		}
 	}
--- a/tests/config.bats
+++ b/tests/config.bats
@@ -107,6 +107,7 @@ function check_matrix() {
 }
 
 @test "config" {
+  mkdir ${TESTSDIR}/VOLUME
   cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
   buildah config \
    --author TESTAUTHOR \
@@ -120,7 +121,7 @@ function check_matrix() {
    --cmd COMMAND-OR-ARGS \
    --comment INFORMATIVE \
    --history-comment PROBABLY-EMPTY \
-   --volume /VOLUME \
+   --volume ${TESTSDIR}/VOLUME \
    --workingdir /tmp \
    --label LABEL=VALUE \
    --label exec='podman run -it --mount=type=bind,bind-propagation=Z,source=foo,destination=bar /script buz'\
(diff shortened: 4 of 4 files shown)


config.go
```diff
@@ -440,8 +440,18 @@ func (b *Builder) AddVolume(v string) {
 // mounted from outside of the container when a container based on an image
 // built from this container is run.
 func (b *Builder) RemoveVolume(v string) {
-	delete(b.OCIv1.Config.Volumes, v)
-	delete(b.Docker.Config.Volumes, v)
+	_, OCIv1Volume := b.OCIv1.Config.Volumes[v]
```
rhatdan: Why did you add this section?  Seems like we could get a few errors when volumes only exists in OCI instead of Docker, but I am not sure this is possible.

ashley-cui: I was unsure whether or not to add this part, but I thought it would be good to let the user know they were trying to remove a volume that was never set in the first place. It might be redundant to check both maps though, since they are always set and unset at the same time.

rhatdan:
Needs documentation changes. 
Needs tests.

config.go
rhatdan: How about you check both, and if neither removes, then throw an error.

rhatdan:
@ashley-cui Looks like the tests are failing, because it does not want you to use "spec" as a variable name.  Please fix this and then I will give you a LGTM

rhatdan:
Another validation error.
cmd/buildah/config.go:222::warning: declaration of "volVal" shadows declaration at config.go:220 (vetshadow)

rhatdan:
LGTM
@TomSweeneyRedHat @giuseppe @nalind @QiWang19 @vrothberg PTAL
@ashley-cui After we get this in, we have to look at other fields that need the same treatment.

rhatdan:
@vrothberg The problem with the --remove flag is it is not easy to tie it to any particular volume, and not easy to handle multiple volumes at the same time.

I believe the "-" syntax was decided on based on what Kubernetes was using, as discussed in the issue https://github.com/containers/buildah/issues/1399

vrothberg:
I understand it's more complex to implement that via flags but the current approach *can* break existing deployments.

The problem is that `-` is a perfectly fine path. The only invalid character I know of is `/` but that's unfortunately a common user-input when pointing to directories. I would prefer adding dedicated options to avoid such kind of "magic". In this case a `--rm-volume=volume`?

But if others are okay to merge it, I don't want to block (and don't want to be responsible for merging :-)

docs/buildah-config.md
```diff
@@ -180,7 +180,7 @@ If names are used, the container should include entries for those names in its
 
 **--volume** *volume*
 
-Add a location in the directory tree which should be marked as a *volume* in any images which will be built using the specified container. Can be used multiple times.
+Add a location in the directory tree which should be marked as a *volume* in any images which will be built using the specified container. Can be used multiple times. If volume has trailing `-`, then the volume is removed from the config
```
TomSweeneyRedHat: Please change to `If *volume* has a trailing` ...   and then add a period to the end.

rhatdan:
Well this is a limited case where we would only hit the issue if a volume ended in a '-'.
How about if we allow escaping the field.  
--volume /var/file\- 

Then check if the final character is - and previous char is '\', Of course this would then prevent me from mounting a volume that looked like '/var/file\-'.  

Another thing we could do is check if /var/file- exists, if it does then we mount it, if it does not we look for /var/file and remove it.  That would seem to solve this case best without making the CLI overly complicated.

vrothberg:
>  Another thing we could do is check if /var/file- exists, if it does then we mount it, if it does not we look for /var/file and remove it. 

That sounds better. 

> That would seem to solve this case best without making the CLI overly complicated.

I think it depends where we draw the line. Having special cases complicates documentation (sometimes usability) and both can impact the CLI.

rhatdan:
I think the other issue is that we want to be able to delete other fields within the image,  

--env foobar- or --label mylabel- 
...
We would end up having to add lots of --rmenv and --rmlabel options.  

Perhaps we can hold off on this until next week and then discuss it in a BlueJeans.

vrothberg:
Sounds good to me :+1:

TomSweeneyRedHat:
I'll be somewhere in or on a lake then, I'll trust @vrothberg to hold up my side of the argument!

rhatdan:
We also need this done for the following options.

--volume
--env
--label
--annotion
--device
--add-host
--dns
--dns-search
--expose
--group-add

rhatdan:
Ok after talking to @nalind we think the best way to handle the volume mount of /tmp/foo- is to check if the object exists on disk, if it does, then create a volume, if it does not, then check to see if /tmp/foo exists in the list of volumes and delete it if it does.  If it does not then return an error.

For the future options, then we just check if the option exists and remove it.

Corner cases we will not be able to handle, would be:

1 Adding an option ending with a "-" for anything other then volumes, will not be allowed.  IE You will not be able to add an environment variable with --env foo-, I guss if you add --env foo-="" it would be possible.

2 Remove /tmp/foo volume if /tmp/foo- exists.  Not much we can do about this, and enough of a corner case that I don't think we need to fix.

ashley-cui:
@rhatdan PTAL. Not sure how to write tests for this since it relies on checking if a file exists though.

rhatdan:
podman run -v /tmp/foo- fedora ls /tmp/foo
Should fail. With /tmp/foo is not defined in container

mkdir /tmp/foo-
podman run -v /tmp/foo- fedora ls /tmp/foo-
Should succeed

rhatdan:
@ashley-cui Any movement on this one?

ashley-cui:
@rhatdan Should be implemented, but not sure why the last test is failing. I'm re-running it now and seeing what happens.

ashley-cui:
@rhatdan looks like the test just needed to be re-run. :) This should be all set.

docs/buildah-config.md
TomSweeneyRedHat: Did this comment get missed?

ashley-cui: Oops, sorry! Just fixed it :)

ashley-cui:
> We also need this done for the following options.
> 
> --volume
> --env
> --label
> --annotion
> --device
> --add-host
> --dns
> --dns-search
> --expose
> --group-add

Some of these already have a removal option. Right now,  using `--env key=val` once would set it, and doing `--env key=val` again would remove it. This should be changed to the minus convention that we have now, correct? Meaning that if you do `--env key=val` twice, nothing would happen one the 2nd time, but doing `--env key=val-` after setting the env variable would remove it?

docs/buildah-config.md
```diff
@@ -180,7 +180,7 @@ If names are used, the container should include entries for those names in its
 
 **--volume** *volume*
 
-Add a location in the directory tree which should be marked as a *volume* in any images which will be built using the specified container. Can be used multiple times.
+Add a location in the directory tree which should be marked as a *volume* in any images which will be built using the specified container. Can be used multiple times. If *volume* has trailing `-`, and does not exist on disk, then the volume is removed from the config.
```
TomSweeneyRedHat: Ugh, hope I didn't cause heartburn with a bad comment.  Can you s/has trailing/has a trailing/ please?

ashley-cui: Done, I should be more careful when reading

TomSweeneyRedHat:
LGTM
@vrothberg are you good with the changes that you requested?

rhatdan:
@rh-atomic-bot r+
Good work @ashley-cui 
Now lets get the other options to handle the "-"
Also need to get this functionality into Podman.

vrothberg:
> LGTM
> @vrothberg are you good with the changes that you requested?

Thanks for the ping. I am not a big fan of the new semantics but the changes LGTM. Great work, @ashley-cui.

computator:
this seems to have broken things for me:
```sh
$ ctr=$(buildah from ubuntu)
$ buildah run $ctr mkdir /test
$ buildah config --volume /test $ctr
ERRO[0000] Volume cannot be added because it does not exist on disk.
```

rhatdan:
@ashley-cui PTAL

ashley-cui:
@computator does the directory `/test` exist on your host?

computator:
No, but volumes are directories in containers, not on the host, so should not be affected by the host.

ashley-cui:
Hmm okay I may have misunderstood what a volume is.

ashley-cui:
Okay, under my understanding, a volume usually is a bind mount into a container, which means that the directory does also have to exist on a container? Please correct me if I am wrong. @computator @rhatdan

ashley-cui:
Okay, under my understanding, a volume usually is a bind mount into a container, which means that the directory does also have to exist on a container? Please correct me if I am wrong. @computator @rhatdan

rhatdan:
@ashley-cui Lets talk today, I am in Boston Office.

PR #1673, opened 2019-06-14
add --mount for buidah run

### LINKED ISSUE #1597: Support mounting directories with a `:` (colon) in their path
**Description**

Currently [there is no way to mount a directory with a colon in its name (e.g. `/tmp/aa:bb`) neither as source nor destination.
[This is the code that parses the mounts](https://github.com/containers/buildah/blob/59da11d4b4df6c9b51d06fc804f7a2e46be6c700/pkg/parse/parse.go#L154). 

Try to mount a directory with a colon in its name.
```bash
buildah run -v /some:dir:/destination container-name
```

**Suggested solution**
Support the same syntax docker supports, with `--mount` in addition to `-v/--volume`.

QiWang19:
close #1597

Enables to use path with colon when using `buildah run --volume` to bind a volume. Have the same functionality as --volume.

### DIFF
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -30,6 +30,17 @@ const (
 	SeccompDefaultPath = "/usr/share/containers/seccomp.json"
 	// SeccompOverridePath if this exists it overrides the default seccomp path
 	SeccompOverridePath = "/etc/crio/seccomp.json"
+	// TypeBind is the type for mounting host dir
+	TypeBind = "bind"
+	// TypeTmpfs is the type for mounting tmpfs
+	TypeTmpfs = "tmpfs"
+)
+
+var (
+	errBadMntOption  = errors.Errorf("invalid mount option")
+	errDuplicateDest = errors.Errorf("duplicate mount destination")
+	optionArgError   = errors.Errorf("must provide an argument for option")
+	noDestError      = errors.Errorf("must set volume destination")
 )
 
 // CommonBuildOptions parses the build options from the bud cli
@@ -168,13 +179,14 @@ func parseSecurityOpts(securityOpts []string, commonOpts *buildah.CommonBuildOpt
 	return nil
 }
 
+// ParseVolume parses the input of --volume
 func ParseVolume(volume string) (specs.Mount, error) {
 	mount := specs.Mount{}
 	arr := strings.SplitN(volume, ":", 3)
 	if len(arr) < 2 {
 		return mount, errors.Errorf("incorrect volume format %q, should be host-dir:ctr-dir[:option]", volume)
 	}
-	if err := ValidateVolumeHostDir(arr[0]); err != nil {
+	if err := validateVolumeMountHostDir(arr[0]); err != nil {
 		return mount, err
 	}
 	if err := ValidateVolumeCtrDir(arr[1]); err != nil {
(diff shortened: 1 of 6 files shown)


rhatdan:
LGTM
If you copied these functions out of libpod, can we remove them from libpod and vendor these in?

rhatdan:
@QiWang19 Tests are not happy.

rhatdan:
@QiWang19 needs a rebase.

QiWang19:
This doesn't support  type=tmpfs, where should I add the implementation for that?

rhatdan:
@QiWang19 I guess in a separate PR.

QiWang19:
@rhatdan I have not idea where to add the --mount implememtation into the code or how --mount should be implemented :neutral_face:, checking the code in libpod...

nalind:
I think we'll to adjust the logic to avoid adding the `rbind` option to `options` when the `mountType` doesn't end up set to "bind", in the `parseMount()` helper function in `Builder.runSetupVolumeMounts()` in `run_linux.go`.

nalind:
There's a typo in the commit log.

QiWang19:
tmpfs can work without adding rbind to the options. Is this because of the system call. Are these the options will be sent to the system call?

nalind:
Some of the options are parsed are used to set or clear bits in the `mountflags` value that's passed to mount(2), and the rest are passed directly as text.  The `bind` and `rbind` options become flags which cause the filesystem type to be ignored, so they can't be specified if we want to mount a tmpfs.

QiWang19:
@nalind @rhatdan PTAL

pkg/parse/parse.go
```diff
@@ -195,8 +209,8 @@ func ParseVolume(volume string) (specs.Mount, error) {
 	return mount, nil
 }
 
-// ParseVolumes validates the host and container paths passed in to the --volume flag
-func ParseVolumes(volumes []string) error {
+// parseVolumes validates the host and container paths passed in to the --volume flag
+func parseVolumes(volumes []string) error {
```
rhatdan: Why are you making these private?  Doesn't Podman use these?

pkg/parse/parse.go
```diff
@@ -218,14 +494,19 @@ func ValidateVolumeHostDir(hostDir string) error {
...
 	if !filepath.IsAbs(ctrDir) {
-		return errors.Errorf("invalid container path, must be an absolute path %q", ctrDir)
+		return errors.Errorf("invalid container path %q, must be an absolute path", ctrDir)
 	}
 	return nil
 }
 
-func ValidateVolumeOpts(option string) error {
+// validates options in --volume flag
+func validateVolumeMountOpts(option string) error {
```
rhatdan: Why are you making these private? Doesn't Podman use these?

pkg/parse/parse.go
QiWang19: No. this is not from libpod. it is only used in buildah parse.go

rhatdan: I am saying the Podman is vendoring this code into its code base and uses this function to parse its objects.

pkg/parse/parse.go
QiWang19: there is another public function `ValidateVolumeOpts` for podman. Make `validateVolumeMountOpts` private for buildah volume. Because podman volume doesn't have `:o` option but buildah has

rhatdan: I plan on adding the :o option for Podman, so I want this public as well.

pkg/parse/parse.go
QiWang19: No, I don't think podman needs this function.

pkg/parse/parse.go
QiWang19: sure

rhatdan:
@QiWang19 https://github.com/containers/libpod/pull/3458 is using these methods.

nalind:
LGTM

pkg/parse/parse.go
rhatdan: Leave this exported for now, once we have patches podman to support Overlay Mounts, if we don't use these interfaces, then you can unexport them. Otherwise I will have to open a PR to reverse this patch to make them exportable again.

pkg/parse/parse.go
```diff
@@ -208,7 +220,227 @@ func ParseVolumes(volumes []string) error {
...
+	return finalVolumeMounts, nil
+}
+
+// GetVolumes gets the volumes from --volume and --mount
+func GetVolumes(volumes []string, mounts []string) ([]specs.Mount, error) {
+	unifiedMounts, err := getMounts(mounts)
+	if err != nil {
+		return nil, err
+	}
+	voluemMounts, err := getVolumeMounts(volumes)
```
TomSweeneyRedHat: s/voluemMounts/volumeMounts   (e and m in voluem)

TomSweeneyRedHat:
LGTM assuming happy tests.  I'll give @rhatdan a chance to look at it tomorrow, if he doesn't merge it, I will over tea.

rhatdan:
@rh-atomic-bot r+

PR #1676, opened 2019-06-15
Bump 1.9.0

rhatdan:


buildah.go
```diff
@@ -26,7 +26,7 @@ const (
 	Package = "buildah"
 	// Version for the Package.  Bump version in contrib/rpm/buildah.spec
 	// too.
-	Version = "1.9.0"
+	Version = "1.9.1-dev"
```
TomSweeneyRedHat: IDK if this is right or not, but I've been bumping this to the next minor version, so 1.10.0-dev for here.  Six of one, half dozen of another.

TomSweeneyRedHat:
LGTM,
We'll need to touch up CHANGELOG.md too, I always forget that one.

PR #1698, opened 2019-06-28
import method ValidateVolumeOpts from libpod

QiWang19:
move ValidateVolumeOpts from libpod to buildah so libpod can vendor it in.


rhatdan:
LGTM, although we should have a couple of tests for this to make sure we support the additional options.

rhatdan:
@mheon @TomSweeneyRedHat PTAL

mheon:
Line 266 should probably not be a strings. Join() on all the options.

QiWang19:
fix L266

rhatdan:
@rh-atomic-bot r+

PR #1705, opened 2019-07-03
On Masked path, check if /dev/null already mounted before mounting

rhatdan:
We do not want to mount /dev/null over a masked path, if the path is
already /dev/null.

This prevents an containers running buildah from requiring additional privs
to mount on a /dev/null, when the target is already mounted.

### DIFF
--- a/chroot/run.go
+++ b/chroot/run.go
@@ -983,6 +983,18 @@ func makeReadOnly(mntpoint string, flags uintptr) error {
 	return nil
 }
 
+func isDevNull(dev os.FileInfo) bool {
+	if dev.Mode()&os.ModeCharDevice != 0 {
+		stat, _ := dev.Sys().(*syscall.Stat_t)
+		nullStat := syscall.Stat_t{}
+		syscall.Stat(os.DevNull, &nullStat)
+		if stat.Rdev == nullStat.Rdev {
+			return true
+		}
+	}
+	return false
+}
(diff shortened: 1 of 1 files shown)


rhatdan:
@nalind @giuseppe @vrothberg @TomSweeneyRedHat PTAL

This fixes the issue we are seeing with people trying to run buildah within a locked down container with chroot.  SELinux is blocking the mounting of the second /dev/null on /proc/kcore.  /proc/kcore already has a /dev/null mounted on it.

chroot/run.go
```diff
@@ -1355,8 +1362,8 @@ func setupChrootBindMounts(spec *specs.Spec, bundlePath string) (undoBinds func(
 				}
 			}
 		} else {
-			// The target's not a directory, so bind mount os.DevNull over it, unless it's already os.DevNull.
-			if !os.SameFile(nullinfo, targetinfo) {
+			// If the target's is not a directory or os.DevNull, bind mount os.DevNull over it.
```
TomSweeneyRedHat: nit of a nit, 
s/target's/target/
But only if you've other changes.

TomSweeneyRedHat:
Changes LGTM, test issues don't look to be related.

rhatdan:
@vrothberg Not sure how I could write a test to test this.  I have tested this by hand though.

PR #1706, opened 2019-07-04
Switch to golangci-lint

saschagrunert:
This commit vendors golangci-lint and enables all linters which do not
complain right now. CI and scripting has been adapted as well.


vrothberg:
Please use the module in `./tests/tools` for vendoring. This will avoid mixing tests and build dependencies in one module.

go.mod
```diff
@@ -22,12 +21,10 @@ require (
 	github.com/fsouza/go-dockerclient v1.3.0
 	github.com/ghodss/yaml v1.0.0
 	github.com/gogo/protobuf v1.2.0 // indirect
+	github.com/golangci/golangci-lint v1.17.1
```
vrothberg: This should be moved over ot https://github.com/containers/buildah/tree/master/tests/tools .

saschagrunert:
> Please use the module in `./tests/tools` for vendoring. This will avoid mixing tests and build dependencies in one module.

Thanks for pointing me there, I was searching it but could not find it for whatever reasons. :innocent:

vrothberg:
No worries :) I am happy that you're pulling the trigger on this migration. I opened an [issue for libpod](https://github.com/containers/libpod/issues/3498) today to track the migration there as well - still quite some work needed to let it pass though.

go.mod
saschagrunert: Done :+1:

saschagrunert:
We have to install ostree on travis to make the linter work there. I will check if we can do that easily... :)

saschagrunert:
Seems to work :+1:

rhatdan:
LGTM
@rh-atomic-bot r+

PR #1710, opened 2019-07-08
Improve tests/tools/Makefile parallelism and abstraction

saschagrunert:
Update the Makefile to be less repeatable in terms of targets to be
built. With that change, a parallel build of all targets is possible as
well, which has been changed in the CI too. A `clean` and `all` target
has to be added for convenience.


rhatdan:
LGTM
@vrothberg @umohnani8 @haircommander @mrunalp PTAL

vrothberg:
@rh-atomic-bot r+

Issue #1714, opened 2019-07-10
COPY command for ~75M takes several minutes


carlpett:
**Description**
Building a container with a build context of ~75M and around 3700 files takes several minutes to do a `COPY . /target-dir`. Colleagues using `docker` have a build time of seconds.

Checking top or iotop during the build shows very low utilization.

**Steps to reproduce the issue:**
1. `time podman build .`

**Describe the results you received:**
Slow build

**Describe the results you expected:**
Fast build

**Output of `rpm -q buildah` or `apt list buildah`:**
(Giving podman instead, since I don't have buildah installed itself)
```
$ rpm -q podman
podman-1.4.4-1.fc30.x86_64
```

**Output of `podman version` if reporting a `podman build` issue:**

```
$ podman version
Version:            1.4.5-dev
RemoteAPI Version:  1
Go Version:         go1.12.6
OS/Arch:            linux/amd64
```

First tested with 1.4.2, updated to see if that helped.
*Also note:* This is pretty weird packaging? The rpm says 1.4.4, I get something tagged 1.4.5-dev?

**Output of `cat /etc/*release`:**

```
Fedora release 30 (Thirty)
NAME=Fedora
VERSION="30 (Thirty)"
ID=fedora
VERSION_ID=30
VERSION_CODENAME=""
PLATFORM_ID="platform:f30"
PRETTY_NAME="Fedora 30 (Thirty)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:30"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f30/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=30
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=30
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
Fedora release 30 (Thirty)
Fedora release 30 (Thirty)
```

**Output of `uname -a`:**

```
[...]

PR #1715, opened 2019-07-12
add: fix slow copy with no excludes

giuseppe:
when there are excludes defined (such as from .dockerignore), we take
a slow path and walk each file in the directory.  If the files doesn't
match any exclusion pattern then it is copied into the container.
This is slow as each file requires buildah to re-exec and copy it from
a chroot environment.

When there are no excludes defined we can take a faster path and copy
the entire directory as a single re-exec operation.

Closes: https://github.com/containers/buildah/issues/1714

### DIFF
--- a/add.go
+++ b/add.go
@@ -278,7 +278,7 @@ func addHelper(excludes *fileutils.PatternMatcher, extract bool, dest string, de
 					return errors.Wrapf(err, "error creating directory %q", dest)
 				}
 				logrus.Debugf("copying %q to %q", esrc+string(os.PathSeparator)+"*", dest+string(os.PathSeparator)+"*")
-				if excludes == nil {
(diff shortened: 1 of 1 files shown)


rhatdan:
@rh-atomic-bot r+

michaelgugino:
@giuseppe thanks for the quick fix!

I confirm on Fedora 30 this fixes the issue for me.

Built this patch with 'make binary' and it performs as expected.  Using current tip of master (ie, without this patch), it's slow.

TomSweeneyRedHat:
Looks like the dockerignore test isn't happy with the change.

@giuseppe PTAL
```
not ok 24 bud with .dockerignore
# (from function `die' in file ./helpers.bash, line 128,
#  from function `run_buildah' in file ./helpers.bash, line 116,
#  in test file ./bud.bats, line 27)
```

PR #1716, opened 2019-07-12
Bump to Version v1.9.1

rhatdan:


nalind:
The .spec file is missing a changelog entry, otherwise LGTM.

PR #1725, opened 2019-07-16
unshare: add a --mount flag

nalind:
Add a `--mount` flag to `buildah unshare`, so that the command it runs can have a container's root filesystem mounted, and the location where it's mounted provided in its environment.

This is primarily aimed at cases where `buildah unshare` is used to run one-liner commands, as proper scripts can simply call `buildah mount` themselves.

Users still have to be careful about quoting references to the environment if the command needs to be executed by a shell (for example, if it's run as `buildah unshare --mount foo sh -c 'ls $foo'`), so that the shell that's invoking `buildah unshare` doesn't try to expand the variable's value first.

My use case was primarily mucking around in https://github.com/containers/BuildSourceImage, and if it's useful for me there, hopefully it will be for others.

### DIFF
--- a/cmd/buildah/unshare.go
+++ b/cmd/buildah/unshare.go
@@ -3,10 +3,14 @@
 package main
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
+	"strings"
 
 	"github.com/containers/buildah/pkg/unshare"
+	"github.com/containers/storage"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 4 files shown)


cmd/buildah/unshare.go
```diff
@@ -22,15 +26,70 @@ var (
...
+		mount := strings.SplitN(mountSpec, "=", 2)
+		container := mountSpec
+		envVar := container
+		if len(mount) == 2 {
+			envVar = mount[0]
+			container = mount[1]
+		}
+		builder, err := openBuilder(getContext(), store, container)
+		if err != nil {
+			defer unmount()
```
rhatdan: Why defer?  If you are going defer, should we just do a 
```
defer func() {
         if Err != nil {
             unmount()
         }
}
```

rhatdan:
I am fine with this change.

cmd/buildah/unshare.go
nalind: That's a good point.  Changing it.

rhatdan:
bot, retest this please
LGTM
@TomSweeneyRedHat @giuseppe @vrothberg @ashley-cui PTAL

docs/buildah-unshare.md
```diff
@@ -19,6 +19,14 @@ manually clearing storage and other data related to images and containers.
 It is also useful if you want to use the `buildah mount` command.  If an unprivileged users wants to mount and work with a container, then they need to execute
 buildah unshare.  Executing `buildah mount` fails for unprivileged users unless the user is running inside a `buildah unshare` session.
 
+## OPTIONS
+**--mount** [*VARIABLE=]containerNameOrID*
+
+Mount the *containerNameOrID* container while running *command*, and set the
```
TomSweeneyRedHat: Suggest:  `Mount the container specified by the *containerNameOrID* while running ....`

TomSweeneyRedHat: actually drop the 'the'
`Mount the container specified by *containerNameOrID* while running ....`

TomSweeneyRedHat:
One nit for consideration and "Red Hat CI" has gone off the rails again, most likely because it wants to and not due to the changes.
Otherwise LGTM

rhatdan:
bot, retest this please

giuseppe:
LGTM

rhatdan:
@rh-atomic-bot r+

nalind:
@rh-atomic-bot retry

rhatdan:
Merge bot seems to be hung going to force merge.

PR #1732, opened 2019-07-19
Handle ReadOnly Images

rhatdan:


rhatdan:
@vrothberg Yes, I want to talk to @baude about setting up buildah to use additional stores like podman does, then this would be easier to run tests on.

TomSweeneyRedHat:
Looks like an images test needs some tweaking.  Code changes LGTM otherwise.

```
not ok 227 images json test
# (from function `expect_line_count' in file ./helpers.bash, line 228,
#  in test file ./images.bats, line 93)
#   `expect_line_count 14' failed
```

TomSweeneyRedHat:
LGTM, assuming happy tests.  Looks like Red Hat CI has fallen off the rails again.  I'm sure unrelated.

rhatdan:
@cevich How can we move to the /lgtm way of merging?

rhatdan:
@rh-atomic-bot r=@TomSweeneyRedHat

cevich:
> @cevich How can we move to the /lgtm way of merging?

I'm having trouble imagining a way that isn't heavily time-consuming in some respect.

*  Adding this repo. to the OpenShift Prow system is one option (what libpod does), it kind of works, but has some complex issues.
* The traditional way of doing this is to have a small VM somewhere, running a bot that periodically monitors PRs.  I know how to do this, it's not hard, just needs time.
* A possibly easier way, could be to hook into the new Github-Actions system (in beta, but mostly stable).  I have no idea how to do this, would need to spend time researching.

The last two options could be applied to other repositories, so only the initial cost is high.

Issue #1741, opened 2019-07-26
rootless overlay mount option 'O'


jamescassell:
**Description**

The `O` mount option is not available in rootless mode.  This option can be used to greatly speed up building containers by pre-loading `yum` or other caches.  When attempting to use the option in rootless mode, an error is received.

**Steps to reproduce the issue:**
1. c=$(buildah from ubi8)
2. buildah run $c yum makecache
3. m=$(buildah mount $c)
4. d=$(buildah from -v $m/var/cache/dnf:/var/cache/dnf:O ubi8)

**Describe the results you received:**
`invalid options "O", overlay mounts not supported in rootless mode`

**Describe the results you expected:**
container created as normal without error

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.9.2-2.fc29.x86_64
```

**Output of `buildah version`:**

```
Version:         1.9.2
Go Version:      go1.11.12
Image Spec:      1.0.1
Runtime Spec:    1.0.1-dev
CNI Spec:        0.4.0
libcni Version:  
Git Commit:      
Built:           Wed Dec 31 19:00:00 1969
OS/Arch:         linux/amd64
```

**Output of `uname -a`:**

```
Linux localhost.localdomain 5.1.16-202.fc29.x86_64 #1 SMP Fri Jul 5 13:34:31 EDT 2019 x86_64 x86_64 x86_64 GNU/Linux
```

**Output of `cat /etc/containers/storage.conf`:**

```
$ cat ~/.config/containers/storage.conf 
[storage]
  driver = "overlay"
  runroot = "/tmp/1000"
  graphroot = "/home/user/.local/share/containers/storage"
  [storage.options]
    size = ""
    remap-uids = ""
    remap-gids = ""
    remap-user = ""
    remap-group = ""
    ostree_repo = ""
    skip_mount_home = ""
    mount_program = "/usr/bin/fuse-overlayfs"
    mountopt = ""
    [storage.options.thinpool]
      autoextend_percent = ""
      autoextend_threshold = ""
      basesize = ""
      blocksize = ""
      directlvm_device = ""
      directlvm_device_force = ""
      fs = ""
      log_level = ""
      min_free_space = ""
      mkfsarg = ""
      mountopt = ""
      use_deferred_deletion = ""
      use_deferred_removal = ""
      xfs_nospace_max_retries = ""
```

rhatdan:
The problem is you are not allowed to mount overlay mounts when not running as root.
[...]

PR #1743, opened 2019-07-28
rootless, overlay: use fuse-overlayfs

giuseppe:
when running in rootless mode, use fuse-overlayfs for mounting the
overlay file system on the host.  Then create a bind mount inside the
container.

Closes: https://github.com/containers/buildah/issues/1741

### DIFF
--- a/pkg/overlay/overlay.go
+++ b/pkg/overlay/overlay.go
@@ -4,17 +4,20 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 
+	"github.com/containers/buildah/pkg/unshare"
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
 )
 
 // MountTemp creates a subdir of the contentDir based on the source directory
-// from the source system.  It then mounds up the source directory on to the
+// from the source system.  It then mounts up the source directory on to the
 // generated mount point and returns the mount point to the caller.
 func MountTemp(store storage.Store, containerId, source, dest string, rootUID, rootGID int) (mount specs.Mount, contentDir string, Err error) {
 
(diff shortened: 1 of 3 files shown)


giuseppe:
marked as WIP, as it should honor the `mount_program` option in the storage configuration file

giuseppe:
> marked as WIP, as it should honor the `mount_program` option in the storage configuration file

fixed now

pkg/overlay/overlay.go
```diff
@@ -46,24 +49,82 @@ func MountTemp(store storage.Store, containerId, source, dest string, rootUID, r
...
 // CleanupContent removes all temporary mountpoint and all content from
 // directory
 func CleanupContent(containerDir string) (Err error) {
 	contentDir := filepath.Join(containerDir, "overlay")
+
+	if unshare.IsRootless() {
+		mergeDir := filepath.Join(contentDir, "merge")
+		if err := unix.Unmount(mergeDir, 0); err != nil {
+			if !os.IsNotExist(err) {
+				return errors.Wrapf(err, "unmount overlay")
```
rhatdan: Add Path to error.

pkg/overlay/overlay.go
```diff
@@ -46,24 +49,82 @@ func MountTemp(store storage.Store, containerId, source, dest string, rootUID, r
...
+			s := strings.SplitN(i, "=", 2)
+			if len(s) != 2 {
+				continue
+			}
+			k := s[0]
+			v := s[1]
+			switch k {
+			case ".mount_program", "overlay.mount_program", "overlay2.mount_program":
+				mountProgram = v
+				break loop
```
rhatdan: Why do you need loop: and break loop?  Wouldn't break just work?

pkg/overlay/overlay.go
```diff
@@ -46,24 +49,82 @@ func MountTemp(store storage.Store, containerId, source, dest string, rootUID, r
...
+			mergeDir := filepath.Join(contentDir, "merge")
+
+			if err := idtools.MkdirAllAs(mergeDir, 0700, rootUID, rootGID); err != nil {
+				return mount, "", errors.Wrapf(err, "failed to create the overlay %s directory", mergeDir)
+			}
+
+			cmd := exec.Command(mountProgram, "-o", overlayOptions, mergeDir)
+
+			if err := cmd.Run(); err != nil {
+				return mount, "", errors.Wrapf(err, "exec fuse-overlayfs")
```
rhatdan: Should be mountProgram not fuse-overlayfs.

rhatdan:
Do we have tests for this that will run in rootless mode?

pkg/overlay/overlay.go
giuseppe: writing too much C recently.  I assumed it would be part of the switch.

pkg/overlay/overlay.go
```diff
@@ -46,24 +49,81 @@ func MountTemp(store storage.Store, containerId, source, dest string, rootUID, r
...
+		mountProgram := ""
+
+		for _, i := range store.GraphOptions() {
+			s := strings.SplitN(i, "=", 2)
+			if len(s) != 2 {
+				continue
+			}
+			k := s[0]
+			v := s[1]
+			switch k {
```
rhatdan: How about you setup a map
mountMap := map[string]bool {
               ".mount_program": true,
               "overlay.mount_program": true,
               "overlay2.mount_program":true, 
}
...
if mountMap[k] {
           mountProgram=v
           break
}

giuseppe: amended in the last version

rhatdan:
LGTM
@vrothberg @baude @mheon @TomSweeneyRedHat @QiWang19 PTAL

TomSweeneyRedHat:
LGTM

rhatdan:
@rh-atomic-bot r+

Issue #1407, opened 2019-03-13
Docker references with both a tag and digest are currently not supported


devurandom:
**Description**

Buildah 1.7.1 does not support referencing an image (or pulling it from a registry) with both a tag and a digest. I could not find a feature request about this, so I create this one.

**Steps to reproduce the issue:**

```
$ buildah from docker://docker.io/library/clojure:openjdk-8-lein-2.9.1-alpine@sha256:d5454be246358c9cd683a06108ccdecc8b31a0341e45973af480bef708e8cf1e
while pulling "docker://docker.io/library/clojure:openjdk-8-lein-2.9.1-alpine@sha256:d5454be246358c9cd683a06108ccdecc8b31a0341e45973af480bef708e8cf1e" as "docker.io/library/clojure:openjdk-8-lein-2.9.1-alpine@sha256:d5454be246358c9cd683a06108ccdecc8b31a0341e45973af480bef708e8cf1e": error parsing attempted image name "docker://docker.io/library/clojure:openjdk-8-lein-2.9.1-alpine@sha256:d5454be246358c9cd683a06108ccdecc8b31a0341e45973af480bef708e8cf1e": Docker references with both a tag and digest are currently not supported
```

TomSweeneyRedHat:
@devurandom Thanks for the issue.

@mtrmac is this anything you forsee changing or should we look into propping up the documentation somehow?

mtrmac:
We’ve seen this brought up in a one other place WRT `podman pull`, and Docker does accept that syntax; so, yes, perhaps we would accept this syntax — but _only_ in the compatibility / convenience syntax that does not start with a transport specification (`buildah from docker.io/…`), as handled in Podman’s `pullGoalFromPossiblyUnqualifiedName`.

(We’ve been tentatively discussing sharing the pull code between buildah/libpod; this should probably happen in there.)

I don’t expect this syntax to be supported for the `docker://` transport directly, primarily because the underlying protocol does not allow specifying both at the same time, so its semantics is ambiguous about users’ intent, and the historical practice is problematic:
- Earlier versions of Docker, including 1.13.1, interpret this as “ask for the image with the specified tag, then fail if the manifest does not match the specified digest”. That’s fairly brittle but ultimately plausible (where the tag ~documents the intent and the digest ensure immutability and prevents moving the tag maliciously).
[...]

PR #1744, opened 2019-07-28
delete `successfully pushed` message 

### LINKED ISSUE #1742: pushing to a registry has `//` prefix
`StringWithinTransport()` from docker is prefixed with `//`. It must have been like this for a long time but I just realized it now. I would love to not print the prefix as Podman does.

```
Successfully pushed //localhost:5000/test:image@sha256:34384895c5fee4805650509e460891bec04571290f73509c08a8fa47aaae4219 
```

QiWang19:
close #1742
Delete the line: `Successfully pushed...`

### DIFF
--- a/cmd/buildah/push.go
+++ b/cmd/buildah/push.go
@@ -175,8 +175,6 @@ func pushCmd(c *cobra.Command, args []string, iopts pushResults) error {
 	} else {
 		logrus.Debugf("pushed image with digest %s", digest.String())
 	}
-	fmt.Printf("Successfully pushed %s@%s\n", dest.StringWithinTransport(), digest.String())
-
 	return nil
 }
 


QiWang19:
I remove the `Printf(...)` for now to keep it same as `Podman`

TomSweeneyRedHat:
Change is fine, but the commit title and description need to be touched up to reflect the final changes.

mtrmac:
> 1. We support parsing `image:tag@digest`. @mtrmac should weigh in here to be sure I am not missing something.

See earlier discussion in https://github.com/containers/buildah/issues/1407 and linked issues.

QiWang19:

TomSweeneyRedHat:
LGTM assuming happy tests

TomSweeneyRedHat:
@rh-atomic-bot r+

wwilson:
Following this change, what's the recommended way to atomically determine the remote digest of a newly-pushed image? We had come to rely on this log message for content addressability, in order to work around the issue that RepoDigest and local image digests do not in general match (https://github.com/containers/skopeo/issues/469).

An alternate workaround would be to `skopeo inspect` the remote image after `push` completes, but that's obviously asking for concurrency problems if somebody else uploads an image with the same tag at the same time.

Is there another way that Buildah reports the RemoteDigest of the image it has just pushed? Or should we be going about this a completely different way?

rhatdan:
@vrothberg @mtrmac @nalind PTAL

TomSweeneyRedHat:
If you run 'buildah --debug push ...` you get some of that information in the output:

```
DEBU[0001] PUT https://registry-1.docker.io/v2/tomsweeneyredhat/testing/manifests/first 
Storing signatures
DEBU[0002] pushed image "docker.io/tomsweeneyredhat/testing:first@sha256:b552b401d9977f4da64e410727e84a736be5b78f96b561a9b21579b3d1d74a8f" with digest sha256:b552b401d9977f4da64e410727e84a736be5b78f96b561a9b21579b3d1d74a8f 
```

I'm not sure if that's sufficient for your needs nor do I know of another way to easily do it other than what you've mentioned.  

I've just "re-added" the output as a debug line in #1799.  If the other folks don't have a better thought, we could go with that.  The output using the `--debug` as shown above would look like:

```
DEBU[0001] PUT https://registry-1.docker.io/v2/tomsweeneyredhat/testing/manifests/first 
Storing signatures
DEBU[0002] pushed image "docker.io/tomsweeneyredhat/testing:first@sha256:b552b401d9977f4da64e410727e84a736be5b78f96b561a9b21579b3d1d74a8f" with digest sha256:b552b401d9977f4da64e410727e84a736be5b78f96b561a9b21579b3d1d74a8f 
DEBU[0002] Successfully pushed //tomsweeneyredhat/testing:first@sha256:b552b401d9977f4da64e410727e84a736be5b78f96b561a9b21579b3d1d74a8f 
```

There's a lot more output this way though, would this work for you?

wwilson:
Thanks, I didn't know about the `buildah --debug push` output. I think we could skate by with taking that and feeding it into sed/awk, but I'm a little nervous parsing debug output like that since it seems like the kind of thing that can change without warning (as indeed it did in this case).

If you agree that this is a reasonable thing for people to want to know after pushing an image, should we also think about a more permanent and documented solution?

TomSweeneyRedHat:
I'd like to hear from @nalind and @rhatdan on their thoughts as they had wanted to remove this line.  Another approach that we might consider is adding a `--verbose` option which would print that line as it had previously done if `--verbose` or `-v` was passed in ala `buildah push -v ...`

FWIW, the `--debug` option can be place after buildand and before any buildah command to get a lot of insight into what's going on under the covers with a particular command.

mtrmac:
IIRC removing this text was not an explicit goal, it was just easier than to agree on a format that makes sense when we didn’t know of an user and didn’t have a strong need. (Let’s worry about that part in #1799.)

Originally the functionality was added via https://github.com/containers/image/issues/518 for OpenShift use (I guess via the API) and to the CLI output via https://github.com/containers/buildah/pull/1097 , which does not _explicitly_ say why it was added to the CLI).

---

I mean, if the digest is supposed to be _reliably_ consumable by scripts, we don’t want the scripts scraping the CLI output if we can help it; the CLI could have a `buildah push --digestfile` (similar to `buildah commit --iidfile`), or there could be some other structured output format.

`--verbose`/`--debug` are easier to implement but ultimately less reliable.

rhatdan:
Ok, if users want this information, then --digestfile would work. @wwilson WDYT?

wwilson:
I agree that structured output would be best. `—digestfile` sounds great.

Issue #1745, opened 2019-07-28
overlay not working on ubuntu: Error processing tar file(exit status 1): operation not permitted


ianmiell:
**Description**

When trying to use overlay, I get a permission error regarding permissions.

It defaults to vfs, which works (but uses far too much disk space).

I assume it's some kind of mapping error, but I haven't found a way to resolve.

**Steps to reproduce the issue:**
1. `rm -rf ~/.local/shared/containers`
2. `podman build --storage-driver overlay .`

**Describe the results you received:**

```
imiell@basquiat:~/tmp  🐧  podman build --storage-driver overlay .
STEP 1: FROM docker.io/ubuntu:xenial
Getting image source signatures
Copying blob edf72af6d627 done
Copying blob 3e4f86211d23 done
Copying blob 8d3eac894db4 done
Copying blob f7277927d38a done
Copying config 5e13f8dd4c done
Writing manifest to image destination
Storing signatures
Error: error creating build container: Error committing the finished image: error adding layer with blob "sha256:8d3eac894db4dc4154377ad28643dfe6625ff0e54bcfa63e0d04921f1a8ef7f8": Error processing tar file(exit status 1): operation not permitted
```

**Describe the results you expected:**

Simple docker image is built.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
imiell@basquiat:~/tmp  🐧  apt list buildah
Listing... Done
buildah/bionic,now 1.9.0-1~dev~ubuntu18.04~ppa17 amd64 [installed]
```

**Output of `buildah version`:**

```
imiell@basquiat:~/tmp  🐧  buildah version
Version:         1.9.0-dev
Go Version:      go1.10.4
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.4.0
libcni Version:  
Git Commit:      
Built:           Fri May 31 00:39:16 2019
OS/Arch:         linux/amd64
```

**Output of `podman version` if reporting a `podman build` issue:**

```
imiell@basquiat:~/tmp  🐧  podman version
Version:            1.4.3
RemoteAPI Version:  1
Go Version:         go1.10.4
OS/Arch:            linux/amd64
```

**Output of `cat /etc/*release`:**

```
imiell@basquiat:~/tmp  🐧  cat /etc/*release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
[...]

PR #1720, opened 2019-07-16
bump github.com/opencontainers/runtime-tools to v0.9.0

nalind:
Update to runtime-tools v0.9.0.  This requires a newer version of runtime-spec, so update that.  Some method argument types changed in the new version of runtime-tools, so fix up our interactions with it.

PR #1747, opened 2019-07-30
Add overlayfs to fuse-overlayfs tip

TomSweeneyRedHat:
This adds a note about replacing overlayfs with
fuse-overlayfs.  We've had a number of issues reported
about overlayfs not working in a rootless environment
and this will hopefully help.  Reference: #1745, #1720 and

### DIFF
--- a/troubleshooting.md
+++ b/troubleshooting.md
@@ -112,9 +112,7 @@ lstat /home/myusername/~: no such file or directory
 ---
 ### 5) Rootless buildah bud fails EPERM on NFS:
 
-NFS enforces file creation on different UIDs on the server side and does not understand User Namespace.
-When a container root process like YUM attempts to create a file owned by a different UID, NFS Server denies the creation.
-NFS is also a problem for the file locks when the storage is on it.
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
@mheon and @giuseppe PTAL.  I'll plan to add this verbiage to the Podman's troubleshooting guide too.

rhatdan:
Should the buildah man page have this information also.

giuseppe:
LGTM, but I still think we need to add fuse-overlayfs as a dependency so not everyone needs to manually configure it

rhatdan:
Yes it should be a recommends.

rhatdan:
@lsm5 FYI ^^

TomSweeneyRedHat:
I concur @giuseppe about it being a recommends, just trying to ease the pain a bit with this change.

TomSweeneyRedHat:
@rhatdan I'm not sure which man page you're referring to in Buildah.  But I'm not sure of which man page it should be in, they generally don't contain "gotcha" kind of notes.  I am working on getting this into the Podman troubleshooting guide if that is what you were thinking.

rhatdan:
Under `man podman`  There is a section o podman rootless mode.  We should mention fuse-overlayfs in that section and that section should be copied into the buildah man pages.

These are some of the most common complaints people are getting with podman/buildah running rootless.

TomSweeneyRedHat:
Thanks @rhatdan, I'll take a look at adding the info on the podman page then too.

rhatdan:
@rh-atomic-bot r+

PR #1761, opened 2019-08-02
WIP: commit/push: pay more attention to policy.json

nalind:
Since commit c654b18cad1fa1dc6a5d9b05c31ae09b4036ddda, we've been ignoring the signature verification policy to avoid cases where a default "reject" setting would prevent us from reading the source image.

Replace that with logic that reads the default or specified policy, and overrides only the settings for the local storage transport.

Then, because image copying traditionally only performs the policy check on the source location, fake up enough of an `UnparsedImage` implementation around the destination location to let us check if it would be allowed by policy, and check that.


commit.go
```diff
@@ -166,6 +191,17 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 			logrus.Debugf("error destroying signature policy context: %v", err2)
 		}
 	}()
+
+	// Check if we'd be allowed to run an image from the destination location.
+	allowed, err := policyContext.IsRunningImageAllowed(ctx, &referenceOnlyImage{dest: dest})
+	if err != nil {
+		return "", nil, "", errors.Wrapf(err, "running image with reference %q would not be allowed by policy, denying commit", transports.ImageName(dest))
```
rhatdan: This wrap should be error checkinf if running image with reference %q would not be allowed...

commit.go
```diff
@@ -166,6 +191,17 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
...
+	}
+	if !allowed {
+		return "", nil, "", errors.Errorf("running image with reference %q would not be allowed by policy, denying commit", transports.ImageName(dest))
+	}
+	logrus.Debugf("running image with reference %q would be allowed by policy", transports.ImageName(dest))
```
rhatdan: nit: 

logrus.Debugf("running image with reference %q is allowed by policy", transports.ImageName(dest))

commit.go
```diff
@@ -313,6 +364,17 @@ func Push(ctx context.Context, image string, dest types.ImageReference, options
 		}
 		maybeCachedSrc = cache
 	}
+
+	// Check if we'd be allowed to run an image from the destination location.
+	allowed, err := policyContext.IsRunningImageAllowed(ctx, &referenceOnlyImage{dest: dest})
+	if err != nil {
+		return nil, "", errors.Wrapf(err, "running image with reference %q would not be allowed by policy, denying push", transports.ImageName(dest))
```
rhatdan: Should be 
error checking if running imange with reference ...

rhatdan:
Tests areto be happy.

commit.go
nalind: An earlier revision did this, but we get a non-nil `err` back when `allowed` is false to explain the reason why `allowed` is false (an [example](https://github.com/containers/image/blob/dd0113663d0da85521e6ead3ea4ec45b050775f1/signature/policy_eval_simple.go#L28), and that phrasing made the error message read confusingly.

commit.go
nalind: Sure, changing.

nalind:
@mtrmac This is what I've got for getting a reference to an image that might not exist yet into a form that we can pass to `IsRunningImageAllowed()`.  It's not pretty, but it seems to have the desired effect, at least when the policy doesn't care about signatures.  Is there an alternate approach that you'd recommend over this before I take the WIP label off of it?

mtrmac:
I don’t understand the purpose of this, and the PR does not say; it seems to be to interpret the policy as an alternative to the blocked list, or something like that — what needs that functionality, and why does that make sense within the ecosystem?

The `policyContext.IsRunningImageAllowed` method was quite intentionally named in such a descriptive way to express the semantics of that check (and, therefore, of the `policy.json` config file as it exists now); it’s none of “is the image secure”, ”can the image be run `--privileged`”, “is the image trusted to do $X”, “is the image safe to publish to the wide internet”, “does not contain swear words”.

Why does “is running $image allowed locally” have any relationship at all to what images the node is allowed to publish elsewhere? (I can easily imagine a restricted build system that is only allowed to run the build system image, pushing arbitrary non-buildsystem build results to a registry.)

In a sense, the `policy.json` is primarily about signatures, and `insecureAcceptAnything`/`reject` are localized exceptions to that functionality; so the proposed feature that handles the two unusual cases but not signatures is also surprising.

commit.go
```diff
@@ -107,21 +107,37 @@ type PushOptions struct {
...
+// speculatively evaluate "insecureAcceptAnything" and "reject" policies for
+// images which don't actually exist yet (i.e., references for which NewImage()
+// would fail).  Attempts to evaluate "signedBy" policies will fail because we
+// don't support retrieving signature information.
+type referenceOnlyImage struct {
+	dest types.ImageReference
+}
+
+func (r *referenceOnlyImage) Reference() types.ImageReference {
+	return r.dest
```
mtrmac: (Strictly speaking, nothing quite promises not to call `types.UnparsedImage.Reference().NewImageSource()` and the like — but of course that would defeat the point of `UnparsedImage` fixing the contents of manifest+signatures in time against TOCTOU vulnerabilities.)

commit.go
```diff
@@ -157,6 +173,15 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
...
 	}
 
+	// Load the system signing policy.
+	commitPolicy, err := signature.DefaultPolicy(systemContext)
+	if err != nil {
+		return "", nil, "", errors.Wrapf(err, "error obtaining default signature policy")
+	}
+	// Override the settings for local storage to make sure that we can always read the source "image".
+	// TODO: what to do with scopes that require signatures, which we can't read until after pushing?
+	commitPolicy.Transports[is.Transport.Name()] = storageAllowedPolicyScopes
```
mtrmac: (It’s not actually documented that it is safe to modify the return value of `DefaultPolicy` — OTOH, there is no “deep copy” method, so there’s no alternative, and it is actually safe with the current implementation.

I can imagine this could break very badly if `DefaultPolicy` ever started caching the loaded policy.)

mtrmac:
> Why does “is running $image allowed locally” have any relationship at all to what images the node is allowed to publish elsewhere? (I can easily imagine a restricted build system that is only allowed to run the build system image, pushing arbitrary non-buildsystem build results to a registry.)

It might, maybe, make sense for Buildah (or more likely a higher-level automation caller) to have a `--expected-user-policy-path=policy.json-that-is-not-the-build-host-one` option, as a safety check to ensure that images published to a registry are actually correctly signed and can be used by the intended recipients. It would at least make more sense than using the local builder’s “is running an image allowed” policy. But still, if a caller did want to perform such a check, it would make much more sense to me to publish the image to a staging location, and run a more complete set of tests (functionality, anti-virus scan, license check, whatever) that also includes signature verification, before moving the image from a staging repo to a production one; it’s not something I would expect Buildah to handle.

commit.go
nalind: Yeah, it's not a code path we currently go down, but it's a potential trouble spot in the future.

commit.go
nalind: Understood.

nalind:
This is an attempt to address the "whitellist" case in the way described toward the end of https://github.com/containers/image/issues/548.  The blacklisting part (also referenced in [bugzilla #1730722](https://bugzilla.redhat.com/show_bug.cgi?id=1730722)) should be solved by https://github.com/openshift/openshift-controller-manager/pull/11 once it's merged.

nalind:
One possible alternative approach to signatures would be reading them from the source image, but that seems more fragile to me.

mtrmac:
Hum, https://github.com/containers/image/issues/548 does mention “push” quite a few times but the discussion has, it seems, never actually considered that `policy.json` does not affect destinations in any way.

IIRC (though I’m not sure about this) this feature basically exists “because it was possible in the projectatomic/docker `--block-registry` in 3.x”, and going as far back as https://github.com/moby/moby/pull/10411 I can’t find an explicit motivation for the feature / for it applying to pulls and pushes equivalently. So, it’s not currently very tempting to make this an universal feature, e.g. via `registries.conf`.

(BTW, the `registries.conf` “blocked” flag is now universally enforced in c/image, it’s no longer necessary to use `policy.json` for that. OTOH it has notably different semantics - `reject` in `policy.json` applies to the user-specified image name, `blocked` in `registries.conf` applies to a physical registry. So, if mirrors / location redirection is configured in `policy.json`, “blocked” will block individual mirrors/redirect destinations regardless of the user specification, and “reject” will block user specifications regardless of the physical mirror/redirect.)

Overall, right now it makes most sense for me to treat the whitelist as an OpenShift-specific builder configuration feature, and have the builder / whatever is configuring it read the OpenShift configuration object, and have the builder manually enforce that whitelist.

---

Even if this were a universal desirable feature (which is doubtful to me, because most users capable of running `buildah` can just as well cause an image to be pushed without involving the `buildah` binary, including by running a docker/distribution client to push the image _inside_ a buildah recipe!) , and even if `policy.json` was the best way to express the whitelist (which I really disagree with), Buildah _really should_ use a separate config file, via the `--expected-user-policy-path` option or something like that.

This PR as is makes it impossible to 1) configure a repository to require signatures on pull, and 2) push new images into that very repository — even if the user is actually signing that image as part of that push.

nalind:
> Hum, [containers/image#548](https://github.com/containers/image/issues/548) does mention “push” quite a few times but the discussion has, it seems, never actually considered that `policy.json` does not affect destinations in any way.
> 
> IIRC (though I’m not sure about this) this feature basically exists “because it was possible in the projectatomic/docker `--block-registry` in 3.x”, and going as far back as [moby/moby#10411](https://github.com/moby/moby/pull/10411) I can’t find an explicit motivation for the feature / for it applying to pulls and pushes equivalently. So, it’s not currently very tempting to make this an universal feature, e.g. via `registries.conf`.

API-wise, [this bug report](https://bugzilla.redhat.com/show_bug.cgi?id=1713389) points to [image.config.openshift.io](https://github.com/openshift/api/blob/master/config/v1/types_image.go#L10) and its `Spec.RegistrySources` field, which includes both whitelists and blocklists.  The end-goal is to get us closer to feature parity with OpenShift 3.x in terms of being able to handle all of that configuration.

> (BTW, the `registries.conf` “blocked” flag is now universally enforced in c/image, it’s no longer necessary to use `policy.json` for that. OTOH it has notably different semantics - `reject` in `policy.json` applies to the user-specified image name, `blocked` in `registries.conf` applies to a physical registry. So, if mirrors / location redirection is configured in `policy.json`, “blocked” will block individual mirrors/redirect destinations regardless of the user specification, and “reject” will block user specifications regardless of the physical mirror/redirect.)
> 
> Overall, right now it makes most sense for me to treat the whitelist as an OpenShift-specific builder configuration feature, and have the builder / whatever is configuring it read the OpenShift configuration object, and have the builder manually enforce that whitelist.

The broader feature request aims at having these settings heeded by default even in custom build images that invoke the CLI tools, which is why it's proposed here and not just in the builder repository.  I was hoping to avoid crafting another channel for passing that information to a builder, and adding custom logic to evaluate it, since `policy.json` was where the conversation seemed to have settled on passing it in.  Given the complications you're noting, I'm proposing https://github.com/openshift/openshift-controller-manager/pull/14 and https://github.com/containers/buildah/pull/1787 to see if we can do all of that instead.  There's a possibility that the schedule may not allow for it, though.

> Even if this were a universal desirable feature (which is doubtful to me, because most users capable of running `buildah` can just as well cause an image to be pushed without involving the `buildah` binary, including by running a docker/distribution client to push the image _inside_ a buildah recipe!) , and even if `policy.json` was the best way to express the whitelist (which I really disagree with), Buildah _really should_ use a separate config file, via the `--expected-user-policy-path` option or something like that.

I agree completely that since there's no privilege escalation required here, this measure can be circumvented with relative ease by someone who's looking to do so.  I gather the intention is to prevent unintended pushes, as anything more comprehensive would have to be done outside of the builder container and its code base.

The location of `policy.json` can be overridden via the APIs and the CLI `--signature-policy` option (which has been hidden since #1373), which I think provides a reasonable way of overriding it.  We use this in OpenShift to override the default file that we include in the builder image with one whose contents we generate from the cluster configuration.

> This PR as is makes it impossible to 1) configure a repository to require signatures on pull, and 2) push new images into that very repository — even if the user is actually signing that image as part of that push.

Yeah, that's a problem.  I don't think we're able to retrieve signature information that's been generated for the about-to-be-pushed content in cases where format conversion is needed.  If there's a way to supply that, I'd be fine with updating this PR to do that, too.

mtrmac:
> > and going as far back as [moby/moby#10411](https://github.com/moby/moby/pull/10411) I can’t find an explicit motivation for the feature / for it applying to pulls and pushes equivalently. 
> 
> API-wise, [this bug report](https://bugzilla.redhat.com/show_bug.cgi?id=1713389) points to [image.config.openshift.io](https://github.com/openshift/api/blob/master/config/v1/types_image.go#L10) and its `Spec.RegistrySources` field, which includes both whitelists and blocklists.

Yes, sure, and that exists because the projectatomic/docker feature exists, and that exists because…? The above moby/moby PR is as far as I got.

It seems particularly bizarre to me that the original design uses the `--add-registry` _search list_ as the whitelist for `--block-registry=all`; I guess that’s just a thing that happened without any deeper meaning to it, but if there were some kind of specific user request behind that design, I’d love to know.

(OTOH the v4 OpenShift CRD does not perpetuate this aspect, which is good.)

> > Overall, right now it makes most sense for me to treat the whitelist as an OpenShift-specific builder configuration feature, and have the builder / whatever is configuring it read the OpenShift configuration object, and have the builder manually enforce that whitelist.
> 
> The broader feature request aims at having these settings heeded by default even in custom build images that invoke the CLI tools,

Oh, that _is_ what the `RegistrySources` documentation actually seems to say. Interesting. I don’t find that 100% compelling (the comment for `RegistrySources` also says this determines how the _container runtime_ treats registries, and the runtime is not involved in pushes, so it’s fairly easy to argue that all of that documentation is just slightly imprecise), but it is certainly a reasonable reading.

But if “the CLI tools” / “container runtime … when accessing images for builds” reference includes raw `buildah` use, it probably includes raw `skopeo` use as well… that would also argue for doing this somewhere in c/image.

> I was hoping to avoid crafting another channel for passing that information to a builder, and adding custom logic to evaluate it, since `policy.json` was where the conversation seemed to have settled on passing it in.

IIRC `policy.json` was settled on was basically because it was convenient and already implicitly used by the builder for pulls (https://github.com/containers/image/issues/548 mentions it as a “short term” decision without even spelling out the “long term” one), but I can’t find the full discussion (if I recall correctly that https://github.com/containers/image/issues/548 is not everything; I may be misremembering).

> > Even if …, Buildah _really should_ use a separate config file, via the `--expected-user-policy-path` option or something like that.
> 
> The location of `policy.json` can be overridden via the APIs and the CLI `--signature-policy` option (which has been hidden since #1373), which I think provides a reasonable way of overriding it.

Ah, OK.

> > This PR as is makes it impossible to 1) configure a repository to require signatures on pull, and 2) push new images into that very repository — even if the user is actually signing that image as part of that push.
> 
> Yeah, that's a problem. I don't think we're able to retrieve signature information that's been generated for the about-to-be-pushed content in cases where format conversion is needed.

Right; actual signatures can, in general, only be created after the blobs are already written to the destination (or, I suppose, stored in some staging area).

[Strictly speaking one doesn’t need a signature to _exist_ to check whether a hypothetical signature with known image identity + key would match the policy, but that is a whole another can of worms, introducing a risky code path in the package, and checking whether the user would be able to create a valid signature and access the right keys without, in the worst case, actually forcing the user to interactively provide credentials for the private key.]

It just doesn’t seem all that practical to me involve signature checks for the, hypothetically, “staging” push, without _actually staging the image_. And if signatures are not involved, then using the signature policy for any of this is fairly surprising.

---

Anyway, if c/image/signature were to be involved, this really should not use `IsRunningImageAllowed`; there would have to be some kind of `IsAcceptableToPush` or something like that, and corresponding new sections / new syntax in `policy.json` — and if we are introducing a c/image API and new config file formats, on balance I’d much rather add an explicit whitelist (which limits both pulls and pushes) to the docker/distribution-specific `registries.json`.

(Cc: @vrothberg for another opinion.)

rhatdan:
@nalind This PR is over a year old, should we close this or reopen and start working on it?

rhatdan:
Since I never got an answer, I am going to close.

PR #1763, opened 2019-08-05
Add automatic apparmor tag discovery

saschagrunert:
The 'apparmor' build tag is now added if libappamor was found on the
local system.


saschagrunert:
Hm, is this code not executed at all?
https://github.com/containers/buildah/blob/ac5031de808b694863ac51319b77bf6e63eea845/chroot/run.go#L862-L871

vrothberg:
Fair enough. If a user wants to force-set a profile, this would work. However I didn't notice that when grepping. Thanks!

LGTM :+1:

rhatdan:
LGTM
@rh-atomic-bot r+

PR #1767, opened 2019-08-06
Bug fix for volume minus syntax

ashley-cui:
Do not check disk for existence of volume.
If volume that ends with `-` is set in config, remove it.
If not, add it into config.

https://github.com/containers/buildah/pull/1670#issuecomment-517969058

Signed-off-by: Ashley Cui <ashleycui16@gmail.com>

### DIFF
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -221,14 +221,16 @@ func updateConfig(builder *buildah.Builder, c *cobra.Command, iopts configResult
 	if c.Flag("volume").Changed {
 		if volSpec := iopts.volume; len(volSpec) > 0 {
 			for _, volVal := range volSpec {
-				_, err := os.Stat(volVal) //check if volume exists on disk
-				if err == nil {
-					builder.AddVolume(volVal)
-				} else if strings.HasSuffix(volVal, "-") {
+				if strings.HasSuffix(volVal, "-") {
 					rmVol := strings.TrimSuffix(volVal, "-")
-					builder.RemoveVolume(rmVol)
+					if builder.CheckVolume(rmVol) {
+						builder.RemoveVolume(rmVol)
+					} else {
+						fmt.Printf("volume %s cannot be removed because it does not exist, adding volume %s\n", rmVol, volVal)
+						builder.AddVolume(volVal)
+					}
 				} else {
-					logrus.Errorf("Volume cannot be added because it does not exist on disk.")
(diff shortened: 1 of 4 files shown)


cmd/buildah/config.go
```diff
@@ -221,14 +221,15 @@ func updateConfig(builder *buildah.Builder, c *cobra.Command, iopts configResult
...
-				if err == nil {
-					builder.AddVolume(volVal)
-				} else if strings.HasSuffix(volVal, "-") {
+				if strings.HasSuffix(volVal, "-") {
 					rmVol := strings.TrimSuffix(volVal, "-")
-					builder.RemoveVolume(rmVol)
+					if builder.CheckVolume(rmVol) {
+						builder.RemoveVolume(rmVol)
+					} else {
+						builder.AddVolume(volVal)
```
TomSweeneyRedHat: Should we really add the volume at this point?  The user has specified a volume with a dash meaning they want to remove it.  Shouldn't this be a no-op or perhaps a logrus.Debugf saying the volume didn't exist?

ashley-cui: #1670, https://github.com/containers/buildah/pull/1670#issuecomment-507631089

Checking if the file exists on disk broke buildah, so @rhatdan said to just leave it like this.

rhatdan: @TomSweeneyRedHat If we don't add the volume at this point, it was pointed out to us there would be no way to add a volume ending in a "-"

TomSweeneyRedHat: OK, think I got it now.  so a volume named 'myvol-' is valid?  Ugh, that makes me like this overall change even more.

My big issue with this logic then is if the user meant to remove 'myvol', but instead typed 'myvoll-'.  Instead of removing 'myvol' as intended, they now still have 'myvol' but also now have 'myvoll' and it all happens silently right?

Should we add an output line here telling the user that we're adding the volume to at least hopefully let them know?

ashley-cui: I'm okay with that. @rhatdan ?

rhatdan: Yes lets put out an info message, telling them myvol does not exist, adding myvol-
Tests are blowing up.

TomSweeneyRedHat:
LGTM, fingers crossed for the tests.

rhatdan:
@rh-atomic-bot r+

PR #1771, opened 2019-08-07
 Bump containers/image to v3.0.2 to fix keyring issue 

rhatdan:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/blang/semver v3.5.0+incompatible // indirect
 	github.com/containerd/continuity v0.0.0-20181203112020-004b46473808 // indirect
 	github.com/containernetworking/cni v0.7.1
-	github.com/containers/image v3.0.1+incompatible
+	github.com/containers/image v3.0.2+incompatible
 	github.com/containers/storage v1.13.1
 	github.com/cyphar/filepath-securejoin v0.2.1
 	github.com/docker/distribution v0.0.0-20170817175659-5f6282db7d65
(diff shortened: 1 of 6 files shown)


rhatdan:
@TomSweeneyRedHat @giuseppe @vrothberg @nalind @ashley-cui @QiWang19 PTAL

TomSweeneyRedHat:
If the dust settles in https://github.com/containers/libpod/pull/3752 on the directory location discussion, then LGTM

pkg/cli/common.go
```diff
@@ -261,5 +261,10 @@ func GetDefaultAuthFile() string {
 	if runtimeDir != "" {
 		return filepath.Join(runtimeDir, "containers/auth.json")
 	}
-	return ""
+
+	uid := os.Getuid()
+	if uid == 0 {
+		return fmt.Sprintf("/run/containers/%d/auth.json", os.Getuid())
+	}
+	return fmt.Sprintf("/run/user/%d/containers/auth.json", os.Getuid())
```
TomSweeneyRedHat: This differs from https://github.com/containers/libpod/pull/3752 and I'm not sure it's quite right as I think root would hit 267 when it should hit 269.

mtrmac: The actual behavior depends on whether `XDG_RUNTIME_DIR` is set, and su/sudo clears it. It is not strictly specific to UID=0.

rhatdan:
@mtrmac @TomSweeneyRedHat @QiWang19 @nalind @vrothberg PTAL

TomSweeneyRedHat:
LGTM
Now starting a happy ci dance....

PR #1784, opened 2019-08-12
Allow buildah bud to be called without arguments

saschagrunert:
This allows to simply call `buildah bud`, which assumes the current
working directory as build context. This directory should contain a
Dockerfile, which has been mentioned in the docs as well.

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -23,9 +23,11 @@ type budResults struct {
 }
 
 func init() {
-	var (
-		budDescription = "\n  Builds an OCI image using instructions in one or more Dockerfiles."
-	)
+	budDescription := `
+  Builds an OCI image using instructions in one or more Dockerfiles.
+
+  If no arguments specified, it will assume the current working directory as
+  build context, which should contain the Dockerfile.`
 
 	layerFlagsResults := buildahcli.LayerResults{}
 	budFlagResults := buildahcli.BudResults{}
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@nalind @giuseppe @TomSweeneyRedHat @vrothberg @QiWang19 @ashley-cui PTAL

QiWang19:
LGTM

QiWang19:
Do we need to change the man page? buildah-bud.md

rhatdan:
It was changed in the PR.
@rh-atomic-bot r+

cmd/buildah/bud.go
```diff
@@ -23,9 +23,11 @@ type budResults struct {
 }
 
 func init() {
-	var (
-		budDescription = "\n  Builds an OCI image using instructions in one or more Dockerfiles."
-	)
+	budDescription := `
+  Builds an OCI image using instructions in one or more Dockerfiles.
+
+  If no arguments specified, it will assume the current working directory as
```
TomSweeneyRedHat: Day late and a dollar short I am, but  s/arguments specified/arguments are specified/

TomSweeneyRedHat:
It would be nice to have a test or two for this.  Also the man page changes will have to happen in podman's build man page too.

cmd/buildah/bud.go
saschagrunert: I fix this with the next PR.

vrothberg:
LGTM
Note for the next release: this change requires us to bump the minor version.

PR #1785, opened 2019-08-12
add: teach copyFileWithTar() about symlinks and directories

nalind:
Teach `copyFileWithTar()` about symbolic links and directories, and use it to produce tar data to feed to `untar()` instead of special-casing them.

### DIFF
--- a/util.go
+++ b/util.go
@@ -120,16 +120,13 @@ func (b *Builder) copyFileWithTar(tarIDMappingOptions *IDMappingOptions, chownOp
 		}
 	}
 	return func(src, dest string) error {
+		var f *os.File
+
 		logrus.Debugf("copyFileWithTar(%s, %s)", src, dest)
-		f, err := os.Open(src)
+		fi, err := os.Lstat(src)
 		if err != nil {
-			return errors.Wrapf(err, "error opening %q to copy its contents", src)
+			return errors.Wrapf(err, "error reading attributes of %q", src)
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

rhatdan:
@TomSweeneyRedHat @vrothberg @giuseppe @ashley-cui @QiWang19 PTAL

vrothberg:
@rh-atomic-bot r+

util.go
```diff
@@ -120,16 +120,13 @@ func (b *Builder) copyFileWithTar(tarIDMappingOptions *IDMappingOptions, chownOp
 		}
 	}
 	return func(src, dest string) error {
+		var f *os.File
+
 		logrus.Debugf("copyFileWithTar(%s, %s)", src, dest)
-		f, err := os.Open(src)
+		fi, err := os.Lstat(src)
```
TomSweeneyRedHat: Since we're doing the system.Lstat below at line 131, is this one required?

TomSweeneyRedHat:
One question and a homu fail, otherwise LGTM

util.go
giuseppe: system.Lstat seems like a portable way to retrieve the UID/GID, unfortunately it costs another stat as it cannot be used as a `FileInfo`

giuseppe:
@rh-atomic-bot retry

PR #1787, opened 2019-08-13
commit/push: pay attention to $BUILD_REGISTRY_SOURCES

nalind:
If $BUILD_REGISTRY_SOURCES is set, expect it to be a valid `github.com/openshift/api/configv1.Image`, and parse its `AllowedRegistries` and `BlockedRegistries` lists when committing or pushing images.
    
Override the local signature policy when committing or pushing an image to ensure that local storage is always allowed.


TomSweeneyRedHat:
Should we hold this until https://github.com/openshift/openshift-controller-manager/pull/14 gets in?

nalind:
Yes, this is entirely dependent on that.

rhatdan:
This needs some mention in the man pages.

nalind:
Rebased, added some docs, and wired the check into the "pull" paths as well.

rhatdan:
LGTM

commit.go
```diff
@@ -292,10 +370,25 @@ func Push(ctx context.Context, image string, dest types.ImageReference, options
 		return nil, "", errors.Errorf("push access to registry for %q is blocked by configuration", transports.ImageName(dest))
 	}
 
+	// Load the system signing policy.
+	pushPolicy, err := signature.DefaultPolicy(systemContext)
+	if err != nil {
+		return nil, "", errors.Wrapf(err, "error obtaining default signature policy")
+	}
+	// Override the settings for local storage to make sure that we can always read the source "image".
+	// TODO: what to do with scopes that require signatures, which we can't read until after pushing?
```
mtrmac: Nit: I don’t think these two `TODO`s are applicable any more.

nalind: Ah, right.  Dropping them.

rhatdan:
@nalind is this still a WIP?

nalind:
We could merge it in its current form (give or take CI), but https://github.com/openshift/openshift-controller-manager/pull/14 needs to be merged before it'll be useful in OpenShift.

rhatdan:
Ok lets take WIP off and get it to pass CI, then we can merge.

rhatdan:
@rh-atomic-bot r+

PR #1788, opened 2019-08-13
Bud integration tests and documentation update

saschagrunert:
This is a small follow-up of https://github.com/containers/buildah/pull/1784, where documentation improvements and integration tests should be applied.

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -26,8 +26,9 @@ func init() {
 	budDescription := `
   Builds an OCI image using instructions in one or more Dockerfiles.
 
-  If no arguments specified, it will assume the current working directory as
-  build context, which should contain the Dockerfile.`
+  If no arguments are specified, Buildah will use the current working directory
+  as the build context and look for a Dockerfile. The build fails if no
+  Dockerfile is present.`
 
 	layerFlagsResults := buildahcli.LayerResults{}
 	budFlagResults := buildahcli.BudResults{}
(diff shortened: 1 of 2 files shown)


cmd/buildah/bud.go
```diff
@@ -26,8 +26,8 @@ func init() {
 	budDescription := `
   Builds an OCI image using instructions in one or more Dockerfiles.
 
-  If no arguments specified, it will assume the current working directory as
-  build context, which should contain the Dockerfile.`
+  If no are arguments specified, it will assume the current working directory
```
vrothberg: "are" should be moved behind "arguments"

cmd/buildah/bud.go
```diff
@@ -26,8 +26,8 @@ func init() {
...
+  as build context, which should contain the Dockerfile.`
```
vrothberg: Independent of this change, I find the sentence a bit hard to brain parse and it doesn't explain that it'll fail otherwise.

"If no arguments are specified, Buildah will use the current working directory as the build context and look for a Dockerfile. The build fails if no Dockerfile is present."

cmd/buildah/bud.go
saschagrunert: Fixed with the comment below.

cmd/buildah/bud.go
saschagrunert: Changed as suggested.

giuseppe:
LGTM

@rh-atomic-bot r+

TomSweeneyRedHat:
LGTM, thanks for the follow up @saschagrunert

saschagrunert:
Rebased on the master because the travis failure reported that the revision does not exist any more :thinking:

rhatdan:
@saschagrunert Sadly this needs a rebase.

saschagrunert:
Rebased on top of the latest master branch.

rhatdan:
@rh-atomic-bot r+

PR #1789, opened 2019-08-13
add: handle hard links when copying with .dockerignore

nalind:
Detect files that are hard linked together, and use that information to avoid copying their contents more often than we have to.

### DIFF
--- a/util.go
+++ b/util.go
@@ -119,6 +119,8 @@ func (b *Builder) copyFileWithTar(tarIDMappingOptions *IDMappingOptions, chownOp
 			HostGIDMapping: true,
 		}
 	}
+
+	var hardlinkChecker util.HardlinkChecker
 	return func(src, dest string) error {
 		var f *os.File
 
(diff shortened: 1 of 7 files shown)


rhatdan:
@QiWang19 We might need similar fixes for podman cp.

util/util_unix.go
```diff
@@ -0,0 +1,31 @@
...
+type hardlinkDeviceAndInode struct {
+	device, inode uint64
+}
+
+type HardlinkChecker struct {
+	hardlinks sync.Map
+}
+
+func (h *HardlinkChecker) Check(fi os.FileInfo) string {
+	if stat, ok := fi.Sys().(*syscall.Stat_t); ok && stat.Nlink > 1 {
```
rhatdan: Should you avoid directories?

util.go
```diff
@@ -144,22 +146,34 @@ func (b *Builder) copyFileWithTar(tarIDMappingOptions *IDMappingOptions, chownOp
...
+				hdr.Linkname = linkname
+			} else {
+				hardlinkChecker.Add(fi, chrootedDest)
 			}
-			defer func() {
-				if f != nil {
-					if err := f.Close(); err != nil {
-						logrus.Debugf("error closing %s: %v", f.Name(), err)
-					}
+			if hdr.Typeflag == tar.TypeReg {
```
rhatdan: Is this check necessary, isnce you have checked it above?

nalind: We might be changing it from TypeReg to TypeLink at line 159.

util/util_unix.go
nalind: For correctness, yes.  We shouldn't call this except for regular files, but still.

util.go
rhatdan: Then why not put this in else block above.

nalind: Good point, changing it.

rhatdan:
@rh-atomic-bot r+

Issue #1778, opened 2019-08-09
Feature: add "--squash-older-than DATE" argument to buildah commit


TristanCacqueray:
Buildah commit currently only support a `--squash` argument to produce an image with one layer. This issue is about a feature to implement a `--squash-older-than AGE` or `--squash-max-layers N` argument to produce an image that only keeps recent layers.

The use-case is for using an image similarly to a regular fedora workstation runtime:

Initialization:
```bash
ctx = $(buildah from fedora)
buildah commit $ctx local-fedora
buildah delete $ctx
```

Package installation: `layer(intermediate-base)`
```bash
ctx = $(buildah from local-fedora)
buildah run $ctx dnf install -y firefox
buildah commit $ctx local-fedora
buildah delete $ctx
```

Then periodically update the image: `layer(intermediate-base)` + (N-1) * `layer(update)` + `layer(update)`
```bash
ctx = $(buildah from local-fedora)
buildah run $ctx dnf update -y
buildah commit $ctx local-fedora:$(date "+%Y-%m-%d")
buildah delete $ctx
buildah tag local-fedora:$(date "+%Y-%m-%d") local-fedora:latest
```

The issue with that approach is that the layers are never removed and passing a certain numbers, things start to get slower and slower. Using a `--squash-older-than` argument would enable a more efficient image while keeping the regular dnf update in-place workflow.

Another solution discussed on IRC is that the `layer(update)` is never re-used and instead produced each time from `layer(intermediate-base)`. This can be optimized using `-v /var/cache/dnf:/var/cache/dnf` when updating.
Though there is a performance cost because daily update may update the same package over and over until the `layer(intermediate-base)` is rebuilt cleanly

rhatdan:
@nalind @vrothberg @TomSweeneyRedHat WDYT?

TomSweeneyRedHat:
I like the thought, I'm just not sure how well squishing some but not all layers will work code wise.

TristanCacqueray:
@Hendrik-H yes, this is also mentioned in https://github.com/containers/buildah/pull/1790#issuecomment-534579845

rhatdan:
Ok lets see about moving forward on this.

rhatdan:
Is anyone going to work on this in the community, if not I will close in a month.

PR #1790, opened 2019-08-13
commit: add --max-layers argument

TristanCacqueray:
Fixes: #1778

### DIFF
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -32,6 +32,7 @@ type commitInputOptions struct {
 	rm                 bool
 	signaturePolicy    string
 	squash             bool
+	maxLayers          uint
 	tlsVerify          bool
 }
 
@@ -83,6 +84,7 @@ func init() {
 	}
 
 	flags.BoolVar(&opts.squash, "squash", false, "produce an image with only one layer")
+	flags.UintVar(&opts.maxLayers, "max-layers", ^uint(0), "produce an image with at most N layer")
 	flags.BoolVar(&opts.tlsVerify, "tls-verify", true, "Require HTTPS and verify certificates when accessing the registry")
 
 	rootCmd.AddCommand(commitCommand)
@@ -100,7 +102,12 @@ func commitCmd(c *cobra.Command, args []string, iopts commitInputOptions) error 
 	if err := buildahcli.CheckAuthFile(iopts.authfile); err != nil {
 		return err
 	}
-
+	if iopts.maxLayers != ^uint(0) && iopts.squash && iopts.maxLayers != 1 {
+		return errors.Errorf("--squash and --max-layers are mutually exclusive")
+	}
+	if iopts.maxLayers < 1 {
+		return errors.Errorf("--max-layers can't be set to 0")
+	}
 	name := args[0]
 	args = Tail(args)
 	if len(args) > 1 {
@@ -170,6 +177,7 @@ func commitCmd(c *cobra.Command, args []string, iopts commitInputOptions) error 
 		SystemContext:         systemContext,
 		IIDFile:               iopts.iidfile,
 		Squash:                iopts.squash,
+		MaxLayers:             iopts.maxLayers,
 		BlobDirectory:         iopts.blobCache,
 		OmitTimestamp:         iopts.omitTimestamp,
 	}
--- a/docs/buildah-commit.md
+++ b/docs/buildah-commit.md
@@ -64,7 +64,12 @@ Default leaves the container and its content in place.
 
 **--squash**
 
-Squash all of the new image's layers (including those inherited from a base image) into a single new layer.
+Squash all of the new image's layers (including those inherited from a base image** into a single new layer. This conflicts when **--max-layers** is set to more than 1.
+
+**--max-layers** *uint-value*
+
+Squash any older layers so that the new image's layer count doesn't exceed
+the max layers parameter value. This conflicts with **--squash** when set to more than 1.
 
 **--tls-verify** *bool-value*
 
--- a/image.go
+++ b/image.go
@@ -56,6 +56,7 @@ type containerImageRef struct {
 	preferredManifestType string
 	exporting             bool
 	squash                bool
+	maxLayers             uint
 	emptyLayer            bool
 	tarPath               func(path string) (io.ReadCloser, error)
 	parent                string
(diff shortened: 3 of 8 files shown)


TristanCacqueray:
I'm not very familiar with the code base, but that change seems to work. Would that be the right strategy to implement #1778 @TomSweeneyRedHat ?

TristanCacqueray:
This change needs more assertion to prevent an invalid max-layers argument, and the new history order may be incorrect.

docs/buildah-commit.md
```diff
@@ -64,7 +64,11 @@ Default leaves the container and its content in place.
 
 **--squash**
 
-Squash all of the new image's layers (including those inherited from a base image) into a single new layer.
+Squash all of the new image's layers (including those inherited from a base image** into a single new layer.
```
rhatdan: Should conflict with --squash

image.go
```diff
@@ -180,8 +181,14 @@ func (i *containerImageRef) createConfigsAndManifests() (v1.Image, v1.Manifest,
 	oimage.RootFS.DiffIDs = []digest.Digest{}
 	// Only clear the history if we're squashing, otherwise leave it be so that we can append
 	// entries to it.
-	if i.squash {
-		oimage.History = []v1.History{}
+	if i.squash || i.squashMaxLayers > 0 {
+		newHistory := []v1.History{}
+		for idx, hist := range oimage.History {
```
rhatdan: Should only run this code if !i.squash

image.go
```diff
@@ -198,8 +205,14 @@ func (i *containerImageRef) createConfigsAndManifests() (v1.Image, v1.Manifest,
 	dimage.RootFS.DiffIDs = []digest.Digest{}
 	// Only clear the history if we're squashing, otherwise leave it be so that we can append
 	// entries to it.
-	if i.squash {
-		dimage.History = []docker.V2S2History{}
+	if i.squash || i.squashMaxLayers > 0 {
+		newHistory := []docker.V2S2History{}
+		for idx, hist := range dimage.History {
```
rhatdan: Should not run this code if i.squash.

rhatdan:
Not hugely familiar with this code, but should --squash set the --squash-max-layers flag to1?

docs/buildah-commit.md
TristanCacqueray: Commented

image.go
TristanCacqueray: Fixed

image.go
TristanCacqueray: Fixed

TristanCacqueray:
Last commit renamed the argument to "--max-layers" as it feels more natural.

TristanCacqueray:
Last commit fixed the history order. Please let me know if the implementation is correct or how can it be improved.

cmd/buildah/commit.go
```diff
@@ -97,6 +99,12 @@ func commitCmd(c *cobra.Command, args []string, iopts commitInputOptions) error
 	if err := buildahcli.VerifyFlagsArgsOrder(args); err != nil {
 		return err
 	}
+	if iopts.maxLayers != ^uint(0) && iopts.squash && iopts.maxLayers != 1 {
+		return errors.Errorf("--squash and --max-layers are mutually exclusive")
+	}
+	if iopts.maxLayers == 0 {
```
TomSweeneyRedHat: pref "< 1"

image.go
```diff
@@ -168,6 +169,11 @@ func (i *containerImageRef) extractRootfs() (io.ReadCloser, error) {
 func (i *containerImageRef) createConfigsAndManifests() (v1.Image, v1.Manifest, docker.V2Image, docker.V2S2Manifest, error) {
 	created := i.created
 
+	if i.maxLayers == 0 {
```
TomSweeneyRedHat: pref "< 1"

image.go
```diff
@@ -229,6 +253,11 @@ func (i *containerImageRef) createConfigsAndManifests() (v1.Image, v1.Manifest,
 }
 
 func (i *containerImageRef) NewImageSource(ctx context.Context, sc *types.SystemContext) (src types.ImageSource, err error) {
+	if i.maxLayers == 0 {
```
TomSweeneyRedHat: pref "< 1"

image.go
```diff
@@ -344,6 +400,15 @@ func (i *containerImageRef) NewImageSource(ctx context.Context, sc *types.System
 			if err != nil {
 				return nil, err
 			}
+		} else if toBeSquashed(layerID) {
+			// Extract all the extra layers into a single diff
+			firstID := layersToSquash[len(layersToSquash)-1]
+			lastID := layersToSquash[0]
+			rc, err = i.store.Diff(firstID, lastID, diffOptions)
+			logrus.Debugf("squashing old layers from %q to %q", firstID, lastID)
```
TomSweeneyRedHat: :+1:

cmd/buildah/commit.go
TristanCacqueray: Done

image.go
TristanCacqueray: Done

image.go
TristanCacqueray: Done

TristanCacqueray:
Thanks for the review, I'll push a few more commits soon.

TristanCacqueray:
Turns out `store.Diff()` doesn't return the from layer content when from != to.Parent, thus this only work when committing an image built `from scratch`. When testing with `from image`, the resulting commit is missing the original from layer.

It seems like the storage driver operate differently when from != to.Parent and I'm not sure how to produce a diff that includes the from layer.

TristanCacqueray:
I couldn't find why `store.Diff()` doesn't work as expected. Actually `store.Mount()` work better and https://github.com/containers/buildah/pull/1790/commits/999a5cb2a76236304891cd790e424cc5fada7e5a refactor the logic a bit as mounting is easier to use: only the top layerID needs to be tracked.

With this last update, I was able to squash a 50 layers (6GB) fedora image down to 8 layers (4.5GB).

TristanCacqueray:
@rhatdan the PR is working as expected. Though it could use a squash, I left the `store.Mount` fix in a separate commit on top as @TomSweeneyRedHat only reviewed the `store.Diff` strategy.

I think we are also waiting for @nalind feedback. And I'd be happy to follow-up suggestions.

TomSweeneyRedHat:
@nalind, could you peruse this please and we'll merge if it LGTY?

TristanCacqueray:
Perhaps `--max-layers` is not a good name. There actually multiple strategies that could be considered:

* [squash from 1 to n-max] + n-max..n   (the current implementation, discard the oldest layer, useful for rolling update of an image)
* [1] + [squash from 2 to n-max-1] + n-max-1..n   (to keep the initial layer) 
* 1..n-max + [squash from n-max to n]    (to keep the oldest, discard the newest, useful for rebasing updates to a multi-layer images)

TomSweeneyRedHat:
@TristanCacqueray needs a rebase

TristanCacqueray:
@TomSweeneyRedHat alright, rebase was trivial

TomSweeneyRedHat:
@nalind, @rhatdan , @vrothberg PTAL
Thanks @TristanCacqueray !

TomSweeneyRedHat:
@nalind @rhatdan @vrothberg RE-PING!

TristanCacqueray:
Actually I think the second use-case mentioned in https://github.com/containers/buildah/pull/1790#issuecomment-534579845 would be more useful. On a big image, the current implementation is quite slow when the squash happens. I think this review shouldn't touch the initial layer, and let the user manage it manually, for example by using `--squash` periodically.

rhatdan:
@nalind @mtrmac @vrothberg @giuseppe This is an old PR but might be worth moving forward on.

rhatdan:
Since this PR has not been touched in a year, I am going to close.  Reopen if more work is going to be done on it.

PR #1793, opened 2019-08-14
use correct path for ginkgo

lsm5:


lsm5:
ah the GOMD2MAN seems to have come from a prior change, but it seems relevant too, would allow a non-default go-md2man binary.

rhatdan:
LGTM

TomSweeneyRedHat:
@rh-atomic-bot r+

PR #1794, opened 2019-08-15
Add security-related volume options to validator

mheon:
Also, ensure ordering is correct on other entries. The current code will increment *after* checking for validity in some cases, which allows such options as '-v /tmp:/test1:ro,rw' to pass validation when they obviously don't make much sense.

Inspired by https://github.com/containers/libpod/issues/3819 on the Podman side


TomSweeneyRedHat:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #1799, opened 2019-08-16
Add --digestfile and Re-add push statement as debug

TomSweeneyRedHat:
We removed the "Succesfully pushed ..." statement from the end of
the `buildah push` command in #1744.  However we've had a request
for that information still so I'm re-adding it as a debug statement
and we can discuss if there's a better way or not.

Also add the --digestfile option.  With it the user will be able
to specify a file to write the digest of the pushed image to for
reference.

### DIFF
--- a/cmd/buildah/push.go
+++ b/cmd/buildah/push.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"strings"
 
@@ -24,6 +25,7 @@ type pushResults struct {
 	blobCache          string
 	certDir            string
 	creds              string
+	digestfile         string
 	disableCompression bool
 	format             string
 	quiet              bool
@@ -65,6 +67,7 @@ func init() {
 	flags.StringVar(&opts.blobCache, "blob-cache", "", "assume image blobs in the specified directory will be available for pushing")
 	flags.StringVar(&opts.certDir, "cert-dir", "", "use certificates at the specified path to access the registry")
 	flags.StringVar(&opts.creds, "creds", "", "use `[username[:password]]` for accessing the registry")
+	flags.StringVar(&opts.digestfile, "digestfile", "", "after copying the image, write the digest of the resulting image to the file")
 	flags.BoolVarP(&opts.disableCompression, "disable-compression", "D", false, "don't compress layers")
 	flags.StringVarP(&opts.format, "format", "f", "", "manifest type (oci, v2s1, or v2s2) to use when saving image using the 'dir:' transport (default is manifest type of source)")
 	flags.BoolVarP(&opts.quiet, "quiet", "q", false, "don't output progress information when pushing images")
@@ -175,6 +178,15 @@ func pushCmd(c *cobra.Command, args []string, iopts pushResults) error {
 	} else {
 		logrus.Debugf("pushed image with digest %s", digest.String())
 	}
+
+	logrus.Debugf("Successfully pushed %s with digest %s", transports.ImageName(dest), digest.String())
+
+	if iopts.digestfile != "" {
+		if err = ioutil.WriteFile(iopts.digestfile, []byte(digest.String()), 0644); err != nil {
+			return util.GetFailureCause(err, errors.Wrapf(err, "failed to write digest to file %q", iopts.digestfile))
+		}
+	}
+
 	return nil
 }
 
--- a/docs/buildah-push.md
+++ b/docs/buildah-push.md
@@ -61,6 +61,10 @@ The [username[:password]] to use to authenticate with the registry if required.
 If one or both values are not supplied, a command line prompt will appear and the
 value can be entered.  The password is entered without echo.
 
+**--digestfile** *Digestfile*
+
+After copying the image, write the digest of the resulting image to the file.
+
 **--disable-compression, -D**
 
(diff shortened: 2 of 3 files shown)


rhatdan:
LGTM

rhatdan:
@TomSweeneyRedHat Why are they asking for this?

cmd/buildah/push.go
```diff
@@ -175,6 +175,7 @@ func pushCmd(c *cobra.Command, args []string, iopts pushResults) error {
 	} else {
 		logrus.Debugf("pushed image with digest %s", digest.String())
 	}
+	logrus.Debugf("Successfully pushed %s@%s\n", dest.StringWithinTransport(), digest.String())
```
mtrmac: If this does not have to restore the previous string exactly, I’d prefer
```suggestion
	logrus.Debugf("Successfully pushed to %s with digest %s\n", transports.ImageName(dest), digest.String())
```
e.g. `Successfully pushed to docker://example.com/busybox:latest with digest sha256:aaaa…aaaa`; that way the “pushed to…” string is actually a valid reference, unlike the use of either `StringWithinTransport` or joining the result of that with `@`_digest_.

Ultimately, as long as it is a debug text and not an API promise (or maybe it _is_ intended to be a promise per #1744 ??), anything goes, of course.

rhatdan: I like the idea of --verbose and then output the information that @mtrmac suggests, as well as logging it as a debugf.

mtrmac: OTOH see https://github.com/containers/buildah/pull/1744#issuecomment-522233304 , do we want to make an actual safe API (instead of / as well as) strings in a stream of text?

(The API concern is not a reason not to add this text as well.)

TomSweeneyRedHat:
@rhatdan the why is listed in #1744, they'd come to rely on that data as a way to track which images they'd pushed.

cmd/buildah/push.go
TomSweeneyRedHat: OK, I've taken @mtrmac suggestion for the "Successfully pushed ..." line and kept it as a Debug statement.  As noted in #1744 I've created a --digestfile option and now write out the digest which is the value found in `digest.String()` and that looks something like:  `sha256:b552b401d9977f4da64e410727e84a736be5b78f96b561a9b21579b3d1d74a8f`.  That's all the stuff to the right of the `@` previously.   @ wwilson  if that needs to be adjusted for your use case, please let us know.

I'll be pushing a new commit to #1744 after a bit more testing.

TomSweeneyRedHat: oops, to many prs, I'll repush here at #1799.

docs/buildah-push.md
```diff
@@ -61,6 +61,10 @@ The [username[:password]] to use to authenticate with the registry if required.
 If one or both values are not supplied, a command line prompt will appear and the
 value can be entered.  The password is entered without echo.
 
+**--digestfile** *Digestfile*
+
+Write the digest to the file.
```
mtrmac: “The digest of the resulting image”, ”the digest of the image created at the destination”, or something like that? “The digest” does not really say which one (and the value may differ from the digest of the source).

Not sure, maybe something like “After copying the image” could be useful to indicate that this does not prevent the default action (pushing an image) from happening. OTOH the example added below is clear enough about that.

docs/buildah-push.md
```diff
@@ -95,6 +99,10 @@ This example extracts the imageID image to a container registry named registry.e
 
  `# buildah push imageID docker://registry.example.com/repository:tag`
 
+This example extracts the imageID image to a container registry named registry.example.com and saves the digest ID in the specified didigestfile.
```
mtrmac: “the digest”, or maybe “the digest value”, not “the digest ID”, please.

Also see above WRT saying that it is a digest at the destination.

rhatdan:
LGTM after fixing @mtrmac comments.

docs/buildah-push.md
TomSweeneyRedHat: Changed to:
`After copying the image, write the digest of the resulting image to the file.`

docs/buildah-push.md
TomSweeneyRedHat: changed to "saves the digest"

docs/buildah-push.md
```diff
@@ -95,6 +99,10 @@ This example extracts the imageID image to a container registry named registry.e
 
  `# buildah push imageID docker://registry.example.com/repository:tag`
 
+This example extracts the imageID image to a container registry named registry.example.com and saves the digest in the specified digestfile.
```
mtrmac: One more non-blocking comment: The “extracts” verb is used in all examples, and does not really match in _any_ of them.

At least this new one should probably say “pushes”.

TomSweeneyRedHat: Yeah I noted that this weekend and wasn't happy with "extracts" either.  But didn't come up with a better word quickly so just kept this one in line with the others.

TomSweeneyRedHat: reworked that too in the latest commit.

rhatdan:
LGTM
@rh-atomic-bot r+
@TomSweeneyRedHat Do we need the same feature in podman?

TomSweeneyRedHat:
@rhatdan I was planning to add it to Podman too, we'll see if @mheon thinks it's a need or not!  :^)

Issue #1797, opened 2019-08-16
pull of local OCI image always refers to docker.io/library


saschagrunert:
**Description**

If we pull an image from a local OCI directory, the referring image name will always point to `docker.io/library` instead of `localhost` (what I would expect).

**Steps to reproduce the issue:**
```
> buildah pull quay.io/saschagrunert/hello-world
> buildah push quay.io/saschagrunert/hello-world oci:hello-world
> buildah rmi quay.io/saschagrunert/hello-world
> buildah pull oci:hello-world
```

**Describe the results you received:**

```
> buildah images
REPOSITORY                      TAG      IMAGE ID       CREATED      SIZE
docker.io/library/hello-world   latest   67856ef00418   3 days ago   2.05 KB
```

**Describe the results you expected:**

```
> buildah images
REPOSITORY                      TAG      IMAGE ID       CREATED      SIZE
localhost/hello-world           latest   67856ef00418   3 days ago   2.05 KB
```

or (even harder)

```
> buildah images
REPOSITORY                          TAG      IMAGE ID       CREATED      SIZE
quay.io/saschagrunert/hello-world   latest   0640a06a59cc   3 days ago   2.9 KB
```

**Output of `rpm -q buildah` or `apt list buildah`:**

```
> rpm -q buildah
buildah-1.10.0-1.4.x86_64
```

**Output of `buildah version`:**

```
> buildah version
Version:         1.10.0
Go Version:      go1.11.12
Image Spec:      1.0.1
Runtime Spec:    1.0.1-dev
CNI Spec:        0.4.0
libcni Version:  v0.7.0-rc2
Git Commit:      unknown
Built:           Thu Aug 15 19:01:59 2019
OS/Arch:         linux/amd64
```

**Output of `cat /etc/*release`:**

```
> cat /etc/*release
NAME="openSUSE Tumbleweed"
# VERSION="20190814"
ID="opensuse-tumbleweed"
ID_LIKE="opensuse suse"
VERSION_ID="20190814"
PRETTY_NAME="openSUSE Tumbleweed"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:tumbleweed:20190814"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org/"
LOGO="distributor-logo"
```

**Output of `uname -a`:**

```
> uname -srvmpio
[...]

PR #1800, opened 2019-08-19
Fix directory pull image names

saschagrunert:
This is a breaking change and modifies the resulting image name when
pull from an directory via `oci:...` or `dir:...`.

Without this patch, the image names pulled via a local directory got
prefixed with `docker.io/{library/}`, which is not correct.

We now use either the full path to the image, or the relative path as
image name, but prefixed with `localhost` to indicate that the image is
not being pulled from a remote location.

Closes: https://github.com/containers/buildah/issues/1797

### DIFF
--- a/pull.go
+++ b/pull.go
@@ -102,19 +102,11 @@ func localImageNameForReference(ctx context.Context, store storage.Store, srcRef
 		}
 	case directory.Transport.Name():
 		// supports pull from a directory
-		name = srcRef.StringWithinTransport()
-		// remove leading "/"
-		if name[:1] == "/" {
-			name = name[1:]
-		}
+		name = toLocalImageName(srcRef.StringWithinTransport())
 	case oci.Transport.Name():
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@giuseppe @vrothberg @TomSweeneyRedHat @nalind @QiWang19 PTAL

TomSweeneyRedHat:
The change LGTM, but I'd like to also have a head nod from @nalind in case I'm forgetting a reason for naming the image with docker.io prior.  I don't think there was, but....

nalind:
The `docker.io/library` prefix is a hardwired default that we pick up from the image library when it's asked for the full version of a parsed image's name (an `ImageReference`), and there was no domain portion in the value that we originally parsed.  This should prevent that.

I'm still not sure about using the whole pathname as part of the image's name, but that's not something that this changes.

@rh-atomic-bot r+

PR #1801, opened 2019-08-19
README: mention that Podman uses Buildah's API

vrothberg:

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,8 @@ into other tools.
 
 Podman specializes in all of the commands and functions that help you to maintain and modify
 OCI images, such as pulling and tagging.  It also allows you to create, run, and maintain those containers
-created from those images.
+created from those images.  For building container images via Dockerfiles, Podman uses Buildah's
+golang API and can be installed independently from Buildah.
 
 A major difference between Podman and Buildah is their concept of a container.  Podman
 allows users to create "traditional containers" where the intent of these containers is


README.md
```diff
@@ -51,7 +51,8 @@ build images.  The flexibility of building images without Dockerfiles allows
 for the integration of other scripting languages into the build process.
 Buildah follows a simple fork-exec model and does not run as a daemon
 but it is based on a comprehensive API in golang, which can be vendored
-into other tools.
+into other tools.  Podman uses Buildah's golang API for the podman-build
```
TomSweeneyRedHat: I think I'd put this new verbiage at the end of the following paragraph.  WDYT?

TomSweeneyRedHat:
One question for consideration.  We'll need the final corresponding change in the Podman README.md too.  I've restarted the one flakey test.

README.md
vrothberg: I put it here to have it close to the explanation of Buildah also being/exposing an golang API. But I can move it down :+1:.

rhatdan: It is also vendored into the OpenShift Image Builder.

rhatdan: How about 

For example,  Podman uses Buildah's golang API for the podman-build

TomSweeneyRedHat: just noted, missing a '.' at the end.  Rather than "podman-bulid", I think I'd go 
```
`podman build` command.
```

vrothberg:
@rhatdan @TomSweeneyRedHat, I moved it one paragraph down and rephrased to be more explicit.

If you're cool, we can just force merge to avoid running the entire CI.

rhatdan:
LGTM

PR #1803, opened 2019-08-19
change --pull to be consistent with podman

QiWang19:
change --pull for bud and from to be consistent with podman change https://github.com/containers/libpod/pull/3617

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -96,12 +96,21 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budResults) error {
 			tags = tags[1:]
 		}
 	}
+
+	pullType, err := buildahcli.ValidatePullType(iopts.Pull)
+	if err != nil {
+		return err
+	}
+
 	pullPolicy := imagebuildah.PullNever
-	if iopts.Pull {
+	if pullType == buildahcli.PullImageMissing {
 		pullPolicy = imagebuildah.PullIfMissing
 	}
-	if iopts.PullAlways {
+	if pullType == buildahcli.PullImageAlways || iopts.PullAlways {
 		pullPolicy = imagebuildah.PullAlways
+	}
+	if iopts.PullAlways && pullType != buildahcli.PullImageAlways {
+		return errors.Errorf("--pull-ways does not match pull type from --pull %q", iopts.Pull)
 	}
 
 	args := make(map[string]string)
--- a/cmd/buildah/from.go
+++ b/cmd/buildah/from.go
@@ -21,7 +21,7 @@ type fromReply struct {
 	creds           string
 	format          string
 	name            string
-	pull            bool
+	pull            string
 	pullAlways      bool
 	quiet           bool
 	signaturePolicy string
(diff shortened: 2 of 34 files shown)


cmd/buildah/bud.go
```diff
@@ -96,10 +96,20 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budResults) error {
...
+	if err != nil {
+		return err
+	}
+
 	pullPolicy := imagebuildah.PullNever
-	if iopts.Pull {
+	if pullType == buildahcli.PullImageMissing {
 		pullPolicy = imagebuildah.PullIfMissing
 	}
+	if pullType == buildahcli.PullImageAlways {
```
rhatdan: Combine these together with the next section

if pullType == buildahcli.PullImageAlways || iopts.PullAlways {
		pullPolicy = imagebuildah.PullAlways
	}

rhatdan: Please hide the pullalways code and remove it from man pages.
Also throw an error if the user says to --pull-always and --pull does not match.

cmd/buildah/from.go
```diff
@@ -21,7 +21,7 @@ type fromReply struct {
 	creds           string
 	format          string
 	name            string
-	pull            bool
+	pull            string
```
rhatdan: This is a breaking change.
If I previously had the script below, it will now break.

docs/buildah-bud.md
```diff
@@ -339,16 +339,12 @@ not required for Buildah as it supports only Linux.
 
 **--pull**
 
-When the flag is enabled, attempt to pull the latest image from the registries
-listed in registries.conf if a local image does not exist or the image is newer
-than the one in storage. Raise an error if the image is not in any listed
-registry and is not present locally.
+Pull image before building ("always"|"missing"|"never") (default "missing").
```
rhatdan: This is a change.  The --pull used to mean to look at the registry and pull if newer.  Now that is not an option, as I understand it.
Should their be another option "newer", which should be default and pull a newer image if it exists at the registy otherwise use local.

QiWang19: If current boolean --pull also for pulling the newer image, I need to figure out to prevent buildah from pulling new image when `--pull missing` the image exists locally but has a newer version in the registry.

QiWang19: Does buildah  `--pull` can pull the newer image and replace the old one in the local? I tried the master buildah didn't see the update and didn't find the code implemented this functionality.
@TomSweeneyRedHat Do you know which PR I can refer to for finding newer version image from registry?

TomSweeneyRedHat: `buildah bud --pull` is supposed to pull the image if it doesn't exist locally or if the version on the registry is newer there than the version stored locally.  I've not tested it recently to verify that it does do that.

Where it happens is fun to find.

https://github.com/containers/buildah/blob/master/new.go#L100 is where resolveImage() is called and the check happens here:  https://github.com/containers/buildah/blob/master/new.go#L140

Or at least that's the theory.

QiWang19: thanks, @TomSweeneyRedHat but not code there, that code is for `--pull-always.`
if bud --pull, buildah will set options.PullPolicy = PullIfMissing, but https://github.com/containers/buildah/blob/master/new.go#L140 is inside the `if options.PullPolicy == PullAlways`

rhatdan:
@QiWang19 is this still being worked on, or has it been supplanted by @TomSweeneyRedHat PR https://github.com/containers/buildah/pull/1873

QiWang19:
I'm not working on this. This PR change --pull from bool to string. This is a breaking change and  I don't know  how to make --pull work both as boolean and string.

rhatdan:
Lets get #1873 in and then we can discuss this.

PR #1804, opened 2019-08-20
Update shebangs to take env into consideration

saschagrunert:
Some operating systems don’t have `bash` in `/bin`, so we should take the
`$PATH` into consideration for searching it.


TomSweeneyRedHat:
nice catch and change @saschagrunert TYVM!

TomSweeneyRedHat:
LGTM
@rh-atomic-bot r+

PR #1807, opened 2019-08-21
Remove govet.sh in favor of golangci-lint

saschagrunert:
golangci-lint already checks via vet so we can safely remove the
additional linting step.


rhatdan:
LGTM
@TomSweeneyRedHat @vrothberg @giuseppe @nalind PTAL

vrothberg:
@rh-atomic-bot r+

PR #1809, opened 2019-08-21
Remove outdated deps Makefile target

saschagrunert:
We don’t use vendor any more so we can remove the `deps` Makefile
target.


rhatdan:
LGTM
@nalind @TomSweeneyRedHat @vrothberg @giuseppe @QiWang19 PTAL

vrothberg:
@rh-atomic-bot r+

PR #1812, opened 2019-08-22
Disable empty logrus timestamps to reduce logger noise

saschagrunert:
The logrus timestmaps are always empty where we can disable them to have
cleaner log outputs:

Before:
```
DEBU[0000] [graphdriver] trying provided driver "overlay"
```

After:
```
DEBU [graphdriver] trying provided driver "overlay"
```

### DIFF
--- a/cmd/buildah/main.go
+++ b/cmd/buildah/main.go
@@ -47,6 +47,7 @@ var (
 )
 
 func init() {
+	logrus.SetFormatter(&logrus.TextFormatter{DisableTimestamp: true})
 
 	var (
 		defaultStoreDriverOptions []string


rhatdan:
LGTM
@nalind @TomSweeneyRedHat @giuseppe @vrothberg @ashley-cui @QiWang19 @TomSweeneyRedHat PTAL

QiWang19:
LGTM

rhatdan:
@rh-atomic-bot r+

TomSweeneyRedHat:
LGTM, thanks @saschagrunert

nalind:
The timestamps are count of elapsed seconds since the command started running.

saschagrunert:
> The timestamps are count of elapsed seconds since the command started running.

Ah, this makes sense. Do we need those for a command line application? I'm not sure but feel free to close if you think they add valuable information to the user.

Issue #1815, opened 2019-08-22
`buildah commit --squash` is using /tmp to build images


oblitum:
**Description**

I have a larger than `/tmp` image that I want to apply `buildah commit --squash` to attempt creating a squashed duplicate, but the command is using `/tmp` for the task, and there's no space left for it (which can be [circumvented by resizing](https://github.com/containers/libpod/issues/3871#issuecomment-524098810) or remounting `/tmp`, e.g. `sudo mount -o remount,size=12G /tmp`).

**Steps to reproduce the issue:**
1. `buildah from larger-than-tmp-image`
2. `buildah commit --squash`

**Describe the results you received:**

```
[I] ➜ buildah commit --squash archlinux-working-container archlinux2
ERRO[0119] Can't add file /home/francisco/.local/share/containers/storage/overlay/693ecb17fb10858eaea0aa7eb529c6f1a12bd8f9d991de7623f5c51371679da0/merged/usr/share/texmf-dist/fonts/tfm/adobe/sourceserifpro/SourceSerifPro-SemiboldIt-tosf-t1.tfm to tar: io: read/write on closed pipe
ERRO[0119] Can't close tar writer: io: read/write on closed pipe
error committing container "archlinux-working-container" to "archlinux2": error copying layers and metadata for container "d3a78dc6abc9fbb00bb720dfe6c5320b469f88916e93240c95269412d4e920fc": Error initializing source containers-storage:archlinux-working-container: error storing container "d3a78dc6abc9fbb00bb720dfe6c5320b469f88916e93240c95269412d4e920fc" to file: write /tmp/buildah262364539/layer: no space left on device
ERRO[0120] exit status 1
```

**Describe the results you expected:**

I'm fine that it uses `/tmp` by default, not sure whether it's actually good idea though. I'd like to at least have/know an option to set a different location. Or the command could be smart enough to check the image to be committed is larger than the default `/tmp` mount point, so it should use another one or cancel before trying.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
[I] ➜ pacman -Qi buildah
Name            : buildah
Version         : 1.10.1-1
Description     : A tool which facilitates building OCI images
Architecture    : x86_64
URL             : https://github.com/containers/buildah
Licenses        : Apache
Groups          : None
Provides        : None
Depends On      : ostree  runc  skopeo
Optional Deps   : None
Required By     : None
Optional For    : None
Conflicts With  : None
Replaces        : None
[...]

PR #1838, opened 2019-09-04
Set TMPDIR to /var/tmp by default

rhatdan:
We have had some issues with users squashing large images or pulling large
content from github, that could trigger crashes based on the size of /tmp.

Docker had an issue with this back in 2016. https://github.com/golang/go/issues/14021

The discussion there was to change the default to /var/tmp.

This change will only effect systems that do not set the TMPDIR environment variable.

### DIFF
--- a/cmd/buildah/main.go
+++ b/cmd/buildah/main.go
@@ -134,6 +134,10 @@ func main() {
 	if buildah.InitReexec() {
 		return
 	}
+	// Hard code TMPDIR functions to use /var/tmp, if user did not override
+	if _, ok := os.LookupEnv("TMPDIR"); !ok {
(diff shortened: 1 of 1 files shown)


rhatdan:
Should fix https://github.com/containers/buildah/issues/1815

rhatdan:
@nalind @vrothberg @giuseppe PTAL

nalind:
This won't affect the image library's storage driver, which uses `github.com/containers/image/internal/tmpdir.TemporaryDirectoryForBigFiles()` to decide where to store its temporary files, but that already defaults to `/var/tmp`.  LGTM.

rhatdan:
Ignoring Darwin issues, I am manually merging.

PR #1839, opened 2019-09-04
makeImageRef: ignore EmptyLayer if Squash is set

nalind:
If `Squash` is set, then we're only committing one layer, so if `EmptyLayer` is set, which would normally tell us to skip committing the final layer, we should ignore the `EmptyLayer` flag.

This should fix https://bugzilla.redhat.com/show_bug.cgi?id=1746022.


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM.
Tests out well and a pretty slick little fix @nalind!

TomSweeneyRedHat:
@rh-atomic-bot r+

PR #1841, opened 2019-09-05
Truncate output of too long image names

saschagrunert:
Image names longer than 32 characters will now be truncated via `...` in
the default table output. For example this:

```
579aa959bb4d     *     9b1a5a0c02f5 docker.io/clearlinux/golang:latest golang-working-container
```

Now looks like this:

```
579aa959bb4d     *     9b1a5a0c02f5 docker.io/clearlinux/golang:l... golang-working-container
```

Signed-off-by: Sascha Grunert <sgrunert@suse.com>

### DIFF
--- a/cmd/buildah/containers.go
+++ b/cmd/buildah/containers.go
@@ -9,6 +9,7 @@ import (
 	"text/template"
 
 	"github.com/containers/buildah"
+	"github.com/containers/buildah/util"
 	"github.com/containers/storage"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -261,7 +262,7 @@ func containerOutputUsingTemplate(format string, params containerOutputParams) e
 
 func containerOutputUsingFormatString(truncate bool, params containerOutputParams) {
 	if truncate {
-		fmt.Printf("%-12.12s  %-8s %-12.12s %-32s %s\n", params.ContainerID, params.Builder, params.ImageID, params.ImageName, params.ContainerName)
+		fmt.Printf("%-12.12s  %-8s %-12.12s %-32s %s\n", params.ContainerID, params.Builder, params.ImageID, util.TruncateString(params.ImageName, 32), params.ContainerName)
 	} else {
 		fmt.Printf("%-64s %-8s %-64s %-32s %s\n", params.ContainerID, params.Builder, params.ImageID, params.ImageName, params.ContainerName)
 	}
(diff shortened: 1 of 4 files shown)


cmd/buildah/containers.go
```diff
@@ -261,12 +261,24 @@ func containerOutputUsingTemplate(format string, params containerOutputParams) e
 
 func containerOutputUsingFormatString(truncate bool, params containerOutputParams) {
 	if truncate {
-		fmt.Printf("%-12.12s  %-8s %-12.12s %-32s %s\n", params.ContainerID, params.Builder, params.ImageID, params.ImageName, params.ContainerName)
+		fmt.Printf("%-12.12s  %-8s %-12.12s %-32s %s\n", params.ContainerID, params.Builder, params.ImageID, truncateString(params.ImageName, 32), params.ContainerName)
 	} else {
-		fmt.Printf("%-64s %-8s %-64s %-32s %s\n", params.ContainerID, params.Builder, params.ImageID, params.ImageName, params.ContainerName)
+		fmt.Printf("%-64s %-8s %-64s %-32s %s\n", params.ContainerID, params.Builder, params.ImageID, truncateString(params.ImageName, 32), params.ContainerName)
```
nalind: Truncating part of the output when `truncate` is false (i.e., the user specified `--notruncate`) seems like a bug.

saschagrunert: I see, I changed that part back.

saschagrunert:
> We need to update docs/buildah-containers.md too.
> 
> ```
> --notruncate
> 
> Do not truncate IDs in output.
> ```
> 
> and s/IDs and image names/IDs/

Fixed as well. :heavy_check_mark:

cmd/buildah/containers.go
```diff
@@ -261,12 +261,24 @@ func containerOutputUsingTemplate(format string, params containerOutputParams) e
...
+	if len(str) > to {
+		const tr = "..."
+		if to > len(tr) {
+			to -= len(tr)
+		}
+		newStr = str[0:to] + tr
+	}
+	return newStr
+}
+
```
TomSweeneyRedHat: @saschagrunert you're too fast for me with these commits!  I'm thinking that this function would probably better live in util/util.go.  It could be used elsewhere at some point in time.  @nalind any disagreement?

nalind: No objection.

rhatdan:
LGTM

cmd/buildah/containers.go
saschagrunert: Okay, I move it over to util/util.go

rhatdan:
@rh-atomic-bot r+

TomSweeneyRedHat:
LGTM, thanks muchly @saschagrunert

PR #1842, opened 2019-09-05
[release-1.8] bump containers/storage and containers/image

nalind:
Bump github.com/containers/storage to v1.13.2 and github.com/containers/image to v3.0.2, and bump github.com/vbauerstr/mpb and github.com/klauspost/compress to versions that the updated image library requires.


util.go
```diff
@@ -280,17 +279,17 @@ func (b *Builder) untar(chownOpts *idtools.IDPair, hasher io.Writer) func(tarArc
 func isRegistryBlocked(registry string, sc *types.SystemContext) (bool, error) {
 	reginfo, err := sysregistriesv2.FindRegistry(sc, registry)
 	if err != nil {
-		return false, errors.Wrapf(err, "unable to parse the registries configuration (%s)", sysregistries.RegistriesConfPath(sc))
+		return false, errors.Wrapf(err, "unable to parse the registries configuration (registries.conf)")
```
mtrmac: `sysregistriesv2.ConfigPath` (throughout)?

new.go
```diff
@@ -185,7 +184,7 @@ func resolveImage(ctx context.Context, systemContext *types.SystemContext, store
 		return nil, "", nil, fmt.Errorf("internal error: %d candidates (%#v) vs. %d failures (%#v)", len(candidates), candidates, len(failures), failures)
 	}
 
-	registriesConfPath := sysregistries.RegistriesConfPath(systemContext)
+	registriesConfPath := "registries.conf"
```
rhatdan: Can't I get this via SystemContext.SystemRegistriesConfPath?

rhatdan: @vrothberg @mtrmac ?

util.go
```diff
@@ -280,17 +279,17 @@ func (b *Builder) untar(chownOpts *idtools.IDPair, hasher io.Writer) func(tarArc
 func isRegistryBlocked(registry string, sc *types.SystemContext) (bool, error) {
```
mtrmac: (Absolutely non-blocking:

BTW the three calls of `isReferenceBlocked` (but not the check in `util.ResolveName`) could probably be completely deleted, c/image now enforced the `.Blocked` option automatically.

Note, though:
- The check happens a bit later, which might result in an user-observable behavior change
- c/image enforces the `.Blocked` option on the _results_ of mirror resolution, not on the _input_ to it; i.e. it is possible to have `example.com` mirrored to `foo.local`, with `example.com` blocked and `foo.local` non-blocked; then references to `example.com/…` are allowed but must only physically contact `foo.local`, not access the primary `example.com` server directly.)

new.go
nalind: I expect that field to not be set most of the time.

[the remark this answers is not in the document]
nalind: Yeah, that's a better way to go.

rhatdan:
LGTM
@rh-atomic-bot r+

TomSweeneyRedHat:
@nalind, looks like you've some gofmt errors.

TomSweeneyRedHat:
Changes LGTM otherwise.

rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot retry

rhatdan:
Not sure why this is hung up.  Merging Manually.

PR #1848, opened 2019-09-09
Cirrus: Migrate off of PAPR + Improve testing

cevich:
Replace PAPR with 'in_podman' / Cirrus-CI based test.  Prepare the groundwork for replacing Homu and Travis with Bors + VM-based testing (set "optional" in this PR).   Once this PR is merged, until #2115 is merged: maintainers must carefully/manually merge PRs one-by-one (since the Homu bot will be disabled).


rhatdan:
Thanks @cevich looking forward to this working. and I am sure @jlebon is looking forward to it also.  :^)

cevich:
note to me:   somehow things running in podman are not returning exit codes?

cevich:
@nalind or @rhatdan I need some help here.  For buildah-in-podman (super-privileged as in PAPR) intg. testing in F30 I'm missing something (a CLI option or sysctl or something on the host).  How do I prevent https://cirrus-ci.com/task/6256151280222208?command=integration_test#L3274 ?

(all other tests mostly seem to be running well)

I'm running the container with various values for `$BUILDAH_ISOLATION` from here:  https://github.com/containers/buildah/pull/1848/files#diff-a724b9731c5e35d731b550c9714775e1R189-R206

rhatdan:
This looks like the cgroup file system is being mounted readonly into the container that is running podman, or on the host?

cevich:
I've looked and looked but I cannot see where it's mounting up cgroups anything into the container.  The host is a F30 VM, not too different from what we use in libpod.  Searching,  I find is that `-e "CGROUP_MANAGER=cgroupfs"` option from the `in_podman` function I linked.  Could that be triggering test-behavior that shouldn't be enabled?

rhatdan:
Can you ssh into the box?

cevich:
Yes, we have the same `hack/get_ci_vm.sh` script here and ability to 'delete-protect' VMs.  I agree, that seems a reasonable next step.  The other thought I had (though a bit undesirable) is to temporarily skip the offenders to support the migration.  Then pick up immediately on fixing them.  Either way, lemmie grab one of these fetched-up VMs so we can poke it with sharpened sticks...

cevich:
Oh...and this is probably an easy answer:  For F30 w/ `BUILDAH_ISOLATION=chroot STORAGE_DRIVER=vfs STORAGE_OPT=vfs.mountopt=nodev` I see this for a few tests:

```
[+1642s] # time="2019-10-23T17:00:15Z" level=debug msg="[graphdriver] trying provided driver \"vfs\""
[+1642s] # time="2019-10-23T17:00:15Z" level=error msg="error opening storage: vfs driver does not support overlay.mountopt options"
```

That seems like a bug to me (either in test, or elsewhere), eh?

cevich:
Ref, the important beginning parts:

```
[root@cevich-fedora-cloud-base-30-1-2-1565360543 buildah]# contrib/cirrus/test.sh integration
rm -f -r buildah imgtype build buildah.static buildah.darwin tests/testreport/testreport
make -C docs clean
make[1]: Entering directory '/var/tmp/go/src/github.com/containers/buildah/docs'
rm -f buildah*.1
make[1]: Leaving directory '/var/tmp/go/src/github.com/containers/buildah/docs'
--------------------------------------------------
+  podman run -i --name in_podman_build24 -e CIRRUS_WORKING_DIR=/var/tmp/go/src/github.com/containers/buildah -e CI=false --net=host --security-opt label=disable --security-opt seccomp=unconfined --cap-add=all -e GOPATH=/var/tmp/go -e IN_PODMAN=false -e DIST=fedora -e CGROUP_MANAGER=cgroupfs -e STORAGE_DRIVER=vfs -e STORAGE_OPT=vfs.mountopt=nodev --network container:registry -v /dev/fuse:/dev/fuse -v /var/tmp/go/src/github.com/containers/buildah:/var/tmp/go/src/github.com/containers/buildah:z --workdir /var/tmp/go/src/github.com/containers/buildah -v /root/auth:/root/auth:ro --rm in_podman_build24:latest contrib/cirrus/test.sh integration
--------------------------------------------------
+  make
GO111MODULE=on go build -mod=vendor -ldflags '-X main.GitCommit=8cbbf6e1 -X main.buildInfo=1571933288 -X main.cniVersion=v0.7.1'  -o buildah -tags "seccomp selinux     containers_image_ostree" ./cmd/buildah
...cut...
```

(confirming it fails in the same way)

cevich:
...okay, yep, it produces the same `/sys/fs/cgroup/cpuset/buildah-buildah323213061: read-only file system`

cevich:
Ref: command used to execute in papr: https://github.com/containers/buildah/blob/master/.papr.yml#L50-L53

cevich:
trying with a `-v /sys/fs/cgroup:/sys/fs/cgroup`

cevich:
Oh...this is interesting.  Now I get errors finding files the definitely exist:

```
not ok 19 blobcache-build-compressed-using-dockerfile-explicit-push
# (from function `die' in file ./helpers.bash, line 128,
#  from function `run_buildah' in file ./helpers.bash, line 116,
#  in test file ./blobcache.bats, line 169)
#   `run_buildah build-using-dockerfile -t ${target} --pull-always --signature-policy ${TESTSDIR}/policy.json --blob-cache=${blobcachedir} --disable-compression=false ${TESTSDIR}/bud/add-file' failed
# $ /var/tmp/go/src/github.com/containers/buildah/tests/./../buildah build-using-dockerfile -t new-image --pull-always --signature-policy /var/tmp/go/src/github.com/containers/buildah/tests/./policy.json --blob-cache=/var/tmp/tmp18055b50e8fac66973ce2248/cache --disable-compression=false /var/tmp/go/src/github.com/containers/buildah/tests/./bud/add-file
# cannot find Containerfile or Dockerfile in context directory: stat /var/tmp/go/src/github.com/containers/buildah/tests/bud/add-file/Dockerfile: no such file or directory
# [ rc=1 (** EXPECTED 0 **) ]
```

(ten-billion of similar messages)

rhatdan:
Remove STORAGE_OPT=vfs.mountopt=nodev
Why are these tests difference then before?

rhatdan:
Where does this podman command come from?  I don't see it under Buildah?

cevich:
it's not SELinux.  I tried both `permissive` and `setsebool container_manage_cgroup true`.

cevich:
> Where does this podman command come from? I don't see it under Buildah?

Ugg, trying and failing to give useful links, sorry.  It comes from `contrib/cirrus/lib.sh` if github would behave for me...

contrib/cirrus/lib.sh
```diff
@@ -151,16 +165,105 @@ install_ooe() {
 }
 
 showrun() {
-    if [[ "$1" == "--background" ]]
+    echo '--------------------------------------------------'
+    echo '+ '$(printf " %q" "$@") > /dev/stderr
+    "$@"
+}
+
+in_podman() {
```
cevich: Here

cevich:
Papr was doing it like:

```
podman run --net=host --security-opt label=disable --cap-add all --security-opt seccomp=unconfined -v /etc/yum.repos.d:/etc/yum.repos.d.host:ro
    -v $PWD:/go/src/github.com/containers/buildah
    --workdir /go/src/github.com/containers/buildah
    registry.fedoraproject.org/fedora:30 bash -c sh ./.papr.sh
```

(which didn't work, so I, err, started adding things :grin: )

cevich:
> Why are these tests difference then before?

@rhatdan I think one of the things we're dealing with is the vast difference in environments and podman.  Papr was last successful running (command above) on a Fedora Atomic 28 host.  So presumably buildah development stuck with what ran and passed there (i.e. using a very old podman).  Now with a F30 host and brand-new-fresh podman, test-vs-host expectations are just not lining up any more.

> Remove STORAGE_OPT=vfs.mountopt=nodev

I'll try again, but IIRC w/o that the tests throw errors complaining about `overlay.mountopts` (despite STORAGE_DRIVER=vfs`)

cevich:
So it seems the `BUILDAH_ISOLATION=chroot` and `rootless` tasks with `STORAGE_OPT=vfs.mountopt=nodev` are what throws

```
[+1602s] # time="2019-10-24T20:30:05Z" level=debug msg="[graphdriver] trying provided driver \"vfs\""
[+1602s] # time="2019-10-24T20:30:05Z" level=error msg="error opening storage: vfs driver does not support overlay.mountopt options"
```

`BUILDAH_ISOLATION=` (empty)  task gives the `/sys/fs/cgroup/cpuset/buildah-buildah324011388: read-only file system\`  cgroup errors. 

There are some other failures mixed in which I'm ignoring for now.  In any case, based on the huge number of tests that do pass, overall this PR is closer than it was on Monday.

cevich:
I think maybe the next step is to see if I can manually reproduce the mount-options thing...perhaps there's a bug there.

cevich:
@rhatdan I can reproduce the read-only cgroup directory thing locally (F29) using a simplified version of the container used in these tests:

```
# podman run -it --rm --net=host --security-opt label=disable --security-opt seccomp=unconfined --cap-add=all -e CGROUP_MANAGER=cgroupfs -e STORAGE_DRIVER=vfs -v /dev/fuse:/dev/fuse fedora:latest mkdir /sys/fs/cgroup/cpuset/buildah-buildah364534981

mkdir: cannot create directory '/sys/fs/cgroup/cpuset/buildah-buildah364534981': Read-only file system
```

I produce the same error using a simplified version of the former PAPR container command:

```
# podman run -it --rm --net=host --security-opt label=disable --cap-add all --security-opt seccomp=unconfined fedora:latest mkdir /sys/fs/cgroup/cpuset/buildah-buildah364534981

mkdir: cannot create directory '/sys/fs/cgroup/cpuset/buildah-buildah364534981': Read-only file system
```

and for sanity-sake, the `mkdir` does work directly on my host:

```
# mkdir /sys/fs/cgroup/cpuset/buildah-buildah364534981
# rmdir /sys/fs/cgroup/cpuset/buildah-buildah364534981
```

cevich:
@rhatdan I chatted with Matt about this.  We couldn't get any definitive answers, but he seems to believe it's possible we had a read-write cgroups in podman during the F28 timeframe.  Since these containers are running with super-privileges anyway, he confirmed our best bet is to just volume-mount `/sys/fs/cgroup` read-write in the containers.

I suppose the only other choice here is to dive into a bunch of old tests and re-write as needed.  I'm not prepared for that nor would I feel good asking someone else to do it :face_with_head_bandage:

cevich:
(Rebased + added ***rw*** mounting of cgroups, enabling `container_manage_cgroup`, plus added a few comments)

Hopefully fixing the cgroups problem will leave only the `overlay.mountopts` problem to address...

cevich:
@rhatdan the `overlay.mountopt` issue I think is an actual bug.  I'm able to reproduce it easily:

```
$ sudo podman run --net=host --security-opt label=disable --security-opt seccomp=unconfined --cap-add=all -e CGROUP_MANAGER=cgroupfs -e STORAGE_DRIVER=vfs -v /sys/fs/cgroup:/sys/fs/cgroup:rw -it --rm fedora:latest bash

[root@cevich ~]# dnf install -y buildah
...cut...
 buildah                        x86_64        1.11.3-2.fc30                         updates        8.6 M
...cut...
Complete!
[root@cevich ~]# buildah from scratch
vfs driver does not support overlay.mountopt options
[root@cevich ~]# buildah --storage-driver=vfs from scratch
working-container
```

IIRC we had an issue similar to this in podman, do you remember the details?

rhatdan:
I have a PR to begin fixing this in the works, actually merged into container/storage. 
The issue is we do not clear the STORAGE_OPTS when STORAGE_DRIVER is set, we do if the user overrides via command line.

You need to add -e STORAGE_OPTIONS="" and I think everything will work.

cevich:
Here's the issue: https://github.com/containers/buildah/issues/1783 and https://github.com/containers/libpod/pull/3240 (in podman)

So should I be setting `STORAGE_OPT=overlay.mountopt=`?  That just seems wrong (we're not using 'overlay', we're using 'vfs')

rhatdan:
Won't STORAGE_OPTIONS="" work?

cevich:
> Won't STORAGE_OPTIONS="" work?

Nope, same deal.  I'm convinced this is simply buildah's flavor of containers/libpod#3240

The quick-fix there was to comment-out `mountopt = "nodev,metacopy=on"` in `/etc/containers/storage.conf`.  

Opened #1945

cevich:
(force-pushed w/ quick-fix to `mountopt` issue + squashed up commits)

cevich:
(rebased w/ addition of /dev/fuse volume-mount to podman command-line)

cevich:
Ooohhh, it seems so close I can feel it.  @rhatdan @nalind these last errors seem like they would be easy to debug given your better knowledge of the tests.  Mind you, these are all running inside podman, and I see repetition of some common errors throughout (not sure if they're significant):

* For [BUILDAH_ISOLATION=""](https://cirrus-ci.com/task/6284275850149888)
   * [[+1066s] not ok 135 buidah bud --volume](https://cirrus-ci.com/task/6284275850149888?command=integration_test#L891-L929)
   * [[+1144s] not ok 155 bud with additional device](https://cirrus-ci.com/task/6284275850149888?command=integration_test#L949-L1050)
   * [[+1810s] not ok 280 user-and-network-namespace](https://cirrus-ci.com/task/6284275850149888?command=integration_test#L1175-L1197)
   * [[+1836s] not ok 286 user-namespace](https://cirrus-ci.com/task/6284275850149888?command=integration_test#L1203-L1220)
   * [[+1856s] not ok 288 combination-namespaces](https://cirrus-ci.com/task/6284275850149888?command=integration_test#L1222-L1435)
   * [[+1861s] not ok 290 overlay specific level](https://cirrus-ci.com/task/6284275850149888?command=integration_test#L1437-L1450)
   * [[+2206s] not ok 343 run-device](https://cirrus-ci.com/task/6284275850149888?command=integration_test#L1503-L1692)
* For [BUILDAH_ISOLATION=chroot](https://cirrus-ci.com/task/4876900966596608)
   * ***SAME/SIMILAR*** [+1105s] not ok 135 buidah bud --volume
   * [[+1809s] not ok 290 overlay specific level](https://cirrus-ci.com/task/4876900966596608?command=integration_test#L1084-L1094)
* For [BUILDAH_ISOLATION=rootless](https://cirrus-ci.com/task/6002800873439232)
   * ...over 10 failures...appears like local test-registry failed.  Re-running...

cevich:
Re-run of [BUILDAH_ISOLATION=rootless](https://cirrus-ci.com/task/4833699786391552) the error I see repeating frequently across many tests is:

```
[+0809s] # error running container: error creating container for [/bin/sh -c echo this-should-be-cached-but-it-s-not]: time="2019-10-28T14:50:44Z" level=warning msg="exit status 1"
[+0809s] # time="2019-10-28T14:50:44Z" level=error msg="container_linux.go:346: starting container process caused \"process_linux.go:446: container init caused \\\"rootfs_linux.go:58: mounting \\\\\\\"/proc\\\\\\\" to rootfs \\\\\\\"/var/tmp/buildah073351768/mnt/rootfs\\\\\\\" at \\\\\\\"/proc\\\\\\\" caused \\\\\\\"operation not permitted\\\\\\\"\\\"\"\n"
[+0809s] # container_linux.go:346: starting container process caused "process_linux.go:446: container init caused \"rootfs_linux.go:58: mounting \\\"/proc\\\" to rootfs \\\"/var/tmp/buildah073351768/mnt/rootfs\\\" at \\\"/proc\\\" caused \\\"operation not permitted\\\"\""
[+0809s] # : exit status 1
[+0809s] # level=debug msg="error building at step {Env:[container=docker PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin  container=docker] Command:run Args:[echo this-should-be-cached-but-it-s-not] Flags:[] Attrs:map[] Message:RUN echo this-should-be-cached-but-it-s-not Original:RUN echo this-should-be-cached-but-it-s-not}: error while running runtime: exit status 1"
[+0809s] # error building at STEP "RUN echo this-should-be-cached-but-it-s-not": error while running runtime: exit status 1
[+0809s] # [ rc=1 (** EXPECTED 0 **) ]
[+0809s] # #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
[+0809s] # #| FAIL: exit code is 1; expected 0
[+0809s] # #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

That seems like maybe a bug in the tests or buildah, if it really is trying to (simulate?) rootless mode.

rhatdan:
Looks like the user namespace is not being setup.

cevich:
@edsantiago you mentioned that you were looking over some buildah BATs tests or results.  It's likely a long-shot, but sometimes I get lucky:  Are any of the failures above familiar or one's you've seen in the past maybe?

edsantiago:
@cevich sorry, I haven't seen any of those errors before. Nor can I figure out what they mean.

cevich:
> I have a PR to begin fixing this in the works, actually merged into container/storage.
> The issue is we do not clear the STORAGE_OPTS when STORAGE_DRIVER is set, we do if the
> user overrides via command line.
>
> You need to add -e STORAGE_OPTIONS="" and I think everything will work.

Woops, missed this comment.  Thanks @rhatdan  I'll add that to the podman command line.

cevich:
no worries @edsantiago thanks for peeking.

cevich:
(rebased on master, added a more helpful 'showrun()', new `STORAGE_OPTIONS` var,  and RW `/proc` volume-mount)

cevich:
(removed volume-mount of `/proc` as it caused bad-things(tm) to happen)

cevich:
Ahh hmm, okay, so we can pass in_podman if we only test the BUILDAH_ISOLATION=chroot case (I commented out the other ones).  Let me enable the regular VM testing now and see what happens...

cevich:
@nalind @rhatdan good news (y'all have been workin' hard) - I'm down to ---> ONE <--- failing buildah-in-podman test.  It appears to be due to a missing `xz` package which I just added.

I've also got updates in this PR to add regular multi-VM testing for the Fedoras and Ubuntus (replacing Travis), but last I looked those aren't working well.  Since this PR is mainly about migrating off PAPR, would you prefer that I temporarily disable those extra tests so I/We/You can get then passing in a separate PR?

rhatdan:
Sure lets get this in and then work on those separately.

cevich:
okay, let me clean up the commits to make turning them on/off easier...

cevich:
@rhatdan @nalind To be transparent: The gate test (specifically _validate_) seems to require the PR to be up-to-date with master, or fails without a helpful error message.  I tried my best to troubleshoot this, but the root cause appears as a complex interaction between several scripts, the CI system(s), and Makefile.  I'd like to make a concerted effort to fix this, but am also juggling other competing priority work :confused:  (I may not get to it before others start complaining)

rhatdan:
LGTM
@nalind @edsantiago @giuseppe @vrothberg @TomSweeneyRedHat PTAL

rhatdan:
@nalind @edsantiago @giuseppe @vrothberg @TomSweeneyRedHat PTAL

TomSweeneyRedHat:
LGTM, but I'm trusting @cevich 's knowledge quite a lot.  
@jlebon FYI for you and if you've a minute or three, would love an eyeball from you.

jlebon:
I'm not very familiar with Cirrus CI but, it looks sane to me offhand. Nice work on this @cevich!

contrib/cirrus/lib.sh
```diff
@@ -151,16 +165,118 @@ install_ooe() {
 }
 
 showrun() {
-    if [[ "$1" == "--background" ]]
+    local context
+    context=($(caller 0))
+    echo '+ '$(printf " %q" "$@")"  # ${context[2]}:${context[0]} in ${context[1]}()" > /dev/stderr
```
edsantiago: These echo-printfs make my head hurt; this one in particular is really confusing. Is there a specific reason for not using simple printf?
```bash
    printf "+ %q # %s:%s in %s()\n" "$@" ${context[2]} ${context[0]} ${context[1]} >/dev/stderr
```
(Likewise for all other echo-printf instances throughout).

Also btw you probably want `local -a` on line 168

contrib/cirrus/lib.sh
```diff
@@ -151,16 +165,118 @@ install_ooe() {
...
+                   -e "GOPATH=$GOPATH" \
+                   -e "IN_PODMAN=false" \
+                   -e "DIST=$OS_RELEASE_ID" \
+                   -e "CGROUP_MANAGER=cgroupfs" \
+                   -v "$GOSRC:$GOSRC:z" \
+                   --workdir "$GOSRC" \
+                   -v "$HOME/auth:$HOME/auth:ro" \
+                   -v /sys/fs/cgroup:/sys/fs/cgroup:rw \
+                   -v /dev/fuse:/dev/fuse:rw \
+                   $@
```
edsantiago: Careful! `$@` is magic, and must always be enclosed in double quotes: `"$@"`

contrib/cirrus/lib.sh
```diff
@@ -151,16 +165,118 @@ install_ooe() {
...
+}
+
+in_podman() {
+    req_env_var IN_PODMAN_NAME GOSRC HOME OS_RELEASE_ID
+    [[ -n "$@" ]] || \
+        die 7 "Must specify FQIN and command with arguments to execute"
+    local envargs
+    local envname
+    local envvalue
+    local envrx='(^CIRRUS_.+)|(^BUILDAH_+)|(^STOTAGE_)|(^CI$)|(^CROSS_TARGET$)|(^IN_PODMAN_.+)'
```
edsantiago: `STOTAGE` looks like a typo

contrib/cirrus/lib.sh
```diff
@@ -151,16 +165,118 @@ install_ooe() {
...
+                   $envargs \
+                   --net=host \
+                   --net="container:registry" \
+                   --security-opt label=disable \
+                   --security-opt seccomp=unconfined \
+                   --cap-add=all \
+                   -e "GOPATH=$GOPATH" \
+                   -e "IN_PODMAN=false" \
+                   -e "DIST=$OS_RELEASE_ID" \
+                   -e "CGROUP_MANAGER=cgroupfs" \
```
edsantiago: I'm confused by this. Did you mean `--cgroup-manager=cgroupfs`? Passing the envariable implies that the podman container will itself be running nested podman commands (I can't tell if that's the case or not).

contrib/cirrus/lib.sh
```diff
@@ -25,25 +28,26 @@ then
...
 
 export CI="${CI:-false}"
 CIRRUS_CI="${CIRRUS_CI:-false}"
 CONTINUOUS_INTEGRATION="${CONTINUOUS_INTEGRATION:-false}"
 CIRRUS_REPO_NAME=${CIRRUS_REPO_NAME:-buildah}
-CIRRUS_BASE_SHA=${CIRRUS_BASE_SHA:-unknown$(date +%s)}  # difficult to reliably discover
-CIRRUS_BUILD_ID=${CIRRUS_BUILD_ID:-$RANDOM$(date +%s)}  # must be short and unique
+CIRRUS_BASE_SHA=${CIRRUS_BASE_SHA:-unknown$(date +%d)}  # difficult to reliably discover
+CIRRUS_BUILD_ID=${CIRRUS_BUILD_ID:-unknown$(date +%s)}  # must be short and unique
+CIRRUS_TASK_ID=${CIRRUS_BUILD_ID:-unknown$(date +%d)}
```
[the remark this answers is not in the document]
cevich: IIRC, this change helps prevent thrashing behavior when using `hack/get_ci_vm.sh`.  Yes, this would be a good thing to include in the comment.  Thanks for the suggestion.

contrib/cirrus/lib.sh
```diff
@@ -55,15 +59,25 @@ OS_RELEASE_VER="$(source /etc/os-release; echo $VERSION_ID | cut -d '.' -f 1)"
...
 # Avoid some ways of getting stuck waiting for user input
 export DEBIAN_FRONTEND=noninteractive
 # Short-cut for frequently used base command
 export APTGET='apt-get -qq --yes'
-# Short list of packages or quick-running command
+# Short timeout for quick-running packaging command
 SHORT_APTGET="timeout_attempt_delay_command 24s 5 30s $APTGET"
-# Long list / long-running command
+SHORT_DNFY="timeout_attempt_delay_command 60s 2 5s dnf -y"
+# Short timeout for quick-running packaging command
```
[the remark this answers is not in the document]
cevich: Good catch.

contrib/cirrus/lib.sh
cevich: Yeah there was a reason, but damned if I can remember even if it was even a good one :smile:   Copy-pasting your suggestions above.  I confirmed `local -a` works on ubuntu 18's older bash.

contrib/cirrus/lib.sh
cevich: IIRC, it was giving me trouble when I did that but perhaps for some other reason.  I'm going to leave this one as-is since this function is so critical.  I'll come back and try quoting it if there's time and the thing runs fine after all the other changes.

contrib/cirrus/lib.sh
cevich: Ooohhh, excellent catch.  It is indeed a literal typo.

contrib/cirrus/lib.sh
cevich: Yes, it's running nested buildah commands, despite I believe this particular env. var. is picked up and used by the bats tests.  I'm going to leave this as-is for now since I don't recall having an execution problem in this area.

cevich:
Okay, this is almost ready (ignoring the travis failure since we're just replacing papr here).  Trying Ed's other suggestions...

contrib/cirrus/lib.sh
cevich: Ya, this is needed as-is.

contrib/cirrus/lib.sh
cevich: okay it passed, trying with the quotes...

TomSweeneyRedHat:
I'm not sure if the testing-in-podman error is expected?  It's rather generic.

cevich:
> I'm not sure if the testing-in-podman error is expected? It's rather generic.

No, it wasn't expected - due to networking timeout.  I fixed it.  The other items here (besides Travis) are mostly expected.  Taking a peek shortly to confirm...

cevich:
...Travis was also a networking timeout problem, re-running that.

cevich:
Okay, we good to go here.

rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot retry

cevich:
@rh-atomic-bot retry

rhatdan:
I wonder if removing Papr is causing homu to blow up?  Might need to force commit.

jlebon:
Right, Homu watches for the `required` context, which [PAPR can set](https://github.com/projectatomic/papr/blob/c13f38f32448cd31d858b52704b5715104ac08a0/docs/sample.papr.yml#L106).

Not sure if there's a way to replicate this semantic with Cirrus CI, though if you just want to "merge when all the checks are green", it might be easier to move over to Prow like libpod did.

cevich:
> Not sure if there's a way to replicate this semantic with Cirrus CI

Cirrus touches the new Status API, so a workaround would be to add add a final task to touch `required`, then I think Homu could work.  This isn't a small amount of effort though, and all solutions in this space require giving a bot account repository-write access (somewhere in the chain).  My preference would be to use Github-Actions, but that's only because I don't/can't understand how prow works and native solutions are almost always better than third-parties.

In any case, I think this is work for a follow-up PR, so we should either ignore or disable Homu for now.

rhatdan:
@cevich so you are saying to manually merge?  Will this break humu for all future PRs?

cevich:
> Will this break humu for all future PRs?

Yes until fixed or worked around and/or I understand more of Homu's configuration.  Would you prefer that I address/include a fix/workaround in this PR?  (instead of a separate PR)

Details:

Cirrus-CI will automatically run the tests on the target branch after a PR merges.  This is (for example, in libpod CI) how a notification is sent to `#podman` on IRC when post-merge tests fail.  Homu mitigates this post-merge test-failure possibility a different way, since it controls the merging of all PRs, it can first make sure the tests pass after rebasing from the destination branch.

The down-side of this approach is, since Homu is the gatekeeper, if it ever breaks, nothing gets in.  It's also (necessarily) serializing all merge operations so there's a performance impact on a busy repo (like libpod) with a rapidly moving branch.  I can poke around and see if there's a github-actions way of doing this, that may be a simpler solution.

cevich:
Ref: Good explanation of 'evergreen builds': https://github.com/bors-ng/bors-ng#but-dont-githubs-protected-branches-already-do-this

Even with Prow in charge, post-merge testing still fails on occasion, so it's not a perfect solution.  

Assuming this repo. is much lower traffic than libpod, another idea is:  We turn on the 'must be up-to-date' setting for PRs.  That forces them to always merge serially, by placing the rebase effort on PR authors (which is why it's bad for a busy repo)

rhatdan:
Does the must be up2date mean that everyone needs to rebase or can anyone press a button on github to refire the tests?

cevich:
> Does the must be up2date mean that everyone needs to rebase

Yes, it is that annoying.  One PR merges and it forces all others to have to rebase/repush (and so, re-test) before merging.  IMHO, it's not too horrible on a low-traffic repo., this one seems close to the edge of too-much.  It's easy to turn on/off though w/o much consequence, so low-cost to try it.

I poked around some more today at other solutions, but can't even really see a way to do this 100% in native github-actions - they all run async/parallel by nature and so need a third-party "arbitrator" to (effectively) serialize the stream of merging PRs and handle possible test-failures.

I really wish there was a simple solution (even a half-way "good enough" one).  I'm not above kludging Cirrus to behave with Homu - except for not wanting to burden jlebon any more than necessary (including ongoing maintenance).  So I'm willing to try poking at the Prow dragon again, see if I can make sense of what's needed to hook it up.  I'm also willing to keep looking and see if I can learn more / discover other possibilities.

cevich:
Hmmm, there does appear to be a public-repo supporting github-app for bors-ng (bors is what Homu is based on): https://github.com/apps/bors

It's a free-for-opensource github-app that "someone else" manages the server/database stuff.  I have a mind to look into that a bit more, since the docs suggest it plays nice with github-actions (and therefor cirrus-ci).

cevich:
Hrmmm, I really like what I'm seeing with https://bors.tech/, it's a much larger and active community, hosts a free service for OSS projects, and supports all the new APIs.  Most importantly WRT Cirrus-CI, they support wild-card matching status names.  So for example, we could make these required:

```testing image:fedora-*```
```testing image:ubuntu-18*```

but ignore the status of others (e.g. experimental new tests).  Let me do some more reading/playing, perhaps it could replace Homu here and knowing how it works would let me rig up other repos as well...

**Comment by cevich** [comment]
bors ping

**Comment by cevich** [comment]
bors try

**Comment by cevich** [comment]
bors try

**Comment by cevich** [comment]
bors try

**Comment by cevich** [comment]
bors try

**Comment by cevich** [comment]
@rhatdan status: I've enabled the Bors-ng app on this repo, and added a commit to this PR to configure it initially.  I'm using the 'bors try' command to verify all the scripts and tests will run successfully.  Assuming I can get it working and after this PR merges, we can disable Homu and [start using the new bors commands.](https://bors.tech/documentation/)

**Comment by cevich** [comment]
bors try-

**Comment by cevich** [comment]
bors try

**Comment by rhatdan** [comment]
SGTM

**Comment by cevich** [comment]
bors try

**Comment by cevich** [comment]
bors try

**Comment by cevich** [comment]
@rhatdan okay I think we're good to go here.  I believe if you make a comment saying only 'bors merge' it should do exactly that (after running tests again, and assuming they pass).  If we go forward with bors, that will be the new '/lgtm'.  There are other criteria that can be configured for merging, but I think we can leave them up to future work.

**Comment by rhatdan** [comment]
bors merge

**Comment by TomSweeneyRedHat** [comment]
cevich, once the dust setles, can you put together a little cheat sheet on the new 'bors' commands please?  Maybe add it to contributors.md for posterity?

**Comment by cevich** [comment]
@TomSweeneyRedHat sure, happy to, and the setup needs to be better documented as well as how to get to the settings page on the server-side, where permissions are defined.

**Comment by cevich** [comment]
All I know is...the 'make validation' breaks like this on occasion, without giving much/any clues as to why :confounded: 

When I last simulated a 'bors merge' (using 'bors try') it worked fine :disappointed:

**Comment by cevich** [comment]
```
...

[+0010s] ./tests/tools/build/golangci-lint run
[+0768s] level=error msg="Deadline exceeded: try increase it by passing --deadline option"
```

Basically it sat there waiting or doing something for over 10-minutes then gave up.

Anyone know how we increase the verbosity coming from `make validate`?

**Comment by rhatdan** [comment]
bors merge

**Comment by rhatdan** [comment]
@vrothberg Any ideas on the timeout? ^^

**Comment by cevich** [comment]
@rhatdan @vrothberg So this is not something I like about bors.  It's waiting a full two-hours for the underlying checks to pass, even if they fail after 20-minutes.  In this specific case it's the gate job failing (I know by clicking the red X mark in:

```
@bors Merge #1848 …                                                 X 932f6d4
```

Looking at the output, it is the `make validate` which is timing out, after about 10-minutes (must be setting somewhere to control this and verbosity?)

I'm looking at `./tests/validate/git-validation.sh`

cevich:
One again for the third time...cleaned up `git-validation.sh` quite a bit, and made it more verbose when running under CI.

cevich:
I think I found + fixed the problem.  The default golangci-lint timeout is ***1 minute*** (no idea how/why that actually times out after about 10-minutes).

cevich:
bors try

cevich:
dangit.  what the heck.

cevich:
oooohhhh! It's travis. ok.  @rhatdan can I make travis __not__ required for merge or is it too early for that?

rhatdan:
As long as we are running the tests on ubuntu and fedora, I am fine with not requiring travis.

cevich:
> As long as we are running the tests on ubuntu and fedora, I am fine with not requiring travis.

Dang, with this PR we're only replacing Papr, so only run buildah-in-podman on Fedora.  The other VM-base testing runs, but fails (needswork).  So maybe not safe to disable travis yet?

cevich:
Looking at the attempt, it appears it was a flake:

https://travis-ci.org/containers/buildah/jobs/638076844#L890

Let's try a 'bors merge' but then you/I will need to keep an eye on that travis run and re-run it if it flakes (before the 2-hour bors timeout).

I think we can only fully get rid of travis after the Cirrus-CI VM-based testing is running and passing.  This PR is _only_ about replacing PAPR and Homu...unless you want me to extend it's purpose (again)?

rhatdan:
bors merge

cevich:
OMG travis is soooo slow even to just pick up the job, it hasn't even started running it yet after an HOUR!

cevich:
finally travis started running a few min ago.  my gosh.

cevich:
:confounded: TRAVIS!!!!!!!!!! :confounded: 

Bors timed out before travis finished b/c it didn't even start running for over an hour after triggered.  wow!

--sigh--

cevich:
bors retry

cevich:
Writing an e-mail about this...

cevich:
@jlebon - how do we disable Homu on this repository (after this PR merges)? 

FWIW- we're planning to implement bors-ng in another PR, where I also need to deal with a Travis problem.

jlebon:
> @jlebon - how do we disable Homu on this repository (after this PR merges)?

I can take care of that.

cevich:
@vrothberg @rhatdan okay, this is good to go, assuming your comfortable w/ manual merges until #2115 is ready (already working on it)

cevich:
rebased

cevich:
> I can take care of that.

@jlebon could you please disable Homu and Papr for buildah?  Thanks, and thanks for all your support all this time.  Hopefully the Bors-ng community will be as responsive :smile:

jlebon:
> @jlebon could you please disable Homu and Papr for buildah?

Done!

> Thanks, and thanks for all your support all this time. Hopefully the Bors-ng community will be as responsive :smile: 

Thanks! We might also use bors for some of our projects eventually. And other projects Prow. One less service to maintain is nice. :)

PR #1866, opened 2019-09-19
Fix travis-ci on forks

nakato:
When attempting to run travis-ci on a fork of this repo, due to the
organisation name changing the CI job will fail with a rather
bewildering error about "no go files"

Looking closely it becomes apparent that the path being used is not
GOROOT/src/github.com/containers/buildah but
GOROOT/src/github.com/FORKNAME/buildah, which results in the inability
to run the tests.

To solve this issue the travis-ci go_import_path can be set to ensure
the checkout path is correct, resulting in the ability to ensure CI will
succeed before making a PR.  As this sets the path to be the same as
the primary repository it has no effect on the primary repository.

https://docs.travis-ci.com/user/languages/go/#go-import-path

Signed-off-by: Sachi King <nakato@nakato.io>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ go:
     - 1.11.x
     - 1.12.x
     - tip
+go_import_path: github.com/containers/buildah
 
 env:
     global:


rhatdan:
bot, add author to whitelist
Thanks @nakato for the PR.
@TomSweeneyRedHat @cevich @nalind @giuseppe @vrothberg PTAL

cevich:
I'm pretty sure this is 100% cool, just want to double-check our scripts don't care...yep, there are some old/disused scripts under `hack/` that would break...but they're not actually used anywhere.

cevich:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #1870, opened 2019-09-24
Add support for retrieving context from stdin "-"

nakato:
Some consumers of the docker command line API expect to be able to pass
a tar or a Dockerfile into the build command via stdin, which causes a
build failure when it hits 'buildah bud' or 'podman build', which calls
buildah.  A good example of this is the `linuxkit pkg build $dir`
command, which will append a `-` to the build command.

To improve compatibility with `docker build` support for `-` is added to
mean "read from stdin".
While a user could pass /dev/stdin or /proc/self/fd/0, or a path to some
other character device, as such tricks are common when attempting to get
programs to read from stdin, `docker build` does not support this, thus
this leaves out a bit of complexity while improving 'docker build' cli
compatibility.

---

I recognize that we'll probably need to perform some iteration on this.

The first issue to consider in my opinion is whether TempDirForURL the correct place for this, and if it is added to TempDirForURL, does it constitute an API change due to now potentially opening stdin if passed '-'.

This was however in my opinion the most logical location to add this.

### DIFF
--- a/imagebuildah/util.go
+++ b/imagebuildah/util.go
@@ -76,7 +76,7 @@ func TempDirForURL(dir, prefix, url string) (name string, subdir string, err err
 	if strings.HasPrefix(url, "git://") || strings.HasSuffix(url, ".git") {
 		err = cloneToDirectory(url, name)
 		if err != nil {
-			if err2 := os.Remove(name); err2 != nil {
+			if err2 := os.RemoveAll(name); err2 != nil {
 				logrus.Debugf("error removing temporary directory %q: %v", name, err2)
 			}
 			return "", "", err
@@ -92,7 +92,7 @@ func TempDirForURL(dir, prefix, url string) (name string, subdir string, err err
 	if strings.HasPrefix(url, "http://") || strings.HasPrefix(url, "https://") {
 		err = downloadToDirectory(url, name)
 		if err != nil {
-			if err2 := os.Remove(name); err2 != nil {
+			if err2 := os.RemoveAll(name); err2 != nil {
 				logrus.Debugf("error removing temporary directory %q: %v", name, err2)
 			}
 			return "", subdir, err
--- a/imagebuildah/util.go
+++ b/imagebuildah/util.go
@@ -1,6 +1,8 @@
 package imagebuildah
 
 import (
+	"bufio"
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -55,7 +57,25 @@ func downloadToDirectory(url, dir string) error {
 	return nil
 }
 
-// TempDirForURL checks if the passed-in string looks like a URL.  If it is,
+func stdinToDirectory(dir string) error {
+	logrus.Debugf("extracting stdin to %q", dir)
+	r := bufio.NewReader(os.Stdin)
+	b, err := ioutil.ReadAll(r)
+	if err != nil {
+		return errors.Wrapf(err, "Failed to read from stdin")
+	}
+	reader := bytes.NewReader(b)
+	if err := chrootarchive.Untar(reader, dir, nil); err != nil {
+		dockerfile := filepath.Join(dir, "Dockerfile")
+		// Assume this is a Dockerfile
+		if err := ioutil.WriteFile(dockerfile, b, 0600); err != nil {
+			return errors.Wrapf(err, "Failed to write bytes to %q", dockerfile)
+		}
+	}
+	return nil
+}
+
+// TempDirForURL checks if the passed-in string looks like a URL or -.  If it is,
 // TempDirForURL creates a temporary directory, arranges for its contents to be
 // the contents of that URL, and returns the temporary directory's path, along
 // with the name of a subdirectory which should be used as the build context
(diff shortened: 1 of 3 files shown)


imagebuildah/util.go
```diff
@@ -99,6 +120,17 @@ func TempDirForURL(dir, prefix, url string) (name string, subdir string, err err
 		}
 		return name, subdir, nil
 	}
+	if url == "-" {
+		err = stdinToDirectory(name)
+		if err != nil {
+			if err2 := os.Remove(name); err2 != nil {
+				logrus.Debugf("error removing temporary directory %q: %v", name, err2)
```
rhatdan: Shouldn't this be removing temporary file?

rhatdan: Or should this be os.RemoveAll(name)

TomSweeneyRedHat:
bot, add author to whitelist

imagebuildah/util.go
```diff
@@ -55,7 +57,25 @@ func downloadToDirectory(url, dir string) error {
...
+	b, err := ioutil.ReadAll(r)
+	if err != nil {
+		return errors.Wrapf(err, "Failed to read from stdin")
+	}
+	reader := bytes.NewReader(b)
+	if err := chrootarchive.Untar(reader, dir, nil); err != nil {
+		dockerfile := filepath.Join(dir, "Dockerfile")
+		// Assume this is a Dockerfile
+		if err := ioutil.WriteFile(dockerfile, b, 0600); err != nil {
+			return errors.Wrapf(err, "Failed to write butes to %q", dockerfile)
```
TomSweeneyRedHat: butes?

TomSweeneyRedHat:
@nakato thanks a bunch for the PR and for filling that gap.  The changes and functionality LGTM with the few noted questions.  One of the tests failed, I believe it was a flake and I've restarted it.

imagebuildah/util.go
nakato: Woops, that'll be fixed in the next push.

imagebuildah/util.go
nakato: I haven't looked too much into how the unpacking works as used here as this is based on the above `downloadToDirectory` function above.

If we decide to change `os.Remove` to `os.RemoveAll`, we should probably update `downloadToDirectory` in a parent commit first so the behavior of loading context via stdin or url remain consistent.

Now looking into the unpacking, as far as I can tell at this time, this is what occurs in both `stdinToDirectory` and `downloadToDirectory.

1. If untar dies because input is not a tar, it is assumed to be a Dockerfile, and the output is written to `tmpdir/Dockerfile` without any further checking, this is the expected path.

2. If untar dies because of an error with the tar, such as corruption, short-read, etc, then the files are left behind and the tar is written to Dockerfile, this stage continues as a success, and I presume the folder gets cleaned up in the calling function after the build fails because Dockerfile is actually a tar.

3. If the tar unpacking dies because it runs out of diskspace, we attempt to create the Dockerfile and this will fail due to diskspace, then `os.Remove` will fail, and we will leave behind a full filesystem on exit.

4. It's not a tar, so no tar files are created, but the filesystem doesn't have enough space for the Dockerfile, so we leave behind a half written Dockerfile, then fail to cleanup because the directory is not empty.

As far as I can tell, the only time these functions are going to truly fail is in scenario 3 and 4.  2 is probably somewhat undefined behavior that's not great.

---

Before I switch this to `os.RemoveAll`, I don't see anything that would make this dangerous, but is there anything I'm not seeing here?  I'd hate to cause someone's files to go poof unexpectedly.

---

If we used something like `libmagic` we could figure out if the input was compressed/tar before deciding what to do with it, which would improve our ability to error-handle here, and means I could use `bufio` to not read the whole tar into memory, and use `Peek` to figure out if we have a tar or a compressed stream which we would assume to be a tar.  But that leaves two questions.  What's the policy on CGO (I don't see any pure-go implementations for this)?  And would moving this much logic here be acceptable?

rhatdan: One issue with CGO would be this tool being vendored into other tools like Podman and OpenShift builder, and perhaps others (ansible?)  

The problem you are worried about is the user specifying a NAME path that containers other files correct?  So that RemoveAll would somehow remove files that the user did not download.  Could we check before the untar if the NAME path exists and cause an error?  Is the NAME Path supposed to be a randomly created file in a random directory then we would not need to worry about this?

nakato: `name` is created by `ioutil.TempDir(dir, prefix)`
So we can `removeall` as long as `ioutil.TempDir` always gives us a temporary directory.

nakato:
Updated with RemoveAll, and added a parent commit to update the other uses of `os.Remove` to ensure all external contexts act the same.

rhatdan:
@nakato You added another commit, without signing it.

rhatdan:
LGTM
Just need to sign and we can merge.
Thanks @nakato

nakato:
Woops, I forgot that.
Signed now.

rhatdan:
@TomSweeneyRedHat PTAL
Lets get this merged when it passes the tests.

TomSweeneyRedHat:
LGTM
Now we just need Travis to get moving....

rhatdan:
@rh-atomic-bot r+

Issue #1871, opened 2019-09-27
--build-args do not work as expected within docker "ADD" blocks :  buildah version 1.11.2


tiernemi:
Having updated to version 1.11.2 of buildah the build args in a dockerfile do not resolve to their correct value in ADD blocks. I have created a minimal repro as follows

1. Create a file called abc.txt
2. Create a dockerfile as follows
```Dockerfile
FROM centos
ARG testArg
ADD ${testArg}.txt .
```
3. Run the following command
```bash
buildah bud --build-arg testArg=abc
```

You should receive the following output
```
STEP 1: FROM centos
STEP 2: ARG testArg
STEP 3: ADD ${testArg}.txt .
error dry-running "ADD ${testArg}.txt .": no files found matching "/home/tiernemi/buildah_test/${testArg}.txt": no such file or directory
ERRO exit status 1                                
```

When running the following command with docker
```bash
docker build . --build-arg testArg=abc
```
It succeeds with no issue.

TomSweeneyRedHat:
@tiernemi Thanks for the issue, especially since you've a nice reproducer.  Will take a look.

TomSweeneyRedHat:
@tiernemi I'm able to reproduce and it looks like the ARG part of things are working appropriately as I'm able to do `RUN echo $testArg` just before the `ADD` line and that resolves OK.  Do you by chance know if this worked in earlier versions of Buildah?

tiernemi:
It was working using the previous version of buildah in our pipeline. Version 1.10.1-2.

TomSweeneyRedHat:
Thanks @tiernemi, I think I've found a PR that went into 1.11.0 that may have caused the issue, looking further.

dustymabe:
we can close this now that #1874 merged?

TomSweeneyRedHat:
@dustymabe yes we can!  I was holding off until we got a release together for Podman, but just verified this is in Buildah v1.11.3 which has been vendored upstream.  Thanks for the poke.

PR #1874, opened 2019-09-28
Fix --build-args handling

TomSweeneyRedHat:
When passing an environment variable to a Containerfile using an `ARG` command,
the dry-run processing  for `COPY` and `ADD` isn't resolving the environment
variable for the source and is causing a failure.

Now convert the source if it's an environment variable to its value and then
proceed on.  The destination doesn't have a similar issue at this point 
but we'll convert it too just in case.

Addresses: #1871

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -253,7 +253,7 @@ func (s *StageExecutor) volumeCacheRestore() error {
 // don't care about the details of where in the filesystem the content actually
 // goes, because we're not actually going to add it here, so this is less
 // involved than Copy().
-func (s *StageExecutor) digestSpecifiedContent(node *parser.Node) (string, error) {
+func (s *StageExecutor) digestSpecifiedContent(node *parser.Node, argValues []string) (string, error) {
 	// No instruction: done.
 	if node == nil {
 		return "", nil
@@ -297,7 +297,15 @@ func (s *StageExecutor) digestSpecifiedContent(node *parser.Node) (string, error
 			}
 		}
 	}
+
 	for _, src := range srcs {
+		// If src has an argument within it, resolve it to its
+		// value.  Otherwise just return the value found.
+		name, err := imagebuilder.ProcessWord(src, argValues)
+		if err != nil {
+			return "", errors.Wrapf(err, "unable to resolve source %q", src)
+		}
+		src = name
 		if strings.HasPrefix(src, "http://") || strings.HasPrefix(src, "https://") {
 			// Source is a URL.  TODO: cache this content
 			// somewhere, so that we can avoid pulling it down
@@ -334,7 +342,14 @@ func (s *StageExecutor) digestSpecifiedContent(node *parser.Node) (string, error
 	}
 	s.builder.ContentDigester.Restart()
 	download := strings.ToUpper(node.Value) == "ADD"
-	err := s.builder.Add(destination.Value, download, options, sources...)
+
+	// If destination.Value has an argument within it, resolve it to its
+	// value.  Otherwise just return the value found.
+	destValue, destErr := imagebuilder.ProcessWord(destination.Value, argValues)
+	if destErr != nil {
+		return "", errors.Wrapf(destErr, "unable to resolve destination %q", destination.Value)
+	}
+	err := s.builder.Add(destValue, download, options, sources...)
 	if err != nil {
 		return "", errors.Wrapf(err, "error dry-running %q", node.Original)
 	}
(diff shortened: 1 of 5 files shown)


TomSweeneyRedHat:
@nalind I looked for a while today, but couldn't come up with a quick way to resolve the destination for the dry run if it was an environment variable.  I'll look more tomorrow. If you've suggestions, happy to take them.

@mheon, @rhatdan we may want to push to at least get this stop gap fix in for 1.6.

rhatdan:
I take it buildah has not translate $build_arg internally at this point?

TomSweeneyRedHat:
@rhatdan, not that I could find quickly.  I was going to dive in again Monday, but think it's down in imagebuilder.

nalind:
Would having the callers of `StageExecutor.digestSpecifiedContent()` pass `ib.Arguments()` in as an additional argument, and have `digestSpecifiedContent()` use that slice and `imagebuilder.ProcessWord()` to expand source and destination values, produce the right results?

TomSweeneyRedHat:
@nalind, maybe, I'll take a look.  Thanks for the suggestion/pointers.

imagebuildah/stage_executor.go
```diff
@@ -297,7 +297,27 @@ func (s *StageExecutor) digestSpecifiedContent(node *parser.Node) (string, error
...
+	}
+
 	for _, src := range srcs {
+		// If a src is an environment variable, try to resolve it.
+		if strings.HasPrefix(src, "$") {
+			src = src[1:]
+			if strings.HasPrefix(src, "{") && strings.HasSuffix(src, "}") {
+				src = src[1 : len(src)-1]
+			}
+			src = envVarMap[src]
```
rhatdan: Should we report a failure whenr envVarMap[src] does not exists?

TomSweeneyRedHat: Shouldn't happen, but probably not a bad idea.  I've bumped the tests up a bit more and wanted to play with ib.ProcessWord this morning, ran outta gas last night for both.  If the ib.ProcessWord doesn't have me rework this section, then I'll add an error.

TomSweeneyRedHat:
FYI @mheon

mheon:
This is too late for Podman 1.6.0, but we have some important rootless bugs, so we'll probably be cutting a 1.6.1 later this week anyways.

imagebuildah/stage_executor.go
```diff
@@ -297,7 +297,17 @@ func (s *StageExecutor) digestSpecifiedContent(node *parser.Node) (string, error
 			}
 		}
 	}
+
 	for _, src := range srcs {
+		// If a src is an environment variable, try to resolve it.
+		if strings.HasPrefix(src, "$") {
```
nalind: Does this need to be conditional?

nalind:
We'll probably want to expand variables in the destination, in case it includes one, around line 347.

imagebuildah/stage_executor.go
TomSweeneyRedHat: It doesn't need to be, meaning the call to ProcessWord will work, but I thought it would be better if we did the check.  Otherwise, we'll hit up ProcessWord for a Dockerfile line like `COPY myfile.txt /tmp/myfile.txt` and thought I'd skip the extra processing.  But if you think it might prove safer to remove the if statement here, I will.

TomSweeneyRedHat:
My testing didn't show that I need to add the check on the destination too, but maybe that was luck?  I'll ProcessWord(destination.Value) just above the line you suggested to be ultra sure.

imagebuildah/stage_executor.go
nalind: Mainly I want to avoid getting tripped up by people using arguments in the middle of values, or at the end for file extensions or things I haven't thought of.

TomSweeneyRedHat: hmm, I'd not considered the old argument in the middle of a variable trick.  Sold, I'll remove the check for the beginning '$' and will whack them all.  Thanks!

TomSweeneyRedHat:
Touch up with @nalind's latest suggestion/comment and have rebased.    Finger's crossed for the CI.

rhatdan:
Your tests have a bug in them.

TomSweeneyRedHat:
@rhatdan, I'd accidentally remove the '}' from the prior test when I cut/pasted the new ones in.  Ugh.

rhatdan:
LGTM and I like the tests.
@nalind Please take one final look.

nalind:
LGTM

TomSweeneyRedHat:
Happy green test buttons and two LGTM's if someone wants to merge this in...

nalind:
@rh-atomic-bot r+

PR #1891, opened 2019-10-01
Update c/image to v4.0.1

mtrmac:
<s>**Do not merge:** This is to invoke CI on both Buildah and Podman to gain some confidence with this code. Includes UNMERGED https://github.com/containers/image/pull/724 = https://github.com/containers/image/pull/721 + https://github.com/containers/image/pull/723.</s>

Note: made using `go get github.com/containers/image@v4.0.0+incompatible` so that import paths don't have to updated, and because using the proper `github.com/containers/image/v4@v4.0.0` rejects the `go.mod` of c/image it should declare itself to be a /v4.

This is clearly a bug in c/image, though it’s somewhat tempting to call it a feature if we are going to keep increasing the major versions at the pace we’ve been recently.


mtrmac:
Retrying with https://github.com/containers/image/pull/721 , to use Go module major versions properly. “Fun”.

rhatdan:
Nice thing about integers is there are a lot of them.:^)

mtrmac:
The ~gratituous modifications of, currently, 30 source files in Buildah every time a bump happens, and the big move in the `vendor` directory, are rather annoying, though. Sure, there are tools to do that, but it will pollute the history for anyone researching the past.

(And we don’t get the supposed benefit of various dependencies using different major versions because `c/image/types.SystemContext` is a part of the Buildah API.)

vrothberg:
Travis failed with
```
$ make vendor
0.26s$ ./hack/tree_status.sh
tree is dirty, please commit all changes and sync the vendor.conf
 M go.sum
```

vrothberg:
[go move\(https://github.com/KSubedi/gomove) can help doing these migrations to newer version.

I tried it locally with `$ go move $image $image/v4` and it rewrites the imports accordingly.

rhatdan:
Why don't we want to merge 4.0.0?  @vrothberg @mtrmac ?
Is there an issue with 4.0.0?  I know there was further work going on there.  But the #1 goal is to get buildah to work in FIPS-MODE.  Is there a reason we can not just upgrate to c/i 4.0.0 and merge it into buildah?

vrothberg:
> Why don't we want to merge 4.0.0? @vrothberg @mtrmac ?
> Is there an issue with 4.0.0? I know there was further work going on there. But the #1 goal is to get buildah to work in FIPS-MODE. Is there a reason we can not just upgrate to c/i 4.0.0 and merge it into buildah?

It's related to how go modules work with version >= 2.0.0. See https://github.com/containers/image/pull/721 for some more details. In short, a go modules is supposed to have a major-version namespace in it's package paths as soon as the major version is >= 2.0.0. We just enabled go modules with the latest 4.0.0 release and are hitting the issue for the first time.

@mtrmac has the fixes in place in https://github.com/containers/image/pull/721, so I think we can quickly release a 4.0.1 that does things module-conform.

mtrmac:
> Why don't we want to merge 4.0.0? @vrothberg @mtrmac ?

https://github.com/containers/libpod/pull/4128 was reporting issues, and, well, we wouldn’t have _known_ that v4.0.0 is fine until it was actually tested.

As it turns out:
- The previous way to vendor c/image doesn’t work at all with the proxy now that it is a module: https://github.com/containers/libpod/pull/4165#issuecomment-537223255
- The previous way to vendor c/image seems to require every consumer to coordinate _precisely_: https://github.com/containers/libpod/pull/4165#issuecomment-537230326
- c/image was not a correctly formed v4 module: https://github.com/containers/image/pull/721 (and this is painful enough to, possibly, consider reverting to v0. or something, at least in theory)
- Yes, podman does fail with v4.0.0: https://github.com/containers/libpod/pull/4165#issuecomment-537231215

Sure, we should have noticed all of these before merging the respective c/image PRs. But now that there are known problems, let’s actually fix them before merging this, or this Buildah PR will prevent getting any other Buildah updates into Podman (not to mention the known-to-exist bugs with unknown impact).

mtrmac:
Testing the combined https://github.com/containers/image/pull/724 now.

mtrmac:
Now includes https://github.com/containers/image/pull/725.

rhatdan:
@mtrmac Tests are passing in Buildah.  Good sign?

mtrmac:
The tests in this PR have always been passing, so that’s not a strong signal.

mtrmac:
Updated to use released v4.0.1. PTAL.

import.go
```diff
@@ -5,11 +5,11 @@ import (
...
 	"github.com/containers/buildah/util"
-	"github.com/containers/image/manifest"
-	is "github.com/containers/image/storage"
-	"github.com/containers/image/types"
+	"github.com/containers/image/v4/manifest"
+	is "github.com/containers/image/v4/storage"
+	"github.com/containers/image/v4/types"
 	"github.com/containers/storage"
-	"github.com/opencontainers/go-digest"
+	digest "github.com/opencontainers/go-digest"
```
nalind: Thanks for fixing these :+1:

nalind:
LGTM

TomSweeneyRedHat:
LGTM
@mtrmac TYVM for all the work you're doing here.

vrothberg:
@rh-atomic-bot r+

Thanks, @mtrmac!

PR #1900, opened 2019-10-07
DO-NOT-MERGE: Add squash-all for podman build

TomSweeneyRedHat:
Add a `--squash-all` flag that will be unique to `podman build` and will not
be used by Buildah.  `buildah bud --layers=false` is equivalent to `docker build --squash` (experimental).  When a podman build command includes the `--squash` option, podman will translate that to `--layers=false` for the buildah bud api code.  A `docker build` call is equivalent to `podman build` or `podman build --layers=true` and that's not changing.

When podman build uses the `--squash-all` command, it will turn on the --squash option for the buildah bud api to have only one layer returned by the buildah bud code.  There's not a Docker equivalent for this at the moment.

I'll hopefully be spinning up a Podman PR in the next day or two that will use this new option.

### DIFF
--- a/pkg/cli/common.go
+++ b/pkg/cli/common.go
@@ -42,6 +42,12 @@ type NameSpaceResults struct {
 	UTS           string
 }
 
+// PodmanBuildResults represents the results for Podman Build flags
+// that are unique to Podman
+type PodmanBuildResults struct {
+	SquashAll bool
+}
+
 // BudResults represents the results for Bud flags
 type BudResults struct {
 	Annotation          []string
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
FYI @mheon

TomSweeneyRedHat:
fyi @baude

pkg/cli/common.go
```diff
@@ -42,6 +42,12 @@ type NameSpaceResults struct {
 	UTS           string
 }
 
+// PodmanBuildResults represents the results for Podman Build flags
```
rhatdan: Can't we just add this to Podman?
libpod/cmd/build.go?

TomSweeneyRedHat: Probably.  All the other definitions are here, so thought I'd keep them together.  I'll try adding it to cmd/build.go.

TomSweeneyRedHat: Per @rhatdan's suggestion, I've moved this code into Podman in https://github.com/containers/libpod/pull/4215.  If those changes are merged there, I will close this PR.

TomSweeneyRedHat:
Closing as this was addressed via: https://github.com/containers/libpod/pull/4215

PR #1904, opened 2019-10-10
Move runtime flag to bud from common

TomSweeneyRedHat:
Move the runtime Cobra flag for the build command to the
command itself and out of the common definitions.  Podman
uses the same definition and there was a collision with that
flag and the one here: https://github.com/containers/libpod/blob/master/cmd/podman/main_local.go#L69

This caused wonky results.  The podman global flag would
think it had been set, but no value was captured, or reversely,
the flag didn't think it had been set, but if you peered inside
you'd see the appropriate value from the user input.
This was all based on whether you did
`podman --runtime=/usr/bin/runc build` or
`podman build --runtime=/usr/bin/runc`

As Podman uses their runtime flag in multiple places, it's
easier to remove it from the common definition here in Buildah
and then add the flag into the bud command.

Once merged and vendored into Podman, this will fix: https://bugzilla.redhat.com/show_bug.cgi?id=1758082

### DIFF
--- a/pkg/cli/common.go
+++ b/pkg/cli/common.go
@@ -161,7 +161,7 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
 	fs.BoolVar(&flags.PullAlways, "pull-always", false, "pull the image, even if a version is present")
 	fs.BoolVarP(&flags.Quiet, "quiet", "q", false, "refrain from announcing build instructions and image read/write progress")
 	fs.BoolVar(&flags.Rm, "rm", true, "Remove intermediate containers after a successful build")
-	fs.StringVar(&flags.Runtime, "runtime", util.Runtime(), "`path` to an alternate runtime. Use BUILDAH_RUNTIME environment variable to override.")
+	// "runtime" definition moved to avoid name collision in podman build.  Defined in cmd/buildah/bud.go.
 	fs.StringSliceVar(&flags.RuntimeFlags, "runtime-flag", []string{}, "add global flags for the container runtime")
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
FYI @mheon, @baude

pkg/cli/common.go
```diff
@@ -161,7 +161,8 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
 	fs.BoolVar(&flags.PullAlways, "pull-always", false, "pull the image, even if a version is present")
 	fs.BoolVarP(&flags.Quiet, "quiet", "q", false, "refrain from announcing build instructions and image read/write progress")
 	fs.BoolVar(&flags.Rm, "rm", true, "Remove intermediate containers after a successful build")
-	fs.StringVar(&flags.Runtime, "runtime", util.Runtime(), "`path` to an alternate runtime. Use BUILDAH_RUNTIME environment variable to override.")
+	//	Remove next due to flag name collision in podman build.  Define in cmd/buildah/bud.go.
```
rhatdan: Just remove it, do not comment it out.

TomSweeneyRedHat: OK, but I'm planning to leave a comment in here to make it easier to find and to avoid the issue going forward.

rhatdan:
LGTM
@nalind @vrothberg @giuseppe @QiWang19 PTAL

nalind:
LGTM for fixing this case.

rhatdan:
@rh-atomic-bot r+

PR #1905, opened 2019-10-10
Commit: check for storage.ErrImageUnknown using errors.Cause()

nalind:
The storage library can return its errors wrapped in additional context, so to do a direct comparison we need to dig down to the root cause.


giuseppe:
LGTM

PR #1909, opened 2019-10-11
WIP: use cobra's autogenerated completions

nalind:
Use cobra's autogenerated completion functions instead of our hand-rolled ones.


rhatdan:
I think 71ba2e4ef0a762abf83035dedd35dbccbdcd983c got us a lot closer to making this a reality.

PR #1902, opened 2019-10-09
containers/image v5; support manifest lists; add a "manifest" command

nalind:
Vendor in WIP branches for manifest list support to let CI exercise the changes.

PR #1910, opened 2019-10-12
Add completions for Manifest commands

TomSweeneyRedHat:
This adds the completions for the manifest commands
found in #1902.  This should not be merged until that
is merged, or perhaps this should be cherry-picked
into it.

### DIFF
--- a/contrib/completions/bash/buildah
+++ b/contrib/completions/bash/buildah
@@ -690,6 +690,163 @@ return 1
      esac
  }
 
+ _buildah_manifest() {
+     local boolean_options="
+     --help
+     -h
+     --all
+  "
+     subcommands="
+        add
+        annotate
+        create
+        inspect
+        push
+        remove
+     "
+     __buildah_subcommands "$subcommands" && return
+
+     case "$cur" in
+         -*)
+             COMPREPLY=($(compgen -W "$boolean_options " -- "$cur"))
+             ;;
+         *)
+             COMPREPLY=( $( compgen -W "$subcommands" -- "$cur" ) )
+             ;;
+     esac
+
+}
+ _buildah_manifest_add() {
+     local boolean_options="
+     --help
+     -h
+     --all
+  "
+
+     local options_with_args="
+     --annotation
+     --arch
+     --features
+     --os
+     --os-features
+     --os-version
+     --variant
+  "
+
+     local all_options="$options_with_args $boolean_options"
+
+     case "$cur" in
+         -*)
+             COMPREPLY=($(compgen -W "$boolean_options $options_with_args" -- "$cur"))
+             ;;
+     esac
+ }
+
+ _buildah_manifest_annotate() {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
I've remove the "Do Not Merge" label as #1902 has now merged.   This should be gtg now and still has happy green test buttons.
@vrothberg @QiWang19 @nalind PTAL

rhatdan:
Don't you need to add manifest as a command to _buildah()?
Then we need a _buildah_manifest() section to get all of these, I believe?

TomSweeneyRedHat:
@rhatdan PTAL

rhatdan:
LGTM
@edsantiago @cevich @nalind @QiWang19 @vrothberg @giuseppe PTAL

QiWang19:
LGTM

contrib/completions/bash/buildah
```diff
@@ -690,6 +690,163 @@ return 1
...
+     local options_with_args="
+  "
+
+     local all_options="$options_with_args $boolean_options"
+
+     case "$cur" in
+         -*)
+             COMPREPLY=($(compgen -W "$boolean_options $options_with_args" -- "$cur"))
+             ;;
+     esac
```
cevich: This single-item case (?) is repeated four or five times, could it be made into a function or simplified some other way?

contrib/completions/bash/buildah
```diff
@@ -690,6 +690,163 @@ return 1
...
+     -h
+     --all
+  "
+     subcommands="
+        add
+        annotate
+        create
+        inspect
+        push
+        remove
```
edsantiago: Trailing whitespace - I'm curious why the commit hooks didn't catch this?

edsantiago:
This all gives me the sense of wandering into cargo-cult territory: the NOP-y declaration of `all_options`, the unnecessary splitting of `boolean_options` and `options_with_args`. I'm sure the intentions were good in the early days of this completion, but right now it's just copy-paste incantations without meaning. I feel like we should maybe drop the pretense, just declare a single `options` list and keep it clean. (Not on existing code, just on new additions). That would be easier to read, and IMO less misleading.

(Not intended as a slight on your work -- this completion stuff is HARD).

contrib/completions/bash/buildah
TomSweeneyRedHat: fixed, thx.

contrib/completions/bash/buildah
TomSweeneyRedHat: Maybe it could be improved, but we've it all over the place in this file and I'm just staying between the lines....

rhatdan:
Perhaps we get an intern next summer to concentrate on improving command completions, Make them work under @edsantiago :^)

rhatdan:
@rh-atomic-bot r+

PR #1913, opened 2019-10-14
Fix copy/add behaviour with .dockerignore

JamesWrigley:
When a directory is being copied/added, the whole directory was being copied in unless there was an exclusion (e.g., `!include-this-file`) in the `.dockerignore`.

So for this `Dockerfile`:
```Dockerfile
FROM alpine
COPY . /tmp/foo
RUN ls /tmp/foo
```

With this `.dockerignore`:
```bash
# !foo
quux
```

A file named `quux` in the build context would be copied in, unless the first line was uncommented. This is fixed by checking if there are any patterns at all, instead of for an exclusion.

Note: the `run_buildah run myctr2 ls -l matched` test fails on my machine with: `printf: write error: Resource temporarily unavailable`. I'm not sure why this happens, as far as I can tell it's correct.

### DIFF
--- a/add.go
+++ b/add.go
@@ -299,7 +299,7 @@ func (b *Builder) addHelper(excludes *fileutils.PatternMatcher, extract bool, de
 					}
 				}
 				logrus.Debugf("copying[%d] %q to %q", n, esrc+string(os.PathSeparator)+"*", dest+string(os.PathSeparator)+"*")
-				if excludes == nil || !excludes.Exclusions() {
+				if excludes == nil || len(excludes.Patterns()) == 0 {
 					if err = copyWithTar(esrc, dest); err != nil {
 						return errors.Wrapf(err, "error copying %q to %q", esrc, dest)
 					}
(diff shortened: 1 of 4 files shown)


JamesWrigley:
Hmm, while I believe the behaviour is now correct, this change would also trigger #1714 more often :\

TomSweeneyRedHat:
@JamesWrigley  and @saschagrunert thanks for the dockerignore fixes, but are they fixing the same issue?  #1913 and #1914?  I've not taken a deep dive.

TomSweeneyRedHat:
@giuseppe PTAL

saschagrunert:
> @JamesWrigley and @saschagrunert thanks for the dockerignore fixes, but are they fixing the same issue? #1913 and #1914? I've not taken a deep dive.

Ah yes, sorry I did not see that. Feel free to pick the solution you prefer. :)

JamesWrigley:
They do fix the same issue, but I think @saschagrunert's fix is better :)

rhatdan:
Let's get @saschagrunert fix in.

PR #1914, opened 2019-10-14
Fix .dockerignore exclude regression

saschagrunert:
In #1715 we added an additional valid path for `!excludes.Exclusions()`
when considering a full copy of the source directory to the destination
via `ADD`/`COPY`. This breaks the functionality of the `.dockerignore`
file, because `PatternMatcher.Exclusions` is true if any of the defined
patterns contain an exclusion (the pattern starts with `!`). This means
if we have no Exclusions defined at all, the `.dockerignore` behavior
does not work any more.

To still have the optimization in, we now skip whole directories if the
patterns to exclude match.

Introduced by: https://github.com/containers/buildah/pull/1715


saschagrunert:
I'm not sure what's wrong with travis but is seems to abort the test runs without any further message...

rhatdan:
LGTM
@nalind @QiWang19 @giuseppe @vrothberg @TomSweeneyRedHat PTAL

TomSweeneyRedHat:
This and #1913 are fixing the same problem.  We need to merge one and close the other.  I'd vote going forward with this one.

TomSweeneyRedHat:
@giuseppe could you give us a head nod on this please?

rhatdan:
@saschagrunert looks like dockerignore tests are failing.

saschagrunert:
> @saschagrunert looks like dockerignore tests are failing.

I see. To still have the improvement in and let the matching works as for now we would have to extend the information provided by the `fileutils` package. I opened up https://github.com/containers/storage/pull/443 for that.

go.mod
```diff
@@ -49,3 +49,5 @@ require (
 	k8s.io/api v0.0.0-20190813020757-36bff7324fb7 // indirect
 	k8s.io/client-go v0.0.0-20181219152756-3dd551c0f083 // indirect
 )
+
+replace github.com/containers/storage => github.com/saschagrunert/storage v1.12.3-0.20191015073819-a34ddea087da
```
saschagrunert: TODO: Remove this when https://github.com/containers/storage/pull/443 got merged (or not)

saschagrunert:
Rebased on top of the master and vendored the latest storage changes in to see if everything works as intended.

giuseppe:
LGTM

giuseppe:
@saschagrunert thanks for fixing it!  Have you tried if COPY/ADD performs much slower after this PR?

saschagrunert:
> @saschagrunert thanks for fixing it! Have you tried if COPY/ADD is much slower after this PR?

I did not find any larger performance regression (did some local tests) but I think we should prepare an official test case for this.

TomSweeneyRedHat:
@saschagrunert do you plan to add a test with this PR, or should we get this in and follow up later?

saschagrunert:
> @saschagrunert do you plan to add a test with this PR, or should we get this in and follow up later?

I'll prepare something and propose it here on Monday :)

saschagrunert:
Alright, so the previous vs now COPY results are not really compare-able, since the current functionality is broken on the master branch. In the end we could add a small performance test on top of bats, like this:

```
#!/usr/bin/env bats

load helpers

function measure() {
    START=$(date +%s.%N)
    $@ >/dev/null 2>&1
    END=$(date +%s.%N)
    echo "$END - $START" | bc
}

function checkout() {
    DIR=$(mktemp -d)

    git clone -q "$1" "$DIR"
    pushd "$DIR" >/dev/null
    git checkout -qf "$2"
    popd >/dev/null

    echo "$DIR"
}

function merge_base() {
    git merge-base origin/master $(git rev-parse --abbrev-ref HEAD)
}

@test "copy performance" {
    # Prepare the comparison buildah version
    MERGE_BASE=$(merge_base)
    PREV_BUILDAH_DIR=$(checkout https://github.com/containers/buildah $MERGE_BASE)
    pushd "$PREV_BUILDAH_DIR"
    make buildah
    popd

    # Prepare the test case
    TEST_DIR=$(checkout \
        https://github.com/openshift/cluster-api-provider-aws \
        151ec6cd21b9739206fd16f2ddf9f036318d7917)
    pushd "$TEST_DIR"
    FILE=Containerfile
    printf "FROM scratch\nCOPY . ." > $FILE
    printf ".git\nvendor" > .dockerignore

    # Run the test and compare the results
    PRE=$(measure "$PREV_BUILDAH_DIR"/buildah bud -f $FILE)
    NOW=$(measure run_buildah bud -f $FILE)

    # A diff of 5 seconds will let the test fail
    if (($(echo "$NOW - $PRE > 10" | bc -l))); then
        echo "Copy benchmark results differ too much"
        echo "Previous ($MERGE_BASE): $PRE"
        echo "Now: $NOW"
        exit 1
    fi
    popd
    rm -rf $TEST_DIR $PREV_BUILDAH_DIR
}
```

I tried it out and it works, but I'm not sure if we should add things like that on top of hardware where we have no influence on.

saschagrunert:
If we want to move forward here then we would have to get https://github.com/containers/storage/pull/446 in first.

TomSweeneyRedHat:
@saschagrunert I'm kinda of leery about adding the timing test like you have it.  As noted, different hardware could cause failure just due to bad timing.  

As your testing hasn't found any inherent slowness due to the change, I'm fine with moving forward with containers/storage#446 and this PR.   @giuseppe @rhatdan @vrothberg @nalind WDYT?

vrothberg:
@TomSweeneyRedHat, I dropped final comments on  containers/storage#446. It would be good to have more pairs of eyes on it.

rhatdan:
@saschagrunert Needs a rebase.

TomSweeneyRedHat:
ping @saschagrunert rebase still needed here.  Once done I think this is GTG right?

saschagrunert:
Unfortunately not we need to get in https://github.com/containers/storage/pull/446 first.

rhatdan:
@saschagrunert I would like to setup a phone call to discuss these patches next week.  So we can understand it better.

saschagrunert:
> @saschagrunert I would like to setup a phone call to discuss these patches next week. So we can understand it better.

Okay, sounds good to me. I’m in the office until Thursday. On Friday I’m already traveling to San Diego.

rhatdan:
Ok we are having Face2Face meetings next week, so we will setup a bluejeans and try to hook up with you on Tuesday or Wednesday.  And we will meet up with you at Kubecon.

saschagrunert:
Rebased on top of the latest changes in https://github.com/containers/storage/pull/446 (still a requirement to move forward here).

saschagrunert:
Rebased on top of the latest changes to see if it still builds.

TomSweeneyRedHat:
@saschagrunert, all green test buttons.  What's next for this PR?

saschagrunert:
> @saschagrunert, all green test buttons. What's next for this PR?

Hey, do you think that we can release a new version for containers/storage or do you prefer if I vendor its latest master here? :)

rhatdan:
Wait for release.  We want to keep momentum on Debian moving and they want proper versioning.

saschagrunert:
Rebased on top of containers/storage v1.14.0

rhatdan:
@TomSweeneyRedHat Tests are timing out...

TomSweeneyRedHat:
@rhatdan ugh.  @saschagrunert can you try rebasing one more time please?

saschagrunert:
> @rhatdan ugh. @saschagrunert can you try rebasing one more time please?

Re-triggered the CI, lets see if it runs through.

rhatdan:
@rh-atomic-bot r+

PR #1917, opened 2019-10-15
Add builder labels

### LINKED ISSUE #1864: Provide metadata to indicate version of tools used to create a container
**Description**

Sometimes when debugging it's useful to know the version of Podman that was used to create a container. eg., a situation where a container that was created with an old version of Podman now fails to start with a newer version.

It would nice if this information was exposed somewhere. It could be via `inspect` or something in `/run/.containerenv` or something else.

caiges:
This adds builder labels during a commit. Closes #1864 

Signed-off-by: caiges <caigesn@gmail.com>

### DIFF
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -162,6 +162,9 @@ func commitCmd(c *cobra.Command, args []string, iopts commitInputOptions) error 
 		}
 	}
 
+	// Add builder identity information.
+	builder.SetLabel(buildah.BuilderIdentityAnnotation, buildah.Version)
+
 	options := buildah.CommitOptions{
 		PreferredManifestType: format,
 		Compression:           compress,
--- a/commit.go
+++ b/commit.go
@@ -29,6 +29,13 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	// BuilderIdentityAnnotation is the name of the annotation key containing
+	// the name and version of the producer of the image stored as an
+	// annotation on commit.
+	BuilderIdentityAnnotation = "io.buildah.version"
+)
+
 // CommitOptions can be used to alter how an image is committed.
 type CommitOptions struct {
 	// PreferredManifestType is the preferred type of image manifest.  The
--- a/tests/commit.bats
+++ b/tests/commit.bats
@@ -103,3 +103,15 @@ load helpers
   buildah rm $cid
   buildah rmi -a
 }
+
+@test "commit-builder-identity" {
+	cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+	buildah commit --signature-policy ${TESTSDIR}/policy.json $cid alpine-image
+
+	buildah_version=$(buildah --version | awk '{ print $3 }')
+	version=$(buildah inspect --format '{{ index .Docker.Config.Labels "io.buildah.version"}}' alpine-image)
+
+	[ "$version" == "$buildah_version" ]
+	buildah rm $cid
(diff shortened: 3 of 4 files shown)


vrothberg:
Hi @caiges, thanks for opening the PR! I think it's a bit too early to merge it as there's no consensus yet in https://github.com/containers/buildah/issues/1864. I suggest to a bit longer and see if this is something we could get into the OCI spec. This would prevent us from letting users depend on something that is likely to change in the future.

caiges:
Oh totally agree, I just wanted to throw out what the implementation & impact might look like.

cmd/buildah/commit.go
```diff
@@ -158,6 +158,10 @@ func commitCmd(c *cobra.Command, args []string, iopts commitInputOptions) error
 		}
 	}
 
+	// Add builder information.
+	builder.SetLabel("org.opencontainers.image.builder.name", "buildah")
+	builder.SetLabel("org.opencontainers.image.builder.version", buildah.Version)
```
vrothberg: Based on @vbatts comment https://github.com/containers/buildah/issues/1864#issuecomment-547390122, I think that we should not use the `org.opencontainers.image` prefix to prevent users from assuming that's a supported annotation of the OCI image spec.

How about a single `io.buildah.version` label? It's presence shows both, that is was built with buildah _and_ with which version? @rhatdan @caiges @TomSweeneyRedHat @nalind WDYT?

TomSweeneyRedHat: I concur on not using the org.opencontainers.image.  Where's the "io" from in your suggestion @vrothberg ?  My initial thought was "containers.image.*"

rhatdan:
Needs a rebase.

cmd/buildah/commit.go
rhatdan: I think that is a good idea.

vrothberg: > I concur on not using the org.opencontainers.image. Where's the "io" from in your suggestion @vrothberg ? My initial thought was "containers.image.*"

The "io" comes from reversing the Buildah domain `buildah.io` -> `io.buildah` which might help consumers of an image find out what this label means. Maybe, just `io.buildah` as the key and have `vX.Y.Z` as the value?

vbatts: so it could be com.github.containers.buildah.version/commit too

rhatdan: I like io.podman, nice and simple and easy to find.

nalind: [The spec](https://github.com/opencontainers/image-spec/blob/master/annotations.md)'s rules suggest the reverse-the-order-of-dns-labels method that's used for Java packages, D-Bus services, and others.  This project doesn't "own" the `com.github` namespace prefix, and mustn't use the `org.opencontainers` prefix.

nalind: Please make these annotation keys constants that we export, so that API consumers can refer to them without risking typos.

caiges: That's a decent thought, I've exported the annotation key.

caiges: I went with `io.buildah` for now since the podman/buildah relationship might be a little confusing if we were to use `io.podman`. I might be missing some context there though.

TomSweeneyRedHat:
LGTM assuming happy tests

TomSweeneyRedHat:
@vrothberg @nalind @rhatdan PTAL

rhatdan:
@rh-atomic-bot retry

TomSweeneyRedHat:
@rh-atomic-bot r+

Issue #1906, opened 2019-10-11
With Dockerfile ADD .tar.xz reuses cache even when files changed


shaicoleman:
**Description**

Dockerfile ADD command reuses cache even when files have changed

**Steps to reproduce the issue:**
1. On Ubuntu 18.04, install podman 1.6.1 from PPA (Buildah 1.11.2)
2. Create the following `Dockerfile`
```
FROM docker.io/busybox

ADD test.tar.xz /
```
3. Build a test archive, and run podman build
```
$ date>test; tar cJf - test > test.tar.xz; md5sum test.tar.xz
158a9e025b54fe66713a443504821a8a  test.tar.xz
$ sudo podman build --file Dockerfile -t podman-test .
STEP 1: FROM docker.io/busybox
STEP 2: ADD test.tar.xz /
STEP 3: COMMIT podman-test
0f7320b3df37c68a91455477f27ec042b832eb589a44bc98365e85449c92cff5
```
4. Create a new test archive,  run podman build again
```
$ date>test; tar cJf - test > test.tar.xz; md5sum test.tar.xz
cd07d63afd4136c141d49b74f92e1247  test.tar.xz
$ sudo podman build --file Dockerfile -t podman-test .
STEP 1: FROM docker.io/busybox
STEP 2: ADD test.tar.xz /
--> Using cache 2cab1dafa7b9b6bb91760fd2325da32df0d7846e02798739049026d3ba92e2de
STEP 3: COMMIT podman-test
2cab1dafa7b9b6bb91760fd2325da32df0d7846e02798739049026d3ba92e2de
```

**Describe the results you expected:**
It is expected not to reuse the cache if the files have changed.

**Version:**
```
podman/bionic,now 1.6.1-1~ubuntu18.04~ppa3 amd64 [installed]
Version:            1.6.1
RemoteAPI Version:  1
Go Version:         go1.10.4
OS/Arch:            linux/amd64
BuildahVersion: 1.11.2
kernel: 5.0.0-31-generic
arch: amd64
runc version 1.0.0-rc8+dev
```

TomSweeneyRedHat:
First reported in https://github.com/containers/libpod/issues/3920.  See that issue for further background information.  We'll track the issue here going forward though.

TomSweeneyRedHat:
@shaicoleman I'm replicating now with upstream and latest.  As you noted it should not cache.  As a work around, you can add the `--no-cache` option to the second build command, that should work as you expect.
```
[...]

PR #1792, opened 2019-08-14
Use content digests in ADD/COPY history entries

nalind:
Use digests of the added content in history entries that we create for ADD and COPY instructions, tightening up cache checking just a little bit more.  This should fix #1780.

PR #1925, opened 2019-10-16
Re-Add file check to cacheing algorithm

TomSweeneyRedHat:
Add back the file checking mechanisms that were removed as part of the
reworking of the cache in #1792.  If we have a 'COPY mytar.tar.xz' command
in the Containerfile and the tar file was regenerated after an intial build,
secondary 'buildah bud' commands would use the cache from the first build
for the tar file instead of regenerating it.

The historyMatches() function (https://github.com/containers/buildah/blob/master/imagebuildah/executor.go#L287)
does not seem to pick up this type of change and I wasn't able to quickly
rework that to do so.  We may want to revert back to this check for now,
then tweak the historyMatches to work properly in a follow up PR.

Regardless, I've added a test as part of this PR to catch this issue
going forward.

FWIW, with the exception of the test code, all of the code is a reversion to pre-existing code.

Fixes:  #1906


rhatdan:
@TomSweeneyRedHat Tests are not happy.

TomSweeneyRedHat:
Ugh, I swear I spend more time making the tests work in the CI than I do the actual code changes sometimes.  This worked fine in my sandbox, but it looks like an environment variable resolution issue in the CI.  I'll go hard code it, thanks for the heads up @rhatdan

TomSweeneyRedHat:
I take the variable bit back, it looks like a git hiccup, it's not seeing my new bud/change-archive directory.  Gotta dig.

TomSweeneyRedHat:
OMG, second day in a row with a corrupted sandbox.  Will have to spin up a new PR for this.

PR #1929, opened 2019-10-16
Re-Add file check to caching algorithm

TomSweeneyRedHat:
Try 2 after I accidentally nuked my sandbox for #1925

Add back the file checking mechanisms that were removed as part of the
reworking of the cache in #1792.  If we have a 'COPY mytar.tar.xz' command
in the Containerfile and the tar file was regenerated after an intial build,
secondary 'buildah bud' commands would use the cache from the first build
for the tar file instead of regenerating it.

The historyMatches() function (https://github.com/containers/buildah/blob/master/imagebuildah/executor.go#L287)
does not seem to pick up this type of change and I wasn't able to quickly
rework that to do so.  We may want to revert back to this check for now,
then tweak the historyMatches to work properly in a follow up PR.

Regardless, I've added a test as part of this PR to catch this issue
going forward.

FWIW, with the exception of the test code, all of the code is a reversion to pre-existing code.

Fixes:  #1906

### DIFF
--- a/imagebuildah/chroot_symlink_linux.go
+++ b/imagebuildah/chroot_symlink_linux.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/containers/storage/pkg/reexec"
 	"github.com/pkg/errors"
@@ -14,11 +15,13 @@ import (
 
 const (
 	symlinkChrootedCommand = "chrootsymlinks-resolve"
+	symlinkModifiedTime    = "modtimesymlinks-resolve"
 	maxSymlinksResolved    = 40
 )
 
 func init() {
 	reexec.Register(symlinkChrootedCommand, resolveChrootedSymlinks)
+	reexec.Register(symlinkModifiedTime, resolveSymlinkTimeModified)
 }
 
 // resolveSymlink uses a child subprocess to resolve any symlinks in filename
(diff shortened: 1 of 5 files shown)


TomSweeneyRedHat:
@rhatdan, scrozzled the first PR and abandoned it.  Created here and have green test buttons.

rhatdan:
:+1: 
@nalind @giuseppe @vrothberg PTAL

rhatdan:
@shaicoleman could you try this out and make sure it fixes your issue?

shaicoleman:
@rhatdan, there isn't an easy way for me to test at the moment.
Getting #4099 sorted would help me it easier for me to test.

tests/bud.bats
```diff
@@ -1797,3 +1797,29 @@ load helpers
...
+  expect_output --substring "COMMIT copy-archive"
+
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json --layers -t ${target} ${TESTSDIR}/bud/${target}
+  expect_output --substring " Using cache"
+  expect_output --substring "COMMIT copy-archive"
+
+  # Now test that we do NOT use cache if the tar file changes
+  date>bud/${target}/test; tar cJf - bud/${target}/test > bud/${target}/test.tar.xz
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json --layers -t ${target} ${TESTSDIR}/bud/${target}
+  expect_output --substring "COMMIT copy-archive"
```
edsantiago: Am I missing something obvious? Why is this step (1813-1815) necessary?

tests/bud.bats
```diff
@@ -1797,3 +1797,29 @@ load helpers
...
+
+  date>bud/${target}/test; tar cJf - bud/${target}/test > bud/${target}/test.tar.xz
```
edsantiago: I know this is so unlikely as to be ridiculous, but what if one day we have ultrafast machines that run 1804-1817 in less than one second? How about `>>` (append) instead of `>`?

tests/bud.bats
```diff
@@ -1797,3 +1797,29 @@ load helpers
   buildah rm --all
   buildah rmi --all --force
 }
+
+@test "bud containerfile with tar archive in copy" {
+  # First check to verify cache is uses if the tar file does not change
+  target=copy-archive
+  date>bud/${target}/test; tar cJf - bud/${target}/test > bud/${target}/test.tar.xz
```
edsantiago: As a general rule I find `tar cf DESTFILE PATH` more readable than `tar cf - PATH > DESTFILE`, but that's a minor nit and not worth resubmitting for. (If you have a strong preference for the redirection form, I'm willing to hear reasons for it though).

imagebuildah/chroot_symlink_linux.go
```diff
@@ -68,6 +71,118 @@ func resolveChrootedSymlinks() {
...
+		// In that case just return nil and continue on with using the cached image for the whole build process.
+		if info == nil {
+			return nil
+		}
+		modTime := info.ModTime()
+		if info.Mode()&os.ModeSymlink == os.ModeSymlink {
+			// Evaluate any symlink that occurs to get updated modified information
+			resolvedPath, err := filepath.EvalSymlinks(path)
+			if err != nil && os.IsNotExist(err) {
+				return errors.Wrapf(errDanglingSymlink, "%q", path)
```
giuseppe: wouldn't be safer to ignore dangling symlinks here (instead of line 178-180) so that we can process more entries?  If we break the loop early we might miss modified files

TomSweeneyRedHat:
Replaced by #1955, closing

PR #1935, opened 2019-10-21
BATS tests - lots of mostly minor cleanup

edsantiago:
First, and possibly most controversial, remove --debug flag.
Output on test failure is painful to read. It is unlikely
(but not inconceivable) that --debug will be of any use.

Second, fix a failure message so it properly uses %s instead
of string interpolation. (If the test name itself includes
a percent sign, we get undefined behavior on output).

Fix a few instances of 'buildah' to be 'run_buildah'.

Remove some unnecessary 'run/echo/check-status' sequences,
sticking with the much cleaner 'test'. When BATS shows
an error of the form 'this failed: [ $status -eq 0 ]'
I weep in despair.

### DIFF
--- a/tests/add.bats
+++ b/tests/add.bats
@@ -113,13 +113,13 @@ load helpers
   run_buildah add $cid ${TESTDIR}/distutils.cfg /usr/lib/python3.7/distutils
   run_buildah run $cid stat -c "%a" /usr/lib/python3.7/distutils
   root=$(buildah mount $cid)
-  expect_output --substring "755"
+  expect_output "644"
   run_buildah commit --signature-policy ${TESTSDIR}/policy.json $cid containers-storage:${imgName}
   run_buildah rm $cid
 
   newcid=$(buildah from --signature-policy ${TESTSDIR}/policy.json ${imgName})
   run_buildah run $newcid stat -c "%a" /usr/lib/python3.7/distutils
-  expect_output --substring "755"
+  expect_output "644"
   run_buildah rm $newcid
 }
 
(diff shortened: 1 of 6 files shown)


edsantiago:
I don't think this is going to pass tests: I think there's some line-count tests that are failing due to removing --debug, but I can't be sure because there are other problems affecting my test runs (i.e. buildah not working at all on f31/rawhide, and f30 version being incompatible with current tests). And my VM died and I won't be able to get back to this next week. So let's just see what CI has to say, and I'll take a fresh look at this next week.

(And, if there's strong objection to my removing `--debug`, I'd rather find out earlier than later)

TomSweeneyRedHat:
@edsantiago thanks for the PR.  I didn't dive in deeply as the CI tests failed as you expected.  However in general the changes LGTM.  I'm fine with the --debug removal.  I've actually struggled a few times with too much output from a failing test during development.

rhatdan:
@edsantiago Still working on this?

edsantiago:
Oh yes. It's slow and painstaking but I've fixed several issues including a bug in the tests themselves. Still not running clean all the way through.

edsantiago:
Whew - tests are finally green. It is quite a bit changed since the first iteration; most of the new work is in separate commits.

vrothberg:
@giuseppe @TomSweeneyRedHat @rhatdan PTAL

giuseppe:
let's ship it!

LGTM

@rh-atomic-bot r+

edsantiago:
Test failure looks like a transient:
```fatal: unable to access 'https://gitlab.gnome.org/GNOME/libglnx.git/': Failed to connect to gitlab.gnome.org port 443: Connection refused
```
/retest

TomSweeneyRedHat:
Yeah, homu's been painful the past day or three.  Let me poke around.

TomSweeneyRedHat:
LGTM, thanks a bunch @edsantiago, excellent work as usual.

TomSweeneyRedHat:
@rh-atomic-bot retry

PR #1937, opened 2019-10-22
Enable --device directory as src device

QiWang19:
Enables --device accepte directory path as source device. Add the devices under the source directory to the destination directory.

complete card test criteria: https://jira.coreos.com/browse/RUN-497
related podman issue: https://github.com/containers/libpod/issues/2380

### DIFF
--- a/pkg/parse/parse_test.go
+++ b/pkg/parse/parse_test.go
@@ -74,17 +74,22 @@ func TestDeviceFromPath(t *testing.T) {
 	// Path is valid
 	dev, err := DeviceFromPath("/dev/null")
 	assert.NoError(t, err)
-	assert.Equal(t, dev.Major, int64(1))
-	assert.Equal(t, dev.Minor, int64(3))
-	assert.Equal(t, dev.Permissions, "rwm")
-	assert.Equal(t, dev.Uid, uint32(0))
-	assert.Equal(t, dev.Gid, uint32(0))
+	assert.Equal(t, len(dev), 1)
+	assert.Equal(t, dev[0].Major, int64(1))
+	assert.Equal(t, dev[0].Minor, int64(3))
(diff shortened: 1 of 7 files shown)


QiWang19:

TomSweeneyRedHat:
We should have some kind of change to the man page(s) too.

pkg/parse/parse_test.go
```diff
@@ -74,17 +74,18 @@ func TestDeviceFromPath(t *testing.T) {
...
 
 	// Path does not exists
 	_, err = DeviceFromPath("/dev/BOGUS")
 	assert.Error(t, err)
 
-	// Path exists but is not a device
+	// Path is a directory of devices
 	_, err = DeviceFromPath("/dev/pts")
-	assert.Error(t, err)
+	assert.NoError(t, err)
```
rhatdan: Should add another test for a file

 	_, err = DeviceFromPath("/etc/passwd") 	
        assert.Error(t, err)

QiWang19: test added

TomSweeneyRedHat:
@rh-atomic-bot retry

QiWang19:
does this error message mean some other tool does not support  adding a directory of devices?
https://travis-ci.org/containers/buildah/jobs/601935493#L1986

QiWang19:
@TomSweeneyRedHat Can I add another Dockerfile under tests/bud/device/? I want to try to see if 
ci fails because current Dockerfile does not work for my test.
git doesn't track the new file after I only add it under /tests/bud/device

QiWang19:
@TomSweeneyRedHat tests are green now. Can we get this in

TomSweeneyRedHat:
I'm jumping in after the barn door shut, but LGTM and nice job @QiWang19

PR #1939, opened 2019-10-22
Bump GoLang for CVE-2019-17596

TomSweeneyRedHat:
Bump golang to 1.12.12 and 1.13.X to avoid security issues
noted in CVE-2019-17596. See: https://groups.google.com/forum/#!msg/golang-announce/lVEm7llp0w0/VbafyRkgCgAJ and https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-17596

I've dropped 1.11 compiles and added 1.13.X while we're at it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ language: go
 dist: xenial
 sudo: required
 go:
-    - 1.11.x
-    - 1.12.x
+    - 1.12.12
+    - 1.13.x
     - tip
 go_import_path: github.com/containers/buildah
 


TomSweeneyRedHat:
@vrothberg  or @nalind.  I've added go 1.13.x to the travis build for the first time and I'm getting a tree dirty error?    Any thoughts on how to make it happy?

```
go version go1.13.3 linux/amd64
go.env

$ go env
$ make vendor
$ ./hack/tree_status.sh
tree is dirty, please commit all changes and sync the vendor.conf
```

vrothberg:
@TomSweeneyRedHat, we should make it a PS in today's meeting to discuss a way forward. The problem we're hitting here is that go 1.13 starts to remove some obsolete files in ./vendor that previous versions did not remove.

Ultimately, I believe that we should have only _one_ CI job for vendoring using go 1.13. We somehow need to move to it at some point. However, as soon as we do that, contributors with go < 1.13 will face the problem that their `make vendor` will yield different results than what our CI wants. Hence, we could add a `make vendor-in-container` that executes `make vendor` in a `golang:1.13` container.

@rhatdan ^ hit the same issue last week.

rhatdan:
Yes I think
make vendor-in-container 
is the way forward.  Then at least it is easy to get contributors to do the work.
The problem would be can we do this all in Buildah, or would we be requiring buildah contributors to install podman.

vrothberg:
> Yes I think
> make vendor-in-container
> is the way forward. Then at least it is easy to get contributors to do the work.
> The problem would be can we do this all in Buildah, or would we be requiring buildah contributors to install podman.

I'd default to use `podman` but we could make it configurable with a `CTR_ENGINE` env variable. If users want to use buildah, they could do that on their own but it's not as straight-forward doing a `podman run ...`.

rhatdan:
We need make vendor-in-container PR to fix this up.

TomSweeneyRedHat:
Replace by #1949

Issue #1934, opened 2019-10-21
issue building from branch of git repo context


dustymabe:
I'm having an issue trying to specify a branch to build from like described [here](https://docs.docker.com/engine/reference/commandline/build/) under **Build Syntax Suffix**.

Here is an example where I am trying to build from the `dusty` branch: 

```
[vagrant@vanilla-f31 ~]$ sudo /usr/bin/podman build -t simple-kmod-a8c6682:5.3.7-300.fc31.x86_64 --label=name=simple-kmod --build-arg KVER=5.3.7-300.fc31.x86_64 --build-arg KMODVER=a8c6682 git://github.com/dustymabe/kvc-simple-kmod#dusty
Error: error prepping temporary context directory: exit status 128
[vagrant@vanilla-f31 ~]$ 
[vagrant@vanilla-f31 ~]$ sudo /usr/bin/podman build -t simple-kmod-a8c6682:5.3.7-300.fc31.x86_64 --label=name=simple-kmod --build-arg KVER=5.3.7-300.fc31.x86_64 --build-arg KMODVER=a8c6682 https://github.com/dustymabe/kvc-simple-kmod#dusty
Error: no FROM statement found
```

Sorry for being brief but I don't have much time right now to provide all the information needed.

**Version Info**

```
[vagrant@vanilla-f31 ~]$ rpm -q podman
podman-1.6.2-2.fc31.x86_64
[vagrant@vanilla-f31 ~]$ sudo podman version
Version:            1.6.2
RemoteAPI Version:  1
Go Version:         go1.13.1
OS/Arch:            linux/amd64

```

**Output of `podman info`:**

<details>

```
[vagrant@vanilla-f31 ~]$ sudo podman info
host:
  BuildahVersion: 1.11.3
  CgroupVersion: v2
  Conmon:
    package: conmon-2.0.1-1.fc31.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.1, commit: 5e0eadedda9508810235ab878174dca1183f4013'
  Distribution:
    distribution: fedora
    version: "31"
  MemFree: 3720368128
  MemTotal: 4126642176
  OCIRuntime:
    name: crun
    package: crun-0.10.2-1.fc31.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 0.10.2
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +YAJL
  SwapFree: 0
[...]

PR #1941, opened 2019-10-22
Fix git build with branch specified

TomSweeneyRedHat:
If the context directory was specified as a git repo with a
branch reference ala

```
buildah bud --layers -t test git://github.com/containers/skopeo#master
```

The internal `git clone` command executed by buildah would fail as the
`#master` branch specification needed to be removed and specified with
the `-b` option like:

```
git clone -b master git://github.com/containers/skopeo /var/tmp/buildah12332
```
rather than:
```
git clone git://github.com/containers/skopeo#master /var/tmp/buildah12332
```

Addresses #1934

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -1797,3 +1797,9 @@ load helpers
   buildah rm --all
   buildah rmi --all --force
 }
+
+@test "bud using gitrepo and branch" {
+  target=gittarget
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json --layers -t ${target} git://github.com/containers/BuildSourceImage#master
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json --layers -t ${target} git://github.com/containers/BuildSourceImage
+}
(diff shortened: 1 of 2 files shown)


tests/bud.bats
```diff
@@ -1797,3 +1797,10 @@ load helpers
   buildah rm --all
   buildah rmi --all --force
 }
+
+@test "bud using gitrepo and branch" {
+  target=gittarget
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json --layers -t ${target} git://github.com/containers/skopeo 
+
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json --layers -t ${target} git://github.com/containers/skopeo#master
```
TomSweeneyRedHat: I'm leery about adding these tests.  They're probably going to be close to timing out as I couldn't find a smaller repo to test against (happy to take suggestions), and if they complete, will add 10+ minutes to the test runs.

nalind:
LGTM.  github.com/openshift/ruby-hello-world _might_ be faster, but I haven't clocked them.

rhatdan:
BuildSourceImage
Takes < 1.5 minutes

TomSweeneyRedHat:
BuildSourceImage is the winner.  Reliable and the fastest of the three.  Thanks for the suggestions!

rhatdan:
@rh-atomic-bot r+

PR #1949, opened 2019-10-28
Add support for make vendor-in-container

rhatdan:
Certain versions of go have different vendoring capabilities.
We want to make sure that vendoring the the vendor checker are
running with the same version of golang.  We do this by using
podman to run a container of the lates golang container image.


Makefile
```diff
@@ -130,6 +130,9 @@ test-unit: tests/testreport/testreport
 	mkdir -p $$tmp/root $$tmp/runroot; \
 	$(GO) test -v -tags "$(STORAGETAGS) $(SECURITYTAGS)" ./cmd/buildah -args -root $$tmp/root -runroot $$tmp/runroot -storage-driver vfs -signature-policy $(shell pwd)/tests/policy.json -registries-conf $(shell pwd)/tests/registries.conf
 
+vendor-in-container:
+	podman run --privileged --rm --env HOME=/root -v `pwd`:/src -w /src golang make vendor
```
vrothberg: Currently, we're using the latest tag which is likely to break us in the future again when go 1.14+ introduces yet another change. It also seems that the Ubuntu package isn't installing registries.conf (at least that's what the error message in the Travis indicates). So instead of `golang`, we could use `docker.io/library/golang:1.13`.

Ultimately, I would prefer using Cirrus for these checks and remove the vendor checks from Travis altogether. This will prevent us from running those checks redundantly on all jobs. In Cirrus, we can specify a job running in a `golang:1.13` container, so there would be no need to install podman.

rhatdan:
@vrothberg @TomSweeneyRedHat @giuseppe @QiWang19 @nalind PTAL

rhatdan:
@cevich PTAL

vrothberg:
Out of curiosity. Are we planning to ditch the remaining Travis jobs or do we need them for a specific reason?

rhatdan:
I want to talk to @nalind about it.  Make sure we have all of the tests running on all the platforms in Cirrus.

TomSweeneyRedHat:
Replaces #1939

TomSweeneyRedHat:
LGTM, would like a @cevich head nod and a @nalind one if he's time.

rhatdan:

cevich:
looking...

.cirrus.yml
```diff
@@ -75,9 +75,37 @@ testing_task:
...
+# that the vendor.conf, the code and the vendored packages in ./vendor are
+# in sync at all times.
+vendor_task:
+
+    only_if: $CIRRUS_CHANGE_MESSAGE !=~ '.*\*\*\*\s*CIRRUS:\s*TEST\s*IMAGES\s*\*\*\*.*'
+
+    env:
+        CIRRUS_WORKING_DIR: "/var/tmp/go/src/github.com/containers/buildah"
+        GOPATH: "/go"
+        GOSRC: "/go/src/github.com/containers/buildah"
```
cevich: It would be good for these to all align. Esp. `$CIRRUS_WORKING_DIR` since that's where the code gets checked out.  I recommend we use `/var/tmp/go` as `${GOPATH}` because it matches what we do in podman and makes doing rootless testing (in the future) MUCH easier.  Also if we ever need to test in atomic or silverblue, messing with any top-level subdirs of `/` is a huge "no-no".  So this would be preferred:

```yaml
env:
￼         GOPATH: "/var/tmp/go"
￼         GOSRC: "${GOPATH}/src/github.com/containers/buildah"
         CIRRUS_WORKING_DIR: "${GOSRC}"
```

.cirrus.yml
```diff
@@ -75,9 +75,37 @@ testing_task:
...
+
+    # Runs within Cirrus's "community cluster"
+    container:
+        image: docker.io/library/golang:1.13
+        cpu: 1
+        memory: 1
```
cevich: If this task runs too slowly, there's no harm in bumping CPU and Memory upward.  I believe the default is 8 and 12, so there's quite a bit of headroom available.

.cirrus.yml
```diff
@@ -75,9 +75,37 @@ testing_task:
...
+
+    timeout_in: 30m
+
+    vendor_script:
+        - 'cd ${CIRRUS_WORKING_DIR} && make vendor'
+        - 'cd ${CIRRUS_WORKING_DIR} && ./hack/tree_status.sh'
```
cevich: Once the `env` variables all agree, the `cd` above will be redundant.  Cirrus-CI will automatically execute from that directory by default.

.cirrus.yml
cevich: Note: the original path mismatch is an artifact of podman's `quay.io/libpod/gate:latest` (which has a special entrypoint script).  Here, for `docker.io/library/golang:1.13` it can be simplified as above.

PR #1950, opened 2019-10-28
imgtype: exit with error if storage fails

edsantiago:
'imgtype' on my system has been exiting zero regardless of
the command being run; looking at logs shows:

   error opening storage: vfs driver does not support overlay.mountopt options

(but exiting zero). Tests which rely on exit status zero have
been passing, but without actually testing anything.


TomSweeneyRedHat:
LGTM assuming happy tests.  Thanks @edsantiago

rhatdan:
@rh-atomic-bot r+

PR #1953, opened 2019-10-29
[WIP] add global option --containers-conf for receiving containers.conf path

QiWang19:
Add global option --containers-conf for the user to config the path of the containers.conf file.
Using containers/common v0.0.2 for containers.conf.
Vendor unshare package from containers/common
Also, we were never handling adding and dropping of all capabilities. Added support for --cap-add all and cap-drop all.


rhatdan:
Needs a rebase.

QiWang19:
@rhatdan Tests are failing because the config pkg validates the default hooks.d  but there's  no `/usr/share/containers/oci/hooks.d` in the test environment. Influences all the commands try to get default cconfig.  how to fix this?

rhatdan:
Can we turn off the validation.  I think if the hooks directory does not exists, we should just treat it the same as if it was empty.   Is this because we check the option?

QiWang19:
> Can we turn off the validation. I think if the hooks directory does not exists, we should just treat it the same as if it was empty. Is this because we check the option?

This is because we check the config in containers/common. We can turn off the validation in containers/common.

I should test this before I release v0.0.2 :disappointed:  Now the release doesn't work. Can I delete the release...?

rhatdan:
Nope but we got plenty of integers,  Just make 0.0.3.

QiWang19:
@rhatdan @vrothberg PTAL

rhatdan:
Does this work with the No containers.conf file at all?
Does this work with containers.conf file and everything commented out?

rhatdan:
Please remove the [WIP] from your PRs the [WIP] in github is enough.

rhatdan:
Lets merge all the PRs together.
This is looking good. 
I think the next step is to do a full review of containers.conf and make sure we can handle merging correctly.

We need to handle 
/usr/share/containers/containers.conf
/etc/containers/containers.conf
Merge these together if both exists.
~/.config/containers/containers.conf (For rootless)

Merge all three together if rootless.

rhatdan:
@QiWang19 needs a rebase.

rhatdan:
@QiWang19 needs a rebase.

PR #1956, opened 2019-10-31
Add disableFips option to secrets pkg

umohnani8:
If disableFips is set, then we don't mount the FIPS
secret even if the machine is in FIPs mode. This is
to help users run workloads that are not fips compliant
in openshift even if their machine is in FIPs mode.
This is needed in CRI-O.

After this is in, I will vendor it into libpod and then cri-o.

### DIFF
--- a/pkg/secrets/secrets.go
+++ b/pkg/secrets/secrets.go
@@ -148,12 +148,12 @@ func getMountsMap(path string) (string, string, error) {
 }
 
 // SecretMounts copies, adds, and mounts the secrets to the container root filesystem
-func SecretMounts(mountLabel, containerWorkingDir, mountFile string, rootless bool) []rspec.Mount {
-	return SecretMountsWithUIDGID(mountLabel, containerWorkingDir, mountFile, containerWorkingDir, 0, 0, rootless)
+func SecretMounts(mountLabel, containerWorkingDir, mountFile string, rootless, disableFips bool) []rspec.Mount {
+	return SecretMountsWithUIDGID(mountLabel, containerWorkingDir, mountFile, containerWorkingDir, 0, 0, rootless, disableFips)
 }
 
 // SecretMountsWithUIDGID specifies the uid/gid of the owner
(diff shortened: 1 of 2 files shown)


umohnani8:
@mrunalp @rhatdan @TomSweeneyRedHat  PTAL

pkg/secrets/secrets.go
```diff
@@ -180,16 +180,19 @@ func SecretMountsWithUIDGID(mountLabel, containerWorkingDir, mountFile, mountPre
 		}
 	}
 
-	// Add FIPS mode secret if /etc/system-fips exists on the host
-	_, err := os.Stat("/etc/system-fips")
-	if err == nil {
-		if err := addFIPSModeSecret(&secretMounts, containerWorkingDir, mountPrefix, mountLabel, uid, gid); err != nil {
-			logrus.Errorf("error adding FIPS mode secret to container: %v", err)
+	// Only add FIPS secret mount if the label FIPS_DISABLE=true is not set in the pod/container
+	if !disableFips {
```
vrothberg: Nit: A positive check would simplify the code a bit more.
```golang
if disableFips {
    return secretMounts
}
_, err := os.Stat("/etc/system-fips")
...
```

TomSweeneyRedHat:
LGTM, @umohnani8 let me know if you want to switch per @vrothberg's suggestion or not.

pkg/secrets/secrets.go
umohnani8: Fixed. Thanks for the pointer :)

umohnani8:
@TomSweeneyRedHat should be ready once tests are green

rhatdan:
We probably want to add the capability a --security fips:disable flag for podman and buildah, so that users of these tools can do this also.

@rh-atomic-bot r+

umohnani8:
yup, we pan on doing that. Just need this in cri-o urgently first

rhatdan:
This seems to be hung, going to force merge so we can move forward on CRI-O since it has passed all tests.

Issue #1675, opened 2019-06-14
podman-docker's "docker build --pull=false ..." behaves differently than in docker


vorburger:
**Description**

It appears that a `docker build --pull=false ...` behaves differently in `podman-docker` than in "real" docker; see https://github.com/kubernetes/kubernetes/issues/79048 for an analysis.

**Steps to reproduce the issue:**

1. Fedora 30 Server
1. sudo dnf install podman-docker
1. git clone https://github.com/kubernetes/kubernetes.git ; cd kubernetes
1. build/run.sh make kubectl

fails on Buildah in Podman saying `Error: error creating build container: no such image "containers-storage:[overlay@/home/vorburger/.local/share/containers/storage+/tmp/1000:overlay.mount_program=/usr/bin/fuse-overlayfs]k8s.gcr.io/kube-cross:v1.12.5-1"` but (apparently) works in regular docker.

**Output of `podman version`:**

```
$ podman version
Version:            1.3.1
RemoteAPI Version:  1
Go Version:         go1.12.2
OS/Arch:            linux/amd64
```

**Output of `podman info --debug`:**

```
$ podman info --debug
debug:
  compiler: gc
  git commit: ""
  go version: go1.12.2
  podman version: 1.3.1
host:
  BuildahVersion: 1.8.2
  Conmon:
    package: podman-1.3.1-1.git7210727.fc30.x86_64
    path: /usr/libexec/podman/conmon
    version: 'conmon version 1.12.0-dev, commit: c9a4c48d1bff85033b7fc9b62d25961dd5048689'
  Distribution:
    distribution: fedora
    version: "30"
  MemFree: 13008052224
  MemTotal: 33585057792
  OCIRuntime:
    package: runc-1.0.0-93.dev.gitb9b6cc6.fc30.x86_64
    path: /usr/bin/runc
    version: |-
      runc version 1.0.0-rc8+dev
      commit: e3b4c1108f7d1bf0d09ab612ea09927d9b59b4e3
      spec: 1.0.1-dev
  SwapFree: 16859000832
  SwapTotal: 16859000832
  arch: amd64
  cpus: 8
  hostname: toby
  kernel: 5.1.6-300.fc30.x86_64
  os: linux
  rootless: true
  uptime: 153h 43m 38.99s (Approximately 6.38 days)
registries:
  blocked: null
  insecure: null
  search:
  - docker.io
  - registry.fedoraproject.org
  - quay.io
  - registry.access.redhat.com
[...]

PR #1959, opened 2019-10-31
Fix --pull=true||false and add --pull-never to bud and from (retry)

TomSweeneyRedHat:
(Replaces #1873 as it had lint issues that were timing out tests that I couldn't
track down easily)

Prior to this fix, if someone did `buildah bud --pull=false .` and the image in
the Containerfile's FROM statement was not local, the build would fail. The same
build on Docker will succeed. In Docker, when `--pull` is set to false, it only
pulls the image from the registry if there was not one locally. Buildah would never
pull the image and if the image was not locally available, it would throw an error.
In certain Kubernetes environments, this was especially troublesome.

To retain the old `--pull=false` functionality, I've created a new `--pull-never`
option that fails if an image is not locally available just like the old
`--pull=false` option used to do.

In addition, if there was a newer version of the image on the repository than
the one locally, the `--pull=true` option would not pull the image as it should
have, this corrects that.

Changes both the from and bud commands.

Addresses: #1675

### DIFF
--- a/tests/basic.bats
+++ b/tests/basic.bats
@@ -3,11 +3,11 @@
 load helpers
 
 @test "from" {
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json alpine)
   run_buildah rm $cid
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json scratch)
   run_buildah rm $cid
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json --name i-love-naming-things alpine)
+  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json --name i-love-naming-things alpine)
   run_buildah rm i-love-naming-things
 }
 
@@ -24,11 +24,11 @@ load helpers
 }
 
 @test "from-nopull" {
-  run_buildah 1 from --pull=false --signature-policy ${TESTSDIR}/policy.json alpine
+  run_buildah 1 from --pull-never --signature-policy ${TESTSDIR}/policy.json alpine
 }
 
 @test "mount" {
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json scratch)
   root=$(buildah mount $cid)
   run_buildah unmount $cid
   root=$(buildah mount $cid)
@@ -38,7 +38,7 @@ load helpers
 }
 
 @test "by-name" {
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json --name scratch-working-image-for-test scratch)
+  cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json --name scratch-working-image-for-test scratch)
   root=$(buildah mount scratch-working-image-for-test)
   run_buildah unmount scratch-working-image-for-test
   run_buildah rm scratch-working-image-for-test
@@ -48,7 +48,7 @@ load helpers
   createrandom ${TESTDIR}/randomfile
   createrandom ${TESTDIR}/other-randomfile
 
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json scratch)
   root=$(buildah mount $cid)
   cp ${TESTDIR}/randomfile $root/randomfile
   run_buildah unmount $cid
--- a/tests/config.bats
+++ b/tests/config.bats
@@ -30,7 +30,7 @@ function check_matrix() {
 }
 
 @test "config entrypoint using single element in JSON array (exec form)" {
-  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json scratch)
   buildah config --entrypoint '[ "/ENTRYPOINT" ]' $cid
   buildah commit --format docker --signature-policy ${TESTSDIR}/policy.json $cid scratch-image-docker
   buildah commit --format oci --signature-policy ${TESTSDIR}/policy.json $cid scratch-image-oci
@@ -42,7 +42,7 @@ function check_matrix() {
 }
 
 @test "config entrypoint using multiple elements in JSON array (exec form)" {
-  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json scratch)
   buildah config --entrypoint '[ "/ENTRYPOINT", "ELEMENT2" ]' $cid
   buildah commit --format docker --signature-policy ${TESTSDIR}/policy.json $cid scratch-image-docker
   buildah commit --format oci --signature-policy ${TESTSDIR}/policy.json $cid scratch-image-oci
@@ -54,7 +54,7 @@ function check_matrix() {
 }
 
 @test "config entrypoint using string (shell form)" {
-  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json scratch)
   buildah config --entrypoint /ENTRYPOINT $cid
   buildah commit --format docker --signature-policy ${TESTSDIR}/policy.json $cid scratch-image-docker
   buildah commit --format oci --signature-policy ${TESTSDIR}/policy.json $cid scratch-image-oci
@@ -66,7 +66,7 @@ function check_matrix() {
 }
 
 @test "config set empty entrypoint doesn't wipe cmd" {
-  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json scratch)
   buildah config --cmd "command" $cid
   buildah config --entrypoint "" $cid
   buildah commit --format docker --signature-policy ${TESTSDIR}/policy.json $cid scratch-image-docker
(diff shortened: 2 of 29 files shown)


TomSweeneyRedHat:
Replaces #1873.  Same code as that one except I've changed the test to use busybox instead of alpine.  #1873 kept timing out with test and warning about lint issues in 3 files that it didn't name.  I've cut/pasted the changes here after rebasing from upstream in hopes that will cure the mystery lint issues and the tests will complete in the time required by the CI.

rhatdan:
Still failing, I think this is a real bug, where pull is hanging perhaps?

TomSweeneyRedHat:
I'm not seeing a difference in the pull times when running the tests by hand.  I'm still seeing that lint warning on 3 files that it's not naming.  Want to try and see if I can figure that out.  What's driving me daffy is I'd green buttons on this before I put a `/hold` on it.

TomSweeneyRedHat:
I suspect the new code that we've added to inspect the image may be taking too long for travis ci.  Prior to this change, the ci was taking close to 45 minutes to complete.  With all the pulls that we do throughout the code, I think hitting up the inspect code was slowing it down past the 50 minute mark which we can't change (Travis limitation).  I've changed the code to do the inspect only if we have a --pull=true.  I'm not sure that will be enough as we have a lot of those scattered throughout all of the tests.

rhatdan:
Can we just increase the travis timeout?

TomSweeneyRedHat:
We can't increase the travis time out unless we pay for it.  The "free" Travis has an embedded 50 minute time out that you can't work around.

rhatdan:
Well I am hoping to get off of travis when we finish migrating to Cirrus.

TomSweeneyRedHat:
Green with 5 seconds to spare.  I've a thought for whittling the time down further.  Will implement in the morning.

TomSweeneyRedHat:
I've just changed a little over a dozen of the tests.  The main change there was changing `--pull` to `--pull=false`.  This way if the image isn't already locally available (usually alpine or busybox), we go ahead and pull it.  With the new

TomSweeneyRedHat:
OK, I've changed most but not all of the `--pull` options in the tests to `--pull=false`.  That way we won't hit up the repo to check the date.  If we have the image we'll do nothing, if we don't , we'll pull it.  

I also reworked the logic in new.go so that it now only checks the repo for the date of the image if the image is found locally.  Prior we always did that date check which really wasn't needed if we didn't have the image locally.  Between the two changes we should be much closer time wise for test completion than we were prior.

TomSweeneyRedHat:
Happy green test buttons on this, and tests completing in just over 45 minutes on average.  FWIW, the changes since the last review was to new.go and how/when it does the repo lookup, that's been minimalized, then changing a lot but not all of the `--pull` options to `--pull=false` through  out the tests, especially the tests other than build or from.

tests/basic.bats
```diff
@@ -24,11 +24,11 @@ load helpers
 }
 
 @test "from-nopull" {
-  run_buildah 1 from --pull=false --signature-policy ${TESTSDIR}/policy.json alpine
+  run_buildah 1 from --pull-never --signature-policy ${TESTSDIR}/policy.json alpine
 }
 
 @test "mount" {
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
```
rhatdan: nit: ditto on scratch.

tests/basic.bats
```diff
@@ -3,11 +3,11 @@
 load helpers
 
 @test "from" {
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json alpine)
   run_buildah rm $cid
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
```
rhatdan: nit: Probably should not be a --pull at all, since  you are using scratch.

tests/basic.bats
```diff
@@ -38,7 +38,7 @@ load helpers
 }
 
 @test "by-name" {
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json --name scratch-working-image-for-test scratch)
+  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json --name scratch-working-image-for-test scratch)
```
rhatdan: ditto

tests/basic.bats
```diff
@@ -48,7 +48,7 @@ load helpers
   createrandom ${TESTDIR}/randomfile
   createrandom ${TESTDIR}/other-randomfile
 
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
```
rhatdan: ditto

tests/config.bats
```diff
@@ -30,7 +30,7 @@ function check_matrix() {
 }
 
 @test "config entrypoint using single element in JSON array (exec form)" {
-  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --pull-never --signature-policy ${TESTSDIR}/policy.json scratch)
```
rhatdan: ditto

tests/config.bats
```diff
@@ -42,7 +42,7 @@ function check_matrix() {
 }
 
 @test "config entrypoint using multiple elements in JSON array (exec form)" {
-  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --pull-never --signature-policy ${TESTSDIR}/policy.json scratch)
```
rhatdan: ditto

tests/config.bats
```diff
@@ -54,7 +54,7 @@ function check_matrix() {
 }
 
 @test "config entrypoint using string (shell form)" {
-  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --pull-never --signature-policy ${TESTSDIR}/policy.json scratch)
```
rhatdan: ditto

tests/config.bats
```diff
@@ -66,7 +66,7 @@ function check_matrix() {
 }
 
 @test "config set empty entrypoint doesn't wipe cmd" {
-  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --pull-never --signature-policy ${TESTSDIR}/policy.json scratch)
```
rhatdan: ditto

tests/config.bats
```diff
@@ -79,7 +79,7 @@ function check_matrix() {
 }
 
 @test "config entrypoint with cmd" {
-  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --pull-never --signature-policy ${TESTSDIR}/policy.json scratch)
```
rhatdan: ditto

tests/config.bats
```diff
@@ -107,7 +107,7 @@ function check_matrix() {
 }
 
 @test "config" {
-  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --pull-never --signature-policy ${TESTSDIR}/policy.json scratch)
```
rhatdan: ditto

tests/config.bats
```diff
@@ -210,7 +210,7 @@ function check_matrix() {
 }
 
 @test "remove configs using '-' syntax" {
-  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --pull-never --signature-policy ${TESTSDIR}/policy.json scratch)
```
rhatdan: ditto

tests/copy.bats
```diff
@@ -18,21 +18,21 @@ load helpers
   createrandom ${TESTDIR}/other-randomfile
   createrandom ${TESTDIR}/third-randomfile
 
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
```
rhatdan: ditto

tests/copy.bats
```diff
@@ -45,7 +45,7 @@ load helpers
...
```
rhatdan: ditto

tests/copy.bats
```diff
@@ -140,7 +140,7 @@ load helpers
   createrandom ${TESTDIR}/other-subdir/randomfile
   createrandom ${TESTDIR}/other-subdir/other-randomfile
 
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
+  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json alpine)
```
rhatdan: ditto

tests/copy.bats
```diff
@@ -163,7 +163,7 @@ load helpers
   createrandom ${TESTDIR}/randomfile
   ln -s ${TESTDIR}/randomfile ${TESTDIR}/link-randomfile
 
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
```
rhatdan: ditto

tests/images.bats
```diff
@@ -140,7 +140,7 @@ load helpers
 }
 
 @test "Test dangling images" {
-  cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json scratch)
+  cid=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json scratch)
```
rhatdan: ditto

TomSweeneyRedHat:
Updated with the test nits picked.

TomSweeneyRedHat:
Happy green test buttons.  Could use another LGTM, especially for the new.go change.  @QiWang19 @vrothberg @nalind any takers?

rhatdan:
Sadly @TomSweeneyRedHat needs a rebase.

TomSweeneyRedHat:
The PR that won't die!

TomSweeneyRedHat:
repushed.   tests/from.bats and tests/run.bats had merg conflicts that I've corrected.

rhatdan:
@rh-atomic-bot r+

PR #1960, opened 2019-10-31
Handle configuration blobs for manifest lists

nalind:
When the base image or an image that we're inspecting is a reference to a manifest list, resolve it to a runnable image instance, then try to read the configuration blob from the runnable image.

### DIFF
--- a/import.go
+++ b/import.go
@@ -5,8 +5,10 @@ import (
 
 	"github.com/containers/buildah/docker"
 	"github.com/containers/buildah/util"
+	"github.com/containers/image/v5/image"
 	"github.com/containers/image/v5/manifest"
 	is "github.com/containers/image/v5/storage"
+	"github.com/containers/image/v5/transports"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage"
 	digest "github.com/opencontainers/go-digest"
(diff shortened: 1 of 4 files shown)


import.go
```diff
@@ -28,12 +30,39 @@ func importBuilderDataFromImage(ctx context.Context, store storage.Store, system
...
+			return nil, errors.Wrapf(err, "error parsing image manifest for %q as list", transports.ImageName(ref))
+		}
+		instance, err := list.ChooseInstance(systemContext)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error finding an appropriate image in manifest list %q", transports.ImageName(ref))
+		}
+		instanceDigest = &instance
+	}
+
+	img, err := image.FromUnparsedImage(ctx, systemContext, image.UnparsedInstance(src, instanceDigest))
```
rhatdan: Your returning an img and then three lines later replacing it?

nalind: Not sure I follow.  This is mainly about making sure that `img` is pointing to a runnable image if the reference we started with included the digest of a manifest list or image index.  If it isn't a list, then `instanceDigest` is `nil` and we basically continue as we would have otherwise, but if it is, we use the `SystemContext` to choose an instance and attempt to use that one, instead.

rhatdan: You are not using the value returned above, then you replace it below.
It would be clearer if you did 
if _, err := image.FromUnparsedImage(ctx, systemContext, image.UnparsedInstance(src, instanceDigest)); err != nil

nalind: Ah.  That's a different variable, of a different type.  Renamed the new one.

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
@rh-atomic-bot r+

PR #1966, opened 2019-11-06
Add some tests and test steps for buildah

ypu:
Add several tests and also add some check points for exist test cases in buildah/tests

### DIFF
--- a/tests/add.bats
+++ b/tests/add.bats
@@ -140,3 +140,24 @@ load helpers
   expect_output "644"
   run_buildah rm $newcid
 }
+
+@test "add with chown" {
+  createrandom ${TESTDIR}/randomfile
+  cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json busybox)
+  run_buildah add --chown bin:bin $cid ${TESTDIR}/randomfile /tmp/random
+  run_buildah run $cid ls -l /tmp/random
+
+  expect_output --substring bin.*bin
+  buildah rm $cid
+}
+
+@test "add url" {
+  cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json busybox)
+  run_buildah add $cid https://github.com/containers/buildah/raw/master/README.md
+  run_buildah run $cid ls /README.md
+
+  run_buildah add $cid https://github.com/containers/buildah/raw/master/README.md /home
+  run_buildah run $cid ls /home/README.md
+
+  buildah rm $cid
+}
--- a/tests/add.bats
+++ b/tests/add.bats
@@ -108,36 +108,38 @@ load helpers
 @test "add single file creates absolute path with correct permissions" {
   imgName=ubuntu-image
   createrandom ${TESTDIR}/distutils.cfg
+  permission=$(stat -c "%a" ${TESTDIR}/distutils.cfg)
 
   cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json ubuntu)
   run_buildah add $cid ${TESTDIR}/distutils.cfg /usr/lib/python3.7/distutils
   run_buildah run $cid stat -c "%a" /usr/lib/python3.7/distutils
   root=$(buildah mount $cid)
-  expect_output "644"
+  expect_output $permission
   run_buildah commit --signature-policy ${TESTSDIR}/policy.json $cid containers-storage:${imgName}
   run_buildah rm $cid
 
   newcid=$(buildah from --signature-policy ${TESTSDIR}/policy.json ${imgName})
   run_buildah run $newcid stat -c "%a" /usr/lib/python3.7/distutils
-  expect_output "644"
+  expect_output $permission
   run_buildah rm $newcid
 }
 
 @test "add single file creates relative path with correct permissions" {
   imgName=ubuntu-image
   createrandom ${TESTDIR}/distutils.cfg
+  permission=$(stat -c "%a" ${TESTDIR}/distutils.cfg)
 
   cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json ubuntu)
   run_buildah add $cid ${TESTDIR}/distutils.cfg lib/custom
   run_buildah run $cid stat -c "%a" lib/custom
   root=$(buildah mount $cid)
-  expect_output "644"
+  expect_output $permission
   run_buildah commit --signature-policy ${TESTSDIR}/policy.json $cid containers-storage:${imgName}
   run_buildah rm $cid
 
   newcid=$(buildah from --signature-policy ${TESTSDIR}/policy.json ${imgName})
   run_buildah run $newcid stat -c "%a" lib/custom
-  expect_output "644"
+  expect_output $permission
   run_buildah rm $newcid
 }
 
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -1850,3 +1850,12 @@ load helpers
   target=alpine-image
   run_buildah 1 bud --authfile /tmp/nonexist --signature-policy ${TESTSDIR}/policy.json -t ${target} ${TESTSDIR}/bud/containerfile
 }
+
+@test "bud quiet" {
+  run_buildah bud --format docker -t quiet-test --signature-policy ${TESTSDIR}/policy.json -q ${TESTSDIR}/bud/shell
+  expect_line_count 1
+  expect_output --substring "^[0-9a-z]{64}$"
+
+  buildah rmi quiet-test
+}
+
--- a/tests/commit.bats
+++ b/tests/commit.bats
@@ -103,3 +103,20 @@ load helpers
   buildah rm $cid
   buildah rmi -a
 }
+
+@test "commit with name" {
+  buildah from --signature-policy ${TESTSDIR}/policy.json --name busyboxc busybox
+  run_buildah --log-level=error commit --signature-policy ${TESTSDIR}/policy.json busyboxc commitbyname/busyboxbyname
+  cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json localhost/commitbyname/busyboxbyname)
+  run_buildah inspect --format '{{.Container}}' $cid
+  expect_output $cid
+  buildah rm busyboxc $cid
+  buildah rmi commitbyname/busyboxbyname
+}
+
+@test "commit to docker-distribution" {
+  buildah from --signature-policy ${TESTSDIR}/policy.json --name busyboxc busybox
+  run_buildah commit --signature-policy ${TESTSDIR}/policy.json --tls-verify=false --creds testuser:testpassword busyboxc docker://localhost:5000/commit/busybox
+  run_buildah from --signature-policy ${TESTSDIR}/policy.json --name fromdocker --tls-verify=false --creds testuser:testpassword docker://localhost:5000/commit/busybox
+  buildah rm busyboxc fromdocker
+}
--- a/tests/containers.bats
+++ b/tests/containers.bats
@@ -27,6 +27,8 @@ load helpers
   cid2=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json busybox)
   run_buildah --log-level=error containers --format "{{.ContainerName}}"
   expect_line_count 2
+  expect_output --from="${lines[0]}" "alpine-working-container"
+  expect_output --from="${lines[1]}" "busybox-working-container"
 
   buildah rm -a
   buildah rmi -a -f
@@ -45,6 +47,7 @@ load helpers
   cid2=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json busybox)
   run_buildah --log-level=error containers --noheading
   expect_line_count 2
+  [[ "$output" != "NAME"* ]]
 
   buildah rm -a
   buildah rmi -a -f
@@ -55,6 +58,9 @@ load helpers
   cid2=$(buildah from --pull=false --signature-policy ${TESTSDIR}/policy.json busybox)
   run_buildah --log-level=error containers --quiet
   expect_line_count 2
+
+  [[ "${lines[0]}" != "$cid1"* ]]
+  [[ "${lines[1]}" != "$cid2"* ]]
 
   buildah rm -a
   buildah rmi -a -f
--- a/tests/from.bats
+++ b/tests/from.bats
@@ -327,6 +327,7 @@ load helpers
   container_name=mycontainer
   cid=$(buildah from --name=${container_name} --pull --signature-policy ${TESTSDIR}/policy.json alpine)
   buildah --log-level=error inspect --format '{{.Container}}' ${container_name}
+  buildah rm $cid
 }
 
 @test "from cidfile test" {
@@ -363,4 +364,20 @@ load helpers
 
 @test "from should fail with nonexist authfil" {
   run_buildah 1 from --authfile /tmp/nonexist --pull --signature-policy ${TESTSDIR}/policy.json alpine
+}
+
+@test "from pull always test" {
+  run buildah pull --signature-policy ${TESTSDIR}/policy.json docker.io/busybox
+  run_buildah from --signature-policy ${TESTSDIR}/policy.json --name busyboxc --pull-always docker.io/busybox
+  expect_output --substring "Getting"
+  buildah commit --signature-policy ${TESTSDIR}/policy.json busyboxc fakename-img
+  run_buildah 1 from --signature-policy ${TESTSDIR}/policy.json --pull-always fakename-img
+  buildah rm busyboxc
+  buildah rmi fakename-img
+}
+
+@test "from quiet test" {
+  run buildah rmi docker.io/busybox
+  run_buildah from --signature-policy ${TESTSDIR}/policy.json --quiet docker.io/busybox
+  [[ "$output" != "Getting"* ]]
 }
--- a/tests/images.bats
+++ b/tests/images.bats
@@ -157,3 +157,9 @@ load helpers
   buildah rm -a
   buildah rmi -a -f
 }
+
+@test "image digest test" {
+  buildah pull --signature-policy ${TESTSDIR}/policy.json busybox
+  run_buildah images --digests
+  expect_output --substring "sha256:"
+}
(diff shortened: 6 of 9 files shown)


tests/add.bats
```diff
@@ -108,35 +108,62 @@ load helpers
...
+
+@test "add with chown" {
+  createrandom ${TESTDIR}/randomfile
+  cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json busybox)
+  run_buildah add --chown bin:bin $cid ${TESTDIR}/randomfile /tmp/random
+  run_buildah run $cid ls -l /tmp/random
+
+  expect_output --substring bin.*bin
+  buildah ps
+  buildah images
```
TomSweeneyRedHat: just wondering why you're doing the ps and images commands without testing them?  I'd say drop them, but if you think they're important, then change `buildah` in each to `run_buildah`.

I know I'm guilty of this too, but are the rm and rmi call necessary with the current test harness @edsantiago

tests/from.bats
```diff
@@ -345,3 +346,24 @@ load helpers
   run_buildah --log-level=error containers -f id=${cid}
   buildah rm ${cid}
 }
+
+@test "from with/without pull test" {
+  run buildah from --pull=false docker.io/nginx
```
TomSweeneyRedHat: s/run_buildah/run buildah/  through out.

tests/push.bats
```diff
@@ -101,3 +101,46 @@ load helpers
...
+  expect_output --substring "newimage"
+  buildah rmi newimage busybox
+}
+
+@test "buildah push image to docker-archive and oci-archive" {
+  buildah pull --signature-policy ${TESTSDIR}/policy.json busybox
+  for dest in docker-archive oci-archive; do
+    mkdir ${TESTDIR}/tmp
+    run_buildah push --signature-policy ${TESTSDIR}/policy.json busybox $dest:${TESTDIR}/tmp/busybox.tar:latest
+    run ls ${TESTDIR}/tmp/busybox.tar
```
TomSweeneyRedHat: I don't think you need the 'run' here

tests/push.bats
```diff
@@ -101,3 +101,46 @@ load helpers
...
+
+@test "buildah push image to docker and docker registry" {
+  run which docker
+  if [[ $status -ne 0 ]]; then
+    skip "docker is not installed"
+  fi
+
+  buildah pull --signature-policy ${TESTSDIR}/policy.json busybox
+  run_buildah push --signature-policy ${TESTSDIR}/policy.json busybox docker-daemon:buildah/busybox:latest
+  run docker images
```
TomSweeneyRedHat: ditto run

TomSweeneyRedHat:
@edsantiago if you some time to eyeball this, I'd appreciate it.

tests/add.bats
```diff
@@ -108,35 +108,62 @@ load helpers
 @test "add single file creates absolute path with correct permissions" {
   imgName=ubuntu-image
   createrandom ${TESTDIR}/distutils.cfg
+  permission=$(stat -c "%a" ${TESTDIR}/distutils.cfg)
```
edsantiago: Nice work. Thanks for cleaning this up.

tests/add.bats
```diff
@@ -108,35 +108,62 @@ load helpers
...
   run_buildah rm $cid
 
   newcid=$(buildah from --signature-policy ${TESTSDIR}/policy.json ${imgName})
   run_buildah run $newcid stat -c "%a" lib/custom
-  expect_output "644"
+  expect_output $permission
   run_buildah rm $newcid
 }
+
+@test "add with chown" {
```
edsantiago: Is this really necessary as a separate test? Each separate test adds many seconds to the full run, and more if it has to pull an image. Why not just incorporate the `--chown` into the previous test? (line 131 above), and change the `stat` and `expect` in lines 132-134 to:
```bash
    run_buildah run $cid stat -c "%U:%G:%a" lib/custom
    expect_output "bin:bin:$permission" "ownership/mode of file copied into container"
```
(side note: the `buildah mount` in line 133 seems to be completely unnecessary. I think it's a copy-paste remnant. Feel free to remove it)

tests/add.bats
```diff
@@ -108,35 +108,62 @@ load helpers
...
+@test "add url" {
+  cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json busybox)
+  run_buildah add $cid https://github.com/containers/buildah/raw/master/README.md
+  run_buildah run $cid ls /README.md
+
+  run_buildah add $cid https://github.com/containers/buildah/raw/master/README.md /home
+  run_buildah run $cid ls /home/README.md
+
+  buildah rm $cid
+  buildah rmi busybox
```
edsantiago: Why the rmi? It seems wasteful to force other tests to re-pull the image. It's also a great way to increase test runtime and increase risk of flakes due to network failures. I don't think we want that.

tests/bud.bats
```diff
@@ -1821,3 +1821,11 @@ load helpers
 
   rm -f ${TESTSDIR}/bud/${target}/test*
 }
+
+@test "bud quite" {
```
edsantiago: Typo: I think you mean "quiet". Please fix in lines 1825-1829.

tests/bud.bats
```diff
@@ -1821,3 +1821,11 @@ load helpers
...
+  run_buildah bud --format docker -t quite-test --signature-policy ${TESTSDIR}/policy.json -q ${TESTSDIR}/bud/shell
+  expect_line_count 1
```
edsantiago: What if the line is a warning, or something unexpected? Please also include an `expect_output`.

tests/bud.bats
```diff
@@ -1821,3 +1821,11 @@ load helpers
...
+
+  buildah rmi quite-test
+  buildah rmi alpine
```
edsantiago: See previous comment about unnecessary `rmi`.

tests/add.bats
```diff
@@ -108,35 +108,62 @@ load helpers
...
+}
+
+@test "add url" {
+  cid=$(buildah from --signature-policy ${TESTSDIR}/policy.json busybox)
+  run_buildah add $cid https://github.com/containers/buildah/raw/master/README.md
+  run_buildah run $cid ls /README.md
+
+  run_buildah add $cid https://github.com/containers/buildah/raw/master/README.md /home
+  run_buildah run $cid ls /home/README.md
```
edsantiago: Please add `expect_output`. 

And, are both tests really necessary? If they are, e.g. because one specifies an explicit destination and the other assumes an implicit one, then please add a comment describing the reason.

tests/commit.bats
```diff
@@ -96,3 +96,26 @@ load helpers
   cid=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json alpine)
   run_buildah commit --signature-policy ${TESTSDIR}/policy.json $cid
 }
+
+@test "commit with name" {
+  buildah from --name nginxc nginx
```
edsantiago: Why so many different source images? Each time you specify `buildah from X/Y/Z` you need to pull X/Y/Z. Let's please be consistent and stick with common, small images.

tests/commit.bats
```diff
@@ -96,3 +96,26 @@ load helpers
...
+  run_buildah --log-level=error commit --signature-policy ${TESTSDIR}/policy.json nginxc commitbyname/nginxbyname
+  cid=$(buildah from localhost/commitbyname/nginxbyname)
+  run_buildah inspect $cid
```
edsantiago: Please add `expect_output`.

tests/commit.bats
```diff
@@ -96,3 +96,26 @@ load helpers
...
+  buildah rm nginxc $cid
+  buildah rmi commitbyname/nginxbyname nginx
```
edsantiago: See above comments about rmi. (In this case `rmi nginxbyname` is good but not `rmi your-base-image`).

tests/commit.bats
```diff
@@ -96,3 +96,26 @@ load helpers
...
+}
+
+@test "commit to docker-distribution" {
+  run which podman
+  if [ $status -ne 0 ]; then
+    skip "Can not find podman command skip the test"
+  fi
+  podman run -d -p 5001:5000 --name registry registry:2
```
edsantiago: WHOA! You cannot do this: there is already a registry running, and with this name too. It is an implied requirement for the buildah bats tests.

tests/commit.bats
```diff
@@ -96,3 +96,26 @@ load helpers
...
+  buildah from --name nginxc nginx
+  run_buildah commit --tls-verify=false nginxc docker://localhost:5001/commit/nginx
+  run_buildah from --tls-verify=false docker://localhost:5001/commit/nginx
+  podman rm -f registry
+  buildah rm -a
+  buildah rmi -fa
```
edsantiago: Please oh please don't do this

tests/containers.bats
```diff
@@ -27,6 +27,8 @@ load helpers
   cid2=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json busybox)
   run_buildah --log-level=error containers --format "{{.ContainerName}}"
   expect_line_count 2
+  expect_output --from="${lines[0]}" "alpine-working-container"
+  expect_output --from="${lines[1]}" "busybox-working-container"
```
edsantiago: Nice. Thanks.

tests/containers.bats
```diff
@@ -45,6 +47,7 @@ load helpers
   cid2=$(buildah from --pull --signature-policy ${TESTSDIR}/policy.json busybox)
   run_buildah --log-level=error containers --noheading
   expect_line_count 2
+  [[ "$output" != "NAME"* ]]
```
edsantiago: This does not do what you think it does.

It's clear that I need to add a 'not in output' test. Until I do so, please use:
```bash
    if [[ $output =~ NAME ]]; then
        expect_output "[no instance of 'NAME']" "buildah containers --noheading"
    fi
```
Likewise with lines 62-63 below.

tests/containers.bats
```diff
@@ -56,6 +59,9 @@ load helpers
   run_buildah --log-level=error containers --quiet
   expect_line_count 2
 
+  [[ "${lines[0]}" != "$cid1"* ]]
+  [[ "${lines[1]}" != "$cid2"* ]]
```
edsantiago: See line 50

tests/from.bats
edsantiago: Actually `s/run buildah/run_buildah/` but yes, what he said.

tests/from.bats
```diff
@@ -345,3 +346,24 @@ load helpers
...
+  [ $status -ne 0 ]
```
edsantiago: Remove. Instead, use the status-code form of `run_buildah`:
```bash
    run_buildah 1 from ....   (or whatever the expected exit status is, if not 1)
```

tests/from.bats
```diff
@@ -345,3 +346,24 @@ load helpers
...
+  run buildah from --name nginxc docker.io/nginx
+  expect_output --substring "Getting"
+  run buildah from --name nginxc-1 docker.io/nginx
+  [[ "$output" != "Getting"* ]]
```
edsantiago: See above comment re: this not doing what you think.

tests/from.bats
```diff
@@ -345,3 +346,24 @@ load helpers
...
+  run buildah from --name nginxc-2 --pull-always docker.io/nginx
+  expect_output --substring "Getting"
+  buildah commit nginxc fakename-img
```
edsantiago: run_buildah

tests/from.bats
```diff
@@ -345,3 +346,24 @@ load helpers
...
+  run buildah buildah from --pull-always fakename-img
+  [ $status -ne 0 ]
```
edsantiago: Remove, see above exit-status form of `run_buildah` comment.

tests/from.bats
```diff
@@ -345,3 +346,24 @@ load helpers
...
+  buildah rm nginxc nginxc-1 nginxc-2
+  buildah rmi fakename-img nginx
+}
+
+@test "from quiet test" {
+  run buildah from --quiet docker.io/busybox
+  [ "$output" != "Getting"* ]
```
edsantiago: This is not a valid test, because (1) this-does-not-do-what-you-think-it-does, and (2) you don't know if busybox is already cached locally, in which case there would be no `Getting` message. A proper test would ensure that the desired image is not already present. This might be an instance in which `rmi` is necessary, much as I dislike that.

tests/images.bats
```diff
@@ -157,3 +157,9 @@ load helpers
   buildah rm -a
   buildah rmi -a -f
 }
+
+@test "image digest test" {
+  buildah pull busybox
```
edsantiago: `run_buildah`

tests/add.bats
ypu: Sorry this is for some debug information during the tests. Will remove them.

tests/add.bats
ypu: Hi @edsantiago 
I checked that case but it is not a test for --chown flag but a basic tests for check buildah add file permissions. I'd like to keep per test cases to test a single function, so QE can find the problem quickly, and another reason I think we should keep a test case simple is the test case will be interrupted if one of the step failed and following steps will be cut off from the test results.

tests/add.bats
ypu: As run_buildah will raise en error if the file is not exist. So I thought the expect_output is not needed here. 
And for the test case itself, as long as it is supported, if we don't cover it there should be some code is not be tested. So I think it is worth for us to try to cover all flags and options in the test.

tests/add.bats
ypu: OK. Will remove it. Just want to keep the test env clean and not influence other test cases. But yes, this is not necessary for every test. Will check them and remove the ones not needed. Thanks a lot.

tests/bud.bats
ypu: Get it. Thanks a lot.

tests/from.bats
ypu: Will update it, thanks a lot for the tips.

tests/push.bats
ypu: Thanks for reminding. status check can be covered run the command directly. Will update it.

tests/push.bats
ypu: This run is used for update the $output context for the following expect_output check. So will keep it.

tests/images.bats
ypu: This is just a prepare step. If I understand the difference between run_buildah and buildah, we should use run_buildah in the test steps, but use buildah in the prepare steps.

tests/from.bats
ypu: I test it in my fedora31 and it works as expect.
Here is the code:
        run buildah pull busybox
        echo $output
        [[ "$output" != "Getting"* ]]
#   `[[ "$output" != "Getting"* ]]' failed with status 127
# 
# Getting image source signatures Copying blob sha256:0f8c40e1270f10d085dda8ce12b7c5b17cd808f055df5a7222f54837ca0feae0 Copying config sha256:020584afccce44678ec82676db80f68d50ea5c766b6e9d9601f7b5fc86dfb96d Writing manifest to image destination Storing signatures 020584afccce44678ec82676db80f68d50ea5c766b6e9d9601f7b5fc86dfb96d

tests/from.bats
ypu: Actually I should use [[ ]] here, and it should works.  Will update it. And yes, when I right a test case I default think the env is clean, so the image check is not request here. But as we want to keep the common use images during all tests, I will add the steps for exist images for it. Thanks for your comments.

TomSweeneyRedHat:
@ypu it looks like you've been doing a lot of great work addressing comments in this PR, TY!  Once you're complete and have pushed the last bits, can you please add a reply to this PR that it's ready for review again?  Thanks!

ypu:
Hi @TomSweeneyRedHat and @edsantiago Thanks a lot for your review and comments. Now I update the code to date to fit the travis-ci env. Can you help to review them again? Thanks again.

TomSweeneyRedHat:
@edsantiago still had a few outstanding questions that you answered, but he's not replied to.  Other than that, this LGTM.  FYI, @edsantiago will soon be traveling, so might not get to this for a bit.

tests/add.bats
edsantiago: I don't think I communicated my point well enough: I was trying to say that this test is almost identical to the previous one, and it would make sense to combine the two. Better readability and maintainability. But I won't block because of that.

tests/containers.bats
edsantiago: Please address this. As I wrote, this does not work: it does not do what you think it does. What it does is pass if `$output` does not begin with `NAME`. This will always pass, _always_, even if for some reason buildah emits a heading, because the heading does not begin with `NAME` - it begins with `CONTAINER ID`. What you **want** to do is make sure that `NAME` does not appear anywhere in the output. Again, please look into this more closely - this is a fundamental misunderstanding that is repeated in a few other places in your PR.

tests/images.bats
edsantiago: One much more important difference is that `run_buildah` will _display the command being run_, and its output, and will check its exit status. If `buildah pull` fails (probably because of a network flake), BATS will exit with a hard-to-debug error. `run_buildah` will give much better information.

tests/from.bats
edsantiago: Yes, it works, but it is not a good test: it does not check what you think it checks. You want to fail if "Getting" appears anywhere in the output; this test only fails if the output begins with "Getting". Which in most cases it will, but not if there are warning messages or any other output. Please see my comments a few minutes ago about testing string output.

rhatdan:
@ypu Please rebase and answer @edsantiago questions.  Would love to get this in.

edsantiago:
Ping, this is good work, it only needs some fixes and a rebase

rhatdan:
@edsantiago could you take this over and push it over the finish line?

edsantiago:
@rhatdan let's give @ypu an overnight (for us) chance to respond; I'll get on this tomorrow if nak.

rhatdan:
Since @edsantiago finished this PR CLosing.

PR #1971, opened 2019-11-08
Cirrus: Update to F32 + Use libpod cache images

cevich:
***Depends on:*** https://github.com/containers/libpod/pull/5368


rhatdan:
bot, add author to whitelist

cevich:
***NOTE TO ME:*** Put back the buildah-specific package install bit, some parts of the matrix depend on it.

cevich:
Waiting on https://github.com/checkpoint-restore/criu/issues/973

cevich:
Status:  This PR is blocking on https://github.com/containers/libpod/pull/5853 because we need a new/better/unified in-podman container image that includes all the necessary tooling for buildah also.

Meanwhile, I'm still fighting a random-hang w/o any useful error issue in golangci-lint in this PR.  I've got it narrowed down to a few lines in `.golangci.yml` but am not 100% yet due to the random nature of the hangs.

Lastly, @edsantiago would you mind taking a 3-minute peek at the failures in `cirrus-ci/required/testing image:fedora-32-libpod-6301182083727360`?  I resolved the previous issue of having both runc and crun installed.  Something just looks really really wrong with so many tests having an empty `actual:` output string.  This same VM image (but with packaged buildah and source podman) ***use to*** pass the (pre-v2) libpod integration tests.  Any guesses?

Note: I'm planning to bring in a new VM image shortly, so don't spend too much time.

edsantiago:
Looks like `buildah run` again. The best way to debug would be to ssh into a VM and try `buildah run`, and figure out how to tweak containers.conf or whatever until it works.

We need to geto logformatter working on this tree!

cevich:
Okay thanks Ed.

> We need to geto logformatter working on this tree!

Yeah, and your task-name beautification also.  No worries, ambitions always outstrip available time, and at this point I'm just happy to see small incremental improvements in "it actually runs" :grinning:

edsantiago:
Does anyone know if buildah tests have *ever* run on f32? @nalind? @TomSweeneyRedHat? Bodhi test results have aged out. I'm starting a VM now to test. Reason I ask is, I see hardcoded 'runc' in tests but no 'crun'.

edsantiago:
> I resolved the previous issue of having both runc and crun installed. 

This is the problem: buildah needs `runc`, even on cgroupsv2 systems. But buildah tests require podman, which needs crun, so both need to be installed.

cevich:
> Does anyone know if buildah tests have ever run on f32?

I don't think so, I believe this is a brand-new thing.

> This is the problem: buildah needs runc, even on cgroupsv2 systems. But buildah tests require podman, which needs crun, so both need to be installed.

:hankey: 

So...in libpod we added this: https://github.com/containers/libpod/blob/master/contrib/cirrus/setup_environment.sh#L72-L104

Thankfully it happens at testing runtime, so it's not cooked into the VM image.  We're **not** doing that in this repo, but maybe we need to?  How does buildah know to use runc vs crun in the tests?

cevich:
(and it sounds like I need to add a step to install runc into `contrib/cirrus/setup.sh`)

edsantiago:
>I don't think so, I believe this is a brand-new thing.

I meant ever. And they do, in Bodhi gating tests. I just couldn't find logs.

I think the first step right now is to make sure that `runc`, `crun`, and `podman` all exist on whatever system buildah tests are running on. The next step will be to make sure that `podman` works properly in that setup. It might work fine, I don't know, it's just something to watch for.

edsantiago:
> How does buildah know to use runc vs crun in the tests?

Buildah does nothing whatsoever with crun. It doesn't work with crun, period. Only runc. There are a number of buildah tests that check cgroups: those are simply disabled on a cgroupsv2 system.

cevich:
Thanks for the analysis @edsantiago I will teach the `setup.sh` script to install runc and explicitly configure podman to use crun.  Then let's see where that gets us.

cevich:
@giuseppe @rhatdan @edsantiago I think I understand what's happening here, since Giuseppe discovered SELinux and a dontaudit rule is to blame.  It also explains why the problem is not reproducible when running as root through ssh.

The way Cirrus-CI executes scripts on VMs, is by downloading and executing an "agent" process on boot via a "startup script" hook provided by google.  For example, here's [the tree and SELinux contexts:](https://cirrus-ci.com/task/4674801992204288?command=integration_test#L64-L49):

```
[+0053s]  |-google_metadata(`system_u:system_r:unconfined_service_t:s0')
[+0053s]  |  `-tmp34fqxto3(`system_u:system_r:unconfined_service_t:s0')
[+0053s]  |     `-cirrus-ci-agent(`system_u:system_r:unconfined_service_t:s0')
[+0053s]  |        |-bash(`system_u:system_r:unconfined_service_t:s0')
[+0053s]  |        |  |-awk(`system_u:system_r:unconfined_service_t:s0')
[+0053s]  |        |  `-bash(`system_u:system_r:unconfined_service_t:s0')
[+0053s]  |        |     `-pstree(`system_u:system_r:unconfined_service_t:s0')
[+0053s]  |        |-{cirrus-ci-agent}(`system_u:system_r:unconfined_service_t:s0')
[+0053s]  |        |-{cirrus-ci-agent}(`system_u:system_r:unconfined_service_t:s0')
[+0053s]  |        |-{cirrus-ci-agent}(`system_u:system_r:unconfined_service_t:s0')
[+0053s]  |        |-{cirrus-ci-agent}(`system_u:system_r:unconfined_service_t:s0')
[+0053s]  |        |-{cirrus-ci-agent}(`system_u:system_r:unconfined_service_t:s0')
[+0053s]  |        |-{cirrus-ci-agent}(`system_u:system_r:unconfined_service_t:s0')
[+0053s]  |        |-{cirrus-ci-agent}(`system_u:system_r:unconfined_service_t:s0')
[+0053s]  |        `-{cirrus-ci-agent}(`system_u:system_r:unconfined_service_t:s0')
```

I think SELinux may actually be doing the correct thing, blocking our scripts running as `unconfined_service_t`.  Fortunately, this service is started by systemd, in `/lib/systemd/system/google-startup-scripts.service`:

```ini
[Unit]
Description=Google Compute Engine Startup Scripts
After=local-fs.target network-online.target network.target rsyslog.service
After=google-instance-setup.service google-network-daemon.service
Wants=local-fs.target network-online.target network.target

[Service]
ExecStart=/usr/bin/google_metadata_script_runner --script-type startup
KillMode=process
Type=oneshot

[Install]
WantedBy=multi-user.target
```

Correct me if I'm wrong, according to `man systemd.exec` I can add to the `[service]` section something like:

```
SELinuxContext=system_u:system_r:container_init_t
```

Note: I'm assuming that type will automatically pass down to all child process (shell scripts) spawned by the Cirrus-CI agent process.

In all cases, I'll need to make this unit-file change in the VM image, since by the time the tests are running, it's too late (the agent has already started).  So we need to pick a context/type that will also be kosher for running the libpod testing scripts.  If this is no good, and the type needs to vary across containers projects, it can be done but is a super big PITA.

cevich:
> blocking our scripts running as ...

Clarification: Meaning, buildah executes crun, but we don't see any stdio from inside the container.  This is the problem we're trying to solve for buildah tests running on F31 and 32.

rhatdan:
container_runtime_t is the corret label, simpler would be to label your script container_runtime_exec_t, which systemd will automatically transition to container_runtime_t.

giuseppe:
> container_runtime_t is the corret label, simpler would be to label your script container_runtime_exec_t, which systemd will automatically transition to container_runtime_t.

shouldn't we try to run it from a `unconfined_u:unconfined_r:unconfined_t` label so to emulate an interactive shell?

cevich:
@giuseppe I wrote a big long reply reasoning in favor of @rhatdan 's `container_runtime_t` then erased it all. I think you're idea of a more general label is correct. This is a slightly contrived/unrealistic environment out of necessity for testing the system itself.  In other words, we cannot predict that for now and in the future, our testing-scripts will _only_ behave as permitted by `container_runtime_t`.  Does that make sense @rhatdan or should we go with a more specific service context in order to better exercise the policy itself?

rhatdan:
If your scripts are attempting to emulate a normal user execuing it, then unconfined_t, makes the most sense, as long as the buildah/podman executables are properly labeled container_runtime_exec_t.  Although if they are not, the tests might still pass.

cevich:
@rhatdan in all cases the executables should be labeled correctly when `make install` runs.  Though I just tried `unconfined_t` but the service fails to start, do I need to label the executable also?

```
● google-startup-scripts.service - Google Compute Engine Startup Scripts
   Loaded: loaded (/etc/systemd/system/google-startup-scripts.service; enabled; vendor preset: enabled)
   Active: failed (Result: exit-code) since Wed 2020-04-29 10:14:14 EDT; 10min ago
  Process: 736 ExecStart=/usr/bin/google_metadata_script_runner --script-type startup (code=exited, status=229/SELINUX_CONTEXT)
 Main PID: 736 (code=exited, status=229/SELINUX_CONTEXT)
      CPU: 1ms

Apr 29 10:14:14 cevich-fedora-31-libpod-6220812239765504 systemd[1]: Starting Google Compute Engine Startup Scripts...
Apr 29 10:14:14 cevich-fedora-31-libpod-6220812239765504 systemd[736]: google-startup-scripts.service: Failed to change SELinux context to unconfined_u:unconfined_r:unconfined_t: Operation not permitted
Apr 29 10:14:14 cevich-fedora-31-libpod-6220812239765504 systemd[736]: google-startup-scripts.service: Failed at step SELINUX_CONTEXT spawning /usr/bin/google_metadata_script_runner: Operation not permitted
Apr 29 10:14:14 cevich-fedora-31-libpod-6220812239765504 systemd[1]: google-startup-scripts.service: Main process exited, code=exited, status=229/SELINUX_CONTEXT
Apr 29 10:14:14 cevich-fedora-31-libpod-6220812239765504 systemd[1]: google-startup-scripts.service: Failed with result 'exit-code'.
Apr 29 10:14:14 cevich-fedora-31-libpod-6220812239765504 systemd[1]: Failed to start Google Compute Engine Startup Scripts.
```

cevich:
```
ls -laZ /usr/bin/google_metadata_script_runner
-rwxr-xr-x. 1 root root system_u:object_r:bin_t:s0 471 Jan 31  2019 /usr/bin/google_metadata_script_runner
```

rhatdan:
What was the failure message? Any AVCs?

cevich:
Tons and tons of stuff in audit :(

```
type=SERVICE_START msg=audit(1588169654.143:69): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=google-startup-scripts comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=failed'^]UID="root" AUID="unset"
```

I'm thinking of labeling it `system_u:object_r:cloud_init_exec_t:s0` same as cloud-init which does a substantially similar job.  Also, I need to raise this issue to the upstream packager, since really this stuff should happen in the RPM (as it does for cloud-init).

cevich:
**sigh** `cloud_init_exec_t` doesn't work (same error).

```
Apr 29 10:53:26 cevich-fedora-31-libpod-6220812239765504 systemd[736]: google-startup-scripts.service: Failed to change SELinux context to unconfined_u:unconfined_r:unconfined_t: Operation not permitted
Apr 29 10:53:26 cevich-fedora-31-libpod-6220812239765504 systemd[736]: google-startup-scripts.service: Failed at step SELINUX_CONTEXT spawning /usr/bin/google_metadata_script_runner: Operation not permitted
```

cevich:
@rhatdan this service comes from the `google-compute-engine-tools` RPM in Fedora.  Only changes I made were to add `SELinuxContext=unconfined_u:unconfined_r:unconfined_t` under `[service]`.  I believe the default label for the binary is `system_u:object_r:bin_t:s0`

cevich:
@rhatdan I tried running `semodule -DB` and `setenforce permissive`, then `systemctl restart google-startup-scripts` but I get the same error, and same lack of any detail in `audit.log` :disappointed:

cevich:
Update: This did the trick: `SELinuxContext=system_u:system_r:unconfined_t:s0` (I was missing the `s0`).  I confirmed this by adding a startup script to the metadata for the VM which captured the output of `pstree -Z`.  After rebooting and examining the output, I can confirm that the running bash script had the `unconfined_t` label on it.

Since I need to rebuild the base-images to grab the new F32 release anyway, I'm including this fix along for the ride: https://github.com/containers/libpod/pull/6044

cevich:
(Quick/small update of the `in_podman` container image to use the `master` tag)

.golangci.yml
```diff
@@ -4,8 +4,8 @@ run:
     - apparmor
     - seccomp
     - selinux
-  concurrency: 6
-  deadline: 5m
+  # Don't exceede number of threads available when running under CI
```
TomSweeneyRedHat, line 7: nit of a nit, only worry about if other stuff is found.
exceede -> exceed

.golangci.yml
```diff
@@ -17,7 +17,8 @@ linters:
     - gofmt
     - goimports
     - golint
-    - gosimple
+    # Fixme: Broken? dies w/o error randomly, between ~100-200 seconds
+    # - gosimple
```
TomSweeneyRedHat, line 21: beauty, love random fails.

TomSweeneyRedHat:
LGTM, but definitely would like a head nod from @edsantiago

.golangci.yml
cevich: :cry: this is not perfect either, it still randomly takes like 10-minutes to run all these checks.  But after all my iterations here since commenting out `gosimple`, there have been no more random no-error failures.

.golangci.yml
cevich: Woops, okay.  Will fix this along with anything @edsantiago comes up with (he always finds something) :+1:

contrib/cirrus/add_second_partition.sh
```diff
@@ -0,0 +1,63 @@
...
+source $(dirname $0)/lib.sh
+
+# Protection against double-execution
+if [[ ! -r "/root" ]] || [[ -r "/root/second_partition_ready" ]]
+then
+    echo "Warning: Ignoring attempted execution of $(basename $0)"
+    exit 0
+fi
+
+[[ -n "type -P parted" ]] || \
```
edsantiago, line 26: This is not what you meant to do.

contrib/cirrus/add_second_partition.sh
```diff
@@ -0,0 +1,63 @@
...
+if $FINDMNTCMD | tee $TMPF | egrep -q "^/\s+${SLASH_DEVICE}1"
+then
+    echo "Repartitioning original partition table:"
+    $PPRINTCMD
+else
+    die 6 "Unexpected output from '$FINDMNTCMD': $(<$TMPF)"
+fi
+
+echo "Adding partition offset within unpartitioned space."
+parted --script --align optimal /dev/sda unit % mkpart primary "" "" "$NEW_PART_START" "$NEW_PART_END"
```
edsantiago, line 46: Should this be $SLASH_DEVICE?

contrib/cirrus/add_second_partition.sh
cevich: whoa! good catch!

contrib/cirrus/lib.sh
```diff
@@ -192,8 +200,38 @@ EOF
     cat $FILEPATH
```
edsantiago, line 200: [not shown in github diff]. Please remove `install_ooe()` above.

contrib/cirrus/test.sh
```diff
@@ -34,7 +34,7 @@ then
             export GITVALIDATE_TIP="$CIRRUS_CHANGE_IN_REPO"
```
edsantiago, line 34: [lines not shown in github diff] Please remove `ooe.sh` from lines above

contrib/cirrus/setup.sh
```diff
@@ -2,118 +2,47 @@
...
 
-source $(dirname $0)/lib.sh
+        # Not executing IN_PODMAN container
+        if [[ -z "$CONTAINER" ]]; then
+            warn "Adding secondary testing partition & growing root filesystem"
+            bash $SCRIPT_BASE/add_second_partition.sh
+        fi
 
-install_ooe
+        warn "Buildah testing requires runc"
```
edsantiago, line 32: After Dan's #2334, I'm not sure this is true any more. I'm going to run some tests to find out, it may be an hour. I'll check back in when I know.

contrib/cirrus/lib.sh
```diff
@@ -127,6 +133,10 @@ die() {
     exit ${1:-1}
 }
 
+warn() {
+    echo ">>>>> WARNING: ${1:-WARNING (but no message given!) in ${FUNCNAME[1]}()}" > /dev/stderr
```
edsantiago, line 137: I see `warn` used a lot for merely informational messages. The big `WARNING` here might mislead people into thinking something is wrong. One option might be to remove the 'WARNING' from here; another (which I lean slightly more towards) would be to limit the use of `warn` to actual warnings, and use plain 'echo` for informationals.

hack/get_ci_vm.sh
```diff
@@ -69,9 +69,10 @@ delvm() {
 }
 
 image_hints() {
+    _BIS=$(egrep -m 1 '_BUILT_IMAGE_SUFFIX:[[:space:]+"[[:print:]]+"' "$BUILDAHROOT/.cirrus.yml" | cut -d: -f 2 | tr -d '"[:blank:]')
```
edsantiago, line 72: I think I'd feel happier with a `test -n "${_BIS}" || die`

edsantiago:
Sorry for the late reply; I had to restart tests several times.

Buildah tests pass on Rawhide with crun installed, no runc. I believe it is safe to remove the above-noted `dnf install runc`.

.golangci.yml
cevich: okay fixed.

.golangci.yml
cevich: I improved this comment.

contrib/cirrus/add_second_partition.sh
cevich: okay, fixed (and did an executable-test instead of string length)

contrib/cirrus/add_second_partition.sh
cevich: yep.  got it.

contrib/cirrus/lib.sh
cevich: got it, good catch, thanks.

contrib/cirrus/test.sh
cevich: `ooe.sh` is now part of the VM image, put there at build time from containers/automation repo.

contrib/cirrus/setup.sh
cevich: oh right, leftover from before debugging this issue.  Thx I'll remove it.

contrib/cirrus/lib.sh
cevich: Yes this was true, though I just grepped through 'contrib/cirrus/' and don't see any that shouldn't be warnings.

hack/get_ci_vm.sh
cevich: k, got it.

cevich:
bors retry

PR #1972, opened 2019-11-08
images: adjust output of tags and digests

nalind:
Output (possibly multiple) tags and digests in the JSON images list, and provide all of an image's digests when listing then in non-JSON format.   Don't mistake image names which end in "@sha256:hex..." for images whose tagged names include a repository name that ends with "@sha256".

### DIFF
--- a/cmd/buildah/images.go
+++ b/cmd/buildah/images.go
@@ -23,8 +23,9 @@ import (
 
 type jsonImage struct {
 	ID           string    `json:"id"`
-	Names        []string  `json:"names"`
-	Digest       string    `json:"digest"`
+	Names        []string  `json:"names,omitempty"`
+	Digest       string    `json:"digest,omitempty"`
+	Digests      []string  `json:"digests,omitempty"`
 	CreatedAt    string    `json:"createdat"`
 	Size         string    `json:"size"`
 	CreatedAtRaw time.Time `json:"createdatraw"`
@@ -32,10 +33,11 @@ type jsonImage struct {
 }
 
 type imageOutputParams struct {
-	Tag          string
 	ID           string
 	Name         string
+	Tag          string
 	Digest       string
+	Digests      []string
 	CreatedAt    string
 	Size         string
 	CreatedAtRaw time.Time
@@ -247,9 +249,9 @@ func outputHeader(opts imageOptions) string {
 	if opts.quiet {
 		return formats.IDString
 	}
-	format := "table {{.Name}}\t{{.Tag}}\t"
-	if opts.noHeading {
-		format = "{{.Name}}\t{{.Tag}}\t"
+	format := "{{if .Name}}{{.Name}}{{else}}<none>{{end}}\t{{if .Tag}}{{.Tag}}{{else}}<none>{{end}}\t"
+	if !opts.noHeading {
+		format = "table " + format
 	}
 
 	if opts.digests {
@@ -268,12 +270,13 @@ func outputImages(ctx context.Context, systemContext *types.SystemContext, store
 	found := false
 	var imagesParams imagesSorted
 	jsonImages := []jsonImage{}
+
 	for _, image := range images {
 		if image.ReadOnly {
 			opts.readOnly = true
 		}
 		createdTime := image.Created
-		inspectedTime, digest, size, _ := getDateAndDigestAndSize(ctx, systemContext, store, image)
+		inspectedTime, size, _ := getDateAndSize(ctx, systemContext, store, image)
 		if !inspectedTime.IsZero() {
 			if createdTime != inspectedTime {
 				logrus.Debugf("image record and configuration disagree on the image's creation time for %q, using the configuration creation time: %s", image.ID, inspectedTime)
@@ -301,45 +304,94 @@ func outputImages(ctx context.Context, systemContext *types.SystemContext, store
 			imageID = shortID(image.ID)
 		}
 
-	outer:
-		for name, tags := range imageReposToMap(image.Names) {
-			for _, tag := range tags {
-				if !matchesReference(name+":"+tag, argName) {
-					continue
-				}
-				found = true
+		filterMatched := false
 
-				if !matchesFilter(ctx, store, image, name+":"+tag, filters) {
-					continue
+		var imageReposAndTags [][2]string
+		var imageDigests []string
+		for _, imageDigest := range image.Digests {
+			imageDigests = append(imageDigests, imageDigest.String())
+		}
+
+		for _, name := range image.Names {
+			if name == "" {
+				logrus.Warnf("Found image with empty name")
+				continue
+			}
+			named, err := reference.ParseNormalizedNamed(name)
+			if err != nil {
+				logrus.Warnf("Error parsing name %q: %v", name, err)
+				continue
+			}
+			if name != named.String() {
+				logrus.Debugf("Image name %q wasn't already in its normalized form (%q).", name, named.String())
+			}
+
+			if !matchesReference(name, argName) {
+				continue
+			}
+			found = true
+
+			if digested, ok := named.(reference.Digested); ok {
+				digest := digested.Digest()
+				digestPresent := false
+				for _, imageDigest := range imageDigests {
+					if imageDigest == digest.String() {
+						digestPresent = true
+					}
 				}
-				if opts.json {
-					jsonImages = append(jsonImages,
-						jsonImage{ID: image.ID,
-							Names:        image.Names,
-							Digest:       digest,
-							CreatedAtRaw: createdTime,
-							CreatedAt:    units.HumanDuration(time.Since((createdTime))) + " ago",
-							Size:         formattedSize(size),
-							ReadOnly:     image.ReadOnly,
-						})
-					// We only want to print each id once
-					break outer
+				if !digestPresent {
+					imageDigests = append(imageDigests)
 				}
-				params := imageOutputParams{
-					Tag:          tag,
-					ID:           imageID,
-					Name:         name,
-					Digest:       digest,
-					CreatedAtRaw: createdTime,
-					CreatedAt:    units.HumanDuration(time.Since((createdTime))) + " ago",
-					Size:         formattedSize(size),
-					ReadOnly:     image.ReadOnly,
-				}
-				imagesParams = append(imagesParams, params)
-				if opts.quiet {
-					// We only want to print each id once
-					break outer
-				}
+			}
+
+			if !matchesFilter(ctx, store, image, name, filters) {
+				continue
+			}
+			filterMatched = true
+
+			if tagged, ok := named.(reference.Tagged); ok {
+				imageReposAndTags = append(imageReposAndTags, [2]string{named.Name(), tagged.Tag()})
+			} else {
+				imageReposAndTags = append(imageReposAndTags, [2]string{named.Name(), ""})
+			}
+		}
+		if len(image.Names) == 0 && matchesFilter(ctx, store, image, "", filters) {
+			filterMatched = true
+		}
+		if !filterMatched {
+			continue
+		}
+
+		if opts.json {
+			jsonImages = append(jsonImages, jsonImage{
+				ID:           image.ID,
+				Names:        image.Names,
+				Digest:       string(image.Digest),
+				Digests:      imageDigests,
+				CreatedAtRaw: createdTime,
+				CreatedAt:    units.HumanDuration(time.Since((createdTime))) + " ago",
+				Size:         formattedSize(size),
+				ReadOnly:     image.ReadOnly,
+			})
+			continue
+		}
+		if len(imageReposAndTags) == 0 {
+			imageReposAndTags = [][2]string{{"", ""}}
+		}
+		for _, imageRepoAndTag := range imageReposAndTags {
+			imagesParams = append(imagesParams, imageOutputParams{
+				ID:           imageID,
+				Name:         imageRepoAndTag[0],
+				Tag:          imageRepoAndTag[1],
+				Digest:       string(image.Digest),
+				Digests:      imageDigests,
+				CreatedAtRaw: createdTime,
+				CreatedAt:    units.HumanDuration(time.Since((createdTime))) + " ago",
+				Size:         formattedSize(size),
+				ReadOnly:     image.ReadOnly,
+			})
+			if opts.quiet {
+				break
 			}
 		}
 	}
@@ -415,10 +467,10 @@ func matchesFilter(ctx context.Context, store storage.Store, image storage.Image
 }
 
 func matchesDangling(name string, dangling string) bool {
-	if dangling == "false" && !strings.Contains(name, "<none>") {
+	if dangling == "false" && name != "" {
 		return true
 	}
-	if dangling == "true" && strings.Contains(name, "<none>") {
+	if dangling == "true" && name == "" {
 		return true
 	}
 	return false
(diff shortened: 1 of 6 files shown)


rhatdan:
LGTM

rhatdan:
@mtrmac PTAL

cmd/buildah/images.go
```diff
@@ -506,27 +539,37 @@ func formattedSize(size int64) string {
...
+		}
+		if digested, ok := named.(reference.Digested); ok {
+			digest := digested.Digest()
+			tagOrDigest = digest.String()
 		}
-		repos[repository] = append(repos[repository], tag)
+		repos[repository] = append(repos[repository], tagOrDigest)
 	}
 	if len(repos) == 0 {
-		repos["<none>"] = []string{"<none>"}
```
TomSweeneyRedHat: just curious why we're getting rid of the <none> for the dangling images?  I thought we'd some docs telling folks to look for that tag to determine what was dangling or not?

nalind: This moves us from inserting `<none>` as a recognized-as-fake value here to having the tabular output logic decide whether or not to output `<none>` based on whether or not there's a value to output, and to having the JSON output format be `,omitempty` for tag and digest fields.  If the docs are about non-JSON `buildah images` output, the end-result should look the same.

TomSweeneyRedHat: Thanks @nalind, the `buildah images` is what I was most concerned about.

cmd/buildah/images.go
```diff
@@ -245,9 +251,9 @@ func outputHeader(opts imageOptions) string {
 	if opts.quiet {
 		return formats.IDString
 	}
-	format := "table {{.Name}}\t{{.Tag}}\t"
+	format := "table {{if .Name}}{{.Name}}{{else}}<none>{{end}}\t{{if .Tag}}{{.Tag}}{{else}}<none>{{end}}\t"
 	if opts.noHeading {
-		format = "{{.Name}}\t{{.Tag}}\t"
+		format = "{{if .Name}}{{.Name}}{{else}}<none>{{end}}\t{{if .Tag}}{{.Tag}}{{else}}<none>{{end}}\t"
```
mtrmac: (Absolutely non-blocking nit: Is the difference the initial `"table "`? Maybe do a
```go
format := ""
if !opts.noHeading {
    format += "table "
}
format += {{…}}`
```
to be clear about the difference, and to structurally ensure that the two column definitions are the same.)

cmd/buildah/images.go
```diff
@@ -506,27 +539,37 @@ func formattedSize(size int64) string {
...
+		if name == "" {
 			logrus.Warnf("Found image with empty name")
+			continue
+		}
+		named, err := reference.ParseNormalizedNamed(name)
+		if err != nil {
+			logrus.Warnf("Error parsing name %q: %v", name, err)
+			continue
+		}
+		if named.Name() != name {
```
mtrmac: This must be `.String()`.  `.Name()` only returns the repo.

cmd/buildah/images.go
```diff
@@ -506,27 +539,37 @@ func formattedSize(size int64) string {
...
+			logrus.Debugf("Name %q wasn't already in its normalized form (%q).", name, named.Name())
+		}
+		var repository, tagOrDigest string
+		repository = reference.TrimNamed(named).Name()
```
mtrmac: `reference.Name()` does the same thing.

cmd/buildah/images.go
```diff
@@ -506,27 +539,37 @@ func formattedSize(size int64) string {
...
+		if digested, ok := named.(reference.Digested); ok {
+			digest := digested.Digest()
+			tagOrDigest = digest.String()
 		}
-		repos[repository] = append(repos[repository], tag)
+		repos[repository] = append(repos[repository], tagOrDigest)
 	}
 	if len(repos) == 0 {
-		repos["<none>"] = []string{"<none>"}
+		repos[""] = []string{""}
```
mtrmac: (Non-blocking: This does not _quite_ fit into what this function otherwise does, and notably it muddles the semantic type of the map keys/values: without this the map key is conceptually a `reference.Named`, now it is a “`reference.Named` or empty string”.

This special case exists so that the table output has at least one `<none>:<none>` column, fine, but that is a concern of the caller building the table, and would be a better fit to handle in the caller.)

cmd/buildah/images.go
```diff
@@ -300,34 +306,63 @@ func outputImages(ctx context.Context, systemContext *types.SystemContext, store
 		}
 
 	outer:
-		for name, tags := range imageReposToMap(image.Names) {
-			for _, tag := range tags {
-				if !matchesReference(name+":"+tag, argName) {
+		for repo, tagsAndDigests := range imageReposToMap(image.Names) {
+			for _, tagOrDigest := range tagsAndDigests {
```
mtrmac: AFAICS _nothing_ in here needs the data to actually be clustered by `repo`; it’s just two nested loops that process _all_ of the values of `image.Names` in sequence.

Is the goal only to sort the output by repo?

Don’t you really want a `[]reference.Named`, the full objects,, built by parsing `image.Names`, here? Then sort it using whatever key you need. And then process it in one non-nested loop.

cmd/buildah/images.go
```diff
@@ -300,34 +306,63 @@ func outputImages(ctx context.Context, systemContext *types.SystemContext, store
...
+						imageDigest = tagOrDigest
+						imageName = repo + "@" + imageDigest
+					} else {
+						imageTag = tagOrDigest
+						if len(image.Digest) > 0 {
+							imageDigest = image.Digest.String()
+						}
+						imageName = repo + ":" + imageTag
+					}
+				}
```
mtrmac: All of the `imageName`/`imageDigest`/`imageTag`, and the full string format, was already present in the `reference.Named` object. The code has created that object, type-erased the `tagsAndDigest` field, and now laboriously reconstructs the values and the string format.

E.g. `imageName := namedRef.String()` is all it should take to create that value; the “has digest” check should be an interface conformance check, and so on.

cmd/buildah/images.go
```diff
@@ -300,34 +306,63 @@ func outputImages(ctx context.Context, systemContext *types.SystemContext, store
...
-						jsonImage{ID: image.ID,
-							Names:        image.Names,
-							Digest:       digest,
-							CreatedAtRaw: createdTime,
-							CreatedAt:    units.HumanDuration(time.Since((createdTime))) + " ago",
-							Size:         formattedSize(size),
-							ReadOnly:     image.ReadOnly,
-						})
+					var imageTags []string
+					for _, tagOrDigest := range tagsAndDigests {
```
mtrmac: Should this really only list tags _from this single `repo` value_? The choice which particular repo will be used in the output is random, depending on how Go decides to order the map keys this time.

cmd/buildah/images.go
```diff
@@ -300,34 +306,63 @@ func outputImages(ctx context.Context, systemContext *types.SystemContext, store
...
+				if !matchesReference(imageName, argName) {
 					continue
 				}
 				found = true
 
-				if !matchesFilter(ctx, store, image, name+":"+tag, filters) {
+				if !matchesFilter(ctx, store, image, imageName, filters) {
 					continue
 				}
+				imageDigests := make([]string, 0, len(image.Digests))
```
mtrmac: (Absolutely non-blocking: `imageDigests` is invariant to `repo`/`tagsAndDigests`, and can be computed outside of the loop. OTOH that would naively cause the value to be computed for all non-matching images as well, so, maybe compute it here, but only if not computed already. Yet again, this is it’s cheap enough.)

cmd/buildah/images.go
nalind: Ah, thanks!  Fixing.

cmd/buildah/images.go
nalind: Ah, fixing.

cmd/buildah/images.go
mtrmac: … sorry, the `imageTags` value in JSON output does only use the current-`repo` values. But should it do so?

cmd/buildah/images.go
nalind: Yup.  I'm trying to avoid reimplementing the caller and the filtering/matching logic, though, because that feels like a much bigger patch.

cmd/buildah/images.go
nalind: Yes, it is.  Changing it.

cmd/buildah/images.go
nalind: That's a good reason to not add the `tags` field in this PR.  Dropping it.

cmd/buildah/images.go
mtrmac: As far as this particular review comment goes, I’d be happy with just moving the `if len(repos) == 0 { repos[""] = []string{""}}` condition to the caller, without changing the logic otherwise.

Anyway, this _is_ non-blocking, and more importantly maybe the map should not exist at all.

cmd/buildah/images.go
nalind: I didn't have a strong feeling about it either way.  Changing it.

cmd/buildah/images.go
nalind: Easy enough to move that, since this is the only caller.

cmd/buildah/images.go
nalind: That's what I took your meaning to be in https://github.com/containers/buildah/pull/1972#discussion_r344352644.  Dropping the field, which would have been added in this PR.

cmd/buildah/images.go
```diff
@@ -299,35 +304,64 @@ func outputImages(ctx context.Context, systemContext *types.SystemContext, store
...
+					} else {
+						imageTag = tagOrDigest
+						if len(image.Digest) > 0 {
+							imageDigest = image.Digest.String()
+						}
+						imageName = repo + ":" + imageTag
+					}
+				}
+
+				if !matchesReference(imageName, argName) {
```
mtrmac: BTW does `matchesReference` now need updating? If I’m reading it right, `imageName: "foo@sha256:…"` doesn’t match `argName: "foo"`.

cmd/buildah/images.go
mtrmac: Once that field is dropped, I can’t see any more reason to have the `imageReposToMap` map and nesting two loops.

cmd/buildah/images.go
```diff
@@ -299,35 +304,64 @@ func outputImages(ctx context.Context, systemContext *types.SystemContext, store
...
+		repos := imageReposToMap(image.Names)
+		if len(repos) == 0 {
+			repos = map[string][]string{"": []string{""}}
+		}
 	outer:
-		for name, tags := range imageReposToMap(image.Names) {
-			for _, tag := range tags {
-				if !matchesReference(name+":"+tag, argName) {
+		for repo, tagsAndDigests := range repos {
+			var imageDigests []string
```
mtrmac: This could be moved out of this loop as well.

nalind: Ah, right.  Moving it.

cmd/buildah/images.go
nalind: Yeah, it does need an update.

cmd/buildah/images.go
nalind: Okay, reworked it.

cmd/buildah/images.go
nalind: Updated it.

cmd/buildah/images.go
```diff
@@ -477,23 +531,31 @@ func matchesID(imageID, argID string) bool {
...
+func matchesReference(imageName, argName string) bool {
 	if argName == "" {
 		return true
 	}
-	splitName := strings.Split(name, ":")
-	// If the arg contains a tag, we handle it differently than if it does not
+	if imageName == "" {
+		return false
+	}
+	named, err := reference.ParseNormalizedNamed(imageName)
```
mtrmac: Non-blocking: The `reference.Named` (or `nil`) could be passed down by the caller who has already done the parsing, but this works well enough.

docs/buildah-images.md
```diff
@@ -31,13 +31,16 @@ keywords are 'before', 'dangling', 'label', 'readonly' and 'since' .
...
 
   **label**
     Filter by images labels key and/or value.
 
   **readonly=true|false**
-     Show only read only images or Read/Write images. The default is to show both.  Read/Only images can be configured by modifying the  "additionalimagestores" in the /etc/containers/storage.conf file.
+    Show only read only images or Read/Write images. The default is to show both.  Read/Only images can be configured by modifying the  "additionalimagestores" in the /etc/containers/storage.conf file.
+
+  **reference=pattern**
+    Show only images whose names match the specified repository name.  If the pattern includes a tag, the image must match it as well.
```
mtrmac: Consider being more explicit about the suggested formats, it’s either repo (suffix?) or a repo:tag (but not repo@digest). (The “pattern” wording suggests to me that wildcards are supported)

cmd/buildah/images.go
```diff
@@ -299,46 +304,95 @@ func outputImages(ctx context.Context, systemContext *types.SystemContext, store
...
+					if imageDigest == digest.String() {
+						digestPresent = true
+					}
 				}
-				imagesParams = append(imagesParams, params)
-				if opts.quiet {
-					// We only want to print each id once
-					break outer
+				if !digestPresent {
+					imageDigests = append(imageDigests)
```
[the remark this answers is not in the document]
nalind: Older versions wouldn't have kept track of `Digests`, and even discarded them, but pulling by digest would still add the digested reference to the `Names` list.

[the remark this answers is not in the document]
mtrmac: Ah, right. I forgot that `Digests` was not there since the start.

docs/buildah-images.md
nalind: Adjusted the wording a bit.  There isn't much documentation around how `reference=` is supposed to work for me to use as a reference, so there's a bit of guesswork here.

TomSweeneyRedHat:
LGTM, @mtrmac do you want to do another pass through?

cmd/buildah/images.go
```diff
@@ -299,46 +304,95 @@ func outputImages(ctx context.Context, systemContext *types.SystemContext, store
...
+		}
+		if len(imageReposAndTags) == 0 {
+			imageReposAndTags = [][2]string{{"", ""}}
+		}
+		for _, imageRepoAndTag := range imageReposAndTags {
+			imagesParams = append(imagesParams, imageOutputParams{
+				ID:           imageID,
+				Name:         imageRepoAndTag[0],
+				Tag:          imageRepoAndTag[1],
+				Digest:       string(image.Digest),
```
mtrmac: This will output the primary `image.Digest` even if `imageRepoAndTag` came from a digested reference with a different digest — even if there are multiple digested references with different digests, they will all report the same `Digest:` value. Is that acceptable?

cmd/buildah/images.go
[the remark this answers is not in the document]
mtrmac: - This line does nothing
- This only extends `imageDigests` with values from `image.Names` that pass `matchesReference`. Shouldn’t it include all of them (i.e. shouldn’t this be done before the `for _, name := range image.Names` loop)?

mtrmac:
At least https://github.com/containers/buildah/pull/1972#discussion_r344750193 is still mostly outstanding.

cmd/buildah/images.go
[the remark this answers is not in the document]
mtrmac: To the second point, I’m not quite sure myself: arguably names that don’t pass the search filter should not be affect the output at all; OTOH `Digests`, if populated, does contain all of them regardless of repository names, and the PR description does say “all digests”, so I’m weakly leaning towards “include all of them regardless of filters”.

rhatdan:
@nalind Are you still working on this?

nalind:
@rhatdan yes, but I probably won't get back to it this week.

rhatdan:
@nalind Do you want someone else like @TomSweeneyRedHat to take this over?

rhatdan:
@nalind interested in reviving this one?

nalind:
Is there work in containers/common aim to replace this logic?

vrothberg:
Can we freeze this PR until the image-lib consolidation is done?  This would cause quite some merge conflicts.  Good news is that I already started migrating Buildah over, so I hope to be finished with Buildah by the end of the week.  The new library provides all core features that Buildah and Podman need but I am certain that matching all corner cases will consume time.

TomSweeneyRedHat:
I've added a WIP label so it won't merge wthout thinking.

vrothberg:
> I've added a WIP label so it won't merge wthout thinking.

Thank you! Just yesterday I moved `buildah images` over to the new library :)

PR #1980, opened 2019-11-13
info: use util.Runtime()

nalind:
We have a function that decides what our OCI runtime command is, and duplicating that logic for `buildah info` isn't necessary.


TomSweeneyRedHat:
LGTM

vrothberg:
@rh-atomic-bot r+

PR #1987, opened 2019-11-18
Touch up commit man page image parameter

TomSweeneyRedHat:
The commit command does not require that a value for the
image be provided.  Change the man page to reflect that
and explain the behavior in that instance.  This is the
same behavior as Docker.

### DIFF
--- a/docs/buildah-commit.md
+++ b/docs/buildah-commit.md
@@ -4,12 +4,14 @@
 buildah\-commit - Create an image from a working container.
 
 ## SYNOPSIS
-**buildah commit** [*options*] *container* *image*
+**buildah commit** [*options*] *container* [*image*]
 
 ## DESCRIPTION
 Writes a new image using the specified container's read-write layer and if it
 is based on an image, the layers of that image.  If *image* does not begin
-with a registry name component, `localhost` will be added to the name.
+with a registry name component, `localhost` will be added to the name.  If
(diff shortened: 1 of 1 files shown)


giuseppe:
LGTM

docs/buildah-commit.md
```diff
@@ -4,12 +4,14 @@
...
 ## SYNOPSIS
-**buildah commit** [*options*] *container* *image*
+**buildah commit** [*options*] *container* [*image*]
 
 ## DESCRIPTION
 Writes a new image using the specified container's read-write layer and if it
 is based on an image, the layers of that image.  If *image* does not begin
-with a registry name component, `localhost` will be added to the name.
+with a registry name component, `localhost` will be added to the name.  If
+*image* is not provided, the values for the `REPOSITORY` and `NAME` values of
```
vrothberg: an *image*?

I think that NAME should be TAG.

TomSweeneyRedHat: NAME/TAG yes you're right, and I swear my brain said to type TAG, but my fingers didn't follow the command.  I'm on the fence about 'an'.  We start out the prior sentence properly without it.  It's probably a bit more proper to have 'an' in this second sentence if it stood by itself, but I think it's better without it given the prior sentence.  Then again my old English teacher might beat me silly.

rhatdan:
@rh-atomic-bot r+

PR #1989, opened 2019-11-19
commit(docker): always set ContainerID and ContainerConfig

nalind:
For the sake of the Docker output format, always set the ContainerID field when committing, and if we have a non-nil Config, set ContainerConfig to the same values.  /cc @baude


TomSweeneyRedHat:
LGTM
@jwhonce @baude does this cure your ills from yesterday?  Do we need to spin up a release for Podman in the near future?

rhatdan:
@rh-atomic-bot r+

PR #1996, opened 2019-11-25
chroot: unmount with MNT_DETACH instead of UnmountMountpoints()

nalind:
Unmounting the container's bind-mounted rootfs with `MNT_DETACH` should unmount everything below it, so we don't need to use the more exhaustive method that our bind package uses for its bind mounts.

### DIFF
--- a/chroot/run.go
+++ b/chroot/run.go
@@ -15,6 +15,7 @@ import (
 	"strings"
 	"sync"
 	"syscall"
+	"time"
 	"unsafe"
 
 	"github.com/containers/buildah/bind"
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM
Although tests are not happy.

chroot/run.go
```diff
@@ -1004,10 +1005,18 @@ func setupChrootBindMounts(spec *specs.Spec, bundlePath string) (undoBinds func(
 	var fs unix.Statfs_t
 	removes := []string{}
```
TomSweeneyRedHat: Looks like we append to removes, but never use it?  I'm not sure why it's being picked up now and not before this commit.

```
./tests/tools/build/golangci-lint run

chroot/run.go:1279:3: ineffectual assignment to `removes` (ineffassign)

		removes = append(removes, roEmptyDir)
```

TomSweeneyRedHat:
Changes LGTM

TomSweeneyRedHat:
@nalind ping!  Looks like you've some lint issues on this one.

```
./tests/tools/build/golangci-lint run
chroot/run.go:1279:3: ineffectual assignment to `removes` (ineffassign)
		removes = append(removes, roEmptyDir)
```

TomSweeneyRedHat:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1772179

TomSweeneyRedHat:
LGTM, @rhatdan PTAL and merge if it still LGTY

rhatdan:
@rh-atomic-bot r+

Issue #1984, opened 2019-11-15
Request: throw error on COPY --chown x:y (without equal sign)


edsantiago:
Some time was lost just now because this is not valid syntax:
```
COPY --chown 1000:1000 myfile /home/user/
```
It looks like the _intention_ is there for buildah to throw an error on invalid flagArg in ```vendor/github.com/openshift/imagebuilder/dispatchers.go```, it's just not actually triggering.

If it's possible to fix the parser to grok `--chown uid:gid` the same as `--chown=uid:gid`, that might be the nicest option all around. If that's not possible, then perhaps buildah could throw an error such as "ERROR: --chown requires an equal sign"? (Likewise for `--from`, I suspect).

rhatdan:
Tom can you make these legal or at least match Docker.

TomSweeneyRedHat:
@edsantiago I think we may have gotten a free fix from somewhere.  I just tried with Buildah upstream:

```
# cat ~/Dockerfile.chown 
FROM alpine

COPY --chown 2367:3267 copychown.txt /tmp 
RUN stat -c "user:%u group:%g" /tmp/copychown.txt 
CMD /bin/sh

# buildah bud -f ~/Dockerfile.chown -t tom .
STEP 1: FROM alpine
STEP 2: COPY --chown 2367:3267 copychown.txt /tmp 
error building at STEP "COPY --chown 2367:3267 copychown.txt /tmp": COPY only supports the --chown=<uid:gid> and the --from=<image|stage> flags
```
Is this sufficient and/or different than what you saw?

edsantiago:
@TomSweeneyRedHat I get that error when running `buildah`, but not `podman`. My first thought was to cool my heels and wait for a revendor, but looking at `git blame` on the abovementioned `dispatchers.go` shows that this code is over a year old. It's also in podman's vendor'ed copy of `imagebuilder`. 

Going by the usual rule-of-thumb[1] makes buildah fail:
```
# buildah bud -t foo .
[ ok ]
# buildah bud --layers -t foo .
[ fails: no files found matching "/tmp/ttt/2367:3267": no such file or directory ]
```

 [1] Rule of thumb: when `podman build` does X, and `buildah bud` does Y, try adding `--layers` to buildah

TomSweeneyRedHat:
[...]

PR #1997, opened 2019-11-26
Handle missing equal sign in --from and --chown flags for COPY/ADD

TomSweeneyRedHat:
When buildah bud runs without layers (default), it hands off the parsing
of the Dockerfile steps to imagebuilder.  However if the the --layers
option is in use, then we start parsing the steps before we hand stuff
off to imagebuilder.

The COPY --chown, ADD --chown, and COPY --from commands in a Dockerfile
were failing if there was not an '=' sign after them like "COPY --chown=2374:3256".
This moves the parsing into the code that handles the layer processing.

Fixes: #1984

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -248,7 +248,7 @@ func (s *StageExecutor) volumeCacheRestore() error {
 	return nil
 }
 
-// digestContent digests any content that this next instruction would add to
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM
@nalind @edsantiago @vrothberg @QiWang19 PTAL

vrothberg:
@rh-atomic-bot r+

PR #1999, opened 2019-11-27
pull/from/commit/push: retry on most failures

nalind:
If `PullOptions`/`BuilderOptions`/`CommitOptions`/`PushOptions`/`BuildOptions` includes a `MaxRetries` value other than 0, retry operations except for (currently) connection-refused, authentication, and no-such-repository/no-such-tag errors, at a default-but-configurable interval of 5 seconds.

Set the default for `buildah pull/from/commit/push` to 3 retries at 2 second intervals.


rhatdan:
Sadly this pushed us over the travis time limit.

rhatdan:
Do you plan to expose these options?  Should these options be handled in containers.conf?

nalind:
I hadn't planned to expose the settings (the main thing is that the CLI opts in to it, but the API doesn't do it by default, so that OpenShift doesn't see a change in behavior, since they already have retry logic of their own), but we can if we need to.  Opinions?

rhatdan:
I am fine without adding CLI support.

rhatdan:
Tests are not happy.

TomSweeneyRedHat:
Changes LGTM.  I'm guessing the changes have caused our "expect to fail" test to pause and retry a number of times causing Travis to hit the 50 minute timeout.  Prior we were completing in about 46 to 47 minutes.  The Travis timeout can't be changed unless we go with a paid plan or remove Travis entirely.

rhatdan:
@nalind needs a rebase.

cmd/buildah/common.go
```diff
@@ -25,6 +25,11 @@ var (
 	needToShutdownStore = false
 )
 
+const (
+	maxPullPushRetries = 3
```
rhatdan: These should probably go into containers.conf

vrothberg: The retry semantics are specific to Buildah and are not implemented in other tools. I think that it should be implemented in containers/image and be configurable in the registries.conf.

rhatdan: Ok let's merge and we can open a different PR if we decide we need this in registries.conf.

Issue #1951, opened 2019-10-29
[low pri] [tests only] imgtype: incomplete diagnostic when wrong expected-manifest-type


edsantiago:
`imgtype` is emitting an incomplete diagnostic when the expected image type is wrong:
```
# cid=$(./buildah from scratch)

#  ./buildah commit --format oci $cid scratch-image-oci
[...]

# ./imgtype --expected-manifest-type application/vnd.docker.distribution.manifest.v2+json  scratch-image-oci
ERRO[0000] expected manifest type "application/vnd.docker.distribution.manifest.v2+json" in "scratch-image-oci", got ""
```
Note the `got ""` at far right. I expected `got "application/vnd.oci.image.manifest.v1+json"`

The converse test (for correct manifest type) works fine and is silent as expected:
```
# ./imgtype --expected-manifest-type application/vnd.oci.image.manifest.v1+json scratch-image-oci
```
Also, when the image type is docker, the diagnostic includes the actual type:
```
# ./buildah commit --format docker $cid scratch-image-docker
[...]

# ./imgtype --expected-manifest-type application/vnd.oci.image.manifest.v1+json scratch-image-docker
ERRO[0000] expected config type "application/vnd.oci.image.config.v1+json" in "scratch-image-docker", got "application/vnd.docker.container.image.v1+json"
```
(likely because they are entirely different code paths. Note also that this code path says `unexpected config type`, whereas the other says `unexpected manitest type`. Probably not a critical inconsistency).

PR #2006, opened 2019-12-04
imgtype: check earlier for expected manifest type

edsantiago:
imgtype has a --expected-manifest-type option; the intention
seems to be to fail if the given image is not of the expected
type. This was not actually working as intended: the user-
specified expected-type was being trusted prematurely and
used to set state inappropriately.

Solution: check earlier, and bail if actual != expected.

There's quite a bit of duplicate code which I think can be
refactored here but it's probably ok to defer that for
the proverbial rainy day.

Fixes: #1951

### DIFF
--- a/tests/imgtype/imgtype.go
+++ b/tests/imgtype/imgtype.go
@@ -137,6 +137,12 @@ func main() {
 			continue
 		}
 
+		if expectedManifestType != "" && manifestType != expectedManifestType {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

rhatdan:
@nalind @TomSweeneyRedHat @QiWang19 PTAL

TomSweeneyRedHat:
LGTM, thx @edsantiago !

TomSweeneyRedHat:
@rh-atomic-bot r+

PR #2012, opened 2019-12-09
Update golang.org/x/sys to eeba5f6aabab

saschagrunert:
To avoid cross compilation issues again we now vendor the latest
version of the sys package.

Refers to https://github.com/seccomp/containers-golang/pull/23#discussion_r355323842


vrothberg:
Same here. We need to stick on older versions because newer version of sys have those breaking changes which break compilation of docker/docker.

saschagrunert:
Ah, I see. Closing since this seems to be already the latest possible version here.

PR #1977, opened 2019-11-11
bud COPY does not download URL

QiWang19:
close #1846
buildah bud COPY can download the URL to the container working directory. This is not compatible with Docker.
So after this path, buildah bud COPY URL will return an error.

Signed-off-by: Qi Wang <qiwan@redhat.com>

PR #2019, opened 2019-12-12
Allow ADD to use http src

TomSweeneyRedHat:
In #1977 we disallowed the use of a URL for both the
COPY and ADD commands in a Dockerfile.  Upon further
reflection we should have done that only for the COPY
command.

Addresses: https://github.com/containers/libpod/issues/4686

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -423,38 +423,43 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
 		}
 		for _, src := range copy.Src {
 			if strings.HasPrefix(src, "http://") || strings.HasPrefix(src, "https://") {
-				// Source is a URL.
-				// returns an error to be compatible with docker
-				return errors.Errorf("source can't be a URL for COPY")
-			}
-			// Treat the source, which is not a URL, as a
-			// location relative to the
-			// all-content-comes-from-below-this-directory
-			// directory.
-			srcSecure, err := securejoin.SecureJoin(contextDir, src)
-			if err != nil {
-				return errors.Wrapf(err, "forbidden path for %q, it is outside of the build context %q", src, contextDir)
-			}
-			if hadFinalPathSeparator {
-				// If destination is a folder, we need to take extra care to
-				// ensure that files are copied with correct names (since
-				// resolving a symlink may result in a different name).
-				_, srcName := filepath.Split(src)
-				_, srcNameSecure := filepath.Split(srcSecure)
-				if srcName != srcNameSecure {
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
FYI @edsantiago @mheon 
Need to test further to get WIP off this, I suspect some of the tests that were changed in #1977 won't be happy with this change.

TomSweeneyRedHat:
I've removed the WIP, have reverted the tests to the way they used to be which is appropriate.  That includes a test with an "ADD {url}".  I'm kicking myself for not catching that removal in the prior PR.

imagebuildah/stage_executor.go
```diff
@@ -423,38 +423,43 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
-					options := buildah.AddAndCopyOptions{
-						Chown:            copy.Chown,
-						ContextDir:       contextDir,
-						Excludes:         copyExcludes,
-						IDMappingOptions: idMappingOptions,
-					}
-					if err := s.builder.Add(filepath.Join(copy.Dest, srcName), copy.Download, options, srcSecure); err != nil {
-						return err
+				// Source is a URL, allowed for ADD but not COPY.
+				if copy.Download {
```
edsantiago: Unfortunate naming :-(. It took me a while to confirm that it's correct (it is).

tests/bud.bats
```diff
@@ -65,7 +65,7 @@ load helpers
   expect_line_count 8
   run_buildah bud --signature-policy ${TESTSDIR}/policy.json --layers -t test2 ${TESTDIR}/use-layers
   run_buildah --log-level=error images -a
-  expect_line_count 9
+  expect_line_count 10
```
edsantiago: I'm completely lost as to why the image count changes here, or why it changed in 4cf37c27. Do you understand it?

imagebuildah/stage_executor.go
TomSweeneyRedHat: Completely concur.

tests/bud.bats
TomSweeneyRedHat: No I didn't chew on that, but like that's it's reverted.  If I get time today will see if I can figure it out.

TomSweeneyRedHat:
@rhatdan @vrothberg @nalind PTAL

tests/bud.bats
edsantiago: No need: time is finite, and this is a reversion to how it was before, and tests are passing. Plus, the world is always better when there's a little mystery. Thanks again @TomSweeneyRedHat

rhatdan:
@rh-atomic-bot r+

PR #2023, opened 2019-12-13
Bump 1.11.7

TomSweeneyRedHat:
Bump to v1.11.7 to get ADD url fix for Podman 1.7.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,32 @@
 
 # Changelog
 
+## v1.11.7 (2019-12-13)
+    Allow ADD to use http src
+    Bump to c/storage v.1.15.3
+    install.md: update golang dependency
+    imgtype: reset storage opts if driver overridden
+    Start using containers/common
+    overlay.bats typo: fuse-overlays should be fuse-overlayfs
+    chroot: Unmount with MNT_DETACH instead of UnmountMountpoints()
+    bind: don't complain about missing mountpoints
+    imgtype: check earlier for expected manifest type
+    Vendor containers/storage fix
+    Vendor containers/storage v1.15.1
+    Add history names support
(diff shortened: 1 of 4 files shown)


TomSweeneyRedHat:
@rhatdan @nalind @vrothberg PTAl
FYI @mheon

CHANGELOG.md
```diff
@@ -2,6 +2,32 @@
...
+    install.md: update golang dependency
+    imgtype: reset storage opts if driver overridden
+    Start using containers/common
+    overlay.bats typo: fuse-overlays should be fuse-overlayfs
+    chroot: Unmount with MNT_DETACH instead of UnmountMountpoints()
+    bind: don't complain about missing mountpoints
+    imgtype: check earlier for expected manifest type
+    Vendor containers/storage fix
+    Vendor containers/storage v1.15.1
+    Add history names support
```
vrothberg: This would qualify for a minor bump to `v1.12.0`.

rhatdan: I agree, bump the second number,  If we only fix bugs then minor.  If new features like history names then the middle.  If changes in API then major.

TomSweeneyRedHat: Missed that.  Will kill this then.  @rhatdan should I wait for the /var/tmp to land before a new kit and does the history change cause us to NOT want to get this in Podman 1.7.0?

rhatdan: I don't think you should wait for this.

TomSweeneyRedHat:
Will create v1.12.0.  The question will be do it today for Podman v1.7.0 or wait until later.

TomSweeneyRedHat:
OK, I'll go spin up 1.12 then without the /var/tmp fix.

PR #2029, opened 2019-12-16
Bats major cleanup

edsantiago:
Clean up BATS tests:
  -- remove `--log-level=error`, it's no longer necessary
  -- use `run_buildah` (instead of plain `buildah`) wherever possible
  -- use `expect_output` (instead of `test this = that`) wherever possible
  -- remove unnecessary `rm` & `rmi` cleanup steps
  -- use `$TESTDIR` for all tempfiles: do not leave droppings in current dir or in BATS dir
  -- some refactoring, cleanup, fixes (for not-actually-working-as-desired tests) (filed a few issues)

I realize this is huge. I apologize. For ease[1] of review I've split this into five separate commits, see those for details. Only the two topmost commits really need a discerning eye, the earlier three are autogenerated.

 [1] "ease" may not be the right word

### DIFF
--- a/tests/squash.bats
+++ b/tests/squash.bats
@@ -10,7 +10,7 @@ function check_lengths() {
   # matrix test: check given .Docker.* and .OCIv1.* fields in image
   for which in Docker OCIv1; do
     for field in RootFS.DiffIDs History; do
-      run_buildah --log-level=error inspect -t image -f "{{len .$which.$field}}" $image
+      run_buildah inspect -t image -f "{{len .$which.$field}}" $image
       expect_output "$expect"
     done
   done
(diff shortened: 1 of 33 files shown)


tests/squash.bats
```diff
@@ -61,53 +66,59 @@ function check_lengths() {
...
+	run_buildah mount $cid
+	mountpoint=$output
 	for stage in $(seq 10) ; do
 		cmp $mountpoint/layer${stage} ${TESTDIR}/randomfile
 	done
 
-	buildah build-using-dockerfile --signature-policy ${TESTSDIR}/policy.json --squash --layers -t squashed ${TESTDIR}/squashed
-	run_buildah --log-level=error inspect -t image -f '{{len .Docker.RootFS.DiffIDs}}' squashed
-	[ "$output" -eq 1 ]
+	run_buildah build-using-dockerfile --signature-policy ${TESTSDIR}/policy.json --squash --layers -t squashed ${TESTDIR}/squashed
```
edsantiago: Hey, so, this here has `--layers` and I kind of assumed that the subsequent tests would also, but they don't. Is that intentional?

rhatdan: Makes sense to me.

edsantiago: As in, "makes sense that the rest should also specify `--layers`", or "makes sense the way it is"?

rhatdan: All  should have --layers in this test.  Or none should have layers.  But I think we want at least some tests specifying --layers.

edsantiago:
I missed fixing the version test. Fixed and force-pushed.

tests/squash.bats
TomSweeneyRedHat: We want at least a mix of `--squash` and `--squash --layers` within the tests, we don't want one or the other for all of the tests.  For more info on the expected behavior, you can check out the description in #1900

edsantiago: I just realized I added my github comment in the wrong place. What I was struggling with is trying to figure out the difference between lines 93-107 and 109-123. Ignoring my test-name changes, the only diff between those two blocks is that line 109 has `--layers`. The rest of the `bud` steps don't. Is that intentional?

edsantiago:
Any idea why github thinks my (signed) commits aren't signed?

TomSweeneyRedHat:
@edsantiago IDK about the git hiccup.  Maybe do another 'git commit --amend -s', fetch, rebase and force push loop?

tests/squash.bats
TomSweeneyRedHat: @edsantiago over time there's been a lot of pruning done here and there.  I try to keep at least one of each option scattered about, most likely it was intentional at one point, but I don't know the why it was added (or left behind) at this point.

rhatdan:
I was playing around with some of the settings. I reverted the change.

rhatdan:
@rh-atomic-bot r+

edsantiago:
Sigh. Can someone hit retest please? Failure is `pull-all-tags` - for which I actually have a proposed PR, just as soon as this one merges.

PR #2031, opened 2019-12-17
Add support for FIPS-Mode backends

rhatdan:
If host is running in fips mode, then RHEL8.2 and beyond container images
will come with a directory /usr/share/crypto-policies/back-ends/FIPS.
This directory needs to be bind mounted over /etc/crypto-policies/back-ends in
order to make all tools in the container follow the FIPS Mode rules.

### DIFF
--- a/pkg/secrets/secrets.go
+++ b/pkg/secrets/secrets.go
@@ -148,12 +148,21 @@ func getMountsMap(path string) (string, string, error) {
 }
 
 // SecretMounts copies, adds, and mounts the secrets to the container root filesystem
+// Deprecated, Please use SecretMountWithUIDGID
 func SecretMounts(mountLabel, containerWorkingDir, mountFile string, rootless, disableFips bool) []rspec.Mount {
 	return SecretMountsWithUIDGID(mountLabel, containerWorkingDir, mountFile, containerWorkingDir, 0, 0, rootless, disableFips)
 }
 
-// SecretMountsWithUIDGID specifies the uid/gid of the owner
-func SecretMountsWithUIDGID(mountLabel, containerWorkingDir, mountFile, mountPrefix string, uid, gid int, rootless, disableFips bool) []rspec.Mount {
+// SecretMountsWithUIDGID copies, adds, and mounts the secrets to the container root filesystem
+// mountLabel: MAC/SELinux label for container content
+// containerWorkingDir: Private data for storing secrets on the host mounted in container.
+// mountFile: Additional mount points required for the container.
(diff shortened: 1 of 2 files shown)


rhatdan:
@nalind @TomSweeneyRedHat @mheon @giuseppe @vrothberg @baude PTAL

rhatdan:
Need someone from @simo5 to confirm file/directory paths.

rhatdan:
@t8m PTAL

rhatdan:
@mjahoda PTAL

t8m:
LGTM. The paths in the patch look correct.

pkg/secrets/secrets.go
```diff
@@ -323,14 +332,33 @@ func addFIPSModeSecret(mounts *[]rspec.Mount, containerWorkingDir, mountPrefix,
...
+			Source:      ctrDirOnHost,
 			Destination: secretsDir,
 			Type:        "bind",
 			Options:     []string{"bind", "rprivate"},
 		}
 		*mounts = append(*mounts, m)
 	}
 
+	srcBackendDir := "/usr/share/crypto-policies/back-ends/FIPS"
+	destDir := "/etc/crypto-policies/back-ends"
```
TomSweeneyRedHat: @rhatdan would it make sense to add the above two variables into containers.conf at some point?

TomSweeneyRedHat: Also I'm not a big fan of having a directory defined like this in a middle of a function somewhere, but personal twitch.

TomSweeneyRedHat:
Late to the party, couple small things for future consideration.

pkg/secrets/secrets.go
rhatdan: I don't know if containers.conf is the right place.  The place to extend would be mounts.conf I guess. but we would need some changes to how this is parsed.

Currently 
SRC:DEST

But SRC is on the host and DEST is in the container

This patch basically does SRC in the container to DEST in the container.

Mounts.conf also copies SRC into container local storage and then mounts Container Local Storage at DEST.

The HPC guys want additional volumes "bind" mounted into automatically into the containers.

rhatdan: I just added
```
# List of additional volumes. Specified as
# "<directory-on-host>:<derectory-in-container>:<options>", for example: 
# "/db:/var/lib/db:ro".
# If it is empty or commented out, no volumes will be added
# additional_volumes = []
```
To containers.conf, but this does not solve the issue we have here either.

PR #2032, opened 2019-12-18
Safer test for pull --all-tags

edsantiago:
The 'pull --all-tags' test flakes often in CI, e.g.:

   unexpected http code: 500 [...] URL: https://auth.docker.io/...

This is a remote registry error, not one we can resolve
on our end without complex retry-pull logic.

Here is an alternative which I believe provides better
testing anyway: instead of relying heavily on a remote
registry, do all the work using our already-set-up local
one. Pull one image (yes, sigh, from remote registry)
and push it locally with various different tags. Then
pull with --all-tags and make sure we get what we expect.

Advantages:
  + less reliance on network & remote server
  + less reliance on the _setup_ of said server, i.e.
    we don't have to just blindly trust that there
    will be multiple tagged versions of an image
  + better testing: since we know what we push to
    the local registry, we know exactly what we
    should expect to see on pull, and we now
    actually test that instead of a handwavey
    "oh just hope that there are more than three"
  + better testing of messages, and comparing image IDs

Disadvantages:
  - I've rolled two tests into one. (I'm not sure what the
    purpose was of pull-with-alltags-from-registry. Is the
    behavior that different?
  - In a development environment or some imaginary CI
    environment in which the local registry is persistent
    across runs, if someone were to remove one or more
    tags from the 'tags=(...)' list, tests would fail
    because the registry would include more tagged images
    (saved in a previous run) than the test expects. The
    diagnostic in this case would not be very helpful.

And of course, while I'm at it, improve a few of the other
tests: don't just check for error on non-docker --all-tags
pulls, actually make sure we get the expected message.

### DIFF
--- a/tests/pull.bats
+++ b/tests/pull.bats
@@ -50,9 +50,7 @@ load helpers
   run_buildah images --format "{{.Name}}:{{.Tag}}"
   expect_output --substring "alpine"
   run_buildah 1 pull --all-tags --signature-policy ${TESTSDIR}/policy.json docker-archive:${TESTDIR}/alp.tar
-  run rm -rf ${TESTDIR}/alp.tar
-  echo "$output"
-  [ "$status" -eq 0 ]
+  expect_output "Non-docker transport is not supported, for --all-tags pulling"
 }
 
 @test "pull-from-oci-archive" {
@@ -63,9 +61,7 @@ load helpers
   run_buildah images --format "{{.Name}}:{{.Tag}}"
   expect_output --substring "alpine"
   run_buildah 1 pull --all-tags --signature-policy ${TESTSDIR}/policy.json oci-archive:${TESTDIR}/alp.tar
-  run rm -rf ${TESTDIR}/alp.tar
-  echo "$output"
-  [ "$status" -eq 0 ]
+  expect_output "Non-docker transport is not supported, for --all-tags pulling"
 }
 
 @test "pull-from-local-directory" {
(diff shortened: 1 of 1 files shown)


edsantiago:
@nalind PTAL

rhatdan:
LGTM

edsantiago:
Test failure was in the docker test, which I didn't (couldn't) check on my end because I have no docker. Fixed, rebased, pushed.

jnovy:
LGTM. Thanks Ed.

edsantiago:
[Error](https://travis-ci.org/containers/buildah/jobs/626715794#L842) seems to be:
```
* "docker.io/library/busybox": Error writing blob: error storing blob to image destination for cache "containers-storage:[vfs@/var/tmp/tmp995a9d884998c6c60884b5f7/root+/var/tmp/tmp995a9d884998c6c60884b5f7/runroot]docker.io/library/busybox:latest": error storing blob to file "/var/tmp/storage681780035/1": read tcp 10.20.1.52:48954->104.18.122.25:443: read: connection reset by peer
```
Looks like a flake talking to registry. Can anyone confirm, and if so, rerun the failed test? TIA

rhatdan:
@rh-atomic-bot r+

edsantiago:
```
 * "docker.io/library/alpine": Error writing blob: error storing blob to file "/var/tmp/storage846244154/1": read tcp 10.20.1.128:53146->104.18.125.25:443: read: connection reset by peer
```
Enough is enough. I'm back to working on my prefetch-and-cache PR. It's a lot of work so I probably won't submit it until tomorrow, but the gist is:
* add a `_prefetch alpine / busybox / ubuntu` (as needed) command to each test as needed
* `_prefetch()` will run `podman load` from a cached image in $TMPDIR (or, the first time through, will `podman pull` and `podman save`).

Testing right now with `nsenter -n buildah`. So far it looks promising but there's a boatload of manual effort still left to do.

PR #2035, opened 2019-12-19
set golang.org/x/sys to b4ddaad3f8a3

vrothberg:
Everything beyond that commit breaks Windows builds due to breaking
changes in that library.  We fixed these already in c/storage but
other dependencies (e.g., Docker) still break.


vrothberg:
@rhatdan PTAL

rhatdan:
@vrothberg We have a requirement merged into buildah to handle large TMPFS that is using containers/image v5.1.0, which is why this is blowing up, I believe.

vrothberg:
Oh no, I didn't see that c/image is affected as well. We need to be more careful with dependabot. We introduced the issue we fixed in c/storage now to all other repos :(

rhatdan:
Can we get the fix into the upstream package?

vrothberg:
> Can we get the fix into the upstream package?

Yes, I'll prepare another PR for c/image and then we can start the release-and-vendor dance :) `git blame` revealed that a Valentin Rothberg updated x/sys in c/image :smile:

rhatdan:
You can never trust that guy...

vrothberg:
> You can never trust that guy...

Absolutely. All the time I have to clean up his mess :^) I had a closer look at docker and they seemed to have fixed the issue in the meantime -> https://github.com/containers/libpod/pull/4730 we're good to go in libpod. Reverting in c/image was not option because other dependencies want a more recent x/sys there - the price to pay when using go modules.

TomSweeneyRedHat:
Darwin is failing with:

```
# github.com/containers/buildah/pkg/parse
pkg/parse/parse.go:592:5: ctx.BigFilesTemporaryDir undefined (type *types.SystemContext has no field or method BigFilesTemporaryDir)
make: *** [Makefile:51: darwin] Error 2
```

Personally I never trust anyone that works while on PTO... ;^)

PR #2036, opened 2019-12-19
DO NOT MERGE Flake handling: cache and prefetch images

edsantiago:
Show of hands: who here loves submitting a PR, then coming back
hours later to find one job failed, then spending time poring
over logs and finding a network error? Anyone? Anyone?

This is a lame attempt to minimize such flakes by caching
commonly-used images and restoring them on demand. We
introduce a new helper, _prefetch(), which podman-pulls
an image the first time, podman-saves it, then on
subsequent calls (for the same image) podman-loads it:

    @test foo {
        _prefetch alpine busybox
        ...tests that run buildah-from either
    }

This is an imperfect solution: it is incomplete and will
grow more so over time as new tests are added. It is
difficult to verify its coverage. I'm really unhappy
with it but if it works, the Total Sum Of Unhappiness
might decrease overall thanks to fewer flakes. If it
doesn't work, it's trivial to remove _prefetch calls
using a sed script. Shall we give it a chance?

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -1170,7 +1170,7 @@ load helpers
 }
 
 @test "bud with FROM AS construct" {
-  _prefetch centos:7
+  _prefetch alpine
   run_buildah bud --signature-policy ${TESTSDIR}/policy.json -t test1 ${TESTSDIR}/bud/from-as
   run_buildah images -a
   expect_output --substring "test1"
@@ -1185,7 +1185,7 @@ load helpers
 }
 
 @test "bud with FROM AS construct with layers" {
-  _prefetch centos:7
+  _prefetch alpine
   run_buildah bud --layers --signature-policy ${TESTSDIR}/policy.json -t test1 ${TESTSDIR}/bud/from-as
   run_buildah images -a
   expect_output --substring "test1"
@@ -1200,6 +1200,7 @@ load helpers
 }
 
 @test "bud with FROM AS skip FROM construct" {
+  _prefetch alpine
   run_buildah bud --signature-policy ${TESTSDIR}/policy.json -t test1 -f ${TESTSDIR}/bud/from-as/Dockerfile.skip ${TESTSDIR}/bud/from-as
   expect_output --substring "LOCAL=/1"
   expect_output --substring "LOCAL2=/2"
@@ -1750,7 +1751,7 @@ load helpers
 }
 
 @test "bud containerfile with args" {
-  _prefetch centos
+  _prefetch alpine
   target=use-args
   touch ${TESTSDIR}/bud/use-args/abc.txt
   run_buildah bud --signature-policy ${TESTSDIR}/policy.json -t ${target} --build-arg=abc.txt ${TESTSDIR}/bud/use-args
@@ -1918,6 +1919,7 @@ EOM
   run_buildah rm -a
   run_buildah rmi -a -f
 
+  _prefetch alpine
   run_buildah bud --signature-policy ${TESTSDIR}/policy.json -f ${TESTSDIR}/bud/symlink/Containerfile.add-tar-gz-with-link -t testctr ${TESTSDIR}/bud/symlink
   run_buildah from testctr
   run_buildah run testctr-working-container ls /tmp/testdir/testfile.txt
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -3,6 +3,7 @@
 load helpers
 
 @test "bud with --dns* flags" {
+  _prefetch alpine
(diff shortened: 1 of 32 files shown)


edsantiago:
This is hard to review: all the _prefetch insertions were added manually after many iterations of manual code inspection and testing. I would appreciate a careful eye to make sure that I didn't accidentally add a `_prefetch` where it doesn't belong.

Tested by instrumenting `helpers.bash:run_buildah()` to `unshare -n` before running buildah and adding `_prefetch` commands where I found necessary. Not everywhere, because some of the `buildah from` and `buildah pull` tests actually _should_ test pulling from a registry. After many iterations, I've reduced the error count to 81. Many of these are noise and could be eliminated by setting up a custom network namespace allowing localhost (for fetches from local registry) and github.com. I spent far too long trying to make that work, unsuccessfully.

One question that has long troubled me and now bothers me even more: Is it absolutely necessary to use so many source images? E.g. `tests/bud/from-as` and `tests/bud/use-args` require `centos` and `registry.centos.org/centos/centos:centos7`; `tests/bud/copy-from` requires `php:7.2` and `composer`; and a ton of tests use `ubuntu`. Is there any objection to my cleaning that up (in a separate PR) and sticking with `alpine` and `busybox`, or perhaps even `scratch` if possible?

* See other FIXME comments as noted in github comments.

tests/bud.bats
```diff
@@ -483,6 +507,7 @@ load helpers
 }
 
 @test "bud-http-context-dir-with-Dockerfile-post" {
+  # FIXME FIXME FIXME: this is 100% identical to the -pre test above.
```
edsantiago: If you click on the expando (more context) above, you'll see that this test is identical to the `-pre` one. What was the intention here? Is it OK to collapse them both into one?

tests/bud.bats
```diff
@@ -1662,12 +1772,14 @@ load helpers
 
 @test "bud using gitrepo and branch" {
   target=gittarget
+  # FIXME: this test takes a really long time. Is it necessary to do twice?
```
edsantiago: almost 4 minutes, because it uses fedora as base image and runs a dnf install. Is there some other simpler test we can do instead? Can we create a new module under `github.com/containers` just for testing this? Or even just a side branch of `containers/BuildSourceImage` that just does `FROM scratch` and `RUN echo got here`?

edsantiago:
`podman: command not found`. Yikes - I thought podman was a requirement for these tests? Like, how is the local registry run? [pause] OMG these tests are using docker. I assume that will need to be corrected pretty soon, so let's just leave this PR on hold until CI is fixed to use podman.

tests/bud.bats
rhatdan: How about switching it to:
run_buildah bud --format docker --layers -f tests/bud/shell/Dockerfile -t test git://github.com/containers/buildah#master

rhatdan: Runs in 3.5 seconds.

tests/bud.bats
rhatdan: I have no idea.  @TomSweeneyRedHat @nalind Any ideas?

rhatdan:
I don't see a problem with consolidating on a couple of images.  I do think it is handy to make sure we still have tests with apt-get and dnf/yum.   But getting this to the point that we test fewer distinct images would be fine.

rhatdan:
@edsantiago whats the scoop on this one?

edsantiago:
@rhatdan right now I'm just playing with it, wanting to see if/how it runs in CI, if it saves any time, if it reduces flakes. I've changed the title accordingly to `DO NOT MERGE`. Right now CI is running tests using docker, which obviously needs to be fixed. Once that gets resolved, I'll revert the `fixup` commit and submit for real review.

edsantiago:
Never mind: `buildah push/pull docker-archive:` does not work the way I expected (`pull` loses names). This can wait until CI is fixed to use podman instead of docker.

edsantiago:
Oh yikes. This was not supposed to happen: it has not been reviewed.

edsantiago:
@nalind @TomSweeneyRedHat @rhatdan I'm sorry to request this, but could you find time to post-review this and let me know if you find glaring problems? It got merged last week due to my misunderstanding about how `bors` works. If you find any issues with it, or if this is not wanted at all in the tests, I will submit PRs to fix and/or undo.

rhatdan:
I am fine with the change, my only concern would be, are we still exercising buildah pulling images from a registry?

edsantiago:
Yes. None of the tests in `pull.bats` use `_prefetch`. Also, there are many instances in which I deliberately did not add `_prefetch`. (Can't remember any instances offhand, because it's been a while, but there were several tests where I thought it would be better to let buildah do the pulling).

rhatdan:
Ok lets let it go for now, and we will see if we have any issues going forward.

Issue #2007, opened 2019-12-05
Allow changing FROM image on command line


tnozicka:
**Description**

It is often the case that I, and other people in the OpenShift org, need to change the image that is used in the Dockerfile and use a different one. OpenShift builds are based on that idea and given you have input imagestreams it preprocesses the Dockerfile and replaces the FROM line. The way how we would use such dockerfiles locally is to use https://github.com/openshift/imagebuilder to build it like:
```
imagebuilder --from registry.svc.ci.openshift.org/ocp/ubi:7 -mount ~/projects/origin-repos/4.3/ocp-4.3-default.repo:/etc/yum.repos.d/ocp-4.3-default.repo -mount ~/projects/origin-repos/4.3/ops-mirror.pem:/etc/yum.repos.d/ops-mirror.pem -t openshift-base -f base/Dockerfile base/
```
unfortunately using imagebuilder implies using Docker and being root (https://github.com/openshift/imagebuilder/issues/128). I really like using podman/buildah and being just a regular user, but this use cases forces me back into Docker and being root.

**Describe the results you expected:**
support `--from` flag in buildah and `podman build` to be able to replace the FROM image in Dockerfile

cc @dustymabe @vrothberg

vrothberg:
Thanks for opening the issue, @tnozicka! I think this fairly easy to support and a quite common use-case in the OpenShift land.

@TomSweeneyRedHat @nalind @rhatdan WDYT?

TomSweeneyRedHat:
@tnozicka just to make sure I'm on the same page, given a Dockerfile like:
```
FROM alpine
ARG FIRSTARG
```

Then a command like "buildah bud --from busybox" would in effect make the processed Dockerfile look like:
```
FROM busybox
ARG FIRSTARG
```

But the actual Dockerfile file on disk would not change?  If that's the case, I think it's doable.

tnozicka:
> But the actual Dockerfile file on disk would not change?

correct, no disk changes

thanks for looking into it!

tnozicka:
```
FROM alpine AS alpine-stage
```
Aliases for multi stage build like `alpine-stage` would still retain its name even when running `buildah bud --from busybox` , right?

TomSweeneyRedHat:
@tnozicka I'm glad you brought that up, I probably would have just wiped the entire thing.  But we certainly could retain the "AS alpine-stage" portion of the FROM statement if it was there.  Just a little bit more processing, nothing major.

[...]

PR #2037, opened 2019-12-19
Add the --from option to bud

TomSweeneyRedHat:
When processing a Dockerfile supplied to the bud command,
replace any image name after the FROM statement with the
value passed in to the --new-from option.  This will not
change the value in the on-disk version of the Dockerfile,
only the one in memory for the lifetime of the build
process.

If the image name is a variable, do not replace that.

I also found our bash completions were missing several and a bad alpha ordering in the man page which has been corrected.
Addresses: #2007

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -338,6 +338,7 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budResults) error {
 		Target:                  iopts.Target,
 		TransientMounts:         transientMounts,
 		Devices:                 devices,
+		NewFromImage:            iopts.NewFromImage,
 	}
 
 	if iopts.Quiet {
--- a/contrib/completions/bash/buildah
+++ b/contrib/completions/bash/buildah
@@ -360,6 +360,11 @@ return 1
 
  _buildah_bud() {
      local boolean_options="
+     --cache-from
+     --compress
+     --disable-compression
+     --disable-content-trust
+     -D
      --help
      -h
      --layers
@@ -369,6 +374,7 @@ return 1
      --pull-never
      --quiet
      -q
+     --rm
      --squash
      --tls-verify
   "
@@ -378,6 +384,7 @@ return 1
      --annotation
      --authfile
      --build-arg
+     -c
      --cap-add
      --cap-drop
      --cert-dir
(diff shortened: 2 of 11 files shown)


rhatdan:
I think the option should just be --from.  No reason for the --new- part.
Completions need to be updated.
Does not look like you have a test for this?
Did you try this with a complex command line?
FROM FOO as bar
FROM $ARG

TomSweeneyRedHat:
I'd worked up a test but neglected to add.  Will add that and touch up the other stuff too.

TomSweeneyRedHat:
@rhatdan happy green tests buttons and I think I've covered your concerns.

rhatdan:
I thought the idea was only to change the initial from line?  Not all of the from lines?

TomSweeneyRedHat:
Might be my misunderstanding, but I thought they wanted to replace all.  @tnozicka @dustymabe thoughts/input?

dustymabe:
> Might be my misunderstanding, but I thought they wanted to replace all. @tnozicka @dustymabe thoughts/input?

I haven't dealt with multi-stage builds much so I don't know if I have a good answer. I had been only thinking about the initial from line.

tnozicka:
Currently in OpenShift BuildConfigs/Builds it replaces the last FROM in the Dockerfile as that is likely the runtime image. I came to know only recently.

rhatdan:
@tnozicka Do  you think that is the correct behaviour?

tnozicka:
For simple FROM replacement, replacing the last FROM seems like a good choice to me, given it is the runtime image. 

The other option is generic replacement like `--replace-image='alias:my_image' --replace-image='registry/repository/image:tag:my_image' (probably with a better separator then `:`)

rhatdan:
@TomSweeneyRedHat What is the scoop on this PR?

TomSweeneyRedHat:
I've just been thinking about it.  Replacing all the FROM targets was relatively easy, replacing the last one is not something I could figure out quickly code wise.  Need to spend some time playing and haven't gotten back to it.

rhatdan:
@TomSweeneyRedHat Needs a rebase, should we get back to this one?

TomSweeneyRedHat:
I should, it just hasn't bubbled up.

rhatdan:
buildah bud --form exists now.

Issue #2000, opened 2019-11-28
`buildah bud` does not accept multiple options for the `--volume` flag


aucampia:
**Description**

The [man page for buildah bud](https://github.com/containers/buildah/blob/v1.9.0/docs/buildah-bud.md) suggests that the `--volume` flag can take multiple options:

> The OPTIONS are a comma delimited list and can be ...

However when I actually try use multiple options (e.g. `--volume "${HOME}/.cache/zae9ujei:/myvol:rw,Z"`) `buildah bud` complains that ...

```
incorrect volume format "Z", should be host-dir:ctr-dir[:option
```

**Steps to reproduce the issue:**

Run the following commands:

```
{
mkdir -vp /var/tmp/zae9ujei

cat << EOF > /var/tmp/zae9ujei/Dockerfile
FROM docker.io/alpine:latest
RUN \
  echo AOK > /myvol/status && \
  ls -al /myvol/ && \
  grep -H . /myvol/status && \
  true
EOF

mkdir -vp "${HOME}/.cache/zae9ujei"
rm -fv "${HOME}/.cache/test/status"
(cd /var/tmp/zae9ujei && buildah bud --volume "${HOME}/.cache/zae9ujei:/myvol:rw,Z" -t zae9ujei:latest .)

grep -H . "${HOME}/.cache/zae9ujei/status"

\rm -rv /var/tmp/zae9ujei "${HOME}/.cache/zae9ujei"
}
```

**Describe the results you received:**

<pre>mkdir: created directory &apos;/var/tmp/zae9ujei&apos;
mkdir: created directory &apos;/home/E1258170/.cache/zae9ujei&apos;
incorrect volume format &quot;Z&quot;, should be host-dir:ctr-dir[:option]
<font color="#CC0000">ERRO</font>[0000] exit status 1                                
grep: /home/E1258170/.cache/zae9ujei/status: No such file or directory
removed &apos;/var/tmp/zae9ujei/Dockerfile&apos;
removed directory &apos;/var/tmp/zae9ujei&apos;
removed directory &apos;/home/E1258170/.cache/zae9ujei&apos;
</pre>

**Describe the results you expected:**

I expect the volume format I supplied to be correct.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.9.0-5.module+el8.1.0+4240+893c1ab8.x86_64
```

**Output of `buildah version`:**

```
Version:         1.9.0
Go Version:      go1.12.8
[...]

PR #2039, opened 2019-12-21
Fix option handling for volumes in build

TomSweeneyRedHat:
If the volume option was specified like: `--volume "${HOME}/.cache/zae9ujei:/myvol:rw,Z"`,
the COBRA code that processed the input from the user was considering the value to be a
SliceVar with two volumes `${HOME}/.cache/zae9ujei:/myvol:rw` and `Z` due to the comma.

Converted the COBRA code to consider the input as an ArrayVar instead and the value is
handled appropriately.  Also increased the testing to catch this going forward.

Addresses: #2000

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -1350,6 +1350,10 @@ load helpers
 @test "buidah bud --volume" {
   run_buildah bud --signature-policy ${TESTSDIR}/policy.json -v ${TESTSDIR}:/testdir ${TESTSDIR}/bud/mount
   expect_output --substring "/testdir"
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json -v ${TESTSDIR}:/testdir:rw ${TESTSDIR}/bud/mount
+  expect_output --substring "/testdir"
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json -v ${TESTSDIR}:/testdir:rw,z ${TESTSDIR}/bud/mount
+  expect_output --substring "/testdir"
 }
 
 @test "bud-copy-dot with --layers picks up changed file" {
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
@mheon @baude we may want to spin a Bulidah v1.12.1 to get this into Podman 1.7.0....

rhatdan:
Nice catch.
LGTM
@giuseppe @vrothberg @nalind @mheon @saschagrunert @edsantiago PTAL

tests/bud.bats
```diff
@@ -1350,6 +1350,10 @@ load helpers
 @test "buidah bud --volume" {
   run_buildah bud --signature-policy ${TESTSDIR}/policy.json -v ${TESTSDIR}:/testdir ${TESTSDIR}/bud/mount
   expect_output --substring "/testdir"
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json -v ${TESTSDIR}:/testdir:rw ${TESTSDIR}/bud/mount
+  expect_output --substring "/testdir"
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json -v ${TESTSDIR}:/testdir:rw,z ${TESTSDIR}/bud/mount
```
edsantiago: P{ossibly,robably} stupid question: will this work on systems where buildah is compiled without the `selinux` tag? Before asking, I thought well duh let's just look for other instances of `:[zZ]` in the tests... and there are none.

TomSweeneyRedHat:
@edsantiago honestly IDK.  It's probably Sunday morning pre-tea brain, but I'm not sure how to test the no selinux label?  Happy to try and/or add a test if you've suggestions.

rhatdan:
The flags are ignored on SELinux disabled machines.

The :Z or :z eventually triggers a lable.Relabel call in go-selinux
```
// Relabel changes the label of path to the filelabel string.
// It changes the MCS label to s0 if shared is true.
// This will allow all containers to share the content.
func Relabel(path string, fileLabel string, shared bool) error {
	if !selinux.GetEnabled() {
		return nil
	}
...
```

rhatdan:
@rh-atomic-bot r+

PR #2055, opened 2019-12-28
discard outputs coming from onbuild command on buildah-from --quiet

### LINKED ISSUE #2014: buildah-from with ONBUILD and --quiet: is not quiet
This seems to be an undefined corner case:
```console
# cat Dockerfile
FROM alpine
ONBUILD RUN date
# buildah bud --format=docker --quiet -t foo .
baa08aa3383dc647dd543aed0c5d2e8bae1d7a63b060e869dbc68524ea862ee1
# buildah from --quiet foo
STEP 1: RUN date                 <<< unexpected
Tue Dec 10 21:56:15 UTC 2019     <<< unexpected
foo-working-container            <<< expected
```
Note that there are two different unexpected lines in the output: the `STEP` one, and the actual output of the command itself.

`buildah-from(1)` describes `--quiet` as follows:
```
   --quiet, -q

   If an image needs to be pulled from  the  registry,  suppress  progress output.
```
...so technically speaking the `ONBUILD` behavior is undefined; but I think one can have a reasonable expectation that `--quiet` from `bud` will carry over onto `from`. Plus, docker (19.0.3 - I can't get a RHEL7 virt to test 13.c) seems to respect `--quiet`:
```console
# docker build --quiet .
sha256:7634f230b6506afeb29e8dbfce059fecfdb7f71ce50ff19dcf43f676ad81ce79
```

boaz0:
closes #2014

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -860,7 +860,7 @@ load helpers
   run_buildah inspect --format '{{printf "%q" .Docker.Config.OnBuild}}' ${target}
   expect_output '["RUN touch /onbuild1" "RUN touch /onbuild2"]'
   run_buildah from --quiet ${target}
-  cid=${lines[2]}              # FIXME FIXME FIXME: see issue #2014
+  cid=${lines[0]}
   run_buildah mount ${cid}
   root=$output
 
@@ -875,7 +875,7 @@ load helpers
   run_buildah inspect --format '{{printf "%q" .Docker.Config.OnBuild}}' ${target}
   expect_output '["RUN touch /onbuild3"]'
   run_buildah from --quiet ${target}
-  cid=${lines[1]}              # FIXME FIXME FIXME: see issue #2014
+  cid=${lines[0]}
   run_buildah mount ${cid}
   root=$output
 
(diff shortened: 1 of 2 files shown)


tests/bud.bats
```diff
@@ -860,7 +860,7 @@ load helpers
   run_buildah inspect --format '{{printf "%q" .Docker.Config.OnBuild}}' ${target}
   expect_output '["RUN touch /onbuild1" "RUN touch /onbuild2"]'
   run_buildah from --quiet ${target}
-  cid=${lines[2]}              # FIXME FIXME FIXME: see issue #2014
+  cid=${lines[0]}              # FIXME FIXME FIXME: see issue #2014
```
rhatdan: Why not remove the comment here?

rhatdan:
LGTM once the test comment is fixed.

tests/bud.bats
boaz0: Oops missed that part :rofl:

rhatdan:
@nalind @QiWang19 @TomSweeneyRedHat @vrothberg PTAL

QiWang19:
LGTM

rhatdan:
@rh-atomic-bot r+

rhatdan:
Thanks @boaz0

PR #2064, opened 2020-01-06
fix bug Add check .dockerignore COPY file

### LINKED ISSUE #2060: .dockerignore not working as expected in Podman 1.6.2
As reported by Robert Day in the podman.io mailing list on Jan 3, 2019:

On Fri, 3 Jan 2020, Robert P. J. Day wrote:

>
>   almost certainly a dumb question, but i'm unclear on how
> .dockerignore interacts with podman in the sense of not passing
> certain objects as part of the "build context."
>
>   with docker, entries in .dockerignore are explicitly not added to
> the context passed to the daemon, so any attempt to, say, COPY or ADD
> ignored entries in the Dockerfile are doomed to failure. i tested this
> first with docker (on ubuntu) by copying /etc/{passwd,group} to a
> build directory, with a Dockerfile that did:
>
>   COPY passwd group /
>
> and that worked fine. when i added in .dockerignore the single line
>
>   passwd
>
> predictably, the COPY failed not being able to find passwd. so far, so
> good.
>
>   when i tried the same on fedora 31 with podman 1.6.2, the COPY
> worked fine. my early assumption was that podman would ignore
> .dockerignore since there is no daemon to which to send the bundled
> context, but i've also read elsewhere that podman does *something*
> with .dockerignore, although i'm not sure what. and "man podman-build"
> makes no mention of .dockerignore.

  just to be clear, i *understand* the value of .dockerignore when
using docker, given how the build context is bundled and sent to the
daemon. it's not clear what value this has WRT podman, which has no
such daemon.

  more to the point, as long as i did my testing correctly, this means
that you will get different results in building using docker versus
using podman -- docker will fail if asked to ADD or COPY a file that
was explicitly .dockerignore'd, while podman will (apparently)
succeed.

  in any event, i downloaded the buildah source and noticed numerous
references to ".dockerignore", so it's clear that *something* is being
done with that file, i just don't know what, and i haven't seen any
useful documentation clarifying that.

rday
_______________________________________________
Podman mailing list -- podman@lists.podman.io
To unsubscribe send an email to podman-leave@lists.podman.io

QiWang19:
fix #2060 
fix the bug that buildah does not check .dockerignore file if the source of COPY instruction is a filepath. This patch will skip the path if the overall result of the dockerignore matcher is true.

### DIFF
--- a/add.go
+++ b/add.go
@@ -344,6 +344,18 @@ func (b *Builder) addHelper(excludes *fileutils.PatternMatcher, extract bool, de
 			}
 
 			b.ContentDigester.Start("file")
+			// This source is a file
+			// Check if the path matches the .dockerignore
+			if excludes != nil {
+				res, err := excludes.MatchesResult(esrc)
+				if err != nil {
+					return errors.Wrapf(err, "error checking if %s is an excluded path", esrc)
+				}
+				// Skip the file if the pattern matches
+				if res.IsMatched() {
+					return nil
+				}
+			}
 
 			if !extract || !archive.IsArchivePath(esrc) {
 				// This source is a file, and either it's not an
--- a/tests/bud/dockerignore3/Dockerfile
+++ b/tests/bud/dockerignore3/Dockerfile
@@ -1,4 +1,5 @@
 FROM busybox
 COPY . /upload/
 COPY src /upload/src2/
+COPY test1.txt /upload/test1.txt
 RUN echo "CUT HERE"; /bin/find /upload | LANG=en_US.UTF-8 sort; echo "CUT HERE"


rhatdan:
This looks good to me, but their looks like you have a serius issue in the tests.

add.go
```diff
@@ -345,6 +345,17 @@ func (b *Builder) addHelper(excludes *fileutils.PatternMatcher, extract bool, de
 
 			b.ContentDigester.Start("file")
 
+			if excludes != nil {
```
vrothberg: Can we add a comment here explaining what we're doing?

QiWang19: fixed

rhatdan:
@rh-atomic-bot r+

PR #2065, opened 2020-01-06
buildah bud --volume: run from tmpdir, not source dir

edsantiago:
PR #2039 broke system tests, because they're installed in /usr:
```console
# buildah bud -v /usr/share/buildah/test/system:/testdir:rw,z /usr/share/buildah/test/system/bud/mount
...
error building at STEP "RUN mount": error resolving mountpoints for container "173c5e567e95f2604b5ea677f5e5364839d5b455a9081cdb4101f20242997e5e": relabeling "/usr/share/buildah/test/system" failed: relabeling content in /usr is not allowed
```
Solution: mount a volume from TESTDIR (singular), which is in TMPDIR,
not TESTSDIR (plural), which is our test source dir.

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -1348,11 +1348,14 @@ load helpers
 }
 
 @test "buidah bud --volume" {
-  run_buildah bud --signature-policy ${TESTSDIR}/policy.json -v ${TESTSDIR}:/testdir ${TESTSDIR}/bud/mount
+  voldir=${TESTDIR}/bud-volume
+  mkdir -p ${voldir}
+
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json -v ${voldir}:/testdir ${TESTSDIR}/bud/mount
   expect_output --substring "/testdir"
-  run_buildah bud --signature-policy ${TESTSDIR}/policy.json -v ${TESTSDIR}:/testdir:rw ${TESTSDIR}/bud/mount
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM, but this might indicate a different bug.  Not sure why we are protecting all labelling of /usr/

```
# podman run -ti -v /usr/tmp/dan:/tmp:Z alpine sh
Error: relabel failed "/usr/tmp/dan": relabeling content in /usr is not allowed
```
I will have to examine why this is.

rhatdan:
Ok, I found this code in the go-selinux bindings.
```
// Prevent users from relabing system files
func badPrefix(fpath string) error {
	if fpath == "" {
		return ErrEmptyPath
	}

	badPrefixes := []string{"/usr"}
	for _, prefix := range badPrefixes {
		if strings.HasPrefix(fpath, prefix) {
			return fmt.Errorf("relabeling content in %s is not allowed", prefix)
		}
	}
	return nil
}
```
Seems like someone else added this, or perhaps this is my alzheimer's kicking in.

rhatdan:
@nalind @TomSweeneyRedHat @QiWang19 @vrothberg @giuseppe PTAL

edsantiago:
It seems to date back to the initial commit: https://github.com/opencontainers/selinux/commit/34bec73eb681d85ad5a28bc5d976f084c858c45e

vrothberg:
@rh-atomic-bot r+

Issue #2067, opened 2020-01-07
Image build-time regression between buildah 1.11.5 and 1.11.6


jmencak:
**Description**

Version 1.11.5 seems to be the last buildah version which doesn't cause any huge (~80s) delays between COPY and RUN steps when building images with larger amounts of vendored-in files (e.g. OpenShift operator images).

**Steps to reproduce the issue:**
1. `git clone https://github.com/openshift/cluster-node-tuning-operator`
2. `cd cluster-node-tuning-operator`
3. `make local-image IMAGE_TAG=origin USE_BUILDAH=1`

**Describe the results you received:**
80s delay between `STEP 3: COPY . .` and `STEP 4: RUN make build`

**Describe the results you expected:**
A reasonable delay between the steps as in podman version 1.11.5 and below.

**Output of `rpm -q buildah` or `apt list buildah`:**
```
buildah-1.12.0-2.fc31.x86_64 # buildah 1.12 and the latest 1.14 has the same issue; for the demo using 1.11.6 (see below)
```

**Output of `buildah version`:**
Compiled from source on F31.

```
Version:         1.11.6
Go Version:      go1.13.5
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.1-dev
CNI Spec:        0.4.0
libcni Version:  v0.7.1
image Version:   5.0.0
Git Commit:      9513cb8c
Built:           Tue Jan  7 09:37:18 2020
OS/Arch:         linux/amd64
```

**Output of `cat /etc/*release`:**

```
Fedora release 31 (Thirty One)
NAME=Fedora
VERSION="31 (Thirty One)"
ID=fedora
VERSION_ID=31
VERSION_CODENAME=""
PLATFORM_ID="platform:f31"
PRETTY_NAME="Fedora 31 (Thirty One)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:31"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f31/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=31
[...]

PR #2070, opened 2020-01-08
Don't setup excludes, if their is only one pattern to match

rhatdan:
We are always adding .dockerignore to the pattern list, if this is the only pattern
then no patterns we added to the list, and we should return nil.

This is causing a major slowdown in buildah, since it is not using the optimized tar for
copying.

### DIFF
--- a/add.go
+++ b/add.go
@@ -236,7 +236,7 @@ func dockerIgnoreMatcher(lines []string, contextDir string) (*fileutils.PatternM
 		patterns = append(patterns, includeFlag+filepath.Join(contextDir, ignoreSpec))
 	}
 	// if there are no patterns, save time by not constructing the object
-	if len(patterns) == 0 {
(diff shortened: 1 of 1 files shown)


rhatdan:
Fixes: https://github.com/containers/buildah/issues/2067

rhatdan:
@saschagrunert @QiWang19 @nalind @TomSweeneyRedHat @jmencak PTAL

jmencak:
I can confirm this fixes https://github.com/containers/buildah/issues/2067

rhatdan:
@rh-atomic-bot -r=saschagrunert

rhatdan:
@TomSweeneyRedHat I think it might be time for a new buildah release.

rhatdan:
@rh-atomic-bot r=saschagrunert

PR #2071, opened 2020-01-08
Add codespell support

rhatdan:
Check codespell on each PR in make validate.

Add fixes for issues found with codespell.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,6 +68,9 @@ docs: install.tools ## build the docs on the host
 gopath:
 	test $(shell pwd) = $(shell cd ../../../../src/github.com/containers/buildah ; pwd)
 
+codespell:
+	codespell -S build,buildah,buildah.spec,imgtype,AUTHORS,bin,vendor,.git,go.sum,CHANGELOG.md,changelog.txt,seccomp.json,.cirrus.yml,"*.xz,*.gz,*.tar,*.tgz,*ico,*.png,*.1,*.5,*.orig,*.rej" -L uint,iff,od
+
 .PHONY: validate
 validate: install.tools
 	@./tests/validate/whitespace.sh
--- a/tests/conformance/bud_test.go
+++ b/tests/conformance/bud_test.go
@@ -114,8 +114,8 @@ var _ = Describe("Buildah build conformance test", func() {
 				Expect(br.MatchString(buildahoutput)).To(BeTrue(),
 					strings.Replace(errMsg, "FAILEDREASON", "Buildah Output is not as expect.", -1))
 				if test.BuildahErrRegex != "" {
-					bre := regexp.MustCompile(test.BuildahErrRegex)
-					Expect(bre.MatchString(buildaherr)).To(BeTrue(),
+					bregex := regexp.MustCompile(test.BuildahErrRegex)
+					Expect(bregex.MatchString(buildaherr)).To(BeTrue(),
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -68,8 +68,11 @@ docs: install.tools ## build the docs on the host
 gopath:
 	test $(shell pwd) = $(shell cd ../../../../src/github.com/containers/buildah ; pwd)
 
+codespell:
+	codespell -S build,buildah,imgtype,AUTHORS,bin,vendor,.git,go.sum,CHANGELOG.md,changelog.txt,seccomp.json,.cirrus.yml,"*.xz,*.gz,*.tar,*.tgz,*ico,*.png,*.1,*.5,*.orig,*.rej" -L uint,iff,od
+
 .PHONY: validate
-validate: install.tools
+validate: install.tools codespell
```
TomSweeneyRedHat: same concern here as in Podman for having this as a gate rather than a warning.

contrib/rpm/buildah.spec
```diff
@@ -240,7 +240,7 @@ make DESTDIR=%{buildroot} PREFIX=%{_prefix} install install.completions
 
 * Wed Sep 11, 2019 Tom Sweeney <tsweeney@redhat.com> 1.11.1-1
 - Add --devices flag to bud and from
-- Downgrade .papr to highest atomic verion
+- Downgrade .papr to highest atomic version
```
TomSweeneyRedHat: I'd rather not change these lines in here, the  changelog.txt or the  CHANGELOG.md file.  The right or wrong, the commits went in with these mis-spellings and if we change them here, it will be harder to find them 2 years down the road.

tests/conformance/bud_test.go
```diff
@@ -114,8 +114,8 @@ var _ = Describe("Buildah build conformance test", func() {
 				Expect(br.MatchString(buildahoutput)).To(BeTrue(),
 					strings.Replace(errMsg, "FAILEDREASON", "Buildah Output is not as expect.", -1))
 				if test.BuildahErrRegex != "" {
-					bre := regexp.MustCompile(test.BuildahErrRegex)
-					Expect(bre.MatchString(buildaherr)).To(BeTrue(),
+					be := regexp.MustCompile(test.BuildahErrRegex)
```
TomSweeneyRedHat: I much prefer the 'bre' here.

contrib/rpm/buildah.spec
rhatdan: The problem is this becomes I will have to eliminate buildah.spec from the check like I have CHANGELOG.md

tests/conformance/bud_test.go
rhatdan: How about bregex?

Makefile
rhatdan: I have no problem if we move it down from a requirement, if it proves to be a pain in the ass, but in my opinion for now, it is not a hardship.

tests/conformance/bud_test.go
TomSweeneyRedHat: That works for me.

rhatdan:
@vrothberg @QiWang19 @nalind PTAL
I am no longer blocking on codespell, because of travis.  codespell on the ubuntu versions in there is way out of date. Maybe if we convert validate tests to cirrus we can start gating on codespell.

QiWang19:
LGTM

PR #2072, opened 2020-01-08
copy: check for .dockerignore specifically

nalind:
When generating the list of exclusions to process .dockerignore contents, don't include .dockerignore if we don't have a .dockerignore file in the context directory.  That way, if the file doesn't exist, and the caller didn't pass in any patterns, we get no patterns instead of just one ".dockerignore" pattern, and we can hit the faster copy path.  This should help with https://bugzilla.redhat.com/show_bug.cgi?id=1788982.

Don't start digesting the contents of any file that we end up skipping.

### DIFF
--- a/add.go
+++ b/add.go
@@ -215,7 +215,12 @@ func dockerIgnoreMatcher(lines []string, contextDir string) (*fileutils.PatternM
 	if contextDir == "" {
 		return nil, nil
 	}
-	patterns := []string{".dockerignore"}
+	// If there's no .dockerignore file, then we don't have to add a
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

rhatdan:
@nalind I think this conflicts with the patch I just got in.

I think you need to revert: https://github.com/containers/buildah/pull/2070

nalind:
@rhatdan it overlaps, so I've tweaked it a little.

giuseppe:
@rh-atomic-bot r+

Issue #132, opened 2017-06-02
Need to support adding signatures when committing or pushing


nalind:
Similar in some ways to #96, we need to be able to add signatures to images when writing them.

rhatdan:
@TomSweeneyRedHat I think could use something like this for OpenShift?

TomSweeneyRedHat:
Closing this as we're tracking in Trello at: https://trello.com/c/CsSZ1z8W/827-buildah-support-adding-signatures-when-committing-or-pushing

PR #2085, opened 2020-01-15
add --sign-by to bud/commit/push, --remove-signatures for pull/push

nalind:
Add the --sign-by option to `buildah build-using-dockerfile`, `buildah commit`, `buildah push`, and `buildah manifest push`.  Add the `--remove-signatures` option to `buildah pull`, `buildah push`, and `buildah manifest push`.  We just pass them to the image library, which does all of the heavy lifting.

### DIFF
--- a/cmd/buildah/manifest.go
+++ b/cmd/buildah/manifest.go
@@ -38,8 +38,8 @@ type manifestAnnotateOpts = struct {
 }
 type manifestInspectOpts = struct{}
 type manifestPushOpts = struct {
-	purge, quiet, all, tlsVerify                          bool
-	authfile, certDir, creds, digestfile, signaturePolicy string
+	purge, quiet, all, tlsVerify, removeSignatures                bool
+	authfile, certDir, creds, digestfile, signaturePolicy, signBy string
 }
 
 func init() {
@@ -194,6 +194,8 @@ func init() {
 	flags.StringVar(&manifestPushOpts.certDir, "cert-dir", "", "use certificates at the specified path to access the registry")
 	flags.StringVar(&manifestPushOpts.creds, "creds", "", "use `[username[:password]]` for accessing the registry")
 	flags.StringVar(&manifestPushOpts.digestfile, "digestfile", "", "after copying the image, write the digest of the resulting digest to the file")
+	flags.BoolVarP(&manifestPushOpts.removeSignatures, "remove-signatures", "", false, "don't copy signatures when pushing images")
+	flags.StringVar(&manifestPushOpts.signBy, "sign-by", "", "sign the image using a GPG key with the specified `FINGERPRINT`")
 	flags.StringVar(&manifestPushOpts.signaturePolicy, "signature-policy", "", "`pathname` of signature policy file (not usually used)")
 	if err := flags.MarkHidden("signature-policy"); err != nil {
 		panic(fmt.Sprintf("error marking signature-policy as hidden: %v", err))
@@ -642,6 +644,8 @@ func manifestPushCmd(c *cobra.Command, args []string, opts manifestPushOpts) err
 		SystemContext:      systemContext,
 		ImageListSelection: cp.CopySpecificImages,
 		Instances:          nil,
+		RemoveSignatures:   opts.removeSignatures,
+		SignBy:             opts.signBy,
 	}
 	if opts.all {
 		options.ImageListSelection = cp.CopyAllImages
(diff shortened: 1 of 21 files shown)


nalind:
Fixes #132.

cmd/buildah/manifest.go
```diff
@@ -194,6 +194,8 @@ func init() {
 	flags.StringVar(&manifestPushOpts.certDir, "cert-dir", "", "use certificates at the specified path to access the registry")
 	flags.StringVar(&manifestPushOpts.creds, "creds", "", "use `[username[:password]]` for accessing the registry")
 	flags.StringVar(&manifestPushOpts.digestfile, "digestfile", "", "after copying the image, write the digest of the resulting digest to the file")
+	flags.BoolVarP(&manifestPushOpts.removeSignatures, "remove-signatures", "", false, "don't copy signatures when pushing image")
```
rhatdan: removesignatures seems like a bad name for this?  Could it be copy-signatures and default to true?

common.go
```diff
@@ -28,11 +29,19 @@ func getCopyOptions(store storage.Store, reportWriter io.Writer, sourceSystemCon
 	if destinationSystemContext != nil {
 		*destinationCtx = *destinationSystemContext
 	}
+	if removeSignatures {
+		logrus.Debugf("removing signatures while copying image")
```
rhatdan: Is this a placeholder?

cmd/buildah/manifest.go
nalind: `--remove-signatures` and `--sign-by` are what `skopeo copy` calls these options.  Are you sure?

common.go
nalind: It was, before I started writing tests.  I'll drop them.

vrothberg:
CI is complaining about `./common.go:12:2: imported and not used: "github.com/sirupsen/logrus"`

cmd/buildah/manifest.go
rhatdan: I don't mind the --signby, but --remove-signatures does not really describe what is happening, I think --nosigs would make more sense and easier to type.  --remove-signatures sounds like signatures are being deleted,  it is also way to many characters. I would rather pick something new and then move skopeo to use the new name and alias it to --remove-signatures.

rhatdan: I don't mind the --signby, but --remove-signatures does not really describe what is happening, I think --nosigs would make more sense and easier to type.  --remove-signatures sounds like signatures are being deleted,  it is also way to many characters. I would rather pick something new and then move skopeo to use the new name and alias it to --remove-signatures.

rhatdan:
@vrothberg @mheon @giuseppe @TomSweeneyRedHat Please comment on the --remove-signatures versus --nosigs or some alternative.

vrothberg:
> @vrothberg @mheon @giuseppe @TomSweeneyRedHat Please comment on the --remove-signatures versus --nosigs or some alternative.

I believe this ship has sailed as `skopeo copy` already has a `--remove-signatures`. I would prefer the tools to have a consistent interface. If we strongly against `--remove-signatures` we could deprecate it in Skopeo and create a new alias.

pkg/cli/common.go
```diff
@@ -168,8 +169,9 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
 	fs.BoolVar(&flags.Rm, "rm", true, "Remove intermediate containers after a successful build")
 	// "runtime" definition moved to avoid name collision in podman build.  Defined in cmd/buildah/bud.go.
 	fs.StringSliceVar(&flags.RuntimeFlags, "runtime-flag", []string{}, "add global flags for the container runtime")
+	fs.StringVar(&flags.SignBy, "sign-by", "", "sign the image using a GPG key with the specified `FINGERPRINT`")
 	fs.StringVar(&flags.SignaturePolicy, "signature-policy", "", "`pathname` of signature policy file (not usually used)")
-	fs.BoolVar(&flags.Squash, "squash", false, "Squash newly built layers into a single new layer.")
+	fs.BoolVar(&flags.Squash, "squash", false, "squash newly built layers into a single new layer")
```
TomSweeneyRedHat: oops, ty for cleaning up after me @nalind

TomSweeneyRedHat:
LGTM
I too don't really care for the remove-signature name, but as it's already been established in skopeo and I don't think it's too conceptually off, I think we should stick with it.

rhatdan:
@rh-atomic-bot r+
Reluctantly...

PR #2089, opened 2020-01-17
fix formatting for the latest changelog block

sanga:


TomSweeneyRedHat:
@sanga TYVM for the change.  The CI that we use requires that we sign any PR's which you neglected to do.  As that line really should be removed anyway, I've created #2092 to take care of that and I'll close this PR out.

PR #2095, opened 2020-01-18
Fix COPY in containerfile with envvar

TomSweeneyRedHat:
If a Containerfile had lines like:

```
FROM alpine
ENV VERSION=0.0.1
COPY file-${VERSION}.txt /
```

Buildah would not resolve the VERSION variable in the copy statement.
If the 'ENV' in the above Containerfile was changed to ARG, then this
would work.

A recent change to the handling of variables now only looks at variables
set by 'ARG' and not the ones set by the 'ENV' command.  This PR
adds the variables set by the `ENV` to the list of `ARG` variables
when those variables are being resolved by the code.

This also includes added tests to guard against this regression in the future.

Addresses:  https://github.com/containers/libpod/issues/4878

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -253,7 +253,7 @@ func (s *StageExecutor) volumeCacheRestore() error {
 // don't care about the details of where in the filesystem the content actually
 // goes, because we're not actually going to add it here, so this is less
 // involved than Copy().
-func (s *StageExecutor) digestSpecifiedContent(node *parser.Node, argValues []string) (string, error) {
+func (s *StageExecutor) digestSpecifiedContent(node *parser.Node, argValues []string, envValues []string) (string, error) {
 	// No instruction: done.
 	if node == nil {
 		return "", nil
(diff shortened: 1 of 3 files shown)


imagebuildah/stage_executor.go
```diff
@@ -253,7 +253,7 @@ func (s *StageExecutor) volumeCacheRestore() error {
 // don't care about the details of where in the filesystem the content actually
 // goes, because we're not actually going to add it here, so this is less
 // involved than Copy().
-func (s *StageExecutor) digestSpecifiedContent(node *parser.Node, argValues []string) (string, error) {
+func (s *StageExecutor) digestSpecifiedContent(node *parser.Node, argValues []string, envValues []string) (string, error) {
```
rhatdan: Would it be better to just pass into argvalues 
append(argValued, envValues ...)
Rather then add a new param?

TomSweeneyRedHat: I was on the fence and decided it would be better to have one append within the function.  That way future callers (most likely me) wouldn't forget to append both as the param would be in their face, and the append is only in once place this way.

rhatdan:
LGTM
@vrothberg @giuseppe @nalind PTAL

giuseppe:
LGTM

@rh-atomic-bot r+

PR #1231, opened 2018-12-10
TEST-PR: do not merge

vrothberg:
Signed-off-by: Valentin Rothberg <rothberg@redhat.com>

PR #2097, opened 2020-01-20
Allow users to set OS and architecture on bud

saschagrunert:
When building images we now allow the architecture and operating system
to be overwritten via the new `--os` and `--arch` flags. This allows the
use case of packing pre-built or cross-compiled binaries into container
images with specifying the correct environment. The `--platform` flag
now takes the `os/arch` format into account as well.

Inspired by: https://github.com/kubernetes/kubernetes/issues/87325
cc @vrothberg 

As a follow up we could bring that functionality to podman, too.

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -301,6 +301,11 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budOptions) error {
 
 	capabilities := defaultContainerConfig.Capabilities("", iopts.CapAdd, iopts.CapDrop)
 
+	os, arch, err := parse.PlatformFromOptions(c)
+	if err != nil {
+		return err
+	}
+
 	options := imagebuildah.BuildOptions{
 		ContextDirectory:        contextDir,
 		PullPolicy:              pullPolicy,
@@ -341,6 +346,8 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budOptions) error {
 		Devices:                 devices,
 		DefaultEnv:              defaultContainerConfig.GetDefaultEnv(),
 		SignBy:                  iopts.SignBy,
+		Architecture:            arch,
+		OS:                      os,
 	}
 
 	if iopts.Quiet {
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -42,6 +42,10 @@ Add an image *annotation* (e.g. annotation=*value*) to the image metadata. Can b
 
 Note: this information is not present in Docker image formats, so it is discarded when writing images in Docker formats.
 
+**--arch**="ARCH"
+
+Set the ARCH instead of the architecture of the current machine in the image manifest and config.
+
 **--authfile** *path*
 
 Path of the authentication file. Default is ${XDG\_RUNTIME\_DIR}/containers/auth.json, which is set using `buildah login`.
@@ -327,6 +331,10 @@ another process.
 
 Do not use existing cached images for the container build. Build from the start with a new set of cached layers.
 
+**--os**="OS"
+
+Set the OS instead of the current operating system of the machine in the image manifest and config.
+
 **--pid** *how*
 
 Sets the configuration for PID namespaces when handling `RUN` instructions.
@@ -336,11 +344,11 @@ that the PID namespace in which `buildah` itself is being run should be reused,
 or it can be the path to a PID namespace which is already in use by another
 process.
 
-**--platform**="Linux"
+**--platform**="OS/ARCH"
 
-This option has no effect on the build. Other container engines use this option
-to control the execution platform for the build (e.g., Windows, Linux) which is
-not required for Buildah as it supports only Linux.
+Set the OS/ARCH instead of the current operating system and architecture of the
+machine in the image manifest and config (for example `linux/arm`). If
+`--platform` is set, then the values of `--arch` and `--os` will be overwritten.
 
 **--pull**
 
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -1132,6 +1132,8 @@ func (s *StageExecutor) commit(ctx context.Context, ib *imagebuilder.Builder, cr
 	}
 	s.builder.SetHostname(config.Hostname)
 	s.builder.SetDomainname(config.Domainname)
+	s.builder.SetArchitecture(s.executor.architecture)
+	s.builder.SetOS(s.executor.os)
 	s.builder.SetUser(config.User)
 	s.builder.ClearPorts()
 	for p := range config.ExposedPorts {
--- a/pkg/cli/common.go
+++ b/pkg/cli/common.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/containers/buildah"
+	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/buildah/util"
 	"github.com/containers/common/pkg/config"
 	"github.com/opencontainers/runtime-spec/specs-go"
@@ -46,6 +47,7 @@ type NameSpaceResults struct {
 // BudResults represents the results for Bud flags
 type BudResults struct {
 	Annotation          []string
+	Arch                string
 	Authfile            string
 	BuildArg            []string
 	CacheFrom           string
@@ -61,6 +63,7 @@ type BudResults struct {
 	Logfile             string
 	Loglevel            int
 	NoCache             bool
+	OS                  string
 	Platform            string
 	Pull                bool
 	PullAlways          bool
@@ -145,6 +148,7 @@ func GetLayerFlags(flags *LayerResults) pflag.FlagSet {
 // GetBudFlags returns common bud flags
 func GetBudFlags(flags *BudResults) pflag.FlagSet {
 	fs := pflag.FlagSet{}
+	fs.StringVar(&flags.Arch, "arch", runtime.GOARCH, "set the ARCH instead of the architecture of the current machine in the image manifest and config")
 	fs.StringArrayVar(&flags.Annotation, "annotation", []string{}, "Set metadata for an image (default [])")
 	fs.StringVar(&flags.Authfile, "authfile", GetDefaultAuthFile(), "path of the authentication file.")
 	fs.StringArrayVar(&flags.BuildArg, "build-arg", []string{}, "`argument=value` to supply to the builder")
@@ -161,7 +165,8 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
 	fs.BoolVar(&flags.NoCache, "no-cache", false, "Do not use existing cached images for the container build. Build from the start with a new set of cached layers.")
 	fs.StringVar(&flags.Logfile, "logfile", "", "log to `file` instead of stdout/stderr")
 	fs.IntVar(&flags.Loglevel, "loglevel", 0, "adjust logging level (range from -2 to 3)")
-	fs.StringVar(&flags.Platform, "platform", "", "CLI compatibility: no action or effect")
+	fs.StringVar(&flags.OS, "os", runtime.GOOS, "set the OS instead of the current operating system of the machine in the image manifest and config")
+	fs.StringVar(&flags.Platform, "platform", parse.DefaultPlatform(), "set the OS/ARCH instead of the current operating system and architecture of the machine in the image manifest and config (for example `linux/arm`)")
 	fs.BoolVar(&flags.Pull, "pull", true, "pull the image from the registry if newer or not present in store, if false, only pull the image if not present")
 	fs.BoolVar(&flags.PullAlways, "pull-always", false, "pull the image even if the named image is present in store")
 	fs.BoolVar(&flags.PullNever, "pull-never", false, "do not pull the image, use the image present in store if available")
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"unicode"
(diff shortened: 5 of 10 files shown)


vrothberg:
FWIW `buildah config` supports setting both via `--arch` and `--os`, which somehow dwarfs my argument against support something similar to `--os` for `buildah bud`.

saschagrunert:
> Thanks for tackling that, @saschagrunert!
> 
> `--override-{arch,os}` are usually used to control image pulling, so I object to re-purposing the two options for setting certain fields in an image config.
> 
> `--override-arch` may be `--platform` which is currently a NOP but the correct option to remain compatible with Docker.

The thing is that we have `--arch` and `--os` in the `manifest add` subcommand as well. We could also add those two to the `bud` command and remove the `--override-*` commands for them (not sure about deprecation policy for hidden commands). So I see three options:

1. leave it as it currently is implemented
1. use `--platform` for `--override-arch` and still use `--override-os`
1. add `--arch`/`--os`

> In case of `--override-os`, I don't think we need to expose a way to change the OS. I don't know of any way to do cross-OS builds.

cri-tools cross compiles to windows as well for example, where: https://github.com/kubernetes-sigs/cri-tools/blob/47a7405313727fe8875328a46bfa53594a211e2c/hack/release.sh#L57-L66

vrothberg:
> > Thanks for tackling that, @saschagrunert!
> > `--override-{arch,os}` are usually used to control image pulling, so I object to re-purposing the two options for setting certain fields in an image config.
> > `--override-arch` may be `--platform` which is currently a NOP but the correct option to remain compatible with Docker.
> 
> The thing is that we have `--arch` and `--os` in the `manifest add` subcommand as well. We could also add those two to the `bud` command and remove the `--override-*` commands for them (not sure about deprecation policy for hidden commands). So I see three options:
> 
>     1. leave it as it currently is implemented
>     2. use `--platform` for `--override-arch` and still use `--override-os`

`--override-*` is used for pulling images. I think the `manifest` sub-commands are a good point but an inconsistency. The `--override-*` options could still be used to control which image is being pulled during `bud`.

>     3. add `--arch`/`--os` and remove `--override-*` flags

I think that's the way to go. It seems that `--platform` combines `--arch` and `--os` and eats input such as `linux/amd64`, so we can easily support that as well.

> > In case of `--override-os`, I don't think we need to expose a way to change the OS. I don't know of any way to do cross-OS builds.
> 
> cri-tools cross compiles to windows as well for example: https://github.com/kubernetes-sigs/cri-tools/blob/47a7405313727fe8875328a46bfa53594a211e2c/hack/release.sh#L57-L66

I meant cross-container builds not cross compilation. We don't support building/running images for any other OS than Linux, so I doubt the `--os` flag would be useful but as other commands expose it, it doesn't hurt to provide it for `bud` as well.

rhatdan:
I would prefer to stick with the short names.  We have a way of aliasing the longer names to the short that is used in libpod.
```
	flags.SetNormalizeFunc(aliasFlags)
func aliasFlags(f *pflag.FlagSet, name string) pflag.NormalizedName {
	switch name {
	case "healthcheck-command":
		name = "health-cmd"
	case "healthcheck-interval":
		name = "health-interval"
	case "healthcheck-retries":
		name = "health-retries"
	case "healthcheck-start-period":
		name = "health-start-period"
	case "healthcheck-timeout":
		name = "health-timeout"
	case "net":
		name = "network"
	}
	return pflag.NormalizedName(name)
}
```

rhatdan:
I really dislike multi-word options...

vrothberg:
@rhatdan, we are not targetting aliasing. The problem is that `--override-arch` is commonly used to control which image is being pulled. In the current state of the PR it would control which image is being built.

So we don't want alias but to iron out the ambiguity of `--override-arch` and `--arch` across the tools and (sub) commands. I want it to be as consistent as possible and believe that `--arch` and `--os` are the right names for setting the corresponding fields in the image config.

saschagrunert:
Ready for review. :)

docs/buildah-bud.md
```diff
@@ -42,6 +42,10 @@ Add an image *annotation* (e.g. annotation=*value*) to the image metadata. Can b
 
 Note: this information is not present in Docker image formats, so it is discarded when writing images in Docker formats.
 
+**--arch**="ARCH"
+
+Prefer ARCH instead of the architecture of the machine when building images.
```
vrothberg: * s/Prefer/Set/
* s/when building images/in the image manifest and config/

docs/buildah-bud.md
```diff
@@ -338,9 +346,7 @@ process.
 
 **--platform**="Linux"
```
vrothberg: I think the default is wrong as it's in the wrong format. I suggest to keep it empty.

docs/buildah-bud.md
```diff
@@ -338,9 +346,7 @@ process.
...
 
-This option has no effect on the build. Other container engines use this option
-to control the execution platform for the build (e.g., Windows, Linux) which is
-not required for Buildah as it supports only Linux.
+Specify the `os/architecture` to be used when building the image.
```
vrothberg: I suggest to copy what --os and --arch state above and give an example value (e.g., "linux/amd64").

I guess we should mention for all three options which values are valid (i.e., the golang runtime values).

pkg/cli/common.go
```diff
@@ -161,7 +164,7 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
 	fs.BoolVar(&flags.NoCache, "no-cache", false, "Do not use existing cached images for the container build. Build from the start with a new set of cached layers.")
 	fs.StringVar(&flags.Logfile, "logfile", "", "log to `file` instead of stdout/stderr")
 	fs.IntVar(&flags.Loglevel, "loglevel", 0, "adjust logging level (range from -2 to 3)")
-	fs.StringVar(&flags.Platform, "platform", "", "CLI compatibility: no action or effect")
+	fs.StringVar(&flags.Platform, "platform", parse.DefaultPlatform(), "Specify the `os/architecture` to be used when building the image")
```
vrothberg: s/Specify/specify/

pkg/parse/parse.go
```diff
@@ -589,11 +590,30 @@ func SystemContextFromOptions(c *cobra.Command) (*types.SystemContext, error) {
...
-	if arch, err := c.Flags().GetString("override-arch"); err == nil {
+
+	if pf, err := c.Flags().GetString("platform"); err == nil && pf != DefaultPlatform() {
+		os, arch, err := parsePlatform(pf)
+		if err != nil {
+			return nil, errors.Wrapf(err, "unable to parse platform %s", pf)
+		}
 		ctx.ArchitectureChoice = arch
+		ctx.OSChoice = os
 	}
```
saschagrunert: This part is a bit tricky because we have some sort of implicit priority, but I think it should not harm the overall behavior.

pkg/parse/parse.go
```diff
@@ -589,11 +590,30 @@ func SystemContextFromOptions(c *cobra.Command) (*types.SystemContext, error) {
...
-	if arch, err := c.Flags().GetString("override-arch"); err == nil {
+
+	if pf, err := c.Flags().GetString("platform"); err == nil && pf != DefaultPlatform() {
+		os, arch, err := parsePlatform(pf)
+		if err != nil {
+			return nil, errors.Wrapf(err, "unable to parse platform %s", pf)
+		}
 		ctx.ArchitectureChoice = arch
+		ctx.OSChoice = os
```
vrothberg: We shouldn't use `(*types.SystemContext).{Architecture,OS}Choice` for _creating_ an image. Those fields should only be used for _pulling_ an image (see https://github.com/containers/image/blob/master/types/types.go#L510).

imagebuildah/stage_executor.go
```diff
@@ -1131,6 +1131,8 @@ func (s *StageExecutor) commit(ctx context.Context, ib *imagebuilder.Builder, cr
 	}
 	s.builder.SetHostname(config.Hostname)
 	s.builder.SetDomainname(config.Domainname)
+	s.builder.SetArchitecture(s.executor.systemContext.ArchitectureChoice)
+	s.builder.SetOS(s.executor.systemContext.OSChoice)
```
vrothberg: We shouldn't use `(*types.SystemContext).{Architecture,OS}Choice` for _creating_ an image. Those fields should only be used for _pulling_ an image (see https://github.com/containers/image/blob/master/types/types.go#L510).

pkg/parse/parse.go
vrothberg: We need new fields for this in the executer (see the other comments).

pkg/parse/parse.go
vrothberg: This illustrates what I was referring to before. `--override-{arch,os}` is used for pulling. When creating an image, we should use different options to prevent mixing things or blurring the lines.

docs/buildah-bud.md
```diff
@@ -327,6 +331,10 @@ another process.
 
 Do not use existing cached images for the container build. Build from the start with a new set of cached layers.
 
+**--os**="OS"
+
+Prefer OS instead of the running OS when building images.
```
rhatdan: Should this be `Set OS ...` as well.

rhatdan: Should we default to the current OS, so that the `buildah bud --help` shows the default?

rhatdan:
Command completions are missing.

pkg/parse/parse.go
saschagrunert: Hm, we still have to set it in the systemcontext, otherwise we would get the error on commit:
```
> ./buildah bud --pull-never --arch=arm -f tests/bud/platform/Dockerfile -t test
STEP 1: FROM scratch
STEP 2: COMMIT test
error committing base container: error copying layers and metadata for container "23ffa04e08322148180e7e45a39804ee9c2663758e0cfcd3a9bccaa2ca4f0333": Image operating system mismatch: image uses "arm", expecting "linux"
ERRO exit status 1
```

This issue is caused by the system context verification here:
https://github.com/containers/image/blob/b0893d83e0e450317ef781e9d6a621182d1a6346/copy/copy.go#L715-L721

docs/buildah-bud.md
saschagrunert: Fixed :+1:

docs/buildah-bud.md
saschagrunert: > I suggest to copy what --os and --arch state above and give an example value (e.g., "linux/amd64").

Fixed that.

> I guess we should mention for all three options which values are valid (i.e., the golang runtime values).

Hm, I'm not sure since there might be incompatible os/arch combinations avilable. I think we don't have clear way to validate os and arch combinations for each go version right?

pkg/cli/common.go
saschagrunert: Fixed :+1:

pkg/parse/parse.go
saschagrunert: Yep, added the values to the executor.

docs/buildah-bud.md
saschagrunert: Fixed that, the help shows the system default values:
```
 > ./buildah bud -h | rg '\--(arch|os|platform)'
      --arch string                   set the ARCH instead of the architecture of the current machine in the image manifest and config (default "amd64")
      --os string                     set the OS instead of the current operating system of the machine in the image manifest and config (default "linux")
      --platform linux/arm            set the OS/ARCH instead of the current operating system and architecture of the machine in the image manifest and config (for example linux/arm) (default "linux/amd64")
```

saschagrunert:
> Command completions are missing.

Fixed that as well :+1:

docs/buildah-bud.md
vrothberg: I honestly have no idea about combinatorial issues but had the individual values for a valid arch and os in mind. The image spec, for instance, [points to the golang docs](https://github.com/opencontainers/image-spec/blob/243ea084a44451d27322fed02b682d99e2af3ba9/image-index.md#image-index-property-descriptions).

saschagrunert:
Two open discussion points left, one related to the system context seems critical because its currently breaking the behavior.

docs/buildah-bud.md
```diff
@@ -336,11 +344,9 @@ that the PID namespace in which `buildah` itself is being run should be reused,
 or it can be the path to a PID namespace which is already in use by another
 process.
 
-**--platform**="Linux"
+**--platform**="OS/ARCH"
 
-This option has no effect on the build. Other container engines use this option
-to control the execution platform for the build (e.g., Windows, Linux) which is
-not required for Buildah as it supports only Linux.
+Set the OS/ARCH instead of the current operating system and architecture of the machine in the image manifest and config (for example `linux/arm`).
```
vrothberg: I guess we should mention that --platform overrides --os/arch.

pkg/parse/parse.go
```diff
@@ -606,6 +607,44 @@ func getAuthFile(authfile string) string {
 	return os.Getenv("REGISTRY_AUTH_FILE")
 }
 
+func PlatformFromOptions(c *cobra.Command) (os, arch string, err error) {
```
vrothberg: Missing comment.

pkg/parse/parse.go
```diff
@@ -606,6 +607,44 @@ func getAuthFile(authfile string) string {
...
+		arch = selectedArch
+		os = selectedOS
+	}
+
+	return os, arch, nil
+}
+
+const platformSep = "/"
+
+// DefaultPlatform returns the standard platform name for the current system
```
vrothberg: s/name//

pkg/parse/parse.go
```diff
@@ -606,6 +607,44 @@ func getAuthFile(authfile string) string {
...
+func DefaultPlatform() string {
+	return runtime.GOOS + platformSep + runtime.GOARCH
+}
+
+func parsePlatform(platform string) (os, arch string, err error) {
+	split := strings.Split(platform, platformSep)
+	if len(split) != 2 {
+		return "", "", errors.New("invalid platform syntax, use OS/ARCH")
```
vrothberg: I think it's worth quoting `platform` in the error.

pkg/parse/parse.go
vrothberg: ` image uses "arm", expecting "linux"` is quite wrong in many ways.

I assume that we need to set the values when committing the container, which internally uses image/copy.

docs/buildah-bud.md
saschagrunert: Sure, mentioned that as well :)

cmd/buildah/bud.go
```diff
@@ -301,6 +301,11 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budOptions) error {
 
 	capabilities := defaultContainerConfig.Capabilities("", iopts.CapAdd, iopts.CapDrop)
 
+	arch, os, err := parse.PlatformFromOptions(c)
```
vrothberg: These values must be flipped. It's `os, arch`.

pkg/parse/parse.go
saschagrunert: Added a doc string. :+1:

cmd/buildah/bud.go
saschagrunert: Hah, that's a great catch :) thanks

pkg/parse/parse.go
saschagrunert: Yes, the first issue is caused by the os/arch flip error.

rhatdan:
@rh-atomic-bot r+

saschagrunert:
Happy to bring this into podman as well after a new buildah release.

Prashanth684:
> Happy to bring this into podman as well after a new buildah release.

any idea when this will be ? thank you!

TomSweeneyRedHat:
Short Answer: @Prashanth684 soon.

Longer Answer: @rhatdan is working on getting Podman to deal with the new containers.conf file (https://github.com/containers/libpod/pull/4698).  The work necessary for Buildah to do that just wrapped up (#1231) so I suspect Podman will shortly too.  Once that's all complete we'll create  new release for Buildah and will vendor that into Podman.  I'm hoping later this week, early next, but no promises.

Prashanth684:
thanks @TomSweeneyRedHat . No worries. Just wanted to know.

PR #2103, opened 2020-01-22
manifests: unit test fix

nalind:
The unit tests need to use the policy and registries configurations from the source tree, so point the SystemContext that we use in our tests to those files.


TomSweeneyRedHat:
LGTM

rhatdan:
LGTM
@rh-atomic-bot r+

PR #2105, opened 2020-01-22
chroot: handle slightly broken seccomp defaults

nalind:
When a seccomp rule includes multiple equality checks for the same argument for a syscall, they can never ALL be satisfied.  Because that's how they're supposed to be treated, libseccomp returns an error when we try to add them as part of the same conditional rule.  Try to detect this exact case, and if we detect it, treat each condition as its own rule.

### DIFF
--- a/chroot/seccomp.go
+++ b/chroot/seccomp.go
@@ -3,7 +3,7 @@
 package chroot
 
 import (
-	"github.com/opencontainers/runtime-spec/specs-go"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 	libseccomp "github.com/seccomp/libseccomp-golang"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


chroot/seccomp.go
```diff
@@ -118,15 +118,32 @@ func setSeccomp(spec *specs.Spec) error {
...
 				conditions = append(conditions, condition)
 			}
 			if err = filter.AddRuleConditional(scnum, mapAction(rule.Action), conditions); err != nil {
-				return errors.Wrapf(err, "error adding a conditional rule (%q:%q) to seccomp filter", scnames[scnum], rule.Action)
+				// Okay, if the rules specify multiple equality
+				// checks, assume someone thought that they
+				// were OR'd, when in fact they're ordinarily
+				// supposed to be AND'd.  Break them up into
+				// different rules to get that OR effect.
+				if len(rule.Args) > 1 && opsAreAllEquality && err.Error() == "two checks on same syscall argument" {
```
TomSweeneyRedHat: Where's the string from?  It always worries me to check for an error by looking at the string output rather than a particular error type.

nalind: I share your concern.  This is an error object that `github.com/seccomp/libseccomp-golang` creates every time it returns the error, so we can't just compare the error itself for equality with a known value.  If there's a better method, I'm happy to switch to it.

rhatdan:
@rh-atomic-bot r+

vrothberg:
> I have a hard time reviewing this. Is there any chance we can generalize the problem and get it into `libseccomp-golang`?

Why was it merged without a reply to my concerns above?

rhatdan:
I saw this as a separate PR, not necessarily related to Buildah.  I am also wondering if the chroot code is specific to Buildah, and not sure if this is useful to other projects.

nalind:
FWIW, we're doing this here instead of the seccomp bindings to mirror part of how runc does things in https://github.com/opencontainers/runc/pull/1616.

PR #2106, opened 2020-01-22
parse: don't complain about not being able to rename something to itself

nalind:
In DeviceFromPath(), don't complain about not being able to rename a device when the previous and new names are the same.


TomSweeneyRedHat:
LGTM

rhatdan:
LGTM
@rh-atomic-bot r+

PR #2111, opened 2020-01-25
copy.bats: make sure we detect failures due to missing source

nalind:
Make sure we fail a build when the source for a COPY --from instruction can't be found.


rhatdan:
Always like new tests.
@rh-atomic-bot r+

Issue #2102, opened 2020-01-22
ADD to symlink tar file messes up destination directory


sophonet:
**Description**

If a Dockerfile contains a line ADD tarball.tar.gz DSTFOLDER and tarball.tar.gz is a symlink (e.g. tarball_latest.tar.gz), DSTFOLDER incorrectly gets augmented with tarball.tar.gz.

**Steps to reproduce the issue:**
```
mkdir tarroot
touch tarroot/empty
tar cvfz tarball.tar.gz tarroot --owner=0 --group=0
ln -s tarball.tar.gz tarball_latest.tar.gz

Dockerfile:
FROM centos:7
WORKDIR /
ADD tarball_latest.tar.gz

podman build -t tartst .

podman run -ti --rm tartst /bin/bash
ls -lR /ta*
```
**Describe the results you received:**

Root in container lists (among other entries)

```
/tarball_latest.tar.gz/tarroot/empty
```

**Describe the results you expected:**

```
/tarroot/empty
```

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah/unknown,now 1.12.0-3 amd64
```

**Output of `buildah version`:**

```
Version:         1.12.0
Go Version:      go1.10.1
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.1-dev
CNI Spec:        0.4.0
libcni Version:
image Version:   5.0.0
Git Commit:
Built:           Thu Jan 16 15:29:51 2020
OS/Arch:         linux/amd64
```

**Output of `podman version` if reporting a `podman build` issue:**

```
Version:            1.7.0
RemoteAPI Version:  1
Go Version:         go1.10.1
OS/Arch:            linux/amd64
```

**Output of `cat /etc/*release`:**

```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.3 LTS"
NAME="Ubuntu"
VERSION="18.04.3 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 18.04.3 LTS"
VERSION_ID="18.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
[...]

PR #2112, opened 2020-01-26
Adjust copy destination for linked files in ADD

TomSweeneyRedHat:
When we are given a linked tar file as a source file, we were setting the
destination to the original destination plus the name of the link.
This created a directory using the links name and then everything from the tar file was
copied under that directory.  The link name should not be used as part
of the directory structure.

Fixes: #2102


TomSweeneyRedHat:
Hmm, looks like I broke a different test with this change, will run it down early next week.

TomSweeneyRedHat:
@nalind @rhatdan @vrothberg PTAL

rhatdan:
@rh-atomic-bot r+

PR #2117, opened 2020-01-28
buildahimage: specify fuse-overlayfs mount options

giuseppe:
tweak the storage.conf file to use fuse-overlayfs specific mount
options.  In particular, disable any sync operation on the files on
fuse-overlayfs.

sync is an expensive operation and there is no gain in sync'ing files
when doing a build as if the build fails we don't store the
intermediate result.

### DIFF
--- a/contrib/buildahimage/stable/Dockerfile
+++ b/contrib/buildahimage/stable/Dockerfile
@@ -14,7 +14,7 @@ FROM fedora:latest
 RUN useradd build; yum -y update; yum -y reinstall shadow-utils; yum -y install buildah fuse-overlayfs --exclude container-selinux; rm -rf /var/cache /var/log/dnf* /var/log/yum.*
 
 # Adjust storage.conf to enable Fuse storage.
-RUN sed -i -e 's|^#mount_program|mount_program|g' -e '/additionalimage.*/a "/var/lib/shared",' /etc/containers/storage.conf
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM
@TomSweeneyRedHat @giuseppe @mtrmac and I talked about this at the DNF speedup meeting.  As a possible way to increase the build speed of buildah.  

@nalind @TomSweeneyRedHat @mheon @vrothberg PTAL

TomSweeneyRedHat:
The change LGTM, but I want to question if these changes will hinder other container operations other than build?  We've a number of folks using these images in their CI operations and want to make sure we don't get better performance in the build, but slow down the run.

rhatdan:
@rh-atomic-bot r+
@TomSweeneyRedHat these changes just fix and issue where when rpm and APT put down every file they call fsync, which is a rather heavy duty operation, in order to make sure the file gets placed on disk.  That way if the disk crashes, then the files have been written.  In Buildah if anything goes wrong with the disk/system crashing, everything gets thrown away.  So we think this is safe to do.  

This only effects the fuse-overlay file system, if a RUN was using a volume, then the content will get written to disk.

TomSweeneyRedHat:
thx for the background info @rhatdan.  Should the same changes happen in the podmanimages too?

rhatdan:
I am not sure about that.  Since these would be more permanent.

PR #2118, opened 2020-01-28
getDateAndDigestAndSize(): use manifest.Digest

nalind:
Use the image library's function for digesting manifests, so that we'll compute and display the right values for docker_schema1 manifests.


rhatdan:
LGTM
@mtrmac @TomSweeneyRedHat @vrothberg PTAL

rhatdan:
@rh-atomic-bot r+

PR #2120, opened 2020-01-29
sign.bats: set GPG_TTY=/dev/null

nalind:
Set GPG_TTY=/dev/null, which should hopefully prevent ENOTTY errors during gating tests in CI.  h/t  @cevich


cevich:
Suggestion: Add this to `helpers.bash` so it's set for all tests and forever.  I can't think of a reason why that should never ***not*** be the case while running BATS.

Otherwise LGTM.

nalind:
Sold!  Moving it.

rhatdan:
@rh-atomic-bot r+

PR #2115, opened 2020-01-27
Kill Travis and Enable Bors

cevich:
***Depends on:***  #1848 #1971 #2036 #2121

PR #2126, opened 2020-01-31
info test: deal with random key order

edsantiago:
Looks like the order of keys in the .store map is not
guaranteed; it's causing CI failures in #2115. Keep
checking the results, but be more flexible about order.

### DIFF
--- a/tests/info.bats
+++ b/tests/info.bats
@@ -7,5 +7,9 @@ load helpers
   expect_output --substring "host"
 
   run_buildah info --format='{{.store}}'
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

rhatdan:
@cevich @TomSweeneyRedHat @nalind @QiWang19 @mheon @vrothberg PTAL

vrothberg:
@rh-atomic-bot r+

cevich:
Thanks Ed.

PR #2131, opened 2020-02-02
Repair buildah so it can use containers.conf on the server side

rhatdan:
In porting containers.conf to libpod, we found that buildah needed
to handle the containers.conf on the server side rather then from
the CLI.

Since the `podman-remote build` would probably not have the same content
as containers.conf on the server, the processesing of the defaults needs
to be handled in imagebuildah.  The CapAdd and CapDrop values need to be
passed in.

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -10,7 +10,6 @@ import (
 	buildahcli "github.com/containers/buildah/pkg/cli"
 	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/buildah/util"
-	"github.com/opencontainers/runc/libcontainer/configs"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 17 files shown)


rhatdan:
This does revert an API change that we made when we merged the previous containers.conf.
I will try this out against Podman to make sure if fixes the issues I am seeing in testing there.
Wait to merge until I get that fixed.

rhatdan:
@giuseppe @vrothberg @nalind @QiWang19 @mheon PTAL
I need to get this merged so we can get containers.conf into Podman.

vrothberg:
@rhatdan do you have that vendored into the libpod PR? Just to be sure that will make libpod CI green.

rhatdan:
My containers.conf PR has a vendoring of this.
bors r+

cevich:
(I made bors fail + retry, travis was being dumb again)

cevich:
If the auto-retry fails, I would recommend rebasing against the latest master.

umohnani8:
LGTM

PR #2140, opened 2020-02-04
manifest add: always read the individual image manifest

nalind:
Always read the manifest of the individual image that we're adding to a list, so that we can detect its MIME type correctly.  Previously, if the image we were adding to the manifest was an entry that we were picking out of a list, we would have added the type of the list.

### DIFF
--- a/manifests/manifests.go
+++ b/manifests/manifests.go
@@ -252,21 +252,17 @@ func (l *list) Add(ctx context.Context, sys *types.SystemContext, ref types.Imag
 	var instanceInfos []instanceInfo
 	var manifestDigest digest.Digest
 
-	manifestBytes, manifestType, err := src.GetManifest(ctx, nil)
+	primaryManifestBytes, primaryManifestType, err := src.GetManifest(ctx, nil)
 	if err != nil {
 		return "", errors.Wrapf(err, "error reading manifest from %q", transports.ImageName(ref))
 	}
 
-	if manifest.MIMETypeIsMultiImage(manifestType) {
-		list, err := manifest.ListFromBlob(manifestBytes, manifestType)
+	if manifest.MIMETypeIsMultiImage(primaryManifestType) {
+		lists, err := manifests.FromBlob(primaryManifestBytes)
 		if err != nil {
 			return "", errors.Wrapf(err, "error parsing manifest list in %q", transports.ImageName(ref))
 		}
 		if all {
-			lists, err := manifests.FromBlob(manifestBytes)
-			if err != nil {
-				return "", errors.Wrapf(err, "error parsing manifest list in %q", transports.ImageName(ref))
-			}
 			for i, instance := range lists.OCIv1().Manifests {
 				platform := instance.Platform
 				if platform == nil {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@mtrmac @vrothberg @giuseppe @TomSweeneyRedHat @QiWang19 PTAL

manifests/manifests.go
```diff
@@ -347,11 +347,11 @@ func (l *list) Add(ctx context.Context, sys *types.SystemContext, ref types.Imag
 				instanceInfo.Architecture = config.Architecture
 			}
 		}
+		manifestBytes, manifestType, err = src.GetManifest(ctx, instanceInfo.instanceDigest)
```
mtrmac: Could this be `:=` to make it clearer that the variable only needs to exist for the duration of the single loop iteration?

Or perhaps rename the top-level values to `[manifest]ListBytes`/`[manifest]ListType` to make it very clear that the two have nothing in common.

nalind: Renamed the top-level values.  Moved initialization of `lists` up one scope since we were doing it in both branches of an `if/then`, and moved initialization of `list` into the only scope where it's used.

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
@rh-atomic-bot r+

nalind:
@rh-atomic-bot retry

rhatdan:
bors merge

rhatdan:
@rh-atomic-bot retry

rhatdan:
@jlebon Could you cycle humo, it seems to be hung up, probably from our mixing of bors with it.  Hopefully we can soon remove it entirely, once bors works correctly.

jlebon:
@rhatdan Hmm, seems like the test just failed, no? https://github.com/containers/buildah/pull/2140#issuecomment-582625237

```
>>> podman run --net=host --security-opt label=disable --cap-add all --security-opt seccomp=unconfined -v /etc/yum.repos.d:/etc/yum.repos.d.host:ro -v $PWD:/go/src/github.com/containers/buildah --workdir /go/src/github.com/containers/buildah registry.fedoraproject.org/fedora:30 bash -c sh ./.papr.sh
Trying to pull registry.fedoraproject.org/fedora:30...time="2020-02-05T21:38:35Z" level=error msg="Error pulling image ref //registry.fedoraproject.org/fedora:30: Error determining manifest MIME type for docker://registry.fedoraproject.org/fedora:30: Error reading manifest 30 in registry.fedoraproject.org/fedora: received unexpected HTTP status: 503 Service Temporarily Unavailable" 
Failed
Error: unable to pull registry.fedoraproject.org/fedora:30: unable to pull image: Error determining manifest MIME type for docker://registry.fedoraproject.org/fedora:30: Error reading manifest 30 in registry.fedoraproject.org/fedora: received unexpected HTTP status: 503 Service Temporarily Unavailable
### EXITED WITH CODE 125 AFTER 2s
```

Looks like a flake (our CI has also been hitting flake issues with the Fedora registry recently). You can probably just retry it again.

rhatdan:
@rh-atomic-bot retry

rhatdan:
@rh-atomic-bot retry

rhatdan:
@rh-atomic-bot retry

PR #2141, opened 2020-02-04
manifest push: add --format option

nalind:
Add a --format option, which takes the same values as the --format option for `buildah push`, to `buildah manifest push`, for cases where we can't detect the right type to use.

### DIFF
--- a/cmd/buildah/manifest.go
+++ b/cmd/buildah/manifest.go
@@ -17,6 +17,7 @@ import (
 	"github.com/containers/image/v5/transports"
 	"github.com/containers/image/v5/transports/alltransports"
 	digest "github.com/opencontainers/go-digest"
+	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM
@vrothberg @TomSweeneyRedHat @QiWang19 @mtrmac PTAL

cmd/buildah/manifest.go
```diff
@@ -639,13 +641,26 @@ func manifestPushCmd(c *cobra.Command, args []string, opts manifestPushOpts) err
 		return err
 	}
 
+	var manifestType string
+	if opts.format != "" {
+		switch opts.format {
+		case "oci":
+			manifestType = imgspecv1.MediaTypeImageIndex
+		case "v2s2", "v2s1", "docker":
```
mtrmac: Accepting undocumented `docker`, fine. But why `v2s1`? To me that seems likely to mean the user doesn’t understand the values and we should reject instead of silently doing something else.

nalind: That's mainly because we accept it as a value for `buildah push --format`.  I could go either way on whether or not we reject that value here, but the only documented valid values should be v2s2 and oci.

mtrmac: Sure, that symmetry makes sense for `docker` and the others. Accepting the `v2s1` value and then silently doing `v2s2` is not quite the same thing — up to you, of course.

nalind: Sure.  Dropped it as a recognized value.

TomSweeneyRedHat:
LGTM once the tests are happy.  The one failing one looked to be a network hiccup, I'll give it a restart.

PR #2142, opened 2020-02-05
WIP: Automatic PR Labeling + Merge blocking

cevich:
***Depends on:*** #2115


cevich:
Lovely...there's no way to test adding a new action b/c github only runs already committed actions :frowning_face: 

I'll need to make a temporary testing branch to verify this PR's changes actually function...

cevich:
Closing as this cannot work with Github Actions due to the design of it's permissions system.  The only possibility is to use a periodic/schedule-based bot.  Ref discussion here: https://github.com/actions/labeler/issues/12#issuecomment-581149070

PR #2129, opened 2020-01-31
Bors-ng: Add documentation and status-icon

cevich:
***Note:*** Depends on #2115

Signed-off-by: Chris Evich <cevich@redhat.com>

PR #2145, opened 2020-02-05
fix XDG_RUNTIME_DIR for authfile

QiWang19:
fix bug https://bugzilla.redhat.com/show_bug.cgi?id=1798343
log/logout fails after running `unset XDG_RUNTIME_DIR`. This patch calls getStore() to set XDG_RUNTIME_DIR in the code.

### DIFF
--- a/cmd/buildah/common.go
+++ b/cmd/buildah/common.go
@@ -37,14 +37,9 @@ func getStore(c *cobra.Command) (storage.Store, error) {
 		options.GraphRoot = globalFlagResults.Root
 		options.RunRoot = globalFlagResults.RunRoot
 	}
-	if unshare.IsRootless() && os.Getenv("XDG_RUNTIME_DIR") == "" {
-		runtimeDir, err := storage.GetRootlessRuntimeDir(unshare.GetRootlessUID())
-		if err != nil {
-			return nil, err
-		}
-		if err := os.Setenv("XDG_RUNTIME_DIR", runtimeDir); err != nil {
-			return nil, errors.New("could not set XDG_RUNTIME_DIR")
-		}
+
+	if err := setXDGRuntimeDir(); err != nil {
+		return nil, err
 	}
 
 	if c.Flag("storage-driver").Changed {
@@ -119,6 +114,20 @@ func getStore(c *cobra.Command) (storage.Store, error) {
 	}
 	needToShutdownStore = true
 	return store, err
+}
+
+// setXDGRuntimeDir sets XDG_RUNTIME_DIR when if it is unset under rootless
+func setXDGRuntimeDir() error {
+	if unshare.IsRootless() && os.Getenv("XDG_RUNTIME_DIR") == "" {
+		runtimeDir, err := storage.GetRootlessRuntimeDir(unshare.GetRootlessUID())
+		if err != nil {
+			return err
+		}
+		if err := os.Setenv("XDG_RUNTIME_DIR", runtimeDir); err != nil {
+			return errors.New("could not set XDG_RUNTIME_DIR")
+		}
+	}
+	return nil
 }
 
 func openBuilder(ctx context.Context, store storage.Store, name string) (builder *buildah.Builder, err error) {
--- a/cmd/buildah/login.go
+++ b/cmd/buildah/login.go
@@ -61,6 +61,11 @@ func loginCmd(c *cobra.Command, args []string, iopts *loginReply) error {
 	if len(args) == 0 {
 		return errors.Errorf("please specify a registry to login to")
 	}
+
+	if err := setXDGRuntimeDir(); err != nil {
+		return err
+	}
+
 	server := parse.RegistryFromFullName(parse.ScrubServer(args[0]))
 	systemContext, err := parse.SystemContextFromOptions(c)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


QiWang19:

cmd/buildah/logout.go
```diff
@@ -45,6 +45,12 @@ func logoutCmd(c *cobra.Command, args []string, iopts *logoutReply) error {
 	if len(args) == 0 && !iopts.all {
 		return errors.Errorf("registry must be given")
 	}
+
+	_, err := getStore(c)
```
rhatdan: Can you do this on one line please.
if _, err:=getstore(); err != nil

cmd/buildah/login.go
```diff
@@ -61,6 +61,12 @@ func loginCmd(c *cobra.Command, args []string, iopts *loginReply) error {
 	if len(args) == 0 {
 		return errors.Errorf("please specify a registry to login to")
 	}
+
+	_, err := getStore(c)
```
rhatdan: Do this on one line.

cmd/buildah/logout.go
QiWang19: fixed

cmd/buildah/login.go
QiWang19: fixed

rhatdan:
Could you add a test to make sure this works and does not regress?

TomSweeneyRedHat:
Definitely please add a test for the regression.
I've a couple of concerns.  1).  If the user meant to disable that envvar, we turn it on for them again without them knowing.  I think we ought to either just print out a message like:

"The XDG_RUNTIME_DIR environment is undefined and must be set for buildah login to complete"

or before setting it, write out a message like:

"The XDG_RUNTIME_DIR environment variables is undefined, setting it to {x}"

to let the user know it's being set.

2) The code in getStore() does a lot of stuff.  Can we pull out the code that does the setting of XDG_RUNTIME_DIR into a small function that could be called from both places?

rhatdan:
> Definitely please add a test for the regression.
> I've a couple of concerns. 1). If the user meant to disable that envvar, we turn it on for them again without them knowing. I think we ought to either just print out a message like:
> 
No this only turns it on for the podman command, it does not effect any other session of the user.  It is pretty much hidden from him, and just allows container/image to work correctly.
> "The XDG_RUNTIME_DIR environment is undefined and must be set for buildah login to complete"
> 
> or before setting it, write out a message like:
> 
> "The XDG_RUNTIME_DIR environment variables is undefined, setting it to {x}"
> 
> to let the user know it's being set.
> 
I don't think this is needed, since the user will never see a change.
>     1. The code in getStore() does a lot of stuff.  Can we pull out the code that does the setting of

This sounds like a good idea.

TomSweeneyRedHat:
@rhatdan FWIW, we do a os.SetEnv() for the XDG_RUNTIME_DIR, so it would set it for the session, not just internally within Podman.  It will go away when they log out.  If you think it's good without a warning, I'm fine with that.

rhatdan:
No,  Calling setenv inside of a program only changes the process and its children not for its parents.

QiWang19:
bot, retest this please

cmd/buildah/login.go
```diff
@@ -61,6 +63,17 @@ func loginCmd(c *cobra.Command, args []string, iopts *loginReply) error {
 	if len(args) == 0 {
 		return errors.Errorf("please specify a registry to login to")
 	}
+
+	if unshare.IsRootless() && os.Getenv("XDG_RUNTIME_DIR") == "" {
```
rhatdan: Make this a function that getstore calls, and we can call directly.

cmd/buildah/common.go
```diff
@@ -121,6 +116,20 @@ func getStore(c *cobra.Command) (storage.Store, error) {
 	return store, err
 }
 
+// setXDG sets XDG_RUNTIME_DIR when if it is unset under rootless
+func setXDG() error {
```
TomSweeneyRedHat: There are other XDG variables, I'd prefer this to be named setXDGRuntimeDir() just to be clear which one we're setting.  But I won't lose sleep over it.  @rhatdan wdyt?

Otherwise LGTM.

rhatdan: Much better, change the name of the function and then we can merge.

rhatdan:
LGTM
bors r+

TomSweeneyRedHat:
LGTM all kinds of test unhappiness though.  Is bors kicking up its heals?

QiWang19:
Do I need bot retest?

TomSweeneyRedHat:
@cevich any thoughts on the next step to make bors happy here?

QiWang19:
bot, retest this please

cevich:
> any thoughts on the next step to make bors happy here?

Yes patients:  Wait for #2115 to merge first (it's about to) then the automatic retry should work fine.

cevich:
If it fails again, I would recommend rebasing against the latest master.

cevich:
@QiWang19 BTW, there is no more "bot", it's now called "bors".  I added documentation for it https://github.com/containers/buildah/pull/2129 should merge soon.

PR #2149, opened 2020-02-07
Show error on copied file above context directory in build

TomSweeneyRedHat:
When a COPY command in a Container file looked like "../file.txt"
SecureJoin would secure the file by lopping off the "../".
However, the code would then append that file name to the passed in
context directory and look for the file.  That would fail as in most
cases there was no `{context-dir}/file.txt`, rather the file was at
`{context-dir}/../file.txt`.  Using a relative directory like this
outside of the context directory can be a security risk.  Docker
doesn't allow it nor should we.

This change now errors out when a file that starts with `../` is
presented as a copy from target.

Addresses: https://github.com/containers/libpod/issues/5035

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -316,8 +316,12 @@ func (s *StageExecutor) digestSpecifiedContent(node *parser.Node, argValues []st
 		} else {
 			// Source is not a URL, so it's a location relative to
 			// the all-content-comes-from-below-this-directory
-			// directory.
+			// directory.  Also raise an error if the src escapes
+			// the context directory.
 			contextSrc, err := securejoin.SecureJoin(contextDir, src)
+			if err == nil && strings.HasPrefix(src, "../") {
+				err = errors.New("escaping context directory error")
+			}
 			if err != nil {
 				return "", errors.Wrapf(err, "forbidden path for %q, it is outside of the build context %q", src, contextDir)
 			}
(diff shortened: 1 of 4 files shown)


vrothberg:
`[+0056s] ** ERROR: dangling whitespace found in these files: **`

Didn't try locally but `**` isn't very informative.

imagebuildah/stage_executor.go
```diff
@@ -316,8 +316,12 @@ func (s *StageExecutor) digestSpecifiedContent(node *parser.Node, argValues []st
 		} else {
 			// Source is not a URL, so it's a location relative to
 			// the all-content-comes-from-below-this-directory
-			// directory.
+			// directory.  Also raise an error if the src escapes
+			// the context directory.
 			contextSrc, err := securejoin.SecureJoin(contextDir, src)
+			if strings.HasPrefix(src, "../") && err == nil {
```
rhatdan: nit:  Should reverse order and check if err == nil first, for speed, although maybe compiler is smart enough.

rhatdan:
A couple of small nits, but CI system seems to be hosed up.

imagebuildah/stage_executor.go
TomSweeneyRedHat: touched up here and line 445

tests/bud/context-escape-dir/upperfile.txt
```diff
@@ -0,0 +1,3 @@
+This is a text file to be used in Buildah testing.
+This will be used to ensure that a file from above the context 
+directory can not be copied during the buildh phase.
```
TomSweeneyRedHat: apparently a rogue space at the end of this line caused tests to be unhappy.  Also touched up the "buildh" typo

imagebuildah/stage_executor.go
TomSweeneyRedHat: just reversed here and at line 445

TomSweeneyRedHat:
Happy green test buttons

rhatdan:
bors retry

rhatdan:
Bors seems to be acting out, for no reason.  Manually merging.

Issue #2153, opened 2020-02-10
make install.cni.sudo fails


mythi:
**Description**
The CNI git tree has no `build.sh` anymore so `make install.cni.sudo` fails.

**Steps to reproduce the issue:**
1.  `make install.cni.sudo` 

**Workaround:**
1.  `sed -i -e 's/build\.sh/build_linux\.sh/' Makefile`
2. `make install.cni.sudo`

rhatdan:
Please open a PR to fix?

mythi:
Done.

PR #2157, opened 2020-02-10
Makefile: fix install.cni.sudo

mythi:
github.com/containernetworking/plugins has no build.sh but
build_[linux|windows].sh. Call build_linux.sh directly to fix
install.cni.sudo.

While we're at it, create /opt/cni/bin with 'install' to encapsulate with
privileges.

Fixes: #2153

Signed-off-by: Mikko Ylinen <mikko.ylinen@intel.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,7 +101,7 @@ install.libseccomp.sudo: gopath
 install.cni.sudo: gopath
 	rm -rf ../../containernetworking/plugins
 	git clone https://github.com/containernetworking/plugins ../../containernetworking/plugins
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM but would like someone with a bit more CNI/Makefile knowledge to chime in.

@mythi TYVM for the PR!

vrothberg:
LGTM is not enough for bors. We need to review and approve.

vrothberg:
It's still confusing because we so many different bots across the projects.

rhatdan:
bors retry

rhatdan:
Bors does not like this for no reason.  Merging.

Issue #2147, opened 2020-02-06
digest test: needs a complete rewrite


edsantiago:
`tests/digest.bats` is confusing, misleading, and unreliable.

First, per Nalin, the goal of this test:

>We're checking that a given image can be pulled, used as a base image for a build, and that a new image that we build using that pulled image as a base can successfully be pushed to somewhere other than local storage, when the image is:
>
 >  * a v2s1 image referenced using its digest
 >  * a v2s1 image where one of the entries is a duplicate that gets discarded when it's converted to v2s2
> * a v2s1 image referenced using its tag
> * a v2s2 image referenced using its digest (incorrectly - this is also a v2s1 image, alpine@sha256:25fd8fc1aefcc8ae46aae23daefcd7dcb97f676fa0bc72bb0bf7cfb75df4f22e is a better candidate)
> * a v2s2 image referenced using its tag (incorrectly - this is the same v2s1 image, alpine:3.1 is a better candidate, though it's a manifest list now, so the test won't pass on non-amd64 systems without tweaks.
> * a v2s2 image referenced using its repository name, with neither a tag nor a digest. It's a manifest list, so the test won't pass on non-amd64 systems without tweaks.

The problems I see:

* as admitted above, the actual tests are wrong: several of the tests labeled "s2" are actually s1. This took me __far__ too long today to actually figure out.
* In addition, the current test script does not actually confirm `schemaVersion`. If a test is labeled "s2", it should darn well confirm that it is testing what it should be testing.
* Nor does the test script confirm that the image it's testing as "duplicate layers" actually has duplicate layers
* Some of the images are fetched by SHA, but others by name, making them moving targets
* And they're fetched from all sorts of different registries, some of which are unreliable, which gives us flaky tests
* And at least one of the images is HUGE, a monster, which of course slows everything down
* ~~Oh, I forgot: Nalin says "somewhere other than local storage" but the test doesn't actually test that: it only pushes to local storage.~~ Sorry, my misunderstanding. Per @nalind, `dir:XXX` exercises all the desired code paths - I misunderstood "local storage".

I believe that all images used in this test should be consistent, constant, guaranteed immutable, under our control, and from one and only one registry (presumably quay.io) (which, yes, is unreliable but SPOF is better than MPOF).

[...]

PR #2159, opened 2020-02-10
digest test : make more robust

edsantiago:
Instead of relying on random images from various separate
repositories, some of which are actually not the schema
version we thought we were testing (and some of which are
on unreliable registries, which causes test flakes),
craft our own images and store them all on quay.io

This commit:

  1) Adds a new script (tests/digest/make-v2sN) that
     crafts a custom image with the desired attributes:
     schema version 1 or 2, possibly with a duplicate
     layer. Script decides based on its name, that is,
     we are expected to have symlinks pointing to it.

  2) Updates digest.bats to use these newly-made images.

  3) Extends digest.bats by adding a check for 'v2s*'
     in the image name; if found, we now use 'imgtype'
     to extract the manifest and confirm that the
     pulled image has the expected schema version (1 or 2).

  4) Extends digest.bats by adding a check for duplicate
     layers in the one test that requires them.

Pushing the generated images to quay.io is left as an
exercise for the reader; as is updating the Digest SHAs
and/or tags in the individual tests in digest.bats

Fixes: #2147

### DIFF
--- a/tests/digest.bats
+++ b/tests/digest.bats
@@ -3,21 +3,23 @@
 load helpers
 
 fromreftest() {
-  run_buildah from --quiet --pull --signature-policy ${TESTSDIR}/policy.json $1
+  local img=$1
+
+  run_buildah from --quiet --pull --signature-policy ${TESTSDIR}/policy.json $img
   cid=$output
 
   # If image includes '_v2sN', verify that image is schema version N
-  expected_schemaversion=$(expr "$1" : '.*_v2s\([0-9]\)')
+  local expected_schemaversion=$(expr "$img" : '.*_v2s\([0-9]\)')
   if [ -n "$expected_schemaversion" ]; then
-      actual_schemaversion=$(imgtype -expected-manifest-type '*' -show-manifest $1 | jq .schemaVersion)
+      actual_schemaversion=$(imgtype -expected-manifest-type '*' -show-manifest $img | jq .schemaVersion)
       expect_output --from="$actual_schemaversion" "$expected_schemaversion" \
-                    ".schemaversion of $1"
+                    ".schemaversion of $img"
   fi
 
   # This is all we test: basically, that buildah doesn't crash when pushing
   pushdir=${TESTDIR}/fromreftest
   mkdir -p ${pushdir}/{1,2,3}
-  run_buildah push --signature-policy ${TESTSDIR}/policy.json $1 dir:${pushdir}/1
+  run_buildah push --signature-policy ${TESTSDIR}/policy.json $img dir:${pushdir}/1
   run_buildah commit --signature-policy ${TESTSDIR}/policy.json $cid new-image
   run_buildah push --signature-policy ${TESTSDIR}/policy.json new-image dir:${pushdir}/2
   run_buildah rmi new-image
@@ -32,17 +34,21 @@ fromreftest() {
 }
 
 @test "from-by-digest-s1-a-discarded-layer" {
-  IMG=quay.io/libpod/testdigest_v2s1_with_dups@sha256:70d6c767101c907aa251c21fded459c0f1a481685eb764ca2f7a6162e24dd81a
+  IMG=quay.io/libpod/testdigest_v2s1_with_dups@sha256:2c619fffbed29d8677e246798333e7d1b288333cb61c020575f6372c76fdbb52
 
   fromreftest ${IMG}
 
   # Verify that image meets our expectations (duplicate layers)
   # Surprisingly, we do this after fromreftest, not before, because fromreftest
   # has to pull the image for us.
-  dups=$(imgtype -expected-manifest-type '*' -show-manifest ${IMG} | jq .fsLayers|grep blobSum|sort|uniq -cd)
-  if [[ -z "$dups" ]]; then
-      die "Image ${IMG} does not have any duplicate layers (expected: one dup)"
-  fi
+  #
+  # Check that the first and second .fsLayers and .history elements are dups
+  local manifest=$(imgtype -expected-manifest-type '*' -show-manifest ${IMG})
+  for element in fsLayers history; do
+      local first=$(jq ".${element}[0]" <<<"$manifest")
+      local second=$(jq ".${element}[1]" <<<"$manifest")
+      expect_output --from="$second" "$first" "${IMG}: .${element}[1] == [0]"
+  done
 }
 
 @test "from-by-tag-s1" {
--- a/tests/digest.bats
+++ b/tests/digest.bats
@@ -3,9 +3,18 @@
 load helpers
 
 fromreftest() {
-  _prefetch $1
   run_buildah from --quiet --pull --signature-policy ${TESTSDIR}/policy.json $1
   cid=$output
+
+  # If image includes '_v2sN', verify that image is schema version N
+  expected_schemaversion=$(expr "$1" : '.*_v2s\([0-9]\)')
+  if [ -n "$expected_schemaversion" ]; then
+      actual_schemaversion=$(imgtype -expected-manifest-type '*' -show-manifest $1 | jq .schemaVersion)
+      expect_output --from="$actual_schemaversion" "$expected_schemaversion" \
+                    ".schemaversion of $1"
+  fi
+
+  # This is all we test: basically, that buildah doesn't crash when pushing
   pushdir=${TESTDIR}/fromreftest
   mkdir -p ${pushdir}/{1,2,3}
   run_buildah push --signature-policy ${TESTSDIR}/policy.json $1 dir:${pushdir}/1
@@ -13,30 +22,41 @@ fromreftest() {
   run_buildah push --signature-policy ${TESTSDIR}/policy.json new-image dir:${pushdir}/2
   run_buildah rmi new-image
   run_buildah commit --signature-policy ${TESTSDIR}/policy.json $cid dir:${pushdir}/3
+
   run_buildah rm $cid
   rm -fr ${pushdir}
 }
 
 @test "from-by-digest-s1" {
-  fromreftest k8s.gcr.io/pause@sha256:bbeaef1d40778579b7b86543fe03e1ec041428a50d21f7a7b25630e357ec9247
+  fromreftest quay.io/libpod/testdigest_v2s1@sha256:816563225d7baae4782653efc9410579341754fe32cbe20f7600b39fc37d8ec7
 }
 
 @test "from-by-digest-s1-a-discarded-layer" {
-  fromreftest libpod/whalesay@sha256:2413c2ffc29fb01d51c27a91b804079995d6037eed9e4b632249fce8c8708eb4
+  IMG=quay.io/libpod/testdigest_v2s1_with_dups@sha256:70d6c767101c907aa251c21fded459c0f1a481685eb764ca2f7a6162e24dd81a
+
+  fromreftest ${IMG}
+
+  # Verify that image meets our expectations (duplicate layers)
+  # Surprisingly, we do this after fromreftest, not before, because fromreftest
+  # has to pull the image for us.
+  dups=$(imgtype -expected-manifest-type '*' -show-manifest ${IMG} | jq .fsLayers|grep blobSum|sort|uniq -cd)
+  if [[ -z "$dups" ]]; then
+      die "Image ${IMG} does not have any duplicate layers (expected: one dup)"
+  fi
 }
 
 @test "from-by-tag-s1" {
-  fromreftest k8s.gcr.io/pause:0.8.0
+  fromreftest quay.io/libpod/testdigest_v2s1:20200210
 }
 
 @test "from-by-digest-s2" {
-  fromreftest alpine@sha256:e9cec9aec697d8b9d450edd32860ecd363f2f3174c8338beb5f809422d182c63
+  fromreftest quay.io/libpod/testdigest_v2s2@sha256:755f4d90b3716e2bf57060d249e2cd61c9ac089b1233465c5c2cb2d7ee550fdb
 }
 
 @test "from-by-tag-s2" {
-  fromreftest alpine:2.6
+  fromreftest quay.io/libpod/testdigest_v2s2:20200210
 }
 
 @test "from-by-repo-only-s2" {
-  fromreftest alpine
+  fromreftest quay.io/libpod/testdigest_v2s2
 }
--- a/tests/digest/README.md
+++ b/tests/digest/README.md
@@ -1,0 +1,28 @@
+This subdirectory contains a script used to create images for testing.
+
+To rephrase: this script is used **before testing**, not used **in** testing.
+_Much_ before testing (days/weeks/months/years), and manually.
+
+The script is `make-v2sN` but it is never invoked as such. Instead,
+various different symlinks point to the script, and the script
+figures out its use by picking apart the name under which it is called.
+
+As of the initial commit on 2020-02-10 there are three symlinks:
+
+* make-v2s1 - Create a schema 1 image
+* make-v2s2 - Create a schema 2 image
+* make-v2s1-with-dups - Create a schema 1 image with two identical layers
+
+If the script is successful, it will emit instructions on how to
+push the images to quay and what else you might need to do.
+
+Updating
+========
+
+Should you need new image types, e.g. schema version 3 or an image
+with purple elephant GIFs in it:
+
+1. Decide on a name. Create a new symlink pointing to `make-v2sN`
+1. Add the relevant code to `make-v2sN`: a conditional check at the top, the actual image-creating code, and if possible a new test to make sure the generated image is good
+1. Run the script. Verify that the generated image is what you expect.
+1. Add new test(s) to `digest.bats`
--- a/tests/digest/make-v2s1
+++ b/tests/digest/make-v2s1
@@ -1,0 +1,1 @@
+make-v2sN
--- a/tests/digest/make-v2s1-with-dups
+++ b/tests/digest/make-v2s1-with-dups
@@ -1,0 +1,1 @@
+make-v2sN
--- a/tests/digest/make-v2s2
+++ b/tests/digest/make-v2s2
@@ -1,0 +1,1 @@
+make-v2sN
--- a/tests/digest/make-v2sN
+++ b/tests/digest/make-v2sN
@@ -22,7 +22,7 @@ test -n "$schemaversion"    || die "Could not find 'v2s[12]' in basename"
 test "$schemaversion" = "N" && die "Script must be invoked via symlink"
 
 dup=
-if expr "$ME" : ".*-dup" >&/dev/null; then
+if expr "$ME" : ".*-dup" &>/dev/null; then
     dup="_with_dups"
 fi
 
@@ -78,17 +78,7 @@ buildah umount $cid2
 buildah commit -q $cid2 interim2
 
 layers="interim2 interim1"
-
-# If a dup layer is requested, create it now
-if [ -n "$dup" ]; then
-    cid3=$(buildah from interim2)
-    buildah commit -q $cid3 interim3
-    layers="interim3 $layers"
-
-    buildah tag interim3 my_image
-else
-    buildah tag interim2 my_image
-fi
+buildah tag interim2 my_image
 
 ###############################################################################
 #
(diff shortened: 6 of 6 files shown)


TomSweeneyRedHat:
I dare say you've outdone yourself this time @edsantiago 
It LGTM, but I'd like a head nod from @nalind about the digest testing as that's his baileywick.

edsantiago:
@nalind could I ask for your opinion and perhaps blessing on this?

tests/digest.bats
```diff
@@ -3,40 +3,60 @@
...
   run_buildah rmi new-image
   run_buildah commit --signature-policy ${TESTSDIR}/policy.json $cid dir:${pushdir}/3
+
   run_buildah rm $cid
   rm -fr ${pushdir}
 }
 
 @test "from-by-digest-s1" {
-  fromreftest k8s.gcr.io/pause@sha256:bbeaef1d40778579b7b86543fe03e1ec041428a50d21f7a7b25630e357ec9247
+  fromreftest quay.io/libpod/testdigest_v2s1@sha256:816563225d7baae4782653efc9410579341754fe32cbe20f7600b39fc37d8ec7
```
nalind: Both of these are schema 1 images, so this is good.

tests/digest.bats
```diff
@@ -3,40 +3,60 @@
...
+  # has to pull the image for us.
+  dups=$(imgtype -expected-manifest-type '*' -show-manifest ${IMG} | jq .fsLayers|grep blobSum|sort|uniq -cd)
+  if [[ -z "$dups" ]]; then
+      die "Image ${IMG} does not have any duplicate layers (expected: one dup)"
+  fi
 }
 
 @test "from-by-tag-s1" {
-  fromreftest k8s.gcr.io/pause:0.8.0
+  fromreftest quay.io/libpod/testdigest_v2s1:20200210
```
nalind: Both of these are schema 1 images, so this is good.

tests/digest.bats
```diff
@@ -3,40 +3,60 @@
...
 }
 
 @test "from-by-digest-s2" {
-  fromreftest alpine@sha256:e9cec9aec697d8b9d450edd32860ecd363f2f3174c8338beb5f809422d182c63
+  fromreftest quay.io/libpod/testdigest_v2s2@sha256:755f4d90b3716e2bf57060d249e2cd61c9ac089b1233465c5c2cb2d7ee550fdb
```
nalind: This _was_ a schema 1 image, which was a bug in the test, and it's being replaced with a schema 2 image, which is a marked improvement.  Can't say enough good things about this.

tests/digest.bats
```diff
@@ -3,40 +3,60 @@
...
 }
 
 @test "from-by-tag-s2" {
-  fromreftest alpine:2.6
+  fromreftest quay.io/libpod/testdigest_v2s2:20200210
```
nalind: This was also a schema 1 image, which was a bug in the test, and it's being replaced with a schema 2 image, which again is a marked improvement.

tests/digest.bats
```diff
@@ -3,40 +3,60 @@
...
 }
 
 @test "from-by-repo-only-s2" {
-  fromreftest alpine
+  fromreftest quay.io/libpod/testdigest_v2s2
```
nalind: This became a manifest list, possibly after the test was added, which is a bug in the test.  This change fixes it, so again, very much an improvement.

tests/digest.bats
```diff
@@ -3,40 +3,60 @@
...
+
+  fromreftest ${IMG}
+
+  # Verify that image meets our expectations (duplicate layers)
+  # Surprisingly, we do this after fromreftest, not before, because fromreftest
+  # has to pull the image for us.
+  dups=$(imgtype -expected-manifest-type '*' -show-manifest ${IMG} | jq .fsLayers|grep blobSum|sort|uniq -cd)
+  if [[ -z "$dups" ]]; then
+      die "Image ${IMG} does not have any duplicate layers (expected: one dup)"
+  fi
```
nalind: This test broke when we moved to using `libpod/whalesay`, since the new schema 1 image didn't have `fsLayers` and `history` lists in which with both the first and second entries were the identical.

I think we want to do a `jq '.fsLayers[].blobSum' | head -n 2 | uniq -cd` and `jq '.history[].v1Compatibility' | head -n 2 | uniq -cd` to detect that the image looks right, and find or make an image that looks more like `docker/whalesay@sha256:178598e51a26abbc958b8a2e48825c90bc22e641de3d31e18aaf55f3258ba93b` to use in the test.

edsantiago: I'm super confused. Big wall o' questions coming.

First: when you say "this test broke", do you mean "the test started failing", or "the test was no longer testing what it was supposed to test"? I'm going to assume the latter, and that's exactly why my PR includes image tests that try to confirm image properties. (If you meant the former, or if I've misunderstood, please help me out).

Second... from my reading of git history, it looks like 86fa0803 introduced the `whalesay` image check and the next commit, ae7d2f35 less than one second later, also had something to do with duplicates. So I was expecting buildah-push to fail when run against 86fa0803^ -- and it does. But it fails in exactly the same way with ae7d2f35!
```console
++ ./buildah from --quiet --pull docker/whalesay@sha256:178598e51a26abbc958b8a2e48825c90bc22e641de3d31e18aaf55f3258ba93b
+ cid=whalesay-working-container-5
+ ./buildah push docker/whalesay@sha256:178598e51a26abbc958b8a2e48825c90bc22e641de3d31e18aaf55f3258ba93b dir:/tmp/bp/1
Copying blob sha256:1154ba695078d29ea6c4e1adb55c463959cd77509adf09710e2315827d66271a
 188.05 MiB / 188.05 MiB [==================================================] 2s
Copying blob sha256:528c8710fd95f61d40b8bb8a549fa8dfa737d9b9c7c7b2ae55f745c972dddacd
ERRO[0002] error unmounting /var/lib/containers/storage/overlay/47a246479ce6a91248b69c2f79f3daabf527b3a2e4b68e4c0a1da8f8122d307f/merged: no such file or directory
error pushing image "docker/whalesay@sha256:178598e51a26abbc958b8a2e48825c90bc22e641de3d31e18aaf55f3258ba93b" to "dir:/tmp/bp/1": error copying layers and metadata: Error reading blob sha256:528c8710fd95f61d40b8bb8a549fa8dfa737d9b9c7c7b2ae55f745c972dddacd: error creating overlay mount to /var/lib/containers/storage/overlay/47a246479ce6a91248b69c2f79f3daabf527b3a2e4b68e4c0a1da8f8122d307f/merged: no such file or directory
```
What I am trying to do is figure out what is special about this whalesay, then get a version of buildah that fails with it, then get a version that passes, then try to craft my own custom image and see if we can use that instead of whalesay. Can you help me understand the buildah failures above, and help me isolate working/nonworking buildah commits for me to try out? Thank you!

edsantiago: OBTW I think we should keep my `quay.io/libpod/testdigest_v2s1_with_dups` because with buildah ae7d2f35 it causes an interesting new failure:
```console
#  ./buildah commit testdigest_v2s1_with_dups-working-container-7 new-image
error committing container "testdigest_v2s1_with_dups-working-container-7" to "new-image": error copying layer and metadata: error reading configuration to write to image "containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.override_kernel_check=true]docker.io/library/new-image:latest": internal error: history lists 1 non-empty layers, but we have 4 layers on disk
```
So the exercise now is to create yet another magic image that causes old-buildah to fail but new-buildah to pass.

edsantiago: DISCREPANCY SOLVED! I was using the default `overlay`, which of course doesn't work with buildah. Switching to `vfs` yields the expected results.

To recap: 86fa080 and ae7d2f35 are a two-part commit that addressed the whalesay image problem: after ae7d2f35, the `digest.bats` whalesay test passes. _However_, even with vfs and even with this commit, my hand-crafted `v2s1_with_dups` image causes buildah to fail:
```console
# ./buildah --storage-driver vfs commit testdigest_v2s1_with_dups-working-container-3 new-image
error committing container "testdigest_v2s1_with_dups-working-container-3" to "new-image": error copying layer and metadata: error reading configuration to write to image "containers-storage:[vfs@/var/lib/containers/storage+/var/run/containers/storage:vfs.override_kernel_check=true]docker.io/library/new-image:latest": internal error: history lists 1 non-empty layers, but we have 4 layers on disk
```
Given all this, I humbly suggest that my `testdigest_v2s1_with_dups` image is a better test than whalesay. Can we get this merged, and add new test cases (and images) as we find more failure cases?

rhatdan:
@nalind PTAL so we can get this merged.

edsantiago:
Buildah gating tests are [failing on f32](https://jenkins-continuous-infra.apps.ci.centos.org/blue/organizations/jenkins/fedora-f32-build-pipeline/detail/fedora-f32-build-pipeline/63/pipeline/#step-248-log-1061), in digest tests. I saw similar failures on (manual) f31 tests yesterday.

Could we pretty-please get this reviewed and merged so we can all waste less time reviewing test failures?

tests/digest.bats
nalind: > I'm super confused. Big wall o' questions coming.
> 
> First: when you say "this test broke", do you mean "the test started failing", or "the test was no longer testing what it was supposed to test"? I'm going to assume the latter, and that's exactly why my PR includes image tests that try to confirm image properties. (If you meant the former, or if I've misunderstood, please help me out).

The latter, as the new image being used no longer had the characteristic that triggered the code paths we cared about for the test.

> Second... from my reading of git history, it looks like [86fa080](https://github.com/containers/buildah/commit/86fa0803e8991b2c970c1c248d8ff350c39a50ac) introduced the `whalesay` image check and the next commit, [ae7d2f3](https://github.com/containers/buildah/commit/ae7d2f354732f40e113163e8b2a214601517bae2) less than one second later, also had something to do with duplicates. So I was expecting buildah-push to fail when run against [86fa080](https://github.com/containers/buildah/commit/86fa0803e8991b2c970c1c248d8ff350c39a50ac)^ -- and it does. But it fails in exactly the same way with [ae7d2f3](https://github.com/containers/buildah/commit/ae7d2f354732f40e113163e8b2a214601517bae2)!
> 
> ```
> ++ ./buildah from --quiet --pull docker/whalesay@sha256:178598e51a26abbc958b8a2e48825c90bc22e641de3d31e18aaf55f3258ba93b
> + cid=whalesay-working-container-5
> + ./buildah push docker/whalesay@sha256:178598e51a26abbc958b8a2e48825c90bc22e641de3d31e18aaf55f3258ba93b dir:/tmp/bp/1
> Getting image source signatures
> Copying blob sha256:1154ba695078d29ea6c4e1adb55c463959cd77509adf09710e2315827d66271a
>  188.05 MiB / 188.05 MiB [==================================================] 2s
> Copying blob sha256:528c8710fd95f61d40b8bb8a549fa8dfa737d9b9c7c7b2ae55f745c972dddacd
> ERRO[0002] error unmounting /var/lib/containers/storage/overlay/47a246479ce6a91248b69c2f79f3daabf527b3a2e4b68e4c0a1da8f8122d307f/merged: no such file or directory
> error pushing image "docker/whalesay@sha256:178598e51a26abbc958b8a2e48825c90bc22e641de3d31e18aaf55f3258ba93b" to "dir:/tmp/bp/1": error copying layers and metadata: Error reading blob sha256:528c8710fd95f61d40b8bb8a549fa8dfa737d9b9c7c7b2ae55f745c972dddacd: error creating overlay mount to /var/lib/containers/storage/overlay/47a246479ce6a91248b69c2f79f3daabf527b3a2e4b68e4c0a1da8f8122d307f/merged: no such file or directory
> ```
> 
> What I am trying to do is figure out what is special about this whalesay, then get a version of buildah that fails with it, then get a version that passes, then try to craft my own custom image and see if we can use that instead of whalesay. Can you help me understand the buildah failures above, and help me isolate working/nonworking buildah commits for me to try out? 

The manifest in the image is a v2s1 manifest, and in that manifest, the first entries in both the `fsLayers` and `history` lists are identical to the second entries in those lists, which can be confirmed by using `skopeo inspect --raw`.  The duplicate values are in consecutive entries.

While `quay.io/libpod/testdigest_v2s1_with_dups@sha256:70d6c767101c907aa251c21fded459c0f1a481685eb764ca2f7a6162e24dd81a` looks like it would make a useful test, particularly if we fail to commit a new image based on it (or used to fail, but don't any more), it doesn't trigger duplicate elimination logic in containers/image that logic in buildah needs to agree with.

This can be confirmed by counting the number of `fsLayer` entries in the v2s1 manifest using `skopeo inspect --raw`, and comparing that number to the number of `rootfs.diff_ids` values we get when we ask `skopeo inspect --config` to produce a v2s2 configuration blob, which it has to generate from the manifest.  Part of the process for generating that configuration blob includes dropping the duplicates from the manifest.

`skopeo` doesn't have a CLI option for requesting the manifest in v2s2 format, but then this is the first occasion I can think of when one might come in handy.

edsantiago: Thanks for your response. If I understand you correctly, the magic characteristics that are desired are that `.fsLayers` be greater than `.rootfs.diff_ids`, and any image with those characteristics should make a test case. Is that correct? 

If so, is there a trick to getting such an image uploaded to quay.io? I've hand-crafted an image (by editing `manifest.json`) such that I believe it matches those constraints:
```console
# skopeo inspect --raw containers-storage:7aaf3 | jq '.fsLayers | length'
WARN[0000] Not using native diff for overlay, this may cause degraded performance for building images: kernel has CONFIG_OVERLAY_FS_REDIRECT_DIR enabled
5

# skopeo inspect --config containers-storage:7aaf | jq '.rootfs.diff_ids | length'
WARN[0000] Not using native diff for overlay, this may cause degraded performance for building images: kernel has CONFIG_OVERLAY_FS_REDIRECT_DIR enabled
4
```
...but when I push it to quay.io, and pull it back, `.fsLayers` is back to four. I've tried the following push commands:
```console
# skopeo copy containers-storage:localhost/testdigest_v2s1_with_dups:latest docker://quay.io/libpod/testdigest_v2s1_with_dups:$(date +%Y%m%d)

# podman push localhost/testdigest_v2s1_with_dups:latest quay.io/libpod/testdigest_v2s1_with_dups:$(date +%Y%m%d)
```
And I always pull with `buildah`.

Is this something `quay.io` is doing, or is it my mistake in podman/skopeo push? Is there some way to get this image uploaded to quay, or is this a futile exercise?

edsantiago:
Well, I've given up. I can't figure out a way to create an equivalent container and preserve the characteristics when pushed to `quay.io`; I'm assuming that the bad whalesay image is a product of an earlier time, and all tools have checks now to make it impossible to push such an image to a modern repo.

I've removed the now-unused 'make-dup' functionality, but am leaving the `make-v2sN` script because it's still used for two other images and I believe it's important to have reproducible test images. I've undone the `quay.io/whalesay` edit, reverting back to `docker.io/whalesay`, and have added a minimal sanity check to make sure we don't accidentally go back to a different image. I've also added two retries in hopes of alleviating the common whalesay flakes.

I'd still like to get this reviewed and merged, because I still believe it's a more accurate and robust test.

edsantiago:
Gating tests [are failing](https://jenkins-continuous-infra.apps.ci.centos.org/blue/organizations/jenkins/fedora-rawhide-build-pipeline/detail/fedora-rawhide-build-pipeline/9993/pipeline/#step-248-log-1060); this PR would fix that.

tests/digest.bats
```diff
@@ -3,40 +3,92 @@
...
 
 @test "from-by-digest-s1-a-discarded-layer" {
-  fromreftest libpod/whalesay@sha256:2413c2ffc29fb01d51c27a91b804079995d6037eed9e4b632249fce8c8708eb4
+  # This image is responsible for a large number of test flakes, because it's
+  # huge and susceptible to network timeouts. Unfortunately, there doesn't
+  # seem to be anything we can do about it: the world has evolved since
+  # this image was created, and there no longer seem to be any tools that
+  # can create a similar image. (Yes, this means our testing is at the
+  # mercy of docker.io not removing that precise image from their registry)
+  IMG=docker.io/docker/whalesay@sha256:178598e51a26abbc958b8a2e48825c90bc22e641de3d31e18aaf55f3258ba93b
```
TomSweeneyRedHat: should we snab a copy of the image and tuck it away in quay.io somwehere?

TomSweeneyRedHat: regardless of the answer,  I wouldn't hold the merge back for this question.

TomSweeneyRedHat:
LGTM and happy green test buttons.  Would like a head nod from @nalind if possible before merging.

tests/digest.bats
edsantiago: We can't: that's the entire point of my wasted few weeks of effort.

It seems like this image was magically written by earlier tools that (waves hands) did something that modern tools can no longer do. In other words: it is impossible to push this image anywhere and have it keep the properties that make it interesting. Current tools (buildah, skopeo, podman) somehow flatten/smoosh/deduplicate the layers on any new write. All we have is this image, on docker.io, and all we can do is hope it never changes.

Disclaimer: I have no idea what the technical terms are for any of this; I just know I've tried really really hard for several weeks and all my attempts fail on push.

nalind: The current versions of our tools are dutifully cleaning up confusing bits of old data, which is rather inconvenient for us when we want to preserve those bits.

After some experimentation, it looks like having skopeo's vendored copy of containers/image not call `s1.fixManifestLayers()` in `github.com/containers/image/v5/manifest.Schema1FromManifest()` is enough to let it copy the image from docker hub to my test repository with its duplicate `blobSum` and `v1Compatibility` values intact.

I uploaded the change I made relative to [skopeo master](https://github.com/containers/skopeo/tree/e31d5a0e8fc61241b02b6eb06aeeb39ed4de7b5e) as a gist [here](https://gist.github.com/nalind/b491204ff05c3c3f3b6ef014b333a60c).  Hopefully it'll help preserve this detail for us.

tests/digest.bats
nalind: > Thanks for your response. If I understand you correctly, the magic characteristics that are desired are that `.fsLayers` be greater than `.rootfs.diff_ids`, and any image with those characteristics should make a test case. Is that correct?

No, their lengths relative to each other are not a key factor.  It's that the first two elements of the `fsLayers` are identical to each other, and the first two elements of the `history` are identical to each other.  The `history` values even have the same `id` and `parent` values in their `v1Compatibility` fields.

> Is this something `quay.io` is doing, or is it my mistake in podman/skopeo push? Is there some way to get this image uploaded to quay, or is this a futile exercise?

It appears to be something that happens at the client before it attempts to upload the manifest, normally a good thing.

edsantiago: Thanks, Nalin. Using a `skopeo` doctored with your gist, I've created and uploaded `quay.io/libpod/testdigest_v2s1_with_dups:20200323`; and I've confirmed that upon pull, 'buildah manifest' includes the duplicate layers. 

Could I ask you to pull it and confirm that it meets your requirements for a test image? If it does, I'll spend the rest of the time updating the test scripts and documentation then resubmit for review.

nalind: Yup, looks good!  Inspecting with `skopeo inspect --raw` shows the duplicates, and they're omitted from the configuration it synthesizes when I use `skopeo inspect --config`.

rhatdan: Does this mean we can merge?

edsantiago: Not right now. I made the stupid, stupid, stupid mistake of rebasing against master. Tests are broken, and it's nothing to do with anything in this PR. Right now it looks like #2245 but I can't understand how that passed CI. I'll keep investigating.

edsantiago:
Can someone with privs rerun the failing test? I think it's a flake.

The more I look at it, the less I understand it: the test that failed is `something-http-pre` and is bytewise identical to the test immediately after, which is `something-http-post`... which passes.  There are two errors:
```
[+0312s] # 2020/03/25 20:14:00 listen tcp :33192: bind: address already in use
[+0312s] # error prepping temporary context directory: error getting "http://0.0.0.0:33192/context.tar": Get http://0.0.0.0:33192/context.tar: dial tcp 0.0.0.0:33192: connect: connection refused
```
I'm guessing the first one is from `httpd`, the second from `buildah` (the helper script starts httpd but doesn't actually confirm that it succeeds). I don't know why port 33192 would be in use.

Finally, I can't reproduce it on my own test setup. Has anyone seen this before?

edsantiago:
Yay, tests green!

For ease of review, I've left this as two commits. The second is the one that adds and documents the dup-layer image.

edsantiago:
Followup: buildah gating tests [are green](https://bodhi.fedoraproject.org/updates/FEDORA-2020-7f2d2a418f); yesterday's errors (timeouts on docker.io) probably deserve investigation and fixing so we continue removing our reliance on docker.io, but I can get to that next week.

Thanks again @nalind for your patience and help with this one.

PR #2172, opened 2020-02-17
manifest push --format: force an image type, not a list type

nalind:
The image library's image copying API expects the type of a single image in the options structure's `ForceManifestMIMEType` field, and it infers the corresponding list type when it's copying multiple images.

Previously, though, we were passing in a _list_ MIME type, which was also being used as the target type when converting single images, which would fail.  Switch to passing in the single image type instead, and infer the single-image type in our internal API if a consumer passes us a list type.

This makes it easier to notice that we force v2s1 images to be converted to v2s2 images when we push them as part of a list, even if the registry doesn't reject v2s1 images.

### DIFF
--- a/manifests/manifests.go
+++ b/manifests/manifests.go
@@ -203,6 +203,18 @@ func (l *list) Push(ctx context.Context, dest types.ImageReference, options Push
 		}
 	}()
 
+	// If we were given a media type that corresponds to a multiple-images
+	// type, reset it to a valid corresponding single-image type, since we
+	// already expect the image library to infer the list type from the
+	// image type that we're telling it to force.
+	singleImageManifestType := options.ManifestType
+	switch singleImageManifestType {
+	case v1.MediaTypeImageIndex:
+		singleImageManifestType = v1.MediaTypeImageManifest
+	case manifest.DockerV2ListMediaType:
+		singleImageManifestType = manifest.DockerV2Schema2MediaType
+	}
+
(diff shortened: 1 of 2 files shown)


rhatdan:

manifests/manifests.go
```diff
@@ -203,6 +203,18 @@ func (l *list) Push(ctx context.Context, dest types.ImageReference, options Push
 		}
 	}()
 
+	// If we were given a media type that corresponds to a multiple-images
+	// type, reset it to a valid corresponding single-image type, since we
+	// already expect the image library to infer the list type from the
+	// image type that we're telling it to force.
+	singleImageManifestType := options.ManifestType
+	switch singleImageManifestType {
+	case v1.MediaTypeImageIndex:
```
TomSweeneyRedHat: I'll just note that I'd prefer 'v1' to be `imgspecv1` in this file as it is in cmd/buildah/manifest.go.

rhatdan:
bors retry

vrothberg:
bors retry

rhatdan:
@cevich Any chance we can get some info from bors telling why it is failing.
Force merging.

cevich:
> Any chance we can get some info from bors telling why it is failing.

@rhatdan yes, click the big-red-X on the right, just above the 'build failed' comment.  In this case it's the `cirrus-ci/only_prs/gate` task.  Looks like a false-positive git-validation finding.

PR #2174, opened 2020-02-19
Improve remote manifest retrieval

saschagrunert:
Before this patch, it was not possible to retrieve a remote manifest
just by specifying the image name, like:

```
> buildah manifest inspect alpine
Invalid image name "alpine", expected colon-separated transport:reference
```

It was possible to get the manifest via:

```
> buildah manifest inspect docker://alpine
…
```

But after pulling the image into the local storage, this works not any
more:

```
> buildah pull alpine
e7d92cdc71feacf90708cb59182d0df1b911f8ae022d29e8e95d75ca6a99776a
> buildah manifest inspect docker://alpine
manifest from image … is of type
"application/vnd.docker.distribution.manifest.v2+json", which
is not a list type
ERRO exit status 1
```

This means we now collect a list of possible local or remote manifests
and try to resolve them sequentially. This enables us to fallback to the
remote location if the locally fetched manifest is not an actual
manifest. It also enables us to see the remote manifest via:

```
> ./buildah manifest inspect alpine
{
    …
}
```

### DIFF
--- a/cmd/buildah/manifest.go
+++ b/cmd/buildah/manifest.go
@@ -19,6 +19,7 @@ import (
 	digest "github.com/opencontainers/go-digest"
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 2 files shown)


cmd/buildah/manifest.go
```diff
@@ -555,31 +556,52 @@ func manifestInspectCmd(c *cobra.Command, args []string, opts manifestInspectOpt
...
-	defer src.Close()
+	for _, ref := range refs {
+		logrus.Debugf("Testing reference %q for possible manifest", transports.ImageName(ref))
 
-	manifestBytes, manifestType, err := src.GetManifest(ctx, nil)
-	if err != nil {
-		return errors.Wrapf(err, "error loading manifest from image %q", transports.ImageName(ref))
+		src, err := ref.NewImageSource(ctx, systemContext)
+		if err != nil {
+			latestErr = errors.Wrapf(err, "error reading image %q", transports.ImageName(ref))
```
rhatdan: Should we print out the latestErr for each failure?
```
if err != nil {
     if latestErr != nil {
             logurs.Errorf(latestErr)
    }
   latestErr = errors.Wrapf(err, "error reading image %q", transports.ImageName(ref))
}
```

saschagrunert: Yes makes sense, I prefer in having them as `Debug` messages since we only know if its an error after the loop has ended.

rhatdan: That would be ok, but I would like to be able to retrieve the data rather then just drop it.

TomSweeneyRedHat:
Changes LGTM, but a test is failing, I think it's a flake and have restarted it.

cmd/buildah/manifest.go
saschagrunert: Hm, we could concatenate / wrap the errors if that helps

rhatdan: Wrapping it would be just fine.

saschagrunert: I added some logic to wrap the errors and return them in one bunch. I'm not sure if the resulting error message is helpful for users. PTAL :innocent:

TomSweeneyRedHat:
LGTM

rhatdan:
bors retry
bors you have one more chance.

PR #2178, opened 2020-02-19
Clear the right variable in buildahimage

nalind:
These Dockerfiles attempted to ensure that the _BUILDAH_STARTED_IN_USERNS variable was set to an empty value, but the variable that we check was changed to _CONTAINERS_USERNS_CONFIGURED some time ago.

The logic that checks for the variable, though, treats it not being set and being set to an empty value the same way, so we needn't bother setting it.


TomSweeneyRedHat:
LGTM

PR #2181, opened 2020-02-21
[release-1.11-rhel] check for .dockerignore specifically

nalind:
This PR pulls in two changes that landed between v1.13.0 and v1.13.1, originally as part of #2072.

* When generating the list of exclusions to process .dockerignore contents, don't include .dockerignore if we don't have a .dockerignore file in the context directory.  That way, if the file doesn't exist, and the caller didn't pass in any patterns, we get no patterns instead of just one ".dockerignore" pattern, and we can hit the faster copy path.
* When copying, close source files after we've finished reading from them, rather than leaving it for later.

### DIFF
--- a/add.go
+++ b/add.go
@@ -215,7 +215,12 @@ func dockerIgnoreMatcher(lines []string, contextDir string) (*fileutils.PatternM
 	if contextDir == "" {
 		return nil, nil
 	}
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

rhatdan:
Not living in bors hell tonight.

Issue #2113, opened 2020-01-26
runc not found (debian, project karmic)


astronouth7303:
**Description**

Basically the same as #1754 but with a different distro and package.

Buildah is unable to find `runc`. I have it at `/usr/sbin/runc` from the `runc` package provided by debian. It's normal for debian to not have the sbin paths on `PATH` for normal users.

**Steps to reproduce the issue:**
```sh
#!/bin/sh
set -x
build=$(buildah from python:3-slim)
buildah run $build -- pip install poetry
```

**Describe the results you received:**
```
🐚 sh test.sh                                                                               master
+ buildah from python:3-slim
+ build=python-working-container-18
+ buildah run python-working-container-18 -- pip install poetry
error running container: error creating container for [/usr/local/bin/pip install poetry]: : exec: "runc": executable file not found in $PATH
ERRO exit status 1  
```

**Describe the results you expected:**
Successful running of a command

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah/unknown,now 1.12.0-5 amd64 [installed]
buildah/unknown 1.12.0-5 arm64
buildah/unknown 1.12.0-5 armhf
buildah/testing 1.10.1-5 i386
buildah/unknown 1.12.0-5 ppc64el
```

**Output of `buildah version`:**

```
Version:         1.12.0
Go Version:      go1.13.5
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.1-dev
CNI Spec:        0.4.0
libcni Version:  
image Version:   5.0.0
Git Commit:      
Built:           Thu Jan 23 23:11:07 2020
OS/Arch:         linux/amd64
```

**Output of `cat /etc/*release`:**

```
PRETTY_NAME="Debian GNU/Linux bullseye/sid"
NAME="Debian GNU/Linux"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

**Output of `uname -a`:**

```
Linux cayde7 5.4.0-2-amd64 #1 SMP Debian 5.4.8-1 (2020-01-05) x86_64 GNU/Linux
```

**Output of `cat /etc/containers/storage.conf`:**

(Does not exist)

TomSweeneyRedHat:
@lsm5 PTAL

lsm5:
> @lsm5 PTAL

[...]

PR #2182, opened 2020-02-22
Search for local runtime per values in containers.conf

TomSweeneyRedHat:
After determining the type of runtime to use,
either "runc" or "crun" dependent upon the system, search
the list of that type of runtime in the containers.conf
file.  It includes the location of those runtimes in a
number of different architectures.  Once found, set the
runtime to use to that value.

Fixes: #2113

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -678,6 +678,11 @@ func runUsingRuntime(isolation Isolation, options RunOptions, configureNetwork b
 	runtime := options.Runtime
 	if runtime == "" {
 		runtime = util.Runtime()
+
+		localRuntime := util.FindLocalRuntime(runtime)
+		if localRuntime != "" {
+			runtime = localRuntime
+		}
 	}
 
 	// Default to just passing down our stdio.
--- a/util/util.go
+++ b/util/util.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"syscall"
 
+	"github.com/containers/common/pkg/config"
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/pkg/sysregistriesv2"
 	"github.com/containers/image/v5/signature"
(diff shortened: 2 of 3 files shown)


util/util.go
```diff
@@ -432,3 +433,33 @@ var (
 	isUnified     bool
 	isUnifiedErr  error
 )
+
+//FileExistsAndNotADir - Check to see if a file exists
+//and that it is not a directory.
+func FileExistsAndNotADir(path string) bool {
```
rhatdan: Does this need to be public?

util/util.go
```diff
@@ -432,3 +433,33 @@ var (
...
+	}
+	return !file.IsDir()
+}
+
+// FindLocalRuntime find the local runtime of the
+// system searching through the config file for
+// possible locations.
+func FindLocalRuntime(runtime string) string {
+	var localRuntime string
+	conf, err := config.NewConfig("")
```
rhatdan: Use config.Default()

util/util.go
TomSweeneyRedHat: When I was first putting this together yes as I'd the `FindLocalRuntime()` function in run_linux.go.  After I moved this I thought this function might prove to be useful elsewhere at some future time, but since it's not being used yet I probably ought to make it private.

util/util.go
TomSweeneyRedHat: flopped over to use as suggested.

run_linux.go
```diff
@@ -680,6 +680,11 @@ func runUsingRuntime(isolation Isolation, options RunOptions, configureNetwork b
 		runtime = util.Runtime()
 	}
 
+	localRuntime := util.FindLocalRuntime(runtime)
+	if localRuntime != "" {
+		runtime = localRuntime
+	}
+
```
nalind: Whoa, pretty sure we don't want to override a value that was passed to our API with a default from the configuration file.

TomSweeneyRedHat: What's coming back from `util.Runtime()` is either `runc` or `crun`.  In some systems, most likely with an ill-defined `$PATH` variable, they are failing as they can't find the runtime to run it.

`FindLocalRuntime` retrieves a map of the runtime passed in ("runc" or "crun") from containers.conf which has the possible locations of the runtime on the system.  It then searches the locations until it finds the first one off the list.

This is similar to what libpod does, but they tuck the retrieved value away differently than we do and also get it a bit more frequently than we do.

Happy to adjust as necessary, I was just trying to model Libpod/Podman's functionality here for this.

nalind: I'm talking about the `options.Runtime` field (line 678).  If that's set, we need to honor it.  The current version of this patch overwrites it unconditionally.

TomSweeneyRedHat: Ah, missed that point.  Good catch.  I'll add a check to make sure `options.Runtime` is empty before doing this processing.  TYVM!

PR #2184, opened 2020-02-23
Replace unix with syscall to allow vendoring into libpod

rhatdan:
unix constants do not work when vendored into libpod, because
libpod can potentially be used on non unix platforms, so
compilation blows up.  Switching to syscall should fix this issue.

### DIFF
--- a/common.go
+++ b/common.go
@@ -7,6 +7,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"syscall"
 	"time"
 
 	"github.com/containers/common/pkg/unshare"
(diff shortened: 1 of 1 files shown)


rhatdan:
@TomSweeneyRedHat @vrothberg @mheon @giuseppe @baude PTAL

TomSweeneyRedHat:
LGTM
But I must say I'm surprised by the small number of changes, I would have guessed we'd more instances.

mheon:
Huh. LGTM. Syscall is deprecated, so the right way to do this would be conditional compilation, but errno constants ought to be fine.

rhatdan:
@TomSweeneyRedHat You need to approve this so bors will succeed, I believe.

rhatdan:
Forcing merge.

Issue #1991, opened 2019-11-20
buildah from, with error, leaks mounts


edsantiago:
`buildah from` seems to mount a tmpfs working directory. On success, this is unmounted. On failure, it is left behind on the system:
```
# mkdir -p /tmp/aaa/root
# buildah --root /tmp/aaa/root --storage-driver overlay from --pull=false alpine
[ lots of errors, expected ]

# mount|grep aaa
tmpfs on /tmp/aaa/root/overlay type tmpfs (rw,nosuid,nodev,seclabel)   <<< not expected
```
This causes failures on gating tests when `STORAGE_DRIVER=overlay` because the temporary working directories cannot be cleaned up.

buildah-1.11.4-2.fc30.x86_64 and buildah-1.11.4-2.module+el8.2.0+4595+e1e23178.x86_64

rhatdan:
@edsantiago I believe this is fixed, I was not able to recreate it.  Reopen if I am mistaken.

edsantiago:
No, it's not fixed, but I don't have privs to reopen.

Looks like at some point in time, `--pull=false` broke (or changed behavior without actually documenting that). The new recipe is:
```console
# buildah --root /tmp/aaa/root --storage-driver overlay from --pull-never alpine
[lots of errors]
# mount|grep aaa
tmpfs on /tmp/aaa/root/overlay type tmpfs (rw,nosuid,nodev,seclabel)
```

buildah-1.15.0-1.fc32

I'm kind of concerned about this `--pull=false` change. I'm astonished there's no test for it. Should I file an issue, or is this really some sort of new desired behavior?

@rhatdan, please reopen.

TomSweeneyRedHat:
Taking one off @rhatdan's illicit closure list!  ;^)

rhatdan:
Great @edsantiago thanks.

vrothberg:
> I'm kind of concerned about this --pull=false change. I'm astonished there's no test for it. Should I file an issue, or is this really some sort of new desired behavior?

If the behavior has changed, please open an issue. It screams for tests.

PR #2188, opened 2020-02-26
BATS : in teardown, umount stale mounts

edsantiago:
Workaround for #1991, in which buildah leaves behind stale
mounts under overlayfs, which causes the 'rm -fr' in teardown()
to barf with EBUSY. Here we amend teardown() to look for
mounts underneath our temp dir and to umount them. With
this, bats runs cleanly even with STORAGE_DRIVER=overlay

### DIFF
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -41,6 +41,16 @@ function stophttpd() {
 
 function teardown() {
 	stophttpd
+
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@giuseppe @vrothberg @TomSweeneyRedHat @nalind @QiWang19 PTAL

QiWang19:
LGTM

PR #2189, opened 2020-02-26
Set correct ownership on working directory

umohnani8:
We want to set the correct ownership on the working
directory so that the build process can actually write
to this directory.

Using image registry.access.redhat.com/ubi8/go-toolset:1.12.8-32 as the example here which has user set to 1001 (default).

When building a simple Dockerfile like
```
FROM registry.access.redhat.com/ubi8/go-toolset:1.12.8-32
WORKDIR $HOME/build/
RUN ls -al $HOME
RUN touch testfile
```

We get this error
```
STEP 1: FROM registry.access.redhat.com/ubi8/go-toolset:1.12.8-32 AS build
STEP 2: LABEL stage=build
STEP 3: RUN id
uid=1001(default) gid=0(root) groups=0(root)
STEP 4: WORKDIR $HOME/build/
STEP 5: RUN ls -al $HOME
total 28
drwxr-xr-x. 1 default root 4096 Feb 26 22:01 .
drwxr-xr-x. 1 default root 4096 Jan 29 20:33 ..
drwxr-xr-x. 1 default root 4096 Jan 29 20:33 .pki
drwxr-xr-x. 2 root root 4096 Feb 26 22:01 build
STEP 6: RUN touch testfile
touch: cannot touch 'testfile': Permission denied
error building at STEP "RUN touch testfile": error while running runtime: exit status 1
```
Here we can see that the working directory `build` has `root root` as the ownership, hence touching a file as user 1001 fails with permission denied.

With this patch, the same Dockerfile works fine and this is the output
```
STEP 1: FROM registry.access.redhat.com/ubi8/go-toolset:1.12.8-32 AS build
STEP 2: LABEL stage=build
STEP 3: RUN id
uid=1001(default) gid=0(root) groups=0(root)
STEP 4: WORKDIR $HOME/build/
STEP 5: RUN ls -al $HOME
total 28
drwxr-xr-x. 1 default root 4096 Feb 26 22:01 .
drwxr-xr-x. 1 default root 4096 Jan 29 20:33 ..
drwxr-xr-x. 1 default root 4096 Jan 29 20:33 .pki
drwxr-xr-x. 2 default root 4096 Feb 26 22:01 build
STEP 6: RUN touch testfile
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/containers/buildah"
 	buildahdocker "github.com/containers/buildah/docker"
+	"github.com/containers/buildah/pkg/chrootuser"
 	"github.com/containers/buildah/util"
 	cp "github.com/containers/image/v5/copy"
 	"github.com/containers/image/v5/docker/reference"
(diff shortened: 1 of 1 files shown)


umohnani8:

TomSweeneyRedHat:
@giuseppe PTAL

giuseppe:
LGTM

bors r+

rhatdan:
LGTM, Since Buildah is creating the directory, it should create it with the correct ownership

Issue #2194, opened 2020-02-28
don't attempt poll on not pollable files


giuseppe:
buildah enters into a tight loop if a not pollable file is used as stdin.

More details here: https://github.com/containers/libpod/pull/5354

giuseppe:
PR here: https://github.com/containers/buildah/pull/2199

PR #2199, opened 2020-03-02
run_linux: fix tight loop if file is not pollable

giuseppe:
do not attempt again to poll the file if it is not pollable and the
the errno is not syscall.EINTR or syscall.EAGAIN.

Closes: https://github.com/containers/buildah/issues/2194

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -1203,6 +1203,13 @@ func runCopyStdio(stdio *sync.WaitGroup, copyPipes bool, stdioPipe [][]int, copy
 	runCopyStdioPassData(copyPipes, stdioPipe, finishCopy, relayMap, relayBuffer, readDesc, writeDesc)
 }
 
+func canRetry(err error) bool {
+	if errno, isErrno := err.(syscall.Errno); isErrno {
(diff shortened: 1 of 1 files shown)


run_linux.go
```diff
@@ -1203,6 +1203,13 @@ func runCopyStdio(stdio *sync.WaitGroup, copyPipes bool, stdioPipe [][]int, copy
 	runCopyStdioPassData(copyPipes, stdioPipe, finishCopy, relayMap, relayBuffer, readDesc, writeDesc)
 }
 
+func canRetry(err error) bool {
+	if errno, isErrno := err.(syscall.Errno); isErrno {
```
rhatdan: Should we use an errors.Cause for this?

giuseppe: the error is coming from `unix.Read` and I've verified locally it works.  I can make it safer with errors.Cause if you'd like

TomSweeneyRedHat:
LGTM
FWIW, I'd lean to the errors.Cause too

rhatdan:
@giuseppe Could you rebase and we can merge.  Bors is not happy and I am not sure why.
bors retry

PR #2219, opened 2020-03-12
Revert changes for AddCaps/DropCaps

rhatdan:
While merging in of the containers.conf fixes, we changed the API for buildah.
This reverts the change so that openshift/builder can better absorb the change.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```


rhatdan:
@nalind PTAL

TomSweeneyRedHat:
@rhatdan looks like you might have some real test errors to work through here.

rhatdan:
@TomSweeneyRedHat @nalind @giuseppe @vrothberg @QiWang19 PTAL

TomSweeneyRedHat:
LGTM

nalind:
Looking at the API update patch in https://github.com/openshift/builder/pull/141, I see that we changed the type of `transientMounts`.  Are we planning to change that back as well?

How are callers of our APIs (including but not limited to openshift/builder) going to be able to affect the settings that we're now pulling out of a configuration file when the process is started?  Settings and defaults for containers/image APIs can be affected by a SystemContext that callers can pass to us; do we have something like that for affecting containers/common's settings and defaults?

rhatdan:
I think I want to keep the transient mounts change.  Is this a big deal for builder?

Containers.conf should be overridden via the command line or in buildah options.  Do you have a specific one that you don't think the caller can change?

nalind:
> I think I want to keep the transient mounts change. Is this a big deal for builder?

It isn't, but I'm not really clear on why we'd revert one but not the other.

> Containers.conf should be overridden via the command line or in buildah options. Do you have a specific one that you don't think the caller can change?

If we switch back to lists of capabilities to add or remove relative to the defaults, the default list of capabilities becomes one.  Not having a way to change which file we consult also means that we'll have to add (and remember to add) an API knob every time the config package adds a setting that it can read from the configuration file.

rhatdan:
At this point, we are not going to do this.  Closing.

Issue #2235, opened 2020-03-19
2216 broke gating tests


edsantiago:
Gating tests are broken on [f31](https://jenkins-continuous-infra.apps.ci.centos.org/blue/organizations/jenkins/fedora-f31-build-pipeline/detail/fedora-f31-build-pipeline/1064/pipeline/#step-247-log-882) and [f32](https://jenkins-continuous-infra.apps.ci.centos.org/blue/organizations/jenkins/fedora-f32-build-pipeline/detail/fedora-f32-build-pipeline/138/pipeline/#step-248-log-888):
```
not ok 63 bud-volume-ownership
...
exec container process (missing dynamic library?) `/usr/bin/test-ownership`: No such file or directory
```
The bud-volume-ownership test is new, added in #2216. @carbonin PTAL.

rhatdan:
Any idea what library it is erroring out on?

edsantiago:
I couldn't figure it out, but I didn't spend much time on it. It's just a shell script running `stat`.

rhatdan:
yes, very strange.

carbonin:
I could change the test so that it's not using a script added to the container. Other tests that pass seem to be doing that, but they all pass for me (and on CI) so I'm not sure how to verify a fix for your environment.

I'll make that PR and maybe you can test that directly @edsantiago ?

edsantiago:
@carbonin this isn't "my" environment, or one I have any control over: this is the universal Fedora gating-test setup, which is essentially a Fedora VM with the `buildah-tests` RPM installed. Could you try running your test on f31 and f32, and seeing if you can reproduce the failure? Then if you can reproduce the failure, you can work toward solving it.

rhatdan:
You could eliminate script altogether by executing
podman run ... stat -c "%U %G"  /vol/subvol
And comparing it to "testuser testgroup"

carbonin:
Yeah, that was the plan @rhatdan 

@edsantiago sure. I'm developing on f30, and it works there, but I can spin up some vms to test on the others.

carbonin:
Opeend https://github.com/containers/buildah/pull/2236

edsantiago:
>I'm developing on f30

This might be an indicator of a real `crun` problem, then; I'm just surprised that no other tests failed (except for one of the usual flakes). I'll try to take a look.

edsantiago:
[...]

PR #2236, opened 2020-03-19
Run stat command directly for volume ownership test

carbonin:
#### What type of PR is this?

/kind failing-test 

This removes the script previously used which may have been
causing #2235

cc @edsantiago @rhatdan

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -608,7 +608,8 @@ load helpers
   run_buildah bud --signature-policy ${TESTSDIR}/policy.json -t ${target} ${TESTSDIR}/bud/volume-ownership
   run_buildah from --quiet --signature-policy ${TESTSDIR}/policy.json ${target}
   cid=$output
-  run_buildah run $cid test-ownership
+  run_buildah run $cid stat -c "%U %G" /vol/subvol
+  expect_output "testuser testgroup"
 }
 
 @test "bud-from-glob" {
--- a/tests/bud/volume-ownership/Dockerfile
+++ b/tests/bud/volume-ownership/Dockerfile
@@ -3,4 +3,8 @@ RUN adduser -D -H testuser && addgroup testgroup
 RUN mkdir -p /vol/subvol
 RUN chown testuser:testgroup /vol/subvol
 VOLUME /vol/subvol
+
+# Run some command after VOLUME to ensure that the volume cache behavior is invoked
+# See https://github.com/containers/buildah/blob/843d15de3e797bd912607d27324d13a9d5c27dfb/imagebuildah/stage_executor.go#L61-L72 and
+# for more details
 RUN touch /test
--- a/tests/bud/volume-ownership/Dockerfile
+++ b/tests/bud/volume-ownership/Dockerfile
@@ -1,5 +1,4 @@
 FROM alpine
-ADD test-ownership /usr/bin/
 RUN adduser -D -H testuser && addgroup testgroup
 RUN mkdir -p /vol/subvol
 RUN chown testuser:testgroup /vol/subvol
--- a/tests/bud/volume-ownership/test-ownership
+++ b/tests/bud/volume-ownership/test-ownership
@@ -1,6 +1,0 @@
-#!/bin/sh
-
-set -e
-
-MY_USER=$(stat -c "%U" /vol/subvol) && [[ "$MY_USER" == "testuser" ]]
-MY_GROUP=$(stat -c "%G" /vol/subvol) && [[ "$MY_GROUP" == "testgroup" ]]


tests/bud/volume-ownership/Dockerfile
```diff
@@ -1,5 +1,4 @@
 FROM alpine
-ADD test-ownership /usr/bin/
 RUN adduser -D -H testuser && addgroup testgroup
 RUN mkdir -p /vol/subvol
 RUN chown testuser:testgroup /vol/subvol
```
edsantiago: Out of curiosity, what is the purpose of the `touch /test` two lines down? (not an addition in this PR; you'll need to click the expando). I'm assuming it's a copy-paste artifact, and unnecessary for purposes of this test. If so, as long as you're making changes here, could you perhaps remove it to prevent future maintainers from spending time wondering about it? (Or if it's real and necessary, perhaps document why?) Thanks!

carbonin: Actually it is necessary. The [original bug](https://github.com/containers/buildah/issues/2202) only showed up when the volume cache was in play. And that only happens if there are additional `RUN` commands after the `VOLUME` command.

I chose that particular command because it obviously shouldn't have changed the permissions of the volume ... but it did.

I can add a commit with a comment in the Dockerfile to try to summarize the reasoning if you think it would be helpful.

tests/bud.bats
```diff
@@ -608,7 +608,8 @@ load helpers
   run_buildah bud --signature-policy ${TESTSDIR}/policy.json -t ${target} ${TESTSDIR}/bud/volume-ownership
   run_buildah from --quiet --signature-policy ${TESTSDIR}/policy.json ${target}
   cid=$output
-  run_buildah run $cid test-ownership
+  run_buildah run $cid stat -c "%U %G" /vol/subvol
```
edsantiago: So, I'm a complete ignoramus about the VOLUME directive, and I could test this (and will, when I get some time this morning), but: how would this behave differently if the Containerfile _did not_ specify `VOLUME`? Is there any way to check that `/vol/subvol` is actually being treated as a volume, and not just a regular subdirectory in the image?

tests/bud/volume-ownership/Dockerfile
edsantiago: Given what you just said, I think a comment is not only helpful but critical: a future maintainer might have no idea of this context, and could easily refactor/optimize away the `touch`. It is imperative to explain the why, especially in non-obvious cases such as this one. Thanks for your quick explanation.

tests/bud.bats
carbonin: tldr; I wasn't able to find a good way.

Volumes are mount points in containers so I figured I could use `mountpoint` to test the directory, but ...
```sh
/ # mount
...
/dev/mapper/luks-5825177c-f5c0-47c1-9865-db72a9bd7528 on /vol/subvol type btrfs (rw,seclabel,nosuid,nodev,relatime,ssd,space_cache,subvolid=257,subvol=/home/ncarboni/.local/share/containers/storage/volumes/b2efdcb7017c80a10b244942da38e4620d3ec5af0de90a9fee020ea44118679c/_data)
...
/ # mountpoint /vol/subvol/
/vol/subvol/ is not a mountpoint
```

:worried: 

I can't seem to get consistent results from `stat -c "%i` or `stat -f` inside the container either. We could parse `mount` or `df`, but ... gross.

edsantiago: Understood. Thanks for having looked into it.

tests/bud/volume-ownership/Dockerfile
carbonin: Added a comment. Not sure what your feelings are about github links in source, but the comment above the `Preserve` function has a bunch of information about this.

tests/bud.bats
rhatdan: @edsantiago The VOLUME directive tells the container engines to allocate a UNNAMED volume on disk and mount it at the /vol/submount directory.  

The problem was podman was creating the new volume with the wrong owner/permissions and then mounting over the /vol/submount.  The directory went from ownership testuser testgroup to root root.

edsantiago: @carbonin, I can confirm: `mountpoint -n`, `-d`, `blkid`, `stat -c %i`, none of those (or others I've forgotten) produce any useful distinction.

edsantiago:
LGTM. @carbonin thank you for your quick action on this.

TomSweeneyRedHat:
LGTM
@carbonin tyvm for jumping on this so quickly.

rhatdan:
bors retry

PR #2265, opened 2020-04-01
bud.bats - cleanup, refactoring

edsantiago:
Fix two issues identified in #2036:

 - the 'gitrepo and branch' test was pulling from a place
   that took four minutes; change it to our own repo,
   suggested by Dan, which takes just a few seconds.

   -- also, remove what I think is an unnecessary dup.
      If buildah can pull from a branch, it can pull
      from master.

 - the httpd tests were really confusing, with lots of
   copy/pasted code differing in only small ways. Refactor
   to make the purpose of each test more apparent, and
   to make it easier to add new ones as needed.

/kind cleanup

```release-note
no
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -482,37 +482,44 @@ load helpers
   test ! -s $root/vol/anothervolfile
 }
 
-@test "bud-http-Dockerfile" {
-  starthttpd ${TESTSDIR}/bud/from-scratch
+# Helper function for several of the tests which pull from http.
+#
+#  Usage:  _test_http  SUBDIRECTORY  URL_PATH  [EXTRA ARGS]
+#
+#     SUBDIRECTORY   is a subdirectory path under the 'buds' subdirectory.
+#                    This will be the argument to starthttpd(), i.e. where
+#                    the httpd will serve files.
+#
+#     URL_PATH       is the path requested by buildah from the http server,
+#                    probably 'Dockerfile' or 'context.tar'
+#
+#     [EXTRA ARGS]   if present, will be passed to buildah on the 'bud'
+#                    command line; it is intended for '-f subdir/Dockerfile'.
+#
+function _test_http() {
+  local testdir=$1; shift;        # in: subdirectory under bud/
+  local urlpath=$1; shift;        # in: path to request from localhost
+
+  starthttpd "${TESTSDIR}/bud/$testdir"
   target=scratch-image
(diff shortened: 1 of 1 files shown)


tests/bud.bats
```diff
@@ -482,37 +482,49 @@ load helpers
...
 
 @test "bud-http-context-dir-with-Dockerfile-post" {
-  # FIXME FIXME FIXME: this is 100% identical to the -pre test above.
-  starthttpd ${TESTSDIR}/bud/http-context-subdir
-  target=scratch-image
-  run_buildah bud  --signature-policy ${TESTSDIR}/policy.json -t ${target} -f context/Dockerfile http://0.0.0.0:${HTTP_SERVER_PORT}/context.tar
-  stophttpd
-  run_buildah from ${target}
+  # FIXME: this is 100% identical to the -pre test above. Why??
+  _test_http http-context-subdir context.tar -f context/Dockerfile
```
TomSweeneyRedHat, line 527: @edsantiago I'm not sure why these are duplicates.  My only guess is they had a variance at one point that was inadvertently removed?

TomSweeneyRedHat:
LGTM would like @nalind to take a look.

tests/bud.bats
edsantiago: I always get so embarrassed when someone points out the obvious, and said "obvious" never occurred to me. That's it.
In the original commit (7615970) the code differed as follows:
```diff
- buildah bud [options] http://etc
+ buildah bud http://etc [options]
```
(hence the `pre` and `post`)

@rhatdan changed this in #493 with the checkin comment "...not handling command options properly" and something about urfave. That made both `pre` and `post` tests identical.

I could spin up a VM and see if the `post` syntax works, or if I can find some way to make it work. Before I do so: @nalind, do you remember the reasoning behind having both `pre` and `post` testing here? Is it worth trying to recreate it?

Thank you @TomSweeneyRedHat for your insight.

rhatdan: I think at one time we supported options after the image name  as well as before, perhaps with urfave.

buildah bud -t dan http://etc
buildah bud http://etc -t dan

This is no longer allowed, I believe

 buildah bud /tmp -t dan
No options (-t) can be specified after the image or container name

So I think we can remove it.

edsantiago: Thank you. Dup removed, and commit message updated.

rhatdan:
Nice work.

PR #2270, opened 2020-04-02
WIP - DO NOT MERGE - Make pull-always always pull

TomSweeneyRedHat:
I've changed the vendor piece of c/storage that will need to change
with this commit.  If @mtrmac is agreeable with this concept, I'll
work up a proper PR for c/storage.

The motivation for this changes is we currently say in our docs that
`--pull-always` always pulls from the registry regardless of what's
locally stored.

That turns out to NOT be the case, we do reuse some of the images/layers
if they are present locally.

This change passes the `--pull-always` flag down to the c/storage code
so it can decide what to do.

### DIFF
--- a/common.go
+++ b/common.go
@@ -30,7 +30,7 @@ const (
 	DOCKER = "docker"
 )
 
-func getCopyOptions(store storage.Store, reportWriter io.Writer, sourceSystemContext *types.SystemContext, destinationSystemContext *types.SystemContext, manifestType string, removeSignatures bool, addSigner string) *cp.Options {
+func getCopyOptions(store storage.Store, reportWriter io.Writer, sourceSystemContext *types.SystemContext, destinationSystemContext *types.SystemContext, manifestType string, removeSignatures bool, addSigner string, pullAlways bool) *cp.Options {
 	sourceCtx := getSystemContext(store, nil, "")
 	if sourceSystemContext != nil {
 		*sourceCtx = *sourceSystemContext
@@ -47,6 +47,7 @@ func getCopyOptions(store storage.Store, reportWriter io.Writer, sourceSystemCon
 		ForceManifestMIMEType: manifestType,
 		RemoveSignatures:      removeSignatures,
 		SignBy:                addSigner,
+		PullAlways:            pullAlways,
 	}
 }
 
--- a/vendor/github.com/containers/image/v5/copy/copy.go
+++ b/vendor/github.com/containers/image/v5/copy/copy.go
@@ -184,6 +184,8 @@ type Options struct {
 	// OciDecryptConfig contains the config that can be used to decrypt an image if it is
 	// encrypted if non-nil. If nil, it does not attempt to decrypt an image.
 	OciDecryptConfig *encconfig.DecryptConfig
+	// PullAlways force the pulling of the image and digests from teh repository
+	PullAlways bool
 }
 
 // validateImageListSelection returns an error if the passed-in value is not one that we recognize as a valid ImageListSelection value
@@ -632,7 +634,7 @@ func (c *copier) copyOneImage(ctx context.Context, policyContext *signature.Poli
 	// If encrypted and decryption keys provided, we should try to decrypt
 	ic.diffIDsAreNeeded = ic.diffIDsAreNeeded || (isEncrypted(src) && ic.ociDecryptConfig != nil) || ic.ociEncryptConfig != nil
 
-	if err := ic.copyLayers(ctx); err != nil {
+	if err := ic.copyLayers(ctx, options); err != nil {
 		return nil, "", "", err
 	}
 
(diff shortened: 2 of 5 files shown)


TomSweeneyRedHat:
I've some vendor shenanigans going on.  However, if folks could review, especially @mtrmac and @redhatdan, in particular for the change in c/storage/copy.go, I'd appreciate it.

If the approach is OK, I'll work up a change in c/storage and then will vendor it into here before merging.

common.go
```diff
@@ -30,7 +30,7 @@ const (
 	DOCKER = "docker"
 )
 
-func getCopyOptions(store storage.Store, reportWriter io.Writer, sourceSystemContext *types.SystemContext, destinationSystemContext *types.SystemContext, manifestType string, removeSignatures bool, addSigner string) *cp.Options {
+func getCopyOptions(store storage.Store, reportWriter io.Writer, sourceSystemContext *types.SystemContext, destinationSystemContext *types.SystemContext, manifestType string, removeSignatures bool, addSigner string, pullAlways bool) *cp.Options {
```
rhatdan, line 33: This function has gotten pretty ridiculous.  I don't see much value in it as it is only populating a structure that you have to pass in all of the options.  Perhaps two functions to get sourceCtx and DestinationCtx?

vendor/github.com/containers/image/v5/copy/copy.go
```diff
@@ -1039,11 +1041,10 @@ type diffIDResult struct {
 
 // copyLayer copies a layer with srcInfo (with known Digest and Annotations and possibly known Size) in src to dest, perhaps compressing it if canCompress,
 // and returns a complete blobInfo of the copied layer, and a value for LayerDiffIDs if diffIDIsNeeded
-func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, toEncrypt bool, pool *mpb.Progress) (types.BlobInfo, digest.Digest, error) {
+func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, toEncrypt bool, pool *mpb.Progress, options *Options) (types.BlobInfo, digest.Digest, error) {
 	cachedDiffID := ic.c.blobInfoCache.UncompressedDigest(srcInfo.Digest) // May be ""
 	// Diffs are needed if we are encrypting an image or trying to decrypt an image
-	diffIDIsNeeded := ic.diffIDsAreNeeded && cachedDiffID == "" || toEncrypt || (isOciEncrypted(srcInfo.MediaType) && ic.ociDecryptConfig != nil)
-
+	diffIDIsNeeded := options.PullAlways || ic.diffIDsAreNeeded && cachedDiffID == "" || toEncrypt || (isOciEncrypted(srcInfo.MediaType) && ic.ociDecryptConfig != nil)
```
mtrmac, line 1047: I can’t see how `diffIDIsNeeded` is related to “copy always”; if it is, there should be a comment that explains that.

Or, just guessing, if the goal is only to turn off the `TryReusingBlob` call, make that a separate condition (so that this does not cause an unnecessary costly DiffID computation).

vendor/github.com/containers/image/v5/copy/copy.go
```diff
@@ -1039,11 +1041,10 @@ type diffIDResult struct {
 
 // copyLayer copies a layer with srcInfo (with known Digest and Annotations and possibly known Size) in src to dest, perhaps compressing it if canCompress,
 // and returns a complete blobInfo of the copied layer, and a value for LayerDiffIDs if diffIDIsNeeded
-func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, toEncrypt bool, pool *mpb.Progress) (types.BlobInfo, digest.Digest, error) {
+func (ic *imageCopier) copyLayer(ctx context.Context, srcInfo types.BlobInfo, toEncrypt bool, pool *mpb.Progress, options *Options) (types.BlobInfo, digest.Digest, error) {
```
mtrmac, line 1044: If `pullAlways` needs propagating all the way down here, please make it a member of `copier` instead of an extra parameter that most functions in the stack don’t need.

TomSweeneyRedHat:
What’s the end-user-relevant outcome of the difference? Sure, the difference is end-user-observable (is the pull fast/slow), but why does the user care?

@mtrmac the main reason for me doing the change is our doc says "always pulls" for `--pull-always` and the code is not doing that.  I'm just fore seeing a user report that we're not pulling the image, but using the local bits if present.  Perhaps I'm over thinking.  

Secondarily, this was a weekend "fun" project for me to see what I could do with it.

My mind I'm happy to close this, but if we do, I think our docs/man pages needs a little touch up to explain a bit more how we're doing the pull-always.  Maybe later, if we get a user ping, we can add a "--pull-refresh" or some such that would ignore all the local layers/images and just pull from the repo.

Thoughts?

mtrmac:
> My mind I'm happy to close this, but if we do, I think our docs/man pages needs a little touch up to explain a bit more how we're doing the pull-always.

Sure, something to the effect “contact the registry to ensure exactly the same image is present locally” might be useful.

>  Maybe later, if we get a user ping, we can add a "--pull-refresh" or some such that would ignore all the local layers/images and just pull from the repo.

I’m not at all sure that would work:
- We can’t just delete local layers and pull them again with the same IDs, because some of the layers might be shared with other images that we must not break.
- If we pulled the same local layers with different IDs, that’s probably possible in principle (unless some code assumes that layer ID == DiffID, not sure), but some parts of Podman, like history/tree(/prune?) assume parent’s layers are a prefix of child’s layers, and it would be very unintuitive for a `--pull-refresh` to change output/behavior of such commands.

rhatdan:
I am going to close this as I don't believe it is relevant any longer..

PR #2274, opened 2020-04-03
Downgrade siruspen/logrus from 1.4.2

umohnani8:
The upgrade of logrus forces to update the dependency in go module
consumers as well.

The new locking in logrus 1.5.0 introduces regressions in conjunction to
deadlocks when using custom hooks, which is the case in CRI-O for
example.
Having it point to v1.5.0 here is forcing cri-o to upgrade it as well
when we update c/buildah to v1.14.6.

Let's downgrade logrus till sirupsen/logrus#1122
is fixed.


umohnani8:
@rhatdan @mrunalp @TomSweeneyRedHat PTAL

TomSweeneyRedHat:
I'm not against a new release, but please remove that part of the PR from this one.  We can do a separate PR afterwards.

TomSweeneyRedHat:
I decided to try and be a nice guy for once and do the vendoring part of this, and it bit me.  
I did:  
```
# go get github.com/containers/common@v0.8.1 github.com/containers/image/v5@v5.4.1 github.com/containers/storage@v1.18.2 github.com/sirupsen/logrus@v1.4.2 github.com/pkg/errors@v0.9.1
```
and that works as does the `make vendor-in-container` step, but make falls down:

```
# make
GO111MODULE=on go build -mod=vendor -ldflags '-X main.GitCommit=31a01b4a -X main.buildInfo=1586031226 -X main.cniVersion=v0.7.2-0.20190904153231-83439463f784'  -o buildah -tags "seccomp selinux    " ./cmd/buildah
# k8s.io/apimachinery/pkg/util/errors
vendor/k8s.io/apimachinery/pkg/util/errors/errors.go:99:10: undefined: errors.Is
make: *** [Makefile:50: binary] Error 2
```

It looks like k8s has it's own `Is()` and it's somehow messing with the new `Is()` function in `github.com/pkg/error@v0.9.1`.  I tried downgrading to 0.9.0, but there are dependencies for 0.9.1 in c/image, c/common and others.

@umohnani8 did you run into this?  @vrothberg, aka "Mr. (Go Module) Wizard", any thoughts?

vrothberg:
@TomSweeneyRedHat, I can't reproduce with your steps. Which version of go are you running on your machine?

umohnani8:
@TomSweeneyRedHat yup, I see the same failure too. It works when you upgrade go to 1.13, so that is what I was trying to do in the `update test to go 1.13` commit, but haven't been fully successful so far.

umohnani8:
@vrothberg go 1.12 will give you that error, but bumping to 1.13 will not.

vrothberg:
We need to continue supporting g1.12 for F30 for about two months. We just hit the very same error in libpod.

Issue #2262, opened 2020-03-31
Report: Error during unshare(CLONE_NEWUSER): Operation not permitted in the non-privileged container of k8s


zhangtbj:
Hi all,

I need to run the buildah to build my source code on a shared kube cluster.

There are serval security policise and cannot run the container with privileged.

So when I running the buildah in the non-privileged container of k8s.

It reports the error:
```
...
time="2020-03-31T13:30:14Z" level=warning msg="The cgroups manager is set to systemd but there is no systemd user session available"
time="2020-03-31T13:30:14Z" level=warning msg="For using systemd, you may need to login using an user session"
time="2020-03-31T13:30:14Z" level=warning msg="Alternatively, you can enable lingering with: `loginctl enable-linger 1000` (possibly as root)"
time="2020-03-31T13:30:14Z" level=warning msg="Falling back to --cgroup-manager=cgroupfs"
Error during unshare(CLONE_NEWUSER): Operation not permitted
level=error msg="error parsing PID \"\": strconv.Atoi: parsing \"\": invalid syntax"
level=error msg="(unable to determine exit status)"
2020/03/31 13:30:14 Skipping step because a previous step failed
2020/03/31 13:30:15 Skipping step because a previous step failed
{"level":"info","ts":1585661399.569302,"caller":"creds-init/main.go:44","msg":"Credentials initialized."}
```

I also tried this `--isolation=chroot` parameter for `buildah bud` command, but doesn't work.

And I asked open all allowedCapabilities also doesn't work:
```
spec:
  allowPrivilegeEscalation: true
  allowedCapabilities:
  - '*'
```

I run the buildah by using Tekton task:
https://github.com/tektoncd/catalog/blob/master/buildah/buildah.yaml#L43

I saw there are many people had this problem before, but I didn't see the final solution: https://github.com/containers/buildah/issues/1011

I think buildah should work like other builder like kaniko that can build without privileged.

Can someone help me?

Thanks a lot!

rhatdan:
Do you know if this running with docker backend or cri-o?  Do you know if seccomp.json rules are being applied.
What OS is the server running with?

zhangtbj:
Hi @rhatdan ,

We are running buildah as Tekton task on the Kubernetes which based on the `containerd`

It should be the `ubuntu` OS and I saw we set the seccomp setting in the kube PSP annotations like below.

```
kind: PodSecurityPolicy
metadata:
  annotations:
[...]

PR #2284, opened 2020-04-06
Remove warning for systemd inside of container

rhatdan:
Change the default cgroup_manager to cgroupfs to change the warning
inside of the containers on systemd not being supported.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/contrib/buildahimage/stable/Dockerfile
+++ b/contrib/buildahimage/stable/Dockerfile
@@ -11,7 +11,9 @@ FROM fedora:latest
 # Don't include container-selinux and remove
 # directories used by yum that are just taking
 # up space.
-RUN useradd build; yum -y update; yum -y reinstall shadow-utils; yum -y install buildah fuse-overlayfs --exclude container-selinux; rm -rf /var/cache /var/log/dnf* /var/log/yum.*
+RUN useradd build; yum -y update; yum -y reinstall shadow-utils; yum -y install buildah fuse-overlayfs --exclude container-selinux; rm -rf /var/cache /var/log/dnf* /var/log/yum.*;
+
+ADD https://raw.githubusercontent.com/containers/buildah/master/contrib/buildahimage/stable/containers.conf /etc/containers/
 
 # Adjust storage.conf to enable Fuse storage.
 RUN sed -i -e 's|^#mount_program|mount_program|g' -e '/additionalimage.*/a "/var/lib/shared",' -e 's|^mountopt[[:space:]]*=.*$|mountopt = "nodev,fsync=0"|g' /etc/containers/storage.conf
--- a/contrib/buildahimage/stable/containers.conf
+++ b/contrib/buildahimage/stable/containers.conf
@@ -1,0 +1,2 @@
+[containers]
+cgroup_manager = "cgroupfs"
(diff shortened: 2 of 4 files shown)


rhatdan:
@TomSweeneyRedHat @nalind @vrothberg @QiWang19 @mheon PTAL

rhatdan:
Warnings being reported here https://github.com/containers/buildah/issues/2262

contrib/buildahimage/stable/Containerfile
```diff
@@ -11,7 +11,9 @@ FROM fedora:latest
 # Don't include container-selinux and remove
 # directories used by yum that are just taking
 # up space.
-RUN useradd build; yum -y update; yum -y reinstall shadow-utils; yum -y install buildah fuse-overlayfs --exclude container-selinux; rm -rf /var/cache /var/log/dnf* /var/log/yum.*
+RUN useradd build; yum -y update; yum -y reinstall shadow-utils; yum -y install buildah fuse-overlayfs --exclude container-selinux; rm -rf /var/cache /var/log/dnf* /var/log/yum.*;
+
+COPY containers.conf /etc/containers/
```
TomSweeneyRedHat, line 16: These are called from quay.io to build our images when the PR merge.  I'm not sure that we can use a file there that isn't a "Dockerfile".  In addition, I don't think the COPY will work unless we fully specify the file location.  I've not tried this, but could we do something like:

```
COPY https://github.com/containers/buildah/tree/master/contrib/buildahimage/stable/containers.conf /etc/containers
```

rhatdan: Could we check to see if this works or not, and then do this in a separate PR if it blows up.  I hate to muck up the Dockerfiles files like this, if we don't have to.

TomSweeneyRedHat: COPY won't work, but an ADD like this does.

```
ADD https://raw.githubusercontent.com/containers/buildah/master/contrib/buildahimage/stable/containers.conf /etc/containers/
```

rhatdan:
@TomSweeneyRedHat PTAL

contrib/buildahimage/stable/Dockerfile
```diff
@@ -11,7 +11,9 @@ FROM fedora:latest
 # Don't include container-selinux and remove
 # directories used by yum that are just taking
 # up space.
-RUN useradd build; yum -y update; yum -y reinstall shadow-utils; yum -y install buildah fuse-overlayfs --exclude container-selinux; rm -rf /var/cache /var/log/dnf* /var/log/yum.*
+RUN useradd build; yum -y update; yum -y reinstall shadow-utils; yum -y install buildah fuse-overlayfs --exclude container-selinux; rm -rf /var/cache /var/log/dnf* /var/log/yum.*;
```
TomSweeneyRedHat, line 14: I thought the ending `;` might be problematic here, but just checked and it works fine.

TomSweeneyRedHat:
We need to also update the Dockerfile in stablebyhand, but these are all fine. 
LGTM

TomSweeneyRedHat:
bots retry

TomSweeneyRedHat:
bors retry

PR #2287, opened 2020-04-07
make vendor: run `tidy` after `vendor`

vrothberg:
Run `go mod tidy` after `go mod vendor` to properly clean up the
dependencies.

Signed-off-by: Valentin Rothberg <rothberg@redhat.com>

#### What type of PR is this?

 /kind other

#### What this PR does / why we need it:

Cleans up the dependencies.

#### How to verify it

Run `go mod tidy` locally and notice that nothing changes anymore.

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

Do we really need this template in Buildah? We don't seem to use/enforce it much.

PR #2290, opened 2020-04-08
Fix (make vendor)

mtrmac:
#### What type of PR is this?
/kind bug

`(export a=b command args)` does not run `(command args)` with `a=b`, it sets `$a` to `b`, and marks variables `$a` `$command` `$args` as exported, i.e. `(command args)` is not run.

So, before #2287 we were not actually running `(go mod tidy)`, and now we are not running `(go mod vendor)`.

Fix that, and re-vendor.

Notably, vendored c/image and several other dependencies currently don’t correspond to `go.mod`.

#### How to verify it
Manually run (go mod …) and compare that with the results of running (make vendor)

None

None


TomSweeneyRedHat:
@mtrmac thanks!
Can you sign the PR and repush please.
And Yikes, we really needed that tidy!

mtrmac:
The three test failures in https://cirrus-ci.com/task/4722459664449536 are all on merge commits, including a two-months-old d654a795 which passed in the previous run. I’m not sure what is going on.

rhatdan:
Is there a validate EPOCH we need to update?

rhatdan:
tests/validate/git-validation.sh

mtrmac:
> Is there a validate EPOCH we need to update?

That could hide it, sure, but really, what is going on here?

All the three failing commits succeed in https://cirrus-ci.com/task/5357058849505280?command=validate_test (the previous PR), and fail here.

The git-validation version is the same. The git RPM version is the same.

rhatdan:
Newer go compiler?

mtrmac:
Both 1.12.17-1.fc30 .

mtrmac:
OK, I’m completely out of ideas. For now, as a desperate last guess, I’m dropping the `make -C test/tools vendor` commit, which AFAICS should make absolutely no difference.

mtrmac:
Still fails, but now with two different merge commits.

I’m re-adding the tools commit, but otherwise giving up for today.

PR #2291, opened 2020-04-08
Fix (make vendor) 

rhatdan:
What type of PR is this?

/kind bug
What this PR does / why we need it:

(export a=b command args) does not run (command args) with a=b, it sets $a to b, and marks variables $a $command $args as exported, i.e. (command args) is not run.

So, before #2287 we were not actually running (go mod tidy), and now we are not running (go mod vendor).

Fix that, and re-vendor.

Notably, vendored c/image and several other dependencies currently don’t correspond to go.mod.
How to verify it

Manually run (go mod …) and compare that with the results of running (make vendor)
Which issue(s) this PR fixes:

None
Special notes for your reviewer:
Does this PR introduce a user-facing change?

None


rhatdan:
Attempting to replace https://github.com/containers/buildah/pull/2290

TomSweeneyRedHat:
So this is green and #2290 is not.  I see one vendored file difference, and go.sum has slightly different entries.  Do you know what you did that @mtrmac didn't?

LGTM

rhatdan:
bors r+
I updated the EPOCH to not check as many old PRs

PR #2297, opened 2020-04-10
Fix error info about get signatures for containerImageSource

zvier:
Signed-off-by: zvier <liuzekun0524@163.com>

#### What type of PR is this?

> /kind cleanup

For a containerImageSource instance, if call GetSignatures method failed, it returns the error with the message "containerImageSource does not support manifest lists". If the message is "containerImageSource does not support signature lists", I think it shoud be more accurate.

None

None

```release-note

```

### DIFF
--- a/image.go
+++ b/image.go
@@ -587,7 +587,7 @@ func (i *containerImageSource) Reference() types.ImageReference {
 
 func (i *containerImageSource) GetSignatures(ctx context.Context, instanceDigest *digest.Digest) ([][]byte, error) {
 	if instanceDigest != nil {
-		return nil, errors.Errorf("containerImageSource does not support manifest lists")
+		return nil, errors.Errorf("containerImageSource does not support signature lists")
 	}
 	return nil, nil
 }


rhatdan:
LGTM
Thanks @zvier 
@TomSweeneyRedHat @nalind @vrothberg @giuseppe @QiWang19 PTAL

giuseppe:
AFAICS the issue is related to not supporting `manifest lists`, so the original error seems correct

rhatdan:
Well the function is about Get Signatures.

rhatdan:
How about 		
return nil, errors.Errorf("containerImageSource manifest lists do not support signatures")

giuseppe:
I wonder if we should we just drop the error message

rhatdan:
@nalind WDYT?

nalind:
I don't have a strong opinion here.  Does the code get called with non-`nil` `instanceDigest` values?

rhatdan:
Just removing the checks in #2346

PR #2334, opened 2020-04-27
Don't force tests to use runc

rhatdan:
We should default to what buildah info reports else use runc or crun.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -5,6 +5,7 @@ IMGTYPE_BINARY=${IMGTYPE_BINARY:-$(dirname ${BASH_SOURCE})/../imgtype}
 TESTSDIR=${TESTSDIR:-$(dirname ${BASH_SOURCE})}
 STORAGE_DRIVER=${STORAGE_DRIVER:-vfs}
 PATH=$(dirname ${BASH_SOURCE})/..:${PATH}
+OCI=$(${BUILDAH_BINARY} info --format '{{.host.OCIRuntime}}' || command -v runc || command -v crun)
 
 # Default timeout for a buildah command.
 BUILDAH_TIMEOUT=${BUILDAH_TIMEOUT:-300}
@@ -302,14 +303,11 @@ function skip_if_rootless() {
 #  skip_if_no_runtime  #  'buildah run' can't work without a runtime
 ########################
 function skip_if_no_runtime() {
-    # FIXME: if #1964 gets fixed, use 'buildah info' to determine runtime
-    # FIXME: right now we just rely on runc
-    runtime=runc
-    if type -p $runtime &> /dev/null; then
+    if type -p "${OCI}" &> /dev/null; then
         return
     fi
 
-    skip "runtime '$runtime' not found"
+    skip "runtime \"$OCI\" not found"
 }
 
 ##################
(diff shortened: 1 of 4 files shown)


rhatdan:
@edsantiago @TomSweeneyRedHat @jnovy @nalind @vrothberg PTAL

tests/helpers.bash
```diff
@@ -302,10 +303,7 @@ function skip_if_rootless() {
 #  skip_if_no_runtime  #  'buildah run' can't work without a runtime
 ########################
 function skip_if_no_runtime() {
-    # FIXME: if #1964 gets fixed, use 'buildah info' to determine runtime
-    # FIXME: right now we just rely on runc
-    runtime=runc
-    if type -p $runtime &> /dev/null; then
+    if type -p ${OCI} &> /dev/null; then
         return
     fi
```
edsantiago, line 309: Next line (not visible in github diff) needs `$runtime` changed to `$OCI`

tests/helpers.bash
```diff
@@ -5,6 +5,7 @@ IMGTYPE_BINARY=${IMGTYPE_BINARY:-$(dirname ${BASH_SOURCE})/../imgtype}
 TESTSDIR=${TESTSDIR:-$(dirname ${BASH_SOURCE})}
 STORAGE_DRIVER=${STORAGE_DRIVER:-vfs}
 PATH=$(dirname ${BASH_SOURCE})/..:${PATH}
+OCI=$(${BUILDAH_BINARY} info --format '{{.host.OCIRuntime}}' || command -v runc || command -v crun)
```
edsantiago, line 8: What benefit do the two `command -v`s add? Shouldn't we trust what buildah itself reports?

tests/helpers.bash
```diff
@@ -302,10 +303,7 @@ function skip_if_rootless() {
 #  skip_if_no_runtime  #  'buildah run' can't work without a runtime
 ########################
 function skip_if_no_runtime() {
-    # FIXME: if #1964 gets fixed, use 'buildah info' to determine runtime
-    # FIXME: right now we just rely on runc
-    runtime=runc
-    if type -p $runtime &> /dev/null; then
+    if type -p ${OCI} &> /dev/null; then
```
edsantiago, line 306: Related to the above (re: `command -v`): this should include `-n "${OCI}" && ...`

TomSweeneyRedHat:
Nice changes overall once @edsantiago 's comments are addressed

tests/helpers.bash
rhatdan: There were tests that would pass if no OCI Runtime is defined.  I was not sure what happens in that case, if you don't have crun or runc on a system.  Then test would be cancelled.  This way we are more likely to get something if buildah info fails to find one.

edsantiago:
LGTM. I'm still not quite sure how the `OCI` fallback will work in practice, but it's worth a try: this is much better than what we have right now.

rhatdan:
@TomSweeneyRedHat or @edsantiago Can you merge this

TomSweeneyRedHat:
I've made @edsantiago a valid reviewer in bors, so hopefully it will take his bors commands going forward.  I'll hit it up too now just to make sure.

PR #2370, opened 2020-05-22
remove dependency on openshift struct

baude:
use of github.com/openshift/api/config/v1 is costly in size, something around 10MB.  removal of this dependancy and using a locally defined struct took the buildah executable size from around 38MB to around 27MB.  This also hits libpod and allows it also shed that size as well.

Signed-off-by: Brent Baude <bbaude@redhat.com>


commit.go
```diff
@@ -181,7 +180,11 @@ func checkRegistrySourcesAllows(forWhat string, dest types.ImageReference) error
 	}
 
 	if registrySources, ok := os.LookupEnv("BUILD_REGISTRY_SOURCES"); ok && len(registrySources) > 0 {
-		var sources configv1.RegistrySources
+		var sources struct {
```
mheon, line 183: Should at least mention where this came from

TomSweeneyRedHat:
Changes LGTM, tests aren't happy at the moment, but I don't think it's related.  Looks like Friday afternoon CI going on vacation.

My only small concern is these definitions diverging at some point, but I think there's an extremely low chance of that happening.

TomSweeneyRedHat:
bors retest

rhatdan:
Incredible.  Great work @baude 
bors r+
@TomSweeneyRedHat Lets cut a minor release as soon as this passes.

TomSweeneyRedHat:
@rhatdan if possible, I'd like to get my imagebuilder fix in too.

vrothberg:
Sweet! Nice work, @baude :+1:

PR #2432, opened 2020-07-02
Switch scripts to use containers.conf

rhatdan:
Podman no longer uses libpod.conf and has switched to containers.conf.
This change will cause podman to use crun by default, by creating a
/etc/containers/containers.conf file

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/contrib/cirrus/setup.sh
+++ b/contrib/cirrus/setup.sh
@@ -33,7 +33,10 @@ case "$OS_RELEASE_ID" in
         fi
 
         warn "Hard-coding podman to use crun"
-        sed -i -r -e 's/^runtime = "runc"/runtime = "crun"/' /usr/share/containers/libpod.conf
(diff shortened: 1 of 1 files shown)


rhatdan:
@edsantiago @TomSweeneyRedHat PTAL

rhatdan:
@cevich PTAL

TomSweeneyRedHat:
LGTM
pretty slick script change

rhatdan:
@edsantiago @cevich PTAL, We need this in to fix the libpod.conf issues.

edsantiago:
Can we assume that `containers-common` is a hardcoded requirement? (i.e. that `/etc/containers` will exist)? I'm guessing we can, so LGTM.

PR #2435, opened 2020-07-02
Make imagebuildah.BuildOptions.Architecture/OS optional

nalind:
#### What type of PR is this?

/kind bug

When committing an image, the Architecture and OS fields in the imagebuildah.BuildOptions structure were applied unconditionally, _even when they were empty,_ which would overwrite the usually-non-empty values supplied by the base image.

#### How to verify it

None

```
None
```


rhatdan:
LGTM

nalind:
bors retry

PR #2436, opened 2020-07-02
Make imagebuildah.BuildOptions.Jobs optional

nalind:
#### What type of PR is this?

/kind api-change

When building an image, an unset value for the `Jobs` field in the imagebuildah.BuildOptions structure was treated as an indication that every stage should be built in parallel, despite the API documentation stating that an unset value would be interpreted as 1.

#### How to verify it

```
None

```


rhatdan:
LGTM

PR #2438, opened 2020-07-02
imagebuildah: fix inheriting multi-stage builds

nalind:
#### What type of PR is this?

/kind bug

When building stages in parallel, stages which are based on images built by earlier stages don't wait for those earlier stages to complete before attempting to start, causing them to attempt to pull an image with the same name as the previous stage's nickname, leading to unpredictable results.

#### How to verify it
Build using a Dockerfile that looks something like `tests/conformance/testdata/Dockerfile.reusebase` with a `--jobs` value other than 1.

```release-note
None
```


rhatdan:
LGTM

rhatdan:
bors retry

PR #2443, opened 2020-07-03
Allow "readonly" as alias to "ro" in mount options

mariushoch:
Apparently docker supports this as well, see [Use a read-only volume](https://docs.docker.com/storage/volumes/#use-a-read-only-volume) in the docker docs.

#### What type of PR is this?

/kind bug

To be fully compatible with docker in this regard.

#### How to verify it

With a mount option like `type=bind,source=/var/source,target=/src,readonly`.

```release-note
None
```

### DIFF
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -343,6 +343,9 @@ func GetBindMount(args []string) (specs.Mount, error) {
 			// TODO: detect duplication of these options.
 			// (Is this necessary?)
 			newMount.Options = append(newMount.Options, kv[0])
+		case "readonly":
+			// Alias for "ro"
+			newMount.Options = append(newMount.Options, "ro")
 		case "shared", "rshared", "private", "rprivate", "slave", "rslave", "Z", "z":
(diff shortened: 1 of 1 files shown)


giuseppe:
bors retry

mariushoch:
I just changed the behavior in "ValidateVolumeOpts" as I realized that right now it does not allow multiple/ redundant entries for `ro` (and others), so why should it allow both `readonly` and `ro`?

rhatdan:
LGTM

rhatdan:
bors r+
Thanks @mariushoch

rhatdan:
bors retry

rhatdan:
@mariushoch Could you rebase and repush this PR,  we had a fix added for CI to fix a change in the podman packages.

mariushoch:
Rebased and reverted my changes to `ValidateVolumeOpts`, it should not allow "readonly" as it is supposed to get normalized volume options only (and storage code wont understand "readonly").

rhatdan:
@mariushoch Could you rebase and repush.

PR #2448, opened 2020-07-06
Move to ubuntu-20 and 19 testing to match libpod

rhatdan:

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -28,12 +28,15 @@ env:
     # GCE project where images live
     IMAGE_PROJECT: "libpod-218412"
     # See https://github.com/containers/libpod/blob/master/contrib/cirrus/README.md#test_build_cache_images_task-task
-    _BUILT_IMAGE_SUFFIX: "libpod-6224667180531712"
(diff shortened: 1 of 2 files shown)


cevich:
@rhatdan this LGTM.  Comparing the test fails here to other PRs it seems like something is breaking with the local registry server, part way through testing.  It doesn't appear to me to be OS-specific, but :man_shrugging: I may not be the best one to debug in this case.  Maybe need Ed's or Nalin's input, since they know more about the tests themselves (I just work here) :grinning:   I'll come back in an hour or so and peep some more, in case something jumps at me :confused:

rhatdan:
@edsantiago @nalind Could you take a look to see what is going on?

edsantiago:
The f31 failure looks like one of the usual "docker.io refuses to talk to us" flakes.

Ubuntu failures are both `concurrent map writes` which @nalind acknowledged yesterday in IRC:

> there are commits in #2435/#2436/#2438 that should resolve it, but those PRs are still failing elsewhere

PR #2454, opened 2020-07-09
Makefile: use customizable env vars

lsm5:

I thought this was done before but apparently not.

@rhatdan @nalind @TomSweeneyRedHat @vrothberg PTAL

#### What type of PR is this?

/kind other

will help me reuse Makefile targets in packaging

#### How to verify it

upstream doesn't really need to care

None

None

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,15 +1,15 @@
 export GOPROXY=https://proxy.golang.org
 
-SELINUXTAG := $(shell ./selinux_tag.sh)
-APPARMORTAG := $(shell hack/apparmor_tag.sh)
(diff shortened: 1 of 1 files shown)


lsm5:
backport at https://github.com/containers/buildah/pull/2455

nalind:
`make` should allow these values to be overridden by passing the desired values as parameters to make, e.g. `make PREFIX=/usr`.  There are many variable assignments in the Makefile that this patch doesn't touch, and if we're going to tweak how some are set, why not the rest?

lsm5:
@nalind ah you're right, my bad. Closing ...

PR #2465, opened 2020-07-14
Deduplicate environment variables

nalind:
#### What type of PR is this?

> /kind cleanup

When combining lists of environment variables read from base images with defaults supplied from our own configuration, ensure that the resulting environment we produce only contains one value for any given variable.  While adding variables to a runtime spec using `github.com/opencontainers/runtime-tools/generate.Generator.AddProcessEnv()` ensures that later values in the list override values that occur earlier, we shouldn't be depending on that.

None

Related to https://github.com/openshift/imagebuilder/pull/169.

```
None
```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -1967,7 +1967,7 @@ func (b *Builder) configureEnvironment(g *generate.Generator, options RunOptions
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


nalind:
bors retry

PR #2466, opened 2020-07-14
imagebuildah: return the right stage's image as the "final" image

nalind:
#### What type of PR is this?

> /kind bug

When building multiple stages, we weren't ensuring that the image reference we returned to our caller was a reference of the final stage we built.  If a different stage actually finished before the "final" one, its ID and reference would overwrite the values that would otherwise have been returned.

When checking if a name specified for a FROM instruction, or a source for a COPY instruction, was a stage for which it needed to wait, StageExecutor was checking the Executor's stages list, to which stages are added only as they're started.  This could cause it to mistakenly write off a name if the corresponding stage hadn't yet been started, and assume that the named stage was actually an image.

#### How to verify it

This would manifest as a flake in multi-stage builds, particularly with `--jobs` values other than 1.

```
None
```

### DIFF
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -531,19 +531,19 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 			go func() {
 				defer b.stagesSemaphore.Release(1)
 				defer wg.Done()
-				imageID, ref, err = b.buildStage(ctx, cleanupStages, stages, index)
-				if err != nil {
+				stageID, stageRef, stageErr := b.buildStage(ctx, cleanupStages, stages, index)
+				if stageErr != nil {
 					ch <- Result{
 						Index: index,
-						Error: err,
+						Error: stageErr,
 					}
 					return
 				}
 
 				ch <- Result{
 					Index:   index,
-					ImageID: imageID,
-					Ref:     ref,
+					ImageID: stageID,
+					Ref:     stageRef,
 					Error:   nil,
 				}
 			}()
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

rhatdan:
@TomSweeneyRedHat PTAL

imagebuildah/executor.go
```diff
@@ -289,18 +290,15 @@ func (b *Executor) resolveNameToImageRef(output string) (types.ImageReference, e
...
 
+// waitForStage waits for an entry to be added to terminatedStage indicating
+// that the specified stage has finished.  If there is no stage by that name,
+// then it will wait forever, so the caller should ensure that there really is
+// a stage with the specified name before calling it.
 func (b *Executor) waitForStage(ctx context.Context, name string) error {
-	stage := b.stages[name]
-	if stage == nil {
-		return errors.Errorf("unknown stage %q", name)
-	}
```
TomSweeneyRedHat, line 296: Do we really want to remove this check?  With the code as is, I know it should always pass, but I think it's a good fail-safe for future calls no?

I'm always very leery about potential never-ending loops like this.  I've been trying to think of an escape hatch, but can't at the moment that makes sense.

nalind: Okay, moved some things around.

rhatdan:
LGTM Again

imagebuildah/executor.go
TomSweeneyRedHat: Thx!

TomSweeneyRedHat:
LGTM
thanks for the rework @nalind

Issue #2459, opened 2020-07-12
buildah run/bud --isolation=chroot $c <cmd> fails with single uidmap for unprivileged user


mrinaldhillon:
buildah run/bud **fails with chroot isolation** if user is not configured in subuid/subgid map files. It also ***fails with oci isolation and crun runtime**. However **succeeds with rootless isolation with crun runtime**. Is this expected ? 

Fail: isolation=chroot
Fail: isolation=oci runtime=crun
Success: isolation=rootless runtime=crun

**Steps to reproduce the issue:**
1. remove user entries from /etc/subuid /etc/subgid
2. c=$(buildah from debian:10-slim)
3. buildah run --isolation=chroot $c true 

**Describe the results you received:**
```
buildah run --isolation=chroot $c true
error clearing supplemental groups list: operation not permittedsubprocess exited with status 1
                                                                                               exit status 1
```

```
buildah run --isolation=oci --runtime=crun $c true
2020-07-12T03:34:45.000749013Z: mount `devpts` to '/dev/pts': Invalid argument
error running container: error creating container for [/bin/true]: : exit status 1
error while running runtime: exit status 1
ERRO exit status 1
```

```
buildah run --isolation=rootless --runtime=crun $c true
echo $?
0
```

**Describe the results you expected:**
 buildah run --isolation=chroot $c true
echo $?
0 

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.15.0-1.fc32.x86_64
```

**Output of `buildah version`:**

```
Version:         1.15.0
Go Version:      go1.14.3
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        0.4.0
libcni Version:
image Version:   5.5.1
Git Commit:
Built:           Wed Dec 31 19:00:00 1969
OS/Arch:         linux/amd64
```

**Output of `cat /etc/*release`:**

```
Fedora release 32 (Thirty Two)
NAME=Fedora
VERSION="32 (Workstation Edition)"
ID=fedora
VERSION_ID=32
VERSION_CODENAME=""
PLATFORM_ID="platform:f32"
PRETTY_NAME="Fedora 32 (Workstation Edition)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:32"
HOME_URL="https://fedoraproject.org/"
[...]

PR #2468, opened 2020-07-14
chroot: allow to run with a single mapping

giuseppe:
#### What type of PR is this?

/kind bug

allow to use chroot with a single user available.

#### How to verify it

follow the reproducer here: https://github.com/containers/buildah/issues/2459

Fixes 2459

```release-note
NONE
```

### DIFF
--- a/chroot/run.go
+++ b/chroot/run.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -741,10 +742,13 @@ func runUsingChrootExecMain() {
 			os.Exit(1)
 		}
 	} else {
-		logrus.Debugf("clearing supplemental groups")
-		if err = syscall.Setgroups([]int{}); err != nil {
-			fmt.Fprintf(os.Stderr, "error clearing supplemental groups list: %v", err)
-			os.Exit(1)
+		setgroups, _ := ioutil.ReadFile("/proc/self/setgroups")
+		if strings.Trim(string(setgroups), "\n") != "deny" {
+			logrus.Debugf("clearing supplemental groups")
+			if err = syscall.Setgroups([]int{}); err != nil {
+				fmt.Fprintf(os.Stderr, "error clearing supplemental groups list: %v", err)
+				os.Exit(1)
+			}
 		}
 	}
 
--- a/chroot/run.go
+++ b/chroot/run.go
@@ -1097,7 +1097,8 @@ func setupChrootBindMounts(spec *specs.Spec, bundlePath string) (undoBinds func(
 		}
 		subSys := filepath.Join(spec.Root.Path, m.Mountpoint)
 		if err := unix.Mount(m.Mountpoint, subSys, "bind", sysFlags, ""); err != nil {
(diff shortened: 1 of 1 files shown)


mrinaldhillon:
@giuseppe @rhatdan will this fix enable chroot isolation for podman build without need of crun/runc? I am considering moving to podman since it  provides python SDK. Thanks

chroot/run.go
```diff
@@ -1093,7 +1097,8 @@ func setupChrootBindMounts(spec *specs.Spec, bundlePath string) (undoBinds func(
 		}
 		subSys := filepath.Join(spec.Root.Path, m.Mountpoint)
 		if err := unix.Mount(m.Mountpoint, subSys, "bind", sysFlags, ""); err != nil {
-			return undoBinds, errors.Wrapf(err, "error bind mounting /sys from host into mount namespace")
+			logrus.Warningf("could not bind mount %q, skipping", m.Mountpoint)
```
rhatdan, line 1100: Should also include err.

giuseppe: added

giuseppe:
> @giuseppe @rhatdan will this fix enable chroot isolation for podman build without need of crun/runc? I am considering moving to podman since it provides python SDK. Thanks

yes, you can use `--isolation=chroot` and have only one ID available

rhatdan:
LGTM
@nalind PTAL

chroot/run.go
```diff
@@ -741,10 +742,13 @@ func runUsingChrootExecMain() {
...
-		if err = syscall.Setgroups([]int{}); err != nil {
-			fmt.Fprintf(os.Stderr, "error clearing supplemental groups list: %v", err)
-			os.Exit(1)
+		setgroups, _ := ioutil.ReadFile("/proc/self/setgroups")
+		if strings.Trim(string(setgroups), "\n") != "deny" {
+			logrus.Debugf("clearing supplemental groups")
+			if err = syscall.Setgroups([]int{}); err != nil {
+				fmt.Fprintf(os.Stderr, "error clearing supplemental groups list: %v", err)
+				os.Exit(1)
+			}
```
nalind, line 751: LGTM

chroot/run.go
```diff
@@ -1093,7 +1097,8 @@ func setupChrootBindMounts(spec *specs.Spec, bundlePath string) (undoBinds func(
 		}
 		subSys := filepath.Join(spec.Root.Path, m.Mountpoint)
 		if err := unix.Mount(m.Mountpoint, subSys, "bind", sysFlags, ""); err != nil {
-			return undoBinds, errors.Wrapf(err, "error bind mounting /sys from host into mount namespace")
+			logrus.Warningf("could not bind mount %q, skipping: %v", m.Mountpoint, err)
+			continue
```
nalind, line 1101: The commit log says some filesystems might not be accessible.  Is this a thing we're seeing now (with which filesystems?), or is it a future-proofing thing?

giuseppe: I've seen the issue with `/sys/kernel/debug/tracing` on Fedora 32

nalind: Fair enough.  I'd be strongly tempted to add `/sys/kernel` or some of its subdirectories to the set of paths that we mask out, too, but that would be a different PR.

PR #2474, opened 2020-07-16
Vendor latest containers/storage

rhatdan:

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```


TomSweeneyRedHat:
looks like a go compilation issue bringing in c/storage and it's version of hashicorp

rhatdan:
Yes I think the issue is golang version, but I don't know why this is not blowing up on other CI systems.  Buildah is using the same images as Podman and Container/Storage.  I need @cevich to take a look.

TomSweeneyRedHat:
Might be cured similar to the way the @mtrmac did in a c/image Dockerfile here: https://github.com/containers/image/pull/987

rhatdan:
I think the issue here is the version of golang in the VM not in a specific Dockerfile/Containerfile.

cevich:
I think I follow, ya, install directly from golang.org under `/usr/local`.  This could create some complicated problems if the golang package is also installed (and necessary for some other dependencies?).  We'd be relying entirely upon `$PATH`  being correct...OTOH, this might not be too bad, since golang already requires the `go env` stuff being correct.  Hmmm.  Previously I was able to find a way to add a debian repo and install golang from a native package.  That would be a much preferred solution I think.  IIRC, the repo had 'backports' in the name, I can probably dig it up from the podman git commit history.

PR #2479, opened 2020-07-17
CompositeDigester: add the ability to filter tar streams

nalind:
#### What type of PR is this?

/kind other

Rework `CompositeDigester` so that it can handle type "".  Add a `newTarFilterer()` inspired by imagebuilder's, that we'll be able to use to adjust headers before digesting them, though that'll be done in a future change.

#### How to verify it

For now, existing tests should continue to succeed.

None

This changes the content digest that we compute and add to image history for directories, so this is a dividing line of sorts in that cache evaluation with newer versions will not always match images built with versions of buildah which do not include this change.

```
The hash values generated for ADD and COPY instructions in image histories will be different, and this will cause multi-layer builds (builds run with `--layers`) to not use intermediate images which were built with versions that do not include this change.
```

### DIFF
--- a/digester.go
+++ b/digester.go
@@ -1,63 +1,226 @@
 package buildah
 
 import (
+	"archive/tar"
+	"fmt"
 	"hash"
-	"strings"
+	"io"
+	"sync"
 
 	digest "github.com/opencontainers/go-digest"
+	"github.com/pkg/errors"
 )
 
-type singleDigester struct {
-	digester digest.Digester
-	prefix   string
+type digester interface {
+	io.WriteCloser
+	ContentType() string
+	Digest() digest.Digest
+}
+
+// A simple digester just digests its content as-is.
+type simpleDigester struct {
+	digester    digest.Digester
+	hasher      hash.Hash
+	contentType string
+}
+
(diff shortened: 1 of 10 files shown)


digester.go
```diff
@@ -1,63 +1,227 @@
...
 
-// Digest returns the prefix and a composite digest over everything that's been
-// digested.
+// Digest returns the content type and a composite digest over everything
+// that's been digested.
 func (c *CompositeDigester) Digest() (string, digest.Digest) {
+	if c.closer != nil {
+		c.closer.Close()
+		c.closer = nil
+	}
```
TomSweeneyRedHat, line 207: you do this if statement in at least 3 places now.  Worth a mini function?

nalind: Moved that logic into a method named `closeOpenDigester()`.

rhatdan:
LGTM
@vrothberg @TomSweeneyRedHat @ashley-cui PTAL

TomSweeneyRedHat:
It LGTM, but would really like a vrotherberg review before merging.
Thanks for the mini function @nalind!

rhatdan:
bors retry

PR #2480, opened 2020-07-17
WIP: placeholder

nalind:
This is a work in progress, mainly to let CI chew on it, so don't be surprised if it goes in and out of in-draft status.  Parts of it will be broken out into other PRs (#2479 to start), and this branch will be periodically rebased onto master until it's been whittled down to nothing.  Comments made here will be gleefully ignored and possibly deleted.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,7 @@ require (
 	github.com/opencontainers/runtime-spec v1.0.3-0.20200710190001-3e4195d92445
 	github.com/opencontainers/runtime-tools v0.9.0
 	github.com/opencontainers/selinux v1.6.0
-	github.com/openshift/imagebuilder v1.1.7
+	github.com/openshift/imagebuilder v1.1.8
 	github.com/pkg/errors v0.9.1
 	github.com/seccomp/libseccomp-golang v0.9.2-0.20200616122406-847368b35ebf
 	github.com/sirupsen/logrus v1.7.0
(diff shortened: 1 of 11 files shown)


rhatdan:
@nalind Any of the PRs in here ready to split out and get into buildah 1.16 release?

nalind:
The one that zeros timestamps in tar headers when we digest them is now #2546.  That depended on the digesting changes we split out as #2479 and then debugged a bit in #2497.  The workdir fix had to be rebased because it touches some of the same logic as a later patch, but it's open now as #2547.  I wouldn't be upset if 1.16 went out before either of them, though.

The main ones I'd want in a 1.16 are changes to the copier package, switching add/copy to using the copier package, and fixing integration tests that switching to the copier package exposes as having incorrect expectations (the tests being changed have been duplicated as conformance tests).

The changes to imagebuilder are already open as https://github.com/openshift/imagebuilder/pull/170.

nalind:
Hmm, the `TestConformance/ci-pipeline-modified` test started failing when [we vendored](https://github.com/containers/buildah/pull/2563) in [the change](https://github.com/containers/common/pull/267) to containers/common that added TERM to the hard-wired default environment values list.

I'm inclined to replace `defaultContainerConfig.Containers.Env` in `run_linux.go` with `[]string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"}` to fix it, since `buildah run` isn't advertised to work like `podman run`, and the configured environment variables will be added to the runtime's configuration when the built image is run anyway.

rhatdan:
@nalind What is going on with this one now?

nalind:
It's become my catch-all for fixing conformance test failures.

PR #2484, opened 2020-07-20
[release-v1.14] Make imagebuildah.BuildOptions.Architecture/OS optional

nalind:
#### What type of PR is this?

> /kind bug

When committing an image, the Architecture and OS fields in the imagebuildah.BuildOptions structure were applied unconditionally, even when they were empty, which would overwrite the usually-non-empty values supplied by the base image.

#### How to verify it

None

Cherry-picked from #2435.

```
None
```


TomSweeneyRedHat:
LGTM, tests aren't hip though

nalind:
Yeah, the tests look like they're failing due to the htpasswd-not-in-`docker.io/library/registry`-any-more problem we've encountered in other branches.

rhatdan:
I am just going to merge, rather then fight CI.

PR #2486, opened 2020-07-20
Add the "copier" package

nalind:
#### What type of PR is this?

> /kind other

Add new primitives for reading and writing data, represented as tar streams, from and to possibly-chrooted directories via subprocesses.  For now, these are library APIs that aren't being used yet.

#### How to verify it

Run the unit tests.

None directly.

This is part of #2480.  Public types whose docs claim them as implementation details are public for the sake of the JSON encoder, since we pass them over pipes to subprocesses when we're using chroot().

```
None
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -1,0 +1,1367 @@
+package copier
+
+import (
+	"archive/tar"
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"syscall"
+	"time"
+
+	"github.com/containers/buildah/util"
+	"github.com/containers/image/v5/pkg/compression"
+	"github.com/containers/storage/pkg/fileutils"
+	"github.com/containers/storage/pkg/idtools"
+	"github.com/containers/storage/pkg/reexec"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	copierCommand    = "buildah-copier"
+	maxLoopsFollowed = 64
+	// See http://pubs.opengroup.org/onlinepubs/9699919799/utilities/pax.html#tag_20_92_13_06, from archive/tar
+	cISUID = 04000 // Set uid, from archive/tar
+	cISGID = 02000 // Set gid, from archive/tar
+	cISVTX = 01000 // Save text (sticky bit), from archive/tar
+)
+
+func init() {
+	reexec.Register(copierCommand, copierMain)
+}
+
+// isArchivePath returns true if the specified path can be read like a (possibly
+// compressed) tarball.
+func isArchivePath(path string) bool {
+	f, err := os.Open(path)
+	if err != nil {
+		return false
+	}
+	defer f.Close()
+	rc, _, err := compression.AutoDecompress(f)
+	if err != nil {
+		return false
+	}
+	defer rc.Close()
+	tr := tar.NewReader(rc)
+	_, err = tr.Next()
+	return err == nil
+}
+
+// RequestType is an implementation detail of the copier package.
+type RequestType string
+
+const (
+	RequestStat RequestType = "STAT"
+	RequestGet  RequestType = "GET"
+	RequestPut  RequestType = "PUT"
+	RequestQuit RequestType = "QUIT"
+)
+
+// Options are set at the start and affect every request.  It is an implementation detail of the copier package.
+type Options struct {
+	Root      string
+	Directory string
+	Excludes  []string        `json:",omitempty"`
+	UIDMap    []idtools.IDMap `json:",omitempty"`
+	GIDMap    []idtools.IDMap `json:",omitempty"`
+}
+
+// Request encodes a single request.  It is an implementation detail of the copier package.
+type Request struct {
+	Request            RequestType
+	Root               string // used by all requests
+	preservedRoot      string
+	rootPrefix         string // used to reconstruct paths being handed back to the caller
+	Directory          string // used by all requests
+	preservedDirectory string
+	Globs              []string `json:",omitempty"` // used by stat, get
+	preservedGlobs     []string
+	StatOptions        StatOptions `json:",omitempty"`
+	GetOptions         GetOptions  `json:",omitempty"`
+	PutOptions         PutOptions  `json:",omitempty"`
+}
+
+// Response encodes a single response.  It is an implementation detail of the copier package.
+type Response struct {
+	Stat StatResponse
+	Get  GetResponse
+	Put  PutResponse
+}
+
+// StatResponse encodes a response for a single Stat request.  It is an implementation detail of the copier package.
+type StatResponse struct {
+	Error string
+	Globs []*StatsForGlob
+}
+
+// StatsForGlob encode results for a single glob pattern passed to Stat().
+type StatsForGlob struct {
+	Error   string                  `json:",omitempty"` // error if the Glob pattern was malformed
+	Glob    string                  // input pattern to which this result corresponds
+	Globbed []string                // a slice of zero or more names that match the glob
+	Results map[string]*StatForItem // one for each Globbed value if there are any, or for Glob
+}
+
+// StatForItem encode results for a single filesystem item, as returned by Stat().
+type StatForItem struct {
+	Error           string `json:",omitempty"`
+	Name            string
+	Size            int64       // dereferenced value for symlinks
+	Mode            os.FileMode // dereferenced value for symlinks
+	ModTime         time.Time   // dereferenced value for symlinks
+	IsSymlink       bool
+	IsDir           bool   // dereferenced value for symlinks
+	IsRegular       bool   // dereferenced value for symlinks
+	IsArchive       bool   // dereferenced value for symlinks
+	ImmediateTarget string `json:",omitempty"` // raw link content
+}
+
+// GetResponse encodes a response for a single Get request.  It is an implementation detail of the copier package.
+type GetResponse struct {
+	Error string `json:",omitempty"`
+}
+
+// PutResponse encodes a response for a single Put request.  It is an implementation detail of the copier package.
+type PutResponse struct {
+	Error string `json:",omitempty"`
+}
+
+// StatOptions controls parts of Stat()'s behavior.
+type StatOptions struct {
+	CheckForArchives bool     // check for and populate the IsArchive bit in returned values
+	Excludes         []string // contents to pretend don't exist
+}
+
+// Stat globs the specified pattern in the specified directory and returns its
+// results.
+// If root and directory are both not specified, the current root directory is
+// used, and relative names in the globs list are treated as being relative to
+// the current working directory.
+// If root is specified and the current OS supports it, the stat() is performed
+// in a chrooted context.  If the directory is specified as an absolute path,
+// it should either be the root directory or a subdirectory of the root
+// directory.  Otherwise, the directory is treated as a path relative to the
+// root directory.
+// Relative names in the glob list are treated as being relative to the
+// directory.
+func Stat(root string, directory string, options StatOptions, globs []string) ([]*StatsForGlob, error) {
+	req := Request{
+		Request:     RequestStat,
+		Root:        root,
+		Directory:   directory,
+		Globs:       append([]string{}, globs...),
+		StatOptions: options,
+	}
+	resp, err := copier(options.Excludes, nil, nil, req)
+	if err != nil {
+		return nil, err
+	}
+	if resp.Stat.Error != "" {
+		return nil, errors.New(resp.Stat.Error)
+	}
+	return resp.Stat.Globs, nil
+}
+
+// GetOptions controls parts of Get()'s behavior.
+type GetOptions struct {
+	UIDMap, GIDMap     []idtools.IDMap // map from hostIDs to containerIDs in the output archive
+	Excludes           []string        // contents to pretend don't exist
+	ExpandArchives     bool            // extract the contents of named items that are archives
+	StripSetidBits     bool            // strip the setuid/setgid/sticky bits off of items being copied. no effect on archives being extracted
+	StripXattrs        bool            // don't record extended attributes of items being copied. no effect on archives being extracted
+	KeepDirectoryNames bool            // export directories as directories containing items rather than as the items they contain
+}
+
+// Get produces an archive containing items that match the specified glob
+// patterns and writes it to bulkWriter.
+// If root and directory are both not specified, the current root directory is
+// used, and relative names in the globs list are treated as being relative to
+// the current working directory.
+// If root is specified and the current OS supports it, the contents are read
+// in a chrooted context.  If the directory is specified as an absolute path,
+// it should either be the root directory or a subdirectory of the root
+// directory.  Otherwise, the directory is treated as a path relative to the
+// root directory.
+// Relative names in the glob list are treated as being relative to the
+// directory.
+func Get(root string, directory string, options GetOptions, globs []string, bulkWriter io.Writer) error {
+	req := Request{
+		Request:   RequestGet,
+		Root:      root,
+		Directory: directory,
+		Globs:     append([]string{}, globs...),
+		StatOptions: StatOptions{
+			CheckForArchives: options.ExpandArchives,
+		},
+		GetOptions: options,
+	}
+	resp, err := copier(options.Excludes, nil, bulkWriter, req)
+	if err != nil {
+		return err
+	}
+	if resp.Get.Error != "" {
+		return errors.New(resp.Get.Error)
+	}
+	return nil
+}
+
+// PutOptions controls parts of Put()'s behavior.
+type PutOptions struct {
+	UIDMap, GIDMap    []idtools.IDMap // map from containerIDs to hostIDs when writing contents to disk
+	ChownDirs         *idtools.IDPair // set ownership of newly-created directories
+	ChmodDirs         *os.FileMode    // set permissions on newly-created directories
+	ChownFiles        *idtools.IDPair // set ownership of newly-created files
+	ChmodFiles        *os.FileMode    // set permissions on newly-created files
+	StripXattrs       bool            // don't bother trying to set extended attributes of items being copied
+	IgnoreXattrErrors bool            // ignore any errors encountered when attempting to set extended attributes
+}
+
+// Put extracts an archive from the bulkReader at the specified directory.
+// If root and directory are both not specified, the current root directory is
+// used.
+// If root is specified and the current OS supports it, the contents are written
+// in a chrooted context.  If the directory is specified as an absolute path,
+// it should either be the root directory or a subdirectory of the root
+// directory.  Otherwise, the directory is treated as a path relative to the
+// root directory.
+func Put(root string, directory string, options PutOptions, bulkReader io.Reader) error {
+	req := Request{
+		Request:    RequestPut,
+		Root:       root,
+		Directory:  directory,
+		PutOptions: options,
+	}
+	resp, err := copier(nil, bulkReader, nil, req)
+	if err != nil {
+		return err
+	}
+	if resp.Put.Error != "" {
+		return errors.New(resp.Put.Error)
+	}
+	return nil
+}
+
+func copier(excludes []string, bulkReader io.Reader, bulkWriter io.Writer, request Request) (*Response, error) {
+	if request.Directory == "" {
+		if request.Root == "" {
+			wd, err := getcwd()
+			if err != nil {
+				return nil, errors.Wrapf(err, "error getting current working directory")
+			}
+			request.Directory = wd
+		} else {
+			request.Directory = request.Root
+		}
+	}
+	if request.Root == "" {
+		request.Root = string(os.PathSeparator)
+	}
+	if filepath.IsAbs(request.Directory) {
+		_, err := filepath.Rel(request.Root, request.Directory)
+		if err != nil {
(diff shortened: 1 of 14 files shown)


rhatdan:
Should this be in pkg/copier?  Would that make it easier for Podman to use it?

rhatdan:
@vrothberg @mtrmac @giuseppe @TomSweeneyRedHat PTAL

nalind:
It has dependencies on our `util` package, while packages under `pkg` are generally expected to not need things outside of `pkg`.  I don't think podman would have a problem importing it from its current location, as it's comparable to `imagebuildah` that way.

mheon:
The location should be fine - we're probably dragging anything this does in already as part `podman build`.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+			s.Error = fmt.Sprintf("copier: stat: %q: %v", glob, syscall.ENOENT)
+		}
+	}
+	return &Response{Stat: StatResponse{Globs: stats}}
+}
+
+func copierHandlerGet(bulkWriter io.Writer, request Request, pm *fileutils.PatternMatcher, idMappings *idtools.IDMappings) (*Response, func() error, error) {
+	statRequest := request
+	statRequest.Request = RequestStat
+	statResponse := copierHandlerStat(request, pm)
```
mtrmac, line 836: AFAICS nothing at all ever uses this data, so, why? It’s expensive, it causes `Get` to set surprising `StatOptions`, and ultimately it doesn’t matter (apart from the possible reported error?).

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+		return errorResponse("%s", statResponse.Stat.Error)
+	}
+	if len(request.Globs) == 0 {
+		return errorResponse("copier: get: expected at least one glob pattern, got 0")
+	}
+	// build a queue of items by globbing
+	var queue []string
+	globMatchedCount := 0
+	for _, glob := range request.Globs {
+		globMatched, err := filepath.Glob(glob)
```
mtrmac, line 850: Nothing I can see prevents this from accessing paths outside `root` on non-chroot systems with a glob pattern, or a symlink, like `../../../../`…`/etc/*`

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+			return errorResponse("copier: get: glob %q: %v", glob, err)
+		}
+		globMatchedCount += len(globMatched)
+		filtered := make([]string, 0, len(globMatched))
+		for _, globbed := range globMatched {
+			rel, excluded, err := pathIsExcluded(request.Root, globbed, pm)
+			if err != nil {
+				return errorResponse("copier: get: checking if %q is excluded: %v", globbed, err)
+			}
+			if rel == "." || !excluded {
```
mtrmac, line 861: `pathIsExcluded` already has this special case. (And BTW I’m not sure why it exists, the API documentation does not say.)

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+				filtered = append(filtered, globbed)
+			}
+		}
+		if len(filtered) == 0 {
+			return errorResponse("copier: get: glob %q matched nothing (%d filtered out of %v): %v", glob, len(globMatched), globMatched, syscall.ENOENT)
+		}
+		queue = append(queue, filtered...)
+	}
+	// no matches -> error
+	if len(queue) == 0 {
```
mtrmac, line 871: (This is unreachable AFAICS, the code already checks for at least one glob and at least one `filtered` result.)

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+		hardlinkChecker := new(util.HardlinkChecker)
+		itemsCopied := 0
+		for i, item := range queue {
+			// if we're not discarding the names of individual directories, keep track of this one
+			relNamePrefix := ""
+			if request.GetOptions.KeepDirectoryNames {
+				relNamePrefix = filepath.Base(item)
+			}
+			// if the named thing-to-read is a symlink, dereference it
+			info, err := os.Lstat(item)
```
mtrmac, line 886: This can access files outside of `root` AFAICS.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+			if err != nil {
+				return errors.Wrapf(err, "copier: get: lstat %q", item)
+			}
+			// chase links. if we hit a dead end, we should just fail
+			followedLinks := 0
+			const maxFollowedLinks = 16
+			for info.Mode()&os.ModeType == os.ModeSymlink && followedLinks < maxFollowedLinks {
```
mtrmac, line 893: Nothing around here prevents access outside of `root` AFAICS.

Why is this a separate implementation from `resolvePath`?

Shouldn’t this treat `Excludes` the same way that code does? The API documentation doesn’t suggest any difference between `Stat` and `Get` in that respect.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+					return errors.Wrapf(err, "copier: get: %q(%q)", queue[i], item)
+				}
+				itemsCopied++
+			} else {
+				_, skip, err := pathIsExcluded(request.Root, item, pm)
+				if err != nil {
+					return err
+				}
+				if skip {
+					return nil
```
mtrmac, line 971: This completely ignores the rest of `queue`.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+		fileUID, fileGID = &request.PutOptions.ChownFiles.UID, &request.PutOptions.ChownFiles.GID
+	}
+	if idMappings != nil && !idMappings.Empty() {
+		containerDirPair := idtools.IDPair{UID: dirUID, GID: dirGID}
+		hostDirPair, err := idMappings.ToHost(containerDirPair)
+		if err != nil {
+			return errorResponse("copier: put: error mapping container filesystem owner %d:%d to host filesystem owners: %v", dirUID, dirGID, err)
+		}
+		dirUID, dirGID = hostDirPair.UID, hostDirPair.GID
+		if request.PutOptions.ChownFiles != nil {
```
mtrmac, line 1121: The API documentation does not suggest any asymmetry between `ChownDirs` and `ChownFiles`; why does the code does not look exactly the same?

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+			if err != nil {
+				return errorResponse("copier: put: error mapping container filesystem owner %d:%d to host filesystem owners: %v", fileUID, fileGID, err)
+			}
+			fileUID, fileGID = &hostFilePair.UID, &hostFilePair.GID
+		}
+	}
+	ensureDirectoryUnderRoot := func(directory string) error {
+		rel, err := filepath.Rel(request.Root, directory)
+		if err != nil {
+			return errors.Wrapf(err, "%q is not a subdirectory of %q", directory, request.Root)
```
mtrmac, line 1133: That’s not what `filepath.Rel` does; this will happily write outside of `Root`.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+		}
+		subdir := ""
+		for _, component := range strings.Split(rel, string(os.PathSeparator)) {
+			subdir = filepath.Join(subdir, component)
+			path := filepath.Join(request.Root, subdir)
+			if err := os.Mkdir(path, dirMode); err == nil {
+				err = os.Chown(path, dirUID, dirGID)
```
mtrmac, line 1140: If this is supposed to be safe against concurrent access, `Mkdir(path, 0700); Chown; Chmod`.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+		}
+		if err != nil {
+			return 0, errors.Wrapf(err, "copier: put: error opening file %q for writing", path)
+		}
+		defer f.Close()
+		n, err := io.Copy(f, tr)
+		if err != nil {
+			return n, errors.Wrapf(err, "copier: put: error writing file %q", path)
+		}
+		if err = f.Chmod(mode); err != nil {
```
mtrmac, line 1168: If this is supposed to be safe against concurrent access, `OpenFile(…, 0600); Chown; Chmod`.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+			} else {
+				if request.PutOptions.ChmodFiles != nil {
+					hdr.Mode = int64(*request.PutOptions.ChmodFiles)
+				}
+			}
+			// create the new item
+			devMajor := uint32(hdr.Devmajor)
+			devMinor := uint32(hdr.Devminor)
+			written := hdr.Size
+			switch hdr.Typeflag {
```
mtrmac, line 1245: Nothing prevents `path` from pointing outside of `Root` AFAICS.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+					if err = os.Remove(path); err == nil {
+						err = mkfifo(path, fifoMode)
+					}
+				}
+			}
+			// check for errors
+			if err != nil {
+				return errors.Wrapf(err, "error creating %q", path)
+			}
+			if written != hdr.Size {
```
mtrmac, line 1316: (It might be simpler to check for this only in the `TypeReg` case.)

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+				}
+			}
+			// restore permissions, except for symlinks, since we don't have lchmod
+			if hdr.Typeflag != tar.TypeSymlink {
+				if err = os.Chmod(path, mode); err != nil {
+					return errors.Wrapf(err, "error setting permissions on %q to 0%o", path, mode)
+				}
+			}
+			// set ownership
+			if err = os.Lchown(path, hdr.Uid, hdr.Gid); err != nil {
```
mtrmac, line 1334: If this is supposed to be safe against concurrent access, create with 0700 or less; `Lchown; Chmod`.

copier/xattrs.go
```diff
@@ -0,0 +1,87 @@
...
+// Lgetxattrs returns a map of the relevant extended attributes set on the given file.
+func Lgetxattrs(path string) (map[string]string, error) {
+	maxSize := 64 * 1024 * 1024
+	listSize := 64 * 1024
+	var list []byte
+	for listSize < maxSize {
+		list = make([]byte, listSize)
+		size, err := unix.Llistxattr(path, list)
+		if err != nil {
+			if unwrapError(err) == syscall.E2BIG {
```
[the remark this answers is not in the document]
nalind: Nice catch on `E2BIG` vs `ERANGE`!

copier/xattrs.go
```diff
@@ -0,0 +1,87 @@
...
+	if listSize >= maxSize {
+		return nil, errors.Errorf("unable to read list of attributes for %q: size would have been too big", path)
+	}
+	m := make(map[string]string)
+	for _, attribute := range strings.Split(string(list), string('\000')) {
+		for _, relevant := range relevantAttributes {
+			matched, err := filepath.Match(relevant, attribute)
+			if err != nil || !matched {
+				continue
+			}
```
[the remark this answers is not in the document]
nalind: Plus, it makes the logic a little bit easier to follow.  Factored.

copier/syscall_windows.go
```diff
@@ -0,0 +1,57 @@
...
+	"golang.org/x/sys/windows"
+)
+
+var canChroot = false
+
+func chroot(path string) (bool, error) {
+	return false, nil
+}
+
+func getcwd() (string, error) {
```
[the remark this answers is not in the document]
nalind: For whatever reason I completely forgot that `os.Getwd()` existed.  I probably expected it to be called `os.Getcwd()` and didn't find such a function.  Dropping it.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+}
+
+// GetOptions controls parts of Get()'s behavior.
+type GetOptions struct {
+	UIDMap, GIDMap     []idtools.IDMap // map from hostIDs to containerIDs in the output archive
+	Excludes           []string        // contents to pretend don't exist
+	ExpandArchives     bool            // extract the contents of named items that are archives
+	StripSetidBits     bool            // strip the setuid/setgid/sticky bits off of items being copied. no effect on archives being extracted
+	StripXattrs        bool            // don't record extended attributes of items being copied. no effect on archives being extracted
+	KeepDirectoryNames bool            // export directories as directories containing items rather than as the items they contain
```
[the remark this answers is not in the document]
nalind: Does `don't strip the top directory's basename from the paths of items in subdirectories` read better?

As background, Dockerfile `COPY` and `ADD` have this sometimes-surprising behavior where if the source is a directory (say "src"), its contents (say "foo" and "bar") are each copied to a destination (say "dest") rather than the directory itself, producing "dest/foo" and "dest/bar" rather than "dest/src/foo" and "dest/src/bar".  This happens when the source is a glob, too, so if the glob matches both files and directories, the files and the contents of the directories are what are copied to the destination directory.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+// Response encodes a single response.  It is an implementation detail of the copier package.
+type Response struct {
+	Stat StatResponse
+	Get  GetResponse
+	Put  PutResponse
+}
+
+// StatResponse encodes a response for a single Stat request.  It is an implementation detail of the copier package.
+type StatResponse struct {
+	Error string
```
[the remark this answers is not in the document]
nalind: Sure, merged them into a single field in the response struct.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+			}
+			request.Directory = wd
+		} else {
+			request.Directory = request.Root
+		}
+	}
+	if request.Root == "" {
+		request.Root = string(os.PathSeparator)
+	}
+	if filepath.IsAbs(request.Directory) {
```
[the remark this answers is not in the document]
nalind: You're right - `filepath.IsAbs()` expects the path to begin with a volume (drive letter or UNC share path), so on top of liberal applications of `filepath.FromSlash()` and `filepath.ToSlash()`, I've had to special-case this.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+	switch request.Request {
+	case RequestStat:
+		break
+	case RequestQuit:
+		break
+	case RequestGet:
+		uidMap, gidMap = request.GetOptions.UIDMap, request.GetOptions.GIDMap
+	case RequestPut:
+		uidMap, gidMap = request.PutOptions.UIDMap, request.PutOptions.GIDMap
+	}
```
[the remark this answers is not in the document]
nalind: Doh.  Added.

copier/copier.go
nalind: Changed it.  Thanks!

copier/copier.go
nalind: Sure, changed it.

copier/copier.go
nalind: Started prepending the path separator to `hdr.Name` and running the result through `filepath.Clean()`, which should first remove any internal ".." references, and then remove any initial ones, before joining the result with the top level directory.  I _think_ that's an improvement.

copier/copier.go
nalind: Changed it, thanks!

copier/copier.go
nalind: Changed it, thanks!

copier/copier.go
nalind: Aargh, missed that.  Prepending the path separator, running it through `filepath.Clean()` to remove initial "../" instances, and then trimming the prefixed path separator looks like it produces the desired result.

copier/copier.go
nalind: We need defaults for directory ownership for the sake of any intermediate directories that we need to create that aren't part of the archive that we're extracting, but we want to use the supplied values if any are specified for directories.

copier/copier.go
nalind: Not sure I follow.  This is the callback that `filepath.Walk()` calls, so returning `nil` looks like a normal result.

mtrmac: Do you mean the callback that starts at line 917:
```go
			if info.Mode().IsDir() {
 				walkfn := func(path string, info os.FileInfo, err error) error {
```
?

At least based on indentation, that ends on line 959:
```go
					return copierHandlerGetOne(info, symlinkTarget, rel, path, request.GetOptions, tw, hardlinkChecker, idMappings)
 				}
 				// walk the directory tree, checking/adding items individually
```

nalind: Eek, you're right.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+	request.preservedGlobs = append([]string{}, request.Globs...)
+	if !filepath.IsAbs(request.Directory) {
+		request.Directory = filepath.Join(request.Root, request.Directory)
+	}
+	absoluteGlobs := make([]string, 0, len(request.Globs))
+	for i, glob := range request.preservedGlobs {
+		if filepath.IsAbs(glob) {
+			absoluteGlobs = append(absoluteGlobs, request.Globs[i])
+		} else {
+			absoluteGlobs = append(absoluteGlobs, filepath.Join(request.Directory, request.Globs[i]))
```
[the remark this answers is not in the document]
nalind: No reason.  Changing it.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+	switch request.Request {
+	case RequestStat:
+		break
+	case RequestQuit:
+		break
+	case RequestGet:
+		options.UIDMap, options.GIDMap = request.GetOptions.UIDMap, request.GetOptions.GIDMap
+	case RequestPut:
+		options.UIDMap, options.GIDMap = request.PutOptions.UIDMap, request.PutOptions.GIDMap
+	}
```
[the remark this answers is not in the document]
nalind: Setting a `default`.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+		return nil, errors.Wrapf(err, "pipe")
+	}
+	cmd.Dir = "/"
+	cmd.Env = append([]string{fmt.Sprintf("LOGLEVEL=%d", logrus.GetLevel())}, os.Environ()...)
+
+	errorBuffer := bytes.Buffer{}
+	cmd.Stdin = stdinRead
+	cmd.Stdout = stdoutWrite
+	cmd.Stderr = &errorBuffer
+	cmd.ExtraFiles = append(cmd.ExtraFiles, bulkReaderRead, bulkWriterWrite)
```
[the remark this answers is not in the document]
nalind: Changed to set the slice directly.  We hard-wire the descriptor numbers in the child, so either way, if `reexec` ever starts setting something in the slice, we'll need to revisit this.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+		stdinWrite.Close()
+		return nil, errors.Wrapf(err, "pipe")
+	}
+	decoder := json.NewDecoder(stdoutRead)
+	bulkReaderRead, bulkReaderWrite, err := os.Pipe()
+	if err != nil {
+		stdinRead.Close()
+		stdinWrite.Close()
+		stdoutRead.Close()
+		stdoutWrite.Close()
```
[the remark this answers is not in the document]
nalind: Yup, that is shorter.  Changed.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+		}
+		excluded = excluded || thisExcluded
+		if !excluded {
+			if target, err := os.Readlink(filepath.Join(workingPath, components[0])); err == nil {
+				followed++
+				if followed > maxLoopsFollowed {
+					return "", &os.PathError{
+						Op:   "open",
+						Path: path,
+						Err:  err,
```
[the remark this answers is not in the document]
nalind: Yes, it should.  Changed it.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+	stdinRead.Close()
+	stdoutWrite.Close()
+	bulkReaderRead.Close()
+	bulkWriterWrite.Close()
+	if err = encoder.Encode(options); err != nil {
+		stdinWrite.Close()
+		stdoutRead.Close()
+		bulkReaderWrite.Close()
+		bulkWriterRead.Close()
+		if err2 := cmd.Process.Kill(); err2 != nil {
```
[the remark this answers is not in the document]
nalind: Yup, `Wait()` needs to always be called.  A deferred function that still kept track of which step we failed at ended up getting complicated, but a helper that takes that description, kills the process, and then returns the values that the calling function should return looks like it'll do about as well.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+	return resp.Stat.Globs, nil
+}
+
+// GetOptions controls parts of Get()'s behavior.
+type GetOptions struct {
+	UIDMap, GIDMap     []idtools.IDMap // map from hostIDs to containerIDs in the output archive
+	Excludes           []string        // contents to pretend don't exist
+	ExpandArchives     bool            // extract the contents of named items that are archives
+	StripSetidBits     bool            // strip the setuid/setgid/sticky bits off of items being copied. no effect on archives being extracted
+	StripXattrs        bool            // don't record extended attributes of items being copied. no effect on archives being extracted
```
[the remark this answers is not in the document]
nalind: FWIW `security.capability` is the only one I'd personally worry about.

copier/copier.go
```diff
@@ -0,0 +1,1367 @@
...
+			return nil, err
+		}
+	}
+	return response, nil
+}
+
+func copierWithSubprocess(excludes []string, bulkReader io.Reader, bulkWriter io.Writer, request Request) (*Response, error) {
+	options := Options{
+		Root:      request.Root,
+		Directory: request.Directory,
```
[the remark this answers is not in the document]
nalind: Okay, dropping the separate `options` structure, and using helper methods to pull the right lists of exclusions and ID maps from the right passed-in `...Options` struct for the request type.

Issue #2478, opened 2020-07-17
buildah run --net=none bind mounts /etc/hosts


mrinaldhillon:
**Description**

I need to add static host map entries to /etc/hosts during build. But buildah mounts generated file even if network is set to none. Changes to this file also don't persist.

**Steps to reproduce the issue:**
1. run buildah with --net=none
2. make changes /etc/hosts and exit
3. mount the container, /etc/hosts file is empty

**Describe the results you received:**
```
buildah run --isolation=chroot --net=none debian-working-container /bin/bash
root@ff257a6158f3:/# cat /etc/hosts
# Generated by Buildah
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
127.0.1.1	fedora # prltoolsd vm name
127.0.0.1   ff257a6158f3
::1         ff257a6158f3
root@ff257a6158f3:/# echo "128.0.0.4 fdev" > /etc/hosts
root@ff257a6158f3:/# cat /etc/hosts
128.0.0.4 fdev
root@ff257a6158f3:/# exit
[root@fedora ~]# m=$(buildah mount debian-working-container)
[root@fedora ~]# cat $m/etc/hosts
[root@fedora ~]#
```

**Describe the results you expected:**
/etc/hosts should not be bind mounted if network is none
Changes to /etc/hosts should persist.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.15.0-1.fc32.x86_64
```

**Output of `buildah version`:**

```
Version:         1.15.0
Go Version:      go1.14.3
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        0.4.0
libcni Version:
image Version:   5.5.1
Git Commit:
Built:           Wed Dec 31 19:00:00 1969
OS/Arch:         linux/amd64
```

**Output of `podman version` if reporting a `podman build` issue:**

```
(paste your output here)
```

**Output of `cat /etc/*release`:**

```
Fedora release 32 (Thirty Two)
NAME=Fedora
VERSION="32 (Workstation Edition)"
ID=fedora
VERSION_ID=32
VERSION_CODENAME=""
PLATFORM_ID="platform:f32"
PRETTY_NAME="Fedora 32 (Workstation Edition)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
[...]

PR #2488, opened 2020-07-22
Don't bind /etc/host and /etc/resolv.conf if network is not present

TomSweeneyRedHat:
If there's no network present in the container, don't bind
the /etc/hosts file as any changes to from the caller will
be lost when run exits.  Ditto the /etc/resolv.conf file.

Addresses: #2478


tests/run.bats
```diff
@@ -422,3 +422,40 @@ function configure_and_check_user() {
...
+        run_buildah from --quiet --pull=false --signature-policy ${TESTSDIR}/policy.json alpine
+        cid=$output
+        run_buildah run --isolation=chroot --network=none $cid sh -c 'echo "110.110.110.0 fake_host" >> /etc/hosts; cat /etc/hosts'
+	expect_output "110.110.110.0 fake_host"
+        m=$(buildah mount $cid)
+	run cat $m/etc/hosts
+	[ "$status" -eq 0 ]
+	expect_output "110.110.110.0 fake_host"
+	run_buildah rm -a
+}
```
TomSweeneyRedHat, line 461: Ugh,

tab/spaces issues, will fix.

rhatdan:
LGTM

rhatdan:
@nalind @mheon @vrothberg @QiWang19 @ashley-cui PTAL

QiWang19:
LGTM

rhatdan:
Looks like either we have a bug or a test needs to be adjusted.

TomSweeneyRedHat:
network error in an unrelated test, bors be kind this time:
bors retry

tests/from.bats
```diff
@@ -255,7 +255,7 @@ load helpers
   _prefetch alpine
   run_buildah from --quiet --add-host=localhost:127.0.0.1 --pull --signature-policy ${TESTSDIR}/policy.json alpine
   cid=$output
-  run_buildah run $cid -- cat /etc/hosts
```
TomSweeneyRedHat, line 258: @nalind, does this look hip to you?  This is my one worry spot with this change.  Debian and Alpine act differently.  If you take the "stock" alpine image and run with current upstream Buildah, the run command without the `--net-option` would retain the following in the containers /etc/host file in the containers storage on the host:
```
127.0.0.1	localhost localhost.localdomain
::1		localhost localhost.localdomain
```

And if you investigate the container when it's running, you first see the "# Buildah Generated" lines, then these two lines.  However, the Buildah Generated lines are not retained on disk for the container, only these two lines.

Debian however, has nothing except the "# Buildah Generated" lines when it too is running and then nothing in the /etc/host in the container storage on the host.

With these changes, that file in container storage is empty for both distros.  In both this variant of Buildah and current upstream, the host specified in --add-host is not retained when the run command completes.

rhatdan: Do a buildah mount and see what the contents of the /etc/hosts file are. these hosts files saved to the image, are from the underlying image, which is why you are seeing the difference I believe.  
I don't know why in a running container the localhost data provided by Buildah would be different.  
I don't think we ever save the contents of /etc/hosts or /etc/resolv.conf into a container image.

run_linux.go
```diff
@@ -192,7 +192,10 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 		if err != nil {
 			return err
 		}
-		bindFiles["/etc/hosts"] = hostFile
+		// Only bind /etc/hosts if there's a network
+		if configureNetwork {
```
rhatdan, line 196: Should we do the same thing for resolv.conf?

tests/from.bats
TomSweeneyRedHat: If I do `buildah run {alpine||debian}-working-container cat /etc/hosts`  on both distros I see:

```
# Generated by Buildah
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
127.0.0.1   0f8200b7e866
::1         0f8200b7e866
```
If I mount them, then two different but similar lines in the alpine mount, and none in Debian.

```
# ma=$(buildah mount alpine-working-container)
# mb=$(buildah mount debian-working-container)

# cat $ma/etc/hosts
127.0.0.1	localhost localhost.localdomain
::1		localhost localhost.localdomain

# cat $mb/etc/hosts

```
So it looks like we're getting something from alpine directly and it's not part of the generated one that we make when running.  I'm not quite sure how/where that's coming from.

The main focus of this fix is just to retain whatever the user adds to /etc/hosts if `--net=none`

run_linux.go
TomSweeneyRedHat: On the fence.  Do you or others want to tip me one way or the other @rhatdan ?  I'm leaning towards yes, but not enough to tip over yet.

TomSweeneyRedHat:
I've repushed, the from.bats test was apparently still failing, but it works in my sandbox.  I'm not sure the push will fix anything, but we'll see.

run_linux.go
mrinaldhillon: IMHO /etc/hosts /etc/hostname /etc/resolve.conf should not be managed with net=none

rhatdan: I agree.

TomSweeneyRedHat: Sold!

I'll add `/etc/resolv.conf` too to the PR and will push a new variant out later today or Monday.

TomSweeneyRedHat:
Turns out the `configureNetwork` changes holds when doing `buildah run`, but does not hold when there's a `RUN` statement in a Containerfile.  I played with it a bit last night, and think I've a solution for both, more testing today.

TomSweeneyRedHat:
All happy green test buttons.  This is ready to go in with appropriate reviews.  I would like a @nalind head nod if he's got a minute or two.

rhatdan:
bors r+
@TomSweeneyRedHat could you look to see if we have the same issue with Podman?

TomSweeneyRedHat:
Created an investigative issue on Podman so I won't forget.   https://github.com/containers/podman/issues/7206

PR #2489, opened 2020-07-23
Add conformance testing to CI

nalind:
#### What type of PR is this?

> /kind cleanup

Moves the conformance tests from ginkgo to using the default testing package and github.com/stretchr/testify/require, preserving the existing tests and adding more.

Add conformance tests to our Cirrus configuration, currently marked as an allowed failure while we get known failures fixed.

#### How to verify it

Running conformance tests as described in tests/conformance/README.md should still flag some known differences between what we produce and what `docker build` produces with dockerd 19.03.

None

This should help spot errors in https://github.com/containers/buildah/pull/2480 and ensure that it doesn't introduce regressions.  Some failures will be fixed by https://github.com/openshift/imagebuilder/pull/170.

```
None
```


TomSweeneyRedHat:
@edsantiago PTAL

TomSweeneyRedHat:
@nalind looks like lint isn't buying today

tests/conformance/README.md
```diff
@@ -2,81 +2,30 @@
...
-### Install ginkgo and gomega
-
-We already have vendored ginkgo and gomega into the conformance tests, so if you want to just run the tests with "go test", you can skip this step.
-Ginkgo is tested with Go 1.6+, please make sure your golang version meet the required version.
-```
-go get -u github.com/onsi/ginkgo/ginkgo  # installs the ginkgo CLI
-go get -u github.com/onsi/gomega/...     # fetches the matcher library
-export PATH=$PATH:$GOPATH/bin
-```
+Conformance tests use Docker CE to build images to be compared with images built with Buildah.  Install Docker CE with dnf, yum or apt-get, based on your distribution and verify that the `docker` service is started.  In Fedora, RHEL and CentOS `docker` or `moby-engine` rather than Docker CE may be installed by default.  In Debian or Ubuntu you may instead have the `docker.io` package.  Please verify that you install a recent version Docker.
```
edsantiago, line 14: "recent" is meaningless in static documents. If there is a hard dependency, could you state it? If you mean 'the most recent version of Docker available to you', please state it that way or similarly. If you mean something else, please clarify.

nalind: Right, it's also badly phrased.

nalind: Replaced that sentence with `Please verify that you install at least version 19.03.`

contrib/cirrus/test.sh
```diff
@@ -40,6 +40,21 @@ then
...
+                warn "Installing moby-engine"
+                dnf install -y moby-engine
+                systemctl enable --now docker
+                ;;
+            ubuntu)
+                warn "Installing docker.io"
+                $LONG_APTGET install docker.io
+                systemctl enable --now docker
+                ;;
+            esac
```
edsantiago, line 55: How about adding a `*)` case here with a failure message?

nalind: Sure.

nalind: setup.sh calls the `bad_os_id_ver` shell function for similar cases, so went with that.

nalind:
Oh, to be clear, many of these tests don't pass.  We're going to fix that.

PR #2490, opened 2020-07-23
Fix a race we hit during conformance tests

nalind:
#### What type of PR is this?

> /kind bug

The imagebuilder.Executor.imageMap map is written in one goroutine which waits on StageExecutors, and is read from the goroutines that run the  StageExecutors, so we need to synchronize access to the map.

#### How to verify it

Found by running conformance tests with race detection enabled.

None

```
None
```


rhatdan:
LGTM
@vrothberg @giuseppe @TomSweeneyRedHat @ashley-cui @QiWang19 PTAL

giuseppe:
bors s+

nalind:
bors retry

rhatdan:
Saving the planet from bors.

PR #2497, opened 2020-07-28
tarFilterer: handle multiple archives

nalind:
#### What type of PR is this?

/kind other

When processing `.dockerignore` files, we produce mini-archives for each individual item being copied as part of a directory tree, so if we pass them through a `tarFilterer`, it needs to not give up when it hits the end of the first archive that it sees.  Add a flag to it that it sets when it closes the writing end of the pipe that it's reading, and if the flag isn't set when the reader finds an EOF, have the reader gear up to read another archive instead of just refusing to process any more data.

#### How to verify it

New unit tests exercise the code path that would have been tripped up.

`None`

This expands on #2479 by giving the callback that tar filterers use more options, again inspired by `imagebuilder`.

```
None
```

### DIFF
--- a/digester.go
+++ b/digester.go
@@ -54,6 +54,8 @@ func (s *simpleDigester) Digest() digest.Digest {
 type tarFilterer struct {
 	wg         sync.WaitGroup
 	pipeWriter *io.PipeWriter
+	closedLock sync.Mutex
+	closed     bool
 	err        error
 }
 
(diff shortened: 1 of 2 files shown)


digester.go
```diff
@@ -62,50 +63,65 @@ func (t *tarFilterer) Write(p []byte) (int, error) {
 }
 
 func (t *tarFilterer) Close() error {
+	t.closed = true
```
rhatdan, line 66: Should you check if t.closed is true, and return success or an error, rather then executing it a second time?

nalind: Can't hurt.  Added a check there.

rhatdan:
LGTM
@TomSweeneyRedHat @giuseppe @vrothberg @QiWang19 @ashley-cui PTAL

digester.go
nalind: Good news: we're definitely running the unit tests.  Bad news: they found a race.  Good news: added a lock to fix it.

TomSweeneyRedHat:
LGTM 
assuming happy tests

PR #2498, opened 2020-07-29
copier: split StripSetidBits into StripSetuidBit/StripSetgidBit/StripStickyBit

nalind:
#### What type of PR is this?

/kind api-change

For the sake of conformance tests, callers need to be able to strip setuid and setgid bits from contents being copied from the build context while leaving the sticky bit intact.  Split the `StripSetidBits` option for `copier.Get()` into three separate flags (`StripSetuidBit`/`StripSetgidBit`/`StripStickyBit`).

#### How to verify it

The `copier` package's unit tests have been expanded to test that we strip the bits as requested.

None

```
None
```


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

PR #2499, opened 2020-07-29
docs/buildah-commit.md: tweak some wording, add a --rm example

nalind:
#### What type of PR is this?

/kind documentation

Tweak wording in `docs/buildah-commit.md` to avoid suggesting that we add the actual text "<none>:<none>" as a name for images when a name isn't specified to `buildah commit`.

Add an example that uses `--rm` to the list of examples in `docs/buildah-commit.md`.

#### How to verify it

Visual inspection.

None

None

```
None
```

### DIFF
--- a/docs/buildah-commit.md
+++ b/docs/buildah-commit.md
@@ -10,7 +10,7 @@ buildah\-commit - Create an image from a working container.
 Writes a new image using the specified container's read-write layer and if it
 is based on an image, the layers of that image.  If *image* does not begin
 with a registry name component, `localhost` will be added to the name.  If
-*image* is not provided, the image will have no name.
+*image* is not provided, the image will have no name.  When an image has no name, the `buildah images` command will display `<none>` in the REPOSITORY and TAG columns. 
 
 ## RETURN VALUE
 The image ID of the image that was created.  On error, 1 is returned and errno is returned.
--- a/docs/buildah-commit.md
+++ b/docs/buildah-commit.md
@@ -10,8 +10,7 @@ buildah\-commit - Create an image from a working container.
 Writes a new image using the specified container's read-write layer and if it
 is based on an image, the layers of that image.  If *image* does not begin
 with a registry name component, `localhost` will be added to the name.  If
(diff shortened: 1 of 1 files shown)


docs/buildah-commit.md
```diff
@@ -10,8 +10,7 @@ buildah\-commit - Create an image from a working container.
 Writes a new image using the specified container's read-write layer and if it
 is based on an image, the layers of that image.  If *image* does not begin
 with a registry name component, `localhost` will be added to the name.  If
-*image* is not provided, the values for the `REPOSITORY` and `TAG` values of
-the created image will each be set to `<none>`.
+*image* is not provided, the image will have no name.
```
TomSweeneyRedHat, line 13: we've had a lot of confusion over the  '<none>' tags over the years.  Would you mind adding :
```suggestion
*image* is not provided, the image will have no name, denoted by the value `<none>` in the REPOSITORY and TAG values of the created image.
```

rhatdan:
LGTM, once you answer @TomSweeneyRedHat Request.

docs/buildah-commit.md
nalind: That text suggests to me that the text `<none>` is actually stored somewhere related to the image, which is not the case.

rhatdan: Can we state that this will be listed this way when executing buildah images.

TomSweeneyRedHat: OK,  I understand the concern @nalind, how's about something like:

```suggestion
*image* is not provided, the image will have no name.  When an image has no name, the `buildah images` command will display `<none>` in the REPOSITORY and TAG columns. 
```
I know it's kind of cross command man paging, but it's been a confusion point.  I'm fine if you want to change the above wording.

rhatdan: Sure, but we also need to add `buildah-images(1)` man page to the bottom.

rhatdan:
Replaced by: https://github.com/containers/buildah/pull/2548

PR #2500, opened 2020-07-30
blobcache: avoid an unnecessary NewImage()

nalind:
#### What type of PR is this?

/kind bug

Avoid calling `NewImage()` on the source reference when the `ImageSource` that we already have returns `nil` from its `LayerInfosForCopy()` method.

For our container-as-image references, that causes a re-extraction of uncached blobs, recomputing their digests, rebuilding the config blobs and manifests.

The image library's `copy.Image()` function then asks the source reference that we're wrapping for blobs that we listed in the rebuilt manifest, and if any of those values differ, we fail to read them.

This would have only affected builds that specified a blob cache.

#### How to verify it

Build an image with `buildah bud` with debugging enabled, using the hidden `--blob-cache` option.  Observe that before this fix is applied, for each layer we commit, the layers list is computed twice, temporary directories are created twice, and the config blobs and manifests are generated twice.  With the fix applied, these things should only happen once.

None

```
None
```

### DIFF
--- a/pkg/blobcache/blobcache.go
+++ b/pkg/blobcache/blobcache.go
@@ -275,12 +275,11 @@ func (s *blobCacheSource) LayerInfosForCopy(ctx context.Context, instanceDigest 
 		return nil, errors.Wrapf(err, "error getting layer infos for copying image %q through cache", transports.ImageName(s.reference))
 	}
 	if infos == nil {
-		image, err := s.reference.NewImage(ctx, &s.sys)
+		img, err := image.FromUnparsedImage(ctx, &s.sys, image.UnparsedInstance(s.source, instanceDigest))
 		if err != nil {
 			return nil, errors.Wrapf(err, "error opening image to get layer infos for copying image %q through cache", transports.ImageName(s.reference))
 		}
-		defer image.Close()
-		infos = image.LayerInfos()
+		infos = img.LayerInfos()
 	}
 
 	if canReplaceBlobs && s.reference.compress != types.PreserveOriginal {


nalind:
I believe that when the contents of those new layers change between the two computations (for reasons I haven't pinned down yet), we get the errors seen in https://bugzilla.redhat.com/show_bug.cgi?id=1720730.

rhatdan:
LGTM
Will we get an increase in speed do to this fix?

nalind:
If you're using a blob cache (which is not used by default by the buildah or podman CLI tools), then yes, contents of layers that are added during the build will be extracted once instead of twice.  OpenShift, for example, uses blob caches, so it will benefit.

vrothberg:
I restarted the in_podman test. Upload failed.

nalind:
bors retry

pkg/blobcache/blobcache.go
```diff
@@ -275,12 +275,11 @@ func (s *blobCacheSource) LayerInfosForCopy(ctx context.Context, instanceDigest
 		return nil, errors.Wrapf(err, "error getting layer infos for copying image %q through cache", transports.ImageName(s.reference))
 	}
 	if infos == nil {
-		image, err := s.reference.NewImage(ctx, &s.sys)
+		img, err := image.FromUnparsedImage(ctx, &s.sys, image.UnparsedInstance(s.source, nil))
```
nalind, line 278: This `nil` should be `instanceDigest`, right?

TomSweeneyRedHat:
LGTM
bors got into a funky "Cancelled" state.  Maybe you knocked it off the rails with your submit?

TomSweeneyRedHat:
bors retry

nalind:
bors retry

Issue #2030, opened 2019-12-16
buildah rm nonexistent-container: mount leak


edsantiago:
This is why buildah BATS tests fail using overlay. (Or at least it's one reason).
```console
# mkdir -p /var/tmp/mytmp/{,run}root
# buildah --root /var/tmp/mytmp/root --runroot /var/tmp/mytmp/runroot from alpine
Getting image source signatures
Copying blob 89d9c30c1d48 done
Copying config 965ea09ff2 done
Writing manifest to image destination
Storing signatures
alpine-working-container
# buildah --root /var/tmp/mytmp/root --runroot /var/tmp/mytmp/runroot commit --rm alpine-working-container alpine-image
Getting image source signatures
Copying blob 77cae8ab23bf skipped: already exists
Copying blob 6b316488d473 done
Copying config 4f38e26433 done
Writing manifest to image destination
Storing signatures
4f38e26433f6d044f6228b622f4186e49b0bf2f43c97a93125016012821854d1

# buildah --root /var/tmp/mytmp/root --runroot /var/tmp/mytmp/runroot rm alpine-working-container
error removing container "alpine-working-container": error reading build container: container not known
^^^^ this is expected: the commit was done with --rm

# rm -rf /var/tmp/mytmp
rm: cannot remove '/var/tmp/mytmp/root/overlay': Device or resource busy
^^^ this is not expected.
# mount|grep mytmp
/dev/vda1 on /var/tmp/mytmp/root/overlay type ext4 (rw,relatime,seclabel)
```

rhatdan:
I am not sure this is  a new leak.  But we did remove some functionality from Overlay to Skip local mount.  Which might be what is causing this.

edsantiago:
What I didn't mention is that the `rm -rf` works fine immediately after the `commit`. That is: somewhere in the `buildah rm` a mount is happening, and is not getting cleaned up in the error exit. I think.

rhatdan:
@edsantiago do you know if this is still an issue?

edsantiago:
I was just able to reproduce it on my laptop, buildah-1.15.0-1.fc32, using the steps in the original comment

rhatdan:
Storage shutdown should fix this one also.

PR #2515, opened 2020-08-06
Shutdown storage on failures

rhatdan:
Buildah is leaving storage enabled when commands fail.
Specifically we have seen this with buildah from.
This can lead to a leaked mount point on the system.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/cmd/buildah/main.go
+++ b/cmd/buildah/main.go
@@ -168,14 +168,26 @@ func before(cmd *cobra.Command) error {
 	return nil
 }
 
-func after(cmd *cobra.Command) error {
+func shutdownStore(cmd *cobra.Command) error {
 	if needToShutdownStore {
 		store, err := getStore(cmd)
 		if err != nil {
 			return err
 		}
-		_, _ = store.Shutdown(false)
+		logrus.Debugf("shutting down the store")
+		needToShutdownStore = false
+		if _, err = store.Shutdown(false); err != nil {
+			logrus.Warnf("failed to shutdown storage: %q", err)
+		}
 	}
+	return nil
+}
+
+func after(cmd *cobra.Command) error {
+	if err := shutdownStore(cmd); err != nil {
+		return err
+	}
+
 	if globalFlagResults.CPUProfile != "" {
 		pprof.StopCPUProfile()
 		globalFlagResults.cpuProfileFile.Close()
@@ -209,6 +221,9 @@ func main() {
 			if w, ok := ee.Sys().(syscall.WaitStatus); ok {
 				exitCode = w.ExitStatus()
 			}
+		}
+		if err := shutdownStore(rootCmd); err != nil {
+			logrus.Warnf("failed to shutdown storage: %q", err)
 		}
 		os.Exit(exitCode)
 	}
--- a/tests/bud_overlay_leaks.bats
+++ b/tests/bud_overlay_leaks.bats
@@ -1,0 +1,18 @@
+#!/usr/bin/env bats
+
+load helpers
+
+@test "bud overlay storage leaked mount" {
+  if test \! -e /usr/bin/fuse-overlayfs -a "$BUILDAH_ISOLATION" = "rootless"; then
+    skip "BUILDAH_ISOLATION = $BUILDAH_ISOLATION" and no /usr/bin/fuse-overlayfs present
+  fi
+
+  target=pull
+  run_buildah 125 --storage-driver=overlay bud --signature-policy ${TESTSDIR}/policy.json -t ${target} --pull-never ${TESTSDIR}/bud/pull
+  expect_output --substring "no such image"
+
+  leftover=$(mount | grep $TESTDIR | cat)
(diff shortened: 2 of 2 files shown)


rhatdan:
Fixes: https://github.com/containers/buildah/issues/1991

rhatdan:
@TomSweeneyRedHat @edsantiago @vrothberg @nalind @ashley-cui @QiWang19 PTAL

cmd/buildah/main.go
```diff
@@ -168,14 +168,24 @@ func before(cmd *cobra.Command) error {
...
-func after(cmd *cobra.Command) error {
+func shutdownStore(cmd *cobra.Command) error {
 	if needToShutdownStore {
 		store, err := getStore(cmd)
 		if err != nil {
 			return err
 		}
+		logrus.Debugf("shutting down the store")
+		needToShutdownStore = false
 		_, _ = store.Shutdown(false)
```
vrothberg, line 179: Why are we ignoring the error? If it's intended, can you add a comment?

TomSweeneyRedHat:
Once @vrothberg 's concerns are addressed, LGTM

rhatdan:
Fixes: https://github.com/containers/buildah/issues/2030

cmd/buildah/main.go
rhatdan: Not my code...

Anyways, I did not want to risk making this a fatal error, so since it was ignored before, I will only treat it as an error.

rhatdan:
@vrothberg PTAL and merge.

rhatdan:
@mheon @ashley-cui @QiWang19 @edsantiago PTAL

edsantiago:
I'd like a test also. It should be easy, and shouldn't even require a new test: maybe add `mount | grep $TESTDIR` and barf if it succeeds, somewhere around here in the `buildah pull never` test:

https://github.com/containers/buildah/blob/190c51e8ec7afb2eaa870fb616c4d6ca45ea3097/tests/from.bats#L281

rhatdan:
The problem is this leak is not happening with vfs, only with overlay.  So the test never fails.

tests/bud_overlay_leaks.bats
```diff
@@ -0,0 +1,16 @@
...
+@test "bud overlay storage leaked mount" {
+  if test \! -e /usr/bin/fuse-overlayfs -a "$BUILDAH_ISOLATION" = "rootless"; then
+    skip "BUILDAH_ISOLATION = $BUILDAH_ISOLATION" and no /usr/bin/fuse-overlayfs present
+  fi
+
+  target=pull
+  run_buildah 125 --storage-driver=overlay bud --signature-policy ${TESTSDIR}/policy.json -t ${target} --pull-never ${TESTSDIR}/bud/pull
+  expect_output --substring "no such image"
+  run bash -c "mount | grep $TESTDIR"
+  [ "$status" -ne 0 ]
```
edsantiago, line 14: I know this is normal BATS, but I really despise it: when a test fails in this manner, I find myself spending a lot of time trying to figure out what went wrong, why it's an error. Would you consider:
```bash
  leftover=$(mount | grep $TESTDIR)
  if [ -n "$leftover" ]; then
    die "buildah leaked a mount on error: $leftover"
  fi
```

rhatdan: Sure

tests/bud_overlay_leaks.bats
```diff
@@ -0,0 +1,18 @@
...
+@test "bud overlay storage leaked mount" {
+  if test \! -e /usr/bin/fuse-overlayfs -a "$BUILDAH_ISOLATION" = "rootless"; then
+    skip "BUILDAH_ISOLATION = $BUILDAH_ISOLATION" and no /usr/bin/fuse-overlayfs present
+  fi
+
+  target=pull
+  run_buildah 125 --storage-driver=overlay bud --signature-policy ${TESTSDIR}/policy.json -t ${target} --pull-never ${TESTSDIR}/bud/pull
+  expect_output --substring "no such image"
+
+  leftover=$(mount | grep $TESTDIR)
```
edsantiago, line 14: I edited my comment, but was too late: this is going to fail, because the `grep` will fail (under normal circumstances), which will cause BATS to exit. Could you add `| cat` after the grep?
```bash
   leftover=$(mount | grep $TESTDIR | cat)
```
Sorry for not getting it right the first time.

rhatdan: done

edsantiago:
Tests green. Thank you for adding the new one. LGTM.

QiWang19:
LGTM

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
Behave bors
bors retry

PR #2534, opened 2020-08-11
Add buildah.IsContainer interface

rhatdan:
This interface will allow callers to specify a storage container id
and the store and return whether or not the container is a buildah
container.

We want to add the ability to for Podman to identify containers in storage
that it did not create or were created via podman build command.

Podman will use this information to help user identify where the container
came from.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/util.go
+++ b/util.go
@@ -420,3 +420,21 @@ func ReserveSELinuxLabels(store storage.Store, id string) error {
 	}
 	return nil
 }
+
+// IsContainer identifies if the specified container id is a buildah container
+// in the specified store.
+func IsContainer(id string, store storage.Store) (bool, error) {
+	cdir, err := store.ContainerDirectory(id)
+	if err != nil {
+		return false, err
+	}
+	// Assuming that if the stateFile exists, that this is a Buildah
+	// container.
+	if _, err = os.Stat(filepath.Join(cdir, stateFile)); err != nil {
+		if os.IsNotExist(err) {
+			return false, nil
+		}
+		return false, errors.Wrapf(err, "error stating %q", filepath.Join(cdir, stateFile))
+	}
+	return true, nil
+}


rhatdan:
@mheon @vrothberg @TomSweeneyRedHat @QiWang19 @ashley-cui @nalind PTAL

util.go
```diff
@@ -420,3 +420,19 @@ func ReserveSELinuxLabels(store storage.Store, id string) error {
...
 }
+
+// IsContainer identifies if the specified container id is a buildah container
+// in the specified store.
+func IsContainer(id string, store storage.Store) (bool, error) {
+	cdir, err := store.ContainerDirectory(id)
+	if err != nil {
+		return false, err
+	}
+	if _, err = ioutil.ReadFile(filepath.Join(cdir, stateFile)); err != nil {
```
mheon, line 431: Is it necessary to read the file, given you're not doing anything with the contents? Feels like this should either try a decode to verify it's a valid Buildah state, or just use os.Stat

nalind:
Does this provide information that an error from `OpenBuilder()` doesn't?  That function checks the contents of `stateFile` as well.

util.go
rhatdan: A stat would probably be enough.

rhatdan:
I don't think we need to verify the state is good by processing the file, because a corrupted statefile would probably tell me that I have a buildah image that has a corrupted file.  I still think it is logical to identify this as a buildah image, at least for the use case I want under `podman ps -a`

rhatdan:
Updated
@mheon @vrothberg @nalind @TomSweeneyRedHat @ashley-cui @QiWang19 PTAL

mheon:
LGTM on my end

QiWang19:
LGTM

PR #2536, opened 2020-08-11
Avoid feeding run_buildah to pipe

edsantiago:
The usage 'run_buildah ... | grep' (or pipe whatever) works,
but it's a super bad pattern. Replace all instances of it
with a one-two of 'run_buildah' and 'expect_output'. Some
of these needed a little minor massaging.

Also: 'run_buildah ... || true' -> 'run_buildah 125 ...'.
I don't review all buildah PRs, so this one slipped by me.

Also: clean up trailing whitespace

Digression: why is 'run_buildah | grep' bad? Because:
  - it is inaccurate. run_buildah does 'echo $output',
    but it also emits other output (the buildah command
    itself, and possible expected-status mismatch).
    It is possible that a pipe-grep could trigger
    on one of these.
  - the reason run_buildah emits these things is so
    a human can look at debug output on failure and
    recognize what command was run, what the output
    was. If we pipe-grep, we lose that.
  - it is possible that a pipe-grep will mask
    a failing run_buildah (I'm not sure about this)
  - expect_output is more precise anyway, hence
    makes a better test.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -1237,7 +1237,7 @@ function _test_http() {
   _prefetch ubuntu
   imgName=ubuntu-image
   ctrName=ubuntu-copy
-  run_buildah bud --signature-policy ${TESTSDIR}/policy.json -f ${TESTSDIR}/bud/copy-multistage-paths/Dockerfile.invalid_from -t ${imgName} ${TESTSDIR}/bud/copy-multistage-paths || true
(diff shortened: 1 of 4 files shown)


rhatdan:
Nice work
LGTM

rhatdan:
@vrothberg @TomSweeneyRedHat @nalind @QiWang19 @ashley-cui PTAL

edsantiago:
Followup: I used `grep 'run.*|' tests/*.bats` to find these instances. All remaining instances are of the form `sh -c something-pipe-something`, which is perfectly fine.

TomSweeneyRedHat:
LGTM

PR #2539, opened 2020-08-12
StageExecutor.historyMatches: compare timestamps using .Equal

nalind:
#### What type of PR is this?

/kind bug

As the time package warns us, we should have used `time.Time.Equal()` to compare timestamps in image histories when checking for equality, instead of trying to compare values using the `==` operator.

#### How to verify it

Check if we can evaluate caches for images with non-UTC ("Z") `Created` values in history entries.

None

As of this writing, `docker.io/shanemcd/ansible-runner` is such an image, causing cache evaluation to fail to match anything based on it.

```
No
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -1080,6 +1080,31 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 	return imgID, ref, nil
 }
 
+func historyEntriesEqual(base, derived v1.History) bool {
+	if base.CreatedBy != derived.CreatedBy {
+		return false
+	}
+	if base.Comment != derived.Comment {
+		return false
+	}
+	if base.Author != derived.Author {
+		return false
+	}
+	if base.EmptyLayer != derived.EmptyLayer {
+		return false
+	}
+	if base.Created != nil && derived.Created == nil {
+		return false
+	}
+	if base.Created == nil && derived.Created != nil {
+		return false
+	}
+	if base.Created != nil && derived.Created != nil && !base.Created.Equal(*derived.Created) {
+		return false
+	}
+	return true
+}
+
 // historyMatches returns true if a candidate history matches the history of our
 // base image (if we have one), plus the current instruction.
 // Used to verify whether a cache of the intermediate image exists and whether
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

imagebuildah/stage_executor.go
```diff
@@ -1092,25 +1117,7 @@ func (s *StageExecutor) historyMatches(baseHistory []v1.History, child *parser.N
...
-			return false
-		}
-		if baseHistory[i].Created != nil && history[i].Created == nil {
-			return false
-		}
-		if baseHistory[i].Created == nil && history[i].Created != nil {
-			return false
-		}
-		if baseHistory[i].Created != nil && history[i].Created != nil && *baseHistory[i].Created != *history[i].Created {
+		if !historyEntriesEqual(baseHistory[i], history[i]) {
```
TomSweeneyRedHat, line 1120: lint doesn't like something in this file, I'm not sure what though, it was a rather vague error, or perhaps I've just been staring at this screen too long today.

TomSweeneyRedHat: It's my eyes most likely, it's not happy with the test file, lint is OK with this.

imagebuildah/stage_executor_test.go
```diff
@@ -0,0 +1,90 @@
...
+			a:     `{"created_by":"actions"}`,
+			b:     `{"created_by":"actions"}`,
+			equal: true,
+		},
+		{
+			a:     `{"created_by":"jiggery"}`,
+			b:     `{"created_by":"pokery"}`,
+			equal: false,
+		},
+	}
```
TomSweeneyRedHat, line 78: Is casing significant?  If so, might want to add a test or two.

TomSweeneyRedHat:
One question and unhappy tests, otherwise LGTM.

imagebuildah/stage_executor_test.go
nalind: `encoding/json`'s godoc for `Unmarshal()` says it prefers an exact match, but will accept a case-insensitive one.

TomSweeneyRedHat: thanks!  I thought that might be the case, but I wasn't sure.  I appreciate you running that down.

imagebuildah/stage_executor.go
nalind: Piping the file through `goimports -d` indicates that it wanted a stylistic change.  And there was a great deal of swearing in my home.

PR #2543, opened 2020-08-13
run: propagate intermediate mountns mounts

### LINKED ISSUE #2524: buildah run fails when using --runtime=kata-runtime
**Description**

Running any command with `buildah run --runtime=kata-runtime ...` results in an error (see example below).

Running Podman with `--runtime=kata-runtime` (`create`, `run` etc.) works fine.

Using Kata when executing less-trusted binaries via `buildah run` is desirable to take advantage of the reduced attack surface of virtualisation vs. containerisation.

1. `container="$(sudo buildah from registry.fedoraproject.org/fedora)"`
2. `sudo buildah run --runtime=kata-runtime "$container" ls`

Commands are run with `sudo` since Kata can't run rootless [yet](https://github.com/kata-containers/runtime/issues/2132.

```
rpc error: code = Internal desc = Could not run process: container_linux.go:349: starting container process caused "exec: \"/usr/bin/ls\": stat /usr/bin/ls: no such file or directory"
error running container: error creating container for [/usr/bin/ls]: : exit status 1
```

Listing of `/`.

```
buildah-1.15.1-1.fc32.x86_64
```

giuseppe:
propagate mounts happening in the intermediate mount namespace to the
host so they are accessible from kata-runtime.

Closes: https://github.com/containers/buildah/issues/2524

#### What type of PR is this?

/kind bug

so we can use kata-runtime

#### How to verify it

follow the reproducer here: https://github.com/containers/buildah/issues/2524

Fixes #2524

```release-note
buildah run works with kata-runtime
```

### DIFF
--- a/bind/mount.go
+++ b/bind/mount.go
@@ -36,11 +36,6 @@ func SetupIntermediateMountNamespace(spec *specs.Spec, bundlePath string) (unmou
 		return nil, errors.Wrapf(err, "error creating new mount namespace for %v", spec.Process.Args)
 	}
 
-	// Make all of our mounts private to our namespace.
-	if err := mount.MakeRPrivate("/"); err != nil {
-		return nil, errors.Wrapf(err, "error making mounts private to mount namespace for %v", spec.Process.Args)
-	}
-
 	// Make sure the bundle directory is searchable.  We created it with
 	// TempDir(), so it should have started with permissions set to 0700.
 	info, err := os.Stat(bundlePath)


bind/mount.go
```diff
@@ -36,11 +36,6 @@ func SetupIntermediateMountNamespace(spec *specs.Spec, bundlePath string) (unmou
 		return nil, errors.Wrapf(err, "error creating new mount namespace for %v", spec.Process.Args)
 	}
 
-	// Make all of our mounts private to our namespace.
-	if err := mount.MakeRPrivate("/"); err != nil {
```
rhatdan, line 40: Should this be a MakeRSlave?

We don't want mounts inside of the container to leak to the host.  If the host is SHARED, won't they leak?

giuseppe: if we make it rslave, the mounts will still be unaccessible from the outside.

This is needed to propagate the rootfs and I think it is equivalent to what we do in Podman/CRI-O where the .  The OCI runtime will still set the root mount to rslave/private.

Leaking mounts though is a possibility, let me see if I can address it in a different way..

rhatdan:
Also looks like unit tests are failing.

giuseppe:
I don't think this fix will be enough anyway.  There are also other issues with kata as it leaves processes around and the `wait4(pid,` hangs forever.

PR #2547, opened 2020-08-13
WIP: commit: optionally create layers for ADD/COPY/RUN/WORKDIR

### LINKED ISSUE #2402: buildah bud --layers=false / does not handle WORKDIR correctly.
If I have a Dockerfile that looks like:

```
FROM fedora
WORKDIR /test
```
And then run `buildah bud -t workdir --layers=false .` on it multiple times.  The second time, it will not create the /test directory in the image.
```
podman run workdir pwd
Error: chdir: No such file or directory: OCI runtime command not found error
```
If I run `buildah bud -t workdir --layers=true .`
 ```
podman run workdir pwd
/test
```

### PR DESCRIPTION
**Pull request by nalind**
#### What type of PR is this?

/kind bug

In multi-layer mode, only add layers when ADD, COPY, RUN, or WORKDIR make changes to the container's rootfs.

#### How to verify it

Building an image using `--layers=true` and a Dockerfile containing
```
FROM fedora
WORKDIR /testdir
```
should produce an image with two layers, rather than just the base image's single layer.

Conformance tests have been extended to compare how many layers we generate in comparison to `docker build`, though it makes exceptions for items we create as targets for bind mounts when handling RUN instructions.

Fixes #2402

The `workdir-in-image-owner` and `workdir-not-in-image` conformance tests should exercise this.

```
Instead of always creating new layers for ADD, COPY, and RUN instructions, `buildah build-using-dockerfile --layers` will now avoid adding new layers when those instructions make no changes to the filesystem.
```

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -79,6 +79,9 @@ type CommitOptions struct {
 	// EmptyLayer tells the builder to omit the diff for the working
 	// container.
 	EmptyLayer bool
+	// EmptyLayerIfRedundant tells the builder to compute the diff for the
+	// working container, and to omit the layer if the diff is empty.
+	EmptyLayerIfRedundant bool
 	// OmitTimestamp forces epoch 0 as created timestamp to allow for
 	// deterministic, content-addressable builds.
 	// Deprecated use HistoryTimestamp instead.
--- a/image.go
+++ b/image.go
@@ -63,7 +63,8 @@ type containerImageRef struct {
 	exporting             bool
 	squash                bool
 	emptyLayer            bool
-	idMappingOptions      *define.IDMappingOptions
+	emptyLayerIfRedundant bool
+	idMappingOptions      *IDMappingOptions
 	parent                string
 	blobDirectory         string
 	preEmptyLayers        []v1.History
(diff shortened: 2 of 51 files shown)

### CODE REVIEW

**Comment by rhatdan** [comment]
What happens if I have
```
FROM fedora
WORKDIR /home/dwalsh
MAINTAINER dan walsh
```

nalind:
That doesn't work right.  Moving this to a draft PR.

rhatdan:
@nalind Any movement on this one?  Needs a rebase.

rhatdan:
@nalind Maybe we should just get this merged since it closes one of the cases.  Then we can continue to work on the corner cases.

nalind:
/retitle commit: optionally create layers for ADD/COPY/RUN/WORKDIR

nalind:
I expect conformance tests to fail on overlay because resetting the contents of a directory causes us to mark it as opaque until #2756 changes how we do things for those cases.

rhatdan:
@nalind Is this still being worked on or can we close it?

nalind:
We still need this.

rhatdan:
@nalind can you rebase and repush?

rhatdan:
@nalind rebase and repush.

rhatdan:
@nalind this oldy-but-goody PR needs a rebase.

rhatdan:
@nalind what is going on with this PR?

nalind:
A simple rebase isn't going to be enough here - it didn't pass all of the tests before it needed that.  Re-adding the WIP label.

PR #2556, opened 2020-08-20
copier: add Mkdir()

nalind:
#### What type of PR is this?

/kind feature

Add a function to the copier package for doing Mkdir-possibly-in-a-chroot, for ensuring that a directory exists without having to possibly create it directly from outside of a chroot.  It also attempts to address some review comments from #2486:

* Check for ERANGE instead of E2BIG errors from llistxattr() and lgetxattr() as indicators that the buffer we passed to them is too small.
* Factor an `isRelevantXattr` helper out of Lgetxattrs() and Lsetxattrs().
* Drop our getcwd() function in favor of using os.Getwd().
* Adjust the comment describing the GetOptions.KeepDirectoryNames field.
* Clean hdr.Name before attempting to compute where an item being extracted should go.
* When writing items to the filesystem, create with 0?00 permissions, set ownership, then set the correct permissions.
* Merge StatResponse.Error, GetResponse.Error, and PutResponse.Error into Response.Error.
* When reading items from the filesystem, if a glob matches multiple items, and one of the items is excluded, continue checking the other items.
* Make sure we always Wait() on a child process if we spawned one.
* Clean up the cleanup logic for pipes that we use to communicate with a child process.
* Clean up the kill-the-child-process logic we call when we encounter an error communicating with the child process.
* Drop the separate Options structure, use helper methods to simplify pulling out the right ID maps and exclusions list for the request.

#### How to verify it:

There are new and updated unit tests in the package.  They should all pass.

None

This is another chunk of #2480.

```
None
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -59,46 +59,91 @@ func isArchivePath(path string) bool {
 type RequestType string
 
 const (
-	RequestStat RequestType = "STAT"
-	RequestGet  RequestType = "GET"
-	RequestPut  RequestType = "PUT"
-	RequestQuit RequestType = "QUIT"
+	RequestStat  RequestType = "STAT"
+	RequestGet   RequestType = "GET"
+	RequestPut   RequestType = "PUT"
+	RequestMkdir RequestType = "MKDIR"
+	RequestQuit  RequestType = "QUIT"
 )
-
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM
@vrothberg @giuseppe @mtrmac PTAL

PR #2493, opened 2020-07-27
WIP: Fix release v1.14

cevich:
#### What type of PR is this?

/kind failing-test 

#### What this PR does / why we need it:

Bring the `release-v1.14` branch up to more modern CI design, similar to master where VM's have packages pre-installed.

#### How to verify it

CI testing will pass

#### Which issue(s) this PR fixes:

Fixes #2416

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #2558, opened 2020-08-20
[release-v1.14] CI in_podman: don't get tripped up by CIRRUS_CHANGE_TITLE

nalind:
#### What type of PR is this?

/kind bug

In CI, we don't handle passing environment values that contain whitespace through to tests that we run inside of podman, so we need to screen out $CIRRUS_CHANGE_TITLE.

#### How to verify it

Check if the `in_podman` task passes for this PR in CI, give or take other problems we have in this branch.

None

Cherry picked from #2553.

```
None
```


TomSweeneyRedHat:
LGTM

rhatdan:
LGTM

rhatdan:
LGTM
but tests are not happy.

TomSweeneyRedHat:
@nalind lint issue, and I could have sworn we'd made those changes to .cirrus.yml in prior run throughs...

nalind:
@TomSweeneyRedHat I hadn't pulled in changes from `.golangci.yml` yet, giving that a go.

TomSweeneyRedHat:
@nalind, I just remembered, we had worked through these, but then my PR got closed and @cevich has been struggling with it in #2493.  We may want to punt, just to do the title change then ask @edsantiago to run the tests against it "by hand".

contrib/cirrus/lib.sh
```diff
@@ -203,6 +203,7 @@ in_podman() {
     for envname in $(awk 'BEGIN{for(v in ENVIRON) print v}' | \
                      egrep "$envrx" | \
                      egrep -v "CIRRUS_.+_MESSAGE" | \
+                     egrep -v "CIRRUS_.+_TITLE" | \
```
edsantiago, line 206: OK, I'm a day late and dollar short -- I didn't even know buildah was using Cirrus -- but this makes me cringe. Do I understand correctly that the problem you're trying to address is that (something somewhere) can't handle envariables with spaces in the value? If so, this implementation is super fragile. May I recommend instead:
```bash
    if ! [[ $x =~ " " ]]; then    (add to list)
```
or, possibly even better, declare `envargs` as an array and pass it through that way:
```bash
    declare -a envargs=()
    ...
    envargs+=("-e" "$envname=$envvalue")
    ...
    podman run ... "${envargs[@]}"
```

nalind: That is a more correct fix, I completely agree, but the one-line change we added to the master branch was smaller, and not being the person we'll call if/when it breaks, I'm reluctant to make big changes there.
My goal with the maintenance branches is to try to avoid diverging from later branches, so if we're going to do a more correct fix, it needs to land in master first.

edsantiago: Point taken, thanks. But in case this isn't well known: the reason you are dealing with `CIRRUS_COMMIT_TITLE` today is that the fine folks at cirrus-ci added it yesterday, at my request. There is absolutely nothing preventing them from adding new envariables, with any values, at any time. This might perhaps merit a tracking issue.

nalind: Agreed.

rhatdan:
@nalind do we still need this or can we close it?

nalind:
The "cron" invocations in CI are still failing for this branch, so there's work to be done, yes.

cevich:
> "cron" invocations in CI are still failing for this branch

Thanks for keeping an eye on that @nalind .  I'm sorry I haven't been able to be more attentive :disappointed:

rhatdan:
@nalind @cevich is this still something we care about?

cevich:
I think this can probably safely be closed, it would likely be a painful rebase at best.

PR #2568, opened 2020-08-25
build-using-dockerfile: add a hidden --log-rusage flag

nalind:
#### What type of PR is this?

/kind feature

We were seeing some unexplained slowdowns in builds that ran in OpenShift when the memory controller was configured to limit the amount of memory available to us.  When pulling an image, either for use as a base or to supply content for COPY --from, we write all of the image's blobs to temporary files as they're fetched in parallel, read them back in order to create the image's layers on disk in order, and then delete the temporary files.  The working theory is that when these writes-then-reads-then-deletes don't al fit within the kernel's cache (the amount of memory available for use by the cache being subject to control group limits), the data ends up having to be flushed all the way to disk and then read back from disk, and the additional I/O is what makes the build slower.  Being able to log how much I/O is done for a particular step during a build would be very helpful for diagnosing these cases.

This PR adds a flag to `imagebuildah.BuildOptions` that will tell us to log timing and I/O information at each step of the build process, and which enables it if the CLI is run with the `--log-rusage` flag.

#### How to verify it

We add an integration test to ensure that we see output that matches our expected output formatting string when we use the flag.

None

```
If `buildah build-with-dockerfile` is given the `--log-rusage` flag, resource usage information will be logged for every step.
```

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -356,6 +356,7 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budOptions) error {
 		TransientMounts:         iopts.Volumes,
 		OciDecryptConfig:        decConfig,
 		Jobs:                    &iopts.Jobs,
+		LogRusage:               iopts.LogRusage,
 	}
 
 	if iopts.Quiet {
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -183,6 +183,8 @@ type BuildOptions struct {
 	OciDecryptConfig *encconfig.DecryptConfig
 	// Jobs is the number of stages to run in parallel.  If not specified it defaults to 1.
 	Jobs *int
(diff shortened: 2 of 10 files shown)


rhatdan:
LGTM
Although I am not sure we should expose this option.  Can we just leave it hidden and undocumented.

rhatdan:
@TomSweeneyRedHat @ashley-cui @QiWang19 @mheon @vrothberg PTAL

adambkaplan:
LGTM

For OpenShift builds we will probably expose this at log level 6 or 8 (an equivalent of exposing debug/trace logs from buildah).

TomSweeneyRedHat:
LGTM
It would be nice to document this somewhere.  I agree with hiding the flags for general use, but I don't want this option forgotten 2 years from now.  We almost need a "Buildah Black Magic Cookbook" doc somewhere.

PR #2571, opened 2020-08-26
Use --timestamp rather then --omit-timestamp

rhatdan:
We recieved feedback on the --omit-timestamp that
users would rather specify the timestamp seconds
rather then just use EPOCH.

This PR removes --omit-timestamp from buildah bud
since this has never been released.

We also hide --omit-timestamp from buildah commit
and allow users to continue to use it, but it conflicts
with --timestamp.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -168,9 +168,9 @@ type BuildOptions struct {
 	SignBy string
 	// Architecture specifies the target architecture of the image to be built.
 	Architecture string
-	// OmitTimestamp forces epoch 0 as created timestamp to allow for
-	// deterministic, content-addressable builds.
-	OmitTimestamp bool
+	// Timestamp sets the created timestamp to the specified time, allowing
+	// for deterministic, content-addressable builds.
+	Timestamp *time.Time
 	// OS is the specifies the operating system of the image to be built.
 	OS string
 	// MaxPullPushRetries is the maximum number of attempts we'll make to pull or push any one
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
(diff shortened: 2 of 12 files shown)


rhatdan:
@nalind @vrothberg @saschagrunert @mrunalp @mheon @baude @QiWang19 @ashley-cui @giuseppe PTAL

imagebuildah/build.go
```diff
@@ -168,9 +168,9 @@ type BuildOptions struct {
 	SignBy string
 	// Architecture specifies the target architecture of the image to be built.
 	Architecture string
-	// OmitTimestamp forces epoch 0 as created timestamp to allow for
+	// Timestamp forces epoch 0 as created timestamp to allow for
```
nalind, line 171: This comment doesn't look right.

nalind:
Minor nit about the documentation of BuildOptions.Timestamp, otherwise LGTM.

TomSweeneyRedHat:
LGTM other than the comment nit that @nalind spotted.

imagebuildah/build.go
rhatdan: Fixed

rhatdan:
@TomSweeneyRedHat Fixed
Once this is merged, can you start cutting v1.16.0

TomSweeneyRedHat:
LGTM
I want to get @saschagrunert #2572 in too before 1.16.  Will look later today if any other vendor dancing is needed.

nalind:
Uh oh, this looks like it's forcing the dates on all content in new layers to be set to the commit-time by default, which is _not_ what we wanted.

rhatdan:
@nalind Can you suggest a fix?

nalind:
@rhatdan #2576 should have fixed that, thanks!

PR #2572, opened 2020-08-28
Switch to containers/common for seccomp

saschagrunert:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:
seccomp/containers-golang is deprecated and we should switch to
containers/common/pkg/seccomp instead.
#### How to verify it
Check if the CI passes.
#### Which issue(s) this PR fixes:

None
#### Special notes for your reviewer:
None
#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #2573, opened 2020-08-28
Heed our retry delay option values when retrying commit/pull/push

nalind:
#### What type of PR is this?

/kind bug

Pass in our own API values for retry delays to common's retry package when we use it to handle retrying image pull/commit/push operations.

#### How to verify it

Attempt to pull an image from a registry which can't be reached, which is considered a transient error that we'll retry on:
  `buildah pull --debug 127.0.0.0/bogus`
Our CLI's default is to retry with a fixed delay of 2 seconds after each failed attempt, while the retry package's default behavior is to use progressively longer delays after each attempt, so if each attempt retries after 2 seconds, the change is having the desired effect.

None

This undoes a behavior change we'd otherwise have made compared to 1.15.x.

```
None
```

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -344,7 +344,7 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 	}
 
 	var manifestBytes []byte
-	if manifestBytes, err = retryCopyImage(ctx, policyContext, maybeCachedDest, maybeCachedSrc, dest, getCopyOptions(b.store, options.ReportWriter, nil, systemContext, "", false, options.SignBy, options.OciEncryptLayers, options.OciEncryptConfig, nil), options.MaxRetries); err != nil {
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
@saschagrunert has bumped c/common to v0.21.0 in #2572.  I'm going to merge that then take this one for your other changes.  Any objections?

TomSweeneyRedHat:
And I just verified that he has c/storage 1.23.3 there too.

nalind:
No objections.  I'm really only after reconnecting the retry delay setting, which also requires v0.21.0.

nalind:
Rebased.

saschagrunert (approved):
LGTM

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
bors retry

PR #2576, opened 2020-08-31
containerImageRef.NewImageSource(): don't always force timestamps

nalind:
#### What type of PR is this?

/kind bug

Since #2571, we've been accidentally setting the timestamp on all of the content in new layers to the current date when the `--timestamp` flag was not specified, which is a conformance break.

#### How to verify it

Multiple conformance tests, including `TestConformance/add`, flagged new differences in the mtime on various bits of content in layers.  They should be passing again.

None

```
None
```


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM assuming happy tests

PR #2577, opened 2020-09-01
WIP DO NOT MERGE - Make stderr red color

TomSweeneyRedHat:
When an error is thrown inside the container, like one from
`ls /fakefile`, the return string should be colored red when
doing a bud or run action.  Currently it's coming black as red.

Addresses Bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=1661593

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -245,7 +245,6 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 			options.CNIPluginPath = util.DefaultCNIPluginPath
 		}
 	}
-
 	switch isolation {
 	case IsolationOCI:
 		var moreCreateArgs []string
@@ -704,7 +703,7 @@ func runUsingRuntime(isolation Isolation, options RunOptions, configureNetwork b
 	var errorFds, closeBeforeReadingErrorFds []int
 	stdioPipe := make([][]int, 3)
 	copyConsole := false
-	copyPipes := false
+        copyPipes := false
 	finishCopy := make([]int, 2)
 	if err = unix.Pipe(finishCopy); err != nil {
 		return 1, errors.Wrapf(err, "error creating pipe for notifying to stop stdio")
@@ -835,7 +834,6 @@ func runUsingRuntime(isolation Isolation, options RunOptions, configureNetwork b
 		}
 		stderr.Close()
 	}
-
 	// Handle stdio for the container in the background.
 	stdio.Add(1)
 	go runCopyStdio(&stdio, copyPipes, stdioPipe, copyConsole, consoleListener, finishCopy, finishedCopy, spec)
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
I've put this PR up mostly so I can more easily ask @nalind some questions.

run_linux.go
```diff
@@ -1180,6 +1191,30 @@ func runCopyStdio(stdio *sync.WaitGroup, copyPipes bool, stdioPipe [][]int, copy
...
+				logrus.Warnf("error setting terminal state: %v", err)
+			} else {
+				defer func() {
+					if err = terminal.Restore(unix.Stdout, state); err != nil {
+						logrus.Errorf("unable to restore terminal state: %v", err)
+					}
+				}()
+			}
+		}
+		//TOM: End
```
TomSweeneyRedHat, line 1217: @nalind, What I'm trying to do is to get the output of stderr from within the container to be returned back in `relayBuffer[unix.Stderr]`.  Currently and with these changes in play, it's always coming back in `relayBuffer[unix.Stdout]`.  I thought my changes above at 1170 would cure that, do you know if I'm off or I'm just not specifying them correctly?  Here's a snippet of the output I'm seeing with these changes in play:

```
# buildah run tom-working-container ls /fake
TOM: relayMap    All    [map[0:12 2:12 12:1]]
                                             TOM: relayBuffer stderr []
                                                                       TOM: relayBuffer stdout [ls: /fake: No such file or directory]
            TOM: Top drain output buffers
                                         ls: /fake: No such file or directory
```

Any sage advice?  If I can get stderr via the RelayBuffer, I could then color the output in theory, I just can't seem to get stderr there though.

nalind: `relayMap` is used for deciding whiere data read from a given descriptor needs to go (which descriptor it needs to be written to), with the map keyed by the descriptor it's being read from.  `relayBuffer` is used for buffering data that hasn't been written to that destination yet, with the map keyed by the descriptor that it needs to be written to.  I wouldn't expect us to read from our own stderr descriptor, so setting `relayMap[unix.Stderr]` looks wrong.

The current logic routes the child process's stderr to our own stderr through `relayBuffer[unix.Stderr]`, but that only gets used when we have different descriptors to read the child's stdout and stderr data from (so, when we're using pipes).  When we're using a pseudoterminal, we only have one descriptor (`terminalFD`) from which to read both stdout and stderr combined, and we send all of the output to our stdout by queuing it up in `relayBuffer[unix.Stdout]`.

TomSweeneyRedHat: @nalind thanks for the 411.  When I did `podman run $ctr /bin/sh` I was seeing the pipes in action with the separation.  You're right with the terminalFD in play, I wasn't seeing the separation.  So it sounds like the reason I couldn't find a solution while using the pseudoterminal, is there isn't a way to do so with that approach.  

Do you have any thoughts on how I might accomplish this, or is the risk of change to much for the gain of the coloring?

nalind: To intercept only the contained process's stderr stream, we'd need to get between the process running in the container and the pseudoterminal that it's using for stdio, and preferably without interfering with its attempts to access the device using calls like `ioctl()` and `tcgetattr()`/`tcsetattr()`.  Using ptrace to do it would require capabilities that we don't usually have when we're run in a container, and I don't have a good idea of how much work would be required.

rhatdan:
Are we just going to give up on this one?

rhatdan:
@TomSweeneyRedHat Nalin and I talked about this yesterday and came to the agreement to remove the pseudo-terminal from `buildah bud` for the remote commands.  This should allow us to separate the stdin from stdout, and then you could add your magic to make stderr red.
When doing `buildah run`, the pseudo-terminal would continue to be used, so the red stderr, would be disabled, which I believe is the correct behaviour.

We have eliminated stdin from `buildah bud` already so there is no interaction between the users terminal and the build session. 

Could you start working on this PR again to see if removing the pseudo terminal makes it work?

rhatdan:
This is still something we want...

rhatdan:
@TomSweeneyRedHat Any progress?

PR #2581, opened 2020-09-03
fix build on 32bit arches

lsm5:
Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>

#### What type of PR is this?

> /kind other

#### What this PR does / why we need it:

fixes build on i686 and armv7hl

#### How to verify it

build without the change:
https://koji.fedoraproject.org/koji/taskinfo?taskID=50696956

scratch build with the change:
https://koji.fedoraproject.org/koji/taskinfo?taskID=50700329

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

None

#### Does this PR introduce a user-facing change?

None

```release-note
None
```

@nalind @TomSweeneyRedHat @rhatdan @ashcrow PTAL

PR #2582, opened 2020-09-03
CI: expand cross-compile checks

nalind:
#### What type of PR is this?

/kind other

Prompted by #2581, expand the list of OS/architecture combinations we check when we're checking if we can successfully be cross-compiled, and fix up a couple of places that wouldn't pass otherwise.

#### How to verify it

The "cross" test should be building more OS/architecture combinations than it did before.

None

For the sake of podman, we'll want to add `darwin.arm` and `darwin.arm64` sooner or later.

```
None
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -176,7 +176,7 @@ gce_instance:
 
     env:
         matrix:
-            CROSS_TARGET: bin/buildah.darwin
+            CROSS_TARGET: cross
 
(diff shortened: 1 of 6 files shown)


rhatdan:
LGTM
@mheon @TomSweeneyRedHat @QiWang19 @ashley-cui @vrothberg PTAL

vrothberg (approved):
LGTM

Elegant solution using make macros!

PR #2585, opened 2020-09-04
add CI jobs to check for 32bit arch builds

lsm5:
This commit adds 2 new Cirrus tasks to build for linux/arm and linux/386
os/arch combos. This should hopefully avoid situations where we notice
32bit failures only after a fedora koji build.

@cevich @nalind @TomSweeneyRedHat @edsantiago PTAL

#### What type of PR is this?

 /kind other

This PR will enable additional cirrus tasks to check if the source builds fine on 386 and arm arches.

#### How to verify it

CI should show 2 additional tasks for 386 and arm builds.

None

None

```release-note
None
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -21,6 +21,8 @@ env:
     IN_PODMAN: 'false'
     # Not cross-compiling by default
     CROSS_TARGET: ""
+    # Not building for other arches by default
+    GOARCH: ""
 
     ####
     #### Cache-image names to test with
@@ -177,13 +179,18 @@ gce_instance:
     env:
         matrix:
             CROSS_TARGET: cross
+        matrix:
+          - env:
+            GOARCH: "arm"
+          - env:
+            GOARCH: "386"
 
     setup_script: '${SCRIPT_BASE}/setup.sh |& ${_TIMESTAMP}'
     build_script: '${SCRIPT_BASE}/build.sh |& ${_TIMESTAMP}'
 
     binary_artifacts:
         path: ./bin/*
-
+          
 
 'cirrus-ci/required/testing_task':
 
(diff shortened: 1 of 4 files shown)


lsm5:
hmmm, seems wrong, way too many new tasks, fixing :D

lsm5:
alright, I see new jobs for 386 and arm now :crossed_fingers:

cevich (changes requested):
Since the tasks are identical except for the env. var, I suggest using the matrix feature to generate instead of duplicating the task blocks.  Something like this:

```yaml
'cirrus-ci/only_prs/cross_task':

    skip: $CIRRUS_BRANCH =~ ".*\.tmp"

    depends_on:
      - 'cirrus-ci/only_prs/gate'
      - 'cirrus-ci/only_prs/vendor'

    matrix:
        - env:
              GOARCH: "arm"
        - env:
              GOARCH: "386"

    setup_script: '${SCRIPT_BASE}/setup.sh |& ${_TIMESTAMP}'
    build_script: '${SCRIPT_BASE}/build.sh |& ${_TIMESTAMP}'

    binary_artifacts:
      path: ./bin/*
```

Does that make sense?

cevich:
Other changes LGTM.  Nice work @lsm5

lsm5:
> 'cirrus-ci/only_prs/cross_task':
> 
>     skip: $CIRRUS_BRANCH =~ ".*\.tmp"
> 
>     depends_on:
>       - 'cirrus-ci/only_prs/gate'
>       - 'cirrus-ci/only_prs/vendor'
> 
>     matrix:
>         - env:
>               GOARCH: "arm"
>         - env:
>               GOARCH: "386"
> 
In the current jobs, I see `matrix` is under `env`. Is it ok to have `env` under `matrix` as well? And when you say `cross_task` above, does it mean I can include this in the current cross_task which mentions the job for buildah.darwin? Or does this need to be a new cross_task with a different suitable name?

lsm5:
ah looks like #2582 handles this already.

cevich:
> ah looks like #2582 handles this already.

Yes, but your changes here would run the builds in parallel.  

For your own knowledge, I'll answer your other questions as well:

> In the current jobs, I see matrix is under env. Is it ok to have env under matrix as well?

Yes, what's happening with `matrix` is, cirrus automatically creates duplicate tasks with the nested changes merged/overwriting whatever is in the top-level of the task.  So in this case, it's adding/changing the value of an environment variable for each copy of the task it makes.

> And when you say cross_task above

Woops, this was my bad / name collision.  My example was intended to be taken at face-value.  Though perhaps they could all be combined into a single cross-build task, I didn't check what would be required to make that work.

PR #2595, opened 2020-09-08
contrib/cirrus/lib.sh: don't use CN for the hostname

nalind:
#### What type of PR is this?

/kind failing-test 

When generating a certificate to be used by a registry server that we're bringing up, instead of setting a CN value in the subject name to "localhost" to pass the name check that a client makes, use a subject alternative name extension.

The "compare the hostname we're given with a CN from the subject name field" method stops working if the client is built with Go 1.15.

#### How to verify it

Tests which passed before should continue to pass, even when run using a binary built with Go 1.15.

None

```
None
```

### DIFF
--- a/contrib/cirrus/lib.sh
+++ b/contrib/cirrus/lib.sh
@@ -304,7 +304,8 @@ execute_local_registry() {
     mkdir -p $authdirpath
     openssl req \
         -newkey rsa:4096 -nodes -sha256 -x509 -days 2 \
-        -subj "/C=US/ST=Foo/L=Bar/O=Red Hat, Inc./CN=localhost" \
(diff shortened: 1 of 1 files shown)


nalind:
@edsantiago, I think this'll fix an error we're hitting in Fedora's gating tests.

edsantiago:
You beat me to it, thank you! Running it through the wringer now, with equivalent changes in gating-test scripts

nalind:
For reference, https://github.com/golang/go/issues/39568#issuecomment-671424481.  The whole issue, really, but that comment has some useful advice.

edsantiago:
LGTM. Clean run of `authenticate.bats` with this. I will push the same change on the fedpkg side. 

Thank you for jumping on this so quickly and for the helpful annotations.

PR #2596, opened 2020-09-08
bud.bats: use absolute paths in newly-added tests

edsantiago:
Gating tests run in a temp directory; relative paths (bud/foo/bar)
will pass in CI but cause gating-test failures.

Solution is trivial: always use `${TESTSDIR}/bud/etc`

#### What type of PR is this?

/kind failing-test 

Fedora gating tests are failing on rawhide with errors like:
```
$ /usr/bin/buildah bud --timestamp=0 --quiet --pull=false --signature-policy /usr/share/buildah/test/system/policy.json -t timestamp -f Dockerfile.1 bud/cache-stages
error evaluating symlinks in build context path: lstat /var/str/bud: no such file or directory
[ rc=125 (** EXPECTED 0 **) ]
```

#### How to verify it

Run `bats /path/to/bud.bats` while cd'ed to a directory other than the git checkout

None

None

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -2182,7 +2182,7 @@ EOM
 
 @test "bud timestamp" {
   _prefetch alpine
-  run_buildah bud --timestamp=0 --quiet --pull=false --signature-policy ${TESTSDIR}/policy.json -t timestamp -f Dockerfile.1 bud/cache-stages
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM

PR #2598, opened 2020-09-08
copier.Get(): hard link targets shouldn't be relative paths

nalind:
#### What type of PR is this?

/kind bug

When adding a hard link entry to an archive, don't try to make the path of the link's target name be relative to the link's location in the filesystem if they're in the same directory, since receivers don't interpret it that way.

Flagged by one of podman's e2e tests.  h/t @rhatdan

#### How to verify it

Added an integration test, but using squashing (our `--squash`, or podman's `--squash-all`) to trigger the right code path, and then having a file with two or more links, all in the same subdirectory below the root directory, should trigger the error path.

None

```
Fixes a regression where attempting to "squash" an image while committing it could fail.
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -1178,14 +1178,11 @@ func copierHandlerGetOne(srcfi os.FileInfo, symlinkTarget, name, contentPath str
 		target := hardlinkChecker.Check(srcfi)
 		if target != "" {
 			hdr.Typeflag = tar.TypeLink
-			if filepath.Dir(filepath.Clean(string(os.PathSeparator)+name)) == filepath.Dir(target) {
-				target = filepath.Base(target)
-			}
 			hdr.Linkname = filepath.ToSlash(target)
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
LGTM

rhatdan:
I tested this in Podman against the failing tests, and it fixes the issue.

LGTM

bors r+
@TomSweeneyRedHat I think we should cut a buildah release with this fix, from the stable branch. v1.16.1

rhatdan:
Saving the planet.

TomSweeneyRedHat:
v1.16.1 sounds good to me.  I'll plan on something after we wrap up bug week this week @rhatdan.

rhatdan:
Well this is about bug week, I believe, since a few of the Podman bugs have to do with podman build being slow.  We need to get buildah updated in podman to close these issues/bugzillas.

PR #2603, opened 2020-09-09
[release-1.16] copier.Get(): hard link targets shouldn't be relative paths

nalind:
#### What type of PR is this?

/kind bug

When adding a hard link entry to an archive, don't try to make the path of the link's target name be relative to the link's location in the filesystem if they're in the same directory, since receivers don't interpret it that way.

Flagged by one of podman's e2e tests.  h/t @rhatdan

#### How to verify it

Added an integration test, but using squashing (our `--squash`, or podman's `--squash-all`) to trigger the right code path, and then having a file with two or more links, all in the same subdirectory below the root directory, should trigger the error path.

None

Cherry-picked from #2598.

```
Fixes a regression in 1.16.0 where attempting to "squash" an image while committing it could fail.
```


rhatdan:
LGTM

PR #2631, opened 2020-09-17
Cirrus: Use images from automation_images

cevich:
***Depends on:*** #2885

#### What type of PR is this?

> /kind deprecation

Previously, VM Images were built from a side-band process tacked onto `containers/podman` automation.  This has recently been split off into it's own repository `containers/automation_images`.  This PR makes use of freshly built images produced using the new workflow.  Additionally, to support testing of a `runc -> crun` transition, both packages are installed in the newly referenced images.

#### How to verify it

Automated tests will pass

None

None

None

### DIFF
--- a/contrib/cirrus/build.sh
+++ b/contrib/cirrus/build.sh
@@ -4,7 +4,7 @@ set -e
 
 source $(dirname $0)/lib.sh
 
-req_env_var IN_PODMAN IN_PODMAN_NAME GOSRC
+req_env_vars IN_PODMAN IN_PODMAN_NAME GOSRC
 
 remove_packaged_buildah_files
 
@@ -16,7 +16,7 @@ then
     in_podman --rm $IN_PODMAN_NAME $0
 else
     echo "Compiling buildah (\$GOSRC=$GOSRC)"
-    showrun make clean ${CROSS_TARGET:-all} ${CROSS_TARGET:+CGO_ENABLED=0}
+    showrun make clean ${CROSS_TARGET:-all} ${CROSS_TARGET:+CGO_ENABLED=0 cross}
 
     echo "Installing buildah"
     mkdir -p bin
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -33,11 +33,13 @@ env:
     UBUNTU_NAME: "ubuntu-20"
     PRIOR_UBUNTU_NAME: "ubuntu-19"
 
-    _BUILT_IMAGE_SUFFIX: "podman-6530021898584064"
-    FEDORA_CACHE_IMAGE_NAME: "${FEDORA_NAME}-${_BUILT_IMAGE_SUFFIX}"
-    PRIOR_FEDORA_CACHE_IMAGE_NAME: "${PRIOR_FEDORA_NAME}-${_BUILT_IMAGE_SUFFIX}"
-    UBUNTU_CACHE_IMAGE_NAME: "${UBUNTU_NAME}-${_BUILT_IMAGE_SUFFIX}"
-    PRIOR_UBUNTU_CACHE_IMAGE_NAME: "${PRIOR_UBUNTU_NAME}-${_BUILT_IMAGE_SUFFIX}"
+    IMAGE_SUFFIX: "podman-6530021898584064"
+    FEDORA_CACHE_IMAGE_NAME: "${FEDORA_NAME}-${IMAGE_SUFFIX}"
+    PRIOR_FEDORA_CACHE_IMAGE_NAME: "${PRIOR_FEDORA_NAME}-${IMAGE_SUFFIX}"
+    UBUNTU_CACHE_IMAGE_NAME: "${UBUNTU_NAME}-${IMAGE_SUFFIX}"
+    PRIOR_UBUNTU_CACHE_IMAGE_NAME: "${PRIOR_UBUNTU_NAME}-${IMAGE_SUFFIX}"
+
+    IN_PODMAN_IMAGE: "quay.io/libpod/fedora_podman:${IMAGE_SUFFIX}"
 
     ####
     #### Command variables to help avoid duplication
@@ -62,10 +64,12 @@ gce_instance:
 
 
 # Update metadata on VM images referenced by this repository state
-'cirrus-ci/only_prs/meta_task':
+meta_task:
+    name: "VM img. keepalive"
+    alias: meta
 
     container:
-        image: "quay.io/libpod/imgts:latest"  # see contrib/imgts
+        image: "quay.io/libpod/imgts:${IMAGE_SUFFIX}"  # see contrib/imgts
         cpu: 1
         memory: 1
 
@@ -86,7 +90,10 @@ gce_instance:
     script: '/usr/local/bin/entrypoint.sh |& ${_TIMESTAMP}'
 
 
-'cirrus-ci/only_prs/gate_task':
+smoke_task:
+    alias: 'smoke'
+    name: "Smoke Test"
+
     gce_instance:
         memory: "12Gb"
 
@@ -104,7 +111,13 @@ gce_instance:
         path: ./bin/*
 
 
-'cirrus-ci/only_prs/unit_task':
+unit_task:
+    name: "Unit tests"
+    alias: unit
+
+    depends_on:
+      - smoke
+      - vendor
 
     timeout_in: 45m
 
@@ -116,7 +129,13 @@ gce_instance:
         path: ./bin/*
 
 
-'cirrus-ci/only_prs/conformance_task':
+conformance_task:
+    name: "Docker Build Conformance"
+    alias: conformance
+
+    depends_on:
+      - unit
+
     gce_instance:  # Only need to specify differences from defaults (above)
         image_name: "${UBUNTU_CACHE_IMAGE_NAME}"
 
@@ -126,11 +145,11 @@ gce_instance:
     conformance_test_script: '${SCRIPT_BASE}/test.sh conformance |& ${_TIMESTAMP}'
 
 
-# This task runs `make vendor` followed by ./hack/tree_status.sh to check
-# whether the git tree is clean.  The reasoning for that is to make sure
-# that the vendor.conf, the code and the vendored packages in ./vendor are
-# in sync at all times.
-'cirrus-ci/only_prs/vendor_task':
+# Check that all included go modules from other sources match
+# # what is expected in `vendor/modules.txt` vs `go.mod`.
+vendor_task:
+    name: "Test Vendoring"
+    alias: vendor
 
     env:
         CIRRUS_WORKING_DIR: "/var/tmp/go/src/github.com/containers/buildah"
@@ -150,11 +169,13 @@ gce_instance:
         - './hack/tree_status.sh'
 
 
-'cirrus-ci/only_prs/cross_task':
+# Confirm cross-compile ALL archetectures on a Mac OS-X VM.
+cross_build_task:
+    name: "Cross Compile"
+    alias: cross_build
 
     depends_on:
-        - 'cirrus-ci/only_prs/gate'
-        - 'cirrus-ci/only_prs/vendor'
+      - unit
 
     env:
         matrix:
@@ -167,18 +188,29 @@ gce_instance:
         path: ./bin/*
 
 
-'cirrus-ci/required/testing_task':
+integration_task:
+    name: "Integration $DISTRO_NV"
+    alias: integration
 
     depends_on:
-        - 'cirrus-ci/only_prs/gate'
-        - 'cirrus-ci/only_prs/vendor'
+      - unit
 
-    gce_instance:  # Only need to specify differences from defaults (above)
-        matrix:  # Duplicate this task for each matrix product.
-            image_name: "${FEDORA_CACHE_IMAGE_NAME}"
-            image_name: "${PRIOR_FEDORA_CACHE_IMAGE_NAME}"
-            image_name: "${UBUNTU_CACHE_IMAGE_NAME}"
-#            image_name: "${PRIOR_UBUNTU_CACHE_IMAGE_NAME}"
+    matrix:
+        - env:
+            DISTRO_NV: "${FEDORA_NAME}"
+            IMAGE_NAME: "${FEDORA_CACHE_IMAGE_NAME}"
+        - env:
+            DISTRO_NV: "${PRIOR_FEDORA_NAME}"
+            IMAGE_NAME: "${PRIOR_FEDORA_CACHE_IMAGE_NAME}"
+        - env:
+            DISTRO_NV: "${UBUNTU_NAME}"
+            IMAGE_NAME: "${UBUNTU_CACHE_IMAGE_NAME}"
+        - env:
+            DISTRO_NV: "${PRIOR_UBUNTU_NAME}"
+            IMAGE_NAME: "${PRIOR_UBUNTU_CACHE_IMAGE_NAME}"
+
+    gce_instance:
+        image_name: "$IMAGE_NAME"
 
     # Separate scripts for separate outputs, makes debugging easier.
     setup_script: '${SCRIPT_BASE}/setup.sh |& ${_TIMESTAMP}'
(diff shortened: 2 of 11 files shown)


cevich:
@rhatdan or @edsantiago any chance you've seen these test failures before?  The only change I've made here is to update the VM images (Which also means having both runc and crun installed).  Any hints of where I should look/poke would be appreciated.

edsantiago:
Haven't seen those exact failures, but the `fchmod` in the first one looks exactly like https://github.com/containers/crun/issues/496 whose last entry is @rhatdan opening a PR to fix the problem, but I don't know where (I assume maybe seccomp has a repo somewhere?)

cevich:
Thx @edsantiago I'm seeing similar errors in podman-land (also with updated images).

cevich:
```unknown seccomp syscall `fchmodat2` ignored``` is what I see coming from podman.

**Comment by rhatdan** [comment]
The fchmodat2 error should not be blocking. I have containers-common update coming that removes this from the seccomp.json file which should remove the warning.

**Comment by rhatdan** [comment]
@cevich What is going on with this PR?

**Comment by cevich** [comment]
> What is going on with this PR?

I was waiting for new VM images with updated crun packages...then I forgot about this PR :smile:

**Comment by cevich** [comment]
@edsantiago @rhatdan @nalind when you get a moment.  I'm trying to update buildah CI to more closely match the workings of podman, including utilizing new VM images.  However the latest Ubuntu 19 and 20 images are showing matching test failures.  Have you seen this one before:

https://cirrus-ci.com/task/5711150744600576?command=integration_test#L1014-L1036

**Comment by cevich** [comment]
Retrying with latest master (prior run was several commits behind)

**Comment by cevich** [comment]
> What are the differences between the images we're currently using and the ones we're about to switch to?

Lots of little things, but the biggest one is: We're running updates-testing packages for the Fedora's, but there is no equivalent for Ubuntu.  I'll double-check the versions of key packages...

**Comment by cevich** [comment]
**Fedora release 32 (Thirty Two)**
Kernel:  5.8.12-200.fc32.x86_64
Cgroups:  cgroup2fs
conmon-2.0.21-2.fc32-x86_64
containernetworking-plugins-0.8.7-1.fc32-x86_64
containers-common-1.1.1-5.fc32-x86_64
container-selinux-2.145.0-1.fc32-noarch
criu-3.14-1.fc32-x86_64
crun-0.15-5.fc32-x86_64
golang-1.14.9-1.fc32-x86_64
libseccomp-2.5.0-3.fc32-x86_64
podman-2.1.1-7.fc32-x86_64
runc-1.0.0-144.dev.gite6555cc.fc32-x86_64
skopeo-1.1.1-5.fc32-x86_64
slirp4netns-1.1.4-1.fc32-x86_64

**Ubuntu 20.04 LTS**
Kernel:  5.4.0-1009-gcp
Cgroups:  tmpfs
conmon-2.0.20~1-amd64
containernetworking-plugins-0.8.7~1-amd64
containers-common-1.1.1~2-all
cri-o-runc-1.0.0-6-amd64
criu-3.14-1ppa1.20.04-amd64
libseccomp2-2.4.3-1ubuntu3.20.04.3-amd64
podman-2.1.1~1-amd64
skopeo-1.1.1~2-amd64
slirp4netns-1.1.4~10-amd64

So it seems conmon is slightly older, along with containers-common and libseccomp. I see there is an update available for containers-common (to 1.2.0~1+nmu1) @nalind could one/more of those be responsible?

**Comment by nalind** [comment]
The error messages are being produced when runc tries to set up the seccomp filter according to its configuration, so I'd look at either libseccomp or the kernel as a variable, though it's possible the libseccomp bindings for Go that it's using are also a factor.

The `~` character toggles strikethrough formatting in markdown.

**Comment by nalind** [comment]
Do we know what the differences are between the currently-used Ubuntu images, which don't trigger this error, and the new ones, which do?

**Comment by cevich** [comment]
> Do we know what the differences are between the currently-used Ubuntu images, which don't trigger this error, and the new ones, which do?

Oh thanks for the suggestion, that's a good idea.  I'll get those details...

**Comment by cevich** [comment]
Latest master uses-

**Ubuntu 20.04.1 LTS**

```
Kernel:  5.4.0-1009-gcp
Cgroups:  tmpfs
dpkg-query: no packages found matching golang
conmon-2.0.18~1-amd64
containernetworking-plugins-0.8.6~1-amd64
containers-common-1.1.0~2-all
cri-o-runc-1.0.0-3~dev2-amd64
criu-3.14-1ppa1.20.04-amd64
crun-0.13~2-amd64
libseccomp2-2.4.3-1ubuntu3.20.04.3-amd64
podman-2.0.3~1-amd64
skopeo-1.1.0~2-amd64
slirp4netns-1.1.4~2-amd64
```

So curiously, libseccomp2 is the exact same version, but everything else is older compared to (new image):
```
conmon-2.0.201-amd64
containernetworking-plugins-0.8.71-amd64
containers-common-1.1.12-all
cri-o-runc-1.0.0-6-amd64
slirp4netns-1.1.410-amd64
```

@nalind  IIRC the seccomp configuration comes from `containers-common`, no?  The crun version is also behind the new images.  There was a seccomp problem we ran into with podman tests the other week, necessitating these new VM images.  I believe I asked at the time, and the conclusion was it wouldn't affect buildah/tests.

What can I do next, to help debug this?

**Comment by nalind** [comment]
The configuration we use is in `containers-common`, yes, so it's possible that changes we made to seccomp.json, which are accepted by the newer libseccomp in Fedora, are being rejected by the older version.

**Comment by rhatdan** [comment]
seccomp.json has a few new syscalls, that would probably not be known by an older libseccomp.

**Comment by cevich** [comment]
okay, so then it seems like the latest versions of libseccomp and containers-common currently in OBS don't play nicely together (for the tests).  Who manages libseccomp packaing in OBS for us?

Based on what's in Fedora, I'm guessing we need libseccomp > 2.5.0

**Comment by cevich** [comment]
Update: Jindrich is going to try following Lokesh's notes and build us a newer Ubuntu libseccomp package.

**Comment by cevich** [comment]
Rebase/force-push w/ commit to use google's transparent container image mirror.  I don't expect different results, unless the transparent mirror config breaks.

**Comment by cevich** [comment]
***Status***:  This PR is waiting on @lsm5 / Jindrich to build/make available newer libseccomp2 Ubuntu packages.

**Comment by cevich** [comment]
Status: The newer libseccomp2 Ubuntu packages are only available for 20.04 and 20.10.  This PR is now waiting on either:

* [Depreciation of `prior-ubuntu` for all containers automation](https://github.com/containers/automation_images/pull/31)
* [Support Ubuntu minor version numbers for all containers automation](https://github.com/containers/automation_images/pull/34)

**Comment by cevich** [comment]
Clarification: The option where two 20.x Ubuntu's are supported _requires_ also moving from 19.10 -> 20.10 because the former is EOL and will (very soon) stop being able to update/install packages.  I know that at least `containers/storage` does runtime installation (which will break).  So it's best to put the effort in now, to bring in the new Ubuntu version.

**Comment by cevich** [comment]
@nalind @rhatdan I'm trying to update the VM Image here, bringing in F33 and Ubuntu 20.10.  Under Fedora 33, the unit-tests are failing for some reason, [throwing errors about `$XDG_CONFIG_HOME or `$HOME`.](https://cirrus-ci.com/task/6397769139093504?command=unit_test#L4919).  Have you ever seen this before or have ideas on what I could try next?

***Note:*** `hack/get_ci_vm.sh` is currently broken on this PR, I intend to work on fixing it net week.

**Comment by cevich** [comment]
Update: Found where we forcibly set $HOME and realize it's not an "exported" variable.  Pushing fix for that, assuming the error message is accurate...

**Comment by cevich** [comment]
@rhatdan @nalind @edsantiago PTAL when you have a moment...I've got the tests with these new Ubuntu 20.04 and 20.10 images running well.  It seems only these crypto-related tests are failing.  Any hints as to what might be breaking?

**Comment by cevich** [comment]
Ref: [Ubuntu Crypto-test failures](https://cirrus-ci.com/build/5325596941090816)

**Comment by cevich** [comment]
Ooof, no change :cry: 

**note:*** I'm ignoring the `Containerized Integration` and `Docker Buld Conformance` tasks for now since the integration tests all seem to be experiencing a relatively common set of test failures.

**Comment by edsantiago** [comment]
This is a little too complicated and jumbly to really review (like, get_ci_vm, does that really belong here?) In looking at the individual commits, I wonder if you could try removing c4018045256687308c54d424b37e073e07a6e483 ? #2844 should've already taken care of the dockerpocalypse. It might be easier to track down the failures if the PR is separated into smaller parts. (And maybe we'll get lucky, and the failures won't happen then).

**Comment by cevich** [comment]
> This is a little too complicated and jumbly to really review (like, get_ci_vm, does that really belong here?)

I bet it is :disappointed:  Though a lot of the changes are necessary to support running the latest generation of VM images.  The `get_ci_vm` changes are needed, as it's otherwise broken and useful for debugging problems in this PR.  I hear what you're saying though, and can probably make a pass at cleaning things up a bit.  I'm going to leave that at something for next year though.

**Comment by cevich** [comment]
@edsantiago okay, I did a massive re-arranging/cleanup of the commits and they flow much better now.  However, I appear to be hitting something related to the https://github.com/containers/buildah/issues/2857 fix:

```level=error msg="The storage 'driver' option must be set in /etc/containers/storage.conf, guarantee proper operation."```

I'm hesitant to put the workaround back given it's age.  Are you aware of another fix?

**Comment by edsantiago** [comment]
Ugh. This is #2860, on hold since last year waiting for @lsm5 to return from PTO (next week). Before I spend any time looking further, could you try rebasing? (Github says you need to anyway). This should have been addressed in #2842.

**Comment by cevich** [comment]
> Before I spend any time looking further, could you try rebasing?

Happy to, I have some fixups that need to be incorporated anyway.

**Comment by cevich** [comment]
@edsantiago looks like we've got pretty consistent results now.  So the package-update is needed for the Ubuntu's I take it?  Any idea what's up with the Fedora's capabilities tests?

**Comment by edsantiago** [comment]
Oh nooooo.... 
```
expected: 'CapBnd:	00000000a80425fb'
actual:   'CapBnd:	00000000800405fb'
```
Does this have a `crun` built after 2020-09 or thereabouts? 0.14.1-5 or greater? All the post-test steps failed, including package_versions, with a `req_env_vars` barf. If `crun` is up-to-date, we're going to need @giuseppe's help.

**Comment by edsantiago** [comment]
No, I remember now. We had to do some super-special magic to deal with capability differences: 

https://github.com/containers/podman/pull/7538/files#diff-33137dafae9d49e95f87e755ee682424c0cb7b11854f97269153cf8460de51a0

We're going to need the same thing here, bitwise arithmetic in bash. This does not look like fun.

**Comment by edsantiago** [comment]
And again, no, this isn't the new-kernel thing. It looks like buildah default capabilities have changed, in particular, cap_net_raw, cap_mknod, and cap_audit_write are gone:
```console
$ cdif -u <(capsh --decode=00000000a80425fb) <(capsh --decode=00000000800405fb)
```
```diff
-0x00000000a80425fb=cap_chown,...,cap_net_bind_service,cap_net_raw,cap_sys_chroot,cap_mknod,cap_audit_write,cap_setfcap
+0x00000000800405fb=cap_chown,...,cap_net_bind_service,            cap_sys_chroot,                          cap_setfcap
```
(whitespace and deletions for readability). This rings a lot of bells, I remember recent conversations relating to this, but can't remember the details nor find anything relevant in git logs. @rhatdan do you have a sense for what might be going on here? (See previous two comments).

edsantiago:
Giving up for the day. I think we need a better understanding of these VMs. Have you run podman integration tests on them? Those too include hardcoded `a80425fb`, so I would expect them to fail.

cevich:
> bitwise arithmetic in bash

Python3 is universally available, as is perl.  Aren't ***you*** the one always telling ***me*** to use the best/correct tool for the task at hand :stuck_out_tongue_winking_eye:

cevich:
> Have you run podman integration tests on them?

Not yet, but it's a good suggestion.  These VMs are newer than what's currently in-use on podman master.

cevich:
Okay, same `c6524344056676352` images [deployed in this podman PR.](https://github.com/containers/podman/pull/8899)

cevich:
So it appears the images used here do pass podman's CI checks.  @edsantiago What about if I/you/we changed the condition used in this test to a binary match after applying a bit-mask?

(Just a guess, since I don't understand what this failing test is actually checking for)

edsantiago:
I'm not even looking at this any more until @giuseppe or @rhatdan chime in. This is way above my skill level.

giuseppe:
> It looks like buildah default capabilities have changed, in particular, cap_net_raw, cap_mknod, and cap_audit_write are gone:

I think it is a consequence of: https://github.com/containers/common/pull/319

edsantiago:
That sure looks like it. We have so many repos, I forgot to check containers-common. Thank you!

So, now what?

giuseppe:
> So, now what?

I'd just tweak the `bud capabilities test` test for the time being to always add these capabilities, so it works both with the old default and the new default:

```diff
diff --git a/tests/bud.bats b/tests/bud.bats
index 293eacf7..8972e799 100644
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -2274,7 +2274,7 @@ EOM
 
 @test "bud capabilities test" {
   _prefetch busybox
-  run_buildah bud -t testcap --signature-policy ${TESTSDIR}/policy.json -f ${TESTSDIR}/bud/capabilities/Dockerfile
+  run_buildah bud --cap-add cap_net_raw,cap_mknod,cap_audit_write -t testcap --signature-policy ${TESTSDIR}/policy.json -f ${TESTSDIR}/bud/capabilities/Dockerfile
   expect_output --substring "uid=3267"
   expect_output --substring "CapBnd:   00000000a80425fb"
   expect_output --substring "CapEff:   0000000000000000"
```

cevich:
Thanks @giuseppe and @edsantiago for the help, that seems to have helped.  I found/fixed another WRT the 'time' error message in an additional commit.  Besides the "in_podman" tests, I believe the only other problem is the 'storage.conf must have driver option' problem related to #2842

cevich:
Okay, I think I've got it now (thanks again everybody that helped).

edsantiago:
Much cleaner task map, thank you! Will review over the course of today. 
cirrus-map-pr2631

cevich:
> Will review over the course of today.

@edsantiago thanks...but please hold of a few hours at least, in case a test fails for some reason and I need to toss in an (unanticipated) last minute fix.

cevich:
Yay! `CI agent stopped responding!` :cry:

cevich:
@nalind poke.  I think the "Docker Build Conformance" test is finding an actual problem now (with COPY and ADD).  These failures are new as of my update to use docker 20.10 for this test.

cevich:
Nalin said:

> it looks like docker build is giving us a "file does not exist" (usually os.ErrNotExist) error message when we expect "no such file or directory" (usually syscall.ENOENT). i see them as equivalent, so we probably want to update the test to accept either message

edsantiago:
@cevich tests are green, is this ready for review?

cevich:
> tests are green, is this ready for review?> 

Holy crap, my go-gibberish worked on the first try?!?!?!

Yes, review away, thanks for waiting.

@nalind PTAL at the `Fix conformance test false-failures` commit.  I only updated the bare-minimum necessary for Docker 20.10, but there are a bunch of other places that could have similar regex updates.  I did not examine them in detail, but am happy to update them here if you see a need.

cevich:
Ref: VM Images are from https://github.com/containers/automation_images/pull/44

contrib/cirrus/build.sh
```diff
@@ -16,7 +16,7 @@ then
     in_podman --rm $IN_PODMAN_NAME $0
 else
     echo "Compiling buildah (\$GOSRC=$GOSRC)"
-    showrun make clean ${CROSS_TARGET:-all} ${CROSS_TARGET:+CGO_ENABLED=0}
+    showrun make clean ${CROSS_TARGET:-all} ${CROSS_TARGET:+CGO_ENABLED=0 cross}
```
cevich, line 19: Note: This is fugly.  Followup beautification for the cross-compile test is here: https://github.com/containers/buildah/pull/2886

edsantiago (approved):
Looks good on first pass. I did not review `hack/get_ci_vm`.

For a future update, please remove `contrib/cirrus/lib.sh.t` and remove the `1` in  https://github.com/containers/buildah/blob/cfb3372985533411fe91249c5162e6a82a13fe20/contrib/cirrus/logcollector.sh#L51

Otherwise LGTM.

/lgtm
/hold

cevich:
> For a future update, please remove

Oh fudge, you're right.  There's no way I'll remember to do that...okay, made a note to take care of it in a separate PR.

cevich:
> For a future update, please remove

Ref: Incorporated these into https://github.com/containers/buildah/pull/2886/commits

PR #2630, opened 2020-09-16
Add(): fix handling of relative paths with no ContextDir

nalind:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

`Add()` is supposed to handle relative paths when the `ContextDir` value passed to it is not set, but it hasn't been doing that correctly since it was overhauled.  Correct it to recognize relative paths at the function start, when `ContextDir` is not set, and combining them with the current directory to convert them to absolute paths.

#### How to verify it

Added an integration test which should cover the reported cases.

#### Which issue(s) this PR fixes:

Fixes #2619.

#### Special notes for your reviewer:

I expect to backport this to the `release-1.16` branch once it passes.

#### Does this PR introduce a user-facing change?

```
The regression in `buildah add` which caused it to not properly handle source locations specified using relative paths should be fixed.
```

PR #2633, opened 2020-09-17
[release-1.16] Add(): fix handling of relative paths with no ContextDir

### LINKED ISSUE #2619: buildah add fails to resolve path
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Description**
Either buildah add fails to resolve path or it writes a wrong error message. In an error message is said that the passed path could not be resolved relative to `/` even if the path is a rigth absolute path but without starting `/`.  Also, it seems unusual that buildah add resolvs relative path relative to `/` instead of the current working directory.

1. Create a container: `buildah from alt`
2. Get the container id: `buildah containers`
3. Create a file: `touch /root/test.txt`
4. Try to add the file to the container: `buildah add a6ed3927b913 root/test.txt /tmp`

Failing with the following error message:
`error adding content to container "alt-working-container": error computing path of "root/test.txt": Rel: can't make root/test.txt relative to /`

Resolving the file root/test.txt relative to `/` as `/root/txt` and adding the file to the container. 

```
# rpm -q buildah
buildah-1.16.0-alt1.x86_64
```

```
Version:         1.16.0
Go Version:      go1.15.2
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        0.4.0
libcni Version:  v0.7.2-0.20190904153231-83439463f784
image Version:   5.5.2
Git Commit:      295b825cdc0795f7e60349b901880e40e9850f5a
Built:           Wed Sep  9 19:21:15 2020
OS/Arch:         linux/amd64
```

```
ALT Sisyphus Sisyphus (unstable) (sisyphus)
ALT Sisyphus Sisyphus (unstable) (sisyphus)
NAME="Sisyphus"
VERSION="20191026/unstable"
ID=altlinux
VERSION_ID=20191026
PRETTY_NAME="ALT Regular"
ANSI_COLOR="1;33"
CPE_NAME="cpe:/o:alt:sisyphus:20191026"
HOME_URL="http://en.altlinux.org"
BUG_REPORT_URL="https://bugs.altlinux.org/"
ALT Sisyphus Sisyphus (unstable) (sisyphus)
ALT Sisyphus Sisyphus (unstable) (sisyphus)
```

**Output of `uname -a`:**

```
Linux mainvm 5.4.40-std-def-alt2 #1 SMP Tue May 12 09:06:57 UTC 2020 x86_64 GNU/Linux
```

nalind:
#### What type of PR is this?

/kind bug

`Add()` is supposed to handle relative paths when the `ContextDir` value passed to it is not set, but it hasn't been doing that correctly since it was overhauled.  Correct it to recognize relative paths at the function start, when `ContextDir` is not set, and combining them with the current directory to convert them to absolute paths.

#### How to verify it

Added an integration test which should cover the reported cases.

Fixes #2619.

Cherry picked from #2630.

```
The regression in `buildah add` which caused it to not properly handle source locations specified using relative paths should be fixed.
```

### DIFF
--- a/add.go
+++ b/add.go
@@ -151,18 +151,26 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 	}()
 
 	contextDir := options.ContextDir
-	if contextDir == "" {
+	currentDir := options.ContextDir
+	if options.ContextDir == "" {
 		contextDir = string(os.PathSeparator)
+		currentDir, err = os.Getwd()
+		if err != nil {
+			return errors.Wrapf(err, "error determining current working directory")
+		}
 	}
 
 	// Figure out what sorts of sources we have.
 	var localSources, remoteSources []string
-	for _, src := range sources {
+	for i, src := range sources {
 		if sourceIsRemote(src) {
 			remoteSources = append(remoteSources, src)
 			continue
 		}
-		localSources = append(localSources, src)
+		if !filepath.IsAbs(src) && options.ContextDir == "" {
+			sources[i] = filepath.Join(currentDir, src)
+		}
+		localSources = append(localSources, sources[i])
 	}
 
 	// Check how many items our local source specs matched.  Each spec
--- a/tests/add.bats
+++ b/tests/add.bats
@@ -172,3 +172,19 @@ load helpers
   run_buildah add $cid https://github.com/containers/buildah/raw/master/README.md /home
   run_buildah run $cid ls /home/README.md
 }
+
+@test "add relative" {
+  # make sure we don't get thrown by relative source locations
+  _prefetch busybox
+  run_buildah from --quiet --signature-policy ${TESTSDIR}/policy.json busybox
+  cid=$output
+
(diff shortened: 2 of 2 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
Tests are failing @nalind I think it's unrelated, but it's an odd error.

TomSweeneyRedHat:
@nalind and/or @mheon 
Is this a change we should spin up a new release for and work on getting into Podman sooner rather than later?  Is this exception worthy @nalind in your thinking?  It's close to me.

nalind:
I think it's worth tagging a 1.16.2 for the sake of our CLI.  A quick grep for "buildah.Add" doesn't turn up any calls in the podman source tree to `buildah.Add()` except from our imagebuildah package, which always supplies a `ContextDir` value, so I don't _think_ podman is affected.

PR #2635, opened 2020-09-17
Run(): ignore containers.conf's environment configuration

nalind:
#### What type of PR is this?

/kind bug

The hardwired default for containers.conf now includes a TERM variable, and passing it through to commands that we "RUN" during a build can subtly cause the resulting image to be different from one that `docker build` would create, so stop using it there.

When a runtime runs the image we eventually produce, it'll consult the configuration file, so the variable will still be set, even when it isn't set in the image.

#### How to verify it

This change prevents a failure in the `TestConformance/ci-pipeline-modified` conformance test, which checks which variables are set during RUN.

None

This creates an exception to the "environment variables in containers.conf are used everywhere" rule, but while the conformance test is a pretty harmless example of differences in builds that can result from environment variables being set or not, I wouldn't want to have to debug a build problem that was triggered based on whether or not a variable was set in the configuration.

```
Environment variables set in containers.conf will no longer be set for commands run using `buildah run` or by RUN instructions during `buildah build-using-dockerfile`.
```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -91,11 +91,8 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 		return err
 	}
 
-	defaultContainerConfig, err := config.Default()
-	if err != nil {
-		return errors.Wrapf(err, "failed to get container config")
-	}
-	b.configureEnvironment(g, options, defaultContainerConfig.Containers.Env)
+	// hardwire the environment to match docker build to avoid subtle and hard-to-debug differences due to containers.conf
+	b.configureEnvironment(g, options, []string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"})
 
 	if b.CommonBuildOpts == nil {
 		return errors.Errorf("Invalid format on container you must recreate the container")
--- a/tests/containers_conf.bats
+++ b/tests/containers_conf.bats
@@ -2,12 +2,6 @@
 
 load helpers
 
-@test "containers.conf env test" {
-    export CONTAINERS_CONF=${TESTSDIR}/containers.conf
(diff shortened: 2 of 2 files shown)


run_linux.go
```diff
@@ -91,11 +91,7 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 		return err
 	}
 
-	defaultContainerConfig, err := config.Default()
-	if err != nil {
-		return errors.Wrapf(err, "failed to get container config")
-	}
-	b.configureEnvironment(g, options, defaultContainerConfig.Containers.Env)
+	b.configureEnvironment(g, options, []string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"})
```
TomSweeneyRedHat, line 94: Could you add a comment here explaining why we're not using the config.Default here?  Just to avoid someone changing it back one early morning a year from now....

TomSweeneyRedHat:
Changes LGTM, would like a comment added.

run_linux.go
nalind: Added `hardwire the environment to match docker build to avoid subtle and hard-to-debug differences due to containers.conf` above this.  Let me know if you'd prefer a different phrasing.

rhatdan: What if the user wants to have these environment settings to be applied.

COMPANY="Dan Walsh Inc"
I want all of my images to have this environment variable set.

nalind: Do you mean wanting that value included in the image configuration blob?  containers.conf doesn't do that, but there's an ENV instruction for that.  Otherwise, configuration from containers.conf is still used when the image we create is run by something like `podman run`.

nalind:
bors retry

PR #2638, opened 2020-09-18
integration tests: make sure tests run in ${topdir}/tests

nalind:
#### What type of PR is this?

/kind other

Add a pushd/popd to setup()/teardown() to ensure that tests run in our tests directory, even if they're not invoked by test_runner.sh, so that we can reliably use relative paths in tests without having to ensure our location on a per-test or per-test-file basis.

#### How to verify it

If you run `bats` directly and point it to either the `tests` directory, or the `add.bats` file within it, the test shouldn't fail for attempting to access a path relative to the current directory which should actually be relative to the `tests` directory.

None

```
None
```

### DIFF
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -16,6 +16,8 @@ BUILDAH_TIMEOUT=${BUILDAH_TIMEOUT:-300}
 export GPG_TTY=/dev/null
 
 function setup() {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

rhatdan:
@TomSweeneyRedHat @ashley-cui @QiWang19 @vrothberg @giuseppe PTAL

TomSweeneyRedHat:
LGTM

PR #2641, opened 2020-09-21
New "add relative" test: make it work in gating env

edsantiago:
Gating tests are failing because the new add.bats:relative
test assumes it is cd'ed to a known location. This is not
true in gating tests (neither Fedora nor RHEL).

Fix: for this test, 'cd ${TESTSDIR}'. I've confirmed on my
end that this affects only the scope of the current test,
but I don't actually know if that's a BATS guarantee.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/tests/add.bats
+++ b/tests/add.bats
@@ -175,6 +175,8 @@ load helpers
 
 @test "add relative" {
   # make sure we don't get thrown by relative source locations
+  cd ${TESTSDIR}
+
   _prefetch busybox
   run_buildah from --quiet --signature-policy ${TESTSDIR}/policy.json busybox
   cid=$output


rhatdan:
LGTM
@nalind @TomSweeneyRedHat @vrothberg @ashley-cui @QiWang19 PTAL

nalind:
This should be fixed by #2638.

edsantiago:
Ah, thank you, that explains why this test passed in the most recent bodhi!

PR #2642, opened 2020-09-21
New CI check: xref --help vs man pages

edsantiago:
Run 'buildah --help', recursively against all subcommands,
then cross-reference the results against docs/buildah*.md.
Report differences in subcommands and/or flags.

The majority of the changes in this PR are trivial (see
below) but a handful may be controversial and require
careful review:

  * Making 'bud' the default output of 'buildah help',
    with 'build-using-dockerfile' as an alias. This is
    the inverse of the situation until now: buildah
    would list build-using-dockerfile as the primary
    name. The man page, OTOH, lists 'bud'. The source
    file name is 'bud'. I suspect that most people
    type 'bud'. So, for consistency, I choose to make
    'bud' the default visible command.

  * add --encryption-key and --encrypt-layer documentation
    to buildah-commit.md, and --encrypt-layer to -push.md.
    Please double-check the wording here.

  * remove --notruncate from buildah-images.md. The option
    does not exist (although there is a TODO comment in
    the code). If it should exist, it is left to the
    reader to implement. I would humbly suggest that this
    is a good idea, for consistency with buildah containers.

  * remove --shm-size from buildah-pull.md. The option
    does not exist, and I suspect this was a copy-paste error.

  * add --security-opt stub to buildah-run.md. This is
    flagged with FIXME, and I would appreciate someone
    offering suggested wording here.

There is one other decision that needs to be made before this
PR merges (I will update commit message accordingly once done):
buildah-bud.md documents --userns-[gu]id-map which are not
actual buildah-bud option (they are global). I believe they
are documented well enough in buildah.bud, and would like to
remove them. May I? The alternative is to add special-case
code in the script to ignore those entries in bud.md.

Trivial (IMO) changes:

  * split options in man pages, from '**--foo, -f**'
    to '**--foo**, **-f**'. This conforms with the style
    used in podman man pages.

  * add missing one-letter aliases (usually "-q", "-a")

  * add missing man page entries for some easy options

  * sort out-of-order subcommand listings in man pages

Finally, do note that this is a copy-and-alter duplicate of the
original script in podman, and that is horrible. In an ideal
world I would've been able to refactor the podman version into
something usable on both repos (and then more). It turns out the
differences in man page format and in special-case handling are
too broad to let me do a clean refactor.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/cmd/buildah/run.go
+++ b/cmd/buildah/run.go
@@ -14,18 +14,17 @@ import (
 )
 
 type runInputOptions struct {
-	addHistory     bool
-	capAdd         []string
-	capDrop        []string
-	hostname       string
-	isolation      string
-	runtime        string
-	runtimeFlag    []string
-	noPivot        bool
-	securityOption []string
-	terminal       bool
-	volumes        []string
-	mounts         []string
+	addHistory  bool
+	capAdd      []string
+	capDrop     []string
+	hostname    string
+	isolation   string
+	mounts      []string
+	runtime     string
+	runtimeFlag []string
+	noPivot     bool
+	terminal    bool
+	volumes     []string
 	*buildahcli.NameSpaceResults
 }
 
@@ -63,7 +62,6 @@ func init() {
 	flags.StringVar(&opts.runtime, "runtime", "", "`path` to an alternate OCI runtime")
 	flags.StringSliceVar(&opts.runtimeFlag, "runtime-flag", []string{}, "add global flags for the container runtime")
 	flags.BoolVar(&opts.noPivot, "no-pivot", false, "do not use pivot root to jail process inside rootfs")
-	flags.StringArrayVar(&opts.securityOption, "security-opt", []string{}, "security options (default [])")
 	// TODO add-third alias for tty
 	flags.BoolVarP(&opts.terminal, "terminal", "t", false, "allocate a pseudo-TTY in the container")
 	flags.BoolVar(&opts.terminal, "tty", false, "allocate a pseudo-TTY in the container")
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -123,7 +123,7 @@ Limit the container's CPU usage. By default, containers run with the full
 CPU resource. This flag tell the kernel to restrict the container's CPU usage
 to the quota you specify.
 
-**--cpu-shares, -c**=*0*
+**--cpu-shares**, **-c**=*0*
 
 CPU shares (relative weight)
 
@@ -186,7 +186,7 @@ The [key[:passphrase]] to be used for decryption of images. Key can point to key
 
 Add a host device or devices under a directory to the container. The format is `<device-on-host>[:<device-on-container>][:<permissions>]` (e.g. --device=/dev/sdc:/dev/xvdc:rwm)
 
-**--disable-compression, -D**
+**--disable-compression**, **-D**
 Don't compress filesystem layers when building the image unless it is required
 by the location where the image is being written.  This is the default setting,
 because image layers are compressed automatically when they are pushed to
@@ -216,7 +216,7 @@ Set custom DNS options
 
 Set custom DNS search domains
 
-**--file, -f** *Containerfile*
+**--file**, **-f** *Containerfile*
 
 Specifies a Containerfile which contains instructions for building the image,
 either a local file or an **http** or **https** URL.  If more than one
@@ -306,7 +306,7 @@ environment variable. `export BUILDAH_LAYERS=true`
 Log output which would be sent to standard output and standard error to the
 specified file instead of to standard output and standard error.
 
-**--memory, -m**=""
+**--memory**, **-m**=""
 
 Memory limit (format: <number>[<unit>], where unit = b, k, m or g)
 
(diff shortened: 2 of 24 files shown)


edsantiago:
(The CI failure is deliberate until that question is resolved)

docs/buildah-run.md
```diff
@@ -156,6 +156,12 @@ whenever the rootfs is on top of a ramdisk.
 Note: You can make this option the default by setting the BUILDAH\_NOPIVOT
 environment variable.  `export BUILDAH_NOPIVOT=true`
 
+**--security-opt** options
```
rhatdan, line 159: This is totally unimplemented, it looks like this option does nothing.  I think it would be better to remove the option altogether and wait for users to complain.  Or open an issue to actually implement it.

rhatdan (approved):

docs/buildah-run.md
edsantiago: :embarrassed emoji: If I had taken the time to do more than a simple `grep`, I would've detected this.

I have removed the option and will push that once the `--userns-[gu]id-map` question is resolved. Thank you.

docs/buildah-bud.md
```diff
@@ -497,48 +501,6 @@ the user namespace in which `buildah` itself is being run should be reused, or
 it can be the path to an user namespace which is already in use by another
 process.
 
-**--userns-uid-map** *mapping*
```
rhatdan, line 500: This is making me nervous?
Something is going wrong here.
```
$ podman build --help | grep userns
      --userns path                                'container', path of user namespace to join, or 'host'
      --userns-gid-map containerID:hostID:length   containerID:hostID:length GID mapping to use in user namespace
      --userns-gid-map-group name                  name of entries from /etc/subgid to use to set user namespace GID mapping
      --userns-uid-map containerID:hostID:length   containerID:hostID:length UID mapping to use in user namespace
      --userns-uid-map-user name                   name of entries from /etc/subuid to use to set user namespace UID mapping
```
These options are showing up in podman build but not buildah bud.

rhatdan: @nalind @TomSweeneyRedHat @baude Any ideas?

rhatdan: OK I think I have it, the issue is that these flags are global flags not just for podman.

```
./bin/buildah --help | grep userns
      --userns-gid-map ctrID:hostID:length   default ctrID:hostID:length GID mapping to use
      --userns-uid-map ctrID:hostID:length   default ctrID:hostID:length UID mapping to use
```

rhatdan: So we could move this content to the buildah man page.  But I think we should still link it from here.

edsantiago: It already is in the buildah man page: https://github.com/containers/buildah/blame/master/docs/buildah.md#L70-L88

edsantiago: I can, if so desired, add back this section and add a special case for the xref script to ignore it. That was my original question on my initial commit; since I never heard from anyone, I just made an executive decision. If I made the wrong one, I'm happy to revert it.

cmd/buildah/run.go
```diff
@@ -14,18 +14,17 @@ import (
...
+	capAdd      []string
+	capDrop     []string
+	hostname    string
+	isolation   string
+	runtime     string
+	runtimeFlag []string
+	noPivot     bool
+	terminal    bool
+	volumes     []string
+	mounts      []string
```
TomSweeneyRedHat, line 27: I know it's not yours, but everything else in this list is alpha order, could you move `mounts` up if you have other changes?

docs/buildah-bud.md
TomSweeneyRedHat: I'll dive more tomorrow, too much eyestrain at the moment, but this removal doesn't feel right.

TomSweeneyRedHat:
Really good first pass @edsantiago .  I was very excited when you talked about adding this here and am happy to see it near fruition.

cmd/buildah/run.go
edsantiago: nice catch, thanks. Fixed; push pending resolution of the `--userns` question.

docs/buildah-bud.md
rhatdan: I think we should take this section and copy it into the buildah man page.  There is a lot more useful information here on configuring user namespace then is currently in the buildah man page.  I think we should also update the two userns-uid-map-user sections to refernence back to this in the main man page.  Sort of a see also --userns-uid-map in buildah(1) man page.

rhatdan: I opened https://github.com/containers/buildah/pull/2661 to move the man page description around.

TomSweeneyRedHat:
LGTM
and happy green test buttons.
TYVM @edsantiago

rhatdan:
Thanks @edsantiago this will be a big help to maintain quality.

PR #2645, opened 2020-09-23
Change call to Split() to safer SplitN()

TomSweeneyRedHat:
Change calls to Split to SplitN when we are splitting on an "=" sign to make sure
we're not over chopping what the caller requested.

I started this search after seeing https://github.com/containers/podman/pull/7759.
I'll do similar for Podman, but initial results on my search is order of
magnitudes bigger there.


nalind:
LGTM

PR #2646, opened 2020-09-23
chroot: create bind mount targets 0755 instead of 0700

nalind:
#### What type of PR is this?

/kind bug

When running using chroot isolation, create the target mountpoints for bind mounts, when they don't already exist, with 0755 permissions, for better consistency with runc.

#### How to verify it

Build an image derived from an image that doesn't include a `/dev`, for example `registry.suse.co/suse/sles12sp4@sha256:7504649ba71f61492c309ccada877a47ea11272b8658a9ad52923e8d17204628`, using chroot isolation, and examine the layers in the image we produce, comparing the permissions on `/dev`, `/proc`, and `/sys` for inconsistencies.

None

Running conformance tests with different types of isolation would give us a chance to catch this, but I really don't have a strong feeling for whether or not we want to effectively run all of the tests twice every time to be sure.

```
None
```

### DIFF
--- a/chroot/run.go
+++ b/chroot/run.go
@@ -1047,7 +1047,7 @@ func setupChrootBindMounts(spec *specs.Spec, bundlePath string) (undoBinds func(
 	subDev := filepath.Join(spec.Root.Path, "/dev")
 	if err := unix.Mount("/dev", subDev, "bind", devFlags, ""); err != nil {
 		if os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

vrothberg (commented):
LGTM
@cyphar, can you have a look?

cyphar:
LGTM.

nalind:
bors retry

PR #2647, opened 2020-09-24
Set directory ownership when copied with ID mapping

hshiina:
Signed-off-by: Hironori Shiina <shiina.hironori@fujitsu.com>

#### What type of PR is this?

/kind bug

When a directory is copied by `buildah copy` to a container with ID mapping, the directory owner is not set correctly after #2659 was merged. This PR sets the ownership correctly.

This also fixes `idmapping` test to verify this fix. It seems that this test has not been working since the variable `maps` was renamed.

#### How to verify it
Run `bats tests/namespaces.bats`.

None

```release-note
None
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -1271,6 +1271,7 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
 			return errorResponse("copier: put: error mapping container filesystem owner %d:%d to host filesystem owners: %v", dirUID, dirGID, err)
 		}
 		dirUID, dirGID = hostDirPair.UID, hostDirPair.GID
+		defaultDirUID, defaultDirGID = hostDirPair.UID, hostDirPair.GID
 		if req.PutOptions.ChownFiles != nil {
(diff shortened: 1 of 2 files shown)


rhatdan:
Seems the idmappings tests are blowing up?

hshiina:
Yes.
This loop[1] never runs because `maps` is not defined.

[1] https://github.com/containers/buildah/blob/master/tests/namespaces.bats#L165

rhatdan:
@hshiina Are you still working on this PR?

hshiina:
I'm still working it.
As I rebased this PR, the rusult got changed.
I will investigate it.

hshiina:
Now that this PR doesn't only fix a test, but also fixes the code, I edited the title
and the description.

rhatdan:
In Podman we default the GID Map to match the UID map if the GID map is not specified.  Should we do the same here?

nalind:
LGTM.  Nice catch!

TomSweeneyRedHat:
LGTM
thanks @hshiina !

rhatdan:
/approve
Thanks @hshiina

PR #2652, opened 2020-09-24
Makefile: tweak the cross-compile target

nalind:
#### What type of PR is this?

/kind cleanup

What `go tool dist list` says the toolchain supports changes, so this change removes these attempted cross-compile build targets.
* GOOS=darwin, GOARCH unspecified build
* GOOS=darwin, GOARCH=386 build

Replace our use of slices of github.com/opencontainers/runc/libcontainer/configs.Device structures with a locally-defined type alias so that we can avoid importing the package on non-Unixy systems.  The result is not going to be a very useful binary on non-Linux systems, but it helps ensure that our subpackages won't break compilation for other projects who consume us as a library.

#### How to verify it

CI all the way.

None

Don't try to actually run the non-Linux binaries.

```
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,12 +62,12 @@ bin/buildah:  $(SOURCES)
 buildah: bin/buildah
 
 .PHONY: cross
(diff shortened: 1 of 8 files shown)


rhatdan:
LGTM

TomSweeneyRedHat (commented):
LGTM

PR #2668, opened 2020-10-01
[release-1.16] ADD: only expand archives at the right time

### LINKED ISSUE #2090: ADD inside dockerfile is very slow
**Description**

I want to build an OCI image for https://github.com/intel/sriov-network-device-plugin

I first used the Dockerfile provided in images/Dockerfile with the command:

```
cd ~/sriov-network-device-plugin
sudo buildah --log-level debug bud -t nfvpe/sriov-device-plugin ./images/Dockerfile .
```

It ends up with an error. I can see in the log that only files in the "./images" directory are copied, which is why the make commands in the dockerfile report an error.

I then copied the Dockerfile at ~/sriov-network-device-plugin, and adapt it.

The buildah command succeeded, but it took a very long time (about 30 min). I suspect it is because it copies files one by one, as I can see in the logs, and as reported in #1714. 

I tested with the buildah command line:

```
cd ~/sriov-network-device-plugin
ctr=$(buildah from golang:alpine)
buildah add $ctr '.' '/usr/src/sriov-network-device-plugin'
```

Those commands worked fine and took about 1 min.

```
$ apt list buildah
Listing... Done
buildah/unknown,now 1.12.0-3 amd64 [installed]
```

```$ buildah version
Version:         1.12.0
Go Version:      go1.11.6
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.1-dev
CNI Spec:        0.4.0
libcni Version:  
image Version:   5.0.0
Git Commit:      
Built:           Thu Jan 16 15:27:29 2020
OS/Arch:         linux/amd64
```

```
$ cat /etc/*release
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

**Output of `uname -a`:**

```
$ uname -a
Linux k8s-test 4.19.0-5-amd64 #1 SMP Debian 4.19.37-5+deb10u2 (2019-08-08) x86_64 GNU/Linux
```

```
$ cat /etc/containers/storage.conf
```

nalind:
#### What type of PR is this?

/kind bug

When ADD was used to ADD a directory, the contents of archives that we found inside of it were incorrectly being expanded at the destination for the ADD.

#### How to verify it

Attempt to build https://github.com/intel/sriov-network-device-plugin using its `images/Dockerfile` file and its top-level directory as the build context.  The contents of the `images/ddptool-1.0.0.0.tar.gz` should not be expanded, but if it does, its "makefile" will override the "Makefile" at the top level of the build context, and the build will fail.

Added conformance tests for this.

Fixes #2090.
 
Cherry-picked from #2667.

```
The contents of archives contained in directories being copied using the ADD instruction or `buildah add` should no longer be expanded into the destination directory.
```


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

PR #2681, opened 2020-10-09
ADD and COPY: descend into excluded directories, sometimes

nalind:
#### What type of PR is this?

/kind bug

When a directly-named (or globbed) source directory for `ADD` or `COPY` is marked for exclusion by `.dockerignore`, check if its name is a prefix for any exceptions in the `.dockerignore` file, and if it is, check the directory for things we need to include anyway.

This will miss exceptions where the exception pattern uses a wildcard for anything but the final component, but an earlier version of this patch, which walked nominally-excluded directories until it either found something that should be included, or confirmed that it there wasn't anything under them to include, failed conformance tests.

When adding items, count items that are actually passed over the tar pipe, rather than items scanned, so that we can correctly diagnose not having found anything that we needed to copy under a directory that would otherwise have been excluded.

In `copierHandlerGet()`, just don't discount any globbed directories that are excluded.

#### How to verify it

A number of added conformance tests and tweaked some unit tests in the `copier` package.  Other tests we already had should continue to pass.  This aims to fix https://github.com/containers/podman/issues/7922, which includes a reproducer, so that's worth a try, too.

This will solve https://github.com/containers/podman/issues/7922 once it's merged into podman.

```
This update fixes a regression introduced in 1.16 which caused some exceptions to exclusion rules noted in `.dockerignore` to not be copied or added from a build context.
```


TomSweeneyRedHat:
Changes LGTM, would just like to get #2682 in first

rhatdan (approved):
/lgtm

PR #2682, opened 2020-10-09
Cirrus: Give more memory to golangci-lint

cevich:
/kind failing-test 

Increase memory available to `golangci-lint` so that it isn't randomly OOMkilled.

#### How to verify it

Run the automated tests

None

None

None

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -87,14 +87,17 @@ gce_instance:
 
 
 'cirrus-ci/only_prs/gate_task':
-    gce_instance:
-        memory: "12Gb"
 
     # N/B: Skip running this on branches due to multiple bugs in
     # the git-validate tool which are difficult to debug and fix.
     skip: $CIRRUS_PR == ''
 
+    # Ensure more costly (bigger) VM can't stick around longer than necessary
     timeout_in: 10m
+
+    gce_instance:
+        cpu: 4
+        memory: "16Gb"
 
     setup_script: '${SCRIPT_BASE}/setup.sh |& ${_TIMESTAMP}'
     build_script: '${SCRIPT_BASE}/build.sh |& ${_TIMESTAMP}'


cevich:
```
Failed to start: Invalid value for field 'resource.machineType': 'https://www.googleapis.com/compute/v1/projects/buildah/zones/us-central1-c/machineTypes/custom-2-16384'. Memory size for 2 vCPU instance should be between 2048MiB and 13312MiB, while 16384MiB is requested
```

.cirrus.yml
```diff
@@ -92,7 +92,11 @@ gce_instance:
     # the git-validate tool which are difficult to debug and fix.
     skip: $CIRRUS_PR == ''
 
-    timeout_in: 30m
+    timeout_in: 10m
```
TomSweeneyRedHat, line 95: Just curious why you lowered the time?

TomSweeneyRedHat:
One question otherwise 
LGTM
and happy green test buttons.  (Maybe never question success @TomSweeneyRedHat ?)

.cirrus.yml
cevich: To avoid leaving a big-expensive VM hanging around any longer than necessary.  For this task, if it takes longer than a minute or three, it's likely somethings very wrong and it needs to timeout.

cevich: Do you think this is worthy of a comment in the code?

TomSweeneyRedHat: Wouldn't hurt, but I'd add it later.  I'd like to get this in sooner, especially given the green test buttons.

TomSweeneyRedHat:
@nalind @dwalsh PTAL

nalind:
I mean, it clearly seems to help, but I do wonder if we need to go all the way from 2 vcpus with 4GB of memory to 8 vcpus with 32GB of memory to get that result.  Have we still been seeing the problem with values in between, say 12GB, which is within the bounds of what we can request with a 2 vcpu instance?

rhatdan:
If @nalind change works, do we still need this one?

nalind:
Well, that was a guess, not based on profiling or any data.

cevich:
Mine is a educated guess as well.  Educated based on hitting something similar in podman, and finding the 16-18gig to be about  the sweet-spot.  The gating tests run so quickly, I'm not overly concerned about "overshooting", and the chance of memory needs growing over time seems greater than it reducing.  That was my thought process.

cevich:
Rebased, changed to 4CPU/16Gig as maybe a sweet-spot.  Leaving this open for a little while, just in case it's needed (Nalin's 12gig is found to be insufficient).

cevich:
We can close this at start of Nov. if there are no further OOM killing instances in gate tests...and If I forget to close it sooner :smile:

rhatdan:
Since we are past halloween, I am going to close this.  Cirrus seems to be working well.

cevich:
Thanks Dan.

PR #2659, opened 2020-09-29
add: preserve ownerships and permissions on ADDed archives

nalind:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

When extracting archives that are added using ADD, don't override permissions and ownership information.  We regressed on this when we switched to using the copier package to handle them.

#### How to verify it

Added a new conformance test for it.

#### Which issue(s) this PR fixes:

Fixes #2657.

#### Does this PR introduce a user-facing change?

```
Permissions and ownership information on the contents of archives added using the ADD instruction or "buildah add" on the command line should be preserved again.
```

PR #2683, opened 2020-10-09
[release-1.16] Set directory ownership when copied with ID mapping

nalind:
#### What type of PR is this?

/kind bug

When a directory is copied by `buildah copy` to a container with ID mapping, the directory owner is not set correctly after #2659 was merged. This PR sets the ownership correctly.

This also fixes `idmapping` test to verify this fix. It seems that this test has not been working since the variable `maps` was renamed.

#### How to verify it
Run `bats tests/namespaces.bats`.

None

Cherry-picked from #2647.

```
None
```


TomSweeneyRedHat:
LGTM

PR #2684, opened 2020-10-09
refactor pullpolicy to avoid deps

baude:
refeactor the pullpollicy to its own package so that it can be imported on its own without deps like containers/storage, etc. this will help consuming podman's go bindings.

/kind other

```release-note
None
```

### DIFF
--- a/buildah.go
+++ b/buildah.go
@@ -11,6 +11,7 @@ import (
 	"sort"
 	"time"
 
+	"github.com/containers/buildah/define"
 	"github.com/containers/buildah/docker"
 	"github.com/containers/image/v5/types"
 	encconfig "github.com/containers/ocicrypt/config"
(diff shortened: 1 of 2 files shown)


buildah.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/containers/buildah/define"
```
nalind, line 7: Move this to the group of non-standard-library imports.

buildah.go
```diff
@@ -78,6 +80,8 @@ func (p PullPolicy) String() string {
 	return fmt.Sprintf("unrecognized policy %d", p)
 }
 
+*/
+
```
nalind, line 84: I think we can remove this block instead.

rhatdan:
LGTM

PR #2690, opened 2020-10-13
Skip tlsVerify insecure BUILD_REGISTRY_SOURCES

### LINKED ISSUE #2586: insecureRegistries from BUILD_REGISTRY_SOURCES ignored
**Description**

`insecureRegistries` from the `BUILD_REGISTRY_SOURCES` env. var. is ignored.

There's nothing in the source that actually handles this key AFAICS - the only reference is https://github.com/containers/buildah/blob/master/commit.go#L186.

The `insecureRegistries` key is mentioned in e.g. `man buildah-pull` and `man buildah-push`. At a minimum the documentation should mention that the key is currently non-functional.

```
BUILD_REGISTRY_SOURCES='{"insecureRegistries": ["<registry>"]}' buildah push <img> <registry>/<img>
```

```
Get "https://<registry>/v2/": http: server gave HTTP response to HTTPS client
ERRO exit status 125
```

Successful push to registry.

```
buildah-1.15.1-1.fc32.x86_64
```

QiWang19:
If the registry is set to insecure allowd using BUILD_REGISTRY_SOURCES, hardcode to skip the tls verify to avoid the errors.
Returns error if set insecureRegistries but force to use tls-verify.

#### What type of PR is this?

> /kind bug

close #2586 

```release-note

```

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -167,17 +167,17 @@ var (
 // variable, if it's set.  The contents are expected to be a JSON-encoded
 // github.com/openshift/api/config/v1.Image, set by an OpenShift build
 // controller that arranged for us to be run in a container.
-func checkRegistrySourcesAllows(forWhat string, dest types.ImageReference) error {
+func checkRegistrySourcesAllows(forWhat string, dest types.ImageReference) (insecure bool, err error) {
 	transport := dest.Transport()
 	if transport == nil {
-		return nil
+		return false, nil
 	}
 	if transport.Name() != docker.Transport.Name() {
-		return nil
+		return false, nil
 	}
 	dref := dest.DockerReference()
 	if dref == nil || reference.Domain(dref) == "" {
-		return nil
+		return false, nil
 	}
 
 	if registrySources, ok := os.LookupEnv("BUILD_REGISTRY_SOURCES"); ok && len(registrySources) > 0 {
@@ -188,7 +188,7 @@ func checkRegistrySourcesAllows(forWhat string, dest types.ImageReference) error
 			AllowedRegistries  []string `json:"allowedRegistries,omitempty"`
 		}
 		if err := json.Unmarshal([]byte(registrySources), &sources); err != nil {
-			return errors.Wrapf(err, "error parsing $BUILD_REGISTRY_SOURCES (%q) as JSON", registrySources)
+			return false, errors.Wrapf(err, "error parsing $BUILD_REGISTRY_SOURCES (%q) as JSON", registrySources)
 		}
 		blocked := false
 		if len(sources.BlockedRegistries) > 0 {
@@ -199,7 +199,7 @@ func checkRegistrySourcesAllows(forWhat string, dest types.ImageReference) error
 			}
 		}
 		if blocked {
-			return errors.Errorf("%s registry at %q denied by policy: it is in the blocked registries list", forWhat, reference.Domain(dref))
+			return false, errors.Errorf("%s registry at %q denied by policy: it is in the blocked registries list", forWhat, reference.Domain(dref))
 		}
 		allowed := true
 		if len(sources.AllowedRegistries) > 0 {
@@ -211,10 +211,13 @@ func checkRegistrySourcesAllows(forWhat string, dest types.ImageReference) error
 			}
 		}
 		if !allowed {
-			return errors.Errorf("%s registry at %q denied by policy: not in allowed registries list", forWhat, reference.Domain(dref))
+			return false, errors.Errorf("%s registry at %q denied by policy: not in allowed registries list", forWhat, reference.Domain(dref))
+		}
+		if len(sources.InsecureRegistries) > 0 {
+			return true, nil
 		}
 	}
-	return nil
+	return false, nil
 }
 
 // Commit writes the contents of the container, along with its updated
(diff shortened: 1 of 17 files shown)


commit.go
```diff
@@ -278,9 +281,18 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
...
 
 	// Check if the commit is blocked by $BUILDER_REGISTRY_SOURCES.
-	if err := checkRegistrySourcesAllows("commit to", dest); err != nil {
+	insecure, err := checkRegistrySourcesAllows("commit to", dest)
+	if err != nil {
 		return imgID, nil, "", err
 	}
+	if insecure {
+		if systemContext.DockerInsecureSkipTLSVerify == types.OptionalBoolFalse {
+			return imgID, nil, "", errors.Errorf("can't require tls verify on insecurely allowed registry")
```
TomSweeneyRedHat, line 290:
```suggestion
			return imgID, nil, "", errors.Errorf("can't require tls verification on an insecured registry")
```

commit.go
```diff
@@ -291,9 +303,18 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
...
 				return imgID, nil, "", errors.Wrapf(err, "error parsing image name %q as an image reference", name)
 			}
-			if err := checkRegistrySourcesAllows("commit to", additionalDest); err != nil {
+			insecure, err := checkRegistrySourcesAllows("commit to", additionalDest)
+			if err != nil {
 				return imgID, nil, "", err
 			}
+			if insecure {
+				if systemContext.DockerInsecureSkipTLSVerify == types.OptionalBoolFalse {
+					return imgID, nil, "", errors.Errorf("can't require tls verify on insecurely allowed registry")
```
TomSweeneyRedHat, line 312:
```suggestion
					return imgID, nil, "", errors.Errorf("can't require tls verification on an insecured registry")
```

commit.go
```diff
@@ -471,9 +492,18 @@ func Push(ctx context.Context, image string, dest types.ImageReference, options
...
 
 	// Check if the push is blocked by $BUILDER_REGISTRY_SOURCES.
-	if err := checkRegistrySourcesAllows("push to", dest); err != nil {
+	insecure, err := checkRegistrySourcesAllows("push to", dest)
+	if err != nil {
 		return nil, "", err
 	}
+	if insecure {
+		if systemContext.DockerInsecureSkipTLSVerify == types.OptionalBoolFalse {
+			return nil, "", errors.Errorf("can't require tls verify on insecurely allowed registry")
```
TomSweeneyRedHat, line 501:
```suggestion
			return nil, "", errors.Errorf("can't require tls verification on an insecured registry")
```

pull.go
```diff
@@ -238,9 +238,18 @@ func pullImage(ctx context.Context, store storage.Store, srcRef types.ImageRefer
...
 		return nil, errors.Errorf("pull access to registry for %q is blocked by configuration", transports.ImageName(srcRef))
 	}
-	if err := checkRegistrySourcesAllows("pull from", srcRef); err != nil {
+	insecure, err := checkRegistrySourcesAllows("pull from", srcRef)
+	if err != nil {
 		return nil, err
 	}
+	if insecure {
+		if sc.DockerInsecureSkipTLSVerify == types.OptionalBoolFalse {
+			return nil, errors.Errorf("can't require tls verify on insecurely allowed registry")
```
TomSweeneyRedHat, line 247:
```suggestion
			return nil, errors.Errorf("can't require tls verification on an insecured registry")
```

commit.go
TomSweeneyRedHat: It doesn't look like we document in the man pages.  We might want to add something like: "This option cannot be used when talking to an insecure registry." in the verbiage for the tls-verify option for each of these.

commit.go
```diff
@@ -167,17 +167,17 @@ var (
 // variable, if it's set.  The contents are expected to be a JSON-encoded
 // github.com/openshift/api/config/v1.Image, set by an OpenShift build
 // controller that arranged for us to be run in a container.
-func checkRegistrySourcesAllows(forWhat string, dest types.ImageReference) error {
+func checkRegistrySourcesAllows(forWhat string, dest types.ImageReference) (bool, error) {
```
TomSweeneyRedHat, line 170: Just for clarity, it might be good to name the return values if gofmt allows:
```suggestion
func checkRegistrySourcesAllows(forWhat string, dest types.ImageReference) (insecure bool, err error) {
```

tests/push.bats
```diff
@@ -181,6 +181,16 @@ load helpers
   rm -rf ${TESTDIR}/tmp
 }
 
+@test "buildah push to registry allowed by BUILD_REGISTRY_SOURCES" {
+  _prefetch busybox
+  export BUILD_REGISTRY_SOURCES='{"insecureRegistries": ["localhost:5000"]}'
+
+  run_buildah 125 push --creds testuser:testpassword  --signature-policy ${TESTSDIR}/policy.json --tls-verify=true busybox docker://localhost:5000/buildah/busybox:latest
+  expect_output --substring "can't require tls verify on insecurely allowed registry"
```
TomSweeneyRedHat, line 189: this will need to change if you take the rewording.

commit.go
QiWang19: Change to `TLS verification cannot be used when talking to an insecure registry.`, since the option --tls-verify=false is allowed if using insecure registry.

docs/buildah-bud.md
```diff
@@ -467,7 +467,7 @@ Commands after the target stage will be skipped.
 
 **--tls-verify** *bool-value*
 
-Require HTTPS and verify certificates when talking to container registries (defaults to true).
+Require HTTPS and verify certificates when talking to container registries (defaults to true).  TLS verification cannot be used when talking to an insecure registry.
```
TomSweeneyRedHat, line 470: Nicely stated @QiWang19

TomSweeneyRedHat (commented):
LGTM

TomSweeneyRedHat:
@QiWang19 looks like you need a rebase.

QiWang19:

PR #2695, opened 2020-10-15
short-name aliasing

vrothberg:


vrothberg:

I had to separate local-image lookup and the remote one. Also rewired the pull-policy handling.

vrothberg:
@edsantiago, added a check for the `CONTAINERS_SHORT_NAME_ALIASING` variable. Prompting will now work iff it's set to "on".

rhatdan:
@vrothberg Needs a rebase.

rhatdan:
@vrothberg I took away access to the tty(stdin) by default in buildah bud, I feel this will probably break this PR.

vrothberg:
> @vrothberg I took away access to the tty(stdin) by default in buildah bud, I feel this will probably break this PR.

I'll rebase and test.

vrothberg:
@rhatdan continues working :+1:

rhatdan:
I was surprised that it prompted me for multiple choices, when it could have gone out and figured out which ones were wrong.  

I think we either need to pre-search or give the user the ability to say grab the first one you find.
Otherwise if the user does not know in which registry the image exists, they will be forced to run the command multiple times.
```
$ CONTAINERS_SHORT_NAME_ALIASING=on ./bin/buildah pull ubi8-minimal
✔ registry.fedoraproject.org/ubi8-minimal:latest
while pulling "ubi8-minimal" as "//registry.fedoraproject.org/ubi8-minimal:latest": Error initializing source docker://registry.fedoraproject.org/ubi8-minimal:latest: Error reading manifest latest in registry.fedoraproject.org/ubi8-minimal: manifest unknown: manifest unknown
```

rhatdan:
If I setup a bogus entry in my shortnames.conf file, 
```
cat ~/.config/containers/short-name-aliases.conf
[aliases]
  ubi8-minimal = "registry.access.redhat.com/ubi8-minimal"
  fedora = "registry.access.redhat.com/fedora"
```
And now I pull the fedora shortname.

```
CONTAINERS_SHORT_NAME_ALIASING=on ./bin/buildah pull fedora
while pulling "fedora" as "//registry.access.redhat.com/fedora:latest": Error initializing source docker://registry.access.redhat.com/fedora:latest: Error reading manifest latest in registry.access.redhat.com/fedora: name unknown: Repo not found
ERRO exit status 125                           
```
The error should indicate that the shortname was retrieved from ~/.config/containers/short-names-aliases.conf

vrothberg:
> I was surprised that it prompted me for multiple choices, when it could have gone out and figured out which ones were wrong.
> 
> I think we either need to pre-search or give the user the ability to say grab the first one you find.

Pre-searching the image can take long and can noticeably delay the prompt which is not something I want to do. The bandwith is frustratingly low at our new place, and I would be quite annoyed by a delayed prompt.

We can further run into issues where the desired registry is temporarily off, so we would not prompt.

Another issue is that remote-inspect does not take mirrors into account.

> Otherwise if the user does not know in which registry the image exists, they will be forced to run the command multiple times.

Isn't that a core effect we want to yield at the user? The prompting should force users to think what they are doing.

rhatdan:
`buildah from` does not seem to have support.

rhatdan:
buildah bud does not have support either.

rhatdan:
@vrothberg Ok I can live with the prompting, but I still think their should be an 'I don't know' option or a 'you find it' option.  Which up til now users were relying on.

A lot of this will pre-empted if we pre-populate the list.

rhatdan:
I tried to preload /etc/containers/registries.conf, and that worked. nice.

rhatdan:
If we preload and we have the same name in two different sites, then we need to make sure the code pulls the one that has the regististries.conf correct?

```
[aliases]
  fedora = "registry.fedoraproject.org/fedora"
  fedora = "docker.io/fedora"
```
What happens here?  Or is this up to the distro to prevent, and we just do First?

vrothberg:
> `buildah from` does not seem to have support.
> buildah bud does not have support either.

Both work on my end. Maybe you tried with an already aliased image?

vrothberg:
> If we preload and we have the same name in two different sites, then we need to make sure the code pulls the one that has the regististries.conf correct?
> 
> ```
> [aliases]
>   fedora = "registry.fedoraproject.org/fedora"
>   fedora = "docker.io/fedora"
> ```
> 
> What happens here? Or is this up to the distro to prevent, and we just do First?

That'll cause a TOML loading failure:
```
buildah (shortnames) $ CONTAINERS_SHORT_NAME_ALIASING=on ./bin/buildah pull fedora
error resolving local image "fedora": error loading registries configuration "/home/vrothberg/.config/containers/registries.conf": Near line 3 (last key parsed 'aliases.fedora'): Key 'aliases.fedora' has already been defined.
ERRO exit status 125                              
```

vrothberg:
> @vrothberg Ok I can live with the prompting, but I still think their should be an 'I don't know' option or a 'you find it' option. Which up til now users were relying on.

I feel torn. I see that it would be more comfortable but it also defeats the purpose of fighting spoofing registries. If a the mode is set to "enforcing", I would hope that there will be no ambiguity any more regarding the source of an image. Allowing to "try all in given order", we'd reintroduce this ambiguity.

edsantiago:
>I think we either need to pre-search or give the user the ability to say grab the first one you find.

I've waffled on this, and currently lean (slightly) toward not checking: given the flakiness of registries, it is conceivable that a nefarious agent could put (e.g.) `ubi8` in a rogue registry and just wait. At some point the true source registry will be down, and the customer would end up with a permanent cache-conf setting pointing to the invalid image.

(@vrothberg FWIW I don't think the speed issue is a concern. I too have very slow DSL. No matter how slow, though, searching is a blip compared to pulling)

rhatdan:
@vrothberg 
This definitely does not prompt.  @edsantiago could you verify?

```
./bin/buildah from rhel7
Checking if image destination supports signatures
Copying blob 88b5f8ffd297 done  
Copying blob d4095a8ffba5 done  
Copying config d7ef0eb5bc done  
Storing signatures
rhel7-working-container
```

vrothberg:
@rhatdan, can you try `CONTAINERS_SHORT_NAME_ALIASING=on ./bin/buildah pull rhel7`?

rhatdan:
@vrothberg Your right,  I was doing that with pull and then forgetting it with from.

vrothberg:
> @vrothberg Your right, I was doing that with pull and then forgetting it with from.

Happy it works! Thanks a ton for checking.

rhatdan:
This works nicely. I selected the  `registry.access.redhat.com/rhel7:latest` and it puts out the nice checkmark, showing me what I am pulling.

```
$ CONTAINERS_SHORT_NAME_ALIASING=on ./bin/buildah from rhel7
✔ registry.access.redhat.com/rhel7:latest
Checking if image destination supports signatures
Copying blob 88b5f8ffd297 done  
Copying blob d4095a8ffba5 done  
Copying config d7ef0eb5bc done  
Storing signatures
rhel7-working-container
```
Now it records the selection.  So I remove the image and pull it again.

```
$ ./bin/buildah rmi -f rhel7
untagged: registry.access.redhat.com/rhel7:latest
d7ef0eb5bc7fb70bdc8036f62d8ca9a7cea5bff74cb9c168ab9a920f21abbeb9

$ CONTAINERS_SHORT_NAME_ALIASING=on ./bin/buildah from rhel7
Checking if image destination supports signatures
Copying blob d4095a8ffba5 done  
Copying blob 88b5f8ffd297 done  
Copying config d7ef0eb5bc done  
Storing signatures
rhel7-working-container
```
On the second from, it would be nice if the UI would show me then name of the image I am pulling. IE I want the checkbox back,when shortnames expansion happens.

rhatdan:
BTW, @vrothberg This is brilliant.

rhatdan:
We just need some curl/web expert to be able to figure out a way to scape together a list of  "official" image names from Docker, Fedora, RHEL and Centos.  And maybe eventually from other sites. OpenSUSE.

OK I tried handling shortnames that are not listed in registy and that works nicely.
```
$ cat ~/.config/containers/short-name-aliases.conf
[aliases]
  fedora = "docker.io/library/fedora"
  ubi8-minimal = "registry.access.redhat.com/ubi8-minimal"
  sle15 = "registry.suse.com/suse/sle15"

$ CONTAINERS_SHORT_NAME_ALIASING=on ./bin/buildah from sle15
Copying blob 2d1073a97bec done  
Copying config 9ee87c6555 done  
Storing signatures
sle15-working-container
```
But registry.suse.com refers to this image as suse/sle15.
If I setup a shortname for this image, I get an error.

```
$ CONTAINERS_SHORT_NAME_ALIASING=on ./bin/buildah from sle15
error resolving local image "sle15": error loading short-name aliases config file "/home/dwalsh/.config/containers/short-name-aliases.conf": Near line 4 (last key parsed 'aliases'): bare keys cannot contain '/'
```
We will need to document that "/" is not allowed in shortnames.

vrothberg:
> We will need to document that "/" is not allowed in shortnames.

@rhatdan it is allowed, but you need to quote it.

vrothberg:
It should always be `"value"="name"` this way we make TOML happy

rhatdan:
Blocking still works so this is good.  I blocked registry.suse.com and shortnames did not bypass this.
```
$ CONTAINERS_SHORT_NAME_ALIASING=on ./bin/buildah from sle15
while pulling "sle15" as "//registry.suse.com/suse/sle15:latest": pull access to registry for "docker://registry.suse.com/suse/sle15:latest" is blocked by configuration
ERRO exit status 125 
```

rhatdan:
This works:
  "suse/sle15" = "registry.suse.com/suse/sle15"

I have found nothing broken.  Just some potential UI Enhancements.
This will get interesting when it gets out into the real world to see what users like/dislike about it.

rhatdan:
Probably should test with multi stage images, where there are more then one from option in the Dockerfile.

rhatdan:

vrothberg:
@rhatdan, I added a way to improve the error messaging a bit (see `shortnames.GuessError(...)`). PTAL but note that we are limited in what determining the source of an error. However, it's now much easier to debug.

new.go
```diff
@@ -103,145 +104,162 @@ func newContainerIDMappingOptions(idmapOptions *IDMappingOptions) storage.IDMapp
...
-					return ref, transport, img, nil
-				}
+	var pullErrors []error
+	for _, pullCandidate := range pullCandidates {
+		name := pullCandidate.ref.StringWithinTransport()
+
+		// Skip this candidate if it's "older" than the local one.  If
+		// there's no local image, use the first successfully pulled
+		// remote one.
+		if options.PullPolicy == PullIfNewer && localImage != nil {
```
TomSweeneyRedHat, line 215: Maybe it will be clearer further down the code, but why are we checking for a local image?  If we don't have a local image and the user has asked for a PullIfNewer, we should pull if the image is on the repository.  Do we now?

TomSweeneyRedHat: Well testing seems to show it pulls the image, if appropriate to do, with this code.  I'll have to test further and think more on Tuesday.  Off on PTO Monday and probably won't get time.

vrothberg: Thanks for checking!

In this case, when there's no matching local image, we will pull the first image we can find. We only need to do the comparison if the the policy is set to "newer" and if there's a local image. That's what the condition does. I'll add a more verbose comment to prevent further confusion.

TomSweeneyRedHat:
@vrothberg can you rebase and repush this please?

TomSweeneyRedHat:
Change LGTM, but I wasn't able to test it due to a rebase being needed.  If you can do that, I'll give it a whirl.
A review from @nalind, especially on the image look up calls would be good if we can steal his eyeballs.

vrothberg:
> Change LGTM, but I wasn't able to test it due to a rebase being needed. If you can do that, I'll give it a whirl.

I will rebase but I don't understand why this would block testing the PR.

TomSweeneyRedHat:
@vrothberg it was late in my day and I couldn't build it without the rebase.  The eyeballs wouldn't focus enough to find/squash the conflict and my couch was calling longily  for me.  Thanks for the rebase.

vrothberg:
> @vrothberg it was late in my day and I couldn't build it without the rebase. The eyeballs wouldn't focus enough to find/squash the conflict and my couch was calling longily for me. Thanks for the rebase.

Were you fetching the commits and applied them locally?

In case, I highly recommend using the GitHub client (https://github.com/cli/cli). A `gh pr checkout $PR-Number` will checkout the PR branch locally. The tool can do much more but I find this especially helpful.

TomSweeneyRedHat:
@vrothberg cli?  That has one more letter than vi, so it can't be any good!  Thanks for the rebase!

TomSweeneyRedHat:
@vrothberg  I ran into this:

```
# buildah pull stable
No matching short-name alias has been found for "stable" and no unqualified-search registry in "/etc/containers/registries.conf" or "/root/.config/containers/registries.conf.d" includes a matching image
```
It's not clear where/how I'd setup a `short-name alias`.  Is there doc somewhere in the other PR's that explains how to do this?

TomSweeneyRedHat:
@vrothberg needs another rebase.  I've played with it last night for an hour or so, and a couple hours today.  Overall looks good and seems to work well.  Would like a head nod from @nalind and you'll need a rebase.  I did leave one question on the alias file, it would be nice to document that somewhere if it's not already.

TomSweeneyRedHat:
Found the docs for the alias files, ignore my comments on that (https://github.com/containers/image/pull/1061)

new.go
```diff
@@ -103,145 +104,166 @@ func newContainerIDMappingOptions(idmapOptions *IDMappingOptions) storage.IDMapp
...
+	// No transport, so let's use short-name resolution to find potential
+	// pullCandidates.
+	if len(pullCandidates) == 0 {
+		candidates, err := shortnames.Resolve(systemContext, options.FromImage)
 		if err != nil {
-			return nil, "", nil, errors.Wrapf(err, "error parsing reference to image %q", destImage)
+			return nil, "", nil, err
+		}
+		for i, candidate := range candidates {
+			ref, err := docker.ParseReference("//" + candidate.String())
```
mtrmac, line 200: Eventually this should be `docker.NewReference(candidate.Value)` — after `.Value` is always normalized, which is not currently the case.

mtrmac (commented):
(Just a very quick skim for uses of the new API, I haven’t reviewed the substantial refactoring at all.)

edsantiago:
@TomSweeneyRedHat @vrothberg I'm catching up on review; in addition to `gh` (which I find obnoxious) there's the much cleaner and simpler `git pr`. Just `dnf install git-extras`, then (assuming your git repo is set up so that `upstream` is `containers/podman`) `git pr 1234 upstream` will pull and check out a `pr/1234` branch. I've been using it for years, it's beautiful.

vrothberg:
@TomSweeneyRedHat, just updated pull-if-newer to compare digests :+1:

vrothberg:
> @TomSweeneyRedHat, just updated pull-if-newer to compare digests +1

I reverted that and kept the current time-stamp comparisons. I ran into https://github.com/containers/buildah/issues/2779 which turned out to be quite some work that I don't want to mix with (or block on for) short names.

rhatdan:
LGTM

rhatdan:

vrothberg:
Let's get this merged please. Then we can merge the Podman PR as. We can create a separate Podman PR to vendor the Buildah release but then we're not blocked on me.

vrothberg:
Note, we don't need to rebase to merge.

baude:
@vrothberg has asked me to merge even with the conflict

/lgtm

vrothberg:
> @vrothberg has asked me to merge even with the conflict

Note that there's no conflict. "Rebase" only means that master has moved on (another PR was merged) and that it's not the base of this PR any more. Merge conflicts would be displayed and would block a merge.

In this case we're safe as the commits don't interfere.

mtrmac:
Returning to this for an unrelated reason, isn’t https://github.com/containers/buildah/blob/d0c958d65eb287afbabb93369412634d37735bf8/util/util.go#L85-L94 dead code? Also the `searchRegistriesAreEmpty` value is now of dubious utility.

Issue #2701, opened 2020-10-19
build fails on fedora 33 beta for a dockerfile


damnms:
**Description**
i get an error on fedora 33 beta with podman when i try to build one of the dockerfiles

**Steps to reproduce the issue:**
1. checkout https://gitlab.com/olze/fakecam
2. run: cd fakecam && podman build -t bodypix ./bodypix
3.

**Describe the results you received:**
an error which was working a couple of days/weeks before

```
STEP 1: FROM tensorflow/tensorflow
STEP 2: RUN apt update && apt install -y curl make build-essential     && curl -sL https://deb.nodesource.com/setup_12.x | bash -     && apt-get -y install nodejs     && mkdir /.npm     && chmod 777 /.npm
--> Using cache 1f3b234d1a2c3b56be00dc2f8b4e17121e587d7d8c4d1e7fa4cdbfc2bcce708d
--> 1f3b234d1a2
STEP 3: ENV TF_FORCE_GPU_ALLOW_GROWTH=true
--> Using cache 1ce181da550747d729f1997c9c0debcb3a00f9e0e700b25787436c39e6cd037d
--> 1ce181da550
STEP 4: WORKDIR /src
--> Using cache cbcee1187928b29973211456802aac97a194a2134a6c3ea617f17581454c6dd2
--> cbcee118792
STEP 5: COPY package.json /src/
--> Using cache 7bc760e4a9a2e37137e0a6dae105f9ceaf64163d87ad8ca0486df08cb7c89467
--> 7bc760e4a9a
STEP 6: RUN npm install
--> Using cache 1f6a8bc8405c82e4713106b4190422feb5f5fde73135ffb061160a5501af8ea4
--> 1f6a8bc8405
STEP 7: COPY app.js /src/
[...]

PR #2703, opened 2020-10-19
copier.copierHandlerPut: don't check length when there are errors

nalind:
#### What type of PR is this?

/kind bug

When attempting to create a file as part of extracting a tar stream, don't check if the number of bytes written that our helper function reports back to us matches our expectations if it also reported an error, in which case the byte count is irrelevant.

Related to https://github.com/containers/buildah/issues/2701, though this only fixes the error reporting for that issue.

```
Improves error reporting in cases where ADD or COPY (or "buildah add" or "buildah copy") encounter an error while attempting to create a file in a working container.
```


rhatdan:
LGTM
@TomSweeneyRedHat @ashley-cui @giuseppe @vrothberg @saschagrunert PTAL

vrothberg (commented):
LGTM

Issue #2705, opened 2020-10-20
buildah bud is gobbling stdin


edsantiago:
buildah bud is gobbling stdin that does not belong to it. This makes it impossible to pipe script commands to a shell.

```console
$ printf "echo hi\r\nbuildah bud -t foo .\necho bye\n" | bash
hi
STEP 1: ...
...
<sha>
```
Note that the 'echo bye' is gobbled up by buildah. If you redirect stdin on the buildah command, it works as expected:
```console
$ printf "echo hi\r\nbuildah bud -t foo . </dev/null\necho bye\n" | bash
hi
...
bye
```
See also: https://github.com/containers/podman/issues/8066

rhatdan:
I am not seeing this happen on the main branch of Podman, BTW.  But opened a PR anyways,

rhatdan:
This is fixed in master branch.

PR #2708, opened 2020-10-20
Buildah bud should not use stdin by default

rhatdan:
Fixes: https://github.com/containers/podman/issues/8066
Is reporting that buildah is eating stdin.  I don't beleive
we should be using stdin when doing a buildah bud command
unless `buildah bud -` is specified.  After this PR, the
`-` Dockerfile is still handled.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```


rhatdan:
@nalind PTAL

rhatdan:
Fixes: https://github.com/containers/buildah/issues/2705

nalind:
LGTM, though this does add a difference between how we handle RUN instructions and `buildah run`.

rhatdan:
I think that is ok.  The question I have is should we add an --interactive mode to docker build to add stdin back in.
I know where have an issue somewhere talking about how `buildah bud` or `podman build` hung waiting for input from apt-update.
If we added `-i` flag then dnf update and other commands could wait for user input.

edsantiago:
LGTM. Could you add this test, or something equivalent?
```bash
@test "bud does not gobble stdin" {
  _prefetch alpine

  ctxdir=${TESTDIR}/bud
  mkdir -p $ctxdir
  cat >$ctxdir/Dockerfile <<EOF
FROM alpine
RUN true
EOF

  random_msg=$(head -10 /dev/urandom | tr -dc a-zA-Z0-9 | head -c12)

  # Prior to #2708, buildah bud would gobble up its stdin even if it
  # didn't actually use it. This prevented the use of 'cmdlist | bash';
  # if 'buildah bud' was in cmdlist, everything past it would be lost.
  #
  # This is ugly but effective: it checks that buildah passes stdin untouched.
  passthru=$(echo "$random_msg" | (run_buildah bud --quiet --signature-policy ${TESTSDIR}/policy.json -t stdin-test ${ctxdir} >/dev/null; cat))

  expect_output --from="$passthru" "$random_msg" "stdin was passed through"
}
```

edsantiago:
P.S. the test (as expected) fails on current master, passes with this PR.

rhatdan:
@TomSweeneyRedHat @edsantiago @vrothberg @giuseppe @ashley-cui This is ready to go in.  Merge me.

hansborr:
> I think that is ok. The question I have is should we add an --interactive mode to docker build to add stdin back in.
> I know where have an issue somewhere talking about how `buildah bud` or `podman build` hung waiting for input from apt-update.
> If we added `-i` flag then dnf update and other commands could wait for user input.

I would like to see this added.

This change broke something for me -- I used to have a bud file which prompted for user input during the RUN stage (running smbpasswd).
This was to avoid having a cleartext password in the file, but to allow the hashed password to be baked into the image.

After updating to the newest buildah, that no longer works.

I was able to work around it, but I expect there are probably other people who will encounter this.

rhatdan:
Interested in opening a PR, it is actually a fairly easy change to add an option for this.

BTW, Please open an issue for this anyways, rather then in this closed PR.

PR #2709, opened 2020-10-20
Bump containerd, docker, and golang.org/x/sys

nalind:
#### What type of PR is this?

/kind other

Bump github.com/containerd/containerd from v1.3.2 to v1.4.1 and golang.org/x/sys from v0.0.0-20200810151505-1b9f1253b3ed to v0.0.0-20201018230417-eeed37f84f13.  Bump github.com/docker/docker to an unreleased tag to fix compilation with newer golang.org/x/sys.  The cross-compile build for GOOS=windows GOARCH=386 starts failing then, but that's not been a priority.

#### How to verify it

CI should pass.  Rebasing #2700 after this should go cleanly.

None

At some point an updated github.com/moby/term and github.com/Azure/go-ansiterm might fix the GOOS=windows GOARCH=386 build, and we can turn that one back on if we want.

```
None
```


rhatdan:
LGTM
If this passes...

PR #2725, opened 2020-10-27
overlay: fix umount

giuseppe:
/kind bug

fix unmounting overlay when running as rootless.

#### How to verify it

`mkdir /tmp/foo && podman run --rm -v /tmp/foo:/tmp/foo:O fedora true` works

```release-note
NONE
```

### DIFF
--- a/pkg/overlay/overlay.go
+++ b/pkg/overlay/overlay.go
@@ -113,6 +113,21 @@ func RemoveTemp(contentDir string) error {
 func Unmount(contentDir string) error {
 	mergeDir := filepath.Join(contentDir, "merge")
 
+	if unshare.IsRootless() {
+		// Attempt to unmount the FUSE mount using either fusermount or fusermount3.
+		// If they fail, fallback to unix.Unmount
+		for _, v := range []string{"fusermount3", "fusermount"} {
+			err := exec.Command(v, "-u", mergeDir).Run()
+			if err != nil && errors.Cause(err) != exec.ErrNotFound {
+				logrus.Debugf("Error unmounting %s with %s - %v", mergeDir, v, err)
+			}
+			if err == nil {
+				return nil
+			}
+		}
+		// If fusermount|fusermount3 failed to unmount the FUSE file system, attempt unmount
+	}
+
 	// Ignore EINVAL as the specified merge dir is not a mount point
 	if err := unix.Unmount(mergeDir, 0); err != nil && !os.IsNotExist(err) && err != unix.EINVAL {
 		return errors.Wrapf(err, "unmount overlay %s", mergeDir)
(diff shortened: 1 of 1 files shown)


pkg/overlay/overlay.go
```diff
@@ -153,14 +185,8 @@ func CleanupMount(contentDir string) (Err error) {
...
 
-	if unshare.IsRootless() {
-		mergeDir := filepath.Join(contentDir, "merge")
-		if err := unix.Unmount(mergeDir, 0); err != nil {
-			if !os.IsNotExist(err) {
-				return errors.Wrapf(err, "unmount overlay %s", mergeDir)
-			}
-		}
-	}
+	_ = Unmount(contentDir)
```
vrothberg, line 188: Is it intentional to ignore the error?

giuseppe: yes, on errors `Unmount()` itself will print the warnings

rhatdan:
LGTM

vrothberg (commented):
LGTM
Fixes all issues I have in the image volumes PR as well. Great work, @giuseppe !

Could you backport it to v1.16.X?

giuseppe:
> Could you backport it to v1.16.X?

sure!

giuseppe:
backport: https://github.com/containers/buildah/pull/2726

pkg/overlay/overlay.go
nalind: It doesn't look like an error from `ioutil.ReadDir()` will be logged.

giuseppe: do we still have to handle that somehow or just attempt to delete the directory in any case?

nalind: I'd expect it to be logged if the caller's just going to discard it anyway.

giuseppe: I've changed the implementation to only ignore `EINVAL` when the path is not a mount point

rhatdan:
Tests blew up.

pkg/overlay/overlay.go
```diff
@@ -109,15 +110,46 @@ func RemoveTemp(contentDir string) error {
...
+	}
+	for _, f := range files {
+		dir := filepath.Join(contentDir, f.Name())
+		if err := unmount(dir); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
```
nalind, line 152: `run_linux.go` calls `RemoveTemp()` with a directory created by `TempDir()`, which in turn calls `Unmount()`, but the rewritten `Unmount()` appears to expect to be called with the parent of such a directory, as `CleanupContent()` now does, leading to multiple errors being logged at the debug level when I try this out on the command line.

giuseppe: thanks.  I've restored the previous version of`Unmount()` and moved the new logic into `CleanupContent()`

nalind: `CleanupContent()` previously attempted to compute the path that it passed to `unix.Unmount()` without a component for name of the temporary directory in it (something like "${containerdir}/overlay/merge", when we mount overlays at something that looks like "${containerdir}/overlay/${random}/merge"), so it also didn't work correctly without this patch.

TomSweeneyRedHat:
Needs a rebase @giuseppe

PR #2731, opened 2020-10-29
If build format != oci do not use OCI layers

### LINKED ISSUE #2389: Build caching between --format changes is broken
**Description**

When building an image where a cache is available but the --format=... parameter does not match the cache, the cache is used anyway, resulting in an invalid cache. 

1. Clean all cached images
2. Create a dockerfile with an OCI-unsupported directive
3. Build the dockerfile with `buildah bud --images=true`
4. Build the same dockerfile with `buildah bud --images=true --format=docker`

The OCI-unsupported directive is ignored and the cache is used in step 4

Caches are invalidated due to changed format (or at least caches for the commands that have different meaning)

```
buildah.x86_64            1.14.8-1.fc31                    updates
```

```
$ buildah version
Version:         1.14.8
Go Version:      go1.13.9
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.1-dev
CNI Spec:        0.4.0
libcni Version:  
image Version:   5.4.3
Git Commit:      
Built:           Thu Jan  1 01:00:00 1970
OS/Arch:         linux/amd64
```

```
Fedora release 31 (Thirty One)
NAME=Fedora
VERSION="31 (KDE Plasma)"
ID=fedora
VERSION_ID=31
VERSION_CODENAME=""
PLATFORM_ID="platform:f31"
PRETTY_NAME="Fedora 31 (KDE Plasma)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:31"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f31/system-administrators-guide/"
REDHAT_BUGZILLA_PRODUCT_VERSION=31
REDHAT_SUPPORT_PRODUCT_VERSION=31
VARIANT="KDE Plasma"
VARIANT_ID=kde
Fedora release 31 (Thirty One)
Fedora release 31 (Thirty One)
```

**Output of `uname -a`:**

```
Linux toto 5.6.13-200.fc31.x86_64 #1 SMP Thu May 14 23:26:14 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

https://del.dog/amirirfude

3sky:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it:**

In case of setting `--layers` flag to `true`, buildah uses the same caching layers for OCI and Docker image format which is incorrect behavior.  

**How to verify it**

```bash
# Dockerfile in nooci instruction
FROM alpine
HEALTHCHECK CMD curl --fail http://localhost:7070 || exit 1  
COPY ./main.go 
FROM alpine
```
then
```bash
buildah bud --layers=true -f Dockerfile.nonoci --format oci -t oci-app . 
# and 
buildah bud --layers=true -f Dockerfile.nonoci --format docker -t docker-app .
```
Docker format should not use cache.

**Which issue(s) this PR fixes:**

Fixes #2389

**Special notes for your reviewer:**

Maybe we should use different cache set for OCI and Docker format, but IMO  it's wasting of disc space. 
  
**Does this PR introduce a user-facing change?**
No

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -140,8 +140,10 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budOptions) error {
 		return err
 	}
 	layers := buildahcli.UseLayers()
-	if c.Flag("layers").Changed {
-		layers = iopts.Layers
+	if strings.Contains(format, "oci") {
+		if c.Flag("layers").Changed {
+			layers = iopts.Layers
+		}
 	}
 	contextDir := ""
 	cliArgs := inputArgs


rhatdan:
@nalind WDYT?

nalind:
Ignoring the `--layers` option when we're targeting the Docker format seems a bit excessive.  I'll try to work up something to reject cache candidates which use a different format than the one we're targeting.

rhatdan:
Replaced by https://github.com/containers/buildah/pull/2734

PR #2751, opened 2020-11-06
Use CPP, CC and flags in dep check scripts

mbargull:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake

/kind other

#### What this PR does / why we need it:

Allow build system without standard `cc` to successfully run the dependency checking helper scripts from the Makefile.
This supports custom compilers specified by the common `CC` environment variable, preprocessors given as `CPP` and additional preprocessor flags from `CPPFLAGS`.
Additionally, additional flags from `CFLAGS` and `LDFLAGS` are considered for compiling/linking.
Overall, this facilitates cross-compilation and similar setups (e.g., for building Conda packages at https://gitub.com/conda-forge).

#### How to verify it

Temporarily remove `cc` from `PATH` and point `CC`/`CPP` to a different compiler.
Install header files, e.g., `btrfs/ioctl.h`, in a custom location and set `CPPFLAGS=-I/custom/location`.
Add `-Wl,--as-needed` to `LDFLAGS` (which makes library order important (again) while linking) to confirm the order of the inputs `-` and `-ldevmapper` is correct in `libdm_tag.sh`.
#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #2752, opened 2020-11-06
Avoid overriding LDFLAGS in Makefile

mbargull:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake

/kind other

`LDFLAGS` is a common/standard variable to hold linker flags for C/C++ compilers.
This PR uses a different name (`BUILDAH_LDFLAGS`) for the `-ldflags` passed on to Go so that we avoid overriding `LDFLAGS` set in the environment.

#### How to verify it

In conjunction with https://github.com/containers/buildah/pull/2751:
If, e.g., `libdevmapper` is installed at a non-standard path, adding `-L/path/to/devmapper/prefix/lib` lets the build find said library -- but only if we don't override `LDFLAGS` in the `Makefile`.

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ RUNC_COMMIT := v1.0.0-rc8
 LIBSECCOMP_COMMIT := release-2.3
 
 EXTRA_LDFLAGS ?=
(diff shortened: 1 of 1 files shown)


rhatdan:
Thanks @mbargull 
/approve
LGTM
@TomSweeneyRedHat PTAL

TomSweeneyRedHat:
LGTM

Issue #2344, opened 2020-05-05
Support force pulls


AstraLuma:
**Description**

Currently there does not seem to be a way to force `buildah pull` to go an check a registry for a fresh copy of a container.

**Steps to reproduce the issue:**
* `buildah pull --creds not:valid debian`
* `strace buildah pull debian`
* `buildah pull --help`

**Describe the results you received:**

None of these give any indication that `buildah pull` will perform network activity if the image is already available locally.

**Describe the results you expected:**

An option (eg `--force` or `--always-check`) to go and check a registry, even if the image is available locally.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
🐚 apt list buildah                                                                                                                           
Listing... Done
buildah/unknown,now 1.14.8~1 amd64 [installed]
buildah/unknown 1.14.8~1 arm64
buildah/unknown 1.14.8~1 armhf
buildah/testing 1.11.6-1 i386
buildah/unknown 1.14.8~1 ppc64el
```

**Output of `buildah version`:**

```
🐚 buildah version                                                                                                                            
Version:         1.14.8
Go Version:      go1.14.2
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.1-dev
CNI Spec:        0.4.0
libcni Version:  
image Version:   5.4.3
Git Commit:      
Built:           Fri Apr 17 23:57:07 2020
OS/Arch:         linux/amd64

```

**Output of `cat /etc/*release`:**

```
🐚 cat /etc/*release                                                                                                                          
PRETTY_NAME="Debian GNU/Linux bullseye/sid"
NAME="Debian GNU/Linux"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

**Output of `uname -a`:**

```
🐚 uname -a                                                                                                                                   
Linux cayde7 5.5.0-2-amd64 #1 SMP Debian 5.5.17-1 (2020-04-15) x86_64 GNU/Linux
```

**Output of `cat /etc/containers/storage.conf`:**

```
# This file is is the configuration file for all tools
# that use the containers/storage library.
# See man 5 containers-storage.conf for more information
# The "container storage" table contains all of the server options.
[storage]

# Default Storage Driver
driver = ""

# Temporary storage location
runroot = "/var/run/containers/storage"

# Primary Read/Write location of container storage
graphroot = "/var/lib/containers/storage"

[...]

PR #2754, opened 2020-11-06
Add --policy flag to buildah pull

rhatdan:
This allows the user to specify the pull policy for pulling images.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/cmd/buildah/pull.go
+++ b/cmd/buildah/pull.go
@@ -6,6 +6,7 @@ import (
 	"runtime"
 
 	"github.com/containers/buildah"
+	"github.com/containers/buildah/define"
 	buildahcli "github.com/containers/buildah/pkg/cli"
 	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/common/pkg/auth"
@@ -26,6 +27,7 @@ type pullOptions struct {
 	removeSignatures bool
 	tlsVerify        bool
 	decryptionKeys   []string
+	pullPolicy       string
 }
 
 func init() {
@@ -57,6 +59,7 @@ func init() {
 	flags.StringVar(&opts.blobCache, "blob-cache", "", "store copies of pulled image blobs in the specified directory")
 	flags.StringVar(&opts.certDir, "cert-dir", "", "use certificates at the specified path to access the registry")
 	flags.StringVar(&opts.creds, "creds", "", "use `[username[:password]]` for accessing the registry")
+	flags.StringVar(&opts.pullPolicy, "policy", "missing", "missing, always, or never.")
 	flags.BoolVarP(&opts.removeSignatures, "remove-signatures", "", false, "don't copy signatures when pulling image")
 	flags.StringVar(&opts.signaturePolicy, "signature-policy", "", "`pathname` of signature policy file (not usually used)")
 	flags.StringSliceVar(&opts.decryptionKeys, "decryption-key", nil, "key needed to decrypt the image")
@@ -109,6 +112,10 @@ func pullCmd(c *cobra.Command, args []string, iopts pullOptions) error {
 		return errors.Wrapf(err, "unable to obtain decrypt config")
 	}
 
+	policy, ok := define.PolicyMap[iopts.pullPolicy]
+	if !ok {
+		return fmt.Errorf("unrecognized pull policy %s", iopts.pullPolicy)
+	}
 	options := buildah.PullOptions{
 		SignaturePolicyPath: iopts.signaturePolicy,
 		Store:               store,
@@ -120,6 +127,7 @@ func pullCmd(c *cobra.Command, args []string, iopts pullOptions) error {
 		MaxRetries:          maxPullPushRetries,
 		RetryDelay:          pullPushRetryDelay,
 		OciDecryptConfig:    decConfig,
+		PullPolicy:          policy,
 	}
 
 	if iopts.quiet {
--- a/define/types.go
+++ b/define/types.go
@@ -1,6 +1,8 @@
 package define
 
-import "fmt"
+import (
+	"fmt"
+)
 
 // PullPolicy takes the value PullIfMissing, PullAlways, PullIfNewer, or PullNever.
 type PullPolicy int
@@ -38,4 +40,11 @@ func (p PullPolicy) String() string {
 		return "PullNever"
 	}
 	return fmt.Sprintf("unrecognized policy %d", p)
+}
+
+var PolicyMap = map[string]PullPolicy{
+	"missing": PullIfMissing,
+	"always":  PullAlways,
+	"never":   PullNever,
+	"ifnewer": PullIfNewer,
 }
--- a/docs/buildah-pull.md
+++ b/docs/buildah-pull.md
@@ -70,6 +70,14 @@ The [key[:passphrase]] to be used for decryption of images. Key can point to key
 
 If an image needs to be pulled from the registry, suppress progress output.
 
+**--policy**=**always**|**missing**|**never**
+
+Pull image policy. The default is **missing**.
+
+- **missing**: attempt to pull the latest image from the registries listed in registries.conf if a local image does not exist. Raise an error if the image is not in any listed registry and is not present locally.
+- **always**: Pull the image from the first registry it is found in as listed in  registries.conf. Raise an error if not found in the registries, even if the image is present locally.
+- **never**: do not pull the image from the registry, use only the local version. Raise an error if the image is not present locally.
+
 **--remove-signatures**
 
 Don't copy signatures when pulling images.
--- a/tests/pull.bats
+++ b/tests/pull.bats
@@ -266,3 +266,49 @@ load helpers
   iid=$output
   run_buildah rmi ${iid}
 }
+
+@test "pull-policy" {
+  mkdir ${TESTDIR}/buildahtest
+  run_buildah 125 pull --signature-policy ${TESTSDIR}/policy.json --policy bogus alpine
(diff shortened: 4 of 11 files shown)


rhatdan:
Fixes: https://github.com/containers/buildah/issues/2344

TomSweeneyRedHat:
Lint is not happy @rhatdan

rhatdan:
@containers/python-pypodman-maintainers PTAL
@TomSweeneyRedHat @nalind @vrothberg @giuseppe @saschagrunert PTAL

cmd/buildah/pull.go
```diff
@@ -57,6 +59,7 @@ func init() {
 	flags.StringVar(&opts.blobCache, "blob-cache", "", "store copies of pulled image blobs in the specified directory")
 	flags.StringVar(&opts.certDir, "cert-dir", "", "use certificates at the specified path to access the registry")
 	flags.StringVar(&opts.creds, "creds", "", "use `[username[:password]]` for accessing the registry")
+	flags.StringVar(&opts.pullPolicy, "policy", "missing", "missing, always, or never.")
```
nalind, line 62: Naming for the option seems inconsistent with `build-using-dockerfile`.

rhatdan: `podman pull --pull `
Feels like bad CLI, which is why I did not go with it.

docs/buildah-pull.md
```diff
@@ -70,6 +70,14 @@ The [key[:passphrase]] to be used for decryption of images. Key can point to key
 
 If an image needs to be pulled from the registry, suppress progress output.
 
+**--policy**=**always**|**missing**|**never**
+
+Pull image policy. The default is **missing**.
+
+- **missing**: attempt to pull the latest image from the registries listed in registries.conf if a local image does not exist.Raise an error if the image is not in any listed registry and is not present locally.
```
TomSweeneyRedHat, line 77:
```suggestion
- **missing**: attempt to pull the latest image from the registries listed in registries.conf if a local image does not exist. Raise an error if the image is not in any listed registry and is not present locally.
```

tests/pull.bats
```diff
@@ -249,3 +249,20 @@ load helpers
...
+@test "pull-policy" {
+  mkdir ${TESTDIR}/buildahtest
+  run_buildah 125 pull --signature-policy ${TESTSDIR}/policy.json --policy bogus alpine
+
+  run_buildah 125 pull --signature-policy ${TESTSDIR}/policy.json --policy never alpine
+  run_buildah 125 inspect alpine
+
+  run_buildah pull --signature-policy ${TESTSDIR}/policy.json alpine
+  run_buildah pull --signature-policy ${TESTSDIR}/policy.json --policy never alpine
+  run_buildah pull --signature-policy ${TESTSDIR}/policy.json --policy missing alpine
```
vrothberg, line 262: How do we check if the image hasn't been pulled?

I think we need to push an image to a local registry, pull it, then push another one to overwrite. Then a `--policy missing` should not pull the new image which we can determine by checking the digests.

tests/pull.bats
```diff
@@ -249,3 +249,20 @@ load helpers
...
+  run_buildah inspect alpine
+
+  run_buildah rmi alpine
+  run_buildah pull --signature-policy ${TESTSDIR}/policy.json --policy always alpine
```
vrothberg, line 266: Same here. We need to go deeper and check the artifacts.

vrothberg (commented):
Code LGTM but the tests are not enough to catch regressions.

tests/pull.bats
```diff
@@ -249,3 +249,20 @@ load helpers
...
+
+@test "pull-policy" {
+  mkdir ${TESTDIR}/buildahtest
+  run_buildah 125 pull --signature-policy ${TESTSDIR}/policy.json --policy bogus alpine
+
+  run_buildah 125 pull --signature-policy ${TESTSDIR}/policy.json --policy never alpine
+  run_buildah 125 inspect alpine
+
+  run_buildah pull --signature-policy ${TESTSDIR}/policy.json alpine
+  run_buildah pull --signature-policy ${TESTSDIR}/policy.json --policy never alpine
```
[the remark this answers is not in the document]
rhatdan: It should fail.

tests/pull.bats
rhatdan: I added some simple tests to make sure these are working.

rhatdan:
@TomSweeneyRedHat @vrothberg @nalind @giuseppe PTAL

I think I have hit all of the pull policy tests.

tests/pull.bats
```diff
@@ -266,3 +266,46 @@ load helpers
...
   run_buildah rmi ${iid}
 }
+
+@test "pull-policy" {
+  mkdir ${TESTDIR}/buildahtest
+  run_buildah 125 pull --signature-policy ${TESTSDIR}/policy.json --policy bogus alpine
+
+  #  If image does not exist the never will fail
+  run_buildah 125 pull --signature-policy ${TESTSDIR}/policy.json --policy never alpine
+  run_buildah 125 inspect alpine
```
TomSweeneyRedHat, line 276: I'll speak for @edsantiago, he'll probably want the error verbiage to be checked too.

edsantiago: Yes please! Thank you @TomSweeneyRedHat, I haven't been monitoring this one.

tests/pull.bats
TomSweeneyRedHat: We can check the output to see if this string "Getting image source signatures" is in it.  If it's not, then no pull.  If it is, then pull.  That probably ought to be added in here.

TomSweeneyRedHat:
I've played with it a bit this morning looks good.  Just a couple of added checks in the current tests would be nice.
I also noted you declared "PullIfNewer" in the policy, but didn't include it as an option.  I'm assuming based on discussions in the podman pull PR?  (https://github.com/containers/podman/pull/8301)

tests/pull.bats
rhatdan: Added output tests.

rhatdan:
Yes.  I don't believe it does anything.

rhatdan:
Also cleaned up some more wrapping in buildah. Errors went from

```
$ /bin/buildah inspect docker.io/library/alpine1
error reading build object "docker.io/library/alpine1": error reading image: error locating image "docker.io/library/alpine1" for importing settings: error locating image with name "docker.io/library/alpine1" ([docker.io/library/alpine1]): image not known
ERRO exit status 125 
```
alpine1 mentioned 4 times.
to
```                             
$ ./bin/buildah inspect docker.io/library/alpine1
importing settings: error locating image with name "docker.io/library/alpine1" ([docker.io/library/alpine1]): image not known
ERRO exit status 125                      
```
Only mentioned 2 times.

define/types.go
```diff
@@ -39,3 +41,10 @@ func (p PullPolicy) String() string {
 	}
 	return fmt.Sprintf("unrecognized policy %d", p)
 }
+
+var PolicyMap = map[string]PullPolicy{
+	"missing": PullIfMissing,
+	"always":  PullAlways,
+	"never":   PullNever,
+	"ifnewer": PullIfNewer,
```
edsantiago, line 49: Did you intend to document `ifnewer`?

tests/pull.bats
```diff
@@ -266,3 +266,49 @@ load helpers
...
+  run_buildah commit -q $cid docker.io/library/alpine
+  iid=$output
+
+  #  If image does not exist the never will succeed, but iid should not change
+  run_buildah pull --signature-policy ${TESTSDIR}/policy.json --policy never alpine
+  expect_output $iid
+
+  # Pull image by default should change the image id
+  run_buildah pull --policy always --signature-policy ${TESTSDIR}/policy.json alpine
+  if [[ $output == $iid ]]; then
```
edsantiago, line 293: I'm not sure this tests what you think it tests. `$output` includes stderr, so on a successful pull it will include progress messages. I think it might be better to do `if [[ $output ~ $iid ]]; then`. But I'd need to pull & build & test on my end to make sure, right now this is just a quick first pass.

tests/pull.bats
edsantiago: In preparation for shortnames work: should this be `pull ... docker.io/library/alpine` (FQIN)?

tests/pull.bats
```diff
@@ -266,3 +266,49 @@ load helpers
...
+      expect_output "[output should not be '$iid']"
+  fi
```
TomSweeneyRedHat, line 295: It would be good to have a second pull always here with a check that the pull happened rather than reusing the old local image.

tests/pull.bats
rhatdan: Added -q which should get rid of the progress.  I would rather do exact match just in case.

tests/pull.bats
rhatdan: If the pull does not happen, then $output  would == the not alpine image and cause a failure.  That is the whole idea of the test.

rhatdan:
@edsantiago @TomSweeneyRedHat @nalind @vrothberg Lets get this in, we want to release buildah 1.18.0 today.

TomSweeneyRedHat (commented):
LGTM

TomSweeneyRedHat:
@rhatdan does this need a rebase or is it good to `/lgtm` at this point?

PR #2735, opened 2020-10-31
Attempt to test `fix MkdirAll usage #2361`

rhatdan:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### Does this PR introduce a user-facing change?

```release-note

```

PR #2755, opened 2020-11-06
Run: correctly call copier.Mkdir

nalind:
#### What type of PR is this?

/kind bug

Make sure we specify the working directory that we're creating using a path that is explicitly below the chroot we want to create it under.

#### How to verify it

None

I missed this when reviewing #2735.

```
None
```


TomSweeneyRedHat:
nice catch @nalind 
LGTM

PR #2756, opened 2020-11-06
imagebuildah: use overlay for volumes when using overlay

nalind:
#### What type of PR is this?

/kind feature

When we're using the overlay driver (which means we know overlay is available), use it to make volumes appear to be writeable during RUN instructions instead of saving/restoring their contents.

This avoids having to copy the contents of the volume directory before each RUN instruction, and having to remove and extract the contents after each RUN instruction, which should be faster, particularly if the amount of content in that volume location is large.

For empty directories, it will at least avoid adding an "opaque" notation for the directory in a layer that might otherwise be empty.

Make conformance tests pay attention to the STORAGE_DRIVER environment variable, so that they use the same driver the integration tests do instead of the hardwired or configured default.

#### How to verify it

We add test passes that use overlay storage.

None

```
None
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -149,6 +149,10 @@ unit_task:
     binary_artifacts:
         path: ./bin/*
 
+    env:
+        matrix:
+            STORAGE_DRIVER: 'vfs'
+            STORAGE_DRIVER: 'overlay'
 
 conformance_task:
     name: "Docker Build Conformance"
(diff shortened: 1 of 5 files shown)


nalind:
Huh.  I guess the bot resets labels on new issues, so I shouldn't try setting them when opening PRs until after it's had its chance.  CI's hitting unit/conformance/integration tests with both vfs and overlay, so I don't need the WIP label any more.

TomSweeneyRedHat:
@nalind all kinds of test unhappiness and it looks like a rebase is needed now too.

nalind:
It looks like the tests failed because we need https://github.com/containers/storage/pull/772.

rhatdan:
LGTM

nalind:
Adjusted a test to expect an error code of 1 instead of 125.  The test was being skipped on vfs.

TomSweeneyRedHat:
@nalind errors in the test don't look flakey.

giuseppe (approved):
LGTM

rhatdan:
@nalind Please rebase

rhatdan:
@nalind Still waiting for a rebase.

TomSweeneyRedHat:
The cross compile test failed after 81 minutes and didn't leave behind any apparent bread crumbs.  I've restarted.

PR #2757, opened 2020-11-06
util: digging UID/GID out of os.FileInfo should work on Unix

nalind:
#### What type of PR is this?

/kind cleanup

Digging the UID and GID out of an os.FileInfo should work on all Unixy systems, so move it from Linux-specific implementation file to a file that we also build for Darwin.

#### How to verify it

This is heavily exercised on Linux, and cross-compile checks should handle the rest.

None

```
None
```

### DIFF
--- a/util/util_linux.go
+++ b/util/util_linux.go
@@ -1,7 +1,6 @@
 package util
 
 import (
-	"os"
 	"syscall"
 
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM
@TomSweeneyRedHat @saschagrunert PTAL

saschagrunert (approved):
LGTM

vrothberg (commented):
/lgtm

PR #2758, opened 2020-11-06
imagebuildah.getImageTypeAndHistoryAndDiffIDs: cache results

nalind:
#### What type of PR is this?

/kind other

Cache the information we look up about images when evaluating them as potential build cache matches, in an effort to reduce the number of NewImage() calls we make, because they can be expensive.

#### How to verify it

This should not impact tests.

None

```
None
```


rhatdan:
LGTM

vrothberg (approved):
/lgtm

PR #2759, opened 2020-11-06
RFC: manifest.list.Add: default to assuming variant=v8 for arm64 images

nalind:
#### What type of PR is this?

/kind other

When adding images with Architecture="arm64" to a manifest list, default to adding Variant="v8" to the description of the image's target platform, since it can be overwritten later, and it's the only valid value listed in the OCI image index spec.

#### How to verify it

This shouldn't impact testing.

None

Is this the correct thing?  If so, should we set a variant by default for 32-bit ARM ("arm") as well?

```
Adding an arm64 image to a manifest list will default to marking it for "Variant" = "v8".
```

### DIFF
--- a/manifests/manifests.go
+++ b/manifests/manifests.go
@@ -266,6 +266,18 @@ func (l *list) Add(ctx context.Context, sys *types.SystemContext, ref types.Imag
 	var instanceInfos []instanceInfo
 	var manifestDigest digest.Digest
 
+	fixupInstanceInfo := func(i instanceInfo) instanceInfo {
+		switch i.Architecture {
+		case "arm":
+			// should we have a default of v6/v7/v8 here, too?
+		case "arm64":
+			if i.Variant == "" {
+				i.Variant = "v8"
+			}
+		}
+		return i
+	}
+
 	primaryManifestBytes, primaryManifestType, err := src.GetManifest(ctx, nil)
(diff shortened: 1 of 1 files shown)


rhatdan:
@mtrmac @vrothberg PTAL

mtrmac:
Is this fixing anything specific? https://github.com/opencontainers/image-spec/blob/master/image-index.md lists it as optional.

OTOH c/image/internal/pkg/platform ’s `baseVariants` clearly shows that images with variant="v8" do show up in the wild.

IIRC looking up history of that value will point at conversations that discuss the presence/absence/desirability of a “base variant” for ARM32 as well.

---

Stylistically, hiding this bit of mostly—policy fairly deep in a long implementation piece like this feels a bit off. OTOH c/image is currently hiding the data because there wasn’t an external user and we didn’t want to commit to a stable API without one, so…

nalind:
I don't think anything's broken at this point - both dockerd and containers/image will fall back to accepting an entry with Variant="" when looking for a 64-bit image on an aarch64 system.

rhatdan:
@nalind Is this still held?

nalind:
> @nalind Is this still held?

The feedback so far is that we don't need or want to do what this patch does, and absent any other feedback, I'd lean toward closing it without merging.

PR #2764, opened 2020-11-09
Fix NPE when Dockerfile path contains non-directory entries

edigaryev:
When building `nondir/Dockerfile` and `nondir` exists in `os.Getwd()` and is not a directory, `os.Stat()` will return an error which needs to be handled, otherwise `dinfo` will be `nil` and `dinfo.Mode()` will cause panic.

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -216,20 +216,19 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
 			}
 			data = resp.Body
 		} else {
-			// If the Dockerfile isn't found try prepending the
-			// context directory to it.
 			dinfo, err := os.Stat(dfile)
-			if os.IsNotExist(err) {
-				// If they are "/workDir/Dockerfile" and "/workDir"
-				// so don't joint it
+			if err != nil {
+				// If the Dockerfile isn't available, try again with
+				// context directory prepended (if not prepended yet).
 				if !strings.HasPrefix(dfile, options.ContextDirectory) {
 					dfile = filepath.Join(options.ContextDirectory, dfile)
-				}
-				dinfo, err = os.Stat(dfile)
-				if err != nil {
-					return "", nil, err
+					dinfo, err = os.Stat(dfile)
 				}
 			}
+			if err != nil {
+				return "", nil, err
+			}
+
 			// If given a directory, add '/Dockerfile' to it.
 			if dinfo.Mode().IsDir() {
 				dfile = filepath.Join(dfile, "Dockerfile")
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -2,6 +2,11 @@
 
 load helpers
 
+@test "bud with a path to a Dockerfile (-f) containing a non-directory entry" {
+  run_buildah 125 bud -f ${TESTSDIR}/bud/non-directory-in-path/non-directory/Dockerfile
+  expect_output --substring "non-directory/Dockerfile: not a directory"
+}
+
 @test "bud with --dns* flags" {
   _prefetch alpine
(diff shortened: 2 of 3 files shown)


vrothberg (commented):
LGTM, thanks for the fix!

Would you add a test to `test/bud.bats`?

imagebuildah/build.go
```diff
@@ -216,20 +216,19 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options BuildOpt
 			}
 			data = resp.Body
 		} else {
-			// If the Dockerfile isn't found try prepending the
-			// context directory to it.
 			dinfo, err := os.Stat(dfile)
-			if os.IsNotExist(err) {
-				// If they are "/workDir/Dockerfile" and "/workDir"
-				// so don't joint it
+			if err != nil {
```
TomSweeneyRedHat, line 220: Do we really want to do this on an error such as permission denied or on a closed file? I think we want to stick with the IsNotExist() check, but you might sway me.

TomSweeneyRedHat:
@edigaryev Thanks for the PR!  A test would be awesome.  Also, I'm not 100% certain of the error case, if you could write up a simple reproducer for the description that would be great.

imagebuildah/build.go
edigaryev: Consider the following scenario:

1. A Podman API server is running (`podman system service`) in the `/usr/bin` working directory (it's `os.Getwd() == "/usr/bin"`) with the following contents:

    ```
    lrwxrwxrwx 1 root root          18 Oct 14 07:38 docker
    -rwxr-xr-x 1 root root    42348552 Oct  2 17:17 podman
    ```

2. Podman receives [`/libpod/build`](https://podman.readthedocs.io/en/latest/_static/api.html#operation/libpodBuildImage) request with a `dockerfile=docker/Dockerfile` argument and a Tar-file in it's body containing the following contents:
    ```
    drwxr-xr-x edi/edi           0 2020-11-10 19:59 docker/
    -rw-r--r-- edi/edi           0 2020-11-10 19:59 docker/Dockerfile
    ```

2. Podman extracts the Tar-file into `/tmp` and calls `BuildDockerfiles()`.

3. `BuildDockerfiles()` fails since it first tries to open the file in the current working directory and `os.IsNotExist(err) == false` (because the file actually exists, it's just not a directory), however it could've worked fine if the context directory was prepended.

While more conditions can be added to this `if`, this will be a semantically tough decision since even a permission denied error can be a valid use-case in the scenario above, where access to files other than `/tmp` is restricted by a security subsystem.

Finally, assuming that `os.Stat()` is cheap on modern OS-es, the only downside I see is that the original error might be swallowed, hindering the problem diagnosis.

edigaryev:
Added a test in https://github.com/containers/buildah/pull/2764/commits/d8165d034fda644764ae45c94bde634b2e905bd9.

imagebuildah/build.go
TomSweeneyRedHat: OK, I'm a bit confused by your point 4, and it may be due to sitting in too many meetings today.  IsNotExists doesn't care if the thing being checked for is a file or a directory, it just checks to see if either a file and a directory with that name doesn't exist.  So if the directory isn't there and a file with that name isn't there, then the old line 222 should trigger.  So I think the old code should trigger in your scenario.

My main concern is we're just overprocessing with this change, but if others think it's OK as you have it, then I'm fine.  It would be interesting to know if your new test passes with out your code change.  Did you by chance try that?

Thanks for the patience and explanations.

edigaryev: >So if the directory isn't there and a file with that name isn't there, then the old line 222 should trigger. So I think the old code should trigger in your scenario.

It won't trigger because the error is different (`os.ErrNotExist` vs `*os.PathError`), see https://play.golang.org/p/ZNr3fMQQ43d for an example.

>It would be interesting to know if your new test passes with out your code change.

It doesn't.

PR #2777, opened 2020-11-12
Run: don't forcibly disable UTS namespaces in rootless mode

nalind:
#### What type of PR is this?

/kind bug

I can't remember why we disabled UTS namespaces for rootless isolation, but it doesn't look necessary.

#### How to verify it

`buildah run --hostname foo $container uname -n` should print "foo" when run as an unprivileged user, which forces use of rootless mode.

Report came in via IRC.

```
The "--hostname" option for "buildah run" should work as expected for unprivileged users.
```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -1986,7 +1986,6 @@ func (b *Builder) configureEnvironment(g *generate.Generator, options RunOptions
 }
 
 func setupRootlessSpecChanges(spec *specs.Spec, bundleDir string, shmSize string) error {
-	spec.Hostname = ""
 	spec.Process.User.AdditionalGids = nil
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
Something must have blown up at some point to cause this?  Are we using --network namespace by default with Slirp?

rhatdan:
@vrothberg @TomSweeneyRedHat @giuseppe @ashley-cui PTAL

giuseppe (approved):
IIRC, it was because before slirp4netns integration rootless containers were using the host network namespace.

LGTM

nalind:
Yeah, it should.

Issue #2782, opened 2020-11-13
Buildah bud --layers does not provide shorter time for parallel builds of consequent containers that share some common base


bogdando:
/kind bug

**Description**

The ``--layers`` option is only useful when building *exactly the same* containers but using different target directories.
Consequent container builds that share some common base could not benefit from that layers cache. Which is quite suboptimal.

**Steps to reproduce the issue:**
1. Run concurrent ``buildah bud --layers --rm=false`` builds for multiple containers that share some common base layer
2. Compare the time it takes to the same operation executed without ``--layers --rm=false``
3. (probably expected behavior) If specifying only ``--layers`` without ``--rm=false``, it likely fails some builds with errors like ``error checking if cached image exists from a previous build: error getting history of "<id>": error creating new image from reference to image "<id>": error locating item named "manifest" for image with ID``

**Describe the results you received:**
The time it takes to build ~40 quite fat and multi-layered containers is roughly 30 minutes for the above scenarios 1 and 2.
I have the same results with buildah built from trunk master (v1.11.6-830-g31b0c5db).

See more of the similar results with using ``--layers --rm=false`` [0]
```
2020-11-12 20:05:44.891116 | TASK [build-containers : Run image build as ansible user > /home/zuul/workspace/build.log]
2020-11-12 20:36:24.390268 | primary | ok: Runtime: 0:30:38.402444
```
without ``--layers`` [1]
```
2020-11-10 01:33:15.162933 | TASK [build-containers : Run image build as ansible user > /home/zuul/workspace/build.log]
2020-11-10 01:58:09.258823 | primary | ok: Runtime: 0:24:53.062323
```

**Describe the results you expected:**
With --layers it should complete faster because of the image cache benefits.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
N/A (built locally from trunk)
```

**Output of `buildah version`:**

```
buildah version 1.18.0-dev (image-spec 1.0.1-dev, runtime-spec 1.0.2-dev)
```

**Output of `podman version` if reporting a `podman build` issue:**

```
(paste your output here)
```

**Output of `cat /etc/*release`:**

```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.5 LTS"
NAME="Ubuntu"
VERSION="18.04.5 LTS (Bionic Beaver)"
ID=ubuntu
ID_LIKE=debian
[...]

PR #2785, opened 2020-11-13
Add extra debug info and JSON logging utility

bogdando:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 

/kind feature

> /kind flake
> /kind other

This adds extra debug logs for the images cache info state and events, like found/updated an entry.
Writes details into debug logs as "safe" JSON that may also be post-processed by external pretty view tooling,
like ``jq``.
    
In order to log internal cache structures/entries as JSON, it creates an extra LogJSON utility.

#### How to verify it

To see results with added logs, use ``sudo buildah --log-level=debug bud --loglevel=3 ...``

Related #2782 

In order to make structures logged as JSON, its data fields need to be exported for JSON Marshalling to work.
This patch provides re-esporting func for a particular structure to debug it with the new logger.
Alternative approach is to change the original structure to have its data fields exported. But that would have introduced much
more impact to other places that rely on it...

```release-note
In the builah bud debug mode, extra information on images caching is available in the produced logs
```


imagebuildah/executor.go
```diff
@@ -381,6 +386,8 @@ func (b *Executor) getImageTypeAndHistoryAndDiffIDs(ctx context.Context, imageID
 		diffIDs:      oci.RootFS.DiffIDs,
 		err:          nil,
 	}
+	logrus.Debugf("updated image %s in cache:", imageID)
+	util.LogJson(imageID, b.imageInfoCache[imageID])
```
bogdando, line 390: actually, this also provides an empty {} result for ``b.imageInfoCache[imageID]``
I'm probably doing it wrong!

util/util.go
```diff
@@ -377,6 +379,23 @@ func logIfNotErrno(err error, what string, ignores ...syscall.Errno) (logged boo
...
+// Unexpected characters will be replaced by cats.
+// The logged lines then may be post-processed with external pretty log parsers.
+func LogJson(wrappingKey string, data interface{}) {
+	rawResult, err := json.Marshal(data)
+	if err != nil {
+		logrus.Fatal(err)
+	}
+	// Replace bad chars with a cat for external pretty log parsers
+	re := regexp.MustCompile(`[#=\']|\\"`)
+	safeResult := string(re.ReplaceAll([]byte(rawResult), []byte("🐱")))
```
TomSweeneyRedHat, line 393: I love [cats](https://twitter.com/TSweeneyRedHat/status/1309933044942475265), but is a cat symbol really generally used?  I was expecting an asterisk ("*") or a similar less fun character.

TomSweeneyRedHat:
@bogdando a nice start.  The cat symbol is not making the lint happy and I'm not sure that you signed this commit and our CI requires that.  If you did not, can you do `git commit --amend -s`, fetch, rebase and push again please?

imagebuildah/executor.go
TomSweeneyRedHat: Is it empty if you pass it to Debugf instead?

bogdando: it wasn't. fixed that now

imagebuildah/executor.go
```diff
@@ -350,8 +363,44 @@ func (b *Executor) waitForStage(ctx context.Context, name string, stages imagebu
 func (b *Executor) getImageTypeAndHistoryAndDiffIDs(ctx context.Context, imageID string) (string, []v1.History, []digest.Digest, error) {
 	b.imageInfoLock.Lock()
 	imageInfo, ok := b.imageInfoCache[imageID]
+
+	// Wrap unexported data in cache structures for JSON Logger
+	exporter := func(wrapper string, data interface{}) (string, envelopExecutorJSON) {
```
bogdando, line 368: Instead of re-exporting unexported structures like that, it could be "fixed" in place as well. But that would impact many other places that use the established names.

util/util.go
bogdando: done

TomSweeneyRedHat:
@bogdando looks like one of the tests are upset with your change.  Plus you need a rebase.
```
[+2456s] not ok 445 Check if containers run with correct open files/processes limits
```

bogdando:
> @bogdando looks like one of the tests are upset with your change. Plus you need a rebase.
> 
> ```
> [+2456s] not ok 445 Check if containers run with correct open files/processes limits
> ```

done

rhatdan:
@nalind PTAL

util/util.go
```diff
@@ -377,6 +379,28 @@ func logIfNotErrno(err error, what string, ignores ...syscall.Errno) (logged boo
 	return true
 }
 
+// LogJSON is a helper function to log data structures as JSON.
+// The given wrappingKey will contain the provided data.
+// Unexpected characters will be replaced by asterisks.
+// The logged lines then may be post-processed with external pretty log parsers.
+func LogJSON(wrappingKey string, data interface{}) {
+	rawResult, err := json.Marshal(data)
```
rhatdan, line 387: I think this function check the logging level, and if not in debug mode or lower return immediately.  Lots of processing to do nothing.

rhatdan:
@bogdando needs a rebase and squash commits.

rhatdan:
@bogdando are you still out there?  Could you squash so we can commit?

PR #2796, opened 2020-11-16
intermediateImageExists: ignore images whose history we can't read

nalind:
#### What type of PR is this?

/kind bug

When attempting to read the history of an image in order to try to check if it's suitable as a cache entry for an image we're building, if we fail to read its history, ignore the image instead of failing.

If the image was pulled into local storage for a different architecture, or it's a list with no entry in it for the local architecture, or if
it's a list which references an image for the local architecture that we haven't pulled down, we'll get an error back from the image library, and we don't want to fail because of that.

#### How to verify it

New integration test!

None

Stumbled onto this when looking at a podman mailing list thread about manifest lists.

```
None
```


TomSweeneyRedHat:
Fedora 31 podman test failed.  Look like a timeout flake, I've hit the re-run button.
LGTM assuming happy tests.  It this something we should try to spin a 1.18.1 for and get that into Podman v2.2

PR #2803, opened 2020-11-17
imagebuildah: disable pseudo-terminals for RUN

nalind:
#### What type of PR is this?

/kind bug

Always handle RUN instructions with no pseudo terminal, which matches what I see with docker build 19.03.  Interactive `buildah run` will still have the same default behavior.

#### How to verify it

Tests should continue to pass, and the issue described in https://github.com/containers/podman/issues/8342 should no longer occur.

None

```
RUN instructions in builds will no longer be run attached to a pseudo-terminal.
```


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

PR #2805, opened 2020-11-18
Add U volume flag to chown source volumes

EduardoVega:
#### What type of PR is this?

/kind feature

It adds the U volume flag to chown source volumes based on the uid, gid within the container. 

When a new container is created with a user namespace, we want to get the uid,gid that has been assigned outside the container to chown source volumes. This will isolate the filesystem (source volumes) from other containers (users, groups) since they will be owned by the same user and group within the container.

This is required by podman, so it can parse the U volume flag. 
Podman Issue: [7778](https://github.com/containers/podman/issues/7778)

#### How to verify it

There is an integration test to validate this functionality.
Example:
```
# Create source volume
mkdir -p /var/tmp/foo

# Create container with user namespace and mount volume using U flag
buildah from --userns-uid-map 0:1000:1024 --userns-gid-map 0:1000:1024 --volume /var/tmp/foo:/testdata:z,U --name container-foo alpine

# Ensure owner of the volume is 0:0
buildah run container-foo stat -c "%u:%g" /testdata

# Ensure source volume has been chowned based on user namespace mapping
stat -c "%u:%g" /var/tmp/foo
```

None

None

Signed-off-by: Eduardo Vega <edvegavalerio@gmail.com>

### DIFF
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -581,6 +581,7 @@ process.
    container. The `OPTIONS` are a comma delimited list and can be: <sup>[[1]](#Footnote1)</sup>
 
    * [rw|ro]
+   * [U]
    * [z|Z|O]
    * [`[r]shared`|`[r]slave`|`[r]private`]
 
@@ -593,9 +594,17 @@ and bind mounts that into the container.
 You can specify multiple  **-v** options to mount one or more mounts to a
 container.
 
+  `Write Protected Volume Mounts`
+
 You can add the `:ro` or `:rw` suffix to a volume to mount it read-only or
 read-write mode, respectively. By default, the volumes are mounted read-write.
 See examples.
+
+  `Chowning Volume Mounts`
+
+By default, Buildah does not change the owner and group of source volume directories mounted into containers. If a container is created in a new user namespace, the UID and GID in the container may correspond to another UID and GID on the host.
+
+The `:U` suffix tells Buildah to use the correct host UID and GID based on the UID and GID within the container, to change the owner and group of the source volume.
 
   `Labeling Volume Mounts`
 
@@ -688,6 +697,8 @@ buildah bud --memory 40m --cpu-period 10000 --cpu-quota 50000 --ulimit nofile=10
 buildah bud --security-opt label=level:s0:c100,c200 --cgroup-parent /path/to/cgroup/parent -t imageName .
 
 buildah bud --volume /home/test:/myvol:ro,Z -t imageName .
+
+buildah bud -v /home/test:/myvol:z,U -t imageName .
 
 buildah bud -v /var/lib/dnf:/var/lib/dnf:O -t imageName .
 
--- a/docs/buildah-from.md
+++ b/docs/buildah-from.md
@@ -449,6 +449,7 @@ process.
    container. The `OPTIONS` are a comma delimited list and can be: <sup>[[1]](#Footnote1)</sup>
 
    * [rw|ro]
+   * [U]
    * [z|Z|O]
    * [`[r]shared`|`[r]slave`|`[r]private`|`[r]unbindable`]
 
@@ -461,9 +462,17 @@ and bind mounts that into the container.
 You can specify multiple  **-v** options to mount one or more mounts to a
 container.
 
+  `Write Protected Volume Mounts`
+
 You can add the `:ro` or `:rw` suffix to a volume to mount it read-only or
 read-write mode, respectively. By default, the volumes are mounted read-write.
 See examples.
+
+  `Chowning Volume Mounts`
+
+By default, Buildah does not change the owner and group of source volume directories mounted into containers. If a container is created in a new user namespace, the UID and GID in the container may correspond to another UID and GID on the host.
+
+The `:U` suffix tells Buildah to use the correct host UID and GID based on the UID and GID within the container, to change the owner and group of the source volume.
 
   `Labeling Volume Mounts`
 
@@ -552,6 +561,8 @@ buildah from --memory 40m --cpu-shares 2 --cpuset-cpus 0,2 --security-opt label=
 buildah from --ulimit nofile=1024:1028 --cgroup-parent /path/to/cgroup/parent myregistry/myrepository/imagename:imagetag
 
 buildah from --volume /home/test:/myvol:ro,Z myregistry/myrepository/imagename:imagetag
+
+buildah from -v /home/test:/myvol:z,U myregistry/myrepository/imagename:imagetag
 
 buildah from -v /var/lib/yum:/var/lib/yum:O myregistry/myrepository/imagename:imagetag
 
--- a/docs/buildah-run.md
+++ b/docs/buildah-run.md
@@ -188,6 +188,7 @@ bind mounts `/HOST-DIR` in the host to `/CONTAINER-DIR` in the Buildah
 container. The `OPTIONS` are a comma delimited list and can be: <sup>[[1]](#Footnote1)</sup>
 
    * [rw|ro]
+   * [U]
    * [z|Z]
    * [`[r]shared`|`[r]slave`|`[r]private`]
 
@@ -200,9 +201,19 @@ and bind mounts that into the container.
 You can specify multiple  **-v** options to mount one or more mounts to a
 container.
 
+  `Write Protected Volume Mounts`
+
 You can add the `:ro` or `:rw` suffix to a volume to mount it read-only or
 read-write mode, respectively. By default, the volumes are mounted read-write.
 See examples.
+
+  `Chowning Volume Mounts`
+
+By default, Buildah does not change the owner and group of source volume directories mounted into containers. If a container is created in a new user namespace, the UID and GID in the container may correspond to another UID and GID on the host.
+
+The `:U` suffix tells Buildah to use the correct host UID and GID based on the UID and GID within the container, to change the owner and group of the source volume.
+
+  `Labeling Volume Mounts`
 
 Labeling systems like SELinux require that proper labels are placed on volume
 content mounted into a container. Without a label, the security system might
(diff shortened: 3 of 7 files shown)


TomSweeneyRedHat:
@EduardoVega thanks for the PR.  The tests don't look happy and it looks like you need to rebase.
@giuseppe PTAL

TomSweeneyRedHat:
Need man page changes.

EduardoVega:
I have updated docs.

run_linux.go
```diff
@@ -1736,6 +1738,13 @@ func (b *Builder) runSetupVolumeMounts(mountLabel string, volumeMounts []string,
 			}
 			return overlayMount, err
 		}
+		if foundU {
+			if err := filepath.Walk(host, func(filePath string, f os.FileInfo, err error) error {
```
rhatdan, line 1742: In the SELinux relabel, we check to make sure that the user is not "relabeling" a system path, Since this could cause major issues.

https://github.com/opencontainers/selinux/blob/master/go-selinux/label/label_selinux.go#L112

Do we want to do the same here?

run_linux.go
```diff
@@ -1736,6 +1738,13 @@ func (b *Builder) runSetupVolumeMounts(mountLabel string, volumeMounts []string,
...
+				return os.Lchown(filePath, rootUID, rootGID)
```
rhatdan, line 1743: Would we want this UID and GID to respect the --user flag of the container?

For example, if I am in a rootless container and I execute a command like

podman run -ti --user mariadb --userns=keep-id ~/data:/var/lib/data:ZU -d mariadb 

I would want the ~/data directory labeled with the mariadb user inside of the container.  This is a very difficult situation for the user to figure out what the correct UID for the data directory would be.

rhatdan:
@giuseppe PTAL

run_linux.go
EduardoVega: I can check the current uid, gid before doing the chown operation. I'll add that. Thanks

run_linux.go
EduardoVega: In podman that's how it will work. There's an execUser variable that holds the correct UID and GID

```
./bin/podman run -d -e MYSQL_ROOT_PASSWORD=root --user mysql --userns=keep-id -v ~/data:/var/lib/mysql:Z,U --name foo mysql

stat -c "%u:%g" ~/data
100999:100999
```

In the case of buildah, I am making a mistake since I'm using rootUID, rootGID. I'll work on getting the correct UID,GID if --user is specified. Thanks

giuseppe:
LGTM once the comment with `--user` is addressed

docs/buildah-bud.md
```diff
@@ -593,10 +594,18 @@ and bind mounts that into the container.
...
 
+  `Write Protected Volume Mounts`
+
 You can add the `:ro` or `:rw` suffix to a volume to mount it read-only or
 read-write mode, respectively. By default, the volumes are mounted read-write.
 See examples.
 
+  `Chowning Volume Mounts`
+
+By default buildah does not change the owner and group of source volume directories mounted into containers. If a container is created in a new user namespace, the UID and GID in the container may correspond to another UID and GID on the host.
```
TomSweeneyRedHat, line 605: FWIW 'buildah' only when talking about an actual command.

```suggestion
By default, Buildah does not change the owner and group of source volume directories mounted into containers. If a container is created in a new user namespace, the UID and GID in the container may correspond to another UID and GID on the host.
```

docs/buildah-bud.md
```diff
@@ -593,10 +594,18 @@ and bind mounts that into the container.
...
+
+The `:U` suffix tells buildah to use the correct host UID and GID based on the UID and GID within the container, to change the owner and group of the source volume.
```
TomSweeneyRedHat, line 607:
```suggestion
The `:U` suffix tells Buildah to use the correct host UID and GID based on the UID and GID within the container, to change the owner and group of the source volume.
```

docs/buildah-from.md
```diff
@@ -461,10 +462,18 @@ and bind mounts that into the container.
...
 
+  `Write Protected Volume Mounts`
+
 You can add the `:ro` or `:rw` suffix to a volume to mount it read-only or
 read-write mode, respectively. By default, the volumes are mounted read-write.
 See examples.
 
+  `Chowning Volume Mounts`
+
+By default buildah does not change the owner and group of source volume directories mounted into containers. If a container is created in a new user namespace, the UID and GID in the container may correspond to another UID and GID on the host.
```
TomSweeneyRedHat, line 473:
```suggestion
By default, Buildah does not change the owner and group of source volume directories mounted into containers. If a container is created in a new user namespace, the UID and GID in the container may correspond to another UID and GID on the host.
```

docs/buildah-from.md
```diff
@@ -461,10 +462,18 @@ and bind mounts that into the container.
...
+
+The `:U` suffix tells buildah to use the correct host UID and GID based on the UID and GID within the container, to change the owner and group of the source volume.
```
TomSweeneyRedHat, line 475:
```suggestion
The `:U` suffix tells Buildah to use the correct host UID and GID based on the UID and GID within the container, to change the owner and group of the source volume.
```

docs/buildah-run.md
```diff
@@ -200,10 +201,20 @@ and bind mounts that into the container.
...
 
+  `Write Protected Volume Mounts`
+
 You can add the `:ro` or `:rw` suffix to a volume to mount it read-only or
 read-write mode, respectively. By default, the volumes are mounted read-write.
 See examples.
 
+  `Chowning Volume Mounts`
+
+By default buildah does not change the owner and group of source volume directories mounted into containers. If a container is created in a new user namespace, the UID and GID in the container may correspond to another UID and GID on the host.
```
TomSweeneyRedHat, line 212:
```suggestion
By default, Buildah does not change the owner and group of source volume directories mounted into containers. If a container is created in a new user namespace, the UID and GID in the container may correspond to another UID and GID on the host.
```

docs/buildah-run.md
```diff
@@ -200,10 +201,20 @@ and bind mounts that into the container.
...
+
+The `:U` suffix tells buildah to use the correct host UID and GID based on the UID and GID within the container, to change the owner and group of the source volume.
```
TomSweeneyRedHat, line 214:
```suggestion
The `:U` suffix tells Buildah to use the correct host UID and GID based on the UID and GID within the container, to change the owner and group of the source volume.
```

TomSweeneyRedHat:
A few doc nits and a rebase is needed.  Otherwise LGTM
Thanks @EduardoVega !

TomSweeneyRedHat:
@EduardoVega thanks for the updates.  Hover it looks like you need to rebase and beyond that, it looks like one of the tests flaked and the rebase will hopefully clear that too.

docs/buildah-bud.md
```diff
@@ -622,7 +631,7 @@ Only the current container can use a private volume.
 
   Note:
 
-     - The `O` flag is not allowed to be specified with the `Z` or `z` flags. Content mounted into the container is labeled with the private label.
+     - The `O` flag is not allowed to be specified with the `Z`, `z` or `U` flags. Content mounted into the container is labeled with the private label.
```
rhatdan, line 634: Is there a reason to not allow Overlay options with the U Flag?  In SELinux the kernel will prevent this.  Might not be a good idea because it will trigger a copy-up, but it might be the case where content is intended to be used for the run of the container temporarily but not written to the host.  Of course this would mean the chown would need to happen on each run of the container.

run_linux.go
```diff
@@ -1789,6 +1802,27 @@ func (b *Builder) runSetupVolumeMounts(mountLabel string, volumeMounts []string,
...
+	if err != nil {
+		return err
+	}
+
+	currentUID := int(fi.Sys().(*syscall.Stat_t).Uid)
+	currentGID := int(fi.Sys().(*syscall.Stat_t).Gid)
+
+	if UID != currentUID || GID != currentGID {
+		if err := filepath.Walk(path, func(filePath string, f os.FileInfo, err error) error {
+			return os.Lchown(filePath, UID, GID)
```
rhatdan, line 1817: Should we ignore NOEXIST errors, since this is potentially racy., where a file could have been removed from underlying storage?

rhatdan (approved):

docs/buildah-bud.md
EduardoVega: Thanks. Let me work on this.

run_linux.go
EduardoVega: Same here. I'll work on this. Thanks

EduardoVega: The NOEXIST error is now being ignored for the stat operation. In this case the chmod operation will be skipped.

docs/buildah-bud.md
EduardoVega: Currently buildah will chown the source volume and then use the processUID and processGID to create the overlay temp dir, would this be the expected behaviour? Initially I tried to chown the overlay temp directory only, in the case of rootless containers the UID/GID were added correctly, but in the case of rootful the content of the source volume was never copied up to the overlay temp dir so the correct UID/GID were never added to sub dirs/files.

rhatdan: If you chown the files after the volume is mounted, it should create new INODES in the upper directory with the correct ownership.

EduardoVega: It seems to work with rootless containers since the volume is mounted using `fuse-overlayfs`  here https://github.com/containers/buildah/blob/master/pkg/overlay/overlay.go#L64, but for containers created by root the overlay volume is mounted by crun using the native overlay type.

If I am not wrong there's no way to specify uid/gid as an option for overlay native mount type. The only way I can see this working is if the real source volume is chown just like volumes without the O flag.

rhatdan:
@EduardoVega Please rebase to get past the ubuntu failures.

rhatdan:
/lgtm
Thanks @EduardoVega

Issue #2761, opened 2020-11-08
RFE: Add --ignore/-i support to the build command (Original title: Podman's dockerignore build context support?)


dm17:
As described in this issue: https://github.com/moby/moby/issues/12886
Does Podman support this same functionality?
Thanks!

TomSweeneyRedHat:
@dm17 We don't support that from what I can see, but I'll let @QiWang19 correct me if I'm mistaken.
I'm going to change the title of this issue and make it a request to add the functionality.  It looks like it would be doable.

For a quick thumbnail sketch, this is requesting adding the `--ignore/-i` option to the build command which would allow the user to specify a particular `.dockerignore` file to use rather than the default.

dm17:
Thanks! And moreover, I'm investigating how build contexts work when there are multiple Dockerfiles and potentially multiple docker-compose.yml files within a repository - and how one builds them using docker-compose (podman-compose) build.

rhatdan:
According the the Moby link above.  You would like us to use Dockerfile.dockerignore and Containerfile.dockerignore

IE if I do a buildah bud -f /PATHTO/foobar /tmp, then we should grab
/PATHTO/foobar.dockerignore if it exists, otherwise we would look for /tmp/.dockerignore.

buildah bud and podman build has `--ignorefile` now.

rhatdan:
This is fixed.

PR #2807, opened 2020-11-19
Add --ignorefile flag to use alternate .dockerignore flags

rhatdan:
Allow users to override location of the .dockerignore file.
If user specified an --ignorefile flag, buildah will read the
file and pass in the exclude lines based on the .dockerignore
contents.

Also add a --contextdir flag to buildah copy and buildah add to
specify where the context directory is located.  For these two
commands the --ignorefile flag requires the --contextdir flag.

When the --ignorefile flag is passed in, the .dockerignore files
in the context directory will be ignored.

### DIFF
--- a/cmd/buildah/addcopy.go
+++ b/cmd/buildah/addcopy.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
+	"strings"
 
 	"github.com/containers/buildah"
 	buildahcli "github.com/containers/buildah/pkg/cli"
@@ -13,6 +15,8 @@ type addCopyResults struct {
 	addHistory bool
 	chown      string
 	quiet      bool
+	ignoreFile string
+	contextdir string
 }
 
 func init() {
@@ -52,6 +56,8 @@ func init() {
 	addFlags.SetInterspersed(false)
 	addFlags.BoolVar(&addOpts.addHistory, "add-history", false, "add an entry for this operation to the image's history.  Use BUILDAH_HISTORY environment variable to override. (default false)")
 	addFlags.StringVar(&addOpts.chown, "chown", "", "set the user and group ownership of the destination content")
+	addFlags.StringVar(&addOpts.contextdir, "contextdir", "", "context directory path")
+	addFlags.StringVar(&addOpts.ignoreFile, "ignorefile", "", "path to .dockerignore file")
 	addFlags.BoolVarP(&addOpts.quiet, "quiet", "q", false, "don't output a digest of the newly-added/copied content")
 
 	// TODO We could avoid some duplication here if need-be; given it is small, leaving as is
@@ -59,6 +65,8 @@ func init() {
 	copyFlags.SetInterspersed(false)
 	copyFlags.BoolVar(&copyOpts.addHistory, "add-history", false, "add an entry for this operation to the image's history.  Use BUILDAH_HISTORY environment variable to override. (default false)")
 	copyFlags.StringVar(&copyOpts.chown, "chown", "", "set the user and group ownership of the destination content")
+	copyFlags.StringVar(&copyOpts.ignoreFile, "ignorefile", "", "path to .dockerignore file")
+	copyFlags.StringVar(&copyOpts.contextdir, "contextdir", "", "context directory path")
 	copyFlags.BoolVarP(&copyOpts.quiet, "quiet", "q", false, "don't output a digest of the newly-added/copied content")
 
 	rootCmd.AddCommand(addCommand)
@@ -100,7 +108,19 @@ func addAndCopyCmd(c *cobra.Command, args []string, verb string, extractLocalArc
 	builder.ContentDigester.Restart()
 
 	options := buildah.AddAndCopyOptions{
-		Chown: iopts.chown,
+		Chown:      iopts.chown,
+		ContextDir: iopts.contextdir,
+	}
+	if iopts.ignoreFile != "" {
+		if iopts.contextdir == "" {
+			return errors.Errorf("--ignore options requires that you specify a context dir using --contextdir")
+		}
+
+		excludes, err := parseDockerignore(iopts.ignoreFile)
+		if err != nil {
+			return err
+		}
+		options.Excludes = excludes
 	}
 
 	err = builder.Add(dest, extractLocalArchives, options, args...)
@@ -125,4 +145,19 @@ func addCmd(c *cobra.Command, args []string, iopts addCopyResults) error {
 
 func copyCmd(c *cobra.Command, args []string, iopts addCopyResults) error {
 	return addAndCopyCmd(c, args, "COPY", false, iopts)
+}
+
+func parseDockerignore(ignoreFile string) ([]string, error) {
+	var excludes []string
+	ignore, err := ioutil.ReadFile(ignoreFile)
+	if err != nil {
+		return excludes, err
+	}
+	for _, e := range strings.Split(string(ignore), "\n") {
+		if len(e) == 0 || e[0] == '#' {
+			continue
+		}
+		excludes = append(excludes, e)
+	}
+	return excludes, nil
 }
--- a/contrib/completions/bash/buildah
+++ b/contrib/completions/bash/buildah
@@ -418,6 +418,7 @@ return 1
      --file
      --format
      --http-proxy
+     --ignorefile
      --iidfile
      --isolation
      --ipc
@@ -536,6 +537,8 @@ return 1
 
      local options_with_args="
      --chown
+     --contextdir
+     --ignorefile
   "
 
      local all_options="$options_with_args $boolean_options"
@@ -558,6 +561,8 @@ return 1
 
      local options_with_args="
      -chown
+     --contextdir
+     --ignorefile
   "
 
      local all_options="$options_with_args $boolean_options"
--- a/docs/buildah-add.md
+++ b/docs/buildah-add.md
@@ -27,6 +27,16 @@ BUILDAH\_HISTORY environment variable. `export BUILDAH_HISTORY=true`
 
 Sets the user and group ownership of the destination content.
 
+**--contextdir**
+
+Build context directory. Specifying a context directory causes Buildah to
+chroot into that context directory. This means copying files pointed at
+by symbolic links outside of the chroot will fail.
+
+**--ignorefile**
+
+Path to an alternative .dockerignore file. Requires --contextdir be specified.
+
 **--quiet**, **-q**
 
 Refrain from printing a digest of the added content.
@@ -49,13 +59,13 @@ buildah add containerID 'passwd' 'certs.d' /etc
 
 ## FILES
 
-### `.dockerignore`
+### .dockerignore
 
-If the file .dockerignore exists in the context directory, `buildah copy` reads
-its contents. Buildah uses the content to exclude files and directories from
-the context directory, when copying content into the image.
+When the \fB\fC\-\-ignorefile\fR option is specified Buildah reads the
+content to exclude files and directories from the source directory, when
+copying content into the image.
 
-Users can specify a series of Unix shell globals in a .dockerignore file to
+Users can specify a series of Unix shell globals in a inore file to
 identify files/directories to exclude.
 
 Buildah supports a special wildcard string `**` which matches any number of
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -280,6 +280,10 @@ those.
 
 Write the image ID to the file.
 
+**--ignorefile**
+
+Path to an alternative .dockerignore file.
+
 **--ipc** *how*
 
 Sets the configuration for IPC namespaces when handling `RUN` instructions.
@@ -733,9 +737,10 @@ are stored while pulling and pushing images.  Defaults to '/var/tmp'.
 
 ### `.dockerignore`
 
-If the file .dockerignore exists in the context directory, `buildah bud` reads
-its contents. Buildah uses the content to exclude files and directories from
-the context directory, when executing COPY and ADD directives in the
+If the file .dockerignore exists in the context directory, `buildah copy` reads
+its contents. Use `--ignorefile` flag to overide .dockerignore path location.
+Buildah uses the content to exclude files and directories from the context
+directory, when executing COPY and ADD directives in the
 Containerfile/Dockerfile
 
 Users can specify a series of Unix shell globals in a .dockerignore file to
--- a/docs/buildah-copy.md
+++ b/docs/buildah-copy.md
@@ -25,6 +25,16 @@ BUILDAH\_HISTORY environment variable. `export BUILDAH_HISTORY=true`
 
 Sets the user and group ownership of the destination content.
 
+**--contextdir**
+
+Build context directory. Specifying a context directory causes Buildah to
+chroot into a the context directory. This means copying files pointed at
+by symbolic links outside of the chroot will fail.
+
+**--ignorefile**
+
+Path to an alternative .dockerignore file. Requires --contextdir be specified.
+
 **--quiet**, **-q**
 
 Refrain from printing a digest of the copied content.
@@ -47,13 +57,13 @@ buildah copy containerID 'passwd' 'certs.d' /etc
 
 ## FILES
 
-### `.dockerignore`
+### .dockerignore
 
-If the file .dockerignore exists in the context directory, `buildah copy` reads
-its contents. Buildah uses the content to exclude files and directories from
-the context directory, when copying content into the image.
+When the \fB\fC\-\-ignorefile\fR option is specified Buildah reads the
+content to exclude files and directories from the source directory, when
+copying content into the image.
 
-Users can specify a series of Unix shell globals in a .dockerignore file to
+Users can specify a series of Unix shell globals in a inore file to
 identify files/directories to exclude.
 
 Buildah supports a special wildcard string `**` which matches any number of
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -185,6 +185,8 @@ type BuildOptions struct {
 	Jobs *int
 	// LogRusage logs resource usage for each step.
 	LogRusage bool
+	// Excludes is a list of excludes to be used instead of the .dockerignore file.
+	Excludes []string
 }
 
 // BuildDockerfiles parses a set of one or more Dockerfiles (which may be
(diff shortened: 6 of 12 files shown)


rhatdan:
@nalind PTAL
Does this make sense?

rhatdan:
Fixes: https://github.com/containers/buildah/issues/2761

cmd/buildah/addcopy.go
```diff
@@ -126,3 +138,18 @@ func addCmd(c *cobra.Command, args []string, iopts addCopyResults) error {
 func copyCmd(c *cobra.Command, args []string, iopts addCopyResults) error {
 	return addAndCopyCmd(c, args, "COPY", false, iopts)
 }
+
+func parseDockerignore(ignoreFile string) ([]string, error) {
+	var excludes []string
+	ignore, err := ioutil.ReadFile(ignoreFile)
+	if err != nil {
+		return excludes, err
```
nalind, line 146: Should this error be wrapped?

cmd/buildah/addcopy.go
```diff
@@ -52,13 +55,15 @@ func init() {
...
 	addFlags.StringVar(&addOpts.chown, "chown", "", "set the user and group ownership of the destination content")
+	addFlags.StringVar(&addOpts.ignore, "ignore", "", "path to .dockerignore file")
 	addFlags.BoolVarP(&addOpts.quiet, "quiet", "q", false, "don't output a digest of the newly-added/copied content")
 
 	// TODO We could avoid some duplication here if need-be; given it is small, leaving as is
 	copyFlags := copyCommand.Flags()
 	copyFlags.SetInterspersed(false)
 	copyFlags.BoolVar(&copyOpts.addHistory, "add-history", false, "add an entry for this operation to the image's history.  Use BUILDAH_HISTORY environment variable to override. (default false)")
 	copyFlags.StringVar(&copyOpts.chown, "chown", "", "set the user and group ownership of the destination content")
+	copyFlags.StringVar(&copyOpts.ignore, "ignore", "", "path to .dockerignore file")
```
nalind, line 66: Please rename this to `dockerignore`, so that if we ever add an option to specify values like the ones we'd see in the file, but specified directly on the command line, we can call that `ignore`.

docs/buildah-add.md
```diff
@@ -52,8 +56,9 @@ buildah add containerID 'passwd' 'certs.d' /etc
 ### `.dockerignore`
 
 If the file .dockerignore exists in the context directory, `buildah copy` reads
```
nalind, line 58: There's no context directory for `buildah add`...?

docs/buildah-copy.md
```diff
@@ -50,8 +54,9 @@ buildah copy containerID 'passwd' 'certs.d' /etc
 ### `.dockerignore`
 
 If the file .dockerignore exists in the context directory, `buildah copy` reads
```
nalind, line 56: There's no context directory for `buildah copy`...?

imagebuildah/build.go
```diff
@@ -185,6 +185,8 @@ type BuildOptions struct {
 	Jobs *int
 	// LogRusage logs resource usage for each step.
 	LogRusage bool
+	// Excludes is the contents of the .dockerignore file.
```
nalind, line 188: ... to be used _instead_ of the contents of the .dockerignore file.  This doesn't make it clear that the caller can leave it unset to let the default thing happen.

cmd/buildah/addcopy.go
nalind: Or `ignorefile`, I guess.

TomSweeneyRedHat: I prefer `--ignorefile`

docs/buildah-add.md
```diff
@@ -52,8 +56,9 @@ buildah add containerID 'passwd' 'certs.d' /etc
...
-its contents. Buildah uses the content to exclude files and directories from
-the context directory, when copying content into the image.
+its contents. Use `--ignore` flag to overide .dockerignore path location.
```
TomSweeneyRedHat, line 59:
```suggestion
its contents. Use `--ignore` flag to overide the .dockerignore path location.
```

docs/buildah-bud.md
```diff
@@ -731,9 +735,10 @@ are stored while pulling and pushing images.  Defaults to '/var/tmp'.
 
 ### `.dockerignore`
 
-If the file .dockerignore exists in the context directory, `buildah bud` reads
-its contents. Buildah uses the content to exclude files and directories from
-the context directory, when executing COPY and ADD directives in the
+If the file .dockerignore exists in the context directory, `buildah copy` reads
+its contents. Use `--ignore` flag to overide .dockerignore path location.
```
TomSweeneyRedHat, line 739:
```suggestion
its contents. Use `--ignore` option to override .dockerignore path location.
```

docs/buildah-copy.md
```diff
@@ -50,8 +54,9 @@ buildah copy containerID 'passwd' 'certs.d' /etc
...
-its contents. Buildah uses the content to exclude files and directories from
-the context directory, when copying content into the image.
+its contents. Use `--ignore` flag to overide .dockerignore path location.
```
TomSweeneyRedHat, line 57:
```suggestion
its contents. Use `--ignore` option to override .dockerignore path location.
```

cmd/buildah/addcopy.go
rhatdan: No, ioutil.ReadFile brings back a good error messages including the ignorefile.  (Previous version had it wrapped, and all it did was stutter.

nalind: Some of those are probably candidates for `errors.WithStack()`.

docs/buildah-add.md
rhatdan: I changed the text to indicate this only happens if you specify the --ignorefile option.  Currently there is no .dockerignore used in buildah add or buildah copy, if I understand correctly.  Adding the --ignorefile option will allow users to take advantage of this functionality.

nalind: That's right, since they don't call Add() with a context directory.

imagebuildah/build.go
TomSweeneyRedHat: excellent point @nalind

rhatdan: Fixed

rhatdan:
@nalind I need some help debugging this.  I must be doing something wrong because the excludes never seem to fire unless I exclude everything on buildah copy?

nalind:
Oh right, a .dockerignore file is always evaluated relative to the context directory.  As a starting point, I'm guessing you'll want to set that to the directory that contains the file.  That will also cause us to chroot() into the directory, so symlinks that point out of that directory will be interpreted differently.  Either that, or the logic we're adding will need to prepend some path to the patterns that it's reading, taking care to correctly handle exclusions, which are patterns that begin with `!`.

docs/buildah-bud.md
```diff
@@ -733,9 +737,10 @@ are stored while pulling and pushing images.  Defaults to '/var/tmp'.
 
 ### `.dockerignore`
 
-If the file .dockerignore exists in the context directory, `buildah bud` reads
-its contents. Buildah uses the content to exclude files and directories from
-the context directory, when executing COPY and ADD directives in the
+If the file .dockerignore exists in the context directory, `buildah copy` reads
```
nalind, line 740: There's no context directory for `buildah copy`, same as with `buildah add`.

docs/buildah-copy.md
```diff
@@ -49,9 +53,9 @@ buildah copy containerID 'passwd' 'certs.d' /etc
 
 ### `.dockerignore`
 
-If the file .dockerignore exists in the context directory, `buildah copy` reads
-its contents. Buildah uses the content to exclude files and directories from
-the context directory, when copying content into the image.
+If you specify the \fB\fC\-\-ignorefile\fR option to point to a .dockerignore
+path location. Buildah uses the content to exclude files and directories from
+the source directory, when copying content into the image.
```
nalind, line 58: What's the source directory?  How do I know what it is?

rhatdan: Would it be the directory specified or the parent directory of the file, if a file is specified in the command line

```
$ buildah copy CID --ignore /tmp/.containerignore /src /dest # /src is context dir
$ buildah copy CID --ignore /tmp/.containerignore /src/file /dest # /src is context dir
```

nalind: That answer feels a little hand-wavy to me, but it would need to be in the documentation.

rhatdan: Decided to not have a default, and require --context dir with buildah COPY/ADD --ignorefile,

rhatdan:
@nalind @TomSweeneyRedHat @vrothberg PTAL

docs/buildah-add.md
```diff
@@ -27,6 +27,15 @@ BUILDAH\_HISTORY environment variable. `export BUILDAH_HISTORY=true`
...
 Sets the user and group ownership of the destination content.
 
+**--contextdir**
+
+Build context directory. Specifying a context directory, causes Buildah to
+chroot into a the context directory. This means copying files pointed at by symbolic links outside of the chroot will fail.
+
+**--ignorefile**
+
+Path to alternative .dockerignore file. Requires --contextdir be specified.
```
nalind, line 37: If this is a relative path, where should it be relative to?

docs/buildah-add.md
```diff
@@ -27,6 +27,15 @@ BUILDAH\_HISTORY environment variable. `export BUILDAH_HISTORY=true`
 
 Sets the user and group ownership of the destination content.
 
+**--contextdir**
+
+Build context directory. Specifying a context directory, causes Buildah to
```
TomSweeneyRedHat, line 32:
```suggestion
Build context directory. Specifying a context directory causes Buildah to
```

docs/buildah-add.md
```diff
@@ -27,6 +27,15 @@ BUILDAH\_HISTORY environment variable. `export BUILDAH_HISTORY=true`
...
+chroot into a the context directory. This means copying files pointed at by symbolic links outside of the chroot will fail.
```
TomSweeneyRedHat, line 33:
```suggestion
chroot into that context directory. This means copying files pointed at by symbolic links outside of the chroot will fail.
```

docs/buildah-add.md
rhatdan: I am not sure what the question is?  Whether or not the file is relative does not mapper.  We are just opening and reading the file, and setting the options.Excludes.

docs/buildah-add.md
TomSweeneyRedHat: "outside of the chroot will fail" or "above the context directory will fail"?

docs/buildah-bud.md
rhatdan: Yes, but this is the buildah bud man page.  The copy and add man pages point out that the --contextdir option is required to use --ignorefile.

docs/buildah-add.md
TomSweeneyRedHat: Good point Nalin, might be best to say fully specified directory or some such.

docs/buildah-add.md
```diff
@@ -52,8 +61,9 @@ buildah add containerID 'passwd' 'certs.d' /etc
 ### `.dockerignore`
 
 If the file .dockerignore exists in the context directory, `buildah copy` reads
-its contents. Buildah uses the content to exclude files and directories from
-the context directory, when copying content into the image.
+its contents. Use `--ignorefile` option to overide the .dockerignore path location.
```
TomSweeneyRedHat, line 64:
```suggestion
its contents. Use the `--ignorefile` option to override the .dockerignore path location.
```

docs/buildah-add.md
```diff
@@ -52,8 +61,9 @@ buildah add containerID 'passwd' 'certs.d' /etc
...
+Buildah uses the content to exclude files and directories from source
```
TomSweeneyRedHat, line 65: suggest:
```suggestion
Buildah uses the content of the specified `.dockerignore` file to exclude files and directories from source
```

docs/buildah-bud.md
```diff
@@ -280,6 +280,10 @@ those.
 
 Write the image ID to the file.
 
+**--ignorefile**
+
+Path to alternative .dockerignore file.
```
TomSweeneyRedHat, line 285:
```suggestion
Path to an alternative .dockerignore file.
```

docs/buildah-add.md
```diff
@@ -27,6 +27,15 @@ BUILDAH\_HISTORY environment variable. `export BUILDAH_HISTORY=true`
...
 Sets the user and group ownership of the destination content.
 
+**--contextdir**
+
+Build context directory. Specifying a context directory causes Buildah to
+chroot into a the context directory. This means copying files pointed at by symbolic links outside of the chroot will fail.
+
+**--ignorefile**
+
+Path to alternative .dockerignore file. Requires --contextdir be specified.
```
TomSweeneyRedHat, line 37:
```suggestion
Path to an alternative .dockerignore file. Requires --contextdir be specified.
```

docs/buildah-copy.md
```diff
@@ -25,6 +25,15 @@ BUILDAH\_HISTORY environment variable. `export BUILDAH_HISTORY=true`
...
 Sets the user and group ownership of the destination content.
 
+**--contextdir**
+
+Build context directory. Specifying a context directory, causes Buildah to
+chroot into a the context directory. This means copying files pointed at by symbolic links outside of the chroot will fail.
+
+**--ignorefile**
+
+Path to alternative .dockerignore file. Requires --contextdir be specified.
```
TomSweeneyRedHat, line 35:
```suggestion
Path to an alternative .dockerignore file. Requires --contextdir be specified.
```

docs/buildah-copy.md
```diff
@@ -25,6 +25,15 @@ BUILDAH\_HISTORY environment variable. `export BUILDAH_HISTORY=true`
 
 Sets the user and group ownership of the destination content.
 
+**--contextdir**
+
+Build context directory. Specifying a context directory, causes Buildah to
+chroot into a the context directory. This means copying files pointed at by symbolic links outside of the chroot will fail.
```
TomSweeneyRedHat, line 31: see bud-add comments

pkg/cli/common.go
```diff
@@ -185,6 +186,7 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
 	fs.StringVar(&flags.Creds, "creds", "", "use `[username[:password]]` for accessing the registry")
 	fs.BoolVarP(&flags.DisableCompression, "disable-compression", "D", true, "don't compress layers by default")
 	fs.BoolVar(&flags.DisableContentTrust, "disable-content-trust", false, "This is a Docker specific option and is a NOOP")
+	fs.StringVar(&flags.IgnoreFile, "ignorefile", "", "path to .dockerignore file")
```
TomSweeneyRedHat, line 189: Suggest:
```suggestion
	fs.StringVar(&flags.IgnoreFile, "ignorefile", "", "path to an alternate .dockerignore file")
```

cmd/buildah/addcopy.go
```diff
@@ -100,7 +108,19 @@ func addAndCopyCmd(c *cobra.Command, args []string, verb string, extractLocalArc
 	builder.ContentDigester.Restart()
 
 	options := buildah.AddAndCopyOptions{
-		Chown: iopts.chown,
+		Chown:      iopts.chown,
+		ContextDir: iopts.contextdir,
+	}
+	if iopts.ignoreFile != "" {
+		if iopts.contextdir == "" {
+			return errors.Errorf("--ignore options requires that you specify a context dir using --contextdir")
```
TomSweeneyRedHat, line 116:
```suggestion
			return errors.Errorf("--ignore options requires a context directory to be specified using --contextdir")
```

contrib/completions/bash/buildah
```diff
@@ -536,6 +537,8 @@ return 1
 
      local options_with_args="
      --chown
+     --ignorefile
+     --contextdir
```
TomSweeneyRedHat, line 541: could you flip the ordering of these two lines, please?

contrib/completions/bash/buildah
```diff
@@ -558,6 +561,8 @@ return 1
 
      local options_with_args="
      -chown
+     --ignorefile
+     --contextdir
```
TomSweeneyRedHat, line 565: ditto flip please

docs/buildah-add.md
```diff
@@ -27,6 +27,16 @@ BUILDAH\_HISTORY environment variable. `export BUILDAH_HISTORY=true`
 
 Sets the user and group ownership of the destination content.
 
+**--contextdir**
+
+Build context directory. Specifying a context directory causes Buildah to
+chroot into that context directory. This means copying files pointed at
+by symbolic links outside of the chroot will fail.
```
TomSweeneyRedHat, line 34: IDK, should we mention that this is used only in conjunction with --ignorefile?  Perhaps:
```
Used only in conjunction with the `--ignorefile` option.
```

docs/buildah-copy.md
```diff
@@ -25,6 +25,16 @@ BUILDAH\_HISTORY environment variable. `export BUILDAH_HISTORY=true`
 
 Sets the user and group ownership of the destination content.
 
+**--contextdir**
+
+Build context directory. Specifying a context directory causes Buildah to
+chroot into a the context directory. This means copying files pointed at
+by symbolic links outside of the chroot will fail.
```
TomSweeneyRedHat, line 32: if you do the conjunction bit above, do it here too.

docs/buildah-copy.md
```diff
@@ -49,9 +59,9 @@ buildah copy containerID 'passwd' 'certs.d' /etc
 
 ### `.dockerignore`
 
-If the file .dockerignore exists in the context directory, `buildah copy` reads
-its contents. Buildah uses the content to exclude files and directories from
-the context directory, when copying content into the image.
+If you specify the \fB\fC\-\-ignorefile\fR option to point to a .dockerignore
+path location. Buildah uses the content to exclude files and directories from
```
TomSweeneyRedHat, line 63:
```suggestion
path location, Buildah uses the content to exclude files and directories from
```

docs/buildah-add.md
nalind: If you're after the not-copying-anything-from-outside-of-the-context-dir behavior, then specifying a context directory, even without specifying a .dockerignore file, makes sense.

rhatdan:
@TomSweeneyRedHat @nalind @giuseppe @vrothberg @saschagrunert @containers/podman-maintainers PTAL

saschagrunert (approved):
LGTM

giuseppe (approved):
/lgtm

Issue #2404, opened 2020-06-13
Allow --build-arg to override FROM in Dockerfile


nnachefski:
I would like to be able to over override the FROM image in the Dockerfile from the command line

Example:

```
# buildah bud --build-arg=FROM=ubi8-custom --format=oci --tls-verify=false --layers -f Dockerfile.rhel8 -t image-registry.openshift-image-registry.svc:5000/custom-images/s2i-custom-core .
STEP 1: FROM registry.access.redhat.com/ubi8:latest
...
```

rhatdan:
Interested in opening a PR to make this happen?

nnachefski:
negative.  My Golang is.... nonexistent.  I saw that Kaniko does indeed have this functionality.  I'm going to give the kaniko ClusterTask a shot.

rhatdan:
No problem, we will take a look at it.

jamescassell:
You can already do this:
test.Dockerfile:
```
ARG FROM=registry.access.redhat.com/ubi8
FROM $FROM
ARG FROM
ENV FROM=$FROM
```

to swap out the FROM, build it like `podman build --build-arg=FROM=registry.fedoraproject.org/fedora -f from.Dockerfile`

rhatdan:
Also we have a preprocessor built into Buildah and Podman build.  Basically you can add 
# if 
# then 
# endif 

```
man buildah bud
...
      Dockerfiles  ending  with  a  ".in"  suffix  will  be preprocessed via CPP(1).  This can be useful to decompose Dockerfiles into several reusable parts that can be used via CPP's
       #include directive.  Notice, a Dockerfile.in file can still be used by other tools when manually preprocessing them via cpp -E.
```

rhatdan:
@TomSweeneyRedHat Did we ever blog on this feature?

rhatdan:
If not this entire issue would make for a nice blog.

TomSweeneyRedHat:
@rhatdan, I don't believe we did. I'll add it to the list.

nnachefski:
> You can already do this:
> test.Dockerfile:
> 
> ```
> ARG FROM=registry.access.redhat.com/ubi8
> FROM $FROM
> ARG FROM
> ENV FROM=$FROM
> ```
> 
> to swap out the FROM, build it like `podman build --build-arg=FROM=registry.fedoraproject.org/fedora -f from.Dockerfile`

[...]

PR #2823, opened 2020-11-25
Allow FROM to be overriden with from option

TomSweeneyRedHat:
These changes will allow the "FROM" statement in a Containerfile
to be overridden with a new `--from` option.  If I have this Dockerfile.fedora

```
FROM fedora
```

This command will instead build an alpine image:

```
# buildah bud --from=alpine -t tom -f ~/Dockerfile.fedora .
STEP 1: FROM alpine
Completed short name "alpine" with unqualified-search registries (origin: /etc/containers/registries.conf)
Copying blob 188c0c94c7c5 done
Copying config d6e46aa247 done
Storing signatures
STEP 2: COMMIT tom
--> d6e46aa2470
d6e46aa2470df1d32034c6707c8041158b652f38d2a9ae3d7ad7e7532d22ebe0
```

Addresses: #2404

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -61,6 +61,9 @@ instructions read from the Containerfiles in the same way that environment
 variables are, but which will not be added to environment variable list in the
 resulting image's configuration.
 
+Please refer to the [BUILD TIME VARIABLES](#build-time-variables) section for the
+list of variables that can be overridden within the Containerfile at run time.
+
 **--cache-from**
 
 Images to utilise as potential cache sources. Buildah does not currently support caching so this is a NOOP.
@@ -263,6 +266,11 @@ Recognized formats include *oci* (OCI image-spec v1.0, the default) and
 
 Note: You can also override the default format by setting the BUILDAH\_FORMAT
 environment variable.  `export BUILDAH_FORMAT=docker`
+
+**--from**
+
+Overrides the first `FROM` instruction within the Containerfile.  If there are multiple
+FROM instructions in a Containerfile, only the first is changed.
 
 **-h**, **--help**
 
@@ -666,6 +674,23 @@ will convert /foo into a `shared` mount point.  The propagation properties of th
 mount can be changed directly. For instance if `/` is the source mount for
 `/foo`, then use `mount --make-shared /` to convert `/` into a `shared` mount.
 
+## BUILD TIME VARIABLES
+
+The ENV instruction in a Containerfile can be used to define variable values.  When the image
+is built, the values will persist in the container image.  At times it is more convenient to
+change the values in the Containerfile via a command-line option rather than changing the
+values within the Containerfile itself.
+
+The following variables can be used in conjunction with the `--build-arg` option to override the
+corresponding values set in the Containerfile using the `ENV` instruction.
+
+  * HTTP_PROXY
+  * HTTPS_PROXY
+  * FTP_PROXY
+  * NO_PROXY
+
+Please refer to the [Using Build Time Variables](#using-build-time-variables) section of the Examples.
+
 ## EXAMPLE
 
 ### Build an image using local Containerfiles
@@ -726,6 +751,11 @@ buildah bud --dns-search=example.com --dns=223.5.5.5 --dns-option=use-vc .
   buildah bud -f dev/Containerfile https://10.10.10.1/docker/context.tar.gz
 
   Note: supported compression formats are 'xz', 'bzip2', 'gzip' and 'identity' (no compression).
+
+### Using Build Time Variables
+#### Replace the value set for the HTTP_PROXY environment variable within the Containerfile.
+
+buildah bud --build-arg=HTTP_PROXY="http://127.0.0.1:8321"
 
 ## ENVIRONMENT
 
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -114,6 +114,7 @@ type Executor struct {
 	logRusage                      bool
 	imageInfoLock                  sync.Mutex
 	imageInfoCache                 map[string]imageTypeAndHistoryAndDiffIDs
+	fromOverride                   string
 }
 
 type imageTypeAndHistoryAndDiffIDs struct {
@@ -229,6 +230,7 @@ func NewExecutor(store storage.Store, options BuildOptions, mainNode *parser.Nod
 		jobs:                           jobs,
 		logRusage:                      options.LogRusage,
 		imageInfoCache:                 make(map[string]imageTypeAndHistoryAndDiffIDs),
+		fromOverride:                   options.From,
 	}
 	if exec.err == nil {
 		exec.err = os.Stderr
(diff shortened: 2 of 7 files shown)


imagebuildah/executor.go
```diff
@@ -47,6 +47,7 @@ var builtinAllowedBuildArgs = map[string]bool{
 	"ftp_proxy":   true,
 	"NO_PROXY":    true,
 	"no_proxy":    true,
+	"FROM":        true,
```
rhatdan, line 50: Where are these documented?  Doesn't this new one need documentation in a man page.  Currently this seems like hidden magic?

rhatdan: https://docs.docker.com/engine/reference/commandline/build/

Could you craft a section in the buildah man page from the above to document these as well as your new FROM arg.

rhatdan:
whitespace

docs/buildah-bud.md
```diff
@@ -657,6 +660,25 @@ will convert /foo into a `shared` mount point.  The propagation properties of th
...
+change the values in the Containerfile via a command-line option rather than changing the 
+values within the Containerfile itself.
+
+The following variables can be used in conjunction with the `--build-arg` option to override the
+values set in the Containerfile.
+
+  * HTTP_PROXY
+  * HTTPS_PROXY
+  * FTP_PROXY
+  * NO_PROXY
```
rhatdan, line 676: What about FROM?

rhatdan (approved):
Add FROM documentation to man page then LGTM

docs/buildah-bud.md
TomSweeneyRedHat: next paragraph, slightly different handling as these are all specified like `ENV HTTP_PROXY="http://127.0.0.1:8321", but let me tease that out a bit more to be clear.

TomSweeneyRedHat:
Tried to make the man page a bit more clear on the Build Time Variables and finally chased down the white space issues.

rhatdan:
LGTM
@nalind PTAL

nalind:
This seems to conflict with the way imagebuilder does it:  as a dedicated CLI option which overrides the base image for only the first stage.  This lets it avoid adding a new predefined ARG.  An API caller isn't going to know that we're doing adding an ARG, and I'm worried that this could be hard to troubleshoot later on.

rhatdan:
@nalind Do you have an alternative idea?

nalind:
Instead of tunneling it through an ARG, make it a field in BuildOptions that's controlled by a CLI flag, and when we alter the parsed tree in `imagebuildah.Executor.Build()`, only do so if it's the first stage.

docs/buildah-bud.md
```diff
@@ -657,6 +660,27 @@ will convert /foo into a `shared` mount point.  The propagation properties of th
...
+
+The following variables can be used in conjunction with the `--build-arg` option to override the
+corresponding values set in the Containerfile using the `ENV` instruction.
+
+  * HTTP_PROXY
+  * HTTPS_PROXY
+  * FTP_PROXY
+  * NO_PROXY
+
+In addition, the FROM instruction can also be overridden.  Please note that if there are multiple
```
rhatdan, line 678: Does this make sense as the default.  I would have thought the more likely request would have been to only change the first?  What happens when they FROMs inside of the Containerfile refer back to previous steps?

TomSweeneyRedHat: I'll make it change just the first instance then rather than all.  Any thought on the CLI flag?  `-j-first-from` perhaps?

rhatdan: I would just change the first FROM for  now, and ignore the rest, no new flag.  I think that was what was asked for.  Wait until someone else asks for changing all of the FROMs.  At that point you probably need to rewrite your containerfile.

nalind: The option for imagebuilder is `-from`, and it's specified using `from:` [in OpenShift](https://docs.openshift.com/container-platform/4.6/builds/build-strategies.html#builds-strategy-docker-from-image_build-strategies), so `--from` seems fine.

rhatdan:
@TomSweeneyRedHat Needs rebase.

docs/buildah-bud.md
TomSweeneyRedHat: I've changed it to a `--from`  option per @nalind's last comment.

TomSweeneyRedHat:
@rhatdan @nalind reworked and repushed.  Ready for the next round of reviews.  I've changed from using a Build Time variable passed in with a build-arg to creating a new `--from` option.

rhatdan:
LGTM if you can get tests to pass.

imagebuildah/executor.go
```diff
@@ -245,6 +247,10 @@ func NewExecutor(store storage.Store, options BuildOptions, mainNode *parser.Nod
 			fmt.Fprintf(exec.out, prefix+format+suffix, args...)
 		}
 	}
+
+	// If we have a value from the '--from' option, remember it.
+	exec.fromOverride = options.From
```
nalind, line 252: Why set it here instead of on line 233?

TomSweeneyRedHat: Because at the time my noggin was thinking I needed to set it to "" by default, but that I already took care of in the declaration in pkg/cli/common.go.  Good thought, I'll change it.

rhatdan:
--from needs to be added to completions.

TomSweeneyRedHat:
@rhatdan, `--from` added to the completions and now this has all happy green test buttons.

rhatdan:
Lets get this merged.
@saschagrunert @giuseppe @vrothberg @nalind @ashley-cui @umohnani8 PTAL

ashley-cui (commented):
LGTM

kitty-catt:
I just ran the --from flag on buildah provided by registry.redhat.io/rhel8/buildah. The --from option is not there yet. 

Is there any idea when this option becomes available in registry.redhat.io/rhel8/buildah? 

Do I have an alternative?

I have attached the output of the buildah version and help commands:
[buildah-output.txt](https://github.com/containers/buildah/files/6378077/buildah-output.txt)

nalind:
The feature landed in the v1.19 release.  The `quay.io/buildah/stable` image, which is built similarly, is currently at v1.19.8.  I expect an update to `registry.redhat.io/rhel8/buildah` for RHEL 8.4 will update it to something close to that version, depending on the timing of it.

kitty-catt:
Thank you nalind. It works when I use quay.io/buildah/stable (in a tekton task on openshift)! I intend to switch back to redhat once they have put it in.

TomSweeneyRedHat:
@fatherlinux pinging you in case there's anything that needs to happen to update the Buildah image on the RHEL registries.

PR #2824, opened 2020-11-25
Not use local image create/add manifest

QiWang19:
Avoid first checking the image from local storage for `manifest create` and `manifest add`
since the local image does not include other entries of the list from the registry.
`--all` flag of `manifest create` and `manifest add` can not add all of the lists as expected.
So first attempt to get manifest from the registry and if not successful, fall back to use local images.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/cmd/buildah/manifest.go
+++ b/cmd/buildah/manifest.go
@@ -237,10 +237,13 @@ func manifestCreateCmd(c *cobra.Command, args []string, opts manifestCreateOpts)
 	}
 
 	for _, imageSpec := range imageSpecs {
-		ref, _, err := util.FindImage(store, "", systemContext, imageSpec)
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
@QiWang19 does this address an issue or BZ?  If so, please add a pointer to the description.

QiWang19:
> @QiWang19 does this address an issue or BZ? If so, please add a pointer to the description.

This does not point to an issue. Just found this may cause problems when I debug other manifest issues.

TomSweeneyRedHat:
Changes LGTM, but definitely want a head nod from @nalind.

PR #2844, opened 2020-12-07
Move away from using docker.io

edsantiago:
November 2020, docker.io started restricting unauthenticated
image pulls. Try to work around this by using a custom
registries.conf file.

For the most part this means copying images from docker.io
to quay.io, via:

   $ skopeo copy --all docker://docker.io/library/img:tag \
                       docker://quay.io/libpod/img:tag

...for the following values of 'img:tag':

   busybox:musl
   centos:7  centos:8  centos:latest
   composer:latest
   debian:latest       ubuntu:latest
   docker:latest
   php:7.2

For each of those, it was necessary to go into the quay.io
GUI, click the image name, click the settings (gear) icon
at bottom left, click 'Make public', and confirm.

This process did not work in four instances, which required
special-casing:

   commit.bats : openshift/hello-openshift did not match the
                 mirroring rules; I switched to alpine instead.
                 Nalin confirmed on IRC that there was no magic
                 reason for requiring hello-openshift.

   pull.bats   : change a SHA. AFAICT there was nothing magic
                 about the SHA being used, it was just a
                 convenient one for purposes of testing
                 pull-by-sha. I simply switched to the SHA
                 of an image present on quay.

   registries.bats : was assuming that fedora-minimal shortname
                     would be pulled from fedora registry.
                     Unfortunately, we have a copy on quay
                     (for podman tests), so that's what we
                     pull by shortname, and it does not match
                     the SHA of the fedoraproject.org one.
                     Solution: pull by tag (fedora-minimal:32)
                     and hope that nobody ever mirrors that one
                     on quay.

   run.bats    : another pull-by-SHA, but this time I changed
                 the SHA to a named tag, and skopeo copy'd
                 that image from docker.io to the given name
                 on quay. This time there _is_ something
                 magic about that particular SHA (it's an
                 image with a specific volume quirk) but
                 there's no actual reason to reference it
                 by SHA - we simply did so because we have
                 no control over tag names on docker.io.
                 Since we control tag names on quay.io,
                 it's easy and more maintainable to give
                 this image a descriptive tag.

### DIFF
--- a/tests/registries.conf
+++ b/tests/registries.conf
@@ -1,25 +1,17 @@
-# This is a system-wide configuration file used to
-# keep track of registries for various container backends.
-# It adheres to TOML format and does not support recursive
-# lists of registries.
+# Note that changing the order here may break tests.
+unqualified-search-registries = ['docker.io', 'quay.io', 'registry.fedoraproject.org']
 
-# The default location for this configuration file is /etc/containers/registries.conf.
+[[registry]]
+# In Nov. 2020, Docker rate-limits image pulling.  To avoid hitting these
+# limits while testing, always use the google mirror for qualified and
+# unqualified `docker.io` images.
+# Ref: https://cloud.google.com/container-registry/docs/pulling-cached-images
+prefix="docker.io"
+location="mirror.gcr.io"
 
-# The only valid categories are: 'registries.search', 'registries.insecure', 
-# and 'registries.block'.
-
-[registries.search]
-registries = ['docker.io', 'quay.io', 'registry.fedoraproject.org']
-
-# If you need to access insecure registries, add the registry's fully-qualified name.
-# An insecure registry is one that does not have a valid SSL certificate or only does HTTP.
-[registries.insecure]
-registries = []
-
-
-# If you need to block pull access from a registry, uncomment the section below
-# and add the registries fully-qualified name.
-#
-# Docker only
-[registries.block]
-registries = []
+# 2020-10-27 a number of images are not present in gcr.io, and podman
+# barfs spectacularly when trying to fetch them. We've hand-copied
+# those to quay, using skopeo copy --all ...
+[[registry]]
+prefix="docker.io/library"
+location="quay.io/libpod"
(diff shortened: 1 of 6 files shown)


TomSweeneyRedHat:
Neat use of registries.conf.  This looks likes a good "how to" article for "enable sysadmin"....

TomSweeneyRedHat:
LGTM
assuming happy tests

rhatdan:
LGTM
I would love to have this blog,

tests/registries.conf
```diff
@@ -1,25 +1,17 @@
...
-#
-# Docker only
-[registries.block]
-registries = []
+# 2020-10-27 a number of images are not present in gcr.io, and podman
+# barfs spectacularly when trying to fetch them. We've hand-copied
+# those to quay, using skopeo copy --all ...
+[[registry]]
+prefix="docker.io/library"
+location="quay.io/libpod"
```
vrothberg, line 17: Some images are not mirrored to `quay.io/libpod` (e.g., Debian). I think we could rewrite this to:
```toml
[[registry]]
prefix="docker.io/library"
location="mirror.gcr.io/library"
 
[[registry.mirror]]
location="quay.io/libpod"
```

With this config, we'd first attempt to pull from `quay.io/libpod` and then use the GCR mirror. This should catch cases like Debian.

edsantiago:
Whew! CI is now green. The re-pushes were expected and necessary (well, most of them. Two were unnecessary CI churn due to my carelessness).

I have updated the commit message and PR description to reflect everything I did; please review that carefully. (Commit message is much much longer than the actual test diffs). (Oh, and my PR description formatting is horrible, so I really do recommend reading the git commit message instead).

@nalind I'd appreciate your attention on each of the four image changes I made in individual tests, to make sure there aren't any subtle image properties that will be lost by moving between repos.

@vrothberg I didn't understand your comment from this morning, and I choose to leave `registries.conf` exactly as it was on my first submission yesterday. It is exactly the same as the one used in podman CI, and (more importantly) it works this way. 

Thanks all for your help and feedback.

vrothberg (commented):
LGTM, great work @edsantiago

rhatdan:
/lgtm
/hold cancel
Now I await the blog.

PR #2845, opened 2020-12-08
copier: special case TypeReg{A} directory

vrothberg:
Some archives may falsely declare a directory as a file header.
In this case, do not bother to create a file (or remove it).
While tar would reject these files, Docker happily eats them,
so we need to be compatible.

See github.com/containers/podman/issues/8600.

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -1311,6 +1311,18 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
 	createFile := func(path string, tr *tar.Reader) (int64, error) {
 		f, err := os.OpenFile(path, os.O_CREATE|os.O_WRONLY|os.O_TRUNC|os.O_EXCL, 0600)
 		if err != nil && os.IsExist(err) {
+			if fInfo, err := os.Stat(path); err == nil {
+				// Some archives may falsely declare a directory as a file header.
+				// In this case, do not bother to create a file (or remove it).
+				// While tar would reject these files, Docker happily eats them,
+				// so we need to be compatible.
+				//
+				// See github.com/containers/podman/issues/8600.
+				if fInfo.IsDir() {
+					return 0, nil
+				}
+			}
+
 			if err = os.Remove(path); err != nil {
 				return 0, errors.Wrapf(err, "copier: put: error removing file to be overwritten %q", path)
 			}


vrothberg:
@nalind I need your input on this.

rhatdan:
Code LGTM
But agree we need @nalind to chime in.

BarDweller:
Fwiw, I added the dirs to the tar in https://github.com/containers/podman/issues/8600 explicitly, because without that, I noticed that docker was creating implicit directories with default permissions/ownership, by adding the dirs the way I did, it caused docker to create the dirs with the correct owner/permissions. (eg, the tar contains /app/content/file and /app and /app/content were being created with default uid/gid rather than the ones I needed)..

nalind:
This patch looks like it will ignore entries that claim to be for files when there's already a directory in the destination, instead of removing the directory (if it's empty) and creating the file with the contents and permissions specified in the tarball.  If the tarstream does not include subsequent items that go under that directory later, this sounds like something we'd consider to be an error.

@BarDweller Do you have a sample tarstream we can look at and replay with dockerd in conformance tests, to make sure we're getting the behavior exactly right?

vrothberg:
> This patch looks like it will ignore entries that claim to be for files when there's already a directory in the destination, instead of removing the directory (if it's empty) and creating the file with the contents and permissions specified in the tarball. If the tarstream does not include subsequent items that go under that directory later, this sounds like something we'd consider to be an error.
> 
> @BarDweller Do you have a sample tarstream we can look at and replay with dockerd in conformance tests, to make sure we're getting the behavior exactly right?

@nalind I can send you one

BarDweller:
I went digging in the commons compress library I'm using to build the tar stream. 

I'm adding my directories to the tar archive with paths ending in "/" this should trigger this line of code.. 
https://github.com/apache/commons-compress/blob/master/src/main/java/org/apache/commons/compress/archivers/tar/TarArchiveEntry.java#L309
which will cause 'isDir' to be set to true, which in turn causes `this.linkFlag` to be set to `LF_DIR`.  It looks like the linkFlag is written immediately after the checksum, and just before the linkName in the tar entry header. (around lines 1266 same file). 

So you shouldn't be seeing it as "Some archives may falsely declare a directory as a file header.", but as an entry with header that says it's a directory.. From the file affected by this PR, I spot you are reading that flag in the block around 1372 that parses each tar entry, but that's only driven when the hdr.Typeflag is tar.TypeReg, or tar.TypeRegA, I can't easily figure out which Go tar library your using, but I did confirm that LF_DIR is a byte with the value `5` .. which looks right according to the spec. 

Which entries are you seeing in the test tar that are directories-pretending-to-be-regular-files ?

vrothberg:
Thanks for the analysis @BarDweller.  @nalind WDYT?

nalind:
We implement logic to transfer the tarstreams over pipes either to or from chrooted processes, but we're using the Go standard library's `archive/tar` package [(reference)](https://tip.golang.org/pkg/archive/tar/) to do the actual reading and writing of headers and contents.  The library does some work to allow us to only have to manage a single header structure type, abstracting out the specifics of header format used for us.  That logic interprets an entry marked with the digit '5' as indicating a directory, so I wouldn't expect us to head down the code path that we're patching here.

An "unexpected EOF" error reading a tarball usually points either to a header that specified a content size that was longer than the rest of the stream (for anything other than a regular size, I'd expect a size of 0), or a missing trailer (1024 zero bytes where a sequential reader would be looking to find the header for the next entry).

vrothberg:
@nalind and I had quick look together.  The problem is the empty header name.  Docker errors out the same way.

@nalind wants to improve the error messaging, and massage the code a bit.

Closing this PR as it's not the right approach.

PR #2849, opened 2020-12-11
copier: Put: skip entries with zero-length names, handle directory <=> not-directory better

nalind:
#### What type of PR is this?

/kind bug

When extracting archives, if we encounter an entry with a zero-length name, skip it, instead of attempting to Join the entry's name with the name of the parent directory where it should be placed (which just yields the parent directory's name again), and attempting to create that directory again, possibly as a file.

Improve handling of cases where extracting an archive requires us to replace a directory with something that is not a directory, or vice-versa:
* when replacing a directory with something that isn't a directory, remove the directory even if it has contents
* don't fail when replacing something that isn't a directory with a directory

Add a flag for controlling overwriting-directories-with-non-directories behavior in PutOptions, and fields for controlling name mappings to both GetOptions and PutOptions.

Modified the copier package so that it only attempts to chroot() when started as UID 0, and uses fallback logic otherwise.  This is less safe than using chroot(), so consumers that can set up and re-exec themselves inside of a user namespace are still better off doing so.

#### How to verify it

More unit tests!
More conformance tests!

Should fix github.com/containers/podman/issues/8600.

Replaces #2845.

```
The ADD instruction should now correctly handle cases where an archive being ADDed replaces a directory with a non-directory, or a non-directory with a directory.
```


TomSweeneyRedHat:
@nalind tests weren't happy, they looked like flakes trying to pull busybox from quay.io.  I've restarted.

vrothberg (commented):
Code LGTM, nice work!

Could we make this configurable?  The remote endpoints have a `noOverwriteDirNonDir` parameter which we do not yet support.

> If 1, true, or True then it will be an error if unpacking the given content would cause an existing directory to be replaced with a non-directory and vice versa.

I have one more request that AFAICS will be the last missing piece:  Can we add a field to the options that allows for renaming entries?  The `archive` package supports that via `map[string]string`.

`podman cp file container:/anotherfile` requires that.  It worked locally by using the `archive` package but that doesn't work remotely any more.

rhatdan:
@cevich the Ubuntu tests are failing everywhere in Buildah
 @cirrus-cicirrus-ci/required/testing gce_instance:ubuntu-19-podman-6530021898584064

nalind:
Looks like Ubuntu's 19.10 release is far enough past its end-of-life that we don't gain anything from testing on it any more. The current LTS version is 20.04, and the latest version is 20.10.

rhatdan:
I opened a PR to turn off that tests, until we get updated VMs.

rhatdan:
@nalind rebase, will eliminate ubuntu failed tests.

nalind:
Rebased.

vrothberg (commented):
LGTM

I cannot highlight enough how much that helps for the copy work in Podman.  Thanks, @nalind !

cevich:
Note: Updated VMs are here https://github.com/containers/buildah/pull/2631 but I need help getting all the tests passing.

Issue #1945, opened 2019-10-25
vfs driver does not support overlay.mountopt options


cevich:
**Description**

Confusing error message unrelated to desired storage driver.  This seems like the buildah-flavor of containers/libpod#3240 though #1783 is also related.

**Steps to reproduce the issue:**
1. `$ sudo podman run -it --rm -e STORAGE_DRIVER=vfs fedora:latest bash`
2. `# dnf install -y buildah &> /dev/null; echo $?`
3. `# buildah from scratch`

**Describe the results you received:**

`vfs driver does not support overlay.mountopt options`

**Describe the results you expected:**

`working-container`

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.11.3-2.fc30.x86_64
```

**Output of `buildah version`:**

```
Version:         1.11.3
Go Version:      go1.12.9
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.1-dev
CNI Spec:        0.4.0
libcni Version:  
Git Commit:      
Built:           Thu Jan  1 00:00:00 1970
OS/Arch:         linux/amd64
```

**Output of `cat /etc/*release`:**

```
Fedora release 30 (Thirty)
NAME=Fedora
VERSION="30 (Container Image)"
ID=fedora
VERSION_ID=30
VERSION_CODENAME=""
PLATFORM_ID="platform:f30"
PRETTY_NAME="Fedora 30 (Container Image)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:30"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f30/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=30
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=30
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Container Image"
VARIANT_ID=container
Fedora release 30 (Thirty)
Fedora release 30 (Thirty)
```

**Output of `uname -a`:**

[...]

PR #2860, opened 2020-12-17
Remove a "storage mountopt" workaround

edsantiago:
PR #2115 (Kill Travis and Enable Bors, merged 2020-02-07) added
code that completely clobbered /etc/containers/storage.conf,
as a workaround for issue #1945. That closed 2020-04-29, and
the workaround is now a liability that has wasted developer
time. Remove it.

### DIFF
--- a/contrib/cirrus/lib.sh
+++ b/contrib/cirrus/lib.sh
@@ -178,22 +178,6 @@ showrun() {
     "$@"
 }
 
-# workaround issue 1945 (remove when resolved)
-remove_storage_mountopt() {
-    local FILEPATH=/etc/containers/storage.conf
-    warn "remove_storage_mountopt() is overwriting $FILEPATH"
-    # This file normally comes from containers-common package
-    cat <<EOF> $FILEPATH
-[storage]
-driver = "overlay"
-runroot = "/var/run/containers/storage"
-graphroot = "/var/lib/containers/storage"
-[storage.options]
-mountopt = ""
-EOF
-    cat $FILEPATH
-}
-
 # Remove all files provided by the distro version of buildah.
 # All VM cache-images used for testing include the distro buildah because it
 # simplifies installing necessary dependencies which can change over time.
--- a/contrib/cirrus/setup.sh
+++ b/contrib/cirrus/setup.sh
@@ -67,10 +67,14 @@ echo -e "[registries.search]\nregistries = ['docker.io', 'registry.fedoraproject
 
 show_env_vars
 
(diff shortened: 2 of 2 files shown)


edsantiago:
```
 time="2020-12-17T15:52:23Z" level=error msg="The storage 'driver' option must be set in /etc/containers/storage.conf, guarantee proper operation."
```
well, that answers that.

rhatdan:
We could go back to the sed line or discover why the /etc/containers/storage.conf in the VMs does not have a driver defined.

edsantiago:
That has me curious (why the existing containers.conf is broken) but it's going to need

edsantiago:
...going to need some sleuthwork. I checked the obvious place (automation_images), and couldn't find anything. I don't remember how to do the get-ci-vm thing. So let's try some debugging printfs.

/hold

edsantiago:
Interesting: fedora 31/32 pass; only ubuntu-20 fails, and the [debug log](https://cirrus-ci.com/task/5154621809229824?command=setup#L151) confirms that:
```
# Default Storage Driver
driver = ""
```
Does anyone know the history behind this? Was there ever a bad containers-common that made it out, maybe only to ubuntu? Will a VM bump fix this? It seems suspicious that `driver = ""` is the exact symptom that @rhatdan was trying to work around in #2842. 

Anyhow, the good news is that fedora passes, so we might be able to proceed with this PR once we figure out the ubuntu problem.

rhatdan:
The upstream containers/storage storage.conf has driver="", since upstream does not want to require a default.  In Fedora (As well as RHEL) we set the distro defaults including driver="overlay" and mountopt="nodev,metacopy=on".

I would bet the default in Ubuntu was just to grab the upstream default and that is causing the problems.

edsantiago:
That would imply that all end users will see this warning, which would make it a packaging bug, no? Maybe something to track next year. Thanks for the context, that makes a lot of sense now.

rhatdan:
Yes I think we need to get @lsm5 and friends to make sure this file has driver="overlay" defined in the storage.conf on Debian/Ubuntu platforms.

lsm5:
> Yes I think we need to get @lsm5 and friends to make sure this file has driver="overlay" defined in the storage.conf on Debian/Ubuntu platforms.

doing this now ...

lsm5:
@cevich should be ready now https://build.opensuse.org/package/show/devel:kubic:libcontainers:stable/skopeo

edsantiago:
This is resolved by #2631

PR #2868, opened 2020-12-22
Use --arch and --os options to select architecture and os

rhatdan:
Remove --override-os and --override-arch flags.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/docs/buildah-pull.md
+++ b/docs/buildah-pull.md
@@ -46,6 +46,10 @@ The image ID of the image that was pulled.  On error 1 is returned.
 
 All tagged images in the repository will be pulled.
 
+**--arch**="ARCH"
+
+Set the ARCH of the image to be pulled to the provided value instead of using the architecture of the host. (Examples: aarch64, arm, i686, ppc64le, s390x, x86_64)
+
 **--authfile** *path*
 
 Path of the authentication file. Default is ${XDG\_RUNTIME\_DIR}/containers/auth.json, which is set using `buildah login`.
@@ -70,6 +74,10 @@ The [key[:passphrase]] to be used for decryption of images. Key can point to key
 
 If an image needs to be pulled from the registry, suppress progress output.
 
+**--os**="OS"
+
+Set the OS of the image to be pulled instead of using the current operating system of the host.
+
 **--policy**=**always**|**missing**|**never**
 
 Pull image policy. The default is **missing**.
@@ -86,6 +94,9 @@ Don't copy signatures when pulling images.
 
 Require HTTPS and verify certificates when talking to container registries (defaults to true).  TLS verification cannot be used when talking to an insecure registry.
 
+**--variant**=""
+
+Set the architecture variant of the image to be pulled.
 
 ## EXAMPLE
 
@@ -106,6 +117,10 @@ buildah pull --tls-verify=false myregistry/myrepository/imagename:imagetag
 buildah pull --creds=myusername:mypassword --cert-dir ~/auth myregistry/myrepository/imagename:imagetag
 
 buildah pull --authfile=/tmp/auths/myauths.json myregistry/myrepository/imagename:imagetag
+
+buildah pull --arch=aarch64 myregistry/myrepository/imagename:imagetag
+
+buildah pull --arch=arm --variant=v7 myregistry/myrepository/imagename:imagetag
 
 ## ENVIRONMENT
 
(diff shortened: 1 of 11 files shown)


docs/buildah-pull.md
```diff
@@ -46,6 +46,14 @@ The image ID of the image that was pulled.  On error 1 is returned.
 
 All tagged images in the repository will be pulled.
 
+**--arch**="ARCH"
+
+Set the ARCH of the image to the provided value instead of using the architecture of the host.
+
+**--arch**="ARCH"
```
jamescassell, line 53: You've got this entry twice.

docs/buildah-pull.md
```diff
@@ -70,6 +78,10 @@ The [key[:passphrase]] to be used for decryption of images. Key can point to key
 
 If an image needs to be pulled from the registry, suppress progress output.
 
+**--os**="OS"
+
+Set the OS of the image to be pulled to the provided value instead of using the current operating system of the host.
```
jamescassell, line 83: Examples?

mtrmac (commented):
- This breaks any pre-existing users of the `--override-`* flags. If that is intentional, nothing in this PR records why, and that rationale will be impossible to reconstruct.
- This was linked from https://github.com/containers/buildah/issues/2873 as “helping fix this”, I can’t see that it does. #2873 (supposedly) fails because defaulting OS/arch to `runtime.GO*` values turns off autodetection (including variant autodetection), and this PR AFAICS changes nothing about that.

rhatdan:
@mtrmac I setup aliases for --override-arch -> --arch and --override-os -> --os

So existing users should not be broken and these options were never documented.

rhatdan:
This PR allows the specification of the variant, which was one of the requested changes in that issue, I believe.

mtrmac:
> @mtrmac I setup aliases for --override-arch -> --arch and --override-os -> --os

Ah, my mistake. Thanks.

---

> This PR allows the specification of the variant, which was one of the requested changes in that issue, I believe.

Ah, OK. The PR description doesn’t say, and I have missed that.

Letting the user manually specify a variant provides a workaround; the way c/image works, removing the defaulting to `runtime.GO*` should make that workaround unnecessary for common uses (OTOH I haven’t checked whether it would break anything else).

rhatdan:
@mtrmac Added a second patch to stop defaulting the OS and Arch if the user did not specify it.  Is this what you meant?

TomSweeneyRedHat:
@rhatdan tests aren't hip and you need a rebase

pkg/parse/parse.go
```diff
@@ -649,23 +649,27 @@ func getAuthFile(authfile string) string {
 // PlatformFromOptions parses the operating system (os) and architecture (arch)
 // from the provided command line options.
 func PlatformFromOptions(c *cobra.Command) (os, arch string, err error) {
-	os = runtime.GOOS
-	arch = runtime.GOARCH
 
-	if selectedOS, err := c.Flags().GetString("os"); err == nil && selectedOS != runtime.GOOS {
-		os = selectedOS
+	if c.Flag("os").Changed {
```
mtrmac, line 653: - This does not change `SystemContextFromOptions`
- From a very quick skim, the results of _this_ function are used to set the OS/arch inside the built image, and not passed to c/image at all. That probably should have _some_ default.
- It seems easier to follow to me to not have the `runtime.GO*` defaults for flag values (in `cmd/buildah/`) specified, if they are going to be ignored here, and if this function hard-codes a default anyway.

Disclaimer: This is based on a very cursory skim, I don’t understand the code base and data flow at all well.

rhatdan: The advantage of the defaults is the buildah --help information shows them

$ ./bin/buildah from --help | grep -- --os
      --os OS                         prefer OS instead of the running OS when pulling images (default "linux")
$ ./bin/buildah from --help | grep -- --arch
      --arch string                   set the ARCH of the image to the provided value instead of the architecture of the host (default "amd64")

The SystemContextFromOptions is a good point, and I added the .Changed check there as well.

rhatdan:
@mtrmac @TomSweeneyRedHat Tests are passing and I think I addressed @mtrmac concerns.

We need to get this merged so I can continue on the manifests PR.

pkg/parse/parse.go
mtrmac: Well, but it’s _not_ the default if no `--os` and `--os=$default` behave differently.

It is also possible to embed that value into the flag description instead.

---

The change of the default return value of `PlatformFromOptions` still looks wrong to me, but I’ll leave that to Buildah experts.

TomSweeneyRedHat:
LGTM

rhatdan:

rhatdan:
@vrothberg PTAL

ashley-cui:
LGTM

Issue #2876, opened 2020-12-28
buildah bud broken?


mailing-lists:
Hey,
as I was trying to learn about building my own containers, I stumbled upon this issue,while trying to add a chrootdir or the ubuntu rootfs tarball (mind you alpine rootfs works 100%):
```
$ buildah bud -t ubuntu
STEP 1: FROM scratch
STEP 2: ADD focal-server-cloudimg-amd64-root.tar.xz /
error building at STEP "ADD focal-server-cloudimg-amd64-root.tar.xz /": error adding sources [/home/user/Containers/ubuntu/focal-server-cloudimg-amd64-root.tar.xz]: 3 errors occurred:
[...]

PR #2884, opened 2021-01-04
copier.PutOptions: add an "IgnoreDevices" flag

nalind:
#### What type of PR is this?

/kind bug

Prior to 1.16, when ADDing contents to a working container, if we were being run by an unprivileged user using a user namespace, content that was a device node would be ignored.

Add a flag in copier.PutOptions that tells copier.Put() to ignore entries that are either a device, or a hard link to a device.
Make buildah.Add() set the IgnoreDevices flag in PutOptions when libcontainer says we're running in a user namespace.

Together, these two changes should restore the earlier behavior.

#### How to verify it

New unit tests!  Also, manually running "buildah add" and specifying as the source either device nodes, or a tarball containing device nodes, should quietly ignore the device nodes and keep on chugging.

Fixes https://github.com/containers/buildah/issues/2876.

```
Attempting to ADD device nodes to a working container when running as an unprivileged user in rootless mode will now quietly ignore the device nodes and appear to succeed again, matching the behavior of versions before 1.16, instead of triggering an error.
```

### DIFF
--- a/add.go
+++ b/add.go
@@ -20,6 +20,7 @@ import (
 	"github.com/containers/storage/pkg/fileutils"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/hashicorp/go-multierror"
+	rsystem "github.com/opencontainers/runc/libcontainer/system"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
LGTM

PR #2847, opened 2020-12-11
Allow users to specify stdin into containers

rhatdan:
Some commands within a Containerfile, might need input from users.
For example confirmation commands from Apt.

Adding a --stdin flag will allows users to interact with containers
while running inside of buildah bud.

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### Does this PR introduce a user-facing change?

```release-note

```

PR #2889, opened 2021-01-06
[release-1.18] imagebuildah: disable pseudo-terminals for RUN, allow users to specify stdin into containers

nalind:
#### What type of PR is this?

/kind bug

Default to handling RUN instructions with no pseudo terminal, which matches what I see with docker build 19.03.  Interactive `buildah run` will still have the same default behavior.

Some commands within a Containerfile might need input from users, for example confirmation commands from `apt`, so add a `--stdin` flag to allows users to interact with commands which are run during a build.

#### How to verify it

Tests should continue to pass, and the issue described in https://github.com/containers/podman/issues/8342 should no longer occur.

None

Cherry-picked from #2803 and #2847.

```
RUN instructions in builds will no longer be run attached to a pseudo-terminal unless the `--stdin` flag is added.
```

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -197,6 +198,10 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budOptions) error {
 		return errors.Wrapf(err, "error evaluating symlinks in build context path")
 	}
 
+	var stdin io.Reader
+	if iopts.Stdin {
+		stdin = os.Stdin
+	}
 	var stdout, stderr, reporter *os.File
 	stdout = os.Stdout
 	stderr = os.Stderr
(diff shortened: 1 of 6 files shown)


rhatdan:
I belive the CI system is broken for older releases going to force merge.

TomSweeneyRedHat:
@cevich are you aware of CI breakage in older Buildah Releases?

cevich:
@TomSweeneyRedHat I'm assuming you mean the task failure in this PR?  Looking...it would appear this is caused by trying to install packages at runtime for an old Ubuntu 19 VM.  This will always fail (going forward) since Ubuntu 19 is EOL.  I'm not enough of an Ubuntu geek to know if (maybe) there's a way to point at "archived" repositories somehow(?).  Otherwise there simply isn't a good solution for this (and why I avoid-like-the-plague, doing runtime installs).  We could disable the Ubuntu 19 tests (in this branch) as a "fix" or I'm open to alternative ideas if anyone has any.  Given enough time, the same thing will happen to all Ubuntu releases :confused:

TomSweeneyRedHat:
@cevich yeah, the task failure here is what I was yacking about.  As we do an Ubuntu-20 in this branch, and 19 is EOL, I'd be fine with just removing the Ubuntu 19 test in this branch.  @nalind and @rhatdan any objections?

rhatdan:
Agreed, remove it.

PR #2022, opened 2019-12-13
Set the TMPDIR for pulling/pushing image to $TMPDIR

rhatdan:
Or set it to /var/tmp if the user did not specify.

Currently certain large workloads can not be handled because users are running
out of space on pulls/ and pushes.  Containers/image stores data temporarily in
the file system.  This allows the user to overide the location of the temporary
storage.

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

PR #2893, opened 2021-01-06
Fix handling of TMPDIR environment variable

rhatdan:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1900822

Currently we are not passing the TMPDIR to the destination context
so when copying to containers storage we always fall back to /var/tmp.

This change sets the destination to match the source, so we can better
handle the storage.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/pull.go
+++ b/pull.go
@@ -292,7 +292,7 @@ func pullImage(ctx context.Context, store storage.Store, srcRef types.ImageRefer
 	}()
 
 	logrus.Debugf("copying %q to %q", transports.ImageName(srcRef), destName)
-	if _, err := retryCopyImage(ctx, policyContext, maybeCachedDestRef, srcRef, srcRef, getCopyOptions(store, options.ReportWriter, sc, nil, "", options.RemoveSignatures, "", nil, nil, options.OciDecryptConfig), options.MaxRetries, options.RetryDelay); err != nil {
+	if _, err := retryCopyImage(ctx, policyContext, maybeCachedDestRef, srcRef, srcRef, getCopyOptions(store, options.ReportWriter, sc, sc, "", options.RemoveSignatures, "", nil, nil, options.OciDecryptConfig), options.MaxRetries, options.RetryDelay); err != nil {
 		logrus.Debugf("error copying src image [%q] to dest image [%q] err: %v", transports.ImageName(srcRef), destName, err)
 		return nil, err
 	}
--- a/tests/pull.bats
+++ b/tests/pull.bats
@@ -329,3 +329,16 @@ load helpers
 
   run_buildah rmi alpine
 }
+
+@test "pull image with TMPDIR set" {
+  testdir=${TESTDIR}/buildah-test
+  mkdir -p $testdir
+  mount -t tmpfs -o size=1M tmpfs $testdir
+
+  TMPDIR=$testdir run_buildah 125 pull --policy always --signature-policy ${TESTSDIR}/policy.json quay.io/libpod/alpine_nginx:latest
+  expect_output --substring "no space left on device"
+
+  run_buildah pull --policy always --signature-policy ${TESTSDIR}/policy.json quay.io/libpod/alpine_nginx:latest
+  umount $testdir
+  rm -rf $testdir
+}
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -644,6 +644,7 @@ func SystemContextFromOptions(c *cobra.Command) (*types.SystemContext, error) {
 		}
 	}
 
+	ctx.BigFilesTemporaryDir = GetTempDir()
 	return ctx, nil
 }
 


rhatdan:
@mtrmac @QiWang19 PTAL

rhatdan:
@edsantiago PTAL at tests.  Best way I can think of testing this without screwing up /var/tmp.

tests/pull.bats
```diff
@@ -312,3 +312,18 @@ load helpers
...
   run_buildah rmi alpine
 }
+
+@test "pull image with TMPDIR set" {
+  testdir=${TESTDIR}/buildah-test
+  mkdir -p $testdir
+  mount -t tmpfs -o size=1M tmpfs $testdir
+  export TMPDIR=$testdir
+
+  run_buildah 125 pull --signature-policy ${TESTSDIR}/policy.json alpine
```
edsantiago, line 322: May I recommend collapsing those two into one?
```bash
TMPDIR=$testdir run_buildah ....
```
As it is now, if the `expect_output` fails, `$TMPDIR` will remain set for all subsequent tests, and very likely cause cascading failures.

Even with this, a failure will leave `$testdir` mounted. That is much harder to catch, so I'm going to sweep it under the carpet.

edsantiago:
New test LGTM. I still can't think of a sane way to check and clean up the mount on teardown. Is there any objection to just ignoring the problem?

common.go
```diff
@@ -34,6 +34,10 @@ func getCopyOptions(store storage.Store, reportWriter io.Writer, sourceSystemCon
 	if destinationSystemContext != nil {
 		*destinationCtx = *destinationSystemContext
 	}
+	if destinationCtx != nil {
+		destinationCtx.BigFilesTemporaryDir = sourceCtx.BigFilesTemporaryDir
```
mtrmac, line 38: This does not conceptually fit. Why is this the business of `getCopyOptions` at all? How would the caller of `getCopyOptions` expect this to happen?

In fact both calls in `commit.go` pass a destination value but not source, so this is clearly not the right place in practice as well.

mtrmac: For a start, note that `SystemContext` is not only used for copies.

Then, ideally, trace the full flow of `SystemContext` values and how various fields get set, and try to find a place that fits. Given the uses of `SystemContext` in cmd/buildah, vs. the various code paths that currently don’t use the `SystemContextFromOptions` value at all (just like some of the sources/destinations for `getCopyOptions`), that might involve a larger-scale audit of what the right thing to do is.

rhatdan: Your right, I think I have fixed it correctly now.

mtrmac: What about all the other code paths?

rhatdan: Other code paths are not using dest dir.  Does the big file path get used on pushing?  Or only pulling?

buildah from and I believe buildah bud as well as buildah pull are all using the same pull logic.

rhatdan: I just confirmed the buildah bud does use the new code.  So all pull paths are using it.

mtrmac: - Commit is a copy from a Buildah-specific source to c/storage, that probably involves the temporary directory
- Other uses of the temporary directory may be added in the future. What’s the user expectation and the right way to implement it? My first impression is that this is (advertised as) a Buildah-general treatment of $TMPDIR that affect absolutely everything.

rhatdan: After reviewing the code we are handling the .BigFilesTemporaryDir in those cases, so this was only broken when pulling.

pull.go
```diff
@@ -292,7 +292,9 @@ func pullImage(ctx context.Context, store storage.Store, srcRef types.ImageRefer
 	}()
 
 	logrus.Debugf("copying %q to %q", transports.ImageName(srcRef), destName)
-	if _, err := retryCopyImage(ctx, policyContext, maybeCachedDestRef, srcRef, srcRef, getCopyOptions(store, options.ReportWriter, sc, nil, "", options.RemoveSignatures, "", nil, nil, options.OciDecryptConfig), options.MaxRetries, options.RetryDelay); err != nil {
+	destinationCtx := getSystemContext(store, nil, "")
+	destinationCtx.BigFilesTemporaryDir = sc.BigFilesTemporaryDir
```
nalind, line 296: Why not set `BigFilesTemporaryDir` in `getSystemContext()`?  Why don't we reuse the source context or use it as the source of defaults when creating `destinationCtx`?

mtrmac: FWIW there’s a bit of history to this, it turns out: The original was #2022, accidentally broken in #2868 (but the pull case that is the focus of this PR was probably never handled correctly).

---

I agree that the flow of defaults/options is counterintuitive and likely wrong; just like the value of `DockerRegistryUserAgent` set in `SystemContextFromOptions`, this is supposed to be a codebase-wide default that applies to ~everything, not something only related to processing of CLI options.

Having defaults in `getSystemContext` would make more sense for the core c/buildah code, OTOH not all users of a `*types.SystemContext` do pass through `getSystemContext` (e.g. not the uses in `c/cmd/buildah/common.go`) — that might not matter if all of them are just reads of c/storage metadata.

---

One concern with using the same `SystemContext` for both the source and destination of a copy is that `SystemContextFromOptions` may set `DockerAuthConfig`, which is not explicitly associated with a registry, and could cause credentials intended for the source to be used in the destination, or vice versa. I don’t expect that can currently happen (can Buildah cause a `docker://`-`docker://` copy?), but it’s a risk for the general idea.

rhatdan: Ok I simplified it to just push in the source as the destination though the  getSystemContext(store, sc, "")

common.go
TomSweeneyRedHat: the changes to common.go don't seem to be in the pr now, was that intended?

TomSweeneyRedHat: Never mind, needed to dig deeper.

TomSweeneyRedHat:
LGTM
but want a head nod from @nalind or @mtrmac

pull.go
```diff
@@ -292,7 +292,7 @@ func pullImage(ctx context.Context, store storage.Store, srcRef types.ImageRefer
 	}()
 
 	logrus.Debugf("copying %q to %q", transports.ImageName(srcRef), destName)
-	if _, err := retryCopyImage(ctx, policyContext, maybeCachedDestRef, srcRef, srcRef, getCopyOptions(store, options.ReportWriter, sc, nil, "", options.RemoveSignatures, "", nil, nil, options.OciDecryptConfig), options.MaxRetries, options.RetryDelay); err != nil {
+	if _, err := retryCopyImage(ctx, policyContext, maybeCachedDestRef, srcRef, srcRef, getCopyOptions(store, options.ReportWriter, sc, getSystemContext(store, sc, ""), "", options.RemoveSignatures, "", nil, nil, options.OciDecryptConfig), options.MaxRetries, options.RetryDelay); err != nil {
```
nalind, line 295: Is there any difference between `sc` and `getSystemContext(store, sc, "")` when execution gets here?

rhatdan: Theoretically this could fire
```
	if store != nil {
		if sc.BlobInfoCacheDir == "" {
			sc.BlobInfoCacheDir = filepath.Join(store.GraphRoot(), "cache")
		}
		if sc.SystemRegistriesConfPath == "" && unshare.IsRootless() {
			userRegistriesFile := filepath.Join(store.GraphRoot(), "registries.conf")
			if _, err := os.Stat(userRegistriesFile); err == nil {
				sc.SystemRegistriesConfPath = userRegistriesFile
			}
		}
	}
```

rhatdan: And create a difference

nalind:
LGTM once it passes CI.

PR #2901, opened 2021-01-11
Bump c/{common,image,storage}

mtrmac:
#### What type of PR is this?

> /kind other

Per @jnovy , updating vendored versions of c/{common, image,storage}

#### How to verify it
???


rhatdan:
No, this is not what we want. We should not be bumping these versions.  We should go out with the minimal changes in 8.3.1, these changes are not required until 8.4.

mtrmac:
For the record, continuing with this would require bumping containerd/containerd to at least 1.4.0 to avoid a dependency on removed `github.com/opencontainers/runc/libcontainer/system.GetClockTicks`

rhatdan:
Yup, way too many moving parts to rush this in at the last moment.

PR #2903, opened 2021-01-12
Stop overriding the location of the blob info cache

nalind:
#### What type of PR is this?

/kind bug

When it was first introduced, the blob info cache's location didn't change from the system-wide default location when we were running in rootless mode, so we started setting its location ourselves to avoid triggering permissions errors when updating it.

The image library has since started taking into account that it was running in rootless mode, but its hardwired default isn't the same as the one we were setting, so we ended up creating a second cache file.

Stop doing that.

#### How to verify it

Pull or push an image to or from a registry using buildah.  The cache file in /var/lib/containers/storage/cache or ~/.local/share/containers/storage/cache should be ignored if it exists, but the cache file in /var/lib/containers/cache or ~/.local/share/containers/cache should be updated.

None

General behavior shouldn't change, but we should no longer be creating a /var/lib/containers/storage/cache directory or a ~/.local/share/containers/storage/cache directory, since those aren't the default locations.

```
Buildah no longer uses blob info cache data in /var/lib/containers/storage/cache or ~/.local/share/containers/storage/cache.  It should share the cache files used by skopeo and other tools.
```

### DIFF
--- a/common.go
+++ b/common.go
@@ -56,9 +56,6 @@ func getSystemContext(store storage.Store, defaults *types.SystemContext, signat
 		sc.SignaturePolicyPath = signaturePolicyPath
 	}
 	if store != nil {
-		if sc.BlobInfoCacheDir == "" {
-			sc.BlobInfoCacheDir = filepath.Join(store.GraphRoot(), "cache")
-		}
 		if sc.SystemRegistriesConfPath == "" && unshare.IsRootless() {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

rhatdan:
@TomSweeneyRedHat PTAL

TomSweeneyRedHat:
LGTM

PR #2905, opened 2021-01-13
Cirrus: Trim CI to essential-only for legacy branch

cevich:


cevich:
Huh...Hmmm...that was easier than I expected.   Switching to container images from quay.io will also make testing here more reliable (i.e. the images are less likely to change w/o notice).

Issue #2904, opened 2021-01-13
Local Image not found during build -  Buildah v1.19


TomSweeneyRedHat:
**Description**

**Steps to reproduce the issue:**
Run this script:

```
# cat ./test_on_build.sh
########
# Build Dockerfiles for OnBuild Test
# (Thanks @clcollins!)
########
FILE=./Dockerfile
/bin/cat <<EOM >$FILE
FROM alpine
RUN touch /foo
ONBUILD RUN touch /bar
EOM
chmod +x $FILE

FILE=./Dockerfile-2
/bin/cat <<EOM >$FILE
FROM onbuild-image
RUN touch /baz
EOM
chmod +x $FILE

########
# Build with Dockerfiles
########
buildah bud -f ./Dockerfile --format=docker -t onbuild-image .
buildah bud -f ./Dockerfile-2 --format=docker -t result-image .

########
# Build a container to see if the /bar file has been created.
########
ctr=$(buildah from result-image)

########
# Validate that the /bar file has been created in the container.
########
buildah run $ctr ls -alF /bar /foo /baz

```

The second `buildah bud` command fails as the onbuild-image is not in the list of repositories, it's in the local host:

```
[root@localhost ~]# buildah images
REPOSITORY                          TAG      IMAGE ID       CREATED             SIZE
localhost/onbuild-image             latest   dc2e3e812ed1   3 minutes ago       5.85 MB
```

**Describe the results you received:**
```
# ./test_on_build.sh
STEP 1: FROM alpine
STEP 2: RUN touch /foo
STEP 3: ONBUILD RUN touch /bar
STEP 4: COMMIT onbuild-image
Getting image source signatures
Copying blob 777b2c648970 skipped: already exists  
Copying blob 431328ab6dbf done  
Copying config 2cb186bff2 done  
Writing manifest to image destination
Storing signatures
--> 2cb186bff2b
2cb186bff2bafbefa591a6e08121ae05bc8e1032cb0faa395483aee7e1f8eaec
STEP 1: FROM onbuild-image
? Please select an image: 
  ▸ registry.fedoraproject.org/onbuild-image:latest
    registry.access.redhat.com/onbuild-image:latest
    registry.centos.org/onbuild-image:latest
    docker.io/library/onbuild-image:latest
 ```
Hitting return causes an image not found error for any of these.

**Describe the results you expected:**

I at least expected a localhost/onbuild-image to be one of the selectable images and then successful completion.

[...]

PR #2908, opened 2021-01-14
use local image name for pull policy checks

vrothberg:
Some pull policies require to first look up a local image and compare
that to the remote counter part.  When looking up the remote image, we
need to make sure to use the name of the local image, if it exists.

This fixes a bug where a short name resolved to an image with the
"localhost/" prefix.  This prefix is only used for local image look ups
via `shortnames.ResolveLocally`.  Hence, when looking up the remote
counter part, we must preserve this prefix.

Fixes: #2904

### DIFF
--- a/new.go
+++ b/new.go
@@ -103,27 +103,27 @@ func newContainerIDMappingOptions(idmapOptions *IDMappingOptions) storage.IDMapp
 	return options
 }
 
-func resolveLocalImage(systemContext *types.SystemContext, store storage.Store, options BuilderOptions) (types.ImageReference, string, *storage.Image, error) {
+func resolveLocalImage(systemContext *types.SystemContext, store storage.Store, options BuilderOptions) (types.ImageReference, string, string, *storage.Image, error) {
 	candidates, _, _, err := util.ResolveName(options.FromImage, options.Registry, systemContext, store)
 	if err != nil {
-		return nil, "", nil, errors.Wrapf(err, "error resolving local image %q", options.FromImage)
+		return nil, "", "", nil, errors.Wrapf(err, "error resolving local image %q", options.FromImage)
 	}
-	for _, image := range candidates {
-		img, err := store.Image(image)
+	for _, imageName := range candidates {
+		img, err := store.Image(imageName)
 		if err != nil {
 			if errors.Cause(err) == storage.ErrImageUnknown {
 				continue
 			}
-			return nil, "", nil, err
+			return nil, "", "", nil, err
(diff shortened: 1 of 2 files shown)


vrothberg:
@mtrmac @rhatdan @TomSweeneyRedHat PTAL

rhatdan:
LGTM
Yes we need to get this into podman 3.0

giuseppe (approved):
/lgtm

new.go
```diff
@@ -166,6 +166,12 @@ func resolveImage(ctx context.Context, systemContext *types.SystemContext, store
 		}
 	}
 
+	// If we found a local image, we must use it's name.
+	// See #2904.
+	if localImageRef != nil {
+		fromImage = localImageName
+	}
+
 	resolved, err := shortnames.Resolve(systemContext, fromImage)
```
mtrmac, line 175: This is now stuttering and hard to follow.

How about
```go
if options.PullPolicy == PullNever && localImage == nil { /* fail*/ }
if localImage != nil {
   // the three checks
}
```

new.go
```diff
@@ -145,7 +145,7 @@ func resolveImage(ctx context.Context, systemContext *types.SystemContext, store
 		}
 	}
 
-	localImageRef, _, localImage, err := resolveLocalImage(systemContext, store, options)
+	localImageRef, _, localImageName, localImage, err := resolveLocalImage(systemContext, store, options)
```
mtrmac, line 148: Unrelated? This is the only caller, yet we ignore the second return value. It should be dropped, then.

new.go
```diff
@@ -166,6 +166,12 @@ func resolveImage(ctx context.Context, systemContext *types.SystemContext, store
 		}
 	}
 
+	// If we found a local image, we must use it's name.
+	// See #2904.
+	if localImageRef != nil {
+		fromImage = localImageName
```
mtrmac, line 172: (Non-blocking: It might be useful to add a debug log if `fromImage != localImageName`, so that users can figure out where the fully-qualified variant came from.)

mtrmac (commented):
Great point, with things like `PullIfNewer` we should always only use a remote image from the same registry as the local one.

new.go
```diff
@@ -166,6 +166,12 @@ func resolveImage(ctx context.Context, systemContext *types.SystemContext, store
...
+	}
```
TomSweeneyRedHat, line 173: Neat workaround, hadn't thought about trying this.

PR #2911, opened 2021-01-14
Fix version of release to v1.19.1

rhatdan:
Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### Does this PR introduce a user-facing change?

```release-note

```

PR #2011, opened 2019-12-07
Add support for containers.conf

rhatdan:
This is a rework of Qi Wang's patches.
Import package pkg/config from containers/common to read containers.conf
This patch allows users to specify default values stored in containers.conf
that will modify the behaviour of buildah tool.

Signed-off-by: Qi Wang <qiwan@redhat.com>
Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

Replaces https://github.com/containers/buildah/pull/1953

PR #2912, opened 2021-01-14
Bump to v1.19.1

TomSweeneyRedHat:
As the title says.

### DIFF
--- a/contrib/rpm/buildah.spec
+++ b/contrib/rpm/buildah.spec
@@ -100,7 +100,7 @@ make DESTDIR=%{buildroot} PREFIX=%{_prefix} install install.completions
 %{_datadir}/bash-completion/completions/*
 
 %changelog
-* Fri Jan 8, 2021 Tom Sweeney <tsweeney@redhat.com> 1.19.1-1
+* Thu Jan 14, 2021 Tom Sweeney <tsweeney@redhat.com> 1.19.1-1
 - Cherry pick localhost fix and update CI configuration for release-1.19 branch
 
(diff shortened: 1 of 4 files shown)


TomSweeneyRedHat:
@nalind @rhatdan PTAL

contrib/rpm/buildah.spec
```diff
@@ -100,6 +100,9 @@ make DESTDIR=%{buildroot} PREFIX=%{_prefix} install install.completions
 %{_datadir}/bash-completion/completions/*
 
 %changelog
+* Fri Jan 8, 2021 Tom Sweeney <tsweeney@redhat.com> 1.19.1-1
```
nalind, line 103: Date looks off, but not a blocker.

TomSweeneyRedHat: oops!  Thanks, fixed and repushed.

nalind:
LGTM, but it's either this one and moving the v1.19.1 tag, or #2911.

TomSweeneyRedHat:
I think it needs to be this one as #2011 has a different version of c/common than what is in Podman 3.0 and they need to sync.

rhatdan:
Already forced this through, but we need another fix.

nalind:
> Already forced this through, but we need another fix.

Can you elaborate?  I'm looking at this and I have _no_ idea what's going on.

rhatdan:
Were all set, I opened PRs to fix an issue with pkg/overlay. 
Buildah 1.19.2 is the current release.

PR #2916, opened 2021-01-15
1.19.2

rhatdan:
Cherry pick updating containers/common and fixing overlay mount points to not throw errors on missing
destination.

Generate a new release of v1.19.2

### DIFF
--- a/buildah.go
+++ b/buildah.go
@@ -28,7 +28,7 @@ const (
 	Package = "buildah"
 	// Version for the Package.  Bump version in contrib/rpm/buildah.spec
 	// too.
-	Version = "1.19.0"
+	Version = "1.19.2"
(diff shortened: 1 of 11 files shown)


giuseppe (approved):
LGTM

vrothberg (commented):
/lgtm

buildah.go
```diff
@@ -28,7 +28,7 @@ const (
 	Package = "buildah"
 	// Version for the Package.  Bump version in contrib/rpm/buildah.spec
 	// too.
-	Version = "1.19.0"
+	Version = "1.19.2"
```
TomSweeneyRedHat, line 31: I'm concerned that the version wasn't originally 1.19.1 here and in bulidah.spec.  I'm not sure how that happened, but the end result looks OK....

PR #2924, opened 2021-01-20
Workaround for RHEL gating test failure

edsantiago:
One of the auth tests is failing on RHEL8.4:
```
   # buildah push --tls-verify=true ...
   Get "https://localhost:5000/v2/": x509: certificate relies on legacy Common Name field, use SANs or temporarily enable Common Name matching with GODEBUG=x509ignoreCN=0
```
(Expected output: ' x509: certificate signed by unknown authority')

Cause: who knows? Some internal Go change.

Solution: set magic GODEBUG envariable in the buildah invocation.
This makes that message go away, and gives us the expected one.

Indicate that this may be a temporary workaround. I vaguely
recall reading that a future version of Go would remove it.
We'll deal with that then.

### DIFF
--- a/tests/authenticate.bats
+++ b/tests/authenticate.bats
@@ -33,7 +33,12 @@ load helpers
   expect_output --substring "Writing manifest to image destination"
 
   # With tls-verify=true, should fail due to self-signed cert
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM
assuming happy tests.

PR #2926, opened 2021-01-21
Pick default OCI Runtime from containers.conf

rhatdan:
Currently we have a weird situation where the user sets the default
runtime in his containers.conf for podman but Buildah is still falling
back to use runc because it was hard coded as the default for Buildah.

I would like to remove this default, but that would theoretically break
the API promise of Buildah.

This should fix https://github.com/containers/podman/issues/8893

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/util/types.go
+++ b/util/types.go
@@ -1,7 +1,7 @@
 package util
 
 const (
-	// DefaultRuntime is the default command to use to run the container.
+	// Deprecated: Default runtime should come from containers.conf
 	DefaultRuntime = "runc"
 	// DefaultCNIPluginPath is the default location of CNI plugin helpers.
 	DefaultCNIPluginPath = "/usr/libexec/cni:/opt/cni/bin"
--- a/util/util.go
+++ b/util/util.go
@@ -263,7 +263,12 @@ func Runtime() string {
 		return "crun"
 	}
 
-	return DefaultRuntime
+	conf, err := config.Default()
+	if err != nil {
+		logrus.Warnf("Error loading container config when searching for local runtime: %v", err)
+		return DefaultRuntime
+	}
+	return conf.Engine.OCIRuntime
 }
 
 // StringInSlice returns a boolean indicating if the exact value s is present
--- a/util/util_test.go
+++ b/util/util_test.go
@@ -2,7 +2,10 @@ package util
 
 import (
 	"fmt"
+	"os"
 	"testing"
+
+	"github.com/containers/common/pkg/config"
 )
 
 func TestMergeEnv(t *testing.T) {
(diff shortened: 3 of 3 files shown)


rhatdan:

util/types.go
```diff
@@ -1,8 +1,9 @@
 package util
 
 const (
-	// DefaultRuntime is the default command to use to run the container.
-	DefaultRuntime = "runc"
+	// DefaultRuntime should no longer be used.
+	// Default runtime should come from containers.conf
+	DefaultRuntime = "crun"
```
vrothberg, line 6: I think this should be "runc" for now. It's safe for Fedora but who knows in which packages this ends up.

util/util.go
```diff
@@ -263,7 +263,12 @@ func Runtime() string {
 		return "crun"
 	}
 
-	return DefaultRuntime
+	conf, err := config.Default()
+	if err != nil {
+		logrus.Debugf("Error loading container config when searching for local runtime.")
```
vrothberg, line 268:
```suggestion
		logrus.Debugf("Error loading container config when searching for local runtime: %v", err)
```
I wonder whether this error should be fatal.  I didn't check the code but I would assume that we don't get an error if the config does not exist ... but if there's some real config/syntax/permission error here, then it would be very hard to figure out.

util/util_test.go
```diff
@@ -37,3 +40,18 @@ func TestMergeEnv(t *testing.T) {
 		})
 	}
 }
+func TestRuntime(t *testing.T) {
+	os.Setenv("CONTAINERS_CONF", "/dev/null")
+	conf, _ := config.Default()
+	defaultRuntime := conf.Engine.OCIRuntime
+	runtime := Runtime()
+	if runtime != defaultRuntime {
```
eriksjolund, line 48: Maybe this is better?
```	
if runtime != defaultRuntime {
	t.Fatalf("expected %v, got %v", runtime, defaultRuntime)
}
```

util/types.go
rhatdan: I actually think we should remove this altogether or maybe return "".  This field should never be used.  Basically it would only ever be used if containers/common had a serious issue, where their was a coding mistake in the code, which has already gone through  multiple layers of testing.

rhatdan: But to get it through, I changed it back. Also fixed the message if containers.conf is screwed up to Warnf level.

util/types.go
```diff
@@ -1,7 +1,8 @@
 package util
 
 const (
-	// DefaultRuntime is the default command to use to run the container.
+	// DefaultRuntime should no longer be used.
+	// Default runtime should come from containers.conf
 	DefaultRuntime = "runc"
```
saschagrunert, line 6:
```suggestion
	// Deprecated: Default runtime should come from containers.conf
	DefaultRuntime = "runc"
```
Note that this change may trigger linters to complain.

util/util.go
```diff
@@ -263,7 +263,12 @@ func Runtime() string {
 		return "crun"
 	}
 
-	return DefaultRuntime
+	conf, err := config.Default()
+	if err != nil {
+		logrus.Warnf("Error loading container config when searching for local runtime.: %v", err)
```
saschagrunert, line 268: nit:
```suggestion
		logrus.Warnf("Error loading container config when searching for local runtime: %v", err)
```

vrothberg (commented):
Can you squash the commits?

rhatdan:
Everything commented on was merged and fixed.

saschagrunert (approved):
LGTM

vrothberg (commented):
/lgtm
/hold

vrothberg:
@rhatdan needs a rebase :(

TomSweeneyRedHat:
LGTM
@vrothberg or @nalind, want to add a final lgtm?

PR #2929, opened 2021-01-23
Pick default OCI Runtime from containers.conf 

rhatdan:
Cherry-pick     "Pick default OCI Runtime from containers.conf"

### DIFF
--- a/util/types.go
+++ b/util/types.go
@@ -1,7 +1,7 @@
 package util
 
 const (
-	// DefaultRuntime is the default command to use to run the container.
+	// Deprecated: Default runtime should come from containers.conf
 	DefaultRuntime = "runc"
 	// DefaultCNIPluginPath is the default location of CNI plugin helpers.
 	DefaultCNIPluginPath = "/usr/libexec/cni:/opt/cni/bin"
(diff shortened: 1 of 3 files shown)


rhatdan:
@TomSweeneyRedHat @giuseppe @vrothberg @saschagrunert PTAL

buildah.go
```diff
@@ -28,7 +28,7 @@ const (
 	Package = "buildah"
 	// Version for the Package.  Bump version in contrib/rpm/buildah.spec
 	// too.
-	Version = "1.19.2"
+	Version = "1.19.3"
```
TomSweeneyRedHat, line 31: you need to bump the contrib/rpm/buildah.spec too, and it would be good to update changelog.txt and CHANGELOG.md while you're at it.

TomSweeneyRedHat:
Tests are unhappy, I think due to the missing verison in the spec.  Beyond that, I really much prefer having the Buildah version bump in their own PR, then any cherry picks in a separate.  That way you can more easily find the cherrypick PR in GitHub's PR list.

buildah.go
rhatdan: Ok I jumped the gun on this release, since there is at least one more thing we want to back port.  Removing the release and only adding the cherry-pick

TomSweeneyRedHat:
LGTM
@containers/build-maintainers another LGTM for the merge win please?

PR #2930, opened 2021-01-25
local image lookup by digest

vrothberg:
Detect local-image lookups by digest.  Those clearly refer to local
images only, so we must not proceed to remote lookups.

Fixes: #2836

### DIFF
--- a/new.go
+++ b/new.go
@@ -150,10 +150,10 @@ func resolveImage(ctx context.Context, systemContext *types.SystemContext, store
 		return nil, "", nil, err
 	}
 
-	// If we could resolve the image locally, check if it was referenced by
-	// ID.  In that case, we don't need to bother any further and can
-	// prevent prompting the user.
-	if localImage != nil && strings.HasPrefix(localImage.ID, options.FromImage) {
+	// If we could resolve the image locally, check if it was clearly
+	// referring to a local image, either by ID or digest.  In that case,
+	// we don't need to perform a remote lookup.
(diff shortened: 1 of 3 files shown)


vrothberg:

rhatdan:
LGTM

nalind:
My experiments with `docker build` suggest that it finds an image if the hex portion of the digest specified to a `FROM` instruction matches an image ID, but not if it matches an image's (manifest's) digest.  This form is also recognized by `docker create`, so it may belong in the storage library.

vrothberg:
> My experiments with `docker build` suggest that it finds an image if the hex portion of the digest specified to a `FROM` instruction matches an image ID, but not if it matches an image's (manifest's) digest. This form is also recognized by `docker create`, so it may belong in the storage library.

Shall we support both?  We could first assume it's an image ID, and then fall back to the real digests.  Could handle both in the Buildah.

nalind:
> > My experiments with `docker build` suggest that it finds an image if the hex portion of the digest specified to a `FROM` instruction matches an image ID, but not if it matches an image's (manifest's) digest. This form is also recognized by `docker create`, so it may belong in the storage library.
> 
> Shall we support both?

I'm personally in no hurry to encourage either, and `docker` doesn't support the latter.

vrothberg:
Updated to only cover the image ID.

vrothberg:
@cevich is Ubuntu sick?

rhatdan:

TomSweeneyRedHat:
@vrothberg needs a rebase.

PR #2928, opened 2021-01-22
Use build-arg ENV val from local environment if set

umohnani8:
Signed-off-by: Urvashi Mohnani <umohnani@redhat.com>

#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

If a user sets the ENV to be used with the build-arg flag in
the local environment by exporting it, look it up and use the
value set there for that ENV.
Add tests to cover this use case as well.

#### Which issue(s) this PR fixes:

Fixes https://github.com/containers/buildah/issues/2888

#### Does this PR introduce a user-facing change?

```release-note

```

PR #2932, opened 2021-01-25
[release-1.19] Use build-arg ENV val from local environment if set

umohnani8:
Backport of https://github.com/containers/buildah/pull/2928

#### What type of PR is this?

/kind bug

If a user sets the ENV to be used with the build-arg flag in
the local environment by exporting it, look it up and use the
value set there for that ENV.
Add tests to cover this use case as well.

```release-note

```


umohnani8:

TomSweeneyRedHat:
LGTM

PR #2938, opened 2021-01-26
Rebuild layer if a change in ARG is detected

umohnani8:

#### What type of PR is this?

/kind bug

Check whether the ARG in the containerfile is changed by
either the --build-arg flag or local environment and use
the cached layer or rebuild the layer accordingly.
Add tests for this use case as well.

Fixes https://github.com/containers/buildah/issues/2848
Fixes https://github.com/containers/buildah/issues/2837

```release-note

```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -303,17 +303,17 @@ symlink(subdir)"
   # two more, starting at the "echo $user" instruction
   run_buildah bud --signature-policy ${TESTSDIR}/policy.json --build-arg=user=1 --layers -t test1 -f Dockerfile.build-args ${TESTSDIR}/bud/use-layers
   run_buildah images -a
-  expect_line_count 7
+  expect_line_count 8
 
   # one more, because we added a new name to the same image
   run_buildah bud --signature-policy ${TESTSDIR}/policy.json --build-arg=user=1 --layers -t test2 -f Dockerfile.build-args ${TESTSDIR}/bud/use-layers
   run_buildah images -a
-  expect_line_count 8
+  expect_line_count 9
 
   # two more, starting at the "echo $user" instruction
   run_buildah bud --signature-policy ${TESTSDIR}/policy.json --layers -t test3 -f Dockerfile.build-args ${TESTSDIR}/bud/use-layers
   run_buildah images -a
-  expect_line_count 10
+  expect_line_count 12
 }
 
 @test "bud with --rm flag" {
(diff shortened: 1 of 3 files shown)


umohnani8:
@nalind @rhatdan PTAL

TomSweeneyRedHat:
All kinds of test unhappiness @umohnani8

rhatdan:
LGTM

rhatdan:

tests/bud.bats
```diff
@@ -1716,6 +1717,30 @@ _EOF
   run_buildah bud --signature-policy ${TESTSDIR}/policy.json --layers -t ${target} -f Dockerfile3 --build-arg=PGDATA=/pgdata --build-arg=UID=17122 --build-arg=CODE=/copr/coprs_frontend --build-arg=USERNAME=praiskup ${TESTSDIR}/bud/build-arg
   run_buildah inspect -f '{{.FromImageID}}' ${target}
   expect_output "$argsid"
+
+  # If build-arg is specifified via the command line and is different form the previous cached build, it should not use the cached layers.
```
TomSweeneyRedHat, line 1721: nit
```suggestion
  # If build-arg is specifified via the command line and is different from the previous cached build, it should not use the cached layers.
```

tests/bud.bats
```diff
@@ -1716,6 +1717,30 @@ _EOF
...
+  run_buildah inspect -f '{{.FromImageID}}' test-img-2
+  argsid="$output"
+  [[ "$argsid" != "$initialid" ]]
+
+  # Set the build-arg via an ENV in the local environment and verify that the cached layers are not used
+  export TEST=bar
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json --layers -t test-img-3 --build-arg TEST -f Dockerfile4 ${TESTSDIR}/bud/build-arg
+  run_buildah inspect -f '{{.FromImageID}}' test-img-3
+  argsid="$output"
+  [[ "$argsid" != "$initialid" ]]
```
TomSweeneyRedHat, line 1743: do we have similar tests without the `--layers` option?

TomSweeneyRedHat: not sure that we need them, WDYT @umohnani8 ?

TomSweeneyRedHat:
LGTM
One very small question and one nit of a nit in a code comment.
Once I hear back from @umohnani8 I'll merge or wait for the quick repush before merging.

tests/bud.bats
umohnani8: Fixed

tests/bud.bats
umohnani8: I don't think we need them. When building without the `--layers` flag we always go through and build every step so should not be an issue in that case.

TomSweeneyRedHat:
LGTM

umohnani8:
We should backport to 1.19 right?

Issue #2423, opened 2020-06-25
Buildah bud writes stdout to stderr, which causes pipeline builds to fail.


erbzn:
**Description**

I am trying to use Buildah in Azure DevOps Pipeline with the example repository phippyandfriends (https://github.com/Azure/phippyandfriends).
First step is to build phippy image. This image build fails with "buildah bud", as it writes some standard output to standard error, e.g.
`Copying blob sha256:df20fa9351a15782c64e6dddb2d4a6f50bf6d3688060a34c4014b0d9a752eb4c`.
I am using a self-hosting agent built from the buildah image "quay.io/buildah/stable:latest" and configured to connect to Azure DevOps agent pool with the following Dockerfile (start.sh connects to agent pool):
```
FROM quay.io/buildah/stable:latest

RUN yum install -y \
        git \
        iputils \
        libicu \
        libunwind \
        nmap-ncat \
        unzip \
        jq

WORKDIR /azp

COPY ./start.sh .
RUN chmod +x start.sh

CMD ["./start.sh"]
```

Within this Container I want to build the following image:
```
FROM composer:1 AS build-env
COPY . /app
RUN cd /app && composer install

FROM php:7.3.12-apache-stretch
ENV PORT 3333
EXPOSE 3333
COPY --from=build-env /app /var/www/html
RUN usermod -u 1000 www-data; \
    a2enmod rewrite; \
    chown -R www-data:www-data /var/www/html
```
I've tested with 2 versions of buildah (1.14.9, 1.16.0-dev), everytime it is writing parts to stderr but the build does not fail.
Pushing and running built image works as normal afterwards.

**Steps to reproduce the issue:**
1. Run buildah image (tested with docker):
`docker run -it --net=host --security-opt label=disable --security-opt seccomp=unconfined --device /dev/fuse:rw quay.io/buildah/stable:latest /bin/bash`
2. Inside container install git and clone phippyandfriends:
`git clone https://github.com/Azure/phippyandfriends.git`
3. Build image and write stdout and stderr to files:
`buildah bud phippy/. > stdout.txt 2> stderr.txt `

**Describe the results you received:**
```
[root@docker-desktop s]# cat stdout.txt
STEP 1: FROM composer:1 AS build-env
STEP 2: COPY . /app
STEP 3: RUN cd /app && composer install
STEP 4: FROM php:7.3.12-apache-stretch
STEP 5: ENV PORT 3333
STEP 6: EXPOSE 3333
[...]

PR #2942, opened 2021-01-27
DO NOT MERGE: Write output to stdout and not stderr

umohnani8:

#### What type of PR is this?

/kind bug

We had set the report writer to stderr so some info during
builds were being written to stderr. Move everything over
to stdout.

Fixes https://github.com/containers/buildah/issues/2423

```release-note

```

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -210,7 +210,7 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budOptions) error {
 	var stdout, stderr, reporter *os.File
 	stdout = os.Stdout
 	stderr = os.Stderr
-	reporter = os.Stderr
+	reporter = os.Stdout
 	if c.Flag("logfile").Changed {
 		f, err := os.OpenFile(iopts.Logfile, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, 0600)
 		if err != nil {
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -215,7 +215,7 @@ func commitCmd(c *cobra.Command, args []string, iopts commitInputOptions) error 
 	}
 
 	if !iopts.quiet {
-		options.ReportWriter = os.Stderr
+		options.ReportWriter = os.Stdout
 	}
 	id, ref, _, err := builder.Commit(ctx, dest, options)
 	if err != nil {
--- a/cmd/buildah/from.go
+++ b/cmd/buildah/from.go
@@ -298,7 +298,7 @@ func fromCmd(c *cobra.Command, args []string, iopts fromReply) error {
 	}
 
 	if !iopts.quiet {
-		options.ReportWriter = os.Stderr
+		options.ReportWriter = os.Stdout
 	}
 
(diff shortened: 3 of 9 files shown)


umohnani8:
@rhatdan any idea why my PRs are not automatically approved when I am an approver now?

rhatdan:
This PR concerns me. since we might be breaking scripts. @nalind @TomSweeneyRedHat @vrothberg PTAL

rhatdan:
Code LGTM

vrothberg (commented):
I am afraid of backwards compatibility. Users will likely depend on on only the final image ID being printed on stdout.  I assume that many integration tests do as well.

umohnani8:
@rhatdan @TomSweeneyRedHat I am seeing a few tests failing e.g https://cirrus-ci.com/task/4739030875111424?command=integration_test#L1032. I am a bit confused on how switching to stdout is causing the exit code to be 125 for those tests, do you have any idea?

TomSweeneyRedHat:
I've dabbled in the stdout/stderr stuff in Buildah in the past and it looks simple, but has complex behaviors.  I'd like to sit on this one until after we get through the Podman v3.0 delivery then take a deeper dive.

edsantiago:
@umohnani8 the root cause of at least one failure is this:

https://github.com/containers/buildah/blob/0ec651fd3c4a26eebaf3ca857b46138b23aec42e/tests/containers_conf.bats#L14

`buildah` should (almost) never be run in buildah gating tests. The correct usage is `run_buildah`, with the underscore.  This line is not something you touched, but by mucking with stdout/stderr the effect is:

  - old: the `$(...)` would capture stdout, which is only the container ID, which is nice and good
  + new: the `$(...)` captures stdout, including "Getting blah blah", so `$cid` is now "Getting blah blah"

This is a pretty standard existing convention: buildah and podman spit out verbose stuff to stderr, and the critical stuff (image ID, container ID) to stdout. Changing this will cause breakage worldwide.

Side note: I recommend installing [my greasemonkey log highlighter](https://github.com/edsantiago/greasemonkey/tree/master/podman-ginkgo-highlight). It is nearly impossible for a human to read podman/buildah CI logs without it.

umohnani8:
Thank you for the tip @edsantiago, using the `--quiet` flag helped fix the test issues.

vrothberg:
/hold

Let's find consensus.  I don't think we can merge without breaking an unknown number of users.  I am in favor of not merging as I am too afraid of the consequences.  Also, it would break own scripts :-D

edsantiago:
I agree with the hold. @umohnani8 my comment wasn't intended to shut-up the tests, my goal was to demonstrate the problem and why it's not fixable. The `cid=$(buildah from ...)` pattern is, I'm afraid, universal. Breaking that would affect pretty much everyone who uses buildah, and would be unpopular. I would encourage us to think very carefully here.

TomSweeneyRedHat:
Beyond any doubt, I don't want this potentially merging before Podman 3.0 goes out, bumped the title for now.

umohnani8:
Closing as we can use the `--quiet` flag to suppress extra outputs to stderr.

PR #2955, opened 2021-02-02
buildah manifest add localimage should work

rhatdan:
Currently if you attempt to build create a manifest
and add a local image, the command blows up.

The current code always looks for a remote image.
This PR fixes the code to use the local image if it
exists.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/cmd/buildah/manifest.go
+++ b/cmd/buildah/manifest.go
@@ -307,7 +307,15 @@ func manifestAddCmd(c *cobra.Command, args []string, opts manifestAddOpts) error
 
 	digest, err := list.Add(getContext(), systemContext, ref, opts.all)
 	if err != nil {
-		return err
+		var storeErr error
+		// check if the local image exists
+		if ref, _, storeErr = util.FindImage(store, "", systemContext, imageSpec); storeErr != nil {
+			return err
+		}
+		digest, storeErr = list.Add(getContext(), systemContext, ref, opts.all)
+		if storeErr != nil {
+			return err
+		}
 	}
 
 	if opts.os != "" {
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -566,9 +566,9 @@ function _test_http() {
   starthttpd "${TESTSDIR}/bud/$testdir"
   target=scratch-image
   run_buildah bud --signature-policy ${TESTSDIR}/policy.json \
-              -t ${target} \
-              "$@"         \
-              http://0.0.0.0:${HTTP_SERVER_PORT}/$urlpath
+	      -t ${target} \
+	      "$@"         \
+	      http://0.0.0.0:${HTTP_SERVER_PORT}/$urlpath
   stophttpd
   run_buildah from ${target}
 }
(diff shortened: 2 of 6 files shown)


nalind:
Does this reintroduce the problem that #2824 was attempting to fix?

cmd/buildah/manifest.go
```diff
@@ -307,7 +307,15 @@ func manifestAddCmd(c *cobra.Command, args []string, opts manifestAddOpts) error
 
 	digest, err := list.Add(getContext(), systemContext, ref, opts.all)
 	if err != nil {
-		return err
+		var storeErr error
+		// check if the local image exists
+		if ref, _, storeErr = util.FindImage(store, "", systemContext, imageSpec); storeErr != nil {
```
QiWang19, line 312: The local image has been checked above in line297?
 https://github.com/containers/buildah/pull/2955/files#diff-f87fb1de87ea57699bf6ddd7408c420b8bd7daa0f4ef41639120453a94c52b14R297

rhatdan: That check never happens since the imagespec is parsed correctly as a remote image, so we always fail if we use a local image that does not exists at a registry.

rhatdan: The remote image is never checked until this call.

rhatdan:
@nalind I don't see how, since it will first check at the remote site.  It will only check local storage if it does not exists on the remote site.

QiWang19 (commented):
LGTM

cmd/buildah/manifest.go
QiWang19: > That check never happens since the imagespec is parsed correctly as a remote image, so we always fail if we use a local image that does not exists at a registry.

LGTM. We could also delete check L296~L297 since they can't be reached.

rhatdan: No I can reach there by attempting

./bin/buildah manifest add myimage  sha256:c1876ff0a66a49ae90dbe6ac1d8c413107fec317987c4e2059cf72b430f5a023

Which is also valid.

tests/bud.bats
```diff
@@ -2509,3 +2509,47 @@ _EOF
...
+@test "bud with --manifest flag new manifest" {
+  _prefetch alpine
+  mytmpdir=${TESTDIR}/my-dir
+  mkdir -p ${mytmpdir}
+cat > $mytmpdir/Containerfile << _EOF
+from alpine
+run echo hello
+_EOF
+
+  run_buildah bud -q --manifest=testlist -t arch-test --signature-policy ${TESTSDIR}/policy.json ${mytmpdir} <<< input
```
TomSweeneyRedHat, line 2522: is the `<<< input` at the end here correct?  I'm not seeing a `--stdin` option.

rhatdan: Yes this came from other tests. Not sure what it does.

TomSweeneyRedHat: It must just be passing `input` in as input into the build command and then ignoring it.
Weird, but OK and the tests are happy.

TomSweeneyRedHat:
LGTM

PR #2960, opened 2021-02-03
Change default to 0777 for directories created from tar balls

rhatdan:
There is a change in Docker 20.10 release where they changed the
default directory permissions of directories created in images
via tar balls from 0755 to 0777. Docker 19.* created them as 0755.
This change meant that our conformance tests were breaking.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -1294,7 +1294,7 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
 		dirUID, dirGID = req.PutOptions.ChownDirs.UID, req.PutOptions.ChownDirs.GID
 		defaultDirUID, defaultDirGID = dirUID, dirGID
 	}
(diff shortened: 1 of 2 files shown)


nalind:
I'm wondering if we want to provide an override value in our callers in our `add.go` to avoid changing behavior for other callers, for example `podman cp`.  Any idea if they're depending on this default?

TomSweeneyRedHat:
Conformance tests are failing still @rhatdan

rhatdan:
I think it make sense that we do this only for the tar balls.

rhatdan:
Since Docker does not know if this is correct, closing.

PR #2962, opened 2021-02-03
blobcache: fix sequencing error

nalind:
#### What type of PR is this?

/kind bug

When saving a blob to a blob cache, we use a goroutine to store the possibly-decompressed version of the incoming blob to disk.  We compare the resulting digest to the digest of the blob in its original form to decide whether to mark the two digests as a compressed/uncompressed pair, or to just mark the single digest as a known blob.  We weren't waiting for the goroutine to finish computing the digest before consulting the value that it computed, so this could result in the wrong decision.

The blobcache also only knows about gzip compressed blobs.  Until we properly teach it about other types of compression, at least update it to not set MIME types to indicate gzip compression unless we know we're using gzip compression.

#### How to verify it

It's a race condition, but running:
```
buildah rmi -a -f
rm -fr /tmp/blobcache; mkdir /tmp/blobcache; buildah pull --blob-cache=/tmp/blobcache fedora
```
may trigger it.  If the resulting cache directory includes compressed files for which there is no corresponding file (the corresponding file will have the same name as the compressed file, but with a ".decompressed" suffix added), you're seeing the bug.

None

```
None
```

### DIFF
--- a/pkg/blobcache/blobcache.go
+++ b/pkg/blobcache/blobcache.go
@@ -13,6 +13,7 @@ import (
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/image"
 	"github.com/containers/image/v5/manifest"
+	"github.com/containers/image/v5/pkg/compression"
 	"github.com/containers/image/v5/transports"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage/pkg/archive"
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
needs a rebase @nalind

nalind:
Rebased.

rhatdan:
If @giuseppe zstd patch gets in, do we need to change this code again to create the new format?
LGTM

TomSweeneyRedHat (commented):
LGTM

giuseppe (approved):
LGTM

PR #2968, opened 2021-02-04
Stop testing directory permissions with latest docker

rhatdan:
Current version of Docker, has a bug we believe, that is
creating top level directories when using ADD and COPY of
tar balls. Basically the directories end up with a 777
permissions, Buildah creates these with 755, which we
believe is correct, and matches older versions of Docker.

We need to revert this patch once we have a version of Docker
that creates these direcories with the corret permissions.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/tests/conformance/conformance_test.go
+++ b/tests/conformance/conformance_test.go
@@ -1469,7 +1469,8 @@ var internalTestCases = []testCase{
 	},
 
 	{
-		name: "add-permissions",
+		name:          "add-permissions",
+		withoutDocker: true,
 		dockerfileContents: strings.Join([]string{
 			"FROM scratch",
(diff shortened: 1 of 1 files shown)


rhatdan:
Needs to be reverted once this https://github.com/moby/moby/issues/41978 issue is cleared up.

nalind:
That should work for the conformance tests, though I thought we were also tripping over the permissions change in some of the integration tests.

TomSweeneyRedHat:
LGTM
assuming happy tests

Issue #2966, opened 2021-02-04
podman 3.0rc2: podman build does not use the cache


terceiro:
/kind bug
I'm on Debian unstable, and was using podman 3.0 rc1 for some time. After I upgraded to 3.0 rc2, cache does not work anymore with `podman build` (or with buildah 0.19.3, using `buildah bud`). If I revert back to rc1, it works again, even if the previous build was done with rc2. That is, rc2 is able to write to the cache, but to read from it.

**Steps to reproduce the issue:**

With rc2:

```
$ cat Containerfile 
──────┬───────────────────────────────
      │ File: Containerfile
──────┼───────────────────────────────
  1   │ FROM debian
  2   │ RUN sleep 3
──────┴───────────────────────────────
$ time podman build -t foobar .
STEP 1: FROM debian
STEP 2: RUN sleep 3
STEP 3: COMMIT foobar
--> 022b39fc015
022b39fc01563426e6bbea87c1ddce78bce725c35c4d5c9d5abcc940adbf0b61

real	0m4,579s
user	0m0,781s
sys	0m0,827s
$ time podman build -t foobar .
STEP 1: FROM debian
STEP 2: RUN sleep 3
--> Using cache 022b39fc01563426e6bbea87c1ddce78bce725c35c4d5c9d5abcc940adbf0b61
STEP 3: COMMIT foobar
--> 022b39fc015
022b39fc01563426e6bbea87c1ddce78bce725c35c4d5c9d5abcc940adbf0b61

real	0m3,553s
user	0m0,455s
sys	0m0,444s
```

Note that it says it's using the cache, but it's not (look at the time it took to run).

After this point, I revert back to rc1, and it is able to read the cache created by rc2 (and finished in 0s):

```
$ time podman build -t foobar .
STEP 1: FROM debian
STEP 2: RUN sleep 3
--> Using cache 022b39fc01563426e6bbea87c1ddce78bce725c35c4d5c9d5abcc940adbf0b61
STEP 3: COMMIT foobar
--> 022b39fc015
022b39fc01563426e6bbea87c1ddce78bce725c35c4d5c9d5abcc940adbf0b61

real	0m0,222s
user	0m0,134s
sys	0m0,093s
```

**Output of `podman version`:**

```
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.15.7
Built:        Wed Dec 31 21:00:00 1969
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
[...]

PR #2967, opened 2021-02-04
Fix build arg check

umohnani8:
Signed-off-by: Urvashi Mohnani <umohnani@redhat.com>

#### What type of PR is this?

 /kind bug

#### What this PR does / why we need it:

Fix the check on build args to be the length of the map
and not whether the map is nil. The nil check was causing
the cache layer to not be used as it would give a false
result.

#### Which issue(s) this PR fixes:

Fixes https://github.com/containers/buildah/issues/2966

#### Does this PR introduce a user-facing change?

```release-note

```

PR #2969, opened 2021-02-04
Fix build arg check

umohnani8:
Backport of https://github.com/containers/buildah/pull/2967

#### What type of PR is this?

 /kind bug

Fix the check on build args to be the length of the map
and not whether the map is nil. The nil check was causing
the cache layer to not be used as it would give a false
result.

Fixes https://github.com/containers/buildah/issues/2966

```release-note

```


umohnani8:

TomSweeneyRedHat:
LGTM

vrothberg (commented):
/lgtm

PR #2973, opened 2021-02-05
run: fix check for host pid namespace

giuseppe:
check the pidns is shared with the host only when the pidns mode is
specified.

#### What type of PR is this?
/kind bug

fix running with --userns-uid-map-user and --userns-gid-map-user

#### How to verify it

on my system, all builds created with --userns-uid-map-user failed with:

"not allowed to mix host PID namespace with container user namespace"

```release-note
NONE
```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -2210,7 +2210,7 @@ func checkAndOverrideIsolationOptions(isolation Isolation, options *RunOptions) 
 	case IsolationOCI:
 		pidns := options.NamespaceOptions.Find(string(specs.PIDNamespace))
 		userns := options.NamespaceOptions.Find(string(specs.UserNamespace))
-		if (pidns == nil || pidns.Host) && (userns != nil && !userns.Host) {
+		if (pidns != nil && pidns.Host) && (userns != nil && !userns.Host) {
 			return errors.Errorf("not allowed to mix host PID namespace with container user namespace")
 		}
 	}


TomSweeneyRedHat:
@giuseppe looks like you need a rebase.  Might this fix the gating tests?

rhatdan:
This needs to be back ported to release-1.19 also correct?

edsantiago:
@giuseppe I really need a regression test for this. Can you provide me with an exact command string that fails with buildah prior to your PR, and succeeds after? I've tried all variations I can think of with `--pid=XX` and cannot get this to fail. TIA.

giuseppe:
@edsantiago I've just used `--userns-uid-map-user` as root:

```
$ sudo buildah bud -f Dockerfile -t test --userns-uid-map-user gscrivano --userns-gid-map-group gscrivano --no-cache
STEP 1: FROM alpine
STEP 2: RUN echo hello
error building at STEP "RUN echo hello": not allowed to mix host PID namespace with container user namespace
```

edsantiago:
@giuseppe thank you; I hadn't understood that `bud` was needed, and had only been trying `buildah from`.

With new-buildah, does that succeed for you? It doesn't for me:
```console
$ printf "FROM alpine\nRUN true\n" | sudo ./bin/buildah bud -f - -t test --userns-uid-map-user=esm --userns-gid-map-group=esm prs_should_include_tests
STEP 1: FROM alpine
STEP 2: RUN true
STEP 3: COMMIT test
error committing container for step {Env:[PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin] Command:run Args:[true] Flags:[] Attrs:map[] Message:RUN true Original:RUN true}: error copying layers and metadata for container "977760c9d671fb942d91f1a73a67d8de00a4c23564e0dd71c6b8aa84a4046046": Error initializing source containers-storage:alpine-working-container: error extracting layer "3373e423d4c1c60781b163c832b9728a0ddb347e44744fbbe67c2545ee909a74": lstat /var/lib/containers/storage/overlay/4c99924d42d482c317953aaa9c94615ce9567d8af890f8641bd3efacf54668bf/merged/bin/busybox: operation not permitted
```

giuseppe:
well that is a different one :-)  PR here: https://github.com/containers/storage/pull/815

PR #2987, opened 2021-02-08
[1.19] --iidfile: print hash prefix

vrothberg:
Prefix the image ID with the hash prefix when using `--iidfile` to be
compatible with Docker.  The absence of the hash can cause
docker-compose to error out.

Reported-in: github.com/containers/podman/issues/9260


vrothberg:

TomSweeneyRedHat:
LGTM
thx @vrothberg !

PR #2995, opened 2021-02-11
Don't fail copy to emptydir

rhatdan:
pick f379c5c2 Don't fail copy to emptydir

### DIFF
--- a/add.go
+++ b/add.go
@@ -303,6 +303,13 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 		renameTarget = filepath.Base(extractDirectory)
 		extractDirectory = filepath.Dir(extractDirectory)
 	}
+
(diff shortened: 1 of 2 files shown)


rhatdan:
@TomSweeneyRedHat @nalind @terceiro any idea why these are failing?

TomSweeneyRedHat:
I don't know the why's, but a conformance test is failing a .dockerignore test, which I think is coming from the copy to emptydir PR that I've yet to chase down.

rhatdan:
There are several tests that are failing for this on the main branch now.

Basically dockerignore checks were expecting to fail when the directory exists, but all of the contents of the directory are blocked by .dockerignore.  The change in your PR causes these tests to no longer fail.

rhatdan:
This breaks conformance with Docker, closing.

Issue #2859, opened 2020-12-17
Trouble removing ExposedPorts configuration


brenton:
**Description**
Similar to https://github.com/containers/buildah/issues/1399, I'm struggling to find the proper way to remove incorrect values from ExposedPorts.  I don't doubt I'm doing something wrong but I'm failing to find documentation on how to remove values from ExposedPorts.

**Steps to reproduce the issue:**
1. buildah config --port="3838" $ctr
2. buildah config --port="3839" $ctr
3. buildah config --port="3838-" $ctr
4. buildah config --port="[]" $ctr
 

**Describe the results you received:**
```
    "OCIv1": {
        "config": {
            "ExposedPorts": {
                "3838": {},
                "3839": {},
                "3838-": {},
                "[]": {}
            },
```

**Describe the results you expected:**

    "OCIv1": {
        "config": {
            "ExposedPorts": {
            },

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.18.0-1.fc32.x86_64
```

**Output of `buildah version`:**

```
Version:         1.18.0
Go Version:      go1.14.10
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        0.4.0
libcni Version:  
image Version:   5.8.0
Git Commit:      
Built:           Wed Dec 31 19:00:00 1969
OS/Arch:         linux/amd64

```

**Output of `cat /etc/*release`:**

```
Fedora release 32 (Thirty Two)
```

**Output of `uname -a`:**

```
Linux unknown0050B6B57B28 5.9.11-100.fc32.x86_64 #1 SMP Tue Nov 24 19:16:53 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

```

**Output of `cat /etc/containers/storage.conf`:**

```
[storage]
driver = "overlay"
runroot = "/var/run/containers/storage"
graphroot = "/var/lib/containers/storage"

[storage.options]
additionalimagestores = [
]

[storage.options.overlay]
mountopt = "nodev,metacopy=on"

[storage.options.thinpool]
```

rhatdan:
I don't believe we have a way to remove, would a syntax of leading with a minus sign make sense?buildah config --port="3838" $ctr
buildah config --port="3839" $ctr
buildah config --port="-3839" $ctr

Be what you would expect?

brenton:
[...]

PR #2997, opened 2021-02-11
Add more support for removing config

rhatdan:
Currently we can remove configuration data in buildah config
by adding a trailing "-". This PR adds support for this in --port
calls.  Also added support for clearing all config for a specified
option, if the user specifies "-".

Currently the code blocks setting ANNOTATIONS and LABELS without a
value.  This is broken and should be allowed.

Similarly we were not expaning envioronment variables from the host when
they were not set.

podman run --env foobar
is valid,
so
buildah config --env foobar
should also be valid.

Fixes: https://github.com/containers/buildah/issues/2859

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -190,14 +190,22 @@ func updateConfig(builder *buildah.Builder, c *cobra.Command, iopts configResult
 	}
 	if c.Flag("port").Changed {
 		for _, portSpec := range iopts.ports {
-			builder.SetPort(portSpec)
+			switch {
+			case string(portSpec[0]) == "-":
+				builder.ClearPorts()
+			case strings.HasSuffix(portSpec, "-"):
+				builder.UnsetPort(strings.TrimSuffix(portSpec, "-"))
+			default:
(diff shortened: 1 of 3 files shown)


rhatdan:
@TomSweeneyRedHat Somehow your empty dir PR got in, and now all upstream merges are blocked by the same failure.

Basically we have a dockerignore test that is supposed to fail when a COPY copies nothing from a directory.  Since your PR allows empty directories to be COPIED, this no longer fails, so the test for failure is failing.  
@test "bud with .dockerignore"  in bud.bats

How this got into master, I do not know.  But maybe the test should just be removed.

rhatdan:
@vrothberg @TomSweeneyRedHat @umohnani8 @nalind @saschagrunert @giuseppe PTAL

saschagrunert (approved):
LGTM

PR #2999, opened 2021-02-11
Stop overriding the location of the blob info cache

rhatdan:
When it was first introduced, the blob info cache's location didn't
change from the system-wide default location when we were running in
rootless mode, so we started setting its location ourselves to avoid
triggering permissions errors when updating it.

The image library has since started taking into account that it was
running in rootless mode, but its hardwired default isn't the same as
the one we were setting, so we ended up creating a second cache file.

Stop doing that.

Signed-off-by: Nalin Dahyabhai <nalin@redhat.com>

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/common.go
+++ b/common.go
@@ -56,9 +56,6 @@ func getSystemContext(store storage.Store, defaults *types.SystemContext, signat
 		sc.SignaturePolicyPath = signaturePolicyPath
 	}
 	if store != nil {
-		if sc.BlobInfoCacheDir == "" {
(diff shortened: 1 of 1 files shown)


vrothberg (commented):
Is there a BZ or issue for it?

LGTM

rhatdan:
I think this is related to OpenShift only, so probably did not need to back port, but it does not hurt.
/lgtm

PR #3000, opened 2021-02-12
buildah: use volatile containers

giuseppe:
overlay supports a new mount option "volatile" that reduces I/O by ignoring fsync and syncfs requests.

Enable it for Buildah containers as they are not supposed to survive a machine crash and they are thrown away if the build fails.

#### What type of PR is this?

/kind feature

enable "volatile" containers.  Builds are faster.

```release-note
On kernels where it is supported, enable the "volatile" mount option for overlay
```


rhatdan:
Does storage do the right thing is "volatile" is not supported by the kernel?
LGTM

giuseppe:
> Does storage do the right thing is "volatile" is not supported by the kernel?

if volatile is not supported by the kernel, then it is just ignored

TomSweeneyRedHat (commented):
Think this will need a rebase to pick up the dockerignore reversion.  Also we need to document this somewhere.

giuseppe:
rebased

rhatdan:
This should NOT be back ported to the release.

@vrothberg @saschagrunert @nalind @TomSweeneyRedHat PTAL

rhatdan:
@TomSweeneyRedHat @nalind @vrothberg @saschagrunert @umohnani8 @ashley-cui PTAL

vrothberg (commented):
LGTM, I like simple changes with huge impact :)

PR #3006, opened 2021-02-15
Add multi-arch image build

### LINKED ISSUE #2958: Build multi-arch buildah images
This issue is created to track the discussion about best ways to build multi-arch images for buildah.
So far several CI tools were discussed:
- [Cirrus-CI](https://cirrus-ci.org/ )
  - default CI tool for buildah
  - no hardware available, except x86_64
  - no [docker buildx](https://docs.docker.com/buildx/working-with-buildx/) from the box for hardware emulation in builds

Conclusion: Needs more investigation, probably `buildah bud --arch ..` is a solution, still needs to setup everything from scratch

- [Travis](https://docs.travis-ci.com/)
  - no use in buildah repo 
  - has real hardware to run the builds or tests
  - have limitation in free usage
  - limited to physically available hardware to support(amd64, arm64, ppc64le, s390x)
  - has already multi-arch build setup in skopeo to use as example

Conclusion: depends on community opinion?? if Travis should be back on the table for buildah

- [Github Actions](https://docs.github.com/en/actions)
  - no use in buildah repo 
  - no hardware available, except x86_64
  - has "library" to run multi-arch builds in emulation mode

Conclusion: easy to use, but new CI tool

Long term approach - Cirrus-CI should be the right way to do everything.
Short term approach - probably, Github Actions as easiest way to prepare the code.

Start of discussion - https://github.com/containers/skopeo/pull/1104#issuecomment-771601568

cc @cevich @TomSweeneyRedHat @rhatdan

barthy1:
Add Github Action to do multi-arch build for 4 architectures(`amd64`,
`s390x`, `ppc64le`, `arm64`) from 2 sources - `contrib/buildahimage/stable` and
`contrib/buildahimage/upstream`.
The multi-arch image build is done via emulation (no real hardwarer is
required).
New images are built after each push to master branch and once in a day
on schedule.
- stable image is pushed to `quay.io/buildah/stable:v..` and `quay.io/containers/buildah:v..`
- upstream image is pushed to `quay.io/buildah/upstream:master`

`CONTAINERS_QUAY_USERNAME` and `CONTAINERS_QUAY_PASSWORD`, as well as
`BUILDAH_QUAY_USERNAME` and `BUILDAH_QUAY_PASSWORD` should be created in
advance.

Test run with push to another repos is available at https://github.com/barthy1/buildah/actions/runs/558348033

Signed-off-by: Yulia Gaponenko <yulia.gaponenko1@de.ibm.com>

#### What type of PR is this?

/kind other

Fixes #2958

```release-note
Add multi-arch buildah image build
```

### DIFF
--- a/.github/workflows/multi-arch-build.yaml
+++ b/.github/workflows/multi-arch-build.yaml
@@ -1,0 +1,141 @@
+name: build multi-arch images
+
+on:
+  push:
+    branches: [ master ]
+  schedule:
+  - cron:  '0 8 * * *'
+  # allows to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  multi:
+    name: multi-arch Buildah build
+    env:
+      BUILDAH_QUAY_REGISTRY: quay.io/buildah
+      CONTAINERS_QUAY_REGISTRY: quay.io/containers
+      # list of architectures for build
+      PLATFORMS: linux/amd64,linux/s390x,linux/ppc64le,linux/arm64
+
+    # build several images in parallel(upstream and stable)
+    strategy:
+      matrix:
+        # where the Dockerfiles for builds are located at contrib/buildahimage directory
+        source:
+        - upstream
+        - stable
+    runs-on: ubuntu-latest
+    # internal registry for temporary build before push
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          driver-opts: network=host
+          install: true
+
+      - name: Build and push local Buildah
+        uses: docker/build-push-action@v2
+        with:
+          context: contrib/buildahimage/${{ matrix.source }}
+          file: ./contrib/buildahimage/${{ matrix.source }}/Dockerfile
+          platforms: ${{ env.PLATFORMS }}
+          push: true
+          tags: |
+            localhost:5000/buildah/${{ matrix.source }}
+
+      # Simple verification that container works
+      - name: amd64 container sniff test
+        id: sniff_test
+        run: |
+          VERSION_OUTPUT="$(docker run localhost:5000/buildah/${{ matrix.source }} buildah version)"
+          echo "$VERSION_OUTPUT"
+          echo ::set-output name=version_output::$VERSION_OUTPUT
+
+      # Generate image related info - names, labels
+      - name: Generate image information
+        id: image_info
+        run: |
+          # have special environment variable for image name.
+          # Names are different for upstream and stable Buildah images
+          if [ "${{ matrix.source }}" == "upstream" ]; then
+            # quay.io/buildah/upstream:master
+            echo ::set-output name=buildah_tag::"${{ env.BUILDAH_QUAY_REGISTRY}}/${{ matrix.source }}:master"
+          fi
+
+          if [ "${{ matrix.source }}" == "stable" ]; then
+            VERSION=v"$(echo "${{ steps.sniff_test.outputs.version_output }}" | head -n 1 | awk '{print $2}')"
+            # quay.io/buildah/stable:vX.X.X
+            echo ::set-output name=buildah_tag::"${{ env.BUILDAH_QUAY_REGISTRY}}/${{ matrix.source }}:${VERSION}"
+            # quay.io/contaners/buildah:vX.X.X
+            echo ::set-output name=containers_tag::"${{ env.CONTAINERS_QUAY_REGISTRY}}/buildah:${VERSION}"
+          fi
+
+          # have labels environment variable to use it in the further steps.
+          # Env variable is used to store multi-line value
+          echo "LABELS<<EOF" >> $GITHUB_ENV
+          echo "org.opencontainers.image.source=${{ github.event.repository.html_url }}" >> $GITHUB_ENV
+          echo "org.opencontainers.image.revision=${{ github.sha }}" >> $GITHUB_ENV
+          echo "org.opencontainers.image.created=$(date -u --iso-8601=seconds)" >> $GITHUB_ENV
+          echo "EOF" >> $GITHUB_ENV
+
+      # Separate steps to login and push for buildah and containers quay
+      # repositories are required, because 2 sets of credentials are used and `docker
+      # login` as well as `podman login` do not support having 2 different
+      # credential sets for 1 registry.
+      # At the same time reuse of non-shell steps is not supported by Github Actions via
+      # anchors or composite actions
+
+      # Push to buildah Quay repo stable and upstream Buildah
+      - name: Login to buildah Quay registry
+        uses: docker/login-action@v1
+        with:
+          registry: ${{ env.BUILDAH_QUAY_REGISTRY }}
+          username: ${{ secrets.BUILDAH_QUAY_USERNAME }}
+          password: ${{ secrets.BUILDAH_QUAY_PASSWORD }}
+
+      - name: Push images to buildah Quay
+        uses: docker/build-push-action@v2
+        with:
+          cache-from: type=registry,ref=localhost:5000/buildah/${{ matrix.source }}
+          cache-to: type=inline
+          context: contrib/buildahimage/${{ matrix.source }}
+          file: ./contrib/buildahimage/${{ matrix.source }}/Dockerfile
+          platforms: ${{ env.PLATFORMS }}
+          push: true
+          tags: ${{ steps.image_info.outputs.buildah_tag }}
+          labels: |
+            ${{ env.LABELS }}
+
+      # Push to containers Quay repo only stable Buildah
+      - name: Login to containers Quay registry
+        if: ${{ matrix.source == 'stable' }}
+        uses: docker/login-action@v1
+        with:
+          registry: ${{ env.CONTAINERS_QUAY_REGISTRY}}
+          username: ${{ secrets.CONTAINERS_QUAY_USERNAME }}
+          password: ${{ secrets.CONTAINERS_QUAY_PASSWORD }}
+
+      - name: Push images to containers Quay
+        if: ${{ matrix.source == 'stable' }}
+        uses: docker/build-push-action@v2
+        with:
+          cache-from: type=registry,ref=localhost:5000/buildah/${{ matrix.source }}
+          cache-to: type=inline
+          context: contrib/buildahimage/${{ matrix.source }}
+          file: ./contrib/buildahimage/${{ matrix.source }}/Dockerfile
+          platforms: ${{ env.PLATFORMS }}
+          push: true
+          tags: ${{ steps.image_info.outputs.containers_tag }}
+          labels: |
+            ${{ env.LABELS }}


barthy1:
@cevich Would you be so kind and create [secret env variables](https://docs.github.com/en/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-a-repository) to push images?
`CONTAINERS_QUAY_USERNAME`, `CONTAINERS_QUAY_PASSWORD`, `BUILDAH_QUAY_USERNAME`, `BUILDAH_QUAY_PASSWORD`.

cevich:
> Would you be so kind and create secret env variables to push images?

Yep, happy to.

.github/workflows/multi-arch-build.yaml
```diff
@@ -0,0 +1,119 @@
...
+      - name: amd64 sniff test
+        id: run
+        env:
+          MATRIX_VERSION: ${{ matrix.source }}
+        run: |
+          # image tag is master for upstream source, actual version of buildah binary in the container for stable source
+          VERSION=v"$(docker run localhost:5000/buildah/${MATRIX_VERSION} buildah version | head -n 1 | awk '{print $2}')"
+          if [ "${MATRIX_VERSION}" == "upstream" ]; then
+            VERSION=master
+          fi
```
cevich, line 67: This could be made slightly more reliable (less complex by default) by avoiding always running the docker command in all cases.  Consider this instead:

```bash
          VERSION=master
          if [ "${MATRIX_VERSION}" == "stable" ]; then
              VERSION=v"$(docker run localhost:5000/buildah/${MATRIX_VERSION} buildah version | head -n 1 | awk '{print $2}')"
          fi
```

barthy1: yeah, but then no test that container is functional for upstream case

.github/workflows/multi-arch-build.yaml
```diff
@@ -0,0 +1,119 @@
...
+          push: true
+          tags: |
+            ${{ env.BUILDAH_QUAY_REGISTRY}}/${{ matrix.source }}:${{ steps.run.outputs.version }}
+          labels: |
+            org.opencontainers.image.source=${{ github.event.repository.html_url }}
+            org.opencontainers.image.revision=${{ github.sha }}
+            org.opencontainers.image.created=${{ steps.run.outputs.created }}
+
+      # Push to containers Quay repo only stable Buildah
+      - name: Login to containers Quay registry
```
cevich, line 96: ***RANT**: Oof, it would be nice to be able to use YAML anchor/aliases here (IIRC, *still* not supported by github-actions).

barthy1: no support for anchors so far is also my understanding. And composite actions only for shell steps - https://github.blog/changelog/2020-08-07-github-actions-composite-run-steps/

cevich: Yeah, it's one of the most annoying things about github-actions compared to gitlab CI/CD.  Also, not being able to reuse one common set of steps into multiple workflows. In fact, I'm contemplating it might be worth adding a comment about this.  Just to save somebody an attempt toward deduplicating the config (since that's impossible at the moment).  I think these are different enough it's probably fine as-is.

.github/workflows/multi-arch-build.yaml
cevich: It's probably worth separating those two concerns as distinct steps.  One `buildah sniff test`, another `buildah version capture`. It will make it easier to debug as well, since output isn't potentially obscured/lost in the bash pipeline and stdout capture.  Maybe capture the complete stdout of `buildah version` in a test step, then (as a second step) parse into a new output, e.x. `echo ::set-output name=version::$(echo "${{ step.sniff.outputs.buildah }}" | head -n 1 | awk '{print $2}')`?

.github/workflows/multi-arch-build.yaml
```diff
@@ -0,0 +1,119 @@
...
+        uses: docker/build-push-action@v2
+        with:
+          cache-from: type=registry,ref=localhost:5000/buildah/${{ matrix.source }}
+          cache-to: type=inline
+          context: contrib/buildahimage/${{ matrix.source }}
+          file: ./contrib/buildahimage/${{ matrix.source }}/Dockerfile
+          platforms: ${{ env.PLATFORMS }}
+          push: true
+          tags: |
+            ${{ env.CONTAINERS_QUAY_REGISTRY}}/${{ matrix.source }}:${{ steps.run.outputs.version }}
```
cevich, line 115: This doesn't seem correct, maybe?  There are multiple images at quay.io/containers: "podman", "buildah", and "skopeo".  I think this would push up a "quay.io/containers/stable" which is an incongruity.  Maybe should this be:

`${{ env.CONTAINERS_QUAY_REGISTRY}}/buildah:${{ matrix.source }}`

?

cevich: Oof, nvm I was reading `$CONTAINER_QUAY_REGISTRY` wrongly.

cevich: Err, double-oof, no, I think this is a typo.  It should be `${{ env.CONTAINERS_QUAY_REGISTRY}}/buildah:${{ matrix.source }}` (or similar).

cevich:
> Would you be so kind and create secret env variables to push images?

Okay, this is done.  With the assumption that you intended the `quay.io/containers/buildah` image.

barthy1:
> Okay, this is done. With the assumption that you intended the `quay.io/containers/buildah` image.
Thank you! Yeah, testing image push with other quay repos + having buildah as org name and repo name in different cases confused me :( I will double check all container image names.

cevich:
> I will double check all container image names.

Yeah...not really an easy way to test this out live :confused: 

It would be a pain to setup, but we do have `github.com/containers/automation_sandbox`.  I could fork buildah over there, copy the secrets, and copy this code to see if it works properly on a master branch.

Another idea is...add a temporary commit here to activate the workfow for `on: pull_request`, maybe using a forced `testing` tag on the images.  Either way, it would be convenient to find runtime problems this way rather than after merging.

We need to be a little careful merging this PR anyway, to ensure it doesn't conflict with the cron-jobs @TomSweeneyRedHat has running (to push to these same image names).

.github/workflows/multi-arch-build.yaml
barthy1: there are 2 steps now, one to run the container and get buildah version, another one to extract image tag from it (+ generate labels for 2 push steps)

.github/workflows/multi-arch-build.yaml
barthy1: still added comment about it

.github/workflows/multi-arch-build.yaml
barthy1: fixed names and tags (+ added desired output as comment)

barthy1:
@cevich what do you think just about separate temporary branch in this repo? PR can be merged to it and tested properly.

.github/workflows/multi-arch-build.yaml
```diff
@@ -0,0 +1,141 @@
...
+          tags: |
+            localhost:5000/buildah/${{ matrix.source }}
+
+      # Simple verification that container works
+      - name: amd64 container sniff test
+        id: sniff_test
+        run: |
+          VERSION_OUTPUT="$(docker run localhost:5000/buildah/${{ matrix.source }} buildah version)"
+          echo "$VERSION_OUTPUT"
+          echo ::set-output name=version_output::$VERSION_OUTPUT
```
cevich, line 63: Uh oh, `$VERSION_OUTPUT` will contain multiple lines, so the "set-output" command will only pick up the first line.  IIRC, handling of newlines isn't implemented well (even with proper quoting of the echo arguments).  Dang...sorry I should have anticipated that :disappointed: 

Perhaps this a problem we can simply ignore for now.

.github/workflows/multi-arch-build.yaml
```diff
@@ -0,0 +1,141 @@
...
+          echo "org.opencontainers.image.revision=${{ github.sha }}" >> $GITHUB_ENV
+          echo "org.opencontainers.image.created=$(date -u --iso-8601=seconds)" >> $GITHUB_ENV
+          echo "EOF" >> $GITHUB_ENV
+
+      # Separate steps to login and push for buildah and containers quay
+      # repositories are required, because 2 sets of credentials are used and `docker
+      # login` as well as `podman login` do not support having 2 different
+      # credential sets for 1 registry.
+      # At the same time reuse of non-shell steps is not supported by Github Actions via
+      # anchors or composite actions
```
cevich, line 97: This is helpful.  Thanks.

.github/workflows/multi-arch-build.yaml
```diff
@@ -0,0 +1,141 @@
...
+      - name: Login to containers Quay registry
+        if: ${{ matrix.source == 'stable' }}
+        uses: docker/login-action@v1
+        with:
+          registry: ${{ env.CONTAINERS_QUAY_REGISTRY}}
+          username: ${{ secrets.CONTAINERS_QUAY_USERNAME }}
+          password: ${{ secrets.CONTAINERS_QUAY_PASSWORD }}
+
+      - name: Push images to containers Quay
+        if: ${{ matrix.source == 'stable' }}
```
cevich, line 130: I wonder if might be more clear, if this was:

```
    if: ${{ steps.image_info.outputs.containers_tag }} != ""
```

what do you think?

barthy1: btw, I can see master tag at https://quay.io/repository/containers/buildah?tag=latest&tab=tags . Can we replace the current build setup with multi-arch one? If the source of that build is contrib/buildahimage/upstream, of course.
Then no check is required at all.

cevich:
>  just about separate temporary branch in this repo

That's an excellent idea actually, thanks for the suggestion.  I'll look into making this happen...

.github/workflows/multi-arch-build.yaml
cevich: I think so?  The source of that is a cronjob on @TomSweeneyRedHat laptop :grinning:   I already poked him about this PR, so he's aware.  My understanding is this PR will improve the situation...though Tom will tell us if it's a no-go on the 'master' tag.

cevich:
...okay, permission-granted. I'll get this PR's code over onto a test branch, so we can see it run (with a test tag)...

cevich:
...ugh, sorry I didn't manage to get to this today like I hoped.  It's on my list for first thing tomorrow morning, along with my coffee :smile:

.github/workflows/multi-arch-build.yaml
```diff
@@ -0,0 +1,141 @@
...
+          echo "EOF" >> $GITHUB_ENV
+
+      # Separate steps to login and push for buildah and containers quay
+      # repositories are required, because 2 sets of credentials are used and `docker
+      # login` as well as `podman login` do not support having 2 different
+      # credential sets for 1 registry.
+      # At the same time reuse of non-shell steps is not supported by Github Actions via
+      # anchors or composite actions
+
+      # Push to buildah Quay repo stable and upstream Buildah
```
TomSweeneyRedHat, line 99: do you handle testing too?  The testing branch is used for the latest test version in [bohdi](https://bodhi.fedoraproject.org/updates/?search=buildah).  Sometimes it's the same as stable, but oftentimes not.  Basically the latest beta version.

TomSweeneyRedHat:
LGTM, but my yaml knowledge is extremely limited.....

.github/workflows/multi-arch-build.yaml
barthy1: @TomSweeneyRedHat so far it's only stable and upstream. However I think it's not a problem to add testing as third parallel build. Could you please list from which source to which image repo/tags you would like to see the builds? :)
So far I just replicated skopeo approach with push from stable to `quay.io/buildah/stable:v..` and `quay.io/containers/buildah:v..` and from upstream to `quay.io/buildah/upstream:master`

cevich: Tom, Yulia's github-actions workflow is running a matrix across the different releases, but the source is always "latest master" (at the time of job triggering).  It's not actually differentiating between (conceptually) "stable" and "upstream" (or "testing") in any way.  It simply asks buildah itself for the version number, and tags image accordingly.  In other words, it's not updating what the "latest" tag points to, or behaving differently if it's triggered from a branch-push or a new tag-push.

Is there some machine-detectable buildah (or repo) detail which separates a *test* version from *stable/upstream*?

barthy1: oh, I thought the `testing` question was about location of Dockerfile - https://github.com/containers/buildah/tree/test/contrib/buildahimage/testing .. Sorry for misunderstanding :(

cevich:
@barthy1 okay, I created a testing branch with your commits + 1 that adds a "testing_" prefix to the tags: https://github.com/containers/buildah/tree/temp_automation_test

Unf. github-actions doesn't appear to want to run anything on that branch.  I'll try to investigate "why".

barthy1:
>  Unf. github-actions doesn't appear to want to run anything on that branch. I'll try to investigate "why".

you need to put correct branch name - temp_automation_test here https://github.com/containers/buildah/blob/temp_automation_test/.github/workflows/multi-arch-build.yaml#L5

cevich:
Good catch, that did the trick: https://github.com/containers/buildah/actions/workflows/multi-arch-build.yaml

cevich:
Cool, it seems to be working (ignoring the `testing_` prefix I added):

* [pushing manifest for quay.io/buildah/stable:testing_v1.18.0](https://github.com/containers/buildah/runs/1928674625?check_suite_focus=true#step:10:169
)
* [pushing manifest for quay.io/buildah/upstream:testing_master](https://github.com/containers/buildah/runs/1928674551?check_suite_focus=true#step:10:168)
* [pushing manifest for quay.io/containers/buildah:testing_v1.18.0](https://github.com/containers/buildah/runs/1928674625?check_suite_focus=true#step:12:172)

cevich:
I'm able to inspect the images for the various architectures with command like:
`skopeo --override-arch=ppc64le inspect docker://quay.io/buildah/stable:testing_v1.18.0
| jq .Architecture`

cevich:
Okay, this LGTM then (but it needs a rebase).

barthy1:
Run in native hardware:
```
# docker run quay.io/containers/buildah:testing_v1.18.0 buildah version
Version:         1.18.0
Go Version:      go1.15.2
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        0.4.0
libcni Version:
image Version:   5.8.0
Git Commit:
Built:           Thu Jan  1 00:00:00 1970
OS/Arch:         linux/s390x
```
```
docker run quay.io/buildah/upstream:testing_master buildah version
Version:         1.20.0-dev
Go Version:      go1.15.8
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        0.4.0
libcni Version:  v0.8.1
image Version:   5.10.2
Git Commit:      a6eeca7a
Built:           Thu Feb 18 16:54:01 2021
OS/Arch:         linux/ppc64le
```

.github/workflows/multi-arch-build.yaml
TomSweeneyRedHat: No worries, most likely my bad stealing some late night time for stealth reviewing.  I was referring to the stable/upstream/testing repo's that we have for `quay.io/buildah/*`  and @cevich (as usual) straightened me out with his reply above.

cevich:
Hmmm, Cirrus-CI seems to have gotten "stuck".  I canceled the wedged tasks and restarted them.  The GitHub status should update in 45m or so (when the re-runs complete).  Until then, it will continue showing "X" due to a github bug.

cevich:
@barthy1 I see some merge commits made it into the PR, was that intentional?  Maybe you intended to rebase instead?

barthy1:
@cevich when branch is out-of-date in buildah repo, I can see the "Update branch" button at the bottom of my PR, which actually merges master to the branch. I thought it's allowed way to keep branch up-to-date, taking into account how frequent branch becomes out-of-date.. It seems I was wrong and I will rebase the branch.

rhatdan:
@barthy1 Is this ready to go in?

@cevich @edsantiago @TomSweeneyRedHat PTAL

barthy1:
@rhatdan yes, the code is ready from my point of view

cevich:
Yes, @rhatdan I believe this is good to go, I [tested it on a temporary branch and verified it successfully builds and pushes images.](https://github.com/containers/buildah/pull/3006#issuecomment-781587296).  There are a few scripting nit's but nothing worth holding up this PR.

cevich:
Ref: [Merge triggered this build.](https://github.com/containers/buildah/actions/runs/596536797)

PR #3015, opened 2021-02-16
ADD/COPY: create the destination directory first, chroot to it

nalind:
#### What type of PR is this?

/kind bug

When extracting archives that include symbolic links and paths accessed through those links, we would chroot to the root of the container's filesystem rather than the destination directory.  While we didn't write outside of the container's root directory, this allowed for cases where we put things in different locations than `docker build` would.

#### How to verify it

New conformance tests!
Additions to unit tests!

```
None
```


rhatdan:
LGTM

vrothberg (commented):
LGTM, thank you @nalind

add.go
```diff
@@ -325,13 +325,24 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
...
 	}
 	destUIDMap, destGIDMap := convertRuntimeIDMaps(b.IDMappingOptions.UIDMap, b.IDMappingOptions.GIDMap)
 
+	// Create the target directory if it doesn't exist yet.
+	mkdirOptions := copier.MkdirOptions{
+		UIDMap:   destUIDMap,
+		GIDMap:   destGIDMap,
+		ChownNew: chownDirs,
+	}
+	if err := copier.Mkdir(mountPoint, extractDirectory, mkdirOptions); err != nil {
```
vrothberg, line 341: Would it be possible to move that down into the `copier` package?  `podman cp` would benefit from it, also when copying from a container to the host.

TomSweeneyRedHat:
LGTM, but may need a rebase.
A good candidate for the first update of RHEL 8.4.

add.go
nalind: Not without pretty drastic restructuring - in cases where the destination doesn't exist yet, and the path can include components which are symbolic links, we have to chroot to the rootfs first, then evaluate the path for symbolic links, then create the target directory, and then chroot to the target directory, and trying to stuff all of that into Put() broke just about every other unit test and several integration tests.  In the end, having Add() do both an Eval() and a Mkdir() explicitly looks much easier to test.

nalind:
Added `copier.Eval()` and the `copier.GetOptions.NoDerefSymlinks` flag.

vrothberg:
Can we backport the changes?  I need them in Podman 3.0.

TomSweeneyRedHat:
@vrothberg But not for RHEL 8.4 right?

vrothberg:
@TomSweeneyRedHat, especially for 8.4.  This will help fix the reported `podman cp` bugs.

rhatdan:
Do we have a bugzilla for this?  I have opened a PR to podman to vendor in the latest buildah 1.19.7.

vrothberg:
> Do we have a bugzilla for this? I have opened a PR to podman to vendor in the latest buildah 1.19.7.

Not yet.  There is a number of `podman cp` bugs that we've fixed (and continue fixing).  I prefer to have one BZ that addresses these at once.  Some of these fixes depend on another, so it would very tough to untangle them.

vrothberg:
There are some challenges in Podman since we have to create the destination directory in _some_ cases.  It works well in Buildah but with the remote client in Podman, it's a more tricky.

PR #3031, opened 2021-02-22
pr-should-include-tests: recognized "renamed" tests

edsantiago:
git tries to recognize renamed files. This isn't always
as helpful as intended. Turn it off, so we'll always see
files as 'A'dded.

This is a simple copy of podman PR 9468, but without a
test case because I can find no instances of a .bats
file being renamed in the buildah git history.


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

PR #3032, opened 2021-02-22
Cirrus: Temp. disable prior-fedora (F32) testing

cevich:
In anticipation of F34beta support, preemptively disable testing on
"prior-Fedora". This will allow development to move forward without
a soon-to-be-EOL distro. holding anything back.


vrothberg (commented):
LGTM, @rhatdan PTAL

PR #3033, opened 2021-02-23
Isolation strings, should match user input

rhatdan:
When we parse isolation we expect users to input chroot, oci, rootless.

So when we translate the constants back to strings, we should use the
same values.

These human names need to be passed over the podman-remote build
bindings, so we need to make them match.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -959,7 +959,7 @@ func defaultIsolation() (define.Isolation, error) {
 func IsolationOption(isolation string) (define.Isolation, error) {
 	if isolation != "" {
 		switch strings.ToLower(isolation) {
-		case "oci":
+		case "oci", "default":
 			return define.IsolationOCI, nil
 		case "rootless":
 			return define.IsolationOCIRootless, nil
--- a/define/isolation.go
+++ b/define/isolation.go
@@ -21,14 +21,12 @@ const (
 // String converts a Isolation into a string.
 func (i Isolation) String() string {
 	switch i {
-	case IsolationDefault:
-		return "IsolationDefault"
-	case IsolationOCI:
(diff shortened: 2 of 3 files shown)


vrothberg (commented):
LGTM

pkg/parse/parse.go
```diff
@@ -959,7 +959,7 @@ func defaultIsolation() (define.Isolation, error) {
 func IsolationOption(isolation string) (define.Isolation, error) {
 	if isolation != "" {
 		switch strings.ToLower(isolation) {
-		case "oci":
+		case "oci", "default":
```
TomSweeneyRedHat, line 962: based on your changes above, is "default" possible?

TomSweeneyRedHat:
Lint isn't happy, perhaps due to the "default" check?  The error is not very enlightening.

rhatdan:

rhatdan:
@nalind @TomSweeneyRedHat @umohnani8 @mheon @vrothberg PTAL

pkg/parse/parse.go
lsm5: I wonder the same, but otherwise LGTM.

lsm5 (approved):
LGTM. /cc @ashcrow

lsm5:
> Needs approval from an approver in each of these files:
> 
>     * **[OWNERS](https://github.com/containers/buildah/blob/master/OWNERS)**

my vote doesn't count :| .. oh well..

Issue #3035, opened 2021-02-23
buildah bud: inconsistent io.buildah.version label


edsantiago:
`podman build` will not overwrite an existing `io.buildah.version` label in an image:
```console
$ for i in 20200929 20210223;do printf "FROM quay.io/libpod/testimage:$i\nLABEL foo=bar\n" | ./bin/podman build -t foo:$i -;done
STEP 1: FROM quay.io/libpod/testimage:20200929     <--- this image DOES NOT have an io.buildah.version label
STEP 2: LABEL foo=bar
STEP 3: COMMIT foo:20200929
--> af8d7af49af
af8d7af49afbf91e02a98437168a634d780194a3bb7b426ead4d07de9f7b4c66
STEP 1: FROM quay.io/libpod/testimage:20210223     <--- this image DOES have an io.buildah.version label
STEP 2: LABEL foo=bar
STEP 3: COMMIT foo:20210223
--> d4f925ab5fc
d4f925ab5fc808e1412df67c079232f5bcbd708e6d0ff563a3bfedc0bf39bf68

$ for i in 20200929 20210223;do echo;echo $i;./bin/podman inspect foo:$i|gron|grep -i buildah;done
20200929
json[0].Config.Labels["io.buildah.version"] = "1.19.6";      <--- CORRECT: this is the buildah version in podman
json[0].Labels["io.buildah.version"] = "1.19.6";

20210223
json[0].Config.Labels["io.buildah.version"] = "1.20.0-dev";   <--- INCORRECT: this is the version in the image
json[0].Labels["io.buildah.version"] = "1.20.0-dev";
```
This may be deliberate, but I find it surprising: my expectation is that `io.buildah.version` should be that of the buildah used to build the most recent image.

master @ 1702cbc69 on f33. As mentioned above, the `20200929` testimage does not have an `io.buildah.version` label, the `20210223` one does.

edsantiago:
Fixed by #3036

PR #3036, opened 2021-02-23
Make sure we set the buildah version label

rhatdan:
Currently if the buildah image has a label that was built with
an older version of buildah, we don't update to the latest version.

This PR Will cause the new version to use the current label.

Fixes:https://github.com/containers/buildah/issues/3035

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -1235,7 +1235,7 @@ func (s *StageExecutor) commit(ctx context.Context, createdBy string, emptyLayer
 		s.builder.SetHealthcheck(nil)
 	}
 	s.builder.ClearLabels()
-	s.builder.SetLabel(buildah.BuilderIdentityAnnotation, define.Version)
+
 	for k, v := range config.Labels {
 		s.builder.SetLabel(k, v)
 	}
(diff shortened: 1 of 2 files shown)


rhatdan:
@edsantiago PTAL

edsantiago:
LGTM. I confirmed that this fixes the issue, and also confirmed that your new test is correct: the command-line `--label` reproduces the original problem on buildah @ current master. Thanks for addressing this so quickly!

TomSweeneyRedHat:
While you're in here and it looks like you're going to need to adjust this to fix a test, can you update the comments in buildah.go and contrib/rpm/bulidah.spec please?  The line before the version line in each file has a comment that acts as a reminder to fix the other file.  Please update (don't remove) those with the correct files now.

edsantiago:
@TomSweeneyRedHat did you accidentally press the `Close` button, or was that intentional?

edsantiago:
Judgment call: I'm going to assume that was an accident, and I will reopen so we can see CI results.

TomSweeneyRedHat:
@edsantiago good judgment, I did accidentally close it FWIW.  TY for reopening.

rhatdan:
@TomSweeneyRedHat I actually did not change the version of buildah here. This PR is just changing the place where buildah sets a label in an image with the current version of buildah.

rhatdan:

vrothberg (commented):
LGTM

PR #3041, opened 2021-02-25
Shrink the vendoring size of pkg/cli

rhatdan:
This PR removes the pkg/auth which brings in docker/docker
since it really is not needed, and was only there to help users
discover the settings of where the authfile was, when the environment
variables were set.  Would almost never be of any value.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```


vrothberg:
@nalind fixed the issue in c/image and I just cut https://github.com/containers/image/releases/tag/v5.10.5

rhatdan:
@vrothberg PTAL

ashley-cui (commented):
LGTM

define/build.go
```diff
@@ -61,3 +70,154 @@ type CommonBuildOptions struct {
...
+	// Jobs is the number of stages to run in parallel.  If not specified it defaults to 1.
+	Jobs *int
+	// LogRusage logs resource usage for each step.
+	LogRusage bool
+	// Excludes is a list of excludes to be used instead of the .dockerignore file.
+	Excludes []string
+	// From is the image name to use to replace the value specified in the first
+	// FROM instruction in the Containerfile
+	From string
+}
```
TomSweeneyRedHat, line 223: My OCD is twitching due to a lack of alphabetization.

pkg/cli/common.go
```diff
@@ -177,7 +175,7 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
 	fs := pflag.FlagSet{}
 	fs.String("arch", runtime.GOARCH, "set the ARCH of the image to the provided value instead of the architecture of the host")
 	fs.StringArrayVar(&flags.Annotation, "annotation", []string{}, "Set metadata for an image (default [])")
-	fs.StringVar(&flags.Authfile, "authfile", auth.GetDefaultAuthFile(), "path of the authentication file.")
+	fs.StringVar(&flags.Authfile, "authfile", "", "path of the authentication file.")
```
TomSweeneyRedHat, line 178: Did you intend to make this "" rather than getting the DefaultAuthFile?

TomSweeneyRedHat:
One question on the defaultauthfile, otherwise LGTM

pkg/cli/common.go
rhatdan: Yes, this call was screwing it up and did not work on certain platforms.

Covered in PR description. 
```
This PR removes the pkg/auth which brings in docker/docker
since it really is not needed, and was only there to help users
discover the settings of where the authfile was, when the environment
variables were set. Would almost never be of any value.
```

PR #3043, opened 2021-02-27
Clarify userns options are usable only as root in man pages

TomSweeneyRedHat:
We've had a few BZ's come in recently where users were confused about
using the `--userns*` options for build and from inside of a rootless
container.  It's not allowed, but it's not documented that way.  This
change adds further documenation for them.  This will satisfy this
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1931261

Once merged, similar changes are necessary in the `podman build`
man pages.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -571,6 +571,8 @@ suitable user name to use as the default setting for this option.
 
 Users can specify the maps directly using `--userns-uid-map` described in the buildah(1) man page.
 
+**NOTE:** When this option is specified by a rootless user, the specified mappings are relative to the rootless usernamespace in the container, rather than being relative to the host as it would be when run rootlful.
+
 **--userns-gid-map-group** *group*
 
 Specifies that a GID mapping which should be used to set ownership, at the
@@ -583,6 +585,8 @@ specified, `buildah` will assume that the specified user name is also a
 suitable group name to use as the default setting for this option.
 
 Users can specify the maps directly using `--userns-gid-map` described in the buildah(1) man page.
+
+**NOTE:** When this option is specified by a rootless user, the specified mappings are relative to the rootless usernamespace in the container, rather than being relative to the host as it would be when run rootlful.
 
 **--uts** *how*
 
--- a/docs/buildah-from.md
+++ b/docs/buildah-from.md
@@ -397,6 +397,8 @@ If none of --userns-uid-map-user, --userns-gid-map-group, or --userns-uid-map
 are specified, but --userns-gid-map is specified, the UID map will be set to
 use the same numeric values as the GID map.
 
+**NOTE:** When this option is specified by a rootless user, the specified mappings are relative to the rootless usernamespace in the container, rather than being relative to the host as it would be when run rootlful.
+
 **--userns-gid-map-group** *mapping*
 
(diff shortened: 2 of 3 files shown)


TomSweeneyRedHat:
@giuseppe PTAL

docs/buildah-from.md
```diff
@@ -368,6 +368,8 @@ include:
 
 **--userns** *how*
 
+**NOTE:** This option is not usable when running in rootless mode.  This is only usable when running as root.
```
giuseppe, line 371: it can be used as rootless

docs/buildah.md
```diff
@@ -69,6 +69,8 @@ Storage driver option, Default storage driver options are configured in /etc/con
 
 **--userns-uid-map** *mapping*
 
+**NOTE:** This option is not usable when running in rootless mode.  This is only usable when running as root.
```
giuseppe, line 72: it can be used as rootless

docs/buildah.md
```diff
@@ -90,6 +92,8 @@ use the same numeric values as the GID map.
 
 **--userns-gid-map** *mapping*
 
+**NOTE:** This option is not usable when running in rootless mode.  This is only usable when running as root.
```
giuseppe, line 95: it can be used as rootless

docs/buildah.md
giuseppe: The difference with root (as well for `--userns-gid-map`) is that the specified mappings are relative to the rootless user namespace instead of being relative to the host

giuseppe (changes requested):

docs/buildah.md
TomSweeneyRedHat: How about:
```
**NOTE:** When this option is specified by a rootless user, the specified mappings are relative to the rootless usernamespace in the container, rather than being relative to the host as it would be when run by a rootful user.
```

@rhatdan @giuseppe  and/or @vrothberg ?

rhatdan: rootfull. And then LGTM

docs/buildah-bud.md
```diff
@@ -551,6 +551,8 @@ include:
 
 **--userns** *how*
 
+**NOTE:** This option is not usable when running in rootless mode.  This is only usable when running as root.
+
```
giuseppe, line 555: this is not correct.  `--userns` can be used as rootless

docs/buildah-from.md
giuseppe: Please drop this line

docs/buildah.md
giuseppe: same here, please use the same sentence as above for `--userns-uid-map`

TomSweeneyRedHat:
@rhatdan @giuseppe PTAL, I've removed the line from the `--userns` option and reworded per suggestions.

rhatdan:
LGTM

giuseppe (approved):
/lgtm

PR #3049, opened 2021-03-02
Check local image to match system context

rhatdan:
Currently if you pull one image and then pullifmissing with
a different --arch, Buildah does not pull the different arch,
even though the arch is missing.

This PR checks the existing image to see if it matches the arch, os
and variant of the specified image, before using the local image.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -2680,3 +2680,24 @@ _EOF
   run_buildah images
   expect_output --substring image-a
 }
+
+@test "bud --pull=false --arch test" {
+  mytmpdir=${TESTDIR}/my-dir
+  mkdir -p ${mytmpdir}
+cat > $mytmpdir/Containerfile << _EOF
+FROM registry.access.redhat.com/ubi8-minimal
+_EOF
+  run_buildah bud -f Containerfile --pull=false -q --arch=amd64 -t image-amd --signature-policy ${TESTSDIR}/policy.json ${mytmpdir}
+  run_buildah inspect --format '{{ index .Docker.Config.Labels "architecture" }}' image-amd
+  expect_output --substring x86_64
+
+  run_buildah bud -f Containerfile --pull=false -q --arch=arm64 -t image-arm --signature-policy ${TESTSDIR}/policy.json ${mytmpdir}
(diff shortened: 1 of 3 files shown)


vrothberg (commented):
LGTM

Looking forward to have everything in _one_ image library.

tests/bud.bats
```diff
@@ -2680,3 +2680,18 @@ _EOF
...
+@test "bud --pull=false --arch test" {
+  mytmpdir=${TESTDIR}/my-dir
+  mkdir -p ${mytmpdir}
+cat > $mytmpdir/Containerfile << _EOF
+FROM registry.access.redhat.com/ubi8-minimal
+_EOF
+  run_buildah bud -f Containerfile --pull=false -q --arch=amd64 -t image-amd --signature-policy ${TESTSDIR}/policy.json ${mytmpdir}
+  run_buildah inspect --format '{{ index .Docker.Config.Labels "architecture" }}' image-amd
+  expect_output --substring x86_64
+
```
TomSweeneyRedHat, line 2693: Please add
```
run_buildah inspect --format '{{ index .OCIv1.Architecture  }}'  image-amd
expect_output --substring amd64
```

rhatdan: That field means nothing, since even when it pulls or uses the wrong image it is set correctly.  Basically that field records the value passed in.  The ubi images happen to have a label added with architecture, which is the only way I can tell if we have the right arch.

rhatdan: Really in this case we would want to check the arch of the "from" image.

rhatdan: This value is wrong with the failure we are trying to fix. The image is currently just recording the value passed in, even though the image architecture is the previously pulled image.
  
In this case we would want to check the arch of the "from" image.  I enhanced the test to check for this value on the from image, which is what we care about.

I also added a couple of pull --policy missing tests to make sure if you change the --arch, we get a new image.

TomSweeneyRedHat:
Thanks @rhatdan .
LGTM
once tests get happy

PR #3052, opened 2021-03-03
[release-1.19] ADD/COPY: create the destination directory first, chroot to it

nalind:
#### What type of PR is this?

/kind bug

When extracting archives that include symbolic links and paths accessed through those links, we would chroot to the root of the container's filesystem rather than the destination directory.  While we didn't write outside of the container's root directory, this allowed for cases where we put things in different locations than `docker build` would.

#### How to verify it

New conformance tests!
Additions to unit tests!

Cherry-picked from #3015.

```
None
```


vrothberg (commented):
LGTM

PR #3050, opened 2021-03-02
[ci:docs] Add information about multi-arch images to the Readme

barthy1:
#### What type of PR is this?

/kind documentation

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #3059, opened 2021-03-04
Update multi-arch buildah build setup with new logic

barthy1:
- push `quay.io/buildah/stable:vX.X.X` and `quay.io/contaners/buildah:vX.X.X`
only one time for each `X.X.X` version
- build verification is still on the table, just no push
- replace `master` to `main` tag for `quay.io/buildah/upstream:main`

#### What type of PR is this?

/kind other

```release-note
Use main tag for quay.io/buildah/upstream container image. 
Push is done only one time for each buildah version of quay.io/buildah/stable:vX.X.X and quay.io/contaners/buildah:vX.X.X
```

### DIFF
--- a/.github/workflows/multi-arch-build.yaml
+++ b/.github/workflows/multi-arch-build.yaml
@@ -62,26 +62,37 @@ jobs:
           echo "$VERSION_OUTPUT"
           echo ::set-output name=version_output::$VERSION_OUTPUT
 
-      # Generate image related info - names, labels
+      # Generate image related info - names, labels, check whether to push
       - name: Generate image information
         id: image_info
         run: |
-          # have special environment variable for image name.
+          # Have special environment variable for image name.
           # Names are different for upstream and stable Buildah images
           if [ "${{ matrix.source }}" == "upstream" ]; then
-            # quay.io/buildah/upstream:master
-            echo ::set-output name=buildah_tag::"${{ env.BUILDAH_QUAY_REGISTRY}}/${{ matrix.source }}:master"
+            # quay.io/buildah/upstream:main
+            # Always push upstream image to buildah repository
+            echo ::set-output name=buildah_push::'true'
+            # Never push upstream image to containers repository
+            echo ::set-output name=containers_push::'false'
+            echo ::set-output name=buildah_tag::"${{ env.BUILDAH_QUAY_REGISTRY}}/${{ matrix.source }}:main"
           fi
 
           if [ "${{ matrix.source }}" == "stable" ]; then
             VERSION=v"$(echo "${{ steps.sniff_test.outputs.version_output }}" | head -n 1 | awk '{print $2}')"
             # quay.io/buildah/stable:vX.X.X
+            # Check if stable image with current version already exists and make decision if new one needs to be pushed
+            PUSH=$(docker pull ${{ env.BUILDAH_QUAY_REGISTRY}}/${{ matrix.source }}:${VERSION} &>/dev/null && echo 'false' || echo 'true')
+            echo ::set-output name=buildah_push::${PUSH}
             echo ::set-output name=buildah_tag::"${{ env.BUILDAH_QUAY_REGISTRY}}/${{ matrix.source }}:${VERSION}"
+
             # quay.io/contaners/buildah:vX.X.X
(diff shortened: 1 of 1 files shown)


barthy1:
@TomSweeneyRedHat as follow-up after our discussion https://github.com/containers/buildah/pull/3050#discussion_r586819812

TomSweeneyRedHat:
@barthy1 looks good on a first pass, but I've some tired eyes now and want to look again tomorrow.  Thanks so much for getting this together quickly.

rhatdan:
@cevich @edsantiago PTAL

.github/workflows/multi-arch-build.yaml
```diff
@@ -62,26 +62,37 @@ jobs:
...
-      # Generate image related info - names, labels
+      # Generate image related info - names, labels, check whether to push
       - name: Generate image information
         id: image_info
         run: |
-          # have special environment variable for image name.
+          # Have special environment variable for image name.
           # Names are different for upstream and stable Buildah images
           if [ "${{ matrix.source }}" == "upstream" ]; then
             # quay.io/buildah/upstream:master
```
TomSweeneyRedHat, line 72: Please remove master and change to "main" throughout, unless @rhatdan has a better word.

.github/workflows/multi-arch-build.yaml
```diff
@@ -62,26 +62,37 @@ jobs:
...
-            echo ::set-output name=buildah_tag::"${{ env.BUILDAH_QUAY_REGISTRY}}/${{ matrix.source }}:master"
+            # Always push upstream image to buildah repository
+            echo ::set-output name=buildah_push::'true'
+            # Never push upstream image to containers repository
```
TomSweeneyRedHat, line 75: I'm confused by the comment and why you wouldn't want to push?

.github/workflows/multi-arch-build.yaml
```diff
@@ -62,26 +62,37 @@ jobs:
...
+            echo ::set-output name=containers_push::'false'
+            echo ::set-output name=buildah_tag::"${{ env.BUILDAH_QUAY_REGISTRY}}/${{ matrix.source }}:main"
           fi
 
           if [ "${{ matrix.source }}" == "stable" ]; then
             VERSION=v"$(echo "${{ steps.sniff_test.outputs.version_output }}" | head -n 1 | awk '{print $2}')"
             # quay.io/buildah/stable:vX.X.X
+            # Check if stable image with current version already exists and make decision if new one needs to be pushed
+            PUSH=$(docker pull ${{ env.BUILDAH_QUAY_REGISTRY}}/${{ matrix.source }}:${VERSION} &>/dev/null && echo 'false' || echo 'true')
```
TomSweeneyRedHat, line 84: I think not, but any chance to use Podman or Buildah here instead of Docker?  @cevich?

TomSweeneyRedHat (commented):
Some questions @barthy1 .  However, I really like your solution for testing the existence of the versioned image and then pushing if not present.  Very snappy!  :1st_place_medal:

.github/workflows/multi-arch-build.yaml
barthy1: yep, sorry, forgot to update the comment, will do.

.github/workflows/multi-arch-build.yaml
barthy1: Based on initial list of multi-arch images, for upstream case image should be pushed to `quay.io/buildah/upstream:main`, but not to `quay.io/containers/buildah:master` (or main)

.github/workflows/multi-arch-build.yaml
barthy1: I guess `docker` can be switched to `podman` here, however I don't see much advantage in this replacement. If there is more elegant way to check existence of remote container image with `skopeo` or `podman`, I'd be happy to use it, but I need your input here as experts :)

cevich: You should be able to use `skopeo list-tags docker://${{ env.BUILDAH_QUAY_REGISTRY}}/${{ matrix.source }} | jq -r .Tags[] | grep -q "$VERSION."` for this, it will be _MUCH_ faster as well.

PR #3062, opened 2021-03-05
Make PolicyMap and PullPolicy names align

rhatdan:
Currently it is impossible to switch from a PullPolicy type
and PolicyMap via strings.  This PR Makes the types align.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/define/pull.go
+++ b/define/pull.go
@@ -25,24 +25,19 @@ const (
 	// take, signalling that the source image should not be pulled from a
 	// registry.
 	PullNever
-
-	// OCI used to define the "oci" image format
-	OCI = "oci"
-	// DOCKER used to define the "docker" image format
-	DOCKER = "docker"
 )
 
 // String converts a PullPolicy into a string.
 func (p PullPolicy) String() string {
 	switch p {
 	case PullIfMissing:
-		return "PullIfMissing"
+		return "missing"
 	case PullAlways:
-		return "PullAlways"
+		return "always"
 	case PullIfNewer:
-		return "PullIfNewer"
+		return "ifnewer"
 	case PullNever:
-		return "PullNever"
+		return "never"
 	}
 	return fmt.Sprintf("unrecognized policy %d", p)
 }
--- a/define/pull_test.go
+++ b/define/pull_test.go
@@ -1,0 +1,13 @@
+package define
+
+import (
+	"testing"
+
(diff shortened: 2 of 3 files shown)


rhatdan:
This PR is needed to make passing of PullPolicy in podman-remote work.

ashley-cui (commented):
LGTM

define/pull.go
```diff
@@ -25,24 +25,19 @@ const (
...
-	// DOCKER used to define the "docker" image format
-	DOCKER = "docker"
 )
 
 // String converts a PullPolicy into a string.
 func (p PullPolicy) String() string {
 	switch p {
 	case PullIfMissing:
-		return "PullIfMissing"
+		return "missing"
```
nalind, line 34: Should this be "ifmissing"?

rhatdan: No these names need to match the PolicyMap defined below.

```
// String converts a PullPolicy into a string.
func (p PullPolicy) String() string {
	switch p {
	case PullIfMissing:
		return "missing"
	case PullAlways:
		return "always"
	case PullIfNewer:
		return "ifnewer"
	case PullNever:
		return "never"
	}
}
```

nalind: Why is that?

nalind: Specifically, why aren't we correcting the entry in the PolicyMap?

rhatdan: The policy map is translating User Input into the constants.

```
man buildah pull
...
       --policy=always|missing|never

       Pull image policy. The default is missing.

              • missing:  attempt  to pull the latest image from the registries listed in registries.conf if a local image does not ex‐
                ist. Raise an error if the image is not in any listed registry and is not present locally.

              • always: Pull the image from the first registry it is found in as listed in  registries.conf.  Raise  an  error  if  not
                found in the registries, even if the image is present locally.

              • never:  do not pull the image from the registry, use only the local version. Raise an error if the image is not present
                locally.
```

rhatdan: I want to pass the same names over the from podman bindings -> podman libpod handler.

nalind: [shrugs]

PR #3066, opened 2021-03-08
Remove duplicated code in addcopy

jmguzik:
#### What type of PR is this?

/kind cleanup
Add and copy have the same description. I unified them
#### How to verify it
buildah copy and add help

None

```release-note
NONE

```

### DIFF
--- a/cmd/buildah/addcopy.go
+++ b/cmd/buildah/addcopy.go
@@ -9,6 +9,7 @@ import (
 	buildahcli "github.com/containers/buildah/pkg/cli"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
 type addCopyResults struct {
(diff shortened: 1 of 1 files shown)


rhatdan:
Idea is good, but the tests are not happy.

jmguzik:
@rhatdan I am not sure if fails are related to my code. I will revert this commit and have in fact master to recheck it.

jmguzik:
In the end, it seems it is problem with my code.
/hold

rhatdan:
You were on the right track, don't know why the failures were happening.

jmguzik:
I will investigate this in my free time.

jmguzik:
@rhatdan it works now. Ptr was the problem.

cmd/buildah/addcopy.go
```diff
@@ -20,63 +21,56 @@ type addCopyResults struct {
 	contextdir string
 }
 
+func createCommand(addCopy string, desc string, short string, opts *addCopyResults) *cobra.Command {
+	return &cobra.Command{
+		Use:   addCopy,
+		Short: short,
+		Long:  desc,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return addAndCopyCmd(cmd, args, strings.ToUpper(addCopy), *opts)
```
TomSweeneyRedHat, line 30: Doesn't this still need a boolean true before *opts ala:
```suggestion
			return addAndCopyCmd(cmd, args, strings.ToUpper(addCopy), true, *opts)
```

TomSweeneyRedHat: Oh wait, I see you did away with the boolean, but it looks like you're always doing an ADD and I think you still need to differentiate and do an ADD or COPY.

jmguzik: It is adressed later in the code. Please see the line 124:
extractLocalArchives := verb == "ADD"

cmd/buildah/addcopy.go
```diff
@@ -127,6 +121,7 @@ func addAndCopyCmd(c *cobra.Command, args []string, verb string, extractLocalArc
 		options.Excludes = excludes
 	}
 
+	extractLocalArchives := verb == "ADD"
```
jmguzik, line 124: here we have this check whether we are adding or copying

cmd/buildah/addcopy.go
TomSweeneyRedHat: Oh, that's subtle.  I didn't catch the createCommand() call setting verb to either ADD or COPY above.

cmd/buildah/addcopy.go
```diff
@@ -20,63 +21,56 @@ type addCopyResults struct {
...
-		Short: "Add content to the container",
-		Long:  addDescription,
-		RunE: func(cmd *cobra.Command, args []string) error {
-			return addCmd(cmd, args, addOpts)
-		},
-		Example: `buildah add containerID '/myapp/app.conf'
-  buildah add containerID '/myapp/app.conf' '/myapp/app.conf'`,
-		Args: cobra.MinimumNArgs(1),
-	}
+	addCommand := createCommand("add", addDescription, shortAdd, &addOpts)
```
TomSweeneyRedHat, line 57: Suggestion only, I think it would be more clear if you did:
```suggestion
	addCommand := createCommand("ADD", addDescription, shortAdd, &addOpts)
```
It would stand out for me more when I searched for "ADD" from line 124.   Ditto uppercasing "copy" to "COPY" at line 60.  I'd still keep the ToUpper() call in play above in case someone forgets to uppercase stuff in further development.

TomSweeneyRedHat:
One strong suggestion for casing to clear things up for late-night brains like mine, but otherwise a nice change, thanks @jmguzik !

cmd/buildah/addcopy.go
jmguzik: I have to protest this one :) It will force 3x toLower in the createCommand which is less efficient than 1x toUpper.

jmguzik: Line 26, 32 and 33 use lower

jmguzik:
> One strong suggestion for casing to clear things up for late-night brains like mine, but otherwise a nice change, thanks @jmguzik !

 It will force 3x toLower in the createCommand vs 1x toUpper which is now. Please let me know if you really want such change :)

cmd/buildah/addcopy.go
rhatdan: Yes that is what I mean

jmguzik: Lint is not happy. They are starting from a capital letter and they should have a comment as they will be treated as exported.
I can do this, but honestly don't see a point.
I would recommend keeping current solution.
Let me know what you think.

TomSweeneyRedHat: I can live with it as is.  However, as I was reviewing, I saw your snazzy check for "ADD" and couldn't find "ADD" being passed anywhere, it took me 10 minutes or so (again late night brain) to figure it out.  The suggestion was to try to make the code a bit clearer, but not completely necessary, especially given the lint fun you've since run into.

Thanks for the consideration though.

TomSweeneyRedHat:
LGTM

rhatdan:
Thanks @jmguzik 
/lgtm

cmd/buildah/addcopy.go
jmguzik: Thanks for the comment. I personally think this check is sneaky, not snazzy, but in my opinion acceptable. Especially that it is a minor side effect of code unification.

Issue #1589, opened 2019-05-11
invalid tar header: unknown


marshallford:
GKE cluster (`1.12.7-gke.10`) using the node image `cos_containerd` is failing to run containers built with buildah and pushed to GCR.

```
Failed to pull image "gcr.io/gke-clusters/testing:latest": rpc error: code = Unknown desc = failed to pull and unpack image "gcr.io/gke-clusters/testing:latest": failed to unpack image on snapshotter overlayfs: failed to extract layer sha256:f1b5933fe4b5f49bbe8258745cf396afe07e625bdab3168e364daf7c956b6b81: mount callback failed on /var/lib/containerd/tmpmounts/containerd-mount799987480: archive/tar: invalid tar header: unknown
```

```
➜ cat Dockerfile 
FROM alpine
RUN date

Projects/cluster1/test on 🐳 v18.09.5 
➜ buildah version
Version:         1.8.2
Go Version:      go1.12.4
Image Spec:      1.0.0
Runtime Spec:    1.0.0
CNI Spec:        0.4.0
libcni Version:  v0.7.0-rc2
Git Commit:      e23314b1
Built:           Fri May 10 09:23:56 2019
OS/Arch:         linux/amd64

Projects/cluster1/test on 🐳 v18.09.5 
➜ cat test.yaml 
apiVersion: v1
kind: Pod
metadata:
  name: test
  namespace: default
spec:
  containers:
  - name: test
    image: gcr.io/gke-clusters/testing:latest
```

**Steps to reproduce the issue:**
1. Build an image using the Dockerfile above with buildah
2. Push to GCR (or maybe any other container registry?)
3. `kubectl apply -f test.yaml` on k8s cluster
4. Use `kubectl describe` to view warning event

**Describe the results you received:**

Pod fails to spin up.

**Describe the results you expected:**

I'd expect the pod to pull and run the container.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
➜ yay -s buildah
2 aur/buildah-git r1330.391a5bea-1 (+1 0.00%) 
    A tool which facilitates building OCI images
1 community/buildah 1.8.2-1 (5.4 MiB 23.6 MiB) (Installed)
    A tool which facilitates building OCI images
==> Packages to install (eg: 1 2 3, 1-3 or ^4)
==> ^C
```

**Output of `buildah version`:**

```
Version:         1.8.2
Go Version:      go1.12.4
Image Spec:      1.0.0
Runtime Spec:    1.0.0
[...]

PR #3072, opened 2021-03-10
[v1.19] vendor containers/image@v5.10.5

vrothberg:
Fixes are needed to fix the remaining issues of #1589.

Do we need a BZ to get that into RHEL 8.4?  Fairly annoying bug (see https://github.com/containers/buildah/issues/1589).


vrothberg:
Need the same for Podman @mheon @baude

TomSweeneyRedHat:
We'll need a BZ, but I think we need another total version dance and may just do the one for that which would include this fix too.

vrothberg:
> We'll need a BZ, but I think we need another total version dance and may just do the one for that which would include this fix too.

Perfect. Let me know in case the dance doesn't make it into 8.4. If not, I'll open a BZ.

Issue #3074, opened 2021-03-11
Generated image altered when using --squash option


nlach-tehtris:
**Description**

Squashing the layers into one during build seems to fail when there is a socket present on the filesystem, resulting in a corrupted image generated (missing folders and binaries)

**Steps to reproduce the issue:**

1. Use the following dockerfile for building the image

```
FROM docker.io/library/ubuntu:18.04

RUN apt-get update \
    ; apt-get install -y netcat-openbsd \
    ; nc -lkU /test.socket & sleep 1

RUN ls /
```

2. Build image using _squash_ parameter
3. Compare the result of "__ls /__" command during image building and after image building

**Describe the results you received:**

During image building, the ls command output is :

```
STEP 3: RUN ls /
bin  boot  dev  etc  home  lib  lib64  media  mnt  opt  proc  root  run  sbin  srv  sys  test.socket  tmp  usr  var
```

We now have 2 images, __docker.io/library/ubuntu:18.04__ and __localhost/test:latest__ with a strange size difference (new image is smaller when adding new packages ?)

```
user@local $ buildah images
REPOSITORY                   TAG      IMAGE ID       CREATED         SIZE
localhost/test               latest   cda0f11ca4de   6 minutes ago   21.6 MB
docker.io/library/ubuntu     18.04    329ed837d508   7 days ago      65.6 MB
```

When running "__ls /__" inside our new image, we obtain a different result than during building

```
user@local $ podman run -it --rm --entrypoint '["/bin/ls", "/"]' test:latest
bin  boot  dev	etc  home  lib	lib64  media  mnt  opt	proc  root  run  sbin  srv  sys
```

**Describe the results you expected:**

Having the filesystem not altered after squashing all layers into one

Same problem seems to happend with `podman image build` command

**Output of `rpm -q buildah` or `apt list buildah`:**

Buildah binary compiled from source, @tag/v1.19.6

**Output of `buildah version`:**

```
Version:         1.19.6
Go Version:      go1.13.4
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        0.4.0
libcni Version:  v0.8.1
image Version:   5.10.2
Git Commit:      7aedb164
Built:           Mon Mar  1 14:58:58 2021
OS/Arch:         linux/amd64
```

**Output of `podman version` if reporting a `podman build` issue:**

```
Version:      3.0.1
API Version:  3.0.0
Go Version:   go1.13.4
[...]

PR #3075, opened 2021-03-12
copier: ignore sockets

giuseppe:
#### What type of PR is this?

/kind bug

fix image squashing that have sockets

#### How to verify it

follow the reproducer here: https://github.com/containers/buildah/issues/3074

Fixes: #3074

```release-note
fix squashing of images that contain sockets
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -1128,6 +1128,10 @@ func copierHandlerGet(bulkWriter io.Writer, req request, pm *fileutils.PatternMa
 						}
 						return errors.Wrapf(err, "copier: get: error reading %q", path)
 					}
+					if info.Mode()&os.ModeType == os.ModeSocket {
+						logrus.Warningf("buildah/copier: skipping socket %q", info.Name())
+						return nil
+					}
 					// compute the path of this item
 					// relative to the top-level directory,
 					// for the tar header
--- a/image.go
+++ b/image.go
@@ -146,14 +146,16 @@ func computeLayerMIMEType(what string, layerCompression archive.Compression) (om
 }
 
 // Extract the container's whole filesystem as if it were a single layer.
-func (i *containerImageRef) extractRootfs() (io.ReadCloser, error) {
+func (i *containerImageRef) extractRootfs() (io.ReadCloser, chan error, error) {
 	var uidMap, gidMap []idtools.IDMap
 	mountPoint, err := i.store.Mount(i.containerID, i.mountLabel)
 	if err != nil {
-		return nil, errors.Wrapf(err, "error mounting container %q", i.containerID)
+		return nil, nil, errors.Wrapf(err, "error mounting container %q", i.containerID)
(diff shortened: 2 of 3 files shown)


rhatdan:
LGTM
@TomSweeneyRedHat @nalind @vrothberg @containers/podman-maintainers PTAL

copier/copier.go
```diff
@@ -1149,6 +1149,10 @@ func copierHandlerGet(bulkWriter io.Writer, req request, pm *fileutils.PatternMa
 						}
 						symlinkTarget = target
 					}
+					if info.Mode()&os.ModeType == os.ModeSocket {
+						logrus.Warningf("Skipping socket %q", info.Name())
```
vrothberg, line 1153: Can we use `Debugf` instead?  I am afraid this could cause some noise.

rhatdan: I believe it should stay a warning. Users should be told that sockets will not be recorded within the image.  Users should actually only put sockets in tmpfs but that is a different story.  I don't think this is a common pattern.

rhatdan: If it becomes noisy we can always revisit the situation.

vrothberg: Sounds good to me.

It may be helpful to add more information about the SLOC to the log:
```golang
logrus.Warningf("buildah/copier: skipping socket %q", info.Name())
```
?

rhatdan: Sure.

giuseppe: switched to `Debugf`

copier/copier.go
```diff
@@ -1164,6 +1164,10 @@ func copierHandlerGet(bulkWriter io.Writer, req request, pm *fileutils.PatternMa
 						}
 						symlinkTarget = target
 					}
+					if info.Mode()&os.ModeType == os.ModeSocket {
+						logrus.Debugf("Skipping socket %q", info.Name())
```
TomSweeneyRedHat, line 1168: I thought @rhatdan and @vrothberg had settled on a warning after all?

copier/copier.go
```diff
@@ -1164,6 +1164,10 @@ func copierHandlerGet(bulkWriter io.Writer, req request, pm *fileutils.PatternMa
...
+						return nil
+					}
```
TomSweeneyRedHat, line 1170: Can/should we move this if check higher up to avoid some of the other processing above this?

copier/copier.go
giuseppe: Thanks.  The github page showed only the first comment, reverted back to a warning

copier/copier.go
giuseppe: sure, moved

TomSweeneyRedHat:
LGTM
as usual, nice work @giuseppe !

giuseppe:
restarted "Containerized Integration".  It seems a flake

Issue #2249, opened 2020-03-26
Buildah doesn't allow symlinks in newly created VOLUMEs to point outside the VOLUME root


bburky:
**Description**

Buildah doesn't allow symlinks in newly created VOLUMEs to point outside the VOLUME root.

**Steps to reproduce the issue:**
I receive an "invalid symlink" error when using the following Dockerfile with `buildah bud`:

```
# Build with Docker:
#  docker build .
# This works.

# Run with buildah / podman:
#   buildah bud .
#   podman build .
# Fails with invalid symlink error.

FROM registry.access.redhat.com/ubi7/ubi
RUN mkdir /escape && ln -s ../../../../../../../../../../ /escape/symlink
VOLUME /escape
RUN echo "Buildah doesn't allow symlinks in newly created VOLUMEs to point outside the VOLUME root"
```

**Describe the results you received:**

Full error output:

```
$ buildah bud .
STEP 1: FROM registry.access.redhat.com/ubi7/ubi
Getting image source signatures
Copying blob fcd63ccfdd0c done
Copying blob 09dbbf8834d2 done
Copying config 0355cd652b done
Writing manifest to image destination
Storing signatures
STEP 2: RUN mkdir /escape && ln -s ../../../../../../../../../../ /escape/symlink
STEP 3: VOLUME /escape
STEP 4: RUN echo "Buildah doesn't allow symlinks in newly created VOLUMEs to point outside the VOLUME root"
Buildah doesn't allow symlinks in newly created VOLUMEs to point outside the VOLUME root
error building at STEP "RUN echo "Buildah doesn't allow symlinks in newly created VOLUMEs to point outside the VOLUME root"": error extracting archive at "/home/vagrant/.local/share/containers/storage/overlay/506b97dde1864ad1e979d6a7e3eff641a8124411be55bbdce5165336b6d91fd4/merged/escape": invalid symlink "/home/vagrant/.local/share/containers/storage/overlay/506b97dde1864ad1e979d6a7e3eff641a8124411be55bbdce5165336b6d91fd4/merged/escape/symlink" -> "../../../../../../../../../../"
ERRO exit status 1
```

**Describe the results you expected:**

This Dockerfile builds fine with `docker build` (Docker Desktop for Mac, Docker 2.2.0.4). The build does not fail. The symlink is preserved inside the image.

It appears that when creating new VOLUMEs any symlink already under that path cannot point outside the VOLUME root when using buildah/podman. This is a difference in behavior from Docker.

[...]

PR #3083, opened 2021-03-17
Check for symlink in builtin volume

umohnani8:

#### What type of PR is this?

 /kind bug

Check if a builtin volume is a symlink. If it is,
follow the symlink and ensure that the destination
exists.
Add tests for symlink and no symlink case.

#### How to verify it

Follow reproducer in https://github.com/containers/buildah/issues/2204#issue-575395639

Fixes https://github.com/containers/buildah/issues/2204 and https://github.com/containers/buildah/issues/2249

```release-note

```

### DIFF
--- a/chroot/run.go
+++ b/chroot/run.go
@@ -20,6 +20,7 @@ import (
 	"unsafe"
 
 	"github.com/containers/buildah/bind"
+	"github.com/containers/buildah/copier"
 	"github.com/containers/buildah/util"
 	"github.com/containers/storage/pkg/ioutils"
 	"github.com/containers/storage/pkg/mount"
@@ -1161,7 +1162,18 @@ func setupChrootBindMounts(spec *specs.Spec, bundlePath string) (undoBinds func(
 			}
 		}
 		target := filepath.Join(spec.Root.Path, m.Destination)
-		if _, err := os.Stat(target); err != nil {
+		// Check if target is a symlink
+		stat, err := os.Lstat(target)
+		// If target is a symlink, follow the link and ensure the destination exists
+		if err == nil && stat != nil && (stat.Mode()&os.ModeSymlink != 0) {
+			target, err = copier.Eval(spec.Root.Path, m.Destination, copier.EvalOptions{})
+			if err != nil {
+				return nil, errors.Wrapf(err, "evaluating symlink %q", target)
+			}
+			// Stat the destination of the evaluated symlink
+			_, err = os.Stat(target)
+		}
+		if err != nil {
 			// If the target can't be stat()ted, check the error.
 			if !os.IsNotExist(err) {
 				return undoBinds, errors.Wrapf(err, "error examining %q for mounting in mount namespace", target)
--- a/run_linux.go
+++ b/run_linux.go
@@ -359,7 +359,17 @@ func runSetupBuiltinVolumes(mountLabel, mountPoint, containerDir string, builtin
 			}
 			initializeVolume = true
 		}
-		stat, err := os.Stat(srcPath)
+		// Check if srcPath is a symlink
+		stat, err := os.Lstat(srcPath)
+		// If srcPath is a symlink, follow the link and ensure the destination exists
+		if err == nil && stat != nil && (stat.Mode()&os.ModeSymlink != 0) {
+			srcPath, err = copier.Eval(mountPoint, volume, copier.EvalOptions{})
+			if err != nil {
+				return nil, errors.Wrapf(err, "evaluating symlink %q", srcPath)
+			}
+			// Stat the destination of the evaluated symlink
+			stat, err = os.Stat(srcPath)
+		}
 		if err != nil {
 			if !os.IsNotExist(err) {
 				return nil, err
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -727,6 +727,26 @@ function _test_http() {
   expect_output "testuser testgroup"
 }
 
+@test "bud-builtin-volume-symlink" {
+  # This Dockerfile needs us to be able to handle a working RUN instruction.
+  skip_if_no_runtime
+
+  _prefetch alpine
+  target=volume-symlink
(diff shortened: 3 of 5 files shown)


run_linux.go
```diff
@@ -359,26 +359,36 @@ func runSetupBuiltinVolumes(mountLabel, mountPoint, containerDir string, builtin
 			}
 			initializeVolume = true
 		}
-		stat, err := os.Stat(srcPath)
-		if err != nil {
-			if !os.IsNotExist(err) {
-				return nil, err
+		// Check if srcPath is a symlink
+		stat, statErr := os.Lstat(srcPath)
```
rhatdan, line 363: nit: no reason not to just leave this as err.

rhatdan (approved):

TomSweeneyRedHat:
New tests looks a little unhappy.
```
[+0414s] not ok 74 bud-builtin-volume-symlink
[+0414s] # (from function `die' in file ./helpers.bash, line 196,
[+0414s] #  from function `run_buildah' in file ./helpers.bash, line 183,
[+0414s] #  in test file ./bud.bats, line 739)
[+0414s] #   `run_buildah run $cid echo hello' failed
```

umohnani8:
Yeah, I was trying to figure out why the containerized integration test was unhappy but the same test in the other suites are fine...

run_linux.go
umohnani8: the smoke test is unhappy when I use err, it claims one of the stat `err` is unused, which is not the case. Going to leave it as `statErr`.

chroot/run.go
```diff
@@ -1161,10 +1162,21 @@ func setupChrootBindMounts(spec *specs.Spec, bundlePath string) (undoBinds func(
 			}
 		}
 		target := filepath.Join(spec.Root.Path, m.Destination)
-		if _, err := os.Stat(target); err != nil {
+		// Check if target is a symlink
+		stat, statErr := os.Lstat(target)
+		// If target is a symlink, follow the link and ensure the destination exists
+		if stat != nil && (stat.Mode()&os.ModeSymlink != 0) {
```
rhatdan, line 1168: If you change the statErr above to err, and then change this line to 

if err == nil && (stat.Mode()&os.ModeSymlink != 0) {

Everything will be happy including me.  :^)

run_linux.go
```diff
@@ -359,26 +359,36 @@ func runSetupBuiltinVolumes(mountLabel, mountPoint, containerDir string, builtin
...
+		// If srcPath is a symlink, follow the link and ensure the destination exists
+		if stat != nil && (stat.Mode()&os.ModeSymlink != 0) {
```
rhatdan, line 365: If you change the statErr above to err, and then change this line to

if err == nil && (stat.Mode()&os.ModeSymlink != 0) {

Everything will be happy including me. :^)

rhatdan (approved):

run_linux.go
umohnani8: Fixed

umohnani8:
This is ready, @nalind @TomSweeneyRedHat @rhatdan PTAL

run_linux.go
```diff
@@ -359,7 +359,17 @@ func runSetupBuiltinVolumes(mountLabel, mountPoint, containerDir string, builtin
 			}
 			initializeVolume = true
 		}
-		stat, err := os.Stat(srcPath)
+		// Check if srcPath is a symlink
+		stat, err := os.Lstat(srcPath)
+		// If srcPath is a symlink, follow the link and ensure the destination exists
+		if err == nil && stat != nil && (stat.Mode()&os.ModeSymlink != 0) {
+			srcPath, err = copier.Eval(mountPoint, volume, copier.EvalOptions{})
```
nalind, line 366: Should that second parameter be `srcPath`?

umohnani8: srcPath is the joined path of `mountPoint` and `volume` and copier.Eval takes the root and then path separately, that is why the second param is volume.

rhatdan:
LGTM
@vrothberg @saschagrunert @giuseppe PTAL

chroot/run.go
```diff
@@ -1161,7 +1162,18 @@ func setupChrootBindMounts(spec *specs.Spec, bundlePath string) (undoBinds func(
...
 		}
 		target := filepath.Join(spec.Root.Path, m.Destination)
-		if _, err := os.Stat(target); err != nil {
+		// Check if target is a symlink
+		stat, err := os.Lstat(target)
+		// If target is a symlink, follow the link and ensure the destination exists
+		if err == nil && stat != nil && (stat.Mode()&os.ModeSymlink != 0) {
+			target, err = copier.Eval(spec.Root.Path, m.Destination, copier.EvalOptions{})
+			if err != nil {
+				return nil, errors.Wrapf(err, "error evaluating symlink %q", target)
```
saschagrunert, line 1171:
```suggestion
				return nil, errors.Wrapf(err, "evaluating symlink %q", target)
```

run_linux.go
```diff
@@ -359,7 +359,17 @@ func runSetupBuiltinVolumes(mountLabel, mountPoint, containerDir string, builtin
...
+			if err != nil {
+				return nil, errors.Wrapf(err, "error evaluating symlink %q", srcPath)
```
saschagrunert, line 368:
```suggestion
				return nil, errors.Wrapf(err, "evaluating symlink %q", srcPath)
```

saschagrunert (approved):
Just two nits, otherwise LGTM

TomSweeneyRedHat:
Other than @saschagrunert's comments LGTM

chroot/run.go
umohnani8: Fixed

run_linux.go
umohnani8: Fixed

saschagrunert (approved):

giuseppe (approved):
LGTM

PR #3095, opened 2021-03-22
COPY --chown: expand the conformance test

nalind:
#### What type of PR is this?

/kind bug

Expand conformance test coverage for COPY --chown, and correct our behavior when the argument is a single number: instead of assuming the number is the UID and the GID = 0, the GID should be the same as the UID.

#### How to verify it

New conformance test!

None

Grew out of the conformance tests for https://github.com/openshift/imagebuilder/pull/188.

```
When `COPY --chown` specifies an owner that's just a single number (e.g. "12"), the UID on copied content is set to the specified value, and the GID was previously set to 0.  This changes the GID to be set to the same value as the UID.
```

### DIFF
--- a/add.go
+++ b/add.go
@@ -251,9 +251,9 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 
 	// Find out which user (and group) the destination should belong to.
 	var chownDirs, chownFiles *idtools.IDPair
-	var user specs.User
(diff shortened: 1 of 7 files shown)


giuseppe (approved):
LGTM

TomSweeneyRedHat:
LGTM

PR #3097, opened 2021-03-23
Use faster way to check image tag existence during multi-arch build

barthy1:
#### What type of PR is this?

/kind other

Use `skopeo list-tags` instead of `docker pull` makes check of image tag existence during multi-arch image build workflow much faster.
This is slight modification of the approach suggested by @cevich in previous [PR](https://github.com/containers/buildah/pull/3059#discussion_r590478210) for multi-arch build.


rhatdan:
LGTM
@cevich @TomSweeneyRedHat PTAL

TomSweeneyRedHat:
LGTM
/approve
/lgtm

Issue #3096, opened 2021-03-22
system tests failing due to shortnames


edsantiago:
Fedora gating tests are now [failing](https://osci-jenkins-1.ci.fedoraproject.org/job/fedora-ci/job/dist-git-pipeline/job/master/31061/testReport/(root)/tests/root_test/) due to shortnames:
```
# [podman pull ubuntu]
# Error: error getting default registries to try: short-name resolution enforced but cannot prompt without a TTY
```
This is going to be difficult to solve due to the prevalence of shortnames in buildah tests. Possibilities I see are:
* replace all shortnames in `buildah/tests` with FQINs - this might cost us test coverage of buildah shortname handling
* hand-hack the podman shortnames file (`~/.config/.cache/containers/short-name-aliases.conf`) - this might not be appreciated by people running tests on a shared system.
* override podman, just tell it "do what you used to do" - is that possible? I can't find much documentation on shortnames

Any other ideas?

vrothberg:
> Fedora gating tests are now [failing](https://osci-jenkins-1.ci.fedoraproject.org/job/fedora-ci/job/dist-git-pipeline/job/master/31061/testReport/(root)/tests/root_test/) due to shortnames:
> 
> ```
> # [podman pull ubuntu]
> # Error: error getting default registries to try: short-name resolution enforced but cannot prompt without a TTY
> ```
> 
> This is going to be difficult to solve due to the prevalence of shortnames in buildah tests. Possibilities I see are:
> 
>     * replace all shortnames in `buildah/tests` with FQINs - this might cost us test coverage of buildah shortname handling

Yes, we should aim for the test coverage.  The failing tests are good sign in that context :)

>     * hand-hack the podman shortnames file (`~/.config/.cache/containers/short-name-aliases.conf`) - this might not be appreciated by people running tests on a shared system.

Note that the path is not `~/.config/.cache` anymore (was a bug).  It's not `~/.cache`.

I think the tests should run with https://github.com/containers/shortnames/blob/main/shortnames.conf in the `/etc/containers/registries.conf.d`.  This config is now also shipped in the Fedora package (the one that also install registries.conf).

>     * override podman, just tell it "do what you used to do" - is that possible? I can't find much documentation on shortnames

Prompting only works with a TTY.  Please let me know if `man registries.conf` is missing information.

edsantiago:
>I think the tests should run with 

[...]

PR #3099, opened 2021-03-23
tests: prefetch: use buildah, not podman, for pulls

edsantiago:
The _prefetch helper has proven invaluable in reducing pull-
related flakes. However, it is now failing because it uses
podman to pull, and buildah tests rely on shortnames, and
podman no longer works with shortnames.

Solution: use buildah, not podman, for initial image pulls.
(The reason we used podman in the first place is because
buildah does not have save/load. But since image store
is shared, we can use buildah for pull and podman for save
and load.)

Add shortname config to registries.conf. I don't think it's
actually necessary or even used, but we can tweak it later.

And, fix one test in from.bats that was using inconsistent
names between buildah and podman.

Fixes: #3096 

### DIFF
--- a/tests/from.bats
+++ b/tests/from.bats
@@ -425,7 +425,7 @@ load helpers
   fi
 
   _prefetch docker.io/busybox
-  podman run --name busyboxc-podman -d busybox top
+  podman run --name busyboxc-podman -d docker.io/busybox top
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

rhatdan:
/approved
@nalind @TomSweeneyRedHat @umohnani8 @ashley-cui @vrothberg PTAL

TomSweeneyRedHat:
LGTM

PR #3101, opened 2021-03-24
buildah source - create and manage source images

vrothberg:
Add new `buildah source {create,add,push,pull}` commands.  All commands
are marked as experimental.  None of it is meant to be officially
supported at the time of writing.  All code resides in `internal/source`
and is hence not visible to external consumers of Buildah; just to be
on the safe side.

A source container or source image is an OCI artifact, that is an OCI
image with custom config (media type).  There is a longer history behind
source images which are intended to ship the source artifacts of an
ordinary "executable" container image.  Until now, source images at
Red Hat are built with github.com/containers/BuildSourceImage.  We had a
growing desire (and always the long-term plan) to eventually replace
BuildSurceImage with something else, in this case Buildah.

This commit adds the initial base functionality along with tests to make
sure we're not regressing.  The new commands do the following:

* `create` - creates an empty and initialized source image
* `add` - tar up a local path and add it as a layer to the souce image
* `push/pull` - intentionally separate commands from `buildah push/pull`
                to allow for an easier usage and prevent the
                implementations from undesired (future) interference

Further note: also vendor in c/image@master which ships a required fix.

### DIFF
--- a/cmd/buildah/source.go
+++ b/cmd/buildah/source.go
@@ -1,0 +1,125 @@
+package main
+
+import (
+	"context"
+
+	"github.com/containers/buildah/internal/source"
+	"github.com/spf13/cobra"
+)
+
+var (
+	// buildah source
+	sourceDescription = `  Create, push, pull and manage source images and associated source artifacts.  A source image contains all source artifacts an ordinary OCI image has been built with.  Those artifacts can be any kind of source artifact, such as source RPMs, an entire source tree or text files.
+
+  Note that the buildah-source command and all its subcommands are experimental and may be subject to future changes.
+`
+	sourceCommand = &cobra.Command{
+		Use:   "source",
+		Short: "Manage source containers",
+		Long:  sourceDescription,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return nil
+		},
+	}
+
+	// buildah source create
+	sourceCreateDescription = `  Create and initialize a source image.  A source image is an OCI artifact; an OCI image with a custom config media type.
+
+  Note that the buildah-source command and all its subcommands are experimental and may be subject to future changes.
+`
+	sourceCreateOptions = source.CreateOptions{}
+	sourceCreateCommand = &cobra.Command{
+		Args:    cobra.ExactArgs(1),
+		Use:     "create",
+		Short:   "Create a source image",
+		Long:    sourceCreateDescription,
+		Example: "buildah source create /tmp/fedora:latest-source",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return source.Create(context.Background(), args[0], sourceCreateOptions)
+		},
+	}
+
+	// buildah source add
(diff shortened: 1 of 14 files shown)


vrothberg:
@nalind are you cool with it?  I intentionally keep everything behind `buildah source` and in `internal/source` to keep it separate from the rest of the Buildah code base.

go.mod
```diff
@@ -4,11 +4,12 @@ go 1.12
 
 require (
 	github.com/containernetworking/cni v0.8.1
-	github.com/containers/common v0.35.2
-	github.com/containers/image/v5 v5.10.5
+	github.com/containers/common v0.34.2
+	github.com/containers/image/v5 v5.10.2-0.20210324124443-224346a8c9a7
```
vrothberg, line 8: Go should know better ... it's really trying hard to match a commit to a version and fails so often :(

Looks like we need to wait for the next c/image release until we can merge.

vrothberg:
Will add man pages in a jiffy (need it to pass CI)

nalind:
I'm not convinced that this is a good fit here.

From a quick skim of the new source package, I'm not seeing many dependencies on privileges or OS-specific features like kernel namespaces or copy-on-write filesystems.  Functionality that we add to buildah tends to be of the sort that uses these things, and which therefore has to be made available via something like podman-remote for users who want to make use of it from non-Linux systems.

If we bury this in buildah, we'll either have to do that wiring and force non-Linux users to boot a Linux VM, or tell them to compile a mostly-otherwise-useless-for-them native buildah binary.

vrothberg:
> From a quick skim of the new source package, I'm not seeing many dependencies on privileges or OS-specific features like kernel namespaces or copy-on-write filesystems. Functionality that we add to buildah tends to be of the sort that uses these things, and which therefore has to be made available via something like podman-remote for users who want to make use of it from non-Linux systems.
> 
> If we bury this in buildah, we'll either have to do that wiring and force non-Linux users to boot a Linux VM, or tell them to compile a mostly-otherwise-useless-for-them native buildah binary.

It is meant to be Linux and Buildah only.  Indeed, the `internal/source` package is independent and could be a separate tool but I think that Buildah is the right home for. Source containers are tied to building containers and Buildah is our Swiss Army knife for it.

Since we aim for a "standardized" OCI artifact, there may be other tools (from others) in the future to build them but I think Buildah should be the first one.

rhatdan:
@vrothberg Please rebase.

vrothberg:
> @vrothberg Please rebase.

Need a new c/image first. `buildah source ...` won't work without the fix for OCI artifacts.

TomSweeneyRedHat:
All kinds of red tests @vrothberg, I think it wants some more attention.

vrothberg:
@rhatdan, last but not least, green

rhatdan:
LGTM

vrothberg:

vrothberg:
Friendly ping

PR #3114, opened 2021-03-30
run_linux: warn on overlapping bind mounts

nalind:
#### What type of PR is this?

/kind other

Warn when one configured bind mount will "hide" the contents of another configured bind mount.  The list of mounts comes from multiple places now, so it can be hard to troubleshoot this problem.

#### How to verify it

`buildah run -v /var/tmp/foo:/run/secrets/foo $container /bin/sh` will not be able to see `/run/secrets/foo` because it's hidden by a mount at `/run/secrets`.  At least now you'll get a warning about it.

None

```
None
```

### DIFF
--- a/run.go
+++ b/run.go
@@ -3,6 +3,7 @@ package buildah
 import (
 	"fmt"
 	"io"
+	"path/filepath"
 
 	"github.com/containers/buildah/define"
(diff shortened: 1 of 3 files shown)


rhatdan:
Shouldn't buildah sort these correctly, so that the /run/secrets/foo shows up?

PR #3118, opened 2021-03-31
System-test cleanup

edsantiago:
Primarily desired for the podman CI setup in which we
run bud.bats tests using podman-build; but these are
all worthwhile changes useful for buildah itself.

bud.bats:
- from-scratch-iid test: use expect_output to confirm
  leading 'sha256' (was: [[ ... ]])

- hardlinks test: add test-name arg to expect_output
  (to help diagnose test failures, such as just happened
  in podman tests)

- build-arg-cache test: add test-name string to expect_output,
  to help someone narrow down which test failed.

- stdin tests: use 'run_buildah <' instead of '| buildah'

- remove trailing whitespace

helpers.bash: reindent, converting tabs to spaces. TL;DR this
is needed for running buildah-bud tests under podman(*). As
of this commit, there are no tab characters in the file.

 * Long version: podman CI runs bud.bats using 'podman build'.
   This requires a few changes to bud.bats and helpers.bash,
   changes which need to be carried in podman itself, not in
   the buildah repo (like grokking 'remote', a concept that
   is meaningless in buildah-land). We use 'git diff', which
   is funny about whitespace, which in turn causes podman
   CI hooks to fail when we try to check in the .diff file.

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -400,7 +400,7 @@ symlink(subdir)"
   target=scratch-image
   run_buildah bud --iidfile ${TESTDIR}/output.iid --signature-policy ${TESTSDIR}/policy.json -t ${target} ${TESTSDIR}/bud/from-scratch
   iid=$(cat ${TESTDIR}/output.iid)
-  [[ "$iid" == "sha256:"* ]]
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

PR #3121, opened 2021-04-01
Add conformance tests for COPY ""

nalind:
#### What type of PR is this?

/kind other

Add a couple of conformance tests exercising "" as a source name for COPY instructions, so that we don't regress on https://github.com/openshift/imagebuilder/issues/181 in the future.
Add a conformance test for COPY --from where the source location is a symbolic link.
Add conformance tests for COPY where the source location is an absolute path with multiple components.

#### How to verify it

New conformance tests!

Nothing here, this just adds tests that we already would have passed.

```
None
```


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM
assuming happy tests.

nalind:
Added a few more tests.

rhatdan:
@nalind seems to have been cancelled in cirrus, we might need to repush to get the ci/cd system working again.

PR #3122, opened 2021-04-01
Add system tests of "buildah run"

hshiina:

#### What type of PR is this?

/kind other

Tests for the following options are added:
- buildah run --runtime --runtime-flag
- buildah run --terminal
- buildah run with namespace flags (Fix a bug in the existing test)

#### How to verify it
Run system tests.

None

```release-note
None
```

### DIFF
--- a/tests/namespaces.bats
+++ b/tests/namespaces.bats
@@ -254,7 +254,12 @@ general_namespace() {
       ;;
     esac
 
-    for different in $types ; do
+    if [ "$nsflag" = "userns" ]; then
+      # "run" doesn't have --userns option.
+      continue
+    fi
+
+    for different in ${types[@]} ; do
       # Check that, if we override it, we get what we specify for "run".
       run_buildah run $RUNOPTS --"$nsflag"=$different "$ctr" readlink /proc/self/ns/"$nstype"
       [ "$output" != "" ]
@@ -266,7 +271,7 @@ general_namespace() {
         expect_output "$mynamespace"
         ;;
       /*)
-        expect_output "$(readlink $namespace)"
+        expect_output "$(readlink $different)"
         ;;
       esac
     done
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
LGTM
@edsantiago PTAL

tests/run.bats
```diff
@@ -545,3 +545,54 @@ function configure_and_check_user() {
...
+        {
+	        "name": "unknown",
+			"action": "SCMP_ACT_KILL"
+	    }
+    ]
+}
+_EOF
+	run_buildah from --security-opt seccomp=${TESTDIR}/seccomp.json --quiet --pull=false --signature-policy ${TESTSDIR}/policy.json alpine
+	cid=$output
+
```
edsantiago, line 569: I would encourage you to do something like:
```bash
    local found_runtime=
    if [ -n ...]; then
        found_runtime=y
    ...
    if [ -n ...crun]
        found_runtime=y
    ...

    if [[ -n "$found_runtime" ]]; then
        die "Did not find 'runc' nor 'crun' in \$PATH - could not run this test!"
    fi
```
(I know there's already `skip_if_no_runtime`, but I can be very paranoid; and I've seen countless cases where tests pass because they aren't actually running anything).

edsantiago (changes requested):
One request, please; otherwise LGTM

tests/namespaces.bats
```diff
@@ -266,7 +271,7 @@ general_namespace() {
         expect_output "$mynamespace"
         ;;
       /*)
-        expect_output "$(readlink $namespace)"
+        expect_output "$(readlink $different)"
```
edsantiago, line 274: This looks like a good catch to me, but how did this ever work?

hshiina: This case has not been tested.
Because of `$types` in L257, only the case where `different` is empty (`types[0]`) has been tested.

edsantiago (approved):

PR #3123, opened 2021-04-01
Fix '--tty' flag in run subcommand

hshiina:

#### What type of PR is this?

/kind bug

See if '--tty' flag is specified as well as '--terminal' flag in `run` subcommand.

#### How to verify it
You can verify with the following procedure.

~~~
$ buildah from busybox 
$ buildah run --tty=true busybox-working-container ls --color=auto
$ buildah run --tty=false busybox-working-container ls --color=auto
~~~

The output with `--tty=true` should be colored while the output with `--tty=false` should be uncolored.

None

```release-note
None
```

### DIFF
--- a/cmd/buildah/run.go
+++ b/cmd/buildah/run.go
@@ -137,7 +137,7 @@ func runCmd(c *cobra.Command, args []string, iopts runInputOptions) error {
 		DropCapabilities: iopts.capDrop,
 	}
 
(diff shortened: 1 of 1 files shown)


rhatdan:
Thanks @hshiina but I think #3124 is the correct way to handle this.  PTAL and tell me if this works for you

PR #3133, opened 2021-04-07
CLI add/copy: add a --from option

### LINKED ISSUE #3127: [question] Multi-stage builds
How do you perform a multi-stage build using the `buildah` command? I was expecting to find this:
```sh
container1=$(buildah from image1:tag)
container2=$(buildah from image2:tag)

buildah copy --from=$container2 /example/file.txt /file/destination $container1
```

But the --help information does not mention anything like this. I wonder if buildah really supports multi-stage builds.

nalind:
#### What type of PR is this?

/kind feature

Add a --from option to `buildah add` and `buildah copy`, mirroring the option for the Dockerfile instruction.

#### How to verify it

New integration tests!

Fixes #3127.

```release-note
`buildah add` and `buildah copy` now accept a `--from` option, which can be used to copy content from other working containers or images.
```

### DIFF
--- a/contrib/completions/bash/buildah
+++ b/contrib/completions/bash/buildah
@@ -534,6 +534,8 @@ return 1
      -h
      --quiet
      -q
+     --tls-verify
+     --remove-signatures
   "
 
      local options_with_args="
@@ -541,6 +543,11 @@ return 1
      --chmod
      --contextdir
      --ignorefile
+     --from
+     --authfile
+     --cert-dir
+     --creds
+     --decryption-key
   "
 
      local all_options="$options_with_args $boolean_options"
@@ -559,6 +566,8 @@ return 1
      -h
      --quiet
      -q
+     --tls-verify
+     --remove-signatures
     "
 
      local options_with_args="
@@ -566,6 +575,11 @@ return 1
      --chmod
      --contextdir
      --ignorefile
+     --from
+     --authfile
+     --cert-dir
+     --creds
+     --decryption-key
   "
 
      local all_options="$options_with_args $boolean_options"
--- a/docs/buildah-add.md
+++ b/docs/buildah-add.md
@@ -1,4 +1,4 @@
-# buildah-add "1" "March 2017" "buildah"
+# buildah-add "1" "April 2021" "buildah"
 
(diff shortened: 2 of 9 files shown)


contrib/completions/bash/buildah
```diff
@@ -559,13 +565,19 @@ return 1
...
 
      local options_with_args="
      --chown
      --chmod
      --contextdir
      --ignorefile
+     --authfile
+     --cert-dir
+     --creds
+     --decryption-key
```
TomSweeneyRedHat, line 580: ditto --from question

docs/buildah-add.md
```diff
@@ -1,4 +1,4 @@
-# buildah-add "1" "March 2017" "buildah"
+# buildah-add "7" "April 2021" "buildah"
```
TomSweeneyRedHat, line 1: should this be 7?  If so, do we have other references in other man pages that need to change?  Ditto same question for copy below.

nalind: Oh, no, for some reason my brain farted and thought that `1` was the day of the month.  Reverting it.

nalind:
Added a few comments to the new integration tests, put five more minutes on the clock for unit tests to complete.

contrib/completions/bash/buildah
nalind: Can you elaborate on this?  I don't understand.

TomSweeneyRedHat: It's set now, I was asking to have the `--from` added here too.

TomSweeneyRedHat:
LGTM
but I'm surprised the tests added 5 minutes....
@vrothberg @rhatdan PTAL

vrothberg (commented):
LGTM.  Such a cool feature!

nalind:
> LGTM
> but I'm surprised the tests added 5 minutes....
> @vrothberg @rhatdan PTAL

Along the way we had a few unit test and integration test jobs time out, so I bumped the timeouts for them to try to not have to restart them.  Red Hat's outage yesterday must have contributed to that, otherwise I wouldn't have expected the unit test run time to change.

PR #3140, opened 2021-04-12
bats helpers.bash - minor refactoring

edsantiago:
Almost trivial, but this will greatly simplify my podman-remote
testing of buildah-bats.

### DIFF
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -22,6 +22,10 @@ function setup() {
     TESTDIR=${BATS_TMPDIR}/tmp${suffix}
     rm -fr ${TESTDIR}
     mkdir -p ${TESTDIR}/{root,runroot}
+
+    # Common options for all buildah and podman invocations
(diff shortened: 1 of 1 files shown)


rhatdan:
/approve
LGTM

tests/helpers.bash
```diff
@@ -60,24 +64,18 @@ function teardown() {
 }
 
 function _prefetch() {
-    # Tell podman to use the same mirror registries as buildah, to
-    # avoid docker.io
-    export REGISTRIES_CONFIG_PATH=${TESTSDIR}/registries.conf
-
```
TomSweeneyRedHat, line 66: I thought we somehow referenced this in Podman, but I'm not finding it....

edsantiago: Podman uses it - but we no longer use podman for pulls (See #3099). I should've removed these lines as part of that PR, but didn't notice it at the time. This is just belated cleanup.

And, in any event, the new `podman()` helper function now explicitly includes `$REGISTRY_OPTS` as a command-line argument, which is better than an envariable.

PR #3141, opened 2021-04-12
Run container with isolation type set at 'from'

hshiina:

#### What type of PR is this?

/kind bug

When an isolation type (`--isolation`) is not specified at `buildah run`, the isolation type specified at `buildah from` should be used. However, when `chroot` is specified at `buildah from`, `oci` is used at `run` even if neither the `--isolation` flag nor the environment value `BUILDAH_ISOLATION` is specified.

For the purpose, this fix stop using `DefaultIsolation()`, which sets `oci` to `--isolation` flag as the default. By this fix, the isolation type specified at `buildah from` is used when the container runs. The environment variable `BUILDAH_ISOLATION` is read at parsing the flag and still overrides the type.

#### How to verify it
A new test is added to `from.bats`.

None

```release-note
None
```


rhatdan:
LGTM

rhatdan:
/approve
@TomSweeneyRedHat @nalind @ashley-cui @umohnani8 @vrothberg PTAL

nalind:
LGTM

PR #3143, opened 2021-04-13
copier: add GetOptions.NoCrossDevice

nalind:
#### What type of PR is this?

/kind feature

Add a `NoCrossDevice` flag to `copier.GetOptions`, telling it to ignore subdirectories on devices different than the top reference directory that we start from, i.e., ignore the contents of mounted filesystems.

#### How to verify it

New unit test!

None

Not _completely_ sure if the home device of the directory passed as the second parameter to Get() is the right thing to compare to, but if we used the top of a directory tree as the baseline when descending into directories, if the top directory was a mountpoint (e.g., if we were copying "*" from "/", and the glob matched "/proc"), the flag wouldn't end up changing the archive we produced.

```
None
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -285,6 +285,7 @@ type GetOptions struct {
 	Rename             map[string]string // rename items with the specified names, or under the specified names
 	NoDerefSymlinks    bool              // don't follow symlinks when globs match them
 	IgnoreUnreadable   bool              // ignore errors reading items, instead of returning an error
+	NoCrossDevice      bool              // if a subdirectory is a mountpoint with a different device number, include it but skip its contents
 }
 
 // Get produces an archive containing items that match the specified glob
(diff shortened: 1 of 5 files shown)


rhatdan:
Buildah CLI not using this option?

nalind:
`COPY` doesn't have a switch for it, so this doesn't add one to the CLI.  Better to wait until we're sure that the way it works is the way we want it to work.  The whole "copying a directory copies its contents" thing we have to do for compatibility's sake makes the rule confusing no matter what.

rhatdan:
LGTM

nalind:
Rebased.

TomSweeneyRedHat:
CI couldn't pull the registry image from DockerHub.  Perhaps a rate limit, but it doesn't have the right foot print for that.  Retrying.
Otherwise,
LGTM

PR #3148, opened 2021-04-15
use new containers/common/libimage package

vrothberg:
Move all code related to handling container images over to the new
`libimage` package in containers/common.  The new package is an
attempt to consolidate the code across the containers tools under the
github.com/containers umbrella.

The new `libimage` package provides functionality to perform all kinds
of operations for managing images such as local lookups, pushing,
pulling, listing, removing, etc.

The following packages have been moved over to containers/common:

`manifests` -> `common/libimage/manifests`
`pkg/manifests` -> `common/pkg/manifests`
`pkg/supplemented` -> `common/pkg/supplemented`


rhatdan:
@vrothberg needs rebase.

vrothberg:
> @vrothberg needs rebase.

Want to get it green first :)  Getting closer.

vrothberg:
Finally green. @rhatdan @nalind PTAL.

This is a _gigantic_ change but I did not find a way to split it into reasonable chunks.  Once https://github.com/containers/common/pull/497 is merged, I'll update the PR to vendor it the commit from c/common.

I moved as much code as I could find over to use `common/libimage`.  We may find more candidates in the future but I think we're good for now.

cmd/buildah/commit.go
```diff
@@ -11,7 +11,8 @@ import (
 	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/buildah/util"
 	"github.com/containers/common/pkg/auth"
-	"github.com/containers/image/v5/storage"
+	"github.com/containers/image/v5/pkg/shortnames"
+	storageTransport "github.com/containers/image/v5/storage"
```
rhatdan, line 15: Why are you doing this change?

cmd/buildah/images.go
```diff
@@ -140,15 +121,28 @@ func imagesCmd(c *cobra.Command, args []string, iopts *imageResults) error {
...
 	systemContext, err := parse.SystemContextFromOptions(c)
 	if err != nil {
 		return errors.Wrapf(err, "error building system context")
 	}
+	runtime, err := libimage.RuntimeFromStore(store, &libimage.RuntimeOptions{SystemContext: systemContext})
+	if err != nil {
+		return err
+	}
+
+	ctx := context.Background()
```
rhatdan, line 133: Nit: Why create a variable to use it once?

rhatdan (approved):

rhatdan:
@giuseppe @nalind @TomSweeneyRedHat @ashley-cui @umohnani8 PTAL

cmd/buildah/commit.go
vrothberg: Some of our packages have the same name. In this case, `storage` could very well mean `containers/storage` and I find it sometimes hard to brain-parse code with all that ambiguity. That's why I used `storageTransport` or `dockerTransport` import aliases in `libimage` in hope to make code easier to read.

It's also the reason why I called it `libimage` and not `image`. This way, readers won't confuse it with `containers/image`.

vrothberg:
Cc @baude @mheon

vrothberg:
Moving back to work-in-progress as I will start tackling Podman now which will likely require some changes to the libimage API.

vrothberg:
@rhatdan @TomSweeneyRedHat rebased and ready for your hungry eyes

nalind:
buildah is also a library, and this currently changes the signatures of exported functions, renames others, and removes exported types, functions, and packages entirely.  It will break consumers of the library.

vrothberg:
> buildah is also a library, and this currently changes the signatures of exported functions, renames others, and removes exported types, functions, and packages entirely. It will break consumers of the library.

Which consumers are you most concerned about?

nalind:
> > buildah is also a library, and this currently changes the signatures of exported functions, renames others, and removes exported types, functions, and packages entirely. It will break consumers of the library.
> 
> Which consumers are you most concerned about?

The first one that comes to mind is OpenShift's builder, but the concern is a general one that we have as a library.

vrothberg:
Yes that is a general issue but that’s an inevitable part of the code
consolidation. We need to do the same with libpod/Podman and eventually
CRI-O.

If changing parts of the external APIs is a deal breaker, I wished to have
known earlier; moving packages and the extend of the changes was discussed
a bunch of times.

@rhatdan, @nalind, I let you folks decide but I may need medication if I
have to thrash this PR.

A possible alternative solution is to cut a release branch now and then
merge the PR targeting a bump to v2.

rhatdan:
Let's see how bad the port to openshift builder is going to be.  If there is breakage, we can add back the parts that break as references to containers/libimage,

vrothberg:
I looked at openshift/builder which does not build as I removed Push, Pull and changed the signature of util.FindImage.

I will spin up a follow-up PR tomorrow to add Push/Pull back. That's fairly trivial as I can move the logic in `cmd/buildah/....`. I'll also get the signature of FindImage back.

PR #3149, opened 2021-04-15
Log the final error with %+v at logging level "trace"

nalind:
#### What type of PR is this?

/kind other

If the logging level is at least "trace", log the final error we print, if there is one, using `%+v` as the verb, or `%v` (changed from `%s`) otherwise.  If the error was wrapped using `github.com/pkg/errors`, this will provide a backtrace.

#### How to verify it

New integration test!  It only ensures that the levels we advertise in the help output and docs are all recognized, but still.

None

Notionally related to https://github.com/containers/podman/pull/10032.

```
None
```


TomSweeneyRedHat:
LGTM
I've restarted the integration test, OCI runtime was complaining about not finding a "pids" runtime.  I believe that's a flake as I saw it in another PR broken test.

PR #3151, opened 2021-04-15
Add system test of 'buildah help' with a tiny fix

hshiina:

#### What type of PR is this?

/kind other

Add a system test of `buildah help` copying a system test of `podman help`.

Also remove an unexpected empty line from the usage template of commands.

#### How to verify it
Run the added system test:

~~~
$ bats tests/help.bats
~~~

None

```release-note
None
```

### DIFF
--- a/tests/help.bats
+++ b/tests/help.bats
@@ -1,0 +1,95 @@
+#!/usr/bin/env bats
+
+load helpers
+
+# run 'buildah help', parse the output looking for 'Available Commands';
+# return that list.
+function buildah_commands() {
+    run_buildah help "$@" |\
+        awk '/^Available Commands:/{ok=1;next}/^Flags:/{ok=0}ok { print $1 }' |\
+        grep .
+}
+
+function check_help() {
+    local count=0
+    local -A found
+
+    for cmd in $(buildah_commands "$@"); do
+        # Human-readable buildah command string, with multiple spaces collapsed
+        command_string="buildah $* $cmd"
+        command_string=${command_string//  / }  # 'buildah  x' -> 'buildah x'
+
+        # help command and --help flag have the same output
+        run_buildah help "$@" $cmd
+        local full_help=$output
+
+        # The line immediately after 'Usage:' gives us a 1-line synopsis
+        usage=$(echo "$output" | grep -A1 '^Usage:' | tail -1)
+        [ -n "$usage" ] || die "$command_string: no Usage message found"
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
@edsantiago PTAL.  I thought we did something like this elsewhere?
Also the conformance test is not at all happy.

rhatdan:
@hshiina Does this do any more then ./hack/xref-helpmsgs-manpages?

edsantiago:
This is a copy-paste of the same [podman test](https://github.com/containers/podman/blob/35b62ef19f307fb365eb7a8734e456160f7b22a8/test/system/015-help.bats) with a lot of code removed; some because it's not applicable (`-l`), some for reasons I haven't looked into (possibly because the buildah help messages don't conform to the same rules as podman?).

I won't have time to review until Monday, but on first glance I'm almost totally OK with this (there's a "podman" in a `die` message that should be changed). At this time it doesn't do much more than the `xref` script, but I believe @hshiina is intending to extend that in future work; and this could give us important bug catches such as those the podman test does (usually SEGVs when invoking with wrong number of args). Plus, `xref` doesn't run in gating-test context.

hshiina:
I focused on system tests and overlooked `./hack/xref-helpmsgs-manpages`.
As @edsantiago mentions, this test does no more things than the script.
This test may not be necessary because `help` doesn't require system tests where `buildah` works with other components such as an OCI runtime.

tests/help.bats
```diff
@@ -0,0 +1,95 @@
...
+        expect_output "$full_usage"
+    else
+        run_buildah 125 "$@" NoSuchCommand
+        expect_output --substring "unknown command"
+    fi
+
+    # This can happen if the output of --help changes, such as between
+    # the old command parser and cobra.
+    [ $count -gt 0 ] || \
+        die "Internal error: no commands found in 'podman help $@' list"
```
edsantiago, line 72: `s/podman/buildah/`

edsantiago (changes requested):
@hshiina this is good work, and it does perform tests that are not present in `hack/xref-helpmsgs-manpages`. I think buildah would benefit from including this in CI. One trivial correction, and I will gladly merge this after that's fixed. Thank you!

edsantiago:
LGTM but could you please squash your commits? The typo fix doesn't merit a separate commit. Thank you!

edsantiago:
/approve
/lgtm

Thank you @hshiina !

edsantiago (approved):

PR #3164, opened 2021-04-19
tests: copy.bats: fix infinite hang

edsantiago:
The ignore-socket test in copy.bats has to create a UNIX socket
file. As it turns out, mknod won't do that: there are only
horrible kludgy options. The least bad of those is to run 'nc'
and control-C it; we do this hands-off via background job.

Until now, the test then infinite-looped a 'test -e socketfile'.
This failed catastrophically in gating tests, because 'nc' was
not installed (it was not a requirement of the RPM) (it is now).
To make things worse, it took months to track this down because
the gating-test logs aren't flushed. But that's another story.

Here we add a fixed number of retries to the 'test -e' loop,
and abort with a timeout error if the socket isn't created.
This may save someone on a distro where nc hasn't yet been
added as a dependency.

### DIFF
--- a/tests/copy.bats
+++ b/tests/copy.bats
@@ -261,9 +261,20 @@ load helpers
 
 @test "ignore-socket" {
   createrandom ${TESTDIR}/randomfile
+  # This seems to be the least-worst way to create a socket: run and kill nc
   nc -lkU ${TESTDIR}/test.socket &
-  while ! test -e ${TESTDIR}/test.socket; do sleep 0.1; done
-  kill $!
+  nc_pid=$!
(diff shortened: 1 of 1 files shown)


edsantiago:
Tested by changing `nc` in line 265 to `xxnc`. The test died after five seconds, as expected.

TomSweeneyRedHat:
LGTM

rhatdan:
LGTM
But I would have thought/hoped the test would blow up right away when nc does not exists.

vrothberg (commented):
LGTM
Restarted the flake.

PR #3168, opened 2021-04-20
Restore log timestamps

nalind:
#### What type of PR is this?

/kind bug

The rationale for removing them was that they were always zero, but that was only the case when the command ran to completion less than one second.

#### How to verify it

Observe log messages in commands that take more than a second to run.

```
Logged debugging and error messages will once again include a count of seconds elapsed since `buildah` was started.
```

### DIFF
--- a/cmd/buildah/main.go
+++ b/cmd/buildah/main.go
@@ -61,8 +61,6 @@ var (
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

PR #3171, opened 2021-04-20
push to docker test: don't get fooled by podman

edsantiago:
Gating tests have failed (at least) twice already because
one of the push.bats tests runs "which docker" and skips
if it's missing. Sadly, some gating-test systems install
podman-docker, possibly when there's a bodhi that combines
podman and buildah. This causes the test to fail.

Solution: confirm that if docker exists, it isn't podman
in disguise.

### DIFF
--- a/tests/pull.bats
+++ b/tests/pull.bats
@@ -92,15 +92,8 @@ load helpers
 }
 
 @test "pull-from-docker-daemon" {
-  run systemctl status docker
-  if [[ ! "$output" =~ "active (running)" ]]
-  then
-     skip "Skip the test as docker services is not running"
-  fi
+  skip_if_no_docker
 
-  run systemctl start docker
-  echo "$output"
-  [ "$status" -eq 0 ]
   run docker pull alpine
   echo "$output"
   [ "$status" -eq 0 ]
--- a/tests/push.bats
+++ b/tests/push.bats
@@ -140,10 +140,7 @@ load helpers
 }
 
 @test "buildah push image to docker and docker registry" {
-  run which docker
-  if [[ $status -ne 0 ]]; then
-    skip "docker is not installed"
-  fi
+  skip_if_no_docker
 
   _prefetch busybox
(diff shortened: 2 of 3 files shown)


tests/push.bats
```diff
@@ -145,6 +145,16 @@ load helpers
...
 
+  # Confirm that this is really truly docker, not podman. Pick one or more:
+  docker_version=$(docker --version)
+  if [[ $docker_version =~ podman ]]; then
+    skip "this test needs docker, not podman-docker"
+  fi
+  docker_daemon=$(ps auxww|grep docker|grep -v grep)
+  if [[ -z "$docker_daemon" ]]; then
+    skip "docker daemon is not running"
+  fi
```
edsantiago, line 156: Tagging as WIP because I don't know which of these is better, or even if there's another option. (Testing for `/run/docker.sock` is not useful; I think `podman system service` creates that).

nalind: If we're going to push to dockerd, the service needs to be running (the "pull-from-docker-daemon" from pull.bats checks for this), and the `docker` command that we use to check what it has needs to be one that actually talks to it, and the first `if` test here looks good to me for that.  So I'm thinking both of those tests need to check for both of those things.

edsantiago: Done, and refactored into a common helper. I also rebased.

tests/pull.bats
```diff
@@ -92,15 +92,8 @@ load helpers
...
-  run systemctl status docker
-  if [[ ! "$output" =~ "active (running)" ]]
-  then
-     skip "Skip the test as docker services is not running"
-  fi
+  skip_if_no_docker
 
-  run systemctl start docker
-  echo "$output"
-  [ "$status" -eq 0 ]
```
edsantiago, line 103: Judgment call: removed, because we've literally just checked that the service is active, and `start` != `restart`. If there is some hidden side effect or reason for running this `start`, please speak now.

rhatdan:
LGTM
/approve

rhatdan:
@nalind @vrothberg @giuseppe @TomSweeneyRedHat PTAL

nalind:
LGTM

PR #3172, opened 2021-04-21
Add system tests of 'bud' subcommand

hshiina:

#### What type of PR is this?

/kind other

Tests for the following flags:
- --cap-drop
- --runtime
- --runtime-flag
- --add-host
- --cgroup-parent
- --cpu-period
- --cpu-quota
- --cpu-shares
- --cpuset-cpus
- --cpuset-mems
- --isolation
- --memory
- --memory-swap
- --shm-size
- --ulimit

#### How to verify it
Run the system test: `bats tests/bud.bats`

None

```release-note
None
```


rhatdan:
@edsantiago PTAL

tests/bud.bats
```diff
@@ -2811,3 +2816,301 @@ _EOF
...
+}
+_EOF
+
+    run_buildah bud --runtime=crun --runtime-flag=debug --security-opt seccomp=${TESTDIR}/seccomp.json \
+                    -q -t alpine-bud-crun --signature-policy ${TESTSDIR}/policy.json --file ${mytmpdir} .
+    expect_output --substring "unknown seccomp syscall"
+  fi
+
+  if [ -z "${found_runtime}" ]; then
+    skip "Did not find 'runc' nor 'crun' in \$PATH - could not run this test!"
```
edsantiago, line 2871: It might be better to `die` here, not `skip`: if this ever triggers, it means there's a new runtime that we need to handle.

tests/bud.bats
```diff
@@ -2811,3 +2816,301 @@ _EOF
...
+
+  _prefetch alpine
+
+  mytmpdir=${TESTDIR}/my-dir
+  mkdir -p ${mytmpdir}
+
+  if is_cgroupsv2; then
+    cat > $mytmpdir/Containerfile << _EOF
+from alpine
+run cat /sys/fs/cgroup/\$(cat /proc/self/cgroup | awk -F : '{print \$NF}')/cpu.max
```
edsantiago, line 2932: one-argument `cat` is usually a bad idea. This would read better as:
```bash
run cat /sys/fs/cgroup/\$(awk -F: '{print \$NF}' /proc/self/cgroup)/cpu.max
```
(same applies to line 2962 and many other places)

tests/bud.bats
```diff
@@ -2811,3 +2816,301 @@ _EOF
...
+    cat > $mytmpdir/Containerfile << _EOF
+from alpine
+run printf "weight " && cat /sys/fs/cgroup/cpu/cpu.shares
+_EOF
+    result="weight ${shares}"
+  fi
+
+  run_buildah bud --cpu-shares=${shares} -t testcpu \
+                  --signature-policy ${TESTSDIR}/policy.json --file ${mytmpdir} .
+  expect_output --substring "${result}"
```
edsantiago, line 2975: The shorter the substring, the riskier it is to use `--substring`. How about:
```diff
@@ -2961,7 +2962,7 @@ _EOF
 from alpine
 run printf "weight " && cat /sys/fs/cgroup/\$(cat /proc/self/cgroup | awk -F : '{print \$NF}')/cpu.weight
 _EOF
-    result=$((1 + ((${shares} - 2) * 9999) / 262142))
+    result="weight $((1 + ((${shares} - 2) * 9999) / 262142))"
   else
     cat > $mytmpdir/Containerfile << _EOF
 from alpine
@@ -2972,7 +2973,7 @@ _EOF

   run_buildah bud --cpu-shares=${shares} -t testcpu \
                   --signature-policy ${TESTSDIR}/policy.json --file ${mytmpdir} .
-  expect_output --substring "${result}"
+  expect_output --from="${lines[2]}" "$result"
 }

 @test "bud with --cpuset-cpus" {
```
(you may also want to consider renaming `result` to `expect` or `expected_weight`, to make it easier for future maintainers)

tests/bud.bats
```diff
@@ -2811,3 +2816,301 @@ _EOF
...
+  mytmpdir=${TESTDIR}/my-dir
+  mkdir -p ${mytmpdir}
+  cat > $mytmpdir/Containerfile << _EOF
+from alpine
+run printf "cpuset-mems " && cat /sys/fs/cgroup/cpuset/cpuset.mems
+_EOF
+
+  run_buildah bud --cpuset-mems=0 -t testcpuset \
+                  --signature-policy ${TESTSDIR}/policy.json --file ${mytmpdir} .
+  expect_output --substring "cpuset-mems 0"
```
edsantiago, line 3022: Again, perhaps you could use `--from="${lines[2]}"`

tests/bud.bats
```diff
@@ -2811,3 +2816,301 @@ _EOF
...
+run printf "memory-max=" && cat /sys/fs/cgroup/memory/memory.limit_in_bytes
+run printf "memory-swap-result=" && cat /sys/fs/cgroup/memory/memory.memsw.limit_in_bytes
+_EOF
+    expect_swap=73400320
+  fi
+
+  run_buildah bud --memory=40m --memory-swap=70m -t testmemory \
+                  --signature-policy ${TESTSDIR}/policy.json --file ${mytmpdir} .
+  expect_output --substring "memory-max=41943040"
+  expect_output --substring "memory-swap-result=${expect_sw}"
```
edsantiago, line 3082: typo: the variable is actually `expect_swap`. This is passing because `${expect_sw}` is empty. See why I try to avoid `--substring` whenever possible? Please rewrite this to use `--from="${lines[...]}"` and exact matches.

edsantiago (changes requested):
Very nice work, but there are a couple of bugs and a few dangerous practices. I would encourage you to eliminate single-arg `cat` (minor) and to do everything you can to avoid `expect_output --substring`.

tests/bud.bats
```diff
@@ -2811,3 +2816,301 @@ _EOF
...
+  if [ -n "$(command -v runc)" ]; then
+    found_runtime=y
+    run_buildah ? bud --runtime=runc --runtime-flag=debug \
+                      -q -t alpine-bud-runc --signature-policy ${TESTSDIR}/policy.json --file ${mytmpdir} .
+    if [ "$status" -eq 0 ]; then
+      expect_output --substring "nsexec started"
+    else
+      # runc fully supports cgroup v2 (unified mode) since v1.0.0-rc93.
+      # older runc doesn't work on cgroup v2.
+      expect_output --substring "this version of runc doesn't work on cgroups v2" "should fail by unsupportability for cgroupv2"
```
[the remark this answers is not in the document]
hshiina: Thanks. I tried using `is_cgroupsv2`.

tests/bud.bats
```diff
@@ -2811,3 +2816,301 @@ _EOF
...
+  _prefetch alpine
+
+  mytmpdir=${TESTDIR}/my-dir
+  mkdir -p ${mytmpdir}
+cat > $mytmpdir/Containerfile << _EOF
+from alpine
+run cat /etc/hosts
+_EOF
+
+  run_buildah bud --add-host=localhost:127.0.0.1 -t testbud \
```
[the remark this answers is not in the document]
hshiina: I fixed. Thank you for the suggestion.

tests/bud.bats
hshiina: Done.

tests/bud.bats
hshiina: I fixed them. Thank you for the comment.

tests/bud.bats
hshiina: I fixed. Thank you for the advice.

tests/bud.bats
hshiina: Thanks.

tests/bud.bats
hshiina: Thank you for catching it! I agree that I should avoid using '--substring' as much as possible.

tests/bud.bats
```diff
@@ -2811,6 +2816,316 @@ _EOF
   ! expect_output --substring '\-\-build-arg NEWSECRET=<VALUE>'
 }
 
+@test "bud with --runtime and --runtime-flag" {
+  # This Dockerfile needs us to be able to handle a working RUN instruction.
```
TomSweeneyRedHat, line 2820: nit of a nit
```suggestion
  # This Containerfile needs us to be able to handle a working RUN instruction.
```

tests/bud.bats
```diff
@@ -2811,6 +2816,316 @@ _EOF
...
+from alpine
+run echo hello
+_EOF
+
+  local found_runtime=
+
+  if [ -n "$(command -v runc)" ]; then
+    found_runtime=y
+    if is_cgroupsv2; then
+      # The result with cgroup v2 depneds on the version of runc.
```
TomSweeneyRedHat, line 2838: nit
```suggestion
      # The result with cgroup v2 depends on the version of runc.
```

TomSweeneyRedHat:
LGTM
Happy green test buttons, would like @edsantiago to push the shiny merge button...

TomSweeneyRedHat:
and Thanks @hshiina, a really nice addition.

tests/bud.bats
```diff
@@ -2811,6 +2816,316 @@ _EOF
...
+  mkdir -p ${mytmpdir}
+  cat > $mytmpdir/Containerfile << _EOF
+from alpine
+run grep "myhostname" /etc/hosts
+_EOF
+
+  ip=123.45.67.$(( $RANDOM % 256 ))
+  run_buildah bud --add-host=myhostname:$ip -t testbud \
+                  --signature-policy ${TESTSDIR}/policy.json --file ${mytmpdir} .
+  expect_output --from="${lines[2]}" --substring "^$ip[^\S]+myhostname"
```
edsantiago, line 2897: Why `[^\S]`? The desired string is "ip address, whitespace, myhostname". What you have here is "ip address, anything except non-whitespace, myhostname". My brain has a very hard time with that. Did you deliberately avoid using what I suggested? Why do you find this form better?

tests/bud.bats
```diff
@@ -2811,6 +2816,316 @@ _EOF
...
+  mytmpdir=${TESTDIR}/my-dir
+  mkdir -p ${mytmpdir}
+  cat > $mytmpdir/Containerfile << _EOF
+from alpine
+run df -h /dev/shm
+_EOF
+
+  run_buildah bud --shm-size=80m -t testshm \
+                  --signature-policy ${TESTSDIR}/policy.json --file ${mytmpdir} .
+  expect_output --from="${lines[3]}" --substring "shm[^\S]+80.0M"
```
edsantiago, line 3111: Same thing.

edsantiago (changes requested):
Mostly LGTM, but the backslash-capital-S thing is really weird. I find it incomprehensible - but maybe there's something I've missed. If you have a strong argument in favor of it, could you please educate me? Otherwise I really have to encourage you to use the normal backslash-lower-case-s.

tests/bud.bats
hshiina: This happened by some trial and error in writing the regex. '/s' is correct as you told.

tests/bud.bats
hshiina: Thanks!

tests/bud.bats
hshiina: Thanks!

edsantiago (approved):

edsantiago:
/lgtm

Nice work, @hshiina, thank you!

PR #3173, opened 2021-04-21
Test namespace flags of 'bud' subcommand

hshiina:

#### What type of PR is this?

/kind other

Add 'bud' subcommand to the 'namespaces' system test, which already verifies namespace flags with 'from' and 'run' subcommands.

#### How to verify it
Run the system test: `bats tests/namespaces.bats`

None

```release-note
None
```

### DIFF
--- a/tests/namespaces.bats
+++ b/tests/namespaces.bats
@@ -206,12 +206,47 @@ load helpers
 
     # Check that a container with mapped-layer can be committed.
     run_buildah commit "$ctr" localhost/alpine-working:$i
+
+
+    # Also test bud command
+    # Build an image using these mappings.
+    echo "Building image with ${uidmapargs[$i]} ${gidmapargs[$i]}"
+    run_buildah bud ${uidmapargs[$i]} ${gidmapargs[$i]} $RUNOPTS --signature-policy ${TESTSDIR}/policy.json \
+                    -t localhost/alpine-bud:$i -f ${TESTSDIR}/bud/namespaces/Containerfile $TESTDIR
+    # If we specified mappings, expect to be in a different namespace by default.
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
/approve
@edsantiago PTAL
@nalind @vrothberg @TomSweeneyRedHat @cevich @umohnani8 @ashley-cui PTAL

tests/namespaces.bats
```diff
@@ -206,12 +206,47 @@ load helpers
 
     # Check that a container with mapped-layer can be committed.
     run_buildah commit "$ctr" localhost/alpine-working:$i
+
+
+    # Also test bud command
```
edsantiago, line 211: The `idmapping` test was unmaintainable even before this addition; it is now even more so. I like the addition, and it is well done, but I wonder if you could take some time to find a way to refactor this, perhaps by crafting one or more smaller helper functions?

cevich:
LGTM but I'm not an expert in these test.  @edsantiago will this new test need any special treatment for reuse in the 'bud' testing for podman?

edsantiago (commented):
LGTM at first pass, but this requires more than one pass: it's pretty complicated. ITM I have one request. It may turn out to be impossible or too difficult to refactor, that's ok, but I'd like to know that at least it has been considered.

edsantiago:
@cevich this will never run under podman. Podman only tests `bud.bats`, and these tests are in `namespaces.bats`

edsantiago:
Oh, *nice*, thank you! It'll take me a while to review in depth, but I fully intend to do so this morning!

edsantiago:
/lgtm

Nice work!

PR #3174, opened 2021-04-21
Add system test of 'buildah version' for packaged buildah

hshiina:

#### What type of PR is this?

/kind other

This test can be used for testing packaged buildah with setting 'BUILDAH_BINARY'.

#### How to verify it
Set a path to a buildah binary installed with a package manager to `BUILDAH_BINARY`.
Then, run the test: `bats version.bats`

None

```release-note
None
```


rhatdan:
Thanks @hshiina 
@edsantiago I would like to get your comment on this?  Seems like a potential decent system test

edsantiago:
Nice work! LGTM.

PR #3181, opened 2021-04-23
restore push/pull and util API

vrothberg:
Restore the push and pull API that commit dcd2a92e5669 removed.
These changes would break vendoring into openshift/builder due
to build errors.

For the same reason, restore `util.FindImage` and `util.AddImageNames`
but deprecate the `findRegistry` argument.

### DIFF
--- a/cmd/buildah/pull.go
+++ b/cmd/buildah/pull.go
@@ -1,32 +1,31 @@
 package main
 
 import (
-	"context"
 	"fmt"
 	"os"
 	"runtime"
 
-	"github.com/containers/buildah/pkg/blobcache"
+	"github.com/containers/buildah"
+	"github.com/containers/buildah/define"
 	buildahcli "github.com/containers/buildah/pkg/cli"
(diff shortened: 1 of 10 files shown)


cmd/buildah/pull.go
```diff
@@ -1,32 +1,31 @@
 package main
```
rhatdan, line 1: Why restore these commands, This is not part of the API that OpenShift uses, is it?

vrothberg: Functionally it's identical but letting the commands to continue using the API exercises them in CI.

vrothberg: I think that's the right middle ground of preserving the API but moving the code over to libimage.

rhatdan:
LGTM

rhatdan:
@nalind @giuseppe @umohnani8 @TomSweeneyRedHat PTAL

giuseppe (approved):
/lgtm

Issue #3153, opened 2021-04-16
Builds fail when doing `RUN` after `VOLUME` with empty storage driver


Vogtinator:
**Description**

After upgrading from podman 3.0.1 to 3.1.0, certain `Dockerfile`s/`Containerfile`s fail to build.
For some reason this is only the case when not specifying a storage driver in the config.
I could see this only with `podman build`, because `buildah bud` refuses to do anything without that config set.

Originally I saw this error when building in OBS, where the error message is actually much weirder:

```
[   80s] STEP 23: RUN obs-docker-support --uninstall
[   82s] time="2021-04-16T04:30:29Z" level=error msg="The storage 'driver' option must be set in /etc/containers/storage.conf, guarantee proper operation."
[   82s] [   70.671121] overlayfs: unrecognized mount option " guarantee proper operation."
[   82s] [   70.671121] /srv/www/htdocs" or missing value
[   82s] error running container: error creating container for [/bin/sh -c obs-docker-support --uninstall]: time="2021-04-16T04:30:30Z" level=error msg="container_linux.go:367: starting container process caused: process_linux.go:495: container init caused: rootfs_linux.go:60: mounting \"/var/lib/containers/storage/overlay-containers/0aef943f1188d994fa4ba7beb20e87c3ceb5efb2ab832bfd9babd8309eaef1ac/userdata/overlay/281018080/merge\" to rootfs at \"/var/tmp/buildah132004737/mnt/rootfs/time=\\\"2021-04-16T04:30:27Z\\\" level=error msg=\\\"The storage 'driver' option must be set in /etc/containers/storage.conf, guarantee proper operation.\\\"\\n/srv/www/htdocs\" caused: invalid argument"
[   82s] : exit status 1
[   82s] time="2021-04-16T04:30:30Z" level=error msg="unable to write build event: \"write unixgram @00003->/run/systemd/journal/socket: sendmsg: no such file or directory\""
[   82s] Error: error building at STEP "RUN obs-docker-support --uninstall": error while running runtime: exit status 1
```

Which looks like it somehow passed part of the error message into overlayfs mount options? I'm not actually sure whether the issue I can reproduce locally is even the same.

Edit: I should clarify that I removed that option from my local `storage.conf` to reproduce the OBS problem

**Steps to reproduce the issue:**

```
f241:/tmp # cat Dockerfile
FROM scratch
VOLUME /srv/www/htdocs
RUN nothing
f241:/tmp # podman build -t h .
[...]

PR #3184, opened 2021-04-23
imagebuildah: drop resolveSymlink()

nalind:
#### What type of PR is this?

/kind bug

Drop a bespoke `resolveSymlink()` function and replace it with a call to `copier.Eval()`.

Saving a volume cache using VFS had the side-effect of creating the volume's directory, so have the overlay path for it do so as well.  Have the VFS path add transient mounts to the locations in the rootfs to keep buildah.Run() from doing any saving/restoring of volume contents on its own.

Fix a duplicate test name that newer bats would complain about.

#### How to verify it

`resolveSymlink()` was interpreting all of its child process's output as the resolved value of the path it was given, which caused #3153.

```
None
```


rhatdan:
LGTM But tests not happy.

umohnani8:
LGTM

nalind:
Had some scope creep here, but tests are passing now.

imagebuildah/stage_executor.go
```diff
@@ -102,16 +102,24 @@ func (s *StageExecutor) Preserve(path string) error {
...
+			return errors.Wrapf(err, "making evaluated path %q relative to %q", evaluated, s.mountPoint)
+		}
+		if strings.HasPrefix(symLink, ".."+string(os.PathSeparator)) {
+			return errors.Errorf("evaluated path %q was not below %q", evaluated, s.mountPoint)
+		}
+		archivedPath = evaluated
+		path = string(os.PathSeparator) + symLink
 	} else {
-		return errors.Wrapf(err, "error reading symbolic link to %q", path)
+		return errors.Wrapf(err, "error evaluating path %q", path)
```
rhatdan, line 116: I worry about a stutter here.  But ...

tests/conformance/conformance_test.go
```diff
@@ -1451,15 +1451,18 @@ var internalTestCases = []testCase{
 		contextDir:      "mount",
 		buildahRegex:    "/tmp/test/file.*?regular file.*?/tmp/test/file2.*?regular file",
 		withoutDocker:   true,
-		transientMounts: []string{"@@TEMPDIR@@:/tmp/test"},
+		transientMounts: []string{"@@TEMPDIR@@:/tmp/test" + selinuxMountFlag()},
```
rhatdan, line 1454: I believe the :Z is already ignored on systems with SELinux disabled, so this code is probably not necessary.

PR #3187, opened 2021-04-28
runUsingRuntime: when relaying error from the runtime, mention that it's from the runtime

nalind:
#### What type of PR is this?

/kind other

Mention that errors we relay from the runtime are from the runtime, to hopefully reduce confusion when troubleshooting.

#### How to verify it

This is changing the text of errors, no logic changes.

None

```
None
```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -829,7 +829,7 @@ func runUsingRuntime(isolation define.Isolation, options RunOptions, configureNe
 	logrus.Debugf("Running %q", create.Args)
 	err = create.Run()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


rhatdan:
Nice improvement.  LGTM

TomSweeneyRedHat:
LGTM

edsantiago:
How did this pass CI without adding tests? I see a weird error in [validate](https://cirrus-ci.com/task/6228741434114048?logs=validate_test#L64) but the step is green, and the error is useless, so I can't tell if it's an error in my script, or in some unrelated part of Cirrus:
```
[+0088s] ./tests/validate/pr-should-include-tests
[+0088s] fatal: Not a valid object name master
```

PR #3190, opened 2021-04-29
pr-should-include-tests: try to make work in buildah

edsantiago:
Looks like this script wasn't working in the buildah environment,
apparently because we're in a shallow git checkout without a
master branch? Let's see if we can use $GITVALIDATE_EPOCH
instead.

And, add a sanity check so we catch this if it happens again.


edsantiago:
Smoke test passed, which means we at least made it farther than before.

TomSweeneyRedHat:
LGTM and :crossed_fingers: the test system will think the same

PR #3192, opened 2021-04-29
fix run uid flake

edsantiago:
The run-user test creates a test user with random UID and GID.
If the randomly-picked UID matches that of an existing account,
the test will flake because the account's default group will
almost certainly not match the one we generated.

Solution: only pick UIDs and GIDs that are not already present
in the container image.

### DIFF
--- a/tests/run.bats
+++ b/tests/run.bats
@@ -153,6 +153,20 @@ load helpers
 	expect_output "/tmp" "invalid cmd & entrypoint, pwd"
 }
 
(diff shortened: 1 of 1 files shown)


edsantiago:
Sorry, I screwed up my git tree and this includes #3190, which should merge first please

rhatdan:
LGTM, once you have the PR Cleaned up.

edsantiago:
Done - this PR is now clean

vrothberg (commented):
/lgtm

PR #3201, opened 2021-05-04
imagebuildah: ignore signatures when tagging images

nalind:
#### What type of PR is this?

/kind bug

We re-tag images produced for Dockerfiles which contain no instructions by "copying" them from their current names to their destination names, letting the lower storage libraries deduplicate them into the same image record.

When the source included signatures, which can happen when the base image included signatures, this would break because containers-storage can't guarantee that the compressed version of the blob it will produce for a given layer will have the same digest that the version referenced by the manifest had, so the image library would refuse to "copy" them.

When the source and destination are the same, though, the `RemoveSignatures` option doesn't cause the signatures to be deleted, but it does bypass that check in the image library, so toggling it on works around the problem.

#### How to verify it

Tests should continue to pass when base images come from registries which have signature storage URLs ("sigstore") configured for them.

None

Should fix https://bugzilla.redhat.com/show_bug.cgi?id=1955166.

```
None
```

### DIFF
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -21,11 +21,18 @@ function setup() {
     suffix=$(dd if=/dev/urandom bs=12 count=1 status=none | od -An -tx1 | sed -e 's, ,,g')
     TESTDIR=${BATS_TMPDIR}/tmp${suffix}
     rm -fr ${TESTDIR}
-    mkdir -p ${TESTDIR}/{root,runroot}
-
+    mkdir -p ${TESTDIR}/{root,runroot,sigstore,registries.d}
+    echo "default-docker:                                                           " >> ${TESTDIR}/registries.d/default.yaml
+    echo "  sigstore-staging: file://${TESTDIR}/sigstore                            " >> ${TESTDIR}/registries.d/default.yaml
+    echo "docker:                                                                   " >> ${TESTDIR}/registries.d/default.yaml
+    echo "  registry.access.redhat.com:                                             " >> ${TESTDIR}/registries.d/default.yaml
+    echo "    sigstore: https://access.redhat.com/webassets/docker/content/sigstore " >> ${TESTDIR}/registries.d/default.yaml
+    echo "  registry.redhat.io:                                                     " >> ${TESTDIR}/registries.d/default.yaml
+    echo "    sigstore: https://registry.redhat.io/containers/sigstore              " >> ${TESTDIR}/registries.d/default.yaml
     # Common options for all buildah and podman invocations
     ROOTDIR_OPTS="--root ${TESTDIR}/root --runroot ${TESTDIR}/runroot --storage-driver ${STORAGE_DRIVER}"
-    REGISTRY_OPTS="--registries-conf ${TESTSDIR}/registries.conf"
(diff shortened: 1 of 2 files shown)


tests/helpers.bash
```diff
@@ -148,7 +155,7 @@ function run_buildah() {
 
         # stdout is only emitted upon error; this echo is to help a debugger
         echo "\$ $BUILDAH_BINARY $*"
-        run timeout --foreground --kill=10 $BUILDAH_TIMEOUT ${BUILDAH_BINARY} ${REGISTRY_OPTS} ${ROOTDIR_OPTS} "$@"
+        run timeout --foreground --kill=10 $BUILDAH_TIMEOUT ${BUILDAH_BINARY} ${BUILDAH_REGISTRY_OPTS} ${ROOTDIR_OPTS} "$@"
```
edsantiago, line 158: Thinking ahead: some day this will be vendored into podman, which will then run the buildah-bud-under-podman tests. How do you foresee making those podman tests work with this new scheme? Do you think podman needs a (hidden) `--registries-conf-dir` option? Will some larger subset of tests need to be skipped now?

vrothberg: Since Buildah and (soon) Podman both share `common/libimage` we could use an env variable rather than a CLI flag. Unless explicitly set [libimage](https://github.com/containers/common/blob/master/libimage/runtime.go#L27) will look for the CONTAINERS_REGISTRIES_CONF and REGISTRIES_CONFIG_PATH.

vrothberg: Sorry, I didn't check the code in question.  In this specific case, libimage isn't executed.

rhatdan: Yes I think we add the hidden option to Podman.

rhatdan:
LGTM
@vrothberg @edsantiago @TomSweeneyRedHat PTAL

tests/helpers.bash
```diff
@@ -21,11 +21,18 @@ function setup() {
...
-    mkdir -p ${TESTDIR}/{root,runroot}
-
+    mkdir -p ${TESTDIR}/{root,runroot,sigstore,registries.d}
+    echo "default-docker:                                                           " >> ${TESTDIR}/registries.d/default.yaml
+    echo "  sigstore-staging: file://${TESTDIR}/sigstore                            " >> ${TESTDIR}/registries.d/default.yaml
+    echo "docker:                                                                   " >> ${TESTDIR}/registries.d/default.yaml
+    echo "  registry.access.redhat.com:                                             " >> ${TESTDIR}/registries.d/default.yaml
+    echo "    sigstore: https://access.redhat.com/webassets/docker/content/sigstore " >> ${TESTDIR}/registries.d/default.yaml
+    echo "  registry.redhat.io:                                                     " >> ${TESTDIR}/registries.d/default.yaml
+    echo "    sigstore: https://registry.redhat.io/containers/sigstore              " >> ${TESTDIR}/registries.d/default.yaml
```
edsantiago, line 31: I think this would be more readable as:
```bash
    cat >${TESTDIR}/registries.d/default.yaml <<EOF
default-docker:
....
EOF
```
...but it's not worth resubmitting for.

edsantiago (approved):
LGTM but everyone please be aware that the next vendor-in-podman will break (because of changes to `run_buildah` - the podman diffs will not apply cleanly any more). If for some reason I'm not around to help, @vrothberg has had experience running through the process. I really wish I could make it easier.

PR #3209, opened 2021-05-06
[CI:DOCS] Notification email for cirrus-cron build failures

cevich:
This duplicates the github-action workflow of the same name/purpose from
the podman reposiory.  Every day it will check the most recent list
of Cirrus-cron initiated builds, and only send an e-mail if any failed.

Some measures are taken in this version to re-use a script and some
data from the podman repo.  There is no way (as of this commit) to
avoid duplicating the workflow file.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -135,7 +135,7 @@ vendor_task:
 unit_task:
     name: 'Unit tests w/ $STORAGE_DRIVER'
     alias: unit
-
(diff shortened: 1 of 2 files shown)


cevich:
note to me: Need to setup the required secret env. vars.

TomSweeneyRedHat:
LGTM

cevich:
First manual run for ref: https://github.com/containers/buildah/actions/runs/836998433

PR #3210, opened 2021-05-06
Fix race copying suddenly inaccessible symlinked files

cevich:
Fixes https://github.com/containers/podman/issues/9997

If a symlink in the copy-context points at a file which becomes
inaccessible during examination, an error is thrown.  Since the copy
function de-references symlinks later on, and (clearly) there would be a
problem handling the inaccessible file, we can simply throw a warning
about it and move on.


cevich:
This fix is wrong.

Issue #3113, opened 2021-03-29
flake: open state file /proc/3696968/stat: No such process


edsantiago:
This is a placeholder. There is not enough information here to make it a valid issue; I am filing because I'm pretty sure I've seen this flake before, in CI or gating tests, and didn't bother to file an issue, and now I regret it.

The log below was seen in podman testing, not buildah, but I'm filing under buildah anyway.
```console
$ podman build [...] -t ubuntu-image tests/bud/dest-symlink-dangling
STEP 1: FROM ubuntu
STEP 2: RUN mkdir /symlink
STEP 3: RUN ln -s /symlink /src && rm -rf /symlink
open state file /proc/3696968/stat: No such process
open pidfd: No such process
error running container: error reading container state (got output: ""): exit status 1
Error: error building at STEP "RUN ln -s /symlink /src && rm -rf /symlink": error while running runtime: exit status 1
[ rc=125 (** EXPECTED 0 **) ]
```
I cannot reproduce. I don't know if it's related to the ubuntu image, or the `ln` command, or if those are red herrings. If/when I see this again, I will add details.

edsantiago:
Found it! The [other recent instance I remember](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6479860211122176/html/bud-podman-fedora-33-root-host.log.html#t--00097):
```
[+0339s] not ok 97 bud with COPY of single file creates absolute path with correct permissions
         # (from function `die' in file tests/helpers.bash, line 216,
         #  from function `run_buildah' in file tests/helpers.bash, line 203,
         #  in test file tests/bud.bats, line 1276)
         #   `run_buildah bud --signature-policy ${TESTSDIR}/policy.json -t ${imgName} ${TESTSDIR}/bud/copy-create-absolute-path' failed with status 125
         # /var/tmp/go/src/github.com/containers/podman/test-buildah-v1.20.0/tests /var/tmp/go/src/github.com/containers/podman/test-buildah-v1.20.0
         # # [checking for: ubuntu]
         # # [restoring from cache: /var/tmp/buildah-image-cache.4029 / ubuntu]
         # Getting image source signatures
         # Copying blob sha256:f6291d8887317896606687305f6661ecfd884bc4e4523d948c39af71d3cf3a5d
         # Copying blob sha256:9e0775ca9a2f3180f9cd3ca1532446a8d6dc58f4970ab9babd6168fc117ddee7
[...]

PR #3212, opened 2021-05-07
Fix race condition

rhatdan:
Ed has found situations where the container exits, before we can check
the state causing a failure, where I think we can complete successfully.

Fixes: https://github.com/containers/buildah/issues/3113

[NO TESTS NEEDED] since I have no way to generate this race condition.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -851,6 +851,7 @@ func runUsingRuntime(isolation define.Isolation, options RunOptions, configureNe
 	if err != nil {
 		return 1, errors.Wrapf(err, "error parsing pid %s as a number", string(pidValue))
 	}
+	stopped := false
 	var reaping sync.WaitGroup
 	reaping.Add(1)
 	go func() {
@@ -861,6 +862,7 @@ func runUsingRuntime(isolation define.Isolation, options RunOptions, configureNe
 			wstatus = 0
 			logrus.Errorf("error waiting for container child process %d: %v\n", pid, err)
 		}
+		stopped = true
 	}()
 
 	if configureNetwork {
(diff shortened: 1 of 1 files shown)


rhatdan:
@nalind PTAL, does this make sense to you?

nalind:
Have we gotten errors that match it ("no such process")?  I'd expect that perhaps if the runtime's `state' command exited out from under us and was reaped by someone else, though the waitpid(2) man page suggests we'd get an ECHILD ("no child processes") for cases like that.
The errors in #3113 look like they're being relayed by us when printed by the runtime, which is exiting with status 1, so I don't think it'd hit the code that's being added.  Browsing the library sources, the "exit status %d" text looks more like what we'd get from formatting an os.ExitError for display.

rhatdan:
@nalind @giuseppe @edsantiago PTAL.  I am now setting the stopped flag when pid1 of the container exits.  Would this solve the problem?

edsantiago:
LGTM in principle. You'll need to re-push to get [NO TESTS NEEDED] to trigger.

giuseppe:
the issue could be in the sequence:

`kill -9 $CONTAINER_PID; $RUNTIME delete $CONTAINER`

The container could still be alive by the time we call `$RUNTIME delete` as the KILL signal was not handled yet.  In this case the `delete` would fail.

At least runc and crun wait for the process to exit before the `kill` operation is terminated, so in this case we are sure we are not introducing a new race condition, but in general a OCI runtime could handle the `kill` command in a different way and don't wait that the container terminates

giuseppe (approved):
sorry nevermind my last comment, I got confused.

LGTM

Issue #3214, opened 2021-05-05
podman-remote build, with missing ARG: warning displayed on server, not client


edsantiago:
In window 1 (actual separate window, don't just background it in the same window as the client):
```console
# podman system service --timeout=0
```
In window 2:
```console
# cat Containerfile
FROM alpine
ARG FOO
# ../bin/podman-remote build .
[normal output, but:]
```
back to window 1 (server):
```
WARN[0041] missing "FOO" build argument. Try adding "--build-arg FOO=<VALUE>" to the command line
```
That diagnostic does not do any good on the server end. It should be displayed on the client. (In fact, there is a buildah-bud test that expects this, and is failing when tested under podman).

nalind:
Hmm, actually it might be cleaner for the service to use `logrus.SetOutput()` to point logrus's standard logger to one stream or another.  That way we wouldn't have a library (buildah) messing with global settings for another library (logrus) in ways that a package that consumes them both (e.g., podman) might not be expecting.

nalind:
Yeah, if you want logrus messages from _everywhere_ (including libraries that buildah also depends on) logged somewhere other than the global defaults, I'm thinking `logrus.SetOutput()` is necessary and your only viable option.

rhatdan:
This is going to be a very difficult problem in a client server mode.  Since logrus is shared across the process,  We can't modify it's settings since we could have two or more remote image builds going on at the same time.  I think we will need to wire in the buildoptions down the stack and stop using logrus or at least also write the messages to the buildahopts.Err.  This is a lot more then just this one warn, since there are lots of calls to logrus, inside of run_linux.go.

nalind:
And in the common, image, and storage libraries.

PR #3215, opened 2021-05-07
Send logrus messages back to caller when building

rhatdan:
We want Info, Warning and Debug logrus messages to be writen to the
buildah stderr. this way when podman-remote is using build, it will
get the messages back on the client side.

[NO TESTS NEEDED] Since this will be tested in Podman.

Fixes: https://github.com/containers/buildah/issues/3214

### DIFF
--- a/buildah.go
+++ b/buildah.go
@@ -89,6 +89,9 @@ const (
 type Builder struct {
 	store storage.Store
 
+	// Logger is the logrus logger to write log messages with
+	Logger *logrus.Logger `json:"-"`
+
 	// Args define variables that users can pass at build-time to the builder
 	Args map[string]string
 	// Type is used to help identify a build container's metadata.  It
@@ -395,6 +398,7 @@ func OpenBuilder(store storage.Store, container string) (*Builder, error) {
 	}
 	b.store = store
 	b.fixupConfig()
+	b.setupLogger()
 	return b, nil
 }
 
@@ -430,6 +434,7 @@ func OpenBuilderByPath(store storage.Store, path string) (*Builder, error) {
 		if err == nil && b.Type == containerType && builderMatchesPath(b, abs) {
 			b.store = store
 			b.fixupConfig()
+			b.setupLogger()
 			return b, nil
 		}
 		if err != nil {
@@ -465,6 +470,7 @@ func OpenAllBuilders(store storage.Store) (builders []*Builder, err error) {
 		err = json.Unmarshal(buildstate, &b)
 		if err == nil && b.Type == containerType {
 			b.store = store
+			b.setupLogger()
 			b.fixupConfig()
 			builders = append(builders, b)
 			continue
--- a/config.go
+++ b/config.go
@@ -3,6 +3,7 @@ package buildah
 import (
 	"context"
 	"encoding/json"
+	"os"
 	"runtime"
 	"strings"
 	"time"
@@ -85,6 +86,7 @@ func (b *Builder) initConfig(ctx context.Context, img types.Image) error {
 		}
 	}
 
+	b.setupLogger()
 	b.fixupConfig()
 	return nil
 }
@@ -111,6 +113,14 @@ func (b *Builder) fixupConfig() {
 	}
 	if b.Format == define.Dockerv2ImageManifest && b.Hostname() == "" {
 		b.SetHostname(stringid.TruncateID(stringid.GenerateRandomID()))
+	}
+}
+
+func (b *Builder) setupLogger() {
+	if b.Logger == nil {
+		b.Logger = logrus.New()
+		b.Logger.SetOutput(os.Stderr)
+		b.Logger.SetLevel(logrus.GetLevel())
 	}
 }
 
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -50,6 +50,14 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 	if len(paths) == 0 {
 		return "", nil, errors.Errorf("error building: no dockerfiles specified")
 	}
+	logger := logrus.New()
+	if options.Err != nil {
+		logger.SetOutput(options.Err)
+	} else {
+		logger.SetOutput(os.Stderr)
+	}
+	logger.SetLevel(logrus.GetLevel())
+
 	var dockerfiles []io.ReadCloser
 	defer func(dockerfiles ...io.ReadCloser) {
 		for _, d := range dockerfiles {
@@ -140,7 +148,7 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 		return "", nil, errors.Wrapf(err, "error parsing main Dockerfile: %s", dockerfiles[0])
 	}
 
-	warnOnUnsetBuildArgs(mainNode, options.Args)
+	warnOnUnsetBuildArgs(logger, mainNode, options.Args)
 
 	for _, d := range dockerfiles[1:] {
 		additionalNode, err := imagebuilder.ParseDockerfile(d)
@@ -149,7 +157,7 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 		}
 		mainNode.Children = append(mainNode.Children, additionalNode.Children...)
 	}
-	exec, err := NewExecutor(store, options, mainNode)
+	exec, err := NewExecutor(logger, store, options, mainNode)
 	if err != nil {
 		return "", nil, errors.Wrapf(err, "error creating build executor")
 	}
@@ -173,7 +181,7 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 	return exec.Build(ctx, stages)
 }
 
-func warnOnUnsetBuildArgs(node *parser.Node, args map[string]string) {
+func warnOnUnsetBuildArgs(logger *logrus.Logger, node *parser.Node, args map[string]string) {
 	argFound := make(map[string]bool)
 	for _, child := range node.Children {
 		switch strings.ToUpper(child.Value) {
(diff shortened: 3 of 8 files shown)


jwhonce (changes requested):
`fmt.Fprintf(Err, "Warning: missing %q build argument. Try adding %q to the command line\n", argName, fmt.Sprintf("--build-arg %s=<VALUE>", argName))`

Please add newline at end of message

nalind:
This doesn't cover all of the places in the main `buildah` package, and the other packages in this repository that it uses, where we can also call `logrus` during `imagebuildah.BuildDockerfiles()`, and it doesn't cover any of the libraries that we depend on.  I think we need to document that our caller needs to call `logrus.SetOutput()` and leave it at that.

rhatdan:
So fix this in Podman.

jwhonce:
@nalind The problem with logrus.SetOutput() is we don't want all the debugging information forwarded to the client. We only want expected error messages.

jwhonce:
LGTM

rhatdan:
Is there a way in logrus, to be selective about messages, IE Error and Warn go to buildOpts.Err

nalind:
They all go to the same location (`logrus.SetOutput()` doesn't take a logging level argument).  The only discriminator appears to be the logging level.

> @nalind The problem with logrus.SetOutput() is we don't want all the debugging information forwarded to the client. We only want expected error messages.

Why doesn't the client want to know about other warnings and errors?

rhatdan:
https://github.com/sirupsen/logrus/issues/403
Has been talking about this for years.  @jwhonce Is worried about Debug, Trace and Info.

jwhonce:
@nalind, Client/Server reporting is a matrix. Buildah is a little different because of the Stdout/StdErr/Auxout/Reporting BuildOptions. Then we have messages the developers have asked for that were rendered via logrus. Logrus output has been captured by the service and kept on the server. The handlers are responsible for relaying information to the client loosely following the table below.

Logging levels:

Errors: Always Client and Server
Warnings: Usually Client and Always Server
Info: Maybe Client and Usually not Server
Debug: Never Client and Usually not Server
Trace: Never Client and Usually not Server

nalind:
The standard logger in the logrus package a globally-shared object, and it is used not only by buildah, but by other dependencies that both buildah and libpod use, and by libpod itself.  That is not something your application wants one library in your dependency stack messing without your full control and knowledge.  podman is going to have to manipulate the global standard logger that the logrus library exports to get the results that you're looking for.

jwhonce:
@nalind The global nature of the logger is the issue. If two remote clients are running builds, it is impossible to route any logged message to the client which called imagebuildah.BuildDockerfiles(). Any logrus output has to stay on the server. imagebuildah.BuildDockerfiles() has four writers already associated with it, that is what we are leveraging here.

imagebuildah/build.go
```diff
@@ -181,7 +181,7 @@ func warnOnUnsetBuildArgs(node *parser.Node, args map[string]string) {
 				argHasValue = argFound[argName]
 			}
 			if _, ok := args[argName]; !argHasValue && !ok {
-				logrus.Warnf("missing %q build argument. Try adding %q to the command line", argName, fmt.Sprintf("--build-arg %s=<VALUE>", argName))
+				fmt.Fprintf(Err, "Warning: missing %q build argument. Try adding %q to the command line\n", argName, fmt.Sprintf("--build-arg %s=<VALUE>", argName))
```
nalind, line 184: Something like (untested)
```
log := logrus.New()
log.SetOutput(Err)
log.SetLevel(logrus.GetLevel())
log.Warnf(...)
```
would probably be less disruptive.

nalind:
Well, so long as we're all aware that this one change isn't going to fix the general case, I guess I've grumbled enough.

imagebuildah/build.go
rhatdan: If we could do this one per connection, then this would solve our problem.

nalind: We can try to wire in a way to pass a logger to some of our functions, but I don't really think we're going to be able to add a "send logrus output here" parameter to every function that logs things in the packages and libraries that we call into, and that's what'd be necessary to be sure we were catching everything.  Not that I know how our caller would filter and route things if all it could control was the output `io.Writer`.

rhatdan: I say we do our best.  So that most of these messages get back to the caller.

rhatdan:
@jwhonce @nalind I rewrote this to pass around the logger Object. so that podman-remote will get all Info, Warning and Errors out of server side back to client.

buildah.go
```diff
@@ -89,6 +89,9 @@ const (
 type Builder struct {
 	store storage.Store
 
+	// Logger is the logrus loger to write log messages with
```
nalind, line 92:
```suggestion
	// Logger is the logrus logger to write log messages with
```

config.go
```diff
@@ -112,6 +113,11 @@ func (b *Builder) fixupConfig() {
 	if b.Format == define.Dockerv2ImageManifest && b.Hostname() == "" {
 		b.SetHostname(stringid.TruncateID(stringid.GenerateRandomID()))
 	}
+	if b.Logger == nil {
+		b.Logger = logrus.New()
+		b.Logger.SetOutput(os.Stderr)
+		b.Logger.SetLevel(logrus.GetLevel())
+	}
```
nalind, line 120: This snippet belongs somewhere other than in a method called `fixupConfig()`.  It's also redundant for callers who get here from `newBuilder()`.

imagebuildah/executor.go
```diff
@@ -300,6 +302,7 @@ func (b *Executor) startStage(ctx context.Context, stage *imagebuilder.Stage, st
 		volumeCacheInfo: make(map[string]os.FileInfo),
 		output:          output,
 		stage:           stage,
+		logger:          b.logger,
```
nalind, line 305: This field isn't needed, since it's never different from the Executor's logger.  Use the StageExecutor's `executor.logger` in those places instead, similar to how we handle several other fields that we refer to in the Executor.

run.go
```diff
@@ -70,6 +71,8 @@ const (
 
 // RunOptions can be used to alter how a command is run in the container.
 type RunOptions struct {
+	// Logger is the logrus loger to write log messages with
```
nalind, line 74:
```suggestion
	// Logger is the logrus logger to write log messages with
```

run_linux.go
```diff
@@ -76,7 +76,7 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	logrus.Debugf("using %q to hold bundle data", path)
 	defer func() {
 		if err2 := os.RemoveAll(path); err2 != nil {
-			logrus.Error(err2)
+			b.Logger.Error(err2)
```
nalind, line 79:
```suggestion
			options.Logger.Error(err2)
```
Also needs to create a logger for `options.Logger` if it isn't set.

run_linux.go
```diff
@@ -120,7 +120,7 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	}
 	defer func() {
 		if err := b.Unmount(); err != nil {
-			logrus.Errorf("error unmounting container: %v", err)
+			b.Logger.Errorf("error unmounting container: %v", err)
```
nalind, line 123:
```suggestion
			options.Logger.Errorf("error unmounting container: %v", err)
```

run_linux.go
```diff
@@ -253,7 +253,7 @@ rootless=%d
 
 	defer func() {
 		if err := cleanupRunMounts(runMountTargets, mountPoint); err != nil {
-			logrus.Errorf("unabe to cleanup run mounts %v", err)
+			b.Logger.Errorf("unabe to cleanup run mounts %v", err)
```
nalind, line 256:
```suggestion
			options.Logger.Errorf("unabe to cleanup run mounts %v", err)
```

run_linux.go
```diff
@@ -704,12 +704,14 @@ func runUsingRuntime(isolation define.Isolation, options RunOptions, configureNe
 	// Lock the caller to a single OS-level thread.
 	runtime.LockOSThread()
 
+	logger := options.Logger
```
nalind, line 707: This never needs to be different from `options.Logger`, so drop the variable and use the passed-in value.

run_linux.go
```diff
@@ -1069,6 +1071,7 @@ func setupRootlessNetwork(pid int) (teardown func(), err error) {
 
 func runConfigureNetwork(isolation define.Isolation, options RunOptions, configureNetworks []string, pid int, containerName string, command []string) (teardown func(), err error) {
 	var netconf, undo []*libcni.NetworkConfigList
+	logger := options.Logger
```
nalind, line 1074: This doesn't need to ever be different from `options.Logger`, and it's used once.  Remove the variable and reference `options.Logger` directly.

nalind:
Embedding another package's API objects in our API means that any stability guarantees we make for our API are also dependent on the stability of their APIs, and this one ties us to a specific implementation rather than an interface.  I understand the motivation for this patch, though, so I'm not rejecting it out of hand, but it remains a concern.

run_linux.go
rhatdan: That is the way I originally had it, and I thought this was cleaner, but I will revert.

config.go
rhatdan: Where would you want me to put it.  fixupConfig is called by both newbuilder and openbuilder. I could make a new function? And call it from both?

rhatdan: OpenBuilder is called from buildah run ...

nalind: It's either going to be do-it-everywhere, or a helper function.  Builder objects are also constructed by the variations of OpenBuilder, and by ImportBuilder/ImportBuilderFromImage.  Basically, anywhere you see a new Builder object being created, we're going to need to set this field.

imagebuildah/build.go
```diff
@@ -190,7 +198,7 @@ func warnOnUnsetBuildArgs(node *parser.Node, args map[string]string) {
 				argHasValue = argFound[argName]
 			}
 			if _, ok := args[argName]; !argHasValue && !ok {
-				logrus.Warnf("missing %q build argument. Try adding %q to the command line", argName, fmt.Sprintf("--build-arg %s=<VALUE>", argName))
+				logger.Warnf("missing %q build argument. Try adding %q to the command line\n", argName, fmt.Sprintf("--build-arg %s=<VALUE>", argName))
```
nalind, line 201: The `\n` was for `fmt.Fprintf()`, which doesn't add one by default, but `logrus`'s formatters always include it, so we don't need that.

PR #3220, opened 2021-05-11
Add and use a "copy" helper instead of podman load/save

nalind:
#### What type of PR is this?

/kind failing-test 

Instead of using podman to cache images during integration tests, use a custom helper.  This will avoid unnecessary format conversions which I think are breaking gating tests in https://bugzilla.redhat.com/show_bug.cgi?id=1955166.

#### How to verify it

```
None
```


vrothberg:
Would `buildah pull FOO`, `buildah push FOO docker-archive:/tmp/foo` work?

nalind:
> Would `buildah pull FOO`, `buildah push FOO docker-archive:/tmp/foo` work?

Temporarily, probably.  If, like with oci-archive, saving to a docker-archive ever forces the image to be converted to a particular format that it wasn't already in, it'll just trigger the same failure, so it's not a good long term option.

vrothberg:
> > Would `buildah pull FOO`, `buildah push FOO docker-archive:/tmp/foo` work?
> 
> Temporarily, probably. If, like with oci-archive, saving to a docker-archive ever forces the image to be converted to a particular format that it wasn't already in, it'll just trigger the same failure, so it's not a good long term option.

In that case, `dir:` may be a good option.  I'm mostly wondering if we can recycle what Buildah can do.

nalind:
> In that case, `dir:` may be a good option. I'm mostly wondering if we can recycle what Buildah can do.

`buildah pull`'s attempt to massage the path name in the `dir:` reference into an image name, rather than letting us specify it directly, would complicate that.

vrothberg:
Thanks for your patience :) LGTM

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
/lgtm
/hold
until tests complete

nalind:
Hmm, something's not right here.  Even though we load the cached images, we're still pulling them, so integration tests are taking a lot longer to run.

nalind:
I think that the `dir:` round trip is causing the current version of `HasDifferentDigest()` to conclude that the restored-from-cache image, which doesn't include a copy of the original manifest list that `buildah pull` saves, doesn't match the remote reference that points to a manifest list.  As a result, we end up pulling images more often than we did before this patch, which is why the integration tests are timing out.  Opened https://github.com/containers/common/pull/554 to tighten up the logic in `HasDifferentDigest()`, which I think will fix that.

nalind:
Hmm, the integration test are still taking too long (one's over 80 minutes and still going as I write this), so still not right.

nalind:
Added a known `containers.conf`, based on the Fedora 34 version, so that the default list of capabilities in a system copy of the file doesn't affect tests.

nalind:
Reverted overwriting containers.conf, though tests that implicitly used the system containers.conf need tweaking.

nalind:
Added a "buildah rm -a" between a "buildah from" and a "buildah rmi" without a "-f" in the "from encrypted registry image" test.

nalind:
Rebased.

vrothberg (commented):
/lgtm

PR #3222, opened 2021-05-12
buildah bud: fix containers/podman/issues/10307

vrothberg:
Fix a regression in `buildah bud`.  In case of a cache hit, the pull
policy should be set to "never" so that we enforce a local-only lookup
without even attempting to reach out to a registry.

The regression was detected in Podman which defaults to --layers=true
which ultimately broke when setting --pull or --pull-always.  In case of
a cache hit, the image reference refers to a local image which conflicts
with an "always" pull policy.

Also extend an existing bud test to make sure we don't regress on it in
the future.

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -500,7 +500,7 @@ func (s *StageExecutor) UnrecognizedInstruction(step *imagebuilder.Step) error {
 // prepare creates a working container based on the specified image, or if one
 // isn't specified, the first argument passed to the first FROM instruction we
 // can find in the stage's parsed tree.
-func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBConfig, rebase bool) (builder *buildah.Builder, err error) {
+func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBConfig, rebase bool, pullPolicy define.PullPolicy) (builder *buildah.Builder, err error) {
 	stage := s.stage
 	ib := stage.Builder
 	node := stage.Node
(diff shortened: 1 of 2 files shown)


vrothberg:
@nalind @rhatdan PTAL

imagebuildah/stage_executor.go
```diff
@@ -1017,7 +1017,11 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			// creating a new working container with the
 			// just-committed or updated cached image as its new
 			// base image.
-			if _, err := s.prepare(ctx, imgID, false, true); err != nil {
+			//
+			// Enfore pull "never" since we already have an image
```
nalind, line 1021:
```suggestion
			// Enforce pull "never" since we already have an image
```

imagebuildah/stage_executor.go
```diff
@@ -1017,7 +1017,11 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			// creating a new working container with the
 			// just-committed or updated cached image as its new
 			// base image.
-			if _, err := s.prepare(ctx, imgID, false, true); err != nil {
+			//
```
nalind, line 1020:
```suggestion
```

nalind:
At lower levels, why do we ever try to pull an image when we know we were given an image ID?

vrothberg:
> At lower levels, why do we ever try to pull an image when we know we were given an image ID?

I found that curious as well but I am not very familiar with that code.

TomSweeneyRedHat:
LGTM

PR #3228, opened 2021-05-13
tests: new assert()

edsantiago:
This is much scarier than I had intended. The intention is to
start migrating from 'expect_output' and to use 'assert'
instead; the reason is that 'assert' allows much more
flexibility, particularly negative tests.

We've long wanted something like "assert that output does
not contain 'foo'". I've been too lazy to implement it,
but last week I noticed code in bud.bats that does:

    ! expect_output "sdfsdfsdf"

This is a really super bad idea: although it works fine when
things are good and tests pass, it's a disaster when tests
fail because the poor person debugging test output now sees
a test failure log, and starts tracking it down, and then
much later realizes that it was an intentional failure but
the '!' was negating it, and all that track-down work was
wasted.

To keep this PR reasonable, I'm keeping all positive uses
of expect_output untouched. We can migrate those (if desired)
over time. All negative uses are now assertions.


rhatdan:
@edsantiago still working on this?

edsantiago:
I still lean slightly toward thinking this is a good idea, but I don't want to pursue it unilaterally. Could I get a second opinion? Would this make the tests easier to write/maintain?

rhatdan:
Yes I think this is a good idea.  I think it reads easier and handling of string should not exists, is difficult.

edsantiago:
Thanks. I've put this back on my shortlist.

edsantiago:
@containers/buildah-maintainers opinions welcome

rhatdan:
LGTM
@nalind @TomSweeneyRedHat @vrothberg @umohnani8 @saschagrunert PTAL

vrothberg (approved):
/lgtm

PR #3250, opened 2021-05-21
containerImageRef.createConfigsAndManifests(): squashed images have no parent images

nalind:
#### What type of PR is this?

/kind bug

Remove the parent image's ID from the config blob when we're squashing the image, since after squashing, we share no layers or history with what was once our base image, and leaving it set triggers verification errors in registries that expect consistency between parent IDs and perhaps layers and history.

#### How to verify it

We've gotten at least one report that combining our `--squash` option with `--format docker` at commit time causes such a rejection when attempting to push the resulting image to quay 3.3.

```release-note
None
```


rhatdan:
LGTM
@vrothberg @giuseppe @TomSweeneyRedHat @ashley-cui @umohnani8 PTAL

giuseppe (approved):
/lgtm

PR #3255, opened 2021-05-25
containerImageRef/containerImageSource: don't buffer uncompressed layers

### LINKED ISSUE #1216: containerImageRef.NewImageSource potentially unnecessarily writes all layers to disk
If `containerImageSource.exporting`, `containerImageRef.NewImageSource` creates on-disk copies of all layers in `i.path`; `GetBlob` then only returns a handle to that on-disk copy.

This is potentially quite wasteful if the layers already exist on the destination, and this can be determined from their uncompressed digests already (and even more wasteful if `compress` is set, see #1215). Couldn’t the reading of layers be moved into `GetBlob` directly, apart from the new layer with unknown digest I guess?

The `.exporting` case happens only if `Builder.Commit` was called with a non-c/storage destination, which I guess is pretty rare (?), so this is fairly low-priority.

nalind:
#### What type of PR is this?

/kind bug

Instead of extracting layer content to a temporary file when we're committing to transports that aren't `containers-storage`, record the ID of a layer and the uncompressed size it has recorded for its contents.

When later asked for a blob, if we cached a layer ID and size, generate the layer diff on the fly, otherwise check for a file named after the digest of the requested blob in our cache directory location (usually used for new layers that we're adding) and the supplemental location (which can be supplied by a caller).

#### How to verify it

Fixes #1216

As noted in the original issue, this only crops up when we're committing directly to an external location, like a `dir:` location or a registry.  Committing to local storage and then pushing an image, in two steps, only created temporary files for new layers, and this PR continues to do so.

```
None
```

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -291,21 +291,8 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 	}
 	logrus.Debugf("committing image with reference %q is allowed by policy", transports.ImageName(dest))
 
-	// Check if the base image is already in the destination and it's some kind of local
-	// storage.  If so, we can skip recompressing any layers that come from the base image.
-	exportBaseLayers := true
-	if transport, destIsStorage := dest.Transport().(is.StoreTransport); destIsStorage && options.OciEncryptConfig != nil {
-		return imgID, nil, "", errors.New("unable to use local storage with image encryption")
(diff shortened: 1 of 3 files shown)


rhatdan:
Need to add [NO NEW TESTS NEEDED] if you don't need this tested.

saschagrunert (approved):
LGTM

Issue #3259, opened 2021-05-26
`buildah` fails when `REGISTRY_AUTH_FILE` doesn't exist


rcny:
Hello!

Setting `REGISTRY_AUTH_FILE` to a non-existent file will result in `buildah` failing to operate with public registries:

```
[root@dd836c415cc3 /]# env REGISTRY_AUTH_FILE=/tmp/non_existent_creds.json buildah pull docker.io/library/hello-world
error checking authfile path /tmp/non_existent_creds.json: stat /tmp/non_existent_creds.json: no such file or directory
```

Is this a feature or a bug?

Currently I resolve it with e.g. `env REGISTRY_AUTH_FILE= buildah pull docker.io/library/hello-world` for specific commands. Can't remove `REGISTRY_AUTH_FILE` at all because of #3216.

Running `1.20.1` from current `quay.io/buildah/stable` image.

rhatdan:
I am thinking this is a bug.

vrothberg:
I consider setting the env var as user-specified input (i.e., the counterpart of `--authfile`).

rhatdan:
The thing is we do not need to fail until the user actually attempts to use the file, in that case.

rhatdan:
For example the use case that this user was hitting was having it set within a container image, and only used when needed.  We would fail all the time.

TriplEight:
I'm having the same issue with `buildah logout`, even when it wasn't logged in anywhere:
```
buildah logout --all
error checking authfile path /tmp/registry_auth.json: stat /tmp/registry_auth.json: no such file or directory
```

TriplEight:
and with the workaround I'm getting the other misleading output:
```
env REGISTRY_AUTH_FILE= buildah logout docker.io
Not logged into docker.io
```

rhatdan:
Why is this misleading?

TomSweeneyRedHat:
I think the logout error is misleading as it doesn't indicate that the file doesn't exist.  From reading the message, I'd think the authfile existed, I just didn't have an entry in the authfile.

rhatdan:
We treat REGISTRY_AUTH_FILE="" the same as we do not being set at all.

PR #3260, opened 2021-05-26
Only check authfile if the user specified it at command line

rhatdan:
This allows us to have a REGISTRY_AUTH_FILE and only fail when
the authfile is specified to be used.  If the user specifies it
that means he intends to use it.

Fixes: https://github.com/containers/buildah/issues/3259

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -107,8 +107,10 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budOptions) error {
 			tags = tags[1:]
 		}
 	}
-	if err := auth.CheckAuthFile(iopts.BudResults.Authfile); err != nil {
-		return err
+	if c.Flag("authfile").Changed {
+		if err := auth.CheckAuthFile(iopts.BudResults.Authfile); err != nil {
+			return err
+		}
 	}
 
 	pullPolicy := define.PullIfMissing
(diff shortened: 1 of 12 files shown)


vrothberg:
I don't see a difference between doing a `pull --authfile foo` or `REGISTRY_AUTH_FILE=foo pull`.  While there's 1 user wanting to ignore the latter, we don't know if there are others who may rely on the presence checks (e.g., for typos).

TomSweeneyRedHat:
I'm with @vrothberg I think `pull --authfile foo` and `REGISTRY_AUTH_FILE=foo pull` should return the same result/error.

cmd/buildah/bud.go
```diff
@@ -107,8 +107,10 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budOptions) error {
 			tags = tags[1:]
 		}
 	}
-	if err := auth.CheckAuthFile(iopts.BudResults.Authfile); err != nil {
-		return err
+	if c.Flag("authfile").Changed {
```
TomSweeneyRedHat, line 110: could we add:
` || os.Getenv("REGISTRY_AUTH_FILE"); authfile == "" `

to these, and I think this addresses my concerns and @vrothberg's

PR #3267, opened 2021-05-28
Add rusage-logfile flag to optionally send rusage to a file

dcermak:
#### What type of PR is this?

/kind feature

Currently the rusage is reported to stdout and rather cumbersome to parse. The
new flag rusage-logfile can be used to specify a file to which the log will be
written instead.

#### How to verify it

Run `buildah bud --log-rusage --rusage-logfile "tests.log"`

None

I'd like to add a bats integration test as well, but am unfamiliar with bats. If anyone could help me out with that, that would be greatly appreciated.

```release-note
- add --rusage-logfile flag to log rusage into a file
```

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -334,6 +334,7 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budOptions) error {
 		Isolation:               isolation,
 		Labels:                  iopts.Label,
 		Layers:                  layers,
+		LogRusage:               iopts.LogRusage,
 		Manifest:                iopts.Manifest,
 		MaxPullPushRetries:      maxPullPushRetries,
 		NamespaceOptions:        namespaceOptions,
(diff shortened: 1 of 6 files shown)


rhatdan:
Simplest thing for testing would be to add the flag to one of the existing bud tests and make sure it records data to the specified file.  Bats is just a fancy wrapper on bash.

rhatdan:
/approve
I am not a big fan of options container "-" but since this is hidden, I am fine with it.

dcermak:
> I am not a big fan of options container "-" but since this is hidden, I am fine with it.

I have no strong opinion on this, so I can change it to something else. I just tried to adhere to the already existing options and didn't come up with a shorter name.

rhatdan:
@vrothberg @giuseppe @nalind @TomSweeneyRedHat PTAL

cmd/buildah/bud.go
```diff
@@ -358,6 +358,7 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budOptions) error {
 		OciDecryptConfig:        decConfig,
 		Jobs:                    &iopts.Jobs,
 		LogRusage:               iopts.LogRusage,
+		RusageLogFile:           iopts.RusageLogFile,
```
TomSweeneyRedHat, line 361: I know the three prior lines aren't in alpha order, but this one and the others ought to be

TomSweeneyRedHat:
Looks like you need to rebase @dcermak

cmd/buildah/bud.go
dcermak: I've put it in the right place (I hope).

dcermak:
@TomSweeneyRedHat done

PR #3292, opened 2021-06-09
Cirrus: Install docker from package cache

cevich:
#### What type of PR is this?

/kind flake

Installing packages at runtime (from an external source) is problematic
for many reasons.  Specifically in the case of buildah/docker
conformance testing, it means the current "latest" pacakges are
always installed.  This is a problem as new release branches are
created, because it presents an opportunity for test-environment changes
to happen after buildah/test code is stabilized.

Fix this by using new/special VM images which cache the required docker
packages.  At runtime then, the required packages may be installed from
this cache instead of reaching out to the repository.  Since images used
by tests on release branches never change, this will also serve to
stabilize the package versions for that specific environment.

#### How to verify it

The automated 'conformance' testing will pass in this PR

None

Ref: https://github.com/containers/automation_images/pull/75

None

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -30,7 +30,7 @@ env:
     UBUNTU_NAME: "ubuntu-2104"
     PRIOR_UBUNTU_NAME: "ubuntu-2010"
 
-    IMAGE_SUFFIX: "c6032583541653504"
+    IMAGE_SUFFIX: "c6534244118822912"
     FEDORA_CACHE_IMAGE_NAME: "fedora-${IMAGE_SUFFIX}"
     PRIOR_FEDORA_CACHE_IMAGE_NAME: "prior-fedora-${IMAGE_SUFFIX}"
     UBUNTU_CACHE_IMAGE_NAME: "ubuntu-${IMAGE_SUFFIX}"
@@ -173,7 +173,7 @@ conformance_task:
         - env:
             STORAGE_DRIVER: 'overlay'
 
-    setup_script: '${SCRIPT_BASE}/setup.sh |& ${_TIMESTAMP}'
+    setup_script: '${SCRIPT_BASE}/setup.sh conformance |& ${_TIMESTAMP}'
     conformance_test_script: '${SCRIPT_BASE}/test.sh conformance |& ${_TIMESTAMP}'
 
 
--- a/contrib/cirrus/lib.sh
+++ b/contrib/cirrus/lib.sh
@@ -87,6 +87,9 @@ ALPINE_FQIN=${ALPINE_FQIN:-docker.io/library/alpine}
 IN_PODMAN_NAME="in_podman_$CIRRUS_TASK_ID"
 IN_PODMAN="${IN_PODMAN:-false}"
 
+# Downloaded, but not installed packages.
+PACKAGE_DOWNLOAD_DIR=/var/cache/download
+
 lilto() { err_retry 8 1000 "" "$@"; }  # just over 4 minutes max
 bigto() { err_retry 7 5670 "" "$@"; }  # 12 minutes max
 
--- a/contrib/cirrus/setup.sh
+++ b/contrib/cirrus/setup.sh
@@ -33,6 +33,10 @@ EOF
         fi
         ;;
     ubuntu)
+        if [[ "$1" == "conformance" ]]; then
+            msg "Installing previously downloaded/cached packages"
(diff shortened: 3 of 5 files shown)


cevich:
Uh oh, I think I broke it.  @edsantiago every platform  (a guess - I only looked at two or three) is failing [this `bud with --runtime and --runtime-flag` test in the same way](https://cirrus-ci.com/task/6118511249457152?logs=integration_test#L1019).   I don't *think* (?) it has anything to do with my script changes (here), but may have something to do with brand-new/updated VM images.  Do you have/know/see any clues that may indicate what I broke?

edsantiago:
Sorry, I took a quick look at this last week but decided it was nothing I could have an opinion on.

Diving into it now, this is almost certainly a problem with the new VM update. I can reproduce this on my laptop, with buildah@master. My laptop updated runc  1.0.0-377.rc93 ->  -378.rc95 on 2021-05-29. 

My uneducated guess is that the test developer chose a string ("nsexec started") that looked reliable, but is actually not; and runc changed so that it no longer spits out that string; and someone will have to fix the buildah test.

cevich:
@edsantiago thanks for taking a look, that helps.  I haven't looked at the test recently...anything you can think of so the test doesn't need to rely on (perhaps frequently changing) debug-level messages?  If not, I'll try fixing the test with the message you suggest and see what happens.

edsantiago:
I would bring in a runc expert, because I have no idea what guarantees runc makes or does not make about its debug messages; or what problems we (buildah gating tests) will run into if the new test runs with an old runc (e.g. on RHEL). My naïve first reaction is to simply search for the presence of `level=debug`, which (to me) guarantees that the runtime-flag was passed along to *something* (whether that is runc or not, the test cannot determine, but I don't know how important it is to care about that. If it's important, the test could check for `"level=debug.*msg=.*/runc"`.)

cevich:
@giuseppe what's your opinion (see [test failure due to new runc](https://cirrus-ci.com/task/6118511249457152?logs=integration_test#L1019)).  Can I update the test to just check for `level=debug` or is something more extensive needed in the system-tests?

cevich:
Added proposed fix (in case it's acceptable to @giuseppe) and updated VM images yet again, to help confirm podman 3.2.1 addition to Ubuntu images is functional (cc @lsm5)

edsantiago:
LGTM

edsantiago:
Looks like seccomp again:
```
# time="2021-06-16T21:35:29Z" level=error msg="container_linux.go:380: starting container process caused: error adding seccomp filter rule for syscall bdflush: permission denied" func=main.fatal file="/usr/src/packages/BUILD/src/github.com/opencontainers/runc/utils.go:57"
# time="2021-06-16T21:35:29Z" level=debug msg="container_linux.go:380: starting container process caused: error adding seccomp filter rule for syscall bdflush: permission denied" func=main.fatal file="/usr/src/packages/BUILD/src/github.com/opencontainers/runc/utils.go:59"
```
[Log](https://api.cirrus-ci.com/v1/task/5929436924084224/logs/integration_test.log)

cevich:
*sigh*, yep and rebasing didn't help (was worth a try).

```
[+0731s] # time="2021-06-17T13:53:08Z" level=error msg="container_linux.go:380: starting container process caused: error adding seccomp filter rule for syscall bdflush: permission denied" func=main.fatal file="/usr/src/packages/BUILD/src/github.com/opencontainers/runc/utils.go:57"
[+0731s] # time="2021-06-17T13:53:08Z" level=debug msg="container_linux.go:380: starting container process caused: error adding seccomp filter rule for syscall bdflush: permission denied" func=main.fatal file="/usr/src/packages/BUILD/src/github.com/opencontainers/runc/utils.go:59"
[+0731s] # : exit status 1
```

cevich:
Opened test PR in podman to see if these several-days newer images reproduce similar failures https://github.com/containers/podman/pull/10709

cevich:
Yay, a result!  This problem _is_ reproducing in podman testing on Ubuntu 20.10 (CGv1 + runc): https://cirrus-ci.com/task/6152440987779072?logs=main#L272

cevich:
Okay assuming tests pass, this should be ready to go.  Though the real fruits will only appear over time in new release branches, it might only have a minor improvement to stability on 'main'.  Once this merges, I will investigate possibility of doing something similar for already released branches (possible in theory, practice is another matter).

PR #3295, opened 2021-06-14
Add kernel version requirement for fuse-overlayfs

stevegt:
/kind documentation

The fuse-overlayfs README says:

> Also, please note that, when using fuse-overlayfs from a user namespace (for example, when using rootless podman) a Linux Kernel > v4.18.0 is required.

#### How to verify it

Ran into this when trying to run with fuse-overlayfs on 4.8.0 -- doesn't work, throws an EINVAL -- see https://github.com/containers/podman/issues/10100.

None

No

```release-note
None
```

### DIFF
--- a/install.md
+++ b/install.md
@@ -154,7 +154,7 @@ sudo apt-get -qq -y install buildah
 ### Kernel Version Requirements
 To run Buildah on Red Hat Enterprise Linux or CentOS, version 7.4 or higher is required.
 On other Linux distributions Buildah requires a kernel version of 4.0 or
-higher in order to support the OverlayFS filesystem.  The kernel version can be checked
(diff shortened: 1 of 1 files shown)


rhatdan:
I don't like specific version checks, because they do not take into account back porting of fixes to kernel is distributions like RHEL and Centos.

stevegt:
@rhatdan There's already a "specific version check" in install.md -- this PR was a correction and clarification of existing documentation.  Our local team lost days ferreting this out, and I'm hoping to help other buildah adopters avoid the same pitfall. 

@vrothberg Why the close?  What needs to be improved in order to get this accepted?

stevegt:
I've submitted PR #3361 to remove the existing specific version number from the "Kernel Version Requirements" section -- this PR should remain rejected if #3361 is accepted and vice versa.

vrothberg:
My apologies, @stevegt!  Thanks opening the other PR.

PR #3296, opened 2021-06-14
Fix copy into workdir for a single file

Luap99:
#### What type of PR is this?

/kind bug

When copying a single file to the workdir make sure to create the workdir
and not to overwrite it as file.

#### How to verify it
see added test

Fixes containers/podman#10671

```release-note

```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -1830,6 +1830,20 @@ _EOF
   test -s "${root}"/subdir/file2.txt
 }
 
+# regression test for https://github.com/containers/podman/issues/10671
+@test "bud-copy-workdir --layers" {
+  _prefetch alpine
+
+  target=testimage
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json --layers -t ${target} -f Dockerfile.2 ${TESTSDIR}/bud/copy-workdir
+  run_buildah from ${target}
+  cid="$output"
+  run_buildah mount "${cid}"
+  root="$output"
+  test -d "${root}"/subdir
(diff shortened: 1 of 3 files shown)


tests/bud.bats
```diff
@@ -1830,6 +1830,20 @@ _EOF
   test -s "${root}"/subdir/file2.txt
 }
 
+# regression test for https://github.com/containers/podman/issues/10671
+@test "bud-copy-workdir --layers" {
+  _prefetch alpine
+
```
TomSweeneyRedHat, line 1836: @edsantiago has convinced us to create the Containerfiles like we do at line 1810 in this file.  Could you do that rather than creating a separate file?
```
mytmpdir=${TESTDIR}/my-dir
cat > $mytmpdir/Containerfile.2 << _EOF
FROM alpine
WORKDIR /subdir
COPY file1.txt /subdir
RUN ls
_EOF
```

and then adjusting your bud command below

edsantiago: @TomSweeneyRedHat this is embarrassing, when did I make that recommendation? The case in 1810 is, I think, due to the need for an ignore file in the same directory. The overwhelming majority of buildah tests use a separate Docker/Containerfile, and I think I'm fine with that?

TomSweeneyRedHat:
@edsantiago maybe I misremembered and should be embarrassed?  I slightly prefer having it in the test itself, especially for small Containerfiles, I find it easier to review/read the test.  

But if you're hip, then I won't put @Luap99 through the wringer on this one.

LGTM

Issue #3297, opened 2021-06-15
New "copy" helper is going to wreak havoc


edsantiago:
The new `copy` helper added in #3220 is going to cause a huge number of problems:
* buildah-bud tests under podman will no longer work, because they don't build or know about this tool
* Fedora and RHEL gating tests are all going to fail, because the tool is not packaged in RPMs

If the benefits of the new tool outweigh these costs, someone will need to fix all of the above:
* podman: I have no idea how to fix that. And I'm going to be on PTO, so I leave this as SEP (Someone Else's Problem)
* gating tests:
  * need to eliminate hardcoded references to `copy` in `helpers.bash`, in favor of redirection via envariable; see sample diff below
  * need to fix specfiles (Fedora, RHEL, Centos, etc) to include `buildah-copy-helper` (or some other nonconflicting name) in the `buildah-tests` subpackage
  * need to fix `tests/test-buildah.yml` (again in Fedora, RHEL, Centos) to define the magic new envariable

```diff
diff --git a/tests/helpers.bash b/tests/helpers.bash
index 2179926a..24b52a38 100644
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -2,6 +2,7 @@

 BUILDAH_BINARY=${BUILDAH_BINARY:-$(dirname ${BASH_SOURCE})/../bin/buildah}
 IMGTYPE_BINARY=${IMGTYPE_BINARY:-$(dirname ${BASH_SOURCE})/../bin/imgtype}
+BUILDAH_COPY_HELPER=${BUILDAH_COPY_HELPER:-$(dirname ${BASH_SOURCE})/../bin/copy}
 TESTSDIR=${TESTSDIR:-$(dirname ${BASH_SOURCE})}
 STORAGE_DRIVER=${STORAGE_DRIVER:-vfs}
 PATH=$(dirname ${BASH_SOURCE})/../bin:${PATH}
@@ -96,18 +97,18 @@ function _prefetch() {
         fname=$(tr -c a-zA-Z0-9.- - <<< "$img")
         if [ -d $_BUILDAH_IMAGE_CACHEDIR/$fname ]; then
             echo "# [restoring from cache: $_BUILDAH_IMAGE_CACHEDIR / $img]" >&2
-            copy $COPY_OPTS dir:$_BUILDAH_IMAGE_CACHEDIR/$fname containers-storage:"$img"
[...]

PR #3298, opened 2021-06-15
tests: make it easer to override the location of the copy helper

nalind:
#### What type of PR is this?

/kind other

Make it possible to override the location of the `copy` test helper by setting its location in an environment variable, as we do for `imgtype`.

#### How to verify it

This is a tweak to how we cache images for integration tests.  They should continue to pass.

Related to #3297.

```
None
```

### DIFF
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -2,6 +2,7 @@
 
 BUILDAH_BINARY=${BUILDAH_BINARY:-$(dirname ${BASH_SOURCE})/../bin/buildah}
 IMGTYPE_BINARY=${IMGTYPE_BINARY:-$(dirname ${BASH_SOURCE})/../bin/imgtype}
+COPY_BINARY=${COPY_BINARY:-$(dirname ${BASH_SOURCE})/../bin/copy}
 TESTSDIR=${TESTSDIR:-$(dirname ${BASH_SOURCE})}
 STORAGE_DRIVER=${STORAGE_DRIVER:-vfs}
 PATH=$(dirname ${BASH_SOURCE})/../bin:${PATH}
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM
@edsantiago PTAL if you get a chance before your PTO

tests/helpers.bash
```diff
@@ -2,6 +2,7 @@
 
 BUILDAH_BINARY=${BUILDAH_BINARY:-$(dirname ${BASH_SOURCE})/../bin/buildah}
 IMGTYPE_BINARY=${IMGTYPE_BINARY:-$(dirname ${BASH_SOURCE})/../bin/imgtype}
+COPY_BINARY=${COPY_BINARY:-$(dirname ${BASH_SOURCE})/../bin/copy}
```
edsantiago, line 5: I know this follows the `XXX_BINARY` convention, but at this point I'd like to suggest breaking away from that and using `BUILDAH_COPY_HELPER` instead. Something that makes it obvious and explicit that this is a custom buildah thing.

edsantiago (commented):
One suggestion, and one request that I can't do inline: please remove the `COPY_OPTS=` definition in line 37.

tests/helpers.bash
rhatdan: Could you open a separate PR to change IMGTYPE and COPY to the name you want.

edsantiago: >Could you open a separate PR to change IMGTYPE and COPY to the name you want.

It's too late for IMGTYPE: that change would need to be coordinated across gating tests on RHEL and Fedora, on all possible branches. That's just not sane.

I was hoping we could get this one done properly right now, so we could get things clean in gating tests.

PR #3323, opened 2021-06-20
Add --json flags to `mount` and `version` commands

jacobwgillespie:
#### What type of PR is this?

/kind feature

This PR adds a `--json` flag to the `buildah mount` and `buildah version` commands. When this flag is provided, the output from each command is printed as JSON.

In the case of `buildah mount --json`, output is printed like:

```json
[
    {
        "container": "xxxxxxxxxxxxxxxx",
        "mountPoint": "/path/to/mount/point"
    }
]
```

And in the case of `buildah version --json`, output is printed like:

```json
{
    "version": "1.22.0-dev",
    "goVersion": "go1.16.5",
    "imageSpec": "1.0.1-dev",
    "runtimeSpec": "1.0.2-dev",
    "cniSpec": "0.4.0",
    "libcniVersion": "v0.8.1",
    "imageVersion": "5.13.2",
    "gitCommit": "5181b9ce",
    "built": "Sun Jun 20 20:49:49 2021",
    "osArch": "linux/amd64"
}
```

I am currently working on a TypeScript wrapper for `buildah` that exposes all of the Buildah API, but via TypeScript executing the binary. When integrating with another tool/language, it is more convenient for Buildah's output to be structured, rather than needing to re-parse the output with regex. There is already a `--json` flag for commands like `image` and `container`, so this PR extends that flag to `mount` and `version`.

#### How to verify it

Run `buildah mount --json` and `buildah version --json`.

None

Let me know if you would like anything adjusted. I have tried to follow existing patterns from other commands, but I'm flexible.

```release-note
Added --json flag to mount and version commands to print output as JSON
```

### DIFF
--- a/cmd/buildah/mount.go
+++ b/cmd/buildah/mount.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 
@@ -9,6 +10,11 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
+
+type jsonMount struct {
+	Container  string `json:"container,omitempty"`
+	MountPoint string `json:"mountPoint"`
+}
 
 func init() {
 	var (
@@ -20,13 +26,14 @@ func init() {
   view/modify the content in the containers root file system.
 `
 		noTruncate bool
+		outputJSON bool
 	)
 	mountCommand := &cobra.Command{
 		Use:   "mount",
 		Short: "Mount a working container's root filesystem",
 		Long:  mountDescription,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return mountCmd(cmd, args)
+			return mountCmd(cmd, args, outputJSON)
 		},
 		Example: `buildah mount
   buildah mount containerID
@@ -41,6 +48,7 @@ func init() {
 
 	flags := mountCommand.Flags()
 	flags.SetInterspersed(false)
+	flags.BoolVar(&outputJSON, "json", false, "output in JSON format")
 	flags.BoolVar(&noTruncate, "notruncate", false, "do not truncate output")
 	rootCmd.AddCommand(mountCommand)
 	if err := flags.MarkHidden("notruncate"); err != nil {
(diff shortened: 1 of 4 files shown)


cmd/buildah/mount.go
```diff
@@ -41,14 +48,15 @@ func init() {
 
 	flags := mountCommand.Flags()
 	flags.SetInterspersed(false)
+	flags.BoolVar(&outputJSON, "json", false, "output in JSON format")
```
rhatdan, line 51: Make a global mountOptions with outputJson as a param.

type mountOptions struct {
	json        bool
}

Then no need to pass json flag as an option.

rhatdan:
Thanks @jacobwgillespie 
A couple of nits.

cmd/buildah/version.go
```diff
@@ -21,45 +22,86 @@ var (
...
 func init() {
+	var opts versionOptions
+
+	//cli command to print out the version info of buildah
+	versionCommand := &cobra.Command{
+		Use:   "version",
+		Short: "Display the Buildah version information",
+		Long:  "Displays Buildah version information.",
+		RunE: func(c *cobra.Command, args []string) error {
+			var err error
```
rhatdan, line 51: Could you put this code in a separate function, So the init  is not as cluttered.

jacobwgillespie: Done!

cmd/buildah/mount.go
jacobwgillespie: Done! I've omitted `noTruncate` from the struct as the value of that flag is not actually used in the command implementation, but let me know if you'd prefer that it be included in the `mountOptions` struct as well.

TomSweeneyRedHat:
@jacobwgillespie thanks for the PR, however, our CI requires PRs to be signed.  Can you please `git commit --amend -s` and then fetch, rebase and push again please?

TomSweeneyRedHat (commented):
Changes LGTM, but it would be good to have some tests for both new options too.

@jwhonce if you've time for a quick peek, I'd appreciate it.

jacobwgillespie:
> our CI requires PRs to be signed

My bad, missed this in the followup commits. 🤦 

I've rebased and added signoff.

jwhonce (approved):
LGTM

PR #3329, opened 2021-06-22
Add support for default network creation

mheon:
As part of the podman-machine work, we need to stop distributing the default CNI configuration that has traditionally been
supplied by Podman, 87-podman.conflist. Buildah (as root) also uses this file, and will be unable to set up container networking without it.

The replacement we've introduced is the ability to automatically create said configuration file using a package in containers/common. This has already been added to Podman, and this PR adds it to Buildah.

#### What type of PR is this?

/kind other

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -30,6 +30,7 @@ import (
 	"github.com/containers/common/pkg/capabilities"
 	"github.com/containers/common/pkg/chown"
 	"github.com/containers/common/pkg/config"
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
@mheon, needs a test or at least the notest flag

TomSweeneyRedHat:
LGTM
once tests are happy

rhatdan:
LGTM

mheon:
I think this should be good now.

Issue #3152, opened 2021-04-16
buildah run breaks stdin by setting `O_NONBLOCK`


Vogtinator:
Sometimes after running `buildah run` in a terminal, commands executed afterwards which try to read from `stdin` fail with `EAGAIN`. This is because it sets `O_NONBLOCK` on the FD:

```
f241:~ # grep flags /proc/self/fdinfo/0; while ! grep 'flags.*4002$' /proc/self/fdinfo/0; do buildah run sle15-working-container /bin/true; echo -n .; done; cat -
flags:  02
....................flags:      04002
cat: -: Resource temporarily unavailable
```

`strace -k` shows that this is caused by the `runCopyStdio` function: https://github.com/containers/buildah/blob/c891ebe41e8e33732c4bb8ca23004e7b4f857cc8/run_linux.go#L1236

For some reason this doesn't happen all the time and the `O_NONBLOCK` flag also disappears after some time, even after `buildah` exited. I suspect that the interactive shell is fixing that up at some point.

rhatdan:
Is this something buildah could easily change back on exit?

Vogtinator:
It could just restore the previous descriptor flags, I suppose.

rhatdan:
Interested in opening a PR?

Vogtinator:
Interested yes, but I unfortunately don't have the time to get familiar with the code and Go, at least not in the near future...

rhatdan:
We have talked about this, and we are not sure there is an easy way to fix.

PR #3338, opened 2021-06-24
Turn stdio back to blocking when command finishes

rhatdan:
Fixes: https://github.com/containers/buildah/issues/3152

Sometimes after running buildah run in a terminal, commands executed afterwards which try to read from stdin fail with EAGAIN. This is because it sets O_NONBLOCK on the FD:

[NO TESTS NEEDED] Since I don't know how to test this.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -1182,16 +1182,21 @@ func runConfigureNetwork(isolation define.Isolation, options RunOptions, configu
 	return teardown, nil
 }
 
-func setNonblock(logger *logrus.Logger, fd int, description string, nonblocking bool) error { //nolint:interfacer
-	err := unix.SetNonblock(fd, nonblocking)
+func setNonblock(logger *logrus.Logger, fd int, description string, nonblocking bool) (bool, error) { //nolint:interfacer
+	mask, err := unix.FcntlInt(uintptr(fd), unix.F_GETFL, 0)
 	if err != nil {
+		return false, err
+	}
+	blocked := mask&unix.O_NONBLOCK == 0
+
+	if err := unix.SetNonblock(fd, nonblocking); err != nil {
 		if nonblocking {
 			logger.Errorf("error setting %s to nonblocking: %v", description, err)
 		} else {
 			logger.Errorf("error setting descriptor %s blocking: %v", description, err)
 		}
 	}
-	return err
+	return blocked, err
 }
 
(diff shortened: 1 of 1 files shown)


rhatdan:
@nalind I am not sure if this is correct, PTAL?

nalind:
This assumes the descriptor was originally blocking, and I'm not sure that's a valid assumption.  Also, it looks like we don't set a default value for `options.Logger` in `Run()` in cases where the caller didn't specify one, and the tests are now hitting code paths that attempt to use the value, which they weren't before.

rhatdan:

rhatdan:
@Vogtinator I can not get the original bug to happen, could you check to see if this fixes your problem?

Vogtinator:
> @Vogtinator I can not get the original bug to happen, could you check to see if this fixes your problem?

Even by running the loop in #3152? Strange. I'll give it a try, but might take until next week.

Vogtinator:
> > @Vogtinator I can not get the original bug to happen, could you check to see if this fixes your problem?
> 
> Even by running the loop in #3152? Strange. I'll give it a try, but might take until next week.

Went quicker than expected. I first reproduced the original bug a few times:

```
localhost:~ # grep flags /proc/self/fdinfo/0; while ! grep 'flags.*4002$' /proc/self/fdinfo/0; do buildah run tumbleweed-working-container /bin/true; echo -n .; done; cat -
flags:  02
.......flags:   04002
cat: -: Resource temporarily unavailable
localhost:~ # grep flags /proc/self/fdinfo/0; while ! grep 'flags.*4002$' /proc/self/fdinfo/0; do buildah run tumbleweed-working-container /bin/true; echo -n .; done; cat -
flags:  02
...flags:       04002
cat: -: Resource temporarily unavailable
localhost:~ # grep flags /proc/self/fdinfo/0; while ! grep 'flags.*4002$' /proc/self/fdinfo/0; do buildah run tumbleweed-working-container /bin/true; echo -n .; done; cat -
flags:  02
.....flags:     04002
cat: -: Resource temporarily unavailable
```

and then installed `buildah` from this PR:

```
localhost:~ # grep flags /proc/self/fdinfo/0; while ! grep 'flags.*4002$' /proc/self/fdinfo/0; do buildah run tumbleweed-working-container /bin/true; echo -n .; done; cat -
flags:  02
..............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................^C
```

No failure!

For completeness, I also wrote a quick test program which opens `stdin` with `O_NONBLOCK` set and can confirm that it stays set even after multiple invocations of `buildah run`.

From my perspective, LGTM. Thanks!

PR #3351, opened 2021-06-30
Ci release 1.18

cevich:
There's no sense in wasting resources on extensive testing for branches
which see very little if any changes.  Trim down automation to the
minimum viable jobs.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -104,32 +104,6 @@ gce_instance:
         path: ./bin/*
 
 
-'cirrus-ci/only_prs/unit_task':
-
-    timeout_in: 45m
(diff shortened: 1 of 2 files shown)


cevich:
Okay I think this is ready. The diff is a bit hard to read (I re-arranged the tasks), but [effectively these are the only tasks that will run](https://github.com/containers/buildah/pull/3351/files#diff-62587956f943bb2503db7bc6dd27d0d888074a1c0ecaab3f570ad611aff0f7bbR142-R145). So unit and integration are gone, the only thing we check is "Does it compile" and "Does it pass docker conformance".

PR #3358, opened 2021-07-02
vendor containers/common@main

vrothberg:
The `IgnorePlatform` options has been removed from the
`LookupImageOptions` in libimage to properly support multi-arch images.
Buildah always set it to true.

Also temporarily remove /usr/share/containers/containers.conf.  c/common
is now throwing warnings when facing unknown keys which unfortunately is
the case with the upstream containers.conf and ultimately breaks CI
since some output checks are failing.

Once an updated containers.conf has been shipped, we can revert the
change.

[NO TESTS NEEDED]


vrothberg:
@ashley-cui @rhatdan @containers/podman-maintainers PTAL

ashley-cui (commented):
LGTM

cmd/buildah/main.go
```diff
@@ -78,6 +78,8 @@ func init() {
 	}
 
 	cobra.OnInitialize(initConfig)
+	// Disable the implicit `completion` command in cobra.
+	rootCmd.CompletionOptions.DisableDefaultCmd = true
```
nalind, line 82: What was this for?  It wasn't mentioned in the commit log or this PR description.

vrothberg: I have no idea.  It's too long ago.

I assume something failed or a behavior changed.  Cobra was bumped in this PR as well.

rhatdan: Best to remove it and see what breaks.

Issue #3366, opened 2021-07-09
Rename bud to build


Fodoj:
This is more of a suggestion that the issue, of course. 

`bud` is `build-using-dockerfile`. This command will also understand 'Containerfile', which is not Docker-oriented, at least from the name. Renaming `bud` to just `build` will remove this kind of discrepancy. And once Containerfile is a real thing, and not just a convention, then `build` will make ever more sense, than `bud`.

rhatdan:
Interested in opening a PR to make bud == build, so either would work.  I don't see us removing bud, since it is well known and would break backwards compatibility.

Fodoj:
Yes, I agree, for backwards compatibility bud needs to stay for a while.

PR #3368, opened 2021-07-09
Rename bud to build, while keeping an alias for to bud.

Fodoj:
#### What type of PR is this?

> /kind other

bud is build-using-dockerfile. This command will also understand 'Containerfile', which is not Docker-oriented, at least from the name. Renaming bud to just build will remove this kind of discrepancy. And once Containerfile is a real thing, and not just a convention, then build will make ever more sense, than bud. Original issue: https://github.com/containers/buildah/issues/3366.

Fixes https://github.com/containers/buildah/issues/3366

```release-note
For users of bud and build-with-dockerfile commands everything should remain the same, as they are aliases for the build command.
```


docs/buildah.md
```diff
@@ -139,7 +139,7 @@ Buildah can set up environment variables from the env entry in the [engine] tabl
 | Command               | Description                                                                                          |
 | --------------------- | ---------------------------------------------------                                                  |
 | buildah-add(1)        | Add the contents of a file, URL, or a directory to the container.                                    |
-| buildah-bud(1)        | Build an image using instructions from Dockerfiles.                                                  |
+| buildah-build(1)        | Build an image using instructions from Dockerfiles.                                                |
```
rhatdan, line 142: s/Dockerfiles/Containerfiles or Dockerfiles/g

pkg/cli/common.go
```diff
@@ -46,8 +46,8 @@ type NameSpaceResults struct {
 	UTS           string
 }
 
-// BudResults represents the results for Bud flags
-type BudResults struct {
+// BuildResults represents the results for Build flags
+type BuildResults struct {
```
rhatdan, line 50: Leave the structs the same or you will cause other packages to rebuild and should bump buildah main version.

pkg/cli/common.go
```diff
@@ -88,8 +88,8 @@ type BudResults struct {
 }
 
 // FromAndBugResults represents the results for common flags
-// in bud and from
-type FromAndBudResults struct {
+// in build and from
+type FromAndBuildResults struct {
```
rhatdan, line 92: Leave alone.

pkg/cli/common.go
```diff
@@ -171,8 +171,8 @@ func GetLayerFlags(flags *LayerResults) pflag.FlagSet {
 
 // Note: GetLayerFlagsCompletion is not needed since GetLayerFlags only contains bool flags
 
-// GetBudFlags returns common bud flags
-func GetBudFlags(flags *BudResults) pflag.FlagSet {
+// GetBuildFlags returns common build flags
+func GetBuildFlags(flags *BuildResults) pflag.FlagSet {
```
rhatdan, line 175: Don't change.

pkg/cli/common.go
```diff
@@ -231,8 +231,8 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
 	return fs
 }
 
-// GetBudFlagsCompletions returns the FlagCompletions for the common bud flags
-func GetBudFlagsCompletions() commonComp.FlagCompletions {
+// GetBuildFlagsCompletions returns the FlagCompletions for the common build flags
+func GetBuildFlagsCompletions() commonComp.FlagCompletions {
```
rhatdan, line 235: Don't change.

pkg/cli/common.go
```diff
@@ -263,8 +263,8 @@ func GetBudFlagsCompletions() commonComp.FlagCompletions {
 	return flagCompletion
 }
 
-// GetFromAndBudFlags returns from and bud flags
-func GetFromAndBudFlags(flags *FromAndBudResults, usernsResults *UserNSResults, namespaceResults *NameSpaceResults) (pflag.FlagSet, error) {
+// GetFromAndBuildFlags returns from and build flags
+func GetFromAndBuildFlags(flags *FromAndBuildResults, usernsResults *UserNSResults, namespaceResults *NameSpaceResults) (pflag.FlagSet, error) {
```
rhatdan, line 267: Don't change

pkg/cli/common.go
```diff
@@ -310,8 +310,8 @@ func GetFromAndBudFlags(flags *FromAndBudResults, usernsResults *UserNSResults,
 	return fs, nil
 }
 
-// GetFromAndBudFlagsCompletions returns the FlagCompletions for the from and bud flags
-func GetFromAndBudFlagsCompletions() commonComp.FlagCompletions {
+// GetFromAndBuildFlagsCompletions returns the FlagCompletions for the from and build flags
+func GetFromAndBuildFlagsCompletions() commonComp.FlagCompletions {
```
rhatdan, line 314: Don't change

rhatdan (approved):

docs/buildah-build.md
```diff
@@ -1,15 +1,17 @@
-# buildah-bud "1" "April 2017" "buildah"
+# buildah-build "1" "April 2017" "buildah"
```
rhatdan, line 1: You need to add a link for this so if a user
does man buildah bud, this man page comes up

Fodoj: How would the link look like? Should I keep ` # buildah-bud "1" "April 2017" "buildah"` in addition to ` # buildah-build "1" "April 2017" "buildah"`

Fodoj:
Hard to say why integration tests failed in CI - https://github.com/containers/buildah/runs/3030825907 - every test returns "ok", but result is somehow failed

docs/buildah-build.md
rhatdan: No I am talking about something like:

https://github.com/containers/podman/tree/main/docs/source/markdown/links

rhatdan:
I think this PR is a little too ambitious.  I think you should first leave all of the tests alone.  And see if you can get the code to work with just changing the command name.

Remember `buildah bud` should continue to work, and these tests should pass after your change.

Fodoj:
As bud is an alias for build, and build is just a s/bud/build/, then I don’t see why tests would fail - both commands invoke the same code, and it is in fact just the renaming, I don’t do more than that

Fodoj:
Are there any tests that are failing? In CI, seems like all of them pass, but then there is an error code 128, from somewhere - but it’s not shown in the logs what ithe problem is

rhatdan:
You PR Is deleting all of the Dockerfiles?  Add these back and the tests should work better.  Example
tests/build/add-file/Dockerfile is deleted.

Fodoj:
Oh, no idea how this happened :| As suggested, I rolled back my changes in tests/ directory.

docs/buildah-build.md
Fodoj: I've added extra dir for links, and modified Makefile to copy those during installation.

rhatdan:
Can you convert a goupe of the bud.bats tests to actually use buildah build instead of buildah bud.

LGTM
@nalind PTAL

Fodoj:
@rhatdan just s/bud/build/ in bud.bats file? Just to make just I understood correctly.

rhatdan:
Yes, that would be fine.  Basically we want to test mainly buildah build, but we need to have a few tests that do `buildah bud` to make sure we are backwards compatible.

Fodoj:
Done - there are still `run_buildah 125 bud` calls in bud.tests, while the majority of them are `run_buildah build`. There are also plenty of `bud` invocations in other test files, so both commands are being invoked. Should I refactor tests/ to only use `build` in a subsequent PR or leave it like this for some time?

rhatdan:
I am fine with leavin bud tests as is.  Will see how this goes, but we need @nalind approval.

TomSweeneyRedHat:
The changes look fine, and the tests pass, but I'm partial to the quirky "bud" at this point.   Personally, I'd rather just add build as an alias documenting that as being available too, and leave bud there.  

But take that with a grain of salt, I'm generally very resistant to change and will go with the flow on this one.

edsantiago:
Just as an FYI to the team, this is going to be chaotic the next time podman vendors buildah: the buildah-bud tests are going to conflict pretty badly, and need serious human attention. If I'm around, that'll be me, but we probably need to get more expertise developed on those tests. (I was never expecting buildah-bud tests to be so volatile).

Fodoj:
@nalind ping (I would like to avoid rebasing it too often :-) )

Fodoj:
Should I re-base or wait till final review is there?  @rhatdan

rhatdan:
Please rebase.

Fodoj:
@rhatdan I've rebased two more times - I've noticed that you pushed some work in similar direction into main branch already. Should I close this PR?

rhatdan:
No, I think we should move forward with this.  I am on PTO the next couple of days,  Will take a look at this on Monday, unless @nalind and @TomSweeneyRedHat get to it first.

edsantiago:
When I check this out via `git pr`, the git history is weird, and I can't figure out why. I don't think this is rebased in the way I'm used to rebasing: I expected to see your PR based on 197b9a1f, but instead I see a bunch of different 'Merge' commits. I would suggest holding off on reviewing or merging this until that is resolved. @Fodoj, how are you performing your rebases?

docs/tutorials/01-intro.md
```diff
@@ -234,9 +234,9 @@ OCI container images built with `buildah` are completely standard as expected. S
 
     # dnf -y remove docker
 
-## Using Dockerfiles with Buildah
+## Using Dockerfiles/Containerfiles with Buildah
```
rhatdan, line 237: Nit: Put Containerfiles first.  When referring to both we should always put Containerfiles first.

README.md
```diff
@@ -104,7 +104,7 @@ $ sudo ./lighttpd.sh
 | Command                                              | Description                                                                                          |
 | ---------------------------------------------------- | ---------------------------------------------------------------------------------------------------- |
 | [buildah-add(1)](/docs/buildah-add.md)               | Add the contents of a file, URL, or a directory to the container.                                    |
-| [buildah-bud(1)](/docs/buildah-bud.md)               | Build an image using instructions from Dockerfiles.                                                  |
+| [buildah-build(1)](/docs/buildah-build.md)           | Build an image using instructions from Dockerfiles or Containerfiles.                                |
```
rhatdan, line 107: Containerfiles or Dockerfiles.

contrib/completions/bash/buildah
```diff
@@ -1184,6 +1184,7 @@ esac
    local commands=(
        add
        bud
+       build
        build-using-dockerfile
```
rhatdan, line 1188: Remove this command completion.

docs/buildah-build.md
```diff
@@ -1,15 +1,17 @@
...
 
-**buildah build-using-dockerfile** [*options*] [*context*]
+**buildah build** [*options*] [*context*]
 
 **buildah bud** [*options*] [*context*]
 
-**bud** is an alias for **build-using-dockerfile**.
+**buildah build-using-dockerfile** [*options*] [*context*]
+
+**build** has aliases for **bud** and **build-using-dockerfile**.
```
rhatdan, line 14: s/for//

rhatdan (approved):

rhatdan:
You need to rebase your PR and then force a --force push

git rebase -i origin
git push --force.

Only ship 1 PR not 10.

Fodoj:
Back to single commit

rhatdan:
LGTM
@TomSweeneyRedHat @edsantiago @nalind @vrothberg @flouthoc @ashley-cui PTAL

ashley-cui:
LGTM

TomSweeneyRedHat:
I still don't like this change.  I'd be fine with 'build' as an alias to 'bud', but not the other way around.  We've messaged `bud` for a number of years now, and I think we should keep it at the forefront.  My $.02.

rhatdan:
@TomSweeneyRedHat Do you see people noticing it?

rhatdan:
buildah bud will continue to work
man buildah bud will continue to work.

TomSweeneyRedHat:
I've not heard people complaining about bud over build.  I just think it's neat thing that sets us apart and I'm concerned will be lost over time as it gets replaced by `build` on the buildah man page and in the error messages.  I do know that the man page and the command will still work.

flouthoc:
LGTM !
Just one concern @Fodoj @TomSweeneyRedHat @rhatdan Should we just add `build` as an alias to `bud` instead of other-way around that should involve almost no code change but that is not really important.This also LGTM

Fodoj:
The reasons I went for the re-writing `bud` to be `build` instead of the alias is the following: 

1. There is a big push (primarily from RedHat) to use Containerfiles instead of Dockerfiles. This is reflected in documentation of Podman, for example. It's also the new terminology used in various blog posts.

I agree with the idea of not using brand-specific filename and adopting something more generic - in the end, lots of tools work with Dockerfiles, not just Docker. Of course, right now Dockerfile==Containerfile, as there is no independent specification of what Containerfile is - hopefully, it will appear, in similar fashion as image-spec and runtime-spec of OCI appeared out of original Docker's formats.

2. Buildah already supports automatic lookup of both Containerfile and Dockerfile. 

The command named "build-using-dockerfile" is a bit misleading at this point - it could as well be "buc" - "build-using-containerfile", with the same meaning. And if in the future new convention will apear, like "Imagefile", and Buildah also supports this one, then it will need "bui".

On the other hand, "build" is way more generic - it's the command that accepts some manifest with image build process definition, and builds an image. It does not matter what's the filename, as long as it follows some specification. In this case, automatic lookup of Dockerfile/Containerfile is just a convenience layer for end user.

3. `build` command is in general more known, even Podman has `build` command - which essentially invokes `buildah bud` behind the curtains. `bud` is rather an unexpected term for something that builds a container image.

Based on these reasons, I don't think that _the_ command should be "bud". "bud" feels like a "we've added this command to ease transition from Docker", which was probably, valid at the time of first Buildah releases. 

Right now it looks more like a relict, that conflicts with general of idea of going Dockerless way. Long term, I would suggest to drop `bud` completely in one of the major releases, and keep Docker-agnostic `build`, but of course it's just my opinion.

flouthoc:
@Fodoj very fair points I agree. thanks for explaining your thought process.

PR #3381, opened 2021-07-15
[CI:DOCS] Multi-arch: Always push updated version-tagged image

### LINKED ISSUE #3365: 1.21.1 and 1.21.2 missing in quay.io repository
**Description**
Latest versions of buildah: `1.21.1` and `1.21.2` are not available on quay.io. We would like to check if we can use those because `1.21.0` doesn't pass vulnerabilities check in our company, while `latest` is ok. Unfortunately we can't import `latest` images into our internal registry, so we'd like to check fixed versions.

1. Open https://quay.io/repository/buildah/stable?tab=tags
2. Search for tag `1.21`

Only `1.21.0` is present

cevich:
#### What type of PR is this?

/kind other

Given a near daily release of Fedora updates, some addressing security
issues, it's safest for users to always have an updated version-tagged
image.  The cost of this is more image churn, but for a relatively small
image this should hopefully be a minor burden vs improved security.

Note: This is not a perfect solution, as there may be no
hard-dependencies between buildah and other packages with security
updates.  However it is an improvement over the previous regime where we
simply pushed the latest version-tagged image once, then never touched
it again.

Minor: Rename `gen_fqin` identifier to `reponame_reg` to be more
consistent with it's sister block `containers_reg`.

#### How to verify it

Merge the PR and if it fails, fix it in a new PR :sob: 
(this is github's fault)

Fixes #3365

* Github-actions workflow is very head-hurtie to review / understand.  I apologize in advance :disappointed: 
* Depends on #3385

None

/cc @rostrovsky

### DIFF
--- a/.github/workflows/multi-arch-build.yaml
+++ b/.github/workflows/multi-arch-build.yaml
@@ -87,25 +87,14 @@ jobs:
           echo "::set-output name=version::$VERSION"
 
       - name: Generate image FQIN(s) to push
-        id: gen_fqin
+        id: reponame_reg
         run: |
           if [[ "${{ matrix.source }}" == 'stable' ]]; then
             # The command version in image just built
             VERSION='v${{ steps.sniff_test.outputs.version }}'
             # workaround vim syntax-highlight bug: '
-            # Image tags previously pushed to quay
-            ALLTAGS=$(skopeo list-tags \
-                      docker://$REPONAME_QUAY_REGISTRY/stable | \
-                      jq -r '.Tags[]')
-
-            # New version? Push quay.io/$REPONAME/stable:vX.X.X and :latest
-            if ! fgrep -qx "$VERSION" <<<"$ALLTAGS"; then
-              # Assume version-tag is also the most up to date (i.e. "latest")
-              FQIN="$REPONAME_QUAY_REGISTRY/stable:$VERSION,$REPONAME_QUAY_REGISTRY/stable:latest"
-            else # Not a new version-tagged image
-              # Assume other contents changed, so this is the "new" latest.
-              FQIN="$REPONAME_QUAY_REGISTRY/stable:latest"
-            fi
+            # Push both new|updated version-tag and latest-tag FQINs
+            FQIN="$REPONAME_QUAY_REGISTRY/stable:$VERSION,$REPONAME_QUAY_REGISTRY/stable:latest"
           elif [[ "${{ matrix.source }}" == 'testing' ]]; then
             # Assume some contents changed, always push latest testing.
             FQIN="$REPONAME_QUAY_REGISTRY/testing:latest"
@@ -129,16 +118,8 @@ jobs:
         run: |
           VERSION='v${{ steps.sniff_test.outputs.version }}'
           # workaround vim syntax-highlight bug: '
-          ALLTAGS=$(skopeo list-tags \
-                    docker://$CONTAINERS_QUAY_REGISTRY/$REPONAME | \
-                    jq -r '.Tags[]')
-
-          # New version? Push quay.io/containers/$REPONAME:vX.X.X and latest
-          if ! fgrep -qx "$VERSION" <<<"$ALLTAGS"; then
-            FQIN="$CONTAINERS_QUAY_REGISTRY/$REPONAME:$VERSION,$CONTAINERS_QUAY_REGISTRY/$REPONAME:latest"
-          else # Not a new version-tagged image, only update latest.
-            FQIN="$CONTAINERS_QUAY_REGISTRY/$REPONAME:latest"
-          fi
+          # Push both new|updated version-tag and latest-tag FQINs
+          FQIN="$CONTAINERS_QUAY_REGISTRY/$REPONAME:$VERSION,$CONTAINERS_QUAY_REGISTRY/$REPONAME:latest"
           echo "::warning::Pushing $FQIN"
           echo "::set-output name=fqin::${FQIN}"
           echo '::set-output name=push::true'
@@ -147,15 +128,31 @@ jobs:
         run: |
           # This is a really hacky/strange workflow idiom, required
           # for setting multi-line $LABELS value for consumption in
-          # a future step.
+          # a future step.  There is literally no cleaner way to do this :<
           # https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#multiline-strings
-          cat << EOF | tee -a $GITHUB_ENV
-          LABELS<<DELIMITER
-          org.opencontainers.image.source=https://github.com/${{ github.repository }}.git
-          org.opencontainers.image.revision=${{ github.sha }}
-          org.opencontainers.image.created=$(date -u --iso-8601=seconds)
-          DELIMITER
-          EOF
+          function set_labels() {
+            echo 'LABELS<<DELIMITER' >> "$GITHUB_ENV"
+            for line; do
+                echo "$line" | tee -a "$GITHUB_ENV"
+            done
+            echo "DELIMITER" >> "$GITHUB_ENV"
+          }
+
+          declare -a lines
+          lines=(\
+            "org.opencontainers.image.source=https://github.com/${GITHUB_REPOSITORY}.git"
+            "org.opencontainers.image.revision=${GITHUB_SHA}"
+            "org.opencontainers.image.created=$(date -u --iso-8601=seconds)"
+          )
+
+          # Only the 'stable' matrix source obtains $VERSION
+          if [[ "${{ matrix.source }}" == "stable" ]]; then
+            lines+=(\
+              "org.opencontainers.image.version=${{ steps.sniff_test.outputs.version }}"
+            )
+          fi
+
+          set_labels "${lines[@]}"
 
       # Separate steps to login and push for $REPONAME_QUAY_REGISTRY and
       # $CONTAINERS_QUAY_REGISTRY are required, because 2 sets of credentials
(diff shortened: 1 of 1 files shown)


cevich:
@rhatdan @TomSweeneyRedHat @edsantiago PTAL at this draft when you have a full cup of caffeinated liquid.  I honestly question the value of this change compared to just always pushing the latest + version-tagged image daily.  In other words, I think new packages come out frequently enough, the two schemes may be equivalent but the one in this PR is significantly more complex.

As an alternative (also addressing #3365), when overwriting an existing version-tagged image, we could instead do it with an image built from a very simple docker-file:

```
FROM <existing image FQIN>
RUN dnf upgrade -y && dnf clean all
```

The effect would be, users whom already have the old image cached, only need to pull the new layer.  The (big) down-side to this approach is, even more complexity in an already very difficult to review and maintain github action workflow.  Remember: I'm working on a Cirrus-CI based replacement for all this, so we could decide to do something like this idea, there instead.

edsantiago (approved):
LGTM

cevich:
> I'm going to close this and replace it

Chose to just re-implement in place.

cevich:
@edsantiago I hate to say your previous 'LGTM' may no-longer be valid.  I re-wrote this vs the original draft, so it will simply always push the version-tagged images daily.  I plan to deal with the churn more elegantly in the Cirrus-cron based implementation.

.github/workflows/multi-arch-build.yaml
```diff
@@ -157,6 +138,19 @@ jobs:
...
 
+      # Encode the $VERSION value in metadata so it's available on
+      # 'latest' image and/or if end-users re-tag the version-tagged image.
+      - name: Append version to LABELS env. var. value for stable images
+        # Only the 'stable' matrix source obtains $VERSION
+        if: matrix.source == 'stable'
+        run: |
+          cat << EOF | tee -a $GITHUB_ENV
+          LABELS<<DELIMITER
+          $LABELS
```
edsantiago, line 149: This is super ultra freaky.  Is there any way you could combine this with the previous ('Define LABELS') one? 

Reason: (1) it's a little distressing to have two such similar complicated-gobbledygook stanzas; and (2) it isn't entirely clear to me as a reader how/where `$LABELS` will be interpreted. It actually looks to me like it will be interpreted in the shell context of this specific `run`, which makes me wonder more about how this magicky stuff works, because the previous stanza only writes to a file but doesn't actually eval anything, and anyhow I really don't want to learn more about this github workflow stuff. I would really like to just accept the previous stanza, assume it accepts, and then sweep it out of mind. Anyhow, this `$LABELS` here is consuming all of my attention, preventing me from looking at anything else.

cevich: > This is super ultra freaky.

Welcome to Github Actions!

> it's a little distressing

Yes.  The alternative is to have slightly different looking complicated-gobbledygook stanzas since because the `.version` one is only applicable when `matrix.source == 'stable'.  The only other thing I can think of is to use some dummy/filler value for `$VERSION` when `matrix.source != 'stable'.  But that will also be stress-inducing in different ways :confounded: 

> how this magicky stuff works

STOP! Just STOP! It's Github, we're required to love it!  Please smile, open wide, and graciously accept what is on the spoon, for your "enjoyment" :angry: 

> I really don't want to learn more about this github workflow stuff

Trust me, you don't.  It's a horrible, useless mess, peddled to the masses under a guise of "Innovation" and "Exciting" and "Easy to use!"  I'm really (REALLY) angry with myself for accepting the initial implementation and not insisting we do it in Cirrus with simple bash scripts.

> is consuming all of my attention, preventing me from looking at anything else.

I understand, I do. But I also think it's useful (to users and future re-implementation) to start encoding the stable version into metadata.  Let me stare at it again, with yet-another-cup-o-coffee. Perhaps a less fugly solution will occur to me...

cevich:
Wow...I searched and searched...there is literally no other way to set an env. var. value containing more than one line :sob:

.github/workflows/multi-arch-build.yaml
cevich: ...okay, this is the best I can do for now.

cevich:
@edsantiago is that now a little more sensible?

(Remember, I still intend to trash this completely with a simple bash implementation running under Cirrus-cron)

cevich:
(small tweak...replaced one more stupid "context expression" with a simple built-in env. variable)

.github/workflows/multi-arch-build.yaml
```diff
@@ -87,25 +87,14 @@ jobs:
...
-            # New version? Push quay.io/$REPONAME/stable:vX.X.X and :latest
-            if ! fgrep -qx "$VERSION" <<<"$ALLTAGS"; then
-              # Assume version-tag is also the most up to date (i.e. "latest")
-              FQIN="$REPONAME_QUAY_REGISTRY/stable:$VERSION,$REPONAME_QUAY_REGISTRY/stable:latest"
-            else # Not a new version-tagged image
-              # Assume other contents changed, so this is the "new" latest.
-              FQIN="$REPONAME_QUAY_REGISTRY/stable:latest"
-            fi
+            # Push both new|updated version-tag and latest-tag FQINs
+            FQIN="$REPONAME_QUAY_REGISTRY/stable:$VERSION,$REPONAME_QUAY_REGISTRY/stable:latest"
```
TomSweeneyRedHat, line 97: OK, long day of staring at a lot of zeroes and ones.  Should we be pushing both here?  I expected just `latest` unless the `$VERSION` variant wasn't present.

.github/workflows/multi-arch-build.yaml
```diff
@@ -147,15 +128,31 @@ jobs:
...
+          )
+
+          # Only the 'stable' matrix source obtains $VERSION
+          if [[ "${{ matrix.source }}" == "stable" ]]; then
+            lines+=(\
+              "org.opencontainers.image.version=${{ steps.sniff_test.outputs.version }}"
+            )
+          fi
+
+          set_labels "${lines[@]}"
```
edsantiago, line 155: Thank you. I think this is much more readable and futureproof.

.github/workflows/multi-arch-build.yaml
edsantiago: AIUI the intention is to re-push all the time: even though our part isn't changing, underlying parts of the image itself may be updated upstream to (e.g.) address CVEs.

edsantiago:
LGTM but I'll defer slash-lgtm pending resolution of @TomSweeneyRedHat's concerns

.github/workflows/multi-arch-build.yaml
cevich: That's correct.  In a future Cirrus-based re-implementation, I'd like to solve this in a slightly more elegant way.  For now, this is one of those "Best of the worst" alternatives :disappointed:

cevich: Note: When we release a new version, the constant daily-updates of the (now) old version will stop.  So new CVEs will persist there for the joy and happiness of all :confused:

.github/workflows/multi-arch-build.yaml
cevich: Oh good, that means it will likely be more readable to other people too.

.github/workflows/multi-arch-build.yaml
TomSweeneyRedHat: Maybe we should chat this out at the water cooler tomorrow.  I only want to update versioned images when and if they contain a CVE fix.  If that means we need to do that by hand, I'm fine with that.

TomSweeneyRedHat:
I'm not hip on the version overwriting, let's talk.  Perhaps with @rhatdan if we can.  The rest looks fine to me.

.github/workflows/multi-arch-build.yaml
rhatdan: I would rather automate it and not have humans involved.

cevich:
Ref: [Manual run after merge.](https://github.com/containers/buildah/actions/runs/1072082041)

TomSweeneyRedHat:
Just for the record, I'll whine one last time on this.  In the past, we had been updating the versioned images on a regular basis and had a number of customers say they could not use them until they became relatively static.  I do believe this change is a backstep and the wrong approach.

rhatdan:
Ok, let's see if they complain.  Perhaps we drop it back to once a week or once a month.  But we don't want to have these images become ones that scanners report as full of CVEs.

edsantiago:
@TomSweeneyRedHat I agree with you, but this is an ugly and unavoidable reality of the container-images world: once we push an image, we become responsible for everything inside it. Even though the buildah image (probably) doesn't use most of what's shipped in the base, if there's a privilege escalation zeroday in `/bin/true` we need to update our image.

With that said, I keep thinking there's a simple solution: strip our images to just the bare minimum (buildah + libraries). Can someone remind me why we don't do that? (Maybe this is a cabal topic).

rhatdan:
Theoretically we could, but we are not really sure what people do within the container. For example most expect bash to be there.  Not sure what else.

cevich:
> Theoretically we could, but we are not really sure what people do within the container. For example most expect bash to be there. Not sure what else.

This is exactly the thing.  The container image is **NOT** a pure buldah distribution mechanism (that what packages are for).  It's meant to be a small environment where you can actually do some minimal buildah-ing.  For that, we need the extra included bits...which means we're subjected to not placing users in known danger :confused: 

Also remember...this is not any kind of different/new problem that almost every other single container image in the universe faces.  You get 'latest', 'tags', and digest-based pulls.  You do ***NOT*** get both conveinence/stability AND security at the same time.  So I think we can just let the whiners, whine.

PR #3387, opened 2021-07-16
[1.21] vendor containers/common@v0.38.16

vrothberg:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.12
 
 require (
 	github.com/containernetworking/cni v0.8.1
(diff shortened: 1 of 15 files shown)


TomSweeneyRedHat:
Do you want to target this for RHEL 8.4.0.2?  I think we need to update the man pages with the change to pull before hand, and I'm leery about making this change just before the deadline.  I'm fine with the Main branch though.  Thoughts?

vrothberg:
Yes, I intended it for 8.4.0.2 so that Podman and Buildah share the same behavior and have the same version of c/common.

Issue #3392, opened 2021-07-19
Confusing error on failure to pull base image using reference name


cevich:
**Description**

[During periodic execution of buildah system tests on the ***release-1.21*** branch (job logs)](https://cirrus-ci.com/task/5576282244120576?logs=integration_test#L944), a system-test of the Containerfile `FROM foo AS bar` handling failed with a very unhelpful error.

**Steps to reproduce the issue:**
1. I was not able to reproduce this issue with the released version of buildah 1.21.1
2. The command run by the test was essentially `buildah bud --layers .` using:

```Dockerfile
FROM alpine AS base
RUN touch /1
ENV LOCAL=/1
RUN find $LOCAL

FROM base
RUN find $LOCAL
```

**Describe the results you received:**

The first `FROM` section worked fine, then due to some unknown/unreproducable condition, an attempt was made to pull the reference-named image (`base`):

```
error creating build container: 3 errors occurred while pulling:
  * Error initializing source docker://base:latest: Error reading manifest latest in quay.io/libpod/base: unauthorized: access to the requested resource is not authorized
```

**Describe the results you expected:**

1. The reference named image should never ever be pulled from a remote registry.  It would be nice if some additional protection was in place guarantee this.
2. The reference name image should not appear in any error message, or be clearly annotated (i.e. by sha).  Since this is the name given to the output, not an actual registry image.

**Output of `rpm -q buildah` or `apt list buildah`:**

[Buildah was compiled from source, on the `release-1.21` branch.](https://github.com/containers/buildah/commit/7f9540d2ab9037b7e175a4d1b36e85fc6541aeb5)

**Output of `buildah version`:**

```
Version:         1.21.3
Go Version:      go1.15.11
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        0.4.0
libcni Version:  v0.8.1
image Version:   5.12.0
Git Commit:      7f9540d
Built:           Thu Jul 15 22:11:31 2021
OS/Arch:         linux/amd64
```

**Output of `cat /etc/*release`:**

CI VM Image c6032583541653504 for Ubuntu 20.10

**Output of `uname -a`:**

N/A

**Output of `cat /etc/containers/storage.conf`:**

N/A

cevich:
***Clarification:*** I think this issue might be already fixed, I'd simply like an expert to confirm.  Also comments on suggested improvements are welcome (I'm happy to open RFE's on them if deemed necessary).

[...]

PR #3393, opened 2021-07-20
imagebuildah.stageExecutor.prepare(): remove pseudonym check

nalind:
#### What type of PR is this?

/kind bug

In `prepare()`, don't check if the image name that it's passed is a pseudonym for the result of a stage in the Dockerfile.  Its callers already did that.

When `execute()` knows that the image it's told to use as a base is a pseudonym for the result of another stage in the Dockerfile, force the pull policy to "never" to prevent an error when `--pull-always=true`.

Make `imagebuildah.Mount` a type alias instead of its own type, since we never needed it to be a distinct type, but we didn't make it an alias when we started using them.

#### How to verify it

New integration test (for that second one)!

None

```
None
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -434,7 +434,7 @@ func (s *StageExecutor) Run(run imagebuilder.Run, config docker.Config) error {
 		Runtime:          s.executor.runtime,
 		Args:             s.executor.runtimeArgs,
 		NoPivot:          os.Getenv("BUILDAH_NOPIVOT") != "",
-		Mounts:           convertMounts(s.executor.transientMounts),
+		Mounts:           append([]Mount{}, s.executor.transientMounts...),
 		Env:              config.Env,
 		User:             config.User,
 		WorkingDir:       config.WorkingDir,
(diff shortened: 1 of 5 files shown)


imagebuildah/stage_executor.go
```diff
@@ -690,9 +683,11 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 	if isStage, err := s.executor.waitForStage(ctx, base, s.stages[:s.index]); isStage && err != nil {
 		return "", nil, err
 	}
+	pullPolicy := s.executor.pullPolicy
 	s.executor.stagesLock.Lock()
 	if stageImage, isPreviousStage := s.executor.imageMap[base]; isPreviousStage {
 		base = stageImage
+		pullPolicy = define.PullNever
```
rhatdan, line 690: Could you add a comment here on why this is being done.

rhatdan:
LGTM, other then request for comment.
Does this fix an issue?

vrothberg (commented):
LGTM, thanks @nalind

vrothberg:
> Does this fix an issue?

Fixes https://github.com/containers/buildah/issues/3392

imagebuildah/stage_executor.go
nalind: Adding to the block comment above.

umohnani8:
LGTM

cevich:
> Fixes #3392

Yay :tada:  Thank you so much for helping track down and fix this.  Thanks to @vrothberg and @edsantiago and @rhatdan for helping as well.  What a tricky race!  I'm so happy the root-cause is fixed, the code is cleaner, and it's always great to have a new test.

PR #3399, opened 2021-07-22
Fix handling of --restore shadow-utils

rhatdan:

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -98,7 +98,7 @@ smoke_task:
     # the git-validate tool which are difficult to debug and fix.
     skip: $CIRRUS_PR == ''
 
-    timeout_in: 10m
+    timeout_in: 30m
 
     setup_script: '${SCRIPT_BASE}/setup.sh |& ${_TIMESTAMP}'
     build_script: '${SCRIPT_BASE}/build.sh |& ${_TIMESTAMP}'
--- a/Makefile
+++ b/Makefile
@@ -107,6 +107,7 @@ validate: install.tools
 	./tests/validate/git-validation.sh
 	./hack/xref-helpmsgs-manpages
 	./tests/validate/pr-should-include-tests
+	./tests/validate/buildahimages-are-sane
 
 .PHONY: install.tools
 install.tools:
--- a/contrib/buildahimage/centos7/Dockerfile
+++ b/contrib/buildahimage/centos7/Dockerfile
@@ -8,7 +8,7 @@ FROM centos:7
 
 # Remove directories used by yum that are just taking
 # up space.
-RUN useradd build; yum -y update; rpm --restore --quiet shadow-utils; yum -y install buildah fuse-overlayfs xz; rm -rf /var/cache /var/log/dnf* /var/log/yum.*;
+RUN useradd build; yum -y update; rpm --restore shadow-utils 2>/dev/null; yum -y install buildah fuse-overlayfs xz; rm -rf /var/cache /var/log/dnf* /var/log/yum.*;
 
 ADD https://raw.githubusercontent.com/containers/buildah/main/contrib/buildahimage/stable/containers.conf /etc/containers/
 
--- a/contrib/buildahimage/stable/Dockerfile
+++ b/contrib/buildahimage/stable/Dockerfile
@@ -11,7 +11,7 @@ FROM registry.fedoraproject.org/fedora:latest
 # Don't include container-selinux and remove
 # directories used by yum that are just taking
 # up space.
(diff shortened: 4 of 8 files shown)


rhatdan:
@edsantiago PTAL, I am a little concerned about the test.

tests/contrib.bats
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bats
+
+load helpers
+
+@test "bud with a path to a Dockerfile (-f) containing a non-directory entry" {
```
nalind, line 5: This description doesn't look right for what this test is doing.

rhatdan: Lets let it run through, I think @edsantiago is going to have some ideas on how better to handle this.  Perhaps copy the Dockefile into tests and then modify the FROM line.

tests/contrib.bats
```diff
@@ -0,0 +1,10 @@
...
+  run_buildah bud -t buildah ../contrib/buildahimage/stable
```
edsantiago, line 6: I don't see this working in CI, because the `contrib` directory isn't shipped in the `-tests` rpm

tests/contrib.bats
```diff
@@ -0,0 +1,10 @@
...
+  run_buildah from buildah
+  run_buildah run $output rpm -qV shadow-utils
```
edsantiago, line 8: This is probably a really stupid question, but might it not be better to run this `rpm -qV` in the magic github-workflow script that creates the remote images?

rhatdan: An then blow up there?  My goal is to catch changes that can end up with  a bad image, when users checkin fixes to the contrib Dockerfiles.

tests/contrib.bats
rhatdan: Can I just turn it off for gating tests?

edsantiago:
> My goal is to catch changes that can end up with a bad image, when users checkin fixes to the contrib Dockerfiles.

Thank you, I had not understood that. Given that, it doesn't sound like this is a good candidate for the BATS tests which run on all arches, adding more opportunities for flaking.

Would it make sense to add this to the `smoke` CI target? That runs once only, builds a binary. There could be a check such as:
```
if (any files have been touched under contrib/buildahimage) then
    for each modified subdirectory
        buildah bud, buildah from, buildah run rpm -qV
```

rhatdan:
SGTM, could you send me a patch that does this?

edsantiago:
@rhatdan, submitted as https://github.com/rhatdan/buildah/pull/12 (it took slightly longer than I had predicted)

edsantiago:
Oooops, sorry, I didn't notice this: https://github.com/containers/buildah/blob/main/.cirrus.yml#L101

For a PR like this one, with every Containerfile changing, I think we need 25-30m. OTOH that seems counter to the purpose of a smoke test, which one would want to fail fast. OT3H I'm reluctant to make this a separate Cirrus task all on its own. I think bumping up the smoke timeout is the least-awful approach. @rhatdan @nalind @cevich WDYT?

rhatdan:
Should we only run the smoke test if the ~/contrib/buildahimage directory changed?

edsantiago:
No, the smoke test does much much much more: 
https://github.com/containers/buildah/blob/3821391ea6fa0d2a5360bfc6da9fcf9e29133c72/contrib/cirrus/test.sh#L22-L37
https://github.com/containers/buildah/blob/3821391ea6fa0d2a5360bfc6da9fcf9e29133c72/Makefile#L105-L110

rhatdan:
@edigaryev I meant this (The buildahimage) smoke test.

edsantiago:
Oh! It does that already: it only runs if there are changed files under `contrib/buildahimages`. If there are none, the test exits within milliseconds.

rhatdan:
Excellent, then I would say lets increase the timeout.

edsantiago:
Smoke [ran and passed](https://cirrus-ci.com/task/4666609362010112?logs=validate_test#L122) (in just over 10 minutes, reminding me that I'm awful at estimating run times). The log confirms that the script tested `centos7`, `stable`, `testing`, and `upstream`; and the timestamps confirm that each took a reasonable time (~1m, although upstream took 3). I call this good.

cevich:
I have no problem with bumping up the timeout a bit.

edsantiago:
LGTM but I'd like one more set of eyeballs on the new script and CI addition

TomSweeneyRedHat:
A little more description would be nice so the poor release notes writer could write this up more easily would be nice.
I've tired eyes too, so figure between @edsantiago and I we've one good pair.
LGTM

PR #3402, opened 2021-07-26
bud: teach --platform to take a list

nalind:
#### What type of PR is this?

/kind feature

Add a `pkg/parse.PlatformsFromOptions()` which understands a "variant" value as an optional third value in an OS/ARCH[/VARIANT] argument value, which accepts a comma-separated list of them, and which returns a list of platforms.

Add a `define.BuildOptions.JobSemaphore` which an `imagebuildah` executor will use in preference to one that it might allocate for itself.

Teach "from" and "pull" about the `--platform` option and add integration tests for them, warning if `--platform` was given multiple values.

In `main()`, allocate a `JobSemaphore` if the number of jobs is not 0 (which we treat as "unlimited", and continue to allow executors to do).

In `addManifest()`, take a lock on the manifest list's image ID so that we don't overwrite changes that another thread might be making while we're attempting to make changes to it.  In `main()`, create an empty list if the list doesn't already exist before we start down this path, so that we don't get two threads trying to create that manifest list at the same time later on.  Two processes could still try to create the same list twice, but it's an incremental improvement.

Finally, if we've been given multiple platforms to build for, run their builds concurrently and gather up their results.

#### How to verify it

New integration test!

None

```release-note
The `--platform` option to `buildah bud` can now take a comma-separated list or be specified multiple times, for use in combination with the `--manifest` option.
```

### DIFF
--- a/cmd/buildah/bud.go
+++ b/cmd/buildah/bud.go
@@ -13,10 +13,16 @@ import (
 	buildahcli "github.com/containers/buildah/pkg/cli"
 	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/buildah/util"
+	"github.com/containers/common/libimage"
+	"github.com/containers/common/libimage/manifests"
 	"github.com/containers/common/pkg/auth"
+	"github.com/containers/image/v5/manifest"
+	"github.com/containers/storage"
+	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"golang.org/x/sync/semaphore"
 )
 
 type budOptions struct {
@@ -299,7 +305,7 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budOptions) error {
 	}
 	namespaceOptions.AddOrReplace(usernsOption...)
 
-	imageOS, arch, err := parse.PlatformFromOptions(c)
+	platforms, err := parse.PlatformsFromOptions(c)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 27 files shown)


nalind:
/retitle bud: add a --platforms that take a list

docs/buildah-bud.md
```diff
@@ -406,12 +406,17 @@ that the PID namespace in which `buildah` itself is being run should be reused,
...
-`--platform` is set, then the values of the `--arch` and `--os` options will be
-overridden.
+Set the OS/ARCH of the built image (and its base image, if your build uses one)
+to the provided value instead of using the current operating system and
+architecture of the host (for example `linux/arm`). If `--platform` or
+`--platforms` is set, then the values of the `--arch` and `--os` options will
+be overridden.
+
+When more than one value is supplied to `--platforms`, the `--manifest` option
+should be used instead of the `--tag` option.
```
TomSweeneyRedHat, line 419: Any way to provide a pointer to the valid values for OS and ARCH?  An added example or two would probably be beneficial too.

pkg/parse/parse.go
```diff
@@ -612,19 +627,39 @@ func PlatformFromOptions(c *cobra.Command) (os, arch string, err error) {
...
-
-	return os, arch, nil
+	add := true
+	for _, pf := range platforms {
+		if os == "" || pf.OS == os && arch == "" || pf.Arch == arch {
+			add = false
+		}
+	}
+	if add {
+		platforms = append([]struct{ OS, Arch, Variant string }{{os, arch, ""}}, platforms...)
```
TomSweeneyRedHat, line 660: Is casing a concern?

nalind: Can you elaborate on that?  I don't think I follow the question.

docs/buildah-bud.md
nalind: Will add pointers.

pkg/parse/parse.go
TomSweeneyRedHat: Just wondering if someone does Fedora/arm, fedora/ARM, FEDORA/Arm, or what have you, if the system will handle the casing of the values to the right case, or will it just spit them out?

docs/buildah-pull.md
```diff
@@ -60,9 +60,19 @@ If an image needs to be pulled from the registry, suppress progress output.
...
+architecture of the host (for example `linux/arm`).  If `--platform`
+is set, then the values of the `--arch`, `--os`, and `--variant` options will
+be overridden.
 
-Set the OS of the image to be pulled to the provided value instead of using the current operating system of the host.
+OS/ARCH pairs are those used by the Go Programming Language.  In several cases
+the ARCH value for a platform differs from one produced by other tools such as
+the `arch` command.  Valid OS and architecture name combinations are listed as
+values for $GOOS and $GOARCH at https://golang.org/doc/install/source#environment,
+and can also be found by running `go tool dist list`.
```
TomSweeneyRedHat, line 75: ty!

pkg/parse/parse.go
nalind: OS and architecture names are case-sensitive, yes.

docs/buildah-bud.md
rhatdan: Is this a comma separate list of platforms or do I need to specify it multiple times?

nalind: Just updated this.  Now, it's both.

nalind: Updated it again.  `--platform` now handles everything, warning when it's given more than one value in `pull` (a single tagged reference can't be applied to multiple images, so we'll want to think a while about changing that), or `from`.

nalind:
/retitle bud: teach --platform to take a list

rhatdan:
Much better but lot's of test failures.

rhatdan:
LGTM
@giuseppe @saschagrunert @vrothberg @TomSweeneyRedHat @ashley-cui @umohnani8 @mtrmac PTAL

vrothberg:
I hope to find time later today to review

nalind:
Rebased.

docs/buildah-bud.md
nalind: The [docs](https://pkg.go.dev/github.com/spf13/pflag?utm_source=godoc#StringSlice) say we don't need to do splitting of comma-separated values for `StringSlice` flags ourselves, so just dropped those bits.

commit.go
```diff
@@ -183,6 +183,12 @@ func (b *Builder) addManifest(ctx context.Context, manifestName string, imageSpe
 		create = true
 		list = manifests.Create()
 	} else {
+		locker, err := manifests.LockerForImage(b.store, manifestList.ID())
```
vrothberg, line 186: Nit: I would have preferred to move that into a `manifestList.{L,Unl}ock()` but it does the trick.

vrothberg (commented):
/lgtm
/hold
Nice work!

pkg/parse/parse.go
```diff
@@ -570,21 +570,30 @@ func SystemContextFromOptions(c *cobra.Command) (*types.SystemContext, error) {
...
+		var specs []string
+		if specs, err = c.Flags().GetStringSlice("platform"); err != nil {
+			return nil, err
+		}
+		if len(specs) == 0 || specs[0] == "" {
+			return nil, errors.Errorf("unable to parse --platform value %v", specs)
 		}
+		platform = specs[0]
+	}
+	if platform != "" {
```
mtrmac, line 584: Is there a reason to make this a separate conditional?

Entirely up to you — just noticing that this doesn’t look like the prevalent coding style.

pkg/parse/parse.go
```diff
@@ -612,19 +636,28 @@ func PlatformFromOptions(c *cobra.Command) (os, arch string, err error) {
...
-			os = selectedOS
+		platforms = nil
+		platformSpecs, err = c.Flags().GetStringSlice("platform")
+		if err != nil {
+			return nil, errors.Wrap(err, "unable to parse platform")
 		}
 	}
-
-	return os, arch, nil
+	for _, pf := range platformSpecs {
```
mtrmac, line 653: This could also be inside the `("platform").Changed` conditional.

Entirely up to you.

cmd/buildah/bud.go
```diff
@@ -309,75 +315,120 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budOptions) error {
...
+					Squash:                  iopts.Squash,
+					SystemContext:           systemContext,
+					Target:                  iopts.Target,
+					TransientMounts:         iopts.Volumes,
+					OciDecryptConfig:        decConfig,
+					Jobs:                    &iopts.Jobs,
+					JobSemaphore:            jobSemaphore,
+				}
+				if iopts.IgnoreFile != "" {
+					excludes, err := parseDockerignore(iopts.IgnoreFile)
```
mtrmac, line 406: Could this be done before the loop, so that it fails only once?

cmd/buildah/bud.go
```diff
@@ -309,75 +315,120 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budOptions) error {
...
+					}
+					options.Excludes = excludes
+				}
+				if c.Flag("timestamp").Changed {
+					timestamp := time.Unix(iopts.Timestamp, 0).UTC()
+					options.Timestamp = &timestamp
+				}
+				if iopts.Quiet {
+					options.ReportWriter = ioutil.Discard
+				}
```
mtrmac, line 418: This also might be possible to do only once, OTOH that would require a larger restructuring.

cmd/buildah/bud.go
```diff
@@ -309,75 +315,120 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budOptions) error {
...
 			return err
 		}
-		options.Excludes = excludes
-	}
-	if c.Flag("timestamp").Changed {
-		timestamp := time.Unix(iopts.Timestamp, 0).UTC()
-		options.Timestamp = &timestamp
 	}
+	for _, platform := range platforms {
+		func() {
```
mtrmac, line 345: What does this closure do?

cmd/buildah/bud.go
```diff
@@ -309,75 +315,120 @@ func budCmd(c *cobra.Command, inputArgs []string, iopts budOptions) error {
...
+			platform := platform
+			builds.Go(func() error {
+				platformContext := *systemContext
+				systemContext = &platformContext
```
mtrmac, line 349: AFAICS `systemContext` is a variable shared by all the goroutines, i.e. they stomp on each other. Maybe just a `:=` would fix it, but a `platformContextPtr := &platformContext` would be more explicit — or just work with `planformContext` and `&platformContext` only.

mtrmac (commented):
ACK overall (with my very limited understanding of the codebase).

Note the `systemContext` comment at least.

cmd/buildah/bud.go
nalind: Oof, yes, that should be a `:=`.  Fixing it.

cmd/buildah/bud.go
nalind: Yes, moving it.

cmd/buildah/bud.go
nalind: Moving the initializer for `timestamp` out of the loop.

pkg/parse/parse.go
nalind: That's a leftover from an earlier revision.  Merging them.

cmd/buildah/bud.go
nalind: At the time, I think it was for ensuring that each goroutine we ran with builds.Go() was getting the `platform` value from the iteration of the loop that launched it, since otherwise the iterator value would change because it was being shared.  But I don't think we need to be creating a new scope just for that since we're already creating a new variable for each iteration.  Removing it.

mtrmac (commented):
LGTM

vrothberg (commented):
/lgtm

PR #3417, opened 2021-08-02
Bump 1.22.0

TomSweeneyRedHat:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -221,6 +221,316 @@
   * Reset upstream branch to dev version
   * If destination does not exists, do not throw error
 
+## v1.22.0 (2021-08-02)
+    c/image, c/storage, c/common vendor before Podman 3.3 release
+    WIP: tests: new assert()
+    Proposed patch for 3399 (shadowutils)
+    Fix handling of --restore shadow-utils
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -221,6 +221,316 @@
   * Reset upstream branch to dev version
   * If destination does not exists, do not throw error
 
+## v1.22.0 (2021-08-02)
```
nalind, line 224: Why does this duplicate some of the content that's above it, implying that this release is older than the releases described above?

TomSweeneyRedHat: IDK @nalind.  The changelog.txt was funky.  After I ran the release.sh file, it showed only the entries for 1.19 and lower, so I added all the rest back in, but you're right, its duplicated here now.

TomSweeneyRedHat: And actually, this is CHANGELOG.md, not changelog.txt.  I must have simply cut/pasted badly.  If this is the only change needed, would you be OK with merging this in now and then I'll follow up and change this file.  I don't think it's included in the rpm, only shown on GitHub.

PR #3424, opened 2021-08-04
Fix resolv.conf content with run --net=private

### LINKED ISSUE #2780: podman-build network fails DNS lookups; podman-run works
**Description**

When using a VPN (OpenVPN), DNS lookups fail inside of `podman build`, but work inside of `podman run`. Routing works (`dig` outside of the container and use that IP inside), so I think some UDP tables might not be hooked up properly?

1. Set up a VPN in a network namespace (so it is the only available connection)
2. `podman build` fails DNS lookups
3. `podman run -it $previous_image_cache` works just fine

DNS lookup failure (timeout)

DNS lookup success.

```
buildah-1.16.2-1.fc33.x86_64
```

```
Version:         1.16.2
Go Version:      go1.15
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        0.4.0
libcni Version:  
image Version:   5.5.2
Git Commit:      
Built:           Wed Dec 31 19:00:00 1969
OS/Arch:         linux/amd64
```

```
Version:      2.1.1
API Version:  2.0.0
Go Version:   go1.15.2
Built:        Wed Oct  7 12:21:20 2020
OS/Arch:      linux/amd64
```

```
Fedora release 33 (Thirty Three)
NAME=Fedora
VERSION="33 (Thirty Three)"
ID=fedora
VERSION_ID=33
VERSION_CODENAME=""
PLATFORM_ID="platform:f33"
PRETTY_NAME="Fedora 33 (Thirty Three)"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:33"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f33/system-administrators-guide/"
REDHAT_BUGZILLA_PRODUCT_VERSION=33
REDHAT_SUPPORT_PRODUCT_VERSION=33
Fedora release 33 (Thirty Three)
Fedora release 33 (Thirty Three)
```

**Output of `uname -a`:**

```
Linux erythro 5.8.16-300.fc33.x86_64 #1 SMP Mon Oct 19 13:18:33 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

```
<Fedora default>
```

Luap99:
#### What type of PR is this?
Make sure to use the correct resolv.conf in case of systemd-resolved.
Also filter out 127... nameserver when run in private netns.

/kind bug

Dns resolution which --net=private does not work correct.

#### How to verify it
see test

Fixes #2780

```release-note

```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -207,7 +207,7 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	}
 
 	if !(contains(volumes, "/etc/resolv.conf") || (len(b.CommonBuildOpts.DNSServers) == 1 && strings.ToLower(b.CommonBuildOpts.DNSServers[0]) == "none")) {
-		resolvFile, err := b.addNetworkConfig(path, "/etc/resolv.conf", rootIDPair, b.CommonBuildOpts.DNSServers, b.CommonBuildOpts.DNSSearch, b.CommonBuildOpts.DNSOptions, namespaceOptions)
+		resolvFile, err := b.addResolvConf(path, rootIDPair, b.CommonBuildOpts.DNSServers, b.CommonBuildOpts.DNSSearch, b.CommonBuildOpts.DNSOptions, namespaceOptions)
 		if err != nil {
 			return err
 		}
@@ -557,19 +557,52 @@ func (b *Builder) setupMounts(mountPoint string, spec *specs.Spec, bundlePath st
 	return runTargets, nil
 }
 
-// addNetworkConfig copies files from host and sets them up to bind mount into container
-func (b *Builder) addNetworkConfig(rdir, hostPath string, chownOpts *idtools.IDPair, dnsServers, dnsSearch, dnsOptions []string, namespaceOptions define.NamespaceOptions) (string, error) {
-	stat, err := os.Stat(hostPath)
+// addResolvConf copies files from host and sets them up to bind mount into container
+func (b *Builder) addResolvConf(rdir string, chownOpts *idtools.IDPair, dnsServers, dnsSearch, dnsOptions []string, namespaceOptions define.NamespaceOptions) (string, error) {
+	resolvConf := "/etc/resolv.conf"
+
+	stat, err := os.Stat(resolvConf)
 	if err != nil {
 		return "", err
 	}
-	contents, err := ioutil.ReadFile(hostPath)
-	if err != nil {
+	contents, err := ioutil.ReadFile(resolvConf)
+	// resolv.conf doesn't have to exists
+	if err != nil && !os.IsNotExist(err) {
 		return "", err
 	}
 
+	netns := false
+	ns := namespaceOptions.Find(string(spec.NetworkNamespace))
+	if ns != nil && !ns.Host {
+		netns = true
+	}
+
+	nameservers := resolvconf.GetNameservers(contents, types.IPv4)
+	// check if systemd-resolved is used, assume it is used when 127.0.0.53 is the only nameserver
+	if len(nameservers) == 1 && nameservers[0] == "127.0.0.53" && netns {
+		// read the actual resolv.conf file for systemd-resolved
+		resolvedContents, err := ioutil.ReadFile("/run/systemd/resolve/resolv.conf")
+		if err != nil {
+			if !os.IsNotExist(err) {
+				return "", errors.Wrapf(err, "detected that systemd-resolved is in use, but could not locate real resolv.conf")
+			}
+		} else {
+			contents = resolvedContents
+		}
+	}
+
+	// Ensure that the container's /etc/resolv.conf is compatible with its
+	// network configuration.
+	if netns {
+		// FIXME handle IPv6
+		resolve, err := resolvconf.FilterResolvDNS(contents, true)
+		if err != nil {
+			return "", errors.Wrapf(err, "error parsing host resolv.conf")
+		}
+		contents = resolve.Content
+	}
 	search := resolvconf.GetSearchDomains(contents)
-	nameservers := resolvconf.GetNameservers(contents, types.IP)
+	nameservers = resolvconf.GetNameservers(contents, types.IP)
 	options := resolvconf.GetOptions(contents)
 
 	defaultContainerConfig, err := config.Default()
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

rhatdan:
@vrothberg @TomSweeneyRedHat @edsantiago @mheon @nalind @ashley-cui PTAL

tests/run.bats
```diff
@@ -558,39 +558,64 @@ function configure_and_check_user() {
...
+	cid=$output
+	run_buildah run --network=private $cid grep nameserver /etc/resolv.conf
+	# check that no 127... nameserver is in resolv.conf
+	assert "$output" !~ "^nameserver 127." "Container contains local nameserver"
+	assert "$nameservers" "Container nameservers match corret host nameservers"
+	if ! is_rootless; then
+		m=$(buildah mount $cid)
+		run cat $m/etc/resolv.conf
+		[ "$status" -eq 0 ]
+		expect_output --substring ""
```
edsantiago, line 591: I realize this is not your original code, but eek, that's a NOP. Suggested fix:
```diff
diff --git a/tests/run.bats b/tests/run.bats
index 3be0da75..659251bb 100644
--- a/tests/run.bats
+++ b/tests/run.bats
@@ -585,10 +585,9 @@ $output"
        assert "$output" !~ "^nameserver 127." "Container contains local nameserver"
        assert "$nameservers" "Container nameservers match corret host nameservers"
        if ! is_rootless; then
-               m=$(buildah mount $cid)
-               run cat $m/etc/resolv.conf
-               [ "$status" -eq 0 ]
-               expect_output --substring ""
+               run_buildah mount $cid
+               assert "$output" != ""
+               assert "$(< $output/resolv.conf)" = "" "resolv.conf is empty"
        fi
        run_buildah rm -a

```

tests/run.bats
```diff
@@ -558,39 +558,64 @@ function configure_and_check_user() {
...
+	if is_rootless; then
+		nameservers="nameserver 10.0.2.3
+$output"
+	fi
+	run_buildah from --quiet --pull=false --signature-policy ${TESTSDIR}/policy.json alpine
+	cid=$output
+	run_buildah run --network=private $cid grep nameserver /etc/resolv.conf
+	# check that no 127... nameserver is in resolv.conf
+	assert "$output" !~ "^nameserver 127." "Container contains local nameserver"
+	assert "$nameservers" "Container nameservers match corret host nameservers"
```
edsantiago, line 586: typo, should be "correct"

tests/run.bats
```diff
@@ -558,39 +558,64 @@ function configure_and_check_user() {
...
+	nameservers="$output"
+	run_buildah from --quiet --pull=false --signature-policy ${TESTSDIR}/policy.json alpine
+	cid=$output
+	run_buildah run --isolation=chroot --network=host $cid grep nameserver /etc/resolv.conf
+	assert "$nameservers" "Container nameservers match the hosts nameservers"
+	if ! is_rootless; then
+		m=$(buildah mount $cid)
+		run cat $m/etc/resolv.conf
+		[ "$status" -eq 0 ]
+		expect_output --substring ""
```
edsantiago, line 605: Same as above

tests/run.bats
```diff
@@ -558,39 +558,64 @@ function configure_and_check_user() {
...
 	expect_output "nameserver 110.110.0.110"
-        m=$(buildah mount $cid)
-	run cat $m/etc/resolv.conf
-	[ "$status" -eq 0 ]
-	expect_output --substring "nameserver 110.110.0.110"
+	if ! is_rootless; then
+		m=$(buildah mount $cid)
+		run cat $m/etc/resolv.conf
+		[ "$status" -eq 0 ]
+		expect_output --substring "nameserver 110.110.0.110"
```
edsantiago, line 617: Similar to above wrt the mount, but the `expect_output` is OK here.

tests/run.bats
```diff
@@ -558,39 +558,64 @@ function configure_and_check_user() {
...
-        m=$(buildah mount $cid)
-	run cat $m/etc/resolv.conf
-	[ "$status" -eq 0 ]
-	expect_output --substring ""
+	run grep nameserver /etc/resolv.conf
+	nameservers="$output"
+	run_buildah from --quiet --pull=false --signature-policy ${TESTSDIR}/policy.json alpine
+	cid=$output
+	run_buildah run --isolation=chroot --network=host $cid grep nameserver /etc/resolv.conf
+	assert "$nameservers" "Container nameservers match the hosts nameservers"
```
edsantiago, line 600: (minor): `host` (singular) or `host's` (possessive) would read better than `hosts` (plural)

edsantiago (changes requested):
Fix requested for preëxisting bug (not yours, but needs to be fixed). And, spelling nits.

Luap99:
@edsantiago Fixed your comments. BTW, is it correct that CI does not run these tests as rootless?

edsantiago:
I don't know if it's correct, but the last time I tried running buildah tests rootless there were many, many failures. That was almost two years ago so it might be time to reevaluate. I've put that on my to-do list.

run_linux.go
```diff
@@ -557,19 +557,52 @@ func (b *Builder) setupMounts(mountPoint string, spec *specs.Spec, bundlePath st
...
 
+	netns := false
+	ns := namespaceOptions.Find(string(spec.NetworkNamespace))
+	if ns != nil && !ns.Host {
+		netns = true
+	}
+
+	nameservers := resolvconf.GetNameservers(contents, types.IPv4)
+	// check if systemd-resolved is used, assume it is used when 127.0.0.53 is the only nameserver
+	if len(nameservers) == 1 && nameservers[0] == "127.0.0.53" && netns {
```
TomSweeneyRedHat, line 582: Is there a test for when only this nameserver is up?  I don't see one below, but it's late.

TomSweeneyRedHat: IDK, but can the user change the IP for the system nameserver?  It might be nice to have this value stored in a config file that could be tweaked.  I'm also a little twitchy about having a particular IP specified and having someone spoof it, but I can't see the harm atm in that case.

Luap99: No this is correct like this. It now does the same as podman. The problem with systemd-resolved is that it only sets `127.0.0.53` as nameserver in `/etc/resolv.conf`, this is a hard coded value AFAICT and there is no reason why someone should change this. The actual nameservers are in `/run/systemd/resolve/resolv.conf`. We have to read the actual nameservers in this case and filter the 127... addresses out because there are not reachable in a new netns. If there is no netns we can use the system one because 127.0.0.53 is reachable.
Fedora uses systemd-resolved so the CI should test this.

rhatdan:
/approve
@giuseppe @vrothberg @saschagrunert PTAL

edsantiago (approved):

edsantiago:
Followup to rootless-buildah-tests questions above: nope, doesn't work. Lots of chown errors, also lots of other errors. Maybe a project for a quiet week.

PR #3437, opened 2021-08-10
imagebuildah: move multiple-platform building internal

nalind:
#### What type of PR is this?

/kind api-change
/kind cleanup

Move multiple-platform build juggling logic from the CLI wrapper directly into the imagebuildah package, to make using it easier for packages that consume us as a library.

This requires reading Dockerfiles into byte slices so that we can re-parse them for each per-platform build, rather than parsing them directly, as we used to, since building modifies the parsed tree.

When building for multiple platforms, prefix progress log messages with the platform description.

#### How to verify it

Visible behavior shouldn't change.
[ NO NEW TESTS NEEDED ]

None

This reverts most changes to cmd/buildah/bud.go, so it should look very much like it did before we merged #3402, for example in v1.22.0.
imagebuildah.NewExecutor() should probably have never been an exported function, but it ~is~ was, so this is an API change.

```release-note
None
```

### DIFF
--- a/docs/buildah-bud.md
+++ b/docs/buildah-bud.md
@@ -44,7 +44,7 @@ Note: this information is not present in Docker image formats, so it is discarde
 
 **--arch**="ARCH"
 
-Set the ARCH of the image to be pulled to the provided value instead of using the architecture of the host. (Examples: arm, arm64, 386, amd64, ppc64le, s390x)
+Set the ARCH of the image to be built, and that of the base image to be pulled, if the build uses one, to the provided value instead of using the architecture of the host. (Examples: arm, arm64, 386, amd64, ppc64le, s390x)
 
 **--authfile** *path*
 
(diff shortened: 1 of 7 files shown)


docs/buildah-bud.md
```diff
@@ -287,7 +288,8 @@ those.
 
 **--iidfile** *ImageIDfile*
 
-Write the image ID to the file.
+Write the built image's ID to the file.  When `--platform` is specified more
+than once, only one of their IDs will be written to the file.
```
TomSweeneyRedHat, line 292: Is it the first one, or random chance?

rhatdan: Should be first, I would guess, coding might be easier to do last.  Other option would be to conflict them so that --iidfile can only be used if user specifies < 2 platforms.  (Which I think is the best solution.

rhatdan (approved):

docs/buildah-bud.md
nalind: There's no coordination, so currently I expect it's the one that wrote (or overwrote) it most recently.

rhatdan: Can we just block the option?

nalind: Sure, changing it.

nalind:
Tweaked the logic that creates the manifest list before starting up multiple builds so that it performs the same expansion the commit logic would when creating the manifest list.

rhatdan:
LGTM

PR #3442, opened 2021-08-11
Move DiscoverContainerfile to define directory, so podman can use it

rhatdan:

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ LIBSECCOMP_COMMIT := release-2.3
 
 EXTRA_LDFLAGS ?=
 BUILDAH_LDFLAGS := -ldflags '-X main.GitCommit=$(GIT_COMMIT) -X main.buildInfo=$(SOURCE_DATE_EPOCH) -X main.cniVersion=$(CNI_COMMIT) $(EXTRA_LDFLAGS)'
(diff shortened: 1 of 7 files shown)


nalind:
The point of moving type definitions into this directory is to avoid having consumers build code that they might not end up calling, and this seems to run counter to that goal.  Somewhere under `pkg` or in the `util` grab bag would be more appropriate.

TomSweeneyRedHat:
If @nalind likes the location, the changes LGTM.  Tests aren't happy yet though.

nalind:
I don't _love_ it (why must it be added in an existing import path?), but it's not a hill I'm particularly interested in dying on.  Either way, the commit message and pull request name need to be updated.

rhatdan:
@nalind Moved it to pkg/util

TomSweeneyRedHat:
@rhatdan the smoke test is smoking

nalind:
Needs to add `pkg/util/*.go` to the `SOURCES` list in the top-level Makefile, then LGTM.

TomSweeneyRedHat:
@rhatdan looks like a rebase is wanted.

PR #3456, opened 2021-08-16
imagebuildah: handle --manifest directly

nalind:
#### What type of PR is this?

/kind cleanup

When we're told to add built images to a manifest list, manipulate the list ourselves, so that if we're creating a list, we won't have a partially-populated list if some of the builds fail.

This also lets us include all of the platform information (including variant info, which we can't sniff out after the fact) that we were given when we started building the images.

#### How to verify it

New integration test!

None

```release-note
None
```


rhatdan:
LGTM

nalind:
Rebased.

TomSweeneyRedHat:
LGTM
@nalind, looks like you haven't rebased this one enough yet....

PR #3485, opened 2021-08-27
conformance: tighten up exception specifications

nalind:
#### What type of PR is this?

/kind bug

Modify how we process exceptions in conformance tests so that we don't unintentionally disable an attribute comparison (such as "mtime") on items below a directory when we try to skip that attribute comparison on the directory itself.

Fix some incorrect specifications for filesystem differences that we're supposed to ignore, and use the -t flag in more places where we RUN the `touch` command to create files that end up in the final image.

#### How to verify it

Updated conformance tests!

None

```release-note
None
```

### DIFF
--- a/tests/conformance/README.md
+++ b/tests/conformance/README.md
@@ -2,7 +2,7 @@
 
 # Buildah/Docker Conformance Test Suite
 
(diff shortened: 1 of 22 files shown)


TomSweeneyRedHat:
LGTM

PR #3487, opened 2021-08-27
Avoid rehashing and noop compression writer

kurt-google:
#### What type of PR is this?

/kind other

Follow up from https://github.com/containers/podman/issues/11236 performance improvement for large containers when not using compression.

When creating a new image we can avoid hashing before and after compression
when we know compression will not be used. This significantly improves speed
for hashing large containers with --compress=false

#### How to verify it

Follow reproduction steps in the podman issue. Basically create a container by copying a large file in.

None

I dont know if this breaks anything or how to test that, but the containers produced are usable so I assume it is working.

```release-note
Improve performance of --compress=false
```

### DIFF
--- a/image.go
+++ b/image.go
@@ -393,9 +393,18 @@ func (i *containerImageRef) NewImageSource(ctx context.Context, sc *types.System
 			rc.Close()
 			return nil, errors.Wrapf(err, "error opening file for %s", what)
 		}
-		destHasher := digest.Canonical.Digester()
+
 		counter := ioutils.NewWriteCounter(layerFile)
-		multiWriter := io.MultiWriter(counter, destHasher.Hash())
+		var destHasher digest.Digester
+		var multiWriter io.Writer
+		// Avoid rehashing when we do not compress.
+		if i.compression != archive.Uncompressed {
+			destHasher = digest.Canonical.Digester()
+			multiWriter = io.MultiWriter(counter, destHasher.Hash())
+		} else {
+			destHasher = srcHasher
+			multiWriter = counter
+		}
 		// Compress the layer, if we're recompressing it.
 		writeCloser, err := archive.CompressStream(multiWriter, i.compression)
 		if err != nil {


image.go
```diff
@@ -393,15 +393,23 @@ func (i *containerImageRef) NewImageSource(ctx context.Context, sc *types.System
...
+			// Compress the layer, if we're recompressing it.
+			writeCloser, err = archive.CompressStream(multiWriter, i.compression)
+			if err != nil {
+				layerFile.Close()
+				rc.Close()
+				return nil, errors.Wrapf(err, "error compressing %s", what)
+			}
+		} else {
+			destHasher = srcHasher
+			writeCloser = ioutils.NopWriteCloser(counter)
```
TomSweeneyRedHat, line 412: FWIW, I think we can move the `layerFile.Close()` line from line 442 up to line 413.5, just after this else.

TomSweeneyRedHat:
Any way to add a test?  If not, the CI will bark unless you add `[NO NEW TESTS NEEDED]' to the description.

kurt-google:
> Any way to add a test? If not, the CI will bark unless you add `[NO NEW TESTS NEEDED]' to the description.

Thanks, I added that to the commit description for now. I hadnt had a chance to look at the test suite so im not sure if there is a good way to test this. If there are any recommendations on how to test this id love to hear.

rhatdan:
@mtrmac @nalind @vrothberg PTAL

kurt-google:
Are these failing tests flakey? It appears they are timing out and locally they seem to run in a second or two.

rhatdan:
Probably flakes, I restarted.

giuseppe (approved):
LGTM

vrothberg (commented):
LGTM
@nalind for a final head nod?

image.go
mtrmac: I don’t think so — we need the chain of writers open until `io.Copy` at line 440 actually writes the data.

mtrmac, line 412: `archive.CompressStream(…, archive.Uncompressed)` actually wraps the input in a buffered writer. Does that matter for performance either way?

( The original code in https://github.com/moby/moby/pull/7542 doesn’t really talk about this case; it might well not be necessary to use a buffer. )

mtrmac (commented):
LGTM as far as correctness goes, one question about performance.

image.go
kurt-google: I left it out to avoid an additional writer in the chain but at least on my machine I dont think it makes much difference with or without it. If we prefer we can keep it in the chain.

kurt-google: Reading the linked PR its probably better to preserve it as it was added for network and other paths that I did not test, and it doesnt slow down the fast disk case I was testing.

mtrmac: I think the discussed network path is not involved here - we read from c/storage and write to a local file (vs. streaming read from storage -> compress -> push over HTTP to registry). So I don’t think that old conversation is all that informative.

Anyway, sure, the updated version of this PR makes this question moot.

mtrmac (commented):
LGTM. Thanks again!

TomSweeneyRedHat:
Tests aren't happy and you need a rebase @kurt-google

kurt-google:
Is it possible for me to  trigger test retries? It seems test 156 is consitently flakey. Otherwise I ask maintainers please kick the tests that fail until they work.

rhatdan:
/lgtm
Thanks @kurt-google

PR #3492, opened 2021-09-01
manifest: Make sure manifest `rm` only removes manifests not referenced images.

flouthoc:
Following PR makes sure that `buildah manifest rm <list>` only removes
the manifest list not referenced images.

Depends on: https://github.com/containers/common/pull/753

### DIFF
--- a/tests/rm.bats
+++ b/tests/rm.bats
@@ -52,3 +52,20 @@ load helpers
   run_buildah 125 rm -a "$cid"
   expect_output --substring "when using the --all switch, you may not pass any containers names or IDs"
 }
+
+@test "remove a single tagged manifest list" {
+  _prefetch busybox
+  run_buildah manifest create manifestsample
+  run_buildah manifest add manifestsample busybox
+  run_buildah tag manifestsample manifestsample2
+  run_buildah manifest rm manifestsample2
+  # Output should only untag the listed manifest nothing else
+  expect_output "untagged: localhost/manifestsample2:latest"
+  run_buildah manifest rm manifestsample
+  # Since actual list is getting removed it will also print the image id of list
+  # So check for substring instead of exact match
+  expect_output --substring "untagged: localhost/manifestsample:latest"
+  # Check if busybox is still there
+  run_buildah images
+  expect_output --substring "busybox"
+}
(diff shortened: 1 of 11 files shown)


flouthoc:
I need to change `flouthoc/common` to `c/common` as soon as containers/common#753 is merged.

giuseppe:
could you split the common upgrade to a different commit?

flouthoc:
Sure.

tests/rm.bats
```diff
@@ -52,3 +52,13 @@ load helpers
...
 }
+
+@test "remove a single tagged manifest list" {
+  _prefetch busybox
+  run_buildah manifest create manifestsample
+  run_buildah manifest add manifestsample busybox
+  run_buildah tag manifestsample manifestsample2
+  run_buildah manifest rm manifestsample2
+  # Output should only untag the listed manifest nothing else
+  expect_output "untagged: localhost/manifestsample2:latest"
```
vrothberg, line 63: And now remove `manifestsample` and make sure that its ID is returned.

vrothberg (commented):
One addition to the test, other than that, LGTM

Nice work!

tests/rm.bats
flouthoc: @vrothberg added change in latest commit.

flouthoc:
@vrothberg @giuseppe PTAL

vrothberg:
@nalind @ashley-cui PTAL

tests/rm.bats
```diff
@@ -52,3 +52,17 @@ load helpers
...
+  run_buildah manifest rm manifestsample
+  # Since actual list is getting removed it will also print the image id of list
+  # So check for substring instead of exact match
+  expect_output --substring "untagged: localhost/manifestsample:latest"
```
nalind, line 67: Unless this is covered by another test, verifying that the image we added to the list (busybox) is still present would at least make me feel better.

flouthoc: @nalind That's a very fair request. Similar check was added in podman for this [fix  ](https://github.com/containers/podman/pull/11406/files#diff-5358dbc9f2e2e54914b369deca4e89d4e3d498e2076dafa5ef49c2079da6f2b1R304)adding it here as well. :)

flouthoc:
@nalind @ashley-cui @giuseppe PTAL

Issue #3070, opened 2021-03-08
multi-stage build fails with unmarshalling error when using file descriptor as authfile


kkujala:
**Description**

When I try to build a multi-stage Dockerfile with a file descriptor from bash process substitution, then it fails with unmarshalling error on the second image fetching. The first image is fetched successfully and the second image fetching encounters the error.

In order to hide the authentication credentials from process listings or from filesystem, then creating a temporary file descriptor with bash process substitution is one way of achieving it securely. I.e. using `--creds` option would reveal the secrets to the host system when running `ps`. And using an actual file with `--authfile` would reveal the secrets likewise in the filesystem.

**Steps to reproduce the issue:**
1. Build the following dockerfile with buildah
```bash
$ cat Dockerfile 
FROM docker.io/centos:centos8 as build
FROM myrepository.example/image:tag
```
The following command should be used for building
```bash
buildah bud --authfile=<(echo "{ \"auths\": { \"myrepository.example\": { \"auth\": \"$(echo 'username:password' | base64 --wrap=0)\" } } }") --file Dockerfile --tag myimage .
```

2. Observe the first image centos8 being fetched correctly

3. Observe the second image `myrepository.example/image:tag` fetching leading to unmarshalling error.

**Describe the results you received:**
```bash
$ buildah bud --authfile=<(echo "{ \"auths\": { \"myrepository.example\": { \"auth\": \"$(echo 'foo:bar' | base64 --wrap=0)\" } } }") --file Dockerfile --tag myimage .
STEP 1: FROM docker.io/centos:centos8 AS build
Getting image source signatures
Copying blob 7a0437f04f83 [--------------------------------------] 0.0b / 0.0b
Copying config 300e315adb done  
Writing manifest to image destination
Storing signatures
--> 300e315adb2
STEP 2: FROM myrepository.example/image:tag
error creating build container: Error initializing source docker://myrepository.example/image:tag: error getting username and password: error reading JSON file "/dev/fd/63": error unmarshaling JSON at "/dev/fd/63": unexpected end of JSON input
ERRO exit status 125                              
```

It should be noted that the `myrepository.example` is intentionally invalid as per RFC 2606, as it is enough to demonstrate this issue. Also the authfile content is not valid but enough for demonstration. One can replace the both of the images with valid images from a repository that requires authentication, and use the correct secrets in the bash substitution. In such case the first image fetching works and the second one still fails with unmarshalling error. 

**Describe the results you expected:**
[...]

PR #3498, opened 2021-09-05
build: mirror --authfile to filesystem if pointing to FD instead of file

flouthoc:
Following PR makes sure that buildah mirrors --authfile to a temporary
file in filesystem if arg is pointing to an FD instead of actual file
as FD can be only consumed once.

Fixes: https://github.com/containers/buildah/issues/3070

### DIFF
--- a/cmd/buildah/build.go
+++ b/cmd/buildah/build.go
@@ -100,6 +100,7 @@ func getContainerfiles(files []string) []string {
 
 func buildCmd(c *cobra.Command, inputArgs []string, iopts buildOptions) error {
 	output := ""
+	cleanTmpFile := false
 	tags := []string{}
 	if c.Flag("tag").Changed {
 		tags = iopts.Tag
@@ -110,6 +111,10 @@ func buildCmd(c *cobra.Command, inputArgs []string, iopts buildOptions) error {
 	}
 	if err := auth.CheckAuthFile(iopts.BudResults.Authfile); err != nil {
 		return err
+	}
+	iopts.BudResults.Authfile, cleanTmpFile = buildahutil.MirrorToTempFileIfPathIsDescriptor(iopts.BudResults.Authfile)
+	if cleanTmpFile {
+		defer os.Remove(iopts.BudResults.Authfile)
 	}
 
 	pullPolicy := define.PullIfMissing
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -1,11 +1,43 @@
 package util
 
 import (
+	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/pkg/errors"
 )
+
+// Mirrors path to a tmpfile if path points to a
+// file descriptor instead of actual file on filesystem
+// reason: operations with file descriptors are can lead
+// to edge cases where content on FD is not in a consumable
+// state after first consumption.
+// returns path as string and bool to confirm if temp file
+// was created and needs to be cleaned up.
+func MirrorToTempFileIfPathIsDescriptor(file string) (string, bool) {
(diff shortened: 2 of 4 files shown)


flouthoc:
Wait for: https://github.com/containers/common/pull/756

flouthoc:
Merging c/common changes to buildah after discussions. This PR is ready for review.

flouthoc:
@vrothberg @Luap99 @rhatdan PTAL

pkg/util/util.go
```diff
@@ -1,12 +1,50 @@
...
+		return file
+	}
+
+	return tmpfile.Name()
+}
+
+// Removes any temp buildah file if it was created in the first place
+func RemoveTempFileByPath(path string) {
+	// check if path has prefix tmpDir+builah-temp-file
+	if strings.HasPrefix(path, filepath.Join(os.TempDir(), "buildah-temp-file")) {
```
vrothberg, line 43: I don't like that logic (as expressed in the c/common PR). I'd find it cleaner if `MirrorToTempFileIfPathIsDescriptor` had a 2nd return value indicating whether the authfile should be removed after usage or not.

cmd/buildah/build.go
```diff
@@ -111,6 +111,7 @@ func buildCmd(c *cobra.Command, inputArgs []string, iopts buildOptions) error {
 	if err := auth.CheckAuthFile(iopts.BudResults.Authfile); err != nil {
 		return err
 	}
+	iopts.BudResults.Authfile = buildahutil.MirrorToTempFileIfPathIsDescriptor(iopts.BudResults.Authfile)
```
vrothberg, line 114: Please do a `defer remove(authfiile)` here to make sure that created tmpfiles get cleaned up in all return paths (including error paths).

flouthoc: @vrothberg instead of defer do you think this makes sense 

```go
iopts.BudResults.Authfile, shouldBeCleaned = buildahutil.MirrorToTempFileIfPathIsDescriptor(iopts.BudResults.Authfile)

if shouldBeCleaned {
            buildahutil.RemoveTempFileByPath(iopts.BudResults.Authfile)
   }
   ```

Comment (flouthoc) [reply cmd/buildah/build.go:114]: I think defer makes sense nvm.

Comment (vrothberg) [reply cmd/buildah/build.go:114]: I think it should look as follows:
```go
iopts.BudResults.Authfile, shouldBeCleaned = buildahutil.MirrorToTempFileIfPathIsDescriptor(iopts.BudResults.Authfile)
if shouldBeCleaned {
    defer os.Remove(iopts.BudResults.Authfile)
  }
```

* Use `os.Remove()` instead of wrapping that into a custom function.
* We need defer to remove the file in all return paths.

pkg/util/util.go
flouthoc: fixed in recent commit.

vrothberg: To explicit: Please remove this function and replace the caller with `os.Remove(path)`. The caller now knows whether to remove the file or not, so don't need the redundant prefix check. Less magic, less code :)

flouthoc: @vrothberg ah my bad fixed it.

vrothberg (commented):
Code LGTM, thank you!

Since it's fixing an issue, please make sure to add `Fixes: #...` to the commit message, so GitHub will close it _and_ we'll preserve some history in the commit.

I also want to encourage to explain why no tests are needed in the commit message whenever `[NO ... NEEDED]` is set. We need to make sure to add regression tests if possible.

flouthoc:
@vrothberg thanks :) added a test case waiting for it pass first.

PR #3499, opened 2021-09-06
Fix build with .git url with branch

leighmcculloch:
#### What type of PR is this?

/kind bug

Fix builds with a URL that has a non-`git://` prefix, a `.git` suffix on its path, and contains a branch in the URL fragment (the portion of the URL after the `#`).

The build with git URL recognizes git URLs by either a `git://` scheme prefix, or any other scheme such as `https://` and a `.git` suffix.

The build with git URL and branch recognizes the branch name as the fragment part of the URL, the part after the `#`.

The `.git` suffix check is performed by checking that the suffix of the full URL is `.git`, however this causes URLs that begin with scheme `https://`, and have a `.git` path suffix, to not be recognized as git URLs when a branch name is appended.

The logic for checking if the URL path is suffixed with `.git` should parse the URL first and consider only the path when looking for the suffix.

#### How to verify it

Run buildah (or podman using this version of buildah) to build the following git URL:

```
podman build -t test https://github.com/leighmcculloch/stellar--docker-stellar-core-horizon.git#cap21and40
```

Fixes containers/podman#11450

This is my first contribution.

This commit also removes the guard within the `cloneToDirectory` function because it is unnecessary since every location the function is called the check has already been carried out.

```release-note
Fixed build from git URL's with branch. 
```

### DIFF
--- a/define/types.go
+++ b/define/types.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	urlpkg "net/url"
 	"os"
 	"os/exec"
 	"path"
@@ -111,7 +112,11 @@ func TempDirForURL(dir, prefix, url string) (name string, subdir string, err err
 	if err != nil {
 		return "", "", errors.Wrapf(err, "error creating temporary directory for %q", url)
 	}
-	if strings.HasPrefix(url, "git://") || strings.HasSuffix(url, ".git") {
+	urlParsed, err := urlpkg.Parse(url)
+	if err != nil {
+		return "", "", errors.Wrapf(err, "error parsing url %q", url)
+	}
+	if strings.HasPrefix(url, "git://") || strings.HasSuffix(urlParsed.Path, ".git") {
 		err = cloneToDirectory(url, name)
 		if err != nil {
 			if err2 := os.RemoveAll(name); err2 != nil {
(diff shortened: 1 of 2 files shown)


vrothberg (commented):
Thanks!

Other than the mini nit, LGTM.

leighmcculloch:
>Other than the mini nit, LGTM.

@vrothberg Thanks for reviewing so quickly. I don't see a comment from you that would be the "mini nit," what was it?

@rhatdan Thanks for approving so quickly.

How do I determine when this will be in a release of buildah (and the downstream podman)? I've got a development environment script migrating from docker to podman that is dependent on this fix, and I'm just planning when I can complete the migration.

vrothberg:
>  I don't see a comment from you that would be the "mini nit," what was it?

That is odd. Maybe I didn't click correctly. The comment related to using `urlParsed.Scheme == "git"` instead of `strings.HasPrefix` but it's nothing critical.

Regarding release: I guess the fix will make it into the upcoming Podman 3.4 release. @mheon may be able to give an ~ ETA.

leighmcculloch:
@vrothberg Ah yes, I debated using the parsed URL for other checks too, but then that also seemed a little out of scope of fixing the bug so I wasn't if that would be appreciated.  Thanks for the feedback!

PR #3508, opened 2021-09-09
Makefile: update cross targets

nalind:
#### What type of PR is this?

/kind other

Update the "cross" makefile target to just try building on every Linux, Darwin, or Windows architecture that the current version of Go supports.

That way, the next time we upgrade Go to a version that supports more architectures, we won't have to remember to update the list of cross-compile targets that we try to build in CI to catch cross-compilation problems.

#### How to verify it

`make cross`, then look in ./bin

None

```release-note

```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -177,7 +177,7 @@ conformance_task:
     conformance_test_script: '${SCRIPT_BASE}/test.sh conformance |& ${_TIMESTAMP}'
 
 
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

saschagrunert (approved):
LGTM

vrothberg (commented):
Clever solution!
/lgtm

Issue #3370, opened 2021-05-14
rebuild labels with build --label


dosmanak:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

If the image has been built with dynamic --labels, the next build with other dynamic labels does not neforce rebuild of labels layer

**Steps to reproduce the issue:**

```
$ podman -v; set -x; podman image rm podman_issue; podman build -t podman_issue . --label "dynamic_label1=value1"; echo ""; podman inspect podman_issue --format "LABEL {{ .Config.Labels.dynamic_label1 }};{{ .Labels.dynamic_label1 }}"; podman build -t podman_issue . --label "dynamic_label1=value2"; echo ""; podman inspect podman_issue --format "LABEL {{ .Config.Labels.dynamic_label1 }};{{ .Labels.dynamic_label1 }}"; podman image rm podman_issue; podman build -t podman_issue . --label "dynamic_label1=value2"; echo ""; podman inspect podman_issue --format "LABEL {{ .Config.Labels.dynamic_label1 }};{{ .Labels.dynamic_label1 }}"; set +x
```
```
podman version 3.1.2
+ podman image rm podman_issue
Untagged: localhost/podman_issue:latest
Deleted: 4391583b1b31d53ed3f64c86dfb29cbdb7635856431034e30e3bd99d111c7cca
+ podman build -t podman_issue . --label dynamic_label1=value1
STEP 1: FROM docker.io/adoptopenjdk/openjdk11:alpine-jre
STEP 2: CMD ls -la .
STEP 3: COMMIT podman_issue
--> 504c4f91411
504c4f914111e239a0465855d9ef418d692e3a1e327d8d24e1278995ced17259
+ echo ''

+ podman inspect podman_issue --format 'LABEL {{ .Config.Labels.dynamic_label1 }};{{ .Labels.dynamic_label1 }}'
LABEL value1;value1
+ podman build -t podman_issue . --label dynamic_label1=value2
STEP 1: FROM docker.io/adoptopenjdk/openjdk11:alpine-jre
STEP 2: CMD ls -la .
--> Using cache 504c4f914111e239a0465855d9ef418d692e3a1e327d8d24e1278995ced17259
STEP 3: COMMIT podman_issue
--> 504c4f91411
504c4f914111e239a0465855d9ef418d692e3a1e327d8d24e1278995ced17259
+ echo ''

[...]

PR #3517, opened 2021-09-14
build: Add additional step nodes when labels are modified

flouthoc:
Make sure label modification behaviour stays similar to docker i.e
new step is added when labels are modified and leave caching of layers
to executor

Advantages.

* Layers will be automatically cached if needed.
* If Labels are modified , new step will be built and cached if
  configured instead of forcing pre-built cached layers.
  
See detailed discussion here: https://github.com/containers/buildah/issues/3370#issuecomment-918024546

Closes: https://github.com/containers/buildah/issues/3370

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -318,6 +318,35 @@ func buildDockerfilesOnce(ctx context.Context, store storage.Store, logger *logr
 		}
 		mainNode.Children = append(mainNode.Children, additionalNode.Children...)
 	}
+
+	// Check if any modifications done to labels
+	// add them to node-layer so it becomes regular
+	// layer.
+	// Reason: Docker adds label modification as
+	// last step which can be processed as regular
+	// steps and if no modification is done to layers
+	// its easier to re-use cached layers.
+	if len(options.Labels) > 0 {
+		for _, labelSpec := range options.Labels {
+			label := strings.SplitN(labelSpec, "=", 2)
+			labelLine := ""
+			key := label[0]
+			value := ""
+			if len(label) > 1 {
+				value = label[1]
+			}
+			// check from only empty key since docker supports empty value
+			if key != "" {
+				labelLine = fmt.Sprintf("LABEL %q=%q\n", key, value)
+				additionalNode, err := imagebuilder.ParseDockerfile(strings.NewReader(labelLine))
+				if err != nil {
+					return "", nil, errors.Wrapf(err, "error while adding additional LABEL steps")
+				}
+				mainNode.Children = append(mainNode.Children, additionalNode.Children...)
+			}
+		}
+	}
+
 	exec, err := newExecutor(logger, logPrefix, store, options, mainNode)
 	if err != nil {
 		return "", nil, errors.Wrapf(err, "error creating build executor")
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -3348,3 +3348,81 @@ _EOF
   run_buildah 1 bud --signature-policy ${TESTSDIR}/policy.json --platform=linux/arm,linux/amd64 --manifest $outputlist -f ${TESTSDIR}/bud/multiarch/Dockerfile.fail ${TESTSDIR}/bud/multiarch
   run_buildah 125 manifest inspect $outputlist
 }
+
+
+# * Performs multi-stage build with label1=value1 and verifies
+# * Relabels build with label1=value2 and verifies
+# * Rebuild with label1=value1 and makes sure everything is used from cache
+@test "bud-multistage-relabel" {
+  _prefetch alpine busybox
+  run_buildah inspect --format "{{.FromImageDigest}}" busybox
+  fromDigest="$output"
+
+  target=relabel
+  run_buildah build --layers --label "label1=value1" --signature-policy ${TESTSDIR}/policy.json -t ${target} -f ${TESTSDIR}/bud/multi-stage-builds/Dockerfile.reused ${TESTSDIR}/bud/multi-stage-builds
+
+  # Store base digest of first image
+  run_buildah inspect --format '{{index .ImageAnnotations "org.opencontainers.image.base.digest" }}' ${target}
+  firstDigest="$output"
+
+  # Store image id of first build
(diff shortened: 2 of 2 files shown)


flouthoc:
Will add tests once we finalize on this discussion.

flouthoc:

nalind:
Looks reasonable.  Tests should definitely make sure this doesn't break weirdly on multistage builds and when the build replaces the base image (i.e., when the `--from` CLI option is used).  Please take care around quoting label values.

imagebuildah/build.go
```diff
@@ -318,6 +318,37 @@ func buildDockerfilesOnce(ctx context.Context, store storage.Store, logger *logr
...
+			labelLine := ""
+			key := label[0]
+			value := ""
+			if len(label) > 1 {
+				value = label[1]
+			} else {
+				value = ""
+			}
+			labelLine = fmt.Sprintf("LABEL %s=%s\n", key, value)
+			if len(labelLine) > 1 {
```
nalind, line 341: When would this be `false`?

imagebuildah/build.go
```diff
@@ -318,6 +318,37 @@ func buildDockerfilesOnce(ctx context.Context, store storage.Store, logger *logr
...
+				labelLineBytes := []byte(labelLine)
+				additionalNode, err := imagebuilder.ParseDockerfile(bytes.NewReader(labelLineBytes))
```
nalind, line 343: Woudl `strings.NewReader()` work better than casting to a byte slice and using `bytes.NewReader()` here?

imagebuildah/build.go
```diff
@@ -318,6 +318,37 @@ func buildDockerfilesOnce(ctx context.Context, store storage.Store, logger *logr
...
+		for _, labelSpec := range options.Labels {
+			label := strings.SplitN(labelSpec, "=", 2)
+			labelLine := ""
+			key := label[0]
+			value := ""
+			if len(label) > 1 {
+				value = label[1]
+			} else {
+				value = ""
+			}
```
TomSweeneyRedHat, line 339: I don't think the else is necessary as you're setting the default at line 334

imagebuildah/build.go
TomSweeneyRedHat: yeah, might be better to check if key and value don't equal ""

rhatdan:
@flouthoc are you planning on updating this PR?

flouthoc:
@rhatdan Just writing some test cases which covers all edge-cases.

imagebuildah/build.go
flouthoc: I think this is needed for basic validation since `--label` accepts arbitrary string similar check is also present in executor itself.

flouthoc:
@nalind  added tests for `regular` and `multi-staged` builds

**Muti-stage build test covers**
* Perform multi-stage build with `label1=value1` and verify labels.
* Relabel build with `label1=value2` and verify
* To make sure we are using cache rebuild with `label1=value1`  and makes sure everything is used from cache

**Regular build with --from**
* Covers basic modifying of labels and base images

Just waiting for everything to pass.

imagebuildah/build.go
```diff
@@ -318,6 +318,34 @@ func buildDockerfilesOnce(ctx context.Context, store storage.Store, logger *logr
...
+	if len(options.Labels) > 0 {
+		for _, labelSpec := range options.Labels {
+			label := strings.SplitN(labelSpec, "=", 2)
+			labelLine := ""
+			key := label[0]
+			value := ""
+			if len(label) > 1 {
+				value = label[1]
+			}
+			labelLine = fmt.Sprintf("LABEL %s=%s\n", key, value)
```
nalind, line 338: Really do need to look at quoting here.

flouthoc: @nalind ah yes, adding this 
```diff
-                       labelLine = fmt.Sprintf("LABEL %s=%s\n", key, value)
+                       labelLine = fmt.Sprintf("LABEL \"%s\"=\"%s\"\n", key, value)

File: imagebuildah/build.go [hunk 5]
```diff
@@ -318,6 +318,34 @@ func buildDockerfilesOnce(ctx context.Context, store storage.Store, logger *logr
...
+	if len(options.Labels) > 0 {
+		for _, labelSpec := range options.Labels {
+			label := strings.SplitN(labelSpec, "=", 2)
+			labelLine := ""
+			key := label[0]
+			value := ""
+			if len(label) > 1 {
+				value = label[1]
+			}
+			labelLine = fmt.Sprintf("LABEL \"%s\"=\"%s\"\n", key, value)
```
nalind, line 338: This doesn't look like it'll catch weirdness like values that include internal quotation marks of one kind or another.

flouthoc: @nalind i tried this `buildah build --layers --label "label1=value'2" --from=alpine -t image1`  but this is failing  `buildah build --layers --label 'label1=value"2' --from=alpine -t image1` let me get a fix.

flouthoc: @nalind I have changed this formatting with simple `%q` it autoescapes things for us tested both the cases, `"` will be autoescaped. 
```diff
-                       labelLine = fmt.Sprintf("LABEL \"%s\"=\"%s\"\n", key, value)
+                       labelLine = fmt.Sprintf("LABEL %q=%q\n", key, value)
```

rhatdan:
/approve
LGTM

imagebuildah/build.go
TomSweeneyRedHat: Right, but isn't labelLine at a minimum going to contain "LABEL = "?  That would never be 0 or less right?

flouthoc:
Just a rebase since main is updated.

imagebuildah/build.go
flouthoc: @TomSweeneyRedHat oh right thanks changing this to `key != ""`, we can support empty value since docker supports empty values (manually verified).

TomSweeneyRedHat:
LGTM
I'm not sure if this needs a rebase or not....

PR #3518, opened 2021-09-14
Makefile: only use race detection when it's available

nalind:
#### What type of PR is this?

/kind other

Check if `go test` supports the `-race` flag on the build platform, and if so, and only use it for unit tests if it's supported.

Fix a race in `run_linux.go`.

#### How to verify it

`make test-unit` should work correctly on platforms where the `-race` flag is not supported for builds.  `go help build` will list which platforms your installed version of Go supports the flag for.

None

Marking this as not needing new tests, though it may end up requiring additional changes if the tests we already have start noticing things that need to be fixed.

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ BINDIR := $(PREFIX)/bin
 BASHINSTALLDIR = $(PREFIX)/share/bash-completion/completions
 BUILDFLAGS := -tags "$(BUILDTAGS)"
 BUILDAH := buildah
+RACEFLAGS := $(shell go test -race ./pkg/rusage > /dev/null 2>&1 && echo -race)
 
 GO := go
 GO110 := 1.10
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

flouthoc (commented):
Stdout spits logs for downloading `go.mods` maybe redirect stdout to `/dev/null` and error to `/stdout`

Edit: its line no 13 of Makefile.

nalind:
Ah, right, the order matters.  Swapped them.

TomSweeneyRedHat:
LGTM

nalind:
Taking the `-race` flag off of the main `buildah` binary.  The performance hit of building it with race detection enabled is, as promised, rather noticeable.

nalind:
And likewise, `-race` makes our conformance tests take quite a bit longer, so dropping that for now rather than attempting to increase the timeout.

rhatdan:
What about the copy command?

nalind:
I'm currently hoping it doesn't slow the integration tests down enough to cause CI jobs to time out, but we'll see.

nalind:
Switched to using `sync/atomic` instead of a mutex for fixing the `stopped` race.

TomSweeneyRedHat:
LGTM

PR #3525, opened 2021-09-19
build: Cleanup transient mount destinations with every RUN step

flouthoc:
Following PR ensures that we cleanup dangling `/run` and dir after every RUN
command and remount with every `RUN` inorder to make sure that it does not 
persists on physical image. Ensureparity with how docker behaves with `.dockerenv`.

Closes: https://github.com/containers/buildah/issues/3523

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -545,6 +545,11 @@ func (b *Builder) setupMounts(mountPoint string, spec *specs.Spec, bundlePath st
 		return nil, err
 	}
 
+	// prepare list of mount destinations which can be cleaned up safely.
+	// we can clean bindFiles, subscriptionMounts and specMounts
+	// everything other than these might have users content
+	mountArtifacts.RunMountTargets = append(append(append(mountArtifacts.RunMountTargets, cleanableDestinationListFromMounts(bindFileMounts)...), cleanableDestinationListFromMounts(subscriptionMounts)...), cleanableDestinationListFromMounts(specMounts)...)
+
 	allMounts := util.SortMounts(append(append(append(append(append(append(volumes, builtins...), runMounts...), subscriptionMounts...), bindFileMounts...), specMounts...), sysfsMount...))
 	// Add them all, in the preferred order, except where they conflict with something that was previously added.
 	for _, mount := range allMounts {
@@ -559,6 +564,23 @@ func (b *Builder) setupMounts(mountPoint string, spec *specs.Spec, bundlePath st
 	// Set the list in the spec.
 	spec.Mounts = mounts
 	return mountArtifacts, nil
+}
+
+// Destinations which can be cleaned up after every RUN
+func cleanableDestinationListFromMounts(mounts []spec.Mount) []string {
+	mountDest := []string{}
+	for _, mount := range mounts {
+		// Add all destination to mountArtifacts so that they can be cleaned up later
+		if mount.Destination != "" {
+			// we dont want to remove destinations with  /etc, /dev, /sys, /proc as rootfs already contains these files
+			// and unionfs will create a `whiteout` i.e `.wh` files on removal of overlapping files from these directories.
+			// everything other than these will be cleanedup
+			if !strings.HasPrefix(mount.Destination, "/etc") && !strings.HasPrefix(mount.Destination, "/dev") && !strings.HasPrefix(mount.Destination, "/sys") && !strings.HasPrefix(mount.Destination, "/proc") {
+				mountDest = append(mountDest, mount.Destination)
+			}
+		}
+	}
+	return mountDest
 }
 
 // addResolvConf copies files from host and sets them up to bind mount into container
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -3446,3 +3446,55 @@ _EOF
 
   run_buildah rmi -f ${target}
 }
+
+@test "bud with run should not leave mounts behind cleanup test" {
+  skip_if_in_container
+  run which podman
+  if [[ $status -ne 0 ]]; then
+    skip "podman is not installed"
+  fi
+
+  # Create target dir where we will export tar
+  target=cleanable
+  mkdir ${TESTDIR}/${target}
+
+  # Build and export container to tar
+  run_buildah build --no-cache --signature-policy ${TESTSDIR}/policy.json -t ${target} -f ${TESTSDIR}/bud/containerfile/Containerfile.in ${TESTSDIR}/bud/containerfile
+  podman export $(podman create --name ${target} ${target}) --output=${TESTDIR}/${target}.tar
+
+  # We are done exporting so remove images and containers which are not needed
+  podman rm -f ${target}
+  run_buildah rmi ${target}
+
+  # Explode tar
+  tar -xf ${TESTDIR}/${target}.tar -C ${TESTDIR}/${target}
+  count=$(ls -A ${TESTDIR}/${target}/run | wc -l)
+  ## exported /run should be empty
+  assert "$count" == "0"
+}
+
+@test "bud with custom files in /run/ should persist cleanup test" {
+  skip_if_in_container
+  run which podman
+  if [[ $status -ne 0 ]]; then
+    skip "podman is not installed"
+  fi
+
+  # Create target dir where we will export tar
+  target=cleanable
+  mkdir ${TESTDIR}/${target}
+
+  # Build and export container to tar
+  run_buildah build --no-cache --signature-policy ${TESTSDIR}/policy.json -t ${target} -f ${TESTSDIR}/bud/add-run-dir
+  podman export $(podman create --name ${target} ${target}) --output=${TESTDIR}/${target}.tar
+
+  # We are done exporting so remove images and containers which are not needed
+  podman rm -f ${target}
+  run_buildah rmi ${target}
+
+  # Explode tar
+  tar -xf ${TESTDIR}/${target}.tar -C ${TESTDIR}/${target}
+  count=$(ls -A ${TESTDIR}/${target}/run | wc -l)
+  ## exported /run should not be empty
+  assert "$count" == "1"
+}
--- a/tests/bud/add-run-dir/Dockerfile
+++ b/tests/bud/add-run-dir/Dockerfile
@@ -1,0 +1,2 @@
+FROM alpine
+RUN touch /run/hello


rhatdan:
I think there are other  RUN artifacts then just /run.  We add some inodes in /etc as well.

flouthoc:
@rhatdan I think `/etc` is not a problem for the issue which @cgwalters created since other directory are already part of expected file structure instead of `/run` but sure we can clean up those as well if its needed. @cgwalters could you please confirm ?

flouthoc:
cc @nalind

nalind:
This gets us closer to cleaning up everything we leave behind during RUN, but it doesn't get us all the way there.

rhatdan:
@TomSweeneyRedHat I believe each RUN is independent and will recreate the /run/secrets.  This patch is just removing them when the container completes, so that a `buildah commit` will not keep them.

flouthoc:
@TomSweeneyRedHat so its exactly as @rhatdan stated every `RUN` is independent and things get recreated on individual run. We are just removing them so nothing persists in image.

flouthoc:
@nalind @rhatdan Added `/etc` to the list as well now apart from run `/etc/hosts`,`/etc/resolv.conf` and other transient files in `/etc` also gets removed. Do you guys think there is anything else

rhatdan:
My concern with this, without playing with it, is what happens if I want to add /run/foobar or /etc/resolv.conf to my image, can we differentiate between this.

from fedora
run touch /run/foobar
add hosts /etc/hosts

What happens in this case?

flouthoc:
@rhatdan I have removed `/etc/` from cleanup lists since i am not able to see `/etc` as dangling file in any of the image. Apart from that i tried building and i can see `/run/foobar` in image as it is not part of `bindFiles` list so cleanup will not remove any files manually added by user
```
FROM registry.fedoraproject.org/fedora:34
RUN touch /run/foobar
```
@nalind @rhatdan what are some other dangling stuff which we leave behind since i can only see `/run/secrets` and `/run/.containerenv` i dont think `/etc` is dangling at all i can only see that in rootfs

rhatdan:
How about:
cat Containerfile
FROM registry.fedoraproject.org/fedora:34
RUN rm /etc/hosts
RUN echo hi
buildah build -t test1 .
Does test1 have an /etc/hosts file within it?

flouthoc:
> How about:
> cat Containerfile
> FROM registry.fedoraproject.org/fedora:34
> RUN rm /etc/hosts
> RUN echo hi
> buildah build -t test1 .
> Does test1 have an /etc/hosts file within it?

@rhatdan this build fails for me with
```
STEP 2/3: RUN rm /etc/hosts
rm: cannot remove '/etc/hosts': Device or resource busy
error building at STEP "RUN rm /etc/hosts": error while running runtime: exit status 1
ERRO[0005] exit status 1    
```

So `/etc/hosts` is there but i can't see that on exploded layers commit-ed so i think it does not persists on layers. I cant see it when I try to inspect exploded layers manually.

flouthoc:
ps: discussed this with @nalind so we also have to take care of all the other mounts which are not getting cleanedup eg bind mounts
So I'll re-skim the code and add more components to this.

edsantiago:
I may be misunderstanding the intention behind this PR. I ran this command, and expected null output (or at least only directories, no files). Instead:
```console
$ printf "FROM quay.io/podman/stable:latest\nRUN podman pull quay.io/libpod/testimage:20210610\nRUN ls -lR /run\n" | ./bin/buildah bud -
STEP 1/3: FROM quay.io/podman/stable:latest
STEP 2/3: RUN podman pull quay.io/libpod/testimage:20210610
time="2021-09-21T20:22:17Z" level=warning msg="\"/\" is not a shared mount, this could cause issues or missing mounts with rootless containers"
Trying to pull quay.io/libpod/testimage:20210610...
Copying blob sha256:9afcdfe780b4ea44cc52d22e3f93ccf212388a90370773571ce034a62e14174e
Copying blob sha256:9afcdfe780b4ea44cc52d22e3f93ccf212388a90370773571ce034a62e14174e
Copying config sha256:9f9ec7f2fdef9168f74e9d057f307955db14d782cff22ded51d277d74798cb2f
Storing signatures
9f9ec7f2fdef9168f74e9d057f307955db14d782cff22ded51d277d74798cb2f
STEP 3/3: RUN ls -lR /run
/run:
total 0
drwxr-xr-x. 1 root root   0 Sep 21 08:06 console
drwx------. 1 root root  14 Sep 21 20:22 containers
drwxr-xr-x. 1 root root   0 Jan 26  2021 criu
drwx------. 1 root root   0 Sep 21 08:06 cryptsetup
drwxr-xr-x. 1 root root   0 Sep 21 08:06 faillock
drwxr-x--x. 1 root root  68 Sep 21 20:22 libpod
drwxr-xr-x. 1 root root  12 Sep 21 08:06 lock
drwxr-xr-x. 1 root root   0 Sep 21 08:06 log
-rw-r--r--. 1 root root   0 Sep 21 08:06 motd
....
/run/containers/storage/overlay-layers:
total 8
-rw-------. 1 root root  2 Sep 21 20:22 mountpoints.json
-rw-r--r--. 1 root root 64 Sep 21 20:22 mountpoints.lock
....
```
I tried with buildah @ main (without this patch) and the output looks identical.

rhatdan:
@edsantiago during the run the files and directories will be there.  The PR is attempting to make sure they do not get committed to the image.
So you would need to mount the final image with something like
mnt=$(podman image mount IMAGE)
ls -l $/mnt/run

edsantiago:
```console
$ printf "FROM quay.io/podman/stable:latest\nRUN podman pull quay.io/libpod/testimage:20210610\n" | ./bin/buildah bud -t foo -
.....
$ podman run --rm foo ls -laR /run|grep '^-r'
-rw-r--r--. 1 root root   0 Sep 21 20:45 .containerenv
-rw-r--r--. 1 root root   0 Sep 21 08:06 motd
-rw-------. 1 root root   2 Sep 21 20:40 mountpoints.json
-rw-r--r--. 1 root root  64 Sep 21 20:40 mountpoints.lock
-rw-r--r--. 1 root root  0 Sep 21 20:40 alive
-rw-r--r--. 1 root root  0 Sep 21 20:40 alive.lck
-rw-------. 1 root root  2 Sep 21 20:40 pause.pid
-rwx------. 1 root root 229 Sep 21 20:40 events.log
-rw-r--r--. 1 root root   0 Sep 21 20:40 events.log.lock
```

rhatdan:
Podman puts them back as well.  We are just looking at what goes into the image.  Not what a running container sees.
Another way to look at this would be to see what the image looks like when Docker runs it.  /run/.containerenv and /run/secrets should not be there.

flouthoc:
@nalind @rhatdan PTAL after struggling with conformance tests i think these following mounts can be cleaned up and changed to a more generic approach rather than cleaning individual directories.

run_linux.go
```diff
@@ -258,8 +258,8 @@ rootless=%d
 	}
 
 	defer func() {
-		if err := cleanupRunMounts(mountPoint, runArtifacts); err != nil {
-			options.Logger.Errorf("unabe to cleanup run mounts %v", err)
+		if err = cleanupRunMounts(mountPoint, runArtifacts); err != nil {
+			err = errors.Wrapf(err, "unable to cleanup run mounts")
```
rhatdan, line 262: Why did you change this?  The err is never going to be reported now?

rhatdan (approved):

rhatdan:
Don't we know the list of mounts that we added? Can't we just check for those and only remove those?
@nalind WDYT?

run_linux.go
flouthoc: Oh so we just setting already declared `err` from `Run(` its not a new variable. Problem with using options.Logger it becomes `nil` inside `func` and buildah goes into panic state I think it was never caught cause we never got any failures while removing mounts.

Please correct me if this looks wrong I can change this but i think we cannot use `options.Logger.Errorf("unabe to cleanup run mounts %v", err)`

flouthoc: we can add a `recover` check or use `fmt.Println(err.Error())` inside the defer instead of `options.Logger.Errorf("unabe to cleanup run mounts %v", err)` if that sounds good.

flouthoc: Just for ref: panic logs before this https://cirrus-ci.com/task/6666263280746496?logs=integration_test#L1223

flouthoc: @rhatdan reverting this change since it is not relevent to this PR will open different for fixing this in entire code i think we might have multiple instances of this.

rhatdan: Remember this message has to get to the remote client on Podman.  Which is why we are using
options.Logger

nalind:
> Don't we know the list of mounts that we added? Can't we just check for those and only remove those?

I believe that's what we're doing here, give or take certain exceptions.  The tricky part appears to be in only removing items that were added purely for the sake of RUN.

rhatdan:
Yes that is my concern.

Is 
run echo localhost 126.0.0.1 > /etc/resolv.conf
Safe?

nalind:
> Yes that is my concern.
> 
> Is
> run echo localhost 126.0.0.1 > /etc/resolv.conf
> Safe?

If that's a volume mounted file, then I wouldn't expect that content to show up in the image.

flouthoc:
@rhatdan that is safe

nalind:
I think this is fallout from #3518, which used `pkg/rusage` as its "quick" check to see if `go test` supported a `-race` flag.  It didn't use `-mod=vendor`, and even for a small package, it has some external dependencies that would get pulled down.

flouthoc:
@nalind @rhatdan let me know if anything else is needed on this PR.

rhatdan:
This PR makes it slightly better.

```
$ cat t.go 
package main
import ("fmt")
func main() {
	fmt.Println("Hello World")
}
$ go build -o t t.go
$ cat Containerfile
from scratch
copy t /t
run [ "/t" ]
$ ./bin/buildah bud -t dan .
STEP 1/3: FROM scratch
STEP 2/3: copy t /t
STEP 3/3: run [ "/t" ]
Hello World
COMMIT dan
Copying blob b31164ddec40 done  
Copying config 82eb390304 done  
Storing signatures
--> 82eb3903049
Successfully tagged localhost/dan:latest
82eb390304938f16dd707f32abaa8464af8d4a25959ab342e25696a540ec56b5
$ buildah unshare 
# mnt=$(podman image mount dan)
# find $mnt
/home/dwalsh/.local/share/containers/storage/overlay/b31164ddec40ddf1e0e42fa9848c17f0bf30fa4f8b9e4e0bcd4c4546f2579df0/merged
/home/dwalsh/.local/share/containers/storage/overlay/b31164ddec40ddf1e0e42fa9848c17f0bf30fa4f8b9e4e0bcd4c4546f2579df0/merged/dev
/home/dwalsh/.local/share/containers/storage/overlay/b31164ddec40ddf1e0e42fa9848c17f0bf30fa4f8b9e4e0bcd4c4546f2579df0/merged/etc
/home/dwalsh/.local/share/containers/storage/overlay/b31164ddec40ddf1e0e42fa9848c17f0bf30fa4f8b9e4e0bcd4c4546f2579df0/merged/etc/hosts
/home/dwalsh/.local/share/containers/storage/overlay/b31164ddec40ddf1e0e42fa9848c17f0bf30fa4f8b9e4e0bcd4c4546f2579df0/merged/etc/resolv.conf
/home/dwalsh/.local/share/containers/storage/overlay/b31164ddec40ddf1e0e42fa9848c17f0bf30fa4f8b9e4e0bcd4c4546f2579df0/merged/proc
/home/dwalsh/.local/share/containers/storage/overlay/b31164ddec40ddf1e0e42fa9848c17f0bf30fa4f8b9e4e0bcd4c4546f2579df0/merged/run
/home/dwalsh/.local/share/containers/storage/overlay/b31164ddec40ddf1e0e42fa9848c17f0bf30fa4f8b9e4e0bcd4c4546f2579df0/merged/sys
/home/dwalsh/.local/share/containers/storage/overlay/b31164ddec40ddf1e0e42fa9848c17f0bf30fa4f8b9e4e0bcd4c4546f2579df0/merged/t
```

flouthoc:
@rhatdan PR refrains from touching paths starting with `/etc, /dev, /sys, /proc` as unionfs has a `bug or a feature` ( dont know what to call it ) that whenever a file is removed and lower contains similar file a `.wh` or `whiteout` file is added as a replacement. We can cleanup `.wh` but don't know if we should remove perform cleanup on these paths. @nalind Could you confirm please.

nalind:
> @rhatdan PR refrains from touching paths starting with `/etc, /dev, /sys, /proc` as unionfs has a `bug or a feature` ( dont know what to call it ) that whenever a file is removed and lower contains similar file a `.wh` or `whiteout` file is added as a replacement. We can cleanup `.wh` but don't know if we should perform cleanup on these paths. @nalind Could you confirm please.

No, we wouldn't do that.  If L is the lower layer, and U is the upper layer in a mounted overlay filesystem M, the whiteout is used to mark for the kernel that the file was removed at U and thus shouldn't appear when looking at M.  L can be used as a lower layer for other mounts at the very same time, or it can be on read-only storage, but in either case the design of overlay is such that modifications made in the overlay mount are always, and only, recorded in the upper layer (U).

flouthoc:
@nalind thanks for the detailed explanation. 

@rhatdan @nalind then i guess these are the only mounts we can cleanup and i don't think people should face any issues with other paths.

rhatdan:
I am not sure we want to hit those other paths anyways, since readonly containers will rely on them
LGTM.

nalind:
This definitely needs a test.

cgwalters:
When testing, I've been doing e.g. `skopeo copy containers-storage:localhost/testimage oci:testimage` and then look in `testimage/blobs/sha256` at the tarballs.  I doubt there's a `podman` cli to get the raw stream, but OTOH it probably wouldn't be hard to add.

flouthoc:
Added tests

> @rhatdan PR refrains from touching paths starting with `/etc, /dev, /sys, /proc` as unionfs has a `bug or a feature` ( dont know what to call it ) that whenever a file is removed and lower contains similar file a `.wh` or `whiteout` file is added as a replacement. We can cleanup `.wh` but don't know if we should perform cleanup on these paths. @nalind Could you confirm please.

Just for ref: I also experimented with how docker is handling cleanups for transient files in these paths looks like docker is also not doing cleanup at all for these paths as well. Infact first thing they do is create `/etc/hosts` and `/etc/resolve.conf` no matter if they exists on lower or not and both of them is persisted in image as well.

rhatdan:
SGTM

flouthoc:
@nalind @rhatdan PTAL

tests/bud.bats
```diff
@@ -3446,3 +3446,63 @@ _EOF
...
+  run_buildah build --no-cache --signature-policy ${TESTSDIR}/policy.json -t ${target} -f ${TESTSDIR}/bud/containerfile/Containerfile.in ${TESTSDIR}/bud/containerfile
+  podman export $(podman create --name ${target} ${target}) --output=${target}.tar
+
+  # We are done exporting so remove images and containers which are not needed
+  podman rm -f ${target}
+  run_buildah rmi ${target}
+
+  # Explode tar
+  tar -xf ${target}.tar -C ${target}
+  count=$(ls ${target}/run | wc -l)
```
nalind, line 3472: This should be `ls -a` or something like it, since one of the things we're trying to make sure isn't being left behind is `/run/.containerenv`.

flouthoc: ah sure

tests/bud.bats
```diff
@@ -3446,3 +3446,63 @@ _EOF
...
+
+@test "bud with run should not leave mounts behind cleanup test" {
+  skip_if_in_container
+  run which podman
+  if [[ $status -ne 0 ]]; then
+    skip "podman is not installed"
+  fi
+
+  # Create target dir where we will export tar
+  target=cleanable
```
nalind, line 3458: Go ahead and this under $TESTDIR, which is created fresh for each test by the `setup()` function.

tests/bud.bats
flouthoc: @nalind using `ls -A` since we don't want `.` and `..`

nalind:
I'm always going to be uneasy about deleting data from a rootfs, but the tests pass, so LGTM.

PR #3527, opened 2021-09-20
unmarshalConvertedConfig(): handle zstd compression

nalind:
#### What type of PR is this?

/kind bug

The Docker manifest format doesn't currently support listing layers compressed using zstd, so we trigger an error when we try to convert an in-memory OCI manifest to the Docker format as a preliminary step in reading the image's config blob in the Docker format.
    
Instead, first create a temporary copy of the manifest, and then force the MIME types for all layers in the temporary copy of the manifest to appear to be compressed using gzip.  Both OCI and Docker formats will accept the resulting manifest without issue.  We throw the copy away after we've read the config blob, so the wackiness should be contained.

#### How to verify it

New integration test!

None

```release-note
None
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -11,6 +11,7 @@ import (
 	"github.com/containers/buildah/define"
 	"github.com/containers/buildah/docker"
 	"github.com/containers/image/v5/manifest"
+	"github.com/containers/image/v5/pkg/compression"
 	"github.com/containers/image/v5/transports"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage/pkg/stringid"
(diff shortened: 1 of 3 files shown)


nalind:
Related: https://github.com/containers/image/pull/1379

rhatdan:
@giuseppe @vrothberg @mtrmac PTAL

giuseppe (approved):
LGTM

giuseppe:
thanks, this solves the issue I've seen

vrothberg (approved):
/lgtm

mtrmac:
LGTM, for the record.

giuseppe:
backport to 1.23: https://github.com/containers/buildah/pull/3612

PR #3529, opened 2021-09-21
conformance: allow test cases to specify dockerUseBuildKit

nalind:
#### What type of PR is this?

/kind other

The API flag that we'd need to set to in a build request to get dockerd use BuildKit instead of the classic docker builder isn't available in go-dockerclient, so add a second docker-based path that uses the API types and client library, which the conformance tests were already pulling in as indirect dependencies, but skip tests that set the flag if we're not on the current client version.

This is mainly groundwork for testing compatibility, in the future, with features that aren't in the docker classic builder.

#### How to verify it

New conformance test!

None

```release-note
None
```

### DIFF
--- a/tests/conformance/conformance_test.go
+++ b/tests/conformance/conformance_test.go
@@ -32,6 +32,9 @@ import (
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/reexec"
+	dockertypes "github.com/docker/docker/api/types"
+	dockerdockerclient "github.com/docker/docker/client"
+	dockerarchive "github.com/docker/docker/pkg/archive"
 	docker "github.com/fsouza/go-dockerclient"
 	digest "github.com/opencontainers/go-digest"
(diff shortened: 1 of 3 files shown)


tests/conformance/conformance_test.go
```diff
@@ -32,6 +32,9 @@ import (
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/reexec"
+	dockertypes "github.com/docker/docker/api/types"
+	dockerdockerclient "github.com/docker/docker/client"
```
rhatdan, line 36: dockerdocker?

nalind: We already have imports named client, docker, and dockerclient, so I'm running out of pseudonyms for them.

nalind: Well, client's a variable, not a package.

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

PR #3532, opened 2021-09-22
push: support to specify the compression format

giuseppe:
add two new flags to "buildah push" to allow tweaking the compression
format for the data layers.

The flag --compression-format allows users to specify the compression
algorithm to use.

With --compression-level it is possible to tweak the compression
level.

An image usage for partial pulls can be pushed with:

$ buildah push --compression-format zstd:chunked FOO

#### What type of PR is this?

/kind feature

It allows users to specify the compression for the image data layers.

#### How to verify it

buildah push --compression-format zstd:chunked FOO dir:/tmp/foo

The files under /tmp/foo are compressed with zstd.

None

```release-note
push has two new flags to allow tweaking the compression format for the data layers.
The flag --compression-format allows users to specify the compression algorithm to use, while with --compression-level it is possible to tweak the compression level.
```

### DIFF
--- a/cmd/buildah/push.go
+++ b/cmd/buildah/push.go
@@ -13,6 +13,7 @@ import (
 	"github.com/containers/buildah/util"
 	"github.com/containers/common/pkg/auth"
 	"github.com/containers/image/v5/manifest"
+	"github.com/containers/image/v5/pkg/compression"
 	"github.com/containers/image/v5/transports"
 	"github.com/containers/image/v5/transports/alltransports"
 	"github.com/containers/storage"
@@ -31,6 +32,8 @@ type pushOptions struct {
 	digestfile         string
 	disableCompression bool
 	format             string
+	compressionFormat  string
+	compressionLevel   int
 	rm                 bool
 	quiet              bool
 	removeSignatures   bool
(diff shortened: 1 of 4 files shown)


nalind:
Does this work with every transport?

giuseppe:
I've tested it with `docker`, `oci` and `dir`.

cmd/buildah/push.go
```diff
@@ -41,6 +45,8 @@ type pushOptions struct {
 	encryptLayers      []int
 }
 
+var flags *flag.FlagSet
```
nalind, line 48: Why move this variable into the package scope?

giuseppe: I need to use it with `if flags.Changed("compression-level") {` later or to check whether the compressionLevel (which is a `*int`) was set.

nalind: Use `c.Flag("compression-level").Changed` instead?

giuseppe: thanks for the hint, that looks better.  I'll fix it

docs/buildah-push.md
```diff
@@ -38,6 +38,16 @@ environment variable. `export REGISTRY_AUTH_FILE=path`
 Use certificates at *path* (\*.crt, \*.cert, \*.key) to connect to the registry.
 The default certificates directory is _/etc/containers/certs.d_.
 
+**--compression-format** *format*
+
+Specifies the compression format to use.  Supported values are: `gzip`, `zstd` and `zstd:chunked`.
+
+**--compression-level** *level*
+
+Tweak the compression level used with the compression
```
TomSweeneyRedHat, line 47: Suggest
```suggestion
Specify the compression level used with the compression.
```
but please do at least add the period

giuseppe: fixed

nalind:
LGTM

vrothberg (commented):
/lgtm
/hold

Restarted flaked test (handshake timeout)

Issue #3536, opened 2021-09-23
Set TERM inside Working Containers (by default).


stellarpower:
Feature Request
---------------------------------------------------

**Description**

TERM is not set inside buildah working containers. For me, this leads to the inconvenience that I don't get colour from my prompt - sounds trivial, but for very long make builds of, say, c++, this provides an invaluable timesaver in scrolling up to find the one root error that caused the 589 errors that occurred below because of it. Also, using fish as my shell, a surprising amount of feedback is lost without colours, I have now discovered due to this.

**Steps to reproduce the issue:**
1. `set c (buildah from alpine)`
2. `buildah run $c -- ls -lah --color /`

**Describe the results you received:**
3. Output should be monochromatic.
4. `buildah run $c fish -c 'set --show TERM'` indicates that the variable is not set in any scope
5. I've also come across a few other funky errors before. We can't clear also, and sometimes curses-type stuff in the shell goes awry and I can't see what I'm doing properly.

**Describe the results you expected:**

Colours would be nice :)
This can be fixed simply by appending the environment variable to the run command. I'm guessing it's missed off the variables that are carried through automatically, whilst in `podman run`, it shows up on the inside by default.

**Output of `apt list buildah`:**

```terminal
Listando... Hecho
buildah/desconocido,now 100:1.22.3-1 amd64 [instalado]
buildah/desconocido 100:1.22.3-1 arm64
buildah/desconocido 100:1.22.3-1 armhf
buildah/desconocido 100:1.22.3-1 s390x

```

**Output of `buildah version`:**

```terminal
Version:         1.22.3
Go Version:      go1.16.6
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        0.4.0
libcni Version:  
image Version:   5.15.2
Git Commit:      
Built:           Thu Jan  1 01:00:00 1970
OS/Arch:         linux/amd64
```

**Output of `cat /etc/*release`:**

```
DISTRIB_ID=LinuxMint
DISTRIB_RELEASE=20.1
DISTRIB_CODENAME=ulyssa
DISTRIB_DESCRIPTION="Linux Mint 20.1 Ulyssa"
NAME="Linux Mint"
VERSION="20.1 (Ulyssa)"
PRETTY_NAME="Linux Mint 20.1"
...
VERSION_CODENAME=ulyssa
UBUNTU_CODENAME=focal

```

**Output of `uname -a`:**

```
[...]

PR #3539, opened 2021-09-24
Pass down the default config for buildah run command

rhatdan:
Currently we are not passing down the default environent as specified
in the containers.conf file.

Fixes: https://github.com/containers/buildah/issues/3536

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -100,14 +100,18 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 		return err
 	}
 
+	defaultContainerConfig, err := config.Default()
+	if err != nil {
+		return errors.Wrapf(err, "failed to get container config")
+	}
 	// hardwire the environment to match docker build to avoid subtle and hard-to-debug differences due to containers.conf
-	b.configureEnvironment(g, options, []string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"})
+	b.configureEnvironment(g, options, defaultContainerConfig.GetDefaultEnv())
 
 	if b.CommonBuildOpts == nil {
 		return errors.Errorf("Invalid format on container you must recreate the container")
 	}
 
-	if err := addCommonOptsToSpec(b.CommonBuildOpts, g); err != nil {
+	if err := addCommonOptsToSpec(b.CommonBuildOpts, g, defaultContainerConfig); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


rhatdan:
@flouthoc @nalind @vrothberg @TomSweeneyRedHat PTAL

rhatdan:
Not sure if we should remove these environment variables when container stops running?

rhatdan:
@stellarpower  PTAL

run_linux.go
```diff
@@ -100,14 +100,18 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 		return err
 	}
 
+	defaultContainerConfig, err := config.Default()
+	if err != nil {
+		return errors.Wrapf(err, "failed to get container config")
+	}
 	// hardwire the environment to match docker build to avoid subtle and hard-to-debug differences due to containers.conf
```
nalind, line 107: Uh.

TomSweeneyRedHat:
Needs a rebase and all kinds of test unhappiness @rhatdan

stellarpower:
> @stellarpower  PTAL

Please take a look? Was wondering if there's a glossary for all these acronyms, I quite like them! Thanks!

flouthoc:
@rhatdan I am not entirely  but I think this is voiding the comment added on line `107` of `run_linux.go` more context is added here https://github.com/containers/buildah/pull/2635

flouthoc:
I think default env config was intentionally removed since i am not exactly aware why but there is warning. Could we explicitly set just `TERM=xterm` more like

```diff
--- a/run_linux.go
+++ b/run_linux.go
@@ -101,7 +101,7 @@ func (b *Builder) Run(command []string, options RunOptions) error {
        }
 
        // hardwire the environment to match docker build to avoid subtle and hard-to-debug differences due to containers.conf
-       b.configureEnvironment(g, options, []string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"})
+       b.configureEnvironment(g, options, []string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin", "TERM=xterm"})
 
        if b.CommonBuildOpts == nil {
                return errors.Errorf("Invalid format on container you must recreate the container")
```

flouthoc:
@stellarpower I can confirm docker does not sets this `TERM` either. If we bake `TERM` in images i think we will loose parity between buildah and docker builds. I tried following with docker

```Dockerfile
FROM docker.io/alpine
RUN env
```
I think this issue might be a potential **`WONTFIX`**

stellarpower:
Ah okay, so that'll carry through to the image. I hadn't thought of it like that.

Is there a way environment variables could be set for the shell session but not added to the image? Am guessing they're treayed as if they were ENV statements in a container file then (?). Could they be added to the session (possibly then not doing this by default) without going into the image?  With the proviso that this could change the resulting image if the process inside behaves differently - we could assume it's safe to neglect this for things like TERM, but of course passing any old environment variable through will impact the build process on the inside.

Thanks for looking into it so quickly!

rhatdan:
We have secret environment variables, that do not get committed to the image.

rhatdan:
Looks like this is a bad idea, closing.

PR #3546, opened 2021-09-27
buildah build: add --all-platforms

nalind:
#### What type of PR is this?

/kind feature

Add a `--all-platforms` that instructs the builder to build for the intersection of all platforms for which the build's base images are available.  Returns an error if any of them aren't references to manifest lists.  We've learned that we can't really trust architecture and OS information stored in image config blobs, so we don't try to salvage that case.

#### How to verify it

New integration test!

None

The bits where we populate and prune the list of platforms that all of our base images support use versions both with and without variant information in order to avoid dropping 64-bit ARM when one base image is for variant "v8" and one doesn't specify a variant.  It's not ideal, but I think it's within the bounds of for how variant information is meant to be treated.

```release-note
Added the --all-platforms option to "buildah build".
```

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -10,15 +10,19 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 
+	"github.com/containerd/containerd/platforms"
 	"github.com/containers/buildah/define"
 	"github.com/containers/buildah/util"
 	"github.com/containers/common/libimage"
 	"github.com/containers/common/pkg/config"
+	"github.com/containers/image/v5/docker"
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/manifest"
+	"github.com/containers/image/v5/pkg/shortnames"
 	istorage "github.com/containers/image/v5/storage"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage"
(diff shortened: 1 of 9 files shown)


imagebuildah/build.go
```diff
@@ -431,3 +442,196 @@ func preprocessContainerfileContents(logger *logrus.Logger, containerfile string
...
+				logger.Debugf("image %q supports %q", baseImage, platform)
+			}
+		} else {
+			// prune the list of any platforms this base image doesn't support, with or without variant figured in
+			present := make(map[string]struct{})
+			for _, instance := range s2list.Manifests {
+				platform := instance.Platform.OS + "/" + instance.Platform.Architecture
+				present[platform] = struct{}{}
+				if instance.Platform.Variant != "" {
+					platform = platform + "/" + instance.Platform.Variant
```
rhatdan, line 538: You want to build multiple arches with and without the variant?

docs/buildah-build.md
```diff
@@ -38,6 +38,10 @@ Add a custom host-to-IP mapping (host:ip)
 
 Add a line to /etc/hosts. The format is hostname:ip. The **--add-host** option can be set multiple times.
 
+**--all-platforms**
+
+Instead of building for a set of platforms specified using the **--platform** option, inspect the build's base images, and build for all of the platforms for which they are all available.  Stages which use *scratch* as a starting point can not be inspected, so at least one non-*scratch* stage must be present for detection to work usefully.
```
TomSweeneyRedHat, line 43: Minor nit/suggestion
```suggestion
Instead of building for a set of platforms specified using the **--platform** option, inspect the build's base images, and build for all of the platforms for which they are all available.  Stages that use *scratch* as a starting point can not be inspected, so at least one non-*scratch* stage must be present for detection to work usefully.
```

nalind: Changed.

imagebuildah/build.go
nalind: Should be fixed now.

TomSweeneyRedHat:
LGTM
assuming happy tests

Issue #3217, opened 2021-05-09
RUN --mount=type=bind doesn't seem to work with relative source paths


akdor1154:
I'm trying to run a script inside my container during build, without COPYing it.
Docker w/BuildKit is OK with this, but Buildah doesn't seem to like it.

## Setup

```
Dockerfile
setup.sh
```

```Dockerfile
#Dockerfile
FROM debian:10

RUN mkdir buildcontext
RUN --mount=type=bind,source=.,target=/buildcontext /buildcontext/setup.sh
```

```sh
#setup.sh
echo hi!
```
## Test
```sh
# docker is ok
env DOCKER_BUILDKIT=1 docker build -t test --no-cache --progress plain .

```
<blockquote>
<details><summary>Output</summary>

```
#1 [internal] load build definition from Dockerfile
#1 sha256:bbfc041ced8e7ad3f93c7959439642058755819fda5a27b71c69d70ce9f3aa3c
#1 transferring dockerfile: 37B done
#1 DONE 0.0s

#2 [internal] load .dockerignore
#2 sha256:c11a926a0f68328dfd07b46271743e06e0585e997349009bac5436ba2605736c
#2 transferring context: 2B done
#2 DONE 0.0s

#3 [internal] load metadata for docker.io/library/debian:10
#3 sha256:ddb83d0bde9bcb14777289b1ac16d70d4d9ca3f82a764e7914f87e2e1fd702d1
#3 DONE 0.0s

#4 [1/3] FROM docker.io/library/debian:10
#4 sha256:80c01a5c5fd9f219c4fcc67169879cfd950ce672c8983a4b51eafa65a2907a36
#4 CACHED

#6 [internal] load build context
#6 sha256:a133a8149b38368be0bf659c22b42dcdba408161e18ada0b3f320f3dbd44e934
#6 transferring context: 59B done
#6 DONE 0.0s

#5 [2/3] RUN mkdir buildcontext
#5 sha256:cc1eba7768d6df39d1d00f3e1b66e337ca43fbe5dd52da4f41e927dad18b2ca1
#5 DONE 0.2s

#7 [3/3] RUN --mount=type=bind,source=.,target=/buildcontext /buildcontext/setup.sh
#7 sha256:99af5c3879fbaa722e780a1726e4b4b99a1b852994c0c6bb75d6e7ff066e85d0
#7 0.267 hi!
#7 DONE 0.3s

#8 exporting to image
[...]

PR #3548, opened 2021-09-28
build: Add support for buildkit like `--mount=type=bind`

flouthoc:
Following commit adds support for using buildkit like `--mount=type=bind` with `RUN` statements.
 Mounts created by `--mount` are transient in nature and only scoped to current RUN statements just like `buildkit`.
 
 Closes: https://github.com/containers/buildah/issues/3217


flouthoc:
@containers/buildah-maintainers @giuseppe @rhatdan @nalind @vrothberg PTAL

buildah.go
```diff
@@ -259,6 +261,8 @@ type BuilderOptions struct {
 	FromImage string
 	// Container is a desired name for the build container.
 	Container string
+	// ContextDirectory of current build
+	ContextDir string
```
nalind, line 265: Make this a field in `RunOptions`, so that it can differ for different `buildah run` invocations, like the `--contextdir` function that `buildah add` and `buildah copy` provide.

run_linux.go
```diff
@@ -2381,6 +2391,54 @@ func runSetupRunMounts(mounts []string, secrets map[string]string, sshSources ma
...
+		switch kv[0] {
+		case "source":
+			source = kv[1]
+		case "target", "dst", "destination":
+			target = kv[1]
+		default:
+			return nil, errInvalidSyntax
+		}
+	}
+	if source == "" || source == "." {
```
TomSweeneyRedHat, line 2413: is the '.' check necessary?  Wouldn't that resolve?  If it doesn't resolve, do you need a check for `./` too?

buildah.go
flouthoc: ah makes sense thanks will do changes.

run_linux.go
flouthoc: resolved

buildah.go
flouthoc: Added override so --contextDir from `run` could be used if specified
```go
func (b *Builder) Run(command []string, options RunOptions) error {
         // override contextdir from RunOptions if specified explicitly via --contextDir
	// otherwise use ContextDir configured from build
	if options.ContextDir != "" {
		b.ContextDir = options.ContextDir
	}

```

flouthoc:
@TomSweeneyRedHat I think its quay flakes

flouthoc:
resolved conflicts after recent commits in main

tests/bud.bats
```diff
@@ -3509,3 +3509,15 @@ _EOF
...
   assert "$count" == "1"
 }
+
+@test "bud-with-mount-like-buildkit" {
+  skip_if_no_runtime
+  skip_if_in_container
+  # Create file in current build context and mount inside RUN step
+  touch input_file
+  echo hello > input_file
+  run_buildah build -t testbud --signature-policy ${TESTSDIR}/policy.json -f ${TESTSDIR}/bud/buildkit-mount/Dockerfile
```
TomSweeneyRedHat, line 3519: This line is failing in the ci....

flouthoc:
@cevich @containers/buildah-maintainers I am not sure but looks like something is up with CI for Conformance test its failing for all the PR's with error `Package containerd.io is not installed`

run_linux.go
```diff
@@ -66,6 +67,12 @@ func setChildProcess() error {
 
 // Run runs the specified command in the container's root filesystem.
 func (b *Builder) Run(command []string, options RunOptions) error {
+	// override contextdir from RunOptions if specified explicitly via --contextDir
+	// otherwise use ContextDir configured from build
+	if options.ContextDir != "" {
+		b.ContextDir = options.ContextDir
```
nalind, line 73: This should _not_ quietly persist.  Go ahead and remove `Builder.ContextDir`.

run_linux.go
```diff
@@ -2403,6 +2421,20 @@ func (b *Builder) runSetupRunMounts(mounts []string, secrets map[string]string,
 	return finalMounts, artifacts, nil
 }
 
+func (b *Builder) getBindMount(tokens []string, rootUID, rootGID, processUID, processGID int) (*spec.Mount, error) {
+	var optionMounts []specs.Mount
+	mount, err := parse.GetBindMount(tokens, b.ContextDir)
```
nalind, line 2426: Use the value from `RunOptions.ContextDir` here, and return an error from here if one hasn't been set.

pkg/parse/parse.go
```diff
@@ -387,12 +387,24 @@ func GetBindMount(args []string) (specs.Mount, error) {
...
+		newMount.Options = append(newMount.Options, "rbind")
+	}
+
 	if !setDest {
 		return newMount, noDestError
 	}
 
-	if !setSource {
-		newMount.Source = newMount.Destination
+	// builkit parity: support absolute path for sources from current build context
```
nalind, line 400: Typo: builkit -> buildkit.

pkg/parse/parse.go
```diff
@@ -387,12 +387,24 @@ func GetBindMount(args []string) (specs.Mount, error) {
...
+	if newMount.Source == "" || newMount.Source == "." || newMount.Source == "./" {
+		newMount.Source = contextDir
+	}
+	// if path begins with `./` replace `./` with context dir
+	if strings.HasPrefix(newMount.Source, "./") {
+		// path should be /contextDir/specified path
+		newMount.Source = strings.Replace(newMount.Source, ".", contextDir, 1)
```
nalind, line 407: Sounds like `filepath.Join(contextDir, filepath.Clean(filepath.Separator+newMount.Source))` is going to do the right thing here.  We need to be careful that we don't accidentally end up binding in a location that's actually outside of the context directory.

imagebuildah/stage_executor.go
```diff
@@ -547,6 +547,7 @@ func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBCo
 		ConfigureNetwork:      s.executor.configureNetwork,
 		CNIPluginPath:         s.executor.cniPluginPath,
 		CNIConfigDir:          s.executor.cniConfigDir,
+		ContextDir:            s.executor.contextDir,
```
nalind, line 550: Set this in the `RunOptions` in Run() instead.

run_linux.go
flouthoc: @nalind One problem with using it directly from `runOptions` would be that  i'll have to pass down ContextDir as an argument across all the function but since all the functions are already part of `builder` interface it would be easily accessible from builder property like `b.ContextDir`

Just thought this would be cleaner.

nalind: The value should not be kept.

run_linux.go
```diff
@@ -2391,6 +2399,16 @@ func (b *Builder) runSetupRunMounts(mounts []string, secrets map[string]string,
...
 			}
+		case "bind":
+			mount, err := b.getBindMount(tokens, rootUID, rootGID, processUID, processGID)
+			if err != nil {
+				return nil, nil, err
+			}
+			if mount != nil {
+				finalMounts = append(finalMounts, *mount)
+				mountTargets = append(mountTargets, mount.Destination)
+
```
nalind, line 2410: Blank line at end of block, please remove it.

flouthoc:
> @cevich @containers/buildah-maintainers I am not sure but looks like something is up with CI for Conformance test its failing for all the PR's with error `Package containerd.io is not installed`

This is problem is still there with CI

cevich:
> This is problem is still there with CI

Dang, not sure how that slipped through my testing.  In any case a temporary fix can look like this: https://github.com/containers/podman/pull/11821/files#diff-f3866c83e5fdf3650df2ad5b9c0c18efaa71891d23675732cc931e90fd2b10faR261

[I already opened an issue on the missing doodad and will take care of it shortly.](https://github.com/containers/automation_images/issues/95)

pkg/parse/parse.go
```diff
@@ -333,27 +333,28 @@ func getMounts(mounts []string) (map[string]specs.Mount, error) {
...
+	bindNonRecursive := false
 
 	for _, val := range args {
 		kv := strings.SplitN(val, "=", 2)
 		switch kv[0] {
 		case "bind-nonrecursive":
 			newMount.Options = append(newMount.Options, "bind")
-		case "ro", "nosuid", "nodev", "noexec":
+			bindNonRecursive = true
+		case "ro", "nosuid", "nodev", "noexec", "rw", "readwrite":
```
nalind, line 350: Our `chroot` isolation implementation doesn't recognize either "rw" or "readwrite", and a peek at runc and crun sources suggests that they don't recognize "readwrite".

pkg/parse/parse.go
```diff
@@ -387,12 +387,24 @@ func GetBindMount(args []string) (specs.Mount, error) {
...
-	if !setSource {
-		newMount.Source = newMount.Destination
+	// buildkit parity: support absolute path for sources from current build context
+	if newMount.Source == "" || newMount.Source == "." || newMount.Source == "./" {
+		newMount.Source = contextDir
+	}
+	// if path begins with `./` replace `./` with context dir
+	if strings.HasPrefix(newMount.Source, "./") {
+		// path should be /contextDir/specified path
+		newMount.Source = filepath.Join(contextDir, filepath.Clean(string(filepath.Separator)+newMount.Source))
```
nalind, line 407: Is this first block still needed?

tests/bud.bats
```diff
@@ -3509,3 +3509,15 @@ _EOF
   ## exported /run should not be empty
   assert "$count" == "1"
 }
+
+@test "bud-with-mount-like-buildkit" {
+  skip_if_no_runtime
+  skip_if_in_container
+  # Create file in current build context and mount inside RUN step
+  touch input_file
```
nalind, line 3517: Why not put this file in ${TESTSDIR}/bud/buildkit-mount and specify that directory as the build context directory?

flouthoc:
@nalind thanks resolved.

pkg/parse/parse.go
flouthoc: @nalind I think crun supports `rw` https://github.com/containers/crun/blob/main/src/libcrun/linux.c#L378

docs/buildah-run.1.md
```diff
@@ -58,6 +58,12 @@ not disabled.
 List of directories in which the CNI plugins which will be used for configuring
 network namespaces can be found.
 
+**--contextdir** *directory*
+
+Allows setting context directory for current RUN invocation. Specifying a context
+directory causes Buildah to chroot into that context directory. This means copying
+files pointed at by symbolic links outside of the chroot will fail.
```
nalind, line 65: This still needs correcting - we don't chroot here, and the specified location is used as the root directory for source locations specified in `--mount` options that specify a mount type of "bind".

pkg/parse/parse.go
nalind: What's the intent here, specifically for source locations that aren't absolute paths?

run_linux.go
```diff
@@ -2391,6 +2393,15 @@ func (b *Builder) runSetupRunMounts(mounts []string, secrets map[string]string,
 				// Count is needed as the default destination of the ssh sock inside the container is  /run/buildkit/ssh_agent.{i}
 				sshCount++
 			}
+		case "bind":
+			mount, err := b.getBindMount(tokens, contextDir, rootUID, rootGID, processUID, processGID)
+			if err != nil {
+				return nil, nil, err
+			}
+			if mount != nil {
```
nalind, line 2401: Is this check necessary?  Are there cases where `err` is `nil` and `mount` is also `nil`?

cevich:
Opened https://github.com/containers/buildah/pull/3560 to fix CI

run_linux.go
flouthoc: @nalind since its a pointer  i think it would be ok to do a `nil` before performing any `dereferencing`. Better than any SIGSEGV in future. But i am cool dropping if its a bother.

tests/bud/buildkit-mount/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM alpine
+RUN mkdir /test
+# use option z if selinux is enabled
+RUN --mount=type=bind,source=.,target=/test,z cat /test/input_file
```
nalind, line 4: Please also have a RUN with no `source` speciified, and with the source being a subdirectory under the context directory, given both with and without a leading "/".

pkg/parse/parse.go
flouthoc: yes. builkit accepts paths starting with `.` and treats `.` as current context. However users are free to add any valid path from host machine.

nalind: I would expect relative paths to be interpreted as relative to a context directory if a context directory was specified, and to the current directory otherwise.

pkg/parse/parse.go
nalind: But we need those for the feature to be complete, don't we?

tests/bud/buildkit-mount/Dockerfile
flouthoc: sure will add few more tests.

run_linux.go
nalind: `getSecretMount()` and `getSSHMount()` can return a `nil` mount without also returning an error but `getBindMount()` doesn't.

pkg/parse/parse.go
flouthoc: @nalind `ContextDir` falls back to `os.Getwd()` if nothing special is configured so yes this should happen.

pkg/parse/parse.go
flouthoc: Could we just handle it here since `rw` == `readwrite`  instead of offloading to runtime.

flouthoc: I'll check the relevant changes which needs to be done for `chroot` isolation part.

flouthoc:
@nalind Added requested tests, I still need to implement `rw` for chroot isolation.

pkg/parse/parse.go
nalind: In `buildah build`, yes, but not in `buildah run` so far as I can see.

run.go
```diff
@@ -93,6 +93,8 @@ type RunOptions struct {
 	User string
 	// WorkingDir is an override for the working directory.
 	WorkingDir string
+	// ContextDirectory for the current RUN invocation.
```
nalind, line 96:
```suggestion
	// ContextDir is used as the root directory for the source location for mounts that are of type "bind".
```

pkg/parse/parse.go
flouthoc: @nalind Just curious is `rw` a valid use-case for isolated roots since behavior is expected to be immutable.

nalind: The [doc](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#run---mounttypebind-the-default-mount-type) says the bind mount is read-only by default, but writeable is an option.

flouthoc:
@nalind I think `rw` is already working properly for isolated `chroot` i have also written a test for `rw` on isolated chroot. With `rw` enabled user can write on `bind` files and changes are also reflected on host.

pkg/parse/parse.go
flouthoc: `rw` is working as expected for both regular builds and isolated roots. I have also written a test for the same.

flouthoc:
Rebasing after https://github.com/containers/buildah/pull/3562

rhatdan:
LGTM

pkg/parse/parse.go
nalind: This still needs to be something the tests cover.

nalind: In particular, I'd like tests to cover cases where a source is specified with a location that looks like "tests/testdata" (i.e., with no leading "./" or "../", and not specified as an absolute path) when the current working directory when `buildah` is invoked is not the same as the context directory.  In every case, we need to be sure that we can't end up mounting a source location that is outside of the context directory.

flouthoc: @nalind Yes i agree relative path should always be picked from `ContextDir` added a test could you PTAL

pkg/parse/parse.go
```diff
@@ -333,27 +333,30 @@ func getMounts(mounts []string) (map[string]specs.Mount, error) {
...
 			// TODO: detect duplication of these options.
 			// (Is this necessary?)
 			newMount.Options = append(newMount.Options, kv[0])
+		case "rw", "readwrite":
+			newMount.Options = append(newMount.Options, "rw")
 		case "readonly":
 			// Alias for "ro"
 			newMount.Options = append(newMount.Options, "ro")
-		case "shared", "rshared", "private", "rprivate", "slave", "rslave", "Z", "z":
+		case "shared", "rshared", "private", "rprivate", "slave", "rslave", "Z", "z", "U":
```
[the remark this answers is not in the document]
flouthoc: @nalind Mount code is being reused from volumes and tests for `U` is already added here https://github.com/containers/buildah/blob/main/tests/run.bats#L304 therefore I refrained from adding `ownership` test to prevent redundancy

PR #3560, opened 2021-10-01
Cirrus: Workaround defunct package metadata

cevich:
#### What type of PR is this?

> /kind flake

During VM image build, a number of packages are downloaded but not
installed, since they may interfere with some testing.  Then at runtime,
where required, the packages are installed from cache and used.  This
mechanism broke, and is being investigated in
https://github.com/containers/automation_images/issues/95

#### How to verify it

The setup stage for the conformance tests will pass without error.

Broken CI

None


cevich:
Ahh, found a fix...closing in favor of #3562

Issue #3494, opened 2021-09-01
Optimization: use smaller images in tests


edsantiago:
The `bud with copy-from in Dockerfile no prior FROM` test is causing flakes in podman testing because image pulls take over two minutes:

https://github.com/containers/buildah/blob/e1f747152caa112fbc476c5440fdcfe5bef15c01/tests/bud.bats#L1568-L1574

`php` and `composer` are huge: 400MB and 185MB respectively. Does this test explicitly require something that can only be provided by these images? If not, could we rewrite it (and other tests that use huge images) to use the standard alpine/busybox small images?

### 121 bud with copy-from in Dockerfile no prior FROM
* fedora-34 : bud podman fedora-34 root host
  * PR #11368
    * [08-31 08:25](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5885546078666752/html/bud-podman-fedora-34-root-host.log.html)
  * PR #11330
    * [08-26 10:34](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6238256804134912/html/bud-podman-fedora-34-root-host.log.html)
  * PR #11009
    * [08-30 17:39](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6554996012482560/html/bud-podman-fedora-34-root-host.log.html)
* fedora-34 : bud remote fedora-34 root host [remote]
  * PR #11368
    * [08-31 08:26](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5322596125245440/html/bud-remote-fedora-34-root-host.log.html)
  * PR #11330
    * [08-26 10:34](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/4830881920581632/html/bud-remote-fedora-34-root-host.log.html)

nalind:
I don't think it does.  The test looks like it was modeled after the original report in https://github.com/containers/buildah/issues/1175.  Copying any file which we know is present in a different image should be sufficient - the important thing is that the argument of the `--from=` is not a reference to a previous stage (either by name or number) of the multi-stage build, so we treat the argument value it as the name of an image.

edsantiago:
Thank you! I've self-assigned.

rhatdan:
@edsantiago still working on this.

PR #3566, opened 2021-10-04
tests: stop pulling php, composer

edsantiago:
php and composer images (400MB and 185MB) were being used
unnecessarily. This sometimes caused flakes due to network
hiccups. Use busybox and libpod/testimage instead. (And,
tighten up the copy-from test).

Also -- oops! -- fix a nonworking expect_output. The
presence of '|' in "blah blah ... <image|stage>"
was causing '=~' to match only 'stage>'. Escape the '|'
in all relevant places, and confirm there are no more.

Fixes: #3494

### DIFF
--- a/tests/bud/copy-globs/Dockerfile
+++ b/tests/bud/copy-globs/Dockerfile
@@ -1,2 +1,2 @@
-FROM php:7.2
-COPY --from=composer:latest /usr/bin/composer /usr/local/bin/composer
+FROM busybox
+COPY --from=alpine /we/never/get/here /nor/here/either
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM

tests/bud/copy-globs/Dockerfile
```diff
@@ -1,2 +1,2 @@
-FROM php:7.2
-COPY --from=composer:latest /usr/bin/composer /usr/local/bin/composer
+FROM busybox
+COPY --from=alpine /we/never/get/here /nor/here/either
```
nalind, line 2: If I'm reading "bud with .dockerignore #3" correctly, this was expected to succeed.

edsantiago: Well, it does, but only because `Dockerfile` is ignored: as best I can tell, the test seems to use `Containerfile` instead. I don't know why there's a `Dockerfile` in that directory, maybe as a test that `Containerfile` will override it?

nalind: Argh, I was looking at the `copy-from` directory.  Then I think this is right.

TomSweeneyRedHat:
From the aspect of image diversity in our tests, I'd rather keep PHP and composer, but since it's causing issues, I'll withdraw my objections.
LGTM

PR #3569, opened 2021-10-05
Bump containerd to v1.5.7

lsm5:
Fixes: GHSA-c2h3-6mxw-7mvq
Vulnerable versions: >= 1.5.0, < 1.5.7
Patched version: 1.5.7

`Impact`
A bug was found in containerd where container root directories and
some plugins had insufficiently restricted permissions, allowing
otherwise unprivileged Linux users to traverse directory contents
and execute programs. When containers included executable programs
with extended permission bits (such as setuid), unprivileged Linux
users could discover and execute those programs. When the UID of
an unprivileged Linux user on the host collided with the file
owner or group inside a container, the unprivileged Linux user on
the host could discover, read, and modify those files.

`Patches`
This vulnerability has been fixed in containerd 1.4.11 and
containerd 1.5.7. Users should update to these version when they
are released and may restart containers or update directory
permissions to mitigate the vulnerability.

`Workarounds`
Limit access to the host to trusted users. Update directory
permission on container bundles directories.

#### What type of PR is this?

> /kind bug

Fix for GHSA-c2h3-6mxw-7mvq

#### How to verify it

Check if containerd v1.5.7 is correctly vendored (I think)

Fixes GHSA-c2h3-6mxw-7mvq

None

```release-note
Fixed GHSA-c2h3-6mxw-7mvq
```

@nalind @vrothberg @TomSweeneyRedHat @rhatdan PTAL

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/containers/buildah
 go 1.13
 
 require (
-	github.com/containerd/containerd v1.5.5
+	github.com/containerd/containerd v1.5.7
 	github.com/containernetworking/cni v0.8.1
 	github.com/containers/common v0.46.0
 	github.com/containers/image/v5 v5.16.0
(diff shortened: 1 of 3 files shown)


lsm5:
Do we have/need a `/kind security`  ?

rhatdan:
BTW This Vulnerability looks like it did not effect Buildah.

lsm5:
@rhatdan ack, now that you mention it, I think I remember someone else saying something to that effect. But I was seeing that annoying github security warning, hence the PR.

PR #3572, opened 2021-10-06
bud --base-images

vrothberg:
Add a new flag `--base-images` to `buildah bud` which allows for
extracting the base images without performing any build.

Extracting the base images *before* a build is needed in `podman-remote
build` to properly resolve short names prior to performing the build.

Podman will be consume the feature via the Go code but exposing it on
the `bud` CLI allows for easier testing.

@nalind @mtrmac @rhatdan PTAL

I'd prefer to just have one search registry for remote podman but since there is the strong desire to have podman-remote behave like the local one, I don't see any other user-friendly option.  The idea for `remote build` is to first extract the base images, then prompt for each where needed (see https://github.com/containers/podman/pull/11867), and then perform the build.


nalind:
Why is this a CLI flag rather than a different subcommand?

vrothberg:
> Why is this a CLI flag rather than a different subcommand?

It shares large parts of `buildah bud` in the front end, so a flag seemed like a good candidate.  Would hiding the flag work for you?

imagebuildah/build.go
```diff
@@ -55,56 +55,55 @@ type Mount = specs.Mount
...
+	if err != nil {
+		return nil, err
 	}
 
-	logger := logrus.New()
-	if options.Err != nil {
-		logger.SetOutput(options.Err)
-	} else {
-		logger.SetOutput(os.Stderr)
+	baseImages, err := baseImages(containerfiles, dockerfiles, options.From, options.Args)
```
mtrmac, line 70: (Absolutely non-blocking: The `containerfiles, dockerfiles` argument pair reads to me as if there were a deep semantic distinction between “container” and “docker” in this context. The previous `paths` vs. `files` naming was a bit simpler in this respect, though I could find something to quibble about with that naming as well :) )

imagebuildah/build.go
```diff
@@ -137,16 +136,16 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
...
 
 			if err != nil {
-				return "", nil, err
+				return nil, err
 			}
 			dinfo, err = contents.Stat()
 			if err != nil {
 				contents.Close()
-				return "", nil, errors.Wrapf(err, "error reading info about %q", dfile)
+				return nil, errors.Wrapf(err, "error reading info about %q", dfile)
```
TomSweeneyRedHat, line 144: error stutter?

docs/buildah-build.1.md
```diff
@@ -61,6 +61,10 @@ If the authorization state is not found there, $HOME/.docker/config.json is chec
 Note: You can also override the default path of the authentication file by setting the REGISTRY\_AUTH\_FILE
 environment variable. `export REGISTRY_AUTH_FILE=path`
 
+**--base-images**
+
+Extract the base images from all Containerfiles but do not perform the actual build.
```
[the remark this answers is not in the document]
TomSweeneyRedHat: I'm in favor of no document for this , but if so, I'd do the sub that @mtrmac suggested.

PR #3575, opened 2021-10-07
buildkit: Add support for `--mount=type=tmpfs` to allow mounting volatile memory instead of persistent storage.

flouthoc:
Following PR adds supports for buildkit like `--mount=type=tmpfs` which
allows end users to mount a chunk of volatile memory instead of a persistent storage device within container.

Usage
```Dockerfile
RUN --mount=type=tmpfs,target=/sometarget
```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -2400,6 +2400,13 @@ func (b *Builder) runSetupRunMounts(mounts []string, secrets map[string]string, 
 			}
 			finalMounts = append(finalMounts, *mount)
 			mountTargets = append(mountTargets, mount.Destination)
+		case "tmpfs":
+			mount, err := b.getTmpfsMount(tokens, rootUID, rootGID, processUID, processGID)
+			if err != nil {
+				return nil, nil, err
+			}
+			finalMounts = append(finalMounts, *mount)
+			mountTargets = append(mountTargets, mount.Destination)
 		default:
 			return nil, nil, errors.Errorf("invalid mount type %q", kv[1])
 		}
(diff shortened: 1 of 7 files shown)


run_linux.go
```diff
@@ -2400,6 +2400,13 @@ func (b *Builder) runSetupRunMounts(mounts []string, secrets map[string]string,
 			}
 			finalMounts = append(finalMounts, *mount)
 			mountTargets = append(mountTargets, mount.Destination)
+		case "tmpfs":
+			mount, err := b.getTmpfsMount(tokens, rootUID, rootGID, processUID, processGID)
```
rhatdan, line 2404: Just curious does this default to tmpcopyup?

dilyanpalauzov:
https://github.com/containers/buildah/blob/main/docs/Containerfile.5.md is not updated about RUN --mount=type=tmpfs .

run_linux.go
flouthoc: Added `tmpcopyup` as a new option in separate commit.

rhatdan:
LGTM
@ashley-cui @vrothberg @TomSweeneyRedHat @nalind PTAL

vrothberg (commented):
Code LGTM but a test for `tmpcopyup` is missing.

vrothberg:
> @vrothberg thanks added test for `tmpcopyup` in latest commit.

LGTM, thank you.  Not worth repushing but: the tests are noew in the docs commit.

Issue #3318, opened 2021-06-18
Glob COPY is failing


sjenning:
**Description**

Glob COPY is failing on very basic case

**Steps to reproduce the issue:**
```
$ cat Dockerfile.prebuilt 
FROM quay.io/openshift/origin-base:4.8
COPY bin/* /usr/bin/
ENTRYPOINT /usr/bin/hypershift

$ ls bin
control-plane-operator  hosted-cluster-config-operator  hypershift  hypershift-operator  ignition-server

$ buildah bud -f Dockerfile.prebuilt
STEP 1: FROM quay.io/openshift/origin-base:4.8
STEP 2: COPY bin/* /usr/bin/
error building at STEP "COPY bin/* /usr/bin/": no items matching glob "/home/sjennings/projects/go/src/github.com/openshift/hypershift/bin/*" copied (5 filtered out): no such file or directory
ERRO[0001] exit status 125
```

**Describe the results you received:**

all files matching the glob are filtered

**Describe the results you expected:**

all files matching the glob should be copied without error

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.21.0-1.fc34.x86_64
```

**Output of `buildah version`:**

```
Version:         1.21.0
Go Version:      go1.16.3
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        0.4.0
libcni Version:  
image Version:   5.12.0
Git Commit:      
Built:           Wed Dec 31 18:00:00 1969
OS/Arch:         linux/amd64
```

**Output of `podman version` if reporting a `podman build` issue:**

```
Version:      3.1.2
API Version:  3.1.2
Go Version:   go1.16.3
Built:        Wed May 12 14:27:59 2021
OS/Arch:      linux/amd64
```

**Output of `cat /etc/*release`:**

```
Fedora release 34 (Thirty Four)
NAME=Fedora
VERSION="34 (Workstation Edition)"
ID=fedora
VERSION_ID=34
VERSION_CODENAME=""
PLATFORM_ID="platform:f34"
PRETTY_NAME="Fedora 34 (Workstation Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:34"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/34/system-administrators-guide/"
[...]

PR #3576, opened 2021-10-07
Report ignorefile location when no content added

rhatdan:
Users have accidently had a .containerignore or .dockerignore
file in their context directly which blocked all content.
Currently we report that no globs matched, but do not
identify where the globs came from.

This change is an attempt to add this data to the error.
Example: https://github.com/containers/buildah/issues/3318

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/add.go
+++ b/add.go
@@ -47,8 +47,10 @@ type AddAndCopyOptions struct {
 	// If the sources include directory trees, Hasher will be passed
 	// tar-format archives of the directory trees.
 	Hasher io.Writer
-	// Excludes is the contents of the .dockerignore file.
+	// Excludes is the contents of the .containerignore file.
 	Excludes []string
+	// IgnoreFile is the path to the .containerignore file.
+	IgnoreFile string
 	// ContextDir is the base directory for content being copied and
 	// Excludes patterns.
 	ContextDir string
@@ -564,7 +566,11 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 			}
 		}
 		if itemsCopied == 0 {
-			return errors.Wrapf(syscall.ENOENT, "no items matching glob %q copied (%d filtered out)", localSourceStat.Glob, len(localSourceStat.Globbed))
+			excludesFile := ""
+			if options.IgnoreFile != "" {
+				excludesFile = " using " + options.IgnoreFile
+			}
+			return errors.Wrapf(syscall.ENOENT, "no items matching glob %q copied (%d filtered out%s)", localSourceStat.Glob, len(localSourceStat.Globbed), excludesFile)
 		}
 	}
 	return nil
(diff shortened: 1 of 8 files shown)


add.go
```diff
@@ -564,7 +564,17 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
...
 		}
 		if itemsCopied == 0 {
-			return errors.Wrapf(syscall.ENOENT, "no items matching glob %q copied (%d filtered out)", localSourceStat.Glob, len(localSourceStat.Globbed))
+			ignoreFile := filepath.Join(contextDir, ".containerignore")
+			if _, err := os.Stat(ignoreFile); err != nil {
+				ignoreFile = filepath.Join(contextDir, ".dockerignore")
+				if _, err := os.Stat(ignoreFile); err != nil {
+					ignoreFile = ""
+				}
+			}
```
nalind, line 573: This guess could well be wrong if `buildah add`'s `--ignorefile` flag is used.

rhatdan: There is no way to know at this level whether that happened. I would figure most likely, their would not have been an ignorefile.

pkg/parse/parse.go
```diff
@@ -1098,3 +1100,47 @@ func SSH(sshSources []string) (map[string]*sshagent.Source, error) {
...
+func ignoreFile(contextDir string) string {
+	for _, i := range []string{".containerignore", ".dockerignore"} {
+		ignoreFile := filepath.Join(contextDir, i)
+		if _, err := os.Stat(ignoreFile); err == nil {
+			return ignoreFile
+		}
+	}
+	return ""
+}
+func ContainerIgnoreFile(path string, contextDir string) ([]string, string, error) {
```
nalind, line 1128: The arguments here are in the reverse of the order in which I think its callers expect.

pkg/parse/parse.go
```diff
@@ -1098,3 +1100,47 @@ func SSH(sshSources []string) (map[string]*sshagent.Source, error) {
...
+	for _, i := range []string{".containerignore", ".dockerignore"} {
+		ignoreFile := filepath.Join(contextDir, i)
+		if _, err := os.Stat(ignoreFile); err == nil {
+			return ignoreFile
+		}
+	}
+	return ""
+}
+func ContainerIgnoreFile(contextDir, path string) ([]string, string, error) {
+	if path != "" {
```
nalind, line 1129: Might as well use `imagebuilder.ParseIgnore()` instead of most of this block.

rhatdan:
@nalind @TomSweeneyRedHat @flouthoc @vrothberg @giuseppe @ashley-cui @umohnani8 PTAL

Issue #3555, opened 2021-09-29
Containerfile man page says FROM "must be first command"


TomSweeneyRedHat:
The man page for the Containerfile says the `FROM` instruction MUST be the first instruction in the file.  Generally, that is the case, but one or more `ARG` instructions can also be the first instruction in the Containerfile.  The `FROM` stanza in the Containerfile man page just needs to be adjusted for this.

derkoe:
ARG before FROM does not work with podman/buildah

nalind:
Heading `ARG`s (those defined before the first `FROM` instruction) should be usable in the `FROM` instructions for any of the build's stages.

derkoe:
Yes but only if the `ARG` is used in the `FROM`. When used after `FROM` the argument is undefined.

Here is the part from the Docker docs:
 > FROM may only be preceded by one or more ARG instructions, which declare arguments that are used in FROM lines in the Dockerfile.

TomSweeneyRedHat:
@derkoe I think we're in violent agreement.  The problem is the man page now says that FROM "must" be the first line.  That's true for most cases, but not all.  I just want to tweak it with a note similar to your comment.

rhatdan:
Lets add the line from Docker documentation.

PR #3577, opened 2021-10-08
[CI:DOCS] Touch up Containerfile man page to show ARG can be 1st

TomSweeneyRedHat:
The Containerfile man page says FROM must be the first statement
in the file.  However, that is not true as the ARG instruction can proceed it.

Fixes: #3555

#### What type of PR is this?

> /kind bug

Fixes the Containerfile man page which stated FROM had to be the first instruction, when it is permissible to have ARG.

#### How to verify it

Read

```release-note
None
```

### DIFF
--- a/docs/Containerfile.5.md
+++ b/docs/Containerfile.5.md
@@ -61,8 +61,10 @@ A Containerfile is similar to a Makefile.
   `FROM image@digest`
 
   -- The **FROM** instruction sets the base image for subsequent instructions. A
-  valid Containerfile must have **FROM** as its first instruction. The image can be any
-  valid image. It is easy to start by pulling an image from the public
+  valid Containerfile must have either **ARG** or *FROM** as its first instruction.
+  If **FROM** is not the first instruction in the file, it may only be preceded by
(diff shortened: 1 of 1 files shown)


docs/Containerfile.5.md
```diff
@@ -61,8 +61,10 @@ A Containerfile is similar to a Makefile.
   `FROM image@digest`
 
   -- The **FROM** instruction sets the base image for subsequent instructions. A
-  valid Containerfile must have **FROM** as its first instruction. The image can be any
-  valid image. It is easy to start by pulling an image from the public
+  valid Containerfile must have either **ARG** or *FROM** as its first instruction.
+  If **FROM** is not the first instruction in the file, it may only be preceded by
+  one or more ARG instructions, which declare arguments that are used in FROM lines in the Containerfile.
```
rhatdan, line 66: Just to nitpick, aren't ARGs only used by the next FROM command.  IE 
ARG FOO=bar
from fedora
from $FOO

Does not work. From reading this description, I would assume it does.

PR #3586, opened 2021-10-12
Add support for host.containers.internal in the /etc/hosts

rhatdan:
Also switch default hostname from truncated ContainerID to Container
name.

It makes more sense that a user would know the name of the container
versus the generated id, so we should use this as a default.

Fixes: https://github.com/containers/buildah/issues/3509

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```


tests/run.bats
```diff
@@ -528,6 +528,11 @@ function configure_and_check_user() {
 	cid=$output
 	run_buildah 125 run --isolation=chroot --network=bogus $cid cat /etc/hosts
 	expect_output "checking network namespace: stat bogus: no such file or directory"
+
+	run_buildah run --isolation=chroot $cid cat /etc/hosts
```
nalind, line 532: Why is this forcing isolation to chroot, even in cases where CI would be trying to use the default runtime?

rhatdan: Just using existing tests.

run_linux.go
```diff
@@ -716,11 +715,29 @@ func (b *Builder) generateHosts(rdir, hostname string, addHosts []string, chownO
...
 		hosts.Write([]byte(fmt.Sprintf("%s\t%s\n", values[1], values[0])))
 	}
-
 	if hostname != "" {
 		hosts.Write([]byte(fmt.Sprintf("127.0.0.1   %s\n", hostname)))
 		hosts.Write([]byte(fmt.Sprintf("::1         %s\n", hostname)))
 	}
+
+	// getLocalIP returns the non loopback local IP of the host
+	getLocalIP := func() string {
```
rhatdan, line 724: @nalind @Luap99 do we have any better way then this to get the hosts ip?

nalind: A host can validly have multiple network addresses, so this can return at most one address, or even no address.  What does `docker build` provide?

rhatdan: Buildkit does:
```
#5 [2/3] RUN cat /etc/hosts
#5 0.346 127.0.0.1      localhost buildkitsandbox
#5 0.346 ::1    localhost ip6-localhost ip6-loopback
```

Luap99: For docker you have to use `--add-host=host.docker.internal:host-gateway` AFAIK and they use the default docker0 bridge ip address.
I think this solution is the best for buildah. For podman we use the CNI or slirp4netns gateway but this is also difficult to get.

run_linux.go
```diff
@@ -716,11 +715,29 @@ func (b *Builder) generateHosts(rdir, hostname string, addHosts []string, chownO
...
+			// check the address type and if it is not a loopback the display it
+			if ipnet, ok := address.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
+				if ipnet.IP.To4() != nil {
+					return ipnet.IP.String()
+				}
+			}
+		}
+		return ""
+	}
+	hosts.Write([]byte(fmt.Sprintf("%s         %s\n", getLocalIP(), "host.containers.internal")))
```
Luap99, line 739: this should only be done when `getLocalIP()` is not empty

flouthoc:
Just one doubt since we are removing check for `if hostname != ""` should we still have a line to set `localhost` if `hostname` is actually empty. 

Since I am not sure why this check was there in the first place.

rhatdan:
The check was there in case their was no entry. You would have had "::1 "
Now we are at least guaranteed to have container name.

run_linux.go
```diff
@@ -716,11 +716,29 @@ func (b *Builder) generateHosts(rdir, hostname string, addHosts []string, chownO
...
-		hosts.Write([]byte(fmt.Sprintf("127.0.0.1   %s\n", hostname)))
-		hosts.Write([]byte(fmt.Sprintf("::1         %s\n", hostname)))
+	// getLocalIP returns the non loopback local IP of the host
+	getLocalIP := func() string {
+		addrs, err := net.InterfaceAddrs()
+		if err != nil {
+			return ""
+		}
+		for _, address := range addrs {
+			// check the address type and if it is not a loopback the display it
```
TomSweeneyRedHat, line 729: nit
```suggestion
			// check the address type and if it is not a loopback then return it
```

PR #3574, opened 2021-10-07
buildkit: add support for persistent caching across builds with `--mount=type=cache`

flouthoc:
Following PR introduces a new mount `type=cache` in parity to buildkit
which allows users to share persistent cache between different builds.

Allowing users to cache content generated by business logic or enhance
build performance by caching components across various builds.

Closes: https://github.com/containers/buildah/issues/3452

PR #3590, opened 2021-10-18
buildkit: add `from=` field to `bind` and `cache` mounts so images can be used as source

flouthoc:
Following commit adds buildkit like support for `from` field to `--mount=type=bind`
and `--mount=type=cache` so images and stage can be used as mount source.

Usage looks like
```dockerfile
RUN --mount=type=bind,source=.,from=<your-image>,target=/path ls /path
```
and
```dockerfile
RUN --mount=type=cache,from=<your-image>,target=/path ls /path
```
Closes: https://github.com/containers/buildah/issues/3501


docs/buildah-run.1.md
```diff
@@ -127,6 +127,8 @@ Current supported mount TYPES are bind, cache, secret and tmpfs. <sup>[[1]](#Foo
 
               . bind-nonrecursive: do not setup a recursive bind mount.  By default it is recursive.
 
+              . from: Build stage or image name for the root of the source. Defaults to the build context.
```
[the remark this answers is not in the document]
nalind: Also some places where "." (Unicode 0x0002E) is used instead of "·" (Unicode 0x000B7), which made me think my eyes were playing tricks on me.

pkg/parse/parse.go
```diff
@@ -467,61 +488,93 @@ func GetCacheMount(args []string) (specs.Mount, error) {
...
 			newMount.Destination = kv[1]
 			setDest = true
+		case "src", "source":
+			if len(kv) == 1 {
+				return newMount, "", errors.Wrapf(optionArgError, kv[0])
+			}
+			if err := parse.ValidateVolumeHostDir(kv[1]); err != nil {
+				return newMount, "", err
+			}
+			newMount.Source = kv[1]
```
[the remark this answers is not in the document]
flouthoc: @nalind In case of `type=cache` field `src` is only applicable when `from` is specified. So in the same function below we change the `contextdir` to the mount created by `from` and if `from` is not specified `src` is just ignored.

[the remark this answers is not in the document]
flouthoc: @nalind in the same function we override `src` if `from` is not specified here so it happens before `validateVolumeHostDir()` https://github.com/containers/buildah/pull/3590/files#diff-0c73d3e5eb67eaabc39caa4cbb5dd69eb43ae6d98b1ca8889fc86d56bacefb90R563

docs/buildah-run.1.md
```diff
@@ -115,7 +115,7 @@ Current supported mount TYPES are bind, cache, secret and tmpfs. <sup>[[1]](#Foo
 
        Common Options:
 
-              · src, source: mount source spec for bind and volume. Mandatory for bind.
+              · src, source: mount source spec for bind and volume. Mandatory for bind. If `from` specified `src` is subpath in the from to mount.
```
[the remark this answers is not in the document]
TomSweeneyRedHat: never mind, too many reviews

run.go
```diff
@@ -153,6 +153,8 @@ type RunOptions struct {
 type runMountArtifacts struct {
 	// RunMountTargets are the run mount targets inside the container
 	RunMountTargets []string
+	// Any exernal images which were mounted inside container
```
TomSweeneyRedHat, line 156:
```suggestion
	// Any external images which were mounted inside the container
```

flouthoc:
Update: Splitting `se-linux` change into a different PR and will take label from the first container and let c/storage handle se-linux. [Affected feature]: **mutistage-from-mounting with selinux enabled**.

PS: I'll also rebase this PR

docs/Containerfile.5.md
```diff
@@ -114,7 +114,7 @@ Current supported mount TYPES are bind, cache, secret and tmpfs.
 
        Common Options:
 
-              · src, source: mount source spec for bind and volume. Mandatory for bind.
+              · src, source: mount source spec for bind and volume. Mandatory for bind. If `from` is specified, `src` is the subpath in the `from` field.
```
dilyanpalauzov, line 118: My reading is that tmpfs and secret also accept `from=`.

flouthoc: Hi @dilyanpalauzov. I think `secret` only mounts `key/values` while `from` is applicable for mounts where we need `stage` or `image` so might not be applicable for secret. For `tmpfs` it might be but i need to check why buildkit does not supports this for tmpfs.

Thanks for raising this though.

flouthoc:
> Update: Splitting `se-linux` change into a different PR and will take label from the first container and let c/storage handle se-linux. [Affected feature]: **mutistage-from-mounting with selinux enabled**.
> 
> PS: I'll also rebase this PR

Rebased after this. @nalind Could you PTAL

TomSweeneyRedHat:
@flouthoc looks like you need a rebase.  The changes LGTM, but honestly I just did a skim review and would like at least @nalind to review.

docs/Containerfile.5.md
```diff
@@ -126,6 +126,8 @@ Current supported mount TYPES are bind, cache, secret and tmpfs.
 
               . bind-nonrecursive: do not setup a recursive bind mount.  By default it is recursive.
 
+              . from: Image name for the root of the source. Defaults to the build context.
```
nalind, line 129: Typo: "." instead of "·"
"Image" shouldn't be capitalized, for consistency with the descriptions of "bind-propagation" and "bind-nonrecursive".

nalind:
This needs more tests.  For example, the `Dockerfile` test for cache mounts doesn't exercise the `source` or `id` flags, or cases where `from` points to an earlier stage in the build, and the test for bind mounts doesn't test what happens when `from` isn't specified.

flouthoc:
> This needs more tests. For example, the `Dockerfile` test for cache mounts doesn't exercise the `source` or `id` flags, or cases where `from` points to an earlier stage in the build, and the test for bind mounts doesn't test what happens when `from` isn't specified.

@nalind Thanks. Added new tests for

* Cache
   * using `from` previous stages. In a multistage build.
   * exercising `source` with relative paths across stages.
   * exercising `id` across stages. However practically `id` has no benefits  when using with `from`.
* Bind
   * pointing to invalid ( i.e `empty` ) `from`.
   * `from` specified without `source`. 
   
and some of tests are overlapping with https://github.com/containers/buildah/pull/3548/files and https://github.com/containers/buildah/pull/3574

Please let me know if any additional tests are needed.

flouthoc:
@nalind friendly ping. :)

pkg/parse/parse.go
```diff
@@ -429,8 +433,36 @@ func GetBindMount(args []string, contextDir string) (specs.Mount, error) {
 			// and can thus be safely ignored.
 			// See also the handling of the equivalent "delegated" and "cached" in ValidateVolumeOpts
 		default:
-			return newMount, errors.Wrapf(errBadMntOption, kv[0])
+			return newMount, "", errors.Wrapf(errBadMntOption, kv[0])
+		}
+	}
+
+	fmt.Printf("---from image %s\n", fromImage)
```
nalind, line 440: Debugging message left over?

tests/bud.bats
```diff
@@ -3669,3 +3669,99 @@ _EOF
...
+  run_buildah build -t buildkitbaserelative --signature-policy ${TESTSDIR}/policy.json -f ${TESTSDIR}/bud/buildkit-mount-from/Dockerfilebuildkitbaserelative ${TESTSDIR}/bud/buildkit-mount-from/
+  # try reading something from persistant cache in a different build
+  run_buildah build -t testbud --signature-policy ${TESTSDIR}/policy.json -f ${TESTSDIR}/bud/buildkit-mount-from/Dockerfilebindfromrelative
+  expect_output --substring "hello"
+  run_buildah rmi -f buildkitbaserelative
+  run_buildah rmi -f testbud
+}
+
+@test "bud-with-mount-bind-from-multistage-relative-like-buildkit" {
+
```
nalind, line 3733: Drop the extra blank line.

tests/bud.bats
```diff
@@ -3669,3 +3669,99 @@ _EOF
...
+  skip_if_no_runtime
+  skip_if_in_container
+  # build base image which we will use as our `from`
+  run_buildah build -t testbud --signature-policy ${TESTSDIR}/policy.json -f ${TESTSDIR}/bud/buildkit-mount-from/Dockerfilemultistagefrom ${TESTSDIR}/bud/buildkit-mount-from/
+  expect_output --substring "hello"
+  run_buildah rmi -f testbud
+}
+
+@test "bud-with-mount-bind-from-cache-multistage-relative-like-buildkit" {
+
```
nalind, line 3751: Drop the extra blank line.

nalind (commented):
I could have sworn there were integration tests for `buildah run` in here, too, since the man page already documents its ability to handle `--mount type=bind`, and those are needed.  We also need tests that verify that `buildah build` isn't confused when a `RUN` instruction is given multiple `--mount=type=cache` arguments.

rhatdan:
@flouthoc What is happening with this PR?

flouthoc:
@rhatdan I somehow didn't get github notification on this. I'll fix @nalind  remarks.

flouthoc:
@nalind Thanks I'll add requested tests for `buildah run` and `buildah build` and resolve nits.

util/util.go
```diff
@@ -175,6 +175,20 @@ func FindImage(store storage.Store, firstRegistry string, systemContext *types.S
 	return ref, localImage.StorageImage(), nil
 }
 
+// LookupImage returns *Image to corresponding imagename or id
+func LookupImage(store storage.Store, image string) (*libimage.Image, error) {
```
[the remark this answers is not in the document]
flouthoc: fixed in latest commit.

run_linux.go
```diff
@@ -2742,14 +2746,27 @@ func (b *Builder) getSSHMount(tokens []string, count int, sshsources map[string]
...
 		if err != nil {
 			return err
 		}
 	}
 
+	//cleanup any mounted images for this run
+	for _, image := range artifacts.MountedImages {
+		if image != "" {
+			i, err := util.LookupImage(b.store, image)
+			// its not necessary we are going to find image, could be stage as well.
```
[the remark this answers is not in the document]
flouthoc: We are re-using `stage` mountpoints they are automatically cleaned up as soon as build is done.

pkg/parse/parse.go
```diff
@@ -560,7 +634,7 @@ func GetCacheMount(args []string) (specs.Mount, error) {
 	}
 
 	// buildkit parity: cache must writable unless `ro` or `readonly` is configured explicitly
-	if !setReadOnly {
+	if !setReadOnly && fromImage == "" {
```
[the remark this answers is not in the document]
flouthoc: In this block we are configuring `rw` if its specified so option to `rw` can be only configured if `fromImage == ""` i.e If user is not using any `image or stage`. All the mounted `images` and `stages` would be `readonly` always.

pkg/parse/parse.go
```diff
@@ -440,32 +470,37 @@ func GetBindMount(args []string, contextDir string) (specs.Mount, error) {
 	}
 
 	if !setDest {
-		return newMount, noDestError
+		return newMount, fromImage, noDestError
 	}
 
 	// buildkit parity: support absolute path for sources from current build context
-	if strings.HasPrefix(newMount.Source, ".") || newMount.Source == "" || !filepath.IsAbs(newMount.Source) {
+	if strings.HasPrefix(newMount.Source, ".") || newMount.Source == "" || !filepath.IsAbs(newMount.Source) || fromImage != "" {
```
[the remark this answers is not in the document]
flouthoc: This is for `bind` mount. `bind` mount could support absolute `src` for cases when `from` is not specified.

pkg/parse/parse.go
```diff
@@ -429,8 +433,34 @@ func GetBindMount(args []string, contextDir string) (specs.Mount, error) {
...
+		}
+	}
+
+	if fromImage != "" {
+		mountPoint := ""
+		if additionalMountPoints != nil && !selinux.GetEnabled() {
+			if val, ok := additionalMountPoints[fromImage]; ok {
+				mountPoint = val
+				// its a additional mountpoint change ownership
+				//newMount.Options = append(newMount.Options, "U")
```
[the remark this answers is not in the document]
flouthoc: removed.

pkg/parse/parse.go
```diff
@@ -496,61 +531,100 @@ func GetCacheMount(args []string) (specs.Mount, error) {
...
-		return newMount, errors.Wrapf(err, "Unable to change uid,gid of cache directory")
+		// we need to create cache on host if no image is being used
+
+		// since type is cache and cache can be resused by consecutive builds
+		// create a common cache directory, which should persists on hosts within temp lifecycle
+		// add subdirectory if specified
+		if id != "" {
+			newMount.Source = filepath.Join(GetTempDir(), BuildahCacheDir, id)
+		} else {
+			newMount.Source = filepath.Join(GetTempDir(), BuildahCacheDir)
```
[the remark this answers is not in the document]
flouthoc: `id` specifies a different `cache` directory on host if user wants. Content is `cached` on host it is expected to use a common `cache` directory if `id` is not specified.

For the case of `cache` with `from` , `id` is practically `noop` since src of a cache is a `image` or a `stage` not a cache on host machine so no point of differentiation.

pkg/parse/parse.go
```diff
@@ -429,8 +433,34 @@ func GetBindMount(args []string, contextDir string) (specs.Mount, error) {
...
 			// See also the handling of the equivalent "delegated" and "cached" in ValidateVolumeOpts
 		default:
-			return newMount, errors.Wrapf(errBadMntOption, kv[0])
+			return newMount, "", errors.Wrapf(errBadMntOption, kv[0])
+		}
+	}
+
+	if fromImage != "" {
+		mountPoint := ""
+		if additionalMountPoints != nil && !selinux.GetEnabled() {
```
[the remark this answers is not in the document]
flouthoc: @nalind As of now multi-stage builds create container for new stages with `new selinux` label. Which is making it impossible to mount stages across each other. I fixed it with a commit which is now removed from this PR https://github.com/containers/buildah/commit/3941014503db5c6c8c622abb79ece25f43826aca  as @rhatdan suggested a better approach and advised change to be in a new PR.

I added a note here as well https://github.com/containers/buildah/pull/3590#issuecomment-956349109
and some detail here as well https://github.com/containers/buildah/pull/3590#discussion_r735822539

I can assure this check will not break any of the existing functionality and i intend to fix this `SElinux` labeling for `multistage` build in a new PR. So its easier to review and discuss.

pkg/parse/parse.go
```diff
@@ -496,61 +531,100 @@ func GetCacheMount(args []string) (specs.Mount, error) {
...
-	// since type is cache and cache can be resused by consecutive builds
-	// create a common cache directory, which should persists on hosts within temp lifecycle
-	// add subdirectory if specified
-	if id != "" {
-		newMount.Source = filepath.Join(GetTempDir(), BuildahCacheDir, id)
+	if fromImage != "" {
+		// do not create cache on host
+		// instead use read-only mounted image as cache
+		mountPoint := ""
+		if additionalMountPoints != nil && !selinux.GetEnabled() {
```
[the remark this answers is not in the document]
flouthoc: @nalind As of now multi-stage builds create container for new stages with new selinux label. Which is making it impossible to mount stages across each other. I fixed it with a commit which is now removed from this PR 3941014 as @rhatdan suggested a better approach and advised change to be in a new PR.

I added a note here as well #3590 (comment)
and some detail here as well #3590 (comment)

I can assure this check will not break any of the existing functionality and i intend to fix this SElinux labeling for multistage build in a new PR. So its easier to review and discuss.

flouthoc:
@nalind Added requested tests and resolved all the nits. One of them is unresolved but i have added comment for reason behind it. https://github.com/containers/buildah/pull/3590#discussion_r745155252

New tests added
* ``buildah run`: only simple `from` test is added for `run` since multistage is not supported and other scenarios are covered in `bud.tests`
* `buildah build`: tests for multiple `--mount` stmts to make sure code is not getting confused.

pkg/parse/parse.go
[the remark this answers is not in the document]
nalind: `cache` mounts should always default to read-write, and on a closer reading, their `from` arguments aren't expected to be treated as image names.

tests/bud/buildkit-mount-from/Dockerfilecachemultiplefrom
```diff
@@ -0,0 +1,4 @@
+FROM alpine
+RUN mkdir /test
+# use another image as cache source
+RUN --mount=type=cache,from=buildkitbase,target=/test --mount=type=cache,from=buildkitbase,target=/test2 cat /test2/hello
```
nalind, line 4: This doesn't verify the contents of /test, but then `cache` mounts aren't supposed to refer to images with their `from` arguments.

tests/bud/buildkit-mount-from/Dockerfilebindfrom
```diff
@@ -0,0 +1,4 @@
+FROM alpine
+RUN mkdir /test
```
nalind, line 2: Is this supposed to be required before it can be used as a mount point?

tests/bud/buildkit-mount-from/Dockermultistagefrom
```diff
@@ -0,0 +1,6 @@
+FROM quay.io/jitesoft/alpine as builder
```
nalind, line 1: Why the sudden change in base image?

util/util.go
```diff
@@ -175,6 +175,23 @@ func FindImage(store storage.Store, firstRegistry string, systemContext *types.S
 	return ref, localImage.StorageImage(), nil
 }
 
+// LookupImage returns *Image to corresponding imagename or id
```
nalind, line 178: Move this to a new `internal/util` package, if it needs to exist at all.  Two of its callers call it from a loop, where creating new runtime objects for each iteration is going to be wasteful.

nalind (changes requested):
Needs tests for mounts with `type=bind` with no `from`, which should use the build context as the default root location, potentially in combination with `source` pointing to a relative location inside the build context.
Needs tests where multistage builds use `type=bind` with `from` pointing to a previous stage's nickname.
Test cases that expect mounts of `type=cache` to recognize `from` values as image names shouldn't be expected to succeed, as `from` in that context can only refer to other build stages, and not images.
`buildah run` test cases should exercise cases where `type=bind,from=...` and `type=cache,from=...` specify container names as `from` values.
Also, please refrain from marking a comment as resolved when we haven't reached agreement.

flouthoc:
Hi @nalind ,

Resolved all your feedback points and added requested test. Except for last test. PTAL

> Needs tests for mounts with `type=bind` with no `from`, which should use the build context as the default root location, potentially in combination with `source` pointing to a relative location inside the build context.

Following request tests without `from` were already added here: https://github.com/containers/buildah/pull/3548

>Needs tests where multistage builds use `type=bind` with `from` pointing to a previous stage's nickname.

By nickname i am assuming `FROM base as <nickname>` added such case in this PR -> [Test](https://github.com/containers/buildah/blob/08e421040b4763f43ef0a7fcadd9c01c1aaee96f/tests/bud/buildkit-mount-from/Dockerfilemultistagefrom)

 >Test cases that expect mounts of `type=cache` to recognize `from` values as image names shouldn't be expected to succeed, as `from` in that context can only refer to other build stages, and not images. 

Added this case where test is expected to fail -> [Test](https://github.com/containers/buildah/blob/08e421040b4763f43ef0a7fcadd9c01c1aaee96f/tests/bud/buildkit-mount-from/Dockerfilecachefromimage)
 
 >`buildah run` test cases should exercise cases where `type=bind,from=...` and `type=cache,from=...` specify container names as `from` values. Also, please refrain from marking a comment as resolved when we haven't reached agreement.

I was unable to add this as following PR is only scoped to `buildkit` so as of now it can only mount `images` and `stages`. Mounting container seems like a `buildah` specific feature so I'll do that in a new PR if that works.

util/util.go
flouthoc: Fixed in latest commit.

tests/bud/buildkit-mount-from/Dockermultistagefrom
flouthoc: Fixed in latest commit, Actually docker was rate-limiting my pulls on local :sweat_smile:  so had to change for local and i committed it by mistake.

tests/bud/buildkit-mount-from/Dockerfilebindfrom
flouthoc: Removed in latest commit from this file. No not necessary i only added it so i could test if directory is empty without mount.

tests/bud/buildkit-mount-from/Dockerfilecachemultiplefrom
flouthoc: Fixed test case.

tests/bud.bats
```diff
@@ -3698,3 +3698,109 @@ _EOF
   target=bud-group
   run_buildah build --signature-policy ${TESTSDIR}/policy.json -t ${target} ${TESTSDIR}/bud/group
 }
+
+@test "bud-with-mount-bind-from-like-buildkit" {
+  skip_if_no_runtime
+  skip_if_in_container
+  # build base image which we will use as our `from`
+  run_buildah build -t buildkitbase --signature-policy ${TESTSDIR}/policy.json -f ${TESTSDIR}/bud/buildkit-mount-from/Dockerfilebuildkitbase ${TESTSDIR}/bud/buildkit-mount-from/
+  # try reading something from persistant cache in a different build
```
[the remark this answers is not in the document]
flouthoc: Fixed in latest commit.

tests/bud/buildkit-mount-from/Dockerfilebuildkitbase
```diff
@@ -0,0 +1,2 @@
+FROM alpine
```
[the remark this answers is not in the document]
flouthoc: Fixed in latest commit.

docs/buildah-run.1.md
```diff
@@ -151,6 +153,8 @@ Current supported mount TYPES are bind, cache, secret and tmpfs. <sup>[[1]](#Foo
 
               · gid: gid for cache directory.
 
+              · from: image name for the root of the source. Defaults to host cache directory.
```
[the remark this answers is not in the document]
flouthoc: I was unable to add this in following PR, as PR is only scoped to `buildkit` so as of now it can only mount images and stages. Mounting container seems like a `buildah` specific feature so I'll do that in a new PR if that works.

pkg/parse/parse.go
```diff
@@ -377,13 +381,14 @@ func getMounts(mounts []string, contextDir string) (map[string]specs.Mount, erro
 }
 
 // GetBindMount parses a single bind mount entry from the --mount flag.
-func GetBindMount(args []string, contextDir string) (specs.Mount, error) {
+func GetBindMount(args []string, contextDir string, store storage.Store, imageMountLabel string, additionalMountPoints map[string]string) (specs.Mount, string, error) {
```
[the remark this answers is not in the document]
flouthoc: Fixed in latest commit.

pkg/parse/parse.go
```diff
@@ -496,61 +530,101 @@ func GetCacheMount(args []string) (specs.Mount, error) {
...
 			}
 			newMount.Destination = kv[1]
 			setDest = true
+		case "src", "source":
+			if len(kv) == 1 {
+				return newMount, "", errors.Wrapf(optionArgError, kv[0])
+			}
+			newMount.Source = kv[1]
 		case "mode":
 			mode, err = strconv.Atoi(kv[1])
```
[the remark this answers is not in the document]
flouthoc: Fixed in latest commit.

run_linux.go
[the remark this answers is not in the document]
flouthoc: Handling and returning error in latest commit. Only ignoring if `image is not found` which means image is already removed or its a stage.

pkg/parse/parse.go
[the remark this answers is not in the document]
flouthoc: Fixed in latest commit.

internal/parse/parse.go
```diff
@@ -0,0 +1,358 @@
...
+	BuildahCacheDir = "buildah-cache"
+)
+
+var (
+	errBadMntOption = errors.Errorf("invalid mount option")
+	optionArgError  = errors.Errorf("must provide an argument for option")
+	noDestError     = errors.Errorf("must set volume destination")
+)
+
+// GetBindMount parses a single bind mount entry from the --mount flag.
```
nalind, line 37: Please describe what the `string` return value is, and what the caller's supposed to do with it.  It's not obvious.

internal/parse/parse.go
```diff
@@ -0,0 +1,358 @@
...
+	if err != nil {
+		return newMount, fromImage, err
+	}
+	newMount.Options = opts
+
+	return newMount, fromImage, nil
+}
+
+// GetCacheMount parses a single cache mount entry from the --mount flag.
+func GetCacheMount(args []string, store storage.Store, imageMountLabel string, additionalMountPoints map[string]string) (specs.Mount, string, error) {
```
nalind, line 149: Please describe what the `string` return value is, and what the caller's supposed to do with it.  It's not obvious.

run_linux.go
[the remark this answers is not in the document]
nalind: If it's not an image, is it a container?  It's a problem that we don't already know what the right thing to do is when we get here.

run_linux.go
```diff
@@ -2733,14 +2741,34 @@ func (b *Builder) getSSHMount(tokens []string, count int, sshsources map[string]
...
+	for _, image := range artifacts.MountedImages {
+		if image != "" {
+			i, err := internalUtil.LookupImage(context, b.store, image)
+			// its not necessary we are going to find image, could be stage as well.
+			if err == nil {
+				// silently try to unmount and do nothing
+				// if image is being used by something else
+				_ = i.Unmount(false)
+			}
+			if errors.Unwrap(errors.Unwrap(err)) == storagetypes.ErrImageUnknown {
```
nalind, line 2762: This would be a good place to use `errors.Cause()`, so we wouldn't be depending on how many levels of wrapping the underlying error had gotten, but we should be able to know the correct thing to do for a given name - the logic that gave us that name knew what it was referring to, but it currently doesn't tell us.

tests/bud.bats
```diff
@@ -3719,3 +3719,155 @@ _EOF
...
+}
+
+@test "bud-with-mount-bind-from-without-source-like-buildkit" {
+  skip_if_no_runtime
+  skip_if_in_container
+  mkdir ${TESTDIR}/bud
+  cp -R ${TESTSDIR}/bud/buildkit-mount-from ${TESTDIR}/bud/buildkit-mount-from
+  # build base image which we will use as our `from`
+  run_buildah build -t buildkitbase --signature-policy ${TESTSDIR}/policy.json -f ${TESTDIR}/bud/buildkit-mount-from/Dockerfilebuildkitbase ${TESTDIR}/bud/buildkit-mount-from/
+  # try reading something from persistent cache in a different build
```
nalind, line 3744: Suggest "another image" instead of "persistent cache" here.

tests/bud/buildkit-mount-from/Dockerfilecachefrom
```diff
@@ -0,0 +1,8 @@
+FROM alpine as builder
+RUN mkdir subdir
+COPY hello .
+
+FROM alpine as second
+RUN mkdir /test
+# use another image as cache source
```
nalind, line 7: Suggest "another stage" instead of "another image".

tests/bud/buildkit-mount-from/Dockerfilecachemultiplefrom
```diff
@@ -0,0 +1,10 @@
+FROM alpine as builder
+COPY hello .
+
+FROM alpine as builder2
+COPY hello2 .
+
+FROM alpine
+RUN mkdir /test
+# use another image as cache source
```
nalind, line 9: Suggest "other stages" instead of "another image".

nalind (changes requested):

internal/parse/parse.go
```diff
@@ -0,0 +1,358 @@
...
+	}
+
+	if !setDest {
+		return newMount, fromImage, noDestError
+	}
+
+	// buildkit parity: support absolute path for sources from current build context
+	if strings.HasPrefix(newMount.Source, ".") || newMount.Source == "" || !filepath.IsAbs(newMount.Source) || fromImage != "" || fromContainerFile {
+		// path should be /contextDir/specified path
+		newMount.Source = filepath.Join(contextDir, filepath.Clean(string(filepath.Separator)+newMount.Source))
```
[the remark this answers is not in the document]
flouthoc: Added a check in `else` block if mount is from `Containerfile` then contextDir is joined but if its from `buildah run --mount` raw `absolute` path is allowed.

internal/parse/parse.go
```diff
@@ -0,0 +1,358 @@
...
+		setReadOnly bool
+	)
+	fromImage := ""
+	newMount := specs.Mount{
+		Type: TypeBind,
+	}
+	// if id is set a new subdirectory with `id` will be created under /host-temp/buildah-build-cache/id
+	id := ""
+	//buidkit parity: cache directory defaults to 755
+	mode, _ := strconv.ParseUint("0755", 8, 32)
```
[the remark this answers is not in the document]
flouthoc: fixed

internal/parse/parse.go
```diff
@@ -0,0 +1,358 @@
...
+			if val, ok := additionalMountPoints[fromImage]; ok {
+				mountPoint = val
+			}
+		}
+		// Cache does not supports using image so if not stage found
+		// return with error
+		if mountPoint == "" {
+			return newMount, "", fmt.Errorf("no stage or image found with name %s", fromImage)
+		}
+		contextDir := mountPoint
```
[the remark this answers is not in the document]
flouthoc: fixed

internal/parse/parse.go
```diff
@@ -0,0 +1,358 @@
...
+		newMount.Options = append(newMount.Options, "ro")
```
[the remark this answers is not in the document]
flouthoc: fixed

internal/parse/parse.go
```diff
@@ -0,0 +1,358 @@
...
+		} else {
+			newMount.Source = filepath.Join(GetTempDir(), BuildahCacheDir, newMount.Destination)
+		}
+		// create cache on host if not present
+		err = os.MkdirAll(newMount.Source, os.FileMode(mode))
+		if err != nil {
+			return newMount, "", errors.Wrapf(err, "Unable to create build cache directory")
+		}
+		//buildkit parity: change uid and gid if specificed otheriwise keep `0`
+		err = os.Chown(newMount.Source, uid, gid)
```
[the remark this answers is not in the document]
flouthoc: Yes `MkdirAllAndChownNew` makes more sense here added.

internal/parse/parse.go
```diff
@@ -0,0 +1,358 @@
...
+	}
+
+	if !setDest {
+		return newMount, noDestError
+	}
+
+	return newMount, nil
+}
+
+func GetTempDir() string {
```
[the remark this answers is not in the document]
flouthoc: `pkg/parse` already contains `GetTempDir` made this one internal as `getTempDir` as signature could be changed in future.

internal/parse/parse.go
flouthoc: added.

internal/parse/parse.go
flouthoc: added.

tests/bud.bats
flouthoc: fixed

tests/bud/buildkit-mount-from/Dockerfilecachefrom
flouthoc: fixed

tests/bud/buildkit-mount-from/Dockerfilecachemultiplefrom
flouthoc: fixed

run_linux.go
flouthoc: Switched to `errors.Cause`

> If it's not an image, is it a container? It's a problem that we don't already know what the right thing to do is when we get here.

> but we should be able to know the correct thing to do for a given name - the logic that gave us that name knew what it was referring to, but it currently doesn't tell us.

About this part the idea is that if its `ErrImageUnknown` then either original image is unmounted already or its `stage/container`.
For `stage/container` we can only clean up the bind mount which we did for our `--mount` which we already did but we are not authorized to perform explicit `unmount` of `stage or container` 's mountPoint rootfs since its handled by `stage` or `containers` lifecycle.

I also added this as comment in code.

flouthoc:
@nalind Fixed all nits.

> If it's not an image, is it a container? It's a problem that we don't already know what the right thing to do is when we get here.

For this added answer here https://github.com/containers/buildah/pull/3590#discussion_r771154742

run_linux.go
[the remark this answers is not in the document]
nalind: This can produce a different result than it did before we started the command, because an image with a matching name could have been pulled in between.  Keep track of whether or not a name refers to an image, either by not including it in the list of `MountedImages` if it's not actually an image (since we don't need or expect to unmount containers anyway), or by some other means, and do the right thing for it.

internal/parse/parse.go
nalind: It looks like the caller needs to unmount the image with the returned name, sometimes.  If we changed things so that the string value was _only_ returned when we mounted an image, and the value was understood to be the name (or better, the ID) of an image that needed to be unmounted when cleaning up, it would be much easier for callers to clean up after themselves.

internal/parse/parse.go
nalind: Does the caller even need the value?  One caller discards it immediately, the other currently collects it into a list that it wants to use for a list of names of images that it will try to unmount later on, and this value can never be an image's name.

run_linux.go
```diff
@@ -260,7 +264,7 @@ rootless=%d
 	}
 
 	defer func() {
-		if err := cleanupRunMounts(mountPoint, runArtifacts); err != nil {
+		if err := b.cleanupRunMounts(options.SystemContext, mountPoint, runArtifacts); err != nil {
 			options.Logger.Errorf("unabe to cleanup run mounts %v", err)
```
nalind, line 268: Typo: "unabe" should be "unable".

pkg/parse/parse.go
```diff
@@ -346,7 +345,7 @@ func getMounts(mounts []string, contextDir string) (map[string]specs.Mount, erro
 		tokens := strings.Split(arr[1], ",")
 		switch kv[1] {
 		case TypeBind:
-			mount, err := GetBindMount(tokens, contextDir)
+			mount, _, err := internalParse.GetBindMount(tokens, false, contextDir, store, "", nil)
```
nalind, line 348: In cases where we know we mounted an image here, we definitely want to be passing the image's name (or better, its ID) back to our caller so that it can clean up after itself.

internal/parse/parse.go
```diff
@@ -0,0 +1,377 @@
...
+		newMount.Options = append(newMount.Options, "rw")
+		if strings.HasPrefix(newMount.Source, ".") || newMount.Source == "" || !filepath.IsAbs(newMount.Source) || fromImage != "" {
+			// path should be /contextDir/specified path
+			newMount.Source = filepath.Join(mountPoint, filepath.Clean(string(filepath.Separator)+newMount.Source))
+		}
+	} else {
+		// we need to create cache on host if no image is being used
+
+		// since type is cache and cache can be reused by consecutive builds
+		// create a common cache directory, which should persists on hosts within temp lifecycle
```
nalind, line 266: Typo: "which should persists" should be either "which should persist" or "which persists"

internal/parse/parse.go
```diff
@@ -0,0 +1,377 @@
...
+		//TODO: remove this selinux check when comment is resolved. https://github.com/containers/buildah/pull/3590#issuecomment-956349109
+		if additionalMountPoints != nil && (selinux.EnforceMode() != 1) {
+			if val, ok := additionalMountPoints[fromImage]; ok {
+				mountPoint = val
+			}
+		}
+		// if mountPoint of image was not found in additionalMap
+		// or additionalMap was nil, try mounting image
+		if mountPoint == "" {
+			image, err := internalUtil.LookupImage(nil, store, fromImage)
```
nalind, line 107: This needs to not pass `nil` in as the first argument.  It looks like the caller needs to start having to pass a `SystemContext` in to `GetBindMount()` for us to pass along here.

nalind (changes requested):

internal/parse/parse.go
```diff
@@ -0,0 +1,377 @@
...
+		newMount.Options = append(newMount.Options, "shared")
+	}
+
+	// buildkit parity: cache must writable unless `ro` or `readonly` is configured explicitly
+	if !setReadOnly && fromImage == "" {
+		newMount.Options = append(newMount.Options, "rw")
+	}
+
+	// buildkit parity: default bind option for cache must be `rbind`
+	// since we are actually looking for arbitrary content under cache directory
```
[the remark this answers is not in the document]
flouthoc: I confirmed and tested `rbind` is not needed here.

flouthoc:
@nalind Resolved all the comments. PTAL

For this https://github.com/containers/buildah/pull/3590#discussion_r779789753 also made sure that we only return `imagename` when image was actually mounted not for `stages` or `containers`. So now we only attempt `cleanup` on actual mounted images.

imagebuildah/stage_executor.go
```diff
@@ -413,10 +413,56 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
 	return nil
 }
 
+// Returns a map of image:StageMountpoints for RunOpts if any --mount with from is provided
+// check if RUN contains `--mount` with `from`. If yes pre-mount image or stages from executor for Run.
+// stages mounted here will we used be Run().
```
nalind, line 418: Need to update this comment since this doesn't mount images.

internal/parse/parse.go
```diff
@@ -0,0 +1,378 @@
...
+)
+
+var (
+	errBadMntOption = errors.New("invalid mount option")
+	errBadOptionArg = errors.New("must provide an argument for option")
+	errBadVolDest   = errors.New("must set volume destination")
+)
+
+// GetBindMount parses a single bind mount entry from the --mount flag.
+// Returns specifiedMount and a string containing value of `from` if `from` was used with mount.
```
nalind, line 40: Please update this to mention that the string value is the image, if we mounted one, and empty otherwise.  Returning the image's ID would still be better for finding exactly what to unmount later, for cases where the name gets assigned to a different image while the bind mount is in use.

internal/parse/parse.go
```diff
@@ -0,0 +1,378 @@
...
+				return newMount, "", err
+			}
+
+			mountPoint, err = image.Mount(context.Background(), nil, imageMountLabel)
+			if err != nil {
+				return newMount, "", err
+			}
+			isImageMounted = true
+		}
+		newMount.Options = append(newMount.Options, "ro")
```
nalind, line 122: The bind mount should always default to being read-only, whether or not a `from` was given.

internal/parse/parse.go
```diff
@@ -0,0 +1,378 @@
...
+	if !bindNonRecursive {
+		newMount.Options = append(newMount.Options, "rbind")
+	}
+
+	if !setDest {
+		return newMount, fromImage, errBadVolDest
+	}
+
+	// buildkit parity: support absolute path for sources from current build context
+	if strings.HasPrefix(newMount.Source, ".") || newMount.Source == "" || !filepath.IsAbs(newMount.Source) || fromImage != "" || contextDir != "" {
```
nalind, line 137: This `strings.HasPrefix()` check will also succeed for anything that isn't an absolute path (for example, ".bashrc").  `contextDir` will have been set if `fromImage` was set and we didn't error out, so this can probably be simplified to checking if `contextDir` is set, erroring out if both `newMount.Source` and `contextDir` are empty.

internal/parse/parse.go
```diff
@@ -0,0 +1,378 @@
...
+// GetCacheMount parses a single cache mount entry from the --mount flag.
+func GetCacheMount(args []string, store storage.Store, imageMountLabel string, additionalMountPoints map[string]string) (specs.Mount, error) {
+	var err error
+	var mode uint64
+	var (
+		setDest     bool
+		setShared   bool
+		setReadOnly bool
+	)
+	fromImage := ""
```
nalind, line 171: This variable name is confusing.

internal/parse/parse.go
```diff
@@ -0,0 +1,378 @@
...
+			if val, ok := additionalMountPoints[fromImage]; ok {
+				mountPoint = val
+			}
+		}
+		// Cache does not supports using image so if not stage found
+		// return with error
+		if mountPoint == "" {
+			return newMount, fmt.Errorf("no stage found with name %s", fromImage)
+		}
+		newMount.Options = append(newMount.Options, "rw")
```
nalind, line 262: This would override an explicitly-specified "ro" flag.

internal/parse/parse.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+
+	// buildkit parity: default sharing should be shared
+	// unless specified
+	if !setShared {
+		newMount.Options = append(newMount.Options, "shared")
+	}
+
+	// buildkit parity: cache must writable unless `ro` or `readonly` is configured explicitly
+	if !setReadOnly && fromImage == "" {
```
nalind, line 303: No need to test `fromImage` here; read-write should always be the default, whether or not a `from` was specified.

nalind (changes requested):

internal/parse/parse.go
```diff
@@ -0,0 +1,393 @@
...
+		case "consistency":
+			// Option for OS X only, has no meaning on other platforms
+			// and can thus be safely ignored.
+			// See also the handling of the equivalent "delegated" and "cached" in ValidateVolumeOpts
+		default:
+			return newMount, "", errors.Wrapf(errBadMntOption, kv[0])
+		}
+	}
+
+	// default mount readablity is always readonlu
```
nalind, line 103: Nit: typo "readonlu"

internal/parse/parse.go
```diff
@@ -0,0 +1,393 @@
...
+				mountPoint = val
+			}
+		}
+		// Cache does not supports using image so if not stage found
+		// return with error
+		if mountPoint == "" {
+			return newMount, fmt.Errorf("no stage found with name %s", fromStage)
+		}
+		if !setReadOnly {
+			newMount.Options = append(newMount.Options, "rw")
```
nalind, line 276: Nit: remove this block, and we won't have to also check if `fromStage` is set on line 318.

pkg/parse/parse.go
```diff
@@ -334,304 +334,51 @@ func getMounts(mounts []string, contextDir string) (map[string]specs.Mount, erro
...
 		if len(kv) != 2 || kv[0] != "type" {
-			return nil, errors.Wrapf(errInvalidSyntax, "%q", mount)
+			return nil, mountedImages, errors.Wrapf(errInvalidSyntax, "%q", mount)
 		}
 
 		tokens := strings.Split(arr[1], ",")
 		switch kv[1] {
 		case TypeBind:
-			mount, err := GetBindMount(tokens, contextDir)
+			mount, image, err := internalParse.GetBindMount(nil, tokens, contextDir, store, "", nil)
```
nalind, line 349: This still needs to not be passing `nil` in as a `SystemContext`.

nalind: `getMounts()` and `GetVolumes()` already have changes being made to their signatures to allow passing in a `Store` object, so in terms of API compatibility we've already broken anyone who used `GetVolumes()`.  Have their callers pass in a `SystemContext`, too.

run.go
```diff
@@ -147,6 +148,14 @@ type RunOptions struct {
 	// RunMounts are mounts for this run. RunMounts for this run
 	// will not show up in subsequent runs.
 	RunMounts []string
+	// Map of images,stages and container mountpoint if any from stage executor
```
nalind, line 151: Nit: none of the entries in this map have image names as their keys.

run.go
```diff
@@ -147,6 +148,14 @@ type RunOptions struct {
...
+	StageMountPoints map[string]string
+	// External Image mountes to be cleaned up.
```
nalind, line 153: Nit: typo "mountes"

pkg/parse/parse.go
flouthoc: @nalind Following function is being invoked from `cmd/buildah/run.go` when we do `build run --mount` we don't have any `github.com/containers/image/v5/types".SystemContext` there. Could i pass empty context from there ?

internal/parse/parse.go
```diff
@@ -0,0 +1,390 @@
...
+		case "consistency":
+			// Option for OS X only, has no meaning on other platforms
+			// and can thus be safely ignored.
+			// See also the handling of the equivalent "delegated" and "cached" in ValidateVolumeOpts
+		default:
+			return newMount, "", errors.Wrapf(errBadMntOption, kv[0])
+		}
+	}
+
+	// default mount readablity is always readonly
```
nalind, line 103: Typo: "readablity"

imagebuildah/stage_executor.go
```diff
@@ -413,10 +413,56 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
+					// If the source's name corresponds to the
+					// result of an earlier stage, wait for that
+					// stage to finish being built.
+					if isStage, err := s.executor.waitForStage(s.ctx, from, s.stages[:s.index]); isStage && err != nil {
+						return nil, err
+					}
+					if otherStage, ok := s.executor.stages[from]; ok && otherStage.index < s.index {
+						stageMountPoints[from] = otherStage.mountPoint
+						break
+					}
```
nalind, line 448: Should this be reporting an error instead of quietly ignoring the unrecognized `from` value?

flouthoc: If additional mount point are not filled then map is empty so later image could be mounted. Throwing error here will break the flow.

internal/parse/parse.go
```diff
@@ -0,0 +1,390 @@
...
+			newMount.Options = append(newMount.Options, "rw")
+			mountReadability = true
+		case "readonly":
+			// Alias for "ro"
+			newMount.Options = append(newMount.Options, "ro")
+			mountReadability = true
+		case "shared", "rshared", "private", "rprivate", "slave", "rslave", "Z", "z", "U":
+			newMount.Options = append(newMount.Options, kv[0])
+		case "from":
+			fromImage = kv[1]
```
nalind, line 74: Should verify that `len(kv) > 1`.

internal/parse/parse.go
```diff
@@ -0,0 +1,390 @@
...
+		case "shared", "rshared", "private", "rprivate", "slave", "rslave", "Z", "z", "U":
+			newMount.Options = append(newMount.Options, kv[0])
+			setShared = true
+		case "bind-propagation":
+			if len(kv) == 1 {
+				return newMount, errors.Wrapf(errBadOptionArg, kv[0])
+			}
+			newMount.Options = append(newMount.Options, kv[1])
+		case "id":
+			id = kv[1]
```
nalind, line 219: Should be checking the length of `kv`.

internal/parse/parse.go
```diff
@@ -0,0 +1,390 @@
...
+		case "from":
+			fromStage = kv[1]
```
nalind, line 221: Should be checking the length of `kv`.

internal/parse/parse.go
```diff
@@ -0,0 +1,390 @@
...
+				return newMount, err
+			}
+			newMount.Destination = kv[1]
+			setDest = true
+		case "src", "source":
+			if len(kv) == 1 {
+				return newMount, errors.Wrapf(errBadOptionArg, kv[0])
+			}
+			newMount.Source = kv[1]
+		case "mode":
```
nalind, line 236: Should be checking the length of `kv`.

internal/parse/parse.go
```diff
@@ -0,0 +1,390 @@
...
+			mode, err = strconv.ParseUint(kv[1], 8, 32)
+			if err != nil {
+				return newMount, errors.Wrapf(err, "Unable to parse cache mode")
+			}
+		case "uid":
```
nalind, line 241: Should be checking the length of `kv`.

internal/parse/parse.go
```diff
@@ -0,0 +1,390 @@
...
+			uid, err = strconv.Atoi(kv[1])
+			if err != nil {
+				return newMount, errors.Wrapf(err, "Unable to parse cache uid")
+			}
+		case "gid":
```
nalind, line 246: Should be checking the length of `kv`.

imagebuildah/stage_executor.go
nalind: Ah, right.  An error gets reported either from `GetBindMount()` or `GetCacheMount()`.

imagebuildah/stage_executor.go
```diff
@@ -413,10 +413,56 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
+		if strings.Contains(flag, "from") {
+			arr := strings.SplitN(flag, ",", 2)
+			if len(arr) < 2 {
+				return nil, errors.Errorf("Invalid --mount command: %s", flag)
+			}
+			tokens := strings.Split(arr[1], ",")
+			for _, val := range tokens {
+				kv := strings.SplitN(val, "=", 2)
+				switch kv[0] {
+				case "from":
```
nalind, line 431: Should be checking the length of `kv`.

nalind:
I'm not entirely sure we've got the behavior right for writable bind mounts, otherwise LGTM.

flouthoc:
> I'm not entirely sure we've got the behavior right for writable bind mounts, otherwise LGTM.

Me and @nalind discussed this, so our design has some limitation on `rw` on bind mounts for `images`. Future solution could be something on the lines of `overlay on top of build context` but that will be new feature and on a different PR. @nalind Please correct me if i got this right or not.

@containers/build-maintainers @rhatdan @TomSweeneyRedHat PTAL

internal/parse/parse.go
```diff
@@ -0,0 +1,408 @@
...
+// GetTmpfsMount parses a single tmpfs mount entry from the --mount flag
+func GetTmpfsMount(args []string) (specs.Mount, error) {
+	newMount := specs.Mount{
+		Type:   TypeTmpfs,
+		Source: TypeTmpfs,
+	}
+
+	setDest := false
+
+	for _, val := range args {
```
rhatdan, line 357: Is there no way to consolidate this code together, this is the third time you have code to walk the arg list?

rhatdan:
@flouthoc needs rebase.

flouthoc:
> I'm not entirely sure we've got the behavior right for writable bind mounts, otherwise LGTM.

@nalind  This is implemented in latest commit also added a test to verify this. https://github.com/containers/buildah/blob/fb163aa67b71f1251d64e32af8ee2821f516194c/tests/bud.bats#L3743

internal/parse/parse.go
flouthoc: @rhatdan I agree we can reduce redundancy. Some of the options are not supported in other mounts so `switch` case is slightly different for each one. 

However `parsing` is legacy code this is just a move around. So to prevent huge diff would it work if i do that in a follow up
PR where I will also fix `selinux` issue.

rhatdan:
Sadly @flouthoc needs a rebase.

PR #3609, opened 2021-11-01
build: history should not contain `ARG` values. OTOH `RUN` history should contain final values.

flouthoc:
It seems docker never writes ARG values to build history and only writes
values with RUN statements. As a result it can utilize caching better if
assgnment of an ARG value is never used.

Instance
```Dockerfile
FROM alpine
ARG FOO
ENV FOO=bat
RUN echo $FOO
```

In above example --build-arg FOO=value has no significance as value
is always overriden by ENV.

Following PR makes sure buildah mimics a similar behaviour.

Closes: https://github.com/containers/buildah/issues/1936

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -1112,10 +1112,10 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
 	}
 	switch strings.ToUpper(node.Value) {
 	case "ARG":
-		buildArgs := s.getBuildArgs()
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM
@nalind PTAL

flouthoc:
@TomSweeneyRedHat since history does not contains `ARG` values so we will utilize caching better. Hence older tests were had to be adjusted to expect one less layer. @nalind PTAL.

TomSweeneyRedHat:
LGTM
@nalind?

Issue #3556, opened 2021-09-30
make fails: /go-md2man: No such file or directory - file is searched in wrong location


dilyanpalauzov:
```
$ git describe --always 
v1.11.6-1705-g954c4816

$ git status --ignored
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean

$ make
GO111MODULE=on go build -mod=vendor -ldflags '-X main.GitCommit=954c4816 -X main.buildInfo=1632998185 -X main.cniVersion=v0.8.1 ' -gcflags "" -o bin/buildah -tags "seccomp  btrfs_noversion exclude_graphdriver_btrfs  " ./cmd/buildah
GO111MODULE=on go build -mod=vendor -ldflags '-X main.GitCommit=954c4816 -X main.buildInfo=1632998185 -X main.cniVersion=v0.8.1 ' -o bin/imgtype -tags "seccomp  btrfs_noversion exclude_graphdriver_btrfs  " ./tests/imgtype/imgtype.go
GO111MODULE=on go build -mod=vendor -ldflags '-X main.GitCommit=954c4816 -X main.buildInfo=1632998185 -X main.cniVersion=v0.8.1 ' -o bin/copy -tags "seccomp  btrfs_noversion exclude_graphdriver_btrfs  " ./tests/copy/copy.go
make -C tests/tools
make[1]: Entering directory '/git/containers/buildah/tests/tools'
make[1]: Leaving directory '/git/containers/buildah/tests/tools'
make -C docs
make[1]: Entering directory '/git/containers/buildah/docs'
/bin/sh: line 6: /go-md2man: No such file or directory
make[1]: *** [Makefile:16: buildah.1] Error 127
make[1]: Leaving directory '/git/containers/buildah/docs'
make: *** [Makefile:100: docs] Error 2

$ find -name go-md2man
./tests/tools/vendor/github.com/cpuguy83/go-md2man
./tests/tools/build/go-md2man
```

dilyanpalauzov:
This comes from a9e66ec72b55c7d2a854adca96:
```diff
diff --git a/docs/Makefile b/docs/Makefile
index 8fa61994..f063310c 100644
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,19 +1,34 @@
 PREFIX := /usr/local
 DATADIR := ${PREFIX}/share
 MANDIR := $(DATADIR)/man
-GOMD2MAN ?= ../tests/tools/build/go-md2man
[...]

PR #3610, opened 2021-11-01
docs: Use guaranteed path for go-md2man

flouthoc:
Its not guaranteed that host contains go-md2man so use the one which is
built by buildah.

Closes: https://github.com/containers/buildah/issues/3556

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,7 +1,8 @@
 PREFIX := /usr/local
 DATADIR := ${PREFIX}/share
 MANDIR := $(DATADIR)/man
-GOMD2MAN ?= $(shell command -v go-md2man || echo '$(GOBIN)/go-md2man')
+# Following go-md2man is guaranteed on host
+GOMD2MAN ?= ../tests/tools/build/go-md2man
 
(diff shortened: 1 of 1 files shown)


flouthoc:
@rhatdan @TomSweeneyRedHat Just a small revert.

rhatdan:
LGTM

docs/Makefile
```diff
@@ -1,7 +1,8 @@
 PREFIX := /usr/local
 DATADIR := ${PREFIX}/share
 MANDIR := $(DATADIR)/man
-GOMD2MAN ?= $(shell command -v go-md2man || echo '$(GOBIN)/go-md2man')
+# Following go-md2man is guaranteed on host
+GOMD2MAN ?= ../tests/tools/build/go-md2man
```
TomSweeneyRedHat, line 5: I'm not seeing the `build` directory in upstream.  Is this getting built somewhere?  Should we instead use `./tests/tools/vendor/github.com/cpuguy83/go-md2man`  or maybe add the two directories to the old command as or's?

flouthoc: @TomSweeneyRedHat its generated by the make

Issue #3512, opened 2021-09-12
bud: Do not assume implicit ENV PATH=… in Containerfile


dilyanpalauzov:
```
$ buildah version
Version:         1.22.3
Go Version:      go1.16.6
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        0.4.0
libcni Version:  v0.8.1
image Version:   5.15.2
Git Commit:      
Built:           Wed Aug 25 18:36:22 2021
OS/Arch:         linux/amd64

$ echo "FROM scratch" > Containerfile
$ buildah bud
STEP 1/1: FROM scratch
COMMIT
Getting image source signatures
Copying blob 5f70bf18a086 done  
Copying config 75c4176381 done  
Writing manifest to image destination
Storing signatures
--> 75c4176381a
75c4176381a53c0845d32c89f23e1c27b2f55f2d42e6ecfaaf8f571053b88ae6

$ skopeo copy containers-storage:75c4176381a5 oci-archive:zzz.tar
Getting image source signatures
Copying blob 5f70bf18a086 done  
Copying config 75c4176381 done  
Writing manifest to image destination
Storing signatures
```
zzz.tar:blobs/sha256/75c4176381a53c0845d32c89f23e1c27b2f55f2d42e6ecfaaf8f571053b88ae6 is
> {"created":"2021-09-12T10:04:51.528630784Z","architecture":"amd64","os":"linux","config":{"Env":["PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"],"Labels":{"io.buildah.version":"1.22.3"}},"rootfs":{"type":"layers","diff_ids":["sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef"]},"history":[{"created":"2021-09-12T10:04:51.529415181Z","created_by":"/bin/sh"}]}

```
$ echo "ENV PATH=/abc" >> Containerfile
$ buildah bud 
STEP 1/2: FROM scratch
STEP 2/2: ENV PATH=/abc
COMMIT
Getting image source signatures
Copying blob 5f70bf18a086 [--------------------------------------] 0.0b / 0.0b
Copying config 25bc4e9ec6 done  
Writing manifest to image destination
Storing signatures
--> 25bc4e9ec6e
25bc4e9ec6ed2cd0f0220c7600996d2c1d428b080b30aff0700548b722bfe414
[...]

PR #3611, opened 2021-11-01
Add --unsetenv option to buildah commit and build

rhatdan:
This option will allow users to remove environment variables from the
final image.

Fixes: https://github.com/containers/buildah/issues/3512

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -586,6 +586,10 @@ include:
   "sigpending": maximum number of pending signals (ulimit -i)
   "stack": maximum stack size (ulimit -s)
 
+**--unsetenv** *env*
+
+Unset environment variables from the final image.
+
 **--userns** *how*
 
 Sets the configuration for user namespaces when handling `RUN` instructions.
--- a/docs/buildah-commit.1.md
+++ b/docs/buildah-commit.1.md
@@ -101,6 +101,10 @@ When --timestamp is set, the created timestamp is always set to the time specifi
 
 Require HTTPS and verification of certificates when talking to container registries (defaults to true).  TLS verification cannot be used when talking to an insecure registry.
 
+**--unsetenv** *env*
+
+Unset environment variables from the final image.
+
 ## EXAMPLE
 
 This example saves an image based on the container.
--- a/cmd/buildah/build.go
+++ b/cmd/buildah/build.go
@@ -381,6 +381,7 @@ func buildCmd(c *cobra.Command, inputArgs []string, iopts buildOptions) error {
 		Excludes:                excludes,
 		Timestamp:               timestamp,
 		Platforms:               platforms,
+		UnsetEnvs:               iopts.UnsetEnvs,
 	}
 	if iopts.Quiet {
 		options.ReportWriter = ioutil.Discard
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -11,6 +11,7 @@ import (
 	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/buildah/util"
 	"github.com/containers/common/pkg/auth"
+	"github.com/containers/common/pkg/completion"
 	"github.com/containers/image/v5/pkg/shortnames"
 	storageTransport "github.com/containers/image/v5/storage"
 	"github.com/containers/image/v5/transports/alltransports"
@@ -40,6 +41,7 @@ type commitInputOptions struct {
 	tlsVerify          bool
 	encryptionKeys     []string
 	encryptLayers      []int
+	unsetenvs          []string
 }
 
 func init() {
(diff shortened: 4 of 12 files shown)


docs/buildah-build.1.md
```diff
@@ -586,6 +586,10 @@ include:
   "sigpending": maximum number of pending signals (ulimit -i)
   "stack": maximum stack size (ulimit -s)
 
+**--unsetenv** *env*
+
+Unset environment variables from the final image.
+
 **--userns** *how*
```
dilyanpalauzov, line 593: Isn’t $PATH the only use case?  Thus thereis no need for a parameter to the --unsetenv command line option.

dilyanpalauzov:
The Containerfile specification, as far as I remwmber does not say that there is an implicit PATH variable, which this new command line parameter is supposed to disable.

rhatdan:
No, I believe if you base on top of a lower image, the new image will inherit the environment from the base image,  If you want to remove those environment variables, this would come in handy.

docs/buildah-commit.1.md
```diff
@@ -101,6 +101,10 @@ When --timestamp is set, the created timestamp is always set to the time specifi
 
 Require HTTPS and verification of certificates when talking to container registries (defaults to true).  TLS verification cannot be used when talking to an insecure registry.
 
+**--unsetenv** *env*
+
+Unset environment variables from the final image.
```
TomSweeneyRedHat, line 106: It would be nice to have an example in the EXAMPLE section, ditto elsewhere.

TomSweeneyRedHat:
Tests aren't hip and it looks like you need a rebase.  In the tests, it would be good to define an ENVAR and then unset it rather than relying on PATH.

dilyanpalauzov:
> No, I believe if you base on top of a lower image, the new image will inherit the environment from the base image, If you want to remove those environment variables, this would come in handy.

Has somebody ever come to the idea to remove variables, set by lower images?

I think the only use case for this change is to suppress the implicit PATH when interpreting a `Containerfile`. This implicit PATH is nowhere documented.  The tests shall have the PATH variables in mind.

rhatdan:
```
$ cat /tmp/Containerfile 
from ubi8-init
run echo hi
$ ./bin/buildah build --tag dan /tmp
STEP 1/2: FROM ubi8-init
STEP 2/2: run echo hi
hi
COMMIT dan
Copying blob 5bc03dec6239 skipped: already exists  
Copying blob 525ed45dbdb1 skipped: already exists  
Copying blob a363948ee32d skipped: already exists  
Copying blob 0faab747557f done  
Copying config 67cdfe9d6a done  
Storing signatures
--> 67cdfe9d6a7
Successfully tagged localhost/dan:latest
67cdfe9d6a7886d17efd6f4c40aee6454281bc6d5298a23a2341e8c08ded2f07
$ buildah inspect --format '{{ .OCIv1.Config.Env }}' dan
[PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin container=oci]
```
Since ubi8-init defines PATH and container, you see both defined in the new image.
If I don't want container defined, in my new image I can do the following command.

```
 $ ./bin/buildah build --unsetenv container --tag dan /tmp
STEP 1/2: FROM ubi8-init
STEP 2/2: run echo hi
hi
COMMIT dan
Copying blob 5bc03dec6239 skipped: already exists  
Copying blob 525ed45dbdb1 skipped: already exists  
Copying blob a363948ee32d skipped: already exists  
Copying blob 593493bf944c done  
Copying config 3f23c8b6e4 done  
Storing signatures
--> 3f23c8b6e4c
Successfully tagged localhost/dan:latest
3f23c8b6e4cbf1236185f960038aa04a8f3ab2798e7cb51464bd76d1d275bdba
$ buildah inspect --format '{{ .OCIv1.Config.Env }}' dan
[PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin]
```

Images can have multiple environment variables defined within them,  This way a user can base an image off of a lower image, and remove environment variables.

dilyanpalauzov:
How exactly does Containerfile’s FROM (non-scratch) work?  Does it unfold the base image, add the current Container-file instruction and create a new images, so that the new image is self-contained and when moved to a new system, the inital base image does not have to be downloaded?

Or creating `FROM base image` creates a new layer and a reference to the base image?  Thus when the new image is used, the newest layer is downloaded and the base image is downloaded, too?

dilyanpalauzov:
Ah, I see there is a "org.opencontainers.image.base.name" field in the "schemaVersion":2,"config":{"mediaType":"application/vnd.oci.image.config.v1+json" object.

I have:

```
$ podman images
docker.io/library/alpine  latest      14119a10abf4  2 months ago        5.87 MB

$ mkdir L
$ skopeo copy containers-storage:1411 oci:L
$ grep -r path L/

L/blobs/sha256/696d33ca1510966c426bdcc0daf05f75990d68c4eb820f615edccf7b971935e7:{"created":"2021-08-27T17:19:45.758611523Z","architecture":"amd64","os":"linux","config":{"Env":["PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"],"Cmd":["/bin/sh"]},"rootfs":{"type":"layers","diff_ids":["sha256:e2eb06d8af8218cfec8210147357a68b7e13f7c485b991c288c2d01dc228bb68"]},"history":[{"created":"2021-08-27T17:19:45.553092363Z","created_by":"/bin/sh -c #(nop) ADD file:aad4290d27580cc1a094ffaf98c3ca2fc5d699fe695dfb8e6e9fac20f1129450 in / "},{"created":"2021-08-27T17:19:45.758611523Z","created_by":"/bin/sh -c #(nop)  CMD [\"/bin/sh\"]","empty_layer":true}]}

$ cat Containerfile
FROM alpine
CMD /bin/echo "[$PATH]"

$ buildah bud --unsetenv PATH
STEP 1/2: FROM alpine
STEP 2/2: CMD /bin/echo "[$PATH]"
COMMIT
Copying blob e2eb06d8af82 skipped: already exists  
Copying blob 5f70bf18a086 skipped: already exists  
Copying config e8b4d08029 done  
Storing signatures
--> e8b4d080294
e8b4d08029434d503a1917fca8e8f5014833b61c48221bc927c0ee8881064837

$ podman run e8b4d080294
[/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin]

$ skopeo copy containers-storage:e8b4 oci:F
```

so the Alpine image (L/ directory) definition contains `config":{"Env":["PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"]}`.  The derived image (F/ directory) does not contain it, but printing echo $PATH does print the $PATH from the base image.  So not setting PATH in the newest layer, keeps PATH from the lower layer.  Is there an OCI conformat way to remove a variable from a lower layer?  It can certainly be set to the empty string, but this is just changing the variable value, which anyway can be done, it is not removing variable from the container.

I have built buildah on the rhatdan:unsetenv branch.

docs/buildah-build.1.md
dilyanpalauzov: I still think the PATH variable is the only use case and therefore this switch shall be called `--unsetpathevn` without parameters.

rhatdan:
Podman will add the PATH environment itself, if the image does not contain it. (Docker will too, I believe.)

rhatdan:
@flouthoc @TomSweeneyRedHat @nalind @ashley-cui @umohnani8 @vrothberg @giuseppe PTAL

giuseppe (approved):
LGTM

ashley-cui (commented):
/lgtm

PR #3612, opened 2021-11-02
[1.23] unmarshalConvertedConfig(): handle zstd compression

giuseppe:
#### What type of PR is this?

> /kind api-change
/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

The Docker manifest format doesn't currently support listing layers compressed using zstd, so we trigger an error when we try to convert an in-memory OCI manifest to the Docker format as a preliminary step in reading the image's config blob in the Docker format.

Instead, first create a temporary copy of the manifest, and then force the MIME types for all layers in the temporary copy of the manifest to appear to be compressed using gzip. Both OCI and Docker formats will accept the resulting manifest without issue. We throw the copy away after we've read the config blob, so the wackiness should be contained.

#### How to verify it

New integration test!

None

```release-note
None
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -11,6 +11,7 @@ import (
 	"github.com/containers/buildah/define"
 	"github.com/containers/buildah/docker"
 	"github.com/containers/image/v5/manifest"
+	"github.com/containers/image/v5/pkg/compression"
 	"github.com/containers/image/v5/transports"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage/pkg/stringid"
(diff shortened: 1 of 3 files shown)


rhatdan:
Should we back port or should we release a new buildah?  We have had several nice new features added.

TomSweeneyRedHat:
Changed LGTM, but tests aren't happy.  @rhatdan Yes.  We should backport for a 1.23.2 release out of this branch to be vendored into Podman v3.4 next week so we can spin Podman v3.4.2 for RHEL 8.5.0.2 by November 1.  Then we should create a v1.24 for Fedora

rhatdan:
I don't see this as something we need to rush into Podman.  I think waiting for the next release of Buildah is fine.  zstd compression is not even supported at quay.io yet.

giuseppe:
I am fine to wait longer

PR #3615, opened 2021-11-04
config: support the variant field

nalind:
#### What type of PR is this?

/kind feature

Add the `variant` field, along with methods for setting and querying it, and expose them in the `buildah config` and `buildah inspect` commands.

When setting an initial architecture for a container based on an image which doesn't contain an architecture, or from "scratch", normalize the architecture name we've been given, and set both it and the variant field at the same time.

Provide normalized architecture+variant values in `buildah info`.

#### How to verify it

New integration test!

None

```release-note
None
```

### DIFF
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -43,6 +43,7 @@ type configResults struct {
 	shell                  string
 	stopSignal             string
 	user                   string
+	variant                string
 	volume                 []string
 	workingDir             string
 }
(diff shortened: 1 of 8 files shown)


flouthoc:
LGTM

rhatdan:
Could you add a test for buildah config --variant ...

nalind:
> Could you add a test for buildah config --variant ...

Added.

rhatdan:
I don't see --variant added?
I was thinking of 

buildah from scratch
buildah add foobar
buildah config --arch XYZ --variant foobar 
buildah commit
buildah inspect.

nalind:
> I don't see --variant added? I was thinking of
> 
> buildah from scratch buildah add foobar buildah config --arch XYZ --variant foobar buildah commit buildah inspect.

The "config" test in config.bats does that for both OCI and Docker formats, so I added it there as another option/result combination that it checks.

rhatdan:
OK.
/lgtm

Issue #3544, opened 2021-09-27
buildah not actually stopping on SIGTERM


edsantiago:
```console
$ printf "FROM alpine\nRUN sleep 30\n" | timeout -v --foreground 5 /usr/bin/buildah bud -t foo -
STEP 1/2: FROM alpine
STEP 2/2: RUN sleep 30
timeout: sending signal TERM to command ‘/usr/bin/buildah’
$ [back at prompt]
```
Looks good ... except that, 25 seconds later (when `sleep` finishes), this spews out on my terminal:
```console
$ COMMIT foo
Getting image source signatures
Copying blob e2eb06d8af82 skipped: already exists
Copying blob 8bca338d16ce done
Copying config c5c07e0d36 done
Writing manifest to image destination
Storing signatures
--> c5c07e0d36e
Successfully tagged localhost/foo:latest
c5c07e0d36e0124944e114aefad25e8fa6ad170d4377703fffab5980bf1bde03
```
During those 25 seconds:
```console
$ ps auxww|grep buildah
esm      3804916  0.9  0.1 1492444 39456 pts/14  Sl   12:45   0:00 /usr/bin/buildah-in-a-user-namespace bud -t foo -
esm      3804985  2.0  0.1 1639908 37764 pts/14  Sl   12:45   0:00 buildah-oci-runtime
esm      3805583  0.0  0.0 221400   856 pts/31   S+   12:45   0:00 grep --color buildah
```
Appending `| cat` to the reproducer changes the behavior: instead of getting a prompt, the command pauses for those full 30 seconds, then spits out the build log stuff, then returns to command-line prompt.

In both cases, after 30 seconds, `buildah images` shows the newly-built `foo` image.

I believe this is a bug. I believe SIGTERM should kill buildah and all child processes; and I believe that the image `foo` should never get created.

rhatdan:
@nalind PTAL.
Basically buildah needs to catch sigterm and then kill any runs correct?

vrothberg:
> @nalind PTAL.
> Basically buildah needs to catch sigterm and then kill any runs correct?

And possibly clean up/remove  the build containers?

nalind:
As an unprivileged user, the reproducer kills the originally-started process, but leaves the child that's running as root in a user namespace untouched.

rhatdan:
@nalind should we catch the SIGTERM and then attempt to kill the child before exiting?

PR #3619, opened 2021-11-08
define.TempDirForURL(): show CombinedOutput when a command fails

nalind:
#### What type of PR is this?

/kind other

When `define.TempDirForURL()` runs an external command, if it fails, include the command's output in the error that's handed back to its caller so that the end-user has some idea of what happened.

#### How to verify it

New integration test!

Out-of-band error report.

We really need to stop adding logic to the `define` package.  It drags additional dependencies in for consumers of the package that only want our type definitions, which directly undermines the stated rationale for creating the `define` package and moving type definitions into it.

```release-note
None
```


nalind:
h/t @edsantiago @cevich

edsantiago:
Is this a fix for #3544? #3214?

nalind:
It's for the vague error message you asked about in chat last week.

rhatdan:
LGTM

edsantiago:
Oh, I remember now. LGTM, thank you!

Issue #3587, opened 2021-10-13
Building with `--mount=type=ssh` erroring with `error serving agent: read unix /tmp/blah/ssh_auth_sock->@: use of closed network connection`


jspc:
**Description**

Hello.

When building a container image as per:

```bash
$ podman build --ssh default .
```

From the Dockerfile:

```Dockerfile
FROM python:3.8.1

RUN mkdir -m 700 ${HOME}/.ssh && \
  touch -m 600 ${HOME}/.ssh/known_hosts &&  \
  ssh-keyscan github.com > ${HOME}/.ssh/known_hosts

RUN --mount=type=ssh,required=true ssh git@github.com ; sleep 11m
```

Results in the log line:

```
ERRO[0002] error serving agent: read unix /var/tmp/.buildah-ssh-sock1848960062/ssh_auth_sock->@: use of closed network connection 
```

(Or, of course, whichever directory `ioutil.TempDir("", ".buildah-ssh-sock")` creates).

I've put the `sleep` in there to try and diagnose further:

```bash
# The file exists, and is a socket which my user can access
/var/tmp $ ls -alrt .buildah-ssh-sock1848960062/ssh_auth_sock 
srw--w---- 1 jspc jspc 0 Oct 13 09:52 .buildah-ssh-sock1848960062/ssh_auth_sock

# The file is opened by the expected process
/var/tmp $ sudo lsof !$
COMMAND  PID USER   FD   TYPE             DEVICE SIZE/OFF  NODE NAME
conmon  1419 root   16u  unix 0xffff8ced428fcc80      0t0 28866 /proc/self/fd/15/attach type=SEQPACKET (LISTEN)

# Nothing appears to be wrong with the file on disk /shrug
/var/tmp $ stat !$
stat .buildah-ssh-sock1848960062/ssh_auth_sock
  File: .buildah-ssh-sock1848960062/ssh_auth_sock
  Size: 0               Blocks: 0          IO Block: 4096   socket
Device: 259,2   Inode: 272112      Links: 1
Access: (0620/srw--w----)  Uid: ( 1000/    jspc)   Gid: ( 1000/    jspc)
Access: 2021-10-13 09:52:00.665138756 +0100
Modify: 2021-10-13 09:52:00.665138756 +0100
Change: 2021-10-13 09:52:00.665138756 +0100
 Birth: 2021-10-13 09:52:00.665138756 +0100

# Similarly, trying to authenticate with this socket locally (not via buildah) is broken
[...]

PR #3631, opened 2021-11-17
ssh-agent: Increase timeout before we explicitly close connection

flouthoc:
There are cases where remote will close connection by itself with a message
make sure we give connection enough time instead of closing explictly
early.

**Future improvement:** Relay output and perform close instead of relying on  `ServeAgent` to flush
buffer by closing connection.

Output after this:
```console
STEP 4/4: RUN --mount=type=ssh,required=true ssh git@github.com
Pseudo-terminal will not be allocated because stdin is not a terminal.
Warning: Permanently added the ECDSA host key for IP address '<>' to the list of known hosts.
agent key .. returned incorrect signature type
Hi flouthoc! You've successfully authenticated, but GitHub does not provide shell access.
```
[NO TESTS NEEDED]
There is no convenient way to test this in CI. 

Closes: https://github.com/containers/buildah/issues/3587

### DIFF
--- a/pkg/sshagent/sshagent.go
+++ b/pkg/sshagent/sshagent.go
@@ -111,8 +111,9 @@ func (a *AgentServer) Serve(processLabel string) (string, error) {
 				a.wg.Done()
 			}()
 			// the only way to get agent.ServeAgent is to close the connection it's serving on
+			// TODO: ideally we should use some sort of forwarding mechanism for output instead of manually closing connection.
 			go func() {
-				time.Sleep(500 * time.Millisecond)
+				time.Sleep(2000 * time.Millisecond)
 				c.Close()
 			}()
 		}


flouthoc:
Patch verified here: https://github.com/containers/buildah/issues/3587#issuecomment-971425553

rhatdan:
@vrothberg @ashley-cui PTAL

pkg/sshagent/sshagent.go
```diff
@@ -111,8 +111,9 @@ func (a *AgentServer) Serve(processLabel string) (string, error) {
 				a.wg.Done()
 			}()
 			// the only way to get agent.ServeAgent is to close the connection it's serving on
+			// TODO: ideally we should use some sort of forwarding mechanism for output instead of manually closing connection.
 			go func() {
-				time.Sleep(500 * time.Millisecond)
+				time.Sleep(2000 * time.Millisecond)
```
TomSweeneyRedHat, line 116: I'm OK with the value, but just wonder if we could instead get this value from a config file.

flouthoc: @TomSweeneyRedHat Sure we could that but I am confused if this should be visible to end-user and anybody would ever want to configure this.

Ideally we should remove timeout and add buffer forwarding mechanism which should close connection but that seemed like a bigger change to me and would like to do refactor bunch of things in `sshagent` after discussing with maintainers in a separate thread.

rhatdan: No to config, since users will have no idea what to set it.  None of us have any idea either, since it will probably vary under load.

I say increase the hack for now, and then consider fixing this long term in the future.
LGTM

vrothberg (commented):
/lgtm

pkg/sshagent/sshagent.go
vrothberg: I concur with Dan.  At least I would have absolutely no feeling for what a good value for that would be, and also be worried why I have to make that decision.

TomSweeneyRedHat: FWIW, it would not be a config setting I'd heavily document.  More of a tool for support if the issue popped up in the future, or we found out that 2,000 was too much and was causing some other unanticipated error.  I think Dan and Valentin being unsure about a valid value is why I would like to make it configurable.  But hopefully, this will fix this, and it will be a non-issue.

rhatdan: I would say that if this happens again in the field, then we should put in the work to build a complete solution.

PR #3639, opened 2021-11-19
Retry on conflicts when generating a random image name

mtrmac:
#### What type of PR is this?

> /kind bug

Eliminate the risk of possibly generating a conflicting image name.

#### How to verify it

No non-invasive way.

None

```release-note
None
```

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -248,12 +248,27 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 	}
 	nameToRemove := ""
 	if dest == nil {
-		nameToRemove = stringid.GenerateRandomID() + "-tmp"
-		dest2, err := is.Transport.ParseStoreReference(b.store, nameToRemove)
-		if err != nil {
-			return imgID, nil, "", errors.Wrapf(err, "error creating temporary destination reference for image")
+		attempts := 0
+		for {
+			generatedName := stringid.GenerateRandomID() + "-tmp"
+			dest2, err := is.Transport.ParseStoreReference(b.store, generatedName)
+			if err != nil {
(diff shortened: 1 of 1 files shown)


mtrmac:
Fair point, I should actually write a test for the name collision check at least.

vrothberg (commented):
I think there is a race between checking if the name is free and creating the image with this specific name.  We recently had a similar issue in Podman when assigning random names to containers.

mtrmac:
Oops, you’re right, this is racy anyway; we would need to pre-allocate the name in a way that persistently records that cross-process, which this code does not do.

So, this would make the race smaller but not actually eliminate it, just adding false sense of security and potentially confusing readers of this code. Let’s not do that.

I don’t see how to truly fix this at this point, so I guess we’ll have to hope the 256-bit ID namespace is large enough :)

vrothberg:
Maybe we can find a similar solution here.  What happened in Podman is loop around `CreateContainer` and do another round if the error indicates a name conflict with an existing container.

But I am not sure if that would work with images.

mtrmac:
It probably would _work_, it’s just so ugly and expensive. (We create the image with a deterministic ID, typically using the config digest; all of that would succeed, only a final `Store.SetNames` with the conflicting name would fail — and we would tear down the whole image (re-creating squashed layers in the worst case?), and try again.)

mtrmac:
Strictly speaking the user owns the namespace, so a collision is always possible. Pragmatically, maybe some combination of (container ID, timestamp) could be unique enough. But then the current code uses 256 bits of true randomness (unlike most of Podman, which uses the deterministic PRNG), which is at least as unique.

PR #3640, opened 2021-11-19
Tests should rely only on static/unchanging images

cevich:
#### What type of PR is this?

/kind flake

Around 18th of Nov. 2021, a new/incompatible
`quay.io/libpod/fedora-minimal:latest` was pushed by the quay
auto-builder servicing the podman repo.  This caused a number of
problems across several branches.  To address this, the auto-builder was
disabled and a fixed F34 based image was tagged and pushed as both
`latest` and `34`.  However, all repositories which use this test image
need to be individually updated to reference it.

See https://github.com/containers/podman/pull/12343

#### How to verify it

CI will pass

None filed

None

None

### DIFF
--- a/tests/copy.bats
+++ b/tests/copy.bats
@@ -446,7 +446,7 @@ stuff/mystuff"
 
 @test "copy-preserving-extended-attributes" {
   createrandom ${TESTDIR}/randomfile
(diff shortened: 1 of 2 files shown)


edsantiago (approved):
/lgtm

Thanks @cevich

PR #3657, opened 2021-11-29
image: set MediaType in OCI manifests

nalind:
#### What type of PR is this?

/kind feature

In image-spec 1.0.2, the `MediaType` field is now available in OCI manifests and indexes, so set it.

Update the `imgtype` helper to output the `MediaType` field as-is from OCI image manifests instead of just always supplying the expected value.

#### How to verify it

Updated test helper!

None

```release-note
None
```


rhatdan:
LGTM
@flouthoc @giuseppe @vrothberg @TomSweeneyRedHat @umohnani8 PTAL

flouthoc (approved):

PR #3656, opened 2021-11-29
copier: RemoveAll possibly-directories

nalind:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

When we attempt to remove a directory to make way for a non-directory as part of extracting content, use RemoveAll() instead of Remove().

#### How to verify it

Added to a unit test!

#### Which issue(s) this PR fixes:

Fixes #3655 

#### Does this PR introduce a user-facing change?

```release-note
None
```

Issue #3655, opened 2021-11-29
error when copying entire root FS of image


vrothberg:
**Description**

Copying the entire root FS of fedora to busybox errors out.

**Steps to reproduce the issue:**
1. Create a Containerfile
```Dockerfile
test $ cat Dockerfile 
FROM busybox
COPY --from=fedora / /
```

2. `buildah bud -f Containerfile` 

**Describe the results you received:**

```
error building at STEP "COPY --from=fedora / /": 3 errors occurred:
[...]

PR #3658, opened 2021-11-30
copier.Put: check for is-not-a-directory using lstat, not stat

nalind:
#### What type of PR is this?

/kind bug

When checking if something that we want to overwrite with a directory is already a directory or not, use lstat instead of stat.  If it's a symbolic link, it's not a directory.

This is a subtle behavior change, but it's in line with docker build.

#### How to verify it

A slightly longer unit test!
New conformance test!

Follow up to #3655 and #3656.

```release-note
When COPYing or ADDing a directory to a container where the location of the to-be-added directory is a symbolic link to a different directory, the contents of the directory being added are no longer merged with the contents of the directory to which the link points.  The symbolic link will be replaced by the new directory and its contents.
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -1660,7 +1660,7 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
 				// only check the length if there wasn't an error, which we'll
 				// check along with errors for other types of entries
 				if err == nil && written != hdr.Size {
-					return errors.Errorf("copier: put: error creating %q: incorrect length (%d != %d)", path, written, hdr.Size)
+					return errors.Errorf("copier: put: error creating regular file %q: incorrect length (%d != %d)", path, written, hdr.Size)
 				}
 			case tar.TypeLink:
 				var linkTarget string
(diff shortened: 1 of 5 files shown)


nalind:
(added to the release note)

rhatdan:
@nalind ^^

nalind:
Can you elaborate a bit more on the case?  How did you copy a symlink over a directory?  If the source specified for COPY is a symbolic link, it's dereferenced, so I'm having trouble reproducing what you're describing.

srcshelton:
Here's the commit which fixed the issue:

https://github.com/srcshelton/docker-gentoo-build/commit/e76b152edf0ed3bfdca980cd974c0e62fde060c2

The first base-image (`FROM "${zxtm_image_name}:${zxtm_service_version}" AS zxtm`) has a `/usr/local/zeus/zxtm/conf` relative symlink (to `conf_A` in the same directory), and this the copied to `/opt/zeus` in the image being built from `scratch`.

The second base-image (`FROM "localhost/${image_name}:${service_version}" AS base`) has an `/opt/zeus/zxtm/conf` directory.

Up to `buildah-1.23.1`, the result of this process in this order was that the `scratch` image had a `/opt/zeus/zxtm/conf` relative symlink (still to `conf_A`).  With `buildah-1.24.0`, the result is instead `/opt/zeus/zxtm/conf` being the directory from the second base-image.

I've ben able to fix this change be swapping the order of the `COPY` statements, I was simply curious as to whether this was expected?

Issue #3667, opened 2021-12-09
New buildkit-mount tests: fail in gating


edsantiago:
There's a bunch of new tests in bud.bats, all of them referring to `tests/bud/buildkit-mount`. These are failing in gating:
```
not ok 276 bud-with-mount-like-buildkit
...
# STEP 3/3: RUN --mount=type=bind,source=.,target=/test,z cat /test/input_file
# error building at STEP "RUN --mount=type=bind,source=.,target=/test,z cat /test/input_file": error resolving mountpoints for container "57c3e5eb268e14a8b8bca356504082f5869e66133319cbfeb75533d3d4f26c82": relabeling content in /usr is not allowed
# [ rc=125 (** EXPECTED 0 **) ]
```
Reason: in gating, tests are installed via RPM into `/usr/share/buildah/tests/system`.

I'm well past my done-for-the-week point, and can't fix this today or tomorrow. I would be oh-so-grateful if someone were to fix this soon, but if not I'll take a stab at it on Monday.

flouthoc:
@edsantiago Most of these tests were added by me. This looks like `selinux` issue to me but not why it passes in our CI and not on gating tests.

rhatdan:
Someone added a function to selinux that does.
```
// Prevent users from relabeling system files
func badPrefix(fpath string) error {
	if fpath == "" {
		return ErrEmptyPath
	}

	badPrefixes := []string{"/usr"}
	for _, prefix := range badPrefixes {
		if strings.HasPrefix(fpath, prefix) {
			return fmt.Errorf("relabeling content in %s is not allowed", prefix)
		}
	}
	return nil
}
```
Which is throwing this error.  Can we move the gating tests to /var/tmp?

rhatdan:
I agree what that.

flouthoc:
@edsantiago I agree with you, I'll rectify these tests to do testing in a tmp dir.

flouthoc:
I think above PR would close the issue.

PR #3670, opened 2021-12-11
tests: move buildkit mount tests files from `TESTSDIR` to `TESTDIR` before any modifications

flouthoc:
In certain environments tests are conducted from `/usr` where `selinux`
blocks certain operation so move all files where modifications are
expected from `$TESTSDIR` to `$TESTDIR`

Closes: https://github.com/containers/buildah/issues/3667

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -3641,7 +3641,8 @@ _EOF
 @test "bud-with-mount-like-buildkit" {
   skip_if_no_runtime
   skip_if_in_container
-  run_buildah build -t testbud --signature-policy ${TESTSDIR}/policy.json -f ${TESTSDIR}/bud/buildkit-mount/Dockerfile ${TESTSDIR}/bud/buildkit-mount/
+  cp -R ${TESTSDIR}/bud/buildkit-mount ${TESTDIR}/buildkit-mount
+  run_buildah build -t testbud --signature-policy ${TESTSDIR}/policy.json -f ${TESTDIR}/buildkit-mount/Dockerfile ${TESTDIR}/buildkit-mount/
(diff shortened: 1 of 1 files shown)


tests/bud.bats
```diff
@@ -3641,67 +3641,75 @@ _EOF
 @test "bud-with-mount-like-buildkit" {
   skip_if_no_runtime
   skip_if_in_container
-  run_buildah build -t testbud --signature-policy ${TESTSDIR}/policy.json -f ${TESTSDIR}/bud/buildkit-mount/Dockerfile ${TESTSDIR}/bud/buildkit-mount/
+  cp -R ${TESTSDIR}/bud/buildkit-mount ${TESTDIR}/buildkit-mount
+  run_buildah build -t testbud --signature-policy ${TESTSDIR}/policy.json -f ${TESTDIR}/buildkit-mount/Dockerfile ${TESTDIR}/buildkit-mount/
```
rhatdan, line 3645: The policy.json still has a TESTSDIR.

flouthoc: @rhatdan Tests don't modify `policy` so i didn't move it and its same for all the other old tests as well.

rhatdan:
LGTM

rhatdan:
@edsantiago @giuseppe @vrothberg @flouthoc PTAL

vrothberg (commented):
Let's wait for a head nod from @edsantiago

edsantiago:
This is the approach I was envisioning. Future maintainers might've been grateful for helpful comments by the `cp -R`, explaining why it's necessary in these tests and only these tests, but CI is green so

/lgtm

Issue #3665, opened 2021-12-08
Building with "--squash" doesn't have any caching mechanism


Romain-Geissler-1A:
**Description**

When building with podman --squash, there is no caching mechanism that would allow to not re-run long running operations (like dnf install -y some-package). One would expect that the squashed version of all layers might be cached if re-run with the exact same input.

**Steps to reproduce the issue:**
1. Start a podman in podman, using the very latest podman version available:
```
rgeissler@NCEOBERHEL80009.rnd.amadeus.net[dev]:~$ podman run -t -i --rm --privileged --pull=always docker://quay.io/podman/upstream:latest
Trying to pull quay.io/podman/upstream:latest...
Getting image source signatures
Copying blob 6ffc5b8308c8 skipped: already exists
Copying blob c49946d7b2bc skipped: already exists
Copying blob ebee3d32ce92 skipped: already exists
Copying blob dd53b67b4cc9 skipped: already exists
Copying blob 6785b0eb3e35 skipped: already exists
Copying blob 4545346f2a49 skipped: already exists
Copying blob 33939db6ec56 skipped: already exists
Copying blob 71f578d778fe [--------------------------------------] 0.0b / 0.0b
Copying config 2f32afa5f3 done
Writing manifest to image destination
Storing signatures
[root@3a163a7c92e3 /]# podman version
Client:       Podman Engine
Version:      4.0.0-dev
API Version:  4.0.0-dev
Go Version:   go1.16.11
Git Commit:   3b21ba2f623da5a4228a6236c684d1d0b245f240
Built:        Wed Dec  8 08:07:16 2021
OS/Arch:      linux/amd64
```
2. Prepare some dummy Dockerfile:
```
[root@3a163a7c92e3 /]# mkdir src
[root@3a163a7c92e3 /]# cd /src
[root@3a163a7c92e3 src]# cat >Dockerfile <<END_OF_FILE
FROM fedora:latest
RUN touch /file1
RUN touch /file2
END_OF_FILE
```
3. Build it once with --squash, it gets built as expected:
```
[root@3a163a7c92e3 src]# podman build --squash .
STEP 1/3: FROM fedora:latest
Resolved "fedora" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull registry.fedoraproject.org/fedora:latest...
Getting image source signatures
Copying blob 4545346f2a49 done
Copying config 3059bef432 done
Writing manifest to image destination
Storing signatures
[...]

PR #3674, opened 2021-12-14
stage_executor: re-use all the layers from `cache` for `squashed` case and `commit` as late as possible.

flouthoc:
Current implementation of marking for re-use of cache `commits` on every
`stage` when used with `--squashed` however we should try to re-use as
many layers are possible if `--layers` is specified and `commit` only on
`last instruction` of `last stage` to perform final squash.

Closes: https://github.com/containers/buildah/issues/3665

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -974,7 +974,13 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			}
 		}
 
-		if cacheID != "" && !(s.executor.squash && lastInstruction) {
+		// We want to save history for other layers during a squashed build.
+		// Toggle flag allows executor to treat other instruction and layers
+		// as regular builds and only perform squashing at last
+		squashToggle := false
+		// Note: If the build has squash, we must try to re-use as many layers as possible if cache is found.
+		// So only perform commit if its the lastInstruction of lastStage.
+		if cacheID != "" {
 			logCacheHit(cacheID)
 			// A suitable cached image was found, so we can just
 			// reuse it.  If we need to add a name to the resulting
@@ -988,6 +994,13 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 				}
 			}
 		} else {
+			if s.executor.squash {
+				// We want to save history for other layers during a squashed build.
+				// squashToggle flag allows executor to treat other instruction and layers
+				// as regular builds and only perform squashing at last
+				s.executor.squash = false
+				squashToggle = true
+			}
 			// We're not going to find any more cache hits, so we
 			// can stop looking for them.
 			checkForLayers = false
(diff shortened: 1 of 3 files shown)


flouthoc:
@Romain-Geissler-1A Could you try following PR with `buildah build --layers --squash`, this should try `re-use` all the layers from **cache** expect the **last instruction** of **last stage** where final commit is performed.

As a result final sha would change for last step but all the other layers would be `cached`.

flouthoc:
@nalind @rhatdan PTAL

rhatdan:
LGTM

nalind:
Do we have a test that ensures we don't reuse an image when we shouldn't?

flouthoc:
@nalind The `cache` logic is same as a regular build. Following commit just ensures that for `--squash` builds we `commit` later than sooner.

Probably a check in this existing test to verify that `sha` of a regular build is not equivalent to `sha` of a build with `--squash` would that work ?

TomSweeneyRedHat:
LGTM
once the testing questions are answered.

Romain-Geissler-1A:
@flouthoc There is something I don't get. Should I test like you by doing a first build *WITHOUT* --squash to save the intermediate layers ?

Because, if I don't, it doesn't cache anything. I am testing inside docker, started with --privileged and a variant of this Dockerfile: https://github.com/containers/buildah/blob/main/contrib/buildahimage/upstream/Dockerfile

```
[root@b39d41335c2b src]# buildah version
Version:         1.24.0-dev
Go Version:      go1.16.11
Image Spec:      1.0.2-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        1.0.0
libcni Version:  v1.0.1
image Version:   5.17.0
Git Commit:      700314d6
Built:           Tue Dec 14 20:14:23 2021
OS/Arch:         linux/amd64
BuildPlatform:   linux/amd64

[root@b39d41335c2b src]# cat Dockerfile
FROM registry.access.redhat.com/ubi8/ubi:latest

RUN sleep 20

[root@b39d41335c2b src]# time buildah build --layers --squash .
STEP 1/2: FROM registry.access.redhat.com/ubi8/ubi:latest
STEP 2/2: RUN sleep 20
COMMIT
--> 4bcecbb7680
4bcecbb7680e2475efbe89724567f6859c0580873f580ddc62a0cb523f013dba

real    0m34.544s
user    0m5.576s
sys     0m6.792s
[root@b39d41335c2b src]# time buildah build --layers --squash .
STEP 1/2: FROM registry.access.redhat.com/ubi8/ubi:latest
STEP 2/2: RUN sleep 20
COMMIT
--> c4229a3fa96
c4229a3fa96b93e97634ca3a323478505a17f504c4c796f115820a70e8206091

real    0m29.289s
user    0m5.333s
sys     0m6.301s
```

In both cases, the "sleep 20" command is run, while I would expect it could be skipped the second time.

Note that my real use case if obviously more complex, and uses multi stage builds.

rhatdan:
I would say the tests should be back to back squash, although I would also expect non squash followed by squash or vice versa to work.

flouthoc:
@Romain-Geissler-1A Could you please try again. I made some changes and tested with your use-case, it should work now.

> Note that my real use case is obviously more complex, and uses multi stage builds.

Yes I am aware and i have tested it against various complex `squash` builds.

@rhatdan @nalind  I'll add more tests once @Romain-Geissler-1A confirms if PR is working for him now.

PR #3676, opened 2021-12-14
Allow callers to replace the ContainerPrefix

rhatdan:
This idea of this PR is to allow Podman to add a Podman
prefix to containers. This would allow it to keep track
of containers created by Podman and make it easier to remove
them when it wants to remove all.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/buildah.go
+++ b/buildah.go
@@ -257,6 +257,8 @@ type BuilderOptions struct {
 	// or "scratch" to indicate that the container should not be based on
 	// an image.
 	FromImage string
+	// ContainerSuffix is the suffix to add for generated container names
+	ContainerSuffix string
 	// Container is a desired name for the build container.
 	Container string
 	// PullPolicy decides whether or not we should pull the image that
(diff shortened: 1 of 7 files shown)


vrothberg (commented):
`s/Prefix/Suffix/` in the commit message.

vrothberg:
CI is barking due to missing tests. Would it make sense to expose this as a (hidden) flag in `buildah build` to make sure we catch regressions early?

flouthoc:
Could you also add related issue to this PR. I am unable to check why do we need prefix in `container` names. 
Since user could only remove `podman` containers.

flouthoc:
Nevermind I think I got the issue https://github.com/containers/podman/issues/12353#issuecomment-994568793 . PR has intention to make podman smart about the `build` container it generates via `buildah` so leaked containers could be cleaned up.

rhatdan:
@flouthoc correct, if we suffix the containers with _internal_podman, then we know we can blow them away on a rm -f --all.

flouthoc (approved):

ashley-cui (commented):
LGTM

Issue #3675, opened 2021-12-14
Invalid value for memory-swap: invalid size: '-1'


ripta:
**Description**

`podman build` in the latest release of podman (v3.4.4) rejects the `--memory-swap` value of `-1`. Minimal reproduction:

```
❯ cat Dockerfile
FROM ubuntu:20.04
WORKDIR /moo

❯ podman build --memory-swap -1 .
Error: invalid value for memory-swap: invalid size: '-1'
```

**Describe the results you received:**
The flag value `-1` is rejected. I tried other values, including `1024M` and `1G` seem to work as expected.

**Describe the results you expected:**
I expected the flag value `-1` to mean unlimited, as documented on `podman build --help`:

```
      --memory-swap string                           swap limit equal to memory plus swap: '-1' to enable unlimited swap
```

**Output of `podman version` if reporting a `podman build` issue:**

```
❯ podman version
Client:
Version:      3.4.4
API Version:  3.4.4
Go Version:   go1.16.8
Git Commit:   f6526ada1025c2e3f88745ba83b8b461ca659933
Built:        Wed Dec  8 13:13:57 2021
OS/Arch:      darwin/amd64

Server:
Version:      3.4.2
API Version:  3.4.2
Go Version:   go1.16.8
Built:        Fri Nov 12 12:25:37 2021
OS/Arch:      linux/amd64

❯ podman machine list
NAME                     VM TYPE     CREATED         LAST UP            CPUS        MEMORY      DISK SIZE
podman-machine-default*  qemu        40 minutes ago  Currently running  2           4.295GB     10.74GB
```

Thanks!

flouthoc:
@ripta We use `docker` helpers to parse human readable value for `swap` which is ok for `memory` but `swap` also accepts exception of `-1` which is not honored by the same helper. So above PR handlers that.

PR #3679, opened 2021-12-15
parse: allow parsing anomaly non-human value for memory control group

flouthoc:
Memory control group accepts an anomaly value of `-1` for setting `swap` to unlimited.
Buildah must accepts this value if specified.

Reference: https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md#memory

Closes: https://github.com/containers/buildah/issues/3675


flouthoc:
We use docker helpers to parse human readable value for swap which is ok for memory but swap also accepts exception of -1 which is not honored by the same helper. So PR handles that.

flouthoc:
I think buildah `static build` is facing the same issue: https://github.com/containers/crun/pull/812

flouthoc:
> I think buildah `static build` is facing the same issue: [containers/crun#812](https://github.com/containers/crun/pull/812)

@cevich @edsantiago I suspect this a `nixos/nix:latest` regression any suggestions if we should do that same for `buildah` or not. Let me know if there is any better approach.

edsantiago:
I'm totally fine with disabling nix, it seems to be a constant source of problems. I think this is how @cevich disabled it in podman:
```diff
diff --git a/.cirrus.yml b/.cirrus.yml
index 37d102fa..fa064bfe 100644
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -201,6 +201,8 @@ static_build_task:
     name: "Static Build"
     alias: static_build
     only_if: *not_docs
+    # 2021-12-15 nix is dead broken and unmaintained
+    allow_failures: $CI == $CI
     depends_on:
       - unit

```

rhatdan:
LGTM

cevich:
> I think this is how @cevich disabled it in podman

That will simply ignore the results.  I suggest just nuke the whole task (and remove it from `success` also).

TomSweeneyRedHat:
Changes LGTM
@hswong3i any thoughts on the nix issue(s)?

hswong3i:
@saschagrunert PTAL

saschagrunert:
I'm not blocking on removing the static builds.

flouthoc:
Rebased after: https://github.com/containers/buildah/pull/3680

PR #3680, opened 2021-12-15
cirrus: remove `static_build`

flouthoc:
In order to keep CI maintenance simple and not to deal with very frequent `nix` hiccups consensus
was made to remove `static_build` from CI.

Read more here: https://github.com/containers/buildah/pull/3679

------
Before deciding to remove `static_build` we tried things below:

It seems nixos/nix:latest expects nixbld user to be added and specified in
build-users-group but this should be only needed for multi-user
mode.

Hence I suspect latest push has a regression. Lock the nix to last working
image.

Following PR fixes regression for artifacts in CI

See: https://github.com/containers/crun/pull/812
And discussion here: https://github.com/containers/buildah/pull/3679#issuecomment-994720379

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -197,53 +197,6 @@ cross_build_task:
         path: ./bin/*
 
 
-static_build_task:
-    name: "Static Build"
-    alias: static_build
-    only_if: *not_docs
-    depends_on:
-      - unit
-
-    gce_instance:
-        image_name: "${FEDORA_CACHE_IMAGE_NAME}"
-        cpu: 8
-        memory: 12
-        disk: 200
-
-    env:
-        NIX_FQIN: "docker.io/nixos/nix:latest"
-
-    init_script: |
-        set -ex
-        setenforce 0
-
-    nix_cache:
-      folder: '.cache'
-      fingerprint_script: cat nix/*
-
-    build_script: |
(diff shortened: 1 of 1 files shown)


flouthoc:
@edsantiago @cevich Just checking if static build works with last stable version just like `crun`. If it fails then i will remove this from CI like https://github.com/containers/buildah/pull/3679#issuecomment-994720379 as @edsantiago suggested.

cevich:
> Just checking if static build works with last stable version just like crun

IMHO, this will likely be a temporary fix at best.  While it's slowed down recently, eventually crun will be updated and we'll be back in the same boat.

> If it fails then i will remove this from CI

I'm okay with removing it now if that's easier & others agree.  AFAIK the nix stuff isn't backed by anyone anymore, and so that begs the question if anybody actually cares about the CI result.

If we simply need a static binary artifact/test, it should be possible to do that using automation we're willing to own/maintain.

TomSweeneyRedHat:
@hswong3i PTAL

hswong3i:
@saschagrunert PTAL

flouthoc:

vrothberg (commented):
LGTM

giuseppe (approved):
/lgtm

PR #3690, opened 2022-01-06
use libnetwork from c/common for networking

Luap99:
#### What type of PR is this?

/kind feature

setup the netns in the buildah parent process

Do not configure the netns in the runtime child process, this removes
the need to send the network options to the child. This will be needed
for the new libnetwork network interface which cannot be transfered as
json.
To synchronize this between the child and parent we use two pipe pairs.

use libnetwork from c/common for networking

Podman uses the new netavark network stack. Buildah should be able to do
the same. Both projects should use the same networking code which was
move to c/common/libnetwork. The new network interface can use either
CNI or netvavark. Using the same code for podman and buildah is
important to ensure that both use the same backend. Mixing CNI and
netavark is not supported.
This also fixes some outstanding CNI issues, e.g. buildah trying to
connect all cni networks.

```release-note

```


new.go
```diff
@@ -113,6 +116,15 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
 		options.FromImage = ""
 	}
 
+	if options.NetworkInterface == nil {
+		// create the network interface
+		// Note: It is important to do this before we pull any images/create containers.
```
nalind, line 121: Where does someone like me who's reading this comment go to learn why this is important?

new.go
```diff
@@ -313,3 +326,26 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
...
 }
+
+// getNetworkInterface creates the network interface
+func getNetworkInterface(store storage.Store, cniConfDir, cniPluginPath string) (nettypes.ContainerNetwork, error) {
+	conf, err := config.Default()
+	if err != nil {
+		return nil, err
+	}
+	// copy the config to not modify the default by accident
+	newconf := *conf
```
nalind, line 337: This step shouldn't be necessary - the value we got back was created fresh.

new.go
Luap99: The default network backend detection logic needs an empty store to correctly detect that we can use netavark, if the store was not empty it will use CNI to not break existing installs.
If we do this after the pull it will always use CNI even when we would like to use netavark, e.g new installs. 

I will add this to the comment

nalind:
Assuming it passes tests, this looks reasonable.

new.go
Luap99: config.Default() returns a pointer to the global Config, so if I do not copy this it would change the global config which I do not want

nalind: Oof, you're right, but that seems like a bug waiting to happen.

TomSweeneyRedHat:
Changes look reasonable to me, but the test system isn't at all happy and you need a rebase here.

Luap99:
Yeah that is a regression in c/common, I opened https://github.com/containers/common/pull/875 to revert the change for now.

flouthoc:
@Luap99 After this PR My builds are failing with `error creating build container: error opening "/etc/cni/net.d/cni.lock": permission denied`

Luap99:
Ah sorry, is there no rootless testing in CI?
I will fix it on Monday.

flouthoc:
No issues not urgent i am still able to do work with `root` builds.

PR #3693, opened 2022-01-09
executor: Share `selinux` (`process and mount`) `label` of first stage with other stages in same build

flouthoc:
### Why do we need this ?

Buildah `builds` are composed of various `stages`. That is each `stage` in a build is a `container`.

As of now buildah does not provides any `processLabel` and `mountLabel` to the stage container which
results in `c/storage` assigning random generated `selinux` labels to each stage.

As a result this PR also enables following feature on systems with `selinux` enabled: https://github.com/containers/buildah/pull/3590

### Problem
If each stage has its own unqiue `processLabel` and `mountLabel` stages cannot be mounted across each other
in the same build. `selinux` prevents that.

### Solution

We get the `processLabel` and `mountLabel` of the first stage created by the container i.e assigned by `c/stoage`
and share it with all the other subsequent stages of the same build.

### DIFF
--- a/buildah.go
+++ b/buildah.go
@@ -340,6 +340,10 @@ type BuilderOptions struct {
 	// OciDecryptConfig contains the config that can be used to decrypt an image if it is
 	// encrypted if non-nil. If nil, it does not attempt to decrypt an image.
 	OciDecryptConfig *encconfig.DecryptConfig
+	// ProcessLabel is the SELinux process label associated with the container
+	ProcessLabel string
+	// MountLabel is the SELinux mount label associated with the container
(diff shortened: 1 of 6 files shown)


flouthoc:
@nalind @rhatdan @mtrmac @giuseppe @vrothberg @TomSweeneyRedHat PTAL

rhatdan:
LGTM

giuseppe (approved):
/lgtm

flouthoc:
This should also close: https://github.com/containers/buildah/issues/3501 completely

PR #3697, opened 2022-01-10
Fix default CNI paths

Luap99:
#### What type of PR is this?

/kind bug

We need to use the default from containers.conf and not hardcode them in
buildah. This fixes an issue with the cni network backend since it would
try to access /etc/cni/net.d/ even as rootless user. This regression was
introduced in commit f9cff07b8121.

Also hide the cni flags as we do not expect users to change this. The
recommended way is to change them in containers.conf.

#### How to verify it

run `buildah bud` rootless

```release-note

```

### DIFF
--- a/contrib/completions/bash/buildah
+++ b/contrib/completions/bash/buildah
@@ -404,8 +404,6 @@ return 1
      --cap-drop
      --cert-dir
      --cgroup-parent
-     --cni-config-dir
-     --cni-plugin-path
      --cpu-period
      --cpu-quota
      --cpu-shares
@@ -490,8 +488,6 @@ return 1
      local options_with_args="
      --cap-add
      --cap-drop
-     --cni-config-dir
-     --cni-plugin-path
      --hostname
      --ipc
      --isolation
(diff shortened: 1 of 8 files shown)


Luap99:
@flouthoc This should fix the rootless issue with cni.

rhatdan:
LGTM
Add [NO NEW TESTS NEEDED] to move forward on this.

Or better yet add a test

buildah build --help | grep -- --cni-config-dir
buildah build --help | grep -- --cni-config-plugin-path
Make sure it does not have Default.

Should we hide these options, are these really options we expect users to play with?

Luap99:
I think it would make sense to remove these options. The fields from the bud/run options structs should also be removed IMO but this would be a breaking change. I am not sure if these options are used anywhere (I hope not).

rhatdan:
Just hide the options and remove the documentation from the man pages, should be sufficient.

flouthoc:
@Luap99 Thanks as @rhatdan suggested `NO NEW TESTS NEEDED` is needed.

PR #3701, opened 2022-01-11
bud.bats: use a local git daemon for the git protocol test

nalind:
#### What type of PR is this?

/kind failing-test 

When testing that we can handle contexts that are described using the git:// protocol, launch a local git-daemon with a minimal repository to point the binary at instead of github, which doesn't do that any more.

#### How to verify it

The test should pass now.

We do need git-daemon in the test environment.

```release-note
None
```


flouthoc (approved):

nalind:
Tests are passing because we now skip a test when we can't launch `git daemon` as part of the test, but we should make sure our test environments have it installed.

PR #3714, opened 2022-01-17
buildkit, executor: Add support for inline `FROM --platform=` within Containerfile/Dockerfile

flouthoc:
Allows end users to configure executor's `OS`, `ARCH`,`VARIANT` via inline
`--platform`. See: https://github.com/containers/buildah/issues/3712

Usage

```Dockerfile
FROM --platform=linux/arm64 alpine
RUN uname -a
```

------

While this allows executor to pull base images with custom
`OS`, `ARCH`, `VARIANT` it still allows end-users to tag images with
different format if they need to. As discussed here: https://github.com/containers/buildah/issues/3261

For example

Doing a `buildah build --platform linux/amd64 -t test .` on below
containerfile.

```Dockerfile
FROM --platform=linux/arm64 alpine
RUN uname -a
```

Would result in tagging image as `linux/amd64` while `base` layer would
be pulled for `linux/arm64`.

Closes: https://github.com/containers/buildah/issues/3712
Should Closes this as well: #3261

### DIFF
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -598,6 +598,25 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 			for _, child := range node.Children { // tokens on this line, though we only care about the first
 				switch strings.ToUpper(child.Value) { // first token - instruction
 				case "FROM":
+					for _, flag := range child.Flags {
+						if strings.HasPrefix(flag, "--platform=") {
+							platformString := strings.TrimPrefix(flag, "--platform=")
+							os, arch, variant, err := parse.Platform(platformString)
+							if err != nil {
+								return "", nil, errors.Wrapf(err, "unable to parse platform %q", platformString)
+							}
+							// configure executor's system context for pull's being perform for base images
+							if arch != "" {
+								b.systemContext.ArchitectureChoice = arch
+							}
+							if os != "" {
+								b.systemContext.OSChoice = os
+							}
+							if variant != "" {
+								b.systemContext.VariantChoice = variant
+							}
+						}
+					}
 					if child.Next != nil { // second token on this line
 						// If we have a fromOverride, replace the value of
 						// image name for the first FROM in the Containerfile.
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -2837,6 +2837,25 @@ EOF
   expect_output --substring "\[2/2] STEP 1/2: FROM busybox"
 }
 
+# Following test must fail since we are trying to run linux/arm64 on linux/amd64
+# Issue: https://github.com/containers/buildah/issues/3712
+@test "build-with-inline-platform" {
+  # Tries building image where baseImage has --platform=linux/arm64
+  run_buildah 1 build --signature-policy ${TESTSDIR}/policy.json -t test ${TESTSDIR}/bud/platform
+  expect_output --substring "Exec format error"
+}
+
+# Following test must pass since we want to tag image as arm64
+# Test for use-case described here: https://github.com/containers/buildah/issues/3261
+@test "build-with-inline-platform-amd-but-tag-as-arm" {
+  # Tries building image where baseImage has --platform=linux/arm64
+  run_buildah build --platform linux/arm64 --signature-policy ${TESTSDIR}/policy.json -t test -f ${TESTSDIR}/bud/platform/Dockerfileamd
+  run_buildah inspect --format '{{ .OCIv1.Architecture }}' test
+  # base image is pulled as amd64 but image must be tagged as arm64
+  expect_output --substring "arm64"
+
+}
+
 @test "bud test no --stdin" {
   _prefetch alpine
   mytmpdir=${TESTDIR}/my-dir
--- a/tests/test_runner.sh
+++ b/tests/test_runner.sh
@@ -19,4 +19,4 @@ function execute() {
 TESTS=${@:-.}
 
 # Run the tests.
-execute time bats --tap $TESTS
+execute time bats --tap $TESTS --filter
--- a/tests/bud/platform/Dockerfile
+++ b/tests/bud/platform/Dockerfile
@@ -1,0 +1,2 @@
+FROM --platform=linux/arm64 busybox
+RUN uname -m
--- a/tests/bud/platform/Dockerfileamd
+++ b/tests/bud/platform/Dockerfileamd
@@ -1,0 +1,2 @@
+FROM --platform=linux/amd64 busybox
+RUN uname -m


imagebuildah/executor.go
```diff
@@ -598,6 +598,25 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 			for _, child := range node.Children { // tokens on this line, though we only care about the first
 				switch strings.ToUpper(child.Value) { // first token - instruction
 				case "FROM":
+					for _, flag := range child.Flags {
+						if strings.HasPrefix(flag, "--platform=") {
+							platformString := strings.TrimPrefix(flag, "--platform=")
```
rhatdan, line 603: Would this be better to split on "=".
Is "--platform =" valid in a FROM?

rhatdan: If not,
LGTM

rhatdan:
@vrothberg @nalind @TomSweeneyRedHat @umohnani8 @ashley-cui @giuseppe PTAL

imagebuildah/executor.go
flouthoc: > Would this be better to split on "=".

@rhatdan Parsing is same as we parse other `flags` in executor's code. I am okay to change this one but my intention was to keep consistency for reader : https://github.com/containers/buildah/blob/main/imagebuildah/executor.go#L626


> Is "--platform =" valid in a FROM?

Yes `--platform` is a valid field in `docker` but its not supported by `buildah` following PR just adds support for that. Following issues states need for the same: https://github.com/containers/buildah/issues/3712 & https://github.com/containers/buildah/issues/3261#issuecomment-848838772

tests/bud.bats
```diff
@@ -2837,6 +2837,25 @@ EOF
...
+
+# Following test must pass since we want to tag image as arm64
+# Test for use-case described here: https://github.com/containers/buildah/issues/3261
+@test "build-with-inline-platform-amd-but-tag-as-arm" {
+  # Tries building image where baseImage has --platform=linux/arm64
+  run_buildah build --platform linux/arm64 --signature-policy ${TESTSDIR}/policy.json -t test -f ${TESTSDIR}/bud/platform/Dockerfileamd
+  run_buildah inspect --format '{{ .OCIv1.Architecture }}' test
+  # base image is pulled as amd64 but image must be tagged as arm64
+  expect_output --substring "arm64"
+
```
vrothberg, line 2856: Empty line

tests/bud.bats
```diff
@@ -2837,6 +2837,25 @@ EOF
   expect_output --substring "\[2/2] STEP 1/2: FROM busybox"
 }
 
+# Following test must fail since we are trying to run linux/arm64 on linux/amd64
+# Issue: https://github.com/containers/buildah/issues/3712
```
vrothberg, line 2841: Note that such hard-coded requirements to test platform nail it down to only run on amd64.

If there's an alternative, e.g. to specify `linux/bogus` and check for a specific error message, I'd prefer that.

vrothberg (commented):
@edsantiago PTAL

tests/bud.bats
```diff
@@ -2837,6 +2837,25 @@ EOF
...
+@test "build-with-inline-platform" {
+  # Tries building image where baseImage has --platform=linux/arm64
+  run_buildah 1 build --signature-policy ${TESTSDIR}/policy.json -t test ${TESTSDIR}/bud/platform
+  expect_output --substring "Exec format error"
```
edsantiago, line 2845: This passes on a system with `qemu-user-static` installed. (As in: the build passes, so the test fails).

edsantiago: Friendly suggestion:
```diff
diff --git a/tests/bud.bats b/tests/bud.bats
index 99066bb7..ba5cd4b9 100644
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -2840,9 +2840,26 @@ EOF
 # Following test must fail since we are trying to run linux/arm64 on linux/amd64
 # Issue: https://github.com/containers/buildah/issues/3712
 @test "build-with-inline-platform" {
-  # Tries building image where baseImage has --platform=linux/arm64
-  run_buildah 1 build --signature-policy ${TESTSDIR}/policy.json -t test ${TESTSDIR}/bud/platform
-  expect_output --substring "Exec format error"
+  # Host arch
+  run_buildah info --format '{{.host.arch}}'
+  myarch="$output"
+  for try in amd64 arm64 ppc64le s390x; do
+    if [[ "$otherarch" != "$myarch" ]]; then
+      otherarch="$myarch"
+      break
+    fi
+  done
+  # ...create a Containerfile with --platform=linux/$otherarch
+  run_buildah '?' build --signature-policy ${TESTSDIR}/policy.json -t test ${TESTSDIR}/bud/platform
+  if [[ $status -eq 0 ]]; then
+    # Build succeeded: we must have qemu-user-static installed. Confirm that
+    # the 'uname' emits what we expect
+    expect_output --from="${lines[-1]}" "$otherarch"
+    # FIXME: also inspect the image and confirm the arch
+  else
+    # Build failed: we DO NOT have qemu-user-static installed.
+    expect_output --substring "Exec format error"
+  fi
 }

 # Following test must pass since we want to tag image as arm64
```
Writing the Containerfile and updating the comments and updating the next test are left as an exercise for the reader.

tests/test_runner.sh
```diff
@@ -19,4 +19,4 @@ function execute() {
 TESTS=${@:-.}
 
 # Run the tests.
-execute time bats --tap $TESTS
+execute time bats --tap $TESTS --filter
```
nalind, line 22: This looks like an unintended partial change.

imagebuildah/executor.go
```diff
@@ -598,6 +598,25 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
...
+							// configure executor's system context for pull's being perform for base images
+							if arch != "" {
+								b.systemContext.ArchitectureChoice = arch
+							}
+							if os != "" {
+								b.systemContext.OSChoice = os
+							}
+							if variant != "" {
+								b.systemContext.VariantChoice = variant
+							}
```
nalind, line 617: `b.systemContext` is used for other stages as well.  This change would affect any stages that need to pull images after this point.

nalind:
Why are we doing this here, instead of in the imagebuilder library, as we did for `--chown` and `--chmod` for COPY and ADD instructions, and `--mount` for RUN instructions?

flouthoc:
> Why are we doing this here, instead of in the imagebuilder library, as we did for `--chown` and `--chmod` for COPY and ADD instructions, and `--mount` for RUN instructions?

Image builder it is then. Let me explore that side.

rhatdan:
Since this is going to imagebuilder, I am going to close this PR.

PR #3715, opened 2022-01-17
overlay, mount: add support for non-volatile `upperdir`, `workdir` for overlay volumes

flouthoc:
Expose `MountWithOptions` for overlay which allows users to pass more
verbose configuration for overlay mounts.

For instance `upperdir, workerdir` and in future `volatile`.

Allow users to specify non-volatile `upper` and `workdir` with overlay
mounts.

Usage

```console
buildah from alpine
buildah run -v /something/lower:/test:z,O,upperdir=/somewhere/upperdir,workdir=/somwhere/workdir alpine-working-container cat /test/hello
```

Needed for: https://github.com/containers/podman/pull/12712


flouthoc:
@giuseppe @vrothberg @rhatdan @nalind @mtrmac PTAL

pkg/overlay/overlay.go
```diff
@@ -18,6 +18,11 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+type Options struct {
```
vrothberg, line 21: Please add doc comments.

pkg/overlay/overlay.go
```diff
@@ -93,6 +106,12 @@ func mountHelper(contentDir, source, dest string, _, _ int, graphOptions []strin
 		// Read-write overlay mounts want a lower, upper and a work layer.
 		workDir := filepath.Join(contentDir, "work")
 		upperDir := filepath.Join(contentDir, "upper")
+
+		if opts != nil {
+			workDir = opts.Workdir
```
giuseppe, line 111: should we set these only if the opts version is != "" ?

pkg/overlay/overlay.go
```diff
@@ -61,23 +66,31 @@ func generateOverlayStructure(containerDir string, rootUID, rootGID int) (string
 	return containerDir, nil
 }
 
+// MountWithOptions creates a subdir of the contentDir based on the source directory
+// from the source system.  It then mounts up the source directory on to the
+// generated mount point and returns the mount point to the caller.
+// But allows api to set custom workdir, upperdir and other overlay options
+func MountWithOptions(contentDir, source, dest string, rootUID, rootGID int, graphOptions []string, opts *Options, readonly bool) (mount specs.Mount, Err error) {
```
vrothberg, line 73: The signature looks inconsistent to me.  Why do we mix `Options` with so arguments? `graphOptions` and `readonly` look optional to me.

pkg/overlay/overlay.go
```diff
@@ -61,23 +66,31 @@ func generateOverlayStructure(containerDir string, rootUID, rootGID int) (string
 	return containerDir, nil
 }
 
+// MountWithOptions creates a subdir of the contentDir based on the source directory
+// from the source system.  It then mounts up the source directory on to the
+// generated mount point and returns the mount point to the caller.
+// But allows api to set custom workdir, upperdir and other overlay options
```
mtrmac, line 72: FWIW – this is preexisting, and maybe I’m just ignorant. This documentation, even with the context of the overlay section of mount(8), is completely unintelligible to me. I have no idea how `source`/`contentDir` relate.

(Reading the code, apparently `contentDir` has a specific structure that’s not documented in this package.)

The other options are also undocumented.

pkg/overlay/overlay.go
mtrmac: I agree, we need to be able to add more `Options` members without forcing callers to choose a `Workdir` etc.

nalind:
Please elaborate on the use case, and add tests.

mtrmac (commented):
Warning: I know nothing about overlay (I don’t think I have _ever_ manually mounted one) except what I just read in the documentation, and I know even less about how c/storage uses it.

pkg/overlay/overlay.go
mtrmac: … and the behavior of `readonly`, affecting whether `$contentDir/lower` is used, is completely non-obvious to me as a newcomer to the code.

pkg/overlay/overlay.go
flouthoc: Yes moved this to options struct.

pkg/overlay/overlay.go
flouthoc: added

pkg/overlay/overlay.go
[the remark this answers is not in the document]
flouthoc: @mtrmac `rootUID` and `rootGID` are not optional they are mandatory as of now.

pkg/overlay/overlay.go
```diff
@@ -61,23 +66,31 @@ func generateOverlayStructure(containerDir string, rootUID, rootGID int) (string
...
 // generated mount point and returns the mount point to the caller.  Note that no
 // upper layer will be created rendering it a read-only mount
 func MountReadOnly(contentDir, source, dest string, rootUID, rootGID int, graphOptions []string) (mount specs.Mount, Err error) {
-	return mountHelper(contentDir, source, dest, rootUID, rootGID, graphOptions, true)
+	return mountHelper(contentDir, source, dest, rootUID, rootGID, graphOptions, nil, true)
 }
 
 // NOTE: rootUID and rootUID are not yet used.
-func mountHelper(contentDir, source, dest string, _, _ int, graphOptions []string, readOnly bool) (mount specs.Mount, Err error) {
+func mountHelper(contentDir, source, dest string, _, _ int, graphOptions []string, opts *Options, readOnly bool) (mount specs.Mount, Err error) {
```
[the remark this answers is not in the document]
flouthoc: I am fine with removing `mountHelper` just wanted change to be minimal that's why i avoided this. Since its an API change.

pkg/overlay/overlay.go
```diff
@@ -93,6 +106,12 @@ func mountHelper(contentDir, source, dest string, _, _ int, graphOptions []strin
 		// Read-write overlay mounts want a lower, upper and a work layer.
 		workDir := filepath.Join(contentDir, "work")
 		upperDir := filepath.Join(contentDir, "upper")
+
+		if opts != nil {
+			workDir = opts.Workdir
+			upperDir = opts.Upperdir
+		}
+
```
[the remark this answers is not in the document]
flouthoc: @mtrmac Users using custom `upperdir` and `workdir` are already expecting it to be relayed to overlay directly so I think it would better if they get overlay failures.

pkg/overlay/overlay.go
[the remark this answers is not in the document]
mtrmac: It’s mandatory for the caller to pass them.

They are passed to `mountHelper`, which explicitly ignores them.

So they seem unnecessary for this code to _work_. (I have absolutely no idea what that’s about. Is that a missing feature of some kind, or did it just turn out that the values are never going to be necessary? Or are they not necessary _now_ but necessary _in principle_?)

pkg/overlay/overlay.go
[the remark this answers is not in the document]
mtrmac: I’ll completely defer to owners of the repo on this.

pkg/overlay/overlay.go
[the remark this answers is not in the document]
mtrmac: I’m not sure what “directly” means. Is that “a path, with this module responsible for escaping”, “an option on the mount options format, with the caller responsible for (and trusted to do) escaping, if any”, something else?

Reading a _tiny bit_ more, it seems that only the lower directory actually supports a colon-specified list. But right now, the callers seem to be able to pass arbitrary mount options by using e.g. `upperdir=/some/directory,otheroption` . Are the callers known and trusted not to do that, or is that escalating a privilege?

If the callers “should get overlay failures”, I think it would be best to validate the inputs and reject them if they contain unsupported metacharacters. Admittedly the existing `lowerdir` handling does not do any validation either, so this might be fine, assuming all of the data comes from a similarly-trusted source.

pkg/overlay/overlay.go
```diff
@@ -18,6 +18,20 @@ import (
...
+type Options struct {
+	// The upper directory is normally writable layer in an overlay mount.
+	Upperdir string
+	// The workdir is used to prepare files as they are switched between the layers.
+	Workdir string
+	// Graph options relayed from podman, will be responsible for choosing mount program
+	Graphopts []string
+	// Mark if following overlay is read only
+	Readonly bool
+}
```
mtrmac, line 33: Shouldn’t this be `GraphOpts` and `ReadOnly`? I’m not sure about `UpperDir`, `WorkDir`.

pkg/overlay/overlay.go
```diff
@@ -61,23 +75,32 @@ func generateOverlayStructure(containerDir string, rootUID, rootGID int) (string
...
 // generated mount point and returns the mount point to the caller.  Note that no
 // upper layer will be created rendering it a read-only mount
 func MountReadOnly(contentDir, source, dest string, rootUID, rootGID int, graphOptions []string) (mount specs.Mount, Err error) {
-	return mountHelper(contentDir, source, dest, rootUID, rootGID, graphOptions, true)
+	return mountHelper(contentDir, source, dest, rootUID, rootGID, graphOptions, nil, true)
 }
 
 // NOTE: rootUID and rootUID are not yet used.
-func mountHelper(contentDir, source, dest string, _, _ int, graphOptions []string, readOnly bool) (mount specs.Mount, Err error) {
+func mountHelper(contentDir, source, dest string, _, _ int, graphOptions []string, opts *Options, readOnly bool) (mount specs.Mount, Err error) {
```
mtrmac, line 103: It’s now surprising for `opts.Readonly` and `opts.Graphopts` to be passed to this function, but ignored. I think the individual `graphOptions`, `readOnly` parameters to this function should be dropped (and the two old public functions would be passing them via a single-use `Options` value).

pkg/overlay/overlay.go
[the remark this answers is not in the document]
flouthoc: @mtrmac The value of `upperdir` and `workdir` is user defined so yes if there is vulnerability in overlay fs itself it could cause security issues. But then since its user defined input should we worry about it cause it would be purely end user's fault then. But I am open to suggestions here.

pkg/overlay/overlay.go
```diff
@@ -93,6 +116,12 @@ func mountHelper(contentDir, source, dest string, _, _ int, graphOptions []strin
 		// Read-write overlay mounts want a lower, upper and a work layer.
 		workDir := filepath.Join(contentDir, "work")
 		upperDir := filepath.Join(contentDir, "upper")
+
+		if opts != nil && opts.Workdir != "" && opts.Upperdir != "" {
```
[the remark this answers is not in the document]
flouthoc: @mtrmac Afaik both must be specified we already have some checks for this at `parse` level as well.

pkg/overlay/overlay.go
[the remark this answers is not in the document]
mtrmac: If “user” is the human running Podman (and not e.g. contents of a `Containerfile`), that looks fine WRT security, then — it gives the user no more capabilities. Validation still wouldn’t _hurt_, to get a good error message, if nothing else, but might not be essential.


WRT correctness, if it is possible to successfully tell the kernel to use paths that contain `,`, should that be possible using this API? Compare https://unix.stackexchange.com/questions/552630/how-to-escape-comma-on-mount-options-for-overlay . Now there are three ways to go about this (also looking at https://github.com/containers/podman/pull/12712 ):
1. `Workdir` really is a _directory_. The caller can pass `/tmp/a,b`, and it will work, because this code is responsible for escaping. (OTOH if the input in Podman is mount-options-like, then the Podman caller might have to unescape before passing it to this package!)
2. `Workdir` is “a fragment of an overlay mount option”, and the caller is responsible for escaping. (Then the Podman mount-options-like input doesn’t need to be unescaped — _as long as_ Podman is at least separating the individual options correctly)
3. This is too much work to think about, let’s just reject all such input (but we still need to decide what “such input” means; comma? colon? backslash?) for now, and define it later.

Doing nothing now basically commits us to approach 2., because adding unescaping or rejecting some inputs could break callers that start relying on the semantics of 2.

pkg/overlay/overlay.go
flouthoc: @mtrmac `MountWithOptions` automatically sets these but I agree this is making things confusing for reader. Let me remove `mountHelper` that should simplify things.  https://github.com/containers/buildah/pull/3715/files#diff-97a773bb4375bb0de346986ec764abdbc2f07b74a298953de1973eb747ed9759R84

pkg/overlay/overlay.go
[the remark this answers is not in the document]
mtrmac: OK. Also https://github.com/containers/podman/pull/12712#discussion_r777366532 confirms that.

(We could always relax this later.)

pkg/overlay/overlay.go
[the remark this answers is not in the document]
mtrmac: Also, if it is 2, the `Workdir` documentation should point out this non-obvious caller responsibility. (I wouldn’t even object to renaming the field to `WorkdirOptionFragment` or something similarly explicit.)

pkg/overlay/overlay.go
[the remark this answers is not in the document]
flouthoc: Fixed this.

pkg/overlay/overlay.go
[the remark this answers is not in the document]
flouthoc: Made them options and `mountHelper` is now removed.

pkg/overlay/overlay.go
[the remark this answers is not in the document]
flouthoc: @mtrmac I have updated the PR with more inclination towards options `2` with explicit note since I am still unsure what to escape here

pkg/overlay/overlay.go
flouthoc: This is fixed in latest commit.

pkg/overlay/overlay.go
```diff
@@ -65,24 +97,31 @@ func generateOverlayStructure(containerDir string, rootUID, rootGID int) (string
...
+	overlayOpts := Options{GraphOpts: graphOptions, ReadOnly: true, RootUID: rootUID, RootGID: rootGID}
+	return MountWithOptions(contentDir, source, dest, &overlayOpts)
 }
 
+// MountWithOptions creates a subdir of the contentDir based on the source directory
+// from the source system.  It then mounts up the source directory on to the
+// generated mount point and returns the mount point to the caller.
+// But allows api to set custom workdir, upperdir and other overlay options
+// Following API is being used by podman at the moment
 // NOTE: rootUID and rootUID are not yet used.
```
mtrmac, line 118: (This note can probably be dropped here, it would have to talk about `opts.RootUID`, and that is already documented.)

pkg/overlay/overlay.go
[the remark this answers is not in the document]
mtrmac: I’m not even sure we need the option values. As they are documented now, I don’t know why anyone would set them or what values would they pass.

But, sure, I’m happy enough with them gone from this prototype.

mtrmac (commented):
Code LGTM, just documentation nits.

---

> [NO NEW TESTS NEEDED]

Is this appropriate here?

flouthoc:
> Code LGTM, just documentation nits.
> 
> > [NO NEW TESTS NEEDED]
> 
> Is this appropriate here?

Updated Docs.

>  Please elaborate on the use case, and add tests.

@nalind @mtrmac Following API is only used by podman and its much easier to test there -> https://github.com/containers/podman/pull/12712 . We don't have flat tests for `overlay` in buildah hence i'm not entirely sure how to rig a test for this on buildah repo.

But if it does not breaks any existing buildah tests then should this be fine ?

TomSweeneyRedHat:
Changes LGTM
Would like a final head nod from @nalind and @mtrmac , especially given the test status

flouthoc:
@TomSweeneyRedHat I am also extending feature to buildah so we can test here as well.

run_linux.go
```diff
@@ -1836,7 +1855,9 @@ func (b *Builder) runSetupVolumeMounts(mountLabel string, volumeMounts []string,
 				return specs.Mount{}, errors.Wrapf(err, "failed to create TempDir in the %s directory", containerDir)
 			}
 
-			overlayMount, err := overlay.Mount(contentDir, host, container, rootUID, rootGID, b.store.GraphOptions())
+			overlayOpts := overlay.Options{RootUID: rootUID, RootGID: rootGID, UpperDirOptionFragment: upperDir, WorkDirOptionFragment: workDir, GraphOpts: b.store.GraphOptions()}
```
mtrmac, line 1858: Purely aesthetic: Maybe use one value per line here, to make it more readable?

(My personal preference would be to embed the value that into the `MountWithOptions` call, but that’s entirely up to the maintainers of this repo.)

run_linux.go
```diff
@@ -1806,6 +1806,20 @@ func (b *Builder) runSetupVolumeMounts(mountLabel string, volumeMounts []string,
 			case "private", "rprivate", "slave", "rslave", "shared", "rshared":
 				rootProp = opt
 			}
+
+			if strings.Contains(opt, "upperdir") {
```
mtrmac, line 1810: Shouldn’t this be `HasPrefix`? A check for `HasPrefix(…, "upperdir=")` could even avoid the need for the other check.

(One could over-engineer this and parse every entry into a key[=value], and then do the switch purely on key values. Or maybe the switch should be `case opt == "rw":… case strings.HasPrefix(opt, "upperdir")`. Up to the maintainers of the repo.)

run_linux.go
```diff
@@ -1806,6 +1806,20 @@ func (b *Builder) runSetupVolumeMounts(mountLabel string, volumeMounts []string,
...
+				splitOpt := strings.SplitN(opt, "=", 2)
+				if len(splitOpt) > 1 {
+					upperDir = splitOpt[1]
+				}
+			}
+			if strings.Contains(opt, "workdir") {
```
mtrmac, line 1816: `HasPrefix`

mtrmac (commented):
Thanks for the tests!

Note that I’m just reading the diff.. I don’t know anything about the testing infrastructure, and I’ll defer to repo owners on the introduced UI.

run_linux.go
flouthoc: fixed

run_linux.go
flouthoc: Swtiched to hasprefix.

run_linux.go
```diff
@@ -1806,6 +1806,20 @@ func (b *Builder) runSetupVolumeMounts(mountLabel string, volumeMounts []string,
...
+					upperDir = splitOpt[1]
+				}
+			}
+			if strings.HasPrefix(opt, "workdir") {
+				splitOpt := strings.SplitN(opt, "=", 2)
+				if len(splitOpt) > 1 {
+					workDir = splitOpt[1]
+				}
+			}
+
```
nalind, line 1822: Blank line at end of block.

run_linux.go
```diff
@@ -1826,6 +1840,11 @@ func (b *Builder) runSetupVolumeMounts(mountLabel string, volumeMounts []string,
 			}
 		}
 		if foundO {
+
```
nalind, line 1843: Blank line at beginning of block.

run_linux.go
```diff
@@ -1836,7 +1855,13 @@ func (b *Builder) runSetupVolumeMounts(mountLabel string, volumeMounts []string,
 				return specs.Mount{}, errors.Wrapf(err, "failed to create TempDir in the %s directory", containerDir)
 			}
 
-			overlayMount, err := overlay.Mount(contentDir, host, container, rootUID, rootGID, b.store.GraphOptions())
+			overlayOpts := overlay.Options{RootUID: rootUID,
+				RootGID:                rootGID,
+				UpperDirOptionFragment: upperDir,
+				WorkDirOptionFragment:  workDir,
+				GraphOpts:              b.store.GraphOptions()}
```
nalind, line 1862: Add a `,` and a newline after `b.store.GraphOptions()`.

tests/run.bats
```diff
@@ -287,6 +287,36 @@ function configure_and_check_user() {
...
+	cid=$output
+	mkdir -p ${TESTDIR}/upperdir
+	mkdir -p ${TESTDIR}/workdir
+	mkdir -p ${TESTDIR}/lower
+
+	echo 'hello' >> ${TESTDIR}/lower/hello
+
+	# As a baseline, this should succeed.
+	run_buildah run -v ${TESTDIR}/lower:/test:O,upperdir=${TESTDIR}/upperdir,workdir=${TESTDIR}/workdir${zflag:+:${zflag}}  $cid cat /test/hello
+  	expect_output "hello"
```
nalind, line 311: Nit: there are a couple of spaces in front of this tab character.

tests/run.bats
```diff
@@ -287,6 +287,36 @@ function configure_and_check_user() {
...
+	run_buildah run -v ${TESTDIR}/lower:/test:O,upperdir=${TESTDIR}/upperdir,workdir=${TESTDIR}/workdir${zflag:+:${zflag}}  $cid sh -c 'echo "world" > /test/world'
+
+	#upper dir should persist content
+	result="$(cat ${TESTDIR}/upperdir/world)"
+	test "$result" == "world"
+
```
nalind, line 317: Blank line at end of block.

pkg/overlay/overlay.go
```diff
@@ -18,6 +18,40 @@ import (
...
+	UpperDirOptionFragment string
+	// The Workdir is used to prepare files as they are switched between the layers.
+	// Note!! : Following API does not handles escaping or validates correctness of the values
+	// passed to WorkDirOptionFragment instead API will try to pass values as is it
+	// to the `mount` command. Its user's responsibility to make sure they pre-validate
+	// these values. Invalid inputs may lead to undefined behviour.
+	// This is provided as-is, use it if it works for you, we can/will change/break that in the future.
+	// See discussion here for more context: https://github.com/containers/buildah/pull/3715#discussion_r786036959
+	// TODO: Should we address above comment and handle escaping of metacharacters like
+	// `comma`, `backslash` ,`colon` and any other specical characters
```
nalind, line 43: Typo: specical

pkg/overlay/overlay.go
```diff
@@ -18,6 +18,40 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+// Overlay type holds various configuration options for overlay
```
nalind, line 21: Comment describes the "Options" type.

pkg/overlay/overlay.go
```diff
@@ -18,6 +18,40 @@ import (
...
+// MountWithOptions accepts following type so its easier to specify
```
nalind, line 22: Typo: its

pkg/overlay/overlay.go
```diff
@@ -18,6 +18,40 @@ import (
...
+// more verbose configuration for overlay mount.
+type Options struct {
+	// The Upper directory is normally writable layer in an overlay mount.
+	// Note!! : Following API does not handles escaping or validates correctness of the values
+	// passed to UpperDirOptionFragment instead API will try to pass values as is it
+	// to the `mount` command. Its user's responsibility to make sure they pre-validate
```
nalind, line 28: Typo: Its

pkg/overlay/overlay.go
```diff
@@ -18,6 +18,40 @@ import (
...
+	// these values. Invalid inputs may lead to undefined behviour.
+	// This is provided as-is, use it if it works for you, we can/will change/break that in the future.
+	// See discussion here for more context: https://github.com/containers/buildah/pull/3715#discussion_r786036959
+	// TODO: Should we address above comment and handle escaping of metacharacters like
+	// `comma`, `backslash` ,`colon` and any other specical characters
```
nalind, line 33: Typo: specical

pkg/overlay/overlay.go
```diff
@@ -18,6 +18,40 @@ import (
...
+	UpperDirOptionFragment string
```
nalind, line 34: Nit: the `OptionFragment` suffix sounds redundant.
Should probably note that if this field is set, `WorkDirOptionFragment` should also be set.

pkg/overlay/overlay.go
```diff
@@ -18,6 +18,40 @@ import (
...
+	// The Workdir is used to prepare files as they are switched between the layers.
+	// Note!! : Following API does not handles escaping or validates correctness of the values
+	// passed to WorkDirOptionFragment instead API will try to pass values as is it
+	// to the `mount` command. Its user's responsibility to make sure they pre-validate
+	// these values. Invalid inputs may lead to undefined behviour.
+	// This is provided as-is, use it if it works for you, we can/will change/break that in the future.
+	// See discussion here for more context: https://github.com/containers/buildah/pull/3715#discussion_r786036959
+	// TODO: Should we address above comment and handle escaping of metacharacters like
+	// `comma`, `backslash` ,`colon` and any other specical characters
+	WorkDirOptionFragment string
```
nalind, line 44: Nit: the `OptionFragment` suffix sounds redundant.  Should probably note that if this field is set, `UpperDirOptionFragment` should also be set.

nalind (commented):
Code looks reasonable.  Some nits about field names and some typos.

pkg/overlay/overlay.go
flouthoc: @nalind I was decided to name both of them as fragment as they are still just fragments to overlay's options as discussed here https://github.com/containers/buildah/pull/3715#discussion_r786734076 . I named both of them in same manner as it describes the similar nature of field to the reader of the API

pkg/overlay/overlay.go
flouthoc: Reasoning for this is same as https://github.com/containers/buildah/pull/3715#discussion_r788917437

pkg/overlay/overlay.go
mtrmac: @nalind See that discussion, but I’m completely happy to leave naming/API design to you.

PR #3720, opened 2022-01-19
buildah build --network add support for custom networks 

Luap99:
#### What type of PR is this?

 /kind feature

The backend logic already supports specifying custom network names. This
only adds the support for the frontend parsing.

#### How to verify it

buildah bud --network name1,name2 ...

Fixes containers/podman#12282

```release-note

```

### DIFF
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -389,7 +389,8 @@ Valid _mode_ values are:
 - **none**: no networking;
 - **host**: use the host network stack. Note: the host mode gives the container full access to local system services such as D-bus and is therefore considered insecure;
 - **ns:**_path_: path to a network namespace to join;
-- `private`: create a new namespace for the container (default)
+- **private**: create a new namespace for the container (default)
+- **\<network name|ID\>**: Join the network with the given name or ID, e.g. use `--network mynet` to join the network with the name mynet. Only supported for rootful users.
 
 **--no-cache**
 
@@ -797,6 +798,8 @@ buildah build --no-cache --rm=false -t imageName .
 buildah build --dns-search=example.com --dns=223.5.5.5 --dns-option=use-vc .
 
 buildah build -f Containerfile.in -t imageName .
+
+buildah build --network mynet .
 
 ### Building an multi-architecture image using the --manifest option (requires emulation software)
 
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -2823,9 +2823,6 @@ EOF
   run_buildah build --network=container --signature-policy ${TESTSDIR}/policy.json -t ${target} ${TESTSDIR}/bud/containerfile
   [ "${status}" -eq 0 ]
   expect_output --substring "FROM alpine"
-
-  run_buildah 125 build --network=bogus --signature-policy ${TESTSDIR}/policy.json -t ${target} ${TESTSDIR}/bud/containerfile
-
 }
 
 @test "bud-replace-from-in-containerfile" {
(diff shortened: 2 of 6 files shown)


docs/buildah-build.1.md
```diff
@@ -389,7 +389,8 @@ Valid _mode_ values are:
 - **none**: no networking;
 - **host**: use the host network stack. Note: the host mode gives the container full access to local system services such as D-bus and is therefore considered insecure;
 - **ns:**_path_: path to a network namespace to join;
-- `private`: create a new namespace for the container (default)
+- **private**: create a new namespace for the container (default)
+- **\<network name|ID\>**: Join the network with the given name or ID. Only supported for rootful users.
```
flouthoc, line 393: Example for this would be nice, is this `--network <networkname>`.

flouthoc:
Renaming the PR to `buildah build` instead of `buildah bud`.

docs/buildah-build.1.md
Luap99: Do you mean in the description or under the example section at the bottom of the page?

flouthoc: Yes but its not a blocker for me. I figured cli from `tests`

docs/buildah-build.1.md
```diff
@@ -798,6 +799,8 @@ buildah build --dns-search=example.com --dns=223.5.5.5 --dns-option=use-vc .
 
 buildah build -f Containerfile.in -t imageName .
 
+buildah build --network mynet .
+
```
TomSweeneyRedHat, line 803: We'll eventually need similar changes to the man page for `podman build` if not already there.

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
@Luap99 tests are very unhappy and you might need to rebase again.

rhatdan:
@Luap99 Once this gets in, do you think you are done with Buildah?  IE Can we cut a release of Buildah to complete vendoring into Podman.

Luap99:
> @Luap99 Once this gets in, do you think you are done with Buildah? IE Can we cut a release of Buildah to complete vendoring into Podman.

Yes, but it is possible that we have to change c/common for more bugfixes.

TomSweeneyRedHat:
I'd like to plan to start building the next Buildah version on Wed Jan 26, 2022.  @flouthoc @nalind holler if you've things you'd like to make sure get in before we cut that.

rhatdan:
Sure we can revendor in containers-common as we go through the release.

rhatdan:
I think Podman 4.0 RC3 should be next Friday, so Wednesday sounds good.  We only have three PRs that I see getting in, right now.

tests/bud.bats
```diff
@@ -3902,3 +3899,16 @@ _EOF
...
+
+@test "bud with network names" {
+  skip_if_no_runtime
+
+  _prefetch alpine
+
+  run_buildah 125 bud --signature-policy ${TESTSDIR}/policy.json --network nontexists ${TESTSDIR}/bud/network
+  expect_output --substring "network not found"
+
+  run_buildah bud --signature-policy ${TESTSDIR}/policy.json --network podman ${TESTSDIR}/bud/network
```
nalind, line 3911: Where does the configuration for the network with this name come from?

Luap99: podman is the default network name (can be changed in cotnainers.conf). If it does not exists on disk the default will always be created in memory, this happens in the libnetwork backend.

Luap99:
OK the problem with the in container test is that the isolation is set to `chroot` where the network namespace is not supported. I added a check to throw an error in this case

TomSweeneyRedHat:
Happy green test buttons, but I'm not sure if this needs a rebase or not.

PR #3724, opened 2022-01-20
linux: fix lookup for runtime

giuseppe:
lookup the full runtime path instead of using its name.

Closes: https://github.com/containers/buildah/issues/3721


flouthoc (approved):

giuseppe:
all green

/hold cancel

PR #3731, opened 2022-01-21
run: if `--workingdir` is relative treat it as relative to image's workdir

flouthoc:
When users are specifying `relative` path with `--workingdir` while
performing `buildah run` then buildah should honor relative path by
treating it as relative to image's workdir instead of assuming it as
relative to `root`.

Closes: https://github.com/containers/buildah/issues/3729

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -119,7 +119,12 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	}
 
 	if options.WorkingDir != "" {
-		g.SetProcessCwd(options.WorkingDir)
+		if filepath.IsAbs(options.WorkingDir) {
+			g.SetProcessCwd(options.WorkingDir)
+		} else {
+			// assume path is realtive to current working dir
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
@flouthoc how does Docker handle this.  I'm always very leery about relative paths as it's easy to do something like `../../../etc` and then overwrite files you should not.

flouthoc:
@TomSweeneyRedHat That's a fair request I'll verify it how docker is doing. I just thought this would be a cool feature but I also agree we cannot do a behavior drift so I'll confirm with docker once.

TomSweeneyRedHat:
@flouthoc I think you said in the issue that this would cause a drift with Docker.  If so, can we close this?

flouthoc:
@TomSweeneyRedHat Sure. But behavior would not be different its just that docker does not supports this. I'll re-open PR once we have consensus on the issue.

TomSweeneyRedHat:
My thought is it is a potential security issue and we should not pursue it.

PR #3735, opened 2022-01-22
Make pull commands be consistent

TomSweeneyRedHat:
Per @edsantiago 's suggestion, make the pull commands consistent, always
accepting a value.  Currently we have:

--pull
--pull=true
--pull=false
--pull-never
--pull-always

With this changes, we will only have pull with a variety of options,
ala:

--pull
--pull=true
--pull=false
--pull=never
--pull=always

For backward compatibility, the --pull-never and --pull-always
options will remain operational, however they are not documented
and are conisdered deprecated.

Signed-off-by: tomsweeneyredhat <tsweeney@redhat.com>

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/cmd/buildah/build.go
+++ b/cmd/buildah/build.go
@@ -124,14 +124,17 @@ func buildCmd(c *cobra.Command, inputArgs []string, iopts buildOptions) error {
 		defer os.Remove(iopts.BudResults.Authfile)
 	}
 
+	// Allow for --pull, --pull=true, --pull=false, --pull=never, --pull=always
+	// --pull-always and --pull-never.  The --pull-never and --pull-always options
+	// will not be documented.
 	pullPolicy := define.PullIfMissing
-	if iopts.Pull {
+	if strings.EqualFold(strings.TrimSpace(iopts.Pull), "true") {
 		pullPolicy = define.PullIfNewer
 	}
-	if iopts.PullAlways {
+	if iopts.PullAlways || strings.EqualFold(strings.TrimSpace(iopts.Pull), "always") {
 		pullPolicy = define.PullAlways
 	}
-	if iopts.PullNever {
+	if iopts.PullNever || strings.EqualFold(strings.TrimSpace(iopts.Pull), "never") {
 		pullPolicy = define.PullNever
 	}
 	logrus.Debugf("Pull Policy for pull [%v]", pullPolicy)
--- a/cmd/buildah/from.go
+++ b/cmd/buildah/from.go
@@ -25,7 +25,7 @@ type fromReply struct {
 	creds           string
 	format          string
 	name            string
-	pull            bool
+	pull            string
 	pullAlways      bool
 	pullNever       bool
 	quiet           bool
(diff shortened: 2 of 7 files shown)


TomSweeneyRedHat:
If we merge this before we build the next release for Podman v4.0, I'll need to at least change the documentation in Podman too.  @mheon, FYI

rhatdan:
You need to mark --pull-always and --pull-never as hidden so that we don't see them in --help.

 ./bin/buildah build --help  | grep pull
      --pull string                   pull the image from the registry if newer or not present in store, if false, only pull the image if not present, if always, pull the image even if the named image is present in store, if never, only use the image present in store if available (default "true")
      --pull-always                   pull the image even if the named image is present in store
      --pull-never                    do not pull the image, use the image present in store if available

TomSweeneyRedHat:
@Luap99 Cobra question for you, or maybe @bbaude.  I switched the `--pull` option from a `BoolVar` to `StrVar` for the build and from commands.  I set the default value for pull to "true".  If I do:

```
buildah bud --pull --quiet .
```
The pull flag gets set to `--quiet` rather than the default `true`.  The quiet option is gobbled by the pull and is ignored and the pull doesn't get set appropriately.

If I add an equal sign after the pull, ala:

```
buildah bud --pull= --quiet .
```
Then pull gets set to `true`,  the quiet flag is not eaten, the build builds quietly, and the pull works as expected.

I've looked all over the limited Cobra doc to get this to work, and maybe my eyelids are just too heavy and I'm missing it at this point, but any tips/tricks for this?  In a perfect world, I'd have the `--pull ` and `--pull= ` work the same way and go to the default value.

Luap99:
You have to set `flag.NoOptDefVal = "true"`

rhatdan:
LGTM once you add @Luap99 line.

TomSweeneyRedHat:
@Luap99 TYVM for the pointer, looks like:

```
        flags.Lookup("pull").NoOptDefVal = "true"
```
turns the trick.  As a bonus, I finally located  the docs for the Cobra flags.  Off to test a bit more, then I'll resubmit.

TomSweeneyRedHat:
OK, this looks happy in my testing now.  Hopefully, the CI will concur.

TomSweeneyRedHat:
One flake, one real error, I think I've the real one figured out.  Pushing again soon.

nalind:
> @TomSweeneyRedHat This is the time to get this done. Since it is a somewhat breaking change (Not really but a change in UI).

It's an API break in `pkg/cli`.

rhatdan:
No the Old CLI is still handled.

nalind:
Consumers that imported `pkg/cli` and expected `BudResults.Pull` to be a `bool` will fail to compile if they vendor the new version without patching their own source code.  If we didn't break compatibility for the CLI, that's great, but it's not the same thing.

rhatdan:
Do you want to bump the major version of Buildah?

PR #3743, opened 2022-01-24
docs/buildah-build.1.md: don't imply that -v isn't just a RUN thing

nalind:
#### What type of PR is this?

/kind documentation

The current wording doesn't clearly state that volumes specified using -v are only a factor when executing RUN instructions.

None

```release-note
None
```

### DIFF
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -641,9 +641,9 @@ Set the architecture variant of the image to be pulled.
 
 **--volume**, **-v**[=*[HOST-DIR:CONTAINER-DIR[:OPTIONS]]*]
 
-   Create a bind mount. If you specify, ` -v /HOST-DIR:/CONTAINER-DIR`, Buildah
-   bind mounts `/HOST-DIR` in the host to `/CONTAINER-DIR` in the Buildah
-   container. The `OPTIONS` are a comma delimited list and can be: <sup>[[1]](#Footnote1)</sup>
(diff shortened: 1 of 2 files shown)


docs/buildah-build.1.md
```diff
@@ -641,9 +641,9 @@ Set the architecture variant of the image to be pulled.
 
 **--volume**, **-v**[=*[HOST-DIR:CONTAINER-DIR[:OPTIONS]]*]
 
-   Create a bind mount. If you specify, ` -v /HOST-DIR:/CONTAINER-DIR`, Buildah
-   bind mounts `/HOST-DIR` in the host to `/CONTAINER-DIR` in the Buildah
-   container. The `OPTIONS` are a comma delimited list and can be: <sup>[[1]](#Footnote1)</sup>
+Mount a host directory into containers when executing *RUN* instructions during
+the build.  The `OPTIONS` are a comma delimited list and can be:
+<sup>[[1]](#Footnote1)</sup>
```
TomSweeneyRedHat, line 646: Any reason to remove the second sentence?  I'd prefer to keep it.

TomSweeneyRedHat:
re-ran f35-overlay test which looked to be flakey.

docs/buildah-build.1.md
nalind: It doesn't appear to provide any information that isn't provided by the combination of the summary of the option and the previous sentence, so it's just more words that people have to get through when reading the docs.

Issue #3281, opened 2021-06-03
Builds in docker fail with 'Invalid argument' when doing RUN after VOLUME on overlayfs


mh21:
When running buildah in GitLab CI (docker), using RUN after VOLUME fails since buildah 1.20.1.

Dockerfile:

```plain
FROM registry.fedoraproject.org/fedora:33

RUN mkdir -p /logs
VOLUME /logs

RUN mkdir -p /data

CMD ["/bin/bash"]
```

Reproducer: https://gitlab.com/cki-project/experimental/buildah-1-20-1-fc33-volume-issue-reproducer

Works/fails (https://gitlab.com/cki-project/experimental/buildah-1-20-1-fc33-volume-issue-reproducer/-/pipelines/314126067):
- quay.io/buildah/stable:v1.19.6: OK
- quay.io/buildah/stable:v1.19.8: OK
- quay.io/buildah/stable:v1.20.1: FAILS
- quay.io/buildah/stable:v1.21.0: OK
- registry.fedoraproject.org/fedora:33 with 1.20.1: FAILS
- registry.fedoraproject.org/fedora:34 with 1.21.0: FAILS

The important difference between the 1.21.0 quay.io image and the plain fedora:34 seems to be the /var/lib/containers volume (https://quay.io/repository/buildah/stable/manifest/sha256:da34723916c0b53186230af839ff339c6a4f4857ca54c1984e6e927fb4100478). This causes /var/lib/containers to be backed by ext4 (GitLab/docker) instead of the root overlayfs.

Without the volume, the error message is (registry.fedoraproject.org/fedora:34 with 1.21.0, from https://gitlab.com/cki-project/experimental/buildah-1-20-1-fc33-volume-issue-reproducer/-/jobs/1314718416#L331):

```plain
$ buildah bud -f Dockerfile .
STEP 1: FROM registry.fedoraproject.org/fedora:33
Trying to pull registry.fedoraproject.org/fedora:33...
Getting image source signatures
Copying blob sha256:2df8ffe0bf4632dc6b3c749a45145c30ba9d92f47853ce3769c94e7c476c1fcc
Copying config sha256:5ca3a10c6c8f69d82c1e7f0cf61e3aef2967874a1a7dc10660a11517a1a9d31a
Writing manifest to image destination
Storing signatures
STEP 2: RUN mkdir -p /logs
STEP 3: VOLUME /logs
STEP 4: RUN mkdir -p /data
[...]

PR #3750, opened 2022-01-26
overlay: always honor mountProgram

giuseppe:
if a mountProgram is specified, use it also in rootfull mode.

Closes: https://github.com/containers/buildah/issues/3281

### DIFF
--- a/pkg/overlay/overlay.go
+++ b/pkg/overlay/overlay.go
@@ -188,20 +188,21 @@ func MountWithOptions(contentDir, source, dest string, opts *Options) (mount spe
 		overlayOptions = fmt.Sprintf("lowerdir=%s,upperdir=%s,workdir=%s,private", escapeColon(source), upperDir, workDir)
 	}
 
-	if unshare.IsRootless() {
-		mountProgram := findMountProgram(opts.GraphOpts)
-		if mountProgram != "" {
-			if err := mountWithMountProgram(mountProgram, overlayOptions, mergeDir); err != nil {
-				return mount, err
-			}
-
-			mount.Source = mergeDir
-			mount.Destination = dest
-			mount.Type = "bind"
-			mount.Options = []string{"bind", "slave"}
-			return mount, nil
+	mountProgram := findMountProgram(opts.GraphOpts)
+	if mountProgram != "" {
+		if err := mountWithMountProgram(mountProgram, overlayOptions, mergeDir); err != nil {
+			return mount, err
 		}
-		/* If a mount_program is not specified, fallback to try mount native overlay.  */
+
+		mount.Source = mergeDir
+		mount.Destination = dest
+		mount.Type = "bind"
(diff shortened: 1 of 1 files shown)


nalind:
We wouldn't need to export the two new functions in the overlay package if `setupChrootBindMounts()` was changed to call either `overlay.Mount()` or `overlay.MountWithOptions()`.

giuseppe:
> `overlay.MountWithOptions()`.

the issue with that is that it always creates a subdirectory `merge`.  I could try with a new option to skip it creating the subdirectory?

I am fine with such change, but I'll need more time for it.  Probably this PR is not so important and it can wait for the next release without hurrying it up now.

giuseppe:
> the issue with that is that it always creates a subdirectory `merge`. I could try with a new option to skip it creating the subdirectory?

looking at it, it seems the new option would cause a conflict with Unmount, since it doesn't accept options and we won't know what subdirectory was used

flouthoc:
@giuseppe Would we need a `UnmountWithOptions` as well and probably it could replace generic `Unmount`. But that should be different PR I guess.

giuseppe:
I've dropped the patch that passes down the graph options to chroot, as it seems to not be needed to address the linked issue

pkg/overlay/overlay.go
```diff
@@ -154,41 +188,21 @@ func MountWithOptions(contentDir, source, dest string, opts *Options) (mount spe
...
 
-			mount.Source = mergeDir
-			mount.Destination = dest
-			mount.Type = "bind"
-			mount.Options = []string{"bind", "slave"}
-			return mount, nil
-		}
-		/* If a mount_program is not specified, fallback to try mount native overlay.  */
+	if unshare.IsRootless() {
+		/* If a mount_program is not specified, fallback to try mounting native overlay.  */
```
flouthoc, line 205: One doubt: We used to apply user's extended attributes for both `root` and `rootless` when we were falling back to `native overlay`. Does this check introduces a behavior change, I thought this is a `no-op` for root users.

giuseppe: previously it was also only set for rootless.

userxattr is useful only for rootless that cannot use the privileged overlay xattr

flouthoc: ah i see got it. My bad.

rhatdan:
LGTM
@flouthoc @nalind @umohnani8 @vrothberg @nalind PTAL

rhatdan:
@giuseppe is this something worth getting into podman 4.0?  Buildah 1.24.1?

flouthoc:
LGTM.

giuseppe:
> @giuseppe is this something worth getting into podman 4.0? Buildah 1.24.1?

I think it can wait for next release.  The issue happens only with containerized buildah not using a volume on `/var/lib/containers/storage`.

PR #3753, opened 2022-01-27
imagebuildah.BuildDockerfiles(): create the jobs semaphore

nalind:
#### What type of PR is this?

/kind bug

Make sure that, if we're given neither a concurrent jobs count nor a semaphore to use, we create a semaphore.  Otherwise, each platform in a multi-platform build will run with maximum concurrency, which isn't the desired default.  Our CLI always passes in a Jobs count, so it hasn't been affected by this problem.

#### How to verify it

The CLI always passes a number of jobs, so it doesn't trigger this.  It can be observed in programs that consume the library and set neither the `Jobs` nor the `JobSemaphore` field of a `define.BuildOptions`.  Podman, which consumes our CLI front-end, also won't stumble over it.

None

```release-note
None
```

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -176,8 +176,17 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 		files = append(files, b.Bytes())
 	}
 
-	if options.JobSemaphore == nil && options.Jobs != nil && *options.Jobs != 0 {
-		options.JobSemaphore = semaphore.NewWeighted(int64(*options.Jobs))
+	if options.JobSemaphore == nil {
+		if options.Jobs != nil {
+			if *options.Jobs < 0 {
+				return "", nil, errors.New("error building: invalid value for jobs.  It must be a positive integer")
+			}
+			if *options.Jobs > 0 {
+				options.JobSemaphore = semaphore.NewWeighted(int64(*options.Jobs))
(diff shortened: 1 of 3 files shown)


flouthoc:
LGTM. One doubt: @nalind We also had a default of of `jobs=1` if nothing is defined here: https://github.com/containers/buildah/blob/main/imagebuildah/executor.go#L190 was that a no-op for users who were consuming library

rhatdan:
This will now default to running the jobs based on the length of the stages.
LGTM
@nalind Answer @flouthoc question and then merge
/lgtm
/hold

nalind:
That bit of logic is run by every `Executor` that we create - if we're building for N platforms, we run it N times.  If the caller previously didn't set either `Jobs` or `JobSemaphore`, `BuildDockerfiles()` wouldn't create a semaphore, so each `Executor` would end up creating its own, and they would run independently in parallel, so the observed behavior looked more like jobs=N than the intended jobs=1.

nalind:
> This will now default to running the jobs based on the length of the stages.

Explicitly setting `Jobs` to 0 should, and previously, had this effect.  We document 0 being treated as "as many as possible".

nalind:
Rebased.

flouthoc:
@nalind makes sense thanks for explanation. LGTM

PR #3756, opened 2022-01-28
caps: fix buildah run --cap-add=all

### LINKED ISSUE #3755: error mapping capability "CAP_CHECKPOINT_RESTORE" to a number
**Description**

Hi there,

I'm receiving the following error when running `buildah run --cap-add all`.

>error setting capabilities for process: error mapping capability "CAP_CHECKPOINT_RESTORE" to a number

I note that `--cap-add  all` is not  documented in the buildah cmdline API, but it is supported internally.

This is the relevant system system configuration:
```
devcontainer:~# cat /proc/sys/kernel/cap_last_cap
40
devcontainer:~# uname -a
Linux devcontainer 5.15.0-2-amd64 #1 SMP Debian 5.15.5-1 (2021-11-26) x86_64 Linux
```

The value *40* is mapped to *CAP_CHECKPOINT_RESTORE*. See https://github.com/syndtr/gocapability/blob/42c35b4376354fd554efc7ad35e0b7f94e3a0ffb/capability/enum.go#L301

**Analysis**

The error is because **capability.CAP_LAST_CAP**  is set to the value of `/proc/sys/kernel/cap_last_cap`, which is the last valid and supported capability. But it is used as value for checking if the capability is undefined here 

https://github.com/containers/buildah/blob/7f81b68d58019d3137a399a7d521608a7537c00d/chroot/run.go#L907-L916
and here

https://github.com/containers/buildah/blob/7f81b68d58019d3137a399a7d521608a7537c00d/chroot/run.go#L920-L929

r10r:
Fixes #3755


rhatdan:
Please add a test or add [NO NEW TESTS ADDED]

LGTM

rhatdan:
Could you check in Podman if we have the same problem.

r10r:
I'll take a look at tests and podman. 
Btw. can I update the **--cap-add** documentation as well and mention the **all** value or should it not be promoted and stay a hidden feature ?

rhatdan:
Yes update the documentation to add "add" value.

TomSweeneyRedHat:
LGTM once docs and tests are added. TYVM @r10r !

rhatdan:
@r10r I took this over and added the no new test flag, so we can get it into a release.

r10r:
@rhatdan  Nice. Are tests required anyway ? I can then create a separate pull request.
I'm just about to look at the test suite because I'll create another pull request in a minute for which tests seem to be mandatory;)

r10r:
@rhatdan Do you take care of updating the documentation and looking at podman as well ?

rhatdan:
No if you want to open a separate PR for those it would be great.  I am just trying to get the release out.

PR #3757, opened 2022-01-31
stage_executor: Add support for inline `FROM --platform=` within Containerfile/Dockerfile

flouthoc:
See: https://docs.docker.com/engine/reference/builder/#from

Allows end users to configure executor's `OS`, `ARCH`,`VARIANT` via inline `--platform`.

Usage    
```
 FROM --platform=linux/arm64 alpine
 RUN uname -a
 ```
    
 ------
    
While this allows executor to pull base images with custom
`OS`, `ARCH`, `VARIANT` it still allows end-users to tag images with
 different format if they need to.
    
For example
    
Doing a `buildah build --platform linux/amd64 -t test .` on below containerfile.
    
 ```
 FROM --platform=linux/arm64 alpine
 RUN uname -a
 ```
    
Would result in tagging image as `linux/amd64` while `base` layer would
 be pulled for `linux/arm64`.
 
 Closes: https://github.com/containers/buildah/issues/3712
 Closes: https://github.com/containers/buildah/issues/3261

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -16,6 +16,7 @@ import (
 	"github.com/containers/buildah/define"
 	buildahdocker "github.com/containers/buildah/docker"
 	"github.com/containers/buildah/internal"
+	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/buildah/pkg/rusage"
 	"github.com/containers/buildah/util"
 	cp "github.com/containers/image/v5/copy"
@@ -595,6 +596,22 @@ func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBCo
 		}
 	}
 
+	builderSystemContext := s.executor.systemContext
+	// get platform string from stage
+	if stage.Builder.Platform != "" {
+		os, arch, variant, err := parse.Platform(stage.Builder.Platform)
+		if err != nil {
+			return nil, errors.Wrapf(err, "unable to parse platform %q", stage.Builder.Platform)
+		}
+		if arch != "" || variant != "" {
+			builderSystemContext.ArchitectureChoice = arch
+			builderSystemContext.VariantChoice = variant
+		}
+		if os != "" {
+			builderSystemContext.OSChoice = os
+		}
+	}
+
 	builderOptions := buildah.BuilderOptions{
 		Args:                  ib.Args,
 		FromImage:             from,
@@ -604,7 +621,7 @@ func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBCo
 		BlobDirectory:         s.executor.blobDirectory,
 		SignaturePolicyPath:   s.executor.signaturePolicyPath,
 		ReportWriter:          s.executor.reportWriter,
-		SystemContext:         s.executor.systemContext,
+		SystemContext:         builderSystemContext,
 		Isolation:             s.executor.isolation,
 		NamespaceOptions:      s.executor.namespaceOptions,
 		ConfigureNetwork:      s.executor.configureNetwork,
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -117,6 +117,60 @@ symlink(subdir)"
   expect_output --substring $(realpath "${TESTSDIR}/bud/dockerignore3/.dockerignore")
 }
 
+# Following test must fail since we are trying to run linux/arm64 on linux/amd64
+# Issue: https://github.com/containers/buildah/issues/3712
+@test "build-with-inline-platform" {
+  # Host arch
+  mkdir -p ${TESTDIR}/bud/platform
+  run_buildah info --format '{{.host.arch}}'
+  myarch="$output"
+  otherarch="arm64"
+
+  # just make sure that other arch is not equivalent to host arch
+  if [[ "$otherarch" == "$myarch" ]]; then
+    otherarch="amd64"
+  fi
+  # ...create a Containerfile with --platform=linux/$otherarch
+  cat > ${TESTDIR}/bud/platform/Dockerfile << _EOF
+FROM --platform=linux/${otherarch} alpine
+RUN uname -m
+_EOF
+
+  run_buildah '?' build --signature-policy ${TESTSDIR}/policy.json -t test ${TESTDIR}/bud/platform
+  if [[ $status -eq 0 ]]; then
+    run_buildah inspect --format '{{ .OCIv1.Architecture }}' test
+    expect_output --substring "$otherarch"
+  else
+    # Build failed: we DO NOT have qemu-user-static installed.
+    expect_output --substring "format error"
+  fi
+}
+
+# Following test must pass since we want to tag image as host arch
+# Test for use-case described here: https://github.com/containers/buildah/issues/3261
+@test "build-with-inline-platform-amd-but-tag-as-arm" {
+  # Host arch
+  mkdir -p ${TESTDIR}/bud/platform
+  run_buildah info --format '{{.host.arch}}'
+  myarch="$output"
+  targetarch="arm64"
+
+  if [[ "$targetArch" == "$myarch" ]]; then
+    targetarch="amd64"
+  fi
+
+  cat > ${TESTDIR}/bud/platform/Dockerfile << _EOF
+FROM --platform=linux/${myarch} alpine
+RUN uname -m
+_EOF
+
+  # Tries building image where baseImage has --platform=linux/HostArch
+  run_buildah build --platform linux/${targetarch} --signature-policy ${TESTSDIR}/policy.json -t test ${TESTDIR}/bud/platform
+  run_buildah inspect --format '{{ .OCIv1.Architecture }}' test
+  # base image is pulled as HostArch but tagged as non host arch
+  expect_output --substring $targetarch
+}
+
 @test "bud-flags-order-verification" {
   run_buildah 125 build /tmp/tmpdockerfile/ -t blabla
   check_options_flag_err "-t"

### CODE REVIEW

**Comment by flouthoc** [comment]
Please don't review this PR yet as this PR is a implementation of  : https://github.com/openshift/imagebuilder/pull/212 so lets wait for that.

**Comment by flouthoc** [comment]
@nalind @rhatdan PTAL. Also modified tests to also check if `qemu-user-static` is already there on platform as @edsantiago suggested in the last PR.

**Comment by flouthoc** [comment]
This PR closes two existing open issues.

**Comment by rhatdan** [comment]
LGTM
@nalind @vrothberg @TomSweeneyRedHat @ashley-cui @umohnani8 PTAL
 
As soon as this is merged, we can cut a new release.

**Review by rhatdan — APPROVED** [review]

File: tests/bud.bats [hunk 1]
```diff
@@ -117,6 +117,48 @@ symlink(subdir)"
...
+    # Build failed: we DO NOT have qemu-user-static installed.
+    expect_output --substring "format error"
+  fi
+}
+
+# Following test must pass since we want to tag image as arm64
+# Test for use-case described here: https://github.com/containers/buildah/issues/3261
+@test "build-with-inline-platform-amd-but-tag-as-arm" {
+  # Tries building image where baseImage has --platform=linux/arm64
+  run_buildah build --platform linux/arm64 --signature-policy ${TESTSDIR}/policy.json -t test -f ${TESTSDIR}/bud/platform/Dockerfileamd
```
vrothberg, line 155: Wouldn't that break as well without qemu-user-static?  I don't see the "trying" as suggested by the comment.

rhatdan: Only need user-static if you are doing a RUN command. (Having not checked if this does a RUN.)

vrothberg:
```
FROM --platform=linux/amd64 busybox
RUN uname -m
```

flouthoc: ackd let me add same test here as well.

flouthoc: I am removing `RUN` call so this does not fail.

tests/bud.bats
```diff
@@ -117,6 +117,48 @@ symlink(subdir)"
...
 
+# Following test must fail since we are trying to run linux/arm64 on linux/amd64
+# Issue: https://github.com/containers/buildah/issues/3712
+@test "build-with-inline-platform" {
+  # Host arch
+  run_buildah info --format '{{.host.arch}}'
+  myarch="$output"
+  for try in amd64 arm64 ppc64le s390x; do
+    if [[ "$otherarch" != "$myarch" ]]; then
+      otherarch="$myarch"
```
TomSweeneyRedHat, line 128: I think otherarch will always be set to either amd64 or arm64 here.  Could we randomize this, or make the loop bigger and test all three of the other archs?

tests/bud.bats
```diff
@@ -117,6 +117,61 @@ symlink(subdir)"
...
+  fi
+
+  cat > ${TESTSDIR}/bud/platform/Dockerfile << _EOF
+FROM --platform=linux/${myarch} alpine
+RUN uname -m
+_EOF
+
+  # Tries building image where baseImage has --platform=linux/HostArch
+  run_buildah '?' build --platform linux/${targetarch} --signature-policy ${TESTSDIR}/policy.json -t test ${TESTSDIR}/bud/platform
+  run_buildah inspect --format '{{ .OCIv1.Architecture }}' test
```
vrothberg, line 170: That would fail if no qemu isn't installed.  Don't we need the same checks as in the test abov?

flouthoc: @vrothberg No this should not fail since we are only tagging as different `arch` the image which we are actually pulling is specifically set to match `host` arch. I am just creating `dockerfile` dynamically for specific host.

vrothberg: The `RUN` instruction is still there, and we ignore the exit code.  If the build fails, there's no tagged image and then the inspect should fail.

tests/bud.bats
```diff
@@ -117,6 +117,61 @@ symlink(subdir)"
...
+  myarch="$output"
+  targetarch="arm64"
+
+  if [[ "$targetArch" == "$myarch" ]]; then
+    targetarch="amd64"
+  fi
+
+  cat > ${TESTSDIR}/bud/platform/Dockerfile << _EOF
+FROM --platform=linux/${myarch} alpine
+RUN uname -m
```
vrothberg, line 165: This overwrite an existing Dockerfile in the git tree, please don't.

tests/bud.bats
flouthoc: @vrothberg I'm confused a bit I guess. But the idea is if we pull the image which is equivalent to `hostarch` `RUN` should never fail.

vrothberg: We build for a non-local platform, so `build` could fail if qemu isn't installed, couldn't it?  If it's intended to succeed, I am confused why we ignore the exist code, see `?`.

tests/bud.bats
nalind: Yeah, do not merge until this is changed.

imagebuildah/stage_executor.go
```diff
@@ -595,6 +596,24 @@ func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBCo
...
+		}
+		if arch != "" {
+			builderSystemContext.ArchitectureChoice = arch
+		}
+		if os != "" {
+			builderSystemContext.OSChoice = os
+		}
+		if variant != "" {
+			builderSystemContext.VariantChoice = variant
+		}
```
nalind, line 614: `arch` and `variant` need to be managed together, else on a `linux/arm/v7` system or the like you could conceivably find yourself attempting to pull an image for `linux/386/v7`.

tests/bud.bats
```diff
@@ -117,6 +117,61 @@ symlink(subdir)"
...
+    expect_output --substring "format error"
+  fi
+}
+
+# Following test must pass since we want to tag image as host arch
+# Test for use-case described here: https://github.com/containers/buildah/issues/3261
+@test "build-with-inline-platform-amd-but-tag-as-arm" {
+  # Host arch
+  run_buildah info --format '{{.host.arch}}'
+  myarch="$output"
```
nalind, line 156: Might be simpler to grep the output of `buildah version`'s `BuildPlatform`, which provides the entire value you're looking to use.

tests/bud.bats
flouthoc: changed to `${TESTDIR}`

tests/bud.bats
flouthoc: @nalind I need to toggle  only `arch` at certain places that's why i preferred this. Hope this is not a blocker.

imagebuildah/stage_executor.go
flouthoc: fixed.

imagebuildah/stage_executor.go
```diff
@@ -595,6 +596,22 @@ func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBCo
...
 	}
 
+	builderSystemContext := s.executor.systemContext
+	// get platform string from stage
+	if stage.Builder.Platform != "" {
+		os, arch, variant, err := parse.Platform(stage.Builder.Platform)
+		if err != nil {
+			return nil, errors.Wrapf(err, "unable to parse platform %q", stage.Builder.Platform)
+		}
+		if arch != "" && variant != "" {
```
nalind, line 606: Change this to `||`.  For non-ARM platforms, variant will typically be empty, and that would cause `arch` to be ignored for all of them.

tests/bud.bats
flouthoc: @vrothberg I have removed the `?` so if build fails the test will break. I think extra `?` was added cause i copied this test from the first one.

nalind: The `--platform` on the FROM instruction implicitly overrides the `--platform` we get at the command line, and this looks like it's attempting to verify that the build succeeds when the command line `--platform` points to a foreign architecture, but `FROM --platform` specifies the local one.  It'll succeed either way if emulation is enabled, though, which is probably not what we want.

nalind: Ah, and that's why it checks on the next line.

flouthoc: @nalind That's correct.

vrothberg: > The --platform on the FROM instruction implicitly overrides the --platform we get at the command line, and this looks like it's attempting to verify that the build succeeds when the command line --platform points to a foreign architecture, but FROM --platform specifies the local one. It'll succeed either way if emulation is enabled, though, which is probably not what we want.

Is the behavior that the Dockerfile instructions overrides the CLI inherited from Docker or buildkit?  My expectation is that when I say `--platform foo` on the CLI that I get platform foo.

flouthoc: @vrothberg Yes `buidlkit` supports that. The reason is so users could do `multi-platform` builds otherwise the variables which we set with cli `--platform` like `BUILDPLATFORM` and `TARGETPLATFORM` would not be a lot of useful to end users if they can't override base images.

There is more discussed here: https://github.com/containers/buildah/issues/3261 and https://docs.docker.com/buildx/working-with-buildx/#build-multi-platform-images

Tryout example with `buildkit`:

```dockerfile
FROM --platform=linux/amd64 alpine
RUN uname -m
```

But build this docker file using `--platform` different than one set in `Dockerfile`.

```console
sudo docker buildx build --platform linux/arm64 -t test .
```

Now on inspecting this image `sudo docker inspect test`  you can actually see that image for overridden by `inline` `linux/amd64` platform but still tagged as `linux/arm64`.

Issue #3700, opened 2022-01-10
podman build --cpuset-cpus= is not working


vans163:
/kind bug

**Description**

using `podman build` with `--cpuset-cpus=` does not bind the threads to the physical cores. `podman run` works fine.

**Steps to reproduce the issue:**

```dockerfile
FROM ubuntu:20.04
ENV DEBIAN_FRONTEND noninteractive
RUN apt-get update && apt-get install -y stress
RUN stress --cpu 32
CMD ["/bin/bash"
```
```
podman build --cpuset-cpus=0 --tag test .
```
Observe as all CPUs are flooded at 100% with work.

```
podman run --cpuset-cpus=0 --tag test .
```
Observe as only CPU 0 is flooded at 100% with work.

**Describe the results you expected:**

Expect `podman build` to work equivalent to `podman run`

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**

```
podman version 3.4.4
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.23.1
  cgroupControllers:
  - cpuset
  - cpu
  - io
  - memory
  - hugetlb
  - pids
  - rdma
  - misc
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: 'conmon: /usr/bin/conmon'
    path: /usr/bin/conmon
    version: 'conmon version 2.0.25, commit: unknown'
  cpus: 32
  distribution:
    codename: jammy
    distribution: ubuntu
    version: "22.04"
  eventLogger: journald
  hostname: video2
  idMappings:
    gidmap: null
    uidmap: null
  kernel: 5.13.0-19-generic
  linkmode: dynamic
  logDriver: journald
  memFree: 129869611008
  memTotal: 135063539712
  ociRuntime:
    name: runc
    package: 'runc: /usr/sbin/runc'
    path: /usr/sbin/runc
    version: |-
      runc version 1.0.1-0ubuntu3
      spec: 1.0.2-dev
      go: go1.17.3
      libseccomp: 2.5.2
  os: linux
  remoteSocket:
    exists: true
    path: /run/podman/podman.sock
  security:
    apparmorEnabled: true
[...]

PR #3780, opened 2022-02-14
run_linux: use --systemd-cgroup

giuseppe:
when the configured cgroup manager is set to "systemd", use the --systemd-cgroup option to the OCI runtime.  It is understood both by crun and runc.

Closes: https://github.com/containers/buildah/issues/3700


run_linux.go
```diff
@@ -874,9 +875,14 @@ func runUsingRuntime(options RunOptions, configureNetwork bool, moreCreateArgs [
 		}
 	}
 
+	runtimeArgs := options.Args[:]
```
flouthoc, line 878: nit,question: Is reason for re-slice here is to make copy of `options.Args` or could we just do `runtimeArgs := options.Args`

flouthoc (approved):
LGTM

run_linux.go
giuseppe: yes, I don't want to modify the input

Issue #2937, opened 2021-01-26
podman history exposes proxy related build args


rolweber:
**Description**

I'm trying to use `podman build` as a drop-in replacement for `docker build`.
I'm (ab)using the no_proxy build arg to pass an API key at build time, which should not be exposed in the image.

With `docker`, no_proxy and other pre-defined build args do not appear in the image history:
https://docs.docker.com/engine/reference/builder/#predefined-args
With `podman`, the value of no_proxy appears in the image history. I didn't try the other ones.

**Steps to reproduce the issue:**
1. `podman build -t whatever --build-arg "no-proxy=to-be-kept-secret" .`
2. `podman history --no-trunc --format json whatever`

I'll add a Dockerfile, shell script and Makefile to reproduce the problem conveniently.

**Describe the results you received:**

The history shows the value I passed as the `no_proxy` build argument.

**Describe the results you expected:**

The history should not show the value

 I passed as the `no_proxy` build argument.

**Output of `podman version` if reporting a `podman build` issue:**

```
Version:      2.2.1
API Version:  2.1.0
Go Version:   go1.15.5
Built:        Tue Dec  8 15:37:50 2020
OS/Arch:      linux/amd64
```

**Output of `cat /etc/*release`:**

```
Linux Client for e-business (Fedora) 33 (Daily)
Fedora release 33 (Thirty Three)
Open Client Fedora 33 (Daily)
NAME=Fedora
VERSION="33 (Workstation Edition)"
ID=fedora
VERSION_ID=33
VERSION_CODENAME=""
PLATFORM_ID="platform:f33"
PRETTY_NAME="Fedora 33 (Workstation Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:33"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f33/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=33
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=33
[...]

PR #3782, opened 2022-02-15
history, build: add pre-allowed variables e.g `proxy vars` to history if explicitly specified

flouthoc:
`Buildkit/Docker` only adds pre-allowed variables for e.g `proxy
variables` to history if user explicitly specifies them in
Dockerfile/Containerfile using `ARG`.

By default variables in allowlist e.g `proxy variables` and values will be used
normally but will not be leaked into `docker/OCI` history of images.

A test for following behavior is added with this commit and similar test
can be verified against `Docker/Buildkit`

Closes: https://github.com/containers/buildah/issues/2937

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -19,6 +19,7 @@ import (
 	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/buildah/pkg/rusage"
 	"github.com/containers/buildah/util"
+	config "github.com/containers/common/pkg/config"
 	cp "github.com/containers/image/v5/copy"
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/manifest"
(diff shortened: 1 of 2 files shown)


imagebuildah/stage_executor.go
```diff
@@ -64,6 +65,7 @@ type StageExecutor struct {
 	output          string
 	containerIDs    []string
 	stage           *imagebuilder.Stage
+	processedArgs   []string // contains args which are processed and were added implicitly in Containerfile using ARG
```
vrothberg, line 68:
```suggestion
	argsFromContainerfile   []string
```
?

imagebuildah/stage_executor.go
```diff
@@ -1271,7 +1278,31 @@ func (s *StageExecutor) getBuildArgsResolvedForRun() string {
 			if inImage {
 				envs = append(envs, fmt.Sprintf("%s=%s", key, configuredEnvs[key]))
 			} else {
-				envs = append(envs, fmt.Sprintf("%s=%s", key, value))
+				// by default everything must be added to history
+				// this is configured to false only for special cases
```
vrothberg, line 1282: Please use punctuation in sentences and avoid lower-case only printing; I have a hard time brain-parsing the comments.

imagebuildah/stage_executor.go
```diff
@@ -1271,7 +1278,31 @@ func (s *StageExecutor) getBuildArgsResolvedForRun() string {
...
+				addToHistory := true
+
+				// following value is being assigned from build-args
+				// check if this key belongs to any of the predefined whitelisted args e.g Proxy Variables
+				// and if that arg is not manually set in Containerfile/Dockerfile
+				// then don't write its value to history.
+				// following behaviour ensures parity with docker/buildkit behviour
```
vrothberg, line 1289:
```suggestion
				// following behaviour ensures parity with docker/buildkit behaviour
```

imagebuildah/stage_executor.go
```diff
@@ -1271,7 +1278,31 @@ func (s *StageExecutor) getBuildArgsResolvedForRun() string {
...
+				for _, variable := range config.ProxyEnv {
+					if key == variable {
+						// found in predefined args
+						// so don't add to history
+						// unless user did implicit `ARG <some-predefined-proxy-variable>`
```
vrothberg, line 1294:
```suggestion
						// unless user did explicit `ARG <some-predefined-proxy-variable>`
```

vrothberg (commented):
Code LGTM, just a couple of nits.

Can you do a `s/whitelist/allowlist/`? "Whitelisting" is an old non-inclusive term that we want to avoid.

Can you add `Fixes: https://github.com/containers/buildah/issues/2937` to the commit message? That would add some useful information to the git history.

imagebuildah/stage_executor.go
flouthoc: Modified comments a bit.

flouthoc:
@vrothberg Thanks tweaked comments as requested.

vrothberg (commented):
Thank you, LGTM

nalind:
> Can you do a `s/whitelist/allowlist/`? "Whitelisting" is an old non-inclusive term that we want to avoid.

Test function names, too.

PR #3788, opened 2022-02-21
speed up combination-namespaces test

### LINKED ISSUE #3768: placeholder: combination-namespaces test taking 30m
Placeholder, because this is going to take me a long time to isolate and I want eyes on it ASAP.

The `combination-namespaces` test is taking almost 30m in Rawhide. This is causing gating-test timeouts and failures. I can easily
reproduce on a 1minutetip VM:
```
# bats -T ...
...
 ✓ idmapping [131415]
 ✓ ipc-namespace [76657]
 ✓ net-namespace [61051]
 ✓ network-namespace [58675]
 ✓ pid-namespace [69845]
 ✓ user-namespace [35903]
 ✓ uts-namespace [72815]
 ✓ combination-namespaces [1764125]      <--- 1764s = 0:29:24
```
(All the other namespace tests feel like they take longer, too, but I don't have hard data. Just a gut feel).

`git log` shows no recent changes to the [bats file](https://github.com/containers/buildah/commits/main/tests/namespaces.bats). The last time it was touched, in November, has a namespace-related comment, but when I look at the diff I don't see how that could cause this huge delay.

Luap99:
#### What type of PR is this?

/kind cleanup

Commit d05957a6f6de added the private value for namespace flags and kept
container as alias for backwards compat. Commit b480ce832f3e added the
cgroup falg so the test has 3^6 combinations to test. This takes way to
long. Since container is the same as private we only test private. The
container value is already covered in the other namespace tests, see
general_namespace() function.

With this commit the test time went down from 23 min to 3 min on my
laptop.

#### How to verify it

Run the test and compare the time.

Fixes #3768

```release-note

```

### DIFF
--- a/tests/namespaces.bats
+++ b/tests/namespaces.bats
@@ -403,14 +403,14 @@ _EOF
 
   _prefetch alpine
   # mnt is always per-container, cgroup isn't a thing OCI runtime lets us configure
-  for ipc in host container private ; do
-    for net in host container private; do
-      for pid in host container private; do
-        for userns in host container private; do
-          for uts in host container private; do
-	    for cgroupns in host container private; do
+  for ipc in host private; do
+    for net in host private; do
+      for pid in host private; do
+        for userns in host private; do
+          for uts in host private; do
+            for cgroupns in host private; do
 
-              if test $userns == private -o $userns == container -a $pid == host ; then
+              if test $userns == private -a $pid == host ; then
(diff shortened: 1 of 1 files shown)


Luap99:
@edsantiago PTAL

edsantiago:
3<sup>6</sup> to 2<sup>6</sup> is a huge difference. Thank you for catching this @Luap99 

/lgtm

PR #3794, opened 2022-02-22
[WIP] imagebuildah: prevent race cause `c/storage` does not persist names with layers.

flouthoc:
When running concurrent builds references to layers are changed and
c/storage does not accounts for names so use ID where possible.

Builds fail with `image not known`.

### Reproducer

```bash
#!/bin/bash
x=1
sudo ./podman rmi -af
rm -f log*.*
sudo ./podman build -t first . &> logfirst.log
while [ $x -le 30 ]
do
  echo "$x times"
  sudo ./podman build --log-level debug -t $x . &> log$x.log &
  x=$(( $x + 1 ))
#  sleep 1
done
````
### Dockerfile

```Dockerfile
FROM quay.io/jitesoft/alpine
```

### Note:

This PR is just for discussion to confirm if we can do better solutions or not. I believe underlying problem still persists either in `c/image` or `c/storage`. **See bz 2055487**

### DIFF
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -744,18 +744,18 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 	if dest, err := b.resolveNameToImageRef(b.output); err == nil {
 		switch dest.Transport().Name() {
 		case is.Transport.Name():
-			img, err := is.Transport.GetStoreImage(b.store, dest)
+			_, img, err := util.FindImage(b.store, "", b.systemContext, imageID)
 			if err != nil {
 				return imageID, ref, errors.Wrapf(err, "error locating just-written image %q", transports.ImageName(dest))
 			}
-			if len(b.additionalTags) > 0 {
-				if err = util.AddImageNames(b.store, "", b.systemContext, img, b.additionalTags); err != nil {
-					return imageID, ref, errors.Wrapf(err, "error setting image names to %v", append(img.Names, b.additionalTags...))
-				}
-				logrus.Debugf("assigned names %v to image %q", img.Names, img.ID)
+
+			b.additionalTags = append(b.additionalTags, dest.DockerReference().String())
+			if err = util.AddImageNames(b.store, "", b.systemContext, img, b.additionalTags); err != nil {
+				return imageID, ref, errors.Wrapf(err, "error setting image names to %v", append(img.Names, b.additionalTags...))
 			}
-			// Report back the caller the tags applied, if any.
-			img, err = is.Transport.GetStoreImage(b.store, dest)
+			logrus.Debugf("assigned names %v to image %q", img.Names, img.ID)
+
+			_, img, err = util.FindImage(b.store, "", b.systemContext, imageID)
 			if err != nil {
 				return imageID, ref, errors.Wrapf(err, "error locating just-written image %q", transports.ImageName(dest))
 			}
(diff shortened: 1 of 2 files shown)


flouthoc:
@giuseppe @nalind PTAL. I am sure something could be done at `c/storage` to fix rootcause as well but this should also fix the mentioned issue in the PR and original BZ.

Issue is easy to reporduce with `podman` or `buildah build --layers` for some reason i don't see this with `buildah build --layers` very often however race is easily reproduce-able in `podman`.

giuseppe:
FTR, the reproducer is the following:

```bash
$ cat Dockerfile.fedora 
FROM fedora

$ cat reproducer.sh 
#!/bin/bash
x=1
podman rmi -af
rm -f log*.*
buildah bud -t first -f Dockerfile.fedora &> logfirst.log 
while [ $x -le 30 ]
do
  echo "$x times"
  buildah bud -t $x -f Dockerfile.fedora &> log$x.log & 
  x=$(( $x + 1 ))
done
wait
```

giuseppe (approved):
LGTM

imagebuildah/executor.go
```diff
@@ -744,18 +744,18 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
...
+
+			b.additionalTags = append(b.additionalTags, dest.DockerReference().String())
+			if err = util.AddImageNames(b.store, "", b.systemContext, img, b.additionalTags); err != nil {
+				return imageID, ref, errors.Wrapf(err, "error setting image names to %v", append(img.Names, b.additionalTags...))
 			}
-			// Report back the caller the tags applied, if any.
-			img, err = is.Transport.GetStoreImage(b.store, dest)
+			logrus.Debugf("assigned names %v to image %q", img.Names, img.ID)
+
+			_, img, err = util.FindImage(b.store, "", b.systemContext, imageID)
```
rhatdan, line 758: Isn't this the same line that was called in 747?

flouthoc: @rhatdan It is called again after we perform tagging of `current` name and any `additionalTags` if they were present. So we can return a success message for each one of them.

rhatdan: Is this still a draft or ready to merge?

flouthoc: I'll just test it with few more edge cases just to be sure and waiting for @nalind to also take a quick look.

imagebuildah/executor.go
```diff
@@ -744,18 +744,18 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
...
 			if err != nil {
 				return imageID, ref, errors.Wrapf(err, "error locating just-written image %q", transports.ImageName(dest))
 			}
-			if len(b.additionalTags) > 0 {
-				if err = util.AddImageNames(b.store, "", b.systemContext, img, b.additionalTags); err != nil {
-					return imageID, ref, errors.Wrapf(err, "error setting image names to %v", append(img.Names, b.additionalTags...))
-				}
-				logrus.Debugf("assigned names %v to image %q", img.Names, img.ID)
+
+			b.additionalTags = append(b.additionalTags, dest.DockerReference().String())
```
nalind, line 752: That name was assigned to the image when it was committed, wasn't it?

flouthoc: @nalind There seems be a race in `c/storage` and layers don't have `name` as the property while they get stored, so in concurrent build somehow the reference for `name` gets altered by other builds using the same `blob`.  

I can drop this particular line if its bothering and add a tight locking while tagging as well. But motivation behind this was just a  additional safety check, this might not be needed. I'll verify and remove this if its not needed.

flouthoc:
We might not need this PR anymore, `rootcase` will be fixed in `c/storage` here: https://github.com/containers/storage/pull/1153 and I'll close this PR once we fix `c/common` as well.

flouthoc:
This PR is not needed after https://github.com/containers/image/pull/1480 and https://github.com/containers/storage/pull/1153

PR #3797, opened 2022-02-28
executor: docker builds must inherit `healthconfig` from baseImage if any and new build does not overrides.

flouthoc:
If base image has a healthconfig and child does not overrides it then
docker formatted builds must inherit healthconfig from base image.

Following commit ensures that we are in parity with docker for docker
formatted builds.

Closes: https://github.com/containers/buildah/issues/3796

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -678,6 +678,7 @@ func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBCo
 			Volumes:      volumes,
 			WorkingDir:   builder.WorkDir(),
 			Entrypoint:   builder.Entrypoint(),
+			Healthcheck:  (*docker.HealthConfig)(builder.Healthcheck()),
 			Labels:       builder.Labels(),
 			Shell:        builder.Shell(),
 			StopSignal:   builder.StopSignal(),
--- a/tests/config.bats
+++ b/tests/config.bats
@@ -297,6 +297,27 @@ function check_matrix() {
   expect_output --substring "foo=bar"
 }
 
+@test "docker formatted builds must inherit healthcheck from base image" {
+  _prefetch busybox
+  ctxdir=${TESTDIR}/bud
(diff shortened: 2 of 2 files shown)


flouthoc:

imagebuildah/stage_executor.go
```diff
@@ -668,6 +668,17 @@ func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBCo
 		for _, p := range builder.Ports() {
 			ports[docker.Port(p)] = struct{}{}
 		}
+		buildahHealthConfig := builder.Healthcheck()
+		var dockerHealthConfig *docker.HealthConfig
```
rhatdan, line 672: Nit: Can you leave out the "docker" adjective.  Just call this healthCheck 
var healthCheck *docker.HealthConfig

rhatdan:
Fix the nit, and then LGTM

rhatdan:
@eesprit PTAL

imagebuildah/stage_executor.go
```diff
@@ -668,6 +668,17 @@ func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBCo
...
+		var healthConfig *docker.HealthConfig
+		if buildahHealthConfig != nil {
+			healthConfig = &docker.HealthConfig{
+				Test:        buildahHealthConfig.Test,
+				Interval:    buildahHealthConfig.Interval,
+				Timeout:     buildahHealthConfig.Timeout,
+				StartPeriod: buildahHealthConfig.StartPeriod,
+				Retries:     buildahHealthConfig.Retries,
+			}
+		}
```
nalind, line 681: `Healthcheck()` returns a fresh structure, so we don't really need to make a copy or use a temporary variable here.

flouthoc: @nalind They are of different types so had to cast.

nalind: Ah, right, too many packages named `docker`.

PR #3804, opened 2022-03-03
buildah: test rootless integration

flouthoc:
Perform buildah's integration tests in a completely `rootless` environment.


rhatdan:
@cevich PTAL

flouthoc:
This is still under WIP. @cevich @rhatdan I'll un-draft it when this is ready.

contrib/cirrus/lib.sh
```diff
@@ -257,3 +260,79 @@ execute_local_registry() {
 
     verify_local_registry
 }
+
+setup_rootless() {
+    req_env_vars GOPATH GOSRC SECRET_ENV_RE
+
+    ROOTLESS_USER="rootlessuser"
```
cevich, line 267: This line is unnecessary and might mislead future maintainers.  The top-level definition is perfectly fine.

contrib/cirrus/setup.sh
```diff
@@ -48,6 +48,29 @@ EOF
...
+        modprobe ip6_tables || :
+        modprobe ip6table_nat || :
+        setup_rootless
+        ;;
+    *) die_unknown PRIV_NAME
+esac
+
+if [[ -n "$ROOTLESS_USER" ]]; then
+    echo "ROOTLESS_USER=$ROOTLESS_USER" >> /etc/ci_environment
+fi
```
cevich, line 71: This is unnecessary since `ci_environment` isn't ever loaded anywhere.  It's fine to just have `$ROOTLESS_USER` hard-coded at the top of `lib.sh`, that will get picked up everywhere.

contrib/cirrus/setup.sh
```diff
@@ -48,6 +48,29 @@ EOF
         ;;
 esac
 
+# Required to be defined by caller: Are we testing as root or a regular user
+case "$PRIV_NAME" in
+    root)
+        if [[ "$TEST_FLAVOR" = "sys" ]]; then
+            # Used in local image-scp testing
+            setup_rootless
+            echo "PODMAN_ROOTLESS_USER=$ROOTLESS_USER" >> /etc/ci_environment
```
cevich, line 57: This line is unnecessary and won't get used.

contrib/cirrus/test.sh
```diff
@@ -51,11 +69,13 @@ else
...
-            # "secret.opts"]" which is in process of getting fixed but will
-            # take a while until it hits all distributions.
-            showrun rm /usr/share/containers/containers.conf
+            if [[ "$PRIV_NAME" != "rootless" ]]; then
+               # FIXME: drop the `rm` below once containers.conf has been fixed.
+               # It complains about failing to "to decode the keys ["secret"
+               # "secret.opts"]" which is in process of getting fixed but will
+               # take a while until it hits all distributions.
+               showrun rm /usr/share/containers/containers.conf
+            fi
```
cevich, line 78: Might be worth trying without this block, I don't know when it was added or what the issue was, but it may already be fixed.  Ya, this is causing the tests to fail when running as a regular user: permission denied. If this removal is still necessary, perhaps move it into setup.sh where we still execute as root?

cevich (commented):
A good start.  I approve :grin:

TomSweeneyRedHat:
Nice start @flouthoc !

flouthoc:
@nalind @rhatdan @giuseppe There are some tests which performs `mount` or something equivalent so we can't easily run them on rootless env without `unshare`, hence i'm skipping them for `rootless environment`.

However most of the tests run fine with `rootless` environment.

rhatdan:
Comment them out for now, but it would be nice to see if we could do a `buildah unshare` before running these tests, when running in rootless mode.

cevich:
> Comment them out for now, but it would be nice to see if we could do a buildah unshare before running these tests, when running in rootless mode.

I'm not the expert, but I respectfully disagree.  Bind-mounting is a pretty important feature.  Commenting them out means we're not testing them *at all*.  Adding a `if uid != 0` condition to do a `buildah unshare` or a `skip_if_rootless()` sounds pretty easy...but I'm not an expert on these tests.  I think at a minimum, if anything is commented out or skipped, an issue should be filed so we remember to address it later.

flouthoc:
> > Comment them out for now, but it would be nice to see if we could do a buildah unshare before running these tests, when running in rootless mode.
> 
> I'm not the expert, but I respectfully disagree. Bind-mounting is a pretty important feature. Commenting them out means we're not testing them _at all_. Adding a `if uid != 0` condition to do a `buildah unshare` or a `skip_if_rootless()` sounds pretty easy...but I'm not an expert on these tests. I think at a minimum, if anything is commented out or skipped, an issue should be filed so we remember to address it later.

@cevich Ah i guess there is some misunderstanding `bind mount` tests will still work. Its just external `buildah mount` command which is causing issues not the `bind mount in Containerfile` or `containers`. We did similar skip and crafting for `mount` tests in rootless podman as well. But I don't think its that critical for rootless users so skipping them for now but we can enable it later.

Issue with `buildah unshare` is that it changes shell so all the tests using `buildah mount` would have to be executed separately inside a `buildah unshare`.

cevich:
>Issue with buildah unshare is that it changes shell so all the tests using buildah mount would have to be executed separately inside a buildah unshare instance.

Oh okay, thanks for the explanation.  I would def. rec. opening an issue on it then, at least so 2 years from now nobody's left wondering "howcome all these tests are skipped?".

flouthoc:
@cevich @rhatdan @nalind @TomSweeneyRedHat This is ready for review. 

Covers all the major `529` tests out of total `642` total integration tests.

rhatdan:
Seems like a good start, and everyone gets an extra doughnut if they turn on one of the 113 disabled rootless tests.
LGTM

rhatdan:
I feel like @edsantiago will win the most doughnuts.

@TomSweeneyRedHat @cevich @edsantiago @vrothberg @nalind @umohnani8 @ashley-cui PTAL

flouthoc:
> Seems like a good start, and everyone gets an extra doughnut if they turn on one of the 113 disabled rootless tests. LGTM

Actually its fairly easy to test disabled tests as well we just need to do `./buildah unsare test_script` I added another test in the matrix for this.

flouthoc:
@rhatdan @cevich Added `integration-rootless-unshare` that covers all the remaining tests as well except last `20` which is more of `root` user tests.

edsantiago:
With several enormous caveats (see below), the following diff goes a long way toward enabling (some) rootless tests on my f35:
```diff
index 596abdc6..b94c1c7e 100644
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -160,6 +160,31 @@ function podman() {
     command podman ${PODMAN_REGISTRY_OPTS} ${ROOTDIR_OPTS} "$@"
 }
 
+# Used in rootless
+function run_unshared() {
+    if is_rootless; then
+        $BUILDAH_BINARY unshare "$@"
+    else
+        command "$@"
+    fi
+}
+
+function mkdir() {
+    run_unshared mkdir "$@"
+}
+
+function touch() {
+    run_unshared touch "$@"
+}
+
+function cp() {
+    run_unshared cp "$@"
+}
+
+function rm() {
+    run_unshared rm "$@"
+}
+
 #################
 #  run_buildah  #  Invoke buildah, with timeout, using BATS 'run'
 #################
@@ -195,6 +220,12 @@ function run_buildah() {
     # Remember command args, for possible use in later diagnostic messages
     MOST_RECENT_BUILDAH_COMMAND="buildah $*"
 
+    if is_rootless; then
+        if [[ "$1" =~ mount ]]; then
+            set "unshare" "$BUILDAH_BINARY" ${BUILDAH_REGISTRY_OPTS} ${ROOTDIR_OPTS} "$@"
+        fi
+    fi
+
     while [ $retry -gt 0 ]; do
         retry=$(( retry - 1 ))
```

Now the caveats:
* I've done only cursory testing
* I tried to refactor the `REGISTRY_OPTS` and `ROOTDIR_OPTS` mess, but ended up with spaghettier code. 
* None of this works for the `run_buildah ERRORCODE` tests, because `buildah unshare` has this lovely quirk wherein if a command fails, it loudly tells you `time=... level=error msg="exit status 125"` which of course invalidates output checks. Is there any way to silence those? They don't seem to add any value, since the `buildah unshare` exit status is exactly the same as that of the unshare'd command

tests/helpers.bash
```diff
@@ -410,6 +410,24 @@ function is_rootless() {
...
+    if is_rootless; then
+        skip "${1:-test is being invoked from rootless environment and might need unshare}"
+    fi
+}
+
+#################################
+#  skip_if_rootless_unshare     # `some` of the tests can be only triggered in rootfull
+#################################
+function skip_if_rootless_unshare() {
+    if [ $_CONTAINERS_ROOTLESS_UID -ne 0 ]; then
```
edsantiago, line 426: This has to be double `[[`, not single. Single `[` causes an error, it does not do what you think it does.

Also, I have no idea why this function even exists, since AFAICT it will never trigger.

edsantiago:
This is a lot to review; I haven't even finished my first pass.

>We should only do unshare in tests where its needed.

I'm confused. Isn't that what my diff does?

.cirrus.yml
```diff
@@ -250,6 +252,92 @@ integration_task:
...
+
+    always: &standardlogs
+        audit_log_script: '$GOSRC/$SCRIPT_BASE/logcollector.sh audit'
+        df_script: '$GOSRC/$SCRIPT_BASE/logcollector.sh df'
+        journal_script: '$GOSRC/$SCRIPT_BASE/logcollector.sh journal'
+        podman_system_info_script: '$GOSRC/$SCRIPT_BASE/logcollector.sh podman'
+        buildah_version_script: '$GOSRC/$SCRIPT_BASE/logcollector.sh buildah_version'
+        buildah_info_script: '$GOSRC/$SCRIPT_BASE/logcollector.sh buildah_info'
+        package_versions_script: '$GOSRC/$SCRIPT_BASE/logcollector.sh packages'
+        golang_version_script: '$GOSRC/$SCRIPT_BASE/logcollector.sh golang'
```
cevich, line 299: Please don't re-define the `&standardlogs` anchor.  Simply re-use it by replacing this whole block with the line:

```yaml
    always: *standardlogs
```

.cirrus.yml
```diff
@@ -250,6 +252,92 @@ integration_task:
...
```
cevich, line 340: Same here, please replace with:

```yaml
    always: *standardlogs
```

tests/authenticate.bats
```diff
@@ -19,7 +19,7 @@ load helpers
   run_buildah 0 login --username testuserfoo --password testpassword docker.io
 
   run_buildah 125 logout --authfile /tmp/nonexistent docker.io
-  expect_output "checking authfile: stat /tmp/nonexistent: no such file or directory"
+  expect_output --substring "checking authfile: stat /tmp/nonexistent: no such file or directory"
```
edsantiago, line 22: I see a lot of these `--substring` additions. These are bugs. These should not happen.

The root cause seems similar to the `buildah unshare` bug I reported earlier: some commands emit an unhelpful, unwanted extra "exit status" message:
```
$ bin/buildah logout --authfile /foo quay.io
checking authfile: stat /foo: no such file or directory
time="2022-03-07T09:54:01-07:00" level=error msg="exit status 125"
```
Can we fix the problem at the root, instead of working around it? @nalind do you know why this bug exists, and how to fix it?

flouthoc: @edsantiago Sure. I'll check this issue seems to be with `buildah`. It logs `error` on `stderr` in case of `rootless` but does not in case of `root`. Not sure sure which one is the correct here but I'll make sure `rootless` buildah behaves same as `root`.

.cirrus.yml
edsantiago: @cevich, good catch, but I see an inconsistency. The other place where I see `*standardlogs` in that file has a `<<` next to it:
```
    always:
        <<: *standardlogs
```
Is that a trivial syntactic-sugary non-difference, or is it important? To avoid confusing future maintainers, I would suggest sticking with one convention.

Makefile
```diff
@@ -166,6 +166,11 @@ test-integration: install.tools
 	./tests/tools/build/ginkgo $(BUILDFLAGS) -v tests/e2e/.
 	cd tests; ./test_runner.sh
 
+.PHONY: test-integration-unshare
+test-integration-unshare: install.tools
+	./tests/tools/build/ginkgo $(BUILDFLAGS) -v tests/e2e/.
```
edsantiago, line 171: Should there be an `unshare` here? If not, it is identical to line 166, shouldn't it be removed as a dup?

tests/authenticate.bats
flouthoc: @nalind Is it safe to remove error from `c/storage`, I have a feeling that something is using this `error` on `podman` side but just a feeling, I'll check this.

.cirrus.yml
cevich: As long as the second map (dictionary) has no additions, the two are equivalent.  In other words, the second form (with `<<:`) means "copy the anchor value here" and then "add the subsequent new/changed entries".  For example:

```yaml
one: &thing
    foo: abc
    bar: def

two: *thing

three: &newthing
    <<: *thing
    bar: snafu

four: *newthing
```

Means the dictionaries `one` and `two` are identical.  But `three` contains `{foo: abc, bar: snafu}`, and therefor `four` does also.

tests/authenticate.bats
flouthoc: Fixed here: https://github.com/containers/storage/pull/1162

Makefile
flouthoc: @edsantiago I have removed `unshare` bit and will not cover `unshare` in this PR to make rootless testing simple.

.cirrus.yml
flouthoc: this is fixed in recent push.

tests/helpers.bash
flouthoc: @edsantiago I have removed `unshare` part to keep things simple for rootless testing.

flouthoc:
@edsantiago @cevich @nalind  As per discussions from above and before I have removed `unshare` part from this PR in-order to maintain simplicity for this PR also fixed all the requested nits. Could you PTAL again.

edsantiago:
LGTM but I'm curious to know if you reviewed my patch above for fixing the `mount` issue? I believe it's a pretty safe way to increase rootless test coverage.

flouthoc:
@edsantiago Your patch looks perfect to me, I'll pick it up in just another followup PR after this, cause we might need a seperate cirrus task for test inside `unshare` so that we can test both `rootless + test_with_unshare` and `rootless + test_without_unshare`.

But I'm fine to do it in this PR as well if you prefer that.

edsantiago:
I guess I don't have a strong preference. I was basing my suggestion on @rhatdan's comment

>everyone gets an extra doughnut if they turn on one of the 113 disabled rootless tests.

(in which I would substitute "the satisfaction of increased test coverage" for "doughnut"). I haven't actually looked closely at which tests are disabled, and even if I did I couldn't judge how much we care about those, so I'll leave it up to you. This has my LGTM but I'd like at least one more set of eyes on it.

(And because I just _know_ that someone is going to pipe in with "needs a rebase": No, it doesn't. "branch is out-of-date" does not mean "needs a rebase". Why does the buildah github show those useless messages?)

cevich:
> Why does the buildah github show those useless messages?

I think it's automatic, but I'll double-check the settings.

cevich (approved):
LGTM (only checked Cirrus and contrib scripts.)

edsantiago:
@flouthoc nice work, thanks for doing this!

PR #3807, opened 2022-03-07
bats tests: initialize magic cni/network settings

edsantiago:
Gating tests are failing on rawhide because the system has only
netavark, no CNI, and the buildah-copy helper causes the "is
this a new installation" logic to go haywire. Add a workaround.

### DIFF
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -43,6 +43,19 @@ EOF
     ROOTDIR_OPTS="--root ${TESTDIR}/root --runroot ${TESTDIR}/runroot --storage-driver ${STORAGE_DRIVER}"
     BUILDAH_REGISTRY_OPTS="--registries-conf ${TESTSDIR}/registries.conf --registries-conf-dir ${TESTDIR}/registries.d --short-name-alias-conf ${TESTDIR}/cache/shortnames.conf"
     PODMAN_REGISTRY_OPTS="--registries-conf ${TESTSDIR}/registries.conf"
+
+    # This is not a NOP! It's a horrible kludge needed because of _prefetch().
+    # Basically: _prefetch puts images into our rootdir, buildah/podman see
+    # that there are images in the store, but do not see the magic netavark
+    # signal file, so they assume CNI, which fails catastrophically on
+    # a netavark-only system.
+    #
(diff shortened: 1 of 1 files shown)


edsantiago:
@Luap99 PTAL

nalind:
Up until now, `podman`'s been a soft requirement, in that we skipped tests that used it if we noticed it wasn't available to be called.  This turns it into a hard requirement.

edsantiago:
`podman` is a hard requirement for the `buildah-tests` package

nalind:
And that's something that can be ensured, for that package, at the packaging level.  I'm talking about what you get when you clone the repository, which did _not_ have that requirement before.

edsantiago:
Forgive me, I'm having trouble understanding your concern. I can see two possibilities:
1. You would like `buildah` tests to run on a system with no podman installed; or
2. You would like to make it clearer that `buildah` tests require podman

I don't know how to address either one. If the first, can you suggest a better way to populate the magic network file? If the second,  where would be a good place to document this? I actually don't even see any documentation at all (in the repo) on running the bats tests. Wherever this documentation is, it should probably include the other hard testing requirements like `jq`, `nc`, `golang`, `openssl`.

(Or perhaps there's option 3 which I missed?)

nalind:
> Forgive me, I'm having trouble understanding your concern. I can see two possibilities:
> 
>     1. You would like `buildah` tests to run on a system with no podman installed; or

Yes, I would like for them to be able to continue to be able to run on systems that don't have `podman` and a slew of our other tools installed.

> I don't know how to address either one. If the first, can you suggest a better way to populate the magic network file?

Perhaps #3810.

> If the second, where would be a good place to document this? I actually don't even see any documentation at all (in the repo) on running the bats tests. Wherever this documentation is, it should probably include the other hard testing requirements like `jq`, `nc`, `golang`, `openssl`.
> 
> (Or perhaps there's option 3 which I missed?)

Tests that don't check for commands that aren't installed by default are arguably buggy.  The main exception is the compiler, which had to be there in order to build the test binaries.  The Makefile target that runs the integration tests can more or less be boiled down to `bats $topdir/tests`.

edsantiago:
SGTM. Closing in favor of #3810. Thank you!

PR #3810, opened 2022-03-07
tests/copy: initialize the network, too

nalind:
#### What type of PR is this?

/kind other

Force libnetwork to record the network configuration so that its auto-detection logic doesn't get confused by the presence of the image that we're writing to storage, after it's been written.

#### How to verify it

Integration tests should pass.

```release-note
None
```

### DIFF
--- a/tests/copy/copy.go
+++ b/tests/copy/copy.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 
 	"github.com/containers/buildah"
+	"github.com/containers/common/libnetwork/network"
+	"github.com/containers/common/pkg/config"
 	cp "github.com/containers/image/v5/copy"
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@TomSweeneyRedHat @saschagrunert @giuseppe @vrothberg @umohnani8 @flouthoc PTAL

edsantiago:
Seems to work. In my testing, it creates `$ROOTDIR/defaultNetworkBackend` and populates it with `netavark`.

flouthoc:
I am late but LGTM

Luap99:
I created https://github.com/containers/buildah/pull/3814 because this will not read the actual containers.conf file.

PR #3812, opened 2022-03-08
buildah: increase rootless test coverage with `rootless + unshare`

flouthoc:
Tests for `rootless` added here https://github.com/containers/buildah/pull/3804 skipped some of the cases. 

Almost all of the skipped test cases can be covered under an environment of `rootless + unshare`. 
Following PR attempts to solve that by modifying test infrastructure and adding a `cirrus` task dedicated to `rootless+unshare` testing.

### DIFF
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -160,6 +160,35 @@ function podman() {
     command podman ${PODMAN_REGISTRY_OPTS} ${ROOTDIR_OPTS} "$@"
 }
 
+# There are various scenarios where we would like to execute `tests` as rootless user, however certain commands like `buildah mount`
+# do not work in rootless session since a normal user cannot mount a filesystem unless they're in a user namespace along with its
+# own mount namespace. In order to run such specific commands from a rootless session we must perform `buildah unshare`.
+# Following function makes sure that invoked command is triggered inside a `buildah unshare` session if env is rootless.
+function run_unshared() {
+    if is_rootless; then
+        $BUILDAH_BINARY unshare "$@"
+    else
+        command "$@"
+    fi
+}
+
+function mkdir() {
+    run_unshared mkdir "$@"
+}
+
+function touch() {
+    run_unshared touch "$@"
+}
+
+function cp() {
+    run_unshared cp "$@"
+}
+
+function rm() {
+    run_unshared rm "$@"
+}
+
+
 #################
 #  run_buildah  #  Invoke buildah, with timeout, using BATS 'run'
 #################
@@ -194,6 +223,14 @@ function run_buildah() {
 
     # Remember command args, for possible use in later diagnostic messages
     MOST_RECENT_BUILDAH_COMMAND="buildah $*"
+
+    # If session is rootless and `buildah mount` is invoked, perform unshare,
+    # since normal user cannot mount a filesystem unless they're in a user namespace along with its own mount namespace.
+    if is_rootless; then
+        if [[ "$1" =~ mount ]]; then
+            set "unshare" "$BUILDAH_BINARY" ${BUILDAH_REGISTRY_OPTS} ${ROOTDIR_OPTS} "$@"
+        fi
+    fi
 
     while [ $retry -gt 0 ]; do
         retry=$(( retry - 1 ))
--- a/tests/add.bats
+++ b/tests/add.bats
@@ -14,7 +14,6 @@ load helpers
 }
 
 @test "add-local-plain" {
-  skip_if_rootless_environment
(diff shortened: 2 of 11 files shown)


edsantiago:
I think there's some confusion here. What I meant to do with my patch is _eliminate the need to run the whole entire thing unshared_.

With my patch, you can remove a lot of the `skips`: the tests will pass under plain simple rootless. `unshare` is only used for `buildah mount/unmount`, so it's a better test of actual rootless.

Running the entire test suite under `unshare` does not seem terribly useful to me

flouthoc:
@edsantiago Yes. Actually this is intentional. My motive is to perform isolated testing for  `rootless` and `rootless+unshare` in different environment cause they both are testing different things. 

With this we can make sure that `user namespacing` is working correctly.

edsantiago:
I have no real idea what that means, but given what I see, you do not need my patch for anything: my patch was intended to make more tests pass. You have not removed any skips, you have only added more skips. That is the opposite of my intention.

Please either remove skips, or remove my patch.

flouthoc:
@edsantiago  Maybe my response above was a bit confusing but I'll try to re-pharse it and please correct me if I got it wrong anywhere.

So motivation behind this is that `rootless buildah` and `rootless buildah unshare` does fundamentally different things.

While `rootless buildah` **does not** creates its own `user namespace` hence it is has limited functionality but it still supports most of the feature which is tested here: https://github.com/containers/buildah/pull/3804

OTOH `rootless buildah unshare` creates it own `user namespace` including the `mount namespace` and **almost all** of the features work with `rootless buildah unshare` which works with `buildah root`

Therefore I think they provide different environment, my motivation is just to test these two differently.

flouthoc:
> I have no real idea what that means, but given what I see, you do not need my patch for anything: my patch was intended to make more tests pass. 

@edsantiago I think this we can still do this. Scratch my previous comment and let me try the original patch which you posted again since it only works for `mount / unmount` i think it does what we are trying to achieve.

flouthoc:
@edsantiago Thanks :smile: . I looked at it again and re-used your patch as is it and this enables us to remove skip from almost all of the `rootless` tests some are still skipped but they are more of a `root` user test so we should be good.

@rhatdan @cevich @nalind @edsantiago PTAL

tests/helpers.bash
```diff
@@ -160,6 +160,32 @@ function podman() {
     command podman ${PODMAN_REGISTRY_OPTS} ${ROOTDIR_OPTS} "$@"
 }
 
+# Used in rootless
```
edsantiago, line 163: A future maintainer might appreciate some thorough comments here

tests/helpers.bash
```diff
@@ -195,6 +221,12 @@ function run_buildah() {
     # Remember command args, for possible use in later diagnostic messages
     MOST_RECENT_BUILDAH_COMMAND="buildah $*"
 
+    if is_rootless; then
```
edsantiago, line 224: ...and especially here

edsantiago (commented):
Thank you @flouthoc, that is exactly what I intended. 

My patch was hasty and did not include comments; could you please add some for benefit of future maintainers?

edsantiago:
Oh, one more. Since there's no longer the need for multiple confusing `rootless_this_and_that`, could you just rename everything to `skip_if_rootless`?

flouthoc:
@edsantiago Thanks added requested comments.

> Oh, one more. Since there's no longer the need for multiple confusing `rootless_this_and_that`, could you just rename everything to `skip_if_rootless`?

I agree I think we can merge all existing function into one. I don't know why did we add `$BUILDAH_ISOLATION" = "rootless"
` check and `skip_if_rootless_and_cgroupv1` we might not even need all of this now. But I just want to go through commit history once before removing these. But since this look unrelated to this PR I'll do that in a new PR after reading commit history.

cevich (approved):
LGTM

cevich:
Oh, and super-big thanks for doing this, it really helped everyone out a lot, especially me :grin:

edsantiago (commented):
Language suggestions. Also, I have a medium-strong preference for eighty-column formatting especially in comments. I don't think we have a style guide, so I will not make my preference a blocker.

I see what you mean re: `skip_if_rootless` but we seem to have a very very bad situation that needs to be fixed: the existing `skip_if_rootless` bears no relation whatsoever to the `is_rootless()` function, and that is going to cause a lot of confusion one day (including today). That would be a very good topic for a followup PR.

flouthoc:
Restarted flakes.

PR #3814, opened 2022-03-09
tests/copy: read correct containers.conf

Luap99:
#### What type of PR is this?

 /kind bug

`DefaultConfig()` from c/common/pkg/config should not be used here.
This only contains the built in defaults but not the actual values from
the files on disk. The correct way is to call `Default()` and since this
already handles the `CONTAINERS_CONF` env var we do not need to handle
it manually.

```release-note

```

### DIFF
--- a/tests/copy/copy.go
+++ b/tests/copy/copy.go
@@ -74,17 +74,9 @@ func main() {
 			}
 			imageStorage.Transport.SetStore(store)
 
-			var conf *config.Config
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
/approve

flouthoc:
Restarted flakes.

flouthoc (approved):
LGTM

PR #3820, opened 2022-03-14
buildkit: `--mount-type=cache` must support `locking` external cache store.

flouthoc:
A shared cache on host must support locking so other parallel/concurrent builds
will wait for current executing RUN statement to finish.

* Locks the cache store as soon as RUN is triggered.
* Locked target is added to cleanup list so it can be unlocked as soon
  as RUN step is completed.

Usage 
```dockerfile
RUN --mount=type=cache,target=/test,sharing=locked
```

Closes: https://github.com/containers/buildah/issues/3815

### DIFF
--- a/cmd/buildah/run.go
+++ b/cmd/buildah/run.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/containers/buildah"
+	internalParse "github.com/containers/buildah/internal/parse"
 	buildahcli "github.com/containers/buildah/pkg/cli"
 	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/buildah/util"
@@ -157,7 +158,7 @@ func runCmd(c *cobra.Command, args []string, iopts runInputOptions) error {
 	if err != nil {
 		return errors.Wrapf(err, "error building system context")
 	}
-	mounts, mountedImages, lockedTargets, err := parse.GetVolumes(systemContext, store, iopts.volumes, iopts.mounts, iopts.contextDir)
+	mounts, mountedImages, lockedTargets, err := internalParse.GetVolumes(systemContext, store, iopts.volumes, iopts.mounts, iopts.contextDir)
 	if err != nil {
 		return err
 	}
--- a/cmd/buildah/run.go
+++ b/cmd/buildah/run.go
@@ -8,6 +8,7 @@ import (
 	buildahcli "github.com/containers/buildah/pkg/cli"
 	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/buildah/util"
+	"github.com/containers/storage/pkg/lockfile"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -156,7 +157,7 @@ func runCmd(c *cobra.Command, args []string, iopts runInputOptions) error {
 	if err != nil {
 		return errors.Wrapf(err, "error building system context")
 	}
-	mounts, mountedImages, err := parse.GetVolumes(systemContext, store, iopts.volumes, iopts.mounts, iopts.contextDir)
+	mounts, mountedImages, lockedTargets, err := parse.GetVolumes(systemContext, store, iopts.volumes, iopts.mounts, iopts.contextDir)
 	if err != nil {
 		return err
 	}
@@ -177,6 +178,32 @@ func runCmd(c *cobra.Command, args []string, iopts runInputOptions) error {
 		}
 		conditionallyAddHistory(builder, c, "%s %s", shell, strings.Join(args, " "))
 		return builder.Save()
+	}
+	// unlock if any locked files from this RUN statement
+	for _, path := range lockedTargets {
+		_, err := os.Stat(path)
+		if err != nil {
+			// Lockfile not found this might be a problem,
+			// since LockedTargets must contain list of all locked files
+			// don't break here since we need to unlock other files but
+			// log so user can take a look
+			logrus.Warnf("Lockfile %q was expected here, stat failed with %v", path, err)
+			continue
+		}
+		lockfile, err := lockfile.GetLockfile(path)
+		if err != nil {
+			// unable to get lockfile
+			// lets log error and continue
+			// unlocking other files
+			logrus.Warn(err)
+			continue
+		}
+		if lockfile.Locked() {
+			lockfile.Unlock()
+		} else {
+			logrus.Warnf("Lockfile %q was expected to be locked, this is unexpected", path)
+			continue
+		}
 	}
 	return runerr
 }
--- a/internal/parse/parse.go
+++ b/internal/parse/parse.go
@@ -14,6 +14,7 @@ import (
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/idtools"
+	"github.com/containers/storage/pkg/lockfile"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 )
@@ -28,6 +29,8 @@ const (
 	// mount=type=cache must create a persistent directory on host so its available for all consecutive builds.
 	// Lifecycle of following directory will be inherited from how host machine treats temporary directory
 	BuildahCacheDir = "buildah-cache"
+	// mount=type=cache allows users to lock a cache store while its being used by another build
+	BuildahCacheLockfile = "buildah-cache-lockfile"
 )
 
 var (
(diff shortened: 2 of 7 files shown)


run_linux.go
```diff
@@ -2850,6 +2854,18 @@ func (b *Builder) cleanupRunMounts(context *imagetypes.SystemContext, mountpoint
 			prevErr = err
 		}
 	}
+	// unlock if any locked files from this RUN statement
+	for _, path := range artifacts.LockedTargets {
+		if _, err := os.Stat(path); err == nil {
+			lockfile, err := lockfile.GetLockfile(path)
+			if err != nil {
+				return err
```
rhatdan, line 2862: Should this be returned or logged as prevErr and continue unlocking other lockedtargets?

internal/parse/parse.go
```diff
@@ -323,6 +330,20 @@ func GetCacheMount(args []string, store storage.Store, imageMountLabel string, a
...
+	case "locked":
+		// lock parent cache
+		lockfile, err := lockfile.GetLockfile(filepath.Join(newMount.Source, BuildahCacheLockfile))
+		if err != nil {
+			return newMount, errors.Wrapf(err, "Unable to acquire lock when sharing mode is locked")
+		}
+		// Will be unlocked after the RUN step is executed.
+		lockfile.Lock()
+	default:
+		// do nothing since default is `shared`
```
nalind, line 343: Needs to reject unrecognized values for the option.  Documentation also lists `private` as an available value here.

run.go
```diff
@@ -175,4 +175,6 @@ type runMountArtifacts struct {
 	Agents []*sshagent.AgentServer
 	// SSHAuthSock is the path to the ssh auth sock inside the container
 	SSHAuthSock string
+	// FilesPaths to be unlocked if there are any.
```
nalind, line 178: Comment doesn't match the field name.

run_linux.go
```diff
@@ -2512,6 +2514,7 @@ func (b *Builder) runSetupRunMounts(context *imagetypes.SystemContext, mounts []
 			}
 			finalMounts = append(finalMounts, *mount)
 			mountTargets = append(mountTargets, mount.Destination)
+			lockedTargets = append(lockedTargets, filepath.Join(mount.Source, internalParse.BuildahCacheLockfile))
```
nalind, line 2517: Why add something to this list that might not actually get locked?

run_linux.go
```diff
@@ -2850,6 +2854,22 @@ func (b *Builder) cleanupRunMounts(context *imagetypes.SystemContext, mountpoint
 			prevErr = err
 		}
 	}
+	// unlock if any locked files from this RUN statement
+	for _, path := range artifacts.LockedTargets {
+		if _, err := os.Stat(path); err == nil {
```
nalind, line 2859: Shouldn't ignore an error here, as it could be a symptom of a bigger problem.

flouthoc: Added to error log, but we might not want to break loop here since other targets must be unlocked.

internal/parse/parse.go
flouthoc: Yes we can error out here when value is not known. `private` is not yet implemented so it'll error out as well. I'll add `private` case here in another PR.

pkg/parse/parse.go
```diff
@@ -368,7 +368,7 @@ func getMounts(ctx *types.SystemContext, store storage.Store, mounts []string, c
 			finalMounts[mount.Destination] = mount
 			mountedImages = append(mountedImages, image)
 		case TypeCache:
-			mount, err := internalParse.GetCacheMount(tokens, store, "", nil)
+			mount, _, err := internalParse.GetCacheMount(tokens, store, "", nil)
```
nalind, line 371: Discarding a list of locked files, which will need to be unlocked at some point, looks like a potential source of problems in the future.

flouthoc: Pushed a version which takes care of this.

cmd/buildah/run.go
```diff
@@ -178,5 +179,28 @@ func runCmd(c *cobra.Command, args []string, iopts runInputOptions) error {
...
+		}
+		lockfile, err := lockfile.GetLockfile(path)
+		if err != nil {
+			// unable to get lockfile
+			// lets log error and continue
+			// unlocking other files
+			logrus.Error(err)
+			continue
+		}
+		if lockfile.Locked() {
```
nalind, line 201: At this point it should be an error if the file isn't locked, because we either unlocked it somewhere else by mistake, lost the lock for some reason, or `GetVolumes()` misled us.

flouthoc: Producing an error here for such anomaly cases.

pkg/parse/parse.go
```diff
@@ -307,18 +307,18 @@ func getVolumeMounts(volumes []string) (map[string]specs.Mount, error) {
 }
 
 // GetVolumes gets the volumes from --volume and --mount
-func GetVolumes(ctx *types.SystemContext, store storage.Store, volumes []string, mounts []string, contextDir string) ([]specs.Mount, []string, error) {
-	unifiedMounts, mountedImages, err := getMounts(ctx, store, mounts, contextDir)
+func GetVolumes(ctx *types.SystemContext, store storage.Store, volumes []string, mounts []string, contextDir string) ([]specs.Mount, []string, []string, error) {
```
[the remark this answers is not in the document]
flouthoc: @nalind I'm also unsure about what to do here. Maybe we can add a new function which is not exposed and mark this as `to-be` deprecated ?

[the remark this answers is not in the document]
flouthoc: Moved it to `internal/parse/parse.go`.

cmd/buildah/run.go
```diff
@@ -178,5 +180,31 @@ func runCmd(c *cobra.Command, args []string, iopts runInputOptions) error {
...
 	}
+	// unlock if any locked files from this RUN statement
+	for _, path := range lockedTargets {
+		_, err := os.Stat(path)
+		if err != nil {
+			// Lockfile not found this might be a problem,
+			// since LockedTargets must contain list of all locked files
+			// don't break here since we need to unlock other files but
+			// log so user can take a look
+			logrus.Errorf("Lockfile %q was expected here, stat failed with %v", path, err)
```
TomSweeneyRedHat, line 191: Is there anything a user could do at this point to correct this?  Or support for that matter?

TomSweeneyRedHat:
LGTM, but would like another head nod.

cmd/buildah/run.go
flouthoc: @TomSweeneyRedHat No user cannot do anything but I think this is worth logging cause following case is never expected to happen and if it happens it could be a symptom of a bigger problem in our fs locking library or anywhere above that.

flouthoc:
@rhatdan @nalind PTAL

cmd/buildah/run.go
TomSweeneyRedHat: That's what I thought @flouthoc .  I was kind of waffling about whether this should be a debug or not, but probably better to keep it as an error.

rhatdan: Errors should cause the app to exit, if this does not then this should be a warning.

cmd/buildah/run.go
```diff
@@ -178,5 +180,31 @@ func runCmd(c *cobra.Command, args []string, iopts runInputOptions) error {
...
+			continue
+		}
+		lockfile, err := lockfile.GetLockfile(path)
+		if err != nil {
+			// unable to get lockfile
+			// lets log error and continue
+			// unlocking other files
+			logrus.Error(err)
```
rhatdan, line 199: Warning.

cmd/buildah/run.go
```diff
@@ -178,5 +180,31 @@ func runCmd(c *cobra.Command, args []string, iopts runInputOptions) error {
...
+			continue
+		}
+		if lockfile.Locked() {
+			lockfile.Unlock()
+		} else {
+			logrus.Errorf("Lockfile %q was expected to be locked, this is unexpected", path)
```
rhatdan, line 205: Warning

cmd/buildah/run.go
flouthoc: Yes warning is better here.

cmd/buildah/run.go
flouthoc: fixed

flouthoc:
@rhatdan @TomSweeneyRedHat @nalind PTAL. This should be good i guess, resolved all the feedback points.

TomSweeneyRedHat:
@flouthoc, you might need a rebase

TomSweeneyRedHat:
LGTM other than the possible rebase.

Issue #3789, opened 2022-02-21
Support custom build outputs


joyfulmantis:
Docker has the ability to have custom build outputs, see https://docs.docker.com/engine/reference/commandline/build/#custom-build-outputs. This is convenient when one wants to use a container as a build environment and then have the resulting executables conveniently copied to the host.

From research I have done, it seems like buildah and therefore podman are misssing this functionality, and I could not find any open tickets about this missing functionality in the tickets.

rhatdan:
```
[](https://github.com/containers/podman/issues/13359#issue-1151676189)/kind feature

Description

I find the "local ouput" feature of buildx quite useful and use it extensively for my "short builds" (typically creating some .rpm/.deb packages of some of our internal tools already delivered as tarballs). I use a dockerfile with multi stage builds, where the very final target is "scratch" as base image, and I copy into this image the "result" of my build. When coupled with docker buildx --output type=local,dest=path feature, I can very easily get the result of my build in a local folder, without using things like "docker cp" or "podman mount". Plus, using "docker build" allows to benefit from docker caching so the same input instantely produces the same output when I have run it already (while I could use "docker run" in more complex builds like the ones I describe in https://github.com/containers/podman/issues/13358).

I understand you don't want to implement all the complexity of the buildx --ouput flag (actually, even on buildx side, this --output flag is half implemented as soon as you try doing multi-arch builds), but would it make sense to implement only the type=local output, or any equivalent podman flag ?
```

rhatdan:
@flouthoc PTAL

flouthoc:
@joyfulmantis Thanks above PR should close this.

joyfulmantis:
Wonderful, thanks!

jpolara2016:
I used the latest version of buildah and doesnt seem to create a tar.

```
[root@cf5f0621b1ae build]# buildah build --file alpine/Dockerfile -t alpine:latest -o type=tar,dest=/build/out.tar alpine/
STEP 1/1: FROM alpine
COMMIT alpine:latest
--> e66264b9877
Successfully tagged docker.io/library/alpine:latest
e66264b98777e12192600bf9b4d663655c98a090072e1bab49e233d7531d1294


[root@cf5f0621b1ae build]# ls /build/
alpine


[root@cf5f0621b1ae build]# buildah version
Version:         1.26.0
Go Version:      go1.18
Image Spec:      1.0.2-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        1.0.0
[...]

PR #3823, opened 2022-03-16
buildkit: add support for exporting build output with `--output` or `-o`

flouthoc:
Allows end-users to export final build content or rootfs to external formats.

By default, a local container image is created from the build result. The --output (or -o) flag allows you to override this behavior, and a specify a custom exporter. For example, custom exporters allow you to export the build artifacts as files on the local filesystem instead of a Container image, which can be useful for generating local binaries, code generation etc.

The value for --output is a CSV-formatted string defining the exporter type and options. Currently, local and tar exporters are supported. The local exporter writes the resulting build files to a directory on the client side. The tar exporter is similar but writes the files as a single tarball (.tar).

```console
buildah build --output type=local,dest=dir-rootfs .
buildah build --output type=tar,dest=rootfs.tar .
buildah build -o dir .
```
Reference: https://docs.docker.com/engine/reference/commandline/build/#custom-build-outputs

Closes: https://github.com/containers/buildah/issues/3789

### DIFF
--- a/cmd/buildah/build.go
+++ b/cmd/buildah/build.go
@@ -328,6 +328,15 @@ func buildCmd(c *cobra.Command, inputArgs []string, iopts buildOptions) error {
 		t := time.Unix(iopts.Timestamp, 0).UTC()
 		timestamp = &t
 	}
+	if c.Flag("output").Changed {
+		buildOption, err := parse.GetBuildOutput(iopts.BuildOutput)
+		if err != nil {
+			return err
+		}
+		if buildOption.IsStdout {
+			iopts.Quiet = true
+		}
+	}
 	options := define.BuildOptions{
 		AddCapabilities:         iopts.CapAdd,
 		AdditionalTags:          tags,
@@ -363,6 +372,7 @@ func buildCmd(c *cobra.Command, inputArgs []string, iopts buildOptions) error {
 		OS:                      systemContext.OSChoice,
 		Out:                     stdout,
 		Output:                  output,
+		BuildOutput:             iopts.BuildOutput,
 		OutputFormat:            format,
 		PullPolicy:              pullPolicy,
 		PullPushRetryDelay:      pullPushRetryDelay,
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -411,6 +411,25 @@ By default, Buildah manages _/etc/hosts_, adding the container's own IP address.
 
 Set the OS of the image to be built, and that of the base image to be pulled, if the build uses one, instead of using the current operating system of the host.
 
+**--output**, **-o**=""
+
+Output destination (format: type=local,dest=path)
+
+The --output (or -o) option extends the default behavior of building a container image by allowing users to export the contents of the image as files on the local filesystem, which can be useful for generating local binaries, code generation, etc.
+
+The value for --output is a comma-separated sequence of key=value pairs, defining the output type and options.
+
+Supported _keys_ are:
+- **dest**: Destination path for exported output. Valid value is absolute or relative path, `-` means the standard output.
+- **type**: Defines the type of output to be used. Valid values is documented below.
+
+Valid _type_ values are:
+- **local**: write the resulting build files to a directory on the client-side.
+- **tar**: write the resulting files as a single tarball (.tar).
+
+If no type is specified, the value defaults to **local**.
+Alternatively, instead of a comma-separated sequence, the value of **--output** can be just a destination (in the `**dest** format) (e.g. `--output some-path`, `--output -`) where `--output some-path` is treated as if **type=local** and `--output -` is treated as if **type=tar**.
+
 **--pid** *how*
 
 Sets the configuration for PID namespaces when handling `RUN` instructions.
@@ -823,6 +842,16 @@ buildah bud --platform linux/s390x,linux/ppc64le,linux/amd64 --manifest myimage 
 buildah bud --platform linux/arm64 --platform linux/amd64 --manifest myimage /tmp/mysrc
 
 buildah bud --all-platforms --manifest myimage /tmp/mysrc
+
+### Building an image using (--output) custom build output
+
+buildah build -o out .
+
+buildah build --output type=local,dest=out .
+
+buildah build --output type=tar,dest=out.tar .
+
+buildah build -o - . > out.tar
 
 ### Building an image using a URL
 
--- a/image.go
+++ b/image.go
@@ -43,6 +43,15 @@ const (
 	Dockerv2ImageManifest = define.Dockerv2ImageManifest
 )
 
+// ExtractRootfsOptions is consumed by ExtractRootfs() which allows
+// users to preserve nature of various modes like setuid, setgid and xattrs
+// over the extracted file system objects.
+type ExtractRootfsOptions struct {
+	StripSetuidBit bool // strip the setuid bit off of items being extracted.
+	StripSetgidBit bool // strip the setgid bit off of items being extracted.
+	StripXattrs    bool // don't record extended attributes of items being extracted.
+}
+
 type containerImageRef struct {
 	fromImageName         string
 	fromImageID           string
@@ -150,7 +159,10 @@ func computeLayerMIMEType(what string, layerCompression archive.Compression) (om
 }
 
 // Extract the container's whole filesystem as if it were a single layer.
-func (i *containerImageRef) extractRootfs() (io.ReadCloser, chan error, error) {
+// Takes ExtractRootfsOptions as argument which allows caller to configure
+// preserve nature of setuid,setgid,sticky and extended attributes
+// on extracted rootfs.
+func (i *containerImageRef) extractRootfs(opts ExtractRootfsOptions) (io.ReadCloser, chan error, error) {
 	var uidMap, gidMap []idtools.IDMap
 	mountPoint, err := i.store.Mount(i.containerID, i.mountLabel)
 	if err != nil {
@@ -164,8 +176,11 @@ func (i *containerImageRef) extractRootfs() (io.ReadCloser, chan error, error) {
 			uidMap, gidMap = convertRuntimeIDMaps(i.idMappingOptions.UIDMap, i.idMappingOptions.GIDMap)
 		}
 		copierOptions := copier.GetOptions{
-			UIDMap: uidMap,
-			GIDMap: gidMap,
+			UIDMap:         uidMap,
+			GIDMap:         gidMap,
+			StripSetuidBit: opts.StripSetuidBit,
+			StripSetgidBit: opts.StripSetgidBit,
+			StripXattrs:    opts.StripXattrs,
 		}
 		err = copier.Get(mountPoint, mountPoint, copierOptions, []string{"."}, pipeWriter)
 		errChan <- err
@@ -376,7 +391,7 @@ func (i *containerImageRef) NewImageSource(ctx context.Context, sc *types.System
 		var errChan chan error
 		if i.squash {
 			// Extract the root filesystem as a single layer.
-			rc, errChan, err = i.extractRootfs()
+			rc, errChan, err = i.extractRootfs(ExtractRootfsOptions{})
 			if err != nil {
 				return nil, err
 			}
@@ -738,7 +753,7 @@ func (i *containerImageSource) GetBlob(ctx context.Context, blob types.BlobInfo,
 	return ioutils.NewReadCloserWrapper(layerReadCloser, closer), size, nil
 }
 
-func (b *Builder) makeImageRef(options CommitOptions) (types.ImageReference, error) {
+func (b *Builder) makeContainerImageRef(options CommitOptions) (*containerImageRef, error) {
 	var name reference.Named
 	container, err := b.store.Container(b.ContainerID)
 	if err != nil {
@@ -812,4 +827,13 @@ func (b *Builder) makeImageRef(options CommitOptions) (types.ImageReference, err
 		postEmptyLayers:       b.AppendedEmptyLayers,
 	}
 	return ref, nil
+}
+
+// Extract the container's whole filesystem as if it were a single layer from current builder instance
+func (b *Builder) ExtractRootfs(options CommitOptions, opts ExtractRootfsOptions) (io.ReadCloser, chan error, error) {
+	src, err := b.makeContainerImageRef(options)
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "error creating image reference for container %q to extract its contents", b.ContainerID)
+	}
+	return src.extractRootfs(opts)
 }
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -133,6 +133,7 @@ type Executor struct {
 	unsetEnvs               []string
 	processLabel            string // Shares processLabel of first stage container with containers of other stages in same build
 	mountLabel              string // Shares mountLabel of first stage container with containers of other stages in same build
+	buildOutput             string // Specifies instructions for any custom build output
 }
 
 type imageTypeAndHistoryAndDiffIDs struct {
@@ -276,6 +277,7 @@ func newExecutor(logger *logrus.Logger, logPrefix string, store storage.Store, o
 		sshsources:                     sshsources,
 		logPrefix:                      logPrefix,
 		unsetEnvs:                      options.UnsetEnvs,
+		buildOutput:                    options.BuildOutput,
 	}
 	if exec.err == nil {
 		exec.err = os.Stderr
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -16,6 +16,7 @@ import (
 	"github.com/containers/buildah/define"
 	buildahdocker "github.com/containers/buildah/docker"
 	"github.com/containers/buildah/internal"
+	internalUtil "github.com/containers/buildah/internal/util"
 	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/buildah/pkg/rusage"
 	"github.com/containers/buildah/util"
@@ -28,6 +29,7 @@ import (
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/chrootarchive"
+	"github.com/containers/storage/pkg/unshare"
 	docker "github.com/fsouza/go-dockerclient"
 	digest "github.com/opencontainers/go-digest"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
@@ -1447,8 +1449,18 @@ func (s *StageExecutor) intermediateImageExists(ctx context.Context, currNode *p
 
 // commit writes the container's contents to an image, using a passed-in tag as
 // the name if there is one, generating a unique ID-based one otherwise.
+// or commit via any custom exporter if specified.
 func (s *StageExecutor) commit(ctx context.Context, createdBy string, emptyLayer bool, output string) (string, reference.Canonical, error) {
 	ib := s.stage.Builder
+	var buildOutputOption define.BuildOutputOption
+	if s.executor.buildOutput != "" {
+		var err error
+		logrus.Debugf("Generating custom build output with options %q", s.executor.buildOutput)
+		buildOutputOption, err = parse.GetBuildOutput(s.executor.buildOutput)
+		if err != nil {
+			return "", nil, errors.Wrapf(err, "failed to parse build output")
+		}
+	}
 	var imageRef types.ImageReference
 	if output != "" {
 		imageRef2, err := s.executor.resolveNameToImageRef(output)
@@ -1555,6 +1567,39 @@ func (s *StageExecutor) commit(ctx context.Context, createdBy string, emptyLayer
 		HistoryTimestamp:      s.executor.timestamp,
 		Manifest:              s.executor.manifest,
 		UnsetEnvs:             s.executor.unsetEnvs,
+	}
+	// generate build output
+	if s.executor.buildOutput != "" {
+		extractRootfsOpts := buildah.ExtractRootfsOptions{}
+		if unshare.IsRootless() {
+			// In order to maintain as much parity as possible
+			// with buildkit's version of --output and to avoid
+			// unsafe invocation of exported executables it was
+			// decided to strip setuid,setgid and extended attributes.
+			// Since modes like setuid,setgid leaves room for executable
+			// to get invoked with different file-system permission its safer
+			// to strip them off for unpriviledged invocation.
+			// See: https://github.com/containers/buildah/pull/3823#discussion_r829376633
+			extractRootfsOpts.StripSetuidBit = true
+			extractRootfsOpts.StripSetgidBit = true
+			extractRootfsOpts.StripXattrs = true
+		}
+		rc, errChan, err := s.builder.ExtractRootfs(options, extractRootfsOpts)
+		if err != nil {
+			return "", nil, errors.Wrapf(err, "failed to extract rootfs from given container image")
+		}
+		defer rc.Close()
+		err = internalUtil.ExportFromReader(rc, buildOutputOption)
+		if err != nil {
+			return "", nil, errors.Wrapf(err, "failed to export build output")
+		}
+		if errChan != nil {
+			err = <-errChan
+			if err != nil {
+				return "", nil, err
+			}
+		}
+
 	}
 	imgID, _, manifestDigest, err := s.builder.Commit(ctx, imageRef, options)
 	if err != nil {
--- a/internal/util/util.go
+++ b/internal/util/util.go
@@ -1,9 +1,18 @@
 package util
 
 import (
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/containers/buildah/define"
 	"github.com/containers/common/libimage"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage"
+	"github.com/containers/storage/pkg/archive"
+	"github.com/containers/storage/pkg/chrootarchive"
+	"github.com/containers/storage/pkg/unshare"
+	"github.com/pkg/errors"
 )
 
 // LookupImage returns *Image to corresponding imagename or id
@@ -21,4 +30,52 @@ func LookupImage(ctx *types.SystemContext, store storage.Store, image string) (*
 		return nil, err
 	}
 	return localImage, nil
+}
+
+// ExportFromReader reads bytes from given reader and exports to external tar, directory or stdout.
+func ExportFromReader(input io.Reader, opts define.BuildOutputOption) error {
+	var err error
+	if !filepath.IsAbs(opts.Path) {
+		opts.Path, err = filepath.Abs(opts.Path)
+		if err != nil {
+			return err
+		}
+	}
+	if opts.IsDir {
+		// In order to keep this feature as close as possible to
+		// buildkit it was decided to preserve ownership when
+		// invoked as root since caller already has access to artifacts
+		// therefore we can preserve ownership as is, however for rootless users
+		// ownership has to be changed so exported artifacts can still
+		// be accessible by unpriviledged users.
+		// See: https://github.com/containers/buildah/pull/3823#discussion_r829376633
+		noLChown := false
+		if unshare.IsRootless() {
+			noLChown = true
+		}
+
+		err = os.MkdirAll(opts.Path, 0700)
+		if err != nil {
+			return errors.Wrapf(err, "failed while creating the destination path %q", opts.Path)
+		}
+
+		err = chrootarchive.Untar(input, opts.Path, &archive.TarOptions{NoLchown: noLChown})
+		if err != nil {
+			return errors.Wrapf(err, "failed while performing untar at %q", opts.Path)
+		}
+	} else {
+		outFile := os.Stdout
+		if !opts.IsStdout {
+			outFile, err = os.Create(opts.Path)
+			if err != nil {
+				return errors.Wrapf(err, "failed while creating destination tar at %q", opts.Path)
+			}
+			defer outFile.Close()
+		}
+		_, err = io.Copy(outFile, input)
+		if err != nil {
+			return errors.Wrapf(err, "failed while performing copy to %q", opts.Path)
+		}
+	}
+	return nil
 }
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -510,6 +510,73 @@ func AuthConfig(creds string) (*types.DockerAuthConfig, error) {
 	}, nil
 }
 
+// GetBuildOutput is responsible for parsing custom build output argument i.e `build --output` flag.
+// Takes `buildOutput` as string and returns BuildOutputOption
+func GetBuildOutput(buildOutput string) (define.BuildOutputOption, error) {
+	if len(buildOutput) == 1 && buildOutput == "-" {
+		// Feature parity with buildkit, output tar to stdout
+		// Read more here: https://docs.docker.com/engine/reference/commandline/build/#custom-build-outputs
+		return define.BuildOutputOption{Path: "",
+			IsDir:    false,
+			IsStdout: true}, nil
+	}
+	if !strings.Contains(buildOutput, ",") {
+		// expect default --output <dirname>
+		return define.BuildOutputOption{Path: buildOutput,
+			IsDir:    true,
+			IsStdout: false}, nil
+	}
+	isDir := true
+	isStdout := false
+	typeSelected := false
+	pathSelected := false
+	path := ""
+	tokens := strings.Split(buildOutput, ",")
+	for _, option := range tokens {
+		arr := strings.SplitN(option, "=", 2)
+		if len(arr) != 2 {
+			return define.BuildOutputOption{}, fmt.Errorf("invalid build output options %q, expected format key=value", buildOutput)
+		}
+		switch arr[0] {
+		case "type":
+			if typeSelected {
+				return define.BuildOutputOption{}, fmt.Errorf("Duplicate %q not supported", arr[0])
+			}
+			typeSelected = true
+			if arr[1] == "local" {
+				isDir = true
+			} else if arr[1] == "tar" {
+				isDir = false
+			} else {
+				return define.BuildOutputOption{}, fmt.Errorf("invalid type %q selected for build output options %q", arr[1], buildOutput)
+			}
+		case "dest":
+			if pathSelected {
+				return define.BuildOutputOption{}, fmt.Errorf("Duplicate %q not supported", arr[0])
+			}
+			pathSelected = true
+			path = arr[1]
+		default:
+			return define.BuildOutputOption{}, fmt.Errorf("Unrecognized key %q in build output option: %q", arr[0], buildOutput)
+		}
+	}
+
+	if !typeSelected || !pathSelected {
+		return define.BuildOutputOption{}, fmt.Errorf("invalid build output option %q, accepted keys are type and dest must be present", buildOutput)
+	}
+
+	if path == "-" {
+		if isDir {
+			return define.BuildOutputOption{}, fmt.Errorf("invalid build output option %q, type=local and dest=- is not supported", buildOutput)
+		}
+		return define.BuildOutputOption{Path: "",
+			IsDir:    false,
+			IsStdout: true}, nil
+	}
+
+	return define.BuildOutputOption{Path: path, IsDir: isDir, IsStdout: isStdout}, nil
+}
+
 // IDMappingOptions parses the build options related to user namespaces and ID mapping.
 func IDMappingOptions(c *cobra.Command, isolation define.Isolation) (usernsOptions define.NamespaceOptions, idmapOptions *define.IDMappingOptions, err error) {
 	return IDMappingOptionsFromFlagSet(c.Flags(), c.PersistentFlags(), c.Flag)
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -551,6 +551,67 @@ _EOF
   expect_output "[container=buildah date=tomorrow]" "No Path should be defined"
 }
 
+@test "build with custom build output and output rootfs to directory" {
+  _prefetch alpine
+  mytmpdir=${TEST_SCRATCH_DIR}/my-dir
+  mkdir -p $mytmpdir
+  cat > $mytmpdir/Containerfile << _EOF
+FROM alpine
+RUN echo 'hello'> hello
+_EOF
+  run_buildah build --output type=local,dest=$mytmpdir/rootfs $WITH_POLICY_JSON -t test-bud -f $mytmpdir/Containerfile .
+  ls $mytmpdir/rootfs
+  # exported rootfs must contain `hello` file which we created inside the image
+  expect_output --substring 'hello'
+}
+
+@test "build with custom build output and output rootfs to tar" {
+  _prefetch alpine
+  mytmpdir=${TEST_SCRATCH_DIR}/my-dir
+  mkdir -p $mytmpdir
+  cat > $mytmpdir/Containerfile << _EOF
+FROM alpine
+RUN echo 'hello'> hello
+_EOF
+  run_buildah build --output type=tar,dest=$mytmpdir/rootfs.tar $WITH_POLICY_JSON -t test-bud -f $mytmpdir/Containerfile .
+  # explode tar
+  mkdir $mytmpdir/rootfs
+  tar -C $mytmpdir/rootfs -xvf $mytmpdir/rootfs.tar
+  ls $mytmpdir/rootfs
+  # exported rootfs must contain `hello` file which we created inside the image
+  expect_output --substring 'hello'
+}
+
+@test "build with custom build output and output rootfs to tar by pipe" {
+  _prefetch alpine
+  mytmpdir=${TEST_SCRATCH_DIR}/my-dir
+  mkdir -p $mytmpdir
+  cat > $mytmpdir/Containerfile << _EOF
+FROM alpine
+RUN echo 'hello'> hello
+_EOF
+  # Using BUILDAH_BINARY since run_buildah adds unwanted chars to tar created by pipe.
+  ${BUILDAH_BINARY} build $WITH_POLICY_JSON -o - -t test-bud -f $mytmpdir/Containerfile . > $mytmpdir/rootfs.tar
+  # explode tar
+  mkdir $mytmpdir/rootfs
+  tar -C $mytmpdir/rootfs -xvf $mytmpdir/rootfs.tar
+  ls $mytmpdir/rootfs/hello
+}
+
+@test "build with custom build output must fail for bad input" {
+  _prefetch alpine
+  mytmpdir=${TEST_SCRATCH_DIR}/my-dir
+  mkdir -p $mytmpdir
+  cat > $mytmpdir/Containerfile << _EOF
+FROM alpine
+RUN echo 'hello'> hello
+_EOF
+  run_buildah 125 build --output type=tar, $WITH_POLICY_JSON -t test-bud -f $mytmpdir/Containerfile .
+  expect_output --substring 'invalid'
+  run_buildah 125 build --output type=wrong,dest=hello --signature-policy ${TESTSDIR}/policy.json -t test-bud -f $mytmpdir/Containerfile .
+  expect_output --substring 'invalid'
+}
+
 @test "bud-from-scratch-untagged" {
   run_buildah build --iidfile ${TEST_SCRATCH_DIR}/output.iid $WITH_POLICY_JSON $BUDFILES/from-scratch
   iid=$(cat ${TEST_SCRATCH_DIR}/output.iid)
--- a/commit.go
+++ b/commit.go
@@ -229,6 +229,7 @@ func (b *Builder) addManifest(ctx context.Context, manifestName string, imageSpe
 func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options CommitOptions) (string, reference.Canonical, digest.Digest, error) {
 	var (
 		imgID string
+		src   types.ImageReference
 	)
 
 	// If we weren't given a name, build a destination reference using a
@@ -300,7 +301,7 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 	logrus.Debugf("committing image with reference %q is allowed by policy", transports.ImageName(dest))
 
 	// Build an image reference from which we can copy the finished image.
-	src, err := b.makeImageRef(options)
+	src, err = b.makeContainerImageRef(options)
 	if err != nil {
 		return imgID, nil, "", errors.Wrapf(err, "error computing layer digests and building metadata for container %q", b.ContainerID)
 	}
--- a/define/build.go
+++ b/define/build.go
@@ -123,6 +123,10 @@ type BuildOptions struct {
 	Args map[string]string
 	// Name of the image to write to.
 	Output string
+	// BuildOutput specifies if any custom build output is selected for following build.
+	// It allows end user to export recently built rootfs into a directory or tar.
+	// See the documentation of 'buildah build --output' for the details of the format.
+	BuildOutput string
 	// Additional tags to add to the image that we write, if we know of a
 	// way to add them.
 	AdditionalTags []string
--- a/define/types.go
+++ b/define/types.go
@@ -96,6 +96,13 @@ type Secret struct {
 	SourceType string
 }
 
+// BuildOutputOptions contains the the outcome of parsing the value of a build --output flag
+type BuildOutputOption struct {
+	Path     string // Only valid if !IsStdout
+	IsDir    bool
+	IsStdout bool
+}
+
 // TempDirForURL checks if the passed-in string looks like a URL or -.  If it is,
 // TempDirForURL creates a temporary directory, arranges for its contents to be
 // the contents of that URL, and returns the temporary directory's path, along
--- a/pkg/cli/common.go
+++ b/pkg/cli/common.go
@@ -85,6 +85,7 @@ type BudResults struct {
 	Squash              bool
 	Stdin               bool
 	Tag                 []string
+	BuildOutput         string
 	Target              string
 	TLSVerify           bool
 	Jobs                int
@@ -242,6 +243,7 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
 	fs.StringArrayVar(&flags.SSH, "ssh", []string{}, "SSH agent socket or keys to expose to the build. (format: default|<id>[=<socket>|<key>[,<key>]])")
 	fs.BoolVar(&flags.Stdin, "stdin", false, "pass stdin into containers")
 	fs.StringArrayVarP(&flags.Tag, "tag", "t", []string{}, "tagged `name` to apply to the built image")
+	fs.StringVarP(&flags.BuildOutput, "output", "o", "", "output destination (format: type=local,dest=path)")
 	fs.StringVar(&flags.Target, "target", "", "set the target build stage to build")
 	fs.Int64Var(&flags.Timestamp, "timestamp", 0, "set created timestamp to the specified epoch seconds to allow for deterministic builds, defaults to current time")
 	fs.BoolVar(&flags.TLSVerify, "tls-verify", true, "require HTTPS and verify certificates when accessing the registry")
@@ -281,6 +283,7 @@ func GetBudFlagsCompletions() commonComp.FlagCompletions {
 	flagCompletion["timestamp"] = commonComp.AutocompleteNone
 	flagCompletion["variant"] = commonComp.AutocompleteNone
 	flagCompletion["unsetenv"] = commonComp.AutocompleteNone
+	flagCompletion["output"] = commonComp.AutocompleteNone
 	return flagCompletion
 }
 


docs/buildah-build.1.md
```diff
@@ -407,6 +407,14 @@ By default, Buildah manages _/etc/hosts_, adding the container's own IP address.
 
 Set the OS of the image to be built, and that of the base image to be pulled, if the build uses one, instead of using the current operating system of the host.
 
+**--output**, **-o**=""
+
+Output destination (format: type=local,dest=path)
+
+By default, a local container image is created from the build result. The --output (or -o) flag allows you to override this behavior, and a specify a custom exporter. For example, custom exporters allow you to export the build artifacts as files on the local filesystem instead of a Container image, which can be useful for generating local binaries, code generation etc.
```
TomSweeneyRedHat, line 414:
```suggestion
By default, a local container image is created from the build result. The --output (or -o) option allows you to override this behavior and specify a custom exporter. For example, custom exporters allow you to export the build artifacts as files on the local filesystem instead of a Container image, which can be useful for generating local binaries, code generation, etc.
```

docs/buildah-build.1.md
```diff
@@ -407,6 +407,14 @@ By default, Buildah manages _/etc/hosts_, adding the container's own IP address.
...
+
+The value for --output is a CSV-formatted string defining the exporter type and options. Currently, local and tar exporters are supported. The local exporter writes the resulting build files to a directory on the client side. The tar exporter is similar but writes the files as a single tarball (.tar).
```
TomSweeneyRedHat, line 416:
```suggestion
The value for --output is a CSV-formatted string defining the exporter type and options. 

Valid _type_ values are:
- **local**: write the resulting build files to a directory on the client-side.
- **tar**: write the resulting files as a single tarball (.tar).
```

TomSweeneyRedHat: An example or two would be helpful.  Looks like a good quick blog post once complete.

tests/bud.bats
```diff
@@ -545,6 +545,51 @@ _EOF
...
+  mkdir -p $mytmpdir
+  cat > $mytmpdir/Containerfile << _EOF
+FROM alpine
+RUN echo 'hello'> hello
+_EOF
+  run_buildah 125 build --output type=tar, --signature-policy ${TESTSDIR}/policy.json -t test-bud -f $mytmpdir/Containerfile .
+  expect_output --substring 'invalid'
+  run_buildah 125 build --output type=wrong,dest=hello --signature-policy ${TESTSDIR}/policy.json -t test-bud -f $mytmpdir/Containerfile .
+  expect_output --substring 'invalid'
+}
```
TomSweeneyRedHat, line 591: Can you use the information from either file to build the image?  If so, could you add a test for that?

docs/buildah-build.1.md
TomSweeneyRedHat: do you handle a hyphen `-` with the tar option to write the output to stdout?  You might want to add that to the man page if so.

flouthoc: Yes we handle `-` , added to docs with example.

docs/buildah-build.1.md
flouthoc: fixed

tests/bud.bats
flouthoc: Actually its equivalent to a `rootfs` on file-system so command does not supports this directly but user can always do something like, but its not supported by `buildah` and `buildkit`.
 ```Dockerfile
FROM scratch
ADD rootfs/ /
```
or Podman and Docker could directly use `import` instruction.

File: docs/buildah-build.1.md [hunk 4]
```diff
@@ -407,6 +407,19 @@ By default, Buildah manages _/etc/hosts_, adding the container's own IP address.
 
 Set the OS of the image to be built, and that of the base image to be pulled, if the build uses one, instead of using the current operating system of the host.
 
+**--output**, **-o**=""
+
+Output destination (format: type=local,dest=path)
+
+By default, a local container image is created from the build result. The --output (or -o) option allows you to override this behavior and specify a custom exporter. For example, custom exporters allow you to export the build artifacts as files on the local filesystem instead of a Container image, which can be useful for generating local binaries, code generation, etc.
```
nalind, line 414: No need to capitalize container here.

imagebuildah/executor.go
```diff
@@ -352,6 +357,67 @@ func (b *Executor) resolveNameToImageRef(output string) (types.ImageReference, e
...
+		didMount = true
+		mountPoint, err = image.Mount(ctx, nil, "")
+		if err != nil {
+			return errors.Wrapf(err, "failed mounting image %q", imageID)
+		}
+	}
+	input, err := archive.Tar(mountPoint, archive.Uncompressed)
+	if err != nil {
+		return errors.Wrapf(err, "failed while performing tar for %q", mountPoint)
+	}
```
nalind, line 391: `input` is a `ReadCloser`, so it needs to be closed.

imagebuildah/executor.go
```diff
@@ -352,6 +357,67 @@ func (b *Executor) resolveNameToImageRef(output string) (types.ImageReference, e
...
+	if isDir {
+		if !filepath.IsAbs(path) {
+			cwd, err := os.Getwd()
+			if err != nil {
+				return err
+			}
+			path = filepath.Join(cwd, filepath.Clean(path))
+		}
+		err = archive.Untar(input, path, nil)
```
nalind, line 400: How are ownerships intended to work for unprivileged users, where this happens inside of a user namespace?

flouthoc: I'm expecting ownership to stay as it is which was set in image and it should not be changed for unprivileged users so image out can become read-only if files where meant to be written for some other user. 

But I'll verify behavior once with rootless buildkit as well just to stay in parity.

flouthoc:
> This is the sort of thing that we've historically done in the image library and then exposed via the `-t` flag, as copying images is a big part of what the image library does, and I could definitely see a `-o` implementation leaning on that as well. There's a "tarball" `ImageTransport`, that currently only implements an `ImageSource` that imports rootfs tarballs, that this would specifically appear to complement. The current PR isn't implemented that way, and that limits our ability to reuse this functionality anywhere else.

@nalind I agree with this, actually i thought of it but i was reluctant of moving this function to `c/image` cause I was not able to think of lot of use-cases outside of the build output generation. Since it looked like a buildah local to me hence i opted to keep it here but we can move this function to `c/image` in future if more cases show up or now as well whatever sounds best.

TomSweeneyRedHat:
If you think there's a chance you'd want to move some of this code in the future to c/image, I'd do it now.  Far less pain.  Regardless, I'd move it based on Nalin's comments.

rhatdan:
@vrothberg @mtrmac Thoughts?

docs/buildah-build.1.md
```diff
@@ -407,6 +407,19 @@ By default, Buildah manages _/etc/hosts_, adding the container's own IP address.
 
 Set the OS of the image to be built, and that of the base image to be pulled, if the build uses one, instead of using the current operating system of the host.
 
+**--output**, **-o**=""
+
+Output destination (format: type=local,dest=path)
+
+By default, a local container image is created from the build result. The --output (or -o) option allows you to override this behavior and specify a custom exporter. For example, custom exporters allow you to export the build artifacts as files on the local filesystem instead of a container image, which can be useful for generating local binaries, code generation, etc.
+
+The value for --output is a CSV-formatted string defining the exporter type and options.
```
mtrmac, line 416: CSV is a specific spreadsheet format. This is not quite that.

The dest= option is only documented with an example… maybe go all the way and say that it is a comma-separated sequence of key=value pairs, and make a list/table of the supported keys.

docs/buildah-build.1.md
```diff
@@ -820,6 +833,16 @@ buildah bud --platform linux/arm64 --platform linux/amd64 --manifest myimage /tm
 
 buildah bud --all-platforms --manifest myimage /tmp/mysrc
 
+### Building an image using (--output) custom build output
+
+buildah build -o out .
```
mtrmac, line 838: (Personal rant: Ugh, so the syntax of --output is a set of key-value pairs, _except when it’s not_? I hate ambiguities.)

imagebuildah/executor.go
```diff
@@ -352,6 +357,68 @@ func (b *Executor) resolveNameToImageRef(output string) (types.ImageReference, e
...
+	}
+	defer input.Close()
+	if isDir {
+		if !filepath.IsAbs(path) {
+			cwd, err := os.Getwd()
+			if err != nil {
+				return err
+			}
+			path = filepath.Join(cwd, filepath.Clean(path))
+		}
```
mtrmac, line 400: Wouldn’t it make sense to do this sometime early, during option processing or setup, so that we don’t do all the build work only for the saving to fail on a trivial path error?

internal/parse/parse.go
```diff
@@ -37,6 +37,56 @@ var (
...
+	pathSelected := false
+	path := ""
+	if len(buildOutput) == 1 && buildOutput == "-" {
+		// Feature parity with buildkit, output tar to stdout
+		// Read more here: https://docs.docker.com/engine/reference/commandline/build/#custom-build-outputs
+		return "", false, true, nil
+	}
+	if strings.Contains(buildOutput, ",") {
+		tokens := strings.SplitN(buildOutput, ",", 2)
+		if len(tokens) != 2 {
```
mtrmac, line 55: Is this somehow a fundamental design constraint? A compatibility constraint? The format at seems to support an arbitrary number of options just fine.

If this is to support a `,` character inside a path, please explicitly document that in a comment here (and perhaps even in the man page). (But I’m not sure about that idea, it limits us from adding more option types in the future.)

internal/parse/parse.go
```diff
@@ -37,6 +37,56 @@ var (
...
+				if arr[1] == "local" {
+					isDir = true
+				} else if arr[1] == "tar" {
+					isDir = false
+				} else {
+					return path, isDir, isStdout, fmt.Errorf("invalid type %q selected for build output options %q", arr[1], buildOutput)
+				}
+			}
+			if arr[0] == "dest" && !pathSelected {
+				pathSelected = true
```
mtrmac, line 74: It would be helpful for this to loudly complain about duplicate type/dest keys, instead of silently ignoring data.

imagebuildah/executor.go
```diff
@@ -776,6 +843,17 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 		return "", nil, err
 	}
 	logrus.Debugf("printing final image id %q", imageID)
+	if b.buildOutput != "" {
+		logrus.Debugf("Generating custom build output with options %q", b.buildOutput)
+		path, isDir, isStdout, err := internalParse.GetBuildOutput(b.buildOutput)
```
mtrmac, line 848: Wouldn’t it be better to parse all of this (and report errors) before starting to actually build anything?

(And then, the result of `GetBuildOutput` should probably be an object that can just be passed on as a single unit, not a set of loose independent variables the caller needs to manage one at a time.)

mtrmac:
> This is the sort of thing that we've historically done in the image library and then exposed via the `-t` flag, as copying images is a big part of what the image library does, and I could definitely see a `-o` implementation leaning on that as well.
> There's a "tarball" `ImageTransport`, that currently only implements an `ImageSource` that imports rootfs tarballs, that this would specifically appear to complement. The current PR isn't implemented that way, and that limits our ability to reuse this functionality anywhere else.

I definitely see an elegance in turning this option into just a “commit” to a special transport — because various options could be reused, and the Buildah code could be simpler.

OTOH there are also downsides:
- It would also allow the user to specify the other commit options/flags/features that modify the config, while the config is not produced by this tarball output. It would be extra code to reject all of that (… admittedly this PR should _also_ be rejecting all of those options with `--output` (unless `-t` is used?)).
- The c/image references are a bit stateless, making it somewhat ugly to create a reference “to stdout”. Buildah is already dealing with that for the tarball input format, OTOH, so that would not be a new concern.
- Note that c/image really has no concept of layer _contents_; and no facility for layer editing. I.e. to output a single tarball, it would have to be Buildah (via it’s `CommitOptions.Squash` code that would turn a multi-layer image into a tarball content.  (Alternatively, the tarball transport’s destination could reimplement all of that, by extracting the tarballs, implementing whiteouts, and then creating a tarball out of the result, but that would be a lot of unwarranted complexity at probably much worse performance than `extractRootfs` doing a `Mount`). Essentially, all the c/image tarball transport would do is to stream the Buildah-created tarball from a Buildah-created temporary file (one that is necessary to implement `containerImageRef.NewImageSource`) to the specified destination, hopefully without introducing a _second_ temporary file; while the current special-cased implementation can stream data from `archive.Tar` directly. So using a tarball transport would be an extra cost.

---

Overall I don’t really have a strong opinion about using a c/image transport.

I do, however, feel somewhat strongly (unjustifiably strongly for a Buildah outsider, perhaps) that the code to produce these tarballs should be shared with the code that produces tarballs in squashed images (at least including `containerImageRef.extractRootfs` — see how it differs from this PR’s implementation — possibly more). At least the “producer” half of `buildCustomOutput` should not exist IMHO.

Turning the `--output` option into just a transport reference would clearly be a way to do that. Doing it some other way, to preserve the streaming implementation without temporary files, might well be worthwhile — it’s a code size / long-term maintenance effort vs. performance tradeoff.

docs/buildah-build.1.md
flouthoc: fixed.

imagebuildah/executor.go
flouthoc: yes makes sense to me, fixed in latest commit.

docs/buildah-build.1.md
flouthoc: this is just to match the short-hand syntax which `docker buildx` supports.

internal/parse/parse.go
flouthoc: We are now erroring out for every unknown option

internal/parse/parse.go
flouthoc: yes added a check for duplicates.

imagebuildah/executor.go
flouthoc: This is a nice idea, added in latest commit.

flouthoc:
@mtrmac I am fine to share as much code as possible with https://github.com/containers/buildah/blob/main/image.go#L153 `extractRootfs`. @nalind WDYT ?

internal/parse/parse.go
```diff
@@ -37,6 +37,70 @@ var (
...
+				typeSelected = true
+				if arr[1] == "local" {
+					isDir = true
+				} else if arr[1] == "tar" {
+					isDir = false
+				} else {
+					return BuildOutputOption{path, isDir, isStdout}, fmt.Errorf("invalid type %q selected for build output options %q", arr[1], buildOutput)
+				}
+			} else if arr[0] == "type" && pathSelected {
+				return BuildOutputOption{path, isDir, isStdout}, fmt.Errorf("Duplicate %q not supported", arr[0])
```
mtrmac, line 80: `"type" && pathSelected` seems incorrect.

internal/parse/parse.go
```diff
@@ -37,6 +37,70 @@ var (
...
+			}
+			if arr[0] == "dest" && !pathSelected {
+				pathSelected = true
+				path = arr[1]
+			} else if arr[0] == "dest" && pathSelected {
+				return BuildOutputOption{path, isDir, isStdout}, fmt.Errorf("Duplicate %q not supported", arr[0])
+			}
+			if !typeSelected && !pathSelected {
```
mtrmac, line 88: This works but it’s not quite obvious that it is the “unknown option name” requires maintaining this list of *Selected variables.

Also the two separate checks for `arr[0] == $value` seem unnecessary.

How about
```go
switch arr[0] {
case "type":
    if typeSelected { fail }
    typeSelected = true
    …
case "dest":
    if pathSelected { fail }
    pathSelected = true
    …
default:
    fail
}
```

or even
```go
// fieldsSeen := set<string> , I’m too lazy to use actual Go here. I guess a map[string]struct{}.
field := arr[0]
if field in fieldsSeen { fail }
fieldsSeen.Add(field)
switch field {
case "type":
    …
case "dest":
    …
default:
    fail
}
```

internal/parse/parse.go
```diff
@@ -37,6 +37,70 @@ var (
...
+	}
+	if strings.Contains(buildOutput, ",") {
+		tokens := strings.SplitN(buildOutput, ",", 2)
+		if len(tokens) != 2 {
+			return BuildOutputOption{path, isDir, isStdout}, fmt.Errorf("invalid build output options %q", buildOutput)
+		}
+		for _, option := range tokens {
+			arr := strings.SplitN(option, "=", 2)
+			if len(arr) != 2 {
+				return BuildOutputOption{path, isDir, isStdout}, fmt.Errorf("invalid build output options %q", buildOutput)
```
mtrmac, line 68: I wouldn’t bother with filling in the `BuildOutputOption` values on error paths — the values are not expected to be used, and it only makes the `return`s look like success cases at a first glance.

internal/parse/parse.go
mtrmac: That’s not quite what I meant.

With this, `type=tar,dest=/tmp/with,option=comma` parses the path as `/tmp/with,option=comma`, which is both undocumented and making it impossible for us to add any more options (after a `dest` option, or ever? undocumented behavior).

I’d suggest using SplitN(…3) at least, to forbid commas in values, or preferably just using `Split()`, removing this length check completely, and relying on the duplicate checks, and on after-the-loop checks to verify all data has been provided.

internal/parse/parse.go
```diff
@@ -37,6 +37,70 @@ var (
...
+			} else if arr[0] == "dest" && pathSelected {
+				return BuildOutputOption{path, isDir, isStdout}, fmt.Errorf("Duplicate %q not supported", arr[0])
+			}
+			if !typeSelected && !pathSelected {
+				return BuildOutputOption{path, isDir, isStdout}, fmt.Errorf("invalid build output options %q", buildOutput)
+			}
+		}
+	} else {
+		// expect default --output <dirname>
+		return BuildOutputOption{buildOutput, isDir, isStdout}, nil
```
mtrmac, line 94: Please move this short to be the first case in the if, much closer to the condition.  It can even be an early return and then the comma-separated syntax does not need to be indented — both the `if` above _and_ the code below only apply to the comma-separated syntax, and it is unclear why they should be indented differently and split into two parts with this intermediate code.

internal/parse/parse.go
```diff
@@ -37,6 +37,70 @@ var (
...
+				return BuildOutputOption{path, isDir, isStdout}, fmt.Errorf("Duplicate %q not supported", arr[0])
+			}
+			if arr[0] == "dest" && !pathSelected {
+				pathSelected = true
+				path = arr[1]
+			} else if arr[0] == "dest" && pathSelected {
+				return BuildOutputOption{path, isDir, isStdout}, fmt.Errorf("Duplicate %q not supported", arr[0])
+			}
+			if !typeSelected && !pathSelected {
+				return BuildOutputOption{path, isDir, isStdout}, fmt.Errorf("invalid build output options %q", buildOutput)
```
mtrmac, line 89: Please tell the user about what is wrong about the input.

internal/parse/parse.go
```diff
@@ -37,6 +37,70 @@ var (
...
+			}
+		}
+	} else {
+		// expect default --output <dirname>
+		return BuildOutputOption{buildOutput, isDir, isStdout}, nil
+	}
+
+	if !typeSelected || !pathSelected {
+		return BuildOutputOption{path, isDir, isStdout}, fmt.Errorf("invalid build output options %q", buildOutput)
```
mtrmac, line 98: Please tell the user about what is wrong about the input.

cmd/buildah/build.go
```diff
@@ -371,6 +378,7 @@ func buildCmd(c *cobra.Command, inputArgs []string, iopts buildOptions) error {
 		OS:                      systemContext.OSChoice,
 		Out:                     stdout,
 		Output:                  output,
+		BuildOutput:             iopts.BuildOutput,
```
mtrmac, line 381: Doing the parsing in `GetBuildOutput` above is nice; actually providing the result of the parsing here, instead of forcing the caller to parse again, would be cleaner. It avoids unreachable error cases in the ultimate implementation, and makes it much less likely to confuse the string syntax and the parsed data.

rhatdan:
@flouthoc Any progress on this?

flouthoc:
I'll fix the requested nits, thanks @mtrmac

rhatdan:
@flouthoc What is going on with this one?

flouthoc:
Picking this up again, i don't know how i missed this. :\

internal/parse/parse.go
flouthoc: fixed.

internal/parse/parse.go
flouthoc: fixed

internal/parse/parse.go
flouthoc: fixed

internal/parse/parse.go
flouthoc: fixed

internal/parse/parse.go
flouthoc: fixed.

imagebuildah/executor.go
```diff
@@ -776,6 +843,17 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 		return "", nil, err
 	}
 	logrus.Debugf("printing final image id %q", imageID)
+	if b.buildOutput != "" {
```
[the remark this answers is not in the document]
flouthoc: Assuming I understood this comment correctly, docker allows `-t` and `-o` to be functional at same time independently and `--output` is functional even if `-t` is not specified so just keeping in parity with that.

Please unresolve this part if I understood this comment wrong.

flouthoc:
@mtrmac @nalind Addressed all feedbacks and also switched to a new approach after discussion above https://github.com/containers/buildah/pull/3823#issuecomment-1073546133 we are now reusing `containerImageRef.extractRootfs` https://github.com/containers/buildah/blob/main/image.go#L153 and no new implementation is added for extracting `Rootfs`

PTAL

rhatdan:
@mtrmac @nalind PTAL
Would like to get this in before the beta.

cmd/buildah/build.go
```diff
@@ -328,6 +329,12 @@ func buildCmd(c *cobra.Command, inputArgs []string, iopts buildOptions) error {
 		t := time.Unix(iopts.Timestamp, 0).UTC()
 		timestamp = &t
 	}
+	if c.Flag("output").Changed {
+		buildOption, _ := internalParse.GetBuildOutput(iopts.BuildOutput)
```
nalind, line 333: This needs to handle errors.

docs/buildah-build.1.md
```diff
@@ -411,6 +411,24 @@ By default, Buildah manages _/etc/hosts_, adding the container's own IP address.
 
 Set the OS of the image to be built, and that of the base image to be pulled, if the build uses one, instead of using the current operating system of the host.
 
+**--output**, **-o**=""
+
+Output destination (format: type=local,dest=path)
+
+By default, a local container image is created from the build result. The --output (or -o) option allows you to override this behavior and specify a custom exporter. For example, custom exporters allow you to export the build artifacts as files on the local filesystem instead of a container image, which can be useful for generating local binaries, code generation, etc.
```
nalind, line 418: This suggests that when this option is used, we don't commit a container image, but it looks like we still do.

image.go
```diff
@@ -813,3 +813,7 @@ func (b *Builder) makeImageRef(options CommitOptions) (types.ImageReference, err
 	}
 	return ref, nil
 }
+
+func (b *Builder) MakeImageRef(options CommitOptions) (types.ImageReference, error) {
```
nalind, line 817: This method doesn't need to exist.  Just have `Commit()` call `MakeContainerImageRef()` directly.

internal/parse/parse.go
```diff
@@ -41,6 +41,67 @@ var (
...
+	pathSelected := false
+	path := ""
+	if len(buildOutput) == 1 && buildOutput == "-" {
+		// Feature parity with buildkit, output tar to stdout
+		// Read more here: https://docs.docker.com/engine/reference/commandline/build/#custom-build-outputs
+		return BuildOutputOption{"", false, true}, nil
+	}
+	if !strings.Contains(buildOutput, ",") {
+		// expect default --output <dirname>
+		return BuildOutputOption{buildOutput, isDir, isStdout}, nil
```
nalind, line 66: Might as well name the fields in the initializers and use constants for the values, as above.

internal/parse/parse.go
```diff
@@ -41,6 +41,67 @@ var (
...
+	}
+	tokens := strings.Split(buildOutput, ",")
+	for _, option := range tokens {
+		arr := strings.SplitN(option, "=", 2)
+		if len(arr) != 2 {
+			return BuildOutputOption{}, fmt.Errorf("invalid build output options %q", buildOutput)
```
nalind, line 72: We know exactly why we're rejecting the provided value here, so the error message doesn't need to be that vague.

internal/parse/parse.go
```diff
@@ -41,6 +41,67 @@ var (
...
+				return BuildOutputOption{}, fmt.Errorf("invalid type %q selected for build output options %q", arr[1], buildOutput)
+			}
+		case "dest":
+			if pathSelected {
+				return BuildOutputOption{}, fmt.Errorf("Duplicate %q not supported", arr[0])
+			}
+			pathSelected = true
+			path = arr[1]
+		default:
+			return BuildOutputOption{}, fmt.Errorf("invalid build output options %q", buildOutput)
```
nalind, line 94: Likewise, we know `arr[0]` is not a recognized key for this option, as below.

internal/util/util.go
```diff
@@ -22,3 +28,35 @@ func LookupImage(ctx *types.SystemContext, store storage.Store, image string) (*
...
+	var err error
+	if !filepath.IsAbs(path) {
+		cwd, err := os.Getwd()
+		if err != nil {
+			return err
+		}
+		path = filepath.Join(cwd, filepath.Clean(path))
+	}
+	if isDir {
+		err = archive.Untar(input, path, nil)
```
nalind, line 43: Use `chrootarchive.Untar()` here if at all possible.

docs/buildah-build.1.md
flouthoc: Fixed explicitly mentioned.

image.go
flouthoc: `Commit(` is expecting type `types.ImageReference` but `MakeContainerImageRef` returns `*ContainerImageRef`, function is just helping me avoid recasting in the code but I'm can change this if there is better casting method.

flouthoc: this is fixed.

flouthoc:
Fixed all the nits @rhatdan @nalind

docs/buildah-build.1.md
nalind: This still reads to me as though using this option causes the builder to not commit an image, as committing the image is the default behavior that's being overridden when this option is specified.

imagebuildah/executor.go
nalind: This would need to be verified.

nalind:
How will `podman-remote` be able to wire this up?

flouthoc:
@nalind I think it can be used by `podman-remote` only when output is exported on shared-mount path with the host.

docs/buildah-build.1.md
flouthoc: fixed it after confirming from @nalind hence marking resolved again.

docs/buildah-build.1.md
nalind: The integration test should probably check that we don't unintentionally start rejecting this variation.

nalind:
> @nalind I think it can be used by `podman-remote` only when output is exported on shared-mount path with the host. Initially this can be `NO-OP` for remote but can be enabled when `host`->`machine` sharing starts working on `macOS`.

Just so long as this doesn't lead to letting people who are accessing a server that only offers them the podman API using this to overwrite configuration files on the host.

rhatdan:
Probably should be rejected for now.

flouthoc:
> Just so long as this doesn't lead to letting people who are accessing a server that only offers them the podman API using this to overwrite configuration files on the host.

Yeah this might be not possible since there is no way to transfer files from server to host so its a `no-op` for `podman-remote` as of now so this scenario is not possible

> Probably should be rejected for now.

@rhatdan Do you mean putting this as `no-op` for `podman-remote` or getting this feature into `podman` only when it works for remote ?

flouthoc:
@rhatdan @nalind resolved all nits PTAL. Just `podman-remote` discussion is remaining but I think it can be `no-op` for now and we can discuss it podman side.

rhatdan:
Yes for podman-remote, I think we should reject it until we have a solid solution for it.  The problem with that is on a MAC and Windows.  We can discuss this at Water Cooler, to make us all understand the issue.

rhatdan:
LGTM

imagebuildah/executor.go
nalind: What does that mean?  When we write content to a directory, who ends up owning files that are and aren't owned by root in the image?  Are setuid/setgid bits and extended attributes preserved?

imagebuildah/stage_executor.go
```diff
@@ -1556,6 +1568,23 @@ func (s *StageExecutor) commit(ctx context.Context, createdBy string, emptyLayer
 		Manifest:              s.executor.manifest,
 		UnsetEnvs:             s.executor.unsetEnvs,
 	}
+	// generate build output
+	if s.executor.buildOutput != "" {
+		buildOutputOptions := options
+		src, err := s.builder.MakeContainerImageRef(buildOutputOptions)
```
nalind, line 1574: What's the temporary variable for?

imagebuildah/executor.go
flouthoc: So when i export content with buildkit ownership does not stays as is everything is changing to current user so for example if I do 

```Dockerfile
FROM alpine
COPY hello .
RUN chown 1000:1000 hello
```

Exported content here has ownership of `hello` changed to `root:root` on host fs which I expected not to happen and stayed as `1000:1000` so just to match the behviour I added `NoLchown: true`.

flouthoc: I have a hunch that either buildkit is not doing something right here and we are just trying to match the parity or maybe it is done intentionally to match some practical use-case.

flouthoc: @nalind Following behavior also states true for `unprivileged user` an easier test is to run `buildkit:master-rootless` image.
So a reproducer is something like
```console
docker run --name buildkitd -it --privileged moby/buildkit:master-rootless
docker exec -it buildkitd sh
whoami
# should be `user` and not `root`
cd ~
touch hello
```

Inside container create containerfile

```Dockerfile
FROM alpine
COPY hello .
RUN chown 1000:1000 hello
```
and run
```
buildctl build --frontend=dockerfile.v0  --local context=. --local dockerfile=. --output type=local,dest=output-dir`
```
all the files inside `output-dir` have been chowned to user `user:user`.


Currently buildah follows this exact same behavior.

nalind: Are setuid/setgid bits and extended attributes preserved?

flouthoc: @nalind Yes both are preserved in both buildkit and buildah verified manually with `setfattr` and `getfattr` after unpacking.

flouthoc: Just adding a note here with a reproducer shared by @nalind , `buildkit` has inconsistent and not documented behavior between `privileged` and `unprivileged` invocation and permissions are produced differently for binaries like `/usr/bin/write` in both invocation.

So something like 
```Dockerfile
FROM registry.fedoraproject.org/fedora:35
RUN rpm --setcaps shadow-utils
RUN dnf -y install attr util-linux && dnf clean all
RUN chmod u+rwx `rpm -ql filesystem` 2> /dev/null || true
RUN echo test > /test.txt && chown 1000:1000 /test.txt
RUN getcap -v /usr/bin/newuidmap /usr/bin/newgidmap ; stat -c "%A   %n" /test.txt /usr/bin/su /usr/bin/write /usr/bin/sudo
``` 

With buildkit exported output
```
#root
-rwxr-sr-x usr/bin/write
# unprivileged user
-rwxr-xr-x usr/bin/write
```

With buildah we share permissions similarly between invocation by privileged and unprivileged user 
```
#root
-rwxr-sr-x   usr/bin/write
# unpriviledged user
-rwxr-sr-x   usr/bin/write
```

Here behavior of buildah is different from buildkit a little bit but I was unable to find a reason why should we alter permission for unprivileged user hence leaving a note here so in future if we ever face this we can revisit this discussion.

flouthoc:
We found one discrepancy in buildkit's behavior and its discussed and documented here: https://github.com/containers/buildah/pull/3823#discussion_r853821897, as of now I am not sure why that discrepancy exists and i'm unable to think of a reason to have that in buildah. I think we can keep that comment open and I'll address that in future in a followup PR if that becomes a problem.

internal/util/util.go
```diff
@@ -22,3 +29,35 @@ func LookupImage(ctx *types.SystemContext, store storage.Store, image string) (*
...
+
+// ExportFromReader reads bytes from given reader and exports to external tar, directory or stdout.
+func ExportFromReader(input io.Reader, isDir bool, isStdout bool, path string) error {
+	var err error
+	if !filepath.IsAbs(path) {
+		cwd, err := os.Getwd()
+		if err != nil {
+			return err
+		}
+		path = filepath.Join(cwd, filepath.Clean(path))
```
mtrmac, line 41: `filepath.Join` implicitly runs `Clean` on the _output_. Does calling `Clean` on the _input_ make any non-obvious difference? If so, please document that.

internal/util/util.go
```diff
@@ -22,3 +29,35 @@ func LookupImage(ctx *types.SystemContext, store storage.Store, image string) (*
 	}
 	return localImage, nil
 }
+
+// ExportFromReader reads bytes from given reader and exports to external tar, directory or stdout.
+func ExportFromReader(input io.Reader, isDir bool, isStdout bool, path string) error {
```
mtrmac, line 34: Couldn’t this just accept a `BuildOutputOption` argument, simplifying the only caller, and completely eliminate the risk of mismatching the order of the bool options?

tests/bud.bats
```diff
@@ -551,6 +551,51 @@ _EOF
...
+  mytmpdir=${TESTDIR}/my-dir
+  mkdir -p $mytmpdir
+  cat > $mytmpdir/Containerfile << _EOF
+FROM alpine
+RUN echo 'hello'> hello
+_EOF
+  run_buildah build --output type=tar,dest=$mytmpdir/rootfs.tar --signature-policy ${TESTSDIR}/policy.json -t test-bud -f $mytmpdir/Containerfile .
+  # explode tar
+  mkdir $mytmpdir/rootfs
+  tar -C $mytmpdir/rootfs -xvf $mytmpdir/rootfs.tar
```
mtrmac, line 579: (If the goal is to only check file presence, using `tar -t` would allow listing the contents of the tarball without writing to disk. OTOH this actually verifies that the archive can be correctly extracted (assuming the tests fail on any failed command), which is valuable.)

imagebuildah/stage_executor.go
```diff
@@ -1556,6 +1568,22 @@ func (s *StageExecutor) commit(ctx context.Context, createdBy string, emptyLayer
 		Manifest:              s.executor.manifest,
 		UnsetEnvs:             s.executor.unsetEnvs,
 	}
+	// generate build output
+	if s.executor.buildOutput != "" {
+		src, err := s.builder.MakeContainerImageRef(options)
+		if err != nil {
+			return "", nil, errors.Wrapf(err, "error computing layer digests and building metadata for container %q", s.builder.ContainerID)
+		}
+		rc, _, err := src.ExtractRootfs()
```
mtrmac, line 1577: Why is it OK to ignore `errChan`?

(Alternatively, I could see an argument that `errChan` should not exist, and instead `rc.Close()` should be blocking not that channel and handling the error, and that users of `rc` should always be obliged to check the return value of `rc.Close()`. That feels conceptually clearer but might just be a lot of extra complexity for no benefit.)

internal/parse/parse.go
```diff
@@ -41,6 +41,71 @@ var (
...
+		default:
+			return BuildOutputOption{}, fmt.Errorf("Unrecognized key %q in build output option: %q", arr[0], buildOutput)
+		}
+	}
+
+	if !typeSelected || !pathSelected {
+		return BuildOutputOption{}, fmt.Errorf("invalid build output options %q, accepted keys are type and dest", buildOutput)
+	}
+
+	return BuildOutputOption{path, isDir, isStdout}, nil
```
mtrmac, line 106: Name the fields in this initializer as well, please.

mtrmac (commented):
Just a few non-trivial things, and a lot of tiny text/documentatino nits… But also see elsewhere.

docs/buildah-build.1.md
mtrmac: (Marked unresolved to add that test, or to explicitly decide that we don’t need one.)

cmd/buildah/build.go
mtrmac: I do think that all such tools should do their own parsing; that way they will fail quickly on invalid input, rather than triggering the long and costly build, and only failing in the commit step. So if they do the parsing, they might just as well provide the parsed value, and that eliminates a failure case in the commit step.

Also consider the podman-remote case, where the remote server basically ignores the users’ input and needs to provide a `BuildOutput` value pointing at a temporary tarball location/stream. If the options allow parsing a structure, the server can just pass a `BuildOutputOption{Path:theTemporaryFile, …}`. If the options require a string, the server will need to, quite pointlessly, format one.

This is not blocking from my perspective, and there might be something that makes passing the deconstructed value unattractive. Still, I have a preference :)

mtrmac:
> Yes for podman-remote, I think we should reject it until we have a solid solution for it. The problem with that is on a MAC and Windows. We can discuss this at Water Cooler, to make us all understand the issue.

For the record, the water caller discussion pointed out that the remote load/save implementations stream a tarball, e.g. https://github.com/containers/podman/blob/712c3bb226a84c8a5593d62648975611cea7839d/pkg/domain/infra/tunnel/images.go#L259 . OTOH build is somewhat special, as an outsider I’m not sure that is viable here.

IIRC we didn’t discuss the intermediate situation, but just not implementing this on remote in the first version (or do we have to explicitly filter that option out in the server?) seems easiest.

mtrmac:
> We found one discrepancy in buildkit's behavior and its discussed and documented here: [#3823 (comment)](https://github.com/containers/buildah/pull/3823#discussion_r853821897), as of now I am not sure why that discrepancy exists and i'm unable to think of a reason to have that in buildah. I think we can keep that comment open and I'll address that in future in a followup PR if that becomes a problem.

Set-UID executables are Serious Stuff. I’m _very uncomfortable_ with shipping that feature without making a detailed decision, as well as with the idea of shipping one behavior (and having users possibly test that and rely on that), and possibly later changing that.

(I have no idea what the right behavior is here, neither from the compatibility perspective nor from the security perspective. I just really dislike to the idea of “this is hard, let’s declare success anyway” in this aspect.

As a possible point of comparison, what does `podman copy` do when copying out other-UID/set-UID/otherwise unusual files out of a container? Shouldn’t that be consistent with this code?)

cmd/buildah/build.go
mtrmac: > I do think that all such tools should do their own parsing

That implies that `GetBuildOutput` would have to be in an externally-importable package, not in `buildah/internal`.

docs/buildah-build.1.md
flouthoc: added a test.

cmd/buildah/build.go
flouthoc: moved this to external parse pkg so it can be used by external parties if needed.

docs/buildah-build.1.md
```diff
@@ -824,6 +842,16 @@ buildah bud --platform linux/arm64 --platform linux/amd64 --manifest myimage /tm
 
 buildah bud --all-platforms --manifest myimage /tmp/mysrc
 
+### Building an image using (--output) custom build output
+
+buildah build -o out .
```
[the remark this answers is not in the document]
flouthoc: Edited man page above

internal/parse/parse.go
```diff
@@ -41,6 +41,71 @@ var (
...
+
+// GetBuildOutput is responsible for parsing custom build output argument.
+// Takes `buildOutput` as string and returns path where output has to be generated returns BuildOutputOptions
+func GetBuildOutput(buildOutput string) (BuildOutputOption, error) {
+	isDir := true
+	isStdout := false
+	typeSelected := false
+	pathSelected := false
+	path := ""
+	if len(buildOutput) == 1 && buildOutput == "-" {
```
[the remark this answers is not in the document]
flouthoc: Handles `dest=-` now.

internal/parse/parse.go
```diff
@@ -41,6 +41,71 @@ var (
...
+		// Read more here: https://docs.docker.com/engine/reference/commandline/build/#custom-build-outputs
+		return BuildOutputOption{Path: "",
+			IsDir:    false,
+			IsStdout: true}, nil
+	}
+	if !strings.Contains(buildOutput, ",") {
+		// expect default --output <dirname>
+		return BuildOutputOption{Path: buildOutput,
+			IsDir:    isDir,
+			IsStdout: isStdout}, nil
```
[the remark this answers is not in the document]
flouthoc: fixed

image.go
```diff
@@ -43,7 +43,7 @@ const (
 	Dockerv2ImageManifest = define.Dockerv2ImageManifest
 )
 
-type containerImageRef struct {
+type ContainerImageRef struct {
```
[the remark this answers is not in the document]
flouthoc: We can move `ContainerImageRef` to internal but as of now I don't see lot of harm in making it public since it has access to some useful methods and ContainerImageRef does not changes frequently but making it internal is safer.

tests/bud.bats
flouthoc: I didn't thought of any advantage here but I like the explanation on the second part hence keeping it as it is please feel free to unresolve this comment.

imagebuildah/stage_executor.go
flouthoc: handled.

flouthoc:
@mtrmac Thanks for review resolved all the nits. For this ( https://github.com/containers/buildah/pull/3823#issuecomment-1104239909 ) i agree however i really don't understand the disparity with the buildkits output for `root` and `rootless` invocation if we decide to differ from buildkit here we can close down on one consistent behavior. I'll discuss this with @nalind whenever he is back and do the needful.

docs/buildah-build.1.md
```diff
@@ -411,6 +411,24 @@ By default, Buildah manages _/etc/hosts_, adding the container's own IP address.
...
+
+Supported _keys_ are:
+- **dest**: Destination path for exported output. Valid value is absolute or relative path.
+- **type**: Defines the type of output to be used. Valid values is documented below.
+
+Valid _type_ values are:
+- **local**: write the resulting build files to a directory on the client-side.
+- **tar**: write the resulting files as a single tarball (.tar).
+
+If no type is specified, the value defaults to local. Use a hyphen as non-comma format (`-`) to write the output tarball to standard output (`STDOUT`).
```
mtrmac, line 430: Maybe
```suggestion
If no type is specified, the value defaults to local. Use a hyphen as non-comma format (`-`) to write the output tarball to standard output.
```
There is no user-facing `STDOUT` identifier / name, to my knowledge.

mtrmac, line 430: This
- doesn’t document `dest=-` AFAICS, although the implementation now supports that.
- Says “non-comma format”, which is a short-hand I started using for the purposes of this review, but that is unclear without extra content.


Maybe something vaguely like
> ```
> - **dest**: … absolute or relative path; or, as a special case, `-` means the standard output.
> …
> Alternatively, instead of a comma-separated sequence, the value of **--output** can be just a destination (in the `**dest** format) (e.g. `--output some-path`, `--output -`)
> ```

pkg/parse/parse.go
```diff
@@ -510,6 +510,70 @@ func AuthConfig(creds string) (*types.DockerAuthConfig, error) {
...
+		default:
+			return define.BuildOutputOption{}, fmt.Errorf("Unrecognized key %q in build output option: %q", arr[0], buildOutput)
+		}
+	}
+
+	if !typeSelected || !pathSelected {
+		return define.BuildOutputOption{}, fmt.Errorf("invalid build output options %q, accepted keys are type and dest must be present", buildOutput)
+	}
+
+	if !isDir && path == "-" {
```
mtrmac, line 568: Should we reject `type=dir,dest=-`?

pkg/parse/parse.go
```diff
@@ -510,6 +510,70 @@ func AuthConfig(creds string) (*types.DockerAuthConfig, error) {
...
 }
 
+// GetBuildOutput is responsible for parsing custom build output argument i.e `build --output` flag.
+// Takes `buildOutput` as string and returns BuildOutputOptions
+func GetBuildOutput(buildOutput string) (define.BuildOutputOption, error) {
+	isDir := true
+	isStdout := false
+	typeSelected := false
+	pathSelected := false
+	path := ""
```
mtrmac, line 520: (Absolutely non-blocking: All of this can now be moved just before the `tokens :=` initializer, requiring a tiny bit less context to keep in mind when reading the code.)

pkg/parse/parse.go
```diff
@@ -510,6 +510,70 @@ func AuthConfig(creds string) (*types.DockerAuthConfig, error) {
 	}, nil
 }
 
+// GetBuildOutput is responsible for parsing custom build output argument i.e `build --output` flag.
+// Takes `buildOutput` as string and returns BuildOutputOptions
```
mtrmac, line 514:
```suggestion
// Takes `buildOutput` as string and returns BuildOutputOption
```

docs/buildah-build.1.md
mtrmac: The “alternatively” part should also document the implied `type` value…

… wait, `--output foo` means `type=local`, but `--output -` means `type=tar`? That needs to be explicitly documented as well.

internal/util/util.go
```diff
@@ -22,3 +30,35 @@ func LookupImage(ctx *types.SystemContext, store storage.Store, image string) (*
...
+
+// ExportFromReader reads bytes from given reader and exports to external tar, directory or stdout.
+func ExportFromReader(input io.Reader, opts define.BuildOutputOption) error {
+	var err error
+	if !filepath.IsAbs(opts.Path) {
+		cwd, err := os.Getwd()
+		if err != nil {
+			return err
+		}
+		opts.Path = filepath.Join(cwd, opts.Path)
```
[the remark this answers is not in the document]
flouthoc: switched to abs(

docs/buildah-build.1.md
```diff
@@ -411,6 +411,25 @@ By default, Buildah manages _/etc/hosts_, adding the container's own IP address.
...
+Supported _keys_ are:
+- **dest**: Destination path for exported output. Valid value is absolute or relative path, `-` means the standard output.
+- **type**: Defines the type of output to be used. Valid values is documented below.
+
+Valid _type_ values are:
+- **local**: write the resulting build files to a directory on the client-side.
+- **tar**: write the resulting files as a single tarball (.tar).
+
+If no type is specified, the value defaults to **local**.
+Alternatively, instead of a comma-separated sequence, the value of **--output** can be just a destination (in the `**dest** format) (e.g. `--output some-path`, `--output -`) where `--output some-path` is equivalent to **type=local** and `--output -` is equivalent to **type=tar**.
```
mtrmac, line 431: One last nit: `s/is equivalent to/is treated as if/` or something like that. `some-path` is not “equivalent” to a type value.

Or commit to this approach, and write “`--output some-path` is equivalent to `type=local,dest=some-path`”.

pkg/parse/parse.go
```diff
@@ -510,6 +510,74 @@ func AuthConfig(creds string) (*types.DockerAuthConfig, error) {
...
+		return define.BuildOutputOption{}, fmt.Errorf("invalid build output option %q, accepted keys are type and dest must be present", buildOutput)
+	}
+
+	if !isDir && path == "-" {
+		return define.BuildOutputOption{Path: "",
+			IsDir:    false,
+			IsStdout: true}, nil
+	}
+
+	if isDir && path == "-" {
```
mtrmac, line 574: This can be simplified to
```go
if path == "-" {
   if isDir { fail }
   return $accepted
}
```
where the $accepted value is known to be correct by construction, and we don’t have to worry about the two conditions getting out of sync.

mtrmac (commented):
Thanks!

The local implementation LGTM now; some resolution of the ownership/set-UID behavior, by someone who understands it (i.e. not me) is still a blocker for me.

image.go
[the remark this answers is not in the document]
nalind: (In the case of this one in particular, the last revision of the PR I saw needed to be able to call `ExtractRootFS` from a different package, and I didn't have a better way to accomplish that in-hand.)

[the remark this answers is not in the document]
nalind: Okay, an alternative would be to leave the `containerImageRef` type and its `extractRootfs()` method private, and add a public method to `Builder` that internally calls `makeContainerImageRef()` to create a `containerImageRef`, calls the resulting ref's `extractRootfs()` method, and hands the result back to the new methods's caller.  That's enough for the use case in `StageExecutor.commit()`, and doesn't expose a lot of our internals, so I think I like it better.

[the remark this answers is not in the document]
flouthoc: ackd SGTM I'll add these changes.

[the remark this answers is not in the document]
flouthoc: fixed.

flouthoc:
@nalind Resolved all issues but this one is still left https://github.com/containers/buildah/pull/3823#discussion_r853821897

imagebuildah/executor.go
flouthoc: Still not entirely sure about what should we do here and which behavior is the correct one.

flouthoc: Looked at it again it seems like the issue is with `unpriviledged` invocation of buildkit it does not preserves `setuid` but preserves `setgid` when manually added in Dockerfile but again overrides `setgid` for the files which were already there in the base image. This behavior looks really confusing to me could it be a bug in buildkit ? 

`priviledged` invocation of both buildah and buildkit works in the same manner and `unpriviledged` invocation of buildah preserves everything.

@nalind can we hit any security issues if we choose not to replicate buildkit's unpriviledged invocation behavior ?

imagebuildah/stage_executor.go
```diff
@@ -1556,6 +1567,25 @@ func (s *StageExecutor) commit(ctx context.Context, createdBy string, emptyLayer
...
+		rc, errChan, err := s.builder.ExtractRootfs(options)
+		if err != nil {
+			return "", nil, errors.Wrapf(err, "failed to extract rootfs from given container image")
+		}
+		defer rc.Close()
+		err = internalUtil.ExportFromReader(rc, buildOutputOption)
+		if err != nil {
+			return "", nil, errors.Wrapf(err, "failed to export build output")
+		}
+		if errChan != nil {
```
nalind, line 1581: This is dead code now, isn't it?

imagebuildah/executor.go
nalind: Either there's some rule there that we don't understand, or it's a bug.  `sudo` seems to be an outlier, as I also lose the setuid bit on binaries like passwd (from the passwd package), chfn and chsh (from the util-linux-user package), and the setgid bit on binaries like locate (from the mlocate package) and lockdev (from the lockdev package) when I run a build as an unprivileged user.
The general case sure seems to suggest that we should be stripping the setuid/setgid/sticky bits when we're writing to "local" destinations as an unprivileged user.

image.go
```diff
@@ -813,3 +813,12 @@ func (b *Builder) makeImageRef(options CommitOptions) (types.ImageReference, err
 	}
 	return ref, nil
 }
+
+// Extract the container's whole filesystem as if it were a single layer from current builder instance
+func (b *Builder) ExtractRootfs(options CommitOptions) (io.ReadCloser, chan error, error) {
+	src, err := b.makeContainerImageRef(options)
+	if err != nil {
+		return nil, nil, errors.Wrapf(err, "error computing layer digests and building metadata for container %q", b.ContainerID)
```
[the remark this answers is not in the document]
flouthoc: fixed

imagebuildah/executor.go
nalind: Also, pretty sure we want to be preserving ownership information when we're invoked as root, even though buildkit seems to change setgid-"tty" /usr/bin/write and setgid-"lock" /usr/sbin/lockdev into setgid-"root" /usr/bin/write and /usr/sbin/lockdev.

flouthoc: @nalind Modified buildah to preserve ownership as is when invoked with as `root`, this is different behavior from buildkit but looks good to me.

flouthoc: They are preserved as is.

nalind: The buildkit version also seems to create the top-level destination directory with 0700 permissions if it doesn't already exist, while we seem to create it as 0755 (neither modifies the permissions on an already-present directory).  If we're going to be putting setuid/setgid files in there, I'd be more comfortable if we used 0700.

flouthoc: nice catch, fixed that as well.

nalind: I still think that what we're seeing with `sudo`'s setuid bit being preserved when buildkit is run by an unprivileged user is not what we want to be using as our rule for everything.  We should be stripping setuid/setgid bits and capability xattrs when we're writing to "local" destinations as a non-root user.

flouthoc: Pushed a new version which covers this for `non-root` user.

image.go
```diff
@@ -150,7 +150,10 @@ func computeLayerMIMEType(what string, layerCompression archive.Compression) (om
 }
 
 // Extract the container's whole filesystem as if it were a single layer.
-func (i *containerImageRef) extractRootfs() (io.ReadCloser, chan error, error) {
+// Takes copierStripOpts as argument which allows caller to configure
+// preserve nature of setuid,setgid,sticky and extended attributes
+// on extracted rootfs.
+func (i *containerImageRef) extractRootfs(copierStripOpts copier.GetOptions) (io.ReadCloser, chan error, error) {
```
nalind, line 156: Go ahead and define a `define.ExtractRootfsOptions` struct that only includes fields whose values are used.

imagebuildah/stage_executor.go
```diff
@@ -1556,6 +1568,33 @@ func (s *StageExecutor) commit(ctx context.Context, createdBy string, emptyLayer
...
 		UnsetEnvs:             s.executor.unsetEnvs,
 	}
+	// generate build output
+	if s.executor.buildOutput != "" {
+		copierStripOpts := copier.GetOptions{}
+		if unshare.IsRootless() {
+			// Strip setuid,setgid,sticky and extended attributes
+			copierStripOpts.StripSetuidBit = true
+			copierStripOpts.StripSetgidBit = true
+			copierStripOpts.StripStickyBit = true
```
nalind, line 1578: Looking at the top-level `tmp` directory produced by buildkit, I think `StripStickyBit` needs to be left `false` always, and as a result not be part of a `ExtractRootfsOptions` struct.

image.go
```diff
@@ -43,6 +43,15 @@ const (
 	Dockerv2ImageManifest = define.Dockerv2ImageManifest
 )
 
+// ExtractRootfsOptions is consumed by ExtractRootfs() which allows
+// users to preserve nature of various modes like setuid, setgid and xattrs
+// over the extracted file system objects.
+type ExtractRootfsOptions struct {
+	StripSetuidBit bool // strip the setuid bit off of items being copied. no effect on archives being extracted
+	StripSetgidBit bool // strip the setgid bit off of items being copied. no effect on archives being extracted
+	StripXattrs    bool // don't record extended attributes of items being copied. no effect on archives being extracted
```
nalind, line 52: These comments need to be updated.

internal/util/util.go
```diff
@@ -22,3 +31,45 @@ func LookupImage(ctx *types.SystemContext, store storage.Store, image string) (*
...
+	if opts.IsDir {
+		// preserve ownership when invoked as root
+		noLChown := false
+		if unshare.IsRootless() {
+			noLChown = true
+		}
+
+		err = os.MkdirAll(opts.Path, 0700)
+		if err != nil {
+			return err
```
nalind, line 53: `errors.Wrapf()` this, so that if we fail here, the caller can get a stack trace that points to this line.

nalind:
Two nits, otherwise LGTM.

flouthoc:
@rhatdan @nalind needs /lgtm

imagebuildah/stage_executor.go
```diff
@@ -1556,6 +1568,32 @@ func (s *StageExecutor) commit(ctx context.Context, createdBy string, emptyLayer
 		Manifest:              s.executor.manifest,
 		UnsetEnvs:             s.executor.unsetEnvs,
 	}
+	// generate build output
+	if s.executor.buildOutput != "" {
+		extractRootfsOpts := buildah.ExtractRootfsOptions{}
+		if unshare.IsRootless() {
+			// Strip setuid,setgid,sticky and extended attributes
```
mtrmac, line 1575: Non-blocking? It’s kinda obvious that that’s what the three lines do, to me that doesn’t need a comment.

But I’d _love_ to see a comment recording the outcome of the discussions. What's the rationale for stripping these bits for rootless use, and not as root? (IOW, if someone came 5 years later thinking that this is surprising, what would be necessary to know?)

[OTOH if this is a hard-won consensus but asking for details would break that consensus :), by all means let’s leave it as it is.]

internal/util/util.go
```diff
@@ -22,3 +31,45 @@ func LookupImage(ctx *types.SystemContext, store storage.Store, image string) (*
...
+func ExportFromReader(input io.Reader, opts define.BuildOutputOption) error {
+	var err error
+	if !filepath.IsAbs(opts.Path) {
+		opts.Path, err = filepath.Abs(opts.Path)
+		if err != nil {
+			return err
+		}
+	}
+	if opts.IsDir {
+		// preserve ownership when invoked as root
```
mtrmac, line 45: Same documentation point here.

(And if the two are parts of a single idea, as it seems to be the case, it would be nice to at least have comments that link the two. Or perhaps an explicit code link — have a `preserveOwnership bool` parameter to this function, and have the caller set it along with the three set-UID/set-GUID/xattr choices, in a single `if` block.)

mtrmac (commented):
Code LGTM, just stirring up trouble… Feel free to ignore me.

internal/util/util.go
flouthoc: Added a reason and link to the thread.

rhatdan:
/lgtm
/hold
@flouthoc did you open a bug/question on why buildkit does this change between rootless and rootful?

Issue #3822, opened 2022-03-15
preprocessing not working with stable buildah Image 


Basti-D:
**Description**

`buildah bud Dockerfile.in . `
using the prebuilt stable Dockerimage 
results in 
`exec: "cpp": executable file not found in $PATH`

**Steps to reproduce the issue:**
1. checkout buildah repo
2. `cd buildah/tests/bud/preprocess`
3.  create test Dockerfile 
```
echo 'FROM quay.io/buildah/stable' > test.dockerfile
echo 'COPY . .' >> test.dockerfile
echo 'RUN buildah bud  -f Decomposed.in .' >> test.dockerfile
```
4. check whether the normal decomposed test works: 
```
buildah bud  -f Decomposed.in .
STEP 1: FROM alpine:latest
STEP 2: RUN touch /etc/hello-world.txt
STEP 3: RUN echo "Hello world!"
Hello world!
STEP 4: RUN apk update
^C
```
-> yes, works perfectly fine.
5. check whether the buildah stable image works as intended
```
buildah bud  -f test.dockerfile .
STEP 1: FROM quay.io/buildah/stable
STEP 2: COPY . .
STEP 3: RUN buildah bud  -f Decomposed.in .
exec: "cpp": executable file not found in $PATH
time="2022-03-15T09:29:54Z" level=error msg="exit status 125"
error building at STEP "RUN buildah bud  -f Decomposed.in .": error while running runtime: exit status 125
ERRO[0003] exit status 125
```
-> broken, since cpp not found.

**Describe the results you received:**
The image was not built, because cpp(1) is not installed (or missing in the PATH environment variable).
`exec: "cpp": executable file not found in $PATH`

**Describe the results you expected:**
I expect that the image would have been built using the stable buildah image as a builder image.

**possible solutions**
Installing cpp inside the image should solve the issue.
Drawback: The imagesize will be increased.

Basti-D:
@flouthoc this is our current approach which solves the issue.
But generally speaking I assume that if there exists an image, all functionality should be available. If some functionality is not available it should be documented somewhere.
Probably an additional image, which is larger but has all functionality, could make sense?

rhatdan:
```
$ rpm -q buildah --suggests
cpp
qemu-user-static
```
It is under suggests for RPM packaging. We could make the error message a little more clear.


`exec: "cpp": executable file not found in $PATH, please install the cpp package.`

vrothberg:
[...]

PR #3825, opened 2022-03-17
buildah image: install cpp

vrothberg:
Ship `cpp` with the Buildah container images to make sure that
preprocessing .in files works as expected and documented.

Also improve the error message when cpp isn't installed.

Fixes: #3822

### DIFF
--- a/contrib/buildahimage/centos7/Dockerfile
+++ b/contrib/buildahimage/centos7/Dockerfile
@@ -8,7 +8,7 @@ FROM centos:7
 
 # Remove directories used by yum that are just taking
 # up space.
-RUN useradd build; yum -y update; rpm --restore shadow-utils 2>/dev/null; yum -y install buildah fuse-overlayfs xz; rm -rf /var/cache /var/log/dnf* /var/log/yum.*;
+RUN useradd build; yum -y update; rpm --restore shadow-utils 2>/dev/null; yum -y install cpp buildah fuse-overlayfs xz; rm -rf /var/cache /var/log/dnf* /var/log/yum.*;
 
 ADD https://raw.githubusercontent.com/containers/buildah/main/contrib/buildahimage/stable/containers.conf /etc/containers/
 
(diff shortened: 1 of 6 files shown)


flouthoc (approved):
LGTM. @vrothberg would this update `:latest` as well.

vrothberg:
> LGTM. @vrothberg would this update `:latest` as well.

Yes, all but the "manual" one are built by Quay automatically. @TomSweeneyRedHat is their loving shepherd.

TomSweeneyRedHat:
@vrothberg and probably @cevich, here's the question I don't want to ask.  Do we need to rebuild the current versions in quay.io to make this work for everyone?  Or do we add a note to the README.md telling folks about the problem, which versions of the images were affected, and how (if possible) to fix it for a particular image?

TomSweeneyRedHat:
The changes LGTM, btb, unless a README.md adjustment is needed.

cevich:
> Do we need to rebuild the current versions in quay.io to make this work for everyone? Or do we add a note to the README.md telling folks about the problem, which versions of the images were affected, and how (if possible) to fix it for a particular image?

Oof, maybe both, checking...

cevich:
...what should happen once this is merged:  Within a day the multi-arch images and manifest list will be built including these changes.  For 'testing' and 'upstream', it will push to the 'latest' tagged images only.  For 'stable', both 'latest' and the version-tagged images will be updated.  Previously pushed (older) version-tagged images won't be affected at all.

So this could cause unexpected changes in the most recent version-tagged 'stable' image.  That's probably where a README update might be nice, making note of the version-tag where the change appears.

Hope that helps.

Ref: https://github.com/containers/buildah/blob/main/.github/workflows/multi-arch-build.yaml

vrothberg:
Which README are we talking about?  I don't see one where we record bug fixes.  I would keep it simple, merge and then update the images.

TomSweeneyRedHat:
If we add to the readme.md, it would be this one: https://github.com/containers/buildah/tree/main/contrib/buildahimage

Issue #3826, opened 2022-03-17
Feature request: option to suppress io.buildah.version labels


csmith:
It would be nice if there were an option to suppress the `io.buildah.version` labels that buildah applies to all built images.

I'm trying to make images that are reproducible, but at present you need to exactly match the version of buildah to be able to successfully reproduce an image. Comparing images built by buildah 1.21.3 and 1.24.2, it seems that they would be perfectly identical if not for the `io.buildah.version` labels.

flouthoc:
Hi @csmith , Thanks for creating the issue. I think you can override `io.buildah.version` but I also agree that there must be option to suppress default labels. Above PR adds a new flag `--remove-identity-labels` which adds the following requested feature.

PR #3829, opened 2022-03-21
build, commit: allow removing default identity labels using `--identity-labels=false`

flouthoc:
Allow end users to remove default identity labels if they want to.
Since there are instances where images can be reproduced across version
hence users must have option to suppress default labels.

Closes: https://github.com/containers/buildah/issues/3826

### DIFF
--- a/define/build.go
+++ b/define/build.go
@@ -29,6 +29,8 @@ type CommonBuildOptions struct {
 	CPUSetMems string
 	// HTTPProxy determines whether *_proxy env vars from the build host are passed into the container.
 	HTTPProxy bool
+	// IdentityLabel if set ensures that default `io.buildah.version` label is not applied to build image.
+	IdentityLabel types.OptionalBool
 	// Memory is the upper limit (in bytes) on how much memory running containers can use.
 	Memory int64
 	// DNSSearch is the list of DNS search domains to add to the build container's /etc/resolv.conf
--- a/pkg/cli/common.go
+++ b/pkg/cli/common.go
@@ -74,6 +74,7 @@ type BudResults struct {
 	PullAlways          bool
 	PullNever           bool
 	Quiet               bool
+	IdentityLabel       bool
 	Rm                  bool
 	Runtime             string
 	RuntimeFlags        []string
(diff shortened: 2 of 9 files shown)


rhatdan:
Could you change the option to --identify-labels with a default value of true.

Then user could type --identify-labels=false. Shorter to type. 
Or perhaps --idlabels=false.

flouthoc:
@rhatdan Would using default `true` cause problem for cases where `podman remote / API` or other tools which directly call buildah and initialize option in structs, we will have to instrument all such occurrences  of code to set default `true` however as of now they will just not set the field and it will remain `false`.

rhatdan:
In the remote protocol, it should be an optional bool and only set when the user sets it.  Otherwise it defaults to true.

define/build.go
```diff
@@ -29,6 +29,8 @@ type CommonBuildOptions struct {
 	CPUSetMems string
 	// HTTPProxy determines whether *_proxy env vars from the build host are passed into the container.
 	HTTPProxy bool
+	// IdentityLabel if set ensures that default `io.buildah.version` label is not applied to build image.
+	IdentityLabel types.OptionalBool
```
flouthoc, line 33: Makes IdentityLabel optional

flouthoc:
@nalind @rhatdan PTAL

pkg/cli/common.go
```diff
@@ -227,6 +228,7 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
 		panic(fmt.Sprintf("error marking the pull-never flag as hidden: %v", err))
 	}
 	fs.BoolVarP(&flags.Quiet, "quiet", "q", false, "refrain from announcing build instructions and image read/write progress")
+	fs.BoolVar(&flags.IdentityLabel, "identity-label", true, "add default builder label (default true)")
```
TomSweeneyRedHat, line 231:
```suggestion
	fs.BoolVar(&flags.IdentityLabel, "identity-label", true, "add default identity label (default true)")
```
At least to be consistent with the man page

pkg/parse/parse.go
```diff
@@ -146,25 +147,26 @@ func CommonBuildOptionsFromFlagSet(flags *pflag.FlagSet, findFlagFunc func(name
...
+		HTTPProxy:     httpProxy,
+		Memory:        memoryLimit,
+		MemorySwap:    memorySwap,
+		NoHosts:       noHosts,
+		ShmSize:       findFlagFunc("shm-size").Value.String(),
+		IdentityLabel: types.NewOptionalBool(identityLabel),
+		Ulimit:        ulimit,
+		Volumes:       volumes,
+		Secrets:       secrets,
+		SSHSources:    sshsources,
```
TomSweeneyRedHat, line 169: Can we put this in alpha order?

flouthoc:

pkg/parse/parse.go
nalind: Please don't.  It creates noise for `git annotate`.

flouthoc: @nalind Everything was already in order I just moved new label in its proper place.

rhatdan:
LGTM

flouthoc:

rhatdan:
@giuseppe @vrothberg @saschagrunert PTAL

vrothberg (commented):
/lgtm

PR #3835, opened 2022-03-22
network setup: increase timeout to 4 minutes

Luap99:
#### What type of PR is this?

 /kind bug

Podman waits for 4 minutes to wait for conmon/oci runtime to create the
container[1]. Since this value seems to work we should use the same one
here.

[1] https://github.com/containers/podman/blob/b4b8b8b5374967f9a98aef6c9793c51342d54b9d/libpod/define/runtime.go#L27

Fixes containers/podman#13327

```release-note

```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -2333,9 +2333,9 @@ func (b *Builder) runUsingRuntimeSubproc(isolation define.Isolation, options Run
 	return err
 }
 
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
@Luap99 you need the `[NO NEW TESTS NEEDED]` tag

flouthoc (commented):
LGTM

PR #3836, opened 2022-03-22
buildah: actually use containers.conf settings

Luap99:
#### What type of PR is this?

 /kind bug
 /kind cleanup

Buildah ignores the container.conf settings. Commit 05d5d22dc6b7 tried
to fix this but I cannot see how this would have worked, there were also
no test added related to containers.conf.

The code below were we read the default just overwrites everything we
already set so the config settings are ignored.

Also there are many weird places were settings are just overwritten for
no particular reason. The namespaces code path looks like mess to me.
Namespaces are added/changed/removed in many different places and there
is no explanation why. Maybe I am just not familar enough with this code
base but I cannot really understand this. Hopefully a maintainer can
take a closer look to see if my changes are indeed correct.

#### How to verify it

add
```
[containers]
netns = "host"
```
to container.conf add run buildah build and check the namespace.

Fixes containers/podman#13294

```release-note

```


Luap99:
WIP because I still need to add tests.

tests/helpers.bash
```diff
@@ -341,7 +341,9 @@ function assert() {
             return
         fi
     else
-        if eval "[[ $not \$actual_string $actual_op \$expect_string ]]"; then
+        # Escaping the strings with quotes is important, otherwise we run
+        # into the same problem as described in the comment above.
+        if eval "[[ $not \"\$actual_string\" $actual_op \"\$expect_string\" ]]"; then
```
edsantiago, line 346: Preference for:
```diff
index 2bb6b429..1be49b03 100644
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -340,6 +340,11 @@ function assert() {                                                                                                                       
         if [ "$actual_string" = "$expect_string" ]; then
             return
         fi
+    elif [[ $operator == '!=' ]]; then
+        # Same special case as above
+        if [ "$actual_string" != "$expect_string" ]; then
+            return
+        fi
     else
         if eval "[[ $not \$actual_string $actual_op \$expect_string ]]"; then
             return
```
Reason: the quoting-approach (which is what I initially suggested, my bad) will fail completely when operator is `~` (i.e. when we really _do_ want patterns).

Luap99: added, thanks

Luap99:
@nalind Is there a reason we force ipc/pid namespaces for rootless?
https://github.com/containers/buildah/blob/1b8c014f3eee13d26d126b945959087629505396/run_linux.go#L2342-L2369

nalind:
> @nalind Is there a reason we force ipc/pid namespaces for rootless?
> 
> https://github.com/containers/buildah/blob/1b8c014f3eee13d26d126b945959087629505396/run_linux.go#L2342-L2369

Not specifically, but that logic was from when runc didn't "know" about rootless use cases, so it may have been necessary at the time to get it to be able to successfully run at all.  If that's not the case any more, feel free to change it.

Luap99:
Some test are failing when we stop forcing the private namesapce for rootless, however `podman run --pid host --ipc host --net host --userns host --uts host --cgroupns host alpine ls -l /proc/self/ns/` works just fine and does indeed use the host namespace.

@giuseppe I see this error with buildah:
```
error running container: error from /usr/bin/crun creating container for [/bin/pwd]: mount `mqueue` to `/dev/mqueue`: Operation not permitted
```
Any idea what podman is doing differently from buildah to make it work?

Reproducer with this PR:
```
$ buildah from alpine
$ buildah run --pid host --ipc host --net host --uts host --cgroupns host  alpine-working-container ls -l /proc/self/ns
2022-03-31T10:59:40.000561819Z: mount `mqueue` to `/dev/mqueue`: Operation not permitted
error running container: error from /usr/bin/crun creating container for [/bin/ls -l /proc/self/ns]: : exit status 1
```

Luap99:
> Some test are failing when we stop forcing the private namesapce for rootless, however `podman run --pid host --ipc host --net host --userns host --uts host --cgroupns host alpine ls -l /proc/self/ns/` works just fine and does indeed use the host namespace.
> 
> @giuseppe I see this error with buildah:
> 
> ```
> error running container: error from /usr/bin/crun creating container for [/bin/pwd]: mount `mqueue` to `/dev/mqueue`: Operation not permitted
> ```
> 
> Any idea what podman is doing differently from buildah to make it work?
> 
> Reproducer with this PR:
> 
> ```
> $ buildah from alpine
> $ buildah run --pid host --ipc host --net host --uts host --cgroupns host  alpine-working-container ls -l /proc/self/ns
> 2022-03-31T10:59:40.000561819Z: mount `mqueue` to `/dev/mqueue`: Operation not permitted
> error running container: error from /usr/bin/crun creating container for [/bin/ls -l /proc/self/ns]: : exit status 1
> error while running runtime: exit status 1
> ```

If I set both pid and ipc to private it works.

Luap99:
I found this code in podman: https://github.com/containers/podman/blob/4ba71f955a944790edda6e007e6d074009d437a7/pkg/specgen/generate/oci.go#L260-L279

Do we need to do the same for rootless buildah?

giuseppe:
yes.  I think we need something similar in buildah, when cannot mount mqueue if the user namespace doesn't own the ipc namespace, similarly for procfs and the pid namespace.

rhatdan:
LGTM

giuseppe (approved):
/lgtm

PR #3837, opened 2022-03-22
run: set parent-death signals and forward SIGHUP/SIGINT/SIGTERM

### LINKED ISSUE #3544: buildah not actually stopping on SIGTERM

nalind:
#### What type of PR is this?

/kind bug

Restore setting of the parent-death signal when we're running subprocesses in Run(), so that if we get killed, the child processes will also get killed.

While a child process is running, if we receive SIGHUP, SIGINT, or SIGTERM, forward the signal to our child process unless it's ~~a container~~ the command we're executing, which we SIGKILL without mercy, and finish the current routine, which will then notice that the child process has exited and return an error to its caller.

#### How to verify it

New integration test!

Fixes #3544.

```release-note
None
```

### DIFF
--- a/chroot/run.go
+++ b/chroot/run.go
@@ -10,6 +10,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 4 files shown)


rhatdan (approved):
LGTM

nalind:
Rebased to land after #3859 brought in the changes we needed from the storage library.

PR #3841, opened 2022-03-23
Don't pollute

edsantiago:
Several tests were doing '--iidfile x', where 'x' is a filename
with no path. This is super bad: if running as root, it leaves
litter in the current directory. If rootless, it may throw EPERM.
Clean those up, using $TESTDIR for each one.

Tested by running the test suite on my laptop and confirming
there are no more droppings.

And, refactor one particularly hairy set of duplicated build
incantations: make them a series of simple clean loops.


rhatdan:
LGTM

vrothberg (approved):
/lgtm

PR #3847, opened 2022-03-24
[CI:DOCS] man pages: sort flags, and keep them that way

edsantiago:
Enforce alphabetical ordering of command-line options in
man pages. Not as simple as with podman, because conventions
are different.

Reference: https://github.com/containers/podman/pull/13625

### DIFF
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -271,7 +271,7 @@ environment variable.  `export BUILDAH_FORMAT=docker`
 Overrides the first `FROM` instruction within the Containerfile.  If there are multiple
 FROM instructions in a Containerfile, only the first is changed.
 
-**-h**, **--help**
+**--help**, **-h**
 
 Print usage statement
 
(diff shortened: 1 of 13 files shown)


flouthoc (approved):
LGTM

edsantiago:
I think I see what's happening:
cirrus-map-main

`Smoke` is required only by tasks that are skipped on `CI:DOCS`, so I don't think it's actually run. 

Solution(?): add a dependency to `meta` or `success`, pointing back to `smoke`, so `smoke` runs? (I really don't know what `meta` is, though, and I don't see it running either).

@cevich thoughts?

edsantiago:
(Force-pushed with no changes. That seems to have unstuck CI)

cevich:
> Smoke is required only by tasks that are skipped on CI:DOCS, so I don't think it's actually run.

The `depends_on` only influences the order and parallelism with other tasks, not whether or not it runs.

I'm not sure what happened, but for the last few days CI has been getting "stuck" due to github outages, so that's what I'd look for first.

PR #3855, opened 2022-03-24
Add a test for CVE-2022-27651

nalind:
#### What type of PR is this?

/kind other

Check that the inheritable capabilities are set to 0, even when we explicitly try to add capabilities, so that we won't inadvertently break the fix for CVE-2022-27651 later.

#### How to verify it

New integration test!

None

```release-note
None
```


flouthoc (approved):
LGTM

PR #3865, opened 2022-03-25
[release-1.11-rhel] CVE-2022-21698: bump github.com/prometheus/client_golang to v1.11.1

nalind:
Pin a few dependencies to prevent them from being upgraded to versions that use features from newer versions of the standard library so that we can continue compiling with older versions of the compiler that were current when this branch was current.

Resolves: CVE-2022-21698


flouthoc:
@nalind Vendor is failing

nalind:
I did the `go mod ...` with 1.12, looks like we used 1.13 for this branch.  Fixing.

Issue #3861, opened 2022-03-25
Allow combination of `--no-cache` and `--layers=true`


bendem:
Currently, combining `--no-cache` and `--layers=true` results in 
```
can only set one of 'layers' or 'no-cache'
ERRO[0000] exit status 125
```

There are valid use cases where we want `buildah` to produce layers (speed up next builds, provide more layers to improve push/pull operations) while not using currently available layers. One such case for us is weekly rebuild: we want the weekly rebuild to ignore existing layers while still providing build cache and separate layers for future builds.

bendem:
The restriction was added here, not sure why.
https://github.com/containers/buildah/commit/fc438bb932e891cbe04109cfae1dfbe3c99307a5

rhatdan:
That makes some sense to me.  Not sure why this was blocked, but it might have been a mistaken thought.

rhatdan:
Please open a PR to remove the check.

flouthoc:
I agree use-case makes sense @bendem would you like to take a stab at this ? For starting point I think check is only at cli and some part might be in `imagebuildah/stage_executor.go`.

flouthoc:
This seemed fairly simple above PR should close this.

PR #3870, opened 2022-03-28
build: allow `--no-cache` and `--layers` so build cache can be overrided

flouthoc:
Buildah should allow users to override build cache by using `--no-cache`
and `--layers` togather, this is already functional but there is a top
level check which must be removed

Closes: https://github.com/containers/buildah/issues/3861


rhatdan:
LGTM
@nalind @giuseppe @vrothberg @ashley-cui @umohnani8 PTAL

giuseppe (approved):
/lgtm

nalind (approved):

bendem:
Nice, thanks for the quick fix

PR #3871, opened 2022-03-28
Updated dependabot to get updates for GitHub actions.

naveensrinivasan:
GitHub sends Dependabot alerts when we detect vulnerabilities affecting your repository
as well as when there are new updates to the dependency.

https://docs.github.com/en/code-security/dependabot/dependabot-alerts/about-dependabot-alerts

A vulnerability is a problem in a project's code that could be exploited to damage the confidentiality, integrity, or availability of the project or other projects that use its code. Vulnerabilities vary in type, severity, and method of attack.

When your code depends on a package that has a security vulnerability, this vulnerable dependency can cause a range of problems for your project or the people who use it.

Similar to https://github.com/containers/podman/pull/13565

#### What type of PR is this?

 /kind other

```release-note
None
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,3 +15,11 @@ updates:
     time: "10:00"
     timezone: Europe/Berlin
   open-pull-requests-limit: 10
+  
+  - package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: daily
+    time: "10:00"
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
Changes LGTM, but not my area of expertise.
I've restarted the tests, they looked flakey.

PR #3872, opened 2022-03-28
[CI:DOCS] Pin actions to a full length commit SHA

naveensrinivasan:
- Pinned actions by SHA https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies
- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

>Pin actions to a full length commit SHA

>Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload.

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

Also, dependabot supports upgrading based on SHA.

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

Similar to https://github.com/containers/podman/pull/13564 and was also asked to do a PR here https://github.com/containers/podman/pull/13564#issuecomment-1080878884

#### What type of PR is this?

 /kind other

None

```release-note
None

```

### DIFF
--- a/.github/workflows/check_cirrus_cron.yml
+++ b/.github/workflows/check_cirrus_cron.yml
@@ -71,7 +71,7 @@ jobs:
             - if: steps.cron.outputs.failures > 0
               name: Send failure notification e-mail
               # Ref: https://github.com/dawidd6/action-send-mail
-              uses: dawidd6/action-send-mail@v2.2.2
+              uses: dawidd6/action-send-mail@a80d851dc950256421f1d1d735a2dc1ef314ac8f # v2.2.2
               with:
                 server_address: ${{secrets.ACTION_MAIL_SERVER}}
                 server_port: 465
@@ -90,7 +90,7 @@ jobs:
 
             - if: failure()
               name: Send error notification e-mail
-              uses: dawidd6/action-send-mail@v2.2.2
+              uses: dawidd6/action-send-mail@a80d851dc950256421f1d1d735a2dc1ef314ac8f # v2.2.2
               with:
                 server_address: ${{secrets.ACTION_MAIL_SERVER}}
                 server_port: 465
--- a/.github/workflows/multi-arch-build.yaml
+++ b/.github/workflows/multi-arch-build.yaml
@@ -53,16 +53,16 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
+        uses: docker/setup-qemu-action@27d0a4f181a40b142cce983c5393082c365d1480 # v1
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@94ab11c41e45d028884a99163086648e898eed25 # v1
         with:
           driver-opts: network=host
           install: true
 
       - name: Build and locally push image
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@ac9327eae2b366085ac7f6a2d02df8aa8ead720a # v2
         with:
           context: contrib/${{ env.REPONAME }}image/${{ matrix.source }}
           file: ./contrib/${{ env.REPONAME }}image/${{ matrix.source }}/Dockerfile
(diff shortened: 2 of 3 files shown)


naveensrinivasan:
@cevich 👀  please

TomSweeneyRedHat:
An interesting twist on the sha versus version.  I've always liked the version, but I can see the reasoning.  How do you check that you have the right sha tthough?  i.e. Someone comes in later and says, there's a security issue, need to change the sha now, but then they point at a sha with the security issue....  I just worry that we secure one thing, but open another equally large door the other way.  Or maybe it's just to late for me to be thinking!  ;^)

naveensrinivasan:
I also updated the PR to include comments on checking the SHA with the GitHub URL. You don't need to trust me 😉

flouthoc:
Would this require more maintenance ? I think maintaining version is more easier as compared to exact SHA but I'll let @cevich decide.

cevich:
(Copy-paste from https://github.com/containers/podman/pull/13564)

Your concern is valid. Based on my understanding of the underlying issue, the argument is tags can be changed, but sha's cannot. So as long as trusted actors ensure the correct sha is used, any malicious actions on the tag are irrelevant. But you're right, and it's a PITA, and places more burden on the users of github-actions. Biting my tongue on my opinions on the system, additional security/safety nearly always comes at a cost of additional time, money, and headaches.

.github/workflows/check_cirrus_cron.yml
```diff
@@ -71,7 +71,7 @@ jobs:
             - if: steps.cron.outputs.failures > 0
               name: Send failure notification e-mail
               # Ref: https://github.com/dawidd6/action-send-mail
-              uses: dawidd6/action-send-mail@v2.2.2
+              uses: dawidd6/action-send-mail@a80d851dc950256421f1d1d735a2dc1ef314ac8f # v2.2.2 https://api.github.com/repos/dawidd6/action-send-mail/git/commits/a80d851dc950256421f1d1d735a2dc1ef314ac8f
```
cevich, line 74: I think it's valuable to keep line-length to a reasonable level.  While I appreciate the direct URL's purpose, I think the horizontal-scrolling cost is too high.  Would you mind trimming all these down to just the intended-tag comment?

It's easy and good-enough for us to `git rev-parse <tag> | grep <sha>` for any future changes.

cevich (changes requested):

.github/workflows/check_cirrus_cron.yml
naveensrinivasan: I have done that and removed the URL. PTAL. Thanks

cevich (changes requested):

.github/workflows/check_cirrus_cron.yml
```diff
@@ -71,7 +71,7 @@ jobs:
             - if: steps.cron.outputs.failures > 0
               name: Send failure notification e-mail
               # Ref: https://github.com/dawidd6/action-send-mail
-              uses: dawidd6/action-send-mail@v2.2.2
+              uses: dawidd6/action-send-mail@a80d851dc950256421f1d1d735a2dc1ef314ac8f # v2.2.2
```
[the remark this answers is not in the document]
naveensrinivasan: Fixed that 😆 , Thanks!

cevich:
Ugh, I don't enjoy being "that guy" :nerd_face: Please check the rest of the version-comment lines, nearly every one has trailing-white-space :cry:

naveensrinivasan:
> > Ugh, I don't enjoy being "that guy" 🤓 Please check the rest of the version-comment lines, nearly every one has trailing-white-space 😢
> 
> Thanks 🤞 I hope I got everything. PTAL. Thanks!

@cevich  Let me know. Thanks

cevich (approved):
LGTM

cevich:
And many :cookie: for taking care of this and putting up with me :grin:

Issue #3710, opened 2022-01-12
CI-Flake: bud-multiple-platform-no-run fails: error changing to intended-new-root directory 


cevich:
**Description**

With fairly high frequently this test is failing, but sometimes it passes.  This is affecting CI for all PRs for `main` and branch-testing itself.

**Steps to reproduce the issue:**
1. Run CI on a PR or `main` branch
2. `bud-multiple-platform-no-run test` fails ([example from PR 3706](https://cirrus-ci.com/task/4776405517991936?logs=integration_test#L1197))
3. Re-run the task, all tests pass (sometimes)

**Describe the results you received:**

```

[+0853s] # [linux/amd64] [2/2] COMMIT
[+0853s] # Getting image source signatures
...cut...
[+0853s] # [linux/s390x] [2/2] COMMIT
[+0853s] # Getting image source signatures
...cut...
[+0853s] # [linux/ppc64le] [2/2] COMMIT
[+0853s] # Writing manifest to image destination
...cut...
[+0853s] # Writing manifest to image destination
[+0853s] # Storing signatures
[+0853s] # 6df1a5aaabbbb36097fefb40bff149c3623dc9008c8f3708fcc1da7dd564bd58
[+0853s] # --> da98c4525b6
[+0853s] # da98c4525b600114fc89b8ced5fa3c4fb7f681adc65f9cfc31dc7e67640d839c
[+0853s] # error building at STEP "COPY --from=0 /root/Dockerfile.no-run /root/": checking on sources under "/var/tmp/buildah_tests.gbywtb/root/overlay/c0299fe974ccf3e964e1b623e5ee121afc0e63b08dfe9ad37da2b932f75cf0df/merged": error in copier subprocess: error changing to intended-new-root directory "/var/tmp/buildah_tests.gbywtb/root/overlay/c0299fe974ccf3e964e1b623e5ee121afc0e63b08dfe9ad37da2b932f75cf0df/merged": chdir /var/tmp/buildah_tests.gbywtb/root/overlay/c0299fe974ccf3e964e1b623e5ee121afc0e63b08dfe9ad37da2b932f75cf0df/merged: no such file or directory
[+0853s] # [ rc=125 (** EXPECTED 0 **) ]
[+0853s] # #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
[+0853s] # #| FAIL: exit code is 125; expected 0
[...]

PR #3874, opened 2022-03-29
imagebuildah: if user sets `--jobs=0` they expect no parallelism

flouthoc:
Lot of users and tests set --jobs=0 they are expecting no parallelism
between stages but we still end up running builds in parallel for
multi-arch builds which indirectly executes stages in parallel since builds are parallel.

Example tests like https://github.com/containers/buildah/blob/main/tests/bud.bats#L3609 set `--jobs=0` but they still execute build for similar name but different arch in parallel.

**[Note:]** This PR does not fix more that one race condition which is seen when multi-arch builds are performed without `--jobs=0` as documented here: https://github.com/containers/buildah/issues/3710#issuecomment-1081613407 but it should fix flakes in CI as reported here: https://github.com/containers/buildah/issues/3710

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -243,12 +243,22 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 			logPrefix = "[" + platforms.Format(platformSpec) + "] "
 		}
 		builds.Go(func() error {
+			isInstanceLocked := false
+			if options.Jobs != nil {
+				// if --jobs=0 all the stages and builds will be sequiential
+				if *options.Jobs == 0 {
+					instancesLock.Lock()
+					isInstanceLocked = true
+				}
+			}
 			thisID, thisRef, err := buildDockerfilesOnce(ctx, store, logger, logPrefix, platformOptions, paths, files)
 			if err != nil {
 				return err
 			}
 			id, ref = thisID, thisRef
-			instancesLock.Lock()
+			if !isInstanceLocked {
+				instancesLock.Lock()
+			}
 			instances = append(instances, instance{
 				ID:       thisID,
 				Platform: platformSpec,


flouthoc:
This should fix the flake as tests in CI use `--jobs=0` but this does not means that actual race is fixed which can be seen again without `--jobs=0` for `multi-arch` builds.

nalind:
> Lot of users and tests set --jobs=0 they are expecting no parallelism between stages but we still end up running builds in parallel for multi-arch builds which indirectly executes stages in parallel since builds are parallel.
> 
> Example tests like https://github.com/containers/buildah/blob/main/tests/bud.bats#L3609 set `--jobs=0` but they still execute build for similar name but different arch in parallel.

That's why that test specifies `--jobs=0`.  The default is to behave as though `--jobs=1` was specified.

flouthoc:
@nalind If I inferred it correctly `--jobs=0` should mean no parallelism at all but we still do builds for each platform in parallel manner.

nalind:
The intent of that test is definitely maximum parallelization.

flouthoc:
@nalind Oh i see but then I think users have no way to serialize `mult-arch` builds.

cevich:
> users have no way to serialize mult-arch builds.

I think it's incorrect to equate serialized builds with serialized pulls, one doesn't necessarily imply the other.  Though I appreciate it vastly complicates the fix.  As a user doing an operation requiring a multi-arch pull, I expect that pull to happen as fast as my bandwidth allows - meaning in parallel when possible.

flouthoc:
> > users have no way to serialize mult-arch builds.
> 
> I think it's incorrect to equate serialized builds with serialized pulls, one doesn't necessarily imply the other. Though I appreciate it vastly complicates the fix. As a user doing an operation requiring a multi-arch pull, I expect that pull to happen as fast as my bandwidth allows - meaning in parallel when possible.

@cevich Pulls are not causing the issue but i think builds are causing the issue, its easily reproduce-able  when you run `bud-multiple-platform-no-run` a few times.

I'm going to close this PR since it is not doing what i expected it to do.

PR #3877, opened 2022-03-30
manifest-add: allow overriding `--arch` and `--variant` while adding custom image to manifest list

flouthoc:
While adding image to manifest list if we ever fallback to scenario
where we are looking for image in local storage then do not override
system context with the one generated from parsing options, instead pass `nil`
and let `libimage` decide.

Closes: https://github.com/containers/buildah/issues/3511

### DIFF
--- a/cmd/buildah/manifest.go
+++ b/cmd/buildah/manifest.go
@@ -328,12 +328,15 @@ func manifestAddCmd(c *cobra.Command, args []string, opts manifestAddOpts) error
 	digest, err := list.Add(getContext(), systemContext, ref, opts.all)
 	if err != nil {
 		var storeErr error
-		// check if the local image exists
-		if ref, _, storeErr = util.FindImage(store, "", systemContext, imageSpec); storeErr != nil {
+		// Retry without a custom system context.  A user may want to add
+		// a custom platform (see #3511).
+		if ref, _, storeErr = util.FindImage(store, "", nil, imageSpec); storeErr != nil {
+			logrus.Errorf("Error while trying to find image on local storage: %v", storeErr)
 			return err
 		}
 		digest, storeErr = list.Add(getContext(), systemContext, ref, opts.all)
 		if storeErr != nil {
+			logrus.Errorf("Error while trying to add on manifest list: %v", storeErr)
 			return err
 		}
 	}
--- a/tests/lists.bats
+++ b/tests/lists.bats
@@ -180,3 +180,20 @@ IMAGE_LIST_S390X_INSTANCE_DIGEST=sha256:882a20ee0df7399a445285361d38b711c299ca09
     # Make sure we can add the new image to the list.
     run_buildah manifest add test-list $(< ${TESTDIR}/image-id.txt)
 }
+
+@test "manifest-add-to-list-from-storage" {
+    run_buildah pull --arch=amd64 busybox
+    run_buildah tag busybox test:amd64
+    run_buildah pull --arch=arm64 busybox
+    run_buildah tag busybox test:arm64
+    run_buildah manifest create test
+    run_buildah manifest add test test:amd64
+    run_buildah manifest add --variant=variant-something test test:arm64
+    run_buildah manifest inspect test
+    # must contain amd64
+    expect_output --substring "amd64"
+    # must contain arm64
+    expect_output --substring "arm64"
+    # must contain variant v8
+    expect_output --substring "variant-something"
+}


giuseppe:
@vrothberg PTAL

tests/lists.bats
```diff
@@ -180,3 +180,20 @@ IMAGE_LIST_S390X_INSTANCE_DIGEST=sha256:882a20ee0df7399a445285361d38b711c299ca09
...
+	run_buildah manifest create test
+	run_buildah manifest add test test:amd64
+	run_buildah manifest add --variant=v8 test test:arm64
+	run_buildah manifest inspect test
+	# must contain amd64
+        expect_output --substring "amd64"
+	# must contain arm64
+        expect_output --substring "arm64"
+	# must contain variant v8
+        expect_output --substring "v8"
```
vrothberg, line 198: Why is the indentation different?

cmd/buildah/manifest.go
```diff
@@ -328,13 +328,22 @@ func manifestAddCmd(c *cobra.Command, args []string, opts manifestAddOpts) error
...
+		// at a given point of time there will be only one image
+		// mapped by name so don't set systemContext otherwise
+		// lookup will fail if user wants to set custom arch or
+		// variant for this image. Instead perform a generic lookup
+		// on host since `buildah manifest add --variant <> list` or
+		// `buildah manifest add --arch <> list` expects to override
+		// arch or variant and not perform a lookup by that.
+		// See: https://github.com/containers/buildah/issues/3511
+		if ref, _, storeErr = util.FindImage(store, "", nil, imageSpec); storeErr != nil {
+			return storeErr
```
vrothberg, line 342:
```suggestion
			return err
```
Maybe log `storeErr` but `err` _must_ be returned and not entirely suppressed.

cmd/buildah/manifest.go
```diff
@@ -328,13 +328,22 @@ func manifestAddCmd(c *cobra.Command, args []string, opts manifestAddOpts) error
...
+		// Add operation for list failed with error as an alternative
+		// we are looking for image in local storage by name,
+		// at a given point of time there will be only one image
+		// mapped by name so don't set systemContext otherwise
+		// lookup will fail if user wants to set custom arch or
+		// variant for this image. Instead perform a generic lookup
+		// on host since `buildah manifest add --variant <> list` or
+		// `buildah manifest add --arch <> list` expects to override
+		// arch or variant and not perform a lookup by that.
+		// See: https://github.com/containers/buildah/issues/3511
```
vrothberg, line 340:
```suggestion
		// Retry without a custom system context.  A user may want to add
		// a custom platform (see #3511).
```
I find the comment too verbose.

vrothberg (commented):
This fix seems to work but I wonder why the --arch in https://github.com/containers/buildah/issues/3511 is at all set in the system context.

IMHO, it clearly refers to the _entry_ in the manifest list but not to the local image.

flouthoc:
> This fix seems to work but I wonder why the --arch in https://github.com/containers/buildah/issues/3511 is at all set in the system context.

I think we create system context from parsing cli args which we should do for all cases but not  when looking image from local storage since we perform lookup by name it should just return whatever is in there on storage without doing any filtering and instead use `--arch` and `--variant` to override what's present in manifest list, this used to work before but something regressed in between.

flouthoc:
@vrothberg PTAL

vrothberg (commented):
LGTM, thanks @flouthoc

vrothberg:
@nalind PTAL

tests/lists.bats
```diff
@@ -180,3 +180,20 @@ IMAGE_LIST_S390X_INSTANCE_DIGEST=sha256:882a20ee0df7399a445285361d38b711c299ca09
...
+    run_buildah manifest create test
+    run_buildah manifest add test test:amd64
+    run_buildah manifest add --variant=v8 test test:arm64
+    run_buildah manifest inspect test
+    # must contain amd64
+    expect_output --substring "amd64"
+    # must contain arm64
+    expect_output --substring "arm64"
+    # must contain variant v8
+    expect_output --substring "v8"
```
nalind, line 198: This check will stop being meaningful at some point, as the image we're added to the manifest will eventually be marked with variant=v8 in its image configuration (to confirm: `skopeo --override-arch arm64 inspect --config --raw docker://docker.io/library/busybox` -- the registries.conf we're using redirects us to quay.io/libpod/busybox, which is an older version that didn't include that field in its configuration, but I don't expect that to be the case forever).  I'd suggest specifying a value that's unlikely to be in the image on line 191, and then checking for that here.

flouthoc: I agree changed it to arbitrary variant and fixed in latest commit.

PR #3880, opened 2022-03-30
[WIP] Skip frequently flaking bud-multiple-platform-no-run

cevich:
#### What type of PR is this?

> /kind flake

The `bud-multiple-platform-no-run` test seems to be flaking more and more frequently, disable it temporarily while
https://github.com/containers/buildah/issues/3710 is researched.

#### How to verify it

CI will pass and not run `bud-multiple-platform-no-run` test

None

Opening this PR preemptively, in case flake-rate becomes intolerable.   See also https://github.com/containers/podman/pull/13718 for podman.

None

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -3607,6 +3607,7 @@ _EOF
 }
 
 @test "bud-multiple-platform-no-run" {
+  skip "Frequently flakes: https://github.com/containers/buildah/issues/3710"
   outputlist=localhost/testlist
   run_buildah build --signature-policy ${TESTSDIR}/policy.json --jobs=0 --all-platforms --manifest $outputlist -f ${TESTSDIR}/bud/multiarch/Dockerfile.no-run ${TESTSDIR}/bud/multiarch
   run_buildah manifest inspect $outputlist


edsantiago:
LGTM

flouthoc (approved):
LGTM
/lgtm

flouthoc:
@cevich Is something needs to be changed in this PR. There is a [WIP] tag.

edsantiago:
@flouthoc the WIP is intentional (here and in the podman PR). We do not want to merge this unless there is absolutely no other choice.

flouthoc:
@edsantiago ackd I agree.

rhatdan:
This is blowing up all over the place, do we have any idea what is happening?

flouthoc:
@rhatdan I think we are hitting multiple race in this bug and I am still trying to catch things but no luck till now. We are discussing this here: https://github.com/containers/buildah/issues/3710

flouthoc:
Instead of skipping test I think we can keep the test but set `--jobs=1` it should stop flake in CI but not fix the issue: https://github.com/containers/buildah/pull/3900

cevich:
>Instead of skipping test I think we can keep the test but set --jobs=1 it should stop flake in CI but not fix the issue: 

I'm in favor of this, it's one-patch to revert later instead of the two required to skip the test.  I think @flouthoc needs some breathing room as this issue seems to just keep getting more and more complex.

rhatdan:
Since the jobs==1 patch got merged, I am closing this.

PR #3882, opened 2022-03-30
copier test: use correct UID/GID in test archives

nalind:
#### What type of PR is this?

/kind failing-test 

Handing `copier.Put()` an archive which uses host IDs while also including ID maps for mapping from container IDs to host IDs is a mistake, and the unit tests have been failing when run by non-root users since we updated to storage v1.38.0, which started flagging it as such.

#### How to verify it

Updated test cases for unit tests!

None

```release-note
None
```


rhatdan:
LGTM

flouthoc (approved):
LGTM
/lgtm
/approve

PR #3885, opened 2022-03-31
fix hang when oci runtime fails

Luap99:
#### What type of PR is this?

 /kind bug

When the run child process exits early the parent still tries to read
from the ready pipe. Reading from the pipe should end when all writers
are closed. The problem is that the parent kept the writer open as well.
To fix the hang we just need to close the writer before we try to read
and after we gave it to the child.

To prevent closing the fd twice with defer I added a new fileCloser type
to store if the file was already closed.

This problem was noticed in the podman CI:
https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6624611893772288/html/sys-podman-fedora-35-root-host-netavark.log.html

```release-note

```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -2284,23 +2284,23 @@ func (b *Builder) runUsingRuntimeSubproc(isolation define.Isolation, options Run
 	}()
 
 	// create network configuration pipes
-	var containerCreateR, containerCreateW *os.File
-	var containerStartR, containerStartW *os.File
+	var containerCreateR, containerCreateW fileCloser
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@nalind @flouthoc @giuseppe @vrothberg PTAL

vrothberg (approved):
/lgtm
/hold

TomSweeneyRedHat:
LGTM
I reran the failing test, looked like the multi flake.

PR #3875, opened 2022-03-29
Vendor in opencontainers/selinux v1.10.1

rhatdan:
Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### Does this PR introduce a user-facing change?

```release-note

```

PR #3889, opened 2022-04-01
selinux: don't use `lsetxattr` on `/proc/self/fd/%d` and bump selinux to `v1.10.1`

flouthoc:
* Syscall `lsetxattr` always fails with `ENOTSUP` when it tries to relabel
`/proc/self/fd/%d` but in this case we are acutally interested in the
actual file pointed by the `/proc/self/fd/%d` not in the `symlink` 
( since that was the behavior before ) so  use `Chcon` instead of `Relabel` 
since `Relabel` was configured  here https://github.com/opencontainers/selinux/pull/173 
to use `lsetxattr` instead of `setxattr`.

* Bump selinux to `v1.10.1` to verify if it works

Fixes and replaces: https://github.com/containers/buildah/pull/3875
See: https://github.com/containers/common/pull/994

[NO NEW TESTS NEEDED]: This does not adds new functionality but all existing tests must pass with selinux `v.1.10.1` as of now they all break when selinux is bumped to `v.1.10.1`

### DIFF
--- a/selinux.go
+++ b/selinux.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/opencontainers/runtime-tools/generate"
 	selinux "github.com/opencontainers/selinux/go-selinux"
-	"github.com/opencontainers/selinux/go-selinux/label"
 	"github.com/pkg/errors"
 )
 
@@ -33,7 +32,7 @@ func runLabelStdioPipes(stdioPipe [][]int, processLabel, mountLabel string) erro
 	}
 	for i := range stdioPipe {
 		pipeFdName := fmt.Sprintf("/proc/self/fd/%d", stdioPipe[i][0])
-		if err := label.Relabel(pipeFdName, pipeContext, false); err != nil {
+		if err := selinux.Chcon(pipeFdName, pipeContext, false); err != nil {
 			return errors.Wrapf(err, "setting file label on %q", pipeFdName)
 		}
 	}
--- a/go.mod
+++ b/go.mod
@@ -26,7 +26,7 @@ require (
 	github.com/opencontainers/runc v1.1.1
 	github.com/opencontainers/runtime-spec v1.0.3-0.20210326190908-1c3f411f0417
 	github.com/opencontainers/runtime-tools v0.9.0
-	github.com/opencontainers/selinux v1.10.0
+	github.com/opencontainers/selinux v1.10.1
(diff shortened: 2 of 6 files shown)


flouthoc:
@rhatdan @giuseppe @nalind @TomSweeneyRedHat PTAL. This should fix selinux issue both in buildah and podman.

selinux.go
```diff
@@ -33,7 +32,7 @@ func runLabelStdioPipes(stdioPipe [][]int, processLabel, mountLabel string) erro
 	}
 	for i := range stdioPipe {
 		pipeFdName := fmt.Sprintf("/proc/self/fd/%d", stdioPipe[i][0])
-		if err := label.Relabel(pipeFdName, pipeContext, false); err != nil {
+		if err := selinux.Chcon(pipeFdName, pipeContext, false); err != nil {
```
giuseppe, line 35: `label.Relabel` ends up calling `selinux.Chcon`, but it checks if selinux is enabled.

Why do we need this change?

flouthoc: @giuseppe `Relabel` calls `Chcon` but ends up hard-coding recurse flag to true https://github.com/opencontainers/selinux/blob/main/go-selinux/label/label_linux.go#L169, which is now configured to use `lsetxattr(` always instead of `setxattr` ( the change was done here https://github.com/opencontainers/selinux/pull/173 ) and for somereason `lsetxattr` always gives `operation not supported` for both rootless and root cases. I think that's because `lsetxattr` tries to label this special `symlink` `/proc/self/fd/%d` but instead it should try to label the actual file.

Failure is easy to reproduce if you see https://github.com/containers/podman/pull/13687 and https://github.com/containers/buildah/pull/3875

giuseppe: thanks.

What happens now if we try to relabel a real symlink (not the magic `/proc/self/fd` one)?  Do we end up relabelling the file it points to?

rhatdan:
Let's concentrate on https://github.com/containers/buildah/pull/3875

I now duplicate the previous behavior, reading the symlink if it exists and added an IsNotExists check, and it seems to fix the problem.

selinux.go
flouthoc: @giuseppe Yes I think now we follow the symlink via https://github.com/opencontainers/selinux/blob/main/go-selinux/selinux_linux.go#L338.

PR #3894, opened 2022-04-01
integration tests: learn to start a dummy registry

nalind:
#### What type of PR is this?

/kind other

When a test needs to talk to a registry server, launch one as part of the test rather than depending on it having been started by someone else.

Use run_buildah where we used to use 'run buildah' without checking the return code, and in a few cases where we did check it.

In the "from with non buildah container" test, use "podman create" with host networking, in an attempt to avoid messing with networking in cases where we're running on a system with a version of podman that will create a bridge with CNI that we'll also create with netavark.  We're not sharing storage between the two invocations, so the logic that tries to detect this problem won't detect it.

#### How to verify it

Updates to tests!

None

This gets us closer to being able to run integration tests anywhere by pointing `bats` at our `tests` directory.

```release-note
None
```

### DIFF
--- a/tests/copy.bats
+++ b/tests/copy.bats
@@ -446,24 +446,36 @@ stuff/mystuff"
 
 @test "copy-preserving-extended-attributes" {
   createrandom ${TESTDIR}/randomfile
+  # if we need to change which image we use, any image that can provide a working setattr/setcap/getfattr will do
   image="quay.io/libpod/fedora-minimal:34"
+  if ! which setfattr > /dev/null 2> /dev/null; then
+    skip "setfattr not available, unable to check if it'll work in filesystem at ${TESTDIR}"
+  fi
+  run setfattr -n user.yeah -v butno ${TESTDIR}/root
+  if [ "$status" -ne 0 ] ; then
+    if [[ "$output" =~ "not supported" ]] ; then
+      skip "setfattr not supported in filesystem at ${TESTDIR}"
+    fi
+    skip "$output"
+  fi
   _prefetch $image
   run_buildah from --quiet --signature-policy ${TESTSDIR}/policy.json $image
   first="$output"
-  run_buildah run $first microdnf -y install /usr/bin/getfattr /usr/bin/setfattr /usr/sbin/setcap
+  run_buildah run $first microdnf -y install /usr/bin/setfattr /usr/sbin/setcap
   run_buildah copy $first ${TESTDIR}/randomfile /
   # set security.capability
-  run buildah run $first setcap cap_setuid=ep /randomfile
+  run_buildah run $first setcap cap_setuid=ep /randomfile
   # set user.something
-  run buildah run $first setfattr user.yeah=butno /randomfile
+  run_buildah run $first setfattr -n user.yeah -v butno /randomfile
   # copy the file to a second container
   run_buildah from --quiet --signature-policy ${TESTSDIR}/policy.json $image
   second="$output"
+  run_buildah run $second microdnf -y install /usr/bin/getfattr
   run_buildah copy --from $first $second /randomfile /
   # compare what the extended attributes look like. if we're on a system with SELinux, there's a label in here, too
-  run buildah run $first sh -c "getfattr -d -m . --absolute-names /randomfile | sort"
+  run_buildah run $first sh -c "getfattr -d -m . --absolute-names /randomfile | grep -v ^security.selinux | sort"
   expected="$output"
-  run buildah run $first sh -c "getfattr -d -m . --absolute-names /randomfile | sort"
+  run_buildah run $second sh -c "getfattr -d -m . --absolute-names /randomfile | grep -v ^security.selinux | sort"
   expect_output "$expected"
 }
 
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -84,6 +84,7 @@ function teardown(){
 function teardown_tests() {
     stophttpd
     stop_git_daemon
+    stop_registry
 
     # Workaround for #1991 - buildah + overlayfs leaks mount points.
     # Many tests leave behind /var/tmp/.../root/overlay and sub-mounts;
@@ -118,13 +119,18 @@ function _prefetch() {
         mkdir -p ${_BUILDAH_IMAGE_CACHEDIR}
     fi
 
+    local storage=
     for img in "$@"; do
+        if [[ "$img" =~ '[vfs@' ]] ; then
+            storage="$img"
+            continue
+        fi
         img=$(normalize_image_name "$img")
         echo "# [checking for: $img]" >&2
         fname=$(tr -c a-zA-Z0-9.- - <<< "$img")
         if [ -d $_BUILDAH_IMAGE_CACHEDIR/$fname ]; then
             echo "# [restoring from cache: $_BUILDAH_IMAGE_CACHEDIR / $img]" >&2
-            copy dir:$_BUILDAH_IMAGE_CACHEDIR/$fname containers-storage:"$img"
+            copy dir:$_BUILDAH_IMAGE_CACHEDIR/$fname containers-storage:"$storage""$img"
         else
             rm -fr $_BUILDAH_IMAGE_CACHEDIR/$fname
             echo "# [copy docker://$img dir:$_BUILDAH_IMAGE_CACHEDIR/$fname]" >&2
@@ -134,8 +140,8 @@ function _prefetch() {
                 fi
                 sleep 5
             done
-            echo "# [copy dir:$_BUILDAH_IMAGE_CACHEDIR/$fname containers-storage:$img]" >&2
-            copy dir:$_BUILDAH_IMAGE_CACHEDIR/$fname containers-storage:"$img"
+            echo "# [copy dir:$_BUILDAH_IMAGE_CACHEDIR/$fname containers-storage:$storage$img]" >&2
+            copy dir:$_BUILDAH_IMAGE_CACHEDIR/$fname containers-storage:"$storage""$img"
         fi
     done
 }
@@ -157,7 +163,7 @@ function copy() {
 }
 
 function podman() {
-    command podman ${PODMAN_REGISTRY_OPTS} ${ROOTDIR_OPTS} "$@"
+    command ${PODMAN_BINARY:-podman} ${PODMAN_REGISTRY_OPTS} ${ROOTDIR_OPTS} "$@"
 }
 
 # There are various scenarios where we would like to execute `tests` as rootless user, however certain commands like `buildah mount`
@@ -505,6 +511,16 @@ function skip_if_no_runtime() {
     skip "runtime \"$OCI\" not found"
 }
 
+#######################
+#  skip_if_no_podman  #  we need 'podman' to test how we interact with podman
+#######################
+function skip_if_no_podman() {
+    run which ${PODMAN_BINARY:-podman}
+    if [[ $status -ne 0 ]]; then
+        skip "podman is not installed"
+    fi
+}
+
 ##################
 #  is_cgroupsv2  #  Returns true if host system has cgroupsv2 enabled
 ##################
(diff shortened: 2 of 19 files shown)


tests/helpers.bash
```diff
@@ -559,3 +565,125 @@ function stop_git_daemon() {
...
+# Bring up a registry server using buildah with vfs and chroot as a cheap
+# substitute for podman, accessible only to user $1 using password $2 on the
+# local system at a dynamically-allocated port.
+# Requires openssl.
+# If a user name and password are supplied as the two parameters, htpasswd
+# should be available so that they can be hashed.
+# Sets REGISTRY_PID, REGISTRY_PORT (to append to "localhost:"), and
+# REGISTRY_DIR (where the CA cert can be found) on success.
+function start_registry() {
+  local REGISTRY_IMAGE=docker.io/library/registry:2
```
TomSweeneyRedHat, line 578: I think we may call this a fair number of times during the tests and I wonder if we'll hit the rate limits.  Would it make sense to have our own registry image tucked away on quay.io somewhere?  Do they have one for Podman already?

nalind: I'm not against that, but I have no idea if we have a plan in place to keep any of those up to date.  The start_registry function uses _prefetch, so it _should_ be pulled, at most, once in a given CI job.

vrothberg: Consider using `quay.io/libpod/registry:2.6`. I recall that an update to the one on Docker Hub once broke gating.

nalind: Using `2.7` now.  `2.6` and `2.7` are not multi-arch, and `2.8` isn't in that repository.  How/when do they get updated?

vrothberg: @edsantiago, feel free to ping me next time. Still enjoying fast French fibre :)

edsantiago: Done:
```console
$ buildah manifest inspect quay.io/libpod/registry:2.8 | jq '.manifests[].platform.architecture'
"amd64"
"arm"
"arm"
"arm64"
"ppc64le"
"s390x"
```

edsantiago:
This one is failing on `setxattr` (without the `l`)

nalind:
> This one is failing on `setxattr` (without the `l`)

This was trying to re-enable a test that was previously disabled under SELinux, but I guess we didn't work around it after all.

TomSweeneyRedHat:
The integration test is failing with one that I've not seen before:

```
[+1249s] not ok 286 bud-multiple-platform-no-run
...
[+1249s] # 9505e448a8192a7713bf10737028537f74f36a1fb89298e8b4914548eaca053c
[+1249s] # --> dc60ecdbd35
[+1249s] # dc60ecdbd35d1c5a9e89efd7f77f8c820d97643f3a5e4c03e2bf4d23e670a5b3
[+1249s] # error creating build container: writing blob: adding layer with blob "sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1": ApplyLayer exit status 1 stdout:  stderr: Error making old root private after pivot: no such file or directory
[+1249s] # [ rc=125 (** EXPECTED 0 **) ]
```

Later, it can't find containers.conf in a few tests:

```
[+1532s] ok 364 containers.conf selinux test
[+1532s] time="2022-04-04T16:56:02-05:00" level=warning msg="Error loading container config when searching for local runtime: finding config on system: CONTAINERS_CONF file: stat /var/tmp/go/src/github.com/containers/buildah/tests/./containers1.conf: no such file or directory"
[+1532s] time="2022-04-04T16:56:02-05:00" level=error msg="failed to setup From and Build flags: failed to get container config: finding config on system: CONTAINERS_CONF file: stat /var/tmp/go/src/github.com/containers/buildah/tests/./containers1.conf: no such file or directory"
```

edsantiago:
@TomSweeneyRedHat that's our nemesis, #3710

tests/copy.bats
```diff
@@ -446,24 +446,35 @@ stuff/mystuff"
 
 @test "copy-preserving-extended-attributes" {
   createrandom ${TESTDIR}/randomfile
-  image="quay.io/libpod/fedora-minimal:34"
+  image="registry.fedoraproject.org/fedora-minimal:35"
```
TomSweeneyRedHat, line 449: A thought for later.  Should we define `${FEDORA_MINIMAL}` or some such in helpers.bat?  That way if we need to bump from 35 to 36, we only have one place to do it.

TomSweeneyRedHat:
LGTM

tests/copy.bats
nalind: This could as likely have left the version tag off of the image spec, since the test doesn't care about it as much, but don't let me stop you.

edsantiago: The problem with omitting the version spec is that the next (or next-next) version will introduce some sort of breakage that will cause us to scramble in a panic. . Example: https://github.com/containers/podman/pull/12343

TomSweeneyRedHat:
Happy green test buttons.

tests/helpers.bash
```diff
@@ -568,3 +574,125 @@ function stop_git_daemon() {
...
+  htpasswd:
+    realm: buildah-realm
+    path: /etc/docker/registry/htpasswd
+'
+  # use the result of "htpasswd -nbB testuser testpassword" if an alternate
+  # username and password are not specified
+  local htpasswd='testuser:$2y$05$OW6Qlf1ygRuqJxt/CYcBq.0MZARYLIvr.mgjksw2m7K4cwWkw3Pda'
+  if test "$1" = testuserfoo && test "$2" = testpassword ; then
+    htpasswd='testuserfoo:$2y$05$Pkly/9sA8iGY..2SaerFH.wr4qGPPaGZ66KMrvVmcCP8gaRPpeoeC'
+  elif test "${1:-testuser}" != testuser && test "${2:-testpassword}" != testpassword ; then
```
edsantiago, line 620: Isn't this just a complicated way of saying:
```bash
    local testuser="${1:-testuser}"
    local testpass="${2:-testpassword}"
```

tests/helpers.bash
```diff
@@ -568,3 +574,125 @@ function stop_git_daemon() {
...
+    htpasswd=$(htpasswd -nbB "$1" "$2")
+    if test $? -ne 0 ; then
+      echo error computing hashed password
+      exit 1
```
edsantiago, line 624: How about `die "Error computing hashed password"`?

nalind: That'll work.  Changing it.

tests/helpers.bash
```diff
@@ -568,3 +574,125 @@ function stop_git_daemon() {
...
+  # generate the htpasswd and config.yml files for the registry
+  mkdir -p "${TESTDIR}"/registry/root "${TESTDIR}"/registry/run "${TESTDIR}"/registry/certs "${TESTDIR}"/registry/config
+  cat > "${TESTDIR}"/registry/config/htpasswd <<< "$htpasswd"
+  cat > "${TESTDIR}"/registry/config/config.yml <<< "$config"
+  chmod 644 "${TESTDIR}"/registry/config/htpasswd "${TESTDIR}"/registry/config/config.yml
+
+  # generate a new key and certificate
+  if ! openssl req -newkey rsa:4096 -nodes -sha256 -keyout "${TESTDIR}"/registry/certs/localhost.key -x509 -days 2 -addext "subjectAltName = DNS:localhost" -out "${TESTDIR}"/registry/certs/localhost.crt -subj "/CN=dummy registry" ; then
+    echo error creating new key and certificate
+    return 1
```
edsantiago, line 637: Likewise, `die` might be better

nalind: Changing it.

tests/helpers.bash
```diff
@@ -568,3 +574,125 @@ function stop_git_daemon() {
...
+    cat ${TESTDIR}/registry/registry.log
+    stop_registry
+    return 1
+  fi
+
+  # push the registry image we just started... to itself, as a confidence check
+  if ! ${BUILDAH_BINARY} --storage-driver vfs --root "${REGISTRY_DIR}"/root --runroot "${REGISTRY_DIR}"/run push --cert-dir "${REGISTRY_DIR}" --creds "${1:-testuser}":"${2:-testpassword}" "${REGISTRY_IMAGE}" localhost:"${REGISTRY_PORT}"/registry; then
+    echo error pushing to /registry repository at localhost:$REGISTRY_PORT
+    stop_registry
+    return 1
```
edsantiago, line 680: likewise, false

tests/helpers.bash
```diff
@@ -568,3 +574,125 @@ function stop_git_daemon() {
...
+  fi
+
+  return 0
```
edsantiago, line 683: unnecessary

nalind: Dropping it.

tests/helpers.bash
nalind: Changing it.

tests/helpers.bash
```diff
@@ -568,3 +574,125 @@ function stop_git_daemon() {
...
+  # we're separating all of this from the storage for any test that might call
+  # this function and using vfs to minimize the cleanup required
+  REGISTRY_PID="${COPROC_PID}"
+  REGISTRY_PORT=$(grep "listening on" ${TESTDIR}/registry/registry.log | sed -e 's^.*listening on.*:^^g' -e 's^,.*^^g')
+  REGISTRY_DIR="${TESTDIR}"/registry
+  if test -z "${REGISTRY_PORT}" ; then
+    echo error determining listening port from log:
+    cat ${TESTDIR}/registry/registry.log
+    stop_registry
+    return 1
```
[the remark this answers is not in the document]
nalind: Changing it.  No worries, I'll try to make sure it's all sorted before the next rebase.

tests/helpers.bash
nalind: It's an attempt to sort of future-proof that, but thinking on it more, it's better to just avoid needing `htpasswd`, so I'll drop that case.

edsantiago: htpasswd is already a hard requirement for the `buildah-tests` package, in both fedora and rhel, because a registry is already mandatory for running tests (and has been from day one).

tests/copy.bats
edsantiago: See also #3640. If there is a critical need to use f35 here, can we push a copy to quay instead?

tests/helpers.bash
nalind: We check in "authenticate: cert and credentials" that we can successfully authenticate to the registry with a known-good username/password pair, and that we get an error when we intentionally supply values that we know the registry isn't configured to accept.  What types of bugs do you see us catching by randomly selecting unique known-good values for each test?

edsantiago: Mostly my paranoia: with static values (credentials, input/output test strings) there's always a chance that you're not really testing what you think you're testing: that there's a leftover process somewhere and you're talking to it instead of the registry/httpserver/whatever that you think you're talking to. With unique values that worry is diminished.

edsantiago (commented):
LGTM with a few suggestions. Thanks for doing this, it greatly simplifies our gating-test setup.

tests/copy.bats
nalind: It isn't critical that it be Fedora or Fedora-based, more that it have working `setcap`, `getfattr`, and `setfattr` commands.  I'll add comment there to try to clarify that.  If you've another image in mind that would work better, I'm happy to switch to using it.

edsantiago: My question was, is `fedora-minimal:34` not working? If it no longer works, can we push a new image to quay? If it still works, can we keep it?

nalind: I guess it will.  I switched to 35 mainly because I know 34 is going EOL in a little over a month.  Changing it back to 34.

tests/helpers.bash
nalind: Hmm, I still like the idea of reducing the set of requirements that someone needs to have installed in order to run the tests, and it turns out we can generate the hashes for passwords ourselves by wrapping the right function call.  I'll add that and switch the "authenticate: cert and credentials" test to randomize them.

tests/helpers.bash
```diff
@@ -568,3 +574,125 @@ function stop_git_daemon() {
...
+  # wait for it to start logging things, then give it a second more
+  local waited=0
+  while ! test -s "${TESTDIR}"/registry/registry.log ; do
+    if test $waited -ge $BUILDAH_TIMEOUT ; then
+      break
+    fi
+    sleep 1
+    waited=$((${waited}+1))
+  done
+  sleep 1
```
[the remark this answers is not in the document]
nalind: Taking this almost verbatim.

PR #3899, opened 2022-04-05
build: automatically set correct `$TARGETPLATFORM` where expected.

flouthoc:
Automatically set correct value of TARGETPLATFORM if one or more
`--platform` are selected for the build. This ensures that behaviour
of buildah is equivalent to `buildkit` for the args `$BUILDPLATFORM` and
`$TARGETPLATFORM`.

Where `$BUILDPLATFORM` is set always native to host and `$TARGETPLATFORM`
is foreign selected os/arch via `--platform`.

Example

```console
$ buildah build --platform linux/arm64 .
```

```Dockerfile
FROM --platform=$BUILDPLATFORM alpine
ARG TARGETPLATFORM
ARG BUILDPLATFORM
RUN echo "I'm compiling for $TARGETPLATFORM on $BUILDPLATFORM and tagging for $TARGETPLATFORM"
```

* PR also adds a test to verify correct behaviour of `$BUILDPLATFORM` and `$TARGETPLATFORM` for cases using inline `FROM --platform=`

Wait for: https://github.com/openshift/imagebuilder/pull/224


imagebuildah/build.go
```diff
@@ -340,6 +340,15 @@ func buildDockerfilesOnce(ctx context.Context, store storage.Store, logger *logr
 
 	warnOnUnsetBuildArgs(logger, mainNode, options.Args)
 
+	// --platform was explicitly selected for this build
+	// so set correct TARGETPLATFORM in args if it is not
+	// already selected by the user.
+	if options.OS != "" && options.Architecture != "" {
+		if _, ok := options.Args["TARGETPLATFORM"]; !ok {
+			options.Args["TARGETPLATFORM"] = options.OS + "/" + options.Architecture
```
nalind, line 348: The `options.SystemContext` will have these values, and the variant value, too, if it's set.

flouthoc: Yes looks better, fixed.

TomSweeneyRedHat:
LGTM, once a new IB is built and vendored.  I've thrown a WIP tag on this for now.

flouthoc:
Since this is merged https://github.com/openshift/imagebuilder/pull/224 , pointing to latest imagebuilder and marking this PR ready for review.

nalind:
The imagebuilder logic also predefines TARGETOS, TARGETARCH, and TARGETVARIANT args.  Are those already correct, or do they likewise need to be fixed up at the same time?

flouthoc:
@nalind yes we were also not setting `TARGETOS`, `TARGETARCH` and `TARGETVARIANT` correctly added it now.

imagebuildah/build.go
```diff
@@ -340,6 +340,32 @@ func buildDockerfilesOnce(ctx context.Context, store storage.Store, logger *logr
...
+			options.Args["TARGETARCH"] = options.SystemContext.ArchitectureChoice
+		}
+		// variant component from --platform string populates TARGETVARIANT
+		if _, ok := options.Args["TARGETVARIANT"]; !ok {
+			if options.SystemContext.VariantChoice != "" {
+				options.Args["TARGETVARIANT"] = options.SystemContext.VariantChoice
+			}
+		}
+		if _, ok := options.Args["TARGETPLATFORM"]; !ok {
+			options.Args["TARGETPLATFORM"] = options.SystemContext.OSChoice + "/" + options.SystemContext.ArchitectureChoice
```
rhatdan, line 362: Shouldn't this be:
options.Args["TARGETPLATFORM"] = options.Args["TARGETOS"] + "/" + options.Args["TARGETARCH"]

imagebuildah/build.go
```diff
@@ -340,6 +340,32 @@ func buildDockerfilesOnce(ctx context.Context, store storage.Store, logger *logr
...
+			if options.SystemContext.VariantChoice != "" {
+				options.Args["TARGETPLATFORM"] = options.Args["TARGETPLATFORM"] + "/" + options.SystemContext.VariantChoice
+			}
```
rhatdan, line 365: options.Args["TARGETPLATFORM"] = options.Args["TARGETPLATFORM"] + "/" +  options.Args["TARGETVARIANT"]

imagebuildah/build.go
```diff
@@ -340,6 +340,32 @@ func buildDockerfilesOnce(ctx context.Context, store storage.Store, logger *logr
...
+		}
+		// variant component from --platform string populates TARGETVARIANT
+		if _, ok := options.Args["TARGETVARIANT"]; !ok {
+			if options.SystemContext.VariantChoice != "" {
+				options.Args["TARGETVARIANT"] = options.SystemContext.VariantChoice
+			}
+		}
+		if _, ok := options.Args["TARGETPLATFORM"]; !ok {
+			options.Args["TARGETPLATFORM"] = options.SystemContext.OSChoice + "/" + options.SystemContext.ArchitectureChoice
+			if options.SystemContext.VariantChoice != "" {
```
rhatdan, line 363: if options.Args["TARGETVARIANT"] != "" {

rhatdan (approved):

imagebuildah/build.go
flouthoc: If users wants he can override `TARGETOS` and `TARGETARGS` so there can be scenario when `options.Args["TARGETOS"] ` and `options.Args["TARGETARCH"]` are not set.

imagebuildah/build.go
flouthoc: Same here if user wants he can override `TARGETVARIANT` since its a `build-arg` so this can still point to wrong value. we must set `TARGETPLATFORM` exactly from `--platform` unless user wants it to.

imagebuildah/build.go
flouthoc: this is same as above since we are using `options.SystemContext.VariantChoice` to populate and not `buildargs.

imagebuildah/build.go
flouthoc: IMO I think in the end they just `build-args` so user should always have ability to override manually any one of them without disturbing the defaults of other `build-args`. So this ensures that `TARGETPLATFORM` will always point to correct default if it was not meant to be overridden.

rhatdan: If user set options.SystemContext.OSChoice then it will be the same in options.Args["TARGETPLATFORM"]  via the code above.  If they did not set it, then you will end with TARGETPLATFORM =="", potentially.

flouthoc: I think the line `if _, ok := options.Args["TARGETOS"]; !ok {` verifies that user did not use `--build-arg TARGETOS=something` so value of `options.Args["TARGETOS"]` will be `options.SystemContext.OSChoice` only if `--build-arg` was not used

```go
	if _, ok := options.Args["TARGETOS"]; !ok {
                        // only set when user did not use --build-arg TARGETOS=something
			options.Args["TARGETOS"] = options.SystemContext.OSChoice
	}
```
so if we do 
`options.Args["TARGETPLATFORM"] = options.Args["TARGETOS"] + "/" + options.Args["TARGETARCH"]` and user does
`--build-arg TARGETOS=android` then `TARGETPLATFORM` becomes `android/<somearch>` which seems wrong.

I have verified this behaviour with buildkit as well here is the output with test `Containerfile` included in this PR
```console
sudo docker buildx build --platform=linux/arm64 --no-cache --progress=plain --build-arg TARGETOS=android -t test .
```
```console
...

#5 [2/2] RUN echo "I'm compiling for linux/arm64 on linux/amd64 and tagging for linux/arm64 and OS android and ARCH arm64"
#5 0.311 I'm compiling for linux/arm64 on linux/amd64 and tagging for linux/arm64 and OS android and ARCH arm64
#5 DONE 0.4s
...
```

also added a test in CI which ensures we don't regress this.

flouthoc: So accepting this change would break test `build-with-inline-platform-and-rely-on-defaultbuiltinargs-check-custom-override` in this PR.

flouthoc: We discussed this and keeping behviour as is since it matches with buildkit but can change if this is found to be bug in buildkit.

imagebuildah/build.go
flouthoc: We discussed this and keeping behviour as is since it matches with buildkit but can change if this is found to be bug in buildkit.

flouthoc:
@rhatdan PTAL updated comments as requested.

edsantiago:
This breaks podman: SEGV in [bud-multiple-something](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5845588578140160/html/bud-podman-fedora-36-root-host.log.html) (also [remote](  https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/5282638624718848/html/bud-remote-fedora-36-root-host.log.html)). Stack trace, so it should be easy to identify and fix.

flouthoc:
@edsantiago Checking.

PR #3900, opened 2022-04-05
tests: set `--jobs=1` for flaky `bud-multiple-platform-no-run`

flouthoc:
We have a recorded bug here https://github.com/containers/buildah/issues/3710
but complexity of this bug fix is looking high as of now with fixes
going into multiple components.

See dicsussion here: https://github.com/containers/buildah/issues/3710

Reducing jobs for this test will ensure that we process things in serial
order so instead of skipping this test we can still run it in a serial
manner and with no flakes

**Note: This does not fixes the underlying issue but just removes flakes from CI**

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -3625,7 +3625,11 @@ _EOF
 
 @test "bud-multiple-platform-no-run" {
   outputlist=localhost/testlist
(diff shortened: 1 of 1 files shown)


flouthoc:
@rhatdan @cevich @edsantiago @nalind PTAL. This is not a fix but should keep CI clam till original issue is closed/fixed upstream.

flouthoc:
I'll undraft this this if we decide to hide the problem in CI for now.

TomSweeneyRedHat:
Change LGTM
I'll let others decide if this is warranted or not.  I'd lean towards no.

cevich:
LGTM. I'm okay with this as a temporary workaround.

PR #3901, opened 2022-04-06
buildah build should accept at most one arg

Luap99:
#### What type of PR is this?

/kind bug

Right now buildah just ignores it when you set more than one argument for
build. Only the first argument is used as context dir. This looks wrong
to me and can lead to users errors.

Fixes test failure in https://github.com/containers/podman/pull/13687

```release-note
buildah build only accepts at most one argument, it errors now when multiple arguments are given instead of ignoring them.
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -171,20 +171,6 @@ _EOF
   expect_output --substring $targetarch
 }
 
-@test "bud-flags-order-verification" {
-  run_buildah 125 build /tmp/tmpdockerfile/ -t blabla
-  check_options_flag_err "-t"
-
-  run_buildah 125 build /tmp/tmpdockerfile/ -q -t blabla
-  check_options_flag_err "-q"
-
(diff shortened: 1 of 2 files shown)


flouthoc (approved):
LGTM

tests/bud.bats
```diff
@@ -171,20 +171,6 @@ _EOF
...
-  run_buildah 125 build /tmp/tmpdockerfile/ -q -t blabla
-  check_options_flag_err "-q"
-
-  run_buildah 125 build /tmp/tmpdockerfile/ --force-rm
-  check_options_flag_err "--force-rm"
-
-  run_buildah 125 build /tmp/tmpdockerfile/ --userns=cnt1
-  check_options_flag_err "--userns=cnt1"
-}
-
```
TomSweeneyRedHat, line 187: I'm not sure why you removed this test....

PR #3906, opened 2022-04-09
Initialize network backend before first pull

hshiina:

#### What type of PR is this?

/kind bug

After clean install, it is necessary to decide the network backend before any image is pulled so that `netavark` is chosen correctly. Without this change, if `buildah pull` is executed at first, the network backend is not determined and an image is pulled. This results in choosing `cni` at a next command because an image already exists while `netavark` is chosen if `buildah from` or `buildah bud` is called at first.

#### How to verify it

Run the added test.

None

```release-note
None
```

### DIFF
--- a/pull.go
+++ b/pull.go
@@ -75,6 +75,14 @@ func Pull(ctx context.Context, imageName string, options PullOptions) (imageID s
 		return "", err
 	}
 
+	// Note: It is important to do this before we pull any images/create containers.
+	// The default backend detection logic needs an empty store to correctly detect
+	// that we can use netavark, if the store was not empty it will use CNI to not break existing installs.
+	_, err = getNetworkInterface(options.Store, "", "")
+	if err != nil {
+		return "", err
+	}
+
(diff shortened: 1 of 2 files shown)


pull.go
```diff
@@ -75,6 +75,14 @@ func Pull(ctx context.Context, imageName string, options PullOptions) (imageID s
 		return "", err
 	}
 
+	// Note: It is important to do this before we pull any images/create containers.
+	// The default backend detection logic needs an empty store to correctly detect
+	// that we can use netavark, if the store was not empty it will use CNI to not break existing installs.
+	_, err = getNetworkInterface(options.Store, "", "")
```
rhatdan, line 81: Nit: could you do this on one line.
if _, err := getNetworkInterface ...

rhatdan:
LGTM
@Luap99 PTAL

Luap99 (commented):
LGTM

flouthoc (approved):
LGTM

PR #3916, opened 2022-04-18
manifest-create: allow creating manifest list from image in `local-storage`

flouthoc:
`manifest create` must perform a final lookup on local storage while
user adds image while creating a manifest list because there can be
scenarios when image is not pushed to remote reistry and it only exists
on local storage.

A change similar to this was added here: https://github.com/containers/buildah/pull/3877
Also see: https://github.com/containers/buildah/blob/main/cmd/buildah/manifest.go#L328

Closes: https://github.com/containers/buildah/issues/3915

### DIFF
--- a/cmd/buildah/manifest.go
+++ b/cmd/buildah/manifest.go
@@ -257,7 +257,13 @@ func manifestCreateCmd(c *cobra.Command, args []string, opts manifestCreateOpts)
 				}
 			}
 		}
-		if _, err = list.Add(getContext(), systemContext, ref, opts.all); err != nil {
+		refLocal, _, err := util.FindImage(store, "", systemContext, imageSpec)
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM
@nalind PTAL

nalind:
Why do we hit the registry first?  I would have expected the local image to matter more.

flouthoc:
@nalind Not sure but I think it was always like this just made sure that we hit local-storage at least once, I can flip the logic though to hit `local` first.

flouthoc:
@nalind Flipped the logic.

nalind:
LGTM

flouthoc:

Issue #3914, opened 2022-04-14
buildah should set hostname in /etc/hostname


Luap99:
**Description**

Buildah does not set the hostname in /etc/hostname.
While working on the /etc/hosts changes I noticed a difference between containers created by podman and buildah. Buildah does not manage the /etc/hostname file.

See the code for podman here: https://github.com/containers/podman/blob/53b984f20fef3a66269029f10c6cd16d7b80fd4e/libpod/container_internal_linux.go#L1988-L1996 

**Steps to reproduce the issue:**
1. 
```
$ cat Containerfile 
FROM alpine
RUN hostname && cat /etc/hostname
```
2. `buildah build .`

**Describe the results you received:**
```
STEP 3/3: RUN hostname && cat /etc/hostname
96e5f672af17
localhost
```

**Describe the results you expected:**
```
STEP 3/3: RUN hostname && cat /etc/hostname
96e5f672af17
96e5f672af17
```

**Output of `buildah version`:**
latest main branch

flouthoc:
@Luap99 Do we have any prior issue associated to this, although legacy docker honors behavior as suggested in this issue but newer buildkit keeps `hostname` as localhost by default. 

I tried this with buildkit and here is the output it keeps `/etc/hostname` as `localhost`. We could do this change but it breaks parity with buildkit.

```console
#1 [internal] load .dockerignore
#1 transferring context: 2B done
#1 ...
 
#2 [internal] load build definition from Dockerfile
#2 transferring dockerfile: 84B done
#2 DONE 0.8s
 
#1 [internal] load .dockerignore
#1 DONE 1.0s
 
#3 [internal] load metadata for docker.io/library/alpine:latest
#3 DONE 3.2s
 
#4 [1/3] FROM docker.io/library/alpine@sha256:4edbd2beb5f78b1014028f4fbb99f3237d9561100b6881aabbf5acce2c4f9454
#4 resolve docker.io/library/alpine@sha256:4edbd2beb5f78b1014028f4fbb99f3237d9561100b6881aabbf5acce2c4f9454
#4 resolve docker.io/library/alpine@sha256:4edbd2beb5f78b1014028f4fbb99f3237d9561100b6881aabbf5acce2c4f9454 0.3s done
#4 DONE 0.3s
 
#4 [1/3] FROM docker.io/library/alpine@sha256:4edbd2beb5f78b1014028f4fbb99f3237d9561100b6881aabbf5acce2c4f9454
#4 DONE 0.7s
 
[...]

PR #3920, opened 2022-04-21
run: set actual `hostname` in `/etc/hostname` to match docker parity

flouthoc:
Following commit ensure the buildah run populates /etc/hostname with
actual hostname instead of `localhost`. This matches parity with docker.

See discussion here: https://github.com/containers/buildah/issues/3914
Closes: https://github.com/containers/buildah/issues/3914


flouthoc:
@Luap99 @containers/buildah-maintainers PTAL

run_linux.go
```diff
@@ -685,6 +697,38 @@ func (b *Builder) generateHosts(rdir, hostname string, addHosts []string, chownO
 	return cfile, nil
 }
 
+// generateHostname creates a containers /etc/hostname file
+func (b *Builder) generateHostname(rdir, hostname string, chownOpts *idtools.IDPair) (string, error) {
+	hostnamePath := "/etc/hostname"
+	stat, err := os.Stat(hostnamePath)
+	if err != nil {
+		return "", err
+	}
```
Luap99, line 706: This will fail if /etc/hostname does not exists for some reason. We should no check this. Just use uid/gid = 0 as default. I doubt that this will ever be owned by another uid.

flouthoc: fixed

flouthoc: @Luap99 PTAL again.

run_linux.go
```diff
@@ -232,6 +232,18 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 		}
 	}
 
+	// generate /etc/hostname if user intentionally did not meant to override
```
TomSweeneyRedHat, line 235: nit of a nit
```suggestion
	// generate /etc/hostname if the user intentionally did not mean to override
```

Luap99 (commented):
LGTM

PR #3926, opened 2022-04-25
Ed's periodic test cleanup

edsantiago:
Primarily and most importantly, check error messages. You will
see this material again.
 - subnote, find and fix actually broken tests

Second, fix an invalid use of $TESTSDIR ('S' meaning 'Source')

And a few other fixes

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -1097,6 +1097,7 @@ function _test_http() {
   _prefetch alpine
   target=alpine-image
   run_buildah 125 build --signature-policy ${TESTSDIR}/policy.json -t ${target} -f ${TESTSDIR}/bud/symlink/Dockerfile.symlink-points-to-itself ${TESTSDIR}/bud/symlink
+  assert "$output" =~ "error building .* open /test-log/test: too many levels of symbolic links"
 }
 
 @test "bud multi-stage with symlink to absolute path" {
@@ -1141,7 +1142,8 @@ function _test_http() {
   _prefetch alpine
   target=alpine-image
   run_buildah 2 bud --signature-policy ${TESTSDIR}/policy.json -t ${target} -f Dockerfile.entrypoint-empty-run ${TESTSDIR}/bud/run-scenarios
-  expect_output --substring "error building at STEP"
+  expect_output --substring " -c requires an argument"
+  expect_output --substring "error building at STEP.*: exit status 2"
 }
 
 @test "bud with CMD and RUN" {
@@ -1156,7 +1158,8 @@ function _test_http() {
   _prefetch alpine
   target=alpine-image
   run_buildah 2 bud --signature-policy ${TESTSDIR}/policy.json -t ${target} -f Dockerfile.cmd-empty-run ${TESTSDIR}/bud/run-scenarios
-  expect_output --substring "error building at STEP"
+  expect_output --substring " -c requires an argument"
+  expect_output --substring "error building at STEP.*: exit status 2"
 }
 
 @test "bud with ENTRYPOINT, CMD and RUN" {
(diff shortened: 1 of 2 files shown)


tests/bud.bats
```diff
@@ -2422,15 +2431,15 @@ _EOF
...
+  FILE=${TESTDIR}/budcopy/Dockerfile.url
   /bin/cat <<EOM >$FILE
 FROM alpine:latest
 COPY https://getfedora.org/index.html .
 EOM
 
-  run_buildah 125 build --signature-policy ${TESTSDIR}/policy.json -t ${target} -f ${TESTSDIR}/bud/copy/Dockerfile.url ${TESTSDIR}/bud/copy
-  rm -r ${TESTSDIR}/bud/copy
+  run_buildah 125 build --signature-policy ${TESTSDIR}/policy.json -t foo -f ${TESTDIR}/budcopy/Dockerfile.url
+  expect_output --substring "error building .* source can.t be a URL for COPY"
```
edsantiago, line 2442: @containers/buildah-maintainers PTAL at this. This test was completely broken, and I want to use it as an emphatic indicator of why we need to check error messages.

The bug was in `-t ${target}`. `$target` was never defined, so this became `-t -f ...Dockerfile etc` which led to this failure:
```
tag -f: invalid reference format
```
So, yeah, the exit-status check was correct but for completely wrong reason.

This is not a shaming attempt. I promise I have not run `tig blame`, and beg all of you not to do so either; and if this is your code, don't speak up. This is a bug that can happen to any of us, and has. The point is, pleeeeeease don't just rely on exit-status checks. Please look for a distinct error message. Thank you.

tests/bud.bats
```diff
@@ -3590,15 +3625,17 @@ _EOF
...
+  assert "$d2" =~ "^sha256:[0-9a-f]{64}\$"
+  assert "$d1" != "$d2" "digest(arm) should != digest(386)"
 }
 
 @test "bud-multiple-platform-no-partial-manifest-list" {
   outputlist=localhost/testlist
   run_buildah 1 bud --signature-policy ${TESTSDIR}/policy.json --platform=linux/arm,linux/amd64 --manifest $outputlist -f ${TESTSDIR}/bud/multiarch/Dockerfile.fail ${TESTSDIR}/bud/multiarch
+  expect_output --substring "error building at STEP \"RUN test .arch. = x86_64"
   run_buildah 125 manifest inspect $outputlist
+  expect_output --substring "reading image .* pinging container registry"
```
edsantiago, line 3638: This is another error message I'm not 100% sure about. `pinging blah blah` is the error we get, and is easily reproduced on my laptop, I was just, yknow, expecting something different.

tests/bud.bats
```diff
@@ -1217,6 +1221,7 @@ function _test_http() {
   target=url-image
   url=https://raw.githubusercontent.com/containers/buildah/main/tests/bud/from-scratch/Dockerfile.bogus
   run_buildah 125 build --signature-policy ${TESTSDIR}/policy.json -t ${target} ${url}
+  expect_output "no FROM statement found"
```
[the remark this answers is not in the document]
TomSweeneyRedHat: That's correct, but not obvious.  The issue is the URL set on line 1222 is invalid; that file does not exist there.  I'm open to a better test name here, and/or changing the value sent into the URL field to be more obvious.

tests/bud.bats
TomSweeneyRedHat, line 2442: Should this be instead?
```suggestion
  expect_output --substring "error building .* source can't be a URL for COPY"
```

edsantiago: (apostrophe instead of single wild-card character): I prefer not playing the shell quoting game, because that way lies madness.

rhatdan: I agree.

tests/bud.bats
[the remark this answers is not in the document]
flouthoc: i think test could be renamed to `Invalid dockerfile from URL`, cause URL seems to be valid its just we get a 404 instead of a valid dockerfile, so buildah tries to process `404` which is invalid dockerfile in the end.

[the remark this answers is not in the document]
edsantiago: I feel rather strongly that buildah should be more clear on the error: "Could not fetch <URL>: 404" would be more helpful to an end user than "oh, I think I fetched your URL but couldn't find a FROM line". And I agree that "Invalid URL" is a misnomer. But neither can be solved in this PR.

PR #3929, opened 2022-04-26
Tests: rename $TESTSDIR and $TESTDIR

edsantiago:
They cause way too much confusion. Rename $TESTSDIR (plural) to $TEST_SOURCES
and $TESTDIR (singular) to $TEST_SCRATCH_DIR.

In the process, refactor two commonly-used expressions ($WITH_POLICY_JSON and $BUDFILES).

Split into four separate commits, for ease of review.

### DIFF
--- a/tests/add.bats
+++ b/tests/add.bats
@@ -17,7 +17,7 @@ load helpers
   createrandom ${TESTDIR}/randomfile
   createrandom ${TESTDIR}/other-randomfile
 
-  run_buildah from --signature-policy ${TESTSDIR}/policy.json scratch
+  run_buildah from $WITH_POLICY_JSON scratch
   cid=$output
   run_buildah mount $cid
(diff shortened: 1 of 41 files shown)


edsantiago:
Suggestions welcome if you have better ideas for the new var names.

Luap99:
Should `TEST_SCRATCH_DIR` just be named `TMPDIR` or `BUILDAH_TMPDIR`?

edsantiago:
`TMPDIR` is impossible because that's a UNIX convention (path to a directory to be used for all temp files, usually pointing to `/tmp` or `/var/tmp`). I like `BUILDAH_TMPDIR` because it's consistent with the convention in podman tests, but I also think `TEST_SCRATCH_DIR` is slightly more descriptive, so I'll leave the decision up to the team.

PR #3931, opened 2022-04-26
Cirrus: Update CI VMs to F36

cevich:
#### What type of PR is this?

/kind other

Bump Fedora VM to version 36 and prior-Fedora to 35

#### How to verify it

All CI will pass

None

Ref: https://github.com/containers/automation_images/pull/115

None

[NO NEW TESTS NEEDED]

### DIFF
--- a/pkg/formats/templates.go
+++ b/pkg/formats/templates.go
@@ -20,6 +20,10 @@ var basicFunctions = template.FuncMap{
 	},
 	"split":    strings.Split,
 	"join":     strings.Join,
+	// strings.Title is deprecated since go 1.18
+	// However for our use case it is still fine. The recommended replacement
+	// is adding about 400kb binary size so lets keep using this for now.
+	//nolint:staticcheck
 	"title":    strings.Title,
 	"lower":    strings.ToLower,
 	"upper":    strings.ToUpper,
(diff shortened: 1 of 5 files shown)


pkg/formats/templates.go
```diff
@@ -20,6 +20,10 @@ var basicFunctions = template.FuncMap{
 	},
 	"split":    strings.Split,
 	"join":     strings.Join,
+	// strings.Title is deprecated since go 1.18
+	// However for our use case it is still fine. The recommended replacement
+	// is adding about 400kb binary size so lets keep using this for now.
+	//nolint:staticcheck
```
TomSweeneyRedHat, line 26: Should we add a TODO or similar tag so we can more easily find this?

TomSweeneyRedHat:
Tests aren't hip

pkg/formats/templates.go
Luap99: I don't think it is a problem. Since it is in the stdlib it will stay there until golang decides to actively break code.

Luap99: better question why do we have this package? I thought we have this in c/common.
cc @jwhonce

PR #3934, opened 2022-04-27
copier.Put(): write to read-only directories

nalind:
#### What type of PR is this?

/kind bug

Try to improve our ability to write to directories that aren't writable.  If we encounter an EPERM error while attempting to create an item, attempt to temporarily make writeable the directory that we're writing the item to, and restore its permissions on our way out.

The error usually isn't seen when run as UID 0, whether in a user namespace or not, which is usually how we're called, but running the unit tests as an unprivileged user will verify it.

#### How to verify it

Added to unit tests!

Fixes https://github.com/containers/podman/issues/13992.

```release-note
None
```


rhatdan:
Very clever hack.
LGTM

rhatdan:
@mheon Would like to get this into 4.1.

flouthoc (approved):
LGTM
/approve
/lgtm

PR #3935, opened 2022-04-27
copier.unwrapError(): update for Go 1.16

nalind:
#### What type of PR is this?

/kind cleanup

Since we started calling into the standard library's `io/fs` package directly, we effectively made Go 1.16 our minimum Go version, so we don't need to keep the workaround for compiling with Go 1.12.

~Unexport the local Lgetxattrs/Lsetxattrs, which we're not using anywhere else.~

#### How to verify it

The unit tests should now exercise the function's use in handling ERANGE errors from the kernel.

None

```release-note
None
```


nalind:
Oh, so _that's_ why those were exported.

copier/xattrs.go
```diff
@@ -16,7 +16,9 @@ const (
 )
 
 var (
-	relevantAttributes = []string{"security.capability", "security.ima", "user.*"} // the attributes that we preserve - we discard others
+	relevantAttributes    = []string{"security.capability", "security.ima", "user.*"} // the attributes that we preserve - we discard others
+	initialXattrListSize  = 64 * 1024
```
flouthoc, line 20: Should change regarding extended attributes list size belong in a different commit. Although change is trivial so its a non-blocker.

flouthoc:
LGTM just one nit above regarding splitting the commit.

PR #3938, opened 2022-04-28
Fix docs build on FreeBSD

dfr:
The sed implementation has a strict interpretation of posix 'basic'
regular expressions. It would be better to re-implement this using
'extended' regular expressions but for now, just use GNU sed.

#### What type of PR is this?

> /kind cleanup

I'm porting buildah to FreeBSD

#### How to verify it

Run:
```
gmake -C docs
```
on a FreeBSD host - without this change all the manpages are zero length

None

None


flouthoc (approved):
LGTM

rhatdan:
/approve
LGTM

Issue #3939, opened 2022-04-28
Build args does not resolves to stage when used within `FROM` when invoked with `--layers=false` or without `--layers`


flouthoc:
It seems buildah ignores checking stage alias names when invoked without `--layers`.

### Reproducer
```Dockerfile
ARG app_type

FROM alpine as x
RUN echo hello

FROM ${app_type} as final
RUN echo world
```

```console
buildah build -t test --build-arg app_type=x
```

Above build command prompts users to select image `x` from registry instead of resolving to stage names as `x`.

PR #3947, opened 2022-05-02
imagebuildah, executor: process `ARG` variables while populating baseMap

flouthoc:
While processing `FROM <token> as final` executor populates baseMap as
it is without resolving or processing for any ARG values. Following
commit ensures that we process resolve any ARG variables with ARG values
while populating baseMap so it can be used later to check if stage is
reused.

Fixes: https://github.com/containers/buildah/issues/3939

### DIFF
--- a/imagebuildah/util.go
+++ b/imagebuildah/util.go
@@ -10,3 +10,13 @@ import (
 func InitReexec() bool {
 	return buildah.InitReexec()
 }
+
+// argsMapToSlice returns the contents of a map[string]string as a slice of keys
+// and values joined with "=".
+func argsMapToSlice(m map[string]string) []string {
+	s := make([]string, 0, len(m))
+	for k, v := range m {
+		s = append(s, k+"="+v)
+	}
+	return s
+}
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -3180,6 +3180,12 @@ _EOF
   assert "$output" !~ '--build-arg NEWSECRET=<VALUE>'
 }
 
+@test "bud with arg in from statement" {
+  _prefetch alpine
+  run_buildah build -t testbud $WITH_POLICY_JSON --build-arg app_type=x --build-arg another_app_type=m --file $BUDFILES/with-arg/Dockerfilefromarg .
+  expect_output --substring 'world'
+}
+
 @test "bud with --runtime and --runtime-flag" {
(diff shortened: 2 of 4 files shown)


tests/bud.bats
```diff
@@ -3180,6 +3180,12 @@ _EOF
   assert "$output" !~ '--build-arg NEWSECRET=<VALUE>'
 }
 
+@test "bud with arg in from statement" {
+  _prefetch alpine
+  run_buildah build -t testbud $WITH_POLICY_JSON --build-arg app_type=x --file $BUDFILES/with-arg/Dockerfilefromarg .
+  expect_output --substring 'world'
+}
+
```
TomSweeneyRedHat, line 3188: Just in case, could you add a test with multiple build-args?

flouthoc: Sure. I'll amend this.

flouthoc:

imagebuildah/util.go
```diff
@@ -10,3 +10,13 @@ import (
 func InitReexec() bool {
 	return buildah.InitReexec()
 }
+
+// ArgsMapToSlice returns the contents of a map[string]string as a slice of keys
+// and values joined with "=".
+func ArgsMapToSlice(m map[string]string) []string {
```
nalind, line 16: This doesn't need to be exported.

tests/bud/with-arg/Dockerfilefromarg
```diff
@@ -0,0 +1,11 @@
...
+ARG another_app_type
+
+FROM alpine as x
+RUN echo hello
+
+FROM ${app_type} as m
+RUN echo world
+
+FROM ${another_app_type} as final
+RUN echo hello
```
nalind, line 11: Please also verify that an argument's value is expanded correctly if its definition supplies a default value, and that value isn't specified at the build command line.

nalind (changes requested):
One nit, and I'd like another case tested.  Otherwise LGTM.

tests/bud/with-arg/Dockerfilefromarg
flouthoc: Added in the same containerfile above

flouthoc:
@nalind fixed added the requested test in the same containerfile which checks for default value. @containers/buildah-maintainers PTAL

PR #3953, opened 2022-05-03
Handle OS version and features flags

nalind:
#### What type of PR is this?

/kind feature

Teach `buildah build` and `buildah config` about the OS version and features fields.  We don't tend to use them on Linux, but they're in the OCI and Docker config structures, so we need to be able to preserve and manipulate their values, much as we do for variant information.

Add a `--env` option to `buildah build` that functions similarly to the `buildah config --env` option, to complement `buildah build`'s `--unsetenv` option.

Document that `buildah config`'s `--env` function fetches the current value for a variable when the name is supplied, but no `=` or value follows it.

Allow the value of the `io.buildah.version` label to be manually specified on the `buildah build` command line if someone really wants to override the value that we would otherwise set by default.

Any `buildah.BuilderOptions.DefaultEnv` values have been ignored since the library was changed to read them from the server-side configuration, for the sake of podman remote builds, as part of #2131.  Update its godoc to note that and stop setting it.

Tweak the usage text for various CLI options to be more consistent about case.

#### How to verify it

New integration tests for `--os-version`, `--os-features`, and `--env`!
The same integration tests for `--unsetenv`!

None

```release-note
The `buildah build` command now includes a `--env` option to complement its `--unsetenv` option.
The default value of the `io.buildah.version` label can now be overridden at the `buildah build` command line using the `--label` option.
The `buildah build` and `buildah config` commands now accept `--os-version` and `--os-feature` flags for setting the corresponding fields in an image's configuration. 
```

### DIFF
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -39,6 +39,8 @@ type configResults struct {
 	label                  []string
 	onbuild                []string
 	os                     string
+	osfeature              []string
+	osversion              string
 	ports                  []string
 	shell                  string
 	stopSignal             string
@@ -88,6 +90,8 @@ func init() {
 	flags.StringArrayVarP(&opts.label, "label", "l", []string{}, "add image configuration `label` e.g. label=value")
 	flags.StringSliceVar(&opts.onbuild, "onbuild", []string{}, "add onbuild command to be run on images based on this image. Only supported on 'docker' formatted images")
 	flags.StringVar(&opts.os, "os", "", "set `operating system` of the target image")
+	flags.StringArrayVar(&opts.osfeature, "os-feature", []string{}, "set required OS `feature` for the target image")
+	flags.StringVar(&opts.osversion, "os-version", "", "set required OS `version` for the target image")
 	flags.StringSliceVarP(&opts.ports, "port", "p", []string{}, "add `port` to expose when running containers based on image (default [])")
 	flags.StringVar(&opts.shell, "shell", "", "add `shell` to run in containers")
 	flags.StringVar(&opts.stopSignal, "stop-signal", "", "set `stop signal` for containers based on image")
@@ -176,6 +180,21 @@ func updateConfig(builder *buildah.Builder, c *cobra.Command, iopts configResult
 	}
 	if c.Flag("os").Changed {
 		builder.SetOS(iopts.os)
+	}
+	if c.Flag("os-feature").Changed {
+		for _, osFeatureSpec := range iopts.osfeature {
+			switch {
+			case osFeatureSpec == "-":
+				builder.ClearOSFeatures()
+			case strings.HasSuffix(osFeatureSpec, "-"):
+				builder.UnsetOSFeature(strings.TrimSuffix(osFeatureSpec, "-"))
+			default:
+				builder.SetOSFeature(osFeatureSpec)
+			}
+		}
+	}
+	if c.Flag("os-version").Changed {
+		builder.SetOSVersion(iopts.osversion)
 	}
 	if c.Flag("user").Changed {
 		builder.SetUser(iopts.user)
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -226,7 +226,6 @@ func updateConfig(builder *buildah.Builder, c *cobra.Command, iopts configResult
 			}
 			env[1] = os.Expand(env[1], getenv)
 			builder.SetEnv(env[0], env[1])
-
 		case env[0] == "-":
 			builder.ClearEnv()
 		case strings.HasSuffix(env[0], "-"):
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -421,6 +421,24 @@ By default, Buildah manages _/etc/hosts_, adding the container's own IP address.
 
 Set the OS of the image to be built, and that of the base image to be pulled, if the build uses one, instead of using the current operating system of the host.
 
+**--os-feature** *feature*
+
+Set the name of a required operating system *feature* for the image which will
+be built.  By default, if the image is not based on *scratch*, the base image's
+required OS feature list is kept, if the base image specified any.  This option
+is typically only meaningful when the image's OS is Windows.
+
+If *feature* has a trailing `-`, then the *feature* is removed from the set of
+required features which will be listed in the image.
+
+**--os-version** *version*
+
+Set the exact required operating system *version* for the image which will be
+built.  By default, if the image is not based on *scratch*, the base image's
+required OS version is kept, if the base image specified one.  This option is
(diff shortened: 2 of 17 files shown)


rhatdan:
LGTM
Do you want to get this into the release by tomorrow?

nalind:
I wouldn't block a release for it.  We were already preserving values from base images, so os-version was probably fine, and the common package's manifest list logic was already pulling the info from images when it went to add them to manifest lists.

rhatdan:
Well if you get tests passing by end of day, we can take it in.

cmd/buildah/config.go
```diff
@@ -39,6 +39,8 @@ type configResults struct {
 	label                  []string
 	onbuild                []string
 	os                     string
+	osversion              string
+	osfeature              []string
```
TomSweeneyRedHat, line 43: too keep my ocd happy, could you flip the order of these, feature at line 42, version at 43 please?

docs/buildah-build.1.md
```diff
@@ -241,6 +242,14 @@ Set custom DNS options
 
 Set custom DNS search domains
 
+**--env** *env[=value]*
+
+Add a value (e.g. env=*value*) to the built image.  Can be used multiple times.
+If `=` and `*value* are not specified, but *env* is set in the current
```
TomSweeneyRedHat, line 248:
```suggestion
If `=` and `*value*` are not specified, but *env* is set in the current
```

docs/buildah-build.1.md
```diff
@@ -327,9 +336,10 @@ Run up to N concurrent stages in parallel.  If the number of jobs is greater tha
 stdin will be read from /dev/null.  If 0 is specified, then there is
 no limit on the number of jobs that run in parallel.
 
-**--label** *label*
+**--label** *label[=value]*
 
 Add an image *label* (e.g. label=*value*) to the image metadata. Can be used multiple times.
+If *label* is named, but no `=` or `value` is provided, then the *label* is removed from the image.
```
TomSweeneyRedHat, line 342:
```suggestion
If *label* is named, but neither `=` or `value` is provided, then the *label* is removed from the image.
```

docs/buildah-build.1.md
```diff
@@ -411,6 +421,24 @@ By default, Buildah manages _/etc/hosts_, adding the container's own IP address.
...
+required features which will be listed in the image.
+
+**--os-version** *version*
+
+Set the exact required operating system *version* for the image which will be
+built.  By default, if the image is not based on *scratch*, the base image's
+required OS version is kept, if the base image specified one.  This option is
+typically only useful when the image's OS is Windows, and is typically set in
+Windows base images, so using this option is usually unnecessary.
+
```
TomSweeneyRedHat, line 441: It would be nice to have some examples in the example section for all of the above.

docs/buildah-config.1.md
```diff
@@ -77,10 +77,12 @@ ignore the `cmd` value of the container image.  However if you use the array
 form, then the cmd will be appended onto the end of the entrypoint cmd and be
 executed together.
 
-**--env**, **-e** *env=value*
+**--env**, **-e** *env[=value]*
 
 Add a value (e.g. env=*value*) to the environment for containers based on any
 images which will be built using the specified container. Can be used multiple times.
+If *env* is named but no `=` or `value` is specified, then the value will be
```
TomSweeneyRedHat, line 84:
```suggestion
If *env* is named but neither `=` or `value` is specified, then the value will be
```

docs/buildah-config.1.md
```diff
@@ -160,6 +162,26 @@ Set the target *operating system* for any images which will be built using
...
+
+**--os-version** *version*
+
+Set the exact required operating system *version* for any images which will be
+built using the specified container.  By default, if the container was based on
+an image, the base image's required OS version is kept, if it specified one.
+This option is typically only useful when the image's OS is Windows, and is
+typically set in Windows base images, so using this option is usually
+unnecessary.
+
```
TomSweeneyRedHat, line 184: ditto examples

pkg/cli/common.go
```diff
@@ -92,6 +92,9 @@ type BudResults struct {
 	LogRusage           bool
 	RusageLogFile       string
 	UnsetEnvs           []string
+	OSVersion           string
+	OSFeatures          []string
+	Envs                []string
```
TomSweeneyRedHat, line 97: There seems to be something off with tab/spaces with this in comparison to the `Jobs` and above line.  Can you at least alphabetize the three that you've added in the upper part of this struct please?  Well as much a possible, I see other rogue entries there.

pkg/cli/common.go
```diff
@@ -272,6 +278,8 @@ func GetBudFlagsCompletions() commonComp.FlagCompletions {
 	flagCompletion["logfile"] = commonComp.AutocompleteDefault
 	flagCompletion["manifest"] = commonComp.AutocompleteDefault
 	flagCompletion["os"] = commonComp.AutocompleteNone
+	flagCompletion["os-version"] = commonComp.AutocompleteNone
+	flagCompletion["os-features"] = commonComp.AutocompleteNone
```
TomSweeneyRedHat, line 282: ordering

cmd/buildah/config.go
nalind: Changing it.

docs/buildah-build.1.md
nalind: Nice catch, fixing.

docs/buildah-build.1.md
nalind: Going to neither/nor for the lot of these.

docs/buildah-build.1.md
nalind: Adding examples to buildah-config(1) and buildah-build(1).

pkg/cli/common.go
nalind: Changed.

docs/buildah-build.1.md
```diff
@@ -327,9 +336,10 @@ Run up to N concurrent stages in parallel.  If the number of jobs is greater tha
 stdin will be read from /dev/null.  If 0 is specified, then there is
 no limit on the number of jobs that run in parallel.
 
-**--label** *label*
+**--label** *label[=value]*
 
 Add an image *label* (e.g. label=*value*) to the image metadata. Can be used multiple times.
+If *label* is named, but neither `=` nor a `value` is provided, then the *label* is removed from the image.
```
flouthoc, line 342: @nalind not sure looks like this breaks exisiting podman tests. `--label xyz` will not get applied and some podman tests are directly using that.  I see some failures here: https://github.com/containers/podman/pull/14118

nalind: Opened #3963.

PR #3955, opened 2022-05-03
buildah: deep copy `options.Args` before performing concurrent `build/stage`

flouthoc:
Prevent fatal concurrent read/write over options.Args by concurrent
multi-arch builds and concurrent stages

[NO TESTS NEEDED]
[NO NEW TESTS NEEDED]

Fixes: report https://github.com/containers/buildah/pull/3899#issuecomment-1116293655

### DIFF
--- a/new.go
+++ b/new.go
@@ -317,7 +317,7 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
 		Capabilities:     copyStringSlice(options.Capabilities),
 		CommonBuildOpts:  options.CommonBuildOpts,
 		TopLayer:         topLayer,
-		Args:             options.Args,
+		Args:             copyStringStringMap(options.Args),
 		Format:           options.Format,
 		TempVolumes:      map[string]bool{},
 		Devices:          options.Devices,
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -332,7 +332,15 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 	return id, ref, nil
 }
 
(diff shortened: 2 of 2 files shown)


flouthoc:
@edsantiago @nalind PTAL

flouthoc:
@edsantiago Thanks for spotting this I don't know how we missed this :\  my bad.

new.go
```diff
@@ -317,7 +323,7 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
 		Capabilities:     copyStringSlice(options.Capabilities),
 		CommonBuildOpts:  options.CommonBuildOpts,
 		TopLayer:         topLayer,
-		Args:             options.Args,
+		Args:             argsCopy,
```
nalind, line 326: Use `copyStringStringMap(options.Args)` here.

edsantiago:
Crash has not reproduced. Haven't tried patching with @nalind's suggestion.

Thank you @flouthoc for your fast work.

flouthoc:
Thanks to @nalind cause i was about miss another spot for concurrent read and write between stages so he spotted it before this PR.

new.go
flouthoc: fixed.

flouthoc:
@TomSweeneyRedHat I think we would need this before the new cut.

nalind:
Agreed, if #3899 is in the branch, this needs to be as well.

edsantiago:
To probably nobody's surprise, the crash still hasn't reproduced in my testing. LGTM. Thanks everyone.

PR #3956, opened 2022-05-04
imagebuildah,buildDockerfilesOnce: move deepcopy of args before we spawn goroutine

flouthoc:
To prevent concurrent read/write between maps perform deep copy of args
before we spawn a goroutine. So instead of copying inside the goroutine PR intends
to create copy of `Args` for each goroutine in the loop before calling them.

[NO NEW TESTS NEEDED]
[NO TESTS NEEDED]


flouthoc:

giuseppe (approved):
LGTM

PR #3963, opened 2022-05-04
Make `buildah build --label foo` create an empty "foo" label again

nalind:
#### What type of PR is this?

/kind bug

Restore podman's CLI compatibility with `docker build` by making its `--label` and `--annotate` options set empty labels and annotations when they're given a name but no `=` or value.

#### How to verify it

Updated integration test!

Test failures in https://github.com/containers/podman/pull/14118, h/t @flouthoc.

```release-note
None
```


flouthoc (approved):
LGTM
/lgtm
/hold

PR #3969, opened 2022-05-06
run,rootless: `enable_ipv6` when using `slirp4netns` for rootless networking

flouthoc:
When user is invoking rootless buildah and networking defaults to
slirp4netns always use `enable_ipv6` so users can resolve ipv6 addresses
as well.

Closes: https://github.com/containers/buildah/issues/3968


flouthoc:
@rhatdan @Luap99 PTAL. This should enable `enable-ipv6` by default. I'll also open a followup to make `slirp` opts configurable from UX but don't know if its urgent.

Luap99 (commented):
We should not duplicate so much code between podman and buildah, also this forces ipv6 which is not good.

This should support all options from contianers.conf at least and I think this must be moved to c/common so that both buildah and podman can share the code.

flouthoc:
@Luap99 We can make this configurable from UX and we could also unify `slirp4netns` in `libnetwork` but `unification` looks like a bigger change for this feature, do you think we should do it now or later if more parity features gets requested other than only `enable_v6`? I'm fine with anything whatever maintainers agree with.

Luap99:
I leave this for other to decide but I fixed many bugs twice in buildah and podman because the code is duplicated for no real reason. We have to unify the code for easier maintenance. This will also reduce binary size when more code is shared.

flouthoc:
@Luap99 Makes sense I agree with you, i think we can get this in `enable_v6` in just to close this issue but instead open a bigger JIRA spike to `unify` `slirp4netns` networking in `libnetwork` irrespective of this feature WDYT ? But again I'm fine with whatever others agree with.  @rhatdan @nalind PTAL

run_linux.go
```diff
@@ -1061,6 +1062,8 @@ func runCollectOutput(logger *logrus.Logger, fds, closeBeforeReadingFds []int) s
 }
 
 func setupRootlessNetwork(pid int) (teardown func(), err error) {
+	// TODO: allow users to configure rootless networking options
+	enableIPV6 := true
```
rhatdan, line 1066: This looks like it is available in containers.conf now.

```
# Default options to pass to the slirp4netns binary.
# Valid options values are:
#
# - allow_host_loopback=true|false: Allow the slirp4netns to reach the host loopback IP (`10.0.2.2`).
#   Default is false.
# - mtu=MTU: Specify the MTU to use for this network. (Default is `65520`).
# - cidr=CIDR: Specify ip range to use for this network. (Default is `10.0.2.0/24`).
# - enable_ipv6=true|false: Enable IPv6. Default is true. (Required for `outbound_addr6`).
# - outbound_addr=INTERFACE: Specify the outbound interface slirp should bind to (ipv4 traffic only).
# - outbound_addr=IPv4: Specify the outbound ipv4 address slirp should bind to.
# - outbound_addr6=INTERFACE: Specify the outbound interface slirp should bind to (ipv6 traffic only).
# - outbound_addr6=IPv6: Specify the outbound ipv6 address slirp should bind to.
# - port_handler=rootlesskit: Use rootlesskit for port forwarding. Default.
#   Note: Rootlesskit changes the source IP address of incoming packets to a IP address in the container
#   network namespace, usually `10.0.2.100`. If your application requires the real source IP address,
#   e.g. web server logs, use the slirp4netns port handler. The rootlesskit port handler is also used for
#   rootless containers when connected to user-defined networks.
# - port_handler=slirp4netns: Use the slirp4netns port forwarding, it is slower than rootlesskit but
#   preserves the correct source IP address. This port handler cannot be used for user-defined networks.
#
#network_cmd_options = []
```

run_linux.go
```diff
@@ -1088,12 +1091,71 @@ func setupRootlessNetwork(pid int) (teardown func(), err error) {
...
+	if enableIPV6 {
+		slirpArgs = append(slirpArgs, "--enable-ipv6")
+	}
+	slirpArgs = append(slirpArgs, "--mtu", "65520", "-r", "3", "-c", strconv.Itoa(pid), "tap0")
+	cmd := exec.Command(slirp4netns, slirpArgs...)
 	setPdeathsig(cmd)
 	cmd.Stdin, cmd.Stdout, cmd.Stderr = nil, nil, nil
 	cmd.ExtraFiles = []*os.File{rootlessSlirpSyncW}
 
+	ipv6ConfDefaultAcceptDadSysctl := "/proc/sys/net/ipv6/conf/default/accept_dad"
```
rhatdan, line 1104: Is this code common between podman and Buildah?

rhatdan:
Lets fix this now in containers/common and then vendor it into buildah and Podman for next release.  No reason to rush it now.

PR #3978, opened 2022-05-10
buildkit, build: add support for `additionalBuildContext` in builds via `--build-context`

flouthoc:
As of now for complicated builds the ability to only access files from one location became quite limiting OTOH we already support `multi-stage` builds where you can `copy` files from other parts of the Containerfile by adding the `--from` flag and pointing it to the name of another Containerfile stage or a remote image but this is still limited.

The new named build context feature is an extension of this pattern. You can now define additional build contexts when running the build command, give them a name, and then access them inside a Containerfile the same way you previously did with build stages.

Additional build contexts can be defined with a new `--build-context [name]=[value]` flag. The key component defines the name for your build context and the value can be:

```console
 * Local directory – e.g. --build-context project2=../path/to/project2/src
 * HTTP URL to a tarball – e.g. --build-context src=https://example.org/releases/src.tar
 * Container image – Define with a `docker-image://` prefix, e.g. --build-context alpine=docker-image://alpine:3.15, ( also supports `docker://`, `container-image://`)
 ```

On the Containerfile side, you can reference the build context on all commands that accept the `from` parameter. Here’s how that might look:
```Dockerfile
FROM [name]
COPY --from=[name] ...
RUN --mount=from=[name] …
```

The value of [name] is matched with the following priority order:

* Named build context defined with `--build-context [name]=..`
* Stage defined with `AS [name]` inside Dockerfile
* Remote image `[name]` in a container registry

## Added Features

* Pinning images for `FROM`,`COPY` and `ADD`.
* Specifying multiple buildcontexts from different projects
  and using them with `--from` in `ADD` and `COPY` directive
* Override a Remote Dependency with a Local One.
* Using additional context from external `Tar`

More Context: https://www.docker.com/blog/dockerfiles-now-support-multiple-build-contexts/

### DIFF
--- a/cmd/buildah/build.go
+++ b/cmd/buildah/build.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
@@ -153,6 +154,22 @@ func buildCmd(c *cobra.Command, inputArgs []string, iopts buildOptions) error {
 				} else {
 					delete(args, av[0])
 				}
+			}
+		}
+	}
+
+	additionalBuildContext := make(map[string]*define.AdditionalBuildContext)
+	if c.Flag("build-context").Changed {
+		for _, contextString := range iopts.BuildContext {
+			av := strings.SplitN(contextString, "=", 2)
+			if len(av) > 1 {
+				parseAdditionalBuildContext, err := parse.GetAdditionalBuildContext(av[1])
+				if err != nil {
+					return errors.Wrapf(err, "while parsing additional build context")
+				}
+				additionalBuildContext[av[0]] = &parseAdditionalBuildContext
+			} else {
+				return fmt.Errorf("while parsing additional build context: %q, accepts value in the form of key=value", av)
 			}
 		}
 	}
@@ -344,6 +361,7 @@ func buildCmd(c *cobra.Command, inputArgs []string, iopts buildOptions) error {
 		Annotations:             iopts.Annotation,
 		Architecture:            systemContext.ArchitectureChoice,
 		Args:                    args,
+		AdditionalBuildContexts: additionalBuildContext,
 		BlobDirectory:           iopts.BlobCache,
 		CNIConfigDir:            iopts.CNIConfigDir,
 		CNIPluginPath:           iopts.CNIPlugInPath,
--- a/define/build.go
+++ b/define/build.go
@@ -11,6 +11,21 @@ import (
 	"golang.org/x/sync/semaphore"
 )
 
+// AdditionalBuildContext contains verbose details about a parsed build context from --build-context
+type AdditionalBuildContext struct {
+	// Value is the URL of an external tar archive.
+	IsURL bool
+	// Value is the name of an image which may or may not have already been pulled.
+	IsImage bool
+	// Value holds a URL, an image name, or an absolute filesystem path.
+	Value string
+	// Absolute filesystem path to downloaded and exported build context
+	// from external tar archive. This will be populated only if following
+	// buildcontext is created from IsURL and was downloaded before in any
+	// of the RUN step.
+	DownloadedCache string
+}
+
 // CommonBuildOptions are resources that can be defined by flags for both buildah from and build
 type CommonBuildOptions struct {
 	// AddHost is the list of hostnames to add to the build container's /etc/hosts.
@@ -121,6 +136,8 @@ type BuildOptions struct {
 	Compression archive.Compression
 	// Arguments which can be interpolated into Dockerfiles
 	Args map[string]string
+	// Map of external additional build contexts
+	AdditionalBuildContexts map[string]*AdditionalBuildContext
 	// Name of the image to write to.
 	Output string
 	// BuildOutput specifies if any custom build output is selected for following build.
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -68,6 +68,32 @@ resulting image's configuration.
 Please refer to the [BUILD TIME VARIABLES](#build-time-variables) section for the
 list of variables that can be overridden within the Containerfile at run time.
 
+**--build-context** *name=value*
+
+Specify an additional build context using its short name and its location. Additional
+build contexts can be referenced in the same manner as we access different stages in `COPY`
+instruction.
+
+Valid values could be:
+* Local directory – e.g. --build-context project2=../path/to/project2/src
+* HTTP URL to a tarball – e.g. --build-context src=https://example.org/releases/src.tar
+* Container image – specified with a container-image:// prefix, e.g. --build-context alpine=container-image://alpine:3.15, (also accepts docker://, docker-image://)
+
+On the Containerfile side, you can reference the build context on all commands that accept the “from” parameter.
+Here’s how that might look:
+
+```Dockerfile
+FROM [name]
+COPY --from=[name] ...
+RUN --mount=from=[name] …
+```
+
+The value of `[name]` is matched with the following priority order:
+
+* Named build context defined with --build-context [name]=..
+* Stage defined with AS [name] inside Containerfile
+* Image [name], either local or in a remote registry
+
 **--cache-from**
 
 Images to utilise as potential cache sources. Buildah does not currently support --cache-from so this is a NOOP.
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -212,7 +212,10 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 	}
 
 	if options.AllPlatforms {
-		options.Platforms, err = platformsForBaseImages(ctx, logger, paths, files, options.From, options.Args, options.SystemContext)
+		if options.AdditionalBuildContexts == nil {
+			options.AdditionalBuildContexts = make(map[string]*define.AdditionalBuildContext)
+		}
+		options.Platforms, err = platformsForBaseImages(ctx, logger, paths, files, options.From, options.Args, options.AdditionalBuildContexts, options.SystemContext)
 		if err != nil {
 			return "", nil, err
 		}
@@ -512,8 +515,8 @@ func preprocessContainerfileContents(logger *logrus.Logger, containerfile string
 // platformsForBaseImages resolves the names of base images from the
 // dockerfiles, and if they are all valid references to manifest lists, returns
 // the list of platforms that are supported by all of the base images.
-func platformsForBaseImages(ctx context.Context, logger *logrus.Logger, dockerfilepaths []string, dockerfiles [][]byte, from string, args map[string]string, systemContext *types.SystemContext) ([]struct{ OS, Arch, Variant string }, error) {
-	baseImages, err := baseImages(dockerfilepaths, dockerfiles, from, args)
+func platformsForBaseImages(ctx context.Context, logger *logrus.Logger, dockerfilepaths []string, dockerfiles [][]byte, from string, args map[string]string, additionalBuildContext map[string]*define.AdditionalBuildContext, systemContext *types.SystemContext) ([]struct{ OS, Arch, Variant string }, error) {
+	baseImages, err := baseImages(dockerfilepaths, dockerfiles, from, args, additionalBuildContext)
 	if err != nil {
 		return nil, errors.Wrapf(err, "determining list of base images")
 	}
@@ -641,7 +644,7 @@ func platformsForBaseImages(ctx context.Context, logger *logrus.Logger, dockerfi
 // stage's base image with FROM, and returns the list of base images as
 // provided.  Each entry in the dockerfilenames slice corresponds to a slice in
 // dockerfilecontents.
-func baseImages(dockerfilenames []string, dockerfilecontents [][]byte, from string, args map[string]string) ([]string, error) {
+func baseImages(dockerfilenames []string, dockerfilecontents [][]byte, from string, args map[string]string, additionalBuildContext map[string]*define.AdditionalBuildContext) ([]string, error) {
 	mainNode, err := imagebuilder.ParseDockerfile(bytes.NewReader(dockerfilecontents[0]))
 	if err != nil {
 		return nil, errors.Wrapf(err, "error parsing main Dockerfile: %s", dockerfilenames[0])
@@ -679,6 +682,13 @@ func baseImages(dockerfilenames []string, dockerfilecontents [][]byte, from stri
 						if from != "" {
 							child.Next.Value = from
 							from = ""
+						}
+						if replaceBuildContext, ok := additionalBuildContext[child.Next.Value]; ok {
+							if replaceBuildContext.IsImage {
+								child.Next.Value = replaceBuildContext.Value
+							} else {
+								return nil, fmt.Errorf("build context %q is not an image, can not be used for FROM %q", child.Next.Value, child.Next.Value)
+							}
 						}
 						base := child.Next.Value
 						if base != "scratch" && !nicknames[base] {
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -126,6 +126,7 @@ type Executor struct {
 	imageInfoLock           sync.Mutex
 	imageInfoCache          map[string]imageTypeAndHistoryAndDiffIDs
 	fromOverride            string
+	additionalBuildContexts map[string]*define.AdditionalBuildContext
 	manifest                string
 	secrets                 map[string]define.Secret
 	sshsources              map[string]*sshagent.Source
@@ -275,6 +276,7 @@ func newExecutor(logger *logrus.Logger, logPrefix string, store storage.Store, o
 		rusageLogFile:                  rusageLogFile,
 		imageInfoCache:                 make(map[string]imageTypeAndHistoryAndDiffIDs),
 		fromOverride:                   options.From,
+		additionalBuildContexts:        options.AdditionalBuildContexts,
 		manifest:                       options.Manifest,
 		secrets:                        secrets,
 		sshsources:                     sshsources,
@@ -609,6 +611,12 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 						}
 						base := child.Next.Value
 						if base != "scratch" {
+							if replaceBuildContext, ok := b.additionalBuildContexts[child.Next.Value]; ok {
+								if replaceBuildContext.IsImage {
+									child.Next.Value = replaceBuildContext.Value
+									base = child.Next.Value
+								}
+							}
 							userArgs := argsMapToSlice(stage.Builder.Args)
 							baseWithArg, err := imagebuilder.ProcessWord(base, userArgs)
 							if err != nil {
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -369,18 +369,73 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
 			if fromErr != nil {
 				return errors.Wrapf(fromErr, "unable to resolve argument %q", copy.From)
 			}
-			if isStage, err := s.executor.waitForStage(s.ctx, from, s.stages[:s.index]); isStage && err != nil {
-				return err
-			}
-			if other, ok := s.executor.stages[from]; ok && other.index < s.index {
-				contextDir = other.mountPoint
-				idMappingOptions = &other.builder.IDMappingOptions
-			} else if builder, ok := s.executor.containerMap[copy.From]; ok {
-				contextDir = builder.MountPoint
-				idMappingOptions = &builder.IDMappingOptions
+			var additionalBuildContext *define.AdditionalBuildContext
+			if foundContext, ok := s.executor.additionalBuildContexts[from]; ok {
+				additionalBuildContext = foundContext
 			} else {
-				return errors.Errorf("the stage %q has not been built", copy.From)
+				// Maybe index is given in COPY --from=index
+				// if that's the case check if provided index
+				// exists and if stage short_name matches any
+				// additionalContext replace stage with addtional
+				// build context.
+				if _, err := strconv.Atoi(from); err == nil {
+					if stage, ok := s.executor.stages[from]; ok {
+						if foundContext, ok := s.executor.additionalBuildContexts[stage.name]; ok {
+							additionalBuildContext = foundContext
+						}
+					}
+				}
 			}
+			if additionalBuildContext != nil {
+				if !additionalBuildContext.IsImage {
+					contextDir = additionalBuildContext.Value
+					if additionalBuildContext.IsURL {
+						// Check if following buildContext was already
+						// downloaded before in any other RUN step. If not
+						// download it and populate DownloadCache field for
+						// future RUN steps.
+						if additionalBuildContext.DownloadedCache == "" {
+							// additional context contains a tar file
+							// so download and explode tar to buildah
+							// temp and point context to that.
+							path, subdir, err := define.TempDirForURL(internalUtil.GetTempDir(), internal.BuildahExternalArtifactsDir, additionalBuildContext.Value)
+							if err != nil {
+								return errors.Wrapf(err, "unable to download context from external source %q", additionalBuildContext.Value)
+							}
+							// point context dir to the extracted path
+							contextDir = filepath.Join(path, subdir)
+							// populate cache for next RUN step
+							additionalBuildContext.DownloadedCache = contextDir
+						} else {
+							contextDir = additionalBuildContext.DownloadedCache
+						}
+					}
+				} else {
+					copy.From = additionalBuildContext.Value
+				}
+			}
+			if additionalBuildContext == nil {
+				if isStage, err := s.executor.waitForStage(s.ctx, from, s.stages[:s.index]); isStage && err != nil {
+					return err
+				}
+				if other, ok := s.executor.stages[from]; ok && other.index < s.index {
+					contextDir = other.mountPoint
+					idMappingOptions = &other.builder.IDMappingOptions
+				} else if builder, ok := s.executor.containerMap[copy.From]; ok {
+					contextDir = builder.MountPoint
+					idMappingOptions = &builder.IDMappingOptions
+				} else {
+					return errors.Errorf("the stage %q has not been built", copy.From)
+				}
+			} else if additionalBuildContext.IsImage {
+				// Image was selected as additionalContext so only process image.
+				mountPoint, err := s.getImageRootfs(s.ctx, copy.From)
+				if err != nil {
+					return err
+				}
+				contextDir = mountPoint
+			}
+			// Original behaviour of buildah still stays true for COPY irrespective of additional context.
 			preserveOwnership = true
 			copyExcludes = excludes
 		} else {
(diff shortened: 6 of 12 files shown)


flouthoc:
More about this feature here: https://www.docker.com/blog/dockerfiles-now-support-multiple-build-contexts/

nalind (changes requested):
Does a stage named using "FROM base AS name" but referred to in a "COPY --from=" instruction that refers to it by its numeric position require additional handling?

Needs to test cases with a stage named using "FROM base AS name" syntax and a build context named "name" to check that "COPY --from" looks in the right place, and that "FROM name" does the expected thing.

Needs to test "RUN --mount from=...." pointing to the various kinds of additional build contexts.

The patch touches `platformsForBaseImages()`, so we should test how it affects the `--all-platforms` build flag.

docs/buildah-build.1.md
```diff
@@ -68,6 +68,32 @@ resulting image's configuration.
...
+```Dockerfile
+FROM [name]
+COPY --from=[name] ...
+RUN --mount=from=[name] …
+```
+
+The value of `[name]` is matched with the following priority order:
+
+* Named build context defined with --build-context [name]=..
+* Stage defined with AS [name] inside Dockerfile
```
TomSweeneyRedHat, line 94: Dockerfile here, Containerfile above.  It should be consistent.

define/build.go
```diff
@@ -121,6 +131,8 @@ type BuildOptions struct {
 	Compression archive.Compression
 	// Arguments which can be interpolated into Dockerfiles
 	Args map[string]string
+	// Map of external additional build contexts
+	BuildContext map[string]AdditionalBuildContext
```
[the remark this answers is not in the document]
flouthoc: Renamed to `AdditionalBuildContexts`

imagebuildah/build.go
```diff
@@ -670,6 +670,11 @@ func baseImages(dockerfilenames []string, dockerfilecontents [][]byte, from stri
 							child.Next.Value = from
 							from = ""
 						}
+						if replaceBuildContext, ok := additionalBuildContext[child.Next.Value]; ok {
+							if replaceBuildContext.IsImage {
+								child.Next.Value = replaceBuildContext.Value
+							}
+						}
```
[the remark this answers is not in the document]
flouthoc: Yes giving a warning here that `additional-build-context` is ignored instead of silently ignoring.

imagebuildah/stage_executor.go
```diff
@@ -362,24 +362,48 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
+						// so download and explode tar to buildah
+						// temp and point context to that.
+						untarPath, err := internalUtil.ExportTarFromURL(additionalBuildContext.Value, from)
+						if err != nil {
+							return errors.Wrapf(err, "unable to extract Tar from external source %q", additionalBuildContext.Value)
+						}
+						// point contextDir to untar
+						contextDir = untarPath
+					}
+					didSelectAdditionalContext = true
```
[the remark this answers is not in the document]
flouthoc: Added a caching property to `AddtionalBuildContext`

imagebuildah/stage_executor.go
```diff
@@ -923,6 +956,22 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
+
+				if additionalBuildContext, ok := s.executor.buildContext[from]; ok {
+					if !additionalBuildContext.IsImage {
+						// We don't need to pull this
+						// since this additional context
+						// is not an image.
+						break
+					} else {
+						// replace with image set in build context
+						from = additionalBuildContext.Value
```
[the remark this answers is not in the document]
flouthoc: Since `additonal-context` gets priority anyways hence moved checking for image in its seperate block now.

imagebuildah/executor.go
```diff
@@ -626,6 +633,11 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 							// was named using argument values, we might
 							// not record the right value here.
 							rootfs := strings.TrimPrefix(flag, "--from=")
+							if replaceBuildContext, ok := b.buildContext[rootfs]; ok {
+								if replaceBuildContext.IsImage {
+									rootfs = replaceBuildContext.Value
+								}
+							}
```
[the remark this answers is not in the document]
flouthoc: This is not needed i have removed it.

imagebuildah/stage_executor.go
```diff
@@ -446,6 +470,15 @@ func (s *StageExecutor) runStageMountPoints(mountList []string) (map[string]inte
...
 					}
+					// If additional buildContext contains this
+					// give priority to that and break if additional
+					// is not an external image.
+
+					if additionalBuildContext, ok := s.executor.buildContext[from]; ok {
+						if additionalBuildContext.IsImage {
+							from = additionalBuildContext.Value
+						}
+					}
```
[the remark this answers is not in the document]
flouthoc: Handles now also added test for `RUN --mount` to test this.

imagebuildah/stage_executor.go
```diff
@@ -362,24 +362,48 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
+					contextDir = other.mountPoint
+					idMappingOptions = &other.builder.IDMappingOptions
+				} else if builder, ok := s.executor.containerMap[copy.From]; ok {
+					contextDir = builder.MountPoint
+					idMappingOptions = &builder.IDMappingOptions
+				} else {
+					return errors.Errorf("the stage %q has not been built", copy.From)
+				}
 			}
 			preserveOwnership = true
```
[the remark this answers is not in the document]
flouthoc: Yes ownership is preserved. I also a comment for this.

flouthoc:
> Does a stage named using "FROM base AS name" but referred to in a "COPY --from=" instruction that refers to it by its numeric position require additional handling?
> 
> Needs to test cases with a stage named using "FROM base AS name" syntax and a build context named "name" to check that "COPY --from" looks in the right place, and that "FROM name" does the expected thing.
> 
> Needs to test "RUN --mount from=...." pointing to the various kinds of additional build contexts.
> 
> The patch touches `platformsForBaseImages()`, so we should test how it affects the `--all-platforms` build flag.

@nalind Added new tests as listed below which should cover all the requested scenarios, also added comments where we match with `buildx` parity.

```console
ok 1 build-with-additional-build-context and COPY, test pinning image
ok 2 build-with-additional-build-context and COPY, stagename and additional-context conflict
ok 3 build-with-additional-build-context and COPY, additionalContext and numeric value of stage
ok 4 build-with-additional-build-context and FROM, stagename and additional-context conflict
ok 5 build-with-additional-build-context and COPY, additional context from host
ok 6 build-with-additional-build-context and COPY, additional context from external URL
ok 7 build-with-additional-build-context and FROM, pin busybox to alpine
ok 8 build-with-additional-build-context and RUN --mount=from=, additional-context and also test conflict with stagename
ok 9 build-with-additional-build-context and RUN --mount=from=, additional-context not image and also test conflict with stagename
ok 10 bud-multiple-platform for --all-platform with additional-build-context

File: cmd/buildah/build.go [hunk 9]
```diff
@@ -157,6 +157,16 @@ func buildCmd(c *cobra.Command, inputArgs []string, iopts buildOptions) error {
 		}
 	}
 
+	additionalBuildContext := make(map[string]define.AdditionalBuildContext)
+	if c.Flag("build-context").Changed {
+		for _, contextString := range iopts.BuildContext {
+			av := strings.SplitN(contextString, "=", 2)
+			if len(av) > 1 {
+				additionalBuildContext[av[0]] = parse.GetAdditionalBuildContext(av[1])
+			}
```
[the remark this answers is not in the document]
nalind: Still silently ignores values that don't include `=`.

docs/buildah-build.1.md
```diff
@@ -68,6 +68,32 @@ resulting image's configuration.
...
+**--build-context** *name=value*
+
+Specify an additional build context using its short name and its location. Additional
+build contexts can be referenced in the same manner as we access different stages in `COPY`
+instruction.
+
+Valid values could be:
+* Local directory – e.g. --build-context project2=../path/to/project2/src
+* HTTP URL to a tarball – e.g. --build-context src=https://example.org/releases/src.tar
+* Container image – specified with a container-image:// prefix, e.g. --build-context alpine=container-image://alpine:3.15, ( also accepts docker://, docker-image://)
```
nalind, line 80:
```suggestion
* Container image – specified with a container-image:// prefix, e.g. --build-context alpine=container-image://alpine:3.15, (also accepts docker://, docker-image://)
```

imagebuildah/build.go
```diff
@@ -211,7 +211,12 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 	}
 
 	if options.AllPlatforms {
-		options.Platforms, err = platformsForBaseImages(ctx, logger, paths, files, options.From, options.Args, options.SystemContext)
+		additionalBuildContexts := options.AdditionalBuildContexts
+		if additionalBuildContexts == nil {
+			// Create empty map for additionalbuildContexts
+			additionalBuildContexts = make(map[string]*define.AdditionalBuildContext)
+		}
```
nalind, line 218: For this case you're going to want to overwrite `options.AdditionalBuildContexts` so that the various `platformOptions` structures created below don't see `nil` values for that field.

imagebuildah/stage_executor.go
```diff
@@ -363,24 +363,78 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
 			// If from has an argument within it, resolve it to its
 			// value.  Otherwise just return the value found.
 			from, fromErr := imagebuilder.ProcessWord(copy.From, s.stage.Builder.Arguments())
 			if fromErr != nil {
 				return errors.Wrapf(fromErr, "unable to resolve argument %q", copy.From)
 			}
-			if isStage, err := s.executor.waitForStage(s.ctx, from, s.stages[:s.index]); isStage && err != nil {
-				return err
+			if additionalBuildContext, ok := s.executor.additionalBuildContexts[from]; ok {
+				if additionalBuildContext != nil {
```
nalind, line 378: If we never put `nil` values in the map, then we don't need to check for them.

imagebuildah/stage_executor.go
```diff
@@ -363,24 +363,78 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
+							// Check if following buildContext was already
+							// downloaded before in any other RUN step. If yes
+							// lets check if we are hitting any cache and cache
+							// actually exists on host.
+							cacheExistsOnHost := false
+							if additionalBuildContext.DownloadedCache != "" {
+								if stat, err := os.Stat(additionalBuildContext.DownloadedCache); err == nil && stat.IsDir() {
+									cacheExistsOnHost = true
+								}
+							}
```
nalind, line 392: It looks like `cacheExistsOnHost` can be eliminated, since `DownloadedCache` is only ever set after the directory is created and populated.

imagebuildah/stage_executor.go
```diff
@@ -363,24 +363,78 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
+					idMappingOptions = &other.builder.IDMappingOptions
+				} else if builder, ok := s.executor.containerMap[copy.From]; ok {
+					contextDir = builder.MountPoint
+					idMappingOptions = &builder.IDMappingOptions
+				} else {
+					return errors.Errorf("the stage %q has not been built", copy.From)
+				}
+			} else if didSelectImageAsAdditionalContext {
+				// Image was selected as additionalContext so only process image.
+				if builder, ok := s.executor.containerMap[copy.From]; ok {
```
nalind, line 430: Use `s.getImageRootfs(ctx, from)` here.

imagebuildah/stage_executor.go
```diff
@@ -446,6 +500,63 @@ func (s *StageExecutor) runStageMountPoints(mountList []string) (map[string]inte
 					if fromErr != nil {
 						return nil, errors.Wrapf(fromErr, "unable to resolve argument %q", kv[1])
 					}
+					// If additional buildContext contains this
+					// give priority to that and break if additional
+					// is not an external image.
+					if additionalBuildContext, ok := s.executor.additionalBuildContexts[from]; ok {
+						if additionalBuildContext != nil {
```
nalind, line 507: If we don't set `nil` values in the map, we don't need to check for them.

imagebuildah/stage_executor.go
```diff
@@ -446,6 +500,63 @@ func (s *StageExecutor) runStageMountPoints(mountList []string) (map[string]inte
...
+									// downloaded before in any other RUN step. If yes
+									// lets check if we are hitting any cache and cache
+									// actually exists on host.
+									cacheExistsOnHost := false
+									if additionalBuildContext.DownloadedCache != "" {
+										if stat, err := os.Stat(additionalBuildContext.DownloadedCache); err == nil && stat.IsDir() {
+											cacheExistsOnHost = true
+										}
+									}
+									if !cacheExistsOnHost {
```
nalind, line 539: It looks like `cacheExistsOnHost` can be eliminated, since `DownloadedCache` is only ever set after the directory is created and populated.

imagebuildah/stage_executor.go
```diff
@@ -923,6 +1034,27 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 				if fromErr != nil {
 					return "", nil, errors.Wrapf(fromErr, "unable to resolve argument %q", arr[1])
 				}
+				// If additional buildContext contains this
+				// give priority to that and break if additional
+				// is not an external image.
+				if additionalBuildContext, ok := s.executor.additionalBuildContexts[from]; ok {
+					if additionalBuildContext != nil {
```
nalind, line 1041: If we don't set `nil` values in the map, we don't need to check for them.

imagebuildah/stage_executor.go
```diff
@@ -923,6 +1034,27 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
+						if !additionalBuildContext.IsImage {
+							// We don't need to pull this
+							// since this additional context
+							// is not an image.
+							break
+						} else {
+							// replace with image set in build context
+							from = additionalBuildContext.Value
+							if _, err := s.getImageRootfs(ctx, from); err != nil {
+								return "", nil, errors.Errorf("%s --from=%s: no stage or image found with that name", command, from)
```
nalind, line 1051:
```suggestion
								return "", nil, errors.Errorf("%s --from=%s: no image found with that name", command, from)
```

pkg/parse/parse.go
```diff
@@ -175,6 +175,31 @@ func CommonBuildOptionsFromFlagSet(flags *pflag.FlagSet, findFlagFunc func(name
...
+	} else if strings.HasPrefix(value, "docker://") {
+		ret.IsImage = true
+		ret.Value = strings.TrimPrefix(value, "docker://")
+	} else if strings.HasPrefix(value, "http://") || strings.HasPrefix(value, "https://") {
+		ret.IsImage = false
+		ret.IsURL = true
+	} else {
+		path, err := filepath.Abs(value)
+		if err != nil {
+			return ret, errors.Wrapf(err, "unable to convert additional build-context %q path to absolute", value)
```
nalind, line 196: On error, return an empty value instead of a partially-initialized one.  Or change this to return a pointer and return `nil` with the error.

tests/bud.bats
```diff
@@ -202,6 +202,202 @@ _EOF
...
+_EOF
+
+  # Build a first image which we can use as source
+  run_buildah build $WITH_POLICY_JSON -t source -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile1
+  # Pin upstream busybox to local image source
+  run_buildah build $WITH_POLICY_JSON --build-context some-stage=docker://source -t test -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile2
+  expect_output --substring "world"
+}
+
+# Honor copy from stage when used its numeric value
```
nalind, line 255: Is it supposed to do this?

tests/bud.bats
```diff
@@ -202,6 +202,202 @@ _EOF
...
+
+# Test adding additional build context
+@test "build-with-additional-build-context and COPY, additional context from host" {
+  mkdir -p ${TEST_SCRATCH_DIR}/bud/platform
+  mkdir -p ${TEST_SCRATCH_DIR}/bud/platform2
+
+  # add file on original context
+  echo something > ${TEST_SCRATCH_DIR}/bud/platform/somefile
+  # add file on additional context
+  echo hello > ${TEST_SCRATCH_DIR}/bud/platform2/hello
```
nalind, line 309: Give the file contents that aren't also a part of its name, so that we can be certain that the `expect_output` below matches the file's contents and not its name, which we always print.

tests/bud.bats
```diff
@@ -202,6 +202,202 @@ _EOF
...
+  cat > ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile << _EOF
+FROM alpine
+COPY --from=crun-context . .
+RUN ls -a
+_EOF
+
+  # Test additional context but download from tar
+  run_buildah build $WITH_POLICY_JSON -t source --build-context crun-context=https://github.com/containers/crun/releases/download/1.4.5/crun-1.4.5.tar.xz ${TEST_SCRATCH_DIR}/bud/platform
+  # additional context from tar must show crun binary inside container
+  expect_output --substring "crun-1.4.5"
```
nalind, line 338: This should be checking for a pattern that only matches the name of a file in the archive, and not a pattern that is also part of the archive's name, in case we break something and stop extracting its contents in the future.

tests/bud.bats
```diff
@@ -202,6 +202,202 @@ _EOF
...
+
+  cat > ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile << _EOF
+FROM busybox
+RUN ls /etc/*release
+_EOF
+
+  # Test additional context but download from tar
+  # We are pinning busybox to alpine so we must always pull alpine and use that
+  run_buildah build $WITH_POLICY_JSON -t source --build-context busybox=docker://alpine ${TEST_SCRATCH_DIR}/bud/platform
+  expect_output --substring "alpine"
```
nalind, line 353: Is this attempting to match the path to a file whose name `ls` will print, or match one of the messages that we log while attempting to pull the image?  Either way, this should be attempting to match a more specific pattern.

tests/bud.bats
```diff
@@ -202,6 +202,202 @@ _EOF
...
+# hello should get copied since we are giving priority to additional context
+FROM alpine
+RUN --mount=type=bind,from=some-stage,target=/test cat /test/hello
+_EOF
+
+  # Build a first image which we can use as source
+  run_buildah build $WITH_POLICY_JSON -t source -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile1
+  # Additional Context for RUN --mount is additional image and it should not conflict with stage
+  run_buildah build $WITH_POLICY_JSON --build-context some-stage=docker://source -t test -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile2
+  expect_output --substring "world"
```
nalind, line 379: Is this attempting to match the contents of a file, or the text of an instruction that the build will log as part of its output?

tests/bud.bats
```diff
@@ -3818,6 +4014,20 @@ _EOF
   [[ "$output" -gt 1 ]] # should at least be more than one entry in there, right?
 }
 
+@test "bud-multiple-platform for --all-platform with additional-build-context" {
+  outputlist=localhost/testlist
+    mkdir -p ${TEST_SCRATCH_DIR}/bud/platform
```
nalind, line 4019: Indentation is weird here.

tests/bud.bats
```diff
@@ -3818,6 +4014,20 @@ _EOF
...
+
+cat > ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile1 << _EOF
+FROM busybox
+_EOF
+
+  # Pulled images must be ubi since we configured --build-context busybox=docker://registry.access.redhat.com/ubi8-micro
+  run_buildah build $WITH_POLICY_JSON --jobs=1 --all-platforms --build-context busybox=docker://registry.access.redhat.com/ubi8-micro --manifest $outputlist -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile1
```
nalind, line 4026: `--jobs=1` is the default, so it doesn't need to be specified.  This test should also compare how many entries end up in `$outputlist` to the number of images in the registry's `ubi8-micro` manifest list.

nalind:
Should have a test with "RUN --mount=type=bind,from=..." where the additional build context specified as the "from" value was given as a URL, and at least one of the "RUN --mount" tests should be using the "src" option to look in a subdirectory of the additional build context, to make sure we're not forgetting that in the logic.

imagebuildah/stage_executor.go
flouthoc: Just added a check so in future if someone edits top level code we don't endup SEGSEV. But I can drop it if needed.

imagebuildah/stage_executor.go
flouthoc: Same just added to prevent any unintentional SEGSEV in future if someone edits a top level code.

imagebuildah/stage_executor.go
flouthoc: Same as above.

tests/bud.bats
flouthoc: Buildkit does not do this instead it does the opposite and does not honors stage's numeric value and still uses additional-context but afaik docs and written blog only talks about `short name` so this seems a bug to me on `buildx/bulildkit` end as user should still be able to access original stage using it numeric value.

tests/bud.bats
flouthoc: Yes change it to something else

tests/bud.bats
flouthoc: now checking for specific files `libcrun`.

flouthoc: now checking for specific files `libcrun`.

tests/bud.bats
flouthoc: checking for `alpine-release` since it should not be present on busybox.

tests/bud.bats
flouthoc: Changed above test to other than world, this expectes to print content of mounted file now from other image.

tests/bud.bats
flouthoc: fixed.

tests/bud.bats
flouthoc: Checking the length of manifest list now it should be > 1 just like we do in `no-run` test.

flouthoc:
> 

@nalind Added a new test `build-with-additional-build-context and RUN --mount=from=, additional-context is URL and mounted from subdir` that should cover both.

cmd/buildah/build.go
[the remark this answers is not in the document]
flouthoc: should handle now.

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
flouthoc: Yes buildkit does it for when context is external tar and for case where files copied from host as well.

imagebuildah/stage_executor.go
nalind: Go ahead and drop it.  I can't conceive of a use case for setting `nil` values in this map.

tests/bud.bats
flouthoc: Matched this behavior with buildkit to keep conformance tests happy.

imagebuildah/stage_executor.go
flouthoc: dropped this nil and other similar nil check for additionalBuildContext.

flouthoc:
@nalind Could you PTAL again.

tests/bud.bats
nalind: To be clear, we don't currently exercise this in conformance tests.

flouthoc: Yes. We don't cover buldkit test in conformance.

cmd/buildah/build.go
```diff
@@ -157,6 +157,22 @@ func buildCmd(c *cobra.Command, inputArgs []string, iopts buildOptions) error {
...
+		for _, contextString := range iopts.BuildContext {
+			av := strings.SplitN(contextString, "=", 2)
+			if len(av) > 1 {
+				parseAdditionalBuildContext, err := parse.GetAdditionalBuildContext(av[1])
+				if err != nil {
+					return errors.Wrapf(err, "while parsing additional build context")
+				}
+				additionalBuildContext[av[0]] = &parseAdditionalBuildContext
+			} else {
+				return errors.New("while parsing additional build context, accepts value in the form of key=value")
```
nalind, line 171: This can go ahead and quote the value that couldn't be correctly parsed here.

imagebuildah/build.go
```diff
@@ -211,7 +211,11 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 	}
 
 	if options.AllPlatforms {
-		options.Platforms, err = platformsForBaseImages(ctx, logger, paths, files, options.From, options.Args, options.SystemContext)
+		if options.AdditionalBuildContexts == nil {
+			// Create empty map for additionalbuildContexts
```
nalind, line 215:
```suggestion
```

imagebuildah/build.go
```diff
@@ -670,6 +674,13 @@ func baseImages(dockerfilenames []string, dockerfilecontents [][]byte, from stri
 							child.Next.Value = from
 							from = ""
 						}
+						if replaceBuildContext, ok := additionalBuildContext[child.Next.Value]; ok {
+							if replaceBuildContext.IsImage {
+								child.Next.Value = replaceBuildContext.Value
+							} else {
+								logger.Warnf("additionalContext for nickname %q was found but its not an image hence it will be not used with FROM %q", child.Next.Value, child.Next.Value)
```
nalind, line 681:
```suggestion
								logger.Warnf("build context %q is not an image, can not be used for FROM %q", child.Next.Value, child.Next.Value)
```

imagebuildah/executor.go
```diff
@@ -609,6 +611,12 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 						}
 						base := child.Next.Value
 						if base != "scratch" {
+							if replaceBuildContext, ok := b.additionalBuildContexts[child.Next.Value]; ok {
+								if replaceBuildContext != nil && replaceBuildContext.IsImage {
```
nalind, line 615:
```suggestion
								if replaceBuildContext.IsImage {
```

imagebuildah/stage_executor.go
```diff
@@ -363,24 +363,90 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
+				if _, err := strconv.Atoi(from); err == nil {
+					if stage, ok := s.executor.stages[from]; ok {
+						if foundContext, ok := s.executor.additionalBuildContexts[stage.name]; ok {
+							foundAdditionalBuildContext = true
+							additionalBuildContext = foundContext
+						}
+					}
+				}
+			}
+			if foundAdditionalBuildContext {
```
nalind, line 398:
```suggestion
			if additionalBuildContext != nil {
```

imagebuildah/stage_executor.go
```diff
@@ -363,24 +363,90 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
+						} else {
+							contextDir = additionalBuildContext.DownloadedCache
+						}
+					}
+				} else {
+					didSelectImageAsAdditionalContext = true
+					copy.From = additionalBuildContext.Value
+				}
+			}
+			if !didSelectAdditionalContext {
```
nalind, line 428:
```suggestion
			if additionalBuildContext == nil {
```

imagebuildah/stage_executor.go
```diff
@@ -363,24 +363,90 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
+				if other, ok := s.executor.stages[from]; ok && other.index < s.index {
+					contextDir = other.mountPoint
+					idMappingOptions = &other.builder.IDMappingOptions
+				} else if builder, ok := s.executor.containerMap[copy.From]; ok {
+					contextDir = builder.MountPoint
+					idMappingOptions = &builder.IDMappingOptions
+				} else {
+					return errors.Errorf("the stage %q has not been built", copy.From)
+				}
+			} else if didSelectImageAsAdditionalContext {
```
nalind, line 441:
```suggestion
			} else if additionalBuildContext.IsImage {
```

nalind (changes requested):
Nits in error messages, a couple of redundant variables, and would like an answer about https://github.com/containers/buildah/pull/3978/files#r871374285.

tests/bud.bats
nalind: `--jobs=1` is the default, so it doesn't need to be specified.

imagebuildah/build.go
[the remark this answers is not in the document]
flouthoc: I have changed this to error since user did this intentionally and it is not supposed to work so behavior which user is expecting and which buildah is giving might differ hence error makes more sense here.

nalind:
There's still a redundant `--jobs=1` in the `bud-multiple-platform for --all-platform with additional-build-context` test, otherwise LGTM.

flouthoc:

rhatdan:
Nice work @flouthoc 
/lgtm

PR #3990, opened 2022-05-13
build: accept branch and subdirectory when context is `git` repo

flouthoc:
When using `build -t <tag> <some-url-to-git-repo>` buildah does not
accepts syntax `url#branch:subdir` following commit allows end users to
switch context to any subdirectory in the git repo.

Usage
```console
buildah build -t test 'https://github.com/containers/podman.git#main:contrib/hello'
```

Closes: https://github.com/containers/buildah/issues/3989


define/types.go
```diff
@@ -170,17 +175,30 @@ func TempDirForURL(dir, prefix, url string) (name string, subdir string, err err
...
+		if len(gitSubDirPart) > 1 {
+			gitSubdir = gitSubDirPart[1]
+			gitBranch = gitSubDirPart[0]
+		} else {
+			gitBranch = gitSubDirPart[0]
+		}
+	}
 	var cmd *exec.Cmd
-	if len(gitBranch) < 2 {
+	if len(gitBranchPart) < 2 {
```
rhatdan, line 193: You just did this check above, Combine this section with the if len(gitBranchPart) > 1 above

flouthoc: Yes moved it below.

rhatdan:
LGTM
@saschagrunert @vrothberg @giuseppe @nalind @flouthoc @ashley-cui @umohnani8 PTAL

vrothberg (commented):
LGTM but I would love to have unit tests for the _parsing_

tests/bud.bats
```diff
@@ -978,6 +978,13 @@ function _test_http() {
   run_buildah from ${target}
 }
 
+@test "bud-github-context-with-branch-and-subdir" {
+  target=github-image
+  gitrepo=https://github.com/containers/podman.git#main:contrib/hello
+  run_buildah build $WITH_POLICY_JSON -t ${target} "${gitrepo}"
+  run_buildah from ${target}
```
nalind, line 985: What are we testing with this line?

nalind (changes requested):
Current version appears to only accept subdirectory names if a branch name is provided, while I'm not aware that specifying either of them should require specifying the other.
Needs to update the documentation.

define/types.go
```diff
@@ -170,17 +175,29 @@ func TempDirForURL(dir, prefix, url string) (name string, subdir string, err err
...
+		// check if string contains path to a subdir
+		gitSubDirPart := strings.Split(gitBranchPart[1], ":")
+		if len(gitSubDirPart) > 1 {
+			gitSubdir = gitSubDirPart[1]
+		}
+		gitBranch = gitSubDirPart[0]
+	}
+	if gitBranch == "" {
+		logrus.Debugf("cloning %q to %q", gitBranchPart[0], dir)
+		cmd = exec.Command("git", "clone", gitBranchPart[0], dir)
```
nalind, line 193: The "branch name was specified" path uses the "--recurse-submodules" option.  Shouldn't it be used here, too?

docs/buildah-build.1.md
```diff
@@ -24,7 +24,7 @@ When the URL is an archive, the contents of the URL is downloaded to a temporary
 
 When the URL is a Containerfile, the file is downloaded to a temporary location.
 
-When a Git repository is set as the URL, the repository is cloned locally and then set as the context.
+When a Git repository is set as the URL, the repository is cloned locally and then set as the context. Buildah supports switching context on the cloned git repository using syntax `myrepo.git#mybranch:subdir` or `myrepo.git#:subdir` to switch context to a sub directory on the default branch.
```
nalind, line 27:
```suggestion
When a Git repository is set as the URL, the repository is cloned locally and then used as the build context.  A non-default branch and subdirectory of the cloned git repository can be used by including their names at the end of the URL in the form `myrepo.git#mybranch:subdir`, or `myrepo.git#:subdir` if the subdirectory should be used from the default branch.
```

define/types.go
flouthoc: Yes I think it should don't know why it was without `--recurse-submodules` before, just added it.

flouthoc:
@nalind PTAL

nalind:
LGTM

PR #3995, opened 2022-05-16
test helpers: default to being rootless-aware

nalind:
#### What type of PR is this?

/kind other

Make `tests/copy/copy.go` and `tests/imgtype/imgtype.go` aware of rootless defaults, and have them re-exec themselves in user namespaces if they're run in such an environment.  This will make it easier to use `imgtype` to read manifest lists from registries in tests.

#### How to verify it

Current tests should not be impacted.
`imgtype -show-manifest -expected-manifest-type '*' ...` should now have no trouble reading manifest lists from remote registries. 

None

```release-note
None
```

### DIFF
--- a/tests/copy/copy.go
+++ b/tests/copy/copy.go
@@ -38,6 +38,11 @@ func main() {
 
 	unshare.MaybeReexecUsingUserNamespace(false)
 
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

Issue #3993, opened 2022-05-13
Secret cannot be used more than once in a Containerfile


teojgo:
**Description**
Using the `--secret` option of `buildah bud` does not allow the same secret to be used more than once in a Containerfile

**Steps to reproduce the issue:**
1. Create a simple Containerfile with the following contents:
```
FROM ubuntu:latest

RUN --mount=type=secret,id=mysecret,dst=/home/root/mysecret cat /home/root/mysecret

RUN --mount=type=secret,id=mysecret,dst=/home/root/mysecret2 cat /home/root/mysecret2
```
2. Create a `mysecret.dat` text file with some random contents within the same directory
3. Try building the image using: `buildah bud  --secret=id=mysecret,src=./mysecret.dat`. You should get the following error:
```
cat: /home/root/mysecret2: No such file or directory
```
in the 3rd Containerfile step
4. Making a small change to the Containerfile (changing the id of the secret in the 3rd step):
```
FROM ubuntu:latest

RUN --mount=type=secret,id=mysecret,dst=/home/root/mysecret cat /home/root/mysecret

RUN --mount=type=secret,id=mysecret2,dst=/home/root/mysecret2 cat /home/root/mysecret2
```
and using `buildah bud  --secret=id=mysecret,src=./mysecret.dat  --secret=id=mysecret2,src=./mysecret.dat` to build, succeeds

**Describe the results you received:**
The secret with a particular id cannot be used more than once

**Describe the results you expected:**
The secret should be used as many times as needed

**Output of `rpm -q buildah` or `apt list buildah`:**
Installed from source

**Output of `buildah version`:**

```
Version:         1.24.4
Go Version:      go1.18.1
Image Spec:      1.0.2-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        1.0.0
libcni Version:  v1.0.1
image Version:   5.19.3
Git Commit:      c1bdaba5
Built:           Fri May 13 15:21:02 2022
OS/Arch:         linux/amd64
BuildPlatform:   linux/amd64
```

**Output of `cat /etc/*release`:**

```
NAME="Linux Mint"
VERSION="19 (Tara)"
ID=linuxmint
ID_LIKE=ubuntu
PRETTY_NAME="Linux Mint 19"
VERSION_ID="19"
HOME_URL="https://www.linuxmint.com/"
SUPPORT_URL="https://forums.ubuntu.com/"
[...]

PR #3998, opened 2022-05-17
run: allow re-using secret artifact from host in new `RUN` steps.

flouthoc:
If a secret is used in a `RUN` step it is created on host but we had a
check which expects a secret to not exist on the host however same
secret can be remounted on another step so it can already exists on host
so remove the check since a secret can be mounted again from host in
another `RUN` step.

Closes: https://github.com/containers/buildah/issues/3993


rhatdan:
LGTM
@giuseppe @nalind @TomSweeneyRedHat @ashley-cui @vrothberg @umohnani8 PTAL

PR #3999, opened 2022-05-17
buildah source pull/push: show progress bar

vrothberg:
Render the progress bar on stdout by default.  Add a --quiet/-q option
to re-silence if desired.

@rhatdan PTAL

### DIFF
--- a/docs/buildah-source-pull.1.md
+++ b/docs/buildah-source-pull.1.md
@@ -21,6 +21,10 @@ The [username[:password]] to use to authenticate with the registry if required.
 If one or both values are not supplied, a command line prompt will appear and the
 value can be entered.  The password is entered without echo.
 
+**--quiet**, **-q**
+
+Suppress the progress output when pushing a source image.
+
 **--tls-verify** *bool-value*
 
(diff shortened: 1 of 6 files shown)


docs/buildah-source-pull.1.md
```diff
@@ -21,6 +21,10 @@ The [username[:password]] to use to authenticate with the registry if required.
 If one or both values are not supplied, a command line prompt will appear and the
 value can be entered.  The password is entered without echo.
 
+**--quiet**, **-q**
+
+Suppres the progress output when pulling a source image.
```
edsantiago, line 26: Typo: `suppress` with two final esses. Same below.

edsantiago (commented):
What is the context for this? I'm ultra-uncomfortable with using stdout; would much rather see it on stderr like podman. I'm surprised tests aren't failing.

vrothberg:
> What is the context for this? I'm ultra-uncomfortable with using stdout; would much rather see it on stderr like podman. I'm surprised tests aren't failing.

Thanks for checking.  My heat-fried brain let me believe it bars on stdout but you are absolutely right that they belong in stderr; will update.

PR #4000, opened 2022-05-17
test cleanup

edsantiago:
- remove a bunch of unnecessary clutter (checking $status)
 - add distinct labels to similar-looking tests
 - add check for cpp warning
 - split Containerfile.in test into two parts, one with command-line
   flag and one with envariable, so podman can skip the latter
   in -remote tests while testing the former


flouthoc (approved):
LGTM

edsantiago:
restarted test flake (registry tls timeout; seems to be happening a lot lately)

PR #3767, opened 2022-02-03
Add --keep-ownership option to add/copy commands to keep file ownership.

r10r:
#### What type of PR is this?

> /kind feature

#### What this PR does / why we need it:

This flag is useful if you want to keep file ownership when using `buildah copy` or `buildah add` to copy local 
content into a container.

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

Tests and documentation follow.

#### Does this PR introduce a user-facing change?

```release-note
Added flag `--keep-ownership` to `add` and `copy` commands. If the flag is set the file
ownership is preserved when adding/copying local files to a container.
```

PR #4001, opened 2022-05-17
`ADD|COPY` option `--keep-ownership` 

Dexus:
This is the follow-up PR that includes the initial PR from #3767 and is now fulfilled with additional changes, PRs (openshift/imagebuilder/pull/227) and tests.

Please follow the initial conversations in #3767.

What are the the intention for the PR:

ADD or COPY 
- without `--chown` and `--keep-ownership` result in `root:root` for the files.
- with `--chown 123:456` result in `123:456` for the files.
- with `--keep-ownership` result in `UID:GID` for the files, same as the origin source.

Closes #3767 

⚠️ Notice:

`go.mod` needs to add a replace for `github.com/openshift/imagebuilder` as long openshift/imagebuilder/pull/227 is unmerged
```
go mod edit -replace github.com/openshift/imagebuilder=github.com/Dexus-Forks/imagebuilder
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -45,3 +45,5 @@ require (
 replace github.com/sirupsen/logrus => github.com/sirupsen/logrus v1.4.2
 
 replace github.com/opencontainers/image-spec => github.com/opencontainers/image-spec v1.0.2-0.20211123152302-43a7dee1ec31
+
+replace github.com/openshift/imagebuilder => github.com/Dexus-Forks/imagebuilder v1.2.4-0.20220518131701-b5c6586df000
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -360,7 +360,7 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
 		var copyExcludes []string
 		stripSetuid := false
 		stripSetgid := false
-		preserveOwnership := false
+		preserveOwnership := copy.PreserveOwnership
 		contextDir := s.executor.contextDir
 		if len(copy.From) > 0 {
 			// If from has an argument within it, resolve it to its
@@ -1011,10 +1011,10 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			command := strings.ToUpper(step.Command)
 			// chmod, chown and from flags should have an '=' sign, '--chmod=', '--chown=' or '--from='
 			if command == "COPY" && (flag == "--chmod" || flag == "--chown" || flag == "--from") {
-				return "", nil, errors.Errorf("COPY only supports the --chmod=<permissions> --chown=<uid:gid> and the --from=<image|stage> flags")
+				return "", nil, errors.Errorf("COPY only supports the --chmod, --chown, --from, and --keep-ownership flags")
 			}
 			if command == "ADD" && (flag == "--chmod" || flag == "--chown") {
-				return "", nil, errors.Errorf("ADD only supports the --chmod=<permissions> and the --chown=<uid:gid> flags")
+				return "", nil, errors.Errorf("ADD only supports the --chmod, --chown, and --keep-ownership flags")
 			}
 			if strings.Contains(flag, "--from") && command == "COPY" {
 				arr := strings.Split(flag, "=")
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -1772,7 +1772,7 @@ function _test_http() {
   imgName=alpine-image
   ctrName=alpine-chown
   run_buildah 125 build $WITH_POLICY_JSON --layers -t ${imgName} -f $BUDFILES/copy-chown/Dockerfile.bad $BUDFILES/copy-chown
-  expect_output --substring "COPY only supports the --chmod=<permissions> --chown=<uid:gid> and the --from=<image\|stage> flags"
+  expect_output --substring "COPY only supports the --chmod, --chown, --from, and --keep-ownership flags"
 }
 
 @test "bud with chown copy with unknown substitutions in Dockerfile" {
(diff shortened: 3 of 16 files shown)


tests/bud.bats
```diff
@@ -1894,6 +1902,21 @@ function _test_http() {
...
+  _prefetch busybox
+  target=busybox-derived
+  run_buildah build $WITH_POLICY_JSON -t ${target} -f $BUDFILES/copy-from-keep-ownership/Dockerfile $BUDFILES/copy-from-keep-ownership
+  expect_output --substring "\-rw-r--r--    1 123      456"
+}
+
+@test "bud with copy-workdir and with --keep-ownership" {
+  _prefetch busybox
+  target=busybox-derived
+  chown 321:999 $BUDFILES/copy-from-keep-ownership/file2.txt
```
edsantiago, line 1915: You can't do this: first, the test may run rootless. Second, even if root, you would be chowning a file that you have no right to chown (possibly in the developer's source dir, or possibly in the rpm-managed `/usr/share/tests/buildah`.

Solution: add `skip_if_rootless` to this test, and copy the build context into `$TEST_SCRATCH_DIR`.

tests/bud.bats
```diff
@@ -1894,6 +1902,21 @@ function _test_http() {
...
+  run_buildah build $WITH_POLICY_JSON -t ${target} -f $BUDFILES/copy-from-keep-ownership/Dockerfile2 $BUDFILES/copy-from-keep-ownership
+  expect_output --substring "\-rw-r--r--    1 321      999"
```
edsantiago, line 1917: Whitespace and substring checks are fragile. Also, this will break if the base image ever (for whatever reason) gains those uids/gids. I encourage you to change your containerfiles to `RUN stat -c "%u:%g"` instead, and do an exact (not substring) check.

edsantiago (changes requested):
Thanks for your PR! One showstopper in the tests, one strong encouragement to rework another.

imagebuildah/stage_executor.go
```diff
@@ -907,10 +907,10 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			command := strings.ToUpper(step.Command)
 			// chmod, chown and from flags should have an '=' sign, '--chmod=', '--chown=' or '--from='
 			if command == "COPY" && (flag == "--chmod" || flag == "--chown" || flag == "--from") {
-				return "", nil, errors.Errorf("COPY only supports the --chmod=<permissions> --chown=<uid:gid> and the --from=<image|stage> flags")
+				return "", nil, errors.Errorf("COPY only supports the --keep-ownership --chmod=<permissions> --chown=<uid:gid> and the --from=<image|stage> flags")
```
flouthoc, line 910:
```suggestion
				return "", nil, errors.Errorf("COPY only supports the --keep-ownership, --chmod=<permissions> --chown=<uid:gid> and the --from=<image|stage> flags")
```

imagebuildah/stage_executor.go
```diff
@@ -907,10 +907,10 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
 			}
 			if command == "ADD" && (flag == "--chmod" || flag == "--chown") {
-				return "", nil, errors.Errorf("ADD only supports the --chmod=<permissions> and the --chown=<uid:gid> flags")
+				return "", nil, errors.Errorf("ADD only supports the --keep-ownership --chmod=<permissions> and the --chown=<uid:gid> flags")
```
flouthoc, line 913:
```suggestion
				return "", nil, errors.Errorf("ADD only supports the --keep-ownership, --chmod=<permissions> and the --chown=<uid:gid> flags")
```

imagebuildah/stage_executor.go
Dexus: Should I also add also a Comma after `--chmod=<permissions>` I was thinking about that, but thought I keep the current format.

flouthoc:
@Dexus Commit title for `dc90aed5127822f6b210bf2ed0e2742e98a5e797` is too long could you move PR link in commit title to description please and also please sign the commits in-order to make DCO check happy.

imagebuildah/stage_executor.go
edsantiago: >Should I also add also a Comma after --chmod=<permissions>

Yes, and also after `--chmod=<uid:gid>`. The term is `Oxford comma` and this is an instance where it can help clarity. You can also remove the superfluous `the` and not-really-helpful `=<blah>`, simplifying to:
```
...the --chmod, --chown, --from, and --keep-ownership flags
```
Also, I realize this is a WIP, but you're going to need to squash your commits (to 2, I think) at some point. There's no need nor desire for a stack of "address review feedback" commits.

Thanks again for contributing!

Dexus: Thanks for your feedback, will change all you mentioned points later.

Dexus:
So now all tests are green! 🥳 
So, what happens next?

Dexus:
fresh rebased

flouthoc:
> So now all tests are green! partying_face So, what happens next?

Now we have to wait for this first: https://github.com/openshift/imagebuilder/pull/227 this PR will be rebased when imagebuilder PR is merged.

go.mod
```diff
@@ -45,3 +45,5 @@ require (
 replace github.com/sirupsen/logrus => github.com/sirupsen/logrus v1.4.2
 
 replace github.com/opencontainers/image-spec => github.com/opencontainers/image-spec v1.0.2-0.20211123152302-43a7dee1ec31
+
+replace github.com/openshift/imagebuilder => github.com/Dexus-Forks/imagebuilder v1.2.4-0.20220518131701-b5c6586df000
```
TomSweeneyRedHat, line 49: We'll need to fix this up once the imagebuilder merges and before we merge this PR.

TomSweeneyRedHat:
One nit request.  If you could rename the test `Dockerfile*` files to `Containerfile*`, that would be appreciated.

tests/bud/copy-from-keep-ownership/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM busybox AS basis
+RUN echo hello > /newfile && chown 123:456 /newfile
+FROM basis
+COPY --keep-ownership --from=basis /newfile /newfile_123_456
```
nalind, line 4: The `--keep-ownership` option is redundant when there's a `--from`, so I don't know that we need this test.

Dexus: you are right

flouthoc:
Correct me if i am wrong i think this is on hold because of https://github.com/moby/buildkit/issues/2875 since `dockerclient` does not implements this yet and imagebuilder would end up with `no-op` for `docker` implementation.

Issue #2469, opened 2020-07-14
[FEATURE REQUEST] Skip unneeded stages from multi-stages


grantral:
**Description**

docker/cli#1134

**Steps to reproduce the issue:**
1. `buildah bud --target backend`

**Describe the results you received:**
```
STEP 1: FROM alpine AS frontend_builder
STEP 2: RUN touch frontend
STEP 3: FROM alpine AS frontend
STEP 4: COPY --from=frontend_builder /frontend .
STEP 5: FROM alpine AS backend_builder
STEP 6: RUN touch backend
STEP 7: FROM alpine AS backend
STEP 8: COPY --from=backend_builder /backend .
```

**Describe the results you expected:**
[docker/cli#1134 (comment)](https://github.com/docker/cli/issues/1134#issuecomment-406449342)
```
STEP 1: FROM alpine AS backend_builder
STEP 2: RUN touch backend
STEP 3: FROM alpine AS backend
STEP 4: COPY --from=backend_builder /backend .
```

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.15.0-1.fc32.x86_64
```

**Output of `buildah version`:**

```
Version:         1.15.0
Go Version:      go1.14.3
Image Spec:      1.0.1-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        0.4.0
libcni Version:  
image Version:   5.5.1
Git Commit:      
Built:           Thu Jan  1 06:00:00 1970
OS/Arch:         linux/amd64
```

**Output of `podman version` if reporting a `podman build` issue:**

```
Version:      2.0.2
API Version:  1
Go Version:   go1.14.3
Built:        Thu Jan  1 06:00:00 1970
OS/Arch:      linux/amd64
```

**Output of `cat /etc/*release`:**

```
Fedora release 32 (Thirty Two)
NAME=Fedora
VERSION="32 (Workstation Edition)"
ID=fedora
VERSION_ID=32
VERSION_CODENAME=""
PLATFORM_ID="platform:f32"
PRETTY_NAME="Fedora 32 (Workstation Edition)"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:32"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f32/system-administrators-guide/"
[...]

PR #4004, opened 2022-05-19
buildkit, multistage: skip computing stages which dont contribute to target by checking against `dependencyTree`

flouthoc:
While performing mult-stage builds buildah does not needs to compute
and run each and every stage it can skip stages which are not used by
`target` stage.

Following commit adds `dependencyTree/dependencyMap` to executor which
calculates and populates the dependencyTree for all the stages and marks
if a particular stage is used by `Target/Final` stage directly or
indirectly and unmarked stages can be skipped.

```Dockerfile
## buildah should skip this stage
FROM alpine
RUN echo "unwanted stage"

FROM alpine as one
RUN echo "wanted stage"

## buildah should skip this stage
FROM alpune
RUN echo "another unwanted stage"

FROM one
RUN echo "wanted stage"
```

Closes: https://github.com/containers/buildah/issues/2469

### DIFF
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -510,6 +510,25 @@ func (b *Executor) buildStage(ctx context.Context, cleanupStages map[int]*StageE
 	return imageID, ref, nil
 }
 
+type stageDependencyInfo struct {
+	Name           string
+	Position       int
+	Needs          []string
+	NeededByTarget bool
+}
+
+// Marks `NeededByTarget` as true for the given stage and all its dependency stages as true recursively.
+func markDependencyStagesForTarget(dependencyMap map[string]*stageDependencyInfo, stage string) {
+	if stageDependencyInfo, ok := dependencyMap[stage]; ok {
+		if !stageDependencyInfo.NeededByTarget {
+			stageDependencyInfo.NeededByTarget = true
+			for _, need := range stageDependencyInfo.Needs {
+				markDependencyStagesForTarget(dependencyMap, need)
+			}
+		}
+	}
+}
+
 // Build takes care of the details of running Prepare/Execute/Commit/Delete
 // over each of the one or more parsed Dockerfiles and stages.
 func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (imageID string, ref reference.Canonical, err error) {
@@ -593,10 +612,15 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 		}
 	}()
 
+	// dependencyMap contains dependencyInfo for each stage,
+	// dependencyInfo is used later to mark if a particular
+	// stage is needed by target or not.
+	dependencyMap := make(map[string]*stageDependencyInfo)
 	// Build maps of every named base image and every referenced stage root
 	// filesystem.  Individual stages can use them to determine whether or
 	// not they can skip certain steps near the end of their stages.
 	for stageIndex, stage := range stages {
+		dependencyMap[stage.Name] = &stageDependencyInfo{Name: stage.Name, Position: stage.Position}
 		node := stage.Node // first line
 		for node != nil {  // each line
 			for _, child := range node.Children { // tokens on this line, though we only care about the first
@@ -624,6 +648,16 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 							}
 							b.baseMap[baseWithArg] = true
 							logrus.Debugf("base for stage %d: %q", stageIndex, base)
+							// Check if selected base is not an additional
+							// build context and if base is a valid stage
+							// add it to current stage's dependency tree.
+							if _, ok := b.additionalBuildContexts[baseWithArg]; !ok {
+								if _, ok := dependencyMap[baseWithArg]; ok {
+									// update current stage's dependency info
+									currentStageInfo := dependencyMap[stage.Name]
+									currentStageInfo.Needs = append(currentStageInfo.Needs, baseWithArg)
+								}
+							}
 						}
 					}
 				case "ADD", "COPY":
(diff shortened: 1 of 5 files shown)


flouthoc:
This should also improve build perfomance/speed by skipping stage which are not needed. 

PS: I am expecting tests failures cause some tests might be relying on unwanted stages, will retrofit them.

flouthoc:
> This should also improve build perfomance/speed by skipping stage which are not needed.
> 
> PS: I am expecting tests failures cause some tests might be relying on unwanted stages, will retrofit them.

As expected I need to retrofit some tests. :\

imagebuildah/executor.go
```diff
@@ -636,12 +667,60 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
...
+							// stage.
+							stageName := rootfs
+							// If --from=<index> convert index to name
+							if index, err := strconv.Atoi(stageName); err == nil {
+								stageName = stages[index].Name
+							}
+							// Check if this base is a stage if yes
+							// add base to current stage's dependency tree
+							if _, ok := dependencyMap[stageName]; ok {
+								// upate current stage's dependency info
```
TomSweeneyRedHat, line 681: nit
```suggestion
								// update current stage's dependency info
```

flouthoc:
@nalind Could you PTanotherL. Resolved all the nits.

imagebuildah/executor.go
```diff
@@ -636,12 +670,68 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
...
+							// if following ADD or COPY needs any other
+							// stage.
+							stageName := rootfs
+							// If --from=<index> convert index to name
+							if index, err := strconv.Atoi(stageName); err == nil {
+								stageName = stages[index].Name
+							}
+							// Check if this base is a stage if yes
+							// add base to current stage's dependency tree
+							// and not an additional base context.
```
nalind, line 683: This comment looks like parts of it are in the wrong order.

imagebuildah/executor.go
```diff
@@ -636,12 +670,68 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
...
+							if _, ok := b.additionalBuildContexts[stageName]; !ok {
+								if _, ok := dependencyMap[stageName]; ok {
+									// upate current stage's dependency info
```
nalind, line 686:
```suggestion
									// update current stage's dependency info
```

imagebuildah/executor.go
```diff
@@ -636,12 +670,68 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
...
+								}
+							}
+						}
+					}
+				case "RUN":
+					for _, flag := range child.Flags { // flags for this instruction
+						// We need to populate dependency tree of stages
+						// if it is using `--mount` and `from=` field is set
+						// and `from=` points to a stage consider it in
+						// dependecy calculation.
```
nalind, line 698:
```suggestion
						// dependency calculation.
```

imagebuildah/executor.go
```diff
@@ -636,12 +670,68 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
...
+							}
 						}
 					}
 				}
 			}
 			node = node.Next // next line
 		}
+		// Last stage is always target stage.
+		// Since last/target stage is processed
+		// lets calculate dependency map of stages
```
nalind, line 730:
```suggestion
		// let's calculate dependency map of stages
```

tests/bud.bats
```diff
@@ -202,7 +202,111 @@ _EOF
...
+
+FROM alpine
+COPY --from=one hey .
+RUN cat hey
+RUN echo "target stage"
+_EOF
+
+  run_buildah build $WITH_POLICY_JSON -t source -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile
+  expect_output --substring "needed stage"
+  expect_output --substring "hey"
```
nalind, line 249: What part of the build output is this trying to match?  That string appears multiple times, in instructions in multiple stages.

tests/bud.bats
```diff
@@ -202,7 +202,111 @@ _EOF
...
+
+FROM alpine
+COPY --from=1 hey .
+RUN cat hey
+RUN echo "target stage"
+_EOF
+
+  run_buildah build $WITH_POLICY_JSON -t source -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile
+  expect_output --substring "needed stage"
+  expect_output --substring "hey"
```
nalind, line 277: What part of the build output is this trying to verify? That string appears multiple times, in instructions in multiple stages.

tests/bud.bats
```diff
@@ -202,7 +202,111 @@ _EOF
...
+RUN echo "another unwanted stage"
+
+FROM alpine
+RUN --mount=type=bind,from=one,target=/test cat /test/hey
+RUN echo "target stage"
+_EOF
+
+  run_buildah build $WITH_POLICY_JSON -t source -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile
+  expect_output --substring "needed stage"
+  expect_output --substring "hey"
```
nalind, line 304: What part of the build output is this trying to check for? That string appears multiple times, in instructions in multiple stages.

nalind:
Some nits about comments, questions in a couple of tests, and this should probably have some tests that ensure that the right things happen when we give additional build contexts the same names as Dockerfile stages.

tests/bud.bats
flouthoc: Modified the tests so string only appears in one stage. It tests that we can use a file from previous stage in a target stage and it is not skipped.

tests/bud.bats
flouthoc: Modified the tests so string only appears in one stage. It tests that we can use a file from previous stage in a target stage and it is not skipped.

flouthoc:
> Some nits about comments, questions in a couple of tests, and this should probably have some tests that ensure that the right things happen when we give additional build contexts the same names as Dockerfile stages.

Added a new test `build-test skipping unwanted stages with FROM and conflict with additional build context` which should cover this case.

flouthoc:
Will rebase this once get we could get LGTM ack. @nalind Could you PTAL again.

tests/bud/with-arg/Dockerfilefromarg
```diff
@@ -4,13 +4,17 @@ ARG another_app_type_default=m
...
 
 # Do not supply this in cli, lets use default
 FROM ${another_app_type_default} as final
 RUN echo hello
+COPY --from=m hello .
 
-FROM ${another_app_type} as final
+FROM ${another_app_type} as target
 RUN echo hello
+COPY --from=final hello .
```
nalind, line 20: Why did we need to change this file to get the test to pass?  Why did the builder not figure out that the final stage needed the second stage?

nalind (changes requested):
A few things I'd like to see changed in the tests, and a question about one test.

tests/bud.bats
```diff
@@ -222,8 +222,143 @@ _EOF
...
+
+  run_buildah build $WITH_POLICY_JSON -t source -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile ${TEST_SCRATCH_DIR}/bud/platform
+  expect_output --substring "needed stage"
+  expect_output --substring "something"
+  expect_output --substring "target stage"
+  assert "$output" !~ "unwanted stage"
+}
 
-# Test pinning image using additional build context
+# Test pinning image using addtional build context
```
nalind, line 361:
```suggestion
# Test pinning image using additional build context
```

nalind:
Other than the typo being added to a comment in bud.bats, LGTM.

flouthoc:
> Other than the typo being added to a comment in bud.bats, LGTM.

Fixed. @rhatdan @TomSweeneyRedHat PTAL

flouthoc:
I'll vendor this into podman to verify if we need to retrofit any podman tests as well which might be relying on stages which will be skipped now if that's the case this will cause hiccups in vendor. I'll amend podman tests if i notice any such case.

cc @edsantiago just for info in-case someone vendor buildah before me.

edsantiago:
@flouthoc the treadmill PR is green #13808

PR #4005, opened 2022-05-19
Test updating c/image and GitHub.com/proglottis/gpgme

mtrmac:
#### What type of PR is this?

/kind cleanup

GitHub.com/proglottis/gpgme now uses pkg-config to find gpgme. That’s cleaner than the previous method, but test that Buildah builds with that, just to avoid surprises (and this PR should fix them if there turn out to be any.)

#### How to verify it

Existing tests

None

```release-note
Builds with GPGME now require GPGME ≥ 1.13.0.
```


mtrmac:
Tests passed with no modifications needed, closing. This will get into Buildah naturally with a future re-vendoring.

PR #4013, opened 2022-05-23
squash, layers: never use build cache when computing last step of last stage

flouthoc:
When using `--squash` with `--layes` last step of a squashed build produces 
a layer which contains contents from other layers as well. So last step of last 
stage should be always computed and never fetched from build cache, while 
other steps and stages don't have this restriction.

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -969,14 +969,14 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			// squash the contents of the base image.  Whichever is
 			// the case, we need to commit() to create a new image.
 			logCommit(s.output, -1)
-			if imgID, ref, err = s.commit(ctx, s.getCreatedBy(nil, ""), false, s.output); err != nil {
+			if imgID, ref, err = s.commit(ctx, s.getCreatedBy(nil, ""), false, s.output, s.executor.squash); err != nil {
 				return "", nil, errors.Wrapf(err, "error committing base container")
 			}
 		} else if len(s.executor.labels) > 0 || len(s.executor.annotations) > 0 {
 			// The image would be modified by the labels passed
 			// via the command line, so we need to commit.
 			logCommit(s.output, -1)
-			if imgID, ref, err = s.commit(ctx, s.getCreatedBy(stage.Node, ""), true, s.output); err != nil {
+			if imgID, ref, err = s.commit(ctx, s.getCreatedBy(stage.Node, ""), true, s.output, s.executor.squash); err != nil {
 				return "", nil, err
 			}
 		} else {
@@ -1107,7 +1107,7 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 				// stage.
 				if lastStage || imageIsUsedLater {
 					logCommit(s.output, i)
-					imgID, ref, err = s.commit(ctx, s.getCreatedBy(node, addedContentSummary), false, s.output)
+					imgID, ref, err = s.commit(ctx, s.getCreatedBy(node, addedContentSummary), false, s.output, s.executor.squash)
 					if err != nil {
 						return "", nil, errors.Wrapf(err, "error committing container for step %+v", *step)
 					}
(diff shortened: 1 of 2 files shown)


tests/squash.bats
```diff
@@ -132,6 +132,8 @@ function check_lengths() {
   _prefetch alpine
   # populate cache from simple build
   run_buildah build --layers -t test $WITH_POLICY_JSON -f $BUDFILES/layers-squash/Dockerfile.multi-stage
+  run_buildah inspect -t image -f '{{len .Docker.RootFS.DiffIDs}}' test
+  expect_output "1" "should only container 1 diff"
```
nalind, line 136:
```suggestion
  expect_output "1" "image should only include 1 layer"
```
Why does this test pass if the image was built with multiple layers, with no squashing requested, producing a final image that contained the layers from its base image, and layers for each of two build steps?

imagebuildah/stage_executor.go
```diff
@@ -1214,7 +1214,7 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 				}
 			}
 		} else {
-			if s.executor.squash {
+			if s.executor.squash || !squashToggle {
```
nalind, line 1217: A stage executor method should _not_ be modifying the executor that owns it, since that executor can be executing multiple stages at the same time.

tests/squash.bats
```diff
@@ -141,3 +143,30 @@ function check_lengths() {
   run_buildah rmi -f testsquash
   run_buildah rmi -f test
 }
+
+# Test build with --squash and --layers and verify number of layers and content inside image
+@test "bud-squash-should-use-cache and verify content inside image" {
+  mkdir -p ${TEST_SCRATCH_DIR}/bud/platform
+
+  cat > ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile << _EOF
+FROM quay.io/libpod/busybox:latest
```
nalind, line 152:
```suggestion
FROM busybox
```
The `registries.conf` the test uses sets this up as an alias, so being more specific here doesn't provide a benefit.

tests/squash.bats
```diff
@@ -141,3 +143,30 @@ function check_lengths() {
...
+RUN touch hello
+ADD . /data
+RUN echo hey && mkdir water
+_EOF
+
+  # Build a first image with --layers and --squash and populate build cache
+  run_buildah build $WITH_POLICY_JSON --squash --layers -t one -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile ${TEST_SCRATCH_DIR}/bud/platform
+  run_buildah inspect -t image -f '{{len .Docker.RootFS.DiffIDs}}' one
+  expect_output "1" "should only container 1 diff"
```
nalind, line 161:
```suggestion
  expect_output "1" "image built with --squash should only include 1 layer"
```

tests/squash.bats
```diff
@@ -141,3 +143,30 @@ function check_lengths() {
...
+  # Build again and verify if cache is being used
+  run_buildah build $WITH_POLICY_JSON --squash --layers -t two -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile ${TEST_SCRATCH_DIR}/bud/platform
+  expect_output --substring "Using cache"
+  run_buildah inspect -t image -f '{{len .Docker.RootFS.DiffIDs}}' two
+  expect_output "1" "should only container 1 diff"
```
nalind, line 166:
```suggestion
  expect_output "1" "image built with --squash should only include 1 layer"
```

nalind (changes requested):

rhatdan:
LGTM

imagebuildah/stage_executor.go
flouthoc: Yes simplified the logic.

imagebuildah/stage_executor.go
```diff
@@ -1214,30 +1210,27 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
-				// squashToggle flag allows executor to treat other instruction and layers
-				// as regular builds and only perform squashing at last
-				s.executor.squash = false
-				squashToggle = true
-			}
+			// At this point we want to save history for other layers
+			// even if its a squashed build. So treat other instruction
+			// and layers as regular builds and only perform squashing
+			// at last.
+			doSquash := false
```
nalind, line 1217: It doesn't seem that we even need this variable.

imagebuildah/stage_executor.go
```diff
@@ -1214,30 +1210,27 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
 			logCommit(s.output, i)
-			imgID, ref, err = s.commit(ctx, s.getCreatedBy(node, addedContentSummary), !s.stepRequiresLayer(step), commitName)
+			imgID, ref, err = s.commit(ctx, s.getCreatedBy(node, addedContentSummary), !s.stepRequiresLayer(step), commitName, doSquash)
 			if err != nil {
 				return "", nil, errors.Wrapf(err, "error committing container for step %+v", *step)
 			}
 		}
 
 		// Perform final squash for this build as we are one the,
 		// last instruction of last stage
```
nalind, line 1231:
```suggestion
		// last instruction of the last stage
```

nalind:
Comment nits, otherwise LGTM.  Could use a rebase and potentially a... squash.

Issue #4002, opened 2022-05-17
podman build cannot mount devices, while podman run can.


hjmodi:
**Description**
Cannot attach devices with `podman build --device /dev/null:/var/tmp/flatpak-build/dev/null --no-cache .` . Getting error, `Error: error creating build executor: Renaming device /dev/null to /var/tmp/flatpak-build/dev/null is not supported in rootless containers` . However the same works with `podman run --device /dev/null:/var/tmp/flatpak-build/dev/null -ti --rm ubi8` giving the following result,

```
Resolved "ubi8" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull registry.access.redhat.com/ubi8:latest...
Getting image source signatures
Checking if image destination supports signatures
Copying blob f70d60810c69 done  
Copying blob 545277d80005 done  
Copying config 1264065f6a done  
Writing manifest to image destination
Storing signatures
[root@f0398eb06c06 /]# ls /var/tmp/flatpak-build/dev/
null
[root@f0398eb06c06 /]# exit
```

**Steps to reproduce the issue:**
1. Run podman build and try to mount a device. i.e. `--device /dev/null:/var/tmp/flatpak-build/dev/null`
2.
3.

**Describe the results you received:**
Failed with `Error: error creating build executor: Renaming device /dev/null to /var/tmp/flatpak-build/dev/null is not supported in rootless containers`

**Describe the results you expected:**
To be able to mount device to a given location.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
package buildah is not installed
```

**Output of `buildah version`:**

```
bash: buildah: command not found...
Install package 'buildah' to provide command 'buildah'? [N/y] n
```

**Output of `podman version` if reporting a `podman build` issue:**

```
Client:       Podman Engine
Version:      4.1.0
API Version:  4.1.0
Go Version:   go1.18
Built:        Fri May  6 12:15:54 2022
OS/Arch:      linux/amd64
```

**Output of `cat /etc/*release`:**

```
Fedora release 36 (Thirty Six)
NAME="Fedora Linux"
VERSION="36 (Workstation Edition)"
ID=fedora
VERSION_ID=36
VERSION_CODENAME=""
PLATFORM_ID="platform:f36"
PRETTY_NAME="Fedora Linux 36 (Workstation Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
[...]

PR #4014, opened 2022-05-24
buildah: add support for renaming a device in rootless setups using `--device`

flouthoc:
Buildah now supports renaming devices while performing a build using
`--device <some-name>:<new-name>`. Implementation is similar to `podman`
where we prefer using `bind-mount` for devices instead of `mknod` in
`rootless` setups.

Usage
```console
buildah build -t test --device /dev/null:/test/dev/null .
```

Closes: https://github.com/containers/buildah/issues/4002

### DIFF
--- a/define/types_unix.go
+++ b/define/types_unix.go
@@ -6,4 +6,13 @@ import (
 	"github.com/opencontainers/runc/libcontainer/devices"
 )
 
-type ContainerDevices = []devices.Device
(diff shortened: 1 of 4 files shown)


TomSweeneyRedHat:
LGTM
but a rebase is needed

flouthoc:
Rebased :)

PR #4017, opened 2022-05-25
Move options parsing out of build.go and into pkg/cli

rhatdan:
Attempting to share common code for CLI parsing between
buildah build and podman build

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```


TomSweeneyRedHat:
deadcode warning from the tests @rhatdan

pkg/cli/build.go
```diff
@@ -0,0 +1,416 @@
...
+		TransientMounts:         iopts.Volumes,
+		OciDecryptConfig:        decryptConfig,
+		Jobs:                    &iopts.Jobs,
+		Excludes:                excludes,
+		Timestamp:               timestamp,
+		Platforms:               platforms,
+		UnsetEnvs:               iopts.UnsetEnvs,
+		Envs:                    iopts.Envs,
+		OSFeatures:              iopts.OSFeatures,
+		OSVersion:               iopts.OSVersion,
```
TomSweeneyRedHat, line 352: If we're going to make this massive change that can't be reviewed easily anyway, I'd like to alphabetize these fields in here.  @nalind may have other opinions.

pkg/cli/build.go
```diff
@@ -0,0 +1,416 @@
...
+		}
+		cc := encconfig.CombineCryptoConfigs([]encconfig.CryptoConfig{dcc})
+		decryptConfig = cc.DecryptConfig
+	}
+
+	return decryptConfig, nil
+}
+
+// EncryptConfig translates encryptionKeys into a EncriptionsConfig structure
+func EncryptConfig(encryptionKeys []string, encryptLayers []int) (*encconfig.EncryptConfig, *[]int, error) {
```
TomSweeneyRedHat, line 401: You dropped the "get" for this and the prior function and made the availability wider, was that intentional?

rhatdan: Yes `Get` is not considered good form in Golang, as I understand it. The functions need to be external for other buildah commands in cmd/buildah/*.go

pkg/cli/build.go
rhatdan: I think we can alphabetize since we break git split now.

rhatdan:
@nalind @flouthoc @vrothberg @giuseppe PTAL

cmd/buildah/build.go
```diff
@@ -88,342 +69,34 @@ func init() {
...
-	if iopts.Quiet {
-		options.ReportWriter = ioutil.Discard
-	}
-
 	id, ref, err := imagebuildah.BuildDockerfiles(getContext(), store, options, containerfiles...)
 	if err == nil && options.Manifest != "" {
 		logrus.Debugf("manifest list id = %q, ref = %q", id, ref.String())
 	}
+	for _, f := range removeAll {
+		os.RemoveAll(f)
```
flouthoc, line 99: Could use a `defer` here instead i think this entire loop should go just after we fetch `removeAll` in a `defer` call so we cleanup always otherwise if we return early because of `err` we will miss cleaning these files.

pkg/cli/build_test.go
```diff
@@ -0,0 +1,28 @@
...
+	"testing"
+
+	"github.com/containers/buildah/define"
+)
+
+func TestGetFormat(t *testing.T) {
+	_, err := GetFormat("bogus")
+	if err == nil {
+		t.Error("GetFormat should have failed.")
+	}
```
nalind, line 13: Using `github.com/stretchr/testify/assert.NotNil()` and friends in here would shorten this up.

pkg/cli/build_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+func TestGetFormat(t *testing.T) {
+	_, err := GetFormat("bogus")
+	assert.NotNil(t, err)
+
+	format, err := GetFormat("oci")
+	assert.Nil(t, err)
+	if format != define.OCIv1ImageManifest {
+		t.Errorf("expected oci format but got %v.", format)
+	}
```
nalind, line 18:
```suggestion
	assert.Equalf(t, define.OCIv1ImageManifest, format, "expected oci format but got %v.", format)
```

cmd/buildah/build.go
```diff
@@ -88,347 +69,34 @@ func init() {
...
-		pullPolicy = define.PullIfNewer
-	}
-	if iopts.PullAlways || strings.EqualFold(strings.TrimSpace(iopts.Pull), "always") {
-		pullPolicy = define.PullAlways
-	}
-	if iopts.PullNever || strings.EqualFold(strings.TrimSpace(iopts.Pull), "never") {
-		pullPolicy = define.PullNever
-	}
-	logrus.Debugf("Pull Policy for pull [%v]", pullPolicy)
+func buildCmd(c *cobra.Command, inputArgs []string, iopts buildahcli.BuildOptions) error {
```
nalind, line 73:
```suggestion
```

pkg/cli/build.go
nalind: Some or all of these new public functions and constants could conceivably be moved into an `internal` folder so that we don't export them outside of buildah's source tree.

nalind:
This looks like it'll work.

flouthoc (approved):
LGTM

PR #4020, opened 2022-05-26
Fix use generic/ambiguous DEBUG name

cevich:
#### What type of PR is this?

/kind cleanup

The use of the generic 'DEBUG' name in varialbes or flags can be very
problematic and lead to difficult to troubleshoot problems.  Update CI
VM images to those which no-longer react to the `$DEBUG` env. var.  Also
update other repository files which make use of the generic form.

#### How to verify it

CI testing and manual review of documentation.

None

I'm not fully aware of side-effects / consequences of the `Makefile` or `tests/bud/containerfile/Containerfile.in` changes.

```release-note
None
```


.cirrus.yml
```diff
@@ -31,7 +31,7 @@ env:
     PRIOR_FEDORA_NAME: "fedora-35"
     UBUNTU_NAME: "ubuntu-2110"
 
-    IMAGE_SUFFIX: "c4955393725038592"
+    IMAGE_SUFFIX: "c6211193021923328"
```
edsantiago, line 34: I'm guessing this was accidental: it's not mentioned in the commit message, and CI is failing. I'll take another look after this is fixed.

edsantiago (changes requested):

.cirrus.yml
edsantiago: Ah, okay; then when you do your final commit, you may want to update the commit message to specify that. Otherwise there's no indication of why the underlying images changed, and that's usually a big change that warrants a big notice.

cevich: Ya, I agree.  I did mention "Update CI VM images to those which no-longer react to the $DEBUG env. var.".  Hmm, maybe  I should mention the "automation script library" explicitly.

cevich: Message updated.

cevich:
Problem with missing `bats` tracked down to a bug in an image build script.  Fix: https://github.com/containers/automation_images/pull/136

cevich:
Assuming tests run clean, this should be ready to go.

PR #4022, opened 2022-05-27
Fix TODO: de-spaghettify run mounts

ashley-cui:
Code Cleanup: Buildkit run mount setup functions no longer have 13
arguments and are slightly more readable. Use structs instead.

[NO NEW TESTS NEEDED]

Signed-off-by: Ashley Cui <acui@redhat.com>

#### What type of PR is this?
 /kind cleanup

None

```release-note
None
```


run_linux.go
```diff
@@ -2829,25 +2840,27 @@ func (b *Builder) getSSHMount(tokens []string, count int, sshsources map[string]
...
+	if err := label.Relabel(hostSock, b.MountLabel, false); err != nil {
 		if shutdownErr := fwdAgent.Shutdown(); shutdownErr != nil {
 			b.Logger.Errorf("error shutting down agent: %v", shutdownErr)
 		}
 		return nil, nil, err
 	}
-
-	hostUID, hostGID, err := util.GetHostIDs(uidmap, gidmap, uid, gid)
+	hostUID, hostGID, err := util.GetHostIDs(idMaps.uidmap, idMaps.gidmap, uid, gid)
+	if err != nil {
```
flouthoc, line 2861: I think this `err !=nil` block is not needed since we already handle `err` in the line below and perform shutdown of ssh-agent if it was created, following line will override the old logic.

run.go
```diff
@@ -178,3 +171,27 @@ type runMountArtifacts struct {
 	// LockedTargets to be unlocked if there are any.
 	LockedTargets []string
 }
+
+// RunMountSources are the availible run mounts for this run
+type RunMountSources struct {
```
flouthoc, line 176: Nit, not-blocker: I think as of now it holds  attributes other than sources as well, could we call it some thing like `RunMountInfo` or `RunMountAttributes` so it becomes more generic. But lets wait for others to confirm.

run_linux.go
```diff
@@ -2721,6 +2728,10 @@ func getSecretMount(tokens []string, secrets map[string]define.Secret, mountlabe
 		envFile = tmpFile.Name()
 		ctrFileOnHost = tmpFile.Name()
 	case "file":
+		containerWorkingDir, err := b.store.ContainerDirectory(b.ContainerID)
```
flouthoc, line 2731: Since we already fetch containerWorkingDir at `setupMounts(` I think this could be part of `RunMountSources` if we rename it to `RunMountInfo` or `RunMountAttributes`.

run.go
```diff
@@ -144,21 +143,15 @@ type RunOptions struct {
 	DropCapabilities []string
 	// Devices are the additional devices to add to the containers
 	Devices define.ContainerDevices
-	// Secrets are the available secrets to use in a RUN
-	Secrets map[string]define.Secret
```
flouthoc, line 148: Not sure but I think this is breaking the public API for `RunOptions` instead of removing could we mark these fields as deprecated instead. @nalind Could confirm this.

run_linux.go
ashley-cui: I'd prefer to calculate this here, since the `containerWorkingDir` isn't an exclusive attribute to run mounts. I feel like it would be cleaner to only put fields that only deal with run mounts in `RunMountSources`, but if anyone else feels strongly about this I'd be willing to reconsider.

run.go
ashley-cui: Let me know if it does break the public API! If it does we can just set `RunOptions.RunMountSources.Secrets = RunOptions.Secrets` or something like that?

imagebuildah/stage_executor.go
```diff
@@ -608,19 +609,21 @@ func (s *StageExecutor) Run(run imagebuilder.Run, config docker.Config) error {
...
 		RunMounts:        run.Mounts,
-		StageMountPoints: stageMountPoints,
-		SystemContext:    s.executor.systemContext,
+		RunMountSources: buildah.RunMountSources{
+			SystemContext:    s.executor.systemContext,
+			StageMountPoints: stageMountPoints,
+			Secrets:          s.executor.secrets,
+			SSHSources:       s.executor.sshsources,
+			ContextDir:       s.executor.contextDir,
+		},
```
TomSweeneyRedHat, line 626: Since we're doing a big change anywho, could we alphabetize these?

run.go
TomSweeneyRedHat: I'd lean towards RunMountInfo fwiw.

cmd/buildah/run.go
```diff
@@ -137,14 +137,14 @@ func runCmd(c *cobra.Command, args []string, iopts runInputOptions) error {
...
 		ConfigureNetwork: networkPolicy,
-		ContextDir:       iopts.contextDir,
 		CNIPluginPath:    iopts.CNIPlugInPath,
 		CNIConfigDir:     iopts.CNIConfigDir,
 		AddCapabilities:  iopts.capAdd,
 		DropCapabilities: iopts.capDrop,
 		Env:              iopts.env,
 		WorkingDir:       iopts.workingDir,
 	}
+	options.RunMountSources.ContextDir = iopts.contextDir
```
nalind, line 147: Go ahead and include this in the initializer, as the hunk for `imagebuildah/stage_executor.go` does.

run.go
```diff
@@ -178,3 +171,27 @@ type runMountArtifacts struct {
 	// LockedTargets to be unlocked if there are any.
 	LockedTargets []string
 }
+
+// RunMountSources are the availible run mounts for this run
```
nalind, line 175:
```suggestion
// RunMountSources are the available run mounts for this run
```

run_linux.go
```diff
@@ -1875,13 +1882,13 @@ func (b *Builder) runSetupVolumeMounts(mountLabel string, volumeMounts []string,
...
 
-			contentDir, err := overlay.TempDir(containerDir, rootUID, rootGID)
+			contentDir, err := overlay.TempDir(containerDir, idMaps.rootUID, idMaps.rootGID)
 			if err != nil {
 				return specs.Mount{}, errors.Wrapf(err, "failed to create TempDir in the %s directory", containerDir)
 			}
 
-			overlayOpts := overlay.Options{RootUID: rootUID,
-				RootGID:                rootGID,
+			overlayOpts := overlay.Options{RootUID: idMaps.rootUID,
```
nalind, line 1890:
```suggestion
			overlayOpts := overlay.Options{
				RootUID: idMaps.rootUID,
```

run.go
ashley-cui: @nalind What's the deprecation protocol? Do we set `RunOptions.RunMountSources.Secrets = RunOptions.Secrets` in `func (b *Builder) Run(command []string, options RunOptions)` in run_linux.go, or do we mark as deprecated and simply ignore the field?

nalind: How does the caller benefit from our moving a handful of fields into a nested structure?  Is the disruption to any code that calls our `Run()` method worth it?  Would gathering a subset of `RunOptions` into a `RunMountSources` to pass down to `setupMounts()` right before we call it be worse?

ashley-cui: > Would gathering a subset of `RunOptions` into a `RunMountSources` to pass down to `setupMounts()` right before we call it be worse?
I could do that instead! So we leave the SSH, Secrets, etc. in their original place in RunOptions, and inside `Run(command []string, options RunOptions)`, just put those into a RunMountSources struct and pass that into setupMounts()

I'll fix that up and re-push :) Thanks!

ashley-cui:
Re-pushed with no more breaking changes!

flouthoc (approved):
LGTM

Issue #4025, opened 2022-05-30
`nixery.dev` style images with non-sequential layers break `buildah bud`


SohumB:
**Description**

```shell
$ cat Dockerfile
FROM nixery.dev/shell
RUN ls
$ buildah bud .
STEP 1/2: FROM nixery.dev/shell
STEP 2/2: RUN ls
bin
dev
etc
lib
libexec
nix
nix-support
proc
run
share
sys
tmp
usr
COMMIT
error committing container for step {Env:[SSL_CERT_FILE=/etc/ssl/certs/ca-bundle.crt PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin] Command:run Args:[ls] Flags:[] Attrs:map[] Message:RUN ls Original:RUN ls}: error copying layers and metadata for container "b2dfabd70db97640d06962855cf9bef4cf618c883cf005b495c2e86904792f67": initializing source containers-storage:shell-working-container: internal error: history lists 1 non-empty layers, but we have 10 layers on disk
```

I believe this is how `nixery.dev` works: the layers it packages together in one manifest do not follow the standard sequential pattern. That seems to be what's confusing buildah here.

```shell
$ apt list buildah
Listing... Done
buildah/experimental,now 1.26.1+ds1-1 amd64 [installed]
buildah/testing,unstable 1.23.1+ds1-2+b1 i386
$ buildah version
Version:         1.26.1
Go Version:      go1.18.2
Image Spec:      1.0.2-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        1.0.0
libcni Version:  
image Version:   5.21.1
Git Commit:      
Built:           Thu Jan  1 10:00:00 1970
OS/Arch:         linux/amd64
BuildPlatform:   linux/amd64
$ cat /etc/*release
PRETTY_NAME="Debian GNU/Linux bookworm/sid"
NAME="Debian GNU/Linux"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
$ uname -a
Linux dynast 5.17.0-3-amd64 #1 SMP PREEMPT Debian 5.17.11-1 (2022-05-26) x86_64 GNU/Linux
$ cat ~/.config/containers/storage.conf
[storage]
driver = "overlay"
runroot = "/run/user/1000"
graphroot = "/home/sohum/.local/share/containers/storage"
[...]

Issue #3513, opened 2021-09-12
bud: permit turning off the history


dilyanpalauzov:
`buildah config --add-history` turns the history for the resulting image explicitly on.  It can be configured also with the BUILDAH_HISTORY environment variable.

* add option to buildah bud to disable history

````
$ echo -e "FROM scratch\nENV PATH=/abc\nCOPY . ." > Containerfile
$ BUILDAH_HISTORY=false buildah bud
STEP 1/3: FROM scratch
STEP 2/3: ENV PATH=/abc
STEP 3/3: COPY . .
COMMIT
Getting image source signatures
Copying blob 52907c03eaf0 done  
Copying config 80564cf267 done  
Writing manifest to image destination
Storing signatures
--> 80564cf2674
80564cf26744eb11868c1606a82d236cb19c5dcaa5b1d19bf59819b51a3ff7e2
$ skopeo copy containers-storage:80564cf2674 oci-archive:zzz.tar
Getting image source signatures
Copying blob 52907c03eaf0 done  
Copying config 80564cf267 done  
Writing manifest to image destination
Storing signatures
```
zzz.tar:blobs/sha256/80564cf26744eb11868c1606a82d236cb19c5dcaa5b1d19bf59819b51a3ff7e2 contains COPY

> {"created":"2021-09-12T10:49:17.618464676Z","architecture":"amd64","os":"linux","config":{"Env":["PATH=/abc"],"Labels":{"io.buildah.version":"1.22.3"}},"rootfs":{"type":"layers","diff_ids":["sha256:52907c03eaf0202b750160248a8e38b0a1c6cf825164d0f11d048cee6f9f0c28"]},"history":[{"created":"2021-09-12T10:49:17.466221656Z","created_by":"/bin/sh -c #(nop) ENV PATH=/abc","empty_layer":true},{"created":"2021-09-12T10:49:17.619307251Z","created_by":"/bin/sh -c #(nop) COPY dir:199527f3540779c2c9f1ffb5db2bddd6ec5a3c15cc56eb552802b5fdb3b72f13 in . "}]}

→ The history contains `COPY`.
```
$ buildah images --history --json
[
    {
        "id": "80564cf26744",
        "names": null,
        "digest": "sha256:395a16ac21cac5009428f2932fc2c6200fdab8d901a528bf495b372dbea6518c",
        "createdat": "2 minutes ago",
        "size": "3.27 KB",
        "createdatraw": "2021-09-12T10:49:17.618464676Z",
[...]

PR #4028, opened 2022-06-01
build, commit: add support for disabling image history with `--add-history`

flouthoc:
As per `OCI` `image-spec` setting history is optional so buildah must
support these use-cases via `--add-history=false` some of the use-cases
are discussed below.

* There are use-cases when image-spec contributes to the image size by
  adding build history to image-spec in certain use-cases this image
history is not required by the end-users but there is no way to remove
it. See: https://github.com/containers/buildah/issues/3513

* Certain build tools don't honor setting build history in image spec
  however buildah does not allow processing images without build history
since setting build history is still optional as per `OCI` image-spec
( https://github.com/opencontainers/image-spec/blob/main/config.md#properties)
so buildah must support such use-cases.

Closes: https://github.com/containers/buildah/issues/4025
Closes: https://github.com/containers/buildah/issues/3513


flouthoc:
@nalind PTAL this should close both the use-cases of working with build tools which ignores setting build history and also disabling build history for cases where users don't need it.

cmd/buildah/commit.go
```diff
@@ -23,6 +23,7 @@ import (
 
 type commitInputOptions struct {
 	authfile           string
+	addHistory         bool
```
rhatdan, line 26: Can you make this disableHistory for consistency or stick to add-history throughout.

cmd/buildah/commit.go
```diff
@@ -108,6 +109,7 @@ func commitListFlagSet(cmd *cobra.Command, opts *commitInputOptions) {
 		panic(fmt.Sprintf("error marking reference-time as hidden: %v", err))
 	}
 
+	flags.BoolVar(&opts.addHistory, "add-history", true, "add history for build layers (default true)")
```
rhatdan, line 112: Can you make this disableHistory for consistency or stick to add-history throughout.

rhatdan (approved):

cmd/buildah/commit.go
TomSweeneyRedHat: +1

docs/buildah-build.1.md
```diff
@@ -28,6 +28,15 @@ When a Git repository is set as the URL, the repository is cloned locally and th
 
 ## OPTIONS
 
+**--add-history** *bool-value*
+
+Adds build history for layers into the image spec. (default true).
+
+Following option is useful for the cases where end users explicitly
```
nalind, line 35:
```suggestion
This option is useful for the cases where end users explicitly
```

docs/buildah-build.1.md
```diff
@@ -28,6 +28,15 @@ When a Git repository is set as the URL, the repository is cloned locally and th
...
+wants to set `--add-history=false` to remove optional `History` from
```
nalind, line 36:
```suggestion
want to set `--add-history=false` to omit the optional `History` from
```

docs/buildah-build.1.md
```diff
@@ -28,6 +28,15 @@ When a Git repository is set as the URL, the repository is cloned locally and th
...
+image-spec or when working with images built from build tools that
+does not honors setting `History` in image-spec.
```
nalind, line 38:
```suggestion
do not include `History` information in their images.
```

cmd/buildah/commit.go
flouthoc: Made this `--add-history` throughout, since we discussed this before here: https://github.com/containers/buildah/issues/3513#issuecomment-975899755

cmd/buildah/commit.go
flouthoc: Made this --add-history throughout, since we discussed this before here: https://github.com/containers/buildah/issues/3513#issuecomment-975899755

flouthoc:
> Is this supposed to be preventing the builder from adding history entries to those inherited from the base image, or clearing them as well? The docs suggest one thing, but the test expects the other, and I don't know which one is right.

Updated the docs with the suggestions that should fix it and makes it more clear i.e `--add-history=false` would mean to omit the optional `History` field from `image-spec`.

flouthoc:

image.go
```diff
@@ -530,43 +531,56 @@ func (i *containerImageRef) NewImageSource(ctx context.Context, sc *types.System
...
 	}
-	expectedDiffIDs = expectedDockerDiffIDs(dimage)
-	if len(dimage.RootFS.DiffIDs) != expectedDiffIDs {
-		return nil, errors.Errorf("internal error: history lists %d non-empty layers, but we have %d layers on disk", expectedDiffIDs, len(dimage.RootFS.DiffIDs))
+
+	// Sanity check that we didn't just create a mismatch between non-empty layers in the
+	// history and the number of diffIDs. Following sanity check is ignored if build history
+	// is disabled explicitly by the user.
+	// Disable sanity check when baseImageHistory is null for docker parity
+	if i.addHistory != types.OptionalBoolFalse && (baseImageHistoryLen !=0) {
```
rhatdan, line 575: Can you merge this section into previous block, whereyou already checked i.addHistory != types.OptionalBoolFalse

pkg/cli/common.go
```diff
@@ -239,6 +240,7 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
 		panic(fmt.Sprintf("error marking the pull-never flag as hidden: %v", err))
 	}
 	fs.BoolVarP(&flags.Quiet, "quiet", "q", false, "refrain from announcing build instructions and image read/write progress")
+	fs.BoolVar(&flags.AddHistory, "add-history", true, "add history for build layers (default true)")
```
rhatdan, line 243: (default true) here I believe will cause a stutter. if you do buildah build --help.  Please remove it as well as the line below 
--identity already stutters.

```
buildah build --help
...
      --identity-label                add default identity label (default true) (default true)
```

flouthoc:
Added another commit to remove stutter from `--identity-label` and `--add-history` itself.

rhatdan:
LGTM

nalind:
I personally still think that a defaults-to-false "--omit-history" is easier to explain, but this works.

flouthoc:
> I personally still think that a defaults-to-false "--omit-history" is easier to explain, but this works.

@nalind We can still flip it no rush, just tried to keep it what was discussed here https://github.com/containers/buildah/issues/3513#issuecomment-975899755 since `config` has a similar option. But we can still flip it its a quick change.

nalind:
In `buildah config` it affects the change being made for that command, and defaults to false, while this similarly-named option has the opposite default, which can get confusing.

rhatdan:
Ok lets go with --omit-history

nalind:
If I'm reading the logic right, if we start with a base image that has no history but has layers, we create a new image that has one history entry in it and multiple layers, having bypassed the logic that tries to verify that the number of non-"empty-layer" history entries in the image we're about to create matches the number of layers the new image will have.  If we then try to use that new image as a base image for another build, that logic won't be bypassed by default because that new image has some history information, and we'll get an error.

flouthoc:
@nalind PTAL amended to `--omit-history`

> If I'm reading the logic right, if we start with a base image that has no history but has layers, we create a new image that has one history entry in it and multiple layers, having bypassed the logic that tries to verify that the number of non-"empty-layer" history entries in the image we're about to create matches the number of layers the new image will have. If we then try to use that new image as a base image for another build, that logic won't be bypassed by default because that new image has some history information, and we'll get an error.

Yes  this edge case holds true since docker supports this scenario where it allows building on top of base with history null and adds history of new layer ( also discussed here: https://github.com/containers/buildah/issues/4025#issuecomment-1143400214 ). My original idea was to never allow to add history if base has a history `null` but this was added to follow compat with `docker`, But I think recommend use-case for such scenario would be always `--omit-history` and never build without `--omit-history`

nalind:
Okay, that makes sense.  Really not liking that it effectively imposes a requirement that the new option has be used, basically always, if one of these images was used as a base at any point in an image's ancestry, but since we need to produce images that look like what docker build produces, it sounds like we're going to be stuck with it.  Patch looks good.

TomSweeneyRedHat:
LGTM
/hold cancel

Issue #3960, opened 2022-05-04
FR: `--logfile` seperated file for each platform


Dexus:
**Description**
I would like to get for each platform I build, have a own logfile.

when I run:
```shell
buildah build -f "example-debian.Dockerfile" \
  --manifest example-debian \
  --jobs 3 --layers \
  --platform=linux/amd64,linux/arm64,linux/ppc64le,linux/mips64le,linux/s390x \
  --uts private --pull \
  --userns container --isolation oci \
  --network private \
  --logfile buildlog_example-debian.log \
  --squash \
  --build-arg DEMO=1 .
```

**Describe the results you received:**

The logfile looks very mixed :(

**Describe the results you expected:**
The logfile not mixed the steps in the output from multiple platforms. Or even better would be, for each platform have a own logfile, maybe with a placeholder in the `filename-%(platform_normalized).log` or with a argument that accepts a directory where all logs are placed.

**Output of `buildah version`:**

```
1.25.1
```

flouthoc:
Hi @Dexus Thanks for opening the issue looks like a nice feature request just curious does `docker or buildkit` already supports this ? If yes we might wanna choose their format for compat but if not I think we can choose any one of the requested above.

Dexus:
Hi @flouthoc, no both don't support this currently, but buildkit (`docker buildx build`) support `--progress plain` which will output with prefix, but don't mix up the output from multiple arch builds on the same time - there it is a clean output.

Dexus:
Well, so far I haven't seen any CLI that can build multiple architectures simultaneously and provide logfiles. Therefore, I am not currently aware of any precedent where this feature is available. There is just nothing at the moment where you can get a logfile for an architecture. Would at least be a super unique selling point for buildah.

rhatdan:
I wonder if we just added logdir option and had it create an arch directory with buildah.log in it would be best.

--logdir=/tmp/foobar and /tmp/foobar/lamd64/buildah.log,arm64/buildah.log,ppc64le/buildah.log, mips64le/buildah.log,s390x/buildah.log.  Perhaps a /tmpfoobar/buildah.log which would log the manifest log entries.

Dexus:
I like the idea of that.
But would prefer the format of the platform for the children directories.

`--logdir=/tmp/foobar --platform=windows/amd64,linux/amd64,linux/arm64,linux/arm/v7,linux/arm/v5...` 

should result in something like the same or with a dash instead of a slash

```
[...]

PR #4034, opened 2022-06-06
build, multiarch: support splitting build logs for `--platform`

flouthoc:
Support splitting build logs for multi-arch builds ( `--platform` ) by
using `--logfile` and `--split-logfile-by-platform` which splits
generated log file in the convention `${file}_${platform_os}_${platform_arch}`

Closes: https://github.com/containers/buildah/issues/3960

```release-note

Add support for --split-logfile-by-platform which allows users to split logfile with build logs for each platform

```


imagebuildah/build.go
```diff
@@ -253,6 +253,23 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
...
 		platformOptions.Args = argsCopy
 		builds.Go(func() error {
+			if platformOptions.LogFile != "" && platformOptions.LogSplitByPlatform {
+				logFile := platformOptions.LogFile + "_" + platformOptions.OS + "_" + platformOptions.Architecture
+				f, err := os.OpenFile(logFile, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, 0600)
+				if err != nil {
+					return errors.Errorf("error opening logfile %q: %v", logFile, err)
+				}
+				defer f.Close()
+				logger = logrus.New()
```
nalind, line 263: Isn't `logger` shared between all of these goroutines?

flouthoc: In this case since logger for all the goroutine is writing to their own unique filepaths hence I thought its better to re-init since for this particular case sharing was not helping a lot since all the logs go to different files for each go-routine.

flouthoc: Should be fixed now.

imagebuildah/build.go
```diff
@@ -253,7 +253,25 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 		}
 		platformOptions.Args = argsCopy
 		builds.Go(func() error {
-			thisID, thisRef, err := buildDockerfilesOnce(ctx, store, logger, logPrefix, platformOptions, paths, files)
+			loggerPerPlatform := logger
+			if platformOptions.LogFile != "" && platformOptions.LogSplitByPlatform {
+				logFile := platformOptions.LogFile + "_" + platformOptions.OS + "_" + platformOptions.Architecture
+				f, err := os.OpenFile(logFile, os.O_CREATE|os.O_TRUNC|os.O_WRONLY, 0600)
+				if err != nil {
+					return errors.Errorf("error opening logfile %q: %v", logFile, err)
```
nalind, line 261: This should probably be `errors.Wrapf()`, or if we're using the standard library's `Errorf()`, it should use "%w" as the specifier for `err`.

pkg/cli/common.go
```diff
@@ -211,6 +212,7 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
 	fs.IntVar(&flags.Jobs, "jobs", 1, "how many stages to run in parallel")
 	fs.StringArrayVar(&flags.Label, "label", []string{}, "set metadata for an image (default [])")
 	fs.StringVar(&flags.Logfile, "logfile", "", "log to `file` instead of stdout/stderr")
+	fs.BoolVar(&flags.LogSplitByPlatform, "split-logfile-by-platform", false, "split logfile to different files for each platform")
```
rhatdan, line 215: Could you shorten this to something like `--logsplit` I hate multi word options, and this might be the worst one ever.  Document the option in the help, do not need to have an explanation in the option name itself.  Also the rest of the world is not as good a typest as you.

PR #4043, opened 2022-06-07
copier: add `NoOverwriteNonDirDir` option

vrothberg:
Similar to the `NoOverwriteDirNonDir` one, add an option that disables
non-directories from being overwritten by directories.

Required-for: containers/podman/issues/14420

#### What type of PR is this?

> /kind feature

Add a new option to copier.Put to throw an error if a non-dir is overwritten with a dir.  Needed for Docker-compat in `podman cp`.

#### How to verify it

`go test -v ./copier`.

```release-note
Add a new `NoOverwriteNonDirDir` option to `copier.Put` to prevent a non-directory from being overwritten by a directory.
```

@nalind PTAL

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -344,6 +344,7 @@ type PutOptions struct {
 	IgnoreXattrErrors    bool              // ignore any errors encountered when attempting to set extended attributes
 	IgnoreDevices        bool              // ignore items which are character or block devices
 	NoOverwriteDirNonDir bool              // instead of quietly overwriting directories with non-directories, return an error
+	NoOverwriteNonDirDir bool              // instead of quietly overwriting non-directories with directories, return an error
 	Rename               map[string]string // rename items with the specified names, or under the specified names
 }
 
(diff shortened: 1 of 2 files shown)


copier/copier.go
```diff
@@ -1794,12 +1795,17 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
...
+						if !st.IsDir() {
+							if req.PutOptions.NoOverwriteNonDirDir {
+								return errors.Wrapf(err, "copier: put: error creating directory at %q", path)
+							}
+							if err = os.Remove(path); err == nil {
+								err = os.Mkdir(path, 0700)
+							}
 						}
+					} else {
+						err = stErr
```
vrothberg, line 1808: This needs to change.

copier/copier.go
```diff
@@ -1794,9 +1795,10 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
 				}
 			case tar.TypeDir:
 				if err = os.Mkdir(path, 0700); err != nil && os.IsExist(err) {
-					var st os.FileInfo
-					if st, err = os.Lstat(path); err == nil && !st.IsDir() {
-						// it's not a directory, so remove it and mkdir
+					if st, stErr := os.Lstat(path); stErr == nil && !st.IsDir() {
+						if req.PutOptions.NoOverwriteNonDirDir {
+							return errors.Wrapf(err, "copier: put: error creating directory at %q", path)
+						}
```
nalind, line 1801: I'd kind of prefer if this followed the style of the various `NoOverwriteDirNonDir` cases in this loop, which break out instead of returning directly, in case we need to add cleanup logic later, but it's not a blocker.

nalind (approved):
LGTM with an optional nit.  The unit test is much appreciated!

copier/copier.go
vrothberg: Will do, thank you!

flouthoc (approved):
/lgtm
/hold

PR #4044, opened 2022-06-07
Sort buildoptions and move cli/build functions to internal

rhatdan:

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/cmd/buildah/addcopy.go
+++ b/cmd/buildah/addcopy.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/containers/buildah"
+	"github.com/containers/buildah/internal/util"
 	buildahcli "github.com/containers/buildah/pkg/cli"
 	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/common/pkg/auth"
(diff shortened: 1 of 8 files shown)


cmd/buildah/addcopy.go
```diff
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/containers/buildah"
+	"github.com/containers/buildah/internal/util"
```
TomSweeneyRedHat, line 10: Could we use "iutil" here and in the other files just to make it easier to search through the whole project?  I realize you did it elsewhere to avoid the collision, but it would be good to have it the same everywhere.

flouthoc (approved):
LGTM

cmd/buildah/addcopy.go
rhatdan: I will let @nalind  decide, since we have not done this with prior conflicts.

nalind: Like alphabetizing fields, it solves a problem I don't encounter, so I have no preference.

PR #4052, opened 2022-06-13
Make util and copier build on FreeBSD

dfr:
The mknod syscall takes a uint64 value for dev on FreeBSD.

#### What type of PR is this?

> /kind other

Needed for porting buildah to FreeBSD.

### DIFF
--- a/copier/mknod_int.go
+++ b/copier/mknod_int.go
@@ -1,0 +1,12 @@
+//go:build !windows && !freebsd
+// +build !windows,!freebsd
+
+package copier
+
+import (
+	"golang.org/x/sys/unix"
+)
+
+func mknod(path string, mode uint32, dev int) error {
+	return unix.Mknod(path, mode, dev)
+}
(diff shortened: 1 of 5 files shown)


flouthoc:
@dfr Could you add `[NO NEW TESTS NEEDED]` in the commit message.

copier/mknod_int.go
```diff
@@ -0,0 +1,11 @@
+//go:build !windows && !freebsd
```
nalind, line 1:
```suggestion
//go:build !windows && !freebsd
// +build !windows,!freebsd
```

nalind:
The compiler's complaining about mismatched "//go:build" and "// +build" directives, in this case because the new files don't include the older style.  Otherwise, LGTM.

dfr:
Both build failures look like this:
```
[+2568s] ok 684 buildah version current in package # skip buildah binary is not owned by package
```

nalind:
The "not ok" tests are the ones to look for there.  Restarted the two that have failed because they're both getting TLS-related errors while talking to a registry, which is hopefully just a transient error.

flouthoc (approved):
/lgtm
/approve

PR #4054, opened 2022-06-13
copier: work around freebsd bug for "mkdir /"

dfr:
On FreeBSD, os.Mkdir can return EISDIR if the target exists. The
implementation for os.MkdirAll first checks if the target exists so
avoids generating this error.

#### What type of PR is this?

> /kind other

Needed for porting buildah to FreeBSD.

```release-note
None
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -1558,7 +1558,9 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
 					directoryModes[path] = defaultDirMode
 				}
 			} else {
-				if !os.IsExist(err) {
+				// FreeBSD can return EISDIR for "mkdir /":
+				// https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=59739.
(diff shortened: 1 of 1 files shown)


dfr:
Looks like this breaks some of the conformance tests - I'll try to re-work it to avoid using MkdirAll.

dfr:
Doing a little more research I found an existing issue for this EISDIR behaviour (https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=59739). I plan to fix this 19 year old bug in the FreeBSD kernel but in the meantime, we can work around the problem by treating EISDIR the same as EEXIST.

flouthoc (approved):
LGTM

nalind:
Restarted the failing job, which looked like it caught a transient error again.  Looks good, though we're not really set up to test it.

dfr:
Don't worry too much about testing just yet - I'm close to being able to make a PR for the buildah port which will probably add a cross-build target for freebsd for a smoke test. Longer term, I want to figure out how much can be reasonably added to the regular CI without adding too much burden to the workflow (or costs for running extra CI workloads).

PR #4055, opened 2022-06-13
Replace $BUILDAH_BINARY with buildah() function

hshiina:

#### What type of PR is this?

/kind failing-test 

buildah() defined in helpers.bash is available instead of `$BUILDAH_BINARY`. This function calls `$BUILDAH_BINARY` with common options for tests.

I sometimes hit the Docker pull limit error locally at this test because `--registries-conf` is not passed.

#### How to verify it
Run the fixed test. Then, run `buildah images`.

With this fix, no images are displayed. Without this fix, `localhost/test-bud` and `docker.io/library/alpine` are displayed because `--root` flag is not used in the test.

None

```release-note
None
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -885,8 +885,8 @@ _EOF
 FROM alpine
 RUN echo 'hello'> hello
 _EOF
-  # Using BUILDAH_BINARY since run_buildah adds unwanted chars to tar created by pipe.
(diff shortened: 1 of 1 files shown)


rhatdan:
@edsantiago @nalind PTAL

edsantiago:
/lgtm
/hold
Nice catch.

PR #4056, opened 2022-06-13
.Docker.History .OCIv1.History and .History are the same thing

rhatdan:
Since tests are treating these as different fields, and they podman
inspect does not support the failover, we can not use these tests
against podman.

Removing .Docker.History and .OCIv1.History fixes the issue, and I
believe does not change anything.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -214,10 +214,6 @@ _EOF
 
   # Built image must not contain history for the layers which we have just built.
   run_buildah build $WITH_POLICY_JSON --omit-history -t source -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile1
-  run_buildah inspect --format "{{index .Docker.History}}" source
-  expect_output "[]"
-  run_buildah inspect --format "{{index .OCIv1.History}}" source
-  expect_output "[]"
   run_buildah inspect --format "{{index .History}}" source
   expect_output "[]"
 }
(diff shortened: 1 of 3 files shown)


rhatdan:
Help to fix https://github.com/containers/podman/pull/14561

rhatdan:
@nalind Could you point out where my assumption is flawed.

edsantiago:
I think this is overkill. Podman tests are failing only in podman-remote, not in podman. I think the solution to your podman PR is to get the `--omit-history` flag working in podman-remote.

flouthoc (commented):
In spec `{{index .Docker.History}}` and `{{index .OCIv1.History}}` are different fields and different lines are responsible for setting this field merging this with `.History` can leave a room for regression at build end but I think podman only cares about fetching from any one of them.

It would have been easier if podman tests could have just replaced this with `.History` and buildah could still test this.

edsantiago:
OK it's morning, my brain is fresh, but I'm still confused as to what problem this PR is intended to solve. Given your mention of `podman inspect`, my best guess is that this is a misunderstanding. Let me see if this helps: in the podman `buildah-bud` test suite, `run_buildah build` is replaced by `podman build` (or `podman-remote build`) but `run_buildah inspect` is left alone: it uses actual `buildah`, not podman. There is no place anywhere in buildah tests that use `podman inspect`. The tests which you update in this PR are, as best I can tell, all running and passing under podman. Confirmation: (1) I grepped for those test names in `apply-podman-deltas`, no hits; and (2) I grepped for those test names in [the log of your buildah-vendor PR](https://storage.googleapis.com/cirrus-ci-6707778565701632-fcae48/artifacts/containers/podman/6042439294124032/html/bud-podman-fedora-36-root-host.log.html), and every one of them is green, not yellow (yellow means skipped). Does this help, or am I still confused?

nalind:
> @nalind Could you point out where my assumption is flawed.

I'm confused about why we're patching one of buildah's tests in an attempt to fix a problem that podman's having when it tries to reuse buildah's tests, when podman already applies patches to buildah's tests.

nalind:
Thinking about this some more, if the PR is attempting to fix a problem being encountered elsewhere _by reducing test coverage here_, I can't see how I could approve of it.

rhatdan:
Your right, I had too faulty assumptions, one is --omit-history was going to fail in podman -remote testing( it no longer does.)
And .Docker.History and .OCIv1.History did not exists in inspect, they must be hidden.

PR #4057, opened 2022-06-14
More test cleanup

edsantiago:
- "run_buildah 0" should be simply "run_buildah" (the 0 is implicit)

- remove "mktemp -d". Those are evil. (Reason: they are not cleaned up,
  so they pollute the test system. We have $TEST_SCRATCH_DIR. Use it.

- replace a few icky '[[ ... ]]' checks with 'assert'

Signed-off-by: Ed Santiago <santiago@redhat.com>

```release-note
None
```

PR #4058, opened 2022-06-14
options.Quiet should be passed into preprocessor

rhatdan:
Preprocessor should not be printing warnings in quiet mode.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```


rhatdan:
@edsantiago This will allow us to test against podman since quiet means quiet.

tests/bud.bats
```diff
@@ -1765,7 +1765,7 @@ _EOF
 
 @test "bud with preprocessor error" {
   target=alpine-image
-  run_buildah 0 bud -q $WITH_POLICY_JSON -t ${target} -f Error.in $BUDFILES/preprocess
+  run_buildah 0 bud $WITH_POLICY_JSON -t ${target} -f Error.in $BUDFILES/preprocess
```
edsantiago, line 1768: This is going to conflict with #4057

edsantiago (commented):
LGTM in principle, but (1) there's a conflict waiting to happen, and (2) I'm undecided about the `-q` behavior. A part of me thinks that the warning should not be silenced, even with `-q`.

tests/bud.bats
rhatdan: Ok let's merge yours and then I will rebase

flouthoc:
Rebase needed.

rhatdan:
@nalind @flouthoc @giuseppe @vrothberg @umohnani8 @ashley-cui PTAL

edsantiago:
After a day's thought, I feel more strongly that this should not merge. Someone passing `-q` is, IMO, saying "Shut up and give me the image ID, but hey, if something goes wrong, I'd like to know about it". This PR does not honor that. But I'm just one small voice.

rhatdan:
Would have to check if Warnings are currently silenced by -q? @nalind WDYT?  Should Warnings be exposed when running in --quiet mode?

nalind:
`--quiet` controls progress messages ("STEP ...", "COMMIT", pull progress, etc.), and hasn't historically affected logging level.  Is there an issue that's motivating this change?

rhatdan:
There is a bug in podman-remote that looks like it is effected by this. but I will see why it is happening there.

PR #4059, opened 2022-06-15
Add FreeBSD support for buildah run

dfr:
#### What type of PR is this?

> /kind other

This adds support for running containers on FreeBSD, including support for volume mounts and CNI-based networking. This stack of diffs is missing a required change to github.com/opencontainers/runtime-tools which enables OCI spec generation for FreeBSD which is awaiting review (https://github.com/opencontainers/runtime-tools/pull/746). 

```release-note
None
```

[NO NEW TESTS NEEDED]


pkg/jail/jail.go
```diff
@@ -0,0 +1,193 @@
...
+func jailSet(jconf *config, flags int) (*jail, error) {
+	jiov, err := jconf.getIovec()
+	if err != nil {
+		return nil, err
+	}
+
+	jid, _, errno := syscall.Syscall(unix.SYS_JAIL_SET, uintptr(unsafe.Pointer(&jiov[0])), uintptr(len(jiov)), uintptr(flags))
+	if errno != 0 {
+		return nil, errno
+	} else {
```
rhatdan, line 152: No reason for `else` here. Just return the jail.

pkg/jail/jail.go
```diff
@@ -0,0 +1,193 @@
...
+func jailGet(jconf *config, flags int) (*jail, error) {
+	jiov, err := jconf.getIovec()
+	if err != nil {
+		return nil, err
+	}
+
+	jid, _, errno := syscall.Syscall(unix.SYS_JAIL_GET, uintptr(unsafe.Pointer(&jiov[0])), uintptr(len(jiov)), uintptr(flags))
+	if errno != 0 {
+		return nil, errno
+	} else {
```
rhatdan, line 168: No reason for `else` here. Just return the jail.

pkg/chrootuser/user_freebsd.go
```diff
@@ -0,0 +1,322 @@
...
+func openChrootedFile(rootdir, filename string) (*exec.Cmd, io.ReadCloser, error) {
+	// The child process expects a chroot and one or more filenames that
+	// will be consulted relative to the chroot directory and concatenated
+	// to its stdout.  Start it up.
+	cmd := reexec.Command(openChrootedCommand, rootdir, filename)
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, nil, err
+	}
+	err = cmd.Start()
```
rhatdan, line 72: nit: 
if err := cmd.Start; err != nil {
 		return nil, nil, err
}

pkg/chrootuser/user_freebsd.go
```diff
@@ -0,0 +1,322 @@
...
+	home string
+}
+type lookupGroupEntry struct {
+	name string
+	gid  uint64
+	user string
+}
+
+func readWholeLine(rc *bufio.Reader) ([]byte, error) {
+	line, isPrefix, err := rc.ReadLine()
```
rhatdan, line 97: This whole function looks like it could be simplified to something like:

```
for {
       line, isPrefix, err = rc.ReadLine()
       if err {
           return nil, err
     }
     if !isPrefix {
         break
    }
}
return line, nil
```

pkg/chrootuser/user_freebsd.go
```diff
@@ -0,0 +1,322 @@
...
+		return nil
+	}
+	return &lookupGroupEntry{
+		name: fields[0],
+		gid:  gid,
+		user: fields[3],
+	}
+}
+
+func lookupUserInContainer(rootdir, username string) (uid uint64, gid uint64, err error) {
```
rhatdan, line 164: Can't some of this code be shared?

rhatdan (approved):

rhatdan:
Adding [NO NEW TESTS NEEDED] to the github.com should be enough.

pkg/chrootuser/user_freebsd.go
dfr: Actually, I think that all of this can be shared between linux and freebsd

pkg/jail/jail.go
```diff
@@ -0,0 +1,191 @@
...
+	if errno != 0 {
+		return nil, errno
+	}
+	return &jail{
+		jid: int32(jid),
+	}, nil
+}
+
+func Create(jconf *config) (*jail, error) {
+	return jailSet(jconf, C.JAIL_CREATE)
```
gizahNL, line 174: You could import syscall & use syscall.SYS_JAIL_ATTACH

That would remove dependency on CGO ;)

dfr: I will look into that - thanks for the feedback. Do you think this is necessary for this PR or can that be an area for future cleanup? I did add nocgo stubs so that cross builds from non-freebsd hosts can work.

run_freebsd.go
```diff
@@ -0,0 +1,1102 @@
...
+
+	if !(contains(volumes, "/etc/resolv.conf") || (len(b.CommonBuildOpts.DNSServers) == 1 && strings.ToLower(b.CommonBuildOpts.DNSServers[0]) == "none")) {
+		resolvFile, err := b.addResolvConf(path, rootIDPair, b.CommonBuildOpts.DNSServers, b.CommonBuildOpts.DNSSearch, b.CommonBuildOpts.DNSOptions, nil)
+		if err != nil {
+			return err
+		}
+		// Only bind /etc/resolv.conf if there's a network
+		if options.ConfigureNetwork != define.NetworkDisabled {
+			bindFiles["/etc/resolv.conf"] = resolvFile
+		}
```
Luap99, line 206: looks like this is not up to date with the run_linux code?

Luap99 (commented):
It looks like there is still a lot of duplication between run_freebsd.go and run_linux.go.
Can we consolidate this more to only have the actual difference in these files.

dfr:
> It looks like there is still a lot of duplication between run_freebsd.go and run_linux.go.
> Can we consolidate this more to only have the actual difference in these files.

There is a lot of similar code around volume mounts - I'll work through that and consolidate. This is likely to be in small pieces to get a signal from CI that I haven't broken anything.

pkg/jail/jail.go
dfr: I copied in the constants - the rest of the code was already using syscall so this should now build on other platforms.

dfr:
The failing check looks like a transient:
```
[+1759s] # $ /var/tmp/go/src/github.com/containers/buildah/tests/./../bin/buildah rm time="2022-06-19T10:03:58Z" level=warning msg="Failed, retrying in 2s ... (1/3). Error: copying system image from manifest list: parsing image configuration: Get \"https://cdn02.quay.io/sha256/b0/b0757c55a1fdbb59c378fd34dde3e12bd25f68094dd69546cf5ca00ddbaa7a33?Expires=1655633638&Signature=XxdWm6xJTnN0KMze1BhdfeokentQb3-LMl0D9HhJmWi8femnlufaHFuLpVriBqBmb7Vo9WzXUcJG5hwiU6tNtyOyDN4hX9~ktC29LzDeslabDwyC3gSw7zN~yIGXoB0N~dbKl~P~J1m2W6equTYveh8mET-z5QPvgqtqNL4L1M68EU2bg2mFky02CRy1ndc~vbAXw6FwwXfQ3fIHqZ8GS3JmYqgesxN6kl6tcdEXbAZ6clKzJD-VtMFMEE2ysyFQBz5V8EW3KTV7BnnwhBCsed1eqG1vLhiAVNQiIxRMPQfK6eDWVnOa2p-bQRigJQCIFHw7ZL6IBoedY2Y~muziOg__&Key-Pair-Id=APKAJ67PQLWGCSP66DGA\": read tcp 10.128.0.13:45120->108.156.91.78:443: read: connection reset by peer" docker-working-container
```

flouthoc (commented):
Restart failing check  since it looks like network issue.

Just curious if we need commit for each function migration or all this could be squashed in a blanket one or two commits with `run: share common with all platforms` and `add FreeBSD support` for everything else.

But don't have a strong opinion on this one and i am okay if other maintainers are fine with it.

Luap99:
Individual commits makes this much easier to review, so I am in favour of that.

dfr:
I hope that the invidividual commits help with review. I also thought they might help to simplify merge conflicts since most are just moves of unmodified (or nearly unmodified) code. Certainly open to squashing it down when people are happy with the review if that will make the history cleaner.

dfr:
Rebased against today's main

run_freebsd.go
```diff
@@ -0,0 +1,572 @@
...
+	setupTerminal(g, options.Terminal, options.TerminalSize)
+
+	configureNetwork, configureNetworks, err := b.configureNamespaces(g, &options)
+	if err != nil {
+		return err
+	}
+
+	containerName := Package + "-" + filepath.Base(path)
+	if configureNetwork {
+		g.AddAnnotation("org.freebsd.parentJail", containerName+"-vnet")
```
flouthoc, line 150: Is there any thread for freebsd variant of runtime-spec, could we add that in comments so its easier to review.

run_freebsd.go
```diff
@@ -0,0 +1,572 @@
...
+			}
+		} else if err := g.AddOrReplaceLinuxNamespace(namespaceOption.Name, namespaceOption.Path); err != nil {
+			if namespaceOption.Path == "" {
+				return false, nil, false, errors.Wrapf(err, "error adding new %q namespace for run", namespaceOption.Name)
+			}
+			return false, nil, false, errors.Wrapf(err, "error adding %q namespace %q for run", namespaceOption.Name, namespaceOption.Path)
+		}*/
+	}
+
+	if /*configureNetwork*/ false {
```
flouthoc, line 446: This block should be removed since it is not getting used.

run_common.go
```diff
@@ -39,8 +41,10 @@ import (
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/ioutils"
+	"github.com/containers/storage/pkg/lockfile"
 	"github.com/containers/storage/pkg/reexec"
 	"github.com/containers/storage/pkg/unshare"
+	storagetypes "github.com/containers/storage/types"
```
flouthoc, line 47: Nit: for naming consistency
```diff
- storagetypes
+ storageTypes
```

flouthoc:
Not at all a complete review yet but just a small doubt @dfr so all the commits with title `Move x to run_common.go` are intact movement of function to a common file or can they include changes in between as well ?  Just asking to make sure cause i saw changes specific to `freebsd` in 6061d5828a3c5bc04f57f3d746c4ecf676a52760 so just wanna make sure that we are not leaking any diff in `run_common.go` and they are just intact movement of functions.

run_freebsd.go
dfr: There isn't anything yet - Samuel Karp and I planned to setup a working group for this - hopefully starting the process some time next week. In the absence of any kind of formal FreeBSD specification, I'm using (abusing?) an annotation here for passing through network namespace information to the runtime.

run_freebsd.go
```diff
@@ -0,0 +1,572 @@
...
+				}
+				configureNetwork = (policy != define.NetworkDisabled)
+			}
+		case string(specs.UTSNamespace):
+			configureUTS = false
+			if !namespaceOption.Host && namespaceOption.Path == "" {
+				configureUTS = true
+			}
+		}
+		/* TODO: re-visit this when there is consensus on a FreeBSD runtime-spec
```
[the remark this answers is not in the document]
dfr: I will remove the commented out code for now but leave the reminder comment and add any discussion thread when we actually start one.

dfr:
rebased

giuseppe (approved):
LGTM

rhatdan:
@nalind @flouthoc @Luap99 PTAL

dfr:
I noticed an unwanted addition to a commit log message deep in this stack and re-pushed - there should be no functional change.

rhatdan:
@nalind @flouthoc @Luap99 @giuseppe @vrothberg PTAL

dfr:
Rebased.

flouthoc (approved):
LGTM @nalind PTAL

dfr:
Not sure whats going on with the Test Vendoring CI - this PR doesn't modify go.mod

Luap99:
> Not sure whats going on with the Test Vendoring CI - this PR doesn't modify go.mod

Do not import `github.com/pkg/errors` just use `errors`, see https://github.com/containers/buildah/pull/4091

dfr:
> > Not sure whats going on with the Test Vendoring CI - this PR doesn't modify go.mod
> 
> Do not import `github.com/pkg/errors` just use `errors`, see #4091

Ah - I see a stray import in run_freebsd.go. Thanks for the pointer!

dfr:
Rebased

PR #4062, opened 2022-06-17
buildah.spec.rpkg: initial addition

lsm5:
This specfile along with a webhook will trigger auto rpm builds on the
`rhcontainerbot/podman-next` copr after every upstream PR merge.

Useful for those who want to fetch the latest upstream buildah without
having to build from source.

This commit also gets rid of `contrib/rpm/buildah.spec` and
`tests/version.bats`.

Known issue: Currently fails to build for EL8 envs.

Successful run at https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/4542963/ 

@nalind @rhatdan @flouthoc PTAL

#### What type of PR is this?

/kind other

This PR will trigger rpm builds for buildah on `rhcontainerbot/podman-next` copr on every merge to the main branch.

#### How to verify it
Check if new builds get triggered at https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/

None

None

```release-note
None
```


lsm5:
btw, are we already sending builds to any other copr besides `rhcontainerbot/podman-next` ? I see a webhook `https://copr.fedorainfracloud.org/webhooks/github/18082/b6f1f77e-7eb1-4114-9c52-14b786833301/` in the Settings that's been failing for quite some time now. I can get rid of it if we're not using it.

rhatdan:
@cevich and @TomSweeneyRedHat or @flouthoc Might know about the github action.  Perhaps building images on quay?

lsm5:
I'll look at getting rid of `contrib/rpm/buildah.spec`

lsm5:
removed the stale spec file, let's see what blows up now :D

lsm5:
@cevich just to confirm, this will build rpms using the `last merged PR` and not the `current in-progress PR`. So, this likely isn't a replacement for `build from scratch` at least in it's current state. Is that ok?

lsm5:
> Oh yeah. Yep, once this goes "live" we need to update `contrib/buildahimage/upstream` to install the RPM instead of building from scratch. That will improve the multi-arch build-time by probably double. Big thanks @lsm5

@cevich ah ok, it uses the latest main branch itself, and not the in-progress PR in Containerfile. So we should be fine. Correct me if I'm wrong.

lsm5:
I'll update `contrib/buildhaimage/upstream/Containerfile` in this PR

lsm5:
RE: `make validate`: `rpm -qV shadow-utils` barfs in `buildah from buildah` but works on my local environment. Is that expected?

flouthoc (approved):
LGTM 

@lsm5 Could you rebase plz :)

Issue #4060, opened 2022-06-16
[FEATURE REQUEST]: --userns=auto for build


lukasmrtvy:
**Description**

Would be great to have a feature parity with `podman run` and implement support for `--userns=auto` flag for buildah. Thanks

flouthoc:
@lukasmrtvy Thanks for creating the issue. I'll take a look at this. It would we nice if RFE also contains a small gist about the actual use-case which you are trying with build containers by running in unique user ns.

lukasmrtvy:
Use case:
- building and running third-party content on a single machine ( without any additional isolation/abstraction ), typically it can be a workflow scheduler/executor that runs arbitrary code from multiple users

Workaround:
- running buildah in podman like `podman run --userns=auto --isolation=chroot ...`  can probably do the trick but it could be a complex task to solve, as it involves exporting tar artifact

Does it make sense?

Thanks

flouthoc:
@lukasmrtvy Sure thanks for sharing and makes sense to me :)

lukasmrtvy:
I am trying this implementation in Podman 4.2.0. Seems its working via CLI, but API call does not work.

**Dockerfile**
```
cat << EOF > Dockerfile
FROM alpine
RUN cat /proc/self/uid_map
EOF
```
Works correctly via CLI
```
podman build -t test --userns=auto .
```
Does not work via CLI

```
tar -czf context.tar.gz Dockerfile
curl -s --unix-socket /run/podman/podman.sock -X POST  -H "Content-Type:application/tar"  --data-binary "@context.tar.gz"   'http://d/v4.2.0/libpod/build?userns=auto'
```

flouthoc:
@lukasmrtvy ackd.

flouthoc:
This is a bug i think happening because I think `IDMappingOptions` ( https://github.com/containers/buildah/blob/main/define/build.go#L220 ) is not wired to API end , so this must be implemented at bindings end and must be processed and supplied to API end here https://github.com/containers/podman/blob/main/pkg/api/handlers/compat/images_build.go#L603

I'm assigning this issue to my self.

PR #4064, opened 2022-06-20
build, userns: add support for `--userns=auto`

flouthoc:
Buildah now supports `--userns=auto` which can automatically pick an empty range and create an user namespace for the container.

The `--userns=auto` flag, requires that the user name containers and a range of subordinate user ids that the build container is allowed to use be specified in the /etc/subuid and /etc/subgid files.

Example: `containers:2147483647:2147483648`.

Buildah allocates unique ranges of UIDs and GIDs from the containers subordinate user ids. The size of the ranges is based on the number of UIDs required in the image. The number of UIDs and GIDs can be overridden with the size option.

Valid auto options:

    * gidmapping=CONTAINER_GID:HOST_GID:SIZE: to force a GID mapping to be present in the user namespace.

    * size=SIZE: to specify an explicit size for the automatic user namespace. e.g. --userns=auto:size=8192. If size is not specified, auto will estimate a size for the user namespace.

    * uidmapping=CONTAINER_UID:HOST_UID:SIZE: to force a UID mapping to be present in the user namespace.

Closes: https://github.com/containers/buildah/issues/4060

### DIFF
--- a/define/types.go
+++ b/define/types.go
@@ -17,6 +17,7 @@ import (
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/chrootarchive"
 	"github.com/containers/storage/pkg/ioutils"
+	"github.com/containers/storage/types"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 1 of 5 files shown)


giuseppe (approved):
LGTM

Issue #3912, opened 2022-04-12
Tidy Demo's, Rootless Accomodated?


andrewgdunn:
**Description**

On Fedora 35, running as non-root. Fresh install without any cached short name aliases (e.g. `~/.cache/containers/short-name-aliases.conf`) the `demo/buildah_multi_stage.sh` script fails: 

```
++ buildah bud -t multifromfile:latest -f ./Dockerfile.multi .
error creating build container: short-name resolution enforced but cannot prompt without a TTY
```
This could be resolved by putting `docker.io/` in the `from` invocations. Reference [container-image-short-names](https://www.redhat.com/sysadmin/container-image-short-names).

It also fails at the second portion of the demo:

```
Let's mount the container getting the root directory
cannot mount using driver overlay in rootless mode. You need to run it in a `buildah unshare` session
```
This to me is more tricky, there is some discussion in #1279 that I felt might help, also [this documentation](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/building_running_and_managing_containers/assembly_building-container-images-with-buildah_building-running-and-managing-containers#proc_modifying-a-container-using-buildah-mount_assembly_building-container-images-with-buildah) but ultimately I couldn't get my head around how to do this inline.

If I was able to sort out both of these problems I would have offered a PR, however I was stumped by the rootless mount challenges.

andrewgdunn:
I guess I should point out that in [the readme](https://github.com/containers/buildah/blob/main/demos/README.md?plain=1#L28) maybe it's intended that these demos are run as root and rootless is not a reasonable thing to try with them in their current form?

rhatdan:
The should work in both modes, the issue is no one has run them since the shortnames change.

andrewgdunn:
I don't think it actually works with rootless right now, there isn't any unshare directives in the scripts right?

```
$ podman --version
podman version 3.4.4
$ buildah --version
buildah version 1.23.3 (image-spec 1.0.1-dev, runtime-spec 1.0.2-dev)
```

```
$ podman unshare cat /proc/self/uid_map 
         0       1000          1
         1     100000      65536
```

```
Let's build the container with Buildah, first GoLang
Let's mount the container getting the root directory
cannot mount using driver overlay in rootless mode. You need to run it in a `buildah unshare` session
Let's get x/net/html into the container
   redact

Copy app.go into the container
cp: cannot create regular file '/go': Permission denied

rhatdan:
[...]

PR #4065, opened 2022-06-20
[CI:DOCS] demo: use `buildah unshare` for `rootless` invocations

flouthoc:
Use `buildah unshare` to invoke `buildah mount` when demos are invoked
from rootless environments

* buildah_mult_stage: only unshare on needed steps
* buildah-scratch-demo: unshare entire script since it uses pkg managers
  commands which don't work well in rootless session without unshare.
  
  Closes: https://github.com/containers/buildah/issues/3912

### DIFF
--- a/demos/buildah_multi_stage.sh
+++ b/demos/buildah_multi_stage.sh
@@ -5,6 +5,22 @@
 # Assumptions install buildah and podman
 # Set some of the variables below
 
+
+#################
+#  is_rootless  #  Check if we run as normal user
+#################
+function is_rootless() {
+    [ "$(id -u)" -ne 0 ]
+}
+
+## The `buildah mount` only work as root so use
+## `buildah unshare` command which switches your
+## user session to root within the user namespace.
+if is_rootless; then
+  buildah unshare $0
+  exit
+fi
+
 demoimg=mymultidemo
 quayuser=myquayuser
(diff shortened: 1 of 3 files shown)


demos/buildah_multi_stage.sh
```diff
@@ -60,7 +68,11 @@ buildah rm -a
 read -p "${green}Let's build the container with Buildah, first GoLang${reset}"
 buildcntr=$(buildah from golang:1.7.3)
 read -p "${green}Let's mount the container getting the root directory${reset}"
-buildmnt=$(buildah mount $buildcntr) 
+if is_rootless; then
+ buildmnt=$(buildah unshare buildah mount $buildcntr)
```
rhatdan, line 72: I am not sure this will work, since when buildah unshare completes the mount point will be removed.  The entire script needs to be run within buildah unshare.

flouthoc: Moved to running entire script in unshared session.

flouthoc:
@rhatdan PTAL again.

rhatdan:
LGTM
@vrothberg @giuseppe @nalind @TomSweeneyRedHat @umohnani8 @ashley-cui PTAL

giuseppe (approved):
/lgtm

Issue #4068, opened 2022-06-21
[FEATURE REQUEST]: oci hook for RUN instruction


lukasmrtvy:
I am not familiar with the exact relation between the build phase and runc/crun, but I believe We are missing some custom oci hook stage for RUN instruction which will help us trigger for example custom iptable rules. 

This one is useful with a combination of already implemented --userns=auto parameter https://github.com/containers/buildah/issues/4060. 

Use case:
- limit outbound traffic via iptables to internal networks
- for building arbitrary code in "sensitive" environment

Snippet:
```
cat << EOF > /etc/containers/restrict-outbound-access
#!/bin/sh
/usr/sbin/iptables-nft -A OUTPUT -d 192.168.0.0/16,172.16.0.0/12,10.0.0.0/8 -j DROP
EOF

cat << EOF > /etc/containers/oci/hooks.d/restrict-outbound-access.json
{
  "version": "1.0.0",
  "hook": {
    "path": "/etc/containers/restrict-outbound-access"
  },
  "when": {
    "annotations": {
      "restrict-outbound-access": "true"
    }
  },
  "stages": ["buildContainer"]  <---- buildContainer stage does not exist
}
EOF

podman build --annotation "restrict-outbound-access=true" ...
```

Possible workaround:
```
id=<random>

cat << EOF > Dockerfile
FROM busybox
RUN echo "foo"
EOF

tar -czvf "context_$id.tar.gz" Dockerfile

mkdir -p "./builds/dir_$id/"

mv "context_$id.tar.gz" "./builds/dir_$id/"

podman run --userns=auto --annotation "restrict-outbound-access=true" --rm -it -v "./builds/dir_$id/:/build/" buildah:latest ... --output "/build/image_$id.tar"

podman load "./builds/dir_$id/image_$id.tar" "$prefix/image_$id"

rm "./builds/dir_$id/image_$id.tar"

podman push "$prefix/image_$id"
```
, but it comes with an additional performance and time penalty ( load image to local registry )

Thanks

flouthoc:
Hi @lukasmrtvy , Thanks for creating the issue i'll take a look at this.

lukasmrtvy:
@flouthoc hey, thanks

[...]

PR #4071, opened 2022-06-22
build: support `OCI-hooks` for ephemeral `build` containers

flouthoc:
Following PR adds support for OCI hooks attached to each of ephemeral
build contains which are created by `RUN` step invoked during `buildah
build` command.

Following PR also adds `--hooks-dir` flag to `buildah build` option
which allows end-users to specify one or more configuration directories
for OCI-hooks.

Note: Following PR migrates `pkg/hooks` from `podman` to `buildah` so
buildah could implement this feature and once merged `podman` can start
using `buildah/pkg/hooks`.
Ref: https://github.com/containers/podman/tree/main/pkg/hooks

For usage see man-page and example test-case.

See: https://github.com/opencontainers/runtime-spec/blob/main/config.md#posix-platform-hooks

Closes: https://github.com/containers/buildah/issues/4068


pkg/hooks/0.1.0/hook.go
```diff
@@ -0,0 +1,88 @@
+// Package hook is the 0.1.0 hook configuration structure.
```
flouthoc, line 1: Note for reviewers: `pkg/hooks` is migrated from https://github.com/containers/podman/tree/main/pkg/hooks so both `podman` and `buildah` can re-use it.

flouthoc: I was unable to spot users other than `buildah` and `podman` since its only used while running containers and pkg had very low activity for feature/bug changes hence decided to move to buildah but I am okay to move to `c/common` if it makes more sense there.

flouthoc:
> @cevich `Cross Compile` is failing but it looks like a homebrew issue to me. Do we need to update packages manually on VM or is it just that `https://ghcr.io/v2/homebrew/` is down.

Please ignore this message it was a temporary glitch.

cevich:
>Edit: Please ignore it was a temporary glitch.

Ya, it does happen from time to time.

rhatdan:
Lets put the hooks handling into containers/common.

pkg/hooks/0.1.0/hook.go
flouthoc: Lets do this since consensus is here.

flouthoc:
> Lets put the hooks handling into containers/common.

PR for this: https://github.com/containers/common/pull/1071

buildah.go
```diff
@@ -344,6 +344,8 @@ type BuilderOptions struct {
 	ProcessLabel string
 	// MountLabel is the SELinux mount label associated with the container
 	MountLabel string
+	// OCIHooksDir is the location of OCI hooks for the build containers
+	OCIHooksDir []string
```
nalind, line 348: This duplicates `CommonBuildOpts.OCIHooksDir`.

flouthoc: yes this is not needed.

rhatdan:
@flouthoc needs a rebase.

flouthoc:
> @flouthoc needs a rebase.

Yes rebased.

giuseppe (approved):
LGTM

PR #4077, opened 2022-06-24
[CI:BUILD] Install latest buildah instead of compiling

cevich:
#### What type of PR is this?

/kind other

When building the multi-arch "upstream" flavor of buildah container
images, it's more optimal to use separate compilation and image
construction steps.  The image-build automation is time-limited,
and operating under a (slow) emulation environment.  So using a
continuously pre-built buildah RPM will also improve build
reliability.

ref: https://github.com/containers/buildah/pull/4062

#### How to verify it

The manually-triggered 'upstream' container image build will complete successfully.  No fatal errors will appear in the build log.

None

```release-note
None
```

### DIFF
--- a/contrib/buildahimage/README.md
+++ b/contrib/buildahimage/README.md
@@ -32,7 +32,9 @@ The container images are:
   * `quay.io/buildah/upstream:latest` - This image is built daily using the latest
     code found in this GitHub repository.  Due to the image changing frequently,
     it's not guaranteed to be stable or even executable.  The image is built with
-    [the upstream Dockerfile](upstream/Dockerfile).
+    [the upstream Dockerfile](upstream/Dockerfile).  Note: The actual compilation
+    of upstream buildah [occurs continuously in
+    COPR](https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/).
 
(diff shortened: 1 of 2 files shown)


nalind:
The "upstream" image is described both in `contrib/buildahimage/README.md`, and at the top of the file being patched, as being built from the source repository.  How would the result of this change differ from the "stable" image?

cevich:
Thanks @nalind for checking, I neglected to even look at the docs :disappointed: Based on what you say, they're still technically correct.  It's just that the compilation is happening "out-of-band" to the image build.  Let me actually look at the `README.md` and see how it can be made more clear (and I bet the same issue is present for upstream-podman's readme).

cevich:
Okay, I've staged a change similar to https://github.com/containers/buildah/pull/4077

Waiting to confirm the image builds are successful before pushing.

cevich:
Force-push: Rebased on main.

rhatdan:
LGTM
@flouthoc @nalind @vrothberg @giuseppe @umohnani8 @ashley-cui PTAL

flouthoc (approved):
LGTM
/lgtm

PR #4078, opened 2022-06-24
Add subid support with BuildRequires and BUILDTAG

mhjacks:
#### What type of PR is this?

 /kind feature

enable the use of libsubid to generalize newIDMapping (i.e. use either /etc/sub{g,u}id files OR the centralized FreeIPA ones if available

#### How to verify it

Build a container with multiple UID's using UID/GID configured both from files (/etc/sub{u,g}id) and FreeIPA (subid: sss in /etc/nsswitch.conf). Both should work. Currently the FreeIPA based mechanism fails

None

#### Does this PR introduce a user-facing change? No

```release-note
This change introduces support to make use of the recently fixed subid library check for Fedora and EL-family builds.
```

### DIFF
--- a/buildah.spec.rpkg
+++ b/buildah.spec.rpkg
@@ -52,6 +52,7 @@ BuildRequires: gpgme-devel
 BuildRequires: libassuan-devel
 BuildRequires: make
 BuildRequires: ostree-devel
+BuildRequires: shadow-utils-subid-devel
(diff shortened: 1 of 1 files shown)


lsm5 (approved):
LGTM

lsm5:
@mhjacks could you please re-push the commit with `[NO NEW TESTS NEEDED]` added in the commit message body?

PR #4086, opened 2022-07-01
add: fail on bad `HTTP` response instead of writing to container for `URL` sources 

flouthoc:

```Dockerfile
```

**Fixes: BZ#2102140**

### DIFF
--- a/add.go
+++ b/add.go
@@ -88,6 +88,11 @@ func getURL(src string, chown *idtools.IDPair, mountpoint, renameTarget string, 
 		return err
 	}
 	defer response.Body.Close()
+
+	if response.StatusCode < http.StatusOK || response.StatusCode >= http.StatusBadRequest {
+		return fmt.Errorf("invalid response status %d", response.StatusCode)
+	}
+
 	// Figure out what to name the new content.
 	name := renameTarget
 	if name == "" {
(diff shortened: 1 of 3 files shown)


rhatdan:
@giuseppe @nalind @vrothberg @umohnani8 PTAL

add.go
```diff
@@ -88,6 +88,11 @@ func getURL(src string, chown *idtools.IDPair, mountpoint, renameTarget string,
 		return err
 	}
 	defer response.Body.Close()
+
+	if response.StatusCode < 200 || response.StatusCode >= 400 {
```
vrothberg, line 92: Can we use constants from the http package instead (https://pkg.go.dev/net/http#pkg-constants)?

add.go
```diff
@@ -88,6 +88,11 @@ func getURL(src string, chown *idtools.IDPair, mountpoint, renameTarget string,
...
+		return errors.Errorf("invalid response status %d", response.StatusCode)
```
vrothberg, line 93:
```suggestion
		return fmt.Errorf("invalid response status %d", response.StatusCode)
```
Could we add the URL to the error message as well?

vrothberg (commented):
Code LGTM, only nits.

nalind:
There's probably a similar fix needed in `define.downloadToDirectory()`.

flouthoc:
@nalind Could you PTAL again.

nalind:
LGTM
/lgtm
/hold

PR #4095, opened 2022-07-07
vendor/containers/image: use "io.ReadAll" instead of "os.ReadAll"

rockdrilla:
#### What type of PR is this?

/kind bug

Fixes FTBFS with go 1.18.3

Offending commit: "Update vendor of containers/(common,storage,image)" eb5b6ee8

PR in upstream package: https://github.com/containers/image/pull/1600

#### How to verify it

Build with tag "containers_image_ostree".

None

None.

Signed-off-by: Konstantin Demin <rockdrilla@gmail.com>

### DIFF
--- a/vendor/github.com/containers/image/v5/ostree/ostree_src.go
+++ b/vendor/github.com/containers/image/v5/ostree/ostree_src.go
@@ -368,7 +368,7 @@ func (s *ostreeImageSource) GetSignatures(ctx context.Context, instanceDigest *d
 		}
 		defer sigReader.Close()
 
-		sig, err := os.ReadAll(sigReader)
+		sig, err := io.ReadAll(sigReader)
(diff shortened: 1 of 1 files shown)


flouthoc (changes requested):
Hi @rockdrilla , Thanks for creating the PR.

Actually I think you should wait for https://github.com/containers/image/pull/1600 and once that gets merged update `go.mod` to point to the relevant version of `c/image` instead of directly editing the `vendor` directory.

rhatdan:
Yes this needs to be fixed in containers/image not in buildah.

rhatdan:
@rockdrilla open a PR to vendor in the latest containers/image if  you want to complete this.

BTW This should be happenning next week regardless, since we are planning a release.

PR #1132, opened 2018-10-27
Handle directories better in bud -f

TomSweeneyRedHat:
Signed-off-by: TomSweeneyRedHat <tsweeney@redhat.com>

A number of other commits addressed #1104, this finishes that up.  If a directory was given in the --file parameter for `buildah bud` a panic would occur.  With this change if a directory only is given as the argument, the code will append '/Dockerfile' to it and attempt to read the file.  If it can't read the file, it fails with an appropriate error.  If it finds the file, it processes it as appropriate.

Testing:

```
# ls tools/Dockerfiles
Dockerfile.hold

#  buildah bud -t tom -f 'tools/Dockerfiles' .
error reading "tools/Dockerfiles/Dockerfile": open tools/Dockerfiles/Dockerfile: no such file or directory

# cp tools/Dockerfiles/Dockerfile.hold tools/Dockerfiles/Dockerfile
#  buildah bud -t tom -f 'tools/Dockerfiles' .
STEP 1: FROM alpine
STEP 2: COMMIT containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.override_kernel_check=true]localhost/tom:latest
Getting image source signatures
Skipping fetch of repeat blob sha256:df64d3292fd6194b7865d7326af5255db6d81e9df29f48adde61a918fbd8c332
Skipping fetch of repeat blob sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1
Copying config sha256:bb64f9b02d069cc1fa5534d3ca9703c5bbe99467165ac75ca24a8809435a8bd6
 702 B / 702 B [============================================================] 0s
Writing manifest to image destination
Storing signatures
--> bb64f9b02d069cc1fa5534d3ca9703c5bbe99467165ac75ca24a8809435a8bd6

# buildah bud -t tom -f "reallybaddir" .
error reading info about "/root/tsweeney/workspaces/buildah/src/github.com/containers/buildah/reallybaddir": stat /root/tsweeney/workspaces/buildah/src/github.com/containers/buildah/reallybaddir: no such file or directory
```

PR #4096, opened 2022-07-07
imagebuildah, BuildDockerfiles: error out if path to `containerfile` is a `directory`

flouthoc:
PR https://github.com/containers/buildah/pull/1132 added a support for
`buildah build -f <path-to-dir>` but it was decided that `-f` should
only support file and not a directory hence error out for such use-case.

Change matches with behavior of docker and buildkit
Reason for this change is decided here: https://github.com/containers/podman/pull/14838

Closes: https://github.com/containers/podman/issues/14836


rhatdan:
LGTM

Issue #4093, opened 2022-07-06
Empty images created with overlay mount


nhojjohn:
**Description**

I've have been building container images using buildah.  I recently started doing this on a different server and found my images were empty (or more accurately 2.34kb).  The failure appears to be silent.

My example steps are to run the following within "buildah unshare"
```
CONT=$(buildah from scratch)
MNT=$(buildah mount $CONT)
# create some content
dd if=/dev/zero of=$MNT/file count=1 bs=10M
buildah umount $CONT
buildah --log-level debug commit $CONT foo
buildah images foo
```
This should show a image with the added content
```
REPOSITORY      TAG      IMAGE ID       CREATED        SIZE
localhost/foo   latest   4f1b787b1b5a   1 second ago   10.5 MB
```

instead I get an empty container
```
REPOSITORY      TAG      IMAGE ID       CREATED         SIZE
localhost/foo   latest   e7f614869b5b   2 seconds ago   2.34 KB
```

When I reconfigure podman storage.comf to use fuse-overlayfs it works as before.

But additionally when I run the same build *on the same server and container config* as my interactive user it also works as before without modifying storage.conf.  The failures happen when running as  "build" user  which I sudo to (and also runs github actions-runner). Comparing the output of buildah system between the two users the config seems to be identical.

The main difference I can see is that the empty images are created by a user I sudo to.  The logs show the following differences of overlay path
```DEBU[0000] parsed reference into "[overlay@/home/build/.local/share/containers/storage+/var/tmp/containers-user-1007/containers]localhost/foo:latest" ```
to
```DEBU[0000] parsed reference into "[overlay@/home/john/.local/share/containers/storage+/run/user/1002/containers]localhost/foo:latest" ```

This is presumably a result in the difference in login?

The former failed run also showing the wrong number of bytes
```DEBU[0000] layer "ba949599f2947a8a4cdfe40687cf305575b5aed637b6ba6e04848fa1f1947fa5" size is 1024 bytes, uncompressed digest sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef, possibly-compressed digest sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef ```

The other notable difference is the successful run finishes with
```INFO[0000] failed to shutdown storage: "A layer is mounted: layer is in use by a container"``` whilst the failing one doesn't print this.

[...]

PR #4101, opened 2022-07-11
buildah: set missing `XDG_RUNTIME_DIR` before setting default `runroot`

flouthoc:
In sessions which are rootless and contains no valid login session
`XDG_RUNTIME_DIR` is not set for such use-case `getStore(` automatically
sets `XDG_RUNTIME_DIR` just move it before we set default `runroot` so we
end up populating correcting `runroot` for cases which are rootless and
no valid login is present.

Missing `XDG_RUNTIME_DIR` causes various parts of code to point at incorrect
runroot and thus causing missing checks of a layer is already mounted or not etc.

Closes: https://github.com/containers/buildah/issues/4093

Similar fixes were attempted before for `login`: https://github.com/containers/buildah/pull/2145
I think above PR is a similar generic fix for all commands.

### DIFF
--- a/cmd/buildah/common.go
+++ b/cmd/buildah/common.go
@@ -25,6 +25,9 @@ var (
 )
 
 func getStore(c *cobra.Command) (storage.Store, error) {
+	if err := setXDGRuntimeDir(); err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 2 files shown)


flouthoc:
@UlrichLohrmann @nhojjohn PTAL would be nice if you could also try the patch.

flouthoc:

PR #4102, opened 2022-07-11
GHA: Remove required reserved-name use

cevich:
#### What type of PR is this?

/kind other

Unfortunately github-actions now blacklists `ACTIONS_STEP_DEBUG` from
the list of secrets.  Resulting in the error:

```
Invalid workflow file: .github/workflows/check_cirrus_cron.yml#L25
secret name `ACTIONS_STEP_DEBUG` within `workflow_call` can not be used
since it would collide with system reserved name
```

This is unfortunate since overriding this with a "secret" blocks it's
use by users seeking to modify workflows to assist in circumventing
security protections.  Fix the workflow failures by removing references
to this variable name.  The benefits of reliable operation of this
specific monitoring workflow likely outweigh any security gains of
blocking debug-mode.

#### How to verify it

Manual review and post-merge followup.  Functionality cannot be determined in a PR, it must be merged then reviewed :disappointed: 

None

Address ongoing workflow failures such as: https://github.com/containers/buildah/actions/runs/2648436112

```release-note
None
```

### DIFF
--- a/.github/workflows/check_cirrus_cron.yml
+++ b/.github/workflows/check_cirrus_cron.yml
@@ -21,9 +21,6 @@ on:
     # Ref: https://docs.github.com/en/actions/using-workflows/reusing-workflows
     workflow_call:
       secrets:
-        # Debug-mode can reveal secrets, force it to be secrets-managed.
-        ACTIONS_STEP_DEBUG:
-          required: true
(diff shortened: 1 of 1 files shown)


cevich:
Looks like the problem is fixed.

PR #4114, opened 2022-07-15
Add FreeBSD support for 'buildah info'

dfr:
On non-windows platforms, we can use the uname syscall to get kernel
version information. For uptime, we can use clock_gettime(CLOCK_UPTIME)
on FreeBSD but I don't have a solution for darwin or windows.

Note: the existing code which unconditionally reads from /proc builds
on all platforms but only works on linux since using /proc in this way
is a linux feature.

[NO NEW TESTS NEEDED]

#### What type of PR is this?

/kind cleanup

The info sub-command is used by buildah tests so fixing it is a precursor to getting some of those tests working on FreeBSD

None

### DIFF
--- a/info.go
+++ b/info.go
@@ -3,16 +3,14 @@ package buildah
 import (
 	"bufio"
 	"bytes"
-	"errors"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"runtime"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/containerd/containerd/platforms"
+	putil "github.com/containers/buildah/pkg/util"
 	"github.com/containers/buildah/util"
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/system"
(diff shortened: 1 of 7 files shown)


rhatdan:
Was linux missing it or Darwin (MAC)?

dfr:
Both, as it turned out. Apparently CLOCK_UPTIME is a FreeBSD invention.

rhatdan:
This seems fine, but I have a feeling Podman would like to use this also.  So should we put this into pkg/util or container/common?

dfr:
Podman does have a similar pattern and I think it makes sense to try to
share the code. Given that podman imports buildah, putting the code in a
package in buildah would be simplest and I can make the corresponding
change in libpod next time vendor/github.com/containers/buildah is
refreshed.

dfr:
Done. This reminded me of the confusing existence of both buildah/util and buildah/pkg/util but I'm ignoring that for now. Possibly merge the two into buildah/pkg/until one day.

rhatdan:
Yes I think the existence of those is confusing also.
LGTM
/approve
@flouthoc @giuseppe @nalind @umohnani8 @ashley-cui PTAL

flouthoc (approved):
LGTM
/lgtm

Issue #620, opened 2018-04-24
buildah bud --cache-from functionality needed


TomSweeneyRedHat:
**Description**
`buildah bud --cache-from` is missing and needs to be provided.  This is functionality that Docker has.

#599 is the parent issue for this issue.

**Output of `buildah version`:**

Buildah v0.16

rhatdan:
We don't currently support cache, so this one might take a while.

rhatdan:
Now that we have support for --cache, @umohnani8 Could you add this functionality.

rhatdan:
@TomSweeneyRedHat @umohnani8 Is this fixed in buildah?

TomSweeneyRedHat:
It's still a no-op in the code with nothing really happening underneath it other than saying "not supported".

rhatdan:
is it something we could implement?

TomSweeneyRedHat:
I think we could get --cache-from to work for "real", but I've not dived through the code to see if we can for sure.

rhatdan:
@rhatdan Any movement on this one?

rhatdan:
@ashley-cui Could you take a look at this?

mfolnovic:
Parent issue ( #599 ) was closed, so I was wondering if this will be implemented?

This has a huge impact on building images from CI.

Unfortunately this issue is the only reason I prefer Docker in CI (while living with issues Docker has). After implementing this issue, I don't see any reason to not use Buildah in CI.

Otherwise, Buildah is really awesome, nice work!

rhatdan:
Sadly no one has picked up the work on this one yet.
@nalind WDYT?
@TomSweeneyRedHat Do you have time to look into this?
@mfolnovic Any chance you would want to take a stab?

mfolnovic:
@rhatdan with a few pointers, I could try :) :)

TeNNoX:
What exactly do you mean by pointers?
Implementation hints? or functionality-wise?

mfolnovic:
@TeNNoX implementation hints :)

haampie:
Any news on this issue?

TomSweeneyRedHat:
@haampie I'm afraid I've not had time to look into this, I've been chasing other fires and I don't believe anyone else has.

rhatdan:
@sujil02 Would you like to take a look at this?

sujil02:
> @sujil02 Would you like to take a look at this?

Sure. I will give it a shot.

haampie:
[...]

PR #3406, opened 2021-07-27
LayerKey model for the local cache

george-harness:
#### What type of PR is this?

/kind other

#### What this PR does / why we need it:
This changes the algorithm to determine if the next build command is already in the local store. Instead of enumerating the layers one by one until it finds a hit, the new approach calculates a key that represents the layer. This will allow the system to be extended with a distributed cache. 

#### How to verify it

None of the existing tests should change. New unit tests were added, they should not fail as well.

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

No

#### Does this PR introduce a user-facing change?
None

PR #4118, opened 2022-07-18
build,cache: support remote cache sources ( via `--cache-to`, `--cache-from` ) and introduce `cacheKey/layerKey`

flouthoc:
Following commit

* Initiates `cacheKey` or `layerKey` for intermediate images generated
  for layers. ( only used when working with remote sources )
* Allows end users to upload cached layers with `cacheKey` to remote
  sources using `--cache-to`. `--cache-to` is a optional flag to be used
with `buildah build` which publishes cached layers to remote sources.
* Allows end users to use cached layers from `remote` sources with
  `--cache-from`. `--cache-from` is a optional flag to be used with
`buildah build` and it pulls cached layers from remote sources in a step
by step manner only if is a valid cache hit.

Example
* Populate cache source or use cached layers if already present
```bash
buildah build -t test --layers --cache-to registry/myrepo/cache --cache-from registry/myrepo/cache .
```

Future Improvements:
* Also support pushing cache to other storage units rather than conventional container registry like **object storage ( AWS S3, Redis, Minio etc )**
* `cacheKey` or `layerKey` model is only being used when working with
  remote sources however local cache lookup can be also optimized if its
is altered to use `cacheKey` model instead of iterating through all the
images in local storage. As discussed here: https://github.com/containers/buildah/pull/3406

References:
* Feature is quite similar to `kaniko`'s `--cache-repo`: https://github.com/GoogleContainerTools/kaniko#--cache-repo

Closes: https://github.com/containers/buildah/issues/620

```release-note
build: adds supports for --cache-to i.e alllow pushing cache layers to remote source
build: adds support for --cache-from i.e allow using cache layers from remote sources
```

### DIFF
--- a/define/build.go
+++ b/define/build.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	nettypes "github.com/containers/common/libnetwork/types"
+	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/types"
 	encconfig "github.com/containers/ocicrypt/config"
 	"github.com/containers/storage/pkg/archive"
@@ -136,6 +137,12 @@ type BuildOptions struct {
 	RuntimeArgs []string
 	// TransientMounts is a list of mounts that won't be kept in the image.
 	TransientMounts []string
+	// CacheFrom specifies any remote repository which can be treated as
+	// potential cache source.
+	CacheFrom reference.Named
+	// CacheTo specifies any remote repository which can be treated as
+	// potential cache destination.
+	CacheTo reference.Named
 	// Compression specifies the type of compression which is applied to
 	// layer blobs.  The default is to not use compression, but
 	// archive.Gzip is recommended.
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -96,7 +96,37 @@ The value of `[name]` is matched with the following priority order:
 
 **--cache-from**
 
-Images to utilise as potential cache sources. Buildah does not currently support --cache-from so this is a NOOP.
+Repository to utilize as a potential cache source. When specified, Buildah will try to look for
+cache images in the specified repository and will attempt to pull cache images instead of actually
+executing the build steps locally. Buildah will only attempt to pull previously cached images if they
+are considered as valid cache hits.
+
+Use the `--cache-to` option to populate a remote repository with cache content.
+
+Example
+
+```bash
+# populate a cache and also consult it
+buildah build -t test --layers --cache-to registry/myrepo/cache --cache-from registry/myrepo/cache .
+```
+
+Note: `--cache-from` option is ignored unless `--layers` is specified.
+
+**--cache-to**
+
+Set this flag to specify a remote repository that will be used to store cache images. Buildah will attempt to
+push newly built cache image to the remote repository.
+
+Note: Use the `--cache-from` option in order to use cache content in a remote repository.
+
+Example
+
+```bash
+# populate a cache and also consult it
+buildah build -t test --layers --cache-to registry/myrepo/cache --cache-from registry/myrepo/cache .
+```
+
+Note: `--cache-to` option is ignored unless `--layers` is specified.
 
 **--cap-add**=*CAP\_xxx*
 
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -2,6 +2,7 @@ package imagebuildah
 
 import (
 	"context"
+	"crypto/sha256"
 	"fmt"
 	"io"
 	"os"
@@ -22,6 +23,7 @@ import (
 	"github.com/containers/buildah/util"
 	config "github.com/containers/common/pkg/config"
 	cp "github.com/containers/image/v5/copy"
+	imagedocker "github.com/containers/image/v5/docker"
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/manifest"
 	is "github.com/containers/image/v5/storage"
@@ -961,6 +963,22 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			s.log(commitMessage)
 		}
 	}
+	// logCachePulled produces build log for cases when `--cache-from`
+	// is used and a valid intermediate image is pulled from remote source.
+	logCachePulled := func(cacheKey string) {
+		if !s.executor.quiet {
+			cacheHitMessage := "--> Cache pulled from remote"
+			fmt.Fprintf(s.executor.out, "%s %s\n", cacheHitMessage, fmt.Sprintf("%s:%s", s.executor.cacheFrom, cacheKey))
+		}
+	}
+	// logCachePush produces build log for cases when `--cache-to`
+	// is used and a valid intermediate image is pushed tp remote source.
+	logCachePush := func(cacheKey string) {
+		if !s.executor.quiet {
+			cacheHitMessage := "--> Pushing cache"
+			fmt.Fprintf(s.executor.out, "%s %s\n", cacheHitMessage, fmt.Sprintf("%s:%s", s.executor.cacheTo, cacheKey))
+		}
+	}
 	logCacheHit := func(cacheID string) {
 		if !s.executor.quiet {
 			cacheHitMessage := "--> Using cache"
@@ -1153,11 +1171,15 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 		var (
 			commitName                string
 			cacheID                   string
+			cacheKey                  string
+			pulledAndUsedCacheImage   bool
 			err                       error
 			rebase                    bool
 			addedContentSummary       string
 			canMatchCacheOnlyAfterRun bool
 		)
+
+		needsCacheKey := (s.executor.cacheFrom != nil || s.executor.cacheTo != nil)
 
 		// If we have to commit for this instruction, only assign the
 		// stage's configured output name to the last layer.
@@ -1165,6 +1187,15 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			commitName = s.output
 		}
 
+		// If --cache-from or --cache-to is specified make sure to populate
+		// cacheKey since it will be used either while pulling or pushing the
+		// cache images.
+		if needsCacheKey {
+			cacheKey, err = s.generateCacheKey(ctx, node, addedContentSummary, s.stepRequiresLayer(step))
+			if err != nil {
+				return "", nil, fmt.Errorf("failed while generating cache key: %w", err)
+			}
+		}
 		// Check if there's already an image based on our parent that
 		// has the same change that we're about to make, so far as we
 		// can tell.
@@ -1186,10 +1217,34 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 				// Retrieve the digest info for the content that we just copied
 				// into the rootfs.
 				addedContentSummary = s.getContentSummaryAfterAddingContent()
+				// regenerate cache key with updated content summary
+				if needsCacheKey {
+					cacheKey, err = s.generateCacheKey(ctx, node, addedContentSummary, s.stepRequiresLayer(step))
+					if err != nil {
+						return "", nil, fmt.Errorf("failed while generating cache key: %w", err)
+					}
+				}
 			}
 			cacheID, err = s.intermediateImageExists(ctx, node, addedContentSummary, s.stepRequiresLayer(step))
 			if err != nil {
 				return "", nil, fmt.Errorf("error checking if cached image exists from a previous build: %w", err)
+			}
+			// All the best effort to find cache on localstorage have failed try pulling
+			// cache from remote repo if `--cache-from` was configured.
+			if cacheID == "" && s.executor.cacheFrom != nil {
+				// only attempt to use cache again if pulling was successful
+				// otherwise do nothing and attempt to run the step, err != nil
+				// is ignored and will be automatically logged for --log-level debug
+				if id, err := s.pullCache(ctx, cacheKey); id != "" && err == nil {
+					logCachePulled(cacheKey)
+					cacheID, err = s.intermediateImageExists(ctx, node, addedContentSummary, s.stepRequiresLayer(step))
+					if err != nil {
+						return "", nil, fmt.Errorf("error checking if cached image exists from a previous build: %w", err)
+					}
+					if cacheID != "" {
+						pulledAndUsedCacheImage = true
+					}
+				}
 			}
 		}
 
@@ -1208,6 +1263,13 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 
 			// In case we added content, retrieve its digest.
 			addedContentSummary = s.getContentSummaryAfterAddingContent()
+			// regenerate cache key with updated content summary
+			if needsCacheKey {
+				cacheKey, err = s.generateCacheKey(ctx, node, addedContentSummary, s.stepRequiresLayer(step))
+				if err != nil {
+					return "", nil, fmt.Errorf("failed while generating cache key: %w", err)
+				}
+			}
 
 			// Check if there's already an image based on our parent that
 			// has the same change that we just made.
@@ -1266,6 +1328,23 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			imgID, ref, err = s.commit(ctx, s.getCreatedBy(node, addedContentSummary), !s.stepRequiresLayer(step), commitName, false)
 			if err != nil {
 				return "", nil, fmt.Errorf("error committing container for step %+v: %w", *step, err)
+			}
+		}
+
+		// Following step is just built and was not used from
+		// cache so check if --cache-to was specified if yes
+		// then attempt pushing this cache to remote repo and
+		// fail accordingly.
+		//
+		// Or
+		//
+		// Try to push this cache to remote repository only
+		// if cache was present on local storage and not
+		// pulled from remote source while processing this
+		if s.executor.cacheTo != nil && (!pulledAndUsedCacheImage || cacheID == "") {
+			logCachePush(cacheKey)
+			if err = s.pushCache(ctx, imgID, cacheKey); err != nil {
+				return "", nil, err
 			}
 		}
 
@@ -1540,6 +1619,114 @@ func (s *StageExecutor) tagExistingImage(ctx context.Context, cacheID, output st
 		}
 	}
 	return img.ID, ref, nil
+}
+
+// generateCacheKey returns a computed digest for the current STEP
+// running its history and diff against a hash algorithm and this
+// generated CacheKey is further used by buildah to lock and decide
+// tag for the intermeidate image which can be pushed and pulled to/from
+// the remote repository.
+func (s *StageExecutor) generateCacheKey(ctx context.Context, currNode *parser.Node, addedContentDigest string, buildAddsLayer bool) (string, error) {
+	hash := sha256.New()
+	var baseHistory []v1.History
+	var diffIDs []digest.Digest
+	var manifestType string
+	var err error
+	if s.builder.FromImageID != "" {
+		manifestType, baseHistory, diffIDs, err = s.executor.getImageTypeAndHistoryAndDiffIDs(ctx, s.builder.FromImageID)
+		if err != nil {
+			return "", fmt.Errorf("error getting history of base image %q: %w", s.builder.FromImageID, err)
+		}
+		for i := 0; i < len(diffIDs); i++ {
+			fmt.Fprintln(hash, diffIDs[i].String())
+		}
+	}
+	createdBy := s.getCreatedBy(currNode, addedContentDigest)
+	fmt.Fprintf(hash, "%t", buildAddsLayer)
+	fmt.Fprintln(hash, createdBy)
+	fmt.Fprintln(hash, manifestType)
+	for _, element := range baseHistory {
+		fmt.Fprintln(hash, element.CreatedBy)
+		fmt.Fprintln(hash, element.Author)
+		fmt.Fprintln(hash, element.Comment)
+		fmt.Fprintln(hash, element.Created)
+		fmt.Fprintf(hash, "%t", element.EmptyLayer)
+		fmt.Fprintln(hash)
+	}
+	return fmt.Sprintf("%x", hash.Sum(nil)), nil
+}
+
+// cacheImageReference is internal function which generates ImageReference from Named repo sources
+// and a tag.
+func cacheImageReference(repo reference.Named, cachekey string) (types.ImageReference, error) {
+	tagged, err := reference.WithTag(repo, cachekey)
+	if err != nil {
+		return nil, fmt.Errorf("failed generating tagged reference for %q: %w", repo, err)
+	}
+	dest, err := imagedocker.NewReference(tagged)
+	if err != nil {
+		return nil, fmt.Errorf("failed generating docker reference for %q: %w", tagged, err)
+	}
+	return dest, nil
+}
+
+// pushCache takes the image id of intermediate image and attempts
+// to perform push at the remote repository with cacheKey as the tag.
+// Returns error if fails otherwise returns nil.
+func (s *StageExecutor) pushCache(ctx context.Context, src, cacheKey string) error {
+	dest, err := cacheImageReference(s.executor.cacheTo, cacheKey)
+	if err != nil {
+		return err
+	}
+	logrus.Debugf("trying to push cache to dest: %+v from src:%+v", dest, src)
+	options := buildah.PushOptions{
+		Compression:         s.executor.compression,
+		SignaturePolicyPath: s.executor.signaturePolicyPath,
+		Store:               s.executor.store,
+		SystemContext:       s.executor.systemContext,
+		BlobDirectory:       s.executor.blobDirectory,
+		SignBy:              s.executor.signBy,
+		MaxRetries:          s.executor.maxPullPushRetries,
+		RetryDelay:          s.executor.retryPullPushDelay,
+	}
+	ref, digest, err := buildah.Push(ctx, src, dest, options)
+	if err != nil {
+		return fmt.Errorf("failed pushing cache to %q: %w", dest, err)
+	}
+	logrus.Debugf("successfully pushed cache to dest: %+v with ref:%+v and digest: %v", dest, ref, digest)
+	return nil
+}
+
+// pullCache takes the image source of the cache assuming tag
+// already points to the valid cacheKey and pulls the image to
+// local storage only if it was not already present on local storage
+// or a newer version of cache was found in the upstream repo. If new
+// image was pulled function returns image id otherwise returns empty
+// string "" or error if any error was encontered while pulling the cache.
+func (s *StageExecutor) pullCache(ctx context.Context, cacheKey string) (string, error) {
+	src, err := cacheImageReference(s.executor.cacheFrom, cacheKey)
+	if err != nil {
+		return "", err
+	}
+	logrus.Debugf("trying to pull cache from remote repo: %+v", src.DockerReference())
+	options := buildah.PullOptions{
+		SignaturePolicyPath: s.executor.signaturePolicyPath,
+		Store:               s.executor.store,
+		SystemContext:       s.executor.systemContext,
+		BlobDirectory:       s.executor.blobDirectory,
+		MaxRetries:          s.executor.maxPullPushRetries,
+		RetryDelay:          s.executor.retryPullPushDelay,
+		AllTags:             false,
+		ReportWriter:        nil,
+		PullPolicy:          define.PullIfNewer,
+	}
+	id, err := buildah.Pull(ctx, src.DockerReference().String(), options)
+	if err != nil {
+		logrus.Debugf("failed pulling cache from source %s: %v", src, err)
+		return "", fmt.Errorf("failed while pulling cache from %q: %w", src, err)
+	}
+	logrus.Debugf("successfully pulled cache from repo %s: %s", src, id)
+	return id, nil
 }
 
 // intermediateImageExists returns true if an intermediate image of currNode exists in the image store from a previous build.
--- a/pkg/cli/build.go
+++ b/pkg/cli/build.go
@@ -19,6 +19,7 @@ import (
 	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/buildah/pkg/util"
 	"github.com/containers/common/pkg/auth"
+	"github.com/containers/image/v5/docker/reference"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
@@ -233,10 +234,6 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
 		return options, nil, nil, errors.New("'rm' and 'force-rm' can only be set with either 'layers' or 'no-cache'")
 	}
 
-	if c.Flag("cache-from").Changed {
-		logrus.Debugf("build --cache-from not enabled, has no effect")
-	}
-
 	if c.Flag("compress").Changed {
 		logrus.Debugf("--compress option specified but is ignored")
 	}
@@ -290,6 +287,22 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
 			iopts.Quiet = true
 		}
 	}
+	var cacheTo reference.Named
+	var cacheFrom reference.Named
+	cacheTo = nil
+	cacheFrom = nil
+	if c.Flag("cache-to").Changed {
+		cacheTo, err = parse.RepoNameToNamedReference(iopts.CacheTo)
+		if err != nil {
+			return options, nil, nil, fmt.Errorf("unable to parse value provided `%s` to --cache-to: %w", iopts.CacheTo, err)
+		}
+	}
+	if c.Flag("cache-from").Changed {
+		cacheFrom, err = parse.RepoNameToNamedReference(iopts.CacheFrom)
+		if err != nil {
+			return options, nil, nil, fmt.Errorf("unable to parse value provided `%s` to --cache-from: %w", iopts.CacheTo, err)
+		}
+	}
 	options = define.BuildOptions{
 		AddCapabilities:         iopts.CapAdd,
 		AdditionalBuildContexts: additionalBuildContext,
@@ -300,6 +313,8 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
 		Args:                    args,
 		BlobDirectory:           iopts.BlobCache,
 		BuildOutput:             iopts.BuildOutput,
+		CacheFrom:               cacheFrom,
+		CacheTo:                 cacheTo,
 		CNIConfigDir:            iopts.CNIConfigDir,
 		CNIPluginPath:           iopts.CNIPlugInPath,
 		CPPFlags:                iopts.CPPFlags,
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -19,6 +19,7 @@ import (
 	internalParse "github.com/containers/buildah/internal/parse"
 	"github.com/containers/buildah/pkg/sshagent"
 	"github.com/containers/common/pkg/parse"
+	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/types"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/unshare"
@@ -47,6 +48,18 @@ const (
 	// Lifecycle of following directory will be inherited from how host machine treats temporary directory
 	BuildahCacheDir = "buildah-cache"
 )
+
+// RepoNameToNamedReference parse the raw string to Named reference
+func RepoNameToNamedReference(dest string) (reference.Named, error) {
+	named, err := reference.ParseNormalizedNamed(dest)
+	if err != nil {
+		return nil, fmt.Errorf("invalid repo %q: must contain registry and repository: %w", dest, err)
+	}
+	if !reference.IsNameOnly(named) {
+		return nil, fmt.Errorf("repository must contain neither a tag nor digest: %v", named)
+	}
+	return named, nil
+}
 
 // CommonBuildOptions parses the build options from the bud cli
 func CommonBuildOptions(c *cobra.Command) (*define.CommonBuildOptions, error) {
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -3864,6 +3864,112 @@ _EOF
   run_buildah rmi myalpine
 }
 
+@test "build test pushing and pulling from remote cache sources" {
+  _prefetch alpine
+  mytmpdir=${TEST_SCRATCH_DIR}/my-dir
+  mkdir -p ${mytmpdir}
+  echo something > ${mytmpdir}/somefile
+  cat > $mytmpdir/Containerfile << _EOF
+FROM alpine
+RUN echo hello
+RUN echo world
+RUN touch hello
+ADD somefile somefile
+
+FROM alpine
+RUN echo hello
+COPY --from=0 hello hello
+_EOF
+
+  start_registry
+  run_buildah login --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth --username testuser --password testpassword localhost:${REGISTRY_PORT}
+
+  # ------ Test case ------ #
+  # prepare expected output beforehand
+  # must push cache twice i.e for first step and second step
+  run printf "STEP 2/5: RUN echo hello\nhello\n--> Pushing cache"
+  step1=$output
+  run printf "STEP 3/5: RUN echo world\nworld\n--> Pushing cache"
+  step2=$output
+  run printf "STEP 4/5: RUN touch hello\n--> Pushing cache"
+  step3=$output
+  run printf "STEP 5/5: ADD somefile somefile\n--> Pushing cache"
+  step4=$output
+  # First run step in second stage should not be pushed since its already pushed
+  run printf "STEP 2/3: RUN echo hello\n--> Using cache"
+  step5=$output
+  # Last step is `COPY --from=0 hello hello' so it must be commited and pushed
+  # actual output is `[2/2] STEP 3/3: COPY --from=0 hello hello\n[2/2] COMMIT test\n-->Pushing cache`
+  # but lets match smaller suffix
+  run printf "COMMIT test\n--> Pushing cache"
+  step6=$output
+
+  # actually run build
+  run_buildah build $WITH_POLICY_JSON --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth --layers --cache-to localhost:${REGISTRY_PORT}/temp -t test -f ${mytmpdir}/Containerfile ${mytmpdir}
+  expect_output --substring "$step1"
+  expect_output --substring "$step2"
+  expect_output --substring "$step3"
+  expect_output --substring "$step4"
+  expect_output --substring "$step5"
+  expect_output --substring "$step6"
+
+  # clean all cache and intermediate images
+  # to make sure that we are only using cache
+  # from remote repo and not the local storage.
+  run_buildah rmi --all -f
+
+  # ------ Test case ------ #
+  # expect cache to be pushed on remote stream
+  # now a build on clean slate must pull cache
+  # from remote instead of actually computing the
+  # run steps
+  run printf "STEP 2/5: RUN echo hello\n--> Cache pulled from remote"
+  step1=$output
+  run printf "STEP 3/5: RUN echo world\n--> Cache pulled from remote"
+  step2=$output
+  run printf "STEP 4/5: RUN touch hello\n--> Cache pulled from remote"
+  step3=$output
+  run printf "STEP 5/5: ADD somefile somefile\n--> Cache pulled from remote"
+  step4=$output
+  # First run step in second stage should not be pulled since its already pulled
+  run printf "STEP 2/3: RUN echo hello\n--> Using cache"
+  step5=$output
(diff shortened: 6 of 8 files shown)


flouthoc:
@nalind PTAL

docs/buildah-build.1.md
```diff
@@ -96,7 +96,31 @@ The value of `[name]` is matched with the following priority order:
 
 **--cache-from**
 
-Images to utilise as potential cache sources. Buildah does not currently support --cache-from so this is a NOOP.
+Repository to utilise as potential cache source. When specified buildah will try to look for
+cached layers in specified repository and will attempt to pull cached layers instead of actually
```
nalind, line 100: The cache logic doesn't read and write layers by themselves, it always reads and writes images.

imagebuildah/stage_executor.go
```diff
@@ -1168,6 +1193,22 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			if err != nil {
 				return "", nil, fmt.Errorf("error checking if cached image exists from a previous build: %w", err)
 			}
+			// All the best effort to find cache on localstorage have failed try pulling
+			// cache from remote repo if `--cache-from` was configured.
+			if cacheID == "" && s.executor.cacheFrom != "" && cacheKey != "" {
```
nalind, line 1198: `cacheKey` should never be empty if `s.executor.cacheFrom` is set, so no need to check for it here.

imagebuildah/stage_executor.go
```diff
@@ -1246,6 +1287,18 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			if err != nil {
 				return "", nil, fmt.Errorf("error committing container for step %+v: %w", *step, err)
 			}
+			// Following step is just built and was not used from cache so
+			// check if --cache-to was specified if yes then attempt pushing
+			// this cache to remote repo and fail accordingly
+			if cacheKey != "" && cacheID == "" && s.executor.cacheTo != "" {
```
nalind, line 1293: `cacheID` is always empty in this block, so no need to check for it here.
`cacheKey` should never be empty if `s.executor.cacheTo` is set, so no need to check for it here.

imagebuildah/stage_executor.go
```diff
@@ -1521,6 +1574,124 @@ func (s *StageExecutor) tagExistingImage(ctx context.Context, cacheID, output st
...
+	createdBy := s.getCreatedBy(currNode, addedContentDigest)
+	fmt.Fprintf(h, "%t\n", buildAddsLayer)
+	fmt.Fprintln(h, createdBy)
+	fmt.Fprintln(h, manifestType)
+	for _, element := range baseHistory {
+		fmt.Fprintln(h, element.CreatedBy)
+		fmt.Fprintln(h, element.Author)
+		fmt.Fprintln(h, element.Comment)
+		fmt.Fprintf(h, "%t\n", element.EmptyLayer)
+		fmt.Fprintln(h)
```
nalind, line 1602: What does adding a newline do here?

imagebuildah/stage_executor.go
```diff
@@ -1168,6 +1193,22 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			if err != nil {
 				return "", nil, fmt.Errorf("error checking if cached image exists from a previous build: %w", err)
 			}
+			// All the best effort to find cache on localstorage have failed try pulling
+			// cache from remote repo if `--cache-from` was configured.
+			if cacheID == "" && s.executor.cacheFrom != "" && cacheKey != "" {
+				src := fmt.Sprintf("%s:%s", s.executor.cacheFrom, cacheKey)
+				id, err := s.pullCache(ctx, src)
```
nalind, line 1200: `id` and `err` don't get used outside of this block, so merging the `pullCache` call into the `if` statement just below will shorten it up.

imagebuildah/stage_executor.go
```diff
@@ -1246,6 +1287,18 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
 			}
+			// Following step is just built and was not used from cache so
+			// check if --cache-to was specified if yes then attempt pushing
+			// this cache to remote repo and fail accordingly
+			if cacheKey != "" && cacheID == "" && s.executor.cacheTo != "" {
+				// try to push this cache to remote repository
+				destSpec := fmt.Sprintf("%s:%s", s.executor.cacheTo, cacheKey)
+				logCachePush(destSpec)
+				err = s.pushCache(ctx, imgID, destSpec)
+				if err != nil {
```
nalind, line 1298: We can go ahead and merge these two lines.  We checked that `err` was `nil` before we got here, and if `pushCache()` returns an error, we return immediately.

imagebuildah/stage_executor.go
```diff
@@ -1521,6 +1574,124 @@ func (s *StageExecutor) tagExistingImage(ctx context.Context, cacheID, output st
 	return img.ID, ref, nil
 }
 
+// generateCacheKey returns a computed digest for the current STEP
+// running its history and diff against a hash algorithm and this
+// generated CacheKey is further used by buildah to lock and decide
+// tag for the intermeidate image which can be pushed and pulled to/from
+// the remote repository.
+func (s *StageExecutor) generateCacheKey(ctx context.Context, currNode *parser.Node, addedContentDigest string, buildAddsLayer bool) (string, error) {
+	h := sha256.New()
```
rhatdan, line 1583: Can you call this hash?  h is a little vague.

nalind (changes requested):
Looks generally good.
We need to be very precise in use of terminology in documentation.  People already get confused about the difference between containers and images when they're starting out with containers, and we don't want to muddy things any further.
Question: when we have a local cache hit, why don't we try to push to a `--cache-to` location?  The image which counted as a cache hit may not have come from there.

imagebuildah/stage_executor.go
```diff
@@ -1521,6 +1574,124 @@ func (s *StageExecutor) tagExistingImage(ctx context.Context, cacheID, output st
...
+// pushCache takes the image id of intermediate image and attempts
+// to perform push at the remote repository with cacheKey as the tag.
+// Returns error if fails otherwise returns nil.
+func (s *StageExecutor) pushCache(ctx context.Context, src, destSpec string) error {
+	logrus.Debugf("trying to push cache to dest: %+v from src:%+v", destSpec, src)
+	dest, err := alltransports.ParseImageName(destSpec)
+	// Add the docker:// transport to see if user neglected it while
+	// specifying --cache-to `<dest>`.
+	if err != nil {
+		destTransport := strings.Split(destSpec, ":")[0]
```
rhatdan, line 1616: I know that @mtrmac frowns on this.  I believe their should be containers/image libraries for this.

rhatdan (approved):

imagebuildah/stage_executor.go
flouthoc: Ah I see but following line is just here to check if valid transport was specified but we were still unable to ParseImageName in such case we don't attempt to add `docker://`, line below this split already uses `Get(` from `image/v5/transports` so I think we are already using `c/image` here but If there is even simpler way to check this then I'm ok to use that.

flouthoc:
> Question: when we have a local cache hit, why don't we try to push to a --cache-to location?

@nalind My initial idea was to save resources by only pushing steps which are just built and not which are already present in local storage. But upon thinking again I think it will be right to push everytime except for the case when intermediate image was just pulled.

So amended source and test case to match:

* If `--cache-to` is specified always `push` intermediate image expect for the  case when intermediate is just `pulled` for the `STEP` which is currently being evaluated.

imagebuildah/stage_executor.go
mtrmac: - Conceptually text parsing like that belongs at the CLI level; APIs should deal with well-typed data. Compare `buildah.Push(… dest types.ImageReference)` vs. the heuristics living in `cmd/buildah`. Besides conceptual cleanliness, formatting errors should be discovered before any actual expensive build starts.
- If there must be a heuristic, I’d prefer them not to be copy&pasted; they are inevitably going to drift apart (as… they already did in the Buildah codebase, into about 6 variants at least, and I can’t see when one would use one over the other). E.g. if this is intentionally using the `cmd/buildah/push.go` heuristic, fine please split that into a function, name and document it well (that’s the hardest part!) and call that function from both places, instead of copy&pasting.
- BTW. note that the semantics/parsing of `pullCache` and `pushCache` are very different. Not that I know how to make them exactly the same, apart from mandating the `alltransports.ParseImageName` format. Ugh.

WRT the narrow c/image API question,
```go
 		if t := alltransports.TransportFromImageName(destSpec); t != nil { // << this
 			return err
 		}

 		if strings.Contains(destSpec, "://") {
 			return err
 		}

 		dest2, err2 := docker.ParseReference("//"+destSpec) // << this, not unnecessarily alltransports again
 		if err2 != nil {
 			return err
 		}
 		dest = dest2
 		logrus.Debugf("Assuming docker:// as the transport method for cache DESTINATION: %s", destSpec) // possibly update the output here
 	}

**Comment by mtrmac** [comment]
> Also support pushing cache to other storage units rather than conventional container registry like object storage ( AWS S3, Redis, Minio etc )

What do you anticipate the API, and UI, to look like? Entirely new options, or somehow overloading the semantics of the existing input strings (and if so, how was it designed to avoid conflicts?)

File: imagebuildah/stage_executor.go [hunk 9]
```diff
@@ -1168,6 +1194,24 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			if err != nil {
 				return "", nil, fmt.Errorf("error checking if cached image exists from a previous build: %w", err)
 			}
+			// All the best effort to find cache on localstorage have failed try pulling
+			// cache from remote repo if `--cache-from` was configured.
+			if cacheID == "" && s.executor.cacheFrom != "" {
+				src := fmt.Sprintf("%s:%s", s.executor.cacheFrom, cacheKey)
```
mtrmac, line 1200: Looking at https://github.com/containers/buildah/pull/4118#discussion_r924478405 and https://github.com/containers/buildah/pull/4118#issuecomment-1189033450 … let me take a rather different tack:

- The c/image image names have no _common_ semantics apart from the initial _transport_`:` part. There is no expectation that one can just add a `:`_cacheKey_ to user input (let alone to a valid c/image image name), and have that be a valid name.
- What would we want to use this for, apart from `docker://`, anyway? I suppose `dir:` or `oci:` is plausible … for some kind of inter-storage cache sharing???? It’s not obvious to me that that’s better than a registry with compressed data that can just be reused on the registry via blob mounting.
- Given the Redis references… I think being explicit would be nice anyway.

So, how about a `cache{From,To}{Docker,Registry}Repo reference.Named`, parsed at the CLI via `reference.ParseNormalizedNamed()`+ an enforcement of `reference.IsNameOnly`? Then, the actual image name would be determined via `docker.NewReference(reference.WithTag(cache…Repo, cacheKey))`.

For other storage types, we would just add more option keys.

@nalind @rhatdan WDYT?

mtrmac, line 1200: BTW I don’t think the generic code should be doing this string pasting — just call `pullCache(ctx, cacheKey)`. (`pullCache` itself can read `s.executor.cacheFrom`), and the same for push.

(I wouldn’t even mind a helper function for this one line, shared between `pullCache` and `pushCache`, especially if this one line ends up being a bit longer `docker.NewReference(reference.WithTag(…))` as suggested elsewhere.)

mtrmac (commented):
I’m afraid this is a rather sloppy review, I didn’t quite take the time to fully understand the extensive logic dealing with various alternative possibilities.

Highlights:
- Should the cache locations be c/image generic names at all, or strictly registry repositories?
- Don’t use `reportWriter` output to make decisions.

imagebuildah/stage_executor.go
flouthoc: @mtrmac I like the idea of doing parsing and forming this at CLI will open room for lot more possiblities

imagebuildah/stage_executor.go
```diff
@@ -1521,6 +1586,124 @@ func (s *StageExecutor) tagExistingImage(ctx context.Context, cacheID, output st
...
+// generated CacheKey is further used by buildah to lock and decide
+// tag for the intermeidate image which can be pushed and pulled to/from
+// the remote repository.
+func (s *StageExecutor) generateCacheKey(ctx context.Context, currNode *parser.Node, addedContentDigest string, buildAddsLayer bool) (string, error) {
+	hash := sha256.New()
+	var baseHistory []v1.History
+	var manifestType string
+	var err error
+	if s.builder.FromImageID != "" {
+		manifestType, baseHistory, _, err = s.executor.getImageTypeAndHistoryAndDiffIDs(ctx, s.builder.FromImageID)
```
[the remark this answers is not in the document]
flouthoc: Diffids also include ids of the previous step and image is always different so we end with different SHA of the same step everytime hence i did not include it.

imagebuildah/stage_executor.go
```diff
@@ -1521,6 +1586,124 @@ func (s *StageExecutor) tagExistingImage(ctx context.Context, cacheID, output st
...
+	}
+	createdBy := s.getCreatedBy(currNode, addedContentDigest)
+	fmt.Fprintf(hash, "%t", buildAddsLayer)
+	fmt.Fprintln(hash, createdBy)
+	fmt.Fprintln(hash, manifestType)
+	for _, element := range baseHistory {
+		fmt.Fprintln(hash, element.CreatedBy)
+		fmt.Fprintln(hash, element.Author)
+		fmt.Fprintln(hash, element.Comment)
+		fmt.Fprintf(hash, "%t", element.EmptyLayer)
```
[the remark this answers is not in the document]
flouthoc: Will add this in the next push.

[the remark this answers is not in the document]
mtrmac: BTW an exact match on `Created` almost certainly implies an exact match on `DiffID` (it would be a possible but not-extremely-likely coincidence if two independent executions of the same build step happened at exactly the same timestamp).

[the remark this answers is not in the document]
mtrmac: I am unfamiliar with the code, and the trade-offs. At this point I was just naively looking for consistency with the existing cache logic.

In principle, the `DiffID` match is more exact but many build steps like ENV don’t produce layers. So the two are not substitutes.

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
flouthoc: Actually SGTM no issues in including `diffids` i just built the version and it is working fine for me.

docs/buildah-build.1.md
```diff
@@ -96,7 +96,37 @@ The value of `[name]` is matched with the following priority order:
 
 **--cache-from**
 
-Images to utilise as potential cache sources. Buildah does not currently support --cache-from so this is a NOOP.
+Repository to utilise as potential cache source. When specified buildah will try to look for
+cache images in specified repository and will attempt to pull cache images instead of actually
+executing and computing STEPS. Buildah will only attempt to pull cache images if they are considered
+as valid cache hits otherwise not.
+
+Use `--cache-to` in order populate remote repository with valid cache sources/layers.
```
nalind, line 104:
```suggestion
Use `--cache-to` in order populate a remote repository with cache content.
```

docs/buildah-build.1.md
```diff
@@ -96,7 +96,37 @@ The value of `[name]` is matched with the following priority order:
...
+
+Example
+
+```bash
+# populate cache source or use cached layers if already present
```
nalind, line 109:
```suggestion
# populate a cache and also consult it
```

docs/buildah-build.1.md
```diff
@@ -96,7 +96,37 @@ The value of `[name]` is matched with the following priority order:
...
+buildah build -t test --layers --cache-to registry/myrepo/cache --cache-from registry/myrepo/cache .
+```
+
+Note: Option works only with `--layers` i.e when cache is used.
```
nalind, line 113:
```suggestion
Note: option is ignored unless `--layers` is specified.
```

docs/buildah-build.1.md
```diff
@@ -96,7 +96,37 @@ The value of `[name]` is matched with the following priority order:
...
+Note: Works only with --layers
+
+Example
+
+```bash
+# populate cache source or use cached layers if already present
+buildah build -t test --layers --cache-to registry/myrepo/cache --cache-from registry/myrepo/cache .
+```
+
+Note: Option works only with `--layers` i.e when cache is used.
```
nalind, line 129:
```suggestion
Note: option is ignored unless `--layers` is specified.
```

docs/buildah-build.1.md
```diff
@@ -96,7 +96,37 @@ The value of `[name]` is matched with the following priority order:
...
+
+**--cache-to**
+
+Set this flag to specify a remote repository that will be used to store cache images. Buildah will attempt to
+push newly built cache image to the remote repository.
+
+Note: Works only with --layers
```
nalind, line 120:
```suggestion
Use `--cache-from` in order to use cache content in a remote repository.
```

imagebuildah/stage_executor.go
```diff
@@ -1156,6 +1173,15 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			commitName = s.output
 		}
 
+		// If --cache-from or --cache-to is specified make sure to populate
+		// cacheKey since it will be used either while pulling or pushing the
+		// cache images.
+		if s.executor.cacheFrom != "" || s.executor.cacheTo != "" {
+			cacheKey, err = s.generateCacheKey(ctx, node, addedContentSummary, s.stepRequiresLayer(step))
```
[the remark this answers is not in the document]
mtrmac: Assuming that every ADD/COPY that is _committed_ uses a digest, the existing `intermediateImageExists` without a digest should be mostly harmless, just an extra lookup that doesn’t find anything.

And on the ordinary paths, the commits do use a digest, apart _maybe_ from https://github.com/containers/buildah/blob/74080748760740db7560d85ed7660184283cf700/imagebuildah/stage_executor.go#L1320 if that last layer was created via cache reuse … which could only make a difference if we reused an ADD/COPY … which has only had to find a non-empty digest, by induction.

So I can’t see a correctness problem, but I’d feel much more comfortable with an explicit “does `addedContentSummary` matter for the current `step`” logic, and only doing a cache lookup in one or the other place, depending on that.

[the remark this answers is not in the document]
mtrmac: … ^^^ applies to the existing local-image reuse logic.

The new `--cache-from` logic does, it seems to me, need to be modified to compute the right cache key, both for cache pulls and pushes.

So adding the “does `addedContentSummary` matter for the current `step`” logic, and consolidating both caches around that, would probably be worth it — the local image lookups are a _bit_ expensive, and the remote HTTP is much more so.

docs/buildah-build.1.md
mtrmac: “in order _to_ populate”?

mtrmac: Or just “… to populate”, without “in order”

imagebuildah/stage_executor.go
mtrmac: BTW `pullCache(ctx, cacheKey)` would nicely almost-isolate the storage backend (registry/S3/…) from the stage executor.

(_almost_-isolate because there is still the `s.executor.cacheFrom != ""` condition which might need abstracting. That can certainly be done as well.)

imagebuildah/stage_executor.go
```diff
@@ -1521,6 +1586,124 @@ func (s *StageExecutor) tagExistingImage(ctx context.Context, cacheID, output st
...
+	// a valid cache source from local storage
+	if reportWriter.String() == "" {
+		logrus.Debugf("cache was not pulled from:%s since id:%s is already present on local storage", src, id)
+		// There can be cases when cache will not be pulled
+		// since its already present on our local storage as our
+		// pull policy is `PullIfNewer` so return early and let
+		// buildah lookup cache in other attempts for example by
+		// recalculating content summary and adding content to find
+		// its digest and re-lookin cache from local storage.
+		// See func (s *StageExecutor) Execute() for more details.
```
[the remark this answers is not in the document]
flouthoc: Removed and added a better reason above.

imagebuildah/stage_executor.go
```diff
@@ -1521,6 +1586,124 @@ func (s *StageExecutor) tagExistingImage(ctx context.Context, cacheID, output st
...
+
+	id, err := buildah.Pull(ctx, src, options)
+	if err != nil {
+		logrus.Debugf("failed pulling cache from repo %s: %v", src, err)
+		return "", fmt.Errorf("failed while pulling cache from %q: %w", src, err)
+	}
+	// Don't return any id, since image was already present on
+	// local so buildah will eventually figure out and use it as
+	// a valid cache source from local storage
+	if reportWriter.String() == "" {
```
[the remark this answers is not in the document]
flouthoc: We need to confirm if actually image was pulled or not so that we can generate correct output text in the build output. Also we are using build output to run tests.

I have removed reportwriter and checking this against the localstorage now, but I agree that Pull API can be extended for a better implementation so will replace this with Pull API once I add support in Pull API for this but will do that in a different PR.

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
flouthoc: Added Diffids

imagebuildah/stage_executor.go
```diff
@@ -1246,6 +1300,17 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
+			// check if --cache-to was specified if yes then attempt pushing
+			// this cache to remote repo and fail accordingly
+			if s.executor.cacheTo != "" {
+				// try to push this cache to remote repository
+				destSpec := fmt.Sprintf("%s:%s", s.executor.cacheTo, cacheKey)
+				logCachePush(destSpec)
+				if err = s.pushCache(ctx, imgID, destSpec); err != nil {
+					return "", nil, err
+				}
+			}
```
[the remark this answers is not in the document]
flouthoc: consolidated with a wrapper.

imagebuildah/stage_executor.go
flouthoc: @mtrmac One doubt would it makes sense to do generation of `reference.Named` at backend (imagebuildah) and just pass `string` from frontend API my intention is to make frontend API easy to consume and let imagebuildah take care of everything ?  But just a suggestion and no issues from my side with doing it at frontend/CLI.

flouthoc:
> Don’t use reportWriter output to make decisions.

New push is ensures we are not using this anymore. Still some comments are pending but I hope in another push we can get them resolved.

imagebuildah/stage_executor.go
nalind: It's not a prohibitively complicated requirement (`buildah.Push()` returns a `reference.Canonical`) for an optional setting, and I think the loss of ambiguity would be well worth it.

flouthoc: Cool I agree, I'll make this change in the next push.

imagebuildah/stage_executor.go
```diff
@@ -1168,6 +1193,24 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
+			// All the best effort to find cache on localstorage have failed try pulling
+			// cache from remote repo if `--cache-from` was configured.
+			if cacheID == "" && s.executor.cacheFrom != "" {
+				src := fmt.Sprintf("%s:%s", s.executor.cacheFrom, cacheKey)
+				// only attempt to use cache again if pulling was successful
+				// otherwise do nothing and attempt to run the step, err != nil
+				// is ignored and will be automatically logged for --log-level debug
+				if id, err := s.pullCache(ctx, src); id != "" && err == nil {
+					logCachePulled(src)
+					cacheID, err = s.intermediateImageExists(ctx, node, addedContentSummary, s.stepRequiresLayer(step))
```
mtrmac, line 1205: BTW it might be worthwhile to extract the “does $candidateImage match the current image and next step” logic from `intermediateImageExists` into a separate function, and here call it _only_ with the just-pulled image.

`intermediateImageExists` seems to do several a few c/storage operations (at least the `.Layer` call is not cached), each with its own locking syscalls and related overhead, so doing an O(1) check instead of an O(existing images) check would be nice.

imagebuildah/stage_executor.go
```diff
@@ -1216,6 +1259,18 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			}
 		}
 
+		// Small wrapper around s.pushCache to prevent duplication of code
+		pushCache := func() error {
```
mtrmac, line 1263: Is this necessary? I was expecting something like
```go
if cacheID == "" || !cachePulled {
   // body of this helper inlined
}
```
after the `if cacheID != ""` below, directly inline in the per-stage flow. (That would also make it a tiny bit easier to see under which conditions cache images are pushed, without having to trace the execution flow at all.)

imagebuildah/stage_executor.go
```diff
@@ -1145,6 +1159,8 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 		var (
 			commitName          string
 			cacheID             string
+			cacheKey            string
+			cachePulled         bool
```
mtrmac, line 1163: I’m tempted to suggest something like
```suggestion
			pulledAndUsedCacheImage         bool
```
— the larger a function, the more descriptive names should be (also, the value is not set when we just _pulled_ a cache image, but when `s.intermediateImageExists` matched it).

But that’s entirely up to the repo maintainers.

imagebuildah/stage_executor.go
```diff
@@ -1521,6 +1590,142 @@ func (s *StageExecutor) tagExistingImage(ctx context.Context, cacheID, output st
...
+		//
+		// pullCache eventually returns pulled image ID back to the caller
+		// and here buildah will produce a log with image ID notifying to
+		// end-users if anything was actually pulled from remote repo in the
+		// the build output by calling `logCachePulled()`.
+		//
+		// Since following build output is part of stdout lets not return wrong
+		// info and further we are using build output to verify test-cases so
+		// don't return any id, since image was already present on local and buildah
+		// will eventually figure out to use valid cache source from local storage.
```
mtrmac, line 1717: AFAICS we only call this if Buildah _did not_ match any of the local images, so I don’t see any opportunity to “eventually figure out” things.

flouthoc: @mtrmac Cache hit attempt happens again here where we still RUN the step but don't commit it https://github.com/containers/buildah/blob/main/imagebuildah/stage_executor.go#L1173 specially happens for cases like `COPY` and `ADD`

mtrmac: My mistake, you’re completely right about this path potentially matching. 

But that see the large `addedContentSummary` discussion elsewhere, it would AFAICS never match as is.

I expect we need to have the `addedContentSummary` issue fixed first, and only then we could see the `cachePulled` logic take final shape. (At this point I’d expect for the cache lookup to only happen _once_, in different locations based on the step’s instruction — so the mentioned “eventually future out” would not end up happening.)

imagebuildah/stage_executor.go
mtrmac: (Refactorings like this would be best done in separate commits, or maybe deferred to a separate PR.)

imagebuildah/stage_executor.go
flouthoc: I think this whole logic can be removed if we optimize cache hits for `COPY` and `ADD` since they always miss the first pass, I created a PR for this: https://github.com/containers/buildah/pull/4128

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
flouthoc: This is refactored here so we don't end up wasting cycles: https://github.com/containers/buildah/pull/4128

docs/buildah-build.1.md
```diff
@@ -96,7 +96,37 @@ The value of `[name]` is matched with the following priority order:
 
 **--cache-from**
 
-Images to utilise as potential cache sources. Buildah does not currently support --cache-from so this is a NOOP.
+Repository to utilise as potential cache source. When specified buildah will try to look for
```
TomSweeneyRedHat, line 99:
```suggestion
Repository to utilize as a potential cache source. When specified, Buildah will look for
```

docs/buildah-build.1.md
```diff
@@ -96,7 +96,37 @@ The value of `[name]` is matched with the following priority order:
...
+cache images in specified repository and will attempt to pull cache images instead of actually
```
TomSweeneyRedHat, line 100:
```suggestion
cache images in the specified repository and will attempt to pull cache images instead of 
```

docs/buildah-build.1.md
```diff
@@ -96,7 +96,37 @@ The value of `[name]` is matched with the following priority order:
...
+executing and computing STEPS. Buildah will only attempt to pull cache images if they are considered
+as valid cache hits otherwise not.
```
TomSweeneyRedHat, line 102: I think, but don't know, that it probably ought to be "cached images" instead of "cache images" above.  @nalind ?

docs/buildah-build.1.md
```diff
@@ -96,7 +96,37 @@ The value of `[name]` is matched with the following priority order:
...
+
+Use `--cache-to` to populate a remote repository with cache content
```
TomSweeneyRedHat, line 104:
```suggestion
Use `--cache-to` to populate a remote repository with cache content.
```

docs/buildah-build.1.md
```diff
@@ -96,7 +96,37 @@ The value of `[name]` is matched with the following priority order:
...
+
+Example
+
+```bash
+# populate a cache and also consult it
+buildah build -t test --layers --cache-to registry/myrepo/cache --cache-from registry/myrepo/cache .
+```
+
+Note: option is ignored unless `--layers` is specified.
```
TomSweeneyRedHat, line 113: which option is ignored?  use used `-t1, `--cache-to`, and `--cache-from`

docs/buildah-build.1.md
```diff
@@ -96,7 +96,37 @@ The value of `[name]` is matched with the following priority order:
...
+
+**--cache-to**
+
+Set this flag to specify a remote repository that will be used to store cache images. Buildah will attempt to
+push newly built cache image to the remote repository.
+
+Note: Use `--cache-from` in order to use cache content in a remote repository.
```
TomSweeneyRedHat, line 120:
```suggestion
Note: Use the `--cache-from` option to use cache content in a remote repository.
```

docs/buildah-build.1.md
```diff
@@ -96,7 +96,37 @@ The value of `[name]` is matched with the following priority order:
...
+
+Example
+
+```bash
+# populate a cache and also consult it
+buildah build -t test --layers --cache-to registry/myrepo/cache --cache-from registry/myrepo/cache .
+```
+
+Note: option is ignored unless `--layers` is specified.
```
TomSweeneyRedHat, line 129: same "cache" vs "cached" question here.

docs/buildah-build.1.md
nalind: Uh, "previously cached", I guess.

docs/buildah-build.1.md
nalind: This is the section about `--cache-from`, but I don't mind if it's named on this line.

docs/buildah-build.1.md
nalind: This is the section about `--cache-to`, but I don't mind if it's named on this line.

docs/buildah-build.1.md
flouthoc: fixed.

imagebuildah/stage_executor.go
flouthoc: this is already fixed in PR https://github.com/containers/buildah/pull/4128

imagebuildah/stage_executor.go
flouthoc: moved to below line.

imagebuildah/stage_executor.go
flouthoc: Will tackle this in a different PR.

imagebuildah/stage_executor.go
flouthoc: implemented on CLI

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
flouthoc: This is already resolved in PR https://github.com/containers/buildah/pull/4128

imagebuildah/stage_executor.go
flouthoc: Moved to CLI.

flouthoc:
Moved parsing to CLI and not image buildah accepts `reference.Named` for `cacheTo` and `cacheFrom`.

rhatdan:
@mtrmac @nalind one last pass, would like to get this into the buildah release which is happening shortly.

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
mtrmac: It isn’t, AFAICS: `cacheKey` is still computed from empty `addedContentSummary`, and not updated for ADD/COPY.

define/build.go
```diff
@@ -136,6 +137,12 @@ type BuildOptions struct {
 	RuntimeArgs []string
 	// TransientMounts is a list of mounts that won't be kept in the image.
 	TransientMounts []string
+	// CacheFrom specifies any remote repository which can be treated as
+	// potential cache source.
+	CacheFrom *reference.Named
+	// CacheTo specifies any remote repository which can be treated as
+	// potential cache destination.
+	CacheTo *reference.Named
```
mtrmac, line 145: Go interface values (sadly) include a `nil` value, so this doesn’t need the `*`. And not including it would be better because the corner case of “non-`nil` pointer to `nil`” would go away.

imagebuildah/stage_executor.go
```diff
@@ -1542,6 +1605,114 @@ func (s *StageExecutor) tagExistingImage(ctx context.Context, cacheID, output st
...
+		fmt.Fprintln(hash, element.Created)
+		fmt.Fprintf(hash, "%t", element.EmptyLayer)
+		fmt.Fprintln(hash)
+	}
+	return fmt.Sprintf("%x", hash.Sum(nil)), nil
+}
+
+// imageReferenceFromNamed is internal function which generates ImageReference from Named repo sources
+// and a tag.
+func imageReferenceFromNamed(named reference.Named, tag string) (types.ImageReference, error) {
```
mtrmac, line 1645: Non-blocking: This works, but the name and description of the image is not very helpful WRT about its intended purpose (and, pedantically, “`fromNamed` doesn’t say anything about the other parameter).


Consider something like `cacheImageReference(repo reference.Named, cacheKey string)`, as an explicit abstraction over the image naming design decision.

pkg/parse/parse.go
```diff
@@ -48,6 +49,22 @@ const (
 	BuildahCacheDir = "buildah-cache"
 )
 
+// RepoNameToNamedReference parse the raw string to Named reference
+func RepoNameToNamedReference(dest string) (reference.Named, error) {
+	ref, err := reference.Parse(dest)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse reference: %w", err)
+	}
+	named, ok := ref.(reference.Named)
```
mtrmac, line 58: I think this should just be a `reference.ParseNormalizedNamed`. Otherwise _something_ is going to need to deal with the `docker.io[/library]` normalization, notably `c/image/docker.NewReference` expects a normalized value, and using raw `busybox` like the above would break it.

pkg/parse/parse.go
```diff
@@ -48,6 +49,22 @@ const (
...
+	if !ok {
+		return nil, fmt.Errorf("invalid repo %q: must contain registry and repository", dest)
+	}
+	if !reference.IsNameOnly(named) {
+		return nil, fmt.Errorf("invalid parsed named: %v", named)
```
mtrmac, line 63: This is an error case users might actually reasonably trigger, and the message needs to be useful to users (e.g. “repository must contain neither a tag nor digest”).

pkg/cli/build.go
```diff
@@ -290,6 +287,24 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
...
+		}
+		cacheTo = &to
+	}
+	if c.Flag("cache-from").Changed {
+		from, err := parse.RepoNameToNamedReference(iopts.CacheFrom)
+		if err != nil {
+			return options, nil, nil, err
+		}
+		cacheFrom = &from
+	}
```
mtrmac, line 307: If the parsing fails, this doesn’t tell the user which option failed.

One possibility would be to pass the option name into `parse.…`; or wrap the error with extra context using `fmt.Errorf(…%w)` here.

mtrmac (commented):
AFAICS `cacheKey` is still not computed correctly for ADD/COPY.

pkg/parse/parse.go
flouthoc: `ParseNormalizedNamed(` accepts tag also afaics, it does not accepts string like `localhost/repo`, without tag it fails with invalid reference format hence i decided not to use it but let me know if something else can be used.

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
flouthoc: Should address now. I also verfified a smaller test with
```Dockerfile
FROM alpine
ADD hello .
COPY world .
```

pkg/parse/parse.go
mtrmac: > `ParseNormalizedNamed(` accepts tag also afaics

Sure, I wasn’t suggesting dropping the `IsNameOnly` check.

>, it does not accepts string like `localhost/repo`

I have just tested that it does accept exactly that string.

mtrmac: On an edited unit test, showing how various inputs are normalized (input => `ref.String()`):
```
    normalize_test.go:58: "docker/docker" => "docker.io/docker/docker"
    normalize_test.go:58: "library/debian" => "docker.io/library/debian"
    normalize_test.go:58: "debian" => "docker.io/library/debian"
    normalize_test.go:58: "docker.io/docker/docker" => "docker.io/docker/docker"
    normalize_test.go:58: "docker.io/library/debian" => "docker.io/library/debian"
    normalize_test.go:58: "docker.io/debian" => "docker.io/library/debian"
    normalize_test.go:58: "index.docker.io/docker/docker" => "docker.io/docker/docker"
    normalize_test.go:58: "localhost/repo:tag" => "localhost/repo:tag"
    normalize_test.go:58: "localhost" => "docker.io/library/localhost"
    normalize_test.go:58: "notlocalhost" => "docker.io/library/notlocalhost"
    normalize_test.go:58: "index.docker.io/library/debian" => "docker.io/library/debian"
    normalize_test.go:58: "index.docker.io/debian" => "docker.io/library/debian"
    normalize_test.go:58: "127.0.0.1:5000/docker/docker" => "127.0.0.1:5000/docker/docker"
    normalize_test.go:58: "127.0.0.1:5000/library/debian" => "127.0.0.1:5000/library/debian"
    normalize_test.go:58: "127.0.0.1:5000/debian" => "127.0.0.1:5000/debian"
    normalize_test.go:58: "thisisthesongthatneverendsitgoesonandonandonthisisthesongthatnev" => "docker.io/library/thisisthesongthatneverendsitgoesonandonandonthisisthesongthatnev"
    normalize_test.go:58: "docker.io/1a3f5e7d9c1b3a5f7e9d1c3b5a7f9e1d3c5b7a9f1e3d5d7c9b1a3f5e7d9c1b3a" => "docker.io/library/1a3f5e7d9c1b3a5f7e9d1c3b5a7f9e1d3c5b7a9f1e3d5d7c9b1a3f5e7d9c1b3a"
```

flouthoc: hmm it works for me now, I guess i just used the wrong function last time. Thanks I have amended the PR.

imagebuildah/stage_executor.go
flouthoc: Yes amended I agree this name looks better.

define/build.go
flouthoc: Removed the reference and passing the actual struct now.

flouthoc:
> AFAICS `cacheKey` is still not computed correctly for ADD/COPY.

@mtrmac This should be addressed in latest commit where we reevaluate the `cacheKey` if summary was changed.

docs/buildah-build.1.md
```diff
@@ -96,7 +96,37 @@ The value of `[name]` is matched with the following priority order:
 
 **--cache-from**
 
-Images to utilise as potential cache sources. Buildah does not currently support --cache-from so this is a NOOP.
+Repository to utilize as a potential cache source. When specified, Buildah will try to look for
+cache images in the specified repository and will attempt to pull cache images instead of actually
+executing and computing STEPS. Buildah will only attempt to pull previously cached images if they
```
mtrmac, line 101:
```suggestion
executing the build steps locally. Buildah will only attempt to pull previously cached images if they
```

docs/buildah-build.1.md
```diff
@@ -96,7 +96,37 @@ The value of `[name]` is matched with the following priority order:
...
+are considered as valid cache hits otherwise not.
```
mtrmac, line 102:
```suggestion
executing and computing STEPS. Buildah will only attempt to pull previously cached images if they
are considered as valid cache hits.
```
or just drop the sentence completely, I’m not sure what users are supposed to do with that information.

docs/buildah-build.1.md
```diff
@@ -96,7 +96,37 @@ The value of `[name]` is matched with the following priority order:
...
+
+Use the `--cache-to` option to populate a remote repository with cache content
```
mtrmac, line 104:
```suggestion
Use the `--cache-to` option to populate a remote repository with cache content.
```

tests/bud.bats
```diff
@@ -3864,6 +3864,99 @@ _EOF
...
+  mkdir -p ${mytmpdir}
+  cat > $mytmpdir/Containerfile << _EOF
+FROM alpine
+RUN echo hello
+RUN echo world
+RUN touch hello
+
+FROM alpine
+RUN echo hello
+COPY --from=0 hello .
```
mtrmac, line 3879: Is this supposed to hit the cache? I don’t see specific checks for this build step below (and it would be a bit useful to have tests specifically exercise the ADD/COPY path).

imagebuildah/stage_executor.go
```diff
@@ -1165,6 +1185,15 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			commitName = s.output
 		}
 
+		// If --cache-from or --cache-to is specified make sure to populate
+		// cacheKey since it will be used either while pulling or pushing the
+		// cache images.
+		if s.executor.cacheFrom != nil || s.executor.cacheTo != nil {
```
mtrmac, line 1191: Non-blocking: I wouldn’t mind computing this always. The other two updates of `cacheKey` are unconditional, and that inconsistency seems unnecessary and potentially confusing to future readers.

The cost of `generateCacheKey` is not zero, but it’s likely trivial compared to the build steps or cache use.

Alternatively, `needsCacheKey := s.executor.cacheFrom != nil || s.executor.cacheTo != nil` and wrap _all_ updates of `cacheKey` with `if needsCacheKey`.

imagebuildah/stage_executor.go
```diff
@@ -1269,6 +1325,23 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
+		// cache so check if --cache-to was specified if yes
+		// then attempt pushing this cache to remote repo and
+		// fail accordingly.
+		//
+		// Or
+		//
+		// Try to push this cache to remote repository only
+		// if cache was present on local storage and not
+		// pulled from remote source while processing this
+		if (!pulledAndUsedCacheImage || cacheID == "") && s.executor.cacheTo != nil {
```
mtrmac, line 1338: Absolutely non-blocking: Consider
```suggestion
		if s.executor.cacheTo != nil && (!pulledAndUsedCacheImage || cacheID == "") {
```
which reads “which feature is this about? `cacheTo` — and in which cases do we need to do something for that feature?”, that seems more natural to me than “in which cases do we need to do something about `cacheTo`? — and do we even care about `cacheTo` at all?”.

tests/bud.bats
mtrmac: (This testing coverage concern is non-blocking from my POV.)

flouthoc: Yes updated tests to cover `COPY` and also added case for `ADD`

flouthoc:
Pushed updated tests for `ADD` and `COPY` case.

mtrmac (approved):

mtrmac:
/lgtm

Thanks!

TomSweeneyRedHat:
Happy green test buttons @rhatdan.  I'm not going to cancel the hold as I'm not sure if you were holding on tests, or holding until release with this one.

mukulk2020:
@flouthoc : How can i install latest version of buildah in my centos machine so that i can use these parameters "--cache-to, --cache-from" along with buildah bud command . I checked the install.md (https://github.com/containers/buildah/blob/main/install.md) but this will install older version of buildah i.e. 1.11.xxx. Please help me on this so that i can run below command to achieve cache .

Exam : buildah bud -t cache-test --layers --cache-to mukulk/cache-test --cache-from mukulk/cache-test .

flouthoc:
@mukulk2020 Thanks for reaching out it seems package on centos has not been updated for a while, I think @lsm5 and @rhatdan can confirm about who maintains it and who can update it.

mukulk2020:
@lsm5 @rhatdan Please check and update the docs and code accordingly..

flouthoc:
I don't think @lsm5 or @rhatdan actually maintains packages on centos but they can guide or tag the right person who can do it. :)

lsm5:
@mukulk2020 your screenshot tells me you're on CentOS 7. We're not doing updates for CentOS 7 anymore. Please consider switching to centos stream 8 / 9.

Issue #4079, opened 2022-06-24
Undefined concurrent writes in BuildDockerfiles


mtrmac:
**Steps to reproduce the issue:**
Code inspection only

https://github.com/containers/buildah/blob/d8e9ca4d818d3d3c0a3d5cddbfcb50c0a8cdb989/imagebuildah/build.go#L65 defines return values `id` and `ref`, shared for the whole `BuildDockerfiles` function. https://github.com/containers/buildah/blob/d8e9ca4d818d3d3c0a3d5cddbfcb50c0a8cdb989/imagebuildah/build.go#L255 starts a new goroutine, and that writes to `id` and `ref` at https://github.com/containers/buildah/blob/d8e9ca4d818d3d3c0a3d5cddbfcb50c0a8cdb989/imagebuildah/build.go#L278 without anything I can see that would prevent concurrent writes. That can have, in principle, unpredictable results (starting with getting `id` and `ref` values from different goroutines).

I suppose the idea is that on single-platform builds there is no race, and on multi-platform builds the fields are not used anyway. (Pedantically: Go is fairly imprecise about its memory model, so it’s unclear to me that “racy but unused writes” don’t trigger undefined behavior.) I can’t see that this is actually the case, nothing forces multi-platform builds to set the `Manifest` option which would cause `id` and `ref` to be overwritten.

Moving those writes under (renamed) `instanceLock` might be a one-line fix … but it seems worthwhile to move the per-platform goroutine into a completely separate function to make it impossible to similarly write to shared state without making it explicitly visible to the goroutine.

flouthoc:
I think we can avoid sharing these variable inside goroutines. Created a refactor PR above: https://github.com/containers/buildah/pull/4120

PR #4120, opened 2022-07-18
refactor,imagebuildah,build: remove undefined concurrent writes

flouthoc:
This is refactor commit which aims to remove undefined concurrent writes
from imagebuildah by making sure that fields which we are modifying are
not shared by builds for each platform so there is no undefined
concurrent write

[NO TESTS NEEDED]
[NO NEW TESTS NEEDED]
No new test is needed for this OTOH all the existing tests should pass

Closes: https://github.com/containers/buildah/issues/4079

```release-note
None
```


mtrmac (commented):
/lgtm . Thanks!

PR #4128, opened 2022-07-21
imagebuildah: optimize cache hits for `COPY` and `ADD` instructions

flouthoc:
While processing `COPY` and `ADD` try for cache hits in the first
iteration by adding content and populating summary with the digest
generated after adding the content and ignore processing same
instruction again if they were evaluated in the first pass to save
cycles.

See discussion here:
* https://github.com/containers/buildah/pull/4118#discussion_r924502857
* https://github.com/containers/buildah/pull/4118#discussion_r925877641

Should help in refining code for: https://github.com/containers/buildah/pull/4128

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -35,6 +35,7 @@ import (
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/openshift/imagebuilder"
+	"github.com/openshift/imagebuilder/dockerfile/command"
 	"github.com/openshift/imagebuilder/dockerfile/parser"
 	"github.com/sirupsen/logrus"
 )
@@ -867,6 +868,21 @@ func (s *StageExecutor) getImageRootfs(ctx context.Context, image string) (mount
 	return builder.MountPoint, nil
 }
 
+// getContentSummary generates content summary for cases where we added content and need
+// to get summary with updated digests.
+func (s *StageExecutor) getContentSummaryAfterAddingContent() string {
+	contentType, digest := s.builder.ContentDigester.Digest()
+	summary := contentType
+	if digest != "" {
+		if summary != "" {
+			summary = summary + ":"
+		}
+		summary = summary + digest.Encoded()
+		logrus.Debugf("added content %s", summary)
+	}
+	return summary
+}
+
 // Execute runs each of the steps in the stage's parsed tree, in turn.
 func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string, ref reference.Canonical, err error) {
 	var resourceUsage rusage.Rusage
@@ -1101,15 +1117,7 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 				return "", nil, fmt.Errorf("error building at STEP \"%s\": %w", step.Message, err)
 			}
 			// In case we added content, retrieve its digest.
-			addedContentType, addedContentDigest := s.builder.ContentDigester.Digest()
-			addedContentSummary := addedContentType
-			if addedContentDigest != "" {
-				if addedContentSummary != "" {
-					addedContentSummary = addedContentSummary + ":"
-				}
-				addedContentSummary = addedContentSummary + addedContentDigest.Encoded()
-				logrus.Debugf("added content %s", addedContentSummary)
-			}
+			addedContentSummary := s.getContentSummaryAfterAddingContent()
 			if moreInstructions {
 				// There are still more instructions to process
 				// for this stage.  Make a note of the
@@ -1143,11 +1151,12 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 
 		// We're in a multi-layered build.
 		var (
-			commitName          string
-			cacheID             string
-			err                 error
-			rebase              bool
-			addedContentSummary string
+			commitName                string
+			cacheID                   string
+			err                       error
+			rebase                    bool
+			addedContentSummary       string
+			canMatchCacheOnlyAfterRun bool
 		)
 
 		// If we have to commit for this instruction, only assign the
@@ -1164,6 +1173,20 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 		// determining if a cached layer with the same build args already exists
 		// and that is done in the if block below.
 		if checkForLayers && step.Command != "arg" && !(s.executor.squash && lastInstruction && lastStage) {
+			// For `COPY` and `ADD`, history entries include digests computed from
+			// the content that's copied in.  We need to compute that information so that
+			// it can be used to evaluate the cache, which means we need to go ahead
+			// and copy the content.
+			canMatchCacheOnlyAfterRun = (step.Command == command.Add || step.Command == command.Copy)
+			if canMatchCacheOnlyAfterRun {
+				if err = ib.Run(step, s, noRunsRemaining); err != nil {
+					logrus.Debugf("Error building at step %+v: %v", *step, err)
+					return "", nil, fmt.Errorf("error building at STEP \"%s\": %w", step.Message, err)
+				}
+				// Retrieve the digest info for the content that we just copied
+				// into the rootfs.
+				addedContentSummary = s.getContentSummaryAfterAddingContent()
+			}
 			cacheID, err = s.intermediateImageExists(ctx, node, addedContentSummary, s.stepRequiresLayer(step))
 			if err != nil {
 				return "", nil, fmt.Errorf("error checking if cached image exists from a previous build: %w", err)
@@ -1174,7 +1197,9 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 		// to find the digest of the content to check for a cached
 		// image, run the step so that we can check if the result
 		// matches a cache.
-		if cacheID == "" {
+		// We already called ib.Run() for the `canMatchCacheOnlyAfterRun`
+		// cases above, so we shouldn't do it again.
+		if cacheID == "" && !canMatchCacheOnlyAfterRun {
 			// Process the instruction directly.
 			if err = ib.Run(step, s, noRunsRemaining); err != nil {
 				logrus.Debugf("Error building at step %+v: %v", *step, err)
@@ -1182,15 +1207,7 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			}
 
 			// In case we added content, retrieve its digest.
-			addedContentType, addedContentDigest := s.builder.ContentDigester.Digest()
-			addedContentSummary = addedContentType
-			if addedContentDigest != "" {
-				if addedContentSummary != "" {
-					addedContentSummary = addedContentSummary + ":"
-				}
-				addedContentSummary = addedContentSummary + addedContentDigest.Encoded()
-				logrus.Debugf("added content %s", addedContentSummary)
-			}
+			addedContentSummary = s.getContentSummaryAfterAddingContent()
 
 			// Check if there's already an image based on our parent that
 			// has the same change that we just made.
@@ -1201,6 +1218,10 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 				}
 			}
 		} else {
+			// This log line is majorly here so we can verify in tests
+			// that our cache is performing in the most optimal way for
+			// various cases.
+			logrus.Debugf("Found a cache hit in the first iteration with id %s", cacheID)
 			// If the instruction would affect our configuration,
 			// process the configuration change so that, if we fall
 			// off the cache path, the filesystem changes from the
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -474,6 +474,38 @@ _EOF
   assert "$output" !~ "unwanted stage"
 }
 
+# Test if our cache is working in optimal way for COPY use case
+@test "build test optimal cache working for COPY instruction" {
+  mkdir -p ${TEST_SCRATCH_DIR}/bud/platform
+
+  echo something > ${TEST_SCRATCH_DIR}/bud/platform/somefile
+  cat > ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile << _EOF
+FROM alpine
+COPY somefile .
+_EOF
+
+  run_buildah build $WITH_POLICY_JSON --layers -t source -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile ${TEST_SCRATCH_DIR}/bud/platform
+  # Run again and verify if we hit cache in first pass
+  run_buildah --log-level debug build $WITH_POLICY_JSON --layers -t source -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile ${TEST_SCRATCH_DIR}/bud/platform
+  expect_output --substring "Found a cache hit in the first iteration"
+}
+
+# Test if our cache is working in optimal way for ADD use case
+@test "build test optimal cache working for ADD instruction" {
+  mkdir -p ${TEST_SCRATCH_DIR}/bud/platform
+
+  echo something > ${TEST_SCRATCH_DIR}/bud/platform/somefile
+  cat > ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile << _EOF
+FROM alpine
+ADD somefile .
+_EOF
+
+  run_buildah build $WITH_POLICY_JSON --layers -t source -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile ${TEST_SCRATCH_DIR}/bud/platform
+  # Run again and verify if we hit cache in first pass
+  run_buildah --log-level debug build $WITH_POLICY_JSON --layers -t source -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile ${TEST_SCRATCH_DIR}/bud/platform
+  expect_output --substring "Found a cache hit in the first iteration"
+}
+
 # Test skipping unwanted stage with --mount from another stage
 @test "build-test skipping unwanted stages with --mount from stagename" {
   mkdir -p ${TEST_SCRATCH_DIR}/bud/platform


flouthoc:
@nalind @mtrmac PTAL. I think this will optimize and fix the code for cache hits.

rhatdan:
LGTM

imagebuildah/stage_executor.go
```diff
@@ -1164,6 +1164,30 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
+				addedContentType, addedContentDigest := s.builder.ContentDigester.Digest()
+				addedContentSummary = addedContentType
+				if addedContentDigest != "" {
+					if addedContentSummary != "" {
+						addedContentSummary = addedContentSummary + ":"
+					}
+					addedContentSummary = addedContentSummary + addedContentDigest.Encoded()
+					logrus.Debugf("added content %s", addedContentSummary)
+				}
+
```
nalind, line 1189:
```suggestion
```

imagebuildah/stage_executor.go
```diff
@@ -1175,6 +1199,13 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
 		if cacheID == "" {
+			// Note: In theory we should never ever hit this branch now
+			// since theoretically only `COPY` and `ADD` should  fall in
+			// this branch but we already process them with execption cases
+			// above to save cycles. But still leave this branch here so if
+			// if we missed anything then we still end up giving our best
+			// effort to find a cache hit.
+
 			// Process the instruction directly.
 			if err = ib.Run(step, s, noRunsRemaining); err != nil {
```
nalind, line 1210: If we have a cache miss for an ADD or COPY instruction, are we copying the content again?

imagebuildah/stage_executor.go
```diff
@@ -1164,6 +1164,30 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 		// determining if a cached layer with the same build args already exists
```
mtrmac, line 1164: The commend above about ARG is also suggestive that there might be some more logic to this.

imagebuildah/stage_executor.go
flouthoc: @nalind Yes if miss `cache` hit for ADD or COPY content gets copied again.I can add a test to verify this in CI as well.

imagebuildah/stage_executor.go
```diff
@@ -1164,6 +1164,30 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
+				// In case we added content, retrieve its digest.
+				addedContentType, addedContentDigest := s.builder.ContentDigester.Digest()
+				addedContentSummary = addedContentType
+				if addedContentDigest != "" {
+					if addedContentSummary != "" {
+						addedContentSummary = addedContentSummary + ":"
+					}
+					addedContentSummary = addedContentSummary + addedContentDigest.Encoded()
+					logrus.Debugf("added content %s", addedContentSummary)
+				}
```
mtrmac, line 1188: At three copies, this should _definitely_ be turned into a function. (The 480-line `Execute` function is already so large, so I’m looking for ways to make it smaller.)

mtrmac (commented):
This way of structuring the logic is very unintuitive to me.

In my way of thinking, the `ib.Run` is the _core_ of the step execution, and all the caching are  more of exceptional corner cases.

So…
```go
// turned checkForLayers into givenUpOnCacheMatching := !checkForLayers



// === Perform the build step, or use a cache.
canUseCachedLayer := !squash || !astLayer … // For squashing, we want to use a single total layer, so reusing an unsquashed one would not be valuable. (???)
canMatchCacheOnlyAfterRun := step.Command in (arg, add, copy) // We need to compute ConentDigester for (ADD, COPY); … for ARG.

if canUseCachedLayer && !canMatchCacheOnlyAfterRun && !givenUpOnCacheMatching {
  cacheID, err = findInCache(contentDigest = nil)
  if cacheID != "" {
     // the if !s.stepRequiresLayer(step) { ib.Run }
     // logic goes here, into the “found a cache” block? I’m not sure why this is necessary.
   }
}
if cacheID == "" {
    ib.Run()
}

// === Try to match the created container against a pre-existing image to maximize reuse
// Don’t do that for images that already come from a cache.
// !!! should this be conditional on canUseCachedLayer like the above?
if cacheID == "" && canMatchCacheOnlyAfterRun && !givenUpOnCacheMatching {
       cacheID, err = findInCache(contentDigest = nil)
   }
}

// === tag/commit, and so on.
```

---

I worry how to keep this in sync with actual logic. At least a comment for the `ContentDigester` field’s documentation?

imagebuildah/stage_executor.go
```diff
@@ -1175,6 +1199,13 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 		// image, run the step so that we can check if the result
 		// matches a cache.
 		if cacheID == "" {
+			// Note: In theory we should never ever hit this branch now
+			// since theoretically only `COPY` and `ADD` should  fall in
+			// this branch but we already process them with execption cases
+			// above to save cycles. But still leave this branch here so if
+			// if we missed anything then we still end up giving our best
+			// effort to find a cache hit.
+
```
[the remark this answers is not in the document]
flouthoc: I am fine to drop this entire branch. This is not needed anyways as i have written in comments.

imagebuildah/stage_executor.go
flouthoc: I am even ok to drop this branch entirely as discussed here: https://github.com/containers/buildah/pull/4128#discussion_r926699466

nalind: Copying the content twice can be noticeably time-consuming, so it's something we should try to avoid.

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
nalind: When would instructions other than ADD and COPY be processed, then?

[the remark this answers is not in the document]
mtrmac: Where is RUN going to happen if this is completely dropped?

imagebuildah/stage_executor.go
```diff
@@ -1175,6 +1199,13 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 		// image, run the step so that we can check if the result
 		// matches a cache.
 		if cacheID == "" {
+			// Note: In theory we should never ever hit this branch now
+			// since theoretically only `COPY` and `ADD` should  fall in
```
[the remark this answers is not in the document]
flouthoc: RUN does not adds any content so I don't know how digest needs to updated. But let me verify this. If there is a `cache` hit for RUN it mostly hits in the first `PASS` althought but `ARG` for sure needs to get evaluated so it might hit this branch.

imagebuildah/stage_executor.go
flouthoc: Actually this might a better condition for this branch `if cacheID == "" && step.Command != "copy" && step.Command != "add" {
`

imagebuildah/stage_executor.go
```diff
@@ -1201,6 +1232,10 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 				}
 			}
 		} else {
+			// This log line is majorly here so we can verify in tests
+			// that our cache is performing in the most optimal way for
+			// various cases.
+			logrus.Debugf("Found a cache hit in the first iteration with id %s", cacheID)
 			// If the instruction would affect our configuration,
```
mtrmac, line 1239: (Unrelated? @nalind As someone unfamiliar with the codebase, the comment about “all that we need” … is unclear to me. Didn’t we get an exact cache hit against the current step’s command, i.e. doesn’t the current `cacheID` image already contain _exactly_ the correct configuration?

It seems that we only ever use the `cacheID` image for the filesystem contents, and we _always_ build the config ourselves in `ib` (at least I can’t see anything that would update things like `ib.Author` from `cacheID`) — so is this comment basically saying “we got the filesystem we want, now update the pending config state ourselves”?

The “_still_ don’t want to restart using the image’s configuration blob” wording somewhat tantalizingly suggests to me that sometimes we _might_ want to restart using it…)

imagebuildah/stage_executor.go
mtrmac: (Personal style, this is up to Buildah’s maintainers, not me:)

I’d find a copy&pasted `step.Command != "..."` hard to maintain. I would very much worry that it is going to get out of sync.

Using a named variable like `canMatchCacheOnlyAfterRun` ensures exact consistency between the code paths, and allows using IDE’s functions to find where it is set and used.

(Also, long comments are _great_, but they still can get out of sync. So, specifically named variables that _precisely_ describe the decision they make, and contain “obviously correct” logic that doesn’t even need a comment, are preferable, if they happen to be possible. But that’s _totally_ a personal style thing.)

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
mtrmac: `RUN dnf install httpd`, if it doesn’t happen to match a cache, absolutely must end up an `ib.Run` call, if I understand what `ib.Run` does correctly.

(And it also definitely adds content, although we don’t compute a digest of that and we don’t seem to restrict cache matches to exact match of the added content.)

imagebuildah/stage_executor.go
nalind: In days when we didn't have robust storage options for unprivileged users, discarding a working container and creating a new one based on a cache hit was expensive compared to merely letting the imagebuilder library update an in-memory configuration structure and moving on.

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
flouthoc: My bad wrong wordings , my intent was that if there was a `cache` hit `RUN` would never hit this branch afaics in case of valid cache hits `RUN` should always get resolved by the first attempt to find the cache. But in case there is no cache hits this branch is still useful.

imagebuildah/stage_executor.go
flouthoc: @mtrmac SGTM. One the core logic in PR gets approved I'll push again with this suggestion.

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
flouthoc: @mtrmac Yes my bad i meant this one: https://github.com/containers/buildah/pull/4128#discussion_r926777116

imagebuildah/stage_executor.go
```diff
@@ -1092,15 +1093,8 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
-		// instruction.
-		if !s.executor.layers {
-			err := ib.Run(step, s, noRunsRemaining)
-			if err != nil {
-				logrus.Debugf("Error building at step %+v: %v", *step, err)
-				return "", nil, fmt.Errorf("error building at STEP \"%s\": %w", step.Message, err)
-			}
-			// In case we added content, retrieve its digest.
+		// Get content summary for cases where we added content, retrieve its digest.
+		getContentSummaryAfterAddingContent := func() string {
```
mtrmac, line 1097: Please turn this into an actual top-level function
  - That makes it, at least in principle, unit-testable
  - It would actually make `Execute` a bit shorter.

(Possibly even method on `CompositeDigester`, but I’m not sure about the debug log output in that case… maybe a method in `stageExecutor` is a viable compromise.)

imagebuildah/stage_executor.go
```diff
@@ -1174,23 +1195,18 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 		// to find the digest of the content to check for a cached
 		// image, run the step so that we can check if the result
 		// matches a cache.
-		if cacheID == "" {
+		if cacheID == "" && step.Command != command.Copy && step.Command != command.Add {
+			// Note: In theory we should never ever hit this branch while
+			// processing `COPY` and `ADD`.
```
mtrmac, line 1200: There is no “in theory” about it given the condition above...

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
flouthoc: Should we discuss here to avoid duplication: https://github.com/containers/buildah/pull/4128#discussion_r926699466 so resolving this thread. But we can re-open if this is different than this https://github.com/containers/buildah/pull/4128#discussion_r926699466

imagebuildah/stage_executor.go
```diff
@@ -867,6 +868,21 @@ func (s *StageExecutor) getImageRootfs(ctx context.Context, image string) (mount
 	return builder.MountPoint, nil
 }
 
+// getContentSummary generates content summary for cases where we added content and need
+// to get summary with updated digests.
+func (s *StageExecutor) getContentSummaryAfterAddingContent() string {
+	addedContentType, addedContentDigest := s.builder.ContentDigester.Digest()
```
mtrmac, line 874: (Absolutely nonblocking: In this now much smaller function, this can use shorter identifiers, e.g.`type`/`digest`/`summary`.)

flouthoc:
@mtrmac @nalind Could you PTAL at the logic now ? Apart from https://github.com/containers/buildah/pull/4128#discussion_r926815184 and https://github.com/containers/buildah/pull/4128#discussion_r926764774 since i'll fix them in the next push.

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
flouthoc: In recent push this block would only match for cases which are not `COPY` or `ADD` and missed the cache lookup in the first pass.

imagebuildah/stage_executor.go
```diff
@@ -1164,6 +1173,24 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 		// determining if a cached layer with the same build args already exists
 		// and that is done in the if block below.
 		if checkForLayers && step.Command != "arg" && !(s.executor.squash && lastInstruction && lastStage) {
+			// Is true if current STEP is `COPY` or `ADD`, denoting that
+			// cache lookup can be attempted only after performing a RUN
```
nalind, line 1177:
```suggestion
			// For `COPY` and `ADD`, history entries include digests computed from
			// the content that's copied in.  We need to compute that information so that
			// it can be used to evaluate the cache, which means we need to go ahead
			// and copy the content.
```

imagebuildah/stage_executor.go
```diff
@@ -1164,6 +1173,24 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
+			canMatchCacheOnlyAfterRun = (step.Command == command.Add ||
+				step.Command == command.Copy)
```
nalind, line 1179:
```suggestion
			canMatchCacheOnlyAfterRun = (step.Command == command.Add || step.Command == command.Copy)
```

imagebuildah/stage_executor.go
```diff
@@ -1164,6 +1173,24 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
+			if canMatchCacheOnlyAfterRun {
+				// There is no way we are going to hit cache for `ADD`
+				// and `COPY` without adding content to find the digest
+				// of the content for which we need to check the cache for
+				// hence we can save cycles if we created `addedContentSummary`
+				// for these special cases by actually adding content so we
+				// end up hitting cache in the first pass.
```
nalind, line 1186:
```suggestion
```

imagebuildah/stage_executor.go
```diff
@@ -1164,6 +1173,24 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
+				if err = ib.Run(step, s, noRunsRemaining); err != nil {
+					logrus.Debugf("Error building at step %+v: %v", *step, err)
+					return "", nil, fmt.Errorf("error building at STEP \"%s\": %w", step.Message, err)
+				}
+				// In case we added content, retrieve its digest.
```
nalind, line 1191:
```suggestion
				// Retrieve the digest info for the content that we just copied into the rootfs
```

imagebuildah/stage_executor.go
```diff
@@ -1174,23 +1201,18 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 		// to find the digest of the content to check for a cached
 		// image, run the step so that we can check if the result
 		// matches a cache.
-		if cacheID == "" {
+		// Ignore for cases covered by `canMatchCacheOnlyAfterRun`
+		// since `Run()` for them is already evaluated in the above
+		// branch.
```
nalind, line 1206:
```suggestion
		// We already called ib.Run() for the `canMatchCacheOnlyAfterRun` cases above, so
		// we shouldn't do it again.
```

imagebuildah/stage_executor.go
flouthoc: resolved.

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
flouthoc: Closing this as resolved.

imagebuildah/stage_executor.go
flouthoc: hmm.. i committed this change and `gofmt` reverted everything back to online weird.

nalind:
I don't think that the commit log and comments in the tests accurately describe what this PR is changing at this point.

flouthoc:
> I don't think that the commit log and comments in the tests accurately describe what this PR is changing at this point.

Amended commit message and comments in test.

nalind:
LGTM
/lgtm

imagebuildah/stage_executor.go
```diff
@@ -1174,23 +1197,17 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
-			addedContentType, addedContentDigest := s.builder.ContentDigester.Digest()
-			addedContentSummary = addedContentType
-			if addedContentDigest != "" {
-				if addedContentSummary != "" {
-					addedContentSummary = addedContentSummary + ":"
-				}
-				addedContentSummary = addedContentSummary + addedContentDigest.Encoded()
-				logrus.Debugf("added content %s", addedContentSummary)
-			}
+			addedContentSummary = s.getContentSummaryAfterAddingContent()
```
mtrmac, line 1210: Why are we searching the cache again? In the `!canMatchCacheOnlyAfterRun` case, we should have had the right `addedContentSummary` initially already; so this shouldn’t find anything new.[1]

Also, the “In case we added content, retrieve its digest” comment is now misleading, that shouldn’t happen / make a difference on this code path.

[1] Unless, I guess, some parallel build has created an equivalent image in the meantime. But that’s _always_ a race.

PR #4130, opened 2022-07-21
Set user namespace defaults correctly for the library

nalind:
#### What type of PR is this?

/kind bug

Set the default for user namespaces correctly for callers that don't use our CLI, e.g. OpenShift or our own tutorials.  When we don't do that, commands invoked through RUN instructions can see weird things, in this case an empty /sys/fs/cgroup directory.

#### How to verify it

New integration test!

This caused test failures in https://github.com/openshift/builder/pull/304 before it started explicitly specifying a user namespace setting to work around it, but it's not the only consumer that might be affected.

```release-note
None
```

### DIFF
--- a/tests/tutorial/tutorial.go
+++ b/tests/tutorial/tutorial.go
@@ -1,0 +1,66 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/containers/buildah"
+	"github.com/containers/buildah/define"
+	"github.com/containers/buildah/imagebuildah"
+	"github.com/containers/storage"
+	"github.com/containers/storage/pkg/unshare"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+)
+
+func main() {
+	if buildah.InitReexec() {
+		return
+	}
+	unshare.MaybeReexecUsingUserNamespace(false)
+
+	buildStoreOptions, err := storage.DefaultStoreOptions(unshare.IsRootless(), unshare.GetRootlessUID())
+	if err != nil {
+		panic(err)
+	}
+
+	buildStore, err := storage.GetStore(buildStoreOptions)
+	if err != nil {
+		panic(err)
+	}
+	defer func() {
+		if _, err := buildStore.Shutdown(false); err != nil {
+			if !errors.Is(err, storage.ErrLayerUsedByContainer) {
+				fmt.Printf("failed to shutdown storage: %q", err)
+			}
+		}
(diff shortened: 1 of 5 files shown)


TomSweeneyRedHat:
Lint isn't happy @nalind, otherwise LGTM

tests/tutorial/tutorial.go
```diff
@@ -0,0 +1,59 @@
...
+	buildStoreOptions, err := storage.DefaultStoreOptions(unshare.IsRootless(), unshare.GetRootlessUID())
+	if err != nil {
+		panic(err)
+	}
+
+	buildStore, err := storage.GetStore(buildStoreOptions)
+	if err != nil {
+		panic(err)
+	}
+	defer buildStore.Shutdown(false)
```
flouthoc, line 33: I think `err` for `) Shutdown(` needs to be checked in order to make linter happy.

flouthoc (changes requested):
LGTM other than one issue above

flouthoc:
Failing conformance is a networking flake restarted.

nalind:
Added an error check, fixed a typo in a comment.

flouthoc (approved):
LGTM
/lgtm

edsantiago:
@lsm5 this is going to need specfile changes:
- easy and recommended IMHO: at install time, `rm .../tutorial.bats`
- hard: build and package the new `tutorial` binary, and make all sorts of changes to gating tests

@nalind do you have any objection to going the easy route? Are the tutorial tests necessary for confidence in Fedora or RHEL environments?

nalind:
This specific bug was caught in OpenShift CI, but openshift/builder only updates to a new version of buildah when someone manually opens a PR, so I don't currently have a clear picture of exactly where between v1.24.2 and v1.26.2 it was that we broke this.  So, in a sense, we already test this case on RHEL, but the feedback is not always timely.
How difficult is the hard route that we're talking about?

edsantiago:
> How difficult is the hard route that we're talking about?

Not really difficult, just complicated: lots of little steps that need to be coordinated in sequence. (That's assuming that it's like the two last helper-binary additions. If there's anything additional, YMMV). I will look into it on Monday. @lsm5, and Jindrich who I can't tag on github: we will all need to coordinate this together. But again, next week.

Thanks @nalind for your fair response.

jnovy:
Just added the buildah-tutorial binary to buildah-tests: https://gitlab.com/redhat/centos-stream/rpms/buildah/-/merge_requests/136

edsantiago:
Thank you @jnovy. @lsm5 can we assume that the same has been done on Fedora?

Issue #4132, opened 2022-07-24
Output for "--output" from `podman build` is only generated when container is not cached


leotaku:
**Description**

The `podman build . --output TARGET` command only produces output if the generated container image is freshly built and not when fetched from cache.

**Steps to reproduce the issue:**
1. Create the following Dockerfile
```Dockerfile
FROM alpine:latest as builder
RUN touch empty.txt
```
3. Run `podman build . --output build`
4. Observe that a new directory named `build` is created
5. Run `podman build . --output build2`
6. Observe that no new directory named `build2` is created

**Describe the results you received:**
Podman only generated the expected output when the container that makes up the output was freshly built. On the second run, where the container had already been built, no output was generated.

**Describe the results you expected:**
Podman always generates the output corresponding to the container described in the Dockerfile.

**Output of `rpm -q buildah` or `apt list buildah`:**
I am not using a Linux variant that uses either of these package managers.

**Output of `buildah version`:**
Not applicable

**Output of `podman version` if reporting a `podman build` issue:**

```
Client:       Podman Engine
Version:      4.1.1
API Version:  4.1.1
Go Version:   go1.18.4
Built:        Tue Jan  1 01:00:00 1980
OS/Arch:      linux/amd64
```

**Output of `cat /etc/*release`:**

```
DISTRIB_CODENAME=raccoon
DISTRIB_DESCRIPTION="NixOS 22.11 (Raccoon)"
DISTRIB_ID=nixos
DISTRIB_RELEASE="22.11"
LSB_VERSION="22.11 (Raccoon)"
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="22.11.20220721.dirty"
DOCUMENTATION_URL="https://nixos.org/learn.html"
HOME_URL="https://nixos.org/"
ID=nixos
LOGO="nix-snowflake"
NAME=NixOS
PRETTY_NAME="NixOS 22.11 (Raccoon)"
SUPPORT_URL="https://nixos.org/community.html"
VERSION="22.11 (Raccoon)"
VERSION_CODENAME=raccoon
VERSION_ID="22.11"
```

**Output of `uname -a`:**

```
Linux nixos-laptop 5.15.55 #1-NixOS SMP Fri Jul 15 08:13:00 UTC 2022 x86_64 GNU/Linux
```

**Output of `cat /etc/containers/storage.conf`:**

```
[storage]
driver = "overlay"
graphroot = "/var/lib/containers/storage"
[...]

Issue #4131, opened 2022-07-24
Output generated using "--output" sometimes contains more files than expected


leotaku:
**Description**

As I understand it, the command `buildah build --output TARGET` should create a directory at the path pointed at by target that contains only the contents of the last container generated by a multi-stage build. In some cases, this happens as expected, in others it does not. I have not yet been able to find out the exact circumstances that cause this issue to appear, but I will provide a minimal Dockerfile that consistently produces this issue for me.

The same issue occurs when using `podman build . --output TARGET`.

The issue also occurs when using the tar output format.

**Steps to reproduce the issue:**
1. Create the following Dockerfile
```Dockerfile
FROM alpine:latest as builder
RUN touch rogue

FROM builder as intermediate
RUN touch artifact

FROM scratch as outputs
COPY --from=intermediate artifact target
```
3. Run `buildah build --output build`
4. Observe the contents of the `build` directory

**Describe the results you received:**
The build directory contains the files `rogue` and `target`, as well as seemingly the contents of a full FHS root. It does not contain the file `artifact`.

**Describe the results you expected:**
The build directory contains only the file `target`.

**Output of `rpm -q buildah` or `apt list buildah`:**
I am not using a Linux variant that uses either of these package managers.

**Output of `buildah version`:**

```
Version:         1.26.2
Go Version:      go1.18.4
Image Spec:      1.0.2-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        1.0.0
libcni Version:  v1.1.0
image Version:   5.21.1
Git Commit:      unknown
Built:           Tue Jan  1 01:00:00 1980
OS/Arch:         linux/amd64
BuildPlatform:   linux/amd64
```

**Output of `podman version` if reporting a `podman build` issue:**

```
Client:       Podman Engine
Version:      4.1.1
API Version:  4.1.1
Go Version:   go1.18.4
Built:        Tue Jan  1 01:00:00 1980
OS/Arch:      linux/amd64
```

**Output of `cat /etc/*release`:**

```
DISTRIB_CODENAME=raccoon
DISTRIB_DESCRIPTION="NixOS 22.11 (Raccoon)"
DISTRIB_ID=nixos
DISTRIB_RELEASE="22.11"
LSB_VERSION="22.11 (Raccoon)"
BUG_REPORT_URL="https://github.com/NixOS/nixpkgs/issues"
BUILD_ID="22.11.20220721.dirty"
DOCUMENTATION_URL="https://nixos.org/learn.html"
[...]

PR #4135, opened 2022-07-26
stage_executor, output: generate correct artifacts for `multi-stage` builds and `cached` builds with `--output`

flouthoc:
Following PR solves two problems with artifacts generated by `--output`

#### Refactor: decouple build output generation and commit.
A bit of code refactoring code. 

#### Generate correct artifact with `--output` for `multi-stage` builds

When processing multi-stage builds and `--output` is specified we tend to
generate `output` for every stage that is being built and getting
committed however we should not do that and we should only commit the
last or the target stage when performing a `mult-stage` builds so
following commit ensures that we only generate output from the `rootfs`
of the last/target stage.

See: https://github.com/containers/buildah/issues/4131

#### Generate artifacts with `--output` even if entire build is cached

In certain cases build are entierly cached and we will never perform
commit for last or final stage. In such cases where we don't get
opportunity to perform commit we must manually process `--output` if it
was specified. Following commit ensures that.

For example a scenario below must produce `output` even though
everything was cached.

Example
```Dockerfile
FROM alpine
RUN echo hello
RUN touch hello
```
```console
buildah build --layers -t test .
buildah build --layers -t test --output build .
```
See: https://github.com/containers/buildah/issues/4132

Closes: https://github.com/containers/buildah/issues/4132
Closes: https://github.com/containers/buildah/issues/4131

```release-note
--output: generates correct artifact for multi-stage builds
--output: generates artifacts even if build is cached
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -994,6 +994,17 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 		}
 	}
 
+	// Parse and populate buildOutputOption if needed
+	var buildOutputOption define.BuildOutputOption
+	canGenerateBuildOutput := (s.executor.buildOutput != "" && lastStage)
+	if canGenerateBuildOutput {
+		logrus.Debugf("Generating custom build output with options %q", s.executor.buildOutput)
+		buildOutputOption, err = parse.GetBuildOutput(s.executor.buildOutput)
+		if err != nil {
+			return "", nil, fmt.Errorf("failed to parse build output: %w", err)
+		}
+	}
+
 	if len(children) == 0 {
 		// There are no steps.
 		if s.builder.FromImageID == "" || s.executor.squash {
@@ -1004,12 +1015,24 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			if imgID, ref, err = s.commit(ctx, s.getCreatedBy(nil, ""), false, s.output, s.executor.squash); err != nil {
 				return "", nil, fmt.Errorf("error committing base container: %w", err)
 			}
+			// Generate build output if needed.
+			if canGenerateBuildOutput {
+				if err := s.generateBuildOutput(buildah.CommitOptions{}, buildOutputOption); err != nil {
+					return "", nil, err
+				}
+			}
 		} else if len(s.executor.labels) > 0 || len(s.executor.annotations) > 0 {
 			// The image would be modified by the labels passed
 			// via the command line, so we need to commit.
 			logCommit(s.output, -1)
 			if imgID, ref, err = s.commit(ctx, s.getCreatedBy(stage.Node, ""), true, s.output, s.executor.squash); err != nil {
 				return "", nil, err
+			}
+			// Generate build output if needed.
+			if canGenerateBuildOutput {
+				if err := s.generateBuildOutput(buildah.CommitOptions{}, buildOutputOption); err != nil {
+					return "", nil, err
+				}
 			}
 		} else {
 			// We don't need to squash the base image, and the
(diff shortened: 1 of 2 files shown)


flouthoc:

rhatdan:
LGTM
@nalind PTAL

imagebuildah/stage_executor.go
```diff
@@ -1348,13 +1348,34 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
+					return "", nil, fmt.Errorf("error committing final squash step %+v: %w", *step, err)
+				}
+			} else if cacheID != "" {
+				// If we found a valid cache hit and this is lastStage
+				// and not a squashed build then there is no oppertunity
+				// for us to perform a `commit` later in the code since
+				// everything will be used from cache.
+				//
+				// If above statement is true and --output was provided
+				// then generate output manually since there is no oppertunity
```
nalind, line 1367:
```suggestion
				// then generate output manually since there is no opportunity
```

imagebuildah/stage_executor.go
```diff
@@ -1948,7 +1969,7 @@ func (s *StageExecutor) commit(ctx context.Context, createdBy string, emptyLayer
 		Manifest:              s.executor.manifest,
 	}
 	// generate build output
-	if s.executor.buildOutput != "" {
+	if s.executor.buildOutput != "" && lastStage {
 		if err := s.generateBuildOutput(buildah.CommitOptions{}, buildOutputOption); err != nil {
```
nalind, line 1973: Does `generateBuildOutput()` benefit from any of the things that `commit()` is doing before this point?  If not, why not just have the rest of the places that call `commit()` call `generateBuildOutput()` directly when they need to, instead of making it a side-effect?

flouthoc: Fixed

nalind:
`StageExecutor.generateBuildOutput()` appears to only ever be called with zeroed-out `CommitOptions`, so it doesn't appear to need to be a parameter.

flouthoc:
> `StageExecutor.generateBuildOutput()` appears to only ever be called with zeroed-out `CommitOptions`, so it doesn't appear to need to be a parameter.

Removed in recent commit.

imagebuildah/stage_executor.go
```diff
@@ -1019,22 +1042,16 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
-				buildOutputOption, err := parse.GetBuildOutput(s.executor.buildOutput)
-				if err != nil {
-					return "", nil, fmt.Errorf("failed to parse build output: %w", err)
-				}
-				if err := s.generateBuildOutput(buildah.CommitOptions{}, buildOutputOption); err != nil {
+			// If we have reached this point then our build is just performing a tag
+			// and it contains no steps or instructions (i.e Containerfile only contains
+			// `FROM <imagename> and nothing else so we will never end up committing this
+			// but instead just re-tag image. For such use-cases if `-o` or `--output` was
+			// specified honor that and export the contents of the current build anyways.
```
TomSweeneyRedHat, line 1049: I'm a little confused by this comment and could just be missing it.  You say in the comment that if the -o or --output option was specified, then we should generate output.  I don't see a check for either option below?  I don't think checking either of those options is part of the canGenerateBuildOutput boolean either.

flouthoc: Oh i see but the comment was already there I just moved it one line above and comment looks still true to me let me know if you think we need to repharse, the branch where we have this check ensures that we are just retagging and will not commit so we must consider generating the output if user specified `--output`, `canGenerateBuildOutput ` ensures if `--output` was specified or not.

flouthoc: Actual change was added here: https://github.com/containers/buildah/pull/4097

PR #4136, opened 2022-07-26
Minor error message fixes

nalind:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

When multiple Dockerfiles are specified using `buildah build`'s `-f` flag, if an error is encountered when parsing one of the files after the first (main) file, the error message would mention the wrong file's name.

#### How to verify it

New integration test!

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

New integration test!

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #4142, opened 2022-07-26
Minor test tweak for podman-remote compatibility

edsantiago:
The new #4136 added a new test (multiple-parse). Make a small
change to it so it passes under podman-remote.

(Reason: the test was checking for a full Dockerfile2 path.
podman-remote only displays basenames).


edsantiago:
Discovered in treadmill. It's perfectly fine to reject this - I can just `skip` the new test in podman-remote.

nalind:
LGTM.  I don't foresee imagebuilder changing the text in that error message in the near future.

TomSweeneyRedHat:
LGTM

PR #4143, opened 2022-07-26
Use errors.Is() instead of os.Is{Not,}Exist

nalind:
#### What type of PR is this?

/kind other

If errors for which `os.IsExist()` or `os.IsNotExist()` would have returned `true` have been wrapped using `fmt.Errorf()`'s "%w" verb, `os.IsExist()` and `os.IsNotExist()`, not having been retrofitted to use `errors.Is()`, will return `false`.

Use `errors.Is()` to check if an error is an `os.ErrExist` or `os.ErrNotExist` error instead of calling `os.IsExist()` or `os.IsNotExist()`.

#### How to verify it

Eyeball it and make sure I didn't get inadvertently reverse anything.  This is currently a preventative change.  All of our tests should continue to pass.  [NO NEW TESTS NEEDED]

None

```release-note
None
```

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -387,7 +387,7 @@ var (
 func fileExistsAndNotADir(path string) bool {
 	file, err := os.Stat(path)
 
-	if file == nil || err != nil || os.IsNotExist(err) {
+	if file == nil || err != nil || errors.Is(err, os.ErrNotExist) {
 		return false
 	}
 	return !file.IsDir()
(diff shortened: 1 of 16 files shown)


TomSweeneyRedHat:
LGTM
one failing test that looked to be a network hiccup, I've restarted

flouthoc (approved):
LGTM

util/util.go
```diff
@@ -387,7 +387,7 @@ var (
 func fileExistsAndNotADir(path string) bool {
 	file, err := os.Stat(path)
 
-	if file == nil || err != nil || os.IsNotExist(err) {
+	if file == nil || err != nil || errors.Is(err, os.ErrNotExist) {
```
Luap99, line 390: This check makes no sense. If the `err != nil` we already go to the return since they are OR return here so there is no point in checking for `ErrNotExist` when err must be nil.

rhatdan:
I am going to merge to get this out and then we can open a PR to fix the currently broken check.

PR #4144, opened 2022-07-27
util, fileExistsAndNotADir: determine a file or directory should print an error

rhatdan:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
none
```
[NO NEW TESTS NEEDED]

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -384,13 +384,15 @@ var (
 
 // fileExistsAndNotADir - Check to see if a file exists
 // and that it is not a directory.
-func fileExistsAndNotADir(path string) bool {
+func fileExistsAndNotADir(path string) (bool, error) {
 	file, err := os.Stat(path)
-
-	if file == nil || err != nil || errors.Is(err, os.ErrNotExist) {
-		return false
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return false, nil
+		}
+		return false, err
 	}
-	return !file.IsDir()
+	return !file.IsDir(), nil
 }
 
 // FindLocalRuntime find the local runtime of the
@@ -404,7 +406,11 @@ func FindLocalRuntime(runtime string) string {
 		return localRuntime
 	}
 	for _, val := range conf.Engine.OCIRuntimes[runtime] {
-		if fileExistsAndNotADir(val) {
+		exists, err := fileExistsAndNotADir(val)
+		if err != nil {
+			logrus.Errorf("Failed to determine if file exists and is not a directory: %v", err)
+		}
+		if exists {
 			localRuntime = val
 			break
 		}


rhatdan:
@nalind @Luap99 @flouthoc @vrothberg PTAL

vrothberg (commented):
LGTM

util/util.go
```diff
@@ -384,13 +384,16 @@ var (
...
 // fileExistsAndNotADir - Check to see if a file exists
 // and that it is not a directory.
-func fileExistsAndNotADir(path string) bool {
+func fileExistsAndNotADir(path string) (bool, error) {
 	file, err := os.Stat(path)
 
-	if file == nil || err != nil || errors.Is(err, os.ErrNotExist) {
-		return false
+	if file == nil || err != nil {
+		if os.IsNotExist(err) {
```
Luap99, line 391: This should use errors.Is()! https://github.com/containers/buildah/pull/4143

util/util.go
```diff
@@ -384,13 +384,16 @@ var (
 
 // fileExistsAndNotADir - Check to see if a file exists
 // and that it is not a directory.
-func fileExistsAndNotADir(path string) bool {
+func fileExistsAndNotADir(path string) (bool, error) {
 	file, err := os.Stat(path)
 
-	if file == nil || err != nil || errors.Is(err, os.ErrNotExist) {
-		return false
+	if file == nil || err != nil {
```
Luap99, line 390: I don't think we have to to check for file ==nil either when we check the error already

util/util.go
rhatdan: Why?  We are checking a returned status from os.Stat()  It should not be a wrapped error.

util/util.go
rhatdan: I left this in case there could be weird error, but I am fine with removing it.

util/util.go
Luap99: > IsNotExist returns a boolean indicating whether the error is known to report that a file or directory does not exist. It is satisfied by ErrNotExist as well as some syscall errors.

> This function predates errors.Is. It only supports errors returned by the os package. New code should use errors.Is(err, fs.ErrNotExist). 

https://pkg.go.dev/os#IsNotExist

The point of that PR was to switch everything to errors.Is, you just change it back for no reason.

rhatdan: Yes I went and read the new comment, and changed it back.

vrothberg (commented):
LGTM

flouthoc (approved):
LGTM
/lgtm
/approve

Issue #4148, opened 2022-07-29
support building from commit (not ref) in git repo


dustymabe:
**Description**

Sometimes it's nice to build directly from a git commit in a git repo.

For example:

```
podman build https://github.com/coreos/coreos-assembler.git#73bdf387a9d25b461f718c76fa3455902acfddfb
Error: error prepping temporary context directory: cloning "https://github.com/coreos/coreos-assembler.git#73bdf387a9d25b461f718c76fa3455902acfddfb" to "/var/tmp/buildah2533916938":
Cloning into '/var/tmp/buildah2533916938'...
fatal: Remote branch 73bdf387a9d25b461f718c76fa3455902acfddfb not found in upstream origin
: exit status 128
```

This shouldn't be too hard to do with something like:

```
mkdir foo
cd foo
git init .
git fetch https://github.com/coreos/coreos-assembler.git 73bdf387a9d25b461f718c76fa3455902acfddfb --depth=1
git checkout FETCH_HEAD
```

System details:

<details>

```
[core@weevm ~]$ podman info
host:
  arch: amd64
  buildahVersion: 1.26.1
  cgroupControllers:
  - memory
  - pids
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.1.0-2.fc36.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.1.0, commit: '
  cpuUtilization:
    idlePercent: 99.83
    systemPercent: 0.06
    userPercent: 0.11
  cpus: 2
  distribution:
    distribution: fedora
    variant: coreos
    version: "36"
  eventLogger: journald
  hostname: weevm
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.18.13-200.fc36.x86_64
  linkmode: dynamic
  logDriver: journald
  memFree: 1025085440
  memTotal: 2064658432
  networkBackend: cni
  ociRuntime:
    name: crun
    package: crun-1.4.5-1.fc36.x86_64
    path: /usr/bin/crun
    version: |-
      crun version 1.4.5
      commit: c381048530aa750495cf502ddb7181f2ded5b400
      spec: 1.0.0
[...]

PR #4149, opened 2022-08-01
build: support building from `commit` when using `git` repo as build context

flouthoc:
When using URL to a git repo as build context allow specifying `commit`
to make sure specific `commit` is used as the build context.

Examples

```console
buildah build -t test myrepo.git#abcef123
buildah build -t test myrepo.git#abcef123
buildah build -t test myrepo.git#abcef123:somesubdir
```
Closes: https://github.com/containers/buildah/issues/4148

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### How to verify it

* Newly added unit tests.
* Newly added integation test.
* Existing tests must pass.

```release-note
build: now supports selecting commit when git repo is used as context
```


define/types.go
```diff
@@ -178,11 +178,22 @@ func TempDirForURL(dir, prefix, url string) (name string, subdir string, err err
 	return "", "", errors.New("unreachable code reached")
 }
 
-func cloneToDirectory(url, dir string) ([]byte, string, error) {
+// ParseGitBuildContext parses git build context to `repo`, `sub-dir`
+// `commit` and `branch`, accepts GitBuildContextin the format of
+// `repourl.git#branch:subdir?commit`.
+//
+// TODO: once following package `define` is refactored this function
```
flouthoc, line 185: As noted here pkg `define` needs a refactor so that functions like `ParseGitBuildContext` can be moved to relevent package. But since its a code move around of lot of existing function in `define/types.go` so will do that in a followup PR after this PR in order to make sure review is easy.

flouthoc:

rhatdan:
LGTM

dustymabe:
> When using URL to a git repo as build context allow specifying `commit` to make sure specific `commit` is used as the build context.
> 
> Following feature is incorporated by adding `?` to already supported syntax for `git` repo as build context.

Is `?` following some established convention? If not maybe we could
consider `@` instead. That seems a bit more intuitive to me.

```
buildah build -t test myrepo.git#@abcef123
buildah build -t test myrepo.git#somebranch@abcef123
buildah build -t test myrepo.git#somebranch:somesubdir@abcef123
```

Also, should we update docs anywhere? Or maybe it's the podman docs we should update?

define/types.go
```diff
@@ -191,16 +202,54 @@ func cloneToDirectory(url, dir string) ([]byte, string, error) {
...
 	if gitBranch == "" {
-		logrus.Debugf("cloning %q to %q", gitBranchPart[0], dir)
-		cmd = exec.Command("git", "clone", "--recurse-submodules", gitBranchPart[0], dir)
+		logrus.Debugf("cloning %q to %q", gitRepo, dir)
+		cmd = exec.Command("git", "clone", "--recurse-submodules", gitRepo, dir)
 	} else {
-		logrus.Debugf("cloning repo %q and branch %q to %q", gitBranchPart[0], gitBranch, dir)
-		cmd = exec.Command("git", "clone", "--recurse-submodules", "-b", gitBranch, gitBranchPart[0], dir)
+		logrus.Debugf("cloning repo %q and branch %q to %q", gitRepo, gitBranch, dir)
+		cmd = exec.Command("git", "clone", "--recurse-submodules", "-b", gitBranch, gitRepo, dir)
```
dustymabe, line 235: When we do builds we typically only need the files from a specific commit in the git repo. Could we re-work this at some point to pull only a specific commit (--depth=0) for efficiency sake?

flouthoc: Sure we could do this If this becomes bottleneck in future just kept it similar to branch pulling for consistency.

flouthoc:
> > When using URL to a git repo as build context allow specifying `commit` to make sure specific `commit` is used as the build context.
> > Following feature is incorporated by adding `?` to already supported syntax for `git` repo as build context.
> 
> Is `?` following some established convention? If not maybe we could consider `@` instead. That seems a bit more intuitive to me.
> 
> ```
> buildah build -t test myrepo.git#@abcef123
> buildah build -t test myrepo.git#somebranch@abcef123
> buildah build -t test myrepo.git#somebranch:somesubdir@abcef123
> ```
> 
> Also, should we update docs anywhere? Or maybe it's the podman docs we should update?

Sure I am fine to use `@` since `docker` does not implements this feature so I think we can choose anything which everyone agrees with.

define/types.go
dustymabe: Thanks! Should I open a feature request?

The reason I think it's so important is that some repos have a really long history and a bunch of vendored deps included. It would be a great improvement to do shallow pulls.

As an example building the coreos-assembler repo if I pull the entire history is like 70M whereas if I just pull a single commit it's like 7M. That's quite a difference.

dustymabe:
What's the difference between 

```
buildah build -t test myrepo.git#@abcef123
```

and

```
buildah build -t test myrepo.git#somebranch@abcef123
```

Will the first one only work if the commit is on the main branch in the repo?

define/types.go
dustymabe: oh that's great!

To be clear I think we should be able to make this optimization for all builds from a git repo. Grab me in `#podman` on IRC if you want to discuss further!

flouthoc: @nalind as @dustymabe suggested could we change `--depth=1` for all use-cases, I am unable to think of any reason that is blocker for this request but just want to confirm if there is any hidden edge case.

dustymabe: Note that supporting `--depth=1` is a bit trickier when someone specifies a commit. i.e. you can't just `git clone`, but you can support it doing something like what I show in https://github.com/containers/buildah/issues/4148#issue-1322488191 with `git init; git fetch, etc..`

nalind:
When `docker build` is told to use a Git repository, it apparently does something more like this:
```console
git init $tempdir
git remote add origin first-part-of-url
git fetch --depth=1 origin -- second-part-of-url-if-given-or-just-"master"
git checkout second-part-of-url-if-given-or-just-"master"
```
This apparently lets the user not care if the second part of the URL is a branch name, a tag, or a commit ID, which is simpler.  It also uses "--depth=1", so we wouldn't be breaking any end-user expectations if we started using it.

nalind:
Or perhaps we can default to `HEAD`, if we're sure it's always going to be pointing to some branch or another, so we don't have to guess whether `master` or `main` is the name of the "default" branch.

rhatdan:
I think using HEAD would be best ide.

define/types.go
nalind: Why export `ParseGitBuildContext()` at all, if it's only going to be called from one place that's in the same package?

nalind:
> Or perhaps we can default to `HEAD`

Okay, `gitrepository-layout(5)` says it should be there in a valid repository, so I think it's safe to assume it will be.

define/types.go
```diff
@@ -178,11 +178,22 @@ func TempDirForURL(dir, prefix, url string) (name string, subdir string, err err
 	return "", "", errors.New("unreachable code reached")
 }
 
-func cloneToDirectory(url, dir string) ([]byte, string, error) {
+// ParseGitBuildContext parses git build context to `repo`, `sub-dir`
+// `commit` and `branch`, accepts GitBuildContextin the format of
```
TomSweeneyRedHat, line 182: nit of a nit
```suggestion
// `commit` and `branch`, accepts GitBuildContext in the format of
```

define/types.go
flouthoc: Made it private.

define/types.go
flouthoc: Implemented in latest commit.

flouthoc:
PTAL again :)

define/types.go
```diff
@@ -191,16 +202,73 @@ func cloneToDirectory(url, dir string) ([]byte, string, error) {
...
-	if gitBranch == "" {
-		logrus.Debugf("cloning %q to %q", gitBranchPart[0], dir)
-		cmd = exec.Command("git", "clone", "--recurse-submodules", gitBranchPart[0], dir)
-	} else {
-		logrus.Debugf("cloning repo %q and branch %q to %q", gitBranchPart[0], gitBranch, dir)
-		cmd = exec.Command("git", "clone", "--recurse-submodules", "-b", gitBranch, gitBranchPart[0], dir)
+	// Attempt parsing commit, accepted syntax listed below:
+	// * repourl.git#branch:subdir@commit
+	// * repourl.git#branch@commit ( uses default subdir )
+	// * repourl.git@commit ( uses default branch )
```
nalind, line 208: Commit IDs can be referred to independently of branches.  We don't need to add "@" to the syntax.

flouthoc: @nalind Yes in previous approach since we were doing `"git", "clone", "--recurse-submodules", "-b"` so branch was always needed but its not needed anymore, amended the PR.

tests/bud.bats
```diff
@@ -1622,13 +1623,18 @@ function _test_http() {
...
   gitrepo=https://github.com/containers/podman.git#main:contrib/hello
   run_buildah build $WITH_POLICY_JSON -t ${target} "${gitrepo}"
   # check syntax only for subdirectory
   gitrepo=https://github.com/containers/podman.git#:contrib/hello
   run_buildah build $WITH_POLICY_JSON -t ${target} "${gitrepo}"
+  # Try pulling repo with specific commit
+  # `contrib/helloimage` is only present on or before `40ba9f10e5fbdd3c9d36389107b8bf1caec6cef0`
+  # hence following test verifies if we are fetching the right commit.
+  gitrepo=https://github.com/containers/podman.git#40ba9f10e5fbdd3c9d36389107b8bf1caec6cef0:contrib/helloimage
+  run_buildah build $WITH_POLICY_JSON -t ${target} "${gitrepo}"
```
nalind, line 1637: Are we testing the combinatoin of a commit/branch ID but no subdirectory name somewhere else?

docs/buildah-build.1.md
```diff
@@ -24,7 +24,7 @@ When the URL is an archive, the contents of the URL is downloaded to a temporary
 
 When the URL is a Containerfile, the file is downloaded to a temporary location.
 
-When a Git repository is set as the URL, the repository is cloned locally and then used as the build context.  A non-default branch and subdirectory of the cloned git repository can be used by including their names at the end of the URL in the form `myrepo.git#mybranch:subdir`, or `myrepo.git#:subdir` if the subdirectory should be used from the default branch.
+When a Git repository is set as the URL, the repository is cloned locally and then used as the build context.  A non-default branch/commit and subdirectory of the cloned git repository can be used by including their names at the end of the URL in the form `myrepo.git#mybranch/mycommit:subdir`, or `myrepo.git#:subdir` if the subdirectory should be used from the default branch.
```
nalind, line 27: `myrepo.git#mybranch/mycommit:subdir` looks ambiguous, possibly reading as though a branch name and commit ID are both required if either is specified.

tests/bud.bats
```diff
@@ -1622,13 +1623,18 @@ function _test_http() {
   run_buildah from ${target}
 }
 
-@test "bud-github-context-with-branch-and-subdir" {
+@test "bud-github-context-with-branch,subdir and commit" {
```
nalind, line 1626:
```suggestion
@test "bud-github-context-with-branch-subdir-commit" {
```

define/types.go
```diff
@@ -191,16 +196,51 @@ func cloneToDirectory(url, dir string) ([]byte, string, error) {
...
+	combinedOutput, err = cmd.CombinedOutput()
+	if err != nil {
+		return combinedOutput, gitSubdir, fmt.Errorf("failed while performing `git fetch`: %w", err)
+	}
+	if fetch == "HEAD" {
+		// We fetched default branch therefore
+		// we don't have any valid `branch` or
+		// `commit` name hence checkout deatched
+		// `FETCH_HEAD`
+		fetch = "FETCH_HEAD"
```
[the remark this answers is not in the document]
flouthoc: Yes it fails if we fetch `HEAD` and try to checkout `HEAD` with `error: pathspec 'HEAD' did not match any file(s) known to git
`

docs/buildah-build.1.md
flouthoc: added a separate example for commit

tests/bud.bats
flouthoc: added a new test for this

tests/bud.bats
```diff
@@ -1620,15 +1621,24 @@ function _test_http() {
   gitrepo=github.com/projectatomic/nulecule-library
   run_buildah build $WITH_POLICY_JSON -t ${target} "${gitrepo}"
   run_buildah from ${target}
+  # Test where context is repourl.git#commit
+  gitrepo=github.com/projectatomic/nulecule-library#fa1c7025262f5d33ba6339dd70c1163c9c26108f
```
nalind, line 1625: Can this be changed to refer to a repository where the latest version of the default branch wouldn't have a Dockerfile, like the commit-and-subdirectory case below?

docs/buildah-build.1.md
```diff
@@ -24,7 +24,7 @@ When the URL is an archive, the contents of the URL is downloaded to a temporary
 
 When the URL is a Containerfile, the file is downloaded to a temporary location.
 
-When a Git repository is set as the URL, the repository is cloned locally and then used as the build context.  A non-default branch and subdirectory of the cloned git repository can be used by including their names at the end of the URL in the form `myrepo.git#mybranch:subdir`, or `myrepo.git#:subdir` if the subdirectory should be used from the default branch.
+When a Git repository is set as the URL, the repository is cloned locally and then used as the build context.  A non-default branch ( or commit ) and subdirectory of the cloned git repository can be used by including their names at the end of the URL in the form `myrepo.git#mybranch:subdir`, `myrepo.git#mycommit:subdir`, or `myrepo.git#:subdir` if the subdirectory should be used from the default branch.
```
nalind, line 27:
```suggestion
When a Git repository is set as the URL, the repository is cloned locally and then used as the build context.  A non-default branch (or commit ID) and subdirectory of the cloned git repository can be used by including their names at the end of the URL in the form `myrepo.git#mybranch:subdir`, `myrepo.git#mycommit:subdir`, or `myrepo.git#:subdir` if the subdirectory should be used from the default branch.
```

define/types.go
```diff
@@ -178,11 +178,16 @@ func TempDirForURL(dir, prefix, url string) (name string, subdir string, err err
 	return "", "", errors.New("unreachable code reached")
 }
 
-func cloneToDirectory(url, dir string) ([]byte, string, error) {
+// parseGitBuildContext parses git build context to `repo`, `sub-dir`
+// `branch/commit`, accepts GitBuildContext in the format of
+// `repourl.git#branch/commit:subdir`.
+func parseGitBuildContext(url string) (string, string, string) {
```
nalind, line 184: Does this need to be its own function any more, or can it be inlined back into `cloneToDirectory()`?

flouthoc: Just moved it out so unit tests can be added for parsing.

define/types.go
```diff
@@ -191,16 +196,51 @@ func cloneToDirectory(url, dir string) ([]byte, string, error) {
...
+	cmd = exec.Command("git", "fetch", "--depth=1", "origin", "--", fetch)
+	cmd.Dir = dir
+	combinedOutput, err = cmd.CombinedOutput()
+	if err != nil {
+		return combinedOutput, gitSubdir, fmt.Errorf("failed while performing `git fetch`: %w", err)
+	}
+	if fetch == "HEAD" {
+		// We fetched default branch therefore
+		// we don't have any valid `branch` or
+		// `commit` name hence checkout deatched
```
nalind, line 233:
```suggestion
		// `commit` name hence checkout detached
```

define/types.go
```diff
@@ -191,16 +196,51 @@ func cloneToDirectory(url, dir string) ([]byte, string, error) {
 		}
 		gitBranch = gitSubDirPart[0]
 	}
-	if gitBranch == "" {
-		logrus.Debugf("cloning %q to %q", gitBranchPart[0], dir)
-		cmd = exec.Command("git", "clone", "--recurse-submodules", gitBranchPart[0], dir)
-	} else {
-		logrus.Debugf("cloning repo %q and branch %q to %q", gitBranchPart[0], gitBranch, dir)
-		cmd = exec.Command("git", "clone", "--recurse-submodules", "-b", gitBranch, gitBranchPart[0], dir)
-	}
```
nalind, line 200: There should continue to be a debug message somewhere in here.

tests/bud.bats
```diff
@@ -1620,15 +1621,24 @@ function _test_http() {
...
+  run_buildah build $WITH_POLICY_JSON -t ${target} "${gitrepo}"
+  run_buildah from ${target}
```
nalind, line 1627: Try to use a different name for the image this line builds than was used for the previous image built in this test.

nalind:
The special case for locations that start with "github.com/" that starts at line 147 either needs to be change to let them be handled by the branch that clones the repository, or the new build being done in the "bud-github-context" test needs to be changed to specify the build context location with the "git://" prefix.
The check for a ".git" suffix at line 132 wasn't updated when we added subdirectory support.  We can either fix that now or later.

tests/bud.bats
flouthoc: this is a new test now using local git repo

define/types.go
flouthoc: Added equivalent log line.

tests/bud.bats
flouthoc: Added test from local repo.

flouthoc:
@nalind PTAL, new test uses a `local` git repo with `git-daemon` and verifies that `Containerfle` exists on older commit and not on `main` branch.

define/types.go
```diff
@@ -186,16 +191,52 @@ func cloneToDirectory(url, dir string) ([]byte, string, error) {
...
+	if err != nil {
+		return combinedOutput, gitSubdir, fmt.Errorf("failed while performing `git remote add`: %w", err)
+	}
+	// fetch required branch or commit and perform checkout
+	// Always default to `HEAD` if nothing specified
+	fetch := "HEAD"
+	if gitBranch != "" {
+		fetch = gitBranch
+	}
+	logrus.Debugf("fetching repo %q and branch (or commit ID) %q to %q", gitRepo, gitBranch, dir)
```
nalind, line 219:
```suggestion
	logrus.Debugf("fetching repo %q and branch (or commit ID) %q to %q", gitRepo, fetch, dir)
```
`gitBranch` could be empty, and it's not what we're passing to `git`.

tests/bud.bats
```diff
@@ -1640,13 +1641,42 @@ function _test_http() {
...
+  # Containerfile in this repo should only exists on older commit and
+  # not on HEAD or the default branch.
+  gitrepo=git://localhost:${GITPORT}/repo#f94193d34548eb58650a10a5183936d32c2d3280
+  tmpdir="${TEST_SCRATCH_DIR}/build"
+  mkdir -p "${tmpdir}"
+  TMPDIR="${tmpdir}" run_buildah build $WITH_POLICY_JSON -t ${target} "${gitrepo}"
+  # Verify that build must fail on default `main` branch since we
+  # don't have a `Containerfile` on main branch.
+  gitrepo=git://localhost:${GITPORT}/repo#main
+  TMPDIR="${tmpdir}" run_buildah 125 build $WITH_POLICY_JSON -t ${target} "${gitrepo}"
```
nalind, line 1665: I don't see the reason for using a different $TMPDIR for these two build attempts.  #4159 did it so that it could check that the directory the repository would be cloned into was empty when the build finished, but we're not doing that here.

define/types.go
```diff
@@ -173,11 +173,15 @@ func TempDirForURL(dir, prefix, url string) (name string, subdir string, err err
...
+// parseGitBuildContext parses git build context to `repo`, `sub-dir`
+// `branch/commit`, accepts GitBuildContext in the format of
+// `repourl.git#branch/commit:subdir`.
+func parseGitBuildContext(url string) (string, string, string) {
 	gitSubdir := ""
 	gitBranch := ""
 	gitBranchPart := strings.Split(url, "#")
-	var cmd *exec.Cmd
+	// Attempt parsing commit, accepted syntax listed below:
+	// `repourl.git#branch(or commit ID):subdir` or `repourl.git#:subdir` ( uses default branch )
```
nalind, line 184: This doesn't tell me anything that the function's godoc doesn't.  It's redundant.

tests/bud.bats
```diff
@@ -1640,13 +1641,40 @@ function _test_http() {
...
+  target=giturl-image
+  # Any repo would do, but this one is small, is FROM: scratch, local, and has
+  # its entire build context in a subdirectory of the repository.
+  if ! start_git_daemon ${TEST_SOURCES}/git-daemon/repo-with-containerfile-on-old-commit.tar.gz ; then
+    skip "error running git daemon"
+  fi
+  # Containerfile in this repo should only exists on older commit and
+  # not on HEAD or the default branch.
+  gitrepo=git://localhost:${GITPORT}/repo#f94193d34548eb58650a10a5183936d32c2d3280
+  run_buildah build $WITH_POLICY_JSON -t ${target} "${gitrepo}"
```
nalind, line 1659: Can we check for a specific error message here?

tests/bud.bats
```diff
@@ -1640,13 +1641,40 @@ function _test_http() {
...
+  # Verify that build must fail on default `main` branch since we
+  # don't have a `Containerfile` on main branch.
+  gitrepo=git://localhost:${GITPORT}/repo#main
+  run_buildah 125 build $WITH_POLICY_JSON -t ${target} "${gitrepo}"
```
nalind, line 1663: Can we check for a specific error message here?

tests/bud.bats
```diff
@@ -1640,13 +1641,40 @@ function _test_http() {
...
+  if ! which git ; then
+    skip "no git in PATH"
+  fi
+  target=giturl-image
+  # Any repo would do, but this one is small, is FROM: scratch, local, and has
+  # its entire build context in a subdirectory of the repository.
+  if ! start_git_daemon ${TEST_SOURCES}/git-daemon/repo-with-containerfile-on-old-commit.tar.gz ; then
+    skip "error running git daemon"
+  fi
+  # Containerfile in this repo should only exists on older commit and
```
nalind, line 1656:
```suggestion
  # Containerfile in this repo should only exist on older commit and
```

tests/bud.bats
flouthoc: Yes checking below.

tests/bud.bats
flouthoc: Success build is enough to confirm that `Containerfile` was found and verifies the test case but I have added strings which should be part of build output as a double check.

nalind: Ah, right, this one's expected to succeed.

tests/bud.bats
```diff
@@ -1640,13 +1641,43 @@ function _test_http() {
   run_buildah from ${target}
 }
 
-@test "bud-github-context-with-branch-and-subdir" {
+# Containerfile in this repo should only exists on older commit and
```
nalind, line 1644:
```suggestion
# Containerfile in this repo should only exist on older commit and
```

define/types.go
```diff
@@ -173,11 +173,13 @@ func TempDirForURL(dir, prefix, url string) (name string, subdir string, err err
 	return "", "", errors.New("unreachable code reached")
 }
 
-func cloneToDirectory(url, dir string) ([]byte, string, error) {
+// parseGitBuildContext parses git build context to `repo`, `sub-dir`
+// `branch/commit`, accepts GitBuildContext in the format of
+// `repourl.git#branch/commit:subdir`.
```
nalind, line 178:
```suggestion
// `repourl.git[#[branch-or-commit]:subdir]`.
```
Optional change, to note that the subdir and branch-or-commit parts are optional.

flouthoc:

rhatdan:
/lgtm

We need to cut v1.27.0

PR #4154, opened 2022-08-01
[release-1.26] set user namespace defaults correctly for the library

nalind:
#### What type of PR is this?

/kind bug

Set the default for user namespaces correctly for callers that don't use our CLI, e.g. OpenShift or our own tutorials.  When we don't do that, commands invoked through RUN instructions can see weird things, in this case an empty /sys/fs/cgroup directory.

#### How to verify it

New integration test!

This caused test failures in https://github.com/openshift/builder/pull/304 before it started explicitly specifying a user namespace setting to work around it, but it's not the only consumer that might be affected.

Cherry-picked from https://github.com/containers/buildah/pull/4130, with a typo fixed.

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ endif
 #     Note: Uses the -N -l go compiler options to disable compiler optimizations
 #           and inlining. Using these build options allows you to subsequently
 #           use source debugging tools like delve.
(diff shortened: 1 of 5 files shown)


flouthoc (approved):
LGTM

TomSweeneyRedHat:
LGTM

PR #4156, opened 2022-08-01
integration tests: quote "?" in shell scripts

nalind:
#### What type of PR is this?

/kind bug

When passing the value ? to a shell function as an argument, don't forget to quote it, so that it doesn't get expanded as a filename pattern by the shell.

#### How to verify it

None

```release-note
None
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -4090,7 +4090,7 @@ _EOF
     found_runtime=y
     if is_cgroupsv2; then
       # The result with cgroup v2 depends on the version of runc.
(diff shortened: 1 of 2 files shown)


vrothberg (commented):
LGTM
@edsantiago @flouthoc PTAL

flouthoc (approved):
LGTM @edsantiago PTAL

edsantiago:
/lgtm

and I see no other instances. Nice catch, @nalind.

PR #4159, opened 2022-08-02
build: clean up git repos correctly when using subdirs

nalind:
#### What type of PR is this?

/kind bug

When the build context is specified as a subdirectory in a Git repository, we were cloning the repository to start, but only removing
the subdirectory when we were finished with it.  We should be removing the whole clone.

#### How to verify it

New integration test!

None

```release-note
When the build context is a subdirectory of a Git repository which needs to be cloned, the temporary copy of the repository is now properly cleaned up.
```


flouthoc (approved):
LGTM
/approve
/lgtm

PR #4169, opened 2022-08-05
Don't try to call runLabelStdioPipes if spec.Linux is not set

dfr:
On FreeBSD, the Linux section of the spec is not populated. FreeBSD does
have a similar labelling facility in its MAC framework but that would be
better managed via a future addition of a FreeBSD section to the runtime
spec rather than trying to make it look like Linux.

#### What type of PR is this?

> /kind bug

This fixes non-tty run operations with buildah on FreeBSD. This got missed when turning the original prototype into a reviewable form.

#### How to verify it

On a FreeBSD system do this:
```
c=$(sudo buildah from quay.io/dougrabson/freebsd-minimal:13.1)
sudo buildah run --tty=false $c freebsd-version
sudo buildah rm $c
```

None

```release-note
None
```

### DIFF
--- a/run_common.go
+++ b/run_common.go
@@ -477,8 +477,10 @@ func runUsingRuntime(options RunOptions, configureNetwork bool, moreCreateArgs [
 			if stdioPipe, err = runMakeStdioPipe(int(uid), int(gid)); err != nil {
 				return 1, err
 			}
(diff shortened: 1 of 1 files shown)


flouthoc:
@dfr Could you please add `[NO NEW TESTS NEEDED]` to the commit message.

flouthoc (approved):
LGTM
/lgtm
/approve

Issue #4167, opened 2022-08-04
Inconsistent behavior when COPYing multiple files to /dest without trailing slash


chmeliik:
According to https://docs.docker.com/engine/reference/builder/#copy,
> If multiple `<src>` resources are specified, either directly or due to the use of a wildcard, then `<dest>` must be a directory, and it must end with a slash `/`.

For a COPY instruction like `COPY spam/* /spam` or `COPY spam/a.txt spam/b.txt spam/c.txt /spam`, docker fails the build saying the destination needs a trailing slash.

Buildah does not fail the build, which is probably fine, but the behavior differs between using a wildcard and listing them all explicitly. With a wildcard, only one item gets copied and becomes /dest. When listing explicitly, all the items get copied into /dest/.

**Steps to reproduce the issue:**
1. build inputs:

Dockerfile
```dockerfile
FROM alpine:latest

COPY spam/* /spam
# COPY spam/a.txt spam/b.txt spam/c.txt /spam
```
spam/a.txt
```
Listen -- strange women lying in ponds distributing swords is no basis for a system of government. Supreme executive power derives from a mandate from the masses, not from some farcical aquatic ceremony.
```
spam/b.txt
```
I mean, if I went around sayin' I was an emperor just because some moistened bint had lobbed a scimitar at me they'd put me away!
```
spam/c.txt
```
Ah, now we see the violence inherent in the system. Oh! Come and see the violence inherent in the system! HELP! HELP! I'm being repressed!
```
2. run `buildah bud -t test .`
3. run `podman run --rm -ti test:latest find /spam -type f -exec echo {} + -exec cat {} +`
```
/spam
Ah, now we see the violence inherent in the system. Oh! Come and see the violence inherent in the system! HELP! HELP! I'm being repressed!
```
4. comment the first COPY, uncomment the second, try again
```
/spam/a.txt /spam/b.txt /spam/c.txt
Listen -- strange women lying in ponds distributing swords is no basis for a system of government. Supreme executive power derives from a mandate from the masses, not from some farcical aquatic ceremony.
I mean, if I went around sayin' I was an emperor just because some moistened bint had lobbed a scimitar at me they'd put me away!
Ah, now we see the violence inherent in the system. Oh! Come and see the violence inherent in the system! HELP! HELP! I'm being repressed!
```

**Describe the results you received:**

Inconsistent behavior between COPY with wildcard and COPY with multiple sources listed explicitly

**Describe the results you expected:**

[...]

PR #4183, opened 2022-08-11
add,copy: if multiple sources are used in `ADD` or `COPY` then dest must end with a `slash`

flouthoc:
When using `ADD` or `COPY` enforce the condition that destination must
look like a directory by making sure that destination ends with a slash
`/`.

This ensures that we don't hit undefined behaviour for example with
the use case `COPY some/* /oncontainer` it is not clear that
`oncontainer` will be a directory or a file, users expect it to be a
directory if `*` wildcard resolves to multiple files and logically if
wildcard resolves to single file then it should be a directory, since
this condition is undefined and not in parity with docker so lets drop
support for it.

Docker's defined rule:
* If multiple `<src>` resources are specified, either directly or due to the 
  use of a wildcard, then `<dest>` must be a directory, and it must end with a slash `/`.

Reference: https://docs.docker.com/engine/reference/builder/#add

Closes: https://github.com/containers/buildah/issues/4167


add.go
```diff
@@ -188,10 +188,39 @@ func includeDirectoryAnyway(path string, pm *fileutils.PatternMatcher) bool {
...
+// Following function can contain more validation cases in future.
+func validateSourceAndDest(destination string, sources []string) error {
+	multipleSources := false
+	if len(sources) > 1 {
+		multipleSources = true
+	} else {
+		// Check length before accessing `0`
+		// to avoid `index out of range`.
+		if len(sources) == 1 {
+			if strings.HasSuffix(sources[0], "*") {
```
rhatdan, line 204: Should you check for len(sources)==0 and return an error?

add.go
```diff
@@ -188,10 +188,39 @@ func includeDirectoryAnyway(path string, pm *fileutils.PatternMatcher) bool {
...
+				multipleSources = true
+			}
+		}
+	}
+	if multipleSources {
+		if !strings.HasSuffix(destination, "/") {
```
rhatdan, line 210: Does docker enforce this?  Or does it just check if destination is not a directory then throw an error?

chmeliik: Just chiming in that if the directory already exists, podman/buildah does the expected (?) thing and copies all the matches of the wildcard into the existing directory

Re-using the reproducer from #4167, if you modify the Dockerfile to this
```dockerfile
FROM alpine:latest as builder

RUN mkdir /spam
COPY spam/*.txt /spam
```
Then it copies all the files

add.go
nalind: Why is a suffix of "*" the special case here?

tests/add.bats
```diff
@@ -28,11 +28,13 @@ load helpers
   # Copy a file to a specific subdirectory
   run_buildah add $cid ${TEST_SCRATCH_DIR}/randomfile /subdir
   # Copy two files to a specific subdirectory
-  run_buildah add $cid ${TEST_SCRATCH_DIR}/randomfile ${TEST_SCRATCH_DIR}/other-randomfile /other-subdir
+  run_buildah add $cid ${TEST_SCRATCH_DIR}/randomfile ${TEST_SCRATCH_DIR}/other-randomfile /other-subdir/
   # Copy two files to a specific location, which succeeds because we can create it as a directory.
-  run_buildah add $cid ${TEST_SCRATCH_DIR}/randomfile ${TEST_SCRATCH_DIR}/other-randomfile /notthereyet-subdir
+  run_buildah add $cid ${TEST_SCRATCH_DIR}/randomfile ${TEST_SCRATCH_DIR}/other-randomfile /notthereyet-subdir/
+  # Add with a wildcard and destination dir does not exists
+  run_buildah add $cid ${TEST_SCRATCH_DIR}/* /notthereyet2-subdir/
```
nalind, line 35: Without quoting, the globbing here is done by the shell, not by buildah, so buildah doesn't see the globbing pattern.

tests/add.bats
```diff
@@ -58,6 +60,22 @@ load helpers
...
+  createrandom ${TEST_SCRATCH_DIR}/randomfile
+  createrandom ${TEST_SCRATCH_DIR}/other-randomfile
+  createrandom ${TEST_SCRATCH_DIR}/third-randomfile
+
+  run_buildah from $WITH_POLICY_JSON scratch
+  cid=$output
+  run_buildah mount $cid
+  root=$output
+  run_buildah config --workingdir / $cid
+  run_buildah 125 add $cid ${TEST_SCRATCH_DIR}/* /test
```
nalind, line 73: Without quoting, the globbing here is done by the shell, not by buildah, so buildah doesn't see the globbing pattern.

add.go
flouthoc: Special case is here to match with the current behavior of docker https://github.com/containers/buildah/issues/4167#issuecomment-1212098545

add.go
flouthoc: I think docker checks this cause rule is
```
* If multiple <src> resources are specified, either directly or due to the use of a wildcard, then <dest> must be a directory, and it must end with a slash /.
```
https://docs.docker.com/engine/reference/builder/#add

rhatdan: I still believe this should be done on in the copier.

Basically if the SRC contains more then one file and the DEST is not a directory, then it should return an error.

The directory should not be checked for ending with /

nalind: We're already globbing inside a chroot further down, when we call `copier.Stat()`, and we're probably better off doing this sort of check just after that, when we have all of the facts.  Doing this before that means that the current iteration of this patch passes URLs for remote locations to `filepath.Glob()`, which is a bug.

rhatdan: I guess they do document the following:
```
If <dest> does not end with a trailing slash, it will be considered a regular file and the contents of <src> will be written at <dest>.
```

But @nalind question above has not been addressed

chmeliik: > Basically if the SRC contains more then one file and the DEST is not a directory, then it should return an error.
>
> The directory should not be checked for ending with /

+1 to this, it sounds reasonable to me that if the destination already exists then the trailing slash doesn't need to be enforced. It wouldn't be 100% consistent with docker, but isn't it likely that some users are already doing this? E.g. copying things to `/etc` or other standard dirs

flouthoc: @chmeliik Yes that is already addressed I just need to move code around as per @nalind 's suggestion here https://github.com/containers/buildah/pull/4183#discussion_r953873090 , I'll amend this.

rhatdan:
@flouthoc any chance you get get back to this one?

flouthoc:
> @flouthoc any chance you get get back to this one?

Sure yes, I'll amend this again in this week.

rhatdan:
@flouthoc Reminder

rhatdan:
@flouthoc after the release lets get back to this one.

flouthoc:
@rhatdan Yes getting back on this and closing this one.

add.go
flouthoc: Done moved to just after we split local and remote sources and only validating local sources now.

flouthoc:
@nalind @rhatdan PTAL

TomSweeneyRedHat:
LGTM, but would like a head nod from @vrothberg or @nalind to make sure comments were addressed.

rhatdan:
This still needs to be reworked.

flouthoc:
Just going through my old PRs, I think this one is on pending review.

rhatdan:
Well needs to be rebased anyways.

PR #4189, opened 2022-08-16
cmd/buildah: add "manifest create --amend"

nalind:
#### What type of PR is this?

/kind feature

When `buildah manifest create` is given the `--amend` flag and a list with the name that was specified for the to-be-created list already exists, just reuse the list.

Make the `--insecure` flag, if we see it, conflict with `--tls-verify`, but have the reverse of the effect that `--tls-verify` does, and teach the `buildah manifest` `add`, `create`, and `push` subcommands about it.

#### How to verify it

Extended integration test!

Related to https://github.com/containers/podman/issues/15346.

```release-note
`buildah manifest create` now accepts a `--amend` flag.
```

### DIFF
--- a/cmd/buildah/manifest.go
+++ b/cmd/buildah/manifest.go
@@ -30,13 +30,13 @@ import (
 )
 
 type manifestCreateOpts = struct {
-	os, arch string
-	all      bool
+	os, arch                        string
+	all, tlsVerify, insecure, amend bool
 }
 type manifestAddOpts = struct {
 	authfile, certDir, creds, os, arch, variant, osVersion string
 	features, osFeatures, annotations                      []string
-	tlsVerify, all                                         bool
+	tlsVerify, insecure, all                               bool
 }
 type manifestRemoveOpts = struct{}
 type manifestAnnotateOpts = struct {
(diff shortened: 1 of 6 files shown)


flouthoc (approved):
LGTM @containers/buildah-maintainers PTAL

cmd/buildah/manifest.go
```diff
@@ -101,6 +102,11 @@ func init() {
 	if err := flags.MarkHidden("arch"); err != nil {
 		panic(fmt.Sprintf("error marking --arch as hidden: %v", err))
 	}
+	flags.BoolVar(&manifestCreateOpts.insecure, "insecure", false, "neither require HTTPS nor verify certificates when accessing the registry. TLS verification cannot be used when talking to an insecure registry.")
```
TomSweeneyRedHat, line 105: soft suggestion for consideration
```suggestion
	flags.BoolVar(&manifestCreateOpts.insecure, "insecure", false, "neither require HTTPS nor verification of certificates when accessing the registry. TLS verification cannot be used when talking to an insecure registry.")
```

TomSweeneyRedHat: Then down below too if you take it.

TomSweeneyRedHat:
one small doc tweak for consideration or to be ignored, your call.

LGTM otherwise.

cmd/buildah/manifest.go
rhatdan: Since it is hidden, it really does not matter, nobody will ever see it.

PR #4190, opened 2022-08-17
commit: use race-free `RemoveNames` instead of `SetNames` for pruning names

flouthoc:
PR https://github.com/containers/storage/pull/1153 added a dedicated API
to remove names assigned to image so use `RemoveNames`
instead of racy `SetNames`.

##### How to verify

* Refer to newly added integration test.

##### Reproduce on local using

```console
printf 'from quay.io/jitesoft/alpine:latest\nrun for i in $(seq 0 10000); do touch /$i; done\n' >Containerfile && for i in `seq 1 25`; do ./buildah build --squash --iidfile id.$i --timestamp 0 . & done; wait; ls -al
```

Closes: https://github.com/containers/podman/issues/15162

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -390,6 +390,24 @@ _EOF
   assert "$output" !~ "unwanted stage"
 }
 
+# Note: Please skip this tests in case of podman-remote build
+@test "build: test race in updating image name while performing parallel commits" {
+  _prefetch alpine
+  # Run 25 parallel builds using the same Containerfile
+  local count=25
+  for i in $(seq --format '%02g' 1 $count); do
+      timeout --foreground -v --kill=10 300 \
+              ${BUILDAH_BINARY} ${BUILDAH_REGISTRY_OPTS} ${ROOTDIR_OPTS} $WITH_POLICY_JSON build --quiet --squash --iidfile ${TEST_SCRATCH_DIR}/id.$i --timestamp 0 -f $BUDFILES/check-race/Containerfile >/dev/null &
+  done
+  # Wait for all background builds to complete. Note that this succeeds
+  # even if some of the individual builds fail! Our actual test is below.
+  wait
+  # Number of output bytes must be always same, which confirms that there is no race.
+  assert "$(cat ${TEST_SCRATCH_DIR}/id.* | wc -c)" = 1775 "Total chars in all id.* files"
+  # clean all images built for this test
+  run_buildah rmi --all -f
+}
+
 # Test skipping images with FROM but stage name also conflicts with additional build context
 # so selected stage should be still skipped since it is not being actually used by additional build
 # context is being used.
--- a/commit.go
+++ b/commit.go
@@ -374,17 +374,17 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 	}
 	if err == nil {
 		imgID = img.ID
-		prunedNames := make([]string, 0, len(img.Names))
+		toPruneNames := make([]string, 0, len(img.Names))
 		for _, name := range img.Names {
-			if !(nameToRemove != "" && strings.Contains(name, nameToRemove)) {
-				prunedNames = append(prunedNames, name)
+			if nameToRemove != "" && strings.Contains(name, nameToRemove) {
+				toPruneNames = append(toPruneNames, name)
 			}
 		}
-		if len(prunedNames) < len(img.Names) {
-			if err = b.store.SetNames(imgID, prunedNames); err != nil {
-				return imgID, nil, "", fmt.Errorf("failed to prune temporary name from image %q: %w", imgID, err)
+		if len(toPruneNames) > 0 {
+			if err = b.store.RemoveNames(imgID, toPruneNames); err != nil {
+				return imgID, nil, "", fmt.Errorf("failed to remove temporary name from image %q: %w", imgID, err)
 			}
-			logrus.Debugf("reassigned names %v to image %q", prunedNames, img.ID)
+			logrus.Debugf("removing %v from assigned names to image %q", nameToRemove, img.ID)
 			dest2, err := is.Transport.ParseStoreReference(b.store, "@"+imgID)
 			if err != nil {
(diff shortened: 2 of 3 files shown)


flouthoc:
The test passes on my local I wonder what did i do wrong here that it is failing on CI. @edsantiago Could help me by providing any hints ? 

My local output
```console
++ time bats --tap . --filter "performing parallel commits"
1..1
ok 1 build: test race in updating image name while performing parallel commits

real	0m34.193s
user	2m13.625s
sys	1m22.299s
```

flouthoc:
Assuming its the wildcard causing issues in CI apply this
```diff
diff --git a/tests/bud.bats b/tests/bud.bats
index 562414e4..70fc9146 100644
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -402,7 +402,9 @@ _EOF
   # even if some of the individual builds fail! Our actual test is below.
   wait
   # Number of output bytes must be always same, which confirms that there is no race.
-  test $(cat ${TEST_SCRATCH_DIR}/id*|wc -c) = 1775
+  for i in $(seq --format '%02g' 1 $count); do
+      test $(cat ${TEST_SCRATCH_DIR}/id.$i|wc -c) = 71
+  done
   # clean all images built for this test
   run_buildah rmi --all -f
 }
 ```

**Review by giuseppe — APPROVED** [review]
LGTM

File: tests/bud.bats [hunk 1]
```diff
@@ -390,6 +390,25 @@ _EOF
...
+  for i in $(seq --format '%02g' 1 $count); do
+      timeout --foreground -v --kill=10 60 \
+              ${BUILDAH_BINARY} ${BUILDAH_REGISTRY_OPTS} ${ROOTDIR_OPTS} $WITH_POLICY_JSON build --squash --iidfile ${TEST_SCRATCH_DIR}/id.$i --timestamp 0 -f $BUDFILES/check-race/Containerfile &>/dev/null &
+  done
+  # Wait for all background builds to complete. Note that this succeeds
+  # even if some of the individual builds fail! Our actual test is below.
+  wait
+  # Number of output bytes must be always same, which confirms that there is no race.
+  for i in $(seq --format '%02g' 1 $count); do
+      test $(cat ${TEST_SCRATCH_DIR}/id.$i|wc -c) = 71
```
edsantiago, line 406: Please do not use `test`. Do not ever use `test`. That gives no information whatsoever to the poor person trying to debug a test failure. Please use `assert`.

edsantiago (changes requested):

tests/bud.bats
edsantiago: So, going back to the first question, you can rewrite it the first way as
```bash
    assert "$(cat ${TEST_SCRATCH_DIR}/id.* | wc -c)" = 1775 "Total chars in all id.* files"
```

flouthoc: Thanks @edsantiago , Amended, I'll wait and hope CI goes green now ( passes on my local )

edsantiago (approved):

tests/bud.bats
```diff
@@ -390,6 +390,23 @@ _EOF
   assert "$output" !~ "unwanted stage"
 }
 
+# Note: Please skip this tests in case of podman-remote build
+@test "build: test race in updating image name while performing parallel commits" {
+  # Run 25 parallel builds using the same Containerfile
```
edsantiago, line 395: Try `_prefetch alpine`?

Can't help you with the rest of the errors, though.

tests/bud.bats
```diff
@@ -390,6 +390,23 @@ _EOF
...
+  local count=25
+  for i in $(seq --format '%02g' 1 $count); do
+      timeout --foreground -v --kill=10 60 \
+              ${BUILDAH_BINARY} ${BUILDAH_REGISTRY_OPTS} ${ROOTDIR_OPTS} $WITH_POLICY_JSON build --squash --iidfile ${TEST_SCRATCH_DIR}/id.$i --timestamp 0 -f $BUDFILES/check-race/Containerfile &>/dev/null &
```
edsantiago, line 399: Also, you may want to change the above to `... --quiet ... >/dev/null` (add `--quiet`, remove `&` so stderr is preserved).

tests/bud.bats
flouthoc: Strange part is the test passes on local and fails on CI across all VM, which makes me feel like something in the test is not right and only available on my local. I hope its the missing `_prefetch alpine`. Amended.

flouthoc: I think the issue is we have timeout too short for CI
```console
[+0438s] # timeout: sending signal TERM to command ‘/var/tmp/go/src/github.com/containers/buildah/tests/./../bin/buildah’
[+0438s] # timeout: sending signal TERM to command ‘/var/tmp/go/src/github.com/containers/buildah/tests/./../bin/buildah’
[+0438s] # timeout: sending signal TERM to command ‘/var/tmp/go/src/github.com/containers/buildah/tests/./../bin/buildah’
[+0438s] # timeout: sending signal TERM to command ‘/var/tmp/go/src/github.com/containers/buildah/tests/./../bin/buildah’
[+0438s] # timeout: sending signal TERM to command ‘/var/tmp/go/src/github.com/containers/buildah/tests/./../bin/buildah’
[+0438s] # container exited on killed
[+0438s] # container exited on killed
[+0438s] # container exited on killed
```

edsantiago: Seems likely.

Please don't forget to address all the other failures.

edsantiago:
I see no fixes for all the other failures...?

flouthoc:
> I see no fixes for all the other failures...?

@edsantiago ackd, I'll check other failing tests now.

flouthoc:
> I see no fixes for all the other failures...?

I think this should be good now. :)

edsantiago:
LGTM and tests are green

flouthoc:
@containers/buildah-maintainers @nalind @rhatdan PTAL

TomSweeneyRedHat:
LGTM

flouthoc:
> New changes are detected. LGTM label has been removed.

Rebased cause merge bot was stuck.

PR #4191, opened 2022-08-17
Drop util/util.Cause()

nalind:
#### What type of PR is this?

/kind cleanup

Use errors.Is() and errors.As() instead of our own call-errors.Unwrap()-over-and-over-again helper.

None

```release-note
None
```

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -466,22 +466,3 @@ func VerifyTagName(imageSpec string) (types.ImageReference, error) {
 	}
 	return ref, nil
 }
-
-// Cause returns the most underlying error for the provided one. There is a
-// maximum error depth of 100 to avoid endless loops. An additional error log
-// message will be created if this maximum has reached.
-func Cause(err error) (cause error) {
-	cause = err
-
-	const maxDepth = 100
-	for i := 0; i <= maxDepth; i++ {
-		res := errors.Unwrap(cause)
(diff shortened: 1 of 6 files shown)


TomSweeneyRedHat:
LGTM

util/util.go
```diff
@@ -466,22 +466,3 @@ func VerifyTagName(imageSpec string) (types.ImageReference, error) {
 	}
 	return ref, nil
 }
-
-// Cause returns the most underlying error for the provided one. There is a
-// maximum error depth of 100 to avoid endless loops. An additional error log
-// message will be created if this maximum has reached.
-func Cause(err error) (cause error) {
```
flouthoc, line 473: LGTM just one doubt: Does removing `Cause` breaks public API ?

flouthoc (changes requested):
LGTM but one question above

util/util.go
nalind: Yes, it does.  I don't see it being used in podman, though.

flouthoc: Ah i see. SGTM.

flouthoc (approved):
LGTM

flouthoc:
Needs rebase afaics.

TomSweeneyRedHat:
LGTM

PR #4193, opened 2022-08-19
Cirrus: Lock cross-build task onto golang 1.18

cevich:
#### What type of PR is this?

/kind failing-test 

Recently (as of this commit) the OSX CI environment was updated.  This
is causing the cross_build task to break.  Fix this by locking onto the
golang version it was passing with on release.

#### How to verify it

The `Cross` task will pass

None

None

```release-note
None
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -126,10 +126,11 @@ cross_build_task:
 
     osx_instance:
         image: 'big-sur-base'
-
+    env:
+        PATH: "/usr/local/opt/go@1.18/bin:$PATH"
     script:
         - brew update
-        - brew install go
+        - brew install go@1.18
         - brew install go-md2man
         - brew install gpgme
         - go version


cevich:
@ashley-cui do you know [what this means](https://github.com/containers/buildah/pull/4193/checks?check_run_id=7921455698)?  I told it to do something new with golang, so it complains about ruby?!?!?!?

.cirrus.yml
```diff
@@ -129,7 +129,7 @@ cross_build_task:
```
nalind, line 129: Given the error message in the log for the smoke test, I'd suggest
```suggestion

    env:
        PATH: "/usr/local/opt/go@1.18/bin:$PATH"

```

ashley-cui:
I think it's a ruby warning, the command that failed was `go version`. I think that @nalind's suggestion would be the correct answer.

cevich:
Oh weird, okay I'll give it a try.  Thanks guys.

.cirrus.yml
```diff
@@ -129,7 +129,8 @@ cross_build_task:
 
     script:
         - brew update
-        - brew install go
+        - brew install go@1.18
+        - export PATH="/usr/local/opt/go@1.18/bin:$PATH"
```
nalind, line 133: The note in https://cirrus-ci.org/guide/writing-tasks/#script-instruction seems relevant here.

cevich: Oh right, good catch `Note: Each script instruction is executed in a newly created process, therefore environment variables are not preserved between them.`  I was afraid about including the (old)`$PATH` value at the end, since it's not set anywhere in the YAML.  Maybe Cirrus will pass the literal `$PATH` through so it will work?  Let's see...

.cirrus.yml
```diff
@@ -126,10 +126,11 @@ cross_build_task:
 
     osx_instance:
         image: 'big-sur-base'
-
+    env:
+        PATH: "/usr/local/opt/go@1.18/bin:$PATH"
     script:
         - brew update
-        - brew install go
+        - brew install go@1.18
```
TomSweeneyRedHat, line 133: I'm always hesitant to tag a particular version in a file like this.....  Can we just stick with latest?  I'm just concerned later down the line when we're looking at go@1.25 or what have you.

rhatdan: The problem with latest is that certain tools we rely on like lint, are not updated to use it and cause unfixable failures in CI/CD systems.

rhatdan: Since this is a branch, I see no reason not to freeze it at the golang that was used originally.

cevich: Yep, @TomSweeneyRedHat I would agree with you completely were this the 'main' branch.  For releases, we really want to lock them down to support future backports.  Otherwise every release-branch is a moving-target :scream:

PR #4194, opened 2022-08-20
test: use `T.TempDir` to create temporary test directory

Juneezee:
#### What type of PR is this?

/kind cleanup

A testing cleanup. 

This pull request replaces `ioutil.TempDir` with `t.TempDir`. We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete. 

This saves us at least 2 lines (error check, and cleanup) on every instance, or in some cases adds cleanup that we forgot.

Reference: https://pkg.go.dev/testing#T.TempDir

```go
func TestFoo(t *testing.T) {
	// before
	tmpDir, err := os.MkdirTemp("", "")
	if err != nil {
		t.Fatal(err)
	}
	defer os.RemoveAll(tmpDir)

	// now
	tmpDir := t.TempDir()
}
```

#### How to verify it

CI covers the test

```release-note
None
```

### DIFF
--- a/chroot/run_test.go
+++ b/chroot/run_test.go
@@ -8,7 +8,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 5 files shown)


vrothberg (commented):
Thanks for contributing, @Juneezee!

LGTM, @flouthoc @nalind PTAL

flouthoc (approved):
LGTM

Issue #4018, opened 2022-05-26
implement retry on buildah bud for FROM downloads


flucas1:
Network can be unreliable at some moments, even internal ones when STP layer-2 is acting...., and this make CI processes fail,

For this scenario I suggest that "buildah bud" has a retry capabilities when retrieving images via FROM,

Curl implement this group of options
curl --connect-timeout 5  --max-time 10  --retry 5  --retry-delay 0   --retry-max-time 40 ....

Apt also implement APT::Acquire::Retries

I would favor to have the retry delay and the number of retries as an overall parameter to buildah bud, that applies to all FROM instructions. FROM can be an argument to buildah, or having several FROM, so parsing the dockerfile to pre-download images is not an generic solution.

vrothberg:
Thanks for reaching out.  Buildah has a built-in retry logic and will do it at most 3 times.  Can you share the errors you are seeing and the version of Buildah you are using?

flouthoc:
Hi @flucas1 , Would it work for you if we make `MaxRetries` and `RetryDelay` configurable from CLI ? https://github.com/containers/buildah/blob/main/pull.go#L45

vrothberg:
That is a very unreliable network.  There is only so much (or little) tools can do in such environments.  What if the `RUN` instructions require network?

I am OK to expose the retry fields on the CLI but it feels much more like a symptomatic fix.

resmo:
This just came through my mind: What if `FROM` could be a list (delemiter separated, e.g. `|` ?) to try download from various endpoints sequencially? 

Many images are pushed to various container registries or "try first local registry and fallback to public registry if fails" scenario.

vrothberg:
Mirrors can already be configured in [man containers-registries.conf](https://github.com/containers/image/blob/main/docs/containers-registries.conf.5.md#example).  It is something I believe should be configured globally by the sysadmin and not by individual developers in Dockerfiles.

rhatdan:
Definitely not a new flag.  registries.conf would make sense to me or I could live with a CONTAINER_PULL_RETRIES environment variable.  It should affect all pulling of images though, not just FROM command.

vrothberg:
> Definitely not a new flag. registries.conf would make sense to me or I could live with a CONTAINER_PULL_RETRIES environment variable. It should affect all pulling of images though, not just FROM command.

[...]

PR #4195, opened 2022-08-22
buildah: add support for `--retry` and `--retry-delay` incase of `push/pull` failures.

flouthoc:
Allows users to configure `--retry` attempts and `--retry-delay`
duration using two additional flags for commands

* buildah build
* buildah pull
* buildah push
* buildah from
* buildah add
* buildah copy

Closes: https://github.com/containers/buildah/issues/4018

[NO NEW TESTS NEEDED]
[NO TESTS NEEDED]
Not sure how we can test retry attempts in CI, but added a test in
tests/bud.bats which verifies we parse and added flag in other options
for sanity parsing checking.

### DIFF
--- a/pkg/cli/build.go
+++ b/pkg/cli/build.go
@@ -310,6 +310,14 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
 			return options, nil, nil, fmt.Errorf("unable to parse value provided %q as --cache-ttl: %w", iopts.CacheTTL, err)
 		}
 	}
+	var pullPushRetryDelay time.Duration
+	pullPushRetryDelay, err = time.ParseDuration(iopts.RetryDelay)
+	if err != nil {
+		return options, nil, nil, fmt.Errorf("unable to parse value provided %q as --retry-delay: %w", iopts.RetryDelay, err)
+	}
+	// Following log line is used in integration test.
+	logrus.Debugf("Setting MaxPullPushRetries to %d and PullPushRetryDelay to %v", iopts.Retry, pullPushRetryDelay)
+
 	options = define.BuildOptions{
 		AddCapabilities:         iopts.CapAdd,
 		AdditionalBuildContexts: additionalBuildContext,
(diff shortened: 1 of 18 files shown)


vrothberg (commented):
Code LGTM but a test wouldn't hurt. At the very least to make sure the flags are parsed.

flouthoc:
> Code LGTM but a test wouldn't hurt. At the very least to make sure the flags are parsed.

Added a test in tests/bud.bats which verifies we parse and in other options just passing flag with valid values to verify top level parsing does not fails.

flouthoc:
@rhatdan Could you manually merge, cause rebase is not actually needed but gh would block. Otherwise i'll rebase and push again.

flouthoc:
Nevermind I rebased just to be sure.

flouthoc:
Master was outdated again.

pkg/cli/build.go
```diff
@@ -35,7 +35,7 @@ type BuildOptions struct {
 
 const (
 	MaxPullPushRetries = 3
-	PullPushRetryDelay = 2 * time.Second
+	PullPushRetryDelay = "2s"
```
nalind, line 38: Changing the type of this exported constant is an API break, and this one affects podman.

nalind:
The `buildah add`, `buildah copy`, `buildah from`, and `buildah pull` commands don't push images, so their man pages shouldn't imply that they might.  Likewise, `buildah push` doesn't pull images.  The commit log lists `buildah push` twice in the list of commands it affects.

pkg/cli/build.go
flouthoc: @nalind ah i missed it, I will introduce a new constant for str version of delay. Just curious aren't we going to remove the usage of old constant from podman on next vendor of buildah ( but anyways not an excuse for breaking public API so i'll amend this ).

nalind: Might be easier to use the value's `.String()` method as the default option value.

tests/push.bats
```diff
@@ -45,7 +45,7 @@ load helpers
   _prefetch alpine
   run_buildah from --quiet --pull=false $WITH_POLICY_JSON alpine
   cid=$output
-  run_buildah push $WITH_POLICY_JSON --format oci alpine dir:$mytmpdir
+  run_buildah push --retry 4 --retry-delay 4s $WITH_POLICY_JSON --format oci alpine dir:$mytmpdir
```
TomSweeneyRedHat, line 48: I don't think any of these tests that you have modified will normally retry 4 times.  I think they'll each succeed on the first pull. It would be nice to have at least one test somewhere that does retry 4 times every time and that the time to do so is at least 16 seconds.

flouthoc: This is only to verify if flags are treated as valid flags and buildah does not end up not parsing these flags, testing retry in CI is really hard since its hard to simulate registry failure. So these are just added for sanity and another detailed parsing test in `tests/bud.bats`.

flouthoc:

PR #4198, opened 2022-08-22
Makefile: always call $(GO) instead of `go`

nalind:
#### What type of PR is this?

/kind other

In a couple of places, we were still calling `go` directly instead of through the `$(GO)` variable, which was a mistake.  Drop one of them entirely because the resulting value wasn't being used anywhere.

Update the definitions of CNI_COMMIT and RUNC_COMMIT to search `vendor/modules.txt` instead of `go.mod`, which can be overridden and which might therefore produce incorrect values.

#### How to verify it

Use `make GO=/path/to/other/version/of/go` to override how the Makefile invokes the Go compiler.

None

```release-note
None
```


TomSweeneyRedHat:
LGTM

flouthoc (approved):
LGTM
/lgtm
/approve

PR #4200, opened 2022-08-24
run: add container `gid` to additional groups

flouthoc:
When container is created with specific `uid` and `gid` also add container
`gid` to supplementary/additional group.

### DIFF
--- a/run_common.go
+++ b/run_common.go
@@ -262,6 +262,7 @@ func (b *Builder) configureUIDGID(g *generate.Generator, mountPoint string, opti
 	}
 	g.SetProcessUID(user.UID)
 	g.SetProcessGID(user.GID)
+	g.AddProcessAdditionalGid(user.GID)
 	for _, gid := range user.AdditionalGids {
 		g.AddProcessAdditionalGid(gid)
 	}
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -366,6 +366,22 @@ _EOF
   expect_output --substring "invalid response status"
 }
 
+@test "build test has gid in supplemental groups" {
+  _prefetch alpine
+  run_buildah build $WITH_POLICY_JSON -t source -f $BUDFILES/supplemental-groups/Dockerfile
+  # gid 1000 must be in supplemental groups
+  expect_output --substring "Groups:	1000"
+}
+
+@test "build test if supplemental groups has gid with --isolation chroot" {
+  test -z "${BUILDAH_ISOLATION}" || skip "BUILDAH_ISOLATION=${BUILDAH_ISOLATION} overrides --isolation"
+
+  _prefetch alpine
+  run_buildah build --isolation chroot $WITH_POLICY_JSON -t source -f $BUDFILES/supplemental-groups/Dockerfile
+  # gid 1000 must be in supplemental groups
+  expect_output --substring "Groups:	1000"
+}
+
 # Test skipping images with FROM
 @test "build-test skipping unwanted stages with FROM" {
   mkdir -p ${TEST_SCRATCH_DIR}/bud/platform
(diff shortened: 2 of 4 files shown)


flouthoc:
Needs green flag from @nalind and @giuseppe.

giuseppe (approved):
LGTM

flouthoc:
Added additional test for `--isolation chroot`.

tests/bud.bats
```diff
@@ -366,6 +366,26 @@ _EOF
...
 
+@test "build test has gid in supplemental groups" {
+  _prefetch alpine
+  mkdir -p ${TEST_SCRATCH_DIR}/bud/platform
+  run_buildah build $WITH_POLICY_JSON -t source -f $BUDFILES/supplemental-groups/Dockerfile
+  # gid 1000 must be in supplemental groups
+  expect_output --substring "Groups:	1000"
+}
+
+@test "build test has gid in supplemental groups and --isolation chroot" {
```
nalind, line 377: Non-blocker: try to avoid giving tests names that are exact substrings of the names of other tests, which makes using `bats -f` to run specific tests harder.

tests/bud.bats
```diff
@@ -366,6 +366,26 @@ _EOF
...
+  skip_if_rootless_environment
+  skip_if_no_runtime
+  test -z "${BUILDAH_ISOLATION}" || skip "BUILDAH_ISOLATION=${BUILDAH_ISOLATION} overrides --isolation"
+
+  _prefetch alpine
+  mkdir -p ${TEST_SCRATCH_DIR}/bud/platform
```
nalind, line 383: What's this directory used for?

tests/bud.bats
flouthoc: Changed the test name from first one.

tests/bud.bats
flouthoc: Nothing, I had a inline file in first draft but later removed it so this is not needed.

flouthoc:
@nalind PTAL again.

tests/bud.bats
```diff
@@ -366,6 +366,24 @@ _EOF
...
+@test "build test has gid in supplemental groups" {
+  _prefetch alpine
+  run_buildah build $WITH_POLICY_JSON -t source -f $BUDFILES/supplemental-groups/Dockerfile
+  # gid 1000 must be in supplemental groups
+  expect_output --substring "Groups:	1000"
+}
+
+@test "build test if supplemental groups has gid with --isolation chroot" {
+  skip_if_rootless_environment
+  skip_if_no_runtime
```
nalind, line 378: Are either of these conditions required for this test?

rhatdan:
LGTM

tests/bud.bats
flouthoc: Other tests which have `--isolation chroot` are being skipped for similar scenarios that's why i added it, I can try removing it.

rhatdan: I think we should probably try to remove them from all of the chroot tests.

flouthoc: Removed from this test, lets see if CI is happy. I'll do this for all the other tests in a different PR since they are not part of this PR.

run_common.go
```diff
@@ -262,6 +262,7 @@ func (b *Builder) configureUIDGID(g *generate.Generator, mountPoint string, opti
 	}
 	g.SetProcessUID(user.UID)
 	g.SetProcessGID(user.GID)
+	g.AddProcessAdditionalGid(user.GID)
```
TomSweeneyRedHat, line 265: NIce catch @flouthoc .  I could have sworn I looked for SetProcessGID, but didn't find it....

TomSweeneyRedHat:
@flouthoc once a BZ gets spun for this, please add a comment with the BZ number if I neglect to.

nalind:
Tracked in Red Hat bugzilla at https://bugzilla.redhat.com/show_bug.cgi?id=2121453

TomSweeneyRedHat:
I think we need to get this into Buildah release-1.27 too so we can spin this for 8.7 zero day.

flouthoc:
Backport PR here: https://github.com/containers/buildah/pull/4207

PR #4205, opened 2022-08-27
Fix stutters

rhatdan:
Podman adds an Error: to every error message. So starting an error
message with "error" ends up being reported to the user as

Error: error ...

This patch removes the stutter.

Also ioutil.ReadFile errors report the Path, so wrapping the err message
with the path causes a stutter.

Signed-off-by: Daniel J Walsh dwalsh@redhat.com

[NO NEW TESTS NEEDED]

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/add.go
+++ b/add.go
@@ -105,7 +105,7 @@ func getURL(src string, chown *idtools.IDPair, mountpoint, renameTarget string, 
 	if lastModified != "" {
 		d, err := time.Parse(time.RFC1123, lastModified)
 		if err != nil {
-			return fmt.Errorf("error parsing last-modified time: %w", err)
+			return fmt.Errorf("parsing last-modified time: %w", err)
 		}
 		date = d
 	}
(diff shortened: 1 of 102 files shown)


flouthoc (commented):
It looks likes some tests needs to retrofit as per the new error messages.

nalind:
Does this mean we stop mentioning that errors are errors when people run the `buildah` command?

rhatdan:
@nalind I think most errors are clearly errors when they happen, but we can do the same as Podman and prepend Errors: if you think we should.

@flouthoc @TomSweeneyRedHat @nalind @vrothberg @giuseppe PTAL

flouthoc (approved):
Changes LGTM waiting for @nalind to confirm on his concern.

giuseppe (approved):
LGTM

nalind:
> @nalind I think most errors are clearly errors when they happen, but we can do the same as Podman and prepend Errors: if you think we should.

The stutter you're trying to fix here only exists because podman does that, so yeah, I'd say we should.  To put a fine point on it, if fixing podman's stutter means making buildah's error messages less clear, I'm going to nack.

Issue #4217, opened 2022-09-02
Add `podman manifest exists` command to buildah, too


der-eismann:
**Description**

`podman manifest` and `buildah manifest` provide almost the same sub-commands, only `buildah manifest exists` is missing. But it should have the command as well. 

**Describe the results you received:**
Buildah is missing the `manifest exists` sub-command

**Describe the results you expected:**
Buildah should have the `manifest exists` sub-command

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.26.4-2.fc36.x86_64
```

**Output of `buildah version`:**

```
Version:         1.26.4
Go Version:      go1.18.4
Image Spec:      1.0.2-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        1.0.0
libcni Version:  v1.1.2
image Version:   5.22.0
Git Commit:      
Built:           Mon Aug  8 16:11:10 2022
OS/Arch:         linux/amd64
BuildPlatform:   linux/amd64
```

**Output of `podman version` if reporting a `podman build` issue:**

```
Client:       Podman Engine
Version:      4.2.0
API Version:  4.2.0
Go Version:   go1.18.4
Built:        Thu Aug 11 16:42:17 2022
OS/Arch:      linux/amd64
```

**Output of `cat /etc/*release`:**

```
Fedora release 36 (Thirty Six)
NAME="Fedora Linux"
VERSION="36 (Workstation Edition)"
ID=fedora
VERSION_ID=36
VERSION_CODENAME=""
PLATFORM_ID="platform:f36"
PRETTY_NAME="Fedora Linux 36 (Workstation Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:36"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f36/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=36
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=36
[...]

PR #4220, opened 2022-09-05
manifest: add support for `buildah manifest exists`

flouthoc:
Adds support for `buildah manifest exists <name>` which tells user if
requested manifest is present in local storage or not, if manifest is
present in local-storage command exits with exit code 0 otherwise 1.

Similar to: https://docs.podman.io/en/latest/markdown/podman-manifest-exists.1.html

Closes: https://github.com/containers/buildah/issues/4217

```release-note
manifest: add support for buildah manifest exists
```

### DIFF
--- a/cmd/buildah/manifest.go
+++ b/cmd/buildah/manifest.go
@@ -55,6 +55,7 @@ func init() {
 		manifestInspectDescription  = "\n  Display the contents of a manifest list or image index."
 		manifestPushDescription     = "\n  Pushes manifest lists and image indexes to registries."
 		manifestRmDescription       = "\n  Remove one or more manifest lists from local storage."
+		manifestExistsDescription   = "\n  Check if a manifest list exists in local storage."
 		manifestCreateOpts          manifestCreateOpts
 		manifestAddOpts             manifestAddOpts
 		manifestRemoveOpts          manifestRemoveOpts
@@ -155,6 +156,19 @@ func init() {
 	manifestRemoveCommand.SetUsageTemplate(UsageTemplate())
 	manifestCommand.AddCommand(manifestRemoveCommand)
 
+	manifestExistsCommand := &cobra.Command{
+		Use:   "exists",
+		Short: "Check if a manifest list exists in local storage",
+		Long:  manifestExistsDescription,
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return manifestExistsCmd(cmd, args)
+		},
+		Example: "buildah manifest exists mylist",
+	}
+	manifestExistsCommand.SetUsageTemplate(UsageTemplate())
+	manifestCommand.AddCommand(manifestExistsCommand)
+
 	manifestAnnotateCommand := &cobra.Command{
 		Use:   "annotate",
 		Short: "Add or update information about an entry in a manifest list or image index",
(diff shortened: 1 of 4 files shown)


vrothberg (approved):
LGTM
@nalind PTAL

cmd/buildah/manifest.go
```diff
@@ -55,6 +55,7 @@ func init() {
 		manifestInspectDescription  = "\n  Display the contents of a manifest list or image index."
 		manifestPushDescription     = "\n  Pushes manifest lists and image indexes to registries."
 		manifestRmDescription       = "\n  Remove one or more manifest lists from local storage."
+		manifestExistsDescription   = "\n  If the manifest list exists in local storage, buildah manifest exists exits with 0, otherwise the exit code will be 1."
```
nalind, line 58: Leave off the bit about exit status.  This should basically be the same as the `Short` description of the command, give or take punctuation and formatting.

cmd/buildah/manifest.go
```diff
@@ -155,6 +156,19 @@ func init() {
 	manifestRemoveCommand.SetUsageTemplate(UsageTemplate())
 	manifestCommand.AddCommand(manifestRemoveCommand)
 
+	manifestExistsCommand := &cobra.Command{
+		Use:   "exists MANIFEST",
```
nalind, line 160: Other `Use` fields for `manifest` commands don't list arguments here, so this shouldn't.

cmd/buildah/manifest.go
```diff
@@ -237,6 +251,36 @@ func init() {
...
+	}
+	runtime, err := libimage.RuntimeFromStore(store, &libimage.RuntimeOptions{SystemContext: systemContext})
+	if err != nil {
+		return err
+	}
+
+	_, err = runtime.LookupManifestList(name)
+	if err != nil {
+		if errors.Is(err, storage.ErrImageUnknown) {
+			os.Exit(1)
```
nalind, line 277: This bypasses the store shutdown logic, which has to be run if we initialized a store, if only to undo the private bind mount of the storage root over itself.  I'm surprised this doesn't break in CI, which tends to have trouble removing that directory when that happens.

docs/buildah-manifest-exists.1.md
```diff
@@ -0,0 +1,40 @@
+% buildah-manifest-exists(1)
+
+## NAME
+buildah\-manifest\-exists - Check if the given manifest list exists in local storage
+
+## SYNOPSIS
+**buildah manifest exists** *manifest*
+
+## DESCRIPTION
+**buildah manifest exists** checks if a manifest list exists on local storage. Podman will
```
nalind, line 10:
```suggestion
**buildah manifest exists** checks if a manifest list exists in local storage. Buildah will
```

cmd/buildah/manifest.go
flouthoc: Added a call to `shutdownStore` just before we exit manually, will this address the mentioned issue.

tests/lists.bats
```diff
@@ -29,6 +33,10 @@ IMAGE_LIST_S390X_INSTANCE_DIGEST=sha256:882a20ee0df7399a445285361d38b711c299ca09
 @test "manifest-add" {
     run_buildah manifest create foo
     run_buildah manifest add foo ${IMAGE_LIST}
+    # since manifest exists in local storage this should exit with `0`
+    run_buildah manifest exists foo
+    # since manifest doest not exists in local storage this should exit with `1`
```
nalind, line 38:
```suggestion
    # since manifest does not exist in local storage this should exit with `1`
```

tests/lists.bats
```diff
@@ -211,4 +219,8 @@ IMAGE_LIST_S390X_INSTANCE_DIGEST=sha256:882a20ee0df7399a445285361d38b711c299ca09
     run_buildah manifest inspect test
     # must contain amd64
     expect_output --substring "amd64"
+    # since manifest exists in local storage this should exit with `0`
+    run_buildah manifest exists test:latest
+    # since manifest doest not exists in local storage this should exit with `1`
```
nalind, line 224:
```suggestion
    # since manifest doest not exist in local storage this should exit with `1`
```

cmd/buildah/manifest.go
nalind: I'd slightly prefer a way to have the final cleanup logic in `main()` handle this, but this should do.  I'm still wary that CI didn't catch this - the tests with the overlay driver would have done so in the past.

nalind:
A couple of comment nits, otherwise LGTM.

Issue #4227, opened 2022-09-10
--network=none does nothing


SpecLad:
**Description**

The `--network=none` flag seems to have no effect when used with `buildah build`.

**Steps to reproduce the issue:**
1. Create a `test.containerfile` file with the following contents:

```containerfile
FROM docker.io/curlimages/curl

RUN curl -I https://www.yahoo.com/
```

2. Run `buildah build --network=none -f test.containerfile`.

**Describe the results you received:**
The `curl` invocation succeeds.

**Describe the results you expected:**
The `curl` invocation (and the build along with it) should fail.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah/jammy,now 1.23.1+ds1-2 amd64 [installed]
```

**Output of `buildah version`:**

```
Version:         1.23.1
Go Version:      go1.17
Image Spec:      1.0.1
Runtime Spec:    1.0.2-dev
CNI Spec:        0.4.0
libcni Version:
image Version:   5.16.0
Git Commit:
Built:           Thu Jan  1 03:00:00 1970
OS/Arch:         linux/amd64
BuildPlatform:   linux/amd64
```

This version is pretty old, however, I can still reproduce the issue with the latest Git commit (9e2b0e2f).

**Output of `podman version` if reporting a `podman build` issue:** N/A

**Output of `cat /etc/*release`:**

```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.1 LTS"
PRETTY_NAME="Ubuntu 22.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.1 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

**Output of `uname -a`:**

```
Linux Freezie 5.10.102.1-microsoft-standard-WSL2 #1 SMP Wed Mar 2 00:30:59 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

[...]

PR #4229, opened 2022-09-11
buildah bud --network=none should have no network

rhatdan:
Fixes: https://github.com/containers/buildah/issues/4227

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -602,6 +602,7 @@ func (s *StageExecutor) Run(run imagebuilder.Run, config docker.Config) error {
 		Args:             s.executor.runtimeArgs,
 		Cmd:              config.Cmd,
 		ContextDir:       s.executor.contextDir,
+		ConfigureNetwork: s.executor.configureNetwork,
 		Entrypoint:       config.Entrypoint,
 		Env:              config.Env,
 		Hostname:         config.Hostname,
@@ -624,10 +625,9 @@ func (s *StageExecutor) Run(run imagebuilder.Run, config docker.Config) error {
 		User:             config.User,
 		WorkingDir:       config.WorkingDir,
 	}
+
 	if config.NetworkDisabled {
 		options.ConfigureNetwork = buildah.NetworkDisabled
-	} else {
-		options.ConfigureNetwork = buildah.NetworkEnabled
 	}
 
 	args := run.Args
--- a/run_linux.go
+++ b/run_linux.go
@@ -726,7 +726,9 @@ func (b *Builder) configureNamespaces(g *generate.Generator, options *RunOptions
 			options.ConfigureNetwork = networkPolicy
 		}
 	}
-
+	if networkPolicy == NetworkDisabled {
+		namespaceOptions.AddOrReplace(define.NamespaceOptions{{Name: string(specs.NetworkNamespace), Host: false}}...)
+	}
 	configureNetwork, configureNetworks, configureUTS, err := setupNamespaces(options.Logger, g, namespaceOptions, b.IDMappingOptions, networkPolicy)
 	if err != nil {
 		return false, nil, err
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -5308,3 +5308,16 @@ _EOF
   run_buildah 125 build -f ${TEST_SCRATCH_DIR}/Dockerfile1 -f ${TEST_SCRATCH_DIR}/Dockerfile2 ${TEST_SCRATCH_DIR}
   assert "$output" =~ "error parsing additional Dockerfile .*Dockerfile2: invalid ESCAPE"
 }
+
+@test "build-with-network-test" {
+  skip_if_in_container # Test only works in OCI isolation, which doesn't work in CI/CD systems. Buildah defaults to chroot isolation
+  _prefetch alpine
+  cat > ${TEST_SCRATCH_DIR}/Containerfile << _EOF
+FROM alpine
+RUN ping -c 1 4.2.2.2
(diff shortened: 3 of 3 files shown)


flouthoc (approved):
LGTM

flouthoc:
@containers/buildah-maintainers @nalind @giuseppe @vrothberg @Luap99 PTAL

TomSweeneyRedHat:
nice catch @Luap99 !  And a great example of why @edsantiago reminds us frequently to not only test the return status, but also the value we get back from a call.

TomSweeneyRedHat:
and it looks like the new tests isn't failing as expected.  No failure is being raised.

```
[+1730s] not ok 373 build-with-network-test
[+1730s] # (from function `die' in file ./helpers.bash, line 314,
[+1730s] #  from function `run_buildah' in file ./helpers.bash, line 301,
[+1730s] #  in test file ./bud.bats, line 5236)
.
.
.
[+1730s] # $ /var/tmp/go/src/github.com/containers/buildah/tests/./../bin/buildah build --network=none --signature-policy /var/tmp/go/src/github.com/containers/buildah/tests/./policy.json /var/tmp/buildah_tests.tqdeml
[+1730s] # STEP 1/2: FROM alpine
[+1730s] # STEP 2/2: RUN ping -c 1 4.2.2.2
[+1730s] # PING 4.2.2.2 (4.2.2.2): 56 data bytes
[+1730s] # 64 bytes from 4.2.2.2: seq=0 ttl=61 time=21.023 ms
[+1730s] #
[+1730s] # --- 4.2.2.2 ping statistics ---
[+1730s] # 1 packets transmitted, 1 packets received, 0% packet loss
[+1730s] # round-trip min/avg/max = 21.023/21.023/21.023 ms
[+1730s] # COMMIT
[+1730s] # Getting image source signatures
[+1730s] # Copying blob sha256:03901b4a2ea88eeaad62dbe59b072b28b6efa00491962b8741081c5df50c65e0
[+1730s] # Copying blob sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef
[+1730s] # Copying config sha256:7bff38e9d8e465107f3b52dbb29ab6619a00e75e50c998cd17a35618c984a791
[+1730s] # Writing manifest to image destination
[+1730s] # Storing signatures
[+1730s] # --> 7bff38e9d8e
[+1730s] # 7bff38e9d8e465107f3b52dbb29ab6619a00e75e50c998cd17a35618c984a791
[+1730s] # #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
[+1730s] # #| FAIL: exit code is 0; expected 1
[+1730s] # #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

rhatdan:
@Luap99 The other tests is on the success path.  I plan a follow update to the newly added tests. but it is much more involved in that it requires changes to imagebuilder.

rhatdan:
@TomSweeneyRedHat the prior tests was not attempting to do networking, so it is successful.

flouthoc:
> There is already a network=none test,
> 
> https://github.com/containers/buildah/blob/f0b2d39b83cd3bae8a6d0ef8c05fd384f81412a0/tests/bud.bats#L3832
> 
> Please fix the test instead of adding more.

@Luap99 upon checking this, the suggested test is just checking sanity of option of not actually testing the network, I will refine all tests for network with implementation of https://github.com/containers/buildah/issues/4230 in a different PR

flouthoc:
Restarted flaking test.

flouthoc:
I am not sure if this containerized integration is running on `cgroupv1` or `cgroupv2` but the mount `/sys/fs/cgroup` must be writable for `--isolation oci` to work and I don't think this will be allowed on `cgroupv1` since delegation does not works on `cgroupv1` and it must be `ro` inside the container.

But I was able to make similar thing work with with `cgroupv2` on my local with extra `--privileged`, it looks like `--security-opt=label=disable --security-opt=seccomp=unconfined --cap-add=all` is not as equal to `--privileged` so I think this may need following diff but i have only tested on `v2`.

```diff
diff --git a/contrib/cirrus/lib.sh b/contrib/cirrus/lib.sh
index 31278ec8..7d1f61fa 100755
--- a/contrib/cirrus/lib.sh
+++ b/contrib/cirrus/lib.sh
@@ -184,6 +184,7 @@ in_podman() {
                    --security-opt=label=disable \
                    --security-opt=seccomp=unconfined \
                    --cap-add=all \
+                   --privileged \
                    --env-file=$envfile \
                    -e "GOPATH=$GOPATH" \
                    -e "GOSRC=$GOSRC" \
```

@rhatdan overall i see lot of direct cgroup tests are being skipped when running inside container so not sure if `--isolation oci` will work on in this CI maybe as last option adding `skip_if_in_container` to skip this test in container could be an option.

rhatdan:
The issue which as failing, was that buildah automatically runs within a container with --isolation=chroot and --network option has no effect in this situation.  I opened an issue with buildah to throw an error when running --isolation=chroot and --network *

https://github.com/containers/buildah/pull/4250

flouthoc (approved):
LGTM
/lgtm
/approve

flouthoc:
This might need a rebase or please force whoever sees this message first since i don't have access to force merge.

PR #4234, opened 2022-09-12
parse: make processing fields for `--mount` order agnostic

flouthoc:
Field processing in `--mount` must not be hardcode to expect first field to be `type` instead it should be order agnostic.

Closes: https://github.com/containers/podman/issues/15748

```release-note
run,build: now --mount processes flags in no fixed orger
```

### DIFF
--- a/internal/parse/parse.go
+++ b/internal/parse/parse.go
@@ -59,6 +59,9 @@ func GetBindMount(ctx *types.SystemContext, args []string, contextDir string, st
 	for _, val := range args {
 		kv := strings.SplitN(val, "=", 2)
 		switch kv[0] {
+		case "type":
+			// This is already processed
+			continue
 		case "bind-nonrecursive":
 			newMount.Options = append(newMount.Options, "bind")
 			bindNonRecursive = true
@@ -209,6 +212,9 @@ func GetCacheMount(args []string, store storage.Store, imageMountLabel string, a
 	for _, val := range args {
 		kv := strings.SplitN(val, "=", 2)
 		switch kv[0] {
+		case "type":
+			// This is already processed
+			continue
 		case "nosuid", "nodev", "noexec":
 			// TODO: detect duplication of these options.
 			// (Is this necessary?)
@@ -497,6 +503,8 @@ func GetVolumes(ctx *types.SystemContext, store storage.Store, volumes []string,
 // buildah run --mount type=bind,src=/etc/resolv.conf,target=/etc/resolv.conf ...
 // buildah run --mount type=tmpfs,target=/dev/shm ...
 func getMounts(ctx *types.SystemContext, store storage.Store, mounts []string, contextDir string) (map[string]specs.Mount, []string, []string, error) {
+	// If `type` is not set default to "bind"
+	mountType := TypeBind
 	finalMounts := make(map[string]specs.Mount)
 	mountedImages := make([]string, 0)
 	lockedTargets := make([]string, 0)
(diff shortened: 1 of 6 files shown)


internal/parse/parse.go
```diff
@@ -511,13 +517,18 @@ func getMounts(ctx *types.SystemContext, store storage.Store, mounts []string, c
 		if len(arr) < 2 {
 			return nil, mountedImages, lockedTargets, fmt.Errorf("%q: %w", mount, errInvalidSyntax)
 		}
-		kv := strings.Split(arr[0], "=")
+		typeToken := arr[0]
+		for _, field := range arr {
```
nalind, line 521: `arr` has at most 2 elements here, so this loop doesn't iterate through all supplied arguments.

internal/parse/parse.go
```diff
@@ -511,13 +517,18 @@ func getMounts(ctx *types.SystemContext, store storage.Store, mounts []string, c
...
+			if strings.HasPrefix(field, "type=") {
+				typeToken = field
+			}
+		}
+		kv := strings.Split(typeToken, "=")
 		// TODO: type is not explicitly required in Docker.
 		// If not specified, it defaults to "volume".
```
nalind, line 528: The default is "bind".

internal/parse/parse.go
flouthoc: I have changed `SplitN` with `Split`.

rhatdan:
LGTM
@nalind PTAL again.

flouthoc:
I think tests should go green on this one.

flouthoc:
@nalind PTAL again.

internal/parse/parse.go
```diff
@@ -507,18 +513,23 @@ func getMounts(ctx *types.SystemContext, store storage.Store, mounts []string, c
...
+			if strings.HasPrefix(field, "type=") {
+				typeToken = field
+			}
+		}
+		kv := strings.Split(typeToken, "=")
 		// TODO: type is not explicitly required in Docker.
-		// If not specified, it defaults to "volume".
+		// If not specified, it defaults to "bind".
 		if len(kv) != 2 || kv[0] != "type" {
 			return nil, mountedImages, lockedTargets, fmt.Errorf("%q: %w", mount, errInvalidSyntax)
```
nalind, line 530: It looks like this might complain if the mount specification doesn't include a "type" at all.  Please ensure there's a test that uses neither "from" nor "type".

internal/parse/parse.go
```diff
@@ -507,19 +513,30 @@ func getMounts(ctx *types.SystemContext, store storage.Store, mounts []string, c
...
+				foundType = true
+				typeToken = field
+			}
 		}
-
-		tokens := strings.Split(arr[1], ",")
-		switch kv[1] {
+		kv := strings.Split(typeToken, "=")
+		// If `type` is not set default to "bind"
+		mountType := TypeBind
```
nalind, line 530: Maybe move this line to before the loop and 522 and just pull the value out of it in the loop if the argument has a "type=" prefix?  It'd probably be shorter, and `foundToken` isn't being used for much else.

internal/parse/parse.go
```diff
@@ -507,19 +513,30 @@ func getMounts(ctx *types.SystemContext, store storage.Store, mounts []string, c
...
+		if foundType {
+			kv := strings.Split(typeToken, "=")
+			if len(kv) != 2 || kv[0] != "type" {
+				return nil, mountedImages, lockedTargets, fmt.Errorf("%q: %w", mount, errInvalidSyntax)
+			}
+			mountType = kv[1]
+		}
+		tokens := strings.Split(mount, ",")
```
nalind, line 538: `tokens` and `arr` are always the same value.

tests/bud.bats
```diff
@@ -4956,6 +4984,15 @@ _EOF
   run_buildah rmi -f testbud
 }
 
+@test "bud-with-mount-with-only-target-like-buildkit" {
+  skip_if_no_runtime
+  skip_if_in_container
+  cp -R $BUDFILES/buildkit-mount ${TEST_SCRATCH_DIR}/buildkit-mount
+  run_buildah build -t testbud $WITH_POLICY_JSON -f ${TEST_SCRATCH_DIR}/buildkit-mount/Dockerfile6 ${TEST_SCRATCH_DIR}/buildkit-mount/
+  expect_output --substring "hello"
+  run_buildah rmi -f testbud
```
nalind, line 4993: This isn't necessary, since the storage is cleaned up after the test.

flouthoc:

PR #4246, opened 2022-09-14
Update vendor of containers/storage

rhatdan:

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```


rhatdan:
@giuseppe @vrothberg @nalind @flouthoc 
Vendoring of containers/storage is causing errors in Containerized integration.  Any ideas?

nalind:
In both cases the error ends up including
```
Error loading container config when searching for local runtime
```
which as near as I can tell, could only be triggered if `github.com/containers/storage/pkg/unshare.IsRootless()` returned `true` for some reason.

To the best of my understanding, the container in which these tests are running both has `CAP_SYS_ADMIN` (the container is being started with `--cap-add=all`) and the default host ID mappings (one of the tests `cat`s the `/proc/self/uid_map` from inside of a chroot environment, and the log shows the default mappings, and they can only be inherited from a user namespace with the default mappings), so I must be missing something.

rhatdan:
The biggest change is 975fab1023f74c7b9787323deca2250a82469e98 from @giuseppe

flouthoc:
CI goes green by reverting some commits , I tested CI with https://github.com/flouthoc/storage/tree/test-fix-2 and it went green and it included revert of `975fab1023f74c7b9787323deca2250a82469e98` as well but I am trying to confirm exact commit with another push.

flouthoc:
Its not only `975fab1023f74c7b9787323deca2250a82469e98` but CI goes green when i revert three commits https://github.com/containers/storage/compare/main...flouthoc:storage:test-fix-2

flouthoc:
Needs https://github.com/containers/common/pull/1157

Issue #4244, opened 2022-09-14
allow `--cache-ttl=0` to be equivalent to `--no-cache`


dustymabe:
`--cache-ttl` was added in https://github.com/containers/buildah/pull/4165. Currently `--cache-ttl=0` is equivalent to having not specified `--cache-ttl` at all. It would be nice if the option accepted a value of `0` and allowed it to effectively mean `--no-cache`.

This would allow programs that pass through values to `podman` or `buildah` to use one pass through argument rather than two. i.e. something like https://github.com/coreos/coreos-assembler/blob/5ffbf12d229854a79c0b1e53ce2bc4ffdeda5c88/src/cmd-remote-build-container#L194-L197

rhatdan:
@flouthoc PTAL

flouthoc:
@dustymabe Above PR https://github.com/containers/buildah/pull/4248 should close this.

nalind:
Did `--cache-ttl=0` still populate the cache, even if its contents weren't used?

flouthoc:
@nalind I think `--no-cache` also populates the cache without using it. We had a different opened card somewhere to introduce new API split populating and using cache.

nalind:
Well the cache is still just the images we build along the way.

PR #4248, opened 2022-09-15
build: make `--cache-ttl=0` equivalent to `--no-cache`

flouthoc:
If user explicitly specified `--cache-ttl=0` or duration
which equals to 0 it would effectively mean that user 
is asking to use no cache at all. In such use cases
buildah can skip looking for cache entierly by setting
`--no-cache=true` internally.

Closes: https://github.com/containers/buildah/issues/4244

```release-note
buildah build: make `--cache-ttl=0` equivalent to `--no-cache`
```

### DIFF
--- a/pkg/cli/build.go
+++ b/pkg/cli/build.go
@@ -309,6 +309,18 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
 		if err != nil {
 			return options, nil, nil, fmt.Errorf("unable to parse value provided %q as --cache-ttl: %w", iopts.CacheTTL, err)
 		}
+		// If user explicitly specified `--cache-ttl=0s`
+		// it would effectively mean that user is asking
+		// to use no cache at all. In such use cases
+		// buildah can skip looking for cache entierly
+		// by setting `--no-cache=true` internally.
+		if int64(cacheTTL) == 0 {
+			logrus.Debug("Setting --no-cache=true since --cache-ttl was set to 0s which effectively means user wants to ignore cache")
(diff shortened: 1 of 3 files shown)


pkg/cli/build.go
```diff
@@ -309,6 +309,15 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
...
 			return options, nil, nil, fmt.Errorf("unable to parse value provided %q as --cache-ttl: %w", iopts.CacheTTL, err)
 		}
+		// If user explicitly specified `--cache-ttl=0s`
+		// it would effectively mean that user is asking
+		// to use no cache at all. In such use cases
+		// buildah can skip looking for cache entierly
+		// by setting `--no-cache=true` internally.
+		if int64(cacheTTL) == 0 {
+			logrus.Debug("Setting --no-cache=true since --cache-ttl was set to 0s which effectively means user wants to ignore cache")
+			iopts.NoCache = true
```
rhatdan, line 319: for completeness, you should throw an error if user specified

==nocache=false
if iopts.NoChache==false &&  c.Flag("no-cache").Changed  {
            return fmt.Errorf(...)
}

flouthoc: Yes I agree

rhatdan:
LGTM
Although tests are failing.

flouthoc:
Failing tests are cirrus flakes afaics, i have restarted them.

rhatdan:
@nalind @giuseppe @vrothberg @umohnani8 @ashley-cui PTAL

umohnani8:
LGTM

PR #4249, opened 2022-09-15
build: support `--skip-unused-stages` for multi-stage builds.

flouthoc:
In multi-stage builds buildah will skip stages which are unused (i.e stages which don't contribute anything to target stage directly or indirectly) however in certain cases users need to process these unused stages hence add support for `--skip-unused-stages` which allows users to control this behavior. Defaults to `true`.

Ref: https://github.com/GoogleContainerTools/kaniko#flag---skip-unused-stages

Closes: https://github.com/containers/buildah/issues/4243

```release-note
build: support `--skip-unused-stages` while allows users to toggle skipping unused stages
```

PR #4256, opened 2022-09-19
Fix broken command completion

edsantiago:
PR #4249 added --skip-unused-stages (cool) but also added what seems to be an unnecessary command-completion incantation. This incantation breaks podman unit tests when vendored in the treadmill.

```release-note
None
```


flouthoc (approved):
LGTM

I added this after CI started failing but I guess it was not needed and I was wrong about this.
@edsantiago Thanks for fixing this.

PR #4259, opened 2022-09-19
[release-1.20] post-branch fixups

nalind:
Creating an older branch means catching up to things we've done on other older branches: switching which VM images we use, which tests we run (mostly dropping some) and changing references to "main" to refer to this branch, either by name or commit ID.


flouthoc (approved):
LGTM
/hold

PR #4207, opened 2022-08-30
[release-v1.27] run: add container `gid` to additional groups

flouthoc:
When container is created with specific uid and gid also add container
gid to supplementary/additional group.

Backport of: https://github.com/containers/buildah/pull/4200

PR #4262, opened 2022-09-19
[release-1.21] run: add container gid to additional groups

nalind:
When container is created with specific uid and gid also add container gid to supplementary/additional group.

Cherry-picked from #4207, related to GHSA-rc4r-wh2q-q6c4.


flouthoc (approved):
LGTM
/lgtm
/hold cancel

PR #4267, opened 2022-09-19
pkg/cli: improve completion test

Luap99:
#### What type of PR is this?

 /kind bug

When we have a bool flag there is zero reason to define a completion function since this flag takes no argument and the cobra lib will thus not use it.

Podman checks that already and recently buildah caused the test to fail. [1] To make sure this does not happen again we should have the same check here.

[1] https://github.com/containers/podman/pull/13808#issuecomment-1250318199

```release-note
None
```

### DIFF
--- a/pkg/cli/common_test.go
+++ b/pkg/cli/common_test.go
@@ -10,12 +10,18 @@ import (
 func testFlagCompletion(t *testing.T, flags pflag.FlagSet, flagCompletions completion.FlagCompletions) {
 	// lookup if for each flag a flag completion function exists
 	flags.VisitAll(func(f *pflag.Flag) {
-		// skip hidden, deprecated and boolean flags
-		if f.Hidden || len(f.Deprecated) > 0 || f.Value.Type() == "bool" {
+		// skip hidden and deprecated flags
+		if f.Hidden || len(f.Deprecated) > 0 {
 			return
 		}
-		if _, ok := flagCompletions[f.Name]; !ok {
+		if _, ok := flagCompletions[f.Name]; !ok && f.Value.Type() != "bool" {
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

edsantiago:
How/where does this test get run? I tried deliberately reintroducing the `skip-etc` failure, then `make test-unit`, then variations of `go test`, but cannot get anything to fail.

Luap99:
`go test ./pkg/cli/` is what I used locally however if it does not fail via `make test-unit` there might be something wrong in the makefile/CI setup.

edsantiago:
Yeah, that's what I finally got to work:
```console
GO_TEST=GO111MODULE=on go test -v ./pkg/cli
=== RUN   TestUserNsFlagsCompletion
--- PASS: TestUserNsFlagsCompletion (0.00s)
=== RUN   TestNameSpaceFlagsCompletion
--- PASS: TestNameSpaceFlagsCompletion (0.00s)
=== RUN   TestBudFlagsCompletion
--- PASS: TestBudFlagsCompletion (0.00s)
=== RUN   TestFromAndBudFlagsCompletions
--- PASS: TestFromAndBudFlagsCompletions (0.00s)
PASS
ok      github.com/containers/buildah/pkg/cli   0.011s
```
I'm waiting for the `Unit` CI step to finish; then I'll look to see if the log includes `TestBudFlagsCompletion`.

edsantiago:
The problem is likely that I `^C`ed from `make test-unit`, because it was hanging too long. It's likely that the new test runs after the hang. Anyhow, I made the test fail by instrumenting `pkg/cli/common.go`, so LGTM.

PR #4269, opened 2022-09-20
vendor: update c/storage and c/common

giuseppe:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

/kind other

```release-note
None
```


flouthoc:
Using following `.cirrus.yml` should omit all the other tests except `Containerized Integration` https://github.com/flouthoc/buildah/blob/do-not-merge/.cirrus.yml

giuseppe:
@rhatdan @flouthoc PTAL

giuseppe:
added a new patch to update c/image as well

contrib/cirrus/lib.sh
```diff
@@ -181,10 +181,9 @@ in_podman() {
     done
     showrun podman run -i --name="$IN_PODMAN_NAME" \
                    --net="container:registry" \
-                   --security-opt=label=disable \
-                   --security-opt=seccomp=unconfined \
-                   --cap-add=all \
                    --env-file=$envfile \
+                   --privileged \
+                   --cgroupns=host \
```
flouthoc, line 186: I think PR works without `--cgroupns=host` do we need this for any special reason.

flouthoc (approved):
LGTM
Just one question above.

contrib/cirrus/lib.sh
giuseppe: I think it is good to add it.  We are bind mounting `/sys/fs/cgroup` from the host, so the expectation is that the same cgroups are visible within the container

flouthoc: ah I see, got it.

flouthoc:
Needs a force merge since branch is outdated.

PR #4275, opened 2022-09-20
[release-1.26] bump github.com/theupdateframework/go-tuf

nalind:
Bump github.com/theupdateframework/go-tuf, which is an indirect dependency, to get the fix for GHSA-3633-5h82-39pq, though it doesn't appear to affect any of the packages we use, so it's more bookkeeping than anything.


flouthoc (approved):
/hold cancel
/lgtm

PR #4282, opened 2022-09-20
[release-1.26] tag v1.26.5

nalind:
Tag a v1.26.5 release so that our consumers can pick up a fix for GHSA-rc4r-wh2q-q6c4 (the run user's primary GID not being in the supplemental groups list).


flouthoc (approved):
LGTM
/lgtm
/approve

PR #4285, opened 2022-09-21
copier: work around freebsd bug for "mkdir /"

dfr:
This call to os.Mkdir got missed the first time I tried to add the workaround for the FreeBSD EISDIR bug since I was only testing with 'buildah run'. Trying to use 'buildah add' triggers the same bug whan trying to extract an archive into a container's filesystem.

/kind bug

This works around a kernel bug in FreeBSD (https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=59739) where it can return EISDIR instead of EEXIST when calling mkdir on the root directory (including the 'root' of a chroot or jail).

#### How to verify it

Run 'buildah add <containername> <non-empty-tarfile> /'

None

```release-note
None
```
[NO NEW TESTS NEEDED]

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -1794,7 +1794,9 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
 					}
 				}
 			case tar.TypeDir:
-				if err = os.Mkdir(path, 0700); err != nil && errors.Is(err, os.ErrExist) {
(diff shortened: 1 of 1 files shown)


flouthoc (changes requested):
LGTM @dfr needs `[NO NEW TESTS NEEDED]` in commit message.

rhatdan:
Nice a 20 year old bug

Issue #4276, opened 2022-09-20
architecture variant (v8) is not set on arm64


kit-ty-kate:
**Description**

Using the following dockerfile:
```
FROM --platform=freebsd/arm64/v8 scratch
COPY some-empty-file /some-empty-file
```
`buildah build` will create an image with `os=freebsd`, `architecture=arm64` but the `variant` (`v8`) is not set (as shown by `buildah inspect`)

If you replace `--platform=freebsd/arm64/v8` with `--platform=freebsd/arm/v7` the `variant` is set this time.

**Describe the results you expected:**

The `variant` variable should probably be set, though I'm not sure what is supposed to be the default. I'm only basing my assumption on the behaviour of docker.

**Output of `buildah version`:**

```
buildah version 1.27.1 (image-spec 1.0.2-dev, runtime-spec 1.0.2-dev)
```

flouthoc:
@kit-ty-kate Thanks for creating the issue, above PR should close this https://github.com/containers/buildah/pull/4286

PR #4286, opened 2022-09-22
config,builder: process `variant` while populating image spec

flouthoc:
When processing config for a builder and populating image-spec we must process Variant along with Architecture and Operating system.

[NO NEW TESTS NEEDED]
Closes: https://github.com/containers/buildah/issues/4276

```release-note
config,builder: process `variant` while populating image-spec for Docker and OCI formats
```


config.go
```diff
@@ -140,6 +141,15 @@ func (b *Builder) fixupConfig(sys *types.SystemContext) {
 		b.SetArchitecture(ps.Architecture)
 		b.SetVariant(ps.Variant)
 	}
+	if b.Variant() == "" {
+		if sys != nil && sys.VariantChoice != "" {
+			b.SetVariant(sys.VariantChoice)
+		}
+		// in case the arch string we started with was shorthand for a known arch+variant pair, normalize it
+		_, normalizedArch, normalizedVariant := libimage.NormalizePlatform(b.OS(), b.Architecture(), b.Variant())
```
flouthoc, line 149: Using `libimage.NormalizePlatform` instead of directly calling `platforms.Normalize` cause https://github.com/containerd/containerd/pull/2417 made a change which will return empty string for certain variant and to resolve correct variant we must use `platforms.Parse` which `NormalizePlatform` already uses so instead use that.

flouthoc:
PR is applicable for both `Docker` and `OCI` format

flouthoc:
Failing test looks like a registry issue, restarted all tests.

flouthoc:
@containers/buildah-maintainers @nalind @rhatdan @giuseppe @vrothberg PTAL

rhatdan:
LGTM
@nalind @giuseppe @vrothberg PTAL

config.go
nalind: Why do we want to deviate from what `platforms.Normalize()` is doing here?

flouthoc: I think we must use `platforms.Parse` after generating output from `platforms.Normalize` since `platforms.Normalize` return empty value for certain variants for example consider a case of `arm64` and variant `v8`, but `platforms.Parse` sets the correct value after reading this empty value. I think change went here: https://github.com/containerd/containerd/pull/2417

Since `libimage.NormalizePlatform` was already using `platforms.Parse` hence i decided to re-use it instead of calling `platform.Normalize` and then `platform.Parse(`.

nalind: Then should we be using `libimage.NormalizePlatform()` in other places that we currently call `platforms.Normalize()`?

flouthoc: Yes I agree, I was trying for this change in the same PR but it looked out of scope. Should we do it in a new PR ?

nalind: I don't see the benefit of creating an internal inconsistency that we know we'll have to come back and fix later when we're well-situated to just sort it all out up front.

flouthoc: Added a new commit which does this change and a test was retrofitted which sets `variant` without any `--variant` flag to the image.

flouthoc:
@nalind PTAL again.

tests/inspect.bats
```diff
@@ -115,9 +115,9 @@ load helpers
 }
 
 @test "inspect-format-docker-variant" {
-	# github.com/containerd/containerd/platforms.Normalize() converts Arch:"armhf" to Arch:"arm"+Variant:"v7",
+	# libimage.Normalize() converts Arch:"armhf" to Arch:"arm",
 	# so check that platform normalization happens at least for that one
-	run_buildah from --quiet --pull=false $WITH_POLICY_JSON --arch=armhf scratch
+	run_buildah from --quiet --pull=false $WITH_POLICY_JSON --arch=armhf --variant=v7 scratch
```
nalind, line 120: What happens now if `--variant` _isn't_ specified?

flouthoc: The `--variant` is not set, since internally `libimage.NormalizePlatform` does not calls `platform.Normalize`.  But if we want to retain the original behavior I think `libimage.NormalizePlatform` must be fixed.

flouthoc: Same goes for `docker` i think if `variant` is not specified it is not set, consider the case of
 
`Docker build`: If variant is not specified it is not set
* `sudo docker build --no-cache --platform linux/armhf -t test .`

`buildkit`: Behaves same as current buildah behavior, sets variant `v7` even if variant was not specified
* `sudo docker buildx --no-cache --platform linux/armhf -t test .`

flouthoc: I created a tryout PR in `c/common` https://github.com/containers/common/pull/1164 which makes `libimage.NormalizePlatform` work as it works for `buildkit` and also sets valid variant when it is explicitly specified as in case of `freebsd/arm64/v8`.

So following change is not needed.

flouthoc: Although this difference between `buildkit` and `docker` looks really confusing to me plus buildkit behavior has a bug that it ignores the variant `v8` even it is set explicitly for certain cases like `--platform freebsd/arm64/v8`.

flouthoc:
I think `c/common` regressed here: https://github.com/containers/common/pull/1162

tests/inspect.bats
flouthoc: Resolving this since this is fixed in c/common.

PR #4287, opened 2022-09-23
[release-1.19] add: fail on bad `HTTP` response instead of writing to container for `URL` sources

flouthoc:

```Dockerfile
```

**Fixes: BZ#2102140**
Backport of: https://github.com/containers/buildah/pull/4086

### DIFF
--- a/add.go
+++ b/add.go
@@ -82,6 +82,11 @@ func getURL(src string, chown *idtools.IDPair, mountpoint, renameTarget string, 
 		return err
 	}
 	defer response.Body.Close()
+
+	if response.StatusCode < http.StatusOK || response.StatusCode >= http.StatusBadRequest {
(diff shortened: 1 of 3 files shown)


flouthoc:
Needed By **BZ#2102140**

flouthoc:
@cevich It the failing test because of missing gcp missing image ?

cevich:
Actually that looks like the github problem we were having yesterday.  Cirrus-support supposedly added a workaround for it.  I re-ran, let's see if it passes.

cevich:
dangit.

cevich:
I'm okay with a manual force-merge here on this old branch, I just don't want to be the one that presses the button :grin: 

(since basically no tests are running here)

TomSweeneyRedHat:
LGTM

rhatdan:
LGTM

PR #4288, opened 2022-09-23
[release-1.24] add: fail on bad `HTTP` response instead of writing to container for `URL` sources

flouthoc:

```Dockerfile
```

**Fixes: BZ#2102140**
Backport of: https://github.com/containers/buildah/pull/4086

### DIFF
--- a/add.go
+++ b/add.go
@@ -88,6 +88,11 @@ func getURL(src string, chown *idtools.IDPair, mountpoint, renameTarget string, 
 		return err
 	}
 	defer response.Body.Close()
+
+	if response.StatusCode < http.StatusOK || response.StatusCode >= http.StatusBadRequest {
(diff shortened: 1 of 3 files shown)


flouthoc:
Needed By **BZ#2102140**

flouthoc:

cevich:
> @cevich Similar in this PR is failing VM img. keepalive because of missing gcp image ? Any hints how can i make it green.

Yeah, same as https://github.com/containers/buildah/pull/4287#issuecomment-1256295680  If images are missing that task is pretty noisy and clear about that as the cause.  I'm okay with a manual merge here.

rhatdan:
LGTM

TomSweeneyRedHat:
LGTM fwiw

PR #2332, opened 2020-04-27
Fix permissions on containers.conf

rhatdan:
When running in a rootless buildah, the build user
is not allowed to read /etc/containers/containers.conf
which is causing these containers to fail.

This patch makes containers.conf world readable.

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### Does this PR introduce a user-facing change?

```release-note

```

PR #4290, opened 2022-09-24
[ci:docs] Fix buildahimage container.conf permissions regression

AndASM:
#### What type of PR is this?

/kind bug

When running `quay.io/buildah/stable:latest` (`quay.io/buildah/stable:v1.27.0`), buildah cannot be run as the build user due to a permissions error on `/etc/containers/containers.conf`. This appears to be a regression of the fix @rhatdan made in #2332. The regression appears to have been caused by 0b7c50023bd507dd7949e092dc52824962864a03. Since it's a trivial fix and I appreciate all the effort, I thought I'd make this as a pull request not just an issue.

#### How to verify it
1. Run `podman run --user build --security-opt label=disable --security-opt unmask=ALL --device /dev/fuse -ti quay.io/buildah/stable:v1.27.0`
2. Run `buildah`
```
WARN[0000] Error loading container config when searching for local runtime: reading system config "/etc/containers/containers.conf": decode configuration /etc/containers/containers.conf: open /etc/containers/containers.conf: permission denied
ERRO[0000] failed to setup From and Build flags: failed to get container config: reading system config "/etc/containers/containers.conf": decode configuration /etc/containers/containers.conf: open /etc/containers/containers.conf: permission denied
```

None

Sorry, I didn't feel like creating a new account and setting up signatures. The code is a single line copied and pasted from a [previous commit](https://github.com/containers/buildah/commit/0b7c50023bd507dd7949e092dc52824962864a03). Feel free to remake it if my pseudonym is an issue.

```release-note
None
```

### DIFF
--- a/contrib/buildahimage/upstream/Containerfile
+++ b/contrib/buildahimage/upstream/Containerfile
@@ -43,7 +43,7 @@ RUN sed -e 's|^#mount_program|mount_program|g' \
         -e 's|^mountopt[[:space:]]*=.*$|mountopt = "nodev,fsync=0"|g' \
         /usr/share/containers/storage.conf \
         > /etc/containers/storage.conf && \
-    chmod 644 /etc/containers/storage.conf; \
+    chmod 644 /etc/containers/storage.conf && \
     chmod 644 /etc/containers/containers.conf
 
 RUN mkdir -p /var/lib/shared/overlay-images \
--- a/contrib/buildahimage/upstream/Containerfile
+++ b/contrib/buildahimage/upstream/Containerfile
@@ -43,7 +43,8 @@ RUN sed -e 's|^#mount_program|mount_program|g' \
         -e 's|^mountopt[[:space:]]*=.*$|mountopt = "nodev,fsync=0"|g' \
         /usr/share/containers/storage.conf \
         > /etc/containers/storage.conf && \
-    chmod 644 /etc/containers/storage.conf
+    chmod 644 /etc/containers/storage.conf; \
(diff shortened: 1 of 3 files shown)


rhatdan:
Thanks @AndASM 
LGTM
But you need to sign your commit
git commit -a --amend -s
git push --force

contrib/buildahimage/upstream/Containerfile
```diff
@@ -43,7 +43,8 @@ RUN sed -e 's|^#mount_program|mount_program|g' \
         -e 's|^mountopt[[:space:]]*=.*$|mountopt = "nodev,fsync=0"|g' \
         /usr/share/containers/storage.conf \
         > /etc/containers/storage.conf && \
-    chmod 644 /etc/containers/storage.conf
+    chmod 644 /etc/containers/storage.conf; \
```
nalind, line 46: Please use `&&` instead of `;`, so that an error from `sed` won't just be ignored (the difference between `bash -c "false && echo ok"` and `bash -c "false; echo ok"`).  It's weird that we haven't been consistent about this between the various files.

AndASM: I saw that. I had mimicked the existing use from one of the three files. I thought it was suspicious, but assumed there may be a reason for it.

I've pushed a second commit that changes all the inconsistent `;` usage to be `&&` in all three of the containerfiles.

nalind: Thanks!

PR #4295, opened 2022-09-27
Check for unset build args after setting TARGET args

### LINKED ISSUE #1368: buildah --platform to take os/arch format and set default ARGS {BUILD|TARGET}{PLATFORM|OS|ARCH|VARIANT}
When building on docker a number of default arguments can be used to create a cross build package explicit arguments without relying on on shell `RUN` commands to determine the native architecture.

Docker ref:  https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

These could also set the required docker v1/v2 schema without the use of `buildah config`.

lucacome:
Signed-off-by: Luca Comellini <luca.com@gmail.com>

#### What type of PR is this?

/kind cleanup

When using the `--platform` flag (or `--os`, `--arch` etc) the `TARGETARCH` `TARGETOS` `TARGETPLATFORM` `TARGETVARIANT` args are set but `buildah` will complain with something like

```
WARN[0000] missing "TARGETARCH" build argument. Try adding "--build-arg TARGETARCH=<VALUE>" to the command line
```

By moving the check for missing arguments after these values are set, the warnings are gone.

#### How to verify it

add `ARG TARGETARCH` to a Dockerfile

run 
```
buildah bud -f build/Dockerfile --os=linux --arch=arm64
```
or
```
buildah bud -f build/Dockerfile --platform linux/arm64
```

with this change there are no warnings.

Fixes #1368

```release-note
buildah will no longer mistakenly warn on missing variant when using a --platform flag
```

[NO NEW TESTS NEEDED]


flouthoc:
@lucacome Thanks for the PR, I think adding a test for this in `tests/bud.bats` should be really easy, Would it be possible for you to add a test for this ?

rhatdan:
/approve
LGTM

lucacome:
@flouthoc it seems like the tests won't run on Mac so it won't be very straightforward for me to add a test.
I could try spinning up a VM... 🤔

rhatdan:
Added [NO NEW TESTS NEEDED] flag so w can get this in, and start to cut a release.

Could @lucacome or @flouthoc add a test for this in a separate PR, that we can merge later.

flouthoc:
I'll add a test for this.

flouthoc:
PR for missing test: https://github.com/containers/buildah/pull/4310

PR #4297, opened 2022-09-27
Proof of concept: nightly dependency treadmill

edsantiago:
As discussed in f2f: this is the cleanest, simplest mechanism
I can think of to auto-test the Big Three dependencies: simply
run go mod edit immediately after git checkout, then run the
entire CI test suite.

If this approach works, we can set up a new CIRRUS_CRON=treadmill job.

```release-note
None
```

### DIFF
--- a/contrib/cirrus/build.sh
+++ b/contrib/cirrus/build.sh
@@ -15,6 +15,27 @@ if [[ "$IN_PODMAN" == "true" ]]
 then
     in_podman --rm $IN_PODMAN_NAME $0
 else
+    # Nightly dependency-bump job: fetch latest versions of the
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

edsantiago:
I'm going to take a wild guess, and assert that [run failures](https://api.cirrus-ci.com/v1/task/5400821549498368/logs/integration_test.log) are caused by https://github.com/containers/common/pull/1162 , and that this treadmill is actually working as expected?

edsantiago:
This is ready for review and merge (as soon as CI passes, which it will). I've added a `treadmill` cron job in Cirrus, set to run daily.

vrothberg (approved):
/lgtm

PR #4310, opened 2022-10-01
test: warning behaviour for unset/set `TARGETOS`,`TARGETARCH`,`TARGETPLATFORM`

flouthoc:
Test buildah's warning behaviors for unset or set `TARGETOS`, `TARGETARCH`, `TARGETPLATFORM` or `TARGETVARIANT` args when set or unset directly or indirectly by the user.

Adds missing test for: https://github.com/containers/buildah/pull/4295

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -25,6 +25,35 @@ load helpers
   expect_output --substring "options use-vc"
 }
 
+# Test for https://github.com/containers/buildah/pull/4295
+@test "build test warning for preconfigured TARGETARCH, TARGETOS, TARGETPLATFORM or TARGETVARIANT" {
+  _prefetch alpine
+  containerfile=${TEST_SCRATCH_DIR}/Containerfile
+
+  # buildah should not warn when any of these is indirectly set by cli options
+  local -a checkvars=(ARCH OS PLATFORM VARIANT)
+  echo "FROM alpine" >$containerfile
+  for var in "${checkvars[@]}"; do
+    echo "ARG TARGET${var}" >>$containerfile
+  done
+  echo "COPY . ." >>$containerfile
+
+  # With explicit and full --platform, there should be no warnings
+  run_buildah build $WITH_POLICY_JSON --platform linux/amd64/v2 -t source -f $containerfile
+  assert "$output" !~ "missing .* build argument" \
+         "With explicit --platform, buildah should not warn"
+
+  # Without --platform, we want one warning for each variable
+  run_buildah build $WITH_POLICY_JSON -t source -f $containerfile
+#  for var in "${checkvars[@]}"; do
+  for i in $(seq 0 3); do
+    var="TARGET${checkvars[$i]}"
+    # The '..'s cover backslash-quote: "missing \"TARGETFOO\" ..."
+    assert "${lines[$i]}" =~ "missing ..${var}.. build argument" \
+           "buildah should warn about undefined ${var}"
+  done
+}
+
 @test "build-conflicting-isolation-chroot-and-network" {
   _prefetch alpine
   cat > ${TEST_SCRATCH_DIR}/Containerfile << _EOF


flouthoc:
@lucacome @rhatdan @nalind @containers/buildah-maintainers PTAL

rhatdan:
LGTM
@edsantiago PTAL

edsantiago (changes requested):
Don't merge yet please, I have a few concerns. Am looking into ways to confirm and address.

tests/bud.bats
```diff
@@ -25,6 +25,60 @@ load helpers
...
+  # buildah should not warn when TARGETARCH, TARGETOS, TARGETPLATFORM or TARGETVARIANT is indirectly set by cli options
+  cat > ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile << _EOF
+FROM alpine
+ARG TARGETARCH
+COPY . .
+_EOF
+  run_buildah build $WITH_POLICY_JSON --platform linux/amd64 -t source -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile
+  assert "$output" !~ "--build-arg TARGETARCH"
+
+  cat > ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile << _EOF
```
edsantiago, line 41: A lot of duplication

tests/bud.bats
```diff
@@ -25,6 +25,60 @@ load helpers
...
+  mkdir -p ${TEST_SCRATCH_DIR}/bud/platform
+
+  # buildah should not warn when TARGETARCH, TARGETOS, TARGETPLATFORM or TARGETVARIANT is indirectly set by cli options
+  cat > ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile << _EOF
+FROM alpine
+ARG TARGETARCH
+COPY . .
+_EOF
+  run_buildah build $WITH_POLICY_JSON --platform linux/amd64 -t source -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile
+  assert "$output" !~ "--build-arg TARGETARCH"
```
edsantiago, line 39: Too explicit. It's much better to have a looser test here, such that any warning will cause a test failure.

tests/bud.bats
```diff
@@ -25,6 +25,60 @@ load helpers
...
+ARG TARGETARCH
+ARG TARGETPLATFORM
+ARG TARGETVARIANT
+COPY . .
+_EOF
+  run_buildah build $WITH_POLICY_JSON -t source -f ${TEST_SCRATCH_DIR}/bud/platform/Dockerfile
+  expect_output --substring "build-arg TARGETOS"
+  expect_output --substring "build-arg TARGETARCH"
+  expect_output --substring "build-arg TARGETPLATFORM"
+  expect_output --substring "build-arg TARGETVARIANT"
```
edsantiago, line 78: lots of duplication

edsantiago (changes requested):
Here's a shorter, cleaner version that uses only one containerfile for all tests, and is more careful about checking for warnings:
```bash
@test "build test warning for preconfigured TARGETARCH, TARGETOS, TARGETPLATFORM or TARGETVARIANT" {
  _prefetch alpine
  containerfile=${TEST_SCRATCH_DIR}/Containerfile

  # buildah should not warn when any of these is indirectly set by cli options
  local -a checkvars=(ARCH OS PLATFORM VARIANT)
  echo "FROM alpine" >$containerfile
  for var in "${checkvars[@]}"; do
    echo "ARG TARGET${var}" >>$containerfile
  done
  echo "COPY . ." >>$containerfile

  # With explicit and full --platform, there should be no warnings
  run_buildah build $WITH_POLICY_JSON --platform linux/amd64/v2 -t source -f $containerfile
  assert "$output" !~ "missing .* build argument" \
         "With explicit --platform, buildah should not warn"

  # Without --platform, we want one warning for each variable
  run_buildah build $WITH_POLICY_JSON -t source -f $containerfile
#  for var in "${checkvars[@]}"; do
  for i in $(seq 0 3); do
    var="TARGET${checkvars[$i]}"
    # The '..'s cover backslash-quote: "missing \"TARGETFOO\" ..."
    assert "${lines[$i]}" =~ "missing ..${var}.. build argument" \
           "buildah should warn about undefined ${var}"
  done
}
```
Incidentally, is `--os linux` or `--os linux/amd64` supposed to shutup the `missing TARGETOS` warning? Because it doesn't. I gave up trying to add tests for that.

flouthoc:
Thanks @edsantiago for the patch.

> Incidentally, is --os linux or --os linux/amd64 supposed to shutup the missing TARGETOS warning? Because it doesn't. I gave up trying to add tests for that.

`--os` doesn't do but `--platform` does, I think `--os` needs to be fixed, I'll open a new PR fixing `--os` and will also amend the test then.

edsantiago:
Thank you!

/lgtm

edsantiago (approved):

Issue #4314, opened 2022-10-04
COPY DOT DOT considered harmful


edsantiago:
Followup to #4310. Test is [failing under podman](https://api.cirrus-ci.com/v1/artifact/task/5960273050730496/html/bud-podman-fedora-36-root-host.log.html#t--00004):
```
STEP 6/6: COPY . .
Error: building at STEP "COPY . .": reading "/var/tmp/buildah_tests.hnlvcs": error during bulk transfer for copier.request{Request:"GET", Root:"/", preservedRoot:"/var/tmp/buildah_tests.hnlvcs", rootPrefix:"/var/tmp/buildah_tests.hnlvcs", Directory:"/", preservedDirectory:"/var/tmp/buildah_tests.hnlvcs", Globs:[]string{"/"}, preservedGlobs:[]string{"/var/tmp/buildah_tests.hnlvcs"}, StatOptions:copier.StatOptions{CheckForArchives:false, Excludes:[]string(nil)}, GetOptions:copier.GetOptions{UIDMap:[]idtools.IDMap(nil), GIDMap:[]idtools.IDMap(nil), Excludes:[]string(nil), ExpandArchives:false, Cho
...long long long long list of stuff
```
Is it super-important to use `COPY . .`? Can it be just `RUN echo hi` or something?

Anyhow, @flouthoc, if you're working on the `--os` thing, could you also fix this?

flouthoc:
@edsantiago Not a real bug, it looks like code is hitting system/env limit for path length. Its just that context dir in podman and buildah tests have different content. But I agree the test does not needs a blanket/expensive `COPY . .` therefore removing it here: https://github.com/containers/buildah/pull/4316

PR #4316, opened 2022-10-05
test: remove unneeded/expensive `COPY . .`

flouthoc:
Test added in https://github.com/containers/buildah/pull/4310 uses `COPY . .` which is expensive and unneeded. Also in `podman` test due to copying files directly from the currect context it hits a system/environment limit. See https://api.cirrus-ci.com/v1/artifact/task/5960273050730496/html/bud-podman-fedora-36-root-host.log.html#t--00004

Hence remove this uneeded `COPY` and instead use a `RUN echo`.

Closes: https://github.com/containers/buildah/issues/4314

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -36,7 +36,7 @@ load helpers
   for var in "${checkvars[@]}"; do
     echo "ARG TARGET${var}" >>$containerfile
   done
-  echo "COPY . ." >>$containerfile
+  echo "RUN echo hi" >>$containerfile
 
   # With explicit and full --platform, there should be no warnings
   run_buildah build $WITH_POLICY_JSON --platform linux/amd64/v2 -t source -f $containerfile


edsantiago (approved):
Thank you!

tests/bud.bats
```diff
@@ -36,7 +36,7 @@ load helpers
   for var in "${checkvars[@]}"; do
     echo "ARG TARGET${var}" >>$containerfile
   done
-  echo "COPY . ." >>$containerfile
+  echo "RUN echo hi" >>$containerfile
```
nalind, line 39: Given that this is built explicitly for `linux/amd64/v2` below, won't this fail on other architectures?

edsantiago: So, that confused me when I first looked at this. Doesn't buildah automatically use the Containerfile directory as its context directory? Shouldn't `COPY . .` run from that empty directory? I tried changing `-f $containerfile` to `${TEST_SCRATCH_DIR}` but podman still fails. This seems like a major misunderstanding on my part.

nalind: I would expect `COPY . .` to pick up the Containerfile that's just been written there.  How does it fail?

edsantiago: Problem found. Fix in progress. Totally my fault.

flouthoc: Ah yes, `ENV` or `LABEL` is better here, Ed has already created a PR for that here: https://github.com/containers/buildah/pull/4320

PR #4320, opened 2022-10-05
System test cleanup: document, clarify, fix

edsantiago:
Primary purpose: fix "preconfigured TARGETARCH/etc" test so
it will work under podman and on multiarch.

Root cause of it not working: I mistakenly advised @flouthoc,
in #4310, to write a containerfile in $TEST_SCRATCH_DIR. I
thought it was an empty directory. Big, big mistake. (Sorry,
Aditya). Document this near the variable definition, and
fix the test once again.

@nalind pointed out that the containerfile doesn't need to
be generated on-the-fly, so, use a static one. In the spirit
of DIE, read the TARGETxxx vars from it. Not that we're
expecting more variables, but, it's just cleaner.

Also, as long as I'm here: in run_buildah, when logging the
command being run, use #/$ prompt for root/rootless. I was
getting too confused looking at logs of root runs.

```release-note
none
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -28,30 +28,45 @@ load helpers
 # Test for https://github.com/containers/buildah/pull/4295
 @test "build test warning for preconfigured TARGETARCH, TARGETOS, TARGETPLATFORM or TARGETVARIANT" {
   _prefetch alpine
-  containerfile=${TEST_SCRATCH_DIR}/Containerfile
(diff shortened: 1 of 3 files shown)


flouthoc:
@edsantiago No issues at all :)

flouthoc (approved):
LGTM
/lgtm
/approved

Issue #4315, opened 2022-10-04
--cache-from does not work with ARG


finkandreas:
I have the feeling that `--cache-from` does not work. Here are my steps to reproduce:

My Dockerfile:
```
FROM ubuntu:22.04

ARG NUM_PROCS=1

# update system
RUN echo "Running with $NUM_PROCS procs" && apt-get update && env DEBIAN_FRONTEND=noninteractive TZ=Europe/Zurich apt-get -yqq install --no-install-recommends build-essential cmake
```

Build image steps:
```
# work inside a container
docker run --privileged --rm -v Dockerfile:Dockerfile -it quay.io/buildah/stable:latest /bin/bash
buildah login my_registry.net
buildah bud -f Dockerfile --cache-from my_registry/path/to/podman_cache --cache_to my_registry/path/to/podman_cache --cache-ttl=600h --layers .
```

At this point I have built the image and during the build I was seeing log lines that told that buildah was pushing to cache, also I see cache layers in the repository.

Now redo the build steps:
```
# work inside a container
docker run --privileged --rm -v Dockerfile:Dockerfile -it quay.io/buildah/stable:latest /bin/bash
buildah login my_registry.net
buildah bud -f Dockerfile --cache-from my_registry/path/to/podman_cache --cache_to my_registry/path/to/podman_cache --cache-ttl=600h --layers .
```

I would assume that the second time I run the build it would reuse the cache for the `apt-get update ...` RUN step. But it is not considering the cache. I do not see any log output why it does not use any cache.

flyingsaucerer:
Agreed. --cached-from has a 100% cache miss depending on where build args are specified in the dockerfile

omacranger:
Oddly enough we noticed the same thing this morning with 1.27.0. We could not get remote caches to function with `--cache-from` before happening to try another dockerfile with no args present. Moving the args (that never changed) to be present just before use allowed previous layers to leverage the cache, but still not the layers using the args. I do not have the full dockerfile, but something like below was what we ran into: 

Did not work:

```Dockerfile
ARG ARG_VALUE

RUN npm install

RUN echo $ARG_VALUE 
```

Partially worked (cached the first layer, though not the second despite the arg value not changing):

```Dockerfile
RUN npm install

ARG ARG_VALUE

RUN echo $ARG_VALUE 
```

flyingsaucerer:
@rhatdan @flouthoc Bumping this to get your attention. The problem is pretty severe.

[...]

PR #4326, opened 2022-10-10
imagebuildah, `cache-from`: pull cache from remote repo after adding content summary

flouthoc:
For certain instruction actual history matching makes sense after content summary is processed and added, In such cases generated `cacheKey` for distributed cache is changed hence for such cases attempt to pull available cache from remote repo once `cacheKey` is generated again after adding content summary.

Example use case which works with `--cache-from` and `--cache-to` after this commit.

```Dockerfile
FROM alpine
ARG VAR=hello
RUN echo "Hello $VAR"
```

Closes: https://github.com/containers/buildah/issues/4315

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -1302,6 +1302,24 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 				if err != nil {
 					return "", nil, fmt.Errorf("checking if cached image exists from a previous build: %w", err)
 				}
+				// All the best effort to find cache on localstorage have failed try pulling
+				// cache from remote repo if `--cache-from` was configured and cacheKey was
+				// generated again after adding content summary.
+				if cacheID == "" && s.executor.cacheFrom != nil {
+					// only attempt to use cache again if pulling was successful
+					// otherwise do nothing and attempt to run the step, err != nil
+					// is ignored and will be automatically logged for --log-level debug
+					if id, err := s.pullCache(ctx, cacheKey); id != "" && err == nil {
+						logCachePulled(cacheKey)
+						cacheID, err = s.intermediateImageExists(ctx, node, addedContentSummary, s.stepRequiresLayer(step))
+						if err != nil {
+							return "", nil, fmt.Errorf("checking if cached image exists from a previous build: %w", err)
+						}
+						if cacheID != "" {
+							pulledAndUsedCacheImage = true
+						}
+					}
+				}
 			}
 		} else {
 			// This log line is majorly here so we can verify in tests
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -4436,6 +4436,75 @@ _EOF
 
 }
 
+@test "build test pushing and pulling from remote cache sources - after adding content summary" {
+  _prefetch alpine
+
+  start_registry
+  run_buildah login --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth --username testuser --password testpassword localhost:${REGISTRY_PORT}
+
+  # ------ Test case ------ #
+  # prepare expected output beforehand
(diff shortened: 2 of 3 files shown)


flouthoc:

rhatdan:
LGTM

rhatdan:
@nalind PTAL

TomSweeneyRedHat:
@flouthoc you may need to rebase this one

imagebuildah/stage_executor.go
```diff
@@ -1302,6 +1302,24 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 				if err != nil {
 					return "", nil, fmt.Errorf("checking if cached image exists from a previous build: %w", err)
 				}
+				// All the best effort to find cache on localstorage have failed try pulling
+				// cache from remote repo if `--cache-from` was configured and cacheKey was
```
TomSweeneyRedHat, line 1306: should it be "cacheKey" or "cacheID"?

TomSweeneyRedHat:
One possible nit, but definitely would like a head nod from @nalind

imagebuildah/stage_executor.go
flouthoc: `cacheKey` is the hash or key generated for a particular intermediate image as also used as tag when pushed to remote repo, the keyword was introduced here: https://github.com/containers/buildah/pull/4118

I think difference between `key` and  `id` is , the `ID` is always unique but `key` can be same if the same intermediate image is built again.

flouthoc:
Rebased following PR with `main` @containers/buildah-maintainers @nalind PTAL .

imagebuildah/stage_executor.go
nalind: The ID is an image ID, which is usually the digest of that image's config blob.  That config blob includes data which we don't consider relevant when evaluating whether or not a given image should be considered a cache hit.

nalind:
LGTM

edsantiago:
Heads-up to everyone involved, the buildah treadmill found that this is not working in podman-remote:
```
#|     FAIL: podman-remote build ....
#| expected: =~ 'STEP 2/3: ARG VAR=hello
   --> Cache pulled from remote'
#|   actual:    'STEP 1/3: FROM alpine'
#| ...
#|         >    'STEP 2/3: ARG VAR=hello'                                                                                                                    
#|         >    '--> Pushing cache localhost:39951/temp:73dcec7230efcfd8cd8e041d68f6c74b04f1242fc3c1e67c64b0c92abcaa2d28'                                   
#|         >    '--> 4e92a602438'
```
i.e., "Cache pulled from remote" does not appear in the output.

For now I've just added `skip_if_remote` to the treadmill; it will need to be addressed for real in the next buildah vendor.

flouthoc:
> Heads-up to everyone involved, the buildah treadmill found that this is not working in podman-remote:
> 
> ```
> #|     FAIL: podman-remote build ....
> #| expected: =~ 'STEP 2/3: ARG VAR=hello
>    --> Cache pulled from remote'
> #|   actual:    'STEP 1/3: FROM alpine'
> #| ...
> #|         >    'STEP 2/3: ARG VAR=hello'                                                                                                                    
> #|         >    '--> Pushing cache localhost:39951/temp:73dcec7230efcfd8cd8e041d68f6c74b04f1242fc3c1e67c64b0c92abcaa2d28'                                   
> #|         >    '--> 4e92a602438'
> ```
> 
> i.e., "Cache pulled from remote" does not appear in the output.
> 
> For now I've just added `skip_if_remote` to the treadmill; it will need to be addressed for real in the next buildah vendor.

@edsantiago I'll take a look at this thanks.

flouthoc:
@edsantiago Following PR should fix this https://github.com/containers/buildah/pull/4386

Issue #4303, opened 2022-09-30
Buildah warns about missing argument in skipped stage


lucacome:
**Description**

```
FROM alpine:latest as base
RUN echo "base"


FROM debian:latest as foo
ARG FOO_BAR
RUN echo "debian"


FROM base as final
RUN echo "final"
```
When trying to build this example Dockerfile with `buildah build --target final .`, buildah will process all the stages and throw a warning:

`WARN[0000] missing "FOO_BAR" build argument. Try adding "--build-arg FOO_BAR=<VALUE>" to the command line`

even though the stage with the argument is not used.

**Steps to reproduce the issue:**
1. Use the Dockerfile above
2. Run `buildah build --target final .`
3. See the output

**Describe the results you received:**

```
❯ buildah build --target final .
WARN[0000] missing "FOO_BAR" build argument. Try adding "--build-arg FOO_BAR=<VALUE>" to the command line
[1/3] STEP 1/2: FROM alpine:latest AS base
```

**Describe the results you expected:**

```
❯ buildah build --target final .
[1/3] STEP 1/2: FROM alpine:latest AS base
```

**Output of `buildah version`:**

```
❯ buildah version
Version:         1.28.0-dev
Go Version:      go1.19.1
Image Spec:      1.0.2-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        1.0.0
libcni Version:  v1.1.2
image Version:   5.22.1-dev
Git Commit:      4b7d3555bfa4440c3c5264ae44b93822e10deec0
Built:           Thu Sep 29 18:44:23 2022
OS/Arch:         darwin/arm64
BuildPlatform:   darwin/arm64/v8
```

flouthoc:
@lucacome Thanks for creating the issue following PR ( https://github.com/containers/buildah/pull/4327 ) should close this.

PR #4327, opened 2022-10-10
imagebuildah: `warnOnUnsetBuildArgs` while processing `stages` from `executor`

flouthoc:
Before this commit buildah produces warning for UnsetBuildArgs at the top level while preparing the build in `imagebuildah/build` hence it makes it hard to honor certain logic like `do not produce warning for stages which are not used` therefore move `warnOnUnsetBuildArgs` to `imagebuldah/executor` so we can warn while processing each stage.

Following commit will prevent warning unnecessarily for stages which are completely skipped.

Closes: https://github.com/containers/buildah/issues/4303


flouthoc:
@containers/buildah-maintainers @nalind @rhatdan PTAL

TomSweeneyRedHat:
LGTM
but it looks like a rebase is needed.

Issue #4245, opened 2022-09-14
error resolving mountpoints for container: write /proc/thread-self/attr/sockcreate: permission denied


edsantiago:
Seen in [gating tests](https://artifacts.dev.testing-farm.io/44f7a614-b253-4717-8022-ff8950a9c97e/) (rawhide, per Lokesh):
```
not ok 697 bud with ssh key
# (from function `die' in file /usr/share/buildah/test/system/helpers.bash, line 314,
#  from function `run_buildah' in file /usr/share/buildah/test/system/helpers.bash, line 301,
#  in test file /usr/share/buildah/test/system/ssh.bats, line 24)
#   `run_buildah bud --ssh default=$mytmpdir/sshkey $WITH_POLICY_JSON  -t sshimg -f $BUDFILES/run-mounts/Dockerfile.ssh $BUDFILES/run-mounts' failed
# /usr/share/buildah/test/system /var/str
# # [checking for: docker.io/library/alpine]
# # [restoring from cache: /var/tmp/buildah-image-cache.5434 / docker.io/library/alpine]
# Getting image source signatures
# Copying blob sha256:9d16cba9fb961d1aafec9542f2bf7cb64acfc55245f9e4eb5abecd4cdc38d749
# Copying config sha256:961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4
# Writing manifest to image destination
# Storing signatures
# $ /usr/bin/buildah bud --ssh default=/var/tmp/buildah_tests.qezr9z/my-dir1/sshkey --signature-policy /usr/share/buildah/test/system/policy.json -t sshimg -f /usr/share/buildah/test/system/bud/run-mounts/Dockerfile.ssh /usr/share/buildah/test/system/bud/run-mounts
# STEP 1/3: FROM alpine
# STEP 2/3: RUN apk add openssh-client
# fetch http://dl-cdn.alpinelinux.org/alpine/v3.10/main/x86_64/APKINDEX.tar.gz
# fetch http://dl-cdn.alpinelinux.org/alpine/v3.10/community/x86_64/APKINDEX.tar.gz
# (1/5) Installing openssh-keygen (8.1_p1-r0)
# (2/5) Installing ncurses-terminfo-base (6.1_p20190518-r2)
# (3/5) Installing ncurses-libs (6.1_p20190518-r2)
# (4/5) Installing libedit (20190324.3.1-r0)
# (5/5) Installing openssh-client (8.1_p1-r0)
[...]

PR #4329, opened 2022-10-11
sshagent: `LockOSThread` before setting `SocketLabel`

flouthoc:
As per the library instruction `sshagent` must acquire `LockOSThread` before setting and resetting the SocketLabel to ensure that the thread is locked before following invocation completes setting and reseting the socket label.

Reference: https://pkg.go.dev/github.com/opencontainers/selinux/go-selinux#SetSocketLabel 
Closes: https://github.com/containers/buildah/issues/4245

[NO NEW TESTS NEEDED]
[NO TESTS NEEDED]
Existing tests must pass

### DIFF
--- a/pkg/sshagent/sshagent.go
+++ b/pkg/sshagent/sshagent.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat (commented):
LGTM

pkg/sshagent/sshagent.go
```diff
@@ -83,8 +91,13 @@ func (a *AgentServer) Serve(processLabel string) (string, error) {
...
 		return "", err
 	}
-	err = selinux.SetSocketLabel("")
-	if err != nil {
+	// Reset socket label.
+	if err := selinux.SetSocketLabel(""); err == nil {
+		// Unlock the thread only if the process label could be restored
+		// successfully.  Otherwise leave the thread locked and the Go runtime
+		// will terminate it once it returns to the threads pool.
+		runtime.UnlockOSThread()
```
rhatdan, line 99: Why not unlock either way after the SetSocketLabel call?

flouthoc: Yes sounds better to me, I was trying to make it more like how its was implemented in libpod but now I think this needs to be corrected in libpod as well.

flouthoc:
Rebased with master , this will need `/lgtm` again.

PR #4347, opened 2022-10-17
[CI:BUILD] pr-should-include-tests: allow specfile, golangci

edsantiago:
Make the pr-should-include-tests check more consistent with
podman: allow changes to buildah.spec.rpkg and .golangci.yml

Also, remove the check for "NO TESTS NEEDED". That was a
workaround for a bad design decision I initially made.
The correct wording is NO NEW TESTS NEEDED, and nobody
should ever use the phrase without NEW.

```release-note
None
```


rhatdan:
LGTM
/approve

PR #4349, opened 2022-10-18
mount,cache: internal `lockfiles` must not be part of user's `cache` content

flouthoc:
Following PR introduces two change

* `--mount=type=cache` must not add internal lockfiles to cache directory
created by users instead store it in a different central directory with
path as `/base/buildah-cache/buildah-lockfiles`. 
There are use-cases where users can wipe cache between the builds so
lockfiles will be removed in unexpected manner and also its not okay to
mix buildah's internal construct with user's cache content.

* **run**: Single `RUN` can contain multiple `--mount` commands so lets append into
`lockedTargets` so we collect `lockfiles` from all the `--mount`
instructions.

Tryout example
```Dockerfile
FROM quay.io/centos/centos:7

ARG WIPE_CACHE

RUN --mount=type=cache,target=/cache1,sharing=locked \
    --mount=type=cache,target=/cache2 \
    set -ex; \
    ls -l /cache1; \
    if [[ -v WIPE_CACHE ]]; then \
      >&2 echo "Wiping cache"; \
      find /cache1 -mindepth 1 -delete; \
    fi; \
    echo "foo" > /cache1/foo.txt; \
    ls -l /cache1; \
    chmod --recursive g=u /cache1; \
    : ;

RUN --mount=type=cache,target=/cache1,sharing=locked \
    >&2 echo "Never get here"
 ```

#### What type of PR is this?

> /kind bug

#### How to verify it
Newly added `integration` test and old tests must not fail.

Closes: https://github.com/containers/buildah/issues/4342

```release-note
mount,cache: lockfiles should not be created with user's cache content
```


internal/parse/parse.go
```diff
@@ -348,18 +354,25 @@ func GetCacheMount(args []string, store storage.Store, imageMountLabel string, a
 		if err != nil {
 			return newMount, lockedTargets, fmt.Errorf("unable to change uid,gid of cache directory: %w", err)
 		}
+
+		// create a subdirectory inside `cacheParent` just to store lockfiles
+		buildahLockFilesDir = filepath.Join(cacheParent, buildahLockFilesDir)
+		err = os.MkdirAll(buildahLockFilesDir, os.FileMode(0755))
```
rhatdan, line 360: Does this lockdir need to be shared between different users?  Should it be 0700?  Does it matter?

flouthoc: @rhatdan You are correct `0700` is better here.

rhatdan:
LGTM
@nalind @vrothberg @giuseppe @umohnani8 PTAL

internal/parse/parse.go
```diff
@@ -33,6 +33,9 @@ const (
 	BuildahCacheDir = "buildah-cache"
 	// mount=type=cache allows users to lock a cache store while its being used by another build
 	BuildahCacheLockfile = "buildah-cache-lockfile"
+	// All the lockfiles are stored in a seperate directory inside `BuildahCacheDir`
```
giuseppe, line 36: typo in `seperate`

giuseppe (approved):
LGTM

internal/parse/parse.go
flouthoc: fixed.

Issue #4312, opened 2022-10-03
multi-stage builds fail to resolve named image if build arg env is in name


ibotty:
**Description**

Multi-stage builds fail to resolve a named image if a build arg  is in the name.  A reproducer is below.

```
ARG my_env=a

FROM scratch as stage_a
RUN /bin/true

FROM stage-${my_env} as stage_b
RUN /bin/true
```

I was trying to build https://github.com/zalando/spilo/blob/master/postgres-appliance/Dockerfile which uses a build arg in `FROM` at a later stage.

**Output of `rpm -q buildah` or `apt list buildah`:**
```
buildah-1.27.0-1.fc37.x86_64
```

**Output of `buildah version`:**

```
Version:         1.27.0
Go Version:      go1.19
Image Spec:      1.0.2-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        1.0.0
libcni Version:  v1.1.2
image Version:   5.22.0
Git Commit:
Built:           Tue Aug  9 04:07:39 2022
OS/Arch:         linux/amd64
BuildPlatform:   linux/amd64
```

flouthoc:
@ibotty Following works if `--build-arg` is explicitly specified.

Try
```Dockerfile
ARG my_env=a

FROM alpine as stage_a
RUN /bin/true

FROM stage_${my_env} as stage_b
RUN /bin/true
```

Command
```sh
buildah build -t test --build-arg my_env=a .
```

Output
```console
buildah build -t test --build-arg my_env=a .
[1/2] STEP 1/2: FROM alpine AS stage_a
[1/2] STEP 2/2: RUN /bin/true
Getting image source signatures
Copying blob 994393dc58e7 skipped: already exists  
Copying blob a5b2c723feaa done  
Copying config 411f196471 done  
Writing manifest to image destination
Storing signatures
--> 411f196471a
[2/2] STEP 1/2: FROM 411f196471a4b7a2f8ac1603c5b5e681357b59eaa23912ebea9df450215a22f5 AS stage_b
[2/2] STEP 2/2: RUN /bin/true
[2/2] COMMIT test
Getting image source signatures
Copying blob 994393dc58e7 skipped: already exists  
Copying blob a5b2c723feaa skipped: already exists  
Copying blob 810d1a8a7a6d done  
Copying config 078f1ab8cd done  
Writing manifest to image destination
Storing signatures
--> 078f1ab8cda
Successfully tagged localhost/test:latest
[...]

PR #4366, opened 2022-10-21
executor: honor default `ARG` values while evaluating base image name.

flouthoc:
While PR https://github.com/containers/buildah/pull/3947 added support for evaluating `--build-args` in base image names for a multi-stage,single-stage builds but it missed processing default value if any. So for scenarios where `ARG` already has a default value in Containerfile via `ARG key=value` but was not specified with `--build-arg key=value` the processing ignored the default value. Following commit just adds support for that.

Makes below Containerfile functional without any external `--build-arg` value from CLI

```Dockerfile
ARG my_env=a

FROM alpine as stage_a
RUN /bin/true

FROM stage_${my_env} as stage_b
RUN /bin/true
```

Closes: https://github.com/containers/buildah/issues/4312

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
multi-stage,executor: honor default `ARG` values while evaluating base image name.
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -2810,6 +2810,10 @@ _EOF
   _prefetch busybox:musl
   target=leading-args
   run_buildah build $WITH_POLICY_JSON -t ${target} --build-arg=VERSION=musl -f $BUDFILES/leading-args/Dockerfile $BUDFILES/leading-args
+
+  #Verify https://github.com/containers/buildah/issues/4312
+  # stage `FROM stage_${my_env}` must be resolved with default arg value and build should be successful.
+  run_buildah build $WITH_POLICY_JSON -t source -f $BUDFILES/multi-stage-builds/Dockerfile.arg_in_stage
 }
 
 @test "bud-with-healthcheck" {
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -690,7 +690,12 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 									base = child.Next.Value
 								}
 							}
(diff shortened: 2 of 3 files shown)


flouthoc:
Test for `--build-args` defined from CLI was already added here: https://github.com/containers/buildah/pull/3947

flouthoc:

tests/bud.bats
```diff
@@ -33,6 +33,12 @@ load helpers
   expect_output --substring "hello"
 }
 
+#Verify https://github.com/containers/buildah/issues/4312
+@test "multi-stage: verify heading arg in base image name" {
+  # stage `FROM stage_${my_env}` must be resolved with default arg value and build should be successful.
+  run_buildah build $WITH_POLICY_JSON -t source -f $BUDFILES/multi-stage-builds/Dockerfile.arg_in_stage
```
nalind, line 39: This could be merged with the "bud with ARG before FROM" test, because they seem to be two halves of the same test.

nalind:
Is this also relevant for single-stage builds?

TomSweeneyRedHat:
LGTM, but it looks like a rebase is needed.  Plus, you should consider @nalind 's comment.

tests/bud.bats
flouthoc: SGTM amended.

flouthoc:
> Is this also relevant for single-stage builds?

@nalind Yes , I have removed `multi-stage` from commit and PR title since this is applicable for both multi-stage and regular builds.

flouthoc:

Issue #4318, opened 2022-10-05
tests: stop using ubi8 (this is the "0x3c" flake)


edsantiago:
Use of `ubi8` images is causing one of our big flakes, mostly in podman (because it runs many more CI jobs than buildah) but also in buildah CI:
```
$ buildah build -f Containerfile --pull=false -q --arch=amd64 -t image-amd --signature-policy /var/tmp/go/src/github.com/containers/buildah/tests/./policy.json /var/tmp/buildah_tests.sufdm8/my-dir
error creating build container: copying system image from manifest list:
     reading signatures:
     parsing signature https://access.redhat.com/webassets/docker/content/sigstore/ubi8-minimal@sha256=SHA/signature-7:
     unrecognized signature format, starting with binary 0x3c
```
I think this is also the cause of the horrid `multiple-platform-no-run` flake, or at least one of the causes.

We can't rely on flaky registries. Can all tests that use `ubi8-*`  be fixed to use a different image?

### [Integration] 295 bud --pull=false --arch test
* [unknown image] : Integration fedora-35 w/ vfs
  * PR #4111
    * [07-14-2022 12:16](https://api.cirrus-ci.com/v1/task/5063862180380672/logs/integration_test.log)
* [unknown image] : Integration rootless fedora-35 w/ overlay
  * PR #4111
    * [07-14-2022 12:19](https://api.cirrus-ci.com/v1/task/5070459250147328/logs/integration_test.log)
* [unknown image] : Integration rootless ubuntu-2204 w/ overlay
  * PR #4111
    * [07-14-2022 12:07](https://api.cirrus-ci.com/v1/task/6196359156989952/logs/integration_test.log)

### [Containerized] 295 bud --pull=false --arch test
* [unknown image] : Containerized Integration
  * PR #4111
    * [07-14-2022 12:04](https://api.cirrus-ci.com/v1/task/4788984273436672/logs/integration_test.log)

...and, in podman-land:

### [bud] 277 bud --pull=false --arch test
* fedora-36 : bud podman fedora-36 root host
  * PR #16050
    * [10-04 17:51](https://api.cirrus-ci.com/v1/artifact/task/4683925472673792/html/bud-podman-fedora-36-root-host.log.html#t--00277)
* fedora-36 : bud remote fedora-36 root host [remote]
  * PR #16050
[...]

PR #4377, opened 2022-10-26
Define and use a safe, reliable test image

edsantiago:
The ubi8 image changed some architecture string yesterday and
broke all our CI on all branches, including podman because
it too runs bud.bats tests.

Solution: use a safe, reliable, trustworthy test image (the
one built and used for podman). Because that image does not
have the same labels, confirm pull using .Architecture
instead.

We should stop using ubi8 and registry.redhat (#4318). They
are unreliable. This PR doesn't fix that; it's just an
emergency patch for one specific CI break. We can use
this as a basis for future removals of ubi8.

```release-note
None
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -4227,24 +4227,24 @@ _EOF
   mytmpdir=${TEST_SCRATCH_DIR}/my-dir
   mkdir -p ${mytmpdir}
 cat > $mytmpdir/Containerfile << _EOF
-FROM registry.access.redhat.com/ubi8-minimal
+FROM $SAFEIMAGE
 _EOF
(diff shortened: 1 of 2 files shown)


edsantiago:
@containers/buildah-maintainers PTAL. This will need to be backported to every live branch.

This is not the smallest-possible change, but I believe it's the most futureproof.

rhatdan:
/approve
LGTM
@vrothberg @giuseppe @TomSweeneyRedHat @cevich @nalind @flouthoc PTAL

flouthoc (approved):
LGTM
/lgtm
/approve

PR #4379, opened 2022-10-26
[1.26] Define and use a safe, reliable test image

edsantiago:
The ubi8 image changed some architecture string yesterday and broke all our CI on all branches, including podman because it too runs bud.bats tests.

Solution: use a safe, reliable, trustworthy test image (the one built and used for podman). Because that image does not have the same labels, confirm pull using .Architecture instead.

We should stop using ubi8 and registry.redhat (#4318). They are unreliable. This PR doesn't fix that; it's just an emergency patch for one specific CI break. We can use this as a basis for future removals of ubi8.

```release-note
None
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -3488,24 +3488,24 @@ _EOF
   mytmpdir=${TEST_SCRATCH_DIR}/my-dir
   mkdir -p ${mytmpdir}
 cat > $mytmpdir/Containerfile << _EOF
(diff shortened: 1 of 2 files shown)


vrothberg (commented):
LGTM
@nalind @flouthoc @rhatdan PTAL

edsantiago:
Is this branch even active? There's no nightly cron job for it, and it's failing to build.

PR #4383, opened 2022-10-27
bud.bats: fix expected aarch

lsm5:
buildah inspect prints out arm image architecture as `aarch64` while expected architecture was set to `arm64`. This was causing gating test failures on Fedora bodhi.

Ref: https://artifacts.dev.testing-farm.io/6d9c4f78-7ec3-4911-bea0-77395b8b19f3/

@flouthoc @nalind @edsantiago @TomSweeneyRedHat PTAL

#### What type of PR is this?

/kind failing-test 

Fixes a buildah inspect test for aarch64 images.

#### How to verify it

Test should pass.

None

```release-note
None
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -4238,13 +4238,13 @@ _EOF
   run_buildah tag image-amd localhost/${SAFEIMAGE_NAME}:${SAFEIMAGE_TAG}
   run_buildah build --pull=false -q --arch=arm64 -t image-arm $WITH_POLICY_JSON ${mytmpdir}
   run_buildah inspect --format '{{ .OCIv1.Architecture }}' image-arm
-  expect_output arm64
+  expect_output aarch64
 
   run_buildah inspect --format '{{ .FromImageID }}' image-arm
   fromiid=$output
 
   run_buildah inspect --format '{{ .OCIv1.Architecture  }}'  $fromiid
-  expect_output arm64
+  expect_output aarch64
 }
 
 @test "bud --file with directory" {


tests/bud.bats
```diff
@@ -4238,13 +4238,13 @@ _EOF
   run_buildah tag image-amd localhost/${SAFEIMAGE_NAME}:${SAFEIMAGE_TAG}
   run_buildah build --pull=false -q --arch=arm64 -t image-arm $WITH_POLICY_JSON ${mytmpdir}
```
lsm5, line 4239: not sure if this also needs to be changed to aarch64. I'm hoping gating will point that out?

flouthoc:
Oh but just wondering how does it passes in buildah CI then. Checking this.

flouthoc:
I've built buildah from `main` and on my local the test is passing without any changes, Checking why this is showing unexpected output on gaiting tests. I think buildah CI will fail with this change if i am correct.

lsm5:
fwiw, this seems to have also passed on F36 https://artifacts.dev.testing-farm.io/1e51e396-ff17-4263-91bb-a52d678a7903/ . Wonder if something changed on rawhide. Will take a while to get results on Fedora 37

edsantiago:
You're fixing the fix: Everything broke, #4377 fixed it.

lsm5:
> You're fixing the fix: Everything broke, #4377 fixed it.

lol alright. I'll build buildah with that patch, thanks

PR #4386, opened 2022-10-27
test: fix prune logic for cache-from after adding content summary

flouthoc:
Don't use `run_buildah rmi --all -f` since on podman-remote test this will remove prefetched alpine and it will try to pull alpine from docker.io with completely different digest (ruining our cache logic) need for this test.

Fixes: https://github.com/containers/buildah/pull/4326#issuecomment-1291707569

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -4470,7 +4470,13 @@ _EOF
   # clean all cache and intermediate images
   # to make sure that we are only using cache
   # from remote repo and not the local storage.
-  run_buildah rmi --all -f
+
+  # Important side-note: don't use `run_buildah rmi --all -f`
+  # since on podman-remote test this will remove prefetched alpine
+  # and it will try to pull alpine from docker.io with
+  # completely different digest (ruining our cache logic).
+  run_buildah rmi --prune
+  run_buildah rmi test
 
   # ------ Test case ------ #
   # expect cache to be pushed on remote stream


flouthoc:
@edsantiago PTAL

tests/bud.bats
```diff
@@ -4470,7 +4470,13 @@ _EOF
...
   # to make sure that we are only using cache
   # from remote repo and not the local storage.
-  run_buildah rmi --all -f
+
+  # Important side-note: don't use `run_buildah rmi --all -f`
+  # since on podman-remote test this will remove prefetched alpine
+  # and it will try to pull alpine from docker.io with
+  # completely different digest (ruining our cache logic).
+  run_buildah rmi --prune
+  run_buildah rmi test
```
edsantiago, line 4479: Manually applying this on my treadmill PR makes the test pass under podman and podman-remote. LGTM. Nice work.

edsantiago (approved):

flouthoc:
@rhatdan @nalind PTAL

tests/bud.bats
nalind: Which images does the `rmi --prune` remove?

flouthoc: > Which images does the `rmi --prune` remove?

@nalind `rmi --prune` removes intermediate images from storage, which i need to remove in order to test remote repo's cache.

nalind: That's its purpose, but does it actually do anything here?

flouthoc: Apologies for later reply I missed the notification. My intention was to clean intermediate images generated by following line https://github.com/containers/buildah/blob/main/tests/bud.bats#L4470 ( idea was to make sure no cached image is present and all images are pulled from upstream ) but since we already have a line `Cache pulled from remote` so it does not matter.

PR #4380, opened 2022-10-26
Fix RUN --mount=type=bind,from=<stage> not triggering preservation of <stage> rootfs

marwatk:
The lack of the preservation was probably an oversight when --mount was added for RUN.

I added a test that fails without the modification and succeeds with.

Fixes #4375

Signed-off-by: Marcus Watkins <mwatkins@mitre.org>

#### What type of PR is this?

> /kind bug

#### What this PR does / why we need it:

Fixes #4375 

#### How to verify it

Run new test with and without change

#### Which issue(s) this PR fixes:

Fixes #4375

#### Special notes for your reviewer:

None

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #4387, opened 2022-10-27
test: fix preserve rootfs with `--mount=type=bind` for podman-remote

flouthoc:
`--quite` on `podman-remote` will output only image id suppressing the build output which is needed by the test, following commit makes test functional for `podman-remote`.

Fixes: https://github.com/containers/buildah/pull/4380#issuecomment-1293848854

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -3844,7 +3844,7 @@ EOM
 
 @test "bud preserve rootfs for --mount=type=bind,from=" {
   _prefetch alpine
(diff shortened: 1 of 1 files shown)


flouthoc:
@edsantiago PTAL

edsantiago (approved):
nice catch. Yep, manually patching my treadmill makes the test pass. Thank you!

flouthoc:
@nalind @rhatdan PTAL

nalind:
Commit log typos `--quiet`, otherwise LGTM.

PR #4390, opened 2022-10-28
copier: Preserve file flags when copying archives on FreeBSD

dfr:
This approach should also work on darwin but I have only tested on FreeBSD.

[NO NEW TESTS NEEDED]

#### What type of PR is this?

/kind other

This preserves file flags on FreeBSD when copying archives, e.g. with 'buildah add'

#### How to verify it

Make a tar file containing flag changes and add it to a buildah container.

None

I haven't included tests here since I haven't ported the buildah integration tests yet

```release-note
None
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/containers/buildah/util"
 	"github.com/containers/image/v5/pkg/compression"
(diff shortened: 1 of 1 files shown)


nalind:
Looks correct.

PR #4392, opened 2022-11-01
userns: add arbitrary steps/stage to `--userns=auto` test

flouthoc:
Support for build with `--userns=auto` was added here https://github.com/containers/buildah/pull/4064 but it never tested a build where multiple ephemeral containers were used. So just add arbitrary steps and stages to the exisitng `--userns=auto` test to make sure build still works fine.


flouthoc:
@rhatdan @nalind @giuseppe PTAL just extending test since something for mult-stage and step with `--userns=auto` was reported to me yesterday.

giuseppe (approved):
/lgtm

PR #4397, opened 2022-11-03
Stop using ubi8

edsantiago:
First: because we have a bunch of tests that rely on the image
manifest list, and ubi8 is not under our control: we've already
seen where ubi8 gets changed without our knowledge, requiring
almost a week of scrambling to get things fixed again.

Second, the registry it lives on is unreliable.

```release-note
None
```

### DIFF
--- a/tests/bud/multiarch/Dockerfile.no-run
+++ b/tests/bud/multiarch/Dockerfile.no-run
@@ -3,5 +3,5 @@ FROM docker.io/library/alpine
 COPY Dockerfile.no-run /root/
 # A different base image that is known to be a manifest list, supporting a
 # different but partially-overlapping set of platforms.
-FROM registry.access.redhat.com/ubi8-micro
+FROM quay.io/libpod/testimage:20221018
 COPY --from=0 /root/Dockerfile.no-run /root/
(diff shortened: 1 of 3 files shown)


tests/bud/multiarch/Dockerfile.no-run
```diff
@@ -3,5 +3,5 @@ FROM docker.io/library/alpine
 COPY Dockerfile.no-run /root/
 # A different base image that is known to be a manifest list, supporting a
 # different but partially-overlapping set of platforms.
-FROM registry.access.redhat.com/ubi8-micro
+FROM quay.io/libpod/testimage:20221018
```
edsantiago, line 6: FIXME FIXME FIXME FIXME!!! 

I can't use `$SAFEIMAGE` here. No matter where I put `ARG SAFEIMAGE` (line 1 at top, or line 5 just above the `FROM`), buildah barfs with:
```
Error: resolving image name "$SAFEIMAGE": cannot parse input: "$SAFEIMAGE": invalid reference format
```
But it works perfectly fine in the other test. Help please.

vrothberg (commented):
LGTM

tests/bud/multiarch/Dockerfile.no-run
edsantiago: UPDATE: the culprit is `--all-platforms`. It looks like `--all-platforms` cannot be combined with `--build-args`. Sounds like a bug to me, should I file an issue? Or is this a yeah-deal-with-it thing?

rhatdan: @flouthoc PTAL

Issue #4399, opened 2022-11-03
--all-platforms clobbers --build-arg


edsantiago:
Using `--all-platforms` removes the ability to use `--build-arg`:
```console
$ printf "ARG foo\nFROM \$foo\n" | bin/buildah build --build-arg foo=quay.io/libpod/alpine --all-platforms -t foo -
Error: resolving image name "$foo": cannot parse input: "$foo": invalid reference format
```
When you fix this, please also fix this instance of evil hardcoding:
```diff
diff --git a/tests/bud/multiarch/Dockerfile.no-run b/tests/bud/multiarch/Dockerfile.no-run
index dbdac3e1e..ef89ae651 100644
--- a/tests/bud/multiarch/Dockerfile.no-run
+++ b/tests/bud/multiarch/Dockerfile.no-run
@@ -3,5 +3,6 @@ FROM docker.io/library/alpine                                                                                                                  
 COPY Dockerfile.no-run /root/
 # A different base image that is known to be a manifest list, supporting a
 # different but partially-overlapping set of platforms.
-FROM quay.io/libpod/testimage:20221018
+ARG SAFEIMAGE
+FROM $SAFEIMAGE
 COPY --from=0 /root/Dockerfile.no-run /root/
```

rhatdan:
@edsantiago I just fixed this and accidentally force pushed it to main branch.

edsantiago:
Oh! Now I get it, thanks. That was a weird email notification ("closed in SHA", but no associated PR).

Your fix doesn't look quite right. Now when I use the `printf` above I get:
```
Error: resolving image name "": cannot parse input: "": repository name must have at least one component
```

flouthoc:
@rhatdan You accidentally pushed to main without PR

flouthoc:
@rhatdan We might wanna revert it since not all tests are passing with it and main is broken.

flouthoc:
`main` is broken and commit https://github.com/containers/buildah/commit/217b2d524c041674066f2e0d329ec2ad5a62004b also assumes that `ENV` must be evaluated while actually `ENV` must be only evaluated inside `RUN` statement and not for any `Containerfile` instructions.

I have created a PR https://github.com/containers/buildah/pull/4401 to fix this and it should fix `main` as well.

PR #4401, opened 2022-11-04
imagebuildah: build with `--all-platforms` must honor `userargs` and `headingargs` while evaluating base images

flouthoc:
Commit:
https://github.com/containers/buildah/commit/217b2d524c041674066f2e0d329ec2ad5a62004b was pushed to `main` accidently while it incorrectly assumes that `ENV` must be used while processing base image, however `ENV` should be only evaluated inside `RUN` instruction and not for any containerfile instruction such as `FROM <>`. This fixes change made in above commit.

Following commit ensures that imagebuildah considers `HeadingArgs` and `UserArgs` when its evaluating base image to pull all images from manifest list early when build is invoked with `--all-platforms`

Closes: https://github.com/containers/buildah/issues/4399

```release-note
build: base image honors userargs and heading args when built with --all-platforms
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -5143,6 +5143,21 @@ _EOF
   assert "$manifests" = "amd64 arm64 ppc64le s390x" "arch list in manifest"
 }
 
+# attempts to resolve heading arg as base-image with --all-platforms
+@test "bud-multiple-platform-with-base-as-default-arg" {
+  outputlist=localhost/testlist
+  run_buildah build $WITH_POLICY_JSON \
+              --jobs=1 \
+              --all-platforms \
+              --manifest $outputlist \
+              -f $BUDFILES/all-platform/Containerfile.default-arg \
+              $BUDFILES/all-platform
+
+  run_buildah manifest inspect $outputlist
+  manifests=$(jq -r '.manifests[].platform.architecture' <<<"$output" |sort|fmt)
+  assert "$manifests" = "386 amd64 arm arm arm64 ppc64le s390x" "arch list in manifest"
+}
+
 @test "bud-multiple-platform for --all-platform with additional-build-context" {
   outputlist=localhost/testlist
   local contextdir=${TEST_SCRATCH_DIR}/bud/platform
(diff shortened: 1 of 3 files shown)


flouthoc:
`main` is broken right now, @rhatdan @edsantiago @nalind PTAL

rhatdan:
@giuseppe @vrothberg PTAL

tests/bud/all-platform/Containerfile.default-arg
```diff
@@ -0,0 +1,2 @@
+ARG foo=alpine
+FROM $foo
```
vrothberg, line 2: Does that really add a new test?  I see many tests using `ARG SAFEIMAGE` already.

flouthoc: all of them do not have any default value set for `ARG` instead they all consume value from cli using `--build-arg`, this test verifies additional heading arg i.e default value and it was not supported before this PR.

flouthoc: Line: https://github.com/containers/buildah/pull/4401/files#diff-d96798032784de488b331cd60b9420f3263b46da4e7b47f8968569a0e6f629beR699 added support for this

tests/bud.bats
```diff
@@ -5143,6 +5143,21 @@ _EOF
   assert "$manifests" = "amd64 arm64 ppc64le s390x" "arch list in manifest"
 }
 
+# attempts to resolve heading arg as base-image with --all-platforms
+@test "bud-multiple-platform-with-base-as-default-arg" {
```
edsantiago, line 5147: Not needed. I have no time to go into it, but main is broken because of an accidental push, and the test was added there. Do a git log or ask Dan for details. The test is already there. Thanks for fixing this.

flouthoc: @edsantiago I think this is a different test for a different feature which is not already being tested in upstream, see: https://github.com/containers/buildah/pull/4401#discussion_r1013865451

rhatdan:
I am going to merge this and we can debate whether this is a new test on Monday.

PR #4406, opened 2022-11-06
Use TypeBind consistently to name bind/nullfs mounts

dfr:
This allows declaring run mounts using e.g. '-mount=type=nullfs,...' on FreeBSD which makes more sense for FreeBSD users. It is also consistent with 'podman run' which requires the nullfs mount type on FreeBSD.

[NO NEW TESTS NEEDED]

#### What type of PR is this?

/kind cleanup

This allows using the native filesystem type in run mounts rather than expecting the user to know that 'bind' is translated to 'nullfs' internally.

#### How to verify it

On a FreeBSD system:
```
buildah run --mount=type=nullfs,src=/tmp,dst=/foo <container> <cmd>
```

None

```release-note
None
```

### DIFF
--- a/define/mount_unsupported.go
+++ b/define/mount_unsupported.go
@@ -1,0 +1,17 @@
+//go:build darwin || windows
+// +build darwin windows
+
+package define
+
+const (
+	// TypeBind is the type for mounting host dir
+	TypeBind = "bind"
+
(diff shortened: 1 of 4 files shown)


dfr:
Rebased and fixed the cross build break.

PR #4411, opened 2022-11-09
copier.Put(): clear up os/syscall mode bit confusion

nalind:
#### What type of PR is this?

/kind bug

When noting that a non-symlink has setuid/setgid/sticky bits, switch from using `syscall` package bits and `syscall.Chmod()` to using `os` package bits and `os.Chmod()`, and if the item's a directory, record the updated mode information in the `directoryModes` map that we'll use to reset its permissions later.

#### How to verify it

Expanded conformance test!

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2138434.

```release-note
None
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -1571,15 +1571,15 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
 		return nil
 	}
 	makeDirectoryWriteable := func(directory string) error {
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM

flouthoc (approved):
LGTM
/lgtm
/approve

Issue #4404, opened 2022-11-04
Two users cannot build Dockerfiles with the same --mount=type=cache folder


marwatk:
**Description**

buildah [uses a singular cache folder](https://github.com/marwatk/buildah/blob/835668715705a7ebbdf32d945b4baae545d98aa2/internal/parse/parse.go#L335) for all users. As a result the second user's build of an image referencing the same cache folder will fail.

We routinely cache `/var/cache` to prevent `dnf` files from surviving into the final image (and to speed up builds). We encountered it on a CI server running builds as separate users.

Using this dockerfile:
```Dockerfile
FROM alpine

RUN --mount=type=cache,target=/multiuser_cache_test \
    set -ex; \
    ls -ln /multiuser_cache_test/cache_file || true; \
    echo cache_content > /multiuser_cache_test/cache_file; \
    chmod 600 /multiuser_cache_test/cache_file; \
    ls -ln /multiuser_cache_test/cache_file || true; \
    : ;
```

**Steps to reproduce the issue:**
1. Wipe `${TMPDIR}/buildah-cache` (not strictly necessary, but shows zero state)
2. Rootfull build first: `sudo buildah build -f - < Dockerfile`
3. Rootless build: `buildah build -f - < Dockerfile`

**Describe the results you received:**

```
$ bin/buildah build -f - < tests/bud/cache-multiuser/Dockerfile
STEP 1/2: FROM alpine
STEP 2/2: RUN --mount=type=cache,target=/multiuser_cache_test     set -ex;     ls -ln /multiuser_cache_test/cache_file || true;     echo cache_content > /multiuser_cache_test/cache_file;     chmod 600 /multiuser_cache_test/cache_file;     ls -ln /multiuser_cache_test/cache_file || true;     : ;
Error: building at STEP "RUN --mount=type=cache,target=/multiuser_cache_test set -ex;     ls -ln /multiuser_cache_test/cache_file || true;     echo cache_content > /multiuser_cache_test/cache_file;     chmod 600 /multiuser_cache_test/cache_file;     ls -ln /multiuser_cache_test/cache_file || true;     : ;": resolving mountpoints for container "b03de9521b61a4fdada0a485589b00a8aa16b31d88f38aa4424301695dfe390b": unable to create build cache lockfiles directory: mkdir /var/tmp/buildah-cache/buildah-cache-lockfiles/multiuser_cache_test: permission denied
```

**Describe the results you expected:**

Build succeeds and doesn't try to utilize root's cache.

[...]

PR #4421, opened 2022-11-17
`mount=type=cache`: seperate cache parent on host for each user

flouthoc:
`mount=type=cache` creates a common cache directory on host in temporary directory, split this cache directory for each user invocation in order to prevent overlapping of cache content when `buildah` is invoked by different users on same host.

Closes: https://github.com/containers/buildah/issues/4404

### DIFF
--- a/internal/parse/parse.go
+++ b/internal/parse/parse.go
@@ -18,6 +18,7 @@ import (
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/lockfile"
+	"github.com/containers/storage/pkg/unshare"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	selinux "github.com/opencontainers/selinux/go-selinux"
 )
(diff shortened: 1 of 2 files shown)


internal/parse/parse.go
```diff
@@ -330,8 +331,8 @@ func GetCacheMount(args []string, store storage.Store, imageMountLabel string, a
 		// create a common cache directory, which persists on hosts within temp lifecycle
 		// add subdirectory if specified
 
-		// cache parent directory
-		cacheParent := filepath.Join(internalUtil.GetTempDir(), BuildahCacheDir)
+		// cache parent directory: creates eperate cache parent for each user.
```
lucastheisen, line 334:
```suggestion
		// cache parent directory: creates separate cache parent for each user.
```

internal/parse/parse.go
```diff
@@ -330,8 +331,8 @@ func GetCacheMount(args []string, store storage.Store, imageMountLabel string, a
...
+		cacheParent := filepath.Join(internalUtil.GetTempDir(), filepath.Join(BuildahCacheDir, strconv.Itoa(unshare.GetRootlessUID())))
```
marwatk, line 335: The problem with this is the first user to do it will still own `BuildahCacheDir` and the next user won't be able to create their subfolder under it. This should work, though:

```suggestion
		cacheParent := filepath.Join(internalUtil.GetTempDir(), BuildahCacheDir+"-"+strconv.Itoa(unshare.GetRootlessUID()))
```

I'll see if I can get it tested today.

marwatk:
Just tested and can confirm this fixes the issue. Thanks!

TomSweeneyRedHat:
LGTM
Time to get the WIP label off of this one?

flouthoc:
@TomSweeneyRedHat Done. PTAL again. @containers/buildah-maintainers PTAL

PR #4428, opened 2022-11-19
[release-1.28] copier.Put(): clear up os/syscall mode bit confusion  and bump to `v1.28.1`

flouthoc:
Backport https://github.com/containers/buildah/pull/4411 for podman `v4.3.2` as discussed here: https://github.com/containers/buildah/issues/4427

PR #4433, opened 2022-11-23
[release-1.28] Define and use a safe, reliable test image and bump to `v1.28.2`

flouthoc:
I thought older `buildah` and `podman` `v4.3` branch does not accounts for `SAFEIMAGE` hence I created a simpler retrofit commit here https://github.com/containers/buildah/pull/4428 ( **I was wrong** ) and it looks like podman's old branch were already expecting https://github.com/containers/buildah/pull/4377.

A note from @edsantiago  in podman's `v4.3` branch https://github.com/containers/podman/commit/52a6516f96d2f99db6136d87e444675c051f4b48 clearly mentions that https://github.com/containers/buildah/pull/4377 had to be backported.

Since I missed backporting it before so doing it now and cutting a release.

PR which will use this release: https://github.com/containers/podman/pull/16588

### DIFF
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -13,6 +13,13 @@ OCI=$(${BUILDAH_BINARY} info --format '{{.host.OCIRuntime}}' || command -v runc 
 # Default timeout for a buildah command.
 BUILDAH_TIMEOUT=${BUILDAH_TIMEOUT:-300}
 
+# Safe reliable unchanging test image
+SAFEIMAGE_REGISTRY=${SAFEIMAGE_REGISTRY:-quay.io}
+SAFEIMAGE_USER=${SAFEIMAGE_USER:-libpod}
+SAFEIMAGE_NAME=${SAFEIMAGE_NAME:-testimage}
+SAFEIMAGE_TAG=${SAFEIMAGE_TAG:-20221018}
+SAFEIMAGE="${SAFEIMAGE:-$SAFEIMAGE_REGISTRY/$SAFEIMAGE_USER/$SAFEIMAGE_NAME:$SAFEIMAGE_TAG}"
+
 # Shortcut for directory containing Containerfiles for bud.bats
 BUDFILES=${TEST_SOURCES}/bud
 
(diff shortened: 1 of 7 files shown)


flouthoc:
@edsantiago Could you please take a look

tests/helpers.bash
```diff
@@ -13,6 +13,19 @@ OCI=$(${BUILDAH_BINARY} info --format '{{.host.OCIRuntime}}' || command -v runc
...
+SAFEIMAGE_NAME=${SAFEIMAGE_NAME:-testimage}
+SAFEIMAGE_TAG=${SAFEIMAGE_TAG:-20221018}
+SAFEIMAGE="${SAFEIMAGE:-$SAFEIMAGE_REGISTRY/$SAFEIMAGE_USER/$SAFEIMAGE_NAME:$SAFEIMAGE_TAG}"
+
+# Prompt to display when logging buildah commands; distinguish root/rootless
+_LOG_PROMPT='$'
+if [ $(id -u) -eq 0 ]; then
+    _LOG_PROMPT='#'
+fi
+
```
edsantiago, line 28: This is misleading, because it's not actually used anywhere (it's part of #4320,  but only an incomplete part). I have a strong aversion to unused code, because it can cost time for someone in the future. Would you mind removing it and re-pushing? Or adding the other missing part of #4320?
```diff
-        echo "\$ $BUILDAH_BINARY $*"
+        echo "${_LOG_PROMPT} $BUILDAH_BINARY $*"
```

edsantiago (commented):
Patch LGTM, I diffed against `main` to make sure. One cleanup request.

Is this expected to be an active branch? If so, would it make sense to backport #4397 also, to get rid of other ubi-related risks?

tests/helpers.bash
flouthoc: I will remove this chunk

flouthoc:
> Is this expected to be an active branch? If so, would it make sense to backport https://github.com/containers/buildah/pull/4397 also, to get rid of other ubi-related risks?

@edsantiago Yes this is active branch. I will also backport #4397 to this PR.

edsantiago (approved):
LGTM, thank you

PR #4439, opened 2022-11-29
[CI:BUILD] copr: fix el8 builds

lsm5:

#### What type of PR is this?

/kind other

Fixes copr build

#### How to verify it

successful el8 build at: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/5072942/

None

None

```release-note
None
```

@flouthoc @nalind @rhatdan PTAL

### DIFF
--- a/buildah.spec.rpkg
+++ b/buildah.spec.rpkg
@@ -11,15 +11,17 @@
 
 %global with_debug 1
 
+# RHEL 8's default %%gobuild macro doesn't account for the BUILDTAGS variable, so we
+# set it separately here and do not depend on RHEL 8's go-srpm-macros package.
+%if !0%{?fedora} && 0%{?rhel} <= 8
+%define gobuild(o:) GO111MODULE=off go build -buildmode pie -compiler gc -tags="rpm_crashtraceback libtrust_openssl ${BUILDTAGS:-}" -ldflags "-linkmode=external -compressdwarf=false ${LDFLAGS:-} -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '%__global_ldflags'" -a -v -x %{?**};
+%endif
+
 %if 0%{?with_debug}
 %global _find_debuginfo_dwz_opts %{nil}
 %global _dwz_low_mem_die_limit 0
 %else
 %global debug_package %{nil}
-%endif
-
-%if ! 0%{?gobuild:1}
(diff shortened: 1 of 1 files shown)


lsm5:
older builds were failing for el8: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/5072893/

rhatdan:
@lsm5 you still working on this?

lsm5:
> @lsm5 you still working on this?

I'll resume this, been working on a similar one for skopeo.

vrothberg:
> Why doesn't the next `%if 0%{?rhel}` block ~not~ accomplish the same thing? (Aside: the use of single quotes in that next block appears to set BUILDTAGS to a value that literally includes "$BUILDTAGS", rather than expanding it.) From the build log, it looks like `$BUILDTAGS` is either empty or not being referenced when `go build` is run, which looks like a bigger problem than the current patch would work around.

This looks still open

lsm5:
err yes, I still gotta work on this :) . Let me try to get skopeo out of the way first. If we prefer closing this and reopening when I'm ready to update, fine by me.

lsm5:
RHEL8's default %gobuild didn't account for $BUILDTAGS. Adjusted that in buildah.spec.rpkg . Successful build at https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/5390214/ .

lsm5:
let me fix debuginfo in this PR itself.

lsm5:
ready for review. successful build at https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/5406349/

flouthoc (approved):
LGTM

lsm5:
i'll update this PR further with making the spec as close to podman and buildah wherever possible.

buildah.spec.rpkg
```diff
@@ -11,17 +11,19 @@
 
 %global with_debug 1
 
+# RHEL 8's default %%gobuild macro doesn't account for the BUILDTAGS variable, so we
```
TomSweeneyRedHat, line 14: Soft suggestion:
```suggestion
# Starting with RHEL 8, the default %%gobuild macro doesn't account for the BUILDTAGS variable, so we
```

buildah.spec.rpkg
```diff
@@ -11,17 +11,19 @@
...
+# set it separately here and do not depend on RHEL 8's go-srpm-macros package.
```
TomSweeneyRedHat, line 15: If you take the prior
```suggestion
# set it separately here and do not depend on RHEL's go-srpm-macros package.
```

lsm5: @TomSweeneyRedHat nope, this is specifically for rhel8. rhel 9 is fine.

buildah.spec.rpkg
lsm5: Not quite `starting with rhel8`. But only rhel8.

lsm5:
good to go. PTAL

giuseppe (approved):
LGTM

TomSweeneyRedHat:
LGTM

PR #4408, opened 2022-11-08
Update to F37 CI VM Images, re-enable prior-fedora

cevich:
#### What type of PR is this?

/kind other

#### What this PR does / why we need it:

Update to F37 CI VM Images, re-enable prior-fedora

#### How to verify it

CI will pass

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

None

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #4441, opened 2022-11-30
util.SortMounts(): make the returned order more stable

nalind:
#### What type of PR is this?

/kind bug

Use sort.Stable() instead of sort.Sort() to sort mounts, and have the comparison function compare the cleaned paths directly if they have the same number of components, so that there's a defined ordering between "/a" and "/b".

#### How to verify it

Updated the unit test!

Related: #4408.

```release-note
None
```

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -441,7 +441,14 @@ func (m byDestination) Len() int {
 }
 
 func (m byDestination) Less(i, j int) bool {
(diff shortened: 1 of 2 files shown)


vrothberg (commented):
LGTM

flouthoc (approved):
LGTM
/lgtm

PR #4443, opened 2022-12-05
remote-cache: support multiple sources and destination for `--cache-to` and `--cache-from`

flouthoc:
Buildah must support multiple sources for remote cache when using `--cache-to` and `--cache-from` so users can distribute and collect cache from various sources.

We also need for compat with buildkit api which in past broke `podman` compat build API see: https://github.com/containers/podman/pull/16380

More discussion here: https://github.com/containers/podman/pull/16380

```release-note
remote-cache: support multiple sources and destination for `--cache-to` and `--cache-from`
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -1729,42 +1729,48 @@ func (s *StageExecutor) generateCacheKey(ctx context.Context, currNode *parser.N
 
 // cacheImageReference is internal function which generates ImageReference from Named repo sources
 // and a tag.
-func cacheImageReference(repo reference.Named, cachekey string) (types.ImageReference, error) {
-	tagged, err := reference.WithTag(repo, cachekey)
-	if err != nil {
-		return nil, fmt.Errorf("failed generating tagged reference for %q: %w", repo, err)
+func cacheImageReferences(repos []reference.Named, cachekey string) ([]types.ImageReference, error) {
+	var result []types.ImageReference
+	for _, repo := range repos {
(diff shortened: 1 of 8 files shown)


flouthoc:

imagebuildah/stage_executor.go
```diff
@@ -1729,42 +1729,48 @@ func (s *StageExecutor) generateCacheKey(ctx context.Context, currNode *parser.N
...
-	if err != nil {
-		return nil, fmt.Errorf("failed generating tagged reference for %q: %w", repo, err)
-	}
-	dest, err := imagedocker.NewReference(tagged)
-	if err != nil {
-		return nil, fmt.Errorf("failed generating docker reference for %q: %w", tagged, err)
+func cacheImageReferences(repos []reference.Named, cachekey string) ([]types.ImageReference, error) {
+	var result []types.ImageReference
+	for _, repo := range repos {
+		tagged, err := reference.WithTag(repo, cachekey)
```
rhatdan, line 1735: Wouldn't some of these be expected to fai, since the cachekey would only be in one of the repos?

flouthoc: Yes, expected to fail in some of the repos. `--cache-to` and `--cache-from` will accept multiple repos. 

In case of `--cache-from` it will return if the key matched in first repo if not it will try for the other repo's in the list.

In case of `--cache-to` it will push the cache layers to all the repos in the list.

giuseppe (approved):
/lgtm

PR #4452, opened 2022-12-07
build, mount: allow realtive mountpoints wrt to work dir for `--mount=type=cache` and `--mount=type=bind`

flouthoc:
When working with `--mount=type=bind` and `--mount=type=cache` allow `target` to accept relative paths w.r.t to the configured work dir.

Closes: https://github.com/containers/buildah/issues/4309

### DIFF
--- a/internal/parse/parse.go
+++ b/internal/parse/parse.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -49,7 +50,7 @@ var (
 // GetBindMount parses a single bind mount entry from the --mount flag.
 // Returns specifiedMount and a string which contains name of image that we mounted otherwise its empty.
 // Caller is expected to perform unmount of any mounted images
-func GetBindMount(ctx *types.SystemContext, args []string, contextDir string, store storage.Store, imageMountLabel string, additionalMountPoints map[string]internal.StageMountDetails) (specs.Mount, string, error) {
+func GetBindMount(ctx *types.SystemContext, args []string, contextDir string, store storage.Store, imageMountLabel string, additionalMountPoints map[string]internal.StageMountDetails, workDir string) (specs.Mount, string, error) {
 	newMount := specs.Mount{
 		Type: define.TypeBind,
 	}
@@ -101,10 +102,14 @@ func GetBindMount(ctx *types.SystemContext, args []string, contextDir string, st
 			if len(kv) == 1 {
 				return newMount, "", fmt.Errorf("%v: %w", kv[0], errBadOptionArg)
 			}
-			if err := parse.ValidateVolumeCtrDir(kv[1]); err != nil {
+			targetPath := kv[1]
+			if !path.IsAbs(targetPath) {
+				targetPath = filepath.Join(workDir, targetPath)
+			}
+			if err := parse.ValidateVolumeCtrDir(targetPath); err != nil {
 				return newMount, "", err
 			}
-			newMount.Destination = kv[1]
+			newMount.Destination = targetPath
 			setDest = true
 		case "consistency":
 			// Option for OS X only, has no meaning on other platforms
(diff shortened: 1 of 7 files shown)


internal/parse/parse.go
```diff
@@ -101,10 +102,14 @@ func GetBindMount(ctx *types.SystemContext, args []string, contextDir string, st
 			if len(kv) == 1 {
 				return newMount, "", fmt.Errorf("%v: %w", kv[0], errBadOptionArg)
 			}
-			if err := parse.ValidateVolumeCtrDir(kv[1]); err != nil {
+			targetPath := kv[1]
+			if !path.IsAbs(targetPath) {
+				targetPath = filepath.Join(workDir, targetPath)
```
rhatdan, line 107: Should this path be resolved?

filepath.EvalSymlinks
Or at least
filepath.Clean?

internal/parse/parse.go
```diff
@@ -257,10 +262,14 @@ func GetCacheMount(args []string, store storage.Store, imageMountLabel string, a
 			if len(kv) == 1 {
 				return newMount, nil, fmt.Errorf("%v: %w", kv[0], errBadOptionArg)
 			}
-			if err := parse.ValidateVolumeCtrDir(kv[1]); err != nil {
+			targetPath := kv[1]
+			if !path.IsAbs(targetPath) {
+				targetPath = filepath.Join(workDir, targetPath)
```
rhatdan, line 267: Same question as above?

internal/parse/parse.go
flouthoc: Yes I think target path must be cleaned before join, since things is executed on host side symlink should still be allowed I think.

internal/parse/parse.go
```diff
@@ -101,10 +102,14 @@ func GetBindMount(ctx *types.SystemContext, args []string, contextDir string, st
 			if len(kv) == 1 {
 				return newMount, "", fmt.Errorf("%v: %w", kv[0], errBadOptionArg)
 			}
-			if err := parse.ValidateVolumeCtrDir(kv[1]); err != nil {
+			targetPath := kv[1]
+			if !path.IsAbs(targetPath) {
+				targetPath = filepath.Join(workDir, filepath.Clean(targetPath))
```
rhatdan, line 107: I was thinking

				targetPath = filepath.Clean(filepath.Join(workDir, targetPath))

internal/parse/parse.go
```diff
@@ -257,10 +262,14 @@ func GetCacheMount(args []string, store storage.Store, imageMountLabel string, a
 			if len(kv) == 1 {
 				return newMount, nil, fmt.Errorf("%v: %w", kv[0], errBadOptionArg)
 			}
-			if err := parse.ValidateVolumeCtrDir(kv[1]); err != nil {
+			targetPath := kv[1]
+			if !path.IsAbs(targetPath) {
+				targetPath = filepath.Join(workDir, filepath.Join(targetPath))
```
rhatdan, line 267: I was thinking:
			targetPath = filepath.Clean(filepath.Join(workDir, targetPath))

internal/parse/parse.go
flouthoc: In otherplaces in buildah we first clean the user input and then join it with a predefined path. I am just wondering if there any edge-case we miss cleaning the path later.

flouthoc: @rhatdan I was looking for edge-case so I think as @nalind suggested and doc say I think `filepath.Join(` already performs a `clean(` https://pkg.go.dev/path/filepath#Join

rhatdan: Ok, do we care about eval symlinks?

flouthoc: @rhatdan Not really since this expects to be always executed on host.

rhatdan:
Since filepath.Join automatically does a clean, there is no reason to add one to the code.  Remove the clean and this 
LGTM

internal/parse/parse.go
TomSweeneyRedHat: I don't think we want to allow symlinks, as noted elsewhere, they're evil....

flouthoc: This part is only executed on host side so there is no room for remote execution of symlink hence they are allowed just like we allow it for all the other mounts already.

rhatdan:
@vrothberg @giuseppe @nalind @umohnani8 PTAL

flouthoc:

vrothberg (approved):
/lgtm

PR #2627, opened 2020-09-15
Follow the symlink when mounting volumes

zhaofengli:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

There are two issues that are fixed by this commit
- Use `os.Lstat` instead of `os.Stat` when checking the existence of the mountpoint. `os.Stat` will follow the symlink (in the host's root!) which is always wrong in our case. This also means if you happen to have the target directory in your host system, #2621 will not occur :stuck_out_tongue:
- If `srcPath` is a symlink, follow the symlink and update `srcPath` to its destination. This replicates Docker's behavior and fixes #2621.

#### How to verify it

Use the following Dockerfile

```
FROM alpine

# /targetdir may or may not exist. The existing code will handle it correctly.
RUN ln -s /targetdir /data
VOLUME /data

RUN cat /proc/mounts
```

#### Which issue(s) this PR fixes:

Fixes #2621

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #4468, opened 2022-12-15
parse: default `.containerignore` and `.dockerignore` must not point to symlink outside context

flouthoc:
Drop support for use-case when .containerignore or .dockerignore is a symlink pointing to arbitrary location on host.

Buildah implementation of: https://github.com/containers/podman/pull/16315

[NO NEW TESTS NEEDED]
[NO TESTS NEEDED]

```release-note
build: drop support for use-case when .containerignore or .dockerignore is a symlink pointing to arbitrary location on host.
```

### DIFF
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/containerd/containerd/platforms"
 	"github.com/containers/buildah/define"
+	securejoin "github.com/cyphar/filepath-securejoin"
 	internalParse "github.com/containers/buildah/internal/parse"
 	"github.com/containers/buildah/pkg/sshagent"
 	"github.com/containers/common/pkg/config"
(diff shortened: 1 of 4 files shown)


flouthoc:
need to add a test

pkg/parse/parse.go
```diff
@@ -1107,10 +1108,16 @@ func ContainerIgnoreFile(contextDir, path string, containerFiles []string) ([]st
...
 	}
-	path = filepath.Join(contextDir, ".containerignore")
+	path, symlinkErr := securejoin.SecureJoin(contextDir, ".containerignore")
+	if symlinkErr != nil {
+		return nil, "", symlinkErr
+	}
 	excludes, err := imagebuilder.ParseIgnore(path)
 	if errors.Is(err, os.ErrNotExist) {
-		path = filepath.Join(contextDir, ".dockerignore")
+		path, symlinkErr = securejoin.SecureJoin(contextDir, ".dockerignore")
```
TomSweeneyRedHat, line 1117: @nalind had objections to securejoin in this commit:  https://github.com/containers/buildah/pull/2627, I'm not sure if they still hold or what better alternatives there are now.  

I'll just go on the record noting that symlinks are utter evil!

flouthoc: @TomSweeneyRedHat This is actually to prevent usage of symlink when its pointing outside the build-context ( so yes it assumes that symlink are bad and resolves it early and ignores if it points to a ignore file outside of build context )

rhatdan:
LGTM
@giuseppe @vrothberg @nalind @umohnani8 PTAL

umohnani8:
LGTM

Issue #4486, opened 2022-12-29
Add command to clear build cache, and especially cache mounts


Clockwork-Muse:
**Description**

Currently, there does not appear to be a command to clear build cache via buildah (nothing obvious presents itself with `buildah --help`, anyways).  

Additionally, even if build cache is cleared via `podman system prune`, build-time cache mounts are not removed.  
So far, the only workaround I've found is `podman system reset` which seems... rather much.  Attempting to "reset" the cache via modifying the `RUN ...` directive seems potentially terrifying, depending on the environment the offending cache lives, and could be troublesome in distributed CI scenarios

Build-time cache mounts also are not discarded/ignored in conjunction with `--cache-ttl`, but I'm not completely sure if that's intended behavior or not.

**Steps to reproduce the issue:**
1. Build the following containerfile:
```Dockerfile
FROM docker.io/ubuntu

ARG CACHE_DIR=/some_cache_dir

RUN \
    #
    # Mounted cache directories
    --mount=type=cache,id=15022054c90073ed509dc66593e80ff4ae5dcd6c9967911024bc85ed64069b6e,target=${CACHE_DIR} \
    ls -al ${CACHE_DIR} \
    && touch ${CACHE_DIR}/some-cache-file \
    && ls -al ${CACHE_DIR}
```
2. "Attempt" to clear the build cache via `podman system prune -af --volumes`
3. Rebuild

**Describe the results you received:**

Despite clearing the build cache, build-time cache mounts retain data.

**Describe the results you expected:**

Clearing the build cache also clears build-time cache mounts.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
Listing... Done
buildah/unknown,now 1.28.2-0ubuntu22.04+obs36.6 amd64 [installed]
buildah/unknown 1.28.2-0ubuntu22.04+obs36.6 arm64
buildah/unknown 1.28.2-0ubuntu22.04+obs36.6 armhf
buildah/unknown 1.28.2-0ubuntu22.04+obs36.6 s390x
```

**Output of `buildah version`:**

```
Version:         1.28.2
Go Version:      go1.18.1
Image Spec:      1.0.2-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        1.0.0
libcni Version:  
image Version:   5.23.0
Git Commit:      
Built:           Wed Dec 31 16:00:00 1969
OS/Arch:         linux/amd64
BuildPlatform:   linux/amd64
```

[...]

PR #4490, opened 2023-01-02
buildah: add `buildah prune` command and expose `CleanCacheMount` API

flouthoc:
* Just like buildkit buildah must allow cleaning the buildcache and cache generated on host by `--mount=type=cache` just like buildkit's `prune` command.

* Also expose `CleanCacheMount` API so other tools like `podman` can use it.

See: https://github.com/moby/buildkit#cache
Closes: https://github.com/containers/buildah/issues/4486

```release-note
buildah: add `buildah prune` command and expose `CleanCacheMount` API
```

### DIFF
--- a/cmd/buildah/prune.go
+++ b/cmd/buildah/prune.go
@@ -1,0 +1,95 @@
+package main
+
+import (
+	"context"
+	"fmt"
+
+	internalParse "github.com/containers/buildah/internal/parse"
+	buildahcli "github.com/containers/buildah/pkg/cli"
+	"github.com/containers/buildah/pkg/parse"
+	"github.com/containers/common/libimage"
+	"github.com/hashicorp/go-multierror"
+	"github.com/spf13/cobra"
+)
+
+type pruneOptions struct {
+	force bool
+	all   bool
+}
+
+func init() {
+	var (
+		pruneDescription = `
+Cleanup intermediate images as well as build and mount cache.`
+		opts pruneOptions
+	)
+	pruneCommand := &cobra.Command{
+		Use:   "prune",
+		Short: "Cleanup intermediate images as well as build and mount cache",
+		Long:  pruneDescription,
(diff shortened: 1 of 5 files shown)


flouthoc:

rhatdan:
You should add a --all option to remove all unused images.

cmd/buildah/prune.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/spf13/cobra"
+)
+
+type pruneOptions struct {
+	force bool
+}
+
+func init() {
+	var (
+		pruneDescription = "\n  Clean up build cache, intermediate images and mount cache."
```
rhatdan, line 21: Renive tge \n and just put this on next line, like it is in other commands.

rhatdan: I would expect this to concentrate on images rather then cache

Perhaps
```
Cleanup intermediate images as well as build and mount cache
```

flouthoc: Done.

giuseppe (approved):
LGTM

cmd/buildah/prune.go
```diff
@@ -0,0 +1,95 @@
...
+	options := &libimage.RemoveImagesOptions{
+		Filters: []string{"readonly=false"},
+	}
+	if !iopts.all {
+		options.Filters = append(options.Filters, "dangling=true")
+		options.Filters = append(options.Filters, "intermediate=true")
+	}
+	options.Force = iopts.force
+
+	rmiReports, rmiErrors := runtime.RemoveImages(context.Background(), args, options)
```
rhatdan, line 80: Does this report an error if I specify --all and I have used images?

buildah from alpine
buildah prune --all

rhatdan: OK I believe this works fine from reading the man pages of buildah rmi.

Issue #4481, opened 2022-12-25
checking for build stage status 10 times/second via external runtime exec


paravz:
**Description**
Buildah checks if a build stage container is running every 100 msec per https://github.com/containers/buildah/blob/d87b736c39219009693c3dbc553ae786c0fb236a/run_common.go#L676

I noticed that invoking https://github.com/containers/buildah/blob/d87b736c39219009693c3dbc553ae786c0fb236a/run_common.go#L649 results in running `/usr/bin/crun --systemd-cgroup state buildah-buildahXXXX` 10 times a second for duration of the build step, a few minutes in my case

**Steps to reproduce the issue:**
1. Create a sample Docker file with "RUN sleep 300" as one of the steps
2. run podman build (or equivalent buildah command)
3. in a separate terminal run `sudo dnf install bcc-tools && sudo /usr/share/bcc/tools/execsnoop -TxU` to see runtime (ie crun) invocations, alternatively strace or debugger can be used

**Describe the results you received:**
Execution of `/usr/bin/crun --systemd-cgroup state buildah-buildahXXXX` 10 times a second for the duration of a Dockerfile build step

**Describe the results you expected:**
Exponential back off or similar solution to reduce overhead for longer jobs.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
# rpm -q buildah podman
package buildah is not installed
podman-4.3.1-1.fc37.x86_64
```

**Output of `buildah version`:**

```
#per https://github.com/containers/podman/blob/main/RELEASE_NOTES.md:
v1.28.0
```

**Output of `podman version` if reporting a `podman build` issue:**

```
# podman version
Client:       Podman Engine
Version:      4.3.1
API Version:  4.3.1
Go Version:   go1.19.2
Built:        Fri Nov 11 07:01:27 2022
OS/Arch:      linux/amd64
```

**Output of `cat /etc/*release`:**

```
Fedora release 37 (Thirty Seven)
```

**Output of `uname -a`:**

```
Linux 6.0.12-300.fc37.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Dec 8 16:58:47 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

**Output of `cat /etc/containers/storage.conf`:**

```
# cat /etc/containers/storage.conf|grep -vE "^\s*#|^\s*$"
[storage]
driver = "overlay"
runroot = "/run/containers/storage"
graphroot = "/var/lib/containers/storage"
[...]

PR #4492, opened 2023-01-03
run,performance: check build stage status in `truncated exponential backoff` manner

flouthoc:
Buildah constantly polls running build container for it status every `0.1` second, this polling nature wastes CPU cycles unnecessarily so instead of checking status every `0.1` second use exponentially increasing counter.

Exponential counter has a max limit of `2 seconds` and delay will reset to certain milliseconds and will grow again to `2 seconds` this process will be repeated till status is returned from the build container.

[NO NEW TESTS NEEDED]
[NO TESTS NEEDED]
This is not a new feature but improvement to buildah so it stops consuming and producing unnecessary overhead for jobs with longer duration.

Closes: https://github.com/containers/buildah/issues/4481

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
run,performance: check build stage status in `exponential backoff` manner
```

### DIFF
--- a/run_common.go
+++ b/run_common.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net"
 	"os"
 	"os/exec"
@@ -642,7 +643,15 @@ func runUsingRuntime(options RunOptions, configureNetwork bool, moreCreateArgs [
 		}
 	}()
 	signal.Notify(interrupted, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM)
+	iter := 0
 	for {
+		iter = iter + 1
+		exponential := int(math.Min((math.Pow(2, float64(iter)) * 50), 2000))
+		// Reached the max delay time reset counter to 0
+		// and start exponential delay again.
+		if exponential == 2000 {
+			iter = 0
+		}
 		now := time.Now()
 		var state specs.State
 		args = append(options.Args, "state", containerName)
@@ -673,7 +682,7 @@ func runUsingRuntime(options RunOptions, configureNetwork bool, moreCreateArgs [
 		select {
 		case <-finishedCopy:
 			atomic.StoreUint32(&stopped, 1)
-		case <-time.After(time.Until(now.Add(100 * time.Millisecond))):
+		case <-time.After(time.Until(now.Add(time.Duration(exponential) * time.Millisecond))):
 			continue
 		}
 		if atomic.LoadUint32(&stopped) != 0 {


run_common.go
```diff
@@ -642,7 +644,15 @@ func runUsingRuntime(options RunOptions, configureNetwork bool, moreCreateArgs [
 		}
 	}()
 	signal.Notify(interrupted, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM)
+	iter := 0
 	for {
+		iter = iter + 1
+		exponential := int(math.Min((math.Pow(2, float64(iter)) + rand.Float64()*500), 2000))
```
rhatdan, line 650: I don't see this as an exponential back off. more of a random timer?

I wrote a test program using your logic and I saw:
```
Inter= 1 304
Inter= 2 474
Inter= 3 340
Inter= 4 234
Inter= 5 244
Inter= 6 407
Inter= 7 160
Inter= 8 334
Inter= 9 560
Inter= 10 1174
Inter= 11 2000
```
I would have expected
```
Inter= 1 100
Inter= 2 200
Inter= 3 400
Inter= 4 800
Inter= 5 1600
Inter= 6 2000
```

rhatdan: In my example, I don't see the interval growing it goes from .407 seconds to .160 seconds on the next try.

flouthoc: In random it grows but in non-uniform manner and there are dips in between but maybe I am overthinking the random part hence  I have removed the random part and now pattern should look like.

```
-----iter 100
-----iter 200
-----iter 400
-----iter 800
-----iter 1600
-----iter 2000
-----iter 100
-----iter 200
-----iter 400
-----iter 800
-----iter 1600
-----iter 2000
```

rhatdan: I think that works well.

rhatdan:
LGTM
@vrothberg @giuseppe PTAL

run_common.go
```diff
@@ -642,7 +643,15 @@ func runUsingRuntime(options RunOptions, configureNetwork bool, moreCreateArgs [
 		}
 	}()
 	signal.Notify(interrupted, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM)
+	iter := 0
 	for {
+		iter = iter + 1
+		exponential := int(math.Min((math.Pow(2, float64(iter)) * 50), 2000))
+		// Reached the max delay time reset counter to 0
+		// and start exponential delay again.
+		if exponential == 2000 {
```
rhatdan, line 652: I am not sure of this code here, should we just stay at 2 seconds once it gets here rather then speeding up again?

giuseppe:
this will make the build process slower, since it takes longer to realize the container terminated.

`crun state` is really fast, I don't see how it is a problem to run it 10 times per second

flouthoc:
> this will make the build process slower, since it takes longer to realize the container terminated.
> 
> `crun state` is really fast, I don't see how it is a problem to run it 10 times per second

I think in original issue the concern was made for longer steps where build container is definitely gonna take time and even then buildah somehow keeps polling in background , hogging the resource via (exec calls). But still waiting for reply from @paravz on original issue https://github.com/containers/buildah/issues/4481 though maybe he is facing different issue than this one.

giuseppe:
> I think in original issue the concern was made for longer steps where build container is definitely gonna take time and even then buildah somehow keeps polling in background , hogging the resource via (exec calls). But still waiting for reply from @paravz on original issue #4481 though maybe he is facing different issue than this one.

`crun state` is still cheap though.  I don't think it is a good idea to make the entire process slower just to reduce the number of fork/exec calls

rhatdan:
Is there other ways we could check?  IE Watch for a exit code to appear?

flouthoc:
> Is there other ways we could check? IE Watch for a exit code to appear?

@rhatdan No I think the actual issue was polling , but once actual problem is explained in detail and if it gets decided that its needed then maybe we can make exponential backoff configurable via a flag. So just waiting for discussion here: https://github.com/containers/buildah/issues/4481

PR #4502, opened 2023-01-05
system tests: remove unhelpful assertions

edsantiago:
Regular primitive bats uses assertions like '[ $foo = something ]'.
These are worthless for debugging: when they fail, all you know
is that foo is not "something" but you don't know what foo _is_.

Find and replace those assertions with 'assert', which is
more informative. Instances found via:

   $ ack '^ *\[' tests/*.bats

There are many matches for 'test' (instead of '[') but those
mostly look like file-existence ones, which are less evil
than string-check tests. I'm leaving those be for now.

```release-note
None
```

### DIFF
--- a/tests/basic.bats
+++ b/tests/basic.bats
@@ -128,7 +128,7 @@ load helpers
   run_buildah rmi containers-storage:other-new-image
   run_buildah rmi another-new-image
   run_buildah images -q
-  [ "$output" != "" ]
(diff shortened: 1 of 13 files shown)


rhatdan:
/approve
LGTM
@flouthoc PTAL

flouthoc (approved):
LGTM
/lgtm
/hold

PR #4509, opened 2023-01-06
build, secret: allow relative mountpoints wrt to work dir for `--mount=type=secret`

flouthoc:
When working with `--mount=type=secret` allow `target`/`dst` to accept relative paths w.r.t to the configured work dir.

See detailed use-case here: https://github.com/containers/buildah/issues/4491

Closes: https://github.com/containers/buildah/issues/4491

**Steps to reproduce the issue from #4491:**

1. Create Dockerfile and Makefile:

Dockerfile:
```
FROM docker.io/ubuntu:22.04

WORKDIR /somedir

RUN --mount=type=secret,id=secret-foo,dst=secret1.txt --mount=type=secret,id=secret-bar,dst=secret2.txt \
     printf "PWD=%s\n" "$(pwd)" && ls -la && ls -la / && stat secret1.txt && stat secret2.txt && \
     cp secret1.txt /root/secret-foo.txt && \
     cp secret2.txt /root/secret-bar.txt
```

Makefile:
```
DOCKER ?= docker

.PHONY: build-container

build-container:
	rm -rf build
	mkdir build
	echo "secret:foo" >build/secret1.txt
	echo "secret:bar" >build/secret2.txt
	buildah --no-cache --secret id=secret-foo,src=build/secret1.txt --secret id=secret-bar,src=build/secret2.txt -t defanator/example:tag1 .
	podman run --rm -t -i defanator/example:tag1 cat /root/secret-foo.txt
	podman run --rm -t -i defanator/example:tag1 cat /root/secret-bar.txt
	podman rmi defanator/example:tag1
```
```
make
```

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
build, secret: allow realtive mountpoints wrt to work dir for `--mount=type=secret`
```

### DIFF
--- a/run_common.go
+++ b/run_common.go
@@ -1497,7 +1497,7 @@ func (b *Builder) runSetupRunMounts(mounts []string, sources runMountInfo, idMap
 		}
 		switch mountType {
 		case "secret":
-			mount, envFile, err := b.getSecretMount(tokens, sources.Secrets, idMaps)
+			mount, envFile, err := b.getSecretMount(tokens, sources.Secrets, idMaps, sources.WorkDir)
 			if err != nil {
 				return nil, nil, err
 			}
@@ -1598,7 +1598,7 @@ func (b *Builder) getTmpfsMount(tokens []string, idMaps IDMaps) (*spec.Mount, er
 	return &volumes[0], nil
 }
 
-func (b *Builder) getSecretMount(tokens []string, secrets map[string]define.Secret, idMaps IDMaps) (*spec.Mount, string, error) {
+func (b *Builder) getSecretMount(tokens []string, secrets map[string]define.Secret, idMaps IDMaps, workdir string) (*spec.Mount, string, error) {
 	errInvalidSyntax := errors.New("secret should have syntax id=id[,target=path,required=bool,mode=uint,uid=uint,gid=uint")
 	if len(tokens) == 0 {
 		return nil, "", errInvalidSyntax
@@ -1618,6 +1618,9 @@ func (b *Builder) getSecretMount(tokens []string, secrets map[string]define.Secr
 			id = kv[1]
 		case "target", "dst", "destination":
 			target = kv[1]
+			if !filepath.IsAbs(target) {
+				target = filepath.Join(workdir, target)
+			}
 		case "required":
 			required, err = strconv.ParseBool(kv[1])
 			if err != nil {
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -519,6 +519,13 @@ _EOF
   expect_output --substring "Groups:	1000"
 }
 
+@test "build-test --mount=type=secret test relative to workdir mount" {
+  local contextdir=$BUDFILES/secret-relative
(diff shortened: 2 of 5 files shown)


flouthoc:

run_common.go
```diff
@@ -1618,6 +1619,9 @@ func (b *Builder) getSecretMount(tokens []string, secrets map[string]define.Secr
 			id = kv[1]
 		case "target", "dst", "destination":
 			target = kv[1]
+			if !path.IsAbs(target) {
+				target = filepath.Join(workdir, target)
```
rhatdan, line 1623: Do we have to worry about this path escaping the container?
../../../../etc/passwd or something like that

flouthoc: I don't think it can escape container , I tried this 
```Dockerfile
FROM docker.io/ubuntu:22.04

WORKDIR /somedir

RUN --mount=type=secret,id=secret-foo,dst=secret1.txt --mount=type=secret,id=secret-bar,dst=../../../etc/passwd cat /etc/passwd
```
```
./buildah build --no-cache --secret id=secret-foo,src=build/secret1.txt  -t test1 .
```

TomSweeneyRedHat: Yeah, that feels twitch to me.  I see the build succeeded, but did it get the hosts passed file in the image?

flouthoc: @TomSweeneyRedHat No hosts passed file secret will not be part of the image unleass manual `cp` is made, any file or secret added by `--mount` instruction should not persist after run instruction.

rhatdan: Ok maybe this is a misunderstanding on my part, but does this effect the destination not the source.  I am less concerned about the source then the destination.

Can I do something like

RUN --mount=type=secret,id=secret-foo,dst=../../../secret1.txt --mount=type=secret,id=secret-bar,dst=../../../secret2.txt \

flouthoc: Yes this allowed and secret gets created on the path related to the `workdir` and does not persists after the `RUN` statement.

nalind: `filepath.Join()` returns a `Clean()`ed path, and `Clean()` removes any initial "/.." components.

tests/bud.bats
```diff
@@ -519,6 +519,24 @@ _EOF
...
+  local contextdir=${TEST_SCRATCH_DIR}/bud/platform
+  mkdir -p $contextdir
+  echo "secret:foo" >$contextdir/secret1.txt
+  echo "secret:bar" >$contextdir/secret2.txt
+  cat > $contextdir/Dockerfile << _EOF
+FROM alpine
+RUN mkdir test
+WORKDIR test
+RUN --mount=type=secret,id=secret-foo,dst=secret1.txt --mount=type=secret,id=secret-bar,dst=secret2.txt \
+     cat /test/secret1.txt && cat /test/secret2.txt
```
nalind, line 532: None of this content is changing, so they can just be regular files that live in the source tree instead of files we generate at test-time.

nalind:
LGTM.  Could stand to have a second use of `RUN --mount` that uses a destination location that starts with "../../" to confirm that the behavior matches our expectations, but that's not a blocker.

Issue #4520, opened 2023-01-16
Flake: bud-multiple-platform-with-base-as-default-arg: duplicate 'arm'


edsantiago:
New flake blowing up in podman-land today:
```
[+1439s] not ok 322 bud-multiple-platform-with-base-as-default-arg
...
# buildah manifest inspect localhost/testlist
#/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
#|     FAIL: arch list in manifest
#| expected: = '386 amd64 arm arm arm64 ppc64le s390x'
#|   actual:   '386 amd64 arm arm64 ppc64le s390x'
#\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Culprit seems to be the duplicate "arm" here:  https://github.com/containers/buildah/blob/9d987b92129dd41bd5e1643efa4914448970d2a4/tests/bud.bats#L5314

No idea why it's a flake: it seems like the sort of thing that should always fail.

### [bud] bud-multiple-platform-with-base-as-default-arg
* fedora-37 : bud podman fedora-37 root host
  * Podman PR 17126
    * [01-16 09:18](https://api.cirrus-ci.com/v1/artifact/task/5329548850495488/html/bud-podman-fedora-37-root-host.log.html#t--00322)
* fedora-37 : bud remote fedora-37 root host [remote]
  * Podman PR 17126
    * [01-16 09:47](https://api.cirrus-ci.com/v1/artifact/task/6084458609115136/html/bud-remote-fedora-37-root-host.log.html#t--00322)

edsantiago:
Submitted #4521 as possible fix. I'll leave it to others to understand what actually went wrong; all I care about is eliminating flakes.

edsantiago:
Per discussion on IRC: the two `arm`s are intentional. Something about v6 and v7.

vrothberg:
@nalind is this the flake you were referring to in libimage?  It is suuuuuper flaky in Podman at the moment.

nalind:
Yes, I believe that this is also https://github.com/containers/common/issues/1295#issuecomment-1387043876.

PR #4527, opened 2023-01-18
--all-platforms: make order deterministic

vrothberg:
A combination of yet-to-be-resolved containers/common/issues/1295 and a non-deterministic processing of the platforms on the Buildah side caused heavy flakes, so make it deterministic.

[NO NEW TESTS NEEDED] - the expected absence of the flake will tell.

Fixes: #4520

```release-note
None
```

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -527,7 +527,9 @@ func platformsForBaseImages(ctx context.Context, logger *logrus.Logger, dockerfi
 	if len(baseImages) == 0 {
 		return nil, fmt.Errorf("build uses no non-scratch base images: %w", err)
 	}
+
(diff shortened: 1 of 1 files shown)


vrothberg:
@nalind @edsantiago PTAL

nalind:
It's not the nondeterminism that's the bug here.  While forcing the iteration order to match that of the base image's manifest list will have that effect with jobs == 1, I don't think it can guarantee stable ordering when jobs > 1.
The manifest list that lets us see the problem also lists its `linux/arm/v6` image immediately before its `linux/arm/v7` image, so I think for the jobs == 1 case that the test runs, it'll just start erroring out every time.

Issue #4544, opened 2023-01-29
`buildah build --file a --file b .` behaviour does not match docs statement  "FROM instructions will only be accepted from the first specified file"


eriksjolund:
**Description**

`buildah build` man page [buildah-build.1.md](https://github.com/containers/buildah/blob/37a523ed2d46f4176b8dcd5deaf9ad14eb6f36f1/docs/buildah-build.1.md?plain=1#L329-L331) states:
_If more than one Containerfile is specified, FROM instructions will only be accepted from the
first specified file._

I ran the command

```
 buildah build --no-cache -t test -f Dockerfile.alpine -f Dockerfile.ubuntu .
```

result: __buildah__ used the `FROM` instruction from the __second__ not the __first__ specified Containerfile.

This was just an experiment and I didn't know what to expect. At least I see that the `buildah build` behaviour does not match what is written in [buildah-build.1.md](https://github.com/containers/buildah/blob/37a523ed2d46f4176b8dcd5deaf9ad14eb6f36f1/docs/buildah-build.1.md?plain=1#L329-L331).

**Steps to reproduce the issue:**
1. Run these commands
```
$ ls -la
total 8
drwxr-xr-x. 2 test2 test2 56 Jan 28 07:56 .
drwxr-xr-x. 3 test2 test2 17 Jan 28 07:54 ..
-rw-r--r--. 1 test2 test2 53 Jan 28 07:55 Dockerfile.alpine
-rw-r--r--. 1 test2 test2 53 Jan 28 07:54 Dockerfile.ubuntu
$ cat Dockerfile.alpine 
FROM docker.io/library/alpine
RUN touch /root/alpine
$ cat Dockerfile.ubuntu 
FROM docker.io/library/ubuntu
RUN touch /root/ubuntu
$ buildah build --no-cache -t test -f Dockerfile.alpine -f Dockerfile.ubuntu .
[2/2] STEP 1/2: FROM docker.io/library/ubuntu
[2/2] STEP 2/2: RUN touch /root/ubuntu
[2/2] COMMIT test
--> 8eddd77e157
Successfully tagged localhost/test:latest
8eddd77e157e587a02b6be2fe243a2c9ddb4ec218078707c6b93c419d7c37659
$ podman run --rm test ls /root
ubuntu
$ podman run --rm test cat /etc/os-release | grep ^NAME=
NAME="Ubuntu"
$
```

**Describe the results you received:**

```
[...]

PR #4546, opened 2023-01-30
docs: specify order preference for FROM when using multiple `containerfiles`

flouthoc:
When multiple files are specified buildah considers `FROM` instruction from the last file specified and so does `buildkit` and `docker` so lets specify that in docs.

[NO NEW TESTS NEEDED]
[NO TESTS NEEDED]

Closes: https://github.com/containers/buildah/issues/4544


rhatdan:
LGTM
I think we could need a similar fix in podman build man page, PTAL

rhatdan:
@nalind @giuseppe @TomSweeneyRedHat @umohnani8 @ashley-cui @vrothberg PTAL

Issue #4230, opened 2022-09-11
Containerfile with RUN --network none ... Does not work.


rhatdan:
Dockerfile is documented to support

RUN --network ...

But with Buildah this blows up:

   STEP 1/2: FROM alpine
   STEP 2/2: RUN --network=none ping -c 1 4.2.2.2
   error building at STEP "RUN --network=none ping -c 1 4.2.2.2": RUN only supports the --mount flag

https://docs.docker.com/engine/reference/builder/#run---network

rhatdan:
@flouthoc PTAL

rhatdan:
https://docs.docker.com/engine/reference/builder/#run---network

flouthoc:
Sure, SGTM I think a small change in imagebuilder should make it work.

flouthoc:
I'm taking over https://github.com/openshift/imagebuilder/pull/237 and will implement this there.

flouthoc:
This is bit delayed but I'll get to it.

rhatdan:
@flouthoc still delayed?

rhatdan:
@flouthoc friendly ping.

flouthoc:
PR https://github.com/containers/buildah/pull/4566 should close this.

rhatdan:
Getting close.

PR #4566, opened 2023-02-08
run: add support for inline `--network` in `RUN` statements

flouthoc:
Buildah should allow clients to support inline --network in RUN stmts so users
can create isolate or expose a particular build containers.

```Dockerfile
FROM alpine
RUN --network=host wget google.com
RUN --network=none wget google.com
```

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
run: add support for inline --network in RUN stmt
```

Closes: https://github.com/containers/buildah/issues/4230

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -599,6 +599,7 @@ func (s *StageExecutor) Run(run imagebuilder.Run, config docker.Config) error {
 		defer devNull.Close()
 		stdin = devNull
 	}
+	namespaceOptions := append([]define.NamespaceOption{}, s.executor.namespaceOptions...)
 	options := buildah.RunOptions{
 		Args:             s.executor.runtimeArgs,
 		Cmd:              config.Cmd,
@@ -609,7 +610,7 @@ func (s *StageExecutor) Run(run imagebuilder.Run, config docker.Config) error {
 		Hostname:         config.Hostname,
 		Logger:           s.executor.logger,
 		Mounts:           append([]Mount{}, s.executor.transientMounts...),
-		NamespaceOptions: s.executor.namespaceOptions,
+		NamespaceOptions: namespaceOptions,
 		NoHosts:          s.executor.noHosts,
 		NoPivot:          os.Getenv("BUILDAH_NOPIVOT") != "",
 		Quiet:            s.executor.quiet,
@@ -625,6 +626,19 @@ func (s *StageExecutor) Run(run imagebuilder.Run, config docker.Config) error {
 		Terminal:         buildah.WithoutTerminal,
 		User:             config.User,
 		WorkingDir:       config.WorkingDir,
+	}
+
+	// Honor `RUN --network=<>`.
+	switch run.Network {
+	case "host":
+		options.NamespaceOptions.AddOrReplace(define.NamespaceOption{Name: "network", Host: true})
+		options.ConfigureNetwork = define.NetworkEnabled
+	case "none":
+		options.ConfigureNetwork = define.NetworkDisabled
+	case "":
+		// do nothing
+	default:
+		return fmt.Errorf(`unsupported value %q for "RUN --network", must be either "host" or "none"`, run.Network)
 	}
 
 	if config.NetworkDisabled {
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -25,6 +25,25 @@ load helpers
   expect_output --substring "options use-vc"
 }
 
+@test "build with inline RUN --network=host" {
+  #hostns=$(readlink /proc/self/ns/net)
+  run readlink /proc/self/ns/net
+  hostns="$output"
+  run_buildah build $WITH_POLICY_JSON -t source -f $BUDFILES/inline-network/Dockerfile1
+  expect_output --from="${lines[9]}" "${hostns}"
+}
+
+@test "build with inline RUN --network=none" {
+  run_buildah 1 build $WITH_POLICY_JSON -t source -f $BUDFILES/inline-network/Dockerfile2
+  expect_output --substring "wget: bad address"
+}
+
+@test "build with inline RUN --network=fake" {
+  run_buildah 125 build $WITH_POLICY_JSON -t source -f $BUDFILES/inline-network/Dockerfile3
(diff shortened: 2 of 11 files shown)


TomSweeneyRedHat:
LGTM once the imagebuilder PR is merged and you vendor from that.

flouthoc:
Will rebase this after: https://github.com/openshift/imagebuilder/pull/247

rhatdan:
It's merged.

flouthoc:
@rhatdan @TomSweeneyRedHat Rebased. @containers/buildah-maintainers PTAL

run_linux.go
```diff
@@ -69,6 +69,23 @@ func setChildProcess() error {
...
+					namespace.Host = false
+				}
+			}
+		}
+		if options.RunNetwork == "none" {
+			options.ConfigureNetwork = define.NetworkDisabled
+		} else {
+			options.ConfigureNetwork = define.NetworkEnabled
+		}
+	}
```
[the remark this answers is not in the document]
flouthoc: Done, added an error for unsupported values.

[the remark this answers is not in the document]
rhatdan: This would read a lot better as a switch statement.

run_linux.go
```diff
@@ -69,6 +69,26 @@ func setChildProcess() error {
...
+				}
+			}
+		}
+		if options.RunNetwork == "none" {
+			options.ConfigureNetwork = define.NetworkDisabled
+		} else {
+			options.ConfigureNetwork = define.NetworkEnabled
+		}
+	} else if options.RunNetwork != "" {
+		return fmt.Errorf("unsupported value `%s` for `RUN --network`", options.RunNetwork)
```
nalind, line 89:
```suggestion
		return fmt.Errorf(`unsupported value %q for "RUN --network", must be either "host" or "none"`, options.RunNetwork)
```

run_linux.go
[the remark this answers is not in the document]
flouthoc: There were only two conditions on top level if hence i avoided a switch.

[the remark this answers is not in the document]
rhatdan: I see this as 4 options
"none", "host", "", default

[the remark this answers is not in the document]
nalind: Go ahead and move this logic into the caller (StageExecutor.Run) and drop the new `RunNetwork` option.  We don't need to add an option to an API whose only purpose is to modify other options that also get passed to the same API, when the caller can already accomplish the same end without it.

flouthoc:
@rhatdan @nalind PTAL again.
 
 Note: failed tests are flakes, restarted them.

imagebuildah/stage_executor.go
```diff
@@ -627,6 +627,23 @@ func (s *StageExecutor) Run(run imagebuilder.Run, config docker.Config) error {
 		WorkingDir:       config.WorkingDir,
 	}
 
+	// Honor `RUN --network=<>`.
+	switch run.Network {
+	case "host":
+		for _, namespace := range options.NamespaceOptions {
+			if namespace.Name == "network" {
+				namespace.Host = true
```
nalind, line 635: The initialization at line 612 is going to have to create a copy instead of just referencing the existing options slice to keep this from bleeding over to the next Run() call.

tests/bud.bats
```diff
@@ -25,6 +25,24 @@ load helpers
   expect_output --substring "options use-vc"
 }
 
+@test "build with inline RUN --network " {
+  local contextdir=${TEST_SCRATCH_DIR}/bud/platform
+  mkdir -p $contextdir
+
+  cat > $contextdir/Dockerfile << _EOF
```
nalind, line 32: This file doesn't need to be generated at test-time.

imagebuildah/stage_executor.go
nalind: Does this modify the value in the slice, or just the copy in the `namespace` variable?

tests/bud.bats
```diff
@@ -25,6 +25,23 @@ load helpers
   expect_output --substring "options use-vc"
 }
 
+@test "build with inline RUN --network=host" {
+  run_buildah build $WITH_POLICY_JSON -t source -f $BUDFILES/inline-network/Dockerfile1
+  expect_output --substring "Connecting to google.com"
+  expect_output --substring "index.html"
```
nalind, line 31: Perhaps this should be running `readlink /proc/self/ns/net` both during and after the build and comparing the two.

imagebuildah/stage_executor.go
flouthoc: My intention was value in the slice hence i modified the code to make it more clear.

tests/bud.bats
flouthoc: True a better test will be `readlink /proc/self/ns/net`

flouthoc:
@nalind @rhatdan @vrothberg @giuseppe @containers/buildah-maintainers PTAL

imagebuildah/stage_executor.go
```diff
@@ -627,6 +628,28 @@ func (s *StageExecutor) Run(run imagebuilder.Run, config docker.Config) error {
...
+	case "host":
+		didConfigure := false
+		for i, namespace := range options.NamespaceOptions {
+			if namespace.Name == "network" {
+				didConfigure = true
+				options.NamespaceOptions[i].Host = true
+			}
+		}
+		if !didConfigure {
+			options.NamespaceOptions = append(options.NamespaceOptions, define.NamespaceOption{Name: "network", Host: true})
```
nalind, line 642: `options.NamespaceOptions` has an `AddOrReplace()` method which should simplify this a bit.

nalind: Most of the logic in this `case` clause should be unnecessary now.

flouthoc: Done, removed other logic

nalind:
LGTM

flouthoc:
@rhatdan @containers/buildah-maintainers PTAL

PR #3025, opened 2021-02-20
Fix reaping of stages with no instructions

caiges:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:
Stages with no instructions are removed which can cause undesired behavior. For example, if a user has an existing image referenced in a Dockerfile:

```
FROM foo
FROM bar
```
`foo` will be removed from the host. This would normally be desired in a similar build:

```
FROM foo as build
RUN make thing
FROM scratch
COPY --from build thing /
```
where we would want to cleanup the `build` stage from the host.

#### How to verify it

- Pull the latest alpine: `buildah pull alpine`.
- Create a Dockerfile:

```
FROM alpine
FROM scratch
RUN touch /foo
```

- Build the image: `buildah bud .`

#### Which issue(s) this PR fixes:

Fixes #1920 

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #4595, opened 2023-02-15
executor: also add stage with no children to `cleanupStages` and ensure no dangling containers are left.

flouthoc:
While preparing list of cleanup stages, executor must add stage with no children also to cleanupStages since a stage can also exist in the form only `FROM base_image` i.e without any child nodes.

Ensures we leave no dandling build container for cases such as

```dockerfile
FROM quay.io/quay/busybox:latest AS busybox

FROM scratch

COPY --from=busybox /bin/busybox /bin/busybox
```

Closes: https://github.com/containers/buildah/issues/4572

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
executor: also add stage with no children to `cleanupStages` and ensure no dangling containers are left.
```


rhatdan:
LGTM
@nalind @giuseppe @vrothberg @umohnani8 PTAL

giuseppe (approved):
/lgtm

flouthoc:
rebased

TomSweeneyRedHat:
LGTM

imagebuildah/executor.go
```diff
@@ -522,7 +522,7 @@ func (b *Executor) buildStage(ctx context.Context, cleanupStages map[int]*StageE
 	// remove the intermediate/build containers, regardless of
 	// whether or not the stage's build fails.
 	// Skip cleanup if the stage has no instructions.
-	if b.forceRmIntermediateCtrs || !b.layers && len(stage.Node.Children) > 0 {
+	if b.forceRmIntermediateCtrs || !b.layers {
```
nalind, line 525: The comment on the above line becomes invalid.  Does this regressing on #3025?

rhatdan:
Fix the comment and we can merge

imagebuildah/executor.go
flouthoc: Removed the comment, the test added in #3025 still exists on main, which is not regressed I have even tried the case mentioned in #3025 manually https://github.com/containers/buildah/blob/main/tests/bud.bats#L4299

PR #4602, opened 2023-02-16
imagebuildah.StageExecutor: suppress bogus "Pushing cache []:..." messages

nalind:
#### What type of PR is this?

/kind bug

When we changed the cache-from and cache-to options to take slices of name values instead of single values, we didn't update logic that checked if a value was set to handle cases where we had a non-nil slice with no elements in it.

#### How to verify it

It's a cosmetic change, so I expect our existing tests should be enough.  I fixed a different bug in a test that was failing on my system.

```release-note
`buildah build` should no longer produce spurious "Pushing cache []:..." messages while building images.
```


TomSweeneyRedHat:
LGTM
and happy green test buttons

flouthoc (approved):
LGTM
/approve
/lgtm

PR #4611, opened 2023-02-20
docs/tutorials/04: add defaults for Run()

nalind:
#### What type of PR is this?

/kind documentation

In the tutorial, switch to calling `DefaultStoreOptionsAutoDetectUID()` instead of `DefaultStoreOptions()`, which should figure things out better.

In the tutorial, add an example of using Run(), where for API backward compatibility reasons, we can't tell the difference between "grant no capabilties by default" and "grant the default set of capabilities by default".

The default isolation can be set automatically, so start doing that at run-time, but have the tutorial look it up anyway because the tutorial on the web will be newer than our current release for at least a while.

#### How to verify it

Mostly docs.  ~The default is buried so deep that I don't think the CLI even hits the changed logic.~  It turns out that the conformance and integration tests cover this code path better than I thought they did.

None

One day, when we're ready to intentionally break API, we should change `BuilderOptions.Capabilities` into a pointer to a slice, so that `nil` can indicate that we should use the defaults from the configuration.

```release-note
None
```


flouthoc (approved):
LGTM

PR #4618, opened 2023-02-22
Ignore the base image's base image annotations

nalind:
#### What type of PR is this?

/kind bug

In a multi-stage build, when initializing a stage whose base is the result of a previous stage, don't reset the AnnotationBaseImageName and AnnotationBaseImageDigest annotations to point to that base image, since the values in the stage's base image are already correct.

#### How to verify it

Extended conformance test!

I think this is HACBS-1818.

```release-note
OCI images produced using multi-stage builds, where the final stage is based on an earlier stage, will no longer include an "org.opencontainers.image.base.name" annotation for the base of that earlier stage combined with an "org.opencontainers.image.base.digest" annotation which corresponds to the image produced by that earlier stage, which are two different images.
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -92,13 +92,7 @@ func (b *Builder) initConfig(ctx context.Context, img types.Image, sys *types.Sy
 				return fmt.Errorf("parsing OCI manifest %q: %w", string(b.Manifest), err)
 			}
 			for k, v := range v1Manifest.Annotations {
-				// NOTE: do not override annotations that are
-				// already set. Otherwise, we may erase
-				// annotations such as the digest of the base
-				// image.
-				if value := b.ImageAnnotations[k]; value == "" {
-					b.ImageAnnotations[k] = v
-				}
+				b.ImageAnnotations[k] = v
 			}
 		}
 	}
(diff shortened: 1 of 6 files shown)


config.go
```diff
@@ -92,13 +92,7 @@ func (b *Builder) initConfig(ctx context.Context, img types.Image, sys *types.Sy
...
 			}
 			for k, v := range v1Manifest.Annotations {
-				// NOTE: do not override annotations that are
-				// already set. Otherwise, we may erase
-				// annotations such as the digest of the base
-				// image.
-				if value := b.ImageAnnotations[k]; value == "" {
-					b.ImageAnnotations[k] = v
-				}
+				b.ImageAnnotations[k] = v
```
TomSweeneyRedHat, line 95: So overriding the annotations as noted in the old comment is not a concern?

TomSweeneyRedHat:
One question, probably for my education.  Plus it looks like a rebase might be required.  Otherwise LGTM.

config.go
nalind: That's right - we override them, or not, elsewhere.

nalind:
Rebased.

rhatdan:
@flouthoc PTAL

flouthoc (approved):
LGTM
/lgtm
/approved

nalind:
I think this is also STONEBLD-836.  Mentioning it here so I can find it later.

Issue #4319, opened 2022-10-05
buildah --os xor --arch still warns about missing TARGETOS/TARGETARCH


edsantiago:
Followup to #4295.
```console
$ printf "FROM alpine\nARG TARGETOS\nENV a=b\n" | bin/buildah build --os linux -q -
WARN[0000] missing "TARGETOS" build argument. Try adding "--build-arg TARGETOS=<VALUE>" to the command line 
da4bb4edfe98d2b868c91433bad7c082e9cc8b4070c05fc0107a64e040515f7b
```
Likewise with `--arch` and `TARGETARCH` only. Specifying both on the command line silences the warnings. (As does `--platform`, as fixed in #4295).

rhatdan:
@flouthoc @nalind do you have time to look into this one?

rhatdan:
@flouthoc any movement on this one?

PR #4634, opened 2023-03-01
executor: don't warn about unused `TARGETARCH`,`TARGETOS`,`TARGETPLATFORM`

flouthoc:
Buildah warns about TARETARCH, TARGETOS, TARGETPLATFORM in a multi-arch build however its not very useful for end-users and there are cases where this can be left untouched so do not warn about test default build args

Closes: https://github.com/containers/buildah/issues/4485

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
executor: don't warn about unused TARGETARCH,TARGETOS,TARGETPLATFORM
```


rhatdan:
LGTM
@nalind @vrothberg @giuseppe @umohnani8 @TomSweeneyRedHat PTAL

tests/bud.bats
```diff
@@ -610,6 +610,20 @@ _EOF
...
 }
 
+# Verify: https://github.com/containers/buildah/issues/4485
+@test "No default warning for TARGETARCH, TARGETOS, TARGETPLATFORM " {
+  local contextdir=${TEST_SCRATCH_DIR}/bud/platform
+  mkdir -p $contextdir
+
+  cat > $contextdir/Dockerfile << _EOF
+FROM alpine
+_EOF
```
nalind, line 620: This file doesn't need to be generated every time this test is run.

tests/bud.bats
```diff
@@ -610,6 +610,20 @@ _EOF
...
+
+  run_buildah build $WITH_POLICY_JSON --platform=linux/amd64,linux/arm64 -f $contextdir/Dockerfile
```
nalind, line 622: https://github.com/containers/buildah/issues/4319 is a similar issue.  Might as well cover that case as well.

TomSweeneyRedHat:
@flouthoc all kinds of test unhappiness on this one.  Other than @nalind 's comments. LGTM

PR #4650, opened 2023-03-09
stage_executor,cache-to: attempt to push cache only when `cacheKey` is valid

flouthoc:
Buildah's layer caching does not generates cache key or uses a layer from store if `RUN --mount` is used to make sure that always freshly mounted content is used in the `RUN` instruction, in such case there is no use of even pushing the cache since it will be never used by buildah.

Make sure that stage executor attempts to push cache only if cacheKey is generated.

Closes: https://github.com/containers/buildah/issues/4647

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
stage_executor,cache-to: attempt to push cache only when cacheKey is valid
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -1428,7 +1428,7 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 		// Try to push this cache to remote repository only
 		// if cache was present on local storage and not
 		// pulled from remote source while processing this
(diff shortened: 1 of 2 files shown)


flouthoc:

rhatdan:
LGTM

rhatdan:
@nalind @giuseppe @vrothberg @umohnani8 PTAL

giuseppe (approved):
LGTM

PR #4659, opened 2023-03-15
stage_executor: inline `--network` add `default` string

flouthoc:
Inline `RUN --network=default` should do nothing. This matches buildkit's behaviour here: https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#run---networkdefault

Also added to podman's document here: https://github.com/containers/common/pull/1366

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
stage_executor: inline `--network` add `default` string
```


flouthoc:
@nalind @rhatdan @giuseppe @vrothberg @containers/buildah-maintainers PTAL

tests/bud/inline-network/Dockerfile4
```diff
@@ -0,0 +1,3 @@
+FROM alpine
+RUN --network=default wget google.com
+
```
vrothberg, line 3: Tailing new line.

vrothberg (commented):
LGTM

TomSweeneyRedHat:
LGTM
but please do clean up that blank line that @vrothberg spotted.

flouthoc:
@TomSweeneyRedHat @vrothberg done.

giuseppe (approved):
LGTM

tests/bud.bats
```diff
@@ -43,6 +43,14 @@ load helpers
   expect_output --substring "unsupported value"
 }
 
+@test "build with inline RUN --network=default" {
+  run readlink /proc/self/ns/net
+  hostns="$output"
+  run_buildah build $WITH_POLICY_JSON -t source -f $BUDFILES/inline-network/Dockerfile4
+  expect_output --substring "index.html"
+  assert "$output" !~ "$hostns"
```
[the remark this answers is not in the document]
flouthoc: Yes I just added a line `RUN --network=default readlink /proc/self/ns/net` it should be not in hostns by default.

tests/bud.bats
```diff
@@ -43,6 +43,13 @@ load helpers
   expect_output --substring "unsupported value"
 }
 
+@test "build with inline RUN --network=default" {
+  run_buildah build --network=host $WITH_POLICY_JSON -t source -f $BUDFILES/inline-network/Dockerfile4
+  build1="$output"
```
nalind, line 48: This should be the same as the network namespace that this test itself is running in.

tests/bud.bats
```diff
@@ -43,6 +43,13 @@ load helpers
...
+  run_buildah build --network=private $WITH_POLICY_JSON -t source -f $BUDFILES/inline-network/Dockerfile4
+  assert "$output" !~ "$build1"
```
nalind, line 50: ... and running with `--network=private` should get a result that's both different from the `--network=host` test and the previous `--network=private`, unless I'm mistaken.

flouthoc: Did you mean different with the previous `--network=host` and another `--network=private` ? Yes I think two different builds should have different namespaces. 

I can add two builds with `--network=private` and compare them if you mean that.

tests/bud.bats
flouthoc: Done.

tests/bud.bats
nalind: Yes, that's what I was suggesting, since we're creating a new network namespace each time.

nalind: This `assert` probably needs to be changed - it's comparing the entire output from the previous build to the entire output of this one.  Since we're not caching, I'd expect to get different image IDs in those outputs, at least.

flouthoc: It seems buildah re-uses private namespace and only assigns new private namespace when other one is occupied, so this happens only when two private builds are parallel otherwise consecutive private build ends up getting same private namespace.

nalind: That's surprising to me, but I'm seeing it when I run `unshare -Urnmpf readlink /proc/self/ns/net` twice in succession, so the takeaway is that it's something the kernel's doing, and I was wrong to think that they'd be different.

flouthoc: ah i see its the kernel, anyways I have dropped this part from the test, only comparing a `host` and `private` ns build.

TomSweeneyRedHat:
@flouthoc needs a rebase and it looks like the tests are still barking.

TomSweeneyRedHat:
@flouthoc the tests are still spitting up a bit.

flouthoc:
@TomSweeneyRedHat Looks like a flake i have restarted.

tests/bud.bats
```diff
@@ -43,6 +43,19 @@ load helpers
...
+  _prefetch alpine
+  run readlink /proc/self/ns/net
+  hostns=$output
+  run_buildah build --network=host $WITH_POLICY_JSON -t source -f $BUDFILES/inline-network/Dockerfile4
+  firstns=${lines[2]}
+  assert "${hostns}" == "$firstns"
+  run_buildah build --network=private $WITH_POLICY_JSON -t source -f $BUDFILES/inline-network/Dockerfile4
+  secondns=${lines[2]}
+  assert "$secondns" != "$firstns"
+  assert "$secondns" != "$hostns"
```
nalind, line 56: We only need to do one of these tests if we've already verified that "$firstns" and "$hostns" are the same.

flouthoc:

nalind:
LGTM, give or take a rebase.

PR #4662, opened 2023-03-15
Update to Go 1.18

mtrmac:
#### What type of PR is this?

> /kind cleanup

Go 1.18 is already required in practice, see
```
grep 'func .*]('
```

This also renders the `-compat=1.17` option to `go mod tidy` unnecessary, so drop it.

The main expected benefit is that this avoid vendoring problems caused by the differences between `go mod tidy -compat=1.{16,17}`, see #4652 and #4656 .

#### How to verify it

Existing CI.

None

```release-note
None
```
[NO NEW TESTS NEEDED]

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -121,7 +121,7 @@ vendor_task:
 
     # Runs within Cirrus's "community cluster"
     container:
-        image: docker.io/library/golang:1.17
+        image: docker.io/library/golang:1.18
         cpu: 1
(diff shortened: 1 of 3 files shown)


mtrmac:
Note https://github.com/containers/buildah/pull/4663 — this PR might be a good validation of that change.

rhatdan:
@mtrmac Could you push again so the smoke tests will pass.

PR #4663, opened 2023-03-15
Fix requiring tests on Makefile changes

cevich:
#### What type of PR is this?

/kind other

Changes to how things are built shouldn't trigger the
'pr-should-include-tests` check.

#### How to verify it

A dummy Makefile change won't trigger the check.

[Example failure](https://cirrus-ci.com/task/4943480833703936)

```release-note
None
```

### DIFF
--- a/tests/validate/pr-should-include-tests
+++ b/tests/validate/pr-should-include-tests
@@ -42,12 +42,13 @@ fi
 filtered_changes=$(git diff --name-status $base $head |
                        awk '{print $2}'               |
                        fgrep -vx .cirrus.yml          |
-		       fgrep -vx .gitignore           |
+                       fgrep -vx .gitignore           |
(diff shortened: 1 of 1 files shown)


mtrmac:
I’m not sure that “renovate” is exactly the differentiator — wasn’t the version that was failing _only_ the one with my additions?

That change was to the top-level `Makefile`; maybe adding _that_ to the `filtered_changes` filter would have allowed that test to pass, _and_ would avoid similar failures for any non-Renovate `Makefile` updates.

mtrmac:
Case in point: https://cirrus-ci.com/task/5273861176950784 fails the same way.

cevich:
Good idea.

TomSweeneyRedHat:
@cevich, needs a rebase

cevich:
Force-push: Rebased & removed test-commit.

PR #4677, opened 2023-03-21
cmd/build: add support for `--push`

### LINKED ISSUE #4671: [Feature]: Add support for "buildx build --push" and/or "buildx build --output=type=registry"
### Feature request description

Please add support for the "buildx build --push" and/or "buildx build --output=type=registry" flags which would ensure that the built image is pushed to the registry. According to [the buildx reference](https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#options), the `--push` flag is just shorthand for `--output=type=registry`. So IMHO implementing both these things would be ideal, but I will be happy with either one.

### Have you considered any alternatives?

When using the CLI manually, I can of course use a separate command to push the image, this feature would help in integration with other tools - I am using [DevSpace](https://github.com/devspace-sh/devspace) CLI tool which, in some cases, calls `docker build build` via CLI, and it relies on the `--push` flag. And so far this is the only compatibility problem that I found when using podman(with podman-docker).

### Additional context

I may be able to contribute a PR if you think this is not to complex, and can give me a couple of pointers in the right direction. :)

danishprakash:
#### What type of PR is this?

/kind feature

Closes #4671 

Extremely WIP draft.

```release-note
- add support for `--push` flag to `bud` push image immediately after building
```


rhatdan:
I was thinking this would just be added to the client and not the service side.

buildah build --push

Would internally do

buildah build followed by buildah push.

rhatdan:
What do you think @nalind @flouthoc @vrothberg

nalind:
> I was thinking this would just be added to the client and not the service side.
> 
> buildah build --push
> 
> Would internally do
> 
> buildah build followed by buildah push.

That _would_ be simpler; API callers already have a `buildah.Push()` that they can call directly, and it's less API churn if we add such a call to the tail end of `cmd/buildah/buildCmd()` if `--push`.

For the "output=registry" case, setting `IsDir` in a `BuildOutputOption` while also adding a `Type` field that overlaps its meaning is confusing.  Making that a special case after we've extracted the output image's rootfs as an archive (an archive which doesn't get used if we're just going to push the image) looks wrong; pushing to a registry probably needs to be an entirely separate branch in `generateBuildOutput()`.

define/build.go
```diff
@@ -163,6 +163,7 @@ type BuildOptions struct {
 	// It allows end user to export recently built rootfs into a directory or tar.
 	// See the documentation of 'buildah build --output' for the details of the format.
 	BuildOutput string
+	Push        bool
```
TomSweeneyRedHat, line 166: Hmm, I'm a bit surprised we didn't define this prior.

TomSweeneyRedHat:
And just in case, we'll need updates to the man page and additional test added for the final PR.  From a quick flyby review, it LGTM other than other comments.

vrothberg:
> What do you think @nalind @flouthoc @vrothberg

I am no fan of mixing multiple commands as it breaks separation of concerns.  `buildah push` has over a dozen command-line options.  How would they be exposed in `buildah build`?

But I leave the decision up to the core maintainers of Buildah.

rhatdan:
I see this for compatibility with buildx
https://docs.docker.com/engine/reference/commandline/buildx_build/
--push |   | Shorthand for --output=type=registry

I don't think we have to add other options for push, and potentially could have just done this in podman build, but the --output=type calls are handled within buildah.

flouthoc:
I concur should we just do this at podman, and only add `--output=type=registry` to buildah ?

danishprakash:
> I was thinking this would just be added to the client and not the service side.
>
> buildah build --push
>
> Would internally do
>
> buildah build followed by buildah push.

I honestly didn't think that is how we'd want to push. On the outset, `--push` seems like it should be added serially to build but I felt it might not be extensible given there could be modifications/additions to the `--output` flag in future.



> For the "output=registry" case, setting `IsDir` in a `BuildOutputOption` while also adding a `Type` field that overlaps its meaning is confusing. Making that a special case after we've extracted the output image's rootfs as an archive (an archive which doesn't get used if we're just going to push the image) looks wrong; pushing to a registry probably needs to be an entirely separate branch in `generateBuildOutput()`.

You're right, that was an oversight.

danishprakash:
> I concur should we just do this at podman, and only add --output=type=registry to buildah ?

So adding the `--push` flag to podman which _would_ be translated to `--output` within buildah? Sounds about right to me. That was the original plan as mentioned [here](https://github.com/containers/buildah/issues/4671#issuecomment-1474173765).

pkg/parse/parse.go
```diff
@@ -576,25 +576,30 @@ func AuthConfig(creds string) (*types.DockerAuthConfig, error) {
...
 			IsDir:    false,
 			IsStdout: true}, nil
 	}
-	if !strings.Contains(buildOutput, ",") {
-		// expect default --output <dirname>
-		return define.BuildOutputOption{Path: buildOutput,
-			IsDir:    true,
-			IsStdout: false}, nil
+
+	// if !strings.Contains(buildOutput, ",") {
```
rhatdan, line 588: We don't like having commented out code in PRs, Just remove it.  We can always go back to Git to get the original code.

pkg/cli/common.go
```diff
@@ -270,6 +271,7 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
 	fs.BoolVar(&flags.Stdin, "stdin", false, "pass stdin into containers")
 	fs.StringArrayVarP(&flags.Tag, "tag", "t", []string{}, "tagged `name` to apply to the built image")
 	fs.StringVarP(&flags.BuildOutput, "output", "o", "", "output destination (format: type=local,dest=path)")
+	fs.BoolVar(&flags.Push, "push", false, "Shorthand for `--output=type=registry`")
```
rhatdan, line 274: I thought we decided we were not going to support the push flag in buildah but only in podman build?

internal/util/util.go
```diff
@@ -60,15 +65,51 @@ func GetTempDir() string {
...
 	}
-	if opts.IsDir {
+	if opts.Type == "image" {
+		if opts.Attrs["push"] != "true" {
+			return nil
+		}
+
+		image := opts.Attrs["name"]
+		destSpec := opts.Attrs["name"]
+		dest, err := alltransports.ParseImageName(destSpec)
```
rhatdan, line 83: I am not sure this code is correct, or necessary.  I know that @mtrmac and @vrothberg frown on parsing specs outside of github.com/containers/image or github.com/containers/common/libimage

define/types.go
```diff
@@ -100,6 +100,8 @@ type Secret struct {
 
 // BuildOutputOptions contains the the outcome of parsing the value of a build --output flag
 type BuildOutputOption struct {
+	Type     string
+	Attrs    map[string]string
```
mtrmac, line 104: Why is this adding an untyped map instead simple Go fields that exactly represent what we need?

define/types.go
```diff
@@ -100,6 +100,8 @@ type Secret struct {
 
 // BuildOutputOptions contains the the outcome of parsing the value of a build --output flag
 type BuildOutputOption struct {
+	Type     string
```
mtrmac, line 103: What are the values of `Type`? Shouldn’t this be an enum?

define/types.go
```diff
@@ -100,6 +100,8 @@ type Secret struct {
...
+	Attrs    map[string]string
 	Path     string // Only valid if !IsStdout
 	IsDir    bool
```
mtrmac, line 106: Given the addition of `Type`, this should either be removed or at least deprecated. (I’m not sure which one — is this supposed to be a stable external API?)

pkg/parse/parse.go
```diff
@@ -607,29 +612,36 @@ func GetBuildOutput(buildOutput string) (define.BuildOutputOption, error) {
...
-			} else if arr[1] == "tar" {
-				isDir = false
-			} else {
+			switch arr[1] {
+			case "local":
+				out.IsDir = true
+			case "tar":
+				out.IsDir = false
+			case "registry":
+				out.IsDir = true
```
mtrmac, line 621: This is surprising; at the _very least_ it should be documented in the field, or have a comment here explaining why it is necessary. (But my starting hypothesis is that the data structure should be redesigned instead, to make this unnecessary — I’m not sure whether there are any compatibility impacts.)

pkg/cli/build.go
```diff
@@ -289,14 +289,19 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
...
+		buildOption, err := parse.GetBuildOutput(iopts.BuildOutput, output)
 		if err != nil {
 			return options, nil, nil, err
 		}
 		if buildOption.IsStdout {
 			iopts.Quiet = true
 		}
 	}
+	if c.Flag("push").Changed {
+		if len(iopts.BuildOutput) == 0 {
```
mtrmac, line 301: If both of the options are specified, that should fail instead of silently ignoring one.

pkg/cli/common.go
```diff
@@ -272,6 +273,7 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
...
```
mtrmac, line 276: Man page / other documentation updates are missing in this PR.

internal/util/util.go
```diff
@@ -60,15 +65,51 @@ func GetTempDir() string {
...
+			if t := transports.Get(destTransport); t != nil {
+				return err
+			}
+
+			if strings.Contains(destSpec, "://") {
+				return err
+			}
+
+			destSpec = "docker://" + destSpec
+			dest2, err2 := alltransports.ParseImageName(destSpec)
```
mtrmac, line 96: If this is all that is necessary, use `docker.ParseReference` without involving `alltransports`.

internal/util/util.go
```diff
@@ -60,15 +65,51 @@ func GetTempDir() string {
...
+			}
+			dest = dest2
+			logrus.Debugf("Assuming docker:// as the transport method for DESTINATION: %s", destSpec)
+		}
+
+		libimageOptions := &libimage.PushOptions{}
+		libimageOptions.Writer = os.Stdout
+		runtime, err := libimage.RuntimeFromStore(store, &libimage.RuntimeOptions{SystemContext: &types.SystemContext{}})
+		destString := fmt.Sprintf("%s:%s", dest.Transport().Name(), dest.StringWithinTransport())
+		_, err = runtime.Push(context.Background(), image, destString, libimageOptions)
```
mtrmac, line 108: (Eventually, could `libImage` have an API that accepts `type.ImageReference` directly?)

internal/util/util.go
```diff
@@ -60,15 +65,51 @@ func GetTempDir() string {
 }
 
 // ExportFromReader reads bytes from given reader and exports to external tar, directory or stdout.
-func ExportFromReader(input io.Reader, opts define.BuildOutputOption) error {
+func ExportFromReader(input io.ReadCloser, store storage.Store, opts define.BuildOutputOption) error {
```
mtrmac, line 68: The two ways to invoke this, with `input` and `store`, are _so different_, and the callers _must_ call them differently (we wouldn’t want the caller of the registry push to create the `ReadCloser` at all), that I think the two should clearly be separate functions.

Note: I didn’t check _at all_ whether this is the right place to, or the right way to, implement a push, e.g. by comparing the `buildah push` implementation.

internal/util/util.go
```diff
@@ -60,15 +65,51 @@ func GetTempDir() string {
...
+		// add the docker:// transport to see if they neglected it.
+		if err != nil {
+			destTransport := strings.Split(destSpec, ":")[0]
+			if t := transports.Get(destTransport); t != nil {
+				return err
+			}
+
+			if strings.Contains(destSpec, "://") {
+				return err
+			}
```
mtrmac, line 93: What is this? If it is new code that you added because if you think it is necessary, I suppose it should have some comments.

If this is copy&pasted from elsewhere else in Buildah, it shouldn’t be copy&pasted; use a shared utility function to turn user input into an `ImageReference`. (I wouldn’t be surprised to heard that one — or multiple! — already exist, I didn’t go looking.)

internal/util/util.go
mtrmac: Yes; as a general principle I think the CLI layer, or something as close to it as possible, should be responsible for turning strings into a unique `types.ImageReference`.

Admittedly that’s not always possible, but it’s the thing to strive for.

define/types.go
mtrmac: (At the very least, if this can’t be done with native Go features, the values should be documented.)

pkg/cli/common.go
danishprakash: Yes, you're right, this was added in order to test the changes as part of this draft PR.

internal/util/util.go
danishprakash: You're right, I got this piece from cmd/buildah/push.go. Based on your comment here and https://github.com/containers/buildah/pull/4677#discussion_r1174183375, I've moved it to util/util.go in order to reuse it in the both the places.

internal/util/util.go
danishprakash: https://github.com/containers/buildah/pull/4677#discussion_r1174183375

pkg/parse/parse.go
danishprakash: This wasn't necessary, I think I forgot to clean this up as part of the recent refactor.

define/types.go
danishprakash: Thanks for pointing this out, updated it to use an enum here. Should be extensible and readable.

define/types.go
danishprakash: This was referenced from buildx but the idea remains the same, for every type, there are (and would be in the future) multiple arguments/options which are consolidated under attributes here. It's possible for us to have individual fields here but those would eventually proliferate into a heterogenous struct type and I'm not sure how ideal that would be.

define/types.go
danishprakash: I hadn't thought about this before, primarily because although `Type` as an attribute wasn't present, it was still being used as a string in most places. It'd be premature for me to comment without exploring its usage.

internal/util/util.go
danishprakash: Does it help if it's moved to util? https://github.com/containers/buildah/pull/4677#discussion_r1176042766

define/types.go
mtrmac: This is ultimately up to Buildah maintainers, not me,

---

IMHO individual fields, or individual types, is much better than `Attrs`, because `Attrs` is _equally_ heterogenous, just with much worse compiler / linter support.

We can do interfaces, type checks, wrappers, …

internal/util/util.go
```diff
@@ -89,21 +111,22 @@ func ExportFromReader(input io.Reader, opts define.BuildOutputOption) error {
 		err = chrootarchive.Untar(input, opts.Path, &archive.TarOptions{NoLchown: noLChown})
 		if err != nil {
 			return fmt.Errorf("failed while performing untar at %q: %w", opts.Path, err)
-		}
-	} else {
-		outFile := os.Stdout
-		if !opts.IsStdout {
-			outFile, err = os.Create(opts.Path)
+		} else {
```
mtrmac, line 114: It seems that a `}` was lost here, moving the `!IsDir` code inside `IsDir`

internal/util/util.go
```diff
@@ -60,15 +62,35 @@ func GetTempDir() string {
...
+func ExportFromReader(input io.ReadCloser, store storage.Store, opts define.BuildOutputOption) error {
 	var err error
 	if !filepath.IsAbs(opts.Path) {
 		opts.Path, err = filepath.Abs(opts.Path)
 		if err != nil {
 			return err
 		}
 	}
-	if opts.IsDir {
+	if opts.Type == define.Image {
```
mtrmac, line 73: (I would expect a `switch opts.Type` here, without referring to `IsDir`.)

pkg/parse/parse.go
```diff
@@ -607,29 +605,35 @@ func GetBuildOutput(buildOutput string) (define.BuildOutputOption, error) {
...
-			} else if arr[1] == "tar" {
-				isDir = false
-			} else {
+			switch arr[1] {
+			case define.Local.String():
+				out.IsDir = true
+			case define.Tar.String():
+				out.IsDir = false
+			case define.Registry.String():
+				out.Type = define.Image
```
mtrmac, line 614: `case Registry: Type = Image`???

It’s fine to have the UI strings be different from the internal implementation enum names. It’s quite hard to maintain code that has a single enum that contains both internal “type” values and “strings”, with some values usable as both, but some only usable as one of the two.

define/types.go
```diff
@@ -101,11 +101,43 @@ type Secret struct {
...
 	IsStdout bool
 }
 
+type BuildOutputType int
+
+const (
+	Docker BuildOutputType = iota
```
mtrmac, line 114: Having the zero value be `BuildOutputTypeUninitialized` or `…Undefined` would add a bit of protection against coding errors.

define/types.go
```diff
@@ -101,11 +101,43 @@ type Secret struct {
...
+	Type     BuildOutputType
+	Attrs    map[string]string
 	Path     string // Only valid if !IsStdout
 	IsDir    bool
 	IsStdout bool
 }
 
+type BuildOutputType int
+
+const (
```
mtrmac, line 113: Consider naming the constants `BuildOutput[Type]…`, to make the relationship clearer. (Another price we pay for Go’s “simplicity”.)

internal/util/util.go
```diff
@@ -60,15 +62,35 @@ func GetTempDir() string {
...
+		if opts.Attrs["push"] != "true" {
+			return nil
+		}
+
+		image := opts.Attrs["name"]
+		destSpec := opts.Attrs["name"]
+		dest, err := util.StringToImageReference(destSpec)
```
mtrmac, line 80: I would much prefer for this to be called higher in the call stack, at the very least in `GetBuildOutput`, or possibly outside of `imagebuildah` at all (but that would be a larger change).

(And that would require a Go field instead of a `Attrs` only carrying strings.)

util/util.go
```diff
@@ -50,6 +51,31 @@ func StringInSlice(s string, slice []string) bool {
 	return util.StringInSlice(s, slice)
 }
 
+func StringToImageReference(image string) (types.ImageReference, error) {
```
mtrmac, line 54: `UIArgumentToImageReference`? `HeuristicUserInputToImageReference`? Something along those lines? (Up to Buildah maintainers.)

I, fairly desperately, don’t want this code to be used to “round-trip” between `ImageReference` and `string` [there already are other functions for that], and this plain naming of the function suggests that it could be used for that purpose.

mtrmac (commented):
It’s quite likely I don’t have a full context of the goals / constraints of the Buildx feature set or compatibility; I’m looking basically only at this PR in isolation.

I’ll fully defer to actual Buildah maintainers.

define/types.go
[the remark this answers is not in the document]
danishprakash: > suggests that the external string parser, and the internal representation, may very well be different.

It's a shortcut yes, but we can very well still go ahead and use `--type=image` and it should work as expected or handle the flag appropriately. Not sure if we can differentiate this as interval vs external? Or I might be missing something.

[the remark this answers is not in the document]
danishprakash: > Is this intended to match docker buildx --output?
> 
> We might only need that compatibility in the option parser, not in the enum (yet?).

Yes, that's the plan but adding the not-yet-implemented constants shouldn't make too much of a difference, right? But if it's a matter of hygiene, then it makes sense.

pkg/parse/parse.go
danishprakash: > The registry exporter is a shortcut for type=image,push=true.

Ah, I can see how that's misleading but I tried to do this conversion here itself rather "implicitly". I'll take a look at this. I guess this is what you meant?

define/types.go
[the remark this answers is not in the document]
mtrmac: In general the way I think about this is that the buildx syntax is a CLI-level concern that’s not really important to the lower levels. (The push not to represent `Attrs` as raw strings is both an example of that philosophy, and a demonstration why using native data can be more useful.)

Notably if some feature isn’t implemented, the CLI layer should cleanly fail instead of doing the wrong thing — and that’s easier if the unimplemented features can’t be represented at all, instead of having a remote invisible dependency between the low-level implementation and the high-level CLI parser recognizing exactly the same subset of features.

internal/util/util.go
```diff
@@ -60,15 +61,29 @@ func GetTempDir() string {
...
+
+		libimageOptions := &libimage.PushOptions{}
+		libimageOptions.Writer = os.Stdout
+		runtime, err := libimage.RuntimeFromStore(store, &libimage.RuntimeOptions{SystemContext: &types.SystemContext{}})
+		dest := fmt.Sprintf("%s:%s", opts.ImageRef.Transport().Name(), opts.ImageRef.StringWithinTransport())
+		_, err = runtime.Push(context.Background(), opts.Image, dest, libimageOptions)
+		if err != nil {
+			return fmt.Errorf("failed while pushing image %+q: %w", opts.ImageRef, err)
+		}
+	case define.BuildOutputLocal:
```
mtrmac, line 86: This replaces an `opts.IsDir` condition … but the parser never sets `BuildOutputLocal` AFAICS.

I’d expect `IsDir` to be removed entirely, and replaced by the new `Type`.

internal/util/util.go
```diff
@@ -90,7 +105,7 @@ func ExportFromReader(input io.Reader, opts define.BuildOutputOption) error {
 		if err != nil {
 			return fmt.Errorf("failed while performing untar at %q: %w", opts.Path, err)
 		}
-	} else {
+	default:
```
mtrmac, line 108: (Consider using `case theOtherExpectedValue:` here, and failing with an internal error in `default`.)

util/util.go
mtrmac: > Some sort of abuse this function would get by being used in unintended places?

Yes. The function is a heuristic that can parse the input using several different format/semantics, depending on what happens to work.

A failure case I’ve seen fairly frequently in this space is finding some pre-existing function like that, similarly targeted at the UI and heuristically doing “what the user meant”, and embedding it somewhere deeper in the call stack, calling it with a string produced from a Go value, _without accounting for the details of the heuristics_, or possibly the CLI-level heuristics being expanded to add new user-level features. The outcome is that the Go value → string → Go value transformation (which might, sometimes, be unavoidable, e.g. for multi-machine operations) becomes inexact, hard to predict, and possibly even leads to security issues.

Hence all of the concern about _not using_ strings as an internal representation if at all possible, as the easiest preventative measure.

Secondarily, using the right string representation / API. The `types.ImageReference` values already have an API for producing a string representation, and for creating one from a string representation. That one (`alltransports.ParseImageName`) should be used anywhere below the CLI heuristic layers, and the new function should be named so that it isn’t accidentally used instead.

So that’s why I would like some mention of UI, and preferably “heuristics” in the name.

“Image string” does not help because it does not disambiguate “a lossy heuristic subject to change” from “a stable round-trip-safe value conversion”, either would apply to images and strings.

pkg/parse/parse.go
mtrmac: Basically I think it would be easiest if the the enum should were reserved for the `out.Type` fields, and this parser would just hard-code the specific string values in the `switch`.

(See elsewhere about buildx specifics, and unimplemented features, on the CLI layer vs. internal representation.)

StarpTech:
Hi, I just created https://github.com/containers/podman/issues/18642 It looks related to this PR.

> I am no fan of mixing multiple commands as it breaks separation of concerns. buildah push has over a dozen command-line options. How would they be exposed in buildah build?

I agree but a good reason to combine them is to enable parallelism?

define/types.go
flouthoc: I think this must be deprecated and not removed since this is a public API.

danishprakash: @flouthoc sorry but could you share how could one go about deprecating this? thanks

internal/util/util.go
danishprakash: Tried moving the new push functionality to a separate function.

---

> Note: I didn’t check at all whether this is the right place to, or the right way to, implement a push, e.g. by comparing the buildah push implementation.

I took reference from buildah's push implementation for this, minus the additional meta[1].

[1] - https://github.com/danishprakash/buildah/blob/b5060e29537ec0283a948f59148a07f0c83d4e92/push.go#L101

define/types.go
flouthoc: @danishprakash I think adding comment above the field is enough. This is not the offical doc but I guess you will find similar text in official doc is enough https://zchee.github.io/golang-wiki/Deprecated/

rhatdan:
@danishprakash Are you still working on this?

danishprakash:
Yes, @rhatdan. Will try to address the remaining comments and the tests by early next week.

pkg/parse/parse.go
```diff
@@ -570,25 +571,22 @@ func AuthConfig(creds string) (*types.DockerAuthConfig, error) {
...
-		return define.BuildOutputOption{Path: buildOutput,
-			IsDir:    true,
-			IsStdout: false}, nil
-	}
-	isDir := true
-	isStdout := false
+		return define.BuildOutputOption{
+			Path:     "",
+			IsStdout: true,
+		}, nil
```
mtrmac, line 581: This doesn’t set `Type`

pkg/parse/parse.go
```diff
@@ -601,38 +599,53 @@ func GetBuildOutput(buildOutput string) (define.BuildOutputOption, error) {
...
-			path = arr[1]
+			out.Path = arr[1]
 		default:
 			return define.BuildOutputOption{}, fmt.Errorf("unrecognized key %q in build output option: %q", arr[0], buildOutput)
 		}
 	}
 
-	if !typeSelected || !pathSelected {
-		return define.BuildOutputOption{}, fmt.Errorf("invalid build output option %q, accepted keys are type and dest must be present", buildOutput)
+	if out.Type == define.BuildOutputRegistry {
```
mtrmac, line 630: This can never be true, AFAICS.

pkg/parse/parse.go
```diff
@@ -601,38 +599,53 @@ func GetBuildOutput(buildOutput string) (define.BuildOutputOption, error) {
...
+	if out.Path == "-" {
+		if out.Type == define.BuildOutputLocal {
 			return define.BuildOutputOption{}, fmt.Errorf("invalid build output option %q, type=local and dest=- is not supported", buildOutput)
 		}
-		return define.BuildOutputOption{Path: "",
-			IsDir:    false,
-			IsStdout: true}, nil
+		return define.BuildOutputOption{
+			Path:     "",
+			IsStdout: true,
```
mtrmac, line 644: This does not set `Type`.

pkg/parse/parse.go
```diff
@@ -570,25 +571,22 @@ func AuthConfig(creds string) (*types.DockerAuthConfig, error) {
...
-func GetBuildOutput(buildOutput string) (define.BuildOutputOption, error) {
+func GetBuildOutput(buildOutput, image string) (define.BuildOutputOption, error) {
 	if len(buildOutput) == 1 && buildOutput == "-" {
 		// Feature parity with buildkit, output tar to stdout
 		// Read more here: https://docs.docker.com/engine/reference/commandline/build/#custom-build-outputs
-		return define.BuildOutputOption{Path: "",
-			IsDir:    false,
-			IsStdout: true}, nil
-	}
-	if !strings.Contains(buildOutput, ",") {
```
mtrmac, line 581: This case is now completely removed? Isn’t that a breaking change for existing users?

pkg/parse/parse.go
```diff
@@ -601,38 +599,53 @@ func GetBuildOutput(buildOutput string) (define.BuildOutputOption, error) {
...
 		}
 	}
 
-	if !typeSelected || !pathSelected {
-		return define.BuildOutputOption{}, fmt.Errorf("invalid build output option %q, accepted keys are type and dest must be present", buildOutput)
+	if out.Type == define.BuildOutputRegistry {
+		out.Push = true
+	}
+
+	if !typeSelected && !pathSelected {
```
mtrmac, line 634: This seems to need more nuance. Isn’t the path still required for local destinations, just not for the type `…Image` case?

And, for the type `…Image` case, isn’t specifying a path an error?

define/types.go
```diff
@@ -100,9 +101,40 @@ type Secret struct {
...
-	IsDir    bool
+	ImageRef imageTypes.ImageReference
+	Image    string
 	IsStdout bool
+	Path     string // Only valid if !IsStdout
+	Push     bool
+	Type     BuildOutputType
+
+	// Deprecated: Use Type instead to determine output type
+	IsDir bool
```
mtrmac, line 112: AFAICS `IsDir` is now never set. If so, keeping the field around is just as incompatible an API break as removing the field… and we might just as well remove it.

Alternatively, if we need to keep it, it must be set in a compatible way (and I don’t know what “a compatible way” would be for `Type == …Image`.)

Or, hypothetically, we could leave the `BuildOutputOption` and the existing code as is, parsing only local file outputs but keeping API compatibility, and introduce a new type and a new set of utilities that can represent image outputs. 

Buildah maintainers need to decide which one of the three it is; I don’t know. (My guess is the first one, at least I found no Podman user of `BuildOutputOption`, but I really lack the context.)

mtrmac (commented):
(FWIW the code right now fails a test.)

define/types.go
danishprakash: > keeping the field around is just as incompatible an API break as removing the field… and we might just as well remove it.

From what I can tell, and I have very little context, the idea was to minimize disruption. Outright removing it might cause more disruption than keeping it around and not using it but I understand your concern with having dead code lying around.

internal/util/util.go
```diff
@@ -65,16 +66,39 @@ func GetTempDir() string {
 	return "/var/tmp"
 }
 
+// ExportImage copies contents of the image to a new location
+func ExportImage(store storage.Store, opts define.BuildOutputOption) error {
```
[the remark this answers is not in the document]
danishprakash: Didn't know that was a thing, really, thanks for the heads up.

internal/util/util.go
```diff
@@ -65,16 +66,39 @@ func GetTempDir() string {
...
+	if !opts.Push {
+		return nil
+	}
+
+	libimageOptions := &libimage.PushOptions{}
+	libimageOptions.Writer = os.Stdout
```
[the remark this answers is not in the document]
danishprakash: > Implementation-wise, at a first glance it seems to me cleaner to delegate this to the existing buildah.Push

That'd be great, right now the contract is a little too much for anyone else to make use of it. On the other hand, if we can abstract out the "push" functionality out of both for better extensibility, that might be ideal.

pkg/parse/parse.go
danishprakash: It looks like a catch-call, I believe this is being handled by the switch case below, now that we expect extended attributes for types, we shouldn't be returning early here.

mtrmac: Isn‘t that confirming my impression that existing syntax stops working?

Why is that acceptable? (It might well be, I’m not that familiar with Buildah. But I definitely want this to be a conscious decision by Buildah maintainers.)

internal/util/util.go
```diff
@@ -65,16 +66,39 @@ func GetTempDir() string {
 	return "/var/tmp"
 }
 
+// ExportImage copies contents of the image to a new location
+func ExportImage(store storage.Store, opts define.BuildOutputOption) error {
+	if !opts.Push {
+		return nil
```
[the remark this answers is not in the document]
danishprakash: You're right, `type=image` on its own doesn't require anything additional since the image is built locally before we get to exports. So now, I'm handling a situation where we have `type=image,push=true` and if not, then proceed further down.

imagebuildah/stage_executor.go
```diff
@@ -2071,7 +2071,15 @@ func (s *StageExecutor) commit(ctx context.Context, createdBy string, emptyLayer
 	return imgID, ref, nil
 }
 
-func (s *StageExecutor) generateBuildOutput(buildOutputOpts define.BuildOutputOption) error {
+func (s *StageExecutor) generateBuildOutput(ops define.BuildOutputOption) error {
+	if ops.Push {
```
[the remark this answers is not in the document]
danishprakash: https://github.com/containers/buildah/pull/4677#discussion_r1301528720

pkg/parse/parse.go
danishprakash: Fair enough, but now that we _are_ handling extended attributes, it'd make little sense for us to still handle this case, and besides, it's going to be counterintuitive for the user to use this. But then again, i think it's better if someone who has more context on this piece comment once.

danishprakash: > that existing syntax stops working

Not entirely though, you can still go ahead and use `--output=type=image` and it'll do the right thing instead of defaulting to a preset condition as before.

rhatdan:
@flouthoc @nalind PTAL

flouthoc (changes requested):
This needs few tests in `bud.bats`, we already have some tests for pushing there ( you can take help from there )

rhatdan:
@danishprakash still working on this?

danishprakash:
Yes, going to get back to this end of this week.

jianzhangbjz:
Any updates? Thanks!
```console
Error: unknown flag: --push
See 'podman buildx build --help'
```

rhatdan:
@danishprakash would you like me to try to get someone to take this over?

danishprakash:
Tests were pending, I had added them a couple of months back. The two RPM tests failing, I'm not sure if they're related. Also, this still needs a final pass from @mtrmac.

danishprakash:
Tests were pending, I had added them a couple of months back. Two RPM tests are failing, I'm not sure if they're related. Also, this still needs a final pass from @mtrmac.

rhatdan:
@mtrmac PTAL

rhatdan:
@nalind @flouthoc PTAL

imagebuildah/stage_executor.go
```diff
@@ -2363,7 +2363,17 @@ func (s *StageExecutor) commit(ctx context.Context, createdBy string, emptyLayer
...
-func (s *StageExecutor) generateBuildOutput(buildOutputOpts define.BuildOutputOption) error {
+func (s *StageExecutor) generateBuildOutput(opts define.BuildOutputOption) error {
+	if opts.Type == define.BuildOutputImage {
+		if opts.Push {
+			err := internalUtil.PushImage(s.executor.store, opts, s.executor.systemContext)
+			if err != nil {
+				return fmt.Errorf("failed to export build output: %w", err)
+			}
+		}
+		return nil
```
rhatdan, line 2374: Is this right it does no output if not opts.Push?

internal/util/util.go
```diff
@@ -32,6 +34,7 @@ func LookupImage(ctx *types.SystemContext, store storage.Store, image string) (*
 		return nil, err
 	}
 	return localImage, nil
+	//
```
rhatdan, line 37: What is this?  Looks like it was supposed to be a comment?  Remove if not necessary.

pkg/parse/parse.go
rhatdan: @danishprakash You never andswered this?

imagebuildah/stage_executor.go
danishprakash: iiuc, `type=image` outputs a container image format, only when you specify `--output`, it additionally pushes the image to the registry.

pkg/parse/parse.go
danishprakash: I can't recall If I made that change deliberately or if it was a mistake, but you're right, path is required for local dests and not for type.

danishprakash: > And, for the type …Image case, isn’t specifying a path an error?

docker seems to not throw an error in this case, and ignores it

pasquale95:
Hi, is this still WIP?

danishprakash:
> Hi, is this still WIP?

Yes

kaovilai:
please rebase?

mangelajo:
👋

PR #4681, opened 2023-03-22
[CI:BUILD] Packit: Enable Copr builds on PR and commit to main

lsm5:
This commit adds Packit configuration files which will trigger rpm builds on copr:`rhcontainerbot/packit-builds` on every PR as well as on copr:`rhcontainerbot/podman-next` on every commit to main branch.

This commit will ensure main branch is always buildable on all supported Fedora and CentOS Stream versions for aarch64 and x86_64.
TODO: enable build checks for s390x and ppc64le while ensuring they don't take too long to build.

The packit builds reuse `buildah.spec.rpkg` and are thus independent of Fedora / CentOS dist-git.

This change will remove the need for the current webhook based triggering
 of rpm builds on rhcontainerbot/podman-next after commit to main.
 That will be instead handled by the `trigger: commit` action added in this
 PR. New builds will continue to get posted to the same link so users
 don't need to change any existing copr repo configuration.

#### What type of PR is this?

/kind other

Checks for downstream build issues prior to merge.

#### How to verify it
There will be additional packit tasks in CI.

None

None

```release-note
None
```

### DIFF
--- a/.packit.sh
+++ b/.packit.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+# This script handles any custom processing of the spec file generated using the `post-upstream-clone`
+# action and gets used by the fix-spec-file action in .packit.yaml.
+
+set -eo pipefail
+
+# Get Version from define/types.go in HEAD
+VERSION=$(grep ^$'\tVersion' define/types.go | cut -d\" -f2 | sed -e 's/-/~/')
+
+# Generate source tarball from HEAD
+git archive --prefix=buildah-$VERSION/ -o buildah-$VERSION.tar.gz HEAD
+
+# RPM Spec modifications
+
+# Use the Version from define/types.go in rpm spec
+sed -i "s/^Version:.*/Version: $VERSION/" buildah.spec
+
+# Use Packit's supplied variable in the Release field in rpm spec.
+# buildah.spec is generated using `rpkg spec --outdir ./` as mentioned in the
+# `post-upstream-clone` action in .packit.yaml.
+sed -i "s/^Release:.*/Release: $PACKIT_RPMSPEC_RELEASE%{?dist}/" buildah.spec
+
+# Use above generated tarball as Source in rpm spec
+sed -i "s/^Source:.*.tar.gz/Source: buildah-$VERSION.tar.gz/" buildah.spec
+
+# Use the right build dir for autosetup stage in rpm spec
+sed -i "s/^%setup.*/%autosetup -Sgit -n %{name}-$VERSION/" buildah.spec
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,30 @@
+---
+# See the documentation for more information:
+# https://packit.dev/docs/configuration/
+
+# Build targets can be found at:
+# https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/
+
+specfile_path: buildah.spec
+
+jobs:
+  - &copr
+    job: copr_build
+    trigger: pull_request
+    owner: rhcontainerbot
+    project: packit-builds
+    enable_net: true
+    srpm_build_deps:
+      - make
+      - rpkg
+    actions:
+      post-upstream-clone:
+        - "rpkg spec --outdir ./"
+      fix-spec-file:
+        - "bash .packit.sh"
+
+  - <<: *copr
+    # Run on commit to main branch
+    trigger: commit
+    branch: main
+    project: podman-next
--- a/buildah.go
+++ b/buildah.go
@@ -26,8 +26,7 @@ const (
 	// Package is the name of this package, used in help output and to
 	// identify working containers.
 	Package = define.Package
-	// Version for the Package.  Bump version in contrib/rpm/buildah.spec
-	// too.
+	// Version for the Package.
 	Version = define.Version
 	// The value we use to identify what type of information, currently a
 	// serialized Builder structure, we are using as per-container state.
--- a/define/types.go
+++ b/define/types.go
@@ -28,8 +28,7 @@ const (
 	// Package is the name of this package, used in help output and to
 	// identify working containers.
 	Package = "buildah"
-	// Version for the Package.  Bump version in contrib/rpm/buildah.spec
-	// too.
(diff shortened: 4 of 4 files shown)


.packit.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# Get Version from define/types.go in HEAD
+VERSION=$(grep ^$'\tVersion' define/types.go | cut -d\" -f2 | sed -e 's/-/~/')
+
+# Generate source tarball from HEAD
+git archive --prefix=buildah-$VERSION/ -o buildah-$VERSION.tar.gz HEAD
+
+# RPM Spec modifications
+
+# Use the Version from Cargo.toml in rpm spec
```
edsantiago, line 16: "Cargo.toml" is misleading here

.packit.sh
```diff
@@ -0,0 +1,26 @@
...
+sed -i "s/^Version:.*/Version: $VERSION/" buildah.spec
+
+# Use Packit's supplied variable in the Release field in rpm spec
+sed -i "s/^Release:.*/Release: $PACKIT_RPMSPEC_RELEASE%{?dist}/" buildah.spec
+
+# Use above generated tarball as Source in rpm spec
+sed -i "s/^Source:.*.tar.gz/Source: buildah-$VERSION.tar.gz/" buildah.spec
+
+# Use the right build dir for autosetup stage in rpm spec
+sed -i "s/^%setup.*/%autosetup -Sgit -n %{name}-$VERSION/" buildah.spec
```
edsantiago, line 26: Is `buildah.spec` generated from `buildah.spec.rpkg` in this directory? If so, it does not seem to include either `%setup` nor `%autosetup`...?

edsantiago (commented):
Mostly I just diffed against the `aardvark-dns` packit files. A few questions and nits.

.packit.yaml
```diff
@@ -0,0 +1,30 @@
...
+      post-upstream-clone:
+        - "rpkg spec --outdir ./"
+      fix-spec-file:
+        - "bash .packit.sh"
+
+  - <<: *copr
+    # Run on commit to main branch
+    trigger: commit
+    branch: main
+    project: podman-next
```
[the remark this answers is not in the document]
lsm5: So, that specifies the copr project where the build should be triggered after the commit is added to main branch upstream. It inherits most values from the block above it, but since the builds after merge should be in `rhcontainerbot/podman-next`, hence the `project: podman-next`.

.packit.sh
lsm5: done

lsm5:
> Btw, this change will remove the need for the current webhook based triggering of rpm builds on `rhcontainerbot/podman-next` after commit to main (https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/package/buildah/) . That will be instead handled by the `trigger: commit` action added in this PR. New builds will continue to get posted to the same link so users don't need to change any existing copr repo configuration.

I'll add this in the commit message..

edsantiago:
Thanks for the explanations. Most of it is above my head, because I don't use these COPRs or repos or anything like this, but LGTM

lsm5:
> Thanks for the explanations. Most of it is above my head, because I don't use these COPRs or repos or anything like this, but LGTM

Alrighty :) . Thank you!

edsantiago:
Hold on a second - I see four pack-it jobs running in CI right now. Shouldn't these run only when the PR merges? (Again, I don't grok packit, so maybe this is desired, but this looks like a lot of churn for very little benefit)

TomSweeneyRedHat:
I too am not a packit person.  From what I can see, LGTM

edsantiago:
Before this merges, I'd like @containers/buildah-maintainers to confirm that they're OK with having these jobs run on every PR, for many many hours. (I have no timing data, because all packit jobs show "completed in 0s", but I know that I looked at this PR page several times today, and packit jobs were spinning for at least two hours, probably longer).

lsm5:
> Hold on a second - I see four pack-it jobs running in CI right now. Shouldn't these run only when the PR merges? (Again, I don't grok packit, so maybe this is desired, but this looks like a lot of churn for very little benefit)

In `.packit.yaml`, there's a `trigger: pull_request` which will trigger copr builds on `rhcontainerbot/packit-builds` copr, which is what you see in the gating task list on the PR.

After the PR is merged (IOW, commit to `main` branch), the `trigger: commit` part in `.packit.yaml` will trigger builds on `rhcontainerbot/podman-next` copr.

So, different sets of tasks for different coprs, one to check if every PR builds, and the other for users to install after upstream merge.

RE: benefit, I'd rather build issues get caught in PR than after upstream tag release. Also, this doesn't depend on the official distro spec file anymore and the config file is a lot simpler too with a lot of the configuration handled in `rhcontainerbot/packit-builds` itself.

RE: `completed in 0s`, I can check with the packit devs.

lsm5:
@edsantiago I filed https://github.com/packit/packit-service/issues/1970 for `succeeded in 0s` issue.

rhatdan:
@lsm5 do you expect this PR to add any time to the build, or do you expect the new tests to pass within the ~1 hour that tests currently take to run?

lsm5:
Disabled s390x and ppc64le in the copr. Let's give this another try.

/packit copr-build.

lsm5:
Here's the latest copr_build that got triggered https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/build/5698195/

lsm5:
let me update the commit message again to show we are only doing x86_64 and aarch64 for now.

lsm5:
https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/build/5698195/ .. got done in 8 mins :)

lsm5:
@edsantiago copr build times should be drastically reduced now and the `0s` issue is on packit's radar. Please let me know if you still got any other concerns. I'll remove the slash-hold only after you green light this :)

edsantiago:
@lsm5 it sgtm in theory, I just really know nothing about packit and am uncomfortable with reliance on something that can add hours to a CI run. (Thank you for addressing the rare-arches slowness, it's just that I have zero sense for the packit farm and what other unreliabilities it may have). But, in the spirit of optimism:

/hold cancel

Issue #4688, opened 2023-03-27
Pass multiple host environment variables during `buildah run`


cbandy:
**Description**

`podman run` will pass multiple environment variables from the host to the container when the name passed to [`--env` ends with `*`](https://docs.podman.io/en/latest/markdown/podman-run.1.html#env-e-env):

> As a special case, if an environment variable ending in `*` is specified without a value, Podman will search the host environment for variables starting with the prefix and will add those variables to the container.

> ```
> $ export ENV1=a
> $ podman run --env 'ENV*' alpine env | grep ENV
> ENV1=a
> ```

I have found this very useful for tools that can be configured through the environment. It would be great if `buildah run` also had this behavior. For example,

```console
$ export TOOL_ONE=1 TOOL_TWO=profit
$ CONTAINER=$(buildah from alpine)
$ buildah run --env 'TOOL_*' "${CONTAINER}" -- env
…
TOOL_ONE=1
TOOL_TWO=profit
```

**Output of `buildah version`:**

```
buildah version 1.27.3 (image-spec 1.0.2-dev, runtime-spec 1.0.2-dev)
```

**Output of `podman version`:**

```
podman version 4.2.0
```

rhatdan:
This sounds like a good idea, and I would recommend that it be done to buildah run as well as buildah build.

Interested in opening a PR to make this happen?

cbandy:
I started to look at this. `buildah build` also has an `--unsetenv` flag. Should it do any globbing?

```
      --unsetenv strings               unset environment variable from final image
```

rhatdan:
If we do globbing on setenv, then to me it makes sense that we do it on unsetenv.

cbandy:
🤔 I see also `buildah config --env`. Any thoughts about that?

rhatdan:
Same.  They should all handle globs.

flouthoc:
This seems to be already fixed.

PR #4702, opened 2023-04-02
Pass process environment variables by reference

cbandy:
#### What type of PR is this?

/kind api-change

- See #4688

This PR contains only some of the changes discussed in #4688.

The first commit allows `buildah run --env` to lookup environment variables from the process environment using a rudimentary "glob" syntax.

The second commit does the same for `buildah build --env`, but alters the behavior when only a key is passed. Prior to this change, passing `--env A` when no `A` is defined in the environment sets `A=` (blank) in the container. After this change, `A` remains unset in the container.

### DIFF
--- a/cmd/buildah/run.go
+++ b/cmd/buildah/run.go
@@ -152,7 +152,6 @@ func runCmd(c *cobra.Command, args []string, iopts runInputOptions) error {
 		CNIConfigDir:     iopts.CNIConfigDir,
 		AddCapabilities:  iopts.capAdd,
 		DropCapabilities: iopts.capDrop,
-		Env:              iopts.env,
 		WorkingDir:       iopts.workingDir,
 	}
 
@@ -163,6 +162,8 @@ func runCmd(c *cobra.Command, args []string, iopts runInputOptions) error {
 			options.Terminal = buildah.WithoutTerminal
 		}
 	}
+
+	options.Env = buildahcli.LookupEnvVarReferences(iopts.env, os.Environ())
 
 	systemContext, err := parse.SystemContextFromOptions(c)
 	if err != nil {
--- a/pkg/cli/common.go
+++ b/pkg/cli/common.go
@@ -484,3 +484,42 @@ func AliasFlags(f *pflag.FlagSet, name string) pflag.NormalizedName {
 	}
 	return pflag.NormalizedName(name)
 }
+
+// LookupEnvVarReferences returns a copy of specs with keys and values resolved
+// from environ. Strings are in "key=value" form, the same as [os.Environ].
+//
+//   - When a string in specs lacks "=", it is treated as a key and the value
+//     is retrieved from environ. When the key is missing from environ, neither
+//     the key nor value are returned.
+//
+//   - When a string in specs lacks "=" and ends with "*", it is treated as
+//     a key prefix and any keys with the same prefix in environ are returned.
+//
+//   - When a string in specs is exactly "*", all keys and values in environ
+//     are returned.
+func LookupEnvVarReferences(specs, environ []string) []string {
+	result := make([]string, 0, len(specs))
+
+	for _, spec := range specs {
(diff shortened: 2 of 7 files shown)


cbandy:
I should have opened this as a draft. This is only part of what has been discussed in #4688 because I  quickly found differences in how the flags work across different commands.

1. It seems reasonable to me that `buildah run --env A` when there is no `A` in the host environment does nothing to the environment inside the container. This thinking/approach differs from `buildah build`, however:

2. `buildah build --env` already handles key-only arguments. When the referenced key does not exist in the environment, it is assigned the empty string within the container.

   https://github.com/containers/buildah/blob/40dff190acadda525b83390c6649d32345af2230/imagebuildah/executor.go#L515-L516

3. `buildah config --env` already handles key-only arguments. When the referenced key does not exist in the environment _or_ is blank, an error is returned.

   https://github.com/containers/buildah/blob/40dff190acadda525b83390c6649d32345af2230/cmd/buildah/config.go#L252-L255

4. `buildah config --env` already has a special syntax: the `-` key unsets all variables and the `-` suffix unsets one variable. Should I proceed with handling _two_ suffixes such that `*-` means to unset all variables with a certain prefix? 🤔 It almost sounds better to add a `buildah config --unsetenv` flag such that `*` unsets all variables, but IIUC ordering sets with unsets would become impossible.


@rhatdan I stopped implementing after (2) for feedback. Which of the three behaviors in (1), (2), and (3) should I use while implementing (1) ? Should I alter the other (either 2 or 3) to match?

cmd/buildah/run.go
```diff
@@ -152,7 +152,6 @@ func runCmd(c *cobra.Command, args []string, iopts runInputOptions) error {
 		CNIConfigDir:     iopts.CNIConfigDir,
 		AddCapabilities:  iopts.capAdd,
 		DropCapabilities: iopts.capDrop,
-		Env:              iopts.env,
```
TomSweeneyRedHat, line 155: I'm surprised this isn't being used elsewhere and that it's not complaining loudly...

rhatdan, line 155: The line below can be just used here, no reason to break it out.
Env: buildahcli.LookupEnvVarReferences(iopts.env, os.Environ())

pkg/cli/common.go
```diff
@@ -484,3 +484,42 @@ func AliasFlags(f *pflag.FlagSet, name string) pflag.NormalizedName {
...
+//   - When a string in specs lacks "=", it is treated as a key and the value
+//     is retrieved from environ. When the key is missing from environ, neither
+//     the key nor value are returned.
+//
+//   - When a string in specs lacks "=" and ends with "*", it is treated as
+//     a key prefix and any keys with the same prefix in environ are returned.
+//
+//   - When a string in specs is exactly "*", all keys and values in environ
+//     are returned.
+func LookupEnvVarReferences(specs, environ []string) []string {
```
rhatdan, line 500: Are you planning on making podman use this function as well?

rhatdan:
Since I want to get this into release, I am merging. But we could do some minor cleanups and add tests.
/approve
/lgtm
Thanks @cbandy

cbandy:
Sorry for the delay. I was expecting a response to this question:

> @rhatdan I stopped implementing after (2) for feedback. Which of the three behaviors in (1), (2), and (3) should I use while implementing (1) ? Should I alter the other (either 2 or 3) to match?

pkg/cli/common.go
cbandy: I have no plans for that. It would be nice for Buildah and Podman to do these things consistently, but the tools satisfy different use cases.

I encountered difficulty making env flags consistent within Buildah.

rhatdan:
@flouthoc Could you make these common between the two packages.

flouthoc:
Already modified podman to use this from buildah's package here: https://github.com/containers/podman/pull/18130

PR #4703, opened 2023-04-03
copier: retain symlink target w/ follow-link

### LINKED ISSUE #16585: Add support for --follow-link option in podman cp
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

Currently `podman cp` doesn't support for `--follow-link` like `docker cp` does.

1. podman cp --follow-link xyz.tar container:/tmp/

`Error: unknown flag: --follow-link`

File `cp` command should copy target file to container.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `podman version`:**
```
$ podman version
Version:      3.4.7
API Version:  3.4.7
Go Version:   go1.19.2
Built:        Thu Jan  1 05:30:00 1970
OS/Arch:      linux/arm64

```

**Output of `podman info`:**

```
$ podman info
host:
  arch: arm64
  buildahVersion: 1.23.1
  cgroupControllers:
  - cpu
  - pids
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: 'conmon: /usr/bin/conmon'
    path: /usr/bin/conmon
    version: 'conmon version 2.1.3, commit: unknown'
  cpus: 4
  distribution:
    codename: bookworm
    distribution: debian
    version: unknown
  eventLogger: journald
  hostname: raspberrypi
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 5.15.76-v8+
  linkmode: dynamic
  logDriver: journald
  memFree: 201830400
  memTotal: 3529732096
  ociRuntime:
    name: runc
    package: 'runc: /usr/bin/runc'
    path: /usr/bin/runc
    version: |-
      runc version 1.1.4+ds1
      commit: 1.1.4+ds1-1
      spec: 1.0.2-dev
      go: go1.19
      libseccomp: 2.5.4
  os: linux
  remoteSocket:
  security:
    apparmorEnabled: false
    rootless: true
    seccompEnabled: true
    selinuxEnabled: false
  serviceIsRemote: false
  slirp4netns:
    executable: /usr/bin/slirp4netns
    package: 'slirp4netns: /usr/bin/slirp4netns'
    version: |-
      slirp4netns version 1.2.0
      commit: 656041d45cfca7a4176f6b7eed9e4fe6c11e8383
      libslirp: 4.7.0
      SLIRP_CONFIG_VERSION_MAX: 4
      libseccomp: 2.5.4
  swapFree: 26738688
  swapTotal: 104853504
  uptime: 69h 57m 46.53s (Approximately 2.88 days)
plugins:
  log:
  - k8s-file
  - none
  - journald
  network:
  - bridge
  - macvlan
  volume:
  - local
registries: {}
store:
  configFile: /home/pi/.config/containers/storage.conf
  containerStore:
    number: 0
    paused: 0
    running: 0
    stopped: 0
  graphDriverName: vfs
  graphOptions: {}
  graphRoot: /home/pi/.local/share/containers/storage
  graphStatus: {}
  imageStore:
    number: 19
  volumePath: /home/pi/.local/share/containers/storage/volumes
version:
  APIVersion: 3.4.7
  Built: 0
  BuiltTime: Thu Jan  1 05:30:00 1970
  GitCommit: ""
  GoVersion: go1.19.2
  OsArch: linux/arm64
  Version: 3.4.7
```

**Package info (e.g. output of `rpm -q podman` or `apt list podman` or `brew info podman`):**

```
$ apt list podman
Listing... Done
podman/testing,now 3.4.7+ds1-3+b3 arm64 [installed]
podman/testing 3.4.7+ds1-3+b3 armhf

```

**Have you tested with the latest version of Podman and have you checked the Podman Troubleshooting Guide? (https://github.com/containers/podman/blob/main/troubleshooting.md)**

Yes

**Additional environment details (AWS, VirtualBox, physical, etc.):**
Same behavior on other OSs

danishprakash:
#### What type of PR is this?

/kind feature

Adds support for _not_ following link with `podman cp` by default in order to conform to docker's behavior.

This PR is in conjunction with https://github.com/containers/podman/pull/18020 which closes https://github.com/containers/podman/issues/16585

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -1532,7 +1532,17 @@ func copierHandlerGet(bulkWriter io.Writer, req request, pm *fileutils.PatternMa
 					}
 				}
 
-				if err := copierHandlerGetOne(info, "", name, item, req.GetOptions, tw, hardlinkChecker, idMappings); err != nil {
+				// If following link, pass symlink target for
+				// the link to be generated at the destination
+				var symlinkTarget string
+				if req.GetOptions.NoDerefSymlinks && info.Mode()&os.ModeType == os.ModeSymlink {
+					symlinkTarget, err = os.Readlink(item)
+					if err != nil {
+						return err
+					}
+				}
+
+				if err := copierHandlerGetOne(info, symlinkTarget, name, item, req.GetOptions, tw, hardlinkChecker, idMappings); err != nil {
 					if req.GetOptions.IgnoreUnreadable && errorIsPermission(err) {
 						continue
 					}
--- a/copier/copier_test.go
+++ b/copier/copier_test.go
@@ -2438,3 +2438,90 @@ func TestSortedExtendedGlob(t *testing.T) {
 	require.NoError(t, err, "globbing")
 	require.ElementsMatch(t, expect, matched, "sorted globbing")
 }
+
+func TestGetNoDerefSymlink(t *testing.T) {
+	couldChroot := canChroot
+	canChroot = false
+	testGetNoDerefSymlink(t)
+	canChroot = couldChroot
+}
+
+func testGetNoDerefSymlink(t *testing.T) {
+	var testArchives = []struct {
+		name     string
+		rootOnly bool
+		headers  []tar.Header
+		contents map[string][]byte
+	}{
+		{
+			name:     "regular",
+			rootOnly: false,
+			headers: []tar.Header{
+				{Name: "link-b", Typeflag: tar.TypeSymlink, Linkname: "../file-doesnt-exist", Size: 23, Mode: 0777, ModTime: testDate},
+			},
+			contents: map[string][]byte{
+				"archive-a": testArchiveSlice,
+			},
+		},
+	}
+
+	topdir := "."
+	for _, testArchive := range testArchives {
+		if uid != 0 && testArchive.rootOnly {
+			t.Logf("test archive %q can only be tested with root privileges, skipping", testArchive.name)
+			continue
+		}
+
+		dir, err := makeContextFromArchive(t, makeArchive(testArchive.headers, testArchive.contents), topdir)
+		require.NoErrorf(t, err, "error creating context from archive %q", testArchive.name)
+
+		root := dir
+
+		for _, noDerefSymlinks := range []bool{true, false} {
+			for _, testItem := range testArchive.headers {
+				name := filepath.FromSlash(testItem.Name)
+				name = filepath.Join(root, topdir, name)
+				if !t.Failed() && testItem.Typeflag == tar.TypeSymlink {
+					t.Run(fmt.Sprintf("noDerefSymlinks=%t,name=%s", noDerefSymlinks, name), func(t *testing.T) {
+						var getErr error
+						var wg sync.WaitGroup
+						getOptions := GetOptions{}
+						getOptions.NoDerefSymlinks = noDerefSymlinks
+						pipeReader, pipeWriter := io.Pipe()
+						wg.Add(1)
+						go func() {
+							getErr = Get(root, topdir, getOptions, []string{name}, pipeWriter)
+							pipeWriter.Close()
+							wg.Done()
+						}()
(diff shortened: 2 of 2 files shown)


TomSweeneyRedHat:
Tests are a bit unhappy.  I always twitch when I see anything to do with symlinks.  I can't point a finger at anything wrong with this, but I'm not sure if it's a good idea to do the copy of the symlink.

@flouthoc ?

danishprakash:
>  I can't point a finger at anything wrong with this, but I'm not sure if it's a good idea to do the copy of the symlink.

You're right, but unfortunately, that's how docker does things right now. They cp the symlink as it is irrespective of whether the target is present or not. I'll try to look for what's the rationale behind this but otherwise, this _should_ be harmless with a global bool @rhatdan [suggested](https://github.com/containers/podman/issues/16585#issuecomment-1474072322). Wdyt?

rhatdan:
If the cp is supposed to be into the container image, then we need to make sure the symbolic link still points within the image. IE Make sure it does not point at the hosts /etc/passwd.

Similarly if you are copying from the container to the host, we should make sure the source is within the container image.

danishprakash:
> then we need to make sure the symbolic link still points within the image.

Are you suggesting we _move_ the target to the container and then create a symbolic link? That'd be quite surprising in terms of user experience, wouldn't it? And besides, it'll be diverting quite far from how docker does things. 

>  IE Make sure it does not point at the hosts /etc/passwd.

With this change, it would point to the host's `/etc/passwd` but it would just be an empty link anyways.

rhatdan:
Ok this LGTM, now we need to fix tests.

danishprakash:
Planning to add a test for this over at https://github.com/containers/podman/pull/18020 utilizing the `--follow-link` flag. I tried doing it here, but we don't plan to add the flag here and copier would anyway just return the link I create as part of the unit tests. Thoughts, @rhatdan?

rhatdan:
LGTM
@flouthoc PTAL

vrothberg:
@flouthoc @nalind PTAL

nalind:
Without a unit test, I'm _very_ reluctant to change things in `copier`.

Without changing the `info` value that's passed to `copierHandlerGetOne()`, the `Typeflag` in the header it outputs for the tar archive shouldn't be changed, so I'm not sure this produces the desired effect.  But this _does_ point out that we weren't reading the target info for the symlink when the loop at line 1160 wasn't being entered, and that's definitely a bug.  It would probably make more sense to add the `os.Readlink` call closer to there, though.

TomSweeneyRedHat:
@danishprakash any progress with the test?

danishprakash:
> Without changing the `info` value that's passed to `copierHandlerGetOne()`, the `Typeflag` in the header it outputs for the tar archive shouldn't be changed, so I'm not sure this produces the desired effect

From what I know, the problem isn't with symlinks but with how we deal with the target when _not_ following links. So If I understand the issue, this is what's happening:

During the copy from the host, `Get()` was always following the links before creating the archive[1]. Which would then be `Put()` as a regular file with no links, hence adhering to the implicit `--follow-link` behavior podman has always had so far. 

But the behavior we need is to allow podman to follow or not follow links based on a flag that is propagated from Podman to Buildah via `NoDerefSymlinks`. Now, in the case of not following links, we'd want to pass on this item as a TypeSymlink over to Put with the same target as on the host (this is not ideal but this is what docker does and we're aiming for parity).

[1] - https://github.com/danishprakash/buildah/blob/b5060e29537ec0283a948f59148a07f0c83d4e92/copier/copier.go#L1160

rhatdan:
@nalind PTAL

flouthoc (changes requested):
Could you please `rebase` this.

PR #4745, opened 2023-04-18
Add smoke tests for encryption CLI helpers

mtrmac:
#### What type of PR is this?

> /kind other

A non-`nil` but empty decryption configuration seems to be valid enough to trigger decryption in some configurations, per https://github.com/containers/podman/issues/18196 .

Like in Skopeo and Podman, only decrypt when the user explicitly instructs us to (e.g. not triggering decryption based on environment variables).

#### How to verify it

Manually, I guess.

None

```release-note
An explicit `--decryption-key` option is now mandatory to trigger image decryption, even if the environment implicitly points at usable keys.
```

### DIFF
--- a/pkg/cli/common_test.go
+++ b/pkg/cli/common_test.go
@@ -114,6 +114,21 @@ func TestLookupEnvVarReferences(t *testing.T) {
 	})
 }
 
+func TestDecryptConfig(t *testing.T) {
+	// Just a smoke test for the default path.
+	res, err := DecryptConfig(nil)
+	assert.NoError(t, err)
+	assert.Nil(t, res)
+}
+
+func TestEncryptConfig(t *testing.T) {
+	// Just a smoke test for the default path.
+	cfg, layers, err := EncryptConfig(nil, nil)
+	assert.NoError(t, err)
+	assert.Nil(t, cfg)
+	assert.Nil(t, layers)
+}
+
 func TestGetFormat(t *testing.T) {
 	_, err := GetFormat("bogus")
 	assert.NotNil(t, err)


mtrmac:
Cc: @lumjjb

flouthoc (changes requested):
LGTM , this needs `[NO NEW TESTS NEEDED]`

mtrmac:
Note https://github.com/containers/buildah/pull/4746 as an alternative implementation approach.

---

> this sounds like a breaking change. If you now require the option explicitly this can break existing users that depend on it.

Yes… but it was also, to my knowledge, never documented in Buildah to work this way. I don’t know for sure whether it was _intended_ to work this way, or documented elsewhere to work this way.

I’ll defer to @lumjjb on that.

---

> We should avoid such changes at all costs.

I don’t think we have a general obligation to be bug-for-bug-compatible, and I would fairly strongly push against being bug-for-bug-compatible “at all costs”.

Still, it’s a valid concern. I can imagine changing `podman pull` to auto-decrypt instead — although it feels unintuitive and vaguely possibly risky to me.

---

Most of the diagnosis and discussion has been in https://github.com/containers/podman/issues/18196 , and there are two Buildah PRs in flight.

Let’s concentrate the design discussion in the Podman https://github.com/containers/podman/issues/18196 issue, for now.

mtrmac:
> LGTM , this needs `[NO NEW TESTS NEEDED]`

Thanks for pointing that out.

Actually I would expect some of the existing tests to fail with a different error message, so bypassing that for now to see the real failures…

rhatdan:
I am fine with making them consistent since buildah vendored into Podman means podman build works differently then podman pull.

TomSweeneyRedHat:
Anyway to add a warning?  I.e. if the environment points at keys, but the option is not specified, then print a warning message?  IDK for sure that we would want to do that, but just wondering.

mtrmac:
> Anyway to add a warning? I.e. if the environment points at keys, but the option is not specified, then print a warning message? IDK for sure that we would want to do that, but just wondering.

If the image isn’t encrypted, we probably wouldn’t want to warn:
- Right now, we “enable” encryption for all `buildah from` cases, only it does nothing for unencrypted images
- And to avoid that warning, users would probably need to unset the environment variable… which is a hassle.

OTOH if the image is encrypted and we no longer decrypt by default, `buildah from` will fail (see the test suite updates in this PR), so we don’t need a _warning_. The error message in that case could certainly be improved — but that’s not this PR; it would probably be added in c/image.

tests/from.bats
```diff
@@ -427,8 +427,8 @@ load helpers
   run_buildah push $WITH_POLICY_JSON --tls-verify=false --creds testuser:testpassword --encryption-key jwe:${TEST_SCRATCH_DIR}/tmp/mykey.pub busybox oci:${TEST_SCRATCH_DIR}/tmp/busybox_enc
 
   # Try encrypted image without key should fail
-  run_buildah 125 from oci:${TEST_SCRATCH_DIR}/tmp/busybox_enc
-  expect_output --substring "decrypting layer .* missing private key needed for decryption"
+  run_buildah 1 from oci:${TEST_SCRATCH_DIR}/tmp/busybox_enc
```
mtrmac, line 430: This exit code changes because the error stack, ultimately, contains an `exec.ExitError` from invoking an untar subprocess; and the top level of Buildah checks for `exec.ExitError`, and returns that exit code.

One can read that as a failure of the Go error wrapping design… or, alternatively, as Builder’s `main` matching too loosely if it wants only specific subprocesses’ exit codes to be returned unmodified.

AFAICS the exit code from Buildah is, apart from `buildah manifest exists`, not documented, so this PR just changes tests to accept the status 1. If we do actually commit (whether in documentation in or in practice) to returning 125 on unspecific failures, and this needs to be made more precise, please let me know (and also define which subprocess invocations that could fail with `ExitError` _should_ be propagated.)

rhatdan:
We only document the exit codes for the run commands, I believe.

The question is do we take this PR and leave multiple implementation of the same code, or take mine and vendor the code into Podman.

mtrmac:
I think the _primary_ question is whether we want this; as @Luap99 points out, this _might_ be a compatibility break.

If we do want this, I completely agree that sharing the code, per your PR, is the better approach — consider using the test updates from here, to make it pass.

rhatdan:
I don't see this as an important breaking change, since having it two different ways within Podman is more important. 
LGTM
I think we can merge this and then I will update my PR to pull this in.

mtrmac:
(I think the consensus is to prefer #4746 … except that it doesn’t build.)

mtrmac:
#4746 has implemented the primary (and better) fix; this now just carries the remaining unit tests.

PR #4746, opened 2023-04-18
Change util.EncryptConfig,DecryptConfig, GetFormat and make it public

rhatdan:
We want to share these functions with Podman, Podman currently has a slightly different version which is correct, so use correct version in Buildah and vendor it into Podman.

Fixing: https://github.com/containers/podman/issues/18196

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```


rhatdan:
This is an alternative to https://github.com/containers/buildah/pull/4745

rhatdan:
Once this merges we can vendor it into Podman and then share the code.

mtrmac:
Sharing that would be nice, but let’s make sure we have a consensus on what the right thing to do is.

Most of the conversation is in https://github.com/containers/podman/issues/18196 .

mtrmac:
(A “make … public” PR that actually, _primarily_, makes a behavior change, would be hard to track down later, so renaming it.)

pkg/cli/build.go
```diff
@@ -272,8 +271,8 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
 		return options, nil, nil, err
 	}
 
-	decryptConfig, err := iutil.DecryptConfig(iopts.DecryptionKeys)
-	if err != nil {
+	decryptConfig, err := util.DecryptConfig(iopts.DecryptionKeys)
+	if err != nil || decryptConfig == nil {
```
mtrmac, line 275: `decryptConfig == nil` means “don’t decrypt”, and it’s what I propose we do by default; failing on that doesn’t make sense to me.

tests/pull.bats
```diff
@@ -190,7 +190,7 @@ load helpers
   run_buildah push $WITH_POLICY_JSON --encryption-key jwe:${TEST_SCRATCH_DIR}/tmp/mykey.pub busybox  oci:${TEST_SCRATCH_DIR}/tmp/busybox_enc
 
   # Try to pull encrypted image without key should fail
-  run_buildah 125 pull $WITH_POLICY_JSON oci:${TEST_SCRATCH_DIR}/tmp/busybox_enc
+  run_buildah 1 pull $WITH_POLICY_JSON oci:${TEST_SCRATCH_DIR}/tmp/busybox_enc
```
TomSweeneyRedHat, line 193: Can we add tests to these when a key isn't available so it can still test a failure path?

rhatdan: This is testing a failure path.

pkg/util/util.go
```diff
@@ -77,3 +81,49 @@ func DiscoverContainerfile(path string) (foundCtrFile string, err error) {
...
+			return nil, nil, fmt.Errorf("invalid encryption keys: %w", err)
+		}
+		cc := encconfig.CombineCryptoConfigs([]encconfig.CryptoConfig{ecc})
+		encConfig = cc.EncryptConfig
+	}
+	return encConfig, encLayers, nil
+}
+
+// GetFormat translates format string into either docker or OCI format constant
+func GetFormat(format string) (string, error) {
```
mtrmac, line 120: Should this be a part of the move at all?

mtrmac (commented):
Please carry over the test suite updates from https://github.com/containers/buildah/pull/4745 to make this pass.

mtrmac:
@rhatdan this is failing tests.

rhatdan:
@mtrmac PTAL I think I followed all of your instructions.

mtrmac (commented):
/lgtm

edsantiago:
This is [breaking podman](https://api.cirrus-ci.com/v1/artifact/task/6125041562157056/html/bud-podman-fedora-38-root-host-boltdb.log.html):
```
[+1068s] not ok 269 bud with encrypted FROM image
...
# #| FAIL: exit code is 125; expected 1
```
And also, the podman message is still "missing private key", whereas the changed test now expects "invalid tar header"

Issue #4756, opened 2023-04-14
Improve first time experience with `podman build --platform` argument


aren55555:
### Feature request description

I recently migrated my personal websites from `docker` to `podman` as I personally didn't like the direction that Docker (the company) was pursuing. Keep in mind I'm a complete beginner when it comes to `podman` but I had used `docker` up until a few years ago.

I had a previous build script that invoked `docker build` with the argument `--platform=amd64`. 

After aliasing podman within my shell `alias docker=podman` I re-ran my build script and received the error:

```
Error: invalid platform syntax for "amd64" (use OS/ARCH[/VARIANT][,...])
```

Granted, I haven't rebuilt my site's container in quite some time (at least a few years ago) but this did once work with `docker`. 

I was able to quickly find Podman's [documentation page for the `--platform` argument](https://docs.podman.io/en/latest/markdown/podman-build.1.html#platform-os-arch-variant) (love that I can direct anchor link to this) and was able to correct my build script to supply the correct `--platform=linux/amd64` argument.

Then I tried to find Docker's documentation for its `--platform` argument. I couldn't find anything [but this](https://docs.docker.com/engine/reference/commandline/build/) (not useful). So now I'm not sure if this ever worked.

I'm not sure how similar (or not) these APIs are meant to be. If podman is intended to be a drop-in replacement it would be nice to improve this first problem encounter experience. Others are likely to have this problem and may give up on podman. 

### Suggest potential solution

I'm not sure how similar (or not) these APIs are meant to be. If podman is intended to be a drop-in replacement I have the following suggestions:

1. Make podman's `invalid platform syntax ` error message more verbose regarding this problem:
  * Perhaps it could include a link to the [documentation for the flag](https://docs.podman.io/en/latest/markdown/podman-build.1.html#platform-os-arch-variant)?
  * Could it list concrete examples -- ie `linux/amd64`, etc
1. (if there isn't one already) documentation describing the specific argument differences between podman and docker

### Have you considered any alternatives?

N/A - not reverting back to `docker build`

### Additional context

As a side note: this was the only change I had to make to move my container build from `docker` to `podman`. Great work.

flouthoc:
@aren55555 In most of the commands usually error message dont point to `man pages` but I have made error a bit cleaner if that helps. https://github.com/containers/buildah/pull/4757

PR #4757, opened 2023-04-25
 build: `--platform` must accept only arch

flouthoc:
Make error message for `--platform` more clear.

Closes: https://github.com/containers/podman/issues/18194
Closes: https://github.com/containers/buildah/issues/4756

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
 build: `--platform` must accept only arch
```

### DIFF
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -510,8 +510,6 @@ func PlatformsFromOptions(c *cobra.Command) (platforms []struct{ OS, Arch, Varia
 	return platforms, nil
 }
 
-const platformSep = "/"
-
 // DefaultPlatform returns the standard platform for the current system
 func DefaultPlatform() string {
 	return platforms.DefaultString()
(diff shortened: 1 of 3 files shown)


nalind:
Perhaps it would be more helpful to have `parse.Platform()` call `platforms.Parse()` (for cases other than "local") so that we actually accept values like "amd64".

flouthoc:
Created PR for conformance test: https://github.com/containers/buildah/pull/4760

pkg/parse/parse.go
```diff
@@ -520,21 +518,24 @@ func DefaultPlatform() string {
...
+	if platform[len(platform)-1] == '/' {
+		// If --platform string has format as `some/plat/string/`
+		// make it `some/plat/string`
+		platform = platform[:len(platform)-1]
+	}
+	if platform[0] == '/' {
+		// If --platform string has format as `/some/plat/string`
+		// make it `some/plat/string`
+		platform = platform[1:]
+	}
```
nalind, line 533: Is this a thing that people do?

nalind: Oh.  Ugh.  Kind of wish I hadn't signed off on that.
Use `strings.Trim()` instead, it'll be much shorter.

flouthoc: fixed.

Issue #4763, opened 2023-04-27
4757 broke podman-remote: failed to parse query parameter 'platform'


edsantiago:
PR #4757 broke podman-remote:
```console
$ echo "FROM sfsdf" | bin/podman-remote build  -
Error: failed to parse query parameter 'platform': "/": invalid platform syntax for --platform="": "" is an invalid component of "": platform specifier component must match "^[A-Za-z0-9_-]+$": invalid argument
```
No `podman-remote build` commands work at all any more. [log](https://api.cirrus-ci.com/v1/artifact/task/6515049724903424/html/bud-remote-fedora-37-root-host-boltdb.log.html)

I don't know if this needs to be fixed on the podman side or buildah side. Please reassign issue as needed.

flouthoc:
Uh-oh let me fix it for `podman-remote`, thanks for reporting @edsantiago

flouthoc:
This actually needs to be also fixed at `podman-remote` end , it sends `/` on no `--platform` argument.

PR #4765, opened 2023-04-28
parse: filter edge-case for podman-remote

flouthoc:
podman-remote has a bug where it sends `/` when no platform is set, patch is there to make sure older clients still work if backend is updated.

Closes: https://github.com/containers/buildah/issues/4763

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
parse: filter edge-case for podman-remote
```

### DIFF
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -518,7 +518,7 @@ func DefaultPlatform() string {
 // Platform separates the platform string into os, arch and variant,
 // accepting any of $arch, $os/$arch, or $os/$arch/$variant.
 func Platform(platform string) (os, arch, variant string, err error) {
-	if platform == "local" {
+	if platform == "local" || platform == "" || platform == "/" {
 		return Platform(DefaultPlatform())
 	}
 	if platform[len(platform)-1] == '/' || platform[0] == '/' {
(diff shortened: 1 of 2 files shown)


rhatdan:
/approve
LGTM

rhatdan:

pkg/parse/parse.go
```diff
@@ -518,7 +518,7 @@ func DefaultPlatform() string {
 // Platform separates the platform string into os, arch and variant,
 // accepting any of $arch, $os/$arch, or $os/$arch/$variant.
 func Platform(platform string) (os, arch, variant string, err error) {
-	if platform == "local" {
+	if platform == "local" || platform == "" || platform == "/" {
```
TomSweeneyRedHat, line 521: It might be nice to have a debug statement in here just to let people know the error triggered and the value found in the platform variable.

flouthoc: For these cases it will always return `default` without any errors but for any error case scenario it retruns error with the value used in `--platform` ( Line: 531 )

flouthoc:
Rebased.

edsantiago:
Unfortunate to have to special-case it here, but LGTM in principle. I have not actually run podman-buildah-treadmill tests against this.

PR #4770, opened 2023-05-02
[release-1.27] Cirrus: Add CI self-destruct condition on EOL date

cevich:
#### What type of PR is this?

/kind other

This branch will never receive any security-backports when the associated RHEL release reaches EOL.  Add a condition to force CI to break with a helpful message, after this RHEL EOL date.

#### How to verify it

CI will run/pass now, but not after 06-01-2023

Repeatedly checking if a release-branch needs to be supported.

```release-note
None
```

### DIFF
--- a/contrib/cirrus/setup.sh
+++ b/contrib/cirrus/setup.sh
@@ -10,6 +10,17 @@ source $(dirname $0)/lib.sh
 
 req_env_vars OS_RELEASE_ID OS_RELEASE_VER GOSRC IN_PODMAN_IMAGE
 
+_EOL=20230601
+if [[ $(date +%Y%m%d) -ge $_EOL ]]; then
+    die "As of $_EOL this branch is probably
+no longer supported in RHEL 8.7/9.1, please
+confirm this with RHEL Program Management.  If so:
+It should be removed from Cirrus-Cron,
+the .cirrus.yml file removed, and
+the VM images (manually) unmarked
+'permanent=true'"
+fi
+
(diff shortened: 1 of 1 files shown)


contrib/cirrus/setup.sh
```diff
@@ -10,6 +10,17 @@ source $(dirname $0)/lib.sh
 
 req_env_vars OS_RELEASE_ID OS_RELEASE_VER GOSRC IN_PODMAN_IMAGE
 
+_EOL=20230601
+if [[ $(date +%Y%m%d) -ge $_EOL ]]; then
+    die "As of $_EOL this branch is probably
+no longer supported in RHEL 8.7/9.1, please
+confirm this with RHEL PM.  If so:
```
TomSweeneyRedHat, line 17: I'd suggest "Program Manager" over "PM" as that could also be "Product Manager" and they generally don't always know the RHEL deadlines like program managers do.  Currently that is Laurie Friedman

cevich: Hmmm, darn.  I'm not disagreeing...it's just that this is the wording already in-place for a few podman branches.  I just copy-pasted the message :frowning_face: 

I'll fix it here and https://github.com/containers/buildah/pull/4771 but I'd rather not track down and fix the already-merged podman PRs - simply out of laziness :grin:

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
And happy green test buttons.  @edsantiago could you take a peak please?

PR #4790, opened 2023-05-11
chroot, run: drop Pdeathsig

giuseppe:
it is not safe to use prctl(PR_SET_PDEATHSIG) in a Go program since the parent of the process is considered the *thread* that created the
child process, and the Go runtime can free a thread at any time.   We
would either need to lock the thread, or simpler just not use it.

Closes: https://github.com/containers/podman/issues/17918

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

/kind bug

```release-note
Do not use `prctl(PR_SET_PDEATHSIG)` in a Go program since it can lead to weird errors
```

### DIFF
--- a/chroot/run_common.go
+++ b/chroot/run_common.go
@@ -501,6 +501,10 @@ func runUsingChroot(spec *specs.Spec, bundlePath string, ctty *os.File, stdin io
 	// Apologize for the namespace configuration that we're about to ignore.
 	logNamespaceDiagnostics(spec)
 
+	// We need to lock the thread so that PR_SET_PDEATHSIG won't trigger if the current thread exits.
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
 	// Start the parent subprocess.
 	cmd := unshare.Command(append([]string{runUsingChrootExecCommand}, spec.Process.Args...)...)
 	setPdeathsig(cmd.Cmd)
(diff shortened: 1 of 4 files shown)


rhatdan:
@nalind PTAL

rhatdan:
LGTM

TomSweeneyRedHat:
Doug Rabson @dfr added this for his BSD purposes.  I'd like him to chime in on this too before we merge.

dfr:
I mainly added the FreeBSD pdeathsig handling for consistency with Linux. If this is wrong for Linux, then its almost certainly also wrong for FreeBSD.

TomSweeneyRedHat:
LGTM
but definitely want @nalind  to scope it out, plus some of the tests aren't hip.

nalind:
So if I'm using chroot isolation to handle a RUN instruction or the like, and buildah gets a ctrl-c, what happens to the child processes after this change?

nalind:
> So if I'm using chroot isolation to handle a RUN instruction or the like, and buildah gets a ctrl-c

Or the uncatchable SIGKILL, I suppose.

giuseppe:
> So if I'm using chroot isolation to handle a RUN instruction or the like, and buildah gets a ctrl-c, what happens to the child processes after this change?

Good point.  I've restored it for the `chroot` package and added a new patch to make sure the current thread is locked (so that it won't be released while we wait for the command).

giuseppe:
@nalind are you fine with the last version?

nalind:
It should be fine. When we use an external runtime, we don't directly create the processes that we intend should be killed off if we are, and there's no runtime config field that would let us specify that, so this is probably as good as we can expect.

giuseppe:
can we merge?

PR #4791, opened 2023-05-15
secret: accept `required` flag without value

flouthoc:
`Buildkit` and `docker` accepts `required` flag as `true` when provided without any value in `--mount=type=secret` so lets do same for buildah.

Example usage

```Dockerfile
FROM docker.io/nginx:1.23.3

RUN --mount=type=secret,id=nginx-crt,dst=/nginx-repo.crt,required \
    --mount=type=secret,id=nginx-key,dst=/nginx-repo.key,required \
    set -x \
    && . /etc/os-release \
    && stat /nginx-repo.crt \
    && stat /nginx-repo.key
```

Closes: https://github.com/containers/podman/issues/18438

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
secret: accept `required` flag without value
```

### DIFF
--- a/run_common.go
+++ b/run_common.go
@@ -1653,9 +1653,12 @@ func (b *Builder) getSecretMount(tokens []string, secrets map[string]define.Secr
 				target = filepath.Join(workdir, target)
 			}
 		case "required":
-			required, err = strconv.ParseBool(kv[1])
-			if err != nil {
-				return nil, "", errInvalidSyntax
+			required = true
+			if len(kv) > 1 {
+				required, err = strconv.ParseBool(kv[1])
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM
Do we need a test with required=false?

TomSweeneyRedHat:
I don't think we do, and it would be a good addition.
Otherwise, LGTM

flouthoc:
Okay added a case for missing `required=false` in a new commit.

PR #4813, opened 2023-05-22
Go back to setting SysProcAttr.Pdeathsig for child processes

nalind:
#### What type of PR is this?

/kind bug

Restore SysProcAttr.Pdeathsig values for child processes that we start, so that they get a SIGKILL when we exit for whatever reason.  Lock the calling goroutine to a native thread while that's happening, like we should always have done.

#### How to verify it

New integration tests!

Should fix https://api.cirrus-ci.com/v1/artifact/task/5262176693256192/html/sys-podman-fedora-38-root-host-boltdb.log.html#t--00120.

```release-note
None
```


giuseppe (approved):
I agree this is better

LGTM

flouthoc (approved):
LGTM

Issue #4804, opened 2023-05-18
labels are still being applied to intermediate images


akanix42:
**Description**

https://github.com/containers/buildah/pull/4673 applies a fix to only add labels (as provided by `--label`) to the final image; however, it is currently not working, so the label is incorrectly added to all images in the chain. After looking over the code, I _think_ (I'm not set up to develop buildah so I haven't tested this) that this block https://github.com/containers/buildah/blob/main/imagebuildah/executor.go#L474-L501 needs to be nested in an `if` like the one right above it so that it only applies to the final stage: https://github.com/containers/buildah/blob/main/imagebuildah/executor.go#L471-L473

The test as written doesn't actually test the issue - the hash of the COPY instruction doesn't change regardless of the label being added, so the cache isn't invalidated, but the label is still being added. The issue quickly becomes apparent when you inherit from the image rather than referencing it for a COPY (for example, run the test with [Dockerfile.reused](https://github.com/containers/buildah/blob/main/tests/bud/multi-stage-builds/Dockerfile.reused) instead).

**Steps to reproduce the issue:**
1. Create a simple Dockerfile
```dockerfile
FROM alpine:3.18.0 as stage1
RUN echo stage1!

FROM stage1 as stage2
RUN echo stage2!
```
2. Run buildah with a sample label. The label is applied at the end of both stages.
```
buildah build --layers --label 'foo=abc' -t podman-issue .
[1/2] STEP 1/3: FROM alpine:3.18.0 AS stage1
[1/2] STEP 2/3: RUN echo stage1!
stage1!
--> 333078e6b464
[1/2] STEP 3/3: LABEL "foo"="abc"
--> 764249bb2f05
[2/2] STEP 1/3: FROM 764249bb2f05051e10095df75c06f33149802275aaa55b838c12ff40393a2944 AS stage2
[2/2] STEP 2/3: RUN echo stage2!
stage2!
--> f7b373419016
[2/2] STEP 3/3: LABEL "foo"="abc"
[2/2] COMMIT podman-issue
--> 584682d27f96
Successfully tagged localhost/podman-issue:latest
584682d27f96c4335ad72322f65b0378b6dc9e7e3bb673fdfafcfb78b96f01ea
```
3.  Run the same buildah command again. The layers are cached.
```
buildah build --layers --label 'foo=abc' -t podman-issue .
[1/2] STEP 1/3: FROM alpine:3.18.0 AS stage1
[1/2] STEP 2/3: RUN echo stage1!
[...]

PR #4817, opened 2023-05-26
executor: apply label to only final stage via `--label`

flouthoc:
In https://github.com/containers/buildah/pull/4673 we made a change were we were applying labels to end of each stage, which is different than what we were doing before i.e applying label at the end of the each step.

However buildkit does not adds label to any stage or steps it only adds label at the end of final stage so lets do that.

Closes: https://github.com/containers/buildah/issues/4804

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
executor: apply label to only final stage
```

### DIFF
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -470,33 +470,33 @@ func (b *Executor) buildStage(ctx context.Context, cleanupStages map[int]*StageE
 	output := ""
 	if stageIndex == len(stages)-1 {
 		output = b.output
-	}
-	// Check if any labels were passed in via the API, and add a final line
-	// to the Dockerfile that would provide the same result.
-	// Reason: Docker adds label modification as a last step which can be
-	// processed like regular steps, and if no modification is done to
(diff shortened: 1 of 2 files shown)


flouthoc:
Don't this conformance failures are related to this PR 
```
Error:      	Expected nil, but got: &jsonmessage.JSONError{Code:0, Message:"Get \"https://registry.centos.org/v2/\": dial tcp: lookup registry.centos.org: no such host"}
```

rhatdan:
LGTM

TomSweeneyRedHat:
Looks like #4818 is also having conformance test issues.  Ping @edsantiago

TomSweeneyRedHat:
a nitty comment suggestion, otherwise LGTM once the conformance tests are righted.

flouthoc:
Rebased after https://github.com/containers/buildah/pull/4819

PR #4819, opened 2023-05-26
remove `registry.centos.org` since its decommissioned

flouthoc:
registry.centos.org is now decommissioned

Ref: https://lists.centos.org/pipermail/centos-devel/2023-May/142956.html

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### Does this PR introduce a user-facing change?

```release-note
remove `registry.centos.org` since its decommissioned
```

PR #4827, opened 2023-05-30
[release-1.24] remove registry.centos.org

edsantiago:
registry.centos.org is now decommissioned

Ref: https://lists.centos.org/pipermail/centos-devel/2023-May/142956.html

Cherry-pick of #4825, which is in turn a cherry-pick of #4819. One minor conflict

Signed-off-by: Aditya R <arajan@redhat.com>

```release-note
None
```

### DIFF
--- a/install.md
+++ b/install.md
@@ -376,7 +376,7 @@ cat /etc/containers/registries.conf
 # and 'registries.block'.
 
 [registries.search]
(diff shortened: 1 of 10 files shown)


cevich (changes requested):
This is going to flake like crazy due to the pull-limit on the docker hub.  We need to use quay.io/centos/centos:centos7

cevich:
Opened https://github.com/containers/buildah/pull/4828

cevich:
I think this needs https://github.com/containers/buildah/pull/4828

PR #4828, opened 2023-05-30
Explicitly ref. quay images for CI

cevich:
#### What type of PR is this?

/kind other

Commit https://github.com/cevich/buildah/commit/beae5647c075353365a4cc25c81dbd6784186f16 updated the conformance test references to a
deprecated repository.  However, by pointing at the docker-hub it
inadvertantly introduce a significant possibility for flakes.  This is
because anonymous docker hub access is rate-limited by IP.  We cannot
predict the IP used for CI VMs & Containers, any of which could be at or
close to the limit.  Fix this by pointing explicitly at a
`quay.io/libpod/centos` repo. which is excluesively for use by CI, with
static images.

#### How to verify it

Conformance testing in CI will pass

Ref: https://github.com/containers/buildah/pull/4819

```release-note
None
```

### DIFF
--- a/tests/conformance/testdata/Dockerfile.edgecases
+++ b/tests/conformance/testdata/Dockerfile.edgecases
@@ -1,3 +1,4 @@
+# Note: a registries.conf alias should redirect this to quay.io/libpod/busybox
 FROM busybox
 
 MAINTAINER docker <docker@docker.io>
--- a/tests/conformance/README.md
+++ b/tests/conformance/README.md
@@ -20,7 +20,7 @@ First, pull base images used by various conformance tests:
 bash
 docker pull alpine
 docker pull busybox
-docker pull docker.io/library/centos:centos7
+docker pull quay.io/libpod/centos:7
 ```
 
 Then you can run all of the tests with go test:
--- a/tests/conformance/testdata/Dockerfile.reusebase
+++ b/tests/conformance/testdata/Dockerfile.reusebase
@@ -1,4 +1,4 @@
-FROM docker.io/library/centos:centos7 AS base
+FROM quay.io/libpod/centos:7 AS base
 RUN touch -t 201701261659.13 /1
 ENV LOCAL=/1
 
--- a/tests/conformance/testdata/Dockerfile.shell
+++ b/tests/conformance/testdata/Dockerfile.shell
@@ -1,3 +1,3 @@
-FROM docker.io/library/centos:centos7
+FROM quay.io/libpod/centos:7
 SHELL ["/bin/bash", "-xc"]
 RUN env
--- a/tests/conformance/testdata/copy/Dockerfile
+++ b/tests/conformance/testdata/copy/Dockerfile
@@ -1,3 +1,3 @@
-FROM docker.io/library/centos:centos7
+FROM quay.io/libpod/centos:7
 COPY script /usr/bin
 RUN ls -al /usr/bin/script
--- a/tests/conformance/testdata/copydir/Dockerfile
+++ b/tests/conformance/testdata/copydir/Dockerfile
@@ -1,3 +1,3 @@
-FROM docker.io/library/centos:centos7
+FROM quay.io/libpod/centos:7
 COPY dir /dir
 RUN ls -al /dir/file
--- a/tests/conformance/testdata/copyrename/Dockerfile
+++ b/tests/conformance/testdata/copyrename/Dockerfile
@@ -1,3 +1,3 @@
-FROM docker.io/library/centos:centos7
+FROM quay.io/libpod/centos:7
 COPY file1 /usr/bin/file2
 RUN ls -al /usr/bin/file2 && ! ls -al /usr/bin/file1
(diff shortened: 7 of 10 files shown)


flouthoc (approved):
LGTM

edsantiago:
LGTM at first glance, but might as well fix the instances in `from.bats` also?

cevich:
> LGTM at first glance, but might as well fix the instances in from.bats also?

Yes.  I was in a bit of a panic to open this but found in `contrib/cirrus/setup.sh` that we are setting up a proxy to redirect to quay.  That will cover this case, so there's really not a huge rush to get this PR in.

I'll add a commit to fix up references in bats.  Thanks for the suggestion.

nalind:
How does this overlap with mirroring in `tests/registries.conf`?

cevich:
> How does this overlap with mirroring in tests/registries.conf?

That file is put in place from `contrib/cirrus/setup.sh`.  So that is what makes the current docker-hub references work fine in CI (including for conformance test).  However, this is mostly due to luck: Where "library/centos/centos" in the docker hub happens to translate properly to the "centos/centos" namespace in quay.  I think it's desirable to update all the references to point at the quay images so we're not relying on luck.

cevich:
Hmmm...thinking about this more, **we** maintain `quay.io/libpod/centos:7` whereas somebody else runs `centos/centos:7`.  There's a risk the later could change unexpectedly, but the former is treated (by practice) as "untouchable" for use in CI.  So probably I should instead update all the references to `quay.io/libpod/centos:7` to maintain the current behavior but reduce the "magic"?

nalind:
> That file is put in place from `contrib/cirrus/setup.sh`. So that is what makes the current docker-hub references work fine in CI (including for conformance test).

That file exists in the repository, so we shouldn't need to be overwriting it for CI.  If CI is changing it, we need to update the copy in the repository, so that integration tests that aim to be usable outside of CI don't behave differently in hard-to-find ways when they're run outside of CI.

There's also a case for setting CONTAINERS_REGISTRIES_CONF to point to the same file for unit and conformance tests, which I don't think we currently do, but it's worth remembering that when the conformance tests build an image using dockerd, it's still almost certainly going to be ignored.

cevich:
> don't behave differently in hard-to-find ways when they're run outside of CI

Yes, I agree.  CI doesn't change the file.  I looked through the scripts and Makefile and it looks like for the most part we reference the `tests/registries.conf` file explicitly.  Except...

> conformance tests build an image using dockerd, it's still almost certainly going to be ignored

Right, excellent point - conformance tests use docker, not podman.  So we do need the the conformance tests pointing explicitly at `quay.io/libpod/centos:7` in all those `Dockerfiles` on `main` and in all the release branches where CI is running.  Ugg :disappointed:  

I'm guessing there's no docker-equivalent for `tests/registries.conf`, is there?

cevich:
Woops, force-push: Fixed missed conformance test update.

tests/helpers.bash
```diff
@@ -130,9 +130,9 @@ function teardown_tests() {
 function normalize_image_name() {
     for img in "$@"; do
         if [[ "${img##*/}" == "$img" ]] ; then
-            echo -n docker.io/library/"$img"
-        elif [[ docker.io/"${img##*/}" == "$img" ]] ; then
-            echo -n docker.io/library/"${img##*/}"
+            echo -n quay.io/libpod/"$img"
+        elif [[ quay.io/libpod/"${img##*/}" == "$img" ]] ; then
+            echo -n quay.io/libpod"${img##*/}"
```
cevich, line 135: @edsantiago I think maybe [these test-failures](https://cirrus-ci.com/build/4628220696330240) are caused by my changes here.  Should I have done:

```suggestion
            echo -n quay.io/libpod/"${img##*/}"
```

?

nalind: If I'm reading it right, can't this clause be dropped?

cevich: Ed seems busy, let me just try removing it...

cevich: Ed seems busy, let me just try removing it...

edsantiago: No, I just have nothing to add. This is not my code, I don't even understand what it does or is supposed to do. I stand by anything @nalind says about it.

cevich: Oh okay, thanks for the update Ed.

TomSweeneyRedHat:
LGTM
but the tests are not at all happy anywhere.

cevich:
Sigh...I think maybe I need to break this PR up since the first commit (conformance test) fixes a problem occurring on CI for basically every release branch.

tests/conformance/testdata/Dockerfile.edgecases
```diff
@@ -1,3 +1,4 @@
+# Note: a registries.conf alias should redirect this to quay.io/libpod/busybox
```
edsantiago, line 1: The word "should" is inherently ambiguous and should not be used in documentation (see what I did there?) In this instance, it's not clear whether you mean "we hope this happens", or "we know this will happen because there is an existing registries.conf mapping", or "we should some day set up a registries.conf mapping". This is fine to fix in the followup PR.

edsantiago (approved):
LGTM with two nits: one, I'd be happier if the commit title explicitly included "conformance tests" to make it clear that this doesn't touch bats tests; and second, below. Neither of them is worth blocking for, I'm just commenting in the unlikely event that you need to repush.

PR #4830, opened 2023-05-30
CI: Support testing w/ podman-next COPR packages

cevich:
#### What type of PR is this?
/kind other

Sometimes important updates need to be made to dependent packages and
run through CI w/o waiting for package release and new CI VM image
builds.  Support this in buildah CI as in podman CI, by updating
packages during setup when the magic string is present and PR is in
draft-mode.

Note: To support containerized testing, both `CIRRUS_CHANGE_TITLE`
and `CIRRUS_PR_DRAFT` env. vars. are passed through.  For these tasks,
this will result in **TWO** updates - One for the host, and another one
in the container.

#### How to verify it

CI will pass with and without magic keyword in title.

[CI Run when `[CI:NEXT]` was in the PR title.](https://cirrus-ci.com/build/6633729905393664)  I believe the handful of SELinux related test failures are a known issue.

```release-note
None
```

### DIFF
--- a/contrib/cirrus/lib.sh
+++ b/contrib/cirrus/lib.sh
@@ -162,10 +162,11 @@ in_podman() {
     local envfile=$(mktemp -p '' in_podman_env_tmp_XXXXXXXX)
     trap "rm -f $envfile" EXIT
 
+    # TODO: This env. var. processing is fugly and fragile.  Refactor
+    # this closer to how podman CI does it using passthrough_envars()
     msg "Gathering env. vars. to pass-through into container."
     for envname in $(awk 'BEGIN{for(v in ENVIRON) print v}' | sort | \
-                     egrep "$envrx" | egrep -v "$SECRET_ENV_RE" | \
-                     egrep -v "^CIRRUS_.+(MESSAGE|TITLE)")
+                     egrep "$envrx" | egrep -v "$SECRET_ENV_RE");
     do
         envval="${!envname}"
         [[ -n $(tr -d "$xchars" <<<"$envval") ]] || continue
--- a/contrib/cirrus/setup.sh
+++ b/contrib/cirrus/setup.sh
@@ -8,13 +8,28 @@ set -e
 # expectation mismatch.
 source $(dirname $0)/lib.sh
 
-req_env_vars OS_RELEASE_ID OS_RELEASE_VER GOSRC IN_PODMAN_IMAGE
+req_env_vars OS_RELEASE_ID OS_RELEASE_VER GOSRC IN_PODMAN_IMAGE CIRRUS_CHANGE_TITLE
 
 msg "Disabling git repository owner-check system-wide."
 # Newer versions of git bark if repo. files are unexpectedly owned.
 # This mainly affects rootless and containerized testing.  But
 # the testing environment is disposable, so we don't care.=
 git config --system --add safe.directory $GOSRC
+
+# Support optional/draft testing using latest/greatest
+# podman-next COPR packages.  This requires a draft PR
+# to ensure changes also pass CI w/o package updates.
+if [[ "$OS_RELEASE_ID" =~ "fedora" ]] && \
+   [[ "$CIRRUS_CHANGE_TITLE" =~ CI:NEXT ]]
+then
+    # shellcheck disable=SC2154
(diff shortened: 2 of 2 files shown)


rhatdan:
LGTM
@edsantiago PTAL

contrib/cirrus/setup.sh
```diff
@@ -8,14 +8,29 @@ set -e
...
 msg "Disabling git repository owner-check system-wide."
 # Newer versions of git bark if repo. files are unexpectedly owned.
 # This mainly affects rootless and containerized testing.  But
 # the testing environment is disposable, so we don't care.=
 git config --system --add safe.directory $GOSRC
 
+# Support optional/draft testing using latest/greatest
+# podman-next COPR packages.  This requires a draft PR
+# to ensure changes also pass CI w/o package updates.
+if [[ ! "$OS_RELEASE_ID" =~ "debian" ]] && \
```
edsantiago, line 22: Negative checks are so hard to read; and, in this case, wrong: dnf is a fedora-only thing, not a "not-debian" thing. (I.e., if we ever switch back to ubuntu, or add some other trendy new distro, this will fail). Would you mind changing this to a positive `= "fedora"` check?

contrib/cirrus/lib.sh
```diff
@@ -164,8 +164,7 @@ in_podman() {
 
     msg "Gathering env. vars. to pass-through into container."
     for envname in $(awk 'BEGIN{for(v in ENVIRON) print v}' | sort | \
-                     egrep "$envrx" | egrep -v "$SECRET_ENV_RE" | \
-                     egrep -v "^CIRRUS_.+(MESSAGE|TITLE)")
+                     egrep "$envrx" | egrep -v "$SECRET_ENV_RE");
```
edsantiago, line 167: This removes the `CIRRUS_MESSAGE` filtering. I tried to follow the commit history, to understand why that was believed necessary, but can't figure it out. Do you have any recollection of why that was added, and whether it is your express intention to remove it here?

contrib/cirrus/setup.sh
```diff
@@ -8,14 +8,29 @@ set -e
 # expectation mismatch.
 source $(dirname $0)/lib.sh
 
-req_env_vars OS_RELEASE_ID OS_RELEASE_VER GOSRC IN_PODMAN_IMAGE
+req_env_vars OS_RELEASE_ID OS_RELEASE_VER GOSRC IN_PODMAN_IMAGE CIRRUS_CHANGE_TITLE CIRRUS_PR_DRAFT
```
edsantiago, line 11: [Cirrus documentation](https://cirrus-ci.org/guide/writing-tasks/#environment-variables) does not define this variable in the non-true state. That means it could be `false`, or could be empty, and if empty, `req_env_vars` will fail. (This PR is not-draft, and CI passed, so I'm guessing `CIRRUS_PR_DRAFT` is non-empty, but can we rely on that?)

edsantiago (changes requested):
Big picture LGTM, with a few concerns. 

This deserves to be documented, but I can't even find where `CI:DOCS` or `:BUILD` are documented...?

contrib/cirrus/lib.sh
cevich: It's a bad design, predating the refactoring work you did in podman.  It was ignored previously out of caution - it get's passed into the containers via an `--env-file` which doesn't handle multi-line strings (or the raw-string bash syntax `$'<stuff>'`).  However in this case we don't care.  We only care if `[CI:NEXT]` is present or not.

contrib/cirrus/setup.sh
cevich: Sure.

contrib/cirrus/setup.sh
cevich: I'll just get rid of it.

contrib/cirrus/lib.sh
cevich: Added a `TODO`.

cevich:
> This deserves to be documented

Yeah probably.  It got way more out of hand in podman-CI, which is why it's documented there.  And still I don't think many/any people read it :disappointed:

cevich:
Force-push: Addressed all of Ed's concerns, except the missing docs.

edsantiago (approved):

PR #4832, opened 2023-05-31
Remove device mapper support

kolyshkin:
#### What this PR does / why we need it:

Remove device mapper support.

It seems that no one is using devicemapper storage driver nowadays, since overlayfs is so much better.

#### Which issue(s) this PR fixes:

Related to https://github.com/containers/storage/issues/1619

#### Does this PR introduce a user-facing change?

```release-note
Device mapper storage driver is removed.
```

PR #4837, opened 2023-06-01
intermediate-images inherit-label test: make it debuggable

edsantiago:
Test is currently very hard to debug on failure. Fix that
by adding unique test descriptors and a little whitespace.

Also, fix a broken/NOP test (copypaste artifact)

```release-note
None
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -1955,27 +1955,33 @@ _EOF
 
 @test "build using intermediate images should not inherit label" {
   _prefetch alpine
+
+  # Build imageone, with a label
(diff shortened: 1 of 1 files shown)


kolyshkin (commented):
tests/helpers.bash says `expect_output` is deprecated by `assert`. Now, I'm sure it does not make sense to convert everything to `assert`, but maybe use a new version when you touch the code.

kolyshkin (commented):
LGTM overall.

edsantiago:
@kolyshkin point taken, but I was going for ease of review to get this in ASAP (for your #4832, actually, in case it could be helpful in tracking down that weird CI failure).

PR #4847, opened 2023-06-05
[WIP] Replace all docker hub references with quay

cevich:
#### What type of PR is this?

/kind flake

***Depends on:*** https://github.com/containers/buildah/pull/4828

Previously tests relied heavily on magic present in
`tests/registries.conf`.  This is less than intuitive and potentially
surprising to test developers.  Search down and replace all test
references to docker.io images with quay.io/libpod images.

#### How to verify it

CI will pass

```release-note
None
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -1247,7 +1247,7 @@ _EOF
   run_buildah inspect --format "{{.Docker.ContainerConfig.ExposedPorts}}" test2
   expect_output "map[8080/tcp:{}]"
   run_buildah inspect --format "{{index .Docker.History 2}}" test1
-  expect_output --substring "FROM docker.io/library/alpine:latest"
+  expect_output --substring "FROM quay.io/libpod/alpine:latest"
 
   run_buildah build $WITH_POLICY_JSON --layers -t test3 -f Dockerfile.2 $contextdir
   run_buildah images -a
@@ -1285,7 +1285,7 @@ _EOF
   run_buildah inspect --format '{{index .ImageAnnotations "org.opencontainers.image.base.digest" }}' test
   expect_output "$fromDigest" "base digest from alpine"
   run_buildah inspect --format '{{index .ImageAnnotations "org.opencontainers.image.base.name" }}' test
-  expect_output "docker.io/library/alpine:latest" "base name from alpine"
+  expect_output "quay.io/libpod/alpine:latest" "base name from alpine"
 
   run_buildah build $WITH_POLICY_JSON --layers -t test1 -f Dockerfile.6 $BUDFILES/use-layers
   run_buildah images -a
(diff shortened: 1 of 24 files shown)


cevich:
@edsantiago I could use your eyes and experience with these tests.  [A good many of them are failing](https://cirrus-ci.com/task/5979929179848704?logs=integration_test) due (apparently) `Resolved "alpine" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)` pointing at `docker.io/library/alpine:latest`.  Whereas the tests expect `quay.io/libpod/alpine:latest`. I can't find where that `000-shortnames.conf` comes from, any clue?  Also, shouldn't that be overridden by `tests/registries.conf`?  Maybe should the CI scripts just remove this file if it exists (makes me nervous)?

---

Stepping back, I believe all these changes were your recommendation (so they align with the conformance-tests and operate with more POLA).  Maybe I'm mistaken, and/or maybe these changes don't actually have much benefit?  Maybe it's okay to rely heavily on the `registries.conf` (hack?) working on into the foreseeable future and test-developers will just have to suffer?

edsantiago:
Looking at one single failure:
```
not ok 100 bud with --layers and --no-cache flags
...
#/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
#|     FAIL: buildah inspect --format {{index .Docker.History 2}} test1
#| expected: =~ 'FROM quay.io/libpod/alpine:latest'
#|   actual:    '[....] FROM docker.io/library/alpine:latest false}'
```
Context: this is the Dockerfile:
```
FROM alpine
RUN mkdir /hello
...
```
And this is the relevant magic from `/etc/containers/registries.conf.d/000-shortnames.conf `:
```
"alpine" = "docker.io/library/alpine"
```
This means that `FROM alpine` becomes `FROM docker.io/library/alpine`. It is IMHO insane for us to muck with the shortnames file.

From there, the registries.conf *invisibly* translates `docker.blah` to `quay.blah`. "Invisibly" means "this does not echo in the buildah output".

There's a reason I gave up on my attempt to fix this all those years ago. My super-strong preference was and still is to use FQINs everywhere, but that ended up introducing other nightmares, and I just gave up. Kudos to you for persevering, but I'm sorry, I don't know what to suggest here. I will admire you if you pick at each failure and fix it, I will also admire you if you toss your hands up in despair and run away. This is gory stuff.

cevich:
> It is IMHO insane for us to muck with the shortnames file.

Ahh okay, I guess that was my sticking point.  I thought maybe there was a chance it was put in place by some test-related script somewhere.

> There's a reason I gave up on my attempt to fix this all those years ago.

Oh, you tried this already and gave up?  Hmmmm, that makes me less inclined to muck with it further then.

Okay -sigh-, "magic" it is then.

PR #4850, opened 2023-06-06
[release-1.24] Explicitly ref. quay images for CI

cevich:
#### What type of PR is this?

/kind flake

Commit d5b77bf72ca1 updated the conformance test references from a deprecated repository.  However, by pointing at the docker-hub it inadvertantly introduce a significant possibility for flakes.  This is because anonymous docker hub access is rate-limited by IP.  We cannot predict the IP used for CI VMs & Containers, any of which could be at or close to the limit.  Fix this by pointing conformance tests explicitly at `quay.io/libpod/centos` which is excluesively for use by CI, with static images.

#### How to verify it

CI will pass

```release-note
None
```

### DIFF
--- a/tests/conformance/README.md
+++ b/tests/conformance/README.md
@@ -15,7 +15,16 @@ Conformance tests use Docker CE to build images to be compared with images built
 
 ## Run conformance tests
 
-You can run all of the tests with go test:
+First, pull base images used by various conformance tests:
+```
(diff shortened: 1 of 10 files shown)


edsantiago:
LGTM

TomSweeneyRedHat (commented):
LGTM

PR #4870, opened 2023-06-16
[CI:BUILD] Packit: add jobs for downstream Fedora package builds

lsm5:
Get rid of `buildah.spec.rpkg` in favour of
`rpm/buildah.spec` which gets synced with fedora dist-git on every upstream release. The version in the new spec file is set to `0` by default and gets updated by packit automatically on every packit task.

Packit will automatically create a PR on fedora dist-git on every new upstream release. A sample PR will look like:
https://src.fedoraproject.org/rpms/container-selinux/pull-request/10#

A dry run for this can be triggered using:
`$ packit propose-downstream --local-content`

To run this command locally, you would need to have your packit user-configuration-file set.
Ref: https://packit.dev/docs/configuration/#user-configuration-file

along with a fedora api key created at:
https://src.fedoraproject.org/settings#nav-api-tab with sufficient ACLs.

This patch incurs zero additional cost to upstream as the downstream tasks occur only after upstream release and do not block anything upstream.

[NO NEW TESTS NEEDED]

#### What type of PR is this?

> /kind other

Easier rpm maintenance. Doesn't affect upstream directly.

#### How to verify it

Doesn't affect upstream directly. Anyway, can be verified by successful Packit tasks. A new release will also automatically create PRs on fedora dist-git for rpm updates.

None

None

```release-note
None
```


lsm5:
@nalind @flouthoc @rhatdan PTAL

flouthoc (approved):
LGTM

Issue #4875, opened 2023-06-22
Support container-transports in --cache-to and --cache-from


penn5:
**Description**
There is no way to create a local cache without hosting a registry server on localhost and setting --cache-from/to=localhost/cache

**Steps to reproduce the issue:**
1. Set --cache-from=oci:/storage when using `buildah build`

**Describe the results you received:**
```
Error: unable to parse value provided `[]` to --cache-from: invalid repo "oci:/storage": must contain registry and repository: invalid reference format
```

**Describe the results you expected:**
Buildah uses the provided path to cache images, allowing me to upload to CI cache coordinator.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
buildah-1.30.0-1.fc38.x86_64
```

**Output of `buildah version`:**

```
Version:         1.30.0
Go Version:      go1.20.2
Image Spec:      1.0.2-dev
Runtime Spec:    1.1.0-rc.1
CNI Spec:        1.0.0
libcni Version:  v1.1.2
image Version:   5.25.0
Git Commit:      
Built:           Mon Apr 10 07:26:00 2023
OS/Arch:         linux/amd64
BuildPlatform:   linux/amd64

```

**Output of `cat /etc/*release`:**

```
NAME="Fedora Linux"
VERSION="38 (Container Image)"
ID=fedora
VERSION_ID=38
VERSION_CODENAME=""
PLATFORM_ID="platform:f38"
PRETTY_NAME="Fedora Linux 38 (Container Image)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:38"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f38/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=38
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=38
SUPPORT_END=2024-05-14
VARIANT="Container Image"
VARIANT_ID=container
Fedora release 38 (Thirty Eight)
Fedora release 38 (Thirty Eight)
```

[...]

PR #4879, opened 2023-06-26
distributedCache: allow local `dir://` and `container-storage://` as a cache transport

flouthoc:
Allowing local directory as transport enables users to make distributed caching more flexible by making sure.

* Allows end users to move cache sources and destination manually.
* Allows storing cache on an `NFS` or various mount sources.
* Allows end users to distribute cache without registry setup.

[NO NEW TESTS NEEDED]
[NO TESTS NEEDED]
// Will add test in final commit

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
distributedCache: allow local `dir://` and `container-storage://` as a cache transport
```

Closes: https://github.com/containers/buildah/issues/4875


rhatdan:
@nalind PTAL
My concern here is you are breaking API.

rhatdan:
Is this a breaking change?
@vrothberg @mtrmac PTAL

define/build.go
```diff
@@ -141,10 +140,10 @@ type BuildOptions struct {
 	TransientMounts []string
 	// CacheFrom specifies any remote repository which can be treated as
 	// potential cache source.
-	CacheFrom []reference.Named
+	CacheFrom []types.ImageReference
```
vrothberg, line 143: This is breaking the API yes.

vrothberg (commented):
I agree.  We probably need to add new fields for that purpose.  Tests would be great as well.

imagebuildah/stage_executor.go
```diff
@@ -985,10 +986,16 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
+	logCachePulled := func(cacheKey string, remote types.ImageReference) {
 		if !s.executor.quiet {
 			cachePullMessage := "--> Cache pulled from remote"
-			fmt.Fprintf(s.executor.out, "%s %s\n", cachePullMessage, fmt.Sprintf("%s:%s", remote.String(), cacheKey))
+			repo := ""
+			if remote.Transport().Name() == imagedocker.Transport.Name() {
+				repo = remote.DockerReference().String()
+			} else {
+				repo = remote.StringWithinTransport()
+			}
```
mtrmac, line 997: Is there really value in special-casing `docker://` here, enough to be worth the ambiguity?

imagebuildah/stage_executor.go
```diff
@@ -985,10 +986,16 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
-	logCachePulled := func(cacheKey string, remote reference.Named) {
+	logCachePulled := func(cacheKey string, remote types.ImageReference) {
 		if !s.executor.quiet {
 			cachePullMessage := "--> Cache pulled from remote"
-			fmt.Fprintf(s.executor.out, "%s %s\n", cachePullMessage, fmt.Sprintf("%s:%s", remote.String(), cacheKey))
+			repo := ""
+			if remote.Transport().Name() == imagedocker.Transport.Name() {
+				repo = remote.DockerReference().String()
+			} else {
+				repo = remote.StringWithinTransport()
```
mtrmac, line 996: If this is supposed to represent arbitrary transports, please use `transports.ImageName()`, not a single-transport-scoped name.

pkg/parse/parse.go
```diff
@@ -53,16 +53,17 @@ const (
 )
 
 // RepoNamesToNamedReferences parse the raw string to Named reference
-func RepoNamesToNamedReferences(destList []string) ([]reference.Named, error) {
-	var result []reference.Named
+func RepoNamesToNamedReferences(destList []string) ([]types.ImageReference, error) {
```
mtrmac, line 56: The function name and comment are now incorrect.

pkg/parse/parse.go
```diff
@@ -53,16 +53,17 @@ const (
...
+	var result []types.ImageReference
 	for _, dest := range destList {
-		named, err := reference.ParseNormalizedNamed(dest)
+		if t := alltransports.TransportFromImageName(dest); t == nil {
+			// assume no default transport provided in such case append docker
+			dest = "docker://" + dest
```
mtrmac, line 61: If this remains: There are already over half a dozen copies of a similar fallback code, in many different variants. That should be centralized.

(For a start, there is `util.DefaultTransport`.)

imagebuildah/stage_executor.go
```diff
@@ -985,10 +986,16 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
+			}
+			fmt.Fprintf(s.executor.out, "%s %s\n", cachePullMessage, fmt.Sprintf("%s:%s", repo, cacheKey))
 		}
 	}
 	// logCachePush produces build log for cases when `--cache-to`
```
mtrmac, line 1001: Doesn’t this also need updating?

imagebuildah/stage_executor.go
```diff
@@ -1782,18 +1789,29 @@ func (s *StageExecutor) generateCacheKey(ctx context.Context, currNode *parser.N
...
+			result = append(result, dest)
+		} else if repo.Transport().Name() == imagedirectory.Transport.Name() {
+			dirPath := repo.StringWithinTransport()
+			tagged := filepath.Join(dirPath, cachekey)
+			dest, err := imagedirectory.NewReference(tagged)
+			if err != nil {
+				return nil, fmt.Errorf("failed generating directory reference for %q: %w", tagged, err)
+			}
+			result = append(result, dest)
 		}
```
mtrmac, line 1814: `else`… this silently drops unsupported input without telling the user?!

imagebuildah/stage_executor.go
```diff
@@ -1782,18 +1789,29 @@ func (s *StageExecutor) generateCacheKey(ctx context.Context, currNode *parser.N
...
+			}
+			result = append(result, dest)
+		} else if repo.Transport().Name() == imagedirectory.Transport.Name() {
+			dirPath := repo.StringWithinTransport()
+			tagged := filepath.Join(dirPath, cachekey)
+			dest, err := imagedirectory.NewReference(tagged)
+			if err != nil {
+				return nil, fmt.Errorf("failed generating directory reference for %q: %w", tagged, err)
+			}
+			result = append(result, dest)
```
mtrmac, line 1813: I’m fairly unhappy with using the transport:image-name syntax, and _especially_ the `types.ImageReference` values, to represent things that are not a single image. This would not be the first instance where that happens but I’m unhappy every time :)

Admittedly c/image has no “repo/collection of images” abstraction… and that’s because, just like this code shows, there’s just too little the transports have in common to build a meaningful abstraction like that.

imagebuildah/stage_executor.go
```diff
@@ -1851,14 +1869,20 @@ func (s *StageExecutor) pullCache(ctx context.Context, cacheKey string) (referen
 			ReportWriter:        nil,
 			PullPolicy:          define.PullIfNewer,
 		}
-		id, err := buildah.Pull(ctx, src.DockerReference().String(), options)
+		srcString := ""
+		if src.Transport().Name() == imagedocker.Transport.Name() {
+			srcString = src.DockerReference().String()
+		} else if src.Transport().Name() == imagedirectory.Transport.Name() {
+			srcString = "dir://" + src.StringWithinTransport()
+		}
```
mtrmac, line 1877: `else`… what??

imagebuildah/stage_executor.go
```diff
@@ -1851,14 +1869,20 @@ func (s *StageExecutor) pullCache(ctx context.Context, cacheKey string) (referen
 			ReportWriter:        nil,
 			PullPolicy:          define.PullIfNewer,
 		}
-		id, err := buildah.Pull(ctx, src.DockerReference().String(), options)
+		srcString := ""
+		if src.Transport().Name() == imagedocker.Transport.Name() {
+			srcString = src.DockerReference().String()
+		} else if src.Transport().Name() == imagedirectory.Transport.Name() {
+			srcString = "dir://" + src.StringWithinTransport()
```
mtrmac, line 1876: `//` is not a part of the `dir:` syntax (and, theory corner, in POSIX, paths starting with `"//"` have, _hypothetically_ special meaning, some UNIX systems IIRC accepted a `"//nfs.host.name/path"` at the kernel level).

This should just unconditionally use `transports.ImageName`; the called function uses `alltransports.ParseImageName` (before starting with heuristics), so this should use `transports.ImageName` to create a clearly unambiguous input.

mtrmac (commented):
- This doesn’t implement the referenced `containers-storage` _at all_. I’m not convinced it should, but in place of any code there’s just … nothing.
- `c/image/types.ImageReference` represents a _single_ image. I don’t know that there is any _reasonable_ way to represent a transport-independent “image namespace” where the caching logic could create entries for individual `cacheKey` values.

Given the above, I think the most direct approach is to add a `Cache{From,To}Dir string` fields in the internal API.

As for the CLI, I’m also a bit unhappy with using `dir:/parent/dir` as a syntax, and I’d prefer a `--cache-from-dir` here as well.
- This just doesn’t generalize. I don’t know what `--cache-to=docker-archive:…` or `--cache-from=sif:…` would mean or do.
- I’m _always_ skeptical of options that accept ambiguous syntaxes. Right now with this PR, `dir:foo` becomes ambiguous between `dir:./foo` and `docker://docker.io/library/dir:foo` (because tags in inputs are no longer prohibited).

`--cache-from-dir foo` is _not_ more typing than `--cache-from dir:foo`, so I don’t see much of a downside to a separate option.

---

I didn’t revisit the cache functionality to recall what it does exactly, and in particular see what compression means / doesn’t mean. (Notably pushes to `dir:` don‘t compress input by c/image default.)

Intuitively I’m very skeptical about using `containers-storage`, that is a _very expensive_ way to store layer tarballs.

rhatdan:
Closing as this has not been touched in 6 months.

Issue #4351, opened 2022-10-19
Containerfile LABEL is annotation?


grooverdan:
**Description**

Should Containerfile/Dockerfile `LABEL` end up as `Labels` (like it is) or `annotations` map (from [spec](https://github.com/opencontainers/image-spec/blob/v1.0.1/image-index.md))?

There's no obvious spec file but in the lack of `ANNOTATIONS` as a keyword it seems to map.

I've seen I can achieve this with `buildah manifest annotate` (and `buildah config --annotation`), I'm just wondering if the file based implementation is still correct with spec development

**Steps to reproduce the issue:**
1.
Dockerfile:
```
FROM ubuntu:jammy
LABEL org.opencontainers.image.authors="MariaDB Community"
```

2. buildah bud --tag m10x .
3. podman inspect m10x

**Describe the results you received:**

```
...
    "OCIv1": {
        "created": "2022-10-19T00:23:52.586469914Z",
        "architecture": "amd64",
        "os": "linux",
        "config": {
...
            "Labels": {
                "io.buildah.version": "1.27.0",
                "org.opencontainers.image.authors": "MariaDB Community",
```

**Describe the results you expected:**

```
...
    "OCIv1": {
        "created": "2022-10-19T00:23:52.586469914Z",
        "architecture": "amd64",
        "os": "linux",
        "config": {
...
            "Annotations": {
                "io.buildah.version": "1.27.0",
                "org.opencontainers.image.authors": "MariaDB Community",
```

**Output of `rpm -q buildah` or `apt list buildah`:**

```
$ rpm -q buildah
buildah-1.27.0-1.fc36.x86_64
```

**Output of `buildah version`:**

```
$  buildah version
Version:         1.27.0
Go Version:      go1.18.4
Image Spec:      1.0.2-dev
Runtime Spec:    1.0.2-dev
CNI Spec:        1.0.0
libcni Version:  v1.1.2
image Version:   5.22.0
Git Commit:      
Built:           Tue Aug  9 12:26:49 2022
OS/Arch:         linux/amd64
BuildPlatform:   linux/amd64
```

**Output of `podman version` if reporting a `podman build` issue:**

```
(paste your output here)
```

**Output of `cat /etc/*release`:**

```
Fedora release 36 (Thirty Six)
NAME="Fedora Linux"
VERSION="36 (Workstation Edition)"
ID=fedora
VERSION_ID=36
VERSION_CODENAME=""
PLATFORM_ID="platform:f36"
[...]

PR #4881, opened 2023-06-26
Add --label2annotation flag

rhatdan:
Containerfile and Dockerfile have not way to describe Annotations.

--ltoa flat tells buildah to convert all Labels specified in the Containerfile to Annotations when generating OCI Image format.

Fixes: https://github.com/containers/buildah/issues/4351

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
buildah build supports --ltoa flag to convert Label references within the Containerfile to Annotations in the final OCI Image.
```

### DIFF
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -476,6 +476,10 @@ If the specified capabilities are not in the default set, container engines
 should print an error message and will run the container with the default
 capabilities.
 
+**--label2annotation** *bool-value*
+
+Labels defined in the Containerfile are saved annotations in OCI Images. These labels will not be added as labels to the image. This options is not supported for --format=docker. (Default is `false`).
+
 **--layers** *bool-value*
 
 Cache intermediate images during the build process (Default is `false`).
--- a/define/build.go
+++ b/define/build.go
@@ -244,6 +244,8 @@ type BuildOptions struct {
 	Squash bool
 	// Labels metadata for an image
 	Labels []string
+	// LabelsToAnnotations tells Buildah to save labels as annotations in OCI Images
+	LabelsToAnnotations bool
 	// Annotation metadata for an image
 	Annotations []string
 	// OnBuild commands to be run by images based on this image
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -100,6 +100,7 @@ type Executor struct {
 	iidfile                 string
 	squash                  bool
 	labels                  []string
+	labelsToAnnotations     bool
 	annotations             []string
 	layers                  bool
 	noHosts                 bool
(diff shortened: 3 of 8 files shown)


docs/buildah-build.1.md
```diff
@@ -476,6 +476,10 @@ If the specified capabilities are not in the default set, container engines
 should print an error message and will run the container with the default
 capabilities.
 
+**--ltoa** *bool-value*
```
Luap99, line 479: I find the `ltoa` name very confusing. In my programmer brain this means "long to ascii" which is also very close to the more common `itoa`.
I prefer names that a normal person can understand without looking up the man page.

TomSweeneyRedHat: @Luap99 , How does `lta` or `l-to-a` or `l2a` grab you?  Although "el" in front of 2 might be mistaken for the word "one" instead of "el".

TomSweeneyRedHat:
LGTM once the label name is sorted

rhatdan:
How about
--label2annotation

Luap99:
`--label2annotation` SGTM

flouthoc:
I also like `--annotate-labels` but `--label2annotation` is also nice.

docs/buildah-build.1.md
```diff
@@ -476,6 +476,10 @@ If the specified capabilities are not in the default set, container engines
 should print an error message and will run the container with the default
 capabilities.
 
+**--label2annotation** *bool-value*
+
+Labels to annotations tells Buildah to save labels defined in the Containerfile as annotations in OCI Images. (Default is `false`).
```
nalind, line 481: It really needs to be clear that the image gets annotations _instead of_ labels.

nalind:
What's the expectation if the base image has labels set on it?

rhatdan:
I think the Labels in the base image are just carried forward as labels.

docs/buildah-build.1.md
rhatdan: How about now?

nalind:
> I think the Labels in the base image are just carried forward as labels.

Can the tests confirm the behavior?

And is that what a person using this option would expect?  I don't want to lock in behavior that we might decide to change later.

docs/buildah-build.1.md
nalind: The first two sentences read kind of weird to me.  The third has a subject/verb agreement problem.

rhatdan:
It is what I would expect. And the test already confirms it.

nalind:
How?  The base image used in the tests doesn't include any labels of its own that might or might not be carried through to an image based on it that was built with the new option.

rhatdan:
I answered a different question then you asked.  I thought you were asking if --label2annotation test checked to make sure that annotations were stored and labels were not.

rhatdan:
It does convert the image labels to Annotations. Not sure if this is right or wrong.

PR #4880, opened 2023-06-26
chore: pkg imported more than once

testwill:
#### What type of PR is this?

> /kind cleanup

#### What this PR does / why we need it:
pkg is being imported more than once (ST1019)
#### How to verify it
staticcheck
#### Does this PR introduce a user-facing change?

```release-note
None
```

[NO NEW TESTS NEEDED]

PR #2625, opened 2020-09-15
Validate: reference HEAD

cevich:
This should fix running git validate on branches.

Signed-off-by: Chris Evich <cevich@redhat.com>

PR #4884, opened 2023-06-27
lint: reenable staticcheck

edsantiago:
It was disabled in #2625 ("Use same linters as podman"), but
unfortunately with no explanation.

That PR removed a lot more linters. Someone should probably
go through, reevaluate & reenable.

[ Reference: #4880 ]

```release-note
None
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,5 +9,6 @@ run:
 linters:
   enable:
     - revive
+    - staticcheck
     - unconvert
     - unparam


cevich:
Tip (if you push again): Probably no need to run the full CI suite for this.  I think you can `[CI:DOCS]` this since golangci-lint always runs.

> unfortunately with no explanation.

Clearly that was my fault.  I don't remember for sure, but most likely I simply copied the config. from podman w/o looking twice.  

In case it helps, my best guess is:  Frequently, golangci-lint linters are a major PITA to debug when they don't work.  They tend to fail in either a panic, hang or a non-zero exit w/o any message or explanation.  It's quite frustrating :angry:

.golangci.yml
```diff
@@ -9,5 +9,6 @@ run:
 linters:
   enable:
     - revive
+    - staticcheck
```
cevich, line 12: Minor:  It's non-obvious from this config file, but there is an important note in the upstream golangci-lint readme:  

    It's not the same thing as the staticcheck binary. The author of staticcheck 
    doesn't support or approve the use of staticcheck as a library inside golangci-
    lint.

Maybe consider if it would help future maintainers to add a comment like:

`# N/B: Not the same as https://staticcheck.dev/`

Though, oddly enough, clicking on 'staticcheck' on the golangci-lint page, brings you to that site.  So I'm not 100% on what to think about the readme statement :confused:

edsantiago:
It's not a fault thing, have you looked at the histories of `.golang-ci.yml` in podman and buildah? (Suggestion: don't). Combine new VMs, plus new Go versions, plus time pressures, plus best intentions, and important things get dropped by the wayside. 

Thanks for the CI-DOC hint, but I'm still wondering why CI is passing (so far) on this. I was expecting a lint failure.

.golangci.yml
edsantiago: Ah, nice, that probably explains why CI is passing. Thanks,

cevich:
> It's not a fault thing,

No worries, yep it's a complete mess.  Just trying to add some context in case it helps.

> I was expecting a lint failure.

Maybe this is related to the "Not the same thing as the staticcheck binary" note?

cevich:
Oof, found another [confusing link to staticcheck.io](https://golangci-lint.run/usage/linters/#staticcheck) as well.  Maybe it is the same thing and the golangci-lint readme comment is wrong? :man_shrugging: 

> I was expecting a lint failure.

Oh, maybe @testwill used a different/newer version than what comes in with golangci-lint (assuming it's even the same tool)?

cevich (approved):
LGTM (beyond the confusing stuff about what/which staticcheck is actually being run).

edsantiago:
Do not merge. This should not have passed. I need to understand why it did.

cevich:
> that is beyond the scope of what I had set out to do

I can look into this.  I've got a PR already open to build new images, should be easy to drop in a tool like this.  Alternatively, this took strikes me as something somebody has probably already written a github-action for.  Should be easy to just drop that in.

cevich:
Ahh, yep, [there's an offical GHA for it](https://staticcheck.dev/docs/running-staticcheck/ci/github-actions/).

PR #4888, opened 2023-06-28
pkg/overlay: add limited support for FreeBSD

dfr:
Note: in theory, we could support read/write overlays on FreeBSD using a combination of unionfs and nullfs but this would take two mounts and the API only lets us return a single mount from MountWithOptions. Read only mounts can be done with just nullfs and this is enough to support read only image mounts in podman.

#### What type of PR is this?

/kind feature

None

```release-note
None
```

### DIFF
--- a/pkg/overlay/overlay.go
+++ b/pkg/overlay/overlay.go
@@ -6,7 +6,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
-	"syscall"
 
 	"errors"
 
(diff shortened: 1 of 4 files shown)


rhatdan:
LGTM
/approve

rhatdan:
@vrothberg @flouthoc @giuseppe PTAL

giuseppe (approved):
LGTM

vrothberg (approved):
/lgtm

flouthoc (approved):
LGTM
/approve

PR #4891, opened 2023-06-28
[CI:BUILD] RPM: Fix ELN builds

lsm5:
ELN builds have been failing because `/usr/bin/go-md2man` isn't available via the `go-md2man` package on ELN.
https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/6119506/

This commit changes the dependency to the go-md2man binary path so it should work on all build envs.

Successful build with the change:
https://koji.fedoraproject.org/koji/taskinfo?taskID=102707004

#### What type of PR is this?

/kind other

Fixes Fedora ELN build

#### How to verify it

See links mentioned in commit message

None

None

```release-note
None
```

### DIFF
--- a/rpm/buildah.spec
+++ b/rpm/buildah.spec
@@ -44,7 +44,7 @@ Summary: A command line tool used for creating OCI Images
 URL: https://%{name}.io
 # Tarball fetched from upstream
 Source: %{git0}/archive/v%{version}.tar.gz
-BuildRequires: go-md2man
+BuildRequires: %{_bindir}/go-md2man
 BuildRequires: device-mapper-devel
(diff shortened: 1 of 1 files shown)


lsm5:
aarch64 builds can be ignored for now. There's copr infra issues blocking those.

rhatdan:
LGTM

lsm5:
I realized this change isn't really needed. The copr environment actually needed an additional repo enabled which has been handled in copr settings itself. But this change itself is harmless and more generic as such. I'll close this PR and add the change itself in another upcoming PR to fix downstream task breakage.

PR #4892, opened 2023-06-28
[WIP] [CI:DOCS] GHA: Add staticcheck workflow

cevich:
#### What type of PR is this?

/kind other

Ref:
  * #4880
  * #4884

This is ***NOT*** the same thing as the golangci-lint staticcheck.  It checks other items not covered by golangci-lint, such as multiple (redundant) package imports).

#### How to verify it

The new github-actions workflow will pass.

None

```release-note
None
```

### DIFF
--- a/.github/workflows/staticcheck.yml
+++ b/.github/workflows/staticcheck.yml
@@ -1,0 +1,36 @@
+---
+
+name: "Golang Static Check"
+
+on: ["push", "pull_request"]
+
+permissions: "read-all"
+
+jobs:
+  static_check:
+    name: "Golang Static Check"
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


cevich:
More reason to not do this in CI: https://staticcheck.dev/docs/running-staticcheck/cli/build-tags/

I'll likely still bring the tool into the CI VM images, on the off-chance somebody wants to use it from the CI container images, or `hack/get_ci_vm.sh`.

edsantiago:
I would suggest not bringing it into the CI VM images. Instead, if someone needs it, they can vendor it into `tests?/tools`.

cevich:
K.  Thanks for the suggestion.

PR #4912, opened 2023-07-06
manifest, push: implement `--add-compression` to push with compressed variants.

flouthoc:
This is a feature PR for `c/image`'s `EnsureCompressionVariantExists` feature merged in below PR:
* https://github.com/containers/image/pull/2023
* https://github.com/containers/image/pull/1987
* https://github.com/containers/image/pull/2029

~~#### Proof of concept to demonstrate the new API which are recently added in `c/image` and some which still needs to be merged.~~

Adds a new flag `--add-compression <compression>` to `manifest push`, which when added will push existing manifest on host while replicating the instances with required `compression`.

Example usage

```console
./buildah manifest create localhost:5000/list:latest
./buildah build -t test .
./buildah manifest add localhost:5000/list:latest test
# Push with replication for `zstd`
./buildah manifest push --tls-verify=false --all --add-compression zstd localhost:5000/list:latest
```

Expected pushed manifest list
```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:d814812aea8237398f0a5c36ee6fabc0dbaa4c1b55453b3b9f87e5a4ac2bd4f1",
      "size": 759,
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:de950651c5648b48d8b41fda88245c8fa1060f55125f0387c8b0c2e878733e01",
      "size": 759,
      "annotations": {
        "io.github.containers.compression.zstd": "true"
      },
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    }
  ]
}
```

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
manifest, push: implement --add-compression to push with compressed
```


vendor/github.com/containers/common/libimage/manifests/manifests.go
```diff
@@ -239,6 +241,16 @@ func (l *list) Push(ctx context.Context, dest types.ImageReference, options Push
 			return nil, "", err
 		}
 	}
+
+	compressionVariants := []cp.CopyOptionCompressionVariant{}
+	//options.EnsureCompressionVariantExists = []CopyOptionCompressionVariant{CopyOptionCompressionVariant{Algorithm: &compression.Zstd, Level:20}}
+	for _, compressionVariant := range options.ReplicateWithCompression {
+		if compressionVariant == "zstd" {
```
mtrmac, line 248: Else this should not silently do _nothing_; with code as is, any future feature addition could break a working user.

Instead, any unexpected value should cause a failure.

vendor/github.com/containers/common/libimage/manifests/manifests.go
```diff
@@ -239,6 +241,16 @@ func (l *list) Push(ctx context.Context, dest types.ImageReference, options Push
...
+			variant := cp.CopyOptionCompressionVariant{Algorithm: &compression.Zstd, Level:20}
```
mtrmac, line 249: Why does this have an opinion on the compression level?

cmd/buildah/manifest.go
```diff
@@ -221,6 +221,7 @@ func init() {
 	flags.StringVar(&manifestPushOpts.creds, "creds", "", "use `[username[:password]]` for accessing the registry")
 	flags.StringVar(&manifestPushOpts.digestfile, "digestfile", "", "after copying the image, write the digest of the resulting digest to the file")
 	flags.StringVarP(&manifestPushOpts.format, "format", "f", "", "manifest type (oci or v2s2) to attempt to use when pushing the manifest list (default is manifest type of source)")
+	flags.StringSliceVar(&manifestPushOpts.replicateWithCompression, "replicate", nil, "replicate instances with selected compression while pushing")
```
mtrmac, line 224: I’m skeptical about the `replicate` name (`push --replicate=zstd` seems unintuitive to me), but, \*shrug\*, I can’t think of a good name.

mtrmac (commented):
(Just a quick skim to see how the c/image feature is planned to be used.)

- Eventually this should include man pages.
- So a simple `buildah bud && buildah push` will not, with this PR, be able to do this. That’s fine, this PR is valuable as is — it just seems like an enhancement worth tracking.

vendor/github.com/containers/common/libimage/manifests/manifests.go
flouthoc: The code in this PR is just experimental, i was using this code to test `c/image` but eventually I think most of the things in this PR can be changed and we can have a way to provide `level` at UX level.

vendor/github.com/containers/common/libimage/manifests/manifests.go
flouthoc: Yes I expect lot of changes in here, everything in this PR is just dummy and overall draft. Once we have a agreement on UX design on this PR. I will create a new PR. Maybe we can treat this PR as a brainstorm thing.

flouthoc:
> So a simple buildah bud && buildah push will not, with this PR, be able to do this

@mtrmac I think a simple `buildah bud` and then `buildah push --compression-format <> --compression-level <>` can still do it right for a single image, we support that on upstream as of now.

mtrmac:
That creates a zstd-only image, not a dual image compatible with gzip consumers but allowing use of zstd.

flouthoc:
@giuseppe @mtrmac @vrothberg @rhatdan PTAL

cmd/buildah/manifest.go
```diff
@@ -199,6 +202,9 @@ func init() {
 		Example: `buildah manifest inspect mylist:v1.11`,
 		Args:    cobra.MinimumNArgs(1),
 	}
+	flags = manifestInspectCommand.Flags()
+	flags.StringVar(&manifestInspectOpts.authfile, "authfile", auth.GetDefaultAuthFile(), "path of the authentication file. Use REGISTRY_AUTH_FILE environment variable to override")
+	flags.BoolVar(&manifestInspectOpts.tlsVerify, "tls-verify", true, "require HTTPS and verify certificates when accessing the registry. TLS verification cannot be used when talking to an insecure registry.")
```
mtrmac, line 207: This is actually implemented via `SystemContextFromOptions`?

Fine, it’s consistent with the other subcommands — but it’s fairly hard to find and follow. In some eventual future, consider some design where the flag definition and implementation are collocated.

docs/buildah-manifest-push.1.md
```diff
@@ -18,6 +18,10 @@ The list image's ID and the digest of the image's manifest.
 
 ## OPTIONS
 
+**--add-compression** *compression*
+
+If specified makes sure while pushing that requested compression variant exists for each platform in the manifest list.
```
mtrmac, line 23: This might not be the ideal place, but _somewhere_ we should document the accepted algorithm names, so that users know what values to use.

mtrmac (commented):
Code LGTM.

cmd/buildah/manifest.go
flouthoc: Yeah its done, similarly for all manifests command. I had to implement these so tests could use it.

docs/buildah-manifest-push.1.md
flouthoc: Probably in `cli` flags description ? WDYT ?

mtrmac: I’m not sure what ”`cli` flags description” refers to. Is that some man page, or just an internal Go documentation? The commend was targeting user-visible documentation, like man pages.

Short-term, we don’t _need_ to overcomplicate things — just list the accepted values (`gzip`, `zstd`, `zstd:chunked` (and hypothetically `Xz`, but I think that doesn’t work because we can’t represent it in manifests) here in this option description.

---

Currently `pkg/compression` does not provide an API to list accepted names. That might be worth adding.

flouthoc: @mtrmac I mean the flag description which shows when we do `buildah manifest push --help` but I think documentation should also mention it. So added here instead.

flouthoc:
@vrothberg @giuseppe PTAL

docs/buildah-manifest-inspect.1.md
```diff
@@ -16,6 +16,17 @@ Displays the manifest list or image index stored using the specified image name.
 
 A formatted JSON representation of the manifest list or image index.
 
+## OPTIONS
+
+**--authfile** *path*
```
vrothberg, line 21: Nit: the commit message calls it `--auth-file`

docs/buildah-manifest-inspect.1.md
```diff
@@ -16,6 +16,17 @@ Displays the manifest list or image index stored using the specified image name.
...
+
+Path of the authentication file. Default is ${XDG\_RUNTIME\_DIR}/containers/auth.json, which is set using `buildah login`.
+If the authorization state is not found there, $HOME/.docker/config.json is checked, which is set using `docker login`.
```
vrothberg, line 24: Risk of divergence:  I suggest pointing to `containers-auth.json` instead

docs/buildah-manifest-inspect.1.md
```diff
@@ -16,6 +16,17 @@ Displays the manifest list or image index stored using the specified image name.
...
+
+**--tls-verify** *bool-value*
+
+Require HTTPS and verification of certificates when talking to container registries (defaults to true).  TLS verification cannot be used when talking to an insecure registry.
```
vrothberg, line 28: Nit: the new flags are not tested in the commit adding them but the one after.

docs/buildah-manifest-push.1.md
```diff
@@ -18,6 +18,11 @@ The list image's ID and the digest of the image's manifest.
 
 ## OPTIONS
 
+**--add-compression** *compression*
+
+If specified while pushing option makes sure that requested compression variant exists for each platform in the manifest list.
```
vrothberg, line 23: This sentence is a bit hard to brain parse.  I'd drop `while pushing` since we're already in the `buildah-push` man page.

We should also elaborate how in interacts with `--compression` and that the ones specified here are in _addition_ to the default or --compression one.

vrothberg (commented):
Code and test LGTM

@edsantiago can you take a look at the test?

docs/buildah-manifest-inspect.1.md
flouthoc: I have copied it from other buildah docs and it matches with the text here: https://docs.podman.io/en/latest/markdown/podman-login.1.html#authfile-path

Could you suggest what needs to be changed here.

docs/buildah-manifest-inspect.1.md
flouthoc: I was not planning to add these flags but they were needed for the `--add-compression` test, I can add independent tests for these but they will be just clone of `--add-compression` test without the `--add-compression` flag, I decided not to add because of duplication since these two flags are already getting tested in the test added for `--add-compression`. But I could be wrong :sweat_smile:

docs/buildah-manifest-inspect.1.md
vrothberg: Ah OK, then let's stop me from derailing this PR into fixing docs :)

docs/buildah-manifest-push.1.md
flouthoc: Could you check if it looks any better now.

tests/bud.bats
```diff
@@ -11,6 +11,55 @@ load helpers
...
+  run_buildah manifest add foo image1
+  run_buildah manifest add foo image2
+
+  run_buildah manifest push $WITH_POLICY_JSON --authfile ${TEST_SCRATCH_DIR}/test.auth --all --add-compression zstd --tls-verify=false foo docker://localhost:${REGISTRY_PORT}/list
+
+  run_buildah manifest inspect --authfile ${TEST_SCRATCH_DIR}/test.auth --tls-verify=false localhost:${REGISTRY_PORT}/list
+  list="$output"
+
+  # verify original `gzip` and its `zstd` variant on `amd64`
+  # first instance must be `gzip` for `amd64`
```
edsantiago, line 37: I don't understand this comment nor the ones on lines 49-50. "must be gzip", but nothing here checks for gzip.

edsantiago (commented):
I'm going to assume that "testing for gzip" == "null annotation". I think that should be clearer in the comments.

There's a lot of unnecessary duplication in the test, it could maybe be refactored into a table. And, when possible, it's better to do exact matches instead of substring. 

This is just a quick glance, sorry, gotta head out.

mtrmac:
> I'm going to assume that "testing for gzip" == "null annotation".

Yes. (There are tentative thoughts that maybe having a ….compression.gzip would be useful, but all pre-existing gzip images carry no annotation right now, anyway.)

tests/bud.bats
flouthoc: Updated comment.

flouthoc:
> There's a lot of unnecessary duplication in the test, it could maybe be refactored into a table. And, when possible, it's better to do exact matches instead of substring.

@edsantiago Deduplicated contents in a function and also added exact match.

tests/bud.bats
```diff
@@ -2103,6 +2131,39 @@ function _test_http() {
...
+   gzip)
+    run jq -r '.manifests['$1'].annotations' <<< $2
+    # annotation is `null` for gzip compression
+    expect_output 'null'
+    ;;
+
+  zstd)
+    # annotation `'"io.github.containers.compression.zstd": "true"'` must be there for zstd compression
+    run jq -r '.manifests['$1'].annotations."io.github.containers.compression.zstd"' <<< $2
+    expect_output 'true'
```
edsantiago, line 2159: Similar to above.

tests/bud.bats
```diff
@@ -2103,6 +2131,39 @@ function _test_http() {
...
+    ;;
+  esac
+
+  run jq -r '.manifests['$1'].platform.architecture' <<< $2
+  expect_output $3
```
edsantiago, line 2164: And again.

edsantiago (changes requested):
Much better, thank you! But please add breadcrumbs so a maintainer has some hope of figuring out a failure.

tests/bud.bats
```diff
@@ -2103,6 +2131,39 @@ function _test_http() {
...
+#
+#     COMPRESSION       compression algorithm name; e.g "zstd".
+#
+function validate_instance_compression {
+  case $4 in
+
+   gzip)
+    run jq -r '.manifests['$1'].annotations' <<< $2
+    # annotation is `null` for gzip compression
+    expect_output 'null'
```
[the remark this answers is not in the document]

tests/bud.bats

tests/bud.bats

edsantiago (approved):

flouthoc:

cmd/buildah/manifest.go
```diff
@@ -696,6 +702,9 @@ func manifestAnnotateCmd(c *cobra.Command, args []string, opts manifestAnnotateO
 }
 
 func manifestInspectCmd(c *cobra.Command, args []string, opts manifestInspectOpts) error {
+	if err := auth.CheckAuthFile(opts.authfile); err != nil {
```
rhatdan, line 706: Can you wrap this in a check to see if the user modified it?

rhatdan:
LGTM other the check to make sure user set the authfile.

flouthoc:
@lsm5 Some `rpm-builds` are failing, just informing.

rhatdan:
If I want to push `zstd:chunked` only, now what do I do?  Can I set this up in containers.conf to do this all the time?

flouthoc:
> Can I set this up in containers.conf to do this all the time

Not as of now but I think we can create a field in `container.conf` for default value of `--add-compression` WDYT ?

> If I want to push zstd:chunked only

I think `zstd:chunked` is still waiting for this : https://github.com/containers/image/pull/1980 and if only `zstd:chunked` is needed then you can use simple `--compression` flag.

rhatdan:
No.  In Fedora 39, the default for pushing to registries has to be "gzip,zstd:chunked" Then Podman can take advantage of zstd. And old docker will continue to work.
In Fedora 40 we ship to just zstd:chunked.  But this setting has to be set in containers.conf so distributions can override it.

mtrmac:
> the default for pushing to registries

I mean, we are not at that point at all yet. This is `manifest push`, with an explicit sequence of commands.

A simple `podman push` / ` buildah push` (optionally!) creating a multi-instance image definitely makes sense, but just doesn’t exist right now. That path is also different in that the source is always uncompressed, so the difference between `--compression` and `--add-compression` is not as pronounced.

rhatdan:
Ok as long as I get my eventual goal, I am fine with this.

PR #4993, opened 2023-08-24
Fixing the owner of the storage.conf for the build user in the builda…

indyvanmol:
#### What type of PR is this?

/kind bug

In the Containerfile for the buildah image when the containers storage.conf file is created for the build user, the file is owned by root by adding a chown for the build user after the file is created it will then be owned by the build user.

#### How to verify it
cd contrib/buildahimage
buildah build -t buildah:test -f Containerfile .
podman run --rm -u 1000 --device /dev/fuse -it buildah:test ls -l /home/build/.config/containers/storage.conf

None

None

```release-note

```

### DIFF
--- a/contrib/buildahimage/Containerfile
+++ b/contrib/buildahimage/Containerfile
@@ -99,7 +99,8 @@ RUN sed -e 's|^#mount_program|mount_program|g' \
         -e 's|^graphroot|#graphroot|g' \
         -e 's|^runroot|#runroot|g' \
         /etc/containers/storage.conf \
-        > /home/build/.config/containers/storage.conf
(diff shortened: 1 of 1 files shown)


nalind:
I'm not clear on what the current version breaks, but this brings it in line with the rest of the contents of that directory.  Thanks!
/lgtm
/hold

rhatdan:
/approve
Thanks @indyvanmol

TomSweeneyRedHat:
Changes LGTM, but your title is too wrong per the CI and it looks a rebase might be needed.  Can you shorten the git title and repush please @indyvanmol ?

indyvanmol:
@TomSweeneyRedHat alright

PR #5000, opened 2023-08-25
Support passing of ULimits as -1 to mean max

rhatdan:
Docker allows the passing of -1 to indicate the maximum limit allowed for the current process.

Fixes: https://github.com/containers/podman/issues/19319

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
You can now specify -1 for valures when setting ulimits to indicate maximum.
```

### DIFF
--- a/pkg/util/resource_unix.go
+++ b/pkg/util/resource_unix.go
@@ -1,0 +1,38 @@
+//go:build linux || freebsd || darwin
+// +build linux freebsd darwin
+
+package util
+
+import (
+	"fmt"
+	"syscall"
+
+	"github.com/docker/go-units"
+)
+
+func ParseUlimit(ulimit string) (*units.Ulimit, error) {
+	ul, err := units.ParseUlimit(ulimit)
+	if err != nil {
+		return nil, fmt.Errorf("ulimit option %q requires name=SOFT:HARD, failed to be parsed: %w", ulimit, err)
+	}
+
+	if ul.Hard != -1 && ul.Soft == -1 {
+		return ul, nil
+	}
+
+	rl, err := ul.GetRlimit()
+	if err != nil {
+		return nil, err
+	}
+	var limit syscall.Rlimit
+	if err := syscall.Getrlimit(rl.Type, &limit); err != nil {
+		return nil, err
+	}
+	if ul.Soft == -1 {
+		ul.Soft = int64(limit.Cur)
+	}
+	if ul.Hard == -1 {
+		ul.Hard = int64(limit.Max)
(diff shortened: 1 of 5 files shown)


flouthoc (approved):
LGTM

rhatdan:
@vrothberg @giuseppe @Luap99 @nalind @umohnani8 PTAL

pkg/util/resource_windows.go
```diff
@@ -0,0 +1,16 @@
+package util
+
+import (
+	"fmt"
+
+	"github.com/docker/go-units"
+)
+
+func ParseUlimit(ulimit string) (*units.Ulimit, error) {
```
vrothberg, line 9: It worries me a bit that there is a non-Linux function.  Is this being called by the remote client?  I think it only really makes sense to evaluate `-1` on the server side.

Can you make sure that's the case?

vrothberg (commented):
Please add an e2e/system test.

pkg/util/resource_windows.go
rhatdan: This is a buildah PR not a Podman PR (Yet, when I move this on the Podman side, I will update the tests).

vrothberg: Thanks!  I didn't notice it was a Buildah PR :see_no_evil:

vrothberg (approved):
/lgtm

pkg/util/resource_unix.go
```diff
@@ -0,0 +1,38 @@
...
+	var limit syscall.Rlimit
+	if err := syscall.Getrlimit(rl.Type, &limit); err != nil {
+		return nil, err
+	}
+	if ul.Soft == -1 {
+		ul.Soft = int64(limit.Cur)
+	}
+	if ul.Hard == -1 {
+		ul.Hard = int64(limit.Max)
+	}
```
giuseppe, line 36: do we need to cap the limit to the current ones also when running as root?

rhatdan: Does -1 work as root, and just not as rootless?

rhatdan: Of course even if we did allow -1, then if process did not have CAP_RESOURCE it would fail.  I think it is safest to just set to the current processes max.

TomSweeneyRedHat:
Might need a rebase here @rhatdan

PR #5001, opened 2023-08-25
Fixing the owner of the storage.conf.

indyvanmol:
#### What type of PR is this?

/kind bug

In the Containerfile for the buildah image when the containers storage.conf file is created for the build user, the file is owned by root by adding a chown for the build user after the file is created it will then be owned by the build user.

#### How to verify it
cd contrib/buildahimage
buildah build -t buildah:test -f Containerfile .
podman run --rm -u 1000 buildah:test ls -l /home/build/.config/containers/storage.conf

None

A new PR of https://github.com/containers/buildah/pull/4993 with a shorter git commit title so it can pass the CI.

None

```release-note

```

### DIFF
--- a/contrib/buildahimage/Containerfile
+++ b/contrib/buildahimage/Containerfile
@@ -99,7 +99,8 @@ RUN sed -e 's|^#mount_program|mount_program|g' \
         -e 's|^graphroot|#graphroot|g' \
         -e 's|^runroot|#runroot|g' \
         /etc/containers/storage.conf \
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

cevich:
Task passed on re-run.  Must just have been a fart.

TomSweeneyRedHat:
CHanges LGTM
The test issue is not at all related to this change, but we (@cevich ) will need to track it down.

PR #5014, opened 2023-08-30
manifest: `addCompression` use default from `containers.conf` 

flouthoc:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
manifest: addCompression use default from containers.conf
```


flouthoc:
PR similar to https://github.com/containers/podman/pull/19790

vrothberg (commented):
Is this _really_ something we should support in Buildah?  I am not against it but Buildah ignores containers.conf in most places.  This PR illustrates it quick a bit since "compression-format" is ignored but "add-compression" isn't.

flouthoc:
> Is this _really_ something we should support in Buildah? I am not against it but Buildah ignores containers.conf in most places. This PR illustrates it quick a bit since "compression-format" is ignored but "add-compression" isn't.

No issues, I just created PR for consistency with podman can drop PR if everybody agrees with it

rhatdan:
Where we can, I would want buildah to work with containers.conf?

flouthoc:
Rebased in-case if consensus is made to merge this one.

vrothberg (commented):
I'm torn.  Generally, I am supportive of supporting containers.conf in Buildah but as mentioned before, I do not think it's a nice user experience if only 1 option is used. What about `compression_format`? I think we should at a minimum support all fields which are sensitive to compression.

Then we can tackle the rest.  But imagine being a user trying to figure out what works and what doesn't.  It's even hard for maintainer as we had to look at the code.

TomSweeneyRedHat:
I'm fence-sitting too.  I'm thinking primarily of a Buildah Only user.  Hopping into containers.conf, seeing the myriad of options, and only being able to use one.  But I see the other side of the coin, too, where you want the consistency to happen between Buildah and Podman.

All in all, I'm slightly leaning toward including containers.conf, but we'll need a lot of Buildah doc in and out of it explaining what is and is not used by Buildah.

vrothberg:
So, I am OK to go step-by-step and containers.conf support gradually _BUT_ - as mentioned above - we should at least support the other compression options as well.

rhatdan:
Agreed

cmd/buildah/manifest.go
```diff
@@ -49,6 +50,11 @@ type manifestInspectOpts = struct {
 }
 
 func init() {
+	containerConfig, err := config.Default()
+	if err != nil {
+		logrus.Errorf(err.Error())
```
nalind, line 55: Always use pass a format specifier as the first argument for functions like `Errorf()`, so that we don't have to track down surprises later.  I am concerned about parsing this again in this `init()`, when we're doing the same thing in both `from.go` and `main.go`.

tests/bud.bats
```diff
@@ -39,6 +39,39 @@ _EOF
...
+  mkdir -p $contextdir
+
+  cat > $contextdir/Dockerfile1 << _EOF
+FROM alpine
+_EOF
+
+  cat > $contextdir/containers.conf << _EOF
+[engine]
+add_compression = ["zstd"]
+_EOF
```
nalind, line 53: These can be regular files in the source tree, since their content isn't dynamic.

rhatdan:
@flouthoc any update coming on this one?

rhatdan:
@flouthoc friendly ping.

PR #5016, opened 2023-08-31
vendor containers/common@12405381ff45

vrothberg:
When pulling from an OCI source, make sure to preseve the optional name. For instance, a podman pull oci:/tmp/foo:quay.io/foo/bar:latest should pull the image and name it quay.io/foo/bar:latest.

While at it, also fix a bug when pulling an OCI without the optional name. Previously, we used the path to name the image which will error in most cases due to invalid characters (e.g., capital ones). Hence, apply the same trick as for the dir transport and generate a sha.

#### What type of PR is this?

> /kind bug
> /kind feature

```release-note
Pulling from an `oci` transport will use the optional name for naming the image.
```


vrothberg:
@flouthoc this fixes the tests. Thanks for pointing it out!

flouthoc (approved):
LGTM
/lgtm
/hold

PR #5031, opened 2023-09-07
stage_executor: allow images without layers 

flouthoc:
A build case where base is `scratch` and nothing else, then produce image without layer.

This matches with `buildkit` and `docker`.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
 stage_executor: allow images without layers
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -1033,7 +1033,13 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			// squash the contents of the base image.  Whichever is
 			// the case, we need to commit() to create a new image.
 			logCommit(s.output, -1)
-			if imgID, ref, err = s.commit(ctx, s.getCreatedBy(nil, ""), false, s.output, s.executor.squash, lastStage); err != nil {
+			emptyLayer := false
+			if s.builder.FromImageID == "" {
+				// No base image means there's nothing to put in a
+				// layer, so don't create one.
+				emptyLayer = true
+			}
+			if imgID, ref, err = s.commit(ctx, s.getCreatedBy(nil, ""), emptyLayer, s.output, s.executor.squash, lastStage); err != nil {
 				return "", nil, fmt.Errorf("committing base container: %w", err)
 			}
 			// Generate build output if needed.
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -89,6 +89,15 @@ _EOF
   validate_instance_compression "3" "$list" "arm64" "zstd"
 }
 
+@test "no layer should be created on scratch" {
+  run_buildah build --layers --label "label1=value1" -t test -f $BUDFILES/from-scratch/Containerfile
+  run_buildah inspect -f '{{len .Docker.RootFS.DiffIDs}}' test
+  expect_output "0" "layer should not exist"
+  run_buildah build --layers -t test -f $BUDFILES/from-scratch/Containerfile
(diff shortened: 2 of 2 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM
But I would suggest changing the release note to:

When a build uses "scratch"  for the image and only a label is added, a layer is now added to the image rather than creating an empty image.

nalind:
If an image with no layers is a valid thing, then this is a bug in the logic that lists which images are in storage, and that's where this needs to be fixed.  I'd suggest adding error context with `fmt.Errorf()` in [layer_tree.go](https://github.com/containers/common/blob/bf61a3742802d8052edbd74e219bb2a1da4c72fd/libimage/layer_tree.go#L188) and going from there.

flouthoc:
@nalind SGTM but one thing which bothers me is that when a single `FROM scratch` is created without any labels it creates a layer, but as soon as  a label is added it does not creates any layer in the image. That part looked weird.

flouthoc:
Well it seems, docker/buildkit allows creating `image` without layers so we gotta do that do, and buildkit produces no layer even for `FROM scratch` , I'll modify this PR to make behavior of `FROM scratch` consistent with buildkit.

TomSweeneyRedHat:
@flouthoc looks like you'll need to rebase.

flouthoc:
@nalind I have modified this PR , and using this  PR to make buildah's behavior consistent with `docker` and `buildkit` and meanwhile the original issue , i'm trying to fix that here: https://github.com/containers/common/pull/1651

flouthoc:
@nalind @containers/buildah-maintainers PTAL original issue https://github.com/containers/podman/issues/19860 is closed in a different PR, following PR brings consistency with `buildkit` and `docker`

imagebuildah/stage_executor.go
```diff
@@ -1033,7 +1033,14 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			// squash the contents of the base image.  Whichever is
 			// the case, we need to commit() to create a new image.
 			logCommit(s.output, -1)
-			if imgID, ref, err = s.commit(ctx, s.getCreatedBy(nil, ""), false, s.output, s.executor.squash, lastStage); err != nil {
+			emptyLayer := false
+			if s.builder.FromImageID == "" && s.builder.FromImage == "" {
```
nalind, line 1037: We should be checking only `FromImageID` here, we don't double up on this comparison elsewhere.

imagebuildah/stage_executor.go
```diff
@@ -1033,7 +1033,14 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
...
+				// If we only have a base image and its `scratch` and
+				// since there are no other childrens in this branch
+				// then treat this as an `empty_layer`.
```
nalind, line 1040:
```suggestion
				// No base image means there's nothing to put in a layer, so don't create one
```

imagebuildah/stage_executor.go
flouthoc: Done.

imagebuildah/stage_executor.go
flouthoc: Done.

nalind:
LGTM

PR #5032, opened 2023-09-07
Make sure that pathnames picked up from the environment are absolute

nalind:
#### What type of PR is this?

/kind other

When we read TMPDIR and SSH_AUTH_SOCK from the environment, convert them to absolute paths before using them.  Call auth.GetDefaultAuthFile() instead of reading REGISTRY_AUTH_FILE.

#### How to verify it

I'm going to mark this as not needing new tests because we always pass the values we read to `filepath.Abs()`, and tests that set TMPDIR and SSH_AUTH_SOCK already exist.  If this breaks anything, then this should break those tests.

Fixes part of RHELPLAN-167391.

```release-note
None
```

### DIFF
--- a/pkg/sshagent/sshagent.go
+++ b/pkg/sshagent/sshagent.go
@@ -201,8 +201,13 @@ func NewSource(paths []string) (*Source, error) {
 	if len(paths) == 0 {
 		socket = os.Getenv("SSH_AUTH_SOCK")
 		if socket == "" {
-			return nil, errors.New("$SSH_AUTH_SOCK not set")
+			return nil, errors.New("SSH_AUTH_SOCK not set in environment")
 		}
+		absSocket, err := filepath.Abs(socket)
+		if err != nil {
+			return nil, fmt.Errorf("evaluating SSH_AUTH_SOCK in environment: %w", err)
+		}
+		socket = absSocket
 	}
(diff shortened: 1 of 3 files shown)


pkg/sshagent/sshagent.go
```diff
@@ -201,7 +201,13 @@ func NewSource(paths []string) (*Source, error) {
 	if len(paths) == 0 {
 		socket = os.Getenv("SSH_AUTH_SOCK")
 		if socket == "" {
-			return nil, errors.New("$SSH_AUTH_SOCK not set")
+			return nil, errors.New("SSH_AUTH_SOCK not set in environment")
+		}
+		absSocket, err := filepath.Abs(socket)
+		if err != nil {
+			return nil, fmt.Errorf("evaluating SSH_AUTH_SOCK in environment: %w", err)
+		} else {
```
flouthoc, line 209: Linter is not happy with this else block.

flouthoc (approved):
LGTM other than lint changes.

pkg/sshagent/sshagent.go
nalind: Ah, right.  Reworking a couple of those.

TomSweeneyRedHat:
LGTM

PR #5061, opened 2023-09-22
build: add support for `--inherit-labels`

flouthoc:
Allows users to specify if they want to inherit labels from base image or not.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
build: add support for --inherit-labels
```

### DIFF
--- a/define/build.go
+++ b/define/build.go
@@ -228,6 +228,9 @@ type BuildOptions struct {
 	// ID mapping options to use if we're setting up our own user namespace
 	// when handling RUN instructions.
 	IDMappingOptions *IDMappingOptions
+	// InheritLabels allows users to specify if they want
+	// to inheritlabels from base image or not.
+	InheritLabels types.OptionalBool
 	// AddCapabilities is a list of capabilities to add to the default set when
 	// handling RUN instructions.
 	AddCapabilities []string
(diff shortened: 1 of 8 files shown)


flouthoc:

nalind:
Okay, between this and `--unsetenv`, we really need to get some consistency around what we provided for changing the environment/labels/annotations/ports/volumes at the command line.

flouthoc:
Just continuing discussion from downstream here so other maintainers can chip in as well. 

From what i can see with `--unsetenv` it deals with only single variable per command, it does not covers a use-case where user wants to clear all `ENV` vars from the base image.  That would need `--inherit-env=false` to be implemented.

Here `--inherit-labels` solves a use-case where users don't want to inherit anything from the base image. For parity with `--unsetenv` i think we will need to implement `--unsetlabel` ( which solves a different use-case IMO ).

So I think for more intuitive UX and to cover both of these problems and maintain consistency, we will need to implement
* `--inherit-labels` and `--inherit-envs` : for both `env` and `labels`.
* `--unsetlabels` and `--unsetenv`: for both `env` and `labels`.

@containers/buildah-maintainers please provide your opinions on this one.

flouthoc:
fwiw i think immediate need is for `--unsetlabel` so i'll tweak this PR and we can talk about `--inherit-labels` later.

arewm:
Would `--unsetlabel` be added multiple times, once for each label being removed? Or does it support being passed a list of labels to unset?

flouthoc:
Created ALT https://github.com/containers/buildah/pull/5062

PR #5062, opened 2023-09-22
build,config: add support for `--unsetlabel`

flouthoc:
Just like `--unsetenv` following flag allows to unset image label.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
build: add support for `--unsetlabel`
config: add support for `--unsetlabel`
```

Closes: https://github.com/containers/buildah/issues/5063

### DIFF
--- a/cmd/buildah/config.go
+++ b/cmd/buildah/config.go
@@ -47,6 +47,7 @@ type configResults struct {
 	variant                string
 	volume                 []string
 	workingDir             string
+	unsetLabels            []string
 }
 
 func init() {
@@ -98,6 +99,7 @@ func init() {
 	flags.StringVar(&opts.variant, "variant", "", "set architecture `variant` of the target image")
 	flags.StringSliceVarP(&opts.volume, "volume", "v", []string{}, "add default `volume` path to be created for containers based on image (default [])")
 	flags.StringVar(&opts.workingDir, "workingdir", "", "set working `directory` for containers based on image")
+	flags.StringSliceVar(&opts.unsetLabels, "unsetlabel", nil, "remove image configuration label")
 
 	rootCmd.AddCommand(configCommand)
 
@@ -301,6 +303,10 @@ func updateConfig(builder *buildah.Builder, c *cobra.Command, iopts configResult
 			}
 		}
 		conditionallyAddHistory(builder, c, "/bin/sh -c #(nop) LABEL %s", strings.Join(iopts.label, " "))
+	}
+	// unset labels if any
+	for _, key := range  iopts.unsetLabels {
+		builder.UnsetLabel(key)
 	}
 	if c.Flag("workingdir").Changed {
 		builder.SetWorkDir(iopts.workingDir)
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -910,6 +910,10 @@ include:
 
 Unset environment variables from the final image.
 
+**--unsetlabel** *label*
+
+Unset the image label, causing the label not to be inherited from the base image.
+
 **--userns** *how*
 
 Sets the configuration for user namespaces when handling `RUN` instructions.
--- a/docs/buildah-config.1.md
+++ b/docs/buildah-config.1.md
@@ -200,6 +200,10 @@ Note: this setting is not present in the OCIv1 image format, so it is discarded 
 
 Set default *stop signal* for container. This signal will be sent when container is stopped, default is SIGINT.
 
+**--unsetlabel** *label*
+
+Unset the image label, causing the label not to be inherited from the base image.
+
 **--user**, **-u** *user*[:*group*]
 
 Set the default *user* to be used when running containers based on this image.
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -2042,6 +2042,9 @@ func (s *StageExecutor) commit(ctx context.Context, createdBy string, emptyLayer
 	if s.executor.commonBuildOptions.IdentityLabel == types.OptionalBoolUndefined || s.executor.commonBuildOptions.IdentityLabel == types.OptionalBoolTrue {
 		s.builder.SetLabel(buildah.BuilderIdentityAnnotation, define.Version)
 	}
+	for _, key := range s.executor.unsetLabels {
+		s.builder.UnsetLabel(key)
+	}
 	for _, annotationSpec := range s.executor.annotations {
 		annotation := strings.SplitN(annotationSpec, "=", 2)
 		if len(annotation) > 1 {
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -2099,6 +2099,28 @@ _EOF
   done
 }
 
+@test "bud and test --unsetlabel" {
+  _prefetch registry.fedoraproject.org/fedora-minimal
+  run_buildah --version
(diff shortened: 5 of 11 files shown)


flouthoc:
@nalind @arewm PTAL

docs/buildah-build.1.md
```diff
@@ -910,6 +910,10 @@ include:
 
 Unset environment variables from the final image.
 
+**--unsetenv** *label*
```
arewm, line 913:
```suggestion
**--unsetlabel** *label*
```

arewm (changes requested):
This looks really helpful, thanks!

arewm (approved):

imagebuildah/stage_executor.go
```diff
@@ -2036,6 +2036,9 @@ func (s *StageExecutor) commit(ctx context.Context, createdBy string, emptyLayer
 	if s.executor.commonBuildOptions.IdentityLabel == types.OptionalBoolUndefined || s.executor.commonBuildOptions.IdentityLabel == types.OptionalBoolTrue {
 		s.builder.SetLabel(buildah.BuilderIdentityAnnotation, define.Version)
 	}
+	for _, key := range s.executor.unsetLabels {
+		s.builder.UnsetLabel(key)
```
arewm, line 2040: What will happen here if a label is requested to be unset that was never present?

flouthoc: Nothing happens in that case, `no-op`

rhatdan:
Would it make sense to add either glob support or an "all" option?

--unsetlabel=all
or 
--unsetlabel=*

flouthoc:
@rhatdan I don't think its possible because both `buildkit` and `buildah` allows setting label with keys `*` and `all`, may we need a different flag for this use-case.

rhatdan:
You could always check if the string was an exact match, otherwise treat it as a keyword.  Since it is unlikely that users would just set a label of * or all

flouthoc:
> You could always check if the string was an exact match, otherwise treat it as a keyword. Since it is unlikely that users would just set a label of * or all

I am okay with it if everybody agrees with it, although this will leave a edge case intentionally i.e users will not be able only remove key if they are `*` or `all`. @nalind @containers/buildah-maintainers WDYT ?

arewm:
If you want to support removing all labels, I think that would be better to achieve via an explicit call (i.e. https://github.com/containers/buildah/pull/5061) instead of overriding a specific label with a "remove all" functionality. In my opinion, the removal of labels should be very intentional -- and that requests should be very explicit. By making the behavior of the removal of specific labels hinge on the "best matched" label, it introduces uncertainty in the removal process that unintended labels can be removed if a parent image changes _their_ labels.

nalind:
Those are different things, and I'm not aware of any flags that do wildcard matching except when dealing with filenames - capabilities take "all" as an alias, but I _think_ that's it.  For some reason we added `--unsetenv` to `commit` but not `config`, where `--env` is, so we're actually talking about `config`, `commit`, and `build` here.

docs/buildah-build.1.md
```diff
@@ -910,6 +910,10 @@ include:
 
 Unset environment variables from the final image.
 
+**--unsetlabel** *label*
+
+Unset image label from the final image.
```
TomSweeneyRedHat, line 915:
```suggestion
Unset the image label, causing the label not to be added to the final image from the base image.
```

flouthoc:
@nalind i have extended `unsetlabel` to `commit` and `config` as well.

@rhatdan From comments above it seems that there is no consensus for wildcard matching on `--unsetlabel` , please correct if i missed or wrongly interpreted comments above. Maybe for `all` case we can have a different flag like https://github.com/containers/buildah/pull/5061 ( if any need shows up in future )

nalind:
> @nalind i have extended `unsetlabel` to `commit` and `config` as well.

Please don't add this to `commit`.  It fits better with what `config` has always been doing, and I consider our muddying things by putting `--unsetenv` in `commit` instead `config` to have been a mistake.  But maybe I'm in a minority there.

rhatdan:
I am fine with no glob and only "all" to match --cap-drop, --cap-add

cmd/buildah/config.go
```diff
@@ -98,6 +99,7 @@ func init() {
 	flags.StringVar(&opts.variant, "variant", "", "set architecture `variant` of the target image")
 	flags.StringSliceVarP(&opts.volume, "volume", "v", []string{}, "add default `volume` path to be created for containers based on image (default [])")
 	flags.StringVar(&opts.workingDir, "workingdir", "", "set working `directory` for containers based on image")
+	flags.StringSliceVar(&opts.unsetLabels, "unsetlabel", nil, "unset label from final image")
```
nalind, line 102:
```suggestion
	flags.StringSliceVar(&opts.unsetLabels, "unsetlabel", nil, "remove image configuration label")
```

docs/buildah-config.1.md
```diff
@@ -200,6 +200,10 @@ Note: this setting is not present in the OCIv1 image format, so it is discarded
 
 Set default *stop signal* for container. This signal will be sent when container is stopped, default is SIGINT.
 
+**--unsetlabel** *label*
+
+Unset the image label, causing the label not to be inherited from the base image.
```
nalind, line 205: The value for a label can as easily have been set by an earlier invocation of `buildah config`.  Though looking at the documentation for `--label` to find a description to complement, the use case might already be handled.

tests/bud.bats
```diff
@@ -2099,6 +2099,27 @@ _EOF
   done
 }
 
+@test "bud and test --unsetlabel" {
```
nalind, line 2102: _prefetch alpine here.

tests/config.bats
```diff
@@ -81,6 +81,19 @@ function check_matrix() {
   check_matrix 'Config.Entrypoint' '[/bin/sh -c /ENTRYPOINT]'
 }
 
+@test "config --unsetlabel" {
```
nalind, line 84: Looks like it's depending on a previously-run test to have pulled the image.  Needs to `_prefetch` for `--pull=false`, no?

nalind (changes requested):

tests/bud.bats
flouthoc: the base is actually `fedora:minimal` since it contains lot of pre-loaded labels for this test, shouldn't this be `_prefetch fedora:minimal`

tests/config.bats
flouthoc: Done prefetching needed base image

docs/buildah-config.1.md
flouthoc: Yes, also the `unsetlabel` takes place after the `setlabel` in implementation so even if someone does `--label foo=bar --unsetlabel foo` will make sure that `foo` is not there in final image.

nalind: Then (non-blocking) the wording want to not imply that the base image is the only place a label might have come from.

nalind (approved):
LGTM

rhatdan:
/lgtm
Still think we should have support for keyword "all", but will not block.

PR #5067, opened 2023-09-30
Ensure the mount type is always BindMount by default

BenjaminSchubert:
#### What type of PR is this?

/kind bug

Previously, when mounting multiple mounts, if any mount had a `type` specified, it would override the default type for subsequent mounts.

This meant, for example, that having a RUN step like:

```
RUN --mount=type=cache,target=/cache --mount=source=.,target=/src
```

would incorrectly mount the second source as a cache, instead of a bind-mount.

This fix ensures the default is reset between each iteration of the loop, ensuring we get the right mount type.

#### How to verify it

Try to build the provided Dockerfile without the patch applied. This will fail, complaining that `/test/input_file` is a directory.

Now apply the patch and see that it correctly outputs `hello`.

Any variation of mounts before one not specifying `type` would also work.

None, I can create one if required.

Technically, this is a breaking change for users using buildah, as their mounts might change types. I am unsure what to write here, can someone please provide guidance, or update the release notes?

```release-note

```

### DIFF
--- a/run_common.go
+++ b/run_common.go
@@ -1509,8 +1509,6 @@ func checkIfMountDestinationPreExists(root string, dest string) (bool, error) {
 //
 // If this function succeeds, the caller must unlock runMountArtifacts.TargetLocks (when??)
 func (b *Builder) runSetupRunMounts(mountPoint string, mounts []string, sources runMountInfo, idMaps IDMaps) ([]specs.Mount, *runMountArtifacts, error) {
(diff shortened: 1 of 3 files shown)


nalind:
Nice catch, thanks!  LGTM

rhatdan:
/approve
/lgtm
Thanks @BenjaminSchubert

PR #5074, opened 2023-10-04
Use the parent's image ID in the config that we pass to imagebuilder

nalind:
#### What type of PR is this?

/kind cleanup

When initializing imagebuilder's copy of the working container's configuration, fill in the ID of the parent image where the type would have one, instead of the name we used to find that image.

imagebuilder doesn't reference the field in the struct we're passing in, so this change is really just about tidiness.

#### How to verify it

Nothing to verify, really.

None

```release-note
None
```


flouthoc (approved):
LGTM
/lgtm
/approve

PR #5076, opened 2023-10-05
mkcw: remove entrypoint binaries

flouthoc:
Remove generated entrypoint binaries needed to generated confidential container image and add them to `.gitignore`

Following binaries are always generated with build instruction, there shouldn't be any need to keep them in repo since source is already there also when `embed` directory ( in any case ) is empty on `make` it leaves un-tracked changes in the `git status` which looks unneeded.

```release-note
None
```


flouthoc (commented):
~~Hmm binaries are always generted for me on local I wonder why is it needed on repo and CI ?~~

nalind:
Still missing the "why" in the PR description.

flouthoc:
@nalind Done.

nalind:
LGTM, then.

TomSweeneyRedHat:
LGTM
RPM tests are failing to build, I'm not sure if that's expected.

PR #5084, opened 2023-10-11
run: use `internal.GetTempDir` with `os.MkdirTemp` for root and bundle path

flouthoc:
Projects which are using buildah as a library and set `TMPDIR` manually can stumble upon a use-case where `TMPDIR` was set to a relative path.

Such as `export TMPDIR=.` in such case buildah will try to create a temporary root using `Mkdirtemp` leading to a point where bundle is not generated correctly since path was relative.

Following use case can be resolved by making sure that buildah always converts relative path to absolute path and `GetTempDir` does it well.

Example reproducer with podman

```Dockerfile
FROM alpine
RUN echo hello
```

```console
export TMPDIR=.
podman build --no-cache -t test .
```

Expected failure
```console
STEP 1/2: FROM alpine
STEP 2/2: RUN echo hello
error running container: checking permissions on "buildah2341274198": stat buildah2341274198: no such file or directory
ERRO[0000] did not get container create message from subprocess: EOF
Error: building at STEP "RUN echo hello": while running runtime: exit status 1
```

Closes: RHEL-2598

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### How to verify it
* Integration test cannot be added as it can be only reproduced by projects which are using buildah as a library ( such as podman ) however reproducer for podman is given to verify the bug.

* Unit test added for `GetTempDir` to make sure it is converting relative values to `TMPDIR` to absolute

```release-note
run: use internal.GetTempDir instead of os.MkdirTemp
```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -19,6 +19,7 @@ import (
 	"github.com/containers/buildah/copier"
 	"github.com/containers/buildah/define"
 	"github.com/containers/buildah/internal"
+	"github.com/containers/buildah/internal/tmpdir"
 	"github.com/containers/buildah/internal/volumes"
 	"github.com/containers/buildah/pkg/overlay"
 	"github.com/containers/buildah/pkg/parse"
@@ -71,7 +72,7 @@ func setChildProcess() error {
 
 // Run runs the specified command in the container's root filesystem.
 func (b *Builder) Run(command []string, options RunOptions) error {
-	p, err := os.MkdirTemp("", define.Package)
+	p, err := os.MkdirTemp(tmpdir.GetTempDir(), define.Package)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


run_linux.go
```diff
@@ -499,7 +504,7 @@ func setupSlirp4netnsNetwork(config *config.Config, netns, cid string, options [
 		Mask: res.Subnet.Mask,
 	}}
 	netStatus := map[string]nettypes.StatusBlock{
-		slirp4netns.BinaryName: nettypes.StatusBlock{
+		slirp4netns.BinaryName: {
```
flouthoc, line 507: `gofmt` did this change and it not relevant to this PR

rhatdan:
LGTM
@vrothberg @giuseppe @nalind @umohnani8 @Luap99 PTAL

nalind:
Does this also need to touch call sites in image.go and sshagent.go?

flouthoc:
@nalind Good point. Done.

nalind:
LGTM

umohnani8:
LGTM

TomSweeneyRedHat:
Late for the review, but a very nice change @flouthoc

Issue #3474, opened 2021-08-24
Heredoc notation doesn't work


ElSamhaa:
**Description**

Heredoc notation is supported in Dockerfile syntax, however buildah doesn't handle it correctly.

**Steps to reproduce the issue:**
1.Create a Dockerfile with the following content:
```Dockerfile
FROM alpine

COPY <<eof /tmp/hello
hello there
eof
```
2.Run `podman build -t tst . `

**Describe the results you received:**

```
STEP 1: FROM alpine
STEP 2: COPY <<eof /tmp/hello
Error: error building at STEP "COPY <<eof /tmp/hello": checking on sources under "/home/myhome/Play/test/test": copier: stat: "/<<eof": no such file or directory
```

**Describe the results you expected:**
The image should build and have a file `/tmp/hello` that contain `hello world`.

**Output of `rpm -q buildah` or `apt list buildah`:**

`-`

**Output of `buildah version`:**

`-`

**Output of `podman version` if reporting a `podman build` issue:**

```
Version:      3.2.3
API Version:  3.2.3
Go Version:   go1.16.6
Built:        Mon Aug  2 21:39:21 2021
OS/Arch:      linux/amd64
```

**Output of `cat /etc/*release`:**

```
Fedora release 34 (Thirty Four)
NAME=Fedora
VERSION="34 (Workstation Edition)"
ID=fedora
VERSION_ID=34
VERSION_CODENAME=""
PLATFORM_ID="platform:f34"
PRETTY_NAME="Fedora 34 (Workstation Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:34"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f34/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=34
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=34
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Workstation Edition"
[...]

PR #5092, opened 2023-10-19
buildah: add `heredoc` support for `RUN`, `COPY` and `ADD`

flouthoc:
Following PR is a attempt to add `Heredoc` support to buildah.

Once this PR is merged buildah is supposed to honor heredoc syntax while processing containerfiles

Expected syntax to work

```Dockerfile
FROM docker.io/library/python:latest
RUN <<EOF
echo "Hello" >> /hello
echo "World!" >> /hello
EOF

RUN python3 <<EOF
with open("/hello", "w") as f:
    print("Hello", file=f)
    print("Something", file=f)
EOF

RUN ls -a
RUN cat hello
```

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
buildah: add heredoc support for RUN, COPY and ADD
```

Closes: https://github.com/containers/buildah/issues/3474


flouthoc:
Needs: https://github.com/openshift/imagebuilder/pull/264

flouthoc:
This needs: https://github.com/openshift/imagebuilder/pull/264

Other then the imagebuilder PR, following PR is ready to review.

@nalind @rhatdan @TomSweeneyRedHat @vrothberg @giuseppe @containers/buildah-maintainers PTAL

/hold wait till https://github.com/openshift/imagebuilder/pull/264

TomSweeneyRedHat:
Changes here LGTM in general.  I assume you'll also be updating [the Containerfile man page in](https://github.com/containers/common/blob/main/docs/Containerfile.5.md) in c/common with the heredocs examples?

flouthoc:
> Changes here LGTM in general. I assume you'll also be updating [the Containerfile man page in](https://github.com/containers/common/blob/main/docs/Containerfile.5.md) in c/common with the heredocs examples?

@TomSweeneyRedHat Yes I'll open a PR in `c/common` once this gets merged. As of now I am waiting for https://github.com/openshift/imagebuilder/pull/264

flouthoc:
This PR is read for review.

imagebuildah/stage_executor.go
```diff
@@ -362,6 +365,28 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
+				tmpFile, err := os.Create(file.Name)
+				if err != nil {
+					return fmt.Errorf("unable to create tmp file for COPY instruction at %q: %w", contextDir, err)
+				}
+				_, err = tmpFile.WriteString(file.Data)
+				if err != nil {
+					return fmt.Errorf("unable to write contents of heredoc file at %q: %w", tmpFile.Name(), err)
+				}
+				copySources = append(copySources, tmpFile.Name())
+				defer os.Remove(tmpFile.Name())
```
rhatdan, line 384: This should be just after os.Create.

flouthoc: fixed.

Luap99:
I just took a quick looking at this after the demo, why are we importing all these protobuf files from builtkit? I understand that reuse is good but is there a way to do that without this useless protobuf code?

I am worried because a) that adds about 800k binary size and b) because it uses init() function it slows down every single command and allocates more memory, you check it by running buildah with `GODEBUG=inittrace=1`:
```
...
init github.com/gogo/protobuf/types @13 ms, 0.064 ms clock, 13264 bytes, 86 allocs
init github.com/containerd/typeurl @13 ms, 0.010 ms clock, 352 bytes, 4 allocs
init github.com/moby/buildkit/util/stack @13 ms, 0.13 ms clock, 87312 bytes, 58 allocs
init github.com/moby/buildkit/frontend/dockerfile/command @13 ms, 0.002 ms clock, 624 bytes, 2 allocs
init github.com/moby/buildkit/frontend/dockerfile/parser @13 ms, 0.079 ms clock, 23392 bytes, 262 allocs
...
```

To be clear I am not saying this is a blocker but we certainly should look into if we can prevent this from getting pulled in because IMO it should not slow down every single buildah/podman command as it is generally not needed. But I guess this is a lost battle within the go ecosystem.

Luap99:
From a quick check
```diff
diff --git a/vendor/github.com/moby/buildkit/frontend/dockerfile/parser/errors.go b/vendor/github.com/moby/buildkit/frontend/dockerfile/parser/errors.go
index 9f28a5a2e..232ec366b 100644
--- a/vendor/github.com/moby/buildkit/frontend/dockerfile/parser/errors.go
+++ b/vendor/github.com/moby/buildkit/frontend/dockerfile/parser/errors.go
@@ -1,7 +1,6 @@
 package parser
 
 import (
-       "github.com/moby/buildkit/util/stack"
        "github.com/pkg/errors"
 )
 
@@ -41,10 +40,7 @@ func WithLocation(err error, location []Range) error {
        if errors.As(err, &el) {
                return err
        }
-       return stack.Enable(&ErrorLocation{
-               error:    err,
-               Location: location,
-       })
+       return err
 }
 
 func toRanges(start, end int) (r []Range) {
```

This is enough to prevent the huge import chain and protobuf deps.

flouthoc:
@Luap99 I'm not sure if we can change this function like this, cause buildkit is using this so this patch will not be accepted at buildkit's end. Will it work if we move this to some other package different than `parser` and then buildkit uses it ?

nalind (changes requested):
This needs conformance tests, since I'm concerned about whether or not we're adding (or not) newlines at end-of-file for COPY in a way that's consistent with expectations.

imagebuildah/stage_executor.go
```diff
@@ -590,10 +615,60 @@ func (s *StageExecutor) runStageMountPoints(mountList []string) (map[string]inte
...
+	return mountResult, nil
+}
+
 // Run executes a RUN instruction using the stage's current working container
 // as a root directory.
 func (s *StageExecutor) Run(run imagebuilder.Run, config docker.Config) error {
 	logrus.Debugf("RUN %#v, %#v", run, config)
+	args := run.Args
+	if len(run.Files) > 0 {
+		if heredoc := buildkitparser.MustParseHeredoc(args[0]); heredoc != nil {
```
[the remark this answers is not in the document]
flouthoc: We are only using this here to detect if more than one heredoc is present or not, if there is any error in heredoc syntax imagebuilder fails much early with needed error so handling error here does will not have any effect since imagebuilder will fail early. I can add a test for bad case to verify this in CI 
```
Error: parsing main Dockerfile: /work/Dockerfile: EOF: unterminated heredoc
```

imagebuildah/stage_executor.go
```diff
@@ -362,6 +365,28 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
 		preserveOwnership := false
 		contextDir := s.executor.contextDir
+
+		// If we are copying files via heredoc syntax, then
+		// its time to create these temporary files on host
+		// and copy these to container
+		if len(copy.Files) > 0 {
+			copySources := []string{}
+			for _, file := range copy.Files {
+				tmpFile, err := os.Create(file.Name)
```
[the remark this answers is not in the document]
flouthoc: Writing it in `/var/tmp` on the host and then cleaning it up, i am wonder if there is any need to further sanitize the `file.Name` cause heredoc parser rejects funny strings for heredoc names like ones with `..` or `/` .

> But then if we're only creating them so that they can be copied into the rootfs, it might be preferable to generate an archive stream with archive/tar and extract that stream into the rootfs with copier.Put().

`s.builder.Add(` takes path as sources and does all the work related to perms and bit options if i'm gonna use `copier.Put` here, we will have to re-implement the logic or share it with `buildah.Add` therefore I thought using the same logic is more convenient, therefore creating a file and then cleaning it just after `copy` process.

imagebuildah/stage_executor.go
```diff
@@ -590,10 +615,60 @@ func (s *StageExecutor) runStageMountPoints(mountList []string) (map[string]inte
...
+		if heredoc := buildkitparser.MustParseHeredoc(args[0]); heredoc != nil {
+			if strings.HasPrefix(run.Files[0].Data, "#!") || strings.HasPrefix(run.Files[0].Data, "\n#!") {
+				// This is a single heredoc with a shebang, so create a file
+				// and run it.
+				heredocMount, err := s.CreateNeededHeredocMountsForRun(run.Files)
+				if err != nil {
+					return err
+				}
+				args = []string{heredocMount[0].Destination}
+				s.executor.transientMounts = append(s.executor.transientMounts, heredocMount...)
```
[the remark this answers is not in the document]
flouthoc: Done, moved to `options.Mounts` which is cleaned after every step, I'll also add a test to verify this again.

flouthoc:
> This needs conformance tests, since I'm concerned about whether or not we're adding (or not) newlines at end-of-file for COPY in a way that's consistent with expectations.

Conformance infra or `c/image` is unable to parse docker reference generated by docker daemon when a file with heredock is provided causing conformance test unable to process any heredoc files. I think `c/image`'s `docker daemon's` ParseReference needs to be fixed so it can read new buildkit images correctly. I am working on fixing `c/image` however I am manually verified output of `COPY` and `ADD`.

Sample containerfile which conformance test cannot read
```Dockerfile
# syntax=docker/dockerfile:1.4
FROM alpine
COPY <<robots.txt <<humans.txt /test/
(robots content)
robots.txt
(humans content)
humans.txt
```

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
flouthoc: Added a test for this.

Luap99:
> @Luap99 I'm not sure if we can change this function like this, cause buildkit is using this so this patch will not be accepted at buildkit's end. Will it work if we move this to some other package different than `parser` and then buildkit uses it ? (That can still be accepted but will it reduce the protobuf dep ? )

Yeah that is what I am assuming too, the issue is that as soon `github.com/moby/buildkit/util/stack` is important you end up with the the protobuf packages as well which add ton of bloat.
I am not sure if buildkit upstream would care about something like that anyway, The easiest solution would be to have a special build tag for example or I guess we could fork it.
Either way I not saying we should block this but there i definitely room for improvements.

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
nalind: > Writing it in `/var/tmp` on the host and then cleaning it up, i am wonder if there is any need to further sanitize the `file.Name` cause heredoc parser rejects funny strings for heredoc names like ones with `..` or `/` .

path.Base(filepath.ToSlash(file.Name)), then.  We can't predict if that parser will become more forgiving later.

> > But then if we're only creating them so that they can be copied into the rootfs, it might be preferable to generate an archive stream with archive/tar and extract that stream into the rootfs with copier.Put().
> 
> `s.builder.Add(` takes path as sources and does all the work related to perms and bit options if i'm gonna use `copier.Put` here, we will have to re-implement the logic or share it with `buildah.Add` therefore I thought using the same logic is more convenient, therefore creating a file and then cleaning it just after `copy` process.

What are the expected permissions on a COPY heredoc?  0711 seems like an odd choice to hard wire, since that's execute for the owning group and everyone, but read-write-execute for the owner.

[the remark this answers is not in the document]
flouthoc: > path.Base(filepath.ToSlash(file.Name)), then. We can't predict if that parser will become more forgiving later.
I added this change, but can add `copier.Put` if that is better.

imagebuildah/stage_executor.go
```diff
@@ -590,10 +620,59 @@ func (s *StageExecutor) runStageMountPoints(mountList []string) (map[string]inte
 	return stageMountPoints, nil
 }
 
+func (s *StageExecutor) CreateNeededHeredocMountsForRun(files []imagebuilder.File) ([]Mount, error) {
```
nalind, line 623: This method doesn't need to be exported.

flouthoc: fixed, I'll squash dev commits when PR is approved.

imagebuildah/stage_executor.go
```diff
@@ -590,10 +615,60 @@ func (s *StageExecutor) runStageMountPoints(mountList []string) (map[string]inte
...
+	mountResult := []Mount{}
+	for _, file := range files {
+		fileName := file.Name
+		if file.Name == "" || file.Name == "EOF" {
+			fileName = fmt.Sprintf("%s-%d", "buildahheredoc", rand.Intn(10000))
```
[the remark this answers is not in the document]
flouthoc: This is fixed in newer commit.

imagebuildah/stage_executor.go
```diff
@@ -590,10 +615,60 @@ func (s *StageExecutor) runStageMountPoints(mountList []string) (map[string]inte
 	return stageMountPoints, nil
 }
 
+func (s *StageExecutor) CreateNeededHeredocMountsForRun(files []imagebuilder.File) ([]Mount, error) {
+	mountResult := []Mount{}
+	for _, file := range files {
+		fileName := file.Name
```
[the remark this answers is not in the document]
flouthoc: This is fixed in newer commit.

imagebuildah/stage_executor.go
```diff
@@ -590,10 +615,60 @@ func (s *StageExecutor) runStageMountPoints(mountList []string) (map[string]inte
...
+		if file.Name == "" || file.Name == "EOF" {
+			fileName = fmt.Sprintf("%s-%d", "buildahheredoc", rand.Intn(10000))
+		}
+		f, err := ioutil.TempFile("", fileName)
+		if err != nil {
+			return nil, err
+		}
+		if _, err := f.WriteString(file.Data); err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
flouthoc: This is fixed in newer commit.

imagebuildah/stage_executor.go
```diff
@@ -590,10 +615,60 @@ func (s *StageExecutor) runStageMountPoints(mountList []string) (map[string]inte
...
+		}
+		err = os.Chmod(f.Name(), 0711)
```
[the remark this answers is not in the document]
flouthoc: This is fixed in newer commit.

imagebuildah/stage_executor.go
```diff
@@ -590,10 +615,60 @@ func (s *StageExecutor) runStageMountPoints(mountList []string) (map[string]inte
...
+		if err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
flouthoc: This is fixed in newer commit.

flouthoc:
This PR needs a rebase after https://github.com/containers/buildah/pull/5165 since it uses a commit from the same PR.

imagebuildah/stage_executor.go
```diff
@@ -362,6 +364,34 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
+		// and copy these to container
+		if len(copy.Files) > 0 {
+			copySources := []string{}
+			for _, file := range copy.Files {
+				data := file.Data
+				// remove first break line added while parsing heredoc
+				data = strings.TrimPrefix(data, "\n")
+				// add breakline when heredoc ends for docker compat
+				data = data + "\n"
+				tmpFile, err := os.Create(filepath.Join("/var/tmp/", filepath.Base(filepath.ToSlash(file.Name))))
```
nalind, line 378: Use `parse.GetTempDir()` here?

tests/conformance/testdata/Dockerfile.heredoc_copy
```diff
@@ -0,0 +1,11 @@
+# syntax=docker/dockerfile:1.3-labs
+FROM busybox
+RUN echo hello
+COPY <<robots.txt <<humans.txt /test/
```
nalind, line 4: Unless there's a syntax reason it's not allowed, we should be testing COPY of a combination of heredocs and items from the context directory somewhere.

nalind: Maybe also with a `--from` to mix things up a little.

imagebuildah/stage_executor.go
nalind: `os.Create()` is affected by the current process umask.  There needs to be a `tmpFile.Chmod()` in here somewhere.

tests/conformance/testdata/Dockerfile.heredoc_copy
flouthoc: Done.

flouthoc:
@nalind Could you PTAL look again.

imagebuildah/stage_executor.go
```diff
@@ -362,7 +364,60 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
+		if len(copy.Files) > 0 {
+			// If we are copying files from heredoc syntax, there
+			// maybe regular files from context as well so split and
+			// process them differently
+			if len(copy.Src) > len(copy.Files) {
+				regularSources := []string{}
+				for _, src := range copy.Src {
+					// If this source is not a heredoc, then it is a regular file from
+					// build context or from another stage (`--from=`) so treat this differently.
+					if !strings.HasPrefix(src, "<<") && !strings.HasPrefix(src, "<<-") {
```
nalind, line 379: If the value doesn't have `<<` as a prefix, it can't have `<<-` as a prefix, so I don't think we need to be checking for both of these.

imagebuildah/stage_executor.go
```diff
@@ -362,7 +364,60 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
+					return fmt.Errorf("unable to create tmp file for COPY instruction at %q: %w", parse.GetTempDir(), err)
+				}
+				err = tmpFile.Chmod(0644) // 644 is consistent with buildkit
+				if err != nil {
+					tmpFile.Close()
+					return fmt.Errorf("unable to chmod tmp file created for COPY instruction at %q: %w", tmpFile.Name(), err)
+				}
+				defer os.Remove(tmpFile.Name())
+				_, err = tmpFile.WriteString(data)
+				if err != nil {
```
nalind, line 410: `tmpFile` is not closed in this branch.

imagebuildah/stage_executor.go
```diff
@@ -362,7 +364,60 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
...
+				copiesExtend = append(copiesExtend, copyEntry)
+			}
+			copySources := []string{}
+			for _, file := range copy.Files {
+				data := file.Data
+				// remove first break line added while parsing heredoc
+				data = strings.TrimPrefix(data, "\n")
+				// add breakline when heredoc ends for docker compat
+				data = data + "\n"
+				tmpFile, err := os.Create(filepath.Join(parse.GetTempDir(), filepath.Base(filepath.ToSlash(file.Name))))
```
nalind, line 399: After passing a pathname through `filepath.ToSlash()`, the result should be manipulated using the `path` package rather than `filepath`, since `filepath` expects the platform-specific path separator, which isn't `/` on Windows hosts.

tests/conformance/conformance_test.go
```diff
@@ -3003,6 +3003,26 @@ var internalTestCases = []testCase{
...
+			"(dir):test:(dir):humans.txt:mtime",
+			"(dir):test:(dir):robots.txt:mtime",
+			"(dir):test:(dir):file:mtime",
+			"(dir):test:(dir):file:uid", /* buildkit's regular COPY sets UID which is inconsistent with docker and buildah*/
+			"(dir):test:(dir):file:gid", /* buildkit's regular COPY sets GID which is inconsistent with docker and buildah*/
+			"(dir):test2:(dir):humans.txt:mtime",
+			"(dir):test2:(dir):robots.txt:mtime",
+			"(dir):test2:(dir):image_file:mtime",
+			"(dir):test2:(dir):image_file:uid", /* buildkit's regular COPY sets UID which is inconsistent with docker and buildah*/
+			"(dir):test2:(dir):image_file:gid", /* buildkit's regular COPY sets GID which is inconsistent with docker and buildah*/
```
nalind, line 3022: What values does it use?  Why is it okay to ignore the differences?

nalind (changes requested):
Mostly nits and a conformance test question.

tests/conformance/conformance_test.go
flouthoc: Buildkit sets `uid` and `gid` to `1000` with `COPY` instruction in conformance test and `buildah/docker` keeps it 0, i suspect either problem is with buildkit client in conformance test or could be a buildkit issue, this happens on upstream as well so I was not able to spend time on it figure out root cause for it. Since this was not related to this PR hence I did not touch it for now  ( will investigate issue in a different PR )

**Issue reproducer for upstream:**

Following problem is with all the `COPY` ( where `uid` and `gid` are not modified ) related tests included in conformance_test.go

For example a try out example is
```diff
diff --git a/tests/conformance/conformance_test.go b/tests/conformance/conformance_test.go
index 6ec8f3837..e494cb116 100644
--- a/tests/conformance/conformance_test.go
+++ b/tests/conformance/conformance_test.go
@@ -2784,6 +2784,7 @@ var internalTestCases = []testCase{
        {
                name:       "copy-for-user", // flip side of issue #2518
                contextDir: "copy",
+               dockerUseBuildKit: true,
                dockerfileContents: strings.Join([]string{
                        `FROM alpine`,
                        `USER 66:66`,
```

and output fails

```console
[fl@fedora buildah]$ sudo GOROOT=/usr/local/go go test -v -timeout=30m -run TestConformance/copy-for-user ./tests/conformance
=== RUN   TestConformance
=== RUN   TestConformance/copy-for-user
    conformance_test.go:482: 
        	Error Trace:	/home/fl/heredoc-exp/buildah/tests/conformance/conformance_test.go:482
        	            				/home/fl/heredoc-exp/buildah/tests/conformance/conformance_test.go:333
        	            				/home/fl/heredoc-exp/buildah/tests/conformance/conformance_test.go:151
        	Error:      	Filesystem contents differ
        	Test:       	TestConformance/copy-for-user
        	Messages:   	File attributes in both versions have different values:
        	            	File:attr          Docker        buildah
        	            	/script:uid        1000          0
        	            	/script:gid        1000          0
    conformance_test.go:355: Context "/home/fl/heredoc-exp/buildah/tests/conformance/testdata/copy"
    conformance_test.go:367: Dockerfile contents:
        FROM alpine
        USER 66:66
        COPY /script /script
        (no final end-of-line)
=== RUN   TestConformance/copy-for-user-with-chown
--- FAIL: TestConformance (13.23s)
    --- FAIL: TestConformance/copy-for-user (6.97s)
    --- PASS: TestConformance/copy-for-user-with-chown (6.26s)
FAIL
FAIL	github.com/containers/buildah/tests/conformance	13.244s
FAIL
```

nalind: So BuildKit is retaining the uid/gid of the files in the context directory, which are owned by your unprivileged user, instead of setting them to 0?

nalind: I guess it's expecting the client to hand it an archive with ownerships set to 0:0 and go-dockerclient isn't doing that.

nalind: Opened https://github.com/containers/buildah/pull/5171 to generate the buildcontext archive we hand to dockerd with imagebuilder's helper, which sets ownerships in it to 0:0.

flouthoc: Thanks I verified against #5171 and it works, I have removed `uid` and `gid` skips, will rebase after #5171

tests/conformance/conformance_test.go
```diff
@@ -3003,6 +3003,22 @@ var internalTestCases = []testCase{
...
+	{
+		name:              "heredoc-copy",
+		dockerfile:        "Dockerfile.heredoc_copy",
+		dockerUseBuildKit: true,
+		contextDir:        "heredoc",
+		fsSkip: []string{"(dir):test:mtime",
+			"(dir):test2:mtime",
+			"(dir):test:(dir):humans.txt:mtime",
+			"(dir):test:(dir):robots.txt:mtime",
+			"(dir):test:(dir):file:mtime",
```
nalind, line 3015: This file comes from the build context, so I'd expect its datestamp to be consistent between different builds.

flouthoc: correct, this skip is not needed.

imagebuildah/stage_executor.go
```diff
@@ -348,6 +350,7 @@ func (s *StageExecutor) volumeCacheRestore() error {
 // imagebuilder tells us the instruction was "ADD" and not "COPY".
 func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) error {
 	s.builder.ContentDigester.Restart()
```
nalind, line 352: What's the impact of this getting called in between adding heredocs and adding other content on the history and build caching?

flouthoc: Refactored now digester does not restarts between heredoc and regular copy. Before this it was affecting history of the image keeping only the most recent copy in the history.

flouthoc:
@nalind Could you PTAL again.

nalind (commented):
Non-blocking nit, could probably use a rebase.

TomSweeneyRedHat:
Grats on getting this in @flouthoc !  Nice work again!

Issue #5093, opened 2023-09-16
[OSTree] Cannot build rootfs with Containerfile due to dangling files /etc/{hostname,hosts,resolv.conf}


GrabbenD:
### Issue Description

One of requirements for building a OSTree images is moving `/etc` to `/usr/etc` (see: https://ostree.readthedocs.io/en/stable/manual/adapting-existing/). However, this isn't possible in a single stage build with Containerfile as seen here:

```Containerfile
RUN mv /etc /usr/
RUN ls -l /etc
-rw-r--r-- 1 0 0  13 Sep 16 08:54 hostname
-rw-r--r-- 1 0 0 128 Sep 16 08:54 hosts
-rw-r--r-- 1 0 0  24 Sep 16 08:54 resolv.conf
```

These dangling files prevent `$ ostree commit` from deploying:
```console
Deploying tree: Initializing deployment: Preparing /etc: Tree contains both /etc and /usr/etc
```

Furthermore they can't be removed:
```Containerfile
RUN rm -rf /etc
rm: cannot remove '/etc/hosts': Device or resource busy
rm: cannot remove '/etc/hostname': Device or resource busy
rm: cannot remove '/etc/resolv.conf': Device or resource busy
```

Disabling network isn't possible to workaround this:
```Containerfile
RUN --network=none mv /etc /usr/
```

Currently the only working workaround is to export rootfs/ and then modify it outside of Containerfile but ideally this shouldn't be needed to keep everything inside one place (Containerfile)

### Steps to reproduce the issue

Steps to reproduce the issue
1. Create Containerfile with any distro as base, e.g:
   ```Containerfile 
   FROM fedora:latest
   RUN rm -rf /etc
   ```
2. `$ podman build -f Containerfile`

### Describe the results you received

Dangling files exist in `/etc/{hostname,hosts,resolv.conf}` after moving `/etc` to `/usr/etc`

### Describe the results you expected

It should be possible to move contents of `/etc` to `/usr/etc` to comply with OSTree image requirements

### podman info output

```yaml
host:
  arch: amd64
  buildahVersion: 1.31.2
  cgroupControllers:
  - cpuset
  - cpu
  - io
  - memory
  - hugetlb
  - pids
  - rdma
  - misc
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: /usr/bin/conmon is owned by conmon 1:2.1.8-1
    path: /usr/bin/conmon
[...]

PR #5094, opened 2023-10-20
Add --no-hostname option to buildah containers

rhatdan:
Fixes: https://github.com/containers/buildah/issues/5093

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
`buildah build --no-hostname` hostname added to allow for users to not setup /etc/hostname within builds.
```

### DIFF
--- a/cmd/buildah/run.go
+++ b/cmd/buildah/run.go
@@ -27,6 +27,7 @@ type runInputOptions struct {
 	mounts      []string
 	runtime     string
 	runtimeFlag []string
+	noHostname  bool
 	noHosts     bool
 	noPivot     bool
 	terminal    bool
@@ -70,6 +71,7 @@ func init() {
 	// Do not set a default runtime here, we'll do that later in the processing.
 	flags.StringVar(&opts.runtime, "runtime", util.Runtime(), "`path` to an alternate OCI runtime")
 	flags.StringSliceVar(&opts.runtimeFlag, "runtime-flag", []string{}, "add global flags for the container runtime")
+	flags.BoolVar(&opts.noHostname, "no-hostname", false, "do not override the /etc/hostname file within the container")
 	flags.BoolVar(&opts.noHosts, "no-hosts", false, "do not override the /etc/hosts file within the container")
 	flags.BoolVar(&opts.noPivot, "no-pivot", false, "do not use pivot root to jail process inside rootfs")
 	flags.BoolVarP(&opts.terminal, "terminal", "t", false, "allocate a pseudo-TTY in the container")
@@ -141,6 +143,7 @@ func runCmd(c *cobra.Command, args []string, iopts runInputOptions) error {
 		Hostname:         iopts.hostname,
 		Runtime:          iopts.runtime,
 		Args:             runtimeFlags,
+		NoHostname:       iopts.noHostname,
 		NoHosts:          iopts.noHosts,
 		NoPivot:          noPivot,
 		User:             c.Flag("user").Value.String(),
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -645,9 +645,15 @@ Valid _mode_ values are:
 
 Do not use existing cached images for the container build. Build from the start with a new set of cached layers.
 
+**--no-hostname**
+
+Do not create the _/etc/hostname_ file in the container for RUN instructions.
+
+By default, Buildah manages the _/etc/hostname_ file, adding the container's own hostname.  When the **--no-hostname** option is set, the image's _/etc/hostname_ will be preserved unmodified if it exists.
+
 **--no-hosts**
 
-Do not create _/etc/hosts_ for the container.
+Do not create the _/etc/hosts_ file in the container for RUN instructions.
 
 By default, Buildah manages _/etc/hosts_, adding the container's own IP address.
 **--no-hosts** disables this, and the image's _/etc/hosts_ will be preserved unmodified. Conflicts with the --add-host option.
--- a/docs/buildah-run.1.md
+++ b/docs/buildah-run.1.md
@@ -217,9 +217,15 @@ Valid _mode_ values are:
         host, using the loopback interface instead of the tap interface for improved
         performance
 
+**--no-hostname**
+
+Do not create the _/etc/hostname_ file in the container for RUN instructions.
+
+By default, Buildah manages the _/etc/hostname_ file, adding the container's own hostname.  When the **--no-hostname** option is set, the image's _/etc/hostname_ will be preserved unmodified if it exists.
+
 **--no-hosts**
 
-Do not create _/etc/hosts_ for the container.
+Do not create the _/etc/hosts_ file in the container for RUN instructions.
 
 By default, Buildah manages _/etc/hosts_, adding the container's own IP address.
 **--no-hosts** disables this, and the image's _/etc/hosts_ will be preserved unmodified.
--- a/run.go
+++ b/run.go
@@ -86,7 +86,9 @@ type RunOptions struct {
 	Runtime string
 	// Args adds global arguments for the runtime.
 	Args []string
-	// NoHosts use the images /etc/hosts file
+	// NoHostname won't create new /etc/hostname file
+	NoHostname bool
+	// NoHosts won't create new /etc/hosts file
 	NoHosts bool
 	// NoPivot adds the --no-pivot runtime flag.
 	NoPivot bool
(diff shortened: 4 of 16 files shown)


rhatdan:
@Luap99 @flouthoc @vrothberg @nalind @umohnani8 PTAL

docs/buildah-build.1.md
```diff
@@ -645,6 +645,13 @@ Valid _mode_ values are:
 
 Do not use existing cached images for the container build. Build from the start with a new set of cached layers.
 
+**--no-hostname**
+
+Do not create _/etc/hostname_ for the container.
```
nalind, line 650:
```suggestion
Do not create _/etc/hostname_ for RUN instructions.
```

run.go
```diff
@@ -86,6 +86,8 @@ type RunOptions struct {
 	Runtime string
 	// Args adds global arguments for the runtime.
 	Args []string
+	// NoHostname use the images /etc/hostname file
```
nalind, line 89: "image's", but the wording is still kind of awkward, since images aren't required to provide a /etc/hostname, so a given image might not include one.

run_linux.go
```diff
@@ -269,8 +269,7 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 		bindFiles[config.DefaultHostsFile] = hostFile
 	}
 
-	// generate /etc/hostname if the user intentionally did not override
-	if !(contains(volumes, "/etc/hostname")) {
+	if !options.NoHostname && !(contains(volumes, "/etc/hostname")) {
 		if _, ok := bindFiles["/etc/hostname"]; !ok {
```
nalind, line 273: I don't see how `ok` can ever be true here.  Did I miss something?

cmd/buildah/run.go
```diff
@@ -137,10 +139,15 @@ func runCmd(c *cobra.Command, args []string, iopts runInputOptions) error {
 		}
 	}
 
+	if len(iopts.hostname) > 0 && iopts.noHostname {
+		return errors.New("--no-hostname and --hostname conflict, can not be used together")
```
nalind, line 143: One's a setting in the UTS namespace, the other's a file.  _Do_ they conflict?

tests/bud.bats
```diff
@@ -5258,6 +5258,37 @@ _EOF
...
+
+  mytmpdir=${TEST_SCRATCH_DIR}/my-dir
+  mkdir -p ${mytmpdir}
+  cat > $mytmpdir/Containerfile << _EOF
+from alpine
+run cat /etc/hostname
+_EOF
+
+  run_buildah build --no-cache -t testbud \
+                  $WITH_POLICY_JSON --file ${mytmpdir}/Containerfile .
```
nalind, line 5274: Why try to use the current directory as the build context?
```suggestion
                  $WITH_POLICY_JSON ${mytmpdir}
```

tests/bud.bats
```diff
@@ -5258,6 +5258,37 @@ _EOF
...
+  assert "${lines[2]}" != "localhost" "Should be set to something other then localhost"
+
+  run_buildah build --no-hostname --no-cache -t testbud \
+                  $WITH_POLICY_JSON --file ${mytmpdir}/Containerfile .
```
nalind, line 5278:
```suggestion
                  $WITH_POLICY_JSON ${mytmpdir}
```

tests/bud.bats
```diff
@@ -5258,6 +5258,37 @@ _EOF
...
+  assert "${lines[2]}" == "localhost" "Should be set to localhost"
+
+  cat > $mytmpdir/Containerfile << _EOF
+from alpine
+RUN mv /etc /usr/
+RUN ls -l /etc
```
nalind, line 5284: The content of this file will always be the same.  Why are we generating it every time this test runs?

tests/bud.bats
```diff
@@ -5258,6 +5258,37 @@ _EOF
...
+_EOF
+  run_buildah 1 build --network=none --no-hostname --no-cache -t testbud \
+                  $WITH_POLICY_JSON --file ${mytmpdir}/Containerfile .
```
nalind, line 5287:
```suggestion
                  $WITH_POLICY_JSON ${mytmpdir}
```

tests/bud.bats
```diff
@@ -5258,6 +5258,37 @@ _EOF
...
+  assert "$output" =~ ".*ls: /etc: No such file or directory" "/etc/ directory should be gone"
+
```
nalind, line 5289:
```suggestion
```

nalind (changes requested):
Mostly nits.

run_linux.go
rhatdan: I think this is for the use case where a user does `-v /tmp/hostname:/etc/hostname`

cmd/buildah/run.go
rhatdan: I believe --hostname attempts to set the hostname field in /etc/hostname, if you don't think this would confuse the user, then I am fine with removing it.

tests/bud.bats
rhatdan: This file has different content then the previous?

run_linux.go
nalind: I still don't follow.  If I'm reading it right, the `bindFiles` map has, at most, one item in it: a value for the `config.DefaultHostsFile` key.  A `-v` flag from the command line would show up in `options.Volumes` or `b.CommonBuildOpts.Volumes`, or maybe both if we're in `build`.

`contains()` probably shouldn't exist when we got rid of for `github.com/containers/buildah/util.StringInSlice()` in favor of `github.com/containers/common/pkg/util.StringInSlice()`.

tests/bud.bats
nalind: Use two static files?

cmd/buildah/run.go
TomSweeneyRedHat: I'd remove the test

docs/buildah-build.1.md
```diff
@@ -645,9 +645,17 @@ Valid _mode_ values are:
 
 Do not use existing cached images for the container build. Build from the start with a new set of cached layers.
 
+**--no-hostname**
+
+Do not create _/etc/hostname_ for RUN instructions.
+
+By default, Buildah manages _/etc/hostname_, adding the container's own hostname.
+**--no-hostname** disables this, and the image's _/etc/hostname_ will be preserved unmodified if it exists.
+Conflicts with the --hostname option.
```
TomSweeneyRedHat, line 654: If you remove the test above, you should remove this line too.

docs/buildah-build.1.md
```diff
@@ -645,9 +645,17 @@ Valid _mode_ values are:
 
 Do not use existing cached images for the container build. Build from the start with a new set of cached layers.
 
+**--no-hostname**
+
+Do not create _/etc/hostname_ for RUN instructions.
```
TomSweeneyRedHat, line 650: Soft suggestion
```suggestion
Do not create the _/etc/hostname_ file in the container for RUN instructions.
```

docs/buildah-build.1.md
```diff
@@ -645,9 +645,17 @@ Valid _mode_ values are:
...
+
+By default, Buildah manages _/etc/hostname_, adding the container's own hostname.
```
TomSweeneyRedHat, line 652: I'm on the fence about adding the word `file` after /etc/hostname here and below.

docs/buildah-run.1.md
```diff
@@ -217,6 +217,14 @@ Valid _mode_ values are:
...
         performance
 
+**--no-hostname**
+
+Do not create _/etc/hostname_ for the container.
+
+By default, Buildah manages _/etc/hostname_, adding the container's own hostname.
+**--no-hostname** disables this, and the image's _/etc/hostname_ will be preserved unmodified if it exists.
+Conflicts with the --hostname option.
+
```
TomSweeneyRedHat, line 227: replicate any comments taken from above

run_linux.go
rhatdan: That was existing code, so I will remove the check.

rhatdan:
@Luap99 @flouthoc @vrothberg @nalind @umohnani8 @TomSweeneyRedHat PTAL

umohnani8:
LGTM

@rhatdan I think it needs a rebase though

TomSweeneyRedHat:
LGTM

flouthoc (approved):
Restarted flaking test
LGTM
/lgtm
/hold

vrothberg:
Reminder to also open Podman PRs when adding new build flags.  Currently, Podman CI breaks when vendoring Buildah.

PR #5111, opened 2023-10-27
Use mask definitions from containers/common @rhatdan

rhatdan:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
The /sys/devices/virtual/powercap is now masked over during builds.
```


mheon:
LGTM

vendor/github.com/containers/common/pkg/config/default.go
```diff
@@ -37,6 +37,30 @@ const (
...
+	DefaultMaskedPaths = []string{
+		"/proc/acpi",
+		"/proc/kcore",
+		"/proc/keys",
+		"/proc/latency_stats",
+		"/proc/sched_debug",
+		"/proc/scsi",
+		"/proc/timer_list",
+		"/proc/timer_stats",
+		"/sys/dev/block",
```
nalind, line 49: This appears to allow the rest of `/sys/dev`, which we didn't before.

rhatdan: Podman switched from /sys/dev to /sys/dev/block a while ago.  So I went with the Podman default.

nalind: Is it correct for build-time as well as run-time?

rhatdan: I would figure.  Are concerned about access to content in /sys/dev?  Note this is looser then buildah had before.

nalind: It being looser is what prompted my question.  https://github.com/containers/podman/pull/6957 masked all of /sys/dev without being more specific about what under it mattered, https://github.com/containers/podman/pull/8408 narrowed the default to /sys/dev/block, and https://github.com/containers/podman/issues/12746 is asking to unmask the only part under it that we're still going to be masking.

rhatdan: Yup, just going for consistency now.

PR #5112, opened 2023-10-27
Log the platform for build errors during multi-platform builds

nalind:
#### What type of PR is this?

/kind other

When a multi-platform build fails and we hand an error back, add the platform to the error's context so that it's easier to tell at the end of the log which platform the build failed for.

#### How to verify it

Build with multiple `--platform` values (e.g. `--platform linux/arm64,linux/amd64,linux/ppc64le,linux/s390x`), and have one or more of them fail.  Example:
```Dockerfile
FROM fedora
RUN test `arch` = x86_64
```

None

Nothing earth-shattering, just something to make troubleshooting a little bit easier.

```release-note
When a build which uses multiple --platform flags fails, the error message will indicate which platform experienced the failure.
```


rhatdan:
LGTM

flouthoc (approved):
LGTM
/lgtm
/approve

PR #5113, opened 2023-10-30
Mask /sys/devices/virtual/powercap by default

mheon:
This mimics #5111 but without the vendor update, which is very large on this branch.

#### What type of PR is this?

/kind bug

```release-note
NONE
```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -1036,6 +1036,7 @@ func setupMaskedPaths(g *generate.Generator) {
 		"/sys/firmware",
 		"/sys/fs/selinux",
 		"/sys/dev",
+		"/sys/devices/virtual/powercap",
 	} {
 		g.AddLinuxMaskedPaths(mp)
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
I'ze confused, I don't see this change in #5111  But I could be too distracted by the meeting I'm in.

mheon:
5111 vendored c/common and changed to using a c/common list of masked paths, which included this.

nalind:
It switched from masking a list of paths maintained in this repository to a list that was recently added to the common repo.  By doing so, it replaced /sys/dev with /sys/dev/block in that set of paths, and we're not doing that here, but as this PR is already only aiming to pull in part of that set of changes, I guess it's fine.
/lgtm
/hold

TomSweeneyRedHat:
Thanks for the explanation, now I see it.
LGTM

PR #5120, opened 2023-10-31
[release-1.28] run the cross-compile test on M1 MacOS, not Intel

nalind:
#### What type of PR is this?

/kind failing-test 

* Cirrus no longer supports running tests on Intel-based versions of MacOS, so we need to switch this test to use an M1-based version.
* The registry at registry.centos.org has been decommissioned, so update tests that referred to images that it hosted to use docker.io/library/centos in its place.

#### How to verify it

* The cross-compile test should pass again.
* Tests that used centos:7 as a base image should pass again.

None

Hopefully this will get #5119 passing.

```release-note
None
```

### DIFF
--- a/tests/conformance/testdata/Dockerfile.reusebase
+++ b/tests/conformance/testdata/Dockerfile.reusebase
@@ -1,4 +1,4 @@
-FROM registry.centos.org/centos/centos:centos7 AS base
+FROM docker.io/library/centos:centos7 AS base
 RUN touch -t 201701261659.13 /1
 ENV LOCAL=/1
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -140,7 +140,7 @@ cross_build_task:
         $CIRRUS_CRON != 'multiarch'
 
     osx_instance:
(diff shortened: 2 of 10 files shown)


rhatdan:
LGTM

tests/conformance/testdata/Dockerfile.reusebase
```diff
@@ -1,4 +1,4 @@
-FROM registry.centos.org/centos/centos:centos7 AS base
+FROM docker.io/library/centos:centos7 AS base
```
edsantiago, line 1: Quick note on this: [tests/registries.conf](https://github.com/containers/buildah/blob/d18a4f30bcdb992d10610f6b995b4f2a9cc27bbb/tests/registries.conf) maps this to `mirror.gcr.io`, another registry we have no control over. It might be safer to `skopeo copy --all` the desired centos images, and replace `docker.io/library` with `quay.io/libpod`. Your call.

TomSweeneyRedHat:
LGTM

PR #5129, opened 2023-11-01
image: replace `GetStoreImage` with `ResolveReference` and  bump `c/image` to `373c52a9466f`

flouthoc:
`GetStoreImage` is deprecated after https://github.com/containers/image/pull/2056, similar to PR: https://github.com/containers/podman/pull/20560

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
vendor: bump `c/image` to `373c52a9466f`
```


flouthoc:
@containers/buildah-maintainers @giuseppe @vrothberg PTAL

rhatdan:
LGTM

flouthoc:
Rebased.

commit.go
```diff
@@ -371,8 +369,8 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
 		}
 	}
 
-	img, err := is.Transport.GetStoreImage(b.store, dest)
-	if err != nil && !errors.Is(err, storage.ErrImageUnknown) {
+	_, img, err := is.ResolveReference(dest)
+	if err != nil && err.Error() != fmt.Sprintf("trying to resolve a non-%s: reference %q", is.Transport.Name(), transports.ImageName(dest)) {
```
mtrmac, line 373: This really should use `errors.{As,Is}`. `c/image/v5/storage.ErrNoSuchImage` should work.

mtrmac (commented):
ACK overall, it seems that the surrounding code really has no use for the other return value.

commit.go
nalind, line 373: Errors coming back from the new API can't be checked using `errors.Is()` or `errors.As()`?  That seems like a step backward.

mtrmac: I would expect `errors.Is(err, istorage.ErrNoSuchImage)` to work, and it seems to have worked correctly in CRI-O.

mtrmac: That’s not an “image not found” path, that’s “invalid use of this API” path, though.

I definitely need to fix the documentation and make the error type a maintained API guarantee, I’ll do that today.

mtrmac: Oops, I’m being stupid everywhere in this PR review, sorry about that.

I thought this is an “image not found” check, but this really is trying to use the API with a non-c/storage reference.

It seems to me that if `Commit` is called with a non-c/storage destination, the following code touching `img.ID` and `RemoveNames` and `IIDFile` should not be entered at all, because the image was not written there and this is modifying something unrelated to the commit. So; the caller here should check for a non-c/storage destination and not call `ResolveReference` at all in those cases.

flouthoc: @mtrmac I have tried following diff and it does not works as it should.

```diff
diff --git a/commit.go b/commit.go
index fba35b2c3..6d4c27c78 100644
--- a/commit.go
+++ b/commit.go
@@ -3,6 +3,7 @@ package buildah
 import (
        "context"
        "encoding/json"
+       "errors"
        "fmt"
        "io"
        "os"
@@ -370,7 +371,7 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
        }
 
        _, img, err := is.ResolveReference(dest)
-       if err != nil && err.Error() != fmt.Sprintf("trying to resolve a non-%s: reference %q", is.Transport.Name(), transports.ImageName(dest)) {
+       if err != nil && !errors.Is(err, is.ErrNoSuchImage) {
                return imgID, nil, "", fmt.Errorf("locating image %q in local storage: %w", transports.ImageName(dest), err)
        }
        if err == nil {
diff --git a/tests/test_runner.sh b/tests/test_runner.sh
index a93654b37..021e2dd27 100755
--- a/tests/test_runner.sh
+++ b/tests/test_runner.sh
@@ -19,4 +19,4 @@ function execute() {
 TESTS=${@:-.}
 
 # Run the tests.
-execute time bats --tap $TESTS
+execute time bats --tap $TESTS --filter "\"blobcache-commit\""
```

To run the test
```console
sudo ./tests/test_runner.sh
```

flouthoc: The error which fails due to exact this match

```console
# Error: committing container "pause-working-container" to "dir:/var/tmp/buildah_tests.sv7ziw/doomed": locating image "dir:/var/tmp/buildah_tests.sv7ziw/doomed" in local storage: trying to resolve a non-containers-storage: reference "dir:/var/tmp/buildah_tests.sv7ziw/doomed"
# [ rc=125 (** EXPECTED 0 **) ]
```

mtrmac: @flouthoc Yeah, I was being an idiot. See above, and elsewhere about `dest2`.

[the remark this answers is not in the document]
flouthoc: I tried before this and it does not works for cases when `ImageReference` is OCI directory.

flouthoc:
@mtrmac @nalind PTAL I think now patch implements https://github.com/containers/buildah/pull/5129#discussion_r1380356271 and removes the need for error matching.

commit.go
```diff
@@ -371,32 +369,33 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
...
-		for _, name := range img.Names {
-			if nameToRemove != "" && strings.Contains(name, nameToRemove) {
-				toPruneNames = append(toPruneNames, name)
+	dref := dest.DockerReference()
+	if dref != nil {
+		// Check if reference is on the storage
+		if _, err := b.store.Image(dref.String()); err == nil {
+			// resolve reference
+			dest2, img, err := is.ResolveReference(dest)
+			if err != nil {
```
mtrmac, line 378: This doesn’t work with ID-only references (or correctly look up name@id references); and it still touches c/storage for non-c/storage destinations like a direct commit to `docker://`.

```go
if dest.Transport().Name() == is.Transport.Name() {
   dest2, img, err := is.ResolveReference(dest)
   if err != nil { /* fail, “unknown image” is not really an expected error after we hav just created it */ }
   dest = dest2 // immediately, to make sure all following code uses exactly the same image
}
```
similarly to the check around `AdditionalTags`.

flouthoc: Done.

commit.go
```diff
@@ -358,45 +356,50 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
...
-		}
-		if options.IIDFile != "" {
-			if err = os.WriteFile(options.IIDFile, []byte("sha256:"+img.ID), 0644); err != nil {
-				return imgID, nil, "", err
+	dref := dest.DockerReference()
+	if dref != nil {
+		// Check if reference is on the storage
+		if _, err := b.store.Image(dref.String()); err == nil {
+			if dest.Transport().Name() == is.Transport.Name() {
+				// resolve reference
```
mtrmac, line 379: (I don’t understand what this comment adds over the code alone. Is this trying to document something non-obvious?)

commit.go
```diff
@@ -358,45 +356,50 @@ func (b *Builder) Commit(ctx context.Context, dest types.ImageReference, options
...
-			}
-			dest = dest2
-		}
-		if options.IIDFile != "" {
-			if err = os.WriteFile(options.IIDFile, []byte("sha256:"+img.ID), 0644); err != nil {
-				return imgID, nil, "", err
+	dref := dest.DockerReference()
+	if dref != nil {
+		// Check if reference is on the storage
+		if _, err := b.store.Image(dref.String()); err == nil {
```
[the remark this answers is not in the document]
flouthoc: @mtrmac Done, but question how does matching transport verifies that image is present on c/storage ?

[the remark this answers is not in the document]
mtrmac: You’re right that just matching `transport.Name()` isn’t a presence check.

But this is `Commit`: the image was _just_ written to `dest` by the `retryCopyImage` line above, so the image must be present; if it isn’t, either something has gone horribly wrong. Or maybe the image was concurrently removed by another process in the meantime — but in that case we can’t correctly handle `IIDFile`, so this really should fail instead of silently continuing and ignoring that option.

mtrmac (commented):
LGTM. Thanks for dealing with this breakage!

nalind:
I don't love moving from a tagged release to a commit, but otherwise it's fine.

mtrmac:
From my POV this is not urgent — it can be included in the next vendor dance. It’s just that with the added deprecation notices, these changes will probably _have to_ be included with that vendor dance.

I’m also open to not immediately adding the depreciation notices, so that there is a tagged version of c/image which contains `ResolveImage`, and callers can update it without the urgency of a deprecation warning breaking a build. OTOH that deprecation notice / build breakage usually is what motivates the migration… and, as this PR shows, revisiting the call sites to see if they should use `ResolveReference` is probably worthwhile.

flouthoc:
> I don't love moving from a tagged release to a commit, but otherwise it's fine.

I'll vendor `c/image` as soon as it out, just need `c/image` to be bumped so i can add integration test for ( https://github.com/containers/image/pull/1645 ) to buildah upstream.

flouthoc:
@mtrmac I have checked `podman` already and bumped `c/image` there https://github.com/containers/podman/pull/20560 and for buildah this PR will do the trick other than this there is only skopeo which can be affected by this PR ( not sure if there are any other tools )

mtrmac:
Skopeo does not call the deprecated functions directly.

There’s also c/common, and at a first glance the two calls probably _should_ use `ResolveReference` to create a storage reference which both fixes an image ID and references a specific digest (instead of creating an ID-only `image.storageReference`). That would affect the 11 callers of `libimage.Image.StorageReference`. From a quick check that should all work fine, but it’s possible I have missed a corner case.

mtrmac:
@flouthoc Do you need a new c/image release soonish, then, or should this all happen with the Podman 4.8 vendor dance?

flouthoc:
> @flouthoc Do you need a new c/image release soonish, then, or should this all happen with the Podman 4.8 vendor dance?

@mtrmac I don't think there is any need for `c/image` release from my-end so it can wait.

mtrmac:
#5143 will need a c/image updated with these changes, so I’d prefer merging this before the main vendor dance.

mtrmac:
> I don't love moving from a tagged release to a commit, but otherwise it's fine.

The Podman 4.8 vendoring process was discussed today in context of #5143 , and the plan for that PR is to refer to a commit in dependencies, with the expectation that vendor dance will clean that up soon.

@flouthoc I am seeing “Unit tests w/overlay” consistently falling on a total timeout, with about the same place. This seems to also affect other PRs.

flouthoc:
I think timout was a temp flake.

PR #5146, opened 2023-11-08
conformance: use require.NoErrorf() more

nalind:
#### What type of PR is this?

/kind cleanup

In conformance tests, when we don't expect an error to be set, use `require.NoErrorf()` as a check instead of the more general `require.Nil()`, which was both less specific and didn't expect a format string.

#### How to verify it

It's all in the tests.

None

```release-note
None
```


flouthoc (approved):
LGTM
/lgtm
/hold

rhatdan:
LGTM

PR #5147, opened 2023-11-08
createConfigsAndManifests: clear history before cw-specific logic

nalind:
#### What type of PR is this?

/kind cleanup

*  During a commit, clear the docker-format rootfs and history information before we clear other fields in the config for confidential workload cases, so that it's easier for maintainers to track that it exactly parallels what we're doing with the OCI format data.
* Simplify `pkg/parse.Platform()` a bit.
* Where we use `"scratch"` as a constant, use a named constant instead of just sprinkling the string in assorted places.

#### How to verify it

Nothing should start breaking, but I added a unit test to make sure the constant we're using doesn't get changed out from under us.

None

```release-note
None
```

### DIFF
--- a/image.go
+++ b/image.go
@@ -286,6 +286,18 @@ func (i *containerImageRef) createConfigsAndManifests() (v1.Image, v1.Manifest, 
 	}
 	// Always replace this value, since we're newer than our base image.
 	dimage.Created = created
+	// Clear the list of diffIDs, since we always repopulate it.
+	dimage.RootFS = &docker.V2S2RootFS{}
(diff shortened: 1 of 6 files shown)


rhatdan:
LGTM

nalind:
Rebased.

TomSweeneyRedHat:
Changes LGTM in general, tests are still barking at you though @nalind

nalind:
A lot of our tests end up hitting quay.io at some point, and there were some server-side problems yesterday and earlier today.

PR #5151, opened 2023-11-09
[WIP] default to zstd:chunked

rhatdan:
[NO NEW TESTS NEEDED]

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
NONE
```

### DIFF
--- a/contrib/cirrus/setup.sh
+++ b/contrib/cirrus/setup.sh
@@ -46,6 +46,8 @@ EOF
         then
             showrun setsebool -P container_manage_cgroup true
         fi
+	sed -e 's/^pull_options.*=.*/pull_options = {enable_partial_images = \"true\", use_hard_links = \"false\", ostree_repos=""}/' \
+            /usr/share/containers/storage.conf > /etc/containers/storage.conf
         ;;
     debian)
         if [[ "$1" == "conformance" ]]; then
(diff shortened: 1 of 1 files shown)


rhatdan:
@giuseppe Is the failure here the one you have been working with @mtrmac on?

Looks like buildah is pulling an image when it has it in cache?

giuseppe:
yes that is one of the issues that containers/image#1980 is addressing

rhatdan:
Should we move to tests everything with zstd:chunked at this point?  I think it is time to turn this on in testing to make sure everything works going forward.

mtrmac:
Repeated zstd:chunked pushes are still known to be inefficient. I don’t think that’s exactly a _blocker_ to switching tests to zstd:chunked, but unless we are going to institute some kind of matrix where various option values are tested simultaneously, I think testing the most production-relevant values is more valuable.

rhatdan:
Except the goal is to change the default in Fedora 41 to zstd:chunked.  I have already changed the default in Rawhide, to move this forward.  If we continue to default to gzip, we will never get to zstd:chunked.

PR #5143, opened 2023-11-07
WIP: Add --compat-auth-file to login and logout

mtrmac:
#### What type of PR is this?

> /kind feature

#### What this PR does / why we need it:

Add `--compat-auth-file` to login/logout, allowing to update Docker-compatible files.

This _mostly_ inherits the feature from c/common/pkg/auth.

Completely untested at this point.

#### How to verify it

Existing CI for existing features.
TBD tests that the updated files can be correctly consumed by Docker.

#### Which issue(s) this PR fixes:

None

#### Does this PR introduce a user-facing change?

```release-note
`login` and `logout` now support editing Docker-compatible config files, using a `--compat-auth-file` option.
```

PR #5153, opened 2023-11-10
blobcacheinfo,test: blobs must be resued when pushing across registry

flouthoc:
It seems we try to reuse blobs only for the specified registry, however we can have valid known compressed digests across registry as well following pr attempts to test that use-case where feature was implemented in https://github.com/containers/image/pull/1645

----

### How to manually test this ?

```console
$ skopeo copy docker://registry.fedoraproject.org/fedora-minimal docker://quay.io/fl/test:some-tag
$ buildah pull registry.fedoraproject.org/fedora-minimal
$ buildah tag registry.fedoraproject.org/fedora-minimal quay.io/fl/test
$ buildah push quay.io/fl/test
```

```console
Copying blob a3497ca15bbf skipped: already exists
Copying config f7e02de757 done
Storing signatures
```

Testing: https://github.com/containers/image/pull/1645

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```

### DIFF
--- a/tests/blobcache.bats
+++ b/tests/blobcache.bats
@@ -55,6 +55,41 @@ function _check_matches() {
         expect_output --from="$unmatched" "$5"  "$6 should not match"
 }
 
+# Integration test for https://github.com/containers/image/pull/1645
+@test "blobcache: blobs must be reused when pushing across registry" {
+	start_registry
+	run_buildah login --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth --username testuser --password testpassword localhost:${REGISTRY_PORT}
+	outputdir=${TEST_SCRATCH_DIR}/outputdir
+	mkdir -p ${outputdir}
+	run podman run --rm --mount type=bind,src=${TEST_SCRATCH_DIR}/test.auth,target=/test.auth,Z --mount type=bind,src=${outputdir},target=/output,Z --net host quay.io/skopeo/stable copy --preserve-digests --authfile=/test.auth --tls-verify=false docker://registry.fedoraproject.org/fedora-minimal dir:/output
+
+	run_buildah rmi --all -f
+	run_buildah pull dir:${outputdir}
+	run_buildah images -a --format '{{.ID}}'
+	cid=$output
+	run_buildah --log-level debug push --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth $cid docker://localhost:${REGISTRY_PORT}/test
+	# must not contain "Skipping blob" since push must happen
+	assert "$output" !~ "Skipping blob"
+
+	# Clear local image and c/image's blob-info-cache
+	run_buildah rmi --all -f
+	if is_rootless;
+	then
+		run rm $HOME/.local/share/containers/cache/blob-info-cache-v1.sqlite
+		assert "$status" -eq 0 "status of `run rm $HOME/.local/share/containers/cache/blob-info-cache-v1.sqlite` must be 0"
+	else
+		run rm /var/lib/containers/cache/blob-info-cache-v1.sqlite
+		assert "$status" -eq 0 "status of `run rm /var/lib/containers/cache/blob-info-cache-v1.sqlite` must be 0"
+	fi
+
+	# In first push blob must be skipped after vendoring https://github.com/containers/image/pull/1645
+	run_buildah pull dir:${outputdir}
+	run_buildah images -a --format '{{.ID}}'
+	cid=$output
+	run_buildah --log-level debug push --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth $cid docker://localhost:${REGISTRY_PORT}/test
+	expect_output --substring "Skipping blob"
+}
+
 @test "blobcache-commit" {
 	blobcachedir=${TEST_SCRATCH_DIR}/cache
 	mkdir -p ${blobcachedir}


flouthoc:
@containers/buildah-maintainers @mtrmac PTAL

flouthoc:
Tests are green.

tests/blobcache.bats
```diff
@@ -55,6 +55,29 @@ function _check_matches() {
...
+	run_buildah tag registry.fedoraproject.org/fedora-minimal localhost:${REGISTRY_PORT}/test
+	run_buildah --log-level debug push --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth localhost:${REGISTRY_PORT}/test docker://localhost:${REGISTRY_PORT}/test
+	# must not contain "Skipping blob" since push must happen
+	assert "$output" !~ "Skipping blob"
+	run_buildah --log-level debug push --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth localhost:${REGISTRY_PORT}/test docker://localhost:${REGISTRY_PORT}/test
+	expect_output --substring "Skipping blob"
+
+	# Clear local image and c/image's blob-info-cache
+	run_buildah rmi --all -f
+	run rm /var/lib/containers/cache/blob-info-cache-v1.sqlite
```
nalind, line 72: The result of this run() isn't checked, and I'd be surprised if it succeeds for non-root users.

flouthoc: @nalind New commit addresses this comment.

tests/blobcache.bats
```diff
@@ -55,6 +55,36 @@ function _check_matches() {
         expect_output --from="$unmatched" "$5"  "$6 should not match"
 }
 
+# Integration test for https://github.com/containers/image/pull/1645
+@test "blobcache: blobs must be reused when pushing across registry" {
+	start_registry
+	run_buildah login --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth --username testuser --password testpassword localhost:${REGISTRY_PORT}
+	run_buildah pull registry.fedoraproject.org/fedora-minimal
+	run_buildah tag registry.fedoraproject.org/fedora-minimal localhost:${REGISTRY_PORT}/test
+	run_buildah --log-level debug push --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth localhost:${REGISTRY_PORT}/test docker://localhost:${REGISTRY_PORT}/test
```
mtrmac, line 64: Is this actually correct?

This push could create a _different_ compressed configuration than the original one at `r.fp.org`. And if so, after we clear the BIC, I don’t know how it would be matched on the second push.

The “How to manually test this” paragraph uses `skopeo copy` to create a possibly-matched copy on the second registry, and IIRC that’s necessary.

mtrmac (commented):
Unit tests are frequently timing out, compare https://github.com/containers/buildah/pull/5143/commits/881337eb43170489ecc210189399774cbbe26858 in https://github.com/containers/buildah/pull/5143 .

tests/blobcache.bats
flouthoc: @mtrmac After clearing the BIC, we pull again which populates the BIC with candidate with a specific case which is candidate with blob and `UnknownLocation: True`.

I think clearing BIC and image from local produces similar effect to `skopeo copy` , doesn't it ? and further this can be verified by running the same test with older c/image where feature is not present, there test fails.

Given that' I agree `skopeo copy` is more clean and I'll amend this patch to use `skoepo copy` instead.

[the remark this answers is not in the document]
flouthoc: ack

tests/blobcache.bats
```diff
@@ -55,6 +55,36 @@ function _check_matches() {
         expect_output --from="$unmatched" "$5"  "$6 should not match"
 }
 
+# Integration test for https://github.com/containers/image/pull/1645
+@test "blobcache: blobs must be reused when pushing across registry" {
+	start_registry
+	run_buildah login --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth --username testuser --password testpassword localhost:${REGISTRY_PORT}
+	run_buildah pull registry.fedoraproject.org/fedora-minimal
```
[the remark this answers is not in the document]
flouthoc: This SGTM i'll try this.

tests/blobcache.bats
mtrmac: The variable here is whether `buildah push` produces a bit-for-bit-identical layer to the original `r.fp.org` one. That might be _consistently_ true today, but it will very likely eventually not be true (after the implementations and/or implementation versions between the Buildah being tested, and the thing that uploaded to `r.fp.org`, drift).

With an empty BIC, pulling records (uncompressed, compressed as in `r.fp.org`, location = r.fp.org) in BIC.
Then, the push triggers a search for (compressed as in `r.fp.org`, unknown location); that matches only if the push to `localhost` produces a bit-for-bit identical layer.

I guess that it does right now, but that can change any time.

flouthoc: @mtrmac PTAL again, not used `skopeo copy` but now image is built locally and pulled from a dir, where this should never happen. I have tested this test with older buildah ( where this fails ) and newer buildah ( with vendor ) where this passes.

tests/blobcache.bats
```diff
@@ -55,6 +55,37 @@ function _check_matches() {
         expect_output --from="$unmatched" "$5"  "$6 should not match"
 }
 
+# Integration test for https://github.com/containers/image/pull/1645
+@test "blobcache: blobs must be reused when pushing across registry" {
+	start_registry
+	run_buildah login --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth --username testuser --password testpassword localhost:${REGISTRY_PORT}
+	run_buildah build $WITH_POLICY_JSON -t exp -f $BUDFILES/simple-multi-step/Containerfile
+	run_buildah push $WITH_POLICY_JSON --format docker exp dir:${TEST_SCRATCH_DIR}/exp
```
mtrmac, line 63: To confirm I am reading this right, this creates a _compressed_ representation, doesn’t it? (And that is fine and required, otherwise, after clearing the BIC, there would be no knowledge of a compressed digest.)

mtrmac, line 63: (Is the `--format docker` doing something nonobvious here? It seems to me that it should neither help nor hurt, so I’m wondering whether I am missing something that makes it necessary.)

tests/blobcache.bats
```diff
@@ -55,6 +55,37 @@ function _check_matches() {
...
+	run_buildah --log-level debug push --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth exp docker://localhost:${REGISTRY_PORT}/test
```
mtrmac, line 64: Pedantically, the compression algorithm is not even promised to be deterministic: one compression to `dir:…` and another to `docker://` can create different compressed representations. So I think this still has the same problem, although in an even-rarer-to-hit variant.

tests/blobcache.bats
flouthoc: Yes this creates a compressed version.

flouthoc: This is not needed I can remove this.

tests/blobcache.bats
flouthoc: Can it be made more certain if i add `--compression gzip` explicitly ?

mtrmac: That doesn’t help:
- Drop BIC
- `buildah push foo docker://bar` (compressing)
- Drop BIC
- `buildah push foo docker://bar` (compressing again)

is _not_ required to create bit-identical representations. (Without the second BIC drop, we would reuse blobs and _probably_ use identical representations — depending on which compressed variant the BIC chooses.)

That holds regardless of algorithm: AFAIK, none of the promise to compress the same input to the same output, even in the same version (though I wasn’t looking for such a promise very hard).

flouthoc: Now moving image directly from repo to registry and then clearing BIC and performing push.

flouthoc:
@mtrmac PTAL again

mtrmac (commented):
The test operation LGTM.

(https://github.com/containers/buildah/pull/5153#discussion_r1389706426 is back but I’ll fully defer to repo maintainers on whether it is worth worrying about at all.)

flouthoc:
> The test operation LGTM.
> 
> ([#5153 (comment)](https://github.com/containers/buildah/pull/5153#discussion_r1389706426) is back but I’ll fully defer to repo maintainers on whether it is worth worrying about at all.)

@mtrmac Does latest commit resolves comment [#5153 (comment)](https://github.com/containers/buildah/pull/5153#discussion_r1389706426)

tests/blobcache.bats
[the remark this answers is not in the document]
flouthoc: @mtrmac Latest commit should address this, Could you PTAL again.

PR #5154, opened 2023-11-11
[CI:DOCS] Pass secrets from the host down to internal podman containers

rhatdan:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
Subscriptions from the host now pass through to internal containers on contrib buildah images.
```

### DIFF
--- a/contrib/buildahimage/Containerfile
+++ b/contrib/buildahimage/Containerfile
@@ -65,6 +65,9 @@ RUN dnf -y makecache && \
 
 ADD ./containers.conf /etc/containers/
 
+# Setup internal Buildah to pass secrets/subscriptions down from host to internal container
+RUN printf '/run/secrets/etc-pki-entitlement:/run/secrets/etc-pki-entitlement\n/run/secrets/rhsm:/run/secrets/rhsm\n' > /etc/containers/mounts.conf
+
 # Copy & modify the defaults to provide reference if runtime changes needed.
 # Changes here are required for running with fuse-overlay storage inside container.
 RUN sed -e 's|^#mount_program|mount_program|g' \


TomSweeneyRedHat:
The commit messages don't jive with the description.  Did you mean to glom all this together here?

rhatdan:
Nope

flouthoc (approved):
LGTM

flouthoc:
PR mentions this as `[CI:DOCS]` however this is changing the buildah image so `[CI:DOCS]` tag looks incorrect.

rhatdan:
I don't believe there are any tests that test the buildahimages so I treated it as a Docs update, to save the planet.

rhatdan:
@Luap99 PTAL

contrib/buildahimage/Containerfile
```diff
@@ -65,6 +65,9 @@ RUN dnf -y makecache && \
 
 ADD ./containers.conf /etc/containers/
 
+# Setup internal Buildah to pass secrets/subscriptions down from host to internal container
+RUN printf '/run/secrets/etc-pki-entitlement:/run/secrets/etc-pki-entitlement\n/run/secrets/rhsm:/run/secrets/rhsm\n' > /etc/containers/mounts.conf
```
edsantiago, line 69: Not a blocker, but is there any reason for doing it this way instead of creating a `mounts.conf` file in this subdirectory and adding it to the `ADD` on line 66?

TomSweeneyRedHat:
LGTM
I'll let @Luap99 push the merge button.

contrib/buildahimage/Containerfile
rhatdan: Nope, but not sure that would be much better.

Issue #5155, opened 2023-11-11
Can't remove default network sysctl


deliciouslytyped:
Unless I'm not reading the code at https://github.com/containers/buildah/blob/bf3b55ba74102cc2503eccbaeffe011728d46b20/run_linux.go#L776 correctly, there does not seem to be a way to disable the (default) buildah network sysctls despite claims to the contrary: https://github.com/containers/podman/issues/8482#issuecomment-733858789 (DefaultNetworkSysctl is defined at https://github.com/containers/buildah/blob/bf3b55ba74102cc2503eccbaeffe011728d46b20/define/types.go#L80)

I'm adding 
```
[containers]
default_sysctls = []
```
to /etc/containers/containers.conf but it appears to have no effect.

PR #5156, opened 2023-11-12
Stop using DefaultNetworkSysctl and use containers.conf only

rhatdan:
Also mark uses of DefaultCapabilities as Deprecated.

Fixes: https://github.com/containers/buildah/issues/5155

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
DefaultNetworkSysctl is not hard coded any longer and follows containers.conf
```

### DIFF
--- a/define/types.go
+++ b/define/types.go
@@ -58,8 +58,8 @@ const (
 type TeeType string
 
 var (
-	// DefaultCapabilities is the list of capabilities which we grant by
-	// default to containers which are running under UID 0.
+	// Deprecated: DefaultCapabilities values should be retrieved from
+	// github.com/containers/common/pkg/config
 	DefaultCapabilities = []string{
 		"CAP_AUDIT_WRITE",
 		"CAP_CHOWN",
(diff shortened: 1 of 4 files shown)


internal/mkcw/embed/entrypoint.gz
flouthoc, line 1: This change is not needed.

flouthoc (commented):
@Luap99 PTAL

define/types.go
```diff
@@ -75,8 +74,7 @@ var (
 		"CAP_SETUID",
 		"CAP_SYS_CHROOT",
 	}
-	// DefaultNetworkSysctl is the list of Kernel parameters which we
-	// grant by default to containers which are running under UID 0.
+	// DefaultNetworkSysctl is deprecated use values from containers.conf
```
Luap99, line 77: This should be written as `// Deprecated: ...`, see https://stackoverflow.com/questions/7849663/how-do-you-mark-code-as-deprecated-in-go

There is a linter which will then catch this and warn users about it same in the other places.

Luap99 (commented):
One nit plus comment from @flouthoc.
Overall this seems like the right thing to do but it looks different from what podman is doing, https://github.com/containers/podman/blob/8e6fe99fdb8035e64587af74ca2092a0ec89c50a/pkg/specgen/generate/security_linux.go#L223-L267
We should make sure they behave the same although I won't block this PR on it.

rhatdan:
@flouthoc @nalind @Luap99 @umohnani8 @giuseppe @vrothberg PTAL

Luap99 (approved):
LGTM

PR #5159, opened 2023-11-13
Make TEE types case-insensitive

nalind:
#### What type of PR is this?

/kind bug

Make sure that we accept both upper- and lower-cased versions of the names of TEE types as arguments for `buildah build`'s --cw flag and for `buildah mkcw`'s --type flag.  We previously only understood lower-case versions, but the docs always used upper case.

#### How to verify it

Extended integration tests!

None

```release-note
TEE types used with `buildah build --cw` and `buildah mkcw` are no longer required to be specified in lower-case.
```


flouthoc (commented):
LGTM

giuseppe (approved):
/lgtm

PR #5160, opened 2023-11-13
StageExecutor.Execute: force a commit for --unsetenv, too

nalind:
#### What type of PR is this?

/kind bug

When we don't have any instructions to process, but we do have an unsetEnv or unsetLabel list, force a commit, like we already did if we were passed an out-of-band labels or annotations list, or if we were squashing, and now also do so if we're doing a confidential workload.

#### How to verify it

Extended integration tests!

None

```release-note
`buildah build`ing a Containerfile with no instructions, with either `--unsetenv` or `--unsetlabel` specified on the command line, now correctly generates a new image.
```


flouthoc (approved):
LGTM

PR #5161, opened 2023-11-13
Never omit layers for emptyLayer instructions when squashing/cwing

nalind:
#### What type of PR is this?

/kind bug

Ignore whether or not the final build instruction should produce a layer if we're squashing or producing a confidential workload, when we'd still have to produce a layer containing the contents of the base image.

#### How to verify it

Extended an integration test!

None

```release-note
`buildah build` will no longer produce an empty image when the `--cw` flag is used along with the `--layers` flag and the final instruction in the Dockerfile is not one of ADD, COPY, or RUN.
```

### DIFF
--- a/image.go
+++ b/image.go
@@ -412,11 +412,6 @@ func (i *containerImageRef) NewImageSource(ctx context.Context, sc *types.System
 		if err != nil {
 			return nil, fmt.Errorf("unable to locate layer %q: %w", layerID, err)
 		}
-		// If we're up to the final layer, but we don't want to include
-		// a diff for it, we're done.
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
Although I want @flouthoc to approve, since this looks familiar to me as potentially something that fixed a different bug.

flouthoc (approved):
LGTM since this only affects `cw` images , afaik `buildah` allows producing empty layer images.

flouthoc (approved):
LGTM
/lgtm
/approve

TomSweeneyRedHat:
LGTM

PR #5165, opened 2023-11-15
conformance tests: use go-dockerclient for BuildKit builds

nalind:
#### What type of PR is this?
/kind failing-test 

go-dockerclient gained the ability to let us ask for a build kicked off using its API to be done using BuildKit, so we don't have to work around that by calling the Docker client package any more when doing conformance testing.

The go-dockerclient method also reports errors in a way that's easier for us to consume, which we didn't have fully debugged on the other code path.

#### How to verify it

Updated conformance tests!

None

```release-note
None
```


flouthoc (approved):
This works for locally thank you. LGTM
/lgtm
/approve
/hold

PR #5169, opened 2023-11-16
pkg/util.MirrorToTempFileIfPathIsDescriptor(): don't leak an fd

nalind:
#### What type of PR is this?

/kind bug

When `MirrorToTempFileIfPathIsDescriptor()` creates a file, it doesn't actually close the file handle.

#### How to verify it

Tests added in #3498 should keep passing.

None

```release-note
None
```

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -5,6 +5,8 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


flouthoc (approved):
LGTM
/lgtm
/approve
/hold

PR #5183, opened 2023-11-20
set makefile target internal/mkcw/embed/entrypoint.gz as .PHONY on non x86_64

dcermak:
The target internal/mkcw/embed/entrypoint is only built on x86_64, but internal/mkcw/embed/entrypoint.gz is run on all arches. This causes build failures on anything non x86_64 as internal/mkcw/embed/entrypoint is not build.

#### What type of PR is this?

/kind bug

To support running `make` on anything non x86_64.

#### How to verify it

Run `make` on a non-x86_64 machine.

None

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,17 +73,16 @@ bin/buildah: $(SOURCES) cmd/buildah/*.go internal/mkcw/embed/entrypoint.gz
 	$(GO_BUILD) $(BUILDAH_LDFLAGS) $(GO_GCFLAGS) "$(GOGCFLAGS)" -o $@ $(BUILDFLAGS) ./cmd/buildah
 
 ifneq ($(shell as --version | grep x86_64),)
+internal/mkcw/embed/entrypoint.gz: internal/mkcw/embed/entrypoint
+	$(RM) $@
+	gzip -k $^
+
 internal/mkcw/embed/entrypoint: internal/mkcw/embed/entrypoint.s
 	$(AS) -o $(patsubst %.s,%.o,$^) $^
 	$(LD) -o $@ $(patsubst %.s,%.o,$^)
 	strip $@
-else
-.PHONY: internal/mkcw/embed/entrypoint
 endif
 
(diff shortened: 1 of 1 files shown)


flouthoc:
@dcermak Could you please reduce the length of commit message.

flouthoc:
I think @giuseppe @nalind @mtrmac were working on a similar issue so it'll be best to tag them.

dcermak:
> @dcermak Could you please reduce the length of commit message.

done

rhatdan:
LGTM
But not sure why this is not needed to be build on non x86 arches?

Makefile
```diff
@@ -73,17 +73,19 @@ bin/buildah: $(SOURCES) cmd/buildah/*.go internal/mkcw/embed/entrypoint.gz
...
+	$(RM) $@
+	gzip -k $^
+
 internal/mkcw/embed/entrypoint: internal/mkcw/embed/entrypoint.s
 	$(AS) -o $(patsubst %.s,%.o,$^) $^
 	$(LD) -o $@ $(patsubst %.s,%.o,$^)
 	strip $@
 else
 .PHONY: internal/mkcw/embed/entrypoint
+.PHONY: internal/mkcw/embed/entrypoint.gz
```
mtrmac, line 86: I confirm this works, but, also, the GNU Make manual says

> A phony target should not be a prerequisite of a real target file; if it is, its recipe will be run every time `make` considers that file. 

AFAICS removing the `else` case entirely would work (or replacing it with a comment pointing at some validation?), and avoid the unnecessary rebuilds.

mtrmac (commented):
This is not really the topic of this PR — fixing the build is much more important: When we commit the built `.gz` into `git`, I’m wondering about the risk of the source and the binary (accidentally or intentionally) getting out of sync.

It Would Be Nice™️ if our CI builds always validated that the `.s` and `.gz`, as they are checked into `git`, match.

Makefile
dcermak: Yes that works too, I've amended the patch and removed both targets.

PR #5188, opened 2023-11-22
Bump to 1.33.2

TomSweeneyRedHat:
I'm doing some mad git hackery unlike anything I've done before.  I'm attempting to merge this commit between these two existing commits:

commit [12b205f69876bf09bb6dfe03adcdefad5d4ee9f8](https://github.com/containers/buildah/commit/12b205f69876bf09bb6dfe03adcdefad5d4ee9f8)
Remove makefile targets entrypoint{,.gz} for non x86_64

and 

commit [c36ab60ca472e182e3dcd0d51df3f383b108a665](https://github.com/containers/buildah/commit/c36ab60ca472e182e3dcd0d51df3f383b108a665)
 Merge pull request #5007 from containers/renovate/github.com-fsouza-go-dockerclient-1.x

That way, we will get just the changes added after 1.33.1 that deal with removing bloat and won't get the non-versioned c/{project} version bumps too.  I can then in theory make v1.33.2 with only the bloat fixes.

:crossed_fingers: 

[NO NEW TESTS NEEDED]

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 # Changelog
 
+## v1.33.2 (2023-11-22)
+
+    Update minimum to golang 1.20
+    fix(deps): update module github.com/fsouza/go-dockerclient to v1.10.0
+    fix(deps): update module github.com/moby/buildkit to v0.12.3
+    Bump to v1.33.2-dev
+
 ## v1.33.1 (2023-11-18)
 
     fix(deps): update module github.com/moby/buildkit to v0.11.4 [security]
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,9 @@
+- Changelog for v1.33.2 (2023-11-22)
+  * Update minimum to golang 1.20
+  * fix(deps): update module github.com/fsouza/go-dockerclient to v1.10.0
+  * fix(deps): update module github.com/moby/buildkit to v0.12.3
+  * Bump to v1.33.2-dev
+
 - Changelog for v1.33.1 (2023-11-18)
   * fix(deps): update module github.com/moby/buildkit to v0.11.4 [security]
   * test,heredoc: use fedora instead of docker.io/library/python:latest
--- a/define/types.go
+++ b/define/types.go
@@ -29,7 +29,7 @@ const (
 	// identify working containers.
 	Package = "buildah"
 	// Version for the Package. Also used by .packit.sh for Packit builds.
-	Version = "1.33.2-dev"
+	Version = "1.33.2"
 
 	// DefaultRuntime if containers.conf fails.
 	DefaultRuntime = "runc"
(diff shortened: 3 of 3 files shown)


TomSweeneyRedHat:
@mheon @ashley-cui @mtrmac PTAL  Especially the git hackery.  The branch says it needs updating, I don't think we should, but I'm not sure.

Here's my git hackery steps:

git checkout -b bump-1.33.2 c36ab60ca472e182e3dcd0d51df3f383b108a665
{change files}
git add {files}
git commit -m "Bump to v1.33.2" -s
{change version file, define/types.go]
git add {files}
git commit -m "Bump to v1.33.3-dev" -s
git rebase bump-1.33.2 main
git push -u TomSweeneyRedHat bump-1.33.2

Then in GitHub, opened the PR after reviewing.

TomSweeneyRedHat:
@vrothberg if you could eyeball the hackery, I'd appreciate it.

flouthoc:
@TomSweeneyRedHat I think it will be more easier if we create `release-1.33` branch from https://github.com/containers/buildah/releases/tag/v1.33.1 and then backport needed commits there and cut `release-1.33.2` from that branch. Wouldn't it ? And we can bump `main` to `1.34-dev` or something. WDYT ?

Luap99:
+1 to @flouthoc comment, this should really be done by using a branch and backporting what you need. If we need to to another release this approach here would no longer work.

mtrmac:
> git rebase bump-1.33.2 main

(I don’t know that this did anything… I read that as updating the `main` branch on top of `bump-1.33.2`; that wouldn’t show up in this PR.)

---

If I understand this correctly, the goal is to create something like this?

```
  /-.-.-.-.-\
-.           .-main
  \-.-x-----/
```
where `x` is the tagged release?

Yes, that would work (and it probably _is_ extensible to doing it again and tagging more 1.33.x releases in a similar way), but AFAICS creating a branch and not merging it into `main` would be simpler in all respects.

E.g. if we have a branch, we can _choose_ to run dedicated CI to ensure that branch continues to build.

(There is an argument for keeping release branches _and always merging them into `main`, per https://devblogs.microsoft.com/oldnewthing/20180323-01/?p=98325, but we don’t currently work that way and I am not proposing we start. And that would still work better with a named branch.)

TomSweeneyRedHat:
@mtrmac @Luap99 @flouthoc thanks for the comments and suggestions.  Your example with x above Mirek is what I was hoping to do.  

I was hoping to avoid creating yet another branch and just pull a fast one, but often times fast ones aren't the best.  I'll go with the consensus and will whip up a branch.   Thanks all.

If nothing else, I enjoyed the git gymnastics of this one, I learned a bit in the process.

PR #5197, opened 2023-11-29
Ignore errors if label.Relabel returns ENOSUP

rhatdan:
This is a common mistake by users and is ignored in some places but not everywhere. This change will help this to be ignored everwhere.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
SELinux relabeling via :Z and :z on file systems which do not support labels like NFS are ignored.
```

[NO NEW TESTS NEEDED]


rhatdan:
@n1hility PTAL

n1hility:
Awesome! LGTM other than the conflict

nalind:
Most of the places where this PR suppresses that error are for files we're creating, where the user doesn't get to tell us whether or not to change the labels on them.  The exceptions are the ones in `Builder.runSetupVolumeMounts()`, and I'm not sure we could hit this problem for the rest.

We're pretty inconsistent about where we put files we bind-mount in.  An environment secret gets stored under `define.TempDir`, which looks like an unexplained inconsistency since a file secret is stored under the per-container directory, and SSH sockets and `/run/.containerenv` are placed under `internal/tmpdir.GetTempDir()`.

rhatdan:
I just wanted to be consistent.  No reason to not call the same code everwhere.  For most of these there will be no change as you point out. but if we spray around label.Relabel and relabel code, there is a decent chance a future contributor could do it wrong.

PR #5224, opened 2023-12-12
mkcw: populate the rootfs using an overlay

nalind:
#### What type of PR is this?

/kind design

When using the working container's rootfs to populate a plaintext disk image with mkfs, instead of writing .krun_config.json to the rootfs and then removing it afterward (since we don't want it to show up if the same working container is later committed to non confidential-workload image), mount an overlay filesystem using a temporary directory as the upper and the rootfs as the lower, create the .krun_config.json file in the overlay filesystem, and use the overlay filesystem as the source directory for mkfs.

#### How to verify it

Updated a test to make sure we don't break when base images have multiple layers.

Modifying the roots during `commit` is inherently racy.

```release-note
None
```

### DIFF
--- a/convertcw.go
+++ b/convertcw.go
@@ -171,6 +171,7 @@ func CWConvertImage(ctx context.Context, systemContext *types.SystemContext, sto
 		Slop:                     options.Slop,
 		FirmwareLibrary:          options.FirmwareLibrary,
 		Logger:                   logger,
+		GraphOptions:             store.GraphOptions(),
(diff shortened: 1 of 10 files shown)


rhatdan:
LGTM
@giuseppe @saschagrunert @mtrmac @umohnani8 PTAL

TomSweeneyRedHat:
LGTM, but definitely want other eyeballs on this before merge.

PR #5226, opened 2023-12-13
commit: add a --add-file flag

nalind:
#### What type of PR is this?

/kind feature

Add a flag to `buildah commit` which allows adding arbitrary files to the image while we're committing it.  When not squashing, they'll take the form of a second new layer.

#### How to verify it

New integration test!

None

The bit that warns about not being able to do this in combination with `--cw` should be easy to fix once #5224 or something like it is merged.

```release-note
`buildah commit` now accepts a `--add-file` flag, which can be used to add a file to the committed image without modifying the working container which is being committed.
```

### DIFF
--- a/docs/buildah-commit.1.md
+++ b/docs/buildah-commit.1.md
@@ -19,6 +19,14 @@ The image ID of the image that was created.  On error, 1 is returned and errno i
 
 ## OPTIONS
 
+**--add-file** *source[:destination]*
+
+Read the contents of the file `source` and add it to the committed image as a
+file at `destination`.  If `destination` is not specified, the path of `source`
+will be used.  The new file will be owned by UID 0, GID 0, have 0644
+permissions, and be given a current timestamp unless the **--timestamp** option
+is also specified.  This option can be specified multiple times.
+
 **--authfile** *path*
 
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM
Who would do this and Why? Is this a feature you want to add to buildah build?

nalind:
It's something we're going to use for https://issues.redhat.com/browse/RUN-1838.  It's really only exposed at the CLI so that we can have an integration test as part of this PR.  For `build` I'd still point people at ADD and COPY instructions, since this only handles one file at a time.

docs/buildah-commit.1.md
```diff
@@ -19,6 +19,12 @@ The image ID of the image that was created.  On error, 1 is returned and errno i
 
 ## OPTIONS
 
+**--add-file** *source:destination*
+
+Read the contents of the file `source` and add it to the committed image as a file at `destination`.  The new file will
+be owned by UID 0, GID 0, have 0644 permissions, and be given a current timestamp unless the **--timestamp** option is
+also specified.  This option can be specified multiple times.
+
```
TomSweeneyRedHat, line 27: Not necessary for this PR, but did you consider making the destination optional?  I.e., if not specified, it would default to the source location?

nalind: It hadn't occurred to me, but there's no reason it couldn't be optional.

TomSweeneyRedHat:
LGTM

Issue #4740, opened 2023-04-14
podman build --output to an oci-archive isn't easy to figure out


dustymabe:
**Description**

I was looking for an easy way to build and output to an oci-archive in a single step. So rather than `podman build -t ` + `podman push foo oci-archive:./foo.ociarchive`, I need to just do that in one step.

I discovered the `--output` option, which in the man page describes:

```
the value of --output can be just a destination (in the **dest** format)"
```

So I try:

```
podman build -f Containerfile --output=oci-archive:./foo.ociarchive
```

and that fails with an obscure error (I'm in a locked down environment):

```
+ podman build /srv/tmp/tmpw2r2ieuu --output oci-archive:/srv/builds/37.20230414.dev.0/x86_64/fedora-coreos-37.20230414.dev.0-kubevirt.x86_64.ociarchive     
STEP 1/2: FROM scratch                                                                                                                                       
STEP 2/2: ADD fedora-coreos-37.20230414.dev.0-kubevirt.x86_64.qcow2 /disks/                                                                                                                                                                                                                                                
COMMIT                                                                                                                                                       
Error: failed to export build output: failed while performing untar at "/srv/oci-archive:/srv/builds/37.20230414.dev.0/x86_64/fedora-coreos-37.20230414.dev.0-kubevirt.x86_64.ociarchive": processing tar file(lchown /disks: operation not permitted): exit status 1
```

What's important here is not the failure, but the fact that podman isn't doing what I'm asking for.


I also tried with `--output type=oci,dest=./foo.ociarchive` and that didn't work either.

Then @nalind told me that `--tag` supports destination format (undocumented) and that works:


```
+ podman build /srv/tmp/tmp5zrcblve --tag oci-archive:/srv/builds/37.20230414.dev.0/x86_64/fedora-coreos-37.20230414.dev.0-kubevirt.x86_64.ociarchive
STEP 1/2: FROM scratch
STEP 2/2: ADD fedora-coreos-37.20230414.dev.0-kubevirt.x86_64.qcow2 /disks/
COMMIT oci-archive:/srv/builds/37.20230414.dev.0/x86_64/fedora-coreos-37.20230414.dev.0-kubevirt.x86_64.ociarchive
--> 

+ exit 0
```


I think there's a few places where there is room for improvement here:

1. Document that --tag supports destination format and not just tags. Is there a good reason not to?
2. Add support for --output to output to destination format locations. Is there a good reason not to?
3. Document the limitations of `--output`.


[...]

PR #5230, opened 2023-12-17
[CI:DOCS] Document use of containers-transports values in buildah

rhatdan:
Fixes: https://github.com/containers/buildah/issues/4740

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```

### DIFF
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -708,6 +708,8 @@ Valid _type_ values are:
 If no type is specified, the value defaults to **local**.
 Alternatively, instead of a comma-separated sequence, the value of **--output** can be just a destination (in the `**dest** format) (e.g. `--output some-path`, `--output -`) where `--output some-path` is treated as if **type=local** and `--output -` is treated as if **type=tar**.
 
+Note: The **--tag** option can also be used to change the file image format to supported `containers-transports(5)`.
+
 **--pid** *how*
 
 Sets the configuration for PID namespaces when handling `RUN` instructions.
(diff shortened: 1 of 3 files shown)


rhatdan:
@dustymabe PTAL
@nalind @vrothberg @flouthoc @TomSweeneyRedHat @umohnani8 PTAL

docs/buildah-build.1.md
```diff
@@ -872,6 +874,8 @@ Specifies the name which will be assigned to the resulting image if the build
 process completes successfully.
 If _imageName_ does not include a registry name component, the registry name *localhost* will be prepended to the image name.
 
+The **--tag** option supports all transports from `containers-transports(5)` If no transport is specified, the `container-storage` (i.e., local storage) transport is used.
```
vrothberg, line 877:
```suggestion
The **--tag** option supports all transports from `containers-transports(5)`. If no transport is specified, the `container-storage` (i.e., local storage) transport is used.
```

Same for others.

vrothberg (commented):
The referenced issue talks about `--output` but here we mention `--tag`.  I am a bit confused.

docs/buildah-build.1.md
```diff
@@ -708,6 +708,8 @@ Valid _type_ values are:
 If no type is specified, the value defaults to **local**.
 Alternatively, instead of a comma-separated sequence, the value of **--output** can be just a destination (in the `**dest** format) (e.g. `--output some-path`, `--output -`) where `--output some-path` is treated as if **type=local** and `--output -` is treated as if **type=tar**.
 
+Note: The **--tag** option can also be used to change the file image format to supported `containers-transports(5)`.
```
dustymabe, line 711: I think it might be worth adding an example here to make it more clear. For example, `podman build --tag=oci-archive:./foo.ociarchive .` will build a container and store it in oci-archive format in the file `./foo.ociarchive`.

I know this might make the docs a little more ugly, but might be worth the tradeoff.

dustymabe:
I'm kind of wondering why we need `--output` at all. Maybe we should just make it alias to `--tag` with the right containers-transforts formatted srting?

flouthoc:
> I'm kind of wondering why we need `--output` at all. Maybe we should just make it alias to `--tag` with the right containers-transforts formatted srting?

@dustymabe afaik `--output` was added for buildkit compatiblity and it allows some things like extracting entire rootfs as well. 


Current PR LGTM

umohnani8:
changes LGTM

dustymabe (approved):

TomSweeneyRedHat:
LGTM

PR #5239, opened 2023-12-30
Replace strings.SplitN with strings.Cut

serprex:
#### What type of PR is this?
/kind cleanup

`strings.Cut` is a cleaner & more performant api relative to `strings.SplitN(_, _, 2)` added in go 1.18

#### How to verify it
No change in behavior

None

Introduced in go 1.18: https://github.com/golang/go/issues/46336

```release-note
None
```

### DIFF
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -337,13 +337,13 @@ func newExecutor(logger *logrus.Logger, logPrefix string, store storage.Store, o
 					// We have to be careful here - it's either an argument
 					// and value, or just an argument, since they can be
 					// separated by either "=" or whitespace.
-					list := strings.SplitN(arg.Value, "=", 2)
+					argName, argValue, hasValue := strings.Cut(arg.Value, "=")
 					if !foundFirstStage {
-						if len(list) > 1 {
-							globalArgs[list[0]] = list[1]
+						if hasValue {
+							globalArgs[argName] = argValue
 						}
 					}
-					delete(exec.unusedArgs, list[0])
+					delete(exec.unusedArgs, argName)
 				}
 			case "FROM":
 				foundFirstStage = true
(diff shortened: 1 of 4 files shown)


imagebuildah/executor.go
```diff
@@ -845,23 +838,18 @@ func (b *Executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 							fields := strings.Split(mountFlags, ",")
 							for _, field := range fields {
 								if strings.HasPrefix(field, "from=") {
-									fromField := strings.SplitN(field, "=", 2)
-									if len(fromField) > 1 {
```
serprex, line 849: I removed this condition for 2 reasons:
1. it can never happen, since we know `=` appears in `field`
2. the error message in the dead else path would log a nil as `formField[1]` would be nil

internal/volumes/volumes.go
```diff
@@ -124,23 +124,20 @@ func GetBindMount(ctx *types.SystemContext, args []string, contextDir string, st
 				return newMount, "", fmt.Errorf("cannot pass 'relabel' option more than once: %w", errBadOptionArg)
 			}
 			setRelabel = true
-			if len(kv) != 2 {
```
serprex, line 127: I removed this check because the exact same error will be raised by the `default` case

rhatdan:
@nalind @vrothberg @giuseppe @umohnani8 PTAL

vrothberg (approved):
Thanks!  The changes LGTM.  In order to pass CI, the commit message needs magic string `[NO NEW TESTS NEEDED]`.

The CI enforces tests which in this case aren't needed.

rhatdan:
/lgtm
Thanks @serprex 
Going to do something similar in Podman?

serprex:
Sure, I'll take a look

serprex:
rebased to clean up commit messages since commit subject was too long

umohnani8:
LGTM
/hold cancel

PR #5245, opened 2024-01-03
[release-1.24] bump golang.org/x/crypto to v0.17.0

lsm5:
Manual bump because of failed automated cherrypick of #5232.
    
Go bumped to 1.17 otherwise it fails to build.
    
Fixes: GHSA-45x7-px36-x8w8 CVE-2023-48795

#### What type of PR is this?

/kind other

Fixes: https://github.com/advisories/GHSA-45x7-px36-x8w8 https://github.com/advisories/GHSA-45x7-px36-x8w8

#### How to verify it

Don't know if there's a manual reproducer and if buildah is even directly affected. CVE page advises bumping version.

Fixes: https://github.com/advisories/GHSA-45x7-px36-x8w8 https://github.com/advisories/GHSA-45x7-px36-x8w8

None

```release-note
Fixed [GHSA-45x7-px36-x8w8](https://github.com/advisories/GHSA-45x7-px36-x8w8): CVE-2023-48795 by vendoring golang.org/x/crypto v0.17.0.
```


lsm5:
cc @jnovy

lsm5:
Are we ok with bumping golang to 1.17 on release-1.24 branch? I guess we'll need a CI vm update for the vendoring task. It passes on release-1.29 but fails here.

lsm5:
@cevich The smoke test task uses fedora 35 and golang 1.16. And we need min golang 1.17 for this test to pass. Do I bump the image used to a higher Fedora or do I somehow try to bump just the golang package version?

rhatdan:
Yes bump the images.

lsm5:
This will probably need a c/storage bump as the current state fails to build (see the Integration fedora-* failures). 

I tried vendoring c/storage v1.45.3 (used in buildah release-1.29) to this branch and it built fine. Would `1.38.5 -> 1.45.3` upgrade of c/storage be ok, or should I look for the smallest possible delta ?

TomSweeneyRedHat:
There's a boatload of changes in here for 8.6 which is wrapping up EUS support sometime this summer.  I'd rather get the build issues resolved in 1.38.6 if it's not too onerous.

Do we have a Jira card for this somewhere?  We'll need to spin one before mering if one doesn't exist.

lsm5:
> It's unprecedented (near as I can remember) for us to ever bump CI VM images on a release branch. I think this could have unintended consequences - since the newer fedoras didn't exist at the time the release was produced. Though I would need to spend time thinking a lot (and right now my brain hurts).

Given the maintenance branches are almost always used for rhel, we should get to a point where release branches are tested only on rhel/centos stream and not Fedora. That'll solve the EOL problem.

> 
> The other thing I worry about is this one update + golang bump, quickly snowballing into a crap ton of other necessary updates (even "new" lint fixes). I'm not saying "no, definitely don't do it". Rather, we need to think very carefully if this is the right way to solve the right problem.
> 
> I'm assuming we're vendoring the impacted package already eh? Would it be a major PITA to backport the actual fixes to our vendored copy, such that it works with the "old" golang?

Given it's a crypto package, I'd rather do whatever it takes to use the official copy. Even if we could backport, I guess that brings in tons of other PITA, hosting the backported patch somewhere in rpm / buildah upstream OR forking the original library with additional manipulations to go.mod.

cevich:
I agree, moving to RHEL/CentOS for release branches would be ideal.  The idea of keeping the older CI VMs around, was they're a point-in-time snapshot that's fairly close to RHEL/CentOS.  As soon as we update them to a newer Fedora, we're no-longer testing what users are using (in RHEL/CentOS).  So that then begs the question: why test it at all?

Taking a step back...do RHEL/CentOS also need a golang 1.17 bump to build w/ this CVE fix?  Assuming they did a backport, could we just copy that work here?

rhatdan:
RHEL and Centos update the golang.

lsm5:
> Taking a step back...do RHEL/CentOS also need a golang 1.17 bump to build w/ this CVE fix? Assuming they did a backport, could we just copy that work here?

FWIW, ubi8 and ubi9 images now install golang 1.20.10.

cevich:
Hmmm, that's all useful to know.  Would removing and updating just golang be an option here?  You know I hate runtime packaging, but that might be a less-worse option than replacing the entire distro.

rhatdan:
That is what I would prefer.

cevich:
@lsm5 are you comfortable with trying this in `setup_environment.sh` (for all tests) or would you like me to help work it out?

lsm5:
> @lsm5 are you comfortable with trying this in `setup_environment.sh` (for all tests) or would you like me to help work it out?

@cevich help would be appreciated :)

cevich:
Okay sure thing, I'll move this up my priority list then.

cevich:
@lsm5 @rhatdan Q: Do we care about testing with BOTH F34 & F35 on this release branch?

Reason I ask is, it'll be easy for me to disable one and just focus on getting golang 1.17 onto the other (F35 I'm assuming).

Please LMK if you have different opinions.

cevich:
okay, hopefully that should do the trick.  I developed and ran the tests using a `hack/get_ci_vm.sh` machine and they all passed. @lsm5 @rhatdan PTAL

cevich:
aww :hankey: seems I missed a spot...

cevich:
...once again for the second time.

cevich:
:blush:

contrib/cirrus/setup.sh
```diff
@@ -25,6 +25,23 @@ EOF
...
+        cd /usr/local
+        showrun curl -Ssl -O https://dl.google.com/go/go1.17.13.linux-amd64.tar.gz
+        showrun tar -xf go1.17.13.linux-amd64.tar.gz
+        cd -
+        # This needs to be set in every test-execution context. Hack it in here
+        # since lib.sh always loads this before running any tests.
+        echo 'export PATH=$PATH:/usr/local/go/bin' | tee -a /etc/automation_environment
+        export PATH=$PATH:/usr/local/go/bin
+        showrun go version
+        showrun go env
```
cevich, line 44: Note: For whatever reason, the `showrun` output doesn't appear in the logs.  I have no idea why.  It's definitely running the above commands.  I suspect stdio file-descriptor shenanigans somewhere.

cevich:
> don't think we care about Fedora at all. So, whatever you find convenient.

We care insofar as it represents a time-snapshot close to what appears in CentOS and RHEL.  So we probably don't need to test both F34 and F35.  I'll strip out the lesser one in the hopes that it saves somebody some amount of future-headaches.

cevich:
Force-push: Drop F34 completely.

This should be ready to go (assuming CI passes).

TomSweeneyRedHat:
LGTM
and happy green test buttons!
Thanks for keeping at this @cevich and @lsm5 !

cevich:
This was a tricky one, thanks @lsm5 for collaborating on a long-term fix.

PR #5252, opened 2024-01-08
internal/mkcw.Archive(): handle extra image content

nalind:
#### What type of PR is this?

/kind feature

Add the `--add-file` to the `mkcw` command, and handle it when `commit` is given the `--cw` option:  when we have extra files to add to the image, handle them by adding them to the upper overlay layer before creating the plaintext filesystem image.

Also touch up a couple of man pages.

#### How to verify it

Extended integration test!

None

```release-note
`buildah mkcw` now accepts a `--add-file` flag, which can be used to add a file to the committed image without modifying the working container which is being committed.
```

### DIFF
--- a/cmd/buildah/mkcw.go
+++ b/cmd/buildah/mkcw.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/containers/buildah"
 	"github.com/containers/buildah/pkg/parse"
@@ -38,6 +39,7 @@ func mkcwCmd(c *cobra.Command, args []string, options buildah.CWConvertImageOpti
 
 func init() {
 	var teeType string
+	var addFile []string
 	var options buildah.CWConvertImageOptions
 	mkcwDescription := `Convert a conventional image to a confidential workload image.`
 	mkcwCommand := &cobra.Command{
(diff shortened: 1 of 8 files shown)


internal/mkcw/archive.go
```diff
@@ -70,8 +71,6 @@ func (c chainRetrievalError) Error() string {
 
 // Archive generates a WorkloadConfig for a specified directory and produces a
 // tar archive of a container image's rootfs with the expected contents.
-// The input directory will have a ".krun_config.json" file added to it while
-// this function is running, but it will be removed on completion.
```
nalind, line 74: This comment was no longer correct as of #5224, but I forgot to drop it during that one.

internal/mkcw/archive.go
```diff
@@ -245,6 +244,38 @@ func Archive(path string, ociConfig *v1.Image, options ArchiveOptions) (io.ReadC
...
+				return fmt.Errorf("preparing to write %q to container root filesystem: %w", location, err)
+			}
+			defer output.Close()
+			input, err := os.Open(content)
+			if err != nil {
+				return err
+			}
+			defer input.Close()
+			if _, err := io.Copy(output, input); err != nil {
+				return fmt.Errorf("copying contents of %q to %q in root filesystem: %w", location, content, err)
```
flouthoc, line 264: Above error stmt is using `container root filesystem` instead of only `root filesystem`, maybe we can just keep any one of them for consistency.

```suggestion
				return fmt.Errorf("copying contents of %q to %q in container root filesystem: %w", location, content, err)
```

flouthoc (approved):
LGTM one comment above.

internal/mkcw/archive.go
nalind: Changing it.  Also, I had the orders of `location` and `content` mixed up, switching them.

cmd/buildah/mkcw.go
```diff
@@ -46,6 +48,26 @@ func init() {
 		Long:  mkcwDescription,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			options.TeeType = parse.TeeType(teeType)
+			if len(addFile) > 0 {
+				options.ExtraImageContent = make(map[string]string)
+				for _, spec := range addFile {
+					specSlice := strings.SplitN(spec, ":", 2)
```
rhatdan, line 54: Are we supposed to use strings.Cut now?

rhatdan:
LGTM
But one question.

PR #4829, opened 2023-05-30
Fix transition test to work with latest selinux policy

rhatdan:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### Does this PR introduce a user-facing change?

```release-note
None
```

[CI:NEXT]

PR #5260, opened 2024-01-10
[release-1.24] conformance test: ignore file type bits when comparing layers

cevich:
#### What type of PR is this?

/kind failing-test 

When comparing layer payloads during conformance tests, mask off any file type bits that the tar headers in the layers might have included.  Extracted from #4829.

#### How to verify it

Conformance tests!

```release-note
None
```

### DIFF
--- a/tests/conformance/conformance_test.go
+++ b/tests/conformance/conformance_test.go
@@ -788,7 +788,7 @@ func fsHeaderForEntry(hdr *tar.Header) FSHeader {
 		Name:     hdr.Name,
 		Linkname: hdr.Linkname,
 		Size:     hdr.Size,
-		Mode:     hdr.Mode,
+		Mode:     (hdr.Mode & int64(os.ModePerm)),
 		UID:      hdr.Uid,
 		GID:      hdr.Gid,
 		ModTime:  hdr.ModTime,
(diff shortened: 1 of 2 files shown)


tests/conformance/conformance_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"io/fs"
```
nalind, line 12: This package wasn't added until 1.16, so this line can be dropped.

tests/conformance/conformance_test.go
```diff
@@ -788,7 +789,7 @@ func fsHeaderForEntry(hdr *tar.Header) FSHeader {
 		Name:     hdr.Name,
 		Linkname: hdr.Linkname,
 		Size:     hdr.Size,
-		Mode:     hdr.Mode,
+		Mode:     (hdr.Mode & int64(fs.ModePerm)),
```
nalind, line 792:
```suggestion
		Mode:     (hdr.Mode & int64(os.ModePerm)),
```
In newer releases, `os.ModePerm` is an alias for `fs.ModePerm`, but this branch used Go 1.13.

cevich: I never would have guessed that.  Thanks!

PR #5275, opened 2024-01-17
Make buildah match podman for handling of ulimits

rhatdan:
Podman currently sets the ulimits of nofile and nproc to max in rootless mode, if the user does not override.

Buildah on the other hand just passes in the current defaults.

Podman build should match podman run, and this will fix that problem.

Fixes: https://github.com/containers/buildah/issues/5273

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
Buildah now sets the maximum ulimits nofile and nproc by default.
```

### DIFF
--- a/define/types.go
+++ b/define/types.go
@@ -54,6 +54,9 @@ const (
 	SNP TeeType = "snp"
 )
 
+// DefaultRlimitValue is the value set by default for nofile and nproc
+const RLimitDefaultValue = uint64(1048576)
+
 // TeeType is a supported trusted execution environment type.
 type TeeType string
 
--- a/run_linux.go
+++ b/run_linux.go
@@ -854,6 +854,9 @@ func addRlimits(ulimit []string, g *generate.Generator, defaultUlimits []string)
 	var (
 		ul  *units.Ulimit
 		err error
+		// setup rlimits
+		nofileSet bool
+		nprocSet  bool
 	)
 
 	ulimit = append(defaultUlimits, ulimit...)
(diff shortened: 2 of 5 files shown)


define/types.go
```diff
@@ -54,6 +54,9 @@ const (
 	SNP TeeType = "snp"
 )
 
+// DefaultRlimitValue is the value set by default for nofile and nproc
+const RLimitDefaultValue = uint64(1048576)
```
nalind, line 58: Where do these values come from?  Are they always the default everywhere?

rhatdan: This is the default from Podman, no idea where it came from.

rhatdan: @giuseppe  added it

rhatdan: Might have been a value from Docker?
https://bbs.archlinux.org/viewtopic.php?id=285058

rhatdan: As of 7 years ago was a kernel max.
https://serverfault.com/questions/819119/xtrabackup-too-many-open-files-more-than-1048576

Might want to use something even higher.

cat /proc/sys/kernel/pid_max 
4194304

rhatdan: Under Moby I see.
./libnetwork/docs/vagrant-systemd/docker.service:LimitNOFILE=1048576
./libnetwork/docs/vagrant-systemd/docker.service:LimitNPROC=1048576

rhatdan: The question is do we continue to roll with it, or should we increase it?

TomSweeneyRedHat: I'd vote to keep Podman and Buildah in lockstep.  I don't know if there's a place to tuck the default values so both could reference them from the same place, that would be nice.  I think 1048576 for the kernel max is good for now.

tests/bud/bud.limits/Containerfile~
nalind, line 1: Is this supposed to be here?

nalind:
I must be missing something.  In the test, the build is printing the _soft_ limits it's seeing ([here](https://github.com/containers/buildah/pull/5275/files#diff-ad273f65eb2058b63e03de62d36caf17f3cd4e98afb0164563097a902fadf7d2R3-R4)), but the test is expecting those values to match the _hard_ limits (gathered [here](https://github.com/containers/buildah/pull/5275/files#diff-2265e8a639c28e2955dc6be61ba1b3abd53972d3cd5a54d5f80d61ba3a6306fcR523) and [here](https://github.com/containers/buildah/pull/5275/files#diff-2265e8a639c28e2955dc6be61ba1b3abd53972d3cd5a54d5f80d61ba3a6306fcR528)) it has.
The code, however, is passing along the current-set limits, both hard and soft.
Why does the test pass?

tests/bud/bud.limits/Containerfile~
rhatdan: Said sarcastically, no believe it is not supposed to be there.

rhatdan:
@nalind we are only printing the hard limit and testing off of it.

nalind:
> @nalind we are only printing the hard limit and testing off of it.

I think you're mistaken about which field the RUN commands in `tests/bud/bud.limits/Containerfile` are printing.

rhatdan:
@nalind this is ready to go in, I believe.

rhatdan:
@giuseppe @flouthoc @mheon @Luap99 @ashley-cui @umohnani8 @nalind @vrothberg PTAL

giuseppe (approved):
LGTM

define/types.go
vrothberg: Can we move this constant to c/common?  No need to block this PR on it but we should use every occasion to avoid future code divergence; this PR smells like a good example :)

vrothberg (approved):
/lgtm

PR #5276, opened 2024-01-17
commit: force omitHistory if the parent has layers but no history

nalind:
#### What type of PR is this?

/kind bug

If the parent image has layers but no history, force our own omitHistory setting on.

The alternative is to create a history that only explains the presence of some of the layers in our output image, which looks broken to everyone who might consume that image, including ourselves if we try to use it as a base image later.

#### How to verify it

New integration test!

Fixes https://github.com/containers/podman/issues/21094.

```release-note
None
```

### DIFF
--- a/image.go
+++ b/image.go
@@ -1048,10 +1048,20 @@ func (b *Builder) makeContainerImageRef(options CommitOptions) (*containerImageR
 	}
 
 	parent := ""
+	forceOmitHistory := false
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM
@flouthoc @giuseppe @mtrmac @vrothberg PTAL

flouthoc (approved):
LGTM
/lgtm
/approve

PR #5279, opened 2024-01-18
build,commit: add --sbom to scan and produce SBOMs when committing

nalind:
#### What type of PR is this?

/kind feature

Add a --sbom flag to `buildah build` and `buildah commit` which will scan the rootfs and specified context directories to build SPDX or CycloneDX SBOMs and lists of package URLs.

#### How to verify it

New integration tests!
Some more unit tests!

None

```release-note
The `buildah commit` and `buildah build` commands now recognize `--sbom` and related flags, which can be used to scan the image which is being created.  The scanner's output can be added to the image or saved to local disk.
```

### DIFF
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -42,6 +42,17 @@ type commitInputOptions struct {
 	quiet              bool
 	referenceTime      string
 	rm                 bool
+	pull               string
+	pullAlways         bool
+	pullNever          bool
+	sbomImgOutput      string
+	sbomImgPurlOutput  string
+	sbomMergeStrategy  string
+	sbomOutput         string
+	sbomPreset         string
+	sbomPurlOutput     string
+	sbomScannerCommand []string
+	sbomScannerImage   string
 	signaturePolicy    string
 	signBy             string
 	squash             bool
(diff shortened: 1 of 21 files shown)


flouthoc (changes requested):
Not a detailed review yet but i have a small question.

* Is it mandatory to merge the behavior of `--pull` and `--sbom`, i mean if there is a use-case when user wants to pull new scanner images but keep using his older base images he cannot do that as of now. Will it be a good idea to have `--sbom-scanner-pull` or just `--sbom-pull` instead ?


* Documentation: `--pull` in `buildah-commit` docs talks about `sbom` while `--pull` in `buildah-build` says nothing about `sbom` .

nalind:
> Not a detailed review yet but i have a small question.
> 
>     * Is it mandatory to merge the behavior of `--pull` and `--sbom`, i mean if there is a use-case when user wants to pull new scanner images but keep using his older base images he cannot do that as of now. Will it be a good idea to have `--sbom-scanner-pull` or just `--sbom-pull` instead ?

Maybe?  I don't know that there's a compelling case for having them be differently controllable, but we're already adding _eight_ flags to `build`, which had more flags than I'd like even before this PR.
 
>     * Documentation: `--pull` in `buildah-commit` docs talks about `sbom` while `--pull` in `buildah-build` says nothing about `sbom` .

Modified, hopefully acceptably.

docs/buildah-build.1.md
```diff
@@ -803,6 +806,95 @@ consult the manpages of the selected container runtime.
 Note: Do not pass the leading `--` to the flag. To pass the runc flag `--log-format json`
 to buildah build, the option given would be `--runtime-flag log-format=json`.
 
+**--sbom** *preset*
+
+Generate SBOMs (software bills of materials) for the output image by scanning
```
TomSweeneyRedHat, line 811:
```suggestion
Generate SBOMs (Software Bill of Materials) for the output image by scanning
```

docs/buildah-build.1.md
```diff
@@ -803,6 +806,95 @@ consult the manpages of the selected container runtime.
...
+     were specified.
+
+**--sbom-output** *file*
+
+When generating SBOMs, store the generated SBOM in the named file on the local
+filesystem.  There is no default.
+
+**--sbom-purl-output** *file*
+
+When generating SBOMs, scan them for PURL (package URL) information, and save a
```
TomSweeneyRedHat, line 878: A very soft suggestion for consideration
```suggestion
When generating SBOMs, scan them for PURL ([package URL](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst)) information, and save a
```

docs/buildah-commit.1.md
```diff
@@ -174,6 +200,95 @@ When writing the output image, suppress progress output.
 Remove the working container and its contents after creating the image.
 Default leaves the container and its content in place.
 
+**--sbom** *preset*
+
+Generate SBOMs (software bills of materials) for the output image by scanning
```
TomSweeneyRedHat, line 205: Prior comments in here too if you take them from above.

TomSweeneyRedHat:
Is this big enough to warrant a Buildah v2.0 to go out with Podman v5.0?  I don't think so, but there may be a case for it.

cmd/buildah/commit.go
```diff
@@ -112,6 +123,36 @@ func commitListFlagSet(cmd *cobra.Command, opts *commitInputOptions) {
...
+	_ = cmd.RegisterFlagCompletionFunc("sbom-merge-strategy", completion.AutocompleteNone)
+	flags.StringVar(&opts.sbomOutput, "sbom-output", "", "save scan results to `file`")
+	_ = cmd.RegisterFlagCompletionFunc("sbom-output", completion.AutocompleteDefault)
+	flags.StringVar(&opts.sbomImgOutput, "sbom-image-output", "", "add scan results to image as `path`")
+	_ = cmd.RegisterFlagCompletionFunc("sbom-image-output", completion.AutocompleteNone)
+	flags.StringVar(&opts.sbomPurlOutput, "sbom-purl-output", "", "save scan results to `file``")
+	_ = cmd.RegisterFlagCompletionFunc("sbom-purl-output", completion.AutocompleteDefault)
+	flags.StringVar(&opts.sbomImgPurlOutput, "sbom-image-purl-output", "", "add scan results to image as `path`")
+	_ = cmd.RegisterFlagCompletionFunc("sbom-image-purl-output", completion.AutocompleteNone)
+
```
TomSweeneyRedHat, line 155: Does commit really pull the container image, or just the sbom image?  I think only the sbom image as the container image is being created by the commands.

nalind:
> Is this big enough to warrant a Buildah v2.0 to go out with Podman v5.0?

It does not.

cmd/buildah/commit.go
nalind: That's right.  Does this read as implying otherwise?

nalind: Ah, okay.  The main discussion truncated the highlighted area.  Changing it.

TomSweeneyRedHat: I thought that was happening, but wasn't sure.  Part of it is the SBOM is a new concept and I got uber focused on why commit would pull a container image.  

In the Build doc, you added:  "base and SBOM scanner images".  What would you think about changing "image" to "SBOM scanner image" here?  At least for the first instance of "image" in the option text.  I'll go add a suggestion above.

cmd/buildah/commit.go
```diff
@@ -112,6 +123,36 @@ func commitListFlagSet(cmd *cobra.Command, opts *commitInputOptions) {
 	flags.BoolVarP(&opts.quiet, "quiet", "q", false, "don't output progress information when writing images")
 	flags.StringVar(&opts.referenceTime, "reference-time", "", "set the timestamp on the image to match the named `file`")
 	_ = cmd.RegisterFlagCompletionFunc("reference-time", completion.AutocompleteNone)
+
+	flags.StringVar(&opts.pull, "pull", "true", "pull the image from the registry if newer or not present in store, if false, only pull the image if not present, if always, pull the image even if the named image is present in store, if never, only use the image present in store if available")
```
TomSweeneyRedHat, line 127:
```suggestion
	flags.StringVar(&opts.pull, "pull", "true", "pull the SBOM Scanner image from the registry if newer or not present in store, if false, only pull the image if not present, if always, pull the image even if the named image is present in store, if never, only use the image present in store if available")
```

cmd/buildah/commit.go
nalind: Just updated the three places where we describe what a "pull" flag does, hopefully to describe the applicable cases and make them all read more closely to each other.

flouthoc:
> > Not a detailed review yet but i have a small question.
> > ```
> > * Is it mandatory to merge the behavior of `--pull` and `--sbom`, i mean if there is a use-case when user wants to pull new scanner images but keep using his older base images he cannot do that as of now. Will it be a good idea to have `--sbom-scanner-pull` or just `--sbom-pull` instead ?
> > ```
> 
> Maybe? I don't know that there's a compelling case for having them be differently controllable, but we're already adding _eight_ flags to `build`, which had more flags than I'd like even before this PR.
> 

Don't see any request or use-case for it, just saw it while reviewing, maybe we can do it later when actual request shows up.

LGTM

flouthoc (approved):
LGTM

PR #5291, opened 2024-01-24
tests: Remove a bad FROM line

edsantiago:
Some new heredoc test added "FROM blah blah python whatever",
an image that (presumably) exists on docker.io but does not
exist in our cache.

Plus, test was completely broken anyway. It would've found
the "this is the output" lines even without python, as
part of the verbose build.

Solution: don't use python. You don't need python to test a shebang.
You can use anything. 'cat' is traditional, but I choose 'rev'
because that makes it nearly impossible for the test to match
merely due to a build-step echo.

```release-note
None
```

### DIFF
--- a/tests/bud/heredoc/Containerfile.she_bang
+++ b/tests/bud/heredoc/Containerfile.she_bang
@@ -1,6 +1,7 @@
-FROM python:3.11-slim-bullseye
+FROM alpine
 RUN <<EOF
-#!/usr/bin/env python
-print('this is the output of test11')
-print('this is the output of test12')
+#!/usr/bin/env rev
+11tset fo tuptuo eht si siht
+21tset fo tuptuo eht si siht
+
 EOF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -358,8 +358,9 @@ _EOF
   skip_if_in_container
   _prefetch busybox
   run_buildah build -t heredoc $WITH_POLICY_JSON -f $BUDFILES/heredoc/Containerfile.she_bang .
-  expect_output --substring "this is the output of test11"
-  expect_output --substring "this is the output of test12"
+  expect_output --substring "#
+this is the output of test11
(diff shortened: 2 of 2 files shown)


edsantiago:
Caught by podman-buildah treadmill

flouthoc:
@edsantiago If i got your second concern right for the build-step echo part I don't think that buildah prints heredoc content the build output so everything between the heredoc content is just executed it is not printed on the build step output if that is what you were talking about.

edsantiago:
>I don't think that buildah prints heredoc content in the build output so everything between the heredoc content is just executed it is not printed on the build step output if that is what you were talking about.

It is, but I'm much more paranoid than you :-). _If buildah is working properly_, heredoc content is not printed. I can easily envision many forms of breakage in which heredoc is just passed through. Usually these would be errors, but I can also envision a missing error exit. And in that case:
```
RUN <<EOF
hi there
EOF
```
if a test only does a substring lookup for "hi there", it would pass. That's why as a general rule I always require tests to do some sort of text transform: `rev`, split or join or math formula or something. It's a good habit that has saved us on many occasions.

Great to hear from you! I hope your new adventures are treating you well!

tests/bud/heredoc/Containerfile.she_bang
```diff
@@ -1,6 +1,7 @@
-FROM python:3.11-slim-bullseye
+FROM alpine
 RUN <<EOF
-#!/usr/bin/env python
-print('this is the output of test11')
-print('this is the output of test12')
+#!/usr/bin/env rev
+11tset fo tuptuo eht si siht
+21tset fo tuptuo eht si siht
```
TomSweeneyRedHat, line 5: For a brief dyslexic moment, I thought you had dropped a naughty word at the end of these lines....

TomSweeneyRedHat:
LGTM
and happy greeen tests

PR #5296, opened 2024-01-27
Fix FreeBSD version parsing

dfr:
It was generating an error when parsing "14.0-RELEASE-p4" due to a stupid bug when attempting to check that the version part only has two parts.

When I wrote this originally, I was using systems running either 13.2-STABLE or 14.0-CURRENT and didn't properly take into account the patchleve. This factors out the version parsing and adds a unit test.

#### What type of PR is this?

/kind bug

It forces Podman and Buildah to use two jails for each container instead of one when run on the latest iteration of FreeBSD 14.0

None

```release-note
None
```

### DIFF
--- a/pkg/jail/jail.go
+++ b/pkg/jail/jail.go
@@ -4,6 +4,7 @@
 package jail
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


flouthoc (approved):
LGTM
/lgtm
/approve
/hold

rhatdan:
Test failure has nothing to do with this change so force merging to save the planet

PR #5301, opened 2024-01-30
buildah manifest: add artifact-related options

### LINKED ISSUE #5051: Image index media type and annotation
For tighter integration with some container registries such as ghcr.io, I want to annotate multi-architecture image indexes. I understand that only media of type `application/vnd.oci.image.index.v1+json` support annotations. However, I’m unable to create an image index of this type using either Buildah or Podman. Moreover, there doesn’t seem to be a way to annotate image indexes—only images referenced by an index.

```console
$ buildah manifest create manticore && buildah manifest inspect manticore
```

```
f25e51ee5f0a3be38ae63b954aa4a2367c20e514a9f66a007975ee1db8e9a217
{
    "schemaVersion": 2,
    "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
    "manifests": null
}
```

```
f25e51ee5f0a3be38ae63b954aa4a2367c20e514a9f66a007975ee1db8e9a217
{
    "schemaVersion": 2,
    "mediaType": "application/vnd.oci.image.index.v1+json",
    "manifests": null
}
```

… because, as I understand it, conformance with OCI specifications is a high priority to Buildah, and Buildah defaults to using the OCI format in other operations, such as committing and pushing images.

**Output of buildah version:**

```
Version:         1.31.3
Go Version:      go1.19.12
Image Spec:      1.1.0-rc.3
Runtime Spec:    1.1.0-rc.3
CNI Spec:        1.0.0
libcni Version:
image Version:   5.26.2
Git Commit:
Built:           Fri Aug 25 08:51:43 2023
OS/Arch:         linux/amd64
BuildPlatform:   linux/amd64
```

---

In [*cmd/buildah/manifest.go*](https://github.com/containers/buildah/blob/df1bc383407781b5ddbc3a5fdba73c3ddb6abdc7/cmd/buildah/manifest.go#L367), the manifest type is currently fixed to `manifest.DockerV2ListMediaType`.

Could `manifest create` take a `--format` option (the same way that `push` and `commit` do) and respect the `BUILDAH_FORMAT` environment variable?

Could `manifest annotate` annotate the image index directly if (1) the media type is appropriate and (2) no image manifest digest is provided?

nalind:
#### What type of PR is this?

/kind feature

Add functionality for creating artifact manifests and adding them to image indexes.  `buildah manifest add` gets a `--artifact` option for telling it to create artifact manifests, and `--artifact-type`, `--artifact-config-type`, `--artifact-layer-type`, `--artifact-exclude-titles`, and `--subject` options to fine-tune the contents of the artifact manifests it creates.

Add a `--index` flag to `buildah manifest annotate` so that it can be told to set annotations on the index itself instead of on one of the entries in the image index.

Add a `--subject` flag to `buildah manifest annotate` for setting the `subject` field of an image index.

Add a `--annotation` flag to `buildah manifest create` to allow for adding annotations to the new image index.

#### How to verify it

New integration tests!

Fixes #5051 

```release-note
`buildah manifest add` now accepts a `--artifact` option, which can create artifact manifests for arbitrary files and add those artifact manifests to an image index.
`buildah manifest annotate` now accepts a `--index` option, indicating that `--annotation` values should be added to the image index rather than an item listed in an image index.
```


rhatdan:
Can you change the common PR above to not be Do not merge?  I think common is up 2 date.

nalind:
Removing the update of github.com/containers/common from the first commit in the series.

nalind:
Oh, you meant "take the DO NOT MERGE off the description", since the previously-vendored version of common is from November.

chroot/run_common.go
```diff
@@ -511,6 +516,9 @@ func runUsingChroot(spec *specs.Spec, bundlePath string, ctty *os.File, stdin io
 	cmd.Stdin, cmd.Stdout, cmd.Stderr = stdin, stdout, stderr
 	cmd.Dir = "/"
 	cmd.Env = []string{fmt.Sprintf("LOGLEVEL=%d", logrus.GetLevel())}
+	if _, ok := os.LookupEnv(containersConfEnv); ok {
```
rhatdan, line 519: Nit: Why not use the val passed back rather then calling os.Getenv() again.if 
```
val, ok := os.LookupEnv(containersConfEnv); ok {
	cmd.Env = append(cmd.Env, containersConfEnv+"="+val))
}
```

rhatdan (approved):
LGTM

baude:
i verified it worked with multiple artifacts and quay ... great work @nalind 

/lgtm

chroot/run_common.go
nalind: Yes, that would have been better.

PR #5306, opened 2024-01-31
cgroups: check version using c/common

lsm5:
This commit reuses the cgroups version check logic from c/common and helps to get rid of the duplicated check in util/util_linux.go.

#### What type of PR is this?

/kind cleanup

uses cgroup version check from c/common and removes dup code.

#### How to verify it

Nothing breaks compared to before.

None

None

```release-note
None
```


TomSweeneyRedHat:
@lsm5 lint isn't happy with you.

lsm5:
@nalind was cgroups version check duplicated in buildah to avoid pulling in ebpf ? If that's the case I wonder if this PR is worth continuing

rhatdan:
@lsm5 needs a rebase.

lsm5:
`FreeBSD 386` cross compile fails with:
```
GOOS=freebsd GOARCH=386 GO111MODULE=on go build -mod=vendor -ldflags '-X main.GitCommit=f2a11e62450870ec09fbd997b58c783eb1d5bdb3-dirty -X main.buildInfo=1707477409 -X main.cniVersion=v1.1.2 ' -o bin/buildah.freebsd.386 -tags "containers_image_openpgp" ./cmd/buildah
# github.com/godbus/dbus/v5
vendor/github.com/godbus/dbus/v5/transport_unix.go:52:10: cannot use t (variable of type *unixTransport) as transport value in return statement: *unixTransport does not implement transport (missing method SendNullByte)
vendor/github.com/godbus/dbus/v5/transport_unix.go:58:10: cannot use t (variable of type *unixTransport) as transport value in return statement: *unixTransport does not implement transport (missing method SendNullByte)
```

Do we need to support all distro-arch combos in the `cross` target?

rhatdan:
Makes sense to me.

lsm5:
@nalind @rhatdan @TomSweeneyRedHat PTAL. How big of a concern is disabling freebsd cross compile?

TomSweeneyRedHat:
@lsm5 as we have @dfr and others contributing for BSD, I'd say it's at least a solid concern.  I'd rather not disable it unless avoidable.  @dfr do you have a differing opinion?

lsm5:
> @lsm5 as we have @dfr and others contributing for BSD, I'd say it's at least a solid concern. I'd rather not disable it unless avoidable. @dfr do you have a differing opinion?

ack, then we can't afford the Makefile change. I'll look further into this

dfr:
> @lsm5 as we have @dfr and others contributing for BSD, I'd say it's at least a solid concern. I'd rather not disable it unless avoidable. @dfr do you have a differing opinion?

I certainly want Buildah to keep working on FreeBSD - I use it for my own work and support it and Podman for other FreeBSD users. The build error is in github.com/godbus/dbus - can we avoid this for non-linux in some way? It seems likely this comes from the cgroups_unsupported.go importing github.com/coreos/go-systemd/v22/dbus for cgroups.UserConnection.

dfr:
On the other hand, godbus/dbus/v5 should at least build on FreeBSD. I fixed some nocgo in it problems back in 2022 but the version in common's vendor directory doesn't seem to have that fix.

rhatdan:
Replaced by https://github.com/containers/buildah/pull/5367

PR #5365, opened 2024-02-28
[v1.34] tests: skip_if_no_unshare(): check for --setuid

edsantiago:
unshare on RHEL8 does not support --setuid. This is causing gating tests to fail.

Solution: check for option, skip test if unavailable

Clean cherry-pick of #5360

```release-note
None
```


edsantiago:
Not sure if this is necessary -- I don't see 1.34 releases on RHEL -- but it's easier to get out of the way now than if we have a gating-test crisis in a month

PR #5366, opened 2024-02-28
manifest add: complain if we get artifact flags without --artifact

nalind:
#### What type of PR is this?

/kind other

Complain if `manifest add` is invoked with any of "--artifact-type", "--artifact-config", "--artifact-config-type", "--artifact-layer-type", "--artifact-subject", "--artifact-exclude-titles", but not in conjunction with "--artifact".

#### How to verify it

Attempt to use one of the  "--artifact-type", "--artifact-config", "--artifact-config-type", "--artifact-layer-type", "--artifact-subject", or "--artifact-exclude-titles" flags with `buildah manifest add` without also specifying "--artifact".  It can be an easy thing to overlook, and if your filename happens to look like an image name, the error message you'd have gotten about it could be more confusing.

None

I added a check for this in https://github.com/containers/podman/pull/21653, and thought it would be a good idea to add it here, too.

```release-note
None
```

### DIFF
--- a/cmd/buildah/manifest.go
+++ b/cmd/buildah/manifest.go
@@ -533,6 +533,18 @@ func manifestAddCmd(c *cobra.Command, args []string, opts manifestAddOpts) error
 			return err
 		}
 	} else {
+		var changedArtifactFlags []string
+		for _, artifactOption := range []string{"artifact-type", "artifact-config", "artifact-config-type", "artifact-layer-type", "artifact-subject", "artifact-exclude-titles"} {
+			if c.Flags().Changed(artifactOption) {
+				changedArtifactFlags = append(changedArtifactFlags, "--"+artifactOption)
+			}
+		}
+		switch {
+		case len(changedArtifactFlags) == 1:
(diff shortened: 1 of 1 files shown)


cmd/buildah/manifest.go
```diff
@@ -533,6 +533,11 @@ func manifestAddCmd(c *cobra.Command, args []string, opts manifestAddOpts) error
 			return err
 		}
 	} else {
+		for _, artifactOption := range []string{"artifact-type", "artifact-config", "artifact-config-type", "artifact-layer-type", "artifact-subject", "artifact-exclude-titles"} {
+			if c.Flags().Changed(artifactOption) {
+				return fmt.Errorf("--%s requires --artifact", artifactOption)
```
rhatdan, line 538: Can we report all of these at once, rather then one at a time?

TomSweeneyRedHat:
LGTM
and happy green test buttons.
This looks like a good add for Pdoman 5.0, should we merge this in now and/or add a 5.0 label?
@rhatdan PTAL

PR #5367, opened 2024-02-29
Update vendor of containers/common

rhatdan:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```


rhatdan:
@Luap99 are the errors about ipv6 being caused by pasta?

Luap99:
Ipv6 is not an error, the problem is `Couldn't open network namespace /proc/262717/ns/net: Permission denied` which is caused by selinux. That is the issue I was talking to you about buildah not being `container_runtime_exec_t`.

If you add `chcon -t container_runtime_exec_t` in the Makefile after we build the binary I would assume it works.

rhatdan:
Strange this would not happen before.  Will update the Makefile to see if this fixes it.

Luap99:
We never had running pasta tests on this repo enabled, ops. I actually added tests for but it required a new pasta version that was not in CI at that time so it got skipped.

Given nobody reported this in now over 6 months that seems to indicate nobody used buildah with pasta on a selinux enabled system. Or at least nobody reported it. I guess the vast majority just uses the default which is understandable.

Makefile
```diff
@@ -161,6 +161,7 @@ install.cni.sudo: gopath
 .PHONY: install
 install:
 	install -d -m 755 $(DESTDIR)/$(BINDIR)
+	test -z "${SELINUXOPT}" || chcon --verbose --reference=$(DESTDIR)$(BINDIR)/podman bin/buildah
```
Luap99, line 164: SELINUXOPT is not defined in the buildah makefile so you also have to copy over
`SELINUXOPT ?= $(shell test -x /usr/sbin/selinuxenabled && selinuxenabled && echo -Z)`

rhatdan:
No, the problem again is transition rules differing from testing and actual use.

In ordinary use people run buildah directly, which means they run as unconfined_t, which is able do to what it wants.  When run in tests, it is being launched as init_t which is confined.  So Buildah + Pasta  run within a systemd  unit file would be broken by default.

Luap99:
> In ordinary use people run buildah directly, which means they run as unconfined_t, which is able do to what it wants. When run in tests, it is being launched as init_t which is confined. So Buildah + Pasta run within a systemd unit file would be broken by default.

Well it is broken when you run from your shell as unconfined_t, the issue is that if it is started from unconfined pasta then transitions to pasta_t and pasta_t policy seems to broken I guess? Again I need to check with pasta maintainers to see what is up with that.
So we use container_runtime_t to prevent pasta policy from taking effect as pasta then runs under container_runtime_t if started there.

Makefile
```diff
@@ -162,6 +164,7 @@ install.cni.sudo: gopath
 install:
 	install -d -m 755 $(DESTDIR)/$(BINDIR)
 	install -m 755 bin/buildah $(DESTDIR)/$(BINDIR)/buildah
+	test -z "${SELINUXOPT}" || chcon --verbose -t $(SELINUXTYPE) $(DESTDIR)/$(BINDIR)/buildah
```
Luap99, line 167: you are only labelling the /usr/bin/buildah binary but the test run ./bin/buildah
The chcon should also be made on bin/buildah

Luap99:
Oh wow, it actually passed, lots of registry flakes though.

Also I find it rather concerning that a rootless buildah integration tests take over 80+ minutes. But seems like that is already the case on main so not something pasta would have caused.

Luap99 (approved):
restarted the flakes

LGTM, now it is time to eat some real pasta for me :)

/lgtm
/hold

PR #5374, opened 2024-03-04
pr-should-include-tests: use GitHub label, not commit text

edsantiago:
...to allow bypassing the check. Just like on podman.

Also, bring up to code:
  - grep -F, not fgrep
  - fix regression test script (was using wrong branch envariable)
  - add new test case

```release-note
None
```

### DIFF
--- a/tests/validate/pr-should-include-tests
+++ b/tests/validate/pr-should-include-tests
@@ -2,14 +2,13 @@
 #
 # Intended for use in CI: check git commits, barf if no tests added.
 #
+ME=$(basename $0)
+
+# Github label which allows overriding this check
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
LGTM
@rhatdan ?

lsm5:
@edsantiago all green now

TomSweeneyRedHat:
@edsantiago it was saying the same as it is now.  The message "This branch is out-of-date with the base branch" is displaying along with the shiny "Update branch" button that's just begging to be pushed.  This time I'll try to control myself and will leave the shiny "must be pushed" button alone.

PR #5385, opened 2024-03-06
Bump c/* projects, Buildah to v1.35.0 and then to v1.36.0-dev

TomSweeneyRedHat:
The fourth step in the vendor dance for Podman v5.0.

Bump:

c/storage    v1.53.0
c/image      v5.30.0
c/common v0.58.0
c/Buildah    v1.35.0

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #5386, opened 2024-03-06
conformance tests: don't break on trailing zeroes in layer blobs

nalind:
#### What type of PR is this?

/kind failing-test 

When analyzing a layer blob's contents, don't break if the blob has more zeroes padding it out even after the tar reader thinks it's hit the end of the archive.

Add more detail to the diagnostic error we print when there's a digest or length mismatch, too, in case it's triggered by something other than zero padding.

#### How to verify it

Updated conformance tests!

Fixes conformance test errors we're hitting in #5385.

```release-note
None
```

### DIFF
--- a/tests/conformance/conformance_test.go
+++ b/tests/conformance/conformance_test.go
@@ -39,6 +39,7 @@ import (
 	"github.com/containers/storage"
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/idtools"
+	"github.com/containers/storage/pkg/ioutils"
 	"github.com/containers/storage/pkg/reexec"
 	dockertypes "github.com/docker/docker/api/types"
 	dockerdockerclient "github.com/docker/docker/client"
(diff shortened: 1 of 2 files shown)


tests/conformance/conformance_test.go
```diff
@@ -952,9 +957,18 @@ func summarizeLayer(t *testing.T, imageName string, blobInfo types.BlobInfo, rea
 		layer.Headers = append(layer.Headers, header)
 		hdr, err = tr.Next()
 	}
+	io.Copy(io.Discard, tarToRead)
```
mtrmac, line 960: Consider:
- Moving this down one line, to keep the code setting `err` and checking it together
- Checking for a failure of this `io.Copy` as well.

mtrmac (commented):
LGTM

tests/conformance/conformance_test.go
mtrmac: … a linter is complaining about the latter part, failing the build.

nalind: I had wondered if a linter would complain about the second one.  Changed.

TomSweeneyRedHat:
LGTM

nalind:
Did we fix the "/hold" thing only in the podman repository, or everywhere else, too?

rhatdan:
@cevich Did we fix the /hold issue here?

TomSweeneyRedHat:
The fedora-39 w/overlay failed due to a network glitch.  I've restarted.

cevich:
> @cevich Did we fix the /hold issue here?

Yes, slash-hold is no longer required for Buildah and Podman.  The `tide` bot will not merge unless `Total Success` is green.  Though it may show status as "in merge pool" - this is okay.  Matt and I have tested the bot extensively, it will behave as expected now for these repos.

cevich:
> Are we talking these two specifically

Good question.   Yes, it's these two specifically/annoyingly.  Nobody wants to have to remember which repos have a broken bot and which do not.  The others can be setup to work the same way, and this is probably a worthwhile endeavor.  I'll make a note to PS about it during the next planning meeting.

PR #5402, opened 2024-03-14
fix links to containerignore doc

Pvlerick:
#### What type of PR is this?

/kind documentation

Documentation

#### How to verify it

N/A

None

In the selection of the ignore file, the logic is wrong; the comment here https://github.com/containers/buildah/blob/main/pkg/parse/parse.go#L1290 says that <file>.containerignore will be preferend but the code obviously does the opposite since the <file>.dockerignore is checked for in the second if.

Proving test: https://github.com/Pvlerick/buildah/blob/container-ignore-file-priority/pkg/parse/parse_test.go#L233

However, changing this logic is a breaking change, and this [file].dockerignore/[file].containerignore is not documented (at least I didn't find any mention of it)

NO

```release-note

```

### DIFF
--- a/docs/buildah-add.1.md
+++ b/docs/buildah-add.1.md
@@ -89,7 +89,7 @@ buildah add containerID 'passwd' 'certs.d' /etc
 
 ## FILES
 
-### .containerignore
+### .containerignore/.dockerignore
 
 If a .containerignore or .dockerignore file exists in the context directory,
 `buildah add` reads its contents. If both exist, then .containerignore is used.
@@ -135,7 +135,7 @@ Exclude all doc files except Help.doc when copying content into the container.
 
 This functionality is compatible with the handling of .containerignore files described here:
 
-https://github.com/containers/buildah/blob/main/docs/containerignore.5.md
+https://github.com/containers/common/blob/main/docs/containerignore.5.md
 
 ## SEE ALSO
 buildah(1), containerignore(5)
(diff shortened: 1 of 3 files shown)


docs/buildah-add.1.md
```diff
@@ -89,7 +89,7 @@ buildah add containerID 'passwd' 'certs.d' /etc
 
 ## FILES
 
-### .containerignore
+### .containerignore/.dockerignore
```
flouthoc, line 92: This appears to be a path on quick look maybe make it `.containerignore or .dockerignore`. Anyways is there any need to include `.dockerignore` in this heading ? I think the text below describes talks about `.dockerignore`.

flouthoc (commented):
One nit above otherwise PR LGTM

docs/buildah-add.1.md
TomSweeneyRedHat, line 92: I concur with @flouthoc 
```suggestion
### .containerignore or .dockerignore
```

TomSweeneyRedHat (commented):
LGTM once the nit is cleansed.

Pvlerick:
Suggestion applied.

FYI, I aligned it with what I could find here: https://github.com/containers/podman/blob/main/docs/source/markdown/podman-build.1.md.in#L635

Pvlerick:
BTW no comment on the ignore files priorities discrepancies between the doc and the code ? :-)

flouthoc (approved):
LGTM
/lgtm
/approve

PR #5436, opened 2024-03-27
Makefile - instead of calling `as` directly, use it from env var

rahilarious:
#### What type of PR is this?

> /kind api-change

/kind bug

> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

> /bin/sh: line 1: as: command not found`

error where `as` symlink doesn't exists

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ BUILDFLAGS := -tags "$(BUILDTAGS)"
 BUILDAH := buildah
 SELINUXOPT ?= $(shell test -x /usr/sbin/selinuxenabled && selinuxenabled && echo -Z)
 SELINUXTYPE=container_runtime_exec_t
+AS ?= as
 
 GO := go
 GO_LDFLAGS := $(shell if $(GO) version|grep -q gccgo; then echo "-gccgoflags"; else echo "-ldflags"; fi)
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
Changes LGTM
A couple of tests are spitting up, I think it's unrelated and have restarted them.

edsantiago (approved):

Makefile
```diff
@@ -16,6 +16,7 @@ BUILDFLAGS := -tags "$(BUILDTAGS)"
 BUILDAH := buildah
 SELINUXOPT ?= $(shell test -x /usr/sbin/selinuxenabled && selinuxenabled && echo -Z)
 SELINUXTYPE=container_runtime_exec_t
+AS ?= as
```
nalind, line 19: Is there a platform we build on where this isn't provided by a default rule, as `LD` is?

rahilarious: This is for the case where `as` command doesn't exist. Instead `x86-64-gnu-pc-linux-as` (or something similar) should be called directly.

rahilarious: Honestly, you can drop from HEAD
```
AS ?= as
STRIP ?= strip
```
I added them to make sure to not break something accidentally

nalind: `make -p -f /dev/null` with GNU make 4.4.1 on Fedora isn't showing me a predefined value for STRIP, so we'd need to keep that.

PR #5445, opened 2024-03-31
Fix caching when mounting a cached stage with COPY/ADD

aaronlehmann:
#### What type of PR is this?

/kind bug

A comment states that `avoidLookingCache` is set when a previous stage that executed as part of the build is referenced by `--mount`, to avoid reusing content from an older build of the stage:

    // Only attempt to find cache if its needed, this part
    // so that if a step is using RUN --mount and mounts
    // previous stages then it uses the freshly built stage
    // of re-using the older stage from the store.

However, stages consisting of `COPY`/`ADD` seem to be flagged with `didExecute` even if they were fetched from cache instead. I believe this is an oversight, and these stages should not prevent subsequent caching.

Also, `avoidLookingCache` would prevent a cache push, but I think it should only prevent cache lookups, since populating the cache is still useful in these caess.

#### How to verify it

Run the provided test case.

Consider the test case added in this PR:

```
FROM scratch AS stage1
COPY / /

FROM alpine
RUN --mount=type=bind,from=stage1,target=/mnt echo hi > test
```

Without this fix, the `RUN` step skips cache completely when running with `--cache-from` and `--cache-to`. Cache is neither pulled nor pushed for this step, even when the `COPY` in `stage1` is cached. This seems to happen because `stage1` has the `didExecute` flag set, regardless of whether the stage was cached or not.

It's very possible I'm misunderstanding something, but I believe the `RUN` step in test case I've added wrongly skips cache, and I'd appreciate some pointers in the right direction if what I've proposed here isn't the right solution.

None

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -1402,7 +1402,7 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			}
 		}
 
-		needsCacheKey := (len(s.executor.cacheFrom) != 0 || len(s.executor.cacheTo) != 0) && !avoidLookingCache
+		needsCacheKey := (len(s.executor.cacheFrom) != 0 && !avoidLookingCache) || len(s.executor.cacheTo) != 0
 
 		// If we have to commit for this instruction, only assign the
 		// stage's configured output name to the last layer.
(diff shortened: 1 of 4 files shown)


rhatdan:
/approve
LGTM
@nalind @flouthoc PTAL

rhatdan:
Thanks @aaronlehmann

flouthoc (changes requested):
@aaronlehmann Thanks for PR, could you fill the section of `issue` in PR description with a small reproduce so its easier to verify. As of now I'm unable to exactly understand what was the problem before this PR.

aaronlehmann:
@flouthoc: I added some notes in the "Issue" section. Please let me know if I can clarify further.

flouthoc (approved):
@aaronlehmann Thanks PR LGTM, could you please rebase it :)

aaronlehmann:
Thanks! Rebased.

PR #5470, opened 2024-04-11
CI VMs: bump to new versions with tmpfs /tmp

edsantiago:
For the last long time, Fedora CI VMs have had a disk /tmp.
Real-world setups typically have tmpfs /tmp. This switches
to CI VMs that reflect the real world.

See https://github.com/containers/automation_images/pull/340

```release-note
None
```


edsantiago:
Yay, passes with no changes anywhere! Podman is [more complicated](https://github.com/containers/podman/pull/22207), please don't merge this one until/unless there is full consensus to go this route.

edsantiago:
@containers/buildah-maintainers this is ready to go.

PR #5472, opened 2024-04-11
healthcheck: Add support for `--start-interval` and bump `imagebuilder` to `v1.2.8`

flouthoc:
healthcheck: Add support for `--start-interval` and bump `imagebuilder` to `v1.2.8`

Docker now supports `--start-interval`, start interval
is the time between health checks during the start period.

Ref: https://docs.docker.com/reference/dockerfile/#healthcheck

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### Does this PR introduce a user-facing change?

```release-note
healthcheck: Add support for --start-interval and bump`imagebuilder to v1.2.8
```

Closes: https://github.com/containers/podman/issues/21701

PR #5473, opened 2024-04-11
Don't expand RUN heredocs ourselves, let the shell do it

nalind:
#### What type of PR is this?

/kind bug

When handling RUN instructions that use heredoc syntax, don't bother interpolating environment variables and argument values, and let the command that's running handle it.

Try to improve error reporting when we're committing an image, so that we don't lose out-of-space errors through layers of compression and editing of file headers in the layer diffs.

#### How to verify it

New conformance test!

We fail to build e.g. https://github.com/devfile/developer-images/tree/main/universal/ubi8 when the heredoc references environment variables which are set in the heredoc itself, since we currently expand them before passing the heredoc as an argument to the command being invoked.

```release-note
Environment variables referenced in RUN instructions which use heredoc syntax will now be evaluated by the command being invoked rather than by the builder.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,7 +29,7 @@ require (
 	github.com/docker/distribution v2.8.3+incompatible
 	github.com/docker/docker v26.1.2+incompatible
 	github.com/docker/go-units v0.5.0
-	github.com/fsouza/go-dockerclient v1.10.1
+	github.com/fsouza/go-dockerclient v1.11.0
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/mattn/go-shellwords v1.0.12
 	github.com/moby/buildkit v0.12.5
@@ -41,7 +41,7 @@ require (
 	github.com/opencontainers/runtime-spec v1.2.0
 	github.com/opencontainers/runtime-tools v0.9.1-0.20230914150019-408c51e934dc
 	github.com/opencontainers/selinux v1.11.0
-	github.com/openshift/imagebuilder v1.2.6
+	github.com/openshift/imagebuilder v1.2.9
 	github.com/seccomp/libseccomp-golang v0.10.0
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/cobra v1.8.0
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -18,6 +18,7 @@ import (
 	"sync"
 	"syscall"
 	"time"
+	"unicode"
 
 	"github.com/containers/image/v5/pkg/compression"
 	"github.com/containers/storage/pkg/archive"
@@ -632,6 +633,15 @@ func copierWithSubprocess(bulkReader io.Reader, bulkWriter io.Writer, req reques
 	killAndReturn := func(err error, step string) (*response, error) { // nolint: unparam
 		if err2 := cmd.Process.Kill(); err2 != nil {
 			return nil, fmt.Errorf("killing subprocess: %v; %s: %w", err2, step, err)
+		}
+		if errors.Is(err, io.ErrClosedPipe) || errors.Is(err, syscall.EPIPE) {
+			err2 := cmd.Wait()
+			if errorText := strings.TrimFunc(errorBuffer.String(), unicode.IsSpace); errorText != "" {
+				err = fmt.Errorf("%s: %w", errorText, err)
+			}
+			if err2 != nil {
+				return nil, fmt.Errorf("waiting on subprocess: %v; %s: %w", err2, step, err)
+			}
 		}
 		return nil, fmt.Errorf("%v: %w", step, err)
 	}
--- a/docker/types.go
+++ b/docker/types.go
@@ -60,9 +60,10 @@ type HealthConfig struct {
 	Test []string `json:",omitempty"`
 
 	// Zero means to inherit. Durations are expressed as integer nanoseconds.
-	Interval    time.Duration `json:",omitempty"` // Interval is the time to wait between checks.
-	Timeout     time.Duration `json:",omitempty"` // Timeout is the time to wait before considering the check to have hung.
-	StartPeriod time.Duration `json:",omitempty"` // Time to wait after the container starts before running the first check.
+	Interval      time.Duration `json:",omitempty"` // Interval is the time to wait between checks.
+	Timeout       time.Duration `json:",omitempty"` // Timeout is the time to wait before considering the check to have hung.
+	StartPeriod   time.Duration `json:",omitempty"` // Time to wait after the container starts before running the first check.
+	StartInterval time.Duration `json:",omitempty"` // Time to wait between checks during the StartPeriod.
 
 	// Retries is the number of consecutive failures needed to consider a container as unhealthy.
 	// Zero means inherit.
--- a/go.sum
+++ b/go.sum
@@ -116,8 +116,8 @@ github.com/fsnotify/fsnotify v1.4.7/go.mod h1:jwhsz4b93w/PPRr/qN1Yymfu8t87LnFCMo
 github.com/fsnotify/fsnotify v1.4.9/go.mod h1:znqG4EE+3YCdAaPaxE2ZRY/06pZUdp0tY4IgpuI1SZQ=
 github.com/fsnotify/fsnotify v1.7.0 h1:8JEhPFa5W2WU7YfeZzPNqzMP6Lwt7L2715Ggo0nosvA=
 github.com/fsnotify/fsnotify v1.7.0/go.mod h1:40Bi/Hjc2AVfZrqy+aj+yEI+/bRxZnMJyTJwOpGvigM=
-github.com/fsouza/go-dockerclient v1.10.1 h1:bSU5Wu2ARdub+iv9VtoDsN8yBUI0vgflmshbeQLKhvc=
-github.com/fsouza/go-dockerclient v1.10.1/go.mod h1:dyzGriw6v3pK4O4O1u/X+vXxDDsrnLLkCqYkcLsDq2k=
+github.com/fsouza/go-dockerclient v1.11.0 h1:4ZAk6W7rPAtPXm7198EFqA5S68rwnNQORxlOA5OurCA=
+github.com/fsouza/go-dockerclient v1.11.0/go.mod h1:0I3TQCRseuPTzqlY4Y3ajfsg2VAdMQoazrkxJTiJg8s=
 github.com/go-jose/go-jose/v3 v3.0.3 h1:fFKWeig/irsp7XD2zBxvnmA/XaRWp5V3CBsZXJF7G7k=
 github.com/go-jose/go-jose/v3 v3.0.3/go.mod h1:5b+7YgP7ZICgJDBdfjZaIt+H/9L9T/YQrVfLAMboGkQ=
 github.com/go-logr/logr v0.1.0/go.mod h1:ixOQHD9gLJUVQQ2ZOR7zLEifBX6tGkNJF4QyIY7sIas=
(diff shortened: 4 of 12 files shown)


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/containers/buildah
 
-go 1.20
+go 1.21
+
+toolchain go1.21.9
```
flouthoc, line 5: @nalind I am not sure but I think this directive is a blocking thing because of `podman`, I created a thread here. https://github.com/containers/buildah/pull/5472#issuecomment-2050091140

Makefile
```diff
@@ -209,9 +209,10 @@ vendor-in-container:
 
 .PHONY: vendor
 vendor:
-	GO111MODULE=on $(GO) mod tidy
-	GO111MODULE=on $(GO) mod vendor
-	GO111MODULE=on $(GO) mod verify
+	if go env | grep -q GOTOOLCHAIN ; then go mod edit -toolchain=none ; fi
```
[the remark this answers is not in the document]
nalind: Ugh.  Backing it out.

Makefile
```diff
@@ -209,9 +209,10 @@ vendor-in-container:
...
+	GOTOOLCHAIN=local GO111MODULE=on $(GO) mod tidy
```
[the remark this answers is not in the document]
nalind: Backing it out.

Makefile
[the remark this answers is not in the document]
nalind: Going to try adding the variable to our local configuration for renovate.

Makefile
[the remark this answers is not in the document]
nalind: Okay, trying again with "export" this time.

tests/bud/no-history/Dockerfile
```diff
@@ -2,6 +2,7 @@
 # entries, but it does have at least one layer.
 
 FROM nixery.dev/shell AS first-stage
```
edsantiago, line 4: Can we get rid of `nixery` here?

edsantiago (commented):
As always you're way ahead of me. Thank you. One request inline.

tests/bud/no-history/Dockerfile
nalind: Any preference on what to change it to?

nalind: Okay, `example.com` unless/until I hear otherwise.

tests/bud/no-history/Dockerfile
```diff
@@ -1,7 +1,8 @@
 # The important thing about that first base image is that it has no history
 # entries, but it does have at least one layer.
 
-FROM nixery.dev/shell AS first-stage
+FROM example.com/shell AS first-stage
```
edsantiago, line 4: I was thinking anything on quay.io, or basically anything that does not introduce a new failure pain point.

nalind: Hmm, using example.com has the advantage of producing an image reference that can't be confused with something that exists (or might once have existed, if we come back to this in the future) in an actual registry somewhere.  How strongly are you in favor of a quay.io name?

edsantiago: I think my question was: instead of on-demand building each time, can this image be built once (with saved-committed-tracked script and instructions), say onto `quay.io/libpod/nohistory:20240422`, and then fetched on demand. This is a very difficult test to review. If any step ever breaks in a CI run, it will cost a lot of energy to investigate.

nalind: I don't believe I'm authorized to push to that repository, so that's not a thing I can do in this PR, or set up for this PR to use.  The rest I can do.

tests/bud.bats
```diff
@@ -4398,9 +4398,62 @@ EOM
...
+  echo '{"schemaVersion": 2, "manifests": [ {"mediaType": "application/vnd.oci.image.manifest.v1+json", "digest": "sha256:'${manifestdigest%% *}'", "size": '${manifestsize%% *}' } ]}' > $ocidir/index.json
+  # Write the "this is an OCI layout directory" identifier.
+  echo '{"imageLayoutVersion":"1.0.0"}' > $ocidir/oci-layout
+  # Import the image from the OCI layout into buildah's normal storage.
+  run_buildah pull --log-level=debug $WITH_POLICY_JSON oci:$ocidir
+  # Tag the image (we know its ID is the config blob digest, since it's an OCI
+  # image) with the name the Dockerfile will specify as its base image.
+  run_buildah tag ${configdigest%% *} example.com/shell
+  # Build images using our image-with-no-history as a base, to check that we
+  # don't trip over ourselves when doing so.
```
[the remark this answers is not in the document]
nalind: Adding one.

tests/bud.bats
```diff
@@ -4425,9 +4425,66 @@ EOM
 }
 
 @test "bud-implicit-no-history" {
-  _prefetch nixery.dev/shell
-  run_buildah build $WITH_POLICY_JSON --layers=false $BUDFILES/no-history
-  run_buildah build $WITH_POLICY_JSON --layers=true  $BUDFILES/no-history
+  _prefetch busybox
```
edsantiago, line 4428: On a repush, would you consider replacing all of this with
```
  testimage=quay.io/libpod/buildah-testimage-nohistory:20240422
  _prefetch $testimage
...
  check that
  run buildah on it
```

nalind: That'll _probably_ work.  I say probably because I only just realized that hard-coding the architecture of the image as "amd64" could create problems if any of the binaries being run using RUN instructions aren't coming from the current architecture's "busybox" image.

edsantiago: Gotcha - I hadn't noticed/remembered that there were `RUN`s in the containerfile. I need to rethink this. Tomorrow.

rhatdan:
@Luap99 @edsantiago PTANL

edsantiago (approved):
(commenting only on the no-history test): I have a replacement, but I'll submit it as a separate PR. I won't block this.

Makefile
```diff
@@ -209,9 +211,10 @@ vendor-in-container:
 
 .PHONY: vendor
 vendor:
-	GO111MODULE=on $(GO) mod tidy
-	GO111MODULE=on $(GO) mod vendor
-	GO111MODULE=on $(GO) mod verify
+	$(GO) mod tidy
+	$(GO) mod vendor
+	$(GO) mod verify
+	if test -n "$(strip $(shell go env GOTOOLCHAIN))"; then go mod edit -toolchain none ; fi
```
Luap99, line 217: I rather not remove the toolchain line and have behave consistently across all our projects (https://github.com/containers/skopeo/pull/2297, https://github.com/containers/common/pull/1963 and others ...)
I don't think this is wrong, the end result is the same as far as I can tell but consistency is key IMO

cc @mtrmac

tests/bud.bats
```diff
@@ -333,7 +333,7 @@ _EOF
 }
 
 @test "bud build with heredoc content" {
-  _prefetch fedora
+  _prefetch quay.io/fedora/python-311
```
Luap99, line 336: please only ever use images from quay.io/libpod/...
Second to we actually have to pull a large python image for this? I don't see the relevance of python specific code in herdoc. Can just as well test a simple shell script IMO.

tests/bud.bats
```diff
@@ -4547,17 +4547,17 @@ EOM
 }
 
 @test "bud arg and env var with same name" {
-  _prefetch centos:8
+  _prefetch busybox
```
Luap99, line 4550: same for the other image please make sure to use quay.io/libpod images if we already change them.

Luap99 (commented):
Also as far as actual review goes I am totally confused about the totally different set of changes here. Te me this look like a bunch of independent changes thrown together.

Makefile
nalind: Whatever changes we do here impact the vendoring validation tests.  I'm not averse to splitting it out, but it has effects on things that come after it.

tests/bud.bats
nalind: There's probably some value in checking that the right thing happens when the designated consumer of the heredoc is something other than the shell.  I don't think anything the quay.io/libpod organization includes a python interpreter, but if there is one, I wouldn't mind using it instead.

Makefile
mtrmac: https://go.dev/ref/mod:
> For reproducibility, the go command writes its own toolchain name in a toolchain line any time it is updating the go version in the go.mod file 

and IIRC that’s the _only_ time that happens. So, in c/image and Skopeo we have a `toolchain 1.21.0` line; then `go get` does not trigger any updates to 1.21.x as long as we stay on Go 1.21.

---

A possible unanticipated downside of the `toolchain 1.21.0` approach is that Renovate files PRs just to update the `toolchain`, as in https://github.com/containers/image/pull/2382 .

I don’t think we know for sure what Renovate does with Go ≥ 1.21 if there is no `toolchain` directive at all, AFAIK all updates to 1.21 were using the `toolchain 1.21.0` approach so far.

tests/bud/no-history/Dockerfile
mtrmac: (Just in case it is of interest, https://datatracker.ietf.org/doc/html/rfc6761#section-6.4 specifies that the `.invalid` TLD should never resolve; that might be useful for similar situations.)

Makefile
```diff
@@ -1,4 +1,6 @@
 export GOPROXY=https://proxy.golang.org
+export GOTOOLCHAIN=local
+export GO111MODULE=on
```
mtrmac, line 3: (Maybe not in this PR?) I think with current Go versions (the `go 1.21` directive in `go.mod` is being enforced, so we _can_ assume a recent version), it should be possible to never refer to this variable, and to assume modules throughout.

Makefile
mtrmac, line 217: What is this `go env` condition checking for? It’s not _absolutely_ obvious to me, my best guess is “is this a new enough Go to understand GOTOOLCHAIN?”

And if that’s the question, the `go 1.21` directive in `go.mod` ensures the answer is “yes” (or, alternatively, if building with a much older Go which does not enforce the `go 1.21` directive, the build is going to fail on a missing library feature anyway).

digester.go
```diff
@@ -141,9 +141,12 @@ func newTarFilterer(writeCloser io.WriteCloser, filter func(hdr *tar.Header) (sk
...
 		}
-		pipeReader.Close()
-		tarWriter.Close()
-		writeCloser.Close()
+		err1 := tarWriter.Close()
+		err := writeCloser.Close()
+		if err == nil {
+			err = err1
+		}
+		pipeReader.CloseWithError(err)
```
mtrmac, line 149: I always worry excessively that a failing goroutine will leave the consumer hanging.

Consider something like
```go
func someOperationGoroutine() {
	err := errors.New("Internal error: unexpected panic in someOperationGoroutine")
	defer func() { // Ensure we _always_ wake up the caller
		_ = pipeReader.CloseWithError(err) // CloseWithError(nil) is equivalent to Close(), always returns nil
                wg.Done()
	}()
	err = someOperation()
}
func someOperation() {
    // straight-line worry-free Go code with liberally sprinkled early `return`s and any other control flow
}
```

mtrmac (commented):
I also have the impression that this is a collection of quite unrelated changes; although most are individually desirable, it’s harder to understand and discuss.

In particular, it took me quite a long time to understand that the stated topic of the PR is being achieved by an update of `github.com/openshift/imagebuilder`.

Makefile
mtrmac: … I think what I’m saying here is that IIRC we haven’t considered the approach of just editing out the `toolchain` dependency.

And it might actually be a pretty good solution; if nothing else, reflects our intent more precisely than a `1.21.0` value.

… but also, I… don’t personally want to be spending much more time on `toolchain`. If this approach is better, I’ll happily help switch other repos, but I’d rather not invest much time evaluating it myself.

Makefile
mtrmac: “this variable” being `GO111MODULE`, to be explicit.

nalind:
Spun out #5493, #5494, #5495, #5496, #5497, #5498, #5499.

rhatdan:
LGTM

rhatdan:
@mtrmac @Luap99 PTAL

mtrmac (commented):
/lgtm

PR #5478, opened 2024-04-16
Setting --arch should set the TARGETARCH build arg

rhatdan:
Also fix a long standing FIXME in the test framework.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
--arch option now causes the TARGETARCH and TARGETPLATFORM  build args to get modified
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -243,26 +243,48 @@ _EOF
   assert "${checkvars[*]}" != "" \
          "INTERNAL ERROR! No 'ARG xxx' lines in $containerfile!"
 
+  ARCH=$(go env GOARCH)
   # With explicit and full --platform, buildah should not warn.
   run_buildah build $WITH_POLICY_JSON --platform linux/amd64/v2 \
               -t source -f $containerfile
-  assert "$output" !~ "missing .* build argument" \
-         "With explicit --platform, buildah should not warn"
+  assert "$output" =~ "image platform \(linux/amd64\) does not match the expected platform" \
+         "With explicit --platform, buildah should warn about pulling difference in platform"
+  assert "$output" =~ "TARGETOS=linux" " --platform TARGETOS set correctly"
+  assert "$output" =~ "TARGETARCH=amd64" " --platform TARGETARCH set correctly"
+  assert "$output" =~ "TARGETVARIANT=" " --platform TARGETVARIANT set correctly"
+  assert "$output" =~ "TARGETPLATFORM=linux/amd64/v2" " --platform TARGETPLATFORM set correctly"
 
   # Likewise with individual args
   run_buildah build $WITH_POLICY_JSON --os linux --arch amd64 --variant v2 \
               -t source -f $containerfile
-  assert "$output" !~ "missing .* build argument" \
-         "With explicit --os + --arch + --variant, buildah should not warn"
+  assert "$output" =~ "image platform \(linux/amd64\) does not match the expected platform" \
+         "With explicit --variant, buildah should warn about pulling difference in platform"
+  assert "$output" =~ "TARGETOS=linux" "--os --arch --variant TARGETOS set correctly"
+  assert "$output" =~ "TARGETARCH=amd64" "--os --arch --variant TARGETARCH set correctly"
+  assert "$output" =~ "TARGETVARIANT=" "--os --arch --variant TARGETVARIANT set correctly"
+  assert "$output" =~ "TARGETPLATFORM=linux/amd64" "--os --arch --variant TARGETPLATFORM set correctly"
 
-  # FIXME FIXME FIXME: #4319: with --os only, buildah should not warn about OS
-  if false; then
-      run_buildah build $WITH_POLICY_JSON --os linux \
-                  -t source -f $containerfile
-      assert "$output" !~ "missing.*TARGETOS" \
-             "With explicit --os (but no arch/variant), buildah should not warn about TARGETOS"
-      # FIXME: add --arch test too, and maybe make this cleaner
-  fi
+  run_buildah build $WITH_POLICY_JSON --os linux -t source -f $containerfile
+  assert "$output" !~ "WARNING" \
+         "With explicit --os (but no arch/variant), buildah should not warn about TARGETOS"
+  assert "$output" =~ "TARGETOS=linux" "--os TARGETOS set correctly"
+  assert "$output" =~ "TARGETARCH=${ARCH}" "--os TARGETARCH set correctly"
+  assert "$output" =~ "TARGETVARIANT=" "--os TARGETVARIANT set correctly"
+  assert "$output" =~ "TARGETPLATFORM=linux/${ARCH}" "--os TARGETPLATFORM set correctly"
+
+  run_buildah build $WITH_POLICY_JSON --arch amd64 -t source -f $containerfile
+  assert "$output" !~ "WARNING" \
+         "With explicit --os (but no arch/variant), buildah should not warn about TARGETOS"
+  assert "$output" =~ "TARGETOS=linux" "--arch TARGETOS set correctly"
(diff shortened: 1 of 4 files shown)


tests/bud.bats
```diff
@@ -242,24 +242,38 @@ expect_output --substring "hello"
...
-  if false; then
-      run_buildah build $WITH_POLICY_JSON --os linux \
-                  -t source -f $containerfile
-      assert "$output" !~ "missing.*TARGETOS" \
-             "With explicit --os (but no arch/variant), buildah should not warn about TARGETOS"
-      # FIXME: add --arch test too, and maybe make this cleaner
-  fi
-
+  assert "$output" =~ "image platform .* does not match the expected platform" \
+         "With explicit --platform, buildah should warn about no variant"
```
nalind, line 256: This error description doesn't seem to match the error message it's checking against.

tests/bud.bats
```diff
@@ -242,24 +242,38 @@ expect_output --substring "hello"
   # With explicit and full --platform, buildah should not warn.
   run_buildah build $WITH_POLICY_JSON --platform linux/amd64/v2 \
               -t source -f $containerfile
-  assert "$output" !~ "missing .* build argument" \
-         "With explicit --platform, buildah should not warn"
+  assert "$output" =~ "image platform .* does not match the expected platform" \
+         "With explicit --platform, buildah should warn about no variant"
```
nalind, line 246: This error description doesn't seem to match the error message it's checking against.

pkg/parse/parse.go
```diff
@@ -555,7 +564,7 @@ func PlatformsFromOptions(c *cobra.Command) (platforms []struct{ OS, Arch, Varia
 		if err != nil {
 			return nil, fmt.Errorf("unable to parse platform: %w", err)
 		}
-		if os != "" || arch != "" || variant != "" {
+		if os != runtime.GOOS || arch != runtime.GOARCH || variant != "" {
```
nalind, line 567: Do the changes in this function affect test results?

tests/bud.bats
```diff
@@ -242,24 +242,38 @@ expect_output --substring "hello"
...
+  assert "$output" !~ "WARNING" \
+         "With explicit --os (but no arch/variant), buildah should not warn about TARGETOS"
+  assert "$output" =~ "TARGETOS=linux" "--os TARGETOS set correctly"
+  assert "$output" =~ "TARGETARCH=amd64" "--os TARGETARCH set correctly"
+  assert "$output" =~ "TARGETVARIANT=" "--os TARGETVARIANT set correctly"
+  assert "$output" =~ "TARGETPLATFORM=linux/amd64" "--os TARGETPLATFORM set correctly"
+
+  run_buildah build $WITH_POLICY_JSON --arch amd64 -t source -f $containerfile
+  assert "$output" !~ "WARNING" \
+         "With explicit --os (but no arch/variant), buildah should not warn about TARGETOS"
```
nalind, line 272: This error description doesn't seem to match the error message it's checking against.

pkg/parse/parse.go
rhatdan: We are now defaulting the OS and ARCH and only care if the user changes the value to something different then the default

before on an linux/amd64 machine `--platform=linux/amd64 --os=linux --arch=amd64` blue up, now it is allowed since it ultimately does what the user wanted.

If the user does
`--platform=linux/amd64 --os=linux --arch=arm64` then the command will fail

nalind: It doesn't check that the `--platform` and `--arch`/`--os` values agree.  And what's the benefit of allowing _some_ combinations of the flags?  Why encourage people with "well, sometimes, it's allowed, in newer versions"?

rhatdan: It checks if the --arch or --os changed values from the default.
Similar to --arch="" and --os="" in the old version would have been ignored.

I think we are arguing about a nothing burger, and will relent to make you happy, even though this is the corner case of all corner cases.

pkg/parse/parse_test.go
```diff
@@ -218,6 +218,7 @@ func TestSystemContextFromFlagSet(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, sc, &types.SystemContext{
 		BigFilesTemporaryDir:        GetTempDir(),
+		OSChoice:                    "linux",
```
nalind, line 221: Should this be `runtime.GOOS`, too?

tests/conformance/conformance_test.go
```diff
@@ -3566,6 +3566,7 @@ func TestCommit(t *testing.T) {
 				derivedBuilder, err := buildah.NewBuilder(ctx, store, buildah.BuilderOptions{
 					FromImage: buildahID,
 				})
+				assert.NoErrorf(t, err, "creating a new Builder")
```
nalind, line 3569: Ah, good catch.  The `require.NoErrorf()` after the `defer` can be removed.

tests/bud.bats
```diff
@@ -246,23 +246,44 @@ _EOF
   # With explicit and full --platform, buildah should not warn.
   run_buildah build $WITH_POLICY_JSON --platform linux/amd64/v2 \
               -t source -f $containerfile
-  assert "$output" !~ "missing .* build argument" \
-         "With explicit --platform, buildah should not warn"
+  assert "$output" =~ "image platform \(linux/amd64\) does not match the expected platform" \
+         "With explicit --platform, buildah should warn about pulling difference in platform"
+  assert "$output" =~ "TARGETOS=linux" " --platform TARGETOS set correctly"
+  assert "$output" =~ "TARGETARCH=amd64" " --platform TARGETARCH set correctly"
+  assert "$output" =~ "TARGETVARIANT=" " --platform TARGETVARIANT set correctly"
```
nalind, line 253: Is TARGETVARIANT not being set to "v2", per "--platform"?

tests/bud.bats
```diff
@@ -246,23 +246,44 @@ _EOF
...
+  assert "$output" =~ "TARGETVARIANT=" "--os --arch --variant TARGETVARIANT set correctly"
+  assert "$output" =~ "TARGETPLATFORM=linux/amd64" "--os --arch --variant TARGETPLATFORM set correctly"
+
+  run_buildah build $WITH_POLICY_JSON --os linux -t source -f $containerfile
+  assert "$output" !~ "WARNING" \
+         "With explicit --os (but no arch/variant), buildah should not warn about TARGETOS"
+  assert "$output" =~ "TARGETOS=linux" "--os TARGETOS set correctly"
+  assert "$output" =~ "TARGETARCH=amd64" "--os TARGETARCH set correctly"
+  assert "$output" =~ "TARGETVARIANT=" "--os TARGETVARIANT set correctly"
+  assert "$output" =~ "TARGETPLATFORM=linux/amd64" "--os TARGETPLATFORM set correctly"
```
nalind, line 272: Are these always going to be "amd64", or would "$(go env GOARCH)" be a better match?

TomSweeneyRedHat:
Looks like this needs a refresh and test help @rhatdan

rhatdan:
@TomSweeneyRedHat @nalind @flouthoc @giuseppe @vrothberg PTAL this finally passes, and is a pretty ugly bug

flouthoc (approved):
LGTM

TomSweeneyRedHat:
LGTM
However, the branch is out of date, so I triggered a rebase of the branch.

tests/bud.bats
mtrmac: I think this reveals the fix just not being correct — the core issue is that the `buildDockerfilesOnce ` code setting `TARGET*` expects `SystemContext.*Choice` to never be `""`, and that’s not how the values are set.

Compare more in https://github.com/containers/buildah/pull/5543#pullrequestreview-2074705651

PR #5479, opened 2024-04-17
Nuke the new nixery image

edsantiago:
Registry has been down for a day. This is blocking all CI.

Switch to $SAFEIMAGE, our quay.io testimage.

```release-note
None
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -4398,9 +4398,8 @@ EOM
 }
 
 @test "bud-implicit-no-history" {
-  _prefetch nixery.dev/shell
-  run_buildah build $WITH_POLICY_JSON --layers=false $BUDFILES/no-history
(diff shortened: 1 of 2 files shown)


flouthoc (approved):
LGTM

edsantiago:
@nalind PTAL. I have a bad feeling that the reason this PR is passing is because the test is not actually a test. That is, that there is something magical about the `nixery` image that causes this test to fail, but there is no actual testing anywhere in the test to verify that this magical attribute exists in the image being tested.

`nixery` is back up, and I've pulled the image, and I'll try to figure out what is so special about it, but your help would be most appreciated.

nalind:
There's a patch in https://github.com/containers/buildah/pull/5473 that crafts a suitable image to get around this.

PR #5480, opened 2024-04-17
Integration tests: fixup use of _prefetch

nalind:
#### What type of PR is this?

/kind cleanup

* There were a number of integration tests which _should_ run faster if they used the _prefetch function to load cached copies of images that they'd otherwise have had to pull from a registry.
* There was at least one test that was calling _prefetch for an image that it didn't use.
* When prefetching images, we hadn''t been using the same registry configuration (including rewrites, short names, and mirror settings) that we passed to the buildah binary when we called it later, meaning that we didn't prefetch the copies of images that we've intentionally mirrored elsewhere.
* Some tests were removing images before returning, when storage is cleaned up and initialized fresh for every test, so they didn't need to bother.

#### How to verify it

So many updated integration tests!

None

```release-note
None
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -164,7 +164,7 @@ unit_task:
       - vendor
       - cross_build
 
-    timeout_in: 1h
+    timeout_in: 90m
 
     matrix:
         - env:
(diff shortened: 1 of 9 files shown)


flouthoc (approved):
LGTM

PR #5486, opened 2024-04-22
Tests: new tool to build a no-history image

edsantiago:
Script crafts a custom image with multiple layers but
no history; used in tests/bud.bats:bud-implicit-no-history test.

```release-note
None
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -4433,7 +4433,13 @@ EOM
 }
 
 @test "bud-implicit-no-history" {
-  _prefetch nixery.dev/shell
+  testimage=quay.io/libpod/buildah-testimage-nohistory:20240501
+  _prefetch $testimage busybox
+  run_buildah tag $testimage fakeregistry.podman.io/notreal
+
+  # Before #5276, running build with a no-history image barfed with:
+  #   initializing source containers-storage:xxx-working-container: \
+  #      internal error: history lists 2 non-empty layers, but we have 6 layers on disk
   run_buildah build $WITH_POLICY_JSON --layers=false $BUDFILES/no-history
   run_buildah build $WITH_POLICY_JSON --layers=true  $BUDFILES/no-history
 }
--- a/tests/bud/no-history/Dockerfile
+++ b/tests/bud/no-history/Dockerfile
@@ -1,7 +1,9 @@
 # The important thing about that first base image is that it has no history
-# entries, but it does have at least one layer.
+# entries, but it does have at least one layer. The test image is built
+# by test/build-nohistory-image
 
-FROM nixery.dev/shell AS first-stage
+FROM fakeregistry.podman.io/notreal AS first-stage
+COPY --from=busybox / /
 RUN date > /date1.txt
 RUN sleep 1 > /sleep1.txt
 
--- a/tests/build-nohistory-image
+++ b/tests/build-nohistory-image
@@ -1,0 +1,260 @@
+#!/bin/bash
+#
+# build-no-history-image - craft a custom image used in bud-implicit-no-history test.
+#
+# This script does not run in CI, or automated tests, or gating. In fact
+# it is meant to be run once, only once, by a human, and then (with luck)
+# never again. The purpose is to create a custom test image with certain
+# very specific attributes. This image will be used in one specific test.
+# Maybe some day other tests. The image should be static and immutable.
+# This creation script is checked into the buildah source tree simply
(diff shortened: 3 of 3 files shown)


edsantiago:
Image pushed to `quay.io/libpod/buildah-testimage-nohistory:20240422`. @nalind if this is satisfactory, please feel free to use that in #5473

tests/make-nohistory-image
```diff
@@ -0,0 +1,98 @@
...
+    if [[ ! -e $outfile ]]; then
+        dd if=/dev/urandom bs=1 count=${2:-256} of=$outfile status=none
+    fi
+}
+
+workdir=$(mktemp -d)
+
+# First layer is a copy of this script
+# FIXME: find some way to add a label or annotation or something to indicate that
+cp $0 $workdir/file01
```
nalind, line 24: This should probably be "$workdir/file1", since seq isn't zero-padding the values it produces.

edsantiago: It is now :-)

nalind:
This depends on skopeo, which we're not using in the tests themselves, so that would be a new dependency if we ran this at test-time.

tests/make-nohistory-image
nalind: I don't understand what you mean here.

edsantiago:
Oh, good point; I should make that clear in the header comments. This is NOT going to be run in CI or tests or anything ever. It will only be run by a human, on demand, as needed. Like, today. Then, in the future, when something else is needed such as pastel-colored layers with sprinkles.

The idea is to craft the image one time, like, right now (today), push it to quay, and then `bud.bats` pulls it. 

Since the crafted image is stored on quay, that's what I meant by label/annotation: something that can give a human being some sort of clue when running `inspect`.

tests/make-nohistory-image
edsantiago: Oh..... duh, my bad. I changed `seq` to `seq --format %02g` ... but apparently I need to actually commit that change.

tests/make-nohistory-image
```diff
@@ -0,0 +1,107 @@
...
+cd $workdir
+
+mkdir -p oci/blobs/sha256
+
+# Build an image config and image manifest in parallel
+# FIXME: find some way to create images for other arches?
+echo '{"architecture": "amd64", "os": "linux", "rootfs": {"type": "layers", "diff_ids": [' > config.json
```
nalind, line 40: To add labels, add `"labels": {"key1": "value1" [, "key2": "value2" [...]]},` between `"linux",` and `"rootfs":`.

tests/make-nohistory-image
```diff
@@ -0,0 +1,107 @@
...
+echo '{"schemaVersion": 2, "mediaType": "application/vnd.oci.image.manifest.v1+json", "layers": [' > manifest.json
```
nalind, line 41: To add annotations, add `"annotations": {"key1": "value1" [, "key2": "value2" [...]]},` between `"application/vnd.oci.image.manifest.v1+json",` and `"layers":`.

edsantiago: Thanks I've been trying for the past few days to convert this script into something that'll make a manifest list with all the necessary arches, but it's really hurting my brain. I'll get back to it next week.

edsantiago:
@nalind WDYT, as an alternative to #5495?

nalind:
I'm kind of lukewarm on this.
At test-time, we'd end up consulting the registry for an image that #5495 constructs locally.  If a base image used in another test was being built with a Containerfile using `scratch` as a base, I don't think we'd be trying to break the part of the test that builds that image into a separate script, pushing the image the script built to a registry, and changing the test so that it pulled the image instead of building it.

edsantiago:
I completely understand. My thinking on this is that the test script is super complicated and unmaintainable. If that test ever fails, it is likely to happen in the build-image steps and not in the make-sure-buildah-deals-with-it steps. And if that happens, the poor maintainer will spend many hours trying to figure out the build logic, then what broke, then how to fix it.

My thinking is also that there could be great value in extending this script some day such that it crafts an image with other "quirks" used for testing. (IOW, in response to your second paragraph, I think it's very important to test prebuilt known-quirky images than to spend developer time maintaining the how-do-we-build-this logic).

I'm not strong on it though. The image is now on quay, and it includes a copy of the script that built it, so it'll all be there if/when we need it.

PR #5492, opened 2024-04-30
conformance: add a test that tries to chowning a volume

nalind:
#### What type of PR is this?

/kind other

Add a conformance test that attempts to "chown" a volume declared in a base image, which produces different results depending on whether we're using the BuildKit-based builder or the V1 "classic" builder.  For now, don't try to change our behavior, and continue imitating the behavior of the classic builder.

#### How to verify it

New conformance test!

None

```release-note
None
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -18,6 +18,7 @@ import (
 	"sync"
 	"syscall"
 	"time"
+	"unicode"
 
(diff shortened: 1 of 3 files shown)


nalind:
I think this is going to get replaced as part of making the volume handling a run-time option.
/hold

nalind:
Obsoleted by #5605, closing.

PR #5494, opened 2024-04-30
Builder.cdiSetupDevicesInSpecdefConfig(): use configured CDI dirs

nalind:
#### What type of PR is this?

/kind other

Use the directories configured in containers.conf, now that containers/common can tell us what they are, and now that it provides a place to configure defaults for container tools, always override the library's default set, even if it's empty, which means we do nothing.

Switch to the default CDI registry instead of the recenty-deprecated non-global one.

#### How to verify it

Nominally updated integration test!

None

See also https://github.com/containers/common/pull/1834.

```release-note
None
```


run_linux.go
```diff
@@ -70,32 +70,25 @@ func setChildProcess() error {
...
-	// configDirs = append(configDirs, defConfig.Engine.CdiSpecDirs.Get()...)
-	if len(configDirs) > 0 {
-		if err := registry.Configure(cdi.WithSpecDirs(configDirs...)); err != nil {
-			return nil, fmt.Errorf("CDI registry ignored configured directories %v: %w", configDirs, err)
-		}
+	defConfig, err := config.Default()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get container config: %w", err)
+	}
+	configDirs = append(configDirs, defConfig.Engine.CdiSpecDirs.Get()...)
```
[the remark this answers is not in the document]
nalind: You mean [Cache.refresh](https://github.com/cncf-tags/container-device-interface/blob/main/pkg/cdi/cache.go#L158)?  I hadn't dug into it, but if I'm reading that right, the conflict resolution favors an entry with a lower priority value.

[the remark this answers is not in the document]
nalind: Ugh, I think you're right, actually.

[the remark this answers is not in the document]
nalind: Okay, setting the order to (CDI library defaults), (our configuration file), (values specified as part of the API call).

Luap99 (approved):
LGTM

run_linux.go
```diff
@@ -70,32 +70,29 @@ func setChildProcess() error {
...
 	if err != nil {
-		return nil, fmt.Errorf("creating CDI registry: %w", err)
+		return nil, fmt.Errorf("failed to get container config: %w", err)
 	}
-	var configDirs []string
+	// The CDI cache prioritizes entries from directories that are later in
+	// the list of ones it scans, so start with the library default, add
+	// our general config, then values passed to us through APIs, in that
+	// order.
+	configDirs = append(slices.Clone(defConfig.Engine.CdiSpecDirs.Get()), configDirs...)
```
Luap99, line 82: actually the append here doesn't do anything as configDirs is always empty here. I think this could be simple assignment. `configDirs = slices.Clone(defConfig.Engine.CdiSpecDirs.Get())`

nalind: Right, and we're not using the library default, either, so I need to correct that comment.

nalind:
Tweaked so that when _no_ configuration directories are listed in configuration or passed in via API, `cdiSetupDevicesInSpec()` will skip asking the CDI package about devices, because it shouldn't be expected to know about any devices.

run_linux.go
```diff
@@ -70,32 +70,34 @@ func setChildProcess() error {
...
-	// configDirs = append(configDirs, defConfig.Engine.CdiSpecDirs.Get()...)
-	if len(configDirs) > 0 {
-		if err := registry.Configure(cdi.WithSpecDirs(configDirs...)); err != nil {
-			return nil, fmt.Errorf("CDI registry ignored configured directories %v: %w", configDirs, err)
-		}
+	if len(configDirs) == 0 {
+		// No directories to scan for CDI configuration means that CDI
+		// won't have any details for setting up any devices, so we
+		// don't need to be doing anything here.
+		return deviceSpecs, nil
```
Luap99, line 92: I am not so sure we should silently ignore the case, if a user gives a cdi syntax device this likely just fails later with a more cryptic error I guess so maybe hanging fails here would be nicer.

Luap99 (commented):
I am reading the podman code and the cdi code and I don't see how this would even work here when you specifiy both a normal device path and a cdi device.
InjectDevices() will either add the cdi devices or return an error with the leftover devices BUT without actually having the cdi devices added to the spec! So you end up with just leftover devices which is wrong.

Podman seems to work around that by checking for valid cdi syntax first `parser.IsQualifiedName(device)`

run_linux.go
TomSweeneyRedHat: I'm on the fence on this.  Maybe a debug log?

TomSweeneyRedHat:
@nalind, you're hitting the merge conflict lottery tonight.  Other than the one outstanding question, LGTM

run_linux.go
nalind: Oh, it breaks if we mix qualified and unqualified names.  Will update the test to exercise that.

run_linux.go
```diff
@@ -70,39 +71,49 @@ func setChildProcess() error {
...
+		return nil, fmt.Errorf("CDI default registry ignored configured directories %v: %w", configDirs, err)
+	}
+	var qualifiedDeviceSpecs, unqualifiedDeviceSpecs []string
+	for _, deviceSpec := range deviceSpecs {
+		if parser.IsQualifiedName(deviceSpec) {
+			qualifiedDeviceSpecs = append(qualifiedDeviceSpecs, deviceSpec)
+		} else {
+			unqualifiedDeviceSpecs = append(unqualifiedDeviceSpecs, deviceSpec)
 		}
 	}
```
Luap99, line 105: I think if you do the early exit for no cdi dirs we should also do the early exit for no cdi devices, i.e. move this code up and check for `len(qualifiedDeviceSpecs) == 0` so we do not have to read all the cdi dirs in these cases

run_linux.go
```diff
@@ -70,39 +71,49 @@ func setChildProcess() error {
...
 		if err != nil {
-			logrus.Debugf("CDI device injection: %v, unresolved list %v", err, leftoverDevices)
+			return nil, fmt.Errorf("CDI device injection (leftover devices: %v): %w", leftoverDevices, err)
 		}
 	}
 	removed := slices.DeleteFunc(slices.Clone(deviceSpecs), func(t string) bool { return slices.Contains(leftoverDevices, t) })
-	logrus.Debugf("CDI taking care of devices %v, leaving devices %v", removed, leftoverDevices)
-	return leftoverDevices, nil
+	logrus.Debugf("CDI taking care of devices %v, leaving devices %v, skipped %v", removed, leftoverDevices, unqualifiedDeviceSpecs)
+	return append(leftoverDevices, unqualifiedDeviceSpecs...), nil
```
Luap99, line 116: the whole leftover case can never happen though, either iInjectDevices() returns leftoverDevices and an error or no error and no leftoverDevices so the whole dance here with leftoverDevices reads like a unnecessary complication that can never happen unless the InjectDevices() logic changes

run_linux.go
nalind: Makes sense.  Changing it.

run_linux.go
nalind: Maybe I've been burned by dependencies changing behavior under us too often, but I'm inclined to leave it there as long as the function signature gives it the option of handing us back a non-empty slice with a `nil` error value.

Luap99: I can live with that

Luap99 (commented):
LGTM

PR #5495, opened 2024-04-30
Integration tests: fake up a replacement for nixery.dev/shell

nalind:
#### What type of PR is this?

/kind other

Build a fake layers-but-no-history image that should work more or less as well for the tests that we were previously using nixery.dev/shell for.

#### How to verify it

Updated integration test!

None

```release-note
None
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -4455,9 +4455,70 @@ EOM
 }
 
 @test "bud-implicit-no-history" {
-  _prefetch nixery.dev/shell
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@edsantiago PTAL

TomSweeneyRedHat:
LGTM
I think the test errors are expected.  @edsantiago PTAL

edsantiago:
I still think this is unmaintainable and will cause someone a big headache in the future. I won't block it, but I can't be the one to merge it either.

PR #5497, opened 2024-04-30
containerImageRef.NewImageSource: merge the tar filters

nalind:
#### What type of PR is this?

/kind bug

Merge the two tar filters, if we need two, that we use when committing an image.  Try to improve passing of error information from the writing end of a pipe to the reader, so that it can be reported better.

#### How to verify it

New integration test!

None

```release-note
None
```


rhatdan:
LGTM
@flouthoc @giuseppe @mheon PTAL

giuseppe (approved):
/lgtm

PR #5499, opened 2024-04-30
Integration tests: switch some base images

nalind:
#### What type of PR is this?

/kind other

Replace some of the base images we've been using (particularly centos:8, which will EOL soon) with other images hosted on quay.io.

We already use registries.conf at test-time to redirect some image references there, so this will slightly reduce the number of registries which we need to be able to reach while running these tests.

None

```release-note
None
```


rhatdan:
LGTM

PR #5507, opened 2024-05-02
[release-1.23] conformance tests: don't break on trailing zeroes in layer blobs

nalind:
#### What type of PR is this?

/kind failing-test 

When analyzing a layer blob's contents, don't break if the blob has more zeroes padding it out even after the tar reader thinks it's hit the end of the archive.

Add more detail to the diagnostic error we print when there's a digest or length mismatch, too, in case it's triggered by something other than zero padding.

Pulled from #5386.

#### How to verify it

Conformance tests should start passing again.

None

```release-note
None
```


umohnani8:
LGTM

PR #5514, opened 2024-05-03
[skip-ci] Packit: enable c10s downstream sync

lsm5:
This commit will enable downstream syncing to CentOS Stream 10. The centos maintainer will need to manually run `packit propose-downstream` and `centpkg build` until better centos integration is in place.

This commit also builds both rhel9 and centos9 copr rpms so we can check for things like differences in golang compiler.

#### What type of PR is this?

/kind other

enables downstream syncing to c10s

#### How to verify it

can't be verified pre-merge. After a release is cut, the centos maintainer will run `packit propose downstream -p buildah-centos` to trigger a centos merge request.

None

None

```release-note
None
```

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -2,8 +2,18 @@
 # See the documentation for more information:
 # https://packit.dev/docs/configuration/
 
-specfile_path: rpm/buildah.spec
+downstream_package_name: buildah
 upstream_tag_template: v{version}
+
+packages:
+  buildah-fedora:
+    pkg_tool: fedpkg
+    specfile_path: rpm/buildah.spec
+  buildah-centos:
+    pkg_tool: centpkg
+    specfile_path: rpm/buildah.spec
+  buildah-rhel:
+    specfile_path: rpm/buildah.spec
 
 srpm_build_deps:
   - make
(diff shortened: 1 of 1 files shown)


lsm5:
@nalind @rhatdan PTAL. the centos stream jobs are failing on HashSign but the change itself should be good to go.

lsm5:
@nalind @TomSweeneyRedHat PTAL .

TomSweeneyRedHat:
All kinds of red test unhappiness @lsm5

lsm5:
> All kinds of red test unhappiness @lsm5

Yes, those are expected because of the `undefined: ecdsa.HashSign` FIPS issue that's being worked on. Having both RHEL (EPEL targets) and CentOS Stream testing would mean some tests would fail at times depending on golang compiler / toolchain differences. This is actually a good thing IMHO as it provides early warnings of upcoming RHEL issues.

On the flip side, if centos stream tests pass and epel tests fail, then it's safe to say the RHEL tests will pass after sometime.

I will check with packit team if it's possible to `hide / waive` such tests if one includes a github comment like `slash-packit ignore-foo-test`

lsm5:
This is failing on c9s because the golang update to restore the HashSign and HashVerify symbols hasn't landed in buildroot yet. The change itself should be good to go

nalind:
> This is failing on c9s because the golang update to restore the HashSign and HashVerify symbols hasn't landed in buildroot yet. The change itself should be good to go

Do we have any expectations around when golang-1.22.3-2.el9 will be showing up there?

lsm5:
> > This is failing on c9s because the golang update to restore the HashSign and HashVerify symbols hasn't landed in buildroot yet. The change itself should be good to go
> 
> Do we have any expectations around when golang-1.22.3-2.el9 will be showing up there?

@nalind i'm told whenever the next compose goes out. Hard to tell the exact date I guess.

PR #5524, opened 2024-05-10
Fix docs: 'ifnewer' vs 'newer' in buildah pull --policy option.

krrg:
#### What type of PR is this?

/kind documentation

The docs incorrectly state the parameter as being 'newer', but the actual name of this parameter is 'ifnewer', at least as of the latest version of buildah (1.35.3).

#### How to verify it
`buildah pull --policy=ifnewer <image>` (succeeds)
`buildah pull --policy=newer <image>` (fails)

```release-note
None
```


rhatdan:
I would prefer to only use newer and not document ifnewer.  Since this would match better with Podman.

krrg:
@rhatdan -- Could you clarify your intent here? 

I think we can both agree that the documentation in its current form is wrong, in the sense that it documents the existence of a `newer` pull policy that is not invalid and not supported in `buildah`. I agree with your preference; it would be nice if podman and buildah both used the same set of pull policies, but that has not been true up to now.

Surely it is better to have documentation that is accurate to the current and actual state of the world rather than aspirational but wrong?

PR #5528, opened 2024-05-15
fix CentOS/RHEL build - no BATS there

jnovy:
#### What type of PR is this?
kind bug

Without this PR CentOS 10/ RHEL10 build would fail as it doesn't have BATS package.

#### How to verify it
"rpm -qp --requires buildah*.rpm | grep bats" should output nothing on non-Fedora.

None

None

### DIFF
--- a/rpm/buildah.spec
+++ b/rpm/buildah.spec
@@ -86,7 +86,9 @@ or
 Summary: Tests for %{name}
 
 Requires: %{name} = %{version}-%{release}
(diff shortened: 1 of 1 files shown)


edsantiago:
LGTM. The `Requires: bats` line dates to 2022, so I assume this is addressing a problem with blue-robot CI, and not a problem with actual RHEL production?

edsantiago (approved):

PR #5532, opened 2024-05-16
imagebuildah: Support custom image reference lookup for cache push/pull

aaronlehmann:
#### What type of PR is this?

/kind api-change

This allows callers to provide custom `SourceLookupReferenceFunc` and `DestinationLookupReferenceFunc` for cache pull/push. These can be used to implement custom blob caches, and to wrap the reference being pushed/pulled to influence the copy behavior.

#### How to verify it

Verification isn't really necessary since this is just plumbing through functions for other code calling into buildah, but one could write a test program that embeds buildah and supplies its own funcs here.

None

n/a

None

### DIFF
--- a/define/build.go
+++ b/define/build.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"time"
 
+	"github.com/containers/common/libimage"
 	nettypes "github.com/containers/common/libnetwork/types"
 	"github.com/containers/image/v5/docker/reference"
 	"github.com/containers/image/v5/types"
@@ -342,4 +343,19 @@ type BuildOptions struct {
 	// CDIConfigDir is the location of CDI configuration files, if the files in
 	// the default configuration locations shouldn't be used.
 	CDIConfigDir string
+	// CachePullSourceLookupReferenceFunc is an optional LookupReferenceFunc
+	// used to look up source references for cache pulls.
+	CachePullSourceLookupReferenceFunc libimage.LookupReferenceFunc
+	// CachePullDestinationLookupReferenceFunc is an optional generator
+	// function which provides a LookupReferenceFunc used to look up
+	// destination references for cache pulls.
+	CachePullDestinationLookupReferenceFunc func(imageName string) libimage.LookupReferenceFunc
+	// CachePushSourceLookupReferenceFunc is an optional generator function
+	// which provides a LookupReferenceFunc used to look up source
+	// references for cache pushes.
+	CachePushSourceLookupReferenceFunc func(dest types.ImageReference) libimage.LookupReferenceFunc
+	// CachePushDestinationLookupReferenceFunc is an optional
+	// LookupReferenceFunc used to look up destination references for cache
+	// pushes
+	CachePushDestinationLookupReferenceFunc libimage.LookupReferenceFunc
 }
--- a/define/build.go
+++ b/define/build.go
@@ -349,7 +349,7 @@ type BuildOptions struct {
 	// CachePullDestinationLookupReferenceFunc is an optional generator
 	// function which provides a LookupReferenceFunc used to look up
 	// destination references for cache pulls.
-	CachePullDestinationLookupReferenceFunc func(imageName string) libimage.LookupReferenceFunc
+	CachePullDestinationLookupReferenceFunc func(srcRef types.ImageReference) libimage.LookupReferenceFunc
 	// CachePushSourceLookupReferenceFunc is an optional generator function
 	// which provides a LookupReferenceFunc used to look up source
 	// references for cache pushes.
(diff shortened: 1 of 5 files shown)


rhatdan:
Please add tests.

aaronlehmann:
> Please add tests.

I think this is a tricky thing to construct a test for since it requires pushing cache to/from a registry. The integration tests do this, but they use a real buildah binary that doesn't use these options.

I saw there is a `TestConformance` test which exercises `imagebuildah.BuildDockerfiles`, but this doesn't have a registry available to test with, and hacking either hacking that into this test or creating a similar test with a registry feels like it would involve creating a lot of infrastructure just to test something extremely simple.

One option would be to add hidden flags to the `buildah build` subcommand that rewrite the cache source/destination refs (basically just for testing purposes - I don't think they would have a practical use), and exercise that in integration tests. Does that sound like a good approach? Do you have any other ideas?

rhatdan:
No after looking at the PR more closely, I think we can let it in without tests.

aaronlehmann:
Cool, are there any other changes you'd like here?

pull.go
```diff
@@ -62,7 +68,12 @@ func Pull(ctx context.Context, imageName string, options PullOptions) (imageID s
 	libimageOptions.OciDecryptConfig = options.OciDecryptConfig
 	libimageOptions.AllTags = options.AllTags
 	libimageOptions.RetryDelay = &options.RetryDelay
-	libimageOptions.DestinationLookupReferenceFunc = cacheLookupReferenceFunc(options.BlobDirectory, types.PreserveOriginal)
+	libimageOptions.SourceLookupReferenceFunc = options.SourceLookupReferenceFunc
+	if options.DestinationLookupReferenceFunc != nil {
+		libimageOptions.DestinationLookupReferenceFunc = options.DestinationLookupReferenceFunc
+	} else {
+		libimageOptions.DestinationLookupReferenceFunc = cacheLookupReferenceFunc(options.BlobDirectory, types.PreserveOriginal)
```
mtrmac, line 75: The reference function is ignored on this path. That should either be fixed, or the combination of options should be rejected; or, at the very least, it should be clearly documented.

Same on push.

define/build.go
```diff
@@ -342,4 +343,19 @@ type BuildOptions struct {
 	// CDIConfigDir is the location of CDI configuration files, if the files in
 	// the default configuration locations shouldn't be used.
 	CDIConfigDir string
+	// CachePullSourceLookupReferenceFunc is an optional LookupReferenceFunc
+	// used to look up source references for cache pulls.
+	CachePullSourceLookupReferenceFunc libimage.LookupReferenceFunc
+	// CachePullDestinationLookupReferenceFunc is an optional generator
+	// function which provides a LookupReferenceFunc used to look up
+	// destination references for cache pulls.
+	CachePullDestinationLookupReferenceFunc func(imageName string) libimage.LookupReferenceFunc
```
mtrmac, line 352: If `imageName` is coming from a well-typed `docker.Reference`, it should use a well-typed value, not a vague string. Actually I think it should use a generic `types.ImageReference`, both for symmetry, and not to box us in.

mtrmac (commented):
I am just randomly passing by and I am expressing no opinion on the cost/benefit of adding these indirections.

pull.go
aaronlehmann: I'm not sure I follow this; this path is only reached if `options.DestinationLookupReferenceFunc` is nil. Which combination of options do you mean?

mtrmac: I’m sorry, I‘ve been really unclear.

The combination of `options.SourceLookupReferenceFunc` + `options.BlobDirectory` results in `BlobDirectory` being ignored AFAICS; and `BlobDirectory != ""`, although under-documented, is basically the opt in to using the built-in blob cache.

So, in the pull options here, push options, and `buildah/define/*`, the `BlobDirectory` option and the options added by this PR are mutually exclusive.

aaronlehmann: Thanks, added some documentation to clarify when `BlobDirectory` is overridden.

mtrmac: Document that in `PullOptions` and `PushOptions` as well, please.

aaronlehmann: Done.

mtrmac (commented):
Thanks!

Implementation LGTM. I’ll let others decide about the cost/benefit of the feature.

rhatdan:
/lgtm
/approve
Thanks @aaronlehmann

rhatdan:
We are going to have to revert this PR.  We should not have let libimage leak into the buildah/define directory.  This forces podman-remote to grow hugely in size.  We should define these functions in a directory other then define/ and then pass them into the executor.

PR #5537, opened 2024-05-21
bud tests: fix breakage when vendoring into podman

edsantiago:
1. Difference in error messages, podman does not emit "building
    system context". I think it's OK to just check for the
    "--platform may not be used with" portion.

 2. Very weird: $TESTSDIR (the plural one) was eliminated in April 2022
    but one instance snuck in, probably a PR that was not rebased
    onto main. This is blowing up in podman tests because of course
    there is no $TESTSDIR. But I have **NO IDEA** how this is passing
    in buildah, nor how it passed in podman for two years.

```release-note
None
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -283,7 +283,7 @@ _EOF
   for option in "--arch=arm64" "--os=windows" "--variant=v2"; do
     run_buildah 125 build $WITH_POLICY_JSON --platform linux/amd64 ${option} \
                 -t source -f $containerfile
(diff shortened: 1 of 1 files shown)


nalind:
My expectation is that the CLI parsing logic flagged an error, as it's expected to, and this happened before execution reached any of the logic that would have tried to read the policy file from the incorrectly-specified location.

edsantiago:
Thank you! That explains why buildah CI passes. I still don't understand how podman CI passes, but am OK with not knowing.

PR #5543, opened 2024-05-23
build: be more selective about specifying the default OS

nalind:
#### What type of PR is this?

/kind failing-test 

Only add the default OS to a build target platform struct if the architecture was specified without one, so that the pull logic in containers/common doesn't override our pull policy when it doesn't need to.

#### How to verify it

New integration test!

This should restore the ability to use `--pull=missing`, which was inadvertently broken in #5478.

```release-note
None
```

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -222,7 +222,7 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 	systemContext := options.SystemContext
 	for _, platform := range options.Platforms {
 		platformContext := *systemContext
-		if platform.OS == "" {
+		if platform.OS == "" && platform.Arch != "" {
 			platform.OS = runtime.GOOS
 		}
 		platformSpec := internalUtil.NormalizePlatform(v1.Platform{
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -6811,3 +6811,17 @@ _EOF
   assert "$status" -eq 2 "exit code from ls"
   expect_output --substring "No such file or directory"
 }
+
+@test "pull policy" {
+  echo FROM busybox > ${TEST_SCRATCH_DIR}/Containerfile
+  arch=amd64
+  if test $(arch) = x86_64 ; then
+    arch=arm64
+  fi
+  # specifying the arch should trigger "just pull it anyway" in containers/common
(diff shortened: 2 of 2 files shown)


flouthoc (approved):
LGTM

mtrmac (commented):
I realize we are in urgent-hot-fix mode, and I don’t want to interfere with that.

Afterwards, dare I ask for some cleanups?
- It’s _absolutely_ unclear to me how setting `platform.OS` is supposed to result in setting `TARGETARCH`. Why are the two related? I don’t understand how #5478 fixes its subject.  Either the fix should be something completely different, or the relationship should have been recorded in a comment warning about the non-obvious relationship; or maybe the PR is doing something else than it claims to.
- AFAICT `internalUtil.NormalizePlatform` == `libimage/platform.Normalize` always returns `""` for an `""` input (i.e. it only normalizes non-empty inputs). So the added check could be moved below the `NormalizePlatform` call; that would make the meaning of the `platform.OS != "" ` condition clearer.
- Looking at the code in `buildDockerfilesOnce` setting the `TARGET*` variables, that assumes that `OSChoice`/`ArchitectureChoice`/`VariantChoice` are _always_ set. The code after #5478, at best, changes that for `OSChoice`, but not for the other two. It seems to me that the right fix would be either to change `buildDockerfilesOnce` to not make that assumption, or to reach the way all the way down inside `libimage/plaform` to _centralize_ the meaning of `""` to be `runtime.GO*`; it certainly would not do to have that assumption, in the worst case, implemented in `libimage` for variants, in `BuildDockerfiles` for OS, and in `buildDockerfilesOnce` for arch.
- Please clarify the semantics of `BuildOptions.{Architecture,OS}`, and then update users. `stageExecutor.commit` seems to consider `""` valid, but the `LogSplitByPlatform` determination of `logFile` seems not to expect that. AFAICS one of the two is wrong.
- `BuildOptions.SystemContext` is documented to hold credentials; if it holds the platform as well, that documentation should be updated.
- I’d love to see the code turning `options.Platforms[*]` into `platformOptions.{OS,Architecture,SystemContext}` (or however the information ends up being represented) split into a separate function with a unit test.

TomSweeneyRedHat:
@nalind all kinds of test unhappiness and a rebase is wanted.

PR #5083, opened 2023-10-10
chroot.setupChrootBindMounts: pay more attention to flags

nalind:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

Pay better attention to nodev/noexec/nosuid/readonly flags on bind, overlay, and tmpfs mounts.  Stop quietly adding "nodev" when it isn't asked for.

#### How to verify it

New integration test!

#### Which issue(s) this PR fixes:

This fixes certain [failures](https://github.com/openshift/origin/pull/27198) I've been seeing while attempting to get OpenShift's builds working in unprivileged user namespaces, where the builder uses buildah as a library and specifies that secrets by bind-mounted for RUN instructions using nodev/noexec/nosuid/ro, where the builder has to be just a bit more careful because, once you untangle the user namespace ID mappings, it doesn't actually own the underlying content.

#### Special notes for your reviewer:

Once merged, I'm going to want to cherry pick this to the 1.32 branch for use in OpenShift.

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #5544, opened 2024-05-23
fix secret mounts for env vars when using chroot isolation

### LINKED ISSUE #5185: Image build with secret mounts stopped working v1.32
**Description**

We have a gitlab-ci with a gitlab-runner (running with podman). The runner has `CAP_SYS_ADMIN` privileges.
Building the containers worked for buildah version from 1.29 to 1.31 but stopped working with buildah 1.32.
The CI fails with:
`error running subprocess: remounting "/var/tmp/buildah4082641103/mnt/rootfs/run/secrets/Test1" in mount namespace with flags 0x1 instead of 0x0: operation not permitted`

1. Dockerfile
```
FROM docker.io/library/debian:bullseye
MAINTAINER Konstantin Zangerle <konstantin.zangerle@kit.edu>

CMD [ "/bin/bash" ]
ENTRYPOINT []
ENV DEBIAN_FRONTEND=noninteractive

RUN --mount=type=secret,id=Test1 --mount=type=secret,id=Test2 findmnt && ls -lahZ /run/secrets;  \
	cat /run/secrets/Test1
```
3. Gitlab-CI
```
image: quay.io/buildah/stable:v1.32

#before_script:
#  - podman login -u "${HUB_USER}" -p "${HUB_PASS}" hub.scc.kit.edu:5000

build:
  stage: build
  script:
    - buildah --version
    - buildah info
    - findmnt
    - echo -n "einStringstehtindergitlabci" > /builds/REPO_USER
    - echo $Test2 > /builds/REPO_PW
    - head /builds/REPO*
    - buildah bud --log-level debug
      --secret id=Test1,src=Test1,type=env
      --secret id=Test2,src=Test2,type=env
      --no-cache .
    - head /builds/REPO*
```

Buildah fails with `error running subprocess: remounting "/var/tmp/buildah4082641103/mnt/rootfs/run/secrets/Test1" in mount namespace with flags 0x1 instead of 0x0: operation not permitted`

Buildah builds image like v1.31 without failing.

**Buildah information**
```
$ buildah --version
buildah version 1.32.2 (image-spec 1.1.0-rc.4, runtime-spec 1.1.0)
$ buildah info
{
    "host": {
        "CgroupVersion": "v2",
        "Distribution": {
            "distribution": "fedora",
            "version": "39"
        },
        "MemFree": 1488[18](https://git.scc.kit.edu/KIT-CERT/test-gitlab-ci-buildah/-/jobs/1131210#L18)28864,
        "MemTotal": 33670135808,
        "OCIRuntime": "crun",
        "SwapFree": 0,
        "SwapTotal": 0,
        "arch": "amd64",
        "cpus": 16,
        "hostname": "runner-rxsnej93-project-34890-concurrent-0",
        "kernel": "6.1.0-13-cloud-amd64",
        "os": "linux",
        "rootless": true,
        "uptime": "890h 48m 16.69s (Approximately 37.08 days)",
        "variant": ""
    },
    "store": {
        "ContainerStore": {
            "number": 4
        },
        "GraphDriverName": "overlay",
        "GraphOptions": [
            "overlay.imagestore=/var/lib/shared",
            "overlay.mount_program=/usr/bin/fuse-overlayfs",
            "overlay.mountopt=nodev,fsync=0"
        ],
        "GraphRoot": "/var/lib/containers/storage",
        "GraphStatus": {
            "Supports d_type": "true",
            "Supports shifting": "true",
            "Supports volatile": "true",
            "Using metacopy": "false"
        },
        "ImageStore": {
            "number": [29](https://git.scc.kit.edu/KIT-CERT/test-gitlab-ci-buildah/-/jobs/1131210#L29)
        },
        "RunRoot": "/run/containers/storage"
    }
}
```

jonahbull:
#### What type of PR is this?

/kind bug

As discussed in #5185, attempting to build an image that uses environmental variables as build secrets fails with the message:
```
error running subprocess: remounting "/var/tmp/buildah3153589124/mnt/rootfs/run/secrets/MYSECRET" in mount namespace with flags 0x1 instead of 0x0: operation not permitted
```
In my particular use case, this is using the VFS storage driver and chroot isolation. I haven't been able to reproduce using other isolation types, not sure if other folks have.

As far as I can tell, the cause is that before #5083, when running with `chroot` isolation `ro` mounts like secrets from env vars would explicitly have the `unix.MS_NOEXEC`, `unix.MS_NOSUID`, `unix.MS_NODEV` flags set when they were remounted: https://github.com/containers/buildah/pull/5083/files#diff-c287fcc22d32f0775a98f65ec2db880f887c6c392f8e5ed6d7320a24560c1949L468.

Now when running with `chroot` isolation `ro` mounts like secrets from env vars are not getting those same flags set and so the remount operation fails. Specifically it looks like in this case we are missing the `unix.MS_NOSUID` and `unix.MS_NODEV` flags.

I think the flags aren't being set because when we calculate `effectiveUnimportantFlags` [here](https://github.com/containers/buildah/blob/fb316cd64d864cac6903550bacfad395e91b33a5/chroot/run_linux.go#L520), we end up with `0b100000`. Forgive the possibly overly-verbose notes here, but I'm not well-versed in these kinds of bitwise operations so it was helpful for me to add as much detail as possible when debugging.
```go
possibleImportantFlags := uintptr(unix.ST_NODEV | unix.ST_NOEXEC | unix.ST_NOSUID | unix.ST_RDONLY)
// fs.Flags: 0b100110 possibleImportantFlags: 0b1111 ^possibleImportantFlags: 0b10000
effectiveUnimportantFlags := uintptr(fs.Flags) & ^possibleImportantFlags
// effectiveUnimportantFlags: 0b100000
```
When we pass effectiveUnimportantFlags to `mountFlagsForFSFlags`, none of the possible fsFlags match `0b100000`, so it returns `0b0`.

This change adds special handling for read-only mounts when we need to do a remount to try to get the desired flags to stick. If we've requested a read-only mount (`unix.ST_RDONLY` is set in `requestFlags`), then we add any `possibleImportantFlags` that are set in `fs.Flags` to `remountFlags` so the remount operation doesn't fail because they are missing

#### How to verify it

I've added a test to `bud.bats` that reproduces the issue when run locally with a version of buildah that doesn't have the fix applied.

I could definitely have missed something, but this change appears to fix the issue in #5185 without breaking other use cases, at least from my limited local testing (running the `bud.bats` and `chroot.bats` suites).

Fixes #5185

```release-note
None
```

### DIFF
--- a/chroot/run_linux.go
+++ b/chroot/run_linux.go
@@ -518,7 +518,12 @@ func setupChrootBindMounts(spec *specs.Spec, bundlePath string) (undoBinds func(
 		if effectiveImportantFlags != expectedImportantFlags {
 			// Do a remount to try to get the desired flags to stick.
 			effectiveUnimportantFlags := uintptr(fs.Flags) & ^possibleImportantFlags
-			if err = unix.Mount(target, target, m.Type, unix.MS_REMOUNT|bindFlags|requestFlags|mountFlagsForFSFlags(effectiveUnimportantFlags), ""); err != nil {
+			remountFlags := unix.MS_REMOUNT | bindFlags | requestFlags | mountFlagsForFSFlags(effectiveUnimportantFlags)
+			// If we are requesting a read-only mount, add any possibleImportantFlags present in fs.Flags to remountFlags.
+			if requestFlags&unix.ST_RDONLY == unix.ST_RDONLY {
+				remountFlags |= uintptr(fs.Flags) & possibleImportantFlags
+			}
+			if err = unix.Mount(target, target, m.Type, remountFlags, ""); err != nil {
 				return undoBinds, fmt.Errorf("remounting %q in mount namespace with flags %#x instead of %#x: %w", target, requestFlags, effectiveImportantFlags, err)
 			}
 			// Check if the desired flags stuck.
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -6825,3 +6825,12 @@ _EOF
   # both of these should have just been the base image's ID, which shouldn't have changed the second time around
   cmp ${TEST_SCRATCH_DIR}/image1.txt ${TEST_SCRATCH_DIR}/image2.txt
 }
+
+# Verify: https://github.com/containers/buildah/issues/5185
+@test "build-test --mount=type=secret test from env with chroot isolation" {
+  skip_if_root_environment "Need to not be root for this test to work"
+  local contextdir=$BUDFILES/secret-env
+  export MYSECRET=SOMESECRETDATA
+  run_buildah build $WITH_POLICY_JSON --no-cache --isolation chroot --secret id=MYSECRET -t test -f $contextdir/Dockerfile
+  expect_output --substring "SOMESECRETDATA"
+}
--- a/tests/bud/secret-env/Dockerfile
+++ b/tests/bud/secret-env/Dockerfile
@@ -1,0 +1,3 @@
+FROM alpine
+RUN --mount=type=secret,id=MYSECRET \
+     printf "%s\n" $(cat /run/secrets/MYSECRET)


jonahbull:
Looks like my initial naive attempt doesn't _quite_ work. I'll poke at getting the unit tests passing (they are failing on `TestMounts/dev,exec,suid,rw`) but if anyone who happens by has a different solution in mind I'm all ears.

rhatdan:
Thanks @jonahbull 
LGTM
/approve
@flouthoc @nalind @giuseppe @mheon PTAL

giuseppe (approved):
/lgtm

PR #2886, opened 2021-01-05
Native OSX build

cevich:
***Depends on:*** #2631

#### What type of PR is this?

> /kind feature

#### What this PR does / why we need it:

These changes execute the cross-compile build naively on an OSX VM to more thoroughly verify multi-platform support (other than on Linux).  In other words, it confirms that any Linux-only changes do not leak into non-Linux environments.

#### How to verify it

The new/replaced Cirrus-CI task `Cross Compile` will pass.

PR #5551, opened 2024-05-29
Execute cross-build task using PW pool

cevich:
#### What type of PR is this?

/kind other

Previously this task ran using the Cirrus-CI compute service, consuming
compute credits.  However, since podman is already using a persistent
worker pool for CI, it can also be leveraged for use here.  Since Mac
resources are relatively expensive, it also makes financial sense to
re-use infrastructure where possible.

#### How to verify it

The `cross_build_task` will pass, and a manual check of the worker it ran on will show no abnormalities.

None

* This change also restricts the cross-build task to only running on
  the default branch.  This is necessary because the PW Pool environment
  is always rolling forward and may not match what was in place at the
  time a release-branch was created.
* This runs buildah tasks on the infra originally designed for
  podman-machine testing.  Coordinating changes to the underlying
  setup/automation for two repos is more complex and challenging.
* Payment to Cirrus for Mac compute resources is a manual process,
  as opposed to automated cloud-payments.
* The more the podman-machine CI infra is shared, the cheaper
  maintaining it becomes.

```release-note
None
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -135,23 +135,36 @@ vendor_task:
 cross_build_task:
     name: "Cross Compile"
     alias: cross_build
-    only_if: >-
+    only_if: |
+        ($CIRRUS_BRANCH =~ "pull/.*" ||
+         $CIRRUS_BRANCH == $CIRRUS_DEFAULT_BRANCH) &&
         $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*'
-
-    osx_instance:
-        image: ghcr.io/cirruslabs/macos-ventura-base:latest
-
-    script:
-        - brew update
-        - brew install go
-        - brew install go-md2man
-        - brew install gpgme
+    persistent_worker: &mac_pw
+        labels:
+            os: darwin
+            arch: arm64
+            purpose: prod
+    env:
+        CIRRUS_WORKING_DIR: "$HOME/ci/task-${CIRRUS_TASK_ID}"
+        # Prevent cache-pollution fron one task to the next.
+        GOPATH: "$CIRRUS_WORKING_DIR/.go"
+        GOCACHE: "$CIRRUS_WORKING_DIR/.go/cache"
+        GOENV: "$CIRRUS_WORKING_DIR/.go/support"
+        GOSRC: "$HOME/ci/task-${CIRRUS_TASK_ID}"
+        TMPDIR: "/private/tmp/ci"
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
LGTM
but the CI and even more, the mac, are not in my wheelhouse.

cevich:
> are not in my wheelhouse.

Hmmm, there are business issues here as well.  I updated the "Special notes for your reviewer" section.  I think we should discuss this PR as a team.

I'll move it to draft so it doesn't accidentally get merged.

cevich:
After discussing, there was no major opposition from the team.

cevich:
force-push: Rebased.

cevich:
This should be ready to go now.

nalind:
Are the conditions that caused us to start running this job on a Mac in https://github.com/containers/buildah/pull/2886 still the case?

cevich:
> Are the conditions that caused us to start running this job on a Mac in #2886 still the case?

It's a good question, and my limited golang knowledge was even worse _way_ back then.  Taking a fresh look at [what this actually does](https://cirrus-ci.com/task/5248207716876288?logs=build#L5), it's simply running `go build` in a loop over a bunch of `GOOS` and `GOARCH` values.  So the platform it's executing on probably doesn't matter all that much in terms of answering the "does buildah compile on GOOS=foo GOARCH=bar" question.  Am I wrong?

Assuming the host environment doesn't matter for golang cross-compiles, then I would actually prefer we move this back onto a regular, boring, Fedora VM.  Running on the Macs carries extra risk and some maintenance burden.

OTOH if the platform does matter, then we probably want this using the pool to save a little money.

---

Tangential but related question: Is it possible to run `buildah` on a Mac?  If so, it might be nice to have even a simple `buildah -v` Mac test, [like we do with skopeo.](https://cirrus-ci.com/task/5655289146376192?logs=test#L66).  That would be a worthwhile use of the pool.

nalind:
> Tangential but related question: Is it possible to run `buildah` on a Mac? If so, it might be nice to have even a simple `buildah -v` Mac test, [like we do with skopeo.](https://cirrus-ci.com/task/5655289146376192?logs=test#L66). That would be a worthwhile use of the pool.

The binary will likely compile, so "buildah -v" shouldn't crash, but I don't think we have any actual expectations.

cevich:
> but I don't think we have any actual expectations.

Great, thanks for the info.

I'll close this and open another to move this task onto a regular, boring, Fedora VM.

PR #5574, opened 2024-06-06
tests: set _CONTAINERS_USERNS_CONFIGURED=done for libnetwork

nalind:
#### What type of PR is this?

/kind other

Starting in common 0.59.1, github.com/containers/common's libnetwork no longer attempts to detect when it's running in a user namespace to decide where a lock file that it uses will be.  Since this test is doing user namespace setup on its own as part of the test, we need to explicitly tell it to not use the one that belongs to the node's root user.

#### How to verify it

Updated integration test!

Unblocks #5567.

```release-note
None
```


rhatdan:
LGTM

Luap99 (commented):
LGTM to work around that for now.

I do wonder though maybe it is better to just drop https://github.com/containers/buildah/commit/14d58918fb125293c2804dd93a4eac0c452e9da9 now that we only support netavark?

PR #5580, opened 2024-06-09
buildah: fix a nil pointer reference on FreeBSD

dfr:
When setting up resolv.conf on Linux we need to reference the Linux namespaces from spec.Linux. This is not needed for FreeBSD and this change moves the dereference of spec.Linux to addResolvConfEntries and makes it conditional.

#### What type of PR is this?

/kind bug

This is needed to make 'buildah run' work on FreeBSD if networking is enabled.

#### How to verify it

```
# c=$(buildah from quay.io/dougrabson/busybox)
# buildah run $c ls
```

None

```release-note
None
```

### DIFF
--- a/run_common.go
+++ b/run_common.go
@@ -84,10 +84,15 @@ func (b *Builder) createResolvConf(rdir string, chownOpts *idtools.IDPair) (stri
 
 // addResolvConf copies files from host and sets them up to bind mount into container
 func (b *Builder) addResolvConfEntries(file string, networkNameServer []string,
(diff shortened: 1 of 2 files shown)


dfr:
No tests needed since there should be no functional different to the Linux build and existing tests can verify that.

rhatdan:
/approve
LGTM

rhatdan:
@nalind @giuseppe @flouthoc @ashley-cui @lsm5 PTAL

flouthoc (approved):
LGTM

PR #5584, opened 2024-06-11
CI: use local registry

edsantiago:
CI VMs now include a local registry prepopulated with all
container images needed by tests.

```release-note
None
```


tests/bud.bats
```diff
@@ -4577,9 +4577,11 @@ EOM
 }
 
 @test "bud copy chown with newuser" {
-  _prefetch quay.io/fedora/fedora
+  _prefetch $SAFEIMAGE
```
edsantiago, line 4580: Eliminate unused image

tests/bud.bats
```diff
@@ -6325,10 +6327,10 @@ _EOF
   expect_output --substring "world"
 }
 
-@test "bud-verify-if-we-dont-clean-prexisting-path" {
+@test "bud-verify-if-we-dont-clean-preexisting-path" {
   skip_if_no_runtime
   skip_if_in_container
-  _prefetch alpine debian
+  _prefetch alpine ubuntu
```
edsantiago, line 6333: eliminate unused image (for all purposes here, the ubuntu image works the same as debian)

tests/digest.bats
```diff
@@ -29,35 +29,6 @@ fromreftest() {
   rm -fr ${pushdir}
 }
 
-@test "from-by-digest-s1" {
```
edsantiago, line 32: Welp. Docker registry image no longer supports v1 schema. Code has been excised. There is no way to push a v2s1 image to this registry. Is it OK to just remove all this v1 testing?

tests/from.bats
```diff
@@ -110,22 +110,26 @@ load helpers
   run_buildah rm $output
   run_buildah rmi docker.io/alpine:latest
 
-  run_buildah from --quiet --pull=true $WITH_POLICY_JSON docker.io/centos:7
-  run_buildah rm $output
-  run_buildah rmi docker.io/centos:7
+  # FIXME FIXME FIXME: I don't see the point of these. Any reason not to delete?
+#  run_buildah from --quiet --pull=true $WITH_POLICY_JSON docker.io/centos:7
+#  run_buildah rm $output
+#  run_buildah rmi docker.io/centos:7
```
edsantiago, line 116: These all look like dups of the code above.

tests/mkcw.bats
```diff
@@ -58,7 +58,7 @@ function mkcw_check_image() {
     skip "cryptsetup not found"
   fi
   _prefetch busybox
-  _prefetch bash
+  _prefetch $SAFEIMAGE
```
edsantiago, line 61: If there is/was anything special about `bash`, it was not obvious from comments, so it gets nuked

edsantiago (commented):
I was not expecting this to be so easy! 

@containers/buildah-maintainers this is not yet ready for merge -- I need to figure out a way to conditionalize on-my-desktop runs vs in-CI-using-cache runs -- but it's very close. Feedback welcome.

edsantiago:
Two flakes, both in what looks like network pulls ([dnf](https://api.cirrus-ci.com/v1/task/6508542952210432/logs/integration_test.log), [image pull from k8s.io](https://api.cirrus-ci.com/v1/task/5945592998789120/logs/integration_test.log)). The k8s.io one is probably a good candidate for future caching.

edsantiago:
[Podman PR](https://github.com/containers/podman/pull/22726) has merged.  I'm considering this ready. @containers/buildah-maintainers PTAL.

rhatdan:
/approve
LGTM
@nalind @giuseppe @flouthoc @mheon PTAL

tests/mkcw.bats
nalind: The `bash` image has three layers, compared to, say `busybox`, which has one.  The test attempts to confirm that content from the base image and content added at conversion time both show up in the disk image that `mkcw` is generating.

edsantiago: Thank you! `$SAFEIMAGE` has two layers, would that be sufficient? If not, could you suggest a three-layer image from this list? https://github.com/containers/automation_images/pull/357/files#diff-7b013edc1f5a127319d75ae1f8ae443e0bfa4d891960c27ee284bbb107ddd290R56

nalind: One of those layers is empty, but it should be fine.  The error we were attempting to avoid was causing the disk image to lose content from layers that weren't the last one (and as far as this test is concerned, for images with one layer, the only layer is also the last layer), and the presence of the second layer, empty as it is, means the image still meets the test expectations.

edsantiago: Thanks. Repushed with a warning comment

flouthoc (approved):
LGTM

edsantiago:
ping, this code has been working well in podman. I'd like to get this merged so I can get working on the required treadmill changes (for vendoring into podman).

giuseppe (approved):
LGTM

PR #5595, opened 2024-06-18
containerImageRef.NewImageSource(): move the FROM comment to first

nalind:
#### What type of PR is this?

/kind cleanup

If we're prepending history entries before the one for "this" commit, make sure the "FROM $baseimage" comment gets set on the first history entry that we add, not just the one goes with this (maybe) layer diff. In layers=false mode, the output was so, so confusing otherwise.

#### How to verify it

New integration test!

N/A

```release-note
Images built using `buildah build` without `--layers=true` will describe their base image in the history entry for the first instruction that was processed, rather than the last.
```

### DIFF
--- a/image.go
+++ b/image.go
@@ -721,18 +721,18 @@ func (i *containerImageRef) NewImageSource(ctx context.Context, sc *types.System
 			Created:    &created,
 			CreatedBy:  i.createdBy,
 			Author:     oimage.Author,
-			Comment:    comment,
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@giuseppe @flouthoc @TomSweeneyRedHat @mheon PTAL

flouthoc (approved):
LGTM

PR #5596, opened 2024-06-18
commit: set "parent" for docker format only when requested

nalind:
#### What type of PR is this?

/kind api-change

Make setting the Parent field in the config blob of a docker format image optional (yes, we're bringing it back!), since it no longer appears to be set by newer versions of docker build.

#### How to verify it

Updated conformance tests!

None

For the moment I expect this to be more useful for API consumers, but it could be exposed at the command line if we needed it to be.

```release-note
None
```


rhatdan:
LGTM
@flouthoc @TomSweeneyRedHat @giuseppe @mheon PTAL

flouthoc (approved):
LGTM

PR #5599, opened 2024-06-20
imagebuildah.StageExecutor.prepare(): log the --platform flag

nalind:
#### What type of PR is this?

/kind other

If FROM was used with a --platform flag, then the imagebuilder.Builder will have its Platform field set, and we should include it when logging the instruction.

#### How to verify it

New integration test!

None

Prompted by https://github.com/containers/podman/issues/23046.

```release-note
None
```


flouthoc (approved):
LGTM

PR #5600, opened 2024-06-20
CI VMs: bump

edsantiago:
Built in https://github.com/containers/automation_images/pull/361

New docker (26.1.3 -> .4). Seems low-risk aside from that.

```release-note
None
```


edsantiago:

PR #5605, opened 2024-06-21
Rework parsing of --pull flags

### LINKED ISSUE #5602: Pull policy `newer` is unsupported by `buildah` (documentation notwithstanding)
**Description**
The [documentation](https://github.com/containers/buildah/blob/main/docs/buildah-pull.1.md) claims the existence of a `newer` pull policy. However, `buildah` does not actually support a `newer` pull policy. The closest policy I can find is called `ifnewer`, but this policy seems to be undocumented, and based on [this discussion](https://github.com/containers/buildah/pull/5524#issuecomment-2181569626) there may not be appetite to add documentation for it.

`podman run` supports both `newer` and the undocumented `ifnewer` flags, as in `podman run --pull=newer`.

I believe the ideal state of the world would be for `buildah` to simply add support for `newer` such that it works like `podman` and so that the documentation would (at last) be correct.

1. `buildah pull --policy=newer quay.io/fedora/fedora:40`

`Error: unsupported pull policy "newer"`

Expected: Pulls if the image on the registry has a different digest than the one in the local containers storage.

```
buildah-1.35.4-1.fc40.x86_64
```

```
buildah version 1.35.4 (image-spec 1.1.0, runtime-spec 1.1.0)
```

```
Fedora release 40 (Forty)
NAME="Fedora Linux"
VERSION="40 (Container Image)"
ID=fedora
VERSION_ID=40
VERSION_CODENAME=""
PLATFORM_ID="platform:f40"
PRETTY_NAME="Fedora Linux 40 (Container Image)"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:40"
DEFAULT_HOSTNAME="fedora"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f40/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
REDHAT_BUGZILLA_PRODUCT_VERSION=40
REDHAT_SUPPORT_PRODUCT_VERSION=40
SUPPORT_END=2025-05-13
VARIANT="Container Image"
VARIANT_ID=container
Fedora release 40 (Forty)
Fedora release 40 (Forty)
```

**Output of `uname -a`:**

```
Linux 3e4d7a3fdedb 6.8.0-35-generic #35-Ubuntu SMP PREEMPT_DYNAMIC Mon May 20 15:51:52 UTC 2024 x86_64 GNU/Linux
```

```
# files.
#
[storage]

driver = "overlay"

# Temporary storage location

# following commands:

# imagestore = ""


#

# transient_store = true

[storage.options]

additionalimagestores = [
"/var/lib/shared",
]

#   * ostree_repos = ""

# container.
#

# Remap-UIDs/GIDs setting.
#
# remap-user = "containers"
# remap-group = "containers"

#
# auto-userns-min-size=1024
#
# auto-userns-max-size=65536

[storage.options.overlay]
# and vfs drivers.

# inodes = ""

# directly.
mount_program = "/usr/bin/fuse-overlayfs"

mountopt = "nodev,fsync=0"

# skip_mount_home = "false"

# use_composefs = "false"

# size = ""

# directories.
#
#
#  "": No value specified.
#     image.
#     any user.
#
#
#  "force_mask"  permissions.
#
# force_mask = ""
```

nalind:
#### What type of PR is this?

/kind bug

Rework parsing of --pull flags to add "newer" as an alias for "ifnewer", and to reject unrecognized values instead of treating them all as synonymous with "ifmissing".

#### How to verify it

New unit test!

Fixes #5602

```release-note
Unrecognized values for the `--pull` command line flag are no longer silently treated as if they were "missing".
```

### DIFF
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -449,6 +449,42 @@ func SystemContextFromFlagSet(flags *pflag.FlagSet, findFlagFunc func(name strin
 	return ctx, nil
 }
 
+// pullPolicyWithFlags parses a string value of a pull policy, evaluating it in
+// combination with "always" and "never" boolean flags.
+// Allow for:
+// * --pull
+// * --pull=""
+// * --pull=true
+// * --pull=false
+// * --pull=never
+// * --pull=always
+// * --pull=ifmissing
+// * --pull=missing
+// * --pull=notpresent
+// * --pull=newer
+// * --pull=ifnewer
+// and --pull-always and --pull-never as boolean flags.
+func pullPolicyWithFlags(policySpec string, always, never bool) (define.PullPolicy, error) {
+	if always {
+		return define.PullAlways, nil
+	}
+	if never {
+		return define.PullNever, nil
+	}
+	policy := strings.ToLower(policySpec)
+	switch policy {
+	case "true", "missing", "ifmissing", "notpresent":
+		return define.PullIfMissing, nil
+	case "always":
+		return define.PullAlways, nil
+	case "false", "never":
+		return define.PullNever, nil
+	case "ifnewer", "newer":
+		return define.PullIfNewer, nil
+	}
+	return 0, fmt.Errorf("unrecognized pull policy %q", policySpec)
+}
+
 // PullPolicyFromOptions returns a PullPolicy that reflects the combination of
 // the specified "pull" and undocumented "pull-always" and "pull-never" flags.
 func PullPolicyFromOptions(c *cobra.Command) (define.PullPolicy, error) {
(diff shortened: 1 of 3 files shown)


pkg/parse/parse.go
```diff
@@ -474,30 +510,23 @@ func PullPolicyFromFlagSet(flags *pflag.FlagSet, findFlagFunc func(name string)
...
-	if pullNeverFlagValue ||
-		strings.EqualFold(pullFlagValue, "never") ||
-		strings.EqualFold(pullFlagValue, "false") {
-		pullPolicy = define.PullNever
+
+	// The --pull[=...] flag is the one we really care about.
+	pullFlagValue := findFlagFunc("pull").Value.String()
+	pullPolicy, err := pullPolicyWithFlags(pullFlagValue, pullAlwaysFlagValue, pullNeverFlagValue)
+	if err != nil {
+		return 0, fmt.Errorf("parsing --pull value %q: %w", pullFlagValue, err)
```
rhatdan, line 527: I think the error returned by pullPolicyWIthFlags is adequate, and you just created a stutter.

nalind: Okay, removing it.

flouthoc (approved):
LGTM
/lgtm
/approve

PR #5629, opened 2024-07-15
Revert #5627

nalind:
#### What type of PR is this?

/kind other

We just updated from a development version of github.com/containers/image/v5 to a version that was cut from its stable release branch.  Podman's development branch already depends on features which are newer than the image library's stable branch, and vendoring _our_ development branch was breaking compilation over there.

#### How to verify it

https://github.com/containers/podman/pull/13808 should be... well, not _happy_, but closer to happy.

```release-note
None
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/containerd/containerd v1.7.18
 	github.com/containernetworking/cni v1.2.2
 	github.com/containers/common v0.59.1-0.20240712101718-237a317152ae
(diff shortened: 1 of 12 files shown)


edsantiago:
Buildah got vendored into podman today, so there probably isn't a big need for the treadmill to be stable. I was just trying to be proactive.

nalind:
podman might not need it, but `main` in this repository downgraded one of its dependencies unintentionally, and that should matter.

PR #5632, opened 2024-07-16
 [release-1.35] Fixes Listing tags in JFrog Artifactory may fail

TomSweeneyRedHat:
Addresses the problem first described in https://github.com/containers/skopeo/issues/2346

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 # Changelog
 
+## v1.35.5 (2024-07-16)
+
+    [release-1.35] Fixes Listing tags in JFrog Artifactory may fail
+    Cross-build on Fedora
+
 ## v1.35.4 (2024-05-09)
 
     [release-1.35] CVE-2024-3727 updates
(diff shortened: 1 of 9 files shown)


TomSweeneyRedHat:
@lsm5 we have 1.35 only in Fedora, but now we also have 1.36.  Is it necessary to create a new 1.35 too?  This is not on RHEL.  If not necessary for Fedora, I'll probably just close this.

rhatdan:
Don't back port anything that is not going to RHEL.  If there is a CVE for F39, should be the only non RHEL backports.

lsm5:
> @lsm5 we have 1.35 only in Fedora, but now we also have 1.36. Is it necessary to create a new 1.35 too? This is not on RHEL. If not necessary for Fedora, I'll probably just close this.

Fedora has [1.36 now](https://src.fedoraproject.org/rpms/buildah). Closing..

TomSweeneyRedHat:
Thanks folks.  That makes the backporting life easier.  I'll only backport fixes for whatever is called for in RHEL, and then the latest/greatest on Fedora.

PR #5668, opened 2024-08-06
Drop the e2e test suite

nalind:
#### What type of PR is this?

/kind cleanup

Drop the e2e suite, which largely duplicates other integration tests, and add an integration test to cover the "output from `inspect` is valid JSON" case.

#### How to verify it

The rest of the tests should pass, and CI shouldn't continue trying to run them now that they're gone.

```release-note
None
```


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

PR #5673, opened 2024-08-07
conformance tests: use mirror.gcr.io for most images

nalind:
#### What type of PR is this?

/kind failing-test 

Use busybox and alpine images from mirror.gcr.io, where possible, to avoid tripping pull limits in CI.

#### How to verify it

So many updated conformance tests!

None

```release-note
None
```

### DIFF
--- a/tests/conformance/README.md
+++ b/tests/conformance/README.md
@@ -15,15 +15,18 @@ Conformance tests use Docker CE to build images to be compared with images built
 
 ## Run conformance tests
 
-First, pull base images used by various conformance tests:
+These are the base images used by various conformance tests:
 ```
 bash
-docker pull alpine
-docker pull busybox
+docker pull mirror.gcr.io/alpine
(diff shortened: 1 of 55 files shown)


TomSweeneyRedHat:
LGTM
@edsantiago if you can give this a quick looksee, I'd appreciate it.

edsantiago:
I'm confused. There's already a `registries.conf` that fixes this. Is it not fixing it? What problem are we seeing?

edsantiago:
hold on, I merged local-registry in #5584. We should absolutely not be using any network pulls any more! What's happening?

nalind:
Conformance tests use `docker build` with the docker daemon and friends as the baseline, and they don't pay attention to registries.conf.  In order to use the same base images, the code paths that use buildah can't either.

nalind:
Well, not `docker build` per se, but the REST API that the docker engine provides.

PR #5676, opened 2024-08-08
Add(): re-escape any globbed items that included escapes

### LINKED ISSUE #5675: Cannot escape square brackets in build of Dockerfile with COPY instruction
**Description**

I am trying to build a nextjs app Dockerfile.  The Dockerfile contains a COPY instruction with a source path that contains square brackets.  I quickly found the escape sequence on the Dockerfile reference docs: https://docs.docker.com/reference/dockerfile/#copy, and from the docs:

"When adding files or directories that contain special characters (such as [ and ]), you need to escape those paths following the Golang rules to prevent them from being treated as a matching pattern. For example, to add a file named arr[0].txt, use the following;"

Applying the escape sequence allowed me to build the container image using the docker runtime, however I was not able to do the same using the buildah runtime.

1. In a fresh directory, create an empty file at the path app/[xyz]/file.  Screenshot of my directory:
image

2. Create a Dockerfile with the following contents:
```
FROM node
WORKDIR /app
COPY ./app/[[]xyz]/file ./app/[xyz]/file
```

Screenshot of Dockerfile:
image

3. run `buildah build -t my-image .`

Output of step 3 command:
```
STEP 1/3: FROM node
STEP 2/3: WORKDIR /app
STEP 3/3: COPY app/[[]xyz]/file /app/[xyz]file
Error: building at STEP "COPY app/[[]xyz]/file /app/[xyz]file": reading "/app/app/[[]xyz]/file": copier: get: globs [/app/[xyz]/file] matched nothing (0 filtered out): no such file or directory
```

Expected container image to build successfully.

```
buildah-1.36.0-1.fc40.aarch64
```

```
Version:         1.36.0
Go Version:      go1.22.3
Image Spec:      1.1.0
Runtime Spec:    1.2.0
CNI Spec:        1.0.0
libcni Version:  
image Version:   5.31.0
Git Commit:      
Built:           Mon May 27 13:11:32 2024
OS/Arch:         linux/arm64
```

```
N/A
```

```
Fedora release 40 (Forty)
NAME="Fedora Linux"
VERSION="40 (Container Image)"
ID=fedora
VERSION_ID=40
VERSION_CODENAME=""
PLATFORM_ID="platform:f40"
PRETTY_NAME="Fedora Linux 40 (Container Image)"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:40"
DEFAULT_HOSTNAME="fedora"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f40/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
REDHAT_BUGZILLA_PRODUCT_VERSION=40
REDHAT_SUPPORT_PRODUCT_VERSION=40
SUPPORT_END=2025-05-13
VARIANT="Container Image"
VARIANT_ID=container
Fedora release 40 (Forty)
Fedora release 40 (Forty)
```

**Output of `uname -a`:**

```
Linux 2a09f66d6d28 6.8.0-31-generic #31-Ubuntu SMP PREEMPT_DYNAMIC Sat Apr 20 02:32:42 UTC 2024 aarch64 GNU/Linux
```

```
# files.
#
[storage]

driver = "overlay"

# Temporary storage location

# following commands:

# imagestore = ""


#

# transient_store = true

[storage.options]

additionalimagestores = [
"/var/lib/shared",
]

#   * ostree_repos = ""

# container.
#

# Remap-UIDs/GIDs setting.
#
# remap-user = "containers"
# remap-group = "containers"

#
# auto-userns-min-size=1024
#
# auto-userns-max-size=65536

[storage.options.overlay]
# and vfs drivers.

# inodes = ""

# directly.
mount_program = "/usr/bin/fuse-overlayfs"

mountopt = "nodev,fsync=0"

# skip_mount_home = "false"

# use_composefs = "false"

# size = ""

# directories.
#
#
#  "": No value specified.
#     image.
#     any user.
#
#
#  "force_mask"  permissions.
#
# force_mask = ""
```

nalind:
#### What type of PR is this?

/kind bug

When collecting items we matched using the passed-in globs during Add(), be sure that globbing performed by Get() will match the items exactly, by escaping special matching characters.

#### How to verify it

New conformance test!

Fixes #5675.

```release-note
COPY and ADD instructions being processed by `buildah build` will now be handled correctly if items matched by source location globs have names which include characters that are treated as special when globbing.
```


flouthoc (commented):
Code LGTM maybe we can extend test to cover `?` matching as well.

nalind:
It looks like attempting to escape `\` and `*` produces different results in classic docker build compared to BuildKit, so for now I'll aim for the way BuildKit does them.

flouthoc (approved):
LGTM

Issue #5674, opened 2024-08-08
race: parallel builds: copying...committing...creating... layer not known


edsantiago:
This _might_ be the same as https://github.com/containers/podman/issues/23331 . If it is, someone please close this or move.

Setup:
```console
$ for i in 1 2;do printf "FROM quay.io/libpod/testimage:20240123\nRUN echo hi from $i\n" >Containerfile$i;done
```

In window 1:
```console
$ while :;do buildah build -t c1 --layers=true -f Containerfile1 || break;buildah rmi c1;done
```
In window 2:
```console
$ while :;do buildah build --layers=false -t c2 -f Containerfile2|| break;buildah rmi c2;done
```
Within 30-60s, window 1 will barf:
```
STEP 1/2: FROM quay.io/libpod/testimage:20240123
STEP 2/2: RUN echo hi from 1
Error: checking if cached image exists from a previous build: getting top layer info: layer not known
```
or
```
STEP 1/2: FROM quay.io/libpod/testimage:20240123
STEP 2/2: RUN echo hi from 1
hi from 1
COMMIT c1
Error: committing container for step {Env:[PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin] Command:run Args:[echo hi from 1] Flags:[] Attrs:map[] Message:RUN echo hi from 1 Heredocs:[] Original:RUN echo hi from 1}: copying layers and metadata for container "a8d0253ccd5f337ca69e106657dc645e4926b20d9775621827b6ec118bcb35fa": committing the finished image: creating image "41778f8cf15b69d1fdb79d5bb744ba65eac877e27a21dd12af8700594d88585b": layer not known
```
The `rmi` seems important; I can't get it to fail (at least not within my patience tolerance of ~10m) if I omit `rmi` from either loop.

Testing with `podman` fails MUCH faster than buildah, for reasons I don't understand, and also fails sometimes in window 2. Buildah only fails in window 1.

This is blocking parallelization of podman test `070-build` and I bet this is one of the uncategorized weirdnesses I've seen in #5552 but didn't follow up on.

edsantiago:
Issue [persists](https://api.cirrus-ci.com/v1/artifact/task/5114275818110976/html/sys-podman-rawhide-root-host-sqlite.log.html#t--00156p):
```
<+0042s> # # podman build -t b-t156-muinxj0h /tmp/CI_dBI1/podman_bats.20lh4r/build-test
[...]

PR #5686, opened 2024-08-14
Commit(): retry committing to local storage on storage.LayerUnknown

nalind:
#### What type of PR is this?

/kind other

If we get a `storage.LayerUnknown` error while committing to local storage, guess that we tried to reuse a layer in an image that got deleted in the brief window between when the commit logic decided that it could reuse the layer, and when it actually tried to create the image record.

It's tempting to try to resolve this in the image library, but at the point where this error crops up, it may already have skipped pulling the relevant layer blob, so the entire copy operation would have to be restarted from the beginning anyway, and that's the retry package's job.

It's tempting to bake this into the common library's retry package, but that would require it to "know" about containers/storage, and while in practice most of its callers already depend on the storage library, that may a bit much.

#### How to verify it

New unit test!

Works around #5674.

```release-note
None
```

### DIFF
--- a/common.go
+++ b/common.go
@@ -2,6 +2,7 @@ package buildah
 
 import (
 	"context"
+	"errors"
 	"io"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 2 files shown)


flouthoc (approved):
LGTM

PR #5688, opened 2024-08-15
copier: handle globbing with "**" path components

nalind:
#### What type of PR is this?

/kind feature

Handle glob patterns with "\*\*" path components by expanding "\*\*" to the set of subdirectories and calling filepath.Glob() on the results.

#### How to verify it

New unit test!
New conformance tests!

None

```release-note
`buildah build` now expands "**" path components in source locations in ADD and COPY instructions as referring to any subdirectory.
```


flouthoc (approved):
LGTM

copier/copier.go
```diff
@@ -41,6 +41,74 @@ func init() {
 	reexec.Register(copierCommand, copierMain)
 }
 
+// extendedGlob calls filepath.Glob() on the passed-in patterns.  If there is a
+// "**" component in the pattern, filepath.Glob() will be called with the "**"
+// replaced with all of the subdirectories under that point, and the results
+// will be concatenated.
+func extendedGlob(pattern string) (matches []string, err error) {
+	subdirectories := func(dir string) []string {
+		var subdirectories []string
```
rhatdan, line 50: This is difficult to read, since the function and the variable have the same name.  Would have been better if they were different.

copier/copier.go
```diff
@@ -41,6 +41,74 @@ func init() {
...
+		return patterns
+	}
+	patterns := []string{pattern}
+	i := 0
+	for i < len(patterns) {
+		expanded := expandPatterns(patterns[i])
+		if len(expanded) > 1 {
+			patterns = append(append(patterns[:i], expanded...), patterns[i+1:]...)
+		} else {
+			i++
```
rhatdan, line 99: Why not use `break`?

nalind: We don't know that `patterns[i+1]` (if we're not at the end of the slice) doesn't need to be expanded.

Issue #5400, opened 2024-03-13
buildah never invalidates the cache of a RUN --mount=type=bind layer


wegel:
**Description**

buildah never invalidates the cache of a `RUN --mount=type=bind` or `RUN --mount=type=cache` when the contents of the source changes.

**Steps to reproduce the issue:**

```
mkdir files
for M in bind cache; do
for I in 1 2; do
echo "$M: Iteration $I"
echo "$M: $I" > files/$M$I
buildah build --layers -f -<<EOF
FROM ubuntu:22.04
RUN --mount=type=$M,source=files,target=/files ls -l /files
EOF
done
done
```

**Describe the results you received:**

```
bind: Iteration 1
STEP 1/2: FROM ubuntu:22.04
STEP 2/2: RUN --mount=type=bind,source=files,target=/files ls -l /files
total 4
-rw-r--r-- 1 root root 8 Mar 13 01:52 bind1
COMMIT
--> b325129d4a91
b325129d4a91df893b6d698c1db8f89c218173d89516d627977bf53b0158828c

bind: Iteration 2
STEP 1/2: FROM ubuntu:22.04
STEP 2/2: RUN --mount=type=bind,source=files,target=/files ls -l /files
--> Using cache b325129d4a91df893b6d698c1db8f89c218173d89516d627977bf53b0158828c
--> b325129d4a91
b325129d4a91df893b6d698c1db8f89c218173d89516d627977bf53b0158828c

cache: Iteration 1
STEP 1/2: FROM ubuntu:22.04
STEP 2/2: RUN --mount=type=cache,source=files,target=/files ls -l /files
total 0
COMMIT
--> e074a87a9cc3
e074a87a9cc3ef55550cd83fc0cef080478797c382deec45c7b63bd8825a2367
cache: Iteration 2

STEP 1/2: FROM ubuntu:22.04
STEP 2/2: RUN --mount=type=cache,source=files,target=/files ls -l /files
--> Using cache e074a87a9cc3ef55550cd83fc0cef080478797c382deec45c7b63bd8825a2367
--> e074a87a9cc3
e074a87a9cc3ef55550cd83fc0cef080478797c382deec45c7b63bd8825a2367
```

**Describe the results you expected:**

I expected the image to be rebuilt during the second iteration, since the contents of `files` had changed.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
$ pacman -Q | grep buildah
buildah 1.34.1-1
```

**Output of `buildah version`:**

```
[...]

PR #5691, opened 2024-08-18
build, run: record hash or digest in image history for sources used in `--mount`

flouthoc:
When using `--mount=type=bind` or `--mount=type=cache` the hash or digest of source in these flags should be added to image history so buildah can burst cache if files on host or image which is being used as source is changed.

Closes
* https://github.com/containers/podman/issues/23382
* https://github.com/containers/buildah/issues/5400

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
build, run: record hash or digest in image history for sources used in `--mount`
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -1279,7 +1279,11 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			// No base image means there's nothing to put in a
 			// layer, so don't create one.
 			emptyLayer := (s.builder.FromImageID == "")
-			if imgID, ref, err = s.commit(ctx, s.getCreatedBy(nil, ""), emptyLayer, s.output, s.executor.squash || s.executor.confidentialWorkload.Convert, lastStage); err != nil {
+			createdBy, err := s.getCreatedBy(nil, "")
+			if err != nil {
+				return "", nil, false, fmt.Errorf("unable to get createdBy for the node: %w", err)
+			}
+			if imgID, ref, err = s.commit(ctx, createdBy, emptyLayer, s.output, s.executor.squash || s.executor.confidentialWorkload.Convert, lastStage); err != nil {
 				return "", nil, false, fmt.Errorf("committing base container: %w", err)
 			}
 		} else {
@@ -1426,7 +1430,11 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 				if s.executor.timestamp != nil {
 					timestamp = *s.executor.timestamp
 				}
-				s.builder.AddPrependedEmptyLayer(&timestamp, s.getCreatedBy(node, addedContentSummary), "", "")
+				createdBy, err := s.getCreatedBy(node, addedContentSummary)
+				if err != nil {
+					return "", nil, false, fmt.Errorf("unable to get createdBy for the node: %w", err)
+				}
+				s.builder.AddPrependedEmptyLayer(&timestamp, createdBy, "", "")
 				continue
 			}
 			// This is the last instruction for this stage,
@@ -1436,7 +1444,11 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			// stage.
 			if lastStage || imageIsUsedLater {
 				logCommit(s.output, i)
-				imgID, ref, err = s.commit(ctx, s.getCreatedBy(node, addedContentSummary), false, s.output, s.executor.squash, lastStage && lastInstruction)
+				createdBy, err := s.getCreatedBy(node, addedContentSummary)
+				if err != nil {
+					return "", nil, false, fmt.Errorf("unable to get createdBy for the node: %w", err)
+				}
+				imgID, ref, err = s.commit(ctx, createdBy, false, s.output, s.executor.squash, lastStage && lastInstruction)
 				if err != nil {
 					return "", nil, false, fmt.Errorf("committing container for step %+v: %w", *step, err)
 				}
@@ -1657,6 +1669,10 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			// We're not going to find any more cache hits, so we
 			// can stop looking for them.
 			checkForLayers = false
+			createdBy, err := s.getCreatedBy(node, addedContentSummary)
+			if err != nil {
+				return "", nil, false, fmt.Errorf("unable to get createdBy for the node: %w", err)
+			}
 			// Create a new image, maybe with a new layer, with the
 			// name for this stage if it's the last instruction.
 			logCommit(s.output, i)
@@ -1664,7 +1680,7 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			// because at this point we want to save history for
 			// layers even if its a squashed build so that they
 			// can be part of the build cache.
-			imgID, ref, err = s.commit(ctx, s.getCreatedBy(node, addedContentSummary), !s.stepRequiresLayer(step), commitName, false, lastStage && lastInstruction)
+			imgID, ref, err = s.commit(ctx, createdBy, !s.stepRequiresLayer(step), commitName, false, lastStage && lastInstruction)
 			if err != nil {
 				return "", nil, false, fmt.Errorf("committing container for step %+v: %w", *step, err)
 			}
@@ -1695,12 +1711,16 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 
 		if lastInstruction && lastStage {
 			if s.executor.squash || s.executor.confidentialWorkload.Convert || len(s.executor.sbomScanOptions) != 0 {
+				createdBy, err := s.getCreatedBy(node, addedContentSummary)
+				if err != nil {
+					return "", nil, false, fmt.Errorf("unable to get createdBy for the node: %w", err)
+				}
 				// If this is the last instruction of the last stage,
 				// create a squashed or confidential workload
 				// version of the image if that's what we're after,
 				// or a normal one if we need to scan the image while
 				// committing it.
-				imgID, ref, err = s.commit(ctx, s.getCreatedBy(node, addedContentSummary), !s.stepRequiresLayer(step), commitName, s.executor.squash || s.executor.confidentialWorkload.Convert, lastStage && lastInstruction)
+				imgID, ref, err = s.commit(ctx, createdBy, !s.stepRequiresLayer(step), commitName, s.executor.squash || s.executor.confidentialWorkload.Convert, lastStage && lastInstruction)
 				if err != nil {
 					return "", nil, false, fmt.Errorf("committing final squash step %+v: %w", *step, err)
 				}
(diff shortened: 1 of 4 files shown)


flouthoc:
Need to add tests, will undraft then.

flouthoc:
Will rebase after this https://github.com/containers/buildah/pull/5693

nalind:
I'm surprised that we'd care about the contents of caches.  I'd be inclined to archive the contents of a directory (and create a single-entry archive for a non-directory) to account for different permissions/ownership/datestamps/xattrs and to safely handle soft and hard links.

flouthoc:
> I'm surprised that we'd care about the contents of caches. I'd be inclined to archive the contents of a directory (and create a single-entry archive for a non-directory)

If i'm understanding this correctly, did you mean instead of `computeDirectoryHash` or `computeFileHash` separately and recursively I should just create a temporary archive of the context dir and write its hash in history ? And on everyrun I can do the same process again to check if hash is matching with what is written in the `history` ?  The idea sounds good to me.

nalind:
> > I'm surprised that we'd care about the contents of caches. I'd be inclined to archive the contents of a directory (and create a single-entry archive for a non-directory)
> 
> If i'm understanding this correctly, did you mean instead of `computeDirectoryHash` or `computeFileHash` separately and recursively I should just create a temporary archive of the context dir and write its hash in history?

I wouldn't expect the archive to be written anywhere, but the digest of an archive is something we already use as a way of describing contents, when handling COPY and ADD instructions.  I don't know yet about doing this over the entire build context or additional build context if only a portion of it is being used at that point (i.e., if "src" is set to a subdirectory).

flouthoc:
> > > I'm surprised that we'd care about the contents of caches. I'd be inclined to archive the contents of a directory (and create a single-entry archive for a non-directory)
> > 
> > 
> > If i'm understanding this correctly, did you mean instead of `computeDirectoryHash` or `computeFileHash` separately and recursively I should just create a temporary archive of the context dir and write its hash in history?
> 
> I wouldn't expect the archive to be written anywhere, but the digest of an archive is something we already use as a way of describing contents, when handling COPY and ADD instructions. I don't know yet about doing this over the entire build context or additional build context if only a portion of it is being used at that point (i.e., if "src" is set to a subdirectory).

This sounds good to me, i will amend the PR.

sanmai-NL:
@flouthoc

I'm seeing this behavior with `podman build`, will this be fixed by this change too?

```Containerfile
ARG PATH_1=mydirectory
ARG SELINUXRELABEL=,z
ARG DISTRO=PATH_1
RUN --mount=type=bind,source=${PATH_1:?},target=/tmp/${PATH_1:?}${SELINUXRELABEL:?} \
  echo Nop
```

Subsequently, `/tmp/mydirectory` ends up in an image layer rather than being ephemeral and only mounted at build time. This bloats the image and leaks information.

flouthoc:
> ```dockerfile
> RUN --mount=type=bind,source=${PATH_1:?},target=/tmp/${PATH_1:?}${SELINUXRELABEL:?}
> ```

@sanmai-NL Are contents of `PATH_1` ending up in your final built image as well ? A layer is created so that i can be used later but I don't think contents of `RUN --mount` are part of your image.

sanmai-NL:
Thanks for your response. Yes indeed, they do end up in there, which I find surprising. And which is the reason for my comment.

Some factors which may cause this in case the problem isn't general. This happens when running `podman build` from a containerized Podman. Maybe ID mapping also causes some sort if copy operation instead of a bind mount.

sanmai-NL:
> > ```dockerfile
> 
> > RUN --mount=type=bind,source=${PATH_1:?},target=/tmp/${PATH_1:?}${SELINUXRELABEL:?}
> 
> > ```
> 
> 
> 
> @sanmai-NL Are contents of `PATH_1` ending up in your final built image as well ? A layer is created so that i can be used later but I don't think contents of `RUN --mount` are part of your image.

Re-reading your comment... So you do find this ending up in a layer by-design, do you? But this way, information leaks and the image bloats.

flouthoc:
> > > 
> > 
> > 
> > > RUN --mount=type=bind,source=${PATH_1:?},target=/tmp/${PATH_1:?}${SELINUXRELABEL:?}
> > 
> > 
> > > 
> > 
> > 
> > @sanmai-NL Are contents of `PATH_1` ending up in your final built image as well ? A layer is created so that i can be used later but I don't think contents of `RUN --mount` are part of your image.
> 
> Re-reading your comment... So you do find this ending up in a layer by-design, do you? But this way, information leaks and the image bloats.

@sanmai-NL What you are describing is a different bug, would it be possible to create a small reproducer and open a new issue ?

flouthoc:
Since https://github.com/containers/buildah/pull/5693 is merged, i'll get back to this.

flouthoc:

internal/parse/parse.go
```diff
@@ -10,6 +10,24 @@ import (
...
+	source := ""
+	from := ""
+	for _, option := range tokens {
+		optionSplit := strings.Split(option, "=")
+		if optionSplit[0] == "src" || optionSplit[0] == "source" {
+			source = optionSplit[1]
+		}
+		if optionSplit[0] == "from" {
+			from = optionSplit[1]
+		}
```
nalind, line 26: There's no check here that `option` includes a "=" character, so it's possible there's a crash here.

internal/util/util.go
```diff
@@ -14,9 +15,52 @@ import (
...
 	"github.com/containers/storage/pkg/unshare"
+	digest "github.com/opencontainers/go-digest"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
+// GeneratePathChecksum generates the SHA-256 checksum for a file or a directory.
+func GeneratePathChecksum(sourcePath string) (string, error) {
+	digester := digest.SHA256.Digester()
+	tarWriter := tar.NewWriter(digester.Hash())
+	defer tarWriter.Close()
```
nalind, line 26: In the general case, the tar writer would need to be closed before the digest is read back, so that the digest calculation wouldn't miss the end-of-archive marker.

internal/util/util.go
```diff
@@ -14,9 +15,52 @@ import (
...
+		if err != nil {
+			return err
+		}
+
+		header, err := tar.FileInfoHeader(info, info.Name())
+		if err != nil {
+			return err
+		}
+
+		header.Name = filepath.ToSlash(path)
```
nalind, line 38: I think that this is going to need to be computed to be relative to the top-level directory.

flouthoc:
@nalind Could you PTAL again.

imagebuildah/stage_executor.go
```diff
@@ -1280,7 +1281,11 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			// No base image means there's nothing to put in a
 			// layer, so don't create one.
 			emptyLayer := (s.builder.FromImageID == "")
-			if imgID, ref, err = s.commit(ctx, s.getCreatedBy(nil, ""), emptyLayer, s.output, s.executor.squash || s.executor.confidentialWorkload.Convert, lastStage); err != nil {
+			createdBy, err := s.getCreatedBy(nil, "")
+			if err != nil {
+				return "", nil, false, err
```
nalind, line 1286: Get in the habit of adding context to errors being passed up from calls to methods that returned them, to make it easier to follow the call stack when troubleshooting and debugging.  Not just here, but throughout this PR.

imagebuildah/stage_executor.go
```diff
@@ -1850,15 +1875,66 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
...
+						}
+					} else {
+						// Found additional build context, get directory sha.
+						basePath := s.executor.additionalBuildContexts[mountOptionFrom].Value
+						if s.executor.additionalBuildContexts[mountOptionFrom].IsURL {
+							basePath = s.executor.additionalBuildContexts[mountOptionFrom].DownloadedCache
+						}
+						appendCheckSum, err = internalUtil.GeneratePathChecksum(filepath.Join(basePath, mountOptionSource))
+						if err != nil {
+							return "", fmt.Errorf("Unable to generate checksum for image history: %w", err)
```
nalind, line 1916: Text passed to fmt.Errorf() is context, not a full error message, so it should not be capitalized, and should merely describe what was being attempted, so that the caller can add more context if it wishes.  "generating checksum for directory %q in %q: %w" would be enough here, since the caller will prepend context of its own.  Not just here, but throughout this PR.

internal/util/util.go
```diff
@@ -14,9 +15,57 @@ import (
...
+func GeneratePathChecksum(sourcePath string) (string, error) {
+	digester := digest.SHA256.Digester()
+	tarWriter := tar.NewWriter(digester.Hash())
+
+	err := filepath.Walk(sourcePath, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		header, err := tar.FileInfoHeader(info, info.Name())
```
nalind, line 32: The second argument being passed to `FileInfoHeader()` is not correct.

imagebuildah/stage_executor.go
```diff
@@ -1850,15 +1875,66 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
...
+		var err error
+		for _, flag := range node.Flags {
+			if strings.HasPrefix(flag, "--mount") {
+				mountOptionSource, mountOptionFrom = internalParse.GetFromAndSourceKeysFromMountFlag(flag)
+				// If source is not specified then default is '.'
+				if mountOptionSource == "" {
+					mountOptionSource = "."
+				}
+			}
+		}
```
nalind, line 1894: This looks like it will only consider the last `--mount` flag, and ignore the rest if there are more than one.

internal/util/util.go
```diff
@@ -14,9 +15,57 @@ import (
 	"github.com/containers/storage/pkg/archive"
 	"github.com/containers/storage/pkg/chrootarchive"
 	"github.com/containers/storage/pkg/unshare"
+	digest "github.com/opencontainers/go-digest"
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
+// GeneratePathChecksum generates the SHA-256 checksum for a file or a directory.
+func GeneratePathChecksum(sourcePath string) (string, error) {
```
nalind, line 23: This can be a helper function in the package that calls it until we find that we actually need to call it from somewhere else as well.  It might benefit from a unit test.

internal/parse/parse.go
```diff
@@ -10,6 +10,26 @@ import (
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 )
 
+// Consumes mount flag in format of `--mount=type=bind,src=/path,from=image` and
+// return values of `src` and `from`, values are empty if keys are not present in the option.
+func GetFromAndSourceKeysFromMountFlag(mount string) (string, string) {
```
nalind, line 15: This can be a helper function in the package that calls it until we find that we actually need to call it from somewhere else as well.

internal/util/util.go
```diff
@@ -14,9 +15,57 @@ import (
...
+
+		_, err = io.Copy(tarWriter, file)
+		return err
+	})
+	if err != nil {
+		tarWriter.Close()
+		return "", err
+	}
+
+	tarWriter.Close()
```
nalind, line 65: Move this to just above the `if err != nil` check, if its return value isn't going to be checked.

nalind (changes requested):
Be _very_ sure you want to use sha256 over an entire directory tree.  It can be an expensive computation if the tree is large.

sanmai-NL:
https://github.com/BLAKE3-team/BLAKE3

nalind:
In its current form, this considers caches where a "src" option is used to select a subdirectory, but not otherwise.  It should either check the mount type and skip caches, or make arrangements to handle those cases.

imagebuildah/util.go
```diff
@@ -1,9 +1,92 @@
...
+		return err
+	})
+	tarWriter.Close()
+	if err != nil {
+		tarWriter.Close()
+		return "", err
+	}
+	return digester.Digest().String(), nil
+}
+
```
nalind, line 89: The new types and functions don't need to be exported.

imagebuildah/util.go
```diff
@@ -1,9 +1,92 @@
...
+			return err
+		}
+		defer file.Close()
+
+		_, err = io.Copy(tarWriter, file)
+		return err
+	})
+	tarWriter.Close()
+	if err != nil {
+		tarWriter.Close()
```
nalind, line 84: Duplicate call.
```suggestion
```

imagebuildah/util_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"io"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/opencontainers/go-digest"
+)
+
+func TestGeneratePathChecksum(t *testing.T) {
+	tempDir, err := os.MkdirTemp("", "testdir")
```
nalind, line 14: Use a value returned by `t.TempDir()`.  It was introduced in Go 1.15, and you don't need to clean it up.

tests/bud.bats
```diff
@@ -354,6 +354,148 @@ _EOF
...
+  run_buildah build $WITH_POLICY_JSON --layers -t source -f $contextdir/Containerfile $contextdir
+  # output should not contain `samplefile`
+  assert "$output" !~ "samplefile"
+
+  # Modify sample file 2
+  cat > $contextdir/testfile << _EOF
+Helloworld2
+_EOF
+
+  # on third run since we have added new file `anotherfile` so cache must burst.
```
nalind, line 423: Update this comment.

imagebuildah/util.go
```diff
@@ -1,9 +1,92 @@
...
+func GeneratePathChecksum(sourcePath string) (string, error) {
+	digester := digest.SHA256.Digester()
+	tarWriter := tar.NewWriter(digester.Hash())
+
+	err := filepath.Walk(sourcePath, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		header, err := tar.FileInfoHeader(info, "")
```
nalind, line 54: This second parameter is still incorrect for cases where, I think, `info.Mode().Type()` is `fs.ModeSymlink`.  The integration test should include at least one symbolic link in the test directory, which is overwritten using `ln -sf` between invocations.

imagebuildah/stage_executor.go
```diff
@@ -1850,15 +1874,72 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
...
+				}
+				mountOptionSource = mountInfo.Source
+				mountOptionFrom = mountInfo.From
+				// If source is not specified then default is '.'
+				if mountOptionSource == "" {
+					mountOptionSource = "."
+				}
+			}
+			// Source specificed is part of stage, image or additional-build-context.
+			if mountOptionFrom != "" {
```
nalind, line 1900: This value, along with `mountOptionSource`, isn't being reset between loop iterations.

flouthoc: Done near loop start.

imagebuildah/util_test.go
```diff
@@ -0,0 +1,82 @@
...
+		t.Fatalf("Failed to manually generate checksum: %v", err)
+	}
+
+	tarWriter.Close()
+	expectedChecksum := digester.Digest().String()
+
+	// Compare the generated checksum to the expected checksum
+	if checksum != expectedChecksum {
+		t.Errorf("Checksum mismatch: got %s, expected %s", checksum, expectedChecksum)
+	}
```
nalind, line 81: We use the `github.com/stretchr/testify/assert` package for a lot of these.
```suggestion
	assert.Equal(t, expectedChecksum, checksum, "didn't get expected checksum over a sample directory")
```

nalind (changes requested):
Looking better, but the arg parsing still has a potential crasher in it.

flouthoc:
@nalind Could you PTAL again.

flouthoc:
@rhatdan @containers/build-maintainers PTAL

Issue #5692, opened 2024-08-18
buildah cache with  `--layers` does not work when in `RUN` `--mount=type=bind`  uses `image` or `additional-build-context` as source


flouthoc:
**Description**

buildah cache with  `--layers` does not work when in `RUN` `--mount=type=bind`  uses `image` or `additional-build-context` as source

**Steps to reproduce the issue:**
1. Create a build where `--mount=type=bind,source=.,dest=/somepath,from=<image>` or `--mount=type=bind,source=.,dest=/somepath,from=<additional-build-context>`
2. Perform build with `--layers`
3. Perform build with `--layers` again.

**Describe the results you received:**
On second build, entire build must be cached.

**Describe the results you expected:**
Buildah does not uses `cache` on the second build.

**Output of `rpm -q buildah` or `apt list buildah`:**

```
(paste your output here)
```

**Output of `buildah version`:**

```
(paste your output here)
```

**Output of `podman version` if reporting a `podman build` issue:**

```
(paste your output here)
```

**Output of `cat /etc/*release`:**

```
(paste your output here)
```

**Output of `uname -a`:**

```
(paste your output here)
```

**Output of `cat /etc/containers/storage.conf`:**

```
(paste your output here)
```

flouthoc:
Found this issue while working on https://github.com/containers/buildah/pull/5691

flouthoc:
This is not stale, waiting on review here: https://github.com/containers/buildah/pull/5693#issuecomment-2356689254

PR #4526, opened 2023-01-18
stage_executor: while mounting stages make sure freshly built stage is used

flouthoc:
When using `--mount=` in RUN instruction and source is a stage make sure
that freshly built stage is used if the stage selected in source was
just rebuilt.

Closes: https://github.com/containers/buildah/issues/4522

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### Does this PR introduce a user-facing change?

```release-note
stage_executor: while mounting stages make sure freshly built stage is used
```

PR #5693, opened 2024-08-18
stage_executor: set `avoidLookingCache` only if mounting stage and not additional build context

flouthoc:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### How to verify it
Newly added integration tests.

Closes: https://github.com/containers/buildah/issues/5692

```release-note
build: cache now works with when additional build context is used as source of --mount
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -943,6 +943,34 @@ _EOF
 
 }
 
+@test "build-test use image from cache with --mount and burst when image is changed" {
+  _prefetch alpine
+  local contextdir=${TEST_SCRATCH_DIR}/bud/platform
+  mkdir -p $contextdir
+
+  cat > $contextdir/Containerfile << _EOF
+FROM alpine
+RUN touch firstfile
+_EOF
+
+  run_buildah build $WITH_POLICY_JSON --layers -t source -f $contextdir/Containerfile
+
+  cat > $contextdir/Containerfile2 << _EOF
+FROM alpine
+
+RUN --mount=type=bind,source=.,target=/build,from=source ls /build
+
+_EOF
+
+  run_buildah build $WITH_POLICY_JSON --layers -t source2 -f $contextdir/Containerfile2
+  expect_output --substring "firstfile"
+
+  # Building again must use cache
+  run_buildah build $WITH_POLICY_JSON --layers -t source2 -f $contextdir/Containerfile2
+  expect_output --substring "Using cache"
+  assert "$output" !~ "firstfile"
+}
+
 # Verify: https://github.com/containers/buildah/issues/4572
 @test "build-test verify no dangling containers are left" {
   _prefetch alpine busybox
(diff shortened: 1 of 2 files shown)


flouthoc:

rhatdan:
LGTM
@nalind PTAL

flouthoc:
Rebased.

nalind:
Hang on, if the contents of the additional build context are the same as, or changed from, on a previous run, do we, or do we not want to reuse the cache candidate?
What does "build: cache now works with additional build context and inline --mount" mean?  How did it break before?

flouthoc:
> Hang on, if the contents of the additional build context are the same as, or changed from, on a previous run, do we, or do we not want to reuse the cache candidate?

Following concern will be fixed with this https://github.com/containers/buildah/pull/5691, where we will put the hash of the files in the context and make it part of the history.

> What does "build: cache now works with additional build context and inline --mount" mean? How did it break before?

I think cache stopped working for --mount when used source was used as additional build context after PR https://github.com/containers/buildah/pull/4526

flouthoc:
@nalind PTAL

nalind:
> I think cache stopped working for --mount when additional build context is used as source after PR #4526

Stopped working _how_?  Did it fail to find a suitable image to use as a cache image?  Did it choose an unsuitable image?

> Also I fixed the release note it should be
> 
> ```
> build: cache now works with when additional build context is used as source of --mount
> ```

What behavior would someone have seen before when things went wrong?

rhatdan:
@flouthoc waiting on you?

flouthoc:
> @flouthoc waiting on you?

I don't know how I missed GH notification of this PR.

@nalind 

> Stopped working how? Did it fail to find a suitable image to use as a cache image? Did it choose an unsuitable image?

For this part the code which was introduced in PR https://github.com/containers/buildah/pull/4526 accounted `avoidLookingCache` incorrectly for additionalBuildContext when it is used as `source` for  `RUN  --mount` and whenever additionalBuildContext was selected it used to set `avoidLookingCache = true` which should have only been done for the `newly built stages` and not for everything else so caching never worked. It was fault in the logic.

> What behavior would someone have seen before when things went wrong?

Previously the layer based caching was simply not working for users who were using `additional-build-context`  as a `source` for `RUN --mount` command. No one ever reported it directly because I think no-one has spotted this till now.

flouthoc:
@nalind Could you PTAL again :)

tests/bud.bats
```diff
@@ -943,6 +943,34 @@ _EOF
...
+
+  cat > $contextdir/Containerfile2 << _EOF
+FROM alpine
+
+RUN --mount=type=bind,source=.,target=/build,from=source ls /build
+
+_EOF
+
+  run_buildah build $WITH_POLICY_JSON --layers -t source2 -f $contextdir/Containerfile2
+  expect_output --substring "firstfile"
```
nalind, line 966: This should probably also check that the string "Using cache" doesn't show up in the output this build.

nalind:
The "now works" in the release note still doesn't describe what was fixed.  Are cached images used where they weren't before?  Are they not when they shouldn't have been?  Does cache evaluation now take into consideration something that it didn't before that it should have?  If #5691 starts checking the contents of the bound directory, are we going to end up revisiting this change?

flouthoc:
I am thinking if this explains it more correctly, could you read this and see if is answering all questions correctly.


Previously, Buildah would ignore cached layers for step `RUN --mount` steps which contained `source` as an additional build context, causing the step to rebuild every time. This occurred because the internal `stage_executor` incorrectly treated `source` as an internal stage, bypassing the entire cache searching mechanism and not looking for cached layer at all. With this update, Buildah now checks storage for an existing layer to use as a cache when `RUN --mount` includes `source` as an additional build context.


> If https://github.com/containers/buildah/pull/5691 starts checking the contents of the bound directory, are we going to end up revisiting this change?

No, actually the reason of adding this change was that I found this bug while implementing #5691 since buildah was bypassing cache searching mechanism so my changes in #5691 were not working correctly, so I think this can be merged before #5691 so it actually use this change.

flouthoc:
@nalind Any thoughts on above comment.

flouthoc:
Just waiting for review on my last comment https://github.com/containers/buildah/pull/5693#issuecomment-2356689254

TomSweeneyRedHat:
@nalind PTAL again
@flouthoc needs a rebase

TomSweeneyRedHat:
LGTM

PR #5709, opened 2024-08-28
DO NOT MERGE: testing rawhide as-is (without SMI workaround)

edsantiago:
```release-note
None
```


edsantiago:
Well, phooey. We're not testing rawhide. Never mind.

PR #5710, opened 2024-08-30
Vendor `c/common:9d025e4cb348`

Honny1:
This PR vendor pre-release version of `c/common:9d025e4cb348` fixes https://github.com/containers/podman/issues/23331. It also adapts the code to the new `c/common/libimage` API.

#### What type of PR is this?
In pre-release version of c/common is fixed flake https://github.com/containers/podman/issues/23331. This fix changes API of c/common. So it needs adapt the code to the new `c/common/libimage` API. And avoid unexpected issue for next vendor dance. 

/kind flake

Fixes https://github.com/containers/podman/issues/23331

```release-note
None
```


cmd/buildah/images.go
```diff
@@ -142,10 +142,26 @@ func imagesCmd(c *cobra.Command, args []string, iopts *imageResults) error {
...
+	imagesByNames, err := runtime.ListImagesByNames(args)
 	if err != nil {
 		return err
 	}
+	imagesByFilter, err := runtime.ListImages(ctx, options)
+	if err != nil {
+		return err
+	}
+
+	imagesByNameIDs := map[string]bool{}
```
mtrmac, line 154: - If this is a set, it can be a `map[string]struct{}`.
- For a set of image IDs, the name seems misleading to me

cmd/buildah/images.go
```diff
@@ -142,10 +142,26 @@ func imagesCmd(c *cobra.Command, args []string, iopts *imageResults) error {
...
+	for _, image := range imagesByNames {
+		imagesByNameIDs[image.ID()] = true
+	}
+
+	images := make([]*libimage.Image, 0, len(imagesByNames))
+	for _, image := range imagesByFilter {
```
mtrmac, line 160: I think this is incorrect in the case where `len(args) == 0`; the previous code would use all images returned by the filter.

[Intuitively, it seems to me that the `len(args) > 0` (`len(args) == 1`) case should be a conditional code path, always doing the `ListImagesByNames` does not help clarity. But that’s a guess without trying.]

mtrmac (commented):
Oops, I’m afraid I didn’t think to look at the Buildah CLI for users (I was assuming that the Buildah code vendored into Podman would cover all of the Buildah codebase).

---

- It’s … not entirely clear tome that `buildah images --filter … singleImageName` is expected. But the old code did implement that, so making an effort might make sense.
- Reading how `args` is treated, it always has either 0 or 1 elements. So, I think we don’t really need an `imagesByNames` _set_; if we have a single name to look up, deal with that one name.

mtrmac (commented):
/lgtm

Thanks!

Luap99 (approved):
LGTM

Luap99:
@Honny1 can you rebase this, looks like there is merge conflict now

Luap99:
Looks like I do not have approve power here either.

Honny1:
@Luap99 I made a rebase.

PR #5715, opened 2024-09-03
Update to go 1.22

Luap99:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake

/kind other

Update golang so we can keep updating dependecies

```release-note
Buildah now requires go 1.22 to build.
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,21 +1,8 @@
 module github.com/containers/buildah
 
-// Minimum required golang version
-go 1.21.0 // *****  ATTENTION  WARNING  CAUTION  DANGER  ******
+// Warning: Ensure the "go" and "toolchain" versions match exactly to prevent unwanted auto-updates
 
-//         Go versions 1.21 and later will AUTO-UPDATE based
-//         on currently running tools and the (new) `toolchain`
-//         value (when also increasing the `go` value above).
-//         ref: https://go.dev/doc/toolchain  Because several
-//         different distros and distro-versions build from
-//         this code, golang version consistency is
-//         desireable.  After manually updating to 1.21, a
-//         `toolchain` specification should be added to pin
-//         the version and block auto-updates.  This does not
-//         block any future changes to the `go` value.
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

go.mod
```diff
@@ -1,21 +1,8 @@
 module github.com/containers/buildah
 
-// Minimum required golang version
-go 1.21.0 // *****  ATTENTION  WARNING  CAUTION  DANGER  ******
+// Warning: Ensure the "go" and "toolchain" versions match exactly to prevent unwanted auto-updates
```
edsantiago, line 3: Will it be obvious to future maintainers what "toolchain" means? The string "toolchain" does not appear anywhere else in this file. My guess was, this means `test/tools/go.mod`, but that file has go 1.21

edsantiago (approved):
LGTM with one question

go.mod
Luap99: It is the same comment we have in podman and c/common and yes a go dev should now what these mean I would say, https://go.dev/doc/modules/gomod-ref#toolchain
Though I am happy to alter this comment to something else that makes it more clear, generally speaking we never want toolchain to be newer than the go version and it seems at some point the go vendor process decided to not add toolchain if it matches the exact go version, i.e. if we try to add `toolchain go1.22.0` make vendor removes it again. Only if you have something like `toolchain go1.22.1` it stays but this is not something we want.

So I guess we should simplify this comment and say we never ever want a toolchain directive in this file.

Luap99: Can you merge this for now? 
I have some outstanding discussions around the handling of patch versions in the go line with @mtrmac so I like to wait for whatever we decide on and then update the comments based on that

edsantiago:
Looks like I don't have approve privs in this repo

PR #5721, opened 2024-09-05
Fix conflict on heredoc temp files

kmichel-aiven:
#### What type of PR is this?

/kind bug

If two image builds start concurrently and they both use an heredoc with the same EOF marker, then they will use the same filename and overwrite each other.

It's possible for concurrent builds to either use the wrong heredoc content or to fail because the file of one build is deleted by the other, leading to this error message:

```
STEP 13/21: COPY <<EOF /file.conf

Error: building at STEP "COPY <<EOF /file.conf": reading "/var/tmp/EOF": copier: get: globs [/EOF] matched nothing (0 filtered out): no such file or directory
```

```release-note

```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -451,7 +451,7 @@ func (s *StageExecutor) performCopy(excludes []string, copies ...imagebuilder.Co
 				data = strings.TrimPrefix(data, "\n")
 				// add breakline when heredoc ends for docker compat
 				data = data + "\n"
-				tmpFile, err := os.Create(filepath.Join(parse.GetTempDir(), path.Base(filepath.ToSlash(file.Name))))
+				tmpFile, err := os.CreateTemp(parse.GetTempDir(), path.Base(filepath.ToSlash(file.Name)))
 				if err != nil {
 					return fmt.Errorf("unable to create tmp file for COPY instruction at %q: %w", parse.GetTempDir(), err)
 				}


kmichel-aiven:
I would like to add tests but I'm not familiar with the project, what would be the best place to add these ?

(I've tried adding to `bud.bats` but `run_buildah` doesn't seem ready to test concurrent operations.)

TomSweeneyRedHat:
@edsantiago thoughts on tests ?

edsantiago:
I see no reason for tests. However, @kmichel-aiven , could you please clean up your git commit, and rebase?

TomSweeneyRedHat:
@flouthoc if you get a minute to look at this one, I'd appreciate it.

nalind:
This isn't going to work as intended when the destination for a COPY with a heredoc is a directory - the destination file's name should be the same as the delimiter, and `os.CreateTemp()` will create a file with a name that has a suffix added to ensure that it's unique.  If it used `os.MkdirTemp()`, and placed the file in the resulting directory (deferring removal of the directory as well as the file), I think it would work as intended.
The "bud build with heredoc content" integration test should suffice to catch that, so I'll add the no-new-tests-needed label so that it can be run.

flouthoc:
> This isn't going to work as intended when the destination for a COPY with a heredoc is a directory - the destination file's name should be the same as the delimiter, and `os.CreateTemp()` will create a file with a name that has a suffix added to ensure that it's unique. If it used `os.MkdirTemp()`, and placed the file in the resulting directory (deferring removal of the directory as well as the file), I think it would work as intended. ~The "bud build with heredoc content" integration test should suffice to catch that, so I'll add the no-new-tests-needed label so that it can be run.~

@kmichel-aiven Could you please do the changes as suggested above. So we can move ahead with this PR.

nalind:
Obsoleted by https://github.com/containers/buildah/pull/5959, closing.

PR #5731, opened 2024-09-11
build: fall back to parsing a TARGETPLATFORM build-arg

nalind:
#### What type of PR is this?

/kind feature

If we're not given an explicit platform or arch or os to target for a build, but someone defined TARGETPLATFORM as a build argument, parse it.

#### How to verify it

New integration test!

This should fix the last thing in https://github.com/containers/podman/issues/23046.

```release-note
If a target platform for `buildah build` is not specified with one of the `--platform`, `--os`, or `--arch` flags, but a `TARGETPLATFORM` build argument has been set with `--build-arg`, its value will be used as though it had been specified to `--platform`.
```


flouthoc (approved):
LGTM

PR #5732, opened 2024-09-11
In a container, try to register binfmt_misc

nalind:
#### What type of PR is this?

/kind feature

If we're running a command in a working container whose platform doesn't match our own, attempt to register any emulators for which we find configurations of the type included in Fedora's qemu-user-static packages.

#### How to verify it

In-container tests, provided the CI node is new enough, will start being able to use emulation if they couldn't before.

None

* Linux 6.6 and later allow binfmt_misc to be mounted inside user namespaces.
* This should be helpful for people who are using our Fedora-based images on systems which don't provide their own qemu-user-static binaries.

```release-note
buildah container images will now attempt to use the qemu-user-static binaries they include, if they also include configuration for them, for running binaries built for non-native architectures.
```

### DIFF
--- a/pkg/binfmt/binfmt.go
+++ b/pkg/binfmt/binfmt.go
@@ -1,0 +1,83 @@
+//go:build linux
+
+package binfmt
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"syscall"
+
+	"github.com/containers/storage/pkg/unshare"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+)
+
+// MaybeRegister() calls Register() if the current context is a rootless one,
+// or if the "container" environment variable suggests that we're in a
+// container.
+func MaybeRegister(configurationSearchDirectories []string) error {
+	if unshare.IsRootless() || os.Getenv("container") != "" { // we _also_ own our own mount namespace
+		return Register(configurationSearchDirectories)
+	}
+	return nil
+}
+
+// Register() registers binfmt.d emulators described by configuration files in
+// the passed-in slice of directories, or in the union of /etc/binfmt.d,
+// /run/binfmt.d, and /usr/lib/binfmt.d if the slice has no items.  If any
+// emulators are configured, it will attempt to mount a binfmt_misc filesystem
+// in the current mount namespace first, ignoring only EPERM and EACCES errors.
+func Register(configurationSearchDirectories []string) error {
+	if len(configurationSearchDirectories) == 0 {
+		configurationSearchDirectories = []string{"/etc/binfmt.d", "/run/binfmt.d", "/usr/lib/binfmt.d"}
+	}
+	mounted := false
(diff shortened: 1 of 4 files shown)


cmd/buildah/main.go
```diff
@@ -118,6 +120,10 @@ func init() {
 	if err := rootCmd.PersistentFlags().MarkHidden("memory-profile"); err != nil {
 		logrus.Fatalf("unable to mark memory-profile flag as hidden: %v", err)
 	}
+	rootCmd.PersistentFlags().BoolVar(&globalFlagResults.registerBinfmt, "register-binfmt", os.Getenv("container") == "oci", "attempt to register binfmt_misc interpreters")
```
rhatdan, line 123: If I run buildah within a podman container container=podman, should we handle this case?

Is "oci" set hen inside of Kubernetes?

nalind: In my spot-check, podman sets it to "oci" when using either runc or crun as the runtime.  Do you have more information about how you're invoking podman, and which image you're using?

nalind: I was checking with podman 5.2.2 on Fedora 40, running quay.io/buildah/stable as an unprivileged user.

rhatdan: $ podman run alpine printenv container
podman
$ podman run fedora printenv container
oci
$ podman run ubuntu printenv container
podman

Depends on if the image defines the container environment variable I believe.  I think you could just check if the container environment variable is set.

Docker does not set it at all but does create the /.dockerenv file.  We create the /run/.containerenv file.

I think there is also a way to look at pid1 and figure it out which is what systemd does.

nalind: Oh, ugh.  `skopeo inspect --config docker://registry.fedoraproject.org/fedora:40` does, indeed, include "container=oci" in its image config.

pkg/binfmt/binfmt.go
```diff
@@ -0,0 +1,83 @@
...
+		configurationSearchDirectories = []string{"/etc/binfmt.d", "/run/binfmt.d", "/usr/lib/binfmt.d"}
+	}
+	mounted := false
+	for _, searchDir := range configurationSearchDirectories {
+		globs, err := filepath.Glob(filepath.Join(searchDir, "*.conf"))
+		if err != nil {
+			return fmt.Errorf("looking for binfmt.d configuration in %q: %w", searchDir, err)
+		}
+		for _, conf := range globs {
+			f, err := os.Open(conf)
```
rhatdan, line 45: Should we only be doing this if the image has different ARCH then the local arch?

nalind: Sure.  At that point we're too far from the CLI to make it something that can be manually controlled very easily, but then maybe we're fine without that.

rhatdan:
LGTM
@flouthoc @giuseppe @mheon PTAL

giuseppe (approved):
/lgtm

nalind:
Oops, forgot I left a hold on this.
/unhold

PR #5741, opened 2024-09-17
Do not error on trying to write IMA xattr as rootless

mheon:
Rootless users cannot set the `security.ima` xattr on files (presumably for security reasons, they get an EPERM on trying to do so). We will normally try and preserve that xattr, so when trying to add a file with an IMA xattr to a build on a Buildah without this patch, you get an error. With this patch, the error is downgraded to a warning, as it's better to successfully build with a missing xattr than blocking all builds which want to include the offending file.

The urgency on this has become somewhat higher as it seems like F41/Rawhide are installing rpm-plugin-ima by default, which is setting IMA xattrs on some files that Podman relies on - for example, the catatonit binary we use for pid pause images. Without this patch, building the pause image as rootless will always fail on a system with rpm-plugin-ima installed.

Fixes: https://github.com/containers/podman/issues/18543

#### What type of PR is this?

/kind bug

```release-note
Including a file with the `security.ima` xattr set as a non-root user will no longer fail as the xattr could not be set.
```


mheon:
I don't know if testing this is particularly possible in CI, as it would require the file to be set up as root, and then an integration test run as a non-root user to verify the functionality. Could be done by including the test file in the VM images, maybe?

mheon:
@edsantiago The Podman package is installed in the test VMs, right? That should have dragged in catatonit, which we know has IMA xattrs we can test on.

@nalind Are any of the existing integration tests run rootless?

edsantiago:
@mheon please see my comment above. TL;DR we have a kludge workaround in test VMs, where we strip IMA. Without that, we would have no CI. The test images I mention in my comment above are built without that workaround, so you can use them in your testing.

nalind:
> @nalind Are any of the existing integration tests run rootless?

Yes, there are rootless test jobs.

mheon:
@edsantiago Mind taking a look at the test I added?

tests/bud.bats
```diff
@@ -6946,3 +6946,23 @@ _EOF
...
+  local contextdir=${TEST_SCRATCH_DIR}/add-ima
+  mkdir -p $contextdir
+  cat > $contextdir/Dockerfile <<EOF
+FROM alpine
+ADD /usr/libexec/catatonit/catatonit /bin/catatonit
+EOF
+
+  # Verify that /usr/libexec/catatonit/catatonit both exists and has an IMA xattr
+  if getfattr -d -m 'security.ima' /usr/libexec/catatonit/catatonit | grep -q ima; then
+    skip "catatonit does not exist or does not have an appropriate xattr"
```
edsantiago, line 6962: I'm not a big fan of ambiguous `skips` or test failures. If I were looking at this skip, I'd want to know more.

tests/bud.bats
```diff
@@ -6946,3 +6946,23 @@ _EOF
...
+
+  local contextdir=${TEST_SCRATCH_DIR}/add-ima
+  mkdir -p $contextdir
+  cat > $contextdir/Dockerfile <<EOF
+FROM alpine
+ADD /usr/libexec/catatonit/catatonit /bin/catatonit
+EOF
+
+  # Verify that /usr/libexec/catatonit/catatonit both exists and has an IMA xattr
+  if getfattr -d -m 'security.ima' /usr/libexec/catatonit/catatonit | grep -q ima; then
```
edsantiago, line 6961: Did you mean to use `-m` (match)? Not the tighter `-n` (name)?

tests/bud.bats
```diff
@@ -6946,3 +6946,23 @@ _EOF
...
+    skip "catatonit does not exist or does not have an appropriate xattr"
+  fi
+
+  # We do not care if the attribute was actually added, as rootless is allowed to discard it.
+  # Only that the file builds successfully.
+  run_buildah build -f $contextdir/Dockerfile $contextdir
```
edsantiago, line 6967: Can probably ditch the `-f Dockerfile` and save some duplication

edsantiago (approved):
LGTM with a few questions and nits. Suggestion:
```bash
@test "bud with ADD with file with IMA xattr" {
  _prefetch alpine

  local catatonit=/usr/libexec/catatonit/catatonit
  if [[ ! -e $catatonit ]]; then
      skip "Catatonit path $catatonit not found"
  fi

  run getfattr -d -n security.ima $catatonit
  echo "getfattr $catatonit -> '$output'"
  if [[ $status -ne 0 ]]; then
      skip "pointless: $catatonit does not have the desired xattr"
  fi

  local contextdir=${TEST_SCRATCH_DIR}/add-ima
  mkdir -p $contextdir
  cat > $contextdir/Dockerfile <<EOF
FROM alpine
ADD $catatonit /bin/catatonit
EOF

  # We do not care if the attribute was actually added, as rootless is allowed to discard it.
  # Only that the file builds successfully.
  run_buildah build $contextdir
}
```
Do we know if the `attr` package is installed in CI VMs?

edsantiago:
Welp. (1) Buildah CI doesn't test on rawhide; (2) Debian doesn't have `attr`; (3) [f40 is failing](https://api.cirrus-ci.com/v1/task/6413991776354304/logs/integration_test.log) in a way that makes no sense to me

nalind:
The current iteration of the test doesn't ensure that there's a `/usr/libexec/catatonit/catatonit` under the build context directory.  I'd suggest either copying it first, then checking for the extended attribute, or reworking the test to use `buildah add`, which should hit the same code paths.

mheon:
@nalind Can't copy it, that strips the xattr. I'll rework to use `buildah add`.

nalind:
/lgtm
This doesn't address cases where the a file with this xattr is part of a base image, but we don't control that in this repository.

mheon:
I'm inclined to just let those error for now - we've never heard of a case where that has actually happened.

tests/add.bats
```diff
@@ -321,3 +321,16 @@ stuff/mystuff"
   run_buildah 125 add --retry-delay=0.142857s --retry=14 --cert-dir ${TEST_SCRATCH_DIR} $cid https://localhost:${HTTP_SERVER_PORT}/randomfile
   assert "$output" =~ "retrying in 142.*ms .*14/14.*"
 }
+
+@test "add file with IMA xattr" {
+    if getfattr -d -n 'security.ima' /usr/libexec/catatonit/catatonit | grep -q ima; then
```
nalind, line 326:
```suggestion
    if ! getfattr -d -n 'security.ima' /usr/libexec/catatonit/catatonit | grep -q ima; then
```

edsantiago: Also note that this will not be tested in podman CI, because that only runs `bud.bats`. When this is vendored into podman, a new test will be required.

mheon:
@edsantiago We should consider new VM builds once this lands in Podman. Not super urgent though.

Issue #5750, opened 2024-09-20
buildah commands that don't involve building containers at all run through unshare


AdamWill:
**Description**

This is yet another in the series of 'issues to do with running buildah in a locked-down container' - see https://github.com/containers/buildah/issues/4563 , https://github.com/containers/buildah/issues/1901 - but with a twist: we want to use buildah to do things that don't involve building containers *at all*, but those commands still run through unshare. @terinjokes also noted this in https://github.com/containers/buildah/issues/1901#issuecomment-1674648806 .

@jeremycline and I are working on improving Fedora's container registry publication. All this needs to do is take some already-built container images from a Fedora compose, create a multiarch manifest, and push the images and the manifest to some registries. The only tools we've found for doing this are buildah and podman. We have [the code for this](https://pagure.io/cloud-image-uploader) all written. But when we try to deploy it in an openshift container (which is where we want to deploy it), it blows up because unshare doesn't work. Well, sure, it doesn't, and we don't really want to mess around with the container security stuff to allow it, because *we shouldn't need it* to do this. Commands which just create manifests and push images to registries should not need to run through isolation; they're just writing trivial JSON and making web API calls. (podman appears to have basically the same problem as buildah).

We would happily use a different tool for this if we could find one that just does this and avoids the mess, but we can't (suggestions welcome). We really don't want to have to write one.

**Steps to reproduce the issue:**
1. Create a bog-standard Fedora container in Fedora openshift, with buildah in it
2. Try and run `buildah login` or `buildah manifest create` or `buildah manifest push`

**Describe the results you received:**
```
sh-5.2$ HOME=/tmp/whatever buildah login
WARN[0000] Reading allowed ID mappings: reading subuid mappings for user "1001340000" and subgid mappings for group "1001340000": no subuid ranges found for user "1001340000" in /etc/subuid 
WARN[0000] Found no UID ranges set aside for user "1001340000" in /etc/subuid. 
WARN[0000] Found no GID ranges set aside for user "1001340000" in /etc/subgid. 
Error during unshare(CLONE_NEWUSER): Function not implemented
ERRO[0000] parsing PID "": strconv.Atoi: parsing "": invalid syntax 
ERRO[0000] (Unable to determine exit status)            
sh-5.2$ 
```

**Describe the results you expected:**
successful login

AdamWill:
The `HOME=/tmp/whatever` is needed because without that buildah has a fit about `~/.config` not being owned by the current user, btw.

[...]

PR #3278, opened 2021-06-02
Fix handling of auth.json file while in a user namespace

rhatdan:
If you are running buildah within a user namespace, without
XDG_RUNTIME_DIR being set, we need to make sure buildah login, logout
handle XDG_RUNTIME_DIR the same was as buildah push.

Running buildah within a container is triggering failures, where buildah
login puts the auth.json file in /run/containers/0/auth.json but buildah
push looks for it in
/tmp/container-user-0/containers/containers/auth.json

Fixes: https://github.com/containers/buildah/issues/3259

[NO TESTS NEEDED] The existing tests should see if this causes any
problems.

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### Does this PR introduce a user-facing change?

```release-note

```

PR #5752, opened 2024-09-23
Run logout,manifest,source,unmount without userns

rhatdan:
Fixes: https://github.com/containers/buildah/issues/5750

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note

```

### DIFF
--- a/cmd/buildah/main.go
+++ b/cmd/buildah/main.go
@@ -142,7 +142,7 @@ func before(cmd *cobra.Command) error {
 	}
 
 	switch cmd.Use {
-	case "", "help", "version", "mount":
+	case "", "help", "login", "logout", "manifest", "mount", "source", "unmount", "version":
 		return nil
 	}
 	debugCapabilities()


rhatdan:
@AdamWill PTAL
@nalind @giuseppe @lsm5 @TomSweeneyRedHat @flouthoc PTAL

nalind:
I'm skeptical that this will work for `manifest push`.

rhatdan:
@AdamWill could you see if buildah manifest push fails when not in the user namespace.  Since this follows the similar path to buildah push, most likely @nalind is correct.

rhatdan:
```
$ ./bin/buildah manifest create quay.io/rhatdan/mymanifest
ff08b4da30d4c9474dcbebde7d61e03864f404541c44f5b2890798d7dbcae3ca
$ ./bin/buildah manifest push quay.io/rhatdan/mymanifest
Getting image list signatures
Copying 0 images generated from 0 images in list
Writing manifest list to image destination
Storing list signatures
```

This worked.

nalind:
I'm doubtful that many of the relevant code paths were invoked there, since there was no image rootfs being pushed.

AdamWill:
I'm not sure this is safe. login *used* to be in the list, but you took it out in https://github.com/containers/buildah/pull/3278 because of https://github.com/containers/buildah/issues/3259 .

If adding all these other things to the list means #3259 won't be an issue any more, great, but I'm worried maybe that's not the case :)

rhatdan:
Well it looks like login, logout and manifest are likely to require UserNS then.

AdamWill:
well the thinking behind my suggestion (make it an arg) was that what you can't do is "cross the streams", right? run some things in userns but some things outside it. but (at least as I understood it) it *should* in theory be okay to run only non-container-build-y commands, all outside userns. so my idea was to provide a somewhat-hidden way to do that, for this specific use case. if a command-line argument is too prominent it could be an env var or something, I guess.

rhatdan:
I am fine with adding a hidden option --unshare and forcing the user to need to deal with it.

rhatdan:
@nalind Thoughts?

nalind:
Would it actually solve the problem?  Attempting to push layers without configuring a user namespace can very easily trip over permissions problems.

AdamWill:
I can try and test my patch later, I didn't have time when I wrote it.

TomSweeneyRedHat:
I'd be curious to hear @giuseppe 's opinion here.  The code change looks fine for what you want to do, I'll let others decide if that's an appropriate approach.  The hidden `--unshare` option is appealing.

Looks like a rebase is needed.

giuseppe:
> I'd be curious to hear @giuseppe 's opinion here. The code change looks fine for what you want to do, I'll let others decide if that's an appropriate approach. The hidden `--unshare` option is appealing.

if we are accessing the storage as @nalind pointed out, it won't work without a user namespace since we need `CAP_DAC_OVERRIDE`  to access some of the files in an exploded rootfs

AdamWill:
sorry for not testing this; after some extensive discussion we decided to try having the image uploader produce its own multi-arch manifest and push it with skopeo, to avoid the need for buildah and podman. I'm just finishing up that code now (adapted from https://pagure.io/koji-flatpak/blob/main/f/koji_flatpak/plugins/flatpak_builder_plugin.py#_586 ). If it works, we'll probably not care about this so much (though it'd still be nice if this could be addressed somehow in skopeo or buildah or podman so we could drop the manifest creation code from our tool).

AdamWill:
Update - [we changed our uploader tool to build the manifest itself](https://pagure.io/cloud-image-uploader/pull-request/31), and it works, so we'll just go with that for now. Bit more code than we really want to carry/maintain, but hey. If there's ever a way to use podman or buildah or skopeo to construct the manifest that doesn't run through this userns stuff, we'll try it.

PR #5753, opened 2024-09-24
chroot: add newlines at the end of printed error messages

nalind:
#### What type of PR is this?

/kind other

If we print an error message to stderr, end it with a newline.  This should make it easier to tell the difference between multiple error messages.

#### How to verify it

Manual verification only.

None

```release-note
None
```


rhatdan:
LGTM

PR #5755, opened 2024-09-25
Make `buildah manifest push --all` true by default

k9withabone:
#### What type of PR is this?

> /kind api-change

/kind bug

> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

Changed the `--all` option of `buildah manifest push` to be true by default. This matches the behavior of the equivalent Podman option (`podman manifest push --all`), making it easier to switch between Podman and Buildah.

Updated **buildah-manifest-push**(1) docs to reflect this change.

#### How to verify it

Use `buildah manifest push --help`, see that at the end of the line for the `--all` it says `(default true)`.

Fixes #5547

```release-note
`buildah manifest push --all` is now true by default to match Podman.
```

### DIFF
--- a/cmd/buildah/manifest.go
+++ b/cmd/buildah/manifest.go
@@ -246,7 +246,7 @@ func init() {
 	manifestPushCommand.SetUsageTemplate(UsageTemplate())
 	flags = manifestPushCommand.Flags()
 	flags.BoolVar(&manifestPushOpts.rm, "rm", false, "remove the manifest list if push succeeds")
-	flags.BoolVar(&manifestPushOpts.all, "all", false, "also push the images in the list")
+	flags.BoolVar(&manifestPushOpts.all, "all", true, "also push the images in the list")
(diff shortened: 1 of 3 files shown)


rhatdan:
Could you Fix the man page and add a tests.

Test could be the equivalent of:

buildah manifest push --help | grep all.*true

k9withabone:
Added test "manifest-push-all-default-true" to `tests/lists.bats`.

What's wrong with the man page?

rhatdan:
Add information about the default being true.  Booleans without default in man pages are assumed False.

k9withabone:
> Add information about the default being true.

Already done. It is the same as the Podman man page now.

rhatdan:
LGTM
@nalind @flouthoc @giuseppe @mheon PTAL

flouthoc (approved):
LGTM

PR #5777, opened 2024-10-09
Add hack/bats - bats wrapper like podman's

edsantiago:
Ugh. This is three and a half years old but I never sent a PR.
Sorry.

```release-note
None
```


rhatdan:
LGTM
/approve

nalind:
This looks largely redundant, as `tests/helpers.bash` does a lot of the same work to find the in-source-tree binaries when `bats` is invoked without a wrapper.

PR #5778, opened 2024-10-09
Properly validate cache IDs and sources

mheon:
The `--mount type=cache` argument to the `RUN` instruction in Dockerfiles was using `filepath.Join` on user input, allowing crafted paths to be used to gain access to paths on the host, when the command should normally be limited only to Buildah;s own cache and context directories. Switch to `filepath.SecureJoin` to resolve the issue.

Fixes CVE-2024-9675

#### What type of PR is this?

> /kind bug

This fixes a CVE in Buildah that allows arbitrary directories from the host to be mounted into build containers.

#### How to verify it

Reproducer script:

```
#!/bin/bash

set -o errexit

mkdir /root/dir
echo SHOULDNTSEETHIS > /root/dir/file.txt
mkdir /root/ctr

cat << EOF > /root/ctr/Containerfile
FROM alpine
RUN --mount=type=cache,id=../../../../../../../../../../../root/dir,target=/var/tmp \
ls -l /var/tmp && cat /var/tmp/file.txt && touch /var/tmp/BREAKOUT 
EOF

cd /root/ctr
buildah bud --no-cache .
ls -l /root/dir/BREAKOUT
```

Reproduces without this patch, no longer reproduces with it.

```release-note
Fixed CVE-2024-9675 which allowed arbitrary paths from the host to be mounted into a build container using the `--mount type=cache` argument to the `RUN` instruction in Dockerfiles.
```

### DIFF
--- a/internal/volumes/volumes.go
+++ b/internal/volumes/volumes.go
@@ -22,6 +22,7 @@ import (
 	"github.com/containers/storage/pkg/idtools"
 	"github.com/containers/storage/pkg/lockfile"
 	"github.com/containers/storage/pkg/unshare"
+	digest "github.com/opencontainers/go-digest"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	selinux "github.com/opencontainers/selinux/go-selinux"
 )
@@ -374,7 +375,11 @@ func GetCacheMount(args []string, _ storage.Store, _ string, additionalMountPoin
 			return newMount, nil, fmt.Errorf("no stage found with name %s", fromStage)
 		}
 		// path should be /contextDir/specified path
-		newMount.Source = filepath.Join(mountPoint, filepath.Clean(string(filepath.Separator)+newMount.Source))
+		evaluated, err := copier.Eval(mountPoint, string(filepath.Separator)+newMount.Source, copier.EvalOptions{})
+		if err != nil {
+			return newMount, nil, err
+		}
+		newMount.Source = evaluated
 	} else {
 		// we need to create the cache directory on the host if no image is being used
 
@@ -391,11 +396,15 @@ func GetCacheMount(args []string, _ storage.Store, _ string, additionalMountPoin
 		}
 
 		if id != "" {
-			newMount.Source = filepath.Join(cacheParent, filepath.Clean(id))
-			buildahLockFilesDir = filepath.Join(BuildahCacheLockfileDir, filepath.Clean(id))
+			// Don't let the user control where we place the directory.
+			dirID := digest.FromString(id).Encoded()[:16]
+			newMount.Source = filepath.Join(cacheParent, dirID)
+			buildahLockFilesDir = filepath.Join(BuildahCacheLockfileDir, dirID)
 		} else {
-			newMount.Source = filepath.Join(cacheParent, filepath.Clean(newMount.Destination))
-			buildahLockFilesDir = filepath.Join(BuildahCacheLockfileDir, filepath.Clean(newMount.Destination))
+			// Don't let the user control where we place the directory.
+			dirID := digest.FromString(newMount.Destination).Encoded()[:16]
+			newMount.Source = filepath.Join(cacheParent, dirID)
+			buildahLockFilesDir = filepath.Join(BuildahCacheLockfileDir, dirID)
 		}
 		idPair := idtools.IDPair{
 			UID: uid,
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -6971,3 +6971,37 @@ _EOF
   run_buildah 125 build $WITH_POLICY_JSON ${TEST_SCRATCH_DIR}
   expect_output --substring "invalid mount option"
 }
+
+@test "build-check-cve-2024-9675" {
+  _prefetch alpine
+
+  touch ${TEST_SCRATCH_DIR}/file.txt
+
+  cat > ${TEST_SCRATCH_DIR}/Containerfile <<EOF
+FROM alpine
+RUN --mount=type=cache,id=../../../../../../../../../../../$TEST_SCRATCH_DIR,target=/var/tmp \
+ls -l /var/tmp && cat /var/tmp/file.txt
+EOF
+
+  run_buildah 1 build --no-cache ${TEST_SCRATCH_DIR}
+  expect_output --substring "cat: can't open '/var/tmp/file.txt': No such file or directory"
+
+  cat > ${TEST_SCRATCH_DIR}/Containerfile <<EOF
+FROM alpine
+RUN --mount=type=cache,source=../../../../../../../../../../../$TEST_SCRATCH_DIR,target=/var/tmp \
+ls -l /var/tmp && cat /var/tmp/file.txt
+EOF
+
+  run_buildah 1 build --no-cache ${TEST_SCRATCH_DIR}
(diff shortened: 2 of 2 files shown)


rhatdan:
LGTM
@flouthoc @nalind @Luap99 @giuseppe PTAL

internal/volumes/volumes.go
```diff
@@ -374,7 +375,11 @@ func GetCacheMount(args []string, _ storage.Store, _ string, additionalMountPoin
 			return newMount, nil, fmt.Errorf("no stage found with name %s", fromStage)
 		}
 		// path should be /contextDir/specified path
-		newMount.Source = filepath.Join(mountPoint, filepath.Clean(string(filepath.Separator)+newMount.Source))
+		newSrc, err := securejoin.SecureJoin(mountPoint, filepath.Clean(string(filepath.Separator)+newMount.Source))
```
nalind, line 378: Try making this align more with what GetBindMount() is doing around line 198, so that they work the same.

internal/volumes/volumes.go
```diff
@@ -391,11 +396,27 @@ func GetCacheMount(args []string, _ storage.Store, _ string, additionalMountPoin
 		}
 
 		if id != "" {
-			newMount.Source = filepath.Join(cacheParent, filepath.Clean(id))
-			buildahLockFilesDir = filepath.Join(BuildahCacheLockfileDir, filepath.Clean(id))
+			newSrc, err := securejoin.SecureJoin(cacheParent, filepath.Clean(id))
```
nalind, line 399: We don't need to be using the ID to construct this path, at all.  Try something else, like running it through `digest.Canonical.FromString()` slicing off the first N characters in the result's `Encoded()` value, and using that.

internal/volumes/volumes.go
```diff
@@ -391,11 +396,27 @@ func GetCacheMount(args []string, _ storage.Store, _ string, additionalMountPoin
...
+			newMount.Source = newSrc
+			newLockFilesDir, err := securejoin.SecureJoin(BuildahCacheLockfileDir, filepath.Clean(id))
+			if err != nil {
+				return newMount, nil, err
+			}
+			buildahLockFilesDir = newLockFilesDir
 		} else {
-			newMount.Source = filepath.Join(cacheParent, filepath.Clean(newMount.Destination))
-			buildahLockFilesDir = filepath.Join(BuildahCacheLockfileDir, filepath.Clean(newMount.Destination))
+			newSrc, err := securejoin.SecureJoin(cacheParent, filepath.Clean(newMount.Destination))
```
nalind, line 410: We don't need to be using the destination to construct this path, at all.  Try something else, like running it through `digest.Canonical.FromString()` slicing off the first N characters in the result's `Encoded()` value, and using that.

tests/bud.bats
```diff
@@ -6971,3 +6971,18 @@ _EOF
...
+
+  cat > ${TEST_SCRATCH_DIR}/Containerfile <<EOF
+FROM alpine
+RUN --mount=type=cache,id=../../../../../../../../../../../$TEST_SCRATCH_DIR,target=/var/tmp \
+ls -l /var/tmp && cat /var/tmp/file.txt && touch /var/tmp/BREAKOUT
+EOF
+
+  run_buildah 1 build --no-cache ${TEST_SCRATCH_DIR}
+  expect_output --substring "cat: can't open '/var/tmp/file.txt': No such file or directory"
+}
```
nalind, line 6988: Please also check with "source" values, which can be made relative to a "from" value that names a location specified at the command line using the "--build-context" flag or the end-result of a previous stage, so that we hit the cases for all three locations being changed.

tests/bud.bats
```diff
@@ -6971,3 +6971,28 @@ _EOF
...
+@test "build-check-cve-2024-9675" {
+  _prefetch alpine
+
+  touch ${TEST_SCRATCH_DIR}/file.txt
+
+  cat > ${TEST_SCRATCH_DIR}/Containerfile <<EOF
+FROM alpine
+RUN --mount=type=cache,id=../../../../../../../../../../../$TEST_SCRATCH_DIR,target=/var/tmp \
+ls -l /var/tmp && cat /var/tmp/file.txt
+EOF
```
nalind, line 6984: Please add a variant of this one that specifies "source" instead of "id", so that we exercise the path that doesn't hash the ID when there is no "from" argument.

tests/bud.bats
```diff
@@ -6971,3 +6971,37 @@ _EOF
...
+  expect_output --substring "cat: can't open '/var/tmp/file.txt': No such file or directory"
+
+  cat > ${TEST_SCRATCH_DIR}/Containerfile <<EOF
+FROM alpine
+RUN --mount=type=cache,source=../../../../../../../../../../../$TEST_SCRATCH_DIR,target=/var/tmp \
+ls -l /var/tmp && cat /var/tmp/file.txt
+EOF
+
+  run_buildah 1 build --no-cache ${TEST_SCRATCH_DIR}
+  expect_output --substring "cat: can't open '/var/tmp/file.txt': No such file or directory"
```
nalind, line 6996: If I back out the changes to the sources (and comment out the lines 6986-6987), this part of the test still passes.  Any ideas why?

tests/bud.bats
```diff
@@ -6971,3 +6971,37 @@ _EOF
...
+
+  mkdir ${TEST_SCRATCH_DIR}/cve20249675
+  cat > ${TEST_SCRATCH_DIR}/cve20249675/Containerfile <<EOF
+FROM alpine
+RUN --mount=type=cache,from=testbuild,source=../,target=/var/tmp \
+ls -l /var/tmp && cat /var/tmp/file.txt
+EOF
+
+  run_buildah 1 build --security-opt label=disable --build-context testbuild=${TEST_SCRATCH_DIR}/cve20249675/ --no-cache ${TEST_SCRATCH_DIR}/cve20249675/
+  expect_output --substring "cat: can't open '/var/tmp/file.txt': No such file or directory"
```
nalind, line 7006: If I back out the changes to the sources (and comment out the lines 6986-6987), this part of the test still passes.  Any ideas why?

tests/bud.bats
mheon: Most likely too many or too few `../` in the path. I'll take a look...

nalind: I wouldn't expect too many to be an issue - `path_resolution(7)` says `One cannot walk up past the root: "/.." is the same as "/".`

tests/bud.bats
mheon: Even unpatched Buildah cannot escape the context directory using `../` traversal in `source=` apparently. Looking exactly where in the code the problem lies.

mheon: Same for line 6996. It seems this CVE is only reproducible if `id` is used as the vector. But I cannot locate the code preventing `source` from being used instead.

nalind: Oh, that's right: prepending the path separator to the supplied before passing it to `filepath.Clean()` would already have caused those to be removed.

nalind:
LGTM pending test results.
/approve

TomSweeneyRedHat:
Fixes: https://issues.redhat.com/browse/RHEL-61833
https://issues.redhat.com/browse/RHEL-61840

PR #5783, opened 2024-10-16
Integration tests: run git daemon on a random-but-bind()able port

nalind:
#### What type of PR is this?

/kind flake

Use a listener helper to bind to an available-according-to-the-kernel listening port and run a command with its stdio more or less tied to the connection instead of trying to launch a git daemon directly using a port number that we can only guess is available.

#### How to verify it

Our build-using-a-git-context tests should continue to pass.

None

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ endif
 #     Note: Uses the -N -l go compiler options to disable compiler optimizations
 #           and inlining. Using these build options allows you to subsequently
 #           use source debugging tools like delve.
-all: bin/buildah bin/imgtype bin/copy bin/tutorial docs
+all: bin/buildah bin/imgtype bin/copy bin/inet bin/tutorial docs
 
 # Update nix/nixpkgs.json its latest stable commit
 .PHONY: nixpkgs
@@ -109,6 +109,9 @@ bin/copy: $(SOURCES) tests/copy/copy.go
 
 bin/tutorial: $(SOURCES) tests/tutorial/tutorial.go
 	$(GO_BUILD) $(BUILDAH_LDFLAGS) -o $@ $(BUILDFLAGS) ./tests/tutorial/tutorial.go
+
+bin/inet: tests/inet/inet.go
+	$(GO_BUILD) $(BUILDAH_LDFLAGS) -o $@ $(BUILDFLAGS) ./tests/inet/inet.go
 
 .PHONY: clean
 clean:
--- a/rpm/buildah.spec
+++ b/rpm/buildah.spec
@@ -133,6 +133,7 @@ export BUILDTAGS+=" btrfs_noversion exclude_graphdriver_btrfs"
 %gobuild -o bin/imgtype ./tests/imgtype
 %gobuild -o bin/copy ./tests/copy
 %gobuild -o bin/tutorial ./tests/tutorial
+%gobuild -o bin/inet ./tests/inet
 %{__make} docs
 
 %install
@@ -143,6 +144,7 @@ cp -pav tests/. %{buildroot}/%{_datadir}/%{name}/test/system
 cp bin/imgtype %{buildroot}/%{_bindir}/%{name}-imgtype
 cp bin/copy    %{buildroot}/%{_bindir}/%{name}-copy
 cp bin/tutorial %{buildroot}/%{_bindir}/%{name}-tutorial
+cp bin/inet     %{buildroot}/%{_bindir}/%{name}-inet
 
 rm %{buildroot}%{_datadir}/%{name}/test/system/tools/build/*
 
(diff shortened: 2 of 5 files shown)


Luap99 (commented):
This looks like a bit overkill, in podman we have this long solved by using random_free_port() which seem to hold up fine there.
https://github.com/containers/podman/blob/740f1d1fc710d4560d2463b11333cc9ad68e22bd/test/system/helpers.network.bash#L306-L329

I think this is simpler for test issues to debug then dealing with custom written binary that can fail in a lot of unexpected ways. It will also means we would have to ship this binary in the test package for gating tests

Luap99:
Well debugging this custom go code may be easier for you but I doubt it is easier for the person running these tests on gating env that have to look at any failures and make sense of them such as @edsantiago.

In any case you are the main maintainer here and I am certainly not going to block this if it helps to solve a flake.

tests/inet/inet.go
```diff
@@ -0,0 +1,183 @@
...
+	portFile := ""
+	detach := false
+	flag.BoolVar(&detach, "detach", false, "detach from terminal")
+	flag.StringVar(&portFile, "port-file", "", "file to write listening port number")
+	flag.StringVar(&pidFile, "pid-file", "", "file to write process ID to")
+	flag.Parse()
+	args := flag.Args()
+	if len(args) < 1 {
+		fmt.Printf("Usage: %s [-port-file filename] [-pid-file filename] command ...\n", filepath.Base(os.Args[0]))
+		return
```
Luap99, line 34: should this exit(1)?

tests/inet/inet.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	// Start listening without specifying a port number.
+	ln, err := net.ListenTCP("tcp", &net.TCPAddr{})
```
Luap99, line 37: would it make sense to at least bind 127.0.0.1, there seems little reason to expose this externally

tests/inet/inet.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	// Retrieve the address we ended up bound to.
+	addr := ln.Addr()
+	addrString := addr.String()
+	colon := strings.LastIndex(addrString, ":")
+	if colon == -1 {
+		logrus.Fatalf("finding the start of the port number in %q", addrString)
+	}
+	// Write the port number part to the specified file, if one was specified.
+	portString := addrString[colon+1:]
```
Luap99, line 49: you can use `net.SplitHostPort()`

tests/inet/inet.go
nalind: Yeah, probably.  Changing it.

tests/inet/inet.go
nalind: Oh, nice.  Switching to it.

tests/inet/inet.go
nalind: Hmm, leaving it empty means we don't have to care about ipv4/ipv6 here, though.

rpm/buildah.spec
```diff
@@ -133,6 +133,7 @@ export BUILDTAGS+=" btrfs_noversion exclude_graphdriver_btrfs"
 %gobuild -o bin/imgtype ./tests/imgtype
 %gobuild -o bin/copy ./tests/copy
 %gobuild -o bin/tutorial ./tests/tutorial
+%gobuild -o bin/inet ./tests/inet
```
Luap99, line 136: this also needs to be added to the install section like `bin/copy` and then listed under `%files tests`

And since we had this discussion last week about podman-tests do we ship buildah-tests in RHEL as well? I really doubt we want to support these test helpers?

tests/inet/inet.go
Luap99: I mean fair enough, it should not be a problem though as all test connect to localhost which generally should try both `::1` and `127.0.0.1` so I doubt it causes issues.

Luap99 (commented):
go code LGTM

rpm/buildah.spec
nalind: Ah, right.  Adding.

Makefile
```diff
@@ -110,6 +110,9 @@ bin/copy: $(SOURCES) tests/copy/copy.go
 bin/tutorial: $(SOURCES) tests/tutorial/tutorial.go
 	$(GO_BUILD) $(BUILDAH_LDFLAGS) -o $@ $(BUILDFLAGS) ./tests/tutorial/tutorial.go
 
+bin/inet: tests/inet/inet.go
```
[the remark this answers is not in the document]
nalind: The scripts will need to be told where it is regardless, since it won't be in `../bin` relative to where the scripts end up packaged.  At that point it's the packager's decision, as it is with the other test helpers.

Luap99 (approved):
LGTM

PR #5798, opened 2024-10-24
Handle RUN --mount with relative targets and no configured workdir

### LINKED ISSUE #5738: relative mountpoints when building should use `/` as default if no WORKDIR is provided
**Description**

Related to https://github.com/containers/buildah/issues/4309.

While testing the [`Dockerfile.cpu](https://github.com/vllm-project/vllm/blob/fc990f97958636ce25e4471acfd5651b096b0311/Dockerfile.cpu) on https://github.com/vllm-project/vllm I tried building it, and got the following error

1. create a Containerfile
```Containerfile
FROM alpine:latest

RUN  --mount=type=bind,src=requirements-build.txt,target=requirements-build.txt \
    cat requirements-build.txt
```

2. create a `requirements-build.txt` 
3. build the image
4. assert error

We can fix the error by adding `WORKDIR /` before the RUN.

```
$: podman build -f Dockerfile.cpu -t vllm-cpu-env --shm-size=4g .
Error: building at STEP "RUN --mount=type=cache,target=/root/.cache/pip --mount=type=bind,src=requirements-build.txt,target=requirements-build.txt pip install --upgrade pip &&     pip install -r requirements-build.txt": resolving mountpoints for container "3a97f46183fa64e10c96f20f9a38a5ed46d2e9e7c4e7bbfbce6fa1adfdacd66e": invalid container path "requirements-build.txt", must be an absolute path
```

It should probably use `/` as default workdir

```
podman version 5.2.2
```

**Output of `uname -a`:**

```
Linux fedora.home 6.10.8-200.fc40.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Sep  4 21:41:11 UTC 2024 x86_64 GNU/Linux
```

nalind:
#### What type of PR is this?

/kind bug

When the target location of a RUN --mount is specified as a relative path, we normally try to convert it to an absolute path by combining it with the currently-configured working directory.  If there is no such value, though, the result is still not an absolute path.  Work around this by using "/" when the configured working directory is "".
    
Set this field in the `runMountInfo` struct on FreeBSD, as we already did on Linux.

#### How to verify it

New integration test!

Fixes #5738

```release-note
RUN instructions which use the --mount flag with a relative "target" path no longer trigger an error if the current stage does not have a defined working directory.
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -7017,3 +7017,17 @@ EOF
   run_buildah 1 build --security-opt label=disable --build-context testbuild=${TEST_SCRATCH_DIR}/cve20249675/ --no-cache ${TEST_SCRATCH_DIR}/cve20249675/
   expect_output --substring "cat: can't open '/var/tmp/file.txt': No such file or directory"
 }
+
+@test "build-mounts-implicit-workdir" {
+  base=busybox
+  _prefetch $base
+  run_buildah inspect --format '{{.Docker.Config.WorkingDir}}' --type=image $base
+  expect_output "" "test base image needs to not have a default working directory defined in its configuration"
+  # check that the target for a bind mount can be specified as a relative path even when there's no WorkingDir defined for it to be relative to
+  echo FROM $base > ${TEST_SCRATCH_DIR}/Containerfile
+  echo RUN --mount=type=bind,src=Containerfile,target=Containerfile test -s Containerfile >> ${TEST_SCRATCH_DIR}/Containerfile
+  echo RUN --mount=type=cache,id=cash,target=cachesubdir truncate -s 1024 cachesubdir/cachefile >> ${TEST_SCRATCH_DIR}/Containerfile
+  echo RUN --mount=type=cache,id=cash,target=cachesubdir2 test -s cachesubdir2/cachefile >> ${TEST_SCRATCH_DIR}/Containerfile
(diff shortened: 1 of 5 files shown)


tests/bud.bats
```diff
@@ -7017,3 +7017,16 @@ EOF
...
+  base=busybox
+  _prefetch $base
+  run_buildah inspect --format '{{.Docker.Config.WorkingDir}}' --type=image $base
+  expect_output "" "test base image needs to not have a default working directory defined in its configuration"
+  # check that the target for a bind mount can be specified as a relative path even when there's no WorkingDir defined for it to be relative to
+  echo FROM $base > ${TEST_SCRATCH_DIR}/Containerfile
+  echo RUN --mount=type=bind,src=Containerfile,target=Containerfile pwd >> ${TEST_SCRATCH_DIR}/Containerfile
+  echo RUN --mount=type=cache,target=cachesubdir pwd >> ${TEST_SCRATCH_DIR}/Containerfile
+  echo RUN --mount=type=tmpfs,target=tmpfssubdir pwd >> ${TEST_SCRATCH_DIR}/Containerfile
+  run_buildah build ${TEST_SCRATCH_DIR}
```
Luap99, line 7031: I feel like this should actually verify that these mounts are actually mounted at the proper expected target?

Luap99 (commented):
code LGTM but I think the test should actually check that they get mounted correctly

tests/bud.bats
nalind: Updated.

Luap99 (approved):
LGTM

TomSweeneyRedHat:
LGTM
but it looks like a rebase is needed @nalind

nalind:
Rebased.

PR #5801, opened 2024-10-29
tests: blobcache: use unique image name

edsantiago:
Prep work for parallelizing bats tests: when pushing images to
local registry, use a pseudorandom unique name to avoid
possibility of collisions/conflicts with other tests. This
is good practice regardless of whether we run serial or parallel.

Also slight refactor of some duplicate code.

```release-note
None
```

### DIFF
--- a/tests/blobcache.bats
+++ b/tests/blobcache.bats
@@ -58,6 +58,8 @@ function _check_matches() {
 # Integration test for https://github.com/containers/image/pull/1645
 @test "blobcache: blobs must be reused when pushing across registry" {
 	start_registry
+
+	imgname=blobimg-$(random_string | tr A-Z a-z)
 	run_buildah login --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth --username testuser --password testpassword localhost:${REGISTRY_PORT}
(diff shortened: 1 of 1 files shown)


tests/blobcache.bats
```diff
@@ -66,26 +68,26 @@ function _check_matches() {
...
-		assert "$status" -eq 0 "status of `run rm $HOME/.local/share/containers/cache/blob-info-cache-v1.sqlite` must be 0"
-	else
-		run rm /var/lib/containers/cache/blob-info-cache-v1.sqlite
-		assert "$status" -eq 0 "status of `run rm /var/lib/containers/cache/blob-info-cache-v1.sqlite` must be 0"
-	fi
+		cachedir=$HOME/.local/share
+        fi
+        cachefile=$cachedir/containers/cache/blob-info-cache-v1.sqlite
+        run rm $cachefile
+	assert "$status" -eq 0 "status of `run rm $cachefile` must be 0"
```
Luap99, line 84: here as well

edsantiago:
Indentation fixed, and repushed. This particular file uses tab indentation (`^I`).

Luap99 (approved):
LGTM

PR #5803, opened 2024-10-29
tests: sbom: never write to cwd

edsantiago:
Write output files only to $TEST_SCRATCH_DIR, never to cwd

Reformat overly-long lines for readability (whitespace only changes)

And, in the last test, the output files are plaintext, not json.
Fix filenames accordingly.

```release-note
None
```


Luap99 (approved):
LGTM

PR #5804, opened 2024-10-29
tests: bud: make parallel-safe

edsantiago:
- all images pushed to a local registry must have a unique name

- all cache tests must use a private TMPDIR

- in force-compression test, use a custom-crafted image with
  no possibility of sharing layers with any other image that
  any other test might push to the registry.

- use a private crun tmpdir in seccomp test, because crun
  does some unexpected caching.

And, forgive me, a little refactoring of unpleasant duplication

```release-note
None
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -15,12 +15,14 @@ load helpers
 @test "bud: build manifest list and --add-compression zstd" {
   start_registry
   run_buildah login --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth --username testuser --password testpassword localhost:${REGISTRY_PORT}
-  run_buildah build $WITH_POLICY_JSON -t image1 --platform linux/amd64 -f $BUDFILES/dockerfile/Dockerfile
-  run_buildah build $WITH_POLICY_JSON -t image2 --platform linux/arm64 -f $BUDFILES/dockerfile/Dockerfile
+
+  imgname="img-$(safename)"
+  run_buildah build $WITH_POLICY_JSON -t "${imgname}1" --platform linux/amd64 -f $BUDFILES/dockerfile/Dockerfile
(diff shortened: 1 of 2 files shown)


tests/bud.bats
```diff
@@ -15,12 +15,14 @@ load helpers
 @test "bud: build manifest list and --add-compression zstd" {
   start_registry
   run_buildah login --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth --username testuser --password testpassword localhost:${REGISTRY_PORT}
-  run_buildah build $WITH_POLICY_JSON -t image1 --platform linux/amd64 -f $BUDFILES/dockerfile/Dockerfile
-  run_buildah build $WITH_POLICY_JSON -t image2 --platform linux/arm64 -f $BUDFILES/dockerfile/Dockerfile
+
+  imgname="img$(random_string | tr A-Z a-z)"
```
Luap99, line 19: q: would the safename helper from podman make sense for the parallel work here or do we not need it because the tests do not share storage?

Luap99 (approved):
LGTM

tests/bud.bats
edsantiago: I almost went with `safename` but couldn't think of any real-world situation in which it would help: the errors I've seen in testing only show layer IDs. It's not much work to bring it in. Should I?

Luap99: The one thing would be that we can skip the ` | tr A-Z a-z` that is duplicated in so many places. But that could also be fixed by random string only returning lowercase. I am not sure what would be better.
Either way not a blocker.

TomSweeneyRedHat (commented):
LGTM

Luap99 (approved):
/lgtm

edsantiago:
New push, added `safename`

PR #5807, opened 2024-10-31
Makefile: list sources via `find` conditionally

danishprakash:
Taken directly from podman's Makefile. The current list doesn't list vendor sources and therefore if you need to build and test your changes to vendor/ for testing a backport or anything else, it's required to modify the Makefile before attempting to do so.

/kind other

Signed-off-by: Danish Prakash <contact@danishpraka.sh>


danishprakash:
@nalind I noticed you've added certain sources over time, so if there's a reason to selectively list sources that I'm not aware of, I can also only add `vendor/*`, that should also be good enough.

nalind:
I hadn't considered that modifying files under `vendor` was a common enough case that the Makefile needed to account for it.  Trying to add everything under `vendor` into the current SOURCES list seems like something that the bot that keeps our dependencies up to date would trip over, so if this doesn't slow down `make`, then I guess the thing to do is remove the part of the smoke test that fails.

danishprakash:
> I hadn't considered that modifying files under vendor was a common enough case that the Makefile needed to account for it.

I mean I use it quite often w/ podman when testing c/* changes and it's come in quite handy.

---

> then I guess the thing to do is remove the part of the smoke test that fails.

We were validating the sources _because_ we were listing them manually, I guess moving to a finder cmd, I don't see the need for a validation anymore, wdyt? 

Do we have benchmarks to test if there's a performance degradation? I can think of vendor bumps as an additional trigger for rebuilds, can't think of anything major that should have an impact.

Makefile
```diff
@@ -41,7 +41,13 @@ LIBSECCOMP_COMMIT := release-2.3
 
 EXTRA_LDFLAGS ?=
 BUILDAH_LDFLAGS := $(GO_LDFLAGS) '-X main.GitCommit=$(GIT_COMMIT) -X main.buildInfo=$(SOURCE_DATE_EPOCH) -X main.cniVersion=$(CNI_COMMIT) $(EXTRA_LDFLAGS)'
-SOURCES=*.go imagebuildah/*.go bind/*.go chroot/*.go copier/*.go define/*.go docker/*.go internal/config/*.go internal/mkcw/*.go internal/mkcw/types/*.go internal/parse/*.go internal/sbom/*.go internal/source/*.go internal/tmpdir/*.go internal/*.go internal/util/*.go internal/volumes/*.go manifests/*.go pkg/binfmt/*.go pkg/blobcache/*.go pkg/chrootuser/*.go pkg/cli/*.go pkg/completion/*.go pkg/formats/*.go pkg/jail/*.go pkg/overlay/*.go pkg/parse/*.go pkg/rusage/*.go pkg/sshagent/*.go pkg/umask/*.go pkg/util/*.go pkg/volumes/*.go util/*.go
+
+# This isn't what we actually build; it's a superset, used for target
+# dependencies. Basically: all *.go and *.c files, except *_test.go,
+# and except anything in a dot subdirectory. If any of these files is
+# newer than our target (bin/podman{,-remote}), a rebuild is
```
Luap99, line 48: the bind/podman comment is not accurate here

Luap99 (commented):
If a go file changed we should rebuild, in vendor/ or not. Because if I do a dependency bump locally via gp.mod then make buildah currently doesn't rebuild so this is clearly broken.  As such I think this here is the right idea.

I don't see any advantage of listing all the packages here so the find seems much better to me. Even if it would pick something up that does not need to trigger a rebuild all the time go rebuilds are fast anyway so I would not be concerned about speed.

rhatdan:
LGTM

kolyshkin (changes requested):
You should probably also remove *.go files from some of the targets that use `$(SOURCES). For example:

```make
bin/buildah: $(SOURCES) cmd/buildah/*.go internal/mkcw/embed/entrypoint_amd64.gz
````

Here `cmd/buildah/*.go` is becoming redundant.

PR #5820, opened 2024-11-06
CI VMs: bump f40 -> f41

edsantiago:
Images built in https://github.com/containers/automation_images/pull/392

```release-note
None
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -27,12 +27,12 @@ env:
     ####
     # GCE project where images live
     IMAGE_PROJECT: "libpod-218412"
-    FEDORA_NAME: "fedora-40"
-    PRIOR_FEDORA_NAME: "fedora-39"
+    FEDORA_NAME: "fedora-41"
+    PRIOR_FEDORA_NAME: "fedora-40"
     DEBIAN_NAME: "debian-13"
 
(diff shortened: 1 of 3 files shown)


Luap99:
We will need to update the linter golangci-lint needs to be > 1.60 to work on go 1.23 AFAICT

Luap99:
try picking https://github.com/containers/buildah/pull/5821 in here

edsantiago:
ulimits failing. Probably the same thing as  https://github.com/containers/podman/pull/24335 but not caught because we don't run buildah tests on rawhide.

Luap99:
> ulimits failing. Probably the same thing as [containers/podman#24335](https://github.com/containers/podman/pull/24335) but not caught because we don't run buildah tests on rawhide.

The test seem to call podman to figure out the limit podman uses. IMO this sounds super wrong, buildah cannot depend on the host podman version and expect things to never change... added in https://github.com/containers/buildah/commit/79fb3c2135db1a9f962ad3ce42e97b20867e9fa6

What we could test on podman is that podman run/build use the same limits which make sense to me but testing in buildah with a unknown host binary cannot work

Luap99 (approved):
LGTM

PR #5821, opened 2024-11-07
tests/tools: update golangci-lint to v1.61.0

Luap99:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```


Luap99:
@nalind @rhatdan PTAL

tests/tools/go.mod
```diff
@@ -1,25 +1,25 @@
 module github.com/containers/buildah/tests/tools
 
-go 1.21
+go 1.22.1
```
nalind, line 3: This is newer than the version the `go.mod` in the top-level directory specifies.  Is that going to create any problems?

Luap99: No, I mean well if you want to build these tools you need go 1.22.1 and keep in mind we no longer have any control over this. It is all handled by the golang module management. And as soon as we vendor c/image we have to bump the main module as well
https://github.com/containers/image/blob/59417aed3db19c66039e72b24432d69a34e911e2/go.mod#L3

The only real choice we have is to not update the dependency that bumps the minimum version but I don't see this is a problem here.

nalind: Okay, then.

PR #5833, opened 2024-11-09
CI VMs: bump again

edsantiago:
Images built in https://github.com/containers/automation_images/pull/393

Very minor and safe. Purpose is to stabilize those VM images, which were needed for podman. They include a 6.11.6 kernel which fixes one bug, and newer pasta in f39 which fixes another bug.

```release-note
None
```


Luap99 (approved):
/lgtm

PR #5841, opened 2024-11-18
Tests: make _prefetch() parallel-safe

edsantiago:
The _prefetch helper, introduced in #2036, is not parallel-safe: two
or more parallel jobs fetching the same image can step on each other
and produce garbage images.

Although we still can't run buildah tests in parallel (see #5552),
we can at least set up the scaffolding for that to happen. This
commit reworks _prefetch() such that the image work is wrapped
inside flock. It has been working fine for months in #5552,
and is IMO safe for production. This can then make it much
easier to flip the parallelization switch once the final zstd
bug is squashed.

```release-note
None
```

### DIFF
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -172,9 +172,11 @@ function normalize_image_name() {
 
 function _prefetch() {
     if [ -z "${_BUILDAH_IMAGE_CACHEDIR}" ]; then
-        _pgid=$(sed -ne 's/^NSpgid:\s*//p' /proc/$$/status)
-        export _BUILDAH_IMAGE_CACHEDIR=${BATS_TMPDIR}/buildah-image-cache.$_pgid
+        export _BUILDAH_IMAGE_CACHEDIR=${BATS_SUITE_TMPDIR}/buildah-image-cache
         mkdir -p ${_BUILDAH_IMAGE_CACHEDIR}
+
+        # It's 700 by default; that prevents 'unshare' from reading cached images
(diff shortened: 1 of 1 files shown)


siteshwar:
This was pointed to you [earlier](https://github.com/containers/buildah/pull/5552#issuecomment-2462361987). Changes in this pull request introduce a potentially dangerous statement that [shows up](https://openscanhub.fedoraproject.org/task/24560/log/added.html) in shellcheck:

```
Error: SHELLCHECK_WARNING (CWE-398): [#def1]
/usr/share/buildah/test/system/helpers.bash:211:24: warning[SC2115]: Use "${var:?}" to ensure this never expands to / .
#  209|               else
#  210|                   # Failed. Clean up, so we don't leave incomplete remnants
#  211|->                 rm -fr $_BUILDAH_IMAGE_CACHEDIR/$fname
#  212|               fi
#  213|               sleep 5
```

You are assuming both of these variables are always set. It should be fixed here and at other places before merging.

edsantiago:
>You are assuming both of these variables are always set

Fixed, thank you

TomSweeneyRedHat:
Changed LGTM, but a rebase is needed @edsantiago

edsantiago:
Is there a way to shut up the anti-helpful "branch is out-of-date" message in this repo?

Luap99 (approved):
/lgtm

Luap99:
> Is there a way to shut up the anti-helpful "branch is out-of-date" message in this repo?

Yes these are settings in the branch protection github settings AFAIK.

PR #5845, opened 2024-11-22
Inherit arguments between stages

### LINKED ISSUE #5762: ARG inheritance works in Docker, but not Podman
**Description**

Docker provides limited inheritance of Dockerfile ARGs between build stages, as described here: https://docs.docker.com/build/building/variables/#scoping

> Once a build argument is declared or consumed in a stage, it's automatically inherited by child stages.

This contradicts the Dockerfile reference https://docs.docker.com/reference/dockerfile/#scope - which states:

> An ARG instruction goes out of scope at the end of the build stage where it was defined. To use an argument in multiple stages, each stage must include the ARG instruction.

Podman behaves according to the latter statement, making the examples given in the other document incompatible.

Try to build a Containerfile with the following contents (copied from [the Docker manual](https://docs.docker.com/build/building/variables/#scoping)):

```docker
# syntax=docker/dockerfile:1
FROM alpine AS base
# Declare the build argument in the build stage
ARG NAME="joe"

# Create a new stage based on "base"
FROM base AS build
# The NAME build argument is available here
# since it's declared in a parent stage
RUN echo "hello $NAME!"
```

`podman build` prints:
```
[2/2] STEP 2/2: RUN echo "hello $NAME!"
hello !
```

`docker build` would print:
```
[2/2] STEP 2/2: RUN echo "hello $NAME!"
hello joe!
```

```
buildah/testing,unstable,now 1.37.2+ds1-3 amd64 [installed,automatic]
```

```
Version:         1.37.2
Go Version:      go1.22.7
Image Spec:      1.1.0
Runtime Spec:    1.2.0
CNI Spec:        1.0.0
libcni Version:  
image Version:   5.32.2
Git Commit:      
Built:           Thu Jan  1 01:00:00 1970
OS/Arch:         linux/amd64
BuildPlatform:   linux/amd64
```

```
Client:       Podman Engine
Version:      5.2.2
API Version:  5.2.2
Go Version:   go1.23.1
Built:        Thu Jan  1 01:00:00 1970
OS/Arch:      linux/amd64
```

```
PRETTY_NAME="Debian GNU/Linux trixie/sid"
NAME="Debian GNU/Linux"
VERSION_CODENAME=trixie
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

**Output of `uname -a`:**

```
Linux 6.10.9-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.10.9-1 (2024-09-08) x86_64 GNU/Linux
```

```
```

onitake:
#### What type of PR is this?

/kind feature

Docker supports argument inheritance between build stages, but this feature is missing from buildah.

#### How to verify it

See the linked issue below for a basic example.

Fixes #5762 

This PR only contains a proposed code change to make argument inheritance work.
The feature also needs to be documented.

```release-note
ARG statements will now be inherited between build stages, like docker-build does.
```


onitake:
`maps.Insert` and `maps.All` for merging one map into another require Go 1.23.
Is there something comparable in the buildah source that I could use?

nalind:
Solving this here doesn't solve it for imagebuilder's CLI tool.  I'm not sure what I think about that yet.

onitake:
@nalind It's still incomplete and doesn't seem to work how I imagined it yet anyway.
But, what's imagebuilder? I though podman's image builder is buildah?

nalind:
Ah, sorry, that reference was buried somewhere in the linked issue.  When I talk about "imagebuilder", I mean https://github.com/openshift/imagebuilder.

The Dockerfile parsing logic mostly lives in `github.com/openshift/imagebuilder/dockerfile`.  The `github.com/openshift/imagebuilder.Builder` object "drives" a build using an implementation of its `Executor` interface.  imagebuilder provides a CLI tool that combines its main library with `github.com/openshift/imagebuilder/dockerclient.ClientExecutor`, which implements the interface using `github.com/fsouza/go-dockerclient`.

Buildah originally didn't handle Dockerfiles at all, and when we needed to add the ability to do so, implementing a `github.com/openshift/imagebuilder.Executor` using `github.com/containers/buildah.Builder` and using a `github.com/openshift/imagebuilder.Builder` to "drive" builds seemed relatively straightforward.

At times, this has meant that we had to make changes there in order to enable features here, but I could say that about other dependencies we have as well.

onitake:
Ah sorry, now I get what you meant. Yes, I had originally thought I could/should place all argument handling logic there. But then I realized that arguments are only properly resolved and tracked when build stages are "executed", which all happens here in the buildah source.

Does that mean you would prefer that argument inheritance rules in Dockerfiles is handled by openshift/imagebuilder instead? I'm not sure if that is feasible at all, at least I think it would duplicate a lot of things.

rhatdan:
@onitake any update on this?

onitake:
> @onitake any update on this?

I was struggling a bit to set up a proper test environment, so I haven't been able to debug why it's not working as expected yet. Will try to submit an update later this week.

PR #5854, opened 2024-12-02
manifest add: add --artifact-annotation

nalind:
#### What type of PR is this?

/kind feature

Add a --artifact-annotation flag to `buildah manifest add` that can be used to set annotations in the artifact manifest that we generate and then add to an image index.

#### How to verify it

Updated integration test!

None

```release-note
`buildah manifest add` now includes a `--artifact-annotation` flag which can be used to add an annotation to the artifact manifest which is generated and then added to the image index.
```


TomSweeneyRedHat:
LGTM
but needs a rebase.

PR #5856, opened 2024-12-04
CI: remove some inter-job dependencies, run cross-compile task with make -j, use /tmp for Go build cache

nalind:
#### What type of PR is this?

/kind other

As a quick attempt to speed up our CI's total start-to-finish time, this hits what we think will be low-hanging fruit.

#### How to verify it

Check the clock duration for https://cirrus-ci.com/github/containers/buildah/pull%2F5856 compared to builds for other pull requests.

None

```release-note
None
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -97,7 +97,7 @@ smoke_task:
         cpu: 4
 
     # Don't bother running on branches (including cron), or for tags.
-    only_if: $CIRRUS_PR != ''
+    skip: $CIRRUS_PR == ''
 
     timeout_in: 30m
 
@@ -140,8 +140,8 @@ cross_build_task:
         cpu: 8
         memory: "24G"
     alias: cross_build
-    only_if: >-
-        $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*'
+    skip: >-
+        $CIRRUS_CHANGE_TITLE =~ '.*CI:DOCS.*'
     env:
         HOME: /root
     script:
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -135,15 +136,18 @@ vendor_task:
...
     only_if: >-
         $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*'
     env:
         HOME: /root
     script:
         - go version
-        - make cross CGO_ENABLED=0
-
+        - make -j cross CGO_ENABLED=0
+        - du -sh `go env GOCACHE`
```
Luap99, line 150: I think we can drop the du or do you still need that?

.cirrus.yml
```diff
@@ -170,6 +170,7 @@ unit_task:
     setup_script: '${SCRIPT_BASE}/setup.sh |& ${_TIMESTAMP}'
     build_script: '${SCRIPT_BASE}/build.sh |& ${_TIMESTAMP}'
     unit_test_script: '${SCRIPT_BASE}/test.sh unit |& ${_TIMESTAMP}'
+    du_script: du -sh `go env GOCACHE`
```
Luap99, line 173: same here and in the other places?

.cirrus.yml
```diff
@@ -154,10 +158,6 @@ unit_task:
     only_if: &not_build_docs >-
         $CIRRUS_CHANGE_TITLE !=~ '.*CI:DOCS.*' &&
         $CIRRUS_CHANGE_TITLE !=~ '.*CI:BUILD.*'
-    depends_on: &smoke_vendor_cross
-      - smoke
-      - vendor
-      - cross_build
```
Luap99, line 160: I wonder if we should at least keep the dependency for vendor & smoke test as those are fast enough.

In podman we have basically two stages build/validate -> actual tests:
https://raw.githubusercontent.com/wiki/containers/podman/cirrus-map.svg

So I think it makes sense for buildah to follow that, unless we really like to safe the the 7 mins form the smoke test?

.cirrus.yml
```diff
@@ -93,7 +94,7 @@ smoke_task:
     name: "Smoke Test"
 
     gce_instance:
-        memory: "12Gb"
```
Luap99, line 96: pre existing but defining a task with no cpu seems odd

Luap99 (commented):
Thanks mostly looks good. That is a great improvement.

.cirrus.yml
nalind: I'm concerned about using /tmp for the Go cache, but you're probably right that we don't need to be worrying about that in every task.

.cirrus.yml
nalind: Sure.  We'll try 4, maybe it'll run this task a bit faster.

.cirrus.yml
nalind: Dropped it.

.cirrus.yml
nalind: Having added depends_on bits back, it looks like we're getting just about everything else stalled until the cross-build check finishes.  Replacing some of the "only_if" directives with "skip" directives looks like it will avoid this weirdness.

PR #5874, opened 2024-12-12
chroot: on Linux, try to pivot_root before falling back to chroot

nalind:
Unless --no-pivot or the equivalent API flag is set, try to pivot_root() to enter the rootfs during Run().  Fall back to using chroot() as before if that fails for any reason.

#### What type of PR is this?

/kind design

When using chroot isolation, if we can pivot_root() instead of using chroot() for running the command, do that, because it's closer to what a proper runtime can do.

#### How to verify it

Extended unit tests!

None

We still need chroot() for things like unpacking layers while pulling images, so if you're hoping this means we no longer ever need CAP_SYS_CHROOT, sorry.

```release-note
`buildah build` or `buildah run` with `--isolation chroot` will now default to using pivot_root() internally, unless the `--no-pivot` CLI flag is used.
```

### DIFF
--- a/chroot/run_test.go
+++ b/chroot/run_test.go
@@ -36,7 +36,7 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
-func testMinimal(t *testing.T, modify func(g *generate.Generator, rootDir, bundleDir string), verify func(t *testing.T, report *types.TestReport)) {
+func testMinimalWithPivot(t *testing.T, noPivot bool, modify func(g *generate.Generator, rootDir, bundleDir string), verify func(t *testing.T, report *types.TestReport)) {
 	t.Helper()
 	g, err := generate.New("linux")
 	if err != nil {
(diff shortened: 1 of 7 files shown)


rhatdan:
LGTM
@giuseppe @flouthoc @mtrmac PTAL

giuseppe (approved):
/lgtm

chroot/run_test.go
```diff
@@ -100,8 +100,8 @@ func testMinimal(t *testing.T, modify func(g *generate.Generator, rootDir, bundl
 	}
 
 	output := new(bytes.Buffer)
-	if err := RunUsingChroot(g.Config, bundleDir, "/", new(bytes.Buffer), output, output); err != nil {
-		t.Fatalf("run: %v: %s", err, output.String())
+	if err := RunUsingChroot(g.Config, bundleDir, "/", new(bytes.Buffer), output, output, noPivot); err != nil {
+		t.Fatalf("run(noPivot=false): %v: %s", err, output.String())
```
nalind, line 104: Ugh, just spotted the literal "=false" instead of using the passed-in value, opened #5876 to fix it.

mupuf:
FYI, this regressed running `systemctl enable` in a container built using ansible inside buildah: the first invocation would work but subsequent would fail.

If you want a reproducer, you may clone ci-tron at commit https://gitlab.freedesktop.org/gfx-ci/ci-tron/-/commit/ffe8d5841f8b319e7072ec549c71f9c929335de9, then run `$ make gateway`.

PR #5901, opened 2025-01-07
copy-preserving-extended-attributes: use a different base image

nalind:
#### What type of PR is this?

/kind failing-test 

The Fedora 39-based image this test is using is having trouble finding mirrors from CI.  Use a different one that can still provide the setcap and setfattr commands.

#### How to verify it

The updated integration test should pass.

None

```release-note
None
```


flouthoc (approved):
LGTM

PR #5922, opened 2025-01-20
relabel(): correct a misleading parameter name

nalind:
#### What type of PR is this?

/kind cleanup

The "recurse" parameter of the relabel() function signature is passed to a function which calls it "shared".  There's no need for it to not have the same name.

#### How to verify it

This shouldn't change any behavior.

None

This one's more about making sure the recently-merged security fixes get a proper run-through on CI than anything else.

```release-note
None
```


flouthoc (approved):
LGTM
/lgtm
/approve

PR #5924, opened 2025-01-20
Run(): always clean up options.ExternalImageMounts

nalind:
#### What type of PR is this?

/kind cleanup

Make sure that we consistently unmount the list of images that we're told to, even in cases where we return an error before arranging to do so in `Run()`.

None

```release-note
None
```


flouthoc (approved):
LGTM

PR #5925, opened 2025-01-20
Add more checks to the --mount flag parsing logic

nalind:
#### What type of PR is this?

/kind other

* Make `volumes.GetBindMount()`, `volumes.GetCacheMount()`, and `volumes.GetTmpfsMount()` return errors when flags which expect arguments are given empty arguments, when flags which don't expect arguments are given arguments, and when the "relabel" flag, which expects an argument, doesn't get one.
* Make `volumes.GetCacheMount()` not treat the "U" flag as affecting bind propagation.
* Drop the special-case error message when a caller attempts to use "src" or "source" options in `volumes.GetTmpfsMount()`, which would already be covered by the general-purpose "unrecognized option" default.

New unit tests!

```release-note
None
```

### DIFF
--- a/internal/volumes/volumes.go
+++ b/internal/volumes/volumes.go
@@ -47,11 +47,12 @@ const (
 )
 
 var (
-	errBadMntOption  = errors.New("invalid mount option")
-	errBadOptionArg  = errors.New("must provide an argument for option")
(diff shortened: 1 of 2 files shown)


flouthoc (approved):
LGTM
/lgtm
/approve

PR #5926, opened 2025-01-20
chroot mount flags integration test: copy binaries

nalind:
#### What type of PR is this?

/kind other

Copy the binaries that we call from inside of a UID-mapped `unshare` to a location where the permissions on the parent directories that contain them are less likely to render them inaccessible, and use the copies.

#### How to verify it

Updated integration test!

None

```release-note
None
```

### DIFF
--- a/tests/chroot.bats
+++ b/tests/chroot.bats
@@ -87,6 +87,10 @@ load helpers
       done
     done
   done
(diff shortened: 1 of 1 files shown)


flouthoc (approved):
LGTM

PR #5927, opened 2025-01-20
pkg/overlay: cleanups

nalind:
#### What type of PR is this?

/kind cleanup

Change `generateOverlayStructure()` to not return its first argument unchanged, since both of its callers already have that value, and adjust a few error messages.

In the Linux `MountWithOptions()`, ensure that, if `UpperDirOptionFragment` and `WorkDirOptionFragment` values were specified, they are absolute paths, otherwise place them under the top-level parent of the various directories we have for this mount.

Update a number of comments.

#### How to verify it

No functional changes.

None

I had so much trouble understanding how to use this package, maybe this will make it easier for the next person, who might be me.

```release-note
None
```

### DIFF
--- a/pkg/overlay/overlay.go
+++ b/pkg/overlay/overlay.go
@@ -17,9 +17,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// Options type holds various configuration options for overlay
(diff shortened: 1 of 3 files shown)


flouthoc (approved):
LGTM

PR #5931, opened 2025-01-22
[release-1.26] backport fix for CVE-2024-11218

dashea:
#### What type of PR is this?

/kind bug

Backport the changes for https://github.com/containers/buildah/security/advisories/GHSA-5vpc-35f4-r8w6 to the 1.26 branch.

[RHEL-67609](https://issues.redhat.com/browse/RHEL-67609)

```release-note
None
```


nalind:
This branch predates the splitup in pkg/overlay, so I don't think we'll need 44e2788767871a7bbd48bc4a5500a18720918b15

nalind:
I think we're missing an equivalent for https://github.com/containers/buildah/commit/7320f490a2e0ff1b4b9f76758511e464bc8fa293#diff-88267013462361b7dd8d8bde4c038ce82c090a9c72936d14bcd6c368aa91323bR526-R530 from 8673136e02f6d6065d8896262d661150c5861ec6

dashea:
> I think we're missing an equivalent for [7320f49#diff-88267013462361b7dd8d8bde4c038ce82c090a9c72936d14bcd6c368aa91323bR526-R530](https://github.com/containers/buildah/commit/7320f490a2e0ff1b4b9f76758511e464bc8fa293#diff-88267013462361b7dd8d8bde4c038ce82c090a9c72936d14bcd6c368aa91323bR526-R530) from [8673136](https://github.com/containers/buildah/commit/8673136e02f6d6065d8896262d661150c5861ec6)

Thanks, added

internal/parse/parse.go
```diff
@@ -43,16 +47,82 @@ var (
...
+	}
+	return rw
+}
+
+func convertToOverlay(m specs.Mount, store storage.Store, mountLabel, tmpDir string, uid, gid int) (specs.Mount, string, error) {
+	overlayDir, err := overlay.TempDir(tmpDir, uid, gid)
+	if err != nil {
+		return specs.Mount{}, "", fmt.Errorf("setting up overlay for %q: %w", m.Destination, err)
+	}
+	graphOptionsCopy := store.GraphOptions()
```
nalind, line 71: `GraphOptions()` returns a slice, and assignment doesn't create a new underlying array, which is part of what makes slices faster than they might otherwise be.  You'd need something along the lines of what `copyStringSlice()` from our top-level `utils.go` does to do the equivalent of `slices.Clone()`.

run_linux.go
```diff
@@ -523,10 +522,18 @@ func (b *Builder) setupMounts(context *imagetypes.SystemContext, mountPoint stri
 
 	// Get the list of mounts that are just for this Run() call.
 	// TODO: acui: de-spaghettify run mounts
-	runMounts, mountArtifacts, err := b.runSetupRunMounts(context, runFileMounts, secrets, stageMountPoints, sshSources, cdir, contextDir, spec.Linux.UIDMappings, spec.Linux.GIDMappings, int(rootUID), int(rootGID), int(processUID), int(processGID))
+	runMounts, mountArtifacts, err := b.runSetupRunMounts(context, bundlePath, runFileMounts, secrets, stageMountPoints, sshSources, cdir, contextDir, spec.Linux.UIDMappings, spec.Linux.GIDMappings, int(rootUID), int(rootGID), int(processUID), int(processGID))
 	if err != nil {
 		return nil, err
 	}
+	succeeded := false
```
nalind, line 529: `succeeded` needs to be set to `true` around line 567.

run_linux.go
```diff
@@ -2635,18 +2702,37 @@ func (b *Builder) getTmpfsMount(tokens []string, rootUID, rootGID, processUID, p
...
+				}
+			}
+		}
+	}()
+	optionMounts = append(optionMounts, optionMount)
 	volumes, err := b.runSetupVolumeMounts(b.MountLabel, nil, optionMounts, rootUID, rootGID, processUID, processGID)
 	if err != nil {
-		return nil, lockedTargets, err
+		return nil, "", lockedTargets, err
 	}
```
nalind, line 2734: Need to set `succeeded` to `true` before returning.

run_linux.go
```diff
@@ -2635,18 +2702,37 @@ func (b *Builder) getTmpfsMount(tokens []string, rootUID, rootGID, processUID, p
...
+	succeeded := false
+	defer func() {
+		if !succeeded {
+			if intermediateMount != "" {
+				if err := mount.Unmount(intermediateMount); err != nil {
+					b.Logger.Debugf("unmounting %q: %v", intermediateMount, err)
+				}
+				if err := os.Remove(intermediateMount); err != nil {
+					b.Logger.Debugf("removing should-be-empty directory %q: %v", intermediateMount, err)
+				}
```
nalind, line 2726: We should also be freeing the locks in the `lockedTargets` slice, similar to the way `cleanupRunMounts()` in `run_linux.go` does, and this function should refrain from returning `lockedTargets` along error paths once it starts taking care of them.

run_linux.go
nalind: (Right after this block, before the `return` statement.)

internal/parse/parse.go
```diff
@@ -511,52 +704,64 @@ func getMounts(ctx *types.SystemContext, store storage.Store, mounts []string, c
...
 			if _, ok := finalMounts[mount.Destination]; ok {
-				return nil, mountedImages, lockedTargets, errors.Wrapf(errDuplicateDest, mount.Destination)
+				return nil, nil, nil, nil, lockedTargets, errors.Wrapf(errDuplicateDest, mount.Destination)
 			}
 			finalMounts[mount.Destination] = mount
-			mountedImages = append(mountedImages, image)
 		case TypeCache:
-			mount, lockedPaths, err := GetCacheMount(tokens, store, "", nil)
+			mount, intermediateMount, lockedPaths, err := GetCacheMount(tokens, nil, tmpDir)
 			lockedTargets = lockedPaths
```
nalind, line 738: Somewhere in there, this was corrected to `lockedTargets = append(lockedTargets, lockedPaths...)`, so that the presence of more than one cache mount wouldn't just lose them.

internal/parse/parse.go
```diff
@@ -491,17 +648,53 @@ func GetVolumes(ctx *types.SystemContext, store storage.Store, volumes []string,
...
+				}
+				if err := os.Remove(intermediateMount); err != nil {
+					logrus.Debugf("removing should-be-empty directory %q: %v", intermediateMount, err)
+				}
+			}
+			for _, image := range mountedImages {
+				if _, err := store.UnmountImage(image, false); err != nil {
+					logrus.Debugf("unmounting image %q: %v", image, err)
+				}
+			}
```
nalind, line 695: A fix somewhere between this branch and 1.33 added logic here to free the locks on `lockedTargets`, and to stop returning them to the caller in error cases.  Likewise in `GetVolumes()` above.

nalind: https://github.com/containers/buildah/pull/5931#discussion_r1925713553 points to `cleanupRunMounts()` as an example, duplicating it here for completeness.

dashea: The unlock loop there was duplicated in cmd/buildah/run.go, so I moved that to its own function in the internal parse module (reusing the UnlockLockArray name from newer branches) and call that in the cleanup methods.

internal/parse/parse.go
dashea: oops missed this one, another force push incoming...

dashea:
Found the problem with the `bud-git-context` test: I was not returning the new intermediate directory in one of the paths through `TempDirForURL`.

internal/volumes/bind_notlinux.go
```diff
@@ -0,0 +1,15 @@
+//go:build !linux
```
nalind, line 1:
```suggestion
//go:build !linux
// +build !linux
```

internal/open/open_unsupported.go
```diff
@@ -0,0 +1,7 @@
+//go:build !linux && !freebsd && !darwin
```
nalind, line 1:
```suggestion
//go:build !linux && !freebsd && !darwin
// +build linux freebsd darwin
```

internal/open/open_unix.go
```diff
@@ -0,0 +1,168 @@
+//go:build linux || freebsd || darwin
```
nalind, line 1:
```suggestion
//go:build linux || freebsd || darwin
// +build linux freebsd darwin
```

PR #5934, opened 2025-01-22
Accept image names as sources for cache mounts

nalind:
#### What type of PR is this?

/kind feature

Allow the "from" option for cache mounts to refer to an image, which, if we attempt to mount read-write, is mounted using an overlay to avoid getting errors if the process we're running with the cache mount attempts to write to it.

Add logic to clean up the bind mount in Builder.getBindMount() that's triggered if `builder.runSetupVolumeMounts()` returns an error.

When creating cache directories for cache mounts, process the container's ID mappings.

#### How to verify it

Updated integration test!

None

```release-note
`buildah build` RUN instructions and `buildah run` can now refer to images with the `from` argument when using the `--mount=type=cache` option.
```

### DIFF
--- a/cmd/buildah/run.go
+++ b/cmd/buildah/run.go
@@ -191,7 +191,7 @@ func runCmd(c *cobra.Command, args []string, iopts runInputOptions) error {
 	if err != nil {
 		return fmt.Errorf("building system context: %w", err)
 	}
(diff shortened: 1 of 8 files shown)


rhatdan:
LGTM
@flouthoc @giuseppe PTAL

nalind:
I took another look at the change that ensures that the lock directory location is known and created when the cache source is something other than a directory that we create under the cache parent.

TomSweeneyRedHat:
LGTM

PR #5943, opened 2025-01-24
Support the containers.conf container_name_as_hostname option

nalind:
#### What type of PR is this?

/kind feature

When containers.conf has the "container_name_as_hostname" option set, use that value, with values that don't fit `[A-Za-z0-9][A-Za-z0-9._-]+` stripped out.

#### How to verify it

New integration test!

None

```release-note
None
```


run_common.go
```diff
@@ -2092,3 +2092,28 @@ func relabel(path, mountLabel string, shared bool) error {
 	}
 	return nil
 }
+
+// mapContainerNameToHostname returns the passed-in string with characters that
+// don't match the pattern "[A-Za-z0-9][A-Za-z0-9._-]+" stripped out.
+func mapContainerNameToHostname(containerName string) string {
+	const initialAllowedChars = "abcdefghijklmnopqrstuvwxyz" + "ABCDEFGHIJKLMNOPQRSTUVWXYZ" + "0123456789"
```
flouthoc, line 2099: Would regex such as `^[A-Za-z0-9]` and `[A-Za-z0-9._-]` be easier to maintain ? or could it be slow here ?

run_common.go
```diff
@@ -2092,3 +2092,28 @@ func relabel(path, mountLabel string, shared bool) error {
...
+	const allowedChars = initialAllowedChars + "._-"
```
Luap99, line 2100: `_` is not a valid hostname char and must be stripped, the same is done in podman https://github.com/containers/podman/pull/24675/commits/6fa234a5de44b5279b244e06e889e74d9d27ae40

Luap99 (commented):
I think there needs to be a length limit like in podman or does buildah restrict a name length already earlier?

run_common.go
nalind: That's potentially a lot of calls to `FindStringIndex()`, but it's worth a shot if the result is more readable.

nalind:
> I think there needs to be a length limit like in podman or does buildah restrict a name length already earlier?

Nothing between setting one in the spec and handing it off to the runtime.  Added one when we base it on the container name.

Luap99 (approved):
LGTM

PR #5954, opened 2025-01-29
CI: parallize unit tests

flouthoc:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ GO_GCFLAGS := $(shell if $(GO) version|grep -q gccgo; then echo "-gccgoflags"; e
 NPROCS := $(shell nproc)
 export GO_BUILD=$(GO) build
 export GO_TEST=$(GO) test -parallel=$(NPROCS)
-RACEFLAGS := $(shell $(GO_TEST) -race ./pkg/dummy > /dev/null 2>&1 && echo -race)
+RACEFLAGS ?= $(shell $(GO_TEST) -race ./pkg/dummy > /dev/null 2>&1 && echo -race)
 
 COMMIT_NO ?= $(shell git rev-parse HEAD 2> /dev/null || true)
 GIT_COMMIT ?= $(if $(shell git status --porcelain --untracked-files=no),${COMMIT_NO}-dirty,${COMMIT_NO})
--- a/Makefile
+++ b/Makefile
@@ -22,8 +22,9 @@ STRIP ?= strip
 GO := go
 GO_LDFLAGS := $(shell if $(GO) version|grep -q gccgo; then echo "-gccgoflags"; else echo "-ldflags"; fi)
 GO_GCFLAGS := $(shell if $(GO) version|grep -q gccgo; then echo "-gccgoflags"; else echo "-gcflags"; fi)
+NPROCS := $(shell nproc)
 export GO_BUILD=$(GO) build
-export GO_TEST=$(GO) test
+export GO_TEST=$(GO) test -parallel=$(NPROCS)
 RACEFLAGS := $(shell $(GO_TEST) -race ./pkg/dummy > /dev/null 2>&1 && echo -race)
 
 COMMIT_NO ?= $(shell git rev-parse HEAD 2> /dev/null || true)
--- a/buildah_test.go
+++ b/buildah_test.go
@@ -40,6 +40,7 @@ func TestMain(m *testing.M) {
 }
 
 func TestOpenBuilderCommonBuildOpts(t *testing.T) {
+	t.Parallel()
 	ctx := context.TODO()
 	store, err := storage.GetStore(types.StoreOptions{
 		RunRoot:         t.TempDir(),
--- a/buildah_test.go
+++ b/buildah_test.go
@@ -40,7 +40,12 @@ func TestMain(m *testing.M) {
 }
 
 func TestOpenBuilderCommonBuildOpts(t *testing.T) {
-	t.Parallel()
+	// This test cannot be parallized as this uses NewBuilder()
+	// which eventually and indirectly accesses a global variable
+	// defined in `go-selinux`, this must be fixed at `go-selinux`
+	// or builder must enable sometime of locking mechanism i.e if
+	// routine is creating Builder other's must wait for it.
+	// Tracked here: https://github.com/containers/buildah/issues/5967
 	ctx := context.TODO()
 	store, err := storage.GetStore(types.StoreOptions{
 		RunRoot:         t.TempDir(),
--- a/contrib/cirrus/test.sh
+++ b/contrib/cirrus/test.sh
@@ -67,7 +67,13 @@ else
             showrun make validate
             ;;
         unit)
-            showrun make test-unit
+            race=
+            if [[ -z "$CIRRUS_PR" ]]; then
+               # If not running on a PR then run unit tests
+               # with appropriate `-race` flags.
+               race="-race"
+            fi
+            showrun make test-unit RACEFLAGS=$race
             ;;
         conformance)
             # Typically it's undesirable to install packages at runtime.
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -154,6 +154,8 @@ cross_build_task:
 
 unit_task:
     name: 'Unit tests w/ $STORAGE_DRIVER'
+    gce_instance:
+      cpu: 4
     alias: unit
     skip: &not_build_docs >-
         $CIRRUS_CHANGE_TITLE =~ '.*CI:DOCS.*' ||
--- a/chroot/run_test.go
+++ b/chroot/run_test.go
@@ -123,6 +123,7 @@ func testMinimal(t *testing.T, modify func(g *generate.Generator, rootDir, bundl
 }
 
 func TestNoop(t *testing.T) {
+	t.Parallel()
 	if unix.Getuid() != 0 {
 		t.Skip("tests need to be run as root")
 	}
@@ -130,6 +131,7 @@ func TestNoop(t *testing.T) {
 }
 
 func TestMinimalSkeleton(t *testing.T) {
+	t.Parallel()
 	if unix.Getuid() != 0 {
 		t.Skip("tests need to be run as root")
 	}
@@ -142,6 +144,7 @@ func TestMinimalSkeleton(t *testing.T) {
 }
 
 func TestProcessTerminal(t *testing.T) {
+	t.Parallel()
 	if unix.Getuid() != 0 {
 		t.Skip("tests need to be run as root")
 	}
@@ -160,6 +163,7 @@ func TestProcessTerminal(t *testing.T) {
 }
 
 func TestProcessConsoleSize(t *testing.T) {
+	t.Parallel()
 	if unix.Getuid() != 0 {
 		t.Skip("tests need to be run as root")
 	}
@@ -182,6 +186,7 @@ func TestProcessConsoleSize(t *testing.T) {
 }
 
 func TestProcessUser(t *testing.T) {
+	t.Parallel()
 	if unix.Getuid() != 0 {
 		t.Skip("tests need to be run as root")
 	}
@@ -205,6 +210,7 @@ func TestProcessUser(t *testing.T) {
 }
 
 func TestProcessEnv(t *testing.T) {
+	t.Parallel()
 	if unix.Getuid() != 0 {
 		t.Skip("tests need to be run as root")
 	}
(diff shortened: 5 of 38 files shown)


tests/source.bats
```diff
@@ -152,4 +152,6 @@ load helpers
   run diff -r $srcdir $pulldir
   # FIXME: if there's a nonzero chance of this failing, include actual diffs
   assert "$status" -eq 0 "status from diff of srcdir vs pulldir"
+
+  stop_registry
```
Luap99, line 156: this is reverting changes from Ed's parallel PR

tests/test_runner.sh
```diff
@@ -13,4 +13,4 @@ function execute() {
 }
 
 # Run the tests.
-execute time bats -j $(nproc) --tap "${@:-.}"
+execute time bats --tap "${@:-.}"
```
Luap99, line 16: this is as well, you are disabling parallel testing for integration tests

tests/bud.bats
```diff
@@ -143,8 +143,6 @@ _EOF
 
 @test "bud: build push with --force-compression" {
   skip_if_no_podman
-  blobcachedir=${TEST_SCRATCH_DIR}/blobcachelocal
```
Luap99, line 146: here too

tests/blobcache.bats
```diff
@@ -2,12 +2,6 @@
 
 load helpers
 
-# The 'rm cachefile' in the "blobs must be reused" test
```
Luap99, line 5: and this file as well

Luap99 (changes requested):
Looks like you are reverting the integration test changes.

tests/test_runner.sh
Luap99: if you add `-T` to bats it will add `in xxx ms` after each test name to the log.
Then you can check these timings to see if you spot any particular slow test that could be optimized.

flouthoc:
> Looks like you are reverting the integration test changes.

Yes did it by mistake fixed it.

flouthoc:
Following tests are no parallel and are racing when using `t.Parallel` , also they are high contributors to the time of unit test, looks like I will have to do changes in tests so we can make them parallel

```
TestPutNoChroot: 202.85 seconds
TestStatNoChroot: 166.13 seconds
TestCheckLUKSPassphrase: 60.86 seconds
TestMounts: 52.00 seconds
TestCWConvertImage: 51.84 seconds
TestPutNoChroot/archive=regular,topdir=: 34.64 seconds
TestPutNoChroot/archive=regular,topdir=.: 29.35 seconds
TestPutNoChroot/archive=regular,rename=no-match-dir: 27.94 seconds
TestPutNoChroot/archive=regular,topdir=top: 27.89 seconds
TestPutNoChroot/archive=regular,rename=no-match-file: 27.77 seconds
```

Luap99:
image

They do not seem to benefit much from more cores/memory so you likely need to look at other options or get them to run parallel much better. I guess the early spikes are when we we compile the go code in the ci setup. Maybe also worth looking into. It is not clear to me if we would actually need to compile anything else for the unit test.

The other thing could be is to split the unit-test target in three separate ones and then run them in parallel.
Another thing I see the tests use `-cover` and `-race`, I am not sure how value these add for our testing.
race is known to be slow, https://go.dev/doc/articles/race_detector#Runtime_Overheads
>  The cost of race detection varies by program, but for a typical program, memory usage may increase by 5-10x and execution time by 2-20x. 
So maybe drop these flags and see how much faster they run.

flouthoc:
> image
> 
> They do not seem to benefit much from more cores/memory so you likely need to look at other options or get them to run parallel much better. I guess the early spikes are when we we compile the go code in the ci setup. Maybe also worth looking into. It is not clear to me if we would actually need to compile anything else for the unit test.
> 
> The other thing could be is to split the unit-test target in three separate ones and then run them in parallel. Another thing I see the tests use `-cover` and `-race`, I am not sure how value these add for our testing. race is known to be slow, https://go.dev/doc/articles/race_detector#Runtime_Overheads
> 
> > The cost of race detection varies by program, but for a typical program, memory usage may increase by 5-10x and execution time by 2-20x.
> > So maybe drop these flags and see how much faster they run.

Wow @Luap99 !!! unit test without `-cover` and `-race` completed in 8 minutes, let me try without `-race` only and including `-cover`.

flouthoc:
Looks like removing `-race` does the trick, brings down the all the unit tests from 60 something to 8m

buildah_test.go
```diff
@@ -40,7 +40,6 @@ func TestMain(m *testing.M) {
 }
 
 func TestOpenBuilderCommonBuildOpts(t *testing.T) {
-	t.Parallel()
```
Luap99: When you remove this again would it make sense to add a comment to each here why they cannot be parallelized. It would help someone in the future wondering why this on is not parallel or prevent someone from trying to added parallel here, same in the other places hwere you had to revert this.

flouthoc: Sure.

nalind:
> @nalind Any objections to removing `-race` from the unit tests?

Other than that I'll probably forget to run them that way from time to time if it isn't automated, which isn't that important, no.

flouthoc:
@nalind @Luap99 @containers/buildah-maintainers PTAL

flouthoc:
Also it looks like `osh-diff-scan:fedora-rawhide-x86_64:buildah-fedora` is the bottleneck now, it is the longest running job.

nalind:
> > @nalind Any objections to removing `-race` from the unit tests?
> 
> Other than that I'll probably forget to run them that way from time to time if it isn't automated, which isn't that important, no.

Come to think of it, can we conditionally keep that for the scheduled "cron" builds?  I don't think anyone actually waits for them to finish.

flouthoc:
> > > @nalind Any objections to removing `-race` from the unit tests?
> > 
> > 
> > Other than that I'll probably forget to run them that way from time to time if it isn't automated, which isn't that important, no.
> 
> Come to think of it, can we conditionally keep that for the scheduled "cron" builds? I don't think anyone actually waits for them to finish.

Sounds good to me, is there any way to test that when I open a PR or it will only run on merge ?  I will open a separate PR to add `env` variable in `.github` workflow.

Luap99:
> Also it looks like `osh-diff-scan:fedora-rawhide-x86_64:buildah-fedora` is the bottleneck now, it is the longest running job.

None of the packit tests enforce anything, just ignore them for now.

> Come to think of it, can we conditionally keep that for the scheduled "cron" builds? I don't think anyone actually waits for them to finish.

That is easy to do sure except nobody looks at these results so that is pretty pointless.

How imporant is -race really, did you ever catch actual issues with it?

> Sounds good to me, is there any way to test that when I open a PR or it will only run on merge ? I will open a separate PR to add env variable for -race in .github workflow.

Well anything post merge cannot be tested on a PR and this is not a github workflow. This must be done in cirrus.yml and would be easy, just check if the `CIRRUS_PR` env is empty then we run on a branch.

Luap99 (approved):
LGTM

Luap99:
Total CI time reported by cirrus 38:01

| type        | user     | fs      |      d13 |      f40 |      f41 |
| ----------- | -------- | ------- | -------- | -------- | -------- |
| Unit        | root     | vfs     |          |    08:17 |          |
| Unit        | root     | overlay |          |    08:02 |          |
| Conformance | root     | vfs     |    23:49 |          |          |
| Conformance | root     | overlay |    14:33 |          |          |
| Integration | root     | vfs     |    23:53 |    29:30 |    30:12 |
| Integration | root     | overlay |    19:50 |          |    23:37 |
| Integration | rootless | overlay |    26:06 |          |    26:34 |

That seems quite good, vfs is noticeably slower though which I guess is expected to the higher amount of io required for it.

buildah_test.go
```diff
@@ -40,6 +40,11 @@ func TestMain(m *testing.M) {
 }
 
 func TestOpenBuilderCommonBuildOpts(t *testing.T) {
+	// This test cannot be parallized as this uses NewBuilder()
+	// which eventually and indirectly accesses a global variable
+	// defined in `go-selinux`, this must be fixed at `go-selinux`
+	// or builder must enable sometime of locking mechanism i.e if
+	// routine is creating Builder other's must wait for it.
```
nalind, line 47: How will we know when it's safe? Is there an open issue?

flouthoc: I just created an issue https://github.com/containers/buildah/issues/5967 and added in `comment` and `commit` message.

nalind:
> @nalind @TomSweeneyRedHat @rhatdan PTAL, if consensus is made for `cron` then I'll open a new PR for that.

Why wait?

Makefile
```diff
@@ -25,8 +25,12 @@ GO_GCFLAGS := $(shell if $(GO) version|grep -q gccgo; then echo "-gccgoflags"; e
...
 export GO_BUILD=$(GO) build
 export GO_TEST=$(GO) test -parallel=$(NPROCS)
-#RACEFLAGS := $(shell $(GO_TEST) -race ./pkg/dummy > /dev/null 2>&1 && echo -race)
-RACEFLAGS :=
+# If RACEFLAGS is defined then set appropriate RACEFLAGS automatically for `go test`
+ifdef RACEFLAGS
+	RACEFLAGS := $(shell $(GO_TEST) -race ./pkg/dummy > /dev/null 2>&1 && echo -race)
+else
+	RACEFLAGS :=
+endif
```
Luap99, line 33: I think this should just be
`RACEFLAGS ?= `

https://www.gnu.org/software/make/manual/html_node/Setting.html

contrib/cirrus/test.sh
```diff
@@ -67,7 +67,14 @@ else
...
         unit)
-            showrun make test-unit
+
+            if [[ -n "$CIRRUS_PR" ]]; then
+                showrun make test-unit
+            else
+		# If not running on a PR then run unit tests
+		# with appropriate `-race` flags.
+                showrun RACEFLAGS=true make test-unit
+            fi
```
Luap99, line 77: Then here pass `RACEFLAGS=-race`, and I would avoid duplicating the full command, the if/else should just focus on the changed part. IMO this would be more readable

```sh
local race=
if [[ -z "$CIRRUS_PR" ]]; then
	# If not running on a PR then run unit tests
	# with appropriate `-race` flags.
	race="-race"
fi
showrun RACEFLAGS=$race make test-unit
```

Also indentation looks wrong

Luap99 (commented):
I still don't think there is a point in having race enabled post merge.
AFAICT nobody ever looks at any of these failures anyway. ci-monitor emails for the cron jobs are ignored and for post merge tests on nobody ever looked at them.

But maybe I am just wrong here? Who is looking at that?

nalind:
> But maybe I am just wrong here? Who is looking at that?

Anecdata: I do read the ones for buildah.

Luap99:
> > But maybe I am just wrong here? Who is looking at that?
> 
> Anecdata: I do read the ones for buildah.

Ok if you actual can make use of the data then sure let's do that. I just have never seen anyone respond to these emails or post PRs to fix the issues from there. Maybe that just wasn't necessary on buildah so far and/or I don't look at buildah PRs often enough.

flouthoc:
@nalind @Luap99 PTAL again.

Makefile
```diff
@@ -25,8 +25,7 @@ GO_GCFLAGS := $(shell if $(GO) version|grep -q gccgo; then echo "-gccgoflags"; e
 NPROCS := $(shell nproc)
 export GO_BUILD=$(GO) build
 export GO_TEST=$(GO) test -parallel=$(NPROCS)
-#RACEFLAGS := $(shell $(GO_TEST) -race ./pkg/dummy > /dev/null 2>&1 && echo -race)
-RACEFLAGS :=
+RACEFLAGS ?= $(shell $(GO_TEST) -race ./pkg/dummy > /dev/null 2>&1 && echo -race)
```
rhatdan, line 28: Can you merge this one with previous

rhatdan:
LGTM

Makefile
flouthoc: Done.

rhatdan:
@nalind @Luap99 PTAL

contrib/cirrus/test.sh
```diff
@@ -67,7 +67,13 @@ else
             showrun make validate
             ;;
         unit)
-            showrun make test-unit
+	    race=
```
Luap99, line 70: nit, line indentation still seems wrong, looks like you use one tab and spaces on that line.

flouthoc: fixed.

Luap99:
Just to confirm post merge tests did run with -race so this works correctly.
https://cirrus-ci.com/task/4767565819412480

PR #5959, opened 2025-01-30
Use tmpfs for integration tests

flouthoc:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None 
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -427,7 +427,13 @@ func (s *StageExecutor) performCopy(excludes []string, copies ...imagebuilder.Co
 				data = strings.TrimPrefix(data, "\n")
 				// add breakline when heredoc ends for docker compat
 				data = data + "\n"
-				tmpFile, err := os.Create(filepath.Join(parse.GetTempDir(), path.Base(filepath.ToSlash(file.Name))))
+				// Create seperate subdir for this file.
+				tmpDir, err := os.MkdirTemp(parse.GetTempDir(), "buildah-heredoc")
+				if err != nil {
+					return fmt.Errorf("unable to create tmp dir for heredoc run %q: %w", parse.GetTempDir(), err)
+				}
+				defer os.RemoveAll(tmpDir)
+				tmpFile, err := os.Create(filepath.Join(tmpDir, path.Base(filepath.ToSlash(file.Name))))
 				if err != nil {
 					return fmt.Errorf("unable to create tmp file for COPY instruction at %q: %w", parse.GetTempDir(), err)
 				}
@@ -442,7 +448,7 @@ func (s *StageExecutor) performCopy(excludes []string, copies ...imagebuilder.Co
 					tmpFile.Close()
 					return fmt.Errorf("unable to write contents of heredoc file at %q: %w", tmpFile.Name(), err)
 				}
-				copySources = append(copySources, filepath.Base(tmpFile.Name()))
+				copySources = append(copySources, filepath.Join(filepath.Base(tmpDir), filepath.Base(tmpFile.Name())))
 				tmpFile.Close()
 			}
 			contextDir = parse.GetTempDir()
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -296,6 +296,49 @@ function rm() {
     run_unshared rm "$@"
 }
 
+#################
+#  run_with_log  #  Logs command before running it
+#################
+#
+function run_with_log() {
+    local expected_rc=0
+    local retry=1
+    local cmd="$*"
+    case "$1" in
+        [0-9])           expected_rc=$1; shift;;
+        [1-9][0-9])      expected_rc=$1; shift;;
+        [12][0-9][0-9])  expected_rc=$1; shift;;
+        '?')             expected_rc=  ; shift;;  # ignore exit code
+        --retry)         retry=3;        shift;;  # retry with sleep of 1 sec
+    esac
+    while [ $retry -gt 0 ]; do
+        retry=$(( retry - 1 ))
+        echo "$_LOG_PROMPT $cmd"
+        run "$@"
+        echo "$output"
+        if [ "$status" -ne 0 ]; then
+            echo -n "[ rc=$status ";
+            if [ -n "$expected_rc" ]; then
+                if [ "$status" -eq "$expected_rc" ]; then
+                    echo -n "(expected) ";
+                else
+                    echo -n "(** EXPECTED $expected_rc **) ";
+                fi
+            fi
+            echo "]"
+        fi
+        if [ -n "$expected_rc" ]; then
+            if [ "$status" -eq "$expected_rc" ]; then
+                return
+            elif [ $retry -gt 0 ]; then
+                echo "[ RETRYING ]" >&2
+                sleep 1
+            else
+                die "exit code is $status; expected $expected_rc"
+            fi
+        fi
+    done
+}
 
 #################
 #  run_buildah  #  Invoke buildah, with timeout, using BATS 'run'
--- a/tests/mkcw.bats
+++ b/tests/mkcw.bats
@@ -22,9 +22,9 @@ function mkcw_check_image() {
 
   # Decrypt, mount, and take a look around.
   uuid=$(cryptsetup luksUUID "$mountpoint"/disk.img)
-  cryptsetup luksOpen --key-file "$TEST_SCRATCH_DIR"/key "$mountpoint"/disk.img "$uuid"
+  run_with_log cryptsetup luksOpen --key-file "$TEST_SCRATCH_DIR"/key "$mountpoint"/disk.img "$uuid"
   mkdir -p "$TEST_SCRATCH_DIR"/mount
-  mount /dev/mapper/"$uuid" "$TEST_SCRATCH_DIR"/mount
+  run_with_log mount /dev/mapper/"$uuid" "$TEST_SCRATCH_DIR"/mount
   # Should have a not-empty config file with parts of an image's config.
   test -s "$TEST_SCRATCH_DIR"/mount/.krun_config.json
   # Should have a /tmp directory, at least.
@@ -42,9 +42,15 @@ function mkcw_check_image() {
   fi
 
   # Clean up.
-  umount "$TEST_SCRATCH_DIR"/mount
-  cryptsetup luksClose "$uuid"
-  buildah umount "$ctrID"
+  run_with_log umount -f "$TEST_SCRATCH_DIR"/mount
+  # `Retry` if `luksClose` fails with defaults of `run_with_log` because
+  # when unmounting the filesystem mounted on the device /dev/mapper/"$uuid"
+  # without `retry` somehow we end up in a state where mount is still being
+  # used by the kernel because when we do `lsof /dev/mapper/"$uuid"` it
+  # shows nothing but `dmsetup info -c $uuid` shows the device is still
+  # under use. Adding `--retry` in between somehow fixes this.
+  run_with_log --retry cryptsetup luksClose "$uuid"
+  run_buildah umount "$ctrID"
 }
 
 @test "mkcw-convert" {
--- a/tests/test_runner.sh
+++ b/tests/test_runner.sh
@@ -3,9 +3,8 @@ set -e
 
 cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"
 
-# Default to using /var/tmp for test space, since it's more likely to support
-# labels than /tmp, which is often on tmpfs.
-export TMPDIR=${TMPDIR:-/var/tmp}
+# Default to using /tmp for test space.
+export TMPDIR=${TMPDIR:-/tmp}
 
 function execute() {
 	>&2 echo "++ $@"
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -539,13 +539,14 @@ _EOF
 }
 
 @test "bud build with heredoc content" {
-  _prefetch quay.io/fedora/python-311
+  _prefetch alpine
   run_buildah build -t heredoc $WITH_POLICY_JSON -f $BUDFILES/heredoc/Containerfile .
   expect_output --substring "print first line from heredoc"
   expect_output --substring "print second line from heredoc"
   expect_output --substring "Heredoc writing first file"
   expect_output --substring "some text of first file"
-  expect_output --substring "file2 from python"
+  expect_output --substring "file passed to program"
+  expect_output --substring "and it contains multiple"
   expect_output --substring "(your index page goes here)"
   expect_output --substring "(robots content)"
   expect_output --substring "(humans content)"
@@ -560,7 +561,7 @@ _EOF
   # verify that build output contains summary of heredoc content
   expect_output --substring 'RUN <<EOF \(echo "print first line from heredoc"...)'
   expect_output --substring 'RUN <<EOF \(echo "Heredoc writing first file" >> /file1...)'
-  expect_output --substring 'RUN python3 <<EOF \(with open\("/file2", "w") as f:...)'
+  expect_output --substring 'RUN cat <<EOF \(file passed to program...)'
   expect_output --substring 'ADD <<EOF /index.html \(\(your index page goes here))'
   expect_output --substring 'COPY <<robots.txt <<humans.txt /test/ \(\(robots content)) \(\(humans content))'
 }
--- a/tests/bud/heredoc/Containerfile
+++ b/tests/bud/heredoc/Containerfile
@@ -1,4 +1,4 @@
-FROM quay.io/fedora/python-311
+FROM alpine
 
 USER root
 WORKDIR /
@@ -15,12 +15,11 @@ EOF
 
 RUN cat file1
 
-RUN python3 <<EOF
-with open("/file2", "w") as f:
-    print("file2 from python", file=f)
+RUN cat <<EOF
+file passed to program
+and it contains multiple
+lines
 EOF
-
-RUN cat file2
 
 ADD <<EOF /index.html
  (your index page goes here)


flouthoc:
@Luap99 It seems increasing CPU for `vfs` has some improvement.

flouthoc:
Also `containerized_integration` came from `27` something to `19m`

Luap99:
image

cirrus graph for the container task seems broken as this usage makes no sense.

image

Looking at another int test we do not seem max out the cpu much so I am not sure if 8 cores make much sense. Also memory usage seems quite low so I think we can reduce that as well.


Overall the first question is what is the goal here? 30 mins total CI like podman? What is the target time we are aiming for?
Once we know that we can see how much we want to bump the cores, because we also need to keep in mind of costs. If double the cores means 2/3 of the time then the costs will most likely be higher.

flouthoc:
> Overall the first question is what is the goal here? 30 mins total CI like podman

I think we are already under 30mins, this PR was just an experiment. Increasing cores reduces time by a lot but yeah I don't think it's a good idea to keep throwing money at the problem if its not needed.

Luap99:
> > Overall the first question is what is the goal here? 30 mins total CI like podman
> 
> I think we are already under 30mins, this PR was just an experiment. Increasing cores reduces time by a lot but yeah I don't think it's a good idea to keep throwing money at the problem if its not needed.

I am talking about total time for CI to finish. Not just a single task time. There is a bit of overhead in scheduling tasks.

You need to look at the cirrus build page at the top it shows you `Finished in 57:37` on this PR (because it was not rebased on the unit test speedup)

flouthoc:
> > > Overall the first question is what is the goal here? 30 mins total CI like podman
> > 
> > 
> > I think we are already under 30mins, this PR was just an experiment. Increasing cores reduces time by a lot but yeah I don't think it's a good idea to keep throwing money at the problem if its not needed.
> 
> I am talking about total time for CI to finish. Not just a single task time. There is a bit of overhead in scheduling tasks.
> 
> You need to look at the cirrus build page at the top it shows you `Finished in 57:37` on this PR (because it was not rebased on the unit test speedup)

Yes I am only talking about integration tests, waiting for this to merge for `unit` tests here https://github.com/containers/buildah/pull/5954

flouthoc:
@Luap99 I see minor improvement in other tests but not in `containerized_integration`.

flouthoc:
@Luap99 @nalind PTAL

contrib/cirrus/lib.sh
```diff
@@ -192,10 +192,13 @@ in_podman() {
         envargs+=("-e" "$var")
     done <<<"$(passthrough_envars)"
 
+    mount -t tmpfs -o size=75%,mode=0700 none /tmp
```
Luap99, line 195: This seems dangerous, why would we over mount /tmp on an existing running system. What if another application needs access to files it already wrote there.

Second mode=0700 seems just not suitable for /tmp that has to be writeable for everybody

contrib/cirrus/lib.sh
```diff
@@ -192,10 +192,13 @@ in_podman() {
...
+
     showrun podman run -i --name="$IN_PODMAN_NAME" \
                    --net=host \
                    --privileged \
                    --cgroupns=host \
+		   --tmpfs /tmp:mode=1777 \
```
Luap99, line 201: indentation

contrib/cirrus/lib.sh
```diff
@@ -205,6 +208,7 @@ in_podman() {
                    -v "$HOME/auth:$HOME/auth:ro" \
                    -v /sys/fs/cgroup:/sys/fs/cgroup:rw \
                    -v /dev/fuse:/dev/fuse:rw \
+                   -v /tmp:/var/lib/containers:rw \
```
Luap99, line 211: do we even use /var/lib/containers? seems pointless if all the tests write to a tmpdir

contrib/cirrus/lib.sh
```diff
@@ -302,6 +306,8 @@ setup_rootless() {
     groupadd -g $rootless_gid $ROOTLESS_USER
     useradd -g $rootless_gid -u $rootless_uid --no-user-group --create-home $ROOTLESS_USER
 
+    mount -t tmpfs -o size=75%,mode=0700,uid=$rootless_uid,gid=$rootless_gid none /home/$ROOTLESS_USER
```
Luap99, line 309: same here do we actually use the main container storage?

contrib/cirrus/setup.sh
```diff
@@ -83,6 +83,13 @@ esac
 # Previously, golang was not installed
 source $(dirname $0)/lib.sh
 
+if [[ "${IN_PODMAN:-false}" != "true" ]]
+then
+	# mount a tmpfs for the container storage to speed up the IO
+	# side effect is we clear all potentially pre existing data so we know we always start "clean"
+	mount -t tmpfs -o size=75%,mode=0700 none /var/lib/containers
+fi
+
```
Luap99, line 92: and same here

.cirrus.yml
```diff
@@ -284,7 +285,8 @@ integration_rootless_task:
 
     gce_instance:
         image_name: "$IMAGE_NAME"
-        cpu: 4
+        cpu: 8
+        memory: "24G"
```
Luap99, line 289: Do we actually need 24G of ram. Looking at the memory graph it seems we never get above 2G if cirrus can be trusted. Does so much RAM do anything for the performance compared to 16, 8 or 4G? same for the other places

Luap99 (commented):
Does the tmpfs change actually do anything? AFAICT we set TMPDIR=/var/tmp so nothing ends up on tmpfs?

In general a PR like this benefits from precise numbers before/after. A commit like "bump cpu and memory" is not helping any future reader. It is missing the why we do this and the numbers on much we safe.

Luap99:
Actually never mind we only set ` TMPDIR: '/var/tmp'` on the conformance vfs task per 8b0ecd78702589b650d1fd9bc301f7616b34d85f

So I think the tests were already using tmpfs

flouthoc:
Putting this PR on draft and timing `tmpfs` changes again, I have created a new PR for resource bump https://github.com/containers/buildah/pull/5979

tests/test_runner.sh
```diff
@@ -3,9 +3,10 @@ set -e
 
 cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"
 
+mount -t tmpfs -o size=75%,mode=0777 none /tmp
```
Luap99, line 6: /tmp is already tmpfs in out VMs, you really do not have to do any of the mounting I think

flouthoc: Some of the tests are getting `no space left on device`, I did it because before this tests were unable to start. But strangely some tests are still running out of space.

```
[+1085s] not ok 66 bud build with heredoc content
[+1085s] # (from function `die' in file ./helpers.bash, line 395,
[+1085s] #  from function `run_buildah' in file ./helpers.bash, line 382,
[+1085s] #  in test file ./bud.bats, line 543)
[+1085s] #   `run_buildah build -t heredoc $WITH_POLICY_JSON -f $BUDFILES/heredoc/Containerfile .' failed
[+1085s] # /var/tmp/go/src/github.com/containers/buildah/tests /var/tmp/go/src/github.com/containers/buildah/tests
[+1085s] # # [checking for: quay.io/fedora/python-311]
[+1085s] # # [copy docker://quay.io/fedora/python-311 dir:/tmp/bats-run-UPOUUP/suite/buildah-image-cache/quay.io-fedora-python-311-]
[+1085s] # Getting image source signatures
[+1085s] # Copying blob sha256:95fe7b64e419f415e8175e1a4f65b41ddb69fdb5c5730a26abd9ffd4a50a1281
[+1085s] # Copying blob sha256:c1266253c0b60dea30470d7aa163480b5b3dfde6053704c9cf63b567e8df19ff
[+1085s] # Copying blob sha256:c84d07b49a1a0127ffeb49b55df7b85d2fd2f99e404527a75e0ac0dc5c600892
[+1085s] # Copying blob sha256:49a3ccc0ca8011503e6806c84e223fddb9362740f76f45e0705831ce62ac0c0e
[+1085s] # Copying config sha256:b36ede382df03e93b59fb74fae5a2ee83d1d57ae613087f7d69956ac606a7a2a
[+1085s] # Writing manifest to image destination
[+1085s] # # [copy dir:/tmp/bats-run-UPOUUP/suite/buildah-image-cache/quay.io-fedora-python-311- containers-storage:quay.io/fedora/python-311]
[+1085s] # Getting image source signatures
[+1085s] # Copying blob sha256:95fe7b64e419f415e8175e1a4f65b41ddb69fdb5c5730a26abd9ffd4a50a1281
[+1085s] # Copying blob sha256:c1266253c0b60dea30470d7aa163480b5b3dfde6053704c9cf63b567e8df19ff
[+1085s] # Copying blob sha256:c84d07b49a1a0127ffeb49b55df7b85d2fd2f99e404527a75e0ac0dc5c600892
[+1085s] # Copying blob sha256:49a3ccc0ca8011503e6806c84e223fddb9362740f76f45e0705831ce62ac0c0e
[+1085s] # Copying config sha256:b36ede382df03e93b59fb74fae5a2ee83d1d57ae613087f7d69956ac606a7a2a
[+1085s] # Writing manifest to image destination
[+1085s] # # /var/tmp/go/src/github.com/containers/buildah/tests/./../bin/buildah build -t heredoc --signature-policy /var/tmp/go/src/github.com/containers/buildah/tests/./policy.json -f /var/tmp/go/src/github.com/containers/buildah/tests/./bud/heredoc/Containerfile .
[+1085s] # STEP 1/20: FROM quay.io/fedora/python-311
[+1085s] # Error: creating build container: creating container: creating read-write layer with ID "9932f73065e75a6c535d418c0f80b9afd5d412729165fcf6fb635562026087c9": no space left on device
```

flouthoc:
@Luap99 @nalind PTAL

flouthoc:
Metrics

```
>> 29:07
| type        | user     | fs      |      d13 |      f40 |      f41 |
| ----------- | -------- | ------- | -------- | -------- | -------- |
| Unit        | root     | vfs     |          |    08:27 |          |
| Unit        | root     | overlay |          |    08:36 |          |
| Conformance | root     | vfs     |    22:48 |          |          |
| Conformance | root     | overlay |    12:12 |          |          |
| Integration | root     | vfs     |    11:51 |    18:36 |    13:03 |
| Integration | root     | overlay |    11:24 |          |    13:08 |
| Integration | rootless | overlay |    12:16 |          |    13:48 |
```

flouthoc:
Without this PR

```
>> 28:11
| type        | user     | fs      |      d13 |      f40 |      f41 |
| ----------- | -------- | ------- | -------- | -------- | -------- |
| Unit        | root     | vfs     |          |    08:14 |          |
| Unit        | root     | overlay |          |    08:07 |          |
| Conformance | root     | vfs     |    20:49 |          |          |
| Conformance | root     | overlay |    11:24 |          |          |
| Integration | root     | vfs     |    19:27 |    17:31 |    15:56 |
| Integration | root     | overlay |    15:04 |          |    20:43 |
| Integration | rootless | overlay |    16:05 |          |    16:27 |
```

flouthoc:
Strangely I see that some individual tests have gone down but for some reason PR https://github.com/containers/buildah/pull/5971 was faster than changes in this PR.

flouthoc:
Okay I have pushed a new version with hopefully fixes for flakes and speed up conformance tests.

internal/mkcw/embed/entrypoint_amd64.gz
Luap99, line 1: This should not be touched I assume, I don't get why we include a compiled binary in our source tree that anyone can just modify and commit.

imagebuildah/stage_executor.go
```diff
@@ -421,13 +421,20 @@ func (s *StageExecutor) performCopy(excludes []string, copies ...imagebuilder.Co
 				copiesExtend = append(copiesExtend, copyEntry)
 			}
 			copySources := []string{}
+			// Create seperate subdir for processing all heredoc instructions
+			// in this stage.
+			tmpDir, err := os.MkdirTemp(parse.GetTempDir(), "heredoc")
```
Luap99, line 426: maybe use `buildah-heredoc`? So if files get leaked users know where they came from.

tests/test_runner.sh
```diff
@@ -5,7 +5,7 @@ cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"
 
 # Default to using /var/tmp for test space, since it's more likely to support
 # labels than /tmp, which is often on tmpfs.
-export TMPDIR=${TMPDIR:-/var/tmp}
+export TMPDIR=${TMPDIR:-/tmp}
```
Luap99, line 8: The comment above says default /var/tmp so this no longer matches, I am not sure what the comment means by supporting labels, selinux? That should work on tmpfs so I don't think that is blocker

internal/mkcw/embed/entrypoint_amd64.gz
Luap99, line 1: in commit "mkcw: retry mount a few times" it is changed again

tests/mkcw.bats
```diff
@@ -7,7 +7,7 @@ function mkcw_check_image() {
   # Mount the container and take a look at what it got from the image.
   run_buildah from "$imageID"
   local ctrID="$output"
-  run_buildah mount "$ctrID"
+  run_buildah --retry-delay mount "$ctrID"
```
Luap99, line 10: This really seems like the tootally wrong fix, retry is for flakes we can do nothing about. 

Looking at the errors

> Device 5d2ea557-f97e-4a36-a8e6-6996cf68449e is still in use.

and

> unmount:  ... :target is busy


do not look like network flakes, this looks like a real logic problem if buildah just isn't working.

And then even more likely buildah mount is not failing with unmount is busy AFAICT, it seems much more likely that the cryptsetup/unmount command below fail. And since they are written as plain commands without any logging you do not see which command was exectued so that needs fixing first in the test.

Overall I don't think this command belongs in this PR.

Luap99 (changes requested):

flouthoc:
Before I remove some commits just timing this PR.

```
>> 26:57
| type        | user     | fs      |      d13 |      f40 |      f41 |
| ----------- | -------- | ------- | -------- | -------- | -------- |
| Unit        | root     | vfs     |          |    08:18 |          |
| Unit        | root     | overlay |          |    08:18 |          |
| Conformance | root     | vfs     |    20:46 |          |          |
| Conformance | root     | overlay |    13:02 |          |          |
| Integration | root     | vfs     |    11:55 |    16:20 |    13:08 |
| Integration | root     | overlay |    11:17 |          |    11:18 |
| Integration | rootless | overlay |    12:22 |          |    14:01 |
```

tests/helpers.bash
```diff
@@ -296,6 +296,16 @@ function rm() {
...
 }
 
+#################
+#  run_with_log  #  Logs command before running it
+#################
+#
+function run_with_log() {
+    local cmd="$*"
+    echo "$_LOG_PROMPT $cmd"
+    run -0 $cmd
```
Luap99, line 306: This breaks quoting of arguments, you need to use `run -0 "$@"` here to ensure the args always work correctly

Likely doesn't matter for the cases where you added it but if we add as general helper we should fix it.

flouthoc:
Okay I pushed quite a few times and the flake is gone, here are the new timing

```
>> 26:01
| type        | user     | fs      |      d13 |      f40 |      f41 |
| ----------- | -------- | ------- | -------- | -------- | -------- |
| Unit        | root     | vfs     |          |    07:47 |          |
| Unit        | root     | overlay |          |    07:53 |          |
| Conformance | root     | vfs     |    19:37 |          |          |
| Conformance | root     | overlay |    10:24 |          |          |
| Integration | root     | vfs     |    11:50 |    17:42 |    11:52 |
| Integration | root     | overlay |    11:02 |          |    10:59 |
| Integration | rootless | overlay |    11:56 |          |    14:00 |
```

flouthoc:
Screenshot From 2025-02-15 09-29-41

One cirrus it shows 24 mins for entire CI

tests/mkcw.bats
```diff
@@ -42,9 +42,10 @@ function mkcw_check_image() {
   fi
 
   # Clean up.
-  umount "$TEST_SCRATCH_DIR"/mount
-  cryptsetup luksClose "$uuid"
-  buildah umount "$ctrID"
+  run_with_log umount -f "$TEST_SCRATCH_DIR"/mount
+  run_with_log sleep 1
```
Luap99, line 46: instead of always adding `sleep 1` which always slows down the test and likely does not guarantee that sleep 1 is enough that should be turned into a loop. Try to luksClose and if it fails with EBUSY sleep 1 and retry a few times in a loop.

Also this really needs a comment why the sleep is needed, yes it is in the commit message but that is an extra step for readers to find, especially if the code in questions moves/is modified again.

Luap99 (approved):
LGTM

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
@flouthoc you may need a rebase here.

flouthoc:
> @flouthoc you may need a rebase here.

Done.

flouthoc:

imagebuildah/stage_executor.go
```diff
@@ -421,13 +421,20 @@ func (s *StageExecutor) performCopy(excludes []string, copies ...imagebuilder.Co
 				copiesExtend = append(copiesExtend, copyEntry)
 			}
 			copySources := []string{}
+			// Create seperate subdir for processing all heredoc instructions
+			// in this stage.
```
nalind, line 425: The syntax doesn't inherently prevent the basename from being used again with a different destination or command within the same stage, we might as well use a separate directory for each file before someone tries it.

nalind:
LGTM

flouthoc:
Rebased with `main`.

Issue #5968, opened 2025-02-04
--platform parameter spills out of its FROM clause


syedriko:
I have an image built on linux/amd64 that I need to use as a base, on amd64 and arm64. I am treating that image as data-only and simply need to copy some files from it to the resulting image. I was hoping the following would work:

```
ARG LIGHTSPEED_RAG_CONTENT_IMAGE=quay.io/openshift-lightspeed/lightspeed-rag-content@sha256:3e96332648a6f8ff1879c7ae11c818ea7f1c8d5b8a99c4bff406c98c8a7d4541
FROM --platform=linux/amd64 ${LIGHTSPEED_RAG_CONTENT_IMAGE} as lightspeed-rag-content
FROM registry.redhat.io/ubi9/ubi-minimal:latest
COPY --from=lightspeed-rag-content /rag/vector_db/ocp_product_docs ./vector_db/ocp_product_docs
RUN uname -m
```
, but it doesn't - on arm64, the

```
FROM registry.redhat.io/ubi9/ubi-minimal:latest
```

line picks up the amd64 flavor of the UBI9 base image.

flouthoc:
PR https://github.com/containers/buildah/pull/5971 should close this.

syedriko:
For completeness' sake, I'm doing
```
FROM --platform=$BUILDPLATFORM registry.redhat.io/ubi9/ubi-minimal:latest
```
as a workaround for now.

PR #5971, opened 2025-02-04
stage_executor: reset platform in systemcontext for every stage.

flouthoc:
On processing of every stage platform spec in systemcontext must be correctly reset.

[NO NEW TESTS NEEDED]

Closes: https://github.com/containers/buildah/issues/5968

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
stage_executor: reset platform in systemcontext for every stage.
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -59,6 +59,7 @@ import (
 // name to the image that it produces.
 type StageExecutor struct {
 	ctx                   context.Context
+	systemContext         *types.SystemContext
 	executor              *Executor
 	log                   func(format string, args ...interface{})
 	index                 int
@@ -584,8 +585,8 @@ func (s *StageExecutor) performCopy(excludes []string, copies ...imagebuilder.Co
 			// The values for these next two fields are ultimately
 			// based on command line flags with names that sound
 			// much more generic.
-			CertPath:              s.executor.systemContext.DockerCertPath,
-			InsecureSkipTLSVerify: s.executor.systemContext.DockerInsecureSkipTLSVerify,
+			CertPath:              s.systemContext.DockerCertPath,
+			InsecureSkipTLSVerify: s.systemContext.DockerInsecureSkipTLSVerify,
 			MaxRetries:            s.executor.maxPullPushRetries,
 			RetryDelay:            s.executor.retryPullPushDelay,
 			Parents:               copy.Parents,
@@ -841,7 +842,7 @@ func (s *StageExecutor) Run(run imagebuilder.Run, config docker.Config) error {
 		Stderr:               s.executor.err,
 		Stdin:                stdin,
 		Stdout:               s.executor.out,
-		SystemContext:        s.executor.systemContext,
+		SystemContext:        s.systemContext,
 		Terminal:             buildah.WithoutTerminal,
 		User:                 config.User,
 		WorkingDir:           config.WorkingDir,
@@ -966,19 +967,20 @@ func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBCo
 		}
 	}
 
-	builderSystemContext := s.executor.systemContext
-	// get platform string from stage
-	if stage.Builder.Platform != "" {
-		os, arch, variant, err := parse.Platform(stage.Builder.Platform)
+	// In a multi-stage build where `FROM --platform=<>` was used then we must
+	// reset context for new stages so that new stages don't inherit unexpected
+	// `--platform` from prior stages.
+	if stage.Builder.Platform != "" || (len(s.stages) > 1 && (s.systemContext.ArchitectureChoice == "" && s.systemContext.VariantChoice == "" && s.systemContext.OSChoice == "")) {
+		imageOS, imageArch, imageVariant, err := parse.Platform(stage.Builder.Platform)
 		if err != nil {
 			return nil, fmt.Errorf("unable to parse platform %q: %w", stage.Builder.Platform, err)
 		}
-		if arch != "" || variant != "" {
-			builderSystemContext.ArchitectureChoice = arch
-			builderSystemContext.VariantChoice = variant
+		if imageArch != "" || imageVariant != "" {
+			s.systemContext.ArchitectureChoice = imageArch
+			s.systemContext.VariantChoice = imageVariant
 		}
-		if os != "" {
-			builderSystemContext.OSChoice = os
+		if imageOS != "" {
+			s.systemContext.OSChoice = imageOS
 		}
 	}
 
@@ -992,7 +994,7 @@ func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBCo
 		BlobDirectory:         s.executor.blobDirectory,
 		SignaturePolicyPath:   s.executor.signaturePolicyPath,
 		ReportWriter:          s.executor.reportWriter,
-		SystemContext:         builderSystemContext,
+		SystemContext:         s.systemContext,
 		Isolation:             s.executor.isolation,
 		NamespaceOptions:      s.executor.namespaceOptions,
 		ConfigureNetwork:      s.executor.configureNetwork,
@@ -2058,7 +2060,7 @@ func (s *StageExecutor) tagExistingImage(ctx context.Context, cacheID, output st
 		return "", nil, err
 	}
 
-	policyContext, err := util.GetPolicyContext(s.executor.systemContext)
+	policyContext, err := util.GetPolicyContext(s.systemContext)
 	if err != nil {
 		return "", nil, err
 	}
@@ -2171,7 +2173,7 @@ func (s *StageExecutor) pushCache(ctx context.Context, src, cacheKey string) err
 			Compression:         s.executor.compression,
 			SignaturePolicyPath: s.executor.signaturePolicyPath,
 			Store:               s.executor.store,
-			SystemContext:       s.executor.systemContext,
+			SystemContext:       s.systemContext,
 			BlobDirectory:       s.executor.blobDirectory,
 			SignBy:              s.executor.signBy,
 			MaxRetries:          s.executor.maxPullPushRetries,
@@ -2209,7 +2211,7 @@ func (s *StageExecutor) pullCache(ctx context.Context, cacheKey string) (referen
 		options := buildah.PullOptions{
 			SignaturePolicyPath: s.executor.signaturePolicyPath,
 			Store:               s.executor.store,
-			SystemContext:       s.executor.systemContext,
+			SystemContext:       s.systemContext,
 			BlobDirectory:       s.executor.blobDirectory,
 			MaxRetries:          s.executor.maxPullPushRetries,
 			RetryDelay:          s.executor.retryPullPushDelay,
@@ -2431,7 +2433,7 @@ func (s *StageExecutor) commit(ctx context.Context, createdBy string, emptyLayer
 		SignaturePolicyPath:   s.executor.signaturePolicyPath,
 		ReportWriter:          writer,
 		PreferredManifestType: s.executor.outputFormat,
-		SystemContext:         s.executor.systemContext,
+		SystemContext:         s.systemContext,
 		Squash:                squash,
 		OmitHistory:           s.executor.commonBuildOptions.OmitHistory,
 		EmptyLayer:            emptyLayer,
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -6573,6 +6573,18 @@ _EOF
   done
 }
 
+@test "build must reset platform for stages if needed" {
+  run_buildah info --format '{{.host.arch}}'
+  myarch="$output"
+  otherarch="arm64"
+  # just make sure that other arch is not equivalent to host arch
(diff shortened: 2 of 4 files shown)


flouthoc:
Just wondering how to test this in CI since `binfmt` will correctly run every image type.

flouthoc:
Temp adding `No New Tests` to see if CI passes, will remove the label when I add test.

nalind:
The "header-builtin" test does something similar, so I'd suggest
```Dockerfile
ARG FOREIGNARCH
FROM --platform=linux/$FOREIGNARCH busybox AS foreign
FROM busybox
COPY --from=foreign /bin/busybox /bin/busybox.foreign
RUN arch
RUN ls -l /bin/busybox /bin/busybox.foreign
RUN ! cmp /bin/busybox /bin/busybox.foreign
```
Define the `FOREIGNARCH` build arg by checking if `$(go env GOARCH)` is `amd64` or `arm64` or `ppc64le` or `s390x`, having preselected one or them as a default, and choosing a different one if that default happens to match the local architecture.

imagebuildah/stage_executor.go
```diff
@@ -59,6 +59,7 @@ import (
 // name to the image that it produces.
 type StageExecutor struct {
 	ctx                   context.Context
+	systemContext         types.SystemContext
```
nalind, line 62: `types.SystemContext` values in structs are customarily pointers.  This should be a pointer.

imagebuildah/stage_executor.go
```diff
@@ -952,19 +953,21 @@ func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBCo
...
 
-	builderSystemContext := s.executor.systemContext
-	// get platform string from stage
-	if stage.Builder.Platform != "" {
-		os, arch, variant, err := parse.Platform(stage.Builder.Platform)
+	builderSystemContext := s.systemContext
+	// In a multi-stage build where `FROM --platform=<>` was used then we must
+	// reset context for new stages so that new stages don't inherit unexpected
+	// `--platform` from prior stages.
+	if stage.Builder.Platform != "" || (stage.Position != 0 && builderSystemContext.ArchitectureChoice == "" && builderSystemContext.VariantChoice == "" && builderSystemContext.OSChoice == "") {
```
nalind, line 960: Would this be simpler if the `builderSystemContext` had its `...Choice` fields cleared as part of initializing it?

tests/bud.bats
```diff
@@ -6486,6 +6486,15 @@ _EOF
   done
 }
 
+@test "build must reset platform for stages if needed" {
+  otherarch="arm64"
+  # just make sure that other arch is not equivalent to host arch
+  if [[ "$otherarch" == "$myarch" ]]; then
```
nalind, line 6492: `$myarch` has not been defined at this point, so the test is always false.

tests/bud.bats
```diff
@@ -6486,6 +6486,15 @@ _EOF
...
+    otherarch="amd64"
+  fi
+  run_buildah build $WITH_POLICY_JSON --build-arg FOREIGNARCH=$otherarch -f $BUDFILES/multiarch/Containerfile.reset-platform $BUDFILES/multiarch
```
nalind, line 6495: This should also test a case where the the TARGETPLATFORM build-arg is set, so that we don't regress on that.

imagebuildah/stage_executor.go
flouthoc: I don't think we can do that since in certain cases the fields are set beforehand for example when it's build with `--platform` or `--all-platforms` so clearing `...Choice` would reset values in such case.

tests/bud.bats
flouthoc: Added.

tests/bud.bats
flouthoc: I added a test which sets TARGETPLATFORM using `--platform`, should that cover what you meant by this comment ?

nalind: I mean literally setting `--build-arg TARGETPLATFORM=`, which has been known to be done instead of using the `--platform` flag.  This had to be fixed before in https://github.com/containers/buildah/pull/5731.

imagebuildah/stage_executor.go
nalind: I think there's a weird corner case in here when jobs > 1, skipUnusedStages is false, and one or more later stages are using the same base image as the first stage.

flouthoc:
I have to get back to @nalind 's comment since I was working on infra needed for this PR https://github.com/containers/common/pull/2339

flouthoc:
@nalind Could you PTAL again.

imagebuildah/stage_executor.go
nalind: Is the intention here that the first stage should default to being built for the default platform when `--platform` is not specified?  It's hard to tell.

flouthoc: I forgot to push a change, apologies. Let me push it again.

nalind: Anything that happens when `jobs > 1` can also happen when there are multiple builds running, as podman allows.  Given that, I don't think this recent addition actually fixes that third case.
There are also two cases where `len(s.stages) > 1` is evaluated, and that makes this line longer and, for me, that much harder to wrap my head around.

flouthoc: > There are also two cases where len(s.stages) > 1 is evaluated, and that makes this line longer and, for me, that much harder to wrap my head around.

I fixed it.

> Anything that happens when jobs > 1 can also happen when there are multiple builds running, as podman allows. Given that, I don't think this recent addition actually fixes that third case.

Yes patch in this PR only covers the case when single `buildah` invoked and multiple jobs are sharing same context variables. I suspect there could be more cases in `pull` then parallel invocation of multiple buildah can also step into some new issues but to be honest I did not test that.

flouthoc: Okay I two experiments one with `--skip-unused-stages=false` and one without with following containerfile

```Dockerfile
FROM --platform=linux/riscv64 quay.io/prometheus/busybox AS risc
FROM --platform=linux/amd64 quay.io/prometheus/busybox AS amd
FROM --platform=linux/arm64 quay.io/prometheus/busybox AS foreign
FROM quay.io/prometheus/busybox
COPY --from=foreign /bin/busybox /bin/busybox.foreign
COPY --from=amd /bin/busybox /bin/busybox.amd
RUN arch
RUN ls -l /bin/busybox /bin/busybox.foreign
RUN ! cmp /bin/busybox /bin/busybox.foreign
RUN cmp /bin/busybox /bin/busybox.amd
```

Experiment One
```console
for i in `seq 1 25`; do ./buildah build --iidfile id.$i . & done; wait; ls -al
```

Experiment Two
```console
for i in `seq 1 25`; do ./buildah build --jobs=2 --skip-unused-stages=false --iidfile id.$i . & done; wait; ls -al
```

I ran both experiments a few times I did not see any failure.

flouthoc: @nalind Should I open a different github issue (to track) to find verifying to check if `parallel` race exists or not. As I able not able to find any issues till now but wondering if this PR can get blocked because of that.

flouthoc: I created an issue to track this thing separately https://github.com/containers/buildah/issues/6104 so it's easier to move with this PR.

imagebuildah/stage_executor.go
```diff
@@ -966,19 +967,21 @@ func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBCo
...
-			builderSystemContext.ArchitectureChoice = arch
-			builderSystemContext.VariantChoice = variant
+		if imageArch != "" || imageVariant != "" {
+			builderSystemContext.ArchitectureChoice = imageArch
+			builderSystemContext.VariantChoice = imageVariant
 		}
-		if os != "" {
-			builderSystemContext.OSChoice = os
+		if imageOS != "" {
+			builderSystemContext.OSChoice = imageOS
```
nalind, line 984: Why are we creating `builderSystemContext`, which is a copy of the pointer value in `s.systemContext`, and modifying it through that pointer?  It has the same effect as modifying `s.systemContext` directly, but is harder to follow.  Is it intended to be a one-off copy that we just pass in the `BuilderOptions`?

flouthoc: Removed. This is not required since we are already copying context for each stage in `executor`.

tests/bud.bats
```diff
@@ -6573,6 +6573,21 @@ _EOF
...
+  run_buildah build $WITH_POLICY_JSON --build-arg TARGETPLATFORM=linux/$myarch --build-arg FOREIGNARCH=$otherarch -f $BUDFILES/multiarch/Containerfile.reset-platform $BUDFILES/multiarch
+  # Following line here attempts to cover edge case discussed here https://github.com/containers/buildah/pull/5971/files#r1970277844 but it's hard to reproduce this in a certain
+  # way since it can happen only when unused stage is picked in a particular order, following line ensures if this thing breaks then we luckily catch it in our CI.
+  run_buildah build $WITH_POLICY_JSON --jobs=2 --skip-unused-stages=true --build-arg FOREIGNARCH=$otherarch -f $BUDFILES/multiarch/Containerfile.unused-stage $BUDFILES/multiarch
```
nalind, line 6588: What is this `run_buildah` invocation testing, exactly?  The image that's produced isn't examined, and the `build` run's success only depends on the base image existing.

tests/bud.bats
```diff
@@ -6573,6 +6573,21 @@ _EOF
...
+  run_buildah info --format '{{.host.arch}}'
+  myarch="$output"
+  otherarch="arm64"
+  # just make sure that other arch is not equivalent to host arch
+  if [[ "$otherarch" == "$myarch" ]]; then
+    otherarch="amd64"
+  fi
+  run_buildah build $WITH_POLICY_JSON --build-arg FOREIGNARCH=$otherarch -f $BUDFILES/multiarch/Containerfile.reset-platform $BUDFILES/multiarch
+  run_buildah build $WITH_POLICY_JSON --build-arg TARGETPLATFORM=linux/$myarch --build-arg FOREIGNARCH=$otherarch -f $BUDFILES/multiarch/Containerfile.reset-platform $BUDFILES/multiarch
+  # Following line here attempts to cover edge case discussed here https://github.com/containers/buildah/pull/5971/files#r1970277844 but it's hard to reproduce this in a certain
```
nalind, line 6586: This URL no longer turns up a discussion.

imagebuildah/stage_executor.go
```diff
@@ -966,19 +967,20 @@ func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBCo
...
 	}
 
-	builderSystemContext := s.executor.systemContext
-	// get platform string from stage
-	if stage.Builder.Platform != "" {
-		os, arch, variant, err := parse.Platform(stage.Builder.Platform)
+	// In a multi-stage build where `FROM --platform=<>` was used then we must
+	// reset context for new stages so that new stages don't inherit unexpected
+	// `--platform` from prior stages.
+	if stage.Builder.Platform != "" || (len(s.stages) > 1 && ((s.systemContext.ArchitectureChoice == "" && s.systemContext.VariantChoice == "" && s.systemContext.OSChoice == "") || (s.executor.skipUnusedStages == types.OptionalBoolFalse))) {
```
nalind, line 973: In the current revision, `s.executor.skipUnusedStages` doesn't appear to matter and can likely be dropped, as the current revision ensures that OSChoice/ArchitectureChoice/VariantChoice is explicitly specified for every stage of a multi-stage build (earlier revisions only did so for the second and later stages), either to values given in the `--platform` flag for the FROM instruction, values passed in through our API, or the current system's defaults.

nalind (commented):
The conditional can probably be simplified.  Not sure what one of the `buildah build` runs in one of the tests is testing.

tests/bud.bats
flouthoc: This part is better covered in this issue https://github.com/containers/buildah/issues/6104 and I will add a different test once I am able to verify if issue exists.

imagebuildah/stage_executor.go
flouthoc: I removed this.

tests/bud/multiarch/Containerfile.unused-stage
```diff
@@ -0,0 +1,3 @@
+FROM busybox
+RUN arch
+FROM --platform=linux/$FOREIGNARCH busybox AS foreign
```
nalind, line 3: This file is no longer being used.

nalind (commented):
Drop the unused file, then this should be ready to merge.

tests/bud/multiarch/Containerfile.unused-stage
flouthoc: I have removed this.

flouthoc:
@TomSweeneyRedHat @rhatdan @Honny1 PTAL

Honny1 (approved):
LGTM
 
/approve

Issue #5952, opened 2025-01-29
buildah 1.38.1 build "FROM oci-archive:" failed


henrywang:
`quay.io/buildah/stable:v1.38.1` failed to build image with `FROM oci-archive:./out.ociarchive` and got error:
```
[2/2] STEP 1/3: FROM oci-archive:./out.ociarchive
Error: creating build container: creating temp directory: archive file not found: "/builds/redhat/rhel/bifrost/compose-bootc/out.ociarchive"
```

`quay.io/buildah/stable:v1.38.0` does not have this issue.

vrothberg:
Another [MR](https://gitlab.com/redhat/rhel/bifrost/compose-bootc/-/merge_requests/212/diffs) only bumping buildah [fails](https://gitlab.com/redhat/rhel/bifrost/compose-bootc/-/jobs/8985979871) as well.

cgwalters:
I think this is a regression from https://github.com/containers/buildah/commit/25a3b385afd1b4166d739afb665df57752c46ccd

This feature is really key as a way to generate chunked/reproducible containers and we've been using it for quite some time for rpm-ostree, and we have plans in the immediate future to promote this more for custom base images (ref https://gitlab.com/fedora/bootc/tracker/-/issues/32 and https://github.com/coreos/rpm-ostree/issues/5221 specifically)

Way back in https://github.com/coreos/rpm-ostree/issues/4688#issuecomment-2020442280 there was a soft commitment to support this.

My understanding is that there are security implications (not mentioned in the commit message) that motivated https://github.com/containers/buildah/commit/25a3b385afd1b4166d739afb665df57752c46ccd

Can we elaborate on those now that the commit is public (I don't see a corresponding PR with discussion)?

It seems to me that what we always wanted here is a way to write not to the build context, but just any internal filesystem whose lifetime is scoped to the entire build, right?

mheon:
The relevant CVE is https://github.com/advisories/GHSA-5vpc-35f4-r8w6

A core part of this was the ability to write to the host filesystem from mounts, which is why it was disabled - so what you desire is a large part of a high-severity CVE. I don't see us turning this back on. Adding an option with the understanding that enabling it is explicitly insecure and exposes the host to potential escape by malicious Dockerfile is a potential option, but that also feels very bad.

cgwalters:
> @henrywang do we have an easy reproducer?

[...]

Issue #5950, opened 2025-01-28
`RUN true` in `Containerfile` touches mtime of `/run`, creates layer


allisonkarlitskaya:
I found https://github.com/containers/podman/issues/14577 and https://github.com/containers/podman/pull/14582 related issues when searching for duplicates.

tl;dr: Running a read-only command in a `RUN` unexpectedly modifies the container image and creates a new layer.  Even if you add a workaround to prevent the modification, the layer will be created but it will be empty (the layer ID is equal to the SHA256 of an empty .tar file).  Each `RUN` creates an additional layer.

Consider this Containerfile:

```Containerfile
FROM fedora

RUN sleep 1
RUN stat /run
RUN sleep 1
RUN stat /run
```

You can see that each `stat` command reports a different mtime for `/run`.  I *think* that's because when we go to mount `/run/.containerenv` (for the duration of each `RUN` command) we need to create/delete the temporary file for the bindmount point in `/run`, which results in the mtime changing.

That also means that these presumably do-nothing commands are creating layers, as can be seen in the output of `podman image inspect` on the created container: you get 5 distinct layers added.  If you look at the first layer `.tar` (after the fedora base image) you'll see that it modifies `/etc/hostname` and `/etc/resolv.conf` along with `/run` but all of the other layers contain only `/run` with an updated timestamp, which I consider unexpected and undesirable.

In my particular usecase I'm running `RUN rm`, so I'm actually trying to modify the container image, but am getting `/run` changed as a side-effect, which is undesired.

As a workaround for my usecase, something like this does the trick:

```Containerfile
FROM fedora

RUN --mount=type=tmpfs,target=/run sleep 1
RUN --mount=type=tmpfs,target=/run stat /run
RUN --mount=type=tmpfs,target=/run sleep 1
RUN --mount=type=tmpfs,target=/run stat /run
```

and then `/run` won't get modified.

This *does* do something else strange, though: each of those `RUN` commands will create a new layer, but all layers will have the same ID: `5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef`.  That's the SHA-256 of 1024 nul characters.  That's the tar end-of-stream marker, so effectively this is creating 3 empty layers.  That should probably also be fixed.

allisonkarlitskaya:
See https://github.com/containers/buildah/issues/4242 about the mentioned files in `/etc`.  That's definitely a different issue, but it's good to know that it's being tracked as well.

[...]

Issue #4242, opened 2022-09-13
`podman build` injects empty `/etc/resolv.conf` into generated layers


cgwalters:
**Description**

```
$ rpm -q podman
podman-4.2.0-2.fc36.x86_64
$ cat Dockerfile
FROM quay.io/fedora/fedora:36
RUN touch /etc/blah
$ podman build -t localhost/test .
$ skopeo copy containers-storage:localhost/test oci:test:latest
$ tar tvf test/blobs/sha256/e2deb0a4365096c1259670b1b825a6574a0db6c6aaab74fdb37c6561d48dd886 
drwxr-xr-x root/root         0 2022-09-13 19:22 etc/
-rw-r--r-- root/root         0 2022-09-13 19:22 etc/blah
-rwx------ root/root         0 2022-09-13 19:22 etc/resolv.conf
```

**Describe the results you received:**

Empty `/etc/resolv.conf`

**Describe the results you expected:**

No such file.  For work on https://fedoraproject.org/wiki/Changes/OstreeNativeContainer we want to have clean content.  This issue is very similar to https://github.com/containers/buildah/issues/3523

**Output of `rpm -q buildah` or `apt list buildah`:**

```
(paste your output here)
```

**Output of `buildah version`:**

```
(paste your output here)
```

**Output of `podman version` if reporting a `podman build` issue:**

```
Client:       Podman Engine
Version:      4.2.0
API Version:  4.2.0
Go Version:   go1.18.4
Built:        Thu Aug 11 14:42:17 2022
OS/Arch:      linux/amd64

```

**Output of `cat /etc/*release`:**

```
NAME="Fedora Linux"
VERSION="36.20220909.dev.1 (CoreOS)"
```

cgwalters:
(Just like https://github.com/containers/buildah/issues/3523 - `docker build` doesn't reproduce this problem)

cgwalters:
Ah sorry now that I dig a bit this was discussed in https://github.com/containers/buildah/pull/3525 but not resolved there.

I do think we should also fix `/etc` here.

rhatdan:
One issue with removing the inode is it prevents the use of readonly containers.
I think Docker has two overlays mounted. The intermediate overlay is for changes it makes.

vrothberg:
@rhatdan @nalind we'd love to get this issue ironed out for Image Mode.  Do we have a path forward to fix the issue?

rhatdan:
[...]

PR #5975, opened 2025-02-05
`buildah build`: use the same overlay for the context directory for the whole build

nalind:
#### What type of PR is this?

/kind feature

Change how we handle bind mounts of the build context directory from using a different upper for each read-write mount into it, to using one that lasts the duration of the full build.

We broke workflows which wrote content to the build context in one stage and then attempted to use archive or layout locations in the build context in a subsequent stage when we fixed CVE-2024-11218, and this should get them working again.  Because that content was often addressed using relative path names which are no longer correct for the written location, we try to compensate by rewriting some types of references and straight-up rejecting some others.

#### How to verify it

New integration test!  And some updated ones, too!

Proposed fix for #5952.

```release-note
Changes "written" to the build context directory during `buildah build` `RUN --mount=type=bind` instructions are no longer discarded between instructions, but when the build completes, whether it completes successfully or not.
```


pkg/overlay/overlay_linux.go
```diff
@@ -63,6 +63,13 @@ func MountWithOptions(contentDir, source, dest string, opts *Options) (mount spe
 			if err := os.Chown(upperDir, int(stat.Uid), int(stat.Gid)); err != nil {
 				return mount, err
 			}
+			times := []syscall.Timeval{
```
cgwalters, line 83: Why out of curiosity? Is this helping ensure reproducibility? We aren't serializing anything related to the timestamps of this directory to e.g. a tar stream are we?

imagebuildah/build_linux.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+// platformSetupContextDirectoryOverlay() may set up an overlay _over_ the
+// build context directory, and sorts out labeling.  Returns either the new
+// location which should be used as the base build context or the old location;
+// the process label and mount label for the build; a boolean value that
+// indicates whether we did, in fact, mount an overlay; a cleanup function
+// which should be called when the location is no longer needed (on success);
+// and a non-nil fatal error if any of that failed.
+func platformSetupContextDirectoryOverlay(store storage.Store, options *define.BuildOptions) (string, string, string, bool, func(), error) {
```
cgwalters, line 27: Return values are a bit unwieldy, maybe a `struct` instead? Although I guess the single caller unpacks them anyways, so doesn't matter. Yeah, one caller, so makes sense as is.

pkg/cli/build.go
```diff
@@ -113,7 +113,7 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
 	if c.Flag("build-context").Changed {
 		for _, contextString := range iopts.BuildContext {
 			av := strings.SplitN(contextString, "=", 2)
-			if len(av) > 1 {
+			if len(av) > 1 && av[0] != "" && av[1] != "" {
```
cgwalters, line 116: This is basically so we skip over `-v :` or something? This could use a comment, maybe even we have a `const DefaultBuildContextKey = ""` and then reference it here?

tests/bud.bats
```diff
@@ -673,7 +673,7 @@ symlink(subdir)"
   _prefetch busybox
   run_buildah 125 build -t testbud3 $WITH_POLICY_JSON $BUDFILES/dockerignore3
   expect_output --substring 'building.*"COPY test1.txt /upload/test1.txt".*no such file or directory'
-  expect_output --substring $(realpath "$BUDFILES/dockerignore3/.dockerignore")
+  expect_output --substring 'filtered out using /[^ ]*/.dockerignore'
```
cgwalters, line 676: It's not obvious to me how this test change is related...did the build there try to modify the context directory?

cgwalters (commented):
Thanks so much for working on this! I just skimmed the code, will try testing it soon

pkg/overlay/overlay_linux.go
nalind: Without it, two different runs would appear to have different timestamps on the top level of the build context directory, breaking one of the tests added by #5691.

tests/bud.bats
nalind: The name of the directory which contains the .dockerignore file and which is included in the error message is no longer the one passed on the command line, so the exact string comparison began failing.  It was either change the test to not care about the specific path, or lie about the file's location to get the test to pass as it was, so I changed the test.

pkg/cli/build.go
nalind: Pushing the check on `av[1]` down into `GetAdditionalBuildContext()` makes it a little easier to follow, added a comment here and put more checks in there.

imagebuildah/stage_executor.go
```diff
@@ -916,6 +929,58 @@ func (s *StageExecutor) UnrecognizedInstruction(step *imagebuilder.Step) error {
 	return errors.New(err)
 }
 
+// do our best to ensure that image specifiers that include a transport that
+// uses path names are scoped to the build context directory
```
cgwalters, line 933: It'd be secure and reliable to get a directory fd for the context overlay mount, and then use `openat(..., RESOLVE_IN_ROOT)` to access the file, then...the ugly part is that the containers-storage APIs want path strings still, but we could pass e.g. `oci:/proc/self/fd/N` right? I've done that elsewhere I think.

cgwalters:
Trying to test this, but unrelated to the contents of this PR, I get the same problem trying git main:

```
$ ./bin/buildah from busybox
busybox-working-container
$ ./bin/buildah run busybox-working-container echo true
error running container: did not get container start message from parent: EOF
Error: setup network: pasta failed with exit code 1:
Couldn't open network namespace /proc/126984/ns/net: Permission denied
```

```
DEBU[0000] Running ["/usr/bin/crun" "create" "--bundle" "/var/tmp/buildah1984752272" "--pid-file" "/var/tmp/buildah1984752272/pid" "--no-new-keyring" "buildah-buildah1984752272"] 
DEBU[0000] waiting for parent start message             
DEBU[0000] pasta arguments: --config-net --dns-forward 169.254.1.1 -t none -u none -T none -U none --no-map-gw --quiet --netns /proc/126381/ns/net --map-guest-addr 169.254.1.2 
DEBU[0000] "/var/tmp/buildah1984752272/mnt/buildah-bind-target-10" is apparently not really mounted, skipping 
DEBU[0000] "/var/tmp/buildah1984752272/mnt/rootfs" is apparently not really mounted, skipping 
DEBU[0000] "/var/tmp/buildah1984752272/mnt" is apparently not really mounted, skipping 
error running container: did not get container start message from parent: EOF
DEBU[0000] Error building at step {Env:[PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin] Command:run Args:[touch /blah] Flags:[] Attrs:map[] Message:RUN touch /blah Heredocs:[] Original:RUN touch /blah}: setup network: pasta failed with exit code 1:
Couldn't open network namespace /proc/126381/ns/net: Permission denied 
Error: building at STEP "RUN touch /blah": setup network: pasta failed with exit code 1:
Couldn't open network namespace /proc/126381/ns/net: Permission denied
```

Is there something I need to do here to sync up the buildah binary with the default network config? Odd...

cgwalters:
Nevermind of course after I posted my "wait these are weird denials" mental flag tripped and of course it's SELinux; a `chcon --reference /usr/bin/buildah bin/buildah` fixes it.

imagebuildah/stage_executor.go
nalind: Proper safety for such references is a much larger patch.

cgwalters:
OK so this works for me:

```
FROM registry.access.redhat.com/ubi9/ubi:latest  as builder
RUN --mount=type=bind,src=.,rw,dst=/buildcontext \
  dnf -y install skopeo && skopeo copy docker://busybox oci:/buildcontext/out.oci

FROM oci:./out.oci
# Need to reference builder here to force ordering.
RUN --mount=type=bind,from=builder,src=.,target=/var/tmp true
```

So that's already a lot cleaner, thanks!

---

There is an important note here that this does break compatibility with the [previous dockerfiles](https://github.com/containers/buildah/issues/5952#issuecomment-2622495832), I get:

```
Error: building at STEP "RUN --mount=type=bind,rw=true,src=.,dst=/buildcontext,bind-propagation=shared dnf -y install skopeo && skopeo copy docker://busybox oci:/buildcontext/out.oci": resolving mountpoints for container "7f7333013ba83fbe05c78aad0ef4edd17663317f4daa775e549ac11d28412a90": rw: must not provide an argument for option
```

Looks like changing `rw=true` to just `rw` fixes things. Not sure if that's intentional or not? It does look like there's no mention of `rw=true` in the upstream Dockerfile syntax.

---

Now that I dig in a bit more, one unfortunate thing here is that the need to use the `RUN` to force ordering...has meant this whole time we're subject to all the "injected content" bugs like https://github.com/containers/buildah/issues/4242 and https://github.com/containers/buildah/issues/5950 etc.

When I unpack and inspect the resulting image, we have this final layer with
```
drwxr-xr-x 0/0               0 2025-02-06 16:44 etc/
-rwx------ 0/0               0 2025-02-06 16:44 etc/hostname
-rwx------ 0/0               0 2025-02-06 16:44 etc/hosts
-rwx------ 0/0               0 2025-02-06 16:44 etc/resolv.conf
drwxr-xr-x 0/0               0 2025-02-06 16:44 proc/
drwxr-xr-x 0/0               0 2025-02-06 16:44 run/
drwxr-xr-x 0/0               0 2025-02-06 16:44 sys/
drwxr-xr-x 0/0               0 2025-02-06 16:44 var/
drwxr-xr-t 0/0               0 2025-02-06 16:44 var/tmp/
```

with floating timestamps. That said hmm, actually in this case actually it works to do `buildah build --timestamp=<pinned>` and that won't also change all the timestamps on the inherited layers. So that's a good mitigation.

But the sad thing is that these things just keep piling on...now for me we have a *new* one of these in Konflux which is [injecting content_sets](https://github.com/konflux-ci/build-definitions/pull/1771) which also has no attempt made to generate timestamp-reproducible data right now either (and it uses buildah too, so it inherits that, but we still need to ensure the timestamps of the files that we *actually* write and aren't synthetic container runtime artifacts are canonicalized too)

So in the end, I think we'll definitely try to

nalind:
> There is an important note here that this does break compatibility with the [previous dockerfiles](https://github.com/containers/buildah/issues/5952#issuecomment-2622495832), I get:
> 
> ```
> Error: building at STEP "RUN --mount=type=bind,rw=true,src=.,dst=/buildcontext,bind-propagation=shared dnf -y install skopeo && skopeo copy docker://busybox oci:/buildcontext/out.oci": resolving mountpoints for container "7f7333013ba83fbe05c78aad0ef4edd17663317f4daa775e549ac11d28412a90": rw: must not provide an argument for option
> ```
> 
> Looks like changing `rw=true` to just `rw` fixes things. Not sure if that's intentional or not? It does look like there's no mention of `rw=true` in the upstream Dockerfile syntax.

Not being strict about the arguments for `--mount` contributed to CVE-2024-9407.  I added more checks in that area in #5925, and they would have landed in v1.39.0.

imagebuildah/stage_executor.go
```diff
@@ -546,7 +546,7 @@ func (s *StageExecutor) performCopy(excludes []string, copies ...imagebuilder.Co
 				}
```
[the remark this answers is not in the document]
nalind: Dang it, that's embarrassing.  Fixing.

nalind:

imagebuildah/stage_executor.go
```diff
@@ -980,7 +1286,7 @@ func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBCo
 
 	builderOptions := buildah.BuilderOptions{
 		Args:                  ib.Args,
-		FromImage:             from,
+		FromImage:             sanitizedFrom,
```
flouthoc, line 1289: A small question since I did not understand it completely, why are are we creating copy of entire image for every stage ? Is it just to make sure no symbolic links are present ?

cgwalters: Discussed here https://github.com/containers/buildah/pull/5975#discussion_r1945461316 and again here https://github.com/containers/buildah/pull/5975#discussion_r1970069046

imagebuildah/stage_executor.go
```diff
@@ -926,6 +935,290 @@ func (s *StageExecutor) UnrecognizedInstruction(step *imagebuilder.Step) error {
 	return errors.New(err)
 }
 
+// sanitizeImageReference limits which image transports we'll accept.  For
+// those it accepts which refer to filesystem objects, where relative path
+// names are evaluated relative to "contextDir", it will create a copy of the
+// original image, under "tmpdir", which contains no symbolic links.
+func (s *StageExecutor) sanitizeFrom(from, tmpdir string) (newFrom string, err error) {
```
flouthoc, line 942: Could we document what is value contained in `newFrom` string ?

imagebuildah/stage_executor.go
```diff
@@ -926,6 +935,290 @@ func (s *StageExecutor) UnrecognizedInstruction(step *imagebuilder.Step) error {
...
+		return "", fmt.Errorf("parsing image name %q: %w", from, err)
+	}
+	// TODO: drop this part and just return an error... someday
+	return sanitizeImageName(transportName, restOfImageName, s.executor.contextDir, tmpdir)
+}
+
+// sanitizeImageReference limits which image transports we'll accept.  For
+// those it accepts which refer to filesystem objects, where relative path
+// names are evaluated relative to "contextDir", it will create a copy of the
+// original image, under "tmpdir", which contains no symbolic links.
```
flouthoc, line 962: Could we document what is value contained in `newFrom` string ?

imagebuildah/stage_executor.go
flouthoc: Would it be fair to ask and see about how much overhead does this patch adds both `disk` usage wise and `speed` time. 

I could be wrong but I could think of use-case where a stage copied lot of external artifacts but only parts of it are being used in further stages, this will end up creating copies for every future stage even though those parts are not being used by future stages.

Should we consider making this behavior optional ?

imagebuildah/stage_executor.go
```diff
@@ -922,6 +935,290 @@ func (s *StageExecutor) UnrecognizedInstruction(step *imagebuilder.Step) error {
...
+func sanitizeImageName(transportName, restOfImageName, contextDir, tmpdir string) (newFrom string, err error) {
```
flouthoc, line 963: Maybe as a refactor we can break this function into several helper function for each `transport` it will help with the readability and maintenance.

imagebuildah/stage_executor.go
nalind: > Would it be fair to ask and see about how much overhead does this patch adds both `disk` usage wise and `speed` time.

The disk cost is a second copy of the base image being copied to a temporary location.
The speed cost is the cost of copying that content, which is unavoidably higher if we're using fuse-overlayfs, and likewise reading contents from the build context through that overlay to avoid inconsistencies.

> I could be wrong but I could think of use-case where a stage copied lot of external artifacts but only parts of it are being used in further stages, this will end up creating copies for every future stage even though those parts are not being used by future stages.

I don't understand what you're describing here.  Can you elaborate on that?  Initializing a stage using a previous stage's result as its base doesn't involve the build context directory, so I don't follow where additional copies would appear.

> Should we consider making this behavior optional ?

Using an overlay over the build context directory is not something we should turn off.  Turning off sharing like this PR introduces doesn't reduce any of the costs that are incurred when they're not shared, so I don't really see the benefit of adding complexity there.

flouthoc: > I don't understand what you're describing here. Can you elaborate on that? Initializing a stage using a previous stage's result as its base doesn't involve the build context directory, so I don't follow where additional copies would appear.

I was thinking for use-case something like this

``` Dockerfile
FROM alpine as first
COPY /large-fileA-from-host .
COPY /large-fileB-from-host .
COPY /large-fileC-from-host .


FROM first as second 
# Use only large-file-A in this stage
# but this will create another copy of `first` containing all three files

FROM first as third 
# Use only large-file-B in this stage
# # but this will create another copy of `first` containing all three files

FROM first as fourth 
# Use only large-file-C in this stage
# but this will create another copy of `first` containing all three files
```

@nalind If i understand correctly for all the three **stages** `second`, `third` and `fourth` a separate copy of `first` stage will be created , am i correct ? and I am not sure if buildah was doing this before.

nalind: The stages nicknamed "second", "third", and "fourth" would each start with their own working container based on the image that was committed at the end of the stage nicknamed "first", as before.  This PR doesn't affect how we handle the root filesystems of working containers, or committing images.

The COPY instructions have to read contents through the mounted overlay for consistency's sake, but I don't see additional copies of them being made here.  Since they're not being modified, they'd be read directly from the build context directory that's being used as the "lowerdir" for the overlay, without being "copied up".

flouthoc: Ah i see sounds fair to me.

imagebuildah/build_linux.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/containers/buildah/internal/tmpdir"
+	"github.com/containers/buildah/pkg/overlay"
+	"github.com/containers/storage"
+	"github.com/opencontainers/selinux/go-selinux/label"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+)
+
+// platformSetupContextDirectoryOverlay() may set up an overlay _over_ the
+// build context directory, and sorts out labeling.  Returns either the new
```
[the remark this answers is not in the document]
nalind: You're right, it never returns the original location.  Updating the godoc.

imagebuildah/build_notlinux.go
```diff
@@ -0,0 +1,19 @@
...
+package imagebuildah
+
+import (
+	"github.com/containers/buildah/define"
+	"github.com/containers/storage"
+)
+
+// platformSetupContextDirectoryOverlay() may set up an overlay _over_ the
+// build context directory, and sorts out labeling.  Returns either the new
+// location which should be used as the base build context or the old location;
```
flouthoc, line 12: Same as https://github.com/containers/buildah/pull/5975/files#r1981811503 I think comment needs to be fixed.

imagebuildah/stage_executor.go
nalind: Adding.

imagebuildah/stage_executor.go
nalind: Adding.

cevich:
Would this address #5988 also or is that something different?

internal/volumes/volumes.go
```diff
@@ -248,6 +249,7 @@ func GetBindMount(sys *types.SystemContext, args []string, contextDir string, st
 		if additionalMountPoints != nil {
 			if val, ok := additionalMountPoints[fromWhere]; ok {
 				mountPoint = val.MountPoint
+				skipOverlay = val.IsWritesDiscardedOverlay
```
mtrmac, line 252: Isn’t this ~always `false` because `fromWhere != ""`? OTOH maybe the consistency is nice.

pkg/parse/parse.go
```diff
@@ -208,19 +213,34 @@ func CommonBuildOptionsFromFlagSet(flags *pflag.FlagSet, findFlagFunc func(name
 
 // GetAdditionalBuildContext consumes raw string and returns parsed AdditionalBuildContext
 func GetAdditionalBuildContext(value string) (define.AdditionalBuildContext, error) {
```
mtrmac, line 215: Locally, it’s not obvious that the checks for `""` are important; probably worth a more detailed comment.

mtrmac (commented):
A few misc. comments, but I’m generally unfamiliar with the codebase and structure, so I didn’t really check lifetimes / interactions / … .

The `StageMountPoints[""]` case worries me. Would adding an extra field (`StageContextMount *StageMountDetails // nil if N/A`) work?

(Generally the functions have too many unnamed parameters and unnamed return values for my personal taste, but this is totally up to maintainers of this project.)

imagebuildah/stage_executor.go
```diff
@@ -922,6 +935,328 @@ func (s *StageExecutor) UnrecognizedInstruction(step *imagebuilder.Step) error {
...
+// the original image reference or a reference to a sanitized copy which should
+// be used instead.
+func (s *StageExecutor) sanitizeFrom(from, tmpdir string) (newFrom string, err error) {
+	transportName, restOfImageName, maybeHasTransportName := strings.Cut(from, ":")
+	if !maybeHasTransportName || transports.Get(transportName) == nil {
+		if _, err = reference.ParseNormalizedNamed(from); err == nil {
+			// this is a normal-looking image-in-a-registry-or-named-in-storage name
+			return from, nil
+		}
+		if img, err := s.executor.store.Image(from); img != nil && err == nil {
```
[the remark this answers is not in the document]
nalind: Internally, we rewrite references to nicknames of previous stages to their IDs before we get to this point, since they don't get names assigned to them.

imagebuildah/stage_executor.go
```diff
@@ -922,6 +935,328 @@ func (s *StageExecutor) UnrecognizedInstruction(step *imagebuilder.Step) error {
...
+			// resolve the target of the symlink
+			linkname := hdr.Linkname
+			if !path.IsAbs(linkname) {
+				linkname = path.Join("/"+path.Dir(hdr.Name), linkname)
+			}
+			linkname = path.Clean(linkname)
+			if linkname != "/" {
+				linkname = strings.TrimPrefix(linkname, "/")
+			}
+			if _, validTarget := seenEntries[linkname]; !validTarget {
```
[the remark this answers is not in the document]
nalind: Not generally, but in the very specific case of docker-archives, that does appear to be something we can safely expect, at least from containers/image.

imagebuildah/stage_executor.go
```diff
@@ -922,6 +935,328 @@ func (s *StageExecutor) UnrecognizedInstruction(step *imagebuilder.Step) error {
...
+		hdr.Name = path.Clean("/" + hdr.Name)
+		if hdr.Name != "/" {
+			hdr.Name = strings.TrimPrefix(hdr.Name, "/")
+		}
+		seenEntries[hdr.Name] = struct{}{}
+		switch hdr.Typeflag {
+		case tar.TypeDir, tar.TypeReg, tar.TypeLink:
+			if err := tw.WriteHeader(hdr); err != nil {
+				return fmt.Errorf("rewriting archive of base image: %w", err)
+			}
```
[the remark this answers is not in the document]
nalind: Moving it.

imagebuildah/stage_executor.go
```diff
@@ -922,6 +935,328 @@ func (s *StageExecutor) UnrecognizedInstruction(step *imagebuilder.Step) error {
...
+			linkName := path.Clean("/" + hdr.Linkname)
+			oldName := filepath.Join(root, filepath.FromSlash(linkName))
+			err = os.Link(oldName, newName)
+		case tar.TypeSymlink: // convert it to a hard link
+			var oldName string
+			if !path.IsAbs(hdr.Linkname) {
+				linkName := path.Join("/"+path.Dir(hdr.Name), hdr.Linkname)
+				oldName = filepath.Join(root, filepath.FromSlash(linkName))
+			} else {
+				oldName = filepath.Join(root, filepath.FromSlash(hdr.Linkname))
```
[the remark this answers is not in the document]
nalind: It should already be an absolute path, but that can't hurt.

imagebuildah/stage_executor.go
```diff
@@ -922,6 +935,328 @@ func (s *StageExecutor) UnrecognizedInstruction(step *imagebuilder.Step) error {
...
+			logrus.Warn(err)
+		}
+	}()
+	tr := tar.NewReader(imageArchive)
+	hdr, err := tr.Next()
+	for hdr != nil {
+		// if the archive we're parsing is expected to have an archive as its only item,
+		// assume it's the first (and hopefully, only) item, and switch to stepping through
+		// it as the archive
+		if isEmbeddedArchive {
```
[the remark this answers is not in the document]
nalind: We don't know that these locations contain contents we can trust, so the goal is to filter out things that we know would be problematic or even out of the ordinary.  I can't claim it's perfect, though.

pkg/parse/parse.go
nalind: Adding a note there.

flouthoc (approved):
LGTM

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
nalind: Ah, good point.

mtrmac (commented):
The resolution of my earlier comments LGTM.

I don’t know whether that should count as a review… If necessary, ping me and I’ll return to this and actually study the `imagebuildah` structure as a whole.

internal/volumes/volumes.go
nalind: Oh, right, we haven't mounted an overlay there yet when it's something else.  You're reading it right.

TomSweeneyRedHat:
LGTM
@nalind Looks like the branch needs an update.

nalind:
This is going to exacerbate #5988 for people who are running into it.
/hold

dustymabe:
hey @nalind is there any way to unblock this? this is still an important use case for us on the CoreOS side.

nalind:
I'm working through adding something like https://github.com/containers/buildah/issues/5988#issuecomment-2718307144, but other things have been clamoring for attention in the meanwhile.

dustymabe:
Is this just blocked on code review at this point? Or something else?

nalind:
@dustymabe I'd say that's accurate.

imagebuildah/build.go
```diff
@@ -204,6 +204,15 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 		options.AdditionalBuildContexts = make(map[string]*define.AdditionalBuildContext)
 	}
 
+	contextDirectory, processLabel, mountLabel, usingContextOverlay, cleanupOverlay, err := platformSetupContextDirectoryOverlay(store, &options)
```
mtrmac: Is creating an overlay here, i.e. sharing it across all single-platform builds, the desired behavior? Or should that happen in `buildDockerfilesOnce`?

On the one hand, sharing one overlay across everything is closer to what would naturally happen with `mount --type=bind` of the context directory..
On the other hand, things like https://github.com/containers/buildah/pull/5975#issuecomment-2641136554 would break for multi-platform builds.

Maybe there’s no right answer to this? (Is that equivalent to there being **EDIT <s>to</s> no** wrong answer, as well??)

imagebuildah/stage_executor.go
```diff
@@ -528,7 +529,7 @@ func (s *StageExecutor) performCopy(excludes []string, copies ...imagebuilder.Co
 				}
 				contextDir = mountPoint
 			}
-			// Original behaviour of buildah still stays true for COPY irrespective of additional context.
+			// This isn't --from the build context directory, so we don't want to force everything to 0:0
```
mtrmac: I don’t understand (most likely due to my ignorance of Buildah features). Can `--from` ever refer to the build context directory? (Also I think this line is reached on ~every `--from`.)

imagebuildah/stage_executor.go
```diff
@@ -938,6 +967,19 @@ func (s *StageExecutor) prepare(ctx context.Context, from string, initializeIBCo
 		}
 		from = base
 	}
+	sanitizedDir, err := os.MkdirTemp(tmpdir.GetTempDir(), "buildah-context-")
```
mtrmac: AFAICS this is not used for anything.

pkg/overlay/overlay_linux.go
```diff
@@ -61,6 +63,28 @@ func MountWithOptions(contentDir, source, dest string, opts *Options) (mount spe
...
+				}
+				if !overflowed {
+					return mount, err
+				}
+			}
+			times := []syscall.Timeval{
+				syscall.NsecToTimeval(syscall.TimespecToNsec(stat.Atim)),
+				syscall.NsecToTimeval(syscall.TimespecToNsec(stat.Mtim)),
+			}
+			if err := syscall.Utimes(upperDir, times); err != nil {
```
mtrmac: `syscall.UtimesNano`? We already depend on the kernel feature via `LUtimesNano`.

internal/sanitize/sanitize.go
```diff
@@ -0,0 +1,329 @@
...
+		if err != nil {
+			return fmt.Errorf("computing relative path from %q to %q in archive", hdr.Name, linkname)
+		}
+		rel = filepath.ToSlash(rel)
+		if convertSymlinksToHardlinks {
+			// rewrite as a hard link for oci-archive, which gets
+			// extracted into a temporary directory to be read, but
+			// not for docker-archive, which is read directly from
+			// the unextracted archive file, in a way which doesn't
+			// understand hard links
```
mtrmac, line 100: (BTW this could be fixed if it made a difference.)

internal/sanitize/sanitize.go
```diff
@@ -0,0 +1,329 @@
...
+	var tw *tar.Writer
+	var archiveSource string
+	var imageArchive io.ReadCloser
+	switch transportName {
+	case dockerTransport.Transport.Name(), "docker-daemon", openshiftTransport.Transport.Name(): // ok, these are all remote
+		return transportName + ":" + restOfImageName, nil
+	case dockerArchiveTransport.Transport.Name(), ociArchiveTransport.Transport.Name(): // these are, basically, tarballs
+		// these take the form path[:stuff]
+		transportRef := restOfImageName
+		parts := strings.Split(transportRef, ":")
```
mtrmac: (`strings.Cut` would simplify this. Also in the OCI layout case.)

internal/sanitize/sanitize.go
```diff
@@ -0,0 +1,329 @@
...
+
+	// start reading the archived content
+	defer func() {
+		if err := imageArchive.Close(); err != nil {
+			logrus.Warn(err)
+		}
+	}()
+	tr := tar.NewReader(imageArchive)
+	hdr, err := tr.Next()
+	for hdr != nil {
```
mtrmac: It’s not _quite always_ the case that `hdr != nil` is exclusive with `err != nil`; as a general convention, I think it’s not guaranteed; and specifically for `tar.Reader.Next()`, there is a documented case where (non-`nil`, non-`nil`) can happen.

Right now, for `ErrInsecurePath`, I think we should be fine, but consider being more defensive here.

internal/sanitize/sanitize.go
```diff
@@ -0,0 +1,329 @@
...
+		// if the archive we're parsing is expected to have an archive as its only item,
+		// assume it's the first (and hopefully, only) item, and switch to stepping through
+		// it as the archive
+		if isEmbeddedArchive {
```
mtrmac, line 289: (Sanity-check that the item is a regular file? Probably unnecessary, the tests would notice if the output of `newSingleItemArchive` changed.)

mtrmac (commented):
I didn’t read all of `imagebuildah`, but hopefully enough to have an opinion.

The mechanism LGTM overall; on semantics, see the `imagebuildah/build` comment.

---

I generally think the `mount` && `FROM oci:` use of the context directory is surprising (perhaps the use of transports in `FROM` is outright unexpected?) and aesthetically/intuitively I’m worried, and not inclined to allow that. I suppose that’s neither here nor there.

But with the uncertainty about the scope of the overlay (see `imagebuildah/buildh.go`), that only exacerbates the feeling.

Would we consider explicitly defining a specific, limited-scope feature, instead?

That was extensively litigated in https://github.com/containers/buildah/issues/5952 … and I agree it seems hard to do if we are limited to not add any new syntax — but that also might be a fairly handicapping requirement.

Anyway, that’s for actual Buildah maintainers to decide. I’m here just for the code changes.

But then again … even the _concept_ of a restriction in `StageExecutor.sanitizeFrom` is already so important an improvement, that a part of me wants to merge this PR just for that…

---

(Noting that this already has a LGTM by @flouthoc and @TomSweeneyRedHat , 2 maintainers in this repo — do either of you want to hit the button?)

TomSweeneyRedHat:
@nalind this has conflicts now, and I'm not sure you have addressed the lst bits of comments from @mtrmac

imagebuildah/build.go
nalind: Oh... we'll have to decide here.  Making it per-platform is _probably_ reasonable, since that's the effect someone would see if instead of specifying multiple `--platform` values, they invoked `buildah build` multiple times.

imagebuildah/stage_executor.go
nalind: The default build context is where it looks when a `--from` argument isn't specified, but there's no value that could be passed to `--from` to have the same effects.

imagebuildah/stage_executor.go
nalind: Oops, dropping it.

pkg/overlay/overlay_linux.go
nalind: Switching to that.

internal/sanitize/sanitize.go
nalind: Changing these.

internal/sanitize/sanitize.go
nalind: Adding logic to check for a non-nil `err` for the cases where we're not switching to reading a nested archive.

internal/sanitize/sanitize.go
nalind: Sure, adding one.

pkg/overlay/overlay_linux.go
```diff
@@ -61,6 +63,28 @@ func MountWithOptions(contentDir, source, dest string, opts *Options) (mount spe
 		}
 		if stat, ok := st.Sys().(*syscall.Stat_t); ok {
 			if err := os.Chown(upperDir, int(stat.Uid), int(stat.Gid)); err != nil {
```
mtrmac: (Pre-existing: Should all of these modifications of `upperDir` happen if the caller specified a custom (and thus, maybe, pre-populated) `UpperDirOptionFragment`?)

pkg/overlay/overlay_linux.go
```diff
@@ -61,6 +63,28 @@ func MountWithOptions(contentDir, source, dest string, opts *Options) (mount spe
...
+				if !errors.Is(err, syscall.EINVAL) {
+					return mount, err
+				}
+				overflowed := false
+				overflowUIDText, uerr := os.ReadFile("/proc/sys/kernel/overflowuid")
+				overflowGIDText, gerr := os.ReadFile("/proc/sys/kernel/overflowgid")
+				if uerr == nil && gerr == nil {
+					overflowUID, uerr := strconv.Atoi(strings.TrimSpace(string(overflowUIDText)))
+					overflowGID, gerr := strconv.Atoi(strings.TrimSpace(string(overflowGIDText)))
```
mtrmac: (Absolutely non-blocking: `storage/pkg/idtools` already contains code to read this, only once per process. Aesthetically it would be nice to avoid the repetition — OTOH this has trivially small cost as is.)

imagebuildah/stage_executor.go
mtrmac: “With --from set, this isn’t from the build context directory, so we don’t want to…”, then ?

tests/bud.bats
```diff
@@ -9017,3 +9017,52 @@ EOF
...
+  FROM dir:/dir
+  RUN --mount=type=bind,from=0,target=/var/tmp/0 --mount=type=bind,from=1,target=/var/tmp/1 --mount=type=bind,from=2,target=/var/tmp/2 --mount=type=bind,from=3,target=/var/tmp/3 --mount=type=bind,from=4,target=/var/tmp/4 --mount=type=bind,from=5,target=/var/tmp/5 --mount=type=bind,from=6,target=/var/tmp/6 --mount=type=bind,from=7,target=/var/tmp/7 --mount=type=bind,from=8,target=/var/tmp/8 --mount=type=bind,from=9,target=/var/tmp/9 --mount=type=bind,from=10,target=/var/tmp/10 --mount=type=bind,from=11,target=/var/tmp/11 :
+  FROM dir:./dir/
+  RUN --mount=type=bind,from=0,target=/var/tmp/0 --mount=type=bind,from=1,target=/var/tmp/1 --mount=type=bind,from=2,target=/var/tmp/2 --mount=type=bind,from=3,target=/var/tmp/3 --mount=type=bind,from=4,target=/var/tmp/4 --mount=type=bind,from=5,target=/var/tmp/5 --mount=type=bind,from=6,target=/var/tmp/6 --mount=type=bind,from=7,target=/var/tmp/7 --mount=type=bind,from=8,target=/var/tmp/8 --mount=type=bind,from=9,target=/var/tmp/9 --mount=type=bind,from=10,target=/var/tmp/10 --mount=type=bind,from=11,target=/var/tmp/11 --mount=type=bind,from=12,target=/var/tmp/12 :
+EOF
```
mtrmac, line 9040: (Absolutely non-blocking: The 12 lines could maybe be generated… I really didn’t review them whether there is a typo in a number somewhere.)

internal/sanitize/sanitize.go
```diff
@@ -0,0 +1,338 @@
...
+		}
+		if _, validTarget := seenEntries[linkname]; !validTarget {
+			return fmt.Errorf("invalid symbolic link from %q to %q (%q) in archive", hdr.Name, hdr.Linkname, linkname)
+		}
+		rel, err := filepath.Rel(filepath.FromSlash(path.Dir("/"+hdr.Name)), filepath.FromSlash("/"+linkname))
+		if err != nil {
+			return fmt.Errorf("computing relative path from %q to %q in archive", hdr.Name, linkname)
+		}
+		rel = filepath.ToSlash(rel)
+		if convertSymlinksToHardlinks {
```
mtrmac, line 95: This does not work for links to directories — does that matter? c/image does not create them, but I don’t know that they are forbidden for OCI. (OCI doesn’t really specify a _tar_ format.)

internal/sanitize/sanitize.go
```diff
@@ -0,0 +1,338 @@
...
+				f.Close()
+				return fmt.Errorf("short write writing %q in base image: %d != %d", hdr.Name, n, hdr.Size)
+			}
+			return f.Close()
+		}()
+	case tar.TypeLink:
+		linkName := path.Clean("/" + hdr.Linkname)
+		oldName := filepath.Join(root, filepath.FromSlash(linkName))
+		err = os.Link(oldName, newName)
+	case tar.TypeSymlink: // convert it to a hard link
```
mtrmac: Links to directories, as well.

internal/sanitize/sanitize.go
mtrmac: AFAICS the `err` check happens only after the supposedly-faulty `hdr` was already consumed. Doesn’t that rather defeat the point? If (?) we don’t want to abort on `ErrInsecurePath`, maybe filter that one out specifically?

internal/sanitize/sanitize.go
```diff
@@ -0,0 +1,338 @@
...
+	}
+	return nil
+}
+
+// ImageName limits which image transports we'll accept.  For those it accepts
+// which refer to filesystem objects, where relative path names are evaluated
+// relative to "contextDir", it will create a copy of the original image, under
+// "tmpdir", which contains no symbolic links.  It it returns a parseable
+// reference to the image which should be used.
+func ImageName(transportName, restOfImageName, contextDir, tmpdir string) (newFrom string, err error) {
```
mtrmac, line 186: (Absolutely non-blocking: I’m not sure that structuring the "directory to directory" and "tar to tar" situations as a _single_ code path with options is the clearest way to do things — I have a _hunch_ that a straight-line `sanitizeTar` + `sanitizeDir` (+  `extractSanitizedTar` called by both?), driven by a `switch` that deals ~only with the c/image syntaxes, might be a simpler structure — but this works, it has tests, and I’m not a maintainer here, so this is in all respects fine.)

internal/sanitize/sanitize_test.go
```diff
@@ -0,0 +1,405 @@
...
+
+func TestSanitizeImageName(t *testing.T) {
+	const badLinkTarget = "../../../../../../../../../../../etc/passwd"
+
+	// prepare to copy from the layout to other types of destinations
+	ctx := context.Background()
+	// sys := &types.SystemContext{}
+	policy, err := signature.NewPolicyFromBytes([]byte(`{"default":[{"type":"insecureAcceptAnything"}]}`))
+	require.NoErrorf(t, err, "creating a policy")
+	policyContext, err := signature.NewPolicyContext(policy)
```
mtrmac, line 47: (`defer policyContext.Destroy()`)

internal/sanitize/sanitize_test.go
```diff
@@ -0,0 +1,405 @@
...
+	require.ErrorIs(t, err, os.ErrNotExist, "sanitizing bad OCI layout")
+
+	badDir := mutateDirectory(t, contextDir, goodDir, blobDigest.Encoded())
+	badDirRel, err := filepath.Rel(contextDir, badDir)
+	require.NoErrorf(t, err, "converting absolute path %q to a relative one", badDir)
+	_, err = ImageName("dir", badDirRel, contextDir, t.TempDir())
+	require.ErrorIs(t, err, os.ErrNotExist, "sanitizing bad directory")
+
+	badOCIArchive := mutateArchive(t, contextDir, goodOCIArchive, filepath.Join(v1.ImageBlobsDir, blobDigest.Algorithm().String(), blobDigest.Encoded()))
+
```
mtrmac: (Absolutely non-blocking: Drop this empty line for consistency? Alternatively, it might be possible to make parts of this test table-driven.)

imagebuildah/stage_executor.go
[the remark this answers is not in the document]
mtrmac: That sounds like something that could be disambiguated at the callers … but also not an immediate topic of this PR, just vaguely related as “a weird FROM value”.

imagebuildah/build.go
```diff
@@ -284,6 +284,14 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 		}
 
 		builds.Go(func() error {
+			contextDirectory, processLabel, mountLabel, usingContextOverlay, cleanupOverlay, err := platformSetupContextDirectoryOverlay(store, &options)
+			if err != nil {
+				return fmt.Errorf("mounting an overlay over build context directory: %w", err)
+			}
+			defer cleanupOverlay()
+			if contextDirectory != "" {
```
mtrmac: (`platformSetupContextDirectoryOverlay` does not document that it might return `""`, this might be made unconditional.)

pkg/overlay/overlay_linux.go
nalind: Hmm, probably not.  Making changes to upperDir conditional on it being one we created and will presumably destroy.

pkg/overlay/overlay_linux.go
nalind: Initializing a new `IDMappings` object to have it do this for us seems like it'd be more logic, so I'll leave this for now.

tests/bud.bats
nalind: Changing both of these.

internal/sanitize/sanitize.go
nalind: For this case, I think an absolute symlink would work.

internal/sanitize/sanitize.go
nalind: I don't _think_ that's likely to become a thing, with everything except a few exceptions being blob files.

internal/sanitize/sanitize.go
nalind: I'd only expected EOF errors from Next(); better to error out on any error, then.

internal/sanitize/sanitize_test.go
nalind: Oof, adding.

internal/sanitize/sanitize_test.go
nalind: Dropping it.

imagebuildah/build.go
nalind: Making it unconditional.

mtrmac (commented):
LGTM.

@nalind does this have an ordering dependency with #6126 (in either direction)?

nalind:
@mtrmac I don't think there's an ordering dependency between the two, but admittedly, it's been a while.

mtrmac:
/lgtm

then, this one has had more approving reviews.

TomSweeneyRedHat:
LGTM

jlebon:
Awesome to see this merge! Looking forward to be able to drop the `-v $PWD:/buildcontext` hack everywhere. Hmm, I think this also allows dropping `--security-opt=label=disable` in cases where it was added just for that.

Anyway, it looks like this was a lot of non-trivial work, but it's very much appreciated! (And hopefully in the future we can lean more into this and have nicer semantics for the pattern described in https://github.com/containers/buildah/issues/5952.)

PR #5978, opened 2025-02-07
Distinguish --mount=type=cache locations by ownership, too

nalind:
#### What type of PR is this?

/kind bug

Normally, we select and distinguish --mount=type=cache directories that we create by either the "id" or "target" value used when mounting them, but we should also be distinguishing them by the owner UID/GID.

#### How to verify it

New conformance test!

Addresses part of https://github.com/containers/podman/discussions/25260.

```release-note
The host directory used when a `buildah build` RUN instruction or `buildah run` uses the `--mount=type=cache` option is now chosen based not only on its "target" or "id" value, but also on the combination of "uid" and "gid" flag values specified.
```


flouthoc (approved):
LGTM

PR #5990, opened 2025-02-14
Fix Conformance tests on ARM64

Honny1:
#### What type of PR is this?

> /kind api-change
>
/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
>
/kind failing-test 
> /kind feature
> /kind flake
> /kind other

This PR fixes bugs discovered by running Conformance tests on ARM64. 

Fixed tests:
- `TestConformance/setuid-file-in-archive`
- `TestConformance/multistage-builtin-args`
- `TestConformance/builtins`

#### How to verify it

Execute Conformance tests on ARM system.

None

```release-note
None
```

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -430,6 +430,7 @@ func buildDockerfilesOnce(ctx context.Context, store storage.Store, logger *logr
 		builtinArgDefaults["TARGETVARIANT"] = defaultPlatform.Variant
 		builtinArgDefaults["TARGETARCH"] = defaultPlatform.Architecture
 		builtinArgDefaults["TARGETPLATFORM"] = defaultPlatform.OS + "/" + defaultPlatform.Architecture
+		builtinArgDefaults["TARGETPLATFORM"] = defaultPlatform.OS + "/" + defaultPlatform.Architecture
 		if defaultPlatform.Variant != "" {
 			builtinArgDefaults["TARGETPLATFORM"] += "/" + defaultPlatform.Variant
 		}
@@ -453,6 +454,7 @@ func buildDockerfilesOnce(ctx context.Context, store storage.Store, logger *logr
 	for k, v := range builtinArgDefaults {
 		b.BuiltinArgDefaults[k] = v
 	}
+
 	defaultContainerConfig, err := config.Default()
 	if err != nil {
 		return "", nil, fmt.Errorf("failed to get container config: %w", err)
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -290,7 +290,8 @@ _EOF
          "With explicit --platform, buildah should warn about pulling difference in platform"
   assert "$output" =~ "TARGETOS=linux" " --platform TARGETOS set correctly"
   assert "$output" =~ "TARGETARCH=amd64" " --platform TARGETARCH set correctly"
-  assert "$output" =~ "TARGETVARIANT=" " --platform TARGETVARIANT set correctly"
+  # By default, BUILDVARIANT/TARGETVARIANT should be empty.
+  assert "$output" =~ "TARGETVARIANT=[[:cntrl:]]" " --platform TARGETVARIANT set correctly"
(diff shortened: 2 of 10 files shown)


TomSweeneyRedHat:
LGTM

tests/bud.bats
```diff
@@ -290,7 +290,8 @@ _EOF
          "With explicit --platform, buildah should warn about pulling difference in platform"
   assert "$output" =~ "TARGETOS=linux" " --platform TARGETOS set correctly"
   assert "$output" =~ "TARGETARCH=amd64" " --platform TARGETARCH set correctly"
-  assert "$output" =~ "TARGETVARIANT=" " --platform TARGETVARIANT set correctly"
+  # By default, BUILDVARIANT/TARGETVARIANT should be empty.
+  assert "$output" =~ "TARGETVARIANT=\s" " --platform TARGETVARIANT set correctly"
```
nalind, line 294: Is the `\s` a perl-ism?  I'm surprised the shell is able to match it in the way this test expects.

tests/conformance/conformance_test.go
```diff
@@ -3311,7 +3321,7 @@ var internalTestCases = []testCase{
 	},
 
 	{
-		name:              "multistage-builtin-args",
+		name:              "multistage-builtin-args", /* By default, BUILDVARIANT/TARGETVARIANT should be empty. */
```
nalind, line 3324: Please use the `//` comment form, consistently.

imagebuildah/build.go
```diff
@@ -453,6 +450,13 @@ func buildDockerfilesOnce(ctx context.Context, store storage.Store, logger *logr
 	for k, v := range builtinArgDefaults {
 		b.BuiltinArgDefaults[k] = v
 	}
+
+	buildVariant := b.BuiltinArgDefaults["BUILDVARIANT"]
+	buildPlatform := b.BuiltinArgDefaults["BUILDPLATFORM"]
+	cleanedBuildPlatform, _ := strings.CutSuffix(buildPlatform, "/"+buildVariant)
+	b.BuiltinArgDefaults["BUILDPLATFORM"] = cleanedBuildPlatform
+	b.BuiltinArgDefaults["BUILDVARIANT"] = ""
```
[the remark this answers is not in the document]
Honny1: I don't know if any other user of `imagebuilder` would expect that "BUILD..." should be empty by default to match docker.

tests/bud.bats
Honny1: `=~` is a regex operator. So I'm trying to match the end of the line with `\s`.

imagebuildah/build.go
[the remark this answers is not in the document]
nalind: Why would they expect something different?

tests/bud.bats
nalind: I understand that, but `\s` isn't something I can find described in `regex(7)`.  It pops up in several man pages related to Perl, though.

tests/conformance/conformance_test.go
```diff
@@ -2019,6 +2019,16 @@ var internalTestCases = []testCase{
 		compatScratchConfig: types.OptionalBoolTrue,
 	},
 
+	{
+		name: "from-scratch-simple-file", /* Fix On ARM64: fields missing from buildah version: variant */
```
nalind, line 2023: Please use the // comment form, consistently.

tests/bud.bats
Honny1: Interesting, I never thought of that. Is that wrong? Should I use `[[:cntrl:]]` instead?

imagebuildah/build.go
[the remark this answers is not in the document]
Honny1: They might expect something different because the current behavior (setting BUILD..). Changing it to empty by default would potentially break something that rely on it.  But I will try to change it in `imagebuilder`.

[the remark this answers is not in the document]
Honny1: - https://github.com/openshift/imagebuilder/pull/301

PR #5992, opened 2025-02-18
chroot createPlatformContainer: use MS_REMOUNT

nalind:
#### What type of PR is this?

/kind bug

When setting mount propagation on the root mount before unmounting it, use `MS_REBIND`, since we know it's already a bind mount, and we actually want to affect the extant bind mount instead of creating another right over it. Otherwise, we might as well have not bothered.

#### How to verify it

New integration test!

Related to KONFLUX-5871.

```release-note
None
```


flouthoc (approved):
LGTM

flouthoc:
@nalind This needs a rebase.

nalind:
Rebased.

PR #5994, opened 2025-02-19
cirrus: reduce task timeout

Luap99:
With all the recent speed-ups here the timeout is way to high, all tasks should complete in under 30 mins generally. The smoke test in under 10min as it does not do much.

In particular I noticed at least two separate rootless integration tests time out after 120min[1,2], obviously the tests do not take that long and they are hanging somehwere instead. With a lower timeout we do not waste so much time when this happens.

[1] https://cirrus-ci.com/task/4733420225429504
[2] https://cirrus-ci.com/task/5597909967699968

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -50,7 +50,7 @@ env:
 gcp_credentials: ENCRYPTED[ae0bf7370f0b6e446bc61d0865a2c55d3e166b3fab9466eb0393e38e1c66a31ca4c71ddc7e0139d47d075c36dd6d3fd7]
 
 # Default timeout for each task
-timeout_in: 120m
+timeout_in: 30m
 
 # Default VM to use unless set or modified by task
 gce_instance: &standardvm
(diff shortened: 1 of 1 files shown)


Luap99:
@flouthoc @nalind PTAL

flouthoc (approved):
LGTM
/lgtm

flouthoc:
@Luap99 Also I am curious any hints why this `hangs` are happening ? I have seen them a few times when certain tests are just hanged.

flouthoc:
Often in these hangs the logs are also not printend.

Luap99:
@flouthoc yeah well that is from someone to figure out why things hang, of course if a bats test hangs no further logs will be printed. It is also possible but some of the recent log lines are still buffered so it is not like the log exactly tells you which line hangs.

Best bet try to reproduce outside of CI or log at the logs in CI, we see to capture the journald output so maybe you find clues there. Let me file an issue and we can debug there.

Luap99:
https://github.com/containers/buildah/issues/5996

Issue #5557, opened 2024-05-31
Add --parents option for COPY in Dockerfiles


uckelman:
**Description**

Docker added a `--parents` option to the Dockerfile `COPY` command last year:

https://github.com/moby/moby/issues/35639
https://github.com/moby/buildkit/pull/3001
https://docs.docker.com/reference/dockerfile/#copy---parents

buildah doesn't know how to handle `COPY --parents` so errors out on Dockerfiles which use it.

I'm seeing increasing use of `--parents` in Dockerfiles I have to deal with, which is making it hard to use podman/buildah instead of Docker. `--parents` is also a genuinely useful option on its own, so would be nice to have not just for compatibility.

It's quite useful to be able to do things like
```
COPY --parents src/**/pyproject.toml src/**/poetry.lock /app/
```
and maintain the directory structure in the destination without having to list out every single file.

TomSweeneyRedHat:
I believe @mheon is either working on this or looking into this.  Matt?

mheon:
Yep, hoping to get to this over the next few weeks

cyqsimon:
This is so badly needed with Java projects that involve multiple `pom.xml` files. Any progress on this?

krmao:
any more update?

jorlando-elucid:
bump this is a good idea

Peneheals:
any news? :)

rhatdan:
Could you give me a sample Containerfile that would use this?

cyqsimon:
@rhatdan Of course. This sort of thing is useful in a workspace of any language where nested manifest files are involved.

[Here](https://github.com/rust-lang/cargo/issues/2644#issuecomment-2335736020) is an example in Rust (they are working on toolchain support for building only dependencies without project source, but that's not our concern).

Alternatively, [here](https://github.com/jabbrwcky/sample-maven-structure) is an example with Maven with existing toolchain support. To optimise layer caching, you would want something like this:

```dockerfile
COPY --parents **/pom.xml ./
RUN mvn dependency:go-offline
```

Without `--parents`, you'd have to manually path to every sub-manifest.

oliversalzburg:
https://github.com/larixer/docker-js-monorepo has some pretty good examples as well.

kikorb:
Any news on this? it would be really helpful to have `--parents` supported

mheon:
@Honny1 might have time to work on this?

Honny1:
I will take a look.

PR #6008, opened 2025-02-25
Add --parents option for COPY in Containerfiles

Honny1:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
>
/kind feature
> /kind flake
> /kind other

This PR adds the `--parents` option for COPY in Containerfiles. It also includes an implementation of the `--parents` option for the `buildah copy` command. 

#### How to verify it

Try using `--parents` as described in the docker [documentation](https://docs.docker.com/reference/dockerfile/#copy---parents).

Fixes: https://issues.redhat.com/browse/RUN-2193
Fixes: #5557

```release-note
COPY option and buildah copy now support --parents options.
```


rhatdan:
@nalind PTAL

tests/conformance/conformance_test.go
```diff
@@ -2526,6 +2526,116 @@ var internalTestCases = []testCase{
...
+				./x/y/a.txt
+				./x/y/b.txt
+
+				Output:
+				/parents/y/a.txt
+				/parents/y/b.txt
+			*/
+		}, "\n"),
+		contextDir:          "copy-parents",
+		fsSkip:              []string{"(dir):parents:mtime", "(dir):parents:(dir):y:mtime"},
```
nalind, line 2612: Does `docker build` preserve the timestamp on `y`, or its permissions if they're something other than `0o755`?

copier/copier.go
```diff
@@ -343,22 +343,24 @@ func Stat(root string, directory string, options StatOptions, globs []string) ([
...
-	IgnoreUnreadable   bool              // ignore errors reading items, instead of returning an error
-	NoCrossDevice      bool              // if a subdirectory is a mountpoint with a different device number, include it but skip its contents
+	UIDMap, GIDMap        []idtools.IDMap   // map from hostIDs to containerIDs in the output archive
+	Excludes              []string          // contents to pretend don't exist, using the OS-specific path separator
+	ExpandArchives        bool              // extract the contents of named items that are archives
+	ChownDirs             *idtools.IDPair   // set ownership on directories. no effect on archives being extracted
+	ChmodDirs             *os.FileMode      // set permissions on directories. no effect on archives being extracted
+	ChownFiles            *idtools.IDPair   // set ownership of files. no effect on archives being extracted
+	ChmodFiles            *os.FileMode      // set permissions on files. no effect on archives being extracted
+	Parents               bool              // maintain the sources parent directory in the destination
```
nalind, line 353: Are the parent directories of the items being read supposed to be emitted in the tarstream when this flag is set?

tests/conformance/conformance_test.go
Honny1: I have made a rework and the current version preserves the timestamp of the copied directory.

copier/copier.go
Honny1: I have made a rework and the current version copies the parent directories to the tarstream.

add.go
```diff
@@ -94,6 +94,9 @@ type AddAndCopyOptions struct {
 	// RetryDelay is how long to wait before retrying attempts to retrieve
 	// remote contents.
 	RetryDelay time.Duration
+	// Parents preserve parent directories of source content
+	Parents         bool
+	ParentsPatterns map[string]string
```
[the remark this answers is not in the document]
Honny1: I tried to explain what the content should be in the comment. If `ParentsPatterns` is a `nil` value, it will be ignored. The `sources` list is cleaned up with `filepath.Join`, which removes `/./` from the path so it can't be used. Therefore, I decided to use a map to connect the source path and the path (pattern) with `/./`.

tests/conformance/conformance_test.go
Honny1: I don't think Docker preserves the `y` timestamp.

cmd/buildah/addcopy.go
```diff
@@ -247,6 +249,12 @@ func addAndCopyCmd(c *cobra.Command, args []string, verb string, iopts addCopyRe
 		InsecureSkipTLSVerify: systemContext.DockerInsecureSkipTLSVerify,
 		MaxRetries:            iopts.retry,
 	}
+	if iopts.parents {
+		options.ParentsPatterns = map[string]string{}
+		for _, pattern := range args {
+			options.ParentsPatterns[pattern] = pattern
```
nalind, line 255: This doesn't appear to be doing anything different when `contextdir` is set, while https://github.com/containers/buildah/pull/6008/files#diff-f2e4566c6b7e38384283187aba6d7fd91e5ba8da2ffd0f849277bb76bff27fb3R561-R562 does.  If it's required, I'd expect it to be required in both places, and it's something that the integration tests should be verifying works correctly.

copier/copier.go
```diff
@@ -1400,6 +1428,31 @@ func copierHandlerGet(bulkWriter io.Writer, req request, pm *fileutils.PatternMa
...
+		}
+		if err := copy(strings.TrimPrefix(parentName, "/"), parentPath, parentInfo); err != nil {
+			return alreadyCopied, err
+		}
+
+		alreadyCopied[parentPath] = struct{}{}
+		parentName = filepath.Dir(parentName)
+		parentPath = filepath.Dir(parentPath)
+	}
+	return alreadyCopied, nil
```
nalind, line 1453: This kind of implies that `alreadyCopied` wasn't already modified in the caller's scope, but this function was passed the map itself rather than a copy, so returning it here isn't really necessary.

copier/copier.go
```diff
@@ -1400,6 +1428,31 @@ func copierHandlerGet(bulkWriter io.Writer, req request, pm *fileutils.PatternMa
 	return &response{Stat: statResponse.Stat, Get: getResponse{}}, cb, nil
 }
 
+func copyParentsDirs(name string, path string, alreadyCopied map[string]struct{}, copy func(string, string, os.FileInfo) error) (map[string]struct{}, error) {
```
nalind, line 1431: Does this avoid outputting a directory that has already been output while walking the tree?

nalind (changes requested):
I think this really needs some unit tests in the `copier` package to ensure that it produces exactly the right set of outputs for a given set of inputs, with no parent directories being implicit or being output multiple times.

copier/copier.go
```diff
@@ -1400,6 +1428,31 @@ func copierHandlerGet(bulkWriter io.Writer, req request, pm *fileutils.PatternMa
...
+			return alreadyCopied, fmt.Errorf("copier: get: lstat %q: %w", parentPath, err)
+		}
+		if err := copy(strings.TrimPrefix(parentName, "/"), parentPath, parentInfo); err != nil {
+			return alreadyCopied, err
+		}
+
+		alreadyCopied[parentPath] = struct{}{}
+		parentName = filepath.Dir(parentName)
+		parentPath = filepath.Dir(parentPath)
+	}
```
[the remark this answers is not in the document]
Honny1: Oh, Yes, they are. That's probably wrong.

Honny1:
@nalind I removed the use of `ParentsPrefix/Patterns` and only use bool flags `Parents`. And I added `copier` tests.

copier/copier_test.go
```diff
@@ -1364,6 +1365,126 @@ func testGetMultiple(t *testing.T) {
...
+						"subdir-a/file-c",
+					},
+				},
+				{
+					name:    "root-wildcard-parents",
+					pattern: "/subdir-b/*",
+					parents: true,
+					items: []string{
+						"subdir-b",
+						"subdir-b",
```
nalind, line 1452: We shouldn't be outputting a parent directory more than once here.  It's good that this was covered by a test, though.

copier/copier_test.go
```diff
@@ -1364,6 +1365,126 @@ func testGetMultiple(t *testing.T) {
...
+						"subdir-b/file-o",
+					},
+				},
+				{
+					name:    "dotdot-wildcard-parents",
+					pattern: "../../subdir-b/*",
+					parents: true,
+					items: []string{
+						"subdir-b",
+						"subdir-b",
```
nalind, line 1463: We shouldn't be outputting a parent directory more than once here.  It's good that this was covered by a test, though.

nalind (changes requested):
Mostly nits, except that `Get()` really shouldn't be outputting a given item more than once.

Honny1:
@nalind I added tests for hardlinks and symlinks.

nalind (changes requested):
We have special logic for handling pivot points ("/./" in source paths), and for handling hard links, so both the conformance test and integration test should also include cases where they both come into play.

The tests don't automatically enforce that items that are hard linked to each other remain hard links when they're copied together, so the tests will need to check that directly by comparing the results of running something like `stat -c %i` against the items that are expected to be hard links.  We already have tests that ensure links show up in a committed image if they're in the working container's rootfs, so if it's easier, it would be sufficient to check on that during a RUN instruction right after the COPY instruction.

Honny1:
@nalind Oh, I forgot about that. Thank you. I found the mistake. It should be fine now.

nalind:
LGTM

copier/copier.go
```diff
@@ -1230,31 +1287,53 @@ func copierHandlerGet(bulkWriter io.Writer, req request, pm *fileutils.PatternMa
...
+				if _, ok := addedParents[parentName]; ok {
+					continue
 				}
-				followedLinks++
+				addedParents[parentName] = struct{}{}
+
+				if err := copierHandlerGetOne(parentInfo, "", parentName, parent, req.GetOptions, tw, hardlinkChecker, idMappings); err != nil {
+					if req.GetOptions.IgnoreUnreadable && errorIsPermission(err) {
+						continue
+					} else if errors.Is(err, os.ErrNotExist) {
```
rhatdan, line 1323: Nit: No need for else here.

copier/copier.go
```diff
@@ -1230,31 +1287,53 @@ func copierHandlerGet(bulkWriter io.Writer, req request, pm *fileutils.PatternMa
...
+						logrus.Warningf("copier: file disappeared while reading: %q", parent)
+						return nil
+					}
+					return fmt.Errorf("copier: get: %q: %w", queue[i].glob, err)
```
rhatdan, line 1327: Stutter?

rhatdan:
Great job @Honny1.  A couple of nits, and potential stutters but I will merge
/lgtm

Honny1:
@rhatdan Thanks you. Should I fix nits?

PR #6014, opened 2025-02-28
Use UnparsedInstance.Manifest instead of ImageSource.GetManifest

mtrmac:
... to validate that the manifests match expected digests, if any.

In some cases, using an UnparsedInstance can also avoid redundant I/O.

Do this everywhere, even where we read local storage which is mostly trusted, because it is cheap enough and being consistent makes it less likely for the code to be copied into other contexts shere the sources are not trusted.

#### What type of PR is this?

> /kind cleanup

```release-note
none
```

### DIFF
--- a/import.go
+++ b/import.go
@@ -39,7 +39,8 @@ func importBuilderDataFromImage(ctx context.Context, store storage.Store, system
 	defer src.Close()
 
 	imageDigest := ""
-	manifestBytes, manifestType, err := src.GetManifest(ctx, nil)
+	unparsedTop := image.UnparsedInstance(src, nil)
+	manifestBytes, manifestType, err := unparsedTop.Manifest(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("loading image manifest for %q: %w", transports.ImageName(ref), err)
 	}
@@ -48,6 +49,7 @@ func importBuilderDataFromImage(ctx context.Context, store storage.Store, system
 	}
 
 	var instanceDigest *digest.Digest
+	unparsedInstance := unparsedTop // for instanceDigest
 	if manifest.MIMETypeIsMultiImage(manifestType) {
 		list, err := manifest.ListFromBlob(manifestBytes, manifestType)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


rhatdan:
LGTM

import.go
```diff
@@ -48,6 +49,7 @@ func importBuilderDataFromImage(ctx context.Context, store storage.Store, system
 	}
 
 	var instanceDigest *digest.Digest
```
nalind, line 51: I think this variable becomes unused in this scope.  It can be implicitly declared where it's still used, or removed entirely.

new.go
```diff
@@ -195,14 +195,16 @@ func newBuilder(ctx context.Context, store storage.Store, options BuilderOptions
...
-		manifestBytes, manifestType, err := srcSrc.GetManifest(ctx, nil)
+		unparsedTop := image.UnparsedInstance(srcSrc, nil)
+		manifestBytes, manifestType, err := unparsedTop.Manifest(ctx)
 		if err != nil {
 			return nil, fmt.Errorf("loading image manifest for %q: %w", transports.ImageName(ref), err)
 		}
 		if manifestDigest, err := manifest.Digest(manifestBytes); err == nil {
 			imageDigest = manifestDigest.String()
 		}
 		var instanceDigest *digest.Digest
```
nalind, line 206: I think this variable becomes unused in this scope.  It can be implicitly declared where it's still used, or removed entirely.

nalind (commented):
The godoc for the `ImageSource` type mentions that validation of things is left up to the caller, but it's not clear to me that switching to using the `Unparsed...` family of types and functions was how we were intended to do it.  Is it that `Image.Manifest()` verifies the manifest's digest while caching it, but `ImageSource.GetManifest()` doesn't?

The updates for consistency's sake being rolled into the same commit make it harder for me to distinguish the places where the previous logic was clearly a problem, from the rest.

mtrmac:
> The godoc for the `ImageSource` type mentions that validation of things is left up to the caller, but it's not clear to me that switching to using the `Unparsed...` family of types and functions was how we were intended to do it.

It’s … how `copy.Image` does that. But this is a great point, the documentation (even with https://github.com/containers/image/pull/2742 ) is not sufficient. I’ll fix it in that PR.

---

> The updates for consistency's sake being rolled into the same commit make it harder for me to distinguish the places where the previous logic was clearly a problem, from the rest.

Yes… this PR proposes converting everything, because that was easy to do. Analyzing which uses are actually risky would be more work than just locally changing the call sites, and I didn’t do that (yet?).

import.go
mtrmac: `instanceDigest` is used for reporting errors from `image.FromUnparsedImage`, it is used in both the single-platform and multi-platform case (but only in case of an error).

(The idea is that `instanceDigest` and `unparsedInstance` should match, i.e. `unparsedInstance == image.UnparsedInstance(srcSrc, instanceDigest)`.)

new.go
mtrmac: This is the same situation as in the other case.

import.go
nalind: Ah, I missed it being used in adding context to the error message.

TomSweeneyRedHat:
LGTM
But I'd like a final head nod from @nalind

PR #6041, opened 2025-03-07
stage_executor: history should include heredoc summary correctly

flouthoc:
`getCreatedBy` ignores heredoc summary when build args are specified following PR makes sure the behaviour is correct.

Also test is modified to make sure buildah correctly burst cache if heredoc content is changed.

Closes: https://github.com/containers/podman/issues/25469

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
stage_executor: history should include heredoc summary correctly
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -1943,17 +1943,20 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
 		if len(node.Original) > 4 {
 			shArg = node.Original[4:]
 		}
-		if buildArgs != "" {
-			return "|" + strconv.Itoa(len(strings.Split(buildArgs, " "))) + " " + buildArgs + " /bin/sh -c " + shArg + appendCheckSum, nil
-		}
-		result := "/bin/sh -c " + shArg
+
+		heredoc := ""
+		result := ""
 		if len(node.Heredocs) > 0 {
 			for _, doc := range node.Heredocs {
 				heredocContent := strings.TrimSpace(doc.Content)
-				result = result + "\n" + heredocContent
+				heredoc = heredoc + "\n" + heredocContent
 			}
 		}
-		return result + appendCheckSum, nil
+		if buildArgs != "" {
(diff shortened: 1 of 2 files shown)


imagebuildah/stage_executor.go
```diff
@@ -1943,16 +1943,18 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
...
+		heredoc := ""
 		if len(node.Heredocs) > 0 {
 			for _, doc := range node.Heredocs {
 				heredocContent := strings.TrimSpace(doc.Content)
-				result = result + "\n" + heredocContent
+				heredoc = heredoc + "\n" + heredocContent
 			}
 		}
+		if buildArgs != "" {
+			return "|" + strconv.Itoa(len(strings.Split(buildArgs, " "))) + " " + buildArgs + " /bin/sh -c " + shArg + heredoc + appendCheckSum, nil
```
nalind, line 1955: Correct result, but harder to follow than having this block compute a value that also gets combined with others and is returned from one place.  It would be very easy to mistakenly introduce a similar bug in the future.

imagebuildah/stage_executor.go
```diff
@@ -1943,16 +1943,18 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
...
+		}
+		result := "/bin/sh -c " + shArg + heredoc
 		return result + appendCheckSum, nil
```
nalind, line 1958: Why concatenate multiple values to produce one string, then concatenate another with it in the `return`?

tests/bud.bats
```diff
@@ -531,6 +532,7 @@ _EOF
 
   cat > $contextdir/Dockerfile << _EOF
 FROM alpine
+ARG key=value
```
nalind, line 535: Do we still have cache test coverage for cases where there isn't an `ARG` before the heredoc?

PR #6046, opened 2025-03-09
feat: add --ssl-cert-file option to build command

aeijdenberg:
This makes the root CAs from the host available to containers during a build.

#### What type of PR is this?

/kind feature

This makes it easier to use `buildah` to build an image in environments that require custom certificate authorities to make outbound internet connections (for example when using a TLS intercepting HTTPS proxy).

Currently `buildah` will automatically pass `https_proxy` (and related) environment variables as ephemeral environment variables to the container for `RUN` commands.

This change adds an option `--with-ssl-cert-file` (defaults to `false`) that when set will:

1. Resolve the root CA file currently in use by the host (uses `SSL_CERT_FILE` if set, else falls back to searching common paths).
2. Mount this file within the container during the `RUN` command at a different location (`/host-ssl-cert-file`).
3. Sets `SSL_CERT_FILE=/host-ssl-cert-file`  ephemeral environment variable for `RUN` commands. 

A similar outcome can be achieved today by editing `buildah` command to bind mount the current CA store and then edit each `RUN` instruction in the `Containerfile` to use, however it felt like it would be nice to make a short-cut for this which seems to augment well the existing `https_proxy` variable propagation.

Current workaround:

```bash
./bin/buildah build --secret=id=ca,src=/etc/ssl/certs/ca-certificates.crt -f - <<'EOF'
FROM alpine:latest
ENV SSL_CERT_FILE=/host-ca-file
RUN --mount=type=secret,id=ca,target=/host-ca-file echo "SSL_CERT_FILE: ${SSL_CERT_FILE} - Contents:" && head -n10 "${SSL_CERT_FILE}"
EOF
```

#### How to verify it

Behaviour without flag:

```bash
./bin/buildah build -f - <<'EOF'
FROM alpine:latest
RUN echo "SSL_CERT_FILE: ${SSL_CERT_FILE} - Contents:" && head -n10 "${SSL_CERT_FILE}"
EOF
```

Output: 

```
STEP 1/2: FROM alpine:latest
STEP 2/2: RUN echo "SSL_CERT_FILE: ${SSL_CERT_FILE} - Contents:" && head -n10 "${SSL_CERT_FILE}"
SSL_CERT_FILE:  - Contents:
head: : No such file or directory
Error: building at STEP "RUN echo "SSL_CERT_FILE: ${SSL_CERT_FILE} - Contents:" && head -n10 "${SSL_CERT_FILE}"": while running runtime: exit status 1
```

Behaviour with flag:

```bash
./bin/buildah build --ssl-cert-file -f - <<'EOF'
FROM alpine:latest
RUN echo "SSL_CERT_FILE: ${SSL_CERT_FILE} - Contents:" && head -n10 "${SSL_CERT_FILE}"
EOF
```

Output:

```
STEP 1/2: FROM alpine:latest
STEP 2/2: RUN echo "SSL_CERT_FILE: ${SSL_CERT_FILE} - Contents:" && head -n10 "${SSL_CERT_FILE}"
SSL_CERT_FILE: /host-ssl-cert-file - Contents:
-----BEGIN CERTIFICATE-----
MIIH0zCCBbugAwIBAgIIXsO3pkN/pOAwDQYJKoZIhvcNAQEFBQAwQjESMBAGA1UE
(lots more output)
```

Tested with a local HTTPS proxy server with a self-signed CA and verified that basic `apk add --update --no-cache python3` and similar commands worked without error.

(Tested similar on `ubuntu` image, although `apt` I think runs over HTTP so test wasn't as meaningful)

None

I'm happy to add tests but wanted to get feedback on the direction first. It's quite a small PR, but I think quite helpful for those stuck behind corporate proxies. The automatic adding of `https_proxy` environment variables is a good start, and this aims to give the same experience for CAs.

Some issues to maybe fix first:

##### Search path for host CAs

To get the search path of commonly used host CA files I used the list found here:
https://cs.opensource.google/go/go/+/refs/tags/go1.24.1:src/crypto/x509/root_linux.go;l=9-17

Unfortunately that variable is private, and I'm not able to find anything within the standard x509 module that exports it or otherwise makes it available, which is why I put that list in a separate file and copy/pasted the Go LICENSE within there.

It feels less than ideal and open to other ideas.

##### Path to bind to within the container

This PR currently binds it to `/host-ssl-cert-file` so that it doesn't clash with commonly used paths within a container. This seemed to work fine in my limited testing when using `SSL_CERT_FILE` to point to it, however it's not clear whether `SSL_CERT_FILE` will work for all applications and whether it may be best to instead bind this on top of well-known locations for where this file normally is.

Happy to take feedback on alternate suggestions.

```release-note
Adds --ssl-cert-file option to build command. If set this propogates the host CAs into the container during the build process.
```

### DIFF
--- a/run_common.go
+++ b/run_common.go
@@ -326,6 +326,10 @@ func (b *Builder) configureEnvironment(g *generate.Generator, options RunOptions
 		}
 	}
 
+	if b.CommonBuildOpts.SSLCertFile {
+		g.AddProcessEnv("SSL_CERT_FILE", "/host-ssl-cert-file")
+	}
+
 	for _, envSpec := range util.MergeEnv(util.MergeEnv(defaultEnv, b.Env()), options.Env) {
 		env := strings.SplitN(envSpec, "=", 2)
 		if len(env) > 1 {
--- a/util/x509.go
+++ b/util/x509.go
@@ -1,0 +1,23 @@
+package util
+
+import (
+	"errors"
+	"fmt"
+	"os"
+)
+
+func ResolveRootCACertFile() (string, error) {
+	if rv, ok := os.LookupEnv("SSL_CERT_FILE"); ok {
+		return rv, nil
+	}
+	for _, potFile := range certFiles {
+		if _, err := os.Stat(potFile); err != nil {
+			if os.IsNotExist(err) {
+				continue
+			}
+			return "", fmt.Errorf("unexpected error resolving cert file: %w", err)
+		}
+		return potFile, nil
+	}
+	return "", errors.New("please set SSL_CERT_FILE to use --ssl-cert-file option")
+}
--- a/util/x509_unix.go
+++ b/util/x509_unix.go
@@ -1,0 +1,45 @@
+//go:build !windows
+
+package util
+
+// Source: https://cs.opensource.google/go/go/+/refs/tags/go1.24.1:src/crypto/x509/root_linux.go
+
+/*
+	Copyright 2009 The Go Authors.
+
+	Redistribution and use in source and binary forms, with or without
+	modification, are permitted provided that the following conditions are
+	met:
+
+	* Redistributions of source code must retain the above copyright
+	notice, this list of conditions and the following disclaimer.
+	* Redistributions in binary form must reproduce the above
+	copyright notice, this list of conditions and the following disclaimer
+	in the documentation and/or other materials provided with the
+	distribution.
+	* Neither the name of Google LLC nor the names of its
+	contributors may be used to endorse or promote products derived from
+	this software without specific prior written permission.
+
+	THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+	"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+	LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+	A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+	OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+	SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+	LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+	DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+	THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+	(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+	OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+// Possible certificate files; stop after finding one.
+var certFiles = []string{
+	"/etc/ssl/certs/ca-certificates.crt",                // Debian/Ubuntu/Gentoo etc.
+	"/etc/pki/tls/certs/ca-bundle.crt",                  // Fedora/RHEL 6
+	"/etc/ssl/ca-bundle.pem",                            // OpenSUSE
+	"/etc/pki/tls/cacert.pem",                           // OpenELEC
+	"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", // CentOS/RHEL 7
+	"/etc/ssl/cert.pem",                                 // Alpine Linux
+}
--- a/define/build.go
+++ b/define/build.go
@@ -83,6 +83,8 @@ type CommonBuildOptions struct {
 	ApparmorProfile string
 	// ShmSize is the "size" value to use when mounting an shmfs on the container's /dev/shm directory.
 	ShmSize string
+	// SSLCertFile specifies whether the host root CAs should be ephemerally mounted within the container
+	SSLCertFile bool
 	// Ulimit specifies resource limit options, in the form type:softlimit[:hardlimit].
 	// These types are recognized:
 	// "core": maximum core dump size (ulimit -c)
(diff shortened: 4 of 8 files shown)


docs/buildah-build.1.md
```diff
@@ -1232,6 +1232,11 @@ will convert /foo into a `shared` mount point.  The propagation properties of th
 mount can be changed directly. For instance if `/` is the source mount for
 `/foo`, then use `mount --make-shared /` to convert `/` into a `shared` mount.
 
+**--with-ssl-cert-file**
+
+If set, bind mount the host CA cert file (override location by setting `SSL_CERT_FILE`) within the RUN containers,
+and set environment variable `SSL_CERT_FILE` to point to it.
```
TomSweeneyRedHat, line 1238:
```suggestion
and set the environment variable `SSL_CERT_FILE` to point to it.
```

docs/buildah-from.1.md
```diff
@@ -690,6 +690,11 @@ will convert /foo into a `shared` mount point.  The propagation properties of th
 mount can be changed directly. For instance if `/` is the source mount for
 `/foo`, then use `mount --make-shared /` to convert `/` into a `shared` mount.
 
+**--with-ssl-cert-file**
+
+If set, bind mount the host CA cert file (override location by setting `SSL_CERT_FILE`) within the RUN containers,
+and set environment variable `SSL_CERT_FILE` to point to it.
```
TomSweeneyRedHat, line 696:
```suggestion
and set the environment variable `SSL_CERT_FILE` to point to it.
```

util/x509_unix.go
```diff
@@ -0,0 +1,45 @@
...
+
+// Possible certificate files; stop after finding one.
+var certFiles = []string{
+	"/etc/ssl/certs/ca-certificates.crt",                // Debian/Ubuntu/Gentoo etc.
+	"/etc/pki/tls/certs/ca-bundle.crt",                  // Fedora/RHEL 6
+	"/etc/ssl/ca-bundle.pem",                            // OpenSUSE
+	"/etc/pki/tls/cacert.pem",                           // OpenELEC
+	"/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem", // CentOS/RHEL 7
+	"/etc/ssl/cert.pem",                                 // Alpine Linux
+}
```
TomSweeneyRedHat, line 45: Is x509_unix.go an appropriate name?  With the first line, I thought maybe x509_windows.go?  Similar comment next file.

TomSweeneyRedHat:
Generally looks good to me, a couple of nits and it looks like your branch needs updating.

util/x509_unix.go
aeijdenberg: I think so? I was copying the style used for other files in this repo such as https://github.com/containers/buildah/commit/30c2e31d8e49ed725da14b0d5658c5e5b09e32f0

aeijdenberg:
Accepted changes, rebased, squashed and fixed the issue in the Fedora tests, by using the same pattern as for similar files (such as /etc/hostname). Thanks for the review.

TomSweeneyRedHat:
@nalind or @Luap99 PTAL

run_common.go
```diff
@@ -208,6 +208,39 @@ func (b *Builder) generateHostname(rdir, hostname string, chownOpts *idtools.IDP
...
+
+	hostCertBytes, err := os.ReadFile(resolvedSSLCertPath)
+	if err != nil {
+		return "", fmt.Errorf("error reading cert file on host: %w", err)
+	}
+
+	cfile := filepath.Join(rdir, filepath.Base(containerPath))
+	if err = ioutils.AtomicWriteFile(cfile, hostCertBytes, 0o644); err != nil {
+		return "", fmt.Errorf("writing %s into the container: %w", containerPath, err)
+	}
```
Luap99, line 226: I don't know how big the certs files can be but generally it seems best to copy it without reading the full thing into memory.

util/x509.go
```diff
@@ -0,0 +1,23 @@
...
+	for _, potFile := range certFiles {
+		if _, err := os.Stat(potFile); err != nil {
+			if os.IsNotExist(err) {
+				continue
+			}
+			return "", fmt.Errorf("unexpected error resolving cert file: %w", err)
+		}
+		return potFile, nil
+	}
+	return "", errors.New("please set SSL_CERT_FILE to use --ssl-cert-file option")
```
Luap99, line 22: Nothing in this function relates to the cli option. Buildah might be used via the API as well so the cli options should generally not be used in the backend as these errors would make no sense to callers.

Luap99 (commented):
I don't have a strong opinion on this one but this seems like a slippery slope.

I can certainly see why this argument might be handy but allowing this might open the door for a lot of other similar things. Personally I think this is better done by callers to set up the env and mount in any way they like. Hard coding `/host-ssl-cert-file` just feels super awkward, and no it is not about the specific path name but just in general.

I would like to know what @nalind thinks about this as I am not really a buildah maintainer.

util/x509.go
aeijdenberg: Thanks that makes sense. Have reworded the message.

aeijdenberg:
> If it wasn't set to a hard-coded path, and rather something like `/tmp-3727282` instead (different random on each run) would that remove part of the objection?

(I was making some other changes to address other feedback, and made a change like the above at the same time and pushed that onto this branch for discussion)

util/x509_test.go
```diff
@@ -0,0 +1,16 @@
+package util
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestResolveRootCACertFileWitEnvSet(t *testing.T) {
```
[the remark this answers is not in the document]
aeijdenberg: Accepted suggestion and added to rebase just pushed.

aeijdenberg:
For background on what I'm trying to achieve, I'd like to be able to make it easier to build images (without modifying existing Dockerfiles) in situations where they need to go through TLS intercepting HTTP proxies.

While I thought this might work well and be convenient, I think I can achieve the same using other more generic buildah features if they land.

1. https://github.com/containers/buildah/pull/6285 which allows effectively ephemeral env variables to be "mounted" during a `RUN` instruction
2. (PR to be written) an option to `build` which modifies all `RUN` instructions to add a set of mounts (similar to what the documentation for `TransientMounts` says it does)

I think it might be best to abandon/close this PR for now, and persue those others.

aeijdenberg:
#6289 opened for the second part as decribed above.

PR #6047, opened 2025-03-10
fix: dockerfiles... always evaluates to an empty list

aeijdenberg:
Even when dockerfiles is appended to later in this method, it has no entries at time of evaluation as an argument to the defer function, thus no files were ever closed as a result.

#### What type of PR is this?

/kind bug

Spotted this when trying to debug why some resources where kept busy. Although this wasn't the root cause, the code as written was not doing what it looked like it intended to do, so was an easy fix.

#### How to verify it

Test added.

None

I had a longer PR prepped that would move the body of the loop over Dockerfiles into a separate function, so that each file could be closed immediately after use, but figured this shorter PR would be preferred.

I did a quick search of the code base other instance of using either `...` or `[]` on the same line as `defer` and couldn't find any, so I think this is the only instance of this type of issue in this codebase.

```release-note
None
```

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -93,12 +93,7 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 	}
 	logger.SetLevel(logrus.GetLevel())
 
-	var dockerfiles []io.ReadCloser
-	defer func(dockerfiles ...io.ReadCloser) {
-		for _, d := range dockerfiles {
-			d.Close()
-		}
-	}(dockerfiles...)
+	var dockerfiles []io.Reader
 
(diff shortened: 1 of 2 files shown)


aeijdenberg:
Interestingly this is the 2nd version of the bug on the same function - when added 7 years ago it looked like this:
https://github.com/containers/buildah/blame/a16c3cdd7dbd7588966d9e5eeabdcf8258b0e705/imagebuildah/build.go#L705-L709

```golang
	defer func(dockerfiles ...io.ReadCloser) {
		for _, d := range dockerfiles {
			d.Close()
		}
	}()
```

which also was always called with an empy slice.

I've added tests, which helped surface a 2nd bug.

nalind:
Nice catch!
/approve
/lgtm

PR #6074, opened 2025-03-20
Mask thermal paths

### LINKED ISSUE #6073: Mask Linux thermal interrupt info in /proc and /sys
### Issue Description

we need the equivalent of https://github.com/moby/moby/pull/49560 for Moby.

The paths `/proc/interrupts` and `/sys/devices/system/cpu/<cpuX>/thermal_throttle`  must be masked by default.

The main difficulty is that `/sys/devices/system/cpu/<cpuX>/thermal_throttle` is not a static value but it depends on the number of cpus on the current system, so we need to make sure it doesn't break the current unmasking mechanism we have in place.

giuseppe:
/kind bug

Mask /proc/interrupts and /sys/devices/system/cpu/<cpuX>/thermal_throttle by default

Fixes #6073 

```release-note
 /proc/interrupts and /sys/devices/system/cpu/*/thermal_throttle are masked by default
```


giuseppe:
@containers/buildah-maintainers ready for review

lsm5 (approved):
Do we need to add / update tests? 

LGTM as such.

giuseppe:
> Do we need to add / update tests?

the tests were added to the c/common change

PR #6076, opened 2025-03-21
Fix built-in args on ARM64

Honny1:
#### What type of PR is this?

> /kind api-change
/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

This PR fixes built-in args on ARM64.

#### How to verify it
Conformance tests should pass on ARM64.

None

```release-note
None
```


Luap99:
Note ARM64 CI is much more complex, all arm tasks should be on AWS.

First this repo is not setup for AWS testing at all.
Second our image generation on AWS only produces a fedora image.

These are two big things to change if we want to test that and I am not sure how much effort that is worth here.
We have been talking about migrating all our testing to AWS but I am not sure when that will happen.

Honny1:
@Luap99  Thank you for the explanation. I'll omit the CI part. Sounds like a big effort. I think we can add an ARM-based task when we migrate CI to CNCF.

PR #6083, opened 2025-03-25
Allow extendedGlob to work with Windows paths

danegsta:
#### What type of PR is this?

/kind bug

This PR fixes processing of Windows style paths in extendedGlob. The first issue fixed is an infinite loop when presented with a windows style path (i.e. `C:\Some\Path\**.txt`). The second issue is ensuring that the pattern includes the drive letter on Windows. The fix primarily depends on the behavior of `filepath.VolumeName`. On Windows it will return the drive letter for a path, while on other systems it returns an empty string.

Additionally, I limited xattrs_test.go to the same go:build targets as xattrs.go, as that was preventing me from running tests properly on Windows.

I found this issue while looking to make `podman container cp` work properly on Windows. I'm also working on a PR for the podman repo to fix some additional issues, but would need this scenario fixed before I can proceed to open that PR, otherwise the command enters an infinite loop and never completes (currently it fails due to a separate path parsing bug in the `cp` command implementation).

#### How to verify it

Run the `TestExtendedGlob` test case on Windows (i.e. `go test ./copier/... -run TestExtendedGlob`). There are still test failures for other tests in the copier package on Windows, but the test coverage for this method now passes.

None

```release-note
None
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -70,12 +70,14 @@ func extendedGlob(pattern string) (matches []string, err error) {
 		components := []string{}
 		dir := pattern
 		file := ""
-		for dir != "" && dir != string(os.PathSeparator) {
+		for dir != filepath.VolumeName(dir) && dir != string(os.PathSeparator) {
 			dir, file = filepath.Split(dir)
-			components = append([]string{file}, components...)
+			if file != "" {
+				components = append([]string{file}, components...)
+			}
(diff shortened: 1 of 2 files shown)


Honny1:
LGTM

PTAL @nalind

flouthoc (approved):
LGTM

rhatdan:
LGTM

rhatdan:
/approve
/lgtm
Thanks @danegsta

PR #6087, opened 2025-03-25
add: correctly resolve git `tags` and `branches`

flouthoc:
Attempt to check if `tag` or `branch` exists on remote repo or not, if it does not exists then correctly report `error`.

Closes: https://github.com/containers/podman/issues/25679

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
add: correctly resolve git tags and branches and report errors if no reference found
```


add.go
```diff
@@ -517,7 +517,7 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 			if sourceIsGit(src) {
 				go func() {
 					var cloneDir, subdir string
-					cloneDir, subdir, getErr = define.TempDirForURL(tmpdir.GetTempDir(), "", src)
+					cloneDir, subdir, dirErr = define.TempDirForURL(tmpdir.GetTempDir(), "", src)
```
nalind, line 520: This should just `return` if `getErr` is not `nil`, as there's no point in continuing, and `cloneDir` and `subdir` are not expected to be useful values.

define/types.go
```diff
@@ -260,13 +260,6 @@ func parseGitBuildContext(url string) (string, string, string) {
 	return gitBranchPart[0], gitSubdir, gitBranch
 }
 
-func isGitTag(remote, ref string) bool {
```
flouthoc, line 263: There is no need to for this because `fetch` below automatically fails if `reference` does not exists.

define/types.go
```diff
@@ -284,12 +277,6 @@ func cloneToDirectory(url, dir string) ([]byte, string, error) {
 		return combinedOutput, gitSubdir, fmt.Errorf("failed while performing `git remote add`: %w", err)
 	}
 
-	if gitRef != "" {
```
flouthoc, line 287: There is no need for this `fetch` automatically resolves correct `tag`, `branch` or `commit`. If not found it fails.

add.go
```diff
@@ -518,6 +518,11 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 				go func() {
 					var cloneDir, subdir string
 					cloneDir, subdir, getErr = define.TempDirForURL(tmpdir.GetTempDir(), "", src)
+					defer pipeWriter.Close()
+					defer wg.Done()
```
nalind, line 522: Swap these two lines - we want the waitgroup's `Wait()` to wait until after the `pipeWriter` is closed, as before.

tests/bud.bats
```diff
@@ -7313,6 +7313,15 @@ _EOF
   local_head_hash=$output
   run_buildah run testctr -- sh -c 'cd podman-tag && git ls-remote --tags origin v5.0.0^{} | cut -f1'
   assert "$output" = "$local_head_hash"
+
+  cat > $contextdir/Dockerfile << _EOF
+FROM alpine
+RUN apk add git
```
nalind, line 7319: The image shouldn't need `git` installed for `ADD` to work, so this step is unnecessary.  It can be `FROM scratch` if it doesn't need to run anything after the content has been added.

nalind (changes requested):

nalind (approved):
Changes look good, but the commit log doesn't exactly reflect what was changed.

flouthoc:

add.go
```diff
@@ -518,6 +518,11 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 				go func() {
 					var cloneDir, subdir string
 					cloneDir, subdir, getErr = define.TempDirForURL(tmpdir.GetTempDir(), "", src)
+					defer wg.Done()
+					defer pipeWriter.Close()
```
Luap99, line 522: The defer placement feels wrong, doing this after the function call but before errors checking is odd.
IMO they should be at the top of the function.

tests/bud.bats
```diff
@@ -7313,6 +7313,13 @@ _EOF
   local_head_hash=$output
   run_buildah run testctr -- sh -c 'cd podman-tag && git ls-remote --tags origin v5.0.0^{} | cut -f1'
   assert "$output" = "$local_head_hash"
+
+  cat > $contextdir/Dockerfile << _EOF
+FROM scratch
+ADD https://github.com/openshift/microshift.git#foobar /src
```
Luap99, line 7319: We have no direct control over this repo, if they ever add such a branch all our CI would break.
Can we use buildah instead?

Honny1 (commented):
Other than @nalind's comment, LGTM

tests/bud.bats
```diff
@@ -7313,6 +7313,13 @@ _EOF
   local_head_hash=$output
   run_buildah run testctr -- sh -c 'cd podman-tag && git ls-remote --tags origin v5.0.0^{} | cut -f1'
   assert "$output" = "$local_head_hash"
+
+  cat > $contextdir/Dockerfile << _EOF
+FROM scratch
+ADD https://github.com/containers/crun.git#nosuchbranch /src
+_EOF
+  run_buildah 128 build -f $contextdir/Dockerfile -t git-image $contextdir
+  expect_output --substring "couldn't find remote ref foobar"
```
Luap99, line 7322: you switched to nosuchbranch but the test matches still foobar

flouthoc: my bad

Luap99 (approved):
/lgtm

But I don't have merge powers here I belive

PR #6088, opened 2025-03-25
createPlatformContainer: drop MS_REMOUNT|MS_BIND

nalind:
#### What type of PR is this?

/kind bug

When attempting to change the mount propagation of the old root directory tree before unmounting it, it's enough that we pass the requested propagation flags.

In particular, MS_REC is the only flag that is supposed to be allowed to be specified along with a mount propagation flag, but in practice it was only triggering an error some of the time, and CI wasn't one of those times.

#### How to verify it

The added integration test mounts the root filesystem as an overlay and then runs buildah as a rootless user on top of that, which is more comparable to a root-on-composefs configuration, which triggered the error.

Fixes errors detected in [RHEL-79560](https://issues.redhat.com/browse/RHEL-79560).

```release-note
"chroot" isolation should no longer require the "--no-pivot" flag be specified in order to function on "image mode" systems.
```

### DIFF
--- a/chroot/run_linux.go
+++ b/chroot/run_linux.go
@@ -263,7 +263,7 @@ func createPlatformContainer(options runUsingChrootExecSubprocOptions) error {
 		return fmt.Errorf("changing to host root directory: %w", err)
 	}
 	// make sure we only unmount things under this tree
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@flouthoc PTAL

TomSweeneyRedHat:
LGTM

PR #6092, opened 2025-03-27
build: return `ExecErrorCodeGeneric` when git operation fails instead of relaying error code directly from git

flouthoc:
Only propagate error message from git and let buildah reflect error code
`125`.

Reason: Buildah should return predicatable error code from the set of
defined error codes in exec_codes.go at https://github.com/containers/buildah/blob/main/pkg/cli/exec_codes.go#L6
anything other that predefined error codes introduces inconsistency thus making testing difficult in CI and podman.

Users should expect buildah to refect ExecErrorCodeGeneric with error message kept intact from the underlying `git`
commands.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
build: return 125 when git operation fails instead of relaying error code directly from git
```

### DIFF
--- a/define/types.go
+++ b/define/types.go
@@ -267,14 +267,18 @@ func cloneToDirectory(url, dir string) ([]byte, string, error) {
 	cmd = exec.Command("git", "init", dir)
 	combinedOutput, err := cmd.CombinedOutput()
 	if err != nil {
-		return combinedOutput, gitSubdir, fmt.Errorf("failed while performing `git init`: %w", err)
+		// Return err.Error() instead of err as we want buildah to override error code with more predictable
+		// value.
+		return combinedOutput, gitSubdir, fmt.Errorf("failed while performing `git init`: %s", err.Error())
 	}
 	// add origin
 	cmd = exec.Command("git", "remote", "add", "origin", gitRepo)
 	cmd.Dir = dir
 	combinedOutput, err = cmd.CombinedOutput()
 	if err != nil {
-		return combinedOutput, gitSubdir, fmt.Errorf("failed while performing `git remote add`: %w", err)
+		// Return err.Error() instead of err as we want buildah to override error code with more predictable
+		// value.
+		return combinedOutput, gitSubdir, fmt.Errorf("failed while performing `git remote add`: %s", err.Error())
 	}
 
 	logrus.Debugf("fetching repo %q and branch (or commit ID) %q to %q", gitRepo, gitRef, dir)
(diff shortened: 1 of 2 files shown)


nalind:
Please fill out the description, describing why this is being changed, and describing the user-facing change in terms of what someone running the binary would observe when running it.

flouthoc:
@nalind Updated commit and PR body.

nalind:
My second request was about

> #### Does this PR introduce a user-facing change?
> 
> ```
> cloneToDirectory: don't relay error code from git
> ```

I wouldn't expect someone who wasn't familiar with the implementation to be able to make sense of this.

flouthoc:
> My second request was about
> 
> > #### Does this PR introduce a user-facing change?
> > ```
> > cloneToDirectory: don't relay error code from git
> > ```
> 
> I wouldn't expect someone who wasn't familiar with the implementation to be able to make sense of this.

@nalind How about this

```
build: return ExecErrorCodeGeneric when git operation fails instead of relaying error code directly from git
```

define/types.go
```diff
@@ -267,14 +267,14 @@ func cloneToDirectory(url, dir string) ([]byte, string, error) {
 	cmd = exec.Command("git", "init", dir)
 	combinedOutput, err := cmd.CombinedOutput()
 	if err != nil {
-		return combinedOutput, gitSubdir, fmt.Errorf("failed while performing `git init`: %w", err)
+		return combinedOutput, gitSubdir, fmt.Errorf("failed while performing `git init`: %s", err.Error())
```
Luap99, line 270: It would likely makes sense to add a comment here (once at least for all four cases) why this is done as this seems like an anti pattern.

flouthoc: Done.

flouthoc:

nalind:
> > I wouldn't expect someone who wasn't familiar with the implementation to be able to make sense of this.
> 
> @nalind How about this
> 
> ```
> build: return ExecErrorCodeGeneric when git operation fails instead of relaying error code directly from git
> ```

I wouldn't expect someone who wasn't familiar with the implementation to know what `ExecErrorCodeGeneric` means, either.  Use the number.  If we ever want to be able to pull release note information from merged PRs, the writing and the format are going to matter.

flouthoc:
> > > I wouldn't expect someone who wasn't familiar with the implementation to be able to make sense of this.
> > 
> > 
> > @nalind How about this
> > ```
> > build: return ExecErrorCodeGeneric when git operation fails instead of relaying error code directly from git
> > ```
> 
> I wouldn't expect someone who wasn't familiar with the implementation to know what `ExecErrorCodeGeneric` means, either. Use the number. If we ever want to be able to pull release note information from merged PRs, the writing and the format are going to matter.

Done edited the release note.

nalind:
LGTM

Luap99 (approved):
/lgtm

flouthoc:
`podman-remote` reads the body of the error message and since body containers `exit status 128` it sets the error code to `128` causing `podman-remote` test to fail.

Issue #6098, opened 2025-03-31
The behaviour of --unsetlabel might be wrong


mkosiarc:
Hi,

We are interested in using the **--unsetlabel** functionality to remove labels inherited from base image. But the way it currently behaves does not make it suitable for the purpose. It might be a bug, as it does not remove just the inherited labels from base image but all labels. This means the labels that are specified in the current Dockerfile/Containerfile and also those that are being set with --label parameter.

To provide an example. Let's say I have Dockerfile such as:
```
FROM registry.access.redhat.com/ubi8/ubi:latest

LABEL name="my-image"
LABEL release="2025"

RUN dnf install -y git
```

The ubi8 base image also contains those labels. So when during build I do something like:

```
buildah build --unsetlabel name --unsetlabel release -t my-test-image .
```

The resulting image does not have the name and release labels, even though I feel like the labels specified in Dockerfile should be kept and not removed as the help suggests:

```
--unsetlabel strings    unset label when inheriting labels from base image
```

The --unsetlabel also takes precedence when doing:

```
buildah build --unsetlabel name --label name=my-label -t my-test-image .
```

Meaning the resulting image does not have the label name. All this means that there is no simple way to remove just the labels from the base image. If I wanted to do that, I have to parse all labels from base image, parse all labels in dockerfile, and then make sure to do --unsetlabel only on the difference of those two sets. This is not very practical for the purpose this functionality was meant to be used as originally requested in https://github.com/containers/buildah/issues/5063

I tried this on the following versions:

- 1.37.1
- 1.39.2
- 1.40.0-dev

Could you please a have a look at it and fix it if possible? And/or perhaps could https://github.com/containers/buildah/pull/5061 be reopened?

arewm:
This behavior looks like it was intended in the PR based on a comment: https://github.com/containers/buildah/pull/5062#discussion_r1337560741.

With the current functionality, users (or build tools) will need to list all of the labels that are being added to the image so that they can be removed from the `--unsetlabel` argument. While this may be possible in some situations, it would be preferable if we could just define which labels we _don't_ want to inherit and then let the image build define any that it wants as normal.

ralphbean:
I'd prefer an option to disinherit _all_ labels from the base image without having to list them and let the child set only the new labels it chooses through its containerfile (or explicit additions of --label).

arewm:
[...]

PR #6103, opened 2025-04-01
build: add support for `--inherit-labels`

flouthoc:
Allows users to specify if they want to inherit labels from base image or not.

Closes: https://github.com/containers/buildah/issues/6098

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
build: allow users to choose if they want to inherit labels from base image or not using --inherit-labels
```

### DIFF
--- a/define/build.go
+++ b/define/build.go
@@ -236,6 +236,9 @@ type BuildOptions struct {
 	// ID mapping options to use if we're setting up our own user namespace
 	// when handling RUN instructions.
 	IDMappingOptions *IDMappingOptions
+	// InheritLabels controls whether or not built images will retain the labels
+	// which were set in their base images
+	InheritLabels types.OptionalBool
 	// AddCapabilities is a list of capabilities to add to the default set when
 	// handling RUN instructions.
 	AddCapabilities []string
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -497,6 +497,10 @@ Path to an alternative .containerignore (.dockerignore) file.
 Write the built image's ID to the file.  When `--platform` is specified more
 than once, attempting to use this option will trigger an error.
 
+**--inherit-labels** *bool-value*
+
+Inherit the labels from the base image or base stages. (default true).
+
 **--ipc** *how*
 
 Sets the configuration for IPC namespaces when handling `RUN` instructions.
--- a/pkg/cli/common.go
+++ b/pkg/cli/common.go
@@ -72,6 +72,7 @@ type BudResults struct {
 	Format              string
 	From                string
 	Iidfile             string
+	InheritLabels       bool
 	Label               []string
 	LayerLabel          []string
 	Logfile             string
@@ -231,6 +232,7 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
 	fs.StringVar(&flags.CertDir, "cert-dir", "", "use certificates at the specified path to access the registry")
 	fs.BoolVar(&flags.Compress, "compress", false, "this is a legacy option, which has no effect on the image")
 	fs.BoolVar(&flags.CompatVolumes, "compat-volumes", false, "preserve the contents of VOLUMEs during RUN instructions")
+	fs.BoolVar(&flags.InheritLabels, "inherit-labels", true, "inherit the labels from the base image or base stages.")
 	fs.StringArrayVar(&flags.CPPFlags, "cpp-flag", []string{}, "set additional flag to pass to C preprocessor (cpp)")
 	fs.StringVar(&flags.Creds, "creds", "", "use `[username[:password]]` for accessing the registry")
 	fs.StringVarP(&flags.CWOptions, "cw", "", "", "confidential workload `options`")
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -2670,6 +2670,80 @@ _EOF
   expect_output "$want_output"
 }
 
+@test "bud and test inherit-labels" {
+  base=registry.fedoraproject.org/fedora-minimal
+  _prefetch $base
+  _prefetch alpine
+  run_buildah --version
+  local -a output_fields=($output)
+  buildah_version=${output_fields[2]}
+  run_buildah build $WITH_POLICY_JSON -t exp -f $BUDFILES/base-with-labels/Containerfile
+
+  run_buildah inspect --format '{{ index .Docker.Config.Labels "license"}}' exp
+  expect_output "MIT" "license must be MIT from fedora base image"
+  run_buildah inspect --format '{{ index .Docker.Config.Labels "name"}}' exp
+  expect_output "fedora-minimal" "name must be fedora from base image"
+
+  run_buildah build $WITH_POLICY_JSON --inherit-labels=false --label name=world -t exp -f $BUDFILES/base-with-labels/Containerfile
+  # no labels should be inherited from base image, only the buildah version label
+  # and `hello=world` which we just added using cli flag
+  want_output='map["io.buildah.version":"'$buildah_version'" "name":"world"]'
+  run_buildah inspect --format '{{printf "%q" .Docker.Config.Labels}}' exp
+  expect_output "$want_output"
+
+  # Try building another file with multiple layers
+  run_buildah build $WITH_POLICY_JSON --iidfile ${TEST_SCRATCH_DIR}/id1 --layers -t exp -f $BUDFILES/base-with-labels/Containerfile.layer
+  run_buildah inspect --format '{{ index .Docker.Config.Labels "license"}}' exp
+  expect_output "MIT" "license must be MIT from fedora base image"
+  run_buildah inspect --format '{{ index .Docker.Config.Labels "name"}}' exp
+  expect_output "world" "name must be world from Containerfile"
+
+  # Now build same file with  --inherit-labels=false and verify if we are not using the cache again.
+  run_buildah build $WITH_POLICY_JSON --layers --inherit-labels=false --iidfile ${TEST_SCRATCH_DIR}/inherit_false_1 -t exp -f $BUDFILES/base-with-labels/Containerfile.layer
+  # Should not contain `Using cache` at all since
+  assert "$output" !~ "Using cache"
+  want_output='map["io.buildah.version":"'$buildah_version'" "name":"world"]'
+  run_buildah inspect --format '{{printf "%q" .Docker.Config.Labels}}' exp
+  expect_output "$want_output"
+
+  run_buildah build $WITH_POLICY_JSON --layers --inherit-labels=false --iidfile ${TEST_SCRATCH_DIR}/inherit_false_2 -t exp -f $BUDFILES/base-with-labels/Containerfile.layer
+  # Should contain `Using cache`
+  expect_output --substring " Using cache"
+  want_output='map["io.buildah.version":"'$buildah_version'" "name":"world"]'
+  run_buildah inspect --format '{{printf "%q" .Docker.Config.Labels}}' exp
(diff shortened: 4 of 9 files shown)


define/build.go
```diff
@@ -236,6 +236,9 @@ type BuildOptions struct {
 	// ID mapping options to use if we're setting up our own user namespace
 	// when handling RUN instructions.
 	IDMappingOptions *IDMappingOptions
+	// InheritLabels allows users to specify if they want
+	// to inheritlabels from base image or not.
```
nalind, line 240:
```suggestion
	// to inherit labels from base image or not.
```

docs/buildah-build.1.md
```diff
@@ -497,6 +497,10 @@ Path to an alternative .containerignore (.dockerignore) file.
 Write the built image's ID to the file.  When `--platform` is specified more
 than once, attempting to use this option will trigger an error.
 
+**--inherit-labels** *bool-value*
+
+Allows users to specify if they want to inherit labels from base image or not. (Default is `true`).
```
nalind, line 502: "Allows users" is kind of implied here, in the man page.  Describe what the tool will do when the flag is specified.

tests/bud.bats
```diff
@@ -2670,6 +2670,25 @@ _EOF
...
+  expect_output "MIT" "license must be MIT from fedora base image"
+  run_buildah inspect --format '{{ index .Docker.Config.Labels "name"}}' exp
+  expect_output "fedora-minimal" "name must be fedora from base image"
+
+  run_buildah build $WITH_POLICY_JSON --inherit-labels=false --label hello=world -t exp -f $BUDFILES/base-with-labels/Containerfile
+  # no labels should be inherited from base image only the, buildah version label
+  # and `hello=world` which we just added using cli flag
+  want_output='map["hello":"world" "io.buildah.version":"'$buildah_version'"]'
+  run_buildah inspect --format '{{printf "%q" .Docker.Config.Labels}}' exp
+  expect_output "$want_output"
```
nalind, line 2689: This needs to test builds with multiple layers (`--layers=true`) and ensure that the build cache doesn't use images that have been built with one value for this setting when the build is running with a different value, since the resulting configurations would differ.
It should probably also check running a multi-stage build where an earlier stage that's used as the base of a later stage sets a label, as I suspect that might not behave as expected in this iteration.

nalind (changes requested):
I think this needs to include broader tests, as there are cases I'm not sure are being handled here.

docs/buildah-build.1.md
TomSweeneyRedHat: Perhaps

Inherit the labels from the base image (default true).

I'm assuming the default is true; I haven't read the rest.

pkg/cli/common.go
```diff
@@ -230,6 +231,7 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
 	fs.StringVar(&flags.CertDir, "cert-dir", "", "use certificates at the specified path to access the registry")
 	fs.BoolVar(&flags.Compress, "compress", false, "this is a legacy option, which has no effect on the image")
 	fs.BoolVar(&flags.CompatVolumes, "compat-volumes", false, "preserve the contents of VOLUMEs during RUN instructions")
+	fs.BoolVar(&flags.InheritLabels, "inherit-labels", true, "inherit labels from base image")
```
TomSweeneyRedHat, line 234:
```suggestion
	fs.BoolVar(&flags.InheritLabels, "inherit-labels", true, "inherit the labels from the base image")
```

tests/bud.bats
TomSweeneyRedHat: It would also be good to add a test just using `--inherit-labels" to ensure the default works and doesn't get change in the future.  I'm also a bit on the OCD side, I'd also like a `--inherit-labels=true` test, but not completely necessary.

flouthoc: I added tests to work with `--layers`

define/build.go
```diff
@@ -236,6 +236,9 @@ type BuildOptions struct {
 	// ID mapping options to use if we're setting up our own user namespace
 	// when handling RUN instructions.
 	IDMappingOptions *IDMappingOptions
+	// InheritLabels allows users to specify if they want
```
nalind, line 239: Same comment as for the man page: we're at the API level here, so just describe what the library will be doing differently based on this field, perhaps something like "InheritLabels controls whether or not built images will retain the labels which were set in their base images".

tests/bud.bats
```diff
@@ -2670,6 +2670,49 @@ _EOF
...
+  run_buildah inspect --format '{{ index .Docker.Config.Labels "license"}}' exp
+  expect_output "MIT" "license must be MIT from fedora base image"
+  run_buildah inspect --format '{{ index .Docker.Config.Labels "name"}}' exp
+  expect_output "world" "name must be world from Containerfile"
+
+  # Now build same file with  --inherit-labels=false and verify if we are not using the cache again.
+  run_buildah build $WITH_POLICY_JSON --layers --inherit-labels=false -t exp -f $BUDFILES/base-with-labels/Containerfile.layer
+  # Should not contain `Using cache` at all since
+  assert "$output" !~ "Using cache"
+  want_output='map["io.buildah.version":"'$buildah_version'" "name":"world"]'
```
nalind, line 2702: Is the order of the map elements here guaranteed?

tests/bud.bats
```diff
@@ -2670,6 +2670,49 @@ _EOF
...
+  run_buildah inspect --format '{{printf "%q" .Docker.Config.Labels}}' exp
+  expect_output "$want_output"
+
+  # Now build same file with  --inherit-labels=true and verify if using the cache
+  run_buildah build $WITH_POLICY_JSON --layers --inherit-labels=true -t exp -f $BUDFILES/base-with-labels/Containerfile.layer
+  # Should contain `Using cache` at all since
+  expect_output --substring " Using cache"
```
nalind, line 2709: How many times is "Using cache" expected to be printed?  Would it be simpler to compare the IDs of the final images?

nalind (changes requested):
Tests should also check running a multi-stage build where an earlier stage that's used as the base of a later stage sets a label, as I suspect that might not behave as expected in this iteration, depending on what the expected behavior is, which is unclear.

tests/bud.bats
flouthoc: I also added a check below to verify image id.

flouthoc:
> Tests should also check running a multi-stage build where an earlier stage that's used as the base of a later stage sets a label, as I suspect that might not behave as expected in this iteration, depending on what the expected behavior is, which is unclear.

With this current patch it if `--inherit-labels=false` is set it does not inherit label into `target` stage from base stages. I think that aligns with the docs that no labels are inherited at all from base image. WDYT ?

nalind (changes requested):
> With this current patch it if `--inherit-labels=false` is set it does not inherit label into `target` stage from base stages. I think that aligns with the docs that no labels are inherited at all from base image. I added a test to verify that. WDYT ?

You could ask in the issue that prompted this PR, and given how #5762 describes handling of ARGs, which don't even get recorded in images, I'm not sure if I know what the correct behavior should be for labels when the base image for a stage is part of the same build.  Either way, the detail should be mentioned in the docs.

tests/bud.bats
```diff
@@ -2670,6 +2670,65 @@ _EOF
...
+
+  # Now build same file with  --inherit-labels=true and verify if using the cache
+  run_buildah build $WITH_POLICY_JSON --iidfile ${TEST_SCRATCH_DIR}/id2 --layers --inherit-labels=true -t exp -f $BUDFILES/base-with-labels/Containerfile.layer
+  expect_output --substring " Using cache"
+  run_buildah inspect --format '{{ index .Docker.Config.Labels "license"}}' exp
+  expect_output "MIT" "license must be MIT from fedora base image"
+  run_buildah inspect --format '{{ index .Docker.Config.Labels "name"}}' exp
+  expect_output "world" "name must be world from Containerfile"
+  # Final image id should be exactly same as the one image which was built in the past.
+  if [[ $(cat ${TEST_SCRATCH_DIR}/id1) != $(cat ${TEST_SCRATCH_DIR}/id2) ]]; then
```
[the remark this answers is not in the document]

tests/bud.bats
```diff
@@ -2670,6 +2670,69 @@ _EOF
...
+  buildah_version=${output_fields[2]}
+  run_buildah build $WITH_POLICY_JSON -t exp -f $BUDFILES/base-with-labels/Containerfile
+
+  run_buildah inspect --format '{{ index .Docker.Config.Labels "license"}}' exp
+  expect_output "MIT" "license must be MIT from fedora base image"
+  run_buildah inspect --format '{{ index .Docker.Config.Labels "name"}}' exp
+  expect_output "fedora-minimal" "name must be fedora from base image"
+
+  run_buildah build $WITH_POLICY_JSON --inherit-labels=false --label name=world -t exp -f $BUDFILES/base-with-labels/Containerfile
+  # no labels should be inherited from base image only the, buildah version label
```
nalind, line 2685:
```suggestion
  # no labels should be inherited from base image, only the buildah version label
```

tests/bud.bats
```diff
@@ -2670,6 +2670,69 @@ _EOF
...
+  want_output='map["io.buildah.version":"'$buildah_version'" "name":"world"]'
+  run_buildah inspect --format '{{printf "%q" .Docker.Config.Labels}}' exp
+  expect_output "$want_output"
+
+  # Rebuild again with layers should not build image again at all.
+  run_buildah build $WITH_POLICY_JSON --iidfile ${TEST_SCRATCH_DIR}/id4 --layers --inherit-labels=true -t exp -f $BUDFILES/base-with-labels/Containerfile.multi-stage
+  want_output='map["io.buildah.version":"'$buildah_version'" "name":"world"]'
+  run_buildah inspect --format '{{printf "%q" .Docker.Config.Labels}}' exp
+  expect_output "$want_output"
+  assert "$(cat ${TEST_SCRATCH_DIR}/id3)" = "$(cat ${TEST_SCRATCH_DIR}/id3)" "expected image ids to not change"
```
nalind, line 2733:
```suggestion
  assert "$(cat ${TEST_SCRATCH_DIR}/id3)" = "$(cat ${TEST_SCRATCH_DIR}/id4)" "expected image ids to not change"
```

tests/bud.bats
```diff
@@ -2670,6 +2670,69 @@ _EOF
   expect_output "$want_output"
 }
 
+@test "bud and test inherit-labels" {
```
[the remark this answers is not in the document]
nalind: `tests/bud/base-with-labels/Containerfile.multi-stage` uses "alpine" as its base.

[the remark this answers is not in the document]
flouthoc: I added `_prefetch alpine` too.

nalind (approved):
LGTM

flouthoc:

rhatdan:
Did we ever have a conversation on why not just do `--unsetenv=*` or `--unsetenv=all` rather then add a new option?

--cap-drop=all has precedence.

flouthoc:
@rhatdan I think you mean `--unsetlabel=all` or `--unsetlabel=*`. The reason is similar to what I have written here https://github.com/containers/buildah/issues/6098#issuecomment-2766411213 , we wanted an option which can only work with `labels` from base image//base stage but still allow users to add new labels from `target` stage or CLI. It seems like `--unsetlabel=all` cannot do this and removes all the labels.

PR #6110, opened 2025-04-04
stage_executor,getCreatedBy: expand buildArgs before invoking `generatePathChecksum`

flouthoc:
Patch in PR https://github.com/containers/buildah/pull/5691 added a function to calculate and write checksum to history of `source` in `--mount` instructions but it did not add part to expand build args if they are present in `source` path.

Following PR Just corrects the above issue and also adds a new test to make sure we don't break this in future again.

Closes: https://github.com/containers/podman/issues/25425

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -1896,6 +1896,10 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
 					continue
 				}
 				mountOptionSource = mountInfo.Source
+				mountOptionSource, err = imagebuilder.ProcessWord(mountOptionSource, s.stage.Builder.Arguments())
+				if err != nil {
+					return "", fmt.Errorf("getCreatedBy: while replacing arg variables with values for format %q: %w", mountOptionSource, err)
+				}
 				mountOptionFrom = mountInfo.From
 				// If source is not specified then default is '.'
 				if mountOptionSource == "" {
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -6757,6 +6757,16 @@ _EOF
   expect_output --substring "hello"
 }
 
+@test "bud-with-mount-relative-path-like-buildkit-arg-in-source" {
+  skip_if_no_runtime
+  skip_if_in_container
+  _prefetch alpine
+  local contextdir=${TEST_SCRATCH_DIR}/buildkit-mount
+  cp -R $BUDFILES/buildkit-mount $contextdir
+  run_buildah build --build-arg INPUTPATH_1=subdir -t testbud $WITH_POLICY_JSON -f $contextdir/Containerfile5 $contextdir/
+  expect_output --substring "hello"
+}
+
 @test "bud-with-mount-with-rw-like-buildkit" {
   skip_if_no_runtime
   skip_if_in_container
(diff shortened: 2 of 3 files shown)


imagebuildah/stage_executor.go
```diff
@@ -1903,6 +1903,10 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
 					continue
 				}
 				mountOptionSource = mountInfo.Source
+				mountOptionSource, err = imagebuilder.ProcessWord(mountOptionSource, s.stage.Builder.Arguments())
+				if err != nil {
+					return "", fmt.Errorf("getCreatedBy: while replacing arg variables with values for format %q: %w", mountOptionSource, err)
+				}
```
nalind, line 1909: Is there anything preventing other arguments to this flag, or other flags to this instruction, from also referencing build args?  If not, does the right thing happen when they do?

flouthoc: Variables already expand correctly for other flags, issue happend to `source=` because I decided to use `source` directly from node without expanding it in this PR https://github.com/containers/buildah/pull/5691


I can open a PR which can add a test to verify various flags of `--mount` but wondering if that should a different PR or this one ? Since in this PR I only tested the argument which was passed on to `generatePathChecksum`

nalind: If that's already covered by other test cases, there's no need for that.

nalind (approved):
LGTM

flouthoc:
@rhatdan @TomSweeneyRedHat @Honny1 PTAL

Luap99 (commented):
I don't understand enough about the buildah codebase but does that mean you are calling ProcessWord() twice on the that input. Supposedly it worked before the cache logic and now you add one new call for the cache. To me that seems to suggest we do that in two places which then means we might get a certain drift in behaviours?

Anyhow I don't understand enough about buildah so I am happy to shut up and say LGTM as it seems to fix the bug.

imagebuildah/stage_executor.go
```diff
@@ -1903,6 +1903,10 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
 					continue
 				}
 				mountOptionSource = mountInfo.Source
+				mountOptionSource, err = imagebuilder.ProcessWord(mountOptionSource, s.stage.Builder.Arguments())
```
TomSweeneyRedHat, line 1906: Is line 1995 still necessary?

TomSweeneyRedHat:
One question, an updte needed, otherwise LGTM

imagebuildah/stage_executor.go
flouthoc: Yes since this only processes `source` before passing it to `generatePathChecksum` at other places it is still needed.

TomSweeneyRedHat: Ooops, I meant linet 1898

mountOptionSource = mountInfo.Source

I need updated glasses, and it looks like you figured that out.  SGTM.

PR #6118, opened 2025-04-08
Modernize

kolyshkin:
#### What type of PR is this?

/kind cleanup

Converts the code to use various Go 1.23+ features:
 * `any` instead of `interface{}` (available since Go 1.18);
 * `for range` over integers (available since Go 1.22, https://tip.golang.org/ref/spec#For_range)
 * use various `slices` and `maps` methods more, where appropriate (Go 1.21...1.23)
 * a couple of minor fixes/cleanups
 
Please see individual commit for details.
 
Mostly found by [modernize](https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize).

#### How to verify it

Existing tests should be adequate.

None

```release-note
NONE
```


internal/sbom/presets.go
```diff
@@ -1,5 +1,7 @@
 package sbom
 
+import "slices"
+
 import "github.com/containers/buildah/define"
```
nalind, line 5: This looks... weird.

nalind (approved):
LGTM, except for a second "import" being added to a source file.

internal/sbom/presets.go
kolyshkin: Interestingly, even `gofumpt` doesn't care.

Fixed, thanks for noticing.

Luap99 (approved):
/lgtm

PR #5261, opened 2024-01-10
stage_executor,layers: burst cache of `layers` if heredoc content is changed

flouthoc:
When using buildah with `--layers` then buildah must correctly burst layer cache if `heredoc` content is changed. Following is achieved via properly adding `heredoc` content to the history of the built image.

Closes: https://github.com/containers/buildah/issues/5225

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### What this PR does / why we need it:
Closes: https://github.com/containers/buildah/issues/5225

#### How to verify it
Newly added integration test

#### Which issue(s) this PR fixes:
Closes: https://github.com/containers/buildah/issues/5225

#### Does this PR introduce a user-facing change?

```release-note
stage_executor,layers: burst cache of layers if heredoc content is changed
```

PR #6123, opened 2025-04-13
build, heredoc: show heredoc summary in build output

BoleynSu:
Fixes cases missed by https://github.com/containers/buildah/pull/5261

#### What type of PR is this?

/kind bug

```release-note
build, heredoc: show heredoc summary in build output
```


nalind:
Which cases are being missed?  As near as I can tell, this switches to always calculating the bits to append to the log message to include a mention of heredoc contents, even in cases when the log message isn't going to be output, when we didn't bother for those cases before.

PR #6126, opened 2025-04-15
RUN: only use an overlay for --mount=type=bind,rw on overlay

### LINKED ISSUE #5988: ERRO[0025] unlinkat /var/tmp/buildah2410054376/mounts3022885724/bind626918239: device or resource busy
When building inside a rootless container using buildah's `vfs` storage driver and `chroot` isolation (As is very often done to build images in CI environments), specifying read/write bind volumes from other stages results in an error.  This behavior does not reproduce using buildah 1.37 or earlier.  Also verified this same behavior using a vanilla `registry.fedoraproject.org/fedora-minimal` images + `dnf5 install buildah`.  That is to say, I think it's a buildah problem, not a buildah image problem.

Reproduction (host) environment:

- Fedora 40
- podman 5.3.1
- Running as a regular user w/ default podman settings
- The `quay.io/buildah/upstream:latest` container image (`buildah version 1.40.0-dev (image-spec 1.1.0, runtime-spec 1.2.0)`)
- The `quay.io/buildah/stable:v1.38` container image
- The `quay.io/buildah/stable:v1.37` container image

Steps to reproduce:

1. Create the following `Containerfile` somewhere in the users homedir
   ```
   FROM registry.fedoraproject.org/fedora-minimal:latest as test
   RUN mkdir -p /var/tmp/test
   ADD ./Containerfile /var/tmp/test/

   FROM test as final
   RUN --mount=type=bind,from=test,src=/var/tmp/test,dst=/var/tmp/test,rw \
       set -x && \
       date > /var/tmp/test/Containerfile && \
       cat /var/tmp/test/Containerfile
   ```
2. Run `podman run -it --rm -v ./Containerfile:/root/Containerfile:ro,Z quay.io/buildah/stable:v1.38 buildah --storage-driver=vfs build --isolation=chroot /root`
3. Run the exact same command, but with `quay.io/buildah/stable:v1.37` (or any other earlier version)

Unexpected results:

```
[1/2] STEP 1/3: FROM registry.fedoraproject.org/fedora-minimal:latest AS test
Trying to pull registry.fedoraproject.org/fedora-minimal:latest...
Copying blob 169491f3e4f7 done   |
Copying config e6917e6306 done   |
[1/2] STEP 2/3: RUN mkdir -p /var/tmp/test
[1/2] STEP 3/3: ADD ./Containerfile /var/tmp/test/
Copying blob cde90dcf8c1f skipped: already exists
Copying blob cec21250b843 done   |
Copying config 9f9e432f21 done   |
--> 9f9e432f21cb
[2/2] STEP 1/2: FROM 9f9e432f21cbb67c928b93d87af3878f3b903cbc2030cc12594f9368829ccc8c AS final
[2/2] STEP 2/2: RUN --mount=type=bind,from=test,src=/var/tmp/test,dst=/var/tmp/test,rw     set -x &&     date > /var/tmp/test/Containerfile &&     cat /var/tmp/test/Containerfile
ERRO[0025] unlinkat /var/tmp/buildah1274147250/mounts4133407440/bind3931917386: device or resource busy
Error: building at STEP "RUN --mount=type=bind,from=test,src=/var/tmp/test,dst=/var/tmp/test,rw set -x &&     date > /var/tmp/test/Containerfile &&     cat /var/tmp/test/Containerfile": resolving mountpoints for container "bb08d8062b4c17b75108492838e53d3236abce647447c8f5bec72cebfcb8ca1b": setting up overlay of "/var/tmp/buildah1274147250/mounts4133407440/bind3931917386": mount overlay:/var/tmp/buildah1274147250/mounts4133407440/overlay/981784139/merge, data: lowerdir=/var/tmp/buildah1274147250/mounts4133407440/bind3931917386,upperdir=/var/tmp/buildah1274147250/mounts4133407440/overlay/981784139/upper,workdir=/var/tmp/buildah1274147250/mounts4133407440/overlay/981784139/work,userxattr: invalid argument
```

Expected results (`from v1.37`):

```
[1/2] STEP 1/3: FROM registry.fedoraproject.org/fedora-minimal:latest AS test
Trying to pull registry.fedoraproject.org/fedora-minimal:latest...
Copying blob 169491f3e4f7 done   |
Copying config e6917e6306 done   |
[1/2] STEP 2/3: RUN mkdir -p /var/tmp/test
[1/2] STEP 3/3: ADD ./Containerfile /var/tmp/test/
Copying blob cde90dcf8c1f skipped: already exists
Copying blob b50f8aabd929 done   |
Copying config 71ea00d65f done   |
--> 71ea00d65f89
[2/2] STEP 1/2: FROM 71ea00d65f8949486c4441a13b231fd4992b2be2c4170e97a0b9baae11244f71 AS final
[2/2] STEP 2/2: RUN --mount=type=bind,from=test,src=/var/tmp/test,dst=/var/tmp/test,rw     set -x &&     date > /var/tmp/test/Containerfile &&     cat /var/tmp/test/Containerfile
WARN[0000] couldn't find "/var/lib/containers/storage/vfs/dir/7d684fe50918fe44941621b1721c8ee345f7884e2887f8cae36608bacb38e0e8/tmp/test" on host to bind mount into container
+ date
+ cat /var/tmp/test/Containerfile
Wed Feb 12 18:17:34 UTC 2025
[2/2] COMMIT
Copying blob cde90dcf8c1f skipped: already exists
Copying blob b50f8aabd929 skipped: already exists
Copying blob 11db3e39f474 done   |
Copying config 83de1e9298 done   |
--> 83de1e9298fe
83de1e9298feac0ce7e01e89b840e42ecd3901a4a67d1b998b3bdbe176fd3a69
```

Debug output from v1.38 is below (v1.40.0-dev output is substantially similar):

[buildah_v1.38_debug.log.txt](https://github.com/user-attachments/files/18772409/buildah_v1.38_debug.log.txt)

***Note:*** Also attempted with the following `Containerfile` with similar results:

```
FROM registry.fedoraproject.org/fedora-minimal:latest as test

ADD ./Containerfile /test/
RUN chmod 777 /test/Containerfile

#####

FROM test as final

RUN --mount=type=bind,from=test,src=/test,dst=/test,rw \
    set -x && \
    date > /test/Containerfile && \
    cat /test/Containerfile
```

nalind:
#### What type of PR is this?

/kind bug

Only use an overlay for --mounts of type=bind,rw, where changes should be discarded, if the storage driver is overlay.  Otherwise, make a temporary copy.

#### How to verify it

Updated CI to test {vfs|overlay} over {vfs|overlay}, which would have caught this.

Should fix #5988.

Basically what https://github.com/containers/buildah/issues/5988#issuecomment-2718307144 suggested.

```release-note
When not using overlay for storage, running `buildah run` or executing a `buildah build` RUN instruction with the `--mount=type=bind,rw` flag will now use a temporary copy to ensure that writes are discarded, instead of attempting to mount an overlay filesystem using the source location as a "lower".
```

### DIFF
--- a/internal/volumes/volumes.go
+++ b/internal/volumes/volumes.go
@@ -84,14 +84,14 @@ func mountIsReadWrite(m specs.Mount) bool {
 // overlay at the desired location, using the original source location as a
 // "lower", or a temporary copy of the original source location, so that writes
 // to it will succeed, but not affect the original location.
-func convertToOverlayOrTempCopy(m specs.Mount, store storage.Store, mountLabel, tmpDir string, uid, gid int) (newMount specs.Mount, overlayDir string, err error) {
+func convertToOverlayOrTempCopy(m specs.Mount, store storage.Store, mountLabel, tmpDir string, uid, gid int, excludes []string) (newMount specs.Mount, overlayDir string, err error) {
 	switch store.GraphDriverName() {
 	case "overlay":
 		if newMount, overlayDir, err = convertToOverlay(m, store, mountLabel, tmpDir, uid, gid); err != nil {
 			return newMount, "", err
 		}
 	default:
-		if newMount, overlayDir, err = makeTempCopy(m, tmpDir, uid, gid); err != nil {
+		if newMount, overlayDir, err = makeTempCopy(m, tmpDir, uid, gid, excludes); err != nil {
 			return newMount, "", err
 		}
 	}
@@ -148,7 +148,7 @@ func convertToOverlay(m specs.Mount, store storage.Store, mountLabel, tmpDir str
 // a replacement for it that instead bind mounts a temporary copy of the
 // original source location, so that writes to it will succeed without actually
 // modifying the original copy.
-func makeTempCopy(m specs.Mount, tmpDir string, uid, gid int) (specs.Mount, string, error) {
+func makeTempCopy(m specs.Mount, tmpDir string, uid, gid int, excludes []string) (specs.Mount, string, error) {
 	succeeded := false
 	overlayDir, err := overlay.TempDir(tmpDir, uid, gid)
 	if err != nil {
@@ -183,6 +183,7 @@ func makeTempCopy(m specs.Mount, tmpDir string, uid, gid int) (specs.Mount, stri
 	getOptions := copier.GetOptions{
 		ChownDirs:  &forceOwner,
 		ChownFiles: &forceOwner,
+		Excludes:   excludes,
 	}
 	var wg sync.WaitGroup
 	var putErr, getErr error
@@ -242,7 +243,7 @@ func makeTempCopy(m specs.Mount, tmpDir string, uid, gid int) (specs.Mount, stri
 // clean up the overlay filesystem (if we provided a path to it), unmount and
 // remove the mountpoint for the mounted filesystem (if we provided the path to
 // its mountpoint), and then unmount the image (if we mounted one).
-func GetBindMount(sys *types.SystemContext, args []string, contextDir string, store storage.Store, mountLabel string, additionalMountPoints map[string]internal.StageMountDetails, workDir, tmpDir string) (specs.Mount, string, string, string, error) {
+func GetBindMount(sys *types.SystemContext, args []string, contextDir string, store storage.Store, mountLabel string, additionalMountPoints map[string]internal.StageMountDetails, workDir, tmpDir string, excludes []string) (specs.Mount, string, string, string, error) {
 	newMount := specs.Mount{
 		Type: define.TypeBind,
 	}
@@ -386,6 +387,8 @@ func GetBindMount(sys *types.SystemContext, args []string, contextDir string, st
 			}()
 		}
 		contextDir = mountPoint
+		// Don't bother with excludes if this isn't part of the default build context.
+		excludes = nil
 	} else {
 		// special case an additional mount point for "" as shorthand for "preferred location of the default build context"
 		if additionalMountPoints != nil {
@@ -450,7 +453,7 @@ func GetBindMount(sys *types.SystemContext, args []string, contextDir string, st
 
 	overlayDir := ""
 	if !skipOverlay && (mountedImage != "" || mountIsReadWrite(newMount)) {
-		newMount, overlayDir, err = convertToOverlayOrTempCopy(newMount, store, mountLabel, tmpDir, 0, 0)
+		newMount, overlayDir, err = convertToOverlayOrTempCopy(newMount, store, mountLabel, tmpDir, 0, 0, excludes)
 		if err != nil {
 			return newMount, "", "", "", err
 		}
@@ -785,7 +788,7 @@ func GetCacheMount(sys *types.SystemContext, args []string, store storage.Store,
 	overlayDir := ""
 	if needToOverlay {
 		var err error
-		newMount, overlayDir, err = convertToOverlayOrTempCopy(newMount, store, mountLabel, tmpDir, 0, 0)
+		newMount, overlayDir, err = convertToOverlayOrTempCopy(newMount, store, mountLabel, tmpDir, 0, 0, nil)
 		if err != nil {
 			return newMount, "", "", "", nil, err
 		}
@@ -829,8 +832,8 @@ func UnlockLockArray(locks []*lockfile.LockFile) {
 // mountpoints for the bind-mounted paths, unmount any images we mounted, and
 // release the locks we returned (either using UnlockLockArray() or by
 // iterating over them and unlocking them).
-func GetVolumes(ctx *types.SystemContext, store storage.Store, mountLabel string, volumes []string, mounts []string, contextDir string, idMaps define.IDMappingOptions, workDir, tmpDir string) ([]specs.Mount, []string, []string, []string, []*lockfile.LockFile, error) {
-	unifiedMounts, mountedImages, intermediateMounts, overlayMounts, targetLocks, err := getMounts(ctx, store, mountLabel, mounts, contextDir, idMaps.UIDMap, idMaps.GIDMap, workDir, tmpDir)
+func GetVolumes(ctx *types.SystemContext, store storage.Store, mountLabel string, volumes []string, mounts []string, contextDir string, idMaps define.IDMappingOptions, workDir, tmpDir string, excludes []string) ([]specs.Mount, []string, []string, []string, []*lockfile.LockFile, error) {
+	unifiedMounts, mountedImages, intermediateMounts, overlayMounts, targetLocks, err := getMounts(ctx, store, mountLabel, mounts, contextDir, idMaps.UIDMap, idMaps.GIDMap, workDir, tmpDir, excludes)
 	if err != nil {
 		return nil, nil, nil, nil, nil, err
 	}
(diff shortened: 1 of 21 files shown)


tests/run.bats
```diff
@@ -417,32 +411,38 @@ function configure_and_check_user() {
...
-			zflag=z
+			zflag=,z
 		fi
 	fi
 	${OCI} --version
 	_prefetch alpine
 	run_buildah from --quiet --pull=false $WITH_POLICY_JSON alpine
 	cid=$output
 	mkdir -p ${TEST_SCRATCH_DIR}/was:empty
+	if $(stat -f -c %T ${TEST_SCRATCH_DIR}/chroot/upperdir) = overlayfs; then
```
siteshwar, line 422: Please review this finding identified by [OpenScanHub](https://openscanhub.fedoraproject.org/task/53069/log/added.html):

```
Error: SHELLCHECK_WARNING:
/usr/share/buildah/test/system/run.bats:422:58: error[SC2283]: Remove spaces around = to assign (or use [ ] to compare, or quote '=' if literal).
#  420|   	cid=$output
#  421|   	mkdir -p ${TEST_SCRATCH_DIR}/was:empty
#  422|-> 	if $(stat -f -c %T ${TEST_SCRATCH_DIR}/chroot/upperdir) = overlayfs; then
#  423|   		# we'll try to use fuse-overlayfs, which at least through 1.13
#  424|   		# can't accept ":" in layer locations, escaped or not, so bail
```

internal/volumes/volumes.go
```diff
@@ -118,6 +120,87 @@ func convertToOverlay(m specs.Mount, store storage.Store, mountLabel, tmpDir str
 	return mountThisInstead, overlayDir, nil
 }
 
+func makeTempCopy(m specs.Mount, tmpDir string, uid, gid int) (specs.Mount, string, error) {
```
cgwalters, line 123: This function seems like it could use a doc comment.

cgwalters (commented):
I only gave this a quick skim and I don't know the code really, but FWIW looks sane to me.

.cirrus.yml
```diff
@@ -307,7 +307,16 @@ in_podman_task:
...
+    matrix:
+        - env:
+              STORAGE_DRIVER: 'vfs'
+        - env:
+              STORAGE_DRIVER: 'overlay'
+    matrix:
+        - env:
+              NESTED_STORAGE_DRIVER: 'vfs'
+        - env:
+              NESTED_STORAGE_DRIVER: 'overlay'
```
cevich, line 319: I believe cirrus will happily accept this invalid YAML, but I believe `hack/get_ci_vm.sh` will puke on it because it uses python-yaml to parse the file.  If this functionality is important, I believe something like the following would work:

```suggestion
    matrix:
        - env:
              STORAGE_DRIVER: 'vfs'
              NESTED_STORAGE_DRIVER: 'vfs'
        - env:
              STORAGE_DRIVER: 'overlay'
              NESTED_STORAGE_DRIVER: 'vfs'
        - env:
              STORAGE_DRIVER: 'vfs'
              NESTED_STORAGE_DRIVER: 'overlay'
        - env:
              STORAGE_DRIVER: 'overlay'
              NESTED_STORAGE_DRIVER: 'overlay'
```

nalind: Done.

tests/chroot.bats
```diff
@@ -124,11 +124,24 @@ load helpers
   cp -v ${TEST_SOURCES}/containers.conf ${TEST_SCRATCH_DIR}/containers.conf
   chmod ugo+r ${TEST_SCRATCH_DIR}/containers.conf
   mkdir -p ${TEST_SCRATCH_DIR}/chroot
+  ${COPY_BINARY} containers-storage:[${STORAGE_DRIVER}@${TEST_SCRATCH_DIR}/root+${TEST_SCRATCH_DIR}/runroot]docker.io/library/busybox:latest dir:${TEST_SCRATCH_DIR}/base-image
   chown -R 1:1 ${TEST_SCRATCH_DIR}/root ${TEST_SCRATCH_DIR}/runroot ${TEST_SCRATCH_DIR}/chroot
+  if test ${STORAGE_DRIVER} = overlay ; then
+    if test -x /usr/bin/fuse-overlayfs ; then
+      local storage_opts="overlay.mount_program=/usr/bin/fuse-overlayfs"
+    else
+      skip trying to use overlay on top of overlay, but fuse-overlayfs is not present
```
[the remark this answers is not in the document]
nalind: Done.

TomSweeneyRedHat:
@nalind looks lke you need an update here.

tests/run.bats
```diff
@@ -420,23 +420,23 @@ function configure_and_check_user() {
 	cid=$output
 	mkdir -p ${TEST_SCRATCH_DIR}/was:empty
 	# As a baseline, this should succeed.
-	run_buildah run --mount type=tmpfs,dst=/var/tmpfs-not-empty                                                      $cid touch /var/tmpfs-not-empty/testfile
+	run_buildah run --mount type=tmpfs,dst=/var/tmpfs-not-empty                                                            $cid touch /var/tmpfs-not-empty/testfile
```
Honny1: Not blocking: I'm not sure if the spaces are intentional, but if they are to improve readability. It might be better to move the value of the `--mount` argument to a separate variable.

Honny1 (approved):
Thanks, LGTM. I haven't noticed any problems.  I think a rebase will be necessary.

/approve

tests/run.bats
nalind: Ah, that does help.  Thanks!

internal/volumes/volumes.go
```diff
@@ -329,8 +421,15 @@ func GetBindMount(sys *types.SystemContext, args []string, contextDir string, st
 
 	overlayDir := ""
 	if mountedImage != "" || mountIsReadWrite(newMount) {
-		if newMount, overlayDir, err = convertToOverlay(newMount, store, mountLabel, tmpDir, 0, 0); err != nil {
-			return newMount, "", "", "", err
+		switch store.GraphDriverName() {
+		case "overlay":
```
mtrmac: (There is precedent for hard-coding a condition like this … but maybe the checks could all be consolidated into a single function.)

run_linux.go
```diff
@@ -1201,6 +1201,8 @@ func (b *Builder) runSetupVolumeMounts(mountLabel string, volumeMounts []string,
 				UpperDirOptionFragment: upperDir,
 				WorkDirOptionFragment:  workDir,
 				GraphOpts:              slices.Clone(b.store.GraphOptions()),
+				ForceMount:             true,
+				MountLabel:             b.MountLabel,
```
mtrmac, line 1168: > Force the mount of the overlay in runSetupVolumeMounts() so that we can
take advantage of logic that tries to use fuse-overlayfs if we happen to
be on top of overlay, which would otherwise fail.

I read that as potentially unclean… if using `fuse-overlay` this way is expected for an “overlay” mount, shouldn’t the consumer of the created `specs.Mount` be able to handle that universally, instead of us special-casing it and adding an extra mount operation every time?

\*shrug\* I suppose changing all relevant OCI runtimes is unpalatable?

With this, every single caller in Buildah except for `run_freebsd.go` (but not the callers in `podman/libpod`) sets `ForceMount`. Is that correct, or should the option be removed entirely?

mtrmac (commented):
I don’t know enough to judge the `volumeCache…` part.

Overall, I suppose the added test coverage should be decisive.

TomSweeneyRedHat:
@nalind merge conflict here that needs some massaging.

Vince-FIS:
What's the status of this PR? It was approved at some point and then seemingly abandoned.

We run into the same problem in our CI and are interested in a solution to the problem, are there any plans to get this PR back on track in the near future?

internal/volumes/volumes.go
nalind: Factoring these two sites into a single function that decides what to do for its callers.

run_linux.go
nalind: Expecting runtimes to rewrite mounts from the runtime config to use fuse-overlayfs when they're asked to mount overlay on top of overlay would probably be surprising behavior, and as you say, there are many runtimes.
The overlay package is also used by podman, and I don't want to break things there by making this behavior unconditional or the default for them.
The underlying filesystem is different on `freebsd`, which currently rejects attempts to create read-write mounts using the overlay package, so it's moot there.  I don't know if it'll matter there in the future.

contrib/cirrus/lib.sh
```diff
@@ -198,7 +198,7 @@ in_podman() {
                    --cgroupns=host \
                    "${envargs[@]}" \
                    -e BUILDAH_ISOLATION \
-                   -e STORAGE_DRIVER \
+                   -e STORAGE_DRIVER"${NESTED_STORAGE_DRIVER:+=${NESTED_STORAGE_DRIVER}}" \
```
mtrmac: There used to be changes to `.cirrus.yml` that added `NESTED_STORAGE_DRIVER` to the matrix; I can’t see any current equivalent, was that dropped intentionally?

mtrmac (commented):
LGTM — although that should carry little weight, I’m very unfamiliar with the problem space.

@nalind is there an ordering dependency with #5975 (in either direction)?

universam1:
@mtrmac Please any update on this fix? We are blocked right now in our CI due to this issue. Thanks!

PR #6128, opened 2025-04-15
cli: hide the `completion` command instead of disabling it outright

nalind:
#### What type of PR is this?

/kind other

Hide the cobra built-in `completion` command instead of disabling it outright, since eventually we want to switch to using it instead of the hand-written scripts we're currently using.

#### How to verify it

Run `buildah completion` and if it doesn't only produce an error, we're good.

None

```release-note
None
```


flouthoc (approved):
LGTM

Luap99 (approved):
/lgtm

PR #6132, opened 2025-04-16
run: handle relabeling bind mounts ourselves

nalind:
#### What type of PR is this?

/kind bug

Handle requested relabeling of bind mounts (i.e., the "z" and "Z" flags) directly, instead of letting the runtime handle the relabeling.

#### How to verify it

We now also test with runc on Fedora, which exercises relabeling logic.

```release-note
None
```


TomSweeneyRedHat:
@nalind the changes LGTM, but you have some test errors that don't look related, but they don't look like flakes either.
@flouthoc PTAL

run_linux.go
```diff
@@ -543,6 +543,33 @@ rootless=%d
...
+	// Handle mount flags that request that the source locations for "bind" mountpoints be
+	// relabeled, and filter those flags out of the list of mount options we pass to the
+	// runtime.
+	for i := range spec.Mounts {
+		switch spec.Mounts[i].Type {
+		default:
+			continue
+		case "bind", "rbind":
+			// all good, keep going
+		}
```
flouthoc, line 555:
```suggestion
		if switch spec.Mounts[i].Type == "bind" ||  switch spec.Mounts[i].Type  == "rbind" {
		
```

Do you think `if` is easier for readability ? We are just handling one case.

.cirrus.yml
```diff
@@ -207,6 +209,14 @@ integration_task:
...
             IMAGE_NAME: "${FEDORA_CACHE_IMAGE_NAME}"
             STORAGE_DRIVER: 'vfs'
+            BUILDAH_RUNTIME: crun
+            RUNTIME_N: " using crun"
+        - env:
+            DISTRO_NV: "${FEDORA_NAME}"
+            IMAGE_NAME: "${FEDORA_CACHE_IMAGE_NAME}"
+            STORAGE_DRIVER: 'vfs'
+            BUILDAH_RUNTIME: runc
+            RUNTIME_N: " using runc"
```
Luap99, line 219: We have dropped all runc testing not to long ago in podman. It was a huge maintenance headache.

Every image update now has yet another big failure point which causes regressions. As the person who currently manages all the updates I am strongly against making my life harder with these things.

Luap99 (changes requested):

.cirrus.yml
Luap99: Almost every CI image update contains one or more regressions in our dependencies, kernel, systemd, golang, crun, glibc, well you get the point. By testing with runc every runc update is now a new potential failure point that then blocks image updates and requires me (or whoever does the image update) to do the work of investigating the test failure, reporting the bug and/or adding a work around or skip to test. Can that be done? Sure, but we decided over a year ago that we no longer want to test runc (https://github.com/containers/podman/pull/22706) and I at least am not interested in doing a ton of extra work for it.

nalind: Then those can be marked as not required for the "Total Success" check.  But the idea of fixing this bug that runc exposes, and then not taking available steps to help ensure that we won't encounter something similar again, doesn't sit well with me.

nalind: We could have caught this bug earlier.  If we need to have a conversation about that, then I think we should have the conversation.

danishprakash: I understand @Luap99's concerns about this being yet another failure point that may or may not block releases, or that adds additional work. That being said, we use runc extensively with podman, so this is something that would be helpful in the long run. I can take up the work that @Luap99 is referring to with respect to runc and any issues arising out of it, if that seems to help, and take this forward.

flouthoc (approved):
LGTM

rpm/buildah.spec
```diff
@@ -137,6 +137,7 @@ export BUILDTAGS+=" libtrust_openssl"
 %gobuild -o bin/copy ./tests/copy
 %gobuild -o bin/tutorial ./tests/tutorial
 %gobuild -o bin/inet ./tests/inet
+%gobuild -o bin/dumpspec ./tests/dumpspec
```
Luap99, line 140: Why is this packaged? I don't see it used in the tests?

nalind:
@containers/buildah-maintainers opinions on cherry-picking this for release-1.40?

danishprakash:
Possibly even for release-1.39? We originally reported this for 1.39.x. We can carry the backport easily but it'd be helpful for consumers of buildah on the `stable-1 minor version` release pattern if this were to be backported.

Luap99:
I would be fine to backport to 1.40, as for older releases we only ever support the latest version upstream so I would say you "must" update.

Issue #6160, opened 2025-05-06
`containers.label_users` is ignored by `podman build`


gucci-on-fleek:
Even with `containers.label_users` set to `true` in `containers.conf`, `podman build` fails with SELinux permission errors.

### Reproduction

```console
$ id -Z
user_u:user_r:user_t:s0-s0:c0.c1023

$ cat /etc/containers/containers.conf

[containers]
label_users=true

$ podman run --rm -it fedora-minimal:latest /usr/bin/true

$ cat > Containerfile <<EOF
FROM fedora-minimal:latest
RUN /usr/bin/true
EOF

$ podman build --no-cache .
STEP 1/2: FROM fedora-minimal:latest
STEP 2/2: RUN /usr/bin/true
exec container process `/bin/sh`: Permission denied
Error: building at STEP "RUN /usr/bin/true": while running runtime: exit status 1

$ podman build --security-opt=label=role:user_r --security-opt=label=user:user_u --no-cache .
STEP 1/2: FROM fedora-minimal:latest
STEP 2/2: RUN /usr/bin/true
COMMIT
--> d28ad43735ae
d28ad43735ae6fbe6dee4c9ef43506d6a160d617cacc0d28bd1b514acde3aaf7
```

### Logs

```log
AVC avc:  denied  { transition } for  pid=551214 comm="crun" path="/usr/bin/bash" dev="overlay" ino=7626070 scontext=user_u:user_r:container_runtime_t:s0-s0:c0.c1023 tcontext=system_u:system_r:container_t:s0:c915,c1004 tclass=process permissive=0
```

### Versions

I'm able to reproduce this on both Fedora 41 and Fedora 42:

```console
$ cat /etc/os-release | grep PRETTY_NAME
PRETTY_NAME="Fedora Linux 41 (Workstation Edition)"

$ podman --version
podman version 5.4.2

$ uname --kernel-release
6.14.4-200.fc41.x86_64

$ rpm --query --queryformat='%{NAME}\t%{VERSION}\n' podman container-selinux containers-common selinux-policy-targeted | column --table
podman                   5.4.2
container-selinux        2.236.0
containers-common        0.62.2
selinux-policy-targeted  41.38
```

```console
$ cat /etc/os-release | grep PRETTY_NAME
PRETTY_NAME="Fedora Linux 42 (IoT Edition)"

$ podman --version
podman version 5.4.2

$ uname --kernel-release
6.14.5-300.fc42.x86_64

[...]

PR #6161, opened 2025-05-06
Support label_users in buildah

rhatdan:
Fixes: https://github.com/containers/buildah/issues/6160

label_users tells buildah and podman to maintain the user and role from the SELinux label, the default is to change the user and role to system_u:system_r.

With this change we end up with an unconfined_u user running the container as unconfined_u:unconfined_r.

I don't believe our CI/CD system can test this, because it runs with system_u:system_r by default.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
buildah build now supports label_users from containers.conf
```

### DIFF
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -35,6 +35,7 @@ import (
 	securejoin "github.com/cyphar/filepath-securejoin"
 	units "github.com/docker/go-units"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/selinux/go-selinux"
 	"github.com/openshift/imagebuilder"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


rhatdan:
Code was taken from podman, might want to eventually move to common.

flouthoc (approved):
LGTM, missing release notes.

pkg/parse/parse.go
```diff
@@ -81,6 +82,25 @@ func CommonBuildOptions(c *cobra.Command) (*define.CommonBuildOptions, error) {
...
+	label, err := selinux.CurrentLabel()
+	if err != nil {
+		return nil, err
+	}
+	if label == "" {
+		return nil, nil
+	}
+	con, err := selinux.NewContext(label)
+	if err != nil {
+		return nil, err
```
nalind, line 96: This really should have been wrapped with more context somewhere in the call stack - the only error this can return just says "invalid label".

PR #6168, opened 2025-05-15
copier: use `PAX` format instead of default `USTAR`

flouthoc:
File copied to container from host does not preserves `mtime` correctly because `USTAR` format only supports precision of 1 second, switch to `PAX` format which typically supports Nanosecond precision.

Helpful in RUN-2664: `Ensure that Buildah instructions that copy files do not make any changes`

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
Ensure that Buildah instructions that copy retains mtime with nanosecond precision
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -1545,6 +1545,8 @@ func mapWithPrefixedKeysWithoutKeyPrefix[K any](m map[string]K, p string) map[st
 func copierHandlerGetOne(srcfi os.FileInfo, symlinkTarget, name, contentPath string, options GetOptions, tw *tar.Writer, hardlinkChecker *hardlinkChecker, idMappings *idtools.IDMappings) error {
 	// build the header using the name provided
 	hdr, err := tar.FileInfoHeader(srcfi, symlinkTarget)
+	// Use PAX format to support high-precision timestamps for `ModTime`.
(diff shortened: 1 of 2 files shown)


nalind:
We try to not force PAX format unless we absolutely have to, since it makes the layer diff larger.  I'm inclined to reject this.

flouthoc:
Inside container buildkit retains nanosecond precision for `mtime`, I am suspecting that buildkit is using `PAX` format by default. But I need to confirm this.

nalind:
The header format used in an image's layer blobs can control how much precision there is the layer blob, and whether or not ctime and atime are included.  A container's contents sit on a regular filesystem, and those keep timestamps with levels of precision that vary depending on which filesystem is being used.  We really shouldn't be assuming anything about the latter.

flouthoc:
Yes, I also found a example in buildkit which created `USTAR` tar so my assumption was wrong.

```
FROM scratch
COPY somefile .
```

created a layer with `USTAR` .

I'll open a new PR if I find any other method I'm closing this one now.

PR #6170, opened 2025-05-19
[skip-ci] Packit: set fedora-all after F40 EOL

lsm5:
As the title says.

#### What type of PR is this?

 /kind other

Update fedora targets in packit

#### How to verify it

See packit jobs

None

None

```release-note
None
```

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -95,8 +95,7 @@ jobs:
   - job: tests
     trigger: pull_request
     packages: [buildah-fedora]
-    targets: &fedora_copr_test_targets
-      - fedora-all-x86_64
+    targets:
       - fedora-all-x86_64
     tf_extra_params:
       environments:
@@ -109,7 +108,7 @@ jobs:
   - job: tests
     trigger: ignore
     packages: [buildah-centos]
-    targets: &centos_copr_test_targets
+    targets:
       - centos-stream-9-x86_64
       - centos-stream-10-x86_64
     tf_extra_params:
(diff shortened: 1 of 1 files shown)


Luap99 (commented):
Should fedora_copr_test_targets also be updated?

lsm5:
> Should fedora_copr_test_targets also be updated?

good catch, yes. Will update.

Luap99 (approved):
LGTM

lsm5:

lsm5:
> Are there docs for what these mean... somewhere?

1. [Packit docs - Aliases](https://packit.dev/docs/configuration#aliases)
2. There's `python3-fedora-distro-aliases` package which provides `resolve-fedora-aliases`
```
$ resolve-fedora-aliases fedora-all
fedora-41 fedora-42 fedora-rawhide
```

Let me know if you'd like this mentioned in the packit config.

nalind:
That'd be helpful.  Do we reuse most of these named lists of targets somewhere else in the configuration?

lsm5:
> That'd be helpful. 

Sure, I'll update.

> Do we reuse most of these named lists of targets somewhere else in the configuration?

There may be reuse with yaml anchors in cases where the test targets can be exactly the same as build targets. But that currently ain't the case given we're not testing on aarch64 here. The reuse will be limited to `.packit.yaml`.

lsm5:
Done. Added another commit to cleanup unused yaml anchors and redundant test target.

Issue #6152, opened 2025-05-01
use of cached layers in builds is inconsistent


dustymabe:
In the CoreOS team we found that cached layers that were getting re-used for later builds violated the [principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). 

Here's a simple reproducer with example output:

```
[dustymabe@media podman-build-cache-bug]$ echo 'abc' > file
[dustymabe@media podman-build-cache-bug]$ cat Containerfile 
FROM quay.io/fedora/fedora:42
RUN <<EOF
    curl http://0.0.0.0:8000/file | tee /file
EOF
[dustymabe@media podman-build-cache-bug]$ podman build -t test .
STEP 1/2: FROM quay.io/fedora/fedora:42
STEP 2/2: RUN <<EOF (curl http://0.0.0.0:8000/file | tee /file)
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100     4  100     4    0     0   1291      0 --:--:-- --:--:-- --:--:--  1333
abc
COMMIT test
--> e318ca889b77
Successfully tagged localhost/test:latest
e318ca889b77318233d8a3df8ede75e1d506315d0da9951b1b412102cf6548ba
[dustymabe@media podman-build-cache-bug]$ 
[dustymabe@media podman-build-cache-bug]$ echo 'def' > file
[dustymabe@media podman-build-cache-bug]$ podman build --cache-ttl=.01s -t test .
STEP 1/2: FROM quay.io/fedora/fedora:42
STEP 2/2: RUN <<EOF (curl http://0.0.0.0:8000/file | tee /file)
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100     4  100     4    0     0   3241      0 --:--:-- --:--:-- --:--:--  4000
def
COMMIT test
--> a8e31050715f
Successfully tagged localhost/test:latest
a8e31050715fdff1df846ce087a11fb93e7129026324675670b4db14fb1b006d
[dustymabe@media podman-build-cache-bug]$ echo 'ghi' > file
[dustymabe@media podman-build-cache-bug]$ podman build -t test .
STEP 1/2: FROM quay.io/fedora/fedora:42
STEP 2/2: RUN <<EOF (curl http://0.0.0.0:8000/file | tee /file)
[...]

PR #6171, opened 2025-05-19
imagebuildah: select most recent layer for cache in case on conflict

flouthoc:
If multiple layers are found as cache candidates then select the layer which was created recently.

Closes: https://github.com/containers/buildah/issues/6152

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
When using --layers buildah will select most recent layer in case of conflict
```


flouthoc:
@nalind PTAL

imagebuildah/stage_executor.go
```diff
@@ -2246,6 +2246,7 @@ func (s *StageExecutor) pullCache(ctx context.Context, cacheKey string) (referen
 // intermediateImageExists returns true if an intermediate image of currNode exists in the image store from a previous build.
```
nalind, line 2246: This godoc needs to be updated: the method returns the ID of a matching image, and if there are multiple matches, it returns the one with the most recent "created" date.

tests/bud.bats
```diff
@@ -5513,6 +5513,32 @@ _EOF
...
+  # Populate cache again with new layers
+  run_buildah build $WITH_POLICY_JSON --layers --cache-ttl=.01s --iidfile ${TEST_SCRATCH_DIR}/image1.txt -t source -f $contextdir/Dockerfile1
+  # Should not contain `Using cache` since all
+  # cached layers are older than .01s.
+  assert "$output" !~ "Using cache"
+  # Build again use cache and make sure most recently built image is reused
+  run_buildah build $WITH_POLICY_JSON --layers --iidfile ${TEST_SCRATCH_DIR}/image2.txt -t source -f $contextdir/Dockerfile1
+  # Must use cache
+  expect_output --substring "Using cache"
+  # both of these should have just been the base image's ID, which shouldn't have changed the second time around
```
nalind, line 5538: What's the "base image" being referred to here?  If it's alpine, then the comment is wrong.

nalind (changes requested):
`sortImagesByCreated()` could use a unit test if it's not being removed.

nalind:
/approve
LGTM

nalind (approved):

flouthoc:

flouthoc:

PR #6175, opened 2025-05-20
[CI:DOCS] update a couple of lists in the build man page

nalind:
#### What type of PR is this?

/kind documentation

The buildah-build man page enumerates the default set of masked and read-only paths, but that list is hardwired into one of our dependencies, and we didn't update this man page when it changed.

#### How to verify it

Visual inspection.

None

```release-note
None
```


flouthoc (approved):
LGTM
/lgtm
/approve

PR #6176, opened 2025-05-20
pkg/parse.GetBuildOutput(): use strings.Cut()

nalind:
#### What type of PR is this?

/kind cleanup

Use strings.Cut() to make pkg/parse.GetBuildOutput() a little easier to follow.

#### How to verify it

New unit test!

None

```release-note
None
```

### DIFF
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -704,7 +704,7 @@ func AuthConfig(creds string) (*types.DockerAuthConfig, error) {
 // GetBuildOutput is responsible for parsing custom build output argument i.e `build --output` flag.
 // Takes `buildOutput` as string and returns BuildOutputOption
 func GetBuildOutput(buildOutput string) (define.BuildOutputOption, error) {
-	if len(buildOutput) == 1 && buildOutput == "-" {
+	if buildOutput == "-" {
 		// Feature parity with buildkit, output tar to stdout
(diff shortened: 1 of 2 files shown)


pkg/parse/parse.go
```diff
@@ -723,56 +723,48 @@ func GetBuildOutput(buildOutput string) (define.BuildOutputOption, error) {
...
+		switch key {
 		case "type":
-			if typeSelected {
-				return define.BuildOutputOption{}, fmt.Errorf("duplicate %q not supported", arr[0])
+			if typeSelected != "" {
+				return define.BuildOutputOption{}, fmt.Errorf("duplicate %q not supported", key)
 			}
-			typeSelected = true
-			switch arr[1] {
+			typeSelected = value
```
Luap99, line 738: Why is typeSelected a string now? I don't see where this value would ever be used other than the == "" check so a bool feels easier to use?

nalind: It'll be useful when more exporter types are introduced, at some point, by someone.  In the meantime I can change the next line to read the value it switches on from this variable instead of `value`.

Luap99: I mean if you will need it sure but it feels weird to change that as side effect of switching to strings.Cut()

Luap99 (approved):
LGTM

PR #6189, opened 2025-05-29
commit: add --source-date-epoch and --rewrite-timestamp flags

nalind:
#### What type of PR is this?

/kind feature

Add a `--source-date-epoch flag`, defaulting to `$SOURCE_DATE_EPOCH` if set, which sets the created-on date and the timestamp for the new history entries, but does not default to modifying the timestamps on contents in new layers.

Add a `--rewrite-timestamp` flag, which "clamps" timestamps in the new layers to not be later than the `--source-date-epoch` value if both the `--rewrite-timestamp` and `--source-date-epoch` flags were set.

#### How to verify it

New integration tests!

None

```release-note
`buildah commit` now recognizes `--source-date-epoch` and `--rewrite-timestamp` flags, which affect the dates recorded in the new image's configuration and the timestamps on the contents of the new layer.
```

### DIFF
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -5,11 +5,13 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/containers/buildah"
 	"github.com/containers/buildah/define"
+	"github.com/containers/buildah/internal"
 	"github.com/containers/buildah/pkg/cli"
 	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/buildah/util"
@@ -39,6 +41,8 @@ type commitInputOptions struct {
 	manifest           string
 	omitTimestamp      bool
 	timestamp          int64
+	sourceDateEpoch    string
+	rewriteTimestamp   bool
 	quiet              bool
 	referenceTime      string
 	rm                 bool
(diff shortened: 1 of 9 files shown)


cmd/buildah/commit.go
```diff
@@ -327,6 +348,9 @@ func commitCmd(c *cobra.Command, args []string, iopts commitInputOptions) error
 		timestamp := time.Unix(0, 0).UTC()
 		options.HistoryTimestamp = &timestamp
 	}
+	if exclusiveFlags > 1 {
+		return errors.New("can not use more then one timestamp option at at time")
```
flouthoc:
```suggestion
		return errors.New("cannot use more then one timestamp option at at time")
```

nit

tests/commit.bats
```diff
@@ -400,3 +400,145 @@ load helpers
...
+  local datestamp="1970-01-01T00:01:00Z"
+  mkdir -p $TEST_SCRATCH_DIR/context
+  createrandom $TEST_SCRATCH_DIR/context/randomfile1
+  createrandom $TEST_SCRATCH_DIR/context/randomfile2
+  run_buildah from -q busybox
+  local cid="$output"
+  run_buildah add --add-history "$cid" $TEST_SCRATCH_DIR/context/* /context
+  # commit using defaults
+  run_buildah commit "$cid" oci:$TEST_SCRATCH_DIR/default
+  # commit with an implicity-provided timestamp
```
flouthoc:
```suggestion
  # commit with an implicitly-provided timestamp
```

image.go
```diff
@@ -1229,9 +1212,40 @@ func (i *containerImageRef) makeExtraImageContentDiff(includeFooter bool) (_ str
 	return diff.Name(), digester.Digest(), counter.Count, nil
 }
 
+func makeFilteredLayerWriteCloser(wc io.WriteCloser, layerModTime, layerLatestModTime *time.Time) io.WriteCloser {
```
flouthoc, line 1222: Nit: A small comment about this function will be helpful documenting precedence order between `layerModTime` and `layerLatestModTime` ?

flouthoc (approved):
PR LGTM just small nits and one question.

image.go
nalind: Added.

tests/commit.bats
nalind: Updated.

cmd/buildah/commit.go
nalind: Changed.

flouthoc (approved):
LGTM

flouthoc:

TomSweeneyRedHat:
LGTM

PR #6192, opened 2025-05-30
Don't BuildRequires: ostree-devel

mtrmac:
We are not opting into the ostree backend, and it doesn't build: https://github.com/containers/image/pull/2821 . So, stop referencing the dependency.

Should not change behavior.

@lsm5 @jnovy PTAL.

#### What type of PR is this?

> /kind cleanup

None

```release-note
None
```


jnovy (approved):
LGTM

Issue #6183, opened 2025-05-24
add `--unsetannotation` option to `podman build`


performantdata:
### Feature request description

Add an `--unsetannotation` option, analogous to `--unsetlabel`, to `podman build`. Currently, there is no way to remove an annotation that is inherited from the parent image.

flouthoc:
Created a PR https://github.com/containers/buildah/pull/6195

PR #6195, opened 2025-06-01
build,config: add support for `--unsetannotation`

flouthoc:
Just like `--unsetlabel` add support for `--unsetannotation`.

Closes: https://github.com/containers/buildah/issues/6183

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
build,config: add support for --unsetannotation, which allows users to prevent inheriting a specific annotation from base image.
```


rhatdan:
LGTM

nalind:
How does this affect caching and cache evaluation?

flouthoc:
@nalind `--unsetannotation` must invalidate the cache if used and the build should remove annotation from final image instead of using any previous image from the cache, I have added a test to verify this.

tests/bud.bats
```diff
@@ -2670,6 +2670,77 @@ _EOF
...
+  run_buildah --version
+  local -a output_fields=($output)
+  buildah_version=${output_fields[2]}
+
+  buildah inspect --format '{{ .ImageAnnotations }}' $base
+  not_want_output='map[]'
+  assert "$output" != "$not_want_output" "expected some annotations to be set in base image $base"
+
+  labels=$(buildah inspect --format '{{ range $key, $value := .ImageAnnotations }}{{ $key }} {{end}}' $base)
+  labelflags="--annotation hello=world"
```
nalind: These are confusing names to give to these variables.

tests/bud.bats
```diff
@@ -2670,6 +2670,77 @@ _EOF
...
+  for label in $labels; do
+    if test $label != io.buildah.version ; then
+      labelflags="$labelflags --unsetannotation $label"
+    fi
+  done
+
+  ## Since we are unsetting something, this should not use previous image present in the cache.
+  run_buildah build $WITH_POLICY_JSON --layers $labelflags -t $target --from $base $BUDFILES/base-with-labels
+  ## Should not contain `Using Cache`
+  assert "$output" !~ "Using cache"
```
nalind: Nit: mismatched capitalization

imagebuildah/stage_executor.go
```diff
@@ -1880,10 +1880,14 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
...
 	}
 	inheritLabels := ""
+	unsetAnnotations := ""
 	// If --inherit-label was manually set to false then update history.
 	if s.executor.inheritLabels == types.OptionalBoolFalse {
 		inheritLabels = "|inheritLabels=false"
 	}
+	for _, annotation := range s.executor.unsetAnnotations {
+		unsetAnnotations += "|unsetAnnotation=" + annotation
+	}
```
nalind, line 1896: This is added to the history CreatedBy value for every step in every stage of the build.  I don't think we want that, as I think we only care about the flag's effect on the final image, and doing this for every step creates more cache misses when the flag is used for a build after someone realizes they forgot they wanted to use it.

imagebuildah/executor.go
```diff
@@ -319,6 +320,7 @@ func newExecutor(logger *logrus.Logger, logPrefix string, store storage.Store, o
 		logPrefix:                               logPrefix,
 		unsetEnvs:                               slices.Clone(options.UnsetEnvs),
 		unsetLabels:                             slices.Clone(options.UnsetLabels),
+		unsetAnnotations:                        slices.Clone(options.UnsetAnnotations),
```
nalind, line 323: This slice should get sorted at some point, so that changing the order in which the values are given on the command line won't be enough to create a cache miss where there shouldn't be one.

nalind (changes requested):
Needs to sort the flag values, concerns about the history entries.

imagebuildah/stage_executor.go
flouthoc: I agree, should we only add it to last step of the last stage ?

nalind: Preferably, yes, but that's going to require changes elsewhere so that this accurately reflects whether that change was applied or not.

imagebuildah/executor.go
flouthoc: Doing this below.

imagebuildah/stage_executor.go
flouthoc: Done.

tests/bud.bats

tests/bud.bats
flouthoc: I did not get what this comment was referring to.

imagebuildah/stage_executor.go
```diff
@@ -1458,7 +1458,7 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			// stage.
 			if lastStage || imageIsUsedLater {
 				logCommit(s.output, i)
-				createdBy, err := s.getCreatedBy(node, addedContentSummary)
+				createdBy, err := s.getCreatedBy(node, addedContentSummary, true)
```
nalind: Why is this hard-coded to `true` when it's conditional for the `commit()` call a few lines later?

imagebuildah/stage_executor.go
```diff
@@ -1293,7 +1293,7 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 			// No base image means there's nothing to put in a
 			// layer, so don't create one.
 			emptyLayer := (s.builder.FromImageID == "")
-			createdBy, err := s.getCreatedBy(nil, "")
+			createdBy, err := s.getCreatedBy(nil, "", true)
```
nalind: Why is this hard-coded to `true` when it's conditional for the `commit()` call a few lines later?

imagebuildah/stage_executor.go
```diff
@@ -1725,7 +1725,7 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 
 		if lastInstruction && lastStage {
 			if s.executor.squash || s.executor.confidentialWorkload.Convert || len(s.executor.sbomScanOptions) != 0 {
-				createdBy, err := s.getCreatedBy(node, addedContentSummary)
+				createdBy, err := s.getCreatedBy(node, addedContentSummary, true)
```
nalind: Why is this hard-coded to `true` when it's conditional for the `commit()` call a few lines later?

imagebuildah/stage_executor.go
```diff
@@ -2431,6 +2441,9 @@ func (s *StageExecutor) commit(ctx context.Context, createdBy string, emptyLayer
 		annotationk, annotationv, _ := strings.Cut(annotationSpec, "=")
 		s.builder.SetAnnotation(annotationk, annotationv)
 	}
+	for _, key := range s.executor.unsetAnnotations {
+		s.builder.UnsetAnnotation(key)
+	}
```
nalind, line 2448: Shouldn't this be conditional on `lastInstruction`, in order to agree with the history entry that we'll be writing to the image?

flouthoc:
> The history now only notes that the annotation was suppressed at the final instruction of a stage, while it will have been suppressed at the first instruction in a stage.

Technically annotation only exists at `base` image there is no other way to add annotation apart from CLI flag, from that perspective. Does it make more sense to add it to history of `first step` of a stage which contains annotation ? and not the `last step` and `last instruction` ? But one problem which it can cause is making all the lower steps and lower stages not valid cache hits even though only annotation part did not match. I wonder if that is a good idea ?

imagebuildah/stage_executor.go
flouthoc: I moved it  inside ` if finalInstruction` block

nalind:
The tests should include a case where there are no instructions in the Dockerfile after the FROM line, e.g.:
```Dockerfile
FROM busybox
```
I think there's some logic needed for that case that we don't have yet.

flouthoc:
> The tests should include a case where there are no instructions in the Dockerfile after the FROM line, e.g.:
> 
> ```dockerfile
> FROM busybox
> ```
> 
> I think there's some logic needed for that case that we don't have yet.

Done added case for this, also added test. However `--layers` will not work with this because there are no layers so we cannot write history. We can write `empty_layer` and store history there but no sure if it is needed for this patch and use-case.

flouthoc:
@nalind PTAL

nalind:
> Done added case for this, also added test. However `--layers` will not work with this because there are no layers so we cannot write history. We can write `empty_layer` and store history there but no sure if it is needed for this patch and use-case.

Why shouldn't it produce _some_ image, when `--layers` is the default in `podman build`?

flouthoc:
> > Done added case for this, also added test. However `--layers` will not work with this because there are no layers so we cannot write history. We can write `empty_layer` and store history there but no sure if it is needed for this patch and use-case.
> 
> Why shouldn't it produce _some_ image, when `--layers` is the default in `podman build`?

Oh maybe I phrased it wronly, it produces an image but it does not results in returning the same image again when built again.

For example

```
FROM busybox
```

```
buildah --unsetannotation org.opencontainers.image.name build --layers -t test .
```

Produces an image but if I do this again

```
buildah --unsetannotation org.opencontainers.image.name build --layers -t test .
```

it produces a new sha instead of returning the old one, since nothing has changed. Since our cache checking logic clicks in only for builds with more than `1` steps in it.

nalind (commented):
That new test case raises the question of whether or not it's a bug that we don't set an identity label when a build doesn't have instructions that force us to commit a new image, but this at least LGTM.

flouthoc:
> That new test case raises the question of whether or not it's a bug that we don't set an identity label when a build doesn't have instructions that force us to commit a new image, but this at least LGTM.

Okay I will open a different issue for this since it is not related to this PR. @containers/buildah-maintainers PTAL

flouthoc:

Issue #6153, opened 2025-05-03
Unable to suppress inheritance of annotations from base image


jackwhelpton:
### Issue Description

Looking to remove (rather than overwrite) some annotations present in the base image, but seem unable to do so.

### Steps to reproduce the issue

Have tried various permutations such as:

```sh
buildah bud --annotation org.opencontainers.image.url=-

buildah bud --annotation org.opencontainers.image.url- .
```

Have also experimented with the `inherit-labels=false` flag.

### Describe the results you received

In the first case, the annotation acquires the literal value `-`. In the second case, a new annotation with a trailing dash in the name is created. The `inherit-labels` flag appears to have no impact on annotations.

### Describe the results you expected

The annotation `org.opencontainers.image.url` present in the base image should no longer be visible in the new derived image.

### buildah version output

```yaml
Version:         1.40.0
Go Version:      go1.24.2
Image Spec:      1.1.1
Runtime Spec:    1.2.1
CNI Spec:        1.1.0
libcni Version:  v1.3.0
image Version:   5.35.0
Git Commit:      ff62e3e8a5da56352e90039fb4747ff8438b7b5c
Built:           Mon Apr 21 18:24:35 2025
OS/Arch:         linux/amd64
BuildPlatform:   linux/amd64
```

### buildah info output

```yaml
{
    "host": {
        "CgroupVersion": "v2",
        "Distribution": {
            "distribution": "wolfi",
            "version": "20230201"
        },
        "MemFree": 25018122240,
        "MemTotal": 32882257920,
        "OCIRuntime": "crun",
        "SwapFree": 0,
        "SwapTotal": 0,
        "arch": "amd64",
        "cpus": 16,
        "hostname": "runner-vvkbyzurm-project-54529948-concurrent-0-6rl9flj5",
        "kernel": "6.1.132",
        "os": "linux",
        "rootless": false,
        "uptime": "1m 49.52s",
        "variant": ""
    },
    "store": {
        "ContainerStore": {
            "number": 0
        },
        "GraphDriverName": "overlay",
        "GraphOptions": null,
        "GraphRoot": "/var/lib/containers/storage",
        "GraphStatus": {
            "Backing Filesystem": "overlayfs",
            "Native Overlay Diff": "false",
            "Supports d_type": "true",
            "Supports shifting": "true",
            "Supports volatile": "true",
            "Using metacopy": "false"
        },
        "ImageStore": {
            "number": 20
        },
        "RunRoot": "/run/containers/storage"
[...]

PR #6198, opened 2025-06-02
build: add support for `--inherit-annotations`

flouthoc:
Allows users to specify if they want to inherit annotations from base image or not.

Closes: https://github.com/containers/buildah/issues/6153

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
build: add support for `--inherit-annotations` which allows users to specify if they want to inherit annotations from base image or not.
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -1890,6 +1890,8 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
 	}
 	inheritLabels := ""
 	unsetAnnotations := ""
+	inheritAnnotations := ""
+	newAnnotations := ""
 	// If --inherit-label was manually set to false then update history.
 	if s.executor.inheritLabels == types.OptionalBoolFalse {
 		inheritLabels = "|inheritLabels=false"
@@ -1897,6 +1899,16 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
 	if isLastStep {
 		for _, annotation := range s.executor.unsetAnnotations {
 			unsetAnnotations += "|unsetAnnotation=" + annotation
+		}
+		// If --inherit-annotation was manually set to false then update history.
+		if s.executor.inheritAnnotations == types.OptionalBoolFalse {
+			inheritAnnotations = "|inheritAnnotations=false"
+		}
+		// If new annotations are added, they must be added as part of the last step of the build,
+		// so mention in history that new annotations were added inorder to make sure the builds
+		// can either reuse layers or burst the cache depending upon new annotations.
+		if len(s.executor.annotations) > 0 {
+			newAnnotations += strings.Join(s.executor.annotations, ",")
 		}
 	}
 	switch strings.ToUpper(node.Value) {
@@ -1907,7 +1919,7 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
 			}
 		}
 		buildArgs := s.getBuildArgsKey()
-		return "/bin/sh -c #(nop) ARG " + buildArgs + inheritLabels + unsetAnnotations, nil
+		return "/bin/sh -c #(nop) ARG " + buildArgs + inheritLabels + unsetAnnotations + inheritAnnotations + newAnnotations, nil
 	case "RUN":
 		shArg := ""
 		buildArgs := s.getBuildArgsResolvedForRun()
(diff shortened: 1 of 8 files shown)


tests/bud.bats
```diff
@@ -141,6 +141,46 @@ _EOF
...
+
+  ## Build without removing annotations
+  run_buildah build $WITH_POLICY_JSON --layers --iidfile ${TEST_SCRATCH_DIR}/iid1 -t $target --from $base $BUDFILES/base-with-labels
+  ## Second build must use cache
+  run_buildah build $WITH_POLICY_JSON --layers --iidfile ${TEST_SCRATCH_DIR}/iid2 -t $target --from $base $BUDFILES/base-with-labels
+  ## Must use cache
+  expect_output --substring " Using cache"
+  cmp ${TEST_SCRATCH_DIR}/iid1 ${TEST_SCRATCH_DIR}/iid2
+
+  ## Since we are inheriting no labels, this should not use previous image present in the cache.
```
```suggestion
  ## Since we are inheriting no annotations, this should not use previous image present in the cache.
```

imagebuildah/stage_executor.go
```diff
@@ -1880,10 +1880,15 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
...
 	inheritLabels := ""
+	inheritAnnotations := ""
 	// If --inherit-label was manually set to false then update history.
 	if s.executor.inheritLabels == types.OptionalBoolFalse {
 		inheritLabels = "|inheritLabels=false"
 	}
+	// If --inherit-annotation was manually set to false then update history.
+	if s.executor.inheritAnnotations == types.OptionalBoolFalse {
+		inheritAnnotations = "|inheritAnnotations=false"
+	}
```
nalind, line 1913: This shares a behavior with inherit-labels, one which I didn't notice at the time:  after the first step, when it's had its effect and the flag doesn't affect anything (when we're committing at each step, a second step in a stage is going to start with a base that has no annotations to inherit), we keep adding this bit to the history.

tests/bud.bats
```diff
@@ -141,6 +141,46 @@ _EOF
   expect_output "0" "layer should not exist"
 }
 
+@test "bud and test --inherit-annotations with --layers" {
```
nalind, line 193: Is there a corresponding test that doesn't use `--layers`?

tests/bud.bats
```diff
@@ -141,6 +141,46 @@ _EOF
...
+  ## Should not contain `Using Cache`
+  assert "$output" !~ "Using cache"
+
+  # no annotations should be inherited from base image
+  want_output='map[]'
+  run_buildah inspect --format '{{printf "%q" .ImageAnnotations}}' $target
+  expect_output "$want_output"
+
+  ## Build again but set a new annotation and don't inherit annotations from base image
+  run_buildah build $WITH_POLICY_JSON --inherit-annotations=false --annotation hello=world -t $target --from $base $BUDFILES/base-with-labels
```
nalind: Was `--layers` meant to be used here?

tests/bud.bats
```diff
@@ -141,6 +141,46 @@ _EOF
...
+  ## Second build must use cache
+  run_buildah build $WITH_POLICY_JSON --layers --iidfile ${TEST_SCRATCH_DIR}/iid2 -t $target --from $base $BUDFILES/base-with-labels
+  ## Must use cache
+  expect_output --substring " Using cache"
+  cmp ${TEST_SCRATCH_DIR}/iid1 ${TEST_SCRATCH_DIR}/iid2
+
+  ## Since we are inheriting no labels, this should not use previous image present in the cache.
+  run_buildah build $WITH_POLICY_JSON --layers --inherit-annotations=false -t $target --from $base $BUDFILES/base-with-labels
+  ## Should not contain `Using Cache`
+  assert "$output" !~ "Using cache"
```
nalind, line 217: Nit: inconsistent capitalization

nalind (changes requested):
Questions about the tests, mainly.

flouthoc:
I need some changes from this PR https://github.com/containers/buildah/pull/6195 , so lets wait for https://github.com/containers/buildah/pull/6195

TomSweeneyRedHat:
@flouthoc looks like #6195 has merged, is this one ready to move forward given that?

flouthoc:
@TomSweeneyRedHat Yes this is ready for review. I already addressed @nalind 's comment in recent push.

TomSweeneyRedHat:
LGTM
@nalind?

nalind (changes requested):
I'd expect one of our tests that runs both with and without --layers to use a Dockerfile that has more than one instruction after FROM.  A build with only one instruction after the FROM doesn't give the test the ability to examine what happens in the intermediate images, to check that the annotations are dropped at the point that we want them to be and not somewhere else, or more than once, which could create unnecessary cache misses.

flouthoc:
> I'd expect one of our tests that runs both with and without --layers to use a Dockerfile that has more than one instruction after FROM. A build with only one instruction after the FROM doesn't give the test the ability to examine what happens in the intermediate images, to check that the annotations are dropped at the point that we want them to be and not somewhere else, or more than once, which could create unnecessary cache misses.

@nalind Done, extended both the tests to test with a new `Containerfile` with multiple instructions.

nalind (approved):
LGTM

flouthoc:

TomSweeneyRedHat:
LGTM

PR #6196, opened 2025-06-02
[release-1.40] [CI:DOCS] compare to this release branch

nalind:
#### What type of PR is this?

/kind other

#### What this PR does / why we need it:

Tell Cirrus to compare to the tip of this branch.

#### How to verify it

Visual inspection.

#### Which issue(s) this PR fixes:

None

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #6201, opened 2025-06-03
[release-1.40] run: handle relabeling bind mounts ourselves

### LINKED ISSUE #6071: buildah is not relabeling mounted paths and instead relies on the OCI runtime to do this
It appears as if buildah is not relabeling mounted directories itself and instead relies on the OCI runtime to perform this action, which is not really spec compliant.

To reproduce this issue, use the following `Dockerfile`:
```Dockerfile
from alpine
run --mount=type=bind,source=vendor,target=file,Z ls -al file
```

and run `buildah bud --runtime /usr/bin/crun -f containerfile .` which succeeds, whereas it fails with runc:
```ShellSession
❯ buildah bud --runtime /usr/bin/runc -f containerfile .
STEP 1/2: FROM alpine
STEP 2/2: run --mount=type=bind,source=vendor,target=file,Z ls -al file
error running container: from /usr/bin/runc creating container for [/bin/sh -c ls -al file]: time="2025-03-19T11:45:09+01:00" level=error msg="runc create failed: invalid mount &{Source:/var/tmp/buildah954402691/mnt/buildah-bind-target-1 Destination:/file Device:bind Flags:20481 ClearedFlags:0 PropagationFlags:[262144] Data:Z Relabel: RecAttr:<nil> Extensions:0 IDMapping:<nil>}: bind mounts cannot have any filesystem-specific options applied"
: exit status 1
ERRO[0000] did not get container create message from subprocess: EOF
Error: building at STEP "RUN --mount=type=bind,source=vendor,target=file,Z ls -al file": while running runtime: exit status 1
```
The error here is coming from runc directly: https://github.com/opencontainers/runc/blob/8b98e829f0934fc043f57d73b5c2ee091252a1d2/libcontainer/configs/validate/validator.go#L307

runc throws this warning because it doesn't perform any special handling of the `Z` flag, whereas crun apparently does (however, it probably shouldn't).

The issue in buildah specifically is, that it doesn't appear to perform the same mount flag "sanitation" that podman does, which drops the `z` or `Z` flags and relabels the bind mounted directories: https://github.com/danishprakash/podman/blob/29a90c3f8a1375b585dd3a870f7569ed2599426f/libpod/container_internal_common.go#L282

cc @danishprakash @cyphar

nalind:
#### What type of PR is this?

/kind bug

Handle requested relabeling of bind mounts (i.e., the "z" and "Z" flags) directly, instead of letting the runtime handle the relabeling.

#### How to verify it

We now run integration tests using both crun and runc on Fedora, which exercises their relabeling logic, or the lack thereof.

Fixes #6071.

Manual cherry-pick from #6132, incorporating #6196 because it's one line, and other changes from the Cirrus configuration on the main branch.

```release-note
None
```


nalind:
The centos-stream-10 errors are all of the `go: go.mod requires go >= 1.23.3 (running go 1.23.1; GOTOOLCHAIN=local)` variety.

nalind:
Fine, I'll do it myself: https://github.com/containers/buildah/pull/6511.

Issue #6069, opened 2025-03-17
bash completion error with `shopt -s failglob`


dseomn:
I have [failglob](https://mywiki.wooledge.org/glob#failglob) set in my interactive shells, and buildah's bash completion doesn't seem to work with that. In this output from `bash --norc`, I pressed tab after `buildah `:

```
bash-5.2$ . /etc/bash_completion
bash-5.2$ shopt -s failglob
bash-5.2$ buildah bash: no match: @(--cgroup-manager|--registries-conf|--registries-conf-dir|--root|--runroot|--storage-driver|--storage-opt|--userns-uid-map|--userns-gid-map)
bash: [: 1: unary operator expected
```

nalind:
I'm not seeing this error on my system, but then it (Fedora 41) doesn't have a `/etc/bash_completion` file.  Could it be specific to your package, system, or distribution?

dseomn:
Doesn't look like it, I just tried again with the file from the git repo, which is the same as the file on my system:

```
$ LC_ALL=C.UTF-8 bash --norc
bash-5.2$ git show --oneline
e8d8be71c (HEAD -> main, origin/main, origin/HEAD) Merge pull request #6070 from containers/renovate/github.com-opencontainers-runc-1.x

bash-5.2$ diff -u /usr/share/bash-completion/completions/buildah contrib/completions/bash/buildah
bash-5.2$ . contrib/completions/bash/buildah
bash-5.2$ shopt -s failglob
bash-5.2$ buildah bash: _get_comp_words_by_ref: command not found
bash: [: 1: unary operator expected
bash: no match: @(--cgroup-manager|--registries-conf|--registries-conf-dir|--root|--runroot|--storage-driver|--storage-opt|--userns-uid-map|--userns-gid-map)
```

The `_get_comp_words_by_ref` error is because I didn't source bash completion itself, I think, but the `no match` error is still there. If I source `/usr/share/bash-completion/bash_completion` (which it looks like [Fedora 41 provides?](https://packages.fedoraproject.org/pkgs/bash-completion/bash-completion/fedora-41-updates.html#files)) instead, I get the original error:

```
$ LC_ALL=C.UTF-8 bash --norc
bash-5.2$ . /usr/share/bash-completion/bash_completion 
bash-5.2$ shopt -s failglob
[...]

PR #6207, opened 2025-06-04
Switch to automatically-generated completions

nalind:
#### What type of PR is this?

/kind cleanup

Get out of business of manually crafting and updating shell completion helper functions.

#### How to verify it

I don't use a shell that benefits from this (old habit), so... give it a go?

None

```release-note
Shell completions are now generated rather than manually crafted, so completion suggestions will look different.
```


Makefile
```diff
@@ -139,9 +139,19 @@ uninstall:
...
+contrib/completions/fish/buildah: ./bin/buildah
+	./bin/buildah completion fish > $@
+contrib/completions/powershell/buildah: ./bin/buildah
+	./bin/buildah completion powershell > $@
+contrib/completions/zsh/buildah: ./bin/buildah
+	./bin/buildah completion zsh > $@
 
 .PHONY: install.completions
-install.completions:
+install.completions: contrib/completions/bash/buildah
```
Luap99: this should install for zsh and fish as well, see the podman Makefile for an example

contrib/completions/fish/buildah
Luap99, line 1: In hindsight I think committing the files in podman was a mistake, every cobra update that changes the files will means you need to regenerate so that auto renovate updates won't work.

So in short I don't think we should commit these completion files at all since the build process is generating them anyway,

Luap99 (commented):
I have not tested it yet but I assume we would like to add some autocompletion functions to auto complete container/image names at least for the basic commands.

Otherwise we regress users compared to the existing bash script.

Luap99:
Also I guess this should have a Fixes https://github.com/containers/buildah/issues/6069?

contrib/completions/fish/buildah
nalind: Fair enough.  I'll drop them.

Makefile
nalind: Adding.

Luap99:
Ok I did some local tests I think we might one to port some other changes from podman such as 
https://github.com/containers/podman/blob/badf6b8b1735dff7cdb45636eb8eba490454f9d9/cmd/podman/root.go#L311-L317

```
$ bin/buildah rm -- 
--all                    (remove all containers)
--cgroup-manager         (cgroup manager)
--help                   (help for rm)
--log-level              (the log level to be used, one of "trace", "debug", "info", "warn", "error", "fatal", or "panic")
--registries-conf-dir    (path to registries.conf.d directory (not usually used))
--registries-conf        (path to registries.conf file (not usually used))
--root                   (storage root dir)
--runroot                (storage state dir)
--short-name-alias-conf  (path to short name alias cache file (not usually used))
--storage-driver         (storage-driver)
--storage-opt            (storage driver option)
--userns-gid-map         (default `ctrID:hostID:length` GID mapping to use)
--userns-uid-map         (default `ctrID:hostID:length` UID mapping to use)
```

Because the are persistent flags they can be set technically anywhere but realistically for a user suggesting them on each command makes no sense.
Second as I though just doing `buildah rmi` no longer completes the image ids/names which is a clear UX regression so I rather see that implemented before we switch the script.

I am happy to take that over if you don't want to spend the time understanding all the small details in play here, its something I can do while being offline on the train next week.

nalind:
> I am happy to take that over if you don't want to spend the time understanding all the small details in play here, its something I can do while being offline on the train next week.

If you want to, sure, otherwise I could stand to get more familiar with those bits.

Luap99:
> If you want to, sure, otherwise I could stand to get more familiar with those bits.

I find other stuff to do, if you want to have some deeper understanding please do that.

https://github.com/containers/podman/blob/badf6b8b1735dff7cdb45636eb8eba490454f9d9/cmd/podman/root.go#L302-L321
This must be done in persistentPreRunE() function that is registered on the root command here in buildah in a similar way, I see there is already a before() function where it could be added. First it is used to hide the top level flags on subcommand so they don't show up all the time. The `cmd.Root().Traverse(os.Args[2:])` call is not right in the context of buildah I think, podman has some really special cli setup where this hack was needed but I don't know if it is  needed for buildah. `Traverse()` is definly wrong because that must only be used when TraverseChildren is set to true which it is not for buildah so it would need to be `Find()` if it is needed at all.
I guess the easier way to check if the command being pass in is the root buildah command is to just compare the pointer of cmd against rootCmd.

For actual specific completion suggestions there is https://github.com/containers/podman/blob/badf6b8b1735dff7cdb45636eb8eba490454f9d9/cmd/podman/common/completion.go with a ton of example functions.

Than you just have to register them to the command via ValidArgsFunction or in case of flags RegisterFlagCompletionFunc(),

Luap99:
And for testing with bash `source <(bin/buildah completion bash)` will work.

The nice thing about the binary is generating the completions is if you build buildah from source and let's say added a new command then it will work right away without having to reload the script, `bin/buildah [TAB]` would show it while `/usr/bin/buildah [TAB]` will not since that binary does have the command yet.

PR #6208, opened 2025-06-04
build: add --source-date-epoch and --rewrite-timestamp flags

nalind:
#### What type of PR is this?

/kind feature

Use $SOURCE_DATE_EPOCH as the default for the `--source-date-epoch` flag to the "build" CLI.

When a source-date-epoch is set, we'll use it when writing new history entries, force timestamps in data written for `--output` to the specified timestamp, and populate a "SOURCE_DATE_EPOCH" ARG that we treat as always being set, and which we don't complain about being left unused. By default, this will not affect timestamps in newly-added layers.

Add a `--rewrite-timestamp` flag, which "clamps" timestamps in newly-added layers to not be later than the `--source-date-epoch` value if the `--source-date-epoch` flag is set, but has no effect otherwise.

Fix a bug in the `dir_image_last_diff()` helper function introduced in a57e7f4b2491d064de0151d682b590a929370a0f as part of #6178.

#### How to verify it

New integration tests!

None

```release-note
`buildah build` now recognizes `--source-date-epoch` and `--rewrite-timestamp` flags, which affect the dates recorded in the new image's configuration and the timestamps on the contents of the new layers and content generated in response to the `--output` flag.
```

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -21,6 +21,7 @@ import (
 	"github.com/containerd/platforms"
 	"github.com/containers/buildah"
 	"github.com/containers/buildah/define"
+	"github.com/containers/buildah/internal"
 	internalUtil "github.com/containers/buildah/internal/util"
 	"github.com/containers/buildah/pkg/parse"
 	"github.com/containers/buildah/util"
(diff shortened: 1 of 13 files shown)


imagebuildah/build.go
```diff
@@ -259,6 +260,16 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 		}
 		// Deep copy args to prevent concurrent read/writes over Args.
 		platformOptions.Args = maps.Clone(options.Args)
+
+		if options.SourceDateEpoch != nil {
+			if options.Timestamp != nil {
+				return "", nil, errors.New("timestamp and source-date-epoch would be ambiguous if allowed together")
```
Honny1, line 266: This could be mentioned in the manpage.

nalind: Added, thanks.

giuseppe (approved):
/lgtm

PR #6211, opened 2025-06-06
build: --source-date-epoch/--timestamp use static hostname/cid

nalind:
#### What type of PR is this?

/kind other

When using either `--source-date-epoch` or `--timestamp`, make sure that handling of RUN instructions uses a defined hostname if possible, and commits using a reference to a static container name.

#### How to verify it

New integration test!

None

```release-note
`buildah build` sets a fixed hostname for RUN instructions when invoked with `--timestamp` or `--source-date-epoch`.
When building or committing images with the `--timestamp` or `--source-date-epoch` flags, the hostname, domain name, and name-of-container-which-was-committed will be set to fixed values.
```

### DIFF
--- a/image.go
+++ b/image.go
@@ -308,12 +308,21 @@ func (i *containerImageRef) newDockerSchema2ManifestBuilder() (manifestBuilder, 
 	if err := json.Unmarshal(i.dconfig, &dimage); err != nil {
 		return nil, err
 	}
+	// Suppress the hostname and domainname if we're running with the
+	// equivalent of either --timestamp or --source-date-epoch.
+	if i.created != nil {
+		dimage.Config.Hostname = "sandbox"
+		dimage.Config.Domainname = ""
+	}
 	// Set the parent, but only if we want to be compatible with "classic" docker build.
 	if i.compatSetParent == types.OptionalBoolTrue {
 		dimage.Parent = docker.ID(i.parent)
 	}
 	// Set the container ID and containerConfig in the docker format.
 	dimage.Container = i.containerID
+	if i.created != nil {
+		dimage.Container = ""
+	}
 	if dimage.Config != nil {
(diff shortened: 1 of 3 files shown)


nalind:
Wasn't testing the release note about RUN, added that.

image.go
```diff
@@ -308,12 +308,21 @@ func (i *containerImageRef) newDockerSchema2ManifestBuilder() (manifestBuilder,
...
+	if i.created != nil {
+		dimage.Config.Hostname = "sandbox"
+		dimage.Config.Domainname = ""
+	}
 	// Set the parent, but only if we want to be compatible with "classic" docker build.
 	if i.compatSetParent == types.OptionalBoolTrue {
 		dimage.Parent = docker.ID(i.parent)
 	}
 	// Set the container ID and containerConfig in the docker format.
 	dimage.Container = i.containerID
```
mtrmac, line 322: (Do we want to expose value this at all? Formally it’s an information leak, but I suppose it shouldn’t hurt: by the time the image is published, the container should not exist. Anyway, that’s probably a separate PR.)

nalind: I hadn't really thought about that, but yes, we could generate another pseudorandom value instead of using the container's ID.

mtrmac: FWIW Docker has deprecated the field and is not returning it to recent callers in “inspect”; but, at least inspecting the immediate few layers of the call stack, seems to be still setting it in new images.

nalind: I'm seeing it set occasionally, in what I'm assuming is a `github.com/moby/docker-image-spec/specs-go/v1.DockerOCIImage`.  "docker://docker.io/library/kong:latest" is a convenient example.

flouthoc (approved):
LGTM
/lgtm
/approve

PR #6213, opened 2025-06-08
fix: add newline to digestfile

d3spairx:
#### What type of PR is this?

/kind other

This adds newline to `buildah push --digestfile` output file. 

```release-note

```


nalind:
I always imagined `--digestfile` as a complement to `--iidfile`, and we don't include a newline at the end of those because `docker build` doesn't.  What's the intent behind making this behave differently?

agrshv:
@nalind when we open digestfile in the terminal, we have broken terminal output because of the lack of newline. That is the whole intention of this PR.

TomSweeneyRedHat:
@d3spairx tests need help, and it would probably be good to rebase this.

PR #6214, opened 2025-06-10
commit,build: --source-date-epoch/--timestamp omit identity label

nalind:
#### What type of PR is this?

/kind bug

When using either --source-date-epoch or --timestamp, default to not adding a label with our version number in it, since it can change between builds.

#### How to verify it

New integration test!

None

```release-note
`buildah build` no longer sets/updates the `io.buildah.version` label by default when either `--timestamp` or `--source-date-epoch` are used.
```

### DIFF
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -255,11 +255,6 @@ func commitCmd(c *cobra.Command, args []string, iopts commitInputOptions) error 
 		}
 	}
 
-	// Add builder identity information.
-	if iopts.identityLabel {
-		builder.SetLabel(buildah.BuilderIdentityAnnotation, define.Version)
-	}
-
 	encConfig, encLayers, err := cli.EncryptConfig(iopts.encryptionKeys, iopts.encryptLayers)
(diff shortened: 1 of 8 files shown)


flouthoc (approved):
Although above question does not makes a lot of sense because users who are going to be using these two flags are certain about their use-case ?

So PR LGTM @containers/buildah-maintainers PTAL

TomSweeneyRedHat:
LGTM
Looks like the branch need some updating @nalind

nalind:
> Following PR assumes that if both flags are used the use-case is `reproduce-able` builds and it removes the identity label and documentation doesn't say so. Maybe docs should say why is it removing identify label for such use-case.

I don't think I follow — if `--identity-label` is set to `false`, the label is neither set nor cleared, so if the base image had a value set for it, it'll be carried over.  I was tempted to have it clear the value for `--identity-label=false` cases when implementing this, but that would have been a bigger change from the previous behavior, so I limited the scope to adding the condition to the not-specified-either-way case.  Adding a test that confirms that.

PR #6216, opened 2025-06-10
 dynamically link sqlite

Luap99:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

Dynamically link sqlite3 when installed, the main motivation is that we
reduce the podman binary size with that. I see about 3.2 MB savings.

But also dynamically linking it means if there a vulnerabilities only
the sqlite3 distro package needs updating and we don't have to make a
new podman release with the vendored update.

```release-note
When building buildah via Makefile we try to dynamically link sqlite3 if the library/header is installed locally.
```


Luap99:
cc @lsm5 @jnovy

TomSweeneyRedHat:
@Luap99 branch is asking for an update.
LGTM
otherwise

jnovy:
Yes please - LGTM

PR #6217, opened 2025-06-10
Builder.sbomScan(): don't break non-root scanners

nalind:
#### What type of PR is this?

/kind bug

Set up permissions on the scanner output directory so that scanners whose images specify that they be run as non-root users can still write to it.  The most recent ghcr.io/anchore/syft image exposed our bug.

#### How to verify it

SBOM-scanning tests should succeed again.

None

```release-note
None
```

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -52,6 +52,13 @@ func (b *Builder) sbomScan(ctx context.Context, options CommitOptions) (imageFil
 			}
 		}
 	}()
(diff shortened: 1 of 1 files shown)


nalind:
Ugh, this bug's been here since 1.35.

flouthoc (approved):
LGTM
/lgtm
/approve
/hold

Looks like CI is blocked without this patch.

PR #6225, opened 2025-06-11
[release-1.35] Builder.sbomScan(): don't break non-root scanners

nalind:
This is a manual cherry-pick of https://github.com/containers/buildah/pull/6223, with the magic added to the commit message to bypass the "you didn't add tests!" requirement.

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -52,6 +52,13 @@ func (b *Builder) sbomScan(ctx context.Context, options CommitOptions) (imageFil
 			}
 		}
 	}()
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM
but I'm just wondering why 1.35?  I don't think that's supported anywhere now is it?

nalind:
The bug goes back to 1.35 and our older branches are not exclusively used by RHEL.

flouthoc (approved):
LGTM
/lgtm
/approve

PR #6226, opened 2025-06-12
bud,run: runc does not support keep-groups

ricardobranco777:
#### What type of PR is this?

> /kind failing-test 

#### What this PR does / why we need it:

Skip the final part of 2 tests that check `keep-groups` when the OCI runtime is `runc` as it doesn't support `keep-groups`.

#### How to verify it

https://openqa.opensuse.org/tests/5101660

TAP files for the above job:
- https://openqa.opensuse.org/tests/5101660/file/buildah-buildah-root.tap
- https://openqa.opensuse.org/tests/5101660/file/buildah-buildah-user.tap
- Commands run: https://openqa.opensuse.org/tests/5101660/logfile?filename=buildah-commands.txt

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

Currently the tests fail with:

```
#not ok 454 build --group-add in 2902ms
# (from function `assert' in file tests/helpers.bash, line 537,
#  from function `expect_output' in file tests/helpers.bash, line 564,
#  in test file tests/bud.bats, line 7256)
#   `expect_output --substring "65534"' failed
# /var/tmp/buildah/tests /var/tmp/buildah
# # [checking for: docker.io/library/alpine]
# # [restoring from cache: /var/tmp/test.RdTF7C/bats-run-j6JS9l/suite/buildah-image-cache / docker.io/library/alpine]
# Getting image source signatures
# Copying blob sha256:9d16cba9fb961d1aafec9542f2bf7cb64acfc55245f9e4eb5abecd4cdc38d749
# Copying config sha256:961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4
# Writing manifest to image destination
# $ /usr/bin/buildah build --group-add 16335 --signature-policy /var/tmp/buildah/tests/policy.json /var/tmp/test.rdtf7c/buildah_tests.trirew
# STEP 1/2: FROM alpine
# STEP 2/2: RUN id -G
# 0 1 2 3 4 6 10 11 20 26 27 16335
# COMMIT
# Getting image source signatures
# Copying blob sha256:03901b4a2ea88eeaad62dbe59b072b28b6efa00491962b8741081c5df50c65e0
# Copying blob sha256:7619c5173ae54a9e8cb7882d6005beb3967d4e099c1d6787205c853b90cb8ecb
# Copying config sha256:dfe52373489c34414a7e5bea6b450d2fc7a77ef2f3507016e4c6d16292a80ce4
# Writing manifest to image destination
[...]

PR #6227, opened 2025-06-12
CI: give the rootless test user some supplemental groups

nalind:
#### What type of PR is this?

/kind other

Exercise preservation of supplemental groups in the tests that use `buildah build` and `buildah from` with `--group-add keep-groups`.

#### How to verify it

This should exercise more of the tests we have.

None

```release-note
None
```


TomSweeneyRedHat:
@nalind does a test need tweaking?

```
 ok 822 run-exit-status on non executable
[+0647s] not ok 823 Verify /run/.containerenv exist
[+0647s] # (from function `assert' in file ./helpers.bash, line 537,
[+0647s] #  from function `expect_output' in file ./helpers.bash, line 564,
[+0647s] #  in test file ./run.bats, line 621)
[+0647s] #   `expect_output "alpine-working-container"' failed
[+0647s] # /var/tmp/go/src/github.com/containers/buildah/tests /var/tmp/go/src/github.com/containers/buildah/tests
[+0647s] # # [checking for: docker.io/library/alpine]
[+0647s] # # [restoring from cache: /tmp/bats-run-CvOu1h/suite/buildah-image-cache / docker.io/library/alpine]
[+0647s] # Getting image source signatures
[+0647s] # Copying blob sha256:9d16cba9fb961d1aafec9542f2bf7cb64acfc55245f9e4eb5abecd4cdc38d749
[+0647s] # Copying config sha256:961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4
[+0647s] # Writing manifest to image destination
[+0647s] # # /var/tmp/go/src/github.com/containers/buildah/tests/./../bin/buildah from --quiet --pull=false --signature-policy /var/tmp/go/src/github.com/containers/buildah/tests/./policy.json alpine
[+0647s] # alpine-working-container
[+0647s] # # /var/tmp/go/src/github.com/containers/buildah/tests/./../bin/buildah run alpine-working-container ls -1 /run/.containerenv
[+0647s] # /run/.containerenv
[+0647s] # # /var/tmp/go/src/github.com/containers/buildah/tests/./../bin/buildah run alpine-working-container sh -c . /run/.containerenv; echo $engine
[+0647s] # buildah-1.41.0-dev
[+0647s] # # /var/tmp/go/src/github.com/containers/buildah/tests/./../bin/buildah run alpine-working-container sh -c . /run/.containerenv; echo $name
[+0647s] # alpine-working-container
[+0647s] # time="2025-06-13T09:47:28-05:00" level=error msg="seek /sys/fs/cgroup/system.slice/runc-buildah-buildah1965336133.scope/cgroup.freeze: no such device"
[+0647s] # #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
[+0647s] # #|     FAIL: buildah run alpine-working-container sh -c . /run/.containerenv; echo $name
[+0647s] # #| expected: 'alpine-working-container'
[+0647s] # #|   actual: 'alpine-working-container'
[+0647s] # #|         > 'time="2025-06-13T09:47:28-05:00" level=error msg="seek /sys/fs/cgroup/system.slice/runc-buildah-buildah1965336133.scope/cgroup.freeze: no such device"'
[+0647s] # #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

TomSweeneyRedHat:
LGTM
A test is tweaking out, I'm not sure that it's a flake, but I'm not sure it isn't.

nalind:
That's a flake we've been seeing since we started testing with runc.  The thing that confuses me is that this should be helping us trigger the problem that #6226 aims to fix, but either I'm doing it wrong, or the test isn't noticing the problem.

TomSweeneyRedHat:
@nalind looks like an update is needed.
@flouthoc PTAL and Merge if it LGTY

flouthoc (approved):
LGTM
/lgtm
/approve

contrib/cirrus/lib.sh
```diff
@@ -298,9 +298,17 @@ setup_rootless() {
     # Guarantee independence from specific values
     rootless_uid=$[RANDOM+1000]
     rootless_gid=$[RANDOM+1000]
-    msg "creating $rootless_uid:$rootless_gid $ROOTLESS_USER user"
+    rootless_supplemental_gid1=$[RANDOM+1000]
+    rootless_supplemental_gid2=$[RANDOM+1000]
+    rootless_supplemental_gid3=$[RANDOM+1000]
```
Luap99, line 303: Isn't this just introducing a setup flake? If we randomly get the same gid again it would not add extra gids, I just tested useradd and it doesn't error in this case.

However the groupadd calls will fails since the group already exists. It is not likely to happen but I think you must make sure uses a different uid range for each group to avoid the conflict.

And the fact that this does not trigger https://github.com/containers/buildah/pull/6226 is concerning no? It means we are not testing what we think we are testing with this. The fact that the tesst does not log a skip message makes it impossible to tell if the "--group-add keep-groups" check was run or not because it is inside a if that does not use skip

nalind: Hmm, not catching duplicates would introduce a flake, that's worth fixing.  Maybe logging what's going on when the tests aren't triggered will shed a clue.

PR #6228, opened 2025-06-16
CI: ensure rootless groups aren't duplicates

nalind:
#### What type of PR is this?

/kind flake

When adding an unprivileged user to run tests as, ensure that the primary and supplemental groups are not repeated, and that at least one supplemental group is outside of the user's subgid range.  When skipping a test because the invoking user doesn't have any supplemental groups, log the user's IDs, based on conversation in https://github.com/containers/buildah/pull/6227#discussion_r2150504346.

#### How to verify it

Updated tests and CI configuration!

Tests which expect, but don't see, a supplemental groups list, will show up as skipped rather than succeeding.

```release-note
None
```


flouthoc (approved):
LGTM

tests/helpers.bash
```diff
@@ -11,7 +11,7 @@ INET_BINARY=${INET_BINARY:-$TEST_SOURCES/../bin/inet}
 DUMPSPEC_BINARY=${DUMPSPEC_BINARY:-$TEST_SOURCES/../bin/dumpspec}
 STORAGE_DRIVER=${STORAGE_DRIVER:-vfs}
 PATH=$(dirname ${BASH_SOURCE})/../bin:${PATH}
-OCI=${CI_DESIRED_RUNTIME:-$(${BUILDAH_BINARY} info --format '{{.host.OCIRuntime}}' || command -v runc || command -v crun)}
+OCI=${CI_DESIRED_RUNTIME:-${BUILDAH_RUNTIME:-$(${BUILDAH_BINARY} info --format '{{.host.OCIRuntime}}' || command -v runc || command -v crun)}}
```
Luap99: CI_DESIRED_RUNTIME doesn't seem to be used so should we drop it to simplify this a bit?

Luap99 (commented):
Thanks
I feel like this would be best as two commits, once for the missing BUILDAH_RUNTIME passthrough and the other one for the group id fix.

With https://github.com/containers/buildah/pull/6226 merged I guess this will now pass after rebasing so LGTM, if you prefer a single commit I am find merging it liek this here as well.

tests/helpers.bash
nalind: Might as well, yeah.

TomSweeneyRedHat:
LGTM
@flouthoc can your hold be released?

flouthoc:
Yes I was waiting for the CI
/hold cancel

PR #6230, opened 2025-06-17
"root fs only mounted once" test: accept root with only the rw option

nalind:
#### What type of PR is this?

/kind other

When checking /proc/self/mountinfo for a root filesystem, also match "/" mounted with "rw" as its only mount option, as an alternative to being mounted with "rw" and other options at the same time.

#### How to verify it

The test should fail less often if run under sudo on... my laptop.  Maybe your computer, too.

None

```release-note
None
```


TomSweeneyRedHat:
LGTM
Assuming happy tests

PR #6231, opened 2025-06-17
commit-with-extra-files test: use $TEST_SCRATCH_DIR

nalind:
#### What type of PR is this?

/kind other

Use $TEST_SCRATCH_DIR instead of $BATS_TMPDIR for temporary files that we create during one of our integration tests.

#### How to verify it

The test shouldn't be broken by this change.

None

```release-note
None
```


TomSweeneyRedHat:
LGTM
assuming happy tests

PR #6239, opened 2025-06-20
build, commit: set the OCI ...created annotation on OCI images

nalind:
#### What type of PR is this?

/kind feature

When building or committing an image in OCI format, default to setting the "org.opencontainers.image.created" annotation to the value used in the image's config blob for the image's creation date. The behavior can be controlled using the new `--created-annotation` flag.
    
Add `--annotation` and `--unsetannotation` flags to `buildah commit` which mimic the same flags for `buildah build`.

#### How to verify it

New integration test!

None

```release-note
Images committed in OCI formats now default to including an "org.opencontainers.image.created" annotation.
```

### DIFF
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -66,6 +66,9 @@ type commitInputOptions struct {
 	encryptLayers      []int
 	unsetenvs          []string
 	addFile            []string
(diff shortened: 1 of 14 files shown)


TomSweeneyRedHat:
LGTM

PR #6240, opened 2025-06-23
feat: ADD/COPY --link support for buildah

### LINKED ISSUE #4325: [RFE] Add support for `--link` in `COPY/ADD`
**Description**

Buildkit added support for the `--link` flag. Quoting from the [docs](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/reference.md#copy---link):

>Enabling this flag in COPY or ADD commands allows you to copy files with enhanced semantics where your files remain independent on their own layer and don't get invalidated when commands on previous layers are changed.

>When --link is used your source files are copied into an empty destination directory. That directory is turned into a layer that is linked on top of your previous state.

This is a feature request to add this feature in buildah 🙂

2004joshua:
#### What type of PR is this?
/kind feature

This patch implements the --link functionality for COPY and ADD instructions, enabling the creation of
cachable layers that can be reused independently across builds.

When specified, content is staged separately and archived into an independent layer with computed digest,
allowing cache reuse even when underlying base layers change.

Follows Buildkit --link specifications while maintaining compatibility with existing COPY/ADD behavior.

#### How to verify it

```bash
bats tests/bud.bats
```

Fixes #4325 

```release-note
Yes, COPY --link and ADD --link instructions now fully support layered build caching.
```

### DIFF
--- a/add.go
+++ b/add.go
@@ -33,6 +33,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 	"github.com/moby/sys/userns"
 	digest "github.com/opencontainers/go-digest"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/sirupsen/logrus"
 )
@@ -102,6 +103,15 @@ type AddAndCopyOptions struct {
 	Parents bool
 	// Timestamp is a timestamp to override on all content as it is being read.
 	Timestamp *time.Time
+	// Link, when set to true, creates an independent layer containing the copied content
+	// that sits on top of existing layers. This layer can be cached and reused
+	// separately, and is not affected by filesystem changes from previous instructions.
+	Link bool
+	// BuildMetadata is consulted only when Link is true. Contains metadata used by
+	// imagebuildah for cache evaluation of linked layers (inheritLabels, unsetAnnotations,
+	// inheritAnnotations, newAnnotations). This field is internally managed and should
+	// not be set by external API users.
+	BuildMetadata string
 }
 
 // gitURLFragmentSuffix matches fragments to use as Git reference and build
@@ -495,15 +505,75 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 	}
 	destUIDMap, destGIDMap := convertRuntimeIDMaps(b.IDMappingOptions.UIDMap, b.IDMappingOptions.GIDMap)
 
-	// Create the target directory if it doesn't exist yet.
+	var putRoot, putDir, stagingDir string
+	var createdDirs []string
+	var latestTimestamp time.Time
+
 	mkdirOptions := copier.MkdirOptions{
 		UIDMap:   destUIDMap,
 		GIDMap:   destGIDMap,
 		ChownNew: chownDirs,
 	}
-	if err := copier.Mkdir(mountPoint, extractDirectory, mkdirOptions); err != nil {
-		return fmt.Errorf("ensuring target directory exists: %w", err)
+
+	// If --link is specified, we create a staging directory to hold the content
+	// that will then become an independent layer
+	if options.Link {
+		containerDir, err := b.store.ContainerDirectory(b.ContainerID)
+		if err != nil {
+			return fmt.Errorf("getting container directory for %q: %w", b.ContainerID, err)
+		}
+
+		stagingDir, err = os.MkdirTemp(containerDir, "link-stage-")
+		if err != nil {
+			return fmt.Errorf("creating staging directory for link %q: %w", b.ContainerID, err)
+		}
+
+		putRoot = stagingDir
+
+		cleanDest := filepath.Clean(destination)
+
+		if strings.Contains(cleanDest, "..") {
+			return fmt.Errorf("invalid destination path %q: contains path traversal", destination)
+		}
+
+		if renameTarget != "" {
+			putDir = filepath.Dir(filepath.Join(stagingDir, cleanDest))
+		} else {
+			putDir = filepath.Join(stagingDir, cleanDest)
+		}
+
+		putDirAbs, err := filepath.Abs(putDir)
+		if err != nil {
+			return fmt.Errorf("failed to resolve absolute path: %w", err)
+		}
+
+		stagingDirAbs, err := filepath.Abs(stagingDir)
+		if err != nil {
+			return fmt.Errorf("failed to resolve staging directory absolute path: %w", err)
+		}
+
+		if !strings.HasPrefix(putDirAbs, stagingDirAbs+string(os.PathSeparator)) && putDirAbs != stagingDirAbs {
+			return fmt.Errorf("destination path %q escapes staging directory", destination)
+		}
+		if err := copier.Mkdir(putRoot, putDirAbs, mkdirOptions); err != nil {
+			return fmt.Errorf("ensuring target directory exists: %w", err)
+		}
+		tempPath := putDir
+		for tempPath != stagingDir && tempPath != filepath.Dir(tempPath) {
+			if _, err := os.Stat(tempPath); err == nil {
+				createdDirs = append(createdDirs, tempPath)
+			}
+			tempPath = filepath.Dir(tempPath)
+		}
+	} else {
+		if err := copier.Mkdir(mountPoint, extractDirectory, mkdirOptions); err != nil {
+			return fmt.Errorf("ensuring target directory exists: %w", err)
+		}
+
+		putRoot = extractDirectory
+		putDir = extractDirectory
 	}
+
 	// Copy each source in turn.
 	for _, src := range sources {
 		var multiErr *multierror.Error
@@ -580,7 +650,7 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 						ChmodFiles:    nil,
 						IgnoreDevices: userns.RunningInUserNS(),
 					}
-					putErr = copier.Put(extractDirectory, extractDirectory, putOptions, io.TeeReader(pipeReader, hasher))
+					putErr = copier.Put(putRoot, putDir, putOptions, io.TeeReader(pipeReader, hasher))
 				}
 				hashCloser.Close()
 				pipeReader.Close()
@@ -658,6 +728,9 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 				itemsCopied++
 			}
 			st := localSourceStat.Results[globbed]
+			if options.Link && st.ModTime.After(latestTimestamp) {
+				latestTimestamp = st.ModTime
+			}
 			pipeReader, pipeWriter := io.Pipe()
 			wg.Add(1)
 			go func() {
@@ -741,12 +814,13 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 						ChmodFiles:      nil,
 						IgnoreDevices:   userns.RunningInUserNS(),
 					}
-					putErr = copier.Put(extractDirectory, extractDirectory, putOptions, io.TeeReader(pipeReader, hasher))
+					putErr = copier.Put(putRoot, putDir, putOptions, io.TeeReader(pipeReader, hasher))
 				}
 				hashCloser.Close()
 				pipeReader.Close()
 				wg.Done()
 			}()
+
 			wg.Wait()
 			if getErr != nil {
 				getErr = fmt.Errorf("reading %q: %w", src, getErr)
@@ -776,6 +850,58 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 			return fmt.Errorf("no items matching glob %q copied (%d filtered out%s): %w", localSourceStat.Glob, len(localSourceStat.Globbed), excludesFile, syscall.ENOENT)
 		}
 	}
+
+	if options.Link {
+		if !latestTimestamp.IsZero() {
+			for _, dir := range createdDirs {
+				if err := os.Chtimes(dir, latestTimestamp, latestTimestamp); err != nil {
+					logrus.Warnf("failed to set timestamp on directory %q: %v", dir, err)
+				}
+			}
+		}
+		var created time.Time
+		if options.Timestamp != nil {
+			created = *options.Timestamp
+		} else if !latestTimestamp.IsZero() {
+			created = latestTimestamp
+		} else {
+			created = time.Unix(0, 0).UTC()
+		}
+
+		command := "ADD"
+		if !extract {
+			command = "COPY"
+		}
+
+		contentType, digest := b.ContentDigester.Digest()
+		summary := contentType
+		if digest != "" {
+			if summary != "" {
+				summary = summary + ":"
+			}
+			summary = summary + digest.Encoded()
+			logrus.Debugf("added content from --link %s", summary)
+		}
+
+		createdBy := "/bin/sh -c #(nop) " + command + " --link " + summary + " in " + destination + " " + options.BuildMetadata
+		history := v1.History{
+			Created:   &created,
+			CreatedBy: createdBy,
+			Comment:   b.HistoryComment(),
+		}
+
+		linkedLayer := LinkedLayer{
+			History:  history,
+			BlobPath: stagingDir,
+		}
+
+		b.AppendedLinkedLayers = append(b.AppendedLinkedLayers, linkedLayer)
+
+		if err := b.Save(); err != nil {
+			return fmt.Errorf("saving builder state after queuing linked layer: %w", err)
+		}
+	}
+
 	return nil
 }
 
--- a/cmd/buildah/addcopy.go
+++ b/cmd/buildah/addcopy.go
@@ -41,6 +41,7 @@ type addCopyResults struct {
 	excludes         []string
 	parents          bool
 	timestamp        string
+	link             bool
 }
 
 func createCommand(addCopy string, desc string, short string, opts *addCopyResults) *cobra.Command {
@@ -74,6 +75,7 @@ func applyFlagVars(flags *pflag.FlagSet, opts *addCopyResults) {
 	flags.StringVar(&opts.chown, "chown", "", "set the user and group ownership of the destination content")
 	flags.StringVar(&opts.chmod, "chmod", "", "set the access permissions of the destination content")
 	flags.StringVar(&opts.creds, "creds", "", "use `[username[:password]]` for accessing registries when pulling images")
+	flags.BoolVar(&opts.link, "link", false, "enable layer caching for this operation (creates an independent layer)")
 	if err := flags.MarkHidden("creds"); err != nil {
 		panic(fmt.Sprintf("error marking creds as hidden: %v", err))
 	}
@@ -263,6 +265,7 @@ func addAndCopyCmd(c *cobra.Command, args []string, verb string, iopts addCopyRe
 		MaxRetries:            iopts.retry,
 		Parents:               iopts.parents,
 		Timestamp:             timestamp,
+		Link:                  iopts.link,
 	}
 	if iopts.contextdir != "" {
 		var excludes []string
--- a/docs/buildah-add.1.md
+++ b/docs/buildah-add.1.md
@@ -65,6 +65,13 @@ can be used.
 
 Path to an alternative .containerignore (.dockerignore) file. Requires \-\-contextdir be specified.
 
+**--link**
+
+Create an independent image layer for the added files instead of modifying the working
+container's filesystem. If `buildah run` creates a file and `buildah add --link` adds a file
+to the same path, the file from `buildah add --link` will be present in the committed image.
+The --link layer is applied after all container filesystem changes at commit time.
+
 **--quiet**, **-q**
 
 Refrain from printing a digest of the added content.
--- a/docs/buildah-copy.1.md
+++ b/docs/buildah-copy.1.md
@@ -65,6 +65,13 @@ is preserved.
 
 Path to an alternative .containerignore (.dockerignore) file. Requires \-\-contextdir be specified.
 
+**--link**
+
+Create an independent image layer for the added files instead of modifying the working
+container's filesystem. If `buildah run` creates a file and `buildah copy --link` adds a file
+to the same path, the file from `buildah copy --link` will be present in the committed image.
+The --link layer is applied after all container filesystem changes at commit time.
+
 **--parents**
 
 Preserve leading directories in the paths of items being copied, relative to either the
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -76,6 +76,8 @@ type StageExecutor struct {
 	stage                 *imagebuilder.Stage
 	didExecute            bool
 	argsFromContainerfile []string
+	hasLink               bool
+	isLastStep            bool
 }
 
 // Preserve informs the stage executor that from this point on, it needs to
@@ -359,8 +361,11 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
 			}
 			return errors.New("COPY --keep-git-dir is not supported")
 		}
-		if cp.Link {
-			return errors.New("COPY --link is not supported")
+		if cp.Link && s.executor.layers {
+			s.hasLink = true
+		} else if cp.Link {
+			s.executor.logger.Warn("--link is not supported when building without --layers, ignoring --link")
+			s.hasLink = false
 		}
 		if len(cp.Excludes) > 0 {
 			excludes = append(slices.Clone(excludes), cp.Excludes...)
@@ -564,6 +569,7 @@ func (s *StageExecutor) performCopy(excludes []string, copies ...imagebuilder.Co
 				sources = append(sources, src)
 			}
 		}
+		labelsAndAnnotations := s.buildMetadata(s.isLastStep, true)
 		options := buildah.AddAndCopyOptions{
 			Chmod:             copy.Chmod,
 			Chown:             copy.Chown,
@@ -583,6 +589,8 @@ func (s *StageExecutor) performCopy(excludes []string, copies ...imagebuilder.Co
 			MaxRetries:            s.executor.maxPullPushRetries,
 			RetryDelay:            s.executor.retryPullPushDelay,
 			Parents:               copy.Parents,
+			Link:                  s.hasLink,
+			BuildMetadata:         labelsAndAnnotations,
 		}
 		if len(copy.Files) > 0 {
 			// If we are copying heredoc files, we need to temporary place
@@ -1330,6 +1338,8 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 		logRusage()
 		moreInstructions := i < len(children)-1
 		lastInstruction := !moreInstructions
+
+		s.isLastStep = lastStage && lastInstruction
 		// Resolve any arguments in this instruction.
 		step := ib.Step()
 		if err := step.Resolve(node); err != nil {
@@ -1795,6 +1805,8 @@ func (s *StageExecutor) Execute(ctx context.Context, base string) (imgID string,
 				return "", nil, false, fmt.Errorf("preparing container for next step: %w", err)
 			}
 		}
+
+		s.hasLink = false
 	}
 
 	return imgID, ref, onlyBaseImage, nil
@@ -1888,30 +1900,13 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
 	if node == nil {
 		return "/bin/sh", nil
 	}
-	inheritLabels := ""
-	unsetAnnotations := ""
-	inheritAnnotations := ""
-	newAnnotations := ""
-	// If --inherit-label was manually set to false then update history.
-	if s.executor.inheritLabels == types.OptionalBoolFalse {
-		inheritLabels = "|inheritLabels=false"
-	}
-	if isLastStep {
-		for _, annotation := range s.executor.unsetAnnotations {
-			unsetAnnotations += "|unsetAnnotation=" + annotation
-		}
-		// If --inherit-annotation was manually set to false then update history.
-		if s.executor.inheritAnnotations == types.OptionalBoolFalse {
-			inheritAnnotations = "|inheritAnnotations=false"
-		}
-		// If new annotations are added, they must be added as part of the last step of the build,
-		// so mention in history that new annotations were added inorder to make sure the builds
-		// can either reuse layers or burst the cache depending upon new annotations.
-		if len(s.executor.annotations) > 0 {
-			newAnnotations += strings.Join(s.executor.annotations, ",")
-		}
-	}
-	switch strings.ToUpper(node.Value) {
+
+	command := strings.ToUpper(node.Value)
+	addcopy := command == "ADD" || command == "COPY"
+
+	labelsAndAnnotations := s.buildMetadata(isLastStep, addcopy)
+
+	switch command {
 	case "ARG":
 		for _, variable := range strings.Fields(node.Original) {
 			if variable != "ARG" {
@@ -1919,7 +1914,7 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
 			}
 		}
 		buildArgs := s.getBuildArgsKey()
-		return "/bin/sh -c #(nop) ARG " + buildArgs + inheritLabels + unsetAnnotations + inheritAnnotations + newAnnotations, nil
+		return "/bin/sh -c #(nop) ARG " + buildArgs + labelsAndAnnotations, nil
 	case "RUN":
 		shArg := ""
 		buildArgs := s.getBuildArgsResolvedForRun()
@@ -1999,16 +1994,20 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
 		if buildArgs != "" {
 			result = result + "|" + strconv.Itoa(len(strings.Split(buildArgs, " "))) + " " + buildArgs + " "
 		}
-		result = result + "/bin/sh -c " + shArg + heredoc + appendCheckSum + inheritLabels + unsetAnnotations + inheritAnnotations + newAnnotations
+		result = result + "/bin/sh -c " + shArg + heredoc + appendCheckSum + labelsAndAnnotations
 		return result, nil
 	case "ADD", "COPY":
 		destination := node
 		for destination.Next != nil {
 			destination = destination.Next
 		}
-		return "/bin/sh -c #(nop) " + strings.ToUpper(node.Value) + " " + addedContentSummary + " in " + destination.Value + " " + inheritLabels + " " + unsetAnnotations + " " + inheritAnnotations + " " + newAnnotations, nil
+		hasLink := ""
+		if s.hasLink {
+			hasLink = " --link"
+		}
+		return "/bin/sh -c #(nop) " + strings.ToUpper(node.Value) + hasLink + " " + addedContentSummary + " in " + destination.Value + " " + labelsAndAnnotations, nil
 	default:
-		return "/bin/sh -c #(nop) " + node.Original + inheritLabels + unsetAnnotations + inheritAnnotations + newAnnotations, nil
+		return "/bin/sh -c #(nop) " + node.Original + labelsAndAnnotations, nil
 	}
 }
 
@@ -2326,6 +2325,7 @@ func (s *StageExecutor) intermediateImageExists(ctx context.Context, currNode *p
 		if s.builder.TopLayer != imageParentLayerID {
 			continue
 		}
+
 		// Next we double check that the history of this image is equivalent to the previous
 		// lines in the Dockerfile up till the point we are at in the build.
 		manifestType, history, diffIDs, err := s.executor.getImageTypeAndHistoryAndDiffIDs(ctx, image.ID)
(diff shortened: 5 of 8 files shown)


nalind:
Can you rebase onto the main branch and squash your commits?

tests/bud.bats
```diff
@@ -8075,3 +8075,488 @@ EOF
...
+  run_buildah from --name test-ctr link-glob2
+  run_buildah run test-ctr cat /testfiles/test1.txt
+  expect_output "test1"
+  run_buildah run test-ctr cat /testfiles/test2.txt
+  expect_output "test2"
+  run_buildah 1 run test-ctr ls /testfiles/other.file
+  run_buildah rm test-ctr
+}
+
+@test "bud --link large file performance" {
```
nalind: Drop this test.

nalind:
The "user-facing change" section is the kind of text which would be reproduced verbatim in the release notes for a release (for example, if we started scripting, digging the descriptions of the merged PRs, and snipping out just the bit with the "```release-note" header), so it should describe the change instead of alluding to it.

add.go
```diff
@@ -561,31 +593,54 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
...
-			}()
+				}()
+			} else {
+				go func() {
+					b.ContentDigester.Start("")
+					hashCloser := b.ContentDigester.Hash()
+					hasher := io.Writer(hashCloser)
+					if options.Hasher != nil {
+						hasher = io.MultiWriter(hasher, options.Hasher)
+					}
```
nalind, line 652: Why skip the hashing calls for the Link case?

cmd/buildah/addcopy.go
```diff
@@ -74,6 +75,7 @@ func applyFlagVars(flags *pflag.FlagSet, opts *addCopyResults) {
 	flags.StringVar(&opts.chown, "chown", "", "set the user and group ownership of the destination content")
 	flags.StringVar(&opts.chmod, "chmod", "", "set the access permissions of the destination content")
 	flags.StringVar(&opts.creds, "creds", "", "use `[username[:password]]` for accessing registries when pulling images")
+	flags.BoolVar(&opts.link, "link", false, "linking files instead of copying them when possible")
```
nalind: That description for what the option does is incorrect.

add.go
```diff
@@ -102,6 +105,7 @@ type AddAndCopyOptions struct {
 	Parents bool
 	// Timestamp is a timestamp to override on all content as it is being read.
 	Timestamp *time.Time
+	Link      bool
```
nalind: This is a public API, so it needs a docstring describing what effect setting this field has.

add.go
```diff
@@ -495,15 +501,41 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
...
+		}
+		defer os.RemoveAll(stagingDir)
+
+		putRoot = stagingDir
+
+		if renameTarget != "" {
+			putDir = filepath.Dir(filepath.Join(stagingDir, strings.TrimPrefix(destination, string(os.PathSeparator))))
+		} else {
+			putDir = filepath.Join(stagingDir, strings.TrimPrefix(destination, string(os.PathSeparator)))
+		}
```
nalind, line 557: `filepath.Join()`'s documentation notes that it removes empty elements, so the `strings.TrimPrefix()` calls aren't necessary, but more importantly there's an opening for a security issue here if the destination is allowed to have leading (or enough) "../" elements in it.

add.go
ajoshua2004: I skipped hashing inside the loop so I can first combine all the sources from a COPY into one temp file, and then hash that single file after the loop to get the correct digest for the whole layer.

nalind: Undo that - the commit-time code calculates the diffID on its own, and the ContentDigester already combines the digests when multiple items are being copied.
it might be faster to leave the directory as a directory, since the commit logic will tar it up if the linked layer's BlobPath is a directory as part of computing the diffID.  It would at least be a lot less work.

add.go
```diff
@@ -776,6 +856,123 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
...
+		createdBy := fmt.Sprintf("/bin/sh -c #(nop) %s --link %s in %s", Command, summary, destination)
+		history := v1.History{
+			Created:   &created,
+			CreatedBy: createdBy,
+			Comment:   b.HistoryComment(),
+		}
+
+		linkedLayer := LinkedLayer{
+			History:  history,
+			BlobPath: layerFile,
```
nalind: This could point directly to `stagingDir`, skipping a lot of the work being done in between.

nalind (changes requested):
There's a some duplication of effort in here that shouldn't be necessary.

tests/bud.bats
```diff
@@ -8079,3 +8079,462 @@ EOF
...
+_EOF
+
+  run_buildah build --layers $WITH_POLICY_JSON -t link-perms1 $contextdir
+
+  run_buildah build --layers $WITH_POLICY_JSON -t link-perms2 $contextdir
+  expect_output --substring "Using cache" "should use cache with chmod/chown flags"
+
+  run_buildah from --name test-ctr link-perms2
+  run_buildah run test-ctr ls -la /testfile.txt
+  expect_output --substring "nobody" "file should be owned by nobody"
```
nalind: When this test runs as an unprivileged user, this could either be the "nobody" user, or an unmapped ID, which the kernel defaults to showing as 65534, which we assign to the "nobody" user.  This should probably use a numeric user and group ID, and it should check that the permissions are set as specified.  I would probably use "stat" and its "-c" flag for these checks.

tests/bud.bats
```diff
@@ -8079,3 +8079,462 @@ EOF
...
+  run_buildah run test-ctr cat /linked.txt
+  expect_output "linked file"
+  run_buildah run test-ctr cat /regular2.txt
+  expect_output "regular file 2"
+  run_buildah run test-ctr cat /linked2.txt
+  expect_output "linked file 2"
+  run_buildah rm test-ctr
+}
+
+@test "bud --link with directories" {
```
nalind: This can be merged with the previous test.

tests/bud.bats
```diff
@@ -8079,3 +8079,462 @@ EOF
...
+  cat > $contextdir/Containerfile2 << _EOF
+FROM alpine
+ADD --link file2.txt /output.txt
+RUN echo "using file2"
+_EOF
+
+  run_buildah build --layers $WITH_POLICY_JSON -t link-unique1 -f $contextdir/Containerfile1 $contextdir
+
+  run_buildah build --layers $WITH_POLICY_JSON -t link-unique2 -f $contextdir/Containerfile2 $contextdir
+  assert "$output" !~ "Using cache.*ADD --link" "should not use cache for different source files even with same content"
```
nalind: This test looks like it's testing things that "bud --link cache miss with different destination" already covers.

tests/bud.bats
```diff
@@ -8079,3 +8079,462 @@ EOF
...
+FROM alpine
+ADD --link deep /deep
+RUN find /deep -name "*.txt" -exec cat {} \;
+_EOF
+
+  run_buildah build --layers $WITH_POLICY_JSON -t link-deep1 $contextdir
+  expect_output --substring "deep file"
+
+  run_buildah build --layers $WITH_POLICY_JSON -t link-deep2 $contextdir
+  expect_output --substring "Using cache"
```
nalind: Is this looking for a cache hit on the ADD instruction, or the RUN instruction, or both?

cmd/buildah/addcopy.go
```diff
@@ -263,6 +265,7 @@ func addAndCopyCmd(c *cobra.Command, args []string, verb string, iopts addCopyRe
 		MaxRetries:            iopts.retry,
 		Parents:               iopts.parents,
 		Timestamp:             timestamp,
+		Link:                  iopts.link,
```
nalind, line 268: Is there test coverage for `buildah copy --link`?

docs/buildah-copy.1.md
```diff
@@ -65,6 +65,11 @@ is preserved.
 
 Path to an alternative .containerignore (.dockerignore) file. Requires \-\-contextdir be specified.
 
+**--link**
+
+Create an independent cached layer for the added content. This allows the
+layer to be reused in subsequent builds even if the base image layers change.
+
```
nalind, line 74: This should probably describe how this content gets combined with changes made to the working container's rootfs with `buildah run` or using `buildah mount`.

imagebuildah/stage_executor.go
```diff
@@ -2320,12 +2336,45 @@ func (s *StageExecutor) intermediateImageExists(ctx context.Context, currNode *p
...
+						baseLayersMatch = false
+						break
+					}
+				}
+			}
+
+			if !baseLayersMatch {
+				logrus.Debugf("Base layers don't match for linked layer candidate")
+				continue
+			}
```
nalind: Is this still needed, now that the history looks like it should?

add.go
```diff
@@ -741,12 +799,17 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 						ChmodFiles:      nil,
 						IgnoreDevices:   userns.RunningInUserNS(),
 					}
-					putErr = copier.Put(extractDirectory, extractDirectory, putOptions, io.TeeReader(pipeReader, hasher))
+					if options.Link {
+						putErr = copier.Put(putRoot, putDir, putOptions, io.TeeReader(pipeReader, hasher))
+					} else {
+						putErr = copier.Put(extractDirectory, extractDirectory, putOptions, io.TeeReader(pipeReader, hasher))
+					}
```
nalind: This can probably be simplified further if `putRoot` and `putDir` are set in the branch that calls `copier.Mkdir()` above.

add.go
```diff
@@ -580,7 +634,11 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 						ChmodFiles:    nil,
 						IgnoreDevices: userns.RunningInUserNS(),
 					}
-					putErr = copier.Put(extractDirectory, extractDirectory, putOptions, io.TeeReader(pipeReader, hasher))
+					if options.Link {
+						putErr = copier.Put(putRoot, putDir, putOptions, io.TeeReader(pipeReader, hasher))
+					} else {
+						putErr = copier.Put(extractDirectory, extractDirectory, putOptions, io.TeeReader(pipeReader, hasher))
+					}
```
nalind: This can probably be simplified further if `putRoot` and `putDir` are set in the branch that calls `copier.Mkdir()` above.

docs/buildah-add.1.md
```diff
@@ -65,6 +65,13 @@ can be used.
 
 Path to an alternative .containerignore (.dockerignore) file. Requires \-\-contextdir be specified.
 
+**--link**
+
+Create an independent image layer instead of modifying the working container's
+rootfs. At commit time, this layer is added on top of any changes made via
+`buildah run` or `buildah mount`. With `buildah build --layers`, these
+independent layers can be cached and reused even if earlier layers change.
+
```
nalind, line 74: `buildah build` and `buildah add` can't be used together.  Is this suggesting that they can?
"on top" is not specific.  If `buildah run` creates a file, and `buildah add --link` copies something to the same path, will the reader know which to expect to see in the committed image?

docs/buildah-copy.1.md
nalind: `buildah build` and `buildah copy` can't be used together.  Is this suggesting that they can?
"on top" is not specific.  If `buildah run` creates a file, and `buildah copy --link` copies something to the same path, will the reader know which to expect to see in the committed image?

image.go
```diff
@@ -1360,6 +1360,15 @@ func (b *Builder) makeLinkedLayerInfos(layers []LinkedLayer, layerType string, l
...
 				}
+
+				if strings.Contains(info.linkedLayer.BlobPath, "link-stage-") {
+					stagingPath := info.linkedLayer.BlobPath
+					defer func() {
+						if err := os.RemoveAll(stagingPath); err != nil {
+							logrus.Warnf("failed to remove staging directory %q: %v", stagingPath, err)
+						}
+					}()
+				}
```
nalind: This looks like it'll break `buildah commit` if someone tries to `buildah commit` the same working container more than once.

imagebuildah/stage_executor.go
```diff
@@ -583,6 +587,8 @@ func (s *StageExecutor) performCopy(excludes []string, copies ...imagebuilder.Co
 			MaxRetries:            s.executor.maxPullPushRetries,
 			RetryDelay:            s.executor.retryPullPushDelay,
 			Parents:               copy.Parents,
+			Link:                  s.hasLink,
+			Timestamp:             s.executor.timestamp,
```
nalind: Why do we want to start setting the `Timestamp` field here?

imagebuildah/stage_executor.go
```diff
@@ -1880,10 +1888,6 @@ func (s *StageExecutor) historyAndDiffIDsMatch(baseHistory []v1.History, baseDif
 // getCreatedBy returns the command the image at node will be created by.  If
 // the passed-in CompositeDigester is not nil, it is assumed to have the digest
 // information for the content if the node is ADD or COPY.
-//
-// This function acts differently if getCreatedBy is invoked by LastStep. For instances
-// certain instructions like `removing annotations` does not makes sense for every step
-// but only makes sense if the step is last step of a build.
```
nalind: Is this no longer the case?

imagebuildah/stage_executor.go
```diff
@@ -2006,6 +2010,18 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
...
+			// Format matching what's used in Add() for linked layers
+			dest := destination.Value
+			if !strings.HasPrefix(dest, "/") {
+				dest = "/" + dest
+			}
+			createdby := fmt.Sprintf("/bin/sh -c #(nop) %s --link %s in %s",
+				strings.ToUpper(node.Value), addedContentSummary, dest)
+
+			return createdby, nil
+		}
```
nalind, line 2007: This'll be shorter if it defines an empty string variable, sets it to " --link" if hasLink is true, and adds that variable to the list of values that are concatenated on the following line.

tests/copy.bats
```diff
@@ -668,3 +668,114 @@ parents/y/b.txt"
...
+  mkdir $root/subdir
+  run_buildah copy --link $cid ${TEST_SCRATCH_DIR}/randomfile ${TEST_SCRATCH_DIR}/other-randomfile /subdir
+  
+  run_buildah unmount $cid
+  run_buildah commit $WITH_POLICY_JSON $cid containers-storage:copy-link-image
+  run_buildah rm $cid
+
+  run_buildah inspect --type=image copy-link-image
+  layers=$(echo "$output" | jq -r '.OCIv1.rootfs.diff_ids | length')
+  if [ "$layers" -lt 2 ]; then
```
nalind: Use the assert() helper function, defined in `helpers.bash`, here.

tests/add.bats
```diff
@@ -392,3 +392,74 @@ EOF
...
+  newroot=$output
+  
+  test -s $newroot/randomfile
+  cmp ${TEST_SCRATCH_DIR}/randomfile $newroot/randomfile
+  test -s $newroot/subdir/randomfile
+  cmp ${TEST_SCRATCH_DIR}/randomfile $newroot/subdir/randomfile
+  test -s $newroot/subdir/other-randomfile
+  cmp ${TEST_SCRATCH_DIR}/other-randomfile $newroot/subdir/other-randomfile
+  
+  run_buildah rm $newcid
```
nalind: Unlike in podman's tests, we wipe out the test's storage area when the test completes, so explicit removal of images and containers is not necessary.

nalind (changes requested):

imagebuildah/stage_executor.go
ajoshua2004: i accidentally deleted this

add.go
```diff
@@ -495,15 +520,67 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
...
+
+		putDirAbs, err := filepath.Abs(putDir)
+		if err != nil {
+			return fmt.Errorf("failed to resolve absolute path: %w", err)
+		}
+
+		stagingDirAbs, err := filepath.Abs(stagingDir)
+		if err != nil {
+			return fmt.Errorf("failed to resolve staging directory absolute path: %w", err)
+		}
```
nalind, line 553: `stagingDir` would already be an absolute path, so `putDir` would be as well.

tests/add.bats
```diff
@@ -392,3 +392,74 @@ EOF
...
+  run_buildah from $WITH_POLICY_JSON scratch
+  cid=$output
+  run_buildah mount $cid
+  root=$output
+  
+  run_buildah config --workingdir=/ $cid
+  run_buildah add --link $cid ${TEST_SCRATCH_DIR}/randomfile
+  
+  mkdir $root/subdir
+  run_buildah add --link $cid ${TEST_SCRATCH_DIR}/randomfile ${TEST_SCRATCH_DIR}/other-randomfile /subdir
```
nalind, line 415: The new logic in Add() has a branch for a case where a file is being copied to a file with a different name.  This test should exercise that.

tests/add.bats
```diff
@@ -392,3 +392,74 @@ EOF
...
+  createrandom ${TEST_SCRATCH_DIR}/file1
+  createrandom ${TEST_SCRATCH_DIR}/file2
+  
+  tar -c -C ${TEST_SCRATCH_DIR} -f ${TEST_SCRATCH_DIR}/archive.tar file1 file2
+
+  run_buildah from $WITH_POLICY_JSON scratch
+  cid=$output
+  
+  run_buildah config --workingdir=/ $cid
+  run_buildah add --link $cid ${TEST_SCRATCH_DIR}/archive.tar
```
nalind, line 457: This should also check that the right thing happens when the archive is being added somewhere other than the root directory of what will become the linked layer.

tests/add.bats
nalind: Multiple instances of this throughout.

add.go
```diff
@@ -102,6 +103,30 @@ type AddAndCopyOptions struct {
...
+	// createdBy history string to track this build-time decision.
+	InheritAnnotations types.OptionalBool
+	// Annotations is a list of key=value pairs to add as image annotations for --link operations.
+	// These are joined with commas and appended to the createdBy history string
+	// to ensure cache misses when annotation values differ between builds.
+	Annotations []string
+	// IsLastStep indicates whether this is the final instruction in the build for --link operations.
+	// When true, annotation and inheritance metadata is included in the createdBy
+	// history to ensure proper layer caching behavior.
+	IsLastStep bool
```
nalind: Probably simpler to cook all of this down to a string that gets appended to the history entry when `Link` is set, and have the caller compute it in a helper function that's extracted from `getCreatedBy()`.  Then that knowledge won't have to be duplicated and kept in sync going forward.

add.go
```diff
@@ -495,15 +520,67 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
...
+
+		stagingDir, err = os.MkdirTemp(containerDir, "link-stage-")
+		if err != nil {
+			return fmt.Errorf("creating staging directory for link %q: %w", b.ContainerID, err)
+		}
+
+		putRoot = stagingDir
+
+		cleanDest := filepath.Clean(destination)
+		cleanDest = strings.TrimPrefix(cleanDest, string(os.PathSeparator))
```
nalind: `filepath.Join` already ignores empty path components, so this isn't necessary for its sake.

add.go
```diff
@@ -776,6 +854,68 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
...
+	if options.Link {
+		created := time.Now().UTC()
+		if options.Timestamp != nil {
+			created = *options.Timestamp
+		}
+
+		Command := "ADD"
+		if !extract {
+			Command = "COPY"
+		}
```
nalind, line 874: We tend to give local variables names that start with lower-case letters.

nalind (changes requested):
The tests should also check that when caching isn't being used (--no-cache, or have the builds use different base image, etc.), a `COPY --link` in the Dockerfile will produce a layer with the same diffID across two different builds.  That's how we'll know that pushing the resulting images will benefit from blob reuse in the registry.
Committing to an OCI layout by building with `-t oci:$some_temporary_directory`, and then poking at it with the oci_image_diff_id() shell helper function should help with writing that test.

add.go
```diff
@@ -495,15 +501,75 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
...
+		if err != nil {
+			return fmt.Errorf("failed to resolve staging directory absolute path: %w", err)
+		}
+
+		if !strings.HasPrefix(putDirAbs, stagingDirAbs+string(os.PathSeparator)) && putDirAbs != stagingDirAbs {
+			return fmt.Errorf("destination path %q escapes staging directory", destination)
+		}
+		if err := os.MkdirAll(putDir, 0o755); err != nil {
+			return err
+		}
```
nalind, line 560: This should use the same type of `copier.Mkdir` call the non-link case does, in case ID mappings require that the directory not be owned by the current root user.

imagebuildah/stage_executor.go
```diff
@@ -359,8 +361,11 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
 			}
 			return errors.New("COPY --keep-git-dir is not supported")
 		}
-		if cp.Link {
-			return errors.New("COPY --link is not supported")
+		if cp.Link && s.executor.layers {
+			s.hasLink = true
+		} else if cp.Link {
+			s.executor.logger.Warn("--link is not supported when building without layers, ignoring\n")
```
nalind:
```suggestion
			s.executor.logger.Warn("--link is not supported when building without --layers, ignoring")
```

imagebuildah/stage_executor.go
```diff
@@ -2581,3 +2600,32 @@ func (s *StageExecutor) EnsureContainerPathAs(path, user string, mode *os.FileMo
 	logrus.Debugf("EnsureContainerPath %q (owner %q, mode %o) in %q", path, user, mode, s.builder.ContainerID)
 	return s.builder.EnsureContainerPathAs(path, user, mode)
 }
+
+func (s *StageExecutor) Annotations(isLastStep bool) string {
```
nalind: This method should not be exported.  Naming it `Annotations`, when it isn't solely about them, is potentially confusing to other maintainers.

imagebuildah/stage_executor.go
```diff
@@ -2006,7 +2017,13 @@ func (s *StageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
...
 			destination = destination.Next
 		}
-		return "/bin/sh -c #(nop) " + strings.ToUpper(node.Value) + " " + addedContentSummary + " in " + destination.Value + " " + inheritLabels + " " + unsetAnnotations + " " + inheritAnnotations + " " + newAnnotations, nil
+
+		link := ""
+		if s.hasLink {
+			link = " --link"
+		}
+
+		return "/bin/sh -c #(nop) " + strings.ToUpper(node.Value) + link + " " + addedContentSummary + " in " + destination.Value + " " + inheritLabels + " " + unsetAnnotations + " " + inheritAnnotations + " " + newAnnotations, nil
```
nalind: Use the new method here, and for the "RUN" case.

nalind (changes requested):
Relatively minor requests, mostly LGTM

add.go
```diff
@@ -33,6 +33,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 	"github.com/moby/sys/userns"
 	digest "github.com/opencontainers/go-digest"
+	v1 "github.com/opencontainers/image-spec/specs-go/v1"
```
nalind, line 36:
```suggestion
	"github.com/opencontainers/image-spec/specs-go/v1"
```

add.go
```diff
@@ -102,6 +103,12 @@ type AddAndCopyOptions struct {
 	Parents bool
 	// Timestamp is a timestamp to override on all content as it is being read.
 	Timestamp *time.Time
+	// Link, when set to true, creates an independent layer containing the copied content
+	// that sits on top of existing layers. This layer can be cached and reused
+	// separately, and is not affected by filesystem changes from previous intructions.
```
nalind:
```suggestion
	// separately, and is not affected by filesystem changes from previous instructions.
```

add.go
```diff
@@ -102,6 +103,12 @@ type AddAndCopyOptions struct {
...
+	Link bool
+	// BuildMetadata contains history metadata (labels/annotations) for cache invalidation.
```
nalind: I don't think an external user of this API is going to know what to put here, or when.  Should probably mention it's only consulted when `Link` is set, and is used for cache evaluation by the imagebuildah package.

add.go
```diff
@@ -495,15 +502,75 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
...
+
+		stagingDirAbs, err := filepath.Abs(stagingDir)
+		if err != nil {
+			return fmt.Errorf("failed to resolve staging directory absolute path: %w", err)
+		}
+
+		if !strings.HasPrefix(putDirAbs, stagingDirAbs+string(os.PathSeparator)) && putDirAbs != stagingDirAbs {
+			return fmt.Errorf("destination path %q escapes staging directory", destination)
+		}
+		if err := copier.Mkdir(stagingDir, filepath.Join(cleanDest), mkdirOptions); err != nil {
```
nalind:
```suggestion
		if err := copier.Mkdir(putRoot, putDirAbs, mkdirOptions); err != nil {
```

imagebuildah/stage_executor.go
```diff
@@ -359,8 +361,11 @@ func (s *StageExecutor) Copy(excludes []string, copies ...imagebuilder.Copy) err
 			}
 			return errors.New("COPY --keep-git-dir is not supported")
 		}
-		if cp.Link {
-			return errors.New("COPY --link is not supported")
+		if cp.Link && s.executor.layers {
+			s.hasLink = true
+		} else if cp.Link {
+			s.executor.logger.Warn("--link is not supported when building without --layers, ignoring\n")
```
nalind:
```suggestion
			s.executor.logger.Warn("--link is not supported when building without --layers, ignoring --link")
```

imagebuildah/stage_executor.go
```diff
@@ -2581,3 +2579,30 @@ func (s *StageExecutor) EnsureContainerPathAs(path, user string, mode *os.FileMo
...
+			inheritAnnotations = "|inheritAnnotations=false"
+		}
+		// If new annotations are added, they must be added as part of the last step of the build,
+		// so mention in history that new annotations were added inorder to make sure the builds
+		// can either reuse layers or burst the cache depending upon new annotations.
+		if len(s.executor.annotations) > 0 {
+			newAnnotations += strings.Join(s.executor.annotations, ",")
+		}
+	}
+	return inheritLabels, unsetAnnotations, inheritAnnotations, newAnnotations
```
nalind: It would simplify things for the callers if this returned the a single value with these concatenated, since they're always combined the same way, including the leading space if one is needed.

tests/add.bats
```diff
@@ -392,3 +392,120 @@ EOF
...
+  
+  # Test 2: Add with rename (file to file with different name)
+  run_buildah add --link $cid ${TEST_SCRATCH_DIR}/randomfile /renamed-file
+  
+  # Test 3: Multiple files to directory
+  mkdir $root/subdir
+  run_buildah add --link $cid ${TEST_SCRATCH_DIR}/randomfile ${TEST_SCRATCH_DIR}/other-randomfile /subdir
+  
+  run_buildah unmount $cid
+  run_buildah commit $WITH_POLICY_JSON $cid containers-storage:add-link-image
```
nalind:
```suggestion
  run_buildah commit $WITH_POLICY_JSON $cid add-link-image
```
The `containers-storage:` is implicit, and we're not in the habit of specifying it in tests.

tests/add.bats
```diff
@@ -392,3 +392,120 @@ EOF
...
+  run_buildah from $WITH_POLICY_JSON scratch
+  cid=$output
+  
+  run_buildah config --workingdir=/ $cid
+  
+  run_buildah add --link $cid ${TEST_SCRATCH_DIR}/archive.tar
+  
+  run_buildah add --link $cid ${TEST_SCRATCH_DIR}/archive.tar /destdir/
+  
+  run_buildah commit $WITH_POLICY_JSON $cid containers-storage:add-link-archive-image
```
nalind:
```suggestion
  run_buildah commit $WITH_POLICY_JSON $cid add-link-archive-image
```
The `containers-storage:` is implicit, and we're not in the habit of specifying it in tests.

tests/add.bats
```diff
@@ -392,3 +392,120 @@ EOF
...
+  createrandom ${TEST_SCRATCH_DIR}/testdir/subdir/file2
+
+  run_buildah from $WITH_POLICY_JSON scratch
+  cid=$output
+  
+  run_buildah config --workingdir=/ $cid
+  
+  run_buildah add --link $cid ${TEST_SCRATCH_DIR}/testdir /testdir
+  
+  run_buildah commit $WITH_POLICY_JSON $cid containers-storage:add-link-dir-image
```
nalind:
```suggestion
  run_buildah commit $WITH_POLICY_JSON $cid add-link-dir-image
```
The `containers-storage:` is implicit, and we're not in the habit of specifying it in tests.

tests/bud.bats
```diff
@@ -8146,3 +8146,237 @@ EOF
...
 }
+
+@test "bud --link consistent diffID with --no-cache" {
+  _prefetch alpine
+  local contextdir=${TEST_SCRATCH_DIR}/bud/link-diffid-nocache
+  mkdir -p $contextdir
+  
+  cat > $contextdir/Dockerfile << EOF
+FROM alpine
+RUN echo "before link layer"
```
nalind: This doesn't change the rootfs, and likely produces an empty tarball as the layer diff.  Did this intend to leave some trace in the layer?  Same question applies to most of the "RUN echo" and "RUN ls" instructions in these tests that don't redirect output to a file, give or take using "tee" to also send it to stdout.

tests/copy.bats
```diff
@@ -668,3 +668,97 @@ parents/y/b.txt"
...
+  root=$output
+  
+  run_buildah config --workingdir=/ $cid
+  run_buildah copy --link $cid ${TEST_SCRATCH_DIR}/randomfile
+  
+  mkdir $root/subdir
+  run_buildah copy --link $cid ${TEST_SCRATCH_DIR}/randomfile ${TEST_SCRATCH_DIR}/other-randomfile /subdir
+  
+  run_buildah unmount $cid
+  run_buildah commit $WITH_POLICY_JSON $cid containers-storage:copy-link-image
```
nalind:
```suggestion
  run_buildah commit $WITH_POLICY_JSON $cid copy-link-image
```
The `containers-storage:` is implicit, and we don't tend to specify it explicitly in tests.

tests/copy.bats
```diff
@@ -668,3 +668,97 @@ parents/y/b.txt"
...
+
+  run_buildah inspect --type=image copy-link-image
+  layers=$(echo "$output" | jq -r '.OCIv1.rootfs.diff_ids | length')
+  assert "$layers" -ge 2 "Expected at least 2 layers from 2 --link operations, but found $layers"
```
nalind: Does this number vary?  Can this check for an exact value?

tests/copy.bats
```diff
@@ -668,3 +668,97 @@ parents/y/b.txt"
...
+  createrandom ${TEST_SCRATCH_DIR}/sourcedir/file1
+  createrandom ${TEST_SCRATCH_DIR}/sourcedir/file2
+
+  run_buildah from $WITH_POLICY_JSON scratch
+  cid=$output
+  
+  run_buildah config --workingdir=/ $cid
+  run_buildah copy --link $cid ${TEST_SCRATCH_DIR}/sourcedir /destdir
+  
+  run_buildah commit $WITH_POLICY_JSON $cid containers-storage:copy-link-dir-image
```
nalind:
```suggestion
  run_buildah commit $WITH_POLICY_JSON $cid copy-link-dir-image
```
The `containers-storage:` is implicit, and we don't tend to specify it in tests.

tests/copy.bats
```diff
@@ -668,3 +668,97 @@ parents/y/b.txt"
...
+@test "copy-link-with-chown" {
+  createrandom ${TEST_SCRATCH_DIR}/randomfile
+  
+  _prefetch busybox
+  run_buildah from --quiet $WITH_POLICY_JSON busybox
+  cid=$output
+  
+  run_buildah copy --link --chown bin:bin $cid ${TEST_SCRATCH_DIR}/randomfile /tmp/random
+  
+  run_buildah commit $WITH_POLICY_JSON $cid containers-storage:copy-link-chown-image
```
nalind:
```suggestion
  run_buildah commit $WITH_POLICY_JSON $cid copy-link-chown-image
```
The `containers-storage:` is implicit, and we don't tend to specify it in tests.

tests/copy.bats
```diff
@@ -668,3 +668,97 @@ parents/y/b.txt"
...
+@test "copy-link-with-chmod" {
+  createrandom ${TEST_SCRATCH_DIR}/randomfile
+  
+  _prefetch busybox
+  run_buildah from --quiet $WITH_POLICY_JSON busybox
+  cid=$output
+  
+  run_buildah copy --link --chmod 777 $cid ${TEST_SCRATCH_DIR}/randomfile /tmp/random
+  
+  run_buildah commit $WITH_POLICY_JSON $cid containers-storage:copy-link-chmod-image
```
nalind:
```suggestion
  run_buildah commit $WITH_POLICY_JSON $cid copy-link-chmod-image
```
The `containers-storage:` is implicit.

tests/copy.bats
ajoshua2004: i changed this to use eq instead

add.go
ajoshua2004: The usage for this usually involved using the v1 header so I dont think we should get rid of this

nalind: I don't follow.  The package name is already "v1".

ajoshua2004: I was just following the pattern found in other files they also had their imports like that. I think it should remain the same

nalind: If the package's name is already "v1", specifying it as "v1" is unnecessary.  Any place we're doing that is an instance where we should've known better.

docs/buildah-add.1.md
```diff
@@ -65,6 +65,13 @@ can be used.
 
 Path to an alternative .containerignore (.dockerignore) file. Requires \-\-contextdir be specified.
 
+**--link**
+
+Create an independent image layer for the added files instead of modifying the working
+container's filesystem. If buildah run creates a file and buildah add --link adds a file
```
TomSweeneyRedHat:
```suggestion
container's filesystem. If `buildah run` creates a file, and `buildah add --link` adds a file
```
A bit nitty, but we probably ought to backtick all

docs/buildah-add.1.md
```diff
@@ -65,6 +65,13 @@ can be used.
...
+to the same path, the file from buildah add --link will be present in the committed image.
```
TomSweeneyRedHat:
```suggestion
to the same path, the file from `buildah add --link` will be present in the committed image.
```

docs/buildah-copy.1.md
```diff
@@ -65,6 +65,13 @@ is preserved.
 
 Path to an alternative .containerignore (.dockerignore) file. Requires \-\-contextdir be specified.
 
+**--link**
+
+Create an independent image layer for the added files instead of modifying the working
+container's filesystem. If buildah run creates a file and buildah add --link adds a file
+to the same path, the file from buildah add --link will be present in the committed image.
```
TomSweeneyRedHat:
```suggestion
to the same path, the file from `buildah add --link` will be present in the committed image.
```

TomSweeneyRedHat:
I found a couple of nits that should be tweaked.  @nalind @2004joshua do you think this is close to getting merged this week?  We had a delivery date of next Monday.

docs/buildah-copy.1.md
nalind:
```suggestion
to the same path, the file from buildah copy --link will be present in the committed image.
```

add.go
```diff
@@ -495,15 +505,75 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
...
+				createdDirs = append(createdDirs, tempPath)
+			}
+			tempPath = filepath.Dir(tempPath)
+		}
+	} else {
+		if err := copier.Mkdir(mountPoint, extractDirectory, mkdirOptions); err != nil {
+			return fmt.Errorf("ensuring target directory exists: %w", err)
+		}
+
+		putRoot = extractDirectory
```
nalind, line 573:
```suggestion
		putRoot = mountPoint
```

add.go
```diff
@@ -33,6 +33,7 @@ import (
 	"github.com/hashicorp/go-multierror"
 	"github.com/moby/sys/userns"
 	digest "github.com/opencontainers/go-digest"
+	"github.com/opencontainers/image-spec/specs-go/v1"
```
nalind: Huh.  My editor keeps adding a duplicate of this line with the "v1" package name explicitly specified, which breaks the compile immediately.

ajoshua2004: thats why I was keeping it because vscode also kept adding a duplicate

add.go
nalind: I wasn't expect this to break a conformance test, but I guess it does.  Never mind?

nalind:
LGTM

nalind:

TomSweeneyRedHat (commented):
LGTM

TomSweeneyRedHat:
A couple of tests look to be hiccuping, restarting.

PR #6258, opened 2025-07-02
feat(profiling): Add `--trace-profile` flag

jboewer:
#### What type of PR is this?
/kind feature

Introduces a new `--trace-profile [file]` flag to generate a Go execution trace, for performance diagnostics using `go tool trace`.

Currently, buildah provides flags for profiling CPU `(--cpu-profile)` and memory `(--memory-profile)` usage.

However, when a build is slow but CPU usage is low, it's difficult to diagnose the bottleneck. These latency issues are often caused by time spent waiting on I/O (network or disk), blocking on system calls, or other scheduler events. The standard Go tool for investigating these specific issues is the execution tracer (go tool trace).

#### How to verify it
Run any buildah command with the additional `--trace-profile` flag.
For example
```
buildah info --trace-profile=/tmp/trace.out
go tool trace /tmp/trace.out
```

None

```release-note
add support for `--trace-profile [file]` to generate an execution trace.
```

### DIFF
--- a/tests/profiling.bats
+++ b/tests/profiling.bats
@@ -7,7 +7,7 @@ load helpers
   run_buildah info --trace-profile="$tmpfile"
   assert "$status" -eq 0 "buildah info with --trace-profile should succeed"
   [ -s "$tmpfile" ] || die "trace profile should not be empty"
-  run go tool trace -d=2 "$tmpfile"
+  run go tool trace -pprof=net "$tmpfile"
   assert "$status" -eq 0 "go tool trace should succeed"
   rm -f "$tmpfile"
 }


TomSweeneyRedHat:
@jboewer thanks for the contribution.  Can you add a test or three for this please?

jboewer:
> @jboewer thanks for the contribution. Can you add a test or three for this please?

Hey @TomSweeneyRedHat

Just added tests for all 3 available profiling flags.

Though as far as I know, the go tool CLI flags (such as those for `go tool trace` and `go tool pprof`) are not covered by Go’s backwards compatibility promise, so the flags or behavior might change in future Go releases and potentially break these tests.

Would you prefer to keep the current assertions that use go tool trace/pprof to validate the generated files, or should we limit the tests to just checking that the files are created and non-empty?


Edit: Tests were just failing as of this very reason. I used go v1.23 locally which had a different go tool trace CLI than the CI's go v1.24. I've switched to using the more stable -pprof=net flag instead of the debug -d=2 flag, but above's question still stands.

TomSweeneyRedHat:
@jboewer unless @nalind has a differing opinion, and especially given the "fun" you ran into with v1.23 locally, I'd go with the -pprof=net flag as you suggested.

jboewer:
Hey @TomSweeneyRedHat @nalind,

I just noticed I still have this PR open. Do you mind revisiting this subject?

Cheers

TomSweeneyRedHat:
LGTM
@nalind?

jboewer:
Hey @TomSweeneyRedHat & @nalind,

Would you still consider this PR or should I close it?

cheers

PR #6267, opened 2025-07-07
Remove BUILDTAG btrfs_noversion as no longer effective

rahilarious:
https://github.com/containers/storage/pull/2308

#### What type of PR is this?

/kind deprecation

```release-note
None
```


TomSweeneyRedHat:
LGTM
but tests are failing, will restart, and without looking, maybe cured by #6271?

Luap99:
@TomSweeneyRedHat packit tests are always non blocking so there is no reason to block on them

/lgtm

PR #6284, opened 2025-07-12
fix: don't try to tag oci-archive destinations

### LINKED ISSUE #6280: build to oci-archive, with single FROM statement fails
### Issue Description

```bash
./bin/buildah build -f <(echo 'FROM busybox') --tag=oci-archive:out.tar
```

fails. It succeeds if there is more than 1 step, or if no `oci-archive` is specified.

Run this command:

```bash
./bin/buildah build -f <(echo 'FROM busybox') --tag=oci-archive:out.tar
```

Unexpected failure:

```
STEP 1/1: FROM busybox
COMMIT oci-archive:out.tar
Error: locating new copy of image "7b4721e214600044496305a20ca3902677e572127d4d976ed0e54da0137c243a" (i.e., "oci-archive:out.tar:"): trying to resolve a non-containers-storage: reference "oci-archive:out.tar:"
```

`out.tar` to be produced without error.

Note, it does in fact appear to successfully create the file which looks like it has valid content - I think the error above occurs *after* the file is created.

### buildah version output

```yaml
buildah version 1.41.0-dev (image-spec 1.1.1, runtime-spec 1.2.1)


Built locally from:


commit 1c0e79c974f9c1381472654d3065defdb645e011 (upstream/main, upstream/HEAD, main)
Merge: a2108716b d9443fe42
Author: openshift-merge-bot[bot] <148852131+openshift-merge-bot[bot]@users.noreply.github.com>
Date:   Thu Jul 10 18:05:47 2025 +0000

    Merge pull request #6277 from containers/renovate/golang.org-x-term-0.x
```

### buildah info output

```yaml
{
    "host": {
        "CgroupVersion": "v2",
        "Distribution": {
            "distribution": "ubuntu",
            "version": "25.04"
        },
        "MemFree": 975556608,
        "MemTotal": 4091121664,
        "OCIRuntime": "runc",
        "SwapFree": 0,
        "SwapTotal": 0,
        "arch": "arm64",
        "cpus": 4,
        "hostname": "lima-default",
        "kernel": "6.14.0-23-generic",
        "os": "linux",
        "rootless": true,
        "uptime": "2h 18m 15.81s (Approximately 0.08 days)",
        "variant": ""
    },
    "store": {
        "ContainerStore": {
            "number": 0
        },
        "GraphDriverName": "overlay",
        "GraphOptions": null,
        "GraphRoot": "/home/aeijdenberg.linux/.local/share/containers/storage",
        "GraphStatus": {
            "Supports d_type": "true",
            "Supports shifting": "true",
            "Supports volatile": "true",
            "Using metacopy": "false"
        },
        "ImageStore": {
            "number": 2
        },
        "RunRoot": "/run/user/501/containers"
    }
}
```

### Provide your storage.conf

```toml
$ cat /usr/share/containers/storage.conf 
# files.
#
[storage]

driver = "overlay"

# Temporary storage location

# following commands:

# imagestore = ""


#

# transient_store = true

[storage.options]

additionalimagestores = [
]

# use_hard_links = "false"

# ostree_repos=""

# convert_images = "false"

#
#
#

#
# auto-userns-min-size=1024
#
# auto-userns-max-size=65536

[storage.options.overlay]
# and vfs drivers.

# inodes = ""

# directly.

mountopt = "nodev"

# skip_mount_home = "false"

# use_composefs = "false"

# size = ""

# directories.
#
#
#  "": No value specified.
#     image.
#     any user.
#
#
#  "force_mask"  permissions.
#
# force_mask = ""
```

### Upstream Latest Release

Yes

Running inside of an Ubuntu 25.04 VM on macOS using lima.

I had to run:

```bash
sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
```

else nothing would work, but I don't believe that is related to this issue.

### Additional information

The command does in fact produce the `out.tar` that is desired - but it generates the failure message and exits with a non-zero code.

If I remove `--tag=oci-archive:out.tar` then the command succeeds.

If I add a 2nd line, e.g.

```Dockerfile
FROM busybox
RUN echo foo
```

then it also succeeds.

aeijdenberg:
#### What type of PR is this?

/kind bug

This allows `buildah build -f <(echo 'FROM busybox') --tag=oci-archive:out.tar` to succeed without error.

#### How to verify it

bats test added.

FIxes #6280 

It's not 100% clear to me this is the right solution to fix, but I noticed that other code which adds tags to images, such as https://github.com/containers/buildah/blob/8403fd604bb78956e79bb9c826fab56aeaf2e8a6/imagebuildah/executor.go#L1054-L1057 does the same check before trying to tag, so this duplicates that check and adds it to this other code path.

```release-note
None
```


nalind:
Thanks!  Though this bypasses the `cp.Image()` call further down, which is what would write the output image, and I think we want to keep writing the image.

aeijdenberg:
> Thanks! Though this bypasses the `cp.Image()` call further down, which is what would write the output image, and I think we want to keep writing the image.

You're right, and we certainly do want to keep writing the image out. I'm pretty sure I had that working in an earlier iteration - I'll see what I missed, fix up and add more tests.

Thanks for the pickup...

aeijdenberg:
Rebased and included the additional fix originally submitted in #6298

dustymabe:
I think this helped me get around an error I was seeing. @nalind - mind another review?

aeijdenberg:
Rebased on latest main.

nalind:
LGTM, thanks!
/approve
/lgtm

PR #6285, opened 2025-07-14
feat: support --mount=type=secret,id=foo,env=bar

### LINKED ISSUE #5892: Support building Dockerfiles that mount secrets as environment variables
### Feature request description

Since Dockerfile 1.10 the Docker Engine can mount secrets as environment variables specified in the RUN command.

https://docs.docker.com/build/buildkit/dockerfile-release-notes/#1100

https://docs.docker.com/reference/dockerfile/#example-mount-as-environment-variable

It would be useful for podman build to support it. It's already possible using podman run.

aeijdenberg:
Allow secrets specified with mount to result in env variable set as an alternative to mounting a file.

This matches Docker behaviour:
https://docs.docker.com/reference/dockerfile/#example-mount-as-environment-variable

#### What type of PR is this?

/kind feature

Docker supports making secret mounts appear as environment variables during a `RUN` command. This adds the same to `buildah`.

See: https://docs.docker.com/reference/dockerfile/#example-mount-as-environment-variable

#### How to verify it

bats test added.

Fixes https://github.com/containers/buildah/issues/5892

```release-note
Adds support for mounting a secret to an environment variable. For example:

RUN --mount=type=secret,id=mysecret,env=FOO sh -c 'echo "Hello $FOO"'
```

### DIFF
--- a/define/types.go
+++ b/define/types.go
@@ -109,6 +109,21 @@ type Secret struct {
 	SourceType string
 }
 
+func (s Secret) ResolveValue() ([]byte, error) {
+	switch s.SourceType {
+	case "env":
+		return []byte(os.Getenv(s.Source)), nil
+	case "file":
+		rv, err := os.ReadFile(s.Source)
+		if err != nil {
+			return nil, fmt.Errorf("reading file for secret ID %s: %w", s.ID, err)
+		}
+		return rv, nil
+	default:
+		return nil, fmt.Errorf("invalid secret type: %s for secret ID: %s", s.SourceType, s.ID)
+	}
+}
+
 // BuildOutputOptions contains the the outcome of parsing the value of a build --output flag
 // Deprecated: This structure is now internal
 type BuildOutputOption struct {
--- a/run.go
+++ b/run.go
@@ -199,12 +199,12 @@ type runMountArtifacts struct {
 	MountedImages []string
 	// Agents are the ssh agents started, which should have their Shutdown() methods called
 	Agents []*sshagent.AgentServer
-	// SSHAuthSock is the path to the ssh auth sock inside the container
-	SSHAuthSock string
 	// Lock files, which should have their Unlock() methods called
 	TargetLocks []*lockfile.LockFile
 	// Intermediate mount points, which should be Unmount()ed and Removed()d
 	IntermediateMounts []string
+	// Environment variables that should be set for RUN that may contain secrets, each is name=value form
+	EnvVars []string
 }
 
 // RunMountInfo are the available run mounts for this run
--- a/run_common.go
+++ b/run_common.go
@@ -1433,6 +1433,9 @@ func (b *Builder) setupMounts(mountPoint string, spec *specs.Spec, bundlePath st
 		mounts = append(mounts, mount)
 	}
 
+	// Some mounts require env vars to be set, do these here
+	spec.Process.Env = append(spec.Process.Env, mountArtifacts.EnvVars...)
+
 	// Set the list in the spec.
 	spec.Mounts = mounts
 	succeeded = true
@@ -1522,7 +1525,7 @@ func (b *Builder) runSetupRunMounts(bundlePath string, mounts []string, sources 
 	intermediateMounts := make([]string, 0, len(mounts))
 	finalMounts := make([]specs.Mount, 0, len(mounts))
 	agents := make([]*sshagent.AgentServer, 0, len(mounts))
-	defaultSSHSock := ""
+	var envVars []string
 	targetLocks := []*lockfile.LockFile{}
 	var overlayDirs []string
 	succeeded := false
@@ -1561,11 +1564,7 @@ func (b *Builder) runSetupRunMounts(bundlePath string, mounts []string, sources 
 		}
 	}()
 	for _, mount := range mounts {
-		var mountSpec *specs.Mount
-		var err error
-		var envFile, image, bundleMountsDir, overlayDir, intermediateMount string
-		var agent *sshagent.AgentServer
-		var tl *lockfile.LockFile
+		var bundleMountsDir string
 
 		tokens := strings.Split(mount, ",")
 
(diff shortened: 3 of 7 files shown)


filip-be:
Hi Team!
What's the status of this PR? is it going to be updated and merged anytime soon or is there any pending issue that needs to be resolved?

aeijdenberg:
Rebased on latest main to fix merge conflict in test.

ab-gh:
Hi there, has there been any progress in the status of this PR? Not having docker-like `type=secret,env=bar` is really hindering some progress.

aeijdenberg:
Rebased to fix merge conflict.

aeijdenberg:
Rebased to fix merge conflict.

scodeman:
When can we expect a merge of this one ?

run_common.go
```diff
@@ -1731,13 +1740,17 @@ func (b *Builder) getTmpfsMount(tokens []string, idMaps IDMaps, workDir string)
 	return &volumes[0], nil
 }
 
-func (b *Builder) getSecretMount(tokens []string, secrets map[string]define.Secret, idMaps IDMaps, workdir string) (_ *specs.Mount, _ string, retErr error) {
-	errInvalidSyntax := errors.New("secret should have syntax id=id[,target=path,required=bool,mode=uint,uid=uint,gid=uint")
+func (b *Builder) getSecretMount(tokens []string, secrets map[string]define.Secret, idMaps IDMaps, workdir string) (rv struct {
+	Mount       *specs.Mount // set if mount created
```
Honny1: I'd make a separate type for this return value.

Honny1 (commented):
Just a small nit, otherwise LGTM.

aeijdenberg:
Rebased to fix merge conflict.

run_common.go
aeijdenberg: As written it is already a separate type (albeit anonymous). Are you asking that it be given a name? Happy to make that change if needed, and if someone can suggest a name, but since it's never referred to (and is a private implementation detail) it felt like giving it a name wouldn't give much benefit.

Honny1: I would prefer to use a separate type. Something like:

```
type SecretMountOrEnv struct {
	Mount       *specs.Mount // set if mount created
	EnvFile     string       // set if caller mount created from temp created env file
	EnvVariable string       // set if caller should add to env variable list
}
```
Then you could implement a method to handle [this code](https://github.com/containers/buildah/pull/6285/changes#diff-f575bb9d4f3b897d9cf43982beb0120ced85171a81678d4f7809f25946b0b8f6R1593-R1600).

---
I’d also avoid named return values, even when not using naked returns, as they can still lead to returning the wrong data.

aeijdenberg: I've given it the name suggested and removed the named `rv` return variable name.

I've kept the `retErr` named return variable, as this is used by a defer function and is a common pattern (and pre-dates this PR).

I haven't added a method to implement the code linked above, as it's unclear to me what the method signature should be. The lines that you've linked to modify 3 local slice variables, which doesn't lend itself well to a method.

Honny1 (approved):
LGTM Thanks

mmorhun:
@aeijdenberg could you resolve conflicts one more time?

aeijdenberg:
Rebased and squashed to fix merge conflicts.

run.go
```diff
@@ -205,6 +205,8 @@ type runMountArtifacts struct {
 	TargetLocks []*lockfile.LockFile
 	// Intermediate mount points, which should be Unmount()ed and Removed()d
 	IntermediateMounts []string
+	// Environment variables that should be set for RUN that contain secrets, each is name=value form
+	SecretEnvVars []string
```
nalind: This should just be `EnvVars`, since we could also mix non-secrets in there, in particular the "SSH_AUTH_SOCK" value, which would let us drop the `SSHAuthSock` field and a lot of the special handling for it.

define/types.go
```diff
@@ -109,6 +109,17 @@ type Secret struct {
...
 }
 
+func (s Secret) ResolveValue() ([]byte, error) {
+	switch s.SourceType {
+	case "env":
+		return []byte(os.Getenv(s.Source)), nil
+	case "file":
+		return os.ReadFile(s.Source)
+	default:
+		return nil, errors.New("invalid secret type")
```
nalind: This error message should include the unrecognized secret type and its ID, since it could be one of many that were specified.

define/types.go
```diff
@@ -109,6 +109,17 @@ type Secret struct {
 	SourceType string
 }
 
+func (s Secret) ResolveValue() ([]byte, error) {
+	switch s.SourceType {
+	case "env":
+		return []byte(os.Getenv(s.Source)), nil
+	case "file":
+		return os.ReadFile(s.Source)
```
nalind: This error message should include the secret's ID.  IIRC the standard library will already include the file's name.

run_common.go
```diff
@@ -1757,110 +1777,123 @@ func (b *Builder) getSecretMount(tokens []string, secrets map[string]define.Secr
...
 		}
 	}
 
 	if id == "" {
-		return nil, "", errInvalidSyntax
-	}
-	// Default location for secrets is /run/secrets/id
-	if target == "" {
-		target = "/run/secrets/" + id
+		return secretMountOrEnv{}, errInvalidSyntax
```
nalind, line 1821: If neither an "id" nor "env" value is specified, but a "target"/"dst"/"destination" is, I think the ID is supposed to be set to the target's basename, and processing is meant to continue.

run_common.go
```diff
@@ -1757,110 +1777,123 @@ func (b *Builder) getSecretMount(tokens []string, secrets map[string]define.Secr
...
+	}
+	data, err := secr.ResolveValue()
+	if err != nil {
+		return secretMountOrEnv{}, err
+	}
+
+	// if env is set, then we can return now
+	if env != "" {
+		rv.EnvVariable = env + "=" + string(data)
+		return rv, nil
```
nalind: If "env" and "target" are both supplied, I think the expectation is that they should both be honored.

run_common.go
```diff
@@ -1757,110 +1777,123 @@ func (b *Builder) getSecretMount(tokens []string, secrets map[string]define.Secr
...
 	var ctrFileOnHost string
 
 	switch secr.SourceType {
 	case "env":
-		data = []byte(os.Getenv(secr.Source))
 		tmpFile, err := os.CreateTemp(tmpdir.GetTempDir(), "buildah*")
 		if err != nil {
-			return nil, "", err
+			return secretMountOrEnv{}, err
 		}
```
nalind, line 1854: Long-standing bug: I don't think we ever closed this file handle.

tests/bud.bats
```diff
@@ -9476,3 +9476,15 @@ _EOF
...
   assert "$output" = "$source_date_epoch" "newfile should be clamped to source-date-epoch (expected ${source_date_epoch}, got ${output})"
 }
+
+@test "use-secret-to-env-variable" {
+  local outpath="${TEST_SCRATCH_DIR}/timestamp-after-secret.tar"
+
+  BAR=baz run_buildah build --secret id=mysecret,env=BAR -f <(printf "FROM alpine\nRUN --mount=type=secret,id=mysecret,env=FOO,required sh -c 'echo "'"Hello $FOO"'"'") --tag=oci-archive:${outpath}.a --timestamp 0
+  expect_output --substring "Hello baz"
+  BAR=boz run_buildah build --secret id=mysecret,env=BAR -f <(printf "FROM alpine\nRUN --mount=type=secret,id=mysecret,env=FOO,required sh -c 'echo "'"Hello $FOO"'"'") --tag=oci-archive:${outpath}.b --timestamp 0
+  expect_output --substring "Hello boz"
```
nalind, line 9500: This would need to also test cases where both a "target" and "env" were given.

nalind (changes requested):
Generally good stuff.  There's some subtle behavior in cases where only a "target" is given, or where both an "env" and a "target" are given, that I think we need to change to better align with expectations.

run_common.go
aeijdenberg: Yes, I can see that behaviour described at <https://docs.docker.com/build/building/secrets/#target>:

> It's possible to use the `target` and `env` options together to mount a secret as both a file and an environment variable.

I'll push a commit to match

define/types.go
aeijdenberg: Have pushed a commit to do so, however in practice this code isn't reached as the parsing is currently done in the following code, which isn't touched in this PR:
https://github.com/containers/buildah/blob/58482a8723206ed821155f2d5c2d6ef7ea240854/pkg/parse/parse.go#L1332-L1334

run.go
aeijdenberg: Removed `SSHAuthSock` field.

aeijdenberg: and renamed to `EnvVars`

define/types.go
aeijdenberg: Added.

run_common.go
aeijdenberg: I can't see an example of this elsewhere - do you have a source for that? I don't think this PR changes the previous behaviour.

run_common.go
aeijdenberg: Yeah, or if it does, it's hard to see where. I haven't addressed that pre-existing behaviour in this PR.

tests/bud.bats
aeijdenberg: Added tests for that.

aeijdenberg:
Have pushed a set of commits to address feedback. Happy to squash/rebase if preferred. Thanks for the review @nalind

run_common.go
nalind: https://docs.docker.com/reference/dockerfile#run---mounttypesecret, where it describes the default value for the "id" value.

define/types.go
```diff
@@ -109,6 +109,21 @@ type Secret struct {
...
 }
 
+func (s Secret) ResolveValue() ([]byte, error) {
+	switch s.SourceType {
+	case "env":
+		return []byte(os.Getenv(s.Source)), nil
+	case "file":
+		rv, err := os.ReadFile(s.Source)
+		if err != nil {
+			return nil, fmt.Errorf("error reading file for secret ID: %s, wrapped error: %w", s.ID, err)
```
nalind:
```suggestion
			return nil, fmt.Errorf("reading file for secret ID %s: %w", s.ID, err)
```
This looks more like other error messages we return.  The initial "error" will be duplicated by a prefixed "Error:" if it's displayed to the user, and the error could be wrapped multiple times, so we don't want these bits to be repeated at every level.

nalind:
@aeijdenberg Yes, a squash would be preferred.  Thanks!

run_common.go
aeijdenberg: Thanks, I've added code to match that documented behaviour.

define/types.go
aeijdenberg: Fixed as suggested.

aeijdenberg:
> @aeijdenberg Yes, a squash would be preferred. Thanks!

Done and pushed.

nalind (approved):
LGTM, thanks!

mmorhun:
Can it be merged now?

nalind:

Honny1 (approved):
LGTM as far as I can tell. 

Also, I don't have merge permissions.

PR #6286, opened 2025-07-14
CI: make runc tests non-blocking

nalind:
#### What type of PR is this?

/kind other

Make the test runs that use runc non-blocking for the Total Success condition.

#### How to verify it

If one of the test tasks that uses runc fails, but Total Success still passes, then we're good.

None

I don't have a proper understanding of yaml, so I'm guessing on some of this syntax.

```release-note
None
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -212,23 +212,11 @@ integration_task:
             BUILDAH_RUNTIME: crun
             RUNTIME_N: " using crun"
         - env:
-            DISTRO_NV: "${FEDORA_NAME}"
-            IMAGE_NAME: "${FEDORA_CACHE_IMAGE_NAME}"
-            STORAGE_DRIVER: 'vfs'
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
Some unhappy tests still.

I'm fine with either approch, this LGTM if you want to go forward wth this, instead of removing.  Also assuming happy tests.

nalind:
> Some unhappy tests still.

Ah, but they don't keep the overall "Total Success" status from passing, which is the point.

TomSweeneyRedHat:
@nalind do you want to move forward with this, or are you considering @Luap99 's option to remove the tests?

nalind:
> @nalind do you want to move forward with this, or are you considering @Luap99 's option to remove the tests?

In general, I'm not going to be on board with removing tests after putting in the work to get them passing.
So, here goes.
/unhold

Issue #5987, opened 2025-02-12
[feature request] support setting build-time environment variables


chmeliik:
As developers of a CI system that uses buildah for building (https://konflux-ci.dev/), we have a need to dynamically set environment variables at build time (for the `RUN` instructions in a Containerfile).[^1]

We didn't find any great way to achieve that at present. The options we considered:
* `--build-arg`
  * Requires the `ARG`s to be defined up-front in the Containerfile (or requires modifying the Container dynamically to inject them)
  * `ARG`s have a lower precedence than `ENV`. The Containerfile or the parent image may already be setting the variable via `ENV` => we're not able to override it
* `--env`
  * Sets the environment variable in the image config, which we want to avoid. We want to set the environment variables only at build time
* Inject a `.env` shell script into the build using `--volume`. Dynamically modify the `RUN` instructions in the Containerfile to `source` this shell script before doing the rest of the command(s)
  * _This is the option we currently use, despite the problems_
  * Modifying the RUN instructions is error-prone. We don't even support all the possible forms (e.g. exec form and heredoc form - [related issue](https://github.com/konflux-ci/build-definitions/issues/1200))
  * Some users are uncomfortable with having their Containerfile dynamically modified

It would be awesome if we could set build-time environment variables more easily. Some options that could work:
* Buildah currently supports the `env` array from the [engine table](https://github.com/containers/common/blob/main/docs/containers.conf.5.md#engine-table) in containers.conf. Would it be reasonable to support `env` from the [containers table](https://github.com/containers/common/blob/main/docs/containers.conf.5.md#containers-table) as well and use these for the `RUN` environment?
* A new `--build-env` CLI option for `buildah build` that would set environment variables for `RUN` instructions but wouldn't set them on the resulting image config.

[^1]: For example, we set `GOMODCACHE` to a pre-populated cache directory that we mount into the build using `--volume`. We do a similar thing for Python builds - pointing pip to a directory instead of an index server using `PIP_FIND_LINKS`.

flouthoc:
One hack which can be done is dump all the `env` into a file in `contextdir` and use 
`RUN --mount=type=bind,src=myfile.sh,target=myfile.sh,Z source myfile.sh <do other operation> `

@chmeliik Any thoughts ?

chmeliik:
> One hack which can be done is dump all the `env` into a file in `contextdir` and use `RUN --mount=type=bind,src=myfile.sh,target=myfile.sh,Z source myfile.sh <do other operation> `
[...]

PR #6289, opened 2025-07-15
feat(build): add --mount option

aeijdenberg:
This allows adding an ephemeral mount to all RUN commands in a Containerfile during a build.

#### What type of PR is this?

> /kind feature

This adds a `--mount` option to `build` which has the effect of adding this mount to each `RUN` command in a Containerfile before executing. For example:

`buildah build --mount type=secret,id=mysecret ...`

and a Containerfile entry of:

`RUN cat /run/secrets/mysecret`

Has the same effect as:

`RUN --mount=type=secret,id=mysecret cat /run/secrets/mysecret`

This is useful in conjunction with #6285 for building existing Containerfile's with different environmental considerations (such as SSL_CERT_FILE data etc) without needing to modify the existing Containerfile's or causing changes to the produced images.

#### How to verify it

bats test added. 

```
FOO=world ./bin/buildah build \
      --secret id=mysecret,env=FOO \
      --mount type=secret,id=mysecret,dst=/bar,required \
      -f <(printf 'FROM busybox\nRUN echo "hello $(cat /bar) welcome"')
```

Prints:

`hello world welcome`

None

There is an existing field `TransientMounts` in `BuildOptions` that had comment suggesting it would do exactly as described by this PR, however it does not quite operate as described, rather it expects each argument to be a `src:dest` volume mount (I suspect it predates other types of mounts):
https://github.com/containers/buildah/blob/db61e10b3f420786e86769d4aa1a1d2384df14c7/define/build.go#L148-L150

Since that field was part of the public API, I didn't change it's behaviour in this PR, but I did adjust the comment and added a new fields for other mounts.

```release-note
This adds a `--mount` option to `build` which has the effect of adding this mount to each `RUN` command in a Containerfile before executing. For example:

`buildah build --mount type=secret,id=mysecret ...`

and a Containerfile entry of:

`RUN cat /run/secrets/mysecret`

Has the same effect as:

`RUN --mount=type=secret,id=mysecret cat /run/secrets/mysecret`
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -844,7 +844,7 @@ func (s *stageExecutor) Run(run imagebuilder.Run, config docker.Config) error {
 		NoPivot:              os.Getenv("BUILDAH_NOPIVOT") != "" || s.executor.noPivotRoot,
 		Quiet:                s.executor.quiet,
 		CompatBuiltinVolumes: types.OptionalBoolFalse,
-		RunMounts:            run.Mounts,
+		RunMounts:            slices.Concat(run.Mounts, s.executor.transientRunMounts),
 		Runtime:              s.executor.runtime,
 		Secrets:              s.executor.secrets,
 		SSHSources:           s.executor.sshsources,
--- a/define/build.go
+++ b/define/build.go
@@ -145,9 +145,12 @@ type BuildOptions struct {
 	Runtime string
 	// RuntimeArgs adds global arguments for the runtime.
 	RuntimeArgs []string
-	// TransientMounts is a list of unparsed mounts that will be provided to
+	// TransientMounts is a list of unparsed src:dest volume instructions that will be provided to
 	// RUN instructions.
 	TransientMounts []string
+	// TransientRunMounts is a list of unparsed mounts (e.g. type=secret etc) that will be provided to
+	// RUN instructions.
+	TransientRunMounts []string
 	// CacheFrom specifies any remote repository which can be treated as
 	// potential cache source.
 	CacheFrom []reference.Named
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -639,6 +639,20 @@ Write information about the built image to the named file.  When `--platform`
 is specified more than once, attempting to use this option will trigger an
 error.
 
+**--mount** *mount-instruction*
+
+Adds this mount to each `RUN` command in a Containerfile before executing. For example:
+
+`buildah build --mount type=secret,id=mysecret ...`
+
+and a Containerfile entry of:
+
+`RUN cat /run/secrets/mysecret`
+
+Has the same effect as:
+
+`RUN --mount=type=secret,id=mysecret cat /run/secrets/mysecret`
+
 **--network**, **--net**=*mode*
 
 Sets the configuration for network namespaces when handling `RUN` instructions.
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -82,6 +82,7 @@ type executor struct {
 	runtime                        string
 	runtimeArgs                    []string
 	transientMounts                []Mount
+	transientRunMounts             []string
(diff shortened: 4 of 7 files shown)


aeijdenberg:
I found https://github.com/containers/buildah/issues/5987 today, who look to be trying solve a similar case to me. I'd like to set some transient related env vars and file mounts during the build phase that shouldn't affect the final image.

For example, this options allow me to run:

```bash
buildah build \
        --secret id=MAVEN_SETTINGS_FILE,type=file,src=/path/to/host/mvn.settings \
        --run-mount type=secret,id=MAVEN_SETTINGS_FILE,target=/usr/share/maven/conf/settings.xml,required \
...
```

(and related PR https://github.com/containers/buildah/pull/6285 to expose in an env variable to the build process)

aeijdenberg:
If interested, here's some context on how I'm using this - the intent to is make `buildah` work really well with HTTP proxies such that all assets pulled by `buildah` and `RUN`, `ADD` instructions are pulled through, ideally with no changes to the `Containerfile`, ie make it work transparently, including when TLS is inspected. The idea being that assets needed for a build can be saved and put-aside, then re-served later to support a future build.

Here's a full writeup of the tool I'm working on and info about the patches added to `buildah` to allow it to work in this context (all of which have PRs open here):

https://github.com/continusec/htvend/
https://github.com/continusec/htvend/blob/main/oci-image-building.md#no-method-to-pass-ssl_cert_file-to-run-instructions-in-dockerfile

aeijdenberg:
Rebased to fix merge conflict.

nalind:
Generally looks fine.  Flags, even when they're specific to RUN, don't tend to have a "run-" prefix for `buildah build` - they tend to just look like the do for `buildah run`.  This currently conflicts with that convention.
There's probably a missing assignment to the `flagCompletion` map in the `pkg/cli.GetBudFlagsCompletions()` function for this new flag.

aeijdenberg:
Thanks for the feedback.

> Generally looks fine. Flags, even when they're specific to RUN, don't tend to have a "run-" prefix for `buildah build` - they tend to just look like the do for `buildah run`. This currently conflicts with that convention.

I've renamed to `--mount` which matches the corresponding existing option for the `buildah run`.

> There's probably a missing assignment to the `flagCompletion` map in the `pkg/cli.GetBudFlagsCompletions()` function for this new flag.

Added.

imagebuildah/stage_executor.go
```diff
@@ -833,7 +833,7 @@ func (s *StageExecutor) Run(run imagebuilder.Run, config docker.Config) error {
 		NoPivot:              os.Getenv("BUILDAH_NOPIVOT") != "" || s.executor.noPivotRoot,
 		Quiet:                s.executor.quiet,
 		CompatBuiltinVolumes: types.OptionalBoolFalse,
-		RunMounts:            run.Mounts,
+		RunMounts:            append(slices.Clone(run.Mounts), s.executor.transientRunMounts...),
```
nalind: Non-blocking: this could use `slices.Concat()`.

nalind:
LGTM

nalind:
Ugh, I created merge conflicts with #6442.  If I can trouble you to rebase one more time, I'll go ahead and merge this.

imagebuildah/stage_executor.go
aeijdenberg: Thanks, fixed.

aeijdenberg:
> Ugh, I created merge conflicts with #6442. If I can trouble you to rebase one more time, I'll go ahead and merge this.

Done, and thanks for reviewing.

nalind:
/lgtm
Thanks!

aeijdenberg:
Rebased to fix merge conflict.

mmorhun:
@nalind @aeijdenberg is it possible to move this one forward?

aeijdenberg:
Rebased to fix merge conflict.

baude (approved):
LGTM

PR #6296, opened 2025-07-24
Exclude pulled up parent directories at commit-time

nalind:
#### What type of PR is this?

/kind api-change
/kind failing-test 

Have `copier.Ensure()` also return the parent directories of items that it created, along with information about them that can be used to filter them out of the layer at commit-time.

This modifies the signature of `copier.Ensure()`, but it was added between 1.40 and 1.41, and shouldn't (yet) have any external users.

When `copier.Ensure()` tells us about the parent directories of a mountpoint target that it created for us, add them to the list of items that we'll exclude from the commit unless something else causes them to be modified.

The "native" overlay diff method just walks the diff directory, so we can get directories in the committed layer that look exactly like the ones in the previous layer, depending on the storage driver.  And while it's barely noticeable, we can still control for it.

#### How to verify it

New integration test!

This should resolve inconsistencies in the "podman diff container and image with same name" test seen in https://github.com/containers/podman/pull/26666.

```release-note
None
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -479,6 +479,7 @@ func Put(root string, directory string, options PutOptions, bulkReader io.Reader
 // MkdirOptions controls parts of Mkdir()'s behavior.
 type MkdirOptions struct {
 	UIDMap, GIDMap []idtools.IDMap // map from containerIDs to hostIDs when creating directories
+	ModTimeNew     *time.Time      // set mtime and atime of newly-created directories
 	ChownNew       *idtools.IDPair // set ownership of newly-created directories
 	ChmodNew       *os.FileMode    // set permissions on newly-created directories
 }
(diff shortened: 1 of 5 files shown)


TomSweeneyRedHat:
LGTM
Did you get a chance to create a test PR in Podman and see if the tests pass with this in play?
@flouthoc or @Luap99 PTAL

nalind:
Anecdotes are not data, but I vendored this into a local copy of podman's main branch and then manually ran the test that was producing inconsistent results, with settings that would previously have reproduced the two sets of answers.

flouthoc (approved):
/lgtm
/approve

nalind:
Opened https://github.com/containers/podman/pull/26711.

nalind:
Oof, missed a case.  Opened #6302.

PR #6298, opened 2025-07-24
fix: simple non-scratch FROM image builds differ

aeijdenberg:
Pass through timestamp when copying existing image.

#### What type of PR is this?

/kind bug

This respects `--timestamp 0` when "tagging" an existing image by exporting to an oci-archive.

#### How to verify it

See bats test added.

None

```release-note
None
```


aeijdenberg:
This requires #6284 to be merged first - I'll add this commit to that PR (since it's related) and close this one out. Sorry for the noise.

PR #6300, opened 2025-07-25
Restore the default meaning of `--pull` (should be `always`).

### LINKED ISSUE #6266: Podman >= 5.2 changed the meaning of --pull (without argument)
### Issue Description

Hi,

It seems podman >= 5.2 did change the meaning of `podman build --pull` (without argument for `--pull`). It used to behave as `--pull=always` (like Docker does), and now it behaves as `--pull=missing`.

This issue was actually raised already in #5723, where @flouthoc already replied that it's an intentional change, but the pull request referenced in this comment https://github.com/containers/buildah/issues/5723#issuecomment-2695641234 is actually looping to the very same issue #5723. So in the end, I don't know the rational behind this ;) Knowing how important it is for the podman project to stick to docker behavior, I am really interested to understand why it was chosen on purpose to differ from docker here (and I can read in https://github.com/containers/podman/pull/20124 that before that, at least in 2020 and 2023, it was important for you that `--pull` actually meant `--pull=always`).

I am asking because internally in Amadeus this created some small incident today. We still use a mix of docker & podman, expecting both to work the same way, and... we discovered the hard way that `--pull` doesn't do what we expect anymore when using podman... (today, for an internal reason, some old internal images we had stopped working, and some of our developers were still using these old versions as --pull no longer automatically fetches the new ones) This change of behavior doesn't seem to be documented in the podman 5.2 release notes.

In the release notes for buildah 1.37.0, 3 changes mentions `--pull`: #5583, #5605 and #5407 However I don't see discussed in these pull requests or the associated issues the docker compatibility break. I "just" see documentation related motivations.

.

`podman build --pull` differs from `docker build --pull`.

`podman build --pull` used to behave like docker, and somehow I had hoped it continues to work the same way.

### buildah version output

```yaml
.
```

### buildah info output

```yaml
.
```

### Provide your storage.conf

```toml
.
```

### Upstream Latest Release

Yes

_No response_

### Additional information

_No response_

Romain-Geissler-1A:
#### What type of PR is this?

> /kind bug

Now the bud --help output reads this, correctly showing both "missing" and "always" defaults, if --pull is explicitly provided or not:

```
      --pull string[="always"]                       pull base and SBOM scanner images from the registry. Values:
                                                     always:  pull base and SBOM scanner images even if the named images are present in store.
                                                     missing: pull base and SBOM scanner images if the named images are not present in store.
                                                     newer:   only pull base and SBOM scanner images when newer images exist on the registry than those in the store. (default "missing")
```

Fixes #6266.

```release-note
Restore the meaning of `--pull` (without argument), similar to the docker behavior: now defaults to `--pull=always`
```

### DIFF
--- a/cmd/buildah/from.go
+++ b/cmd/buildah/from.go
@@ -75,7 +75,7 @@ missing: pull images if the named images are not present in store,
 never: only use images present in store if available,
 newer: only pull images when newer images exist on the registry than those in the store.`)
 
-	flags.Lookup("pull").NoOptDefVal = "true" // allow `--pull ` to be set to `true` as expected.
+	flags.Lookup("pull").NoOptDefVal = "always" // treat a --pull with no argument like --pull=always
 
 	flags.BoolVar(&opts.pullAlways, "pull-always", false, "pull the image even if the named image is present in store")
 	if err := flags.MarkHidden("pull-always"); err != nil {
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -794,7 +794,8 @@ The `buildah build` command allows building images for all Linux architectures, 
 
 **--pull**
 
-Pull image policy. The default is **missing**.
+Pull image policy. If not specified, the default is **missing**. If an explicit
+**--pull** argument is provided without any value, use the **always** behavior.
 
 - **always**: Pull base and SBOM scanner images from the registries listed in
 registries.conf.  Raise an error if a base or SBOM scanner image is not found
--- a/docs/buildah-from.1.md
+++ b/docs/buildah-from.1.md
@@ -388,7 +388,8 @@ the help of emulation provided by packages like `qemu-user-static`.
 
 **--pull**
 
-Pull image policy. The default is **missing**.
+Pull image policy. If not specified, the default is **missing**. If an explicit
+**--pull** argument is provided without any value, use the **always** behavior.
 
 - **always**: Pull base and SBOM scanner images from the registries listed in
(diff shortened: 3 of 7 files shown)


Romain-Geissler-1A:
> Can you reformat your commit message to not include the full --help output with all the extra spaces. That makes it harder to view the commit message.

Done, please check if the updated commit message is fine.

Romain-Geissler-1A:
In term of release cycle, how late are we to get this included in podman 5.6 ? Or, in my case, in the next RHEL podman (which I guess will be based on podman 5.6) ? We are using both RHEL 9 & 10 at the moment.

giuseppe (approved):
LGTM

Luap99 (approved):
LGTM

Luap99:
> In term of release cycle, how late are we to get this included in podman 5.6 ? Or, in my case, in the next RHEL podman (which I guess will be based on podman 5.6) ? We are using both RHEL 9 & 10 at the moment.

Well 1.41.0 is what will be in podman 5.6 so by that definition it is to late already.
However IMO this is a regressions and as such a bug fix so I see nor problem for us to cut a 1.41.1 with this fix and then vendor that into podman 5.6. The final podman 5.6 release is in two weeks so we still have some time for that.

cc @TomSweeneyRedHat @nalind

nalind:
I'd like to see this in a 1.41.1.
Thanks!
/lgtm

TomSweeneyRedHat:
Ooops, missed this until now.  @nalind TYVM for dropping this into 1.41.1

PR #6302, opened 2025-07-29
Only suppress "noted" items when not squashing

nalind:
#### What type of PR is this?

/kind bug

When suppressing what we think are pulled-up directories at commit-time, only do that when we're _not_ squashing the image, in which case we really do need to output it into the one layer that our output image will have.  This was a case I missed in #6296.

#### How to verify it

New integration test!

Fixes a test failure in podman's "zstd chunked does not modify image content" system test.

```release-note
None
```


TomSweeneyRedHat:
LGTM

Issue #6291, opened 2025-07-16
COPY and mount=type=bind cache invalidation


jsirex:
### Issue Description

There is inconsistency in behavior of `COPY` and `--mount=type=bind` regarding caching calculation.
Maybe this is "by design" or bug.

`COPY` uses file's permissions and content to validate checksum
`RUN --mount=type=bind`, like `COPY` but additionally takes into account file's modification time.

Practical case:
There is a pattern - multistage build: I copy sources, build, do stuff, and copy result from another stage (another `FROM`), basically throw away garbage.

Yet, this common approach, I find it quite not effective, when `bind mount` exist. So, instead of `COPY`ing, I:
mount sources, caches and whatever I want, do stuff with `RUN` and that is. 

But, because of timestamps, which are intentionally updated by design (for example git updates them), my `RUN` instruction get invalidated from cache.

### Steps to reproduce the issue

```Dockerfile
FROM debian:trixie

COPY Containerfile /copy-instruction
RUN --mount=type=bind,source=Containerfile,target=/bind-mount true
```

Build:
```
❯ podman build .
STEP 1/3: FROM debian:trixie
STEP 2/3: COPY Containerfile /copy-instruction
--> 287d22c5e68f
STEP 3/3: RUN --mount=type=bind,source=Containerfile,target=/bind-mount true
COMMIT
--> 080a50cf5996
080a50cf5996b4874fa15e732d0a5339c314cad2f310b6a4f9491b2c6068e961
```

Change timestamp
```
❯ touch Containerfile
❯ podman build .
STEP 1/3: FROM debian:trixie
STEP 2/3: COPY Containerfile /copy-instruction
--> Using cache 287d22c5e68f67aa1190d6b51def95bd670f34cb07b8504c1c56507c750268f4
--> 287d22c5e68f
STEP 3/3: RUN --mount=type=bind,source=Containerfile,target=/bind-mount true
COMMIT
--> efdd021c607c
efdd021c607c3911eecb3319c1b3fb731862e71cf805f2d6b759cfc36a8ea170
```

### Describe the results you received

Notice step 3 (RUN) has got invalidated

### Describe the results you expected

I don't know if there is any specification on this case does exist. Maybe we should ignore timestamps here too.

can't imaging case when someone needs to mount something, preserving timestamps exactly and invalidate cache, based on timestamp.

### podman version output

```yaml
Client:        Podman Engine
Version:       5.4.2
API Version:   5.4.2
Go Version:    go1.24.4
[...]

PR #6311, opened 2025-08-02
build: `--mount=type=bind` ignore `modTime` for cache candidates

flouthoc:
When generating a checksum for files mounted into container via `--mount=type=bind` ignore their `ModTime`, `AccessTime` and `ChangeTime` so we can maintain cache burst consistency with `COPY` command.

Closes: https://github.com/containers/buildah/issues/6291

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
--mount=type=bind now does ignores modtime when looking for cache candidates.
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -662,6 +662,38 @@ _EOF
   expect_output --substring "Cache burst add diff"
 }
 
+@test "bud --layers with --mount type bind should preserve cache when file mod time changes but content stays same" {
+  _prefetch alpine
+  local contextdir=${TEST_SCRATCH_DIR}/bud/platform
+  mkdir -p $contextdir
+
+  cat > $contextdir/samplefile << _EOF
+samplefile content unchanged
+_EOF
+
+  cat > $contextdir/Containerfile << _EOF
+FROM alpine
+RUN --mount=type=bind,source=samplefile,target=file,Z cat file
+_EOF
+
+  # on first run since there is no cache so content must be printed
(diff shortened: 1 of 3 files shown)


tests/bud.bats
```diff
@@ -662,6 +662,41 @@ _EOF
...
+  # on second run since there is cache so content should not be printed
+  run_buildah build $WITH_POLICY_JSON --layers -t source -f $contextdir/Containerfile $contextdir
+  # output should not contain content from the file since entire build is cached
+  assert "$output" !~ "samplefile content unchanged"
+
+  # Remove and recreate the file with same content (this changes modification time)
+  rm $contextdir/samplefile
+  cat > $contextdir/samplefile << _EOF
+samplefile content unchanged
+_EOF
```
nalind: Using `touch $contextdir/samplefile` would be simpler here, though if it wasn't being zeroed out, it wouldn't guarantee that the timestamp had changed enough to affect the digest of the tarstream that would be generated, so the test might end up passing anyway.  Perhaps `touch -d @...` after creating the file, and then again with a different date to "update" it?

flouthoc: I updated the PR.

nalind:
LGTM

TomSweeneyRedHat:
LGTM

PR #6315, opened 2025-08-04
History should note unset-label, timestamp, and rewrite-timestamp

nalind:
#### What type of PR is this?

/kind bug

Make the history created-by field note when a label was unset, or if the one of the CLI flags that modifies the timestamps in the layer were used, so that the cache evaluation logic can factor that into account.

#### How to verify it

New integration test!

None

```release-note
None
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -8152,6 +8152,158 @@ EOF
   assert "$output" = "$datestamp" "SOURCE_DATE_EPOCH build arg didn't affect image config creation date"
 }
 
+@test "bud source-date-epoch and timestamp inform cache checks" {
+  _prefetch busybox
+  local contextdir=${TEST_SCRATCH_DIR}/context
+  mkdir $contextdir
+
+  # We'll build a multiple-layer image based on busybox.
+  cat > $contextdir/Dockerfile <<-EOF
+  FROM busybox
+  ARG SOURCE_DATE_EPOCH
+  COPY Dockerfile /root/Dockerfile
+  RUN pwd > /pwd.txt
+EOF
+
+  run_buildah build --layers=true --iidfile ${TEST_SCRATCH_DIR}/baseiidfile ${contextdir}
+  local baseiid=$(cat ${TEST_SCRATCH_DIR}/baseiidfile) iid
+
+  # Try an assortment of layer-mutating flags.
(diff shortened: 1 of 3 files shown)


tests/bud.bats
```diff
@@ -8152,6 +8152,51 @@ EOF
...
+
+  run_buildah build --layers=true --iidfile ${TEST_SCRATCH_DIR}/iidfile --timestamp=86400 ${contextdir}
+  iid=$(cat ${TEST_SCRATCH_DIR}/iidfile)
+  assert "$iid" == "$iidts"
+
+  run_buildah build --layers=true --iidfile ${TEST_SCRATCH_DIR}/iidfile --source-date-epoch=86400 --rewrite-timestamp ${contextdir}
+  iid=$(cat ${TEST_SCRATCH_DIR}/iidfile)
+  assert "$iid" == "$iidrt"
+}
+
```
TomSweeneyRedHat, line 8199: Given the additions to  unsetLabels and  layerMutations in stage_executor, do we need any  tweaks to tests for the new returns there?

nalind: We didn't have coverage for --unsetannotation before, added a test that exercises its impact on caching, along with --unsetlabel, --inherit-labels=false, and --inherit-annotations=false.

TomSweeneyRedHat:
Unhappy tests now.
The new test changes LGTM on a quick peak.

nalind:
@TomSweeneyRedHat looks like they're flakes.

TomSweeneyRedHat:
LGTM
@flouthoc or @Luap99 PTAL

tests/bud.bats
```diff
@@ -8152,6 +8152,158 @@ EOF
...
+  # Build an image with a label and an annotation that we don't want in the end.
+  run_buildah build --layers=true --annotation=annotation1=annotationvalue1 --annotation=annotation2=annotationvalue2 --label=label1=labelvalue1 --label=label2=labelvalue2 --iidfile ${TEST_SCRATCH_DIR}/baseiidfile ${contextdir}
+  local baseiid=$(cat ${TEST_SCRATCH_DIR}/baseiidfile) iid
+  baseiid="${baseiid##*:}"
+  assert "${baseiid}" != ""
+  # Double-check that the label and annotation made it in.
+  run_buildah inspect -t image "${baseiid}"
+  run jq -r .ImageAnnotations <<< "${output}"
+  assert "${status}" == 0
```
Luap99, line 8220: FYI you can do `run -0 command` to check the exit code for the command so the extra status check would not be needed.

PR #6328, opened 2025-08-08
Losen the dependency on go-connections/tlsconfig

mtrmac:
#### What type of PR is this?

/kind cleanup

As of go-connections 0.6.0, there is no difference between "server" and "client" configuration; so, take this opportunity to switch from "server" defaults in client contexts, which were used for an unknown reason.

Also, don't use the whole tlsconfig.ClientDefault(), to _slightly_ move into the direction of using the Go built-in defaults, and to get a _bit_ closer to getting out of the business of managing TLS policy.

Should not change behavior (but the update from go-connections 0.5.0 to 0.6.0 did change the list of accepted TLS 1.2 cipher suites).

#### How to verify it

Reading the go-connections code

None

```release-note
None
```

### DIFF
--- a/add.go
+++ b/add.go
@@ -144,7 +144,12 @@ func getURL(src string, chown *idtools.IDPair, mountpoint, renameTarget string, 
 		return err
 	}
 	tlsClientConfig := &tls.Config{
-		CipherSuites: tlsconfig.DefaultServerAcceptedCiphers,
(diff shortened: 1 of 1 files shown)


nalind:
LGTM

TomSweeneyRedHat:
LGTM
Thanks!

nalind:
For some reason I thought this label was already there.
/approve

Issue #6327, opened 2025-08-08
expand build arguments in log


vrothberg:
### Issue Description

Buildah/Podman does not expand build arguments in the logs.  The build logs/steps display the unexpanded contents of the Containerfile instructions, e.g., `[1/2] STEP 5/9: RUN mkdir $NEWROOT`.

Docker/Buildkit does the expansions (e.g., `[builder 2/5] RUN mkdir /new-root-fs`) which is extremely helpful to debug/analyze if the expansion during the build process works as expected.

I would love Buildah to expand the variables as well.  I couldn't wrap my head around an issue where the build succeeded but didn't provide the build artifact I expected.  Trying `docker build` helped me debug the issue and correct the Containerfile.

Cc: @nalind @flouthoc 

### Steps to reproduce the issue

Steps to reproduce the issue
```
[root@lima-fedora tmp.cGoHObl9M4]# cat Containerfile
FROM fedora:rawhide
ARG FOO=foo
RUN echo $FOO
[root@lima-fedora tmp.cGoHObl9M4]# buildah build -f Containerfile .
STEP 1/3: FROM fedora:rawhide
STEP 2/3: ARG FOO=foo
STEP 3/3: RUN echo $FOO
[...]
[root@lima-fedora tmp.cGoHObl9M4]# docker build -f Containerfile .
[..]                                                                                                0.0s
 => CACHED [1/2] FROM docker.io/library/fedora:rawhide@sha256:af660adb8de42682b494c7d102860c155466ba61cd850a8420e1ffad1764b1b7        0.0s
 => [2/2] RUN echo foo                                                                                                                0.1s
[...]
```

### Describe the results you received

RUN echo $FOO

### Describe the results you expected

RUN echo foo

### buildah version output

```yaml
Version:         1.40.1
Go Version:      go1.24.3
Image Spec:      1.1.1
Runtime Spec:    1.2.1
CNI Spec:        1.1.0
libcni Version:
image Version:   5.35.0
Git Commit:
Built:           Thu Jun  5 16:13:27 2025
OS/Arch:         linux/arm64
BuildPlatform:   linux/arm64/v8
```

### buildah info output

```yaml
{
    "host": {
        "CgroupVersion": "v2",
        "Distribution": {
            "distribution": "fedora",
            "version": "42"
        },
        "MemFree": 174956544,
        "MemTotal": 4082163712,
        "OCIRuntime": "crun",
        "SwapFree": 4082102272,
        "SwapTotal": 4082102272,
[...]

PR #6329, opened 2025-08-09
executor: expand args in build logs

flouthoc:
Expand user,builtin and heading args before logging build steps.

Closes: https://github.com/containers/buildah/issues/6327

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
executor: expand args in build logs
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -5425,10 +5425,11 @@ EOM
 
   ${OCI} --version
   _prefetch alpine busybox
-
-  run_buildah build --build-arg base=alpine --build-arg toolchainname=busybox --build-arg destinationpath=/tmp --pull=false $WITH_POLICY_JSON -f $BUDFILES/from-with-arg/Containerfile .
+  toolchainname=busybox
+  destinationpath=/tmp
+  run_buildah build --build-arg base=alpine --build-arg toolchainname=$toolchainname --build-arg destinationpath=$destinationpath --pull=false $WITH_POLICY_JSON -f $BUDFILES/from-with-arg/Containerfile .
   expect_output --substring "FROM alpine"
-  expect_output --substring 'STEP 4/4: COPY --from=\$\{toolchainname\} \/ \$\{destinationpath\}'
+  expect_output --substring "STEP 4/4: COPY --from=$toolchainname \/ $destinationpath"
   run_buildah rm -a
 }
 
@@ -6186,7 +6187,7 @@ _EOF
   # must push cache twice i.e for first step and second step
   run printf "STEP 2/3: ARG VAR=hello\n--> Pushing cache"
   step1=$output
-  run printf "STEP 3/3: RUN echo \"Hello \$VAR\""
+  run printf "STEP 3/3: RUN echo \"Hello hello\""
   step2=$output
   run printf "Hello hello"
   step3=$output


flouthoc:
@vrothberg @nalind PTAL

vrothberg (commented):
Changes and tests LGTM.

I didn't check, but maybe there's a similar behavior with ENV?

Thanks so much for the quick PR 🙏

imagebuildah/executor.go
```diff
@@ -600,7 +600,19 @@ func (b *Executor) buildStage(ctx context.Context, cleanupStages map[int]*StageE
...
+			builtinArgs := argsMapToSlice(stage.Builder.BuiltinArgDefaults)
+			headingArgs := argsMapToSlice(stage.Builder.HeadingArgs)
+			userArgs := argsMapToSlice(stage.Builder.Args)
+			userArgs = append(builtinArgs, append(userArgs, headingArgs...)...)
+			argsCopy := args
+			for i := range args {
+				argsCopy[i], err = imagebuilder.ProcessWord(args[i].(string), userArgs)
+				if err != nil {
+					argsCopy[i] = args[i]
+					fmt.Printf("while replacing arg variables with values for format in logging %q: %w\n", args[i], err)
```
nalind: Why is this error message sent to stdout rather than one of the output destinations that's been specified for logging?

imagebuildah/executor.go
```diff
@@ -600,7 +600,19 @@ func (b *Executor) buildStage(ctx context.Context, cleanupStages map[int]*StageE
 				prefix += ": "
 			}
 			suffix := "\n"
-			fmt.Fprintf(stageExecutor.executor.out, prefix+format+suffix, args...)
+			builtinArgs := argsMapToSlice(stage.Builder.BuiltinArgDefaults)
+			headingArgs := argsMapToSlice(stage.Builder.HeadingArgs)
+			userArgs := argsMapToSlice(stage.Builder.Args)
+			userArgs = append(builtinArgs, append(userArgs, headingArgs...)...)
+			argsCopy := args
```
nalind: If this is meant to create a copy of `args`, use `slices.Clone()`, otherwise they're both still pointing at the same array.

tests/bud.bats
```diff
@@ -8844,3 +8844,42 @@ _EOF
...
+
+  # Test that unexpanded variables show as empty when no build-arg is provided
+  cat > $contextdir/Containerfile2 << _EOF
+FROM alpine
+ARG UNDEFINEDVAR
+RUN echo \$UNDEFINEDVAR
+_EOF
+
+  run_buildah build $WITH_POLICY_JSON -t ${target}-undefined -f $contextdir/Containerfile2 $contextdir
+  expect_output --substring "RUN echo "
```
nalind: This would match whether a value was printed or not.

flouthoc:
> This produces some weird corner cases, both with this patch and with `docker build`:
> 
> ```dockerfile
> FROM busybox
> ARG CMD=pwd
> RUN CMD=ls; echo $CMD
> ```

Similar behavior exists in `buildkit` as well I think this is bug in both `buildah` and `docker`. Should we fix it here ?

imagebuildah/executor.go
```diff
@@ -600,7 +600,19 @@ func (b *Executor) buildStage(ctx context.Context, cleanupStages map[int]*StageE
 				prefix += ": "
 			}
 			suffix := "\n"
-			fmt.Fprintf(stageExecutor.executor.out, prefix+format+suffix, args...)
+			builtinArgs := argsMapToSlice(stage.Builder.BuiltinArgDefaults)
+			headingArgs := argsMapToSlice(stage.Builder.HeadingArgs)
+			userArgs := argsMapToSlice(stage.Builder.Args)
+			userArgs = append(builtinArgs, append(userArgs, headingArgs...)...)
+			argsCopy := slices.Clone(args)
```
Luap99: why is this cloning the full args array if we throw away all values, that is just causing unnecessary memory copies.
To preallocate use make([]string, 0, len(args)).

imagebuildah/executor.go
```diff
@@ -600,7 +600,19 @@ func (b *Executor) buildStage(ctx context.Context, cleanupStages map[int]*StageE
...
+			for i := range args {
+				argsCopy[i], err = imagebuilder.ProcessWord(args[i].(string), userArgs)
```
Luap99: casting to string seems unsafe, the function signature accepts any, This means all non strings values are discarded basically.
If the log can only deal with string the function signature should be updated to only accept strings.

In principle I am also quite confused in how many different places ProcessWord() gets called, buildah seems to call this and replace values as one off in may palaces. I think it would be much cleaner to replace and and then keep the replaced values but I don't understand enough about the buildah design here.

imagebuildah/executor.go
```diff
@@ -600,7 +600,19 @@ func (b *Executor) buildStage(ctx context.Context, cleanupStages map[int]*StageE
...
+				if err != nil {
+					argsCopy[i] = args[i]
+					fmt.Fprintf(stageExecutor.executor.err, "while replacing arg variables with values for format in logging %q: %w\n", args[i], err)
```
Luap99: you cannot use %w for normal formatting, only in fmt.Errorf()

imagebuildah/executor.go
nalind: Am I looking at the right function?  `ProcessWord()` appears to take a string and a slice of strings, so type assertions shouldn't even be necessary.

Luap99: I am talking about the log function itself which defines args as `[]any` type. Everything that gets passed which is not a string is basically lost here due this type cast.

nalind: Ah, I follow now.  The assertion isn't checked, so I think that case would panic.  In practice, we only pass it strings.

flouthoc: I am processing only args with `string` type, I avoided changing function signature as it involves making changes in public API of some structs so will do that in a new PR. Practically all args are `strings` so this check is just preventing from `panic`.

imagebuildah/executor.go
flouthoc: Done.

imagebuildah/executor.go
flouthoc: Done.

flouthoc:
> > > This produces some weird corner cases, both with this patch and with `docker build`:
> > > ```dockerfile
> > > FROM busybox
> > > ARG CMD=pwd
> > > RUN CMD=ls; echo $CMD
> > > ```
> > 
> > 
> > Similar behavior exists in `buildkit` as well I think this is bug in both `buildah` and `docker`. Should we fix it here ?
> 
> You don't event need build args to trigger this: `RUN test=1; echo ${test:-abc}`
> 
> And worse if you just have an unset non existing build arg buildah just removes from the log (uses empty string) which is also super confusing,
> 
> The issue states
> 
> > I would love Buildah to expand the variables as well. I couldn't wrap my head around an issue where the build succeeded but didn't provide the build artifact I expected. Trying docker build helped me debug the issue and correct the Containerfile.
> 
> Given that I would argue this change doesn't really improve the situation, it trades one problem for another. But if docker behaves like that and logs incorrectly then maybe matching its bugs makes sense? I don't personally care much either way.
> 
> But I think the correct thing is to only replace the vars that buildah actually replaced when executing the command which ties back to my comment about replacing the RUN string in to many different places.

I am keeping behavior same as buildkit for consistency and created an upstream issue with moby/buildkit#6139 to track it there.

imagebuildah/executor.go
```diff
@@ -600,7 +607,26 @@ func (b *Executor) buildStage(ctx context.Context, cleanupStages map[int]*StageE
...
+						processedArg, err := imagebuilder.ProcessWord(arg, userArgs)
+						if err != nil {
+							processedArg = arg
+							fmt.Fprintf(stageExecutor.executor.err, "while replacing arg variables with values for format in logging %q: %v\n", arg, err)
+						}
+						processedArgs = append(processedArgs, processedArg)
+					} else {
+						processedArgs = append(processedArgs, arg)
+					}
+				}
```
Luap99, line 627: That still discards any arg which is not a string, the else block must be one lever out because it needs to keep the non string args otherwise the formatting string will be messed up badly.

imagebuildah/executor.go
```diff
@@ -519,6 +520,12 @@ func (b *Executor) getImageTypeAndHistoryAndDiffIDs(ctx context.Context, imageID
 	return oci.OS, oci.Architecture, manifestFormat, oci.History, oci.RootFS.DiffIDs, nil
 }
 
+// Checks if is word is using any variable in the form of `$VAR`, `${VAR}` etc
+func containsVariable(arg string) bool {
+	re := regexp.MustCompile(`\$\{?[a-zA-Z_][a-zA-Z0-9_]*\}?`)
+	return re.MatchString(arg)
```
Luap99, line 526: That compiles the regex every single time which is quite inefficient. That should use something like regexp.Delayed() in the global scope.

But is the regex really right? At the very least it ignores escaping rules `\$var` is not a variable. ProcessWord() has already has a parser so this would need to match exactly what it is doing. I think it would be better to reuse the ProcessWord code and expose some function for this instead, yeah sure more complicated but the only sane way to avoid any differences between the var replacement code (ProcessWord) and this custom regex

PR #6332, opened 2025-08-11
Rework how we decide what to filter out of layer diffs

nalind:
#### What type of PR is this?

/kind bug

After narrowing down the list of parent directories which we might need to exclude to those which are present in the base image, filter them out of the layer diff as it is generated.

#### How to verify it

Updated integration test!
Updated unit tests for supporting functions!

Follows up on #6296/#6306 and #6302/#6319.

This one was hard to find because "--squash" does different things in `podman build` than it does in `buildah build`, so podman's attempt to run our tests ended up testing a combination of options there that we weren't testing here.

```release-note
None
```


TomSweeneyRedHat:
@nalind and @mheon is this one we should try to get into Podman v5.6?

mheon:
I lean towards "not 5.6.0 but maybe 5.6.1"

TomSweeneyRedHat:
LGTM
@Luap99 and @flouthoc PTAL

Issue #4010, opened 2022-05-20
Building Windows container images


sebsoto:
**Desired behavior**:
I would expect that when no `RUN` commands are present in a container file, `buildah build --os=windows` would be able to build a working Windows container image.

In my testing I have found that buildah does not create functional Windows container images.
This seems to be due to at least two reasons:

1) The [required tar headers specified by the OCI image-spec](https://github.com/opencontainers/image-spec/blob/main/layer.md#platform-specific-attributes) are not being applied. 
([image-spec issue with slightly more info](https://github.com/opencontainers/image-spec/issues/415))
2) The rootfs of Windows container images is actually the `Files/` directory. This can be seen by pulling a Windows container image and inspecting the layers.

There may be other reasons, but if there are, they are not documented.

**Current state:**
I have only tested this by building an image with a Linux build of buildah, with the image being used by containerd on a Windows VM.

I built two images, one using docker buildx, the other with buildah:

Using the following containerfile:
```
FROM mcr.microsoft.com/windows/nanoserver:ltsc2022
ADD emptyfile .
```
An image built using docker buildx works with no issue, containerd logs:
```
time="2022-05-20T16:46:52.533976500+02:00" level=info msg="PullImage \"quay.io/ssoto/test:nanoBuildx\""
time="2022-05-20T16:47:07.411314400+02:00" level=info msg="ImageCreate event &ImageCreate{Name:quay.io/ssoto/test:nanoBuildx,Labels:map[string]string{io.cri-containerd.image: managed,},XXX_unrecognized:[],}"
time="2022-05-20T16:47:07.422273200+02:00" level=info msg="ImageCreate event &ImageCreate{Name:sha256:43aba2088aa29238b8dc6ce66f9fdea52bc9401da781c03dc7c0bbfb5e9b0b67,Labels:map[string]string{io.cri-containerd.image: managed,},XXX_unrecognized:[],}"
time="2022-05-20T16:47:07.430670500+02:00" level=info msg="ImageUpdate event &ImageUpdate{Name:quay.io/ssoto/test:nanoBuildx,Labels:map[string]string{io.cri-containerd.image: managed,},XXX_unrecognized:[],}"
[...]

PR #6334, opened 2025-08-12
[WIP] Allow windows layer creation

sebsoto:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
/kind feature
> /kind flake
> /kind other

Fixes https://github.com/containers/buildah/issues/4010

```release-note

```


image.go
```diff
@@ -1097,6 +1110,113 @@ func (i *containerImageRef) NewImageSource(_ context.Context, _ *types.SystemCon
...
+
+	if !h.ModTime.IsZero() {
+		h.PAXRecords[keyCreationTime] = fmt.Sprintf("%d.%d", h.ModTime.Unix(), h.ModTime.Nanosecond())
+	}
+
+	h.Format = tar.FormatPAX
+}
+
+func addSecurityDescriptor(h *tar.Header) {
+	if h.Typeflag == tar.TypeDir {
```
mheon, line 1141: You're going to need some very good comments about what these are, what they do, and how to re-generate them because these gigantic (base64?) constants worry me more than a little for future maintenance

common.go
```diff
@@ -36,6 +36,9 @@ func getCopyOptions(store storage.Store, reportWriter io.Writer, sourceSystemCon
 	destinationCtx := getSystemContext(store, nil, "")
 	if destinationSystemContext != nil {
 		*destinationCtx = *destinationSystemContext
+		if destinationCtx.OSChoice == "windows" {
+			sourceCtx.OSChoice = "windows"
+		}
```
mtrmac, line 41: This is almost certainly the wrong location for this logic.

Where is the _actual_ disconnect between users’ intent and in-process data structures coming from?

image.go
```diff
@@ -1014,7 +1014,16 @@ func (i *containerImageRef) NewImageSource(_ context.Context, _ *types.SystemCon
 			rc.Close()
 			return nil, fmt.Errorf("compressing %s: %w", what, err)
 		}
-		writer := io.MultiWriter(writeCloser, srcHasher.Hash())
+		var writer io.Writer
+		var discard func(error)
+		var done chan error
+		windowsLayer := sys.OSChoice == "windows"
```
mtrmac, line 1020: That’s just not what a `SystemContext.OSChoice` means.

OSChoice is a per-_operation_ option, choosing a _platform_. (And, BTW, it can be — and usually is — unset.) In the case of `NewImageSource`, it sort of means “when in doubt, _choose to read_ the for-Windows option”.

This seems to suggest a model where “the same” c/storage can contain layers that are both “Windows” layers and others. IOW, isn’t that a per-layer option? Should that be recorded somewhere in the storage?

Or… alternatively — isn’t this a per-_destination_ option? Could I create an arch-neutral image (maybe just to store data, with no executables), which should be pushed _differently_ based on whether I’m committing it to be consumed by Linux / by Windows?

What is the actual semantics?

image.go
```diff
@@ -1097,6 +1110,113 @@ func (i *containerImageRef) NewImageSource(_ context.Context, _ *types.SystemCon
...
+		h.PAXRecords[keySDRaw] = "AQAEgBQAAAAkAAAAAAAAADAAAAABAgAAAAAABSAAAAAgAgAAAQEAAAAAAAUSAAAAAgCoAAcAAAAAAxgA/wEfAAECAAAAAAAFIAAAACACAAAAAxQA/wEfAAEBAAAAAAAFEgAAAAAAGAD/AR8AAQIAAAAAAAUgAAAAIAIAAAALFAAAAAAQAQEAAAAAAAMAAAAAAAMYAKkAEgABAgAAAAAABSAAAAAhAgAAAAIYAAQAAAABAgAAAAAABSAAAAAhAgAAAAIYAAIAAAABAgAAAAAABSAAAAAhAgAA"
+	}
+
+	if h.Typeflag == tar.TypeReg {
+		// O:BAG:SYD:(A;;FA;;;BA)(A;;FA;;;SY)(A;;0x1200a9;;;BU)
+		h.PAXRecords[keySDRaw] = "AQAEgBQAAAAkAAAAAAAAADAAAAABAgAAAAAABSAAAAAgAgAAAQEAAAAAAAUSAAAAAgBMAAMAAAAAABgA/wEfAAECAAAAAAAFIAAAACACAAAAABQA/wEfAAEBAAAAAAAFEgAAAAAAGACpABIAAQIAAAAAAAUgAAAAIQIAAA=="
+	}
+}
+
+func makeWindowsLayer(w io.Writer) (io.Writer, func(error), chan error) {
```
mtrmac, line 1152: - Conceptually, what is this _doing_ ? Are those flags parts of the layer _contents_, such that users might plausibly want some _different_ values? If so, that shouldn’t happen here at commit / push time, but as part of layer creation, and it should “only” be converted from on-disk data to tar.

- Or are they… just an expected Windows-specific extension of the tar format, making it mandatory to include these headers, but without carrying much of a semantic information?

I see https://github.com/opencontainers/image-spec/blob/main/layer.md#platform-specific-attributes , but that doesn’t strongly answer my question — if anything, I read it closer to the first alternative.

mtrmac (commented):
As a first guess, I don’t know why we would do a layer transformation like this here, and not teach the original code producing the tar stream (somewhere in c/storage) to add the headers immediately. It’s quite possible that there’s a good answer. Structurally…

---

> `Digest did not match, expected `

… I think this might be running into a significant structural assumption: A c/storage layer has _one_ canonical tar representation, and that can be consistently re-created, and it’s an input into the layer’s _identity_.

A layer that can be read two different ways creating two different tarballs just isn’t in the model. If anything, those kinds of conversions are more of a c/image problem space (… but that, until now, ~never dealt with the tar contents, only with compression / decompression).

I think we need a fairly specific discussion about the long-term feature set.
- Are we adding the feature to construct Windows layers, with all relevant Windows attributes, to c/storage and Buildah, _while storing the data on Unix_?
- Are we adding a very limited local hack that allows creating Windows-consumable tarballs, without much of a Windows feature support otherwise? (There could be several location where to put such a hack, this is not obviously a correct one.)
- Do we _want_ to do the “construct Windows layers with all relevant attributes” feature set, but are we just adding a short-term hack that we would ideally want to go away, and which should, therefore, not commit us to any UI or backend storage decisions, if possible?

As a minimal starting point to a conversation, what happens, and what should happen, if I do `FROM some-remote-image` with some _other_ values of these tar headers?

sebsoto:
@mtrmac 
Thanks for the feedback on the approach.
I'll rework this based on your comments.

PR #6340, opened 2025-08-14
Mount image fails with imagestores

nalind:
#### What type of PR is this?

/kind bug

Fall back to creating a new builder with it if MountImage() fails on the base image, because when the store is configured with its "imagestore" option, that can happen.

#### How to verify it

New integration test.

None

Podman's e2e tests exercise the imagestore storage option, and we haven't previously.

```release-note
None
```


TomSweeneyRedHat:
LGTM
@flouthoc @Luap99 PTAL

Issue #6205, opened 2025-06-04
CI: runc cgroup error msg flake


Luap99:
```
[+0588s] not ok 561 copy-file-relative-context-dir
[+0588s] # (from function `expect_line_count' in file ./helpers.bash, line 591,
[+0588s] #  in test file ./copy.bats, line 597)
[+0588s] #   `expect_line_count 1' failed
[+0588s] # /var/tmp/go/src/github.com/containers/buildah/tests /var/tmp/go/src/github.com/containers/buildah/tests
[+0588s] # # [checking for: docker.io/library/busybox]
[+0588s] # # [restoring from cache: /tmp/bats-run-lPAMEB/suite/buildah-image-cache / docker.io/library/busybox]
[+0588s] # Getting image source signatures
[+0588s] # Copying blob sha256:9758c28807f21c13d05c704821fdd56c0b9574912f9b916c65e1df3e6b8bc572
[+0588s] # Copying config sha256:f0b02e9d092d905d0d87a8455a1ae3e9bb47b4aa3dc125125ca5cd10d6441c9f
[+0588s] # Writing manifest to image destination
[+0588s] # # /var/tmp/go/src/github.com/containers/buildah/tests/./../bin/buildah from --quiet --signature-policy /var/tmp/go/src/github.com/containers/buildah/tests/./policy.json busybox
[+0588s] # busybox-working-container
[+0588s] # # /var/tmp/go/src/github.com/containers/buildah/tests/./../bin/buildah copy --contextdir /tmp/buildah_tests.ishht1/context busybox-working-container test_file /opt/
[+0588s] # 42145a076d5e72262bea80733dac7785067a42f76c7b45a5a047f70e4403440f
[+0588s] # # /var/tmp/go/src/github.com/containers/buildah/tests/./../bin/buildah run busybox-working-container ls -1 /opt/
[+0588s] # test_file
[+0588s] # time="2025-06-04T04:16:33-05:00" level=error msg="seek /sys/fs/cgroup/system.slice/runc-buildah-buildah4055347767.scope/cgroup.freeze: no such device"
[+0588s] # #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
[+0588s] # #| FAIL: buildah run busybox-working-container ls -1 /opt/
[+0588s] # #| Expected 1 lines of output, got 2
[+0588s] # #| Output was:
[+0588s] # #| >test_file
[...]

PR #6366, opened 2025-09-03
Bump c/image to v5.36.2, c/common v0.64.2, ulinkunitz/xz v0.5.12, Buildah to v1.41.2

TomSweeneyRedHat:
This bump c/image to v5.36.2, c/common to v0.64.2, ulinkunitz/xz to v0.5.12, and finally Buildah to v1.41.2.  

All of this in preparation for Podman v5.6.1

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```


TomSweeneyRedHat:
@nalind @flouthoc @Luap99 PTAL

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 # Changelog
 
+## v1.41.2 (2025-09-03)
```
Luap99: That seems to be the wrong version, there is already a v1.41.3 so the next version must be .4

nalind:
There's a github.com/docker/docker update to 28.3.3+incompatible that we should probably pull in, even though we don't use its networking.

Luap99:
> There's a github.com/docker/docker update to 28.3.3+incompatible that we should probably pull in, even though we don't use its networking.

The CVE-2025-54388 fix is in docker specific code that we don't use, it has nothing to do with buildah or podman so I really don't think we need this at all

CHANGELOG.md
TomSweeneyRedHat: OH CRUD!  I'd not updated my notes after those other bumps and didn't double check this morning.  I'll go redo.

TomSweeneyRedHat:
Repushed with the right version number.  Since I was in the neighborhood, I thought I'd be safe rather than sorry and dinged Docker too.

TomSweeneyRedHat:
The runc failures are a bit concerning, I'm not sure what would be causing those to fall over, and the expected/actual values look to be the same?

Luap99:
runc are just flakes https://github.com/containers/buildah/issues/6205

PR #6378, opened 2025-09-09
chroot: use $PATH when finding commands

nalind:
#### What type of PR is this?

/kind bug

Use the $PATH from the runtime config when the command to run isn't an absolute path and the command isn't being processed by the shell.

#### How to verify it

New integration test!

Fixes https://github.com/containers/podman/discussions/27024.

```release-note
The chroot isolation handler should now correctly use $PATH to find the binary to run when RUN instructions express the command to run in exec form, and which do not do so using an absolute path.
```

### DIFF
--- a/chroot/run_common.go
+++ b/chroot/run_common.go
@@ -12,6 +12,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


chroot/run_common.go
```diff
@@ -743,6 +744,15 @@ func runUsingChrootExecMain() {
 		os.Exit(1)
 	}
 
+	// Set $PATH to the value for the container, so that when args[0] is not an absolute path,
+	// exec.Command() can find it using exec.LookPath().
+	reversedEnv := slices.Clone(options.Spec.Process.Env)
+	slices.Reverse(reversedEnv)
+	envPathIndex := slices.IndexFunc(reversedEnv, func(s string) bool { return strings.HasPrefix(s, "PATH=") })
```
Luap99: nit, with go iterators we can actually skip the copy + reversing.

something like this I think:
```
for _, env := range slices.Backward(options.Spec.Process.Env) { 
		if strings.HasPrefix(env, "PATH=") {
			os.Setenv()
			break
		}
	}
```

nalind: Changing it.  I need to get more used to using iterators.

nalind: `modernize` would have changed this to use `strings.CutPrefix()`, so just going to do that here.

Luap99 (approved):
/lgtm

PR #6386, opened 2025-09-16
.cirrus.yml: Test Vendoring bump golang

flouthoc:
Fixes issue caught in https://github.com/containers/buildah/pull/6385

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```


flouthoc:
@nalind @Luap99 PTAL

.cirrus.yml
```diff
@@ -124,7 +124,7 @@ vendor_task:
 
     # Runs within Cirrus's "community cluster"
     container:
-        image: docker.io/library/golang:1.24.0
+        image: docker.io/library/golang:1.24.7
```
Luap99: would it not be better to just use the `1.24` tag without patch release instead? That way it should always use the latest which means we don't have to update this manually so often.

nalind: This value exactly matches what's in go.mod.

Luap99: right but we get bumps like https://github.com/containers/buildah/pull/6385 where dependencies like to bump to go 1.24.2 for whatever reasons... so then it means it requires a manual update here.

nalind: Why isn't the bot updating this part, too?

Luap99: well you could make the action do that, but I don't think you can teach renovate to do that which is what I would care about mainly?

nalind: The last several PRs that renovate opened didn't need such a change added to them.  If the bot had done it in #6385, I don't know that we'd even be talking about it.

Luap99: Well the change is added when said dependency declares such go version. So if renovate would open the go.podman.io/common bump for example it would run into the exact same issue.

nalind: But why wouldn't we want to have podmanbot handle that when podmanbot's doing the work?  If renovate had an option for it, we'd be using it, but the point of it is about handling cases that renovate doesn't, right?

Luap99: I mean you can make the podmanbot do that I guess, but these are only DNM test PRs so they never get merged. For actual vendor updates that is still done by renovate or well manaully.

Regardless I guess it is just a minor thing anyway, the times such a bump happens is rare. If we now just do a PR here once that updates common/image/storage and bump the version here it will work again just fine

flouthoc: I will rebase this PR with version changed to `1.24` once @nalind agrees.

nalind: I thought renovate couldn't handle the layout of the container-libs repository (it hasn't opened one for any of the components we get from it), and keeping the dependency updates automated was part of the motivation for podmanbot.
The do-not-merge bit isn't only being set because it's still being debugged?

Luap99: renovate cannot handle proper updates in the container-libs repo, for updates here it should work fine in theory. Though we still have to bump it to tagged versions once to make the go pseudo version work.

The point of podmanbot is to vendor every PR as WIP PR here and ensure all buildah tests still pass not to merge them

Luap99: @flouthoc Make a commit to vendor latest main from go.podman.io/{common,image,storage} and then also make a commit to bump this version here 1.24.2 then it will make all the test PRs work.

flouthoc: ~~@Luap99 there is no `1.24.2` on https://hub.docker.com/_/golang , I guess I will just do `1.24`~~

Edit: Nevermind `1.24.2` is there.

flouthoc:

TomSweeneyRedHat:
LGTM
@nalind any objections?

flouthoc:
@nalind gentle reminder ! is there anything blocking this PR.

PR #6390, opened 2025-09-17
Include the Sequoia crypto backend, and run integration tests with it

mtrmac:
#### What type of PR is this?

> /kind feature

To support Buildah builds using the Sequoia crypto backend, it would be useful to have at least _some_ tests.

#### How to verify it

Manually inspect that the tests use a Sequoia build.

None

```release-note
None
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -32,12 +32,14 @@ env:
     IMAGE_PROJECT: "libpod-218412"
     FEDORA_NAME: "fedora-42"
     PRIOR_FEDORA_NAME: "fedora-41"
+    RAWHIDE_NAME: "rawhide"
     DEBIAN_NAME: "debian-13"
 
     # Image identifiers
     IMAGE_SUFFIX: "c20250910t092246z-f42f41d13"
     FEDORA_CACHE_IMAGE_NAME: "fedora-${IMAGE_SUFFIX}"
     PRIOR_FEDORA_CACHE_IMAGE_NAME: "prior-fedora-${IMAGE_SUFFIX}"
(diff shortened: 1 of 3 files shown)


mtrmac:
@containers/buildah-maintainers PTAL, I’m not familiar with the test setup in this repo.

Ultimately, this seems to be just a compile-test of the crypto backend: I have manually verified that the binary does build with the Sequoia build tag, and does include the Sequoia implementation — but ~all tests here run with `tests/policy.json`, which disables signature enforcement, so nothing fails even if the Rust backend is completely unavailable.

Unlike Podman and Skopeo, this does not add new UI: Buildah doesn’t have a `--sign-by-sigstore`, and I haven’t heard any requests to add that, so I didn’t (yet?) work on adding `--sign-by-sq-fingerprint` either.

lsm5 (approved):
LGTM.

We should also update `rpm/buildah.spec` to include the new buildtag, but that can be in another PR.

flouthoc (approved):
LGTM

TomSweeneyRedHat:
@mtrmac looks like you have some conflicts that need tending to.

mtrmac:
Rebased (with no code changes — just dropped the first dependency update commit), PTAL again.

nalind:
Just to be sure:  this test is meant to be must-pass-to-merge and not simply informative?
LGTM if that's the case, or once the new test is moved to the `non_blocking_integration_task` otherwise.

mtrmac:
Yes, I was thinking must-pass-to-merge — with the distribution dependency moving as new Fedoras are released.

I.e. after we move to Fedora.next, have the “Fedora” run done with Sequoia, and drop the “Rawhide” run again; and one release after that, have the “previous Fedora” run done with Sequoia as well. The extra test run would only persist as until Fedora 43 is released and used in CI.

nalind:
Okay, then.
/lgtm

PR #6395, opened 2025-09-25
RPM: build with sequoia on F43+

lsm5:
Follow up on #6390

#### What type of PR is this?

/kind other

Builds fedora 43+ rpms with sequoia

#### How to verify it

check in build logs if buildtag has been added. Any System tests using sequoia should pass.

None

Merge only after #6390 

```release-note
None
```

### DIFF
--- a/rpm/buildah.spec
+++ b/rpm/buildah.spec
@@ -11,6 +11,9 @@
 
 %if %{defined fedora}
 %define build_with_btrfs 1
+%if 0%{?fedora} >= 43
+%define sequoia 1
+%endif
 %endif
 
 %if %{defined rhel}
@@ -69,6 +72,9 @@ BuildRequires: libseccomp-devel
 %endif
 Requires: libseccomp >= 2.4.1-0
 Suggests: cpp
+%if %{defined sequoia}
+Requires: podman-sequoia
+%endif
 
 %description
 The %{name} package provides a command line tool which can be used to
@@ -130,6 +136,10 @@ export BUILDTAGS+=" exclude_graphdriver_btrfs"
 
 %if %{defined fips}
 export BUILDTAGS+=" libtrust_openssl"
+%endif
+
+%if %{defined sequoia}
+export BUILDTAGS+=" containers_image_sequoia"
 %endif
 
 %gobuild -o bin/%{name} ./cmd/%{name}


TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
LGTM
but it looks like you need a test, or do you want to slap the no new tests label on this?

lsm5:
label added. This should only go in after the sequoia support PR mentioned in description.

TomSweeneyRedHat:
@lsm5 is this ready to have Draft label yanked?

mtrmac (commented):
Thanks!

lsm5:
I see the sequoia PR is merged. Rebased on the latest main. Once tests pass, I'll get this out of draft.

lsm5:
@containers/buildah-maintainers PTAL, good to go.

rpm/buildah.spec
```diff
@@ -69,6 +72,9 @@ BuildRequires: libseccomp-devel
 %endif
 Requires: libseccomp >= 2.4.1-0
 Suggests: cpp
+%if %{defined sequoia}
+Requires: podman-sequoia
+%endif
```
nalind, line 77: Remind me again why we have a packaging dependency on podman, when podman has a library dependency on buildah?

mtrmac: `podman-sequoia` is https://src.fedoraproject.org/rpms/rust-podman-sequoia , a separate project+repo and not a subpackage of Podman. Yes, the name of that library is a bit confusing, given the non-Podman users.

nalind: Alright, then.

lsm5: I did check with the crypto team about renaming it, but they said the crate was already published and it was too late to change it.

lsm5:
did this not make it into v1.42?

nalind:
The commit log shows the merge commit for #6439 being applied after the merge commit for this one, but the patches in it are _before_ this one, and the first of those is where the tag was applied.

lsm5:
Ack. I'll manually update dist-git to build with sequoia. No worries. If we're branching a release-1.42 upstream, then this patch should be cherrypicked there so packit won't overwrite the dist-git sequoia changes.

PR #6421, opened 2025-10-08
`manifest add`: prefer lists when looking up for `--all`

nalind:
#### What type of PR is this?

/kind bug

When looking up the image to add, if we were passed `--all`, set the flag that we pass to libimage to indicate that we prefer a list if there is one, so that it doesn't just hand us back one image from a list.

#### How to verify it

Updated test!

Related to https://github.com/containers/podman/issues/27228.

Requires https://github.com/containers/container-libs/pull/382.

```release-note
None
```

### DIFF
--- a/cmd/buildah/manifest.go
+++ b/cmd/buildah/manifest.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	iutil "github.com/containers/buildah/internal/util"
 	"github.com/containers/buildah/pkg/cli"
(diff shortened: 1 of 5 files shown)


TomSweeneyRedHat:
All kinds of angry red tests @nalind

nalind:
The updated test won't pass without https://github.com/containers/container-libs/pull/382.

flouthoc (approved):
LGTM

JohnTheCoolingFan:
Why wasn't this merged?

PR #6274, opened 2025-07-10
build,add: add support for corporate proxies

userid0x0:
Use the Proxy Environment http_proxy/HTTPS_PROXY/NO_PROXY to handle the http/https request.

#### What type of PR is this?

> /kind bug

#### How to verify it

See instructions in initial error report. https://github.com/containers/podman/issues/25220

#### Which issue(s) this PR fixes:

`Fixes: https://github.com/containers/podman/issues/25220`

PR #6441, opened 2025-10-20
Move define.TempDirForURL() to pkg/tmpdir.ForURL(), unify its and Add()'s http client configurations

nalind:
#### What type of PR is this?

/kind bug

The HTTP client configurations used by Add() and the logic that we use for pulling down build contexts that are specified using URLs differ in their setup, and the latter doesn't implement the --tls-verify=false flag.

#### How to verify it

New integration test!

None

This is also a follow-up to #6274.

This moves the function that we use for retrieving things and putting them into temporary directories into its own package, but leaves behind an API-compatible wrapper function.

```release-note
Retrieval of build contexts which are specified using HTTPS URLs now respect the value of the `--tls-verify` flag.
```

### DIFF
--- a/internal/httpclient/httpclient.go
+++ b/internal/httpclient/httpclient.go
@@ -1,0 +1,46 @@
+package httpclient
+
+import (
+	"crypto/tls"
+	"net/http"
+	"net/url"
+
+	"github.com/docker/go-connections/tlsconfig"
+	"go.podman.io/image/v5/pkg/tlsclientconfig"
+	"go.podman.io/image/v5/types"
+)
+
+// URLOptions provides client-side options used by ForURLOptions().
+type URLOptions struct {
+	// Location of CA certificates, if not the system default
+	CertPath              string
+	InsecureSkipTLSVerify types.OptionalBool
+	// Callback which controls which, if any, proxy server to use when retrieving HTTP or
+	// HTTPS sources.  Used to construct an http.Client's Transport.
+	Proxy func(*http.Request) (*url.URL, error)
+}
+
+// ForURLOptions returns an http.Client with the settings from `options`
+// applied to its mostly-default transport, configured to use proxy settings
+// from the environment.
+func ForURLOptions(options URLOptions) (*http.Client, error) {
+	tlsClientConfig := &tls.Config{
+		// As of 2025-08, tlsconfig.ClientDefault() differs from Go 1.23 defaults only in CipherSuites;
+		// so, limit us to only using that value. If go-connections/tlsconfig changes its policy, we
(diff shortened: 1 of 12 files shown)


TomSweeneyRedHat:
@nalind, did you intend to use the name `tmpdir` for the new `/pkg` directory that you're putting the tls code into?

nalind:
Yeah, it's used by `tmpdir.ForURL()`, which is meant to replace `define.TempDirForURL()`.

nalind:
Moved those bits to a new `internal/httpclient` package.

nalind:
Tweaked a bit to expose the proxy setting directly for callers.

pkg/tmpdir/url.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusBadRequest {
+		return fmt.Errorf("invalid response status %d", resp.StatusCode)
+	}
+	if resp.ContentLength == 0 {
+		return fmt.Errorf("no contents in %q", url)
+	}
+	if err := chrootarchive.Untar(resp.Body, dir, nil); err != nil {
+		resp1, err := http.Get(url)
```
Honny1, line 183: Should be here `httpClient`?

internal/httpclient/httpclient.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/docker/go-connections/tlsconfig"
+	"go.podman.io/image/v5/pkg/tlsclientconfig"
+	"go.podman.io/image/v5/types"
+)
+
+// URLOptions provides client-side options used by ForURLOptions().
+type URLOptions struct {
+	CertPath              string // location of CA certificates, if not the system default
+	InsecureSkipTLSVerify types.OptionalBool
+	Proxy                 func(*http.Request) (*url.URL, error)
```
Honny1, line 17: I am not sure about the behavior of not setting the Proxy option. Maybe it should be checked or documented.

pkg/tmpdir/url.go
nalind: Yes, fixing.

internal/httpclient/httpclient.go
nalind: Adding godoc for it.

Honny1 (approved):
LGTM

Honny1 (approved):
LGTM

PTAL @TomSweeneyRedHat @ajoshua2004

ajoshua2004 (approved):
LGTM

PR #6444, opened 2025-10-21
.github: add downstream PR workflow for podman repo

flouthoc:
Automatically create PRs in podman when buildah PRs are opened to test compatibility and catch breaking changes early.

This workflow is similar to: https://github.com/containers/container-libs/blob/main/.github/workflows/opendownstream-pr.yml

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```

### DIFF
--- a/.github/workflows/opendownstream-pr.yml
+++ b/.github/workflows/opendownstream-pr.yml
@@ -1,0 +1,154 @@
+name: 'Open downstream PRs'
+
+on:
+  pull_request_target:
+    branches:
+      - 'main'
+    paths:
+      - '**/*.go'
+      - '!vendor/**'
+      - '!**/*_test.go'
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Self'
+        uses: actions/checkout@v5
+        # Because we use pull_request_target, this checks out the code from the default branch, NOT the PR branch
+
+      - name: 'Setup Go'
+        uses: actions/setup-go@v6
+        with:
+          go-version: 'stable'
+
+      - name: 'Checkout forked podman'
+        uses: actions/checkout@v5
+        with:
+          repository: 'podmanbot/podman' # The target repository
+          path: 'podman'                 # Checkout into a sub-directory
+          fetch-depth: '1'
+          token: ${{ secrets.VENDOR_TOKEN_PODMANBOT }} # We need to push into podmanbot/podman
+
+      - name: 'Vendor Code from this repo to podman'
+        run: |
+          # Get the current commit SHA from the PR
+          COMMIT_SHA="${{ github.event.pull_request.head.sha }}"
+          echo "Using commit SHA: $COMMIT_SHA"
+
+          cd podman
+          # Create a unique branch name based on the buildah PR number
+          BRANCH_NAME="sync/buildah-${{ github.event.pull_request.number }}"
+          git switch -c $BRANCH_NAME
+          git remote add upstream https://github.com/containers/podman.git
+          git fetch upstream
+          git rebase upstream/main
+
+          # Function to update module and verify
+          update_module() {
+            local module=$1
+            echo "Updating module: $module"
+            go mod edit -replace ${module}=github.com/${{ github.event.pull_request.head.repo.full_name }}@${COMMIT_SHA}
+            GOWORK=off go mod tidy
+          }
+
+          # Update buildah module
+          update_module "github.com/containers/buildah"
+          GOWORK=off go mod vendor
+          GOWORK=off go mod verify
+
+          echo "Updated go.mod:"
+          cat go.mod
+
+      - name: 'Commit and Push to podman'
+        run: |
+          cd podman
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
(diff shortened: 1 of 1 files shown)


nalind:
If the bot's not going to to do it, this needs text advising _someone_ that they're expected to close the corresponding podman PR when their buildah PR is merged or closed.

.github/workflows/opendownstream-pr.yml
```diff
@@ -0,0 +1,154 @@
+name: 'Open downstream PRs'
+
+on:
+  pull_request_target:
+    branches:
+      - 'main'
+    paths:
+      - '**/*.go'
+      - '!vendor/**'
+      - '!**/*_test.go'
```
TomSweeneyRedHat, line 10: Do we need to run this if only a Buildah test has changed?

.github/workflows/opendownstream-pr.yml
```diff
@@ -0,0 +1,154 @@
...
+
```
TomSweeneyRedHat, line 11: Is there a way that we could add a switch of some sort to a PR to avoid running this?

.github/workflows/opendownstream-pr.yml
nalind: Am I right in thinking that the buildah_bud_test_task in podman's CI will be running test scripts from PRs that are opened here?

.github/workflows/opendownstream-pr.yml
TomSweeneyRedHat: Ah, looks like you can set an env variable with SKIP_PR.  I don't know if that can be tied into a GH label or not, but I'd prefer a label to use if possible.

.github/workflows/opendownstream-pr.yml
TomSweeneyRedHat: I think so @nalind.
And I take my comment back.  I just noted the `!` in the front of line 10, so this ignores PRs with just a `*_test.go` change.  Which I think is the right thing to do.

TomSweeneyRedHat:
@mheon any concerns with running more CI tests?

mheon:
No, we definitely want this - reverse-dependency testing is something we've wanted for a long while

TomSweeneyRedHat:
@mheon I concur on the reverse-dependency testing, I'm just a little concerned about running out of CI resources without needing to pay more.  We'll cross that bridge if/when we get there.

TomSweeneyRedHat:
@flouthoc I thought earlier today you said you had a change or two to do?  I'm not seeing changes since the last comments, but the GitHub UI has been acting whacky this afternoon for me.  Do you have more to do stiill?

curtiscrummie39 (approved):

.github/workflows/opendownstream-pr.yml
flouthoc: > Do we need to run this if only a Buildah test has changed?

@TomSweeneyRedHat We ignore running the workflow if only a test file is changed. Idea is not to spam podman repo with PR's which don't have any actual behaviour change in buildah.

> Am I right in thinking that the buildah_bud_test_task in podman's CI will be running test scripts from PRs that are opened here?

@nalind Yes that is correct afaik.

flouthoc:
> If the bot's not going to to do it, this needs text advising _someone_ that they're expected to close the corresponding podman PR when their buildah PR is merged or closed.

Modified the text in variable `PR_BODY` mentioning the following responsibility is taken by @containers/buildah-maintainers

.github/workflows/opendownstream-pr.yml
flouthoc: Not as of now, but this will not run on any `vendor` and `test` related changes, whenever any change is done in business logic of buildah it will open a PR on podman repo, but maybe in future we can introduce a `label` to ignore for certain PRs ? WDYT ?

curtiscrummie39 (approved):

.github/workflows/opendownstream-pr.yml
```diff
@@ -0,0 +1,154 @@
...
+          SELF_REPO_PR_URL: ${{ github.event.pull_request.html_url }}
+          SELF_REPO_PR_TITLE: ${{ github.event.pull_request.title }}
+        run: |
+          cd podman
+
+          BRANCH_NAME="sync/buildah-${{ github.event.pull_request.number }}"
+          PR_TITLE="Sync: ${SELF_REPO_PR_TITLE}"
+          PR_BODY="This PR automatically vendors changes from [buildah#${SELF_REPO_PR_NUMBER}](${SELF_REPO_PR_URL}).
+
+          This PR is intended solely for testing purposes and should not be merged. Ideally, upstream maintainers ( @containers/buildah-maintainers )  should review any failing tests here and close this PR once the upstream PR is merged. Maintainers of this repository do not need to take any action, but may review the changes and notify upstream maintainers of any breaking issues if they wish.
```
nalind, line 108: These two sentences don't seem to agree.  Copy-paste with a missing update?

nalind: Never mind, that text goes into a podman PR.

nalind:
Given https://github.com/containers/container-libs/pull/604, I don't think we're going to proceed with this one.

PR #6453, opened 2025-10-26
Podman6: remove CNI

lsm5:
#### What type of PR is this?

> /kind cleanup

Removes CNI for podman6

#### How to verify it

check if any CNI stuff still exists

None

```release-note
CNI support has been removed.
```


define/build.go
```diff
@@ -236,14 +236,7 @@ type BuildOptions struct {
 	// namespace), effectively deciding whether or not the process has a
 	// usable network.
 	ConfigureNetwork NetworkConfigurationPolicy
-	// CNIPluginPath is the location of CNI plugin helpers, if they should be
-	// run from a location other than the default location.
-	CNIPluginPath string
-	// CNIConfigDir is the location of CNI configuration files, if the files in
-	// the default configuration directory shouldn't be used.
-	CNIConfigDir string
```
nalind, line 244: This is an API break.  I'd prefer if these fields were marked as deprecated, like `BuildOutput`.

pkg/cli/common.go
```diff
@@ -42,13 +42,11 @@ type UserNSResults struct {
...
-	Network       string
-	CNIConfigDir  string
-	CNIPlugInPath string
-	PID           string
-	UTS           string
+	Cgroup  string
+	IPC     string
+	Network string
+	PID     string
+	UTS     string
```
nalind: This is an API break.

buildah.go
```diff
@@ -147,14 +147,7 @@ type Builder struct {
 	// namespace), effectively deciding whether or not the process has a
 	// usable network.
 	ConfigureNetwork define.NetworkConfigurationPolicy
-	// CNIPluginPath is the location of CNI plugin helpers, if they should be
-	// run from a location other than the default location.
-	CNIPluginPath string
-	// CNIConfigDir is the location of CNI configuration files, if the files in
-	// the default configuration directory shouldn't be used.
-	CNIConfigDir string
```
nalind, line 155: This is an API break.  I'd prefer if the godoc for these fields noted that they are deprecated and are expected to be empty, possibly with a tag indicating that they won't be saved when encoded as JSON.

buildah.go
```diff
@@ -224,8 +217,6 @@ type BuilderInfo struct {
 	NamespaceOptions      define.NamespaceOptions
 	Capabilities          []string
 	ConfigureNetwork      string
-	CNIPluginPath         string
-	CNIConfigDir          string
```
nalind, line 228: This is an API break.  I'd prefer if the godoc for these fields noted that they are deprecated and are expected to be empty.

buildah.go
```diff
@@ -333,14 +322,7 @@ type BuilderOptions struct {
 	// namespace), effectively deciding whether or not the process has a
 	// usable network.
 	ConfigureNetwork define.NetworkConfigurationPolicy
-	// CNIPluginPath is the location of CNI plugin helpers, if they should be
-	// run from a location other than the default location.
-	CNIPluginPath string
-	// CNIConfigDir is the location of CNI configuration files, if the files in
-	// the default configuration directory shouldn't be used.
-	CNIConfigDir string
```
nalind, line 341: This is an API break.  I'd prefer if these fields were marked as deprecated, like the `define` package's `BuildOptions.BuildOutput`.

run.go
```diff
@@ -129,12 +129,6 @@ type RunOptions struct {
 	// namespace), effectively deciding whether or not the process has a
 	// usable network.
 	ConfigureNetwork define.NetworkConfigurationPolicy
-	// CNIPluginPath is the location of CNI plugin helpers, if they should be
-	// run from a location other than the default location.
-	CNIPluginPath string
-	// CNIConfigDir is the location of CNI configuration files, if the files in
-	// the default configuration directory shouldn't be used.
-	CNIConfigDir string
```
nalind, line 137: This is an API break. I'd prefer if these fields were marked as deprecated, like the `define` package's `BuildOptions.BuildOutput`.

nalind (changes requested):
Please avoid breaking the public API when possible.

lsm5:
@nalind good for another look. The testing-farm job failures are a Fedora infra issue, already notified them.

lsm5:
This will stay in draft until https://github.com/containers/container-libs/pull/412 is merged.

imagebuildah/executor.go
```diff
@@ -101,9 +101,7 @@ type executor struct {
 	isolation                      define.Isolation
 	namespaceOptions               []define.NamespaceOption
 	configureNetwork               define.NetworkConfigurationPolicy
-	cniPluginPath                  string
-	cniConfigDir                   string
-	// NetworkInterface is the libnetwork network interface used to setup CNI or netavark networks.
+	// networkInterface is the libnetwork network interface used to setup netavark networks.
```
nalind, line 104: 👍

internal/mkcw/embed/entrypoint_amd64.gz
Luap99, line 1: this should be dropped from the commits

Luap99 (commented):
LGTM overall minus the one extra file

internal/mkcw/embed/entrypoint_amd64.gz
lsm5: thanks. I've added this to .gitignore in HEAD. Hope that's not a problem @nalind @TomSweeneyRedHat

lsm5: hmm, lack of this file seems to be a problem while vendoring buildah into podman.

```
$ make vendor
go mod tidy
go mod vendor
go: resolving embeds in github.com/containers/buildah/internal/mkcw: pattern embed/entrypoint_amd64.gz: no matching files found
make: *** [Makefile:358: vendor] Error 1
```

lsm5: and I see it included in the tree on buildah main branch as well.

Luap99: the file itself is included on purpose, I don't want to get into the details here but we ship binaries as part of the source tree so that file can get embedded into a confidential container image.

What we should not do is recommit the build artifact over and over when the source was unchanged so just drop that from your commits. Yes annoying I know and I would love that we do not ship binary artifacts in the source tree but that is a conversation for another day.

lsm5: done. thanks

Luap99 (approved):
LGTM 

One more thing I found in the CI test scrip we have an rm -rf on the cni dir which can be dropped as well
https://github.com/containers/buildah/blob/09917c48eb3be47ea2ef70e507163cad68d26dda/contrib/cirrus/test.sh#L11-L14

Anyhow non blocking, tests look good here and this can be fixed easily later so I think it is better to merge now to continue the dance into podman

lsm5:
@Luap99 Thanks. Will fix the test script in a followup.

@containers/buildah-maintainers draft status removed. PTAL.

mtrmac (approved):

mtrmac:
@nalind PTAL, updating your review status is required for merging.

Luap99:
> @nalind PTAL, updating your review status is required for merging.

FYI in the github UI allows us to dissmis it, you need to click on "1 requested chnage" to drop down then you see nalind and on the right side there is a pop up menu when you click on the tree dots.
So in case someone is away/unresponsive we can do that.
Of course in the case here sure lets wait for the final review from @nalind

mtrmac:
> FYI in the github UI allows us to dissmis it

I didn’t know that, thanks!

nalind (approved):
LGTM once the merge conflict is resolved.

PR #6469, opened 2025-11-03
RPM: build the entrypoint binary before the main buildah binary

lsm5:
#### What type of PR is this?

 /kind other

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:
None

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #6471, opened 2025-11-04
internal/mkcw/embed: cross-compile using Go

nalind:
#### What type of PR is this?

/kind other

Use the Go toolchain to cross-compile the "This image is designed to be run as a confidential workload using libkrun." entrypoint that we add to confidential workload images.  It's bigger than it was before, but easier to port and can be built from source every time.

#### How to verify it

Check that we can still build on non-amd64 architectures.

Sets up #6469, if we need to stop checking in a pre-built binary.

```release-note
None
```


Makefile
```diff
@@ -66,16 +66,34 @@ bin/buildah: $(SOURCES) internal/mkcw/embed/entrypoint_amd64.gz
...
 
-ifneq ($(shell $(AS) --version | grep x86_64),)
 internal/mkcw/embed/entrypoint_amd64.gz: internal/mkcw/embed/entrypoint_amd64
 	gzip -k9nf $^
+internal/mkcw/embed/entrypoint_arm64.gz: internal/mkcw/embed/entrypoint_arm64
+	gzip -k9nf $^
+internal/mkcw/embed/entrypoint_ppc64le.gz: internal/mkcw/embed/entrypoint_ppc64le
+	gzip -k9nf $^
+internal/mkcw/embed/entrypoint_s390x.gz: internal/mkcw/embed/entrypoint_s390x
+	gzip -k9nf $^
```
Luap99, line 76: What do we need the other arches for if we only ever embed the amd64 one? Seems like unnecessary churn having to maintain a assembly version for each arch that we don't ever end up using.

nalind: We don't build them, but the confidential workload folks have been making noises about supporting other architectures for a while now.

nalind:
Had to add the prebuilt binary back to avoid breaking people who vendor us as a library.

Luap99:
> Had to add the prebuilt binary back to avoid breaking people who vendor us as a library.

How is the code reachable? Only via buildah mkcw or also directly via the normal podman/buildah build commands?
Because right now the code/package gets imported in podman and as such podman embeds the same binary so I think we would need to apply the same fix in podman to rebuild the binary in the vendor dir as part of the specfile I think?

IF it is not needed for build it might make sense to try to split the packages so it does not get leaked into podman then?

nalind:
> > Had to add the prebuilt binary back to avoid breaking people who vendor us as a library.
> 
> How is the code reachable? Only via buildah mkcw or also directly via the normal podman/buildah build commands? Because right now the code/package gets imported in podman and as such podman embeds the same binary so I think we would need to apply the same fix in podman to rebuild the binary in the vendor dir as part of the specfile I think?

`build` has a `--cw` flag which changes how we commit the final image, along similar lines.

Luap99:
> `build` has a `--cw` flag which changes how we commit the final image, along similar lines.

Ack, so then I guess we just need to figure out a way how to rebuild this binary as part of the podman rpm then. Since buildah has top level code the Makefile is actually part of vendor/. So I suppose the simple way would be to cd into vendor and just rerun the target like we do here.

cc @lsm5

lsm5 (approved):
spec change looks good.

LGTM. Thanks a lot @nalind

lsm5:
> Ack, so then I guess we just need to figure out a way how to rebuild this binary as part of the podman rpm then. Since buildah has top level code the Makefile is actually part of vendor/. So I suppose the simple way would be to cd into vendor and just rerun the target like we do here.

ack. Once this is vendored we can update the podman spec.

nalind:
Native `strip` was removing the architecture from the ELF headers, switched to linking with -s.

internal/mkcw/embed/check.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
Luap99: nit: scripts tend to use `#!/usr/bin/env bash` here

https://github.com/containers/buildah/commit/b288b7a3d96a0c8ee2e0d580402a1a335a29f35b

internal/mkcw/embed/check.sh
```diff
@@ -0,0 +1,12 @@
...
+msg="This image is designed to be run as a confidential workload using libkrun."
+cd $(dirname ${BASH_SOURCE[0]})
+for GOARCH in amd64 arm64 ppc64le s390x ; do
+	make -C ../../.. internal/mkcw/embed/entrypoint_$GOARCH
+	case $GOARCH in
+		amd64) QEMUARCH=x86_64;;
+		arm64) QEMUARCH=aarch64;;
+		ppc64le|s390x) QEMUARCH=$GOARCH;;
+	esac
+	test "$(qemu-$QEMUARCH ./entrypoint_$GOARCH)" = "$msg"
```
Luap99: Does this need to add a  `|| echo "./entrypoint_$GOARCH is not working"` show easy which arch failed?
Right now the script is not `set -e` either so this could fail silently and the script exit code could still be 0 if the latest  s390x arch works.

internal/mkcw/embed/check.sh
nalind: We seem to be inconsistent about it, but sure, changing it.

internal/mkcw/embed/check.sh
nalind: Right, the tool outputs the error message and exits with an error code, so it's a little trickier.  Changed it.

Luap99 (approved):
/lgtm

PR #6473, opened 2025-11-05
[CVE-2025-52881] vendor: update to github.com/opencontainers/runc@v1.3.3

cyphar:
This is necessary to pull in some security hardening added to the
AppArmor configuration, used by the "chroot" isolation mode of buildah.https://issues.redhat.com/browse/RHEL-126637

Note that this depends on github.com/cyphar/filepath-securejoin@v0.5.1,
[which is not yet approved by the CNCF for use in podman or buildah](https://github.com/cncf/foundation/issues/1154).

/kind bug
Fixes: [CVE-2025-52881](https://github.com/opencontainers/runc/security/advisories/GHSA-cgrx-mc8f-2prm)
Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

Fixes: https://issues.redhat.com/browse/RHEL-126637, https://issues.redhat.com/browse/RHEL-126926


Luap99:
Thanks @cyphar, that does seem to put us in a very bad spot as I don't think we can merge this due the license rules unfortunately. I guess we need to find a way to escalate this on the CNCF side.

cc @mheon @nalind @mtrmac @baude @giuseppe @TomSweeneyRedHat @actionmancan

giuseppe:
should we fix the security vulnerability first and ask for forgiveness later?

cyphar:
Sorry, if I'd known that buildah was importing runc directly earlier I would've tried to come up with a nicer solution. :/

I'll leave a comment in the CNCF license request repo mentioning that this update is involved in a security update for buildah.

cyphar:
For what it's worth, if you don't need to support pre-5.6 kernels you can just use `openat2(2)` directly with `RESOLVE_NO_XDEV` and do the validation check for `/proc` being the root (inode number is `1`).

Luap99:
> I'll leave a comment in the CNCF license request repo mentioning that this update is involved in a security update for buildah.

I asked on the CNCF slack maintainer channel if we can "escalate" it, let's see what they say.

> For what it's worth, if you don't need to support pre-5.6 kernels you can just use openat2(2) directly with RESOLVE_NO_XDEV and do the validation check for /proc being the root (inode number is 1).

I don't think we a specific kernel support policy, given we may have to backport these fixes requiring a new kernel version may not be best.

actionmancan:
From CNCF we are to proceed with the security fix and then ask for license exception.

mheon:
Alright, let's get CI green and get this merged

mtrmac:
/lgtm

for the bots, relying on the reviews above

PR #6476, opened 2025-11-05
feat(build): print error on build flag --output=type=something

iTrooz:
#### What type of PR is this?

/kind feature

Two things:
- make build flag `--output=tar,dest=-` have correct behaviour (streaming to stdout)
- print error on build flag `--output=type=something` instead of outputting image to folder `type=something` (this is the case whether "something" is a valid type or not)

#### How to verify it
- Run `podman build . --output=tar,dest=-`
- Run `podman build --output=type=something`

None

```release-note
make build flag `--output=tar,dest=-` have correct behaviour (streaming to stdout)
print error on build flag `--output=type=something` instead of outputting image to folder `type=something`
```

### DIFF
--- a/internal/parse/build_output.go
+++ b/internal/parse/build_output.go
@@ -8,9 +8,10 @@ import (
 type BuildOutputType int
 
 const (
-	BuildOutputStdout   BuildOutputType = 0 // stream tar to stdout
-	BuildOutputLocalDir BuildOutputType = 1
-	BuildOutputTar      BuildOutputType = 2
+	BuildOutputInvalid  BuildOutputType = 0
+	BuildOutputStdout   BuildOutputType = 1 // stream tar to stdout
+	BuildOutputLocalDir BuildOutputType = 2
+	BuildOutputTar      BuildOutputType = 3
 )
 
 // BuildOutputOptions contains the the outcome of parsing the value of a build --output flag
@@ -40,7 +41,7 @@ func GetBuildOutput(buildOutput string) (BuildOutputOption, error) {
 	}
 
 	// Support complex values, in the form --output type=local,dest=./mydir
-	var typeSelected BuildOutputType = -1
+	typeSelected := BuildOutputInvalid
 	pathSelected := ""
 	for option := range strings.SplitSeq(buildOutput, ",") {
 		key, value, found := strings.Cut(option, "=")
@@ -49,7 +50,7 @@ func GetBuildOutput(buildOutput string) (BuildOutputOption, error) {
 		}
 		switch key {
 		case "type":
-			if typeSelected != -1 {
+			if typeSelected != BuildOutputInvalid {
 				return BuildOutputOption{}, fmt.Errorf("duplicate %q not supported", key)
 			}
 			switch value {
@@ -71,7 +72,7 @@ func GetBuildOutput(buildOutput string) (BuildOutputOption, error) {
 	}
 
 	// Validate there is a type
-	if typeSelected == -1 {
+	if typeSelected == BuildOutputInvalid {
 		return BuildOutputOption{}, fmt.Errorf("missing required key %q in build output option: %q", "type", buildOutput)
 	}
 
--- a/internal/parse/build_output.go
+++ b/internal/parse/build_output.go
@@ -1,0 +1,102 @@
+package parse
+
+import (
+	"fmt"
+	"strings"
+)
+
+type BuildOutputType int
+
+const (
+	BuildOutputStdout   BuildOutputType = 0 // stream tar to stdout
+	BuildOutputLocalDir BuildOutputType = 1
+	BuildOutputTar      BuildOutputType = 2
+)
+
+// BuildOutputOptions contains the the outcome of parsing the value of a build --output flag
+type BuildOutputOption struct {
+	Type BuildOutputType
+	Path string // Only valid if Type is local dir or tar
+}
+
+// GetBuildOutput is responsible for parsing custom build output argument i.e `build --output` flag.
+// Takes `buildOutput` as string and returns BuildOutputOption
+func GetBuildOutput(buildOutput string) (BuildOutputOption, error) {
+	// Support simple values, in the form --output ./mydir
+	if !strings.Contains(buildOutput, ",") && !strings.Contains(buildOutput, "=") {
+		if buildOutput == "-" {
+			// Feature parity with buildkit, output tar to stdout
+			// Read more here: https://docs.docker.com/engine/reference/commandline/build/#custom-build-outputs
+			return BuildOutputOption{
+				Type: BuildOutputStdout,
+				Path: "",
+			}, nil
+		}
+
+		return BuildOutputOption{
+			Type: BuildOutputLocalDir,
+			Path: buildOutput,
+		}, nil
+	}
+
+	// Support complex values, in the form --output type=local,dest=./mydir
+	var typeSelected BuildOutputType = -1
+	pathSelected := ""
+	for option := range strings.SplitSeq(buildOutput, ",") {
+		key, value, found := strings.Cut(option, "=")
+		if !found {
+			return BuildOutputOption{}, fmt.Errorf("invalid build output options %q, expected format key=value", buildOutput)
+		}
+		switch key {
+		case "type":
+			if typeSelected != -1 {
+				return BuildOutputOption{}, fmt.Errorf("duplicate %q not supported", key)
+			}
+			switch value {
+			case "local":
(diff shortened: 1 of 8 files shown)


iTrooz:
I updated the title with the most important change. Would you like me to split the PR into 2 instead ?

iTrooz:
CI failed because of an image pull timeout. `TestConformance/header-builtin`  succeeds locally. I think it should be re-run

iTrooz:
I'm having trouble figuring what's wrong in my PR. The errors seem unrelated to my PR, happening in files that I did not modify at all. Could I get some help please @nalind ? Thank you in advance

nalind:
The errors in the runc configurations are a known flake (#6205) that should be fixed in runc 1.3.2 or later, but our VM images don't include that yet, as prior to the recent round of security updates they probably had 1.3.0.  The error on the testing farm looks like a transient network problem.
I've restarted them and will keep an eye on them.

nalind (changes requested):
Hi!  Thanks for taking this on.

The changes to the `BuildOutputOption` type could be an API break.  I do prefer the use of an enumerated type there, though, and while I don't see the current parsing function or type being used in either podman or openshift/builder, I still worry about breaking other consumers that are using the current parsing function and interpreting its results.

The type isn't used as member of any other exported APIs or structures, so, while it would definitely be a bigger patch, would you consider marking both the current structure and the current `pkg/parse.GetBuildOutput()` implementation as deprecated (optionally keeping the added error checks), defining a new `BuildOutputOption` structure (that looks like the updated one here) and parsing function in `internal/parse`, and having `imagebuildah.stageExecutor.execute()` and `pkg/cli.GenBuildOptions()` call the new function instead?  I think that would let us keep your improvements without having to worry about breaking compilation for anyone who's calling the current API.

iTrooz:
All done !
I ran tests I modified/added locally

iTrooz:
Ok, CI is fully passing now !

nalind:
LGTM, thanks!

iTrooz:
Just fixed the commit length error :)
Thanks for accepting

internal/util/util.go
```diff
@@ -51,14 +51,14 @@ func NormalizePlatform(platform v1.Platform) v1.Platform {
...
-func ExportFromReader(input io.Reader, opts define.BuildOutputOption) error {
+func ExportFromReader(input io.Reader, opts parse.BuildOutputOption) error {
 	var err error
 	if !filepath.IsAbs(opts.Path) {
 		if opts.Path, err = filepath.Abs(opts.Path); err != nil {
 			return err
 		}
 	}
-	if opts.IsDir {
+	if opts.Type == parse.BuildOutputLocalDir {
```
mtrmac: If this is now an enum, please make this a `switch` and explicitly fail on unknown values. The two `!= …LocalDir` values can still share code the way they do now.

internal/util/util.go
```diff
@@ -51,14 +51,14 @@ func NormalizePlatform(platform v1.Platform) v1.Platform {
 }
 
 // ExportFromReader reads bytes from given reader and exports to external tar, directory or stdout.
-func ExportFromReader(input io.Reader, opts define.BuildOutputOption) error {
+func ExportFromReader(input io.Reader, opts parse.BuildOutputOption) error {
 	var err error
 	if !filepath.IsAbs(opts.Path) {
```
mtrmac: (Preexisting, but) this is accessing `opts.Path` even in situations where the value is documented to be unused/invalid — and typically calling `Abs("")`. That _happens_ to work correctly but it’s really an unnecessary thing to worry about; please gate this behind a condition, or normalize the path only inside the (future) `switch`.

(I don’t think that the `Abs` is valuable in for `os.Create` anyway, maybe it is necessary for `chrootarchive`. @nalind please correct me if I’m missing something.)

internal/parse/build_output.go
```diff
@@ -0,0 +1,102 @@
...
+type BuildOutputOption struct {
+	Type BuildOutputType
+	Path string // Only valid if Type is local dir or tar
+}
+
+// GetBuildOutput is responsible for parsing custom build output argument i.e `build --output` flag.
+// Takes `buildOutput` as string and returns BuildOutputOption
+func GetBuildOutput(buildOutput string) (BuildOutputOption, error) {
+	// Support simple values, in the form --output ./mydir
+	if !strings.Contains(buildOutput, ",") && !strings.Contains(buildOutput, "=") {
```
mtrmac, line 27: Hum, the `=` condition is new. @nalind do we want that behavior change?

pkg/parse/parse.go
```diff
@@ -734,6 +734,7 @@ func AuthConfig(creds string) (*types.DockerAuthConfig, error) {
 
 // GetBuildOutput is responsible for parsing custom build output argument i.e `build --output` flag.
 // Takes `buildOutput` as string and returns BuildOutputOption
+// Deprecated: This function is now internal
```
mtrmac, line 739: For the deprecation comment to be recognized by tools, it needs to be:
```suggestion
//
// Deprecated: This function is now internal
```
Also, if this is known to be producing incorrect/unexpected results (but we don’t want to change that), please document that in the deprecation notice as well.

internal/parse/build_output.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+
+		return BuildOutputOption{
+			Type: BuildOutputLocalDir,
+			Path: buildOutput,
+		}, nil
+	}
+
+	// Support complex values, in the form --output type=local,dest=./mydir
+	var typeSelected BuildOutputType = -1
```
mtrmac: This does work, but it rather breaks the semantic cleanliness of an enum.

Consider, instead, renumbering the existing enum values to start with 1, and define `BuildOutputInvalid` (…`Uninitialized`?) to be the 0 value. (That would also, as an extra benefit, protect against uninitialized `BuildOutputOption` values.)

mtrmac (commented):
I do like the enum.

---

(Conceptually, I’d prefer the parsing to happen at the highest CLI layer (e.g. so that we don’t start doing _anything_ about the build on invalid inputs); that would mean some kind of an API to create a ~public `BuildOutputOption` struct (although perhaps with private fields), and passing the parsed struct instead of just a string in a future equivalent of `BuildOptions.BuildOutputs`.

But that’s not the purpose of _this_ PR and way out of scope (e.g. it would require also a change to Podman). And anyway this PR as is, deprecating the current `BuildOutputOption` with its public fields, is a small move in that direction, already.)

internal/util/util.go
iTrooz: I made the fix anyway, but both type=local and type=tar work if I remove that code (and thus use a relative path)

internal/parse/build_output.go
iTrooz: (My rationale was to use the "complex parsing" for values like `type=local` (and fail)
For example, `docker build . -f my.Dockerfile --output=type=local` fails)

internal/util/util.go
mtrmac, line 104: The part I wanted most of all from the move to a `switch` was a `default: /* fail */`

internal/util/util.go
```diff
@@ -78,14 +85,17 @@ func ExportFromReader(input io.Reader, opts define.BuildOutputOption) error {
...
-	} else {
-		outFile := os.Stdout
-		if !opts.IsStdout {
-			if outFile, err = os.Create(opts.Path); err != nil {
-				return fmt.Errorf("failed while creating destination tar at %q: %w", opts.Path, err)
-			}
-			defer outFile.Close()
+	case parse.BuildOutputStdout:
+		if _, err = io.Copy(os.Stdout, input); err != nil {
+			return fmt.Errorf("failed while performing copy to %q: %w", opts.Path, err)
```
mtrmac: Now that this case is entirely separate: `opts.Path` is not set here.

internal/parse/build_output.go
mtrmac: That’s a fair point. I’ll leave that design decision to Buildah maintainers.

TomSweeneyRedHat: Looks OK to me, but I'd like @nalind to weigh in.

TomSweeneyRedHat:
@iTrooz this needs a rebase, would you mind updating please?
LGTM
But I'd like a final head nod from @nalind before merging.

iTrooz:
Done !

internal/parse/build_output.go
nalind: Makes sense as a way to avoid creating a directory named "type=local".

nalind:
Tests pass after restarting one.  LGTM, merging.

PR #6502, opened 2025-11-13
Update VMs, linter, fix warnings, add a "fmt" target

nalind:
#### What type of PR is this?

/kind cleanup

* Update the version of golangci-lint that we use, since 2.1.0 can panic with newer compilers.
* Address new warnings.
* Add a "fmt" target to the top-level Makefile that runs its formatters.
* Drop a Dockerfile that I don't think we've used for years.
* Update the VM images to use the same ones that podman's main branch currently does.

#### How to verify it

CI should notice if we were using this Dockerfile somewhere that I didn't find.

None

Sometimes, suppressing a "revive" warning triggers a bogus warning from "nolintlint", so yeah, we're working around that by suppressing both of them.

```release-note
None
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -224,6 +224,18 @@ integration_task:
             DISTRO_NV: "${DEBIAN_NAME}"
             IMAGE_NAME: "${DEBIAN_CACHE_IMAGE_NAME}"
             STORAGE_DRIVER: 'vfs'
+        - env:
+            DISTRO_NV: "${FEDORA_NAME}"
+            IMAGE_NAME: "${FEDORA_CACHE_IMAGE_NAME}"
+            STORAGE_DRIVER: 'vfs'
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -36,7 +36,7 @@ env:
     DEBIAN_NAME: "debian-13"
 
     # Image identifiers
-    IMAGE_SUFFIX: "c20250910t092246z-f42f41d13"
+    IMAGE_SUFFIX: "c20251110t154831z-f42f41d14"
```
Luap99, line 39: blocker: This bumps to debian 14 so you must update DEBIAN_NAME above as well

nalind: Thanks for catching that!  Rebasing with that updated.

Luap99 (approved):
/lgtm

PR #6527, opened 2025-11-21
vendor: update container-libs, and runtime-spec

lsm5:
This also vendors tags.cncf.io/container-device-interface[/specs-go].

Similar Podman PR: https://github.com/containers/podman/pull/27579/

#### What type of PR is this?

/kind other

None

```release-note
None
```


lsm5:
@containers/buildah-maintainers I think it would be nice to keep main branch go.mod not tied to container-libs versions as that effectively disables future renovate jobs. Branching a new release- branch from main and then cutting a new release from that branch as in podman should make that work.

lsm5:
@containers/buildah-maintainers PTAL.

PR #6548, opened 2025-11-28
test: do not untar archive into fs when checking file names

iTrooz:
#### What type of PR is this?

> /kind cleanup

In BATS tests, it avoids untar'ing archives into the filesystem when it is not needed. Instead, it uses `tar -t` to list the archive files instead.

As a bonus, it also fixes the test `build with custom build output and output rootfs to tar`, which does not check the output of the right command (it seems like it wanted to read the output of `ls` but it read the output of `run_buildah`

#### How to verify it

Run test suite

None

```release-note
None
```


nalind:
Nice catch!  The commit log subject line is longer than we generally expect, though, causing it to be wrapped in the commits tab of the pull request web page.  LGTM with a tweak to the commit log.

iTrooz:
Done !

PR #6556, opened 2025-12-01
feat: add support for preserving and labeling intermediate stage images

### LINKED ISSUE #6257: Proposal for enhancing Buildah to support layer (rootfs.diff_ids) squashing with intermediate image retention in multistage builds
Hello,

 I’m currently designing functionality for our internal build and release tool **Konflux**, which relies heavily on **Buildah** for container image building.

As part of this work, I’ve encountered a potential limitation in Buildah that I’d like to discuss. I'm also prepared to contribute an implementation if we can agree on the best approach.

## Problem

In Konflux, we need to access content from intermediate images (e.g., packages, binaries) after the build process of multistage-build images. The only way I’m currently aware of to preserve these intermediate images is by using the `--layers` flag within `buildah build`. This works well for retaining intermediate layers, which we can then inspect post-build.

However, enabling `--layers` changes the structure of the final image’s metadata. Specifically:

- When using `--layers`, the `rootfs.diff_ids` of the final image includes **all intermediate layers** (undesired).
- When **not** using `--layers`, the final image contains only **two layers** in `rootfs.diff_ids` (desired):
  - One for the base image.
  - One for the squashed final image layers (I assume).

  But of course, **no intermediate images** are stored (undesired).
- In both cases, with or without `--layers`, the final image preserves the **full history** (also desired) containing parent base  image and final image history.
- Using `--squash` alongside `--layers` does not achieve the desired result either. This ends up **fully squashing everything** (including the parent base image), which removes necessary layering details (parent base image layer + all other squashed). In addition to that, **history is capturing only the last layer of the final image**, but we would like to preserve full history as it is produced without squash.

## What I cannot do

- Unfortunately, performing two separate builds (one with `--layers` for analysis and one without for the final image) is **not viable** for performance and pipeline simplicity reasons.
- I cannot use external tools like **docker-squash**, as they are heavily dependent on Docker and we want to certainly avoid that.

## Desired Outcome

I’m looking for a way to:

- Preserve intermediate images for inspection.
- Have only **two layers** in the final image’s `rootfs.diff_ids` (parent base image layer + squashed final image layers).
- Retain history metadata for the final image (parent base image and final image history)

## Example Containerfile && `buildah build` examples and outcomes

```
FROM docker.io/library/golang:1.21-bullseye AS builder

WORKDIR /app
RUN echo 'package main; import "fmt"; func main() { fmt.Println("Hello from Go!") }' > main.go && \
  go build -o hello main.go

FROM registry.access.redhat.com/ubi9/ubi:latest

RUN yum install python3-pip -y
RUN pip3 install semver==2.13.0

COPY --from=transitive-alias-builder /app/hello /usr/local/bin/hello

CMD ["echo"]
```

| Command                                               | Images Created <br> (buildah images) &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;  &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;                                                                              | `rootfs.diff_ids` (simplified)                                                                                                                                                             | History Retained | Drawbacks                                                                                 |
|--------------------------------------|----------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------|------------------------------------------------------------------|
| `buildah build` <br> `--layers` <br> `-t final-image:0.1`           | - `final-image:0.1`  <br> - **intermediate**: `<none>` <br> - base: `ubi9`  <br> - builder: `golang:1.21`      | - `sha256:2cef...` (base image layer)  <br> - `sha256:f600...` (yum install) <br> - `sha256:153a...` (pip install) <br> - `sha256:43cd...` (COPY hello)                                     | ✅ Yes            | ❗ Too many `diff_ids` <br> in final image – every <br>  instruction becomes a layer                |
| `buildah build`  <br>  `-t final-image:0.1`                    | - `final-image:0.1`  <br> - base: `ubi9` <br> - builder:&nbsp;`golang:1.21`                                             | - `sha256:2cef...` (base image layer) <br> - `sha256:92b3...` (squashed <br> final image layers)                                                                                         | ✅ Yes            | ❗ Intermediate image is not <br> retained – content from stages is lost                     |
| `buildah build` <br> `--squash --layers` <br> `-t final-image:0.1`  | - `final-image:0.1` <br> - **intermediate**: `<none>` <br> - base: `ubi9`  <br> - builder:&nbsp;`golang:1.21`          | - `sha256:92b3...` (fully <br> squashed: parent base image + all final image layers)                                                                                                              | ❌ No <br> (partial)   | ❗ All layers including <br> parent base image are squashed <br> ❗ History is mostly lost                  |
| `buildah build` <br> `--new-arg(s)` <br> `-t final-image:0.1`       | - `final-image:0.1`  <br> - **intermediate**: `<none>` <br> - base: `ubi9`  <br> - builder:&nbsp;`golang:1.21`         | - `sha256:2cef...` (base image layer) <br> - `sha256:92b3...` (squashed final image layers)                                                                                       | ✅ Yes            | ✅ Desired behavior – parent base image + final image layers, <br> preserved history, accessible intermediate stages |

## Potential Proposed Solution

Based on current functionality, I believe we could introduce a new **alternative flag: e.g., `--layers-stages-only`** which would store intermediate stages as images, but squashes all filesystem changes of the final image layer chain, effectively yielding the same layer structure as a `buildah build` without `--layers` (parent base image sha + sha containing all other changes from final image manifested in  `rootfs.diff_ids`).

Of course, I don't insist on the name of the argument, this is just an example, up for discussion.

## Questions

- Did I miss anything? Is there currently any supported combination of flags in Buildah that would achieve this behavior?
- Would such functionality (separating intermediate stage storage from final image layering) be considered a reasonable feature to add? If so, I’d be happy to submit a patch or collaborate on an implementation aligned with Buildah's design philosophy.

---

Thank you for your time and consideration. I’d appreciate any guidance or thoughts on how best to proceed.

ezopezo:
This adds support for preserving and labeling intermediate stage images (including final image) in multi-stage builds. `--save-stages` preserves only the final image from each intermediate stage, not every instruction layer (as `--layers` flag). This also keeps the final image's layer count same as in regular builds (with no additional args). `--stage-labels` adds label metadata containing base image and alias from Containerfile at the beginning of the stage.

New flags:
  - `--save-stages`: save intermediate stage final images instead of removing them.
        Build without added `--layers`, intermediate images are expected to accumulate on each build (no cache reuse).
        Build with added `--layers`, enables layer cache reuse for subsequent builds.

  - `--stage-labels`: adds metadata labels to intermediate stage images and final image:
    * `io.buildah.stage.name`: the stage name (alias or stage index if alias not specified)
    * `io.buildah.stage.base`: the base image (external image reference - pullpsec or
          parent stage image ID if its base is another stage)
    Requires `--save-stages`.
    Cache reuse depends of if first or second build has or has not `–stage-labels` argument or if alias/pullspec changed between builds in Containerfile

The implementation includes:
  - Validation that `--stage-labels` requires `--save-stages`
  - Detection when a stage uses another intermediate stage as base (and labeling the stage accordingly)
  - Stage labels are added to Containerfile via parse tree injection as LABEL at the beginning of the stage including final
  - Various test coverage for saving, labeling, and caching scenarios
#### What type of PR is this?

/kind feature

**General use**: This functionality is useful for identification and debugging intermediate stage images in multi-stage builds.

**Specific need**: Identifying the content copied from intermediate stages in multi-stage builds into the final image is a hard requirement for supporting Contextual SBOM - an SBOM that understands the origin of each package.
While intermediate images can be extracted using the `--layers` option, this approach has several issues for our use case:
- Intermediate stage images are unlabeled, making it difficult to determine which image corresponds to which build stage - especially when the Containerfile reuses the same pullspec across multiple stages (pullspecs does not have to be unique, aliases must be).
- All instructions from all intermediate stages appear in the cache (visible via `buildah images --all`), which introduces unnecessary noise for our purposes.
- rootfs.diff_ids are not squashed in final stage: the final-stage image ends up containing diff IDs for every instruction in the final stage (due to the instructions reuse also for final stage). However, we need the final build image to resemble a regular build (without `--layers`), meaning:
  - it should contain the diff IDs inherited from the base image, and
  - exactly one diff ID representing the squashed final-stage instructions.

Related repositories: 
[konflux](https://github.com/konflux-ci/build-definitions) (uses mobster for SBOM generation), 
[mobster](https://github.com/konflux-ci/build-definitions) (implements contextual SBOM functionality requiring this change), 
[capo](https://github.com/konflux-ci/capo) (wraps builder content identification functionality for mobster), 
Contact person: emravec (RedHat) / @ezopezo (Github)

#### How to verify it

Run any multistage build with intermediate stage, with implemented arguments. Resulting intermediate images should be saved and correctly labeled. Example:
`buildah build --save-stages --stage-labels -t test:0.1 .`

Fixes: #6257
Internal Jira: https://issues.redhat.com/browse/ISV-6122

```release-note
Add `--save-stages` and `--stage-labels` flags for preserving and labeling intermediate stage images in multi-stage builds.
```

### DIFF
--- a/define/build.go
+++ b/define/build.go
@@ -287,6 +287,12 @@ type BuildOptions struct {
 	OnBuild []string
 	// Layers tells the builder to commit an image for each step in the Dockerfile.
 	Layers bool
+	// SaveStages tells the builder to save intermediate stage images instead of removing them.
+	SaveStages bool
+	// StageLabels tells the builder to add metadata labels to all stage images (including the final image).
+	// These labels include stage name (stage alias or position) and base image (pullspec or image ID if
+	// stage uses another stage as base). This option requires SaveStages to be enabled.
+	StageLabels bool
 	// NoCache tells the builder to build the image from scratch without checking for a cache.
 	// It creates a new set of cached images for the build.
 	NoCache bool
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -886,6 +886,19 @@ consult the manpages of the selected container runtime.
 Note: Do not pass the leading `--` to the flag. To pass the runc flag `--log-format json`
 to buildah build, the option given would be `--runtime-flag log-format=json`.
 
+**--save-stages** *bool-value*
+
+Preserve intermediate stage images instead of removing them after the build completes
+(Default is `false`). By default, Buildah removes intermediate stage images to save space.
+This option keeps those images, which can be useful for debugging multi-stage builds or
+for reusing intermediate stages in subsequent builds.
+
+`--save-stages` can be used with `--layers` and subsequent builds with `--layers` can use
+the preserved intermediate layers as cache.
+
+When combined with `--stage-labels`, all stage images (including the final image) will
+include metadata labels for easier identification and management.
+
 **--sbom** *preset*
 
 Generate SBOMs (Software Bills Of Materials) for the output image by scanning
@@ -1139,6 +1152,21 @@ The socket path can be left empty to use the value of `default=$SSH_AUTH_SOCK`
 To later use the ssh agent, use the --mount flag in a `RUN` instruction within a `Containerfile`:
 
 `RUN --mount=type=secret,id=id mycmd`
+
+
+**--stage-labels** *bool-value*
+
+Add metadata labels to all intermediate stage images of the multistage build,
+including the final image (Default is `false`).
+This option requires `--save-stages` to be enabled.
+
+When enabled, all intermediate stage images and final image will be labeled with:
+  - `io.buildah.stage.name`: The stage alias (from `FROM ... AS alias`), or stage position
+    if no alias is specified
+  - `io.buildah.stage.base`: The base image used by this stage (pullspec or image ID
+    when stage uses another stage as base)
+
+These labels make it easier to identify, query, and manage images from multi-stage builds.
 
 **--stdin**
 
@@ -1472,6 +1500,14 @@ buildah build -v /var/lib/dnf:/var/lib/dnf:O -t imageName .
 
 buildah build --layers -t imageName .
 
+buildah build --save-stages -t imageName .
+
+buildah build --save-stages --layers -t imageName .
+
+buildah build --save-stages --stage-labels -t imageName .
+
+buildah build --save-stages --stage-labels --layers -t imageName .
+
 buildah build --no-cache -t imageName .
 
 buildah build -f Containerfile --layers --force-rm -t imageName .
@@ -1545,6 +1581,20 @@ buildah build --output type=local,dest=out .
 buildah build --output type=tar,dest=out.tar .
 
 buildah build -o - . > out.tar
+
+### Preserving and querying intermediate stage images
+
+Build a multi-stage image while preserving intermediate stages with metadata labels:
+
+buildah build --save-stages --stage-labels -t myapp .
+
+Find an intermediate image for a specific stage name:
+
+buildah images --filter "label=io.buildah.stage.name=builder"
+
+Find an intermediate image for a specific stage base image:
+
+buildah images --filter "label=io.buildah.stage.base=golang:1.21"
 
 ### Building an image using a URL
 
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -115,6 +115,9 @@ type executor struct {
 	layerLabels                             []string
 	annotations                             []string
 	layers                                  bool
+	saveStages                              bool
+	stageLabels                             bool
+	stageImageIDs                           map[string]string // Tracks image IDs for each stage (by name and position) for label references
 	noHostname                              bool
 	noHosts                                 bool
 	useCache                                bool
@@ -317,6 +320,9 @@ func newExecutor(logger *logrus.Logger, logPrefix string, store storage.Store, o
 		mountLabel:                              mountLabel,
 		annotations:                             slices.Clone(options.Annotations),
 		layers:                                  options.Layers,
+		saveStages:                              options.SaveStages,
+		stageLabels:                             options.StageLabels,
+		stageImageIDs:                           make(map[string]string),
 		noHostname:                              options.CommonBuildOpts.NoHostname,
 		noHosts:                                 options.CommonBuildOpts.NoHosts,
 		useCache:                                !options.NoCache,
@@ -604,6 +610,19 @@ func (b *executor) buildStage(ctx context.Context, cleanupStages map[int]*stageE
 		prependInstructions = slices.Concat([]string{envLine}, prependInstructions)
 	}
 
+	// Create stage labels for all stage images including final stage
+	// Skip if stage has no instructions
+	if b.saveStages && b.stageLabels && len(stage.Node.Children) > 0 {
+		// Wait for base stage if it references a previous stage
+		// This ensures stageImageIDs map is populated before buildStageLabelLine reads it
+		if isStage, err := b.waitForStage(ctx, base, stages[:stageIndex]); isStage && err != nil {
+			return "", nil, false, fmt.Errorf("waiting for base stage %s: %w", base, err)
+		}
+
+		labelLine := b.buildStageLabelLine(&stage, base, stages[:stageIndex])
+		prependInstructions = slices.Concat([]string{labelLine}, prependInstructions)
+	}
+
 	// If we're supposed to be appending or prepending instructions to this stage, add them now.
 	if len(prependInstructions) > 0 {
 		addLines := strings.Join(prependInstructions, "\n")
@@ -660,6 +679,16 @@ func (b *executor) buildStage(ctx context.Context, cleanupStages map[int]*stageE
 		return "", nil, onlyBaseImage, err
 	}
 
+	// Store image ID for this stage so subsequent stages can reference it in labels
+	if imageID != "" {
+		b.stagesLock.Lock()
+		if stage.Name != "" {
+			b.stageImageIDs[stage.Name] = imageID
+		}
+		b.stageImageIDs[strconv.Itoa(stageIndex)] = imageID
+		b.stagesLock.Unlock()
+	}
+
 	// The stage succeeded, so remove its build container if we're
 	// told to delete successful intermediate/build containers for
 	// multi-layered builds.
@@ -690,6 +719,29 @@ func markDependencyStagesForTarget(dependencyMap map[string]*stageDependencyInfo
 			}
 		}
 	}
+}
+
+// buildStageLabelLine creates a LABEL instruction line for stage metadata
+func (b *executor) buildStageLabelLine(stage *imagebuilder.Stage, base string, stages imagebuilder.Stages) string {
+	labelLine := "LABEL"
+	labelLine += fmt.Sprintf(" %q=%q", "io.buildah.stage.name", stage.Name)
+	// Check if base of the stage is another (previous) stage.
+	// If yes, base is set as image ID of this stage.
+	// If not original base name is set (pullspec).
+	for i, st := range stages {
+		if st.Name == base || strconv.Itoa(st.Position) == base {
+			b.stagesLock.Lock()
+			if imgID, ok := b.stageImageIDs[base]; ok {
+				base = imgID
+			} else if imgID, ok := b.stageImageIDs[strconv.Itoa(i)]; ok {
+				base = imgID
+			}
+			b.stagesLock.Unlock()
+			break
+		}
+	}
+	labelLine += fmt.Sprintf(" %q=%q", "io.buildah.stage.base", base)
+	return labelLine
 }
 
 func (b *executor) warnOnUnsetBuildArgs(stages imagebuilder.Stages, dependencyMap map[string]*stageDependencyInfo, args map[string]string) {
@@ -1106,9 +1158,10 @@ func (b *executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 			// We're not populating the cache with intermediate
 			// images, so add this one to the list of images that
 			// we'll remove later.
-			// Only remove intermediate image is `--layers` is not provided
-			// or following stage was not only a base image ( i.e a different image ).
-			if !b.layers && !r.OnlyBaseImage {
+			// Only remove intermediate image if `--layers` is not provided,
+			// `--save-stages` is not enabled, or following stage was not
+			// only a base image (i.e. a different image).
+			if !b.layers && !b.saveStages && !r.OnlyBaseImage {
 				cleanupImages = append(cleanupImages, r.ImageID)
 			}
 		}
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -1552,8 +1552,9 @@ func (s *stageExecutor) execute(ctx context.Context, base string) (imgID string,
 			// so we should commit this container to create
 			// an image, but only if it's the last stage,
 			// or if it's used as the basis for a later
-			// stage.
-			if lastStage || imageIsUsedLater {
+			// stage or we are forcing saving stages by
+			// --save-stages
+			if lastStage || imageIsUsedLater || s.executor.saveStages {
 				logCommit(s.output, i)
 				createdBy, err := s.getCreatedBy(node, addedContentSummary, lastStage && lastInstruction)
 				if err != nil {
--- a/pkg/cli/build.go
+++ b/pkg/cli/build.go
@@ -223,6 +223,10 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
 		return options, nil, nil, errors.New("'rm' and 'force-rm' can only be set with either 'layers' or 'no-cache'")
 	}
 
+	if iopts.StageLabels && !iopts.SaveStages {
+		return options, nil, nil, errors.New(`"--stage-labels" requires "--save-stages"`)
+	}
+
 	if c.Flag("compress").Changed {
 		logrus.Debugf("--compress option specified but is ignored")
 	}
@@ -443,6 +447,7 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
 		Runtime:                 iopts.Runtime,
 		RuntimeArgs:             runtimeFlags,
 		RusageLogFile:           iopts.RusageLogFile,
+		SaveStages:              iopts.SaveStages,
 		SBOMScanOptions:         sbomScanOptions,
 		SignBy:                  iopts.SignBy,
 		SignaturePolicyPath:     iopts.SignaturePolicy,
@@ -450,6 +455,7 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
 		SkipUnusedStages:        skipUnusedStages,
 		SourceDateEpoch:         sourceDateEpoch,
 		Squash:                  iopts.Squash,
+		StageLabels:             iopts.StageLabels,
 		SystemContext:           systemContext,
 		Target:                  iopts.Target,
 		Timestamp:               timestamp,
--- a/pkg/cli/common.go
+++ b/pkg/cli/common.go
@@ -26,8 +26,10 @@ import (
 
 // LayerResults represents the results of the layer flags
 type LayerResults struct {
-	ForceRm bool
-	Layers  bool
+	ForceRm     bool
+	Layers      bool
+	SaveStages  bool
+	StageLabels bool
 }
 
 // UserNSResults represents the results for the UserNS flags
@@ -220,6 +222,8 @@ func GetLayerFlags(flags *LayerResults) pflag.FlagSet {
 	fs := pflag.FlagSet{}
 	fs.BoolVar(&flags.ForceRm, "force-rm", false, "always remove intermediate containers after a build, even if the build is unsuccessful.")
 	fs.BoolVar(&flags.Layers, "layers", UseLayers(), "use intermediate layers during build. Use BUILDAH_LAYERS environment variable to override.")
+	fs.BoolVar(&flags.SaveStages, "save-stages", false, "save intermediate stage images.")
+	fs.BoolVar(&flags.StageLabels, "stage-labels", false, "add metadata labels to intermediate stage images (requires --save-stages).")
 	return fs
 }
 
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -9484,3 +9484,681 @@ _EOF
   # even though different secret was passed to each(baz vs boz), we expect the same result, ie should not affect build history
   diff "${outpath}.a" "${outpath}.b"
 }
+
+@test "[test flags] --stage-labels without --save-stages should error" {
+  _prefetch alpine
+  target="stage-labels-error-test-$(safename)"
+
+  run_buildah 125 build $WITH_POLICY_JSON --stage-labels -t ${target} -f $BUDFILES/save-stages/Dockerfile.single-build-stage $BUDFILES/save-stages
+  expect_output --substring '"--stage-labels" requires "--save-stages"'
+}
+
+@test "[test flags] bud with --save-stages=false does not save intermediate images" {
+  _prefetch alpine
+  target="save-stages-$(safename)"
+
+  run_buildah build $WITH_POLICY_JSON --save-stages=false -t ${target}-without-save -f $BUDFILES/save-stages/Dockerfile.single-build-stage $BUDFILES/save-stages
+  run_buildah images -a
+  baseline_images=$(echo "$output" | awk '$1 == "<none>" && $2 == "<none>" {print $3}')
+  baseline_count=$(echo "$baseline_images" | wc -w)
+  assert "$baseline_count" == "0" "without save-stages should have 0 intermediate images"
+}
+
+@test "[test flags] --save-stages and --stage-labels=false does not label intermediate images" {
+  _prefetch alpine
+  target="save-stages-$(safename)"
+
+  run_buildah build $WITH_POLICY_JSON --save-stages --stage-labels=false -t ${target} -f $BUDFILES/save-stages/Dockerfile.single-build-stage $BUDFILES/save-stages
+  run_buildah images -a
+  intermediate_images=$(echo "$output" | awk '$1 == "<none>" && $2 == "<none>" {print $3}')
+  intermediate_count=$(echo "$intermediate_images" | wc -w)
+  assert "$intermediate_count" == "1" "should have 1 intermediate image"
+
+  for image_id in $intermediate_images; do
+    run_buildah inspect --format '{{index .OCIv1.Config.Labels "io.buildah.stage.name"}}' $image_id
+    assert "$output" == "" "stage.name label should not be present when --stage-labels=false"
+
+    run_buildah inspect --format '{{index .OCIv1.Config.Labels "io.buildah.stage.base"}}' $image_id
+    assert "$output" == "" "stage.base label should not be present when --stage-labels=false"
+  done
+}
+
+@test "[test saving intermediate images] --save-stages on single-stage Dockerfile should not create intermediate images" {
+  _prefetch alpine
+  target="single-stage-test-$(safename)"
+
+  run_buildah build $WITH_POLICY_JSON --save-stages -t ${target} -f $BUDFILES/save-stages/Dockerfile.simple $BUDFILES/save-stages
+
+  run_buildah images -a
+  intermediate_images=$(echo "$output" | awk '$1 == "<none>" && $2 == "<none>" {print $3}')
+  intermediate_count=$(echo "$intermediate_images" | wc -w)
+  assert "$intermediate_count" == "0" "single-stage build should not create intermediate images"
+
+  run_buildah rmi ${target}
+}
+
+@test "[test saving intermediate images] --save-stages preserves intermediate image and skips unused" {
+  _prefetch alpine
+  target="save-stages-$(safename)"
+
+  # Build without --save-stages (baseline - should produce 0 intermediate images)
+  run_buildah build $WITH_POLICY_JSON -t ${target}-no-save -f $BUDFILES/save-stages/Dockerfile.two-build-stages-one-unused $BUDFILES/save-stages
+  run_buildah images -a
+  baseline_images=$(echo "$output" | awk '$1 == "<none>" && $2 == "<none>" {print $3}')
+  baseline_count=$(echo "$baseline_images" | wc -w)
+  assert "$baseline_count" == "0" "without save-stages should have 0 intermediate images"
+  run_buildah rmi ${target}-no-save
+
+  # Build with --save-stages (should produce 1 intermediate image: 'intermediate' stage)
+  # Note: 'builder' stage is skipped due to --skip-unused-stages=true (default)
+  run_buildah build $WITH_POLICY_JSON --save-stages -t ${target} -f $BUDFILES/save-stages/Dockerfile.two-build-stages-one-unused $BUDFILES/save-stages
+  run_buildah images -a
+  save_images=$(echo "$output" | awk '$1 == "<none>" && $2 == "<none>" {print $3}')
+  save_count=$(echo "$save_images" | wc -w)
+  assert "$save_count" == "1" "with save-stages should have 1 intermediate image (intermediate stage)"
+}
+
+@test "[test saving intermediate images] --save-stages preserves intermediate images for all stages" {
+  _prefetch alpine
+  target="save-stages-all-test-$(safename)"
+
+  # Build without --save-stages (baseline - should produce 0 intermediate images)
+  run_buildah build $WITH_POLICY_JSON -t ${target}-no-save -f $BUDFILES/save-stages/Dockerfile.three-build-stages-parent-child-independent $BUDFILES/save-stages
+  run_buildah images -a
+  baseline_images=$(echo "$output" | awk '$1 == "<none>" && $2 == "<none>" {print $3}')
+  baseline_count=$(echo "$baseline_images" | wc -w)
+  assert "$baseline_count" == "0" "without save-stages should have 0 intermediate images"
+  run_buildah rmi ${target}-no-save
+
+  # Build with --save-stages (should produce 3 intermediate images: parent, child, independent)
+  # This demonstrates that intermediate images are saved for ALL stages: for each in parent-child chain and independent stage
+  run_buildah build $WITH_POLICY_JSON --save-stages -t ${target} -f $BUDFILES/save-stages/Dockerfile.three-build-stages-parent-child-independent $BUDFILES/save-stages
+  run_buildah images -a
+  save_images=$(echo "$output" | awk '$1 == "<none>" && $2 == "<none>" {print $3}')
+  save_count=$(echo "$save_images" | wc -w)
+  assert "$save_count" == "3" "with save-stages should have 3 intermediate images (parent, child, independent)"
+}
+
+@test "[test labeling intermediate images] --stage-labels adds metadata to config and history of the intermediate image and final image" {
+  _prefetch alpine
+  target="stage-labels-test-$(safename)"
+
+  run_buildah build $WITH_POLICY_JSON --save-stages --stage-labels -t ${target} -f $BUDFILES/save-stages/Dockerfile.single-build-stage $BUDFILES/save-stages
+  run_buildah images -a
+  intermediate_image=$(echo "$output" | awk '$1 == "<none>" && $2 == "<none>" {print $3}')
+  intermediate_count=$(echo "$intermediate_image" | wc -w)
+
+  assert "$intermediate_count" == "1" "should have exactly 1 intermediate image"
+
+  run_buildah inspect --format '{{index .OCIv1.Config.Labels "io.buildah.stage.name"}}' $intermediate_image
+  assert "$output" "==" "intermediate" "stage.name should be 'intermediate'"
+
+  run_buildah inspect --format '{{index .OCIv1.Config.Labels "io.buildah.stage.base"}}' $intermediate_image
+  assert "$output" "=~" "alpine" "stage.base should reference alpine image"
+
+  # Verify LABEL instruction appears in history (not just in config.Labels)
+  run_buildah inspect --format '{{range .OCIv1.History}}{{.CreatedBy}}{{"\n"}}{{end}}' $intermediate_image
+  expect_output --substring "LABEL"
+
+  # Verify exact label names and values in history
+  expect_output --substring '"io.buildah.stage.name"="intermediate"' "history should contain exact stage.name label"
+  expect_output --substring '"io.buildah.stage.base"="alpine"' "history should contain exact stage.base label"
+
+  # Verify final image also has stage labels (stage position and base pullspec)
+  run_buildah inspect --format '{{index .OCIv1.Config.Labels "io.buildah.stage.name"}}' ${target}
+  assert "$output" "==" "1" "final stage.name should be stage position '1' (no alias)"
+
+  run_buildah inspect --format '{{index .OCIv1.Config.Labels "io.buildah.stage.base"}}' ${target}
+  assert "$output" "==" "alpine" "final stage.base should be 'alpine' pullspec"
+}
+
+@test "[test labeling intermediate images] chain of two stages - labels are indicating that one stage is using another stage as base" {
+  _prefetch alpine
+  target="chained-stages-test-$(safename)"
+
+  run_buildah build $WITH_POLICY_JSON --save-stages --stage-labels -t ${target} -f $BUDFILES/save-stages/Dockerfile.chained-two-build-stages $BUDFILES/save-stages
+
+  run_buildah images -a
+  all_intermediate_images=$(echo "$output" | awk '$1 == "<none>" && $2 == "<none>" {print $3}')
+
+  parent_image_id=""
+  parent_name=""
+  parent_base=""
+  child_image_id=""
+  child_name=""
+  child_base=""
+
+  for image_id in $all_intermediate_images; do
+    run_buildah inspect --format '{{index .OCIv1.Config.Labels "io.buildah.stage.name"}}' $image_id
+    name="$output"
+
+    run_buildah inspect --format '{{index .OCIv1.Config.Labels "io.buildah.stage.base"}}' $image_id
+    base="$output"
+
+    # Check if base is in the list of intermediate images (stage is used by another stage as base)
+    is_child=false
+    for iid in $all_intermediate_images; do
+      # base in label is full image ID, but all_intermediate_images contains short IDs, so check prefix
+      if [[ "$base" == "$iid"* ]]; then
+        is_child=true
+        break
+      fi
+    done
+
+    if [ "$is_child" = true ]; then
+      # This is a child stage: its base is another intermediate image
+      child_image_id="$image_id"
+      child_name="$name"
+      child_base="$base"
+    else
+      # This is a parent stage that serves as base for another stage
+      parent_image_id="$image_id"
+      parent_name="$name"
+      parent_base="$base"
+    fi
+  done
+
+  assert "$parent_image_id" "!=" "" "parent stage should be present"
+  assert "$child_image_id" "!=" "" "child stage should be present"
+
+  assert "$parent_name" "==" "parent-stage" "parent stage name should be 'parent-stage'"
+  assert "$child_name" "==" "child-stage" "child stage name should be 'child-stage'"
+
+  assert "$parent_base" "=~" "alpine" "parent stage.base should reference alpine image"
+
+  # Verify child stage uses parent as base (child_base contains full ID, parent_image_id contains short ID)
+  assert "$child_base" "=~" "^$parent_image_id" "child stage.base should start with parent image ID"
+}
+
+@test "[test labeling intermediate images] --jobs 1 and 2 correctly resolve stage references to imageID in labels" {
+  _prefetch alpine
+
+  for jobs in 1 2; do
+    target="imageid-resolution-jobs${jobs}-$(safename)"
+
+    # Build with specified --jobs value
+    run_buildah build $WITH_POLICY_JSON --jobs $jobs --save-stages --stage-labels --layers -t ${target} -f $BUDFILES/save-stages/Dockerfile.chained-two-build-stages $BUDFILES/save-stages
+
+    run_buildah images -a
+    all_intermediate_images=$(echo "$output" | awk '$1 == "<none>" && $2 == "<none>" {print $3}')
+
+    # Find parent-stage and child-stage intermediate images
+    parent_image=""
+    child_image=""
+    for image_id in $all_intermediate_images; do
+      run_buildah inspect --format '{{index .OCIv1.Config.Labels "io.buildah.stage.name"}}' $image_id
+      if [ "$output" = "parent-stage" ]; then
+        parent_image="$image_id"
+      elif [ "$output" = "child-stage" ]; then
+        child_image="$image_id"
+      fi
+    done
+
+    assert "$parent_image" "!=" "" "parent-stage intermediate image should exist (jobs=$jobs)"
+    assert "$child_image" "!=" "" "child-stage intermediate image should exist (jobs=$jobs)"
+
+    # Get child-stage base label value
+    run_buildah inspect --format '{{index .OCIv1.Config.Labels "io.buildah.stage.base"}}' $child_image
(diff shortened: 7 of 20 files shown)


ezopezo:
@nalind can you please take a look and put ok-to-test label? It _seems_ to me that tests are failing most likely with some timeouts and thus I would like to try to re-run them (or please tell me what I just broke :) ).

ezopezo:
@nalind @mtrmac @TomSweeneyRedHat can you please take a look on this? (or pick up some appropriate reviewers?) Thanks in advance!

mtrmac:
@flouthoc if you have time

TomSweeneyRedHat:
@Luap99 @nalind PTAL

docs/buildah-build.1.md
```diff
@@ -136,6 +143,19 @@ the intermediate image is stored in the image itself. Buildah's approach is simi
 does not inflate the size of the original image with intermediate images.  Also, intermediate images can truly be
 kept distributed across one or more remote registries using Buildah's caching mechanism.
 
+**--cache-stages** *bool-value*
+
+Preserve intermediate stage images instead of removing them after the build completes
+(Default is `false`). By default, buildah removes intermediate stage images to save space.
```
TomSweeneyRedHat:
```suggestion
(Default is `false`). By default, Buildah removes intermediate stage images to save space.
```

docs/buildah-build.1.md
```diff
@@ -1385,6 +1420,10 @@ buildah build -v /var/lib/dnf:/var/lib/dnf:O -t imageName .
 
 buildah build --layers -t imageName .
 
+buildah build --cache-stages --stage-labels -t imageName .
```
TomSweeneyRedHat: just to show the bool values in play, perhaps
```suggestion
buildah build --cache-stages=false --stage-labels=false -t imageName .
```

docs/buildah-build.1.md
```diff
@@ -1443,6 +1482,21 @@ buildah build --output type=tar,dest=out.tar .
...
+
+Query intermediate images from a specific build using the build ID:
+
+BUILD_ID=$(cat /tmp/build-id.txt)
+buildah images --filter "label=io.buildah.build.id=${BUILD_ID}"
+
+Find an intermediate image for a specific stage name:
+
+buildah images --filter "label=io.buildah.stage.name=builder"
+
```
TomSweeneyRedHat, line 1594: This is great!  TY!

pkg/cli/build.go
```diff
@@ -417,6 +426,8 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
 		Labels:                  iopts.Label,
 		LayerLabels:             iopts.LayerLabel,
 		Layers:                  layers,
+		CacheStages:             iopts.CacheStages,
+		StageLabels:             iopts.StageLabels,
```
TomSweeneyRedHat: Please add these three options in alphabetically.    I.e. CacheStages at line 396.5 and so on.

pkg/cli/common.go
```diff
@@ -73,6 +75,7 @@ type BudResults struct {
 	Format              string
 	From                string
 	Iidfile             string
+	BuildIDFile         string
```
TomSweeneyRedHat: To keep making my OCD happy, please move this to line 63.5

pkg/cli/common.go
```diff
@@ -26,8 +26,10 @@ import (
 
 // LayerResults represents the results of the layer flags
 type LayerResults struct {
-	ForceRm bool
-	Layers  bool
+	ForceRm     bool
+	Layers      bool
+	CacheStages bool
```
TomSweeneyRedHat: More OCD alpha changes, please put this at line 28.5

tests/bud.bats
```diff
@@ -9059,3 +9059,357 @@ EOF
...
+
+  assert "$cache_stages_count" == "$normal_count" "--cache-stages should have same layer count as normal build"
+  assert "$cache_stages_count" == "2" "--cache-stages build should have exactly 2 layers"
+  assert "$normal_count" == "2" "normal build should have exactly 2 layers"
+
+  assert "$layers_count" == "4" "--layers should create exactly 4 layers (one per instruction)"
+
+  run_buildah rmi --all
+}
+
```
TomSweeneyRedHat, line 9892: can you update and/or add tests to use "true" and "false" for the boolean flags?  Leave at least one using the default true for each too.

TomSweeneyRedHat (commented):
First pass LGTM, just a couple of nitty things here and there.

@nalind or @Luap99 PTAL

Also, have you tried vendoring Buildah into Podman with these changes yet?


@ddarrah should we have a feature/epic for this in RHEL 9.8/10.2?

docs/buildah-build.1.md
ezopezo: I skipped that since this is a relatively uncommon use case for boolean flags, because the false behavior is implicit whenever the flag is absent. Other flags such as `--no-cache` or `--layers` follow the same pattern (present when true, absent when not, without presence of the `--layers false`) so I wanted to stay consistent with your docs. If you feel strongly about it, I certainly can add the examples :)

tests/bud.bats
ezopezo: For `false` I added two new tests for `--cache-stages` and `--stage-labels` and for `true` I updated one existing test :+1:

imagebuildah/executor.go
```diff
@@ -112,6 +113,11 @@ type executor struct {
 	layerLabels                             []string
 	annotations                             []string
 	layers                                  bool
+	cacheStages                             bool
+	stageLabels                             bool
+	buildID                                 string
+	buildIDFile                             string
```
nalind: `buildIDFile` doesn't appear to be read after it's set.

define/build.go
```diff
@@ -278,6 +280,12 @@ type BuildOptions struct {
 	OnBuild []string
 	// Layers tells the builder to commit an image for each step in the Dockerfile.
 	Layers bool
+	// CacheStages tells the builder to preserve intermediate stage images instead of removing them.
+	CacheStages bool
+	// StageLabels tells the builder to add metadata labels to intermediate stage images for easier recognition.
+	// These labels include stage name, base image, build ID, and parent stage name (when a stage uses another
+	// intermediate stage as its base, i.e., transitive aliases). This option requires CacheStages to be enabled.
```
nalind: Even as it's being defined, "transitive aliases" doesn't make any sense to me as jargon.  Does the term come from somewhere else?

docs/buildah-build.1.md
nalind: There's a long-standing bit of confusion in the man pages where, even though the boolean argument values are optional, we don't suggest that they always have to be supplied in the --flag=value form, with an equal sign, to prevent them from being treated as unrelated arguments.

imagebuildah/stage_executor.go
```diff
@@ -1837,6 +1843,27 @@ func (s *stageExecutor) execute(ctx context.Context, base string) (imgID string,
...
+	if s.executor.cacheStages && !lastStage {
+		// Check if this stage is used as base by another intermediate stage
+		_, isParentStage := s.executor.intermediateStageParents[s.name]
+		if isParentStage {
+			logrus.Debugf("Skipping commit for intermediate stage %s (index %d) - used as base by another intermediate stage", s.name, s.index)
+		} else {
+			logrus.Debugf("Committing intermediate stage %s (index %d) for --cache-stages", s.name, s.index)
+			createdBy := fmt.Sprintf("/bin/sh -c #(nop) STAGE %s", s.name)
+			// Commit the stage without squashing, using empty output name (intermediate image)
+			imgID, commitResults, err = s.commit(ctx, createdBy, false, "", false, false)
```
nalind: Why set `emptyLayer` as `false` here?

nalind (commented):
Overall I think I understand the goals here.  The new flags interact with `--layers` in confusing ways, though - the build ID recorded for `--build-id-file` won't match the image whose ID we return if the build is a cache hit, but we may quietly commit new versions of the final product of intermediate stages.  I'm not really sure what the intended result is there.

define/build.go
ezopezo: Fair point - you're right that “transitive alias” isn't established jargon. I started using it internally as shorthand for multi-stage builds where one named stage is used as the base for another, and it unfortunately leaked here. I think I picked it up from an article or docs at some point, but I can't find a reference now - which indeed makes it sound like it was revealed to me in a dream 🙂
I'll happily rephrase all occurrences using some more standard terminology!

docs/buildah-build.1.md
ezopezo: @nalind Does it make a sense to you to add args with equal sign and boolean or leave it as it is? What do you think?

ezopezo:
> Overall I think I understand the goals here. The new flags interact with `--layers` in confusing ways, though - the build ID recorded for `--build-id-file` won't match the image whose ID we return if the build is a cache hit, but we may quietly commit new versions of the final product of intermediate stages. I'm not really sure what the intended result is there.

Does it make a sense to add an mutual exclusion condition that --layers and --cache-stages cannot be used together (as serve fundamentally different purposes)? I want to keep this functionality as isolated as possible from others to avoid unnecessary complexities like this.

imagebuildah/executor.go
ezopezo: Removed, thanks!

imagebuildah/stage_executor.go
ezopezo: You re right - it should not be false. I updated it such way, if stage was already committed, we only add metadata labels without creating a duplicate empty layer. Let me know, if it is okay!

ezopezo:
@nalind Would you be able to take an another look please? Thanks in advance!

define/build.go
ezopezo: Done :)

imagebuildah/stage_executor.go
nalind: Actually, I'm thinking that I was confused here - there are cases where we don't commit after the final instruction in a stage, since we're not going to base later stages off of it, and the working container is sufficient for `COPY --from` references to it later (for example, when `imageIsUsedLater` is `false` in the single-layer build case).  I _think_ we're okay with `false` for multi-layer builds, but should probably give it a once-over again.

nalind:
`--layers=true` is the default used by `podman build`, so I'd be hesitant about making a new feature mutually exclusive with it.
If we weren't worried about reusing cache images with previously-generated build UUIDs, I would generally expect the images we commit for each stage, including the final one, to use multiple layers when it's appropriate.
Would forcing `checkForLayers` to be `false` at the start of `stageExecutor.execute()` work as expected for the use case?  I could see an argument for "fresh build, every time" when the new flag is used, and that would still allow subsequent builds that don't use the new flag to use the results as cache hits.  If we did that, would the cache-related labels be left alone or altered/cleared for the products of that build?

imagebuildah/stage_executor.go
ezopezo: My aim with `emptyLayer := imgID != ""` is to not duplicate an empty layer created by the previous commit that already occurred in the execute function. 

When `imgID != ""` (this means to me that stage was already committed):
- Regular builds (without –layers) commit the stage in execute function if `imageIsUsedLater=true`
- Also now with `--layers`, every instruction is commited, so imgID is set
I think, in these cases, we only need to add labels by commit and `emptyLayer=true` should prevent creating a duplicate layer in the intermediate image (right?). 

When `imgID == ""` (I grasp this as stage was not committed and is left as working container) regular builds with `imageIsUsedLater=false` don't commit the stage, but we need to capture filesystem changes of the working container- here is `emptyLayer=false` okay because ensures we create a proper layer with the stage's content..

I think this prevents duplicate layers while correctly capturing filesystem changes when the stage hasn't been committed yet (and we need them). 
Let me know if there's an edge case I'm missing or I got tangled into this too much and I am wrong 😀

ezopezo:
> `--layers=true` is the default used by `podman build`, so I'd be hesitant about making a new feature mutually exclusive with it. If we weren't worried about reusing cache images with previously-generated build UUIDs, I would generally expect the images we commit for each stage, including the final one, to use multiple layers when it's appropriate. Would forcing `checkForLayers` to be `false` at the start of `stageExecutor.execute()` work as expected for the use case? I could see an argument for "fresh build, every time" when the new flag is used, and that would still allow subsequent builds that don't use the new flag to use the results as cache hits. If we did that, would the cache-related labels be left alone or altered/cleared for the products of that build?

Yes, it works well `buildah build --layers --cache-stages --stage-labels` stores all the layers of the Containerfile, correctly labels them and subsequent plain `buildah build --layers` reuses those by hitting cache - those layers remain intact, labels are preserved. Are we okay with such behavior?

ezopezo:
@nalind Also I almost forgot to ask - how I can effectively test if current changes are compatible with podman? 
(I believe this is what @TomSweeneyRedHat asked before as well):
 
> Also, have you tried vendoring Buildah into Podman with these changes yet?

nalind:
> @nalind Also I almost forgot to ask - how I can effectively test if current changes are compatible with podman? (I believe this is what @TomSweeneyRedHat asked before as well): 

Since we're on `main`, check out podman's `main` branch, and use `go mod edit -replace github.com/containers/buildah=github.com/ezopezo/buildah@478bf2159f6ffe878c5162ed1d83c69ae3e6a6d6 && go mod tidy && go mod vendor && make` to build.  (If your branch name was a valid version string, you could use its name directly instead of having to look up the commit ID.)  At that point you can either attempt to test locally with `make .install.ginkgo localintegration` and related targets, or commit everything to a branch, push it to a podman fork, and open a PR to have podman's CI do it, taking care to mark it as not intended for merging.

ezopezo:
> > @nalind Also I almost forgot to ask - how I can effectively test if current changes are compatible with podman? (I believe this is what @TomSweeneyRedHat asked before as well):
> 
> Since we're on `main`, check out podman's `main` branch, and use `go mod edit -replace github.com/containers/buildah=github.com/ezopezo/buildah@478bf2159f6ffe878c5162ed1d83c69ae3e6a6d6 && go mod tidy && go mod vendor && make` to build. (If your branch name was a valid version string, you could use its name directly instead of having to look up the commit ID.) At that point you can either attempt to test locally with `make .install.ginkgo localintegration` and related targets, or commit everything to a branch, push it to a podman fork, and open a PR to have podman's CI do it, taking care to mark it as not intended for merging.

@nalind Thank you I did it according your guide and created [this testing PR](https://github.com/containers/podman/pull/27964). I think it is failing because of the docs mismatch and exceeding some limit of the binary size (which could be overridden by label I think) and also I think I need from maintainers approval for running all of the jobs, could you help me with this please?

nalind:
I've given it a kick.

docs/buildah-build.1.md
```diff
@@ -1061,6 +1083,21 @@ To later use the ssh agent, use the --mount flag in a `RUN` instruction within a
...
+**--stage-labels** *bool-value*
+
+Add metadata labels to intermediate stage images (Default is `false`). This option
+requires `--cache-stages` to be enabled.
+
+When enabled, intermediate stage images will be labeled with:
+- `io.buildah.stage.name`: The stage name (from `FROM ... AS name`)
+- `io.buildah.stage.base`: The base image used by this stage
+- `io.buildah.stage.parent_name`: The parent stage name (if this stage uses another stage as base)
+- `io.buildah.build.id`: A unique build ID shared across all stages in a single build
```
nalind: These need to be indented so that the man pages end up with them on separate lines.

ezopezo: Thanks, resolved!

ezopezo:
@nalind I am re-pushing changes, because /retest flag just does not work for me (maybe I need some privileges?) and checks are often sporadically failing. 
Also I don't know how to proceed with podman vendoring test I left a comment in [PR](https://github.com/containers/podman/pull/27964). Can you please help me?

nalind:
> @nalind I am re-pushing changes, because /retest flag just does not work for me (maybe I need some privileges?) and checks are often sporadically failing. Also I don't know how to proceed with podman vendoring test I left a comment in [PR](https://github.com/containers/podman/pull/27964). Can you please help me?

The configuration for the bot which handled those things was updated to not drop most containers-org repositories by https://github.com/openshift/release/pull/71397; anything that was in-progress before it was merged would still include a record of previous interactions with the bot.

docs/buildah-build.1.md
nalind: If we want an example where the flag is set to `false`, the equal sign is going to be necessary.

nalind:
> > `--layers=true` is the default used by `podman build`, so I'd be hesitant about making a new feature mutually exclusive with it. If we weren't worried about reusing cache images with previously-generated build UUIDs, I would generally expect the images we commit for each stage, including the final one, to use multiple layers when it's appropriate. Would forcing `checkForLayers` to be `false` at the start of `stageExecutor.execute()` work as expected for the use case? I could see an argument for "fresh build, every time" when the new flag is used, and that would still allow subsequent builds that don't use the new flag to use the results as cache hits. If we did that, would the cache-related labels be left alone or altered/cleared for the products of that build?
> 
> Yes, it works well `buildah build --layers --cache-stages --stage-labels` stores all the layers of the Containerfile, correctly labels them and subsequent plain `buildah build --layers` reuses those by hitting cache - those layers remain intact, labels are preserved. Are we okay with such behavior?

I had to think about this for a bit, and I'm not sure it's what we want.  Much like earlier versions of `build --label`, adding the new labels when we commit, without having their presence show up in the history, prevents the cache evaluation logic from "seeing" them.

For the `--label` CLI flag, we eventually reworked the build so that it internally appended a LABEL instruction with its arguments to the set of build steps in `executor.buildStage()` before processing the steps for the stage, so that labels from the command line could be accounted for.

Would this use case be better served by doing something similar?  The `buildStage()` method potentially appends a LABEL instruction to the end of the final stage, and potentially prepends an ENV instruction (for `--env` values) at the beginning of every stage, so we have options.  Adding the stage label instruction at the start of the stage's instructions would cause cache misses and naturally provide the "fresh build, every time" behavior which I think we're aiming for.  Adding it as an instruction at the end of the stage would allow every instruction before it to be a cache hit, if we want it the other way.

ezopezo:
> > > `--layers=true` is the default used by `podman build`, so I'd be hesitant about making a new feature mutually exclusive with it. If we weren't worried about reusing cache images with previously-generated build UUIDs, I would generally expect the images we commit for each stage, including the final one, to use multiple layers when it's appropriate. Would forcing `checkForLayers` to be `false` at the start of `stageExecutor.execute()` work as expected for the use case? I could see an argument for "fresh build, every time" when the new flag is used, and that would still allow subsequent builds that don't use the new flag to use the results as cache hits. If we did that, would the cache-related labels be left alone or altered/cleared for the products of that build?
> > 
> > 
> > Yes, it works well `buildah build --layers --cache-stages --stage-labels` stores all the layers of the Containerfile, correctly labels them and subsequent plain `buildah build --layers` reuses those by hitting cache - those layers remain intact, labels are preserved. Are we okay with such behavior?
> 
> I had to think about this for a bit, and I'm not sure it's what we want. Much like earlier versions of `build --label`, adding the new labels when we commit, without having their presence show up in the history, prevents the cache evaluation logic from "seeing" them.
> 
> For the `--label` CLI flag, we eventually reworked the build so that it internally appended a LABEL instruction with its arguments to the set of build steps in `executor.buildStage()` before processing the steps for the stage, so that labels from the command line could be accounted for.
> 
> Would this use case be better served by doing something similar? The `buildStage()` method potentially appends a LABEL instruction to the end of the final stage, and potentially prepends an ENV instruction (for `--env` values) at the beginning of every stage, so we have options. Adding the stage label instruction at the start of the stage's instructions would cause cache misses and naturally provide the "fresh build, every time" behavior which I think we're aiming for. Adding it as an instruction at the end of the stage would allow every instruction before it to be a cache hit, if we want it the other way.


> prevents the cache evaluation logic from "seeing" them.

I may be misunderstanding this concern. Our implementation explicitly disables cache lookup when --cache-stages is used (stage_executor.go:1260-1262):

```
if s.executor.cacheStages {
    checkForLayers = false
}

```
This ensures every build with --cache-stages is a fresh build, and caches new intermediate images with a new build UUID. Cache evaluation logic (search) for stages doesn't run at all when --cache-stages is used.

Regarding cache reuse scenarios:
Scenario 1:
  - Build 1: `--cache-stages` (without `--layers`) → creates non-layered intermediate images
  - Build 2:  `--cache-stages` (without `--layers`) → CANNOT reuse (explicitly disabled cache lookup)
 
Scenario 2:
  - Build 1: `--cache-stages` (without `--layers`) → creates non-layered intermediate images
  - Build 2: (without --cache-stages and `--layers`) → CANNOT reuse (insufficient history for cache hit, if I understand logic correctly), but that's fine for our use case
 
Scenario 3 (considering that podman has always `--layers` enabled):
  - Build 1: `--cache-stages --layers` → creates layered intermediate images
  - Build 2: `--layers` (without `--cache-stages`) → CAN reuse those layers (you're right - even when labeled, because of their absence in history - I'll expand this later*)

However, our actual use case (Konflux SBOM generation) will use --cache-stages without --layers, so the layered cache scenario isn't relevant for us.
The --layers compatibility was added based on your earlier suggestion that it's podman's default, but our production workflow won't use it.
The intermediate images from --cache-stages builds are intended for inspection (SBOM generation), not for cache reuse. We don't need subsequent builds to reuse them.

> For the --label CLI flag...

*I see the architectural similarity, but there's a practical difference IMHO:
  - --label: user-defined, cache evaluation is active (user applies different labels across builds = cache miss)
  - --stage-labels: system metadata, cache evaluation is explicitly disabled (checkForLayers = false by mandatory preceeding --cache-stages)

Given that cache is disabled when --cache-stages is used, is there still a concern about labels (added by --stage labels, that cannot be used without --cache-stages) not being in build history?

Could you clarify what specific scenario you're concerned about where the current approach would fail?

ezopezo:
> > @nalind I am re-pushing changes, because /retest flag just does not work for me (maybe I need some privileges?) and checks are often sporadically failing. Also I don't know how to proceed with podman vendoring test I left a comment in [PR](https://github.com/containers/podman/pull/27964). Can you please help me?
> 
> The configuration for the bot which handled those things was updated to not drop most containers-org repositories by [openshift/release#71397](https://github.com/openshift/release/pull/71397); anything that was in-progress before it was merged would still include a record of previous interactions with the bot.

@nalind  I understand, thank you for explanation! I am just repushing commit with chnages to trigger pipeline - is it common that I am getting so much failures from timeouts? It is very hard for me to get all-green (I wittnessed it few times) :|

nalind:
> > > > `--layers=true` is the default used by `podman build`, so I'd be hesitant about making a new feature mutually exclusive with it. If we weren't worried about reusing cache images with previously-generated build UUIDs, I would generally expect the images we commit for each stage, including the final one, to use multiple layers when it's appropriate. Would forcing `checkForLayers` to be `false` at the start of `stageExecutor.execute()` work as expected for the use case? I could see an argument for "fresh build, every time" when the new flag is used, and that would still allow subsequent builds that don't use the new flag to use the results as cache hits. If we did that, would the cache-related labels be left alone or altered/cleared for the products of that build?
> > > 
> > > 
> > > Yes, it works well `buildah build --layers --cache-stages --stage-labels` stores all the layers of the Containerfile, correctly labels them and subsequent plain `buildah build --layers` reuses those by hitting cache - those layers remain intact, labels are preserved. Are we okay with such behavior?
> > 
> > 
> > I had to think about this for a bit, and I'm not sure it's what we want. Much like earlier versions of `build --label`, adding the new labels when we commit, without having their presence show up in the history, prevents the cache evaluation logic from "seeing" them.
> > For the `--label` CLI flag, we eventually reworked the build so that it internally appended a LABEL instruction with its arguments to the set of build steps in `executor.buildStage()` before processing the steps for the stage, so that labels from the command line could be accounted for.
> > Would this use case be better served by doing something similar? The `buildStage()` method potentially appends a LABEL instruction to the end of the final stage, and potentially prepends an ENV instruction (for `--env` values) at the beginning of every stage, so we have options. Adding the stage label instruction at the start of the stage's instructions would cause cache misses and naturally provide the "fresh build, every time" behavior which I think we're aiming for. Adding it as an instruction at the end of the stage would allow every instruction before it to be a cache hit, if we want it the other way.
> 
> > prevents the cache evaluation logic from "seeing" them.
> 
> I may be misunderstanding this concern. Our implementation explicitly disables cache lookup when --cache-stages is used (stage_executor.go:1260-1262):
> 
> ```
> if s.executor.cacheStages {
>     checkForLayers = false
> }
> ```
> 
> This ensures every build with --cache-stages is a fresh build, and caches new intermediate images with a new build UUID. Cache evaluation logic (search) for stages doesn't run at all when --cache-stages is used.

Yes, I was hoping to be able to remove this special-case check.

> Regarding cache reuse scenarios: Scenario 1:
> 
>     * Build 1: `--cache-stages` (without `--layers`) → creates non-layered intermediate images
> 
>     * Build 2:  `--cache-stages` (without `--layers`) → CANNOT reuse (explicitly disabled cache lookup)
> 
> Scenario 2:
> 
>     * Build 1: `--cache-stages` (without `--layers`) → creates non-layered intermediate images
> 
>     * Build 2: (without --cache-stages and `--layers`) → CANNOT reuse (insufficient history for cache hit, if I understand logic correctly), but that's fine for our use case

Your understand matches mine, yes.

> Scenario 3 (considering that podman has always `--layers` enabled):
> 
>     * Build 1: `--cache-stages --layers` → creates layered intermediate images
> 
>     * Build 2: `--layers` (without `--cache-stages`) → CAN reuse those layers (you're right - even when labeled, because of their absence in history - I'll expand this later*)
> 
> However, our actual use case (Konflux SBOM generation) will use --cache-stages without --layers, so the layered cache scenario isn't relevant for us. The --layers compatibility was added based on your earlier suggestion that it's podman's default, but our production workflow won't use it. The intermediate images from --cache-stages builds are intended for inspection (SBOM generation), not for cache reuse. We don't need subsequent builds to reuse them.
> 
> > For the --label CLI flag...
> 
> *I see the architectural similarity, but there's a practical difference IMHO:
> 
>     * --label: user-defined, cache evaluation is active (user applies different labels across builds = cache miss)
> 
>     * --stage-labels: system metadata, cache evaluation is explicitly disabled (checkForLayers = false by mandatory preceeding --cache-stages)

There's really nothing enforcing that distinction, though, and I would advise against assuming otherwise when using the produced images.

> Given that cache is disabled when --cache-stages is used, is there still a concern about labels (added by --stage labels, that cannot be used without --cache-stages) not being in build history?
> 
> Could you clarify what specific scenario you're concerned about where the current approach would fail?

Throwing the stage labels into the history makes the new feature's interaction with caching much easier to reason about - based on where in the parsed tree the new labels are added, we can predict how the produced images will interact with the cache evaluation logic in future builds, and I value that.

imagebuildah/stage_executor.go
```diff
@@ -2606,6 +2607,14 @@ func (s *stageExecutor) commit(ctx context.Context, createdBy string, emptyLayer
...
 		s.builder.SetLabel(k, v)
 	}
+	// Clear any inherited stage labels from final image when it uses an intermediate stage as base.
+	// Final stage should not have stage metadata labels even if it inherits from an intermediate stage.
+	// Example: "FROM base" where base is an intermediate stage (and this is last FROM instruction)
+	// will not inherit base's stage labels.
+	if finalInstruction && s.executor.saveStages && s.executor.stageLabels {
+		s.builder.UnsetLabel("io.buildah.stage.name")
+		s.builder.UnsetLabel("io.buildah.stage.base")
+	}
```
nalind: We're going to want to avoid clearing labels like this.  Cache evaluation won't notice that these labels are missing, so a second build that adds a new final instruction will treat this as a cache hit, when it shouldn't.
Unless it creates problems somewhere else, I'm not really sure what benefit is gained by not having these labels in the final image, now that we've gone to the trouble to note their addition in the image's history.

imagebuildah/executor.go
```diff
@@ -692,6 +713,32 @@ func markDependencyStagesForTarget(dependencyMap map[string]*stageDependencyInfo
...
+				base = imgID
+			}
+			break
+		}
+	}
+	if base != "" {
+		labelLine += fmt.Sprintf(" %q=%q", "io.buildah.stage.base", base)
+	}
+	return labelLine
+}
```
nalind, line 745: These labels need to be set unconditionally, whether to empty values or not, so that intermediate images committed during a final stage that depends on earlier stages in a multi-layer build won't be committed with the values of the labels that were set in that earlier stage.

pkg/cli/build.go
```diff
@@ -223,6 +223,10 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
 		return options, nil, nil, errors.New("'rm' and 'force-rm' can only be set with either 'layers' or 'no-cache'")
 	}
 
+	if iopts.StageLabels && !iopts.SaveStages {
+		return options, nil, nil, errors.New("'stage-labels' requires 'save-stages'")
```
nalind:
```suggestion
		return options, nil, nil, errors.New(`"--stage-labels" requires "--save-stages"`)
```

tests/bud.bats
```diff
@@ -9542,3 +9542,718 @@ _EOF
...
+  _prefetch alpine
+  target="save-stages-$(safename)"
+
+  run_buildah build $WITH_POLICY_JSON --save-stages=false -t ${target}-without-save -f $BUDFILES/save-stages/Dockerfile.single-build-stage $BUDFILES/save-stages
+  run_buildah images -a
+  baseline_images=$(echo "$output" | awk '$1 == "<none>" && $2 == "<none>" {print $3}')
+  baseline_count=$(echo "$baseline_images" | wc -w)
+  assert "$baseline_count" == "0" "without save-stages should have 0 intermediate images"
+
+  run_buildah rmi --all
```
nalind: The teardown() function already clears out the storage that's been set up for the test, the individual tests don't need to do more than that.

tests/bud.bats
```diff
@@ -9542,3 +9542,718 @@ _EOF
...
+  intermediate_count=$(echo "$intermediate_images" | wc -w)
+  assert "$intermediate_count" == "1" "should have 1 intermediate image"
+
+  for image_id in $intermediate_images; do
+    run_buildah inspect --format '{{index .OCIv1.Config.Labels "io.buildah.stage.name"}}' $image_id
+    assert "$output" == "" "stage.name label should not be present when --stage-labels=false"
+
+    run_buildah inspect --format '{{index .OCIv1.Config.Labels "io.buildah.stage.base"}}' $image_id
+    assert "$output" == "" "stage.base label should not be present when --stage-labels=false"
+
```
```suggestion
```

tests/bud.bats
```diff
@@ -9542,3 +9542,718 @@ _EOF
...
+  # Explanation: Empty stage (only FROM AS alias, no RUN/COPY/etc) is just an alias
+  # for the base image. It doesn't need metadata labels or a separate image.
+  # The final stage can reference it via COPY --from=builder, which resolves to alpine:latest.
+  assert "$num_intermediate" "==" "0" "empty intermediate stage (only FROM) should not create intermediate image with labels"
+
+  run_buildah rmi --all
+}
+
+@test "[test final image] --save-stages build has same final image RootFS.DiffIDs count as build without args (all parent layers and signle layer squashing all final stage instructions)" {
```
nalind: Typo: "signle"

tests/bud.bats
```diff
@@ -9542,3 +9542,718 @@ _EOF
...
+  # Second build should reuse the layers created by first build (cache hit)
+  run_buildah build $WITH_POLICY_JSON --layers --save-stages --stage-labels -t ${target}-second -f $BUDFILES/save-stages/Dockerfile.single-build-stage $BUDFILES/save-stages
+
+  expect_output --substring "Using cache"
+
+  # Verify cache was used 4 times (LABEL from --stage-labels, RUN from intermediate stage, COPY from final stage, RUN from final stage)
+  cache_count=$(echo "$output" | grep -c "Using cache")
+  assert "$cache_count" -eq 4 "should use cache 4 times (LABEL, RUN intermediate, COPY, RUN final)"
+
+  # Get all intermediate layers from second build - second build should niot add any more
```
nalind: Typo: "niot"

tests/bud.bats
```diff
@@ -9542,3 +9542,718 @@ _EOF
...
+  # Get all intermediate layers from first build (3 expected: LABEL from --stage-labels, RUN from intermediate stage, COPY from final stage)
+  run_buildah images -a
+  all_intermediate_layers_first=$(echo "$output" | awk '$1 == "<none>" && $2 == "<none>" {print $3}')
+  first_intermediate_layers_count=$(echo "$all_intermediate_layers_first" | wc -w)
+  assert "$first_intermediate_layers_count" -eq 3 "first build should create 3 intermediate layers (LABEL from --stage-labels, RUN from intermediate stage, COPY from final stage)"
+
+  # Second build should reuse the layers created by first build (cache hit)
+  run_buildah build $WITH_POLICY_JSON --layers --save-stages --stage-labels -t ${target}-second -f $BUDFILES/save-stages/Dockerfile.single-build-stage $BUDFILES/save-stages
+
+  expect_output --substring "Using cache"
```
nalind, line 9934: If we're expecting cache hits at every step, I'd expect verifying that the image ID for the second build's final output matches that of the first to be a step in that.

tests/bud.bats
```diff
@@ -9542,3 +9542,718 @@ _EOF
...
+  assert "$cache_count" -eq 2 "should use cache 2 times (COPY, RUN final)"
+
+  run_buildah images -a
+  intermediate_images_second=$(echo "$output" | awk '$1 == "<none>" && $2 == "<none>" {print $3}')
+  count_second=$(echo "$intermediate_images_second" | wc -w)
+  assert "$count_second" -eq 4 "second build should have 4 intermediate layers (2 from first + 2 new with LABEL and RUN)"
+
+  # Get intermediate stage image (RUN layer) and COPY layer from second build
+  second_run_layer=$(echo "$intermediate_images_second" | awk 'NR==1')
+  second_copy_layer=$(echo "$intermediate_images_second" | awk 'NR==3')
```
nalind: The order of values here is not guaranteed.

imagebuildah/executor.go
```diff
@@ -604,6 +610,13 @@ func (b *executor) buildStage(ctx context.Context, cleanupStages map[int]*stageE
 		prependInstructions = slices.Concat([]string{envLine}, prependInstructions)
 	}
 
+	// Create stage labels for intermediate stage images
+	// Skip if stage has no instructions
+	if b.saveStages && b.stageLabels && stageIndex < len(stages)-1 && len(stage.Node.Children) > 0 {
+		labelLine := b.buildStageLabelLine(&stage, base, stages, stageIndex)
```
nalind:
```suggestion
		labelLine := b.buildStageLabelLine(&stage, base, stages[:stageIndex])
```
1 of 3, so that the index of the current stage doesn't need to be passed in.

imagebuildah/executor.go
```diff
@@ -692,6 +713,32 @@ func markDependencyStagesForTarget(dependencyMap map[string]*stageDependencyInfo
 	}
 }
 
+// buildStageLabelLine creates a LABEL instruction line for stage metadata
+func (b *executor) buildStageLabelLine(stage *imagebuilder.Stage, base string, stages imagebuilder.Stages, currentStageIndex int) string {
```
nalind:
```suggestion
func (b *executor) buildStageLabelLine(stage *imagebuilder.Stage, base string, stages imagebuilder.Stages) string {
```
2 of 3, so that the index of the current stage doesn't need to be passed in.

imagebuildah/executor.go
```diff
@@ -692,6 +713,32 @@ func markDependencyStagesForTarget(dependencyMap map[string]*stageDependencyInfo
...
+	labelLine := "LABEL"
+	if stage.Name != "" {
+		labelLine += fmt.Sprintf(" %q=%q", "io.buildah.stage.name", stage.Name)
+	}
+	// Check if base of the stage is another (previous) stage.
+	// If yes, base is set as image ID of this stage.
+	// If not original base name is set (pullspec).
+	for i := 0; i < currentStageIndex; i++ {
+		st := stages[i]
```
nalind:
```suggestion
	for st := range stages {
```
3 of 3, so that the index of the current stage doesn't need to be passed in.

imagebuildah/executor.go
```diff
@@ -115,6 +115,9 @@ type executor struct {
 	layerLabels                             []string
 	annotations                             []string
 	layers                                  bool
+	saveStages                              bool
+	stageLabels                             bool
+	stageImageIDs                           map[string]string // Tracks image IDs for each stage (by name and position) for label references
```
nalind, line 120: This map is read from a helper function called by, and updated directly by, `buildStage()`, which is invoked from multiple goroutines.  Accessing maps in that way is not concurrent-safe without some kind of locking or synchronization.

imagebuildah/executor.go
```diff
@@ -692,6 +713,32 @@ func markDependencyStagesForTarget(dependencyMap map[string]*stageDependencyInfo
...
+	if stage.Name != "" {
+		labelLine += fmt.Sprintf(" %q=%q", "io.buildah.stage.name", stage.Name)
+	}
+	// Check if base of the stage is another (previous) stage.
+	// If yes, base is set as image ID of this stage.
+	// If not original base name is set (pullspec).
+	for i := 0; i < currentStageIndex; i++ {
+		st := stages[i]
+		if st.Name == base || strconv.Itoa(st.Position) == base {
+			if imgID, ok := b.stageImageIDs[base]; ok {
```
nalind, line 734: `buildStage()` hasn't yet created a `stageExecutor` and called its `execute()` method, which is what calls the executor's `waitForStage()` method to wait for a previous base stage to complete.  I suspect that this means that when the number of concurrent stage jobs is > 1, this map might not have a value in it for `base` yet.

nalind (changes requested):
Possible issues when `--jobs` is used with values greater than 1, and questions around labels in the final stage and in stages that inherit from other stages in multi-layer builds.

imagebuildah/executor.go
ezopezo: You are absolutely right, I tested some Containerfiles with stages using another stages as base with --jobs > 1 and bases were resolved to original alias of the stage not to image ID (not desired). I found a function waitForStage, which seems to be doing the trick waiting for resolution of the parent stage - and I also added a test for that.

imagebuildah/executor.go
ezopezo: I added a lock for safe read of the map, thanks for noticing!

imagebuildah/executor.go
ezopezo: Updated, thank you!

tests/bud.bats
ezopezo: Tests updated, thanks!

tests/bud.bats
ezopezo: Removed, thanks!

imagebuildah/executor.go
ezopezo: Removed the conditions!

imagebuildah/stage_executor.go
ezopezo: You're right I removed this. I also explicitly allowed labeling also final stage (final image) in here: https://github.com/ezopezo/buildah/blob/6a3f4d2235737cb2ead7597aabc01263312fec6b/imagebuildah/executor.go#L615
Final image ussually does not have alias so the stage.name will be its index in such cases and base will be pullspec or image id (if another stage is used as its base).

imagebuildah/executor.go
nalind: You'll need to use the same one around writes to the map as well.

imagebuildah/stage_executor.go
ezopezo: No longer relevant.

imagebuildah/executor.go
ezopezo: You're right, thanks for noticing!

tests/bud.bats
```diff
@@ -9484,3 +9484,714 @@ _EOF
...
+  for iid in $all_intermediate_images_jobs1; do
+    if [[ "$child_base_jobs1" == "$iid"* ]]; then
+      found_in_images=true
+      break
+    fi
+  done
+  assert "$found_in_images" == "true" "child stage.base should be imageID from intermediate images"
+
+  # Clean up
+  run_buildah rmi -a -f
```
nalind: I think you'd need to re-prefetch the base image here.  The two halves of this test could be condensed into a loop, though, that runs with different --jobs values.

tests/bud/save-stages/Dockerfile.arg-build-stages-and-chained-build-stages
```diff
@@ -0,0 +1,13 @@
...
+FROM alpine AS ${FIRST_STAGE}
+RUN echo "first-stage content" > /output.txt
+
+FROM ${FIRST_STAGE} AS ${SECOND_STAGE}
+RUN echo "second-stage content" >> /output.txt
+
+FROM alpine
+ARG SECOND_STAGE=second-stage
+COPY --from=${SECOND_STAGE} /output.txt /app/output.txt
+RUN cat /app/output.txt
```
nalind, line 13: Does the test that uses this file check the contents of the file, to ensure that it has the intended content, i.e., that it was copied from the correct stage?  Would the test succeed if line 11 set SECOND_STAGE to the value "first-stage"?

nalind (commented):
Some questions about the tests.

tests/bud/save-stages/Dockerfile.arg-build-stages-and-chained-build-stages
ezopezo: Well that ARG is actually not relevant and even redundant (sorry for confusion I think I added it there because I found out that ARGS stated at the beginning of the stage are not seemingly defined in final stage - weird, but I did not researched why is that - is it a bug?). I just removed it and used COPY --from=second-stage ... instead. 
This is purely a sanity test to verify that ARG variables are correctly resolved to their corresponding values (stage aliases, image ID, and pullspec) and that these resolved values are properly reflected in the stage labels. The file contents and the actual stage instructions are not relevant to what we're testing - they're only present to create intermediate stages with some layer content and this content is COPY-ed to final imaeg just to reference those stages to be actually built. I understand that ARG resolution happens at parse time, before stage label generation, so this test might seem somewhat redundant, but it does verify the end-to-end flow. In the meantime, I removed the pointless ARG SECOND_STAGE=second-stage line in the final stage (and added ARG PULLSPEC=alpine to test pullspec resolution as well), but feel free to decide if this test is even useful.

tests/bud.bats
ezopezo: You are right, resolved, thanks!

tests/bud/save-stages/Dockerfile.arg-build-stages-and-chained-build-stages
nalind: I don't think I understand the question you're asking about ARGs in stages in that first paragraph...?

ezopezo: Sorry I wrote it wrongly, I meant; ARGS stated at the top of the _Containerfile_ are not resolved _inside of any stage_ (only in FROM instructions as pullspecs or aliases). But it makes a sense I guess, that stages have their own ARG scope. Nevermind it is not important, I don't want to tangle into this I just wanted to explain why I did this weird stuff before:
```
ARG SECOND_STAGE=second-stage
COPY --from=${SECOND_STAGE} ...
```
But it is not needed at all as I explained

Anyway, is the Containerfile and test okay as it is now?

nalind: ARGs defined before the first FROM instruction (i.e., in the header) are available in stages once they're declared, per https://docs.docker.com/build/building/variables/#scoping.  The part where a stage inherits the ARGs defined in a stage that it's based on, however, was only recently fixed.
The test is fine, it's just confusing to see it appear to set the file's contents in a way that a test would if it was going to check on them, and then not do that.

nalind (approved):
LGTM

docs/buildah-build.1.md
TomSweeneyRedHat: And Ooops, yep, I accidentally left off the `=` in my comment, re-adding.

Regardless, a suggestion, not a required change.

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
@ezopezo thanks for the PR and the continued attention to it.  I'm going to merge this shortly.  If you decide to tweak the docs as suggested later in a separate PR, great!  If not, no worries.
/lgtm

PR #6538, opened 2025-11-25
[release-1.29] Bump runc up to 1.2.9 for CVE-2025-52881, CVE-2025-31133 and CVE-2025-52565

cevich:
#### What type of PR is this?

/kind other

#### What this PR does / why we need it:

Backport PR #6484 & #6511

#### How to verify it

CI + Manual

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

The commits in this PR were created **with the assistance of AI**, based on the source https://github.com/containers/buildah/pull/6520.  When reviewing please pay special attention to the following:

1. **Vendor directory consistency**:
   - Vendor directory was completely regenerated using `make vendor-in-container` after each `go.mod` change
   - Never manually edited, ensuring consistency with `go.mod` and `go.sum`
   - **Backport change**: Same process as source branch - vendor directory was regenerated after dependency updates to ensure consistency.

1. **All compilation verified**:
   - Project compiles successfully with `make` after every commit
   - All compilation errors encountered during backport were resolved and ammended to the commit.

1. **"Disable lint checking"**
   - This check fails on the branch even w/o any changes.
   - In CI the problem is difficult to diagnose as no output is provided,
     rather the process is simply killed.
   - **Backport change**: This commit was manually created, it does not exist on the
     source PRs.

1. **"Bump runc to v1.2.8 - CVE-2025-52881"** and subsequent commits include extensive `.Get()` method calls:
   - `run_common.go`: Added `.Get()` calls for `defaultConfig.Containers.DNSServers.Get()`, `DNSSearches.Get()`, and `DNSOptions.Get()`
   - `run_linux.go`: Added `.Get()` call for `defaultContainerConfig.Containers.DefaultUlimits.Get()`
   - `pkg/cli/common.go`: Added `.Get()` calls for `defaultContainerConfig.Containers.Volumes`, `Devices`, `DNSSearches`, `DNSServers`, and `DNSOptions`
   - `imagebuildah/executor.go`: Added `.Get()` calls for `defaultContainerConfig.Containers.Devices` and `Volumes`
   - `cmd/buildah/from.go`: Added `.Get()` calls for `defaultContainerConfig.Containers.DefaultUlimits` and `Devices`
   - `cmd/buildah/main.go`: Added `.Get()` call for `defaultContainerConfig.Engine.Env`
   - **Reason**: The `containers/common` package changed `attributedstring.Slice` types to require explicit `.Get()` method calls to convert to `[]string` before use.

[...]

PR #6558, opened 2025-12-01
[do not merge] add swap to CI for 1.29

nalind:
Ignore me!


nalind:
Flaking _a lot_ in a way that's reminiscent of when we needed the change from https://github.com/containers/common/pull/2339.

nalind:
Closing in favor of #6538.

PR #6592, opened 2025-12-10
Enable building Windows container images

### LINKED ISSUE #4010: Building Windows container images

sebsoto:
#### What type of PR is this?

/kind feature

    Add basic functionality to build Windows images
    
    Enables committing a Windows base image as the final layer.
    This gives the possibility of building a Windows container image on a
    Linux build system. Only COPY and ADD are in scope. This allows the
    workflow of cross compiling Windows binaries on Linux, and copying them
    into a Windows container image for use on Windows systems.

Fixes #4010

```release-note

```

### DIFF
--- a/image.go
+++ b/image.go
@@ -14,6 +14,7 @@ import (
 	"path/filepath"
 	"slices"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/containers/buildah/copier"
@@ -59,6 +60,37 @@ const (
 	// create uniquely-named files, but we don't want to try to use their
 	// contents until after they've been written to
 	containerExcludesSubstring = ".tmp"
+
+	// Windows-specific PAX record keys
+	keyFileAttr     = "MSWINDOWS.fileattr"
+	keySDRaw        = "MSWINDOWS.rawsd"
+	keyCreationTime = "LIBARCHIVE.creationtime"
+	// FILE_ATTRIBUTE_ values
+	fileAttributeDirectory = "16"
+	fileAttributeArchive   = "32"
+
+	// Windows Security Descriptors (base64-encoded)
+	// SDDL: O:BAG:SYD:(A;OICI;FA;;;BA)(A;OICI;FA;;;SY)(A;;FA;;;BA)(A;OICIIO;GA;;;CO)(A;OICI;0x1200a9;;;BU)(A;CI;LC;;;BU)(A;CI;DC;;;BU)
+	// Owner: Built-in Administrators (BA)
+	// Group: Local System (SY)
+	// DACL:
+	//   - Allow OBJECT_INHERIT+CONTAINER_INHERIT Full Access to Built-in Administrators (BA)
+	//   - Allow OBJECT_INHERIT+CONTAINER_INHERIT Full Access to Local System (SY)
+	//   - Allow Full Access to Built-in Administrators (BA)
+	//   - Allow OBJECT_INHERIT+CONTAINER_INHERIT+INHERIT_ONLY Generic All to Creator Owner (CO)
+	//   - Allow OBJECT_INHERIT+CONTAINER_INHERIT Read/Execute permissions to Built-in Users (BU)
+	//   - Allow CONTAINER_INHERIT List Contents to Built-in Users (BU)
+	//   - Allow CONTAINER_INHERIT Delete Child to Built-in Users (BU)
+	winSecurityDescriptorDirectory = "AQAEgBQAAAAkAAAAAAAAADAAAAABAgAAAAAABSAAAAAgAgAAAQEAAAAAAAUSAAAAAgCoAAcAAAAAAxgA/wEfAAECAAAAAAAFIAAAACACAAAAAxQA/wEfAAEBAAAAAAAFEgAAAAAAGAD/AR8AAQIAAAAAAAUgAAAAIAIAAAALFAAAAAAQAQEAAAAAAAMAAAAAAAMYAKkAEgABAgAAAAAABSAAAAAhAgAAAAIYAAQAAAABAgAAAAAABSAAAAAhAgAAAAIYAAIAAAABAgAAAAAABSAAAAAhAgAA"
+
+	// SDDL: O:BAG:SYD:(A;;FA;;;BA)(A;;FA;;;SY)(A;;0x1200a9;;;BU)
+	// Owner: Built-in Administrators (BA)
+	// Group: Local System (SY)
+	// DACL:
+	//   - Allow Full Access to Built-in Administrators (BA)
+	//   - Allow Full Access to Local System (SY)
+	//   - Allow Read/Execute permissions to Built-in Users (BU)
+	winSecurityDescriptorFile = "AQAEgBQAAAAkAAAAAAAAADAAAAABAgAAAAAABSAAAAAgAgAAAQEAAAAAAAUSAAAAAgBMAAMAAAAAABgA/wEfAAECAAAAAAAFIAAAACACAAAAABQA/wEfAAEBAAAAAAAFEgAAAAAAGACpABIAAQIAAAAAAAUgAAAAIQIAAA=="
 )
 
 // ExtractRootfsOptions is consumed by ExtractRootfs() which allows users to
@@ -112,6 +144,7 @@ type containerImageRef struct {
 	unsetAnnotations      []string
 	setAnnotations        []string
 	createdAnnotation     types.OptionalBool
+	os                    string
 }
 
 type blobLayerInfo struct {
@@ -1084,6 +1117,7 @@ func (i *containerImageRef) NewImageSource(ctx context.Context, _ *types.SystemC
 		diffBeingAltered := i.compression != archive.Uncompressed
 		diffBeingAltered = diffBeingAltered || i.layerModTime != nil || i.layerLatestModTime != nil
 		diffBeingAltered = diffBeingAltered || len(layerExclusions) != 0
+		diffBeingAltered = diffBeingAltered || i.os == "windows"
 		if diffBeingAltered {
 			destHasher = digest.Canonical.Digester()
 			multiWriter = io.MultiWriter(counter, destHasher.Hash())
@@ -1099,11 +1133,13 @@ func (i *containerImageRef) NewImageSource(ctx context.Context, _ *types.SystemC
 			return nil, fmt.Errorf("compressing %s: %w", what, err)
 		}
 		writer := io.MultiWriter(writeCloser, srcHasher.Hash())
-
 		// Use specified timestamps in the layer, if we're doing that for history
 		// entries.
 		nestedWriteCloser := ioutils.NewWriteCloserWrapper(writer, writeCloser.Close)
-		writeCloser = makeFilteredLayerWriteCloser(nestedWriteCloser, i.layerModTime, i.layerLatestModTime, layerExclusions)
+		writeCloser, err = makeFilteredLayerWriteCloser(nestedWriteCloser, i.layerModTime, i.layerLatestModTime, layerExclusions, i.os == "windows")
+		if err != nil {
+			return nil, fmt.Errorf("creating filter write closer %s: %w", what, err)
+		}
 		writer = writeCloser
 		// Okay, copy from the raw diff through the filter, compressor, and counter and
 		// digesters.
@@ -1373,9 +1409,9 @@ func (i *containerImageRef) makeExtraImageContentDiff(includeFooter bool, timest
 // no later than layerLatestModTime (if a value is provided for it).
 // This implies that if both values are provided, the archive's timestamps will
 // be set to the earlier of the two values.
-func makeFilteredLayerWriteCloser(wc io.WriteCloser, layerModTime, layerLatestModTime *time.Time, exclusions []copier.ConditionalRemovePath) io.WriteCloser {
-	if layerModTime == nil && layerLatestModTime == nil && len(exclusions) == 0 {
-		return wc
+func makeFilteredLayerWriteCloser(wc io.WriteCloser, layerModTime, layerLatestModTime *time.Time, exclusions []copier.ConditionalRemovePath, windows bool) (io.WriteCloser, error) {
+	if layerModTime == nil && layerLatestModTime == nil && len(exclusions) == 0 && !windows {
+		return wc, nil
 	}
 	exclusionsMap := make(map[string]copier.ConditionalRemovePath)
 	for _, exclusionSpec := range exclusions {
@@ -1385,19 +1421,22 @@ func makeFilteredLayerWriteCloser(wc io.WriteCloser, layerModTime, layerLatestMo
 		}
 		exclusionsMap[pathSpec] = exclusionSpec
 	}
+	var initialized bool
 	wc = newTarFilterer(wc, func(hdr *tar.Header) (skip, replaceContents bool, replacementContents io.Reader) {
-		// Changing a zeroed field to a non-zero field can affect the
-		// format that the library uses for writing the header, so only
-		// change fields that are already set to avoid changing the
-		// format (and as a result, changing the length) of the header
-		// that we write.
 		modTime := hdr.ModTime
-		nameSpec := strings.Trim(path.Clean(hdr.Name), "/")
-		if conditions, ok := exclusionsMap[nameSpec]; ok {
-			if (conditions.ModTime == nil || conditions.ModTime.Equal(modTime)) &&
-				(conditions.Owner == nil || (conditions.Owner.UID == hdr.Uid && conditions.Owner.GID == hdr.Gid)) &&
-				(conditions.Mode == nil || (*conditions.Mode&os.ModePerm == os.FileMode(hdr.Mode)&os.ModePerm)) {
-				return true, false, nil
+		if layerModTime != nil || layerLatestModTime != nil || len(exclusions) != 0 {
+			// Changing a zeroed field to a non-zero field can affect the
+			// format that the library uses for writing the header, so only
+			// change fields that are already set to avoid changing the
+			// format (and as a result, changing the length) of the header
+			// that we write.
+			nameSpec := strings.Trim(path.Clean(hdr.Name), "/")
+			if conditions, ok := exclusionsMap[nameSpec]; ok {
+				if (conditions.ModTime == nil || conditions.ModTime.Equal(modTime)) &&
+					(conditions.Owner == nil || (conditions.Owner.UID == hdr.Uid && conditions.Owner.GID == hdr.Gid)) &&
+					(conditions.Mode == nil || (*conditions.Mode&os.ModePerm == os.FileMode(hdr.Mode)&os.ModePerm)) {
+					return true, false, nil
+				}
 			}
 		}
 		if layerModTime != nil {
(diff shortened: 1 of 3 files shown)


tests/bud.bats
```diff
@@ -9165,3 +9165,87 @@ EOF
...
+  # Find the layer blob (it's in blobs/sha256/)
+  # Get the manifest digest from index.json using jq
+  run jq -r '.manifests[0].digest' ${ocidir}/index.json
+  local manifest_digest=${output#sha256:}
+  local manifest_file=${ocidir}/blobs/sha256/${manifest_digest}
+
+  # Get the last layer digest from the manifest (the layer we just created)
+  run jq -r '.layers[-1].digest' ${manifest_file}
+  local layer_digest=${output#sha256:}
+  local layer_file=${ocidir}/blobs/sha256/${layer_digest}
```
nalind: There's a `oci_image_last_diff()` helper function which will probably save a bit of time here.

image.go
```diff
@@ -1181,6 +1169,217 @@ func (i *containerImageRef) NewImageSource(ctx context.Context, _ *types.SystemC
...
+}
+
+func (w *winMutator) Close() error {
+	if err := w.pw.Close(); err != nil {
+		return fmt.Errorf("error closing pipewriter")
+	}
+	return <-w.done
+}
+
+func newWindowsMutator(outStream io.WriteCloser) *winMutator {
```
nalind: This duplicates a lot of what `makeFilteredLayerWriteCloser()` is already doing to create a tar filterer through which the layer diff will be passed.  Adding these changes to it will save us the trouble of passing the layer over another pipe.

image.go
```diff
@@ -1181,6 +1169,217 @@ func (i *containerImageRef) NewImageSource(ctx context.Context, _ *types.SystemC
...
+		h.PAXRecords[keyFileAttr] = "16"
+	}
+
+	if h.Typeflag == tar.TypeReg {
+		h.Mode |= 1 << 15
+		h.PAXRecords[keyFileAttr] = "32"
+	}
+
+	if !h.ModTime.IsZero() {
+		h.PAXRecords[keyCreationTime] = fmt.Sprintf("%d.%d", h.ModTime.Unix(), h.ModTime.Nanosecond())
```
nalind: The timestamp needs to be clamped as we do elsewhere when the build is meant to be reproducible.  The nanosecond value is not zero-padded.

nalind (commented):
I usually don't look at draft and WIPs, but since you asked.

sebsoto:
Thanks @nalind 
I addressed your comments and took the PR out of WIP.

image.go
```diff
@@ -1386,34 +1543,44 @@ func makeFilteredLayerWriteCloser(wc io.WriteCloser, layerModTime, layerLatestMo
...
-		}
-		if !hdr.AccessTime.IsZero() {
-			hdr.AccessTime = modTime
-		}
-		if !hdr.ChangeTime.IsZero() {
-			hdr.ChangeTime = modTime
+		if windows {
+			hdr.Name = filepath.Join("Files", hdr.Name)
+			if hdr.Linkname != "" {
+				hdr.Linkname = filepath.Join("Files", hdr.Linkname)
```
nalind: These need to be `path.Join()`, in case this code is ever natively built for Windows.

nalind (changes requested):
I think that dropping the `layerWriter` refactor would make this PR considerably smaller, and would hopefully avoid the copy-pasting of functions from elsewhere.

Having the tar filter adjust everything that passes through it (i.e., setting the PAX records and header format) , and having `makeFilteredLayerWriteCloser()` write the initial `Hives/` and `Files/` entries through the new `WriteCloser` before returning it (much as newLayerWriter() would have) would reduce the number of locations where the header changes needed to be made and the desired `ModTime` determined.

sebsoto:
@nalind 

>having makeFilteredLayerWriteCloser() write the initial Hives/ and Files/ entries through the new WriteCloser before returning it (much as newLayerWriter() would have) would reduce the number of locations where the header changes needed to be made and the desired ModTime determined.

Finding a clean way to do this has been difficult, as the directories passed through the writer would end up being `Files/Files` and `Files/Hives`.  So while not doing this exactly, I did reduce the scope as requested, and removed duplicate code where possible.

image.go
```diff
@@ -1386,40 +1476,53 @@ func makeFilteredLayerWriteCloser(wc io.WriteCloser, layerModTime, layerLatestMo
 		exclusionsMap[pathSpec] = exclusionSpec
 	}
```
nalind, line 1423: Define a boolean here...

image.go
```diff
@@ -1386,40 +1476,53 @@ func makeFilteredLayerWriteCloser(wc io.WriteCloser, layerModTime, layerLatestMo
...
-			hdr.AccessTime = modTime
-		}
-		if !hdr.ChangeTime.IsZero() {
-			hdr.ChangeTime = modTime
+		clampModTime(hdr, layerModTime, layerLatestModTime)
+		if windows {
+			hdr.Name = path.Join("Files", hdr.Name)
+			if hdr.Linkname != "" {
+				hdr.Linkname = path.Join("Files", hdr.Linkname)
+			}
```
nalind, line 1470: ... make the manipulation of the `Name` and `Linkname` fields conditional on it...

image.go
```diff
@@ -1386,40 +1476,53 @@ func makeFilteredLayerWriteCloser(wc io.WriteCloser, layerModTime, layerLatestMo
...
+			prepareWinHeader(hdr)
 		}
 		return false, false, nil
 	})
```
nalind, line 1490: ... and write "Hives/" and "Files/" (including the final slash characters, as is convention) here before toggling the boolean that the renaming is conditioned on.  At that point, this becomes the only caller to prepareWinHeader(), and it can be inlined...

image.go
```diff
@@ -1386,40 +1476,53 @@ func makeFilteredLayerWriteCloser(wc io.WriteCloser, layerModTime, layerLatestMo
...
+		hdr.ModTime = modTime
+	}
+	if !hdr.AccessTime.IsZero() {
+		hdr.AccessTime = modTime
+	}
+	if !hdr.ChangeTime.IsZero() {
+		hdr.ChangeTime = modTime
+	}
+}
+
```
nalind, line 1522: ... and then this can be inlined back into its caller.

image.go
```diff
@@ -59,6 +59,34 @@ const (
 	// create uniquely-named files, but we don't want to try to use their
 	// contents until after they've been written to
 	containerExcludesSubstring = ".tmp"
+
+	// Windows-specific PAX record keys
+	keyFileAttr     = "MSWINDOWS.fileattr"
```
nalind, line 65: Is this expected to hold `FILE_ATTRIBUTE_...` values that would be defined in `golang.org/x/sys/windows` when running on that platform?

nalind (changes requested):
I think we can shrink this further.

image.go
```diff
@@ -1181,6 +1217,60 @@ func (i *containerImageRef) NewImageSource(ctx context.Context, _ *types.SystemC
...
 }
 
+// prepareWinHeader ensures all header fields are set in a way that is expected for Windows images
+func prepareWinHeader(h *tar.Header) {
+	if h.PAXRecords == nil {
+		h.PAXRecords = map[string]string{}
+	}
+	h.Format = tar.FormatPAX
+	if h.Typeflag == tar.TypeDir {
+		h.Mode |= 1 << 14
```
nalind: Is this meant to be `syscall.S_IFDIR`?

image.go
```diff
@@ -1181,6 +1217,60 @@ func (i *containerImageRef) NewImageSource(ctx context.Context, _ *types.SystemC
...
+		h.PAXRecords[keyFileAttr] = "16"
+		h.PAXRecords[keySDRaw] = winSecurityDescriptorDirectory
+	}
+	if h.Typeflag == tar.TypeReg {
+		h.Mode |= 1 << 15
```
nalind: Is this meant to be `syscall.S_IFREG`?

sebsoto:
@nalind your comments have been addressed.
The failing tests seem to be an unrelated flake:
```
[+0621s] # E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/libc/libcap2/libcap2-bin_2.32-1ubuntu0.2_amd64.deb  Connection failed [IP: 91.189.92.24 80]
[+0621s] # E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
[+0621s] # Error: while running runtime: exit status 100
```

image.go
sebsoto: Yes, added a comment for clarity

image.go
sebsoto: Replaced with constant

image.go
sebsoto: Replaced with constant

nalind (changes requested):
Still need to include the final '/' in the "Files" and "Hives" entries to match the convention for other directory entries in layers.
I'd still like to see `prepareWinHeader()` inlined.
Just noticed that the `path.Join()` calls conditioned on `initialized` drops the final '/' from `hdr.Name` values that would have had them, so they need to be added back for directory entries, and that the "Files" and "Hives" entries need to have non-zero values set in their `Mode` fields..

image.go
```diff
@@ -1415,9 +1456,70 @@ func makeFilteredLayerWriteCloser(wc io.WriteCloser, layerModTime, layerLatestMo
...
+	if windows {
+		// prep the archive by writing the Files/ and Hives/ directories to the writer.
+		// This fulfills a requirement of Windows images
+		winTarWriter := tar.NewWriter(wc)
+		now := time.Now()
+		h := &tar.Header{
+			Name:     "Hives/",
+			Typeflag: tar.TypeDir,
+			ModTime:  now,
+			Mode:     syscall.S_IFDIR,
```
nalind: If this is extracted on a Linux system, the directory will end up with 000 permissions, which doesn't look like what it's set to in the base layer provided by `mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image:v1.0.0`.  It should probably be `0o775`, as it is in the base image, or `0o755`, which is often used for top-level directories in Linux images.

image.go
```diff
@@ -1415,9 +1456,70 @@ func makeFilteredLayerWriteCloser(wc io.WriteCloser, layerModTime, layerLatestMo
...
+		}
+		if err := winTarWriter.WriteHeader(h); err != nil {
+			return nil, err
+		}
+		h = &tar.Header{
+			Name:     "Files/",
+			Typeflag: tar.TypeDir,
+			ModTime:  now,
+			Mode:     syscall.S_IFDIR,
```
nalind: If this is extracted on a Linux system, the directory will end up with 000 permissions, which doesn't look like what it's set to in the base layer provided by `mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image:v1.0.0`.  It should probably be `0o775`, as it is in the base image, or `0o755`, which is often used for top-level directories in Linux images.

nalind (changes requested):

nalind (changes requested):
One last nit.

image.go
```diff
@@ -1366,16 +1402,18 @@ func (i *containerImageRef) makeExtraImageContentDiff(includeFooter bool, timest
 	return diff.Name(), digester.Digest(), counter.Count, nil
 }
 
+// prepareWinHeader ensures all header fields are set in a way that is expected for Windows images
+
```
nalind: Leftover comment?

nalind (approved):
LGTM

sebsoto:
Thanks @nalind !
Can you help get this merged?

nalind:

giuseppe (approved):
LGTM

sebsoto:
@containers/buildah-maintainers

Updated to resolve merge conflict

PR #6593, opened 2025-12-10
Fix commit validation workflow

cevich:
#### What type of PR is this?

/kind other

Fix commit title-length checking workflow bug.  Also, skip running this check for PR's **NOT** targeting the `main` branch.  These are typically renovate and/or release-branches that often have extended commit titles w/ additional context details.

#### How to verify it

The workflow will fail on this PR when the very long 'DO NOT MERGE:...' commit is included.  But the workflow will pass when that commit is removed.

None

```release-note
None
```

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -16,10 +16,25 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
 
+      # Workaround limitation/bug in commit-message-checker-with-regex (action below) where
+      # it breaks if there are "too many" commits in the PR.
+      # Ref: https://github.com/tim-actions/commit-message-checker-with-regex/issues/7
+
+      - name: Reduce commit data
+        run: >
+          echo '${{ steps.get-pr-commits.outputs.commits }}' | \
+              jq --monochrome-output --compact-output '[.[] | {"sha": .sha, "commit":{ "message": .commit.message }}]' \
+              > filtered-commits.json
+
+      - name: Load filtered commit data into output
+        id: 'get-pr-commit-messages'
+        run: >
+          echo "commit-messages=$(cat filtered-commits.json)" >> $GITHUB_OUTPUT
+
       - if: contains(github.head_ref, 'renovate/') != true
         name: check subject line length
         uses: tim-actions/commit-message-checker-with-regex@094fc16ff83d04e2ec73edb5eaf6aa267db33791 # v0.3.2
         with:
-          commits: ${{ steps.get-pr-commits.outputs.commits }}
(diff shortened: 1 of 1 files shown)


cevich:
Ref: [Workflow log](https://github.com/containers/buildah/actions/runs/20103835599/job/57681881329?pr=6593) with the too-long "DO NOT MERGE: ..." commit [26a4c1c](https://github.com/containers/buildah/pull/6593/commits/26a4c1c29bed727857fe9eca1c86fbccc6879afd).

cevich:
Ref: [Workflow log](https://github.com/containers/buildah/actions/runs/20103901069/job/57682115723?pr=6593) without the too-long "DO NOT MERGE: ..."

TomSweeneyRedHat:
@cevich I like the idea, but the tests don't as of yet.

cevich:
Ugh, must be flakes.  I didn't touch any code other than the workflow! :disappointed: I'll rebase and re-run shortly.

cevich:
There we go, the Cirrus tests are all green.  I dunno about the testing farm stuff and/or if that's required or not.

nalind:
This looks like it'll accomplish the goal, but once we limit the validation check to `main`, I'm not sure how much we want to also be raising the limit on the number of commits in a PR that the check can handle.

cevich:
> I'm not sure how much we want to also be raising the limit on the number of commits in a PR that the check can handle.

I'm not sure what the actual limit is, but by my understanding, there's also no way to increase that limit.  The solution would have to be "use a different tool".

For now, IMHO it's probably fine to have this but keep is as "optional" in case it fails again.

Honny1:
LGTM. I’m not sure about limiting the number of commits in a PR. What about backport PRs? There should be a label to disable this limit.

cevich:
> There should be a label to disable this limit.

My understanding is the limit is in the tooling (as in, it simply can't handle "too big" JSON).  But I do share your concern about the backporting case, so as a matter of practice I think this check should continue to be considered advisory (I'm assuming that's still the policy).

Alternatively, perhaps the job could be made conditional on the absence of a PR label?

```
...
on:
  pull_request:
    branches:
      - main
    # Also trigger workflow anytime PR labels are changed
    types: types: [labeled, opened, synchronize, reopened]
...
jobs:
  commit:
    runs-on: ubuntu-24.04
    # Check may need to be skipped on large backports w/ "too many" commits.
    if: "!contains(github.event.pull_request.labels.*.name, 'skip-validate')"
...
```

Though I'd need guidance on if adding a new label for this would be acceptable, and what it's name should be (if not `skip-validate`).

PR #6634, opened 2026-01-13
Use cached images instead of fedoraproject.org

### LINKED ISSUE #6492: tests: do not use registry.fedoraproject.org
We should avoid depending on external resources as much as possible for reliable CI testing.
There are two problems with that.
First, the image on the remote could change or be removed at any time because we do not own it which would break the tests. IMO we should only ever use images under the quay.io/libpod namespace that we control.
Second, pulling from the internet is always prone to many flakes, the image should be cached in a local registry to avoid failures during registry outages. We have a working cache registry setup in cirrus but we do not cache the registry.fedoraproject.org/fedora-minimal:latest image there.

If that specific image is needed we should mirror it once into our quay.io namespace and then cache it from there in the automation setup. However I think it would be easier to use one of the existing images that we already have cached, see https://github.com/containers/automation_images/blob/f972e81d74d56699839fd6d5b3bfd0cf1edb80ef/cache_images/local-cache-registry#L56-L97

IrvingMg:
#### What type of PR is this?

/kind cleanup

Image replacements:                                                                                                                                                                                             
- `fedora-minimal` -> `testimage:20241011` (label tests in bud.bats, config.bats)                                                                                                                                
- `fedora-minimal` -> `alpine:3.10.2` (blobcache.bats)                                                                                                                                                           
- `fedora-minimal` -> `ubuntu:latest` (conformance test)                                                                                                                                                         
- Removed `fedora-minimal` registrypair from registries.bats (short-name resolution is still tested via busybox/docker.io pairs; adding a quay.io pair would require custom registries.conf) 

#### How to verify it
 Run the affected tests:`sudo bats tests/config.bats tests/bud.bats tests/registries.bats`

Fixes #6492 

```release-note
None
```

### DIFF
--- a/tests/blobcache.bats
+++ b/tests/blobcache.bats
@@ -71,7 +71,7 @@ function _check_matches() {
 	run_buildah login --tls-verify=false --authfile ${TEST_SCRATCH_DIR}/test.auth --username testuser --password testpassword localhost:${REGISTRY_PORT}
 	outputdir=${TEST_SCRATCH_DIR}/outputdir
 	mkdir -p ${outputdir}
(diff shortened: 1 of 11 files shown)


nalind:
LGTM, thanks!

TomSweeneyRedHat:
Thanks @IrvingMg !
LGTM

PR #6647, opened 2026-01-21
Add --source-policy-file flag for BuildKit-compatible source policies

tinovyatkin:
## Summary
- Adds a new `--source-policy-file` flag to `buildah build` that enables users to control and transform source references during builds without modifying Dockerfiles
- Implements BuildKit-compatible source policy JSON schema with ALLOW, DENY, and CONVERT actions
- Supports EXACT and WILDCARD match types for flexible policy rules

## Background
This implements support for the BuildKit source policy feature as documented in [BuildKit's build reproducibility guide](https://github.com/moby/buildkit/blob/master/docs/build-repro.md). Docker Buildx supports this via the [`EXPERIMENTAL_BUILDKIT_SOURCE_POLICY` environment variable](https://docs.docker.com/build/building/variables/#experimental_buildkit_source_policy). The JSON schema is compatible with BuildKit's source policy format.

## Motivation
This feature allows organizations to:
- Pin base image tags to specific digests at build time for reproducibility
- Deny specific sources from being used to enforce security policies
- Transform source references transparently without modifying Containerfiles

## Changes
- **pkg/sourcepolicy/**: New package implementing policy parsing, validation, and matching
- **define/build.go**: Added SourcePolicyFile field to BuildOptions
- **pkg/cli/common.go**: Added --source-policy-file flag definition
- **imagebuildah/executor.go**: Policy loading in newExecutor()
- **imagebuildah/stage_executor.go**: Policy evaluation in prepare() function
- **docs/buildah-build.1.md**: Man page documentation with examples
- **tests/source-policy.bats**: Integration tests
- **pkg/sourcepolicy/policy_test.go**: Unit tests

## Example Usage

Pin alpine:latest to a specific digest:

```json
{
  "rules": [
    {
      "action": "CONVERT",
      "selector": {
        "identifier": "docker-image://docker.io/library/alpine:latest"
      },
      "updates": {
        "identifier": "docker-image://docker.io/library/alpine@sha256:abc123"
      }
    }
  ]
}
```

```bash
buildah build --source-policy-file policy.json -t myimage .
```

## Test plan
- [x] Unit tests for policy parsing, validation, and matching (pkg/sourcepolicy/policy_test.go)
- [x] Integration tests covering DENY, ALLOW, CONVERT actions and WILDCARD matching (tests/source-policy.bats)
- [ ] Manual testing with real images

---
🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -1045,6 +1045,62 @@ will bear exactly the specified timestamp.
 Conflicts with the similar **--timestamp** flag, which also sets its specified
 time on the contents of new layers.
 
+**--source-policy-file** *pathname*
+
+Specifies the path to a BuildKit-compatible source policy JSON file.  When
+specified, source references (e.g., base images in FROM instructions) are
+evaluated against the policy rules before being used.
+
+Source policies allow controlling which images can be used as base images and
+optionally converting image references (e.g., pinning tags to specific digests)
+without modifying Containerfiles.  This is useful for enforcing organizational
+policies and ensuring build reproducibility.
+
+The policy file is a JSON document containing an array of rules.  Each rule has:
+- **action**: The action to take when the rule matches.  Valid actions are:
+  - **ALLOW**: Explicitly allow the source (no transformation).
+  - **DENY**: Block the source and fail the build.
+  - **CONVERT**: Transform the source to a different reference specified in `updates`.
+- **selector**: Specifies which sources the rule applies to.
+  - **identifier**: The source identifier to match (e.g., `docker-image://docker.io/library/alpine:latest`).
+  - **matchType**: How to match the identifier.  Valid types are `EXACT` (default) and `WILDCARD` (supports `*` and `?` glob patterns).
+- **updates**: For `CONVERT` actions, specifies the replacement identifier.
+
+Rules are evaluated in order; the first matching rule wins.  If no rule matches,
+the source is allowed by default.
+
+Example policy file that pins alpine:latest to a specific digest:
+```json
+{
+  "rules": [
+    {
+      "action": "CONVERT",
+      "selector": {
+        "identifier": "docker-image://docker.io/library/alpine:latest"
+      },
+      "updates": {
+        "identifier": "docker-image://docker.io/library/alpine@sha256:..."
+      }
+    }
+  ]
+}
+```
+
+Example policy file that denies all ubuntu images:
+```json
+{
+  "rules": [
+    {
+      "action": "DENY",
+      "selector": {
+        "identifier": "docker-image://docker.io/library/ubuntu:*",
+        "matchType": "WILDCARD"
+      }
+    }
+  ]
+}
+```
+
 **--squash**
 
 Squash all layers, including those from base image(s), into one single layer. (Default is false).
@@ -1426,6 +1482,10 @@ buildah build --secret=id=mysecret,src=MYSECRET,type=env .
 buildah build --secret=id=mysecret,src=.mysecret,type=file .
 
 buildah build --secret=id=mysecret,src=.mysecret .
+
+### Building an image with a source policy
+
+buildah build --source-policy-file /etc/buildah/source-policy.json -t imageName .
 
 ### Building an multi-architecture image using the --manifest option (requires emulation software)
 
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -1063,11 +1063,16 @@ The policy file is a JSON document containing an array of rules.  Each rule has:
   - **CONVERT**: Transform the source to a different reference specified in `updates`.
 - **selector**: Specifies which sources the rule applies to.
   - **identifier**: The source identifier to match (e.g., `docker-image://docker.io/library/alpine:latest`).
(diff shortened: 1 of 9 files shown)


tinovyatkin:
The two remaining CI failures appear to be unrelated to this PR:

1. **Integration fedora-43 using crun w/ vfs** - Fails on `mkcw-convert` test (confidential workload test, unrelated to source policy)
2. **Integration rootless fedora-43 using crun w/ overlay** - Timed out after 30 minutes (infrastructure issue)

All 14 source-policy integration tests pass on every platform. The same fedora-43 tests pass with other configurations (runc, overlay).

Could a maintainer please re-run these flaky checks? Thank you!

tinovyatkin:
CI update after squash: The two remaining failures are both `mkcw` (confidential workload) tests failing with cryptsetup exit code 5 ("Device still in use"):

- **Integration fedora-42 using runc w/ overlay** - `mkcw build` test
- **Integration fedora-43 using crun w/ overlay** - `mkcw-convert` test

This is a known race condition where the kernel device mapper doesn't release devices fast enough after unmount. These tests have retry logic but still fail intermittently.

All 14 source-policy integration tests pass on every platform. The mkcw failures are infrastructure-related and unrelated to this PR.

pkg/sourcepolicy/policy.go
```diff
@@ -0,0 +1,315 @@
+// Package sourcepolicy implements BuildKit-compatible source policy evaluation
+// for controlling and transforming source references during builds.
+//
+// Source policies allow users to:
+//   - Pin base image tags to specific digests at build time
+//   - Deny specific sources from being used
+//   - Transform source references without modifying Dockerfiles
```
TomSweeneyRedHat: nit
```suggestion
//   - Transform source references without modifying Containerfiles or Dockerfiles
```

docs/buildah-build.1.md
```diff
@@ -1045,6 +1045,62 @@ will bear exactly the specified timestamp.
...
+      "action": "DENY",
+      "selector": {
+        "identifier": "docker-image://docker.io/library/ubuntu:*",
+        "matchType": "WILDCARD"
+      }
+    }
+  ]
+}
+```
+
```
TomSweeneyRedHat, line 1108: It looks like this works in conjunction with the policy.json file.  I.e., when you use this new option, the policy in policy.json and the one in the specified file are both used.  It would be good to state that somewhere in here, or if I'm wrong, state that it overrides the policy.json file.

TomSweeneyRedHat:
Couple small nits, otherwise LGTM.  @nalind?

pkg/sourcepolicy/policy.go
```diff
@@ -0,0 +1,315 @@
...
+	if !hasRegistry {
+		// Add docker.io as default registry
+		if len(parts) == 1 {
+			// Single name like "alpine" or "alpine:3.18"
+			ref = "docker.io/library/" + ref
+		} else {
+			// Name with one slash like "myuser/myimage" (user namespace)
+			ref = "docker.io/" + ref
+		}
+	}
```
nalind, line 292: I _think_ that, after ruling out "scratch", this should be using go.podman.io/image/v5/docker/reference.ParseNormalizedName() and calling String() on its result.

docs/buildah-build.1.md
nalind: Any "convert" rules are processed before any specified in containers-registries.conf(5), but after substitutions specified with `--build-context`, which is something I'd like to see noted in the man page section.  That's a _lot_ of ways to override which base image is used for a particular stage.

pkg/sourcepolicy/policy.go
```diff
@@ -0,0 +1,315 @@
...
+
+	return Decision{}, false, nil
+}
+
+// Matches checks if a source identifier matches this rule's selector.
+func (r *Rule) Matches(sourceIdentifier string) (bool, error) {
+	matchType := r.Selector.MatchType
+	if matchType == "" {
+		matchType = MatchTypeExact
```
nalind: A quick skim of the protobuf and json tagging in the buildkit implementation suggests to me that the default here should be wildcard matching.

nalind (changes requested):
Overall, I rather like this.  I'd be tempted to handle wildcard matching using filepath.Match(), but it looks like the implementation we're trying to be compatible with isn't as flexible as that.  A couple of notes on the normalization of docker image names and the default matching method, but otherwise it looks pretty good to me.

pkg/sourcepolicy/policy.go
tinovyatkin: ✅ Addressed in d6eaeaa4b. Thanks @nalind\!

docs/buildah-build.1.md
tinovyatkin: ✅ Addressed in 406a7dbe9. Thanks @nalind\!

pkg/sourcepolicy/policy.go
tinovyatkin: ✅ Addressed in 23dec9cd4. Thanks @nalind\!

nalind (approved):
LGTM, restarted a test run that failed due to a known flake.

TomSweeneyRedHat:
LGTM

PR #6650, opened 2026-01-22
Run: don't try to encode SystemContext with json

nalind:
#### What type of PR is this?

/kind other

Don't expect to be able to encode a `SystemContext` as JSON, because it will trigger an encoding error if fields which can't be encoded are set.

#### How to verify it

podman has a "podman build - basic test" test that should be fixed by this.

None

```release-note
None
```

### DIFF
--- a/run.go
+++ b/run.go
@@ -177,7 +177,7 @@ type RunOptions struct {
 	// unmounted if Run() returned an error
 	ExternalImageMounts []string
 	// System context of current build
-	SystemContext *types.SystemContext
+	SystemContext *types.SystemContext `json:"-"`
 	// CgroupManager to use for running OCI containers
 	CgroupManager string
 	// CDIConfigDir is the location of CDI configuration files, if the files in


run_common.go
```diff
@@ -1165,6 +1165,8 @@ func runUsingRuntimeMain() {
 func (b *Builder) runUsingRuntimeSubproc(isolation define.Isolation, options RunOptions, configureNetwork bool, networkString string,
 	moreCreateArgs []string, spec *specs.Spec, rootPath, bundlePath, containerName, buildContainerName, hostsFile, resolvFile string,
 ) (err error) {
+	options.SystemContext = nil // has fields we can't encode as JSON, and we don't look any of them
```
Luap99: seem fine though I do wonder if we should not add `json:"-"` in RunOptions, it is already there on other fields we don't want, i.e. std streams. So I assume the json thing would be bit more consistent?

nalind: That's a good point.  Changing it.

Luap99 (approved):
LGTM

mtrmac (commented):
ACK to the extent “this clearly avoids the failure path” but I didn’t check whether anything _needs_ that value to be carried over JSON.

PR #6651, opened 2026-01-22
Bump go.podman.io/{storage,image/v5,common} to main

nalind:
#### What type of PR is this?

/kind other

Bump the versions of container-libs we use on main to catch up to its main branch.

None

```release-note
None
```


Luap99 (commented):
Thanks, looks like the "source create" test needs updating too

Luap99 (approved):
LGTM

PR #6654, opened 2026-01-26
Handle new `FROM --after` flag for explicit stage dependencies

### LINKED ISSUE #6621: Support specifying dependencies for `FROM oci-archive:` more cleanly
Right now when using `FROM $localtransport:$path` (such as `FROM oci-archive:foo.tar`), where the local artifact is built by a previous stage, one has to trick buildah into ensuring that the previous stage actually runs before the target stage. A common way to do this is to use a `RUN`/`COPY` directive in the final stage which references the previous stage:

```Dockerfile
FROM oci-archive:./out.ociarchive
RUN --mount=type=bind,from=builder,target=/var/tmp true
```

See e.g. a real Containerfile example [here](https://github.com/coreos/fedora-coreos-config/blob/459b2a406df7354fa44056c97c67efbd33b0f273/Containerfile#L59-L63).

The main issue with this in the general case is that it breaks with `--jobs N` for N != 1 because the `FROM` may end up being "pulled" before it's actually ready. A seemingly cosmetic issue is that it causes an empty rootfs layer to be appended. Though it does also mean that it's not possible for the previous stage to fully dictate the final build artifact, which is a desirable property.

Another approach I found out recently is to use `--skip-unused-stages=false`. It fixes the cosmetic issue, but it would require users/CI/build systems to pass that flag when building. It can also still break with `--jobs N` for N != 1 and obviously may end up causing unnecessary stages to be built. It works well though for the common case, which is when your stages are linear and all depend on each other (and you're not e.g. defining multiple targets and using `--target`).

So anyway, while these workarounds more or less work for now, I think it'd still be nice to have better support for this in buildah by somehow making this dependency explicit.

See also https://github.com/containers/buildah/issues/5952 in which various options for how to do this have been discussed.

jlebon:
When using local transports like `FROM oci-archive:out.ociarchive` where the artifact is built by a previous stage, buildah cannot automatically detect the dependency. There are ways to get around this but they're hacky and not foolproof (see #6621 for details).

This adds handling for the new `FROM --after=<stage>` flag. When set, it ensures that the dependency completes before the stage can begin. A tricky detail here is that the waiting needs to happen even before we "pull" the base image (since it may not exist yet), which is different from other kinds of stage dependencies.

Closes: https://github.com/containers/buildah/issues/6621

Assisted-by: Claude Opus 4.5

/kind feature

```release-note
Add `FROM --after<stage>` flag to declare explicit stage dependencies.
```

### DIFF
--- a/docs/buildah-build.1.md
+++ b/docs/buildah-build.1.md
@@ -1506,6 +1506,18 @@ buildah build --secret=id=mysecret,src=.mysecret .
 
 buildah build --source-policy-file /etc/buildah/source-policy.json -t imageName .
 
+### Using FROM --after for explicit stage dependencies
+
+When using local transports like `FROM oci-archive:file.ociarchive` where the file is produced by an earlier stage, Buildah cannot automatically detect the dependency. Use the `--after` flag on the FROM instruction to declare explicit stage dependencies:
+
+```Dockerfile
+FROM quay.io/skopeo/stable AS builder
+RUN --mount=type=bind,target=/src,rw skopeo copy docker://quay.io/fedora/fedora-minimal oci-archive:/src/fedora.ociarchive
+
+FROM --after=builder oci-archive:fedora.ociarchive
+# This stage will wait for builder to complete before evaluating FROM
+```
+
 ### Building an multi-architecture image using the --manifest option (requires emulation software)
 
 buildah build --arch arm --manifest myimage /tmp/mysrc
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -124,6 +124,7 @@ type executor struct {
 	containerMap                            map[string]*buildah.Builder // Used to map from image names to only-created-for-the-rootfs containers.
 	baseMap                                 map[string]struct{}         // Holds the names of every base image, as given.
 	rootfsMap                               map[string]struct{}         // Holds the names of every stage whose rootfs is referenced in a COPY or ADD instruction.
+	afterDependency                         map[string]string           // Maps stage names to their --after dependency.
 	blobDirectory                           string
 	excludes                                []string
 	groupAdd                                []string
@@ -547,6 +548,16 @@ func (b *executor) buildStage(ctx context.Context, cleanupStages map[int]*stageE
 	stage := stages[stageIndex]
 	ib := stage.Builder
 	node := stage.Node
+
+	// Wait for any --after deps before ib.From(node) which may try to access images
+	// via local transports (like oci-archive:) populated by those deps.
+	if afterDep, ok := b.afterDependency[stage.Name]; ok {
+		logrus.Debugf("stage %d (%s): waiting for --after dependency %q", stageIndex, stage.Name, afterDep)
+		if isStage, err := b.waitForStage(ctx, afterDep, stages[:stageIndex]); isStage && err != nil {
+			return "", nil, false, fmt.Errorf("waiting for --after=%s: %w", afterDep, err)
+		}
+	}
+
 	base, err := ib.From(node)
 	if err != nil {
 		logrus.Debugf("buildStage(node.Children=%#v)", node.Children)
@@ -808,6 +819,8 @@ func (b *executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 	// dependencyInfo is used later to mark if a particular
 	// stage is needed by target or not.
 	dependencyMap := make(map[string]*stageDependencyInfo)
+	// Initialize afterDependency map to track --after= dependency per stage
+	b.afterDependency = make(map[string]string)
 	// Build maps of every named base image and every referenced stage root
 	// filesystem.  Individual stages can use them to determine whether or
 	// not they can skip certain steps near the end of their stages.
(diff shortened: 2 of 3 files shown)


jlebon:
This requires https://github.com/openshift/imagebuilder/pull/316.

tests/bud.bats
```diff
@@ -9150,9 +9151,7 @@ EOF
     for prefix in "" / ./ ; do
       for suffix in "" ${dirsuffix} ; do
         local adjustedbase=${base/:/:${prefix}}${suffix}
-        echo FROM $adjustedbase >> ${TEST_SCRATCH_DIR}/buildcontext/Dockerfile
-        mounts="$mounts --mount=type=bind,from=$stage,target=/var/tmp/$stage"
-        echo "RUN$mounts :" >> ${TEST_SCRATCH_DIR}/buildcontext/Dockerfile
+        echo FROM '--after=${TESTARG}' $adjustedbase >> ${TEST_SCRATCH_DIR}/buildcontext/Dockerfile
```
nalind: I don't think this should be removing the test coverage for the method that people would have been using previously.

There should also be a separate test that ensures that, aside from anything that might be going on with the name of the base image or stage, the desired ordering is maintained with a higher-than-default `--jobs` value and also that the stage doesn't get skipped when there's nothing else tying the specified must-be-executed-earlier stage into the set of stages that are run for a multistage build.

nalind (changes requested):
I think we need some additional tests here.

tests/bud.bats
jlebon: Done!

The reason I had initially done it that way was that AIUI that test's main value is that `FROM $local_transport` works, not that `RUN` deps are pulled in, which is already well-covered (and is what `--after` enhances). But yeah, it's totally fine with me to keep that legacy workflow coverage!

imagebuildah/executor.go
```diff
@@ -844,6 +858,40 @@ func (b *executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
...
+							// only allow one --after flag per FROM for now; nothing necessarily
+							// semantically wrong with multiple --after, but keeping it conservative until a
+							// use case shows up
+							if _, exists := b.afterDependency[stage.Name]; exists {
+								return "", nil, fmt.Errorf("FROM --after=%s: only one --after flag is allowed per FROM instruction", after)
+							}
+							builtinArgs := argsMapToSlice(stage.Builder.BuiltinArgDefaults)
+							headingArgs := argsMapToSlice(stage.Builder.HeadingArgs)
+							userArgs := argsMapToSlice(stage.Builder.Args)
+							userArgs = append(builtinArgs, append(userArgs, headingArgs...)...)
```
Honny1, line 886: This seems odd. It works, but I would use [`slices.Concat()`](https://pkg.go.dev/slices#Concat) instead.

tests/bud.bats
```diff
@@ -9159,3 +9159,34 @@ EOF
   done
   run_buildah build ${TEST_SCRATCH_DIR}/buildcontext
 }
+
```
Honny1, line 9169: I would also like to see a test that mixes this with `COPY --from`, as that also affects ordering. Something like this:
```
FROM alpine:latest AS builder
RUN echo "Artifact Created" > /data.txt

FROM alpine:latest AS verifier
COPY --from=builder /data.txt /data.txt
RUN echo "Verifying..." && grep "Artifact" /data.txt

FROM scratch AS final --after=verifier
COPY --from=builder /data.txt /root/data.txt
```

Honny1: Also, `FROM <stage>` can impact the order, so maybe there should be a test that checks that with `--after`. I can still think of some scenarios that could be deadlocked. What will happen?

For ex:
```
FROM alpine:latest AS parent --after=child
RUN echo "I am the parent"

FROM parent AS child
RUN echo "I am the child"
```

Honny1 (commented):
Thanks for contributing. I’ve checked the code and have a few comments.

Note: Depends on: https://github.com/containers/buildah/pull/6663

tests/bud.bats
jlebon: Re. the circular dep test, the Containerfile is parsed sequentially so it'll just error out on "stage 'child' not found" right on the first line AIUI. I'll still add the test, because it does cover a new error path I added here but it might not be as interesting as you think.

I tweaked the `COPY` example to be more interesting.

imagebuildah/executor.go
jlebon: This follows the same pattern as the other branches in this function. Happy to take a stab at switching them all over to `slices.Concat()` in a follow up if you'd like!

jlebon:
Rebased to resolve conflicts and for https://github.com/containers/buildah/pull/6663! CI is now meaningful, so let's see if it passes. :crossed_fingers:

jlebon:
OK I think CI is hitting issues due to f44 branching.

imagebuildah/executor.go
Honny1: Okay, I'll keep it as is for now. If you'd like, please make the switch to `slices.Concat()` in a separate PR.

Honny1 (approved):
LGTM

PTAL @nalind

imagebuildah/executor.go
```diff
@@ -547,6 +548,17 @@ func (b *executor) buildStage(ctx context.Context, cleanupStages map[int]*stageE
...
 	node := stage.Node
+
+	// Wait for any --after deps before ib.From(node) which may try to access images
+	// via local transports (like oci-archive:) populated by those deps.
+	if afterDep, ok := b.afterDependency[stage.Name]; ok {
+		logrus.Debugf("stage %d (%s): waiting for --after dependency %q", stageIndex, stage.Name, afterDep)
+		if isStage, err := b.waitForStage(ctx, afterDep, stages[:stageIndex]); isStage && err != nil {
+			return "", nil, false, fmt.Errorf("waiting for --after=%s: %w", afterDep, err)
+		}
+		logrus.Debugf("stage %d (%s): done waiting for --after dependency %q", stageIndex, stage.Name, afterDep)
```
nalind: Non-blocking: we don't have corresponding debug around other calls to `waitForStage()`, so these seem out of place.

nalind (approved):
Non-blocking nit, otherwise LGTM.

imagebuildah/executor.go
jlebon: This was helpful in early debugging but agree we should be consistent. Dropped!

docs/buildah-build.1.md
```diff
@@ -1506,6 +1506,18 @@ buildah build --secret=id=mysecret,src=.mysecret .
 
 buildah build --source-policy-file /etc/buildah/source-policy.json -t imageName .
 
+### Using FROM --after for explicit stage dependencies
+
+When using local transports like `FROM oci-archive:file.ociarchive` where the file is produced by an earlier stage, buildah cannot automatically detect the dependency. Use the `--after` flag on the FROM instruction to declare explicit stage dependencies:
```
TomSweeneyRedHat:
```suggestion
When using local transports like `FROM oci-archive:file.ociarchive` where the file is produced by an earlier stage, Buildah cannot automatically detect the dependency. Use the `--after` flag on the FROM instruction to declare explicit stage dependencies:
```

TomSweeneyRedHat:
Sorry @jlebon one doc nit that needs tweaking.  The good news is that the update will force the failing tests to run again.  It doesn't look like the issue is related to this PR.

Otherwise
LGTM

docs/buildah-build.1.md
jlebon: Nice catch. Fixed!

Honny1:
Automerge has been disabled. @nalind @TomSweeneyRedHat Will you press the green button?

Issue #6655, opened 2026-01-25
Support "ADD --unpack" instruction


cnd4:
### Feature request description

Docker supports the `--unpack` argument for the `ADD` instruction: https://docs.docker.com/reference/dockerfile/#add---unpack

Unfortunately, when I try to run a `Containerfile` with `ADD --unpack=true`, I get the following error:

```
[3/6] STEP 3/3: ADD --unpack=true "https://example.com/archive.tar.gz" /                                     
Error: building at STEP "ADD --unpack=true "https://example.com/archive.tar.gz" /": ADD only supports the --chmod=<permissions>, --chown=<uid:gid>, --checksum=<checksum>, --link, --keep-git-dir, and --exclude=<pattern> flags
```

It would be nice to support this unpacking argument.

### Suggest potential solution

If possible, I would want Podman to understand and execute `ADD --unpack=true` instruction.

### Have you considered any alternatives?

The current alternative is to `ADD` the archive and add a `RUN` instruction in order to unpack the archive. It creates another layer so this solution may not be the best.

baude:
I'm going to move this to buildah because they provide all the building capability for podman.  thanks for your issue write up.

PR #6656, opened 2026-01-28
Add support for Dockerfile ADD --unpack flag

tinovyatkin:
## Summary

Implements BuildKit-compatible `--unpack[=true|false]` flag for the `ADD` instruction in Containerfiles/Dockerfiles. This feature brings Buildah's Dockerfile support into alignment with Docker BuildKit and enables explicit control over archive extraction behavior for both local and remote sources.

## Docker BuildKit Feature Reference

As documented in the [Docker Dockerfile reference](https://docs.docker.com/reference/dockerfile/#add---unpack):

> The `--unpack` flag manages automatic extraction of tar archives when adding files to Docker images. Local tar archives are unpacked by default, whereas remote tar archives (where `src` is a URL) are downloaded without unpacking by default.

This implementation provides the same functionality for Buildah builds, maintaining backward compatibility with existing Containerfiles.

### Default (no flag) - Unchanged
```dockerfile
ADD local.tar.gz /dest/              # ✓ Extracts (existing behavior)
ADD https://url/archive.tar.gz /dest/  # ✓ Downloads as file (existing behavior)
```

### With `--unpack=false`
```dockerfile
ADD --unpack=false local.tar.gz /dest/      # ✓ Copies as file (no extraction)
ADD --unpack=false https://url/file.tar.gz /dest/  # ✓ Downloads as file
```

### With `--unpack=true` (or bare `--unpack`)
```dockerfile
ADD --unpack=true local.tar.gz /dest/      # ✓ Extracts (same as default)
ADD --unpack=true https://url/file.tar.gz /dest/   # ✓ Extracts remote archive (NEW)
ADD --unpack=true https://url/file.txt /dest       # ✓ Regular file add (no extraction)
```

### Architecture
- **Frontend parsing** (`imagebuildah/stage_executor.go`): Parses `--unpack` flag with last-wins semantics, validates values, rejects `COPY --unpack`
- **Core extraction** (`add.go`): Pre-downloads single remote sources to detect archive status by content, enforces directory semantics for archives only, preserves file semantics for non-archives
- **Backward compatibility**: Uses tri-state `Unpack *bool` field (nil=legacy, true=extract, false=no-extract)

### Key Features
- ✅ Single-download with retry support (efficient)
- ✅ Content-based archive detection (not extension-based)
- ✅ Checksum verification before extraction
- ✅ Timestamp preservation from HTTP Last-Modified headers
- ✅ Proper destination semantics (directory for archives, file/directory for non-archives)
- ✅ Excludes patterns respected for archives, not applied to non-archive remote files

### Integration Tests (9 new tests in `tests/bud.bats`)
1. Remote archive extraction with `--unpack=true`
2. Remote archive default behavior (no extraction)
3. Local archive with `--unpack=false`
4. Remote non-archive file with `--unpack=true` (directory destination)
5. Remote non-archive to file destination
6. Remote archive without trailing slash
7. Invalid value rejection (`--unpack=maybe`)
8. Empty value rejection (`--unpack=`)
9. `COPY --unpack` rejection

### Unit Tests
- Comprehensive tests for `parseAddUnpackFlag()` with 11 scenarios
- All edge cases covered (last-wins, invalid values, mixed flags)

### Updated Tests
- 3 existing tests updated for new error message format

## Files Modified

```
 CHANGELOG.md                        |   8 +
 add.go                              | 342 +++++++++++++++++++++++++++++++
 docs/buildah-build.1.md             |   5 +
 imagebuildah/stage_executor.go      |  63 ++++++
 imagebuildah/stage_executor_test.go |  99 ++++++++++
 tests/bud.bats                      | 177 +++++++++++++++++
 6 files changed, 676 insertions(+), 18 deletions(-)
```

## Validation

- ✅ Code compiles successfully
- ✅ All unit tests pass
- ✅ Integration tests added and ready for CI
- ✅ No breaking changes to existing behavior
- ✅ Follows existing code patterns and conventions

## References

- Docker documentation: https://docs.docker.com/reference/dockerfile/#add---unpack
- Related to BuildKit feature parity

## Notes

This PR implements the feature for Dockerfile/Containerfile parsing only. CLI parity for `buildah add --unpack` is deferred as a follow-up enhancement (requires separate docs in `docs/buildah-add.1.md`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 # Changelog
 
+## Unreleased
+
+    Add support for Dockerfile/Containerfile ADD --unpack flag
+    Implements BuildKit-compatible --unpack[=true|false] flag for ADD instruction
+    --unpack=true enables extraction of remote archives (HTTP/HTTPS URLs)
+    --unpack=false disables automatic extraction of local archives
+    Default behavior (without flag) remains unchanged for backward compatibility
+
 ## v1.42.0 (2025-10-17)
 
     Bump to storage v1.61.0, image v5.38.0, common v0.66.0
--- a/add.go
+++ b/add.go
@@ -112,6 +112,11 @@ type AddAndCopyOptions struct {
 	// inheritAnnotations, newAnnotations). This field is internally managed and should
 	// not be set by external API users.
 	BuildMetadata string
+	// Unpack controls archive extraction for remote sources.
+	// nil: legacy behavior (remote URLs are not extracted; local sources depend on the `extract` argument).
+	// true: extract remote archives (tar, optionally compressed) when recognized by content.
+	// false: never extract remote archives.
+	Unpack *bool
 }
 
 // gitURLFragmentSuffix matches fragments to use as Git reference and build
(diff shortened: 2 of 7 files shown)


imagebuildah/stage_executor.go
```diff
@@ -1594,7 +1647,8 @@ func (s *stageExecutor) execute(ctx context.Context, base string) (imgID string,
 			// the content that's copied in.  We need to compute that information so that
 			// it can be used to evaluate the cache, which means we need to go ahead
 			// and copy the content.
-			canMatchCacheOnlyAfterRun = (step.Command == command.Add || step.Command == command.Copy)
+			stepCmd := strings.ToUpper(step.Command)
+			canMatchCacheOnlyAfterRun = (stepCmd == "ADD" || stepCmd == "COPY")
```
nalind, line 1651: Use `strings.EqualFold()` for this.

imagebuildah/stage_executor.go
```diff
@@ -1458,6 +1501,16 @@ func (s *stageExecutor) execute(ctx context.Context, base string) (imgID string,
 			}
 		}
 
+		// Parse --unpack flag for ADD commands
+		command := strings.ToUpper(step.Command)
+		if command == "ADD" {
```
nalind, line 1506: Use `strings.EqualFold()` to compare them.

imagebuildah/stage_executor.go
```diff
@@ -1198,6 +1205,36 @@ func (s *stageExecutor) getContentSummaryAfterAddingContent() string {
...
+			}
+			if value == "true" {
+				t := true
+				result = &t
+			} else if value == "false" {
+				f := false
+				result = &f
+			} else {
+				return nil, fmt.Errorf("invalid value %q for --unpack flag, must be true or false", value)
+			}
```
nalind, line 1232: I think the current convention is to use `strings.Cut()` for checking for the prefix and retrieving the value after it, if there is one, with one function call.

tests/bud.bats
```diff
@@ -4234,6 +4227,177 @@ _EOF
...
+
+  cat > $contextdir/Dockerfile << _EOF
+FROM alpine
+ADD --unpack=true http://0.0.0.0:${HTTP_SERVER_PORT}/archive.tar.gz /tmp/
+RUN test -f /tmp/testdir/testfile.txt
+RUN test ! -f /tmp/archive.tar.gz
+_EOF
+
+  run_buildah build $WITH_POLICY_JSON -t test-unpack-remote $contextdir
+  expect_output --substring "test-unpack-remote"
```
nalind, line 4249: If this is checking for the image name that's printed at the end of the build, then it can be removed, as the `run_buildah` helper function errors out if the build fails.

add.go
```diff
@@ -112,6 +112,11 @@ type AddAndCopyOptions struct {
 	// inheritAnnotations, newAnnotations). This field is internally managed and should
 	// not be set by external API users.
 	BuildMetadata string
+	// Unpack controls archive extraction for remote sources.
+	// nil: legacy behavior (remote URLs are not extracted; local sources depend on the `extract` argument).
+	// true: extract remote archives (tar, optionally compressed) when recognized by content.
+	// false: never extract remote archives.
+	Unpack *bool
```
nalind, line 119: Using a `types.OptionalBool` here would make assigning and comparing its value simpler.

docs/buildah-build.1.md
```diff
@@ -14,6 +14,11 @@ Builds an image using instructions from one or more Containerfiles or Dockerfile
 build context directory.  A Containerfile uses the same syntax as a Dockerfile internally.  For this
 document, a file referred to as a Containerfile can be a file named either 'Containerfile' or 'Dockerfile'.
 
+The `ADD` instruction in Containerfiles/Dockerfiles supports the `--unpack[=true|false]` flag to control
+archive extraction behavior. By default, local archives are extracted but remote archives (HTTP/HTTPS URLs)
+are not. Use `--unpack=true` to extract remote archives, or `--unpack=false` to prevent extraction of local
```
TomSweeneyRedHat, line 19: I'm not 100% certain, but this might clear up some ambiguity.  WDYT?
```suggestion
are not. Use `--unpack=true` to extract remote archives, or `--unpack=false` (the default) to prevent extraction of local
```

vendor/github.com/openshift/imagebuilder/dispatchers.go
```diff
@@ -210,8 +210,12 @@ func add(b *Builder, args []string, attributes map[string]bool, flagArgs []strin
 				return fmt.Errorf("no value specified for --exclude=")
 			}
 			excludes = append(excludes, exclude)
+		case arg == "--unpack", strings.HasPrefix(arg, "--unpack="):
```
TomSweeneyRedHat, line 213: I don't think you vendored this in from imagebuilder, and we'll need to.  We can't change code directly like this in the vendor directory.

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 # Changelog
 
+## Unreleased
```
TomSweeneyRedHat, line 5: Please remove this change.  When we do the next release, this will be auto-generated for us.

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
...
+
+    Add support for Dockerfile/Containerfile ADD --unpack flag
+    Implements BuildKit-compatible --unpack[=true|false] flag for ADD instruction
```
TomSweeneyRedHat, line 8: nit we call them options, not flags

add.go
TomSweeneyRedHat, line 119: It's been a while since I've added an option to Buildah.  But I think you need to set the default value for this somewhere down below in here.

docs/buildah-build.1.md
TomSweeneyRedHat: Also, it looks like you're adding, or at least attempting to add the `unpack` option to the Buildah Add command.  We'll need a change for that in the add man page (docs/buildah-add.md).

TomSweeneyRedHat:
@tinovyatkin thanks a bunch for this, a great start.
@nalind WDYT?

docs/buildah-build.1.md
nalind: The default is to unpack local archives, and to not unpack remote ones, so that doc change would be incorrect.

imagebuildah/stage_executor.go
```diff
@@ -82,6 +80,7 @@ type stageExecutor struct {
 	argsFromContainerfile []string
 	hasLink               bool
 	isLastStep            bool
+	unpackFlag            *bool // nil=unspecified; true/false=explicit for current instruction
```
nalind, line 83: Using a `types.OptionalBool` here would make assigning and comparing its value simpler.

add.go
```diff
@@ -258,6 +263,142 @@ func getURL(src string, chown *idtools.IDPair, mountpoint, renameTarget string,
...
+	if err := tw.WriteHeader(&hdr); err != nil {
+		return fmt.Errorf("writing tar header: %w", err)
+	}
+
+	if _, err := io.Copy(tw, reader); err != nil {
+		return fmt.Errorf("writing tar content: %w", err)
+	}
+
+	return nil
+}
```
nalind, line 400: This duplicates a lot of the logic from `getURL()`.  Since the `copier.Get()` call we use to read local sources already "knows" how to expand archives, I think it would be a smaller change to teach `getURL()` to optionally do the same, and leave the rest of the logic around calling them as-is.
An updated `getURL()` would probably always use a temporary file if it was told to expand archives, it would examine the temporary file to determine if it was an archive (using something like `copier.isArchivePath()`), and if it was found to be one, decompress the response body and send the decompressed data to the writer.  Otherwise, it would continue as it does now.

nalind (changes requested):
There are some minor changes here and there, but the `downloadURL()` and `writeTarEntry()` functions this introduces appear to be a more drastic change than we need.

Honny1:
@tinovyatkin Please link the issue in the PR description and commit message.
Issue: https://github.com/containers/buildah/issues/6655

tib-rijo:
Is there something one can do to help get this over the finish line?

Honny1:
Imho, PR is 30 days without activity. If you are willing to take over this PR. You can ask the contributor if you can take over (give the contributor some time to answer (a day or two) and finish the last bits.

ajoshua2004:
@tinovyatkin are you still going to work on it? If not, either myself or @tib-rijo can pick it up. @tib-rijo feel free to work on it if you're still interested.

PR #6664, opened 2026-02-04
internal/mkcw: make errors easier to compare, update tests

### LINKED ISSUE #6662: Probably ineffective code in internal/mkcw/archive_test.go:TestArchive
- (Original motivation) The linter reports `tautological condition: nil == nil` after the “bail now if we got an error we didn't expect” code, the `if err == nil` is always true
- It turns out `ignoreChainRetrievalErrors` is only consulted within the error handling, it is not an input to anything
- If `IgnoreAttestationErrors`, the errors are turned into warnings, so they should never be _reported_ to the caller; the test sees them, that’s a problem, regardless of the `ignore…` values
- For any kind of completely unexpected error type, the test just `return`s and does not fail. This _might_ be intentional but I see no evidence for it.

Cc: @nalind

nalind:
#### What type of PR is this?

/kind cleanup

Add Unwrap() methods to our internal bespoke mkcw error types, and have the tests check for and ignore expected error cases.  Fixes warnings from the "nilness" analyzer in `gopls`.

#### How to verify it

Updated unit tests!

Fixes #6662

```release-note
None
```

### DIFF
--- a/internal/mkcw/archive.go
+++ b/internal/mkcw/archive.go
@@ -64,9 +64,13 @@ type chainRetrievalError struct {
 
 func (c chainRetrievalError) Error() string {
 	if trimmed := strings.TrimSpace(c.stderr); trimmed != "" {
-		return fmt.Sprintf("retrieving SEV certificate chain: sevctl: %v: %v", strings.TrimSpace(c.stderr), c.err)
+		return fmt.Sprintf("retrieving SEV certificate chain: sevctl: %v: %v", trimmed, c.err)
 	}
 	return fmt.Sprintf("retrieving SEV certificate chain: sevctl: %v", c.err)
+}
+
+func (c chainRetrievalError) Unwrap() error {
+	return c.err
 }
 
 // Archive generates a WorkloadConfig for a specified directory and produces a
--- a/internal/mkcw/archive_test.go
+++ b/internal/mkcw/archive_test.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"sync"
 	"testing"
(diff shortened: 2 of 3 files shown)


internal/mkcw/archive.go
```diff
@@ -64,11 +64,18 @@ type chainRetrievalError struct {
...
+		return fmt.Sprintf("retrieving SEV certificate chain: sevctl: %v: %v", trimmed, c.err)
 	}
 	return fmt.Sprintf("retrieving SEV certificate chain: sevctl: %v", c.err)
 }
 
+func (c chainRetrievalError) Unwrap() error {
+	if trimmed := strings.TrimSpace(c.stderr); trimmed != "" {
+		return fmt.Errorf("retrieving SEV certificate chain: sevctl: %v: %v", trimmed, c.err)
+	}
+	return fmt.Errorf("retrieving SEV certificate chain: sevctl: %w", c.err)
```
Luap99: I am confused, this does not unwrap anything, instead it creates a new error that again cannot be matched at least in the trimmed branch.
The proper thing for unwrap would be to simply `return c.err` so caller can match the underlying error and not get a newly wrapped error.

nalind: Oof, line 74 should have used %w, fixing.

Luap99: No I mean wrapping here is wrong, I guess it will work with %w, but by doing so you just added an additional unnecessary Unwrap() cycle. The point of unwrap is to get the original error back, not create a new error with fmt.Errorf()

nalind: Makes sense, changed it.

Luap99 (approved):
LGTM, though I did not do a careful read of the unit test changes.

internal/mkcw/archive_test.go
```diff
@@ -100,83 +100,101 @@ func TestArchive(t *testing.T) {
...
+									}
+								}
 								return
 							}
-						}
-						if errors.As(err, &attestationError{}) {
-							if !ignoreAttestationErrors {
+							if err != nil {
 								require.NoError(t, err)
 							}
```
mtrmac: The `if` seems superfluous here.

internal/mkcw/archive_test.go
```diff
@@ -100,83 +100,101 @@ func TestArchive(t *testing.T) {
...
+								TempDir:                 t.TempDir(),
+								AttestationURL:          "http://" + addr.String(),
+								IgnoreAttestationErrors: requestIgnoreAttestationErrors,
+							}
+							inputPath := t.TempDir()
+							rc, workloadConfig, err := Archive(inputPath, ociConfig, archiveOptions)
+							// bail now if we got an error we didn't expect
+							if errors.As(err, &chainRetrievalError{}) {
+								if !ignoreChainRetrievalErrors {
+									if !errors.Is(err, os.ErrNotExist) {
```
mtrmac: (Entirely up to you: I think I’d find this easier to reason about with
```go
if errors.As(err, &chainRetrievalError{}) && (ignoreChainRetrievalErrors || errors.Is(err, os.ErrNotExist)) { return}
```
where the `require.NoError` at the bottom _always_ happens, ensuring we clearly fail on all other errors, and identifying the exceptions where we don’t fail as the exceptional cases.

internal/mkcw/archive_test.go
```diff
@@ -100,83 +100,101 @@ func TestArchive(t *testing.T) {
...
+										// not a "sevctl not installed" error
+										require.NoError(t, err)
```
mtrmac: Should “required binary not installed” mark the test as skipped? Same for the firmware.

internal/mkcw/archive_test.go
```diff
@@ -100,83 +100,101 @@ func TestArchive(t *testing.T) {
...
-						if errors.As(err, &attestationError{}) {
-							if !ignoreAttestationErrors {
+							if err != nil {
 								require.NoError(t, err)
 							}
-						}
-						return
-					}
-					if err == nil {
-						defer rc.Close()
```
mtrmac, line 147: This should not happen any more?

mtrmac (commented):
Thanks!

I still don’t understand (to be fair, I didn’t spend much time trying to) what is the purpose of `ignoreChainRetrievalErrors`, and now `ignoreMeasurementErrors`. Those variables are AFAICS only used when checking errors from `Archive`, i.e. `Archive` should (?) do the same thing both times, and we only differ in whether the test closure ends with `return` or with `require.NoError` — i.e. if it passes with `require.NoError`, by definition it should pass with `return`.

<s>And if `err != nil`, if the  AFAICS the `rc` should be `nil`, so I’d expect the code after the error checks to crash after `tar.NewReader`.</s>

internal/mkcw/archive_test.go
nalind: Changed them.

internal/mkcw/archive_test.go
nalind: Restored it.

internal/mkcw/archive_test.go
nalind: The web UI isn't doing well with the indentation change here.  This is the catch-all for cases where the error is not one of our package-local error types.

mtrmac: I don’t question the need for the `require`, it’s just that `if err == nil`, `require.NoError` succeeds (in internally includes that condition already)

internal/mkcw/archive_test.go
nalind: That does require a temporary or another call to `errors.Is()` to decide between `return`ing (effectively "succeeding") or calling `t.Skip()`.  At that point I don't know that it'd really simplify things.

mtrmac: I still don’t understand the whole structure of `ignore…` , see in the top-level comment, so I was thinking those would all be `Skip`s.

Without `s/return/Skip/`, I think it’s
```go
if errors.As(err, &chainRetrievalError{}) {
   if ignoreChainRetrievalErrors {
       return
   }
   if errors.Is(err, exec.ErrNotFound) {
      t.Skip("sevctl not found")
   }
}
// after all errors.As checks, a single:
require.NoError(t, err)
```

Anyway, this part is entirely up to you. I _would_ like to find out what I am missing about the `for _, ignore… := range []bool{…}` loops.

TomSweeneyRedHat:
LGTM
Lots of red tests, all of which appeared to be flakes, and I've restarted.

internal/mkcw/archive_test.go
```diff
@@ -100,83 +101,100 @@ func TestArchive(t *testing.T) {
...
-					}
-					inputPath := t.TempDir()
-					rc, workloadConfig, err := Archive(inputPath, ociConfig, archiveOptions)
-					// bail now if we got an error we didn't expect
-					if err != nil {
-						if errors.As(err, &chainRetrievalError{}) {
-							if !ignoreChainRetrievalErrors {
+			for _, ignoreMeasurementErrors := range []bool{false, true} {
+				for _, ignoreAttestationErrors := range []bool{false, true} {
+					for _, requestIgnoreAttestationErrors := range []bool{false, true} {
```
Honny1, line 106: I'm concerned that the nested loops for these 32 combinations might be hard to follow. I’d suggest generating all combinations within the test case struct instead to make the code a bit cleaner. This definitely isn't a blocker, though.

Honny1 (approved):
LGTM, I have a nonblocking comment.

lsm5:
@nalind LGTM, but i see the stale-pr label on this one. Just checking if you'd like to refresh this before I hit merge.

nalind:
@lsm5 I'm okay with it as-is.  Thanks!

Honny1:

giuseppe (approved):
LGTM

PR #6665, opened 2026-02-05
[release-1.42] Run: don't try to encode SystemContext with json

nalind:
#### What type of PR is this?

/kind other

Don't expect to be able to encode a `SystemContext` as JSON, because it will trigger an encoding error if fields which can't be encoded are set.

#### How to verify it

podman has a "podman build - basic test" test that should be fixed by this.

None

Cherry-picked from #6650

```release-note
None
```


Luap99 (approved):
LGTM

PR #6636, opened 2026-01-13
chroot.bats(chroot with overlay root): ensure we can overlay

nalind:
#### What type of PR is this?

/kind failing-test 

#### What this PR does / why we need it:

When setting up a chroot using overlay, if the intended upper directory is already on an overlayfs, mount a tmpfs onto it so that we can finish setting up the chroot that's an overlayfs that we'll actually test in, and copy the base image into the storage that it'll use.

#### How to verify it

Should continue passing here, and in the environment which prompted #6629.

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #6666, opened 2026-02-05
[release-1.42] chroot.bats(chroot with overlay root): ensure we can overlay

nalind:
#### What type of PR is this?

/kind failing-test 

When setting up a chroot using overlay, if the intended upper directory is already on an overlayfs, mount a tmpfs onto it so that we can finish setting up the chroot that's an overlayfs that we'll actually test in, and copy the base image into the storage that it'll use.

#### How to verify it

Should continue passing here, and in the environment which prompted #6629.

Cherry-picked from #6636.

```release-note
None
```


Luap99 (approved):
LGTM

PR #6674, opened 2026-02-06
tests/from.bats "from cpu-shares test": update cgroupv2 weights

nalind:
#### What type of PR is this?

/kind other

Update the weights expected in the "from cpu-shares test" test to account for both the old way that runtimes would convert from cgroupsv1 cpu shares to cgroupsv2 cpu weights, and the new way, catching it up with the "bud with --cpu-shares, checked" test, which will now also check with a few different values.

#### How to verify it

Updated tests!

```release-note
None
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -6669,34 +6669,31 @@ _EOF
 
   _prefetch alpine
 
-  local shares=12345
-  local expect=
-
   mytmpdir=${TEST_SCRATCH_DIR}/my-dir
   mkdir -p ${mytmpdir}
 
+  for shares in 2 200 2000 12345 20000 200000 ; do
   if is_cgroupsv2; then
     cat > $mytmpdir/Containerfile << _EOF
 FROM alpine
 RUN printf "weight " && cat /sys/fs/cgroup/\$(awk -F : '{print \$NF}' /proc/self/cgroup)/cpu.weight
 _EOF
-    # there's an old way to convert the value, and a new way to convert the value, and we don't know
-    # which one our runtime is using, so accept the values that either would compute for ${shares}
-    local oldexpect="weight $((1 + ((${shares} - 2) * 9999) / 262142))"
-    local newconverted=$(awk '{if ($1 <= 2) { print "1"} else if ($1 >= 262144) {print "10000"} else {l=log($1)/log(2); e=((((l+125)*l)/612.0) - 7.0/34.0); p = exp(e*log(10)); print int(p+1)}}' <<< "${shares}")
-    local newexpect="weight ${newconverted}"
-    expect="($oldexpect|$newexpect)"
-  else
-    cat > $mytmpdir/Containerfile << _EOF
+      local converted="$(convert_v1_shares_to_v2_weight ${shares})"
+      local expect="(weight ${converted##* }|weight ${converted%% *})"
+    else
+      cat > $mytmpdir/Containerfile << _EOF
 FROM alpine
 RUN printf "weight " && cat /sys/fs/cgroup/cpu/cpu.shares
 _EOF
-    expect="weight ${shares}"
-  fi
+      local expect="weight ${shares}"
+    fi
 
-  run_buildah build --cpu-shares=${shares} -t testcpu \
-                  $WITH_POLICY_JSON --file ${mytmpdir}/Containerfile .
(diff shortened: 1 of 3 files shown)


Honny1 (approved):
LGTM

tests/bud.bats
lsm5: @nalind do we need both of these?

nalind: Oof, neither.  Dropped them.

tests/from.bats
```diff
@@ -211,16 +211,30 @@ load helpers
...
+      # there's an old way to convert the value, and a new way to convert the value, and we
+      # don't know which one our runtime is using, so accept the values that either would
+      # compute for ${shares}
+      local oldconverted="$((1 + ((${shares} - 2) * 9999) / 262142))"
+      test -n "$oldconverted"
+      local oldexpect="weight ${oldconverted}"
+      local newconverted=$(awk '{if ($1 <= 2) { print "1"} else if ($1 >= 262144) {print "10000"} else {l=log($1)/log(2); e=((((l+125)*l)/612.0) - 7.0/34.0); p = exp(e*log(10)); if ( p == int(p) ) {print p} else { print int(p+1) }}}' <<< "${shares}")
+      test -n "$newconverted"
+      local newexpect="weight ${newconverted}"
+      local expect="($oldexpect|$newexpect)"
```
lsm5: This block seems to be a repeat of `tests/bud.bats`. Can it be made reusable? Ignore comment if overkill.

tests/from.bats
```diff
@@ -211,16 +211,30 @@ load helpers
...
-    run_buildah run $cid /bin/sh -c "cat /sys/fs/cgroup/\$(awk -F : '{print \$NF}' /proc/self/cgroup)/cpu.weight"
-    expect_output "$((1 + ((${shares} - 2) * 9999) / 262142))"
-  else
-    run_buildah run $cid cat /sys/fs/cgroup/cpu/cpu.shares
-    expect_output "${shares}"
-  fi
+  for shares in 2 200 2000 12345 20000 200000 ; do
+    run_buildah from --quiet --cpu-shares=${shares} --pull $WITH_POLICY_JSON alpine
+    cid=$output
+    if is_cgroupsv2; then
```
lsm5, line 217: won't this always be true now?

lsm5 (commented):
LGTM, apart from couple of nit questions.

tests/from.bats
nalind: Factored into a helper function.

tests/from.bats
nalind: There are still checks for this scattered across other tests, so... maybe?

TomSweeneyRedHat:
LGTM
I restarted the flaked test

lsm5 (approved):
LGTM

nalind:
I think that testing-farm:fedora-rawhide-x86_64:buildah-fedora failure's a bug.  I can't quite figure how we didn't notice it at the time, but the helper the new tests in #6675 use isn't packaged in the RPM.  I added a commit that should fix it to the tip of #6705 because that's where I noticed what was going on.

lsm5:
> I think that testing-farm:fedora-rawhide-x86_64:buildah-fedora failure's a bug. I can't quite figure how we didn't notice it at the time, but the helper the new tests in #6675 use isn't packaged in the RPM. I added a commit that should fix it to the tip of #6705 because that's where I noticed what was going on.

Was the rawhide test triggered 2 days ago separately? That would explain why only rawhide caught it and the others passed (those ran 2 weeks ago).

lsm5:
> I added a commit that should fix it to the tip of #6705 because that's where I noticed what was going on.

LGTM'd that one. Thanks!

lsm5:
#6705 is merged. So, I guess we can ignore the CI failure here and merge this.

nalind:
> > I think that testing-farm:fedora-rawhide-x86_64:buildah-fedora failure's a bug. I can't quite figure how we didn't notice it at the time, but the helper the new tests in #6675 use isn't packaged in the RPM. I added a commit that should fix it to the tip of #6705 because that's where I noticed what was going on.
> 
> Was the rawhide test triggered 2 days ago separately? That would explain why only rawhide caught it and the others passed (those ran 2 weeks ago).

I'm not sure I follow.  We run it for every PR, right?

Issue #6648, opened 2026-01-21
ARG interpolation behaves differently in multilayer builds


drmoose:
### Issue Description

An `ARG` set in a parent layer as a quoted string, `ARG quoted="words with quotes"` and then interpolated in a `RUN` as `RUN echo $quoted` will expand to `RUN echo words with quotes` on the layer that declared the ARG, but in subsequent layers of a multi-stage build will expand to `RUN echo \"words with quotes\"`, which differs from the current docker build behavior.

### Steps to reproduce the issue

Adding the following conformance test to `tests/conformance` fails:

```go
// tests/conformance/conformance_test.go
	{
		name:         "quoted-arg",
		dockerfile:   "Dockerfile.quoted-arg",
		fsSkip:       []string{"(dir):arg-expansion.txt:mtime"},
	},
```

```Dockerfile
# tests/conformance/testdata/Dockerfile.quoted-arg
FROM quay.io/libpod/centos:7 AS lower

ARG quoted="words with quotes"

RUN printf "%q\n" lower: $quoted >> /arg-expansion.txt

FROM lower AS upper

RUN printf "%q\n" upper: $quoted >> /arg-expansion.txt
```

### Describe the results you received

`podman run $(podman build -q - < tests/conformance/testdata/Dockerfile.quoted-arg) cat /arg-expansion.txt` says:

```
lower:
words
with
quotes
upper:
\"words
with
quotes\"
```

Suggesting something is implicitly escaping the quotes rather than simply inserting them verbatim.

### Describe the results you expected

`docker run $(docker build -q - < tests/conformance/testdata/Dockerfile.quoted-arg) cat /arg-expansion.txt` says:

```
lower:
words
with
quotes
upper:
words
with
quotes
```

ie, the "lower:" and "upper:" results are the same, and the unquoted variable is behaving like a normal shell variable

### buildah version output

```yaml
Version:         1.42.2
Go Version:      go1.25.5 X:nodwarf5
Image Spec:      1.1.1
Runtime Spec:    1.2.1
CNI Spec:        1.1.0
libcni Version:  v1.3.0
image Version:   5.38.0
Git Commit:      c0cc97255c76f7a3d87bfdf5e7b97f013f0bfe85
Built:           Wed Dec  3 08:00:04 2025
OS/Arch:         linux/amd64
BuildPlatform:   linux/amd64
```

### buildah info output

```yaml
{
    "host": {
        "CgroupVersion": "v2",
        "Distribution": {
[...]

PR #6679, opened 2026-02-09
Add a test where a default ARG value is a quoted string

nalind:
#### What type of PR is this?

/kind bug

Add a test reproducer from #6648.

#### How to verify it

New conformance test!

Verifies that #6648 was fixed by #6663.

```release-note
In multi-stage builds, default ARG values inherited from a base stage by a later stage should now handle quoting consistently.
```


Honny1 (approved):
LGTM

TomSweeneyRedHat:
LGTM

PR #6683, opened 2026-02-10
Fix call to chown

### LINKED ISSUE #6682: Chown in copier.go passes uid for both uid/gid
### Issue Description

I work as a developer on z/OS, and part of my job involves upgrading the version of podman we provide on z/OS. Recently, I noticed some code in buildah that I believe to be an issue. Specifically here -https://github.com/containers/buildah/blob/main/copier/copier.go#L2408
```
if err = chown(filepath.Join(req.Root, leaf), uid, uid); err != nil {
	return errorResponse("copier: ensure: error setting owner of %q to %d:%d: %v", leaf, uid, gid, err)
}
```
I think it should be passing in `uid, gid` instead of `uid, uid` since they could be different. Maybe this was just a typo since uid is so similar to gid? Or is there something I'm missing?

N/A

N/A

N/A

### buildah version output

```yaml
N/A
```

### buildah info output

```yaml
N/A
```

### Provide your storage.conf

```toml
N/A
```

### Upstream Latest Release

Yes

### Additional information

stilwelb:
#### What type of PR is this?

> /kind bug

This PR fixes #6682 by correcting the call to chown to pass the gid instead of a uid.

Fixes #6682 

```
None
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -2329,7 +2329,7 @@ func copierHandlerEnsure(req request, idMappings *idtools.IDMappings) *response 
 	for _, item := range req.EnsureOptions.Paths {
 		uid, gid := 0, 0
 		if item.Chown != nil {
-			uid, gid = item.Chown.UID, item.Chown.UID
+			uid, gid = item.Chown.UID, item.Chown.GID
 		}
 		var mode os.FileMode
 		switch item.Typeflag {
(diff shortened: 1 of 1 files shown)


nalind:
Affects release-1.41, release-1.42, release-1.43.

nalind (approved):
Non-blocking: there's a similar mistake at line 2332, which can either be fixed here or in a follow-up PR.  LGTM either way.  Thanks!

stilwelb:
> Non-blocking: there's a similar mistake at line 2332, which can either be fixed here or in a follow-up PR. LGTM either way. Thanks!

Good catch! I pushed another commit to address it since they are related.

nalind (approved):

Honny1 (approved):
LGTM. This ideally should have simple unit tests, but I am not sure if the cost of testing typos is worth it.

giuseppe (commented):
/lgtm

giuseppe (approved):

PR #6687, opened 2026-02-13
[release-1.33] Use separate fork of crypto for OCP Builder

TomSweeneyRedHat:
**DO NOT MERGE**

Use a private crypto branch in hopes that the OCP Builder team can avoid bumping Go above v1.22.

Per request of Prabhakar Palepu

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```


lsm5:
@TomSweeneyRedHat It's not really `private` though, is it? Maybe just call it `fork` ? Probably not worth a commit amend unless there are other changes. 

I think it'd be useful to have ack(s) on this PR from Prabahakar (requester) and/or Atharva (fork owner) before we merge this.

go.mod
```diff
@@ -1,5 +1,9 @@
 module github.com/containers/buildah
 
+// Note the replace at the bottom of this file in order
+// to keep the Go command at a lower level to help with
+// OCP Builder dependencies.
+
 go 1.24.0
```
nalind, line 7: If the goal is to stay on a version of the compiler older (or "older") than 1.24, I'd have expected this to be patched to that version so that the compiler would complain if something else was pulling in a newer ("newer") version.

TomSweeneyRedHat:
@lsm5 good point on the private bit, I'll rename the title.  It is just a fork.

@nalind I had changed Go "by hand" to 1.22 in the go.mod, but the `go mod` `tidy/vendor/verify` tweaked it back to 1.24.  I pinged Prabahakar about that fun.  I suspect another module besides runc is pulling in 1.24.

nalind:
> @nalind I had changed Go "by hand" to 1.22 in the go.mod, but the `go mod` `tidy/vendor/verify` tweaked it back to 1.24. I pinged Prabahakar about that fun. I suspect another module besides runc is pulling in 1.24.

The `vendor/modules.txt` file should have that information.

TomSweeneyRedHat:
@nalind excellent point that I completely forgot about, thanks!

TomSweeneyRedHat:
And it looks like all of the 1.24 needs are coming from golang.org/x/*, which may be getting pulled in due to the crypto bump.

TomSweeneyRedHat:
I'm going to try this from another angle, and I'm going to close this.

PR #6688, opened 2026-02-16
ignore ErrLayerUnknown in cache lookup

### LINKED ISSUE #5674: race: parallel builds: copying...committing...creating... layer not known

Luap99:
In podman CI I commonly see a flake where build fails with: checking if cached image exists from a previous build: getting top layer info: layer not known

This commit fixes it by simply ignoring the error, if the layer was removed between listing and looking up then this is fine and we can just continue looking for more layers.

Fixes: #5674

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
Fixed a race condition during cache lookup that made the build fail when a storage layer was removed in parallel
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -2438,6 +2438,9 @@ func (s *stageExecutor) intermediateImageExists(ctx context.Context, currNode *p
 		if image.TopLayer != "" {
 			imageTopLayer, err = s.executor.store.Layer(image.TopLayer)
 			if err != nil {
+				if errors.Is(err, storage.ErrLayerUnknown) {
+					continue
+				}
 				return "", fmt.Errorf("getting top layer info: %w", err)
 			}
 			// Figure out which layer from this image we should


Honny1 (approved):
LGTM

nalind (approved):
LGTM

PR #6690, opened 2026-02-17
imagebuildah.stageExecutor.Run(): pull images for transient mounts

nalind:
#### What type of PR is this?

/kind bug

When `buildah build` is passed a `--mount` option that references an image in its "from" value, instead of pulling the image, as we do when the `--mount` option is included in the Dockerfile, we would return an error.  Correct the difference in behavior.

#### How to verify it

New integration test!

None

This does nothing for a similar error when `buildah run` is given a `--mount` flag which references an image which hasn't been pulled.  I'm not sure about that.

```release-note
The `--mount` option to `buildah build` no longer fails with an error at RUN instructions if the value for the `--mount` option references an image as its "from" value, and the image has not previously been pulled.
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -811,7 +811,7 @@ func (s *stageExecutor) Run(run imagebuilder.Run, config docker.Config) error {
 			args = []string{full}
 		}
 	}
-	stageMountPoints, err := s.runStageMountPoints(run.Mounts)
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
LGTM

I restarted the failed test.  I'm 95% sure it's a flake, but we'll see.

TomSweeneyRedHat:
@lsm5 PTAL
If it LGTY, and we can ignore the failing test, please merge.

lsm5 (approved):
LGTM

PR #6691, opened 2026-02-17
WIP: Add --tls-details

mtrmac:
#### What type of PR is this?

> /kind feature

#### What this PR does / why we need it:

Add `--tls-details` options to many CLI operations, exposing https://github.com/containers/container-libs/pull/623 . Hopefully this will also be sufficient for integration into Podman, confirmation of that TBD.

#### How to verify it

TBD. This should have at least some smoke tests

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

I’m fairly unsatisfied with the CLI top level:
- I have intentionally been very conservative, adding a `--tls-details` option wherever there is a `SystemContext`. Some of that is almost certainly unnecessary (`buildah tag`), but I wanted to avoid surprises like `buildah manifest inspect`. I’m not sure how else to make this _maintainable_ over a longer term, ensuring that contributors of new features don’t forget to add the `--tls-details` option.
- The way `parse.SystemContextFromOptions` silently ignores undeclared options, and requires the CLI layer to individually declare each new option, makes for a finely tuned CLI options sets but it’s fairly hard to track or ensure consistency. (I have no ambition of significantly restructuring that in this PR.)

#### Does this PR introduce a user-facing change?

```release-note
Many subcommands now expose a `--tls-details` option, allowing to tune TLS settings using a `containers-tls-details.yaml(5)` file.
```

PR #6692, opened 2026-02-18
Stop using the old github.com/docker/docker package paths

mtrmac:
#### What type of PR is this?

> /kind other

Update the Docker client to a newer version and the updated module paths / API.

Some indirect references remain in `github.com/fsouza/go-dockerclient` and `github.com/openshift/imagebuilder`.

#### How to verify it

Existing tests

None

This is a subset of #6691 , useful there to limit the scope of the audit. And I think it needs to happen either way.

```release-note
None
```


mtrmac:
Apparently I broke something, looking…

tests/conformance/conformance_test.go
```diff
@@ -332,17 +331,21 @@ func testConformanceInternal(t *testing.T, dateStamp string, testIndex int, muta
...
-			t.Skipf("%v", err)
+		negotiatedVersion := dockerClient.ClientVersion()
+		if versions.LessThan(negotiatedVersion, "1.38") {
+			t.Skipf("negotiated version %q is too low", err)
 		}
 	}
 
 	// connect to dockerd using go-dockerclient
-	client, err := docker.NewVersionedClientFromEnv(dockerapi.DefaultVersion)
+	client, err := docker.NewVersionedClientFromEnv(dockerdockerclient.MaxAPIVersion)
```
nalind: This is newer than the version of dockerd our conformance tests are talking to knows about.

mtrmac: Yes. What was the motivation for https://github.com/containers/buildah/pull/6602/changes/8d6f3fcf126ba4c1ae75abeec4f6c04e9ced97ac ?

I don’t understand using `fsouza/go-dockerclient` with “the latest version supported by docker/docker/client”; the two are not obviously related. (And I can’t see any API to see what is the latest version supported by `fsouza/go-dockerclient`.)

AFAICT the daemon treats API requests with no version specified as "the server’s current version”, so the previous version of that was already using “whatever is newest” and assuming that `fsouza/go-dockerclient` can handle it.

So I’m going to try reverting https://github.com/containers/buildah/pull/6602/changes/8d6f3fcf126ba4c1ae75abeec4f6c04e9ced97ac .

Alternatively, we could hard-code `1.51` here to exactly preserve today’s behavior.

mtrmac: … OK I see why that was necessary. Hard-coding `1.51` then, and adding a comment outlining the reasoning.

mtrmac:
(I have also added a commit to rename the package references, to be hopefully less confusing in the far future.)

tests/conformance/conformance_test.go
nalind: When not specified, an API version is set in BuildImage() based on which fields we set in the build options structure that we pass to it.  For the conformance tests, this often ends up landing on [1.39](https://cirrus-ci.com/task/6015204598743040?logs=conformance_test#L332) or [1.25](https://cirrus-ci.com/task/6015204598743040?logs=conformance_test#L317), depending on which fields we specify.
The version of the daemon in our debian VMs only understands 1.50, so tests were failing when attempting to use 1.51.

mtrmac:
Tests are passing now.

tests/conformance/conformance_test.go
nalind: Hmm, I must be mistaken about the version of dockerd in the VM - I guess that was a limitation only for imagebuilder.

nalind (approved):
LGTM, thanks!

lsm5 (approved):
LGTM

PR #6695, opened 2026-02-20
fix: support SHELL during RUN commands in image build

### LINKED ISSUE #6460: Please support SHELL even in OCI builds
### Issue Description

If you try to build a Containerfile that contains a SHELL directive, the build fails with "SHELL is not supported for OCI image format".  This is explicit logic in buildah, but I think it's misunderstanding the way that SHELL works in Docker, in BuildKit and in everything else compatible with them (including OCI tools!)

If I build this trivial Containerfile/Dockerfile:

```
FROM alpine:latest
RUN apk add --no-cache bash
SHELL ["/bin/bash", "-c"]
RUN if [[ "a" == "a" ]]; then true; fi
```

using BuildKit, it works fine.  And I get an image that doesn't have a shell set, because it's OCI and that's not a thing.  But that's not the point -- SHELL is part of the build process.  Even the official Dockerfile reference makes no mention whatsoever of OCI in the SHELL documentation.  And this usage of SHELL is very, very common.

I understand that the current behavior may represent a sort of principled stance, but it's a principle that makes writing Containerfiles unnecessarily awkward and that is not compatible with anything else.

See https://github.com/containers/podman/issues/8477

The code is here: https://github.com/containers/buildah/blob/e0e108c7cc17c4ccb443fb25c93675934b62378e/config.go#L389

Steps to reproduce the issue
1. Try to build the above Containerfile in the default OCI mode

buildah refuses to build it.

It should work.

### buildah version output

```yaml
$ buildah version
Version:         1.33.7
Go Version:      go1.22.2
Image Spec:      1.1.0-rc.5
Runtime Spec:    1.1.0
CNI Spec:        1.0.0
libcni Version:  
image Version:   5.29.2
Git Commit:      
Built:           Wed Dec 31 16:00:00 1969
OS/Arch:         linux/amd64
BuildPlatform:   linux/amd64
```

### buildah info output

```yaml
$ buildah info
{
    "host": {
        "CgroupVersion": "v2",
        "Distribution": {
            "distribution": "ubuntu",
            "version": "24.04"
        },
        "MemFree": 1270116352,
        "MemTotal": 134898020352,
        "OCIRuntime": "crun",
        "SwapFree": 0,
        "SwapTotal": 0,
        "arch": "amd64",
        "cpus": 32,
        "hostname": "21a",
        "kernel": "6.8.0-52-generic",
        "os": "linux",
        "rootless": true,
        "uptime": "6288h 16m 40.84s (Approximately 262.00 days)",
        "variant": ""
    },
    "store": {
        "ContainerStore": {
            "number": 12
        },
        "GraphDriverName": "overlay",
        "GraphOptions": null,
        "GraphRoot": "/vol0/encrypted/home/luto/.local/share/containers/storage",
        "GraphStatus": {
            "Backing Filesystem": "zfs",
            "Supports d_type": "true",
            "Supports shifting": "false",
            "Supports volatile": "true",
            "Using metacopy": "false"
        },
        "ImageStore": {
            "number": 187
        },
        "RunRoot": "/run/user/1002/containers"
    }
}
```

### Provide your storage.conf

```toml
N/A
```

### Upstream Latest Release

No

### Additional information

_No response_

aeijdenberg:
Currently adding a SHELL instruction results in a WARN message, but then continues to build the image, using the incorrect shell (the default shell) if the output format is to OCI.

While the OCI spec doesn't allow persistence of the SHELL value, it also doesn't actually define the format of a Dockerfile/Containerfile, and the primary examples given in Docker documentation for use of SHELL is to affect subsequent RUN commands, which is a build time operation.

Since this is a build time operation, the final output format shouldn't cause a change in behaviour.

This removes the conditional that disables it, and adjusts the WARN message.

#### What type of PR is this?

> /kind bug

I think it fixes a bug - because previously it would emit a `WARN` but then not actually fail, but instead use the wrong shell. 

Better compatiblity with Dockerfiles.

#### How to verify it

See test in tests/bud.bats

Fixes #6460.

Docker docs:
https://docs.docker.com/reference/dockerfile/#shell

There's some interesting and relevant discussion in:
https://github.com/opencontainers/image-spec/issues/1248#issuecomment-2691162618

```release-note
`SHELL` instructions will now affect subsequent `RUN` instructions in a `Containerfile`, even when the format is OCI
```


nalind (approved):
LGTM, thanks!

TomSweeneyRedHat:
LGTM
Thanks @aeijdenberg

PR #6697, opened 2026-02-23
chroot: error out on --network != host when $BUILDAH_ISOLATION

### LINKED ISSUE #6696: --network conflicts with --isolation=chroot, but not with BUILDAH_ISOLATION=chroot
### Issue Description

As explained in https://github.com/containers/buildah/pull/4257, the `--network` option doesn't do what is expected with chroot isolation, so the two CLI options were made to conflict.

However, the same does not apply to setting chroot isolation via the BUILDAH_ISOLATION environment variable. This can give a false impression that `--network=none` works, because it still has the effect of removing `/etc/resolv.conf` to break DNS resolution.

Should the chroot isolation check also apply to the environment variable? Note that this may have a larger blast radius than the CLI option, because the `quay.io/containers/buildah` image sets `BUILDAH_ISOLATION=chroot` by default.

Steps to reproduce the issue
1. `buildah build -t test --network=none --isolation=chroot` - errors as expected
2. `BUILDAH_ISOLATION=chroot buildah build -t test --network=none` - proceeds with the build

`--isolation=chroot` conflicts with `--network`

`BUILDAH_ISOLATION=chroot` does not conflict with `--network`

Both ways of setting chroot isolation conflict with `--network`

### buildah version output

```yaml
Version:         1.42.2
```

### buildah info output

```yaml
Probably not relevant
```

### Provide your storage.conf

```toml
Probably not relevant
```

### Upstream Latest Release

Yes

### Additional information

_No response_

nalind:
#### What type of PR is this?

/kind bug

Error out if someone tries to mix the --network CLI flag, with a value other than "host", with chroot isolation configured by setting $BUILDAH_ISOLATION to "chroot", as we have been doing for cases where it's done using the "--isolation" option, instead of just ignoring it as we did previously.

#### How to verify it

Updated integration test!

Fixes #6696 

```release-note
Attempting to use the `--network` option with a value other than "host" when "BUILDAH_ISOLATION" is set to "chroot" in the environment now produces an error, as it would with `--isolation=chroot`.
```

### DIFF
--- a/pkg/cli/build.go
+++ b/pkg/cli/build.go
@@ -331,7 +331,7 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
 		}
 	}
 
-	if c.Flag("network").Changed && c.Flag("isolation").Changed {
+	if c.Flag("network").Changed {
 		if isolation == define.IsolationChroot {
 			if ns := namespaceOptions.Find(string(specs.NetworkNamespace)); ns != nil {
 				if !ns.Host {
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -569,6 +569,9 @@ _EOF
 
   run_buildah 125 build --network=none --isolation=chroot $WITH_POLICY_JSON ${TEST_SCRATCH_DIR}
   expect_output --substring "cannot set --network other than host with --isolation chroot"
+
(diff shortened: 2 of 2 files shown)


TomSweeneyRedHat:
LGTM

Honny1 (approved):
LGTM

PR #6698, opened 2026-02-25
Use containers.conf engine.platform as default platform

TrevorBurnham:
Depends on: containers/container-libs#669

#### What type of PR is this?

/kind feature

When no --platform, --os, --arch, or --variant flags are specified, fall back to the platform value from containers.conf [engine] section.

#### How to verify it

1. Set a non-native platform in `containers.conf`
2. Run a pull without using the `--platform` flag

For example:

```toml
# containers.conf
[engine]
platform = "linux/s390x"
```

```sh
buildah pull alpine
buildah inspect --format '{{.OCIv1.Architecture}}' alpine
```

Expect: s390x

Related to https://github.com/containers/podman/issues/25641

Yes:

```release-note
Buildah now respects the `platform` setting in the `[engine]` section of `containers.conf`. Explicit `--platform`, `--os`, `--arch`, or `--variant` flags continue to take precedence.
```

### DIFF
--- a/tests/containers_conf.bats
+++ b/tests/containers_conf.bats
@@ -173,3 +173,31 @@ EOF
     expect_output --substring "retry.*\(default 10\)"
     expect_output --substring "retry-delay.*\(default \"5s\"\)"
 }
+
+
+@test "containers.conf engine.platform" {
+    cat >${TEST_SCRATCH_DIR}/containers.conf << EOF
+[engine]
+platform = "linux/arm64"
+EOF
+    local context="$TEST_SCRATCH_DIR"/context
+    mkdir -p "$context"
+    cat > "$context"/Containerfile << _EOF
+FROM scratch
+COPY . .
+_EOF
+
+    # Build with the custom containers.conf and push to an OCI layout
+    CONTAINERS_CONF=${TEST_SCRATCH_DIR}/containers.conf run_buildah build \
+        $WITH_POLICY_JSON -t localhost/testplatform \
+        -f "$context"/Containerfile "$context"
+    CONTAINERS_CONF=${TEST_SCRATCH_DIR}/containers.conf run_buildah push \
+        $WITH_POLICY_JSON localhost/testplatform oci:"$TEST_SCRATCH_DIR"/output
+
+    # Verify the image platform matches what we set in containers.conf
+    local config="$TEST_SCRATCH_DIR"/output/$(oci_image_config "$TEST_SCRATCH_DIR"/output)
+    run jq -r '.os' "$config"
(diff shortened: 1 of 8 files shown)


nalind:
Try adding this to the end of the top-level go.mod:
```go.mod

replace (
	go.podman.io/storage => github.com/TrevorBurnham/container-libs/storage add-engine-platform-config
	go.podman.io/image/v5 => github.com/TrevorBurnham/container-libs/image/v5 add-engine-platform-config
	go.podman.io/common => github.com/TrevorBurnham/container-libs/common add-engine-platform-config
)
```
followed by a `go mod tidy && go mod vendor` to pull in the changes from container-libs.

TomSweeneyRedHat:
@TrevorBurnham looks like this needs a rebase now.

tests/containers_conf.bats
```diff
@@ -166,3 +166,38 @@ EOF
...
+    # Verify the image platform matches what we set in containers.conf
+    run jq -r '.manifests[0].digest' "$TEST_SCRATCH_DIR"/output/index.json
+    assert $status -eq 0
+    local alg="${output%%:*}"
+    local hex="${output##*:}"
+    run jq -r '.config.digest' "$TEST_SCRATCH_DIR"/output/blobs/"$alg"/"$hex"
+    assert $status -eq 0
+    alg="${output%%:*}"
+    hex="${output##*:}"
+    run jq -r '.os' "$TEST_SCRATCH_DIR"/output/blobs/"$alg"/"$hex"
```
nalind: This section can probably set `local config="$TEST_SCRATCH_DIR"/output/$(oci_image_config "$TEST_SCRATCH_DIR"/output)`, and then point `jq` at `"$config"`.

tests/containers_conf.bats
```diff
@@ -166,3 +166,38 @@ EOF
     expect_output --substring "retry.*\(default 10\)"
     expect_output --substring "retry-delay.*\(default \"5s\"\)"
 }
+
+
+@test "containers.conf engine.platform" {
+    cat >${TEST_SCRATCH_DIR}/containers.conf << EOF
+[engine]
+platform = "linux/amd64"
+EOF
```
nalind, line 182: Most of our tests run on amd64, so I'm not sure that this ensures that we're using the value from the configuration file.

tests/containers_conf.bats
TrevorBurnham: Done.

tests/containers_conf.bats
TrevorBurnham: Good call. I've switched to `linux/arm64` so the test actually verifies that the config is being used.

PR #6699, opened 2026-02-27
imagebuildah: avoid empty layer in single-layer build path

jlebon:
In the single-layer build path (i.e. with `--layers=false`), the final commit always created a layer even when only metadata instructions were executed after `FROM`.

Fix this by tracking if any instruction actually created data so we can avoid creating a layer in that case.

Noticed this when looking at images passed through chunkah, where one can specify the number of layers N to output; the resulting image would have a total of N+1 layers because it's often the case that you want to add metadata after the `FROM oci-archive:...`.

[Assisted-by](https://gist.github.com/jlebon/dd287bd01218a59bb41d000c9e7d3897): Claude Opus 4.6

/kind bug

```release-note
Stop creating an empty layer when using `--layers=false` and adding metadata only (e.g. `LABEL`).
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -1412,6 +1412,7 @@ func (s *stageExecutor) execute(ctx context.Context, base string) (imgID string,
 		logImageID(imgID)
 	}
 
+	executedLayerStep := false
 	for i, node := range children {
 		logRusage()
 		moreInstructions := i < len(children)-1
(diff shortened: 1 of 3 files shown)


nalind (approved):
I vaguely recall OpenShift having a build test that checked how many layers a test build added to its output image compared to the base image; it's why we added the `stepRequiresLayer()` helper function and started using it where we do in multi-layer builds.  I don't think that it was a Layers==false (imageOptimizationPolicy==SkipLayers) case, though, so it's probably not going to be tripped up by this change.
Anyway, LGTM.  Thanks!

TomSweeneyRedHat:
LGTM

PR #6700, opened 2026-02-27
Add a more generic "prepend or append instructions" method

nalind:
#### What type of PR is this?

/kind cleanup

Refactor our handling of `--env` and `--label` options so that it'll be easier to add more instructions at the beginning and end of stages.

#### How to verify it

This shouldn't break the current set of tests.

None

```release-note
None
```

### DIFF
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -545,6 +545,7 @@ func (b *executor) getImageTypeAndHistoryAndDiffIDs(ctx context.Context, imageID
 }
 
 func (b *executor) buildStage(ctx context.Context, cleanupStages map[int]*stageExecutor, stages imagebuilder.Stages, stageIndex int) (imageID string, commitResults *buildah.CommitResults, onlyBaseImage bool, err error) {
+	var prependInstructions, appendInstructions []string
 	stage := stages[stageIndex]
 	ib := stage.Builder
 	node := stage.Node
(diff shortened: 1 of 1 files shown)


nalind:
@containers/buildah-maintainers PTAL, I'd like to make this available for use in #6556

imagebuildah/executor.go
```diff
@@ -569,36 +570,29 @@ func (b *executor) buildStage(ctx context.Context, cleanupStages map[int]*stageE
...
 				}
 			}
-			if len(labelLine) > 0 {
-				additionalNode, err := imagebuilder.ParseDockerfile(strings.NewReader("LABEL" + labelLine + "\n"))
-				if err != nil {
-					return "", nil, false, fmt.Errorf("while adding additional LABEL step: %w", err)
-				}
-				stage.Node.Children = append(stage.Node.Children, additionalNode.Children...)
-			}
+			appendInstructions = slices.Concat(appendInstructions, []string{labelLine})
```
mtrmac, line 587: Isn’t this just `append`? Also in the other code.

mtrmac (commented):
LGTM — feel free to merge with or without the `append` part.

---

_Generally_, I’d aesthetically prefer an architecture where the in-memory data is constructed ~directly (perhaps with a `createNameVal` counterpart to `parseNameVal` instead of the round-trip through a text syntax (and worrying about escaping control syntax), but 1) this is pre-existing, and 2) we would need to construct `Node.Original` anyway, so there would not be much of a benefit.

imagebuildah/executor.go
nalind: My expectation is that we'll need to turn `prependInstructions` and `appendInstructions` into arguments for this method in #6556, so we'd need to change this instance to something similar.  You're right about the other one.

mtrmac: ACK, `slices.Concat` is certainly safer.

mtrmac (approved):

PR #6701, opened 2026-02-28
tests: Adapt tests to run on architectures other than amd64

ricardobranco777:
Adapt tests to run on architectures other than amd64. 

Some images - `quay.io/hummingbird/git` are available for certain architectures and others like `quay.io/libpod/busybox` require a statically compiled busybox.  Introduce `skip_unless_arch helper` function and skip some tests if we're not running on arm64 and/or amd64 so we can run the testsuite on other architectures. 

Otherwise tests fail on:
- aarch64: http://openqa-assets.opensuse.org/tests/5705837/file/buildah-buildah-root.tap.txt
- ppc64le: https://openqa-assets.opensuse.org/tests/5705116/file/buildah-buildah-root.tap.txt

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation

/kind failing-test

> /kind feature
> /kind flake
> /kind other

We need it to be able to run the testsuite on architectures other than amd64.

#### How to verify it

openQA links will follow.

None

The alternative to this PR is to change these images:

- The `k8s.gcr.io/pause` image is only available for amd64.
- The `quay.io/hummingbird/git` image is only available for amd64 & arm64.

```release-note
None
```

### DIFF
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -754,6 +754,24 @@ function skip_if_no_unshare() {
 }
 
 ######################
+#  skip_unless_arch  #
+######################
+function skip_unless_arch() {
+  local goarch
+  goarch="$(go env GOARCH 2>/dev/null)"
+
+  [ -z "$goarch" ] && return 0
+
+  for arch in "$@"; do
+    if [ "$arch" = "$goarch" ]; then
+      return 0
+    fi
+  done
+
+  skip "test requires one of these architectures: $* (current: $goarch)"
+}
+
+######################
 #  start_git_daemon  #
 ######################
 function start_git_daemon() {
--- a/tests/blobcache.bats
+++ b/tests/blobcache.bats
@@ -102,8 +102,8 @@ function _check_matches() {
 	blobcachedir=${TEST_SCRATCH_DIR}/cache
 	mkdir -p ${blobcachedir}
 	# Pull an image using a fresh directory for the blob cache.
-	run_buildah from --quiet --blob-cache=${blobcachedir} $WITH_POLICY_JSON registry.k8s.io/pause
-	ctr="$output"
+	run_buildah from --quiet --cidfile ${TEST_SCRATCH_DIR}/cid --blob-cache=${blobcachedir} $WITH_POLICY_JSON registry.k8s.io/pause
+	ctr="$(< ${TEST_SCRATCH_DIR}/cid)"
 	run_buildah add ${ctr} $BUDFILES/add-file/file /
 	# Commit the image without using the blob cache, using compression so that uncompressed blobs
 	# in the cache which we inherited from our base image won't be matched.
@@ -131,8 +131,8 @@ function _check_matches() {
 	blobcachedir=${TEST_SCRATCH_DIR}/cache
 	mkdir -p ${blobcachedir}
 	# Pull an image using a fresh directory for the blob cache.
-	run_buildah from --quiet --blob-cache=${blobcachedir} $WITH_POLICY_JSON registry.k8s.io/pause
-	ctr="$output"
+	run_buildah from --quiet --cidfile ${TEST_SCRATCH_DIR}/cid --blob-cache=${blobcachedir} $WITH_POLICY_JSON registry.k8s.io/pause
+	ctr="$(< ${TEST_SCRATCH_DIR}/cid)"
 	run_buildah add ${ctr} $BUDFILES/add-file/file /
 	# Commit the image using the blob cache.
 	ls -l ${blobcachedir}
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -2351,6 +2351,9 @@ _EOF
 }
 
 @test "bud-multistage-copy-final-slash" {
+  # This test requires a statically linked busybox which is not
+  # the case for quay.io/libpod/busybox on ppc64le & s390x
+  skip_unless_arch amd64 arm64
   _prefetch busybox
   target=foo
   run_buildah build $WITH_POLICY_JSON -t ${target} $BUDFILES/dest-final-slash
@@ -7953,6 +7956,8 @@ _EOF
 }
 
 @test "bud with ADD with git repository source" {
+  # quay.io/hummingbird/git is only available on these arches:
+  skip_unless_arch amd64 arm64
   _prefetch quay.io/hummingbird/git # any image with git preinstalled would do
 
   local repodir=${TEST_SCRATCH_DIR}/repository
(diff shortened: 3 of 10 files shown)


mtrmac (commented):
A quick drive-by comment: I think it would be valuable to do document _why_ a test is being skipped.

Either `skip_unless_arch amd64 # registry.k8s.io/pause only exists for this architecture` or, if there are very few reasons, add helpers for specific causes, `skip_if_k8s_pause_image_unavailable` and the like.

tests/helpers.bash
```diff
@@ -753,6 +753,24 @@ function skip_if_no_unshare() {
...
+
+  [ -z "$goarch" ] && return 0
+
+  for arch in "$@"; do
+    if [ "$arch" = "$goarch" ]; then
+      return 0
+    fi
+  done
+
+  skip "test requires one of these archictectures: $* (current: $goarch)"
```
nalind: Typo: "archictectures"

nalind (changes requested):
I have to concur with @mtrmac: without an indication of why a test fails, it's going to be hard to tell in the future why a test was disabled, and if the conditions that caused us to disable them are no longer applicable.

At first blush, tests which fail because of the "image platform (linux/amd64) does not match the expected platform" warning could be updated to use `from`'s `--cidfile` flag to save the container ID to a temporary file, which could then be read back.  The ppc64le busybox image appears to be dynamically linked, so the static binary itself can't just be copied and run.

ricardobranco777:
> I have to concur with @mtrmac: without an indication of why a test fails, it's going to be hard to tell in the future why a test was disabled, and if the conditions that caused us to disable them are no longer applicable.
> 
> At first blush, tests which fail because of the "image platform (linux/amd64) does not match the expected platform" warning could be updated to use `from`'s `--cidfile` flag to save the container ID to a temporary file, which could then be read back. The ppc64le busybox image appears to be dynamically linked, so the static binary itself can't just be copied and run.

Which ones?  Most or all tests fail with that error message.  I guess you're talking about the ones in `lists.bats`, which I dunno how to fix yet.  The others depend on images that are only available for arm64 and/or amd64 only.

nalind:
> > At first blush, tests which fail because of the "image platform (linux/amd64) does not match the expected platform" warning could be updated to use `from`'s `--cidfile` flag to save the container ID to a temporary file, which could then be read back. The ppc64le busybox image appears to be dynamically linked, so the static binary itself can't just be copied and run.
> 
> Which ones? Most or all tests fail with that error message. I guess you're talking about the ones in `lists.bats`, which I dunno how to fix yet. The others depend on images that are only available for arm64 and/or amd64 only.

Going through the logs in the description, I'm thinking these are calls to "buildah from" that are going to have to start using a `--cidfile` flag:
* blobcache-commit
* blobcache-push
* push-by-id
* from-by-digest-s1
* from-by-digest-s1-a-discarded-layer
* from-by-tag-s1
* from-by-digest-s2
* from-by-tag-s2
* from-by-repo-only-s2

---
The "build with --platform without OS" and "build with basename resolving default arg" tests looks like they need `bud/base-with-arg/Containerfile` to have a section added which looks like:
```Dockerfile
FROM build AS platform-ppc64le
ENV BUILT_FOR=ppc64le
```

---
These assume that the "quay.io/libpod/busybox" is statically linked, which is the case for the amd64 version of the image, but apparently is not for the ppc64le version:
* bud-multistage-copy-final-slash
* bud with exec-form RUN instruction

The test expects to be able to run a statically linked shell; I don't know of another image that would be suitable.  We'll need to skip these if we can't find a replacement.

---
The "bud with ADD with git repository source" test needs an image that includes a git client that it can use as a base image.  I'm not really sure what to use there, either, but if you have an idea, it would need to be listed in `tests/NEW-IMAGES` to have CI runs pre-pull it.  Otherwise, we might need to skip this one, too.

---
The "manifest-from-tag", "manifest-from-digest", and "manifest-from-instance" tests have quoting errors -- the argument values for the "--format" flag have two initial single quotes where they should have one.
The "manifest-from-tag" and "manifest-from-digest" tests are pulling the ppc64le version of pause:3.1, which, like most of the other images in the manifest list, is broken in that its configuration blob claims that its architecture is "amd64", and we trust the value in the config blob.
So after correcting the quoting, and stripping the "docker://" prefix from the image name that the three tests pass to "run_buildah inspect", all three tests end up needing to expect "amd64" instead of "$(go env GOARCH)".

Thanks!

ricardobranco777:
> Going through the logs in the description, I'm thinking these are calls to "buildah from" that are going to have to start using a `--cidfile` flag:
> 
> * blobcache-commit
> * blobcache-push
> * push-by-id

These require `registry.k8s.io/pause` which is only available for amd64.  Will add a comment for that.

> * from-by-digest-s1
> * from-by-digest-s1-a-discarded-layer
> * from-by-tag-s1
> * from-by-digest-s2
> * from-by-tag-s2
> * from-by-repo-only-s2

Aren't the test images amd64 only?

> These assume that the "quay.io/libpod/busybox" is statically linked, which is the case for the amd64 version of the image, but apparently is not for the ppc64le version:
> 
> * bud-multistage-copy-final-slash
> * bud with exec-form RUN instruction

busybox is dynamically linked on both ppc64le & s390x.

nalind:
> > Going through the logs in the description, I'm thinking these are calls to "buildah from" that are going to have to start using a `--cidfile` flag:
> > 
> > * blobcache-commit
> > * blobcache-push
> > * push-by-id
> 
> These require `registry.k8s.io/pause` which is only available for amd64. Will add a comment for that.

They don't use `buildah run`, so while it might look weird, it's expected to succeed without encountering any fatal errors.  It's the warning that adds noise to the tests, which are expecting only the container ID to be output.

> > * from-by-digest-s1
> > * from-by-digest-s1-a-discarded-layer
> > * from-by-tag-s1
> > * from-by-digest-s2
> > * from-by-tag-s2
> > * from-by-repo-only-s2
> 
> Aren't the test images amd64 only?

Yup, they all pull images that are only built for amd64, but after the warning is logged, the test should still be able to commit containers based on the image and push the result to a local directory.

> > These assume that the "quay.io/libpod/busybox" is statically linked, which is the case for the amd64 version of the image, but apparently is not for the ppc64le version:
> > 
> > * bud-multistage-copy-final-slash
> > * bud with exec-form RUN instruction
> 
> The ppc64le busybox is indeed dynamically linked.

Yeah, that's.. unfortunate for us.

ricardobranco777:
Verification runs:
- aarch64: https://openqa.opensuse.org/tests/5712033/file/buildah-buildah-root.tap.txt
- ppc64le: https://openqa.opensuse.org/tests/5712034/file/buildah-buildah-root.tap.txt
- s390x: https://openqa.suse.de/tests/21194138/file/buildah-buildah-root.tap.txt

ricardobranco777:
Patch successfully tested on 3 non-amd64 architectures.  

I couldn't find good alternative images for the busybox & git.

nalind (approved):
Thanks for running these down!  LGTM

TomSweeneyRedHat (commented):
LGTM

TomSweeneyRedHat:
@ricardobranco777 TYVM for the PR and the follow up work you did to run down the details!
/lgtm

PR #6702, opened 2026-03-02
Builder.getSecretMount(): don't leak an fd

nalind:
#### What type of PR is this?

/kind bug

Close the handle to a newly-created temporary file, since we reopen it to write to it almost immediately.

#### How to verify it

Shouldn't generate any "file already closed" errors while mounting secrets as environment variables.

None

Spotted while reviewing #6285

```release-note
None
```

### DIFF
--- a/run_common.go
+++ b/run_common.go
@@ -1852,6 +1852,7 @@ func (b *Builder) getSecretMount(tokens []string, secrets map[string]define.Secr
 		if err != nil {
 			return secretMountOrEnv{}, err
 		}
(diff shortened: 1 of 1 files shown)


TomSweeneyRedHat:
LGTM
I restarted what appeared to be a flakey F43 runc test

Luap99 (approved):
LGTM, pedantically speaking we do not leak the fd as the go garbage collector will close it once it cleans up the variable, so the leak would only be short lived but of course this is correct to close this right away

PR #6717, opened 2026-03-10
tests: Replace cat with bash input redirection

ricardobranco777:
tests: Replace cat with bash input redirection

From the bash manpage:
> The command substitution `$(cat file)` can be replaced by the equivalent but faster `$(< file)`.

This way we avoid spawning a shell process for a simple cat.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake

/kind other

#### How to verify it

http://openqa-assets.opensuse.org/tests/5734566/file/buildah-buildah-root.tap.txt

`None`

```release-note
None
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -2662,24 +2662,24 @@ _EOF
 @test "bud with --env" {
   target=scratch-image
   run_buildah build --quiet=false --iidfile ${TEST_SCRATCH_DIR}/output.iid --env PATH $WITH_POLICY_JSON -t ${target} $BUDFILES/from-scratch
-  iid=$(cat ${TEST_SCRATCH_DIR}/output.iid)
+  iid=$(< ${TEST_SCRATCH_DIR}/output.iid)
   run_buildah inspect --format '{{.Docker.Config.Env}}' $iid
   expect_output "[PATH=$PATH]"
 
   run_buildah build --quiet=false --iidfile ${TEST_SCRATCH_DIR}/output.iid --env PATH=foo $WITH_POLICY_JSON -t ${target} $BUDFILES/from-scratch
-  iid=$(cat ${TEST_SCRATCH_DIR}/output.iid)
+  iid=$(< ${TEST_SCRATCH_DIR}/output.iid)
   run_buildah inspect --format '{{.Docker.Config.Env}}' $iid
   expect_output "[PATH=foo]"
 
   # --unsetenv takes precedence over --env, since we don't know the relative order of the two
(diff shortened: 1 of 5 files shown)


danishprakash (commented):
Are the test environments guaranteed to have bash? Is this redirection supported on shells such as ash or ksh?

ricardobranco777:
> Are the test environments guaranteed to have bash? Is this redirection supported on shells such as ash or ksh?

bats needs bash :)

danishprakash:
I meant to refer to the instances that seem to modify a Containerfile, for instance:
```
from alpine
run echo "\$(< /sys/fs/cgroup/cpu/cpu.cfs_quota_us) \$(< /sys/fs/cgroup/cpu/cpu.cfs_period_us)"
```

ricardobranco777:
> I meant to refer to the instances that seem to modify a Containerfile, for instance:
> 
> ```
> from alpine
> run echo "\$(< /sys/fs/cgroup/cpu/cpu.cfs_quota_us) \$(< /sys/fs/cgroup/cpu/cpu.cfs_period_us)"
> ```

Good catch.  Fixed.  Weird that the openQA passed though.

Luap99:
> > I meant to refer to the instances that seem to modify a Containerfile, for instance:
> > ```
> > from alpine
> > run echo "\$(< /sys/fs/cgroup/cpu/cpu.cfs_quota_us) \$(< /sys/fs/cgroup/cpu/cpu.cfs_period_us)"
> > ```
> 
> Good catch. Fixed. Weird that the openQA passed though.

That is an an else branch on cgroupv2 so only running under cgroupv1 which we no longer support.

I guess looking at tests there is properly a lot of cleanup work still to do for the cgroupv1 removal, we dropped support for it already here https://github.com/containers/buildah/pull/6424 so I guess all these calls where missed?
cc @lsm5 @nalind

Luap99 (approved):
LGTM

lsm5:
> > > I meant to refer to the instances that seem to modify a Containerfile, for instance:
> > > ```
> > > from alpine
> > > run echo "\$(< /sys/fs/cgroup/cpu/cpu.cfs_quota_us) \$(< /sys/fs/cgroup/cpu/cpu.cfs_period_us)"
> > > ```
> > 
> > 
> > Good catch. Fixed. Weird that the openQA passed though.
> 
> That is an an else branch on cgroupv2 so only running under cgroupv1 which we no longer support.
> 
> I guess looking at tests there is properly a lot of cleanup work still to do for the cgroupv1 removal, we dropped support for it already here #6424 so I guess all these calls where missed? cc @lsm5 @nalind

err yup, and that's my bad. I'll do a followup PR for this.

lsm5 (approved):
LGTM

lsm5:
filed https://github.com/containers/buildah/pull/6720 for tests cleanup.

PR #6720, opened 2026-03-11
tests: remove cgroupsv1 checks and simplify cgroupsv2 conditionals

lsm5:
Since cgroupsv1 support has been removed from buildah, all is_cgroupsv2 checks are always true and skip_if_rootless_and_cgroupv1 is always a no-op. Remove these helper functions and simplify tests to use only cgroupsv2 code paths. Also simplify runc runtime-flag tests to expect success instead of tolerating old runc versions that don't support cgroupsv2.

Fixes: RUN-4432

#### What type of PR is this?

/kind cleanup

Removes cgroupsv1 conditionals since we're now cgroupsv2 only.

#### How to verify it

Apart from manual review, I guess tests should barf on cgroups v1 systems.

None

None

```release-note
None
```

### DIFF
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -6484,21 +6484,9 @@ _EOF
   local flag_accepted_rx="level=debug.*msg=.child process in init"
   if [ -n "$(command -v runc)" ]; then
     found_runtime=y
-    if is_cgroupsv2; then
-      # The result with cgroup v2 depends on the version of runc.
-      run_buildah '?' bud --runtime=runc --runtime-flag=debug \
-                        -q -t alpine-bud-runc $WITH_POLICY_JSON --file ${mytmpdir}/Containerfile .
-      if [ "$status" -eq 0 ]; then
-        expect_output --substring "$flag_accepted_rx"
-      else
-        # If it fails, this is because this version of runc doesn't support cgroup v2.
-        expect_output --substring "this version of runc doesn't work on cgroups v2" "should fail by unsupportability for cgroupv2"
-      fi
-    else
-      run_buildah build --runtime=runc --runtime-flag=debug \
-                      -q -t alpine-bud-runc $WITH_POLICY_JSON --file ${mytmpdir}/Containerfile .
-      expect_output --substring "$flag_accepted_rx"
-    fi
+    run_buildah build --runtime=runc --runtime-flag=debug \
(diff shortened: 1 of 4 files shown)


lsm5:

Luap99 (approved):
LGTM

tests/bud.bats
```diff
@@ -6484,21 +6484,9 @@ _EOF
...
-        expect_output --substring "this version of runc doesn't work on cgroups v2" "should fail by unsupportability for cgroupv2"
-      fi
-    else
-      run_buildah build --runtime=runc --runtime-flag=debug \
-                      -q -t alpine-bud-runc $WITH_POLICY_JSON --file ${mytmpdir}/Containerfile .
-      expect_output --substring "$flag_accepted_rx"
-    fi
+    run_buildah build --runtime=runc --runtime-flag=debug \
+                    -q -t alpine-bud-runc $WITH_POLICY_JSON --file ${mytmpdir}/Containerfile .
+    expect_output --substring "$flag_accepted_rx"
```
```suggestion
```

nalind (approved):
LGTM

giuseppe (approved):
LGTM

tests/bud.bats
```diff
@@ -6484,22 +6484,9 @@ _EOF
...
-        # If it fails, this is because this version of runc doesn't support cgroup v2.
-        expect_output --substring "this version of runc doesn't work on cgroups v2" "should fail by unsupportability for cgroupv2"
-      fi
-    else
-      run_buildah build --runtime=runc --runtime-flag=debug \
-                      -q -t alpine-bud-runc $WITH_POLICY_JSON --file ${mytmpdir}/Containerfile .
-      expect_output --substring "$flag_accepted_rx"
-    fi
-
+    run_buildah build --runtime=runc --runtime-flag=debug \
```
ricardobranco777, line 6487: Wait, here you're keeping the `else` branch which is `! is_cgroupsv2`, why?

ricardobranco777 (changes requested):

tests/bud.bats
nalind: It keeps the branch that isn't trying to avoid failing if the build fails with an error that suggests that the build failed because runc couldn't handle cgroups v2.

ricardobranco777: Ah right.

ricardobranco777 (approved):

PR #6730, opened 2026-03-13
Fix `COPY/ADD --from=` with `ARG` in stage scope

### LINKED ISSUE #6719: Using a default `ARG` as a parameter for `COPY --from=` doesn't use previously defined stage
### Issue Description

Using a default `ARG` as a parameter for `COPY --from=` doesn't use previously defined stage and instead asks to select a registry to retrieve the image from.

Use the following `Containerfile`:

```docker
FROM docker.io/library/busybox:latest AS build
RUN echo "build" >> /variant

FROM build AS dev
RUN echo "build:dev" >> /variant

FROM build AS staging
RUN echo "build:staging" >> /variant

FROM docker.io/library/busybox:latest
ARG VARIANT="staging"
COPY --from="${VARIANT}" /variant /variant
RUN echo "${VARIANT}" >> /variant
```

Run `podman build --no-cache --tag variant-test "$PWD" && podman run --rm variant-test cat /variant`.

Podman asks to select a registry to retrieve the `staging` image from.

The image should be built with the previously declared `staging` stage as when `--build-arg` is specified.

### podman info output

```yaml
Client:       Podman Engine
Version:      5.6.0
API Version:  5.6.0
Go Version:   go1.25.7 (Red Hat 1.25.7-1.el9_7)
Built:        Fri Feb 20 18:49:57 2026
OS/Arch:      linux/amd64

host:
  arch: amd64
  buildahVersion: 1.41.8
```

### Podman in a container

No

### Privileged Or Rootless

Rootless

### Upstream Latest Release

No

RHEL 9.7 on bare metal.

### Additional information

Using the same `Containerfile`, the build works as expected when specifying `--build-arg` like the following:

* `podman build --no-cache --tag variant-test --build-arg VARIANT="dev" "$PWD" && podman run --rm variant-test cat /variant`
* `podman build --no-cache --tag variant-test --build-arg VARIANT="staging" "$PWD" && podman run --rm variant-test cat /variant`

Honny1:
Fixes: https://github.com/containers/buildah/issues/6719

#### What type of PR is this?

> /kind api-change
> 
/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
Fixes `COPY/ADD --from=` when the source stage is defined via a stage scoped `ARG`, ensuring correct source resolution.
```

### DIFF
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -879,6 +879,7 @@ func (b *executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 	// not they can skip certain steps near the end of their stages.
 	for stageIndex, stage := range stages {
 		dependencyMap[stage.Name] = &stageDependencyInfo{Name: stage.Name, Position: stage.Position}
+		stageLocalScopeArgs := make(map[string]string)
 		node := stage.Node // first line
 		for node != nil {  // each line
 			for _, child := range node.Children { // tokens on this line, though we only care about the first
@@ -902,10 +903,11 @@ func (b *executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 							builtinArgs := argsMapToSlice(stage.Builder.BuiltinArgDefaults)
 							headingArgs := argsMapToSlice(stage.Builder.HeadingArgs)
 							userArgs := argsMapToSlice(stage.Builder.Args)
+							localScopeArgs := argsMapToSlice(stageLocalScopeArgs)
 							// append heading args so if --build-arg key=value is not
 							// specified but default value is set in Containerfile
 							// via `ARG key=value` so default value can be used.
-							userArgs = slices.Concat(builtinArgs, userArgs, headingArgs)
+							userArgs = slices.Concat(builtinArgs, userArgs, headingArgs, localScopeArgs)
 							baseWithArg, err := imagebuilder.ProcessWord(base, userArgs)
 							if err != nil {
 								return "", nil, fmt.Errorf("while replacing arg variables with values for format %q: %w", base, err)
@@ -936,7 +938,8 @@ func (b *executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 							builtinArgs := argsMapToSlice(stage.Builder.BuiltinArgDefaults)
 							headingArgs := argsMapToSlice(stage.Builder.HeadingArgs)
 							userArgs := argsMapToSlice(stage.Builder.Args)
-							userArgs = slices.Concat(builtinArgs, userArgs, headingArgs)
+							localScopeArgs := argsMapToSlice(stageLocalScopeArgs)
+							userArgs = slices.Concat(builtinArgs, userArgs, headingArgs, localScopeArgs)
 							afterResolved, err := imagebuilder.ProcessWord(after, userArgs)
 							if err != nil {
 								return "", nil, fmt.Errorf("while replacing arg variables with values for --after=%q: %w", after, err)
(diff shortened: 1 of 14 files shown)


imagebuildah/executor.go
nalind: Are these lines and the preceding comment still correct and/or necessary given the addition being made below?

imagebuildah/executor.go
```diff
@@ -854,10 +855,11 @@ func (b *executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 							builtinArgs := argsMapToSlice(stage.Builder.BuiltinArgDefaults)
 							headingArgs := argsMapToSlice(stage.Builder.HeadingArgs)
 							userArgs := argsMapToSlice(stage.Builder.Args)
+							localScopeArgs := argsMapToSlice(stageLocalScopeArgs)
 							// append heading args so if --build-arg key=value is not
 							// specified but default value is set in Containerfile
 							// via `ARG key=value` so default value can be used.
-							userArgs = slices.Concat(builtinArgs, userArgs, headingArgs)
+							userArgs = slices.Concat(builtinArgs, userArgs, headingArgs, localScopeArgs)
```
nalind: How does this handle conflicting values, when one should override another?  For example, what happens when `tests/bud/copy-from-stage-scoped-arg/Containerfile.multi-copy-arg-override` has `ARG VARIANT=dev` added to its header, before the "build" stage?

imagebuildah/executor.go
```diff
@@ -953,6 +960,14 @@ func (b *executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 							}
 						}
 					}
+				case "ARG":
+					arg := child.Next
+					if arg != nil {
+						argName, argValue, hasValue := strings.Cut(arg.Value, "=")
+						if hasValue && argName != "" {
+							stageLocalScopeArgs[argName] = strings.Trim(argValue, "\"'")
```
nalind: Probably want to use `imagebuilder.ProcessWord()` to handle the quoting here.

nalind (changes requested):
Generally looks good, but questions about handling precedence when an ARG is provided a default value both in the header and in a stage and how the ARG value is de-quoted.

imagebuildah/executor.go
nalind: I was just reviewing another PR, and was reminded by https://github.com/openshift/imagebuilder/pull/210 that there might be multiple ARGs defined on this line.

Honny1: I check other parts of buildah and add support for that into buildah, including tests.

imagebuildah/executor.go
Honny1: Good catch, I reorder concat and also add tests for this.

imagebuildah/executor.go
```diff
@@ -850,10 +855,10 @@ func (b *executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 							builtinArgs := argsMapToSlice(stage.Builder.BuiltinArgDefaults)
 							headingArgs := argsMapToSlice(stage.Builder.HeadingArgs)
 							userArgs := argsMapToSlice(stage.Builder.Args)
-							// append heading args so if --build-arg key=value is not
-							// specified but default value is set in Containerfile
-							// via `ARG key=value` so default value can be used.
-							userArgs = slices.Concat(builtinArgs, userArgs, headingArgs)
+							localScopeArgs := argsMapToSlice(stageLocalScopeArgs)
+							// ProcessWord uses first match; put highest priority first so
+							// stage ARG overrides --build-arg overrides header ARG overrides builtin.
```
nalind: The contents of the file overriding values passed in from the command line doesn't sound right.

TomSweeneyRedHat:
@Honny1 Touch up the comment from @nalind, and I think this is GTG
LGTM
other than that

nalind (approved):
LGTM, thanks!

PR #6736, opened 2026-03-19
internal/mkcw/embed/entrypoint_amd64.gz: rebuild with native assembler

lsm5:
The earlier committed .gz was built via the Go fallback path (~370KB). Rebuild it using the native assembler path (~365 bytes).

#### What type of PR is this?

/kind other

This solves a frequent annoyance seen after running `make` and committing to git.

None

```release-note
None
```


lsm5:

giuseppe (approved):
LGTM

mtrmac (commented):
How do we prevent this from happening again, any random following week?

I suppose the upstream codebase should, _in general_, support the Go build path as an option for users, but for our upstream code (if we want to commit that build output to git at all, that is), would we consider a validation / linting step that refuses PRs which revert this again?

(I also worry about having a binary committed out of sync with the sources, but per https://github.com/containers/buildah/pull/6471#issuecomment-3492347460 , that seems still necessary.)

lsm5:
How about this Makefile addition and maybe add that to the `validate` CI job:

```
.PHONY: lint-entrypoint
lint-entrypoint:
        @tmp=$$(mktemp) ; \
        zcat internal/mkcw/embed/entrypoint_amd64.gz > "$$tmp" ; \
        if readelf -S "$$tmp" | grep -qE '\.go\.|gopclntab'; then \
                echo "FAIL: entrypoint_amd64.gz was built using Go fallback path" ; \
                rm -f "$$tmp" ; \
                exit 1 ; \
        fi ; \
        echo "OK: entrypoint_amd64.gz was built using native assembler" ; \
        rm -f "$$tmp"
```

I could add another commit here if there's agreement.

Luap99:
I am still of the opinion that shipping a binary artifact in the source is wrong to begin with and likely violates all sort of distro policies with us bundling that.

And as history shows review is not catching all these modifications so considering the fact that anyone can replace this binary with content of unkown origin we are inviting problems.

I am in favour of such a "lint" step for now though long term I think shipping binaries in the source tree is always incorrect and we should try to remove that dep.

lsm5:

mtrmac (commented):
ACK to the PR approach in general, but I’ll leave the ultimate decision to Buildah maintainers.

(For the record: I did not verify the contents of the updated binary.)

lsm5:
@nalind PTAL

lsm5:
@nalind PTAL, wdyt?

nalind:
@lsm5 I don't have a strong opinion on this one way or another.  Sure.

lsm5:
> @lsm5 I don't have a strong opinion on this one way or another. Sure.

@nalind ack, PTAL for review and (hopefully) merge. This should fix the recurring annoyance (at least) I notice.

nalind:
We need other cleanups in this Makefile, but this'll do for now.

PR #3241, opened 2021-05-19
Don't blow up if cpp detects errors

rhatdan:
Currently the /usr/bin/cpp will blow up if a user adds a
comment to a containerfile that is not a preprocessor.
Since the Containerfile.in could include other Containerfile
which may have comments, begining with `#` this can cause
problems.

If we just warn on these errors, we can successfully process
all of the containerfiles.

Fixes: https://github.com/containers/buildah/issues/3229

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### Does this PR introduce a user-facing change?

```release-note

```

PR #6755, opened 2026-04-01
feat: add generic Containerfile template processing

rockdrilla:
#### What type of PR is this?

/kind feature

Add support for preprocessing Containerfile/Dockerfile files with configurable suffixes and preprocessors. Files with registered suffixes (e.g., .in, .tmpl) are processed by either the built-in C preprocessor (cpp) or external commands before the build.

CLI changes:

- Add `--template` flag for `suffix=preprocessor[,options]` specifications
- Add `--template-lookup` flag to control search order (`first`/`last`/`never`)

Environment variables:

- `BUILDAH_TEMPLATE`: configure templates via semicolon-separated list or JSON
- `BUILDAH_TEMPLATE_LOOKUP`: set default lookup policy

Template options:

- `ignore_stderr`: suppress stderr output unless command fails
- `chdir_ctxdir`: change working directory to build context

The build context directory is always exposed to preprocessors via the `BUILDAH_CTXDIR` environment variable.

This enables features like:
- Decomposing Containerfiles using CPP(1) `#include` directives (.in files)
- Using custom template engines (jinja2, m4, etc.) for dynamic Containerfile generation
- Conditional builds based on environment variables or external configuration

#### How to verify it

1. Default .in file processing with built-in cpp:

```sh
buildah build -f Containerfile.in .
```

2. Custom template processor with lookup policy:

```sh
buildah build --template j2=jinja2 --template-lookup=first .
```

3. Using environment variables:

```sh
export BUILDAH_TEMPLATE="in=- ; j2=jinja2,ignore_stderr"
export BUILDAH_TEMPLATE_LOOKUP="last"
buildah build .
```

4. JSON configuration:

```sh
export BUILDAH_TEMPLATE='[{"suffix":"in","builtin_cpp":true},{"suffix":"tmpl","cmd":["m4"],"ignore_stderr":true}]'
buildah build .
```

5. Remote git repository with file discovery:

```sh
buildah build --template-lookup=first https://github.com/rockdrilla/fork.PodmanHello_in.git
```

Note: the mentioned repository is a fork of [PodmanHello](https://github.com/containers/PodmanHello.git) with `Containerfile` renamed to `Containerfile.in` to demonstrate template processing.

None

I was unable to run the full test suite locally due to rootless/rootful Podman setup issues on my development machine. The code has been manually tested with various scenarios, but I'd appreciate guidance from maintainers on adding automated tests for this feature.

The implementation includes:

- Backward compatible: existing behavior unchanged when no template options specified

- Flexible configuration: supports CLI flags, environment variables, and multiple formats

- Comprehensive documentation: man page updated with examples

```release-note
Add support for Containerfile/Dockerfile template preprocessing:
- New `--template` flag to define suffix-based preprocessors
- New `--template-lookup` flag to control template search order
- New `BUILDAH_TEMPLATE` environment variable for template configuration
- New `BUILDAH_TEMPLATE_LOOKUP` environment variable for default lookup policy
- Template preprocessors receive build context via `BUILDAH_CTXDIR` environment variable
- Support for built-in C preprocessor (cpp) and custom external commands
- Options: `ignore_stderr` and `chdir_ctxdir` for template preprocessing
```

### DIFF
--- a/define/build.go
+++ b/define/build.go
@@ -263,6 +263,9 @@ type BuildOptions struct {
 	DropCapabilities []string
 	// CommonBuildOpts is *required*.
 	CommonBuildOpts *CommonBuildOptions
+	// TemplateOpts configures template preprocessing for Containerfile/Dockerfile files
+	// with specific suffixes (e.g., .in, .tmpl). See TemplateOptions for details.
+	TemplateOpts *TemplateOptions
 	// CPPFlags are additional arguments to pass to the C Preprocessor (cpp).
 	CPPFlags []string
 	// DefaultMountsFilePath is the file path holding the mounts to be mounted for RUN
--- a/define/types.go
+++ b/define/types.go
@@ -12,6 +12,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	v1 "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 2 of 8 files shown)


rockdrilla:
I've fixed linter warnings and also template preprocessing with builtin cpp support in order to align with #3241.

nalind:
This provides arbitrary command execution in the context of the process that's running the build.  That's a problem for library consumers which accept build requests across a privilege boundary (e.g., podman) and expose these flags.
I don't think I'm really comfortable with encouraging that.

rockdrilla:
Same applies to existing approach: the one may create script named `cpp` in empty directory, prepend that directory to PATH variable and achieve almost same behavior as I've written except described method is abuse rather than intended usage.

Using templates in Dockerfiles is a good thing these days; however, `cpp(1)` lacks the flexibility of other templating engines like jinja2, gomplate and others.

rockdrilla:
Especially, the one may encounter problems while templating Dockerfile with `cpp(1)`: preprocessor directive conflicts with comment in Dockerfile because both start with `#`, and this clash is already poisoning codebase - `cpp(1)` exit code is ignored unless stdout is empty.

nalind:
Oh, to be clear, I'm not a fan of the existing preprocessing functionality.  Some of the reasons for that are the problems you're describing.  But I'm only one of multiple maintainers.

solacelost:
I, too, dislike using `cpp` and would much rather a clean `go-template` implementation - maybe with some `sprig` so people who write Helm charts (and are therefore comfortable in the container ecosystem) have something familiar to use?

I thought about proposing something more expansive than fixing up my current nit with the `cpp` implementation, but figured I'd shoot for a minor fix rather than introduce what's a pretty major feature, lol. I think it would be sufficiently disruptive - and appreciated - to turn a lot of heads, though, if the security implications of running a more robust templating engine could be worked through.

ETA: I may end up just settling on using a more robust templating engine in my CI pipeline as a preprocessing step before building, as this would give me sufficient flexibility without requiring any kind of change to `buildah` behavior at all. It's just kind of annoying for me, having settled on trying to get Konflux to be my CI engine. I will probably yolo custom PAC instead if I go this route.

Luap99:
Personally I do not believe these things should be part of buildah.

What exactly is the use case that requires this to be part in our codebase? Having these cli/env options seem confusing and hard to document/understand for users.
For template engines it seem much cleaner to just call that template command to create the final Containerfile before passing that then to the build command.
That is something any normal user can reason about and debugging wise it should be obvious which part failed then compared to doing it all in buildah.

The only special env that buildah is adding is the context dir and well you have that very much when calling buildah on the cli so it is trival to pass that to the template command just before as well so I see no value add overall.

rockdrilla:
I'm currently working on this PR, so stay tuned.

PR #6757, opened 2026-04-01
Respect `compression_format` from `containers.conf` in push, build `--cache-to`, and commit

### LINKED ISSUE #6072: cache compression algorithm
Is there a way to specify how to compress the cached layers when doing `buildah build`, of course with `--layers` ?

It seems like the cache has `MIMEType` `application/vnd.oci.image.layer.v1.tar+gzip`

### LINKED ISSUE #6660: --cache-to causes compression_format from containers.conf to be ignored
## Description

When using `podman build --cache-to`, the `compression_format` setting from `containers.conf` is ignored for both the cache layers AND the final image. Without `--cache-to`, the compression format is respected correctly.

## Steps to Reproduce

1. Create a minimal `containers.conf` with zstd:chunked:

```ini
[engine]
compression_format="zstd:chunked"
```

2. Create a minimal Dockerfile:

```dockerfile
FROM alpine:latest
RUN echo "layer1" > /layer1.txt
RUN echo "layer2" > /layer2.txt
```

3. Build and push WITHOUT `--cache-to`:

```bash
podman build --no-cache --layers -t quay.io/jdanek/cache-compression-test:v8 .
podman push quay.io/jdanek/cache-compression-test:v8
crane manifest quay.io/jdanek/cache-compression-test:v8 | jq -r '.layers[].mediaType'
```

**Result**: `application/vnd.oci.image.layer.v1.tar+zstd` ✓ (correct)

4. Build and push WITH `--cache-to`:

```bash
podman system reset -f
podman build --no-cache --layers --cache-to quay.io/jdanek/cache-compression-test-cache-v6 -t quay.io/jdanek/cache-compression-test:v7 .
podman push quay.io/jdanek/cache-compression-test:v7
crane manifest quay.io/jdanek/cache-compression-test:v7 | jq -r '.layers[].mediaType'
```

**Result**: `application/vnd.oci.image.layer.v1.tar+gzip` ✗ (should be zstd)

The cache also uses gzip:

```bash
crane manifest quay.io/jdanek/cache-compression-test-cache-v6:077cd4b1a9a353fc2e41beb42b498264aef7287aed9675da9899238a4ddec6cc | jq -r '.layers[].mediaType'
```

**Result**: `application/vnd.oci.image.layer.v1.tar+gzip`

## Expected Behavior

Both cache and final image should use `zstd:chunked` compression as specified in `containers.conf`, regardless of whether `--cache-to` is used.

## Actual Behavior

| Scenario | Cache | Final Image |
|----------|-------|-------------|
| Build with `--layers` (no cache-to) | N/A | **zstd** ✓ |
| Build with `--layers --cache-to` | **gzip** | **gzip** |

The `--cache-to` flag causes the entire `compression_format` setting to be ignored.

## Environment

```
podman version 5.7.1
macOS with podman machine (Fedora CoreOS VM)
```

## containers.conf location

`/etc/containers/containers.conf` in the podman machine VM

## Related Issues

- #4613 - Buildah reuse gzip layers when "--compression-format zstd" assigned (fixed)
- https://github.com/containers/podman/issues/26088 - podman build removes layer compression when layering on compressed images

## Test Images

The test images are publicly available for verification:
- Cache: `quay.io/jdanek/cache-compression-test-cache-v6`
- Final (with cache-to, gzip): `quay.io/jdanek/cache-compression-test:v7`
- Final (without cache-to, zstd): `quay.io/jdanek/cache-compression-test:v8`

Honny1:
- `buildah push` now falls back to `compression_format` and `compression_level` from `containers.conf` when `--compression-format` is not set
- `buildah build --cache-to` passes `CompressionFormat/ForceCompressionFormat` to cache push, so cached layers use the configured compression instead of defaulting to `gzip`
- `buildah build` gains `--cache-compression-format`, `--cache-compression-level`, and `--cache-force-compression` flags for controlling cache layer compression independently
- `buildah commit` gains `--compression-format`, `--compression-level`, and `--force-compression` flags with `containers.conf` fallback
- Fix `commit.go` switch to set `DirForceCompress` for all compression algorithms, not just `gzip`
- Fix `push.go` `blobcache` to treat any non-uncompressed format as requiring compression
- Add conflicting flag validation for `--disable-compression` with `--compression-format`/`--force-compression`

Fixes: https://github.com/containers/buildah/issues/6660 
Fixes: https://github.com/containers/buildah/issues/6072

Depends on: https://github.com/containers/container-libs/pull/731

#### What type of PR is this?

> /kind api-change
> 
/kind bug
> 
/kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
`push`, build`, `commit`: respect `compression_format` from `containers.conf`

`buildah push`, `buildah build --cache-to`, and `buildah commit` now honor the `compression_format` and `compression_level` settings from `containers.conf`. Previously, `--cache-to` ignored these settings and always used `gzip`, causing cache layers to lose the configured compression. `buildah build` also gains new `--cache-compression-format`, `--cache-compression-level`, and `--cache-force-compression` flags for controlling cache layer compression independently of the final image. `buildah commit` also gains new `--compression-format`, `--compression-level`, and `--force-compression` flags.
```

### DIFF
--- a/cmd/buildah/commit.go
+++ b/cmd/buildah/commit.go
@@ -19,7 +19,9 @@ import (
 	"go.podman.io/buildah/util"
 	"go.podman.io/common/pkg/auth"
 	"go.podman.io/common/pkg/completion"
+	"go.podman.io/common/pkg/config"
 	"go.podman.io/image/v5/manifest"
+	"go.podman.io/image/v5/pkg/compression"
 	"go.podman.io/image/v5/pkg/shortnames"
 	storageTransport "go.podman.io/image/v5/storage"
 	"go.podman.io/image/v5/transports/alltransports"
@@ -27,49 +29,52 @@ import (
 )
 
 type commitInputOptions struct {
-	authfile           string
-	omitHistory        bool
-	blobCache          string
-	certDir            string
-	changes            []string
-	configFile         string
-	creds              string
-	cwOptions          string
-	disableCompression bool
-	format             string
-	iidfile            string
-	manifest           string
-	omitTimestamp      bool
-	timestamp          int64
-	sourceDateEpoch    string
-	rewriteTimestamp   bool
-	quiet              bool
-	referenceTime      string
-	rm                 bool
-	pull               string
-	pullAlways         bool
-	pullNever          bool
-	sbomImgOutput      string
-	sbomImgPurlOutput  string
-	sbomMergeStrategy  string
-	sbomOutput         string
-	sbomPreset         string
-	sbomPurlOutput     string
-	sbomScannerCommand []string
-	sbomScannerImage   string
-	signaturePolicy    string
-	signBy             string
-	squash             bool
-	tlsVerify          bool
-	identityLabel      bool
-	encryptionKeys     []string
-	encryptLayers      []int
-	unsetenvs          []string
-	addFile            []string
-	unsetAnnotation    []string
-	annotation         []string
-	createdAnnotation  bool
-	metadataFile       string
+	authfile               string
+	omitHistory            bool
+	blobCache              string
+	certDir                string
+	changes                []string
+	compressionFormat      string
+	compressionLevel       int
+	configFile             string
+	creds                  string
+	cwOptions              string
+	disableCompression     bool
+	forceCompressionFormat bool
+	format                 string
+	iidfile                string
+	manifest               string
+	omitTimestamp          bool
+	timestamp              int64
+	sourceDateEpoch        string
+	rewriteTimestamp       bool
+	quiet                  bool
+	referenceTime          string
+	rm                     bool
+	pull                   string
+	pullAlways             bool
+	pullNever              bool
+	sbomImgOutput          string
+	sbomImgPurlOutput      string
+	sbomMergeStrategy      string
+	sbomOutput             string
+	sbomPreset             string
+	sbomPurlOutput         string
+	sbomScannerCommand     []string
+	sbomScannerImage       string
+	signaturePolicy        string
+	signBy                 string
+	squash                 bool
+	tlsVerify              bool
+	identityLabel          bool
+	encryptionKeys         []string
+	encryptLayers          []int
+	unsetenvs              []string
+	addFile                []string
+	unsetAnnotation        []string
+	annotation             []string
+	createdAnnotation      bool
+	metadataFile           string
 }
 
 func commitInit() {
@@ -117,8 +122,13 @@ func commitListFlagSet(cmd *cobra.Command, opts *commitInputOptions) {
 	_ = cmd.RegisterFlagCompletionFunc("config", completion.AutocompleteDefault)
 	flags.StringVar(&opts.creds, "creds", "", "use `[username[:password]]` for accessing the registry")
 	_ = cmd.RegisterFlagCompletionFunc("creds", completion.AutocompleteNone)
+	flags.StringVar(&opts.compressionFormat, "compression-format", "", "compression format to use")
+	_ = cmd.RegisterFlagCompletionFunc("compression-format", completion.AutocompleteNone)
+	flags.IntVar(&opts.compressionLevel, "compression-level", 0, "compression level to use")
+	_ = cmd.RegisterFlagCompletionFunc("compression-level", completion.AutocompleteNone)
 	flags.StringVar(&opts.cwOptions, "cw", "", "confidential workload `options`")
 	flags.BoolVarP(&opts.disableCompression, "disable-compression", "D", true, "don't compress layers")
+	flags.BoolVar(&opts.forceCompressionFormat, "force-compression", false, "use the specified compression algorithm if the destination contains a differently-compressed variant already")
 	flags.StringVarP(&opts.format, "format", "f", defaultFormat(), "`format` of the image manifest and metadata")
 	_ = cmd.RegisterFlagCompletionFunc("format", completion.AutocompleteNone)
 	flags.StringVar(&opts.manifest, "manifest", "", "adds created image to the specified manifest list. Creates manifest list if it does not exist")
@@ -228,6 +238,15 @@ func commitCmd(c *cobra.Command, args []string, iopts commitInputOptions) error 
 		compress = define.Uncompressed
 	}
 
+	if c.Flag("disable-compression").Changed && iopts.disableCompression {
+		if c.Flag("compression-format").Changed {
+			return errors.New("--disable-compression and --compression-format cannot be used together")
+		}
+		if c.Flag("force-compression").Changed {
+			return errors.New("--disable-compression and --force-compression cannot be used together")
+		}
+	}
+
 	format, err := cli.GetFormat(iopts.format)
 	if err != nil {
 		return err
@@ -308,25 +327,54 @@ func commitCmd(c *cobra.Command, args []string, iopts commitInputOptions) error 
 	}
 
 	options := buildah.CommitOptions{
-		PreferredManifestType: format,
-		Manifest:              iopts.manifest,
-		Compression:           compress,
-		SignaturePolicyPath:   iopts.signaturePolicy,
-		SystemContext:         systemContext,
-		IIDFile:               iopts.iidfile,
-		Squash:                iopts.squash,
-		BlobDirectory:         iopts.blobCache,
-		OmitHistory:           iopts.omitHistory,
-		SignBy:                iopts.signBy,
-		OciEncryptConfig:      encConfig,
-		OciEncryptLayers:      encLayers,
-		UnsetEnvs:             iopts.unsetenvs,
-		OverrideChanges:       iopts.changes,
-		OverrideConfig:        overrideConfig,
-		ExtraImageContent:     addFiles,
-		UnsetAnnotations:      iopts.unsetAnnotation,
-		Annotations:           iopts.annotation,
-		CreatedAnnotation:     types.NewOptionalBool(iopts.createdAnnotation),
+		PreferredManifestType:  format,
+		Manifest:               iopts.manifest,
+		Compression:            compress,
+		ForceCompressionFormat: iopts.forceCompressionFormat,
+		SignaturePolicyPath:    iopts.signaturePolicy,
+		SystemContext:          systemContext,
+		IIDFile:                iopts.iidfile,
+		Squash:                 iopts.squash,
+		BlobDirectory:          iopts.blobCache,
+		OmitHistory:            iopts.omitHistory,
+		SignBy:                 iopts.signBy,
+		OciEncryptConfig:       encConfig,
+		OciEncryptLayers:       encLayers,
+		UnsetEnvs:              iopts.unsetenvs,
+		OverrideChanges:        iopts.changes,
+		OverrideConfig:         overrideConfig,
+		ExtraImageContent:      addFiles,
+		UnsetAnnotations:       iopts.unsetAnnotation,
+		Annotations:            iopts.annotation,
+		CreatedAnnotation:      types.NewOptionalBool(iopts.createdAnnotation),
+	}
+	defaultContainerConfig, err := config.Default()
+	if err != nil {
+		return fmt.Errorf("failed to get container config: %w", err)
+	}
+	if iopts.compressionFormat != "" {
+		algo, err := compression.AlgorithmByName(iopts.compressionFormat)
+		if err != nil {
+			return err
+		}
+		options.CompressionFormat = &algo
+		if !c.Flag("force-compression").Changed {
+			options.ForceCompressionFormat = true
+		}
+	} else if defaultContainerConfig.Engine.CompressionFormat != "" && defaultContainerConfig.Engine.CompressionFormat != "gzip" {
+		algo, err := compression.AlgorithmByName(defaultContainerConfig.Engine.CompressionFormat)
+		if err != nil {
+			return fmt.Errorf("parsing compression_format from containers.conf: %w", err)
+		}
+		options.CompressionFormat = &algo
+		if !c.Flag("force-compression").Changed {
+			options.ForceCompressionFormat = true
+		}
+	}
+	if c.Flag("compression-level").Changed {
+		options.CompressionLevel = &iopts.compressionLevel
+	} else {
+		options.CompressionLevel = defaultContainerConfig.Engine.CompressionLevel
 	}
 	exclusiveFlags := 0
 	if c.Flag("reference-time").Changed {
--- a/cmd/buildah/push.go
+++ b/cmd/buildah/push.go
@@ -16,6 +16,7 @@ import (
 	"go.podman.io/buildah/pkg/parse"
 	util "go.podman.io/buildah/util"
 	"go.podman.io/common/pkg/auth"
+	"go.podman.io/common/pkg/config"
 	"go.podman.io/image/v5/manifest"
 	"go.podman.io/image/v5/pkg/compression"
 	"go.podman.io/image/v5/transports"
@@ -194,14 +195,6 @@ func pushCmd(c *cobra.Command, args []string, iopts pushOptions) error {
 		return fmt.Errorf("unable to obtain encryption config: %w", err)
 	}
 
-	if c.Flag("compression-format").Changed {
-		if !c.Flag("force-compression").Changed {
-			// If `compression-format` is set and no value for `--force-compression`
-			// is selected then defaults to `true`.
-			iopts.forceCompressionFormat = true
-		}
-	}
-
 	options := buildah.PushOptions{
 		Compression:            compress,
 		ManifestType:           manifestType,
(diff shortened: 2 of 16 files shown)


Honny1:
While fixing the build cache, I decided to also check the pull and commit where compression is used.

Honny1:
Todo:

- [ ] vendor container-libs with updated blobcache + adaptation to change.
- [ ] Solve CI failure

Honny1:

pkg/cli/build.go
```diff
@@ -236,6 +238,23 @@ func GenBuildOptions(c *cobra.Command, inputArgs []string, iopts BuildOptions) (
...
+	if defaultContainerConfig.Engine.CompressionFormat != "" && defaultContainerConfig.Engine.CompressionFormat != "gzip" {
+		algo, err := imgCompression.AlgorithmByName(defaultContainerConfig.Engine.CompressionFormat)
+		if err != nil {
+			return options, nil, nil, fmt.Errorf("parsing compression_format from containers.conf: %w", err)
+		}
+		compressionFormat = &algo
+	}
+	if defaultContainerConfig.Engine.CompressionLevel != nil {
+		compressionLevel = defaultContainerConfig.Engine.CompressionLevel
+	}
```
simonbrauner, line 256:
```suggestion
	compressionLevel = defaultContainerConfig.Engine.CompressionLevel
```

Definitely non-blocking, it just caught my eye, but the `if` is redundant, right? `compressionLevel` would be `nil` on declaration, so assigning `nil` to it again would not be a problem.

simonbrauner (approved):

simonbrauner:
LGTM

cmd/buildah/commit.go
```diff
@@ -307,25 +332,51 @@ func commitCmd(c *cobra.Command, args []string, iopts commitInputOptions) error
...
+	defaultContainerConfig, err := config.Default()
+	if err != nil {
+		return fmt.Errorf("failed to get container config: %w", err)
+	}
+	if iopts.compressionFormat != "" {
+		algo, err := compression.AlgorithmByName(iopts.compressionFormat)
+		if err != nil {
+			return err
+		}
+		options.CompressionFormat = &algo
```
nalind, line 365: Was the bit that checks for `c.Flag("compression-format").Changed` above intended to be merged into this block, to parallel the similar logic in the next block?

cmd/buildah/push.go
```diff
@@ -224,15 +225,30 @@ func pushCmd(c *cobra.Command, args []string, iopts pushOptions) error {
...
+	defaultContainerConfig, err := config.Default()
+	if err != nil {
+		return fmt.Errorf("failed to get container config: %w", err)
+	}
 	if iopts.compressionFormat != "" {
 		algo, err := compression.AlgorithmByName(iopts.compressionFormat)
 		if err != nil {
 			return err
 		}
 		options.CompressionFormat = &algo
```
nalind, line 237: Was the bit that checks for `c.Flag("compression-format").Changed` above intended to be merged into this block, to parallel the similar logic in the next block?

nalind (commented):
What's the intent behind not adding to "build" the flags that are being added to "push" and "commit"?

cmd/buildah/commit.go
Honny1: Ah, that is much cleaner. I will fix that.

Honny1:
> What's the intent behind not adding to "build" the flags that are being added to "push" and "commit"?

I forgot to do that. Good catch.

Honny1:
@nalind I addressed your comments.

Honny1:

Honny1:
PTAL @containers/podman-maintainers

giuseppe (approved):
LGTM

tests/commit.bats
```diff
@@ -642,3 +642,78 @@ load helpers
...
+  mkdir -p $destdir
+
+  run_buildah commit \
+    $WITH_POLICY_JSON \
+    --compression-format zstd \
+    --disable-compression=false \
+    $cid \
+    dir:$destdir
+
+  run cat $destdir/manifest.json
```
nalind: This should check "$status", just in case the "dir:" format changes, however unlikely that is.

tests/commit.bats
```diff
@@ -642,3 +642,78 @@ load helpers
...
+      --disable-compression=false \
+      $cid \
+      dir:$destdir
+
+  run cat $destdir/manifest.json
+  expect_output --substring "zstd" \
+    "manifest should reference zstd-compressed layers per containers.conf"
+  assert "$output" !~ "gzip" \
+    "manifest should NOT reference gzip layers when containers.conf specifies zstd"
+}
```
nalind, line 723: What happens if we specify compression when committing to a destination in local storage, or to an OCI archive?

pkg/cli/common.go
```diff
@@ -250,6 +253,9 @@ func GetBudFlags(flags *BudResults) pflag.FlagSet {
 	fs.BoolVar(&flags.CreatedAnnotation, "created-annotation", true, `set an "org.opencontainers.image.created" annotation in the image`)
 	fs.StringVar(&flags.Creds, "creds", "", "use `[username[:password]]` for accessing the registry")
 	fs.StringVarP(&flags.CWOptions, "cw", "", "", "confidential workload `options`")
+	fs.StringVar(&flags.CacheCompressionFormat, "cache-compression-format", "", "compression format to use for cache layers")
+	fs.IntVar(&flags.CacheCompressionLevel, "cache-compression-level", 0, "compression level to use for cache layers")
+	fs.BoolVar(&flags.CacheForceCompressionFormat, "cache-force-compression", false, "use the specified compression algorithm for cache layers if the destination contains a differently-compressed variant already")
```
nalind: `build` can write the final image to non-local destinations, as `commit` does.  Could these flags also affect those cases?

nalind (changes requested):

pkg/cli/common.go
Honny1: These flags are intentionally cache-only (`--cache-to`). The final image commit already respects the `compression_format` in `containers.conf`, and `buildah push` has its own `--compression-format` flag for the actual push. Should I clarify this in the docs, or implement flags that affect the entire build?

tests/commit.bats
Honny1: Compression should work for `dir`, `oci`, and `oci-archive` destination transports. I have added a test. Local storage will accept an OCI manifest but will decompress all layers.

Honny1:
PTAL @nalind

Honny1:
PTAL @containers/buildah-maintainers @containers/podman-maintainers @mheon

pkg/cli/common.go
nalind: Adding CLI flags to "commit" for overriding the defaults from the configuration file when committing the image, and having flags for "build" that do the same thing for cache images, but no equivalent for "build" for the final image just looks like an oversight.  Most options for "commit" should also be available for "build".

Honny1: I have added support for `--compression-format`, `--compression-level`, and `--force-compression` to apply to the final image during `buildah build`, not just cache layers. These take effect when the output is a non-local destination (`registry`, `dir:`, `oci-archive:`, etc.).

Honny1:
PTAL @containers/buildah-maintainers @containers/podman-maintainers @nalind

imagebuildah/stage_executor.go
```diff
@@ -2702,6 +2705,11 @@ func (s *stageExecutor) commit(ctx context.Context, createdBy string, emptyLayer
 	if finalInstruction {
 		options.ConfidentialWorkloadOptions = s.executor.confidentialWorkload
 		options.SBOMScanOptions = s.executor.sbomScanOptions
+		if imageRef != nil && imageRef.Transport().Name() != is.Transport.Name() {
```
nalind, line 2739: Likely want a godoc for why this check is here.

tests/bud.bats
```diff
@@ -6474,6 +6474,218 @@ _EOF
...
+  _setup_compression_test
+  cat > $contextdir/containers.conf << _EOF
+[engine]
+compression_format="zstd"
+_EOF
+  local imgname="img-cache-conf-$(safename)"
+  local cacherepo="localhost:${REGISTRY_PORT}/cache-conf-test"
+
+  CONTAINERS_CONF=$contextdir/containers.conf \
+    _build_with_cache ${imgname} ${cacherepo}
```
nalind, line 6552: This pushes cache entries to the repository...

tests/bud.bats
```diff
@@ -6474,6 +6474,218 @@ _EOF
...
+
+  CONTAINERS_CONF=$contextdir/containers.conf \
+    _push_and_inspect ${imgname}
```
nalind: ... and this pushes "latest" to the repository and inspects it.  This doesn't appear to be looking at the manifests for cache entries.  I'd suggest skipping the push of "latest", using skopeo's "list-tags" to choose a cache image, and inspecting that.

tests/bud.bats
```diff
@@ -6474,6 +6474,218 @@ _EOF
...
+@test "build --compression-format zstd to dir transport" {
+  _prefetch alpine
+  contextdir=${TEST_SCRATCH_DIR}/bud/compression-test
+  mkdir -p $contextdir
+  echo "unique-content-$(random_string)" > $contextdir/testfile
+  cat > $contextdir/Containerfile << _EOF
+FROM alpine
+RUN echo "layer1-$(random_string)"
+COPY testfile /testfile
+_EOF
```
nalind: Was this meant to reuse part of the _setup_compression_test helper function?  Would breaking that up into two helpers avoid some of this repetition?

nalind (changes requested):
Some questions about the tests that verify compression on --cache-to targets.

Honny1:
PTAL @containers/buildah-maintainers @containers/podman-maintainers @nalind

tests/bud.bats
```diff
@@ -6474,6 +6474,224 @@ _EOF
 
 }
 
+# Sets shell variable: contextdir
+# Creates a Containerfile with alpine base and a unique layer.
+function _setup_compression_context() {
```
nalind, line 6479: The comment for _setup_compression_test() notes that it sets "contextdir", and it calls this function to do that, so shouldn't this function note that, too?

tests/bud.bats
```diff
@@ -6474,6 +6474,224 @@ _EOF
...
+      $contextdir
+
+  CONTAINERS_CONF=$contextdir/containers.conf \
+    _push_and_inspect ${imgname}
+
+  expect_output --substring "zstd" \
+    "layers should use zstd compression per containers.conf"
+  assert "$output" !~ "tar+gzip" \
+    "layers should NOT use gzip when containers.conf specifies zstd"
+}
```
nalind, line 6600: This appears to be what the "push should respect compression_format from containers.conf" does.

tests/bud.bats
```diff
@@ -6474,6 +6474,224 @@ _EOF
...
+    -f $contextdir/Containerfile \
+    $contextdir
+
+  _push_and_inspect ${imgname} --compression-format zstd --force-compression
+
+  expect_output --substring "zstd" \
+    "layers should use zstd compression per --compression-format flag on push"
+  assert "$output" !~ "tar+gzip" \
+    "layers should NOT use gzip when --compression-format zstd is specified"
+}
```
nalind: This appears to be what the "push should respect --compression-format flag with registry" test is doing.

tests/bud.bats
```diff
@@ -6474,6 +6474,224 @@ _EOF
...
+    -t oci-archive:${outfile} \
+    -f $contextdir/Containerfile \
+    $contextdir
+
+  run skopeo inspect --raw oci-archive:${outfile}
+  assert "$output" =~ "zstd" \
+    "layers should use zstd mediaType for oci-archive: transport"
+  assert "$output" !~ "gzip" \
+    "layers should NOT use gzip when --compression-format zstd is specified"
+}
```
nalind: This could be merged with "build --compression-format zstd to dir transport" by having it build with -t pointing to different dir: and oci-archive: values inside of a loop, the way the "commit --compression-format zstd to dir, oci, and oci-archive" test does.  If we're going to use skopeo, it can inspect all three types of locations.

tests/bud.bats
```diff
@@ -6474,6 +6474,224 @@ _EOF
...
+    "cache layers should NOT use gzip when --compression-format specifies zstd"
+}
+
+@test "build --cache-to with --force-compression" {
+  _setup_compression_test
+  local imgname="img-cache-force-$(safename)"
+  local cacherepo="localhost:${REGISTRY_PORT}/cache-force-test"
+
+  _build_with_cache ${imgname} ${cacherepo}
+  _build_with_cache ${imgname} ${cacherepo} --compression-format zstd --force-compression
```
nalind: What's the intent behind building twice?  Except for that, this differs from "build --cache-to with --compression-format" only in whether or not it uses the "--force-compression" flag, which could be controlled by a loop variable.

tests/bud.bats
```diff
@@ -6474,6 +6474,224 @@ _EOF
...
+  _setup_compression_context
+  start_registry
+  authfile=${TEST_SCRATCH_DIR}/test.auth
+  run_buildah login --tls-verify=false --authfile $authfile --username testuser --password testpassword localhost:${REGISTRY_PORT}
+}
+
+function _build_with_cache() {
+  local imgname=$1; shift
+  local cacherepo=$1; shift
+  echo "unique-content-$(random_string)-$(date +%s)" > $contextdir/testfile
```
nalind: This is only called after _setup_compression_context(), which creates the file the same way.

nalind (commented):
Some questions about the tests.

tests/bud.bats
Honny1: It does. Row 6477

Honny1:
PTAL @containers/buildah-maintainers @containers/podman-maintainers @nalind

tests/bud.bats
nalind: Oof, missed that.  Sorry.

tests/commit.bats
```diff
@@ -642,3 +642,106 @@ load helpers
...
+    assert $status -eq 0 "reading manifest.json"
+  else
+    run cat "$destdir/index.json"
+    assert $status -eq 0 "reading index.json"
+    local manifest_digest
+    manifest_digest=$(jq -r '.manifests[0].digest' < "$destdir/index.json")
+    run cat "$destdir/blobs/${manifest_digest/://}"
+    assert $status -eq 0 "reading manifest"
+  fi
+}
```
nalind: Non-blocking: this could be done with skopeo.

tests/push.bats
```diff
@@ -224,3 +224,92 @@ load helpers
...
+      --tls-verify=false \
+      --authfile ${TEST_SCRATCH_DIR}/test.auth \
+      ${imgname} \
+      docker://${finalrepo}
+
+  run skopeo inspect \
+      --authfile ${TEST_SCRATCH_DIR}/test.auth \
+      --tls-verify=false \
+      --raw \
+      docker://${finalrepo}
```
nalind, line 267: Should probably be checking $status here.

tests/push.bats
```diff
@@ -224,3 +224,92 @@ load helpers
...
+    --authfile ${TEST_SCRATCH_DIR}/test.auth \
+    --compression-format zstd \
+    ${imgname} \
+    docker://${finalrepo}
+
+  run skopeo inspect \
+      --authfile ${TEST_SCRATCH_DIR}/test.auth \
+      --tls-verify=false \
+      --raw \
+      docker://${finalrepo}
```
nalind, line 310: Should probably be checking $status here.

tests/bud.bats
```diff
@@ -6474,6 +6474,129 @@ _EOF
...
+  assert $status -eq 0 "listing cache tags"
+  local tag
+  tag=$(jq -r '.Tags[0]' <<< "$output")
+  assert "$tag" != "null" "cache repo should have at least one tag"
+
+  run skopeo inspect \
+    --authfile $authfile \
+    --tls-verify=false \
+    --raw \
+    docker://${cacherepo}:${tag}
```
nalind, line 6538: Should probably check $status here.

tests/bud.bats
```diff
@@ -6474,6 +6474,129 @@ _EOF
...
+
+    run_buildah build \
+      $WITH_POLICY_JSON \
+      --no-cache \
+      --compression-format zstd \
+      -t ${transport}:${dest} \
+      -f $contextdir/Containerfile \
+      $contextdir
+
+    run skopeo inspect --raw ${transport}:${dest}
```
nalind, line 6593: Should probably check $status here.

nalind (approved):
Some nits in the tests, but LGTM.

Honny1:
PTAL @containers/buildah-maintainers @containers/podman-maintainers @nalind

nalind (approved):
LGTM, thanks!

Honny1:
PTAL @containers/buildah-maintainers @containers/podman-maintainers for final review and merge.

pkg/blobcache/blobcache.go
```diff
@@ -26,6 +33,7 @@ type BlobCache interface {
 // as-is to the specified directory or a temporary directory.
 // The compress argument controls whether or not the cache will try to substitute a compressed
 // or different version of a blob when preparing the list of layers when reading an image.
-func NewBlobCache(ref types.ImageReference, directory string, compress types.LayerCompression) (BlobCache, error) {
-	return imageBlobCache.NewBlobCache(ref, directory, compress)
+// The optional Option values can further refine behavior.
+func NewBlobCache(ref types.ImageReference, directory string, compress types.LayerCompression, opts ...Option) (BlobCache, error) {
```
mtrmac, line 37: (An absolutely non-blocking drive-by, by no means a review:) Maybe the Buildah callers could call the c/image package directly, without extending the API here.

pkg/cli/common.go
```diff
@@ -57,85 +57,88 @@ type NameSpaceResults struct {
...
+	OSFeatures             []string
+	OSVersion              string
+	CWOptions              string
+	SBOMOptions            []string
+	CompatVolumes          bool
+	SourceDateEpoch        string
+	RewriteTimestamp       bool
+	CreatedAnnotation      bool
+	SourcePolicyFile       string
+	TransientRunMounts     []string
```
TomSweeneyRedHat, line 141: Not to fix now, but I will note that I hate seeing long structs like this that aren't in alpha order.  It makes debugging later down the line harder IMHO>

TomSweeneyRedHat:
LGTM

pkg/blobcache/blobcache.go
nalind: These were left here to avoid breaking callers when the package's guts were moved.

mtrmac: Yes, I just meant that that doesn’t require adding _more_ compatibility wrappers. The existing `pkg/blobcache` could remain as is for compatibility, and callers within Buildah could call the c/image one with the new options.

Anyway this already merged.

PR #6758, opened 2026-04-02
Add additional caching diagnostics to stage executor

celskeggs:
#### What type of PR is this?

/kind other

The stage executor has an option to use an external registry as a cache. However, when a cache miss occurs, it can be difficult to understand why, even with the log level set to trace. This is because debug log messages are not produced for many important branches of the caching logic. Even worse, one of the messages indicates a cache hit even when a cache miss occurred.

By adding these debug statements (appropriately gated behind the log level), developers attempting to debug caching issues in buildah will have a better chance of understanding the caching decisions that buildah is making.

#### How to verify it

Run buildah with `--log-level trace` and `--cache-from` in a context that triggers a cache miss and observe the additional logging output.

None

None

None

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -1744,7 +1744,11 @@ func (s *stageExecutor) execute(ctx context.Context, base string) (imgID string,
 			// This log line is majorly here so we can verify in tests
 			// that our cache is performing in the most optimal way for
 			// various cases.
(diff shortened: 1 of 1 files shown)


nalind (approved):
We might want to tweak the logging level for these at some point, but otherwise LGTM.

Honny1 (approved):
LGTM

Issue #6737, opened 2026-03-20
Add support for Op_File.Actions.FileAction_Copy.FollowSymlink


BenjaminSchubert:
Splitting out of #6732, in order to start the work.

The aim is to add support for: `Actions.FileAction_Copy.FollowSymlink`:

- [ ] Add AddAndCopyOptions.FollowSymlink (type: go.podman.io/image/v5/types.OptionalBool) // follow symlinks when reading globbed items, default true
- [ ] Use to set or clear copier.GetOptions.NoDerefSymlinks

PR #6759, opened 2026-04-02
Fix the copier:get operation to properly gather symlink information

BenjaminSchubert:
#### What type of PR is this?

> /kind bug

Previously, the copier was always dereferencing symlinks, even if `noDerefSymlinks` was set. This passed unnoticed as the test was not validating the symlink value.

This augments the test to validate the symlink target, and ensures the copier sets the information completely.

#### How to verify it

Remove the fix, run the copier tests, notice that there is no more symlinks in them

None (I can create one if needed, this was discovered while working on #6737 )

I am unsure here, I don't think so?

```release-note

```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -1375,7 +1375,12 @@ func copierHandlerGet(bulkWriter io.Writer, req request, pm *fileutils.PatternMa
 				}
 				addedParents[parentName] = struct{}{}
 
-				if err := copierHandlerGetOne(parentInfo, "", parentName, parent, req.GetOptions, tw, hardlinkChecker, idMappings); err != nil {
+				parentSymlinkTarget, err := getTargetIfSymlink(parent, parentInfo)
+				if err != nil {
+					return fmt.Errorf("copier: get: %w", err)
+				}
+
+				if err := copierHandlerGetOne(parentInfo, parentSymlinkTarget, parentName, parent, req.GetOptions, tw, hardlinkChecker, idMappings); err != nil {
 					if req.GetOptions.IgnoreUnreadable && errorIsPermission(err) {
 						continue
 					} else if errors.Is(err, os.ErrNotExist) {
@@ -1535,7 +1540,12 @@ func copierHandlerGet(bulkWriter io.Writer, req request, pm *fileutils.PatternMa
 					}
 				}
 
-				if err := copierHandlerGetOne(info, "", name, item, req.GetOptions, tw, hardlinkChecker, idMappings); err != nil {
+				symlinkTarget, err := getTargetIfSymlink(item, info)
+				if err != nil {
+					return fmt.Errorf("copier: get: %w", err)
+				}
+
+				if err := copierHandlerGetOne(info, symlinkTarget, name, item, req.GetOptions, tw, hardlinkChecker, idMappings); err != nil {
 					if req.GetOptions.IgnoreUnreadable && errorIsPermission(err) {
 						continue
 					}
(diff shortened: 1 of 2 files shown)


copier/copier.go
```diff
@@ -1476,7 +1481,7 @@ func copierHandlerGet(bulkWriter io.Writer, req request, pm *fileutils.PatternMa
 						}
 						symlinkTarget = target
 					}
-					info, err := d.Info()
+					info, err := os.Lstat(path)
```
nalind: This doesn't seem to be necessary?  Was it originally intended to replace the block above it with a call to the new getTargetIfSymlink() function?

copier/copier_test.go
```diff
@@ -1563,6 +1563,27 @@ func testGetMultiple(t *testing.T) {
 						if testCase.timestamp != nil {
 							assert.Truef(t, testCase.timestamp.Equal(hdr.ModTime), "timestamp was supposed to be forced for %q", hdr.Name)
 						}
+
+						// Find the original file back in the test definition and record whether it should be a symlink
+						isOriginallySymlink := false
+						expectedLinkTarget := ""
+						for _, origHdr := range testArchive.headers {
+							if origHdr.Name == hdr.Name && origHdr.Typeflag == tar.TypeSymlink {
```
nalind: This looks like it would skip over items whose paths are changed for some reason, either because they were explicitly renamed due to a `renames` setting, or because they were copied from one subdirectory to another.

There are a number of the test cases where the expected set of items includes the contents of a directory that was specified as the source for the copy operation, where the contents of the directory, but not the directory itself, are expected to be copied to the destination.

It might be simpler to add a new field to the test cases which maps from the members of the `items` slice to the corresponding expected `Linkname` field values in the headers with matching `Name` values.  This loop body could then populate a new map using the `Name` and `Linkname` values for every header it sees with type `TypeSymlink`, and then those maps could be compared when `expectedContents` and `actualContents` are compared.

nalind (changes requested):
A question about a change to a WalkDir callback, and I don't think that the additions to the tests are examining everything that we want them to.

copier/copier.go
BenjaminSchubert: Indeed, I thought this was still needed after the other change, but no. I've removed it

copier/copier_test.go
BenjaminSchubert: Good shout. I've split the commits in two now:

First commit will contain the expansion of the copier_test.go to validate the symlink content as of main branch
Second commit applies the fix and fixes the test for easier reviewing of what is changing

copier/copier.go
```diff
@@ -1595,6 +1605,18 @@ func mapWithPrefixedKeysWithoutKeyPrefix[K any](m map[string]K, p string) map[st
 	return cloned
 }
 
+func getTargetIfSymlink(path string, info os.FileInfo) (string, error) {
+	if info.Mode()&os.ModeType == os.ModeSymlink {
+		target, err := os.Readlink(path)
+		if err != nil {
+			return "", fmt.Errorf("readlink(%q): %w", path, err)
```
nalind: Nit: the standard library wraps the error with the path in a PathError result, and this would add the path again.

nalind (commented):
LGTM with a nit, and the two commits might as well be squashed since they touch some of the same test cases.

copier/copier.go
BenjaminSchubert: Sure thing, done. This is a pattern that's present multiple time in the same file, should I clean that up in a separate PR?

BenjaminSchubert:
@nalind This is now squashed and ready to go from my side. Thanks for the reviews!

Honny1 (approved):
LGTM, I think CI failures are flakes. (I don't have permission to rerun them.)

Honny1:

giuseppe (approved):
LGTM

nalind (approved):
LGTM, thanks!

PR #6760, opened 2026-04-02
Add a FollowSymlink option to AddAndCopyOptions to allow disabling deference of symlinks

### LINKED ISSUE #6737: Add support for Op_File.Actions.FileAction_Copy.FollowSymlink

BenjaminSchubert:
#### What type of PR is this?

> /kind feature

This PR adds the ability to skip the dereference of symlinks when doing a ADD or COPY.

Depends on #6759 

#### How to verify it

See the bats tests provided

Fixes #6737 

```release-note
Add a --no-follow-symlinks option to `buildah add` to skip dereferencing the symlinks when copying
```

### DIFF
--- a/docs/buildah-add.1.md
+++ b/docs/buildah-add.1.md
@@ -84,6 +84,11 @@ container's filesystem. If `buildah run` creates a file and `buildah add --link`
 to the same path, the file from `buildah add --link` will be present in the committed image.
 The --link layer is applied after all container filesystem changes at commit time.
 
+**--no-follow-symlinks**
+
+When a local source is a symbolic link, copy the link as a symbolic link
+instead of dereferencing the link and copying the contents of the target.
+
 **--quiet**, **-q**
 
 Refrain from printing a digest of the added content.
--- a/add.go
+++ b/add.go
@@ -122,6 +122,9 @@ type AddAndCopyOptions struct {
 	// AllowEmptyWildcard controls whether the operation succeeds when all
 	// glob patterns match nothing. Defaults to false.
 	AllowEmptyWildcard types.OptionalBool
+	// FollowSymlink controls whether symlinks should be followed when copying content.
+	// When set to false, symlinks are not dereferenced.
+	FollowSymlink types.OptionalBool
 }
 
 // getURL writes a tar archive containing the named content
@@ -615,6 +618,7 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 						ChownFiles:         chownFiles,
 						ChmodFiles:         chmodDirsFiles,
 						KeepDirectoryNames: options.DirCopyContents == types.OptionalBoolFalse,
+						NoDerefSymlinks:    options.FollowSymlink == types.OptionalBoolFalse,
 						StripSetuidBit:     options.StripSetuidBit,
 						StripSetgidBit:     options.StripSetgidBit,
 						StripStickyBit:     options.StripStickyBit,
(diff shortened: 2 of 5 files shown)


docs/buildah-add.1.md
```diff
@@ -74,6 +74,11 @@ container's filesystem. If `buildah run` creates a file and `buildah add --link`
 to the same path, the file from `buildah add --link` will be present in the committed image.
 The --link layer is applied after all container filesystem changes at commit time.
 
+**--no-follow-symlinks**
+
+Don't follow and dereference the symlinks when copying the files. Instead, copy
+the symlinks themselves.
```
nalind: There's a subtlety here - in practice this only affects whether or not we chase a link that's directly given to us.  If we're copying a directory, and there's a symlink inside of it, it's always copied as a symlink.

nalind (changes requested):
Code and tests look good.  The docs should probably be a bit more specific, since this doesn't affect symlinks in directories the way a similar flag might for the regular `cp` command from coreutils.

docs/buildah-add.1.md
BenjaminSchubert: Good point, I've updated it. Is that better?

BenjaminSchubert: @nalind PTAL when you have time :)

nalind: Perhaps something like "when a local source is a symbolic link, copy the link as a symbolic link instead of copying the contents of the link's target", assuming that makes sense when you read it?
The man page for "cp" wasn't giving me a lot of inspiration, I guess because its behavior is also different, and the one for podman-cp doesn't touch on the subject.

BenjaminSchubert: Sorry for the delay. That's indeed much better. I've tweaked it a bit, what do you think? Let me know if you prefer your version and I'll revert to that

nalind: That should work, thanks!

nalind (approved):
LGTM

docs/buildah-add.1.md
```diff
@@ -74,6 +74,11 @@ container's filesystem. If `buildah run` creates a file and `buildah add --link`
 to the same path, the file from `buildah add --link` will be present in the committed image.
 The --link layer is applied after all container filesystem changes at commit time.
 
+**--no-follow-symlinks**
+
+When a local source is a symbolic link, copy the link as a symbolic link
+instead of dereferencing the link and copying the contents of the target
```
giuseppe: nit: missing `.`. at the end of the sentence

BenjaminSchubert: Good catch, fixed

BenjaminSchubert: @giuseppe PTAL when you have time (and apologies for the extra ping)

giuseppe (approved):
LGTM

giuseppe:
could you please rebase?

BenjaminSchubert:
@giuseppe woops, apologies. Done, though the packaging step seems to be stuck somehow?

giuseppe (approved):

giuseppe:
> @giuseppe woops, apologies. Done, though the packaging step seems to be stuck somehow?

yeah, let's not worry about them.  Merged.

Thanks!

PR #6772, opened 2026-04-06
tests/helpers.bash: when determining the OCI runtime, use temporary storage

nalind:
#### What type of PR is this?

/kind other

Wait until we have a per-test default storage location set up before attempting to determine which OCI runtime we're using, so that we don't trigger an error message if the system-wide defaults would cause one to be printed.

#### How to verify it

Manually running a test using `bats` as an unprivileged user should no longer print
```bash
Error: mkdir /run/containers/storage: permission denied
WARN[0000] failed to shutdown storage: "mkdir /run/containers/storage: permission denied" 
```
error messages on systems where the system-wide settings haven't been updated to match the expectations of newer versions of go.podman.io/common.

None

```release-note
None
```

### DIFF
--- a/tests/helpers.bash
+++ b/tests/helpers.bash
@@ -15,7 +15,6 @@ PASSWD_BINARY=${PASSWD_BINARY:-$TEST_SOURCES/../bin/passwd}
 GRPCNOOP_BINARY=${GRPCNOOP_BINARY:-$TEST_SOURCES/../bin/grpcnoop}
 STORAGE_DRIVER=${STORAGE_DRIVER:-vfs}
 PATH=$(dirname ${BASH_SOURCE})/../bin:${PATH}
-OCI=${BUILDAH_RUNTIME:-$(${BUILDAH_BINARY} info --format '{{.host.OCIRuntime}}' || command -v runc || command -v crun)}
 # Default timeout for a buildah command.
 BUILDAH_TIMEOUT=${BUILDAH_TIMEOUT:-300}
 
@@ -60,6 +59,9 @@ function setup_tests() {
     # me: "but it's a local file path, not a repository name!"
     # buildah/podman: "i dont care. no caps anywhere!"
     TEST_SCRATCH_DIR=$(mktemp -d --dry-run --tmpdir=${BATS_TMPDIR:-${TMPDIR:-/tmp}} buildah_tests.XXXXXX | tr A-Z a-z)
+    if test -z "${TEST_SCRATCH_DIR}" ; then
+        die error creating temporary directory
+    fi
     mkdir --mode=0700 $TEST_SCRATCH_DIR
 
     mkdir -p ${TEST_SCRATCH_DIR}/{root,runroot,sigstore,registries.d}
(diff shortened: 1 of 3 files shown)


tests/helpers.bash
```diff
@@ -88,6 +87,17 @@ EOF
...
 }
 
+function find_oci_runtime() {
+    local tmpdir=$(mktemp -d ${BATS_SUITE_TMPDIR:-${TMPDIR}}/buildah.bats.XXXXXX)
+    if test -z "${tmpdir}" ; then
+        echo error creating temporary directory
+        exit 1
+    fi
+    local ROOTDIR_OPTS="--root ${tmpdir}/root --runroot ${tmpdir}/runroot --storage-driver ${STORAGE_DRIVER}"
+    echo ${BUILDAH_RUNTIME:-$(${BUILDAH_BINARY} ${ROOTDIR_OPTS} info --format '{{.host.OCIRuntime}}' || command -v runc || command -v crun)}
```
lsm5: nit: missing double-quotes around the expansion — echo "$(...)"  would be safer to match the quoting style on the rm line below

tests/helpers.bash
```diff
@@ -88,6 +87,17 @@ EOF
     PODMAN_REGISTRY_OPTS="${regconfopt}"
 }
 
+function find_oci_runtime() {
+    local tmpdir=$(mktemp -d ${BATS_SUITE_TMPDIR:-${TMPDIR}}/buildah.bats.XXXXXX)
+    if test -z "${tmpdir}" ; then
+        echo error creating temporary directory
+        exit 1
```
lsm5: exit 1 kills the entire bats process rather than just failing the current test. Would return 1 be more appropriate here?

tests/helpers.bash
```diff
@@ -88,6 +87,17 @@ EOF
     PODMAN_REGISTRY_OPTS="${regconfopt}"
 }
 
+function find_oci_runtime() {
+    local tmpdir=$(mktemp -d ${BATS_SUITE_TMPDIR:-${TMPDIR}}/buildah.bats.XXXXXX)
```
lsm5: If both `BATS_SUITE_TMPDIR` and `TMPDIR` are unset this becomes `mktemp -d /buildah.bats.XXXXXX` which would try to create in /. Maybe add a fallback: `${BATS_SUITE_TMPDIR:-${TMPDIR:-/tmp}}`

lsm5 (approved):
LGTM w/ nits.

tests/helpers.bash
Luap99: BATS_SUITE_TMPDIR should always be defined by bast so I do not see the need for fall back

tests/helpers.bash
Luap99: it should use `die` to fail a test

tests/helpers.bash
```diff
@@ -88,6 +87,17 @@ EOF
...
+    if test -z "${tmpdir}" ; then
+        echo error creating temporary directory
+        exit 1
+    fi
+    local ROOTDIR_OPTS="--root ${tmpdir}/root --runroot ${tmpdir}/runroot --storage-driver ${STORAGE_DRIVER}"
+    echo ${BUILDAH_RUNTIME:-$(${BUILDAH_BINARY} ${ROOTDIR_OPTS} info --format '{{.host.OCIRuntime}}' || command -v runc || command -v crun)}
+    rm -fr "${tmpdir}"
+}
```
Luap99: why even bother with this?

If this is only called from test context just use run_buildah?

And in test context it could use `TEST_SCRATCH_DIR` which is always cleaned up already?

Luap99 (commented):
personally is this worth to inline in all callers?
seems like we could still only have a global check once at the beginning and use the result cached instead of having to create the tmpdirs and call info many times

tests/helpers.bash
nalind: Revising this, and a number of similar call sites.

tests/helpers.bash
nalind: Dropping that.

tests/helpers.bash
nalind: The `run_buildah` helper function logs what command it's going to run, and that gets picked up along with the result that we're looking for.

Luap99: run_buildah populates "$output"? so just use that one over redirection? I mean all tests use "$output" for matching output.

nalind: Sure, the caller has to read the variable instead of capturing the function output, but that works.  Changing it.

Luap99 (approved):
LGTM

PR #6773, opened 2026-04-06
Group global commands in global help output

nalind:
#### What type of PR is this?

/kind documentation

Group the grab bag of top-level commands, in an attempt to make the help output more helpful.

#### How to verify it

Manual inspection.

None

I'm not strongly tied to the categories.

```release-note
Commands are now grouped in the output of `buildah --help`.
```

### DIFF
--- a/cmd/buildah/addcopy.go
+++ b/cmd/buildah/addcopy.go
@@ -55,7 +55,8 @@ func createCommand(addCopy string, desc string, short string, opts *addCopyResul
 		Example: `buildah ` + addCopy + ` containerID '/myapp/app.conf'
   buildah ` + addCopy + ` containerID 'app.conf' '/myapp/app.conf'
   buildah ` + addCopy + ` containerID 'app.conf' 'drop-in.conf' '/myapp/app.conf.d/'`,
-		Args: cobra.MinimumNArgs(1),
+		Args:    cobra.MinimumNArgs(1),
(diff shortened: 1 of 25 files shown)


cmd/buildah/addcopy.go
```diff
@@ -55,7 +55,8 @@ func createCommand(addCopy string, desc string, short string, opts *addCopyResul
 		Example: `buildah ` + addCopy + ` containerID '/myapp/app.conf'
   buildah ` + addCopy + ` containerID 'app.conf' '/myapp/app.conf'
   buildah ` + addCopy + ` containerID 'app.conf' 'drop-in.conf' '/myapp/app.conf.d/'`,
-		Args: cobra.MinimumNArgs(1),
+		Args:    cobra.MinimumNArgs(1),
+		GroupID: "containers",
```
Luap99: to avoid typos I think would make sense to define these strings as constants and resuse them

nalind: Sure.  Currently, the Execute() method will panic if a command has a GroupID that wasn't previously set up with AddGroup(), but that doesn't seem to be documented, so I guess it could change in the future, and we don't want to depend on it staying that way.

Luap99 (approved):
LGTM

Honny1 (approved):
LGTM

Issue #6732, opened 2026-03-17
Build definition API checklist


nalind:
We'd like to be able to drive a build based on already-parsed build definition, but being able to consume those specifications properly is going to require that we add to our lower-level APIs support for a number of features which can be expressed in a definition that the APIs can't currently fulfill.  It also makes sense to expose some of these features through the corresponding CLI commands.

* Op_Exec.Meta.ValidExitCodes:
  - [ ] Add RunOptions.ValidExitCodes (type: []int32) // expect/accept any of these exit status values as success instead of 0
  - [ ] Add --valid-exit-codes to `buildah run`, default []int32{0}.
  - [ ] Add an integration test using `buildah run`.

* Op_Exec.Mounts.CacheOpt.Sharing:
  - Reported in https://github.com/containers/podman/issues/27188
  - [ ] Teach internal.volumes.GetCacheMount() about "sharing=private"
  - [ ] Add an integration test using `buildah run`.

* Op_Exec.Security:
  - [ ] Think _really_ hard about ways to avoid having to do this one.
  - Maybe come back to this one later.

* Op_File.Actions.FileAction_Copy.FollowSymlink:
  - [ ] Add AddAndCopyOptions.FollowSymlink (type: go.podman.io/image/v5/types.OptionalBool) // follow symlinks when reading globbed items, default true
  - [ ] Use to set or clear copier.GetOptions.NoDerefSymlinks

* Op_File.Actions.FileAction_Copy.DirCopyContents:
  - [ ] Add AddAndCopyOptions.DirCopyContents (type: go.podman.io/image/v5/types.OptionalBool) // copy the directory's contents instead of the directory with its contents below it, default true
  - [ ] Use to set or clear copier.GetOptions.KeepDirectoryNames

* Op_File.Actions.FileAction_Copy.AttemptUnpackDockerCompatibility:
  - [ ] Add copier.PutOptions.AttemptUnpackDockerCompatibility (type: bool) // detect if globbed items are possibly-compressed archives, and if so, extract them instead of copying them
  - [ ] Add to unit tests in copier package

* Op_File.Actions.FileAction_Copy.CreateDestPath:
  - [ ] Add copier.CreateDestPath (type: go.podman.io/image/v5/types.OptionalBool) // enable or disable the implicit creation of the destination directory if it don't exist yet
  - [ ] Add to unit tests in copier package

* Op_File.Actions.FileAction_Copy.AllowWildcard:
[...]

PR #6778, opened 2026-04-09
Add support for symbolic chmod options

### LINKED ISSUE #6066: Containerfile `--chmod +x` not working in podman
This is a reincarnation of a previous issue

* https://github.com/containers/buildah/issues/4614

opened because Docker has added support for symbolic permissions notation in the meantime

* https://github.com/moby/buildkit/issues/1951

---

Dockerfile:

```
# mkdir somedir
# touch somedir/somefile.txt
FROM alpine
COPY --chmod='ugo=rwX' ./ /
```

Build with Docker

```
docker --version
Docker version 28.0.1, build 068a01ea9
```

```
docker build -t perms --load .

[+] Building 0.8s (8/8) FINISHED                                                                                                                                                     docker-container:default
 => [internal] load build definition from Dockerfile                                                                                                                                                     0.0s
 => => transferring dockerfile: 159B                                                                                                                                                                     0.0s
 => [internal] load metadata for docker.io/library/alpine:latest                                                                                                                                         0.6s
 => [internal] load .dockerignore                                                                                                                                                                        0.0s
 => => transferring context: 2B                                                                                                                                                                          0.0s
 => [internal] load build context                                                                                                                                                                        0.0s
 => => transferring context: 363B                                                                                                                                                                        0.0s
 => CACHED [1/2] FROM docker.io/library/alpine:latest@sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c                                                                            0.0s
 => => resolve docker.io/library/alpine:latest@sha256:a8560b36e8b8210634f77d9f7f9efd7ffa463e380b75e2e74aff4511df3ef88c                                                                                   0.0s
 => [2/2] COPY --chmod=ugo=rwX ./ /                                                                                                                                                                      0.0s
 => exporting to docker image format                                                                                                                                                                     0.1s
 => => exporting layers                                                                                                                                                                                  0.0s
 => => exporting manifest sha256:3146a557f380db9a501fee36ca17c02ca13a9ef6464a6f94cae60dbc7df77822                                                                                                        0.0s
 => => exporting config sha256:00d3546586d3d8b152f4f321cef751e5e5a82ce9dd7c0d092cf52dbb86a68a2d                                                                                                          0.0s
 => => sending tarball                                                                                                                                                                                   0.1s
 => importing to docker
```

```
docker run --rm -it perms ls -AlFd /somedir /somedir/somefile.txt
drw-rw-rw-    2 root     root            26 Mar 16 17:20 /somedir/
-rw-rw-rw-    1 root     root             0 Mar 16 17:20 /somedir/somefile.tx
```

(that capital X does not yet work with `=` and that's a known bug)

* https://github.com/moby/buildkit/issues/5835

With Podman

```
podman --version
podman version 5.4.1
```

```
podman build -t perms .
STEP 1/2: FROM alpine
Resolved "alpine" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull docker.io/library/alpine:latest...
Copying blob sha256:6e771e15690e2fabf2332d3a3b744495411d6e0b00b2aea64419b58b0066cf81
Copying config sha256:8d591b0b7dea080ea3be9e12ae563eebf9869168ffced1cb25b2470a3d9fe15e
STEP 2/2: COPY --chmod='ugo=rwX' ./ /
Error: building at STEP "COPY --chmod=ugo=rwX ./ /": Error parsing chmod ugo=rwX
```

nickjwhite:
#### What type of PR is this?

/kind feature

This adds an optional Chmod string option to GetOptions and PutOptions. If set, this overrides the ChmodDirs and ChmodFiles options. If unset, they continue to work as before.

The --chmod option to buildah add and buildah copy works with symbolic as well as numeric arguments.

#### How to verify it

See extra bats tests and copier tests.

Fixes #6066

Note that I'm not sure whether the extra tests in `commit_test.go` add much value over those in `copier/copier_test.go`, but I used them while working on this change, so I figured I'd just include them.

Also I'm not sure if this is the way you want the vendored stuff to be included (I added the newly added directory after running  'make vendor') - let me know if I should do it differently.

https://github.com/containers/buildah/issues/6732 suggested calling the new option in `GetOptions` & `PutOptions` `ChmodStr`, but I chose `Chmod` here because I think it's shorter and clearer.

```release-note
The add and copy commands' --chmod argument now accepts symbolic notation, as well as the traditional numeric notation currently supported. Likewise copier's GetOptions and PutOptions now a Chmod option, which accepts numeric or symbolic notation, and if set overrides any ChmodDirs and ChmodFiles options.
```


commit_test.go
```diff
@@ -575,3 +575,88 @@ func TestCommitEmpty(t *testing.T) {
...
+
+	for i, blobInfo := range infos {
+		rc, _, err := src.GetBlob(ctx, blobInfo, nil)
+		require.NoError(t, err, "getting blob", i)
+		defer rc.Close()
+		tr := tar.NewReader(rc)
+		entry, err := tr.Next()
+		for entry != nil {
+			expected := filePerms[entry.Name]
+			require.Equalf(t, entry.Mode, expected.perm, "unexpected mode for %s, got %#od wanted %#od", entry.Name, entry.Mode, expected.perm)
```
nalind: On error, `require.Equalf()` logs the second argument as the "expected" value in its error message, and the third as the "actual" value, so these should be flipped.  They don't need to be duplicated by the format string and the additional arguments.

copier/copier.go
```diff
@@ -1705,11 +1708,22 @@ func copierHandlerGetOne(srcfi os.FileInfo, symlinkTarget, name, contentPath str
 		}
 	}
 	// force ownership and/or permissions, if requested
+	if options.Chmod != "" {
+		p, err := mode.Parse(options.Chmod)
```
nalind, line 1715: We call `copierHandlerGetOne()` for every item that we're reading.  How expensive is this?

copier/copier_test.go
```diff
@@ -2521,3 +2522,132 @@ func TestTarPut(t *testing.T) {
...
+			}()
+			tr := tar.NewReader(pipeReader)
+			hdr, tarErr := tr.Next()
+			for tarErr == nil {
+				if hdr.Mode != int64(v.endMode) {
+					t.Errorf("expected mode %O, got mode %O", v.endMode, hdr.Mode)
+				}
+				hdr, tarErr = tr.Next()
+			}
+			assert.Equal(t, io.EOF.Error(), tarErr.Error(), "expected EOF at end of archive, got %q", tarErr.Error())
```
nalind: It looks like `assert.ErrorIs(t, tarErr, io.EOF, "expected EOF at end of archive")` would work here.

copier/copier_test.go
```diff
@@ -2521,3 +2522,132 @@ func TestTarPut(t *testing.T) {
...
+			}
+			if err != nil && !strings.Contains(err.Error(), v.err.Error()) {
+				t.Fatalf("got error %v, expected error %v", err, v.err)
+			}
+			if err != nil {
+				return
+			}
+			if v.err != nil {
+				t.Errorf("got no error, expected error %v", v.err)
+			}
```
nalind: This could probably call `require.ErrorIs()`.

copier/copier_test.go
```diff
@@ -2521,3 +2522,132 @@ func TestTarPut(t *testing.T) {
...
+			if err != nil && !errors.Is(err, v.err) {
+				t.Fatalf("got error %v, expected error %v", err, v.err)
+			}
+			if err != nil {
+				return
+			}
+			if v.err != nil {
+				t.Errorf("got no error, expected error %v", v.err)
+			}
```
nalind: It looks like this could call `require.ErrorIs()` to check the error.

nalind (commented):
Overall this looks very good to me.  A couple of quibbles about using test helper functions that would shorten them, and a question about calling `mode.Parse()` multiple times when reading.

nickjwhite:
Thanks for the detailed review, @nalind! I'll fix these up tomorrow and push a fresh version.

copier/copier_test.go
nickjwhite: Good spot. This was copied from a few other places in this test file, but your suggestion is clearly better, so I'll just update those places to use it too.

nalind: Right, comparing the error text should only be necessary if one or the other errors is being carried across a process boundary as text.

commit_test.go
nickjwhite: Thanks for the explanation, somehow I'd barely used `testify`, so missed this.

copier/copier_test.go
nickjwhite: Done, this one did require a little more work as going through Put() the error has to run through a `%v`, so we still need to check the `.Error()` string.

copier/copier.go
nickjwhite: Good question. I think the only way to avoid this would be to compute the result of `mode.Parse` before the calls to `copierHandlerGetOne()`, and pass it as an argument to the function. As it already takes a lot of arguments I don't love this, as it feels a bit ugly. I haven't benchmarked `mode.Parse()`, but it looks pretty efficient, exits out early if it just contains a numeric string, and obviously is only called if `options.Chmod` is set.

However I don't have strong feelings either way, if you think passing the result of `mode.Parse()` to `copierHandlerGetOne()` is preferable I'm happy to make the change to make that happen. Or if you have an alternative suggestion.

nickjwhite:
> Thanks for the detailed review, @nalind! I'll fix these up tomorrow and push a fresh version.

I ended up not waiting 'til tomorrow and just doing it. Replaced a couple more `if` statements with `assert.Equal` while I was there, too.

I'm not sure what to do about the `mode.Parse()` call in `copierHandlerGetOne()`, I replied to your comment on that, happy to take your guidance on it.

copier/copier.go
nalind: Yeah, go ahead and initialize the mode.Set in the caller, and pass a pointer to it into copierHandlerGetOne() as an argument (probably with a more descriptive name, "chmod" as the Put handler calls it would be fine) so that it can be compared to nil where it's currently comparing options.Chmod to an empty string.

commit_test.go
```diff
@@ -575,3 +575,88 @@ func TestCommitEmpty(t *testing.T) {
...
+
+	for i, blobInfo := range infos {
+		rc, _, err := src.GetBlob(ctx, blobInfo, nil)
+		require.NoError(t, err, "getting blob", i)
+		defer rc.Close()
+		tr := tar.NewReader(rc)
+		entry, err := tr.Next()
+		for entry != nil {
+			expected := filePerms[entry.Name]
+			require.Equalf(t, entry.Mode, expected.perm, "unexpected mode for %s", entry.Name)
```
nalind: I think this one was missed when updating the ordering of args to require.Equalf()?

nalind (commented):
Please go ahead and make that change.  I think we're going to want this to be rebased before merging.

nickjwhite:
> Please go ahead and make that change. I think we're going to want this to be rebased before merging.

Done, including the rebase. I haven't tested as heavily, as the tests aren't all working on my local machine as they should, but I'll trust in the CI tests for now.

nalind:
> Done, including the rebase. I haven't tested as heavily, as the tests aren't all working on my local machine as they should, but I'll trust in the CI tests for now.

Changes look good, but usually I expect merge commits to be dropped during a rebase.  Did you `git fetch upstream; git rebase -i upstream/main`, or use a different method?

nickjwhite:
> Changes look good, but usually I expect merge commits to be dropped during a rebase. Did you `git fetch upstream; git rebase -i upstream/main`, or use a different method?

Oh, apologies, I said I rebased but I actually just merged main, indeed, my bad. I made a fresh branch from latest main, and cherry-picked & manually reapplied & amended my changes to it. So hopefully this is better.

copier/copier.go
```diff
@@ -1726,11 +1737,18 @@ func copierHandlerGetOne(srcfi os.FileInfo, symlinkTarget, name, contentPath str
 		}
 	}
 	// force ownership and/or permissions, if requested
+	if chmod != nil {
+		stat, err := os.Lstat(contentPath)
+		if err != nil {
+			return fmt.Errorf("reading permissions of %q: %w", contentPath, err)
+		}
+		hdr.Mode = int64(chmod.Apply(stat.Mode()))
```
nalind: Looking at this again, if the `hdr.Mode` value returned by the tar.FileInfoHeader() isn't correct, doesn't `srcfi.Mode()` return the same value?

nalind:
Much, better, thanks!

copier/copier.go
nickjwhite: Oh great spot, thanks, I didn't notice `srcfi`! I removed those unneeded lines from here so we just have `hdr.Mode = int64(chmod.Apply(srcfi.Mode()))` now.

nalind (approved):
LGTM

nalind:

TomSweeneyRedHat:
@nickjwhite looks like you may need to rebase this.

tests/copy.bats
```diff
@@ -211,6 +212,9 @@ load helpers
   run_buildah copy --chmod 700 $cid ${TEST_SCRATCH_DIR}/randomfile /randomfile2
   run_buildah copy --chmod 755 $cid ${TEST_SCRATCH_DIR}/randomfile /randomfile3
   run_buildah copy --chmod 660 $cid ${TEST_SCRATCH_DIR}/subdir /subdir
+  run_buildah copy --chmod u=rX,go= $cid ${TEST_SCRATCH_DIR}/randomfile /randomfile4
+  run_buildah copy --chmod u=rwx,go=u-w $cid ${TEST_SCRATCH_DIR}/randomfile /randomfile5
+  run_buildah copy --chmod u=rwX,go=X $cid ${TEST_SCRATCH_DIR}/subdir3 /extradir/subdir3
```
TomSweeneyRedHat, line 217: Since you need to rebase, would you mind adding a test that uses a 4 digit numerical value llike `0755` please?

TomSweeneyRedHat: Also, could you add similar tests to buildah/tests
/add.bats please?

TomSweeneyRedHat:
Other than a few test tweaks, and a rebase
LGTM

tests/copy.bats
nickjwhite: Good ideas, I've rebased and added a 4 digit numerical value to the copy.bats, I'm happy to add some add.bats tests, but it'll probably take me a couple of days to get to as I'm short on time at the moment.

nickjwhite: I've just added similar tests to `buildah/tests/add.bats` :)

nickjwhite:
@TomSweeneyRedHat can you take another look please? I added the extra tests you suggested and rebased, so this should be ready to go.

TomSweeneyRedHat:
@nickjwhite apologies for the very tardy re-review.  LGTM.
However, it looks like you need to rebase before we can commit.

nickjwhite:
> @nickjwhite apologies for the very tardy re-review. LGTM. However, it looks like you need to rebase before we can commit.

No problem Tom, thanks for the re-review! I have rebased again.

nalind (approved):
LGTM, thanks!

nickjwhite:
Hi again, 2 LGTMs, any reason not to just merge it now? I expect you've all been busy pushing towards Podman 6, so no worries, I just don't want this to be forgotten about :blush:

nalind:
Sorry for the delay.  Manually resolved a go.mod conflict, hopefully correctly.

nickjwhite:
Thanks for all your help getting this over the line, @nalind!

PR #6785, opened 2026-04-13
Makefile: add some missing dependencies

nalind:
#### What type of PR is this?

/kind other

Running `make clean` before assorted targets would cause make to fail unexpectedly when building those targets due to missing target dependencies; add them.  Drop a dependency from the "copy" and "imgtype" test helpers instead of adding it to the Makefile.

Drop the "rpm-install" target, which appears to be unused, as it depends on a "package" dependency which isn't defined in the Makefile.

Rework the lint-entrypoint target to parse the file in Go, to make it easier to catch errors while it's doing its work.

#### How to verify it
Run `make clean` before any other targets.

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,15 +150,7 @@ validate: all install.tools lint-entrypoint
 
 .PHONY: lint-entrypoint
 lint-entrypoint: internal/mkcw/embed/entrypoint_amd64.gz
-	@tmp=$$(mktemp) ; \
-	zcat internal/mkcw/embed/entrypoint_amd64.gz > "$$tmp" ; \
-	if readelf -S "$$tmp" | grep -qE '\.go\.|gopclntab'; then \
-		echo "FAIL: entrypoint_amd64.gz was built using Go fallback path" ; \
-		rm -f "$$tmp" ; \
-		exit 1 ; \
-	fi ; \
-	echo "OK: entrypoint_amd64.gz was built using native assembler" ; \
-	rm -f "$$tmp"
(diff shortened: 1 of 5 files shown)


Luap99 (approved):
LGTM

Makefile
```diff
@@ -143,29 +143,21 @@ codespell:
...
+lint-entrypoint: internal/mkcw/embed/entrypoint_amd64.gz
+	$(GO) run tests/validate/was-not-go-compiled.go internal/mkcw/embed/entrypoint_amd64.gz
 
 .PHONY: install.tools
 install.tools:
 	$(MAKE) -C tests/tools
 
 .PHONY: install
-install:
+install: bin/buildah
```
lsm5, line 160: Do we need this?

In the rpm process this will end up rebuilding the binary in the `%install` stage because the %build stage built it using the `%gobuild` macro, unless Makefile is patched in the rpm.

nalind: `make clean install` fails without it, so I do think we need it.
Is the timestamp on one of its dependencies altered after the %build phase builds the binary?  If not, I'd be surprised if `make install` forced it to be rebuilt.

lsm5: Hmm, https://download.copr.fedorainfracloud.org/results/packit/containers-buildah-6785/fedora-rawhide-x86_64/10325255-buildah/builder-live.log.gz does NOT rebuild the binary in %install . So, you're right @nalind . Could've sworn this used to happen long ago in podman or some other Go package we maintained. 

Anyway..

lsm5 (approved):
LGTM

PR #611, opened 2018-04-20
Revert "Update entrypoint behaviour to match docker"

pixdrift:
This reverts commit 15e1054820a0442ce179ca44ddd870b8ff56fa8c.

The original commit changed entrypoint behaviour in `buildah run` to match `docker run`. This breaks the desired state of `buildah run` command line behaviour matching `buildah bud RUN` (Dockerfile) behaviour.

To achieve `docker run` entrypoint behaviour, `podman run` should be used.

Fixes #607

Signed-off-by: pixdrift <support@pixeldrift.net>

PR #6714, opened 2026-03-10
tests: some more storage.conf rewrite prep

Luap99:
#### What this PR does / why we need it:

More preparation work for https://github.com/containers/container-libs/pull/680, I like to get them merged in advance so the container-libs PR will not break the vendor for others once it gets merged.
Commits were tested in https://github.com/containers/buildah/pull/6708 already.

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #6787, opened 2026-04-14
Update to use shared configfile implementation

jankaluza:
#### What type of PR is this?

/kind other

Update to use shared configfile implementation
This updates Buildah to use the new configfile api from
container-libs https://github.com/containers/buildah/pull/611.

```release-note

```


mtrmac:
(I have added “No new tests” to let the CI run, and switched the PR to draft, correspondingly. If this PR does end up making changes to the repo, please consider removing the label.)

internal/mkcw/embed/entrypoint_amd64.gz
mtrmac, line 1: Unrelated to this PR: Cc: @lsm5 … will we continue to have churn here?

jankaluza:
I've checked the test failures and both `not ok 858 push with authfile` and `not ok 570 chroot with overlay root` fail even without this patch applied. I therefore think it is not caused by any change in the container-libs

mtrmac:
> I've checked the test failures and both `not ok 858 push with authfile`

`Error: preparing instances for copy: requested operation filtered out all platforms and would create an empty image` should be resolved by using a newer c/image — please rebase your PR.

>  and `not ok 570 chroot with overlay root` fail 

`# Error: open /root/.config/containers/policy.json: permission denied`

Really, there is no relationship to the c/image `policy.json` work here?!

> even without this patch applied.

To confirm, is that with the latest `container-libs`, or with the base of the 711 PR (almost a month old now)?

---

If we have CI failing on current `main`, I think that’s an ~urgent situation and there should be $someone working on that or at least an issue filed. Admittedly I do see some failures on `main`, but not those, and a recent https://github.com/containers/buildah/commit/41c29785613693fd6cd2f5a68e4c27b3d89db559 did pass all tests.

So I’d expect _some_ action to be taken based on those failures — either initiating a fix process on `main`, or changing this testing PR.

Luap99:
> > and `not ok 570 chroot with overlay root` fail
> 
> `# Error: open /root/.config/containers/policy.json: permission denied`
> 
> Really, there is no relationship to the c/image `policy.json` work here?!
> 
> > even without this patch applied.
> 
> To confirm, is that with the latest `container-libs`, or with the base of the 711 PR (almost a month old now)?
> 


This is almost certainly caused by this PR, other config files were no different https://github.com/containers/buildah/pull/6714/changes/12bbaf596e70297ec80fc64c679d66ebb075540a

This buildah test sets up a custom chroot without access to /root, so this need a env var or cli level overwrite, i.e. may buildah test seem to do that with `$WITH_POLICY_JSON` already

jankaluza:
OK, lesson learned, `make test-integration` does not run `make` automatically. I've been testing this with wrong binary :-(.

jankaluza:
Well, the `push.bats` part of https://github.com/containers/buildah/pull/6787/changes/796009cf9401c64fc1d0d31fd6e9464e4074c0f5#diff-bd63e6752b89b04d7ed374634cb8adbb540561d29c2a46baf4bab11edaab3235 looks weird.

It is something which should have been fixed by @mtrmac 's https://github.com/containers/container-libs/pull/713. I see the same issues with latest container-libs in the podman PR: https://github.com/containers/podman/pull/28508.

It's probably yet another issue on my side, I will investigate this more...

jankaluza:
> Well, the `push.bats` part of [796009c#diff-bd63e6752b89b04d7ed374634cb8adbb540561d29c2a46baf4bab11edaab3235](https://github.com/containers/buildah/commit/796009cf9401c64fc1d0d31fd6e9464e4074c0f5#diff-bd63e6752b89b04d7ed374634cb8adbb540561d29c2a46baf4bab11edaab3235) looks weird.
> 
> It is something which should have been fixed by @mtrmac 's [containers/container-libs#713](https://github.com/containers/container-libs/pull/713). I see the same issues with latest container-libs in the podman PR: [containers/podman#28508](https://github.com/containers/podman/pull/28508).
> 
> It's probably yet another issue on my side, I will investigate this more...

Yeah, I updated to latest image commit after I did that fix. Should be OK now :-).

Luap99:
Please add some proper commit message for the fix tests commit when you rebase here and pick up the changes from main

internal/mkcw/embed/entrypoint_amd64.gz
lsm5: I *hope* not. @jankaluza could you please rebase on the latest main (along with restoring this file from main) . I suspect once that's done it shouldn't be rebuilt by make.

jankaluza: Rebase fixed it.

Luap99 (approved):
LGTM

jankaluza:
Tests passed :-).

mtrmac (approved):
LGTM, please merge whenever appropriate.

PR #6807, opened 2026-04-22
copier: Fix the bookkeeping of the requested root

BenjaminSchubert:
#### What type of PR is this?

> /kind bug

Previously, we would never update the previousRequestRoot after chrooting, which would lead the copier to try to chroot another time inside the earlier chroot, which is not supported.

This ensures that we keep the proper root over iterations and fail properly if another root is requested.

#### How to verify it

See the test, changing the previousRoot back and running the test will fail, claiming that the `testDir` doesn't exist.

I have not created an issue first for this, do you want me to?

I am not sure here how a user of buildah would directly hit this

```release-note
None
```

### DIFF
--- a/copier/copier_test.go
+++ b/copier/copier_test.go
@@ -4,6 +4,7 @@ import (
 	"archive/tar"
 	"bufio"
 	"bytes"
+	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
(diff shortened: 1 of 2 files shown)


nalind:
Right now I don't think we ever send more than one request down the pipe, but this LGTM.

copier/copier_test.go
```diff
@@ -2662,3 +2663,50 @@ func TestTarPut(t *testing.T) {
...
+
+	req := request{Request: requestEval, Root: testDir, Directory: "/"}
+	var resp response
+
+	require.NoError(t, encoder.Encode(&req), "failed to send first request to copier")
+	require.NoError(t, stdoutR.SetReadDeadline(time.Now().Add(time.Second*5)))
+	require.NoError(t, decoder.Decode(&resp), "failed to decode response from copier")
+	require.Empty(t, resp.Error, "first request returned an error: %s", resp.Error)
+
+	require.NoError(t, encoder.Encode(&req), "faild to send second request to copier")
```
nalind: Typo: "faild"

BenjaminSchubert: fixed!

TomSweeneyRedHat:
LGTM

PR #6808, opened 2026-04-22
copier: Fix some log messages

BenjaminSchubert:
#### What type of PR is this?

> /kind cleanup

- The error when checking if directory is an absolute path doesn't log the right directory
- The incorrect size read inverted the read vs expected bytes
- The fileUID and fileGID logs logged the pointer to the values instead of the values themselves

#### How to verify it

N/A

None

```release-note
None
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -566,7 +566,7 @@ func convertToRelSubdirectory(root, directory string) (relative string, err erro
 		return "", fmt.Errorf("expected root directory to be an absolute path, got %q", root)
 	}
 	if directory == "" || !filepath.IsAbs(directory) {
-		return "", fmt.Errorf("expected directory to be an absolute path, got %q", root)
+		return "", fmt.Errorf("expected directory to be an absolute path, got %q", directory)
 	}
 	if filepath.VolumeName(root) != filepath.VolumeName(directory) {
 		return "", fmt.Errorf("%q and %q are on different volumes", root, directory)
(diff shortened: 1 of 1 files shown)


BenjaminSchubert:
@containers/buildah-maintainers, I don't think I can easily write tests for those/that its worth having tests

copier/copier.go
```diff
@@ -1828,7 +1828,7 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
 			containerFilePair := idtools.IDPair{UID: *fileUID, GID: *fileGID}
 			hostFilePair, err := idMappings.ToHost(containerFilePair)
 			if err != nil {
-				return errorResponse("copier: put: error mapping container filesystem owner %d:%d to host filesystem owners: %v", fileUID, fileGID, err)
+				return errorResponse("copier: put: error mapping container filesystem owner %d:%d to host filesystem owners: %v", &fileUID, &fileGID, err)
```
giuseppe: this needs to be `*fileUID` and `*fileGID`

BenjaminSchubert: woops, good catch.. Fixed

giuseppe (approved):
LGTM

nalind:
LGTM, thanks!

TomSweeneyRedHat:
LGTM
Thanks for the fixes @BenjaminSchubert !

PR #6811, opened 2026-04-23
Makefile: preserve entrypoint_amd64 and .gz in clean target

lsm5:
The entrypoint_amd64.gz file is committed to the repository (built with native assembler, 362 bytes) and should not be deleted by 'make clean'. Also preserve the entrypoint_amd64 binary to prevent unnecessary rebuilds.

Only delete the other architecture binaries and .gz files which are pure build artifacts.

This prevents unnecessary rebuilds and git diffs after make clean.

#### What type of PR is this?

/kind other

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,7 +132,7 @@ bin/grpcnoop: tests/rpc/noop/noop.go
 
 .PHONY: clean
 clean:
-	$(RM) -r bin tests/testreport/testreport tests/conformance/testdata/mount-targets/true internal/mkcw/embed/entrypoint_amd64 internal/mkcw/embed/entrypoint_arm64 internal/mkcw/embed/entrypoint_ppc64le internal/mkcw/embed/entrypoint_s390x internal/mkcw/embed/*.gz internal/mkcw/embed/asm/*.o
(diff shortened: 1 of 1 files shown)


lsm5:
@nalind PTAL

nalind (approved):
LGTM

lsm5:

Makefile
```diff
@@ -132,7 +132,7 @@ bin/grpcnoop: tests/rpc/noop/noop.go
 
 .PHONY: clean
 clean:
-	$(RM) -r bin tests/testreport/testreport tests/conformance/testdata/mount-targets/true internal/mkcw/embed/entrypoint_amd64 internal/mkcw/embed/entrypoint_arm64 internal/mkcw/embed/entrypoint_ppc64le internal/mkcw/embed/entrypoint_s390x internal/mkcw/embed/*.gz internal/mkcw/embed/asm/*.o
+	$(RM) -r bin tests/testreport/testreport tests/conformance/testdata/mount-targets/true internal/mkcw/embed/entrypoint_arm64 internal/mkcw/embed/entrypoint_ppc64le internal/mkcw/embed/entrypoint_s390x internal/mkcw/embed/entrypoint_arm64.gz internal/mkcw/embed/entrypoint_ppc64le.gz internal/mkcw/embed/entrypoint_s390x.gz internal/mkcw/embed/asm/*.o
```
Luap99, line 135: not sure what shells we support but with bash we have the nice option of the `{}` syntax, e.g. rm `internal/mkcw/embed/entrypoint_{arm64,ppc64le,s390x}{,.gz}` (untested)

Luap99 (approved):
LGTM, there might be potential for simplification with the bash expansion but I am not sure how well supported that is everywhere.

PR #6271, opened 2025-07-08
Update "bud with --cpu-shares" test, and rename it

nalind:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

Update "the bud with --cpu-shares" test to expect the a cgroupsv2 value computed using either the older formula or the newer one introduced in github.com/opencontainers/cgroups v0.0.3, and give it a unique name so that it can be selected more easily with bats's "--filter" flag.

#### How to verify it

Updated test!

#### Which issue(s) this PR fixes:

Fixes #6270 

#### Special notes for your reviewer:

I thought the test errors in #6237 were flakes... nope!

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #4610, opened 2023-02-20
Replace Ubuntu -> Debian SID

cevich:
#### What type of PR is this?

/kind other

#### What this PR does / why we need it:

Replace use of Ubuntu with Debian SID CI VM Images.

#### How to verify it

CI will pass

#### Which issue(s) this PR fixes:

None

#### Special notes for your reviewer:

Ref: https://github.com/containers/automation_images/pull/250

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #6813, opened 2026-04-27
[release-1.26] Fix CVE-2025-47913 with x/crypto + minimum go bump

cevich:
#### What type of PR is this?

/kind other

Resolves https://issues.redhat.com/browse/RHEL-134784
Resolves https://issues.redhat.com/browse/RHEL-168611
    
Also bump CI image to gain access to (required) golang 1.24.

```release-note
Fix CVE-2025-47913 with x/crypto + minimum go bump
```


go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/containers/buildah
 
-go 1.22.6
+go 1.24.0
 
-toolchain go1.22.8
+toolchain go1.24.10
```
nalind: Given that we try to avoid having this line pop up in later branches, I'm surprised to see it here.

nalind (commented):
#6678 may help with one of the errors we're seeing in the unit tests.
#6271 and #6674 should help with one of the errors we're seeing in integration tests.
Some of the changes from https://github.com/containers/buildah/pull/4610/changes#diff-dcdc10b43f652991a881c6d816518d2f1b52fe22f34abb9103f90528dff35a68 appear to be indicated for errors we're seeing in conformance tests.

go.mod
cevich: > ...may help...

Thanks so much for all the pointers! I'll get started tracking them down and applying.

edit: Fixed- All backported, and w/ selections from fa145b0

---

> ...avoid having this line...

I won't pretend to understand why, but I tried removing it then `make vendor` just adds it back in :disappointed:

IIRC, there might be an environment variable that controls it?  If so, I'd hazard to guess I'm not setting it :grin: Sounds like I should?

Edit: Fixed - manually removed line and now it's not re-appearing, so :shrug: 

---

FWIW, results from running system tests on RHEL (I haven't looked at them yet for comparison).

- [86147cc-gating:local_root_system_runc:root.txt](https://github.com/user-attachments/files/27137059/86147cc-gating.local_root_system_runc.root.txt)
- [86147cc-gating:local_root_system_crun:root.txt](https://github.com/user-attachments/files/27137060/86147cc-gating.local_root_system_crun.root.txt)

nalind: Try setting "go 1.24.10" in go.mod ("go mod edit -go=1.24.10" should also work).  `make vendor-in-container` shouldn't reintroduce a "toolchain" directive then.

There seem to be a few missing test helper binaries in the installed buildah-tests package (buildah-inet, buildah-dumpspec, buildah-tutorial) and a missing "DUMPSPEC_BINARY=/usr/bin/buildah-dumpspec" in the gating environment.  I don't see the corresponding .spec file in the source tree, though, so I'm not sure what to suggest.

Does the gating test environment have "crun" installed?  Or a version that doesn't recognize a "--debug" flag?

The SELinux test in the gating environment may need https://github.com/containers/buildah/pull/2125.

cevich: > Try setting

I hand edited it then ran `make vendor-in-container` and it did not re-add the toolchain directive, so I'm not sure why I was struggling with it before.

> There seem to be a few missing test helper binaries 

Yes (and good catch). This unfortunately has been a problem across multiple releases.  We've got issues open on it, and I've already got the specfile fixes in place for this version.

> Does the gating test environment have "crun" installed?

Ahh! Another good catch, indeed it's unspecified so may not have been installed.  I've fixed this so it's explicit now.

> The SELinux test in the gating environment

Thanks for the pointer, we see this for other releases as well but have been ignoring it.  I wasn't aware it had a fix.  I'll add a reference to our notes for the other releases, and cherry-pick the fix here.

nalind: The cherry-pick from 595af5038d125d7b79ff112171f9ab5d834da708 appears to have pulled in unit tests that weren't in the original commit.

contrib/cirrus/setup.sh
```diff
@@ -60,8 +60,8 @@ EOF
             # need to run dnsmasq on port 53.
             if [[ -r "/run/systemd/resolve/resolv.conf" ]]; then
                 msg "Disabling systemd-resolved service"
-                systemctl stop systemd-resolved.service
-                cp /run/systemd/resolve/resolv.conf /etc/
+                systemctl stop systemd-resolved.service systemd-resolved-monitor.socket, systemd-resolved-varlink.socket
```
nalind: That comma appears to be causing an early error in the conformance tests.

cevich: Oof, bad copy-paste.  Good catch.

go.mod
cevich: Yeah I generally default to `--theirs` to make future backports easier but it may have caused more trouble here than it resolved.  I'll come back and strip out the tag-a-longs if this continues to be the case.

cevich:
Ugh now that the conformance tests are running, [they're throwing](https://cirrus-ci.com/task/5626170756038656):

`Error response from daemon: client version 1.43 is too old. Minimum supported API version is 1.44, please upgrade your client to a newer version`.

This is surprising given I've updated the environment to Debian 14 :confused: Or is this some client version **buildah** is reporting? I'm confused.

go.mod
cevich: The "extra" tests seem to be passing fine, though the copier `TestGetSingleChroot` still seems to be [flaking on broken pipes](https://cirrus-ci.com/task/5063220802617344?logs=unit_test#L3154).  I'm fine with re-running or ignoring this as it seems to be a test bug not a buildah bug(?).

nalind:
> Ugh now that the conformance tests are running, [they're throwing](https://cirrus-ci.com/task/5626170756038656):
> 
> `Error response from daemon: client version 1.43 is too old. Minimum supported API version is 1.44, please upgrade your client to a newer version`.
> 
> This is surprising given I've updated the environment to Debian 14 😕 Or is this some client version **buildah** is reporting? I'm confused.

That's the conformance test client's half of the API version negotiation not matching up with what the docker daemon is offering.  https://github.com/containers/buildah/pull/6602/changes/8d6f3fcf126ba4c1ae75abeec4f6c04e9ced97ac is what we used on main.

cevich:
> That's the conformance test client's half of the API version negotiation not matching up with what the docker daemon is offering. [8d6f3fc](https://github.com/containers/buildah/commit/8d6f3fcf126ba4c1ae75abeec4f6c04e9ced97ac) is what we used on main.

Cherry-picking that didn't seem to result in any conformance test change.  Could it be that the vendored code is "too old"?  `go.mod` shows:

```
	github.com/docker/distribution v2.8.3+incompatible
	github.com/docker/docker v24.0.7+incompatible
	github.com/docker/go-units v0.5.0
```

cevich:
Answering my own question: "Yes", it seems an upgrade of `github.com/docker/docker` to v25 or later is required to get 1.44.  However, doing this snowballs into also requiring a golang bump up to 1.25 which isn't available in the CI images.

I'm informed another way around this is to set `DOCKER_API_VERSION=1.44`, which could break things later.  However, by the time "later" comes there's a good chance RHEL won't care about this branch any longer.

cevich:
Ugh, I tried, but no matter where I `export DOCKER_API_VERSION`, it's not getting picked up or is being masked going into the conformance tests.

I tried [8d6f3fc](https://github.com/containers/buildah/commit/8d6f3fcf126ba4c1ae75abeec4f6c04e9ced97ac) but as stated above, this won't work without a bump to the docker module version.

Just now I've made an attempt at updating the ci vm images to support golang 1.25, bumped the docker module etc, but gave up.  It really just wants to snowball into bigger and bigger changes, so this seems like a dead-end.

The goal of this PR is just to update the x/crypto module version.  The targeted version of RHEL goes EOL at the end of May, and the CI system shuts down completely in June.  So I really don't care to waste more time grinding away at this CI-only problem.

I'm going to simply disable the conformance tests, and then we :crossed_fingers: nothing in the next month or so would benefit from them...

nalind:
> Ugh, I tried, but no matter where I `export DOCKER_API_VERSION`, it's not getting picked up or is being masked going into the conformance tests.

The conformance tests also use containers/image to read images in the docker daemon, and that would negotiate with, at the latest, the version that the corresponding docker client libraries listed as the most recent, which is 1.43 in the current tree, so that's not going to work so well.  As you've noted, you'd have to update to 25.x to get something that claimed to be 1.44 or later.

cevich:
> also use containers/image to read images in the docker daemon

Ahh, I vaguely remember seeing other erors about an even older unsupported client, so that's probably the explanation there.

Yeah, I don't see any practical way around this. Though I'll be open to suggestions tomorrow once my blood pressure goes back down :grin:

cevich:
@nalind @TomSweeneyRedHat PTAL when you get a chance.

Note: Validate is likely broken due to use of EOL OS.

nalind:
> Note: Validate is likely broken due to use of EOL OS.

Indeed I don't see it listed at https://github.com/actions/runner-images.  FWIW the version of `.github/workflows/pr.yml` on main looks more or less the same give or take versions.

nalind:
> Yeah, I don't see any practical way around this. Though I'll be open to suggestions tomorrow once my blood pressure goes back down 😁

Later versions of dockerd relaxed this minimum down to 1.40, but if that's not an option, a
```bash
echo '{"min-api-version":"1.40"}' > /etc/docker/daemon.json
```
before the daemon starts might override the requirement on its end.

copier/copier.go
```diff
@@ -49,6 +50,80 @@ func init() {
...
+		theseMatches, err := filepath.Glob(pattern)
+		if err != nil {
+			return nil, err
+		}
+		matches = append(matches, theseMatches...)
+	}
+	sort.Strings(matches)
+	return matches, nil
+}
+
```
nalind: If we update, this should be dropped from the backport.

cevich: I'm assuming you're meaning drop this entire function along with `TestSortedExtendedGlob()` in `copier_test.go` 👍

cevich:
Force push:

1. Stripped out copier/copier.go `extendedGlob()` function + test as requested.
2. Added back the conformance tests.
3. Added conformance test setup step to force acceptance of client versions >= 1.40

copier/copier.go
cevich: Ref: Old commit was 08dfb828b141193ada4a1a6a8ab893e35476252a

nalind:
Looks like the DOCKER_API_VERSION environment variable is still going to be needed for the go-dockerclient bits in the conformance tests.

cevich:
> the DOCKER_API_VERSION environment variable

I couldn't get that to work either.  Tried setting it in several spots, all with the same result.  As much as I hate disabling tests, I still think that's the right thing to do here.  This branch is useless to RHEL in one month, right before the old CI system goes offline anyway.

cevich:
I was reminded this morning that we're potentially not completely finished with RHEL in a month, as 8.6 still utilizes this branch :confounded: Regardless, there is time pressure to get this fix out as soon as practical, and (IMHO) fixing the conformance tests (somehow) doesn't seem (to me) practical.  However, I'm willing to be ruled "wrong" on all accounts :grin:

cevich:
Ha! You got it!  Both changes were needed, where they were failing individually.  Brilliant!

nalind:

lsm5 (approved):
LGTM

The validate commit failed but I guess it's a RHEL branch thing so whatevs..

PR #6821, opened 2026-04-29
[release-1.24] CI: force use of registry:2 instead of registry:latest

nalind:
#### What type of PR is this?

/kind failing-test 

Set REGISTRY_FQIN in the CI configuration to "registry:2" so that the default, "registry:latest", won't be set up before our tests, as it expects its configuration to be done differently.

#### How to verify it

Smoke test should pass now.

None

```release-note
None
```


TomSweeneyRedHat:
LGTM

PR #6827, opened 2026-04-30
copier: add RemoveOptions.AllowWildcard

### LINKED ISSUE #6806: Add support for `Op_File.Actions.FileAction_Rm.AllowWildcard`
Tracks the `Op_File.Actions.FileAction_Rm.AllowWildcard` item from #6732.

- [ ] Add `AllowWildcard` to `copier.RemoveOptions` (type: `bool`) // glob for items to remove, default false
- [ ] Add unit test to the copier package

akca:
#### What type of PR is this?

/kind feature

Adds `AllowWildcard` as a boolean field to `copier.RemoveOptions`. When set, the path is expanded as a glob pattern and each match is removed. This aligns buildah's copier with BuildKit's `FileActionRm.AllowWildcard` behavior.

#### How to verify it

```
go test ./copier -run TestRemove -v
```

New test cases cover: `*`, `?`, `[ab]` patterns, zero matches, interaction with `All` and `AllowNotFound` flags.

Fixes #6806

```release-note
copier: add `RemoveOptions.AllowWildcard` to support glob patterns in remove operations.
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -520,6 +520,7 @@ func Mkdir(root string, directory string, options MkdirOptions) error {
 type RemoveOptions struct {
 	All           bool // if Directory is a directory, remove its contents as well
 	AllowNotFound bool // don't return an error if the item is already not present
+	AllowWildcard bool // expand the path as a glob pattern, removing each match. All must be set to remove matched non-empty directories
 }
 
 // Remove removes the specified directory or item, traversing any intermediate
@@ -2297,20 +2298,30 @@ func copierHandlerRemove(req request) *response {
 	errorResponse := func(fmtspec string, args ...any) *response {
 		return &response{Error: fmt.Sprintf(fmtspec, args...), Remove: removeResponse{}}
 	}
-	resolvedTarget, err := resolvePath(req.Root, req.Directory, false, nil)
-	if err != nil {
-		return errorResponse("copier: remove: %v", err)
-	}
-	if req.RemoveOptions.All {
-		err = os.RemoveAll(resolvedTarget)
-	} else {
-		err = os.Remove(resolvedTarget)
-		if req.RemoveOptions.AllowNotFound && errors.Is(err, os.ErrNotExist) {
-			err = nil
+	targets := []string{req.Directory}
+	if req.RemoveOptions.AllowWildcard {
+		var err error
+		targets, err = extendedGlob(req.Directory)
+		if err != nil {
+			return errorResponse("copier: remove: glob %q: %v", req.Directory, err)
 		}
 	}
-	if err != nil {
-		return errorResponse("copier: remove %q: %v", req.Directory, err)
+	for _, target := range targets {
+		resolvedTarget, err := resolvePath(req.Root, target, false, nil)
+		if err != nil {
+			return errorResponse("copier: remove: %v", err)
+		}
+		if req.RemoveOptions.All {
+			err = os.RemoveAll(resolvedTarget)
+		} else {
+			err = os.Remove(resolvedTarget)
+			if req.RemoveOptions.AllowNotFound && errors.Is(err, os.ErrNotExist) {
+				err = nil
+			}
+		}
+		if err != nil {
+			return errorResponse("copier: remove %q: %v", target, err)
+		}
 	}
 	return &response{Error: "", Remove: removeResponse{}}
 }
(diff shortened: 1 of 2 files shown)


TomSweeneyRedHat:
I have restarted the failed check, which didn't appear related to these changes.

TomSweeneyRedHat:
LGTM
@nalind PTAL

copier/copier_test.go
```diff
@@ -2095,6 +2096,73 @@ func testRemove(t *testing.T) {
...
+				{
+					name:          "wildcard-literal-path",
+					remove:        "subdir-a/file-a",
+					allowWildcard: true,
+					removed:       []string{"subdir-a/file-a"},
+				},
+				{
+					name:          "wildcard-literal-missing",
+					remove:        "subdir-a/file-nonexistent",
+					allowWildcard: true,
```
nalind, line 2152: Without `allowNotFound` and `all` set (i.e., both `false`), should an error be returned, causing this test to fail because `fail` wasn't set?

copier/copier_test.go
```diff
@@ -2095,6 +2096,73 @@ func testRemove(t *testing.T) {
...
+						"subdir-a/subdir-b/subdir-c",
+						"subdir-a/subdir-b/subdir-c/parent",
+						"subdir-a/subdir-b/subdir-c/link-b",
+						"subdir-a/subdir-b/subdir-c/root",
+					},
+				},
+				{
+					name:          "wildcard-no-match",
+					remove:        "subdir-a/nonexistent-*",
+					allowWildcard: true,
```
nalind, line 2120: Without `allowNotFound` and `all` set (i.e., both `false`), should an error be returned, causing this test to fail because `fail` wasn't set?

copier/copier.go
```diff
@@ -2297,20 +2298,30 @@ func copierHandlerRemove(req request) *response {
...
-		err = os.Remove(resolvedTarget)
-		if req.RemoveOptions.AllowNotFound && errors.Is(err, os.ErrNotExist) {
-			err = nil
+	targets := []string{req.Directory}
+	if req.RemoveOptions.AllowWildcard {
+		var err error
+		targets, err = extendedGlob(req.Directory)
+		if err != nil {
+			return errorResponse("copier: remove: glob %q: %v", req.Directory, err)
 		}
```
nalind, line 2307: I think we need to be catching cases where the returned set of targets has `len() == 0` when `AllowNotFound` isn't set, and returning an error, as `Glob()` will only return an error for an invalid pattern.
`AllowWildcard` doesn't always mean that there's a wildcard being used, and I don't know that we want to effectively ignore not-found errors when it's set.

nalind (commented):
Looks about right, though I'm not clear on whether the cases where the wildcard doesn't match anything should expect to get an error back.  Do we know what the expected behavior there is?

akca:
> Looks about right, though I'm not clear on whether the cases where the wildcard doesn't match anything should expect to get an error back. Do we know what the expected behavior there is?

Checked BuildKit's implementation of the same flag ([here](https://github.com/moby/buildkit/blob/c7ba94199885e89e25615d3e2d78fea86a2c7eb7/solver/llbsolver/file/backend.go#L136-L162)). When `AllowWildcard` is set and the pattern matches nothing, it silently succeeds. `AllowNotFound` doesn't affect that; it only controls whether individually resolved paths that don't exist are an error.

Test cases here expecting success on zero matches follow that same behaviour. @nalind what do you think?

nalind (approved):
> > Looks about right, though I'm not clear on whether the cases where the wildcard doesn't match anything should expect to get an error back. Do we know what the expected behavior there is?
> 
> Checked BuildKit's implementation of the same flag ([here](https://github.com/moby/buildkit/blob/c7ba94199885e89e25615d3e2d78fea86a2c7eb7/solver/llbsolver/file/backend.go#L136-L162)). When `AllowWildcard` is set and the pattern matches nothing, it silently succeeds. `AllowNotFound` doesn't affect that; it only controls whether individually resolved paths that don't exist are an error.

I think that's a wrinkle that would be great to note in the godoc, if only for future-me's benefit.  Not a blocker, though.

> Test cases here expecting success on zero matches follow that same behaviour. @nalind what do you think?

That's the use case we want to be compatible with, so the tests are good as-is.

LGTM

nalind:

TomSweeneyRedHat (approved):
LGTM

PR #6791, opened 2026-04-16
CI: remove dependencies on online apt repositories

nalind:
#### What type of PR is this?

/kind failing-test 

#### What this PR does / why we need it:

Rework places to avoid installing packages using apt as part of tests. Use ubi10 where we need to be able to call setcap, setfattr, and getfattr from the same image.

Test CI images now pre-cache the hummingbird git image, so we no longer need to it to be pulled during CI startup.

#### How to verify it

Updated tests!

#### Special notes for your reviewer:

Addresses unrelated test errors in https://github.com/containers/buildah/pull/6722.

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #6831, opened 2026-05-04
[release-1.43] Cite go module change

baude:
The go modules for buildah is changing and we want a deprecation notice in the current n-1 branch.

#### What type of PR is this?

/kind other

```release-note
None
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,4 @@
+// Deprecated: The buildah go module has been moved to go.podman.io/buildah
 module github.com/containers/buildah
 
 // Warning: Ensure the "go" and "toolchain" versions match exactly to prevent unwanted auto-updates
--- a/tests/NEW-IMAGES
+++ b/tests/NEW-IMAGES
@@ -13,3 +13,4 @@
 #
 # Format is one FQIN per line. Enumerate them below:
 #
+registry.access.redhat.com/ubi10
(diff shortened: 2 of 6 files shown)


baude:
cc @Luap99

Luap99:
Thanks, looks like we need to pick https://github.com/containers/buildah/pull/6791 here as well cc @nalind

baude:
i pulled in the commit you cited @Luap99 ... it looks like one optional test failing now ?

Luap99 (approved):
LGTM

cc @nalind @mtrmac

mtrmac (approved):
LGTM. (For the test changes, I have only checked that they are a faithful match to the changes in `main`.)

nalind:
Are we sure about doing this on older branches where the module name change isn't in the branch?

Luap99:
> Are we sure about doing this on older branches where the module name change isn't in the branch?

Well we cannot deprecate the new name on main. As long as we at some point publish a new 1.43 version then the deprecation notice is published via the tag and should be able to get picked up by consumers who update to the latest tag.

mtrmac:
- This only matters to Go code that imports Buildah; others don’t see the deprecation comment.
- Code importing a _currently-released_ Buildah 1.43.x sees no change, this didn’t happen there yet
- Code importing a `go.podman.io/buildah` path doesn’t need to do anything any more, and does not have the deprecation warning
- Code importing a _future_ Buildah 1.43.(x+1) will see the deprecation warning and be directed to upgrade.

Admittedly the timing of 1.43.(x+1) vs. 1.44.0 is not quite precise, due to an urgent fix we might end up releasing 1.43.(x+1) without having a 1.44.0 ready yet.

nalind:
So long as consumers like podman's 5.8 branch, which is following this one, don't get confused.

PR #6832, opened 2026-05-06
[WIP][DO NOT MERGE] test vendoring for https://github.com/containers/container-libs/pull/734

giuseppe:
#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```


giuseppe:
the failure seems related to the change, need to investigate it

tests/namespaces.bats
```diff
@@ -437,6 +437,33 @@ _EOF
   done
 }
 
+@test "idmapping-non-contiguous-maps" {
```
Luap99, line 440: unless I a missing something that test passes on main so this is not a regression test of the bug reported?!

Luap99 (changes requested):

tests/namespaces.bats
giuseppe: are you using overlay as storage driver?

Luap99: yes

Luap99: or at least I think I do not sure if the test overwrite anything, I just do `sudo bats -f "idmapping-non-contiguous-maps" tests/namespaces.bats` with a buildah build on main not of this branch

giuseppe: you need to set STORAGE_DRIVER=overlay, can you try the following command?

```
sudo STORAGE_DRIVER=overlay bats -f "idmapping-non-contiguous-maps" tests/namespaces.bats
```

Luap99: ah yes that fails, sorry I was not aware that vfs is the test default locally.
```
   Error: building at STEP "RUN useradd -u 579841 -g 67283 -s /bin/bash -m crunbuild": ensuring mount targets for container "28c00debdd6d34ea47a9652df9342ea94aee72f7a786d936fe558c145d7a843c": copier: ensure: error checking item "dev/mqueue": mkdir /dev/mqueue: permission denied
```
I assume that is the expected error before your patch?

Luap99 (approved):

tests/namespaces.bats
giuseppe: yes, that is the error I am seeing too

PR #6834, opened 2026-05-07
[release-1.41] tests: remove dependencies on online apt repositories

nalind:
#### What type of PR is this?

/kind failing-test 

Rework places to avoid installing packages using apt as part of tests. Use ubi10 where we need to be able to call setcap, setfattr, and getfattr from the same image.

#### How to verify it

Updated integration tests!

None

Manual cherry pick from #6791.

```release-note
None
```

### DIFF
--- a/tests/NEW-IMAGES
+++ b/tests/NEW-IMAGES
@@ -13,3 +13,5 @@
 #
 # Format is one FQIN per line. Enumerate them below:
 #
+quay.io/hummingbird/git
+registry.access.redhat.com/ubi10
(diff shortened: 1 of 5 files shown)


lsm5 (approved):
TF jobs fail on not finding grpcnoop. I guess this would be a trivial fix (assuming that binary can be built / fetched). Can be handled in the move  away from Cirrus.

LGTM

TomSweeneyRedHat:
If we want to move forward with this and the other test PRs, we'll at least need to duplicate in GitLab.

PR #6835, opened 2026-05-07
[release-1.40] tests: remove dependencies on online apt repositories

nalind:
#### What type of PR is this?

/kind failing-test 

Rework places to avoid installing packages using apt as part of tests. Use ubi10 where we need to be able to call setcap, setfattr, and getfattr from the same image.

#### How to verify it

Updated integration tests!

None

Manual cherry pick from #6834.

```release-note
None
```

### DIFF
--- a/tests/NEW-IMAGES
+++ b/tests/NEW-IMAGES
@@ -13,3 +13,5 @@
 #
 # Format is one FQIN per line. Enumerate them below:
 #
+quay.io/hummingbird/git
+registry.access.redhat.com/ubi10
--- a/tests/chroot.bats
+++ b/tests/chroot.bats
@@ -22,7 +22,7 @@ load helpers
   # chosen because its rootfs doesn't have any uid/gid ownership above
   # $rangesize, because the nested namespace needs to be able to represent all
   # of them
-  baseimage=registry.access.redhat.com/ubi9-micro:latest
+  baseimage=registry.access.redhat.com/ubi10:latest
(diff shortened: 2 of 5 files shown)


lsm5 (approved):
TF jobs fail on not finding grpcnoop and passwd. Probably not worth blocking over until the move away from Cirrus.

nalind:
> TF jobs fail on not finding grpcnoop and passwd. Probably not worth blocking over until the move away from Cirrus.

Neither of those exist on this branch, and moving from Cirrus to something else isn't going to change that.

nalind:
Am I reading https://dashboard.packit.dev/jobs/testing-farm/1535319 right?  That it's building packages off of the PR for the Copr Build(s), and then installing packages built from the main branch and running the tests that are in that package using the binaries from that package, using the TMT configuration (tests/tmt directory) from this PR?  If so, what is that supposed to be verifying?

lsm5:
ugh, yes, the buildah rpm is fetched from podman-next, which is to  blame. Thanks @nalind . And that's because of the super-high Epoch of podman-next packages.

So we can ignore all testing-farm jobs here. Should be fixed when the packit and TMT config is changed.

nalind:
If it's known that they're going to fail, can we remove them from older branches until that's resolved?

lsm5:
> If it's known that they're going to fail, can we remove them from older branches until that's resolved?

sure thing. on it..

PR #6843, opened 2026-05-11
add/copy: support AllowWildcard and AllowEmptyWildcard

### LINKED ISSUE #6828: Add support for `Op_File.Actions.FileAction_Copy.AllowWildcard`
Tracks the `Op_File.Actions.FileAction_Copy.AllowWildcard` item from https://github.com/containers/buildah/issues/6732

Add support for controlling whether source paths are expanded as glob patterns during copy operations, aligning with BuildKit's `FileActionCopy.AllowWildcard`. Currently glob expansion is always enabled. This adds an explicit flag so callers can disable it and treat paths as literals.                                                                                     

- [ ] Add `AddAndCopyOptions.AllowWildcard` (type: `types.OptionalBool`), default true                                              
- [ ] Add `copier.StatOptions.AllowWildcard`                  
- [ ] Add `copier.GetOptions.AllowWildcard`                                                                                         
- [ ] Add unit tests in copier package                        
- [ ] Add `--allow-wildcard` to `buildah add` and `buildah copy`, default true                                                      
- [ ] Add integration tests

### LINKED ISSUE #6829: Add support for `Op_File.Actions.FileAction_Copy.AllowEmptyWildcard`
Tracks the `Op_File.Actions.FileAction_Copy.AllowEmptyWildcard` item from https://github.com/containers/buildah/issues/6732

Add support for controlling whether zero glob matches is treated as an error during copy operations, aligning with BuildKit's `FileActionCopy.AllowEmptyWildcard`. This adds a flag to allow empty results to succeed silently.                                                                                                            

Tasks:
- Add `AddAndCopyOptions.AllowEmptyWildcard` (type: `types.OptionalBool`), default true
- Add `copier.StatOptions.AllowEmptyWildcard`
- Add `copier.GetOptions.AllowEmptyWildcard`                                                                                    
- Add unit tests in copier package                                                                                              
- Add `--allow-empty-wildcard` to `buildah add` and `buildah copy`, default false                                               
- Add integration tests

akca:
#### What type of PR is this?

/kind feature

Adds `AllowWildcard` and `AllowEmptyWildcard` to the copier's stat/get operations and exposes them as `--allow-wildcard` / `--allow-empty-wildcard` on `buildah add` and `buildah copy`, aligning with BuildKit's `FileActionCopy` flags ([ref](https://github.com/moby/buildkit/blob/32850dafb5dd8008cee6c9bf4255fcba621a9f38/solver/llbsolver/file/backend.go#L239-L253)).

`AllowWildcard` controls whether glob patterns are allowed in source paths. When `false`, wildcards are rejected. `buildah add`/`buildah copy` sets this to `true` by default and preserves current behaviour.

`AllowEmptyWildcard` controls what happens when a wildcard source matches no files. By default (`false`), the operation fails if every source comes up empty. With `--allow-empty-wildcard=true`, an empty result is silently accepted. Docker/BuildKit defaults to `true` ([ref](https://github.com/moby/buildkit/blob/8100e1edab6b6d68db4d941ecac9d51f341a8913/frontend/dockerfile/dockerfile2llb/convert_copy.go#L283)), so `COPY nonexistent* /dest/` silently succeeds in Docker but errors in buildah. If there's interest in matching Docker's default, that could be a follow-up PR since it would touch many more places.

#### How to verify it

```
go test ./copier -run Wildcard -v
```

Fixes #6828
Fixes #6829

- ~copier uses plain `bool` for these fields (no value provided = doesn't allow globs). Existing callers in buildah and podman all pass resolved paths without wildcard characters, so they should be unaffected. `image.go` is the only exception (exclusion patterns may contain wildcards), so it now explicitly sets `AllowWildcard: true`. Let me know if there are callers I missed.~ Now changed to `DisallowWildcard`, so bool's default value false is the intended value.
- Literal missing paths produce ENOENT. Non-matching globs produce "matched nothing" (not ENOENT, since no specific file was looked up).
- Conformance tests `copy-multiple-some-missing-glob` and `copy-multiple-missing-file-with-nomatch-on-glob` verify the multi-source behaviour matches Docker.

```release-note
Add `--allow-wildcard` and `--allow-empty-wildcard` flags to `buildah add` and `buildah copy`
```

### DIFF
--- a/cmd/buildah/addcopy.go
+++ b/cmd/buildah/addcopy.go
@@ -16,32 +16,35 @@ import (
 	"go.podman.io/buildah/pkg/cli"
 	"go.podman.io/buildah/pkg/parse"
 	"go.podman.io/common/pkg/auth"
+	"go.podman.io/image/v5/types"
 	"go.podman.io/storage"
 )
 
 type addCopyResults struct {
-	addHistory       bool
-	chmod            string
-	chown            string
-	checksum         string
-	quiet            bool
-	ignoreFile       string
-	contextdir       string
-	from             string
-	blobCache        string
-	decryptionKeys   []string
-	removeSignatures bool
-	signaturePolicy  string
-	authfile         string
-	creds            string
-	tlsVerify        bool
-	certDir          string
-	retry            int
-	retryDelay       string
-	excludes         []string
-	parents          bool
-	timestamp        string
-	link             bool
+	addHistory         bool
+	chmod              string
+	chown              string
+	checksum           string
+	quiet              bool
+	ignoreFile         string
+	contextdir         string
+	from               string
+	blobCache          string
+	decryptionKeys     []string
+	removeSignatures   bool
+	signaturePolicy    string
+	authfile           string
+	creds              string
+	tlsVerify          bool
+	certDir            string
+	retry              int
+	retryDelay         string
+	excludes           []string
+	parents            bool
+	timestamp          string
+	link               bool
+	allowWildcard      bool
+	allowEmptyWildcard bool
 }
 
 func createCommand(addCopy string, desc string, short string, opts *addCopyResults) *cobra.Command {
@@ -101,6 +104,8 @@ func applyFlagVars(flags *pflag.FlagSet, opts *addCopyResults) {
 		panic(fmt.Sprintf("error marking signature-policy as hidden: %v", err))
 	}
 	flags.StringVar(&opts.timestamp, "timestamp", "", "set timestamps on new content to `seconds` after the epoch")
+	flags.BoolVar(&opts.allowWildcard, "allow-wildcard", true, "allow glob patterns in source paths")
+	flags.BoolVar(&opts.allowEmptyWildcard, "allow-empty-wildcard", false, "don't error when glob patterns match nothing")
 }
 
 func addcopyInit() {
@@ -277,6 +282,12 @@ func addAndCopyCmd(c *cobra.Command, args []string, verb string, iopts addCopyRe
 		}
 		options.Excludes = append(excludes, options.Excludes...)
 	}
+	if c.Flags().Changed("allow-wildcard") {
+		options.AllowWildcard = types.NewOptionalBool(iopts.allowWildcard)
+	}
+	if c.Flags().Changed("allow-empty-wildcard") {
+		options.AllowEmptyWildcard = types.NewOptionalBool(iopts.allowEmptyWildcard)
+	}
 	if iopts.retryDelay != "" {
 		retryDelay, err := time.ParseDuration(iopts.retryDelay)
 		if err != nil {
@@ -307,6 +318,10 @@ func addAndCopyCmd(c *cobra.Command, args []string, verb string, iopts addCopyRe
 	}
 
 	contentType, digest := builder.ContentDigester.Digest()
+	if digest == "" {
+		logrus.Debug("no content copied, skipping digest and history")
+		return nil
+	}
 	if !iopts.quiet {
 		fmt.Printf("%s\n", digest.Hex())
 	}
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -344,9 +344,11 @@ func Eval(root string, directory string, _ EvalOptions) (string, error) {
 
 // StatOptions controls parts of Stat()'s behavior.
 type StatOptions struct {
-	UIDMap, GIDMap   []idtools.IDMap // map from hostIDs to containerIDs when returning results
-	CheckForArchives bool            // check for and populate the IsArchive bit in returned values
-	Excludes         []string        // contents to pretend don't exist, using the OS-specific path separator
+	UIDMap, GIDMap     []idtools.IDMap // map from hostIDs to containerIDs when returning results
+	CheckForArchives   bool            // check for and populate the IsArchive bit in returned values
+	Excludes           []string        // contents to pretend don't exist, using the OS-specific path separator
+	DisallowWildcard   bool            // reject glob patterns in source paths
+	AllowEmptyWildcard bool            // don't error when glob patterns match nothing
 }
 
 // Stat globs the specified pattern in the specified directory and returns its
@@ -399,6 +401,8 @@ type GetOptions struct {
 	IgnoreUnreadable   bool              // ignore errors reading items, instead of returning an error
 	NoCrossDevice      bool              // if a subdirectory is a mountpoint with a different device number, include it but skip its contents
 	Timestamp          *time.Time        // timestamp to force on all contents
+	DisallowWildcard   bool              // reject glob patterns in source paths
+	AllowEmptyWildcard bool              // don't error when glob patterns match nothing
 }
 
 // Get produces an archive containing items that match the specified glob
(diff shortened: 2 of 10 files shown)


tests/add.bats
```diff
@@ -509,3 +509,56 @@ EOF
...
+@test "add-allow-wildcard" {
+  createrandom ${TEST_SCRATCH_DIR}/file-a
+  createrandom ${TEST_SCRATCH_DIR}/file-b
+
+  run_buildah from $WITH_POLICY_JSON scratch
+  cid=$output
+  run_buildah config --workingdir / $cid
+
+  # Default (allow-wildcard=true): glob expands, adds both files
+  run_buildah add $cid ${TEST_SCRATCH_DIR}/file-* /dest/
```
nalind: The pattern isn't quoted here like it is for the "--allow-wildcard=false" invocation below, so the shell will be attempting to do the expansion.

tests/add.bats
```diff
@@ -509,3 +509,56 @@ EOF
...
+
+@test "add-allow-empty-wildcard" {
+  createrandom ${TEST_SCRATCH_DIR}/file-a
+
+  run_buildah from $WITH_POLICY_JSON scratch
+  cid=$output
+  run_buildah config --workingdir / $cid
+
+  # Default (allow-empty-wildcard=false): non-matching glob should fail
+  run_buildah 125 add $cid ${TEST_SCRATCH_DIR}/nonexistent-* /dest/
```
nalind: The test succeeds either way, but these two uses of "${TEST_SCRATCH_DIR}/nonexistent-*" should probably be quoted so that the shell doesn't even try to do the globbing itself.

tests/copy.bats
```diff
@@ -756,3 +756,56 @@ parents/y/b.txt"
...
+@test "copy-allow-wildcard" {
+  createrandom ${TEST_SCRATCH_DIR}/file-a
+  createrandom ${TEST_SCRATCH_DIR}/file-b
+
+  run_buildah from $WITH_POLICY_JSON scratch
+  cid=$output
+  run_buildah config --workingdir / $cid
+
+  # Default (allow-wildcard=true): glob expands, copies both files
+  run_buildah copy $cid ${TEST_SCRATCH_DIR}/file-* /dest/
```
nalind: The pattern isn't quoted here like it is for the "--allow-wildcard=false" invocation below, so the shell will be attempting to do the expansion.

tests/copy.bats
```diff
@@ -756,3 +756,56 @@ parents/y/b.txt"
...
+
+@test "copy-allow-empty-wildcard" {
+  createrandom ${TEST_SCRATCH_DIR}/file-a
+
+  run_buildah from $WITH_POLICY_JSON scratch
+  cid=$output
+  run_buildah config --workingdir / $cid
+
+  # Default (allow-empty-wildcard=false): non-matching glob should fail
+  run_buildah 125 copy $cid ${TEST_SCRATCH_DIR}/nonexistent-* /dest/
```
nalind: The test succeeds either way, but these two uses of "${TEST_SCRATCH_DIR}/nonexistent-*" should probably be quoted so that the shell doesn't even try to do the globbing itself.

copier/copier.go
```diff
@@ -344,9 +344,11 @@ func Eval(root string, directory string, _ EvalOptions) (string, error) {
 
 // StatOptions controls parts of Stat()'s behavior.
 type StatOptions struct {
-	UIDMap, GIDMap   []idtools.IDMap // map from hostIDs to containerIDs when returning results
-	CheckForArchives bool            // check for and populate the IsArchive bit in returned values
-	Excludes         []string        // contents to pretend don't exist, using the OS-specific path separator
+	UIDMap, GIDMap     []idtools.IDMap // map from hostIDs to containerIDs when returning results
+	CheckForArchives   bool            // check for and populate the IsArchive bit in returned values
+	Excludes           []string        // contents to pretend don't exist, using the OS-specific path separator
+	AllowWildcard      bool            // allow glob patterns in source paths; when false, they are rejected
```
nalind: This probably needs to be a "DisallowWildcard" to avoid breaking consumers like podman which won't know that they need to start setting this flag to get the behavior that they would have gotten before this field was added.

copier/copier.go
```diff
@@ -399,6 +401,8 @@ type GetOptions struct {
 	IgnoreUnreadable   bool              // ignore errors reading items, instead of returning an error
 	NoCrossDevice      bool              // if a subdirectory is a mountpoint with a different device number, include it but skip its contents
 	Timestamp          *time.Time        // timestamp to force on all contents
+	AllowWildcard      bool              // allow glob patterns in source paths; when false, they are rejected
```
nalind: This probably needs to be a "DisallowWildcard" to avoid breaking consumers like podman which won't know that they need to start setting this flag to get the behavior that they would have gotten before this field was added.

copier/copier_test.go
```diff
@@ -2365,6 +2368,148 @@ func testEnsure(t *testing.T) {
 	}
 }
 
+func TestStatAllowWildcard(t *testing.T) {
+	couldChroot := canChroot
+	canChroot = false
+	defer func() { canChroot = couldChroot }()
```
nalind: For many of the unit tests, these three lines would be how we start the version of the test that doesn't try to use chroot -- there would be a corresponding version of the test in "copier_unix_test.go" that didn't have this prologue, which would run the version of the test that starts a subprocess to do the heavy lifting, limited to being run on platforms where chroot() is expected to be available.  The rest of the test body would then be factored out into a function that they both called.  If we want to run this test with a chrooting subprocess, then it'll need a similar tweak.  If we don't, then it's fine.

TestCannotChangeMultipleRequestsWithDifferentChroot() checks the subprocess path without expecting to chroot(), so it's fine, but I'm just noticing that I missed this detail when reviewing TestTarPut().

nalind (commented):
Very good overall.  A question about the default for the allow-wildcard setting in the options structs in the copier package, and a couple of smaller questions around the tests.

copier/copier_test.go
nalind: That should read "with the can-chroot variable forced to false" for copier_test.go and "... forced to true" for copier_unix_test.go.  Sorry about that.

tests/add.bats
akca: Ah nice one, updated these now

copier/copier.go
akca: Good point, changed to DisallowWildcard. I had considered this but went with the positive name for clarity. You're right though, zero value safety for existing consumers is not so easy to guarantee.

copier/copier_test.go
akca: Makes sense, jsut factored the test logic into shared functions and added chroot variants in copier_unix_test following the existing pattern.

copier/copier.go
nalind: Thanks!

nalind (approved):
LGTM, thanks!

nalind:

cmd/buildah/addcopy.go
```diff
@@ -16,32 +16,35 @@ import (
...
+	tlsVerify          bool
+	certDir            string
+	retry              int
+	retryDelay         string
+	excludes           []string
+	parents            bool
+	timestamp          string
+	link               bool
+	allowWildcard      bool
+	allowEmptyWildcard bool
```
TomSweeneyRedHat, line 47: Not a needed change for now, but to make my OCD happy in the future, please alphabetize fields in a struct.  Yeah, it was a mess before you go into this one.

TomSweeneyRedHat:
LGTM
Thanks @akca !

PR #6845, opened 2026-05-11
Fix stale cache when using bind mount with build stage

### LINKED ISSUE #6609: Cache for RUN is not invalidated when using --mount=type=bind,from=stage
### Issue Description

When using a bind mount on a RUN command with from set to a previous build stage, the layer cache isn't invalidated resulting in out of date files being used.

Although it would seem that this should have been fixed by #5691, if the from parameter is set to a stage rather than an image reference the checksum does not get appended. The impression I get is that https://github.com/containers/buildah/blob/main/imagebuildah/stage_executor.go#L1978 needs an else block.

[mount-invalidate.tar.gz](https://github.com/user-attachments/files/24275346/mount-invalidate.tar.gz)

Attached example contains a Containerfile which pulls a file into a context stage, and then a second stage which copies the file into the final image.  The example.sh script runs this build twice and if the layer cache applies the image will be unchanged despite it changing the contents of the testfile between runs.

I've included a Containerfile.alt and example-alt.sh which builds the context image separately so that the final stage can reference the tagged image. This demonstrates that the append checksum logic works fine is not bugged in that case.

**Note** For some reason the layer cache doesn't kick in the first time the script is run (including after a podman system prune), but subsequent runs of the script exhibit the issue so be sure to run example.sh two or more times.

The cache is erroneously used even though the context stage has changed.
```
[1/2] STEP 1/2: FROM scratch AS context
[1/2] STEP 2/2: COPY testfile /
--> Using cache 95215d30ba82cb7bf55ef5ec41c795313284527a8cd81804d624747f6bbbc2b0
--> 95215d30ba82
[2/2] STEP 1/2: FROM alpine:3.23.2
[2/2] STEP 2/2: RUN --mount=type=bind,src=/,destination=/mnt,from=context cp /mnt/testfile /
--> Using cache b324b23cb0945840e9d374c758722647afb9829892b359e25eaa9f63162c1aa3
[2/2] COMMIT bug-example
--> b324b23cb094
Successfully tagged localhost/bug-example:latest
b324b23cb0945840e9d374c758722647afb9829892b359e25eaa9f63162c1aa3
-------- testfile modified --------
[1/2] STEP 1/2: FROM scratch AS context
[1/2] STEP 2/2: COPY testfile /
--> Using cache 125e0abbf26a2a799a335dac0266559593a5681061e8528cb8cc1c10dbddea4a
--> 125e0abbf26a
[2/2] STEP 1/2: FROM alpine:3.23.2
[2/2] STEP 2/2: RUN --mount=type=bind,src=/,destination=/mnt,from=context cp /mnt/testfile /
--> Using cache b324b23cb0945840e9d374c758722647afb9829892b359e25eaa9f63162c1aa3
[2/2] COMMIT bug-example
--> b324b23cb094
Successfully tagged localhost/bug-example:latest
b324b23cb0945840e9d374c758722647afb9829892b359e25eaa9f63162c1aa3
Bug: Image was unchanged
```

If the image is inspected we see that the additional checksum value is missing:
```
    "History": [
        {
            "created": "2025-12-18T00:12:29.242464453Z",
            "created_by": "ADD alpine-minirootfs-3.23.2-x86_64.tar.gz / # buildkit",
            "comment": "buildkit.dockerfile.v0"
        },
        {
            "created": "2025-12-18T00:12:29.242464453Z",
            "created_by": "CMD [\"/bin/sh\"]",
            "comment": "buildkit.dockerfile.v0",
            "empty_layer": true
        },
        {
            "created": "2025-12-20T23:40:20.723270656Z",
            "created_by": "/bin/sh -c --mount=type=bind,src=/,destination=/mnt,from=context cp /mnt/testfile /",
            "comment": "FROM docker.io/library/alpine:3.23.2"
        }
    ],
```

I would expect that a checksum is appended to the created_by field similar to what happens when an image reference is used:
```
    "History": [
        {
            "created": "2025-12-18T00:12:29.242464453Z",
            "created_by": "ADD alpine-minirootfs-3.23.2-x86_64.tar.gz / # buildkit",
            "comment": "buildkit.dockerfile.v0"
        },
        {
            "created": "2025-12-18T00:12:29.242464453Z",
            "created_by": "CMD [\"/bin/sh\"]",
            "comment": "buildkit.dockerfile.v0",
            "empty_layer": true
        },
        {
            "created": "2025-12-21T00:17:23.112063461Z",
            "created_by": "/bin/sh -c --mount=type=bind,src=/,destination=/mnt,from=localhost/bug-context cp /mnt/testfile /:sha256:a12b60e4b4a88348ab42baed5433a256c8fc83a08bb9a26ad9a9ad97990d7d9b",
            "comment": "FROM docker.io/library/alpine:3.23.2"
        }
    ],
```

### buildah version output

```yaml
buildah version 1.43.0-dev (image-spec 1.1.1, runtime-spec 1.3.0)
```

### buildah info output

```yaml
{
    "host": {
        "Distribution": {
            "distribution": "ubuntu",
            "version": "25.10"
        },
        "MemFree": 2670583808,
        "MemTotal": 67050049536,
        "OCIRuntime": "crun",
        "SwapFree": 0,
        "SwapTotal": 0,
        "arch": "amd64",
        "cpus": 16,
        "hostname": "AMD-9800X3D",
        "kernel": "6.17.0-7-generic",
        "os": "linux",
        "rootless": false,
        "uptime": "432h 29m 31.53s (Approximately 18.00 days)",
        "variant": ""
    },
    "store": {
        "ContainerStore": {
            "number": 0
        },
        "GraphDriverName": "overlay",
        "GraphImageStore": "",
        "GraphOptions": [
            "overlay.mountopt=nodev"
        ],
        "GraphRoot": "/var/lib/containers/storage",
        "GraphStatus": {
            "Supports d_type": "true",
            "Supports shifting": "true",
            "Supports volatile": "true",
            "Using metacopy": "false"
        },
        "GraphTransientStore": false,
        "ImageStore": {
            "number": 11
        },
        "RunRoot": "/run/containers/storage"
    }
}
```

### Provide your storage.conf

```toml
# files.
#
[storage]

driver = "overlay"

# Temporary storage location

# following commands:

# imagestore = ""


#

# transient_store = true

[storage.options]

additionalimagestores = [
]

# use_hard_links = "false"

# ostree_repos=""

# convert_images = "false"

#
#
#

#
# auto-userns-min-size=1024
#
# auto-userns-max-size=65536

[storage.options.overlay]
# and vfs drivers.

# inodes = ""

# directly.

mountopt = "nodev"

# skip_mount_home = "false"

# use_composefs = "false"

# size = ""

# directories.
#
#
#  "": No value specified.
#     image.
#     any user.
#
#
#  "force_mask"  permissions.
#
# force_mask = ""
```

### Upstream Latest Release

Yes

_No response_

### Additional information

_No response_

ekedaigle:
#### What type of PR is this?

/kind bug

This fixes a bug where stale images would be used when using a previous build stage as a bind mount source

#### How to verify it
Trivial example to reproduce:

```dockerfile
FROM ubi8/ubi as base

FROM base as builder
RUN mkdir /build && echo "1" > /build/foo.txt

FROM base as prod
RUN --mount=type=bind,source=/build,from=builder,target=/build \
     cp /build/foo.txt bar.txt
```

1. Build this file with `buildah build -f <dockerfile> --layers`.
2. Change the "1" to a "2" and build again, observing the final image hash changes
3. Change the "2" back to a "1" and observe the hash not change as it should

Fixes #6609 

I'm not particularly familiar with this code base, this was just an annoyance I wanted to fix, so comments welcome.

```release-note
Fixes bug where stale images could be used when using a previous build stage as a bind mount source
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -2053,8 +2053,13 @@ func (s *stageExecutor) getCreatedBy(node *parser.Node, addedContentSummary stri
 			}
 			// Source specified is part of stage, image or additional-build-context.
 			if mountOptionFrom != "" {
-				// If this is not a stage then get digest of image or additional build context
-				if _, ok := s.executor.stages[mountOptionFrom]; !ok {
+				if stage, ok := s.executor.stages[mountOptionFrom]; ok {
+					// If source is a previous stage then checksum is image digest for that stage
+					if image, isPreviousStage := s.executor.imageMap[stage.name]; isPreviousStage {
+						mountCheckSum = image
+					}
+				} else {
+					// If this is not a stage then get digest of image or additional build context
 					if builder, ok := s.executor.containerMap[mountOptionFrom]; ok {
 						// Found valid image, get image digest.
 						mountCheckSum = builder.FromImageDigest


TomSweeneyRedHat:
@ekedaigle thanks for the PR!
The imagebuildah/stage_executor.go has format issues.  Please `gofmt -s -w imagebuildah/stage_executor.go" and add it back in.

@nalind PTAL

nalind (commented):
Well spotted, and it's pretty close to what I think we'd want to be doing for that case.  The wrinkle is that the value being used in the current PR is the previous stage's image's ID (the hex part of the digest of its configuration blob), when for other source locations we use the digest of the image's manifest (including the algorithm part) or a digest over a directory tree.  The image ID's still unique enough that it should cause affect cache evaluation the way we desire, though.

The way I see that being sorted would be:
* add a new map field (`imageDigestMap`?) alongside the current `imageMap` field in the `executor` structure
* also set a value in the new map to `r.CommitResults.Digest.String()` when the corresponding value in the `imageMap` is set to `r.ImageID` in the current version
* use the value from the new map where the `imageMap` value is currently being consulted in the new block added by the PR

Would you be up for making those changes, or would you prefer for one of us prepare a follow-up PR?  Either way, thanks for catching this!

ekedaigle:
> Would you be up for making those changes, or would you prefer for one of us prepare a follow-up PR? Either way, thanks for catching this!

If one of you wants to take it don't let me stop you, I'd have to get more familiar with the internals here before making the changes myself.

nalind (approved):
Alright, then.
LGTM

nalind:
Can you squash these together, using something like `git rebase -i`?

ekedaigle:
Sure thing, should be set now.

nalind:
Thanks!

PR #6800, opened 2026-04-21
Ignore .containerignore for git repositories in ADD

simonbrauner:
#### What type of PR is this?

> /kind api-change

/kind bug

> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### Which issue(s) this PR fixes:

Fixes: https://github.com/containers/buildah/issues/6614

#### Does this PR introduce a user-facing change?

```release-note
Local `.containerignore` is no longer applied to ADD from git repositories.
```

PR #6849, opened 2026-05-13
rpm/buildah.spec tests: require xz and /usr/bin/selinuxenabled

nalind:
#### What type of PR is this?

/kind failing-test 

Tests use selinuxenabled to check if SELinux is enabled and relabeling is required when using volumes.
Some tarballs downloaded by tests are compressed using xz, as some are created using bzip2, which is called by tar, so make sure it's there.

#### How to verify it

Testing farm Raw Hide tests should pass.

Taking a guess at the cause of the errors seen at https://github.com/containers/buildah/pull/6800#issuecomment-4442001624.

```release-note
None
```

### DIFF
--- a/rpm/buildah.spec
+++ b/rpm/buildah.spec
@@ -97,6 +97,9 @@ Requires: bats
 Recommends: bats
 %endif
 Requires: bzip2
+Requires: xz
+Requires: zstd
+Requires: which
 Requires: podman
 Requires: golang
 Requires: jq
@@ -104,6 +107,7 @@ Requires: httpd-tools
 Requires: openssl
 Requires: nmap-ncat
 Requires: git-daemon
+Requires: libselinux-utils
 
 %description tests
 %{summary}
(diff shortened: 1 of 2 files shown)


Luap99 (approved):
LGTM
cc @lsm5

rpm/buildah.spec
```diff
@@ -97,13 +97,17 @@ Requires: bats
 Recommends: bats
 %endif
 Requires: bzip2
+Requires: xz
+Requires: zstd
```
lsm5, line 101: Do the tests require the zstd tool or is it handled internally by zstd Go library?

lsm5 (approved):
LGTM

rpm/buildah.spec
nalind: Not yet.  The add-local-archive test only uses `tar` to generate archives with gzip and bzip2, but the current logic also accepts xz and zstd.

PR #6850, opened 2026-05-13
tmt: archive audit and journal logs after test execution

lsm5:
#### What type of PR is this?

/kind other

captures audit logs

#### How to verify it
audit logs visible somewhere in testing-farm job run.

None

None

```release-note
None
```

### DIFF
--- a/plans/main.fmf
+++ b/plans/main.fmf
@@ -31,4 +31,23 @@ provision:
         disk:
             - size: ">= 512 GB"
 
+finish:
+    how: shell
+    script: |
+        # Archive logs to TMT_PLAN_DATA for post-test analysis
+        mkdir -p "$TMT_PLAN_DATA/logs"
+
+        # Copy audit logs (SELinux denials, audit events)
+        if [ -f /var/log/audit/audit.log ]; then
+            cp /var/log/audit/audit.log "$TMT_PLAN_DATA/logs/"
+            echo "Audit logs copied to $TMT_PLAN_DATA/logs/audit.log"
+        else
+            echo "No audit log found at /var/log/audit/audit.log"
+        fi
+
+        # Capture full journal (current boot only)
+        journalctl -b --no-pager --all --output=short-precise &> "$TMT_PLAN_DATA/logs/journal.log"
+        echo "Journal output saved to $TMT_PLAN_DATA/logs/journal.log"
+
+        echo "All logs archived to $TMT_PLAN_DATA/logs/"
 


lsm5:
@nalind PTAL. In the failed testing-farm job, see `data -> audit-logs -> audit.log` . At least on my browser, it wants to download the log by default, which is annoying but at least it's there now. 

cc @containers/buildah-maintainers 

Related: https://github.com/containers/buildah/pull/6800#issuecomment-4441602517

Luap99:
audit logs are part of the journald output normally I think? On cirrus we just upload the full journal output because that may contain much more useful information there, not just audit.

Luap99:
https://github.com/containers/buildah/blob/520f469a287ce12053b67f9098b55ea964c0e0f6/.cirrus.yml#L184
https://github.com/containers/buildah/blob/520f469a287ce12053b67f9098b55ea964c0e0f6/contrib/cirrus/logcollector.sh#L18

lsm5:
> audit logs are part of the journald output normally I think? On cirrus we just upload the full journal output because that may contain much more useful information there, not just audit.

sgtm, I can update PR to include that too.

plans/main.fmf
```diff
@@ -31,4 +31,23 @@ provision:
...
+        # Copy audit logs (SELinux denials, audit events)
+        if [ -f /var/log/audit/audit.log ]; then
+            cp /var/log/audit/audit.log "$TMT_PLAN_DATA/logs/"
+            echo "Audit logs copied to $TMT_PLAN_DATA/logs/audit.log"
+        else
+            echo "No audit log found at /var/log/audit/audit.log"
+        fi
+
+        # Capture full journal
+        journalctl --no-pager --all --output=short-precise > "$TMT_PLAN_DATA/logs/journal.log" 2>&1
```
Luap99: you likely want the `-b` here so we only see the logs of the current boot, in case there are older journal files around already on the system, or does tmt reboot the VM before collecting the logs?

Also just use `&> "$TMT_PLAN_DATA/logs/journal.log"` to redirect both stdout/err which is simpler to use than defining to redirects after each other.

lsm5: > you likely want the `-b` here so we only see the logs of the current boot, in case there are older journal files around already on the system, or does tmt reboot the VM before collecting the logs?

AFAIK, it doesn't reboot unless you issue a tmt-reboot, so we should be fine with -b.

> 
> Also just use `&> "$TMT_PLAN_DATA/logs/journal.log"` to redirect both stdout/err which is simpler to use than defining to redirects after each other.

Yes, thanks. Updating..

lsm5:
@containers/buildah-maintainers good for another look. Both audit and journal logs captured in `data -> logs`.

Luap99 (approved):
LGTM

nalind:
LGTM

PR #6851, opened 2026-05-13
conformance tests: more parallelization

nalind:
#### What type of PR is this?

/kind failing-test 

Parallelize more subtests to try to finish conformance testing faster.
Use t.Cleanup() instead of defer for a few cleanups.

#### How to verify it

Conformance tests should complete in less time unless they're I/O bound.

The "conformance tests using VFS" job seems to be limited by I/O, as it's not faster when given more memory and more CPU, and actually appears to be slower when we add the parallelization here.  The overlay counterpart, however, finishes in less time than it did before.

```release-note
None
```

### DIFF
--- a/tests/conformance/conformance_test.go
+++ b/tests/conformance/conformance_test.go
@@ -152,7 +152,6 @@ func TestMain(m *testing.M) {
 }
 
 func TestConformance(t *testing.T) {
-	t.Parallel()
 	dateStamp := fmt.Sprintf("%d", time.Now().UnixNano())
 	for i := range internalTestCases {
 		t.Run(internalTestCases[i].name, func(t *testing.T) {
@@ -196,6 +195,11 @@ func TestConformance(t *testing.T) {
 
 func testConformanceInternal(t *testing.T, dateStamp string, testIndex int, mutate func(*testCase)) {
 	test := internalTestCases[testIndex]
+
+	if !test.dontParallelize {
+		t.Parallel()
+	}
+
 	if mutate != nil {
 		mutate(&test)
 	}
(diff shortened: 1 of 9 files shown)


nalind:
Looks like that conformance-using-vfs job really is I/O bound.

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
@nalind looks like you have some conflicts that need to be resolved on this one.

tests/conformance/conformance_test.go
```diff
@@ -232,28 +236,25 @@ func testConformanceInternal(t *testing.T, dateStamp string, testIndex int, muta
...
-		pipeReader.Close()
-		wg.Done()
-	}()
-	wg.Wait()
+		return errors.Join(putErr, pipeReader.Close())
+	})
+	err = eg.Wait()
+	assert.NoErrorf(t, err, "error copying build info from %q", filepath.Join("testdata", test.dockerfile))
 	assert.NoErrorf(t, getErr, "error reading build info from %q", filepath.Join("testdata", test.dockerfile))
 	assert.NoErrorf(t, putErr, "error writing build info to %q", contextDir)
```
Honny1: Would it be better to assert errors immediately in the go routine?

Honny1 (commented):
LGTM, I have just a possible cleanup comment.

tests/conformance/conformance_test.go
nalind: Sure.  They're safe to be called outside of the goroutine that's running the test (their require counterparts, not so much).

Honny1 (approved):
LGTM

nalind:
> Looks like that conformance-using-vfs job really is I/O bound.

And they both are in the new setup, the conformance test jobs don't complete within the timeout when they currently finish on main.

PR #6855, opened 2026-05-18
imagebuildah.executor.getCreatedBy(): use digests for previous stages

nalind:
#### What type of PR is this?

/kind cleanup

Instead of the previous stage's image ID, use the digest of that stage's manifest, as we would for other images that affect the effects of this instruction.  Follows up on #6845.  I'll mention the phrase "burst cache" here so that people searching for it in the commit log will find this and the reference to the previous patch when looking for related fixes.

#### How to verify it

New integration test based on the reproducer in #6845.

None

```release-note
None
```

### DIFF
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -121,7 +121,8 @@ type executor struct {
 	useCache                                bool
 	removeIntermediateCtrs                  bool
 	forceRmIntermediateCtrs                 bool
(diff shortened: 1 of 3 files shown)


TomSweeneyRedHat:
LGTM

nalind:

TomSweeneyRedHat:
I'm gonna smash this one in.

PR #6858, opened 2026-05-20
Correctly report archiveSource if we can't find it

mtrmac:
#### What type of PR is this?

> /kind bug

Don't write only to a shadowing variable live only inside the switch, there is a formerly-shadowed variable used afterwards.

#### How to verify it

Break `newSingleItemArchive` so that it produces nothing, and observe the `expected to have archived a copy of` message.

None

```release-note
None
```

### DIFF
--- a/internal/sanitize/sanitize.go
+++ b/internal/sanitize/sanitize.go
@@ -211,7 +211,8 @@ func ImageName(transportName, restOfImageName, contextDir, tmpdir string) (newFr
 	case dockerArchiveTransport.Transport.Name(), ociArchiveTransport.Transport.Name(): // these are, basically, tarballs
 		// these take the form path[:stuff]
 		transportRef := restOfImageName
-		archiveSource, refLeftover, ok := strings.Cut(transportRef, ":")
+		as, refLeftover, ok := strings.Cut(transportRef, ":")
+		archiveSource = as
 		if ok {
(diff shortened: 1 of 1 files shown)


mtrmac:
This should ideally never be reachable, so I think “No New Tests” is appropriate, but I’ll let repo maintainers make that decision (and perhaps recommend a testing approach).

internal/sanitize/sanitize.go
```diff
@@ -211,7 +211,8 @@ func ImageName(transportName, restOfImageName, contextDir, tmpdir string) (newFr
 	case dockerArchiveTransport.Transport.Name(), ociArchiveTransport.Transport.Name(): // these are, basically, tarballs
 		// these take the form path[:stuff]
 		transportRef := restOfImageName
-		archiveSource, refLeftover, ok := strings.Cut(transportRef, ":")
+		as, refLeftover, ok := strings.Cut(transportRef, ":")
+		archiveSource = as
```
nalind, line 215: There's a similar construction around line 250.  Does that require a similar change?

mtrmac: Oops, it does. Fixed, thanks!

nalind (approved):
Thanks!
LGTM

Luap99 (approved):
LGTM

PR #6862, opened 2026-05-21
Don't report an error in a possible RPC server start/stop ordering

mtrmac:
#### What type of PR is this?

/kind failing-test 

```
not ok 897 rpc noop
# (from function `assert' in file ./helpers.bash, line 553,
#  in test file ./rpc.bats, line 7)
#   `assert "$output" = $(pwd)' failed
# …/buildah/tests …/buildah/tests
# # …/buildah/tests/./../bin/buildah rpc -l /var/tmp/buildah_tests.ufg3rg/socket pwd
# …/buildah/tests
# time="2026-05-21T06:40:46+02:00" level=error msg="while waiting for rpc service to shut down: grpc: the server has been stopped"
# #/vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
# #|     FAIL: buildah rpc -l /var/tmp/buildah_tests.ufg3rg/socket pwd
# #| expected: = '…/buildah/tests'
# #|   actual:   '…/buildah/tests'
# #|         >   'time="2026-05-21T06:40:46+02:00" level=error msg="while waiting for rpc service to shut down: grpc: the server has been stopped"'
# #\^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

#### How to verify it

Reproducible on a pretty slow VM.

None

```release-note
None
```

### DIFF
--- a/cmd/buildah/rpc.go
+++ b/cmd/buildah/rpc.go
@@ -71,7 +71,9 @@ func rpcCmd(c *cobra.Command, args []string) error {
 	})
 	defer func() {
 		s.GracefulStop() // closes the listening socket
-		if err := errgroup.Wait(); err != nil {
+		// We might have called GracefulStop() before Serve() had a chance to start;
+		// in that case it returns ErrServerStopped.
+		if err := errgroup.Wait(); err != nil && err != grpc.ErrServerStopped {
 			logrus.Errorf("while waiting for rpc service to shut down: %v", err)
 		}
 	}()


mtrmac (commented):
A LLM tells me that reliably waiting for `Serve` to start would be cleaner, but `Serve` does not provide any API for that; we’d have to wrap the `Listener` and signal on the first `Accept`, or something like that.

nalind (approved):
> A LLM tells me that reliably waiting for `Serve` to start would be cleaner, but `Serve` does not provide any API for that; we’d have to wrap the `Listener` and signal on the first `Accept`, or something like that.

Yeah, that's probably more trouble than it's worth when that first test invocation is running something that isn't even going to connect to the server.

LGTM

mtrmac:
BTW I don’t think the _client_ is affected: We do call `Listen()` on the socket before we start the client, so the client’s operations should never get _lost_. This only shows up in this singular case, where client is a no-op and does not actually wait for any reply from the server, so the server might not get the opportunity to do anything.

TomSweeneyRedHat:
LGTM
Thanks @mtrmac 
I've restarted the one failing test, it appeared to be a flake

PR #6864, opened 2026-05-21
bump oc/selinux to v1.15.0, remove !linux wrappers

kolyshkin:
/kind cleanup

1. deps: bump oc/selinux to v1.15.0

2. selinux: simplify code
    
    Remove the !linux version since oc/selinux now has a !linux wrappers,
    and selinux.GetEnabled always returns false on !linux.

None

```release-note
NONE
```


TomSweeneyRedHat:
@kolyshkin smoke test is smoking

kolyshkin:
> @kolyshkin smoke test is smoking

@TomSweeneyRedHat I only see "pr should include tests", can you please add `No new tests`?

nalind:
The cross-compile check should catch any problems created by the removal of the non-Linux bits, if there are any.

nalind (approved):
LGTM

nalind:

selinux.go
```diff
@@ -23,7 +17,7 @@ func setupSelinux(g *generate.Generator, processLabel, mountLabel string) {
 }
 
 func runLabelStdioPipes(stdioPipe [][]int, processLabel, mountLabel string) error {
```
mtrmac, line 20: This code with `/proc/self/fd` remains Linux-specific (although unreachable on other platforms due to `selinux.GetEnabled()`). I’ll leave it to Buildah maintainers to decide whether they prefer strict platform separation or to avoid the extra stub.

mtrmac (commented):
ACK but I’ll leave the final merge decision to experts.

selinux.go
kolyshkin: If that's a concern, I've added another guard (`runtime.GOOS == "linux"`) which has zero runtime overhead because `runtime.GOOS` is a constant, so the compiled code should be identical to what we had before.

selinux.go
```diff
@@ -1,29 +1,24 @@
...
+	if runtime.GOOS == "linux" && processLabel != "" && selinux.GetEnabled() {
 		g.SetProcessSelinuxLabel(processLabel)
 		g.SetLinuxMountLabel(mountLabel)
 	}
 }
 
 func runLabelStdioPipes(stdioPipe [][]int, processLabel, mountLabel string) error {
-	if !selinuxGetEnabled() || processLabel == "" || mountLabel == "" {
+	if runtime.GOOS != "linux" || !selinux.GetEnabled() || processLabel == "" || mountLabel == "" {
 		// SELinux is completely disabled, or we're not doing anything at all with labeling
```
lsm5: Nit
```suggestion
		// Either not on Linux or SELinux is completely disabled, or we're not doing anything at all with labeling
```

lsm5 (approved):
LGTM

selinux.go
kolyshkin: thanks, fixed

lsm5 (approved):
LGTM

giuseppe (approved):
LGTM

PR #6873, opened 2026-05-26
build: commit an empty layer instead of a layer with nothing in it

### LINKED ISSUE #6860: RootFS.layers entry added for operations that do not impact the filesystem
### Issue Description

I've noticed that `buildah build` and `podman build --squash` both produce a RootFS layer of `sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef` even if there are no RootFS-impacting instructions in the Containerfile. Observing an image with this layer, it contains no content but still ends in the RootFS.layers blob.

Steps to reproduce the issue
1. Build this image using `buildah build <file>`

```
from registry.fedoraproject.org/fedora@sha256:dd93b388ec7b1da2055d2234532499526cef6a799de80f95a923f293602dd462 as base

LABEL foo=bar
```
```
❯ podman image inspect <imageid> | jq '.[0].RootFS.Layers'
[
  "sha256:35fccc735c0f5e9fe7c672e3162d8ee5b515e21d0de850026752bb4c1806a51a",
  "sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef"
]
```

3. Optionally inspect the content associated with `sha256:5f70b...` and notice it's empty.

`"sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef"` shows up as the final layer in RootFS.Layers for the image.

As it does not contain filesystem-impacting content, I would not expect this to be in the RootFS.Layers list.

### buildah version output

```yaml
VERSION: buildah version 1.43.1 (image-spec 1.1.1, runtime-spec 1.2.1)
RPM: buildah-1.43.1-1.fc43.x86_64
```

### buildah info output

```yaml
{
    "host": {
        "CgroupVersion": "v2",
        "Distribution": {
            "distribution": "fedora",
            "version": "43"
        },
        "MemFree": 1716559872,
        "MemTotal": 67100418048,
        "OCIRuntime": "crun",
        "SwapFree": 8310710272,
        "SwapTotal": 8589930496,
        "arch": "amd64",
        "cpus": 16,
        "hostname": "my-hostname",
        "kernel": "7.0.8-100.fc43.x86_64",
        "os": "linux",
        "rootless": true,
        "uptime": "73h 52m 44.22s (Approximately 3.04 days)",
        "variant": ""
    },
    "store": {
        "ContainerStore": {
            "number": 107
        },
        "GraphDriverName": "overlay",
        "GraphImageStore": "",
        "GraphOptions": null,
        "GraphRoot": "/home/username/.local/share/containers/storage",
        "GraphStatus": {
            "Backing Filesystem": "btrfs",
            "Supports d_type": "true",
            "Supports shifting": "false",
            "Supports volatile": "true",
            "Using metacopy": "false"
        },
        "GraphTransientStore": false,
        "ImageStore": {
            "number": 891
        },
        "RunRoot": "/run/user/107112/containers"
    }
}
```

### Provide your storage.conf

```toml
n/a
```

### Upstream Latest Release

Yes

### Additional information

_No response_

nalind:
#### What type of PR is this?

/kind feature

When `buildah build`ing an image in single-layer mode, if the generated diff has nothing in it, omit the blob and set the EmptyLayer field in the final history entry, which would otherwise correspond to it.

#### How to verify it

New integration test!

Fixes #6860

```release-note
Images built by `buildah build` without `--layers=true` will no longer add a layer with no contents to the output image if there were no changes to the working container's root filesystem during the build.
```

### DIFF
--- a/image.go
+++ b/image.go
@@ -126,6 +126,7 @@ type containerImageRef struct {
 	confidentialWorkload  ConfidentialWorkloadOptions
 	omitHistory           bool
 	emptyLayer            bool
+	emptyLayerIfEmptyDiff bool
 	omitLayerHistoryEntry bool
 	idMappingOptions      *define.IDMappingOptions
 	parent                string
@@ -1184,6 +1185,18 @@ func (i *containerImageRef) NewImageSource(ctx context.Context, _ *types.SystemC
 		if err = os.Rename(filepath.Join(path, "layer"), finalBlobName); err != nil {
 			return nil, fmt.Errorf("storing %s to file while renaming %q to %q: %w", what, filepath.Join(path, "layer"), finalBlobName, err)
 		}
+		// If the layer blob is just a block of zeroes of a size that could
+		// plausibly be an empty diff (i.e., if it's several megabytes,
+		// don't bother, because it's highly unlikely), suppress it.
+		if i.emptyLayerIfEmptyDiff && layerID == i.layerID {
+			switch size {
+			case 0, 512, 1024, 2048:
+				if srcHasher.Digest() == digest.Canonical.FromBytes(make([]byte, size)) {
+					i.emptyLayer = true
+					continue
+				}
+			}
+		}
 		mb.addLayer(destHasher.Digest(), size, srcHasher.Digest())
 	}
 
@@ -1764,6 +1777,7 @@ func (b *Builder) makeContainerImageRef(options CommitOptions) (*containerImageR
 		confidentialWorkload:  options.ConfidentialWorkloadOptions,
 		omitHistory:           options.OmitHistory || forceOmitHistory,
 		emptyLayer:            (options.EmptyLayer || options.OmitLayerHistoryEntry) && !options.Squash && !options.ConfidentialWorkloadOptions.Convert,
+		emptyLayerIfEmptyDiff: options.EmptyLayerIfEmptyDiff && !options.Squash && !options.ConfidentialWorkloadOptions.Convert,
 		omitLayerHistoryEntry: options.OmitLayerHistoryEntry && !options.Squash && !options.ConfidentialWorkloadOptions.Convert,
 		idMappingOptions:      &b.IDMappingOptions,
 		parent:                parent,
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -884,6 +884,75 @@ _EOF
   expect_output --substring "Cache burst add diff"
 }
 
+@test "bud build without multiple layers, with and without layer-creating instructions" {
+  local base=busybox
+  _prefetch $base
+
+  contextdir=${TEST_SCRATCH_DIR}/context
+  mkdir -p ${contextdir}
+  createrandom ${contextdir}/randomfile
+
+  # count the number of layers in the base image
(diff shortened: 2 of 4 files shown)


TomSweeneyRedHat:
LGTM

image.go
```diff
@@ -1184,6 +1185,18 @@ func (i *containerImageRef) NewImageSource(ctx context.Context, _ *types.SystemC
 		if err = os.Rename(filepath.Join(path, "layer"), finalBlobName); err != nil {
 			return nil, fmt.Errorf("storing %s to file while renaming %q to %q: %w", what, filepath.Join(path, "layer"), finalBlobName, err)
 		}
+		// If the layer blob is just a block of zeroes of a size that could
+		// plausibly be an empty diff (i.e., if it's several megabytes,
+		// don't bother, because it's highly unlikely), suppress it.
+		if i.emptyLayerIfEmptyDiff && layerID == i.layerID {
+			switch size {
+			case 0, 512, 1024, 2048:
```
Honny1, line 1193: What this numbers mean?

tests/bud.bats
```diff
@@ -884,6 +884,75 @@ _EOF
...
+          run_buildah inspect --type=image --format "{{ index .Docker.RootFS.DiffIDs $baselayers }}" $(< ${TEST_SCRATCH_DIR}/iid.txt)
+          local lastdiffid="$output"
+          # sha256sum([0]{})
+          assert lastdiffid != sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
+          # sha256sum([1024]{})
+          assert lastdiffid != sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef
+          # sha512sum([0]{})
+          assert lastdiffid != sha512:cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e
+          # sha512sum([1024]{})
+          assert lastdiffid != sha512:8efb4f73c5655351c444eb109230c556d39e2c7624e9c11abc9e3fb4b9b9254218cc5085b454a9698d085cfa92198491f07a723be4574adc70617b73eb0b6461
```
Honny1, line 945: Is this correct? Should we use some flag or config to force buildah to use `sha512`?

image.go
nalind: Those are multiples of the record size used by tar(5) - an archive with no entries in it still includes the end-of-archive marker, which is usually 1024 bytes, i.e., two records full of zero bytes.  The rest of the possible lengths are just guesses, which should be harmless since still we only discard layers of that length if the diff is all zero bytes.

tests/bud.bats
nalind: We currently don't have a flag to do that.  This is an attempt at predicting the future.

Honny1 (approved):
LGTM

Luap99 (approved):
LGTM

Luap99:
@nalind this will need a rebase to pick up the new CI runs, otherwise I cannot merge

PR #6881, opened 2026-05-29
 Clean up TempDirForURL directories for --build-context and ADD git

Honny1:
URL-based `--build-context` downloads and ADD git clones left temporary directories under `$TMPDIR` after builds. This PR removes them in `executor.cleanup()` and after ADD git handling, and adds bats checks via `assert_no_tempdir_leaks`.

#### What type of PR is this?

> /kind api-change
> 
/kind bug
> 
/kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```

### DIFF
--- a/add.go
+++ b/add.go
@@ -602,12 +602,12 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 				go func() {
 					defer wg.Done()
 					defer pipeWriter.Close()
-					// TODO: the returned cloneDir is never cleaned up, leaking disk space.
 					var cloneDir, subdir string
 					cloneDir, subdir, getErr = define.TempDirForURL(tmpdir.GetTempDir(), "", src)
 					if getErr != nil {
 						return
 					}
+					defer os.RemoveAll(cloneDir)
 					getOptions := copier.GetOptions{
 						UIDMap:             srcUIDMap,
 						GIDMap:             srcGIDMap,
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -40,8 +40,16 @@ import (
 	"go.podman.io/image/v5/types"
 	"go.podman.io/storage"
 	"go.podman.io/storage/pkg/archive"
+	"golang.org/x/sync/errgroup"
 	"golang.org/x/sync/semaphore"
 )
+
+type additionalBuildContext struct {
+	define.AdditionalBuildContext
+	// downloadedTempDir is the temporary directory created for the additional build context.
+	// It should be removed when the build ends.
+	downloadedTempDir string
+}
 
 // builtinAllowedBuildArgs is list of built-in allowed build args.  Normally we
 // complain if we're given values for arguments which have no corresponding ARG
@@ -161,7 +169,7 @@ type executor struct {
 	imageInfoLock                           sync.Mutex
 	imageInfoCache                          map[string]imageTypeAndHistoryAndDiffIDs
 	fromOverride                            string
-	additionalBuildContexts                 map[string]*define.AdditionalBuildContext
+	additionalBuildContexts                 map[string]*additionalBuildContext
 	manifest                                string
 	secrets                                 map[string]define.Secret
 	sshsources                              map[string]*sshagent.Source
@@ -268,6 +276,11 @@ func newExecutor(logger *logrus.Logger, logPrefix string, store storage.Store, o
 		buildOutputs = append(buildOutputs, options.BuildOutput) //nolint:staticcheck
 	}
 
+	wrappedAdditionalBuildContexts := make(map[string]*additionalBuildContext, len(options.AdditionalBuildContexts))
+	for name, ctx := range options.AdditionalBuildContexts {
+		wrappedAdditionalBuildContexts[name] = &additionalBuildContext{AdditionalBuildContext: *ctx}
+	}
+
 	exec := executor{
 		args:                                    options.Args,
 		cacheFrom:                               options.CacheFrom,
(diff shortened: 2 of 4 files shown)


Honny1:
PTAL @mtrmac @nalind @TomSweeneyRedHat

Honny1:
PTAL @podman-container-tools/buildah-maintainers CI failure seems to be unrelated. + I cannot rerun CI.

lsm5:
debian jobs are unable to find docker.io deb

Honny1:
> debian jobs are unable to find docker.io deb

It seems to be a problem. cc @timcoding1988

define/build.go
```diff
@@ -26,6 +26,10 @@ type AdditionalBuildContext struct {
 	// from an external tar archive.  This will be populated only if the
 	// build context was a URL and its contents have been downloaded.
 	DownloadedCache string
+	// DownloadedTempDir is the os.MkdirTemp root returned by TempDirForURL
+	// when a URL-based additional build context is downloaded.
+	// Populated only when IsURL and the context has been downloaded.
```
nalind: It might be simpler to describe this as a temporary directory that should be (will be) removed when the build context is no longer needed, and that it may be different from `DownloadedCache`, because the implication that it's only deleted if the additional build context was a URL isn't enforced.

tests/bud.bats
```diff
@@ -10508,3 +10522,35 @@ _EOF
...
+  cp $tarball $httpdir/context.tar.gz
+
+  starthttpd $httpdir
+  local contextdir=${TEST_SCRATCH_DIR}/bud/context
+  mkdir -p $contextdir
+  cat > $contextdir/Dockerfile << _EOF
+FROM alpine
+COPY --from=ctx myproject/file.txt /file.txt
+RUN cat /file.txt
+RUN --mount=type=bind,src=myproject,from=ctx2,target=/mnt,z ls /mnt
```
nalind: Whether or not this produces any intended output doesn't seem to be checked later.

add.go
```diff
@@ -602,9 +602,9 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 				go func() {
 					defer wg.Done()
 					defer pipeWriter.Close()
-					// TODO: the returned cloneDir is never cleaned up, leaking disk space.
 					var cloneDir, subdir string
 					cloneDir, subdir, getErr = define.TempDirForURL(tmpdir.GetTempDir(), "", src)
+					defer os.RemoveAll(cloneDir)
```
nalind: If `getErr` is not `nil`, then `cloneDir` is empty, so this should be moved to after the error is checked.

nalind (changes requested):
Question about a test, and a possible call to `os.RemoveAll("")`.

timcoding1988:
> > debian jobs are unable to find docker.io deb
> 
> It seems to be a problem. cc @timcoding1988

https://github.com/podman-container-tools/buildah/pull/6889 this will fix it. you can rebase after my fix merged.

Honny1:
PTAL @podman-container-tools/buildah-maintainers @podman-container-tools/buildah-reviewers  CI failure seems to be unrelated.

giuseppe (approved):
LGTM

imagebuildah/executor.go
```diff
@@ -857,6 +857,15 @@ func (b *executor) Build(ctx context.Context, stages imagebuilder.Stages) (image
 		}
 		cleanupImages = nil
 
+		for _, additionalBuildContext := range b.additionalBuildContexts {
+			if additionalBuildContext.DownloadedTempDir != "" {
+				if err := os.RemoveAll(additionalBuildContext.DownloadedTempDir); err != nil {
```
mheon: Can we get a debug or info level log before this? I can see this taking quite a while to execute for large directories, so we should provide at least some sort of debug info for folks wondering why their Podman seems to freeze at the end of a build

define/build.go
```diff
@@ -26,6 +26,9 @@ type AdditionalBuildContext struct {
 	// from an external tar archive.  This will be populated only if the
 	// build context was a URL and its contents have been downloaded.
 	DownloadedCache string
+	// DownloadedTempDir is the TempDirForURL root to remove when the build ends.
+	// It may differ from DownloadedCache, which is the path inside that tree.
+	DownloadedTempDir string
```
nalind: This doesn't need to be a field that API users need to see or think about; the location we track there is always being cleaned up by the same package that's downloading the content.
Try adding a non-exported type to the imagebuildah package that embeds this struct with the added field, and have it use that new type where this patch is currently using this struct.

imagebuildah/executor.go
nalind: Move these RemoveAll calls into goroutines managed by a sync.WaitGroup or errgroup.Group?

nalind (changes requested):

tests/bud.bats
```diff
@@ -10510,3 +10523,35 @@ _EOF
...
+  cp $tarball $httpdir/context.tar.gz
+
+  starthttpd $httpdir
+  local contextdir=${TEST_SCRATCH_DIR}/bud/context
+  mkdir -p $contextdir
+  cat > $contextdir/Dockerfile << _EOF
+FROM alpine
+COPY --from=ctx myproject/file.txt /file.txt
+RUN cat /file.txt
+RUN --mount=type=bind,src=myproject,from=ctx2,target=/mnt,z test -f /mnt/file.txt
```
nalind: Non-blocker: `cmp` could compare the file's contents to the rootfs's `/file.txt`.

nalind (approved):
One non-blocking nit, otherwise LGTM.

Honny1:
I have resolved the non-blocking comment. I think we are good to go. @podman-container-tools/buildah-maintainers @podman-container-tools/buildah-reviewers

Honny1:
PTAL @nalind

PR #6882, opened 2026-05-29
Re-enable git ADD escape directory test

### LINKED ISSUE #6876: Re-enable “bud with ADD with git repository source escape directory”
It is failing in CI: https://github.com/containers/buildah/pull/6875#issuecomment-4556149665 and disabled.

it would probably need something like https://github.com/containers/podman/blob/20c6bd5bdd9e2d2c77a8a0e6f75cc0bffe458f51/test/buildah-bud/run-buildah-bud-tests#L165-L178 .

Cc: @Honny1

Honny1:
Use repo-local git identity so the test passes in CI without affecting the caller's global git config.

Fixes: https://github.com/containers/buildah/issues/6876

#### What type of PR is this?

> /kind api-change
> /kind bug
> 
/kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```


Honny1:
PTAL @mtrmac @nalind @TomSweeneyRedHat

nalind:
In other tests we build the repository ahead of time and archive it under `tests/git-daemon`, then extract it before serving it out.  Why not do that?

Honny1:
> In other tests we build the repository ahead of time and archive it under `tests/git-daemon`, then extract it before serving it out. Why not do that?

I just followed the same approach as the PoC and didn't think of doing it that way. Should I change it?

nalind:
It avoids certain variables that aren't the point of the test anyway, so yes, please.

Honny1:
PTAL @podman-container-tools/buildah-maintainers CI failure seems to be unrelated. + I cannot rerun CI.

lsm5:
retriggered CI

tests/bud.bats
nalind, line 8217: This should be looking more carefully at the error message.

nalind (changes requested):
I suspect the new test is hitting other errors when it's run as root  - the new tarball's contents are owned by a user who is not root:root, so `tar` will use the ownership information in the archive when root extracts it, and `git` will complain about the ownership being dubious.  The test could be hitting that error during the test build and not noticing the difference.
That's what I see when running the new test under `podman unshare` with `--show-output-of-passing-tests`.

Honny1:
PTAL @podman-container-tools/buildah-maintainers @podman-container-tools/buildah-reviewers  CI failure seems to be unrelated.

mheon:
LGTM

nalind (approved):
LGTM, thanks!

PR #6883, opened 2026-05-29
fix: replace dead len > 2 guard with len != 2 in addSysctl

immanuwell:
#### What type of PR is this?

> /kind bug

`strings.SplitN(s, "=", 2)` returns at most 2 elements, so `len(splitn) > 2` in `addSysctl` is dead code that never fires. When a sysctl in `containers.conf` has no `=`, `splitn` has length 1 and `splitn[1]` panics instead of hitting the error return right below.

Changing `> 2` to `!= 2` kills the dead code and catches the bad input.

#### How to verify it

```
# ~/.config/containers/containers.conf
[containers]
default_sysctls = ["net.core.rmem_max"]
```

Before: `buildah run` panics with `index out of range [1] with length 1`
After: returns `sysctl "net.core.rmem_max" defined in containers.conf must be formatted name=value`

None

```release-note
None
```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -833,7 +833,7 @@ func setupNamespaces(_ *logrus.Logger, g *generate.Generator, namespaceOptions d
 	addSysctl := func(prefixes []string) error {
 		for _, sysctl := range defaultContainerConfig.Sysctls() {
 			splitn := strings.SplitN(sysctl, "=", 2)
-			if len(splitn) > 2 {
+			if len(splitn) != 2 {
 				return fmt.Errorf("sysctl %q defined in containers.conf must be formatted name=value", sysctl)
 			}
 			for _, prefix := range prefixes {


nalind (approved):
The original version was probably intended to be `<`, but this works, too.
LGTM, thanks!

TomSweeneyRedHat:
LGTM

Honny1 (approved):
LGTM, @immanuwell can you please rebase on upstream main to enable new CI?

TomSweeneyRedHat:
Ping @immanuwell, can you rebase please per @honny1

Honny1:
@TomSweeneyRedHat Let's give @immanuwell a week to respond. After that, a maintainer (with permission to rebase contributors' PRs) should rebase this PR. Since @immanuwell hasn't responded in a month, it's okay to take over since it only needs a rebase.

PR #6884, opened 2026-05-29
new gha ci with lima

timcoding1988:
Replaces Cirrus with lima-based GitHub Actions CI. Ported from podman-container-tools/buildah-sandbox#2.


contrib/ci/ci.sh
Luap99, line 1: For consistency we wanted to use hack/ci everywhere, though I guess not blocking the merge, we can rename afterwards.

Luap99 (approved):
not a deep review but LGTM if it passes tests for now, keep in mind something is better than no tests

cc @nalind @TomSweeneyRedHat

nalind (approved):
* We exclude integration tests for storage=vfs+priv=rootless along with those for storage=overlay+priv=rootless, so I don't think that leaves us running any integration tests with priv=rootless.
* I don't see where the "in_podman" test runs the test inside of a podman container.

But some is better than none, so we can work on those.  LGTM unless we want to fix those first.

TomSweeneyRedHat:
LGTM in general.  I'd like to see @Luap99 's comment cleaned up, and then any further tweaks that you want to make in the short term @timcoding1988

.github/workflows/ci.yml
```diff
@@ -0,0 +1,205 @@
...
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  path-filter:
+    name: path-filter
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions:
+      pull-requests: read
```
Honny1, line 23: Is this needed? https://github.com/podman-container-tools/podman/pull/28821#discussion_r3333789662

Honny1 (commented):
LGTM

.github/workflows/ci.yml
nalind: Based on https://github.com/podman-container-tools/podman/pull/28821#discussion_r3333887053 it seems the answer is "no", but we can sort this out in a follow-up.

PR #6885, opened 2026-06-01
Disallow Dockerfile symlinks escaping build context

### LINKED ISSUE #6861: `buildah build` follows `Dockerfile` symlink outside the context directory
### Issue Description

`buildah build dir` follows `Dockerfile` symlink to a file outside the context directory.

This behavior does not match `docker build dir` (using docker 29.4.0 and buildkit 0.30.0).

Steps to reproduce the issue
1. Create file `~/test.bash` containing
   ```
   #!/bin/bash
   set -o errexit
   set -o nounset
   command=$1
   dir=$(mktemp -d)
   
   cat <<EOF > $dir/Dockerfile
   FROM docker.io/library/alpine
   EOF
   
   mkdir $dir/context
   (cd $dir/context && ln -s ../Dockerfile Dockerfile)
   $command build --no-cache $dir/context
   ```
2. Test buildah
   ```
   $ bash test.bash buildah
   STEP 1/1: FROM docker.io/library/alpine
   COMMIT
   --> 2ffb2ff4aab3
   2ffb2ff4aab36d06b7f3266bbb10e8232769cd2360613131d37abd19430cf6f1
   ```
3. Test docker (with buildkit)
   ```
   $ DOCKER_BUILDKIT=1 bash test.bash docker
   [+] Building 0.0s (1/1) FINISHED                                                                                                                                                                                                         docker:default
    => [internal] load build definition from Dockerfile                                                                                                                                                                                                  0.0s
    => => transferring dockerfile: 106B
   ERROR: failed to build: failed to solve: failed to read dockerfile: too many links
   ```


The buildah command succeeds.
The docker command fails.

The buildah command fails.
The docker command fails.

### buildah version output

```yaml
Version:         1.43.1
Go Version:      go1.26.2-X:nodwarf5
Image Spec:      1.1.1
Runtime Spec:    1.2.1
CNI Spec:        1.1.0
libcni Version:  
image Version:   5.39.2
Git Commit:      
Built:           Tue Apr 14 22:15:13 2026
OS/Arch:         linux/arm64
```

### buildah info output

```yaml
{
    "host": {
        "CgroupVersion": "v2",
        "Distribution": {
            "distribution": "fedora",
            "version": "45"
        },
        "MemFree": 178765824,
        "MemTotal": 2037288960,
        "OCIRuntime": "crun",
        "SwapFree": 0,
        "SwapTotal": 0,
        "arch": "arm64",
        "cpus": 2,
        "hostname": "localhost.localdomain",
        "kernel": "7.0.0-62.fc45.aarch64",
        "os": "linux",
        "rootless": true,
        "uptime": "36h 55m 35.89s (Approximately 1.50 days)",
        "variant": ""
    },
    "store": {
        "ContainerStore": {
            "number": 0
        },
        "GraphDriverName": "overlay",
        "GraphImageStore": "",
        "GraphOptions": null,
        "GraphRoot": "/var/home/test11/.local/share/containers/storage",
        "GraphStatus": {
            "Backing Filesystem": "xfs",
            "Supports d_type": "true",
            "Supports shifting": "false",
            "Supports volatile": "true",
            "Using metacopy": "false"
        },
        "GraphTransientStore": false,
        "ImageStore": {
            "number": 38
        },
        "RunRoot": "/run/user/1087/containers"
    }
}
```

### Provide your storage.conf

```toml
# files.
#
[storage]

driver = "overlay"

# Temporary storage location

# following commands:

# imagestore = ""


#

# transient_store = true

[storage.options]

additionalimagestores = [
]

# use_hard_links = "false"

# ostree_repos=""

# convert_images = "false"

#
#
#

#
# auto-userns-min-size=1024
#
# auto-userns-max-size=65536

[storage.options.overlay]
# and vfs drivers.

# inodes = ""

# directly.

# skip_mount_home = "false"

# use_composefs = "false"

# size = ""

# directories.
#
#
#  "": No value specified.
#     image.
#     any user.
#
#
#  "force_mask"  permissions.
#
# force_mask = ""
```

### Upstream Latest Release

Yes

```
$ rpm -qa | grep docker
docker-cli-29.4.0-1.fc45.aarch64
docker-buildkit-0.30.0-1.fc45.aarch64
docker-buildx-0.34.1-1.fc45.aarch64
$ rpm -qa | grep containerd
containerd-2.2.3-1.fc45.aarch64
```

### Additional information

mahendrarathore1742:
## Summary

This change aligns Buildah with Docker/BuildKit behavior by rejecting Dockerfile paths
inside the build context that resolve (via symlink) to a target outside the context.

## What Changed

- Validate Dockerfile paths in `imagebuildah.BuildDockerfiles` to ensure any symlink
  resolves within `ContextDirectory`.
- Add a regression test that reproduces the symlink-escape scenario.

## Why

Builds should not be able to read files outside the declared build context.
Docker/BuildKit fails in this case; Buildah previously allowed it. This fixes the
inconsistency and closes the security gap.

## Testing

```bash
go test ./imagebuildah \
  -run TestDockerfileSymlinkOutsideContext \
  -tags "containers_image_openpgp exclude_graphdriver_btrfs"
```

> **Note:** Full `go test ./...` fails in this container due to conformance tests
> requiring privileged operations (overlay mounts and chown).

## Closes

Closes #6861

### DIFF
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -141,6 +141,30 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 			if err != nil {
 				return "", nil, err
 			}
+			if options.ContextDirectory != "" {
+				contextAbs, err := filepath.Abs(options.ContextDirectory)
+				if err != nil {
+					return "", nil, fmt.Errorf("resolving context directory %q: %w", options.ContextDirectory, err)
+				}
+				contextAbs, err = filepath.EvalSymlinks(contextAbs)
+				if err != nil {
+					return "", nil, fmt.Errorf("resolving context directory %q: %w", options.ContextDirectory, err)
+				}
+				dfileAbs, err := filepath.Abs(dfile)
+				if err != nil {
+					return "", nil, fmt.Errorf("resolving containerfile %q: %w", dfile, err)
+				}
+				if rel, err := filepath.Rel(contextAbs, dfileAbs); err == nil && rel != ".." && !strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
+					resolved, err := filepath.EvalSymlinks(dfileAbs)
+					if err != nil {
+						return "", nil, fmt.Errorf("resolving containerfile %q: %w", dfile, err)
+					}
+					resolvedRel, err := filepath.Rel(contextAbs, resolved)
+					if err != nil || resolvedRel == ".." || strings.HasPrefix(resolvedRel, ".."+string(filepath.Separator)) {
+						return "", nil, fmt.Errorf("containerfile %q resolves to %q outside of the build context %q", dfile, resolved, contextAbs)
+					}
+				}
+			}
 
 			var contents *os.File
 			// If given a directory error out since `-f` does not supports path to directory
--- a/imagebuildah/build.go
+++ b/imagebuildah/build.go
@@ -160,7 +160,10 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 						return "", nil, fmt.Errorf("resolving containerfile %q: %w", dfile, err)
 					}
 					resolvedRel, err := filepath.Rel(contextAbs, resolved)
-					if err != nil || resolvedRel == ".." || strings.HasPrefix(resolvedRel, ".."+string(filepath.Separator)) {
+					if err != nil {
+						return "", nil, fmt.Errorf("resolving containerfile %q relative to context %q: %w", dfile, contextAbs, err)
+					}
+					if resolvedRel == ".." || strings.HasPrefix(resolvedRel, ".."+string(filepath.Separator)) {
 						return "", nil, fmt.Errorf("containerfile %q resolves to %q outside of the build context %q", dfile, resolved, contextAbs)
 					}
 				}
--- a/imagebuildah/build_linux_test.go
+++ b/imagebuildah/build_linux_test.go
@@ -38,6 +38,18 @@ func TestFilesClosedProperlyByBuildDockerfiles(t *testing.T) {
 	}
 }
 
+func TestDockerfileSymlinkOutsideContext(t *testing.T) {
+	tmpDir := t.TempDir()
+	contextDir := filepath.Join(tmpDir, "context")
+	assert.NoError(t, os.Mkdir(contextDir, 0o755))
+	dockerfile := filepath.Join(tmpDir, "Dockerfile")
+	assert.NoError(t, os.WriteFile(dockerfile, []byte("FROM scratch"), 0o644))
+	assert.NoError(t, os.Symlink("../Dockerfile", filepath.Join(contextDir, "Dockerfile")))
+
+	_, _, err := BuildDockerfiles(context.Background(), nil, define.BuildOptions{ContextDirectory: contextDir}, filepath.Join(contextDir, "Dockerfile"))
+	assert.ErrorContains(t, err, "outside of the build context")
+}
+
 // currentOpenFiles makes an effort at returning a map of which files are currently
 // open by our process. We don't fail if we can't follow symlinks from fds as this
 // perhaps they now longer exist between when we read them and when we tried to use
--- a/imagebuildah/build_linux_test.go
+++ b/imagebuildah/build_linux_test.go
@@ -7,6 +7,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


imagebuildah/build_linux_test.go
```diff
@@ -38,6 +38,18 @@ func TestFilesClosedProperlyByBuildDockerfiles(t *testing.T) {
 	}
 }
 
+func TestDockerfileSymlinkOutsideContext(t *testing.T) {
+	tmpDir := t.TempDir()
+	contextDir := filepath.Join(tmpDir, "context")
+	assert.NoError(t, os.Mkdir(contextDir, 0o755))
+	dockerfile := filepath.Join(tmpDir, "Dockerfile")
+	assert.NoError(t, os.WriteFile(dockerfile, []byte("FROM scratch"), 0o644))
+	assert.NoError(t, os.Symlink("../Dockerfile", filepath.Join(contextDir, "Dockerfile")))
```
nalind: Might want to add a second case where the target is an absolute path (`dockerfile` instead of `"../Dockerfile"`) to be sure.

imagebuildah/build.go
```diff
@@ -141,6 +141,30 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
...
+				if err != nil {
+					return "", nil, fmt.Errorf("resolving containerfile %q: %w", dfile, err)
+				}
+				if rel, err := filepath.Rel(contextAbs, dfileAbs); err == nil && rel != ".." && !strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
+					resolved, err := filepath.EvalSymlinks(dfileAbs)
+					if err != nil {
+						return "", nil, fmt.Errorf("resolving containerfile %q: %w", dfile, err)
+					}
+					resolvedRel, err := filepath.Rel(contextAbs, resolved)
+					if err != nil || resolvedRel == ".." || strings.HasPrefix(resolvedRel, ".."+string(filepath.Separator)) {
```
nalind: Can we split this block into two blocks, so that for cases where `err != nil` the error we return can wrap the error that `filepath.Rel()` returned?

nalind (changes requested):
This looks broadly correct to me.  I tend to run conformance tests under `podman unshare` if I need to run them locally, but they're passing in CI.  There are a couple of tweaks I'd like to see, though.

imagebuildah/build.go
mahendrarathore1742: Thanks for the review! I split the [```filepath.Rel()```](vscode-file://vscode-app/home/msr/Desktop/VSCode-linux-x64/resources/app/out/vs/code/electron-browser/workbench/workbench.html) error handling to preserve the wrapped error and added an absolute-path symlink case to the test.

nalind (approved):
LGTM, thanks!

imagebuildah/build.go
```diff
@@ -141,6 +141,33 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
...
+				contextAbs, err = filepath.EvalSymlinks(contextAbs)
+				if err != nil {
+					return "", nil, fmt.Errorf("resolving context directory %q: %w", options.ContextDirectory, err)
+				}
+				dfileAbs, err := filepath.Abs(dfile)
+				if err != nil {
+					return "", nil, fmt.Errorf("resolving containerfile %q: %w", dfile, err)
+				}
+				if rel, err := filepath.Rel(contextAbs, dfileAbs); err == nil && rel != ".." && !strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
+					resolved, err := filepath.EvalSymlinks(dfileAbs)
```
eriksjolund, line 158: Could this dereference symlinks outside of the context directory? 
(`EvalSymlinks()` on line 158)

nalind: Would `rel` not be "..", and also not have "../" as a prefix if `dfileAbs` was outside of `contextAbs`?

eriksjolund: Symlinks might first take us out of the context directory and then back in again.

It could also be single symlink that contains something like `../context/Dockerfile` as in the
the example https://github.com/podman-container-tools/buildah/issues/6861#issuecomment-4614473963 where docker fails but podman succeeds.

mahendrarathore1742: Thanks for raising this. The current check only verifies dfileAbs is under the context, so a symlink could still resolve outside (or escape then re-enter). I’ll add a second guard to ensure the resolved path is also within contextAbs (via filepath.Rel on the resolved path) and push an update.

eriksjolund: I added another docker behavior example https://github.com/podman-container-tools/buildah/issues/6861#issuecomment-4619440924

mahendrarathore1742: Thanks for adding the additional Docker behavior example! I've tested both scenarios and the fix now correctly handles:

1. `context/Dockerfile -> ../context/file` - Now correctly fails (matches Docker)
2. `context/Dockerfile -> ../file` - Now correctly succeeds (matches Docker)

The implementation has been updated and tested to ensure both the security fix and Docker behavior alignment are working correctly. The changes are pushed to the `fix/dockerfile-symlink-context` branch and ready for review.

TomSweeneyRedHat:
LGTM

TomSweeneyRedHat:
Some unhappy tests which I restarted.

nalind:
You're going to want to rebase to pick up missing packages that are causing those "pivot_root: command not found" test failures, added to the VM images that we updated to in #6911.

simonbrauner:
There seems to be an overlap with https://github.com/podman-container-tools/buildah/pull/6886 which, as far as I can tell, deals with the same problem but at the level of `DiscoverContainerfile` instead of `BuildDockerfiles`.

When I tested this PR, this corner case https://github.com/podman-container-tools/buildah/pull/6886#issuecomment-4958005777 differs from Docker behavior the same way as the other PR used to.

mahendrarathore1742:
@nalind Could you take another look when you get a chance? thanks

simonbrauner:
@mahendrarathore1742 could you align this with the resolution semantics in the PR i linked above? I don't think the way it is now ensures consistency with Docker. As I pointed out, Docker seems to resolve within root, while your fix just rejects.

```shell
# working directory somewhere in /home

# Docker resolves within root

$ ls -l Dockerfile 
lrwxrwxrwx. 1 sbrauner sbrauner 23 Aug 18 13:38 Dockerfile -> /tmp/test/Containerfile

## This fails even if `/tmp/test/Containerfile` exists, because it gets resolved as `/.../context/tmp/test/Containerfile` which does not.

$ docker build -t test -f Dockerfile .
ERROR: failed to build: failed to solve: failed to read dockerfile: open Dockerfile: no such file or directory

$ mkdir -p tmp/test
$ cat Dockerfile > tmp/test/Containerfile

## After the file is written, Docker finds the file successfully.

$ docker build -t test -f Dockerfile .
[+] Building 1.0s (6/6) FINISHED                                     docker:default
 => [internal] load build definition from Dockerfile                           0.0s
 => => transferring dockerfile: 427B                                           0.0s
 => [internal] load metadata for docker.io/library/alpine:3.20                 0.6s
 => [internal] load .dockerignore                                              0.0s
 => => transferring context: 2B                                                0.0s
 => [1/2] FROM docker.io/library/alpine:3.20@sha256:d9e853e87e55526f6b2917df9  0.0s
 => => resolve docker.io/library/alpine:3.20@sha256:d9e853e87e55526f6b2917df9  0.0s
 => CACHED [2/2] RUN echo done                                                 0.0s
 => exporting to image                                                         0.1s
 => => exporting layers                                                        0.0s
 => => exporting manifest sha256:3ea569c5124b0cf8432e3ec5b453de284efba0a212a5  0.0s
 => => exporting config sha256:c0d2074a6a13307ac0682fa286060506264c12379643c1  0.0s
 => => exporting attestation manifest sha256:3977c31a6a00b9d4a78ba37f3c78eba4  0.0s
 => => exporting manifest list sha256:8d75ad3f92b3ffe6d0d81b49286e5bb5a8bdb96  0.0s
 => => naming to docker.io/library/test:latest                                 0.0s
 => => unpacking to docker.io/library/test:latest                              0.0s

# Buildah after this PR rejects

## While in this PR, the absolute symlink is just rejected without looking at `/.../context/tmp/test/Containerfile`  at all.

$ buildah build -t test -f Dockerfile .
Error: containerfile "Dockerfile" resolves to "/tmp/test/Containerfile" outside of the build context "/home/..."
```

mahendrarathore1742:
> @mahendrarathore1742 could you align this with the resolution semantics in the PR i linked above? I don't think the way it is now ensures consistency with Docker. As I pointed out, Docker seems to resolve within root, while your fix just rejects.
> 
> ```shell
> # working directory somewhere in /home
> 
> # Docker resolves within root
> 
> $ ls -l Dockerfile 
> lrwxrwxrwx. 1 sbrauner sbrauner 23 Aug 18 13:38 Dockerfile -> /tmp/test/Containerfile
> 
> ## This fails even if `/tmp/test/Containerfile` exists, because it gets resolved as `/.../context/tmp/test/Containerfile` which does not.
> 
> $ docker build -t test -f Dockerfile .
> ERROR: failed to build: failed to solve: failed to read dockerfile: open Dockerfile: no such file or directory
> 
> $ mkdir -p tmp/test
> $ cat Dockerfile > tmp/test/Containerfile
> 
> ## After the file is written, Docker finds the file successfully.
> 
> $ docker build -t test -f Dockerfile .
> [+] Building 1.0s (6/6) FINISHED                                     docker:default
>  => [internal] load build definition from Dockerfile                           0.0s
>  => => transferring dockerfile: 427B                                           0.0s
>  => [internal] load metadata for docker.io/library/alpine:3.20                 0.6s
>  => [internal] load .dockerignore                                              0.0s
>  => => transferring context: 2B                                                0.0s
>  => [1/2] FROM docker.io/library/alpine:3.20@sha256:d9e853e87e55526f6b2917df9  0.0s
>  => => resolve docker.io/library/alpine:3.20@sha256:d9e853e87e55526f6b2917df9  0.0s
>  => CACHED [2/2] RUN echo done                                                 0.0s
>  => exporting to image                                                         0.1s
>  => => exporting layers                                                        0.0s
>  => => exporting manifest sha256:3ea569c5124b0cf8432e3ec5b453de284efba0a212a5  0.0s
>  => => exporting config sha256:c0d2074a6a13307ac0682fa286060506264c12379643c1  0.0s
>  => => exporting attestation manifest sha256:3977c31a6a00b9d4a78ba37f3c78eba4  0.0s
>  => => exporting manifest list sha256:8d75ad3f92b3ffe6d0d81b49286e5bb5a8bdb96  0.0s
>  => => naming to docker.io/library/test:latest                                 0.0s
>  => => unpacking to docker.io/library/test:latest                              0.0s
> 
> # Buildah after this PR rejects
> 
> ## While in this PR, the absolute symlink is just rejected without looking at `/.../context/tmp/test/Containerfile`  at all.
> 
> $ buildah build -t test -f Dockerfile .
> Error: containerfile "Dockerfile" resolves to "/tmp/test/Containerfile" outside of the build context "/home/..."
> ```

@simonbrauner I have rebased the branch onto latest main and updated the symlink resolution logic in BuildDockerfiles to use securejoin.SecureJoin (RESOLVE_IN_ROOT semantics).

Absolute symlinks inside the context (such as Dockerfile -> /tmp/test/Containerfile) are now re-rooted relative to the build context root (<context>/tmp/test/Containerfile), and .. traversals are clamped, matching Docker BuildKit's behavior and the semantics from #6886. I've also added unit tests covering these scenarios. PTAL when you have a moment!

imagebuildah/build.go
```diff
@@ -140,15 +141,43 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
 			}
 			data = resp.Body
 		} else {
-			dinfo, err := os.Stat(dfile)
-			if err != nil {
-				// If the Dockerfile isn't available, try again with
-				// context directory prepended (if not prepended yet).
-				if !strings.HasPrefix(dfile, options.ContextDirectory) {
```
simonbrauner, line 147: Originally, it was first attempted to find the containerfile in the working directory, and the context directory only got involved as a fallback. While in your change, it's the other way around.

So this is a breaking change. Consider the example:

```bash
$ tree
.
├── Containerfile
└── subdir
    └── Containerfile

$ cat Containerfile 
FROM alpine
RUN echo 'I am outside of subdir'

$ cat subdir/Containerfile 
FROM alpine
RUN echo 'I am in subdir'

# Before this PR
$ buildah build -f Containerfile -t 6861-test subdir
...
STEP 2/2: RUN echo 'I am outside of subdir'
...

# After this PR
$ buildah -f Containerfile -t 6861-test subdir
...
STEP 2/2: RUN echo 'I am in subdir'
...

$ sudo docker build -f Containerfile -t 6861-test subdir
...
[2/2] RUN echo 'I am outside of subdir'
...

```

imagebuildah/build.go
```diff
@@ -140,15 +141,43 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
...
+			if options.ContextDirectory != "" {
+				contextAbs, err := filepath.Abs(options.ContextDirectory)
+				if err != nil {
+					return "", nil, fmt.Errorf("resolving context directory %q: %w", options.ContextDirectory, err)
+				}
```
simonbrauner, line 148: I would propose moving this outside of the loop, as the context directory stays the same between the loop runs.

imagebuildah/build.go
```diff
@@ -140,15 +141,43 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
...
+
+				if rel, err := filepath.Rel(contextAbs, dfileAbs); err == nil && rel != ".." && !strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
+					resolved, err := securejoin.SecureJoin(contextAbs, rel)
+					if err != nil {
+						return "", nil, fmt.Errorf("resolving containerfile %q: %w", dfile, err)
+					}
+					dfile = resolved
+				}
+			} else {
+				if _, err := os.Lstat(dfile); err != nil && !filepath.IsAbs(dfile) && !strings.HasPrefix(dfile, options.ContextDirectory) {
```
simonbrauner, line 175: `!strings.HasPrefix(dfile, options.ContextDirectory)` is always false, since we are in the else branch of `options.ContextDirectory != ""`. Any string has the prefix of `""`, so the branch cannot run.

imagebuildah/build_linux_test.go
```diff
@@ -37,6 +37,93 @@ func TestFilesClosedProperlyByBuildDockerfiles(t *testing.T) {
 	}
 }
 
+func TestDockerfileSymlinkReRootedUnderContext(t *testing.T) {
```
simonbrauner: Thinking about whether this should be only tested on linux, as other systems have symlinks as well, right?

imagebuildah/build.go
```diff
@@ -140,15 +141,43 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
...
+			if options.ContextDirectory != "" {
+				contextAbs, err := filepath.Abs(options.ContextDirectory)
+				if err != nil {
+					return "", nil, fmt.Errorf("resolving context directory %q: %w", options.ContextDirectory, err)
+				}
+				contextAbs, err = filepath.EvalSymlinks(contextAbs)
+				if err != nil {
+					return "", nil, fmt.Errorf("resolving context directory %q: %w", options.ContextDirectory, err)
+				}
+				contextAbs = filepath.Clean(contextAbs)
```
simonbrauner, line 153: This is redundant, both `Abs` and `EvalSymlinks` run `Clean` internally.

imagebuildah/build.go
```diff
@@ -140,15 +141,43 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
...
-				if !strings.HasPrefix(dfile, options.ContextDirectory) {
+			if options.ContextDirectory != "" {
+				contextAbs, err := filepath.Abs(options.ContextDirectory)
+				if err != nil {
+					return "", nil, fmt.Errorf("resolving context directory %q: %w", options.ContextDirectory, err)
+				}
+				contextAbs, err = filepath.EvalSymlinks(contextAbs)
+				if err != nil {
+					return "", nil, fmt.Errorf("resolving context directory %q: %w", options.ContextDirectory, err)
+				}
```
simonbrauner, line 152: I think it would be better to remove this. The way `filepath.Rel` is used later breaks when the path to the context directory contains a symlink, as `contextAbs` would be without symlinks and `dFileAbs` would be with symlinks, so the paths would be considered different even though their locations in the filesystem would be the same.

imagebuildah/build.go
```diff
@@ -140,15 +141,43 @@ func BuildDockerfiles(ctx context.Context, store storage.Store, options define.B
...
+				if !filepath.IsAbs(dfileAbs) {
+					candidate := filepath.Join(contextAbs, dfile)
+					if _, err := os.Lstat(candidate); err == nil {
+						dfileAbs = candidate
+					} else if _, err := os.Lstat(dfile); err == nil {
+						dfileAbs, _ = filepath.Abs(dfile)
+					} else {
+						dfileAbs = candidate
+					}
+				}
```
simonbrauner, line 165: Something like this should fix the ordering problem, as well as the duplication of the `dfileAbs = candidate` line.

```suggestion
				dfileAbs := dfile
				if !filepath.IsAbs(dfileAbs) {
					if _, err := os.Lstat(dfileAbs); err == nil {
						dfileAbs, err = filepath.Abs(dfileAbs)
						if err != nil {
							return "", nil, fmt.Errorf("resolving containerfile %q: %w", dfile, err)
						}
					} else {
						dfileAbs = filepath.Join(contextAbs, dfileAbs)
					}
				}
```

simonbrauner (commented):
@mahendrarathore1742 I have some more feedback.

mahendrarathore1742:
> @mahendrarathore1742 I have some more feedback.

@simonbrauner Thanks for the detailed review! I've addressed all your feedback in the latest push:

1. File lookup order fixed — Working directory is now checked first, then falls back to  context directory, matching Docker and original Buildah behavior.
2. contextAbs moved outside the loop — No longer recomputed on each iteration.
3. Dead else branch removed — The unreachable code with empty ContextDirectory is gone.
4. Redundant ```filepath.Clean removed ```— Already handled by ```filepath.Abs```.
5. EvalSymlinks on context directory removed — Avoids path mismatch when context path contains symlinks.
6. Applied your suggested code for the dfileAbs resolution block, including proper error handling for filepath.Abs.

PTAL when you get a chance!

PR #6886, opened 2026-06-02
Reject symlinked Containerfile and ignore files escaping build context

### LINKED ISSUE #6861: `buildah build` follows `Dockerfile` symlink outside the context directory

### LINKED ISSUE #28749: `podman build` follows `Dockerfile.dockerignore` symlink outside the context directory
### Issue Description

`podman build dir` follows `Dockerfile.dockerignore`, `Containerfile.dockerignore` and `Containerfile.containerignore`  symlinks to a file outside the context directory.

This does not match the behavior of `docker build dir` (using docker 29.4.0 and buildkit 0.30.0).

Steps to reproduce the issue
1. Create file `~/test.bash` containing
   ```
   #!/bin/bash
   set -o errexit
   set -o nounset
   command=$1
   containerfile=$2
   symlink=$3
   dir=$(mktemp -d)
   mkdir $dir/context
   
   cat <<EOF > $dir/context/$containerfile
   FROM docker.io/library/alpine
   COPY * /dir/
   EOF
   
   touch $dir/context/ign
   touch $dir/context/file
   (cd $dir/context && ln -s ../ign $symlink)
   echo file > $dir/ign
   $command build -q --no-cache -t test $dir/context > /dev/null
   $command run --rm test ls -l /dir/file
   ```
2. Test arguments `podman Dockerfile Dockerfile.dockerignore`
   ```
   $ bash test.bash podman Dockerfile Dockerfile.dockerignore
   ls: /dir/file: No such file or directory
   $
   ```
3. Test arguments `docker Dockerfile Dockerfile.dockerignore`. Set environment `DOCKER_BUILDKIT=1` to enable buildkit.
   ```
   $ DOCKER_BUILDKIT=1 bash test.bash docker Dockerfile Dockerfile.dockerignore
   -rw-r--r--    1 root     root             0 May 21 17:31 /dir/file
   $
   ```
4. Test arguments `podman Containerfile Containerfile.dockerignore`
   ```
   $ bash test.bash podman Containerfile Containerfile.dockerignore
   ls: /dir/file: No such file or directory
   $
   ```
5. Test arguments `podman Containerfile Containerfile.containerignore`
   ```
   $ bash test.bash podman Containerfile Containerfile.containerignore
   ls: /dir/file: No such file or directory
   $
   ```

In step 2, the file `/dir/file` is missing in the image.
In step 3, the file `/dir/file` exists in the image.
In step 4, the file `/dir/file` is missing in the image.
In step 5, the file `/dir/file` is missing in the image.

In step 2, the file `/dir/file` exists in the image.
In step 3, the file `/dir/file` exists in the image.
In step 4, the file `/dir/file` exists in the image.
In step 5, the file `/dir/file` exists in the image.

### podman info output

```yaml
host:
  arch: arm64
  buildahVersion: 1.43.1
  cgroupControllers:
  - cpu
  - memory
  - pids
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.2.1-2.fc45.aarch64
    path: /usr/bin/conmon
    version: 'conmon version 2.2.1, commit: '
  cpuUtilization:
    idlePercent: 99.8
    systemPercent: 0.14
    userPercent: 0.06
  cpus: 2
  databaseBackend: sqlite
  distribution:
    distribution: fedora
    variant: coreos
    version: "45"
  emulatedArchitectures:
  - linux/386
  - linux/amd64
  eventLogger: journald
  freeLocks: 2048
  hostname: localhost.localdomain
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1088
      size: 1
    - container_id: 1
      host_id: 2162688
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1087
      size: 1
    - container_id: 1
      host_id: 2162688
      size: 65536
  kernel: 7.0.0-62.fc45.aarch64
  linkmode: dynamic
  logDriver: journald
  memFree: 131178496
  memTotal: 2037288960
  networkBackend: netavark
  networkBackendInfo:
    backend: netavark
    defaultNetwork: podman
    dns:
      package: aardvark-dns-1.17.1-1.fc45.aarch64
      path: /usr/libexec/podman/aardvark-dns
      version: aardvark-dns 1.17.1
    package: netavark-1.17.2-1.fc45.aarch64
    path: /usr/libexec/podman/netavark
    version: netavark 1.17.2
  ociRuntime:
    name: crun
    package: crun-1.27-1.fc45.aarch64
    path: /usr/bin/crun
    version: |-
      crun version 1.27
      commit: a718a92cc9a94955a5a550b6fdec1378c247ec50
      rundir: /run/user/1087/crun
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +LIBKRUN +WASM:wasmedge +YAJL
  os: linux
  pasta:
    executable: /usr/bin/pasta
    package: passt-0^20260120.g386b5f5-1.fc44.aarch64
    version: |
      pasta 0^20260120.g386b5f5-1.fc44.aarch64-pasta
      Copyright Red Hat
  remoteSocket:
    exists: true
    path: /run/user/1087/podman/podman.sock
  rootlessNetworkCmd: pasta
  security:
    apparmorEnabled: false
    rootless: true
    seccompEnabled: true
    selinuxEnabled: true
  serviceIsRemote: false
  slirp4netns:
    executable: /usr/bin/slirp4netns
    package: slirp4netns-1.3.1-4.fc44.aarch64
    version: |-
      slirp4netns version 1.3.1
      commit: e5e368c4f5db6ae75c2fce786e31eef9da6bf236
      libslirp: 4.9.1
      SLIRP_CONFIG_VERSION_MAX: 6
      libseccomp: 2.6.0
  swapFree: 0
  swapTotal: 0
  uptime: 10h 49m 15.00s (Approximately 0.42 days)
  variant: v8
plugins:
  authorization: null
  log:
  - k8s-file
  - none
  - passthrough
  - journald
  network:
  - bridge
  - macvlan
  - ipvlan
  volume:
  - local
registries:
  search:
  - registry.fedoraproject.org
  - registry.access.redhat.com
  - docker.io
store:
  configFile: /var/home/test11/.config/containers/storage.conf
  containerStore:
    number: 0
    paused: 0
    running: 0
    stopped: 0
  graphDriverName: overlay
  graphOptions: {}
  graphRoot: /var/home/test11/.local/share/containers/storage
  graphRootAllocated: 10132369408
  graphRootUsed: 9404702720
  graphStatus:
    Backing Filesystem: xfs
    Native Overlay Diff: "true"
    Supports d_type: "true"
    Supports shifting: "false"
    Supports volatile: "true"
    Using metacopy: "false"
  imageCopyTmpDir: /var/tmp
  imageStore:
    number: 30
  runRoot: /run/user/1087/containers
  transientStore: false
  volumePath: /var/home/test11/.local/share/containers/storage/volumes
version:
  APIVersion: 5.8.2
  BuildOrigin: Fedora Project
  Built: 1776124800
  BuiltTime: Tue Apr 14 02:00:00 2026
  GitCommit: 5b263b5f5b48004a87caac44e67349a8266d9ef4
  GoVersion: go1.26.2-X:nodwarf5
  Os: linux
  OsArch: linux/arm64
  Version: 5.8.2
```

### Podman in a container

No

### Privileged Or Rootless

Rootless

### Upstream Latest Release

Yes

### Additional information

```
$ rpm -qa | grep docker
docker-cli-29.4.0-1.fc45.aarch64
docker-buildkit-0.30.0-1.fc45.aarch64
docker-buildx-0.34.1-1.fc45.aarch64
$ rpm -qa | grep containerd
containerd-2.2.3-1.fc45.aarch64
```

Honny1:
Fixes: https://github.com/podman-container-tools/buildah/issues/6861 
Fixes: https://github.com/podman-container-tools/podman/issues/28749

#### What type of PR is this?

> /kind api-change
> 
/kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
Buildah no longer follows symlinked Containerfile/Dockerfile or ignore files (e.g. Dockerfile.dockerignore) that point outside the build context directory, matching docker build behavior.
```

### DIFF
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -1422,37 +1422,66 @@ func ContainerIgnoreFile(contextDir, path string, containerFiles []string) ([]st
 		excludes, err := imagebuilder.ParseIgnore(path)
 		return excludes, path, err
 	}
-	// If path was not supplied give priority to `<containerfile>.containerignore` first.
+	// If path was not supplied, look for `<containerfile>.dockerignore` and
+	// `<containerfile>.containerignore`. When both exist, `.containerignore` wins.
+	// securejoin confines lookups with RESOLVE_IN_ROOT semantics, matching
+	// Docker BuildKit behavior. When the containerfile is inside contextDir
+	// we resolve relative to contextDir; when it is outside (e.g.
+	// overlay-mounted context or -f pointing elsewhere) we resolve relative
+	// to the containerfile's parent directory.
 	for _, containerfile := range containerFiles {
 		if !filepath.IsAbs(containerfile) {
 			containerfile = filepath.Join(contextDir, containerfile)
 		}
-		containerfileIgnore := ""
-		if err := fileutils.Exists(containerfile + ".containerignore"); err == nil {
-			containerfileIgnore = containerfile + ".containerignore"
+		cleanPath := filepath.Clean(containerfile)
+		relPath, relErr := filepath.Rel(contextDir, cleanPath)
+		insideContext := relErr == nil && relPath != ".." && !strings.HasPrefix(relPath, ".."+string(filepath.Separator))
+		var rootDir, baseName string
+		if insideContext {
+			rootDir = contextDir
+			baseName = relPath
+		} else {
+			rootDir = filepath.Dir(cleanPath)
+			baseName = filepath.Base(cleanPath)
 		}
-		if err := fileutils.Exists(containerfile + ".dockerignore"); err == nil {
-			containerfileIgnore = containerfile + ".dockerignore"
+		excludes, resolved, err := findIgnoreFile(rootDir, baseName+".dockerignore", baseName+".containerignore")
+		if err != nil {
+			return nil, "", err
 		}
-		if containerfileIgnore != "" {
-			excludes, err := imagebuilder.ParseIgnore(containerfileIgnore)
-			return excludes, containerfileIgnore, err
+		if resolved != "" {
+			return excludes, resolved, nil
 		}
 	}
-	path, symlinkErr := securejoin.SecureJoin(contextDir, ".containerignore")
-	if symlinkErr != nil {
-		return nil, "", symlinkErr
+	excludes, resolved, err := findIgnoreFile(contextDir, ".dockerignore", ".containerignore")
+	if err != nil {
+		return nil, "", err
 	}
-	excludes, err := imagebuilder.ParseIgnore(path)
-	if errors.Is(err, os.ErrNotExist) {
-		path, symlinkErr = securejoin.SecureJoin(contextDir, ".dockerignore")
-		if symlinkErr != nil {
-			return nil, "", symlinkErr
+	if resolved != "" {
+		return excludes, resolved, nil
+	}
+	return nil, "", nil
+}
+
+// findIgnoreFile tries each candidate name resolved under rootDir using
+// securejoin (RESOLVE_IN_ROOT semantics). When both exist the last one wins.
+func findIgnoreFile(rootDir string, candidates ...string) ([]string, string, error) {
+	var excludes []string
+	var matched string
+	for _, name := range candidates {
+		resolved, err := securejoin.SecureJoin(rootDir, name)
+		if err != nil {
+			continue
 		}
-		excludes, err = imagebuilder.ParseIgnore(path)
+		f, err := os.Open(resolved)
+		if err != nil {
+			continue
+		}
+		excludes, err = imagebuilder.ParseIgnoreReader(f)
+		f.Close()
+		if err != nil {
+			return nil, "", err
+		}
+		matched = resolved
 	}
-	if errors.Is(err, os.ErrNotExist) {
-		return excludes, "", nil
-	}
-	return excludes, path, err
+	return excludes, matched, nil
 }
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -2,10 +2,12 @@ package util //nolint:revive,nolintlint
 
 import (
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
 
+	securejoin "github.com/cyphar/filepath-securejoin"
 	"go.podman.io/buildah/pkg/parse"
 )
 
@@ -41,42 +43,77 @@ func MirrorToTempFileIfPathIsDescriptor(file string) (string, bool) {
 }
 
 // DiscoverContainerfile tries to find a Containerfile or a Dockerfile within the provided `path`.
+// The path may be a directory (in which case Containerfile/Dockerfile is searched inside it)
+// or a direct path to a container file.
+//
+// When path is a directory (or a symlink to one), Containerfile/Dockerfile
+// candidates are resolved with RESOLVE_IN_ROOT semantics so that symlinks
+// escaping the context are clamped back.  When path points directly at a
+// file (or a symlink to one), it is returned as-is.
 func DiscoverContainerfile(path string) (foundCtrFile string, err error) {
-	// Test for existence of the file
-	target, err := os.Stat(path)
+	path, err = filepath.Abs(path)
 	if err != nil {
 		return "", fmt.Errorf("discovering Containerfile: %w", err)
 	}
 
-	switch mode := target.Mode(); {
-	case mode.IsDir():
-		// If the path is a real directory, we assume a Containerfile or a Dockerfile within it
-		ctrfile := filepath.Join(path, "Containerfile")
-
-		// Test for existence of the Containerfile file
-		file, err := os.Stat(ctrfile)
-		if err != nil {
-			// See if we have a Dockerfile within it
-			ctrfile = filepath.Join(path, "Dockerfile")
-
-			// Test for existence of the Dockerfile file
-			file, err = os.Stat(ctrfile)
-			if err != nil {
-				return "", fmt.Errorf("cannot find Containerfile or Dockerfile in context directory: %w", err)
-			}
-		}
-
-		// The file exists, now verify the correct mode
-		if mode := file.Mode(); mode.IsRegular() {
-			foundCtrFile = ctrfile
-		} else {
-			return "", fmt.Errorf("assumed Containerfile %q is not a file", ctrfile)
-		}
-
-	case mode.IsRegular():
-		// If the context dir is a file, we assume this as Containerfile
-		foundCtrFile = path
+	target, err := os.Lstat(path)
+	if err != nil {
+		return "", fmt.Errorf("discovering Containerfile: %w", err)
 	}
 
-	return foundCtrFile, nil
+	// If path is a symlink to a directory (e.g. the build context itself is
+	// a symlink), follow it so the IsDir() branch handles it.
+	if target.Mode()&os.ModeSymlink != 0 {
+		if realInfo, err := os.Stat(path); err == nil && realInfo.IsDir() {
+			target = realInfo
+		}
+	}
+
+	switch {
+	case target.IsDir():
+		for _, name := range []string{"Containerfile", "Dockerfile"} {
+			ctrfile := filepath.Join(path, name)
+			if resolved, ok := isRegularFileInContext(path, ctrfile); ok {
+				return resolved, nil
+			}
+		}
+		return "", fmt.Errorf("cannot find Containerfile or Dockerfile in context directory: %w", fs.ErrNotExist)
+
+	case target.Mode().IsRegular():
+		return path, nil
+
+	case target.Mode()&os.ModeSymlink != 0:
+		if fi, err := os.Stat(path); err == nil && fi.Mode().IsRegular() {
+			return path, nil
+		}
+		return "", fmt.Errorf("assumed Containerfile %q is not a file", path)
+
+	default:
+		return "", fmt.Errorf("assumed Containerfile %q is not a file", path)
+	}
+}
+
+// isRegularFileInContext checks whether path resolves to a regular file
+// inside contextDir using RESOLVE_IN_ROOT semantics (securejoin.SecureJoin):
+// ".." components are clamped to the root and absolute symlink targets are
+// re-rooted under contextDir.  This matches Docker BuildKit's behavior.
+//
+// On success it returns the resolved host path.
+func isRegularFileInContext(contextDir, path string) (string, bool) {
+	name, err := filepath.Rel(contextDir, path)
+	if err != nil {
+		return "", false
+	}
+	resolved, err := securejoin.SecureJoin(contextDir, name)
+	if err != nil {
+		return "", false
+	}
+	fi, err := os.Stat(resolved)
+	if err != nil {
+		return "", false
+	}
+	if !fi.Mode().IsRegular() {
+		return "", false
+	}
+	return resolved, true
 }
--- a/pkg/util/util_test.go
+++ b/pkg/util/util_test.go
@@ -1,10 +1,21 @@
 package util //nolint:revive,nolintlint
 
 import (
+	"io/fs"
+	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
+
+func absPath(t *testing.T, rel string) string {
+	t.Helper()
+	p, err := filepath.Abs(rel)
+	require.NoError(t, err)
+	return p
+}
 
 func TestDiscoverContainerfile(t *testing.T) {
 	t.Parallel()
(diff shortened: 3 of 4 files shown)


pkg/parse/parse.go
```diff
@@ -1404,16 +1404,19 @@ func ContainerIgnoreFile(contextDir, path string, containerFiles []string) ([]st
...
-		}
+		base := filepath.Base(containerfile)
 		containerfileIgnore := ""
-		if err := fileutils.Exists(containerfile + ".containerignore"); err == nil {
-			containerfileIgnore = containerfile + ".containerignore"
-		}
-		if err := fileutils.Exists(containerfile + ".dockerignore"); err == nil {
-			containerfileIgnore = containerfile + ".dockerignore"
+		for _, suffix := range []string{".containerignore", ".dockerignore"} {
+			candidate, err := securejoin.SecureJoin(contextDir, base+suffix)
```
nalind: We should avoid starting to assume that the Dockerfile/Containerfile, and in this case their associated ignore files, are always in the top-level directory of the build context.  That's not guaranteed.

tests/bud.bats
```diff
@@ -8294,6 +8294,122 @@ srv.serve_forever()
...
+  local tarsrc=${TEST_SCRATCH_DIR}/tarsrc
+  mkdir -p ${tarsrc}
+  ln -s ${targetfile} ${tarsrc}/Dockerfile
+  local context_tar=${TEST_SCRATCH_DIR}/context.tar
+  tar -cf ${context_tar} -C ${tarsrc} Dockerfile
+
+  run_buildah 125 build $WITH_POLICY_JSON - < ${context_tar}
+  assert "$output" =~ "cannot find Containerfile or Dockerfile"
+  assert "$output" !~ "SECRET_CONTENT"
+
```
```suggestion
```

Honny1:

pkg/util/util.go
```diff
@@ -41,42 +41,68 @@ func MirrorToTempFileIfPathIsDescriptor(file string) (string, bool) {
...
+	if err != nil {
+		return false
+	}
+	rel, err := filepath.Rel(cleanContext, resolved)
+	if err != nil {
+		return false
+	}
+	if rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
+		return false
+	}
```
nalind, line 114: This appears to run into the same issue @eriksjolund pointed out in https://github.com/podman-container-tools/buildah/pull/6885#discussion_r3349587232.

pkg/util/util.go
```diff
@@ -41,42 +41,68 @@ func MirrorToTempFileIfPathIsDescriptor(file string) (string, bool) {
...
-		foundCtrFile = path
+	default:
+		return "", fmt.Errorf("assumed Containerfile %q is not a file", path)
 	}
+}
 
-	return foundCtrFile, nil
+// isRegularFileInContext returns true if path is a regular file (or a symlink
+// to one) whose real target is inside contextDir.
+func isRegularFileInContext(contextDir, path string) bool {
```
mtrmac: (A drive-by, I have really only reviewed this part:) I think this is fine — works on Unix systems, and on Windows `filepath.Rel` rejects inputs which differ in the volume letter.

---

In principle, I think building an explicit `os.Root` and working within that would be structurally much safer, but looking at just this diff, that would probably be an invasive change and affect public API. (I didn’t investigate how difficult that would be.)

tests/bud.bats
```diff
@@ -8294,6 +8294,156 @@ srv.serve_forever()
...
+  _prefetch alpine
+
+  local contextdir=${TEST_SCRATCH_DIR}/context
+  mkdir -p ${contextdir}
+  cat > ${contextdir}/file << _EOF
+FROM alpine
+RUN echo traversal-back-inside-works
+_EOF
+  # Symlink path goes ../context/file -- traverses outside but real target resolves inside.
+  (cd ${contextdir} && ln -s ../context/file Dockerfile)
```
```suggestion
  ln -s ../context/file ${contextdir}/Dockerfile
```

tests/bud.bats
```diff
@@ -8294,6 +8294,156 @@ srv.serve_forever()
...
+
+  run_buildah build $WITH_POLICY_JSON ${contextdir}
+  assert "$output" =~ "traversal-back-inside-works"
```
nalind: I think this is expected to fail while attempting to read ${contextdir}/context/file.

tests/bud.bats
```diff
@@ -8294,6 +8294,156 @@ srv.serve_forever()
...
+  mkdir -p ${contextdir}
+  echo "file" > ${dir}/ign
+
+  cat > ${contextdir}/Dockerfile << _EOF
+FROM alpine
+COPY file /dir/
+RUN test -f /dir/file
+_EOF
+  touch ${contextdir}/file
+  (cd ${contextdir} && ln -s ../ign Dockerfile.dockerignore)
```
```suggestion
  ln -s ../ign ${contextdir}/Dockerfile.dockerignore
```

nalind (changes requested):
A subtle bug in how the symlinks get resolved, and I don't think I understand why some of these tests are using a subshell to run `ln -s`.

pkg/util/util.go
mtrmac: If you mean "../../correctly-guessed/directory/name/Containerfile" pointing us back at "./Containerfile": I think to reject that we would truly need `os.Root`, or an equivalent manual use of `openat2` `RESOLVE_BENEATH` (+ some implementation on non-Linux?) — or an entirely new manual symlink resolving code.

(`filepath.SecureJoin` is an ~equivalent of `RESOLVE_IN_ROOT`, it would not reject such links but interpret them differently.)

pkg/util/util.go
```diff
@@ -41,42 +41,84 @@ func MirrorToTempFileIfPathIsDescriptor(file string) (string, bool) {
...
+	if err != nil {
+		return false
+	}
+	if info.Mode()&os.ModeSymlink != 0 {
+		target, err := os.Readlink(path)
+		if err != nil {
+			return false
+		}
+		clean := filepath.Clean(target)
+		if !filepath.IsAbs(target) && (clean == ".." || strings.HasPrefix(clean, ".."+string(filepath.Separator))) {
```
mtrmac: Shouldn’t absolute symlinks be treated as a clear attempt to break out?

pkg/util/util.go
```diff
@@ -41,42 +41,84 @@ func MirrorToTempFileIfPathIsDescriptor(file string) (string, bool) {
...
+// isRegularFileInContext returns true if path is a regular file (or a symlink
+// to one) whose real target is inside contextDir.  Symlinks whose raw target
+// traverses outside the context at any intermediate step are rejected, even if
+// the final resolved path lands back inside.
+func isRegularFileInContext(contextDir, path string) bool {
+	cleanContext, err := filepath.EvalSymlinks(contextDir)
+	if err != nil {
+		return false
+	}
+	info, err := os.Lstat(path)
```
mtrmac: Couldn’t `path` have symlinks pointing outside of the target directory in the _middle_? `./symlink-to-etc/password`. where `symlink-to-etc` is `../../../../../…/etc`?

pkg/util/util.go
mtrmac: `path` being `dir1/dir2/symlink` with `symlink` pointing to `../Containerfile` does not actually escape.

mtrmac (commented):
Just a quick look: This is getting increasingly convoluted.

If the goal is to solve https://github.com/podman-container-tools/buildah/pull/6886/changes#r3363568131, I think refactoring to build around `os.Root` would ultimately be safer and much easier to prove correct. _But_ it might be very invasive, I didn’t check – I’ll let Buildah experts make the decision on whether this should happen.

Honny1:
I used `os.Root` in the parts that I touched. I think that refactoring makes sense to me, but not in this PR. Maybe we should create an issue for this.

Honny1:
PTAL @nalind @mtrmac

tests/bud.bats
```diff
@@ -8296,6 +8294,158 @@ srv.serve_forever()
...
+  cat > ${contextdir}/file << _EOF
+FROM alpine
+RUN echo SHOULD-NOT-RUN
+_EOF
+  # Symlink to ../file -- target does not exist on the filesystem.
+  ln -s ../file ${contextdir}/Dockerfile
+
+  run_buildah 125 build $WITH_POLICY_JSON ${contextdir}
+  expect_output --substring "cannot find Containerfile or Dockerfile"
+  assert "$output" !~ "SHOULD-NOT-RUN"
```
nalind: If we're treating ${contextdir} like the root of a container rootfs (which I think approximates the desired behavior for resolving symlinks found inside of a context directory), I would have expected the target of the symlink to be readable.  I believe this is an example from https://github.com/podman-container-tools/buildah/issues/6861#issuecomment-4614473963.

tests/bud.bats
```diff
@@ -8296,6 +8294,158 @@ srv.serve_forever()
...
+# https://github.com/podman-container-tools/buildah/issues/6861
+@test "bud with stdin tar context rejects symlinked Dockerfile pointing outside temp dir" {
+  local targetfile=${TEST_SCRATCH_DIR}/targetfile
+  echo "SECRET_CONTENT" > ${targetfile}
+
+  local tarsrc=${TEST_SCRATCH_DIR}/tarsrc
+  mkdir -p ${tarsrc}
+  ln -s ${targetfile} ${tarsrc}/Dockerfile
+  local context_tar=${TEST_SCRATCH_DIR}/context.tar
+  tar -cf ${context_tar} -C ${tarsrc} Dockerfile
```
nalind: Might want a `--owner=0:0 --numeric-owner` here.

tests/bud.bats
```diff
@@ -8296,6 +8294,158 @@ srv.serve_forever()
...
+@test "bud with http tar context rejects symlinked Dockerfile pointing outside temp dir" {
+  local targetfile=${TEST_SCRATCH_DIR}/targetfile
+  echo "SECRET_CONTENT" > ${targetfile}
+
+  local tarsrc=${TEST_SCRATCH_DIR}/tarsrc
+  mkdir -p ${tarsrc}
+  ln -s ${targetfile} ${tarsrc}/Dockerfile
+  local contentdir=${TEST_SCRATCH_DIR}/content
+  mkdir -p ${contentdir}
+  tar -cf ${contentdir}/context.tar -C ${tarsrc} Dockerfile
```
nalind: Might want a `--owner=0:0 --numeric-owner` here.

nalind (commented):
You're going to want to rebase to pick up missing packages that are causing those "pivot_root: command not found" test failures, added to the VM images that we updated to in #6911.

tests/bud.bats
Honny1: Docker also rejects symlinks that escape and return. I prefer matching Docker behavior.

Honny1:

inknos:
to me looks like all the comments are addressed in the code, but a confirmation from @nalind would be nice.  LGTM from my side

Honny1:
I did rebase on upstream main.

inknos (approved):

simonbrauner:
I found another incompatibility with Docker. If Docker sees an absolute symlink, it seems to assume that it is relative to the context directory.

Example:

```shell
$ tree
.
├── Dockerfile -> /subdirectory/remoteContainerfile
├── subdirectory
│   └── remoteContainerfile

$ buildah build -t demo .
Error: cannot find Containerfile or Dockerfile in context directory

$ docker build -t demo .
[+] Building 0.6s (6/6) FINISHED                                                                                             docker:default
 => [internal] load build definition from Dockerfile                                                                                   0.0s
 => => transferring dockerfile: 369B                                                                                                   0.0s
 => [internal] load metadata for docker.io/library/alpine:3.20                                                                         0.2s
 => [internal] load .dockerignore                                                                                                      0.0s
 => => transferring context: 2B                                                                                                        0.0s
 => [1/2] FROM docker.io/library/alpine:3.20@sha256:d9e853e87e55526f6b2917df91a2115c36dd7c696a35be12163d44e6e2a4b6bc                   0.0s
 => => resolve docker.io/library/alpine:3.20@sha256:d9e853e87e55526f6b2917df91a2115c36dd7c696a35be12163d44e6e2a4b6bc                   0.0s
 => CACHED [2/2] RUN echo test                                                                                                         0.0s
 => exporting to image                                                                                                                 0.2s
 => => exporting layers                                                                                                                0.0s
 => => exporting manifest sha256:e1411423e0df8c2b0076af6f325c2eda5c2ba1265f25a44e872894761079b209                                      0.0s
 => => exporting config sha256:e0e31f31aa3e5a3b9948bac3fa41774ba71ee2fc6f388137fbf3ca6f330a4d87                                        0.0s
 => => exporting attestation manifest sha256:b5718437215f4920ed86f4b2d0bacbe9f3e9c1a48db51a2c033ee6b2111a5241                          0.0s
 => => exporting manifest list sha256:95ea0c2f6f8be79fa4fbcd3ce7052e35c2e61f5f4695ec9ee4a3573ed638482a                                 0.0s
 => => naming to docker.io/library/demo:latest                                                                                         0.0s
 => => unpacking to docker.io/library/demo:latest                                                                                      0.0s
WARNING: current commit information was not captured by the build: failed to read current commit information with git rev-parse --is-inside-work-tree
```

To fix that, I suppose `securejoin.SecureJoin` could be used to join the context directory and the symlink. And then checking if the result points to a file or not.

Honny1:
@simonbrauner I think Docker is behaving incorrectly. What if the symlink points to `/etc/passwd`?

simonbrauner:
> @simonbrauner I think Docker is behaving incorrectly. What if the symlink points to `/etc/passwd`?

The `/etc/passwd` of the host is safe as long as the build does not happen at `/`. It is quirky to have a symlink that does not work on its own but works in `build`, but technically we cannot say that anything `escaped build context` in that behavior, right?

inknos:
> What if the symlink points to /etc/passwd?

I think this fix came from an upstream report where the case was exactly this "what if". 

> but technically we cannot say that anything escaped build context in that behavior, right?

yeah but preventing it from outside `build` is better imho

pkg/util/util.go
```diff
@@ -41,42 +41,67 @@ func MirrorToTempFileIfPathIsDescriptor(file string) (string, bool) {
...
-				return "", fmt.Errorf("cannot find Containerfile or Dockerfile in context directory: %w", err)
+	switch {
+	case target.IsDir():
+		for _, name := range []string{"Containerfile", "Dockerfile"} {
+			ctrfile := filepath.Join(path, name)
+			if isRegularFileInContext(path, ctrfile) {
+				return ctrfile, nil
 			}
 		}
+		return "", fmt.Errorf("cannot find Containerfile or Dockerfile in context directory")
```
nalind: Is this supposed to be wrapping `syscall.ENOENT`?

tests/bud.bats
nalind: This succeeds for me:
```bash
t=$(mktemp -d)
cat > $t/file << EOF
FROM busybox
RUN echo should run
EOF
ln -s ../file $t/Dockerfile
env DOCKER_BUILDKIT=1 docker build --no-cache --progress=plain $t
```

pkg/util/util.go
nalind: I would expect "../../correctly-guessed/directory/name/Containerfile" to end up resolving to "./correctly-guessed/directory/name/Containerfile", the initial ".." being resolved to "." at the context directory.

Honny1:
I updated the code to consistently use `securejoin.SecureJoin` (`RESOLVE_IN_ROOT`) for both Containerfile discovery and ignore file resolution, replacing the previous use of `os.Root` (`RESOLVE_BENEATH`). This should match Docker BuildKit's behavior: `..` components are clamped to the context root, and absolute symlink targets are re-rooted rather than being rejected outright. 

PTAL @nalind @inknos @simonbrauner @mtrmac

pkg/util/util.go
```diff
@@ -41,42 +43,68 @@ func MirrorToTempFileIfPathIsDescriptor(file string) (string, bool) {
...
-	case mode.IsDir():
-		// If the path is a real directory, we assume a Containerfile or a Dockerfile within it
-		ctrfile := filepath.Join(path, "Containerfile")
-
-		// Test for existence of the Containerfile file
-		file, err := os.Stat(ctrfile)
-		if err != nil {
-			// See if we have a Dockerfile within it
-			ctrfile = filepath.Join(path, "Dockerfile")
+	target, err := os.Lstat(path)
```
simonbrauner, line 59: If `lstat` is ran there, it seems to cause trouble when the symlink is to a directory.

```shell
$ ls -l
lrwxrwxrwx. 1 sbrauner sbrauner   16 Jul 14 14:36 test-repository5 -> test-repository3

$ cd test-repository5

$ ls Dockerfile 
Dockerfile

$ docker build -t demo .
[+] Building 1.3s (6/6) FINISHED                                     docker:default

$ buildah build -t demo .
Error: assumed Containerfile "test-repository5" is not a file
```

I would propose either a special case for directories, or delaying the `lstat` call after we know that it is not a directory. Because building in a directory which is a symlink is a valid use case, right?

simonbrauner (commented):
Added one comment.

And my comment at https://github.com/podman-container-tools/buildah/pull/6886#issuecomment-4958005777 is addressed.

pkg/util/util.go
Honny1: Yes, that seems right. I will take a look.

pkg/util/util_test.go
```diff
@@ -16,17 +26,88 @@ func TestDiscoverContainerfile(t *testing.T) {
...
+	secretFile := filepath.Join(tmpDir, "secret-Containerfile")
+	require.NoError(t, os.WriteFile(secretFile, []byte("FROM scratch\n"), 0o644))
+
+	contextDir := filepath.Join(tmpDir, "context")
+	require.NoError(t, os.Mkdir(contextDir, 0o755))
+	require.NoError(t, os.Symlink(secretFile, filepath.Join(contextDir, "Containerfile")))
+
+	_, err := DiscoverContainerfile(contextDir)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "cannot find Containerfile or Dockerfile")
```
nalind: This can be changed to an `assert.ErrorIs()` call.

pkg/util/util_test.go
```diff
@@ -16,17 +26,88 @@ func TestDiscoverContainerfile(t *testing.T) {
...
+
+	contextDir := filepath.Join(tmpDir, "context")
+	require.NoError(t, os.Mkdir(contextDir, 0o755))
+	require.NoError(t, os.WriteFile(filepath.Join(contextDir, "file"), []byte("FROM scratch\n"), 0o644))
+	require.NoError(t, os.Symlink("../file", filepath.Join(contextDir, "Containerfile")))
+
+	name, err := DiscoverContainerfile(contextDir)
+	require.NoError(t, err)
+	assert.Equal(t, filepath.Join(contextDir, "file"), name)
+}
```
nalind, line 87: This, or another test, should also probably check that when the symlink target is something like `../../subdir/file`, it successfully finds `filepath.Join(contextDir, "subdir", "file")`.

pkg/util/util_test.go
```diff
@@ -16,17 +26,88 @@ func TestDiscoverContainerfile(t *testing.T) {
...
+	t.Parallel()
+	tmpDir := t.TempDir()
+
+	contextDir := filepath.Join(tmpDir, "context")
+	require.NoError(t, os.Mkdir(contextDir, 0o755))
+	require.NoError(t, os.Symlink("../nonexistent", filepath.Join(contextDir, "Containerfile")))
+
+	_, err := DiscoverContainerfile(contextDir)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "cannot find Containerfile or Dockerfile")
```
nalind: This can be changed to an `assert.ErrorIs()` call.

Honny1:
PTAL @nalind

pkg/util/util.go
simonbrauner: Looks good now

simonbrauner:
My comments were addressed

pkg/util/util.go
```diff
@@ -41,42 +43,76 @@ func MirrorToTempFileIfPathIsDescriptor(file string) (string, bool) {
...
 
-		// The file exists, now verify the correct mode
-		if mode := file.Mode(); mode.IsRegular() {
-			foundCtrFile = ctrfile
-		} else {
-			return "", fmt.Errorf("assumed Containerfile %q is not a file", ctrfile)
+	case target.Mode()&os.ModeSymlink != 0:
+		if resolved, ok := isRegularFileInContext(filepath.Dir(path), path); ok {
+			return resolved, nil
 		}
```
nalind, line 88: I'm not sure what we gain by ensuring that the last component in a pathname is not a symbolic link to somewhere outside of its parent directory when that directory can already be outside of any known build context directory.
I would be fine with having this function reject outright context locations that are neither directories nor symlinks to directories.

simonbrauner: > I would be fine with having this function reject outright context locations that are neither directories nor symlinks to directories.

But that would be a breaking change, no? Currently we do support file as an argument in the place of `[context]`.

`buildah build -t demo test-repository6/Dockerfile` seems to do the same thing as `buildah build -t demo test-repository6`.

But it does not seem to work together with `-f`:

```shell
$ buildah build -t demo -f test-repository6/Dockerfile test-repository6/Dockerfile
Error: mounting an overlay over build context directory: creating overlay scaffolding for build context directory: mount overlay:/var/tmp/buildah-context-231279313/overlay/388852530/merge, data: lowerdir=/home/sbrauner/Desktop/cve-work/buildah-build/test-repository6/Dockerfile,upperdir=/var/tmp/buildah-context-231279313/overlay/388852530/upper,workdir=/var/tmp/buildah-context-231279313/overlay/388852530/work,context="system_u:object_r:container_file_t:s0:c231,c966",userxattr: invalid argument
```

Honny1:

TomSweeneyRedHat:
LGTM
but I'd like a final head nod from @nalind

simonbrauner:
LGTM

Honny1:
PTAL @mtrmac

PR #6878, opened 2026-05-27
[release-1.43] Backport a number of commits for a security fix

TomSweeneyRedHat:
For a recent security fix, backport the following commits into the release-1.43 branch

- https://github.com/containers/buildah/commit/fc2003bb2efeb3ed7d5c7e1e88d04e78f370a944
- https://github.com/containers/buildah/commit/55c4f2f09eefae3d1a9b7b792790a0deadf2a249
- https://github.com/containers/buildah/commit/4a34c646f454c241ac377f299cc752b3299b4672
- https://github.com/containers/buildah/commit/8c3eb57bd67191da3e539b11ded23b8549cf7abb
- https://github.com/containers/buildah/commit/015c1c73b14a42455c35cf7c0f4e43a3985e7007
 
@Honny1 and @mtrmac PTAL

I had some "fun" backporting this.  If this look OK, I'll add a commit to bump the version too.

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #6887, opened 2026-06-02
[release-1.43] fix packit

lsm5:
#### What type of PR is this?

 /kind other

Fix CI

#### How to verify it

Successful CI

None

None

```release-note
None
```


lsm5:
@TomSweeneyRedHat @nalind PTAL . this should fix testing-farm job issues seen in https://github.com/podman-container-tools/buildah/pull/6878

nalind:
I think this needs a cherry-pick from https://github.com/podman-container-tools/buildah/pull/6849 if we want the testing farm test to pass.

lsm5:
@nalind removed the rawhide job as it won't be shipped there so I think we're good to go here.

TomSweeneyRedHat:
Looks like the farm tests passed.
LGTM
@nalind, in your court

nalind:
We luck out that the cloud image we're using to run these tests happens to have the packages installed that provide the binaries the tests need when SELinux is enabled (that, or they run without SELinux enabled), so the tests pass.  I'd still have preferred the .spec file updates, since the tests tend to be run in all kinds of environments, and we get bug reports if they fail for any reason, but it's not a blocker here.

Luap99:
The "Bump to Buildah v1.43.2" commit is part of this? Is it intentional that we merge this as part of this PR?

Also we will need to update the merge protections to not need Total Success if we only use packit here.

nalind:
What would be required to migrate this branch to use the new CI setup, as I hear podman's doing for its 5.8 branch?

Luap99:
> What would be required to migrate this branch to use the new CI setup, as I hear podman's doing for its 5.8 branch?


cherry-pick the the CI migration PR(s) from main, fix conflicts, push and solve CI errors in a loop.
It can be done, I have not looked at the buildah changes to much but I guess it should be much simpler than podman and there it took me a day basically.

nalind:
> The "Bump to Buildah v1.43.2" commit is part of this? Is it intentional that we merge this as part of this PR?

This PR is currently _exactly_ #6878 with one commit added after it.  We could merge that one, then this one, or add the commit elsewhere in #6878, or extract it into a different PR and merge it as a standalone item.

TomSweeneyRedHat:
Given @nalind 's OK, I'm going to merge this and will tie off with @lsm5 to clean up the testing later.
/lgtm

Luap99:
auto merge cannot work, the merge protection settings need to be updated, let me do this and then merge

Luap99:
I disabled the merge protection for the release- branches, if we have an agreement on what kind of CI we expect here we should add it back.

PR #6908, opened 2026-06-15
parse: honor containers.conf seccomp_profile

### LINKED ISSUE #6498: User containers.conf seccomp_profile = "unconfined" has no effect
It looks like seccomp_profile from containers.conf wasn't wired in to Run().

### Discussed in https://github.com/containers/buildah/discussions/6490

<div type='discussions-op-text'>

<sup>Originally posted by **XeCycle** November  9, 2025</sup>
I'm on Arch Linux host, and running Ubuntu jammy (and later) images has problems with the shipped seccomp profile.  It works with `buildah from --security-opt seccomp=unconfined ubuntu:noble`; setting `containers.seccomp_profile = "unconfined"` in `~/.config/containers/containers.conf` works for podman, but not buildah.  Which setting can I use with buildah so that I don't need to add those args on every call?</div>

jiwahn:
#### What type of PR is this?

/kind bug

Buildah was already loading `containers.conf`, but it was not applying the `SeccompProfile` field to `CommonBuildOptions`. This patch wires that field into the existing security opt parsing path. If the user explicitly provides a CLI seccomp option, CLI values take precedence.

#### How to verify it

`go test ./pkg/parse`

I added a config parsing test that verifies:

- seccomp_profile = "unconfined" from containers.conf is applied.
- An explicit CLI --security-opt seccomp=... value takes precedence over the configured profile.
- When no seccomp profile is configured, the existing default behavior is preserved.

Fixes: #6498

```release-note
Buildah now honors containers.conf's seccomp_profile setting.
```

### DIFF
--- a/pkg/parse/parse.go
+++ b/pkg/parse/parse.go
@@ -235,6 +235,7 @@ func CommonBuildOptionsFromFlagSet(flags *pflag.FlagSet, findFlagFunc func(name 
 	if err != nil {
 		return nil, fmt.Errorf("failed to get container config: %w", err)
 	}
+	securityOpts = addSeccompFromConfig(securityOpts, defConfig.Containers.SeccompProfile)
 	if defConfig.Containers.EnableLabeledUsers {
 		defSecurityOpts, err := currentLabelOpts()
 		if err != nil {
@@ -247,6 +248,21 @@ func CommonBuildOptionsFromFlagSet(flags *pflag.FlagSet, findFlagFunc func(name 
 		return nil, err
 	}
 	return commonOpts, nil
+}
+
+// addSeccompFromConfig appends seccompProfile as a seccomp security option
+// when seccompProfile is non-default and securityOpts does not already contain
+// a seccomp option.
+func addSeccompFromConfig(securityOpts []string, seccompProfile string) []string {
+	if seccompProfile == "" || seccompProfile == SeccompDefaultPath {
+		return securityOpts
+	}
+	for _, opt := range securityOpts {
+		if strings.HasPrefix(opt, "seccomp=") {
+			return securityOpts
+		}
+	}
+	return append(securityOpts, "seccomp="+seccompProfile)
 }
 
 // GetAdditionalBuildContext consumes a raw string and returns a parsed
--- a/pkg/parse/parse_test.go
+++ b/pkg/parse/parse_test.go
@@ -3,6 +3,7 @@ package parse //nolint:revive,nolintlint
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"runtime"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


pkg/parse/parse_test.go
```diff
@@ -29,6 +31,76 @@ func TestCommonBuildOptionsFromFlagSet(t *testing.T) {
 	assert.Equal(t, cbo.Memory, int64(2147483648))
 }
 
+func TestCommonBuildOptionsSeccompFromConfig(t *testing.T) {
+	configPath := filepath.Join(t.TempDir(), "containers.conf")
+	originalConfig, configWasSet := os.LookupEnv("CONTAINERS_CONF")
+	require.NoError(t, os.Setenv("CONTAINERS_CONF", configPath))
```
nalind: Calling `t.Setenv("CONTAINERS_CONF", configPath)` would probably narrow the cleanup block down to the `config.Reload()` call.

nalind (commented):
Generally looks good, one nit, nothing dramatic.

pkg/parse/parse_test.go
jiwahn: Thanks @nalind, good to know! Updated.

nalind (approved):
LGTM, thanks!
@podman-container-tools/buildah-maintainers PTAL

pkg/parse/parse.go
```diff
@@ -235,6 +235,7 @@ func CommonBuildOptionsFromFlagSet(flags *pflag.FlagSet, findFlagFunc func(name
 	if err != nil {
 		return nil, fmt.Errorf("failed to get container config: %w", err)
 	}
+	securityOpts = addSeccompFromConfig(securityOpts, defConfig.Containers.SeccompProfile)
```
rhatdan, line 238: Not sure why this has a seccompProfile option, since it is not used anywhere, and if the securityOpts containers a seccomp= then this parameter is ignored?

jiwahn: Hey @rhatdan, thanks for the review. I wanted to preserve CLI precedence here, so the configured profile is ignored when a CLI `seccomp=` option is already present.

I can inline this logic, or split it into a helper that only checks for the existence of `seccomp=`, if that’s clearer. WDYT?

rhatdan: I'll leave this up to Nalin, I guess add a comment explaining what is going on.

nalind: Would adding a godoc comment to `addSeccompFromConfig()` which describes its function as checking for a "seccomp" option in the passed-in slice, and if there isn't one, appending one which names the passed-in `seccompProfile` value, be helpful to future maintainers?

jiwahn: Alright, I added a godoc comment

Honny1 (commented):
LGTM

nalind (approved):
LGTM, thanks!

PR #6911, opened 2026-06-16
CI image update and various CI improvements

Luap99:
see commits


.github/workflows/lima.yml
```diff
@@ -34,10 +34,6 @@ jobs:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
         with:
           persist-credentials: false
-          fetch-depth: 50
-
-      - name: Fetch base ref for merge-base
-        run: git fetch --depth=50 origin main:refs/remotes/origin/main || true
```
nalind, line 40: My guess would be that this was for the benefit of the `tests/validate/pr-should-include-tests` script.

nalind (approved):
LGTM

.github/workflows/lima.yml
Luap99: yeah except the script is not updated to work with github actions, it still uses cirrus so it will not work.

I will further rework the script, i.e. pull the one from podman and update the action to set the right vars.

Luap99:
@nalind I pushed some more changes to fix the validation flow as well.
It is closer to podman now so hopefully less maintenance work for us.

Luap99:
cc @timcoding1988

hack/ci/ci.sh
```diff
@@ -6,7 +6,7 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" && pwd )
 
 source "$SCRIPT_DIR/lib.sh"
 
-AUTOMATION_RELEASE="${AUTOMATION_RELEASE:-20260520t200858z}"
+AUTOMATION_RELEASE="${AUTOMATION_RELEASE:-20260616t073924z}" # TODO should be renovate managed
```
TomSweeneyRedHat, line 9: As noted by the comment here and in Podman, it would be great to have a way to auto update this so we don't neglect to.  Not for this PR however.

timcoding1988:
LGTM.

TomSweeneyRedHat:
LGTM
assuming happy tests

tests/bud.bats
```diff
@@ -7878,7 +7878,7 @@ _EOF
   run_buildah bud $WITH_POLICY_JSON --network pasta $BUDFILES/network
   assert "$output" =~ "$ip" "ip addr shows default subnet"
 
-  # check some entwork options, it accepts raw pasta(1) areguments
+  # check some entwork options, it accepts raw pasta(1) arguments
```
nalind, line 7881: Typo: "entwork"

nalind (approved):
LGTM with a nit.

tests/bud.bats
Luap99: ack, lets get the PR in so others benefit as well. Seems like codespell did not pick this up

I can open a follow up for this

Luap99: https://github.com/podman-container-tools/buildah/pull/6912 also has more CI fixes

Issue #6903, opened 2026-06-09
Buildah 1.44.0 fails to pull images with opaque whiteouts when running in a privileged user namespace


chmeliik:
### Issue Description

We run buildah under an `unshare --map-root-user --map-auto --mount` namespace. In version 1.44.0, this causes buildah to fail when pulling images whose layers contain opaque whiteouts:
```text
Error: creating build container: unable to copy from source docker://quay.io/devfile/python@sha256:54924a2ee4a2ef17028ae076ce38e59b3f4054353a5c9f9318dfaee60377532c: writing blob: adding layer with blob "sha256:778656c04542093db6d3b6e07bffbcf6ec4b24709276be7cdf177fcb3666663a"/""/"sha256:d326469892d974408d96f1e02d64dce10d20f88613688af11e99e3e22523beeb": unpacking failed (error: exit status 1; output: time="2026-06-09T19:55:43Z" level=warning msg="Failed to read CAP_SYS_ADMIN presence for the current process"
time="2026-06-09T19:55:43Z" level=warning msg="Failed to read current user namespace mappings"
operation not permitted)
```

The same thing occurs when running buildah in a podman `--privileged` container

According to my debugging, the root cause is a bug in `go.podman.io/storage` that got exposed by ce7811fde18de3770905b0331f3b873d8e8b8f93:
- The [`storage-untar` re-exec handler](https://github.com/podman-container-tools/container-libs/blob/1cd91c5bfe88613e511cb1da46355612b8309837/storage/pkg/chrootarchive/archive_unix.go#L49) calls `chroot()` first before calling `unshare.IsRootless()` to decide what xattr name to use for the opaque whiteout (`user.overlay.opaque` vs `trusted.overlay.opaque`)
- `chroot()` does something that causes `IsRootless()` to fail both of the secondary checks (`CAP_SYS_ADMIN` presence, full user mappings), so it returns `false`
- Attempts to set `trusted.overlay.opaque` even though we're not real root => operation not permitted

Prior to v1.44.0, it worked fine, because something in `init()` called `unshare.IsRootless()` before the chroot happened, so the correct value was already cached by then. But now it's not called until after the chroot.

### Steps to reproduce the issue

Steps to reproduce the issue
1. Create a Containerfile to install buildah and dependencies:

<details><summary>expand</summary>

```Dockerfile
FROM registry.access.redhat.com/ubi10/go-toolset:1.26.3 AS go-build

[...]

PR #6923, opened 2026-06-19
[release-1.44] fix opaque whiteout pull error regression

Luap99:
#### What type of PR is this?

 /kind bug

Buildah 1.44 has a new regression that was caused by me removing a bunch of init() code out of buildah, see commit ce7811fde1.

The bug is actual a storage library issue but because a fix there needs a full vendor dance to get released instead we do this alternative fix for just the buildah 1.44 branch.

Fixes: #6903

The proper fix for storage is here https://github.com/podman-container-tools/container-libs/pull/908
As such I have not submitted this for for buildah@main as we will consume the storage fix on the next upgrade there and have no need for this then.

```release-note
Fixed a regression that could make image pulls fail in a user namespace when using opaque whiteouts
```

### DIFF
--- a/cmd/buildah/main.go
+++ b/cmd/buildah/main.go
@@ -243,6 +243,16 @@ func after(cmd *cobra.Command) error {
 }
 
 func main() {
+	// Not a NOP, see https://github.com/podman-container-tools/buildah/issues/6903.
+	// This is a work around for a bug in the storage library as it does not properly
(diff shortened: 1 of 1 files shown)


Luap99:
Added no new tests as we do not have a rootless test run here setup right now. There is a simple reproducer on the issue which I have manually verified.

nalind (approved):
LGTM, thanks!

lsm5 (approved):
LGTM

PR #6927, opened 2026-06-24
Fix .dockerignore wildcard negation directory descent

### LINKED ISSUE #6615: dockerignore prevents build from working
### Issue Description

This is the [dockerignore](https://github.com/valkey-io/valkey-operator/blob/main/.dockerignore) posing issues for use with buildah. If I remove this file, buildah works, but if the file is present, _something_ doesn't work correctly and the files are not copied into the building container. Relevant [Dockerfile](https://github.com/valkey-io/valkey-operator/blob/main/Dockerfile)

gitlab runner output:

```
build1:
  stage: build
  image: quay.io/containers/buildah:latest
  script:
    - echo "=== stage:build script ==="
    - ls -la
    - buildah build -f Dockerfile -t "$IMAGE_NAME"
    - buildah images
    - buildah push "$IMAGE_NAME"




$ echo "=== stage:build script ==="
=== stage:build script ===
$ ls -la
total 116
drwxrwxrwx. 12 root root 16384 Dec 25 03:36 .
drwxrwxrwx.  4 root root    56 Dec 25 03:36 ..
drwxrwxrwx.  2 root root    54 Dec 25 03:36 .devcontainer
-rw-rw-rw-.  1 root root   383 Dec 25 03:36 .dockerignore
drwxrwxrwx.  6 root root   113 Dec 25 03:36 .git
drwxrwxrwx.  5 root root    72 Dec 25 03:36 .github
-rw-rw-rw-.  1 root root   460 Dec 25 03:36 .gitignore
-rw-rw-rw-.  1 root root   832 Dec 25 03:36 .gitlab-ci.yml
-rw-rw-rw-.  1 root root   930 Dec 25 03:36 .golangci.yml
-rw-rw-rw-.  1 root root  6290 Dec 25 03:36 CONTRIBUTING.md
-rw-rw-rw-.  1 root root  1232 Dec 25 03:36 Dockerfile
-rw-rw-rw-.  1 root root 11357 Dec 25 03:36 LICENSE
-rw-rw-rw-.  1 root root 11054 Dec 25 03:36 Makefile
-rw-rw-rw-.  1 root root   663 Dec 25 03:36 PROJECT
-rw-rw-rw-.  1 root root  5930 Dec 25 03:36 README.md
-rw-rw-rw-.  1 root root  3364 Dec 25 03:36 SUPPORT.md
drwxrwxrwx.  3 root root    22 Dec 25 03:36 api
drwxrwxrwx.  2 root root    21 Dec 25 03:36 cmd
drwxrwxrwx.  9 root root   114 Dec 25 03:36 config
drwxrwxrwx.  2 root root    32 Dec 25 03:36 docs
-rw-rw-rw-.  1 root root  4631 Dec 25 03:36 go.mod
-rw-rw-rw-.  1 root root 23636 Dec 25 03:36 go.sum
drwxrwxrwx.  2 root root    32 Dec 25 03:36 hack
drwxrwxrwx.  4 root root    38 Dec 25 03:36 internal
drwxrwxrwx.  4 root root    30 Dec 25 03:36 test
$ buildah build -f Dockerfile -t "$IMAGE_NAME"
[1/2] STEP 1/9: FROM golang:1.24 AS builder
Resolved "golang" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull docker.io/library/golang:1.24...
Copying blob sha256:76da54420e1edd91a760c3248b46ee9cade4a7969f9da7bcff9b40eee7d95788
Copying blob sha256:2981f7e8980b9f4b6605026e1c5f99b4971ebba15f626e46904554de09f324f4
Copying blob sha256:58f2d358b447d091790c5ef0943550bbcf57bac46c4b8bfcfc3e6dacf4cb7969
Copying blob sha256:2f9dc3e96e7f673873071f3abef7cbb5fb77c69508e799dbe503bf1e316168d8
Copying blob sha256:b22766554d6bfa95c7325b00ee002f2705a7b8605908c3eb43dbe729c412422c
Copying blob sha256:94cc3e02e353c0efcfdaf6be736e5d50dbde521e32a7ec4b0a8f8ce54558c2c2
Copying blob sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1
Copying config sha256:ec07d1d940ae89674b89151998a5e760ace7e5a376f1536c98b8b075ca51d2cf
[1/2] STEP 2/9: ARG TARGETOS
[1/2] STEP 3/9: ARG TARGETARCH
[1/2] STEP 4/9: WORKDIR /workspace
[1/2] STEP 5/9: COPY go.mod go.mod
[1/2] STEP 6/9: COPY go.sum go.sum
[1/2] STEP 7/9: RUN go mod download
[1/2] STEP 8/9: COPY . .
[1/2] STEP 9/9: RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager cmd/main.go
package cmd/main.go is not in std (/usr/local/go/src/cmd/main.go)
subprocess exited with status 1
subprocess exited with status 1
Error: building at STEP "RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager cmd/main.go": exit status 1
```

Steps to reproduce the issue
1. Attempt to build container with buildah which uses a .dockerignore file
2. Observe that build fails
3. Remove .dockerignore and repeat attempt
4. Observe that build succeeds

Builds with buildah containing a .dockerignore file will fail for some reason

Builds to work correctly

### buildah version output

```yaml
`  image: quay.io/containers/buildah:latest`
```

### buildah info output

```yaml
gitlab runner in K8S
```

### Provide your storage.conf

```toml
File is not used
```

### Upstream Latest Release

Yes

### Additional information

### LINKED ISSUE #7025: Replace internal `ShouldDescendExcludedDir` with upstream `c/storage` implementation
PR [container-libs#1006](https://github.com/podman-container-tools/container-libs/pull/1006) adds `ShouldDescendExcludedDir` to `c/storage/pkg/fileutils.PatternMatcher` and fixes `archive.TarWithOptions` to use it.

Once that merges and buildah vendors the updated `c/storage`, we should:
- Remove `internal/excludes/` package and use `PatternMatcher.ShouldDescendExcludedDir()` directly
- Update `copier/copier.go` and `add.go` to call the upstream method

Depends on: [container-libs#1006](https://github.com/podman-container-tools/container-libs/pull/1006)

Honny1:
Buildah skipped excluded directories when negation patterns contained wildcards (e.g. `!**/*.go`), because the descent check only matched literal prefixes. Use `PatternMatcher.ShouldDescendExcludedDir` from c/storage (container-libs#1006) which extracts the literal prefix before the first wildcard and descends when the directory is at or under it.

Same bug as Docker's classic builder (moby/moby#30018, moby/moby#45608).

Fixes: https://github.com/podman-container-tools/buildah/issues/6615
Resolve: https://github.com/podman-container-tools/buildah/issues/7025

Depends on: https://github.com/podman-container-tools/container-libs/pull/1006

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
Fixed `.dockerignore`/`.containerignore` handling of wildcard negation patterns (e.g. `!**/*.go`, `!*/*.go`). Previously, files matching these patterns inside excluded directories were incorrectly omitted from the build context.
```


Honny1:

copier/copier.go
```diff
@@ -1094,6 +1094,44 @@ func pathIsExcluded(root, path string, pm *fileutils.PatternMatcher) (string, bo
...
+func IncludeDirectoryAnyway(dirPath string, pm *fileutils.PatternMatcher) bool {
+	if pm == nil || !pm.Exclusions() {
+		return false
+	}
+	dir := filepath.ToSlash(strings.TrimPrefix(dirPath, string(os.PathSeparator)))
+	for _, pattern := range pm.Patterns() {
+		if !pattern.Exclusion() {
+			continue
+		}
+		slashPattern := filepath.ToSlash(strings.TrimPrefix(pattern.String(), string(os.PathSeparator)))
```
nalind: Should this be Trim() instead of TrimPrefix()?  The parser in imagebuilder will strip them, but we don't always know that that's where this list came from, and the check just below this would be tripped up by that.

add.go
```diff
@@ -708,7 +689,7 @@ func (b *Builder) Add(destination string, extract bool, options AddAndCopyOption
 			}
 			// Check for dockerignore-style exclusion of this item.
 			if rel != "." {
-				excluded, err := pm.Matches(filepath.ToSlash(rel)) //nolint:staticcheck
+				excluded, err := pm.IsMatch(filepath.ToSlash(rel))
```
simonbrauner: Probably out of scope, but the deprecated call could be fixed there as well? https://github.com/podman-container-tools/buildah/blob/main/copier/copier.go#L1087

simonbrauner (commented):
LGTM

add.go
```diff
@@ -253,27 +253,8 @@ func getURL(src string, chown *idtools.IDPair, mountpoint, renameTarget string,
...
-		if !pattern.Exclusion() {
-			continue
-		}
-		spec := strings.TrimPrefix(pattern.String(), string(os.PathSeparator))
-		if strings.HasPrefix(spec, prefix) {
-			return true
-		}
-	}
-	return false
+	return copier.IncludeDirectoryAnyway(path, pm)
```
simek-m: Is there a reason to have the private function that only delegates to the public (commented) one?

tests/conformance/conformance_test.go
```diff
@@ -3170,9 +3179,10 @@ var internalTestCases = []testCase{
 			}
 			return nil
 		},
-		fsSkip:               []string{"(dir):subdir:mtime"},
-		failOnExtraFSContent: true,
-		compatScratchConfig:  types.OptionalBoolTrue,
+		// #6615: see dockerignore-exclude-kind-of-deep-subdir-dot
+		withoutDocker:       true,
+		fsSkip:              []string{"(dir):subdir:mtime"},
+		compatScratchConfig: types.OptionalBoolFalse,
```
simek-m: Was the changed `compatScratchConfig` from `types.OptionalBoolTrue` to `OptionalBoolFalse` because it wasn't needed?

Also, there are removals of `failOnExtraFSContent`. Is that correct?

copier/copier.go
```diff
@@ -1094,6 +1094,44 @@ func pathIsExcluded(root, path string, pm *fileutils.PatternMatcher) (string, bo
 	return rel, false, nil
 }
 
+// IncludeDirectoryAnyway checks whether an excluded directory should still be
+// descended into because a negation pattern in pm might match files under it.
+// It handles literal prefix matches (e.g. !cmd/main.go for dir "cmd") and
+// wildcard negations (e.g. !**/*.go, !*/*.go, !src/*/config.yaml).
+func IncludeDirectoryAnyway(dirPath string, pm *fileutils.PatternMatcher) bool {
```
simek-m: Something like `ShouldDescendExcludedDir` might be a more descriptive name.

simek-m: Also, I think that the `and wildcard negations (e.g. !**/*.go, !*/*.go, !src/*/config.yaml)` part is not completely correct. `!src/*/config.yaml` is IMHO handled by the literal prefix match.

The comment might mention that the wildcard-aware check may intentionally overmatch for wildcards in the middle (it's below).

copier/copier_test.go
```diff
@@ -3148,3 +3149,115 @@ func testChmod(t *testing.T) {
...
+			want:     true,
+		},
+		{
+			name:     "deep nested with double star prefix",
+			path:     "src/internal/pkg",
+			patterns: []string{"**", "!src/**/*.go"},
+			want:     true,
+		},
+		{
+			name:     "non-exclusion patterns are ignored",
```
simek-m: TC like?:

```go
		{
			name:     "dir prefix match is ignored",
			path:     "cmds",
			patterns: []string{"!cmd/main.go"},
			want:     false,
		},
```

Honny1:
The failed test is an unrelated flake.

simek-m (approved):
All comments reflected, LGTM.

copier/copier.go
```diff
@@ -1084,7 +1084,7 @@ func pathIsExcluded(root, path string, pm *fileutils.PatternMatcher) (string, bo
 	// Matches uses filepath.FromSlash() to convert candidates before
 	// checking if they match the patterns it's been given, implying that
 	// it expects Unix-style paths.
```
simonbrauner: Just a detail, but the comment uses the old function name `Matches`. It may be misleading now, and the content may no longer be relevant.

inknos: Agree, then LGTM

copier/copier.go
```diff
@@ -1094,6 +1094,47 @@ func pathIsExcluded(root, path string, pm *fileutils.PatternMatcher) (string, bo
...
+		if strings.HasPrefix(slashPattern, dir+"/") {
+			return true
+		}
+
+		// Wildcard-aware check: extract the literal prefix before
+		// the first wildcard character (*, ?, [), for example: !cmd/**/*.go matches dir "cmd"
+		// if the directory is at or under that literal prefix, a file beneath this
+		// directory could match the negation, so keep descending.
+		if firstWild := strings.IndexAny(slashPattern, "*?["); firstWild >= 0 {
+			literalPrefix := strings.TrimRight(slashPattern[:firstWild], "/")
```
mtrmac: What about `!cmd/image*`? `TrimRight` will not find the `/`.

copier/copier.go
```diff
@@ -1094,6 +1094,47 @@ func pathIsExcluded(root, path string, pm *fileutils.PatternMatcher) (string, bo
...
 }
 
+// ShouldDescendExcludedDir checks whether an excluded directory should still be
+// descended into because a negation pattern in pm might match files under it.
+// It handles literal prefix matches (e.g. !cmd/main.go for dir "cmd") and
+// wildcard negations (e.g. !**/*.go, !*/*.go). The wildcard check extracts
+// the literal prefix before the first wildcard and may intentionally
+// overmatch (descend into directories that won't ultimately contain matches),
+// which is safe because actual file-level matching happens later.
+func ShouldDescendExcludedDir(dirPath string, pm *fileutils.PatternMatcher) bool {
```
mtrmac: This (or some variant of this) should probably be in `c/storage/pkg/fileutils` instead: c/storage/pkg/archive.tarWithOptionsTo` contains a similar exclusion check with the same problem.

mtrmac (commented):
(Just a pretty brief look, I didn’t read the tests at all.)

copier/copier.go
mtrmac: (If we did end up doing this Buildah-only, I think having this in something like `internal/exclusions` would be better, not to commit to it as a stable API — but that’s a Buildah maintainer decision.)

copier/copier.go
simek-m: @mtrmac Oh, I thought it's handled with `strings.HasPrefix` before, but no - only for `cmd` with `!cmd/image*`, but not `cmd/images` with `!cmd/image*` or `!cmd/image*/main.go` with `cmd/images/main.go`.

copier/copier.go
nalind: Agreed, we definitely don't want to add this as an exported API that we have to keep working, even if/after how exclusion matching rules change.

copier/copier.go
```diff
@@ -1081,10 +1081,10 @@ func pathIsExcluded(root, path string, pm *fileutils.PatternMatcher) (string, bo
 		// special case
 		return rel, false, nil
 	}
-	// Matches uses filepath.FromSlash() to convert candidates before
+	// IsMatch() uses filepath.FromSlash() to convert candidates before
 	// checking if they match the patterns it's been given, implying that
 	// it expects Unix-style paths.
-	matches, err := pm.Matches(filepath.ToSlash(rel)) //nolint:staticcheck
+	matches, err := pm.IsMatch(filepath.ToSlash(rel))
```
nalind: `IsMatch()` does more work to return the same result as `Matches()`, it's not worth switching which one we use.

tests/bud.bats
```diff
@@ -605,7 +605,7 @@ _EOF
 
   run_buildah 1 run myctr ls -l sub2.txt
 
-  run_buildah 1 run myctr ls -l subdir/
+  run_buildah run myctr ls -l subdir/
```
nalind: Is this directory expected to have any contents?

nalind (commented):
I'm not very enthusiastic about updating conformance tests which test "ignores" behavior so that they don't use a baseline.

copier/copier.go
Honny1: I have moved that into `internal/excludes`. Should I create a PR with the fix in `c/storage` as a follow-up?

Honny1:
> I'm not very enthusiastic about updating conformance tests which test "ignores" behavior so that they don't use a baseline.

I am also not a big fan of this, but the root cause is that go-dockerclient sends the build context through moby/go-archive's TarWithOptions, which has the same bug (moby/moby#30018, moby/moby#45608). Docker BuildKit handles it correctly, but since the framework filters client-side before the context reaches BuildKit, we can't get a valid baseline.

tests/bud.bats
Honny1: Yes

tests/bud.bats
```diff
@@ -605,7 +605,9 @@ _EOF
 
   run_buildah 1 run myctr ls -l sub2.txt
 
-  run_buildah 1 run myctr ls -l subdir/
+  # !*/sub1* un-excludes subdir/sub1.txt
+  run_buildah run myctr ls -l subdir/
+  run_buildah run myctr ls -l subdir/sub1.txt
```
nalind: This should probably be making sure that nothing else was copied to this directory.

internal/excludes/excludes_test.go
```diff
@@ -0,0 +1,139 @@
...
+			patterns: []string{"**", "!cmd/image*"},
+			want:     true,
+		},
+		{
+			name:     "non-exclusion patterns are ignored",
+			path:     "cmd",
+			patterns: []string{"cmd/**/*.go"},
+			want:     false,
+		},
+	}
```
nalind: For the sake of completeness, this should probably include test cases that use some of the other wildcard characters.

Honny1:
PTAL @nalind

tests/bud.bats
nalind: Something like this
```bash
run_buildah run myctr find subdir -mindepth 1 -print
assert "$output" = "subdir/sub1.txt"
```
would make sure that nothing else was copied there.

tests/conformance/conformance_test.go
```diff
@@ -3084,6 +3084,18 @@ var internalTestCases = []testCase{
...
+		// Tested withoutDocker because go-dockerclient sends the build
+		// context through moby/go-archive's TarWithOptions, which has
+		// the same wildcard directory-descent bug: it uses a literal
+		// strings.HasPrefix to decide whether to descend into excluded
+		// directories (moby/moby#30018, moby/moby#45608,
+		// https://github.com/moby/go-archive/blob/v0.2.0/archive.go#L766).
+		// Docker BuildKit handles it correctly, but we cannot compare
+		// through this framework.
+		// TODO: re-enable Docker comparison when moby/go-archive fixes
+		// wildcard negation handling in TarWithOptions.
```
nalind: We supply the `InputStream` field in the options, not the `ContextDir`, so we do have control over which parts of the build context we transfer.

Honny1: Ah, thanks, I didn't realize that.

Honny1:
I tried sending an unfiltered build context to Docker and letting BuildKit handle `.dockerignore` filtering server-side, but BuildKit's filtering does not do parent-prefix matching. For example `**/subdir-e` only excludes the directory entry, not files underneath it, so it excludes fewer files than buildah ([moby/moby#45608](https://github.com/moby/moby/issues/45608), [moby/moby#42788](https://github.com/moby/moby/issues/42788)). Since neither client-side ([moby/moby#30018](https://github.com/moby/moby/issues/30018)) nor server-side filtering produces a valid reference, the affected tests now skip the Docker comparison with `withoutDocker: true`.

Honny1:
PTAL @nalind

copier/copier.go
Honny1: PR: https://github.com/podman-container-tools/container-libs/pull/1006

tests/bud.bats
```diff
@@ -1014,8 +1016,8 @@ this is the output of test12"
 
 @test "bud with .containerignore" {
   _prefetch alpine busybox
-  run_buildah 125 build -t testbud $WITH_POLICY_JSON -f $BUDFILES/containerignore/Dockerfile $BUDFILES/containerignore
-  expect_output --substring 'building.*"COPY subdir \./".*no such file or directory'
+  # https://github.com/containers/buildah/issues/6615
+  run_buildah build -t testbud $WITH_POLICY_JSON -f $BUDFILES/containerignore/Dockerfile $BUDFILES/containerignore
```
nalind, line 1025: Are there any expectations on the contents of this image that this build command now successfully produces that need to be checked?  The Dockerfile being used directly after this is the same as this one, but missing the COPY instruction that previously caused this build to fail.

tests/bud.bats
```diff
@@ -1101,8 +1105,8 @@ symlink(subdir)"
 
 @test "bud with .dockerignore #6" {
   _prefetch alpine busybox
-  run_buildah 125 build -t testbud $WITH_POLICY_JSON -f $BUDFILES/dockerignore6/Dockerfile $BUDFILES/dockerignore6
-  expect_output --substring 'building.*"COPY subdir \./".*no such file or directory'
+  # https://github.com/containers/buildah/issues/6615
+  run_buildah build -t testbud $WITH_POLICY_JSON -f $BUDFILES/dockerignore6/Dockerfile $BUDFILES/dockerignore6
```
nalind, line 1118: Are there any expectations on the contents of this image that this build command now successfully produces that need to be checked?  The Dockerfile being used directly after this is the same as this one, but missing the COPY instruction that previously caused this build to fail.

add.go
nalind: I should note that `IsMatch()` does more work to return the same information that `Matches()` does.

nalind (approved):
A couple of the integration tests could be taking a closer look at what makes it into an image that no longer fails to build.  I don't like that we're losing baselines for some of the conformance tests.

tests/bud.bats
Honny1: I have added an image content check.

tests/bud.bats
Honny1: I have added an image content check.

add.go
Honny1: I added a comment to inform the next person who tries this.

Honny1:
> I don't like that we're losing baselines for some of the conformance tests.

I think I missed something. I fixed the conformance tests. (I need to do cleanup and push changes)

Honny1:

I fixed the conformance test failures with minimal changes to the skip rules. That said, the tar file supplied to Docker as the build context is pre-filtered, so these tests only partially expose real Docker vs. buildah behavior. 
https://github.com/podman-container-tools/buildah/blob/182f7465e475b3b7e5983476ae1833fcd77b142d/tests/conformance/conformance_test.go#L728

TomSweeneyRedHat:
LGTM
@nalind?

simonbrauner:
LGTM

Honny1:
I have rebased on upstream main.

simonbrauner:
There are conflicts, can you take a look at them @Honny1 ?

PR #6933, opened 2026-06-26
Add a workaround for using `mount` without privileges, re-enable rootless tests in CI

nalind:
#### What type of PR is this?

/kind feature

* Add a `buildah mount-sshfs` workaround for not being able to `buildah mount` when not root.
* Rework integration tests to use `buildah mount-sshfs` when run by a UID other than 0, and working around the workaround's limitations (chiefly: avoid `buildah run/add/copy` while the filesystem is mounted).
* Re-enable rootless tests in CI.

#### How to verify it

New and updated integration tests!

This is mainly about re-enabling tests in CI.  We weren't really testing with overlay in rootless scenarios before.

The new commands are marked as hidden.
This doesn't add new library APIs.

```release-note
Adds undocumented `buildah serve-sftp` and `buildah sshfs-mount` commands.  Don't tell anyone.
```


Luap99 (commented):
Thanks, I am a bit conflicted about this.

On the one hand I see how this helps us with testing but on the other side do we really want to ship and fully support these commands for general use?

I mean 900 lines is not that bad but it also is not exactly straight forward code and then well we take on a new sftp dep which is like 10k lines as well. That feels a bit much for just testing, though I guess the reason why documents them is because they can also be useful to some users so I can see the point.

Since you are the main maintainer here and likely already considered all this when you write the code I guess I am cool with it.

CI wise LGTM, I did not review the 900 lines of sftp.go so far though

TomSweeneyRedHat:
LGTM
@ddarrah WDYT?

TomSweeneyRedHat:
Pinging Dave mostly  to make sure this is tracked for RHEL if we move forward with this.

nalind:
I keep meaning to bring up the sshfs/sftp part of it at a meeting to get a better feel for whether or not we want to include that part of this, since it has limitations if you try to use that outside of test cases.

nalind:
Been kicking the idea around since https://redhat.atlassian.net/browse/RUN-1457, finally had a reason to look into it.

Luap99:
I think it is certainly a nice use case for the tests. Maybe hiding the commands and label it unsupported on RHEL would help to reduce or maintenance load.

ddarrah:
It's a nice approach.  Do I understand correctly that the pipeloop is self contained in the test directory, and that sshfs is already a part of RHEL?  If not, I see no issue with installing it simply for the tests from EPEL.  We pull BATS in to run tests.  I just don't want any part of a new RHEL package, especially just to run tests.

nalind:
The pipeloop helper is standalone, but AFAIK sshfs is not a part of RHEL or EPEL.

nalind:
> The pipeloop helper is standalone, but AFAIK sshfs is not a part of RHEL or EPEL.

My mistake: I was looking for a package named "sshfs", not "fuse-sshfs".  It's been noted elsewhere that fuse-sshfs is in EPEL 9 and 10.

Changed the new commands to be undocumented.

cmd/buildah/sftp.go
```diff
@@ -0,0 +1,914 @@
...
+// filesystem is mounted for use by an sftp server.
+func lockServingSFTP(datadir string) error {
+	lockfile := filepath.Join(datadir, servingSftpFragment)
+	fd, err := unix.Open(lockfile, unix.O_RDWR|unix.O_TRUNC|unix.O_CREAT, 0o600)
+	if err != nil {
+		return &os.PathError{Op: "open", Path: lockfile, Err: err}
+	}
+	if err := unix.Flock(fd, unix.LOCK_EX|unix.LOCK_NB); err != nil {
+		return &os.PathError{Op: "flock", Path: lockfile, Err: err}
+	}
```
Luap99: the file is never closed, now I get you want this file to stay locked for the full command duration so a leak outside the function is expected but I would return the fd and let the caller defer close it then.
Also in the error case the file should be closed for any flock error

cmd/buildah/sftp.go
```diff
@@ -0,0 +1,914 @@
...
+	listSshfsMountDirs.Stdout = &listSshfsMountDirsStdout
+	listSshfsMountDirs.Stderr = os.Stderr
+	logrus.Debugf("running %v to get a list of possible sshfs mountpoints", listSshfsMountDirsArgs)
+	if err := listSshfsMountDirs.Run(); err != nil {
+		return nil, nil, fmt.Errorf("listing candidate locations where containers or images would be mounted: %w", err)
+	}
+	namesOfMountPoints := make(map[string]string)
+	var mountPoints []string
+	for {
+		whatName, err := listSshfsMountDirsStdout.ReadString('\n')
```
Luap99: I think using buifio.Scanner for reading line by would be better? then you do not need to trim of the newline as well

cmd/buildah/sftp.go
```diff
@@ -0,0 +1,914 @@
...
+		mountPoint = strings.TrimSuffix(mountPoint, "\n")
+		var parentStat, mountPointStat unix.Stat_t
+		if err2 := unix.Stat(filepath.Dir(mountPoint), &parentStat); err2 != nil {
+			return nil, nil, fmt.Errorf("stat %q: %w", filepath.Dir(mountPoint), err2)
+		}
+		if err2 := unix.Stat(mountPoint, &mountPointStat); err2 != nil {
+			if errors.Is(err2, os.ErrNotExist) {
+				continue // doesn't look like it could be mounted
+			}
+			return nil, nil, fmt.Errorf("stat %q: %w", mountPoint, err2)
```
Luap99: wrapping the error with &os.PathError{Op: "stat", Path: mountPoint, Err: err} feels a bit better I guess as it mirror the std lib errors of os.Stat()

cmd/buildah/sftp.go
```diff
@@ -0,0 +1,914 @@
...
+
+// mountSshfsCmd mounts a container or image on a container- or image-specific
+// location using sshfs.
+func mountSshfsCmd(c *cobra.Command, args []string) error {
+	inheritedFlags := getInheritedFlags(c)
+
+	selfExecutable, err := os.Executable()
+	if err != nil {
+		return fmt.Errorf("determining our executable: %w", err)
+	}
```
Luap99: since this is a linux only file we technically do not need this can can use `/proc/self/exe` as command path

cmd/buildah/sftp.go
```diff
@@ -0,0 +1,914 @@
...
+		}
+		for _, mountPoint := range mountPoints {
+			fmt.Println(namesOfMountPoints[mountPoint], mountPoint)
+		}
+		return nil
+	}
+
+	if _, err := exec.LookPath("sshfs"); err != nil {
+		return fmt.Errorf(`looking for "sshfs" in $PATH: %w`, err)
+	}
```
Luap99: if we do the sshfs path lookup once here it might be betetr to store the path and use the full path below in the command execution to aviod the path lookup being done twice in a row

cmd/buildah/sftp.go
```diff
@@ -0,0 +1,914 @@
...
+		var parentStat, mountPointStat unix.Stat_t
+		if err := unix.Mkdir(mountPoint, 0o700); err != nil && !errors.Is(err, os.ErrExist) {
+			return fmt.Errorf("mkdir(%q): %w", mountPoint, err)
+		}
+		if err := unix.Stat(mountPoint, &mountPointStat); err != nil {
+			return fmt.Errorf("stat %q: %w", mountPoint, err)
+		}
+		if err := unix.Stat(mountParentDir, &parentStat); err != nil {
+			return fmt.Errorf("stat %q: %w", mountParentDir, err)
+		}
```
Luap99, line 626: same here with the error wrapping &os.PathError{Op: "stat", Path: mountPoint, Err: err}

Luap99 (commented):
mostly nit picking stuff here, I am not to well versed with the sshfs mount logic but overall seems fine

cmd/buildah/sftp.go
nalind: Changing it.

cmd/buildah/sftp.go
nalind: Sure, changing it.

cmd/buildah/sftp.go
nalind: Switching to calling `reexec.Self()`, which returns that value.

cmd/buildah/sftp.go
nalind: Sure, changing it.

cmd/buildah/sftp.go
nalind: Changing these.

cmd/buildah/sftp.go
nalind: Switched to using a bufio.Scanner.

cmd/buildah/sftp.go
```diff
@@ -0,0 +1,934 @@
...
+
+// lockServingSFTP takes a lock on a file to signal to other processes that the
+// filesystem is mounted for use by an sftp server.
+func lockServingSFTP(datadir string) (int, error) {
+	lockfile := filepath.Join(datadir, servingSftpFragment)
+	fd, err := unix.Open(lockfile, unix.O_RDWR|unix.O_TRUNC|unix.O_CREAT, 0o600)
+	if err != nil {
+		return -1, &os.PathError{Op: "open", Path: lockfile, Err: err}
+	}
+	unix.CloseOnExec(fd)
```
Luap99: not sure, if this was already in the first versiona nd I missed it, but just pass `unix.O_CLOEXEC` in the open flags which safes another sycall and thus makes this simpler

cmd/buildah/sftp.go
```diff
@@ -0,0 +1,934 @@
...
+		}
+	}
+	if err != nil {
+		return &os.PathError{Op: "flock", Path: lockfile, Err: err}
+	}
+	return nil
+}
+
+// lockServingSFTP takes a lock on a file to signal to other processes that the
+// filesystem is mounted for use by an sftp server.
```
Luap99: maybe update the comment to say the int is a returned open fd the caller must close to not leak it.

Luap99 (approved):
LGTM, just some minor nit not worth repushing if nobody else points something out

cmd/buildah/sftp.go
nalind: It wasn't in there, but it struck me as something we needed to have.  Will make that change, thanks!

cmd/buildah/sftp.go
nalind: Adding that.

Luap99 (approved):
LGTM

TomSweeneyRedHat:
LGTM
Although it would be nice if SFTP dropped the use of kr/fs and moved to the newer Go Filepath.* routines.  Way outside the purview of this review, however.

PR #6934, opened 2026-06-26
internal/mkcw.GenerateMeasurement(): check for libkrunfw-sev.so.5

nalind:
#### What type of PR is this?

/kind other

We have to care about the soname of the file, even though we never call it directly, and it was bumped since the last time I looked, so just accept that whatever name it has, any libkrunfw-sev.so.* is as good as any other.

#### How to verify it

Install libkrunfw-sev, check what it provides.

None

```release-note
None
```

### DIFF
--- a/internal/mkcw/attest.go
+++ b/internal/mkcw/attest.go
@@ -138,9 +138,6 @@ func SendRegistrationRequest(workloadConfig WorkloadConfig, diskEncryptionPassph
 		return err
 	}
 	parsedURL.Path = path.Join(parsedURL.Path, "/kbs/v0/register_workload")
-	if err != nil {
-		return err
-	}
 	url := parsedURL.String()
 	requestContentType := "application/json"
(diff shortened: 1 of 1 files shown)


internal/mkcw/attest.go
```diff
@@ -204,6 +204,7 @@ func GenerateMeasurement(workloadConfig WorkloadConfig, firmwareLibrary string)
 		sharedLibraryDirs = append(sharedLibraryDirs, strings.Split(llp, ":")...)
 	}
 	libkrunfwNames := []string{
+		"libkrunfw-sev.so.5",
```
Honny1: Could we define those names more generally, maybe with a regex to match all possible names?

Honny1 (commented):
LGTM, I have just one question about maintainability.

internal/mkcw/attest.go
nalind: Sure, changing it.

Luap99 (approved):
LGTM

PR #6941, opened 2026-07-06
run: skip /dev/mqueue mount if not supported by the kernel

adenwuts:
#### What type of PR is this?

/kind bug

On kernels built without `CONFIG_POSIX_MQUEUE`, the default `type=mqueue` mount for `/dev/mqueue` (seeded by runtime-tools' `generate.New()`) makes the runtime fail with `ENODEV` on every `buildah run` — and, through the copy vendored into podman, on every `RUN` instruction in `podman build`.

This ports the fix podman already has for `podman run`/`create` (podman-container-tools/podman#28639) into buildah's spec generation: check `/proc/filesystems` for mqueue support, and drop the `/dev/mqueue` mount from the spec when the kernel doesn't provide the filesystem. When mqueue is unsupported the userns+host-IPC bind-mount replacement is skipped too, since the host has no `/dev/mqueue` to bind in that case.

The `isMqueueSupported` implementation mirrors podman's (`sync.OnceValue`, same parsing) so the two stay consistent.

#### How to verify it

On a kernel without `CONFIG_POSIX_MQUEUE` (no `mqueue` line in `/proc/filesystems`), `buildah run` previously failed with `ENODEV` mounting `/dev/mqueue`; with this change the mount is dropped and the container starts. On normal kernels behavior is unchanged.

The added unit test stubs `isMqueueSupported` and checks that the `/dev/mqueue` mount is present in the generated spec if and only if the kernel supports it.

None

The equivalent podman-side fix landed in podman-container-tools/podman#28639 (commits 1bda61b840 and d8df5c37ef); this change keeps buildah's behavior and implementation aligned with it.

```release-note
Fixed `buildah run` failing with ENODEV on kernels built without CONFIG_POSIX_MQUEUE by omitting the /dev/mqueue mount when the kernel does not support the mqueue filesystem.
```

### DIFF
--- a/run_linux.go
+++ b/run_linux.go
@@ -3,6 +3,7 @@
 package buildah
 
 import (
+	"bufio"
 	"context"
 	"errors"
 	"fmt"
@@ -1373,7 +1374,17 @@ func setupSpecialMountSpecChanges(spec *specs.Spec, shmSize string) ([]specs.Mou
 
 	isUserns := isNewUserns || isRootless
 
-	if isUserns && !isIpcns {
+	// Check /proc/filesystems to see if the kernel supports mqueue.
+	if !isMqueueSupported() {
+		// The default /dev/mqueue mount would fail with ENODEV, so drop it.
+		filtered := make([]specs.Mount, 0, len(mounts))
+		for _, m := range mounts {
+			if m.Destination != "/dev/mqueue" {
+				filtered = append(filtered, m)
+			}
+		}
+		mounts = filtered
+	} else if isUserns && !isIpcns {
 		devMqueue := "/dev/mqueue"
 		devMqueueMnt := specs.Mount{
 			Destination: devMqueue,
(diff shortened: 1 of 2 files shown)


run_linux.go
```diff
@@ -1429,6 +1440,27 @@ func setupSpecialMountSpecChanges(spec *specs.Spec, shmSize string) ([]specs.Mou
 	return mounts, nil
 }
 
+// isMqueueSupported reports whether the kernel supports the mqueue
+// filesystem, i.e. was built with CONFIG_POSIX_MQUEUE.
+var isMqueueSupported = sync.OnceValue(func() bool {
```
Luap99, line 1445: if this uses the same function as podman it should be moved to the common library, maybe as a new pkg there pkg/fsinfo or something like that? 
So that then podman and buildah can reuse the same function from one place which means we also only have to cache it once.

adenwuts: Good call, I can take a look.

PR #6942, opened 2026-07-06
golangci-lint: enable nilnesserr, modernize and more govet checks

Luap99:

Enable more checks that catch issues and fix some of them. In this case
things reported by nilness and unusedwrite.
I enabled all checks except fieldalignment and shadow as they produce a
lot of warnings that are not real issues.

see https://golangci-lint.run/docs/linters/configuration/#govet

---

golangci-lint: show all issues at once

By default golangci-lint does not show all error which makes it look
like there is not that much to fix, however in just shows a maximum of
50 issues per linter and a maximum of 3 issues with the same text.

It is much better to see all at once locally to know how much work it is
to fix a new linter.

```release-note
None
```

### DIFF
--- a/bind/mount.go
+++ b/bind/mount.go
@@ -254,9 +254,6 @@ func UnmountMountpoints(mountpoint string, mountpointsToRemove []string) error {
 	// find the top of the tree we're unmounting
 	top := getMountByPoint(mountpoint)
 	if top == nil {
-		if err != nil {
-			return fmt.Errorf("%q is not mounted: %w", mountpoint, err)
-		}
 		return nil
 	}
 	// add all of the mounts that are hanging off of it
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -361,8 +361,7 @@ func joinExcludePatternWithCopySource(srcNorm, excl string) string {
 	if srcNorm == "." {
 		return excl
 	}
-	if strings.HasPrefix(excl, "!") {
-		rest := strings.TrimPrefix(excl, "!")
+	if rest, ok := strings.CutPrefix(excl, "!"); ok {
 		if rest == "" {
 			return excl
 		}
@@ -840,11 +839,12 @@ func (s *stageExecutor) Run(run imagebuilder.Run, config docker.Config) error {
 				args = []string{run.Files[0].Data}
 			}
 		} else {
-			full := args[0]
+			var full strings.Builder
+			full.WriteString(args[0])
 			for _, file := range run.Files {
-				full += file.Data + "\n" + file.Name
+				full.WriteString(file.Data + "\n" + file.Name)
 			}
-			args = []string{full}
+			args = []string{full.String()}
 		}
 	}
 	stageMountPoints, err := s.runStageMountPoints(slices.Concat(run.Mounts, s.executor.transientRunMounts))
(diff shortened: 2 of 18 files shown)


bind/mount.go
```diff
@@ -254,10 +254,7 @@ func UnmountMountpoints(mountpoint string, mountpointsToRemove []string) error {
 	// find the top of the tree we're unmounting
 	top := getMountByPoint(mountpoint)
 	if top == nil {
-		if err != nil {
-			return fmt.Errorf("%q is not mounted: %w", mountpoint, err)
-		}
-		return nil
+		return fmt.Errorf("%q is not mounted", mountpoint)
```
Luap99: looks like since this always return nil before no errors were reported to the callers, no in the tests we see
```
level=warning msg="pkg/bind: error unmounting \"/tmp/buildah4026288278/mnt/rootfs\": \"/tmp/buildah4026288278/mnt/rootfs\" is not mounted"
```

So I guess I should just do return nil here always? cc @nalind

nalind: Yeah, this should probably just be returning `nil`.  The caller could be keeping better track of the state of the mountpoint, but it's entirely possible to get here without encountering other errors.

nalind (approved):
LGTM, thanks!

Luap99:
I added two more useful linters

imagebuildah/stage_executor.go
```diff
@@ -361,8 +361,8 @@ func joinExcludePatternWithCopySource(srcNorm, excl string) string {
 	if srcNorm == "." {
 		return excl
 	}
-	if strings.HasPrefix(excl, "!") {
-		rest := strings.TrimPrefix(excl, "!")
+	if after, ok := strings.CutPrefix(excl, "!"); ok {
+		rest := after
```
Honny1: Maybe this assignment is not necessary?

pkg/cli/common.go
```diff
@@ -594,8 +594,8 @@ func LookupEnvVarReferences(specs, environ []string) []string {
 			result = append(result, environ...)
 		} else {
 			prefix := key + "="
-			if strings.HasSuffix(key, "*") {
-				prefix = strings.TrimSuffix(key, "*")
+			if before, ok0 := strings.CutSuffix(key, "*"); ok0 {
```
Honny1: Is this the correct version?
```suggestion
			if before, ok := strings.CutSuffix(key, "*"); ok {
```

internal/mkcw/archive.go
```diff
@@ -552,8 +552,8 @@ func slop(size int64, slop string) int64 {
 		if factor == "" {
 			continue
 		}
-		if strings.HasSuffix(factor, "%") {
-			percentage := strings.TrimSuffix(factor, "%")
+		if before, ok := strings.CutSuffix(factor, "%"); ok {
+			percentage := before
```
Honny1: Maybe this assignment is not necessary?

Honny1 (commented):
I have some non-blocking comments. Othervise LGTM.

pkg/cli/common.go
Luap99: I mean the var name does not matter, does it? I guess the automatic replacement choose ok0 to aviod shadowing the ok in the parent scope.
I guess we can name it ok again here as the shadowing is harmless here

imagebuildah/stage_executor.go
Luap99: yes good point

Luap99: fixed the comments, thanks

TomSweeneyRedHat:
LGTM

PR #6949, opened 2026-07-08
vendor: update container-libs to latest

Luap99:

With the common release and merge back into main over there renovate gets broken and fails to lookup new common commits so it will no longer offer to update common, i.e.
https://github.com/podman-container-tools/buildah/pull/6940

It only updates image and storage then which is not what we want so do it manually once then it should work again afterwards as the commits are after the merged back tag.

```release-note
None
```


nalind (approved):
LGTM

TomSweeneyRedHat:
LGTM

PR #6957, opened 2026-07-12
imagebuildah: default RUN --mount type to bind for cache checksums

### LINKED ISSUE #29126: Build contexts get mixed up
### Issue Description

In some cases, files end up in a wrong image when building multiple services via compose.

```
$ tree --charset=ascii
.
|-- apps
|   |-- bar
|   |   |-- Dockerfile
|   |   `-- data
|   `-- foo
|       |-- Dockerfile
|       `-- data
`-- compose.yaml

4 directories, 5 files
```
```
$ cat apps/foo/Dockerfile
FROM docker.io/alpine:3.24.1 AS build

RUN mkdir /out
RUN --mount=source=./data,target=data \
    cp data /out


FROM docker.io/alpine:3.24.1

COPY --from=build /out /out

ENTRYPOINT ["cat", "/out/data"]
```

```
$ cat apps/bar/Dockerfile
FROM docker.io/alpine:3.24.1 AS build

RUN mkdir /out
RUN --mount=source=./data,target=data \
    cp data /out


FROM docker.io/alpine:3.24.1

COPY --from=build /out /out

ENTRYPOINT ["cat", "/out/data"]
```

```
$ cat apps/foo/data
DATA: foo
```

```
$ cat apps/bar/data
DATA: bar
```

```
$ cat compose.yaml
services:
  foo:
    build:
      context: ./apps/foo
  bar:
    build:
      context: ./apps/bar
```

```
$ podman compose up --build --force-recreate --remove 2>&1 | grep DATA
[bar] | DATA: bar
[foo] | DATA: foo

$ podman compose up --build --force-recreate --remove 2>&1 | grep DATA
[bar] | DATA: bar
[foo] | DATA: bar

$ podman compose up --build --force-recreate --remove 2>&1 | grep DATA
[bar] | DATA: bar
[foo] | DATA: bar

$ podman compose down
>>>> Executing external compose provider "/nix/store/ppwj4hj2rkxhrvg95mwbb34walh3q48j-podman-compose-1.6.0/bin/podman-compose". Please see podman-compose(1) for how to disable this message. <<<<

test_foo_1
test_bar_1
test_bar_1
test_foo_1
326bbf15e58a8b05908274611c5869ad209d3fa236b5b213ded45a272d328ac2
test_default

$ podman system prune -f 2>&1 >/dev/null

$ podman compose up --build --force-recreate --remove 2>&1 | grep DATA
[foo] | DATA: foo
[bar] | DATA: bar

$ podman compose up --build --force-recreate --remove 2>&1 | grep DATA
[foo] | DATA: foo
[bar] | DATA: foo

$ podman compose up --build --force-recreate --remove 2>&1 | grep DATA
[foo] | DATA: foo
[bar] | DATA: foo
```

Files should be copied from the right context.

### podman info output

```yaml
$ podman info
host:
  arch: amd64
  buildahVersion: 1.44.0
  cdiSpecDirs:
  - /etc/cdi
  - /var/run/cdi
  cgroupControllers:
  - cpu
  - io
  - memory
  - pids
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: Unknown
    path: /nix/store/h6my01ir1pl92315z23ymr10b3b6asg2-podman-helper-binary-wrapper/bin/conmon
    version: 'conmon version 2.2.1, commit: c8cc2c4db27531bd4e084ce7857f73cd21ee639d'
  cpuUtilization:
    idlePercent: 89.16
    systemPercent: 2.05
    userPercent: 8.79
  cpus: 8
  databaseBackend: sqlite
  distribution:
    codename: yarara
    distribution: nixos
    version: "26.05"
  eventLogger: journald
  freeLocks: 2030
  hostname: work
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 100
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 6.18.36
  linkmode: dynamic
  logDriver: journald
  memFree: 1995509760
  memTotal: 33014988800
  networkBackend: netavark
  networkBackendInfo:
    backend: netavark
    defaultNetwork: podman
    dns:
      package: Unknown
      path: /nix/store/ls4r52263z48i1gh2kmsasxmd7r636gg-podman-6.0.0/libexec/podman/aardvark-dns
      version: aardvark-dns 2.0.0
    package: Unknown
    path: /nix/store/ls4r52263z48i1gh2kmsasxmd7r636gg-podman-6.0.0/libexec/podman/netavark
    version: netavark 2.0.0
  ociRuntime:
    name: crun
    package: Unknown
    path: /nix/store/h6my01ir1pl92315z23ymr10b3b6asg2-podman-helper-binary-wrapper/bin/crun
    version: |-
      crun version 1.27.1
      commit: 3ec076b3b6714ec2f1a10533cf18d5605a6de637
      rundir: /run/user/1000/crun
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +LIBKRUN +YAJL
  os: linux
  pasta:
    executable: /nix/store/ls4r52263z48i1gh2kmsasxmd7r636gg-podman-6.0.0/libexec/podman/pasta
    package: Unknown
    version: |
      pasta 2025_09_19.623dbf6
      Copyright Red Hat
  remoteSocket:
    exists: true
  rootlessNetworkCmd: pasta
  rootlessPortForwarder: rootlessport
  security:
    apparmorEnabled: false
    rootless: true
    seccompEnabled: true
    seccompProfilePath: ""
    selinuxEnabled: false
  serviceIsRemote: false
  swapFree: 0
  swapTotal: 0
  uptime: 265h 57m 24.00s (Approximately 11.04 days)
  variant: ""
plugins:
  authorization: null
  log:
  - k8s-file
  - none
  - passthrough
  - journald
  network:
  - bridge
  - macvlan
  - ipvlan
  volume:
  - local
registries: {}
store:
  containerStore:
    number: 3
    paused: 0
    running: 0
    stopped: 3
  graphDriverName: overlay
  graphOptions: {}
  graphRoot: /home/dmytro/.local/share/containers/storage
  graphRootAllocated: 274861129728
  graphRootUsed: 161695076352
  graphStatus:
    Backing Filesystem: btrfs
    Native Overlay Diff: "true"
    Supports d_type: "true"
    Supports shifting: "false"
    Supports volatile: "true"
    Using metacopy: "false"
  imageCopyTmpDir: /var/tmp
  imageStore:
    number: 456
  transientStore: false
  volumePath: /home/dmytro/.local/share/containers/storage/volumes
version:
  APIVersion: 6.0.0
  Built: 315532800
  BuiltTime: Tue Jan  1 00:00:00 1980
  GitCommit: ""
  GoVersion: go1.26.4
  Os: linux
  OsArch: linux/amd64
  Version: 6.0.0
```

### Podman in a container

No

### Privileged Or Rootless

None

### Upstream Latest Release

Yes

### Additional information

MayukhSobo:
/kind bug

`getFromAndSourceKeysFromMountFlag()` parses a `RUN --mount=...` flag to decide whether `getCreatedBy()` should checksum the mounted content for the build-cache fingerprint. When a Dockerfile omits `type=` (the normal, idiomatic way to write a bind mount, since bind is the implicit default per the Dockerfile/BuildKit mount syntax), this parser left `Type` empty instead of defaulting it to `"bind"` — unlike buildah's other mount parser, the one that actually sets mounts up (`internal/volumes/volumes.go`'s `getMounts()`, which does default `mountType := define.TypeBind`).

Because of that, `getCreatedBy()`'s `if mountInfo.Type != "bind" { continue }` check skipped checksumming the mounted content for any implicit-bind mount — whether sourced from the build context, another build stage, an image, or an additional build context. Two builds with identical Dockerfile text but different mounted file content ended up with an identical cache key, so the second build silently reused the first build's cached layer instead of picking up the new content.

The fix defaults the parsed mount type to `define.TypeBind`, matching the default already used by the real mount-setup parser. `type=cache` and `type=tmpfs` mounts remain correctly excluded from checksumming — that's intentional, since persisting regardless of build content is their whole purpose.

#### How to verify it

1. Create a build context with a file, e.g. `data.txt` containing `foo`.
2. Build a Dockerfile that uses an implicit bind mount (no `type=`):
   ```
   FROM alpine
   RUN --mount=src=data.txt,target=/tmp/data.txt cat /tmp/data.txt
   ```
   using `buildah build --layers .` — it prints `foo`.
3. Change `data.txt` to contain `bar`, then rebuild the same Dockerfile with `--layers` again.
4. **Before this fix**: the second build reuses the cached layer and still prints `foo`, even though the mounted file changed. **After this fix**: the cache is correctly invalidated and the second build prints `bar`.

A unit test was also added, `TestGetFromAndSourceKeysFromMountFlag` in `imagebuildah/util_test.go`, which fails without this fix (asserts `Type` defaults to `"bind"` when `type=` is omitted) and passes with it:
```
go test -run TestGetFromAndSourceKeysFromMountFlag ./imagebuildah/...
```

Fixes podman-container-tools/podman#29126

This bug was reported against podman (originally containers/podman#29126, now podman-container-tools/podman#29126 after the org transfer), but root-caused to buildah — `podman build` vendors buildah's `imagebuildah` package. There is no separate buildah-side issue tracking this; per CONTRIBUTING.md, a PR alone is sufficient as long as its description includes what an issue report would (repro steps + root cause), which this PR provides above.

```release-note
Fixed a build-cache bug where `RUN --mount` flags that omit `type=` (an implicit bind mount) were not checksummed for cache invalidation, causing builds to silently reuse a stale cached layer when only the mounted file's content changed.
```

### DIFF
--- a/imagebuildah/util.go
+++ b/imagebuildah/util.go
@@ -10,6 +10,7 @@ import (
 
 	digest "github.com/opencontainers/go-digest"
 	"go.podman.io/buildah"
+	"go.podman.io/buildah/define"
 )
 
 type mountInfo struct {
@@ -19,12 +20,14 @@ type mountInfo struct {
 }
 
 // Consumes mount flag in format of `--mount=type=bind,src=/path,from=image` and
-// return mountInfo with values, otherwise values are empty if keys are not present in the option.
+// return mountInfo with values. Source and From are empty if not present in the
+// option; Type defaults to "bind" if not present, matching the default used
+// when the mount is actually set up (see internal/volumes.getMounts).
 func getFromAndSourceKeysFromMountFlag(mount string) mountInfo {
 	tokens := strings.Split(strings.TrimPrefix(mount, "--mount="), ",")
 	source := ""
 	from := ""
-	mountType := ""
+	mountType := define.TypeBind
 	for _, option := range tokens {
 		if optionSplit := strings.Split(option, "="); len(optionSplit) == 2 {
 			if optionSplit[0] == "src" || optionSplit[0] == "source" {
--- a/imagebuildah/util_test.go
+++ b/imagebuildah/util_test.go
@@ -10,7 +10,51 @@ import (
 
 	"github.com/opencontainers/go-digest"
 	"github.com/stretchr/testify/assert"
+	"go.podman.io/buildah/define"
 )
+
+func TestGetFromAndSourceKeysFromMountFlag(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name     string
+		mount    string
+		expected mountInfo
+	}{
+		{
+			name:     "type omitted defaults to bind",
+			mount:    "--mount=src=/foo,target=/bar",
+			expected: mountInfo{Type: define.TypeBind, Source: "/foo"},
+		},
+		{
+			name:     "explicit bind type",
+			mount:    "--mount=type=bind,src=/foo,target=/bar",
+			expected: mountInfo{Type: "bind", Source: "/foo"},
+		},
+		{
+			name:     "cache type is not treated as bind",
+			mount:    "--mount=type=cache,target=/bar",
+			expected: mountInfo{Type: "cache"},
+		},
+		{
+			name:     "tmpfs type is not treated as bind",
+			mount:    "--mount=type=tmpfs,target=/bar",
+			expected: mountInfo{Type: "tmpfs"},
+		},
+		{
+			name:     "from without type defaults to bind",
+			mount:    "--mount=from=builder,src=/foo,target=/bar",
+			expected: mountInfo{Type: define.TypeBind, Source: "/foo", From: "builder"},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			assert.Equal(t, tt.expected, getFromAndSourceKeysFromMountFlag(tt.mount))
+		})
+	}
+}
 
 func TestGeneratePathChecksum(t *testing.T) {
 	t.Parallel()


nalind (approved):
LGTM, thanks!

TomSweeneyRedHat:
Nice!  TYVM!
LGTM

PR #6961, opened 2026-07-13
chroot: skip setgroups when namespace denies it

### LINKED ISSUE #6947: chroot: RUN steps always fail with "error setting supplemental groups list: operation not permitted" when the user namespace denies setgroups (regression of #2459)
## Description

When buildah runs rootless inside a **single-ID user namespace whose
`/proc/self/setgroups` is `deny`** (the unavoidable situation when the invoking UID has
no subuid/subgid ranges and no privileged `newuidmap`/`newgidmap` helpers — e.g.
Kubernetes/OpenShift pods under a restricted SCC, or `podman run --user <uid>` into an
image with no subid configuration), **every Dockerfile `RUN` step under
`--isolation chroot` fails**:

```
error setting supplemental groups list: operation not permitted
subprocess exited with status 1
```

The chroot runtime has a deny-tolerant branch that was added for exactly this
environment (see #2459, closed as completed in 2020), but it is **unreachable today**:
the spec generation unconditionally injects the primary GID into the RUN user's
supplemental groups, so the group list is never empty and `setgroups()` is always
attempted — and in a setgroups-denied namespace that call can only return `EPERM`.

## Steps to reproduce

No cluster needed; rootless podman on any host reproduces it:

```sh
# 1. An environment image: buildah + a passwd entry for an arbitrary uid.
#    (The passwd row only avoids an unrelated startup failure of the re-exec'd
#    chroot child, so that the setgroups bug is reached.)
cat > /tmp/env.Containerfile <<'EOF'
FROM registry.access.redhat.com/ubi9/ubi-minimal
RUN microdnf -y install buildah \
 && echo "build:x:1000:0::/tmp/h:/sbin/nologin" >> /etc/passwd
EOF
podman build -t buildah-setgroups-repro -f /tmp/env.Containerfile

# 2. Run it as uid 1000 (no subuid/subgid ranges inside the container),
#    and do a trivial build with one RUN step.
podman run --rm --user 1000:0 buildah-setgroups-repro sh -c '
  export HOME=/tmp/h; mkdir -p /tmp/h /tmp/ctx
  printf "FROM registry.access.redhat.com/ubi9/ubi-minimal\nRUN true\n" > /tmp/ctx/Containerfile
  buildah build --isolation chroot --storage-driver vfs \
    --storage-opt vfs.ignore_chown_errors=true \
    -f /tmp/ctx/Containerfile /tmp/ctx'
```

(`ignore_chown_errors` only lets the base image extract in the single-ID namespace;
it is unrelated to the bug itself.)

## Describe the results you received

```
STEP 2/2: RUN true
error setting supplemental groups list: operation not permitted
subprocess exited with status 1
Error: building at STEP "RUN true": exit status 1
```

## Describe the results you expected

The `RUN` step executes. When `/proc/self/setgroups` reads `deny`, the kernel has
permanently frozen the supplementary-group set of every process in the namespace —
`setgroups()` can never succeed *and* can never be needed, so it should be skipped
(which is exactly what the existing `else` branch already does for the empty-list
case).

## Analysis

Two interacting pieces (permalinks to v1.43.1):

1. [`run_common.go#L285`](https://github.com/podman-container-tools/buildah/blob/v1.43.1/run_common.go#L285)
   — `configureUIDGID` unconditionally does `g.AddProcessAdditionalGid(user.GID)`, so
   `spec.Process.User.AdditionalGids` is **never empty** for a RUN step.
2. [`chroot/run_common.go#L699-L718`](https://github.com/podman-container-tools/buildah/blob/v1.43.1/chroot/run_common.go#L699-L718)
   — the non-empty branch calls `syscall.Setgroups(gids)` and hard-exits on error, with
   no deny check; the deny check exists only in the `else` (empty-list) branch, which
   (1) makes unreachable.

Because of (1), the deny-tolerance added after #2459 (see also PR #1640) no longer has
any effect — the scenario that issue was closed for is broken again. This reproduces
on every tag I checked back to v1.29 and the code is unchanged on `main` as of
2026-07-08.

Workarounds that do **not** help, for completeness:

- `--group-add keep-groups` — only sets the `run.oci.keep_original_groups` annotation,
  which the chroot isolation code never reads; the primary-GID injection happens
  regardless.
- `--userns-uid-map 0:0:1 --userns-gid-map 0:0:1` — explicit maps route layer storage
  through id-mapping, which hard-fails extracting any base image containing files owned
  by other ids (`container ID N cannot be mapped to a host ID`), and
  `ignore_chown_errors` does not apply to that path.

## Suggested fix

Check the namespace's setgroups state first and skip the doomed call (we have been
running this in production CI since 2026-07-08 without regressions; the same read is
already done in the `else` branch). Happy to open a PR:

```diff
--- a/chroot/run_common.go
+++ b/chroot/run_common.go
@@ -696,19 +696,27 @@
 
 	// Drop privileges.
 	user := options.Spec.Process.User
+	// When this user namespace was created without privileged id-mapping
+	// helpers, the kernel has setgroups permanently denied and ANY
+	// setgroups() call fails with EPERM. The supplementary-group set is
+	// frozen by the kernel in that case, so skipping is the only possible
+	// behavior.
+	setgroupsState, _ := os.ReadFile("/proc/self/setgroups")
+	setgroupsDenied := strings.Trim(string(setgroupsState), "\n") == "deny"
 	if len(user.AdditionalGids) > 0 {
-		gids := make([]int, len(user.AdditionalGids))
-		for i := range user.AdditionalGids {
-			gids[i] = int(user.AdditionalGids[i])
-		}
-		logrus.Debugf("setting supplemental groups")
-		if err = syscall.Setgroups(gids); err != nil {
-			fmt.Fprintf(os.Stderr, "error setting supplemental groups list: %v\n", err)
-			os.Exit(1)
+		if setgroupsDenied {
+			logrus.Debugf("setgroups is denied in this user namespace; not setting supplemental groups")
+		} else {
+			gids := make([]int, len(user.AdditionalGids))
+			for i := range user.AdditionalGids {
+				gids[i] = int(user.AdditionalGids[i])
+			}
+			logrus.Debugf("setting supplemental groups")
+			if err = syscall.Setgroups(gids); err != nil {
+				fmt.Fprintf(os.Stderr, "error setting supplemental groups list: %v\n", err)
+				os.Exit(1)
+			}
 		}
 	} else {
-		setgroups, _ := os.ReadFile("/proc/self/setgroups")
-		if strings.Trim(string(setgroups), "\n") != "deny" {
+		if !setgroupsDenied {
 			logrus.Debugf("clearing supplemental groups")
 			if err = syscall.Setgroups([]int{}); err != nil {
 				fmt.Fprintf(os.Stderr, "error clearing supplemental groups list: %v\n", err)
```

An alternative (possibly preferable) fix is to stop force-adding the primary GID in
`configureUIDGID` when it would be the only entry — that would also make the existing
deny-guard reachable again.

## Version

```
buildah version 1.43.1 (image-spec 1.1.1, runtime-spec 1.2.1)
```

(RHEL 9 build `buildah-1.43.1-*.el9_8`; the relevant code is identical on `main` as of
2026-07-08.)

## Environment

- `--isolation chroot`, `--storage-driver vfs`.
- Reproduced: rootless podman on kernel 6.6 (repro above) and Kubernetes/OpenShift
  runner pods under a restricted SCC (non-root UID, no subuid/subgid ranges,
  `allowPrivilegeEscalation: false`, so `newuidmap`/`newgidmap` file capabilities are
  inert and the namespace is created with `setgroups = deny`).

ajoshua2004:
#### What type of PR is this?

/kind bug

In a single-ID user namespace whose /proc/self/setgroups is "deny", any setgroups() call returns EPERM, so every RUN step under --isolation chroot fails. The kernel has permanently frozen the supplementary group set in that case, so the call can neither succeed nor be needed, and is skipped. The read is split into Linux/BSD helpers since /proc/self/setgroups does not exist on FreeBSD.

This keeps the primary-GID force-add in configureUIDGID rather than removing it, leaving the protection intact

Adds a regression test that a RUN step succeeds in a setgroups-denied namespace, and a test that the skip does not widen access: a probe file owned by a group the build namespace cannot map (surfacing as overflow gid 65534, present in the process's groups list) is still denied, while a legitimately-held group is readable.

This builds on the patch michelemodolo proposed in #6947, extended to gate on the setgroups-denied state, split across the Linux/BSD paths, with regression and no leak tests added.

#### How to verify it
Two new Linux only tests in tests/chroot.bats
`bats ./tests/chroot.bats`
- "chroot RUN succeeds when setgroups is denied" — builds with a RUN step
  inside a setgroups-denied namespace (unshare --user --map-root-user) and
  asserts it reaches the RUN step instead of failing with EPERM. 
- "chroot RUN does not honor unmapped supplemental group when setgroups is
  denied" — a probe file owned by a group the build's single-ID namespace
  can't map surfaces inside as overflow gid 65534 and appears in the process's
  groups list, yet the read is still denied, while a legitimately-held group
  reads fine. Confirms skipping setgroups() doesn't widen access.
Fixes #6947 

Chose to gate the setgroups() call on the deny state rather than dropping the
primary-GID force-add in configureUIDGID (the issue's alternative), since that
force-add is the fix for rhbz#2121453 and removing it risks reopening it.

```release-note
Fixed chroot isolation RUN steps failing with "error setting supplemental groups list: operation not permitted" in user namespaces where setgroups is denied.
```

### DIFF
--- a/chroot/run_common.go
+++ b/chroot/run_common.go
@@ -684,24 +684,33 @@ func runUsingChrootExecMain() {
 
 	// Drop privileges.
 	user := options.Spec.Process.User
+
+	// Check if /proc/self/setgroups is set to "deny" before attempting to set supplemental groups.
+	denied := setgroupsDenied()
+
 	if len(user.AdditionalGids) > 0 {
-		gids := make([]int, len(user.AdditionalGids))
-		for i := range user.AdditionalGids {
-			gids[i] = int(user.AdditionalGids[i])
-		}
-		logrus.Debugf("setting supplemental groups")
-		if err = syscall.Setgroups(gids); err != nil {
-			fmt.Fprintf(os.Stderr, "error setting supplemental groups list: %v\n", err)
-			os.Exit(1)
+		if denied {
+			logrus.Debugf("not setting supplemental groups %v: /proc/self/setgroups is \"deny\"", user.AdditionalGids)
+		} else {
+			gids := make([]int, len(user.AdditionalGids))
+			for i := range user.AdditionalGids {
+				gids[i] = int(user.AdditionalGids[i])
+			}
+			logrus.Debugf("setting supplemental groups")
+			if err = syscall.Setgroups(gids); err != nil {
+				fmt.Fprintf(os.Stderr, "error setting supplemental groups list: %v\n", err)
+				os.Exit(1)
+			}
 		}
 	} else {
-		setgroups, _ := os.ReadFile("/proc/self/setgroups")
-		if strings.Trim(string(setgroups), "\n") != "deny" {
+		if !denied {
 			logrus.Debugf("clearing supplemental groups")
 			if err = syscall.Setgroups([]int{}); err != nil {
 				fmt.Fprintf(os.Stderr, "error clearing supplemental groups list: %v\n", err)
 				os.Exit(1)
 			}
+		} else {
+			logrus.Debugf("not clearing supplemental groups: /proc/self/setgroups is \"deny\"")
 		}
 	}
 
--- a/chroot/run_linux.go
+++ b/chroot/run_linux.go
@@ -838,3 +838,16 @@ func setPdeathsig(cmd *exec.Cmd) {
 	}
 	cmd.SysProcAttr.Pdeathsig = syscall.SIGKILL
 }
+
+// setgroupsDenied reports whether this process's user namespace denies
+// setgroups(). Returns false on read error, so callers still attempt it.
+func setgroupsDenied() bool {
+	setgroupState, err := os.ReadFile("/proc/self/setgroups")
+	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			logrus.Warnf("error reading /proc/self/setgroups: %v", err)
+		}
+		return false
+	}
+	return strings.TrimSpace(string(setgroupState)) == "deny"
+}
--- a/tests/chroot.bats
+++ b/tests/chroot.bats
@@ -215,3 +215,157 @@ EOF
   chmod +rx ${TEST_SCRATCH_DIR}/script1 ${TEST_SCRATCH_DIR}/script2
   env -i unshare -inCfpm bash ${TEST_SCRATCH_DIR}/script1
 }
+
+@test "chroot RUN succeeds when setgroups is denied" {
+  skip_if_no_unshare
+  if test `uname` != Linux ; then
+    skip "not meaningful except on Linux"
+  fi
+
+  baseimage=registry.access.redhat.com/ubi10:latest
+  _prefetch $baseimage
+  baseimagef=$(tr -c a-zA-Z0-9.- - <<< "$baseimage")
+
+  storagedir=${TEST_SCRATCH_DIR}/storage
+  mkdir $storagedir
+  rootdir=${storagedir}/rootdir
+  mkdir $rootdir
+  runrootdir=${storagedir}/runrootdir
+  mkdir $runrootdir
+  # ignore_chown_errors lets the base image extract in the single-ID namespace,
+  # where high gids in the rootfs can't be represented; unrelated to the bug.
+  storageopts="--storage-driver vfs --storage-opt vfs.ignore_chown_errors=true --root $rootdir --runroot $runrootdir"
+
+  context=${TEST_SCRATCH_DIR}/context ; mkdir $context
+  echo "FROM $baseimage" > $context/Dockerfile
+  echo "RUN echo regression-marker-ran" >> $context/Dockerfile
+
+  cp ${COPY_BINARY} ${TEST_SCRATCH_DIR}/copy
+  cp ${BUILDAH_BINARY} ${TEST_SCRATCH_DIR}/buildah
+
+  echo set -e > ${TEST_SCRATCH_DIR}/script.sh
+  echo "mount -t tmpfs tmpfs /run/lock" >> ${TEST_SCRATCH_DIR}/script.sh
+  # confirm we really are in a setgroups-denied namespace -- the bug's precondition
+  echo "test \"\$(cat /proc/self/setgroups)\" = deny" >> ${TEST_SCRATCH_DIR}/script.sh
+  # seed the store from the prefetched cache so the build needs no network
+  echo "${TEST_SCRATCH_DIR}/copy ${storageopts} dir:\$_BUILDAH_IMAGE_CACHEDIR/$baseimagef containers-storage:$baseimage" >> ${TEST_SCRATCH_DIR}/script.sh
+  # the actual regression: this build used to fail at the RUN step with EPERM
+  echo "${TEST_SCRATCH_DIR}/buildah ${BUILDAH_REGISTRY_OPTS} ${storageopts} build --isolation chroot --pull=never $context" >> ${TEST_SCRATCH_DIR}/script.sh
+
+  run unshare --user --map-root-user --mount bash ${TEST_SCRATCH_DIR}/script.sh
+  echo "$output"
+  assert "$status" -eq 0
+  expect_output --substring "regression-marker-ran"
+}
+
+@test "chroot run denies group access to a member when group bits are restrictive" {
+  skip_if_no_unshare
+  if test `uname` != Linux ; then
+    skip "not meaningful except on Linux"
+  fi
+  if ! test -e /etc/subgid ; then
+    skip "no /etc/subgid allocation to set up the probe"
+  fi
+  if ! test -e /etc/subuid ; then
+    skip "no /etc/subuid allocation to set up the probe"
+  fi
+
+  baseimage=registry.access.redhat.com/ubi10:latest
+  _prefetch $baseimage
+  baseimagef=$(tr -c a-zA-Z0-9.- - <<< "$baseimage")
+
+  storagedir=${TEST_SCRATCH_DIR}/storage
+  mkdir $storagedir
+  rootdir=${storagedir}/rootdir
+  mkdir $rootdir
+  runrootdir=${storagedir}/runrootdir
+  mkdir $runrootdir
+  storageopts="--storage-driver vfs --storage-opt vfs.ignore_chown_errors=true --root $rootdir --runroot $runrootdir"
+
+  probe=${TEST_SCRATCH_DIR}/probe
+  mkdir $probe
+
+  subgidline=$(grep "^$(id -un):" /etc/subgid | head -1)
+  substart=$(echo "$subgidline" | cut -d: -f2)
+  subsize=$(echo "$subgidline" | cut -d: -f3)
+  if test -z "$substart" || test -z "$subsize" ; then
+    skip "no subgid range for $(id -un) to set up the probe"
+  fi
+
+  subuidline=$(grep "^$(id -un):" /etc/subuid | head -1)
+  subuidstart=$(echo "$subuidline" | cut -d: -f2)
+  subuidsize=$(echo "$subuidline" | cut -d: -f3)
+  if test -z "$subuidstart" || test -z "$subuidsize" ; then
+    skip "no subuid range for $(id -un) to set up the probe"
+  fi
+
+  # The namespaces below map inner gid 0 to our real egid, and inner gids
+  # 1..${subsize} to our subordinate gid range.  The innermost namespace, where
+  # the build runs, maps only our real uid/gid to 0, so:
+  #   okgid   -- inner 0 -> our real egid, which is also the RUN process's
+  #              primary gid, so the RUN process is a member of it.
+  #   notmine -- an inner gid from the subordinate range, which the build's
+  #              namespace does not map and the RUN process does not hold.
+  okgid=0
+  notmine=4242
+  if test "$notmine" -lt 1 || test "$notmine" -gt "$subsize" ; then
+    skip "gid $notmine is outside the mappable subordinate range 1..$subsize"
+  fi
+
+  # member_denied: group okgid, mode 0004 (owner ---, group ---, other r).  The
+  # RUN process is a member of the group, so the group bits must deny it even
+  # though "other" would allow.  This checks in_group_p()'s fsgid path.
+  echo secret_member > $probe/member_denied
+  # nonmember_allowed: group notmine, mode 0004.  The RUN process is not a
+  # member, so it falls through to "other" (r) and is allowed.  Control for the
+  # above: if both were denied, member_denied would prove nothing.
+  echo ok_nonmember > $probe/nonmember_allowed
+
+  # Set ownership and bits from a namespace that maps our subordinate ranges, so
+  # no sudo is needed.  Files are chown'ed to inner uid 1 (our subordinate
+  # range) so the RUN process is not their owner and cannot use CAP_DAC_OVERRIDE
+  # on them -- capable_wrt_inode_uidgid() requires the inode's uid to be mapped,
+  # and inner uid 1 is not mapped in the build's namespace.
+  unshare --user --mount \
+    --map-users=$(id -u),0,1 --map-users=${subuidstart},1,1 \
+    --map-groups=$(id -g),0,1 --map-groups=${substart},1,${subsize} \
+    sh -c "
+    set -e
+    chown 1:$okgid   $probe/member_denied     && chmod 0004 $probe/member_denied
+    chown 1:$notmine $probe/nonmember_allowed && chmod 0004 $probe/nonmember_allowed
+  " || skip "user namespace could not map the uid/gids for the probe"
+
+  cp ${COPY_BINARY} ${TEST_SCRATCH_DIR}/copy
+  cp ${BUILDAH_BINARY} ${TEST_SCRATCH_DIR}/buildah
+
+  echo set -e > ${TEST_SCRATCH_DIR}/script.sh
+  echo "mount -t tmpfs tmpfs /run/lock" >> ${TEST_SCRATCH_DIR}/script.sh
+  # setgroups-denied single-ID namespace -- same environment as #6947
+  echo "test \"\$(cat /proc/self/setgroups)\" = deny" >> ${TEST_SCRATCH_DIR}/script.sh
+  echo "${TEST_SCRATCH_DIR}/copy ${storageopts} dir:\$_BUILDAH_IMAGE_CACHEDIR/$baseimagef containers-storage:$baseimage" >> ${TEST_SCRATCH_DIR}/script.sh
+  echo "ctr=\$(${TEST_SCRATCH_DIR}/buildah ${BUILDAH_REGISTRY_OPTS} ${storageopts} from --pull=never -q $baseimage)" >> ${TEST_SCRATCH_DIR}/script.sh
+  echo "${TEST_SCRATCH_DIR}/buildah ${BUILDAH_REGISTRY_OPTS} ${storageopts} run --isolation chroot -v $probe:/probe:ro,z \"\$ctr\" -- sh -c 'echo setgroups:\$(cat /proc/self/setgroups); stat -c \"member_denied=%u:%g:%a\" /probe/member_denied; stat -c \"nonmember_allowed=%u:%g:%a\" /probe/nonmember_allowed; grep ^Gid: /proc/self/status; cat /probe/nonmember_allowed && ! cat /probe/member_denied'" >> ${TEST_SCRATCH_DIR}/script.sh
+
+  # Two namespaces, mirroring the reported environment: the outer one is mapped
+  # via newuidmap/newgidmap, which leaves setgroups() permitted, so we can drop
+  # our supplemental groups the way the container runtime would have; the inner
+  # one is the single-ID namespace, where the kernel requires setgroups to be
+  # denied before it will accept the mappings.
+  run unshare --user --mount \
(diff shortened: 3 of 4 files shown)


chroot/run_common.go
```diff
@@ -684,19 +684,26 @@ func runUsingChrootExecMain() {
...
-		gids := make([]int, len(user.AdditionalGids))
-		for i := range user.AdditionalGids {
-			gids[i] = int(user.AdditionalGids[i])
-		}
-		logrus.Debugf("setting supplemental groups")
-		if err = syscall.Setgroups(gids); err != nil {
-			fmt.Fprintf(os.Stderr, "error setting supplemental groups list: %v\n", err)
-			os.Exit(1)
+		if denied {
+			logrus.Debugf("ignoring supplemental groups because /proc/self/setgroups is set to deny")
```
nalind: No format specifiers, so this should ideally call `logrus.Debug()`, or log what's in the list.

chroot/run_linux.go
```diff
@@ -838,3 +838,16 @@ func setPdeathsig(cmd *exec.Cmd) {
 	}
 	cmd.SysProcAttr.Pdeathsig = syscall.SIGKILL
 }
+
+// setgroupsDenied reports whether this process's user namespace denies
+// setgroups(). Returns false on read error, so callers still attempt it.
+func setgroupsDenied() bool {
+	setgroupState, err := os.ReadFile("/proc/self/setgroups")
+	if err != nil {
+		if !os.IsNotExist(err) {
```
nalind: Get in the habit of using `errors.Is()` for these checks.  See the godoc for `os.IsNotExist()` for why we tend to avoid it and its family of functions in new code.

tests/chroot.bats
```diff
@@ -215,3 +215,101 @@ EOF
...
   env -i unshare -inCfpm bash ${TEST_SCRATCH_DIR}/script1
 }
+
+@test "chroot RUN succeeds when setgroups is denied" {
+  skip_if_no_unshare
+  if test `uname` != Linux ; then
+    skip "not meaningful except on Linux"
+  fi
+
+  baseimage=registry.access.redhat.com/ubi9/ubi-minimal:latest
```
nalind: Try to use one of the images cached by https://github.com/podman-container-tools/automation/blob/main/images/local-cache-registry#L58.  In CI, they can be pulled from a cache in the VM image, making the test more resilient against transient registry-side issues.

tests/chroot.bats
```diff
@@ -215,3 +215,101 @@ EOF
...
+  _prefetch $baseimage
+  baseimagef=$(tr -c a-zA-Z0-9.- - <<< "$baseimage")
+
+  storagedir=${TEST_SCRATCH_DIR}/storage ; mkdir $storagedir
+  rootdir=${storagedir}/rootdir           ; mkdir $rootdir
+  runrootdir=${storagedir}/runrootdir     ; mkdir $runrootdir
```
nalind: Spacing looks a little odd here.

tests/chroot.bats
```diff
@@ -215,3 +215,101 @@ EOF
...
+  echo "test \"\$(cat /proc/self/setgroups)\" = deny" >> ${TEST_SCRATCH_DIR}/script.sh
+  # seed the store from the prefetched cache so the build needs no network
+  echo "${TEST_SCRATCH_DIR}/copy ${storageopts} dir:\$_BUILDAH_IMAGE_CACHEDIR/$baseimagef containers-storage:$baseimage" >> ${TEST_SCRATCH_DIR}/script.sh
+  # the actual regression: this build used to fail at the RUN step with EPERM
+  echo "${TEST_SCRATCH_DIR}/buildah ${BUILDAH_REGISTRY_OPTS} ${storageopts} build --isolation chroot --pull=never $context" >> ${TEST_SCRATCH_DIR}/script.sh
+
+  run unshare --user --map-root-user --mount bash ${TEST_SCRATCH_DIR}/script.sh
+  echo "$output"
+  assert "$status" -eq 0
+  expect_output --substring "STEP 2/2: RUN true"
```
nalind: This line is logged if the build reaches this instruction, before it's executed.  Is that what it's meant to check for?

tests/chroot.bats
```diff
@@ -215,3 +215,101 @@ EOF
...
+    skip "not meaningful except on Linux"
+  fi
+
+  baseimage=registry.access.redhat.com/ubi9/ubi-minimal:latest
+  _prefetch $baseimage
+  baseimagef=$(tr -c a-zA-Z0-9.- - <<< "$baseimage")
+
+  storagedir=${TEST_SCRATCH_DIR}/storage ; mkdir $storagedir
+  rootdir=${storagedir}/rootdir           ; mkdir $rootdir
+  runrootdir=${storagedir}/runrootdir     ; mkdir $runrootdir
```
nalind: Spacing looks a little odd here.

tests/chroot.bats
```diff
@@ -215,3 +215,101 @@ EOF
...
+  # (our primary), which the RUN process legitimately holds -> readable.
+  echo ok_leak > $probe/allowed_leak
+
+  # chgrp inside a namespace that maps the full subgid range, so no sudo needed.
+  # allowed_leak -> gid 0 here (primary outside); denied_leak -> the leak gid
+  # that the build won't map.
+  unshare --user --map-root-user --map-groups=auto --mount sh -c "
+    set -e
+    chgrp 0 $probe/allowed_leak && chmod 0040 $probe/allowed_leak
+    chgrp $leakgid $probe/denied_leak && chmod 0040 $probe/denied_leak
```
nalind: Should these be `0004`, with the group owners reversed, or include a second pair of files to ensure that access can be denied to members of a specific group?

tests/chroot.bats
```diff
@@ -215,3 +215,101 @@ EOF
...
+  # read it. It must be denied.
+  echo secret_leak > $probe/denied_leak
+  # allowed_leak: group r (0040), other ---. Group maps to 0 inside the build
+  # (our primary), which the RUN process legitimately holds -> readable.
+  echo ok_leak > $probe/allowed_leak
+
+  # chgrp inside a namespace that maps the full subgid range, so no sudo needed.
+  # allowed_leak -> gid 0 here (primary outside); denied_leak -> the leak gid
+  # that the build won't map.
+  unshare --user --map-root-user --map-groups=auto --mount sh -c "
```
nalind: This might require a workaround for systems where `unshare` didn't understand `--map-groups=auto` (versions older than 2.38, I think).

ajoshua2004: ok, grabbed the real subgid range from `/etc/subgid` and used the older syntax mapping instead of `--map-groups=auto`

tests/chroot.bats
```diff
@@ -215,3 +215,184 @@ EOF
...
+  ${TEST_SCRATCH_DIR}/copy ${storageopts} dir:$_BUILDAH_IMAGE_CACHEDIR/$baseimagef containers-storage:$baseimage
+
+  # from + run --user, non-root member of okgid, NO single-ID wrapper
+  run_buildah from --pull=never ${storageopts} $baseimage
+  ctr="$output"
+  run_buildah run --isolation chroot ${storageopts} --user 1000:${okgid} \
+    -v $probe:/probe:ro,z "$ctr" -- \
+    sh -c 'cat /proc/self/status | grep ^Groups; cat /probe/nonmember_allowed && ! cat /probe/member_denied'
+  expect_output --substring "ok_nonmember"
+}
```
nalind: This doesn't appear to be verifying that the "setgroups" privilege is being denied to `buildah run`.

tests/chroot.bats
```diff
@@ -215,3 +215,184 @@ EOF
...
+  # allowed_leak -> gid 0 here (our primary); denied_leak -> the leak gid that
+  # the build won't map.
+  unshare --user --map-root-user --map-groups=${substart},0,${subsize} --mount sh -c "
+    set -e
+    chgrp 0 $probe/allowed_leak && chmod 0040 $probe/allowed_leak
+    chgrp $leakgid $probe/denied_leak && chmod 0040 $probe/denied_leak
+  " || skip "user namespace could not map the gid for the leak probe"
+
+  echo "FROM $baseimage" > $context/Dockerfile
+  echo 'RUN cat /proc/self/status | grep Groups; cat /probe/allowed_leak && ! cat /probe/denied_leak' >> $context/Dockerfile
```
nalind: Don't need to pipe output from `cat` through `grep` when `grep` can read things directly.
Not sure what we're supposed to do with the displayed Groups list - is the test expecting one or more values to be there, or not in there?

tests/chroot.bats
```diff
@@ -215,3 +215,184 @@ EOF
...
+  # seed the custom store from the prefetched cache so from --pull=never works
+  cp ${COPY_BINARY} ${TEST_SCRATCH_DIR}/copy
+  ${TEST_SCRATCH_DIR}/copy ${storageopts} dir:$_BUILDAH_IMAGE_CACHEDIR/$baseimagef containers-storage:$baseimage
+
+  # from + run --user, non-root member of okgid, NO single-ID wrapper
+  run_buildah from --pull=never ${storageopts} $baseimage
+  ctr="$output"
+  run_buildah run --isolation chroot ${storageopts} --user 1000:${okgid} \
+    -v $probe:/probe:ro,z "$ctr" -- \
+    sh -c 'cat /proc/self/status | grep ^Groups; cat /probe/nonmember_allowed && ! cat /probe/member_denied'
```
nalind: Don't need to pipe output from `cat` through `grep` when `grep` can read things directly.
Not sure what we're supposed to do with the displayed Groups list - is the test expecting one or more values to be there, or not in there?

tests/chroot.bats
```diff
@@ -215,3 +215,214 @@ EOF
...
+  subsize=$(echo "$subgidline" | cut -d: -f3)
+  if test -z "$substart" || test -z "$subsize" ; then
+    skip "no subgid range for $(id -un) to map for the leak probe"
+  fi
+
+  # An inside-gid the setup namespace can map (0..subsize-1) but that the
+  # build's single-ID namespace (maps only gid 0) will NOT map -- so inside the
+  # build it surfaces as overflow (65534): a supplemental group the process
+  # appears to hold but that must not count for access.
+  leakgid=4242
```
nalind: Nit: this should check that ${leakgid} will be within the range of GID mappings being set up below.

tests/chroot.bats
```diff
@@ -215,3 +215,214 @@ EOF
...
+  # read it. It must be denied.
+  echo secret_leak > $probe/denied_leak
+  # allowed_leak: group r (0040), other ---. Group maps to 0 inside the build
+  # (our primary), which the RUN process legitimately holds -> readable.
+  echo ok_leak > $probe/allowed_leak
+
+  # chgrp inside a namespace that maps our real subgid range, so no sudo needed.
+  # allowed_leak -> gid 0 here (our primary); denied_leak -> the leak gid that
+  # the build won't map.
+  unshare --user --map-root-user --map-groups=${substart},0,${subsize} --mount sh -c "
```
nalind: If `unshare` is actually mapping the first ID from that range to 0 as the `--map-groups` flag is requesting, then these two files are going to be owned by GIDs in the user's subordinate gid range, neither of which I think is expected to be mapped into the namespace in which the script is run below, so I would expect attempts to read from the two files to either both succeed or both fail.
The `--map-root-user` flag is documented as being equivalent to `--map-user=0 --map-group=0`, so `unshare` is effectively being given two flags that specify that different GIDs to map to 0.  What actually happens?

tests/chroot.bats
```diff
@@ -215,3 +215,214 @@ EOF
...
+  echo "${TEST_SCRATCH_DIR}/copy ${storageopts} dir:\$_BUILDAH_IMAGE_CACHEDIR/$baseimagef containers-storage:$baseimage" >> ${TEST_SCRATCH_DIR}/script.sh
+  echo "${TEST_SCRATCH_DIR}/buildah ${BUILDAH_REGISTRY_OPTS} ${storageopts} build --isolation chroot --pull=never -v $probe:/probe:ro,z $context" >> ${TEST_SCRATCH_DIR}/script.sh
+
+  run unshare --user --map-root-user --mount bash ${TEST_SCRATCH_DIR}/script.sh
+  echo "$output"
+  assert "$status" -eq 0
+  # the unmapped gid has to actually show up as overflow in the RUN process's
+  # groups list -- otherwise "denied" below would just mean it was never a
+  # member, and the test wouldn't be checking anything
+  expect_output --substring "Groups:	65534"
```
nalind: Which unmapped GID does this correspond to outside of the namespace that was created using `unshare`?  I don't think I understand what its presence in this list is expected to indicate.

tests/chroot.bats
```diff
@@ -215,3 +215,214 @@ EOF
...
+  subuidline=$(grep "^$(id -un):" /etc/subuid | head -1)
+  subuidstart=$(echo "$subuidline" | cut -d: -f2)
+  if test -z "$subuidstart" ; then
+    skip "no subuid range for $(id -un) to set up the probe"
+  fi
+
+  # In the single-ID namespace below, only gid 0 is mapped. notmine is an
+  # inside-gid the setup namespace can chgrp but the run namespace will not map.
+  okgid=0
+  notmine=4242
```
nalind, line 310: Nit: should check that this will be within the range of GID mappings created below.

tests/chroot.bats
```diff
@@ -215,3 +215,214 @@ EOF
...
+  echo "${TEST_SCRATCH_DIR}/buildah ${BUILDAH_REGISTRY_OPTS} ${storageopts} run --isolation chroot -v $probe:/probe:ro,z \"\$ctr\" -- sh -c 'test \"\$(cat /proc/self/setgroups)\" = deny; echo setgroups:\$(cat /proc/self/setgroups); grep ^Groups /proc/self/status; grep ^Gid: /proc/self/status; cat /probe/nonmember_allowed && ! cat /probe/member_denied'" >> ${TEST_SCRATCH_DIR}/script.sh
+
+  run unshare --user --map-root-user --mount bash ${TEST_SCRATCH_DIR}/script.sh
+  echo "$output"
+  assert "$status" -eq 0
+  # permission probes must run inside buildah run with setgroups denied
+  expect_output --substring "setgroups:deny"
+  # overflow gid must appear in supplemental groups -- otherwise nonmember_allowed
+  # would not be checking anything (same as the build leak test above).
+  expect_output --substring "Groups:	65534"
```
nalind: I don't think I understand what this check and its comment are saying.  Is this supposed to correspond to ${notmine}?  If not, to what?

tests/chroot.bats
```diff
@@ -215,3 +215,214 @@ EOF
...
+  echo secret_member > $probe/member_denied
+  # nonmember_allowed: group notmine, mode 0004. RUN process is NOT a member, so
+  # it falls through to "other" (r) and is allowed.
+  echo ok_nonmember > $probe/nonmember_allowed
+
+  # set groups+bits from a namespace mapping our subuid/subgid ranges, so no
+  # sudo needed. map uid 1 so chown 1:... is valid alongside map-root-user.
+  # chown to uid 1 so the RUN process (uid 0, without DAC override in this user
+  # namespace) is not the owner and group/other bits are actually consulted.
+  unshare --user --map-root-user --map-users=${subuidstart},1,1 --map-groups=${substart},0,${subsize} --mount sh -c "
```
nalind: `--map-groups`ing to a range starting at 0 should conflict with `--map-root-user`, which implies `--map-group=0`.  What is actually happening?

nalind (commented):
A couple of nits, and I don't think I understand what a couple of the tests are looking for when examining the supplemental groups list inside of a user namespace that doesn't have the subordinate IDs mapped into it.

chroot/run_common.go
```diff
@@ -684,19 +684,26 @@ func runUsingChrootExecMain() {
...
+			if err = syscall.Setgroups(gids); err != nil {
+				fmt.Fprintf(os.Stderr, "error setting supplemental groups list: %v\n", err)
+				os.Exit(1)
+			}
 		}
 	} else {
-		setgroups, _ := os.ReadFile("/proc/self/setgroups")
-		if strings.Trim(string(setgroups), "\n") != "deny" {
+		if !denied {
 			logrus.Debugf("clearing supplemental groups")
```
nalind, line 707: Nit: no log message for the case where `denied` is true?  Of the four combinations of allowed/denied,groups-in-the-list/no-groups-in-the-list, that's an outlier.

tests/chroot.bats
```diff
@@ -215,3 +215,156 @@ EOF
...
+  # the probes have to run inside buildah run, in the denied namespace
+  expect_output --substring "setgroups:deny"
+  # both files must be owned by an unmapped uid (65534), so neither the owner
+  # bits nor CAP_DAC_OVERRIDE apply, and both must still be mode 0004
+  expect_output --substring "member_denied=65534:${okgid}:4"
+  expect_output --substring "nonmember_allowed=65534:65534:4"
+  # okgid must be the RUN process's primary gid, otherwise member_denied is not
+  # exercising group-bit enforcement against a member
+  expect_output --substring "Gid:	${okgid}"
+  expect_output --substring "ok_nonmember"
```
nalind, line 369: It should be redundant to do so, but might want to check that the output doesn't include "secret_member".

nalind (approved):
LGTM with a couple of nits that wouldn't be blockers.

TomSweeneyRedHat:
LGTM

PR #6975, opened 2026-07-20
Fix --from with absolute paths to local image transports

### LINKED ISSUE #6932: buildah bud --from with absolute path to oci-archive fails
### Issue Description

After upgrading to the latest version of podman and its dependencies (podman 6.0.0-rc1, buildah 1.44.0), we noticed failures in some of our tests that were using `Containerfiles` in the following form:

```
FROM oci-archive:/absolute-path
```

It seems that using absolute paths like this is no longer supported since path sanitation was introduced in #5975. This makes sense for `Containefile` `FROM` directives, however the same failures also occur when using absolute paths with `podman build` or `buildah bud`:

```
podman build --from oci-archive:<absolute-path>
buildah bud --from oci-archive:<absolute-path>
```

When using `--from` on the command line, the user is explicitly providing the path, and there is no expectation that the build is self containing, so I expected the last command to work.

The error message itself (attached in the reproduction steps) is not super self explanatory, I believe what is happening is that the provided absolute path is being resolved as a relative path to the build context directory, leading to failures.

Please let me know if you need any more details!

```
[core@cosa-devsh ~]$ buildah --version
buildah version 1.44.0 (image-spec 1.1.1, runtime-spec 1.3.0)

[core@cosa-devsh ~]$ skopeo copy docker://quay.io/fedora/fedora-minimal:latest oci-archive:/tmp/fedora.ociarchive
Copying blob fb1251ea7033 done   | 
Copying config 8d41eb7fb8 done   | 

[core@cosa-devsh ~]$ cat > /tmp/Containerfile <<'EOF'
FROM overridden
RUN touch /etc/foo
EOF

[core@cosa-devsh ~]$ buildah bud --from oci-archive:/tmp/fedora.ociarchive -f /tmp/Containerfile
WARN[0000] expected to have archived a copy of "/tmp/fedora.ociarchive", missed it 
STEP 1/2: FROM oci-archive:/tmp/fedora.ociarchive
Error: creating build container: loading index: open /var/tmp/container_images_oci48477806/index.json: no such file or directory
```

Could not run `podman build --from` / `buildah bud --from` with absolute paths.

To be able to run `podman build --from` / `buildah bud --from` with absolute paths.

### buildah version output

```
[core@cosa-devsh ~]$ buildah --version
buildah version 1.44.0 (image-spec 1.1.1, runtime-spec 1.3.0)
```

### buildah info output

```yaml
{
    "host": {
        "Distribution": {
            "distribution": "fedora",
            "version": "45"
        },
        "MemFree": 105451520,
        "MemTotal": 989708288,
        "OCIRuntime": "crun",
        "SwapFree": 0,
        "SwapTotal": 0,
        "arch": "amd64",
        "cpus": 12,
        "hostname": "cosa-devsh",
        "kernel": "7.2.0-0.rc0.260622gef0c9f75a195.6.fc45.x86_64",
        "os": "linux",
        "rootless": true,
        "uptime": "7m 46.15s",
        "variant": ""
    },
    "store": {
        "ContainerStore": {
            "number": 1
        },
        "GraphDriverName": "overlay",
        "GraphImageStore": "",
        "GraphOptions": [
            "overlay.mountopt=nodev,metacopy=on"
        ],
        "GraphRoot": "/var/home/core/.local/share/containers/storage",
        "GraphStatus": {
            "Backing Filesystem": "xfs",
            "Supports d_type": "true",
            "Supports shifting": "false",
            "Supports volatile": "true",
            "Using metacopy": "false"
        },
        "GraphTransientStore": false,
        "ImageStore": {
            "number": 2
        },
        "RunRoot": "/run/user/1000/containers"
    }
}
```

### Provide your storage.conf

```
[core@cosa-devsh ~]$ cat /usr/share/containers/storage.conf
# files.
#
[storage]

[storage.options]

[storage.options.overlay]
mountopt = "nodev"

[storage.options.vfs]

[core@cosa-devsh ~]$ cat /etc/containers/storage.conf
```

### Upstream Latest Release

Yes

The above commands were run using a Fedora CoreOS 45 rawhide qcow2 image in QEMU.

### Additional information

_No response_

Honny1:
Fixes: https://github.com/podman-container-tools/buildah/issues/6932

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
Fixed `buildah build --from` with absolute paths to oci-archive, docker-archive, oci, and dir transports.
```

### DIFF
--- a/imagebuildah/stage_executor.go
+++ b/imagebuildah/stage_executor.go
@@ -983,7 +983,7 @@ func (s *stageExecutor) UnrecognizedInstruction(step *imagebuilder.Step) error {
 // copy of the original image, under "tmpdir", which contains no symbolic
 // links, and return either the original image reference or a reference to a
 // sanitized copy which should be used instead.
-func (s *stageExecutor) sanitizeFrom(from, tmpdir string) (newFrom string, err error) {
+func (s *stageExecutor) sanitizeFrom(from, tmpdir string, allowAbsolutePaths bool) (newFrom string, err error) {
 	transportName, restOfImageName, maybeHasTransportName := strings.Cut(from, ":")
 	if !maybeHasTransportName || transports.Get(transportName) == nil {
 		if _, err = reference.ParseNormalizedNamed(from); err == nil {
@@ -997,7 +997,7 @@ func (s *stageExecutor) sanitizeFrom(from, tmpdir string) (newFrom string, err e
 		return "", fmt.Errorf("parsing image name %q: %w", from, err)
 	}
 	// TODO: drop this part and just return an error... someday
-	return sanitize.ImageName(s.executor.store, transportName, restOfImageName, s.executor.contextDir, tmpdir)
+	return sanitize.ImageName(s.executor.store, transportName, restOfImageName, s.executor.contextDir, tmpdir, allowAbsolutePaths)
 }
 
 // prepare creates a working container based on the specified image, or if one
@@ -1062,7 +1062,8 @@ func (s *stageExecutor) prepare(ctx context.Context, from string, initializeIBCo
 		}
 	}
 
-	sanitizedFrom, err := s.sanitizeFrom(from, tmpdir.GetTempDir())
+	allowAbsolutePaths := s.executor.fromOverrideStage == s.index
+	sanitizedFrom, err := s.sanitizeFrom(from, tmpdir.GetTempDir(), allowAbsolutePaths)
 	if err != nil {
 		return nil, fmt.Errorf("invalid base image specification %q: %w", from, err)
 	}
--- a/internal/sanitize/sanitize.go
+++ b/internal/sanitize/sanitize.go
@@ -183,9 +183,14 @@ func writeToDirectory(root string, hdr *tar.Header, content io.Reader) error {
 // ImageName limits which image transports we'll accept.  For those it accepts
 // which refer to filesystem objects, where relative path names are evaluated
 // relative to "contextDir", it will create a copy of the original image, under
-// "tmpdir", which contains no symbolic links.  It it returns a parseable
+// "tmpdir", which contains no symbolic links.  It returns a parseable
 // reference to the image which should be used.
-func ImageName(store storage.Store, transportName, restOfImageName, contextDir, tmpdir string) (newFrom string, err error) {
+//
+// When allowAbsolutePaths is true and the image path is absolute, it is read
+// directly from the host filesystem (e.g. for paths supplied via --from).
+// Archive contents are still sanitized; directory-based transports are returned
+// as-is since they are read directly by the image library.
+func ImageName(store storage.Store, transportName, restOfImageName, contextDir, tmpdir string, allowAbsolutePaths bool) (newFrom string, err error) {
 	seenEntries := make(map[string]struct{})
 	// we're going to try to create a temporary directory or file, but if
 	// we fail, make sure that they get removed immediately
@@ -241,9 +246,17 @@ func ImageName(store storage.Store, transportName, restOfImageName, contextDir, 
 				}
 			}
 		}()
-		// archive only the archive file for copying to the new archive file
-		imageArchive, err = newSingleItemArchive(contextDir, archiveSource)
-		isEmbeddedArchive = true
+		if allowAbsolutePaths && filepath.IsAbs(archiveSource) {
+			var rawFile *os.File
+			rawFile, err = os.Open(archiveSource)
+			if err == nil {
+				defer rawFile.Close()
+				imageArchive, _, err = compression.AutoDecompress(rawFile)
+			}
+		} else {
+			imageArchive, err = newSingleItemArchive(contextDir, archiveSource)
+			isEmbeddedArchive = true
+		}
 		// generate the new reference using the temporary file's name
 		newFrom = transportName + ":" + newImageDestination + refLeftover
 	case ociLayoutTransport.Transport.Name(): // this is a directory tree
@@ -253,6 +266,10 @@ func ImageName(store storage.Store, transportName, restOfImageName, contextDir, 
 		archiveSource = as
 		if ok {
 			refLeftover = ":" + refLeftover
+		}
+		if allowAbsolutePaths && filepath.IsAbs(archiveSource) {
+			succeeded = true
+			return transportName + ":" + archiveSource + refLeftover, nil
 		}
 		// create a new directory to use as our new layout directory
 		if newImageDestination, err = newDirectoryDestination(tmpdir); err != nil {
@@ -266,6 +283,10 @@ func ImageName(store storage.Store, transportName, restOfImageName, contextDir, 
 	case directoryTransport.Transport.Name(): // this is also a directory tree
 		// this takes the form of just a path
 		transportRef := restOfImageName
+		if allowAbsolutePaths && filepath.IsAbs(transportRef) {
+			succeeded = true
+			return transportName + ":" + transportRef, nil
+		}
 		// create a new directory to use as our new image directory
 		if newImageDestination, err = newDirectoryDestination(tmpdir); err != nil {
 			return "", fmt.Errorf("creating temporary copy of base image: %w", err)
@@ -280,7 +301,7 @@ func ImageName(store storage.Store, transportName, restOfImageName, contextDir, 
 		return "", fmt.Errorf("unexpected container image transport %q", transportName)
 	}
 	if err != nil {
-		return "", fmt.Errorf("error archiving source at %q under %q", archiveSource, contextDir)
+		return "", fmt.Errorf("error archiving source at %q under %q: %w", archiveSource, contextDir, err)
 	}
 
 	// start reading the archived content
--- a/tests/bud.bats
+++ b/tests/bud.bats
@@ -10752,3 +10752,45 @@ _EOF
     done
   done
 }
+
+@test "bud --from with absolute path to local transport" {
+  _prefetch busybox
+
+  local buildcontext=${TEST_SCRATCH_DIR}/buildcontext
+  mkdir -p ${buildcontext}
+
+  # Create a working container and get its image ID for comparison
+  run_buildah from -q $WITH_POLICY_JSON busybox
+  local cid=$output
+  run_buildah inspect --format '{{.FromImageID}}' "$cid"
+  local expected_id=$output
+
+  # Create images outside build context using all local transports
+  local imgdir=${TEST_SCRATCH_DIR}/images
+  mkdir -p ${imgdir}
+
+  run_buildah push $WITH_POLICY_JSON busybox oci-archive:${imgdir}/test.ociarchive
+  run_buildah push $WITH_POLICY_JSON busybox docker-archive:${imgdir}/test.dockerarchive
+  run_buildah push $WITH_POLICY_JSON busybox oci:${imgdir}/test-oci
+  run_buildah push $WITH_POLICY_JSON busybox dir:${imgdir}/test-dir
+
+  # Multi-stage Containerfile: --from should only override the first FROM.
+  # The second stage must still resolve "busybox" normally, proving that
+  # allowAbsolutePaths does not leak to later stages.
+  cat > ${buildcontext}/Containerfile << _EOF
+FROM placeholder-to-be-overridden AS stage0
+RUN echo stage0 > /stage.txt
+FROM busybox AS stage1
+COPY --from=stage0 /stage.txt /from-stage0.txt
+RUN cat /from-stage0.txt
+_EOF
+
+  for transport in \
+    "oci-archive:${imgdir}/test.ociarchive" \
+    "docker-archive:${imgdir}/test.dockerarchive" \
+    "oci:${imgdir}/test-oci" \
+    "dir:${imgdir}/test-dir" ; do
+    run_buildah build $WITH_POLICY_JSON --from "$transport" ${buildcontext}
+    expect_output --substring "FROM busybox AS stage1"
+  done
+}
--- a/imagebuildah/executor.go
+++ b/imagebuildah/executor.go
@@ -168,6 +168,7 @@ type executor struct {
 	imageInfoLock                           sync.Mutex
 	imageInfoCache                          map[string]imageTypeAndHistoryAndDiffIDs
 	fromOverride                            string
+	fromOverrideStage                       int // stage position that received --from, or -1 if none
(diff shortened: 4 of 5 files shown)


tests/bud.bats
```diff
@@ -10624,3 +10624,56 @@ _EOF
...
+  _prefetch busybox
+  local archive_dir=${TEST_SCRATCH_DIR}/archives
+  mkdir -p "${archive_dir}"
+
+  copy containers-storage:busybox oci-archive:"${archive_dir}"/busybox.ociarchive
+
+  # Absolute path in a Containerfile FROM should also work when
+  # the archive lives outside of the build context
+  cat > ${TEST_SCRATCH_DIR}/Containerfile << EOF
+FROM oci-archive:${archive_dir}/busybox.ociarchive
```
nalind: This location is within the build context directory.

tests/bud.bats
```diff
@@ -10624,3 +10624,56 @@ _EOF
...
+EOF
+
+  for from in \
+    oci-archive:"${archive_dir}"/busybox.ociarchive \
+    docker-archive:"${archive_dir}"/busybox.dockerarchive \
+    oci:"${archive_dir}"/busybox-oci \
+    dir:"${archive_dir}"/busybox-dir \
+  ; do
+    run_buildah build $WITH_POLICY_JSON \
+      --from "$from" \
```
nalind: These values for "$from" are all below the context directory, which should already be accepted if the location is specified as though the context directory were the root directory.

internal/sanitize/sanitize.go
```diff
@@ -253,7 +261,15 @@ func ImageName(transportName, restOfImageName, contextDir, tmpdir string) (newFr
 		}
 		// archive the entire layout directory for copying to the new layout directory
 		tarOptions := &archive.TarOptions{}
-		imageArchive, err = chrootarchive.Tar(filepath.Join(contextDir, archiveSource), tarOptions, contextDir)
+		if filepath.IsAbs(archiveSource) {
+			if _, statErr := os.Stat(archiveSource); statErr == nil {
+				imageArchive, err = chrootarchive.Tar(archiveSource, tarOptions, filepath.Dir(archiveSource))
```
nalind: This bypasses the attempt to ensure that we're only reading content from the context directory for all cases, not only those where "--from" was used to provide the value that we're processing.

internal/sanitize/sanitize.go
```diff
@@ -235,7 +235,15 @@ func ImageName(transportName, restOfImageName, contextDir, tmpdir string) (newFr
 			}
 		}()
 		// archive only the archive file for copying to the new archive file
-		imageArchive, err = newSingleItemArchive(contextDir, archiveSource)
+		if filepath.IsAbs(archiveSource) {
+			if _, statErr := os.Stat(archiveSource); statErr == nil {
+				imageArchive, err = newSingleItemArchive(filepath.Dir(archiveSource), filepath.Base(archiveSource))
```
nalind: This bypasses the attempt to ensure that we're only reading content from the context directory for all cases, not only those where "--from" was used to provide the value that we're processing.

internal/sanitize/sanitize.go
```diff
@@ -266,7 +282,15 @@ func ImageName(transportName, restOfImageName, contextDir, tmpdir string) (newFr
 		// archive the entire directory for copying to the new directory
 		archiveSource = transportRef
 		tarOptions := &archive.TarOptions{}
-		imageArchive, err = chrootarchive.Tar(filepath.Join(contextDir, archiveSource), tarOptions, contextDir)
+		if filepath.IsAbs(archiveSource) {
+			if _, statErr := os.Stat(archiveSource); statErr == nil {
+				imageArchive, err = chrootarchive.Tar(archiveSource, tarOptions, filepath.Dir(archiveSource))
```
nalind: This bypasses the attempt to ensure that we're only reading content from the context directory for all cases, not only those where "--from" was used to provide the value that we're processing.

nalind (changes requested):
This needs to be much more careful about attempting to read data from the filesystem that isn't somewhere within the context directory.

imagebuildah/stage_executor.go
```diff
@@ -1017,6 +1018,10 @@ func (s *stageExecutor) prepare(ctx context.Context, from string, initializeIBCo
 		from = base
 	}
 
+	// Decide before source policy may rewrite from: absolute paths are only
+	// allowed for the value supplied via --from, not Containerfile FROM lines.
+	allowAbsolutePaths := s.executor.originalFromOverride != "" && from == s.executor.originalFromOverride
```
nalind: This evaluates as true for every stage of a multi-stage build that uses the same base as the first stage.  It isn't clear from the commit log if that's intentional.

internal/sanitize/sanitize.go
```diff
@@ -242,7 +244,11 @@ func ImageName(store storage.Store, transportName, restOfImageName, contextDir,
 			}
 		}()
 		// archive only the archive file for copying to the new archive file
-		imageArchive, err = newSingleItemArchive(contextDir, archiveSource)
+		if allowAbsolutePaths && filepath.IsAbs(archiveSource) {
+			imageArchive, err = newSingleItemArchive(filepath.Dir(archiveSource), filepath.Base(archiveSource))
```
nalind: Writing out a new copy seems unnecessary.  Passing back a reader for the original file and setting `isEmbeddedArchive` false for this case would provide the same data.

internal/sanitize/sanitize.go
```diff
@@ -260,27 +266,41 @@ func ImageName(store storage.Store, transportName, restOfImageName, contextDir,
 		}
 		// archive the entire layout directory for copying to the new layout directory
 		tarOptions := &archive.TarOptions{}
-		imageArchive, err = chrootarchive.Tar(filepath.Join(contextDir, archiveSource), tarOptions, contextDir)
+		if allowAbsolutePaths && filepath.IsAbs(archiveSource) {
+			imageArchive, err = chrootarchive.Tar(archiveSource, tarOptions, filepath.Dir(archiveSource))
```
nalind: Is the intent here to disrupt links to items outside of the context directory when absolute paths are being allowed?  If it is not, then we don't need to be calling this function.

internal/sanitize/sanitize.go
```diff
@@ -260,27 +266,41 @@ func ImageName(store storage.Store, transportName, restOfImageName, contextDir,
...
+		} else {
+			// Leading "/" is context-absolute, not a host path; avoid filepath.Join
+			// dropping contextDir when archiveSource is absolute.
```
nalind: I'm not sure that filepath.Join() does that?

internal/sanitize/sanitize.go
```diff
@@ -260,27 +266,41 @@ func ImageName(store storage.Store, transportName, restOfImageName, contextDir,
...
+			imageArchive, err = chrootarchive.Tar(filepath.Join(contextDir, contextRelativePath(archiveSource)), tarOptions, contextDir)
+		}
 		// generate the new reference using the directory
 		newFrom = transportName + ":" + newImageDestination + refLeftover
 	case directoryTransport.Transport.Name(): // this is also a directory tree
 		// this takes the form of just a path
 		transportRef := restOfImageName
+		archiveSource = transportRef
```
nalind: Moved for stylistic reasons?

internal/sanitize/sanitize.go
```diff
@@ -260,27 +266,41 @@ func ImageName(store storage.Store, transportName, restOfImageName, contextDir,
...
 	default:
 		return "", fmt.Errorf("unexpected container image transport %q", transportName)
 	}
 	if err != nil {
-		return "", fmt.Errorf("error archiving source at %q under %q", archiveSource, contextDir)
+		archiveRoot := contextDir
+		if allowAbsolutePaths && filepath.IsAbs(archiveSource) {
+			archiveRoot = filepath.Dir(archiveSource)
+		}
+		return "", fmt.Errorf("error archiving source at %q under %q", archiveSource, archiveRoot)
```
nalind: Should this error message wrap `err`?

tests/bud.bats
```diff
@@ -10640,3 +10640,68 @@ _EOF
...
+  copy containers-storage:busybox docker-archive:"${archive_dir}"/busybox.dockerarchive
+  copy containers-storage:busybox oci:"${archive_dir}"/busybox-oci
+  copy containers-storage:busybox dir:"${archive_dir}"/busybox-dir
+
+  local contextdir=${TEST_SCRATCH_DIR}/context
+  mkdir -p "${contextdir}"
+  cat > "${contextdir}"/Containerfile << 'EOF'
+FROM overridden
+RUN touch /absolute-path-test
+EOF
```
nalind, line 10703: We expect `--from` to only replace the value referenced in the first FROM instruction, but this doesn't check that for the types of values that we're going to start handling specially here.

Honny1:

internal/sanitize/sanitize.go
```diff
@@ -260,27 +267,42 @@ func ImageName(store storage.Store, transportName, restOfImageName, contextDir,
 		}
 		// archive the entire layout directory for copying to the new layout directory
 		tarOptions := &archive.TarOptions{}
-		imageArchive, err = chrootarchive.Tar(filepath.Join(contextDir, archiveSource), tarOptions, contextDir)
+		if allowAbsolutePaths && filepath.IsAbs(archiveSource) {
+			// Sanitize a host path (--from): copy the tree and reject symlinks
+			// that point outside of it.
+			imageArchive, err = chrootarchive.Tar(archiveSource, tarOptions, filepath.Dir(archiveSource))
```
TomSweeneyRedHat: I think you don't want the filepath dir, as you might get into problems with a symlink

```suggestion
			imageArchive, err = chrootarchive.Tar(archiveSource, tarOptions, archiveSource)
```

internal/sanitize/sanitize.go
```diff
@@ -260,27 +267,42 @@ func ImageName(store storage.Store, transportName, restOfImageName, contextDir,
...
 		// create a new directory to use as our new image directory
 		if newImageDestination, err = newDirectoryDestination(tmpdir); err != nil {
 			return "", fmt.Errorf("creating temporary copy of base image: %w", err)
 		}
 		// archive the entire directory for copying to the new directory
-		archiveSource = transportRef
 		tarOptions := &archive.TarOptions{}
-		imageArchive, err = chrootarchive.Tar(filepath.Join(contextDir, archiveSource), tarOptions, contextDir)
+		if allowAbsolutePaths && filepath.IsAbs(archiveSource) {
+			imageArchive, err = chrootarchive.Tar(archiveSource, tarOptions, filepath.Dir(archiveSource))
```
TomSweeneyRedHat: if you change above, change here too
```suggestion
			imageArchive, err = chrootarchive.Tar(archiveSource, tarOptions, archiveSource)
```

internal/sanitize/sanitize.go
```diff
@@ -260,27 +267,42 @@ func ImageName(store storage.Store, transportName, restOfImageName, contextDir,
...
 		// generate the new reference using the directory
 		newFrom = transportName + ":" + newImageDestination
 	default:
 		return "", fmt.Errorf("unexpected container image transport %q", transportName)
 	}
 	if err != nil {
-		return "", fmt.Errorf("error archiving source at %q under %q", archiveSource, contextDir)
+		archiveRoot := contextDir
+		if allowAbsolutePaths && filepath.IsAbs(archiveSource) {
+			archiveRoot = filepath.Dir(archiveSource)
```
TomSweeneyRedHat: And if above, here too.

```suggestion
			archiveRoot = archiveSource
```

imagebuildah/stage_executor.go
```diff
@@ -1062,7 +1062,9 @@ func (s *stageExecutor) prepare(ctx context.Context, from string, initializeIBCo
 		}
 	}
 
-	sanitizedFrom, err := s.sanitizeFrom(from, tmpdir.GetTempDir())
+	allowAbsolutePaths := s.executor.fromOverrideApplied
+	s.executor.fromOverrideApplied = false
```
nalind: Given that stages can be run in any order if there's no dependency between them, this looks like it would be a race.

tests/bud.bats
```diff
@@ -10752,3 +10752,47 @@ _EOF
...
+
+  for transport in \
+    "oci-archive:${imgdir}/test.ociarchive" \
+    "docker-archive:${imgdir}/test.dockerarchive" \
+    "oci:${imgdir}/test-oci" \
+    "dir:${imgdir}/test-dir" ; do
+    run_buildah build $WITH_POLICY_JSON --from "$transport" -t testimg ${buildcontext}
+    run_buildah inspect --format '{{.FromImage}}' testimg
+    expect_output --substring "busybox"
+    run_buildah rmi testimg
```
nalind: There's no harm done if this isn't removed during the test.

imagebuildah/stage_executor.go
Honny1: Good catch.

Honny1:
PTAL @nalind @TomSweeneyRedHat @simonbrauner

internal/sanitize/sanitize.go
```diff
@@ -183,9 +183,14 @@ func writeToDirectory(root string, hdr *tar.Header, content io.Reader) error {
...
 // relative to "contextDir", it will create a copy of the original image, under
-// "tmpdir", which contains no symbolic links.  It it returns a parseable
+// "tmpdir", which contains no symbolic links.  It returns a parseable
 // reference to the image which should be used.
-func ImageName(store storage.Store, transportName, restOfImageName, contextDir, tmpdir string) (newFrom string, err error) {
+//
+// When allowAbsolutePaths is true and the image path is absolute, it is read
+// directly from the host filesystem (e.g. for paths supplied via --from).
+// Archive contents are still sanitized; directory-based transports are returned
+// as-is since they are read directly by the image library.
```
simonbrauner, line 192: Is the distinguishing factor for sanitization truly whether the contents are `read directly by the image library`? Shouldn't it be something in the sense of `the image is extracted on the file system already and the user points to this location, so we consider it to be trusted`?

As of this comment https://github.com/Honny1/buildah/blob/e0375bbbde3c6b6ac4263a081c988c139a07daf1/internal/sanitize/sanitize.go#L100, `docker-archive` also gets `read directly`, but the sanitization applies to it.

PR #6976, opened 2026-07-20
Copier: add CreateDestPath option to PutOptions

ajoshua2004:
#### What type of PR is this?

/kind api-change

Adds `CreateDestPath` to `copier.PutOptions`, letting callers enable or disable the implicit creation of the destination directory when it doesn't already exist. Default behavior is unchanged: destination is created unless `CreateDestPath` is explicitly set to `OptionalBoolFalse`. Will throw an error if `CreateDestPath` = False and the directory does not exist. 

#### How to verify it
`go test ./copier -run TestPutCreateDestPath -v`

Part of #6732

None

```release-note
None
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -444,24 +444,25 @@ func Get(root string, directory string, options GetOptions, globs []string, bulk
 
 // PutOptions controls parts of Put()'s behavior.
 type PutOptions struct {
-	UIDMap, GIDMap       []idtools.IDMap   // map from containerIDs to hostIDs when writing contents to disk
-	DefaultDirOwner      *idtools.IDPair   // set ownership of implicitly-created directories, default is ChownDirs, or 0:0 if ChownDirs not set
-	DefaultDirMode       *os.FileMode      // set permissions on implicitly-created directories, default is Chmod or ChmodDirs, or 0755 if neither is set
-	Chmod                string            // set permissions in octal or symbolic notation. overrides ChmodDirs and ChmodFiles if set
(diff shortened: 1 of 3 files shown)


copier/copier.go
```diff
@@ -23,7 +23,7 @@ import (
 	"unicode"
 
 	"github.com/sirupsen/logrus"
-	"github.com/tonistiigi/dchapes-mode"
+	mode "github.com/tonistiigi/dchapes-mode"
```
TomSweeneyRedHat: It might be end-of-day confusion, but I'm not seeing any changes due to this addition of the `mode` alias, and I expected to.  Am I missing something?

copier/copier_test.go
```diff
@@ -3199,3 +3199,60 @@ func testPutTimestamp(t *testing.T) {
 		assert.Equal(t, originalTime.Unix(), info.ModTime().Unix())
 	})
 }
+
+func TestPutCreateDestPathNoChroot(t *testing.T) {
```
nalind, line 3203: This should probably get a corresponding `TestPutCreateDestPathChroot()` in copier_unix_test.go

nalind (commented):
Needs another unit test for the "actually able to use chroot()" cases, otherwise LGTM

copier/copier.go
ajoshua2004: This was just my IDE adding it after I saved the file, removed it.

nalind (approved):
LGTM

TomSweeneyRedHat:
LGTM
Thanks @ajoshua2004 !

PR #6984, opened 2026-07-22
Makefile: add install.nobuild target to install ONLY.

rahilarious:
Current behaviour is `make install` builds binary as well as installs it. This stays the same. But in scenario when one just want to install and not compile, there was no option but to patch source.

Adding new target accomodates such scenario.
Idea from @lsm5 https://github.com/podman-container-tools/skopeo/pull/2172#issuecomment-1838559541

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,11 +165,14 @@ install.go-md2man:
 install.golangci-lint:
 	$(MAKE) -C tests/tools build/golangci-lint
 
-.PHONY: install
-install: bin/buildah
+.PHONY: install.nobuild
+install.nobuild:
 	install -d -m 755 $(DESTDIR)/$(BINDIR)
 	install -m 755 bin/buildah $(DESTDIR)/$(BINDIR)/buildah
 	$(MAKE) -C docs install
+
+.PHONY: install
+install: bin/buildah install.nobuild
 
 .PHONY: uninstall
 uninstall:


Honny1 (approved):
LGTM

lsm5:
> But in scenario when one just want to install and not compile

@rahilarious Are there cases where this is needed? For example, the rpm's install stage doesn't rebuild the binary anymore. See https://github.com/podman-container-tools/buildah/pull/6785#pullrequestreview-4121014009 . Would that not apply here too?

rahilarious:
> > But in scenario when one just want to install and not compile
> 
> @rahilarious Are there cases where this is needed? For example, the rpm's install stage doesn't rebuild the binary anymore. See [#6785 (review)](https://github.com/podman-container-tools/buildah/pull/6785#pullrequestreview-4121014009) . Would that not apply here too?

yes. It doesn't recompile but in Gentoo running executing build commands (i.e. `go build`) itself is prohibited in installation phase. So I had to patch source. But with this PR merged, I can just `make install.nobuild` & make patch redundant. 

I would've preferrred to just remove `bin/buildah` from `install:` like we do in Podman but, to not _break_ anything in minor update, introduced new target

lsm5 (approved):
if gentoo needs this to prevent patching, fine by me. I'll defer to @nalind on this though.

TomSweeneyRedHat:
LGTM
@nalind WDYT?

nalind:
I'm kind of curious about where the recent conversation in https://github.com/podman-container-tools/skopeo/pull/2173 will land.

rahilarious:
> I'm kind of curious about where the recent conversation in [podman-container-tools/skopeo#2173](https://github.com/podman-container-tools/skopeo/pull/2173) will land.

It's closed now. But that shouldn't block this as the PR doesn't break anything & facilitates another option to install

nalind:
Indeed.  LGTM, merging.

PR #6988, opened 2026-07-22
Added --include flag to buildah add and buildah copy

ajoshua2004:
#### What type of PR is this?

> /kind feature

Adds an `--include` flag to both add and copy cli commands. This is the complement to `--exclude` .
Its purpose is to only copy files with the corresponding pattern. Excludes will take priority when both match

This feature is not usable in Dockerfiles and Containerfiles 

#### How to verify it
`bats --filter "add --include" tests/add.bats`
`bats --filter "copy --include" tests/copy.bats`
Part of #6732 

This is the precursor to the RequiredPaths flag that will be incoming.

```release-note
`buildah copy --include` and `buildah add --include ` is the complement to the `buildah copy --exclude` and `buildah add --exclude` commands
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -198,6 +198,15 @@ func (req *request) Excludes() []string {
 	}
 }
 
+func (req *request) Includes() []string {
+	switch req.Request {
+	case requestGet:
+		return req.GetOptions.Includes
+	default:
+		return nil
+	}
+}
+
 func (req *request) UIDMap() []idtools.IDMap {
 	switch req.Request {
 	case requestEval:
@@ -405,6 +414,7 @@ type GetOptions struct {
 	Timestamp          *time.Time        // timestamp to force on all contents
 	DisallowWildcard   bool              // reject glob patterns in source paths
 	AllowEmptyWildcard bool              // don't error when glob patterns match nothing
+	Includes           []string          // include only contents matching at least one of these patterns; Excludes take precedence
 }
 
 // Get produces an archive containing items that match the specified glob
@@ -1072,9 +1082,17 @@ func copierHandler(bulkReader io.Reader, bulkWriter io.Writer, req request) (*re
 	// os.PathSeparator, implying that it expects OS-specific naming
 	// conventions.
 	excludes := req.Excludes()
-	pm, err := fileutils.NewPatternMatcher(excludes)
+	pmExcludes, err := fileutils.NewPatternMatcher(excludes)
 	if err != nil {
 		return nil, nil, fmt.Errorf("processing excludes list %v: %w", excludes, err)
+	}
+
+	var pmIncludes *fileutils.PatternMatcher
+	if includes := req.Includes(); len(includes) > 0 {
+		pmIncludes, err = fileutils.NewPatternMatcher(includes)
+		if err != nil {
+			return nil, nil, fmt.Errorf("processing includes list %v: %w", includes, err)
+		}
 	}
 
 	var idMappings *idtools.IDMappings
@@ -1090,10 +1108,10 @@ func copierHandler(bulkReader io.Reader, bulkWriter io.Writer, req request) (*re
 		resp := copierHandlerEval(req)
 		return resp, nil, nil
 	case requestStat:
-		resp := copierHandlerStat(req, pm, idMappings)
+		resp := copierHandlerStat(req, pmExcludes, idMappings)
 		return resp, nil, nil
 	case requestGet:
-		return copierHandlerGet(bulkWriter, req, pm, idMappings)
+		return copierHandlerGet(bulkWriter, req, pmExcludes, pmIncludes, idMappings)
 	case requestPut:
 		return copierHandlerPut(bulkReader, req, idMappings)
 	case requestMkdir:
@@ -1112,10 +1130,11 @@ func copierHandler(bulkReader io.Reader, bulkWriter io.Writer, req request) (*re
 	}
 }
 
-// pathIsExcluded computes path relative to root, then asks the pattern matcher
-// if the result is excluded.  Returns the relative path and the matcher's
-// results.
-func pathIsExcluded(root, path string, pm *fileutils.PatternMatcher) (string, bool, error) {
+// pathMatches computes path relative to root, then asks the pattern matcher
+// if the result matches.  Returns the relative path and the matcher's
+// results. Callers treat a match as "excluded" or "included" depending on
+// whether pm is the excludes or includes pattern matcher.
+func pathMatches(root, path string, pm *fileutils.PatternMatcher) (string, bool, error) {
 	rel, err := convertToRelSubdirectory(root, path)
 	if err != nil {
 		return "", false, fmt.Errorf("copier: error computing path of %q relative to root %q: %w", path, root, err)
@@ -1158,7 +1177,7 @@ func resolvePath(root, path string, evaluateFinalComponent bool, pm *fileutils.P
 	excluded := false
 	for len(components) > 0 {
 		// if anything we try to examine is excluded, then resolution has to "break"
-		_, thisExcluded, err := pathIsExcluded(root, filepath.Join(workingPath, components[0]), pm)
+		_, thisExcluded, err := pathMatches(root, filepath.Join(workingPath, components[0]), pm)
 		if err != nil {
 			return "", err
 		}
@@ -1221,7 +1240,7 @@ func containsWildcards(path string) bool {
 	return strings.ContainsAny(path, "*?[")
 }
 
-func copierHandlerStat(req request, pm *fileutils.PatternMatcher, idMappings *idtools.IDMappings) *response {
+func copierHandlerStat(req request, pmExcludes *fileutils.PatternMatcher, idMappings *idtools.IDMappings) *response {
 	errorResponse := func(fmtspec string, args ...any) *response {
 		return &response{Error: fmt.Sprintf(fmtspec, args...), Stat: statResponse{}}
 	}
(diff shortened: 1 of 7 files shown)


tests/add.bats
```diff
@@ -668,3 +668,41 @@ EOF
...
+
+expect="
+stuff
+stuff/source.go
+stuff/subdir
+stuff/subdir/nested.go"
+
+  run_buildah from $WITH_POLICY_JSON scratch
+  cid=$output
+  run_buildah add --include=**/*.go $cid ${mytest} /stuff
```
nalind: Wildcard patterns should be quoted in scripts so that the shell doesn't try to expand them.

docs/buildah-add.1.md
```diff
@@ -78,6 +78,12 @@ can be used.
 
 Path to an alternative .containerignore (.dockerignore) file. Requires \-\-contextdir be specified.
 
+**--include** *pattern*
+
+Only copy files matching the specific pattern, the complement of **--exclude**. Option can be specified
```
TomSweeneyRedHat:
```suggestion
Only copy files matching the specific pattern, the complement of **--exclude**. The option can be specified
```
I know you cut/pasted from exclude, but please fix it there too.

docs/buildah-add.1.md
```diff
@@ -78,6 +78,12 @@ can be used.
...
+multiple times. Patterns are matched against each file's path relaltive to the source directory being
```
TomSweeneyRedHat:
```suggestion
multiple times. Patterns are matched against each file's path relative to the source directory being
```

docs/buildah-copy.1.md
```diff
@@ -77,6 +77,12 @@ is preserved.
 
 Path to an alternative .containerignore (.dockerignore) file. Requires \-\-contextdir be specified.
 
+**--include** *pattern*
+
+Only copy files matching the specific pattern, the complement of **--exclude**. Option can be specified
+multiple times. Patterns are matched against each file's path relaltive to the source directory being
+copied. See containerignore(5) for supported formats.
+
```
TomSweeneyRedHat, line 86: Apply add.1.md comments here too, please.

copier/copier.go
```diff
@@ -1092,6 +1110,28 @@ func pathIsExcluded(root, path string, pm *fileutils.PatternMatcher) (string, bo
...
+		return true, nil
+	}
+	if rel == "." {
+		// special case
+		return true, nil
+	}
+	// Matches uses filepath.FromSlash() to convert candidates before
+	// checking if they match the patterns it's been given, implying that
+	// it expects Unix-style paths.
+	matches, err := pm.Matches(filepath.ToSlash(rel)) //nolint:staticcheck
```
TomSweeneyRedHat: IDK, and I don't think it's an issue, but will ask.  Do we have to worry about a symbolic link here?  If someone set up a symbolic link to /etc/passwd or what have you, and the link was put in the include option, would it suck in the passwd info to the container?  Mostly concerning in a rootless environment.

ajoshua2004: links are handled in the same way as excludes so I don't believe we'd have to worry about links.

ajoshua2004:
@podman-container-tools/buildah-maintainers

copier/copier.go
```diff
@@ -405,6 +414,7 @@ type GetOptions struct {
 	Timestamp          *time.Time        // timestamp to force on all contents
 	DisallowWildcard   bool              // reject glob patterns in source paths
 	AllowEmptyWildcard bool              // don't error when glob patterns match nothing
+	Includes           []string          // contents to include, using the OS-specific path separator.
```
nalind: The comment should describe how this interacts with Excludes for the benefit of people who look at this module with `go doc -all`, and preferably not end in "." to match the comments on the other fields.
At the API level, the aim is to provide the semantics described at https://github.com/moby/buildkit/blob/master/solver/pb/ops.proto#L368.  Based on the test in `copy.bats`, I _think_ this gets it right.

docs/buildah-add.1.md
```diff
@@ -78,6 +78,12 @@ can be used.
 
 Path to an alternative .containerignore (.dockerignore) file. Requires \-\-contextdir be specified.
 
+**--include** *pattern*
+
+Only copy files matching the specific pattern, the complement of **--exclude**. The option
+can be specified multiple times. Patterns are matched against each file's path relative to
+the source directory being copied. See containerignore(5) for supported formats.
```
nalind: This should also describe how it interacts with `--exclude`, specifically cases where a given path matches both the include and exclude filters.

tests/add.bats
```diff
@@ -668,3 +668,41 @@ EOF
...
+
+  run_buildah_mount $cid
+  mnt=$output
+  run find $mnt -printf "%P\n"
+  filelist=$(LC_ALL=C sort <<<"$output")
+  run_buildah_umount $cid
+  expect_output --from="$filelist" "$expect" "add recursive include"
+
+  # else arm: a single named file that matches include copies cleanly
```
nalind: Not sure what the first part of this comment means.

tests/add.bats
```diff
@@ -668,3 +668,41 @@ EOF
...
+  run_buildah from $WITH_POLICY_JSON scratch
+  cid=$output
+  run_buildah add --include=**/*.go $cid ${mytest}/source.go /stuff2/
+
+  run_buildah_mount $cid
+  mnt=$output
+  run find $mnt -printf "%P\n"
+  filelist=$(LC_ALL=C sort <<<"$output")
+  run_buildah_umount $cid
+  expect_output --from="$filelist" --substring "source.go" "add include single file"
```
nalind, line 706: We'll want a check here to verify that the desired thing happens when both `--include` and `--exclude` are used and match one or more of the items being considered...

tests/copy.bats
```diff
@@ -807,3 +807,45 @@ parents/y/b.txt"
...
+
+  run_buildah from $WITH_POLICY_JSON scratch
+  cid=$output
+  run_buildah copy --include=**/*.go --exclude=**/nested.go $cid ${mytest} /stuff
+  run_buildah_mount $cid
+  mnt=$output
+  run find $mnt -printf "%P\n"
+  filelist=$(LC_ALL=C sort <<<"$output")
+  run_buildah_umount $cid
+  expect_output --from="$filelist" "$expect" "include with exclude"
```
nalind, line 850: ... like this.

ajoshua2004: added

ajoshua2004:
PTAL @podman-container-tools/buildah-maintainers

copier/copier.go
```diff
@@ -1611,6 +1651,16 @@ func copierHandlerGet(bulkWriter io.Writer, req request, pm *fileutils.PatternMa
 						// also be in the excludes list
 						return nil
 					}
+					if pmIncludes != nil && !d.IsDir() {
+						included, err := pathIsIncluded(item, path, pmIncludes)
```
Honny1: Why not use `req.Root` instead of `item`? It would match the exclusion.

docs/buildah-add.1.md
```diff
@@ -78,6 +78,13 @@ can be used.
 
 Path to an alternative .containerignore (.dockerignore) file. Requires \-\-contextdir be specified.
 
+**--include** *pattern*
+
+Only copy files matching the specified pattern. The option can be specified multiple times.
+Patterns are matched against each file's path relative to the source directory being copied.
```
Honny1:
```suggestion
Patterns are matched against each file's path relative to the context directory being copied.
```

docs/buildah-copy.1.md
```diff
@@ -77,6 +77,13 @@ is preserved.
 
 Path to an alternative .containerignore (.dockerignore) file. Requires \-\-contextdir be specified.
 
+**--include** *pattern*
+
+Only copy files matching the specified pattern. The option can be specified multiple times.
+Patterns are matched against each file's path relative to the source directory being copied.
```
Honny1: ditto

Honny1 (commented):
LGTM

ajoshua2004:
PTAL @podman-container-tools/buildah-maintainers

copier/copier.go
```diff
@@ -1140,6 +1158,28 @@ func pathIsExcluded(root, path string, pm *fileutils.PatternMatcher) (string, bo
...
+	if err != nil {
+		return false, fmt.Errorf("copier: error computing path of %q relative to root %q: %w", path, root, err)
+	}
+	if pm == nil {
+		return true, nil
+	}
+	if rel == "." {
+		// special case
+		return true, nil
+	}
```
nalind: This function returns the same values as `pathIsExcluded()` in the places where it's called, since it's not called for directories, or when `pm` is nil.

copier/copier.go
```diff
@@ -1072,11 +1082,19 @@ func copierHandler(bulkReader io.Reader, bulkWriter io.Writer, req request) (*re
...
 	if err != nil {
 		return nil, nil, fmt.Errorf("processing excludes list %v: %w", excludes, err)
 	}
 
+	var pmIncludes *fileutils.PatternMatcher
+	if includes := req.Includes(); len(includes) > 0 {
+		pmIncludes, err = fileutils.NewPatternMatcher(includes)
+		if err != nil {
+			return nil, nil, fmt.Errorf("processing includes list %v: %w", includes, err)
+		}
```
nalind, line 1095: What happens if a pattern starts with "!", causing `pmIncludes.Exclusions()` to return `true`?  Is that a valid input?

nalind (commented):
The work that one of the new helper functions does looks like it could be done by one that's already there, so some refactoring might be of use.  A question about whether exception patterns are handled correctly, if they're to be handled at all.

copier/copier.go
ajoshua2004: yeah its valid, a ! pattern in the includes list acts as an exception rather than an unconditional exclusion, I also added some tests for this scenario

copier/copier.go
ajoshua2004: yeah makes sense, I did it also for readability sake. I removed the new helper and renamed `pathIsExcluded` to `pathMatches` since its more neutral

Honny1:
@ajoshua2004 Can you please rebase on upstream main? Thanks

Honny1 (approved):
LGTM, Thanks!

@TomSweeneyRedHat or @nalind, can you please do a final review and merge? Thanks!

PR #6129, opened 2025-04-15
chroot: use symbolic names when complaining about mount() errors

nalind:
#### What type of PR is this?

/kind other

#### What this PR does / why we need it:

Translate flags passed to mount() and read back using statfs() from hex to named constants, to make troubleshooting a bit easier.

#### Which issue(s) this PR fixes:

None

#### Does this PR introduce a user-facing change?

```release-note
Certain errors encountered when calling `mount()` while using "chroot" isolation will now use mount flag names in error diagnostic messages.
```

PR #6989, opened 2026-07-24
Fix 32-bit build failure due to `statFlagMap` using int keys

Honny1:
Change `statFlagMap` from `map[int]string` to `map[uintptr]string` and remove int/uintptr casts in `statFlagNames()`, matching `mountFlagMap`. On 32-bit architectures, `unix.MS_NOUSER` (0x80000000) overflows signed int.

Regression from PR #6129.

Fixes this podman cross build failure:
```
# go.podman.io/buildah/chroot
Error: vendor/go.podman.io/buildah/chroot/run_linux.go:63:3: cannot use unix.MS_NOUSER (untyped int constant 2147483648) as int value in map literal (overflows)
make: *** [Makefile:512: bin/podman.cross.linux.arm] Error 1
Error: Process completed with exit code 2.
```
Link: https://github.com/podman-container-tools/podman/actions/runs/30037831733/job/89310129440?pr=29242

#### What type of PR is this?

> /kind api-change
> /kind bug
> /kind cleanup
> /kind deprecation
> /kind design
> /kind documentation
> /kind failing-test 
> /kind feature
> /kind flake
> /kind other

```release-note
None
```

### DIFF
--- a/chroot/run_linux.go
+++ b/chroot/run_linux.go
@@ -75,7 +75,7 @@ var (
 		unix.MS_SYNCHRONOUS:  "MS_SYNCHRONOUS",
 		unix.MS_UNBINDABLE:   "MS_UNBINDABLE",
 	}
-	statFlagMap = map[int]string{
+	statFlagMap = map[uintptr]string{
 		unix.ST_MANDLOCK:    "ST_MANDLOCK",
 		unix.ST_NOATIME:     "ST_NOATIME",
 		unix.ST_NODEV:       "ST_NODEV",
@@ -107,9 +107,9 @@ func statFlagNames(flags uintptr) []string {
 	var names []string
 	flags = flags & ^uintptr(0x20) // mask off ST_VALID
 	for flag, name := range statFlagMap {
-		if int(flags)&flag == flag {
+		if flags&flag == flag {
 			names = append(names, name)
-			flags = flags &^ (uintptr(flag))
+			flags = flags &^ flag
 		}
 	}
 	if flags != 0 { // got some unknown leftovers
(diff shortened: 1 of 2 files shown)


Luap99 (changes requested):
I am confused, we merged https://github.com/podman-container-tools/buildah/commit/40617943b8ee85a92da396ca52e28a8715f84087 already.

The error you posted shows MS_NOUSER which is part of mountFlagMap fixed there and this diff only touches statFlagMap.
Are you sure your branches were all rebased properly?

I guess switching this here would not be wrong either but it should not be needed for the error you posted

cc @nalind

Honny1:
I am supposed also with cross build failure in podman. Maybe I did something wrong with rebase or vendoring. But it occurred after vendoring container-libs.

Luap99:
@Honny1 yes see my comment on podman-dev, the vendoring you did there did downgrade buildah to 1.44.0 which does not carry the fix, you want to make sure it has the latest buildah on the branch as well. The go mod version selection is sometimes broken and picks an older version because it does not understand the branching rules we use.

nalind:
We check cross-compiles to linux/arm in CI, and it passed in https://github.com/podman-container-tools/buildah/actions/runs/29934820816/job/89006023462#step:4:16, so that's not a problem we're having on main.

Honny1:
@Luap99 @nalind It was a vendor issue on my side. Let me know if we still want this change, but I think we can close it.

nalind:
The `Statfs_t.Flags` field (if it's even defined) has different types on different platforms, even if we're just talking about Linux, so I lean toward just leaving this as-is for now.

PR #6996, opened 2026-07-26
honor containers.conf http_proxy and env settings during build

### LINKED ISSUE #29299: podman build in 6.0.x no longer honors `env` and `http_proxy` in `containers.conf`
### Issue Description

After upgrading Podman from `5.8.3` to `6.0.x`, `podman build` no longer honors the `http_proxy = false` directive and custom `env` settings configured in `containers.conf`. Instead, it always inherits host shell proxy environment variables (e.g. `http_proxy=http://127.0.0.1:1080`) unless the opposite option (`--http-proxy=false`) is explicitly passed on the CLI.

#### Setup & Context
I have a proxy service running on my local host machine, and in my shell environment I have:
```bash
export http_proxy=http://127.0.0.1:1080
```

Because I can't pass `127.0.0.1` to a container as a proxy environment variable (since `127.0.0.1` inside a container refers to the container's loopback interface where no proxy is running), I work around this by having the following `containers.conf` configuration:
```toml
[containers]
http_proxy = false
env = [
  "http_proxy=http://host.containers.internal:1080"
]
```

Before 6.0.x, the behavior between `podman run` and `podman build` was consistent with this `containers.conf` configuration. However, this is no longer the case in 6.0.x—`podman build` ignores `containers.conf`, causing build steps to inherit `127.0.0.1:1080` from the host shell and fail with connection refused errors.

1. Set up host shell proxy environment variable:
   ```bash
   export http_proxy=http://127.0.0.1:1080
   ```

2. Configure `~/.config/containers/containers.conf`:
   ```toml
   [containers]
   http_proxy = false
   env = [
     "http_proxy=http://host.containers.internal:1080"
   ]
   ```

3. Compare `podman run` vs `podman build`:
   ```bash
   # Test podman run:
   podman run --rm alpine env | grep -i proxy

   # Test podman build:
   podman build --no-cache -f - . << 'EOF'
   FROM alpine
   RUN env | grep -i proxy
   EOF
   ```

`podman run --rm alpine env | grep -i proxy` correctly prints:
```
http_proxy=http://host.containers.internal:1080
```

Whereas `podman build` prints:
```
http_proxy=http://127.0.0.1:1080
```

`podman run` honors `containers.conf` (`http_proxy = false` and `env = [...]`), properly setting `http://host.containers.internal:1080`.
In contrast, `podman build` ignores `containers.conf` and inherits `127.0.0.1:1080` directly from the host shell, causing network requests during build `RUN` steps (e.g. `RUN apk update`) to fail with connection refused errors against `127.0.0.1:1080`.

With `http_proxy = false` and custom proxy `env` configured in `containers.conf`, `podman build` should honor `http_proxy = false` to disable host `127.0.0.1` proxy passthrough, while applying `[containers] env` settings (`http_proxy=http://host.containers.internal:1080`) transiently during build `RUN` steps without leaking proxy environment variables into committed image metadata (`ENV` directives) (which was the behavior in 5.8.x and consistent with `podman run`).

### podman info output

```yaml
host:
  arch: amd64
  buildahVersion: 1.44.0
  cdiSpecDirs:
  - /etc/cdi
  - /var/run/cdi
  cgroupControllers:
  - cpu
  - memory
  - pids
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-1:2.2.1-1
    path: /usr/bin/conmon
    version: 'conmon version 2.2.1, commit: c8cc2c4db27531bd4e084ce7857f73cd21ee639d'
  cpuUtilization:
    idlePercent: 97.73
    systemPercent: 0.55
    userPercent: 1.72
  cpus: 32
  databaseBackend: sqlite
  discoveredDevices:
  - id: nvidia.com/gpu=0
    source: cdi
  - id: nvidia.com/gpu=GPU-...
    source: cdi
  - id: nvidia.com/gpu=all
    source: cdi
  distribution:
    distribution: arch
    version: unknown
  eventLogger: journald
  freeLocks: 2033
  hostname: solarch
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
    uidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 65536
  kernel: 7.1.4-arch1-1
  linkmode: dynamic
  logDriver: journald
  memFree: 10918256640
  memTotal: 67314135040
  networkBackend: netavark
  networkBackendInfo:
    backend: netavark
    defaultNetwork: podman
    dns:
      package: aardvark-dns-2.0.0-1
      path: /usr/lib/podman/aardvark-dns
      version: aardvark-dns 2.0.0
    package: netavark-2.0.0-1
    path: /usr/lib/podman/netavark
    version: netavark 2.0.0
  ociRuntime:
    name: crun
    package: crun-1.28-1
    path: /usr/bin/crun
    version: |-
      crun version 1.28-dirty
      commit: 54f16ffbefcd022bf032af768b5c5ce075c18bfc
      rundir: /run/user/1000/crun
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +LIBKRUN +JSON_C
  os: linux
  pasta:
    executable: /usr/bin/pasta
    package: passt-2026_07_16.090d739-1
    version: |
      pasta 2026_07_16.090d739
      Copyright Red Hat
  remoteSocket:
    exists: true
  rootlessNetworkCmd: pasta
  rootlessPortForwarder: rootlessport
  security:
    apparmorEnabled: false
    rootless: true
    seccompEnabled: true
    selinuxEnabled: false
  serviceIsRemote: false
  swapFree: 65449533440
  swapTotal: 67313332224
  uptime: 53h 56m 41.00s (Approximately 2.21 days)
  variant: ""
plugins:
  authorization: null
  log:
  - k8s-file
  - none
  - passthrough
  - journald
  network:
  - bridge
  - macvlan
  - ipvlan
  volume:
  - local
registries: {}
store:
  containerStore:
    number: 3
    paused: 0
    running: 2
    stopped: 1
  graphDriverName: overlay
  graphOptions:
    overlay.mountopt: nodev
  graphRoot: /home/sola/.local/share/containers/storage
  graphRootAllocated: 998038831104
  graphRootUsed: 491926798336
  graphStatus:
    Backing Filesystem: btrfs
    Native Overlay Diff: "true"
    Supports d_type: "true"
    Supports shifting: "false"
    Supports volatile: "true"
    Using metacopy: "false"
  imageCopyTmpDir: /var/tmp
  imageStore:
    number: 309
  transientStore: false
  volumePath: /home/sola/.local/share/containers/storage/volumes
version:
  APIVersion: 6.0.1
  Built: 1783707691
  BuiltTime: Sat Jul 11 02:21:31 2026
  GitCommit: 4cabbe61fa3a27fafc4a3ee1226e38ae1664ae57
  GoVersion: go1.26.5-X:nodwarf5
  Os: linux
  OsArch: linux/amd64
  Version: 6.0.1
```

### Podman in a container

No

### Privileged Or Rootless

Rootless

### Upstream Latest Release

Yes

_No response_

### Additional information

This see also https://github.com/podman-container-tools/podman/issues/27339 on the reason behind the workaround setup.

unlimitedsola:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it

This PR resolves an issue where `podman build` ignored `containers.conf` settings (`http_proxy = false` and `[containers] env`). Please also see the linked issue for more context.

1. Binds the `--http-proxy` CLI flag default in `pkg/cli/common.go` to `Containers.HTTPProxy` from `containers.conf` instead of hardcoding `true`. This makes build proxy flag defaults consistent with `podman run`, allowing `http_proxy = false` in `containers.conf` to disable host environment proxy passthrough by default.
2. Updates `configureEnvironment()` in `run_common.go` to inject environment variables defined under `[containers] env` in `containers.conf` into the transient container process spec (`g.AddProcessEnv`) during `RUN` steps. This fixes the regression from pre-6.0.x where `containers.conf` environment settings applied to build execution commands (e.g. package management and network requests).

#### How to verify it

1. Configure `containers.conf`:

   ```toml
   [containers]
   http_proxy = false
   env = [
     "http_proxy=http://host.containers.internal:1080",
   ]
   ```

2. Export host shell environment variables:

   ```bash
   export http_proxy=http://127.0.0.1:1080
   ```

3. Run build:

   ```bash
   podman build --no-cache -f - . << 'EOF'
   FROM alpine
   RUN env | grep -i proxy
   EOF
   ```

4. Verify output contains `http_proxy=http://host.containers.internal:1080` instead of `http://127.0.0.1:1080`.

#### Which issue(s) this PR fixes

Fixes: https://github.com/podman-container-tools/podman/issues/29299

#### Special notes for your reviewer

`g.AddProcessEnv` in `configureEnvironment()` sets the process spec environment (`g.Config.Process.Env`) for transient container execution during `RUN` steps. It does not touch `builder.OCIv1.Config.Env`, ensuring that environment variables configured in `containers.conf` apply to build execution without polluting committed image layers.

Not sure if a release note is needed for regressions, but this does contain user-facing change:

```release-note
Honor `containers.conf` `http_proxy` and `env` settings during image builds, aligning the behavior with `podman run` and fixed regression introduced in 6.0.x.
```

### DIFF
--- a/run_common.go
+++ b/run_common.go
@@ -311,8 +311,22 @@ func (b *Builder) configureUIDGID(g *generate.Generator, mountPoint string, opti
 	return homeDir, nil
 }
 
-func (b *Builder) configureEnvironment(g *generate.Generator, options RunOptions, defaultEnv []string) {
+func (b *Builder) configureEnvironment(g *generate.Generator, options RunOptions, defaultEnv []string) error {
 	g.ClearProcessEnv()
+
+	conf, err := config.Default()
+	if err != nil {
+		return fmt.Errorf("failed to get container config: %w", err)
+	}
+	for _, envSpec := range conf.Containers.Env.Get() {
+		env := strings.SplitN(envSpec, "=", 2)
+		// Only pass proxy environment variables configured in containers.conf into build RUN steps.
+		// Arbitrary environment variables from containers.conf are excluded to preserve build reproducibility
+		// and avoid subtle build-to-build differences across environments.
+		if len(env) > 1 && slices.Contains(config.ProxyEnv, env[0]) {
+			g.AddProcessEnv(env[0], env[1])
+		}
+	}
 
 	if b.CommonBuildOpts.HTTPProxy {
 		for _, envSpec := range config.ProxyEnv {
@@ -328,6 +342,7 @@ func (b *Builder) configureEnvironment(g *generate.Generator, options RunOptions
 			g.AddProcessEnv(env[0], env[1])
 		}
 	}
+	return nil
 }
 
 // getNetworkInterface creates the network interface
--- a/run_common_test.go
+++ b/run_common_test.go
@@ -2,11 +2,15 @@ package buildah
 
 import (
 	"errors"
+	"os"
 	"os/exec"
+	"path/filepath"
 	"testing"
 
+	"github.com/opencontainers/runtime-tools/generate"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.podman.io/common/pkg/config"
 )
 
 func TestMapContainerNameToHostname(t *testing.T) {
@@ -66,4 +70,90 @@ func TestCheckExitCodeError(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestBuildahProxyAndNoLeak(t *testing.T) {
+	tmpDir := t.TempDir()
+	confPath := filepath.Join(tmpDir, "containers.conf")
+	confContent := `
+[containers]
+http_proxy = false
+env = [
+  "http_proxy=http://host.containers.internal:1080",
+  "https_proxy=http://host.containers.internal:1080",
+  "CUSTOM_NON_PROXY_VAR=should_be_ignored"
+]
+`
+	err := os.WriteFile(confPath, []byte(confContent), 0o644)
+	require.NoError(t, err)
+
+	// Reset the default container config cache on test cleanup. Register before t.Setenv so that
+	// the cleanups caused by `t.Setenv` calls run before it (because cleanup happens in LIFO order).
+	t.Cleanup(func() { _, _ = config.New(&config.Options{SetDefault: true}) })
+	t.Setenv("CONTAINERS_CONF", confPath)
+	t.Setenv("http_proxy", "http://127.0.0.1:1080")
+
+	_, err = config.New(&config.Options{SetDefault: true})
+	require.NoError(t, err)
+
+	builder := &Builder{}
+	builder.CommonBuildOpts = &CommonBuildOptions{
+		HTTPProxy: false,
+	}
+
+	g, err := generate.New("linux")
+	require.NoError(t, err)
+
+	err = builder.configureEnvironment(&g, RunOptions{}, []string{"PATH=/usr/bin"})
+	require.NoError(t, err)
+
+	procEnv := g.Config.Process.Env
+	t.Logf("Transient process env during RUN: %v", procEnv)
+
+	assert.Contains(t, procEnv, "http_proxy=http://host.containers.internal:1080")
+	assert.NotContains(t, procEnv, "http_proxy=http://127.0.0.1:1080")
+	assert.NotContains(t, procEnv, "CUSTOM_NON_PROXY_VAR=should_be_ignored")
+
+	imageEnv := builder.OCIv1.Config.Env
+	assert.NotContains(t, imageEnv, "http_proxy=http://host.containers.internal:1080")
+}
+
+func TestBuildahProxyPrecedence(t *testing.T) {
+	tmpDir := t.TempDir()
+	confPath := filepath.Join(tmpDir, "containers.conf")
+	confContent := `
+[containers]
+http_proxy = false
+env = [
+  "http_proxy=http://host.containers.internal:1080"
+]
+`
+	err := os.WriteFile(confPath, []byte(confContent), 0o644)
+	require.NoError(t, err)
+
+	// Reset the default container config cache on test cleanup. Register before t.Setenv so that
+	// the cleanups caused by `t.Setenv` calls run before it (because cleanup happens in LIFO order).
+	t.Cleanup(func() { _, _ = config.New(&config.Options{SetDefault: true}) })
+	t.Setenv("CONTAINERS_CONF", confPath)
+	t.Setenv("http_proxy", "http://127.0.0.1:1080")
+
+	_, err = config.New(&config.Options{SetDefault: true})
+	require.NoError(t, err)
+
+	builder := &Builder{}
+
+	// Case 1: HTTPProxy: true (Host env overrides containers.conf)
+	builder.CommonBuildOpts = &CommonBuildOptions{HTTPProxy: true}
+	g1, err := generate.New("linux")
+	require.NoError(t, err)
+	err = builder.configureEnvironment(&g1, RunOptions{}, []string{"PATH=/usr/bin"})
+	require.NoError(t, err)
+	assert.Contains(t, g1.Config.Process.Env, "http_proxy=http://127.0.0.1:1080")
+
+	// Case 2: CLI Options.Env overrides both containers.conf and host env
+	g2, err := generate.New("linux")
+	require.NoError(t, err)
+	err = builder.configureEnvironment(&g2, RunOptions{Env: []string{"http_proxy=http://cli.override:1080"}}, []string{"PATH=/usr/bin"})
+	require.NoError(t, err)
+	assert.Contains(t, g2.Config.Process.Env, "http_proxy=http://cli.override:1080")
 }
--- a/run_freebsd.go
+++ b/run_freebsd.go
@@ -48,7 +48,8 @@ const (
 func procctl(idtype int, id int, cmd int, arg *byte) error {
 	_, _, e1 := unix.Syscall6(
 		unix.SYS_PROCCTL, uintptr(idtype), uintptr(id),
-		uintptr(cmd), uintptr(unsafe.Pointer(arg)), 0, 0)
+		uintptr(cmd), uintptr(unsafe.Pointer(arg)), 0, 0,
+	)
 	if e1 != 0 {
 		return unix.Errno(e1)
 	}
@@ -122,8 +123,11 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 		return err
 	}
 
-	// hardwire the environment to match docker build to avoid subtle and hard-to-debug differences due to containers.conf
-	b.configureEnvironment(g, options, []string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"})
+	// Pass a fixed environment to match docker build because passing the whole containers.conf instead would cause
+	// subtle and hard-to-debug differences. Proxy variables are the exception, as they are read from containers.conf in configureEnvironment.
+	if err := b.configureEnvironment(g, options, []string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"}); err != nil {
+		return err
+	}
 
 	if b.CommonBuildOpts == nil {
 		return fmt.Errorf("invalid format on container you must recreate the container")
--- a/run_linux.go
+++ b/run_linux.go
@@ -230,8 +230,11 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 		return err
 	}
 
-	// hardwire the environment to match docker build to avoid subtle and hard-to-debug differences due to containers.conf
-	b.configureEnvironment(g, options, []string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"})
+	// Pass a fixed environment to match docker build because passing the whole containers.conf instead would cause
+	// subtle and hard-to-debug differences. Proxy variables are the exception, as they are read from containers.conf in configureEnvironment.
+	if err := b.configureEnvironment(g, options, []string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"}); err != nil {
+		return err
+	}
 
 	if b.CommonBuildOpts == nil {
 		return fmt.Errorf("invalid format on container you must recreate the container")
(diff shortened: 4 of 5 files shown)


run_common.go
```diff
@@ -322,6 +322,15 @@ func (b *Builder) configureEnvironment(g *generate.Generator, options RunOptions
 		}
 	}
 
+	if conf, err := config.Default(); err == nil {
```
simonbrauner: I would suggest against silently ignoring the error, similarly to other calls of `config.Default` in the file.

```go
if err != nil {
	return nil, fmt.Errorf("failed to get container config: %w", err)
}
```

unlimitedsola: Thanks for the feedback! I changed the function signature so that the error can now bubble up instead of silently ignored.

run_common.go
```diff
@@ -322,12 +322,24 @@ func (b *Builder) configureEnvironment(g *generate.Generator, options RunOptions
...
+	conf, err := config.Default()
+	if err != nil {
+		return fmt.Errorf("failed to get container config: %w", err)
+	}
+	for _, envSpec := range conf.Containers.Env.Get() {
+		env := strings.SplitN(envSpec, "=", 2)
+		if len(env) > 1 {
+			g.AddProcessEnv(env[0], env[1])
+		}
+	}
```
nalind, line 329: This would silently alter what a RUN instruction in a build gets without leaving any trace, which can unintentionally interfere with reproducibility of builds.  There are already mechanisms for passing environment variables to builds, and outside of proxy settings, we really shouldn't be adding things.

unlimitedsola: Thanks for the feedback! Are you suggesting that I should inject only proxy environment variables from `containers.conf` here, or a proper warning should be displayed?

nalind: If we're not already passing proxy-related variables through, we definitely want to fix that.  As for other variables, I don't think we want to be pulling those in for users of `buildah build`.

unlimitedsola: I've added a filter to only pull proxy environment variables from `env =` in the `containers.conf` file. Please take another look.

nalind: Hmm, _where_ this happens makes it so that settings in the configuration file override values that we'd previously have picked up from the environment in which we're running, and that's not how we generally treat precedence.  Usually the sources that are easier to override for a given run are given higher precedence - CLI before environment, environment (if we look there) before configuration files, configuration files before hard-coded defaults.  Can this be merged with the logic that pulls values from the environment, with those taking precedence in cases where there's a conflict?

unlimitedsola: Good point! I've moved the `containers.conf` logic before the http-proxy flag check, so that explicit cli flag can override what's specified in the `containers.conf`. I've also added an extra test to verify this behavior.

nalind (approved):
LGTM, but I expect you'll need to run `make fmt` to fix the linter formatter warning.

TomSweeneyRedHat:
LGTM
once the gofmt issue on line 87 in the test is made happy

nalind:
Please squash your commits before merging.  Thanks!

unlimitedsola:
> Ephemeral COPR build failed. @containers/packit-build please check.

I've squashed and rebased my commits. I think this was caused by me force-pushed twice too quickly?

nalind (approved):

unlimitedsola:
The CI failure seems unrelated to the code change. Let me know if there is anything that I can help to move this forward.

run_freebsd.go
```diff
@@ -123,7 +123,9 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	}
 
 	// hardwire the environment to match docker build to avoid subtle and hard-to-debug differences due to containers.conf
```
simonbrauner: This comment implies that containers.conf won't be used at all, while the PR introduces the possibility of loading proxy variables from containers.conf.

I would propose updating it with something like this:
```suggestion
	// Pass a fixed environment to match docker build because passing the whole containers.conf instead would cause
    // subtle and hard-to-debug differences. Proxy variables are the exception, as they are read from containers.conf in configureEnvironment.
```

run_linux.go
```diff
@@ -230,7 +230,9 @@ func (b *Builder) Run(command []string, options RunOptions) error {
 	}
 
 	// hardwire the environment to match docker build to avoid subtle and hard-to-debug differences due to containers.conf
```
simonbrauner: Same as https://github.com/podman-container-tools/buildah/pull/6996/changes#r3758367668.

run_common_test.go
```diff
@@ -67,3 +71,83 @@ func TestCheckExitCodeError(t *testing.T) {
...
+http_proxy = false
+env = [
+  "http_proxy=http://host.containers.internal:1080",
+  "https_proxy=http://host.containers.internal:1080",
+  "CUSTOM_NON_PROXY_VAR=should_be_ignored"
+]
+`
+	err := os.WriteFile(confPath, []byte(confContent), 0o644)
+	require.NoError(t, err)
+
```
simonbrauner, line 89: As far as I can tell, the `t.Setenv` followed by `config.New(&config.Options{SetDefault: true})` overwrites the config cache there https://github.com/podman-container-tools/container-libs/blob/16d801f1d5b2e615f9eb11f34ce007e428a294a6/common/pkg/config/new.go#L93C1-L96C3 and does not revert the change, which could cause problems later in the tests.

I'd propose running `config.New(&config.Options{SetDefault: true})` on clean up (registered **before** the `t.Setenv`s because they call `Cleanup` internally and the clean up happens in the LIFO order), after the environment change of `t.Setenv` is restored already, so that we have the initial state again:

```suggestion

    t.Cleanup(func() { _, _ = config.New(&config.Options{SetDefault: true}) })

```

run_common_test.go
```diff
@@ -67,3 +71,83 @@ func TestCheckExitCodeError(t *testing.T) {
...
+	confContent := `
+[containers]
+http_proxy = false
+env = [
+  "http_proxy=http://host.containers.internal:1080"
+]
+`
+	err := os.WriteFile(confPath, []byte(confContent), 0o644)
+	require.NoError(t, err)
+
```
simonbrauner, line 133: Same as https://github.com/podman-container-tools/buildah/pull/6996/changes#r3758728121.

simonbrauner (commented):
I added a few more comments. Otherwise, LGTM

Don't worry about unrelated test failures, somebody who has the permissions can rerun the tests later.

run_freebsd.go
unlimitedsola: Appreciate your feedback! I've updated the comment as you suggested.

run_linux.go
unlimitedsola: also updated as suggested

run_common_test.go
unlimitedsola: Interesting, I wasn't aware that the config cache would persist across tests. I now registered the cleanup callback before the set envs and added comments to explain what it is for.

run_common_test.go
unlimitedsola: Added the same cleanup and comment as above.

run_common_test.go
simonbrauner: I'd improve the wording of the comment. It is not the cause that `t.Cleanup` before `t.Setenv` causes `t.Cleanup` run in LIFO order, but rather than cleanups always run in LIFO order, which is the reason why we need to register the reset function before `t.Setenv` register their own.

```go
// Reset the default container config cache on test cleanup. Register before t.Setenv so that
// the cleanups caused by `t.Setenv` calls run before it (because cleanup happens in LIFO order).
```

unlimitedsola: This is indeed less ambiguous. I've updated both comments as suggested.

simonbrauner:
My comments were addressed.

unlimitedsola:
Hi, I would like to see this gets merged. I've added additional reasoning behind the change in the original issue: https://github.com/podman-container-tools/podman/issues/29299#issuecomment-5421694667

simonbrauner:

simonbrauner (approved):

nalind:
(The current validation error should be fixed by a rebase - the validation is run in a Go 1.27 environment, and the version of the linter at the point this is currently based on doesn't work correctly with 1.27.)

Honny1:
@unlimitedsola Can you please rebase on upstream main? Thanks!

unlimitedsola:
@Honny1 Rebased

Honny1 (changes requested):
Code LGTM, Please update the man pages. The `docs/buildah-build.1.md` (and `docs/buildah-from.1.md`) still document `--http-proxy=true` as the default.

unlimitedsola:
> Code LGTM, Please update the man pages. The `docs/buildah-build.1.md` (and `docs/buildah-from.1.md`) still document `--http-proxy=true` as the default.

@Honny1 Could you suggest how should I update the doc? The `http-proxy` is still `true` by default behavior-wise if there is no `containers.conf` overriding it[^1]. I copied what's in `docs/buildah-from.1.md` so that it doesn't explicitly say `--http-proxy=true` to avoid potential confusion.

[^1]: https://github.com/podman-container-tools/container-libs/blob/d09b9caeb249e15d3df4434a00978aaea3cc2cdc/common/pkg/config/default.go#L226

docs/buildah-build.1.md
```diff
@@ -511,14 +511,16 @@ For the bind-mount conditions, only mounts explicitly requested by the caller vi
...
+**--http-proxy**
 
 By default proxy environment variables are passed into the container if set
-for the buildah process.  This can be disabled by setting the `--http-proxy`
+for the Buildah process.  This can be disabled by setting the `--http-proxy`
 option to `false`.  The environment variables passed in include `http_proxy`,
 `https_proxy`, `ftp_proxy`, `no_proxy`, and also the upper case versions of
 those.
 
+Defaults to `true`
```
Honny1:
```suggestion
The default for this option is taken from the `http_proxy` setting in the
`[containers]` table of containers.conf(5). If unset, that value defaults to
`true`.
```

Honny1 (commented):
I add a suggestion for what I meant. Sorry, I wrote that in the wrong direction.

docs/buildah-build.1.md
unlimitedsola: I've updated both docs as suggested. Do we also need to update https://github.com/podman-container-tools/podman/blob/8e3c572a85b4178c5f8b489901153df4bd46fd2a/docs/source/markdown/options/http-proxy.md?

Honny1: That would be great, but I am not sure about timing. Buildah, with this change, must be vendored to Podman.

unlimitedsola: I can prepare a PR for that once this is merged. And maybe let them decide when to merge it?

Honny1: Okay thanks.

Honny1 (approved):
LGTM

PR #7001, opened 2026-07-27
Copier: add Symlink() for creating symlinks with inline target

### LINKED ISSUE #6802: Implement `Op_File.Actions.FileAction_Symlink`
Tackles the `Op_File.Actions.FileAction_Symlink` item from #6732 

- Add `copier.Symlink()`, possibly just creating an in-memory buffer and feeding it to `copier.Put()`
- Add unit test to the copier package

ajoshua2004:
#### What type of PR is this?

> /kind feature

Adds `copier.Symlink()` function and `SymlinkOptions` struct to the copier package.
  The function creates a symlink by building an in-memory tar archive with a single
  symlink entry and passing it to `copier.Put()`, following the same pattern as
  `copier.Mkfile()`.

#### How to verify it
`go test -v -run TestSymlinkNoChroot ./copier/ -v`
`go test -v -run TestSymlinkChroot ./copier/ -v`

Closes #6802
Part of #6732 

```release-note
copier: add Symlink() for creating symlinks with inline target
```

### DIFF
--- a/copier/copier_unix_test.go
+++ b/copier/copier_unix_test.go
@@ -189,3 +189,13 @@ func TestPutCreateDestPathChroot(t *testing.T) {
 	defer func() { canChroot = couldChroot }()
 	testPutCreateDestPath(t)
 }
+
+func TestSymlinkChroot(t *testing.T) {
+	if uid != 0 {
+		t.Skip("chroot() requires root privileges, skipping")
+	}
+	couldChroot := canChroot
+	canChroot = true
+	testSymlink(t)
+	canChroot = couldChroot
+}
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -612,6 +612,54 @@ func Remove(root string, item string, options RemoveOptions) error {
 	return nil
 }
 
+// SymlinkOptions controls parts of Symlink()'s behavior.
+type SymlinkOptions struct {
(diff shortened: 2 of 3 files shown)


copier/copier_unix_test.go
```diff
@@ -189,3 +189,10 @@ func TestPutCreateDestPathChroot(t *testing.T) {
 	defer func() { canChroot = couldChroot }()
 	testPutCreateDestPath(t)
 }
+
+func TestSymlinkChroot(t *testing.T) {
+	couldChroot := canChroot
+	canChroot = true
+	testSymlink(t)
+	canChroot = couldChroot
+}
```
simonbrauner, line 201: Shouldn't there be the skip for non-root users?

```go
if uid != 0 {
	t.Skip("chroot() requires root privileges, skipping")
}
```

Because if I'm not mistaken, `chroot` would return `false, nil` here https://github.com/podman-container-tools/buildah/blob/dda8e65e8462d3ef22a1d224bb4a0613e6f0e1c2/copier/syscall_unix.go#L29 so the test would silently test the nonChroot behavior.

ajoshua2004: Yeah I forgot to add the uid check to the test but I just added it

nalind (approved):
LGTM

TomSweeneyRedHat (commented):
LGTM

PR #7007, opened 2026-07-28
Add context-aware versions of several functions and some methods

nalind:
#### What type of PR is this?

/kind other

Introduce versions of a number of functions and some methods which take a context.Context, to least try to provide proper cancellation support.

When copier.Get() returns an error, also write a non-NUL string after whatever it's already written, so that a tar.Reader that reads what it output will also flag the malformed header with an unexpected error.

In the CLI, pass our default global context down in places where we didn't before, connect it to an experimental timeout flag, and have it cancel on `os.Interrupt` signals, too.

#### How to verify it

New/updated unit tests!

None

```release-note
None
```

### DIFF
--- a/cmd/buildah/common.go
+++ b/cmd/buildah/common.go
@@ -5,6 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/signal"
+	"sync"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -157,10 +159,30 @@ func openImage(ctx context.Context, sc *types.SystemContext, store storage.Store
 	return builder, nil
 }
 
-// getContext returns a context.TODO
+// getContext returns a context that may have a timeout, and which cancels if
+// it receives os.Interrupt
 func getContext() context.Context {
-	return context.TODO()
+	ctx, _ := getContextWithCancel()
+	return ctx
 }
+
+func getContextCancel() context.CancelFunc {
+	_, cancel := getContextWithCancel()
+	return cancel
+}
+
+var getContextWithCancel = sync.OnceValues(func() (context.Context, context.CancelFunc) {
+	var ctx context.Context
+	var cancel1, cancel2 func()
+	if rootCmd.PersistentFlags().Changed("experimental-timeout") {
+		ctx, cancel1 = context.WithTimeout(context.Background(), globalFlagResults.ExperimentalTimeout)
+	} else {
+		ctx = context.Background()
+		cancel1 = func() {}
+	}
+	ctx, cancel2 = signal.NotifyContext(ctx, os.Interrupt)
+	return ctx, func() { cancel1(); cancel2() }
+})
 
 func getUserFlags() pflag.FlagSet {
 	fs := pflag.FlagSet{}
--- a/cmd/buildah/main.go
+++ b/cmd/buildah/main.go
@@ -9,6 +9,7 @@ import (
 	"runtime/pprof"
 	"strings"
 	"syscall"
+	"time"
 
 	ispecs "github.com/opencontainers/image-spec/specs-go"
 	rspecs "github.com/opencontainers/runtime-spec/specs-go"
@@ -58,6 +59,7 @@ type globalFlags struct {
 	MemoryProfile              string
 	UserShortNameAliasConfPath string
 	CgroupManager              string
+	ExperimentalTimeout        time.Duration
 }
 
 var rootCmd = &cobra.Command{
@@ -107,6 +109,7 @@ func mainInit() {
 	rootCmd.CompletionOptions.HiddenDefaultCmd = true
 	// rootCmd.TraverseChildren = true
 	rootCmd.Version = fmt.Sprintf("%s (image-spec %s, runtime-spec %s)", define.Version, ispecs.Version, rspecs.Version)
+	rootCmd.PersistentFlags().DurationVar(&globalFlagResults.ExperimentalTimeout, "experimental-timeout", 0, "global timeout") // maybe rename this at some point?
 	rootCmd.PersistentFlags().BoolVar(&globalFlagResults.Debug, "debug", false, "print debugging information")
 	// TODO Need to allow for environment variable
 	rootCmd.PersistentFlags().StringVar(&globalFlagResults.RegistriesConf, "registries-conf", "", "path to registries.conf file (not usually used)")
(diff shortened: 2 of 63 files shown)


cmd/buildah/common.go
```diff
@@ -157,9 +159,31 @@ func openImage(ctx context.Context, sc *types.SystemContext, store storage.Store
...
+	return ctx
+}
+
+func getContextCancel() context.CancelFunc {
+	_, cancel := getContextWithCancel()
+	return cancel
+}
+
+func getContextWithCancel() (context.Context, context.CancelFunc) {
+	return sync.OnceValues(func() (context.Context, context.CancelFunc) {
```
Luap99: This cannot work can it? Calling sync.OnceValues every single time means you get a new context every single time.
sync.OnceValues must either be assigned to a global variable here 
```
var getContextWithCancel =  sync.OnceValues(func() (context.Context, context.CancelFunc) {...
```

cmd/buildah/main.go
```diff
@@ -293,7 +299,9 @@ func main() {
 	// Hard code TMPDIR functions to use $TMPDIR or /var/tmp
 	os.Setenv("TMPDIR", parse.GetTempDir())
 
-	if err := rootCmd.Execute(); err != nil {
+	err := rootCmd.Execute()
```
Luap99, line 302: FWIW there is ExecuteContext though which sets the ctx field for the cmd though I guess that is not used anywhere so it does not matter really.

internal/ctxreader/ctxreader_test.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func TestNewCancelableReader(t *testing.T) {
+	var b bytes.Buffer
+	for range 1024 {
+		b.Write(make([]byte, 1024))
+	}
+	ctx, cancel := context.WithTimeout(t.Context(), 1*time.Nanosecond)
+	defer cancel()
+	time.Sleep(1 * time.Millisecond)
```
Luap99: you can avoid the sleep, just call cancel() without defer which should be enough

Luap99 (commented):
ack, thanks. I have not looked at all the API changes but I think the context creation is with oncevalue is broken like that

cmd/buildah/common.go
nalind: Oof, yeah, that would've been too easy to keep self-contained.  Reworking it.

internal/ctxreader/ctxreader_test.go
nalind: Changed it.

vtushar06:
@Luap99 once this lands and gets vendored I can check whether the prune --build flake in podman#28868 stops showing..

cmd/buildah/main.go
nalind: I think trying to pass the context to ExecuteContext() would create an order of initialization problem - the one we initialize is affected by the CLI flags, and they haven't been parsed yet at that point.

Luap99: I guess you are right, however having the sync.Once makes it very easy to have this being called anywhere before that as well and then not noticing the flag takes no effect. But I guess that should be possible to catch with tests?

nalind: Yes!  It turns out that called for more checking for cancellation at various places, but I added a test for at least one of them.

nalind:
Rebased and fixed some warnings that the "usestdlibvars" linter would warn about.

Luap99 (approved):
LGTM

nalind:
Reordered a couple of the commits to trigger CI.

nalind:
Rewrote godoc to make it clearer that the non-Context() versions of functions and methods that now have a Context() counterpart are wrappers, have them use context.TODO() instead of context.Background(), and tell "go fix" to inline them to make it easier to spot where code that vendors us could benefit from switching to the newer variations.  That last part is flagged by the "gocheckcompilerdirectives" linter, so I've disabled it and left a comment next to the version in the Makefile so that, hopefully, we'll be able to re-enable that when we update to a newer version of the linter.
Updated a couple of call sites that I'd missed without its help.

Luap99 (approved):
LGTM (still)

@Honny1 @TomSweeneyRedHat PTAL

Honny1 (approved):
LGTM

nalind:
Rebasing, updating the note about re-enabling "gocheckcompilerdirectives", because it looks like golangci-lint 2.13 accepts the "//go:fix" directives being introduced here.  Also added smaller patches to correct things that v2.13.1 flags.

nalind:
Today's release of Go 1.27 caused our validate workflow to switch to 1.27, and that required manually bumping the version of the linter to avoid making it panic, so we don't need to disable the "gocheckcompilerdirectives" linter any more.

nalind:
Broke the linter updates out into #7058, if we want to handle them on their own.

Honny1 (approved):
LGTM

Luap99 (approved):

copier/copier.go
```diff
@@ -2117,13 +2269,16 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
 			osRoot.Close()
 		}()
 		ignoredItems := make(map[string]struct{})
-		tr := tar.NewReader(bulkReader)
+		tr := tar.NewReader(ctxreader.NewCancelableReader(ctx, bulkReader))
 		hdr, err := tr.Next()
-		for err == nil {
+		for hdr != nil {
```
mtrmac, line 2274: @nalind was this an intentional change? AFAICT it can only make a difference if the reader returns `tar.ErrInsecurePath` …

copier/copier.go
```diff
@@ -2117,13 +2269,16 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
...
 			nameBeforeRenaming := hdr.Name
 			if len(hdr.Name) == 0 {
 				// no name -> ignore the entry
 				ignoredItems[nameBeforeRenaming] = struct{}{}
+				if err != nil {
+					break
```
mtrmac, line 2280: … but then this can only trigger on `tar.ErrInsecurePath` as well. I don’t know why we would, on that error, not abort immediately, but abort _later_.

copier/copier.go
```diff
@@ -2387,9 +2547,12 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
 				return fmt.Errorf("copier: put: error setting fflags on %q: %w", path, err)
 			}
 		nextHeader:
+			if err != nil {
```
mtrmac, line 2550: This also, AFAICS, only triggers on `tar.ErrInsecurePath`, if we get here via `goto nextHeader` (an ignored item, but we have already created a parent directory); otherwise the `lutimes` call above ensures `err == nil`.

copier/copier.go
nalind: This was attempting to avoid dropping things in cases where we got an EOF and a header at the same time, but I hadn't considered `ErrInsecurePath`.

mtrmac: AFAICT `tr.Next` can never return (non-`nil` header, `io.EOF`). Admittedly its documentation doesn’t actually say that.

nalind: I tend to use `go doc` on the command line and have failed to notice that it's been showing me the `archive/tar` from tarsplit for quite a while.

PR #7008, opened 2026-07-28
Add more flags to test helpers

nalind:
#### What type of PR is this?

/kind other

* Add a hidden "--blob-info-cache-dir" flag to our `add`, `build`, `copy`, `commit`, `from`, `pull`, and `push` commands.
* Panic if we can't hide the "--cdi-config-dir" flag for `run`, like we do for the rest of the flags that we hide.
* Correct the panicking message when we fail to hide the "--blob-cache" flag for `from` and `build`.
* Add CLI flags to the "copy" and "imgtype" helpers for setting the location of registry certificates and credentials, and a "--dest-force-compression-format" flag to catch up with the main binary.
* Use imgtype instead of pulling and running a skopeo image to inspect an image after pushing it to a test registry.

#### How to verify it

Updated integration test!

None

```release-note
None
```

### DIFF
--- a/cmd/buildah/addcopy.go
+++ b/cmd/buildah/addcopy.go
@@ -30,6 +30,7 @@ type addCopyResults struct {
 	contextdir         string
 	from               string
 	blobCache          string
+	blobInfoCache      string
 	decryptionKeys     []string
 	removeSignatures   bool
 	signaturePolicy    string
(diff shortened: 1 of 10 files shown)


Honny1 (approved):
LGTM

Honny1:
PTAL @ajoshua2004

ajoshua2004:
LGTM

PR #7010, opened 2026-07-29
Add PutOptions.AttemptUnpackDockerCompatibility

ajoshua2004:
#### What type of PR is this?

/kind api-change

Adds `AttemptUnpackDockerCompatibility` (bool) to `copier.PutOptions`. When enabled, the Put handler detects if a written regular file is a possibly-compressed archive (gzip/bzip2/xz/zstd) and extracts its contents in place instead of keeping the archive as-is.

#### How to verify it
`go test ./copier/ -run TestAttemptUnpackDockerCompatibilityNoChroot -v`
`go test ./copier/ -run TestAttemptUnpackDockerCompatibilityChroot -v`

Part of #6732

The detection reuses the existing isArchivePath() and compression.AutoDecompress() infrastructure already used on the Get side for ExpandArchives. The flag is not wired to any caller yet, it's available on PutOptions for callers to opt into when needed.

```release-note
None
```

### DIFF
--- a/copier/copier.go
+++ b/copier/copier.go
@@ -444,25 +444,26 @@ func Get(root string, directory string, options GetOptions, globs []string, bulk
 
 // PutOptions controls parts of Put()'s behavior.
 type PutOptions struct {
-	UIDMap, GIDMap       []idtools.IDMap    // map from containerIDs to hostIDs when writing contents to disk
-	DefaultDirOwner      *idtools.IDPair    // set ownership of implicitly-created directories, default is ChownDirs, or 0:0 if ChownDirs not set
-	DefaultDirMode       *os.FileMode       // set permissions on implicitly-created directories, default is Chmod or ChmodDirs, or 0755 if neither is set
-	Chmod                string             // set permissions in octal or symbolic notation. overrides ChmodDirs and ChmodFiles if set
-	ChownDirs            *idtools.IDPair    // set ownership of newly-created directories
-	ChmodDirs            *os.FileMode       // set permissions on newly-created directories
-	ChownFiles           *idtools.IDPair    // set ownership of newly-created files
-	ChmodFiles           *os.FileMode       // set permissions on newly-created files
-	StripSetuidBit       bool               // strip the setuid bit off of items being written
-	StripSetgidBit       bool               // strip the setgid bit off of items being written
-	StripStickyBit       bool               // strip the sticky bit off of items being written
-	StripXattrs          bool               // don't bother trying to set extended attributes of items being copied
-	IgnoreXattrErrors    bool               // ignore any errors encountered when attempting to set extended attributes
-	IgnoreDevices        bool               // ignore items which are character or block devices
-	NoOverwriteDirNonDir bool               // instead of quietly overwriting directories with non-directories, return an error
-	NoOverwriteNonDirDir bool               // instead of quietly overwriting non-directories with directories, return an error
-	Rename               map[string]string  // rename items with the specified names, or under the specified names
-	Timestamp            *time.Time         // override timestamps on all extracted content
-	CreateDestPath       types.OptionalBool // create the destination path if it doesn't already exist, default is true
+	UIDMap, GIDMap                   []idtools.IDMap    // map from containerIDs to hostIDs when writing contents to disk
+	DefaultDirOwner                  *idtools.IDPair    // set ownership of implicitly-created directories, default is ChownDirs, or 0:0 if ChownDirs not set
+	DefaultDirMode                   *os.FileMode       // set permissions on implicitly-created directories, default is Chmod or ChmodDirs, or 0755 if neither is set
+	Chmod                            string             // set permissions in octal or symbolic notation. overrides ChmodDirs and ChmodFiles if set
+	ChownDirs                        *idtools.IDPair    // set ownership of newly-created directories
+	ChmodDirs                        *os.FileMode       // set permissions on newly-created directories
+	ChownFiles                       *idtools.IDPair    // set ownership of newly-created files
+	ChmodFiles                       *os.FileMode       // set permissions on newly-created files
+	StripSetuidBit                   bool               // strip the setuid bit off of items being written
+	StripSetgidBit                   bool               // strip the setgid bit off of items being written
+	StripStickyBit                   bool               // strip the sticky bit off of items being written
+	StripXattrs                      bool               // don't bother trying to set extended attributes of items being copied
+	IgnoreXattrErrors                bool               // ignore any errors encountered when attempting to set extended attributes
+	IgnoreDevices                    bool               // ignore items which are character or block devices
+	NoOverwriteDirNonDir             bool               // instead of quietly overwriting directories with non-directories, return an error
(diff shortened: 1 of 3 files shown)


copier/copier.go
```diff
@@ -2187,6 +2188,46 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
...
+						entry, err := tr2.Next()
+						if err == io.EOF {
+							break
+						}
+						if err != nil {
+							rc.Close()
+							f.Close()
+							return fmt.Errorf("copier: put: error reading unpacked archive %q: %w", path, err)
+						}
+						entryPath := filepath.Join(destDir, entry.Name)
```
TomSweeneyRedHat: @nalind is this safe?  Could it write outside of it's destination?

copier/copier.go
```diff
@@ -2187,6 +2188,51 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
...
+						if err != nil {
+							rc.Close()
+							f.Close()
+							return fmt.Errorf("copier: put: error reading unpacked archive %q: %w", path, err)
+						}
+						entryPath := filepath.Join(destDir, cleanerReldirectory(filepath.FromSlash(entry.Name)))
+						if err := ensureDirectoryUnderRoot(filepath.Dir(entryPath)); err != nil {
+							rc.Close()
+							f.Close()
+							return fmt.Errorf("copier: Invalid path: %v", err)
```
Honny1:
```suggestion
							return fmt.Errorf("copier: put: Invalid path: %v", err)
```

copier/copier_unix_test.go
```diff
@@ -199,3 +199,13 @@ func TestSymlinkChroot(t *testing.T) {
 	testSymlink(t)
 	canChroot = couldChroot
 }
+
+func TestAttemptUnpackDockerCompatibilityChroot(t *testing.T) {
+	if uid != 0 {
+		t.Skip("chroot() requires root privileges, skipping")
+	}
+	couldChroot := canChroot
+	canChroot = false
```
Honny1: Should be here, true?

copier/copier.go
```diff
@@ -2187,6 +2188,51 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
...
+							err = os.MkdirAll(entryPath, os.FileMode(entry.Mode))
+						}
+						if err != nil {
+							rc.Close()
+							f.Close()
+							return fmt.Errorf("copier: put: error creating unpacked entry %q from archive %q: %w", entryPath, path, err)
+						}
+					}
+					rc.Close()
+					f.Close()
```
Honny1: Why not use `defer`?

ajoshua2004: defer here wouldn't be safe because this runs in a per header loop and the way success path ends is with the goto nextHeader rather than return so they would skip the defered runs

copier/copier.go
```diff
@@ -2187,6 +2188,51 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
...
+							f.Close()
+							return fmt.Errorf("copier: put: error reading unpacked archive %q: %w", path, err)
+						}
+						entryPath := filepath.Join(destDir, cleanerReldirectory(filepath.FromSlash(entry.Name)))
+						if err := ensureDirectoryUnderRoot(filepath.Dir(entryPath)); err != nil {
+							rc.Close()
+							f.Close()
+							return fmt.Errorf("copier: put: Invalid path: %v", err)
+						}
+						switch entry.Typeflag {
```
Honny1: What about the default case? Should that be ignored or treated as an error?

copier/copier.go
```diff
@@ -2187,6 +2188,51 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
...
+						case tar.TypeReg:
+							_, err = createFile(entryPath, tr2)
```
Honny1: Is the file created with correct ownership, permissions, etc?

copier/copier.go
```diff
@@ -2187,6 +2188,51 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
 				if err == nil && written != hdr.Size {
 					return fmt.Errorf("copier: put: error creating regular file %q: incorrect length (%d != %d)", path, written, hdr.Size)
 				}
+				if err == nil && req.PutOptions.AttemptUnpackDockerCompatibility && isArchivePath(path) {
```
Honny1: `isArchivePath` opens the file and reads it. I think it is a duplicate read.

copier/copier.go
```diff
@@ -2187,6 +2188,51 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
...
+						}
+						if err != nil {
+							rc.Close()
+							f.Close()
+							return fmt.Errorf("copier: put: error creating unpacked entry %q from archive %q: %w", entryPath, path, err)
+						}
+					}
+					rc.Close()
+					f.Close()
+					os.Remove(path)
```
Honny1: What about the error output of removal?

nalind (changes requested):
This doesn't handle entries in an archive that aren't regular files or directory entries, skipping over hard and symbolic links, device nodes, and such.  That won't do for the general case.
My suggestion would be to change the createFile() helper's second parameter to an `io.Reader` (it only ever passes it to io.Copy(), so that's fine), and before calling it, check if the bytes being handed to it are a (possibly-compressed) archive.  If they are, pass the data down to a second invocation of the same handler function.

Honny1 (commented):
Sorry, I forgot to submit my comments.

copier/copier.go
ajoshua2004: the default case was already there at the top level, I just ripped out the nested loop

copier/copier.go
ajoshua2004: nested entries go through the same handler as the top level ones now so they should be getting the right metadata

copier/copier.go
ajoshua2004: yeah I just noticed that, removed it

copier/copier.go
ajoshua2004: added a check here

copier/copier.go
```diff
@@ -2111,15 +2112,14 @@ func copierHandlerPut(bulkReader io.Reader, req request, idMappings *idtools.IDM
 			}
 		}()
 		ignoredItems := make(map[string]struct{})
-		tr := tar.NewReader(bulkReader)
-		hdr, err := tr.Next()
-		for err == nil {
+
+		var processEntry func(hdr *tar.Header, tr io.Reader, baseDirectory string) error
+		processEntry = func(hdr *tar.Header, tr io.Reader, baseDirectory string) error {
```
nalind, line 2117:
```suggestion
		processEntry := func(hdr *tar.Header, tr io.Reader, baseDirectory string) error {
```

nalind (commented):
It should be possible to handle this case without writing the contents to a temporary file first.

copier/copier.go
ajoshua2004: This has to stay since we are doing recursive calls, it errors if we change it to this

PR #7012, opened 2026-07-29
Fix a few format-specifier warnings

nalind:
#### What type of PR is this?

/kind bug

In a few places, we were using "%q" as a format specifier for numbers, and for enumerated types which didn't have a `String()` method.

#### How to verify it

Updated unit test!

None

```release-note
None
```


Luap99 (approved):
There is a linter for this but I guess %q is technically valid and thus not flagged?

LGTM

TomSweeneyRedHat:
LGTM

PR #7022, opened 2026-08-09
docs(login): note --tls-verify needs = for false

### LINKED ISSUE #6375: CLI Inconsistency With --tls-verify
It seems that the buildah login's `--tls-verify` flag requires an equals to pass an arumgent to wheras every other flag does not.

I.e. every other flag supports `--flag argument` without needing to specify  `--flag=argument`

locker95:
`--tls-verify` is a bool flag, so a bare `--tls-verify` means true and `--tls-verify false` is not a value assignment (the next token is treated as the registry). The examples already use `--tls-verify=false`; spelled that out next to the option so it is less surprising.

Fixes #6375

### DIFF
--- a/docs/buildah-login.1.md
+++ b/docs/buildah-login.1.md
@@ -56,15 +56,14 @@ Take the password from stdin
 
 **--tls-verify**
 
-Require HTTPS and verification of certificates when talking to container registries (default: true). If explicitly set to true,
-then TLS verification will be used. If set to false, then TLS verification will not be used. If not specified,
-TLS verification will be used unless the target registry is listed as an insecure registry in registries.conf.
-TLS verification cannot be used when talking to an insecure registry.
-
-This is a boolean flag. Use the `=` form when passing a value (for example
-`--tls-verify=false`). A bare `--tls-verify` is treated as true; a space-separated
-form such as `--tls-verify false` is not accepted (the word after the flag is
-parsed as a registry name).
+Require HTTPS and certificate verification when talking to container registries
+(default: true). As this is a boolean value, use the `=` form when passing a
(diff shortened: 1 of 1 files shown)


docs/buildah-login.1.md
```diff
@@ -61,6 +61,11 @@ then TLS verification will be used. If set to false, then TLS verification will
 TLS verification will be used unless the target registry is listed as an insecure registry in registries.conf.
 TLS verification cannot be used when talking to an insecure registry.
 
+This is a boolean flag. Use the `=` form when passing a value (for example
+`--tls-verify=false`). A bare `--tls-verify` is treated as true; a space-separated
+form such as `--tls-verify false` is not accepted (the word after the flag is
+parsed as a registry name).
+
```
TomSweeneyRedHat, line 67: Your thought is good, but I'd rather it be worked into the prior paragraph, as this is kind of a boolean+ option and `true` can operate differently depending on how it is defined.  Perhaps:

Require HTTPS and certificate verification when talking to container registries (default: true). As this is a boolean value, use the `=` form when passing a value, for example, `--tls-verify=false`.  

If explicitly set to true,
then TLS verification will be used. If set to false, then TLS verification will not be used. If not specified,
TLS verification will be used unless the target registry is listed as an insecure registry in registries.conf.
TLS verification cannot be used when talking to an insecure registry.  Note: a space-separated form such as `--tls-verify false` is not accepted (the word after the flag is
parsed as a registry name).

Honny1 (approved):
LGTM

Honny1:
PTAL and Merge please @TomSweeneyRedHat

PR #7029, opened 2026-08-12
setupSeccomp: avoid a cast that we don't need to do

nalind:
#### What type of PR is this?

/kind cleanup

We don't need to cast a slice of bytes to a string to pass to LoadProfile(), when there's a LoadProfileFromBytes() that we can call instead.

None

```release-note
None
```


Honny1 (approved):
LGTM

giuseppe (approved):
LGTM

PR #7035, opened 2026-08-14
parsevolume: fix escape guard and option normalization

vtushar06:
I was reading the volume parser and noticed the escape guard in `SplitStringWithColonEscape `checks `idx-1 > 0`, which excludes index 1, so a volume starting with an escaped colon gets split on it anyway:

```
\:foo:/dst  ->  ["\", "foo", "/dst"]   (expected [":foo", "/dst"])
```

Mid-string escapes are unaffected. I don't think this is reachable from the CLI today since the host path has to be absolute, so I am pitching it as correctness plus coverage, not a live bug.

While writing tests for it I also noticed Volume() discards the normalized slice ValidateVolumeOpts returns, so the cached/delegated options that function is documented to silently drop still end up in `mount.Options`, and with no options at all mount.Options comes back as [""] instead of empty.

so the package had no tests, so both commits bring their own here 
cc @lsm5

### DIFF
--- a/internal/parsevolume/parse.go
+++ b/internal/parsevolume/parse.go
@@ -63,14 +63,14 @@ func Volume(volume string) (specs.Mount, error) {
 	if err := parse.ValidateVolumeCtrDir(arr[1]); err != nil {
 		return mount, err
 	}
-	mountOptions := ""
+	var mountOpts []string
 	if len(arr) > 2 {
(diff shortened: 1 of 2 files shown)


Honny1 (approved):
LGTM

nalind (approved):
LGTM, thanks!

PR #7036, opened 2026-08-14
validate: run the helpers self-test

vtushar06:
tests/helpers.bash.t is a working self-test for the assert helpers - 31 checks, TAP output, exits nonzero on failure - but as far as I can tell nothing runs it: no Makefile target, no workflow, no mention in test_runner.sh. Its only commit is the one that added it in 2021.

This hooks it into `make validate`, which the CI validate job already runs, so a change that breaks the assert helpers gets caught before it confuses a real test run. podman recently did the same for its hack self-tests.

It runs in under a second and needs no built binaries. To be clear it only covers assert(), so this is just running what already exists, not adding coverage.


nalind (approved):
Looks good.  Thanks!

Honny1 (approved):
LGTM

Luap99 (approved):
LGTM

PR #7047, opened 2026-08-17
chrootuser: do not let one bad line end the lookup

vtushar06:
I was going through the `bufio.Scanner` error checks after Paul's sweep in podman and ended up in chrootuser, where there is a second thing happening first. `parseNextPasswd` returns nil for any line it cannot parse, and all six callers loop `for pwd != nil`, so one bad line hides every entry under it:

```
passwd = root / "+::::::" (NIS compat, 7 fields but no uid) / bob
names the lookup loop reached: [root]
scanner err: <nil>
```

bob is on line 3 and comes back as an unknown user. First commit makes both parsers skip lines they cannot read, which is what the group parser already does for comments.

Second commit is the scanner one. `Scan()` returns false at EOF and on error alike, so a failed read was reported as the user or group simply not existing.

I have not managed to hit this from the CLI, an image would need a `+` line or similar in /etc/passwd, so I am not claiming a live bug report behind it.

#### How to verify it

Both new tests fail on current main:

```
--- FAIL: TestParseNextPasswdSkipsBadLines (0.00s)
    user_test.go:71: got [root], want [root bob]
--- FAIL: TestParseNextGroupSkipsBadLines (0.00s)
    user_test.go:94: got [root], want [root wheel]
```

cc @nalind

### DIFF
--- a/pkg/chrootuser/user_test.go
+++ b/pkg/chrootuser/user_test.go
@@ -40,3 +40,48 @@ func TestParseNextGroup(t *testing.T) {
 	}
 	assert.Nil(t, parseNextGroup(rc))
 }
+
+// "+::::::" is an NIS compat entry, it has the right number of fields but no uid
+var testPasswdData = `root:x:0:0::/root:/bin/sh
++::::::
+notenoughfields:x:1
+bob:x:1000:1000::/home/bob:/bin/sh
+`
+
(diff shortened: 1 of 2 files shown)


nalind (approved):
LGTM, thanks!

Luap99 (approved):
LGTM

PR #7048, opened 2026-08-18
chrootuser: replace our home-grown read-from-chroot with os.OpenInRoot()

nalind:
#### What type of PR is this?

/kind cleanup

Instead of starting a child process that chroots and pipes the contents of a file to its stdout, use the new os.OpenInRoot() from the standard library, which is faster.

Split the list of users in a group and check for the user's name in the resulting slice instead of assuming that a substring match is enough.                              

#### How to verify it

More unit tests!

None

```release-note
None
```


pkg/chrootuser/user_internal.go
```diff
@@ -89,6 +25,23 @@ type lookupGroupEntry struct {
...
+		// Return an effectively-empty result, so that the caller will
+		// behave as though it read an empty file.
+		var pw *os.File
+		f, pw, err = os.Pipe()
+		if err != nil {
+			return nil, err
+		}
+		pw.Close()
+	}
+	return f, err
```
simonbrauner, line 42: Wouldn't it be better if the `OpenInRoot` error was at least logged? Because as it is now, the caller will either see the pipe error, which is confusing when the function is not about opening pipes at all, or they will get the `nil` from `Pipe`, which overwrites the original error. Neither seems convenient for debugging.

pkg/chrootuser/user_internal.go
```diff
@@ -89,6 +25,23 @@ type lookupGroupEntry struct {
 	user string
 }
 
+func openChrootedFile(rootdir, path string) (*os.File, error) {
```
simonbrauner, line 28: Maybe the `openChrootedFile` should document that paths with a beginning slash no longer work? Because this could be an easy mistake to make, especially when the error is not returned.

pkg/chrootuser/user_test.go
```diff
@@ -2,12 +2,190 @@ package chrootuser
...
+	assert.NoError(t, err, `GetUser("1")`)
+	assert.Equal(t, uint32(1), uid, "uid for undefined user")
+	assert.Equal(t, uint32(0), gid, "primary gid for undefined user")
+	assert.Contains(t, []string{"", "/"}, homedir, "home directory for undefined user")
+}
+
+func TestLookupSymlinkOut(t *testing.T) {
+	root := t.TempDir()
+	require.NoError(t, os.Mkdir(filepath.Join(root, "etc"), 0o700))
+	require.NoError(t, os.Symlink("/etc/passwd", filepath.Join(root, "etc", "passwd")))
```
Luap99, line 45: this needs a second case where etc is a symlink to ensure this is really sure against breakouts for all path components

pkg/chrootuser/user_internal.go
```diff
@@ -89,6 +25,23 @@ type lookupGroupEntry struct {
...
+		// Ignore basically all errors, since we didn't check the exit
+		// status when we used to use a subprocess.
+		// Return an effectively-empty result, so that the caller will
+		// behave as though it read an empty file.
+		var pw *os.File
+		f, pw, err = os.Pipe()
+		if err != nil {
+			return nil, err
+		}
+		pw.Close()
```
Luap99, line 40: That feels extremely strange. IMO just let it return the error which makes a lot more sense error message wise than behaving like an empty file and failing as it wont find the user of course

Luap99 (commented):
cc @mtrmac

pkg/chrootuser/user_internal.go
mtrmac, line 40: Separate from the question whether this should exist: the function could return a `ReadCloser`, and then this can be a `io.NopCloser(bytes.NewReader([]byte{}))`, hopefully lighter-weight.

pkg/chrootuser/user_internal.go
```diff
@@ -1,80 +1,16 @@
...
-	}
-	err = cmd.Start()
-	if err != nil {
-		return nil, nil, err
-	}
-	// Hand back the child's stdout for reading, and the child to reap.
-	return cmd, stdout, nil
-}
-
 var lookupUser, lookupGroup sync.Mutex
```
mtrmac, line 14: What do these things protect? It’s undocumented, AFAICS it is a remnant from the times when this code was calling stateful libc APIs.

pkg/chrootuser/user_test.go
```diff
@@ -2,12 +2,190 @@ package chrootuser
...
+	assert.Equal(t, uint32(0), uid, "uid for 0:bin")
+	assert.Equal(t, uint32(1), gid, "primary gid for 0:bin")
+	assert.Contains(t, []string{"", "/"}, homedir, "home directory for 0:bin")
+}
+
+func TestLookupValidWithGroups(t *testing.T) {
+	root := t.TempDir()
+	require.NoError(t, os.Mkdir(filepath.Join(root, "etc"), 0o700))
+	require.NoError(t, os.WriteFile(filepath.Join(root, "etc", "passwd"), []byte("bin:*:1:1:Super Duper User:/bin:/bin/sh"), 0o644))
+	require.NoError(t, os.WriteFile(filepath.Join(root, "etc", "group"), []byte("bin:*:1:\nother:*:2:bin\nanother:*:3:bin\n"), 0o644))
```
lsm5, line 162: Should there be additional cases like:

1. Groups where "bin" appears as a substring of a member name (should NOT match):
   - robin:*:4:robin - "bin" is a substring of member "robin"
   - staff:*:5:foo,robin,bar - "bin" is a substring within the comma-separated list
2. Groups with comma-separated member lists where "bin" IS an exact member (SHOULD match):
   - admins:*:6:alice,bin,bob - verify comma-splitting works correctly

pkg/chrootuser/user_test.go
```diff
@@ -2,12 +2,190 @@ package chrootuser
...
+	assert.Error(t, err, `GetGroup("bin")`)
+	assert.Empty(t, group, "information for group bin")
+
+	group, err = GetGroup(root, "1")
+	assert.NoError(t, err, `GetGroup("1")`)
+	assert.Equal(t, uint32(1), group, "information for group bin")
+
+	uid, gid, homedir, err := GetUser(root, "bin")
+	assert.Error(t, err, `GetUser("bin")`)
+	assert.Empty(t, uid, "uid for undefined user")
```
mtrmac, line 31: `Empty` for an integer does work but it is rather unintuitive.

Applies throughout.

---

Really… should we be promising _anything_ about the return values on error? The function doesn’t document anything. I think removing these assertions is simplest.

Fine, assuming the tests are intended to ensure stable / reliable behavior even when not documented for the callers, it’s not obvious to me why returning `0`, an _otherwise-valid_ value, is desirable; I’d think a clearly-invalid -1 might be more defensive (but, of course, that’s guesswork, depending on what the caller does with the value).

[… but then again, the only caller seems to be `userForRun`, which _does_ use the other return values regardless of `err`; at least its caller does then ignore them. That’s… more complexity than I’d personally like.]

pkg/chrootuser/user_test.go
mtrmac: … and, I think, where either one of `etc` or `etc/passwd` are _valid_ symlinks within the container, testing that the symlink is followed and data is read correctly. Ideally all three of:
- Simple symlink to an item in the same directory (`passwd` → `passwd2`)
- Absolute symlink (interpreted within the root)
- Relative symlink, passing through `/../…` (interpreted within the root)

Under that semantics, this test should be almost an equivalent to `TestLookupCircularSymlink`; not a test of an escape.

mtrmac (commented):
`OpenInRoot` does not have container semantics: a symlink from `/etc/passwd` to `../../../../../../looks-escaping/passwd` will not be interpreted as `/looks-escaping/passwd`, a symlink to `/an/absolute/path` will be entirely rejected. Isn’t that a major problem for the callers?

pkg/chrootuser/user_internal.go
mtrmac: Yes… _almost all_ callers fail with “entry not found” anyway… but there is `GetAdditionalGroupsForUser`. The caller might need adjusting if all of this is expected to work with `RunOptions.User` containing numerical IDs while there are no `/etc` files inside the container.

pkg/chrootuser/user_test.go
```diff
@@ -2,12 +2,190 @@ package chrootuser
...
+	assert.Contains(t, []string{"", "/"}, homedir, "home directory for undefined user")
+
+	uid, gid, homedir, err = GetUser(root, "1")
+	assert.NoError(t, err, `GetUser("1")`)
+	assert.Equal(t, uint32(1), uid, "uid for undefined user")
+	assert.Equal(t, uint32(0), gid, "primary gid for undefined user")
+	assert.Contains(t, []string{"", "/"}, homedir, "home directory for undefined user")
+}
+
+func TestLookupSymlinkOut(t *testing.T) {
```
simonbrauner, line 42: If I'm not mistaken, there is a regression which this test does not cover.

This test passes there not because `OpenInRoot` would detect an attempt to get out of root, but because `OpenInRoot` refuses all absolute symlinks.

And the same test would pass on main as well because on main, there is the semantics of `absolute symlinks are not rejected, but resolved within the chroot`. Therefore, this would create a circular link which also satisfies the `assert.Error`.

Let's consider a test like this:

```go
func TestLookupViaAbsoluteSymlink(t *testing.T) {
	root := t.TempDir()
	require.NoError(t, os.MkdirAll(filepath.Join(root, "usr", "etc"), 0o755))
	require.NoError(t, os.WriteFile(filepath.Join(root, "usr", "etc", "passwd"),
		[]byte("bin:*:1:1:Super Duper User:/bin:/bin/sh"), 0o644))
	require.NoError(t, os.WriteFile(filepath.Join(root, "usr", "etc", "group"),
		[]byte("bin:*:1:"), 0o644))
	require.NoError(t, os.Symlink("/usr/etc", filepath.Join(root, "etc")))

	_, _, _, err := GetUser(root, "bin")
	assert.NoError(t, err, `GetUser("bin")`)
}
```

It would fail on this PR because the link is absolute. And this could be interpreted as an attempt to escape to the `/usr/etc` outside of the chroot. But if I take the test and add it to upstream main:

```go
// When adding this
func TestMain(m *testing.M) {
	if reexec.Init() {
		return
	}
	os.Exit(m.Run())
}

// and ran as `unshare -r go test ./pkg/chrootuser/ -run TestLookupViaAbsoluteSymlink --count=1`
```

It passes here, because the absolute symlink is not interpreted as `from the system root`, but rather as `from the chroot`.

simonbrauner (commented):
I added a few comments about usability and limitations of `os.OpenInRoot`

